PR #654, opened 2019-03-10
[WASM] Make setStdout arguments more explicit in what they print to

coolreader18:
So a string doesn't mean a CSS selector anymore it means a string enum of, for now, just "console", and `undefined` means nothing, and an `HTMLTextAreaElement` is a textarea console to print to.


palaviv:
You should probably update the documentation in the demo.

palaviv:
This is a great change. I actually needed it at some point when I added the interactive shell :). Why does it have to be a `textarea`?

coolreader18:
That's what we use for the demo, and I think that if you're going to set stdout to an HTML element, it should probably be a textarea due to the way it's displayed. I might add support for other HTML elements just using textContent, though.

palaviv:
> That's what we use for the demo, and I think that if you're going to set stdout to an HTML element, it should probably be a textarea due to the way it's displayed. I might add support for other HTML elements just using textContent, though.

I actually think that we should keep only the function option. The `textarea` is too specific and can be done simply by using a js function.

coolreader18:
Alright, that's sort of vestigial anyway from when the wasm crate was specific to just the demo.

coolreader18:
Oh, and I'll update the docs in a separate PR, cause there's a lot that's changed since they were written.

PR #655, opened 2019-03-10
Use a generator for the mandelbrot demo snippet

coolreader18:
I added a yield in the inner `x` loop, which removes *all* stuttering during the mandelbrot generation, but I'm also concerned that it may be too slow. Should it yield conditionally, or is it fine with one character at a time?

### DIFF
--- a/wasm/demo/snippets/mandelbrot.py
+++ b/wasm/demo/snippets/mandelbrot.py
@@ -1,4 +1,7 @@
-from browser import request_animation_frame
+try:
+    from browser import request_animation_frame
+except:
+    def request_animation_frame(cb): cb()
 
 w = 50.0
 h = 50.0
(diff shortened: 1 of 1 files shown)


wasm/demo/snippets/mandelbrot.py
```diff
@@ -3,36 +3,39 @@
...
+                print('·', end='')
+
+            x += 1
+            yield
+
+        print()
+        y += 1
+        yield
+
+gen = mandel()
```
windelbouwman: Could we make this snippet such that it also works with cpython and rustpython on host? Maybe do a check on import of request_animation_frame. I wanted to use mandelbrot on PC, but it failed.

coolreader18: I think I fixed it, it segfaults for me but I think that's just because I'm testing it on my phone.

coolreader18: Oh, it might have been a stack overflow. Fixed now, anyway.

PR #660, opened 2019-03-11
Typetype clean

cthulahoops:


vm/src/obj/objtype.rs
```diff
@@ -49,33 +115,19 @@ pub fn init(ctx: &PyContext) {
...
-fn _mro(cls: PyObjectRef) -> Option<Vec<PyObjectRef>> {
-    if let Some(PyClass { ref mro, .. }) = cls.payload::<PyClass>() {
-        let mut mro = mro.clone();
-        mro.insert(0, cls.clone());
-        Some(mro)
-    } else {
-        None
-    }
+fn _mro(cls: &PyClassRef) -> Vec<PyObjectRef> {
+    cls.iter_mro().map(|x| x.clone()).collect()
```
adrian17: `.cloned()`?

windelbouwman: Yes, indeed, you can use here `cls.iter_mro().cloned()`. The clippy tool will hint at this.

cthulahoops: Aha! Thanks. It hadn't fully sunk in what ```cloned``` is yet.

vm/src/pyobject.rs
```diff
@@ -1548,6 +1551,19 @@ pub trait PyValue: Any + fmt::Debug {
     fn required_type(ctx: &PyContext) -> PyObjectRef;
 }
 
+impl FromPyObjectRef for PyRef<PyClass> {
```
OddCoincidence: I was planning on deleting `FromPyObjectRef` since it overlaps considerably with `TryFromObject` - could you use the latter instead? Should be more or less the same, just with the unwrap/panic on the caller instead (which we will eventually be able to get rid of).

coolreader18: Get rid of by using `!`?

OddCoincidence: No, from the trait's perspective, it _can_ fail. I was referring to the fact that the `mro` could be a `Vec` of `PyClassRef` rather than `PyObjectRef` so that no downcasting is necessary.

cthulahoops: I thought this might be the controversial bit.

I tried with TryFromObject but it required me to pass the VirtualMachine into a bunch of code that doesn't currently have a VM. The panic should never happen as an object's type should always be a type. I'd be happy with an interface that doesn't require VM but gives me an option/result to deal with?

Will I pass the VM through? 

I suppose the alternative would be to make PyObject.typ a PyClassRef, but that's a pretty invasive change.

OddCoincidence: Oh I forget it takes a `vm` - we can work this out later then.

> I'd be happy with an interface that doesn't require VM but gives me an option/result to deal with?

This will exist soon! (`PyObjectRef::downcast` in my prototype)

> I suppose the alternative would be to make PyObject.typ a PyClassRef, but that's a pretty invasive change.

It is but I definitely think we should do this at some point.

cthulahoops: Okay, I'll merge. Once we have PyClassRef everywhere this will go away anyway.

PR #665, opened 2019-03-12
Type mro

cthulahoops:
Fix this:

```
>>>>> type.__mro__
TypeError: Expected type <class 'type'>, not <class 'NoneType'>
```

### DIFF
--- a/tests/snippets/mro.py
+++ b/tests/snippets/mro.py
@@ -18,3 +18,5 @@ class C(A, B):
     pass
 
 assert (C, A, B, X, Y, object) == C.__mro__
+
+assert type.__mro__ == (type, object)
--- a/vm/src/obj/objproperty.rs
+++ b/vm/src/obj/objproperty.rs
@@ -119,7 +119,7 @@ impl<'a, T> PropertyBuilder<'a, T> {
         }
     }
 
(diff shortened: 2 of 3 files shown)


cthulahoops:
Took me a while to figure out why I couldn't get the property to type check.

@skinny121 I think this is a bug in the PropertyBuilder?

skinny121:
I wanted to make it typesafe so the type of the value returned from 'get' must equal the type 'set' accepted. But it seems to just gets in the way.

cthulahoops:
I thought I'd tried that. (I blame the error messages.)

I can get it to work by making the getter return PyTupleRef instead of PyTuple (adds some noise to the getter). I'd like to just have the setter take PyTuple but get blocked by: 

```
the trait `pyobject::TryFromObject` is not implemented for `obj::objtuple::PyTuple`
```

So, I guess implementing that would work.

I think we want ```PyObjectRef``` in this case anyway, as we want a TypeError complaining about trying to set a read-only attribute, not a TypeError about trying to set it to the wrong type.

windelbouwman:
Strange change, looks like a hack around some other issue in the property factory. We need to take a look at this.

PR #666, opened 2019-03-12
Pyclassref protocols

cthulahoops:
Further moves towards using PyClassRef for types more broadly.

It's bringing in more use of the FromPyObjectRef trait that we want to drop, but they I think they should all start to go away in the next phase.

One thing (I think) we're missing is ```PyTuple<PyClassRef>``` so we can handle conversion of PyClassRef inside other structures.

### DIFF
--- a/vm/src/obj/objtype.rs
+++ b/vm/src/obj/objtype.rs
@@ -15,7 +15,7 @@ use super::objtuple::PyTuple;
 #[derive(Clone, Debug)]
 pub struct PyClass {
     pub name: String,
-    pub mro: Vec<PyObjectRef>,
+    pub mro: Vec<PyClassRef>,
 }
 
 pub type PyClassRef = PyRef<PyClass>;
@@ -48,13 +48,13 @@ struct IterMro<'a> {
 }
 
 impl<'a> Iterator for IterMro<'a> {
-    type Item = &'a PyObjectRef;
+    type Item = &'a PyClassRef;
 
     fn next(&mut self) -> Option<Self::Item> {
         match self.offset {
             None => {
                 self.offset = Some(0);
-                Some(&self.cls.as_object())
+                Some(&self.cls)
             }
             Some(offset) => {
                 if offset < self.cls.mro.len() {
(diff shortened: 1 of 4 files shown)


OddCoincidence:
> One thing (I think) we're missing is PyTuple<PyClassRef> so we can handle conversion of PyClassRef inside other structures.

I had a similar thought for `PyList`, though for `PyTuple` I wonder if we'd want to attempt to map it to rust tuples, e.g. `PyTuple<(PyIntRef, PyStringRef)> given that heterogeneous types are much more common / likely (though I have doubts that this can be implemented nicely in rust today).

vm/src/obj/objtype.rs
```diff
@@ -26,19 +26,35 @@ impl PyValue for PyClass {
     }
 }
 
+impl IdProtocol for PyClassRef {
```
OddCoincidence: Note: in the future we shouldn't need this impl, once `PyObjectRef` is just an alias for `PyRef<dyn PyValue>` we can implement it broadly for all refs.

vm/src/obj/objtype.rs
```diff
@@ -26,19 +26,35 @@ impl PyValue for PyClass {
...
+}
+
 struct IterMro<'a> {
     cls: &'a PyClassRef,
     offset: Option<usize>,
 }
 
 impl<'a> Iterator for IterMro<'a> {
-    type Item = &'a PyObjectRef;
+    type Item = &'a PyClassRef;
```
OddCoincidence: What's the reason for holding `PyClassRef` by reference? Is it just to avoid an extra inc and dec of the ref count?

cthulahoops: Yes, pretty much.

It wasn't a very deliberate decision. My initial work here tended towards just clone everything all the time, and now I'm trying to just take references whenever I can.

cthulahoops:
> I had a similar thought for `PyList`, though for `PyTuple` I wonder if we'd want to attempt to map it to 
> rust tuples, e.g. `PyTuple<(PyIntRef, PyStringRef)>` given that heterogeneous types are much more
> common / likely (though I have doubts that this can be implemented nicely in rust today).

We probably do want to both for tuple though (if possible), as python tends to care more about mutability/hashability than the hetrerogenous/homogenous distinct.

vm/src/obj/objtype.rs
OddCoincidence: Not a big deal, but personally I'd prefer either `&PyClass` or to just clone the `PyClassRef` (which is pretty cheap anyway). Shared refs to reference counted types just seem kind of weird.

PR #667, opened 2019-03-12
Remove Option from PyObject.typ; Refactor type hierarchy initialization.

adrian17:
Also drops Default for PyObject.


OddCoincidence:
This looks great, thanks!

PR #668, opened 2019-03-12
Use default opt-level for release builds.

adrian17:
This makes the release build noticeably faster (~50% faster on random microbenchmarks) while increasing the binary size slightly (less than 10% on x86).
It's a question to wasm guys on whether it's acceptable (as AFAIK this option was added specifically for wasm). If not, a per-target config would be needed.


coolreader18:
I think it's fine for WASM, as if you're including a full Python interpreter on your web page, a 50% performance boost is more important than a couple hundred more kilobytes.

PR #670, opened 2019-03-12
Add PyObjectPayload trait and use it for PyObject.payload

### LINKED ISSUE #664: Debug implementation changed: Trace output has become a lot less useful.
Eg.:

TRACE 2019-03-12T10:29:07Z: rustpython_vm::vm: Invoke: [PyObj Any] PyFuncArgs { args: [[PyObj Any], [PyObj Any]], kwargs: [] }
TRACE 2019-03-12T10:29:07Z: rustpython_vm::vm: vm.call_method [PyObj Any] [PyObj Any] "__subclasscheck__" -> [PyObj Any]
TRACE 2019-03-12T10:29:07Z: rustpython_vm::vm: Invoke: [PyObj Any] PyFuncArgs { args: [[PyObj Any], [PyObj Any], [PyObj Any]], kwargs: [] }

coolreader18:
I'm not sure about the name; I'm willing to change it to pretty much anything but I can't think of anything else right now. The main goal of this is to have a good Debug implementation for PyObject.payload, as discussed in #664.

Edit: this would close #664.


vm/src/pyobject.rs
```diff
@@ -1562,16 +1558,31 @@ impl PyObject {
...
 }
 
-// The intention is for this to replace `PyObjectPayload` once everything is
-// converted to use `PyObjectPayload::AnyRustvalue`.
-pub trait PyValue: Any + fmt::Debug {
+pub trait PyValue: fmt::Debug + 'static {
     fn required_type(ctx: &PyContext) -> PyObjectRef;
 }
 
+pub trait PyValuePayload: Any + fmt::Debug + 'static {
```
OddCoincidence: For the name, how about `PyObjectPayload` - @cthulahoops and @windelbouwman both preferred that, and since this separates it from `PyValue`, it resolves my concerns about `PyObjectPayload` being an unintuitive name from an API perspective.

coolreader18: Sounds good!

vm/src/pyobject.rs
```diff
@@ -1562,16 +1558,31 @@ impl PyObject {
...
+    fn required_type(&self, ctx: &PyContext) -> PyObjectRef;
```
OddCoincidence: I don't think this method is needed. This should basically be a marker trait.

vm/src/pyobject.rs
```diff
@@ -1562,16 +1558,31 @@ impl PyObject {
...
+}
+
+impl<T: PyValue + 'static> PyValuePayload for T {
+    fn required_type(&self, ctx: &PyContext) -> PyObjectRef {
+        T::required_type(ctx)
+    }
+}
+
+impl PyValuePayload for () {
```
OddCoincidence: What's this impl needed for?

vm/src/obj/objobject.rs
```diff
@@ -4,13 +4,19 @@ use super::objtype;
...
+    PyRef, PyResult, PyValue, TypeProtocol,
 };
 use crate::vm::VirtualMachine;
 
 #[derive(Clone, Debug)]
 pub struct PyInstance;
 
+impl PyValue for PyInstance {
+    fn required_type(_ctx: &PyContext) -> PyObjectRef {
+        panic!("no specific type for PyInstance, don't type check me")
```
OddCoincidence: Wouldn't this be the `object` type?

coolreader18: Oh, would it? I'm not really familiar with how the type object system works, I thought the type of PyInstance would be dependent on the class it's an instance of.

vm/src/pyobject.rs
coolreader18: Y'know, I could have sworn there was some function that made a PyObject with the payload as `()`, but I guess not.

vm/src/obj/objobject.rs
OddCoincidence: This method is basically "what type must a value be (or be a subclass of) for it to be valid to extract this payload from it".

coolreader18:
@OddCoincidence any idea why it's failing with a segfault?

OddCoincidence:
Where are you seeing a segfault?

I'm just seeing this in travis:

```
error[E0599]: no method named `downcast_ref` found for type `std::boxed::Box<(dyn rustpython_vm::pyobject::PyObjectPayload + 'static)>` in the current scope
   --> wasm/lib/src/browser_module.rs:174:40
    |
174 |     if let Some(promise) = obj.payload.downcast_ref::<PyPromise>() {
    |                                        ^^^^^^^^^^^^
```

coolreader18:
It's in the pytest tests, `command died with Signals.SIGSEGV`.

OddCoincidence:
Okay, so I did some debugging, here's what I found.

The new `init_type_hierarchy` method @adrian17 added today makes use of a `FromPyObjectRef` impl for `PyRef<PyClass>`. If that fails, it does this:

```rust
            panic!("Error getting inner type: {:?}", obj.typ)
```

but at that point `obj.typ` is __uninitialized memory__. So it segfaults in the debug impl.

```rust
            typ: mem::uninitialized(), // !
```

(So the lesson is basically that `mem::uninitialized` is just ridiculously hard to ever use correctly.)

So why does that impl fail? Because of this line:

```rust
        let payload: &dyn Any = &self.payload;
```

The concrete type for this trait object is a `dyn PyObjectPayload`, so when that impl tries to downcast it to a `PyClass`, it fails because it's actually a `dyn PyObjectPayload`.

coolreader18:
Okay, I fixed by using a as_any() method on the trait as suggested in [this StackOverflow answer](https://stackoverflow.com/a/48979406/9236675).

coolreader18:
Great job figuring that out, by the way; that probably would have taken me an hour.

coolreader18:
Oh boy, now it's a stack overflow. I didn't think this would be that hard.

OddCoincidence:
> Oh boy, now it's a stack overflow. 

I mentioned this on gitter, but in case you didn't see, I noticed that it only occurs with trace logging enabled, so that may be a good clue to help with troubleshooting this.

coolreader18:
That's interesting, I'll look into that.

skinny121:
I have hit a similar issue before. It happened when trying to print Debug representation of a recursive value. This wasn't really a problem with the old PyObjectPayload as it only outputted the type of the payload rather than the value itself.

coolreader18:
Ah, so it's probably because the typ of object_type and type_type point to each other, so it's trying to print that and going on forever.

coolreader18:
Yes, all the tests that are failing are ones that define some function, and it fails just after tracing with `StoreName { that function }`. It then tries to trace the current scope and for whatever reason a function object is recursive. I'll try to fix this once I get home.

coolreader18:
Alright, found the culprit: `Scope`! For whatever reason trying to format the RcList in it was recursive and that caused the stack overflow. I'm not really sure how to fully fix that; for now I'm just formatting with the string `"Scope"`, but I think the `Frame` Debug impl is informative enough for locals, and that's printed after every bytecode instruction.

coolreader18:
The remaining failures are due to recursive objects, e.g. `locals()` or the `recursive` array in `test/snippets/list.py`. What should we do about this? Should we use the ReprGuard for Debug formatting?

skinny121:
If we want a detailed Debug formatting then we kinda do need something similar to that. We probably don't want to use the same guard as it could get weird if it's printed in a __repr__.

coolreader18:
I think for now it's probably fine to do like we did when we had PyObjectPayload as an enum and just have them be very basic strings, e.g "dict", "list", etc.

OddCoincidence:
Rather than changing the `Debug` impls, maybe we can create a new method on `PyValue` called `type_name` or something and use that for the logging?

coolreader18:
I think we *do* want to provide good `Debug` impls for tracing; just not right now, probably in a separate PR, and incrementally.

vm/src/frame.rs
```diff
@@ -1207,26 +1211,30 @@ impl Frame {
 
 impl fmt::Debug for Frame {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        trace!("formatting stack");
```
OddCoincidence: is this needed?

coolreader18: Oh, whoops, that's left over from debugging the recursion. Thanks for catching that!

PR #671, opened 2019-03-13
Remove inner Box from PyObject

OddCoincidence:
(WIP: will need to incorporate changes from #670 and fixup `Debug` / `Display` impls)

This removes the `Box` inside `PyObject`, so that a `PyRef` / `PyObjectRef` only needs to follow a single pointer (just like cpython).

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -782,8 +782,8 @@ impl<T> IntoPyObject for PyRef<T> {
 }
 
 impl<T: ?Sized> fmt::Display for PyRef<T> {
-    fn fmt(&self, _f: &mut fmt::Formatter) -> fmt::Result {
-        Ok(()) // TODO
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        self.obj.fmt(f)
     }
 }
 
@@ -965,9 +965,9 @@ impl BufferProtocol for PyObjectRef {
     }
 }
 
-impl<T: ?Sized> fmt::Debug for PyObject<T> {
-    fn fmt(&self, _f: &mut fmt::Formatter) -> fmt::Result {
-        Ok(()) // TODO
+impl<T: ?Sized + fmt::Debug> fmt::Debug for PyObject<T> {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "[PyObj {:?}]", &self.payload)
     }
 }
 
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -72,7 +72,7 @@ Basically reference counting, but then done by rust.
 /// this reference counting is accounted for by this type. Use the `.clone()`
 /// method to create a new reference and increment the amount of references
 /// to the python object by 1.
-pub type PyObjectRef = Rc<PyObject>;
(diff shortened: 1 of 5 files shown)


vm/src/pyobject.rs
```diff
@@ -695,10 +695,10 @@ impl Default for PyContext {
 /// This is an actual python object. It consists of a `typ` which is the
 /// python class, and carries some rust payload optionally. This rust
 /// payload can be a rust float or rust int in case of float and int objects.
-pub struct PyObject {
+pub struct PyObject<T: ?Sized> {
```
windelbouwman: What is the effect of this template on the code size? Did you check this? This is of interest to the wasm build, since it may be intended for web.

coolreader18: There aren't really any methods on PyObject, so making it generic won't duplicate much code; it's just a container.

coolreader18: We could throw some `#[inline]`s on PyObject::new and PyObject::into_ref but those are the only ones that are really generic.

OddCoincidence: This increases the binary size by about 2.4% for me, which seems acceptable.

(A side note - moving to the new args style should decrease the binary size because there will only be one monomorphization per distinct function signature, whereas `arg_check!` will generate type checking code for each one.)

OddCoincidence:
@coolreader18 it looks like something in `wasm` isn't getting along with this, could you take a look?

```
error[E0277]: the size for values of type `dyn rustpython_vm::pyobject::PyObjectPayload` cannot be known at compilation time
  --> wasm/lib/src/convert.rs:70:39
   |
70 |             let weak_py_obj = wasm_vm.push_held_rc(py_obj).unwrap();
   |                                       ^^^^^^^^^^^^ doesn't have a size known at compile-time
   |
   = help: within `rustpython_vm::pyobject::PyObject<dyn rustpython_vm::pyobject::PyObjectPayload>`, the trait `std::marker::Sized` is not implemented for `dyn rustpython_vm::pyobject::PyObjectPayload`
   = note: to learn more, visit <https://doc.rust-lang.org/book/ch19-04-advanced-types.html#dynamically-sized-types-and-the-sized-trait>
   = note: required because it appears within the type `rustpython_vm::pyobject::PyObject<dyn rustpython_vm::pyobject::PyObjectPayload>`
```

coolreader18:
Ah, you can just put `: ?Sized` on `trait HeldRcInner {}` at the top of `vm_class.rs`.

OddCoincidence:
Yeah, I already tried that - the problem is that now `PyObjectRef` is a fat pointer (because `dyn PyObjectPayload` is unsized) so it cannot be made into a `dyn HeldRcInner` because that would need a "super-fat pointer" (size _and_ vtable), which doesn't exist.

coolreader18:
Hmm, you could probably just change `held_rcs` to be a Vec of PyObjectRefs then. If we needed to use it for something else other than PyObjects we could figure it out once we get there.

coolreader18:
Are you still working on this?

OddCoincidence:
Yeah, but merging in master caused an infinite loop in `linearise_mro` that I haven't had time to debug. I'll get to it this weekend hopefully.

OddCoincidence:
Redone in #671

PR #672, opened 2019-03-13
Use compile::Mode::Single for the demo terminal

coolreader18:
Eventually we should use [`sys.displayhook`](https://docs.python.org/3/library/sys.html#sys.displayhook), but this is fine for now.

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -687,13 +687,10 @@ impl Frame {
                 let expr = self.pop_value();
                 if !expr.is(&vm.get_none()) {
                     let repr = vm.to_repr(&expr)?;
-                    builtins::builtin_print(
-                        vm,
-                        PyFuncArgs {
-                            args: vec![repr],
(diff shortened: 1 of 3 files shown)


coolreader18:
@palaviv do you want to take a look at this? It's your terminal :grinning:

vm/src/frame.rs
```diff
@@ -687,13 +687,10 @@ impl Frame {
...
                     let repr = vm.to_repr(&expr)?;
-                    builtins::builtin_print(
-                        vm,
-                        PyFuncArgs {
-                            args: vec![repr],
-                            kwargs: vec![],
-                        },
-                    )?;
+                    // TODO: implement sys.displayhook
+                    if let Some(print) = vm.ctx.get_attr(&vm.builtins, "print") {
```
windelbouwman: Pro tip: in case of an if-statement, do not forget about the else clause. What happens when we have no print? No worries for now, just a tip.

PR #673, opened 2019-03-13
Add write method to io.StringIO

windelbouwman:


cthulahoops:
Should we use new style declarations for these methods?

windelbouwman:
Is this the way to go forward? How about error checking and typechecking? Is this already completed?

cthulahoops:
I'm finding it pretty good to work with. Produces less boiler plate. Not everything is supported, so sometimes you'll have to fall back to the old style, and sometimes the compiler will throw ridiculous errors at you - but generally good.

adrian17:
> How about error checking and typechecking? Is this already completed?

Yeah, it should work just fine. For example:

```
fn pop(self, i: OptionalArg<isize>, vm: &mut VirtualMachine) -> PyResult {
```
```
>>>>> [5].pop()
5
>>>>> [5].pop(0)
5
>>>>> [].pop([])
TypeError: Expected type <class 'int'>, not <class 'list'>
>>>>> [].pop(1, 2)
TypeError: Expected at most 2 arguments (3 given)
>>>>> [].pop(1234**1234)
OverflowError: Int value cannot fit into Rust isize
```

The missing pieces would be `KwArgs` (it collects args, but doesn't have an API to read them; also no direct positional-or-keyword support... thankfully builtin functions don't use them often) and `PyIterable` (similar story, but that's just a convenience wrapper).

OddCoincidence:
> also no direct positional-or-keyword support

I have some neat plans for this, something like:

```rust
#[derive(FromArgs)]
struct PrintOptions {
    #[pyarg(default_value = " ")]
    sep: String,
    #[pyarg(default_value = "\n")]
    end: String,
    flush: bool,
    file: Option<"whatever type this is I can't remember">
}

fn print(objects: Arg, options: PrintOptions, vm: &mut VirtualMachine) {
    // print the things
}
```

(haven't implemented this yet tho because proc macros have to be in their own crate, which means splitting up vm, which means can't use `self` on PyRef impls without `arbitrary_self_types`, ...)

coolreader18:
Why would it require splitting up `vm`? We could just make another crate rustpython_macro or something.

OddCoincidence:
> Why would it require splitting up vm? We could just make another crate rustpython_macro or something.

Because the `macros` crate would need access to `PyObject`, `PyFuncArgs`, etc. and various `vm` code would need the macros - so it's a circular dependency.

windelbouwman:
Okay, I gave new style objects a shot, it looks good, but there is a problem implementing `required_type`. I need to get the class `io.StringIO`, which I cannot fetch from the context.

OddCoincidence:
If you merge in master, it should be possible to do this now, the way `Match` in `re` does:

```
        vm.class("re", "Match")
```

(Not that this is a good solution - something similar to [@skinny121's prototype](https://github.com/skinny121/RustPython/blob/39421a75cbabdf5fc3b86a779637b999be409b9e/vm/src/pyobject.rs#L353-L365) but just for stdlib could make this a lot cleaner / efficient / less weird).

windelbouwman:
Yes, I see. I did this. It's still weird how python module function access their own scope via a weird import loophole, but this can be another improvement. Maybe we should pass the scope into PyArgs?

windelbouwman:
I think this change is now ready for second review!

vm/src/stdlib/io.rs
```diff
@@ -322,37 +342,25 @@ pub fn io_open(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
 
     //Construct a FileIO (subclass of RawIOBase)
     //This is subsequently consumed by a Buffered Class.
-    let file_args = PyFuncArgs::new(
-        vec![file.clone(), vm.ctx.new_str(modes[0].to_string())],
-        vec![],
-    );
+    let file_args = vec![file.clone(), vm.ctx.new_str(modes[0].to_string())];
```
OddCoincidence: Cool, much less verbose :slightly_smiling_face:

vm/src/stdlib/io.rs
```diff
@@ -33,16 +35,39 @@ fn compute_c_flag(mode: &str) -> u16 {
     }
 }
 
-fn string_io_init(vm: &mut VirtualMachine, _args: PyFuncArgs) -> PyResult {
-    // arg_check!(vm, args, required = [(s, Some(vm.ctx.str_type()))]);
-    // TODO
-    Ok(vm.get_none())
+#[derive(Debug)]
+struct PyStringIO {
```
OddCoincidence: For types in stdlib, I'd prefer we leave off the `Py` prefixes. To me, the purpose of those in the builtin types is to distinguish them from rust types, e.g. `PyString` vs `String`

vm/src/stdlib/io.rs
```diff
@@ -33,16 +35,39 @@ fn compute_c_flag(mode: &str) -> u16 {
...
+
+impl PyValue for PyStringIO {
+    fn class(vm: &mut VirtualMachine) -> PyObjectRef {
+        vm.class("io", "StringIO")
+    }
+}
+
+impl PyStringIORef {
+    fn write(self, data: objstr::PyStringRef, _vm: &mut VirtualMachine) {
+        let data = data.value.clone();
```
OddCoincidence: I don't think this clone is necessary - `&data.value` should work

vm/src/stdlib/io.rs
windelbouwman: At first sight, I tend to disagree, my reason for prefixing all python types with `Py` would be to make clear that those are python types. Unlike for example VirtualMachine or Context. So for me its more than distuingish them from builtin rust types. That being said, I'm fine with both :)

PR #674, opened 2019-03-13
Fix subclasses of string. (Fixes #130)

### LINKED ISSUE #130: Sub-classes of `str` appear to just operate as aliases of `str`
```py
>>>>> class String(str):
.....     def foo(self):
.....         return "foo"
..... 
>>>>> String()
'' 
>>>>> String().foo()
  File <unknown>, line 0, in <module>
AttributeError: RefCell { value: [PyObj class "str"] } object has no attribute foo
>>>>> type(String())
<class 'str'> 
```

cthulahoops:


vm/src/obj/objstr.rs
```diff
@@ -106,8 +106,11 @@ impl PyStringRef {
         }
     }
 
-    fn str(self, _vm: &mut VirtualMachine) -> PyStringRef {
-        self
+    fn str(object: OptionalArg<PyStringRef>, vm: &mut VirtualMachine) -> PyResult<PyStringRef> {
```
adrian17: How does this work? It's a `str.__str__` implementation, but in CPython:
```
>>> str.__str__()
TypeError: descriptor '__str__' of 'str' object needs an argument
```

vm/src/obj/objstr.rs
```diff
@@ -788,16 +791,21 @@ fn perform_format(
...
 // class str(object='')
 // class str(object=b'', encoding='utf-8', errors='strict')
-fn str_new(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
-    if args.args.len() == 1 {
-        return Ok(vm.new_str("".to_string()));
-    }
-
-    if args.args.len() > 2 {
-        panic!("str expects exactly one parameter");
+fn str_new(
```
OddCoincidence: Now that methods exactly match their python names, should we do the same for free-standing functions? i.e. should this just be `str`?

tests/snippets/subclass_str.py
```diff
@@ -0,0 +1,22 @@
...
+y = Stringy(1)
+assert type(y) is Stringy, "Type of Stringy should be stringy"
+assert type(str(y)) is str, "Str of a str-subtype should be a str."
+
+assert y + " other" == "1 other"
+assert y.x == "substr"
+
+## Base strings currently get an attribute dict, but shouldn't.
+# with assertRaises(AttributeError):
+#     "hello".x = 5
```
windelbouwman: I think it is more consistent that base strings have attribute dicts, I believe cpython is wrong here. We can deviate. You can use the platform module to check for rustpython.

vm/src/obj/objstr.rs
cthulahoops: Uh... you're right, I think. I managed to confuse myself somehow. I don't need this at all.

vm/src/obj/objstr.rs
cthulahoops: Maybe. I quite like the convention, it makes it a bit easier to know what's what. I think I'd call that ```new``` rather than ```str``` though.

Of course, that function could be a method.

tests/snippets/subclass_str.py
cthulahoops: It feels like a pretty serious performance question, if we have to allocate dicts for intermediate ints and bools in computations.

cthulahoops: It's important enough that's ```__slots__``` exists so that user classes can take advantage of this.

vm/src/obj/objstr.rs
cthulahoops: But it can't be a method called ```new``` because that clashes with the implementation of new in PyRef.

tests/snippets/subclass_str.py
adrian17: > It feels like a pretty serious performance question, if we have to allocate dicts for intermediate ints and bools in computations.

Not just that; this also makes other common optimizations on builtin types (string interning, small int cache) much weirder and harder to reason about.

```
>>> a = "hello"
>>> b = "hello"
>>> a is b
True
>>> a = 123
>>> b = 123
>>> a is b
True
```

So if in one section of code I do `"hello".x = 5`, this may affect any other code (including third-party library) using the same "hello" literal - as it may or may not be interned. Same with other builtins. That's implicit global state.

vm/src/obj/objstr.rs
OddCoincidence: I really like the idea of it being a `new` method though (feels more idiomatic). We can probably get rid of `PyRef::new` and have a generic `into_ref` on `PyValue`s.

cthulahoops:
The remaining discussions are beyond the scope of this PR, so I think this should be ok to merge?

PR #675, opened 2019-03-13
Remove type parameter from PropertyBuilder

skinny121:
This removes the requirement that properties need to be symmetric.

As discussed in #665, it's useful for properties to be asymmetric, to provide more appropriate error messages and the ability for 'get' to return a value but 'set' to take a reference.


OddCoincidence:
The symmetric thing was a cool idea though :smile:

PR #678, opened 2019-03-14
Py value into ref

cthulahoops:
As discussed in #674.

My plan is:
* eliminate PyRef::new*
* move all/most calls to PyObject::new(...) to this interface.
* use this narrow interface to solve the which objects get attributes problem.

Pushing this push first to make sure we agree on the interface.


cthulahoops:
Remember you need a context for into_ref.

PR #679, opened 2019-03-15
Make it possible to implement PyValue in stdlib

OddCoincidence:
- Renamed `PyValue::required_type()` to `PyValue::class()` which is clearer
- Make a full `&mut VirtualMachine` available rather than just a `&PyContext`, which is needed for implementing types in stdlib

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -164,8 +164,8 @@ pub fn create_type(name: &str, type_type: &PyObjectRef, base: &PyObjectRef) -> P
 pub struct PyNotImplemented;
 
 impl PyValue for PyNotImplemented {
-    fn class(ctx: &PyContext) -> PyObjectRef {
-        ctx.not_implemented().typ()
+    fn class(vm: &mut VirtualMachine) -> PyObjectRef {
+        vm.ctx.not_implemented().typ()
     }
 }
 
@@ -173,8 +173,8 @@ impl PyValue for PyNotImplemented {
 pub struct PyEllipsis;
 
 impl PyValue for PyEllipsis {
-    fn class(ctx: &PyContext) -> PyObjectRef {
-        ctx.ellipsis_type.clone()
+    fn class(vm: &mut VirtualMachine) -> PyObjectRef {
+        vm.ctx.ellipsis_type.clone()
     }
 }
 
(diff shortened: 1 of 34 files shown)


vm/src/stdlib/re.rs
```diff
@@ -111,9 +111,8 @@ struct PyMatch {
 }
 
 impl PyValue for PyMatch {
-    fn required_type(_ctx: &PyContext) -> PyObjectRef {
-        // TODO
-        unimplemented!()
+    fn class(vm: &mut VirtualMachine) -> PyObjectRef {
+        vm.import("re").unwrap().get_attr("Match").unwrap()
```
coolreader18: Maybe add a VirtualMachine `import_expect()` method that panics with a message like "expected module {:?} to be available"? It would cut out a lot of similar `.expect()` messages, and it's good to have a message because in theory it's not absolutely guaranteed that a given module would be available you're calling a type check or something.

OddCoincidence: Sure, using the import here still feels weird to me though, hopefully we can figure out something better in the future.

vm/src/vm.rs
```diff
@@ -103,6 +103,13 @@ impl VirtualMachine {
         &frame.scope
     }
 
+    pub fn class(&mut self, module: &str, class: &str) -> PyObjectRef {
```
windelbouwman: Suggestion to rename this function to `import_from`.

vm/src/stdlib/re.rs
windelbouwman: This is defenitely very weird. I also hope we can figure out a better way.

vm/src/pyobject.rs
```diff
@@ -1574,25 +1575,25 @@ impl PyObject {
 }
 
 pub trait PyValue: fmt::Debug + Sized + 'static {
-    fn required_type(ctx: &PyContext) -> PyObjectRef;
+    fn class(vm: &mut VirtualMachine) -> PyObjectRef;
 
-    fn into_ref(self, ctx: &PyContext) -> PyRef<Self> {
+    fn into_ref(self, vm: &mut VirtualMachine) -> PyRef<Self> {
```
cthulahoops: This change is unfortunate. I was hoping to able to use this in PyContext without a vm.

windelbouwman: At some point pyContext and vm should probably be merged into one. It does not make sense to have both I believe.

PR #683, opened 2019-03-15
Objdict refactor

cthulahoops:
Refactor into new style declarations.

Now contains changes to vm.to_str and vm.to_repr to make them return PyStringRefs too.

### DIFF
--- a/vm/src/obj/objdict.rs
+++ b/vm/src/obj/objdict.rs
@@ -185,9 +185,7 @@ impl PyDictRef {
             for (key, value) in elements {
                 let key_repr = vm.to_repr(&key)?;
                 let value_repr = vm.to_repr(&value)?;
-                let key_str = objstr::get_value(&key_repr);
-                let value_str = objstr::get_value(&value_repr);
-                str_parts.push(format!("{}: {}", key_str, value_str));
+                str_parts.push(format!("{}: {}", key_repr.value, value_repr.value));
             }
 
             format!("{{{}}}", str_parts.join(", "))
--- a/vm/src/obj/objdict.rs
+++ b/vm/src/obj/objdict.rs
@@ -371,34 +371,19 @@ fn dict_get(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
 }
 
 pub fn init(context: &PyContext) {
-    let dict_type = &context.dict_type;
-    context.set_attr(&dict_type, "__len__", context.new_rustfunc(dict_len));
-    context.set_attr(
-        &dict_type,
-        "__contains__",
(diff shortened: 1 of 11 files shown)


cthulahoops:
It's easier than I was thinking. I'll do it properly.

vm/src/obj/objdict.rs
```diff
@@ -174,231 +173,146 @@ fn dict_new(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
...
-            str_parts.push(format!("{}: {}", key_str, value_str));
-        }
+    fn repr(self, vm: &mut VirtualMachine) -> PyResult {
+        let s = if let Some(_guard) = ReprGuard::enter(self.as_object()) {
+            let elements = get_key_value_pairs(self.as_object());
+            let mut str_parts = vec![];
+            for (key, value) in elements {
+                let key_repr = vm.to_repr(&key)?;
+                let value_repr = vm.to_repr(&value)?;
+                let key_str = objstr::get_value(&key_repr);
```
adrian17: (we probably should change `to_repr()` to return `PyResult<PyStringRef>` - `get_value` calls are a bit ugly and they panic when `__repr__` returns a non-string.)

vm/src/obj/objdict.rs
```diff
@@ -174,231 +173,146 @@ fn dict_new(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+    ) -> PyObjectRef {
+        // What we are looking for:
+        let ref key = key.value;
+
+        let elements = self.entries.borrow();
+        if elements.contains_key(key) {
+            elements[key].1.clone()
+        } else {
+            match default {
+                OptionalArg::Present(ref value) => value.clone(),
```
adrian17: you probably don't need to `clone` here:
`OptionalArg::Present(value) => value,`

vm/src/obj/objdict.rs
cthulahoops: Have changed both ```vm.to_str``` and ```vm.to_repr```.

coolreader18:
I'm not at home so I can't do a proper review, but I saw you used `let ref varname = string.value;`, and IIRC clippy would nag you to change it to `let varname = &string.value;`. Otherwise it looks good!

cthulahoops:
> I'm on mobile so I can't do a proper review, but I saw you used `let ref varname = string.value;`, and IIRC clippy would nag you to change it to `let varname = &string.value;`. Otherwise it looks good!

Thanks. I'll have to get into the habit of using clippy, and break the ```let ref``` habit.

Issue #530, opened 2019-02-23
Testing WebAssembly


coolreader18:
There are a few questions if we were to write tests for the WASM library:

- What "framework" to use?:
  - I saw @windelbouwman mention https://wasmer.io in the Gitter chat.
  - wasm-bindgen also has it's own [testing system](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/index.html), which supports testing in headless browsers through WebDrivers.
- How would it integrate with CodeCov? Is it possible to integrate it with CodeCov?

windelbouwman:
It might be good to build the wasm module, and call a few functions using `wasmer-runtime`. I did some tests here: https://github.com/windelbouwman/pzn/blob/master/pzn/src/main.rs#L140

On the other hand, a headless browser like phantomjs might be good to check browser integration.

windelbouwman:
@coolreader18 can this issue be closed now? We have wasm tests. Or is this a discussion? Then we should label it as question.

coolreader18:
Yeah, I guess so.

PR #652, opened 2019-03-10
Add __import__

palaviv:
I will add the builtin module as well so I can test overriding the `__import__`.

PR #684, opened 2019-03-15
Selenium tests for WASM

palaviv:
Not many tests but this should validate we don't break WASM.
I tried using PhantomJS as mentioned in #530 but it did not work...


adrian17:
That seems like a bit of an overkill to me. Wouldn't it possible to load wasm and run tests on Node?

coolreader18:
Yes, with wasm-bindgen-test, as I mentioned in #530, you can run tests in Node.

windelbouwman:
This is very cool! @palaviv nice job!

adrian17:
> Yes, with wasm-bindgen-test, as I mentioned in #530, you can run tests in Node.

So, why would we need selenium tests, then?

windelbouwman:
@adrian17 I agree with you that this test is spinning up a lot of different tools, but I think it is a good crosscheck. Can we do the same with wasm-bindgen-test in a lightweigher way, or do we exclude some code when we do that?

coolreader18:
wasm-bindgen-test uses the native rust testing framework, e.g. `#[test]`, so we'd have to design our tests around that. Another concern is that as far as I know, there's no way to get code coverage for rust compiled to WASM, so that sucks.

palaviv:
@adrian17 I am not familiar with `wasm-bindgen-test`. I wanted to push something to validate the WASM is  good in the CI as I almost broke it on a typo in #652. I agree selenium is very complex and if we can do it by using other tools lets do that.

PR #685, opened 2019-03-15
Refactor symbolic strings

### LINKED ISSUE #463: Refactor Symbolic Strings
There are symbolic strings littered throughout the code base.
e.g. "\_\_repr\_\_", "\<module\>"

These should be hidden behind ~enums~ static constants

ZapAnton:
This PR replaces the occurrences of the symbolic strings like `"__repr__"` with the appropriate static string.

Closes #463

### DIFF
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -325,7 +325,7 @@ fn builtin_hasattr(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
 fn builtin_hash(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
     arg_check!(vm, args, required = [(obj, None)]);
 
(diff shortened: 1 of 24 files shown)


OddCoincidence:
Hmm, I don't really see the benefits of this. Seems like just using the actual names is more clear...

ZapAnton:
Maybe the naming of the variables could be betters, but I think that having only one variable located in one specific place, instead of manipulating the strings all over the code base, should be more optimal for the future refactoring.

OddCoincidence:
I don't think the usual "put in constant to ease future refactoring" argument makes sense here though - these values will never change.

cthulahoops:
For me this seems to make the code harder to read.

ZapAnton:
I will closing this one then. Should #463 be closed too?

PR #686, opened 2019-03-15
Move set type out of PyContext

skinny121:
The main motivation to move type initialisation/declaration out of PyContext is to allow a closer alignment with extension and library types, which can be implemented the exact same way. With this change, the only thing that makes set a special type(with the implementation) is that it is part of builtins module.

### DIFF
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -19,7 +19,8 @@ use crate::obj::objtype;
 
 use crate::frame::Scope;
 use crate::pyobject::{
-    AttributeProtocol, IdProtocol, PyContext, PyFuncArgs, PyObjectRef, PyImmutableClass, PyResult, TypeProtocol,
+    AttributeProtocol, IdProtocol, PyContext, PyFuncArgs, PyImmutableClass, PyObjectRef, PyResult,
+    TypeProtocol,
 };
 
 #[cfg(not(target_arch = "wasm32"))]
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -19,7 +19,7 @@ use crate::obj::objtype;
 
 use crate::frame::Scope;
 use crate::pyobject::{
-    AttributeProtocol, IdProtocol, PyContext, PyFuncArgs, PyObjectRef, PyResult, TypeProtocol,
+    AttributeProtocol, IdProtocol, PyContext, PyFuncArgs, PyObjectRef, PyImmutableClass, PyResult, TypeProtocol,
 };
 
 #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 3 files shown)


vm/src/pyobject.rs
```diff
@@ -1599,6 +1625,33 @@ pub trait PyValue: fmt::Debug + Sized + 'static {
     }
 }
 
+pub trait PyImmutableClass: PyValue {
```
OddCoincidence: What's the reason for this being a separate trait from `PyValue`?

skinny121: My initial thought was that as PyValue::class required the VM, then not all PyValue types are immutable/can be shared between VMs. But depending on what we want the VM/PyContext distinction to be, these methods might move into PyValue.

OddCoincidence: I think they should be part of `PyValue` either way. From the point of view of the VM, they are all "immutable" (i.e. only accessible via shared refs).

skinny121: If I move 
```rust
fn create_type(ctx: &PyContext) -> PyClassRef;
```
Into PyValue then how is that going to work for Regex? As it requires the VM to get the type.

OddCoincidence: It currently uses the vm, but wouldn't everything be moved to use these `get_type` / `set_type` methods you added to `PyContext`? I don't see any reason to have two different ways to register built-in types.

vm/src/builtins.rs
```diff
@@ -793,7 +795,7 @@ pub fn make_module(ctx: &PyContext) -> PyObjectRef {
         "repr" => ctx.new_rustfunc(builtin_repr),
         "reversed" => ctx.new_rustfunc(builtin_reversed),
         "round" => ctx.new_rustfunc(builtin_round),
-        "set" => ctx.set_type(),
+        "set" => objset::PySet::class_from_ctx(ctx),
```
windelbouwman: I would prefer here: `vm.types.set_type()`. I think it is okay to have a collection of builtin types stored in some struct, like the exceptions.

vm/src/obj/objset.rs
```diff
@@ -71,143 +124,85 @@ fn insert_into_set(
...
-    arg_check!(vm, args, required = [(o, Some(vm.ctx.set_type()))]);
-
-    let elements = get_elements(o);
-    let s = if elements.is_empty() {
+fn set_repr(set: PySetRef, vm: &mut VirtualMachine) -> PyResult<String> {
+    let elements = set.elements.borrow();
+    Ok(if elements.is_empty() {
         "set()".to_string()
-    } else if let Some(_guard) = ReprGuard::enter(o) {
+    } else if let Some(_guard) = ReprGuard::enter(set.as_object()) {
```
windelbouwman: Does the reprguard still work when you use as_object?

vm/src/pyobject.rs
```diff
@@ -1599,6 +1625,33 @@ pub trait PyValue: fmt::Debug + Sized + 'static {
...
+
+    fn into_ref_with_context(self, ctx: &PyContext) -> PyRef<Self> {
+        PyRef {
+            obj: PyObject::new(self, Self::class_from_ctx(ctx).into_object()),
+            _payload: PhantomData,
+        }
+    }
+
+    fn class_from_ctx(ctx: &PyContext) -> PyClassRef {
+        if let Some(class) = ctx.get_type::<Self>() {
```
cthulahoops: I'm not sure I like this lazy initialisation of types. If we go down this path every type check will require checking that the type exists and looking it up in a hash table. A struct of builtin types seems simpler and easier.

windelbouwman: Agreed. I'm in favor of keeping the struct with types, like the exception zoo.

skinny121:
Agreed this turned out not to be the best approach, although we need to come up with some mechanism to look up types in modules to make PyValue work. This may or may not them be able to be applied to built-in types.

PR #691, opened 2019-03-16
Convert print() builtin to new args style

OddCoincidence:
Based on #693 to avoid merge conflict.

I decided to convert this because it hits a lot of edge cases and thus helps to validate the model of the new args style.

In addition to converting to the new style, this also fixes a couple bugs:

- `Args` was swallowing the first keyword argument after it finished slurping the remaining positional args.
- `print` was accepting arbitrary / unknown keyword arguments, they are now rejected 

Note:

`KwArgs` is _only_ for functions that accept arbitrary keyword arguments, (i.e. `**kwargs` in python). For functions that accept _specific_ named arguments, a rust struct with an appropriate `FromArgs` implementation must be created. ~~This means some boilerplate right now, but is still cleaner / less error-prone than doing it in the function itself, and will be eliminated in the future with a `#[derive(FromArgs)]` macro.~~ *

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /target
+/*/target
 **/*.rs.bk
 **/*.bytecode
 __pycache__
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -1,0 +1,48 @@
+extern crate proc_macro;
+
+use proc_macro::TokenStream;
+use proc_macro2::TokenStream as TokenStream2;
+use quote::quote;
+use syn::{Data, DeriveInput, Fields};
+
+#[proc_macro_derive(FromArgs)]
+pub fn derive_from_args(input: TokenStream) -> TokenStream {
+    let ast: DeriveInput = syn::parse(input).unwrap();
+
+    let gen = impl_from_args(&ast);
+    gen.to_string().parse().unwrap()
+}
+
+fn impl_from_args(input: &DeriveInput) -> TokenStream2 {
+    // FIXME: This references types using `crate` instead of `rustpython_vm`
+    //        so that it can be used in the latter. How can we support both?
+    let fields = match input.data {
+        Data::Struct(ref data) => {
+            match data.fields {
+                Fields::Named(ref fields) => fields.named.iter().map(|field| {
+                    let name = &field.ident;
+                    quote! {
+                        #name: crate::pyobject::TryFromObject::try_from_object(
+                            vm,
+                            args.take_keyword(stringify!(#name)).unwrap_or_else(|| vm.ctx.none())
+                        )?,
+                    }
+                }),
+                Fields::Unnamed(_) | Fields::Unit => unimplemented!(), // TODO: better error message
+            }
+        }
+        Data::Enum(_) | Data::Union(_) => unimplemented!(), // TODO: better error message
+    };
+
+    let name = &input.ident;
+    quote! {
+        impl crate::function::FromArgs for #name {
+            fn from_args(
+                vm: &mut crate::vm::VirtualMachine,
+                args: &mut crate::function::PyFuncArgs
+            ) -> Result<Self, crate::function::ArgumentError> {
+                Ok(#name { #(#fields)* })
+            }
+        }
+    }
+}
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -4,7 +4,6 @@
 
 use std::char;
 use std::io::{self, Write};
-use std::iter;
 use std::path::PathBuf;
 
 use num_traits::{Signed, ToPrimitive};
@@ -19,7 +18,7 @@ use crate::obj::objstr::{self, PyStringRef};
 use crate::obj::objtype;
 
 use crate::frame::Scope;
-use crate::function::{Args, ArgumentError, FromArgs, PyArg, PyFuncArgs};
+use crate::function::{Args, ArgumentError, FromArgs, PyFuncArgs};
 use crate::pyobject::{
     AttributeProtocol, IdProtocol, PyContext, PyObjectRef, PyResult, TryFromObject, TypeProtocol,
 };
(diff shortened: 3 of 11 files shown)


adrian17:
Side note:

> Note:
> KwArgs is only for functions that accept arbitrary keyword arguments

I feel like at this point too much of the "documentation" on how to use the argument structs is split between pull requests comments and "deduce it from code". For example, I had no idea this was the intenion of KwArgs. Could we have it properly documented somewhere at some point?

vm/src/builtins.rs
```diff
@@ -583,71 +584,84 @@ fn builtin_pow(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+impl FromArgs for PrintOptions {
+    fn from_args<I>(
+        vm: &mut VirtualMachine,
+        args: &mut iter::Peekable<I>,
+    ) -> Result<Self, ArgumentError>
+    where
+        I: Iterator<Item = PyArg>,
+    {
+        let mut sep: Option<PyStringRef> = Some(PyString::from(" ").into_ref(vm));
+        let mut end: Option<PyStringRef> = Some(PyString::from("\n").into_ref(vm));
```
coolreader18: Shouldn't these be initialized to `None`?

vm/src/builtins.rs
```diff
@@ -583,71 +584,84 @@ fn builtin_pow(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+        let mut flush = false;
+
+        while let Some(arg) = args.next() {
```
coolreader18: Why not use a `for..in` loop?

OddCoincidence:
> I feel like at this point too much of the "documentation" on how to use the argument structs is split between pull requests comments and "deduce it from code". For example, I had no idea this was the intenion of KwArgs. Could we have it properly documented somewhere at some point?

@adrian17 `KwArgs` does has the following doc comment:

```
/// A map of keyword arguments to their values.
///
/// A built-in function with a `KwArgs` parameter is analagous to a Python
/// function with `**kwargs`. All remaining keyword arguments are extracted
/// (and hence the function will permit an arbitrary number of them).
///
/// `KwArgs` optionally accepts a generic type parameter to allow type checks
```

But I suppose I could expand it if it's not clear...

OddCoincidence:
Okay, I added (a slightly tweaked version of) this PR comment to the `KwArgs` docs.

vm/src/builtins.rs
OddCoincidence: Thanks, nice catch (an earlier iteration was using `peek`)

vm/src/builtins.rs
OddCoincidence: No - [the defaults for these params are `" "` and `"\n"`](https://docs.python.org/3/library/functions.html#print). `None` in this context corresponds to a _python_ `None`, which means not to print a separator / end.

OddCoincidence:
Okay, I went ahead and created the initial `FromArgs` derive, was easier than I thought.

.gitignore
```diff
@@ -1,4 +1,5 @@
 /target
+/*/target
```
coolreader18: Could probably just be `**/target`

derive/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+    //        so that it can be used in the latter. How can we support both?
+    let fields = match input.data {
+        Data::Struct(ref data) => {
+            match data.fields {
+                Fields::Named(ref fields) => fields.named.iter().map(|field| {
+                    let name = &field.ident;
+                    quote! {
+                        #name: crate::pyobject::TryFromObject::try_from_object(
+                            vm,
+                            args.take_keyword(stringify!(#name)).unwrap_or_else(|| vm.ctx.none())
```
coolreader18: It probably is more efficient to iterate once through instead of each individually, and if we're writing a macro we should think about things like that.

.gitignore
OddCoincidence: I don't really like doing `**` in `.gitignore`, it's too broad.

derive/src/lib.rs
OddCoincidence: Once `kwargs` in `PyFuncArgs` is a map, it'll actually be doing _more_ work by iterating (O(n) vs O(1) times a constant factor). Also another issue with that approach is that you have to store a bunch of locals as `Option`s which further decreases efficiency and also just makes things more complex.

derive/src/lib.rs
```diff
@@ -0,0 +1,48 @@
...
+                }),
+                Fields::Unnamed(_) | Fields::Unit => unimplemented!(), // TODO: better error message
+            }
+        }
+        Data::Enum(_) | Data::Union(_) => unimplemented!(), // TODO: better error message
+    };
+
+    let name = &input.ident;
+    quote! {
+        impl crate::function::FromArgs for #name {
```
windelbouwman: As this crate clearly depends upon rustpython_vm, I would not make a new crate. This makes the code less easy to understand. Why would you create a new crate?

coolreader18: That's required to make a procedural macro. If you look in it's Cargo.toml, you'll see `proc-macro = true`, which means that you can export proc macros from it, while you can't do that from a normal crate.

OddCoincidence: Also note it _doesn't_ depend on `rustpython_vm` (as I originally thought it would) - This is inside a `quote!` macro so it's just generating it on the syntax tree level.

windelbouwman: Can't we set `proc-macro = true` in the rustpython_vm crate?

OddCoincidence: No, rust is pretty strict about the separation of proc macros:

```
error: `proc-macro` crate types cannot export any items other than functions tagged with `#[proc_macro_derive]` currently
  --> vm/src/lib.rs:58:1
   |
58 | pub use self::vm::VirtualMachine;
```

windelbouwman: Is there another way in which we can prevent the new crate? Why do we require the procedural macro?

OddCoincidence: No, I think this is necessary to have fully accurate argument validation for all cases. Right now, neither the `arg_check!` macro or the new style property handle some cases, such as arguments that can be provided either as positional or keyword.

With the macro, we can just do the following:

```rust
#[derive(FromArgs)]
struct CustomArgs {
    #[pyarg(kind = "positional_or_keyword")]
    arg1: Option<PyObjectRef>,
    arg2: ...
}
```

The macro will create a `FromArgs` implementation for this struct that handles all of this complexity, efficiently, and prevents it from being duplicated for every function that accepts arguments this way.

(It's also the idiomatic rust way to handle cases like this.)

coolreader18: The nice thing is that if we want to make another procedural macro, we don't have to make a new crate - all of them can go into `rustpython_derive`.

OddCoincidence: @windelbouwman What are your concerns about adding a new crate?

windelbouwman: My concerns are that the code becomes less readable. The other concern is the splitting of software into seperate packages in general. Have you heard about leftpad in javascript land? Splitting software into multiple pieces is not always a good idea. That being said, I think for rust python, the software could be split into parser, compiler, vm and bytecode for example. Logical chunks, but I am also in favor of a monolithic rustpython crate which contains all.

That being said, I think we can go forward with this change. I'd rather not, but if there is no other way, we should do it!

OddCoincidence: > My concerns are that the code becomes less readable.

I'll add docs on deriving `FromArgs` to `vm/src/function.rs` that explain how this works, so hopefully that will help mitigate this concern.

> Have you heard about leftpad in javascript land?

Yeah, I remember than incident!

Did you know there's also a lib called `is-even`, with [this implementation](https://github.com/jonschlinkert/is-even/blob/master/index.js#L13)? :rofl:  

> That being said, I think for rust python, the software could be split into parser, compiler, vm and bytecode for example. 

Sounds reasonable!

> but I am also in favor of a monolithic rustpython crate which contains all.

Many projects in rust do both, i.e. have a "monolithic" crate, that actually just re-exports the constituent crates to create a cohesive experience for users. The [new `gloo` project](https://rustwasm.github.io/2019/03/12/lets-build-gloo-together.html) takes this approach for example (see the point about creating an umbrella crate).

> That being said, I think we can go forward with this change. I'd rather not create the extra crate, but if there is no other way, we should do it!

Thanks! I'll do what I can to mitigate your concerns. FWIW, I'm not a fan of rust forcing proc-macros into their own crates either - hopefully they'll find a way to make that unnecessary in the future.

PR #694, opened 2019-03-17
Various fixes for getattr/setattr/hasattr

skinny121:
Fixes the following:
- setattr() now calls \_\_setattr\_\_ and implemented object.\_\_setattr\_\_
- Fix \_\_getattr\_\_ getting called with the class instead of the instance
- Fix hasattr catching non-AttributeError exceptions
- Added default paramter to getattr
- object() had a dict
- property not returning itself for class bindings

### DIFF
--- a/vm/src/obj/objnone.rs
+++ b/vm/src/obj/objnone.rs
@@ -1,6 +1,9 @@
 use crate::function::PyFuncArgs;
+use crate::obj::objproperty::PyPropertyRef;
+use crate::obj::objstr::PyStringRef;
 use crate::pyobject::{
-    IntoPyObject, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
+    AttributeProtocol, IntoPyObject, PyContext, PyObjectRef, PyRef, PyResult, PyValue,
+    TryFromObject, TypeProtocol,
 };
 use crate::vm::VirtualMachine;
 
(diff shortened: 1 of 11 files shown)


skinny121:
No, that's fine, I will rebase to remove the formatting commit.

vm/src/obj/objnone.rs
```diff
@@ -39,6 +42,57 @@ impl PyNoneRef {
     fn bool(self, _vm: &mut VirtualMachine) -> PyResult<bool> {
         Ok(false)
     }
+
+    fn get_attribute(self, name: PyStringRef, vm: &mut VirtualMachine) -> PyResult {
```
windelbouwman: This piece I do not get. Do we want to remove it in future again?

windelbouwman:
Btw, I like the fact that on the virtualmachine we have now most methods available, such as setattr, getattr, delattr and friends. They simply invoke the corresponding magic method.

vm/src/obj/objnone.rs
skinny121: What do you mean again? We may be able to remove this in the future once we have slots implemented for the core operations. As we currently don't, we have no way of representing Option<PyObjectRef> as a PyValue, without conflating Some(PyNone) with None. So I had to find a workaround for cases where it mattered, which are properties on None.

windelbouwman:
@skinny121 I want to merge this, but there is still a conflicting file. Could you have a look at this?

skinny121:
@windelbouwman I have resolved the conflict and removed the formatting commit by performing formatting on every commit.

PR #696, opened 2019-03-18
super in class methods

palaviv:
Allow super to work in class methods.
I am not sure how fragile using "self" is but feel free to suggest something else.

### DIFF
--- a/vm/src/obj/objsuper.rs
+++ b/vm/src/obj/objsuper.rs
@@ -8,10 +8,23 @@ https://github.com/python/cpython/blob/50b48572d9a90c5bb36e2bef6179548ea927a35a/
 
 use crate::function::PyFuncArgs;
 use crate::obj::objtype::PyClass;
-use crate::pyobject::{DictProtocol, PyContext, PyResult, TypeProtocol};
+use crate::pyobject::{
+    DictProtocol, PyContext, PyObject, PyObjectRef, PyResult, PyValue, TypeProtocol,
+};
 use crate::vm::VirtualMachine;
 
 use super::objtype;
+
+#[derive(Debug)]
+pub struct PySuper {
+    obj: PyObjectRef,
+}
+
+impl PyValue for PySuper {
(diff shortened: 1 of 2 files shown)


coolreader18:
The thing that seems fragile about using the local `self` variable is that it doesn't have to be named `self` - this is perfectly valid code:

```python
class A:
    def __init__(me, attr):
        me.attr = attr
    def print_attr(me):
        print(me.attr)
A(9).print_attr()
```

I'm not quite sure what a better implementation would be, it may require modifying the `Frame`  struct to hold a reference to the `self`  object.

adrian17:
For comparison, CPython's approach:

https://docs.python.org/3/reference/datamodel.html
https://www.python.org/dev/peps/pep-3135/
> Every function will have a cell named `__class__` that contains the class object that the function is defined in.
>  and `<firstarg>` is the first parameter of the method (normally self for instance methods, and cls for class methods

So for the class:

```
class C:
	def f(self):
		print(__class__) # <class '__main__.C'>
                super() # uses __class__
```

And for the variable... I guess that if we switch to array locals storage, that would work just like in CPython. Before that, we could probably store the object reference in some magic "local variable" with name that isn't a correct Python identifier (so it can't be overwritten)?

vm/src/obj/objsuper.rs
```diff
@@ -80,8 +115,6 @@ fn super_init(vm: &mut VirtualMachine, args: PyFuncArgs) -> PyResult {
         ));
     }
 
-    // TODO: how to store those types?
```
windelbouwman: This TODO can now be solved by using new-style classes as can be seen in objint.rs for example. Implement a struct named `struct PySuper { PyObjectRef obj }` and use this struct as a payload. Note that the init method might need to be changed to a super_new method.

windelbouwman:
@palaviv it great that you start on this, actually it is super!

palaviv:
I am not sure about changing the `Frame`. I will look into the idea of a "magic" local variable.

PR #697, opened 2019-03-19
More pyobject

cthulahoops:
Continuing my project to refactor away direct use of PyObject.


windelbouwman:
Apart from the valid comments from Adrian, this looks good to me.

PR #698, opened 2019-03-19
Use first argument in super

palaviv:
@windelbouwman already merged the previous PR but lets continue the discussion here. What do you guys think about this solution @coolreader18 @adrian17?

### DIFF
--- a/tests/snippets/class.py
+++ b/tests/snippets/class.py
@@ -85,6 +85,20 @@ class Me2(Me):
     def test(me):
         return super().test()
 
+class A():
+    def f(self):
+        pass
+
+class B(A):
+    def f(self):
+        super().f()
+
+class C(B):
+    def f(self):
+        super().f()
+
+C().f()
+
 me = Me2()
 assert me.test() == 100
 
--- a/tests/snippets/class.py
+++ b/tests/snippets/class.py
@@ -75,6 +75,19 @@ c = C()
 assert c.test() == 100
 assert c.test1() == 200
 
+class Me():
+
+    def test(me):
+        return 100
+
+class Me2(Me):
+
+    def test(me):
+        return super().test()
+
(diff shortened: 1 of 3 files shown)


coolreader18:
How does this work with `staticmethod` in comparison to CPython?

adrian17:
In CPython, `super()` doesn't work in a staticmethod anyway, does it?

cthulahoops:
I don't think this works. If you get class from the first argument then uses of ```super``` further up the inheritance chain don't get the right thing and you get an infinite loop.

```
class A(object):
    def f(self):
        pass

class B(A):
    def f(self):
        super().f()

class C(B):
    def f(self):
        super().f()

C().f()
```

```
Running `target/debug/rustpython test.py`

thread 'main' has overflowed its stack
fatal runtime error: stack overflow
```

cthulahoops:
The cell var isn't too hard to add. It's easier to do than describe, so #703.

windelbouwman:
Merged #703, hopefully not too soon, so you can use it here as well.

palaviv:
Thanks @cthulahoops that helped :). I was not sure how to fetch the `__class__` so I added a `get` to `Scope`. I am not sure if that is the best solution...

cthulahoops:
Rather than introducing Scope.get, can't you just use ```load_name``` from the ScopeProtocol?

cthulahoops:
If you're worried about the search going into builtins, I'd add a ```load_cell``` to the scope protocol to search non-local scope only (with the same shape as load_name).

palaviv:
I Added `load_cell`.

cthulahoops:
This looks great! It's excellent to finally have super.

Got some sort of vm mutability issue now (from the frame change?), but other than that I think we're good to go.

PR #699, opened 2019-03-20
Fix a bunch of clippy lints (again)

coolreader18:
The biggest change in this is that I changed the representation of varargs in the bytecode and AST from `Option<Option<_>>` to its own enum in commit ab53883. I sorta jury-rigged it into the lalrpop file because I'm not too familiar with how it functions, so feel free to give critique for that.

### DIFF
--- a/parser/src/ast.rs
+++ b/parser/src/ast.rs
@@ -274,8 +274,8 @@ impl Expression {
 pub struct Parameters {
     pub args: Vec<Parameter>,
     pub kwonlyargs: Vec<Parameter>,
-    pub vararg: Option<Option<Parameter>>, // Optionally we handle optionally named '*args' or '*'
-    pub kwarg: Option<Option<Parameter>>,
+    pub vararg: Varargs, // Optionally we handle optionally named '*args' or '*'
+    pub kwarg: Varargs,
     pub defaults: Vec<Expression>,
     pub kw_defaults: Vec<Option<Expression>>,
 }
@@ -390,4 +390,29 @@ pub enum StringGroup {
     Joined {
         values: Vec<StringGroup>,
     },
+}
+
+#[derive(Debug, PartialEq)]
+pub enum Varargs {
+    None,
+    NoCapture,
+    Capture(Parameter),
+}
+
+impl Default for Varargs {
+    fn default() -> Varargs {
+        Varargs::None
+    }
+}
+
+impl From<Option<Option<Parameter>>> for Varargs {
+    fn from(opt: Option<Option<Parameter>>) -> Varargs {
+        match opt {
+            Some(inner_opt) => match inner_opt {
+                Some(param) => Varargs::Capture(param),
+                None => Varargs::NoCapture,
+            },
+            None => Varargs::None,
+        }
+    }
 }
--- a/parser/src/ast.rs
+++ b/parser/src/ast.rs
@@ -395,8 +395,8 @@ pub enum StringGroup {
 #[derive(Debug, PartialEq)]
 pub enum Varargs {
     None,
-    NoCapture,
-    Capture(Parameter),
+    Unnamed,
+    Named(Parameter),
 }
 
 impl Default for Varargs {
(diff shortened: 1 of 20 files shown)


vm/src/bytecode.rs
```diff
@@ -401,3 +408,23 @@ impl fmt::Debug for CodeObject {
...
+        match varargs {
+            ast::Varargs::None => Varargs::None,
+            ast::Varargs::NoCapture => Varargs::NoCapture,
+            ast::Varargs::Capture(param) => Varargs::Capture(param.arg),
+        }
+    }
+}
+
+impl Varargs {
+    pub fn from_ast_vararg_ref(varargs: &ast::Varargs) -> Varargs {
```
OddCoincidence: Why not `impl<'a> From<&'a ast::Varargs> for Varargs`?

vm/src/obj/objlist.rs
```diff
@@ -203,7 +203,7 @@ impl PyListRef {
                 return Ok(index);
             }
         }
-        let ref needle_str = vm.to_str(&needle)?.value;
+        let needle_str = vm.to_pystr(&needle)?;
```
OddCoincidence: Was removing the `.value` intentional here? Will the `Display` impl for `PyString{,Ref}` do the right thing here?

vm/src/obj/objtype.rs
```diff
@@ -331,6 +331,7 @@ fn linearise_mro(mut bases: Vec<Vec<PyClassRef>>) -> Option<Vec<PyClassRef>> {
     Some(result)
 }
 
+#[allow(clippy::implicit_hasher)]
```
OddCoincidence: We could probably disable this one at the crate level, I don't think it's useful at all

vm/src/bytecode.rs
coolreader18: I tried that, but `.into()` didn't work in `compile.rs`, so I sorta just gave up. It does work manually writing `from(&args.varargs)`, though, so I'll change it to be this.

vm/src/pyobject.rs
```diff
@@ -1118,9 +1118,14 @@ impl PyObject {
     }
 
     #[inline]
-    pub fn payload<T: PyValue>(&self) -> Option<&T> {
+    pub fn payload<T: PyObjectPayload>(&self) -> Option<&T> {
         self.payload.as_any().downcast_ref()
     }
+
+    #[inline]
+    pub fn payload_is<T: PyObjectPayload>(&self) -> bool {
```
OddCoincidence: nit: would prefer this just be called `is`

(I'm thinking once there's a `downcast` or `as` on `PyRef`, we'll move away from using `payload`, so calling this `is` would provide a consistent api)

vm/src/obj/objlist.rs
coolreader18: Oh, looking back on it it's probably not the best thing to do; thanks for pointing it out.

parser/src/ast.rs
```diff
@@ -391,3 +391,28 @@ pub enum StringGroup {
...
+    Capture(Parameter),
+}
+
+impl Default for Varargs {
+    fn default() -> Varargs {
+        Varargs::None
+    }
+}
+
+impl From<Option<Option<Parameter>>> for Varargs {
```
OddCoincidence: Where's the double `Option` coming from? Is there a way to eliminate it altogether?

coolreader18: I didn't really want to mess around in the `python.lalrpop` grammar, as I don't really have any experience with it, so I opted for this instead.

OddCoincidence: Okay, that's fine, was just wondering

vm/src/pyobject.rs
OddCoincidence: Oh, bummer, either is fine for now then

parser/src/ast.rs
windelbouwman: The double option is a sign of a triple option enum. Actually the options are: namedparameter, namedstar, and unnamed star.

parser/src/ast.rs
```diff
@@ -391,3 +391,28 @@ pub enum StringGroup {
         values: Vec<StringGroup>,
     },
 }
+
+#[derive(Debug, PartialEq)]
+pub enum Varargs {
+    None,
+    NoCapture,
```
windelbouwman: Suggestion to rename this into UnNamed and Named.

coolreader18: Sounds good.

PR #700, opened 2019-03-20
Change the WASM example to use rustpython from npm

coolreader18:


wasm/example/package.json
```diff
@@ -1,18 +1,17 @@
 {
     "name": "rustpython-wasm-example",
     "version": "1.0.0",
-    "//1": "`dependencies.rustpython_wasm` would be the version of the npm",
-    "//2": "library in a real app",
     "dependencies": {
-        "rustpython_wasm": "file:../lib/pkg/"
+        "rustpython_wasm": "0.1.0-pre-alpha.1"
```
windelbouwman: Why would you do this? Then we cannot test the current version, but always go through a deployment phase via npm?

coolreader18: This is for the example, not the demo. If someone just wanted to copy/paste the example and use it for their own project, currently, they couldn't do so, as it has a `file:` dependency; they'd have to change it to the version from npm.

PR #701, opened 2019-03-20
Fix the WASM build

coolreader18:
For some reason it was panicking when trying to import `json.{dumps,loads}`, so I just switched to using plain rust functions straight from `json.rs`.

### DIFF
--- a/vm/src/stdlib/json.rs
+++ b/vm/src/stdlib/json.rs
@@ -188,22 +188,36 @@ impl<'de> Visitor<'de> for PyObjectDeserializer<'de> {
     }
 }
 
-pub fn ser_pyobject(vm: &mut VirtualMachine, obj: &PyObjectRef) -> serde_json::Result<String> {
+pub fn ser_pyobject(vm: &mut VirtualMachine, obj: &PyObjectRef) -> PyResult<String> {
     let serializer = PyObjectSerializer { pyobject: obj, vm };
-    serde_json::to_string(&serializer)
+    serde_json::to_string(&serializer).map_err(|err| vm.new_type_error(err.to_string()))
 }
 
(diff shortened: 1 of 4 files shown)


vm/src/stdlib/json.rs
```diff
@@ -188,31 +188,31 @@ impl<'de> Visitor<'de> for PyObjectDeserializer<'de> {
     }
 }
 
+pub fn ser_pyobject(vm: &mut VirtualMachine, obj: &PyObjectRef) -> serde_json::Result<String> {
```
windelbouwman: Please do not expose functions like this. I would prefer to figure out what went wrong, and fix that instead.

vm/src/stdlib/mod.rs
```diff
@@ -1,6 +1,6 @@
 mod ast;
 mod dis;
-mod json;
+pub mod json;
```
windelbouwman: Please keep this code private. Having this code unexposed is a good thing, and avoids leaking internals.

vm/src/stdlib/json.rs
coolreader18: What about as a method on VM? `json` could be a `pub(crate) mod` and it wouldn't really be exposed.

OddCoincidence: I don't really agree with @windelbouwman here - this isn't an implementation detail, it's the actual python function so it's inevitable to expose these at some point to allow developing native extensions without unnecessary overhead.

> The code would also be faster because it's not looking up a function and loading it from a module, it's just calling a Rust function.

Yep! :+1:

coolreader18:
I am going to merge this just in order to stop the CI builds from failing, but I feel like the VM methods are a good compromise for concerns regarding leaking internals.

PR #703, opened 2019-03-20
Add __class__ cell to method scopes.

cthulahoops:
This is a basic implementation that I think should be good enough for ```__super__``` to work.

This should be passed through to the metaclass to allow some more advanced trickery, but I think this is good enough. There's a more optimal implementation than a whole dictionary just for this, which we can sort when we come to optimise local variables.

### DIFF
--- a/tests/snippets/class.py
+++ b/tests/snippets/class.py
@@ -22,15 +22,18 @@ class Bar:
         self.x = x
 
     def get_x(self):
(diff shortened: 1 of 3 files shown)


cthulahoops:
A cellvar is the just cpython name/implementation for variables that captured from non-local, non-global scopes/the optimisation that makes them fast. I may be vague on the details too!

PR #704, opened 2019-03-20
WIP: Class get attr

cthulahoops:
Incomplete work in progress just to show my thinking. I want to make AttributeProtocol the high level magic methods and all way to access attributes, and move Class, Object and Module internals into their respective obj...: modules with suitably forbidding names.

The first step is to start moving users away from the interface I want to kill.


cthulahoops:
Can I mark a PR as a draft after opening it? (I miss GitLab so much sometimes.)

PR #705, opened 2019-03-20
Introduce IntoPyStringRef to make vm.get_attribute more usable.

cthulahoops:
A stand-alone piece of my AttributeProtocol work.

I don't know if an equivalent of ```IntoPyObjectRef``` can be pieced together from the existing function machinery, but I couldn't see how.

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -10,7 +10,7 @@ use crate::format::{FormatParseError, FormatPart, FormatString};
 use crate::function::{OptionalArg, PyFuncArgs};
 use crate::pyobject::{
     IdProtocol, IntoPyObject, PyContext, PyIterable, PyObjectRef, PyRef, PyResult, PyValue,
-    TryFromObject, TypeProtocol,
+    TryFromObject, TryIntoRef, TypeProtocol,
 };
 use crate::vm::VirtualMachine;
 
@@ -24,6 +24,7 @@ pub struct PyString {
     // TODO: shouldn't be public
     pub value: String,
 }
+pub type PyStringRef = PyRef<PyString>;
 
 impl<T: ToString> From<T> for PyString {
     fn from(t: T) -> PyString {
@@ -33,7 +34,20 @@ impl<T: ToString> From<T> for PyString {
     }
 }
 
-pub type PyStringRef = PyRef<PyString>;
+impl TryIntoRef<PyString> for String {
+    fn try_into_ref(self, vm: &mut VirtualMachine) -> PyResult<PyRef<PyString>> {
+        Ok(PyString { value: self }.into_ref(vm))
+    }
+}
+
+impl TryIntoRef<PyString> for &str {
+    fn try_into_ref(self, vm: &mut VirtualMachine) -> PyResult<PyRef<PyString>> {
+        Ok(PyString {
+            value: self.to_string(),
+        }
+        .into_ref(vm))
+    }
+}
 
 impl PyStringRef {
     fn add(self, rhs: PyObjectRef, vm: &mut VirtualMachine) -> PyResult<String> {
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -310,7 +310,7 @@ fn builtin_getattr(
     default: OptionalArg<PyObjectRef>,
     vm: &mut VirtualMachine,
 ) -> PyResult {
-    let ret = vm.get_attribute(obj.clone(), attr.into_object());
+    let ret = vm.get_attribute(obj.clone(), attr);
     if let OptionalArg::Present(default) = default {
         ret.or_else(|ex| catch_attr_exception(ex, default, vm))
     } else {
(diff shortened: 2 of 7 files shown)


cthulahoops:
The wasm test is failing. Anyone understand why?

coolreader18:
It's broken for master too. I think it was something related to imports, but it should be fixed with #701. It's not related to anything you did, I'm pretty sure.

vm/src/obj/objstr.rs
```diff
@@ -33,7 +34,36 @@ impl<T: ToString> From<T> for PyString {
     }
 }
 
-pub type PyStringRef = PyRef<PyString>;
+pub trait IntoPyStringRef {
```
OddCoincidence: Since this returns a `PyResult`, should it actually be called `TryIntoPyStringRef`?

vm/src/obj/objstr.rs
```diff
@@ -33,7 +34,36 @@ impl<T: ToString> From<T> for PyString {
...
+    fn into_pystringref(self, vm: &mut VirtualMachine) -> PyResult<PyStringRef>;
+}
+
+impl IntoPyStringRef for PyStringRef {
+    fn into_pystringref(self, _vm: &mut VirtualMachine) -> PyResult<PyStringRef> {
+        Ok(self)
+    }
+}
+
+impl IntoPyStringRef for String {
```
OddCoincidence: These impls do seem a bit redundant since we already have similar ones for `PyString`. Could we do something like:

```rust
impl<T: Into<PyString>> IntoPyStringRef for T
```

Or to generalize it further:

```rust
impl<V: PyValue, T: Into<V>> IntoPyRef<V> for T
```

cthulahoops: I don't think it works.

Assuming you meant:

```
impl<T: Into<PyString>> TryIntoPyStringRef for T
{
    fn into_pystringref(self, vm: &mut VirtualMachine) -> PyResult<PyStringRef> {
        Ok(PyString::from(self).into_ref(vm))
    }
}
```

It complains because:

```
help: the trait `std::fmt::Display` is not implemented for `T`
```

And fixing that:

```
47 | impl TryIntoPyStringRef for PyStringRef {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ conflicting implementation for `pyobject::PyRef<obj::objstr::PyString>`
```

if you allow it to use the version it derives then it ends up formatting the object using the default formatter and you get: ```"'str' object"``` instead of the string. (I was expected to lose object identity in that case, surprised to lose object value.)

vm/src/obj/objstr.rs
cthulahoops: Sure. If it has to be ```try_into_pystringref``` this gets a bit unwieldy, but I guess there won't too many callers.

vm/src/obj/objstr.rs
cthulahoops: This default implementation for ToString is the issue:

```
impl<T> ToString for T where
    T: Display + ?Sized, 
```

vm/src/obj/objstr.rs
OddCoincidence: Yeah, had the same concern for `TryIntoObject` (that's why I dropped the `Py` and `Ref`, since in practice you know it's referring to a `PyObject` from context and working with the value types instead of refs is pretty exceptional.

cthulahoops: Aha! The analogy to ```TryIntoObject``` is what I need to get the right shape of this. It's ```TryIntoRef<T>```. I this will be really useful eventually, and could replace ```TryFromObject```?

cthulahoops: How does this look now?

OddCoincidence: > I this will be really useful eventually, and could replace TryFromObject?

`TryFromObject` is a bit different, e.g. it wouldn't make sense `TryIntoRef` for `isize`, but that is done for `TryFromObject` so we can accept it as a fn parameter. (Sorry if I caused confusion by typoing `TryFromObject` as `TryIntoObject` above).

But if my plan for `PyObjectRef` to eventually be a type alias for `PyRef<dyn PyObjectPayload>` works out, then this would prevent us from every _needing_ a `TryIntoObject` and `TryFromObject` could be replaced with a `TryFromRef`.

> How does this look now?

This looks great!

cthulahoops: I think creating ```impl TryIntoRef<PyInt> for isize``` does make sense.

They're sort of doing the same thing, in one case allowing us to write flexible parameters for functions that can be called from rust, in the other flexible paremeters for functions that can be called from python.

OddCoincidence: Oh sorry that `isize` example was poorly thought out. I basically just meant that they are going in opposite directions:

`TryIntoRef`: Self -> Ref
`TryFromObject`: Ref -> Self

cthulahoops: I see now. They overlap in the one specific case where I've used TryFromObject (Self is PyRef), but there are other implementations.

PR #709, opened 2019-03-21
Avoid attribute protocol

cthulahoops:
First batch of my big change (breaking it up a bit for review friendliness).

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -10,14 +10,8 @@ extern crate rustyline;
 use clap::{App, Arg};
 use rustpython_parser::error::ParseError;
 use rustpython_vm::{
-    compile,
-    error::CompileError,
-    frame::Scope,
-    import,
-    obj::objstr,
-    print_exception,
-    pyobject::{AttributeProtocol, PyResult},
-    util, VirtualMachine,
+    compile, error::CompileError, frame::Scope, import, obj::objstr, print_exception,
+    pyobject::PyResult, util, VirtualMachine,
 };
 use rustyline::{error::ReadlineError, Editor};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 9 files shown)


cthulahoops:
All this use of AttributeProtocol is using the internal interface rather than the proper vm attribute path.

src/main.rs
```diff
@@ -176,8 +170,9 @@ fn run_shell(vm: &mut VirtualMachine) -> PyResult {
         println!("No previous history.");
     }
 
-    let ps1 = &objstr::get_value(&vm.sys_module.get_attr("ps1").unwrap());
-    let ps2 = &objstr::get_value(&vm.sys_module.get_attr("ps2").unwrap());
+    let sys_module = vm.sys_module.clone();
+    let ps1 = &objstr::get_value(&vm.get_attribute(sys_module.clone(), "ps1").unwrap());
+    let ps2 = &objstr::get_value(&vm.get_attribute(sys_module, "ps2").unwrap());
```
coolreader18: CPython uses an empty string if somebody deletes `sys.ps{1,2}`, we could probably do the same.

vm/src/frame.rs
```diff
@@ -810,13 +810,13 @@ impl Frame {
         // If we're importing a symbol, look it up and use it, otherwise construct a module and return
         // that
         let obj = match symbol {
-            Some(symbol) => module.get_attr(symbol).map_or_else(
-                || {
+            Some(symbol) => match vm.get_attribute(module, symbol.as_str()) {
+                Ok(obj) => Ok(obj),
+                Err(_) => {
```
coolreader18: You could just use `.map_err()` here.

vm/src/import.rs
```diff
@@ -63,8 +64,13 @@ pub fn import_module(
...
 
-fn find_source(vm: &VirtualMachine, current_path: PathBuf, name: &str) -> Result<PathBuf, String> {
-    let sys_path = vm.sys_module.get_attr("path").unwrap();
+fn find_source(
+    vm: &mut VirtualMachine,
+    current_path: PathBuf,
+    name: &str,
+) -> Result<PathBuf, String> {
+    let sys_module = &vm.sys_module;
+    let sys_path = vm.get_attribute(sys_module.clone(), "path").unwrap();
```
coolreader18: This applies to all these changes, but why not just do `vm.get_attribute(vm.sys_module.clone(), ...)`?

cthulahoops: I thought I was running into an immutable borrow while borrowed mutably issue (the one you run into with ```vm.do_something(vm.something())```), but that doesn't happen with a simple member access.

cthulahoops: Fixed.

vm/src/frame.rs
cthulahoops: Yes, that's neater.

cthulahoops:
Review comments addressed. Merging.

PR #710, opened 2019-03-21
Class get attr 2

cthulahoops:
Introduce objtype::get_class_attr as the internal interface for access to attributes where we need to bypass the full attribute chain. Use this interface in place of the attribute protocol where this low level access is needed.

I'd like to eventually make these functions private, perhaps by moving call_method, etc to ```PyClassRef``` but I don't think we can do this yet.

### DIFF
--- a/vm/src/function.rs
+++ b/vm/src/function.rs
@@ -35,6 +35,17 @@ impl From<PyObjectRef> for PyFuncArgs {
     }
 }
 
+impl From<(&Args, &KwArgs)> for PyFuncArgs {
+    fn from(arg: (&Args, &KwArgs)) -> Self {
+        let Args(args) = arg.0;
+        let KwArgs(kwargs) = arg.1;
+        PyFuncArgs {
+            args: args.clone(),
+            kwargs: kwargs.iter().map(|(k, v)| (k.clone(), v.clone())).collect(),
+        }
+    }
+}
+
 impl PyFuncArgs {
     pub fn new(mut args: Vec<PyObjectRef>, kwarg_names: Vec<String>) -> PyFuncArgs {
         let mut kwargs = vec![];
--- a/vm/src/obj/objtype.rs
+++ b/vm/src/obj/objtype.rs
@@ -269,7 +269,8 @@ pub fn type_call(class: PyClassRef, args: Args, kwargs: KwArgs, vm: &VirtualMach
     Ok(obj)
 }
 
-fn class_get_item(class: &PyClassRef, attr_name: &str) -> Option<PyObjectRef> {
+// Very private helper function for class_get_attr
+fn class_get_attr_in_dict(class: &PyClassRef, attr_name: &str) -> Option<PyObjectRef> {
     if let Some(ref dict) = class.as_object().dict {
         dict.borrow().get(attr_name).cloned()
     } else {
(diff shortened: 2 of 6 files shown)


vm/src/function.rs
```diff
@@ -35,6 +35,17 @@ impl From<PyObjectRef> for PyFuncArgs {
     }
 }
 
+impl From<(&Args, &KwArgs)> for PyFuncArgs {
+    fn from(arg: (&Args, &KwArgs)) -> Self {
+        let Args(args) = arg.0;
+        let KwArgs(kwargs) = arg.1;
+        PyFuncArgs {
+            args: args.clone(),
+            kwargs: kwargs.iter().map(|(k, v)| (k.clone(), v.clone())).collect(),
```
windelbouwman: I think you could use `.cloned` here. Clippy wpuld say so, I assume?

vm/src/obj/objtype.rs
```diff
@@ -218,61 +254,54 @@ pub fn type_new_class(
...
     }
+}
 
-    if let Some(attr) = cls.get_attr(&name) {
-        let attr_class = attr.typ();
-        if let Some(descriptor) = attr_class.get_attr("__get__") {
-            let none = vm.get_none();
-            return vm.invoke(descriptor, vec![attr, none, cls.clone()]);
-        }
+fn class_has_item(class: &PyClassRef, attr_name: &str) -> bool {
```
windelbouwman: Rename this function to `class_has_attr` since we are not dealing with items here.

vm/src/obj/objtype.rs
```diff
@@ -218,61 +254,54 @@ pub fn type_new_class(
...
-    let mcl = cls.typ();
-
-    if let Some(attr) = mcl.get_attr(&name) {
-        let attr_class = attr.typ();
-        if attr_class.has_attr("__set__") {
-            if let Some(descriptor) = attr_class.get_attr("__get__") {
-                return vm.invoke(descriptor, vec![attr, cls.clone(), mcl]);
-            }
-        }
+fn class_get_item(class: &PyClassRef, attr_name: &str) -> Option<PyObjectRef> {
```
windelbouwman: Change this function name to `class_get_attr`, since we are getting an attribute.

vm/src/function.rs
cthulahoops: I don't think it can work through the tuple:

```
44 |             kwargs: kwargs.iter().cloned().collect(),
   |                                   ^^^^^^ expected tuple, found reference
```

vm/src/obj/objtype.rs
cthulahoops: It's a helper function for the real class_has_attr. (Same confusion exists in pyobject currently.) It's quite simple and only has two calls. Maybe the simplest thing is just to inline it into class_has_attr.

vm/src/obj/objtype.rs
cthulahoops: As above.

cthulahoops:
I think it'll start to get simpler from here. Most of this complexity is inherent in just how complex attribute access is. There's quite a bit of duplication between the three definitions of getattribute which could possibly be pulled out into a separate functions which might help.

vm/src/obj/objtype.rs
cthulahoops: Inlining is messy.

I think ```class_has_attr_in_dict``` and a comment.

PR #711, opened 2019-03-22
Make the REPL handle line continuation better

### LINKED ISSUE #626: rustpython console closes the stm on the first newline
screenshot_20190307_215347

coolreader18:
Edit: this would resolve #626.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -159,6 +159,14 @@ fn get_history_path() -> PathBuf {
     xdg_dirs.place_cache_file("repl_history.txt").unwrap()
 }
 
+fn get_prompt(vm: &mut VirtualMachine, prompt_name: &str) -> String {
+    vm.sys_module
+        .get_attr(prompt_name)
+        .as_ref()
+        .map(objstr::get_value)
+        .unwrap_or_else(String::new)
+}
+
 fn run_shell(vm: &mut VirtualMachine) -> PyResult {
     println!(
         "Welcome to the magnificent Rust Python {} interpreter",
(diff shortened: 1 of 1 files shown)


src/main.rs
```diff
@@ -176,33 +184,45 @@ fn run_shell(vm: &mut VirtualMachine) -> PyResult {
...
-                        prompt = ps1;
                         input = String::new();
                     }
                 }
             }
             Err(ReadlineError::Interrupted) => {
                 // TODO: Raise a real KeyboardInterrupt exception
                 println!("^C");
-                break;
+                continuing = false;
```
windelbouwman: I think we should quit when hitting CTR+C, right? Now we continue.

src/main.rs
```diff
@@ -176,33 +184,45 @@ fn run_shell(vm: &mut VirtualMachine) -> PyResult {
...
-    let ps1 = &objstr::get_value(&vm.sys_module.get_attr("ps1").unwrap());
-    let ps2 = &objstr::get_value(&vm.sys_module.get_attr("ps2").unwrap());
-    let mut prompt = ps1;
+    let mut prompt = get_prompt(vm, "ps1");
+
+    let mut continuing = false;
 
     loop {
-        match repl.readline(prompt) {
+        if !continuing {
```
windelbouwman: I like the mutable boolean continueing, but I would like to change the prompt into an immutable like this:
```rust
let prompt = if continuing {
  get_prompt(vm, "ps2")
} else {
  get_prompt(vm, "ps1")
};
```
This way, prompt is only set once.

src/main.rs
coolreader18: Oh, right. I should've discussed this before making the change, but usually in shells (iPython, bash/zsh/sh/sshshshshshs), ^C just cancels what you're currently typing, it doesn't completely end the shell like ^D does. I found it to be sort of annoying when testing something in the REPL, so I changed it here; if you disagree with it I can change it back though.

src/main.rs
coolreader18: Ah, thanks for noticing that it could be refactored in that way.

src/main.rs
```diff
@@ -159,6 +159,14 @@ fn get_history_path() -> PathBuf {
     xdg_dirs.place_cache_file("repl_history.txt").unwrap()
 }
 
+fn get_prompt(vm: &mut VirtualMachine, prompt_name: &str) -> String {
+    vm.sys_module
+        .get_attr(prompt_name)
```
cthulahoops: Should be ```vm.get_attribute``` rather than ```AttributeProtocol.get_attr``` here.

src/main.rs
windelbouwman: Okay, in that case, nice job!

src/main.rs
```diff
@@ -170,33 +178,44 @@ fn run_shell(vm: &mut VirtualMachine) -> PyResult {
...
 
-    let ps1 = &objstr::get_value(&vm.get_attribute(vm.sys_module.clone(), "ps1").unwrap());
-    let ps2 = &objstr::get_value(&vm.get_attribute(vm.sys_module.clone(), "ps2").unwrap());
-    let mut prompt = ps1;
+    let mut continuing = false;
 
     loop {
-        match repl.readline(prompt) {
+        let prompt = if continuing {
+            get_prompt(vm, "ps2")
```
windelbouwman: Btw, this also fixes changing `sys.ps1` on the fly!

QGB:
How to run IPython  with ”RustPython" ?

https://stackoverflow.com/questions/16536683/how-to-run-ipython-with-pypy

coolreader18:
I think it would be possible, but I don't think RustPython is at a point yet where it would be able to run a project as complex as IPython.

PR #714, opened 2019-03-22
Convert all &mut VirtualMachine to &VirtualMachine

### LINKED ISSUE #713: Change all occurrences of `&mut VirtualMachine` to `&VirtualMachine`

coolreader18:
Resolves #713.

I did some fiddling in WASM cause I realized we didn't need whatever the heck I had done with `AccessibleVMPtr`, but otherwise it's just a regex `&mut VirtualMachine` -> `&VirtualMachine`.


coolreader18:
What I changed with WASM also stops it from completely invalidating a VM when code panics from inside of it, because it's no longer borrowing an entire RefCell that doesn't get dropped due to WASM's lack of `panic=unwind`.

PR #715, opened 2019-03-22
Make set subclass of frozenset

palaviv:
This add the support in most of frozenset methods.

### DIFF
--- a/vm/src/obj/objset.rs
+++ b/vm/src/obj/objset.rs
@@ -33,7 +33,7 @@ impl fmt::Debug for PySet {
 
 impl PyValue for PySet {
     fn class(vm: &VirtualMachine) -> PyObjectRef {
-        vm.ctx.set_type()
+        vm.ctx.frozenset_type()
     }
 }
 
@@ -157,7 +157,7 @@ fn set_new(
     iterable: OptionalArg<PyObjectRef>,
     vm: &VirtualMachine,
 ) -> PyResult<PySetRef> {
-    if !objtype::issubclass(cls.as_object(), &vm.ctx.set_type()) {
+    if !objtype::issubclass(cls.as_object(), &vm.ctx.frozenset_type()) {
         return Err(vm.new_type_error(format!("{} is not a subtype of set", cls)));
     }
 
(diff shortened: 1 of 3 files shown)


adrian17:
This doesn't look right, because:

```
>>> isinstance(set(), frozenset)
False
>>> set.__subclasses__()
[]
>>> frozenset.__subclasses__()
[]
```

They are separate types.
CPython solves this by having separate types, and just assigning the same methods to both (and having type checks check for both at the same time), and skipping mutating methods for frozenset. So maybe the type checking system could be extended to allow multiple Python types per payload type?

palaviv:
Sounds like a good solution. I think there are a few other types that can use the same code by using the same payload. bytes and bytearray for example.
This is a big refactor so I will open it on a different PR.

OddCoincidence:
> So maybe the type checking system could be extended to allow multiple Python types per payload type?

I'd rather we not do this. I have a plan for proper inheritance support that maintains the 1-1 relationship between payloads and types, which will be beneficial as a fast path to circumvent typechecking in some cases.

adrian17:
Hm, I can't think of any "in between" solution right now :/
Except, well, skipping "idiomatic" type checks and manually checking
`if obj.typ().is(vm.ctx.set_type()) || obj.typ().is(vm.ctx.frozenset_type())`
But that's going to be quite ugly.

palaviv:
@OddCoincidence what do you think we should do in this case? Should we create a `PyFrozenset` and duplicate the functions?

OddCoincidence:
Yes, I don't think they should be sharing methods even if they are similar given that that is no subclass relationship between them afaict.

Additional, sharing a payload between these two has the same disadvantage as the one I mentioned on your other PR with `bytes` and `bytearray`, which is that the immutable version has to pay the cost of mutability (i.e. `RefCell`, eventually a `Mutex`) even though it has no need for it.

windelbouwman:
In the beginning of the project, I made the mistake to confuse types and payloads. I now think they are not the same. A type is a python type, and the payload is its rust internal data. Actually, if we want to support multiple inheritance, we should be able to support multiple payloads. How, for example would you do this:
```python
class FunkyThing(int, str):
    def __repr__(self):
          return int.__repr__(self) + str.__repr__(self)
```

vm/src/obj/objset.rs
```diff
@@ -181,17 +181,18 @@ fn set_new(
 
 fn set_len(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     trace!("set.len called with: {:?}", args);
-    arg_check!(vm, args, required = [(s, Some(vm.ctx.set_type()))]);
+    arg_check!(vm, args, required = [(s, Some(vm.ctx.frozenset_type()))]);
```
windelbouwman: If changing the type here, also rename the function from `set_len` to `frozenset_len`.

OddCoincidence:
If you want to avoid code duplication, you could create an "inner" set type. (very) roughly:

```rust
struct PySetInner {
    // actual implementation
}

impl PySetInner {
  // actual implementation
}

struct PySet {
    value: RefCell<PySetInner>,
}

struct PyFrozenSet {
    value: PySetInner
}

impl PySetRef {
    fn len(self) -> usize {
        self.value.borrow().len()
    }
}

impl PyFrozetSetRef {
    fn len(self) -> usize {
        self.value.len()
    }
}
```

OddCoincidence:
> Actually, if we want to support multiple inheritance, we should be able to support multiple payloads. How, for example would you do this:

I have a plan to support this kind of thing, but I still strongly believe that we should maintain the 1-1 relationship between `PyValue` types and python types.

vm/src/obj/objset.rs
palaviv: After changing `set` and `frozenset` to be separate types this will not look the same. Let's see where the discussion is going and then I will do fixes accordingly.

windelbouwman:
Examples on why payload is different than python type are:
`set` and `frozenset` take `Vec<PyObjectRef>` as payload. But the same holds for `bytes` and `bytearray` which take `Vec<u8>` as payload. Another example are list and tuple which take `Vec<PyObjectRef>` as payload. The member methods operate on this payload.

That being said, the object system has changed significantly, and this maybe a view on the past.

palaviv:
Closed in favor of #725

PR #717, opened 2019-03-22
Allow mutiple types in PyValue

palaviv:
As suggested by @adrian17 #715. This will allow the same `PyValue`  to be used for multiple types. For example:

- `set` and `frozenset`
- `bytes` and `bytearray`

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -185,8 +185,8 @@ pub struct Frame {
 }
 
 impl PyValue for Frame {
-    fn class(vm: &VirtualMachine) -> PyObjectRef {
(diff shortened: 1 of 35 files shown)


OddCoincidence:
I don't think we should do this. See my comment [here](https://github.com/RustPython/RustPython/pull/715#issuecomment-475673869).

In addition, `bytes` and `bytearray` most certainly should not share the same type, because one is mutable and the other is not, so sharing a payload also means sharing the cost of mutability.

palaviv:
Closed. Please look at #725 for what we can do for `bytes` and `bytearray` as well.

PR #718, opened 2019-03-22
Lets update the demo page, since we have a new shiny console added!

windelbouwman:


coolreader18:
I pushed to master a commit that would have webpack print out "keepalive" every 5 minutes because otherwise Travis would end the build for being "inactive" for 10 minutes, and it looks like it worked!

adrian17:
What's happening for these 10 minutes that doesn't print anything?

coolreader18:
I think it's probably not caching `target`, so it's building every single dependency, and with `--release` to boot. We should probably fix that if that's the case, but even so I think the 5 minute keepalive interval is a great good failsafe.

PR #719, opened 2019-03-22
Type __subclasses__()

adrian17:
Wanted to get a feel of the Weak stuff.

### DIFF
--- a/vm/src/obj/objobject.rs
+++ b/vm/src/obj/objobject.rs
@@ -11,7 +11,7 @@ use crate::pyobject::{
 };
 use crate::vm::VirtualMachine;
 
-#[derive(Clone, Debug)]
+#[derive(Debug)]
 pub struct PyInstance;
 
 impl PyValue for PyInstance {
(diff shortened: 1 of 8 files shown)


cthulahoops:
Cool - I didn't know this existed. I like that you've made this work without a gc whereas cpython needs one.

Looks good to me.

adrian17:
> I like that you've made this work without a gc whereas cpython needs one.

Pretty sure I did implemented this the same way as CPython. I'm assuming additional references in CPython come from some other feature.

vm/src/obj/objobject.rs
```diff
@@ -9,7 +9,7 @@ use crate::pyobject::{
 };
 use crate::vm::VirtualMachine;
 
-#[derive(Clone, Debug)]
```
windelbouwman: Less is more. I like it!

cthulahoops:
Yeah, I suppose this has to be the same as CPython.

Can CPython ever reference count a class out existence? Do classes have a reference to themselves in their own mro?

adrian17:
> Do classes have a reference to themselves in their own mro?

Yeah, pretty sure that's the case.

adrian17:
anything blocking merging this?

PR #720, opened 2019-03-22
Use the extend class macro more.

windelbouwman:


windelbouwman:
Haha, yes, you succeeded :P

PR #721, opened 2019-03-22
Change types in PyContext to be PyClassRef's than PyObjectRef's

skinny121:
This is first, in a series of PR's to remove unnecessary usages of PyObjectRef and replace them with more specific/appropriate ref type. 

Majority of the changes in this PR are actually removing the reference in front of the type in context.set_attr calls. In all cases, the type is already borrowed, so there was no need to borrow it again. But it caused problems when I made set_attr be generic.


skinny121:
Yes, the majority of them will.

PR #722, opened 2019-03-23
Remove Box from PyObject

OddCoincidence:


coolreader18:
Ah, it's the same thing again with WASM `held_rcs`.

OddCoincidence:
I always forget about WASM :sweat_smile:

wasm/lib/src/vm_class.rs
```diff
@@ -24,7 +24,7 @@ pub(crate) struct StoredVirtualMachine {
     pub scope: RefCell<Scope>,
     /// you can put a Rc in here, keep it as a Weak, and it'll be held only for
     /// as long as the StoredVM is alive
-    held_rcs: RefCell<Vec<Rc<dyn HeldRcInner>>>,
+    held_objects: RefCell<Vec<PyObjectRef>>,
```
coolreader18: Yeah, this is probably a better system than what I had. You can just remove `HeldRcInner` now.

coolreader18: Yeah, this is a better system than what I had. You can just remove `HeldRcInner` now.

vm/src/pyobject.rs
```diff
@@ -788,9 +791,30 @@ pub trait IdProtocol {
     }
 }
 
-impl IdProtocol for PyObjectRef {
+#[derive(Debug)]
+enum Never {}
```
coolreader18: What's the use of this?

OddCoincidence: We need a concrete payload type to cast a fat pointer (`*const PyObject<dyn PyObjectPayload>`) to a thin pointer (`*const PyObject<SomeConcreteType>`) so we can use its address for the object id. The ideal type would be the never type `!`, but that's not stable, so creating an enum with no variants is the standard workaround.

PR #725, opened 2019-03-23
frozenset

palaviv:
Add `frozenset` with code shared with `set`. After the discussion in #715 I think we can start by adding this and in the future if we will want to allow multiple types for payload.

### DIFF
--- a/vm/src/obj/objset.rs
+++ b/vm/src/obj/objset.rs
@@ -27,7 +27,7 @@ pub type PySetRef = PyRef<PySet>;
 
 #[derive(Default)]
 pub struct PyFrozenSet {
-    elements: RefCell<HashMap<u64, PyObjectRef>>,
+    elements: HashMap<u64, PyObjectRef>,
 }
 pub type PyFrozenSetRef = PyRef<PyFrozenSet>;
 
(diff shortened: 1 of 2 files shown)


vm/src/obj/objset.rs
```diff
@@ -24,21 +25,80 @@ pub struct PySet {
...
-    obj.payload::<PySet>().unwrap().elements.borrow().clone()
+    if let Some(set) = obj.payload::<PySet>() {
+        return set.elements.borrow().clone();
+    } else if let Some(frozenset) = obj.payload::<PyFrozenSet>() {
+        return frozenset.elements.clone();
+    }
+    panic!("Not frozenset or set");
+}
+
+macro_rules! validate_set_or_frozenset {
```
skinny121: I don't think this needs to be a macro. This could be defined as:
```rust
fn validate_set_or_frozenset(vm: &VirtualMachine, obj: &PyObjectRef) -> PyResult<()> {
     ...
}
```
And use ? to propagate the error.

palaviv: Done

windelbouwman: Agreed, although I am in favor of macros in some places, functions are usually preferred. Good review!

PR #726, opened 2019-03-23
Replace special cases in boolval with __bool__ method on types.

cthulahoops:


adrian17:
While it's a nice cleanup in `boolval`, it currently slows down things like array sorting by a factor of 2.
At the very least, a special case for `obj.is(&vm.ctx.true_value)` (and `false`) would be nice, if not more.

What do you think?

OddCoincidence:
I definitely think we need a fast path for _all_ magic methods for _all_ builtin types. We should wait to do this though until the object representation has been stable for a while, because it will heavily impact how this is implemented.

adrian17:
Well, we'd definitely need a fast path for internally calling Rust magic methods without going through a hashmap and `PyFuncArgs`, but I feel like a trivial check for True and False identity is unrelated to object representation here.

windelbouwman:
I would implement fast paths for the parts that really slow the project down. Try to find some big fish, solve that with fast paths, and leave the rest as it is. Maybe we will overhaul the object structure once more. I sincerely hope not, but you never know.

coolreader18:
To that extent, we should probably set up benchmarks on CI soon. It would also let us see how performance of the project improves over time.

cthulahoops:
I agree with letting things settle before we start another round of refactoring. In general, I think we should focus on completeness/correctness first, and then performance.

That said, if I understand the shape of the magic method optimisation it's fairly self contained:

1. Add a member to the PyClass struct for each magic method. (Or just one to start!)
2. Populate from parents on creation.
3. Setattr updates slots in class and children.
4. Access struct directly instead of using call_method.  (one at a time is fine).

(There's a bit extra around argument passing, binding, ```__get__``` and all that, but I think there's a naive first pass that would be a big improvement.)

[An extra special case in boolval is pretty harmless if you want it here.]

PR #727, opened 2019-03-23
Remove attribute protocol from wasm.

cthulahoops:


vm/src/pyobject.rs
```diff
@@ -896,7 +896,7 @@ where
     }
 }
 
-pub trait AttributeProtocol {
```
windelbouwman: Finally! Its no longer accessible!

PR #732, opened 2019-03-23
Support classmethod super

palaviv:


vm/src/obj/objsuper.rs
```diff
@@ -78,6 +79,10 @@ fn super_getattribute(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
         Some(PyClass { ref mro, .. }) => {
             for class in mro {
                 if let Ok(item) = vm.get_attribute(class.as_object().clone(), name_str.clone()) {
+                    if let Some(PyMethod { .. }) = item.payload() {
```
coolreader18: I didn't know that you could use type inference like this! However, given that you're not using a part of the payload, it'd be better to use `item.payload_is::<_>()`

palaviv: Done

PR #733, opened 2019-03-23
[WASM] Add PyDocument and PyElement classes to browser

coolreader18:
I was going to do `.on` for events, but I realized that it would require (a) a different class for every single possible Event subclass, or (b) a `PyJsValue` class with `__getattr__`, which I'm going to do but in a separate PR.

I also fixed that issue from before with importing the `json` module - `builtins.__import__` tried to get the current frame, and if there wasn't anything on the stack it just failed because `vm.current_frame()` used `frames[idx]` which panics instead of `.get(idx)` which returns a result in the case of an invalid index.

### DIFF
--- a/wasm/lib/src/browser_module.rs
+++ b/wasm/lib/src/browser_module.rs
@@ -1,5 +1,3 @@
-use std::path::PathBuf;
-
 use futures::Future;
 use js_sys::Promise;
 use num_traits::cast::ToPrimitive;
@@ -7,11 +5,13 @@ use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
 use wasm_bindgen_futures::{future_to_promise, JsFuture};
 
-use rustpython_vm::function::PyFuncArgs;
-use rustpython_vm::import::import_module;
-use rustpython_vm::obj::{objint, objstr};
+use rustpython_vm::function::{OptionalArg, PyFuncArgs};
+use rustpython_vm::obj::{
+    objint,
+    objstr::{self, PyStringRef},
+};
 use rustpython_vm::pyobject::{
-    AttributeProtocol, PyContext, PyObject, PyObjectRef, PyResult, PyValue, TypeProtocol,
+    PyContext, PyObject, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
 };
 use rustpython_vm::VirtualMachine;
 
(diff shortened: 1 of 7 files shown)


wasm/lib/src/browser_module.rs
```diff
@@ -270,6 +263,65 @@ fn promise_catch(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     Ok(PyPromise::new_obj(promise_type, ret_promise))
 }
 
+#[derive(Debug)]
+struct PyDocument {
```
OddCoincidence: nit: can this just be `Document` (and same for others)? Outside of the core built-in types, the prefixes just seem like noise to me.

wasm/lib/src/browser_module.rs
```diff
@@ -270,6 +263,65 @@ fn promise_catch(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+}
+type PyElementRef = PyRef<PyElement>;
+
+impl PyValue for PyElement {
+    fn class(vm: &VirtualMachine) -> PyObjectRef {
+        vm.class("browser", "Element")
+    }
+}
+
+fn elem_get_attr(
```
OddCoincidence: I know it won't let you do `impl PyElementRef`, but how about `impl PyElement`? In the future we'll be able to make it follow the pattern once `arbitrary-self-types` are allowed.

coolreader18: Oh, I didn't consider that. Sounds good!

OddCoincidence:
> I was going to do .on for events, but I realized that it would require (a) a different class for every single possible Event subclass, or (b) a PyJsValue class with __getattr__, which I'm going to do but in a separate PR.

The [`gloo`](https://github.com/rustwasm/gloo) project is working on a more idiomatic and type safe wrapper for events, so it might be worth waiting to see if we can base it off of that.

wasm/lib/src/browser_module.rs
coolreader18: Sure, for `PyJsValue` I think I'll keep the prefix but for these I'll drop it.

coolreader18:
It'd still require (probably manually) making Python classes for each possible `Event` struct, so I think it's easiest to just go with a `PyJsValue` for now.

coolreader18:
I see what you mean, using `gloo_events` for Python event listeners. That sounds like a good idea, so I'll wait for that.

PR #734, opened 2019-03-23
Some socket and io cleanups

OddCoincidence:
Moves to using `into_ref{,_with_type}` instead of `PyObject::new` since we are moving away from the latter (and its existence is blocking my `PyRef` work). Also changes from panicking to raising an `Overflow` error if `socket.socket` is called with an int that doesn't fit in an `i32` (via `TryFromObject`).

@palaviv I think you wrote most of `socket.rs`, could you review this?


palaviv:
Looks good to me.

Issue #736, opened 2019-03-23
add bytes missing method


jgirardet:
A thread to add  missing bytes method :
I'll try to keep it up to date

- [x] `bytes . __add__`
- [x] `bytes . __contains__` 
- [x] `bytes . __getitem__` 
- [ ] `bytes . __getnewargs__` 
- [ ] `bytes . __init_subclass__` 
- [x] `bytes . __mod__` 
- [x] `bytes . __mul__` 
- [x] `bytes . __reduce__` 
- [x] `bytes . __reduce_ex__` 
- [x] `bytes . __rmod__` 
- [x] `bytes . __rmul__` 
- [x] `bytes . __sizeof__` 
- [x] `bytes . __subclasshook__` 
- [x] `bytes . capitalize` 
- [x] `bytes . center` 
- [x] `bytes . count` 
- [x] `bytes . decode` 
- [x] `bytes . endswith` 
- [x] `bytes . expandtabs` 
- [x] `bytes . find` 
- [x] `bytes . fromhex` 
- [x] `bytes . hex` 
- [x] `bytes . index` 
- [x] `bytes . isalnum` 
- [x] `bytes . isalpha` 
- [x] `bytes . isascii` 
- [x] `bytes . isdigit` 
- [x] `bytes . islower` 
- [x] `bytes . isspace` 
- [x] `bytes . istitle` 
- [x] `bytes . isupper` 
- [x] `bytes . join` 
- [x] `bytes . ljust` 
- [x] `bytes . lower` 
- [x] `bytes . lstrip` 
- [x] `bytes . maketrans` 
- [x] `bytes . partition` 
- [x] `bytes . replace` 
- [x] `bytes . rfind` 
- [x] `bytes . rindex` 
- [x] `bytes . rjust` 
- [x] `bytes . rpartition` 
- [x] `bytes . rsplit` 
- [x] `bytes . rstrip` 
- [x] `bytes . split` 
- [x] `bytes . splitlines` 
- [x] `bytes . startswith` 
- [x] `bytes . strip` 
- [x] `bytes . swapcase` 
- [x] `bytes . title` 
- [x] `bytes . translate` 
- [x] `bytes . upper` 
- [x] `bytes . zfill`

jgirardet:
I start with  __add__, __contains__, __getitem__

coolreader18:
I made your list into checkboxes, so you can check each one off as you do it. Thanks for helping with this!

jgirardet:
[...]

PR #738, opened 2019-03-23
WIP: :add bytes __add__, __contain__, __getitem__

jgirardet:
TODO at #736


vm/src/obj/objbytes.rs
```diff
@@ -201,3 +202,15 @@ fn bytes_iter(obj: PyBytesRef, _vm: &VirtualMachine) -> PyIteratorValue {
...
 }
+
+impl PyBytesRef {
+    fn add(self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        if objtype::isinstance(&other, &vm.ctx.bytes_type()) {
+            let rhs = get_value(&other);
+            let elements: Vec<u8> = self.value.iter().chain(rhs.iter()).cloned().collect();
+            Ok(vm.ctx.new_bytes(elements))
+        } else {
+            Err(vm.new_type_error(format!("Cannot add {} and {}", self.as_object(), other)))
```
OddCoincidence: This should actually return `Ok(vm.ctx.not_implemented())` so that the vm can attempt the mirrored op (radd).

(A lot of our code is still getting this wrong.)

https://docs.python.org/3/library/constants.html#NotImplemented

jgirardet: ok , we keep this kind of message ?

OddCoincidence: If you return `NotImplemented` the vm will generate the message for you (if that's what you're asking)

jgirardet: thats it

vm/src/obj/objbytes.rs
```diff
@@ -201,3 +203,58 @@ fn bytes_iter(obj: PyBytesRef, _vm: &VirtualMachine) -> PyIteratorValue {
...
+    let b_len = b.len();
+    for (n, i) in a.iter().enumerate() {
+        if n + b_len <= a_len && *i == b[0] {
+            if compare_slice(&a[n..n + b_len], b) {
+                return true;
+            }
+        }
+    }
+    false
+}
```
jgirardet: for compare_vec and compare_slice i didn't find an easier way to do it. maybe it exists ?

coolreader18: If compare_slice is just checking that all the elements in each are equal, I think you could just use `slice_a == slice_b`

vm/src/obj/objbytes.rs
```diff
@@ -201,3 +203,58 @@ fn bytes_iter(obj: PyBytesRef, _vm: &VirtualMachine) -> PyIteratorValue {
...
+        }
+    }
+
+    fn contains(self, needle: PyObjectRef, vm: &VirtualMachine) -> PyResult<bool> {
+        if objtype::isinstance(&needle, &vm.ctx.bytes_type()) {
+            if compare_vec(&self.value, &get_value(&needle)) {
+                return Ok(true);
+            } else {
+                return Ok(false);
+            }
```
windelbouwman: You could use here:
```rust
let result = compare_vec(&self.value, &get_value(&needle));
Ok(result)
```
Notice the lack of `if`-statement, and lack of `return` here.

vm/src/obj/objbytes.rs
```diff
@@ -201,3 +203,58 @@ fn bytes_iter(obj: PyBytesRef, _vm: &VirtualMachine) -> PyIteratorValue {
...
+fn compare_slice(a: &[u8], b: &[u8]) -> bool {
+    for (i, j) in a.iter().zip(b.iter()) {
+        if i != j {
+            return false;
+        }
+    }
+    return true;
+}
+
+fn compare_vec(a: &Vec<u8>, b: &Vec<u8>) -> bool {
```
windelbouwman: I suggest to rename this function to `vec_contains`. The word compare is confusing to me.

windelbouwman:
@jgirardet thank you for your time contributing to this project! I left some review comments, but the change looks good to me.

vm/src/obj/objbytes.rs
```diff
@@ -201,3 +203,58 @@ fn bytes_iter(obj: PyBytesRef, _vm: &VirtualMachine) -> PyIteratorValue {
         iterated_obj: obj.into_object(),
     }
 }
+
+fn compare_slice(a: &[u8], b: &[u8]) -> bool {
```
windelbouwman: There should be a rust builtin function to compare slices: https://doc.rust-lang.org/nightly/std/slice/#trait-implementations in this case, this would mean this function is not required.

vm/src/obj/objbytes.rs
```diff
@@ -201,3 +203,58 @@ fn bytes_iter(obj: PyBytesRef, _vm: &VirtualMachine) -> PyIteratorValue {
...
+    }
+    return true;
+}
+
+fn compare_vec(a: &Vec<u8>, b: &Vec<u8>) -> bool {
+    let a_len = a.len();
+    let b_len = b.len();
+    for (n, i) in a.iter().enumerate() {
+        if n + b_len <= a_len && *i == b[0] {
+            if compare_slice(&a[n..n + b_len], b) {
```
windelbouwman: Replace this by `if &a[n..n + b_len] == &b {` if this is possible, slices should implement the `Eq` trait.

jgirardet:
a more general question : should we use args_cjeck! or type_check! every time it's possible or is the if statement : `if objtype::isinstance(&needle, &vm.ctx.bytes_type())` acceptable ?

coolreader18:
What you did is good, because NotImplemented wouldn't be returned if you used arg_check.

windelbouwman:
@jgirardet I think this work is good. Is it still a work in progress, or is it ready for merging?

jgirardet:
Some tests are disabled due to #746

jgirardet:
ARgh I broke everything with git in the rebase/merge...
I'm gonna start a new clean PR to clean things

PR #739, opened 2019-03-23
Introduce Either extractor and convert range.__getitem__

OddCoincidence:
Note: I intentionally implemented `TryFromObject` only for an `Either` of two `PyRef`s rather than any two types than implement `TryFromObject`. This is because doing the latter would pointlessly allocate an error object whenever the argument was type `B`. By restricting this to `PyRef`s, we can just try to downcast, which is a lot cheaper. 

Example of the error messages this returns:

```
>>>>> range(10).__getitem__('')
TypeError: must be int or slice, not str
```

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -1218,6 +1218,50 @@ impl<T: PyValue + 'static> PyObjectPayload for T {
     }
 }
 
+pub enum Either2<A, B> {
+    A(A),
+    B(B),
+}
+
+/// This allows a builtin method to accept arguments that may be one of two
+/// types, raising a `TypeError` if it is neither.
+///
+/// # Example
+///
+/// ```
+/// fn do_something(arg: Either2<PyIntRef, PyStringRef>, vm: &VirtualMachine) {
+///     match arg {
+///         Either2::A(int)=> {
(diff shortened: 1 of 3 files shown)


vm/src/pyobject.rs
```diff
@@ -700,16 +700,23 @@ where
...
 impl PyObject<dyn PyObjectPayload> {
-    pub fn downcast<T: PyObjectPayload>(self: Rc<Self>) -> Option<PyRef<T>> {
+    /// Attempt to downcast this reference to a subclass.
+    ///
+    /// If the downcast fails, the original ref is returned in as `Err` so
+    /// another downcast can be attempted without unnecessary cloning.
+    ///
+    /// Note: The returned `Result` is _not_ a `PyResult`, even though the
+    ///       types are compatible.
+    pub fn downcast<T: PyObjectPayload>(self: Rc<Self>) -> Result<PyRef<T>, PyObjectRef> {
```
coolreader18: I thought `arbitrary self types` hadn't landed yet?

OddCoincidence: It hasn't for user-defined types, but `Box<Self>` and `Rc<Self>` as self types are allowed.

PR #740, opened 2019-03-24
Use specific PyRef where possible - code object/type_new_class

skinny121:
Following on from my previous PRs, this removes unnecessary uses of `PyObjectRef` in `type_new_class` function and in the areas around the code/frame object.

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -11,7 +11,7 @@ use crate::bytecode;
 use crate::function::PyFuncArgs;
 use crate::obj::objbool;
 use crate::obj::objbuiltinfunc::PyBuiltinFunction;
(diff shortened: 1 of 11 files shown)


vm/src/frame.rs
```diff
@@ -562,7 +564,7 @@ impl Frame {
             }
             bytecode::Instruction::MakeFunction { flags } => {
                 let _qualified_name = self.pop_value();
-                let code_obj = self.pop_value();
```
OddCoincidence: I think this should panic (i.e. `self.pop_value().downcast().expect(...) `) rather than raise a `TypeError`, since a non-code object on the stack here would indicate a compiler bug.

PR #743, opened 2019-03-24
Update cargo dependencies

coolreader18:
I updated the WASM convert::* methods because the Uint8Array methods weren't available in the earlier versions of js_sys.


wasm/lib/src/convert.rs
```diff
@@ -125,12 +125,10 @@ pub fn py_to_js(vm: &VirtualMachine, py_obj: PyObjectRef) -> JsValue {
         || objtype::isinstance(&py_obj, &vm.ctx.bytearray_type())
     {
         let bytes = objbytes::get_value(&py_obj);
-        let arr = Uint8Array::new_with_length(bytes.len() as u32);
-        for (i, byte) in bytes.iter().enumerate() {
-            Reflect::set(&arr, &(i as u32).into(), &(*byte).into())
-                .expect("setting Uint8Array value failed");
+        unsafe {
```
OddCoincidence: What's this `unsafe` needed for?

coolreader18: `Uint8Array::view` is an `unsafe fn` because it provides a direct view into the WASM linear memory; if you were to allocate something with Rust that view would probably become invalid. It's safe because I then copy the array using `Uint8Array::slice`.

OddCoincidence: Okay, could you add that in a comment?

coolreader18: Done.

adrian17: (Also, how is this related to cargo dependencies?)

coolreader18: Oh, its a feature only in more recent js-sys versions, and when I tried to update a few weeks ago everything broke. I figured it'd be a bot more involved getting everything to work, but I guess not:smile:

coolreader18: Specifically `Uint8Array::{view,copy_to}`

PR #748, opened 2019-03-25
Attributes and dictionaries

cthulahoops:
Sorry, this is a bit of monster (especially as I had to change most of AST).

This is the controversial change I've been working towards:

After this:

* Instances (including modules) use real dictionaries for their attributes.
* Classes keep their attributes in a hidden PyAttributes.

We should create fewer dictionaries generally.

### DIFF
--- a/vm/src/obj/objmodule.rs
+++ b/vm/src/obj/objmodule.rs
@@ -1,12 +1,10 @@
-use crate::obj::objstr::PyStringRef;
 use crate::obj::objtype::PyClassRef;
-use crate::pyobject::{DictProtocol, PyContext, PyObjectRef, PyRef, PyResult, PyValue};
+use crate::pyobject::{DictProtocol, PyContext, PyRef, PyResult, PyValue};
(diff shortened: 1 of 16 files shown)


vm/src/obj/objmodule.rs
```diff
@@ -18,23 +16,21 @@ impl PyValue for PyModule {
...
+        } else {
+            panic!("Modules should definitely have a dict.");
+        }
     }
 }
 
 pub fn init(context: &PyContext) {
     extend_class!(&context, &context.module_type, {
         "__dir__" => context.new_rustfunc(PyModuleRef::dir),
-        "__setattr__" => context.new_rustfunc(PyModuleRef::set_attr)
```
windelbouwman: Why should a module not have setattr?

cthulahoops: Because its setattr is now the same as the one it will inherit from object anyway.

windelbouwman:
This work looks go to me, do you wish other reviewers to have a look at it?

PR #749, opened 2019-03-25
Tests cleanups

adrian17:
(Didn't touch cpython_bytecode tests, as they don't seem to be working anyway?)
This massively speeds up time of running all tests, to 5-10 seconds max.


tests/snippets/append.py
```diff
@@ -1,2 +0,0 @@
-x = []
```
windelbouwman: Why did you remove this file? I agree the test makes little sense, but still, it is a valid test.

adrian17: Because other tests in `list.py` do the same job. No point testing the same thing twice.

PR #754, opened 2019-03-26
Upgrade lalrpop

adrian17:
Seems like removal of lalrpop-snap in the new version will improve build times a tiny little bit.


coolreader18:
It looks like lalrpop now outputs to `OUT_DIR` , so you should remove `python.rs` from all the gitignores and just put `include!(concat!(env!("OUT_DIR"), "/python.rs"))` in it.

PR #755, opened 2019-03-27
Use new style args for int.__new__

skinny121:
To achieve this I anhanced FromArgs derive proc macro to support positional arguments. Positional-only arguments have the #[positional] attribute in front of it's field and while keyword-only arguments have #[keyword] attribute in front of it's field.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -146,14 +146,19 @@ fn generate_field(field: &Field) -> TokenStream2 {
         quote! {
             .unwrap_or_else(|| #default)
         }
+    } else if attr.optional {
+        quote! {
+            .map(crate::function::OptionalArg::Present)
+            .unwrap_or(crate::function::OptionalArg::Missing)
+        }
     } else {
         let err = match attr.kind {
-            ArgKind::Positional | ArgKind::PositionalKeyword => {
-                quote!(crate::function::ArgumentError::TooFewArgs)
-            }
-            ArgKind::Keyword => quote!(crate::function::ArgumentError::RequiredKeywordArgument(
-                stringify!(#name)
-            )),
+            ArgKind::Positional | ArgKind::PositionalKeyword => quote! {
+                crate::function::ArgumentError::TooFewArgs
+            },
+            ArgKind::Keyword => quote! {
+                crate::function::ArgumentError::RequiredKeywordArgument(tringify!(#name))
+            },
         };
         quote! {
             .ok_or_else(|| #err)?
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -3,9 +3,9 @@ extern crate proc_macro;
 use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
-use syn::{Data, DeriveInput, Field, Fields};
+use syn::{Attribute, Data, DeriveInput, Expr, Field, Fields, Ident, Lit, Meta, NestedMeta};
 
-#[proc_macro_derive(FromArgs, attributes(positional, keyword))]
+#[proc_macro_derive(FromArgs, attributes(pyarg))]
 pub fn derive_from_args(input: TokenStream) -> TokenStream {
     let ast: DeriveInput = syn::parse(input).unwrap();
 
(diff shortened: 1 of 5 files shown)


vm/src/builtins.rs
```diff
@@ -559,11 +559,22 @@ fn builtin_pow(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     }
 }
 
+// Idea: Should we have a 'default' attribute, so we don't have to use OptionalArg's in this case
```
OddCoincidence: I was thinking the same - in that case maybe we should tweak the design a bit to avoid multiple attributes (and conflicts w/ other proc-macros that use `#[default]`) to something like `#[pyarg(positional, default = "<expr>")]`.

vm/src/builtins.rs
```diff
@@ -559,11 +559,22 @@ fn builtin_pow(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+//    #[keyword] #[default(flush)] flush: bool,
+//}
+
 #[derive(Debug, FromArgs)]
 pub struct PrintOptions {
-    sep: Option<PyStringRef>,
-    end: Option<PyStringRef>,
-    flush: bool,
+    #[keyword]
+    sep: OptionalArg<Option<PyStringRef>>,
```
OddCoincidence: I would prefer to avoid using `OptionalArg` in these:

1. `OptionalArg<Option<...>>` feels very weird
2. We can derive whether an arg is optional or not based on the presence of a future `default` attribute.

vm/src/function.rs
```diff
@@ -218,6 +265,32 @@ pub trait FromArgs: Sized {
     fn from_args(vm: &VirtualMachine, args: &mut PyFuncArgs) -> Result<Self, ArgumentError>;
 }
 
+/// Handling behaviour when the argument is and isn't present, used to implement OptionalArg.
+pub trait ArgHandler: Sized {
```
OddCoincidence: I don't really see why this trait should be necessary...

derive/src/lib.rs
```diff
@@ -3,31 +3,64 @@ extern crate proc_macro;
...
-#[proc_macro_derive(FromArgs)]
+#[proc_macro_derive(FromArgs, attributes(positional, keyword))]
 pub fn derive_from_args(input: TokenStream) -> TokenStream {
     let ast: DeriveInput = syn::parse(input).unwrap();
 
     let gen = impl_from_args(&ast);
     gen.to_string().parse().unwrap()
 }
 
+enum ArgType {
```
OddCoincidence: I think this should be called `ArgKind` (since `Type` has a different meaning) and its variants (and the corresponding attributes) named according to the way Python does: https://docs.python.org/3/library/inspect.html#inspect.Parameter.kind (probably a good idea to include this link as well)

vm/src/builtins.rs
skinny121: That does seem better syntax, with the advantage of being only one attribute.

vm/src/builtins.rs
skinny121: Agreed, OptionalArg<Option<...>> is awkward, but wouldn't be needed when I implement the default attribute argument. I'm split about having OptionalArg in these structs, I think in some cases we need to differentiate between a missing argument and passing None.

vm/src/function.rs
skinny121: I didn't want to use the FromArgs-OptionalArg extractor as it deals with PyFuncArgs, not a single argument, which we have in this case.

vm/src/builtins.rs
OddCoincidence: I think we should handle this entirely with attributes rather than reusing `OptionalArg`. In the following, `None` refers to a Python `None` and `Option::None` refers to Rust's none.

Thing is required, and cannot be `None`.

```rust
#[derive(FromArgs)]
struct MyArgs {
    #[pyarg(positional)]
    thing: ThingRef,
}
```

Thing is required, but can be `None`:

```rust
#[derive(FromArgs)]
struct MyArgs {
    #[pyarg(positional)]
    thing: Option<ThingRef>,
}
```

Thing is not required, can be `None`, and will use the `Default` impl for `ThingRef` if not passed, which is `Option::None`.

```rust
#[derive(FromArgs)]
struct MyArgs {
    #[pyarg(positional, default)]
    thing: Option<ThingRef>,
}
```

Thing is not required, cannot be none, and will use the result of <method that returns ThingRef> if the argument is not present.

```rust
#[derive(FromArgs)]
struct MyArgs {
    #[pyarg(positional, default = "<method that returns ThingRef>")]
    thing: ThingRef,
}
```

vm/src/function.rs
OddCoincidence: If we avoid using `OptionalArg` as I suggested in https://github.com/RustPython/RustPython/pull/755#discussion_r269729006, then I think we can go back to just using `TryFromObject` like it did before.

vm/src/builtins.rs
skinny121: I still think in some situations it's not sufficient to just have a default, some functions need to detect if its argument was provided. For example, if the base argument is provided to int.__new__ then the first argument can then only be a string, otherwise, it can be other types like int or float.

CPython implements int.__new__  by using NULL as the default.  https://github.com/python/cpython/blob/master/Objects/longobject.c#L4950

I have added an 'optional 'argument to pyarg, to indicate when we are want to convert into OptionalArg. e.g.
```rust
#[derive(FromArgs)]
struct IntOptions {
    #[pyarg(positional, optional = true)]
    val_options: OptionalArg<PyObjectRef>,
    #[pyarg(positional_keyword, optional = true)]
    base: OptionalArg<u32>,
}
```

OddCoincidence: > For example, if the base argument is provided to int.new then the first argument can then only be a string, otherwise, it can be other types like int or float.

In cases like these, the "function overloading" trick might work, e.g. range does:

```
fn range_new(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
    let range = if args.args.len() <= 2 {
        let (cls, stop) = args.bind(vm)?;
        PyRangeRef::new(cls, stop, vm)
    } else {
        let (cls, start, stop, step) = args.bind(vm)?;
        PyRangeRef::new_from(cls, start, stop, step, vm)
    }?;

    Ok(range.into_object())
}
```

Basically there can be two different `int_new` methods (that accept different types) decided upon based on arity.

derive/src/lib.rs
```diff
@@ -3,31 +3,198 @@ extern crate proc_macro;
...
+/// (https://docs.python.org/3/library/inspect.html#inspect.Parameter.kind)
+enum ParameterKind {
+    PositionalOnly,
+    PositionalOrKeyword,
+    KeywordOnly,
+}
+
+impl ParameterKind {
+    fn from_ident(ident: &Ident) -> ParameterKind {
+        if ident == "positional" {
```
OddCoincidence: Can we change these to use python's terminology as well? "positional_only", "keyword_only", etc.

OddCoincidence:
Thanks, this is looking really great! I'd still like to figure out a way to avoid `OptionalArg` for that edge case at some point, but I think this is ready to merge with one more small nit addressed (and the conflict resolved).

PR #756, opened 2019-03-27
Convert objobject.rs to new args style

adrian17:


vm/src/obj/objobject.rs
```diff
@@ -31,69 +31,31 @@ pub fn new_instance(vm: &VirtualMachine, mut args: PyFuncArgs) -> PyResult {
...
 }
 
-fn object_eq(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
-    arg_check!(
-        vm,
-        args,
-        required = [(_zelf, Some(vm.ctx.object())), (_other, None)]
-    );
-    Ok(vm.ctx.not_implemented())
+fn object_eq(_zelf: PyObjectRef, _other: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
```
OddCoincidence: Not a comment on this PR, but I realized I was wrong when I changed this to return `NotImplemented`, since it looks like cpython does implement this by comparing object ids:

```
>>> object.__eq__(2, 2)
True
```

PR #759, opened 2019-03-28
Convert bytes to new args style

OddCoincidence:


vm/src/obj/objbytes.rs
```diff
@@ -93,111 +92,71 @@ fn bytes_new(
...
+        self.value.hash(&mut hasher);
+        hasher.finish()
+    }
 
-    let byte_vec = get_value(a).to_vec();
-    Ok(vm.ctx.new_int(byte_vec.len()))
-}
+    fn repr(self, _vm: &VirtualMachine) -> String {
+        // TODO: don't just unwrap
+        let data = String::from_utf8(self.value.clone()).unwrap();
```
coolreader18: We'd need a UnicodeDecodeError class for this.

PR #761, opened 2019-03-28
Convert re to new style args

adrian17:


adrian17:
No idea why codecov didn't comment, but I think it's ready to merge now.

PR #762, opened 2019-03-28
Support class and function __doc__

palaviv:
I tried doing this from the lexer but the doc declaration is a legal statement so I did not figured out a way to distinguish between the two.

### DIFF
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -769,6 +769,7 @@ pub fn make_module(ctx: &PyContext) -> PyObjectRef {
 pub fn builtin_build_class_(vm: &VirtualMachine, mut args: PyFuncArgs) -> PyResult {
     let function = args.shift();
     let name_arg = args.shift();
+    let doc = args.shift();
     let bases = args.args.clone();
     let mut metaclass = if let Some(metaclass) = args.get_optional_kwarg("metaclass") {
         PyClassRef::try_from_object(vm, metaclass)?
@@ -801,5 +802,6 @@ pub fn builtin_build_class_(vm: &VirtualMachine, mut args: PyFuncArgs) -> PyResu
         vec![name_arg, bases, namespace.into_object()],
     )?;
     cells.set_item(&vm.ctx, "__class__", class.clone());
+    vm.ctx.set_attr(&class, "__doc__", doc);
     Ok(class)
 }
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -769,7 +769,6 @@ pub fn make_module(ctx: &PyContext) -> PyObjectRef {
 pub fn builtin_build_class_(vm: &VirtualMachine, mut args: PyFuncArgs) -> PyResult {
     let function = args.shift();
     let name_arg = args.shift();
-    let doc = args.shift();
     let bases = args.args.clone();
     let mut metaclass = if let Some(metaclass) = args.get_optional_kwarg("metaclass") {
         PyClassRef::try_from_object(vm, metaclass)?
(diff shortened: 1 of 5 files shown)


vm/src/frame.rs
```diff
@@ -586,6 +588,7 @@ impl Frame {
                 let obj = vm.ctx.new_function(code_obj, scope, defaults);
 
                 vm.ctx.set_attr(&obj, "__annotations__", annotations);
+                vm.ctx.set_attr(&obj, "__doc__", doc);
```
windelbouwman: Instead of implementing setattr here, we could implement it at the compiler part, by using the instruction StoreAttr on the just created function. This will also prevent empty docstrings in for example lambda functions.

vm/src/compile.rs
```diff
@@ -1411,6 +1462,13 @@ impl Compiler {
         // Fetch code for listcomp function:
         let code = self.pop_code_object();
 
+        // function doc
```
windelbouwman: This can be removed if you change the function and class compilation.

vm/src/compile.rs
```diff
@@ -1125,6 +1168,14 @@ impl Compiler {
                 self.compile_expression(body)?;
                 self.emit(Instruction::ReturnValue);
                 let code = self.pop_code_object();
+
+                // function doc
+                self.emit(Instruction::LoadConst {
+                    value: bytecode::Constant::String {
+                        value: "".to_string(),
```
windelbouwman: This empty doc string can be removed if you implement StoreAttr at the functiondef compilation time.

vm/src/compile.rs
```diff
@@ -689,13 +706,34 @@ impl Compiler {
...
+        self.compile_statements(new_body)?;
         self.emit(Instruction::LoadConst {
             value: bytecode::Constant::None,
         });
         self.emit(Instruction::ReturnValue);
 
         let code = self.pop_code_object();
+
+        // function doc
+        self.emit(Instruction::LoadConst {
```
windelbouwman: Remove this empty doc string.

vm/src/compile.rs
```diff
@@ -597,7 +597,20 @@ impl Compiler {
         self.in_loop = false;
         self.in_function_def = true;
         let mut flags = self.enter_function(name, args)?;
-        self.compile_statements(body)?;
+
+        let doc = get_doc(body);
```
windelbouwman: This code is duplicated with the same code in the class definition part. Try to create a function which performs this split.

vm/src/compile.rs
```diff
@@ -644,6 +657,10 @@ impl Compiler {
             });
         }
 
+        self.emit(Instruction::LoadConst {
```
windelbouwman: Modify the code here something like this: ```rust
if let Some(doc_string) = doc_str {
  self.emit(LoadConst { doc_string });
  self.emit(StoreAttr("__doc__"));
}
// else -> no code emitted!
```

File: vm/src/builtins.rs [hunk 7]
```diff
@@ -766,6 +766,7 @@ pub fn make_module(ctx: &PyContext) -> PyObjectRef {
 pub fn builtin_build_class_(vm: &VirtualMachine, mut args: PyFuncArgs) -> PyResult {
     let function = args.shift();
     let name_arg = args.shift();
+    let doc = args.shift();
```
windelbouwman: See my other comments, this code can be removed if you solve the set attr in the compilation phase.

vm/src/compile.rs
windelbouwman: Btw, micropython is a good source of inspiration for these kind of things: https://github.com/micropython/micropython/blob/master/py/compile.c#L3015

palaviv:
Thanks @windelbouwman. I changed the code as you suggested.

PR #764, opened 2019-03-29
Add #[py_class] attribute proc macro

coolreader18:
All the naming is up for debate, and also the solution for using proc macros inside `rustpython_vm`. I think it's a pretty good system, and I don't think there are really any other crates trying to use a proc macro in their own crate and in consumer crates.


skinny121:
In Rust 1.34, which will be released around mid-April, extern_crate_self will be stabilised(https://github.com/rust-lang/rust/pull/57407#issuecomment-452566839). This allows adding:
```rust
extern crate self as rustpython_vm;
```
Which solves the problem, as we can then write rustpython_vm::xxx from within the crate.

coolreader18:
Oh, nice. This sort of hacky solution works for now, then.

adrian17:
Um... what are the benefits of this?
Because I'm not a fan of going back to `zelf`.

coolreader18:
I would have kept `self`, but using `impl PyStringRef` with `#[py_class]` gave some weird internal rustc compiler error. We could put this on the back burner until `arbitrary_self_types` lands.

vm/src/pyobject.rs
```diff
@@ -1259,6 +1259,26 @@ where
     }
 }
 
+pub trait IntoPyClass {
```
OddCoincidence: If the macro generates an implementation of `IntoPyClass`, then shouldn't it be a derive mode macro?

vm/src/obj/objstr.rs
```diff
@@ -48,97 +48,157 @@ impl TryIntoRef<PyString> for &str {
...
+        OptionalArg::Missing => " ",
+    };
+    if rep_str.len() == 1 {
+        Ok(rep_str)
+    } else {
+        Err(vm.new_type_error("The fill character must be exactly one character long".to_string()))
+    }
+}
+
+#[py_class(
```
OddCoincidence: nit: I'd prefer this be called `pyclass` (that's what PyO3 calls it)

vm/src/obj/objstr.rs
```diff
@@ -48,97 +48,157 @@ impl TryIntoRef<PyString> for &str {
...
+    __inside_vm,
+    name = "str",
+    doc = "str(object='') -> str\n\
```
OddCoincidence: It should be possible to use a regular rustdoc comment, and have the macro use it for `__doc__`. IIRC, `structopt` does something similar for command line help messages.

vm/src/obj/objstr.rs
```diff
@@ -48,97 +48,157 @@ impl TryIntoRef<PyString> for &str {
...
+            OptionalArg::Missing => vm.new_str("".to_string()),
+        };
+        if string.class().is(&cls) {
+            TryFromObject::try_from_object(vm, string)
+        } else {
+            let payload = string.payload::<PyString>().unwrap();
+            payload.clone().into_ref_with_type(vm, cls)
+        }
+    }
+    #[py_class(name = "__add__")]
```
OddCoincidence: Since this isn't defining another class, I think it should be `pymethod`.

OddCoincidence:
@adrian17 

> Um... what are the benefits of this?

Some that I can think of are:

- It would provide a very nice way for users to implement their own native extensions for `rustpython`
- We could derive `__doc__` from rust doc comments
- It makes it a bit clearer which methods are python methods and which are rust-only helpers.

@coolreader18 

> We could put this on the back burner until arbitrary_self_types lands.

I think that is the right thing to do. Also, while I think this is a good thing to have (see above reasons) I'm not _totally_ convinced that using it for builtins / rustpython internals is the right thing to do.

adrian17:
Oh, I was about to write a new response, so...

I'm still not seeing benefits fully. Is it just moving type attribute definitions from `init` to attributes next to methods? I'm not sure this is worth all the extra complexity.

> It would provide a very nice way for users to implement their own native extensions for rustpython

How does it make it easier?

> We could derive __doc__ from rust doc comments

I'd be cautious about putting Python-user-visible strings in the place that's normally used for implementation-level comments and docs.

> It makes it a bit clearer which methods are python methods and which are rust-only helpers.

How? Before seeing your comment I was about to say that it's more confusing now; it seems like all methods are now implicitly exposed with their function name by default?

vm/src/pyobject.rs
coolreader18: No, because you can't `#[derive()]` on an `impl` block.

OddCoincidence:
> How does it make it easier?

Less familiarity with `rustpython`-specific APIs would be required. Ideally for native extensions the declarations are as close to just writing pure, idiomatic rust as possible. Having to write a bit of imperative code for each method feels like it's in opposition of that goal. PyO3 does something similar and it's been very positively received.

> I'd be cautious about putting Python-user-visible strings in the place that's normally used for implementation-level comments and docs.

Implementation-level comments would use `//`, docs would use `///`

> How? Before seeing your comment I was about to say that it's more confusing now; it seems like all methods are now implicitly exposed with their function name by default?

Methods should definitely not be implicitly exposed. For example:

```rust
impl MyNativeType {
    fn helper_rust_only() {
        ...
    }

    #[pymethod]
    fn method_callable_from_python(self: PyRef<Self>) {
        self.helper_rust_only();
    }
}
```

The benefit is that you don't have to jump all the way to the `init` / `make_module` / whatever to figure out which ones are actual python methods. (In this example, the `self: PyRef<Self>` also serves that purpose, but I imagine eventually we could make it so we just take `&self` for methods that don't actually need to consume the ref).

adrian17:
> Implementation-level comments would use //, docs would use ///

Alright.

> Methods should definitely not be implicitly exposed. For example:

I was commenting about the current version of the code, which does it implicitly - if methods (and properties? :) ) become explicit, that'll be nice indeed.

> The benefit is that you don't have to jump all the way to the init / make_module / whatever to figure out which ones are actual python methods.

I'm speaking mostly from my minor experience with CPython, where I rarely had a problem of "is this an exposed method?", and more often of "which (magic) methods does this type expose/override?". For this use case I felt like a centralized spot like this allowed for quicker lookups.

https://github.com/python/cpython/blob/2f54908afc5665937d763510b4430f10cf764641/Objects/tupleobject.c#L826-L833

Anyway, with explicit `#[pymethod]` and allowing for `self` I'll definitely feel much warmer towards this :)

coolreader18:
Alright, I'll work on that then.

adrian17:
> I would have kept self, but using impl PyStringRef with #[py_class] gave some weird internal rustc compiler error.

BTW, are you sure this isn't https://github.com/rust-lang/rust/issues/59406 ? Me and @OddCoincidence  encountered this multiple times, it's relatively easy to trigger when doing "covert X to new style args" commits. This appears to be a bug with error reporting, and can be encountered when there's a coding error (referring to `PyXRef::func` when `func` doesn't exist, but there happens to exist `PyYRef::func`). If this indeed is it, it may mean that you can actually make this work with PystringRef, just need to fix some bugs.

coolreader18:
Yes, it was that. How did you avoid it?

adrian17:
More specifically, for example, it occurs when eg I did by accident

```
impl PyTupleRef {
    fn tuple_count(...args...) {...}
}
...
extend_class!(context, tuple_type, {
    "count" => context.new_rustfunc(PyTupleRef::count),
}
```

Note the mismatch between `tuple_count` and `count` names. Rust can't find the `PyTupleRef::count` method, somewhere during the search it somehow stumbles upon `PyListRef::count` and ICEs. The solution was simple, fix the name `tuple_count` -> `count`.

adrian17:
Oh, you did it :)

vm/src/obj/objstr.rs
```diff
@@ -187,7 +247,7 @@ impl PyStringRef {
 
     fn split(
         self,
-        pattern: OptionalArg<Self>,
+        pattern: OptionalArg<PyStringRef>,
```
adrian17: I think you can go back to using `Self` in places like here too, right?

jmhain: Personally I prefer `PyStringRef` here. `Self` makes sense for an argument when you want to emphasize that you are taking another object of the same type, such as for `eq` and such, but here it's more of a coincidence that `pattern` happens to also be a string.

OddCoincidence: Personally I prefer PyStringRef here. Self makes sense for an argument when you want to emphasize that you are taking another object of the same type, such as for eq and such, but here it's more of a coincidence that pattern happens to also be a string.

adrian17: :+1:

coolreader18:
Okay, I think I've addressed all concerns and the current version looks much nicer and more explicit.

vm/src/obj/objstr.rs
```diff
@@ -48,7 +48,39 @@ impl TryIntoRef<PyString> for &str {
     }
 }
 
+#[pyclass(__inside_vm, name = "str")]
```
OddCoincidence: It kinda bothers me that this (and the doc comment) is on the impl block instead of the struct - is there any way to make that work?

coolreader18: Not the `#[pyclass]` attribute for methods, but if we made another trait `PyDocItem` or something then we could put the doc comment on the struct definition. I'd say it's probably not worth it.

jmhain: For the impl block, PyO3 uses `#[pymethods]` and `#[pyclass]` goes on the struct. I think we should imitate that design.

derive/src/lib.rs
```diff
@@ -213,3 +247,180 @@ fn impl_from_args(input: &DeriveInput) -> TokenStream2 {
         }
     }
 }
+
+#[proc_macro_attribute]
+pub fn pyclass(attr: TokenStream, item: TokenStream) -> TokenStream {
```
jmhain: nit: could we put each proc_macro in its own submodule? this file is getting kinda big.

vm/src/obj/objstr.rs
OddCoincidence: For the impl block, PyO3 uses #[pymethods] and #[pyclass] goes on the struct. I think we should imitate that design.

derive/src/lib.rs
OddCoincidence: nit: could we put each proc_macro in its own submodule? this file is getting kinda big.

vm/src/lib.rs
```diff
@@ -30,7 +30,7 @@ extern crate statrs;
 
 extern crate rustpython_parser;
 #[macro_use]
-extern crate rustpython_derive;
+pub extern crate rustpython_derive;
```
OddCoincidence: Why did this have to be made public?

derive/src/lib.rs
```diff
@@ -213,3 +247,180 @@ fn impl_from_args(input: &DeriveInput) -> TokenStream2 {
...
+                    if let Lit::Str(s) = name_value.lit {
+                        class_name = Some(s.value());
+                    } else {
+                        panic!("#[pyclass(name = ...)] must be a string");
+                    }
+                }
+            }
+        }
+    }
+    let class_name = class_name.expect("#[pyclass] must have a name");
```
OddCoincidence: If `#[pyclass]` was on the struct we could default to the name of the struct

vm/src/lib.rs
coolreader18: If we're using it in `rustpython_vm`, we might as well expose it so that consumers of `rustpython_vm` don't have to add it as a separate dependency. I could do `#[doc(hidden)]`, though.

vm/src/obj/objstr.rs
OddCoincidence: Actually I think maybe `#[pyimpl]` would be better? (since the impl block can contain more than just methods)

vm/src/lib.rs
OddCoincidence: Should we reexport the macros specifically then instead of the crate? (I think that's possible as of rust 2018)

vm/src/obj/objstr.rs
coolreader18: Something like:
```rust
#[pyclass]
/// doc
struct PyString {
    value: String,
}
type PyStringRef = PyRef<PyString>;
#[pymethods]
impl PyStringRef {
    // ...
}
```
The problem is that the type referenced in the `impl` block is different from the type in the `struct` declaration, so it won't be able to associate easily between those two proc macros.

OddCoincidence: As a short-term solution, maybe `pymethods` can also take the name of the class?

(When arbitrary-self-types is available, we can go back to `impl PyString` and clean it up.)

coolreader18:
I think that would be possible, and it would make code reuse like that much easier and more intuitive.

vm/src/obj/objstr.rs
coolreader18: I think for now I'll just keep it as is, because I don't think there's a lot of incentive to split it up between the two for not much return.

coolreader18:
I'll work on trait `#[pymethods]` in a separate PR.

vm/src/obj/objstr.rs
OddCoincidence: I really don't think we should merge this without moving the `#[pyclass]` annotation to the type / struct.

- The type is what you are declaring as representing a python class, not the impl block.
- It's inconsistent with similar projects in the ecosystem.
- If we are going to allow `#[pymethod]` on methods in trait impls as well, which one would get the `#[pyclass]`?

In general, I think we need to take more time to work out the kinks for impactful changes like this to avoid unnecessary churn.

coolreader18: > In general, I think we need to take more time to work out the kinks for impactful changes like this to avoid unnecessary churn.

I agree with that, but ~~it's really an implementation problem with doing separate `#[pyclass]` and `#[pyimpl]`. If we transform:~~

```rust
#[pyclass]
/// docstring
struct PyString {
    value: String,
}
```
to
```rust
struct PyString {
    value: String,
}
impl PyClassTrait for PyString {
    const DOC: &'static str = "docstring";
}
```
~~Then there's no way to access that implementation for PyString when `#[pyimpl]` is on a `impl PyStringRef` block - Py. We have three options here:~~
* ~~Keep as is, with the~~

Nevermind, we could `impl<T> PyClassTrait for PyRef<T> where T: PyClassTrait { const DOC: &'static str = T::DOC; }`. I'll do that.

coolreader18: Not quite sure on what to call "`PyClassTrait`" either, do you have any ideas?

windelbouwman: I would prefer to keep rust doc strings for rust, and add python docstrings explicitly, like this:
```rust
#[pymethod(name="add", doc="Adds two things (this is a python docstring)")
// This is rust documentation
fn add(a: PyObjectRef, b: PyObjectRef) -> PyResult {
}
```

vm/src/builtins.rs
```diff
@@ -536,6 +536,7 @@ fn builtin_pow(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
 }
 
 #[derive(Debug, FromArgs)]
+#[__inside_vm]
```
windelbouwman: What does this line do? Could you give it a more descriptive name?

vm/src/pyobject.rs
```diff
@@ -1259,6 +1259,26 @@ where
...
+    const NAME: &'static str;
+    const DOC: Option<&'static str> = None;
+
+    fn _extend_class(ctx: &PyContext, class: &PyClassRef);
+
+    fn extend_class(ctx: &PyContext, class: &PyClassRef) {
```
windelbouwman: Make this function private, and rename the underscore one.

vm/src/pyobject.rs
```diff
@@ -1259,6 +1259,26 @@ where
     }
 }
 
+pub trait IntoPyClass {
+    const NAME: &'static str;
+    const DOC: Option<&'static str> = None;
+
+    fn _extend_class(ctx: &PyContext, class: &PyClassRef);
```
windelbouwman: Renamed this function to not underscore, since this function must be implemented by trait implementors. Allthough this is rust, many coders will come from python background, where the single underscore prefix indicates some privateness.

windelbouwman:
The macro implementation is beyond my understanding. I will have to learn this.

A single list of functions for an object has its benefits. Marking a method as such, also has benefits. I'm fine with both. In general, I'm for keeping things simple at the cost of some extra lines of code.

That being said, I leave it to others to judge this PR. For me, its fine :).

vm/src/builtins.rs
coolreader18: For now it's just to indicate that to refer to `rustpython_vm` we should use `crate::` instead of `::rustpython_vm::`. Once rust-lang/rust#57407 lands, we'll be able to get rid of it.

vm/src/obj/objstr.rs
coolreader18: The thing is, the functions probably aren't going to need Rust documentation, as they're just implementation details for the final Python API.

windelbouwman: Yes, I agree that this is the case. Still I think it is better to explicitly state that the documentation is for the python side.

OddCoincidence: @windelbouwman I disagree.

- Putting the docs in a `#[pyclass(doc = "...")]` is pretty awkward to me because you have to manually specify the newlines. Also multi-line attributes are fairly uncommon and look out of place.
- We can use regular `//` comments for implementation-level comments.
- Using `///` comments also means that developers of native extensions can see the python docs from rustdoc.
- It has precedent in the rust community - many binding generators do the same thing.

coolreader18:
@windelbouwman is the naming better now? I renamed `_extend_class` to `impl_extend_class`, which I feel better explains what it does, while `extend_class` is a little convenience method that should be called from the outside in most cases.

coolreader18:
@adrian17 I have trait support (probably) ready to go under `coolreader18/pyimpl-on-trait` with fbe37d0749130f3ac0cf9637d935797b121cc715, but I'll probably wait on a PR until #765 or #770 are merged so that I can try it out on a real "Protocol".

coolreader18:
@windelbouwman @OddCoincidence Is this good to merge now?

PR #765, opened 2019-03-29
Refactor objset

palaviv:
I tried to make the code sharing between `set` and `frozenset` more nice. I created a shared trait `SetProtocol` that implements all the shared function. I also moved the `set` only functions to the new format.
I was not sure how to move the `new` implementation to `SetProtocol` if anyone has suggestions?

### DIFF
--- a/vm/src/obj/objset.rs
+++ b/vm/src/obj/objset.rs
@@ -195,6 +195,15 @@ where
 
         self.create(vm, elements)
     }
+
+    fn iter(self, vm: &VirtualMachine) -> PyIteratorValue {
+        let items = self.get_elements().values().cloned().collect();
+        let set_list = vm.ctx.new_list(items);
+        PyIteratorValue {
+            position: Cell::new(0),
+            iterated_obj: set_list,
+        }
+    }
 }
 
 impl SetProtocol for PySetRef {
@@ -514,15 +523,6 @@ enum SetCombineOperation {
     Difference,
 }
 
-fn set_iter(zelf: PySetRef, vm: &VirtualMachine) -> PyIteratorValue {
-    let items = zelf.elements.borrow().values().cloned().collect();
-    let set_list = vm.ctx.new_list(items);
-    PyIteratorValue {
-        position: Cell::new(0),
-        iterated_obj: set_list,
-    }
-}
-
 fn frozenset_repr(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     arg_check!(vm, args, required = [(o, Some(vm.ctx.frozenset_type()))]);
 
@@ -583,7 +583,7 @@ pub fn init(context: &PyContext) {
         "__isub__" => context.new_rustfunc(PySetRef::isub),
         "symmetric_difference_update" => context.new_rustfunc(PySetRef::symmetric_difference_update),
         "__ixor__" => context.new_rustfunc(PySetRef::ixor),
-        "__iter__" => context.new_rustfunc(set_iter)
+        "__iter__" => context.new_rustfunc(PySetRef::iter)
     });
 
     let frozenset_type = &context.frozenset_type;
@@ -613,6 +613,7 @@ pub fn init(context: &PyContext) {
         "__len__" => context.new_rustfunc(PyFrozenSetRef::len),
         "__doc__" => context.new_str(frozenset_doc.to_string()),
         "__repr__" => context.new_rustfunc(frozenset_repr),
-        "copy" => context.new_rustfunc(PyFrozenSetRef::copy)
+        "copy" => context.new_rustfunc(PyFrozenSetRef::copy),
+        "__iter__" => context.new_rustfunc(PyFrozenSetRef::iter)
     });
 }
--- a/vm/src/obj/objset.rs
+++ b/vm/src/obj/objset.rs
@@ -231,6 +231,48 @@ impl SetProtocol for PyFrozenSetRef {
     }
 }
 
+impl PySetRef {
+    fn add(self, item: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        insert_into_set(vm, &mut self.elements.borrow_mut(), &item)
+    }
+
+    fn remove(self, item: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        fn remove(
+            vm: &VirtualMachine,
+            elements: &mut HashMap<u64, PyObjectRef>,
+            key: u64,
+            value: &PyObjectRef,
+        ) -> PyResult {
+            match elements.remove(&key) {
+                None => {
+                    let item_str = format!("{:?}", value);
+                    Err(vm.new_key_error(item_str))
+                }
+                Some(_) => Ok(vm.get_none()),
+            }
+        }
+        perform_action_with_hash(vm, &mut self.elements.borrow_mut(), &item, &remove)
+    }
+
+    fn discard(self, item: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        fn discard(
+            vm: &VirtualMachine,
+            elements: &mut HashMap<u64, PyObjectRef>,
+            key: u64,
+            _value: &PyObjectRef,
+        ) -> PyResult {
+            elements.remove(&key);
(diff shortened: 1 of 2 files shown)


vm/src/obj/objset.rs
```diff
@@ -57,6 +57,336 @@ impl PyValue for PyFrozenSet {
     }
 }
 
+trait SetProtocol
+where
+    Self: Sized,
+{
+    fn get_elements(&self) -> HashMap<u64, PyObjectRef>;
```
adrian17: we definitely don't want to copy the entire map just to eg check its size.

palaviv: This is the same as the current implementation. I would prefer not changing the functionality of the code in this PR to make it easier to review. This is a big diff and I think improving in a different PR will be better.

I did add `__iter__` to `frozenset` but used the same code as already existed in `set`.

OddCoincidence:
Sharing functions between python classes with a trait doesn't really seem like the right approach to me, because now the structure of the rust code does not directly map to that of the python code it represents.  Also I don't think this meshes super well with the proc-macro approach to class definition being developed in #764. I understand the desire to have them share code, but imo [duplication is far cheaper than the wrong abstraction](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction?duplication).

adrian17:
> Also I don't think this meshes super well with the proc-macro approach to class definition being developed in #764 

Side note, I asked about it in comment https://github.com/RustPython/RustPython/pull/764#pullrequestreview-220835307 and @coolreader18 thinks it can work.

> Sharing functions between python classes with a trait doesn't really seem like the right approach to me, because now the structure of the rust code does not directly map to that of the python code it represents. 

Now we're going back to my topic from that PR - I do think they don't necessarily should map anyway.
I do imagine that long-term people will want to do stuff like "here's some magic trait/macro that makes the whole iterator type and `__iter__` boilerplate for you, you just need implement `__next__`".

However, I do agree with your last statement that this deduplication may possibly complicate code in the long run, while not giving proportional advantages. We can also see that in https://github.com/RustPython/RustPython/pull/770#discussion_r270683061 - we're again figuring out reference passing, while recent refactors let us drop `get_elements` not worry about this extra layer at all. So yeah, you made me feel less positive about these two PRs.

jgirardet:
I'd be interested in an answer  about it (share a trait between frozentset and set) because I was on the point to do it for bytes  and bytearray.

cthulahoops:
This feels like the wrong abstraction to me too. As does the existing code, I think the new code is probably an improvement.

Everything feels a bit inside out.  It feels to me like we want: a rust set implementation that provides the basic interface for sets in pure rust, and then some repetitive simple wrappers that unwrap/wrap from both types. I think it would be simpler but more repetitive. (This is hard/be sure it makes sense without trying it.)

palaviv:
> Everything feels a bit inside out. It feels to me like we want: a rust set implementation that provides the basic interface for sets in pure rust, and then some repetitive simple wrappers that unwrap/wrap from both types. I think it would be simpler but more repetitive. (This is hard/be sure it makes sense without trying it.)

I am not sure I understand what your suggestion. Are the wrappers written in python?

OddCoincidence:
@cthulahoops is this something like what you were suggesting? I see this "<Type>Inner" pattern in a lot of rust code.

```rust
struct PySet {
    inner: RefCell<PySetInner>,
}

impl PySetRef {
    fn some_set_method(self) {
        self.inner.borrow().some_set_method()
    }
}
```

```rust
struct PyFrozenSet {
    inner: PySetInner,
}

impl PySetRef {
    fn some_set_method(self) {
        self.inner.some_set_method()
    }
}
```

```rust
struct PySetInner {
     // actual set represetation
}

impl PySetInner {
    fn some_set_method(&self) {
        // the real implementation
    }
}
```

palaviv:
This look like a nice way to go @OddCoincidence. I have a few points:

1. How would you implement all the functions that are `set` only. `update` for example.
2. When you will need to create a new object for example in `union`. Would you pass the function the class of the caller?
3. This will create a lot of duplicated wrapper code that will look the same in `PySetRef` and `PyFrozenSet` impl. Using the trait remove all this duplicate code.

OddCoincidence:
> How would you implement all the functions that are set only. update for example.

```rust
impl PySetInner {
    fn update(&mut self, other: PyIterable) {
        // do it
    }
}

impl PySetRef {
    fn update(self, other: PyIterable) {
        self.inner.borrow_mut().update(other)
    }
}

impl PyFrozenSetRef {
    // no update method
}
```
> When you will need to create a new object for example in union. Would you pass the function the class of the caller?

No, PySetInner would never reference PySet or PyFrozenSet, it would operate in terms of itself.

```rust
impl PySetInner {
    fn union(&self, other: &Self) -> PySetInner {
        // do it
    }
}

impl PySetRef {
    fn union(self, other: PySetRef) -> PySet {
        self.inner.borrow().union(&other).into() // via appropriate From impl
    }
}

impl PyFrozenSetRef {
    fn union(self, other: PySetRef) -> PyFrozenSet {
        self.inner.union(&other).into() // via appropriate From impl
    }
}
```

> This will create a lot of duplicated wrapper code that will look the same in PySetRef and PyFrozenSet impl. Using the trait remove all this duplicate code.

I am 100% okay with this. They are different types, the methods should be explicitly specified for each one.

cthulahoops:
Sorry for not being clearer in my last rushed post. Thanks, @OddCoincidence, that's exactly what I was thinking.

cthulahoops:
(I'm not sure I like the name ```inner```, I would have gone with something like ```SetElements``` and ```.elements```, but this does appear to be a common rust naming scheme.)

jgirardet:
Why does this use HashMap instead of Hashset ?

cthulahoops:
So that we can store arbitrary hashable objects in sets. It's a map of hash value to ```PyObjectRef```.

palaviv:
I will start working on `objset` with `inner` as suggested by @OddCoincidence.

OddCoincidence:
> So that we can store arbitrary hashable objects in sets. It's a map of hash value to PyObjectRef.

How do we handle hash collisions?

cthulahoops:
> How do we handle hash collisions?

We don't. (I tried to work out the answer to this myself earlier.)

```
>>>>> {A(43), A(4)}
{4}
```

I guess once we have proper dicts working we can use the same approach to fix sets.

palaviv:
> ```rust
> impl PySetRef {
>     fn union(self, other: PySetRef) -> PySet {
>         self.inner.borrow().union(&other.inner).into() // via appropriate From impl
>     }
> }
> ```

I have started working on the implementation and I seem to have the same problems as before. The `other` needs to be `PyObjectRef` as it can be either `set` or `frozenset`. This keep me again with `get_elements` (or `get_inner` now). I think #774 macro might help here but I am not sure.

OddCoincidence:
> The other needs to be PyObjectRef as it can be either set or frozenset.

Ah, I didn't consider this. Yeah, using `PyObjectRef` is probably the best thing to do right now. We could eventually add a `SetLike` arg extractor or something.

> I think #774 macro might help here but I am not sure.

I think this will help too. @cthulahoops Are you okay with me merging #774 or do you still have concerns with it?

jgirardet:
I thought Pysetref was mandatory since user could call union with any type, not only set or frozenset

OddCoincidence:
Actually, @palaviv for `union` it looks like you should use `PyIterable`.

```
>>> {1, 2, 3}.union([4])
{1, 2, 3, 4}
>>> {1, 2, 3}.union(4)
TypeError: 'int' object is not iterable
```

OddCoincidence:
> I guess once we have proper dicts working we can use the same approach to fix sets.

We might also be able to implement sets as dicts where the value is `None`, kind of like Rust's `HashSet` is implemented internally with `HashMap<T, ()>`.

cthulahoops:
> I think this will help too. @cthulahoops Are you okay with me merging #774 or do you still have concerns with it?

You mostly convinced me. I'm okay with you merging that.

palaviv:
Was implemented using inner in #785.

PR #766, opened 2019-03-30
Fix #746: invalid slice with start or stop =-1 when step<0

### LINKED ISSUE #746: [bug] get_slice_items fails with negative step
I was struggling adding getitem to bytes using get_slice_items but it appears to also fails with string:

```shell 
jimmy@x-or:~/trash/RustPython$ cargo run -- -c "print('result : ', 'abcd'[-1:-5:-1])"
    Finished dev [unoptimized + debuginfo] target(s) in 0.23s                                                                                                                                     
     Running `target/debug/rustpython -c 'print('\''result : '\'', '\''abcd'\''[-1:-5:-1])'`
result :  
jimmy@x-or:~/trash/RustPython$ python -c "print('result : ', 'abcd'[-1:-5:-1])"                                                                                                                    
result :  dcba
```

and  list :
``` shell
jimmy@x-or:~/trash/RustPython$ python -c "print('result : ', [1,2,3,5,6][-1:-5:-1])"                                                                                                              
result :  [6, 5, 3, 2]
jimmy@x-or:~/trash/RustPython$ cargo run -- -c "print('result : ', [1,2,3,5,6][-1:-5:-1])"                                                                                                         
    Finished dev [unoptimized + debuginfo] target(s) in 0.41s                                                                                                                                     
     Running `target/debug/rustpython -c 'print('\''result : '\'', [1,2,3,5,6][-1:-5:-1])'`
result :  []
```
I can"t tell what's wrong for now

jgirardet:
this one was tricky. :sob: 
I let the tests used to track the bug.


tests/snippets/slice_res.py
```diff
@@ -0,0 +1,2252 @@
+SLICES_RES = [
```
coolreader18: Could this be added to a .gitignore, if it's automatically generated during testing? Also, why does this need to be generated as a file? Why can't you just use a generated list?

jgirardet: No the file is generated by cpython to compare cpython results with rustpython

jgirardet: but indeed we can generate it on the fly with a subprocess.checkouput. i'll do it

jgirardet: I was thinking we could generate it on the fly wichh a subprocess.chekoutput but it's not yet implemented so we have to build it before manually.
Maybe adding a fixture/resources directory in tests to put this kind of file

vm/src/obj/objsequence.rs
```diff
@@ -86,8 +85,20 @@ pub trait PySliceableSequence {
                 } else {
                     // calculate the range for the reverse slice, first the bounds needs to be made
                     // exclusive around stop, the lower number
-                    let start = start.as_ref().map(|x| x + 1);
-                    let stop = stop.as_ref().map(|x| x + 1);
+                    let start = start.as_ref().map(|x| {
+                        if *x == (-1).to_bigint().unwrap() {
+                            self.len() + BigInt::one() //.to_bigint().unwrap()
```
windelbouwman: Should the variable x not be used in this equation?

vm/src/obj/objsequence.rs
```diff
@@ -86,8 +85,20 @@ pub trait PySliceableSequence {
...
+                        } else {
+                            x + 1
+                        }
+                    });
+                    let stop = stop.as_ref().map(|x| {
+                        if *x == (-1).to_bigint().unwrap() {
```
windelbouwman: Does it make sense to have a special case for -1? I do not understand what is happening here.

jgirardet: I'll try to explain but it's not easy. look at it , lhs is what python  is doing and rhs is the way we are trying to do it (without fix).
```python
In [26]: "abcd"[:-1:-1], "abcd"[0:][::-1] # stop without fix                    
Out[26]: ('', 'dcba')

In [27]: "abcd"[:-1:-1], "abcd"[5:][::-1] #stop with fix                        
Out[27]: ('', '')

In [28]: "abcd"[-1::-1], "abcd"[:0][::-1] # start without fix                   
Out[28]: ('dcba', '')

In [29]: "abcd"[-1::-1], "abcd"[:4+1][::-1] # start with fix                    
Out[29]: ('dcba', 'dcba')
```
```python
In [43]: for i in range(-5,6): 
    ...:     print(i, "|",   'abcd'[:i:-1], "|", "abcd"[i+1:])                  
-5 | dcba | abcd
-4 | dcb | bcd
-3 | dc | cd
-2 | d | d
-1 |  | abcd
0 | dcb | bcd
1 | dc | cd
2 | d | d
3 |  | 
4 |  | 
5 |  | 

In [44]: for i in range(-5,6): 
    ...:     print(i, "|",   'abcd'[i::-1], "|", "abcd"[:i+1])                  
-5 |  | 
-4 | a | a
-3 | ba | ab
-2 | cba | abc
-1 | dcba | 
0 | a | a
1 | ba | ab
2 | cba | abc
3 | dcba | abcd
4 | dcba | abcd
5 | dcba | abcd
```

look at those example maybe it'll help to understand

jgirardet: i would add that about slice, None is not the same thing as 0. and if think BigInt turns None to 0 ?

tests/snippets/slice_res.py
adrian17: How about building it in `tests_snippets.py` test `setUpClass`, then simply try `import`ing the generated file in one of the snippets?

jgirardet: I was testing with cargo run tests/snippets/builtin_slice.py in compraison with vith python tests/snippets/builtin_slice.py. I don't know how to  use test_snippet.py with it.

adrian17: > I don't know how to use test_snippet.py with it.

`pytest -k builtin_slice` will run (and test setup steps) and test `builtin_slice.py`  for both CPython and RPy.

jgirardet: ah ok thanks.
you mean something like that ? :
```python
def generate_slices():
    a = ["some things"]
    return "SLICE_RES={}".format(a)


@populate('cpython')
# @populate('cpython_bytecode')
@populate('rustpython')
class SampleTestCase(unittest.TestCase):
    @classmethod
    def setUpClass(cls):
        from pathlib import Path
        cls.p = Path(__file__).parent /"snippets"/'some_name_for_slices_res_file.py'
        cls.p.write_text(generate_slices())

    @classmethod
    def tearDownClass(cls):
        cls.p.unlink()
```
since slice_res.py should never change, If find that the previous code add some useless complexity
I really a think a submodule `ressource` inside snippets could be fine and simple.

adrian17: Yeah, something like that.
> I really a think a submodule ressource inside snippets could be fine and simple.
That's fine too. Personally I'd prefer it to be autogenerated (makes repo and searching code simpler; also the generated code could live in `resources` too), but I don't have a strong opinion on that.

jgirardet:
I refactored the tests to autogenerate it.
cpython passes test
rustpython fails.  it's because I import my resource file like this
`from resources.cpython_generated_slices import SLICES_RES, START, END, STEP, LL`
rustpython seems not be able to read this. I'm not sure where to go to try to fix it in rustpython code

coolreader18:
You could use `__import__` and just use a direct path to the resources file.

vm/src/obj/objsequence.rs
windelbouwman: Thank you for your detailed explanation, still, I find this hard to understand.

Did you test with a stop value of `-2`?

jgirardet: yes the added test, tests every possible combination from
 start = range(-7, 7) to   end = range(-7, 7) with   step =range(-5, 5) for sequence of len = 4

tests/snippets/slice_res.py
jgirardet: done

jgirardet:
Only coverage is failing, I don't know why.

cthulahoops:
There's a problem doing a clean build on nightly, so coverage tests are currently failing for everything at the moment.

jgirardet:
Is there anything else todo here ?

PR #767, opened 2019-03-30
Refactor slice.__new__ to new style function

skinny121:
Also made the following changes:
- Allowed slice to contain arbitrary objects
- Implemented calling \_\_index\_\_ if slice property isn't an int


vm/src/obj/objslice.rs
```diff
@@ -1,18 +1,19 @@
...
 use crate::vm::VirtualMachine;
 
-use super::objint;
-use crate::obj::objtype::PyClassRef;
+use crate::obj::objint::PyInt;
+use crate::obj::objtype::{class_has_attr, PyClassRef};
+use num_bigint::BigInt;
 
-#[derive(Debug)]
+#[derive(Debug, FromArgs)]
```
OddCoincidence: Built-in types like `PySlice` shouldn't be implementing `FromArgs`

skinny121: I just thought it was convenient to do so, but now, I realize that it's s bad idea as any functions that accept PySlice would try and use that FromArgs implementation.

vm/src/function.rs
```diff
@@ -46,6 +47,12 @@ impl From<(&Args, &KwArgs)> for PyFuncArgs {
     }
 }
 
+impl FromArgs for PyFuncArgs {
```
OddCoincidence: Not really a fan of this impl because it's not obvious that it will force a pointless allocation of a `PyFuncArgs`. And there's nothing stopping someone from trying to combine this with other extractors.

vm/src/obj/objslice.rs
```diff
@@ -1,18 +1,16 @@
...
+use num_bigint::BigInt;
 
 #[derive(Debug)]
 pub struct PySlice {
-    // TODO: should be private
-    pub start: Option<BigInt>,
-    pub stop: Option<BigInt>,
-    pub step: Option<BigInt>,
+    pub start: Option<PyObjectRef>,
+    pub stop: Option<PyObjectRef>,
```
OddCoincidence: Why is `stop` an `Option`? Seems like it's always `Some`?

vm/src/function.rs
skinny121: `Vec::new()` doesn't allocate, so replacing it with an empty `PyFuncArgs` effectively zeros's the struct. Yes there is nothing to stop people from misusing this, but this is also true for the existing extractors like `Args`, `KwArgs`, and `OptionalArg`.

OddCoincidence:
It looks like rustfmt is the only thing failing, after that is fixed we can merge this

vm/src/function.rs
OddCoincidence: Fair enough, still seems a bit weird to me to have `PyFuncArgs` implement `FromArgs`, but it's not a huge deal

windelbouwman:
@skinny121 @OddCoincidence could we merge this PR? It looks like all review comments were dealed with?

PR #770, opened 2019-03-31
sequence protocol

palaviv:
Same as #765 but for `tuple` and `list`.

### DIFF
--- a/vm/src/obj/objsequence.rs
+++ b/vm/src/obj/objsequence.rs
@@ -1,11 +1,11 @@
-use std::cell::RefCell;
+use std::cell::{Cell, RefCell};
 use std::marker::Sized;
 use std::ops::{Deref, DerefMut, Range};
 
 use num_bigint::BigInt;
 use num_traits::{One, Signed, ToPrimitive, Zero};
 
-use crate::pyobject::{IdProtocol, PyObject, PyObjectRef, PyResult, TypeProtocol};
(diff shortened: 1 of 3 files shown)


vm/src/obj/objsequence.rs
```diff
@@ -137,7 +139,154 @@ impl<T: Clone> PySliceableSequence for Vec<T> {
     }
 }
 
-pub fn get_item(
+pub trait SequenceProtocol
+where
+    Self: Sized,
+{
+    fn get_elements(&self) -> Vec<PyObjectRef>;
```
adrian17: we definitely don't want to copy the entire vec just to eg check its size.

palaviv: Agree. I can't do a `Deref` inside the trait. Do you have a suggestion for something nicer then returning the `Refcell`?

coolreader18: You could return an `&[PyObjectRef]`, or if that wouldn't work an `impl Deref<Target = [PyObjectRef]>`.

vm/src/obj/objsequence.rs
```diff
@@ -137,7 +139,154 @@ impl<T: Clone> PySliceableSequence for Vec<T> {
...
+    fn as_object(&self) -> &PyObjectRef;
+    fn into_object(self) -> PyObjectRef;
```
coolreader18: I think this could be split off into a separate trait, and we could have a generic `impl<T> ... for PyRef<T>`

palaviv:
Should be implemented with inner like #785.

PR #773, opened 2019-04-01
Redo much of whats_left_to_implement.py

coolreader18:
It turns out we had a few methods that CPython doesn't implement, so I removed those. Specifically:

```
bytearray.isascii
bytes.isascii
float.__set_format__
iter.__iter__
iter.__length_hint__
iter.__next__
iter.__setstate__
str.isascii
```

Some of those (the `iter` ones, namely) were because we were just checking if the type had the method, not necessarily if an instance did, so I replaced the types we were passing with instances (e.g. `''` instead of `str`).


tests/snippets/whats_left_to_implement.py
```diff
@@ -1,878 +1,874 @@
...
-    for method in item['methods']:
-        try:
-            if not hasattr(item['type'], method):
-                not_implemented.append((item['name'], method))
-        except NameError:
-            not_implemented.append((item['name'], method))
-
-for r in not_implemented:
-    print(r[0], ".", r[1])
+if not_implemented:
```
windelbouwman: No need for this extra if, just loop over the empty list.

tests/snippets/whats_left_to_implement.py
```diff
@@ -1,878 +1,874 @@
...
+        '__repr__',
+        '__rmul__',
+        '__setattr__',
+        '__sizeof__',
+        '__str__',
+        '__subclasshook__',
+        'count',
+        'index',
+    ]),
+    'object': (object(), [
```
windelbouwman: Do we require an instance? We could do with the class itself right? So, change `object()` to `object`.

coolreader18: For `iter` a couple of attributes were only available on instances, but it's probably fine to give classes for other types.

Issue #771, opened 2019-03-31
Add a `match_class!` macro


OddCoincidence:
## Summary

Add a `match_class!` macro for implementing behavior that varies based on the built-in type of an object.

## Detailed Explanation

Doing this currently with `downcast` is a bit awkward. Because it consumes the `Rc`,
you have two options:

1. Clone the object and then do a series of `if`s.

```rust
impl PyFloat {
    fn eq(self: PyRef<Self>, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
        if let Some(i @ PyInt { .. }) = other.clone().downcast() {
            Ok(PyInt::new(&self.value + i.as_bigint()).into_ref(vm))
        } else if let Some(f @ PyFloat { .. }) = other.downcast() {{
            let output = self.value + f.value.to_bigint().unwrap();
            Ok(PyFloat::new(output))
        } else {
            Ok(vm.ctx.not_implemented())
        }
    }
}
```

2. Use the returned `Err(obj)` of `downcast` to avoid the clone.

```rust
match other.downcast() {
    Ok(i @ PyInt { ..}) => {
        Ok(PyInt::new(&self.value + ).into_ref(vm))
    },
    Err(obj) => match obj.downcast() {
        Ok(f @ PyFloat { .. }) => {
            let output = self.value + f.value.to_bigint().unwrap();
            Ok(PyFloat::new(output))
        },
        _ => Ok(vm.ctx.not_implemented()),
    }
}
```

This avoids a pointless clone at the cost of adding undesirable rightward drift. What we really want is to be able to use a `match`-like syntax that expands to the efficient version above:

```rust
impl PyFloat {
    fn eq(self: PyRef<Self>, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
        match_object!(other,
            i @ PyInt { .. } => {
                Ok(PyInt::new(&self.value + i.as_bigint()).into_ref(vm))
            },
            f @ PyFloat { .. } => {
                let output = self.value + f.value.to_bigint().unwrap();
                Ok(PyFloat::new(output))
            },
            _ => Ok(vm.ctx.not_implemented()),
        )
    }
}
``` 

## Drawbacks, Rationale, and Alternatives

The main drawback is that it is another macro. It should be very straightforward to implement and to understand though.

## Unresolved Questions

Will allowing a full pattern in each match clause encourage making struct fields public, when they really should not be? An alternative would be to just allow the type name in each match arm.

[...]

PR #774, opened 2019-04-02
Add match_class! macro

OddCoincidence:
See #771 for more details and discussion.

### DIFF
--- a/vm/src/macros.rs
+++ b/vm/src/macros.rs
@@ -150,9 +150,8 @@ macro_rules! extend_class {
 
 /// Macro to match on the built-in class of a Python object.
 ///
-/// Like `match`, `match_object!` is an expression, and so must have a default
-/// arm that will be evaluated when the object was not an instance of any of
-/// the classes specified in other arms.
+/// Like `match`, `match_class!` must be exhaustive, so a default arm with
+/// the uncasted object is required.
 ///
 /// # Examples
 ///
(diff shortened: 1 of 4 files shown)


vm/src/macros.rs
```diff
@@ -147,3 +147,90 @@ macro_rules! extend_class {
...
+///     f @ PyFloat => f.to_f64().to_bigint().unwrap(),
+///     obj => panic!("non-numeric object {}", obj)
+/// );
+///
+/// assert!(int_value.is_zero());
+/// ```
+#[macro_export]
+macro_rules! match_class {
+    // The default arm.
+    ($obj:expr, _ => $default:expr) => {
```
coolreader18: Could probably add a `$(,)?` here to allow trailing commas.

vm/src/macros.rs
```diff
@@ -147,3 +147,90 @@ macro_rules! extend_class {
         )*
     }
 }
+
+/// Macro to match on the built-in class of a Python object.
+///
+/// Like `match`, `match_object!` is an expression, and so must have a default
+/// arm that will be evaluated when the object was not an instance of any of
+/// the classes specified in other arms.
```
coolreader18: Should be `match_class!`, right?

coolreader18: Also, is there any way to enable a missing default path? `match` is allowed as just a statement, where everything just leads to a `()` type. Unless this is to ~sorta emulate algebraic types and requiring that all cases are covered? In which case I'd recommend making the documentation clearer here.

OddCoincidence: Yeah it needs to be exhaustive like `match`, I'll improve the docs

PR #779, opened 2019-04-03
Improve trailing comma situation

windelbouwman:


parser/src/python.lalrpop
```diff
@@ -817,7 +795,7 @@ Atom: ast::Expression = {
         // List comprehension:
         e
     },
-    "(" <e:TestList?> <trailing_comma:","?> ")" => {
+    "(" <e:TestList2?> <trailing_comma:","?> ")" => {
```
adrian17: Why do we need this part in the first place?
`TestList` already parses `1, 2, 3` as a tuple, and wrapping it in `()` should be a no-op, like any other expression.

This seems to work for me (passes all tests, didn't look for edge cases):
```
    "(" ")" => ast::Expression::Tuple { elements: vec![] },
    "(" <e:TestList> ")" => { e },
```
And then (and after resolving the issue above) we can drop `TestList2`.

parser/src/python.lalrpop
```diff
@@ -102,7 +102,7 @@ ExpressionStatement: ast::LocatedStatement = {
           }
         }
     },
-    <loc:@L> <expr:TestOrStarExprList> <op:AugAssign> <e2:TestList> => {
+    <loc:@L> <expr:TestOrStarExprList> <op:AugAssign> <e2:TestList2> => {
```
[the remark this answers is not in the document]
windelbouwman: I added a testcase for this. Simply using TestList works.

parser/src/python.lalrpop
windelbouwman: I resolved this, and dropped TestList2. Thank you for the review!

parser/src/ast.rs
```diff
@@ -47,7 +47,7 @@ pub enum Statement {
     Break,
     Continue,
     Return {
-        value: Option<Vec<Expression>>,
+        value: Option<Box<Expression>>,
```
adrian17: couldn't this be simply an `Option<Expression>`?

windelbouwman: Maybe, it's mainly done due to memory size of the different enum variants that `Box` is present. Maybe `Option` also acts as a sort of Box like pointer to reduce the full size of `Return`.

adrian17: Well, other Statement variants already hold a non-boxed Expression, so that shouldn't matter here. IMO either all or none (unless later performance measurements prove otherwise).

PR #780, opened 2019-04-03
Support non-string keys in dictionaries.

cthulahoops:
Much to do. Statement of intent at this stage.

### DIFF
--- a/vm/src/obj/objdict.rs
+++ b/vm/src/obj/objdict.rs
@@ -1,7 +1,6 @@
 use std::cell::{Cell, RefCell};
 use std::collections::HashMap;
 use std::fmt;
-use std::ops::{Deref, DerefMut};
 
 use crate::function::{KwArgs, OptionalArg};
 use crate::pyobject::{
@@ -37,92 +36,14 @@ impl PyValue for PyDict {
     }
 }
 
-pub fn get_elements<'a>(obj: &'a PyObjectRef) -> impl Deref<Target = DictContentType> + 'a {
-    obj.payload::<PyDict>().unwrap().entries.borrow()
-}
-
-pub fn get_mut_elements<'a>(obj: &'a PyObjectRef) -> impl DerefMut<Target = DictContentType> + 'a {
-    obj.payload::<PyDict>().unwrap().entries.borrow_mut()
-}
-
-pub fn set_item(
-    dict: &PyObjectRef,
-    _vm: &VirtualMachine,
-    needle: &PyObjectRef,
-    value: &PyObjectRef,
-) {
-    // TODO: use vm to call eventual __hash__ and __eq__methods.
-    let mut elements = get_mut_elements(dict);
-    set_item_in_content(&mut elements, needle, value);
-}
-
-pub fn set_item_in_content(
-    elements: &mut DictContentType,
-    needle: &PyObjectRef,
-    value: &PyObjectRef,
-) {
-    // XXX: Currently, we only support String keys, so we have to unwrap the
-    // PyObject (and ensure it is a String).
-
-    // TODO: invoke __hash__ function here!
-    let needle_str = objstr::get_value(needle);
-    elements.insert(needle_str, (needle.clone(), value.clone()));
-}
-
 pub fn get_key_value_pairs(dict: &PyObjectRef) -> Vec<(PyObjectRef, PyObjectRef)> {
-    let dict_elements = get_elements(dict);
-    get_key_value_pairs_from_content(&dict_elements)
-}
-
-pub fn get_key_value_pairs_from_content(
-    dict_content: &DictContentType,
-) -> Vec<(PyObjectRef, PyObjectRef)> {
+    let dict_elements = dict.payload::<PyDict>().unwrap().entries.borrow();
     let mut pairs: Vec<(PyObjectRef, PyObjectRef)> = Vec::new();
-    for (_str_key, pair) in dict_content.iter() {
+    for (_str_key, pair) in dict_elements.iter() {
         let (key, obj) = pair;
         pairs.push((key.clone(), obj.clone()));
     }
     pairs
-}
-
-pub fn get_item(dict: &PyObjectRef, key: &PyObjectRef) -> Option<PyObjectRef> {
-    let needle_str = objstr::get_value(key);
-    get_key_str(dict, &needle_str)
-}
-
-// Special case for the case when requesting a str key from a dict:
-pub fn get_key_str(dict: &PyObjectRef, key: &str) -> Option<PyObjectRef> {
-    let elements = get_elements(dict);
-    content_get_key_str(&elements, key)
-}
-
-/// Retrieve a key from dict contents:
-pub fn content_get_key_str(elements: &DictContentType, key: &str) -> Option<PyObjectRef> {
-    // TODO: let hash: usize = key;
-    match elements.get(key) {
-        Some(v) => Some(v.1.clone()),
-        None => None,
-    }
-}
-
-pub fn contains_key_str(dict: &PyObjectRef, key: &str) -> bool {
-    let elements = get_elements(dict);
-    content_contains_key_str(&elements, key)
-}
-
-pub fn content_contains_key_str(elements: &DictContentType, key: &str) -> bool {
-    // TODO: let hash: usize = key;
-    elements.get(key).is_some()
-}
-
-/// Take a python dictionary and convert it to attributes.
-pub fn py_dict_to_attributes(dict: &PyObjectRef) -> PyAttributes {
-    let mut attrs = PyAttributes::new();
-    for (key, value) in get_key_value_pairs(dict) {
-        let key = objstr::get_value(&key);
-        attrs.insert(key, value);
-    }
-    attrs
 }
 
 // Python dict methods:
@@ -137,7 +58,7 @@ impl PyDictRef {
         if let OptionalArg::Present(dict_obj) = dict_obj {
             if objtype::isinstance(&dict_obj, &vm.ctx.dict_type()) {
                 for (needle, value) in get_key_value_pairs(&dict_obj) {
-                    set_item(dict.as_object(), vm, &needle, &value);
+                    dict.set_item(&vm.ctx, &objstr::get_value(&needle), value);
                 }
             } else {
                 let iter = objiter::get_iter(vm, &dict_obj)?;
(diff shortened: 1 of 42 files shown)


windelbouwman:
A small step for you, but a giant leap for the rustpython project!

cthulahoops:
> A small step for you, but a giant leap for the rustpython project!

Just wait until people start claiming that we faked our dictionary support.

vm/src/obj/objdict.rs
```diff
@@ -295,29 +215,43 @@ impl PyDictRef {
...
 
 impl DictProtocol for PyDictRef {
     fn contains_key(&self, k: &str) -> bool {
-        content_contains_key_str(&self.entries.borrow(), k)
+        self.entries.borrow().get(k).is_some()
     }
 
     fn get_item(&self, k: &str) -> Option<PyObjectRef> {
-        content_get_key_str(&self.entries.borrow(), k)
+        match self.entries.borrow().get(k) {
```
windelbouwman: For this situation it is ideal to use the map function of the Option type like this:
```rust
self.entries.borrow().get(k).map(|v| v.1.clone())
```

vm/src/obj/objdict.rs
```diff
@@ -259,9 +180,8 @@ impl PyDictRef {
         }
     }
 
-    fn setitem(self, needle: PyObjectRef, value: PyObjectRef, _vm: &VirtualMachine) {
-        let mut elements = self.entries.borrow_mut();
-        set_item_in_content(&mut elements, &needle, &value)
+    fn setitem(self, needle: PyObjectRef, value: PyObjectRef, vm: &VirtualMachine) {
```
windelbouwman: I suggest to rename one of the functions `setitem` and `set_item`. The names are too similar.

vm/src/obj/objdict.rs
```diff
@@ -295,29 +215,43 @@ impl PyDictRef {
...
 
     fn get_key_value_pairs(&self) -> Vec<(PyObjectRef, PyObjectRef)> {
         get_key_value_pairs(self.as_object())
     }
 
     // Item set/get:
-    fn set_item(&self, ctx: &PyContext, key: &str, v: PyObjectRef) {
-        let key = ctx.new_str(key.to_string());
-        set_item_in_content(&mut self.entries.borrow_mut(), &key, &v);
+    fn set_item(&self, ctx: &PyContext, key_str: &str, v: PyObjectRef) {
```
windelbouwman: Note that in order to set an item, we in future probably require a vm reference, not a context.

vm/src/pyobject.rs
```diff
@@ -948,44 +938,35 @@ pub trait DictProtocol {
...
-        } else {
-            panic!("TODO {:?}", self);
-        }
+    fn set_item<T: IntoPyObject>(
+        &self,
+        key: T,
+        value: PyObjectRef,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyObjectRef> {
+        vm.call_method(self, "__setitem__", vec![key.into_pyobject(vm)?, value])
```
windelbouwman: This is certainly heading in the right direction!

vm/src/pyobject.rs
```diff
@@ -948,44 +938,35 @@ pub trait DictProtocol {
...
-    }
-
-    fn get_item(&self, k: &str) -> Option<PyObjectRef> {
-        if let Some(dict) = self.payload::<PyDict>() {
-            objdict::content_get_key_str(&dict.entries.borrow(), k)
-        } else {
-            panic!("TODO {:?}", k)
-        }
-    }
+pub trait ItemProtocol {
```
windelbouwman: I think this should be the final shape of the `ItemProtocol`. It took 22 iterations, but I think this is the proper signature :).

vm/src/pyobject.rs
```diff
@@ -948,44 +938,35 @@ pub trait DictProtocol {
...
+    // // Move: doesn't really belong in this protocol.
+    // fn get_key_value_pairs(&self, vm: &VirtualMachine) -> PyResult<PyListIterator>;
+    fn get_item<T: IntoPyObject>(&self, key: T, vm: &VirtualMachine) -> PyResult<PyObjectRef>;
```
windelbouwman: Can we use `PyResult` instead of `PyResult<PyObjectRef>`? I believe `PyObjectRef` is the default template parameter.

vm/src/obj/objdict.rs
cthulahoops: Yeah, I'm hoping they'll have merged by the time I get to end of this refactor.

vm/src/obj/objdict.rs
cthulahoops: Will do.

vm/src/obj/objdict.rs
cthulahoops: That's the next step. (Now pushed.)

cthulahoops:
I seem to have failed at git somehow. :(

cthulahoops:
Better.

cthulahoops:
I've done a bunch of evil special-casing to allow unsafe string key insertions on dictionaries without a VirtualMachine available.

Alternatives: 
* unsafe construction of the vm with null-modules, and then we can have an almost finished vm when we're creating the modules.
* restructure the vm so it's okay for the core modules to be missing.
* Some sort of maybe a full dict, maybe only pyattributes magic for modules.

I'm going to come back to the failing wasm tests.

cthulahoops:
After all of that, the easiest thing might be to create the required modules empty, and then populate them once the vm is ready.

That would have been easier - depends on how horrible the unsafe insert magic is. (Do we want fast paths for string keys anyway?)

windelbouwman:
You could re-use the string hash function, and insert the key into the dict using that hash?

cthulahoops:
Yes, that's what this implementation does.

(It's unsafe because it won't interact properly with objects that over eq and hash, but as no such objects exist during initialisation it's absolutely fine.)

cthulahoops:
That was easier than expected!

This is roughly the right shape, just need to clean-up the history.

There's some also some clean up (making ItemProtocol more widely used, possibly killing DictProtocol altogether), but it might be worth getting this lot merged first.

Some operations are inefficient, but I see that as a later phase of work.

vm/src/obj/objdict.rs
```diff
@@ -1,21 +1,22 @@
...
+
 use super::objiter;
 use super::objlist::PyListIterator;
-use super::objstr::{self, PyStringRef};
+use super::objstr;
 use super::objtype;
 use crate::obj::objtype::PyClassRef;
 
-pub type DictContentType = HashMap<String, (PyObjectRef, PyObjectRef)>;
+pub type DictContentType = dictdatatype::Dict;
```
windelbouwman: Ow yeah!

vm/src/pyobject.rs
```diff
@@ -941,51 +913,40 @@ impl<T> TypeProtocol for PyRef<T> {
 }
 
 pub trait DictProtocol {
```
windelbouwman: I'm in favor of removing this trait instead of the item trait. In CPython its called `Item protocol`.

vm/src/stdlib/ast.rs
```diff
@@ -502,80 +507,82 @@ fn expression_to_ast(vm: &VirtualMachine, expression: &ast::Expression) -> PyObj
...
         ast::Expression::Attribute { value, name } => {
             let node = create_node(vm, "Attribute");
 
             let py_value = expression_to_ast(vm, value);
-            vm.ctx.set_attr(&node, "value", py_value);
+            vm.set_attr(&node, "value", py_value).unwrap();
 
             let py_attr = vm.ctx.new_str(name.to_string());
-            vm.ctx.set_attr(&node, "attr", py_attr);
+            vm.set_attr(&node, "attr", py_attr).unwrap();
```
windelbouwman: Instead of unwrap, we should you the `?` operator and return a result instead of a pythonRef.

vm/src/vm.rs
```diff
@@ -58,18 +58,22 @@ impl VirtualMachine {
         let ctx = PyContext::new();
 
         // Hard-core modules:
-        let builtins = builtins::make_module(&ctx);
```
windelbouwman: This is wrong, make_module actually add's builtins to the scope right? New module simply creates an empty module. This might be the source of the build errors.

windelbouwman: Ignore my comment, I now see the code below :).

vm/src/stdlib/ast.rs
cthulahoops: I can't because that function isn't PyResult.

The whole ast module needs a bit of a refactor. Every time I change something I spend most of my time there. I'll do that later though.

vm/src/pyobject.rs
cthulahoops: That's roughly the direction I'm going in. There's a couple of tricky points:

Do we implement ItemProtocol on PyDictRef and if so should it be the full magic method powered version or the direct implementation.

If the full version, how do we name the underlying implementation.

vm/src/stdlib/ast.rs
windelbouwman: Okay, good idea.

vm/src/pyobject.rs
windelbouwman: If you have `PyDictRef` there is not need to go via magic methods (performance win!). If you have `PyObjectRef` you go the magic method path. We can implement `ItemProtocol` twice. Once on `PyObjectRef` and once on `PtDictRef`.

cthulahoops:
> This looks good to me! I wanted this change for a long time. Thanks for implementing it! I suggest to fix the build error and merge this change, since it is a large change.

I think this is now ready to merge once tests pass.

vm/src/pyobject.rs
cthulahoops: The problem is that for subclasses of dict the behaviour is different. It's liable to lead to very subtle bugs in ```dict.into_object().get_item``` and ```dict.get_item``` do different things.

vm/src/obj/objdict.rs
```diff
@@ -37,92 +38,12 @@ impl PyValue for PyDict {
...
-
-/// Take a python dictionary and convert it to attributes.
-pub fn py_dict_to_attributes(dict: &PyObjectRef) -> PyAttributes {
-    let mut attrs = PyAttributes::new();
-    for (key, value) in get_key_value_pairs(dict) {
-        let key = objstr::get_value(&key);
-        attrs.insert(key, value);
-    }
-    attrs
+    dict.payload::<PyDict>()
```
adrian17: can't this take a `PyDictRef` too?
(may be a followup patch)

cthulahoops: The are code paths that support both. A lot of callers have ```PyObjectRef```. I should change the callers to cast to a PyDictRef, and then this is just a simple method on ```PyDictRef```.

I've added to my clean-up list.

Issue #778, opened 2019-04-03
super().__new__(...)  doesn't work


skinny121:
The following doesn't work, but should:

```python
class A(type):
    def __new__(cls):
        super().__new__(cls, "A", (), {})

A()
```
This complains that type_new is getting too many parameters. The extra parameter is a copy of the class of A, that is getting bound within super(), which is not meant to occur.
The parameters to `type.__new__` are `<class "A">, <class "A">, "A", (), {}`. 
Replacing `super()` with `type` does solve the problem.

Lynskylate:
I met same trouble as you when I try to import enum.py. I simplify the code.
```python
class EnumMeta(type):
    def __new__(metacls, cls, bases, classdict):
        enum_class = super().__new__(metacls, cls, bases, classdict)
        return enum_class

class Test(metaclass=EnumMeta):
    def __init__(self):
        pass
"""
>>> t = Test() 
# RustPython Output:
Traceback (most recent call last):
  File "/tmp/reproduce.py", line 17, in <module>
    class Test(metaclass=EnumMeta):
  File "/tmp/reproduce.py", line 14, in __new__
    enum_class = super().__new__(metacls, cls, bases, classdict)
TypeError: type() takes 1 or 3 arguments,
"""
```

windelbouwman:
I stumbled upon the implicit `__class__` variable. This page might be related: https://www.python.org/dev/peps/pep-3135/

youknowone:
Both of the reported code works now.

PR #782, opened 2019-04-04
Add abc module from CPython

skinny121:
This copies over the pure python abc module implementation from CPython, with slight modifications. This is the first step in the Python side of the standard library.

Changes required to abc Python code:
- Changed `super().__new__` with `type.__new__` to workaround #778
- Changed a fronzenset to a set, to work around fronzenset not currently being iterable

Changes made to RustPython to make abc work:
- Made object hashable by default
- Added callback parameter to weakref.ref, which currently doesn't do anything, except not complain if it is passed.

I also copied over the PSF license file, so that we are complying with the terms.


adrian17:
I kinda assumes we were to use https://github.com/beeware/ouroboros for starters (and just fill in core lower-level stdlib till it starts working)?

tests/snippets/test_abc.py
```diff
@@ -0,0 +1,39 @@
+import abc
```
windelbouwman: Did you copy this file from cpython? Please note that cpython has its own test suite for the standard library. We should probably pick that one.

skinny121: Yes I did copy it from cpython. Eventually we should use cpython test suite, but that requires unittest to work, which It won't at this stage.

skinny121:
@adrian17 I think it's useful to start to test core stdlib modules, to check where we aren't quite compatible yet, and ensure we maintain compatibility. We could do this by downloading/installing the stdlib during building, but this would mean we can't patch the stdlib which we probably need to.

coolreader18:
I think ideally we shouldn't need to make patches, but that's probably not feasible at this point in development. Maybe we could have it as an optional test on Travis?

adrian17:
Personally, I was hopeful that it would be possible to start trying to use ouroboros/CPython stdlib/CPython test suite, verbatim; obviously, 99% of code would fail on imports, but that would make a good point of reference to start filling in the blanks.
(and some of CPython test files could be run verbatim already, with just a basic `unittest` skeleton implementation)

tests/snippets/test_abc.py
windelbouwman: unittest is an excellent candidate to port first. This I tried some time back, but too many things were broken at that time.

windelbouwman:
My suggestion would be to make the unittest library work as first library. I'm fine with using ouroboros, using cpython or re-implementing all of the library. The main issue with ouroboros and cpython stdlib is the libraries which are half python, half c. Those are the weird ones. The pure python ones, we can simply copy paste.

PSF-LICENSE
```diff
@@ -0,0 +1,254 @@
+A. HISTORY OF THE SOFTWARE
```
windelbouwman: Maybe this license file can be placed in the Lib folder, together with a readme indicating what it applies to.

skinny121:
I agree unittest should be the first library that we target.

Issue #251, opened 2019-01-08
Implement remaining functions for 'list'


shinglyu:
I'm planning to work on this for the new year. :)

shinglyu:
Note to self: sort has a placeholder but it's not implemented

shinglyu:
Just finished `list.index()`, will add more test and refine the `ValueError` description before creating a PR.

shinglyu:
#254

makarchuk:
I've implemented `list.pop`


Python test for error case is a bit weird for now
```
except Exception as e:
    assert(str(e).startswith("IndexError"))
```

It's because `IndexError` is not in global namespace, which seems to be a reason for another issue, don't you think?

windelbouwman:
Okay, IndexError should be placed in the global namespace as well.

dralley:
It looks like, currently, as per the results of ```dir()```:

Python 3.7 ```list()``` has these methods that RustPython does not have or expose:

'__init_subclass__',
 '__reduce__',
 '__reduce_ex__',
 '__sizeof__',

And RustPython ```list()``` has these methods that CPython does not have or expose:

'__bool__',
 '__dict__',

Although, obviously Python 3.7 lists can be interpreted as bools so I'm not sure why it's missing, maybe they take a shortcut.

youknowone:
now the only unimplemented method is `list.__init_subclass__`, so this issue seems practically done.

PR #783, opened 2019-04-04
Implement list.__delitem__ 

Ryex:
a correct implementation of `list.__delitem__` that supports int, slice, and negative indexing

ref #251

after some discussion on gitter I decided to implement a `PySliceableSequenceMut` trait and use it as the base for this `__delitem__` method as part of this work. For now, it only has the del methods but the intention is to have set methods too to support correct `__setitem__` implementation

## TODO:
- [x] Test cases  /snippits


vm/src/obj/objsequence.rs
```diff
@@ -137,6 +137,107 @@ impl<T: Clone> PySliceableSequence for Vec<T> {
...
+    //fn set_slice<T: PySliceableSequence>(&self, range: Range<usize>, seq: T) -> Self;
+    //fn set_slice_reverse<T: PySliceableSequence>(&self, range: Range<usize>, seq: T) -> Self;
+    //fn set_stepped_slice<T: PySliceableSequence>(&self, range: Range<usize>, step: usize, seq: T) -> Self;
+    //fn set_stepped_slice_reverse<T: PySliceableSequence>(&self, range: Range<usize>, step: usize, seq: T) -> Self;
+
+    fn del_slice(&mut self, range: Range<usize>);
+    fn del_stepped_slice(&mut self, range: Range<usize>, step: usize);
+    fn del_index(&mut self, index: usize);
+
+    fn del_slice_items(&mut self, vm: &VirtualMachine, slice: &PyObjectRef) -> PyResult {
```
adrian17: if the only possible type for `slice` is a `PySlice`, let's convert the reference to `PySliceRef` in `del_item`, and pass it here. Can remove the `payload => panic!` branch then.

Ryex: a good point, done with the latest commit.

vm/src/obj/objsequence.rs
```diff
@@ -183,6 +284,47 @@ pub fn get_item(
     }
 }
 
+pub fn del_item<T: PySliceableSequenceMut>(
```
OddCoincidence: I see that this is modeled after `get_item` above - this is kind of the old way of doing things. Rather than checking here for each kind of sequence payload, I think it would be best for the list implementations of these to be defined directly in the `PyListRef` impl.

adrian17: to be fair, the implementation probably could be reused for `bytearray`?

vm/src/obj/objsequence.rs
```diff
@@ -137,6 +137,107 @@ impl<T: Clone> PySliceableSequence for Vec<T> {
...
+                match step.to_i32() {
+                    Some(1) => {
+                        self.del_slice(range);
+                        Ok(vm.get_none())
+                    }
+                    Some(num) => {
+                        self.del_stepped_slice(range, num as usize);
+                        Ok(vm.get_none())
+                    }
+                    None => {
```
cthulahoops: What is this None case handling?

vm/src/obj/objsequence.rs
OddCoincidence: I don't think this is complex enough to warrant the reuse, especially once it's refactored to use some of the newer conveniences. For example:

- We could create a `SequenceIndex` type that implements `TryFromObject` that is accepted by `list.__getitem__` and `list.__setitem__` (similar to [what I just did for `range`](https://github.com/RustPython/RustPython/pull/784/files#diff-bf99aa0e3eaf69e3d751384530066205R355), but with an i32 instead of a PyintRef).
- That impl would internally use `i32::try_from_object` which will raise "cannot fit" exception rather than doing it manually.
- The checking for the sequence type isn't necessary in the concrete classes. It's basically an extra step that's only necessary because of the shared implementation. To me, this screams ["wrong abstraction"](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction).

cthulahoops: It mostly checks the subscript type - the sequence is just used for the error message. (Is there a neat way to construct that error in the caller and avoid passing sequence through at all?)

OddCoincidence: @cthulahoops did you read my comment above? checking the subscript type seems like a strong candidate for a `TryFromObject` implementation.

cthulahoops: We cross-posted... isn't this pretty similar to the inner type discussed in #765 ? Or is it just the trait you'd like to avoid?

OddCoincidence: I'm not referring to the trait here (although yeah, tbh I'm not super fond of it and think an inner sequence type shared by `list` and `tuple` could be a good idea). I was just referring to `get_item` and this new `del_item` helper, because they seem a bit weird and unnecessary to me.

adrian17: After thinking about this... Maybe it will be better to do it in the simplest way possible (YAGNI and all, even though we know it's going to be also needed for `bytearray` and more), and worry about reuse after we actually have two users of this.

Ryex: Ok, so perhaps I jumped the gun implementing this in one night without waiting for some more input.
So, correct me if I'm wrong but the consensus is to ditch the trait and implement it all on PyListRef?

additionally,  per @OddCoincidence, suggestion, instead of using a `PyObjectRef` for the subscript: the implementation of a `SequenceIndex` enum using an `i32`  and a `PySliceRef` with a `TryFromArgs`  like the work done on #784 would make for far less explicit type checking and panic! branching. 

I like this as it seems much cleaner and more in the realm of 'simplest solution that works' that I ascribe to.

vm/src/obj/objsequence.rs
```diff
@@ -137,6 +137,107 @@ impl<T: Clone> PySliceableSequence for Vec<T> {
...
+        }
+           
+    }
+}
+
+impl<T: Clone> PySliceableSequenceMut for Vec<T> {
+    fn del_slice(&mut self, range: Range<usize>) {
+        self.drain(range);
+    }
+
```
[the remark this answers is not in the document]
Ryex: ah, you are correct I'll double check the code as I thought I was detecting this case and flipping the range around.

vm/src/obj/objsequence.rs
Ryex:
```rust
pub enum SequenceIndex {
    Int(i32),
    Slice(PySliceRef)
}

impl TryFromObject for SequenceIndex {
        match_class!(obj,
            i @ PyInt => i32::try_from_object(vm i),
            s @ PySlice => Ok(SequenceIndex::Slice(s)),
            obj => Err(vm.new_type_error(format!(
                "sequence indices be integers or slices, not {}",
                obj.class(),
            )))
        )
    }
}
```
this is what was ment by internaly using `i32::try_from_obj` correct?

vm/src/obj/objsequence.rs
```diff
@@ -183,6 +284,47 @@ pub fn get_item(
...
+                }
+            }
+            None => {
+                Err(vm.new_index_error("cannot fit 'int' into an index-sized integer".to_string()))
+            }
+        };
+    }
+    
+    if subscript.payload::<PySlice>().is_some() {
+        if sequence.payload::<PyList>().is_some() {
```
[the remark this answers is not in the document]
coolreader18: In addition, it'd be better to use `.payload_is()` instead of `.payload().is_some()`.

vm/src/obj/objsequence.rs
OddCoincidence: Yeah, that's the right idea! The first arm can be simplified with `?`, i.e.:

```i @ PyInt => Ok(SequenceIndex::Int(i32::try_from_object(vm, i.into_object())?),```

Also note the addition of `into_object`: `try_from_object` takes a `PyObjectRef` so the downcasted `i` needs to be casted back to object (kinda weird I know but it can be cleaned up later).

**Comment by cthulahoops** [comment]
Looks good!

(I wonder if some sort of ```iter_slice``` returning an iterator of indexes touched by a slice would avoid get and delete having to repeat all the slice object. But let's leave that as possible further work.)

File: vm/src/obj/objsequence.rs [hunk 6]
```diff
@@ -137,6 +137,24 @@ impl<T: Clone> PySliceableSequence for Vec<T> {
     }
 }
 
+pub enum SequenceIndex {
```
adrian17: could we reuse `RangeIndex`? We'll need a "int-or-slice" typy for most `__getitem__`-like functions.

Ryex: actually, by the time I completed the rework, I realized that, yes, it potentially could have just used `RangeIndex` I only need to convert to `i32` once It came time to check bounds.
```rust
fn delindex(self, index: i32, vm: &VirtualMachine) -> PyResult {
        if let Some(pos_index) = self.get_pos(index) {
//...
}
```
could be
```rust
fn delindex(self, index: PyInt, vm: &VirtualMachine) -> PyResult {
        if let Some(pos_index) = self.get_pos(i32::try_from_object(vm, index.into_object())?) {
//...
}
```

however, I felt it would be best to leave it as is until directed to merge it. it seemed a little strange to be using something called `RangeIndex`.

windelbouwman:
@Ryex thank you for contributing to this project! Nice job in adding deletions!

PR #785, opened 2019-04-04
refactor objset using inner

palaviv:
As suggested by @OddCoincidence in #765. I used `match_class` macro to avoid unwanted clones.

### DIFF
--- a/vm/src/obj/objset.rs
+++ b/vm/src/obj/objset.rs
@@ -375,29 +375,53 @@ impl PySetRef {
     }
 
     fn eq(self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        self.inner.borrow().eq(&get_inner(vm, &other)?, vm)
+        match_class!(other,
+        set @ PySet => self.inner.borrow().eq(&set.inner.borrow(), vm),
+        frozen @ PyFrozenSet => self.inner.borrow().eq(&frozen.inner, vm),
+        other =>  {return Err(vm.new_type_error(format!("{} is not a subtype of set or frozenset", other.class())));},
+        )
     }
 
     fn ge(self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        self.inner.borrow().ge(&get_inner(vm, &other)?, vm)
+        match_class!(other,
+        set @ PySet => self.inner.borrow().ge(&set.inner.borrow(), vm),
+        frozen @ PyFrozenSet => self.inner.borrow().ge(&frozen.inner, vm),
+        other =>  {return Err(vm.new_type_error(format!("{} is not a subtype of set or frozenset", other.class())));},
+        )
     }
 
     fn gt(self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        self.inner.borrow().gt(&get_inner(vm, &other)?, vm)
+        match_class!(other,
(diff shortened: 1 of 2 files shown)


vm/src/obj/objset.rs
```diff
@@ -56,536 +55,688 @@ impl PyValue for PyFrozenSet {
...
+        other =>  {return Err(vm.new_type_error(format!("{} is not a subtype of set or frozenset", other.class())));},
+        )
     }
 
-    for element in get_elements(&other).iter() {
-        let value = vm.call_method(&zelf, "__contains__", vec![element.1.clone()])?;
-        if !objbool::get_value(&value) {
-            elements.insert(element.0.clone(), element.1.clone());
-        }
+    fn union(self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
```
OddCoincidence: Just return the `PySet` directly here, the `IntoPyObject` implementation should take care of the rest

OddCoincidence: I think `other` should be a `PyIterable` since cpython accepts any iterable here:

```
>>> {1, 2, 3}.union([4])
{1, 2, 3, 4}
>>> {1, 2, 3}.union(4)
TypeError: 'int' object is not iterable
```

vm/src/obj/objset.rs
```diff
@@ -56,536 +55,688 @@ impl PyValue for PyFrozenSet {
...
 
-fn set_eq(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
-    set_compare_inner(
-        vm,
-        args,
-        &|zelf: usize, other: usize| -> bool { zelf != other },
-        false,
-    )
-}
+    fn clear(&mut self, vm: &VirtualMachine) -> PyResult {
```
OddCoincidence: This doesn't need to return anything, it doesn't raise any exceptions and returning `None` should be implicit.

vm/src/obj/objset.rs
```diff
@@ -56,536 +55,688 @@ impl PyValue for PyFrozenSet {
...
+        }
+    }
 
-    let hash_value = objint::get_value(&hash);
-    let mut hasher = DefaultHasher::new();
-    hash_value.hash(&mut hasher);
-    let key = hasher.finish();
-    f(vm, elements, key, item)
-}
+    fn contains(&self, needle: PyObjectRef, vm: &VirtualMachine) -> PyResult {
```
OddCoincidence: This should just return a `bool` directly

palaviv:
Thanks for the CR @OddCoincidence. I changed all the functions that should support iterable but that was a bigger change then I thought.

vm/src/obj/objset.rs
palaviv: This was changed.

palaviv: Done

vm/src/obj/objset.rs
palaviv: Done

vm/src/obj/objset.rs
palaviv: I am not sure how as this can fail.

OddCoincidence: Oh sorry missed the `?` - in this case it should return `PyResult<bool>`

palaviv: Done

PR #787, opened 2019-04-05
Add os.{read,write,remove,unlink}

palaviv:
The test should be changed to create the file in some temp folder once we add `tempfile`. In case anyone has a simpler idea that will be more safe please suggest.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1,3 +1,4 @@
+use std::fs;
 use std::fs::File;
 use std::fs::OpenOptions;
 use std::io::{ErrorKind, Read, Write};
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/os.rs
```diff
@@ -113,6 +113,25 @@ fn os_error(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     Err(vm.new_os_error(msg))
 }
 
+fn os_read(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
```
adrian17: These functions probably would be easy to write with new args style, something like:

```
fn os_read(fd: PyIntRef, n: PyIntRef, vm: &VirtualMachine) -> PyResult
```

palaviv: Done.

windelbouwman:
Looks good to me! Testing seems tricky, we can also skip testing if we are going to call file deletion.

windelbouwman:
Maybe I would suggest to not test file deletion and creation, since it depends heavily on the directory where tests are run. It's pretty tricky. Or we should define a seperate test set aside from the snippets which does this kind of functional tests. Maybe we could use BDD or robotframework to test those behaviors with files and network connections.

palaviv:
My plan is to add `mkdir` and `rmdir` and move the tests into a temp directory. I am not sure if we need a new framework but moving them aside and test them in a separate travis might be good.

adrian17:
> I am not sure if we need a new framework but moving them aside and test them in a separate travis might be good.

IMO with temporary directories (as in, system's temporary directory, not `mkdir ./tmp`) that would be good enough in normal tests.

PR #790, opened 2019-04-05
Fix decorator on functions with defaults

skinny121:
Previously, the defaults and decorator function was around the wrong way on the stack, so it tried to call the defaults as if they were the decorator.


windelbouwman:
Nice find!

PR #791, opened 2019-04-06
Dict protocols

cthulahoops:
Finally remove DictProtocol, and clean up a bunch of other dict things.

I've come to the conclusion that ```ItemProtocol``` on ```PyDictRef``` must do the same thing as the version on ```PyObjectRef```.

* It's way too surprising if ```dict.get_item``` and ```dict.as_object().get_item``` are different.
* I haven't found any cases where it's correct to bypass the overridden ```__getitem__```, so the only thing left is the performance boost where we know it's a base dictionary because we just created it. In that case it's clearer to call ```inner_getitem``` explicitly. (We should optimise attribute access first though.)

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -912,12 +912,6 @@ impl<T> TypeProtocol for PyRef<T> {
     }
 }
 
-pub trait DictProtocol {
-    fn get_item<T: IntoPyObject>(&self, key: T, vm: &VirtualMachine) -> Option<PyObjectRef>;
-    fn set_item<T: IntoPyObject>(&self, key: T, value: PyObjectRef, vm: &VirtualMachine);
-    fn del_item<T: IntoPyObject>(&self, key: T, vm: &VirtualMachine);
-}
-
 pub trait ItemProtocol {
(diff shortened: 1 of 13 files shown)


vm/src/pyobject.rs
```diff
@@ -927,6 +921,22 @@ pub trait ItemProtocol {
         vm: &VirtualMachine,
     ) -> PyResult;
     fn del_item<T: IntoPyObject>(&self, key: T, vm: &VirtualMachine) -> PyResult;
+    fn get_item_option<T: IntoPyObject>(
```
windelbouwman: Not a great fan of this method name. But fine for now.

vm/src/pyobject.rs
```diff
@@ -912,12 +912,6 @@ impl<T> TypeProtocol for PyRef<T> {
     }
 }
 
-pub trait DictProtocol {
```
windelbouwman: Glad about its removal!

windelbouwman:
Looks good to me!

vm/src/pyobject.rs
cthulahoops: Me neither. Any better suggestions?

coolreader18: `maybe_get_option`?

PR #793, opened 2019-04-06
WIP: Drop unused IntoPyNativeFunc impl

adrian17:
Should fix latest nightly breakages.

### DIFF
--- a/vm/src/function.rs
+++ b/vm/src/function.rs
@@ -434,12 +434,6 @@ where
     }
 }
 
-impl IntoPyNativeFunc<PyFuncArgs, PyResult> for PyNativeFunc {
-    fn into_func(self) -> PyNativeFunc {
-        self
-    }
-}
-
 // This is the "magic" that allows rust functions of varying signatures to
 // generate native python functions.
 //


adrian17:
...I didn't except this to fail for wasm. Let's go back to the drawing board and/or wait for Rust folks to give some advice.

coolreader18:
Maybe to fix WASM, you could `impl FromArgs for PyFuncArgs` and then make `print_fn` a `Fn(PyFuncArgs, &VirtualMachine)`.

windelbouwman:
@adrian17 what do you wish to do with this PR?

adrian17:
For now, let's just wait. It's the nature of Nightly that things can break periodically, and we shouldn't really try fixing our code as soon as this happens. Let's just give the situation some time to clarify itself, especially now that the Rustc bug just got some activity.
I'd also consider moving Nightly-dependent builds to "allowed failures" on Travis.

adrian17:
PS oh wait, now it's breaking on Beta too? Huh.
Anyway, I'd still wait a little more. Will take a look at it tomorrow, if the issue persists.

windelbouwman:
If the issue persists, we should disable the nightly build, to prevent that we miss other regressions.

PR #794, opened 2019-04-06
Parameterise dictdatatype::Dict by value type.

cthulahoops:
This is in preparation for using this to back set too.

### DIFF
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -9,20 +9,29 @@ use num_traits::ToPrimitive;
 /// And: http://code.activestate.com/recipes/578375/
 use std::collections::HashMap;
 
-#[derive(Default)]
-pub struct Dict {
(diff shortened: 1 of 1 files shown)


vm/src/dictdatatype.rs
```diff
@@ -9,20 +9,29 @@ use num_traits::ToPrimitive;
 /// And: http://code.activestate.com/recipes/578375/
 use std::collections::HashMap;
 
-#[derive(Default)]
-pub struct Dict {
+pub struct Dict<T = PyObjectRef> {
```
windelbouwman: When would you require a different value type? What's the reason to use type parameters?

cthulahoops: The idea is that the backing type for set will be:

```
type SetContentType = Dict<()>
```

windelbouwman: Ah, okay, I get it.

PR #795, opened 2019-04-06
[WASM] Clean up a bit

coolreader18:
Among other things, I removed `AccessibleVM` and just used a `Weak<StoredVirtualMachine>` because the hack around mutability is no longer required, thanks to immutable `VirtualMachine` with #706.


coolreader18:
Could you review this @windelbouwman? It's a lot of converting stuff to new style args and classes, and the Weak change is not that large.

windelbouwman:
Okay, I do not understand fully, but accessible VM removed is a good thing.

rmliddle:
Apologies on being slow to get to this! 
Not sure I could have contributed much here either, this has changed so much since I last worked on it

PR #796, opened 2019-04-06
Dict iters

cthulahoops:


cthulahoops:
It takes 6 additional types to iterate dictionaries!

tests/snippets/dict.py
```diff
@@ -37,11 +37,18 @@ def dict_eq(d1, d2):
         res.add(key)
 assert res == set(['a','b'])
 
+x = {'a': 1, 'b': 2, 'c': 3, 'd': 3}
+del x['c']
+it = iter(x.items())
```
adrian17: could we get more tests of these? Including the "view"-ness? Something like

```
data = {1: 2, 3: 4}
items = data.items()
assert list(items) == [(1, 2), (3, 4)]
data[5] = 6
assert list(items) == [(1, 2), (3, 4), (5, 6)]
```

vm/src/obj/objdict.rs
```diff
@@ -300,9 +361,16 @@ pub fn init(context: &PyContext) {
         "values" => context.new_rustfunc(PyDictRef::values),
         "items" => context.new_rustfunc(PyDictRef::items),
         // TODO: separate type. `keys` should be a live view over the collection, not an iterator.
```
windelbouwman: This TODO could now be removed right?

vm/src/obj/objdict.rs
```diff
@@ -131,56 +130,20 @@ impl PyDictRef {
...
-            .iter()
-            .map(|(k, v)| vm.ctx.new_tuple(vec![k.clone(), v.clone()]))
-            .collect();
-        let items_list = vm.ctx.new_list(items);
-
-        PyListIterator {
-            position: Cell::new(0),
-            list: items_list.downcast().unwrap(),
-        }
+    fn values(self, _vm: &VirtualMachine) -> PyDictValues {
```
windelbouwman: This looks very clean!

vm/src/obj/objdict.rs
```diff
@@ -284,6 +247,104 @@ impl ItemProtocol for PyDictRef {
     }
 }
 
+macro_rules! dict_iterator {
```
windelbouwman: Not sure whether I like this amount of macro-fu, but still, it solves the duplication. Probably this thing has too many macro arguments. Another idea, create a struct which takes a trait which handles the specifics?

tests/snippets/dict.py
windelbouwman: I think this is a good idea to add.

vm/src/obj/objdict.rs
cthulahoops: I'm not sure I see how to get much of the benefit here with traits, especially while avoiding wrong abstraction. The macro feels like the most straight-forward approach.

PR #797, opened 2019-04-06
Add os.{mkdir, mkdirs, rmdir, listdir}

palaviv:
Currently I did not added tests. I want to move all the files tests to a system temp directory. I will need to add the following in order to do that:

- Add `os.getenv` to get windows `%TEMP%`
- Add `os.sep` for correct path building in windows and linux.

I will these two features and then add to `testutils` a context manager of temp folder and move the tests to use that folder.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -169,6 +169,15 @@ fn os_mkdirs(path: PyStringRef, vm: &VirtualMachine) -> PyResult {
     Ok(vm.get_none())
 }
 
+fn os_rmdir(path: PyStringRef, vm: &VirtualMachine) -> PyResult {
+    match fs::remove_dir(&path.value) {
+        Ok(_) => (),
+        Err(s) => return Err(vm.new_os_error(s.to_string())),
+    }
+
+    Ok(vm.get_none())
+}
+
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/os.rs
```diff
@@ -151,6 +151,48 @@ fn os_remove(path: PyStringRef, vm: &VirtualMachine) -> PyResult {
     Ok(vm.get_none())
 }
 
+fn os_mkdir(path: PyStringRef, vm: &VirtualMachine) -> PyResult {
+    match fs::create_dir(&path.value) {
+        Ok(_) => (),
+        Err(s) => return Err(vm.new_os_error(s.to_string())),
+    }
+
+    Ok(vm.get_none())
```
OddCoincidence: With the new args style, we don't need to explicitly return `None` anymore. If you change the return type to `PyResult<()>` you can return `Ok(())`. Also I'd just do this directly in the `match` above so that the early return is not necessary.

coolreader18: You could also just use `map_err` instead of the `match`.

palaviv: Thanks @OddCoincidence and @coolreader18. It is way simpler this way :)

PR #799, opened 2019-04-06
Allow first argument to be a shared ref to a python value

OddCoincidence:
This allows the first argument to a function to be a shared reference to a type implementing `PyValue` _as a special case_. This is useful because it allows calling python methods from rust without consuming or cloning the reference. Eventually, it would be ideal to support shared references in _any_ position, but that is significantly more complex to accomplish due to a combinatorial explosion of trait impls required.

### DIFF
--- a/vm/src/function.rs
+++ b/vm/src/function.rs
@@ -2,7 +2,9 @@ use std::collections::HashMap;
 use std::ops::RangeInclusive;
 
 use crate::obj::objtype::{isinstance, PyClassRef};
-use crate::pyobject::{IntoPyObject, PyObjectRef, PyResult, TryFromObject, TypeProtocol};
+use crate::pyobject::{
+    IntoPyObject, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject, TypeProtocol,
+};
 use crate::vm::VirtualMachine;
 
 use self::OptionalArg::*;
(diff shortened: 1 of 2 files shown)


cthulahoops:
I note that you move all the methods onto PyFloat from PyFloatRef. It doesn't seem to be possible to use the ```&self``` for methods on ```PyFloatRef````? (I guess that mostly isn't necessary?)

OddCoincidence:
> I note that you move all the methods onto PyFloat from PyFloatRef. It doesn't seem to be possible to use the &self for methods on ```PyFloatRef````? (I guess that mostly isn't necessary?)

Yeah it shouldn't be necessary because `PyRef<T>` derefs to `T`, and if we actually do need a full `PyRef`, we should just take it by value. 

I originally started the `impl Py*Ref` pattern so we could use `self` instead of `zelf`, but it causes some problems, mainly that it's not legal for crates outside of `rustpython_vm` to create such an impl block. For the rare case we need a full cloneable reference, we can just use `zelf: PyRef<Self>` until `arbitrary_self_types` is stabilized in rust, at which point we can use `self: PyRef<Self>` there.

OddCoincidence:
> I'm not sure I understand the solution, but very glad you solved it!

I'll have to add more documentation, but in general this is basically working around rust's trait coherence rules, because we can't implement `IntoPyNativeFunc` for both owned and reference types without them overlapping.

cthulahoops:
> I'll have to add more documentation,

Don't worry too much about documentation. This is I'd have to work with the implementations to really get them thing. I understand well enough to use them. (I hope.)

cthulahoops:
> Yeah it shouldn't be necessary because `PyRef<T>` derefs to `T`, and if we actually do need a full `PyRef`, we should just take it by value.

That leaves us back where we started with needing to do ```obj.clone().foo``` if the method needs the PyRef. I guess there will be few enough of those cases that it doesn't matter.

PR #801, opened 2019-04-07
Implement copy and update dict methods

skinny121:


vm/src/obj/objdict.rs
```diff
@@ -44,13 +45,28 @@ impl PyDictRef {
         kwargs: KwArgs,
         vm: &VirtualMachine,
     ) -> PyResult<PyDictRef> {
-        let mut dict = DictContentType::default();
+        let dict = DictContentType::default();
 
+        let entries = RefCell::new(dict);
+        // it's unfortunate that we can't abstract over RefCall, as we should be able to use dict
+        // directly here, but that would require generic associated types
```
cthulahoops: I'm not sure what you mean by this comment?

The natural thing for me for is to have merge take ```&mut DictContentType``` which is simpler and works fine. The short-lived borrows in your version are probably better though at avoiding annoying borrow errors.

tests/snippets/dict.py
```diff
@@ -107,3 +107,19 @@ def __eq__(self, other):
 x['a'] = 2
 x['b'] = 10
 assert list(x) == ['a', 'b']
+
+y = x.copy()
+x['c'] = 12
+assert list(y) == ['a', 'b']
```
cthulahoops: Can we use ```dict_eq``` here? We should implement ```dict.__eq__``` to simplify this.

vm/src/obj/objdict.rs
```diff
@@ -44,13 +45,28 @@ impl PyDictRef {
...
+        dict_obj: OptionalArg<PyObjectRef>,
+        kwargs: KwArgs,
+        vm: &VirtualMachine,
+    ) -> PyResult<()> {
         if let OptionalArg::Present(dict_obj) = dict_obj {
             let dicted: PyResult<PyDictRef> = dict_obj.clone().downcast();
             if let Ok(dict_obj) = dicted {
-                for (key, value) in dict_obj.get_key_value_pairs() {
-                    dict.insert(vm, &key, value)?;
+                let mut dict_borrowed = dict.borrow_mut();
```
cthulahoops: If this borrow moved inside the loop then would we avoid the borrow error and special case in update below?

windelbouwman:
Looks good to me!

cthulahoops:
Fast merged! I was hoping to discuss my review before merging.

vm/src/obj/objdict.rs
skinny121: As you hinted at, the problem with merge taking `&mut DictContentType` is that for the iterable case, arbitrary python code is executed. We can't borrow the `RefCell` over any invocation of arbitrary python code as it may reference the dict we are updating, leading to a panic.

What I originally wanted to do is define merge as:
```rust
fn merge<B: BorrowMut<DictContentType>>(
        dict: B,
        dict_obj: OptionalArg<PyObjectRef>,
        kwargs: KwArgs,
        vm: &VirtualMachine,
    ) -> PyResult<()> {
    ...
}
```
So it could accept both `&mut DictContentType` and `&RefCell`, but I discovered that `RefCell` doesn't actually implement `BorrowMut`, but has it's own `borrow_mut` method.

vm/src/obj/objdict.rs
skinny121: Not exactly, the dict_obj is still borrowed throughout the loop as the `Ref` isn't dropped and for `iter_items` it can't be dropped before the loop, as it borrows new items on each iteration for the dict. If I reverted back to using `get_key_value_pairs`, that special case would be avoided though.

vm/src/obj/objdict.rs
cthulahoops: Ah, I see. I think I prefer the non-generic version in any case. There aren't be enough callers to justify the extra complexity.

vm/src/obj/objdict.rs
cthulahoops: Oh yes, of course, the other borrow.

Once I get my dict iterator patch merged we might be able to use that to iterate the dictionary without keeping the borrow - but it's probably not worth it just to avoid the special case.

PR #802, opened 2019-04-07
use a code generator for not implemented tests

andrew-ld:
Instead of having a static snippet that requires to be manually edited I thought to create a generator by taking all the methods from cpython


tests/not_impl_gen.py
```diff
@@ -0,0 +1,33 @@
...
+    list,
+    memoryview,
+    range,
+    set,
+    str,
+    tuple,
+    object
+]
+
+header = open("generator/not_impl_header.txt")
```
windelbouwman: Maybe a nice idea is to use ninja2 as template engine. This prevents seperate header, footer and body files. This introduces an extra dependency, but I think it is okay.

tests/snippets/whats_left_to_implement.py
```diff
@@ -1,873 +0,0 @@
-import platform
```
windelbouwman: I'd rather keep the generated file checked in into GIT. Please add a note that the file is generated automatically by the generator script, and check it into GIT so that people can run tests straight away. I think it is good to keep the generated file, as it gives us a common reference. If this file is generated on the fly, it depends upon installed cpython what will be generated, so results vary between PC's.

windelbouwman: On the other hand, maybe it is also fine to leave it out of GIT.

andrew-ld: would it be better to insert a warning in the generator or in the generated file?

windelbouwman: In the generated file, a warning should be put that the file was automatically generated, and that manual editing is not handy.

windelbouwman:
@andrew-ld thank you for contributing to this project!

PR #804, opened 2019-04-07
refactor bytes

jgirardet:
It's the start to refactor bytes and  bytearray like #765 .

- only bytes for now (no bytearray) with new style method
- What to do with byteinner ? change the name ? remove a differet file and put everything in bytes.rs
- tests added
- I struggled with the iterator, I don't know if it can be send to byteinner.rs or not.
- the `__new__` method has been rewritten to deal with more cases, but it's quite horrible to read. interested in ideas to refactor this one. there are many possibilities with various error message so ...
- `b"éùà"` passes now but it shouldn't. I'm not sur where to fix it.

### DIFF
--- a/vm/src/obj/objbyteinner.rs
+++ b/vm/src/obj/objbyteinner.rs
@@ -1,4 +1,3 @@
-use crate::obj::objtype::PyClassRef;
 use crate::pyobject::PyObjectRef;
 
 use crate::function::OptionalArg;
@@ -93,4 +92,33 @@ impl PyByteInner {
             }
         }
     }
+
+    pub fn repr(&self) -> PyResult<String> {
+        let mut res = String::with_capacity(self.elements.len());
+        for i in self.elements.iter() {
+            match i {
+                0..=8 => res.push_str(&format!("\\x0{}", i)),
+                9 => res.push_str("\\t"),
+                10 => res.push_str("\\n"),
+                13 => res.push_str("\\r"),
+                32..=126 => res.push(*(i) as char),
+                _ => res.push_str(&format!("\\x{:x}", i)),
+            }
+        }
+        Ok(res)
+    }
+
+    pub fn len(&self) -> usize {
+        self.elements.len()
+    }
+
+    pub fn eq(&self, other: &PyByteInner, vm: &VirtualMachine) -> PyResult {
+        if self.elements == other.elements {
+            Ok(vm.new_bool(true))
+        } else {
+            Ok(vm.new_bool(false))
+        }
+    }
 }
+// TODO
+// fix b"é" not allowed should be bytes("é", "utf8")
--- a/vm/src/obj/objbyteinner.rs
+++ b/vm/src/obj/objbyteinner.rs
@@ -314,6 +314,39 @@ impl PyByteInner {
             .collect::<String>();
         Ok(vm.ctx.new_str(bla))
     }
+
+    pub fn fromhex(string: String, vm: &VirtualMachine) -> Result<Vec<u8>, PyObjectRef> {
+        // first check for invalid character
+        for (i, c) in string.char_indices() {
+            if !c.is_digit(16) && !c.is_whitespace() {
+                return Err(vm.new_value_error(format!(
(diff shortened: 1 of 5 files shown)


vm/src/obj/objbyteinner.rs
```diff
@@ -0,0 +1,191 @@
...
+                            enc.class().name
+                        )));
+                    }
+                } else {
+                    return Err(vm.new_type_error("encoding without a string argument".to_string()));
+                }
+            } else {
+                return Err(vm.new_type_error("encoding without a string argument".to_string()));
+            }
+        // On ly one argument
```
palaviv: typo: `Only`

vm/src/obj/objbyteinner.rs
```diff
@@ -0,0 +1,191 @@
...
+    }
+
+    pub fn add(&self, other: &PyByteInner, vm: &VirtualMachine) -> PyResult {
+        let elements: Vec<u8> = self
+            .elements
+            .iter()
+            .chain(other.elements.iter())
+            .cloned()
+            .collect();
+        Ok(vm.ctx.new_bytes(elements))
```
palaviv: This should probably return a new `PyByteInner` and not `bytes`.

jgirardet: I finally did it with Vec[u8] instead , easier to call wm.ctx.new_bytes()

jgirardet: But thinking to this, maybe it's more consistent to always return a PyByteInner when it's possible. Anyone has an opinion on it ?

coolreader18:
The tests where you're `except`ing and then `assert`ing true could be changed to use `assertRaises`. Other than that, LGTM

vm/src/obj/objbyteinner.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    pub fn contains_int(&self, int: &PyInt, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
+        if let Some(int) = int.as_bigint().to_u8() {
+            if self.elements.contains(&int) {
+                Ok(vm.new_bool(true))
+            } else {
+                Ok(vm.new_bool(false))
+            }
+        } else {
+            Err(vm.new_value_error("byte mu st be in range(0, 256)".to_string()))
```
adrian17: typo

vm/src/obj/objbyteinner.rs
```diff
@@ -0,0 +1,349 @@
...
+            .chunks(2)
+            .map(|x| x.to_vec().iter().collect::<String>())
+            .map(|x| u8::from_str_radix(&x, 16))
+            .map(|x| x.unwrap())
+            .collect::<Vec<u8>>())
+    }
+}
+
+// TODO
+// fix b"é" not allowed should be bytes("é", "utf8")
```
adrian17: it's not allowed anymore, right? Is this TODO still relevant?

vm/src/obj/objbyteinner.rs
```diff
@@ -0,0 +1,349 @@
...
+        // First handle bytes(string, encoding[, errors])
+        if let OptionalArg::Present(enc) = enc_option {
+            if let OptionalArg::Present(eval) = val_option {
+                if let Ok(input) = eval.downcast::<PyString>() {
+                    if let Ok(encoding) = enc.clone().downcast::<PyString>() {
+                        if encoding.value.to_lowercase() == "utf8".to_string()
+                            || encoding.value.to_lowercase() == "utf-8".to_string()
+                        // TODO: different encoding
+                        {
+                            return Ok(PyByteInner {
```
adrian17: Could you run clippy on this? These seem like redundant `return`s.

vm/src/obj/objbyteinner.rs
jgirardet: no it's ok now

vm/src/obj/objbyteinner.rs
jgirardet: ok

jgirardet:
ok changes done

PR #806, opened 2019-04-08
Initial version of symbol table builder.

windelbouwman:


vm/src/symboltable.rs
```diff
@@ -0,0 +1,272 @@
...
+
+pub enum SymbolRole {
+    Global,
+    Nonlocal,
+    Used,
+    Assigned,
+}
+
+pub struct SymbolTable {
+    // TODO: split-up into nested scopes.
```
cthulahoops: Aha, I was wondering why there was only a single symbol table.

vm/src/symboltable.rs
```diff
@@ -0,0 +1,272 @@
...
+            ast::Statement::FunctionDef {
+                name,
+                body,
+                args,
+                decorator_list,
+                returns,
+            } => {
+                self.scan_expressions(decorator_list);
+                self.register_name(name, SymbolRole::Assigned);
+                for parameter in &args.args {}
```
cthulahoops: Shouldn't these names be marked assigned?

vm/src/symboltable.rs
windelbouwman: There is a nested tree of symbol tables. Each symbol table can have multiple sub scopes, which is done now.

windelbouwman:
This is the first part on the way to support global and nonlocal things. Each variable is scanned for usage, and it is checked if the variable is local or global. This can then help to emit different bytecode depending on variable scopeage.

windelbouwman:
@cthulahoops I added some bytecode, it's now ready for review, or do you think it is fine?

vm/src/bytecode.rs
```diff
@@ -50,9 +56,11 @@ pub enum Instruction {
     },
     LoadName {
         name: String,
+        scope: NameScope,
     },
     StoreName {
         name: String,
+        scope: NameScope,
     },
     DeleteName {
```
cthulahoops: ```DeleteName``` needs a scope too?

vm/src/frame.rs
```diff
@@ -995,19 +1013,27 @@ impl Frame {
...
-            }
-        }
+    fn load_name(
+        &self,
+        vm: &VirtualMachine,
+        name: &str,
+        scope: &bytecode::NameScope,
+    ) -> FrameResult {
+        let value = match scope {
+            bytecode::NameScope::Global => self.scope.globals.get_item(name, vm)?,
```
cthulahoops: Don't bypass the interface of Scope here and access globals directly. Scope should either expose a ```load_global```, or ```Scope::load_name``` should accept a ```bytecode::Scope``` as a parameter. (In which case ```Scope::load_cell```should merge into ```Scope::load_name```.)

vm/src/frame.rs
```diff
@@ -995,19 +1013,27 @@ impl Frame {
...
+            bytecode::NameScope::Local => match self.scope.load_name(&vm, name) {
+                Some(value) => value,
+                None => {
+                    let name_error_type = vm.ctx.exceptions.name_error.clone();
```
cthulahoops: The match/exception creation here should be done outside the ```match scope```. As it is I think load_global will raise ```KeyError```rather than ```NameError```.

vm/src/frame.rs
```diff
@@ -995,19 +1013,27 @@ impl Frame {
...
-                let msg = format!("name '{}' is not defined", name);
-                let name_error = vm.new_exception(name_error_type, msg);
-                Err(name_error)
-            }
-        }
+    fn load_name(
+        &self,
+        vm: &VirtualMachine,
+        name: &str,
+        scope: &bytecode::NameScope,
```
cthulahoops: Can we call this ```name_scope``` to avoid confusion with ```frame::Scope```?

vm/src/symboltable.rs
windelbouwman: I guess so, yes.

vm/src/bytecode.rs
```diff
@@ -38,6 +38,12 @@ bitflags! {
 
 pub type Label = usize;
 
+#[derive(Debug, Clone, PartialEq)]
+pub enum NameScope {
+    Local,
+    Global,
```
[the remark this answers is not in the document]
windelbouwman: Yes, I am planning on extending this later.

vm/src/bytecode.rs
```diff
@@ -50,9 +56,11 @@ pub enum Instruction {
     },
     LoadName {
         name: String,
+        scope: NameScope,
```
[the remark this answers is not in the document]
windelbouwman: Yes, I faced the choice of adding extra opcodes, or adding an operand to the existing name opcodes. I chose the operand option for clarity. I agree that at some point, we can optimize. For example, our label system is not optimal as well. I suggest to optimize later.

vm/src/bytecode.rs
windelbouwman: Yes, it does. Do you want me to implement this now?

vm/src/frame.rs
windelbouwman: Okay, I'll go for `load_global` in that case. Should `globals` be private?

vm/src/frame.rs
windelbouwman: Agreed, moved it out of the match.

vm/src/frame.rs
windelbouwman: Good idea, will do.

vm/src/bytecode.rs
cthulahoops: Now or later is fine.

vm/src/frame.rs
cthulahoops: It's needed by the ```globals``` builtin too. We could provide a getter method for that if we want.

PR #810, opened 2019-04-09
Implement IntoIterator for PyDictRef.

cthulahoops:
Simplify iteration of dictionaries from rust code.

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -393,9 +393,8 @@ impl Frame {
                         // Take all key-value pairs from the dict:
                         let dict: PyDictRef =
                             obj.downcast().expect("Need a dictionary to build a map.");
-                        let dict_elements = dict.get_key_value_pairs();
-                        for (key, value) in dict_elements.iter() {
-                            map_obj.set_item(key.clone(), value.clone(), vm).unwrap();
+                        for (key, value) in dict {
+                            map_obj.set_item(key, value, vm).unwrap();
                         }
                     }
                 } else {
@@ -636,8 +635,7 @@ impl Frame {
                         let kwargs = if *has_kwargs {
                             let kw_dict: PyDictRef =
                                 self.pop_value().downcast().expect("Kwargs must be a dict.");
-                            let dict_elements = kw_dict.get_key_value_pairs();
-                            dict_elements
+                            kw_dict
                                 .into_iter()
                                 .map(|elem| (objstr::get_value(&elem.0), elem.1))
                                 .collect()
@@ -862,8 +860,8 @@ impl Frame {
 
         // Grab all the names from the module and put them in the context
         if let Some(dict) = &module.dict {
-            for (k, v) in dict.get_key_value_pairs().iter() {
-                self.scope.store_name(&vm, &objstr::get_value(k), v.clone());
+            for (k, v) in dict {
+                self.scope.store_name(&vm, &objstr::get_value(&k), v);
             }
         }
         Ok(None)
@@ -1230,8 +1228,7 @@ impl fmt::Debug for Frame {
             .collect::<String>();
         let dict = self.scope.get_locals();
         let local_str = dict
-            .get_key_value_pairs()
-            .iter()
+            .into_iter()
             .map(|elem| format!("\n  {:?} = {:?}", elem.0, elem.1))
             .collect::<String>();
         write!(
--- a/vm/src/obj/objdict.rs
+++ b/vm/src/obj/objdict.rs
@@ -237,34 +237,34 @@ impl ItemProtocol for PyDictRef {
 // Implement IntoIterator so that we can easily iterate dictionaries from rust code.
 impl IntoIterator for PyDictRef {
     type Item = (PyObjectRef, PyObjectRef);
-    type IntoIter = DictIterator;
+    type IntoIter = DictIter;
 
     fn into_iter(self) -> Self::IntoIter {
-        DictIterator::new(self)
+        DictIter::new(self)
     }
 }
 
 impl IntoIterator for &PyDictRef {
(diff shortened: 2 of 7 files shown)


vm/src/obj/objdict.rs
```diff
@@ -247,6 +234,54 @@ impl ItemProtocol for PyDictRef {
     }
 }
 
+// Implement IntoIterator so that we can easily iterate dictionaries from rust code.
+impl IntoIterator for PyDictRef {
```
OddCoincidence: Rather than implementing this for the specific type, could this be implemented for `&PyDict`, and then a blanket impl for `PyRef<T>` (i.e. something like `impl<T> IntoIterator for PyRef<T> where &T: IntoIterator`).

vm/src/obj/objdict.rs
```diff
@@ -247,6 +234,54 @@ impl ItemProtocol for PyDictRef {
...
+impl IntoIterator for &PyDictRef {
+    type Item = (PyObjectRef, PyObjectRef);
+    type IntoIter = DictIterator;
+
+    fn into_iter(self) -> Self::IntoIter {
+        DictIterator::new(self.clone())
+    }
+}
+
+pub struct DictIterator {
```
OddCoincidence: Nit - could we call this `DictIter`? (just `Iter` is more common and also serves to distinguish between iterators for use from rust vs from python)

vm/src/obj/objdict.rs
```diff
@@ -247,6 +234,54 @@ impl ItemProtocol for PyDictRef {
...
+    dict: PyDictRef,
+    position: Cell<usize>,
```
OddCoincidence: Why the `Cell`? `Iterator::next` takes `&mut self`

vm/src/obj/objobject.rs
```diff
@@ -237,7 +237,7 @@ pub fn get_attributes(obj: &PyObjectRef) -> PyAttributes {
 
     // Get instance attributes:
     if let Some(dict) = &obj.dict {
-        for (key, value) in dict.get_key_value_pairs() {
+        for (key, value) in dict.into_iter() {
```
OddCoincidence: I think the `into_iter()` is redundant, `for (key, value) in dict` should be desugared to the same thing

vm/src/stdlib/json.rs
```diff
@@ -64,7 +64,7 @@ impl<'s> serde::Serialize for PyObjectSerializer<'s> {
             serialize_seq_elements(serializer, &elements)
         } else if objtype::isinstance(self.pyobject, &self.vm.ctx.dict_type()) {
             let dict: PyDictRef = self.pyobject.clone().downcast().unwrap();
-            let pairs = dict.get_key_value_pairs();
+            let pairs: Vec<(PyObjectRef, PyObjectRef)> = dict.into_iter().collect();
```
OddCoincidence: nit: the item type is implied by the dict, so this could just be `let pairs: Vec<_> = ...`

vm/src/obj/objdict.rs
cthulahoops: Good tip. I think we want the other way around: ie. ```impl<T> IntoIterator for &PyRef<T>```, otherwise you always clone.

cthulahoops: Or maybe, not... you mean PyDict not PyDictRef.

cthulahoops: But, I don't think you implement it for ```&PyDict``` because you need the PyRef in order to clone.

OddCoincidence: Why do you need to clone? Why not:

```rust
struct DictIter<'a> {
    dict: &'a PyDict,
    ...
}
```

vm/src/obj/objdict.rs
cthulahoops: Originally just because that's how I had to do the python ones. It works here though. I have to get rid of the ```.map``` to avoid ```closure requires unique access to `self` but it is already borrowed```.

(I'm not sure how I feel about ```.map```.)

vm/src/obj/objdict.rs
cthulahoops: I'll give that a go. Thanks.

vm/src/frame.rs
```diff
@@ -393,9 +393,8 @@ impl Frame {
                         // Take all key-value pairs from the dict:
                         let dict: PyDictRef =
                             obj.downcast().expect("Need a dictionary to build a map.");
-                        let dict_elements = dict.get_key_value_pairs();
-                        for (key, value) in dict_elements.iter() {
-                            map_obj.set_item(key.clone(), value.clone(), vm).unwrap();
+                        for (key, value) in dict {
```
windelbouwman: This is awesome, good improvement!

vm/src/obj/objdict.rs
cthulahoops: I get as far as this

```
impl<T> IntoIterator for PyRef<T>
where
    T: IntoIterator + PyValue,
{
    type Item = <T as IntoIterator>::Item;
    type IntoIter = <T as IntoIterator>::IntoIter;

    fn into_iter(self) -> Self::IntoIter {
        self.deref().into_iter()
    }
}
```

which gives:

```
266 |         self.deref().into_iter()
    |         ^^^^^^^^^^^^ cannot move out of borrowed content
```

vm/src/frame.rs
cthulahoops: It's fun, isn't. It's lovely how simple things become if you get right traits in place.

vm/src/obj/objdict.rs
cthulahoops: I think I'll merge what I've got here and open a separate PR for this improvement.

PR #814, opened 2019-04-10
Change #[pymethod(property)] to #[pyproperty]

coolreader18:
I have come to the conclusion that it is impossible to write clean code for a macro implementation.

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -6,11 +6,11 @@ use syn::{Attribute, AttributeArgs, Ident, ImplItem, Item, Lit, Meta, MethodSig,
 
 enum ClassItem {
     Method {
-        item_name: Ident,
+        item_ident: Ident,
         py_name: String,
     },
     Property {
-        item_name: Ident,
+        item_ident: Ident,
         py_name: String,
         setter: bool,
     },
@@ -60,7 +60,7 @@ impl ClassItem {
                     }
                 }
                 item = Some(ClassItem::Method {
-                    item_name: sig.ident.clone(),
+                    item_ident: sig.ident.clone(),
                     py_name: py_name.unwrap_or_else(|| sig.ident.to_string()),
                 });
                 attr_idx = Some(i);
@@ -96,9 +96,9 @@ impl ClassItem {
                     }
                 }
                 let py_name = py_name.unwrap_or_else(|| {
-                    let item_name = sig.ident.to_string();
-                    if item_name.starts_with("set_") {
-                        let name = &item_name["set_".len()..];
+                    let item_ident = sig.ident.to_string();
+                    if item_ident.starts_with("set_") {
+                        let name = &item_ident["set_".len()..];
                         if name.is_empty() {
                             panic!(
                                 "A #[pyproperty(setter)] fn with a set_* name have something \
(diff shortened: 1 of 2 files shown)


OddCoincidence:
Could you also convert a built-in with properties to serve as a test and example?

derive/src/pyclass.rs
```diff
@@ -1,91 +1,138 @@
...
-impl MethodKind {
-    fn to_ctx_constructor_fn(&self) -> Ident {
-        let f = match self {
-            MethodKind::Method => "new_rustfunc",
-            MethodKind::Property => "new_property",
-        };
-        Ident::new(f, Span::call_site())
-    }
+struct ClassItem {
+    item_name: Ident,
```
OddCoincidence: nit: `item_ident`

derive/src/pyclass.rs
```diff
@@ -1,91 +1,138 @@
...
 
-impl MethodKind {
-    fn to_ctx_constructor_fn(&self) -> Ident {
-        let f = match self {
-            MethodKind::Method => "new_rustfunc",
-            MethodKind::Property => "new_property",
-        };
-        Ident::new(f, Span::call_site())
-    }
+struct ClassItem {
```
OddCoincidence: In the spirit of "make invalid states unrepresentable", could this and `ClassKindKind` be combined into an enum like:

```rustt
enum ClassItem {
    Method {
        name: String,
        ...
    },
    Property {
        setter: bool
    }
```

i.e. each variant represents an attribute, each field represents metadata and/or an "argument" to the attribute

derive/src/pyclass.rs
```diff
@@ -1,91 +1,138 @@
...
-                    // implementation for it
-                    pymethod_to_remove.push(i);
+        {
+            let name = meta.name();
+            if name == "pymethod" {
+                if item.is_some() {
+                    panic!("You can only have one #[py*] attribute on an impl item")
+                }
+                let nesteds = meta_to_vec(meta)
+                    .expect("#[pymethod = ...] must be a list, e.g. #[pymethod(...)]");
```
OddCoincidence: I'm having trouble parsing this error message. Is `#[pymethod = ...]` actually allowed? If so, what does it do?

coolreader18: No, it's not. The idea is that the error message will only be shown if somebody did write `#[pymethod = ""]`, so that would be a signal. Although, do we want to allow that? Basically as an alias to `#[pymethod (name = "")]`?

derive/src/pyclass.rs
coolreader18: That feels better, thanks.

derive/src/pyclass.rs
OddCoincidence: I think I'd prefer not to allow that syntax in case we want additional options in the future. e.g. I've been considering the idea of `#[pymethod(constructor)]` that allows `new` methods to be more idiomatic rust.

coolreader18:
Hmm, the test is failing because the `extend_class!` implementation used `PyInt::add` for both `__add__` and `__radd__`. Should we allow `name` to be a list or just duplicate the `add` function? I'm going to go with the latter for now but if you think differently LMK.

coolreader18:
All concerns have been addressed, so I'll merge now.

PR #816, opened 2019-04-11
Guard for changes in dictionary size during iteration.

cthulahoops:


vm/src/obj/objdict.rs
```diff
@@ -313,25 +311,28 @@ macro_rules! dict_iterator {
         #[derive(Debug)]
         struct $iter_name {
             pub dict: PyDictRef,
-            pub position: Cell<usize>,
+            pub position: RefCell<dictdatatype::DictPosition>,
```
coolreader18: If you made `DictPosition` `Copy`, you could keep this as a `Cell`.

coolreader18: Actually, does this even need to be mutable? It looks like you're not changing it in `__next__`.

cthulahoops: Thinking more on this, given the eventual implementation I use position and size in different places so there's no need to bundle them together.

vm/src/obj/objdict.rs
```diff
@@ -315,11 +318,8 @@ impl Iterator for DictIter {
     type Item = (PyObjectRef, PyObjectRef);
 
     fn next(&mut self) -> Option<Self::Item> {
-        match self.dict.entries.borrow().next_entry(self.position) {
-            Some((new_position, key, value)) => {
-                self.position = new_position;
-                Some((key.clone(), value.clone()))
-            }
```
coolreader18: You could just do `.cloned()` here.

cthulahoops: I don't think it works through the tuple.

PR #818, opened 2019-04-11
Fix dir bug - not including object attributes correctly.

cthulahoops:
On master:

```
>>>>> dir(lambda: 1)
["'str' object", ...
```

This was introduced when we got non-string dictionary keys. We now fail in a slightly different way than CPython but it's a an obscure enough edge case that I don't think it matters:

```
>>> a.__dict__[9] = 2
>>> dir(a)
TypeError: '<' not supported between instances of 'str' and 'int'
```

vs

```
>>>>> a.__dict__[9] = 2
>>>>> dir(a)
TypeError: Attribute is not a string: "9"
```

### DIFF
--- a/tests/snippets/builtin_dir.py
+++ b/tests/snippets/builtin_dir.py
@@ -8,6 +8,9 @@ a = A()
 assert "test" in dir(a), "test not in a"
 assert "test" in dir(A), "test not in A"
 
+a.x = 3
+assert "x" in dir(a), "x not in a"
+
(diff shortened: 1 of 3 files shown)


cthulahoops:
This implementation now matches Cpython more closely:

```
>>>>> dir(a)
TypeError: Cannot compare __delattr__ and 'int' object
```

(We need to come up with a better interface for exception formatting!)

PR #819, opened 2019-04-11
Add os.environ

palaviv:
I tried using only the low level calls (`getenv`, `setenv`, `putenv`) but python `os.getenv` read only from the `os.environ`.
I copied `_collection_abc.py` from Cpython and commented out the code that did not work yet.
In `os.py` I copied only the parts related to `_Environ` and commented out the unsupported code as well.


windelbouwman:
Looks good! Probably this is the way to go with the standard library! Nicely done!

PR #821, opened 2019-04-12
Add exception cause

palaviv:
You can now do:
```py
try:
    raise ZeroDivisionError
except ZeroDivisionError as ex:
    raise NameError from ex
```
and the output is:
```py
  File tests/snippets/try_exceptions.py, line 131, in <module>
ZeroDivisionError: No msg

  File tests/snippets/try_exceptions.py, line 133, in <module>
NameError: No msg
```

### DIFF
--- a/tests/snippets/try_exceptions.py
+++ b/tests/snippets/try_exceptions.py
@@ -95,3 +95,35 @@ except TypeError as ex:
     l.append(3)
     print('boom', type(ex))
 assert l == [1, 3]
+
+cause = None
+try:
+    try:
+        raise ZeroDivisionError
+    except ZeroDivisionError as ex:
+        assert ex.__cause__ == None
+        cause = ex
+        raise NameError from ex
+except NameError as ex2:
+    assert ex2.__cause__ == cause
+
+try:
+    raise ZeroDivisionError from None
+except ZeroDivisionError as ex:
+    assert ex.__cause__ == None
+
+try:
+    raise ZeroDivisionError
+except ZeroDivisionError as ex:
+    assert ex.__cause__ == None
+
+try:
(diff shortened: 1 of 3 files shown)


tests/snippets/try_exceptions.py
```diff
@@ -95,3 +95,35 @@ def __init__(self):
...
+
+try:
+    raise ZeroDivisionError
+except ZeroDivisionError as ex:
+    assert ex.__cause__ == None
+
+try:
+    raise ZeroDivisionError from 5
+except TypeError:
+    pass
```
coolreader18: You probably want to use `assertRaises`  here.

vm/src/frame.rs
```diff
@@ -1204,6 +1191,30 @@ impl Frame {
...
+        } else if let Ok(exception) = PyClassRef::try_from_object(vm, exception) {
+            if objtype::issubclass(&exception, &vm.ctx.exceptions.base_exception_type) {
+                let exception = vm.new_empty_exception(exception)?;
+                Ok(exception)
+            } else {
+                let msg = format!(
+                    "Can only raise BaseException derived types, not {}",
+                    exception
+                );
+                let type_error_type = vm.ctx.exceptions.type_error.clone();
```
coolreader18: You can use `vm.new_type_error` here.

coolreader18:
Looks good! A couple things, and then I'll merge.

tests/snippets/try_exceptions.py
palaviv: Done

vm/src/frame.rs

palaviv:
Fixed. implicit chaining coming soon :)

PR #822, opened 2019-04-12
add capitalize bytes.center + fix str.center

jgirardet:
Add bytes.center
fix str.center
add some tests

introduce `is_byt`e to get value of a byte object : byte or byterray.
I'cant get out of a clone() in this function. maybe you have an idea ?

Should follow in another  PR : is_bytes_like : which basically would do the same thing, including memory view.

also added a get_value() to PyBytes as convenient way to deal with PyBytes Value in other function. Id'l like it be also added to byterray when refactoring it with PyByteInner.

### DIFF
--- a/vm/src/obj/objbyteinner.rs
+++ b/vm/src/obj/objbyteinner.rs
@@ -1,4 +1,5 @@
 use crate::pyobject::PyObjectRef;
+use num_bigint::BigInt;
 
 use crate::function::OptionalArg;
 
@@ -13,7 +14,11 @@ use std::hash::{Hash, Hasher};
 use super::objint;
 use super::objsequence::PySliceableSequence;
 use crate::obj::objint::PyInt;
+use num_integer::Integer;
 use num_traits::ToPrimitive;
+
+use super::objbytearray::{get_value as get_value_bytearray, PyByteArray};
+use super::objbytes::PyBytes;
 
 #[derive(Debug, Default, Clone)]
 pub struct PyByteInner {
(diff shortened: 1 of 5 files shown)


vm/src/obj/rien.rs
```diff
@@ -0,0 +1 @@
+use crate::obj:
```
windelbouwman: What is the purpose of this file?

windelbouwman:
You added two empty files, `Pipfile` and `tests/snippets/resources/__init__.py`, what are they for?

vm/src/obj/rien.rs
jgirardet: Oh sorry i didn't see all the mess in my commit. I 'll fix it tomorrow

jgirardet:
Ok that"s fixed.

windelbouwman:
Looks good to me! We might be able to share some code between `str`, `bytes` and `bytearray`, but I leave that for a later opportunity.

vm/src/obj/objbyteinner.rs
```diff
@@ -346,4 +360,39 @@ impl PyByteInner {
...
+        // merge all
+        let mut res = vec![fillbyte; ln];
+        res.extend_from_slice(&self.elements[..]);
+        res.extend_from_slice(&vec![fillbyte; rn][..]);
+
+        res
+    }
+}
+
+pub fn is_byte(obj: &PyObjectRef) -> Option<Vec<u8>> {
```
adrian17: could this return a ref to the Vec instead of copying it?

PR #823, opened 2019-04-12
Drop IntoPyNativeFunc impl to fix Nightly/Beta breakages.

adrian17:
Let's hope the checks go fine now.

### DIFF
--- a/vm/src/function.rs
+++ b/vm/src/function.rs
@@ -437,12 +437,6 @@ where
     }
 }
 
-impl IntoPyNativeFunc<PyFuncArgs, PyResult> for PyNativeFunc {
-    fn into_func(self) -> PyNativeFunc {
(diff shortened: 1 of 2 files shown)


coolreader18:
It looks like you need to run rustfmt on this, but then it's good to go!

adrian17:
Hmm, I ran rustfmt on this, you can see Travis is showing unrelated chunks of code. Will fix them, of course.

Also, it seems like WASM is now failing again... but for a non-Rust related reason? Am I reading this right?

coolreader18:
Try updating your rustfmt, Rust 1.34 came out yesterday and that might be the reason. WASM is a separate issue though, we can try to fix that with a different PR. I'll merge this I think, cause the would have happened with any new PR.

PR #824, opened 2019-04-13
Run CPython's test_bool.py via unittest

skinny121:
I have managed to run the test_bool.py using unittest on RustPython.

I have copied the unittest package and its standard library dependencies over from CPython, and the pure python version of itertools.py from https://github.com/beeware/ouroboros. 

Again I had to make changes to some of the standard library files to make unittest work. I had to make quite significant changes to unittest itself, but apart from disabling printing of tracebacks, none of the changes should affect the functionality of the library much. Note only limited functionality of unittest actually works at this stage.

In addition to some of my recent PR I have made the following changes to RustPython itself:
- Ability to do an absolute import of a non-top-level module. i.e. `import unittest.case`. This isn't implemented properly but is a hack to get importing of those to happen. 
- Add `function.__call__`
- Add `io.IOBase.flush`
- Add skeleton `_thread` module
- Expose `types.MethodType`
- Add fake `sys.intern` and `sys.warnoptions`
- Make `classmethod` have a dict by default.

This PR includes #817.


vm/src/eval.rs
```diff
@@ -15,7 +13,7 @@ pub fn eval(vm: &VirtualMachine, source: &str, scope: Scope, source_path: &str)
         }
         Err(err) => {
             let syntax_error = vm.context().exceptions.syntax_error.clone();
-            Err(vm.new_exception(syntax_error, err.description().to_string()))
```
coolreader18: You can use `err.to_string()` here.

vm/src/function.rs
```diff
@@ -274,8 +280,16 @@ impl<T> IntoIterator for KwArgs<T> {
 ///
 /// `Args` optionally accepts a generic type parameter to allow type checks
 /// or conversions of each argument.
+#[derive(Clone)]
 pub struct Args<T = PyObjectRef>(Vec<T>);
 
+impl<T: PyValue> Args<PyRef<T>> {
+    pub fn into_tuple(self, vm: &VirtualMachine) -> PyObjectRef {
+        vm.ctx
```
coolreader18: You can change `|obj| obj.into_object()` into `PyRef::into_object`.

vm/src/import.rs
```diff
@@ -25,7 +24,7 @@ fn import_uncached_module(vm: &VirtualMachine, current_path: PathBuf, module: &s
     let file_path = find_source(vm, current_path, module)
         .map_err(|e| vm.new_exception(notfound_error.clone(), e))?;
     let source = util::read_file(file_path.as_path())
-        .map_err(|e| vm.new_exception(import_error.clone(), e.description().to_string()))?;
```
coolreader18: You can just use `.to_string()` instead of `format!`

vm/src/import.rs
```diff
@@ -34,7 +33,7 @@ fn import_uncached_module(vm: &VirtualMachine, current_path: PathBuf, module: &s
     )
     .map_err(|err| {
         let syntax_error = vm.context().exceptions.syntax_error.clone();
```
coolreader18: And here

vm/src/stdlib/io.rs
```diff
@@ -95,6 +95,10 @@ fn io_base_cm_exit(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     Ok(vm.get_none())
 }
 
+fn io_base_flush(_zelf: PyObjectRef, _vm: &VirtualMachine) -> () {
+    ()
```
coolreader18: Is this meant to do nothing? Also you can remove the return type, by default it's `()`

vm/src/stdlib/thread.rs
```diff
@@ -0,0 +1,32 @@
...
+use super::super::pyobject::PyObjectRef;
+use crate::function::PyFuncArgs;
+use crate::pyobject::PyResult;
+use crate::vm::VirtualMachine;
+
+fn rlock_acquire(vm: &VirtualMachine, _args: PyFuncArgs) -> PyResult {
+    Ok(vm.get_none())
+}
+
+fn rlock_release(_zelf: PyObjectRef, _vm: &VirtualMachine) -> () {
```
coolreader18: Same with the return type here.

vm/src/stdlib/io.rs
skinny121: The only thing this base method is meant to do is check if it's closed and if so then raise a ValueError. But as IOBase doesn't yet have a `close` method or attribute, there is currently nothing for `flush` to do. Subclasses are meant to override this and do a real flush.

skinny121:
Only 13 of the 28 tests in test_bool.py actually pass, but that is good for this stage.

skinny121:
I have added a Python script that displays the diff between the files that we have copied over from CPython, and the original files within the CPython source repository. Hopefully, this will allow us to keep track of the changes we make to them to ensure we don't diverge too much.

cthulahoops:
This is awesome work, thank you!

I think there's a bit too much in here for a single PR though, making it a bit hard to review. Could we get some of this split up a bit: rustpython fixes, addition of cpython library code, creation of test harness.

I'm quite scared about copying code from cpython, especially when we then go on to patch the code. Is there an alternative to maintaining a fork of the cpython stdlib? Are we preferring it to ouroboros / maintaining a fork of parts of both? Is it possible to work off a checkout of one 0or the other? (I've been mostly staying out of the stdlib side of things so not sure exactly where this is going.)

adrian17:
Yeah, this is way too huge. Looks like it has quite a lot of hacks and commented code sprinkled in, that'd be hard to sort out later. Mixing library code from CPython and ouroboros will get really messy really fast too.
Also, the commit list seems to include changes from other PRs?

Anyway, I actually was also experimenting with moving `test_bool.py` several weeks ago - if that's the main goal, it was quite easy to do with a very basic self-implemented `unittest` mock and without adding any other libraries.

windelbouwman:
This is a great step! Nice job! I propose splitting this work up into sizeable pieces, and merge the work one by one, as adrian suggested. I consider this PR as a proof of concept from which we can draw.

Another note, I feel we should seperate tests using the standard library in a different folder than the test snippets. This way, we can run the test snippets without any standard library what so ever. What do you say?

windelbouwman:
Another question, should we consider creating a new attempt along the same lines of ouroboros?

skinny121:
Ok, I will split this PR up more.

I approached this as more of a proof of concept to see where our shortcomings are. Over time once we fix these shortcomings we could move to use CPython standard library directly. But until then we need a place to make changes to the standard library to work around issues. But any ideas are welcome to how to make this work short and long term.

@windelbouwman Possibly, but one of the issues I discovered was our import machinery was lacking. Fortunately, `importlib` package inside the standard library contains a pure python implementation of the import system. But that would mean we would require a bit of the standard library to be able to do anything. Is this a problem?

windelbouwman:
I don't think having some logic for importing in importlib would be a problem. In fact, I think it is probably good to implement some specifics in python itself. As I understood, a lot of how python works, is actually implemented in python itself. Notable examples of this are the importing scheme and the REPL.

coolreader18:
Do you think this could be closed now @skinny121? I think now we have a clearer view of what we want to do in order to run `unittest`, and we're almost there at this point.

skinny121:
Yes, agree, we are making good progress towards being able to run `unittest` without modifications, although there is still a bit of work to go.

PR #827, opened 2019-04-13
str: improve {is,}title impl and add tests

OddCoincidence:
The existing implementation didn't actually title-case, it just naively capitalized the first letter of each word. Also note that this is _still_ not 100% on-par with cpython because it doesn't handle unicode titlecase characters, e.g. digraphs that consist of an uppercase and a lowercase letter. It looks like rust used to support this on nightly [but removed it because they couldn't add corresponding methods on strings without a `unicode-segmentation` dependency](https://github.com/rust-lang/rust/issues/26555). I'll see if I can extract that support into its own crate.

Also note that I added the tests as _rust_ unit tests; this is getting significantly less painful with recent refactors and I think we should do it more.

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -1038,7 +1038,7 @@ mod tests {
             ("Greek Ωppercases ...", "greek ωppercases ..."),
         ];
         for (title, input) in tests {
-            assert_eq!(PyString::from(input).title(&vm), String::from(title));
+            assert_eq!(PyString::from(input).title(&vm).as_str(), title);
         }
     }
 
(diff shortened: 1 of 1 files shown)


vm/src/obj/objstr.rs
```diff
@@ -967,22 +1020,55 @@ fn adjust_indices(
...
+        let vm = VirtualMachine::new();
+
+        let tests = vec![
+            (" Hello ", " hello "),
+            ("Hello ", "hello "),
+            ("Hello ", "Hello "),
+            ("Format This As Title String", "fOrMaT thIs aS titLe String"),
+            ("Format,This-As*Title;String", "fOrMaT,thIs-aS*titLe;String"),
+            ("Getint", "getInt"),
+            ("Greek Ωppercases ...", "greek ωppercases ..."),
```
coolreader18: Nit: you don't need to allocate using `String::from`, you can just directly compare `title` and `&PyString::...`. I think.

PR #829, opened 2019-04-14
RustPython changes for #824

skinny121:
Split from #824 as requested. These are the changes to RustPython itself to support importing and running unittest.

The changes are:
- Ability to do an absolute import of a non-top-level module. i.e. import unittest.case. This isn't implemented properly but is a hack to get importing of those to happen.
- Add function.__call__
- Add io.IOBase.flush
- Add skeleton _thread module
- Expose types.MethodType
- Add fake sys.intern and sys.warnoptions
- Make classmethod have a dict by default.


windelbouwman:
Looks good to me too!

PR #831, opened 2019-04-14
Exception reraise and context

palaviv:
This PR add support for reraise:
```py
>>>>> try:
.....     raise NameError
..... except:
.....     raise
..... 
  File <stdin>, line 4, in <module>
  File <stdin>, line 2, in <module>
NameError: No msg
```

And exception context:
```py
>>>>> try:
.....     raise ZeroDivisionError
..... except ZeroDivisionError as ex:
.....     raise NameError
..... 
  File <stdin>, line 2, in <module>
ZeroDivisionError: No msg

During handling of the above exception, another exception occurred:

  File <stdin>, line 4, in <module>
NameError: No msg
```

### DIFF
--- a/tests/snippets/try_exceptions.py
+++ b/tests/snippets/try_exceptions.py
@@ -126,18 +126,14 @@ try:
 except ZeroDivisionError as ex:
     assert type(ex.__cause__) == NameError
 
-try:
+with assertRaises(NameError):
     try:
         raise NameError
     except:
         raise
-except NameError:
-    pass
 
-try:
+with assertRaises(RuntimeError):
     raise
-except RuntimeError:
-    pass
 
 context = None
 try:
--- a/tests/snippets/try_exceptions.py
+++ b/tests/snippets/try_exceptions.py
@@ -125,3 +125,15 @@ try:
 except ZeroDivisionError as ex:
     assert type(ex.__cause__) == NameError
 
+try:
+    try:
+        raise NameError
+    except:
+        raise
+except NameError:
+    pass
+
+try:
+    raise
+except RuntimeError:
+    pass
--- a/tests/snippets/try_exceptions.py
+++ b/tests/snippets/try_exceptions.py
@@ -137,3 +137,8 @@ try:
     raise
 except RuntimeError:
     pass
+
+try:
+    raise ZeroDivisionError
+except ZeroDivisionError as ex:
+    assert ex.__context__ == None
--- a/tests/snippets/try_exceptions.py
+++ b/tests/snippets/try_exceptions.py
@@ -161,3 +161,13 @@ try:
 except ZeroDivisionError as ex:
     assert type(ex.__cause__) == NameError
     assert ex.__context__ == None
+
+try:
+    try:
+        raise ZeroDivisionError
+    except ZeroDivisionError as ex:
+        pass
+    finally:
+        raise NameError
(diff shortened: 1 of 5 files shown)


tests/snippets/try_exceptions.py
```diff
@@ -125,3 +126,48 @@ def __init__(self):
 except ZeroDivisionError as ex:
     assert type(ex.__cause__) == NameError
 
+try:
+    try:
+        raise NameError
+    except:
+        raise
+except NameError:
```
coolreader18: Use `assertRaises` here

vm/src/exceptions.rs
```diff
@@ -20,12 +20,22 @@ fn exception_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
 
 // print excption chain
 pub fn print_exception(vm: &VirtualMachine, exc: &PyObjectRef) {
+    let mut had_casue = false;
```
coolreader18: Typo: `had_cause`

vm/src/frame.rs
```diff
@@ -715,18 +715,38 @@ impl Frame {
...
+                    },
                     1 | 2 => self.get_exception(vm, false)?,
-                    0 | 3 => panic!("Not implemented!"),
+                    3 => panic!("Not implemented!"),
                     _ => panic!("Invalid parameter for RAISE_VARARGS, must be between 0 to 3"),
                 };
-                info!("Exception raised: {:?} with cause: {:?}", exception, cause);
+                let context = match argc {
+                    0 => vm.get_none(), // We have already got the exception,
+                    _ => match in_exc {
```
coolreader18: You could just use an if/else here

coolreader18:
Looks good!

tests/snippets/try_exceptions.py
palaviv: Done

vm/src/exceptions.rs
palaviv: Fixed

vm/src/frame.rs
palaviv: Changed. Not sure if it looks better.

palaviv:
Thanks. Pushed the changes.

vm/src/frame.rs
coolreader18: Oh sorry I meant for matching on `in_exc`, the line selection for reviews for the GitHub client I'm using is weird.

palaviv: I think it looks better with the match. Would you be fine with it staying as it is?

coolreader18: Sure, np

vm/src/compile.rs
```diff
@@ -24,6 +24,7 @@ struct Compiler {
     current_qualified_path: Option<String>,
     in_loop: bool,
     in_function_def: bool,
+    in_exc_handler: bool,
```
adrian17: This looks like a thing that should be handled at runtime, rather than at compile time. As of now, while this code does handle
```python
try:
	1/0
except:
	raise
```
, it doesn't handle anything more complex, like
```python
def f():
    raise

try:
    1/0
except:
    f()
```

palaviv: I agree. CPython has a stack of exceptions so we can do something similar.

PR #832, opened 2019-04-14
str: proper titlecase support

OddCoincidence:
Follow-up to #827 that adds proper titlecase support using a new crate https://github.com/OddCoincidence/unicode-casing.


vm/Cargo.toml
```diff
@@ -23,6 +23,7 @@ regex = "1"
 rustc_version_runtime = "0.1.*"
 statrs = "0.10.0"
 caseless = "0.2.1"
+unicode-casing = { git = "https://github.com/OddCoincidence/unicode-casing" }
```
windelbouwman: Could you create a release from this package, so we can refer to a pinned version?

vm/src/obj/objstr.rs
```diff
@@ -413,12 +414,12 @@ impl PyString {
         for c in self.value.chars() {
             if c.is_lowercase() {
                 if !previous_is_cased {
-                    title.extend(c.to_uppercase());
+                    title.extend(c.to_titlecase());
```
windelbouwman: How can a single character be title cases? I guess something else is meant than python's titlecase function?

OddCoincidence: This surprised me too, but there indeed exist single codepoint titlecase characters (I realized this from reading the cpython tests for `title` / `istitle`). [Some examples](https://github.com/OddCoincidence/unicode-casing/blob/master/src/lib.rs#L104-L105).

vm/Cargo.toml
OddCoincidence: You mean on crates.io? I was going to see if the rust unicode folks wanted control over this package before submitting. It's fine if you want to wait for that, but I can also add the commit sha here if that makes you feel better.

windelbouwman: Ah, okay, yes, rust unicode group having this package would be good. Let's keep it like this, until this is effect. Maybe add a todo?

vm/src/obj/objstr.rs
windelbouwman: "Latin Capital Letter D with Small Letter Z with Caron". Unicode is never boring! :)

PR #833, opened 2019-04-14
Add bytes.count and bytes.join

jgirardet:
-  add bytes.count
- add IsByte trait to BigInt to always return error if not an u8 : If you are ok with this, I could modify some more things to use it.
- add is_bytes_like : reusable function to check for a "python" bytes like object.
TODO : check if slice argument is an object with a  `__index__` method. I don't know how to do it, need some indication.
- add bytes.join

### DIFF
--- a/tests/snippets/bytes.py
+++ b/tests/snippets/bytes.py
@@ -158,6 +158,7 @@ with assertRaises(TypeError):
 with assertRaises(TypeError):
     b"b".center(2, b"ba")
 b"kok".center(5, bytearray(b"x"))
+b"kok".center(-5,)
 
 # count
 assert b"azeazerazeazopia".count(b"aze") == 3
@@ -177,6 +178,7 @@ assert b"azeazerazeazopia".count(b"aze", -13, -10) == 1
 assert b"azeazerazeazopia".count(b"aze", 1, 10000) == 2
 with assertRaises(ValueError):
     b"ilj".count(3550)
+assert b"azeazerazeazopia".count(97) == 5
 
 # join
 assert (
--- a/tests/snippets/bytes.py
+++ b/tests/snippets/bytes.py
@@ -158,3 +158,11 @@ with assertRaises(TypeError):
 with assertRaises(TypeError):
     b"b".center(2, b"ba")
 b"kok".center(5, bytearray(b"x"))
+
+# count
+assert b"azeazerazeazopia".count(b"aze") == 3
+assert b"azeazerazeazopia".count(b"az") == 4
+assert b"azeazerazeazopia".count(b"a") == 5
+assert b"123456789".count(b"") == 10
+assert b"azeazerazeazopia".count(bytearray(b"aze")) == 3
+assert b"azeazerazeazopia".count(memoryview(b"aze")) == 3
--- a/tests/snippets/bytes.py
+++ b/tests/snippets/bytes.py
@@ -166,3 +166,14 @@ assert b"azeazerazeazopia".count(b"a") == 5
 assert b"123456789".count(b"") == 10
 assert b"azeazerazeazopia".count(bytearray(b"aze")) == 3
 assert b"azeazerazeazopia".count(memoryview(b"aze")) == 3
+assert b"azeazerazeazopia".count(memoryview(b"aze"),1,9 ) == 1
+assert b"azeazerazeazopia".count(b"aze", None, None) == 3
+assert b"azeazerazeazopia".count(b"aze", 2, None) == 2
+assert b"azeazerazeazopia".count(b"aze", 2) == 2
+assert b"azeazerazeazopia".count(b"aze", None, 7) == 2
+assert b"azeazerazeazopia".count(b"aze", None, 7) == 2
+assert b"azeazerazeazopia".count(b"aze", 2, 7) == 1
+assert b"azeazerazeazopia".count(b"aze", -13, -10) == 1
+assert b"azeazerazeazopia".count(b"aze", 1, 10000) == 2
+with assertRaises(ValueError):
+    b"ilj".count(355)
--- a/tests/snippets/bytes.py
+++ b/tests/snippets/bytes.py
@@ -166,7 +166,7 @@ assert b"azeazerazeazopia".count(b"a") == 5
 assert b"123456789".count(b"") == 10
 assert b"azeazerazeazopia".count(bytearray(b"aze")) == 3
 assert b"azeazerazeazopia".count(memoryview(b"aze")) == 3
-assert b"azeazerazeazopia".count(memoryview(b"aze"),1,9 ) == 1
+assert b"azeazerazeazopia".count(memoryview(b"aze"), 1, 9) == 1
 assert b"azeazerazeazopia".count(b"aze", None, None) == 3
 assert b"azeazerazeazopia".count(b"aze", 2, None) == 2
 assert b"azeazerazeazopia".count(b"aze", 2) == 2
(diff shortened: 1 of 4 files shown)


vm/src/obj/objbyteinner.rs
```diff
@@ -387,6 +389,73 @@ impl PyByteInner {
...
+            {
+                total += 1;
+            }
+            i_start += 1;
+        }
+        Ok(total)
+    }
+
+    pub fn join(&self, iter: PyIterable, vm: &VirtualMachine) -> PyResult {
+        let mut refs = vec![];
```
windelbouwman: Could you give a better name to this variable? Suggestions: `joined_bytes`, `result_bytes`?

tests/snippets/bytes.py
```diff
@@ -158,3 +158,30 @@
...
+assert b"azeazerazeazopia".count(b"aze", None, None) == 3
+assert b"azeazerazeazopia".count(b"aze", 2, None) == 2
+assert b"azeazerazeazopia".count(b"aze", 2) == 2
+assert b"azeazerazeazopia".count(b"aze", None, 7) == 2
+assert b"azeazerazeazopia".count(b"aze", None, 7) == 2
+assert b"azeazerazeazopia".count(b"aze", 2, 7) == 1
+assert b"azeazerazeazopia".count(b"aze", -13, -10) == 1
+assert b"azeazerazeazopia".count(b"aze", 1, 10000) == 2
+with assertRaises(ValueError):
+    b"ilj".count(3550)
```
windelbouwman: Could you add a valid test case for counting an int value inside bytes?

vm/src/obj/objbytes.rs
```diff
@@ -269,6 +268,28 @@ impl PyBytesRef {
...
+
+    #[pymethod(name = "count")]
+    fn count(
+        self,
+        sub: PyObjectRef,
+        start: OptionalArg<PyObjectRef>,
+        end: OptionalArg<PyObjectRef>,
+        vm: &VirtualMachine,
+    ) -> PyResult {
+        match is_bytes_like(&sub) {
```
windelbouwman: I suggest to move this logic about sub being an integer value into the inner object, so we can re-use it for `bytearray`.

vm/src/obj/objbytes.rs
```diff
@@ -269,6 +268,28 @@ impl PyBytesRef {
...
+            Some(value) => Ok(vm.new_int(self.inner.count(value, start, end, vm)?)),
+            None => match_class!(sub,
+                i @ PyInt => {
+                    Ok(vm.new_int(self.inner.count(vec![i.as_bigint().is_byte(vm)?], start, end, vm)?))},
+                obj => {Err(vm.new_type_error(format!("argument should be integer or bytes-like object, not {}", obj)))}),
+        }
+    }
+
+    #[pymethod(name = "join")]
+    fn join(self, iter: PyIterable, vm: &VirtualMachine) -> PyResult {
```
windelbouwman: Maybe also add this method to bytearray.

jgirardet: for now i'd like just focus on bytes method. Bytearray will need after a big cleanup

vm/src/obj/objbyteinner.rs
```diff
@@ -396,3 +465,23 @@ pub fn is_byte(obj: &PyObjectRef) -> Option<Vec<u8>> {
     j @ PyByteArray => Some(get_value_bytearray(&j.as_object()).to_vec()),
     _ => None)
 }
+
+pub fn is_bytes_like(obj: &PyObjectRef) -> Option<Vec<u8>> {
```
adrian17: Same note about binary predicate as below.
Also, could this return a reference instead of copying?

vm/src/obj/objbyteinner.rs
```diff
@@ -387,6 +389,73 @@ impl PyByteInner {
...
+                }
+                Some(value) => refs.extend(value),
+            }
+        }
+
+        Ok(vm.ctx.new_bytes(refs))
+    }
 }
 
 pub fn is_byte(obj: &PyObjectRef) -> Option<Vec<u8>> {
```
adrian17: similarly to below, but I think I posted that review late:
could this return a reference instead of copying?

vm/src/obj/objbyteinner.rs
```diff
@@ -396,3 +465,23 @@ pub fn is_byte(obj: &PyObjectRef) -> Option<Vec<u8>> {
...
+    match_class!(obj.clone(),
+
+    i @ PyBytes => Some(i.get_value().to_vec()),
+    j @ PyByteArray => Some(get_value_bytearray(&j.as_object()).to_vec()),
+    k @ PyMemoryView => Some(k.get_obj_value().unwrap()),
+    _ => None)
+}
+
+pub trait IsByte: ToPrimitive {
+    fn is_byte(&self, vm: &VirtualMachine) -> Result<u8, PyObjectRef> {
```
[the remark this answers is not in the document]
jgirardet: 2 things about it :
- first you have to clone sub : `match_class:(sub.clone.....`
- second the error given by try_from_ object is not the intended : overflow error vs value_error. so I think the try_as_u8 or try_as_bytes would be better

vm/src/obj/objbyteinner.rs
jgirardet: not so easy...
```
Some(&objbytes::get_value(&x))                                                                                                                                                                                              
    |         ^^^^^^-----------------------^                                                                                                                                                                                              
    |         |     |                                                                                                                                                                                                                     
    |         |     temporary value created here                                                                                                                                                                                          
    |         returns a value referencing data owned by the current function 
```
this error : https://doc.rust-lang.org/error-index.html#E0515
I 'm not sure I can return anything else than a Owned value, thats to say a Vec.

adrian17: Hm, main big issue is that it's hard to mix immutable Python objects (that you can just borrow from) and mutable structures (which give you a Ref/RefMut instead of a normal reference). Similar issue exists for List/Tuple. Anyway, maybe not now, but long-term it'll probably need solving.

adrian17: As for your particular error, I think it's easy - `objbytes::get_value` already returns a reference, so I'm guessing you just don't need a `&` there.

jgirardet: no it won"t
 ```
error[E0308]: mismatched types                                                                                                                                                                                                            
   --> vm/src/obj/objbyteinner.rs:489:26                                                                                                                                                                                                  
    |                                                                                                                                                                                                                                     
489 |     _i @ PyBytes => Some(objbytes::get_value(&obj)),                                                                                                                                                                                
    |                          ^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                  
    |                          |                                                                                                                                                                                                          
    |                          expected &[u8], found opaque type                                                                                                                                                                          
    |                          help: consider borrowing here: `&objbytes::get_value(&obj)`                                                                                                                                                
    |                                                                                                                                                                                                                                     
    = note: expected type `&[u8]`                                                                                                                                                                                                         
               found type `impl lazy_static::Deref`   
```
trying to modify objbytes::getvalue  return type to `&[u8]` gives back the previous  E0515 error.

adrian17: Anyway, I agree that it's a hard thing; it can come later, but it definitely should be done at some point.

jgirardet:
ok last commit does : 

move is_byte to try_as_byte
move trait IsByte to ByteOr
refactor center
refactor count
fix center with negative value
add test for count(integer

jgirardet:
I close this for now. it appears that I must refactor  center to fit with ljust and rjust and count(to match kwargs)

PR #834, opened 2019-04-14
Make macros comply fully with Rust 2018

dkaste:
These changes make it possible to successfully call the exported macros from outside the `rustpython_vm` crate.


coolreader18:
Looks great, thanks for contributing! (The Azure Pipelines build is failing on master, it's not related to the changes in this PR)

PR #839, opened 2019-04-16
Add nonlocal support.

windelbouwman:
@cthulahoops I'm a bit confused about the load_cell name. Is it the same as the load_deref opcode in cpython?

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -163,7 +163,7 @@ impl NameProtocol for Scope {
             .iter()
             .skip(1)
             .next()
-            .unwrap()
+            .expect("no outer scope for non-local")
             .set_item(name, value, vm)
             .unwrap();
     }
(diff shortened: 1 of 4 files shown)


cthulahoops:
> @cthulahoops I'm a bit confused about the load_cell name. Is it the same as the load_deref opcode in cpython?

Yes. I think cell, deref, and nonlocal all refer to roughly the same thing. (A cell stores variables for non-local access, and you deref the cell?)

I suggest we rename ```load_cell``` to ```load_nonlocal```.

vm/src/frame.rs
```diff
@@ -157,6 +158,16 @@ impl NameProtocol for Scope {
         None
     }
 
+    fn store_cell(&self, vm: &VirtualMachine, name: &str, value: PyObjectRef) {
+        self.locals
+            .iter()
+            .skip(1)
+            .next()
+            .unwrap()
```
cthulahoops: This panic here is an interesting case. It looks like cpython raises ```SyntaxError``` is it can't find the relevant variable at compile time. You can't create a variable in an outer scope with ```nonlocal```, only change it. (Unlike ```global```.)

We can't do better than this yet, but worth using ```.expect("no outer scope for non-local")```?

windelbouwman: Allright, good suggestion! A next topic for me would be to detect more errors at compile time, like cpython does, and raise proper syntax errors.

PR #840, opened 2019-04-17
Fix os.unsetenv test for windows

coolreader18:
Should we add a `cfg` to not put `unsetenv` into `os` if on Windows?

### DIFF
--- a/tests/snippets/stdlib_os.py
+++ b/tests/snippets/stdlib_os.py
@@ -55,7 +55,6 @@ assert ENV_KEY not in os.environ
 assert os.getenv(ENV_KEY) == None
 
 if os.name == "posix":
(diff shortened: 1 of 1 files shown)


windelbouwman:
It still is a mistery to me why we do not have unsetenv on windows. I think we set the symbol on the os module explicitly (https://github.com/RustPython/RustPython/blob/master/vm/src/stdlib/os.rs#L217).

Anyway, I will merge this fix to repair the windows build.

palaviv:
This should have worked on windows... I planned on using the `environ` to get `%TEMP%` to create a temp directory for tests but now I am not sure if that will work on windows. Any guesses why this is not working on windows?

PR #842, opened 2019-04-17
Add some benchmark scripts.

windelbouwman:


benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,31 @@
+
```
coolreader18: Should there be something in here to compile RustPython like `test_snippets.py` does?

windelbouwman: Probably, I don't know. Maybe also not. I think it is okay if it picks the latest release build.

benchmarks/test_benchmarks.py
```diff
@@ -0,0 +1,31 @@
...
+# Benchmark scripts:
+benchmarks = [
+    ['benchmarks/nbody.py'],
+    ['benchmarks/mandelbrot.py'],
+]
+
+@pytest.mark.parametrize('exe', pythons)
+@pytest.mark.parametrize('args', benchmarks)
+def test_bench(exe, args, benchmark):
```
adrian17: is this being run on every `pytest` invocation?

windelbouwman: No, it is not run on travis, but when you run pytest in the benchmark directory, it will run test_bench for each snippet and each interpreter.

PR #843, opened 2019-04-17
support bytes creation from hex and ascii

jgirardet:
this allows b"blabla\x0f\x1dblabla"


parser/src/lexer.rs
```diff
@@ -1105,6 +1105,84 @@ where
         let tok_end = self.get_pos();
         Ok((tok_start, ty, tok_end))
     }
+
+    fn lex_byte(&self, s: String) -> Result<Vec<u8>, LexicalError> {
```
windelbouwman: This function could be a function outside of the `Lexer` struct, since self was never referenced. I would prefer to extract this function out of the struct.

parser/src/lexer.rs
```diff
@@ -1105,6 +1105,84 @@ where
...
+                'r' => {
+                    if escape {
+                        res.push(13);
+                    } else {
+                        res.push(114)
+                    }
+                    escape = false;
+                }
+                x => {
+                    if hex_on {
```
windelbouwman: I think this if statement must be around the match like this:
```rust
if hex_on {
  process digit
} else {
   match c {
     // the rest
   }
}

File: parser/src/lexer.rs [hunk 3]
```diff
@@ -1105,6 +1105,84 @@ where
...
+                                res.push(u8::from_str_radix(&hex_value, 16).unwrap());
+                                hex_on = false;
+                                hex_value.clear();
+                            }
+                        } else {
+                            return Err(LexicalError::StringError);
+                        }
+                    } else {
+                        if escape {
+                            res.push(92);
```
windelbouwman: Always push 92 here? Not `c`?

jgirardet: yes backslash (92) was seen at previous loop but not added since you need the following char to know if its a "pure" backslash or a backslash of a \xXX. So I add the "pure" backslash at next loop before to add `c` if we are not `hex_on`

coolreader18: Maybe you could do `'\\' as u8` to make it more clear what character you're adding.

jgirardet: I thought everyone knows ascii table by heart :grin:

jgirardet: following clippy i'll use b'x' notation : https://rust-lang.github.io/rust-clippy/master/index.html#char_lit_as_u8

parser/src/lexer.rs
jgirardet: done, with a bigger match, i think, it'll be clearer

PR #844, opened 2019-04-17
Improve syntax error with line information.

windelbouwman:


tests/snippets/invalid_syntax.py
```diff
@@ -0,0 +1,16 @@
...
+
+yield 2
+"""
+
+try:
+    compile(src, 'test.py', 'exec')
+except SyntaxError as ex:
+    assert ex.lineno == 5
+else:
+    raise AssertionError("Must throw syntax error")
```
coolreader18: You can use `assertRaises` here

vm/src/error.rs
```diff
@@ -1,10 +1,26 @@
 use rustpython_parser::error::ParseError;
+use rustpython_parser::lexer::Location;
 
 use std::error::Error;
 use std::fmt;
 
 #[derive(Debug)]
-pub enum CompileError {
+pub struct CompileError {
+    pub error: CompileErrorType,
```
coolreader18: Maybe something like `problem`  or `issue` for naming here?

tests/snippets/invalid_syntax.py
windelbouwman: I also want to check the `lineno` attribute, so thats why I did not use it.

PR #847, opened 2019-04-18
add other bytes method

jgirardet:
some refactoring
some more methods


jgirardet:
I have   : split and rsplit, expandstab ,  partition/reparition, spitlines,  zfill, replace
waiting in another branch. do you  want I push it now, or do you prefer review just this one which is already bi enough  ?

coolreader18:
I think maybe save that for another PR, yes.

vm/src/obj/objbyteinner.rs
```diff
@@ -94,7 +98,148 @@ impl PyByteInner {
...
         }
     }
+}
+
+#[derive(FromArgs)]
+pub struct ByteInnerFindOptions {
+    #[pyarg(positional_only, optional = false)]
+    sub: PyObjectRef,
+    #[pyarg(positional_only, optional = true)]
+    start: OptionalArg<PyObjectRef>,
```
palaviv: This can be `PyIntRef`.

vm/src/obj/objbyteinner.rs
```diff
@@ -94,7 +98,148 @@ impl PyByteInner {
...
+    end: OptionalArg<PyObjectRef>,
+}
+
+impl ByteInnerFindOptions {
+    pub fn get_value(
+        self,
+        elements: &[u8],
+        vm: &VirtualMachine,
+    ) -> PyResult<(Vec<u8>, Range<usize>)> {
+        let sub = match try_as_bytes_like(&self.sub) {
```
palaviv: There are many cases where we can get "bytes like" or `int`. Maybe consider making some sort of abstraction for this case?

vm/src/obj/objbyteinner.rs
jgirardet: It could be  Int or None so I have to accept PyObjectRef

vm/src/obj/objbyteinner.rs
```diff
@@ -384,14 +551,290 @@ impl PyByteInner {
...
-pub fn is_byte(obj: &PyObjectRef) -> Option<Vec<u8>> {
+pub fn try_as_byte(obj: &PyObjectRef) -> Option<Vec<u8>> {
     match_class!(obj.clone(),
 
     i @ PyBytes => Some(i.get_value().to_vec()),
     j @ PyByteArray => Some(get_value_bytearray(&j.as_object()).to_vec()),
     _ => None)
 }
+
+pub fn try_as_bytes_like(obj: &PyObjectRef) -> Option<Vec<u8>> {
```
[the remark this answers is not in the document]
jgirardet: thanks for the review. This was a problem from beginning for me.
REturning PyResult is the easier. I didn't do it it since I wanted to have strictly the same error messages than cpython (which are very inconsistent) but since it's not strictly   needed PyResult is better.
I answer to your following comment  about many times it could be byutes like or int or None. I hesitate to add a `int' and a '`None' flag to `try_as_bytes_like to directly handle those cases.

About your proposal  of `BytesLike`. I'm not sure to understand : What would be inside that struct ? what would  TryFromObject return ? Vec<u8> ? ByteLikes ? PyByteInner ?

vm/src/obj/objbyteinner.rs
palaviv: Can you do `OptinalArg<PyIntRef>`?

vm/src/obj/objbyteinner.rs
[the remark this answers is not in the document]
palaviv: > About your proposal of BytesLike. I'm not sure to understand : What would be inside that struct ? what would TryFromObject return ? Vec ? ByteLikes ? PyByteInner ?

I think `PyByteInner` would be great options here.

> I answer to your following comment about many times it could be byutes like or int or None. I hesitate to add a int' and a 'None' flag to `try_as_bytes_like to directly handle those cases.

I agree with what you say. Maybe using `Either<BytesLike, PyIntRef>`?

vm/src/obj/objbyteinner.rs
coolreader18: Or maybe `OptionalArg<Option<PyIntRef>>`? I'm not sure if `Option` is `FromArgs` yet, but it should be the specified type or Python `None`.

vm/src/obj/objbytes.rs
```diff
@@ -218,56 +214,119 @@ impl PyBytesRef {
...
+        self.inner.startsendswith(suffix, start, end, false, vm)
+    }
+
+    #[pymethod(name = "find")]
+    fn find(self, options: ByteInnerFindOptions, vm: &VirtualMachine) -> PyResult<isize> {
+        self.inner.find(options, false, vm)
+    }
+
+    #[pymethod(name = "index")]
+    fn index(self, options: ByteInnerFindOptions, vm: &VirtualMachine) -> PyResult {
```
palaviv: This can return `PyResult<isize>` as well

vm/src/obj/objbyteinner.rs
```diff
@@ -174,42 +319,52 @@ impl PyByteInner {
...
+    pub fn contains(&self, needle: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        match try_as_bytes_like(&needle) {
+            Some(value) => self.contains_bytes(&value, vm),
+            None => match_class!(needle,
+                i @ PyInt => self.contains_int(&i, vm),
+                obj => {Err(vm.new_type_error(format!("a bytes-like object is required, not {}", obj)))}),
+        }
+    }
+
+    fn contains_bytes(&self, other: &[u8], vm: &VirtualMachine) -> PyResult {
```
palaviv: This can return `bool` as there is not exceptions that can be raised

vm/src/obj/objbyteinner.rs
```diff
@@ -174,42 +319,52 @@ impl PyByteInner {
...
+        {
+            if let Some(idx) = self.elements.get_pos(int.as_bigint().to_i32().unwrap()) {
             Ok(vm.new_int(self.elements[idx]))
         } else {
             Err(vm.new_index_error("index out of range".to_string()))
         }
-    }
-
-    pub fn getitem_slice(&self, slice: &PyObjectRef, vm: &VirtualMachine) -> PyResult {
+    },
```
palaviv: Something is very weird with the indentation here...

vm/src/obj/objbyteinner.rs
```diff
@@ -319,6 +474,18 @@ impl PyByteInner {
         new
     }
 
+    pub fn swapcase(&self, _vm: &VirtualMachine) -> Vec<u8> {
+        let mut new: Vec<u8> = Vec::new();
```
palaviv: You know the size of the result vector so you should initialize the vector with that size.

vm/src/obj/objbyteinner.rs
jgirardet: using  `OptionalArg<Option<PyIntRef>>`, you mean this implementation in impl  block (line 126)?
```
let start = match self.start {
            OptionalArg::Present(Some(int)) => Some(int.as_bigint().clone()),
            _ => None,
        };
```

vm/src/obj/objbyteinner.rs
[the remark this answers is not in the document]
jgirardet: do you mean something like this :
```rust

impl TryFromObject for PyByteInner {
        match_class!(obj.clone(),

        i @ PyBytes => Ok(PyByteInner{elements: i.get_value().to_vec()}),
        // j @ PyByteArray => Some(get_value_bytearray(&j.as_object()).to_vec()),
        // k @ PyMemoryView => Some(k.get_obj_value().unwrap()),
        _ => Err(vm.new_type_error(format!(
                            "a bytes-like object is required, not {}",
                            obj
                        ))))
    }
}
```

where would you use this `Either<BytesLike, PyIntRef>` ?

vm/src/obj/objbyteinner.rs
palaviv: I think you can use `into_option` of `OptionalArg` to make this code a little nicer but this looks good. This will move all the type checks to the `FromArgs`.

vm/src/obj/objbyteinner.rs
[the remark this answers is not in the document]
palaviv: What you wrote looks great.

> where would you use this Either<BytesLike, PyIntRef> ?

For example in:

- `ByteInnerFindOptions.sub`
- `contains.needle`

My point here is to move as much as possible of the type check to the shared code already exist.

palaviv:
Sorry for the late response @jgirardet. I will try and finish going over all the PR soon.

jgirardet:
np  @palaviv 
It took time but I finally understood how it works. thank for your help.
I added it :
```
impl<B: PyValue> TryFromObject for Either<PyByteInner, PyRef<B>>
and
impl TryFromObject for PyByteInner 
```
it's far better like it.

vm/src/obj/objbyteinner.rs
jgirardet: I let it  for now. 
```rust
let start = match self.start {
            OptionalArg::Present(Some(int)) => Some(int.as_bigint().clone()),
            _ => None,
        };
```
instead of 
```rust
let start = match self.start.into_option() {
            Some(Some(int)) => Some(int.as_bigint().clone()),
            _ => None,
        };
```

using into_option() isn't much shorter and I find it less explicit. what do you think ?

vm/src/obj/objbyteinner.rs
jgirardet: not relevant since I merged contains_bytes and contains_int

vm/src/obj/objbyteinner.rs
jgirardet: refactored

vm/src/obj/objbyteinner.rs
jgirardet: done

vm/src/obj/objbyteinner.rs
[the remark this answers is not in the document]
jgirardet: done

palaviv:
I think you can do `TryFromObject` for `PyByteInner` and use `Either` default one instead of creating one for `Either<PyByteInner, PyRef<B>>`.

vm/src/obj/objbyteinner.rs
palaviv: They both look great to me.

jgirardet:
PyByteInner doesn't implement PyValue so I can't use the default Either

vm/src/obj/objbyteinner.rs
windelbouwman: The rust clippy tool would probably suggest to use something like this:
```rust
let start = if let Some(Some(int)) = self.start.into_option() {
   Some(int.as_bigint().clone())
} else {
   None
};
```

windelbouwman:
I think this change is ready for merge now? Thank you for reviewing the change.

vm/src/obj/objbyteinner.rs
jgirardet: I use clippy and non suggestion about it

windelbouwman: It's especially handy when you have a match with only two options (basically an if-else case).

PR #848, opened 2019-04-19
Add PyJsValue to WASM

### LINKED ISSUE #808: How should JS values bind into Python?
This does not require knowledge of the WASM crate's internals, only a high level knowledge of Python and JS objects.

A couple questions about adding a `JsValue` class to WASM:

* When passing in a JS value into python, where should the line be between `JsValue` and builtin Python type? e.g. JS strings and numbers should be converted to their corresponding Python primitives, but what about arrays/lists? `TypedArrays`/`bytes`/`bytearray`s?
* Should functions be `JsValue`s too? Should JsValue have a `__call__` method?
* Should we have a `__getattr__` method on `JsValue`?
    * Should we throw an error when a `JsValue` doesn't have an attribute, or just return `None`?
* How should we transfer `kwargs` to JS functions? Right now it provides them as the `this` variable, but should we change that?
* How should errors transfer/convert between the two languages?

Not all of these questions are necessarily hard to answer, but it's good to have them on the table.

coolreader18:
It's not done yet, hence the draft PR, but the basic concept is working.

Discussion at/resolves #808 

cc @Ryex

### DIFF
--- a/wasm/lib/src/objjsvalue.rs
+++ b/wasm/lib/src/objjsvalue.rs
@@ -8,12 +8,13 @@ use rustpython_vm::VirtualMachine;
 
 use crate::convert;
 
-fn get_prop(value: &JsValue, name: &str, vm: &VirtualMachine) -> Option<PyObjectRef> {
+fn get_prop(value: JsValue, name: &str, vm: &VirtualMachine) -> Option<PyObjectRef> {
     let name: &JsString = &name.into();
-    if Reflect::has(value, name).expect("Reflect.has failed") {
-        Some(convert::js_to_py(
+    if Reflect::has(&value, name).expect("Reflect.has failed") {
+        Some(convert::js_to_py_with_this(
             vm,
-            Reflect::get(value, name).expect("Reflect.get failed"),
+            Reflect::get(&value, name).expect("Reflect.get failed"),
+            Some(value),
         ))
     } else {
         None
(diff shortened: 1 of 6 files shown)


coolreader18:
One thing is: How should we handle shared methods between `PyJsValue` and `PyJsFunction`? I was thinking that this *might* be a good place to bring back the idea of trait protocols, since this can't really be extracted out into an `inner` type? @OddCoincidence I think you have some opinions on abstractions like this, WDYT?

wasm/lib/src/objjsvalue.rs
```diff
@@ -0,0 +1,154 @@
...
+    } else {
+        None
+    }
+}
+
+fn set_prop(value: &JsValue, name: &str, val: PyObjectRef, vm: &VirtualMachine) {
+    Reflect::set(value, &name.into(), &convert::py_to_js(vm, val)).expect("Reflect failed");
+}
+
+#[pyclass(name = "JsValue")]
```
OddCoincidence: I mentioned [here](https://github.com/RustPython/RustPython/issues/808#issuecomment-485042199) also, but I think this should actually be called `JsObject`, with primitives always be converted to their Python correspondents.

Also, I think the name exposed to Python for this and other JS types should be `JS` (fully capitalized), as that's the convention set by [PEP 8](https://www.python.org/dev/peps/pep-0008/).

> Note: When using acronyms in CapWords, capitalize all the letters of the acronym. Thus HTTPServerError is better than HttpServerError.

PR #852, opened 2019-04-21
Add os.scandir

palaviv:
Currently support `name, path, is_dir, is_file`.
I will add `stat` and `symlink` support in separate PRs.

### DIFF
--- a/tests/snippets/stdlib_os.py
+++ b/tests/snippets/stdlib_os.py
@@ -85,12 +85,16 @@ with TestWithTempDir() as tmpdir:
 	names = set()
 	paths = set()
 	dirs = set()
+	files = set()
 	for dir_entry in os.scandir(tmpdir):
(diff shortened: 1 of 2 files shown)


tests/snippets/stdlib_os.py
```diff
@@ -73,3 +75,26 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 	assert os.read(fd, len(CONTENT2)) == CONTENT2
 	assert os.read(fd, len(CONTENT3)) == CONTENT3
 	os.close(fd)
+
+	fname2 = tmpdir + os.sep + FILE_NAME2
```
windelbouwman: We are in need of `os.path.join` :).

palaviv: It's all python code in path. I will look if it can be copied as is.

PR #855, opened 2019-04-21
Add check for nonlocal on top level.

windelbouwman:


tests/snippets/global_nonlocal.py
```diff
@@ -21,3 +22,37 @@ def y():
...
+
+try:
+    exec(src)
+except Exception as ex:
+    # print(ex)
+    pass
+else:
+    raise RuntimeError('Must raise')
+
+# Invalid syntax:
```
cthulahoops: Are we ready to test invalid nonlocal in a function too?

tests/snippets/global_nonlocal.py
```diff
@@ -21,3 +22,37 @@ def y():
...
 res = x()
 assert res == 3, str(res)
+
+# Invalid syntax:
+src = """
+b = 2
+global b
+"""
+
+try:
```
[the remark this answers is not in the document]
windelbouwman: Yes, this is a good idea!

tests/snippets/global_nonlocal.py
windelbouwman: I think so, what is the wrong case here?

cthulahoops: I was thinking simply:
```
def f():
    nonlocal x
    return x
```

windelbouwman: Yes, this is checked. This will raise a SyntaxError.

windelbouwman:
@cthulahoops I think this change is fine now?

PR #857, opened 2019-04-22
Add a store of Rust classes to PyContext

coolreader18:
I extracted this from #848, and I plan to add an `extends` capability to `#[pyclass]`, but that also depends on this feature and I didn't want to put all of that in #848. Also, I feel like this is a good compromise for how to store classes like `json.JSONDecodeError` - builtin types are still stored as struct fields, but you also have the capability to map an arbitrary Rust type to a class. It's certainly more performant than looking it up as a string key of a module dict!

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -1,7 +1,8 @@
 use std::any::Any;
+use std::any::TypeId;
 use std::cell::Cell;
 use std::cell::RefCell;
-use std::collections::HashMap;
+use std::collections::hash_map::{Entry, HashMap};
 use std::fmt;
 use std::marker::PhantomData;
 use std::mem;
@@ -162,6 +163,7 @@ pub struct PyContext {
     pub weakproxy_type: PyClassRef,
     pub object: PyClassRef,
     pub exceptions: exceptions::ExceptionZoo,
+    rust_classes: RefCell<HashMap<TypeId, PyClassRef>>,
 }
 
 pub fn create_type(name: &str, type_type: &PyClassRef, base: &PyClassRef) -> PyClassRef {
@@ -367,6 +369,7 @@ impl PyContext {
             weakproxy_type,
             type_type,
             exceptions,
+            rust_classes: RefCell::default(),
         };
         objtype::init(&context);
         objlist::init(&context);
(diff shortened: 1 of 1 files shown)


vm/src/pyobject.rs
```diff
@@ -407,6 +410,36 @@ impl PyContext {
...
+
+    #[inline]
+    pub fn add_class<T: PyClassImpl + 'static>(&self) -> PyClassRef {
+        let class = T::make_class(self);
+        self._add_class(TypeId::of::<T>(), class.clone());
+        class
+    }
+
+    pub fn _get_class(&self, typeid: &TypeId) -> Option<PyClassRef> {
+        self.rust_classes.borrow_mut().get(typeid).cloned()
```
windelbouwman: No need to use borrow_mut here, we are not mutating the mapping.

windelbouwman:
@coolreader18 thank's for your work on this. Getting a member from the module itself is not solved well at this point. I have mixed feelings about this change. On the one hand, it solves the issue of getting a rust type from the VM in a better way, rather then getting it via a get module hack. On the other hand, we still require the module if we want to retrieve a variable or function.

One other remark, with this route we store all rust classes of all modules on the pycontext, this potentially exposes all classes to all modules. I don't think this is the way to go.

I think is better to provide a mechanism such that each function in a module has access to its containing module. Maybe by extending the magic argument types with a `OwnModule` type like this:

```
#[pymethod(name="foo")
fn foo(vm: VirtualMachine, module: OwnModule, x: PyIntRef) {
  // use module here
  module.get_attr('')
}
```

Or maybe we can use a closure to fix this? Something like this:
```
fn make_module() {

SocketClass = py_class!();

#[pymethod(name="foo")
fn foo(vm: VirtualMachine, x: PyIntRef) -> SocketClass {
   SocketClass
}

}

```

In other words, I am not convinced this is the way to go.

coolreader18:
What do you think about struct modules then? i.e. a module would be defined as a struct with specifically defined fields as well as a string-keyed dict, stored somewhere as just a `dyn Any`, and you would call maybe `ctx.get_module::<OsModule>()` to get it. That way, you could only access it if you can get a reference to a type that it corresponds to (which, btw, is how this rust_classes implementation works), so if you make that module struct private, it's completely encapsulated.

windelbouwman:
Yeah, I think that is already better. It sort of duplicates the `sys.modules` dictionary, but in a more fail safe way. Still I feel we could do something with scopes at the module sourcecode file level.

coolreader18:
What you were describing with a `fn` declaration inside of the `make_module` function wouldn't work, because inline function declarations aren't closures; they can't capture any of the variables around them like `||{}` closures can, and defining every function in a module as a closure would be sort of painful.

windelbouwman:
Okay, in that case I prefer to keep the current system with the strings. This also works, although it is a bit weird. Lets wait for the opinions of other reviewers.

coolreader18:
I'm currently working on a struct modules implementation, which would eliminate the need for this. Should we close this for now?

windelbouwman:
Maybe leave this open to allow the others to have a look at this idea. What is the idea with the struct modules?

coolreader18:
A `RefCell<HashMap<TypeId, dyn Any>>` stored on PyContext, and `store_module<T>(&self, module: T)` and `get_module<T>(&self) -> T` methods. I think `T` would be required to be `Clone` so we could get it out of the `HashMap` .

coolreader18:
Or maybe `get_module` could accept a closure so someone could only clone what they need from the stored module.

PR #858, opened 2019-04-22
Fix README not to refer moved file

youknowone:
`cargo run tests/snippets/whats_left_to_implement.py` -> `./whats_left.sh`


coolreader18:
Thanks for catching this!

PR #859, opened 2019-04-22
Add built-in dict.pop

youknowone:


vm/src/obj/objdict.rs
```diff
@@ -228,6 +228,13 @@ impl PyDictRef {
         PyDictRef::merge(&self.entries, dict_obj, kwargs, vm)
     }
 
+    fn pop(self, key: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        match self.entries.borrow_mut().pop(vm, &key) {
```
windelbouwman: You may want to use `map_err` here. Furthermore, I think we do not require the map the result. Can't we do without this match statement and return whatever the `pop` function on entries returns?

vm/src/obj/objdict.rs
```diff
@@ -228,6 +228,13 @@ impl PyDictRef {
...
+            Ok(obj) => Ok(obj),
+            Err(_) => Err(vm.new_key_error(format!("{}", vm.to_pystr(&key)?))),
```
windelbouwman: I would move the creation of this key error to line 207 of dictdatatype.rs.

windelbouwman:
@youknowone thank you for your contribution. I left some remarks, but the change looks good to me! Let me know if you want to modify the changes or not, then we can merge this work.

youknowone:
Thanks for the nice review. I fixed it.

PR #860, opened 2019-04-22
Add os.stat

palaviv:
This is only a limited part of `os.stat`. The windows implementation is missing `st_ino`, `st_dev` and `st_nlink` as they are not yet implemented in `std::os::windows::fs::MetadataExt`.
I have tested the results against Cpython on my machine but I am not sure how to test it in the CI. The windows code is not tested yet so I will let the CI test it.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -277,6 +277,9 @@ struct StatResult {
     st_mode: u32,
     st_ino: u64,
     st_dev: u64,
+    st_nlink: u64,
+    st_uid: u32,
+    st_gid: u32,
 }
 
 impl PyValue for StatResult {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/os.rs
```diff
@@ -272,6 +272,98 @@ fn os_scandir(path: PyStringRef, vm: &VirtualMachine) -> PyResult {
...
+        }
+        .into_ref(vm)
+        .into_object()),
+        Err(s) => Err(vm.new_os_error(s.to_string())),
+    }
+}
+
+#[cfg(all(not(unix), not(windows)))]
+fn os_stat(path: PyStringRef, vm: &VirtualMachine) -> PyResult {
+    unimplemented!();
```
windelbouwman: This is a rust unimplemented. It might be good to use a python NotImplementedError here? I'm not sure which is right.

PR #861, opened 2019-04-22
Add __main__.py support

youknowone:
I tested this code by running `cargo run tests/snippets/dir_main` but no idea how to integrate it into current test system. Any idea?

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -131,6 +131,19 @@ fn run_script(vm: &VirtualMachine, script_file: &str) -> PyResult {
     }
 }
 
+#[test]
+fn test_run_script() {
+    let vm = VirtualMachine::new();
+
+    // test file run
+    let r = run_script(&vm, "tests/snippets/dir_main/__main__.py");
+    assert!(r.is_ok());
+
+    // test module run
+    let r = run_script(&vm, "tests/snippets/dir_main");
+    assert!(r.is_ok());
+}
+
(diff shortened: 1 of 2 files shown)


src/main.rs
```diff
@@ -97,12 +97,36 @@ fn run_script(vm: &VirtualMachine, script_file: &str) -> PyResult {
...
     // Parse an ast from it:
     let file_path = Path::new(script_file);
-    match util::read_file(file_path) {
-        Ok(source) => _run_string(vm, &source, file_path.to_str().unwrap().to_string()),
-        Err(err) => {
-            error!("Failed reading file: {:?}", err.kind());
-            std::process::exit(1);
+    if file_path.is_file() {
+        match util::read_file(file_path) {
+            Ok(source) => _run_string(vm, &source, file_path.to_str().unwrap().to_string()),
```
windelbouwman: Here you duplicated this line of code. Change the code such that first the filename is determined, and then this line is executed. Like this:
```
let file_path = if file_path.is_file() {
  file_path
} else if file_path.is_dir() {
  let file_path = file_path.join("__main__.py");
  if file_path.is_file() {
     file_path
  } else {
    error("No __main__.py found...")
  }
} else {
};

match util::read_file(&file_path) {
  Ok => run_string() 
   ...
};

windelbouwman:
No ideas :). I think it is probably fine not to test this. Maybe you could add a rust unittest?

PR #866, opened 2019-04-24
dict.popitem & dict.setdefault

cthulahoops:


tests/snippets/dict.py
```diff
@@ -174,3 +174,15 @@ def __eq__(self, other):
 
 with assertRaises(KeyError):
     x.pop("not here")
+
+x = {1: 'a'}
+assert (1, 'a') == x.popitem()
+with assertRaises(KeyError):
+    x.popitem()
```
coolreader18: We should check that `x == {}`

tests/snippets/dict.py
```diff
@@ -174,3 +174,15 @@ def __eq__(self, other):
...
+
+x = {'a': 4}
+assert 4 == x.setdefault('a', 0)
+assert 0 == x.setdefault('b', 0)
+assert x['b'] == 0
```
coolreader18: We should check `x['a'] == 4`

PR #869, opened 2019-04-24
Accept a tuple for the second argument of isinstance and issubclass

skinny121:


vm/src/builtins.rs
```diff
@@ -312,19 +313,42 @@ fn builtin_id(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
 
 // builtin_input
 
-fn builtin_isinstance(obj: PyObjectRef, typ: PyClassRef, vm: &VirtualMachine) -> PyResult<bool> {
-    vm.isinstance(&obj, &typ)
+fn builtin_isinstance(
+    obj: PyObjectRef,
+    typ: Either<PyClassRef, PyTupleRef>,
```
windelbouwman: Hmm, I'm not sure about the `Either` type. Why not use `PyObjectRef` and use something like this:
```rust
if let cls = to_PyClassRef(typ) {
   ...
} else if let tuple = to_tuple_ref(typ) {
   ...
} else {
   return typeerror('expected class or tuple of classes')
}
```
My problem is with the mapping of the either type into members A and B. This introduces a mapping.

vm/src/builtins.rs
```diff
@@ -312,19 +313,42 @@ fn builtin_id(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
-    Ok(vm.context().new_bool(issubclass))
+fn builtin_issubclass(
+    subclass: PyClassRef,
+    typ: Either<PyClassRef, PyTupleRef>,
+    vm: &VirtualMachine,
+) -> PyResult<bool> {
+    match typ {
+        Either::A(ref cls) => vm.issubclass(&subclass, cls),
+        Either::B(ref tuple) => {
+            for cls_obj in tuple.elements.borrow().iter() {
```
windelbouwman: Could we use the `any` function here?

vm/src/builtins.rs
skinny121: I changed it to use the `match_class` macro, which is equivalent to the above.

vm/src/builtins.rs
skinny121: If you are referring to the `builtin_any` function, then we would have to convert each result back into a PyObjectRef, so  `builtin_any` understands it. That seems kinda inefficiently and would 
 not really simplify the code.

If you are referring to the `any` function on a rust `Iterator`, then that doesn't handle fallible operations which we need as `isinstance`/`issubclass` can fail.

windelbouwman: I was referring to something like this:
```rust
tuple.elements.borrow.map(|clb_obj| obj_obj.isinstance?).any()
```
so more the rust `any` functionality. Just curiosity from my side. This code looks pretty clean.

skinny121: That won't work as `?` doesn't work within a closure like that, I wish it did work as it would make working with failable operations easier.

skinny121:
I don't know why the last build failed. It appears that something in the WASM test timed out. https://travis-ci.org/RustPython/RustPython/jobs/525513132

I have rebased onto the latest master to see if that will pass or not.

vm/src/builtins.rs
```diff
@@ -312,19 +313,37 @@ fn builtin_id(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
 
 // builtin_input
 
-fn builtin_isinstance(obj: PyObjectRef, typ: PyClassRef, vm: &VirtualMachine) -> PyResult<bool> {
-    vm.isinstance(&obj, &typ)
+fn type_test(
```
windelbouwman: Could we think of a clearer function name, like `single_or_tuple_must_be_true`? Not required for the rest of the change, just the function name which is a little vague.

PR #870, opened 2019-04-25
Dict fromkeys

cthulahoops:


vm/src/obj/objdict.rs
```diff
@@ -206,10 +222,7 @@ impl PyDictRef {
     ) -> PyResult {
         match self.entries.borrow().get(vm, &key)? {
             Some(value) => Ok(value),
-            None => match default {
-                OptionalArg::Present(value) => Ok(value),
-                OptionalArg::Missing => Ok(vm.ctx.none()),
-            },
+            None => Ok(default.unwrap_or_else(|| vm.ctx.none())),
```
coolreader18: You could add an `unwrap_or_none(self, vm: &VirtualMachine)` method for use cases like this.

windelbouwman:
Our new internal API is good! Nice change!

PR #872, opened 2019-04-25
Fix os_stat for MacOS

ypyf:


palaviv:
Nice work.
Are we supporting mac? We should probably add mac to the CI if we do.

PR #873, opened 2019-04-25
Add bytes.maketrans

youknowone:


jgirardet:
hi, 
there is a proposal for maketrans in #847. Please could you review it and tell me If I should remove it or change some things.
A bunch a bytes method are waiting review.
You can find not implemented  bytes method here : #736

youknowone:
Thanks. It looks a lot better. I will close this one

PR #874, opened 2019-04-25
Add a cfg for os.stat for android

coolreader18:


vm/src/stdlib/os.rs
```diff
@@ -378,7 +397,12 @@ fn os_stat(path: PyStringRef, vm: &VirtualMachine) -> PyResult {
     }
 }
 
-#[cfg(all(not(target_os = "linux"), not(target_os = "macos"), not(windows)))]
+#[cfg(not(any(
+    target_os = "linux",
+    target_os = "macos",
```
windelbouwman: Is there an option to use `#if` and `#else` kind of construct?

coolreader18: No, there's really only cfg for conditional compilation, which can only apply to individual items for hygiene's sake. We could in theory have multiple Rust modules that carry the target-specific parts of the `_os` module, but I don't think we have enough `cfg`s to warrant that yet.

PR #876, opened 2019-04-26
Fix json.loads

ypyf:
Prevent crash from call to json.loads. Use `vm.get_attribute` to get the attribute `json.JSONDecodeError` instead of `get_item`.


coolreader18:
Thanks for contributing!

PR #881, opened 2019-04-27
Add __builtins__ to globals()

youknowone:
This feature allows to expand `__builtins__` for program-global

### DIFF
--- a/benchmarks/bench.rs
+++ b/benchmarks/bench.rs
@@ -99,7 +99,7 @@ fn bench_rustpy_nbody(b: &mut test::Bencher) {
     };
 
     b.iter(|| {
-        let scope = vm.ctx.new_scope();
+        let scope = vm.new_scope_with_builtins();
         let res: PyResult = vm.run_code_obj(code.clone(), scope);
         assert!(res.is_ok());
     })
@@ -118,7 +118,7 @@ fn bench_rustpy_mandelbrot(b: &mut test::Bencher) {
     };
 
     b.iter(|| {
-        let scope = vm.ctx.new_scope();
+        let scope = vm.new_scope_with_builtins();
         let res: PyResult = vm.run_code_obj(code.clone(), scope);
         assert!(res.is_ok());
     })
--- a/tests/snippets/builtins_module.py
+++ b/tests/snippets/builtins_module.py
@@ -1,0 +1,16 @@
+from testutils import assertRaises
+
+assert '__builtins__' in globals()
+# assert type(__builtins__).__name__ == 'module'
+with assertRaises(AttributeError):
+    __builtins__.__builtins__
+
+__builtins__.x = 'new'
+assert x == 'new'
+
+# __builtins__ is deletable but names are alive
+del __builtins__
+with assertRaises(NameError):
+    __builtins__
+
+assert print
--- a/tests/snippets/builtins_module.py
+++ b/tests/snippets/builtins_module.py
@@ -8,6 +8,18 @@ with assertRaises(AttributeError):
 __builtins__.x = 'new'
 assert x == 'new'
 
+exec('assert "__builtins__" in globals()', dict())
+exec('assert __builtins__ == 7', {'__builtins__': 7})
+exec('assert not isinstance(__builtins__, dict)')
+exec('assert isinstance(__builtins__, dict)', {})
+
+namespace = {}
+exec('', namespace)
+assert namespace['__builtins__'] == __builtins__.__dict__
+
+# with assertRaises(NameError):
+#     exec('print(__builtins__)', {'__builtins__': {}})
+
 # __builtins__ is deletable but names are alive
 del __builtins__
 with assertRaises(NameError):
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -218,18 +218,27 @@ fn make_scope(
         }
         None => None,
     };
-
     let current_scope = vm.current_scope();
     let globals = match globals {
-        Some(dict) => dict.clone().downcast().unwrap(),
+        Some(dict) => {
+            let dict: PyDictRef = dict.clone().downcast().unwrap();
+            if !dict.contains_key("__builtins__", vm) {
+                let builtins_dict = vm.builtins.dict.as_ref().unwrap().as_object();
+                dict.set_item("__builtins__", builtins_dict.clone(), vm)
+                    .unwrap();
+            }
+            dict
+        }
         None => current_scope.globals.clone(),
     };
+
     let locals = match locals {
         Some(dict) => dict.clone().downcast().ok(),
         None => current_scope.get_only_locals(),
     };
 
-    Ok(Scope::new(locals, globals))
+    let scope = Scope::with_builtins(locals, globals, vm);
+    Ok(scope)
 }
 
 fn builtin_format(
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -98,6 +98,20 @@ impl Scope {
         Scope { locals, globals }
     }
 
+    pub fn with_builtins(
+        locals: Option<PyDictRef>,
+        globals: PyDictRef,
+        vm: &VirtualMachine,
+    ) -> Scope {
+        if !globals.contains_key("__builtins__", vm) {
+            globals
+                .clone()
+                .set_item("__builtins__", vm.builtins.clone(), vm)
+                .unwrap();
+        }
+        Scope::new(locals, globals)
+    }
+
     pub fn get_locals(&self) -> PyDictRef {
         match self.locals.iter().next() {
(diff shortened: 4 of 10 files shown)


vm/src/pyobject.rs
```diff
@@ -643,8 +643,13 @@ impl PyContext {
         objtype::new(self.type_type(), name, vec![base], PyAttributes::new()).unwrap()
     }
 
-    pub fn new_scope(&self) -> Scope {
-        Scope::new(None, self.new_dict())
+    pub fn new_scope(&self, vm: &VirtualMachine) -> Scope {
+        let scope = Scope::new(None, self.new_dict());
+
+        let globals = scope.globals.clone();
+        let _ = globals.set_item("__builtins__", vm.builtins.clone(), vm);
```
windelbouwman: I'm unsure whether this is the right place to set the item. Anyways, the function name should be changed, since the function now no longer creates a new scope. I would suggest to leave the original function in place, and create a second function which creates a scope (by calling the original function) and adds the builtin module. Then we can use this new function when we require a scope with builtins.

In other words: do we always require builtins in the scope?

windelbouwman: You can remove the `let _ =` part here. Underscore is the unused value. In case set_item returns an error, it should be handled by the `?` operator, or by unwrap, such that we panic when this fails. My preference is to handle the error with the `?` operator.

benchmarks/bench.rs
```diff
@@ -99,7 +99,7 @@ fn bench_rustpy_nbody(b: &mut test::Bencher) {
     };
 
     b.iter(|| {
-        let scope = vm.ctx.new_scope();
+        let scope = vm.ctx.new_scope(&vm);
```
cthulahoops: I don't like the method on context taking a ```VirtualMachine```. Given that this now needs vm then I think the method should move to the ```VirtualMachine```.

vm/src/pyobject.rs
cthulahoops: I think ```__builtins__``` get added in every case (except presumably during the creation of builtins itself). Even if we pass a dictionary to as globals, ```__builtins__``` gets added:

```
>>> exec('assert "__builtins__" in globals()', dict())
```

It doesn't get overwritten if already there:

```
exec('assert __builtins__ == 7', {'__builtins__': 7})
```

And the ```__builtins__``` provided is used for resolution of builtin names:

```
>>> exec('print()', {'__builtins__': {}})
  File "<string>", line 1, in <module>
NameError: name 'print' is not defined
```

And it doesn't use set_item to add it to the dictionary. (If you use a custom class for globals you don't get to intercept the addition of builtins.

There's also something odd that ```__builtins__``` is sometimes a module and sometimes a dict.

I'm not sure what the best implementation is. I lean towards not changing ```new_scope``` and doing the insert in the caller (because some callers will have to do special things to add it to globals that lack it.

youknowone:
I have no idea about exec examples.
But I created `Scope::init_builtins()` to do whatever required builtins stuff to the scope in the future.

wasm/lib/src/vm_class.rs
```diff
@@ -27,6 +27,7 @@ impl StoredVirtualMachine {
     fn new(id: String, inject_browser_module: bool) -> StoredVirtualMachine {
         let mut vm = VirtualMachine::new();
         let scope = vm.ctx.new_scope();
+        let _ = scope.init_builtins(&vm);
```
windelbouwman: Instead of ignoring the result, use `.unwrap()` instead. Then the error will be noticed earlier. The init_builtins function is to pass always, so this should not be a problem.

youknowone:
Fixed about `let _ =`

youknowone:
@windelbouwman @cthulahoops  I revised second commit. I am sorry that I made change after review.

Now, `__builtins__` is dict when exec() got dict variable; otherwise module. I think this is now correct rule for `globals()`.
Name resolving is a little bit different. For now, we lookup builtins module directly but not lookup `__builtins__`.

vm/src/builtins.rs
```diff
@@ -218,18 +218,30 @@ fn make_scope(
...
         None => None,
     };
-
     let current_scope = vm.current_scope();
     let globals = match globals {
-        Some(dict) => dict.clone().downcast().unwrap(),
+        Some(dict) => {
+            let dict: PyDictRef = dict.clone().downcast().unwrap();
+            if !dict.contains_key("__builtins__", vm) {
+                let builtins_dict = match &vm.builtins.dict {
```
windelbouwman: Please change this match statement into an `.unwrap()` call. This will panic as well:
```rust
let builtins_dict = &vm.builtins.dict.unwrap().as_object();
```

vm/src/frame.rs
```diff
@@ -90,7 +90,13 @@ impl fmt::Debug for Scope {
 }
 
 impl Scope {
-    pub fn new(locals: Option<PyDictRef>, globals: PyDictRef) -> Scope {
+    pub fn new(locals: Option<PyDictRef>, globals: PyDictRef, vm: &VirtualMachine) -> Scope {
+        if !globals.contains_key("__builtins__", vm) {
```
windelbouwman: I would not set the builtins in this new function, but in the calling sites of this new function, as @cthulahoops suggested.

vm/src/frame.rs
```diff
@@ -90,7 +90,13 @@ impl fmt::Debug for Scope {
 }
 
 impl Scope {
-    pub fn new(locals: Option<PyDictRef>, globals: PyDictRef) -> Scope {
+    pub fn new(locals: Option<PyDictRef>, globals: PyDictRef, vm: &VirtualMachine) -> Scope {
```
windelbouwman: This function can be left as is, because we will add builtins on other locations when required.

vm/src/vm.rs
```diff
@@ -250,6 +250,10 @@ impl VirtualMachine {
         syntax_error
     }
 
+    pub fn new_scope(&self) -> Scope {
+        Scope::new(None, self.ctx.new_dict(), self)
```
windelbouwman: Please remove the last `self` argument.

vm/src/frame.rs
youknowone: I thought @cthulahoops said `__builtins__` is added in every case.

If we still need to keep the `new` clean, I will add `Scope::with_builtins`. Is this ok?

vm/src/builtins.rs
youknowone: When I tried it, I got error like this:
```rust
error[E0507]: cannot move out of an `Rc`
   --> vm/src/builtins.rs:226:38
    |
226 |                 let builtins_dict = &vm.builtins.dict.unwrap().as_object();
    |                                      ^^^^^^^^^^^^^^^^ cannot move out of an `Rc`
```

youknowone:
I reverted `Scope::new()` as before and added a new method `Scope::with_builtins()`

vm/src/builtins.rs
windelbouwman: Maybe this works better:
```rust
let builtins_dict = vm.builtins.dict.as_ref().unwrap().as_object();
```

tests/snippets/builtins_module.py
```diff
@@ -0,0 +1,28 @@
...
+exec('assert "__builtins__" in globals()', dict())
+exec('assert __builtins__ == 7', {'__builtins__': 7})
+exec('assert not isinstance(__builtins__, dict)')
+exec('assert isinstance(__builtins__, dict)', {})
+
+namespace = {}
+exec('', namespace)
+assert namespace['__builtins__'] == __builtins__.__dict__
+
+# with assertRaises(NameError):
```
windelbouwman: What is the issue with this test? Can this test be enabled?

vm/src/vm.rs
```diff
@@ -250,6 +250,10 @@ impl VirtualMachine {
         syntax_error
     }
 
+    pub fn new_scope(&self) -> Scope {
```
windelbouwman: Please rename this function to reflect that it creates a scope and adds builtins. Something like: `new_scope_with_builtins`.

tests/snippets/builtins_module.py
youknowone: In RustPython, `exec('print()')` always succeed without builtins injection (original state).
Because this patch didn't touch anything about name resolving, this code still works - so the test is failed.

vm/src/vm.rs
youknowone: Fixed it

vm/src/builtins.rs
youknowone: Thanks! this code works well

windelbouwman:
@youknowone I will merge this now. Thank you for your patience with this pull request, I think it is now in good shape.

PR #882, opened 2019-04-27
Add type.__name__

youknowone:


vm/src/obj/objtype.rs
```diff
@@ -99,6 +99,10 @@ impl PyClassRef {
         issubclass(&subclass, &self)
     }
 
+    fn name(self, vm: &VirtualMachine) -> PyStringRef {
+        PyString::from(&self.name[..]).into_ref(vm)
+    }
```
coolreader18: You can just have this return `String`, it'll all work itself out,

coolreader18 (approved):

vm/src/obj/objtype.rs
youknowone: In that case, the function will return `self.name.clone()` rather than `self.name`. Is it ok? Actually I have no idea which way is generally preferred.

adrian17: Pretty sure that `PyString::from` also clones the string internally, so there shouldn't be any difference.

youknowone: Thanks all, I fixed it

PR #885, opened 2019-04-27
Add __int__, __float__, __trunc__, __rmod__ to float

youknowone:


vm/src/obj/objfloat.rs
```diff
@@ -163,12 +186,10 @@ impl PyFloat {
         let value = if objtype::isinstance(&arg, &vm.ctx.float_type()) {
             get_value(&arg)
         } else if objtype::isinstance(&arg, &vm.ctx.int_type()) {
-            match objint::get_value(&arg).to_f64() {
-                Some(f) => f,
-                None => {
-                    return Err(
-                        vm.new_overflow_error("int too large to convert to float".to_string())
-                    );
+            match objint::get_float_value(&arg, vm) {
```
windelbouwman: You might be able to replace this match statement with this line:
```rust
objint::get_float_value(&arg, vm)?
```

youknowone: Thanks

PR #887, opened 2019-04-28
objint::to_int to call __int__

youknowone:
Question: Removing PyInt and PyFloat matches or not?

~`float.__int__` in #885~

### DIFF
--- a/tests/snippets/ints.py
+++ b/tests/snippets/ints.py
@@ -74,3 +74,37 @@ with assertRaises(TypeError):
 with assertRaises(TypeError):
     # check that first parameter is truly positional only
     int(val_options=1)
+
+class A(object):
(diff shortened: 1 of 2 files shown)


windelbouwman:
I'm not sure whether to remove the PyFloat case. @cthulahoops what do you say? I suggest to remove the PyFloat case, since we already have float.__int__. We might as well use that function.

youknowone:
All test passes without PyFloat match. I think performance is the only reason to keep the line.

youknowone:
Removing the line will be better for now. Original code doesn't pass the following test code:
https://github.com/RustPython/RustPython/pull/887/files#diff-2bc1a6ee243d9eabcfcf2fa3bc66239dR110

PR #888, opened 2019-04-28
Add trunc, ceil, floor to math module

youknowone:
- ~`float.__trunc__` in #885~

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -8,7 +8,7 @@ use statrs::function::gamma::{gamma, ln_gamma};
 
 use crate::function::PyFuncArgs;
 use crate::obj::objfloat;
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/math.rs
```diff
@@ -172,6 +172,43 @@ fn math_lgamma(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     }
 }
 
+fn inner_rounding(func_name: &str, vm: &VirtualMachine, value: &PyObjectRef) -> PyResult {
```
windelbouwman: Rename this function so that it does not suggest to do rounding. Something like `call_magic_method` or `try_magic_method`.

youknowone: Thanks, fixed it

PR #889, opened 2019-04-28
Convert iterators to pyclass macros

adrian17:


vm/src/obj/objbytearray.rs
```diff
@@ -307,7 +304,8 @@ impl PyByteArrayIteratorRef {
         }
     }
 
-    fn iter(self, _vm: &VirtualMachine) -> Self {
-        self
+    #[pymethod(name = "__iter__")]
+    fn iter(zelf: PyRef<Self>, _vm: &VirtualMachine) -> PyRef<Self> {
+        zelf
```
lycantropos: I'm just a passer-by but why it is `zelf` and not `self`?

coolreader18: The target of the `impl` block is `PyByteArrayIterator`, and in this method we need a reference to a `PyRef`, e.g. a reference-counted smart pointer instead of just an `&self`. In a future upcoming version of Rust there will be a feature called `arbitrary_self_types`, which will let us specify the self parameter as `self: PyRef<Self>`, but because that's not stable yet, we use `zelf` for the time being.

windelbouwman:
@adrian17 I think this change is good. If you resolve the conflict, we can merge it.

PR #890, opened 2019-04-28
 Add exceptions stack to VirtualMachine

### LINKED ISSUE #867: Continue in except clause broken
The following, raises an exception `TypeError: Unsupported method: __next__`:
```python
for _ in [1, 2]:
    try:
        raise ArithmeticError()
    except ArithmeticError as e:
        continue
```
The cause of this is that the exception is put on the top of the stack, but continue expects the iterator to be on the top of the stack.

palaviv:
This resolves #867 and resolves [#837](https://github.com/RustPython/RustPython/pull/831#discussion_r275260542)

### DIFF
--- a/vm/src/compile.rs
+++ b/vm/src/compile.rs
@@ -24,7 +24,6 @@ struct Compiler {
     current_qualified_path: Option<String>,
     in_loop: bool,
     in_function_def: bool,
-    in_exc_handler: bool,
 }
 
 /// Compile a given sourcecode into a bytecode object.
@@ -86,7 +85,6 @@ impl Compiler {
             current_qualified_path: None,
             in_loop: false,
             in_function_def: false,
-            in_exc_handler: false,
         }
     }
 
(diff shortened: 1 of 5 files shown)


vm/src/frame.rs
```diff
@@ -860,6 +860,10 @@ impl Frame {
                 self.push_value(formatted);
                 Ok(None)
             }
+            bytecode::Instruction::PopException {} => {
+                assert!(!vm.pop_exception().is_none());
```
coolreader18: This can just be `assert!(vm.pop_exception().is_some())`.

vm/src/compile.rs
```diff
@@ -330,24 +328,15 @@ impl Compiler {
                     match cause {
                         Some(cause) => {
                             self.compile_expression(cause)?;
-                            self.emit(Instruction::Raise {
-                                argc: 2,
-                                in_exc: self.in_exc_handler,
-                            });
+                            self.emit(Instruction::Raise { argc: 2 });
```
windelbouwman: This looks cleaner already!

vm/src/compile.rs
```diff
@@ -597,8 +583,6 @@ impl Compiler {
         self.emit(Instruction::Jump { target: else_label });
 
         // except handlers:
-        let was_in_exc_handler = self.in_exc_handler;
-        self.in_exc_handler = true;
```
windelbouwman: Its really good to lose this state boolean!

vm/src/frame.rs
```diff
@@ -966,7 +970,8 @@ impl Frame {
         while let Some(block) = self.pop_block() {
             match block.typ {
                 BlockType::TryExcept { handler } => {
-                    self.push_value(exc);
+                    self.push_value(exc.clone());
```
windelbouwman: Its a bit weird to push the exception on the value stack as well as on the exception stack, but I guess we require this for proper value stack unwinding.

vm/src/frame.rs

PR #892, opened 2019-04-29
Add basic float.__round__ without ndigits support

youknowone:


vm/src/obj/objfloat.rs
```diff
@@ -41,6 +43,25 @@ impl From<f64> for PyFloat {
...
+            if value.is_infinite() {
+                Err(vm.new_overflow_error(
+                    "OverflowError: cannot convert float NaN to integer".to_string(),
+                ))
+            } else if value.is_nan() {
+                Err(vm
+                    .new_value_error("ValueError: cannot convert float NaN to integer".to_string()))
+            } else {
+                // unreachable unless BigInt has a bug
+                unsafe { std::hint::unreachable_unchecked() }
```
youknowone: Is this line ok? If not, please tell me an exception for this kind of places.

coolreader18: I think it should be a value error, in the case that the `f64` has a fractional component and can't be converted into an integer.

youknowone: Inside of `to_bigint`, it runs `trunc` before conversion. So I am pretty sure there is no unconvertible fractional component because the fractional simply does not exist.

Though it will not happen, it doesn't mean I am disagreeing to value error. I want to know you still want to make it value error.

coolreader18: I think it'd probably be safer just to go with a `panic!`, but if you're sure that there's no way for that to happen than feel free to do an unreachable hint.

youknowone: I changed it to `unreachable!()` macro instead of the hint one.

coolreader18 (approved):
Thanks for contributing!

PR #894, opened 2019-04-29
Add range.__eq__

youknowone:
I saparated a commit from #880 not to mix independent changes into a PR


coolreader18 (approved):
Looks good!

PR #895, opened 2019-04-29
extend_class for PyRange

youknowone:


vm/src/obj/objrange.rs
```diff
@@ -106,22 +107,9 @@ pub fn init(context: &PyContext) {
                      These are exactly the valid indices for a list of 4 elements.\n\
                      When step is given, it specifies the increment (or decrement).";
 
-    extend_class!(context, range_type, {
-        "__bool__" => context.new_rustfunc(PyRange::bool),
-        "__contains__" => context.new_rustfunc(PyRange::contains),
+    extend_class!(context, &context.range_type, {
         "__doc__" => context.new_str(range_doc.to_string()),
```
adrian17: The macros can also autogenerate `__doc__` from Rust doctrings now.

coolreader18: And you can put `__new__` in the pyimpl as well.

youknowone: Thanks, I will try

youknowone: Fixed them

coolreader18 (approved):

vm/src/obj/objrange.rs
```diff
@@ -355,8 +353,10 @@ impl PyValue for PyRangeIterator {
 
 type PyRangeIteratorRef = PyRef<PyRangeIterator>;
 
-impl PyRangeIteratorRef {
-    fn next(self, vm: &VirtualMachine) -> PyResult<BigInt> {
+#[pyimpl]
+impl PyRangeIterator {
+    #[pymethod(name = "__next__")]
```
coolreader18: What we usually do for methods that need a `PyRef` is keep it in the `pyimpl` and change the `self` parameter to `zelf: PyRef<Self>`. If you want to do that than feel free, otherwise this is good to merge.

youknowone: I am a little bit confused. It is not about `next` but about `iter`, right? like:

```rust
fn iter(zelf: PyRef<Self>, _vm: &VirtualMachine) -> Self
```

coolreader18: Correct, I guess I misclicked where I wanted to place the review. I meant that you can move the `iter` method into the same `impl` block as `next`.

youknowone: Thanks, I fixed it

PR #901, opened 2019-05-01
Add str.encode for utf-8

youknowone:


vm/src/obj/objbyteinner.rs
```diff
@@ -25,6 +25,18 @@ pub struct PyByteInner {
     pub elements: Vec<u8>,
 }
 
+pub fn vec_from_str(value: &str, encoding: &str, vm: &VirtualMachine) -> PyResult<Vec<u8>> {
+    let encoding = encoding.to_lowercase();
+    if encoding == "utf-8" || encoding == "u8" || encoding == "utf8" || encoding == "utf_8" {
```
jgirardet: Thinking bytes.decode I drafted a normalize_encoding like cpython. maybe it could be usable :
```rust
//same algorithm as cpython
pub fn normalize_encoding(encoding: &str) -> String {
    let mut res = String::new();
    let mut punct = false;

    for c in encoding.chars() {
        if c.is_alphanumeric() || c == '.' {
            if punct && !res.is_empty() {
                res.push('_')
            }
            res.push(c.to_ascii_lowercase());
            punct = false;
        } else {
            punct = true;
        }
    }
    res
}
```
you'll see on the same file some ideas to deal with errors.
https://github.com/jgirardet/RustPython/blob/1736d63932d5b7b4a6093eed62019255545888b2/vm/src/stdlib/_codecs.rs#L66

vm/src/obj/objbyteinner.rs
```diff
@@ -25,6 +25,18 @@ pub struct PyByteInner {
     pub elements: Vec<u8>,
 }
 
+pub fn vec_from_str(value: &str, encoding: &str, vm: &VirtualMachine) -> PyResult<Vec<u8>> {
```
jgirardet: maybe implementing `from_string` to pybyteinner ?

vm/src/obj/objstr.rs
```diff
@@ -837,6 +839,33 @@ impl PyString {
             false
         }
     }
+
+    #[pymethod]
+    fn encode(
+        &self,
+        encoding: OptionalArg<PyObjectRef>,
```
jgirardet: use PyStringRef to avoid type checking for "encoding" and "errors" wich will always be string

vm/src/obj/objstr.rs
```diff
@@ -837,6 +839,33 @@ impl PyString {
...
+                } else {
+                    Err(vm.new_type_error(format!(
+                        "encode() argument 1 must be str, not {}",
+                        v.class().name
+                    )))
+                }
+           },
+        )?;
+
+        let bytes = PyBytes::new(objbyteinner::vec_from_str(&self.value, &encoding, vm)?);
```
jgirardet: maybe `vm.ctx.new_bytes`  ?

vm/src/obj/objbyteinner.rs
youknowone: That was one of my early idea but then I found it is hard to turns `PyByteInner` into `PyBytes`. Do you have a good solution for that? For now, I think creating a vector is good enough for this kind of minimal charset transition API. Because it doesn't use any feature of `PyBytesInner` and easy to transit to it once it is required. Tell me if I missed something. I feel like I am very new to the str/bytes (actually for everything in this project, but especially).

vm/src/obj/objstr.rs
youknowone: Thanks, this is exactly what I needed

vm/src/obj/objstr.rs
youknowone: By trying it, I found I can't control the exception message with it. Do you have any idea about it?

vm/src/obj/objbyteinner.rs
youknowone: Both the function and the codec module looks really good. And I just recognized it is not a part of 847 😞 . Will you create a small PR with the part to adapt? Or do you prefer I import some required parts from your work? Just waiting for future and editing later will work too if it will be finally merged in future.

jgirardet: I 'm waiting #847 to be merge, and next there is another big PR (depending of 847) with other bytes method. I 'll wait everything is reviewed/merge( it's much work) and then I will propose something for bytes.decode.
For no just take what you need, i'll see after.

vm/src/obj/objbyteinner.rs
jgirardet: there is a try_from_object in #847 which creates a PyByteInner from a PyObjectRef(which could be bytes, byterray, memoryview). I think it make the API more consitent to use use a PyByteInner::from_string and the use PyByteInner.elements 
It's a personal opinion and I'm not in charge of this project.

vm/src/obj/objstr.rs
jgirardet: RustPython doesn't aim to have exactly the same error message as cpython. But it's better to let the type checking "automatic" in function type.

vm/src/obj/objbyteinner.rs
youknowone: I added this function in `objbyteinner.rs`, thanks

vm/src/obj/objbyteinner.rs
youknowone: added PyBytes::from_string and PyByteInner::from_string

vm/src/obj/objstr.rs
youknowone: I am not sure about error messages. If we can keep them the same or similar with small cost, why not? Is there priority about this? @windelbouwman

vm/src/obj/objbyteinner.rs
```diff
@@ -314,6 +323,20 @@ impl ByteInnerSplitlinesOptions {
 }
 
 impl PyByteInner {
+    pub fn from_string(value: &str, encoding: &str, vm: &VirtualMachine) -> PyResult<Self> {
+        let normalized = normalize_encoding(encoding);
+        if normalized == "utf_8" || normalized == "utf8" || normalized == "u8" {
```
jgirardet: `u8` is in encoding.aliases so I think it is redundant here.

jgirardet: looking at cpython code unicode.encode uses a getdefaultencoding function which return "utf-8" (which is a different behaviour than bytes.decode). You could do it add it  and at the same time add the getdefaultencoding to the sys module

youknowone: About `u8`, we don't have codec for now. So it is not redundant at this time. We can refactor it later once we have codec. About getdefaultencoding, it is good point, Thanks. I will go with it for next step.

vm/src/obj/objstr.rs
windelbouwman: The error messages to my opinion do not have to be similar to cpython. While I worked with rust, I find the error messages super useful in rust, maybe we could try something new in rustpython as well with regards to error messages. So in this case, I agree with @jgirardet to use the type checker with `PyStringRef` in this case.

windelbouwman:
@youknowone I guess this PR looks pretty good. Could you have a look at the conflicted file? Then I could merge this.

PR #903, opened 2019-05-01
Improve handling of lexical errors.

windelbouwman:


windelbouwman:
This change resolves some panics, and adds location information to lexical errors.

Issue #358, opened 2019-02-06
Expected methods on bool


veera83372:
The following should be implemented `src/vm/obj/objbool`

- [ ] bool . \_\_ceil__
- [ ] bool . \_\_dir__
- [ ] bool . \_\_floor__
- [ ] bool . \_\_getnewargs__
- [ ] bool . \_\_index__
- [ ] bool . \_\_init_subclass__
- [ ] bool . \_\_int__
- [ ] bool . \_\_radd__
- [ ] bool . \_\_rand__
- [ ] bool . \_\_rdivmod__
- [ ] bool . \_\_reduce__
- [ ] bool . \_\_reduce_ex__
- [ ] bool . \_\_rfloordiv__
- [ ] bool . \_\_rlshift__
- [ ] bool . \_\_rmod__
- [ ] bool . \_\_rmul__
- [ ] bool . \_\_ror__
- [ ] bool . \_\_round__
- [ ] bool . \_\_rpow__
- [ ] bool . \_\_rrshift__
- [ ] bool . \_\_rsub__
- [ ] bool . \_\_rtruediv__
- [ ] bool . \_\_rxor__
- [ ] bool . \_\_setattr__
- [ ] bool . \_\_sizeof__
- [ ] bool . \_\_subclasshook__
- [ ] bool . \_\_trunc__
- [ ] bool . denominator
- [ ] bool . from_bytes
- [ ] bool . imag
- [ ] bool . numerator
- [ ] bool . real
- [ ] bool . to_bytes

cthulahoops:
Most (I suspect All?) of these should not be implemented, bool subclasses int and uses the implementation from there.

List:

```
<method '__ceil__' of 'int' objects>
<method '__dir__' of 'object' objects>
<method '__floor__' of 'int' objects>
<method '__getnewargs__' of 'int' objects>
<slot wrapper '__index__' of 'int' objects>
<built-in method __init_subclass__ of type object at 0x9c99e0>
<slot wrapper '__int__' of 'int' objects>
<slot wrapper '__radd__' of 'int' objects>
<slot wrapper '__rand__' of 'bool' objects>
<slot wrapper '__rdivmod__' of 'int' objects>
<method '__reduce__' of 'object' objects>
<method '__reduce_ex__' of 'object' objects>
<slot wrapper '__rfloordiv__' of 'int' objects>
<slot wrapper '__rlshift__' of 'int' objects>
<slot wrapper '__rmod__' of 'int' objects>
[...]

PR #906, opened 2019-05-02
Implement bool.__ror__, __rand__ and __rxor__

sapir:
For issue #358


sapir:
Implemented incorrectly, fixing

PR #907, opened 2019-05-02
Preallocate output string of str.title() with required size

Jongy:


vm/src/obj/objstr.rs
```diff
@@ -425,7 +425,7 @@ impl PyString {
     /// uppercase character and the remaining characters are lowercase.
     #[pymethod]
     fn title(&self, _vm: &VirtualMachine) -> String {
-        let mut title = String::new();
+        let mut title = String::with_capacity(self.value.len());
```
adrian17: Could we also do the same in `__mul__`?

Jongy: Yeah, and in a few other spots as well. I have it ready but I intended to do the other batch in another PR.

Jongy: Ehh I'll just add another commit

Jongy: Actually, it gets a bit messy so I'll do that in another PR

PR #909, opened 2019-05-02
Add bool methods

sapir:
For #358

### DIFF
--- a/vm/src/obj/objbool.rs
+++ b/vm/src/obj/objbool.rs
@@ -43,8 +43,11 @@ The class bool is a subclass of the class int, and cannot be subclassed.";
         "__new__" => context.new_rustfunc(bool_new),
         "__repr__" => context.new_rustfunc(bool_repr),
         "__or__" => context.new_rustfunc(bool_or),
+        "__ror__" => context.new_rustfunc(bool_ror),
         "__and__" => context.new_rustfunc(bool_and),
+        "__rand__" => context.new_rustfunc(bool_rand),
         "__xor__" => context.new_rustfunc(bool_xor),
+        "__rxor__" => context.new_rustfunc(bool_rxor),
         "__doc__" => context.new_str(bool_doc.to_string())
     });
 }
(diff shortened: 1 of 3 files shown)


vm/src/obj/objbool.rs
```diff
@@ -71,6 +77,72 @@ fn bool_repr(vm: &VirtualMachine, args: PyFuncArgs) -> Result<PyObjectRef, PyObj
...
+
+fn do_bool_xor(vm: &VirtualMachine, lhs: &PyObjectRef, rhs: &PyObjectRef) -> PyResult {
+    if objtype::isinstance(lhs, &vm.ctx.bool_type())
+        && objtype::isinstance(rhs, &vm.ctx.bool_type())
+    {
+        let lhs = get_value(lhs);
+        let rhs = get_value(rhs);
+        (lhs ^ rhs).into_pyobject(vm)
+    } else {
+        Ok(lhs.payload::<PyInt>().unwrap().and(rhs.clone(), vm))
```
windelbouwman: This should probably use `.xor` instead of `.and`?

vm/src/obj/objbool.rs
```diff
@@ -71,6 +77,72 @@ fn bool_repr(vm: &VirtualMachine, args: PyFuncArgs) -> Result<PyObjectRef, PyObj
...
+    }
+}
+
+fn bool_xor(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
```
windelbouwman: This method seems duplicate with the one marked by `pymethod` in the class. There are at this point several styles to define object methods. The latest and the way to go is to use the `pymethod` markers on the class. If you mark a method as such, it will be available on the object if you use `PyBool::extend_class`. objint.rs is a good example file where this is used.

windelbouwman: Okay, ignore my comment :). I was not aware that also objint.rs was modified.

windelbouwman:
@sapir thank you for contributing to this project. This is appreciated! I left one remark in your change.

vm/src/obj/objbool.rs
sapir: oops

Issue #219, opened 2018-12-06
Add the "file" parameter to builtin "print()"


shinglyu:
This is useful is we want to re-map the STDOUT to WASM's console (or something else) in a clean way.

apatrushev:
Looks like this can be closed. Tested with following example on latest master:
```python
class A:
    def write(*args, **kwargs):
        print(args, kwargs)
print('a', file=A())
```
```bash
$ cargo run -- test.py
(<A object at 0x7fa77fcd5eb0>, 'a') {}
(<A object at 0x7fa77fcd5eb0>, '\n') {}
```
Llater found that test case was implemented in PR at `tests/snippets`.

PR #910, opened 2019-05-02
Print file parameter

sapir:
for #219

### DIFF
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -22,8 +22,8 @@ use crate::obj::objtype::{self, PyClass, PyClassRef};
 use crate::frame::Scope;
 use crate::function::{Args, KwArgs, OptionalArg, PyFuncArgs};
 use crate::pyobject::{
-    IdProtocol, ItemProtocol, PyIterable, PyObjectRef, PyResult, PyValue, TryFromObject,
-    TypeProtocol,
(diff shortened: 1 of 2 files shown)


vm/src/builtins.rs
```diff
@@ -559,32 +559,60 @@ pub struct PrintOptions {
...
-        } else {
-            write!(stdout_lock, " ").unwrap();
+    if let Some(file) = options.file {
+        let sep = match options.sep {
+            Some(sep) => sep.into_pyobject(vm).unwrap(),
+            _ => PyString::from(" ").into_pyobject(vm).unwrap(),
+        };
+
+        let mut first = true;
+        for object in objects {
```
windelbouwman: This for loop is now duplicated, including the `first` variable. I would change this into something like this:

Enter pseudo rust code:
```rust

// Create some sort of closure function:
let write_func = if let Some(file) = options.file {
  get_
} else {
  func_with_write_macro
}

// From this point, we have the availability over a function which does the proper printing, and can use this function at will.

// Now loop over the objects and print them using the closure function:
let mut first = true;
for object in objects {
  
  if first {
    first = false;
  } else {
    write_func(sep.clone())
  }
  write_func(object)
}
```

windelbouwman: Additional information: https://doc.rust-lang.org/book/ch13-01-closures.html

sapir: better now?

windelbouwman: Nice job!

PR #913, opened 2019-05-03
Preallocate strings in a smarter way + add str.rmul

Jongy:


vm/src/obj/objstr.rs
```diff
@@ -193,6 +198,11 @@ impl PyString {
         }
     }
 
+    #[pymethod(name = "__rmul__")]
+    fn rmul(&self, val: PyObjectRef, vm: &VirtualMachine) -> PyResult<String> {
+        self.mul(val, vm)
```
Jongy: Downside with this simple impl is that the raised `TypeError` shows the values swapped (As per the order of `__mul__`, not `__rmul__`).

On the other hand, the shown message (still?) doesn't match the one given by CPython, so we can be less pedantic about it.

Issue #659, opened 2019-03-11
[IDEA] Web imports for WASM


palaviv:
When using our WASM code currently doing import to anything that is not packaged in the build (builtins, stdlib) will not work. I suggest we will allow import in a way very similar to any other web resource.
For example when we will run the following from `http://test.org/py.html`
```js
rp.pyEval("import test")
```
The WASM import will do a HTTP GET request to `http://test.org/test.py` and import the content of the file.

We can also allow the users to add other paths to `sys.path` so the browser will try to look for the imported file there as well. I think that the paths should be treated like html paths:

- `path/to/import`: will be relative to the current file.
- `/path/to/import `: will be absolute to the server root.
- 'http://rustpypi.org/imports': will look in the `imports` path at domain `rustpypi.org`.

We should probably add the js VirtualMachine a `setSysPath` function like `setStdout`.

coolreader18:
Related to #228

palaviv:
> Related to #228 

I think that the solution at pyodide is a little more complex then what I suggest. I believe we should start from something simple like importing pure python code first.

coolreader18:
We could also use the PyPI API to fetch packages: https://warehouse.readthedocs.io/api-reference/json/; i.e. `import x` will look up x on PyPI and try to load it all somehow. The problem with both of these is that `import` is synchronous, as is `vm.exec()`, while fetch and related APIs are async.

palaviv:
I think that we should start with synchronous fetch maybe by using `XMLHttpRequest`. After that I think we can try and do something more complex. We can send an asynchronous request when we encounter the import but wait for the import to finish only on the first use.

coolreader18:
What if we just had a WasmVM method/browser module export that accepted a module URL and returned a promise for when it was loaded? That way you would either load it from JS or do ```python
def main():
    import module
browser.load_module("module", "https://module.url").then(main)

palaviv:
This will work great with new code someone writes specifically for RustPython WASM. I think that we would like a solution that will allow a user to use existing code in the regular way he used to.
I can't actually think about how to do what I suggest right now so I think starting with your suggestion and synchronous fetch would be best. We might want to add the pyEval a parameter with list of modules the user can ask to be loaded before the code is executed:
```js
rp.pyEval("import module; module.cool_func()", {"imports": {"module": "https://module.url"}})
```

coolreader18:
[...]

PR #914, opened 2019-05-03
Add browser.load_module

palaviv:
This is initial work on #659.
This allowed you to do:
```py
def f():
    import test
    print(test.func())

import browser
browser.load_module("test", "http://localhost:8080/test.py").then(f)
```
when test.py is:
```py
def func():
    return "testing"
```

### DIFF
--- a/wasm/lib/src/browser_module.rs
+++ b/wasm/lib/src/browser_module.rs
@@ -366,10 +366,7 @@ fn browser_load_module(module: PyStringRef, path: PyStringRef, vm: &VirtualMachi
                 .expect("val to be of type Response");
             response.text()
         })
-        .and_then(move |text_value: Promise| {
-            // Convert this other `Promise` into a rust `Future`.
-            JsFuture::from(text_value)
-        })
+        .and_then(JsFuture::from)
         .and_then(move |text| {
             let stored_vm = &weak_vm
                 .upgrade()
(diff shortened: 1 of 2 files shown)


wasm/lib/src/browser_module.rs
```diff
@@ -342,6 +347,45 @@ fn browser_prompt(
...
+    let future = JsFuture::from(request_prom)
+        .and_then(move |val| {
+            let response = val
+                .dyn_into::<web_sys::Response>()
+                .expect("val to be of type Response");
+            response.text()
+        })
+        .and_then(move |text_value: Promise| {
+            // Convert this other `Promise` into a rust `Future`.
+            JsFuture::from(text_value)
```
coolreader18: Instead of a closure, this can just be `JsFuture::from`

palaviv: I am not sure I understand the comment. I am using `JsFuture::from`

windelbouwman:
This API looks pretty cool to me! Only I am wondering whether we can implement this import logic in python itself using something like import lib?

palaviv:
I was thinking about that as well but I got into a bootstrap problem. How would we import the python code that will do the import in python?

wasm/lib/src/browser_module.rs
coolreader18: Instead of `move |text| {...}`, you can just put `JsFuture::from`, because the closure just forwards all it's arguments to `from`, so the closure isn't necessary.

palaviv: Done

palaviv:
CPython actually have a[ precompiled python code](https://github.com/python/cpython/blob/master/Python/importlib.h) that is made in the make process. But that is probably too complex for this stage.

coolreader18:
This looks good to go then!

PR #915, opened 2019-05-03
add  bytes split, expendtab, splitlines, zfil, replace

jgirardet:


vm/src/obj/objbyteinner.rs
```diff
@@ -783,3 +1002,483 @@ pub enum ByteInnerPosition {
...
+            index -= 1;
+        }
+    }
+    splitted.reverse();
+    splitted
+}
+
+#[cfg(test)]
+
+// needed for dev. Same as python tests in bytes.py. should it be kept ?
```
windelbouwman: I would prefer the python test snippets for now. Using unit tests makes refactoring difficult. I think we are at the final iteration, but you never know if we want to change the internal design again, and then these tests will probably be deleted.

PR #916, opened 2019-05-03
rewrite pybytearray with pybyteinner

jgirardet:
This is a rewrite of bytearray with pybyteinner.
Everything is the same as in #847 , except :
- some refactoring for eq, gt, ge, le, lt and add to DRY
- adapt clear, pop and append (which where already here)

Actual issues : 
- I'm struggling with  get_value and get_value_mut. For now i'm just using a public `inner` and call directly `borrow` or `borrow_mut` (see changes in io.rs). I need some help on it.
- the return type of `append`...

### DIFF
--- a/tests/snippets/bytearray.py
+++ b/tests/snippets/bytearray.py
@@ -1,53 +1,368 @@
-#__getitem__ not implemented yet
-#a = bytearray(b'abc')
-#assert a[0] == b'a'
-#assert a[1] == b'b'
+from testutils import assertRaises
 
-assert len(bytearray([1,2,3])) == 3
+# new
+assert bytearray([1, 2, 3])
+assert bytearray((1, 2, 3))
+assert bytearray(range(4))
+assert bytearray(3)
+assert b"bla"
+assert (
+    bytearray("bla", "utf8") == bytearray("bla", encoding="utf-8") == bytearray(b"bla")
+)
+with assertRaises(TypeError):
+    bytearray("bla")
+with assertRaises(TypeError):
+    bytearray("bla", encoding=b"jilj")
 
-assert bytearray(b'1a23').isalnum()
-assert not bytearray(b'1%a23').isalnum()
+assert bytearray(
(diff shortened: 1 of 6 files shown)


windelbouwman:
I had another idea about this. We face several types:
- `str`
- `bytes`
- `bytearray`

.. with a large set of similar methods:
- 'toupper`
- `swapcase`

.. and many more methods.

I was thinking about some sort of a trait on the inner object. Would it be possible to de-duplicate the logic between all three python types by putting the logic in a trait like this:

```rust
pub struct ByteInner {}
pub struct StrInner {}

pub trait StrMethods {
  // This must be specifically implemented for ByteInner and StrInner:
  fn Vec<char> get_chars();
  fn set_chars(chars: Vec<char>);

  // These are generic methods:
  fn to_upper() {
    let chars = self.get_chars();
   // perform magic to upper here :)
    self.set_chars(modified_chars);
  }
  // All other similar methods like swapcase and friends go here.
}

impl StrMethods for ByteInner {
  // implement proper get_chars and set_chars
}

impl StrMethods for StrInner {
  // implement proper get_chars and set_chars
}

```

Would this be possible?

adrian17:
While most bytes and bytearray methods can be merged, most bytes and str methods can't.

```
>>> 'ą'.upper()
'Ą'
>>> 'ą'.encode() == 'ą'.encode().upper()
True
```

jgirardet:
i'm not against the idea of the trait. but it adds much complexity due to many edge cases (like previous example).
But what would you think of some generic function for example like the split and split_reverse which would accept  [u8] or str ?

windelbouwman:
I'm for generic functions which can take [u8] or str arguments. We could use this for a whole series of string functions. Maybe we could do some templating with parameters like `u8` and `char`.

windelbouwman:
Something like this:
```rust
fn ljust<T>(src: &[T], width: i32, fill_char: T) -> [T] {
   // Implement ljust here...
}
```

jgirardet:
I quickly looked into it. Even it's intuitive, it's definitely as easy/smart than it seems.
Even  they seems very similar str and [u8] have many differences. Here as the mains issues :

- Most of the time, you want merge slice or str and you want a different return type (String or Vec<u8>).
- due to this you won't merge by the same way  : collect::<Vec<u8>>() for slice but clearly you prefer use format! for String.
- In function that increment an initial empty result, you will use  Vec.extend_from_slice where you will use push_str for String.
- some methods already exists for str (ex : find, trim, ends_wih, matches) and are well implemented so would we switch for other one ?
- edge cases ascii / unicode
- you iterate strings with chars() but iter() for vec.
- repeat is nightly for slice but stable for str

for all of this, I'm not sure we should try to do it. Even trying to reduce duplicate code,  it (IMHO) will add to much complexity.

windelbouwman:
I'm for readability at the expense of duplication. Maybe we will figure a way in the future to do this.

PR #917, opened 2019-05-04
Extend os.stat

palaviv:
- Add `DirEntry.stat`.
- Add Cpython `Lib/stat.py`
- Add tests for stat mode.


vm/src/stdlib/os.rs
```diff
@@ -380,12 +386,33 @@ fn os_stat(path: PyStringRef, vm: &VirtualMachine) -> PyResult {
     }
 }
 
+// Copied from CPython fileutils.c
+#[cfg(windows)]
+fn attributes_to_mode(attr: u32) -> u32 {
```
windelbouwman: Wow, this is a weird mapping :). Nice find!

PR #918, opened 2019-05-04
Except block

palaviv:
This solution come to fix the following problem:
```py
def g():
    try:
        1/0
    except ArithmeticError:
        return 5

try:
    g()
    raise NameError
except NameError as ex:
    assert ex.__context__ == None
```
When we exit an except with `return` or `yield` we need to pop the exception from the exception stack.

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -203,6 +203,7 @@ enum BlockType {
         end: bytecode::Label,
         context_manager: PyObjectRef,
     },
+    ExceptHandler {},
 }
 
 pub type FrameRef = PyRef<Frame>;
@@ -862,6 +863,12 @@ impl Frame {
             }
             bytecode::Instruction::PopException {} => {
                 assert!(vm.pop_exception().is_some());
+                let block = self.pop_block();
+                assert!(block.is_some());
+                match block.unwrap().typ {
+                    BlockType::ExceptHandler {} => (),
+                    _ => assert!(false),
+                };
                 Ok(None)
             }
         }
@@ -937,6 +944,9 @@ impl Frame {
                         }
                     }
                 }
+                BlockType::ExceptHandler { .. } => {
+                    vm.pop_exception();
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


vm/src/frame.rs
```diff
@@ -203,6 +203,7 @@ enum BlockType {
         end: bytecode::Label,
         context_manager: PyObjectRef,
     },
+    ExceptHandler {},
```
windelbouwman: Is the syntax without the `{}` valid? I would assume so, like this:
```rust
ExceptHandler,
```

vm/src/frame.rs
```diff
@@ -862,6 +863,12 @@ impl Frame {
             }
             bytecode::Instruction::PopException {} => {
                 assert!(vm.pop_exception().is_some());
+                let block = self.pop_block();
+                assert!(block.is_some());
+                match block.unwrap().typ {
+                    BlockType::ExceptHandler {} => (),
+                    _ => assert!(false),
+                };
```
windelbouwman: I would change this into:
```rust
let block = self.pop_block.unwrap(); // this asserts that the block is_some.
if let BlockType::ExceptHandler = block.typ {
  ()
} else {
  panic!("Block type must be ExceptHandler here.")
}
```

Reasons for this:
- `assert!` might be ignored during release builds, where panic is not. I'm not sure about this.
- A match with one match, and an else case is clearer with the `if let` syntax in my opinion.

vm/src/frame.rs
windelbouwman: Does CPython has a concept like this as well? Did you take the idea from there?

palaviv: No. I just wanted something that will allow me to know I finished the except handler.

vm/src/frame.rs
palaviv: Done :)

vm/src/frame.rs

vm/src/frame.rs
```diff
@@ -861,8 +862,13 @@ impl Frame {
                 Ok(None)
             }
             bytecode::Instruction::PopException {} => {
-                assert!(vm.pop_exception().is_some());
-                Ok(None)
+                let block = self.pop_block().unwrap(); // this asserts that the block is_some.
+                if let BlockType::ExceptHandler = block.typ {
+                    assert!(vm.pop_exception().is_some());
```
windelbouwman: You can use the same unwrap trick in this case.

PR #920, opened 2019-05-04
Repeat bytes

sapir:


tests/snippets/bytearray.py
```diff
@@ -71,3 +71,17 @@
 a.append(100)
 assert len(a) == 6
 assert a.pop() == 100
+
+
+# repeat by multiply
+a = bytearray(b'abcd')
```
youknowone: Would adding `assert a * -1 == bytearray(b'')` be good for future?

vm/src/obj/objbytes.rs
```diff
@@ -328,6 +329,29 @@ impl PyBytesRef {
             .ctx
             .new_bytes(self.inner.strip(chars, ByteInnerPosition::Right, vm)?))
     }
+
+    #[pymethod(name = "__mul__")]
+    fn repeat(self, n: isize, vm: &VirtualMachine) -> PyResult {
```
jgirardet: I should put as many logic as possible inside a `repeat` or a `mul` method in PyByteInner.
I think it should accept a `PyIntRef` instead of isize.

sapir: ok

tests/snippets/bytearray.py
sapir: added

PR #924, opened 2019-05-05
Builtin ord improvements

Jongy:


youknowone:
The unit tests for `ord()` is in `tests/snippets/builtin_ord.py`

Jongy:
> The test snippets for ord() are in tests/snippets/builtin_ord.py

The "0 characters" case is already tested, however the exception message itself is not tested (And that's what I fixed, to match CPython's behavior).
The question is, do we want to test such exact behavior regarding raised exceptions?

youknowone:
I thought the error message is changed and so now wrong messages are written in the code.
But it seems the messages are not tested as you said. I am sorry to bothering you.

I don't know much about the test policy. Maybe other people will explain about it.

windelbouwman:
There is no test policy set in stone. We have some snippets, hitting most of the code paths, but testing is not very strict.

PR #926, opened 2019-05-05
Fix number hashes for small numbers

youknowone:


tests/snippets/floats.py
```diff
@@ -98,6 +98,18 @@
...
+
+assert 1.0 == 1
+assert 1.0 == True
+assert 0.0 == 0
+assert 0.0 == False
+assert hash(1.0) == hash(1)
+assert hash(1.0) == hash(True)
+assert hash(0.0) == hash(0)
+assert hash(0.0) == hash(False)
+assert hash(1.0) != hash(1.0000000001)
```
adrian17: could we also get a "real world" test case, for example this?
```
assert 5.0 in {3, 4, 5}
```

youknowone: Added a few cases. The test added in dict.py also depends on hash

vm/src/obj/objint.rs
```diff
@@ -398,10 +398,12 @@ impl PyInt {
     }
 
     #[pymethod(name = "__hash__")]
-    fn hash(&self, _vm: &VirtualMachine) -> u64 {
```
windelbouwman: I'm unsure whether we must use `u64` or `usize` for hashes. What was your reason to switch from `u64` to `usize`?

youknowone: Just to unify to other __hash__ impls, because some of them returns u64 and the others returns usize for now.

Switching to `isize` is also an option because it is Py_ssize_t in CPython.

youknowone: #927 met a problem related to hash size. If we can't decide the type at this time, I suggest to introduce a new type alias `type HashValue = u64` (or usize) and keep watching how it goes.

youknowone: @windelbouwman I fixed type definition based on CPython. Will you check it again?

cthulahoops (approved):
314159? Weird.

This looks good to me. What a strange bit edge case.

youknowone:
rebased due to conflict

PR #927, opened 2019-05-06
Fix set() hash collision

youknowone:


cthulahoops:
We should reuse the implementation from dictdatatype rather than copying the code into set. The Dict struct has a type parameter so that we can use ```Dict<()>``` for sets.

vm/src/dictdatatype.rs
```diff
@@ -25,10 +34,10 @@ impl<T> Default for Dict<T> {
 }
 
 #[derive(Clone)]
-struct DictEntry<T> {
-    hash: usize,
-    key: PyObjectRef,
-    value: T,
+pub struct DictEntry<T> {
```
cthulahoops: I don't think we should be exposing the internals of the dictionary implementation here, this should be private. I think that we just need ```delete_or_insert``` and maybe ```delete_if_exists``` instead of  ```unchecked_*```

youknowone: Fixed

youknowone:
Any blocking issue here?

coolreader18:
I don't think so, merging.

PR #928, opened 2019-05-06
Add bytearray.count

talhof8:


tests/snippets/bytearray.py
```diff
@@ -71,3 +71,21 @@
 a.append(100)
 assert len(a) == 6
 assert a.pop() == 100
+
+a = bytearray(b'hey there, what is up?')
+assert(a.count(b'h', 3))
+assert(a.count(bytearray(b'h'), 3))
+assert(a.count(104, 3))
+try:
```
windelbouwman: For this type of contraption, we have this:
```python
from testutils import assertRaises

with assertRaises(TypeError):
    raise TypeError  here
```

windelbouwman:
@talhof8 thank you for contributing to this project!

tests/snippets/bytearray.py
talhof8: Seems about right, resolved :)

talhof8:
https://github.com/RustPython/RustPython/pull/916 added `bytearray.count()`, so this PR is no longer relevant.

PR #931, opened 2019-05-06
add bytes/byterray title

jgirardet:


vm/src/obj/objbyteinner.rs
```diff
@@ -975,6 +975,30 @@ impl PyByteInner {
 
         Ok(res)
     }
+
+    pub fn title(&self) -> Vec<u8> {
+        let mut res = vec![];
+        let mut spaced = true;
```
coolreader18: Should this be using the `unicode-casing` crate?

jgirardet: which one ?

coolreader18: It might not be necessary because it's all just `u8`s, but see the objstr implementation of `title()` for how unicode-casing is used.

jgirardet: got it. but ascii make it useless so think it's simpler to let it as it is.

PR #939, opened 2019-05-09
add from_bytes for int

stwen77:
Hi,
I want to take part in the project, while most object has someone to maintain, would you like to list some tasks for new comers?


windelbouwman:
Hi @stwen77 thank you for contributing to this project! Help is appreciated! I'll try to add some issues with good starter tasks. A good starter is to add methods to objects. It will get you started with the tooling and also is very useful!

PR #941, opened 2019-05-09
add from_bytes for int

stwen77:


vm/src/obj/objint.rs
```diff
@@ -489,6 +490,42 @@ impl PyInt {
...
+        for (key, value) in kwargs.into_iter() {
+            if key == "signed" {
+                signed = match_class!(value,
+
+                    b @ PyInt => !b.as_bigint().is_zero(),
+                    _ => false,
+                );
+            }
+        }
+        let x;
```
windelbouwman: I was not aware that this was possible in rust!

PR #942, opened 2019-05-09
Define `os_unix_stat_inner` only when required

Jongy:
Avoids warn when building on Windows.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -357,6 +357,7 @@ impl StatResultRef {
     }
 }
 
+#[cfg(unix)]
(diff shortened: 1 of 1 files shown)


coolreader18:
Is this necessary? There's no runtime cost to an unused macro, and this requires that we now have two `cfgs` to update and keep in sync if we were to add another `os_stat` implementation.

palaviv:
How about changing the config to `#[cfg(unix)]`?

coolreader18:
That sounds good too.

Jongy:
> How about changing the config to #[cfg(unix)]?

Cool, I actually took example from `os_stat`, that's why this funny list of `target_os`..

Issue #190, opened 2018-11-07
Add missing methods to `str` class.


windelbouwman:
The `str` class is still missing several methods. In CPython there exist a lot of handy methods you can use on strings. To fix this issue, head over to the file `vm/src/obj/objstr.rs` and add a method you found was missing.

This is a good first issue. If you completed your mission, please create another good starter issue for others to get their hands dirty with the sourcecode!

mcon:
Looks like `string_expected_methods` includes everything that `dir('foobar')` contains in cpython, and `objstr.rs` has an implementation of each of those - think this ticket can be closed.

ZapAnton:
I would disagree.

`isdecimal` is not explicitly implemented, and `''.isdecimal()` returns `True`, which is not correct
`maketrans` is also not implemented

mcon:
My mistake, you're correct - misinterpreted `whats_left_to_implement` - looks like the following all need doing:
```
string . __dir__
string . __doc__
string . __format__
string . __ge__
string . __getnewargs__
string . __init_subclass__
string . __iter__
string . __le__
string . __mod__
string . __reduce__
string . __reduce_ex__
string . __rmod__
string . __rmul__
string . __setattr__
string . __sizeof__
string . __subclasshook__
string . encode
string . format_map
string . isdecimal
string . isprintable
string . maketrans
string . translate
```

epellis:
Looks like the following (maybe more) have already been implemented:
```
string . isdecimal
string . __ge__
string . __doc__
string . __format__
```

yanganto:
implement str.isprintable[#1024](https://github.com/RustPython/RustPython/pull/1024)

longwusha:
@windelbouwman  What other work remains waiting for completion and implementation under this question? I am happy to contribute my spare time to this issue.

windelbouwman:
@longwusha did you run the what is left script to check missing methods on the str type?

longwusha:
@windelbouwman 

When I run the what is left script, I got these missing methods on the str type:

    str.__getnewargs__
    str.__init_subclass__
    str.__iter__
    str.__rmod__
    str.__sizeof__
    str.__subclasshook__
    str.format_map

I am not sure is any other people are working on these.

longwusha:
@windelbouwman 

As we can see, some one was trying to implement str.__iter__ in #593 . But it seems that he has no interest or time to continue this work, so I want to start to implement other missing methods on the str type based on his work.

[...]

PR #943, opened 2019-05-09
[WIP] implement str.translate and str.maketrans

Furyzer0:
Currently `str.translate` is working. This is my first contribution to this project so I would like to know if there is something wrong in my code.
Ref: #190

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -1201,26 +1201,28 @@ mod tests {
     }
 
     #[test]
-    fn str_translate() {
+    fn str_maketrans_and_translate() {
         let vm = VirtualMachine::new();
 
         let table = vm.context().new_dict();
         table
-            .set_item(97, vm.new_str("🎅".to_owned()), &vm)
+            .set_item("a", vm.new_str("🎅".to_owned()), &vm)
             .unwrap();
-        table.set_item(98, vm.get_none(), &vm).unwrap();
+        table.set_item("b", vm.get_none(), &vm).unwrap();
         table
-            .set_item(99, vm.new_str("xda".to_owned()), &vm)
+            .set_item("c", vm.new_str("xda".to_owned()), &vm)
             .unwrap();
+        let translated = PyString::maketrans(
+            table.into_object(),
+            OptionalArg::Missing,
+            OptionalArg::Missing,
+            &vm,
+        )
+        .unwrap();
         let text = PyString::from("abc");
-        let translated = text.translate(table.into_object(), &vm).unwrap();
+        let translated = text.translate(translated, &vm).unwrap();
         assert_eq!(translated, "🎅xda".to_owned());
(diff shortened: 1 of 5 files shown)


vm/src/obj/objstr.rs
```diff
@@ -1104,4 +1139,34 @@ mod tests {
             assert!(!PyString::from(s).istitle(&vm));
         }
     }
+
+    #[test]
+    fn str_translate() {
```
youknowone: We also add python snippets under `tests/snippets`

youknowone:
Instead of manual formatting, run `cargo fmt --all`. The broken test is that one.

vm/src/obj/objstr.rs
```diff
@@ -829,6 +830,40 @@ impl PyString {
...
     }
+
+    // https://docs.python.org/3/library/stdtypes.html#str.translate
+    // Make it work for only subscribtable types
+    #[pymethod]
+    fn translate(&self, table: PyObjectRef, vm: &VirtualMachine) -> PyResult<String> {
+        let mut result = String::new();
+        for c in self.value.chars() {
+            let args = PyFuncArgs::new(vec![vm.new_int(c as i32)], vec![]);
+            match vm.call_method(&table, "__getitem__", args) {
```
adrian17: `match table.get_item(c as i32, vm) {`

Furyzer0: This looks way clearer!

Furyzer0:
> Instead of manual formatting, run `cargo fmt --all`. The broken test is that one.

I was already using `cargo fmt`.

coolreader18:
Are you using the latest rust toolchain?

vm/src/obj/objstr.rs
Furyzer0: > Are you using the latest rust toolchain?

You are right my default toolchain is stable which has different `rustfmt` than  `nightly`. Now formatting with `cargo +nightly fmt` and it seems all formatting problemsa are fixed now!

whats_left.sh
```diff
@@ -8,4 +8,4 @@ python3 not_impl_gen.py
 
 cd ..
 
-cargo run -- tests/snippets/whats_left_to_implement.py
+cargo +nightly run -- tests/snippets/whats_left_to_implement.py
```
windelbouwman: Please leave this file as it is. Other people also depend upon this one.

windelbouwman:
@Furyzer0 thank you for contributing to this project!

whats_left.sh
Furyzer0: Ok I wasn't intented to push this change but I forgot that I changed it. I will revert it back with my next commit.

Furyzer0:
It seems both functions work and all tests have passed. Tomorrow I will add the unit test for `str.maketrans` in the Rust module then it can be merged if there is no problem with the code.

PR #948, opened 2019-05-10
Itertools

Jongy:


adrian17:
(For simplicity, couldn't we use the `itertools` implementation from ouroboros? It'd probably have a higher compatibility than we have now too.)

vm/src/stdlib/itertools.rs
```diff
@@ -0,0 +1,73 @@
...
+    fn iter(zelf: PyRef<Self>, _vm: &VirtualMachine) -> PyRef<Self> {
+        zelf
+    }
+}
+
+pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
+    let ctx = &vm.ctx;
+
+    let count = ctx.new_class("count", ctx.object());
+    PyItertoolsCount::extend_class(ctx, &count);
```
coolreader18: This can be `PyItertoolsCount::make_class(ctx)` if you put `#[pyclass(name = "count")]` on the struct declaration.

windelbouwman:
@adrian17 this was also my first idea, but it seems a lot of modules are implemented in C in cpython, so we could do the same with rustpython.

vm/src/stdlib/itertools.rs
Jongy: I'll make use of this trick in the following PRs, thanks :)

PR #951, opened 2019-05-10
Fix str.isidentifier() and add tests snippets

youknowone:


vm/src/obj/objstr.rs
```diff
@@ -834,7 +834,8 @@ impl PyString {
     fn isidentifier(&self, _vm: &VirtualMachine) -> bool {
         let value = &self.value;
         // a string is not an identifier if it has whitespace or starts with a number
-        if !value.chars().any(|c| c.is_ascii_whitespace())
+        if !value.is_empty()
```
windelbouwman: Should we use the XID unicode package here? 

See also:
- https://github.com/RustPython/RustPython/blob/master/parser/src/lexer.rs#L591
- file:///usr/share/doc/python/html/reference/lexical_analysis.html#identifiers
- file:///usr/share/doc/python/html/library/stdtypes.html?highlight=isidentifier#str.isidentifier

windelbouwman: I think we could do something like this:
```rust
return chars.first().is_xid_start() && chars.skip(1).all(|c| c.is_xid_continuation());
```

youknowone: Fixed with unicode-xid.

windelbouwman: Nice job!

PR #957, opened 2019-05-10
Binascii

Jongy:


Jongy:
After finishing this I found about `PyByteInner.hex`, `PyByteInner.fromhex` methods, I guess we should merge these impls? I'm not too fluent with Rust to be able to tell which of them is more efficient, though...

sanxiyn:
You can't name the test containing `import binascii` as `binascii.py`, due to the way Python import works. Consider renaming to `stdlib_binascii.py` as done for other similar cases.

Jongy:
> You can't name the test containing import binascii as binascii.py, due to the way Python import works. Consider renaming to stdlib_binascii.py as done for other similar cases.

Actually it does work (As it also does in https://github.com/RustPython/RustPython/pull/948), I guess because the imports for builtin modules are attempted first?

Anyway I agree it's a bad practice and if there's a convention to use `stdlib_*.py`, so I'll use.

PR #960, opened 2019-05-11
Add sys.getfilesystemencoding, sys.getfilesystemencodeerrors

youknowone:
Related features: `os.fsencode()` `os.fsdecode()` (with this PR + #901)


windelbouwman:
@youknowone this change looks okay to me I guess, do we really need this? Is this cpython specific logic? In any case, if you could resolve the conflict, I could merge this work.

youknowone:
Fixed. The behavior can be changed but the functions are defined in pep-529
https://www.python.org/dev/peps/pep-0529/

Issue #958, opened 2019-05-11
Implement import by using importlib


skinny121:
The standard library for Python contains importlib, which contains a pure* Python implementation of the import machinery. This is lucky as we then don't have to implement it all in Rust. But we have to bootstrap it and implement all of the functionality it uses from builtin modules.

This is blocked by the following:
- [ ] compile should accept bytes as well as strings
- [x] _os.fspath - https://docs.python.org/3/library/os.html#os.fspath
- [x] sys.implementation.cache_tag  - This just needs to exist we don't support caching bytecode at the moment
- [ ] sys.pycache_prefix  - same as above
- [x] sys.flags
- [x] str.endswith needs to support tuple argument
- [ ] _io.FileIO \_\_enter\_\_ and \_\_exit\_\_ methods
- [x] sys.platform
- [x] os.getcwd
- [x] os.stat st_mtime attribute
- [ ] _warnings module
- possibly more

\* Uses built-in module _imp to actually create module objects.

palaviv:
An initial implementation can be found in #1018.

palaviv:
Another thing that would help this is supporting `{!r}` in string format.

palaviv:
This can be closed not that #1018 was merged

apatrushev:
`level` and `globals` argument is not provided and relative import is not working.
As example `import unittest` is broken at:
```python
from .result import TestResult
```

apatrushev:
Additional investigation - CPython does level magic in compile:
```bash
python -c "import dis; dis.dis(compile('from .second import value', '', 'exec'))"
```
```
  1           0 LOAD_CONST               0 (1)
              2 LOAD_CONST               1 (('value',))
              4 IMPORT_NAME              0 (second)
              6 IMPORT_FROM              1 (value)
              8 STORE_NAME               1 (value)
[...skipped...]
```
(note second dot)
```bash
python -c "import dis; dis.dis(compile('from ..second import value', '', 'exec'))"
```
```
  1           0 LOAD_CONST               0 (2)
              2 LOAD_CONST               1 (('value',))
              4 IMPORT_NAME              0 (second)
              6 IMPORT_FROM              1 (value)
              8 STORE_NAME               1 (value)
[...skipped...]
```

palaviv:
@apatrushev please look at the implementation of the relative import at #1050.

apatrushev:
> @apatrushev please look at the implementation of the relative import at #1050.

[...]

PR #962, opened 2019-05-11
Add st_{a,m,c}time to os.stat

skinny121:
For #958.

### DIFF
--- a/tests/snippets/stdlib_os.py
+++ b/tests/snippets/stdlib_os.py
@@ -143,6 +143,13 @@ with TestWithTempDir() as tmpdir:
 	print(stat_res.st_gid)
 	print(stat_res.st_size)
 	assert stat_res.st_size == len(CONTENT2) + len(CONTENT3)
+	print(stat_res.st_atime)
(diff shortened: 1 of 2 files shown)


tests/snippets/stdlib_os.py
```diff
@@ -143,6 +143,13 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 	print(stat_res.st_gid)
 	print(stat_res.st_size)
 	assert stat_res.st_size == len(CONTENT2) + len(CONTENT3)
+	print(stat_res.st_atime)
+	print(stat_res.st_ctime)
+	print(stat_res.st_mtime)
+	# test that it all of these times are greater than the 10 May 2019, when this test was written
+	assert stat_res.st_atime > 1557500000
```
palaviv: I think you should check here that `test1` file is older then `test2` file.
I would also check that after changing the file the modification time is increased.

skinny121:
I also implemented os.fsync to assist with testing, the updating of the file time attributes.

PR #964, opened 2019-05-11
Add os.supports_* placeholders

youknowone:


vm/src/stdlib/os.rs
```diff
@@ -512,33 +577,128 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
          "st_size" => ctx.new_property(StatResultRef::st_size),
     });
 
-    py_module!(vm, "_os", {
-        "open" => ctx.new_rustfunc(os_open),
+    let fd_funcs: Vec<(&str, PyObjectRef, Option<bool>, Option<bool>, Option<bool>)> = vec![
```
palaviv: Why is the `support_*` an `Option`? This can be either `true` of `false`. You convert the `None` to `false` later

tests/snippets/stdlib_os.py
```diff
@@ -161,3 +161,8 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
 	assert os.path.basename(fname) == FILE_NAME
 	assert os.path.dirname(fname) == tmpdir
+
+# supports
+assert isinstance(os.supports_fd, set)
+assert isinstance(os.supports_dir_fd, set)
+assert isinstance(os.supports_follow_symlinks, set)
```
palaviv: Can you test that `os.stat` (or any other compatible function) is in `os.supports_*`?

vm/src/stdlib/os.rs
```diff
@@ -512,33 +577,128 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
...
+        // name, func_obj, supports_fd, supports_dir_fd, supports_follow_symlinks
+        ("open", ctx.new_rustfunc(os_open), None, Some(false), None),
+        // access Some Some None
+        // chdir Some, None None
+        // chflags Some, None Some
+        // chmod Some Some Some
+        // chown Some Some Some
+        // chroot Some None None
+        (
```
palaviv: This is not very readable in the current way. It is hard to remember what parameter is which. How about changing this to be a vector of:
```rust
struct FdFunc {
   name: String,
   func_obj: PyObjectRef,
   supports_fd: boolean,
   supports_dir_fd: boolean,
   supports_follow_symlinks: boolean,
}
```

vm/src/stdlib/os.rs
```diff
@@ -512,33 +577,128 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
...
+        // truncate Some None None
+        (
+            "unlink",
+            ctx.new_rustfunc(os_remove),
+            Some(false),
+            Some(false),
+            None,
+        ),
+        // utime Some Some Some
+        // walk Some None
```
palaviv: Some of this function like `walk` are implemented in pure python. Maybe consider removing the placeholders?

vm/src/stdlib/os.rs
```diff
@@ -451,13 +499,25 @@ fn os_stat(path: PyStringRef, vm: &VirtualMachine) -> PyResult {
...
 #[cfg(unix)]
-fn os_symlink(src: PyStringRef, dst: PyStringRef, vm: &VirtualMachine) -> PyResult<()> {
+fn os_symlink(
+    src: PyStringRef,
+    dst: PyStringRef,
+    dir_fd: DirFd,
+    vm: &VirtualMachine,
+) -> PyResult<()> {
     use std::os::unix::fs as unix_fs;
+    let src = make_path(vm, src, &dir_fd);
```
palaviv: The `dir_fd` in symlink is only relevant for the `dst`. The underlying implementation use [`symlinkat`](https://linux.die.net/man/2/symlinkat).

vm/src/stdlib/os.rs
youknowone: It is a kind of documentation describing it is varies by platform(Some) or always not supported(None)

tests/snippets/stdlib_os.py
youknowone: Nothing is insdie of the set for now. It is just placeholders. See fd_funcs which only includes `None` and `Some(false)`

palaviv: Cool. Did not see that.

vm/src/stdlib/os.rs
palaviv: I am not sure I understand...
Let's take `os.symlink` for example. on linux system `dir_fd` is supported and on windows it is not. In that case we will have `Some(true)` on linux and `Some(false)` on windows. While `follow_symlinks` will always be `None`. Is there any difference between `Some(false)` and `None`? Would there be a different error message somewhere I am missing?

youknowone: There is no difference between `None` and `Some(false)` for implementations. As you pointed out, `None` marks `os.symlink` doesn't have any option about `follow_symlinks`. I think it is a little bit confusing because it is just `Some(false)` for now.
Once we get actual values for each platforms, the `None` part always will be `None`. Because Python spec doesn't have any value about it. Otherwise, the `Some` part will be something like `Some(platform_supports_os_symlink)` because Python spec describes the value must be `true` or `false` by platforms.

palaviv: Is `platform_supports_os_symlink` a function in your example or a boolean value?

youknowone: It is a value. (or result of a function) I think the values are platform-specific constants defined in compile time.

vm/src/stdlib/os.rs
youknowone: I fixed this and other comments, thanks!

vm/src/stdlib/os.rs
palaviv: My opinion is that this `Option` is not needed. I think using `platform_supports_os_symlink` or `false` would be more readable.

PR #965, opened 2019-05-11
Itertools takewhile

Jongy:
After https://github.com/RustPython/RustPython/pull/949

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -5,8 +5,9 @@ use std::ops::{AddAssign, SubAssign};
 use num_bigint::BigInt;
 
 use crate::function::OptionalArg;
+use crate::obj::objbool;
 use crate::obj::objint::{PyInt, PyIntRef};
-use crate::obj::objiter::new_stop_iteration;
+use crate::obj::objiter::{call_next, get_iter, new_stop_iteration};
 use crate::obj::objtype::PyClassRef;
 use crate::pyobject::{PyClassImpl, PyObjectRef, PyRef, PyResult, PyValue};
 use crate::vm::VirtualMachine;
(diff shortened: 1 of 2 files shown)


Jongy:
Seems like this needs to be rebased...

vm/src/stdlib/itertools.rs
```diff
@@ -0,0 +1,212 @@
...
+        .into_object())
+    }
+
+    #[pymethod(name = "__next__")]
+    fn next(&self, vm: &VirtualMachine) -> PyResult {
+        if *self.stop_flag.borrow() {
+            return Err(new_stop_iteration(vm));
+        }
+
+        let obj = call_next(vm, &self.iterable);
```
windelbouwman: Please use the `?` operator here, like this:
```rust
let obj = call_next(vm, &self.iterable)?;
```
The match statement can be replaced by the `?` suffix.

vm/src/stdlib/itertools.rs
```diff
@@ -0,0 +1,212 @@
...
+            Ok(verdict) => match objbool::boolval(vm, verdict) {
+                Ok(true) => Ok(obj),
+                Ok(false) => {
+                    *self.stop_flag.borrow_mut() = true;
+                    Err(new_stop_iteration(vm))
+                }
+                Err(err) => Err(err),
+            },
+            Err(err) => Err(err),
+        }
```
windelbouwman: Please replace the above by this:
```rust
let verdict = vm.invoke(self.predicate.clone(), vec![obj.clone()])?;
let verdict = objbool::boolval(vm, verdict)?;
if verdict {
  Ok(obj)
} else {
                    *self.stop_flag.borrow_mut() = true;
                    Err(new_stop_iteration(vm))
}

Jongy:
> Please check my comments about the usage of the ? operator for error handlings. It can be applied at multiple places in your change and makes the code much more compact.

Hmm, cool. Didn't know of this one. I'm learning Rust on the way :)

Thanks for the instructive comments.

PR #966, opened 2019-05-11
Move os.fspath to _os

palaviv:
For #958.


Lib/os.py
```diff
@@ -133,34 +133,3 @@ def getenv(key, default=None):
     The optional second argument can specify an alternate default.
     key, default and the result are str."""
     return environ.get(key, default)
-
-
-def fspath(path):
```
youknowone: Will keeping this funciton as `_fspath` just as cpython help to test new os_fspath works correctly?
Probably we can test 2 functions return same value.

palaviv: We are testing against CPython function

coolreader18:
@palaviv I tried to merge master into your branch and I messed it up; would you mind adding `use crate::obj::objtype;` to the top of `os.rs`? I'll merge then.

PR #968, opened 2019-05-12
Add walrus operator to lexer and try to make lexer code cleaner.

windelbouwman:


cthulahoops:
The walrus operator is 3.8 only? Are we targetting that? I tend to feel like we should target 3.7 rather chasing the latest shiny version. (I may be biased because the walrus operator feels like a mistake for the language.)

The lexer refactor is great. It would be nice to split the refactor and feature into separate commits.

coolreader18:
Should we merge this now? @cthulahoops would you be good with that? It's been 20 days since this was opened and we should merge before there are conflicts.

windelbouwman:
I will remove the walrus operator from this change to split the refactor and the walrus operator, so I would rather wait for this.

windelbouwman:
@cthulahoops could you re-review?

Mr-Andersen:
Excuse me, I don't understand, why Python actually needs "walrus operator". Why don't they implement assignment as binary operator?

cthulahoops:
> Excuse me, I don't understand, why Python actually needs "walrus operator". Why don't they implement assignment as binary operator?

To avoid the confusion between ```==``` and ```=``` which is a common error in other languages. ie.

```
if x = 5:
   ...
```

I always thought the lack of an assignment operator was a strength of python. The walrus operator seems like an odd thing to introduce. It'll save a couple of lines of code in it's few legitimate uses, but it'll get lots of use where it sholudn't.

PR #969, opened 2019-05-12
itertools islice

Jongy:
After https://github.com/RustPython/RustPython/pull/965.

After this branch, `Lib/reprlib.py` is importable, which leaves only `collections.namedtuple`  (Itself requiring lots of stuff) and we can use `Lib/functools.py` :)

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -3,14 +3,16 @@ use std::cmp::Ordering;
 use std::ops::{AddAssign, SubAssign};
 
 use num_bigint::BigInt;
+use num_traits::ToPrimitive;
 
 use crate::function::{OptionalArg, PyFuncArgs};
 use crate::obj::objbool;
+use crate::obj::objint;
 use crate::obj::objint::{PyInt, PyIntRef};
 use crate::obj::objiter::{call_next, get_iter, new_stop_iteration};
 use crate::obj::objtype;
 use crate::obj::objtype::PyClassRef;
-use crate::pyobject::{PyClassImpl, PyObjectRef, PyRef, PyResult, PyValue};
+use crate::pyobject::{IdProtocol, PyClassImpl, PyObjectRef, PyRef, PyResult, PyValue};
 use crate::vm::VirtualMachine;
 
 #[pyclass(name = "chain")]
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -181,6 +184,141 @@ impl PyItertoolsTakewhile {
...
+    fn new(_cls: PyClassRef, args: PyFuncArgs, vm: &VirtualMachine) -> PyResult {
+        let (iter, start, stop, step) = match args.args.len() {
+            0 | 1 => {
+                return Err(vm.new_type_error(format!(
+                    "islice expected at least 2 arguments, got {}",
+                    args.args.len()
+                )))
+            }
+
+            2 => {
```
Jongy: Argument binding is too complicated now, I want to improve this bit

vm/src/stdlib/itertools.rs
```diff
@@ -181,6 +184,141 @@ impl PyItertoolsTakewhile {
...
+    }
+
+    #[pymethod(name = "__next__")]
+    fn next(&self, vm: &VirtualMachine) -> PyResult {
+        while *self.cur.borrow() < *self.next.borrow() {
+            call_next(vm, &self.iterable)?;
+            *self.cur.borrow_mut() += 1;
+        }
+
+        if self.stop.is_some() && *self.cur.borrow() >= self.stop.unwrap() {
```
sapir: You can avoid the `is_some`+`unwrap` here.
```rust
if let Some(stop) = self.stop {
  if *self.cur.borrow() >= stop {
    return Err;
  }
```
or maybe
```rust
if self.stop.map_or(false, |stop| *self.cur.borrow() >= stop) {
  return Err;
}
```
or
```rust
if self.stop.map(|stop| *self.cur.borrow() >= stop).unwrap_or(false) {
  return Err;
}
```

vm/src/stdlib/itertools.rs
```diff
@@ -181,6 +184,141 @@ impl PyItertoolsTakewhile {
...
+                (iter, 0usize, stop, 1usize)
+            }
+            _ => {
+                let (iter, start, stop, step): (
+                    PyObjectRef,
+                    PyObjectRef,
+                    PyObjectRef,
+                    PyObjectRef,
+                ) = args.bind(vm)?;
+                let stop = if !stop.is(&vm.get_none()) {
```
sapir: The entire `if` statement can be refactored to a `pyobject_to_opt_usize` function, that returns a `PyResult<Option<usize>>`, then you can reuse it a few times.

vm/src/stdlib/itertools.rs
sapir: Maybe something like

```rust
// Get the argument indices
let (start, stop, step) = match args.args.len() {
  2 => (None, Some(1), None),
  3 => (Some(1), Some(2), None),
  4 => (Some(1), Some(2), Some(3)),
  0 | 1 => return Err("not enough args"),
  _ => return Err("too many args"),
};

// I don't know how you retrieve an argument by index.
let iter: PyObjectRef = get_arg(0).unwrap();
let start: usize = match start {
  Some(i) => get_arg(i).to_usize().ok_or_else(|| Err(...))?,
  _ => 0,
};
...
```

vm/src/stdlib/itertools.rs
```diff
@@ -181,6 +184,141 @@ impl PyItertoolsTakewhile {
...
+        if self.stop.is_some() && *self.cur.borrow() >= self.stop.unwrap() {
+            return Err(new_stop_iteration(vm));
+        }
+
+        let obj = call_next(vm, &self.iterable)?;
+        *self.cur.borrow_mut() += 1;
+
+        // TODO is this overflow check required? attempts to copy CPython but it's
+        // currently broken.
+        let (next, ovf) = (*self.next.borrow()).overflowing_add(self.step);
```
sapir: You can call `borrow_mut` once for this bit, then reuse the borrow.
How about
```rust
let next = self.next.borrow_mut();
*next = *next.checked_add(self.step).or(self.stop).ok_or_else(|| new_vm_error())?;
```

But why are you checking for overflow here, if you didn't check for overflow in the `+= 1` lines?

vm/src/stdlib/itertools.rs
```diff
@@ -181,6 +184,141 @@ impl PyItertoolsTakewhile {
...
+impl PyValue for PyItertoolsIslice {
+    fn class(vm: &VirtualMachine) -> PyClassRef {
+        vm.class("itertools", "islice")
+    }
+}
+
+fn pyobject_to_usize(obj: PyObjectRef, vm: &VirtualMachine) -> Option<usize> {
+    let is_int = objtype::isinstance(&obj, &vm.ctx.int_type());
+    if is_int {
+        objint::get_value(&obj).to_usize()
```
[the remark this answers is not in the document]
Jongy: All the `get_value` / `borrow_value` family of functions basically do `obj.payload::<T>().unwrap()` so if you pass a wrong type, we panic.

vm/src/stdlib/itertools.rs
```diff
@@ -181,6 +184,141 @@ impl PyItertoolsTakewhile {
...
+                        return Err(vm.new_value_error("Indices for islice() must be None or an integer: 0 <= x <= sys.maxsize.".to_string()));
+                    }
+                    start.unwrap()
+                } else {
+                    0usize
+                };
+
+                let step = if !step.is(&vm.get_none()) {
+                    let step = pyobject_to_usize(step, &vm);
+                    if !step.is_some() {
```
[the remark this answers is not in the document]
Jongy: Mhm, nice

vm/src/stdlib/itertools.rs
Jongy: Basically followed CPython's `itertools` on this, which doesn't check for an overflow here.

Technically CPython's `itertools` is wrong because after an overflow on `next` when `stop=None`, `next` is reset to `-1` which is incorrect because `cnt` itself will overflow successfully to `ssize_t`'s min (If it won't be optimized away due to its UB being), and so the next object return by `islice` will be after a random step.

Now all of this is purely hypothetical here since we use `usize` which will usually take the form of 64 bits.

vm/src/stdlib/itertools.rs
```diff
@@ -181,6 +184,141 @@ impl PyItertoolsTakewhile {
...
+        objint::get_value(&obj).to_usize()
+    } else {
+        None
+    }
+}
+
+#[pyimpl]
+impl PyItertoolsIslice {
+    #[pymethod(name = "__new__")]
+    fn new(_cls: PyClassRef, args: PyFuncArgs, vm: &VirtualMachine) -> PyResult {
```
coolreader18: Try something like:

```rust
fn new(_cls: PyClassRef, iter: PyIterable, start: Option<PyInt>, stop:  OptionalArg<Option<PyInt>>, step: OptionalArg<Option<PyInt>>, vm: &VirtualMachine) -> PyResult {
    let (start, stop, step) = match (start, stop.into_option().and_then(|x| x), step.into_option().and_then(|x| x)) {
        ...
    };
}
```

vm/src/stdlib/itertools.rs
Jongy: Liked the first one

coolreader18:
Are you still working on this @Jongy?

Jongy:
Yeah, though I won't get to it in the upcoming week.

vm/src/stdlib/itertools.rs
Jongy: Changed it so there's no code dup in the handling of `stop` argument.

vm/src/stdlib/itertools.rs
Jongy: Actually I don't think I can use it because it then when calling `islice(x, 5)` then `5` is treated as the `start` argument - not as the `stop` one, as required by CPython.

Jongy: It's indeed nicer to declare the arguments in the function prototype, but in this case it's not much of a a help because the argument order changes depending on the number of them present.

vm/src/stdlib/itertools.rs
Jongy: Not sure it will allow that much of code reuse; The tristate usize/`None`/`Err` has to be checked across all sites anyway, and has to be handled differently (Different default values for `None`, different errors for `Err`...)

Jongy:
Good enough, I assume.

coolreader18:
Yeah, that looks good to me. I'll merge this once tests pass.

PR #981, opened 2019-05-15
Add math.frexp

youknowone:


vm/src/stdlib/math.rs
```diff
@@ -209,6 +209,24 @@ fn math_floor(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     }
 }
 
+fn math_frexp(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
```
adrian17: `fn math_frexp(value: PyObjectRef, vm: &VirtualMachine) -> PyResult`?

youknowone: Thanks, I fixed it

coolreader18 (approved):

PR #985, opened 2019-05-18
Add exit/quit builtin function

RobertBerglund:


vm/src/builtins.rs
```diff
@@ -613,6 +613,24 @@ impl Printer for std::io::StdoutLock<'_> {
...
+    arg_check!(vm, args, required = [], optional = [(object, None)]);
+    match object {
+        Some(value) => match i32::try_from_object(&vm, value.clone()) {
+            Ok(code) => std::process::exit(code),
+            _ => {
+                let stdout = io::stdout();
+                let mut printer: Box<dyn Printer> = Box::new(stdout.lock());
+                printer.write(vm, value.clone())?;
+                printer.write(vm, "\n".into_pyobject(vm).unwrap())?;
+                printer.flush(vm)?;
```
coolreader18: Instead of all this, you can just do:

```rust
println!("{}", vm.to_str(&obj)?.as_str());
```

vm/src/builtins.rs
```diff
@@ -613,6 +613,24 @@ impl Printer for std::io::StdoutLock<'_> {
     }
 }
 
+pub fn builtin_exit(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
+    arg_check!(vm, args, required = [], optional = [(object, None)]);
```
coolreader18: Instead of using `arg_check!()`, we have a new system where we can do type checking based on the function declaration's parameters. I'd recommend making the function signature be:

```rust
pub fn builtin_exit(object: OptionalArg<PyObjectRef>, vm: &VirtualMachine) -> PyResult<()>
```

That will handle optional arg checking for you.

RobertBerglund: Thanks !
I am wondering why some functions take a &VirtualMachine as the first argument and others as the last argument.
Is there any rule for ordering the arguments ?

vm/src/builtins.rs
RobertBerglund: Thanks !
I made a new commit with the suggested changes.
I noticed that std::process::exit won't work in the wasm build for obvious reasons. Any suggestions on what could be done in this case ?

coolreader18: It looks like the test is passing for wasm, or do you mean that you tried it out, and it doesn't work? If that's the case, I'd say use `panic!()` to exit the program.

vm/src/builtins.rs
coolreader18: The functions that take `&VirtualMachine` as their first argument are all old-style python function definitions. Notice how they're all a `fn(&VirtualMachine, PyFuncArgs)` and use the `arg_check!` macro. The newer style is to specify argument types right in the function signature itself, and then have the `FromArgs` implementation for the type handle the type-checking. If it's a more complex function signature that you need to represent, e.g. with named arguments, you can use the `#[derive(FromArgs)]` procedural macro, a good example of which is the [`PrintOptions` struct](https://github.com/RustPython/RustPython/blob/8359c6e0310f2019b9e0618155d77e2a7b4c9074/vm/src/builtins.rs#L574-L584). The `&VirtualMachine` argument coming before or after doesn't really matter much, it's just that that's how it happened to be implemented for each style of function declaration.

vm/src/builtins.rs
RobertBerglund: Indeed, panic! will cause the program to exit, but will terminate with code 101 and std::process::exit() will cause a panic in the wasm built anyway.
I also noticed that the exit function won't print anything in the wasm build, what could I use instead of println! in this case ?

coolreader18: > std::process::exit() will cause a panic in the wasm build anyway (maybe this is ok ?)

It is ok; `exit` as a function doesn't really exist in wasm because when a wasm program aborts, it still doesn't quit the entire web page or anything like that.

> I also noticed that the builtin_exit function won't print anything in the wasm build, what could I use instead of println! in this case ?

I think it's fine for now, it doesn't really matter for now. Python in the browser isn't well defined, so it's fine if it doesn't work exactly how it would on the command line.

RobertBerglund: Thanks for the clarification.

PR #986, opened 2019-05-18
Add `itertools.chain`

Jongy:
Required for `Lib/collections/__init__.py`.

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -4,13 +4,71 @@ use std::ops::{AddAssign, SubAssign};
 
 use num_bigint::BigInt;
 
-use crate::function::OptionalArg;
+use crate::function::{OptionalArg, PyFuncArgs};
 use crate::obj::objbool;
 use crate::obj::objint::{PyInt, PyIntRef};
 use crate::obj::objiter::{call_next, get_iter, new_stop_iteration};
+use crate::obj::objtype;
 use crate::obj::objtype::PyClassRef;
 use crate::pyobject::{PyClassImpl, PyObjectRef, PyRef, PyResult, PyValue};
 use crate::vm::VirtualMachine;
+
+#[pyclass(name = "chain")]
+#[derive(Debug)]
+struct PyItertoolsChain {
+    iterables: Vec<PyObjectRef>,
+    cur: RefCell<(usize, Option<PyObjectRef>)>,
+}
+
+impl PyValue for PyItertoolsChain {
+    fn class(vm: &VirtualMachine) -> PyClassRef {
+        vm.class("itertools", "chain")
+    }
+}
+
+#[pyimpl]
+impl PyItertoolsChain {
+    #[pymethod(name = "__new__")]
+    fn new(_cls: PyClassRef, args: PyFuncArgs, vm: &VirtualMachine) -> PyResult {
+        Ok(PyItertoolsChain {
+            iterables: args.args,
+            cur: RefCell::new((0, None)),
+        }
+        .into_ref(vm)
+        .into_object())
+    }
+
+    #[pymethod(name = "__next__")]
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -4,14 +4,74 @@ use std::ops::{AddAssign, SubAssign};
...
 use crate::obj::objtype::PyClassRef;
 use crate::pyobject::{PyClassImpl, PyObjectRef, PyRef, PyResult, PyValue};
 use crate::vm::VirtualMachine;
 
+#[pyclass(name = "chain")]
+#[derive(Debug)]
+struct PyItertoolsChain {
+    iterables: Vec<PyObjectRef>,
+    cur: RefCell<usize>,
+    cur_iter: RefCell<Option<PyObjectRef>>,
```
coolreader18: I think you could combine these two `RefCell`s into one holding a tuple (`RefCell<(usize, Option<PyObjectRef>)>`) in order to reduce a bit of the overhead of runtime-checked mutability. Then, in `next()`, you could borrow both of them at the start of the method as so: `let (cur, cur_iter) = self.cur.borrow_mut()`. Then, you won't have any problems with borrows colliding as well.

vm/src/stdlib/itertools.rs
```diff
@@ -4,14 +4,74 @@ use std::ops::{AddAssign, SubAssign};
...
+                        *self.cur.borrow_mut() += 1;
+                        *self.cur_iter.borrow_mut() = None;
+                    } else {
+                        return Err(err);
+                    }
+                }
+            }
+        }
+
+        return Err(new_stop_iteration(vm));
```
coolreader18:
```suggestion
        Err(new_stop_iteration(vm))
```

vm/src/stdlib/itertools.rs
```diff
@@ -4,14 +4,74 @@ use std::ops::{AddAssign, SubAssign};
...
+                Err(err) => {
+                    if objtype::isinstance(&err, &vm.ctx.exceptions.stop_iteration) {
+                        *self.cur.borrow_mut() += 1;
+                        *self.cur_iter.borrow_mut() = None;
+                    } else {
+                        return Err(err);
+                    }
+                }
+            }
+        }
```
coolreader18: I think this could be represented better with recursion than a loop - maybe change the `while` to an `if` with the `new_stop_iteration` in the `else`, and then after incrementing `self.cur`, call `self.next(vm)`. That will also allow you to use `obj.or_else(|err| { ... })` instead of the `match` statement.

vm/src/stdlib/itertools.rs
Jongy: mhm

vm/src/stdlib/itertools.rs
Jongy: mhm. Tho, I thought it'd better if the code was constructed with as little mutable borrows as possible. (After all, to my understanding, the whole hassle of these `RefCell`s is `PyObject`'s `self` being immutable, which I guess is something we'll change in the future? So reconsutrcting the code with long mutable borrows will just get us more work later.

vm/src/stdlib/itertools.rs
Jongy: Hmm, okay. I based this function on CPython, which has a while loop itself. But I guess the likeliness of an semi-infinite recursion here is very low.

Jongy: Actually, the recursion trick doesn't work after I applied your suggestions in https://github.com/RustPython/RustPython/pull/986#discussion_r285356963 :(

Jongy: (Because the mutable borrow persists across calls)

coolreader18: Ah, okay. This seems fine then.

PR #988, opened 2019-05-18
Add sys.flags

skinny121:
For #958. For this, I created a `pystruct_sequence` proc macro as there are quite a few of these 'struct sequences' throughout the c part of the standard library.

### DIFF
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -1,3 +1,5 @@
+#![recursion_limit = "128"]
+
 extern crate proc_macro;
 
 #[macro_use]
@@ -37,4 +39,11 @@ pub fn pyimpl(attr: TokenStream, item: TokenStream) -> TokenStream {
     let attr = parse_macro_input!(attr as AttributeArgs);
     let item = parse_macro_input!(item as Item);
     result_to_tokens(pyclass::impl_pyimpl(attr, item))
+}
+
+#[proc_macro_attribute]
+pub fn pystruct_sequence(attr: TokenStream, item: TokenStream) -> TokenStream {
+    let attr = parse_macro_input!(attr as AttributeArgs);
+    let item = parse_macro_input!(item as Item);
+    result_to_tokens(pyclass::impl_pystruct_sequence(attr, item))
 }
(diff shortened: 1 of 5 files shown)


derive/src/pyclass.rs
```diff
@@ -316,11 +318,111 @@ pub fn impl_pyclass(attr: AttributeArgs, item: Item) -> Result<TokenStream2, Dia
...
+                };
+                methods.push(method);
+                let field_name_str = field_name.to_string();
+                // TODO add doc to the generated property
+                let method_reference = quote! {
+                    class.set_str_attr(
+                        #field_name_str,
+                        ::rustpython_vm::obj::objproperty::PropertyBuilder::new(ctx)
+                            .add_getter(Self::#method_ident)
+                            .create(),
```
coolreader18: You probably don't need to add it as a method that gets put as an item on the struct, you should be able to generate code like:

```rust
class.set_str_attr(
    "#field_name_str",
    ::rustpython_vm::obj::objproperty::PropertyBuilder::new(ctx)
        .add_getter({
            fn getter(zelf: ..., _vm: ...) -> ... { ... }
            getter
        })
        .create(),
);
```

derive/src/pyclass.rs
```diff
@@ -316,11 +318,111 @@ pub fn impl_pyclass(attr: AttributeArgs, item: Item) -> Result<TokenStream2, Dia
...
+                }
+            }
+        };
+        Ok(ret)
+    } else {
+        bail_span!(
+            item,
+            "#[pystruct_sequence] can only be on a struct declaration"
+        )
+    }
```
coolreader18: If you want, I'd say you could put `let struc = if let ... { struc } else { bail_span!(...) };` at the beginning, to help keep the bulk of the code less indented.

derive/src/pyclass.rs
skinny121: I was able to use a closure here, which is quite nice.

derive/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![recursion_limit = "128"]
```
windelbouwman: Why is this required? I have my doubts about changing the recursion limit to this large value. Is there a better way to implement the macros?

coolreader18: This is just for the quote macros, it doesn't have any effect on macros in `rustpython_vm`.

windelbouwman: Okay, its probably fine. The macro derive package is still a bit voodoo to me :)

coolreader18:
Merging.

skinny121:
Done.

PR #991, opened 2019-05-19
Remove RefCell from tuple

Jongy:
I didn't find any reason to have a shared `get_elements`, mostly since a direct check on `type in (tuple, list)` is done anyways before the call.

Removing the RefCell, the API now won't allow returning both a normal reference and a `Ref`, so if we want to keep this API then some larger change has to be made.

### DIFF
--- a/vm/src/import.rs
+++ b/vm/src/import.rs
@@ -67,7 +67,7 @@ pub fn import_file(
 
 fn find_source(vm: &VirtualMachine, current_path: PathBuf, name: &str) -> Result<PathBuf, String> {
     let sys_path = vm.get_attribute(vm.sys_module.clone(), "path").unwrap();
-    let mut paths: Vec<PathBuf> = objsequence::get_elements(&sys_path)
+    let mut paths: Vec<PathBuf> = objsequence::get_elements_list(&sys_path)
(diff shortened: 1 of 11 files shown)


vm/src/import.rs
```diff
@@ -67,7 +67,7 @@ pub fn import_file(
 
 fn find_source(vm: &VirtualMachine, current_path: PathBuf, name: &str) -> Result<PathBuf, String> {
     let sys_path = vm.get_attribute(vm.sys_module.clone(), "path").unwrap();
-    let mut paths: Vec<PathBuf> = objsequence::get_elements(&sys_path)
+    let mut paths: Vec<PathBuf> = objsequence::get_elements_list(&sys_path)
```
windelbouwman: The objsequence file was created with the overlapping functions of tuple and list in mind. Now that we have specific functions for tuple and list, move those two functions to the objtuple and objlist files.

Please move `get_elements_list` into the objlist.rs file and rename it to `get_elements`.
Please also move `get_elements_tuple` into the objtuple.rs file and rename it to `get_elements`.

windelbouwman:
I have one small remark, but this is not essential. Good improvement!

PR #560, opened 2019-02-26
refactor kwargs to HashMap

rickyhan:

PR #993, opened 2019-05-22
Move PyFuncArgs.kwargs to IndexMap, implement PEP-468

mkurnikov:
https://www.python.org/dev/peps/pep-0468/

Revives https://github.com/RustPython/RustPython/pull/560

### DIFF
--- a/vm/src/function.rs
+++ b/vm/src/function.rs
@@ -137,7 +137,9 @@ impl PyFuncArgs {
         self.kwargs.remove(name)
     }
 
-    pub fn remaining_keywords<'a>(&'a mut self) -> impl Iterator<Item = (String, PyObjectRef)> + 'a {
+    pub fn remaining_keywords<'a>(
+        &'a mut self,
+    ) -> impl Iterator<Item = (String, PyObjectRef)> + 'a {
         self.kwargs.drain(..)
     }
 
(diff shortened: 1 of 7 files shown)


vm/src/function.rs
```diff
@@ -132,19 +134,12 @@ impl PyFuncArgs {
...
-        if let Some(index) = self
-            .kwargs
-            .iter()
-            .position(|(arg_name, _)| arg_name == name)
-        {
-            Some(self.kwargs.remove(index).1)
-        } else {
-            None
-        }
+        self.kwargs.remove(name)
```
windelbouwman: Well, that was a good refactor!

vm/src/function.rs
```diff
@@ -79,19 +85,15 @@ impl PyFuncArgs {
...
         }
         default.clone()
     }
 
     pub fn get_optional_kwarg(&self, key: &str) -> Option<PyObjectRef> {
-        for (arg_name, arg_value) in self.kwargs.iter() {
-            if arg_name == key {
-                return Some(arg_value.clone());
-            }
+        if let Some(kwarg_value) = self.kwargs.get(key) {
```
windelbouwman: Could this function be reduced to only `self.kwargs.get(key).cloned()`?

windelbouwman:
I had a minor comment, but the change looks good!

PR #995, opened 2019-05-23
add binascii crc32

stwen77:


vm/src/stdlib/binascii.rs
```diff
@@ -56,6 +58,63 @@ fn binascii_unhexlify(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     Ok(vm.ctx.new_bytes(unhex))
 }
 
+static CRC32_TABLE: [u32; 256] = [
```
adrian17: Are there any CRC32 libraries on crates.io?

mkurnikov: There is https://crates.io/crates/crc, but also note here https://github.com/BurntSushi/rust-snappy/blob/master/src/crc32.rs#L1 (I don't know whether it's relevant anymore, though).

stwen77: > Are there any CRC32 libraries on crates.io?

Hi adrian, I just translate the implement in cpython to rust. Would you recommend to implement by crate crc?

windelbouwman: @stwen77 , thank you for your contribution. As @adrian17 and @mkurnikov suggested, please have a look at this crate: https://crates.io/crates/crc . It contains the crc32 logic so that we do not need to use the crc table. This is required in CPython since C does not have such a good eco system as rust. But in rust we can use many libraries, and I think we should take advantage of this ecosystem.

PR #997, opened 2019-05-24
objstr: Refactored the 'mul' method

ZapAnton:
- Used `str::repeat` instead of manually building the result string
- Rewritten the int type check to return error at the beginning of the
method
- Replaced the `unwrap` calls with the single `unwrap_or`

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -183,22 +183,16 @@ impl PyString {
 
     #[pymethod(name = "__mul__")]
     fn mul(&self, val: PyObjectRef, vm: &VirtualMachine) -> PyResult<String> {
-        if objtype::isinstance(&val, &vm.ctx.int_type()) {
-            let value = &self.value;
-            let multiplier = objint::get_value(&val).to_i32().unwrap();
-            let capacity = if multiplier > 0 {
-                multiplier.to_usize().unwrap() * value.len()
-            } else {
-                0
-            };
-            let mut result = String::with_capacity(capacity);
-            for _x in 0..multiplier {
-                result.push_str(value.as_str());
-            }
-            Ok(result)
-        } else {
-            Err(vm.new_type_error(format!("Cannot multiply {} and {}", self, val)))
+        if !objtype::isinstance(&val, &vm.ctx.int_type()) {
+            return Err(vm.new_type_error(format!("Cannot multiply {} and {}", self, val)));
         }
+        let value = &self.value;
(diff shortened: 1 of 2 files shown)


vm/src/obj/objstr.rs
```diff
@@ -183,22 +183,16 @@ impl PyString {
...
-            Ok(result)
-        } else {
-            Err(vm.new_type_error(format!("Cannot multiply {} and {}", self, val)))
+        if !objtype::isinstance(&val, &vm.ctx.int_type()) {
+            return Err(vm.new_type_error(format!("Cannot multiply {} and {}", self, val)));
         }
+        let value = &self.value;
+        let multiplier = objint::get_value(&val)
+            .to_i32()
+            .map(|multiplier| if multiplier < 0 { 0 } else { multiplier })
```
sapir: you could use `multiplier.max(0)`

ZapAnton: Done

vm/src/obj/objstr.rs
```diff
@@ -183,22 +183,16 @@ impl PyString {
...
-        } else {
-            Err(vm.new_type_error(format!("Cannot multiply {} and {}", self, val)))
+        if !objtype::isinstance(&val, &vm.ctx.int_type()) {
+            return Err(vm.new_type_error(format!("Cannot multiply {} and {}", self, val)));
         }
+        let value = &self.value;
+        let multiplier = objint::get_value(&val)
+            .to_i32()
+            .map(|multiplier| multiplier.max(0))
+            .and_then(|multiplier| multiplier.to_usize())
```
```suggestion
            .and_then(i32::to_usize)
```

silmeth:
You could also switch from `i32` to `isize` and return an `OverflowError` instead of doing `unwrap_or(0)` if the multiplier doesn’t fit in, it seems that’s what CPython does:

```python
>>> 'abc' * 234234234234234234234234234234 # does not fit in i64
OverflowError: cannot fit 'int' into an index-sized integer
>>> 'a' * 2147483650 # does not fit in i32, but works on 64 bit machines
'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa… # etc.
```

vm/src/obj/objstr.rs
ZapAnton: Got an error:
```text
expected signature of `fn(i32) -> _`
found signature of `for<'r> fn(&'r _) -> _`
```

coolreader18: Oh, it's fine then.

PR #998, opened 2019-05-25
Support frozen modules

palaviv:
As part of #958 we need to be able to import frozen modules. This is not the same as in Cpython as we do not compile the code and marshal it. I think that can wait for a later stage.

### DIFF
--- a/vm/src/import.rs
+++ b/vm/src/import.rs
@@ -18,6 +18,8 @@ pub fn import_module(vm: &VirtualMachine, current_path: PathBuf, module_name: &s
     // First, see if we already loaded the module:
     if let Ok(module) = sys_modules.get_item(module_name.to_string(), vm) {
         Ok(module)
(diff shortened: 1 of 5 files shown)


vm/src/import.rs
```diff
@@ -18,6 +18,8 @@ pub fn import_module(vm: &VirtualMachine, current_path: PathBuf, module_name: &s
     // First, see if we already loaded the module:
     if let Ok(module) = sys_modules.get_item(module_name.to_string(), vm) {
         Ok(module)
+    } else if let Some(frozen) = vm.frozen.borrow().get(module_name) {
+        import_file(vm, module_name, "frozen".to_string(), frozen.to_string())
```
coolreader18: I'm not sure how many built-in frozen modules there are going to be, but we might want to change `import_file` to take a `Cow<str, 'static>` if we're going to be holding big modules as an `&'static str`

PR #1000, opened 2019-05-25
objstr: Refactored the join method

ZapAnton:
This replaces the manual building of the mutable String via the `for`
loop with the `slice::join` method.

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -568,17 +568,12 @@ impl PyString {
 
     #[pymethod]
     fn join(&self, iterable: PyIterable<PyStringRef>, vm: &VirtualMachine) -> PyResult<String> {
-        let mut joined = String::new();
-
-        for (idx, elem) in iterable.iter(vm)?.enumerate() {
-            let elem = elem?;
-            if idx != 0 {
-                joined.push_str(&self.value);
-            }
-            joined.push_str(&elem.value)
-        }
-
-        Ok(joined)
+        Ok(iterable
+            .iter(vm)?
+            .filter_map(|elem| elem.ok())
+            .map(|elem| elem.value.to_owned())
+            .collect::<Vec<String>>()
+            .join(&self.value))
     }
 
     #[pymethod]


coolreader18:
Congrats on issue #1000! Looks good, I'll merge once tests pass.

ZapAnton:
The only thing that I am not sure about is that I do not return an error if the iter element an `Err` (e.g. nothing similar to the `let elem = elem?` line). Could this be improved?

coolreader18:
I'd say remove `.filter_map()` and `.collect::<PyResult<String>>()?`

adrian17:
To be sure, let's add this test to the test suite:

```
def f():
	yield 'a'
	yield 'b'
	yield 'c'
	1/0

with assertRaises(ZeroDivisionError):
	''.join(f())
```

By the way, this seems less efficient as we're copying all the strings in the iterable (and temporarily storing them in a vec), right?

coolreader18:
`itertools`  has a join function, not sure if we want to add that as a dependency though.

adrian17:
I think we already have `itertools` as a dependency :D

ZapAnton:
> By the way, this seems less efficient as we're copying all the strings in the iterable (and temporarily storing them in a vec), right?

There is no copying involved here. `to_owned` moves the String which copies only the pointer to String and then stores the pointers into the vec

silmeth:
> There is no string copying involved here. `to_owned` moves the String which copies only the pointer to String and then stores the pointers into the vec

Are you sure? You shouldn’t be able to move a string from a borrowed PyObject, and as far as I can tell, the `String::to_value` comes from [`ToOwned<T>` impl for all `Clone` types](https://doc.rust-lang.org/src/alloc/borrow.rs.html#80-91), which clones the value (that is, clones the whole string on the heap).

coolreader18:
Thanks for contributing, but I think this I'm going to reject this change. Using the iterator along with `to_owned()` clones the strings more than it needs to, and I don't think there's a way around that while using the iterator. If you can find a more idiomatic way to lay out this method with equal performance to the current way, please re-open, but I'm not sure if it's possible.

ZapAnton:
@coolreader18 , without using any unsafe operations, the only way I found to modify the code is the following:
```rust
Ok(itertools::join(
    iterable
    .iter(vm)?
    .collect::<PyResult<Vec<PyStringRef>>>()?
    .iter(),
    &self.value,
))        
```
Does this look acceptable?

coolreader18:
It does look better, but it still allocates an unnecessary `Vec`. The only way around that would be to write a custom iterator `join_result`, which would pretty much look like the `join` method as it is now.

PR #1001, opened 2019-05-25
Check RUSTPYTHONPATH as well as PYTHONPATH env variables

coolreader18:
The motivation for this is so that we could `export RUSTPYTHONPATH=${rustpython_checkout}/Lib` in our dev environments, and then not have to worry about having to do `PYTHONPATH=$PWD/Lib cargo run` or worrying about a permanently set PYTHONPATH with the RustPython `Lib` directory messing something up for python programs. If anyone has come up with a better way of managing this in their dev environment I'd be interested in hearing it.

### DIFF
--- a/vm/src/sysmodule.rs
+++ b/vm/src/sysmodule.rs
@@ -54,15 +54,19 @@ pub fn make_module(vm: &VirtualMachine, module: PyObjectRef, builtins: PyObjectR
         "cache_tag" => ctx.new_str("rustpython-01".to_string()),
     });
 
-    fn get_paths(env_name: &str) -> impl Iterator<Item = std::path::PathBuf> {
-        match env::var_os(env_name) {
-            Some(paths) => env::split_paths(&paths),
+    fn get_paths<'a>(
+        paths: &'a Option<std::ffi::OsString>,
+    ) -> impl Iterator<Item = std::path::PathBuf> + 'a {
+        match paths {
+            Some(paths) => env::split_paths(paths),
             None => env::split_paths(""),
         }
     }
 
-    let path_list = get_paths("RUSTPYTHONPATH")
-        .chain(get_paths("PYTHONPATH"))
+    let rustpy_path = env::var_os("RUSTPYTHONPATH");
+    let py_path = env::var_os("PYTHONPATH");
+    let path_list = get_paths(&rustpy_path)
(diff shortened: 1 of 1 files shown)


vm/src/sysmodule.rs
```diff
@@ -54,18 +54,23 @@ pub fn make_module(vm: &VirtualMachine, module: PyObjectRef, builtins: PyObjectR
...
-        None => vec![],
-    };
+    fn get_paths(env_name: &str) -> impl Iterator<Item = std::path::PathBuf> {
+        match env::var_os(env_name) {
+            Some(paths) => env::split_paths(&paths),
+            None => env::split_paths(""),
+        }
+    }
+
+    let path_list = get_paths("RUSTPYTHONPATH")
```
windelbouwman: Could we figure out this path from the location of the executable? This is how cpython does it, maybe we could do the same?

coolreader18: > This is how cpython does it

Based on [this code](https://github.com/python/cpython/blob/022be02dcfdfd9011415804bb4553a33fa7ec8f3/Python/pathconfig.c#L248), I think it's based on an external config option, which an environment variable effectively is. Even so, with the way that cargo works, the `Lib` directory won't always be in the same place relative to the executable.

coolreader18:
Tests should pass now, @windelbouwman is this good to merge or would you prefer some other solution?

vm/src/sysmodule.rs
```diff
@@ -54,17 +54,26 @@ pub fn make_module(vm: &VirtualMachine, module: PyObjectRef, builtins: PyObjectR
         "cache_tag" => ctx.new_str("rustpython-01".to_string()),
     });
 
-    let path_list = match env::var_os("PYTHONPATH") {
-        Some(paths) => env::split_paths(&paths)
+    let path_list = if cfg!(target_arch = "wasm32") {
```
windelbouwman: Off topic:

Still I believe that having specific platform checks in the rustpython core library is not correct. We can figure out a way later I believe in which we might have ways to insert paths into the vm. Another point in which there is wasm related code in the virtual machine is the wasm_id. I think the rustpython library must be agnostic of wasm, and we can achieve this.

windelbouwman:
I had some idea, but we can implement this in a later PR. The improvement is good!

PR #1002, opened 2019-05-26
bytearray.__setitem__

RobertBerglund:
Work in progress, need to handle memoryviews.


RobertBerglund:
> Work in progress, need to handle memoryviews.

Done

vm/src/obj/objbyteinner.rs
```diff
@@ -426,6 +426,69 @@ impl PyByteInner {
...
+                        }
+                    },
+                    _ => {Err(vm.new_type_error("an integer is required".to_string()))}
+                    );
+                    match result {
+                        Ok(value) => {
+                            self.elements[idx] = value;
+                            Ok(vm.new_int(value))
+                        }
+                        Err(error) => Err(error),
```
windelbouwman: This is a case of error propagation. Please use the `?` operator in this case to collapse the `match` statement into the simpler form:

```rust
let value = result?;
self.elements[idx] = value;
Ok(vm.new_int(value))
```

vm/src/obj/objbyteinner.rs
```diff
@@ -426,6 +426,69 @@ impl PyByteInner {
...
+                    Ok(items) => {
+                        let range = self
+                            .elements
+                            .get_slice_range(&slice.start_index(vm)?, &slice.stop_index(vm)?);
+                        self.elements.splice(range, items);
+                        Ok(vm
+                            .ctx
+                            .new_bytes(self.elements.get_slice_items(vm, slice.as_object())?))
+                    }
+                    Err(error) => Err(error),
```
windelbouwman: Please use the `?` operator to propagate errors as explained above.

vm/src/obj/objbyteinner.rs
```diff
@@ -426,6 +426,69 @@ impl PyByteInner {
         }
     }
 
+    pub fn setitem(
+        &mut self,
+        needle: Either<PyIntRef, PySliceRef>,
+        object: PyObjectRef,
+        vm: &VirtualMachine,
+    ) -> PyResult {
+        match needle {
```
windelbouwman: Since we match on the two types of needle, maybe it makes sense to create two seperate functions to keep the function small in size. This is what I would take a look at, leaving the function as it is is also fine with me :).

vm/src/obj/objbyteinner.rs
RobertBerglund: Thanks, I improved the code as suggested.

vm/src/obj/objbyteinner.rs
RobertBerglund: I agree, thanks for the tip :)

PR #1011, opened 2019-06-01
More _imp

palaviv:
Some more _imp functions. I think this should be enough for _bootstrap.py but I have not tested this yet.

### DIFF
--- a/vm/src/stdlib/imp.rs
+++ b/vm/src/stdlib/imp.rs
@@ -72,6 +72,11 @@ fn imp_init_frozen(name: PyStringRef, vm: &VirtualMachine) -> PyResult {
     }
 }
 
+fn imp_is_frozen_package(_name: PyStringRef, _vm: &VirtualMachine) -> bool {
+    // TODO: Support frozen package.
+    false
+}
+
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/imp.rs
```diff
@@ -25,6 +30,53 @@ fn imp_is_builtin(name: PyStringRef, vm: &VirtualMachine) -> bool {
...
+        if let Some(make_module_func) = vm.stdlib_inits.borrow().get(name) {
+            Ok(make_module_func(vm))
+        } else {
+            Ok(vm.get_none())
+        }
+    }
+}
+
+fn imp_exec_builtin(_mod: PyModuleRef, _vm: &VirtualMachine) -> i32 {
+    // TOOD: Should we do something here?
```
windelbouwman: It might be nice to use `unimplemented` here, or to log an error or warning with a logger?

palaviv: I am actually not sure if we should do something there... The CPython function `PyModule_ExecDef` does some sort of initializing that I don't think we currently need to do.

PR #1014, opened 2019-06-03
Cleanup get_method() error handling

mkurnikov:
1. Make raising `TypeError` from `get_method()` more explicit by splitting it into two methods: one that raises `TypeError` if there's no method, and one that's not (for some cases, like `__missing__` method for dict). 

2. If it's acceptable for method to be absent (`__missing__`, `__iter__` for iterators and others), do not raise `TypeError`. 
(note that if `__get__` descriptor raises an exception, it still gets propagates as before)

3. Make error messages more consistent with CPython. 

4. Add `exc_type` to `info!` exception logging.

### DIFF
--- a/vm/src/obj/objdict.rs
+++ b/vm/src/obj/objdict.rs
@@ -206,11 +206,10 @@ impl PyDictRef {
         if let Some(value) = self.entries.borrow().get(vm, &key)? {
             return Ok(value);
         }
-
-        if let Ok(method) = vm.get_method(self.clone().into_object(), "__missing__") {
+        if let Some(method_or_err) = vm.get_method(self.clone().into_object(), "__missing__") {
+            let method = method_or_err?;
             return vm.invoke(method, vec![key]);
         }
-
         Err(vm.new_key_error(format!("Key not found: {}", vm.to_pystr(&key)?)))
     }
 
--- a/vm/src/vm.rs
+++ b/vm/src/vm.rs
@@ -624,12 +624,24 @@ impl VirtualMachine {
 
     // get_method should be used for internal access to magic methods (by-passing
     // the full getattribute look-up.
-    pub fn get_method(&self, obj: PyObjectRef, method_name: &str) -> PyResult {
+    pub fn get_method_or_type_error(
+        &self,
+        obj: PyObjectRef,
+        method_name: &str,
+        err_msg: String,
+    ) -> PyResult {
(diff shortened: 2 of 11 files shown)


vm/src/vm.rs
```diff
@@ -624,14 +624,26 @@ impl VirtualMachine {
 
     // get_method should be used for internal access to magic methods (by-passing
     // the full getattribute look-up.
-    pub fn get_method(&self, obj: PyObjectRef, method_name: &str) -> PyResult {
+    pub fn get_method_or_type_error(
+        &self,
+        obj: PyObjectRef,
+        method_name: &str,
+        err_msg: String,
```
coolreader18: Could you make `err_msg` here a `FnOnce() -> String` instead of just a `String`? That way we can avoid unnecessary allocations when there's no error.

coolreader18:
Looks good, one small thing and then I'll merge. Thanks for contributing!

vm/src/obj/objdict.rs
```diff
@@ -206,11 +206,10 @@ impl PyDictRef {
         if let Some(value) = self.entries.borrow().get(vm, &key)? {
             return Ok(value);
         }
-
-        if let Ok(method) = vm.get_method(self.clone().into_object(), "__missing__") {
+        if let Some(method_or_err) = vm.get_method(self.clone().into_object(), "__missing__") {
```
windelbouwman: I think it is probably not handy to return `Option<Result<..>>`. Why not reverse the two so that we have the error as the outer type? Like this: `Result<Option<..>>`. Then we can use the `?` operator more intuitively.

windelbouwman:
Before merging, I would like @cthulahoops to have a chance to look at this, since this change touches the attribute related code.

vm/src/obj/objdict.rs
mkurnikov: I did it because `Option` is this case means "whether a method present on the object, or not". And if it's present, then we can check whether `__get__` method raised exception. 

There are three possible states
1. No method, `get_method` returns `None`.
2. Method exists, but `__get__` raised exception, returns `Err(some_error)`.
3. No exception, return method. 

Moving `Option` inside converts it into 4 states: 
1/2. Whether exception was returned from `__get__` or not. 
3/4. Whether method exists or not. 
It doesn't compatible with reality of the `get_method`.

PR #1016, opened 2019-06-04
Fix io.open("r") case and address some TODOs

mkurnikov:


vm/src/stdlib/io.rs
```diff
@@ -294,76 +292,117 @@ fn text_io_base_read(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+
+    let available_modes = hashset!('a', 'r', 'w');
+    let available_types = hashset!('t', 'b');
+    for ch in mode_value.chars() {
+        if ch == '+' {
+            if plus_is_set {
+                return Err(format!("invalid mode: '{}'", mode_value));
+            }
+            plus_is_set = true;
+        } else if available_modes.contains(&ch) {
```
mkurnikov: Should I replace this, and next, `contains` with `if ch == 'r' && ch == 'a'...` and remove `HashSet` usage?

coolreader18: That would be good, I think. You could also do `if let 'a' | 'r' | ... = c { ... }`

mkurnikov:
@coolreader18 It's ready for review/merge, I've made suggested changes and removed irrelevant ones.

PR #1017, opened 2019-06-04
Fix memoryview.obj misimplementation.

michelhe:
The memoryview skeleton did not correctly assign the "obj" attribute;
The "obj" attribute was assigned to the memoryview class and not the memryview object,
leading to a funny bug that can be reproduced as follows:

```python

m1 = memoryview(b'1234')
m2 = memoryview(b'abcd')

print(m1.obj)
```

the result would be the value inside `m2` instead that of `m1`


windelbouwman:
Thanks for contributing! This change looks good!

PR #1018, opened 2019-06-05
Importlib

palaviv:
This is a work in progress on the imprtlib module. I would love to get input as early as possible so please fill free to comment. There is still work to do to make everything works but the steps are:

1. Make `importlib._install` work.
2. Support `bootstrap_external`.
3. Change `__import__` to use the Importers in `sys.meta_path`

A big question I am not sure about is where to initialize the importlib? In Cpython it is done as part of the lifecycle. Where do you think we should do it?


vm/src/importlib.rs
```diff
@@ -0,0 +1,1175 @@
+pub const IMPORTLIB_BOOTSTRAP: &str = r#"
```
coolreader18: Could we put this in its own file? I don't know if it should be in `vm/src` or somewhere else, but it seems easier to use an `include_str!()`  than to have a 1000 line string literal in a rust source file.

palaviv: Way better. Thanks :)

palaviv:
> Should we put a `FIXME` or `XXX` here so that we can easily find it later?

Maybe `XXX` as I do not think we would like to fix this. I do not thin we should split `_os` into `posix` and `nt` as rust give us an abstraction over the OS on most methods.

windelbouwman:
This importlib stuff is pretty complex, thank you for having a look at this. The change looks pretty okay to me. It looks like pretty entangled code of frozen, import and the vm. Does the `concat!` macro work like `os.path.join`?

windelbouwman:
> A big question I am not sure about is where to initialize the importlib? In Cpython it is done as part of the lifecycle. Where do you think we should do it?

I would be my preference to initialize this optionally, seperate from the vm, so that we could re-use the vm without importlib if this is required. So in this case, I would suggest to initialize importlib in `src/main.rs`. In any case it is good to have a seperate importlib.rs file which contains the logic, but it does not have to be used always.

palaviv:
Why would you like to make this optional? For the WASM we can just replace the `PathFinder` for `sys.meta_path` with something specific for WASM.

windelbouwman:
I would like to make this optional for cases where you do not want to to use importlib. Or do you always use importlib? What is the fallback scenario? I'm not familiar with importlib, if it is the only option for importing stuff, it should not be optional, but baked in.

palaviv:
In Cpython you always use the frozen importlib `_bootstrap.__import__` function. I think that we should always use it as well instead of having a fallback to the current code.

vm/src/obj/objmodule.rs
```diff
@@ -34,6 +36,17 @@ impl PyModuleRef {
         }
     }
 
+    fn dict(self, vm: &VirtualMachine) -> PyResult<PyDictRef> {
+        let name_obj = vm.new_str(self.name.clone());
+        if let Some(ref dict) = &self.into_object().dict {
+            let mod_dict = dict.clone();
+            mod_dict.set_item("__name__", name_obj, vm)?;
```
windelbouwman: This set_item will happen each time the dict is retrieved. Is there a way to create the dict when the module is constructed? So in init or new?

palaviv:
This actually pass the tests now :tada: :tada: :tada:

A few comments:

1. The imports are slower then before.
2. We should change the import parsing to pass the `from_list` as one list.
3. I need to fix the change with `name` I did in `objmodule` (Waiting for response on gitter).
4. Why are we using `import` on `vm.class` and `vm.try_class`?

coolreader18:
> Why are we using import on vm.class and vm.try_class?

What's the alternative? We need to know what class a pyclass is associated with, and the best way we have so far is to store it in a module and import it in `class`. I did a little work in storing a `HashMap<TypeId, PyClassRef>` on `VirtualMachine` a while back, but I was busy then and nothing really came of it.

vm/src/obj/objmodule.rs
palaviv: Fixed

palaviv:
Thanks @coolreader18 I just had some infinite loop problems and wanted to understand more the logic.

About the import being slow. When running the following script:
```py
import time

start = time.time()
import os
print(time.time() - start)
```
When I run this without the `--release` flag I get ~13 [sec] while with the flag I get ~0.5 [sec].

palaviv:
After looking more into the slowness of the `import` I saw most of the time we lost is by calling the python function `__import__` on already loaded modules. This change the none `--release` time to ~5[sec] compared to the ~3[sec] of the current master. The `--release` version is 0.17[sec] while in the master it is 0.1 [sec].

I know this is about 1.5-2 times slower then the current import system but I believe we can optimize this at later stage. Do you guys think this is ready to be merged?

coolreader18:
This looks good to me at this point, @windelbouwman is this good to merge?

coolreader18:
Also, I'm sure that imports will speed up even more once we have pyc serialization and use `__pycache__`.

vm/src/vm.rs
```diff
@@ -300,13 +303,28 @@ impl VirtualMachine {
...
     }
 
-    pub fn import(&self, module: &str) -> PyResult {
-        match self.get_attribute(self.builtins.clone(), "__import__") {
-            Ok(func) => self.invoke(func, vec![self.ctx.new_str(module.to_string())]),
-            Err(_) => Err(self.new_exception(
-                self.ctx.exceptions.import_error.clone(),
-                "__import__ not found".to_string(),
-            )),
+    pub fn import(&self, module: &str, from_list: &PyObjectRef) -> PyResult {
```
windelbouwman: It might be good to move this whole function to the `import.rs` file. But that is something we can do later on.

vm/src/import.rs
```diff
@@ -11,19 +11,53 @@ use crate::pyobject::{ItemProtocol, PyResult};
...
+            module_name,
+            format!("frozen {}", module_name),
+            frozen.to_string(),
+        )
+    } else {
+        Err(vm.new_import_error(format!("Cannot import frozen module {}", module_name)))
+    }
+}
+
+pub fn import_builtin(vm: &VirtualMachine, module_name: &str) -> PyResult {
```
windelbouwman: This function name is slightly confusing with the existing function `builtin_import` in the file `builtins.rs`. We should come up with some better name, preferably not having the the word `builtin` in it. Maybe something like internal, or baked in, or rust lib.

windelbouwman:
Looks good to me!

PR #1019, opened 2019-06-07
Add _warnings

palaviv:
This is just a the `_warnings` module and a `warn` method without implementation so I can continue with #958.


vm/src/stdlib/warnings.rs
```diff
@@ -0,0 +1,37 @@
...
+    stacklevel: OptionalArg<u32>,
+}
+
+fn warnings_warn(args: WarnArgs, _vm: &VirtualMachine) {
+    // TODO: Implement correctly
+    let level = match args.stacklevel {
+        OptionalArg::Present(l) => l,
+        OptionalArg::Missing => 1,
+    };
+    println!(
```
coolreader18:
```suggestion
    eprintln!(
```

PR #1020, opened 2019-06-07
Disable the pwd module on android

coolreader18:
It was causing a build error due to some `libc` functions that aren't available on android.


windelbouwman:
Did you actually run rustpython on Android? Is there some manual available on how to do this?

coolreader18:
I use [Termux](https://termux.com/), a terminal emulator/full Linux environment for Android. There's an experimental package repository for it that includes, among other things, Rust development tools. When I first tried it a couple months ago, everything worked right out of the box; just `cargo run` and it compiles and works very well!

PR #1022, opened 2019-06-08
Add line continuation for the WASM demo terminal

coolreader18:
This is an old branch from a few months ago (I just forgot to make a PR), but it still works and there are no merge conflicts!


windelbouwman:
My javascript-fu is weak, but this seems ok to me.

PR #1024, opened 2019-06-09
str.isprintable

yanganto:
Implement `isprintable` function of `str` object

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -1,3 +1,4 @@
+extern crate unicode_categories;
 extern crate unicode_xid;
 
 use std::fmt;
@@ -26,6 +27,8 @@ use super::objnone::PyNone;
 use super::objsequence::PySliceableSequence;
 use super::objslice::PySlice;
 use super::objtype::{self, PyClassRef};
+
+use unicode_categories::UnicodeCategories;
 
 /// str(object='') -> str
 /// str(bytes_or_buffer[, encoding[, errors]]) -> str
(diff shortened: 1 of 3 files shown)


vm/src/obj/objstr.rs
```diff
@@ -519,6 +519,88 @@ impl PyString {
         }
     }
 
+    // Return true if all characters in the string are printable or the string is empty,
```
windelbouwman: If you enhande these comments to the triple `///` format, the comment will be the python doc comment I believe.

vm/src/obj/objstr.rs
```diff
@@ -519,6 +519,88 @@ impl PyString {
...
+    //   * Zl Separator, Line ('\u2028', LINE SEPARATOR)
+    //   * Zp Separator, Paragraph ('\u2029', PARAGRAPH SEPARATOR)
+    //   * Zs (Separator, Space) other than ASCII space('\x20').
+    #[pymethod]
+    fn isprintable(&self, _vm: &VirtualMachine) -> bool {
+        self.value.is_empty()
+            || self.value.chars().all(|c| {
+                match c {
+                    // Other, Control
+                    '\u{0000}'| '\u{0001}'| '\u{0002}'| '\u{0003}'| '\u{0004}'| '\u{0005}'|
```
windelbouwman: The should exist a rust library which has these tables implemented. For example this library: https://crates.io/crates/unicode_categories

windelbouwman:
@yanganto thank you for contributing to this project! I left some review comments in your changes.

vm/src/obj/objstr.rs
yanganto: Thanks

vm/src/obj/objstr.rs
```diff
@@ -519,6 +522,30 @@ impl PyString {
...
+    /// database as following categories are considered printable.
+    ///   * Cc (Other, Control)
+    ///   * Cf (Other, Format)
+    ///   * Cs (Other, Surrogate)
+    ///   * Co (Other, Private Use)
+    ///   * Cn (Other, Not Assigned)
+    ///   * Zl Separator, Line ('\u2028', LINE SEPARATOR)
+    ///   * Zp Separator, Paragraph ('\u2029', PARAGRAPH SEPARATOR)
+    ///   * Zs (Separator, Space) other than ASCII space('\x20').
+    #[pymethod]
```
coolreader18: This isn't necessary, `Iterator::all` returns true if the iterator has no elements.

vm/src/obj/objstr.rs
```diff
@@ -519,6 +522,30 @@ impl PyString {
...
+    fn isprintable(&self, _vm: &VirtualMachine) -> bool {
+        self.value.is_empty()
+            || self.value.chars().all(|c| match c {
```
coolreader18: Should this be a `||`?

yanganto: `"".isprintable()` return `True`, I will remove empty check because `Iterator::all` returns true if the iterator has no elements.

vm/src/obj/objstr.rs
yanganto: Thank you ~

PR #1026, opened 2019-06-09
String/BytesIO

rmliddle:


vm/src/stdlib/io.rs
```diff
@@ -357,7 +407,7 @@ pub fn io_open(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
         optional = [(mode, Some(vm.ctx.str_type()))]
     );
     // mode is optional: 'rt' is the default mode (open from reading text)
-    let mode_string = mode.map_or("rt".to_string(), |m| objstr::get_value(m));
+    let mode_string = mode.map_or("rt".to_string(), |m| get_value(m));
```
mkurnikov: I think `get_value` is too ambiguous name to be used without namespace here. `objstr::get_value` is easier to read. And it's against Rust guidelines. 

Also, could we use just `objstr::get_value` here, without creating new closure?

vm/src/stdlib/io.rs
```diff
@@ -140,7 +190,7 @@ fn file_io_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
         optional = [(mode, Some(vm.ctx.str_type()))]
     );
 
-    let rust_mode = mode.map_or("r".to_string(), |m| objstr::get_value(m));
+    let rust_mode = mode.map_or("r".to_string(), |m| get_value(m));
```
mkurnikov: Same as below.

vm/src/stdlib/io.rs
```diff
@@ -45,32 +45,82 @@ impl PyValue for PyStringIO {
...
+    fn read(self, _vm: &VirtualMachine) -> String {
+        let data = self.data.borrow().clone();
+        self.data.borrow_mut().clear();
+        data
+    }
 }
 
-fn string_io_new(cls: PyClassRef, vm: &VirtualMachine) -> PyResult<PyStringIORef> {
+fn string_io_new(
+    cls: PyClassRef,
```
coolreader18: `object` should be an `OptionalArg<Option<PyStringRef>>`

vm/src/stdlib/io.rs
```diff
@@ -357,7 +407,7 @@ pub fn io_open(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
         optional = [(mode, Some(vm.ctx.str_type()))]
```
coolreader18: This could probably just be a `match` to avoid unnecessary allocation.

vm/src/stdlib/io.rs
```diff
@@ -45,32 +45,82 @@ impl PyValue for PyStringIO {
...
+        let data = self.data.borrow().clone();
+        self.data.borrow_mut().clear();
+        data
+    }
 }
 
-fn string_io_new(cls: PyClassRef, vm: &VirtualMachine) -> PyResult<PyStringIORef> {
+fn string_io_new(
+    cls: PyClassRef,
+    object: OptionalArg<PyObjectRef>,
```
coolreader18: `object` should be an `OptionalArg<Option<PyStringRef>>`

vm/src/stdlib/io.rs
coolreader18: This could probably just be a `match` in order to avoid unnecessary allocation.

coolreader18:
LGTM, do you want to merge this now @rmliddle?

rmliddle:
@coolreader18 The behaviour of the implementation here in particular around read is quite inconsistent with that of CPython, just working on a few changes to make this better! Should have that ready for review shortly

PR #1027, opened 2019-06-09
Feature: str.format accepting !r, !s and !a

alanjds:
From: https://gitter.im/rustpython/Lobby?at=5cfa898ebf4cbd167c484cf0

As the Python docs, when a `{!r}` occur, it runs `__repr__` over the argument before applying the `:` rules. The current implementation of `{!a}` is not correct, as simple acts as `!r`.

### DIFF
--- a/vm/src/format.rs
+++ b/vm/src/format.rs
@@ -4,6 +4,24 @@ use std::cmp;
 use std::str::FromStr;
 
 #[derive(Debug, Copy, Clone, PartialEq)]
+pub enum FormatPreconversor {
+    Str,
+    Repr,
+    Ascii,
+}
+
+impl FormatPreconversor {
+    fn from_char(c: char) -> Option<FormatPreconversor> {
+        match c {
+            's' => Some(FormatPreconversor::Str),
+            'r' => Some(FormatPreconversor::Repr),
+            'a' => Some(FormatPreconversor::Ascii),
+            _ => None,
+        }
+    }
+}
+
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub enum FormatAlign {
     Left,
     Right,
@@ -56,6 +74,7 @@ pub enum FormatType {
 
 #[derive(Debug, PartialEq)]
 pub struct FormatSpec {
+    preconversor: Option<FormatPreconversor>,
     fill: Option<char>,
     align: Option<FormatAlign>,
     sign: Option<FormatSign>,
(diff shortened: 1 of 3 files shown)


alanjds:
@palaviv: Can you please take a look if it suits your needs?

vm/src/format.rs
```diff
@@ -3,6 +3,49 @@ use num_traits::Signed;
...
+
+    pub fn from_str(text: &str) -> Option<FormatPreconversor> {
+        let mut chars = text.chars();
+        if chars.next() != Some('!') {
+            return None;
+        }
+
+        match chars.next() {
+            None => None, // Should fail instead?
+            Some(c) => FormatPreconversor::from_char(c),
```
coolreader18: This can just be `chars.next().map(FormatPreconversor::from_char)`

palaviv:
This looks great to me.  Thanks @alanjds this will help me very much in #1018.

vm/src/format.rs
alanjds: Hum... this still returns None if `chars.next()` is None?

Asking because `FormatPreconversor::from_char` is not prepared to handle None as input.

alanjds: Another aspect: `// Should fail instead?`

I do not know how to raise exceptions. Ideally, `{!x}` should fail, but this code will pass for now:
```py
# CPy 3.6
>>> '{!x}'.format(123)
  File "<ipython-input-2-b294f18f6498>", line 1, in <module>
    '{!x}'.format(123)
ValueError: Unknown conversion specifier x
```
```py
# RustPython
>>>>> '{!x}'.format(123)
'123'
```

silmeth: `next().map(FormatPreconversor::from_char)` would just map the inner value of the `Option` if present – would do exactly what you did (pass the inner value to `from_char()` if it exists, or return None if the `Option` is `None`).

To raise an exception, you’d need to signal an error somehow (eg. return a `Result<Option<FormatPreconversor>, String>` instead of an `Option<FormatPreconversor>`) and then check for the result in `call_object_format()` and in case of an error return, return a new vm error there (eg. by `return Err(vm.new_value_error(…))`). Or just by mapping the error:

```rust
let (preconversor, new_format_spec) = FormatPreconversor::parse_and_consume(format_spec);
let preconversor = preconversor.map_err(|msg| vm.new_value_error(msg))?;
// …
```

But then you’d probably want to return `Result<(Option<FormatPreconversor>, &str), String>` rather than `(Result<Option<FormatPreconversor>, String>, &str)`, so that it’d look like:

```rust
let (preconversor, new_format_spec) = FormatPreconversor::parse_and_consume(format_spec)
    .map_err(|msg| vm.new_value_error(msg))?;
```

PR #1029, opened 2019-06-13
Split off bytecode compilation into a separate crate

### LINKED ISSUE #1023: [RFC] Split off bytecode compilation into a separate crate
## Summary

Make a new crate, `rustpython_bytecode`, that holds the bytecode data structures and does the compilation from the AST from `rustpython_parser` to bytecode.

## Detailed Explanation

This would likely be easy to implement, just move `bytecode.rs` and `compile.rs` into a separate crate. This would allow us to let `rustpython_derive` depend on `rustpython_bytecode` and we would be able to compile Python to bytecode at compile time. Then we would be able to create frozen modules more efficiently: just run the bytecode, no need to parse the source code at runtime.

I believe @windelbouwman said that it would be nice to not fragment RustPython into too many crates, citing npm and `left-pad` as an argument against doing so. I believe that this is a fairly logical place to split up RustPython, allowing us separation of concerns. In addition, Rust's crate system and cargo allow many crates to be depended upon easily, and many large players in Rust's ecosystem have modularized, allowing consumers to only depend on what they need (see: tokio, actix, amethyst, wasmer, etc). We're not in that exact situation, but still.

An alternative to this would just be a `build.rs` file for `rustpython` that depends on `rustpython_vm` and compiles source code there, but that feels less elegant and would have to be copied for `rustpython_wasm` or anyone that wants to embed RustPython into their app.

## Unresolved Questions

There may be some other solution to this compile-time compilation I haven't thought of, if anyone could think of something else I'd be interested to hear it.

coolreader18:
This only splits the crate, it doesn't add any macros like I discussed in #1023.

Closes #1023.


coolreader18:
@windelbouwman I decided against both `rustpython_compiler` and `rustpython_bytecode` because there likely won't be many cases where you would need just the functionality of `rustpython_bytecode` by itself, as a standalone crate with essentially only data structure definitions. If you do need that, you could still just use `rustpython_compiler` and DCE would take care of the extra compilation functionality.

windelbouwman:
The usecase for having a seperate compiler and bytecode crate is to be able to execute bytecode on the vm without requiring the compiler. I know that this is done in micropython, to reduce size on the target, they precompile the python code on host and run it on the target.

We can do this later as you suggested, first we need to port it to microcontrollers.

coolreader18:
That makes sense. Also, I'm not sure why builds are failing; it seems to be a timeout issue with caching so I deleted the PR's cache to try to see if that will work.

coolreader18:
Alright, tests are passing. Can we merge this?

coolreader18:
I'm going to merge this so that #1030 can work with the new code structure.

PR #1030, opened 2019-06-13
Import all from_list in one call

palaviv:
Currently when doing `from module import a,b,c,d,e` we would call `vm.import` one time for each symbol. This change the parser so we will call `vm.import` once for all symbols.

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -357,8 +357,8 @@ impl Frame {
             }
             bytecode::Instruction::Import {
                 ref name,
-                ref symbol,
-            } => self.import(vm, name, symbol),
+                ref symbols,
+            } => self.import(vm, name, symbols),
             bytecode::Instruction::ImportStar { ref name } => self.import_star(vm, name),
             bytecode::Instruction::LoadName {
                 ref name,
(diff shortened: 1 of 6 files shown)


vm/src/frame.rs
```diff
@@ -907,25 +907,23 @@ impl Frame {
...
-        // that
-        let obj = match symbol {
-            Some(symbol) => vm.get_attribute(module, symbol.as_str()).map_err(|_| {
-                let import_error = vm.context().exceptions.import_error.clone();
-                vm.new_exception(import_error, format!("cannot import name '{}'", symbol))
-            }),
-            None => Ok(module),
-        };
+    fn import(&self, vm: &VirtualMachine, module: &str, symbols: &Vec<String>) -> FrameResult {
+        let mut from_list = vec![];
```
coolreader18:
```rust
let from_list = symbols.iter().map(...);
```

vm/src/compile.rs
```diff
@@ -226,29 +226,51 @@ impl Compiler {
...
+                            if import_star {
+                                self.emit(Instruction::ImportStar {
+                                    name: module.clone(),
+                                });
+                            } else {
+                                self.emit(Instruction::Import {
+                                    name: module.clone(),
+                                    symbols: symbols_strings,
+                                });
+                                names.iter().for_each(|name| self.store_name(&name));
```
coolreader18: Wouldn't this still import the original symbol names, but just add the aliases on top of them?

vm/src/compile.rs
```diff
@@ -226,29 +226,51 @@ impl Compiler {
...
-                        }
-                        _ => {
+                    if let Some(alias) = alias {
+                        self.emit(Instruction::Import {
+                            name: module.clone(),
+                            symbols: vec![],
+                        });
+                        self.store_name(&alias);
+                    } else {
+                        if symbols.is_empty() {
```
coolreader18: This can be collapsed into an `else if`

vm/src/compile.rs
coolreader18: Wait, nevermind, I realized what this does.

vm/src/compile.rs
```diff
@@ -226,29 +226,51 @@ impl Compiler {
...
-                                    Some(symbol) => symbol.clone(),
-                                    None => module.clone(),
-                                },
-                            };
-                            self.store_name(&name);
+                            self.store_name(&module.clone());
+                        } else {
+                            let mut import_star = false;
+                            let mut symbols_strings = vec![];
+                            let mut names = vec![];
```
coolreader18:
```rust
let import_star = symbols.iter().any(|ast::ImportSymbol { symbol, .. }| symbol == "*");
let (mut names, symbols_strings): (Vec<String>, Vec<String>) = symbols
    .into_iter().map(|ast::ImportSymbol { symbol, alias }| (alias.unwrap_or_else(|| symbol.to_string()), symbol))
    .unzip()
    .collect();
names.reverse();
```
I think that should work, and it has the benefit of using `length_hint()` from Vec so that it doesn't have to make a bunch of separate allocations.

vm/src/frame.rs

vm/src/compile.rs
palaviv: Done

vm/src/compile.rs
palaviv: Nice. This is much simpler now :+1:

PR #1032, opened 2019-06-14
Printf-style formatting

michelhe:
This is a WIP on `str % tuple` formatting.
Currently on the parser itself is available for reviewing, I will be happy to take your comments and improve it.

Example usage:
```rust
let format_string = match CFormatString::from_str("Hello, My name is %s and I'm %d years old. My favorite hex number is %#08x") {
        Ok(fmt) => fmt,
        Err(_) => { panic!("got an error") }
    };
    println!("{:?}", format_string.format_parts);
```

will output:
```
[(0, Literal("Hello, My name is ")), (18, Spec(CFormatSpec { mapping_key: None, flags: (empty), min_field_width: None, precision: None, format_type: String(Str), format_char: 's' })), (20, Literal(" and I\'m ")), (29, Spec(CFormatSpec { mapping_key: None, flags: (empty), min_field_width: None, precision: None, format_type: Number(Decimal), format_char: 'd' })), (31, Literal(" years old. My favorite hex number is ")), (69, Spec(CFormatSpec { mapping_key: None, flags: ALTERNATE_FORM | ZERO_PAD, min_field_width: Some(8), precision: None, format_type: Number(Hex(Lowercase)), format_char: 'x' }))]

```

### DIFF
--- a/vm/src/cformat.rs
+++ b/vm/src/cformat.rs
@@ -19,6 +19,7 @@ pub enum CFormatErrorType {
 // also contains how many chars the parsing function consumed
 type ParsingError = (CFormatErrorType, usize);
 
+#[derive(Debug, PartialEq)]
 pub struct CFormatError {
     pub typ: CFormatErrorType,
     pub index: usize,
@@ -77,6 +78,7 @@ pub struct CFormatSpec {
     pub precision: Option<usize>,
     pub format_type: CFormatType,
     pub format_char: char,
+    chars_consumed: usize,
 }
 
 impl CFormatSpec {
@@ -86,7 +88,12 @@ impl CFormatSpec {
             .collect::<String>()
     }
 
-    pub fn fill_string(&self, string: String, fill_char: char, num_prefix_chars: Option<usize>) -> String {
+    pub fn fill_string(
+        &self,
+        string: String,
+        fill_char: char,
+        num_prefix_chars: Option<usize>,
+    ) -> String {
         let mut num_chars = string.len();
         if let Some(num_prefix_chars) = num_prefix_chars {
             num_chars = num_chars + num_prefix_chars;
@@ -125,11 +132,6 @@ impl CFormatSpec {
     pub fn format_number(&self, num: &BigInt) -> String {
         use CFormatCase::{Lowercase, Uppercase};
         use CNumberType::*;
-        let fill_char = if self.flags.contains(CConversionFlags::ZERO_PAD) {
-            '0'
-        } else {
-            ' '
-        };
         let magnitude = num.abs();
         let prefix = if self.flags.contains(CConversionFlags::ALTERNATE_FORM) {
             match self.format_type {
@@ -154,7 +156,6 @@ impl CFormatSpec {
             _ => unreachable!(), // Should not happen because caller has to make sure that this is a number
         };
 
-
         let sign_string = match num.sign() {
             Sign::Minus => "-",
             _ => {
@@ -168,16 +169,22 @@ impl CFormatSpec {
             }
         };
 
-
         let prefix = format!("{}{}", sign_string, prefix);
-        let magnitude_filled_string = if !self.flags.contains(CConversionFlags::LEFT_ADJUST) {
-            // '-' flags overrides '0' flag
-            self.fill_string(magnitude_string, fill_char, Some(prefix.len()))
-        } else {
-            magnitude_string
-        };
 
-        format!("{}{}", prefix, magnitude_filled_string)
+        if self.flags.contains(CConversionFlags::ZERO_PAD) {
+            let fill_char = if !self.flags.contains(CConversionFlags::LEFT_ADJUST) {
+                '0'
+            } else {
+                ' ' // '-' overrides the '0' conversion if both are given
+            };
+            format!(
+                "{}{}",
+                prefix,
+                self.fill_string(magnitude_string, fill_char, Some(prefix.len()))
+            )
+        } else {
+            self.fill_string(format!("{}{}", prefix, magnitude_string), ' ', None)
+        }
     }
 }
 
@@ -223,11 +230,9 @@ impl FromStr for CFormatString {
                     index = index + consumed;
                     new_text
                 })
-                .map_err(|(e, consumed)| {
-                    CFormatError {
-                        typ: e,
-                        index: index + consumed,
-                    }
+                .map_err(|(e, consumed)| CFormatError {
+                    typ: e,
+                    index: index + consumed,
                 })?;
         }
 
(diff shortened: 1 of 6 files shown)


vm/src/obj/objstr.rs
```diff
@@ -429,6 +430,29 @@ impl PyString {
...
+    #[pymethod(name = "__mod__")]
+    fn modulo(&self, values: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        let format_string_text = &self.value;
+
+        match CFormatString::from_str(format_string_text.as_str()) {
+            Ok(format_string) => perform_clike_format(vm, format_string, values.clone()),
+            Err(err) => match err.typ {
+                CFormatErrorType::UnsupportedFormatChar(c) => Err(vm.new_value_error(format!(
+                    "unsupported format character '{}' ({:#x}) at index {}",
+                    c, c as u8, err.index
```
windelbouwman: In the `c as u8` we assume ascii, which might not be the case. Consider `c as u32` at least. Is there a way to get the unicode code point from a char as `usize`? This would be nice!

vm/src/obj/objstr.rs
```diff
@@ -429,6 +430,29 @@ impl PyString {
...
     }
 
+    #[pymethod(name = "__mod__")]
+    fn modulo(&self, values: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        let format_string_text = &self.value;
+
+        match CFormatString::from_str(format_string_text.as_str()) {
+            Ok(format_string) => perform_clike_format(vm, format_string, values.clone()),
+            Err(err) => match err.typ {
+                CFormatErrorType::UnsupportedFormatChar(c) => Err(vm.new_value_error(format!(
```
windelbouwman: Instead of rendering the `CFormatErrorType` to a string, implement the `Display` trait for `CFormatErrorType`, or add a method to `CFormatErrorType` called `to_string` or something like that. Then this code can resolve to:
```rust
fn module() {

    let format_string = CFormatString::from_str(format_string_text)
             .map_err(|err| vm.new_value_error(err.to_string()))?;

    perform_clike_format(vm, format_string_text)

}
```

vm/src/obj/objstr.rs
```diff
@@ -429,6 +430,29 @@ impl PyString {
         }
     }
 
+    #[pymethod(name = "__mod__")]
+    fn modulo(&self, values: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        let format_string_text = &self.value;
+
+        match CFormatString::from_str(format_string_text.as_str()) {
```
windelbouwman: the `.as_str()` suffix should not be required, since `&str` is a good type which is often compatible.

vm/src/cformat.rs
```diff
@@ -0,0 +1,357 @@
...
+    Literal(String),
+    Spec(CFormatSpec),
+}
+
+#[derive(Debug, PartialEq)]
+pub struct CFormatString {
+    pub format_parts: Vec<(usize, CFormatPart)>,
+}
+
+impl CFormatString {
```
windelbouwman: This "class" contains mostly static methods. This impl block might not be required, and we can use simple functions for the parsing, right?

vm/src/cformat.rs
```diff
@@ -0,0 +1,357 @@
...
+                        index: index + consumed,
+                    }
+                })?;
+        }
+
+        Ok(CFormatString {
+            format_parts: parts,
+        })
+    }
+}
```
windelbouwman: I propose to add some (perhaps 3 or 7) test cases in this file as unittests.

windelbouwman:
@michelhe thanks for implementing this! The change looks good! I left some comments / ideas.

vm/src/obj/objstr.rs
michelhe: Thanks for the good advice!
I had in mind though that the cformat crate will only do parsing and formatting, and won't have to do with any python internals by itself.

michelhe:
Thanks for the comments :)
Can you please also check that i'm using `clone` for PyObjectRef correctly?

michelhe:
> @windelbouwman Thanks for the comments :)
> Can you please also check that i'm using `clone` for PyObjectRef correctly?

Woops, I haven't pushed these changes yet ^^

michelhe:
I've rebased on the current master. It seems that on my `master` bitflags was still a dependency.

vm/src/cformat.rs
michelhe: done :)

vm/src/cformat.rs
michelhe: Yep, thanks

vm/src/obj/objstr.rs
windelbouwman: You're right that the parsing of format strings does not require any python objects. Are you going to implement the change I suggested above?

michelhe: Yeah, I've pushed it

windelbouwman:
This change looks good to me! One thing left to do is add some python snippets as python test cases. @michelhe do you like to do this in this pull request, or do you first want to merge this work and then add some test snippets?

vm/src/obj/objstr.rs
```diff
@@ -1072,6 +1090,200 @@ fn call_object_format(vm: &VirtualMachine, argument: PyObjectRef, format_spec: &
...
+                    Some(key) => {
+                        // TODO: change the KeyError message to match the one in cpython
+                        call_getitem(vm, &values_obj, &vm.ctx.new_str(key.to_string()))?
+                    }
+                    None => {
+                        let mut elements = objtuple::get_value(&values_obj)
+                            .into_iter()
+                            .skip(auto_index);
+
+                        let mut try_quantity_from_tuple = |q: &mut Option<CFormatQuantity>| {
```
windelbouwman: Lambda functions are usually for small functions. This is a pretty big function, so I would suggest to create a function for this. This will have as a side effect that this whole function becomes somewhat smaller.

vm/src/obj/objstr.rs
```diff
@@ -1072,6 +1090,200 @@ fn call_object_format(vm: &VirtualMachine, argument: PyObjectRef, format_spec: &
...
+
+        // convert `values_obj` to a new tuple if it's not a tuple
+        if !objtype::isinstance(&values_obj, &vm.ctx.tuple_type()) {
+            vm.ctx.new_tuple(vec![values_obj.clone()])
+        } else {
+            values_obj.clone()
+        }
+    };
+
+    let mut auto_index: usize = 0;
```
windelbouwman: Keeping this index seems somewhat weird to me. I would expect some sort of a `zip` operation over the format parts and the python objects to turn them into strings, then join all strings together. Off course it is not as simple as that since we have string literals in between, but maybe there is a way to store a sequence of pairs or string literals.

windelbouwman: Maybe creating a simple enum would do the trick:
```rust
enum StringPart {
  String,
  FormattedObject(FormatString, PyObject),
}
```

vm/src/obj/objstr.rs
michelhe: The reason I used a closure is to take advantage of the access it has to the context of the bigger function. Making it it's own function will require to pass it lots of arguments. 
Have a better idea? :/

michelhe:
> This change looks good to me! One thing left to do is add some python snippets as python test cases. @michelhe do you like to do this in this pull request, or do you first want to merge this work and then add some test snippets?

I prefer to implement the python snippets on this PR.
Also, there is the missing floats formatting which I think can wait further down the line, and as a temporary placeholder we can simply delegate %f specifiers to the float's `repr`

vm/src/obj/objstr.rs
michelhe: > Keeping this index seems somewhat weird to me. I would expect some sort of a `zip` operation over the format parts and the python objects to turn them into strings, then join all strings together. Off course it is not as simple as that since we have string literals in between, but maybe there is a way to store a sequence of pairs or string literals.

It's indeed not as simple because the `values` object can be a mapping type, and not just a tuple.
I'll think of something tomorrow.

michelhe: @windelbouwman 
I don't have any elegant solution at the moment, the problems with zipping are
* `try_update_quantity_from_tuple` (asterik in format strings)
* `values_obj` might be a dict so it can't be converted to an iterator

vm/src/obj/objstr.rs
```diff
@@ -1072,6 +1090,210 @@ fn call_object_format(vm: &VirtualMachine, argument: PyObjectRef, format_spec: &
...
+                // try to get the object
+                let obj: PyObjectRef = match &format_spec.mapping_key {
+                    Some(key) => {
+                        // TODO: change the KeyError message to match the one in cpython
+                        call_getitem(vm, &values_obj, &vm.ctx.new_str(key.to_string()))?
+                    }
+                    None => {
+                        let mut elements = objtuple::get_value(&values_obj).into_iter().skip(tuple_index);
+
+                        tuple_index = try_update_quantity_from_tuple(
```
windelbouwman: I would have a suggestion to overcome the index variable. Pseudocode is like this:
```rust
let element_iterator = tuple_get_elements().into_iter();  // Do this only once, do not use the index
for part in format_parts {
    match part {
         CFormatPart(spec) => {
                let obj = match element_iterator.next() {   // Get next element from object iterator.
                    Some(obj) => { do_format(spec, obj) },
                    None => { panic!("Not enough elements!"); }
                }
         }
    }
}

// Check here is the element_iterator is exhausted.
```

vm/src/obj/objstr.rs
```diff
@@ -1072,6 +1090,210 @@ fn call_object_format(vm: &VirtualMachine, argument: PyObjectRef, format_spec: &
...
+        .format_parts
+        .iter()
+        .any(|(_, part)| CFormatPart::has_key(part))
+        && format_string
+            .format_parts
+            .iter()
+            .filter(|(_, part)| CFormatPart::is_specifier(part))
+            .all(|(_, part)| CFormatPart::has_key(part));
+
+    let values_obj = if mapping_required {
```
windelbouwman: If `mapping_required` is used a lot, this might be a hint that there are actually two functions required. As a rule of thumb, if you have a function with a boolean as argument, just create two functions for the true and false case.

michelhe: Let's merge the current work in.
I will re-work the entire parsing-formatting later.
I have in mind to make CFormatString into an iterator and try to zip it with the tuple/dict like you suggested.

Issue #102, opened 2018-08-29
Compilation broken on rust nightly


OddBloke:
This is due to lalrpop's dependency on petgraph; a new version of petgraph has been released to address the issue and fixing it in lalrpop is being tracked at https://github.com/lalrpop/lalrpop/issues/400

OddBloke:
I've also opened up https://github.com/RustPython/RustPython/pull/101 to make it easier to catch issues like this in future.

PR #1034, opened 2019-06-14
Allow compiling Python code to bytecode at compile-time

coolreader18:
As discussed in #102.

I went with just json in a string literal and `serde::from_str` after trying 1) a serde serializer that outputs rust source code, which didn't work with certain types that didn't serialize to a direct rust representation of them, and 2-2.5) outputting `serde_json::Value`s, either with the `serde_json::json!` macro which didn't work due to too much recursion, and then just outputting what `json!` would have expanded to anyway, which took 20 minutes and then crashed rustc with a SIGKILL. So I figured that if a json string literal works and compiles, then we might as well go with that.

### DIFF
--- a/derive/src/compile_bytecode.rs
+++ b/derive/src/compile_bytecode.rs
@@ -51,11 +51,17 @@ struct PyCompileInput {
     metas: Vec<Meta>,
 }
 
+struct PyCompileResult {
+    code_obj: CodeObject,
+    lazy_static: bool,
+}
+
 impl PyCompileInput {
-    fn compile(&self) -> Result<CodeObject, Diagnostic> {
+    fn compile(&self) -> Result<PyCompileResult, Diagnostic> {
         let mut source_path = None;
         let mut mode = None;
         let mut source: Option<CompilationSource> = None;
+        let mut lazy_static = false;
 
         fn assert_source_empty(source: &Option<CompilationSource>) -> Result<(), Diagnostic> {
             if let Some(source) = source {
@@ -108,11 +114,16 @@ impl PyCompileInput {
                         });
                     }
                 }
+                Meta::Word(ident) => {
+                    if ident == "lazy_static" {
+                        lazy_static = true;
+                    }
+                }
                 _ => {}
             }
         }
 
-        source
+        let code_obj = source
             .ok_or_else(|| {
                 Diagnostic::span_error(
                     self.span.clone(),
@@ -122,7 +133,12 @@ impl PyCompileInput {
             .compile(
                 &mode.unwrap_or(compile::Mode::Exec),
                 source_path.unwrap_or_else(|| "frozen".to_string()),
-            )
+            )?;
+
+        Ok(PyCompileResult {
+            code_obj,
+            lazy_static,
+        })
     }
 }
 
@@ -140,16 +156,32 @@ impl Parse for PyCompileInput {
 pub fn impl_py_compile_bytecode(input: TokenStream2) -> Result<TokenStream2, Diagnostic> {
     let input: PyCompileInput = parse2(input)?;
 
-    let code_obj = input.compile()?;
+    let PyCompileResult {
+        code_obj,
+        lazy_static,
+    } = input.compile()?;
+
     let bytes = bincode::serialize(&code_obj).expect("Failed to serialize");
     let bytes = LitByteStr::new(&bytes, Span::call_site());
 
     let output = quote! {
         ({
-            use bincode;
-            bincode::deserialize(#bytes).expect("Deserializing CodeObject failed")
+            use ::bincode;
+            bincode::deserialize::<::rustpython_vm::bytecode::CodeObject>(#bytes).expect("Deserializing CodeObject failed")
         })
     };
 
-    Ok(output)
+    if lazy_static {
+        Ok(quote! {
+            ({
+                use ::lazy_static::lazy_static;
+                lazy_static! {
+                    static ref STATIC: ::rustpython_vm::bytecode::CodeObject = #output;
+                }
+                &*STATIC
+            })
+        })
+    } else {
+        Ok(output)
+    }
 }
--- a/derive/src/compile_bytecode.rs
+++ b/derive/src/compile_bytecode.rs
@@ -1,3 +1,18 @@
+//! Parsing and processing for this form:
+//! ```ignore
+//! py_compile_input!(
+//!     // either:
+//!     source = "python_source_code",
+//!     // or
+//!     file = "file/path/relative/to/$CARGO_MANIFEST_DIR",
+//!
+//!     // the mode to compile the code in
+//!     mode = "exec", // or "eval" or "single"
+//!     // the path put into the CodeObject, defaults to `None`
+//!     source_path = "frozen",
+//! )
+//! ```
+
 use crate::{extract_spans, Diagnostic};
 use bincode;
 use proc_macro2::{Span, TokenStream as TokenStream2};
@@ -46,6 +61,7 @@ impl CompilationSource {
     }
 }
 
+/// This is essentially just a comma-separated list of Meta nodes, aka the inside of a MetaList.
 struct PyCompileInput {
     span: Span,
     metas: Vec<Meta>,
--- a/derive/src/compile_bytecode.rs
+++ b/derive/src/compile_bytecode.rs
@@ -1,13 +1,13 @@
 use super::Diagnostic;
-use proc_macro2::TokenStream as TokenStream2;
+use bincode;
+use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::quote;
 use rustpython_compiler::{bytecode::CodeObject, compile};
-use serde_json;
 use std::env;
 use std::fs;
 use std::path::{Path, PathBuf};
 use syn::parse::{Parse, ParseStream, Result as ParseResult};
-use syn::{self, parse2, Ident, Lit, Meta, MetaList, NestedMeta, Token};
+use syn::{self, parse2, Ident, Lit, LitByteStr, Meta, MetaList, NestedMeta, Token};
 
 struct BytecodeConst {
     ident: Ident,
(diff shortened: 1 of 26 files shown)


coolreader18:
Now I'm using bincode, which is hopefully as efficient as you can get for this situation.

derive/src/compile_bytecode.rs
```diff
@@ -0,0 +1,135 @@
+use super::Diagnostic;
```
windelbouwman: What does the name `derive` of the package mean? Maybe we could create a seperate package called `python_precompiler` or something like that, which can be used to precompile python code into bytecode.

parser/src/lexer.rs
```diff
@@ -74,8 +74,8 @@ pub enum LexicalErrorType {
 
 #[derive(Clone, Debug, Default, PartialEq, Serialize, Deserialize)]
 pub struct Location {
-    row: usize,
-    column: usize,
+    pub row: usize,
```
windelbouwman: These have been public before, but then made private with the accessor function `get_row` and `get_column`.

derive/src/compile_bytecode.rs
coolreader18: The `_derive` suffix is sort of a standard for proc macro crates because previously the only procedural macros were custom derives. We have non-derive proc macros already, and I was thinking maybe we could rename the crate to `rustpython_macros`. I don't think that proc macros in the same crate as other code is going to happen very soon, so I'd say it's not really temporary.

parser/src/lexer.rs
coolreader18: Ah yeah, that's a leftover from the syn serde serializer.

vm/src/frozen.rs
```diff
@@ -1,25 +1,20 @@
...
-    let mut modules = HashMap::new();
-    modules.insert("__hello__".to_string(), HELLO);
-    modules.insert("_frozen_importlib".to_string(), IMPORTLIB_BOOTSTRAP);
-    modules.insert(
-        "_frozen_importlib_external".to_string(),
-        IMPORTLIB_BOOTSTRAP_EXTERNAL,
+",
     );
-    modules
+    static ref IMPORTLIB_BOOTSTRAP = file("../Lib/importlib/_bootstrap.py");
```
windelbouwman: This `file` function call is pretty mind blowing :). It is a very magical function, I still need time to process how this works, but anyway, I would like to change the name of the "function" in something more descriptive like `compile_file` or `include_file_as_bytecode!`.

derive/src/compile_bytecode.rs
```diff
@@ -0,0 +1,135 @@
...
+            &source,
+            &mode.unwrap_or(compile::Mode::Exec),
+            source_path.unwrap_or_else(|| "frozen".to_string()),
+        )
+        .map_err(|err| err_span!(source_lit, "Compile error: {}", err))
+    }
+}
+
+impl Parse for BytecodeConst {
+    /// Parse the form `static ref IDENT = metalist(...);`
```
windelbouwman: I suggest to use a normal macro to introduce a marker token for the second stage.

First, use a macro something like this: `include_file_as_bytecode!("my_file.py")` which will expand to `compile_file_sentinel "my_file.py"` then the matching parse can use the `compile_file_sentinel` as a token to match on, so that users of the macro do not need to use `static ref IDENT ...`, but this will then also work: `let bytecode = include_file_as_bytecode!("other_file");`!

Would this be possible? Did I explain this idea well enough?

derive/src/compile_bytecode.rs
```diff
@@ -0,0 +1,135 @@
...
+                            Lit::Str(s) => s.value(),
+                            _ => bail_span!(name_value.lit, "source_path must be string"),
+                        })
+                    }
+                }
+                _ => {}
+            }
+        }
+
+        let source = if meta.ident == "file" {
```
windelbouwman: Rather than implementing this as a rust syntax extension (which is very cool by the way;) . The code might be simpler to understand if we use a procedural macro here. The syntax extension seems overkill to me, or did I miss something?

derive/src/compile_bytecode.rs
windelbouwman: Temporary is a fun word, and it happens often that the temporary solution becomes the final one. Agreed to keep it like this. If derive is the convention, let's stick with that.

derive/src/compile_bytecode.rs
coolreader18: I would have done this in the first place, but the problem is that procedural macros can't be used in expression position. There is a way to fix that, see: https://github.com/dtolnay/proc-macro-hack, but I wasn't sure if that was necessary. I suppose we could have `rustpython_vm`  be the declaration crate? Should I try that?

coolreader18:
I think this is a lot cleaner, the syntax now is:
```rust
py_compile_bytecode!(
    // if you include this, it returns an `&'static CodeObject`, otherwise it just
    // makes a new CodeObject each time
    lazy_static,
    // either:
    source = "python_source_code",
    // or
    file = "file/path/relative/to/$CARGO_MANIFEST_DIR",
    // the mode to compile the code in
    mode = "exec", // or "eval" or "single"
    // the path put into the CodeObject, defaults to "frozen"
    source_path = "frozen",
)
```

vm/src/import.rs
```diff
@@ -83,25 +81,30 @@ pub fn import_file(
...
-    // trace!("Code object: {:?}", code_obj);
+    import_codeobj(vm, module_name, code_obj)
+}
 
+pub fn import_codeobj(vm: &VirtualMachine, module_name: &str, code_obj: CodeObject) -> PyResult {
     let attrs = vm.ctx.new_dict();
     attrs.set_item("__name__", vm.new_str(module_name.to_string()), vm)?;
+    let file_path = &code_obj.source_path;
     if !file_path.starts_with("frozen") {
         // TODO: Should be removed after precompiling frozen modules.
```
palaviv: The `TODO` should probably be removed if  we keep the way we find this is a frozen module the same.

palaviv:
> Now I'm using bincode, which is hopefully as efficient as you can get for this situation.

Should we change the marshal to used bincode as well? Would this be the best option for `.pyc` files as well?

palaviv:
> I think this is a lot cleaner, the syntax now is:
> 
> ```rust
> py_compile_bytecode!(
>     // if you include this, it returns an `&'static CodeObject`, otherwise it just
>     // makes a new CodeObject each time
>     lazy_static,
>     // either:
>     source = "python_source_code",
>     // or
>     file = "file/path/relative/to/$CARGO_MANIFEST_DIR",
>     // the mode to compile the code in
>     mode = "exec", // or "eval" or "single"
>     // the path put into the CodeObject, defaults to "frozen"
>     source_path = "frozen",
> )
> ```

I am not sure I understand the purpose of `lazy_static`. This seems to always be set. At what cases you would not set it?

derive/src/compile_bytecode.rs
windelbouwman: Yes, that would be good if the resulting macro is easier to understand.

vm/src/vm.rs
```diff
@@ -53,7 +53,7 @@ pub struct VirtualMachine {
     pub frames: RefCell<Vec<FrameRef>>,
     pub wasm_id: Option<String>,
     pub exceptions: RefCell<Vec<PyObjectRef>>,
-    pub frozen: RefCell<HashMap<String, &'static str>>,
+    pub frozen: RefCell<HashMap<&'static str, &'static bytecode::CodeObject>>,
```
windelbouwman: What is the reason for the static lifetime parameter? Is a static lifetime good practice? Can't we use `HashMap<&str, bytecode::CodeObject>`?

vm/src/frozen.rs
```diff
@@ -1,25 +1,19 @@
...
-    modules.insert("__hello__".to_string(), HELLO);
-    modules.insert("_frozen_importlib".to_string(), IMPORTLIB_BOOTSTRAP);
-    modules.insert(
-        "_frozen_importlib_external".to_string(),
-        IMPORTLIB_BOOTSTRAP_EXTERNAL,
-    );
-    modules
+pub fn get_module_inits() -> HashMap<&'static str, &'static CodeObject> {
+    hashmap! {
+        "__hello__" => py_compile_bytecode!(
```
windelbouwman: This looks pretty neat!

derive/src/compile_bytecode.rs
```diff
@@ -0,0 +1,187 @@
+use crate::{extract_spans, Diagnostic};
```
windelbouwman: Perhaps you could add here some documentation text describing the purpose and usage of this code. Something like how you described the macro usage in this pull request would be handy to keep as a reference in our source code as well.

coolreader18:
I think I've decided that using lazy_static! is sort of premature optimization, `bincode` is fast enough in deserializing and we rarely use more than one `VirtualMachine` in a program, so I'm just going to make frozen be a `HashMap<String, CodeObject>`.

> Should we change the marshal to used bincode as well? Would this be the best option for .pyc files as well?

I think marshal already is using bincode, but I think the ultimate goal is to have our bytecode be the same as CPython's (@windelbouwman correct me if I'm wrong), so using bincode might be too flakey for that, as its format changes with pretty much any change we make to the data structure definition. It might be a good temporary solution though.

vm/src/import.rs
coolreader18: What should the behavior be for this? Only set __file__ if not a frozen module? Cause doing `frozen *` as the module name does seem sort of hacky.

palaviv: Yes. Frozen modules should not have a `__file__` attribute. This can be solved in a later stage but the `TODO` should be changed.

coolreader18: I just changed CodeObject.source_path to an `Option` in another branch and I'm going to open a PR for it once this merges.

windelbouwman:
> I think the ultimate goal is to have our bytecode be the same as CPython's (@windelbouwman correct me if I'm wrong)

I will correct you here ;), the CPython bytecode is a python internal and ever changing bytecode. In rustpython we will have similar, but not compatible bytecode. Writing a compiler from one form of bytecode to the other would be doable though, so if we need this, we can make it. However, the bytecode of CPython is a moving target, so I suggest to not be compatible. I like the way we use bincode to serialize our codeobject, it works great!

PR #1035, opened 2019-06-15
Append newline to scripts too, if not exists yet

mkurnikov:


src/main.rs
```diff
@@ -74,14 +74,19 @@ fn main() {
     handle_exception(&vm, result);
 }
 
-fn _run_string(vm: &VirtualMachine, source: &str, source_path: String) -> PyResult {
+fn _run_string(vm: &VirtualMachine, mut source: String, source_path: String) -> PyResult {
+    // This works around https://github.com/RustPython/RustPython/issues/17
```
windelbouwman: I'm unsure if we should keep adding work arounds. The actual fix in the lexer might be easy to achieve by adding a newline or two over there and emitting the indentation levels still in process. If you wish, you can have a look at this, I think it is not much more work than this workaround.

mkurnikov: Totally makes sense, I'll look at it tomorrow.

mkurnikov: @windelbouwman What do you mean by "emitting the indentation levels still in process"? Is it not just the same "add newline character if not exists", but in the https://github.com/RustPython/RustPython/blob/master/parser/src/lexer.rs#L143 now?

windelbouwman: This could be a nice place yo inject some extra newlines. Just append several newlines characters in this iterator chain, and it should be good.

What I mean by emitting indentation levels is that the lexer keeps track of indentation levels here https://github.com/RustPython/RustPython/blob/master/parser/src/lexer.rs#L53 which must be emitted as dedent tokens. Newline characters trigger this, depending upon the new indentation level.

Just appending extra newlines (at the point you indicated) to the source characters is simpler.

PR #1037, opened 2019-06-16
Make CodeObject.source_path an Option<String>

coolreader18:


cthulahoops:
I'm surprised by this. When is there no source path?

coolreader18:
With frozen modules, try this with cpython:
```py
import __hello__
__hello__.__file__
```
The previous solution was to check if the module name started with `"frozen"`, which was sorta hacky.

palaviv:
I am not sure this is the best solution as now our stacktrace for frozen modules look like this:
```
File "None", line 1093, in __import__
  File "None", line 1014, in _gcd_import
  File "None", line 991, in _find_and_load
  File "None", line 975, in _find_and_load_unlocked
  File "None", line 686, in _load_unlocked
  File "None", line 677, in _load_unlocked
  File "None", line 671, in _load_unlocked
  File "None", line 774, in exec_module
  File "None", line 925, in get_code
  File "None", line 923, in get_code
  File "None", line 1009, in _cache_bytecode
  File "None", line 1038, in set_data
  File "None", line 1034, in set_data
  File "None", line 139, in _write_atomic
  File "None", line 131, in _write_atomic
```
I think that a better solution will add some Enum to the module that will indicate if it is `frozen`/ `built-in`/`regular`. What do you think?

coolreader18:
That does seem like a better idea, I think I might revert this change then.

coolreader18:
#1047

PR #1039, opened 2019-06-18
[WASM] Add a JsValue class

### LINKED ISSUE #808: How should JS values bind into Python?

coolreader18:
Attempt number 2, after #848. Here's the design for this implementation:

- `_js` module - contains the `JsValue` class which wraps a `wasm_bindgen::JsValue` and provides methods like `get_prop()`, `set_prop()`, and `typeof()`.
- `_window` - provides a single export, a `JsValue` wrapping `window`. Can be disabled by the JS code using `rustpython_wasm`.
- `browser` - a (eventually) Python frozen module that imports `window` from `_window` and uses it to provide all the functionality that the `browser` native module does now.
- (eventually) `jslib` - a frozen module that provides a JsProxy similar in functionality to the `JsValue` discussed in #808 among some other things (like a Promise wrapper).

Closes #808.

### DIFF
--- a/vm/src/lib.rs
+++ b/vm/src/lib.rs
@@ -66,3 +66,8 @@ mod vm;
 pub use self::exceptions::print_exception;
 pub use self::vm::VirtualMachine;
 pub use rustpython_compiler::*;
+
+#[doc(hidden)]
+pub mod __exports {
+    pub use bincode;
+}
--- a/vm/src/stdlib/mod.rs
+++ b/vm/src/stdlib/mod.rs
@@ -34,6 +34,7 @@ use crate::pyobject::PyObjectRef;
 pub type StdlibInitFunc = Box<dyn Fn(&VirtualMachine) -> PyObjectRef>;
 
 pub fn get_module_inits() -> HashMap<String, StdlibInitFunc> {
+    #[allow(unused_mut)]
     let mut modules = hashmap! {
         "ast".to_string() => Box::new(ast::make_module) as StdlibInitFunc,
         "binascii".to_string() => Box::new(binascii::make_module),
--- a/derive/src/compile_bytecode.rs
+++ b/derive/src/compile_bytecode.rs
@@ -163,7 +163,7 @@ pub fn impl_py_compile_bytecode(input: TokenStream2) -> Result<TokenStream2, Dia
 
     let output = quote! {
         ({
-            use ::bincode;
+            use ::rustpython_vm::__exports::bincode;
             bincode::deserialize::<::rustpython_vm::bytecode::CodeObject>(#bytes)
                 .expect("Deserializing CodeObject failed")
         })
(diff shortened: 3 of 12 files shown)


coolreader18:
Ok, I think this is good for review as an initial implementation; everything works that did before, and although it's not all converted to the new system, I think that's best done in follow-up PRs.

vm/src/lib.rs
```diff
@@ -66,3 +66,8 @@ mod vm;
 pub use self::exceptions::print_exception;
 pub use self::vm::VirtualMachine;
 pub use rustpython_compiler::*;
+
+#[doc(hidden)]
+pub mod __exports {
```
windelbouwman: I think the same can be achieved by using `extern crate bincode`. Cargo will make sure that both the vm and the wasm crate use the same version of the bincode crate, so no need to import the bincode crate via the `rustpython_vm`.

vm/src/stdlib/mod.rs
```diff
@@ -34,6 +34,7 @@ use crate::pyobject::PyObjectRef;
 pub type StdlibInitFunc = Box<dyn Fn(&VirtualMachine) -> PyObjectRef>;
 
 pub fn get_module_inits() -> HashMap<String, StdlibInitFunc> {
+    #[allow(unused_mut)]
```
windelbouwman: Why was this added? We're returning an immutable HashMap here right?

windelbouwman:
@rmliddle could you take a view on this change?

vm/src/stdlib/mod.rs
coolreader18: A little bit downwards there are some `#[cfg]`s that conditionally add modules to the map, and if your target matches neither of those `cfg`s, you get an unused mut warning.

windelbouwman: Ah, I understand now. Thanks!

vm/src/lib.rs
coolreader18: I did this here because this pattern is used by other macros in the ecosystem, here are a few examples:
- https://docs.rs/serde/1.0.92/src/serde/lib.rs.html#236-242
- https://docs.rs/wasm-bindgen/0.2.47/src/wasm_bindgen/lib.rs.html#801
- https://docs.rs/darling/0.9.0/src/darling/lib.rs.html#84-95

Issue #1025, opened 2019-06-09
unicode literals not work 


yanganto:
## Feature
In Python source code, specific Unicode code points can be written using the `\u` escape sequence, which is followed by four hex digits giving the code point. The `\U` escape sequence is similar but expects eight hex digits, not four.

cpython
```python
Python 3.7.3 (default, Mar 26 2019, 21:43:19)
[GCC 8.2.1 20181127] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> "\u0037"
'7'
>>> "\xac"
'¬'
```
RustPython
```python
Welcome to the magnificent Rust Python 0.0.1 interpreter 😱 🖖
>>>>> "\u0037"
'\\u0037'
>>>>> "\xac"
'\\xac'
```

## Python Documentation
https://docs.python.org/3/howto/unicode.html#unicode-literals-in-python-source-code

windelbouwman:
This is a lexer issue. I think the lexer only takes ascii into account at this point for strings, or it does not replace escape code properly.

yanganto:
May I implement this feature?  It is essential if I want to implement some the test for `str.isprintable` #1024

windelbouwman:
Sure! Give it your best shot :)

ProgVal:
Hi, I don't really know how RustPython works, but unicode literals are tricky to get right (if your goal is to be fully compatible with CPython).

CPython supports unpaired surrogate code points in unicode literals, which are invalid UTF-8.

In the toy rust interpreter I wrote a while ago, I use the `wtf8` crate to store this ill-formed UTF-8: https://github.com/ProgVal/rust-python-parser/commit/cd6159b70a546a0f4780a295d74c27e640319765

See also: https://simonsapin.github.io/wtf-8/

Have fun! :)

michelhe:
> This is a lexer issue. I think the lexer only takes ascii into account at this point for strings, or it does not replace escape code properly.

True, note that other types of escaping like `\x` are not implemented.

yanganto:
@ProgVal  Thank you,  I almost there.

PR #1043, opened 2019-06-18
support unicode literal

yanganto:
support Unicode literal as same as CPython #1025 
- support Unicode literal `\x` with 2 digits
- support Unicode literal `\u` with 4 digits
- support Unicode literal `\U` with 8 digits

### DIFF
--- a/parser/src/lexer.rs
+++ b/parser/src/lexer.rs
@@ -13,6 +13,7 @@ use std::collections::HashMap;
 use std::str::FromStr;
 use unic_emoji_char::is_emoji_presentation;
 use unicode_xid::UnicodeXID;
+use wtf8;
 
 #[derive(Clone, Copy, PartialEq, Debug)]
 struct IndentationLevel {
@@ -67,6 +68,7 @@ pub struct LexicalError {
 #[derive(Debug)]
 pub enum LexicalErrorType {
     StringError,
+    UnicodeError,
     NestingError,
     UnrecognizedToken { tok: char },
     OtherError(String),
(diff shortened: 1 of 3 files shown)


yanganto:
The error handler needs to check.

parser/src/lexer.rs
```diff
@@ -456,6 +458,27 @@ where
         }
     }
 
+    fn uncicode_literal(&mut self, literal_number: usize) -> Result<char, LexicalError> {
+        let location = self.get_pos();
+        let mut it: Vec<u32> = Vec::new();
```
windelbouwman: I suggest to pick a better name for the variable `it`. Perhaps `digits` would be better?

parser/src/lexer.rs
```diff
@@ -513,6 +536,18 @@ where
...
+                                Ok(c) => string_content.push(c),
+                                Err(e) => return Err(e),
+                            },
+                            Some('U') => match self.uncicode_literal(8) {
+                                Ok(c) => string_content.push(c),
+                                Err(e) => return Err(e),
+                            },
+                            Some('x') if !is_bytes => match self.uncicode_literal(2) {
+                                Ok(c) => string_content.push(c),
+                                Err(e) => return Err(e),
```
windelbouwman: The pattern `Err(e) => return Err(e)` can be short circuit like this:
```rust
Some('x') if !is_bytes => string_content.push(self.unicode_literal(2)?),
```

parser/src/lexer.rs
```diff
@@ -456,6 +458,27 @@ where
...
+                None => {
+                    return Err(LexicalError {
+                        error: LexicalErrorType::UnicodeError,
+                        location: location,
+                    })
+                }
+            }
+        }
+        let mut p: u32 = 0u32;
+        for i in 1..=literal_number {
```
windelbouwman: In this case, I suggest to use `enumerate` to iterate over the parsed digits and construct the numerical value.
```rust
for (i, d) in enumerate it {
   p += d << i*4;
}
```

windelbouwman:
@yanganto great job! I left some comments for improvements, but the change look good!

parser/src/lexer.rs
michelhe: Will the compiler optimize it out though?

michelhe: Anyway, we can even do something like this

```rust
let p = it.iter().enumerate().fold(|acc, (i, d) acc + d << i *4)
```
But only if this all line ends up being optimized to a loop

parser/src/lexer.rs
yanganto: Thanks.  I refactor this with `?`.

parser/src/lexer.rs
yanganto: Thanks for the suggestion,  I remove the `it` and just put digit into `p`.

parser/src/lexer.rs
yanganto: The `it` is removed, but I learn the powerful `fold` method, I will try to use it next time.    Thank you. @michelhe

PR #1044, opened 2019-06-18
Move PyObject de/serialization to its own module

coolreader18:
Also use [serde-wasm-bindgen](https://github.com/cloudflare/serde-wasm-bindgen) for de/serialization for converting between JS and PyObjects in `rustpython_wasm` instead of using JSON as an intermediary.


windelbouwman:
What is the reason for this refactor? Could we re-use the serialization / deserialization code in the marshall module?

Btw, I think it is a good refactor. Especially having the wasm code not importing into stdlib code is a good thing.

vm/src/ser_de.rs
```diff
@@ -0,0 +1,199 @@
...
+    {
+        Ok(self.vm.ctx.new_bool(value))
+    }
+
+    fn visit_seq<A>(self, mut access: A) -> Result<Self::Value, A::Error>
+    where
+        A: serde::de::SeqAccess<'de>,
+    {
+        let mut seq = Vec::with_capacity(access.size_hint().unwrap_or(0));
+        while let Some(value) = access.next_element_seed(self.clone())? {
```
windelbouwman: It's probably a later topic, since you just copy pasted this code from json.rs, but what happens with a nested data structure like this?
```python
L = []
L.append(L)
serialize(L) --> BOOM?
```

vm/src/ser_de.rs
```diff
@@ -0,0 +1,199 @@
+use std::fmt;
```
windelbouwman: It would be neat if we could use this same code for `marshall.rs`, but then use bincode instead of json!

windelbouwman: I'm not sure about the filename `ser_de.rs`. I would suggest `serialization.rs`. Other ideas are welcome!

wasm/lib/src/convert.rs
```diff
@@ -130,10 +132,9 @@ pub fn py_to_js(vm: &VirtualMachine, py_obj: PyObjectRef) -> JsValue {
             view.slice(0, bytes.len() as u32).into()
         }
     } else {
-        match vm.serialize(&py_obj) {
-            Ok(json) => js_sys::JSON::parse(&json).unwrap_or(JsValue::UNDEFINED),
-            Err(_) => JsValue::UNDEFINED,
-        }
+        rustpython_vm::ser_de::PyObjectSerializer::new(vm, &py_obj)
```
windelbouwman: Instead of creating the new struct `PyObjectSerializer`, consider creating a function `serialize(vm, &py_obj, &serde_wasm_bindgen::Serializer::new())`. It's probably best if `PyObjectSerializer` is a private struct which does not leave the `ser_de` module.

vm/src/ser_de.rs
coolreader18: Right now, a segfault. CPython errors with a `ValueError: Circular reference detected`.

vm/src/ser_de.rs
coolreader18: What about `py_serde.rs`?

wasm/lib/src/convert.rs
coolreader18: I think it's better to do both, because when we use `serde::to_string()` in `json`, it expects a type that's `Deserialize`, and the alternative is to manually pass an `&mut String` to `serde::Serializer::new()`, which seems uglier.

vm/src/py_serde.rs
```diff
@@ -0,0 +1,222 @@
...
+        // implement those for now
+        Ok(self.vm.ctx.new_int(value))
+    }
+
+    fn visit_u64<E>(self, value: u64) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        // The JSON deserializer always uses the i64/u64 deserializers, so we only need to
+        // implement those for now
```
silmeth: The default implementations of `visit_u32`, `visit_u16`, etc. forward to `visit_u64` anyway, so it’s probably good as is even for other deserializers that might use those.

coolreader18:
Merging, more work can be done on the serde implementations in separate PRs.

Issue #55, opened 2018-08-09
Support relative imports


OddBloke:
#54 implements most (perhaps all?) of the import syntax, but it doesn't include support for relative imports (e.g. `from .foo import bar`, `from ..spam import eggs`).

PR #1050, opened 2019-06-21
Support relative import

palaviv:
Should solve #55.


coolreader18:
I'm not sure why WASM's not passing; I restarted the build and im hoping it was just a fluke.

palaviv:
When I run the test locally it passes.

palaviv:
> I'm not sure why WASM's not passing; I restarted the build and im hoping it was just a fluke.

It is not a fluke for sure...
I am starting to debug the test to see the problem. I think we should find a better test for the wasm then the selenium.

vm/src/frame.rs
```diff
@@ -912,7 +912,9 @@ impl Frame {
             .iter()
             .map(|symbol| vm.ctx.new_str(symbol.to_string()))
             .collect();
-        let module = vm.import(module, &vm.ctx.new_tuple(from_list))?;
+        let level = module.chars().take_while(|char| *char == '.').count();
```
windelbouwman: Instead of splitting of the level here, it's better to modify the parser to be able to handle this such that have AST nodes with a level inside. This way, the level is already known at parse time, and is passed in as a second argument. See also: https://greentreesnakes.readthedocs.io/en/latest/nodes.html#imports

windelbouwman: This might be a big change tough, so it can be done in a follow up pull request.

palaviv: Sure. I will do that in a different PR.

palaviv:
I am not very familiar with javascript... The only thing I get in the error message is `[object Opaque]`.
I do not actually import anything in my test and especially nothing relative...
If anyone has any idea why is this failing or what I can do to understand why I would appreciate any help :sos:

windelbouwman:
Debugging travis can be a pain. Is the issue not present on the master branch?

windelbouwman:
Did you try to reproduce the scenario locally?

coolreader18:
Try rebasing to master, #1039 added an import function specifically for wasm that might fix the issue.

coolreader18:
@windelbouwman https://github.com/RustPython/RustPython/pull/1050#issuecomment-504637581

palaviv:
> Did you try to reproduce the scenario locally?

It passes and everything look good on the demo on locally.



> Try rebasing to master, #1039 added an import function specifically for wasm that might fix the issue.

Already did... I hoped it will be resolved by that as well.

I will try removing stuff from the change to pinpoint the problem with the WASM.

palaviv:
> I will try removing stuff from the change to pinpoint the problem with the WASM.

The problematic line is the copy of the local variables as parameter of `__import__`. My guess is that I am doing something wrong there but I am not sure what... Could one of you have a look @windelbouwman @coolreader18?

coolreader18:
That's so strange, because as far as I can tell based on the tests that `import` function isn't being called at all.

vm/src/vm.rs
```diff
@@ -314,7 +314,7 @@ impl VirtualMachine {
                     func,
                     vec![
                         self.ctx.new_str(module.to_string()),
-                        self.get_locals().clone().as_object().clone(),
```
coolreader18: Once you do re-add this, it can just be `self.get_locals().into_object()`

coolreader18:
Aha! In the demo page, you can look at the global variable `__RUSTPYTHON_ERROR_STACK` and see (a slice):
```
core::panicking::panic_fmt::h0d6d5c8b201e3246@http://localhost:5000/2884fcd643f7322b02df.module.wasm:wasm-function[36148]:0xf3aeae
core::option::expect_failed::hae9ebb676e842a6f@http://localhost:5000/2884fcd643f7322b02df.module.wasm:wasm-function[29113]:0xeb4d33
core::option::Option<T>::expect::he72bf3761fad8028@http://localhost:5000/2884fcd643f7322b02df.module.wasm:wasm-function[11921]:0xc26318
rustpython_vm::vm::VirtualMachine::current_scope::h49f16e5d2524e3d7@http://localhost:5000/2884fcd643f7322b02df.module.wasm:wasm-function[16313]:0xd2cfa1
rustpython_vm::vm::VirtualMachine::get_locals::hbe0204e1c9c7fbdf@http://localhost:5000/2884fcd643f7322b02df.module.wasm:wasm-function[13198]:0xc7f2d4
rustpython_vm::vm::VirtualMachine::import::h1dd82715101d7e1b@http://localhost:5000/2884fcd643f7322b02df.module.wasm:wasm-function[1520]:0x5f1021
rustpython_vm::vm::VirtualMachine::try_class::hbeb6ef59ad4b76fe@http://localhost:5000/2884fcd643f7322b02df.module.wasm:wasm-function[2274]:0x6f4c29
```
Which is from here in `vm.rs`:
```rust
pub fn current_scope(&self) -> Ref<Scope> {
    let frame = self
        .current_frame()
        .expect("called current_scope but no frames on the stack");
    Ref::map(frame, |f| &f.scope)
}
```
Not sure exactly what to do about this, but that's the root of the problem. Odd. Does `get_locals()` pop a frame off the stack or something?

palaviv:
Thanks @coolreader18. Are you seeing this when running locally or on travis? How can I get `__RUSTPYTHON_ERROR_STACK`?
Maybe this happens because `StoredVirtualMachine` has a different `Scope`?

palaviv:
@coolreader18 can you share the full stacktrace please?

palaviv:
My guess is this:
We are failing in the following cal to `pyresult_to_jsresult` in `WASMVirtualMachine.run`:
```rust
let result = vm.run_code_obj(code, scope.borrow().clone());
convert::pyresult_to_jsresult(vm, result)
```
When we run the code in the line before we acutally push a frame to the `vm` and then we pop it when we finish (`vm.run_frame`). So when we call `pyresult_to_jsresult` our vm does not have any frames. I will change the code to pass locals only if we have frames and we will see if that works.

coolreader18:
Hallelujah! Do you want to merge this now?

palaviv:
> Hallelujah! Do you want to merge this now?

Very very very much :smiley:

windelbouwman:
Can we improve on error reporting so that we might be able to pinpoint these issues easier in the future?

coolreader18:
That's probably doable, I'll look into it.

PR #1054, opened 2019-06-22
KeyError should have passed key object as a first element of .args

mkurnikov:


tests/snippets/dict.py
```diff
@@ -95,8 +95,15 @@
 x[2] = 2
 x[(5, 6)] = 5
 
-with assertRaises(KeyError):
-    x["not here"]
+assert_raises(KeyError,
+              lambda: x['not here'],
+              "'not here'")
```
palaviv: I am not sure the `msg` `'not here'` here is actually helpful. The output is:
```py
AssertionError: KeyError was not raised: 'not here'
```
Please consider removing the `msg` or making it more helpful.

tests/snippets/dict.py
```diff
@@ -95,8 +95,15 @@
...
+assert_raises(KeyError,
+              lambda: x[10],
+              "10")
+assert_raises(KeyError,
+              lambda: x[10.0],
+              "10.0")
```
palaviv: You should had some test that you actually return the `__repr__` of an object and not the `__str__`:
```py
class A: 
    def __repr__(self): 
        return "repr"

    def __str__(self): 
        return "str"

a = A()

try:
    x[a]
except KeyError as err:
    assert str(err) == "repr"
```

vm/src/obj/objmappingproxy.rs
```diff
@@ -28,7 +28,8 @@ impl PyMappingProxy {
         if let Some(value) = objtype::class_get_attr(&self.class, key.as_str()) {
             return Ok(value);
         }
-        Err(vm.new_key_error(format!("Key not found: {}", key)))
+        let key_repr = vm.to_repr(key.as_object())?.to_string();
```
palaviv: `key` is already a string here. There is not need to convert it to a object and the calling `__repr__` on it.

tests/snippets/dict.py
mkurnikov: Thanks, I got the idea of the function wrong.

mkurnikov:
Blocked by https://github.com/RustPython/RustPython/pull/1056

mkurnikov:
@palaviv I addressed your comments, it's ready for review. 

It turned out a bit more complicated, as `KeyError` really returns passed object, not it's string representation. 

Not sure about the last commit, but in my opinion it describes what method does better (but diff is big for such a small enhancement).

tests/snippets/testutils.py
```diff
@@ -20,19 +20,27 @@ def assert_raises(exc_type, expr, msg=None):
...
+        return self.raised
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         if exc_type is None:
             failmsg = '{} was not raised'.format(self.expected.__name__)
             assert False, failmsg
         if not issubclass(exc_type, self.expected):
             return False
+
+        self.raised.exception = exc_val
```
palaviv: Why do we need the `Raised` class? You can just do:
```py
self.raised_exception = exc_val
```

mkurnikov: Yeah, you're right.

mkurnikov:
> Nice work. I am just not sure about vm.new_exception_with_msg maybe we can keep vm.new_excpetion and do the conversion inside? What do you think?

Another `new_exception_obj` method will still be required though. I'll try, thanks for the review.

mkurnikov:
@palaviv I addressed comments.

Issue #945, opened 2019-05-09
builtin ord() doesn't work with bytes and bytearray types


sapir:
In CPython, `ord(b'a') == 97` and `ord(bytearray(b'a')) == 97` but RustPython fails with
```
TypeError: argument of type <class 'str'> is required for parameter 1 (string) (got: <class 'bytes'>)
```

dralley:
Looks like this can be closed

PR #1055, opened 2019-06-23
add bytearray and bytes input types for ord()

romab1998:
fix for #945


windelbouwman:
Hi @romab1998 thank you for your contribution! The change looks good. Could you in addition also add a small test snippet in the `test/snippets` directory to test this behavior?

romab1998:
@windelbouwman I've added snippet, but Travis CI went mad and build stuck (even before rust wast installed). Can we somehow rerun it?

coolreader18:
We have to fix that issue, it's been happening recently; just timing out when fetching and unpacking the cache. I restarted the build for you, it should work now.

PR #1056, opened 2019-06-24
Add args attribute to exceptions, make __str__ and __repr__ compatible with CPython

mkurnikov:


mkurnikov:
Are we focusing python 3.6 or 3.7? On 3.7 the line
```
assert repr(exc) == "KeyError('message')"
```
runs ok, on 3.6 (and consequently on CI) it fails.

PR #1057, opened 2019-06-26
Add CPython library modules needed for unittest

coolreader18:
The big blocker for the next set of modules (`logging`, `traceback`)
requires `re`, so I think someone should try implementing that in full
(using CPython's `re.py`, preferably). I'll open an issue for that.


windelbouwman:
This change is mostly good, except that the strings.py module is broken. Perhaps revert that module, and keep the rust version? The other modules require adding, so these are good.

PR #1059, opened 2019-06-26
Add initial unicodedata module.

windelbouwman:


vm/src/stdlib/unicodedata.rs
```diff
@@ -0,0 +1,94 @@
...
+    character: PyStringRef,
+    default: OptionalArg<PyObjectRef>,
+    vm: &VirtualMachine,
+) -> PyResult {
+    let my_char = extract_char(character, vm)?;
+
+    if let Some(name) = Name::of(my_char) {
+        Ok(vm.new_str(name.to_string()))
+    } else {
+        match default {
```
palaviv: You can use `unwrap_or` here to make it nicer.

windelbouwman: Okay, but I'll keep it this way, its probably clear enough.

PR #1064, opened 2019-06-27
Add initial hashlib module.

windelbouwman:


windelbouwman:
Hmm, apparantly rust-crypto is abandoned.. Too bad, it is a nice library!

coolreader18:
Should we use `hashlib.py` for this? Then we'd only have to implement the raw hashing functions.

palaviv:
> Hmm, apparantly rust-crypto is abandoned.. Too bad, it is a nice library!

I have a open pull request to `rust-crypto` since Juny 2016 :disappointed:

palaviv:
> Should we use `hashlib.py` for this? Then we'd only have to implement the raw hashing functions.

I am not sure we need `hashlib.py`. The main goal is to maintain the `algorithms_available` and `algorithms_guaranteed`. I think in the current stage we will support a finite state of hashes that are written in pure rust so we will know the list.

windelbouwman:
I would also say that in this case, we can have a rust only library. hashlib.py is merely a pass through interface.

vm/src/stdlib/hashlib.rs
```diff
@@ -0,0 +1,292 @@
...
+        "sha3_384" => Ok(PyHasher::new("sha3_384", HashWrapper::sha3_384())),
+        "sha3_512" => Ok(PyHasher::new("sha3_512", HashWrapper::sha3_512())),
+        // TODO: "shake128" => Ok(PyHasher::new("shake128", HashWrapper::shake128())),
+        // TODO: "shake256" => Ok(PyHasher::new("shake256", HashWrapper::shake256())),
+        "blake2b" => Ok(PyHasher::new("blake2b", HashWrapper::blake2b())),
+        "blake2s" => Ok(PyHasher::new("blake2s", HashWrapper::blake2s())),
+        other => Err(vm.new_value_error(format!("Unknown hashing algorithm: {}", other))),
+    }?;
+
+    match data {
```
palaviv: How about:
```rust
if let  OptionalArg::Present(data) = data {
     hasher.update(data, vm)?;
}
```

windelbouwman: Ah, yes. I'll do that! Thank you!

PR #1065, opened 2019-06-27
Give more information in the error messages for failed wasm tests

coolreader18:


vm/src/vm.rs
```diff
@@ -303,33 +303,31 @@ impl VirtualMachine {
...
-                    "__import__ not found".to_string(),
-                )),
-            }
-        }
+        let sys_modules = self.get_attribute(self.sys_module.clone(), "modules")?;
+        sys_modules.get_item(module.to_string(), self).or_else(|_| {
+            let import_func = self
+                .get_attribute(self.builtins.clone(), "__import__")
+                .map_err(|_| {
+                    self.new_exception(
```
palaviv: I feel bad commenting on this as I wrote this but if you make it nicer we can use `vm.new_import_error` :smiling_imp:

wasm/demo/package.json
```diff
@@ -24,7 +25,7 @@
         "build": "webpack",
         "dist": "webpack --mode production",
         "test": "cd ../tests; pipenv run pytest",
-        "ci": "start-server-and-test dev http-get://localhost:8080 test"
+        "ci": "start-server-and-test 'npm run build && serve dist/ -np 8080' :8080 test"
```
palaviv: It seems like it does not recognize the `-np` param in travis. Maybe we need a newer version there?

coolreader18: No, I just thought it supported flag concatenation without checking, it should just be `-n -p 8080`.

vm/src/vm.rs
```diff
@@ -303,33 +303,26 @@ impl VirtualMachine {
...
-                    ],
-                ),
-                Err(_) => Err(self.new_exception(
-                    self.ctx.exceptions.import_error.clone(),
-                    "__import__ not found".to_string(),
-                )),
-            }
-        }
+        let sys_modules = self.get_attribute(self.sys_module.clone(), "modules")?;
+        sys_modules.get_item(module.to_string(), self).or_else(|_| {
```
windelbouwman: Why did you change the original code to a lambda function with the `or_else` statement? Is there a performance benefit of using a lambda function? Personally I find the code a little harder to understand when a lot of lambda functions are involved.

vm/src/vm.rs
```diff
@@ -303,33 +303,26 @@ impl VirtualMachine {
...
+            let import_func = self
```
windelbouwman: First evaluating `import_func` is a good idea!

vm/src/vm.rs
```diff
@@ -303,33 +303,26 @@ impl VirtualMachine {
...
+                .get_attribute(self.builtins.clone(), "__import__")
+                .map_err(|_| self.new_import_error("__import__ not found".to_string()))?;
+            self.invoke(
+                import_func,
+                vec![
+                    self.ctx.new_str(module.to_string()),
+                    if let Some(frame) = self.current_frame() {
```
windelbouwman: Maybe fetch the current frame first, just like fetching the `import_func` first, to reduce the size of the `self.invoke` expression.

vm/src/vm.rs
coolreader18: I personally find it more readable because I think the intent is clearer. "Get it from `sys.modules`, or else if that fails try calling the import function.

windelbouwman: Hmm, I guess it is a matter of personal preference/opinion? Maybe we could settle on having both ways, and leaving the code in either style when touching the code? Does that sound good to you? I'll do my best on this strategy at least.

PR #1070, opened 2019-06-29
Add "Ellipsis" for Ellipsis

corona10:
Currently, only "..." works for Ellipsis type.
This PR will fix the RustPython to work correctly on this code.
```python
a = Ellipsis
b = ...
a is b
```

### DIFF
--- a/tests/snippets/ellipsis.py
+++ b/tests/snippets/ellipsis.py
@@ -3,6 +3,14 @@
 a = ...
 b = ...
 c = type(a)()  # Test singleton behavior
+d = Ellipsis
+e = Ellipsis
 
 assert a is b
 assert b is c
+assert b is d
+assert d is e
+
+assert Ellipsis is ...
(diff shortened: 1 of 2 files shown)


parser/src/lexer.rs
```diff
@@ -99,6 +99,7 @@ pub fn get_keywords() -> HashMap<String, Tok> {
 
     // Alphabetical keywords:
     keywords.insert(String::from("..."), Tok::Ellipsis);
+    keywords.insert(String::from("Ellipsis"), Tok::Ellipsis);
```
windelbouwman: Ellipsis is a builtin name. See my other comment. You might want to add this testcase:
```python
assert Ellipsis is ...
Ellipsis = 2
assert Ellipsis is not ...
```

corona10:
@windelbouwman Thank's for the review. Please take a look

windelbouwman:
Perfection!

PR #1073, opened 2019-06-29
Test namedtuple

youknowone:


tests/snippets/test_namedtuple.py
```diff
@@ -0,0 +1,36 @@
+code = 'def __new__(_cls, hits, misses, maxsize, currsize): return _tuple_new(_cls, (hits, misses, maxsize, currsize))'
+n = {'_tuple_new': tuple.__new__, '__name__': 'namedtuple__CacheInfo'}
+exec(code, n)
+assert '__new__' in n
+
+def namedtuple(typename, field_names, *, rename=False, defaults=None, module=None):
```
windelbouwman: What is the reason for this test snippet? Could you add a comment to it why this test is present?

youknowone: Sorry, in short, it is useless. I didn't realize this PR doesn't have any actual commits after rebase.

PR #1074, opened 2019-06-29
Implement int.__rmod__

JimJeon:
Extracted `inner_mod` function and used in `int.__mod__` and `int.__rmod__`
and also added test snippets of `int.__mod__` and `int.__rmod__`.

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -145,7 +145,7 @@ fn inner_mod(int1: &PyInt, int2: &PyInt, vm: &VirtualMachine) -> PyResult {
         Err(vm.new_zero_division_error("integer modulo by zero".to_string()))
     }
 }
- 
+
 #[pyimpl]
 impl PyInt {
     #[pymethod(name = "__eq__")]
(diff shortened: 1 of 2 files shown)


vm/src/obj/objint.rs
```diff
@@ -369,12 +377,18 @@ impl PyInt {
...
-            }
+            let other = other.payload::<PyInt>().unwrap();
+            inner_mod(self, &other, vm)
+        } else {
+            Ok(vm.ctx.not_implemented())
+        }
+    }
+
+    #[pymethod(name = "__rmod__")]
+    fn rmod_(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
```
sanxiyn: `mod` is a Rust keyword, but `rmod` is not. You can just use `rmod` instead of `rmod_`.

JimJeon: Thank you

palaviv:
@JimJeon rustfmt is failing. Please fix that and I will merge this :smile:

Issue #1075, opened 2019-06-29
1-1j print wrong


corona10:
```
Welcome to the magnificent Rust Python 0.0.1 interpreter 😱 🖖
>>>>> 1-1j
(1+-1j) # should be 1-1j
```

I am going to fix this issue.

PR #1076, opened 2019-06-29
Fix complex representation for negative imaginary number case.

corona10:
fixes: #1075

### DIFF
--- a/vm/src/obj/objcomplex.rs
+++ b/vm/src/obj/objcomplex.rs
@@ -204,7 +204,7 @@ impl PyComplex {
         if re == 0.0 {
             format!("{}j", im)
         } else {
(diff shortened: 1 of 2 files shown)


vm/src/obj/objcomplex.rs
```diff
@@ -204,7 +204,11 @@ impl PyComplex {
         if re == 0.0 {
             format!("{}j", im)
         } else {
-            format!("({}+{}j)", re, im)
+            if im < 0.0 {
```
youknowone: Maybe sign forcing is possible? `format!("({}{:+}j)", re, im)`

corona10: @youknowone Thanks for the comment. It works well

PR #1084, opened 2019-06-29
fix float–int cross-type comparisons

### LINKED ISSUE #1069: Bad cross-type comparison between float and int
```python
>>>>> a = 1.0; b = 1
>>>>> while True: a *= 2; b *= 2; assert a > b - 1, b
AssertionError: 18014398509481984
>>>>> 2**54
18014398509481984
>>>>> 2.**54 > 2**54 - 1
False
```

The loop should not fail, probably except for the eventual MemoryError or so. We currently compare them by converting int to float, and thus by losing precision. Note that the other way also doesn't work because we then lose fractional bits. The corresponding [CPython code](https://github.com/python/cpython/blob/80097e089ba22a42d804e65fbbcf35e5e49eed00/Objects/floatobject.c#L341-L535) is understandably complex.

silmeth:
Fixes #1069


windelbouwman:
Looks good to me! Thanks!

PR #1086, opened 2019-06-30
Move bytecode into own crate.

windelbouwman:
Reasons for moving bytecode into own crate:

- Ability to precompile python code as bytecode and run them in VM only wasm/microcntroller build to reduce size.
- Cleaner seperation between compiler / vm / ast.


bytecode/src/bytecode.rs
```diff
@@ -4,19 +4,39 @@
...
+    row: usize,
+    column: usize,
+}
+
+impl Location {
+    pub fn new(row: usize, column: usize) -> Self {
+        Location { row, column }
+    }
+
+    pub fn get_row(&self) -> usize {
```
silmeth: I know this is not specific to this pull request (the same code exists for a long time in `ast` module), but Rust naming guidelines encourage to not use the `get_` prefix in getter methods, so perhaps it’d be good to change those to `fn row(&self)` and `fn column(&self)` here and in `ast`?

See:
* https://doc.rust-lang.org/1.0.0/style/style/naming/README.html#getter/setter-methods-[rfc-344]
* https://rust-lang-nursery.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

windelbouwman: You're absolutely right. I'll do that.

PR #1093, opened 2019-06-30
Change get_row() into row()

windelbouwman:
As commented in #1086


windelbouwman:
Has anyone an idea what is wrong with the travis build?

coolreader18:
I think it was something weird with caching, I cleared them and restarted the build and it's good.

Issue #1094, opened 2019-06-30
How should we list contributors in Cargo.toml(s)?


coolreader18:
Now that we're going to start publishing our crates (#838), how what should we put for `authors`  in the `Cargo`s`.toml`? All the crates have different authors listed, usually just whoever first created that package. What should we change them to to keep them consistent across everything? [Relevant section in cargo docs](https://doc.rust-lang.org/cargo/reference/manifest.html#the-authors-field-optional)

windelbouwman:
As I wrote a fairly large portion of the code, I don't mind being put as author. I prefer however something like rustpython team or something like that. Is this possible?

windelbouwman:
As package names I would prefer this:
- `rustpython-parser`
- `rustpython-compiler`
- `rustpython-bytecode`
- `rustpython-vm`

windelbouwman:
I propose to release rustpython-parser as the first package, since this is the least changing one.

coolreader18:
I think we could, maybe something like:
```toml
authors = ["RustPython contributors <github.com/RustPython/RustPython/contributors>"]
```

windelbouwman:
I propose:
```toml
authors = ["RustPython Team"]
```

coolreader18:
Sounds good to me!

windelbouwman:
I'll make a PR for this.

hung44zzz:
Unscribe me

PR #1096, opened 2019-07-01
Underscore2hyphen

windelbouwman:
Update authors field and package name in cargo.toml files.


windelbouwman:
As discussed in #1094

coolreader18:
I added some more reviewers, not everyone needs to review but it'd be good to have at least one other person look over this change.

PR #1097, opened 2019-07-02
Add deque to collections

coolreader18:


palaviv:
@coolreader18 can you add a snippet test for this code?

coolreader18:
Alright, I think this is all ready for review.

vm/src/stdlib/collections.rs
```diff
@@ -0,0 +1,346 @@
...
+use crate::pyobject::{IdProtocol, PyClassImpl, PyIterable, PyObjectRef, PyRef, PyResult, PyValue};
+use crate::vm::ReprGuard;
+use crate::VirtualMachine;
+use itertools::Itertools;
+use std::cell::{Cell, RefCell};
+use std::collections::VecDeque;
+
+#[pyclass(name = "deque")]
+#[derive(Debug, Clone)]
+struct PyDeque {
```
windelbouwman: Are there coming more collection types? If so, it might be handy to create a subfolder with seperate files per type. Not sure if you want to do this at this point.

vm/src/stdlib/collections.rs
```diff
@@ -0,0 +1,346 @@
...
+    deque: RefCell<VecDeque<PyObjectRef>>,
+    maxlen: Cell<Option<usize>>,
+}
+
+impl PyValue for PyDeque {
+    fn class(vm: &VirtualMachine) -> PyClassRef {
+        vm.class("_collections", "deque")
```
windelbouwman: We really need a better solution for this construction in general. Maybe we could open an issue for it, asking other people for ideas on how to solve this?

vm/src/stdlib/collections.rs
coolreader18: I think that was the only one that needed a native implementation, the rest are implemented by the python collections module. It would be good to eventually optimize and add the native implementations, though.

PR #1102, opened 2019-07-03
Remove old import

palaviv:


vm/src/import.rs
```diff
@@ -46,39 +43,6 @@ pub fn import_builtin(vm: &VirtualMachine, module_name: &str) -> PyResult {
...
-    let sys_modules = vm.get_attribute(vm.sys_module.clone(), "modules").unwrap();
-
-    // First, see if we already loaded the module:
-    if let Ok(module) = sys_modules.get_item(module_name.to_string(), vm) {
-        Ok(module)
-    } else if vm.frozen.borrow().contains_key(module_name) {
-        import_frozen(vm, module_name)
-    } else if vm.stdlib_inits.borrow().contains_key(module_name) {
-        import_builtin(vm, module_name)
-    } else if cfg!(feature = "rustpython-compiler") {
```
windelbouwman: The point of this feature flag was to check if rustpython was built with bytecode compiler support. If not, the module cannot be loaded from disc. I think importlib.py has a similar mechanism. Please refer to the function `imp_exec_builtin` in `imp.rs`. Maybe the logic deleted here, can be replicated over there?

vm/src/stdlib/thread.rs
```diff
@@ -38,7 +36,7 @@ fn get_ident(_vm: &VirtualMachine) -> u32 {
 }
 
 fn allocate_lock(vm: &VirtualMachine) -> PyResult {
-    let module = import::import_module(vm, PathBuf::default(), "_thread")?;
+    let module = vm.import("_thread", &vm.ctx.new_tuple(vec![]), 0)?;
     let lock_class = vm.get_attribute(module.clone(), "RLock")?;
```
windelbouwman: Please replace this code by `let lock_class = vm.class("_thread", "RLock")`. This will call `vm.import`

vm/src/import.rs
palaviv: Only the `external_importers` use compile now. So I changed `init_importlib` to not install them if there is not compiler.

vm/src/stdlib/thread.rs
palaviv: Done

PR #1104, opened 2019-07-04
Remove importlib frames from traceback

palaviv:
Previously the traceback was:
```py
  File "/tmp/a.py", line 3, in <module>
    import b
  File "_frozen_importlib", line 1093, in __import__
  File "_frozen_importlib", line 1014, in _gcd_import
  File "_frozen_importlib", line 991, in _find_and_load
  File "_frozen_importlib", line 975, in _find_and_load_unlocked
  File "_frozen_importlib", line 686, in _load_unlocked
  File "_frozen_importlib", line 677, in _load_unlocked
  File "_frozen_importlib", line 671, in _load_unlocked
  File "_frozen_importlib_external", line 778, in exec_module
  File "_frozen_importlib", line 219, in _call_with_frames_removed
  File "/tmp/b.py", line 1, in <module>
    1/0
ZeroDivisionError: 'integer division by zero'
```
Now it is:
```py
  File "/tmp/a.py", line 3, in <module>
    import b
  File "/tmp/b.py", line 1, in <module>
    1/0
ZeroDivisionError: 'integer division by zero'
```

### DIFF
--- a/vm/src/import.rs
+++ b/vm/src/import.rs
@@ -4,8 +4,8 @@
 
 use crate::bytecode::CodeObject;
 use crate::frame::Scope;
-use crate::obj::objcode;
-use crate::pyobject::{ItemProtocol, PyResult, PyValue};
+use crate::obj::{objcode, objsequence, objstr, objtype};
+use crate::pyobject::{ItemProtocol, PyObjectRef, PyResult, PyValue};
 use crate::vm::VirtualMachine;
 #[cfg(feature = "rustpython-compiler")]
 use rustpython_compiler::compile;
@@ -81,4 +81,36 @@ pub fn import_codeobj(
         Scope::with_builtins(None, attrs, vm),
     )?;
     Ok(module)
+}
+
+pub fn remove_importlib_frames(vm: &VirtualMachine, exc: &PyObjectRef) -> PyObjectRef {
+    let always_trim = objtype::isinstance(exc, &vm.ctx.exceptions.import_error);
+
+    if let Ok(tb) = vm.get_attribute(exc.clone(), "__traceback__") {
+        if objtype::isinstance(&tb, &vm.ctx.list_type()) {
+            let mut tb_entries = objsequence::get_elements_list(&tb).to_vec();
+            tb_entries.reverse();
+            let mut new_tb = Vec::with_capacity(tb_entries.len());
+
+            for tb_entry in tb_entries.iter() {
+                let mut current_chunk = vec![];
+                let location_attrs = objsequence::get_elements_tuple(&tb_entry);
+                let file_name = objstr::get_value(&location_attrs[0]);
+                if file_name != "_frozen_importlib" && file_name != "_frozen_importlib_external" {
+                    current_chunk.clear();
+                    new_tb.push(tb_entry.clone())
+                } else {
+                    current_chunk.push(tb_entry.clone());
+                    let run_obj_name = objstr::get_value(&location_attrs[0]);
+                    if run_obj_name == "_call_with_frames_removed" || always_trim {
+                        new_tb.append(&mut current_chunk);
+                    }
+                };
+            }
+            new_tb.reverse();
+            vm.set_attr(exc, "__traceback__", vm.ctx.new_list(new_tb))
+                .unwrap();
+        }
+    }
+    exc.clone()
 }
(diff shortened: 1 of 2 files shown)


palaviv:
This is a adaptation of the [`remove_importlib_frames`](https://github.com/python/cpython/blob/7f41c8e0dd237d1f3f0a1d2ba2f3ee4e4bd400a7/Python/import.c#L1484) from Cpython.

vm/src/import.rs
```diff
@@ -144,3 +144,35 @@ fn find_source(vm: &VirtualMachine, current_path: PathBuf, name: &str) -> Result
         None => Err(format!("No module named '{}'", name)),
     }
 }
+
+pub fn remove_importlib_frames(vm: &VirtualMachine, exc: &PyObjectRef) -> PyObjectRef {
```
windelbouwman: Are you sure you want to add this? This might be a good opportunity to keep the rustpython interpreter transparent to users, and show them what is actually going on. What is your opinion on this?

vm/src/import.rs
```diff
@@ -144,3 +144,35 @@ fn find_source(vm: &VirtualMachine, current_path: PathBuf, name: &str) -> Result
...
+    let always_trim = objtype::isinstance(exc, &vm.ctx.exceptions.import_error);
+
+    if let Ok(tb) = vm.get_attribute(exc.clone(), "__traceback__") {
+        if objtype::isinstance(&tb, &vm.ctx.list_type()) {
+            let mut tb_entries = objsequence::get_elements_list(&tb).to_vec();
+            tb_entries.reverse();
```
windelbouwman: It looks like the list is reversed twice. This might not be required.

vm/src/import.rs
```diff
@@ -144,3 +144,35 @@ fn find_source(vm: &VirtualMachine, current_path: PathBuf, name: &str) -> Result
...
+            let mut new_tb = Vec::with_capacity(tb_entries.len());
```
windelbouwman: Since this function is basically a filtering of a list, this might be an excellent opportunity to use the `filter_map` function, or `filter` function on rust iterators, and then use `collect` on the filtered iterator.

Something along this idea:

```rust
let new_tb = tb_entries.iter()
    .filter(|tb_entry| {
       // fetch filename etc..
       // funky filtering here, if file_name != frozen etc...
    })
    .collect();


```

vm/src/import.rs
```diff
@@ -144,3 +144,35 @@ fn find_source(vm: &VirtualMachine, current_path: PathBuf, name: &str) -> Result
...
+            for tb_entry in tb_entries.iter() {
+                let mut current_chunk = vec![];
+                let location_attrs = objsequence::get_elements_tuple(&tb_entry);
+                let file_name = objstr::get_value(&location_attrs[0]);
+                if file_name != "_frozen_importlib" && file_name != "_frozen_importlib_external" {
+                    current_chunk.clear();
+                    new_tb.push(tb_entry.clone())
+                } else {
+                    current_chunk.push(tb_entry.clone());
+                    let run_obj_name = objstr::get_value(&location_attrs[0]);
```
windelbouwman: This is the same as `file_name` earlier on.

vm/src/import.rs
```diff
@@ -144,3 +144,35 @@ fn find_source(vm: &VirtualMachine, current_path: PathBuf, name: &str) -> Result
...
+    let always_trim = objtype::isinstance(exc, &vm.ctx.exceptions.import_error);
+
+    if let Ok(tb) = vm.get_attribute(exc.clone(), "__traceback__") {
+        if objtype::isinstance(&tb, &vm.ctx.list_type()) {
+            let mut tb_entries = objsequence::get_elements_list(&tb).to_vec();
+            tb_entries.reverse();
+            let mut new_tb = Vec::with_capacity(tb_entries.len());
+
+            for tb_entry in tb_entries.iter() {
+                let mut current_chunk = vec![];
```
windelbouwman: I think this variable can be removed right?

vm/src/import.rs
palaviv: I think that it gives too much information. I will add a `TODO` that when running on `verbose` this function will do nothing.

vm/src/import.rs
palaviv: This is only a simple filter if the error is from outside the `importlib` code. That is why we remove only chunks that end in `_call_with_frames_removed`.

vm/src/import.rs
palaviv: Nice catch. Thanks.

vm/src/import.rs
palaviv: Fixed.

vm/src/import.rs
palaviv: Done

PR #1106, opened 2019-07-05
Pass more information in user defined parse error

palaviv:
Before the message was:
```py
SyntaxError: 'Got unsupported token(s) at line 0'
```
And now it is:
```py
SyntaxError: 'Got unexpected token \t at line 4 column 25'
```

### DIFF
--- a/compiler/src/error.rs
+++ b/compiler/src/error.rs
@@ -1,4 +1,4 @@
-use rustpython_parser::error::ParseError;
+use rustpython_parser::error::{ParseError, ParseErrorType};
 use rustpython_parser::lexer::Location;
 
 use std::error::Error;
@@ -13,8 +13,8 @@ pub struct CompileError {
 impl From<ParseError> for CompileError {
     fn from(error: ParseError) -> Self {
         CompileError {
-            error: CompileErrorType::Parse(error),
-            location: Default::default(), // TODO: extract location from parse error!
+            error: CompileErrorType::Parse(error.error),
+            location: error.location,
         }
     }
 }
@@ -28,7 +28,7 @@ pub enum CompileErrorType {
     /// Expected an expression got a statement
     ExpectExpr,
     /// Parser error
-    Parse(ParseError),
+    Parse(ParseErrorType),
     SyntaxError(String),
     /// Multiple `*` detected
     StarArgs,
@@ -56,10 +56,7 @@ impl fmt::Display for CompileError {
         }?;
 
         // Print line number:
-        match &self.error {
-            CompileErrorType::Parse(..) => Ok(()),
-            _ => write!(f, " at line {:?}", self.location.row()),
-        }
+        write!(f, " at {}", self.location)
     }
 }
 
--- a/compiler/src/error.rs
+++ b/compiler/src/error.rs
@@ -56,7 +56,10 @@ impl fmt::Display for CompileError {
         }?;
 
         // Print line number:
-        write!(f, " at line {:?}", self.location.row())
+        match &self.error {
+            CompileErrorType::Parse(..) => Ok(()),
+            _ => write!(f, " at line {:?}", self.location.row()),
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


compiler/src/error.rs
```diff
@@ -56,7 +56,10 @@ impl fmt::Display for CompileError {
         }?;
 
         // Print line number:
-        write!(f, " at line {:?}", self.location.row())
+        match &self.error {
+            CompileErrorType::Parse(..) => Ok(()),
```
windelbouwman: Hmm. This exception for the parse error is not ideal, but we can improve on this later.

parser/src/error.rs
```diff
@@ -24,7 +24,7 @@ pub enum ParseError {
     /// Parser encountered an unexpected token
     UnrecognizedToken(TokSpan, Vec<String>),
     /// Maps to `User` type from `lalrpop-util`
-    Other,
+    Other(LexicalError),
```
windelbouwman: It might be good to rename this error variant into `Lexical(LexicalError)`.

parser/src/error.rs
```diff
@@ -34,8 +34,7 @@ impl From<InnerError<Location, Tok, LexicalError>> for ParseError {
             // TODO: Are there cases where this isn't an EOF?
             InnerError::InvalidToken { location } => ParseError::EOF(Some(location)),
             InnerError::ExtraToken { token } => ParseError::ExtraToken(token),
-            // Inner field is a unit-like enum `LexicalError::StringError` with no useful info
-            InnerError::User { .. } => ParseError::Other,
+            InnerError::User { error } => ParseError::Other(error),
```
windelbouwman: This is very good to propagate the lexer errors!

parser/src/lexer.rs
```diff
@@ -74,6 +75,26 @@ pub enum LexicalErrorType {
...
+        match self.error {
+            LexicalErrorType::StringError => write!(f, "Got unexpected string"),
+            LexicalErrorType::UnicodeError => write!(f, "Got unexpected unicode"),
+            LexicalErrorType::NestingError => write!(f, "Got unexpected nesting"),
+            LexicalErrorType::UnrecognizedToken { tok } => {
+                write!(f, "Got unexpected token {}", tok)
+            }
+            LexicalErrorType::OtherError(ref msg) => write!(f, "{}", msg),
+        }?;
+        write!(
```
windelbouwman: I would prefer to add the line and column info later on, so in the `Display` method of the `CompilerError` struct.

compiler/src/error.rs
windelbouwman: Please implement the todo at this line: https://github.com/RustPython/RustPython/blob/ef1d5433df31a3c4ba7e58250b75de85ca76eaa1/compiler/src/error.rs#L17

Then we have proper locations on all compiler errors (where possible).

palaviv: Done

parser/src/error.rs
palaviv: Done

parser/src/lexer.rs
palaviv: Done

windelbouwman:
This change looks very good now. The only thing left is to take a look at the wasm build.

palaviv:
I am trying to fix the wasm but the error currently look like:
```py
SyntaxError: Error parsing Python code: Got unexpected token: DoubleStar at Location { row: 1, column: 8 } at line 0
```
Not sure why the `fmt::Display` I defined is not working.

wasm/lib/src/vm_class.rs
```diff
@@ -281,30 +281,13 @@ impl WASMVirtualMachine {
...
-                            let _ =
-                                Reflect::set(&js_err, &"endrow".into(), &(loc.row() as u32).into());
-                            let _ = Reflect::set(
-                                &js_err,
-                                &"endcol".into(),
-                                &(loc.column() as u32).into(),
-                            );
-                        }
-                    }
+                    let _ =
```
windelbouwman: Well, adding the `location` attribute clearly was beneficial!

PR #1114, opened 2019-07-07
Enable clippy in CI

mkurnikov:


.travis.yml
```diff
@@ -39,6 +39,17 @@ matrix:
...
+    - name: Lint code with clippy
+      language: rust
+      rust: stable
+      cache: cargo
+      before_script:
+        - rustup component add clippy
+      script:
+        - cargo clippy --all -- -Dwarnings
+      env:
+        - JOBCACHE=3
```
windelbouwman: This variable must contain a unique number to prevent cache issues. Please pick a number other than the other values used elsewhere in the matrix.

.travis.yml
```diff
@@ -39,6 +39,17 @@ matrix:
       env:
         - JOBCACHE=3
 
+    - name: Lint code with clippy
```
windelbouwman: Does it make sense to combine the rustfmt check with this check to reduce the amount of different jobs? We could add `cargo clippy --all` to the rustfmt job above.

mkurnikov: Is there a way to make them appear in a different rows in the Travis output? I'd like to spare user from looking inside the Travis row to see what's going on. He sees that formatter failed, goes to his own terminal, runs it, creates new commit.

windelbouwman: Okay, that sounds fair to me.

PR #1115, opened 2019-07-07
Improve f-string syntax. Fixes #1099

### LINKED ISSUE #1099: f-string with bool expression and if-expression inside
All these fail to parse
```
f'{True or True}'
f'{1 == 1}'
f'{"0" if True else "1"}'
```

windelbouwman:


mkurnikov:
I tested it locally with various patterns, change indeed fixes the issue, thank you!

PR #1117, opened 2019-07-07
make exceptions repr tests Python 3.7-friendly by eval-repr round-trip

### LINKED ISSUE #1088: exceptions.py tests failing on >= 3.7 because repr changed
CPython 3.7 [changed exception’s `repr`](https://docs.python.org/3.7/whatsnew/3.7.html) implementation to get rid of a trailing comma:

> `repr` for [`BaseException`](https://docs.python.org/3.7/library/exceptions.html#BaseException) has changed to not include the trailing comma. Most exceptions are affected by this change. (Contributed by Serhiy Storchaka in [bpo-30399](https://bugs.python.org/issue30399).)

This makes the test case from `exceptions.py` fail on CPython 3.7 and later.

Should we get rid of the tests for repr, or perhaps just test substrings (that `repr` string contains both the type and the message)?

silmeth:
Closes #1088

### DIFF
--- a/tests/snippets/exceptions.py
+++ b/tests/snippets/exceptions.py
@@ -1,29 +1,37 @@
+def exceptions_eq(e1, e2):
+    return type(e1) is type(e2) and e1.args == e2.args
+
+def round_trip_repr(e):
+    assert exceptions_eq(e, eval(repr(e)))
+
 # KeyError
 empty_exc = KeyError()
 assert str(empty_exc) == ''
-assert repr(empty_exc) == 'KeyError()'
+round_trip_repr(empty_exc)
 assert len(empty_exc.args) == 0
 assert type(empty_exc.args) == tuple
 
 exc = KeyError('message')
 assert str(exc) == "'message'"
-assert repr(exc) == "KeyError('message',)"
+round_trip_repr(exc)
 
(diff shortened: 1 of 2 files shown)


tests/snippets/exceptions.py
```diff
@@ -1,29 +1,37 @@
+def exceptions_eq(e1, e2):
+    return type(e1) is type(e2) and e1.args == e2.args
+
+def round_trip_repr(e):
+    assert exceptions_eq(e, eval(repr(e)))
```
coolreader18: Nitpicky, but maybe this should return and the assert should be where it's called? That way an error would show the line with the exception variable instead of just the parameters.

silmeth: I based it on [the `round_trip_test()` from `json_snippet.py`](https://github.com/RustPython/RustPython/blob/master/tests/snippets/json_snippet.py#L4-L8) where it has similar logic. But I agree it’d be a bit nicer to move asserts out of the function. Shall I change it just here, or in both snippets (I haven’t found any other round-tripy utils).

silmeth:
Also corrected the last test case in `json_snippets.py` – because of a typo (`*` instead of `**`) it did not test what it was supposed to test…

PR #1118, opened 2019-07-08
Pass the Lib directory's path to rustpython at build time

coolreader18:


build.rs
```diff
@@ -0,0 +1,5 @@
+fn main() {
+    let mut lib_path = std::env::current_dir().unwrap();
+    lib_path.push("Lib");
+    println!("cargo:rustc-env=RUSTPYTHON_LIB_DIR={}", lib_path.display());
```
windelbouwman: Hmm, I understand how this would work, but I am not sure if we should do this now. At the moment i like the flexibility of the RUSTPYTHONPATH variable, which will let you point to other stdlibs as well. Another point, if we do this, we fix the path at build time. How would we be able to install rustpython into /usr/lib/rustpython37?

All in all, I think this might be good, but maybe for a later point in time.

src/main.rs
```diff
@@ -81,7 +90,7 @@ fn _run_string(vm: &VirtualMachine, source: &str, source_path: String) -> PyResu
     vm.run_code_obj(code_obj, Scope::with_builtins(None, attrs, vm))
 }
 
-fn handle_exception(vm: &VirtualMachine, result: PyResult) {
+fn handle_exception<T>(vm: &VirtualMachine, result: PyResult<T>) {
```
windelbouwman: What does this change do?

windelbouwman:
@cthulahoops @palaviv what is your opinion on this change?

build.rs
coolreader18: That's fair, I wasn't too sure about this change either. I'm fine with closing it, but I think we should think about some way of distributing the stdlib.

coolreader18:
This was just sort of an experiment, and while it did "work", I agree that this is a pretty sketchy way of doing it.

build.rs
windelbouwman: I like the idea of pre-compiling it into the vm. Another idea I had was to create additional crates for libraries, so that the vm can include it via feature flags. This way, you could trim the vm down to a minimum if required.

Let's try to get some logging and unittest modules to work, and figure this out along the road.

Issue #547, opened 2019-02-25
File IO issues


silmeth:
There a few differences between the way CPython handles file IO and the way RustPython does.

1. RustPython does not currently offer any `close()` methods, opened files get leaked and can be closed only by invoking `os.close(file.raw.fileno)` (or `os.close(file.buffer.raw.fileno)` for files in text mode) after importing the `os` module.
2. RustPython currently doesn’t offer text writing mode (which is the default for writing in CPython), so in CPython:
   ```python3
   >>> file = open('/tmp/tst', 'w')
   >>> file
   <_io.TextIOWrapper name='/tmp/tst' mode='w' encoding='UTF-8'>
   ```
   one can try to open a file for text writing in RustPython with `wt` mode, but then writing fails:
   ```python3
   >>>>> file = open('/tmp/tst', 'wt')
   >>>>> file.write('\ntest string')
   Traceback (most recent call last):
     File <stdin>, line 0, in <module>
   AttributeError: 'TextIOWrapper' object has no attribute 'write'
   ```
   and files for writing are by default opened in binary mode.
3. RustPython’s `fileno` is a property, while in CPython it is a *method* returning the underlying file descriptor.
4. CPython **doesn’t** rely on the `fileno()` value for file interactions, instead it uses [internal file handle](https://github.com/python/cpython/blob/bcda8f1d42a98d9022736dd52d855be8e220fe15/Modules/_io/fileio.c#L58) while [`fileno()` just returns it as a Python `int`](https://github.com/python/cpython/blob/bcda8f1d42a98d9022736dd52d855be8e220fe15/Modules/_io/fileio.c#L543-L556), so one can modify it, and the file object will still be referring to the original file:
   ```python3
   >>> file = open('/tmp/test', 'wb')
   >>> file.write(b'writing to the original file\n')
   29
   >>> file.raw.fileno = lambda: 1  # stdout descriptor
   >>> file.write(b'again writing to the original file\n')
   35
   >>> 
   ```
   while in RustPython, changing the `fileno` property changes the underlying file:
   ```python3
   >>>>> file = open('/tmp/test', 'wb')
   >>>>> file.write(b'writing to the original file\n')
   29
   >>>>> file.raw.fileno = 1  # stdout descriptor
   >>>>> file.write(b'this will appear in the stdout instead\n')
   this will appear in the stdout instead
   39
   >>>>> 
   ```
5. Some file attributes in CPython are protected from modifications, eg.:
   ```python3
[...]

PR #1119, opened 2019-07-08
make TextIOBase writable, handle malformed utf-8 in read()

silmeth:
This:
* adds `UnicodeError` and subtypes,
* makes `TextIOWrapper` throw a `UnicodeDecodeError` on malformed utf-8 in a file,
* makes `TextIOWrapper` writable if underlying buffer is writable,
* makes `TextIOWrapper` not readable if underlying buffer is writable.

When the wrapper is not readable/writable, CPython throws an [`io.UnsupportedOperation`](https://docs.python.org/3/library/io.html#io.UnsupportedOperation) which inherits from both `OSError` *and* `ValueError`. I just throw `ValueError` here – do we have multiple inheritance yet to do a proper `UnsupportedOperation`?

Also, all the [`UnicodeError`](https://docs.python.org/3/library/exceptions.html#UnicodeError) types in CPython have more internal structure than the simple exceptions I added here.

Some more things to be done to close the rest of #547 are:
* [ ] implement closing files, buffers, text wrappers,
* [ ] perhaps make file a separate `PyObjectPayload` to hide the underlying Rust `File` internally and no longer depend on `fileno`?,
* [ ] make `fileno` just an accessor method.

But I’d rather try to do them in separate pull requests (or let somebody else do them if anyone gets to it before me).


vm/src/stdlib/io.rs
```diff
@@ -442,16 +442,71 @@ fn text_io_wrapper_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
 fn text_io_base_read(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     arg_check!(vm, args, required = [(text_io_base, None)]);
 
+    let io_module = vm.import("_io", &vm.ctx.new_tuple(vec![]), 0)?;
+    let buffered_reader_class = vm
+        .get_attribute(io_module.clone(), "BufferedReader")
+        .unwrap()
+        .downcast::<PyClass>()
+        .unwrap();
```
coolreader18: All of this can just be

```rust
let buffered_reader_class = vm.try_class("_io", "BufferedReader")?;
```

vm/src/stdlib/io.rs
```diff
@@ -442,16 +442,71 @@ fn text_io_wrapper_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+        args,
+        required = [(text_io_base, None), (obj, Some(vm.ctx.str_type()))]
+    );
+
+    let io_module = vm.import("_io", &vm.ctx.new_tuple(vec![]), 0)?;
+    let buffered_writer_class = vm
+        .get_attribute(io_module.clone(), "BufferedWriter")
+        .unwrap()
+        .downcast::<PyClass>()
+        .unwrap();
```
coolreader18: And here

vm/src/stdlib/io.rs
silmeth: Nice, have missed it after all the refactors since March when I last looked at that :D.

vm/src/stdlib/io.rs
```diff
@@ -442,19 +442,65 @@ fn text_io_wrapper_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+
+    let write = vm
+        .get_method(raw.clone(), "write")
+        .ok_or_else(|| vm.new_attribute_error("BufferedWriter has no write method".to_owned()))
+        .and_then(|it| it)?;
+    let bytes = objstr::get_value(obj).into_bytes();
+
+    let len = vm.invoke(
+        write,
+        PyFuncArgs::new(vec![vm.ctx.new_bytes(bytes.clone())], vec![]),
```
windelbouwman: You could use `vm.invoke(write, vec![vm.ctx.new_bytes(bytes.clone())])` here, since `Into<PyArgs>` is implemented for `Vec`.

vm/src/stdlib/io.rs
```diff
@@ -442,19 +442,65 @@ fn text_io_wrapper_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+    let buffered_writer_class = vm.try_class("_io", "BufferedWriter")?;
+    let raw = vm.get_attribute(text_io_base.clone(), "buffer").unwrap();
+
+    if !objtype::isinstance(&raw, &buffered_writer_class) {
+        // TODO: this should be io.UnsupportedOperation error which derives from ValueError and OSError
+        return Err(vm.new_value_error("not writable".to_string()));
+    }
+
+    let write = vm
+        .get_method(raw.clone(), "write")
```
windelbouwman: Any reason why not using the `vm.call_method` function here?

Issue #1047, opened 2019-06-19
RustPython became extremely slow due to recent changes


michelhe:
A recent change probably made the interpreter extremely slow, this can be demonstrated by a simple benchmark:

```bash
# On current master
target/debug/rustpython tests/snippets/strings.py  2.25s user 0.01s system 98% cpu 2.301 total`

# On HEAD~100
target/debug/rustpython tests/snippets/strings.py  0.13s user 0.00s system 98% cpu 0.139 total
```

I'm currently doing a `git-bisect` to identify which change introduced the poor performance, and I'll update what I find.

michelhe:
Well, It's sadly not as straight as doing a simple git-bisect to find what's causing this, because many commits are broken and either not compiling / panicking, and the "good" execution time for my little benchmark is not consistent.

Anyways, I suspect that maybe some of the "importlib" changes are causing this because this is around the time that execution time skyrockets from ~0.120 to 0.7-2 for me

coolreader18:
That's odd though, because import machinery itself only takes ~15% of the execution time, you can test by changing the top of the file to:
```py
import time
t1 = time.time()
from testutils import assert_raises
print(time.time() -t1)
```
```
$ time target/debug/rustpython tests/snippets/strings.py
0.333484411239624
Hurray!
target/debug/rustpython tests/snippets/strings.py  2.19s user 0.00s system 99% cpu 2.203 total
```
Maybe there was some other change in #1018 that caused this performance regression?

coolreader18:
Although, maybe it's importlib intialization, just running `vm.run_code_obj(<importlib>)`

coolreader18:
Yeah, this is so weird! Look at this:
```
$ time target/debug/rustpython tests/snippets/strings.py
top of main(): SystemTime { tv_sec: 1560999593, tv_nsec: 403368144 }
top of _run_script(): SystemTime { tv_sec: 1560999594, tv_nsec: 477420455 }
top of _run_string(): SystemTime { tv_sec: 1560999594, tv_nsec: 477691635 }
top of strings.py: 1560999594.6324782
Hurray!
target/debug/rustpython tests/snippets/strings.py  1.57s user 0.01s system 99% cpu 1.579 total
```
Most of the delay happens before the file is even loaded or parsed! [Here's an asciinema](https://asciinema.org/a/awOylSSxSdO8TczBgDKeWcsfz) to better illustrate the time at which they're logged.

coolreader18:
[...]

PR #1120, opened 2019-07-08
Add flamegraph profiling using flame and flamer

coolreader18:
This is just an initial change to get the ball rolling for flame profiling, we can add more scopes later.

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -268,8 +268,15 @@ impl Frame {
         }
     }
 
-    #[cfg_attr(feature = "flame-it", flame("Frame"))]
+    // #[cfg_attr(feature = "flame-it", flame("Frame"))]
     pub fn run(&self, vm: &VirtualMachine) -> Result<ExecutionResult, PyObjectRef> {
+        flame_guard!(format!("Frame::run({})", self.code.obj_name));
+        // flame doesn't include notes in the html graph :(
+        flame_note!(
+            flame::StrCow::from("CodeObj name"),
+            self.code.obj_name.clone().into()
+        );
+
         let filename = &self.code.source_path.to_string();
 
         // This is the name of the object being run:
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -268,6 +268,7 @@ impl Frame {
         }
     }
 
+    #[cfg_attr(feature = "flame-it", flame("Frame"))]
     pub fn run(&self, vm: &VirtualMachine) -> Result<ExecutionResult, PyObjectRef> {
         let filename = &self.code.source_path.to_string();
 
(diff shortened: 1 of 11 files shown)


coolreader18:
There's not a ton of information here that would be helpful to optimization (yet), but it's really interesting to see!

image

windelbouwman:
This is pretty cool! Is it mandatory to add the flame-it attribute to each function you wish to track? Can we just follow all functions?

vm/src/frame.rs
```diff
@@ -268,7 +268,15 @@ impl Frame {
         }
     }
 
+    // #[cfg_attr(feature = "flame-it", flame("Frame"))]
     pub fn run(&self, vm: &VirtualMachine) -> Result<ExecutionResult, PyObjectRef> {
+        flame_guard!(format!("Frame::run({})", self.code.obj_name));
+        // flame doesn't include notes in the html graph :(
+        flame_note!(
+            flame::StrCow::from("CodeObj name"),
+            self.code.obj_name.clone().into()
```
windelbouwman: Nice! Now we can follow python objects as well!

vm/src/frame.rs
```diff
@@ -334,6 +342,7 @@ impl Frame {
 
     /// Execute a single instruction.
     #[allow(clippy::cognitive_complexity)]
+    #[cfg_attr(feature = "flame-it", flame("Frame"))]
```
windelbouwman: Do we need to explicitly enable all trace points?

windelbouwman:
Off topic: why does the importlib loading take so long? Is it a whole lot of code?

coolreader18:
re: importlib: The main problem is that it is a fair bit of code, and RustPython isn't all that optimized. See #1047.

re: automatically tagging all fns: see llogiq/flamer#45. I tried to do that initially, and there were a lot of issues with it. Even so, manually tagging all fns in `frame.rs` with `#[flame]` gave way too much irrelevant data, and the file size was huge as well (36MB compared to the 1MB with the current version here). e.g., do we really need to see how long each individual exception creation function takes?

coolreader18:
I had to use a modified local clone of flame to get [this flamegraph](https://gistcdn.githack.com/coolreader18/4b173399907966958f0ea3244ca07593/raw/323cb62d713bf53b7ea5171957ff0c8a9fbdc411/flame-graph.html) due to a bug (TyOverby/flame#47), but something interesting that you can see is that in `_install_external_importers`, when it imports `_frozen_importlib_external`, a fairly large portion of the time spent in `__import__` is before it even starts executing the module (`Frame::run(<module>)`). Maybe we can detect when it's importing `_frozen_importlib_external` from `_frozen_importlib` and use Rust to import it?

coolreader18:
Here it is with a simple optimization to skip `__import__` in that case: [link](https://gistcdn.githack.com/coolreader18/4b173399907966958f0ea3244ca07593/raw/e2c491c73ded65916544cd73e1ef9f387062d572/flame-graph-optimized.html). Shaved off 10ms, not bad!

palaviv:
CPython actually have large parts of the `__import__` function implemented in C. We can do the same in Rust for optimization.

PR #1121, opened 2019-07-09
Add #![doc(html_logo_url = ...)] to crates

coolreader18:
I randomly came across [this crate](https://docs.rs/adapton/0.3.30/adapton/) on docs.rs, and I had no idea that you were able to put logos in rustdoc! I figured we should put our derpy logo there. :smiley:


windelbouwman:
This is excellent. Our logo is really on par :)

PR #1123, opened 2019-07-09
complex: Fix hash function to work correctly

corona10:
Currently, hash(3.1-4.2j) or hash(3.1+4.2j) is crashed due to
overflow error on RustPython.

This patch will fix this issue.

### DIFF
--- a/tests/snippets/builtin_complex.py
+++ b/tests/snippets/builtin_complex.py
@@ -94,6 +94,10 @@ assert hash(complex(3.14)) == hash(float(3.14))
 assert hash(complex(-float('inf'))) == hash(-float('inf'))
 assert hash(1j) != hash(1)
 
+# TODO: Find a way to test platform dependent values
+assert hash(3.1 - 4.2j) == hash(3.1 - 4.2j)
+assert hash(3.1 + 4.2j) == hash(3.1 + 4.2j)
+
 # numbers.Complex
 
 a = complex(3, 4)
--- a/vm/src/obj/objcomplex.rs
+++ b/vm/src/obj/objcomplex.rs
@@ -1,5 +1,6 @@
 use num_complex::Complex64;
 use num_traits::Zero;
+use std::num::Wrapping;
 
 use crate::function::OptionalArg;
 use crate::pyhash;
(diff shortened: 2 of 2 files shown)


tests/snippets/builtin_complex.py
```diff
@@ -93,6 +93,8 @@
 assert hash(complex(3.14)) == hash(float(3.14))
 assert hash(complex(-float('inf'))) == hash(-float('inf'))
 assert hash(1j) != hash(1)
+assert hash(3.1 - 4.2j) == -1152921505020448009
```
silmeth: Is using a fixed value to test hashes a good idea? If I understand [the docs](https://docs.python.org/3/reference/datamodel.html#object.__hash__) correctly, the value might and will be different on different platforms:

> **Note:** `hash()` truncates the value returned from an object’s custom `__hash__()` method to the size of a `Py_ssize_t`. This is typically 8 bytes on 64-bit builds and 4 bytes on 32-bit builds. If an object’s `__hash__()` must interoperate on builds of different bit sizes, be sure to check the width on all supported builds. An easy way to do this is with `python -c "import sys; print(sys.hash_info.width)"`.

corona10: @silmeth Yes, That makes sense Do you have any good ideas?

corona10: If we just want to check whether the hash function of complex is crashed or not
````
python
hash(3.1 - 4.2j)
````
this statement will be enough

WDYT?

Comment (silmeth) [reply tests/snippets/builtin_complex.py]: I think I’d just leave the `hash()` statement alone like you suggested, or alternatively just check that it produces the same value twice (`assert hash(3.1 - 4.2j) = hash(3.1 - 4.2j)`).

**Comment by corona10** [comment]
@silmeth I've updated the PR. PTAL

File: vm/src/obj/objcomplex.rs [hunk 2]
```diff
@@ -254,7 +255,7 @@ impl PyComplex {
     fn hash(&self, _vm: &VirtualMachine) -> pyhash::PyHash {
         let re_hash = pyhash::hash_float(self.value.re);
         let im_hash = pyhash::hash_float(self.value.im);
-
-        re_hash + im_hash * pyhash::IMAG
+        let ret = Wrapping(re_hash) + Wrapping(im_hash) * Wrapping(pyhash::IMAG);
```
windelbouwman: Interesting stuff, this wrapping! Thanks!

Issue #447, opened 2019-02-12
Settle on `Py_ssize_t` type and `sys.maxsize` value


silmeth:
The [`sys.maxsize` constant](https://docs.python.org/3.7/library/sys.html#sys.maxsize) defines the maximal value of the `Py_ssize_t` type, which is a type for indexing collections (and thus, it is the theoretical maximal size of given Python implementation collections).

CPython defines `Py_ssize_t` as C `ssize_t` (signed int of the machine’s pointer size, equivalent to Rust `isize`). And justifies the use of signed type in the [PEP-353](https://www.python.org/dev/peps/pep-0353/#why-not-size-t) with two reasons:
1. Python allows negative indexes in many contexts anyway (so using `isize` makes arithmetic on indices easier – but nothing that can’t be done with `BigInt`s and `usize` anyway),
2. the CPython codebase has many places where it depends on indexing variable being signed (`for` loops checking the variable being non-negative, possibly also some error handling by returning `-1`) – this does not concern RustPython in any way.

At the moment, RustPython defines [`sys.maxsize`](https://github.com/RustPython/RustPython/blob/8835b47d1f0d603b6bb1a5f592a6747832d38d8d/vm/src/sysmodule.rs#L133) as the maximal value of `usize`:
```rust
ctx.set_attr(&sys_mod, "maxsize", ctx.new_int(std::usize::MAX));
```
and documents it as ‘the largest supported length of containers’.

But at the same time, it starts to use `isize` for maximal size of containers. At the moment [`list.insert()`](https://github.com/RustPython/RustPython/blob/8835b47d1f0d603b6bb1a5f592a6747832d38d8d/vm/src/obj/objlist.rs#L279-L286) checks if the index fits in `isize`.

Using `usize` sounds like it makes sense because that is what Rust expects for vector indexing (so all indices must be cast to it eventually), but apparently Rust [`Vec` will panic](https://doc.rust-lang.org/src/alloc/vec.rs.html#995-1006) anyway if you try to grow it over `isize::MAX` bytes (because Rust has [no way to allocate more as a single chunk of memory](https://stackoverflow.com/a/32334789/359949)) – on the other hand it lets you get `usize::MAX`-long vectors of 0-bytes sized values (but those afaik have no use in RustPython).

[...]

PR #1124, opened 2019-07-09
sys: Fix sys.maxsize and add sys.maxunicode

corona10:
Current implementation of sys.maxsize is unsigned type
but AFAIK the type of sys.maxsize should be identical with
Py_ssize_t which is known as signed type.

sys.maxunicode is hardcoded from cpython repository
https://github.com/python/cpython/blob/master/Python/sysmodule.c#L2654


windelbouwman:
The maxsize value is under debate. What should be used might change again later. Please see #447

PR #1126, opened 2019-07-10
Add an speedscope output format for profiling 

coolreader18:
I added some documentation to `README.md`, but essentially it now by default outputs a `flamescope.json` file, which you can upload to https://speedscope.app to view profiling information in a *much* nicer viewer than the html that flame outputs. I authored the `flamescope` crate, so if there are any problems that come up I can quickly fix them.

[Here's](https://www.speedscope.app/#profileURL=https://gist.githubusercontent.com/coolreader18/4d601fb73b9382a5c217f3ce9efcf849/raw/9b98b5814c110cd32e90fb9c3e73f3a598a0c9da/flamescope.json) a demo of speedscope with a RustPython profiling output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,6 +105,23 @@ There also are some unit tests, you can run those with cargo:
 $ cargo test --all
 ```
 
+# Profiling
+
+To profile rustpython, simply build in release mode with the `flame-it` feature.
+This will generate a file `flamescope.json`, which you can then view at
+https://speedscope.app.
+
+```sh
+$ cargo run --release --features flame-it script.py
+$ cat flamescope.json
+{<json>}
+```
+
+You can also pass the `--output-file` option to choose which file to output to
+(or stdout if you specify `-`), and the `--output-format` option to choose if
+you want to output in the speedscope json format (default), text, or a raw html
+viewer (currently broken).
+
 # Using a standard library
 
 As of now the standard library is under construction. You can
--- a/README.md
+++ b/README.md
@@ -59,19 +59,6 @@ Documentation HTML files can then be found in the `target/doc` directory.
 
 If you wish to update the online documentation, push directly to the `release` branch (or ask a maintainer to do so). This will trigger a Travis build that updates the documentation and WebAssembly demo page.
 
-# Code organization
-
-- `parser/src`: python lexing, parsing and ast
-- `vm/src`: python virtual machine
-  - `builtins.rs`: Builtin functions
-  - `compile.rs`: the python compiler from ast to bytecode
-  - `obj`: python builtin types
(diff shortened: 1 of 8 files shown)


coolreader18:
You can also install speedscope from npm and view with `speedscope flamescope.json` as well.

README.md
```diff
@@ -105,6 +105,23 @@ There also are some unit tests, you can run those with cargo:
 $ cargo test --all
 ```
 
+# Profiling
```
windelbouwman: Could you create a seperate document, maybe `development.md` with more in detail information for developers? I already moved the wasm part to a seperate document to keep the first readme file manageable.

src/main.rs
```diff
@@ -55,15 +57,25 @@ fn run() -> Result<(), Box<std::error::Error>> {
...
-                 built with feature 'flame-it'",
-            ),
-    );
+    let app = app
+        .arg(
+            Arg::with_name("profile_output")
+                .long("profile-output")
+                .takes_value(true)
+                .help(
+                    "the file to output the profiling information to. present due to being \
```
windelbouwman: I think you could skip the text "present due to being built with feature flame-it".

src/main.rs
```diff
@@ -55,15 +57,25 @@ fn run() -> Result<(), Box<std::error::Error>> {
...
-            .long("profile-output")
-            .takes_value(true)
-            .help(
-                "the file to output the profile graph to. present due to being \
-                 built with feature 'flame-it'",
-            ),
-    );
+    let app = app
+        .arg(
+            Arg::with_name("profile_output")
```
windelbouwman: Next to the profile output, could we have an option enabling / disabling tracing at runtime in addition to the feature at compile time?

src/main.rs
```diff
@@ -92,13 +104,43 @@ fn run() -> Result<(), Box<std::error::Error>> {
     {
```
windelbouwman: The contents of this block could be put into a seperate function which is appropriately named.

src/main.rs
coolreader18: I'm not sure, it'd require some sort of override of the `flame` dependency, and `flame-it` really shouldn't be enabled for a real release build, just developement. We could just disable outputting the data, I guess?

windelbouwman: No, what I mean is that if the flame-it feature is enabled, you get flame trace options at the command line. One of these options is then wether or not you want to use tracing.

I fully agree that trace should be disabled in release build for end users. It's mostly for ourselves to detect slowness of rustpython.

windelbouwman:
This speedscope thing is really something cool. I look forward to finding performance bottlenecks!

PR #1128, opened 2019-07-10
Add time.gmtime function.

windelbouwman:
Several additions to the time module.


vm/src/stdlib/time_module.rs
```diff
@@ -32,11 +41,209 @@ fn time_time(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+}
+
+fn time_strptime(
+    string: PyStringRef,
+    format: OptionalArg<PyStringRef>,
+    vm: &VirtualMachine,
+) -> PyResult<PyStructTime> {
+    let format: String = match format {
+        OptionalArg::Present(format) => format.value.clone(),
+        OptionalArg::Missing => "%a %b %H:%M:%S %Y".to_string(),
```
coolreader18: You can avoid an allocation here if you do:
```rust
let format = match format {
    OptionalArg::Present(ref format) => &format.value,
    OptionalArg::Missing => "...",
};
```

PR #1129, opened 2019-07-10
Cover up bare trait objects with a dyn

coolreader18:


windelbouwman:
What does this change do? What is dyn? Could you explain a bit to me?

coolreader18:
The older syntax for trait object types was just the bare trait name, e.g. `Box<std::any::Any>`, vs the newer one which prefixes the trait name with the `dyn` keyword (`Box<dyn std::any::Any>`). They have the same meaning, but bare trait objects are already deprecated on nightly (with just a warning) so I figured I'd switch our uses to the new version.

PR #1131, opened 2019-07-10
Add testcase for logging module.

windelbouwman:
Things to note:
- the traceback is difficult, since `f_back` is not working properly. @cthulahoops maybe you could have a look at the `f_back` operation?
- `sys.stderr` is not implemented. How should we tackle this?

### DIFF
--- a/tests/snippets/test_logging.py
+++ b/tests/snippets/test_logging.py
@@ -1,0 +1,14 @@
+
+import io
+import sys
+
+f = io.StringIO()
+sys.stderr = f
+
+import logging
+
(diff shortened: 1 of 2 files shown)


tests/snippets/test_logging.py
```diff
@@ -0,0 +1,14 @@
...
+
+f = io.StringIO()
+sys.stderr = f
+
+import logging
+
+logging.error('WOOT')
+logging.warning('WARN')
+
+print(f.getvalue())
```
palaviv: Can you test that the value is actually what was written in the log?

PR #1134, opened 2019-07-11
Add defaultdict to collections; add pprint.py

coolreader18:
Ok, so I did it in Python because I encountered a ton of broken/fragile class machinery when I tried to do it in Rust: subclassing a built-in class but also being a `PyValue` itself weirdness, inconsistency with CPython in how `__new__` vs `__init__` works (it was calling `__new__` with arguments intended for `__init__`, and I still had to work around that in Python), and the final straw, `__missing__` not being detected from `dict.__getitem__`, while `__missing__` did work in Python. In any case `defaultdict` being a native class is mostly just an implementation detail, and there's really no reason why it can't be implemented in Python. We can look at all that at some point later, but we're *really* close to unittest and I'm focusing on that at the moment.

### DIFF
--- a/Lib/collections/__init__.py
+++ b/Lib/collections/__init__.py
@@ -1277,3 +1277,9 @@ class UserString(_collections_abc.Sequence):
         return self.__class__(self.data.translate(*args))
     def upper(self): return self.__class__(self.data.upper())
     def zfill(self, width): return self.__class__(self.data.zfill(width))
+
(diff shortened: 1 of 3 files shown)


windelbouwman:
Nice, I was hitting this before as well! Nice job!

coolreader18:
Sounds good, I'll do that in the morning.

PR #1135, opened 2019-07-11
Add unittest module python files.

windelbouwman:
The module does not function correctly yet, but I figured we could check it in already.

### DIFF
--- a/Lib/contextlib.py
+++ b/Lib/contextlib.py
@@ -1,0 +1,440 @@
+"""Utilities for with-statement contexts.  See PEP 343."""
+import abc
+import sys
+import _collections_abc
+from collections import deque
+from functools import wraps
+
+__all__ = ["contextmanager", "closing", "nullcontext",
+           "AbstractContextManager",
+           "ContextDecorator", "ExitStack",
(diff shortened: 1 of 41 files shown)


coolreader18:
Ah, I was just about to make a PR that included fnmatch and contextlib, with the async parts of contextlib removed so it didn't panic. Do you want to rebase onto that?

coolreader18:
#1136

palaviv:
Is `import unittest` working? The last time I tried we were missing a few libraries.

windelbouwman:
We still lack the argparse library. Thats where it is stuck now. Also, sys.stdout is missing.

palaviv:
I would prefer not adding code that we can't import yet.

coolreader18:
And `argparse` -> `gettext` -> `locale` -> `_locale`, so we're going to need to implement `_locale` next.

coolreader18:
I had the same sort of thought process as palaviv, as in all the stdlib code we commit should be working or at least be able to be imported, but I'm open to discussion.

windelbouwman:
I agree that the code should import without error. I'll close this pull request.

PR #1136, opened 2019-07-11
CPython lib files 2

coolreader18:


palaviv:
I think it will be easier to detect our changes in the `contextlib` if you will first copy the CPython original and make the changes in a separate commit.

PR #1137, opened 2019-07-11
Optimize import

palaviv:
I implemented CPython [PyImport_ImportModuleLevelObject](https://github.com/python/cpython/blob/59ad110d7a7784d53d0b502eebce0346597a6bef/Python/import.c#L1764) trying to improve the import speed. The problem is that it has almost no effect. For the following script:
```py
import time

start = time.time()
import os
print(time.time() - start)
```
I get after the change 0.11813092231750488 and before 0.12146615982055664.
Do you think we should add this change?

### DIFF
--- a/vm/src/vm.rs
+++ b/vm/src/vm.rs
@@ -15,7 +15,6 @@ use crate::bytecode;
 use crate::frame::{ExecutionResult, Frame, FrameRef, Scope};
 use crate::frozen;
 use crate::function::PyFuncArgs;
-use crate::import;
 use crate::obj::objbool;
 use crate::obj::objbuiltinfunc::PyBuiltinFunction;
 use crate::obj::objcode::{PyCode, PyCodeRef};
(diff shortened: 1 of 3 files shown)


windelbouwman:
I think we should import all import mechanisms in rust. Not because of performance, but for simplicity. We will not require the frozen importlib stuff, but do all in rust. I think one of the reasons that cpython uses python for the importlib, is that C is such a pain to program in. Rust is a lot better than C, so we can probably implement more in rust.

That being said, I'd rather not change the current working importlib stuff for performance, unless it improves performance. For performance, I think it is better to look at the load_name opcode, which is used a lot, and consumes a lot of time (accoring to flame-it).

vm/src/vm.rs
```diff
@@ -311,34 +313,29 @@ impl VirtualMachine {
     }
 
     pub fn import(&self, module: &str, from_list: &PyObjectRef, level: usize) -> PyResult {
-        let sys_modules = self.get_attribute(self.sys_module.clone(), "modules")?;
-        sys_modules
```
windelbouwman: Did you remove the caching mechanism (if name is in `sys_modules`) on purpose?

palaviv: Yes. The first thing the `import` function does is to check for that.

palaviv:
I am not sure I agree with changing all the import mechanism to rust but we can discuss this later. I will close this PR as we don't gain enough from this.

PR #1139, opened 2019-07-12
Add optimize flag to ignore assert statements.

windelbouwman:


coolreader18:
I'm not too sure about passing `optimize` as an argument to `VirtualMachine::new`, I think I'd prefer it to just be a field you set on it once you construct it.

windelbouwman:
> I'm not too sure about passing `optimize` as an argument to `VirtualMachine::new`, I think I'd prefer it to just be a field you set on it once you construct it.

That was my first idea, but then it turned out, that the optimization flag is used by the construction of the builtins. I think it makes sense to pass it by constructor anyways, since changing it half way, might give weird results. In the end, I figured that it would actually be good to pass the flag during VM construction, since it's really a flag that determines how the VM operates for the rest of its existence.

coolreader18:
Maybe as a separate method (`with_optimize`)? Passing a single `bool` as a parameter to a method named `new` isn't very descriptive.

windelbouwman:
Thats true, passing booleans to functions is always confusing. Another option I had in mind was a struct with flags. There will be more flags, such as verbosity, which could be added in this immutable struct which is defined at startup.

vm/src/macros.rs
```diff
@@ -188,7 +188,7 @@ macro_rules! py_namespace {
 /// use rustpython_vm::obj::objint::PyInt;
 /// use rustpython_vm::pyobject::PyValue;
 ///
-/// let vm = VirtualMachine::new();
+/// let vm: VirtualMachine = Default::default();
```
coolreader18: A lot of these can be `VirtualMachine::default()` instead of `: VirtualMachine = Default::default()`

PR #1140, opened 2019-07-13
Add Redox OS support, via platform checks!

coolreader18:
This took me a very long time to get working. If there's anything that looks weird, feel free to point it out, but there's probably a reason it's there.


coolreader18:
I'm also planning to delete the `RustPython/redox` and `RustPython/redox-cookbook` in favor of the installer scripts, as I realized that it's a much bigger pain to maintain a fork of a huge project like Redox than it is to just insert the 2/3 lines of code into the needed files.

src/main.rs
```diff
@@ -68,24 +66,37 @@ fn main() {
...
+        import::init_importlib(&vm, true)?;
+
+        if cfg!(target_os = "redox") {
+            let sys_path = vm.get_attribute(vm.sys_module.clone(), "path")?;
+            vm.call_method(
+                &sys_path,
+                "insert",
+                vec![
+                    vm.ctx.new_int(0),
+                    vm.ctx.new_str("/lib/rustpython/".to_string()),
```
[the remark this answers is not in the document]
coolreader18: I'm not sure, there was no way I could find to add an environment variable to redox from the recipe, and having two separate RustPython binaries seems sorta flakey. I can make the code in main not just specific to redox though.

src/main.rs
```diff
@@ -107,24 +105,43 @@ fn main() {
...
-    let result = if let Some(command) = matches.value_of("c") {
-        run_command(&vm, command.to_string())
-    } else if let Some(module) = matches.value_of("m") {
-        run_module(&vm, module)
-    } else {
-        // Figure out if a script was passed:
-        match matches.value_of("script") {
-            None => run_shell(&vm),
-            Some(filename) => run_script(&vm, filename),
+    let res = (|| {
```
windelbouwman: This closure seems redundant. Please either remove it, or create a seperate function. Seperate function might be better, given the size of this function.

vm/src/obj/objfloat.rs
```diff
@@ -11,7 +11,7 @@ use crate::pyobject::{
     TypeProtocol,
 };
 use crate::vm::VirtualMachine;
-use hexf;
+use hexf_parse;
```
windelbouwman: Why was this required?

coolreader18: `hexf` uses `proc_macro` at runtime (I think?) in order to do things like `hexf!("1e10")`, which wasn't supported for targeting redox because it needs to compile to a dylib and redox only supports staticlibs, so I found that since we don't use its macro we could just switch to `hexf_parse`.

redox_recipe.sh
```diff
@@ -0,0 +1,10 @@
+GIT=https://github.com/RustPython/RustPython
+BRANCH=redox
```
windelbouwman: Consider moving the redox related scripts into a seperate folder.

coolreader18: I could actually probably just embed the recipe into the install script, do you think I should do that?

vm/src/obj/objfloat.rs
windelbouwman: `hexf_parse` also makes more sense for our use-case.

redox_recipe.sh
windelbouwman: No, probably not. Having the recipe seperate is good.

PR #1141, opened 2019-07-13
Add mktime function.

windelbouwman:


vm/src/obj/objsequence.rs
```diff
@@ -170,6 +170,31 @@ impl TryFromObject for SequenceIndex {
     }
 }
 
+/// Get the index into a sequence like type. Get it from a python integer
+/// object, accounting for negative index, and out of bounds issues.
+pub fn get_sequence_index(vm: &VirtualMachine, index: &PyIntRef, length: usize) -> PyResult<usize> {
```
palaviv: I think this and `PySliceableSequence::get_pos` do the same. Can you use the same code for both?

windelbouwman: I wanted to re-use this code, but not implement the full sequence trait. Maybe it can be re-used, so that sliced indexing also works.

PR #1144, opened 2019-07-14
Move processing of PYTHONPATH out of vm.

windelbouwman:


src/main.rs
```diff
@@ -89,34 +116,88 @@ fn main() {
...
+    }
+
+    if matches.is_present("quiet") {
+        settings.quiet = true;
+    }
+
+    settings
+}
+
+/// Get environment variable and turn it into integer.
```
coolreader18: If you want the env variable to be Unicode anyway, you can just use `env::var`

windelbouwman: Cool, I'll do it!

PR #1149, opened 2019-07-15
Add codecs, locale, shutil, errno, and encodings modules

coolreader18:
It turns out that `_locale` wasn't strictly necessary in order to import `locale`, and `_codecs` is just a stub, but `codecs` and `encodings` can still be imported, just sans any functionality. `errno` requires error code definitions for non-linux operating systems, but I don't have access to macOS to see what the values in `errorcode` there are.


windelbouwman:
That's a whole bunch of files :). Did you consider using a rust crate, like https://crates.io/crates/encoding_rs for the encodings? This seems reasonable though, in the sense that this will probably work good enough for now, and we can swap python libraries with their rust counterparts later on.

vm/Cargo.toml
```diff
@@ -59,6 +59,7 @@ unic = "0.9.0"
 maplit = "1.0"
 proc-macro-hack = { version = "0.5", optional = true }
 bitflags = "1.1"
+libc = "0.2"
```
windelbouwman: So this was inevitable:). We all want rust instead of c, but, oh well :P

coolreader18: I could use just raw integer literals instead of the libc constants, they're probably not likely to change often. It's just tough when the `errno` module from CPython looks like [this](https://github.com/python/cpython/blob/master/Modules/errnomodule.c).

coolreader18:
Yeah, that's what I was thinking of doing, but given that there are so many files in this PR, I figured that'd best be left to a separate one.

PR #1151, opened 2019-07-15
Simplify import AST in line with CPython.

windelbouwman:
This does not yet account for the triple dot, but aligns the import AST nodes with cpython.

### DIFF
--- a/bytecode/src/bytecode.rs
+++ b/bytecode/src/bytecode.rs
@@ -88,6 +88,9 @@ pub enum Instruction {
         name: String,
         level: usize,
     },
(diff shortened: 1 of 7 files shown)


coolreader18:
I believe this change caused a regression in importing of the form `from . import names`

windelbouwman:
Yes, you are right, this import scheme does not work. I doubt if it worked before?

coolreader18:
Yeah, probably not. Changing `unwrap()` to `unwrap_or_default()` doesn't panic, but it still doesn't find the module.

PR #1156, opened 2019-07-19
Fix panic on import statement with levels but no module name.

windelbouwman:


vm/src/frame.rs
```diff
@@ -923,15 +923,16 @@ impl Frame {
     fn import(
         &self,
         vm: &VirtualMachine,
-        module: &str,
+        module: &Option<String>,
```
coolreader18: You could accept an `Option<&str>` and then just do `.unwrap_or("")` to avoid unnecessary allocation.

windelbouwman: True, but I suspect this will change rather sooner than later. I do not know how the python importlib expects the module when there are only levels given. Should it be `None` or an empty string?

coolreader18: It is an empty string, the temporary fix made was just changing `module.name.clone().unwrap()` to use `unwrap_or_default()`, and that seemed to work fine.

Issue #1142, opened 2019-07-13
[RFC] Add frozen python Lib to WASM build


palaviv:
## Summary

In the current WASM build we can import the stdlib libraries written in Rust (`time`, `hashlib` for example) but we cannot import any python libs  (`collections` for example). I would suggest to freeze all the modules in the `Lib/` folder to allow them to be imported in WASM.

## Detailed Explanation

When building the WASM build we will add all the modules in `Lib/` to the hashmap in `frozen.rs`.

## Drawbacks, Rationale, and Alternatives

- This will increase the WASM build size.

## Unresolved Questions

- Should we add libraries like `io` and `os`?
- Should we add this as a feature flag to the regular build?

windelbouwman:
What we could also do, it extend the pythonpath with an url, and create wasm import hook. So, `sys.path` will contain `https://demo.rustpython.github.io/stdlib`, and when we import the `logging` module, this url is also queried. If a 404 is returned, it means not found, and import resolution continues.

How is this done in brython and skulpt?

coolreader18:
The only problem with that is that http requests are asynchronous, so we'd have to have some way of pausing execution of a module and then coming back to it once the promise resolves.

palaviv:
@windelbouwman I actually thought the same could be done but as @coolreader18 says that will be a problem with the asynchronous requests.
I think that allowing people to import code in the WASM without having to deal with asynchronous requests and promises will be great. I believe we should add the stdlib as default but allowing people to freeze whatever they want into the binary would be a great feature.

PR #1159, opened 2019-07-20
Allow stdlib freeze

palaviv:
As discussed in #1142.
Adding the current `Lib` folder increase the executable by 5MB. I would like to add this to the demo so it will support python stdlib.
I wanted to allow specifying the location of the stdlib dir with a compilation flag but could not find out how to do that. @coolreader18 any suggestions?

### DIFF
--- a/derive/src/compile_bytecode.rs
+++ b/derive/src/compile_bytecode.rs
@@ -23,7 +23,7 @@ use std::env;
 use std::fs;
 use std::path::PathBuf;
 use syn::parse::{Parse, ParseStream, Result as ParseResult};
-use syn::{self, parse2, Lit, LitByteStr, Meta, Token};
+use syn::{self, parse2, Lit, LitByteStr, LitStr, Meta, Token};
 
 enum CompilationSourceKind {
     File(PathBuf),
@@ -156,14 +156,16 @@ pub fn impl_py_compile_bytecode(input: TokenStream2) -> Result<TokenStream2, Dia
 
     let code_obj = input.compile()?;
 
+    let module_name = LitStr::new(&code_obj.source_path, Span::call_site());
+
     let bytes = bincode::serialize(&code_obj).expect("Failed to serialize");
     let bytes = LitByteStr::new(&bytes, Span::call_site());
 
     let output = quote! {
         ({
             use ::rustpython_vm::__exports::bincode;
-            bincode::deserialize::<::rustpython_vm::bytecode::CodeObject>(#bytes)
-                .expect("Deserializing CodeObject failed")
+            hashmap! { #module_name.into() => bincode::deserialize::<::rustpython_vm::bytecode::CodeObject>(#bytes)
+                .expect("Deserializing CodeObject failed")}
         })
     };
 
--- a/derive/src/compile_bytecode.rs
+++ b/derive/src/compile_bytecode.rs
@@ -236,6 +236,7 @@ pub fn impl_py_compile_bytecode(input: TokenStream2) -> Result<TokenStream2, Dia
     let output = quote! {
         ({
             use ::rustpython_vm::__exports::bincode;
+            use ::rustpython_vm::__exports::hashmap;
             hashmap! {
                 #(#modules),*
             }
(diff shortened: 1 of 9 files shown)


derive/src/compile_bytecode.rs
```diff
@@ -55,10 +65,61 @@ impl CompilationSource {
...
+            }
+            CompilationSourceKind::SourceCode(code) => {
+                hashmap! {module_name.clone() => self.compile_string(code, mode, module_name.clone())?}
+            }
+            CompilationSourceKind::Dir(rel_path) => {
+                let mut path = PathBuf::from(
+                    env::var_os("CARGO_MANIFEST_DIR").expect("CARGO_MANIFEST_DIR is not present"),
+                );
+                path.push(rel_path);
+                self.compile_dir(&path, "".to_string(), mode)?
```
coolreader18:
```suggestion
                self.compile_dir(&path, String::new(), mode)?
```

derive/src/compile_bytecode.rs
```diff
@@ -154,16 +225,22 @@ impl Parse for PyCompileInput {
...
+        quote! {  #module_name.into() => bincode::deserialize::<::rustpython_vm::bytecode::CodeObject>(#bytes)
+                .expect("Deserializing CodeObject failed") }
+    });
 
     let output = quote! {
         ({
             use ::rustpython_vm::__exports::bincode;
-            bincode::deserialize::<::rustpython_vm::bytecode::CodeObject>(#bytes)
-                .expect("Deserializing CodeObject failed")
+            hashmap! {
```
coolreader18: You should probably add `hashmap` to `__exports` and use it here.

windelbouwman:
Btw, I opened this issue in pyckitup: https://github.com/pickitup247/pyckitup/issues/4 they might be interested in this.

derive/src/compile_bytecode.rs
```diff
@@ -55,10 +65,61 @@ impl CompilationSource {
...
+                );
+                path.push(rel_path);
+                self.compile_dir(&path, "".to_string(), mode)?
+            }
+        })
+    }
+
+    fn compile_dir(
+        &self,
+        path: &PathBuf,
```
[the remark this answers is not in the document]
palaviv: Why is this better?

derive/src/compile_bytecode.rs
palaviv: What is the purpose of the `__exports`? This is my first time using procedural macro

palaviv: I think I understand what `__exports` do. Added `hashmap`

palaviv:
> Btw, I opened this issue in pyckitup: [pickitup247/pyckitup#4](https://github.com/pickitup247/pyckitup/issues/4) they might be interested in this.

We should probably add the ability to freeze any directory.


> LGTM. An idea for allowing specifying the stdlib dir, you could maybe allow a `DirFromEnv` `CompilationSourceKind`?

I can get the directory from an environment variable but I hoped for something easier.

coolreader18:
Ah yeah, I think that's the best way to go for configuration, as Rust doesn't have anything like that built in. We've done similar configuration in env variables before, e.g. `BUILDTIME_RUSTPYTHONPATH`

derive/src/compile_bytecode.rs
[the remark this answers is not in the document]
coolreader18: `PathBuf` vs `Path` is akin to `String` vs `str` - an owned, allocated buffer vs a slice. Like you should (almost) always prefer `&str` over `&String`, so should you for `&Path` over `&PathBuf`.

coolreader18:
I sorta glazed over this when I reviewed earlier, but I don't think that it should always return a `HashMap`, it should only do that when the source is `Dir`.

palaviv:
> I sorta glazed over this when I reviewed earlier, but I don't think that it should always return a `HashMap`, it should only do that when the source is `Dir`.

I am not sure about that... I prefer that `py_compile_bytecode` will return the same type on all cases. I think it will be confusing otherwise.

PR #1161, opened 2019-07-21
Add a bunch of stdlib modules from CPython

coolreader18:
This is basically everything that can be imported without erroring. I used this mini-script to add them:

```bash
for modfile in /usr/lib/python3.6/*; do
  mod=$(basename "$modfile")
  modname=${mod%.py}
  RUSTPYTHONPATH=/usr/lib/python3.6 rustpython -c "import $modname" >/dev/null 2>&1 && \
    [[ ! -e ~/Projects/RustPython/Lib/$mod ]] && \
    echo "$modname" && \
    cp -r "$modfile" ~/Projects/RustPython/Lib/"$mod"
done
```


Lib/keyword.py
```diff
@@ -0,0 +1,94 @@
+#! /usr/bin/python3.6
+
+"""Keywords (from "graminit.c")
```
windelbouwman: Note, this module we also have as a rust module. I prefer the rust module, since it picks the keywords directly from the lexer, so it's not a copy of the keywords.

Lib/struct.py
```diff
@@ -0,0 +1,15 @@
+__all__ = [
+    # Functions
+    'calcsize', 'pack', 'pack_into', 'unpack', 'unpack_from',
```
windelbouwman: Note, we do not require this module, since we have one written in rust.

Lib/tokenize.py
```diff
@@ -0,0 +1,803 @@
+"""Tokenization help for Python programs.
+
+tokenize(readline) is a generator that breaks a stream of bytes into
```
windelbouwman: This module we have as a rust module. So we would not require this python version as well.

Lib/token.py
```diff
@@ -0,0 +1,143 @@
...
+
+#  This file is automatically generated; please don't muck it up!
+#
+#  To update the symbols in this file, 'cd' to the top directory of
+#  the python source tree after building the interpreter and run:
+#
+#    ./python Lib/token.py
+
+#--start constants--
+ENDMARKER = 0
```
windelbouwman: This module, I am not sure. I think we can skip this one, since it is used together with the tokenize module, which we already have in rust.

Lib/symbol.py
```diff
@@ -0,0 +1,115 @@
+#! /usr/bin/python3.6
+
+"""Non-terminal symbols of Python grammar (from "graminit.h")."""
+
```
windelbouwman: This module I would leave out, as it is specific to how the grammar works in cpython.

windelbouwman:
Please remove some modules that we implement in rust. I marked the ones that are doubles.

Lib/struct.py
coolreader18: Whoops, I thought I had removed it.

hung44zzz:
If you dont stop this project i will sue you'll

windelbouwman:
@hung44zzz if you no longer want those e-mails, please use the unsubscribe link at the bottom of the e-mail you got.

Issue #1090, opened 2019-06-30
.pyc files and bytecode changes


palaviv:
As [mentioned](https://github.com/RustPython/RustPython/pull/1068#issue-292730561) .pyc are working now when `sys.dont_write_bytecode` is set to `False`.
I would like to make the use of `.pyc` files the default but I am not sure how should we follow changes in the bytecode. CPython has a magic number defined in `_bootstrap_external.py` that changes every time the bytecode is changed. RustPython is changing rapidly and I don't think we would like to update a magic number every change in the bytecode.

I have a few suggestion:

- Use the git commit as the magic: Would allow multiple runs with same `.pyc` without code changes.
- Use a random number on main start: Would allow only the same run of main to use the same `.pyc`.

coolreader18:
I think the commit hash sounds good, once we settle on a consistent bytecode format we can do something else.

coolreader18:
Acutally, it might be good to use the hash of the contents of the `bytecode` directory, since we know that if any of that changes then the bytecode format would likely change.

coolreader18:
Yeah, here's some quick and dirty code that'll hash the `rustpython_bytecode` crate:

```rust
use std::collections::hash_map::DefaultHasher;
use std::fs::File;
use std::hash::Hasher;
use std::io::{self, Read};
use walkdir::WalkDir;

fn main() -> io::Result<()> {
    let mut hasher = DefaultHasher::new();
    for entry in WalkDir::new("../RustPython/bytecode/")
        .into_iter()
        .filter_map(Result::ok)
        .filter(|e| e.file_type().is_file())
    {
        for byte in File::open(entry.path())?.bytes() {
            hasher.write_u8(byte?);
        }
    }
    let hash = hasher.finish();
    println!("{}", hash);
    Ok(())
}
```
It's pretty quick, and could probably be put into a `build.rs` file pretty easily.

windelbouwman:
I like the idea of the commit hash only on the rustpython-bytecode crate. Since this is the code being serialized. Instead of hashing the directory ourselves, is there some flag in the rust compiler which maintains a hash of the sourcecode? Cargo will also hash the sourcecode, so there should be a hash of the code somewhere in the rustcompiler we could use.

windelbouwman:
Maybe something like this would be good: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_data_structures/fingerprint/struct.Fingerprint.html

windelbouwman:
[...]

PR #1166, opened 2019-07-22
Write bytecode

palaviv:
As talked about in #1090. This uses the git hash and not a hash of the only the bytecode but I will leave that for later stage.

### DIFF
--- a/vm/src/import.rs
+++ b/vm/src/import.rs
@@ -1,6 +1,7 @@
 /*
  * Import mechanics
  */
+use rand::Rng;
 
 use crate::bytecode::CodeObject;
 use crate::obj::{objcode, objsequence, objstr, objtype};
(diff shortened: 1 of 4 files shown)


vm/src/import.rs
```diff
@@ -23,6 +25,15 @@ pub fn init_importlib(vm: &VirtualMachine, external: bool) -> PyResult {
         let install_external =
             vm.get_attribute(importlib.clone(), "_install_external_importers")?;
         vm.invoke(install_external, vec![])?;
+        // Set pyc magic number to commit hash. Should be changed when bytecode will be more stable.
+        let sys_modules = vm.get_attribute(vm.sys_module.clone(), "modules")?;
```
coolreader18: Any reason not to use `import_frozen()` here?

coolreader18: Nevermind, that would make a completely new module object. What about `vm.import`? That checks `sys.modules` for the module just like you are here, I believe.

palaviv: Done

PR #1167, opened 2019-07-23
Travis CI: Lint Python code with flake8

cclauss:
These tests focus on Python syntax errors and undefined names.

Also capitalize first words and brand names.

$ __flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics__

On the test selection, this PR does not focus on "_style violations_" (the majority of [flake8 error codes](https://lintlyci.github.io/Flake8Rules) that [__python/black__](https://github.com/python/black) can autocorrect).  Instead these tests focus on runtime safety and correctness:
* E9 tests are about Python syntax errors usually raised because flake8 can not build an Abstract Syntax Tree (AST).  Often these issues are a sign of unused code or code that has not been ported to Python 3.  These would be compile-time errors in a compiled language but in a dynamic language like Python they result in the script halting/crashing on the user.
* F63 tests are usually about the confusion between identity and equality in Python.  Use ==/!= to compare str, bytes, and int literals is the classic case.  These are areas where __a == b__ is True but __a is b__ is False (or vice versa).
* F7 tests are logic errors or syntax errors in type hints
* F82 tests are almost always _undefined names_ which are usually a sign of a typo, missing imports, or code that has not been ported to Python 3.  These also would be compile-time errors in a compiled language but in Python a __NameError__ is raised which will halt/crash the script on the user.

This PR adds the __# noqa: F821__ linter directive on lines where our tests create variables in a nonstandard way or where we are actually testing for undefined or deleted variables.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,8 +53,8 @@ matrix:
     - name: Lint Python code with flake8
       language: python
       python: 3.6
-      cache:
-        - pip
+      cache: pip
+      env: JOBCACHE=9
       install: pip install flake8
       script: flake8 . --count --exclude=./.*,./Lib,./vm/Lib  --select=E9,F63,F7,F82 --show-source --statistics
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 matrix:
   fast_finish: true
   include:
-    - name: Run rust tests
+    - name: Run Rust tests
       language: rust
       rust: stable
       cache: cargo
@@ -16,7 +16,7 @@ matrix:
 
     # To test the snippets, we use Travis' Python environment (because
     # installing rust ourselves is a lot easier than installing Python)
-    - name: python test snippets
+    - name: Python test snippets
       language: python
       python: 3.6
       cache:
@@ -28,7 +28,7 @@ matrix:
         - CODE_COVERAGE=false
       script: tests/.travis-runner.sh
 
-    - name: Check rust code style with rustfmt
+    - name: Check Rust code style with rustfmt
       language: rust
       rust: stable
       cache: cargo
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -50,7 +50,15 @@ matrix:
       env:
         - JOBCACHE=8
 
-    - name: publish documentation
+    - name: Lint Python code with flake8
```
windelbouwman: Please include an environment variable which makes the cache unique. The other jobs use JOBCACHE for this, I think it is good to use JOBCACHE=<unique number here> in this case.

.travis.yml
```diff
@@ -50,7 +50,15 @@ matrix:
...
+      language: python
+      python: 3.6
+      cache:
+        - pip
+      install: pip install flake8
+      script: flake8 . --count --exclude=./.*,./Lib,./vm/Lib  --select=E9,F63,F7,F82 --show-source --statistics
```
windelbouwman: What is the reason to lint the example snippets? Does this give us feedback? The snippets are not of high quality code. They are more intended to test certain python behaviors.

windelbouwman:
@cclauss thank you for contributing to this project! I left some remarks in the code. Please take a look at the travis jobcache variable.

.travis.yml
cclauss: env: JOBCACHE=9

.travis.yml
cclauss: I really like the snippets -- I nice test of Python syntax corner cases.  As I wrote above, I added a few __noqa__ directives where needed.  I think it is a good idea to lint the snippet so that contributors are warned when new test cases  contain syntax errors and undefined names.  It these are intentional then noqa can be added and if not the linter should improve the quality of submissions.

Awesome project!!

PR #1169, opened 2019-07-23
Add some extra profiling trace points.

windelbouwman:
I added some extra tracepoints. This gives a nice hint on where to improve.

To run this, use: `cargo run --features flame-it demo.py` and upload flamescope.json into http://speedscope.app

Also moved the get_item_option function into the PyDictRef struct, since it was only used for that case.


vm/src/dictdatatype.rs
```diff
@@ -271,6 +273,7 @@ enum LookupResult {
     Existing(EntryIndex), // Existing record, index into entries
 }
```
coolreader18: The string passed to `flame` is a prefix, so this would appear in the output as `collection_hash::collection_hash`. I'd recommend changing it to `dictdatatype` or removing the prefix altogether.

vm/src/pyobject.rs
```diff
@@ -1021,6 +1021,8 @@ pub trait ItemProtocol {
         vm: &VirtualMachine,
     ) -> PyResult;
     fn del_item<T: IntoPyObject>(&self, key: T, vm: &VirtualMachine) -> PyResult;
```
coolreader18: This should probably be `flame("ItemProtocol")`?

PR #1170, opened 2019-07-23
Add int ratio properties

Askaholic:
Adds properties to int objects:
- `numerator`
- `denominator`

### DIFF
--- a/tests/snippets/ints.py
+++ b/tests/snippets/ints.py
@@ -52,6 +52,11 @@ with assertRaises(ZeroDivisionError):
 # real/imag attributes
 assert (1).real == 1
 assert (1).imag == 0
+# numerator/denominator attributes
+assert (1).numerator == 1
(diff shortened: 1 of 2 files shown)


windelbouwman:
Cool! Thanks!

tests/snippets/ints.py
```diff
@@ -52,6 +52,11 @@
 # real/imag attributes
 assert (1).real == 1
 assert (1).imag == 0
+# numerator/denominator attributes
+assert (1).numerator == 1
+assert (1).denominator == 1
+assert (10).numerator == 10
+assert (1).denominator == 1
```
peddermaster2: did you meant to write `assert (10).denominator == 1` ?  
Overwise you just repeated from 2 lines above.

Askaholic: Yes that was supposed to be a 10

PR #1171, opened 2019-07-24
Implement str iterator support

mpajkowski:
Hello,

I added iteration support for ```str```.

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -1102,6 +1102,16 @@ impl PyString {
             string: zelf,
         }
     }
+
+    #[pymethod(name = "__reversed__")]
+    fn reversed(zelf: PyRef<Self>, _vm: &VirtualMachine) -> PyStringReverseIterator {
+        let begin = zelf.value.chars().count();
+
+        PyStringReverseIterator {
+            position: Cell::new(begin),
+            string: zelf,
+        }
+    }
 }
 
 impl PyValue for PyString {
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -1,6 +1,7 @@
 extern crate unicode_categories;
 extern crate unicode_xid;
 
+use std::cell::Cell;
 use std::char;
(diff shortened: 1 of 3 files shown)


vm/src/obj/objstr.rs
```diff
@@ -1025,6 +1094,14 @@ impl PyString {
         let encoded = PyBytes::from_string(&self.value, &encoding, vm)?;
         Ok(encoded.into_pyobject(vm)?)
     }
+
+    #[pymethod(name = "__iter__")]
+    fn iter(zelf: PyRef<Self>, _vm: &VirtualMachine) -> PyStringIterator {
+        PyStringIterator {
+            position: Cell::new(0),
```
coolreader18: I believe you're missing the `__reversed__`  method somewhere here?

mpajkowski: Actually, I checked it using CPython (3.7.3):

```
>>> hasattr(str, "__reversed__")
False
```

Should I implement ```__reversed__``` despite this fact or leave it unchanged?

mpajkowski: Thanks for your feedback :) I'll deliver corrections today.

mpajkowski:
Thanks for your feedback :) I'll deliver corrections today.

vm/src/obj/objstr.rs
coolreader18: How is the `PyStringReverseIterator` used then?

mpajkowski: Within ```reverse```:

```
string = "123456789"
reversed_iter = reversed(string)
```

coolreader18: Right, but nowhere in your code do you construct a `PyStringReverseIterator`, so there's no way to actually get one in Python code.

mpajkowski: Ok! I'll fix it in a moment.

mpajkowski:
I corrected lint warning and added ```__reverse__``` method - it seems to be an easiest solution.

PR #1172, opened 2019-07-24
[WIP] Add int __sizeof__ implementation

Askaholic:
Leaving this here in hopes that someone will tell me there's actually a really easy way of doing this!

On my system (x86_64 ubuntu 18) I've determined that the compiler lays out the BigInt structs like this:
```
offset item
-------------
0:    BigUInt
   0:    Vec
      0:    RawVec
         0:   Unique<T>
         8:   usize (capacity)
      15:   usize (len)
23:   Sign
```

So I can simply cast the `&BigInt` to a  `&Vec` and get the capacity from it.

TODO: Test snippets

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -506,7 +506,7 @@ impl PyInt {
         let capacity = unsafe {
             // Grab a reference to the underlying vector. This relies on the compiler always placing
             // the BigUInt and its data vector directly at the start of their containing structs.
-            let data = & *(&self.value as *const BigInt as *const Vec<u8>);
+            let data = &*(&self.value as *const BigInt as *const Vec<u32>);
             data.capacity()
         };
 
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -499,6 +499,21 @@ impl PyInt {
         !self.value.is_zero()
     }
 
+    #[pymethod(name = "__sizeof__")]
+    fn sizeof(&self, _vm: &VirtualMachine) -> usize {
+        // We would like to get self.value.data.data.capacity(), however, this is private so we
+        // need to use a hack instead
+        let capacity = unsafe {
+            // Grab a reference to the underlying vector. This relies on the compiler always placing
+            // the BigUInt and its data vector directly at the start of their containing structs.
+            let data = & *(&self.value as *const BigInt as *const Vec<u8>);
+            data.capacity()
+        };
+
+        // We would like to do std::mem::size_of::<BigDigit>(), but of course BigDigit is private.
+        std::mem::size_of::<Self>() + capacity * std::mem::size_of::<u32>()
+    }
+
     #[pymethod]
     fn bit_length(&self, _vm: &VirtualMachine) -> usize {
         self.value.bits()


coolreader18:
Do we have to account for unused extra capacity in the underlying `Vec`? If not, I'd say just use the [`bits`](https://docs.rs/num-bigint/0.2.2/num_bigint/struct.BigInt.html#method.bits) method to find the size of the integer.

Askaholic:
I'm not really sure where to look for the official spec, but from the python3 docs it sounds like it's supposed to return the amount of memory in bytes that the object is holding. 

https://docs.python.org/3/library/sys.html#sys.getsizeof

windelbouwman:
Are you aware that we implemented `sys.getsizeof` here? https://github.com/RustPython/RustPython/blob/master/vm/src/sysmodule.rs#L101

Maybe you could look at the `mem::size_of_val` function?

Maybe the `__sizeof__` method should be implemented on the `object` type, so that all subtypes have a default implementation?

windelbouwman:
A good test would be to check what cpython reports as a size on various ineteger value. Keep in mind however, that this is probably implementation specific, and that we probably cannot make testcases for it in the tests/snippets directory.

Askaholic:
I don't think `mem::size_of_val` helps here because we would still need a reference to the `BigUInt` data `Vec`. Unfortunately you can't really implement this function in general because you can't know how much memory an arbitrary object owns via a reference/pointer.

The cpython implementation in `longobject.c` looks like this:
```c
offsetof(PyLongObject, ob_digit) + Py_ABS(Py_SIZE(self))*sizeof(digit);
```
And in practice that translates to:
```python
>>> (0).__sizeof__()
24
>>> (1).__sizeof__()
28
>>> (1000000000000000).__sizeof__()
32
>>> (10000000000000000000000000000000000000).__sizeof__()
44
```

And my implementation for RustPython:
```python
>>>>> (0).__sizeof__()
32
>>>>> (1).__sizeof__()
36
>>>>> (1000000000000000).__sizeof__()
40
>>>>> (10000000000000000000000000000000000000).__sizeof__()
48
```
I would guess that the cpython ints are 4 bytes smaller because they combine the sign with the number of digits i.e. -4 digits means 4 digits, but the sign of the number is negative. That's why they use `Py_ABS` in the `__sizeof__` implementation.

I think using `bits` we can basically get the vectors `len` in a safe way. It's still a little ugly because it will do a bunch of extra calculations that we need to undo, and also it gets the length and not the capacity.

windelbouwman:
I think we should search in the direction of modifying `PyObject` or the like. This struct holds the custom payload for many types. Maybe we could ask in the rust forums on how to tackle this in a good way? I think the unsafe implementation, as well the manual calculation with the bits are a bit weird. There must be a proper way to solve this in a generic way I think. Can't you use `size_of_val` on the payload? You could add an extra method to the `PyObject` which will return the size of the payload in bytes?

Askaholic:
Well, the reason why you can't write a function for this in general is the same reason why you can't write a hashing function that works for all objects: Some objects have pointers, and how can you tell which pointers are important and which ones aren't? You may be able to come up with something that works most of the time if the language has good support for introspection, which I have no experience with in Rust.

Really the only problem here is that we're using the `big_num` crate, and this crate completely hides away all implementation details from you. And unless we can get a reference to the `BigUInt.data` field we can't make use of `size_of_val`. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ceee481b5ba8180474a2492af32f999b

I see that in the code I've written here I could use `size_of_val` to replace the `capacity * size_of(u32)` calculation, and I should probably change my pointer cast to `*const Vec<u32>` instead.

As an aside. I've been writing a lot of Python, and of course in python there is no public/private access control (the best you can do is use `@property`). So it's extremely frustrating to me, not being able to even get a read-only reference to a data member. Like, what is the language trying to protect me against by preventing me from reading valid data?

windelbouwman:
Hmm. This is too bad. I suggest to implement the `__sizeof__` method on the generic object type, and use size_of_val anyway. This will give some result, but it is incorrect. Please mark this with a TODO comment, and maybe log a warning. In the mean time, we could ask at the rust forums around on how to solve this.

I would really like to prevent that we require a specific sizeof method for each type.

Does this sound reasonable to you?

Askaholic:
What is it that you want me to mark with TODO?

I have a suspicion that implementing a generic `__sizeof__` that works for ALL types of objects is going to be an even harder problem than implementing `__sizeof__` for only one specific object. I might take a look at some other objects or methods though if I feel motivated.

I don't see a lot of point in incorrectly implementing `__sizeof__` though. I think that will just cause the `whats_left.sh` script to give the illusion of the project being further along thprobleman it is.

BTW, a half baked idea I had for a workaround for ints was to maybe store the data vec directly in `PyInt` and only convert it to a `BigInt` when it's actually needed for arithmetic. Doesn't seem particularly practical to me either though.

The other foolproof solution would be to write our own implementation of BigInt (or maybe to fork the num repo) and give data members sane access controls.

windelbouwman:
Sorry, I accidentally closed-reopened this PR.

The reason for me to skip the `__sizeof__` method from the `whats_left.sh` script is to prevent new contributors from hitting this issue. Maybe we could have a `whats_left_new_contributors.sh` script indicating the more do-able items.

I do not think forking the bigint crate is a good idea. @coolreader18 and @palaviv do you have any ideas on this? For the moment, I would like to park this method, and implement other modules first. I think we can postpone this method for some time.

coolreader18:
Forking `bigint` does seem like it would come with a lot of unnecessary maintenance cost. Also, FWIW, I looked into this a bit ago and I came to the same conclusion; `__sizeof__`  would have to be implemented manually for each type, as it's not possible to tell what is and isn't significant.

windelbouwman:
Please see also this issue: https://github.com/rust-num/num-bigint/issues/98

windelbouwman:
I propose to hold off this issue until the bigint crate has support for querying memory. It appears that there might be an option to do this in future.

windelbouwman:
I filed an issue for this in the rust repo: https://github.com/rust-lang/rust/issues/63073

windelbouwman:
@Askaholic do you agree on closing this pull request, and creating an issue for this topic?

Askaholic:
Yea, good idea!

windelbouwman:
Changed into issue #1250

PR #1173, opened 2019-07-24
stdlib.io: Implement seekable()

corona10:
Seekable API is not yet implemented.
Apparently, this PR is not a perfect implementation of seekable but we can fill the rest of implementation in the future.

### DIFF
--- a/vm/src/stdlib/io.rs
+++ b/vm/src/stdlib/io.rs
@@ -128,6 +128,10 @@ impl PyStringIORef {
         }
     }
 
+    fn seekable(self, _vm: &VirtualMachine) -> bool {
+        true
+    }
+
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/io.rs
```diff
@@ -128,6 +128,10 @@ impl PyStringIORef {
         }
     }
 
+    fn seekable(self, vm: &VirtualMachine) -> PyResult {
+        Ok(vm.ctx.new_bool(true))
+    }
```
coolreader18: Minor thing, but functions like these can just return `bool`, some trait magic will work out how to convert it into a PyObject.

corona10:
@coolreader18 Thanks coolreader, I've updated the PR

PR #1177, opened 2019-07-24
Add an example usecase of the parser.

windelbouwman:


examples/parse_folder.rs
```diff
@@ -0,0 +1,70 @@
...
+extern crate log;
+
+use clap::{App, Arg};
+
+use rustpython_parser::{ast, parser};
+use std::path::Path;
+
+fn main() {
+    env_logger::init();
+    let app = App::new("RustPython")
```
silmeth: Since `RustPython` is the name of the whole project and, presumably, its interpreter as a stand-alone application – shouldn’t this be called something else (or at least – more specific)?

windelbouwman: Oops, copy paste error, thanks!

examples/parse_folder.rs
```diff
@@ -0,0 +1,70 @@
...
+        let entry = entry?;
+        let metadata = entry.metadata()?;
+
+        let path = entry.path();
+        if metadata.is_dir() {
+            let x = parse_folder(&path)?;
+            res.extend(x);
+        }
+
+        if metadata.is_file() && path.extension().map(|s| s.to_str().unwrap()) == Some("py") {
```
coolreader18:
```suggestion
        if metadata.is_file() && path.extension().and_then(|s| s.to_str()) == Some("py") {
```

examples/parse_folder.rs
```diff
@@ -0,0 +1,70 @@
...
+                .help("Folder to scan")
+                .required(true),
+        );
+    let matches = app.get_matches();
+
+    let folder = Path::new(matches.value_of("folder").unwrap());
+    if folder.exists() && folder.is_dir() {
+        println!("Parsing folder of python code: {:?}", folder);
+        let res = parse_folder(&folder).unwrap();
+        println!("Processed {:?} files", res.len());
```
coolreader18: Should it print the parsed file or something?

examples/parse_folder.rs
windelbouwman: Nice!

examples/parse_folder.rs
windelbouwman: I have some ideas for metrics:
- parsed lines per second
- print the amount of functions / classes

For now, I just use it to try and scan my cpython lib folder

PR #1178, opened 2019-07-25
Add empty tuple optimization

coolreader18:
Yeah, not all that big of an optimization, but it seemed dumb to be allocating a new `Rc` for something that's essentially empty.

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -146,6 +146,7 @@ pub struct PyContext {
     pub none: PyNoneRef,
     pub ellipsis: PyEllipsisRef,
     pub not_implemented: PyNotImplementedRef,
+    pub empty_tuple: PyTupleRef,
     pub tuple_type: PyClassRef,
     pub tupleiterator_type: PyClassRef,
     pub set_type: PyClassRef,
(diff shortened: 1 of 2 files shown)


windelbouwman:
Hmm, this is interesting, and I would rather not do this. But actually this solves this example:

```python
>>> a = ()
>>> b = ()
>>> a is b
True
```

Please add a test snippet for this special case.

src/main.rs
```diff
@@ -318,7 +318,7 @@ fn run_command(vm: &VirtualMachine, source: String) -> PyResult<()> {
 
 fn run_module(vm: &VirtualMachine, module: &str) -> PyResult<()> {
     debug!("Running module {}", module);
-    vm.import(module, &vm.ctx.new_tuple(vec![]), 0)?;
+    vm.import(module, &vm.ctx.empty_tuple(), 0)?;
```
windelbouwman: Please do not use this empty tuple directly, and make it a private method on `PyContext`. I think it is better to call the new_tuple function with an empty vector, so we have one way to create a tuple.

vm/src/pyobject.rs
```diff
@@ -604,6 +609,10 @@ impl PyContext {
         self.not_implemented.clone().into_object()
     }
 
+    pub fn empty_tuple(&self) -> PyObjectRef {
```
windelbouwman: I think this method could be private, or merged into the new_tuple function.

PR #1180, opened 2019-07-26
freeze stdlib in WASM

palaviv:
As discussed in #1142. We did not reach a conclusion there if we should make this the default. I can remove the default if we will decide against this.

### DIFF
--- a/wasm/lib/Cargo.toml
+++ b/wasm/lib/Cargo.toml
@@ -10,6 +10,9 @@ edition = "2018"
 [lib]
 crate-type = ["cdylib", "rlib"]
 
+[features]
+freeze-stdlib = ["rustpython-vm/freeze-stdlib"]
+
 [dependencies]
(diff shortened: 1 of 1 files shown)


windelbouwman:
Nice! Now we can use stdlib in wasm right?

wasm/lib/Cargo.toml
```diff
@@ -10,6 +10,10 @@ edition = "2018"
 [lib]
 crate-type = ["cdylib", "rlib"]
 
+[features]
+default = ["freeze-stdlib"]
+freeze-stdlib = ["rustpython-vm/freeze-stdlib"]
```
windelbouwman: Nice usage of options! I think the name is good. Another viable name for this option might be `include-stdlib` to indicate the inclusion of the the standard library for people who are unfamiliar with the concept of freezing. But I think most people will probably know the idea of freeze.

This looks good to me!

PR #1182, opened 2019-07-27
Add support for star expression in tuple.

windelbouwman:


parser/src/python.lalrpop
```diff
@@ -491,16 +499,15 @@ ParameterDefs<ArgType>: (Vec<ast::Parameter>, Vec<ast::Expression>) = {
...
-                        &name.arg
-                    );
+                    panic!("non-default argument follows default argument: {}", &name.arg);
                 }
             }
             names.push(name);
         }
 
-        (names, default_elements)
+        //Ok(
```
palaviv: ?

windelbouwman: I tried to get rid of the `panic!` macro usage, by using a Result, like in the `StringGroup` further down in the grammar. This would be nice, since then we would not panic, but fail with a syntaxerror. I could not get it to work though, and this is a leftover.

PR #1187, opened 2019-07-28
Implement dictionary indexing by trait.

windelbouwman:
This add support for indexing the dictionary by string. So instead of creating a PyStringRef, we now can use rust's String type to index a dict.

### DIFF
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -2,6 +2,7 @@ use crate::obj::objbool;
 use crate::pyhash;
 use crate::pyobject::{IdProtocol, PyObjectRef, PyResult};
 use crate::vm::VirtualMachine;
+use num_bigint::ToBigInt;
 /// Ordered dictionary implementation.
 /// Inspired by: https://morepypy.blogspot.com/2015/01/faster-more-memory-efficient-and-more.html
 /// And: https://www.youtube.com/watch?v=p33CVV29OG8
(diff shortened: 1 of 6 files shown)


vm/src/dictdatatype.rs
```diff
@@ -273,23 +274,63 @@ enum LookupResult {
...
+
+    fn do_eq(&self, vm: &VirtualMachine, other_key: &PyObjectRef) -> PyResult<bool> {
+        let result = vm._eq(self.clone(), other_key.clone())?;
+        objbool::boolval(vm, result)
+    }
+}
+
+/// Implement trait for the str type, so that we can use strings
+/// to index dictionaries.
+impl DictKey for String {
```
coolreader18:
```suggestion
impl DictKey for str {
```

vm/src/dictdatatype.rs
```diff
@@ -273,23 +274,63 @@ enum LookupResult {
...
+    fn do_is(&self, _other: &PyObjectRef) -> bool {
+        // No matter who the other pyobject is, we are never the same thing, since
+        // we are a str, not a pyobject.
+        false
+    }
+
+    fn do_eq(&self, vm: &VirtualMachine, other_key: &PyObjectRef) -> PyResult<bool> {
+        // Fall back to PyString implementation.
+        let s = vm.new_str(self.to_string());
+        s.do_eq(vm, other_key)
```
coolreader18: I feel like we could have a more efficient eq implementation here than reallocating the string.

windelbouwman: Absolutely, this involves checking the specific python type to be of `PyStringRef`, and then doing a string compare. I suspect this happens only in a couple of cases, after the first hash compare is a success. I will implement this though, since I think it is a good idea.

vm/src/dictdatatype.rs
windelbouwman: I tried this, without success. I get this compilation error:
```
 --> vm/src/dictdatatype.rs:363:24
    |
363 |         let val = dict.get(&vm, "x").unwrap().unwrap();
    |                        ^^^ doesn't have a size known at compile-time

```

coolreader18: Oh, I think it's because it's trying to make a double-fat pointer, with both a vtable and a slice length, and it doesn't like that

PR #1188, opened 2019-07-28
builtin: Fix hex format of ascii

corona10:
Some range of Unicode characters should be specially handled

### DIFF
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -66,7 +66,7 @@ fn builtin_ascii(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<String> {
         if c.is_ascii() {
             ascii.push(c);
         } else if (c as i64) < 0x10000 {
-            let hex = format!("\\u{:x}", c as i64);
+            let hex = format!("\\u{:04x}", c as i64);
             ascii.push_str(&hex);
         } else {
             let hex = format!("\\U{:08x}", c as i64);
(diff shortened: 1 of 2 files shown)


corona10:
@windelbouwman 

PTAL

corona10:
@windelbouwman

PTAL

windelbouwman:
Hi @corona10 , what does PTA and PTAL mean?

vm/src/builtins.rs
```diff
@@ -64,10 +64,13 @@ fn builtin_ascii(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<String> {
     let mut ascii = String::new();
     for c in repr.value.chars() {
         if c.is_ascii() {
-            ascii.push(c)
+            ascii.push(c);
```
windelbouwman: The trailing semi column `;` might not be required here.

vm/src/builtins.rs
```diff
@@ -64,10 +64,13 @@ fn builtin_ascii(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<String> {
...
+        } else if (c as i64) < 0x10000 {
```
windelbouwman: Consider converting only once to `i64`, like this:
```rust
if c.is_ascii() {
    ...
} else {
    let c = c as i64;
    let hex = if c < 0x10000 {
        format!("\\u{:04x}", c)
    } else {
        format!("\\U{:08x}", c)    
    };
    ascii.push_str(&hex);
}

corona10: Oh, this is great! Thanks!

corona10:
@windelbouwman 

PTAL stands for 'Please take a look'  :) (PTA was my typo.. sorry)
I 've updated your review, Thanks for the review with cool tips!

PR #1189, opened 2019-07-28
signal support

palaviv:
This is a very initial support in signals. I wanted to get early feedback as this is harder then expected in `Rust`.
This is based on the `CPython` signal handling. This [article](http://ramblings.timgolden.me.uk/2013/05/31/how-does-python-handle-signals-on-windows/) is actually explaining very nicely on how it works.


vm/src/frame.rs
```diff
@@ -163,6 +163,7 @@ impl Frame {
     /// Execute a single instruction.
     #[allow(clippy::cognitive_complexity)]
     fn execute_instruction(&self, vm: &VirtualMachine) -> FrameResult {
+        vm.check_signals();
```
windelbouwman: I think this contraption makes sense, it inserts signals nicely at a defined point of time.

vm/src/vm.rs
```diff
@@ -42,6 +43,23 @@ use num_bigint::BigInt;
 #[cfg(feature = "rustpython-compiler")]
 use rustpython_compiler::{compile, error::CompileError};
 
+// Signal triggers
+// TODO: 64
+pub const NSIG: usize = 10;
+
+pub static mut TRIGGERS: [AtomicBool; NSIG] = [
```
windelbouwman: Maybe you could use `Default::default()` here?

vm/src/vm.rs
```diff
@@ -42,6 +43,23 @@ use num_bigint::BigInt;
...
+    AtomicBool::new(false),
+    AtomicBool::new(false),
+    AtomicBool::new(false),
+    AtomicBool::new(false),
+    AtomicBool::new(false),
+    AtomicBool::new(false),
+    AtomicBool::new(false),
```
windelbouwman: Should this be a boolean flag, or a queue of events? Each event triggers a callback, so if we have multiple events happening in a short time, the boolean flag was already true, so it was not handled. I propose to create a queue of events per signal.

vm/src/vm.rs
```diff
@@ -160,6 +179,7 @@ impl VirtualMachine {
             trace_func,
             use_tracing: RefCell::new(false),
             settings,
+            signal_handlers: RefCell::new(HashMap::new()),
```
windelbouwman: You could use `Default::default()` here instead of `RefCell::new(HashMap::new())`

vm/src/vm.rs
windelbouwman: Having this variable as a static global would prevent multiple VirtualMachine instances. We can solve this, by moving the triggers struct into a RefCell. Place this RefCell into the signals.rs file. In the signals.rs file, create a global cell, with a trigger list. Each vm then can register a trigger list to.

Okay, my explanation is poor. Lets see how this evolves further :).

vm/src/vm.rs
```diff
@@ -1138,6 +1158,21 @@ impl VirtualMachine {
...
         self.exceptions.borrow().last().cloned()
     }
+
+    pub fn check_signals(&self) {
+        for (signum, handler) in self.signal_handlers.borrow().iter() {
+            if *signum as usize >= NSIG {
+                panic!("Signum bigger then NSIG");
+            }
+            let triggerd;
+            unsafe {
```
windelbouwman: Please move this unsafe code into the signal module. This function should just loop over the list of triggered events. Also, you probably want to guard this for loop with some sort of lock. Another solution would be to use the `std::sync::mpsc` fifo, which is thread safe. This might be handy, so that the signal handlers still can add events, while this loop still is processing.

vm/src/vm.rs
```diff
@@ -1138,6 +1158,21 @@ impl VirtualMachine {
     pub fn current_exception(&self) -> Option<PyObjectRef> {
         self.exceptions.borrow().last().cloned()
     }
+
+    pub fn check_signals(&self) {
+        for (signum, handler) in self.signal_handlers.borrow().iter() {
+            if *signum as usize >= NSIG {
+                panic!("Signum bigger then NSIG");
+            }
+            let triggerd;
```
coolreader18: This can just be `let triggerd = unsafe { ... };`

vm/src/vm.rs
palaviv: This will be different from the `CPython` implementation. Maybe we should leave this for as future improvement?

vm/src/vm.rs
palaviv: There is not problem that the an event will be added while I process the list. I will just process it in the next run.

vm/src/vm.rs
palaviv: Did not work. I will create a procedural macro in the future to solve this duplication.

vm/src/frame.rs
palaviv: Do you think this is a good point?

vm/src/stdlib/signal.rs
```diff
@@ -0,0 +1,157 @@
...
+        && !vm.isinstance(&handler, &vm.ctx.builtin_function_or_method_type())?
+    {
+        return Err(vm.new_type_error("Hanlder must be callable".to_string()));
+    }
+    let signal = vm.import("signal", &vm.ctx.new_tuple(vec![]), 0)?;
+    let sig_dfl = vm.get_attribute(signal.clone(), "SIG_DFL")?;
+    let sig_ign = vm.get_attribute(signal, "SIG_DFL")?;
+    let signalnum = signalnum.as_bigint().to_i32().unwrap();
+    let signal_enum = signal::Signal::from_c_int(signalnum).unwrap();
+    let sig_handler = if handler.is(&sig_dfl) {
```
palaviv: @windelbouwman @coolreader18 any suggestion on how should I do this? This don't actually works...

palaviv: Ignore this. I had a typo...

palaviv:
I changed the code to use `libc` directly so it will work on Windows. I will test the code on Windows after adding `os.kill`.

vm/src/frame.rs
```diff
@@ -23,6 +23,12 @@ use crate::vm::VirtualMachine;
 use indexmap::IndexMap;
 use itertools::Itertools;
 
+#[cfg(not(target_arch = "wasm32"))]
+use crate::stdlib::signal::check_signals;
+
+#[cfg(target_arch = "wasm32")]
+fn check_signals(vm: &VirtualMachine) {}
```
windelbouwman: Instead of defining a stub function for wasm, make the code on line 172 also depending on configuration.

vm/src/frame.rs
```diff
@@ -163,6 +169,7 @@ impl Frame {
     /// Execute a single instruction.
     #[allow(clippy::cognitive_complexity)]
     fn execute_instruction(&self, vm: &VirtualMachine) -> FrameResult {
+        check_signals(vm);
```
windelbouwman: Add here something like this:

`#[cfg(not(target_arch = "wasm32"))]`

vm/src/stdlib/signal.rs
```diff
@@ -0,0 +1,183 @@
...
+    handler: PyObjectRef,
+    vm: &VirtualMachine,
+) -> PyResult<Option<PyObjectRef>> {
+    if !vm.isinstance(&handler, &vm.ctx.function_type())?
+        && !vm.isinstance(&handler, &vm.ctx.bound_method_type())?
+        && !vm.isinstance(&handler, &vm.ctx.builtin_function_or_method_type())?
+    {
+        return Err(vm.new_type_error("Hanlder must be callable".to_string()));
+    }
+    let signal = vm.import("signal", &vm.ctx.new_tuple(vec![]), 0)?;
```
windelbouwman: For clarity, you might consider naming this variable `signal_module`.

vm/src/stdlib/signal.rs
```diff
@@ -0,0 +1,183 @@
...
+
+extern "C" fn run_signal(signum: i32) {
+    unsafe {
+        TRIGGERS[signum as usize].store(true, Ordering::Relaxed);
+    }
+}
+
+#[derive(Debug)]
+enum SigMode {
+    Ign,
```
windelbouwman: Does this abbreviate `Ignore`? We might want to use `Ignore`, `Default` and `Handler` here.

vm/src/frame.rs

vm/src/stdlib/signal.rs
palaviv: I removed the enum as it is not needed now that I use the same code for windows and linux.

vm/src/frame.rs
windelbouwman: I think so, we would have to wait for builtin functions to finish, but then again, we do not face threading issues.

vm/src/stdlib/signal.rs
```diff
@@ -0,0 +1,165 @@
...
+    for signum in 1..NSIG {
+        let triggerd = unsafe { TRIGGERS[signum].swap(false, Ordering::Relaxed) };
+        if triggerd {
+            let handler = vm
+                .signal_handlers
+                .borrow()
+                .get(&(signum as i32))
+                .expect("Handler should be set")
+                .clone();
+            vm.invoke(handler, vec![vm.new_int(signum), vm.get_none()])
```
windelbouwman: What should happen when a signal handler throws an exception? Do you want to tackle this now, or implement it later?

palaviv: I would prefer to do it later

PR #1193, opened 2019-07-31
byte.decode

yanganto:
- support ascii, utf-8 encoding


vm/src/obj/objbytes.rs
```diff
@@ -420,6 +422,117 @@ impl PyBytesRef {
...
+                                decode_content.push(replacing_char.unwrap())
+                            }
+                        }
+                    }
+                }
+            }
+            // utf-8 is default encoding
+            _ => {
+                let mut p: u32 = 0u32;
+                let mut remaining_bytes = 0;
```
coolreader18: utf-8 is the default, but cpython throws a `LookupError` if you give it an invalid encoding.

yanganto: Got it, I will implement this.

yanganto: Done :)

yanganto:
@coolreader18 thanks for  your review.

PR #1200, opened 2019-08-03
Add basic constant optimization to the compiler

coolreader18:


compiler/src/compile.rs
```diff
@@ -1481,6 +1481,7 @@ impl Compiler {
                 self.set_label(end_label);
             }
         }
+        self.optimize_instruction();
```
windelbouwman: Instead of calling this function here, call it in the `emit` function. The emit function will be called whenever there is a new function emitted to the output.

windelbouwman:
This is interesting. What you implemented here, is essentially a sort of peephole optimizer. I suggest the following.

- Create a seperate file called `peephole.rs` which contains the optimization code. The compiler.rs file is already pretty large, and should preferably not grow further.

- Modify the compilers emit function such that it emit code not into the code object, but into a peephole optimizer. The peephole optimizer also contains an emit function, which emits into the code object. This way, the peephole optimizer can be inserted between the compiler and the codeobject. To do this, there might be the need for another trait: the output stream. This is a an output for instructions, which takes a two functions: emit and flush.

Pseudo code:
```rust
trait OutputStream {
  fn emit(instruction);
  fn flush();
}

impl OutputStream for CodeObjectStream {
  ...
}


struct PeepholeOptimizer {
   last_instructions: Vec<7>, // keep a window of some amount of instructions, to be able to do optimization onto.
}

impl PeepholeOptimizer
  fn new(down_stream: OutputStream) {
     ...
  }
}

impl OutputStream for PeepholeOptimizer {
 ...
}

```

The compiler will then emit to an outputstream, and we can choose between the optimizing one, and the direct one:

Optimizing path:

compiler -> peephole_stream -> codeobject_stream -> code_object

Normal path:

compiler -> codeobject_stream -> code_object

coolreader18:
@windelbouwman how does it look now? I'm not sure how we should expose choosing the `OutputStream` to the public `rustpython_bytecode` API, or if we should at all. Do you have any ideas?

windelbouwman:
`OutputStream` is an internal of the `rustpython_compiler` crate. It should not be exposed to the `rustpython_bytecode` crate. The bytecode should remain just the code carrier, nothing extras.

compiler/src/peephole.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    fn emit(&mut self, instruction: Instruction, meta: InstructionMetadata) {
+        if self.buffer.is_full() {
+            let (instr, meta) = self.buffer.remove(0);
+            Self::inner_emit(&mut self.inner, instr, meta);
+        }
+        // safe because we just checked that: if full then remove one element from it
+        unsafe { self.buffer.push_unchecked((instruction, meta)) };
```
windelbouwman: Is this a performance improvement? I would remove the unsafe block and take the hit of a couple extra instructions. Note that we might do a search in all files for unsafe code, and this would show up. I think it should be a good goal to have no unsafe code, or at least as little as possible.

compiler/src/compile.rs
```diff
@@ -1840,9 +1847,7 @@ impl Compiler {
 
     // Assign current position the given label
     fn set_label(&mut self, label: Label) {
-        let position = self.current_code_object().instructions.len();
-        // assert!(label not in self.label_map)
-        self.current_code_object().label_map.insert(label, position);
+        self.output_stack.last_mut().unwrap().set_label(label)
```
windelbouwman: You might consider creating a helper function `get_output_stream(&mut self)` for this code: `self.output_stack.last_mut().unwrap()`

compiler/src/lib.rs
```diff
@@ -8,4 +8,6 @@ extern crate log;
 
 pub mod compile;
 pub mod error;
+pub mod output_stream;
```
windelbouwman: This modules could be private.

compiler/src/peephole.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    fn inner_emit(inner: &mut O, instruction: Instruction, meta: InstructionMetadata) {
+        inner.emit(instruction, meta.loc);
+        for label in meta.labels {
+            inner.set_label(label);
+        }
+    }
+
+    fn emit(&mut self, instruction: Instruction, meta: InstructionMetadata) {
```
windelbouwman: Does the method take precedence upon the other emit method in the trait implementation? In any case, please rename this method so that it does not name clash with the method in the trait implementation below.

compiler/src/peephole.rs
```diff
@@ -0,0 +1,182 @@
...
+                }
+            }
+            Instruction::UnpackSequence { size } => {
+                let (arg, arg_meta) = self.pop();
+                match arg {
+                    Instruction::BuildTuple {
+                        size: tup_size,
+                        unpack,
+                    } if !unpack && tup_size == size => {
+                        self.emit(
```
windelbouwman: Does this not cause infinite recursion? Since `emit` calls `optimize` again?

vm/src/frame.rs
```diff
@@ -1170,12 +1173,9 @@ impl Frame {
...
     fn pop_multiple(&self, count: usize) -> Vec<PyObjectRef> {
-        let mut objs: Vec<PyObjectRef> = Vec::new();
-        for _x in 0..count {
-            objs.push(self.pop_value());
-        }
-        objs.reverse();
-        objs
+        let mut stack = self.stack.borrow_mut();
+        let stack_len = stack.len();
+        stack.drain(stack_len - count..stack_len).collect()
```
windelbouwman: I'm not sure whether this makes the code more readable. The previous range from `0..count` was clearer in my opinion. I would rather revert this change.

windelbouwman:
@coolreader18 I left some additional review comments. I apologize for delaying this review, but I think it would be good to have this thing in good code-shape.

compiler/src/peephole.rs
```diff
@@ -0,0 +1,182 @@
...
+                        emitconst!([lhs_meta, rhs_meta], String, lhs);
+                    }
+                    (op, lhs, rhs) => {
+                        self.emit(lhs, lhs_meta);
+                        self.emit(rhs, rhs_meta);
+                        self.emit(Instruction::BinaryOperation { op, inplace }, meta);
+                    }
+                }
+            }
+            Instruction::UnpackSequence { size } => {
```
windelbouwman: This basically starts a second pattern for optimization. Could the two patterns be split into seperate functions? I can see the `optimize` function grow rather large if we continue like this.

compiler/src/peephole.rs
```diff
@@ -0,0 +1,182 @@
...
+                        emitconst!([lhs_meta, rhs_meta], Float, lhs.powf(rhs))
+                    }
+                    (op!(Add), lc!(String, mut lhs), lc!(String, rhs)) => {
+                        lhs.push_str(&rhs);
+                        emitconst!([lhs_meta, rhs_meta], String, lhs);
+                    }
+                    (op, lhs, rhs) => {
+                        self.emit(lhs, lhs_meta);
+                        self.emit(rhs, rhs_meta);
+                        self.emit(Instruction::BinaryOperation { op, inplace }, meta);
```
windelbouwman: We now have `emit` which is more or less the opposite of `pop`. I would suggest to change `emit` in this context, to `push`, since we basically push things back on the stack, and do not emit code to the stream. This also solves some confusion (at least for me ;) ) with the two `emit` methods.

compiler/src/peephole.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+}
+
+impl<O> OutputStream for PeepholeOptimizer<O>
+where
+    O: OutputStream,
+{
+    fn emit(&mut self, instruction: Instruction, loc: Location) {
+        self.optimize(
```
windelbouwman: It would be nice to first place the instruction onto the stack, and then call the optimize function afterwards.

compiler/src/peephole.rs
coolreader18: It doesn't, a method declared directly on a type takes precedence over a method with the same name that's part of a trait implemented for that type. I can see how this might be confusing; I'll rename the `PeepholeOptimizer::emit` method.

vm/src/frame.rs
coolreader18: This is more efficient, it preallocates a Vec to put the elements into and it doesn't have to pop backwards and then reverse().

coolreader18:
I added an OptimizationBuffer trait to allow consumers of the crate to define their own container for instructions and use our optimizations on it; however, I'm not sure how useful that is.

windelbouwman:
Looks good to me!

PR #1201, opened 2019-08-03
Add subprocess

palaviv:
In CPython subprocess is actually a pure python lib that uses `os` functions but I think the [subprocess](https://docs.rs/subprocess/0.1.18/subprocess/index.html) library in rust fits better here.

### DIFF
--- a/vm/src/stdlib/subprocess.rs
+++ b/vm/src/stdlib/subprocess.rs
@@ -3,12 +3,14 @@ use std::time::Duration;
 
 use subprocess;
 
-use crate::function::OptionalArg;
+use crate::function::{OptionalArg, PyFuncArgs};
 use crate::obj::objlist::PyListRef;
 use crate::obj::objsequence;
 use crate::obj::objstr::{self, PyStringRef};
 use crate::obj::objtype::PyClassRef;
 use crate::pyobject::{Either, IntoPyObject, PyObjectRef, PyRef, PyResult, PyValue};
+use crate::stdlib::io::io_open;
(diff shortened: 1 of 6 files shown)


windelbouwman:
W00t, this library was modeled after python's API. This is nice!

vm/src/stdlib/subprocess.rs
```diff
@@ -0,0 +1,182 @@
...
+        },
+        None => Ok(subprocess::Redirection::None),
+    }
+}
+
+fn convert_to_file_io(file: &Option<File>, mode: String, vm: &VirtualMachine) -> PyResult {
+    match file {
+        Some(ref stdin) => io_open(
+            vm,
+            PyFuncArgs::new(
```
windelbouwman: There is a `From` implementation for `PyFuncArgs` for `Vec`, so you might be able to reduce this to:
```rust
io_open(
            vm,
            vec![
                    vm.new_int(raw_file_number(stdin.try_clone().unwrap())),
                    vm.new_str(mode),
            ],
)
```

windelbouwman:
@palaviv this looks good to me, except for one minor comment. Please feel free to merge this.

vm/src/stdlib/subprocess.rs
coolreader18: Did you mean to append a `.into()` on the `vec!` call?

windelbouwman: Yes, you are right. I was confused with the `invoke` function which accepts an `Into<PyFuncArgs>` type.

Issue #1176, opened 2019-07-24
Extend the socket module


windelbouwman:
The `socket` module is still incomplete. Many methods and types are missing. To add methods to this module, edit the file `vm/src/stdlib/socket.rs` accordingly.

mpajkowski:
Ok, I'm working on {{get/set} gid,uid,egid and related stuff.

posutsai:
Hi, I would like to contribute to extending the socket module. Are there any remaining methods which are not assigned to anybody? Should I open another issue and specify which method I want, preventing others from doing the same thing? Or which method is the first priority?

windelbouwman:
@posutsai you could add a series of checkboxes in this issue, and mention what you implement. There is not a fixed list, but there is a helper script to determine the missing methods, called [`whats_left.sh`](https://github.com/RustPython/RustPython/blob/master/whats_left.sh)

I think it is good to mention here what you will start on.

palaviv:
> Hi, I would like to contribute to extending the socket module. Are there any remaining methods which are not assigned to anybody? Should I open another issue and specify which method I want, preventing others from doing the same thing? Or which method is the first priority?

I would recommend trying to implement the methods needed by `socketserver` as it is something that would be great to add. From a quick look we should add:

- [ ] gettimeout
- [ ] setsockopt
- [ ] shutdown

posutsai:
@palaviv 
I start with `setsockopt` function. However, I meet issues at the beginning and I can't find a great existing solution on google. I would like to hear your comment or any suggestion.

According to the document on the official [Python website](https://docs.python.org/3/library/socket.html#socket.socket.setsockopt), `setsockopt` takes three or four arguments. However, as I know, Rust doesn't provide function overloading. It is possible to do it with trait if it takes three arguments in each implementation.

I have referred to the [implementation](https://github.com/RustPython/RustPython/blob/master/vm/src/stdlib/socket.rs#L305) of `sendto` and there is also an optional argument `flag` on the [document](https://docs.python.org/3/library/socket.html#socket.socket.sendto). Seems like current implementation simply ignores the optional argument.

Could you give me some hint to implement the function? Sorry for asking such an elementary question, since I just start to learn rust and try to understand it deeper by solving a real problem.

windelbouwman:
Hi @posutsai ,

Your best option at this point would be to use `OptionalArg` in some way like this:

```rust
[...]

PR #1205, opened 2019-08-04
Extend Socket module

Lynskylate:
For #1176 Add socket method gethostname sethostname inet_aton . 
The convert_nix_error method is a common method, maybe it should be moved to exceptions.rs.

### DIFF
--- a/tests/snippets/stdlib_socket.py
+++ b/tests/snippets/stdlib_socket.py
@@ -115,3 +115,9 @@ with assertRaises(OSError):
 
 with assertRaises(OSError):
 	socket.socket(socket.AF_INET, 1000)
+
+with assertRaises(OSError)	:
+	socket.inet_aton("test")
+
+assert socket.inet_aton("127.0.0.1")==b"\x7f\x00\x00\x01"
+assert socket.inet_aton("127.0.0.1")==b"\xff\xff\xff\xff"
--- a/tests/snippets/stdlib_socket.py
+++ b/tests/snippets/stdlib_socket.py
@@ -120,4 +120,4 @@ with assertRaises(OSError)	:
 	socket.inet_aton("test")
 
 assert socket.inet_aton("127.0.0.1")==b"\x7f\x00\x00\x01"
-assert socket.inet_aton("127.0.0.1")==b"\xff\xff\xff\xff"
+assert socket.inet_aton("255.255.255.255")==b"\xff\xff\xff\xff"
--- a/tests/snippets/stdlib_socket.py
+++ b/tests/snippets/stdlib_socket.py
@@ -116,8 +116,16 @@ with assertRaises(OSError):
 with assertRaises(OSError):
 	socket.socket(socket.AF_INET, 1000)
 
-with assertRaises(OSError)	:
+with assertRaises(OSError):
 	socket.inet_aton("test")
 
 assert socket.inet_aton("127.0.0.1")==b"\x7f\x00\x00\x01"
 assert socket.inet_aton("255.255.255.255")==b"\xff\xff\xff\xff"
+
+
+assert socket.inet_ntoa(b"\x7f\x00\x00\x01")=="127.0.0.1"
+assert socket.inet_ntoa(b"\xff\xff\xff\xff")=="255.255.255.255"
+
+with assertRaises(OSError):
+	socket.inet_ntoa(b"\xff\xff\xff\xff\xff")
+
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -24,7 +24,6 @@ sha2 = "0.8"
 sha3 = "0.8"
 blake2 = "0.8"
 
-gethostname = "0.2.0"
 num-complex = { version = "0.2", features = ["serde"] }
 num-bigint = { version = "0.2", features = ["serde"] }
 num-traits = "=0.2.6"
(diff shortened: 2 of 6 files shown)


vm/src/stdlib/socket.rs
```diff
@@ -372,6 +378,30 @@ fn get_addr_tuple(vm: &VirtualMachine, addr: SocketAddr) -> PyResult {
...
+
+#[cfg(unix)]
+fn socket_sethostname(hostname: PyStringRef, vm: &VirtualMachine) -> PyResult<()> {
+    sethostname(hostname.value.as_str()).map_err(|err| convert_nix_error(vm, err))
+}
+
+fn socket_inet_aton(ip_string: PyStringRef, vm: &VirtualMachine) -> PyResult {
+    match ip_string.as_str().parse::<Ipv4Addr>() {
+        Ok(ip_addr) => {
+            let out_bytes: [u8; 4] = unsafe { transmute(u32::from(ip_addr).to_be()) };
```
coolreader18: Try `u32::from(ip_addr).to_be_bytes()`, so that we can avoid `unsafe` use.

vm/src/stdlib/socket.rs
```diff
@@ -372,6 +378,30 @@ fn get_addr_tuple(vm: &VirtualMachine, addr: SocketAddr) -> PyResult {
     Ok(vm.ctx.new_tuple(vec![ip, port]))
 }
 
+#[cfg(unix)]
+fn socket_gethostname(vm: &VirtualMachine) -> PyObjectRef {
+    let mut buf = [0u8; 1024];
+    match gethostname(&mut buf) {
+        Ok(cstr) => vm.new_str(String::from(cstr.to_str().unwrap())),
+        Err(e) => convert_nix_error(vm, e),
```
coolreader18: Did you mean to wrap these in Ok and Err? As it is now, it'll just return the exception object if there's an error.

vm/src/stdlib/socket.rs
palaviv: How about using [octet](https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html#method.octets)?

tests/snippets/stdlib_socket.py
```diff
@@ -115,3 +115,9 @@
 
 with assertRaises(OSError):
 	socket.socket(socket.AF_INET, 1000)
+
+with assertRaises(OSError)	:
+	socket.inet_aton("test")
+
+assert socket.inet_aton("127.0.0.1")==b"\x7f\x00\x00\x01"
+assert socket.inet_aton("127.0.0.1")==b"\xff\xff\xff\xff"
```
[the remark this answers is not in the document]
Lynskylate: Yeah, Thank you

> I guess you wanted "255.255.255.255"?

Thank you.

vm/src/stdlib/socket.rs
Lynskylate: Great advice :). Done

vm/src/stdlib/socket.rs
Lynskylate: Done

Lynskylate: Done

Lynskylate: Done

vm/src/stdlib/socket.rs
```diff
@@ -372,6 +378,30 @@ fn get_addr_tuple(vm: &VirtualMachine, addr: SocketAddr) -> PyResult {
     Ok(vm.ctx.new_tuple(vec![ip, port]))
 }
 
+#[cfg(unix)]
+fn socket_gethostname(vm: &VirtualMachine) -> PyObjectRef {
```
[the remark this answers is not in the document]
Lynskylate: > The [gethostname](https://docs.rs/gethostname/0.2.0/gethostname/) crate is not very popular but the windows implementation looks good. What do you think about using it?

It sounds great. Does RustPython has any constraint to introduce new dependencies?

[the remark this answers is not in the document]
coolreader18: Not particularly, one of the main things we want to do with RustPython is leverage Rust's large crate ecosystem to make developing the native standard library modules; so, feel free to use it.

vm/src/stdlib/socket.rs
```diff
@@ -372,6 +382,44 @@ fn get_addr_tuple(vm: &VirtualMachine, addr: SocketAddr) -> PyResult {
     Ok(vm.ctx.new_tuple(vec![ip, port]))
 }
 
+#[cfg(unix)]
+fn socket_gethostname(vm: &VirtualMachine) -> PyResult {
```
palaviv: You can use the same function as in windows. The `gothostname` module lib support it.

vm/Cargo.toml
```diff
@@ -24,6 +24,7 @@ sha2 = "0.8"
 sha3 = "0.8"
 blake2 = "0.8"
 
+gethostname = "0.2.0"
```
palaviv: This should not be included in WASM. You can look #1201 for example.

Lynskylate: Thanks for your guidance, it helps a lot.

vm/src/stdlib/socket.rs
Lynskylate: Done :)

palaviv:
Great job @Lynskylate

PR #1208, opened 2019-08-05
Add Redox OS compilation to Travis using Redoxer

coolreader18:
[Redoxer](https://gitlab.redox-os.org/redox-os/redoxer/)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,8 +42,9 @@ version = "0.2"
 name = "rustpython"
 path = "src/main.rs"
 
-# Uncommment when you want to compile/check with redoxer
-# [patch.crates-io]
+[patch.crates-io]
+# REDOX START, Uncommment when you want to compile/check with redoxer
 # time = { git = "https://gitlab.redox-os.org/redox-os/time.git", branch = "redox-unix" }
 # libc = { git = "https://github.com/AdminXVII/libc",	branch = "extra-traits-redox" }
 # nix = { git = "https://github.com/AdminXVII/nix",	branch = "add-redox-support" }
+# REDOX END
--- a/vm/src/stdlib/socket.rs
+++ b/vm/src/stdlib/socket.rs
@@ -4,7 +4,7 @@ use std::io::Read;
 use std::io::Write;
 use std::net::{Ipv4Addr, SocketAddr, TcpListener, TcpStream, ToSocketAddrs, UdpSocket};
 
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "redox")))]
 use nix::unistd::sethostname;
 
 use gethostname::gethostname;
@@ -388,7 +388,7 @@ fn socket_gethostname(vm: &VirtualMachine) -> PyResult {
         .map_err(|err| vm.new_os_error(err.into_string().unwrap()))
 }
 
-#[cfg(unix)]
+#[cfg(all(unix, not(target_os = "redox")))]
 fn socket_sethostname(hostname: PyStringRef, vm: &VirtualMachine) -> PyResult<()> {
     sethostname(hostname.value.as_str()).map_err(|err| convert_nix_error(vm, err))
 }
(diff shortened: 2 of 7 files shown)


Cargo.toml
```diff
@@ -42,8 +42,9 @@ version = "0.2"
 name = "rustpython"
 path = "src/main.rs"
 
-# Uncommment when you want to compile/check with redoxer
-# [patch.crates-io]
+[patch.crates-io]
+# REDOX START, Uncommment when you want to compile/check with redoxer
```
windelbouwman: Maybe a patch file on `Cargo.toml` in the redox folder might be a better option here?

windelbouwman:
I'm not too sure about this change. We should try to keep the pull request builds light weight. Is there an option to run this check once in a while? We could use cron or something to run this now and then? Also, is there a redox capable CI-service?

coolreader18:
The funny thing is that the redox job is actually shorter than both the rust test and Python test jobs; also, I was thinking of opening a merge request to redox to add RustPython to their cookbook, and if we do do that it'd be good to make sure that master always compiles for redox. Although maybe we could base it off release, in which case I could just run this test if we're on release.

Cargo.toml
coolreader18: If it were a patch file, it would only work with the same number of lines; it would break if we were to add more lines e.g. dependencies above where it's switching the lines.

palaviv: I would prefer doing a target dependent dependency like in the `vm` Cargo.toml. Some thing like
`[target.'cfg(not(target_os = "redox"))'.dependencies]` and `[target.'cfg(target_os = "redox")'.dependencies]`

palaviv:
> I'm not too sure about this change. We should try to keep the pull request builds light weight. Is there an option to run this check once in a while? We could use cron or something to run this now and then? Also, is there a redox capable CI-service?

I agree with @windelbouwman. I don't think at this stage we should support another target. In case we decide to add another target to the CI I would suggest `macos`

Cargo.toml
coolreader18: I tried doing `target.cfg().patch` syntax at first, but it didn't work for some reason.

coolreader18:
I did set it to only run as a cron job or when on the release branch, does that seem better?

Cargo.toml
palaviv: Why do we need to use the `target.cfg().patch` and not `target.cfg().dependencies`?

palaviv:
> I did set it to only run as a cron job or when on the release branch, does that seem better?

I am not sure this will be any different from the current status. Changes that break `redox` will enter the code and once in a while you will see the cron job fails and fix them...
As for running the job when doing a release. What would we do in case it fails?

coolreader18:
It would at least signal that something is broken. Also, as I mentioned before, it really does finish quickly, so if we wanted to do it every run we could; I could also change `redoxer build` `redoxer check`.

Also, re: failing on release, we could commit a fix to master through a quick PR when the merge master into release build fails, and then merge to release.

palaviv:
I would suggest doing only the cron job and remove the release check. I think a release should depend  only on supported targets.

coolreader18:
Maybe I could manage a redox-release branch or something that could be the source for the redox recipe.

Cargo.toml
windelbouwman: This might be a nice option, so we can use feature flags, or even let cargo figure out the os, and use the `target_os` flag.

vm/src/stdlib/socket.rs
```diff
@@ -449,6 +449,7 @@ fn extend_module_platform_specific(_vm: &VirtualMachine, module: PyObjectRef) ->
 fn extend_module_platform_specific(vm: &VirtualMachine, module: PyObjectRef) -> PyObjectRef {
     let ctx = &vm.ctx;
 
+    #[cfg(not(target_os = "redox"))]
```
windelbouwman: You might consider modifying line 443 somewhat more above. This function will then be enabled for `not unix or redox`.

coolreader18: No, that's intentional. I'm assuming in the future that there will be unix-only parts of the module that *are* compatible with redox, so once those are added they can all stay in the `unix` `extend_module_platform_specific` fn like for [`os`](https://github.com/RustPython/RustPython/blob/0129bb5c7589cde5aedb5f416aad4830a89b8792/vm/src/stdlib/os.rs).

windelbouwman: Okay.

PR #1210, opened 2019-08-07
Add os.ttyname

ypyf:


vm/src/stdlib/os.rs
```diff
@@ -862,6 +863,22 @@ pub fn os_openpty(vm: &VirtualMachine) -> PyResult {
     }
 }
 
+#[cfg(unix)]
+pub fn os_ttyname(fd: PyIntRef, vm: &VirtualMachine) -> PyResult {
+    use libc::ttyname;
+    unsafe {
```
windelbouwman: Why is this `unsafe` block required? In any case, please try to reduce the scope of this `unsafe` block. Probably only the `libc` related call is unsafe, the other code is safe rust code.

ypyf: Yes, I see. I can fix it.

vm/src/stdlib/os.rs
```diff
@@ -862,6 +863,22 @@ pub fn os_openpty(vm: &VirtualMachine) -> PyResult {
     }
 }
 
+#[cfg(unix)]
+pub fn os_ttyname(fd: PyIntRef, vm: &VirtualMachine) -> PyResult {
+    use libc::ttyname;
+    if let Some(fd) = fd.as_bigint().to_i32() {
+        let name = unsafe { ttyname(fd) };
+        if !name.is_null() {
```
windelbouwman: Note: you might want to swap if-arms if the condition contains a `!` for readability:

```rust
        if name.is_null() {
            Err(vm.new_os_error(Error::last_os_error().to_string()))
        } else {
            let name = unsafe { CStr::from_ptr(name) }.to_str().unwrap();
            Ok(vm.ctx.new_str(name.to_owned()))
        }
```

This is a slight improvement, but not required. I think this change is good now.

PR #1220, opened 2019-08-07
Fix a bunch of problems with getting argparse to actually work

coolreader18:
What's preventing `rustpython -m unittest.__main__` from working now is `sys.std{out,in,err}`, which I believe @yanganto is working on.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -50,7 +50,15 @@ fn parse_arguments<'a>(app: App<'a, '_>) -> ArgMatches<'a> {
         .version(crate_version!())
         .author(crate_authors!())
         .about("Rust implementation of the Python language")
-        .arg(Arg::with_name("script").required(false).index(1))
+        .usage("rustpython [OPTIONS] [-c CMD | -m MODULE | FILE | -] [PYARGS]...")
+        .arg(
+            Arg::with_name("script")
+                .required(false)
+                .multiple(true)
+                .min_values(1)
+                .allow_hyphen_values(true)
+                .value_names(&["script", "args..."]),
+        )
         .arg(
             Arg::with_name("optimize")
                 .short("O")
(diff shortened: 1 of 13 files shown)


vm/src/sysmodule.rs
```diff
@@ -163,6 +167,12 @@ fn sys_exc_info(vm: &VirtualMachine) -> PyResult {
     }))
 }
 
+// TODO: raise a SystemExit here
+fn sys_exit(code: OptionalArg<i32>, _vm: &VirtualMachine) -> PyResult<()> {
+    let code = code.unwrap_or(0);
+    std::process::exit(code)
```
windelbouwman: What does this do in the WASM build? Close browser?

windelbouwman:
This is a big change. It looks pretty okay, I left some minor comment.

src/main.rs
```diff
@@ -47,10 +47,39 @@ fn main() {
 
 fn parse_arguments<'a>(app: App<'a, '_>) -> ArgMatches<'a> {
     let app = app
+        .setting(AppSettings::TrailingVarArg)
         .version(crate_version!())
         .author(crate_authors!())
         .about("Rust implementation of the Python language")
-        .arg(Arg::with_name("script").required(false).index(1))
+        .usage("rustpython [OPTIONS] [-c CMD | -m MODULE | FILE | -] [PYARGS]...")
```
windelbouwman: I'd rather not advertise this usage, since it is not true, we do not support `-`.

vm/src/stdlib/re.rs
```diff
@@ -77,9 +77,8 @@ impl PyValue for PyPattern {
 /// Inner data for a match object.
 #[pyclass(name = "Match")]
 struct PyMatch {
-    start: usize,
-    end: usize,
-    // m: Match<'t>,
+    haystack: PyStringRef,
+    captures: Vec<Option<(usize, usize)>>,
```
windelbouwman: `Option<(usize, usize)` should be a seperate struct, `Capture` for example, to keep thing simple.

PR #1225, opened 2019-08-10
Fix compilation of nested boolean operations

coolreader18:
I ran into this in `unittest/runner.py`, which has this code:

```py
(run, run != 1 and "s" or "", timeTaken)
```

It was duplicating the False from the `!=` expression but not popping it off the stack, so when BuildTuple was called it included that original False as the first element of the tuple.

### DIFF
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -1303,11 +1303,9 @@ impl<O: OutputStream> Compiler<O> {
                             self.emit(Instruction::Pop);
                         }
                         if let Some(false_label) = false_label {
-                            self.emit(Instruction::Duplicate);
                             self.emit(Instruction::JumpIfFalse {
                                 target: false_label,
                             });
-                            self.emit(Instruction::Pop);
                         }
                     }
                 }


compiler/src/compile.rs
```diff
@@ -1303,11 +1303,9 @@ impl<O: OutputStream> Compiler<O> {
                             self.emit(Instruction::Pop);
                         }
                         if let Some(false_label) = false_label {
-                            self.emit(Instruction::Duplicate);
```
windelbouwman: Please add the test case as you mentioned in the pull request comment.

compiler/src/compile.rs
```diff
@@ -1303,11 +1303,9 @@ impl<O: OutputStream> Compiler<O> {
                             self.emit(Instruction::Pop);
```
windelbouwman: I would expect the same change to be required for this line. Still I do not fully grasp the problem.

windelbouwman:
Suggested test snippet:

```python
run = 1
timeTaken = 33
r = (22, run, run != 1 and "s" or "", timeTaken)
assert r == (22, 1, '', 33)
```

On rustpython, this now gives: `(1, False, '', 33)`.

compiler/src/compile.rs
windelbouwman: This `compile_test` is fairly complex. It implements [short circuit evaluation](https://en.wikipedia.org/wiki/Short-circuit_evaluation).

Please check carefully the logic of this function. I believe the `EvalContext::Expression` is for the case where an expression on the stack is desired.

It is probably a good idea to expand on our snippets here, so that we cover all cases.

coolreader18:
Alright, it's taken me a while to fully grasp the issue, but I think I've got it.

Usually, when `and` short-circuits, it leaves a value on the stack, because it doesn't need to think about what's on the rhs. However, when and `and` is the lhs of an `or`, the `or` gives the short-circuit label to be the beginning of the rhs of the `or`, which doesn't expect a value to be on the stack, because when `or` doesn't short-circuit it pops the value off the stack before moving to the rhs. Basically, we jump straight from the lhs of the `and` to the rhs of the `or`, and it's hard to tell just in `compile_test` whether or not we should pop a value off the stack before doing so, because if we were just jumping from the lhs of the `and` to the end of the expression there *should* be a value there, but when we're jumping to the rhs of the `or` there *shouldn't*.

coolreader18:
Do you have any ideas for how to tackle this @windelbouwman? Maybe a flag in `EvalContext::Expression`?

windelbouwman:
No, I don't think an extra flag is a good option. Basically, the switch `Statement/Expression` is the flag. In the case of `Statement`, there must be no expression left on the stack, in the case of `Expression`, there must be a single value on the stack, which caused the True or False result.

I do not have a good solution for this, except that I would try to first think of all the different test cases, and really write them out. Only then we might get the logic right, if we cover all combinations. I think it's about 8 different combinations, so we should do this first.

windelbouwman:
Btw, the longer I look at this `compile_test` function, the more issues I see.

For example:

If you set both `true_label` and `false_label`, in the `EvalContext::Statement` case, the stack is potentially popped twice.

windelbouwman:
I can have a look at it if you like? I looked at the CPython sourcecode, I think we can get ideas from there. I'll check micropython as well how they did it.

cthulahoops:
Sorry about leaving this mess. I got stuck on these bugs and decided to take a break and come back to it (which never happened).

windelbouwman:
@cthulahoops no problem! All will be fine :).

windelbouwman:
Okay, I got a pretty good idea about how to solve this. Will open another pull request.

coolreader18:
I'll close this, then. :)

windelbouwman:
@coolreader18 did you intend to merge this? Or close this pull request?

coolreader18:
Whoops, I meant to close it. I'll undo it.

PR #1227, opened 2019-08-10
Add unix constant for os module

yjhmelody:


vm/src/stdlib/os.rs
```diff
@@ -1051,6 +1051,18 @@ fn extend_module_platform_specific(vm: &VirtualMachine, module: PyObjectRef) ->
         "setgid" => ctx.new_rustfunc(os_setgid),
         "setpgid" => ctx.new_rustfunc(os_setpgid),
         "setuid" => ctx.new_rustfunc(os_setuid),
+        "O_DSYNC" => ctx.new_int(4096),
+        "O_RSYNC" => ctx.new_int(1_052_672),
+        "O_NDELAY" => ctx.new_int(0o0_004_000),
+        "O_NONBLOCK" => ctx.new_int(0o0_004_000),
+        "O_NOCTTY" => ctx.new_int(0o0_000_400),
+        "O_CLOEXEC" => ctx.new_int(0o2_000_000),
+        "POSIX_FADV_NORMAL" => ctx.new_int(0),
```
palaviv: Are we supporting any function that accepts `POSIX_FADV_*` advice?

yjhmelody: Well, I will remove posix_*.

vm/src/stdlib/os.rs
```diff
@@ -1051,6 +1051,18 @@ fn extend_module_platform_specific(vm: &VirtualMachine, module: PyObjectRef) ->
         "setgid" => ctx.new_rustfunc(os_setgid),
         "setpgid" => ctx.new_rustfunc(os_setpgid),
         "setuid" => ctx.new_rustfunc(os_setuid),
+        "O_DSYNC" => ctx.new_int(4096),
```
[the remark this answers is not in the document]
yjhmelody: Oh, yes. And I feel strange about that there are so many constants that can be found in `libc`. Why not use `libc`?

palaviv:
Looking more on this I think that we can use `custom_flags` for windows and linux and remove `FileCreationFlags` and use the flag parameter directly. What do you think?

vm/src/stdlib/os.rs
```diff
@@ -99,7 +101,12 @@ bitflags! {
         const O_CREAT = 0o0000_0100;
         const O_EXCL = 0o0000_0200;
         const O_APPEND = 0o0000_2000;
-        const O_NONBLOCK = 0o0000_4000;
+        const O_NONBLOCK = libc::O_NONBLOCK as u32;
```
palaviv: Let's get all flags from `libc`

yjhmelody: I think if all constants are replaced by libc, the rust and python code should be closer and better.

yjhmelody:
> Looking more on this I think that we can use `custom_flags` for windows and linux and remove `FileCreationFlags` and use the flag parameter directly. What do you think?

There may be a big change here, maybe you can merge it first, then remove BitFlag.

vm/src/obj/objint.rs
```diff
@@ -670,7 +670,7 @@ fn int_new(cls: PyClassRef, options: IntOptions, vm: &VirtualMachine) -> PyResul
 pub fn to_int(vm: &VirtualMachine, obj: &PyObjectRef, base: u32) -> PyResult<BigInt> {
     match_class!(obj.clone(),
         s @ PyString => {
-            i32::from_str_radix(s.as_str(), base)
+            i32::from_str_radix(s.as_str().trim(), base)
```
yjhmelody: I fix a part of #1228

vm/src/obj/objint.rs
```diff
@@ -6,6 +6,7 @@ use num_traits::{One, Pow, Signed, ToPrimitive, Zero};
 
 use crate::format::FormatSpec;
 use crate::function::{KwArgs, OptionalArg, PyFuncArgs};
+use crate::obj::objtype::PyClassRef;
```
windelbouwman: This probably conflicts with the automatic sorting of `cargo fmt`.

vm/src/stdlib/os.rs
windelbouwman: @palaviv Was the `FileCreationFlags` structure created as a sort of wrapper for libc, so that we could implement on other platforms without libc?

windelbouwman:
Hi @yjhmelody , you actually fixed two issues, maybe it makes sense to create two seperate branches for this. The `int` fix seems good, but the other fix might take some more time to discuss.

yjhmelody:
> Hi @yjhmelody , you actually fixed two issues, maybe it makes sense to create two seperate branches for this. The `int` fix seems good, but the other fix might take some more time to discuss.

Well, I try to split it to 2 PR.

vm/src/stdlib/os.rs
palaviv: No. I did not see the `OpenOptionsExt`. Looking now that is a better way to go.

palaviv:
> There may be a big change here, maybe you can merge it first, then remove BitFlag.

Once you use `custom_flags` everything else should go so that should not be a big change. I can help if you have any questions.

yjhmelody:
`custom_flags` only support unix and window. So how to deal with other platform? Just replace `FileCreationFlags` by libc?

palaviv:
> `custom_flags` only support unix and window. So how to deal with other platform? Just replace `FileCreationFlags` by libc? @palaviv

I think unix and windows is good enough. Rust has only these 2 [target family's](https://doc.rust-lang.org/reference/conditional-compilation.html#target_family).

yjhmelody:
@palaviv Fine. I will try to update, and ask for your help when meet any problem.

yjhmelody:
@palaviv 
Unfortunately, libc exports types and constants of platform in compiling time. I cannot use some constants for windows env. And I found `OpenOptionsExt` need different types in win and unix.

vm/src/stdlib/os.rs
```diff
@@ -128,30 +121,15 @@ pub fn os_open(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
+    let mut options = OpenOptions::new();
 
-    if flags.contains(FileCreationFlags::O_WRONLY) {
-        options = options.write(true);
-    } else if flags.contains(FileCreationFlags::O_RDWR) {
-        options = options.read(true).write(true);
+    if cfg!(unix) {
+        let flags = objint::get_value(flags).to_i32().unwrap();
+        options.custom_flags(flags);
     } else {
```
yjhmelody: image


image

image

I feel strange about the result compilint in ci.

palaviv:
The problem is that you use the `cfg!` macro syntax. In this case it compiles all the code for both `unix` and `windows` and check the condition in runtime. What you would want to do is conditional compilation  by creating an inner function that will be called `_set_file_mode` for example and will compile differently for each operating system. You can look at the `raw_file_number` function for exmaple on how to do that.

palaviv:
Hi @yjhmelody,
This change is very good and there is not much left to do. Can I help in any way so we can push this in?

yjhmelody:
@palaviv Sorry for long time not to update code, I will try update tomorrow.

yjhmelody:
Well, I don't know how to fix these errors...

coolreader18:
The issue is that `OpenOptions` errors if you don't have read, write, or append set on it, even if it's in the custom flags. This is because it [stores it in a separate field than the custom flags](https://doc.rust-lang.org/src/std/sys/unix/fs.rs.html#395-428). Try doing `options.read(flags & libc::O_RDONLY != 0).write(flags & libc::O_WRONLY != 0).append(flags & libc::O_APPEND != 0)`, and then call `_set_file_model`.

coolreader18:
Okay, I think this is the last thing: because `O_RDONLY` is usually `0`, and `x & 0` always is `0` (because 0 always matches), `x & 0 != 0` fails. So, what you should actually do is `options.read(flags & libc::O_RDONLY == libc::O_RDONLY); ...etc`. That should fix the issue.

Oh, [clippy even pointed it out](https://travis-ci.org/RustPython/RustPython/jobs/586340753#L215-L222)!

yjhmelody:
I know that. But I have to say my local compiler reported even more errors.
image

coolreader18:
Oh, that might be because those are in Windows-only functions, while CI only runs clippy on Linux. Also, it should be `flags & libc::O_RDONLY == libc::O_RDONLY`, not `flags & libc::O_RDONLY != libc::O_RDONLY`.

coolreader18:
Looks good, thanks again for contributing!

coolreader18:
Oh whoops, I thought the tests were passing. I'll fix this up real quick.

palaviv:
@coolreader18 are you fixing this or should I look into this?

coolreader18:
I fixed it on unix but I have no idea what's wrong with Windows; feel free to look into it.

Issue #1212, opened 2019-08-07
Found NameError case that is not properly handled


cmarshall108:
The result below is from RustPython:

```
>>>>> def t() -> void:
.....     pass
..... 
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:347:21
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

The result below is from CPython 3:
```
>>> def t() -> void:
...     pass
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
NameError: name 'void' is not defined
```

corona10:
I am looking now this issue.

PR #1231, opened 2019-08-11
fix int fn

yjhmelody:
#1228

PR #1232, opened 2019-08-11
gh-1212: Fix scope_for_name to catch NameError properly

corona10:
fix: #1212

### DIFF
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -1877,7 +1877,9 @@ impl<O: OutputStream> Compiler<O> {
     fn lookup_name(&self, name: &str) -> &Symbol {
         // println!("Looking up {:?}", name);
         let scope = self.scope_stack.last().unwrap();
-        scope.lookup(name).unwrap()
+        scope.lookup(name).expect(
+            "The symbol must be present in the symbol table, even when it is undefined in python.",
+        )
     }
 
     // Low level helper functions:
--- a/compiler/src/symboltable.rs
+++ b/compiler/src/symboltable.rs
@@ -244,13 +244,11 @@ impl SymbolTableBuilder {
             } => {
                 self.scan_expressions(decorator_list)?;
                 self.register_name(name, SymbolRole::Assigned)?;
-
-                self.enter_function(args)?;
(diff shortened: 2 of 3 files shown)


corona10:
@windelbouwman 
Please take a look.
The CI test failure looks like not related to this patch.
(I found that https://github.com/RustPython/RustPython/pull/1231 is also failed with the same test)

compiler/src/compile.rs
```diff
@@ -1874,10 +1879,10 @@ impl<O: OutputStream> Compiler<O> {
         assert!(scope.sub_scopes.is_empty());
     }
 
-    fn lookup_name(&self, name: &str) -> &Symbol {
+    fn lookup_name(&self, name: &str) -> Option<&Symbol> {
         // println!("Looking up {:?}", name);
         let scope = self.scope_stack.last().unwrap();
-        scope.lookup(name).unwrap()
```
windelbouwman: The failure of this unwrap is caused elsewhere. The symbol `void` must be present in a symbol table. It might be better to modify `symboltable.rs` to fix this particular issue.

My suggestion:
- change the `unwrap` call into an `except("The symbol must be present in the symbol table, even when it is undefined in python.")` call. In other words, the unwrap here must succeed, otherwise an internal error was made, and we should not leak internal rustpython errors into python runtime errors, but fail with an assertion.
- Move the lines concering the return type annotation (https://github.com/RustPython/RustPython/blob/master/compiler/src/symboltable.rs#L251) before the `enter_function` call. This might also fix the issue.

If anything is unclear, please let me know.

bytecode/src/bytecode.rs
```diff
@@ -63,6 +63,7 @@ pub enum NameScope {
     Local,
     NonLocal,
     Global,
+    Unknown,
```
windelbouwman: Please revert this enum variant. See my other comment.

windelbouwman:
The build might be fixed on the master branch. If you merge the master branch again into your own branch, the build might succeed.

corona10:
@windelbouwman 
Thanks for the review! Please take a look

PR #1234, opened 2019-08-11
Improve the situation regarding boolean operations.

windelbouwman:
This implements the boolean short circuit logic in line with cpython and micropython.

### DIFF
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -15,6 +15,7 @@ use rustpython_parser::{ast, parser};
 
 type BasicOutputStream = PeepholeOptimizer<CodeObjectStream>;
 
+/// Main structure holding the state of compilation.
 struct Compiler<O: OutputStream = BasicOutputStream> {
     output_stack: Vec<O>,
     scope_stack: Vec<SymbolScope>,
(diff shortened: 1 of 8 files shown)


compiler/src/compile.rs
```diff
@@ -1256,66 +1248,120 @@ impl<O: OutputStream> Compiler<O> {
...
+    ///
+    /// The idea is to jump to a label if the expression is either true or false
+    /// (indicated by the condition parameter).
+    fn compile_jump_if(
         &mut self,
         expression: &ast::Expression,
-        true_label: Option<Label>,
-        false_label: Option<Label>,
-        context: EvalContext,
+        condition: bool,
```
coolreader18: I'd recommend moving `condition` to the first argument and removing the second underscore in the function name so it can read better, e.g. you can see `compile_jumpif(true, ...)` compiles a `JumpIfTrue`.

windelbouwman: The way I intended this to read was: `compile_jump_if(expression, true, target)` means: Jump if the expression is true to target.

windelbouwman: I'll leave this for now like it is. Thanks for the idea!

PR #1236, opened 2019-08-12
Add dis example

coolreader18:


vm/src/stdlib/symtable.rs
```diff
@@ -28,7 +27,9 @@ fn symtable_symtable(
     mode: PyStringRef,
     vm: &VirtualMachine,
 ) -> PyResult<PySymbolTableRef> {
-    let mode = get_compile_mode(vm, &mode.value)?;
+    let mode = mode.as_str().parse().map_err(|_| {
+        vm.new_value_error("symtable() arg 3 must be 'exec' or 'eval' or 'single'".to_owned())
```
windelbouwman: I think it is fine to just display the inner error here, and slightly deviate from the exact CPython error message.

vm/src/builtins.rs
```diff
@@ -125,9 +124,11 @@ fn builtin_compile(args: CompileArgs, vm: &VirtualMachine) -> PyResult<PyCodeRef
         Either::B(bytes) => str::from_utf8(&bytes).unwrap().to_string(),
     };
 
-    let mode = get_compile_mode(vm, &args.mode.value)?;
+    let mode = args.mode.as_str().parse().map_err(|_| {
+        vm.new_value_error("compile() mode must be 'exec', 'eval' or 'single'".to_owned())
```
windelbouwman: I think you can use the display method of the parse error for the error message.

examples/dis.rs
```diff
@@ -0,0 +1,91 @@
+/// This an example usage of the rustpython_compiler crate.
+/// This program reads, parses, and compiles a file you provide
```
windelbouwman: Woot! Dis is handy! (pun intended :P).

examples/dis.rs
```diff
@@ -0,0 +1,91 @@
...
+use clap::{App, Arg};
+
+use rustpython_compiler::compile;
+use std::error::Error;
+use std::fs;
+use std::path::Path;
+
+fn main() {
+    env_logger::init();
+    let app = App::new("parse_folders")
```
windelbouwman: You might want to change the app name and about box.

vm/src/vm.rs
```diff
@@ -877,7 +877,7 @@ impl VirtualMachine {
     pub fn compile(
         &self,
         source: &str,
-        mode: &compile::Mode,
+        mode: compile::Mode,
```
windelbouwman: I like this change, but how did you get the idea to change this? Is this more efficient? Since the `Mode` is a simple enum, we can pass it by value easily, I agree.

coolreader18: I've been thinking about it for a while, because as you said, there's really no reason to pass it by reference rather than value. It was also sorta annoying when passing them around to decide whether a function should take an `&Mode` or just a `Mode` and then pass a reference to that to `compile`.

PR #1239, opened 2019-08-12
Fix int [r]divmod, Add rfloordiv

youknowone:
This PR allows `import datetime`

If anyone interested in, the workflow is captured as an example patch for sprint.
https://www.slideshare.net/YunWonJeong/pycon-kr-2019-sprint-rustpython-by-example

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -146,6 +146,17 @@ fn inner_mod(int1: &PyInt, int2: &PyInt, vm: &VirtualMachine) -> PyResult {
     }
 }
 
+fn inner_divmod(int1: &PyInt, int2: &PyInt, vm: &VirtualMachine) -> PyResult {
+    if int2.value.is_zero() {
+        Err(vm.new_zero_division_error("integer division or modulo by zero".to_string()))
+    } else {
+        let (div, modulo) = int1.value.div_mod_floor(&int2.value);
+        Ok(vm
+            .ctx
+            .new_tuple(vec![vm.ctx.new_int(div), vm.ctx.new_int(modulo)]))
+    }
+}
+
 #[pyimpl]
 impl PyInt {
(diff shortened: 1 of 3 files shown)


vm/src/obj/objint.rs
```diff
@@ -732,6 +748,17 @@ fn div_ints(vm: &VirtualMachine, i1: &BigInt, i2: &BigInt) -> PyResult {
     }
 }
 
+#[inline]
+fn divmod(vm: &VirtualMachine, i1: &BigInt, i2: &BigInt) -> PyResult<(BigInt, BigInt)> {
+    if i2.is_zero() {
+        Err(vm.new_zero_division_error("integer division by zero".to_string()))
+    } else {
+        let modulo = (i1 % i2 + i2) % i2;
```
windelbouwman: Did you have a look at this: https://docs.rs/num-bigint/0.2.2/num_bigint/struct.BigInt.html#method.div_rem ?

vm/src/obj/objint.rs
```diff
@@ -270,12 +270,19 @@ impl PyInt {
...
+        } else {
+            Ok(vm.ctx.not_implemented())
+        }
+    }
+
+    #[pymethod(name = "__rfloordiv__")]
+    fn rfloordiv(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        if objtype::isinstance(&other, &vm.ctx.int_type()) {
+            let v1 = get_value(&other);
+            let (div, _) = divmod(vm, v1, &self.value)?;
```
windelbouwman: Did you check on `div_floor`? This might be faster than full divmod.

vm/src/obj/objint.rs
youknowone: div_rem was the original implementation which didn't match to python. I fixed them to div_floor and div_mod_floor, thanks.

PR #1242, opened 2019-08-12
Use redox-release branch for redox recipe

coolreader18:


.travis.yml
```diff
@@ -179,6 +179,19 @@ matrix:
         - bash redox/uncomment-cargo.sh
         - redoxer build --verbose
         - bash redox/comment-cargo.sh
+      deploy:
+        provider: script
+        script:
+          - openssl aes-256-cbc -K $encrypted_441570148e04_key -iv
+            $encrypted_441570148e04_iv -in redox/deploy_key.pem.enc -out
```
windelbouwman: I'm not sure about having a deployment key in the repo with access to the repository. What is the purpose of this deploy action? To merge the master branch into the redox-release branch? It might be good enough to once in a while check the master branch on redox compatibility.

The redox package could then refer to a specific commit that worked, or even better a tag to a released version.

windelbouwman: Btw, it might be nice to link to the redox package from the readme file, so redox users can install rustpython.

coolreader18: Yeah, I suppose I could merge master to it manually every so often; it really wouldn't be that much of a hassle. I'll remove the automatic deployment and just leave the `BRANCH=redox-release`.

coolreader18:
@windelbouwman what do you mean by a link to the package in the README? Redox doesn't have a package manager for the OS, all the packages are built and added to the image when you compile redox, so all you would have to do to enable rustpython while building redox is uncomment a line `filesystem.toml`.

windelbouwman:
Ah, okay, I'm not a redox user (yet), so I did not know there is no package manager.

PR #1243, opened 2019-08-12
Implement zlib module

mpajkowski:
Hello, I implemented following methods from ```zlib``` module:
* ```zlib.compress```
* ```zlib.decompress```
* ```zlib.crc32```
* ```zlib.adler32```


vm/src/stdlib/zlib.rs
```diff
@@ -0,0 +1,150 @@
...
+            $h_update(&mut hasher, &data);
+
+            let checksum: u32 = $h_compute(hasher);
+
+            Ok(vm.new_int(checksum))
+        }
+    };
+}
+
+checksum_fn!(
```
windelbouwman: The macro is used twice, so I would suggest to simply write out the two methods for readability. It's about the same amount on lines, but is somewhat easier to read. What do you think?

mpajkowski: Ok, at the first time i had done it with ```PyFuncArgs``` and ```arg_check!``` - it was much longer then. I'll rewrite it

mpajkowski: ok, done! :smiley:

tests/snippets/stdlib_zlib.py
```diff
@@ -0,0 +1,62 @@
...
+    assert zlib.compress(lorem, level) == text
+
+# default level
+assert zlib.compress(lorem) == zlib.compress(lorem, -1) == zlib.compress(lorem, 6)
+
+# decompression
+for text in compressed_lorem_list:
+    assert zlib.decompress(text) == lorem
+
+assert_raises(zlib.error, lambda: zlib.compress(b"123", -40))
```
windelbouwman: Just a minor though, I find this
```python
with assertRaises(zlib.error):
    zlib.compress(b"123", -40)
```
somewhat easier to read. What do you think?

PR #1245, opened 2019-08-13
Improve symbol table processing.

windelbouwman:
- Renamed SymbolScope to SymbolTable
- Add proper detection of a local variable

### DIFF
--- a/crawl_sourcecode.py
+++ b/crawl_sourcecode.py
@@ -12,6 +12,7 @@ $ diff cpython.txt rustpython.txt
 import ast
 import sys
 import symtable
+import dis
 
 filename = sys.argv[1]
 print('Crawling file:', filename)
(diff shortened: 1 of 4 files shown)


crawl_sourcecode.py
```diff
@@ -67,3 +68,9 @@ def print_table(table, indent=0):
 
 table = symtable.symtable(source, 'a', 'exec')
 print_table(table)
+
+print()
+print('======== dis.dis ========')
+print()
+co = compile(source, filename, 'exec')
```
coolreader18: `dis.dis` returns `None`, it prints the disassembly.

windelbouwman: Allright, will change it later on. This is sort of a scratch script to check for scope detection working properly.

windelbouwman:
@cthulahoops sorry for the too quick merge. I put you on review, since this might enable proper local variables.

PR #1246, opened 2019-08-14
Fix weirdness with importing submodules

coolreader18:


vm/src/frame.rs
```diff
@@ -740,25 +737,23 @@ impl Frame {
...
+    fn import_star(&self, vm: &VirtualMachine) -> FrameResult {
+        let module = self.pop_value();
 
         // Grab all the names from the module and put them in the context
         if let Some(dict) = &module.dict {
             for (k, v) in dict {
-                self.scope.store_name(&vm, &objstr::get_value(&k), v);
+                let k = vm.to_str(&k)?;
+                let k = k.as_str();
+                if !k.starts_with('_') {
```
windelbouwman: Did you check cpython sourcecode how this is implemented? I think we should respect the `__all__` variable at this point right?

vm/src/obj/objmodule.rs
```diff
@@ -18,14 +18,37 @@ impl PyValue for PyModule {
...
+    }
+
+    fn getattribute(self, name: PyStringRef, vm: &VirtualMachine) -> PyResult {
+        match vm.generic_getattribute(self.as_object().clone(), name.clone()) {
+            Ok(Some(val)) => Ok(val),
+            Ok(None) => Err(vm.new_attribute_error(format!(
+                "module '{}' has no attribute '{}'",
+                self.name, name,
+            ))),
+            Err(err) => Err(err),
```
windelbouwman: You can use `?` when you have `Err(err) => Err(err)`.

vm/src/vm.rs
```diff
@@ -407,10 +407,23 @@ impl VirtualMachine {
...
-            .or_else(|_| {
+        // if the import inputs seem weird, e.g a package import or something, rather than just
+        // a straight `import ident`
+        let weird = module.contains('.')
+            || level != 0
+            || objbool::boolval(self, from_list.clone()).unwrap_or(true);
+
+        let module = self.new_str(module.to_owned());
+
+        let sys_module = if weird {
```
windelbouwman: This variable name should be renamed. It is not the `sys` module, but it is optionally a cached module from `sys.modules`.

vm/src/obj/objobject.rs
```diff
@@ -218,39 +218,11 @@ fn object_dict_setter(
...
-    vm: &VirtualMachine,
-) -> PyResult<Option<PyObjectRef>> {
-    if let Some(ref dict) = obj.dict {
-        dict.get_item_option(attr_name, vm)
-    } else {
-        Ok(None)
+    match vm.generic_getattribute(obj.clone(), name.clone()) {
+        Ok(Some(val)) => Ok(val),
+        Ok(None) => Err(vm.new_attribute_error(format!("{} has no attribute '{}'", obj, name))),
+        Err(err) => Err(err),
```
windelbouwman: You could use the `?` operator here.

compiler/src/compile.rs
```diff
@@ -549,6 +548,30 @@ impl<O: OutputStream> Compiler<O> {
         Ok(())
     }
 
+    fn compile_import(
+        &mut self,
+        name: Option<&str>,
+        symbols: Vec<String>,
+        level: usize,
```
windelbouwman: Suggest to move the position of level before the name parameter.

vm/src/frame.rs
windelbouwman: I checked with the python opcodes, and this is indeed correct behavior. Well, that's a surprise for me :).

PR #1251, opened 2019-08-14
Extend AST python module. Add idea for scope detection.

windelbouwman:


compiler/src/symboltable.rs
```diff
@@ -204,8 +204,25 @@ impl SymbolTableAnalyzer {
...
+
+                if symbol.is_assigned || symbol.is_parameter {
                     symbol.scope = SymbolScope::Local;
+                } else {
+                    // TODO: comment this out and make it work properly:
+                    /*
+                    let found_in_outer_scope = self
+                        .tables
+                        .iter()
+                        .any(|t| t.symbols.contains_key(&symbol.name));
```
coolreader18: Why doesn't this work now?

windelbouwman: It works everywhere, except when using `exec` to execute code and providing a global dictionary. I thought I had all cases covered, but this was missed.

PR #1252, opened 2019-08-15
Make func_ref an &PyObjectRef

mpajkowski:
Fixes this [TODO](https://github.com/RustPython/RustPython/blob/master/vm/src/vm.rs#L526)


coolreader18:
Nice, looks like there's still some changes to be made to `rustpython_wasm`.

mpajkowski:
Thanks for review! Now ```rustpython-wasm``` builds flawlessly.

PR #1255, opened 2019-08-15
Add `json.dump` and `json.load`

zsaladin:
Hello.
I am new for rust python.

I have added `json_dump` method. But it works only for StringIO.
Could you give me some advice? (Other IO,, what else)

### DIFF
--- a/vm/src/stdlib/json.rs
+++ b/vm/src/stdlib/json.rs
@@ -39,6 +39,11 @@ pub fn json_loads(string: PyStringRef, vm: &VirtualMachine) -> PyResult {
     })
 }
 
+pub fn json_load(fp: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+    let result = vm.call_method(&fp, "read", vec![])?;
+    json_loads(result.downcast()?, vm)
+}
+
 pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
     let ctx = &vm.ctx;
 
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/json.rs
```diff
@@ -10,6 +10,11 @@ pub fn json_dumps(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<String> {
     serde_json::to_string(&serializer).map_err(|err| vm.new_type_error(err.to_string()))
 }
 
+pub fn json_dump(obj: PyObjectRef, fs: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+    let result = json_dumps(obj, vm).map_err(|err| vm.new_type_error(err.to_string()));
```
windelbouwman: Instead of the `unwrap`, you might consider the `?` operator like this:

```rust
let result = json_dumps(obj, vm)?;
vm.call_method(&fs, "write", vec![vm.new_str(result)])
```

windelbouwman:
Hi @zsaladin , thank you for your contribution!

Some tips:
- Please add a test snippet in `tests/snippets/json_snippet.py` which demonstrates the behavior. For example by using `io.StringIO` and `json.dump`.
- Consider adding the `load` function as well.

youknowone:
@windelbouwman  This function take `StringIO` but doesn't take `BytesIO`. Any idea how to support it?

windelbouwman:
@youknowone StringIO is basically a file opened in text mode. If you have a BytesIO file, you have a file opened in binary mode, so you would need to encode the text first as bytes, and then write it to file. Should the `json.dump` take a file opened in binary mode?

zsaladin:
json.load and json.loads need to support bytes-like objects.
https://docs.python.org/3/library/json.html#json.load

According to cpython [implementation](https://github.com/python/cpython/blob/master/Lib/json/__init__.py#L293), implementationn `json_loads`for bytes-like object should be preceded. Does this need to be done in this PR?

windelbouwman:
@zsaladin what do you mean by "bytes-like object should be preceded"?

zsaladin:
Oh I mean `json_loads` for bytes-like object(bytes, bytearray) should be implemented first. I will make new PR for it.

PR #1257, opened 2019-08-15
Impl socket htonl

SeungUkLee:
For #1176

Add socket method htonl.

### DIFF
--- a/vm/src/stdlib/socket.rs
+++ b/vm/src/stdlib/socket.rs
@@ -21,8 +21,8 @@ use crate::vm::VirtualMachine;
 use crate::obj::objtype::PyClassRef;
 #[cfg(unix)]
 use crate::stdlib::os::convert_nix_error;
-use num_traits::ToPrimitive;
 use num_bigint::Sign;
+use num_traits::ToPrimitive;
 
 #[derive(Debug, Copy, Clone)]
 enum AddressFamily {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/socket.rs
```diff
@@ -409,6 +410,17 @@ fn socket_inet_ntoa(packed_ip: PyBytesRef, vm: &VirtualMachine) -> PyResult {
...
 
+fn socket_htonl(host: PyIntRef, vm: &VirtualMachine) -> PyResult {
+    if host.as_bigint().sign() == Sign::Minus {
+        return Err(
+            vm.new_overflow_error("can't convert negative value to unsigned int".to_string())
+        );
+    }
+
+    let host = host.as_bigint().to_u32().unwrap();
+    Ok(vm.new_int(host.to_be()))
```
windelbouwman: Is there a rust library with the `htonl` and `ntohl` functions?

windelbouwman:
Hi @SeungUkLee this change looks good! I had one question about using a library for the htonl function.

vm/src/stdlib/socket.rs
SeungUkLee: I found `htonl` and `ntohl` in [socket](https://docs.rs/socket/0.0.7/socket/) crate. but it is not very popular.

windelbouwman: Okay, maybe stick to this version, until we find a good socket crate. @palaviv do you have any ideas?

coolreader18: Even from [gnu libc](https://github.com/lattera/glibc/blob/895ef79e04a953cac1493863bcae29ad85657ee1/inet/htonl.c#L24-L34), it essentially is this simple of an implementation. Rust's [`u32::to_be`](https://doc.rust-lang.org/src/core/num/mod.rs.html#2616-2625) looks basically the same to that libc `htonl`.

PR #1260, opened 2019-08-15
implement os.readlink()

HyeonuPark:
this fixes broken pytest on macos, which failed on `os.path.realpath(tmpdir)`


windelbouwman:
@HyeonuPark nicely done! Thanks for contributing!

PR #1261, opened 2019-08-15
Implemented stdlib function 'os.cpu_count()'

kluid:
Hi, I just implemented stdlib 'os.cpu_count()'.


windelbouwman:
@kluid thanks for contributing! This looks good :)

coolreader18:
Looks like it is just 1, judging by this code: https://github.com/seanmonstar/num_cpus/blob/master/src/lib.rs#L434-L452

PR #1262, opened 2019-08-15
Add str.__rmod__() method. #190

ChJR:
CPython returns "NotImplemented" string when arguments supplied.

### DIFF
--- a/tests/snippets/strings.py
+++ b/tests/snippets/strings.py
@@ -294,11 +294,5 @@ assert next(str_iter_reversed) == "1"
 assert next(str_iter_reversed, None) == None
 assert_raises(StopIteration, lambda: next(str_iter_reversed))
 
-assert str.__rmod__('%i', 30) == "NotImplemented"
-
-typeError = False
-try:
-    str.__rmod__(30, '%i')
-except TypeError:
-    typeError = True
-assert typeError
+assert str.__rmod__('%i', 30) == NotImplemented
+assert_raises(TypeError, lambda: str.__rmod__(30, '%i'))
--- a/tests/snippets/strings.py
+++ b/tests/snippets/strings.py
@@ -293,3 +293,12 @@ assert next(str_iter_reversed) == "2"
 assert next(str_iter_reversed) == "1"
 assert next(str_iter_reversed, None) == None
 assert_raises(StopIteration, lambda: next(str_iter_reversed))
+
+assert str.__rmod__('%i', 30) == "NotImplemented"
+
+typeError = False
+try:
+    str.__rmod__(30, '%i')
+except TypeError:
+    typeError = True
+assert typeError
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -526,7 +526,7 @@ impl PyString {
 
         match do_cformat(vm, _format_string, values.clone()) {
             Ok(_format_string) => Ok(vm.ctx.not_implemented()),
-            Err(err) => Err(err)
+            Err(err) => Err(err),
         }
     }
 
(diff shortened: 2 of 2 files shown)


tests/snippets/strings.py
```diff
@@ -293,3 +293,12 @@ def try_mutate_str():
...
 assert_raises(StopIteration, lambda: next(str_iter_reversed))
+
+assert str.__rmod__('%i', 30) == "NotImplemented"
+
+typeError = False
+try:
+    str.__rmod__(30, '%i')
+except TypeError:
+    typeError = True
+assert typeError
```
coolreader18: You can use `assertRaises` from `testutils` for this:

```python
from testutils import assertRaises

...

with assertRaises(TypeError):
    ...
```

vm/src/obj/objstr.rs
```diff
@@ -518,6 +518,18 @@ impl PyString {
...
+    #[pymethod(name = "__rmod__")]
+    fn rmod(&self, values: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        let format_string_text = &self.value;
+        let _format_string = CFormatString::from_str(format_string_text)
+            .map_err(|err| vm.new_value_error(format!("{}", err)))?;
+
+        match do_cformat(vm, _format_string, values.clone()) {
+            Ok(_format_string) => Ok(vm.new_str("NotImplemented".to_string())),
+            Err(err) => Err(err)
+        }
```
coolreader18: I think you misunderstood what you were seeing in the python shell -- in CPython, `str.__rmod__(str, int)` doesn't return `"NotImplemented"`, it returns the actual value `NotImplemented`, which you can access from Rust as `vm.ctx.not_implemented()`.

ChJR: I misunderstood that. I will fix and re-open this pull request.

tests/snippets/strings.py
ChJR: Thank you for sharing. I will use your proposal for this PR.

ChJR:
I improved PR code quality.
Thanks to @coolreader18

vm/src/obj/objstr.rs
```diff
@@ -518,6 +518,18 @@ impl PyString {
         do_cformat(vm, format_string, values.clone())
     }
 
+    #[pymethod(name = "__rmod__")]
+    fn rmod(&self, values: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        let format_string_text = &self.value;
+        let _format_string = CFormatString::from_str(format_string_text)
+            .map_err(|err| vm.new_value_error(format!("{}", err)))?;
```
```suggestion
            .map_err(|err| vm.new_value_error(err.to_string()))?;
```

ChJR: That code found on str.__mod__ too. I suggest that code should be moved to another PR.

coolreader18: Sounds good, thanks for contributing!

vm/src/obj/objstr.rs
```diff
@@ -518,6 +518,18 @@ impl PyString {
...
+
+        match do_cformat(vm, _format_string, values.clone()) {
+            Ok(_format_string) => Ok(vm.ctx.not_implemented()),
```
windelbouwman: Is the call to `do_cformat` required here? Or can we just always return `NotImplemented`?

ChJR: I have just checked your second suggestion, and I found it works.
I used do_cformat just for argument checking, but that method is not required.
I will add a new commit for this PR.

youknowone:
@ChJR there is a tiny style glitch. There is a clippy result in travis.

PR #1265, opened 2019-08-15
Add __reduce__ method to Ellipsis

### LINKED ISSUE #1254: [PyCon Sprint KR Good first bug] Ellipsis.__reduce__() not working
## CPython
```python
>>> Ellipsis.__reduce__()
'Ellipsis'
```

## RustPython
``` python
>>> Ellipsis.__reduce__()
AttributeError: 'EllipsisType' object has no attribute '__reduce__'
```

pbzweihander:
Resolves #1254 

Add `__reduce__` method to `Ellipsis` class.

### DIFF
--- a/vm/src/obj/objellipsis.rs
+++ b/vm/src/obj/objellipsis.rs
@@ -1,5 +1,5 @@
 use crate::function::PyFuncArgs;
-use crate::pyobject::{PyContext, PyResult};
+use crate::pyobject::{PyContext, PyEllipsisRef, PyResult};
 use crate::vm::VirtualMachine;
 
 pub fn init(context: &PyContext) {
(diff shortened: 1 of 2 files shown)


corona10:
LGTM for me

vm/src/obj/objellipsis.rs
```diff
@@ -18,3 +19,8 @@ fn ellipsis_repr(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     arg_check!(vm, args, required = [(_cls, None)]);
     Ok(vm.new_str("Ellipsis".to_string()))
 }
+
+fn ellipsis_reduce(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
+    arg_check!(vm, args, required = [(_cls, None)]);
+    Ok(vm.new_str("Ellipsis".to_string()))
+}
```
coolreader18: I know this is a really simple function, but it's good to know how to write RustPython functions that can take advantage of the `Fn()` type shenanigans we have set up to make the signatures more readable.

First off, it looks like `Ellipsis.__reduce__` does require an `Ellipsis` as its self argument:
```
>>> type(Ellipsis).__reduce__(1)
TypeError: descriptor '__reduce__' requires a 'ellipsis' object but received a 'int'
```

So we want to write a function that takes one parameter and returns a string. We can do that by just writing a rust function that expresses that signature, because we know all the types:

```rust
fn ellipsis_reduce(
    _self: PyEllipsisRef, // from crate::pyobject
    _vm: &VirtualMachine, // necessary to mark that this is a python function
) -> String {
    "Ellipsis".to_string()
}

All the argument checking is done by the `FromArgs` trait in `function.rs`, and the result conversion is done by the `IntoPyObject` trait from `pyobject.rs`.

pbzweihander:
I have updated the signature of `__repr__` and `__reduce__` method. And I also updated the signature of `__new__` method and add subtype-checking.

PR #1267, opened 2019-08-15
Add _w_long, _u_long to bootstrap which is requried by importlib

youknowone:


coolreader18:
Interesting, I wonder where we got our version from that it doesn't have it. Feel free to update it with the full version of `_bootstrap_external` from CPython.

youknowone:
It seems the files came from python 3.8 dev

PR #1268, opened 2019-08-15
object.__subclasshook__

youknowone:
https://docs.python.org/3/library/abc.html#abc.ABCMeta.__subclasshook__


coolreader18:
It looks like CPython just returns NotImplemented in this case: https://github.com/python/cpython/blob/master/Objects/typeobject.c#L4643-L4647

youknowone:
Thanks! fixed and added tests

Issue #1224, opened 2019-08-09
%f formatting is missing


coolreader18:
## Feature

`'%f'` printf-style formatting formats floats, including both sides of the decimal point and an optional precision argument. It should just be adding a format_float function in `cformat.rs`.

## Python Documentation

https://docs.python.org/3/library/stdtypes.html#printf-style-string-formatting

PR #1273, opened 2019-08-15
Add floating point formatting

jmg7173:
#1224
Just supports those things:
* `%f`
* `% f`
* `%+f`
* Integer input is available

Not supported:
* `%#.#f` (TODO)


vm/src/cformat.rs
```diff
@@ -209,6 +209,41 @@ impl CFormatSpec {
             self.fill_string(format!("{}{}", prefix, magnitude_string), ' ', None)
         }
     }
+
+    pub fn format_float(&self, num: f64) -> String {
+        let magnitude = num.abs();
+
+        let sign_string = match num.is_sign_positive() {
```
windelbouwman: Instead of `match` you can use `if` in the case of true and false values like this:
```rust
let sign_string = if num.is_sign_positive() {
    "+" // ... or the other options.
} else {
    "-"
};
```

windelbouwman:
@jmg7173  There is some build failure, could you take a look at that? For the rest, this looks pretty good!

jmg7173:
Thank you! I'll fix it.

vm/src/cformat.rs
jmg7173: Thank you for your comment. I fix it!

coolreader18:
Looks like clippy's failing on the `_ => ...` match arm in `objstr.rs` because you've made the match exhaustive now, so there's no way it'd ever get to the `_` pattern.

PR #1274, opened 2019-08-15
Add int.__(x)xshift__ methods.

ChJR:


vm/src/obj/objint.rs
```diff
@@ -311,18 +341,18 @@ impl PyInt {
...
-                Err(vm.new_overflow_error("the number is too large to convert to int".to_string()))
-            }
-            _ => panic!("Failed converting {} to rust usize", get_value(&other)),
+    #[pymethod(name = "__rlshift__")]
+    fn rlshift(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        if !objtype::isinstance(&other, &vm.ctx.int_type()) {
+            return Ok(vm.ctx.not_implemented());
         }
+
+        let other = other.payload::<PyInt>().unwrap();
```
windelbouwman: It is better to not directly access the payload attribute. Use `get_value` instead.

vm/src/obj/objint.rs
```diff
@@ -165,6 +165,36 @@ fn inner_divmod(int1: &PyInt, int2: &PyInt, vm: &VirtualMachine) -> PyResult {
     }
 }
 
+fn inner_lshift(int1: &PyInt, int2: &PyInt, vm: &VirtualMachine) -> PyResult {
+    if let Some(n_bits) = int2.value.to_usize() {
+        return Ok(vm.ctx.new_int(&int1.value << n_bits));
+    }
+
+    // i2 failed `to_usize()` conversion
+    match &int2.value {
```
windelbouwman: I suggest to put this code into a helper function `get_shift_amount` like this:
```rust
fn get_shift_amount(amount: &PyInt, vm: &VirtualMachine) -> PyResult<usize> {
    if let Some(n_bits) = amount.value.to_usize() {
        Ok(n_bits)
    } else {
        match &amount.value {
              v if *v < BigInt::zero() => Err(vm.new_value_error("negative shift count".to_string())),
              v if *v > BigInt::from(usize::max_value()) => {
                    Err(vm.new_overflow_error("the number is too large to convert to int".to_string()))
              }
             _ => panic!("Failed converting {} to rust usize", amount.value),
        }
    }
}
```
And then use this function in `inner_lshift` (and off course in `inner_rshift`):
```rust
fn inner_lshift(int1: &PyInt, int2: &PyInt, vm: &VirtualMachine) -> PyResult {
    let n_bits = get_shift_amount(int2, vm)?;
    Ok(vm.ctx.new_int(&int1.value << n_bits))
}
```

ChJR:
This approach looks better than I used. There will be another PR soon.
Thank you @windelbouwman

vm/src/obj/objint.rs
ChJR: I found some similar problems in other int methods. I suggest this change to another PR.

PR #1276, opened 2019-08-15
Add functools.reduce

mpajkowski:


palaviv:
@mpajkowski why have you chosen to implement this in Rust instead of using functools.py from CPython stdlib?

windelbouwman:
@palaviv CPython also implements `reduce` in C code, I think this is why.

windelbouwman:
@mpajkowski nice job!

vm/src/stdlib/functools.rs
```diff
@@ -0,0 +1,46 @@
...
+    vm: &VirtualMachine,
+) -> PyResult {
+    let iterator = objiter::get_iter(vm, &sequence)?;
+
+    let start_value = if let OptionalArg::Present(val) = start_value {
+        val
+    } else {
+        objiter::call_next(vm, &iterator).map_err(|err| {
+            if objtype::isinstance(&err, &vm.ctx.exceptions.stop_iteration) {
+                let exc_type = vm.ctx.exceptions.type_error.clone();
```
youknowone: `vm.new_type_error("reduce() of empty sequence with no initial value".to_string())` also works

Issue #1263, opened 2019-08-15
[good first issue] implement os.access


coolreader18:
## Feature

`os.access` lets you check to see whether a user has access to a file or not. It is used by a few libraries in the standard library, namely `webbrowser`, which is depended on by `antigravity`, which is just a fun one IMO.

## Python Documentation

https://docs.python.org/3/library/os.html#os.access

## Implementation

You can look at using [`std::fs::Permissions`](https://doc.rust-lang.org/std/fs/struct.Permissions.html) to implement this.

Nimpruda:
Hi,
I'd like to try and implement this but i'm not sure how to proceed and where to put it. I guess it's in vm/src/stdlib/os.rs ?

Nimpruda:
https://github.com/Nimpruda/RustPython/commit/6f78bca8ee41e33728ceaa180bb455050931b343 
Would something like that do the trick for the R_OK and F_OK? I'm not sure how I am supposed to use vm tho.

kluid:
@Nimpruda If you feel free, Can I try to implement os.access function roughly?

PR #1277, opened 2019-08-15
[WIP]Implement os access

Nimpruda:
Tries to fix https://github.com/RustPython/RustPython/issues/1263
It's my first pull request ever on github. I'd love to have feedback.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -360,11 +360,10 @@ fn os_access(path: PyStringRef, mode: u8, vm: &VirtualMachine) -> PyResult<bool>
 
             match get_right_permission(mode, Uid::from_raw(user_id), Gid::from_raw(group_id)) {
                 Ok(perm) => {
-                    if flags.contains(AccessFlags::R_OK) & perm.is_readable {
-                        Ok(true)
-                    } else if flags.contains(AccessFlags::W_OK) & perm.is_writable {
-                        Ok(true)
-                    } else if flags.contains(AccessFlags::X_OK) & perm.is_executable {
+                    if (flags.contains(AccessFlags::R_OK) & perm.is_readable)
+                        || (flags.contains(AccessFlags::W_OK) & perm.is_writable)
+                        || (flags.contains(AccessFlags::X_OK) & perm.is_executable)
+                    {
                         Ok(true)
                     } else {
                         Ok(false)
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -233,6 +233,7 @@ bitflags! {
 
 }
 
+#[cfg(unix)]
 fn os_access(path: PyStringRef, mode: u8, vm: &VirtualMachine) -> PyResult<bool> {
     let path = path.as_str();
     let file_metadata = fs::metadata(path);
@@ -1046,8 +1047,7 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         "W_OK" => ctx.new_int(AccessFlags::W_OK.bits()),
         "X_OK" => ctx.new_int(AccessFlags::X_OK.bits()),
         "getpid" => ctx.new_rustfunc(os_getpid),
-        "cpu_count" => ctx.new_rustfunc(os_cpu_count),
-        "access" => ctx.new_rustfunc(os_access)
+        "cpu_count" => ctx.new_rustfunc(os_cpu_count)
     });
 
     for support in support_funcs {
@@ -1096,6 +1096,7 @@ fn extend_module_platform_specific(vm: &VirtualMachine, module: PyObjectRef) -> 
         "setgid" => ctx.new_rustfunc(os_setgid),
         "setpgid" => ctx.new_rustfunc(os_setpgid),
         "setuid" => ctx.new_rustfunc(os_setuid),
+        "access" => ctx.new_rustfunc(os_access)
     });
 
     #[cfg(not(target_os = "redox"))]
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/os.rs
```diff
@@ -221,6 +221,31 @@ fn convert_nix_errno(vm: &VirtualMachine, errno: Errno) -> PyClassRef {
     }
 }
 
+fn os_access(path: PyStringRef, mode: PyIntRef, vm: &VirtualMachine) -> PyResult {
```
coolreader18: You can return a `PyResult<bool>` to avoid having to call `vm.new_bool` yourself.

vm/src/stdlib/os.rs
```diff
@@ -221,6 +221,31 @@ fn convert_nix_errno(vm: &VirtualMachine, errno: Errno) -> PyClassRef {
...
+    let mode = mode.as_bigint().to_u8().unwrap();
```
coolreader18: You can accept `mode: u8` as a parameter to have it check that the BigInt fits in a u8 for you.

vm/src/stdlib/os.rs
```diff
@@ -221,6 +221,31 @@ fn convert_nix_errno(vm: &VirtualMachine, errno: Errno) -> PyClassRef {
...
+                    Err(convert_io_error(vm, err))
+                }
+            }
+        },
+        4 => match file_metadata {
+            Ok(metadata) => Ok(vm.new_bool(metadata.permissions().readonly())),
+            Err(err) => Err(convert_io_error(vm, err)),
+        },
+        2 => unimplemented!(),
+        1 => unimplemented!(),
```
coolreader18: I'd recommend extracting out 1, 2, 3, and 4 into `const`s and using them also for `F_OK`, `W_OK`, etc in `make_module`.

vm/src/stdlib/os.rs
Nimpruda: I did that first but I wasn sure it would send it to the context.

vm/src/stdlib/os.rs
Nimpruda: I thought only Python types were accepted by functions.

coolreader18: [Here's](https://github.com/RustPython/RustPython/pull/1265#discussion_r314192418) a little explanation of how the parameter and return types of native functions work for RustPython. Basically, there's a trait `FromArgs` that lets a type be used as a parameter, and a trait `IntoPyObject` that lets a type be used as a return type.

vm/src/stdlib/os.rs
coolreader18: It does, `bool` implements the `IntoPyObject` trait that I mentioned in my other comment, and `IntoPyObject` is implemented for `PyResult<T> where T: IntoPyObject`.

coolreader18:
Looks good, do you want to implement `W_OK` and `X_OK` in this pull request or open a separate one?

vm/src/stdlib/os.rs
```diff
@@ -221,6 +221,37 @@ fn convert_nix_errno(vm: &VirtualMachine, errno: Errno) -> PyClassRef {
     }
 }
 
+// Flags for os_access
+const F_OK: u8 = 0;
+const R_OK: u8 = 4;
+const W_OK: u8 = 2;
+const X_OK: u8 = 1;
```
coolreader18: Actually, sorry; this should be a bitflags struct. You can look at `FileCreationFlags` for an example of this, it would probably look something like this:

```rust
bitflags! {
     pub struct AccessFlags: u8 {
        const F_OK = 0;
        const R_OK = 4;
        const W_OK = 2;
        const X_OK = 1;
    }
}
```

Then in `os_access` you can check to see if the user's checking for a certain flag like this:
```rust
let flags = AccessFlags::from_bits(mode)
    .ok_or_else(|| vm.new_value_error("Unsupported flag".to_string()))?;
if flags.contains(AccessFlags::F_OK) {
    ...
}
```
This is all because you can use `access` like this:
```python
os.access(path, R_OK & W_OK)
```

in order to check multiple permissions on the file.

vm/src/stdlib/os.rs
```diff
@@ -221,6 +221,37 @@ fn convert_nix_errno(vm: &VirtualMachine, errno: Errno) -> PyClassRef {
...
+                    Ok(false)
+                } else {
+                    Err(convert_io_error(vm, err))
+                }
+            }
+        },
+        R_OK => match file_metadata {
+            Ok(metadata) => Ok(metadata.permissions().readonly()),
+            Err(err) => Err(convert_io_error(vm, err)),
+        },
```
coolreader18: `readonly()` checks to see if it has read but *not* write permissions, which isn't what `R_OK` signals. You should probably try to parse the mode of the file using the [`PermissionsExt`](https://doc.rust-lang.org/std/os/unix/fs/trait.PermissionsExt.html) unix-only trait and [conditional compilation](https://doc.rust-lang.org/rust-by-example/attribute/cfg.html)

vm/src/stdlib/os.rs
```diff
@@ -1,13 +1,13 @@
+use num_cpus;
 use std::cell::RefCell;
 use std::ffi::CStr;
 use std::fs::File;
 use std::fs::OpenOptions;
 use std::io::{self, Error, ErrorKind, Read, Write};
+use std::os::unix::fs::MetadataExt;
```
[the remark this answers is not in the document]
Nimpruda: Yes will do when I come back. I have to commit the rest of my work.

Nimpruda:
I couldn't make F_OK working without bypassing all the other ones as it is 0.

coolreader18:
That's fine, i can look into that later.

PR #1278, opened 2019-08-15
Upgrade dependencies

coolreader18:


coolreader18:
Ah, @windelbouwman you're up :smiley:. It's funny to be working with people in 3 vastly different time zones :smile:

windelbouwman:
It's still early here :)

coolreader18:
And it's almost 1 am where I am, and the middle of the afternoon at PyCon KR. Time is crazy, yo :)

PR #1284, opened 2019-08-16
Update factorial module of math

Writtic:
- Implement factorial function with test case
- Deal with Negative input as a ValueError

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -237,13 +237,12 @@ fn math_gcd(a: PyIntRef, b: PyIntRef, vm: &VirtualMachine) -> PyResult {
 fn math_factorial(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     arg_check!(vm, args, required = [(value, None)]);
     let value = objint::get_value(value);
-    if value < &BigInt::zero() {
-        return Err(vm.new_value_error(format!("factorial() not defined for negative values")))
-    } else if value <= &BigInt::one() {
-        return Ok(vm.ctx.new_int(BigInt::from(1u64)))
+    if *value < BigInt::zero() {
+        return Err(vm.new_value_error("factorial() not defined for negative values".to_string()));
+    } else if *value <= BigInt::one() {
+        return Ok(vm.ctx.new_int(BigInt::from(1u64)));
     }
-    let ret: BigInt =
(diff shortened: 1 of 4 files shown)


vm/src/stdlib/math.rs
```diff
@@ -6,13 +6,15 @@
...
+use num_bigint::BigInt;
+use num_traits::{One, Zero};
+
 use crate::function::PyFuncArgs;
 use crate::obj::objint::PyIntRef;
-use crate::obj::{objfloat, objtype};
+use crate::obj::{objfloat, objint, objtype};
 use crate::pyobject::{PyObjectRef, PyResult, TypeProtocol};
-use crate::vm::VirtualMachine;
+use crate::vm::VirtualMachine;// Helper macro:
```
youknowone: Is this an intended change?

vm/src/stdlib/math.rs
```diff
@@ -231,6 +233,20 @@ fn math_gcd(a: PyIntRef, b: PyIntRef, vm: &VirtualMachine) -> PyResult {
     Ok(vm.new_int(a.as_bigint().gcd(b.as_bigint())))
 }
 
+fn math_factorial(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
+    arg_check!(vm, args, required = [(value, None)]);
+    let value = objint::get_value(value);
+    match value {
```
windelbouwman: Maybe instead of a `match` statement, you could choose the `if` statement like this:
```rust
if value <BigInt::zero() {
    ...
} else if value <= BigInt::one() {
    ...
} else {
    ...
}

Writtic: I used `if` statement before but as far as I know rust could compile with zero-cost abstration so, I think, there is no performance issue and I just feel `match` statement looks much more rustic(?).
That's why I use `match` statement.
Is there more specific reason for using `if` statement..?

vm/src/stdlib/math.rs
Writtic: No, I'm going to fix it right away.

vm/src/stdlib/math.rs
Writtic: Oh, but `if` statement is much simpler than `match` in this case. I'll change it right away. thx.

windelbouwman: `match` is indeed very powerful, but in some cases, a simple `if` does the job.

PR #1288, opened 2019-08-16
Update ldexp module of math

Writtic:
- Implement ldexp function with test case

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -8,6 +8,7 @@ use statrs::function::gamma::{gamma, ln_gamma};
 
 use num_bigint::BigInt;
 use num_traits::{One, Zero};
+use num_traits::cast::ToPrimitive;
 
 use crate::function::PyFuncArgs;
 use crate::obj::objint::PyIntRef;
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/math.rs
```diff
@@ -229,6 +230,13 @@ fn math_frexp(value: PyObjectRef, vm: &VirtualMachine) -> PyResult {
     )
 }
 
+fn math_ldexp(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
+    arg_check!(vm, args, required = [(value, None), (i, None)]);
```
windelbouwman: Instead of using `arg_check` you might want to use `PyFloatRef` and `PyIntRef` here, like this:
```rust
fn math_ldexp(value: PyFloatRef, i: PyIntRef, vm: &VirtualMachine) -> PyResult {
    ...
}
```

This will check the types for you. This code will probably crash with a panic when you feed a `str` into the second argument. Did you try this out? When using the `PyIntRef` this will raise a `TypeError` when passing a `str`.

Writtic: I definitely missed it!
I appreciate your advice!
I apply right away.

PR #1290, opened 2019-08-16
Fix panics with int()

### LINKED ISSUE #1228: int('0b100', base=0) panic!
## Feature

`objint.rs`:
```
/// If x is not a number or if base is given, then x must be a string,
/// bytes, or bytearray instance representing an integer literal in the
/// given base.  The literal can be preceded by '+' or '-' and be surrounded
/// by whitespace.  The base defaults to 10.  Valid bases are 0 and 2-36.
/// Base 0 means to interpret the base from the string as an integer literal.

>>> int('0b100', base=0)
```

When I run it in ubuntu 18.04, it will panic.
Not handling error correctly when out of ranges.

mpajkowski:
Resolves #1228 issue

### DIFF
--- a/tests/snippets/ints.py
+++ b/tests/snippets/ints.py
@@ -117,12 +117,22 @@ with assertRaises(ValueError):
 with assertRaises(ValueError):
     int('0bFF', base=10)
 with assertRaises(ValueError):
+    int('0bFF', base=10)
+with assertRaises(ValueError):
     int(b"F\xc3\xb8\xc3\xb6\xbbB\xc3\xa5r")
 with assertRaises(ValueError):
     int(b"F\xc3\xb8\xc3\xb6\xbbB\xc3\xa5r")
 
 # underscore
 assert int('0xFF_FF_FF', base=16) == 16_777_215
+with assertRaises(ValueError):
+    int("_123_")
+with assertRaises(ValueError):
+    int("123_")
+with assertRaises(ValueError):
+    int("_123")
+with assertRaises(ValueError):
+    int("1__23")
 
 # signed
 assert int('-123') == -123
@@ -136,6 +146,7 @@ assert int('10', base=0) == 10
 
 # type byte, signed, implied base
 assert int(b'     -0XFF ', base=0) == -255
+
 
 assert int.from_bytes(b'\x00\x10', 'big') == 16
 assert int.from_bytes(b'\x00\x10', 'little') == 4096
--- a/tests/snippets/ints.py
+++ b/tests/snippets/ints.py
@@ -98,6 +98,8 @@ assert -10 // -4 == 2
 
 assert int() == 0
 assert int(1) == 1
+assert int("101", 2) == 5
+assert int("101", base=2) == 5
 
 # implied base
 assert int('1', base=0) == 1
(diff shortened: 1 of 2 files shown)


tests/snippets/ints.py
```diff
@@ -97,14 +97,40 @@
 assert -10 // -4 == 2
 
 assert int() == 0
-assert int("101", 2) == 5
-assert int("101", base=2) == 5
```
windelbouwman: Its good to keep this check with an explicit base given.

vm/src/obj/objint.rs
```diff
@@ -736,21 +739,22 @@ fn int_new(cls: PyClassRef, options: IntOptions, vm: &VirtualMachine) -> PyResul
...
-                )))
+        string @ PyString => {
+            let s = string.value.as_str().trim();
+            str_to_int(vm, s, base)
+        },
+        bytes @ PyBytes => {
+            let bytes = bytes.get_value();
+            let s = std::str::from_utf8(bytes)
+                .map(|s| s.trim())
+                .map_err(|_| invalid_literal(vm, &bytes.iter().map(|&c| c as char).collect::<String>(), base))?;
```
windelbouwman: You might want to raise a TypeError here and use the rust error message like this:
```rust
.map_err(|e| vm.new_type_error(format!("utf8 decode error: {}", e)))?;
```

windelbouwman:
@mpajkowski nice job!

vm/src/obj/objint.rs
mpajkowski: I changed it to ```TypeError``` and added some tests covering that branch - but it looks that ```cpython``` uses ```ValueError``` right here.

Should we switch to ```TypeError``` or rather stay with ```cpython``` behavior?

tests/snippets/ints.py
mpajkowski: ok, I'll fix it in a moment

vm/src/obj/objint.rs
windelbouwman: Ah, okay, a `ValueError` is a better idea. I meant more the error message which is returned from the `std::from_utf8` function, we might as well use it, to indicate in a better way what went wrong.

```rust
.map_err(|e| vm.new_value_error(format!("utf8 decode error: {}", e)))?;
```

mpajkowski: Ok, so I'll return ```ValueError``` and apply Your suggestion :)

vm/src/obj/objint.rs
```diff
@@ -766,6 +770,49 @@ pub fn to_int(vm: &VirtualMachine, obj: &PyObjectRef, mut base: u32) -> PyResult
     )
 }
 
+fn str_to_int(vm: &VirtualMachine, literal: &str, mut base: u32) -> PyResult<BigInt> {
+    let mut buf = literal.chars().filter(|&c| c != '_').collect::<String>();
```
seeeturtle: cpython has some rules for underscore in integer literal:
there can be only *one* underscore in front of the digit (not prefix)

examples:
```
>>> int('_0x121223', base=16)
ValueError: invalid literal for int() with base 16: '_0x121223'
>>> int('121223_')
ValueError: invalid literal for int() with base 10: '121223_'
>>> int('121__223')
ValueError: invalid literal for int() with base 10: '121__223'
```

vm/src/obj/objint.rs
```diff
@@ -766,6 +770,49 @@ pub fn to_int(vm: &VirtualMachine, obj: &PyObjectRef, mut base: u32) -> PyResult
...
+            } else {
+                base = matched_radix;
+            }
+            buf.drain(radix_range);
+        }
+    }
+
+    // base still not found, use default
+    if base == 0 {
+        base = 10;
```
seeeturtle: if base is 0 cpython doesn't allow base 10 integer starting with zero. the only acceptable case is when it is zero.
```
>>> int('0123', base=0)
ValueError: invalid literal for int() with base 0: '0123'
```
however `BigInt::from_str_radix` allows. there should be check for the case

PR #1291, opened 2019-08-16
Increase performance for get_item_option.

windelbouwman:
I finally got some idea that works in short circuiting the get_item_option for the case of strings. For example, this happens with the load_name opcode.

Idea is to short circuit the get_item logic and prevent several things from happening:
- Creation of the `KeyError`. Simply check for `None` or `Some`.
- Creation of a PyString when we lookup values by `str`.

### DIFF
--- a/vm/src/obj/objdict.rs
+++ b/vm/src/obj/objdict.rs
@@ -200,7 +200,18 @@ impl PyDictRef {
         value: PyObjectRef,
         vm: &VirtualMachine,
     ) -> PyResult<()> {
-        self.entries.borrow_mut().insert(vm, &key, value)
+        self.inner_setitem_fast(&key, value, vm)
+    }
+
+    /// Set item variant which can be called with multiple
+    /// key types, such as str to name a notable one.
+    fn inner_setitem_fast<K: DictKey + IntoPyObject + Copy>(
+        &self,
+        key: K,
+        value: PyObjectRef,
+        vm: &VirtualMachine,
+    ) -> PyResult<()> {
+        self.entries.borrow_mut().insert(vm, key, value)
     }
 
     #[cfg_attr(feature = "flame-it", flame("PyDictRef"))]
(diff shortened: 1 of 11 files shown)


vm/src/obj/objdict.rs
```diff
@@ -205,14 +205,31 @@ impl PyDictRef {
...
+            Ok(value)
+        } else {
+            Err(vm.new_key_error(key.clone()))
+        }
+    }
+
+    /// Return an optional inner item, or an error (can be key error as well)
+    fn inner_getitem_option<K: DictKey + IntoPyObject + Clone>(
+        &self,
+        key: &K,
```
coolreader18: I'd say to accept just a `K` in order to avoid a double allocation when you `key.clone()`.

coolreader18: Here's an idea: implement `DictKey for Cow<str>`, since we only sometimes have to get an owned version of the `DictKey`.

vm/src/vm.rs
```diff
@@ -936,7 +936,7 @@ impl VirtualMachine {
         }
 
         let attr = if let Some(ref dict) = obj.dict {
-            dict.get_item_option(name_str.clone(), self)?
+            dict.get_item_option(&name_str.value, self)?
```
coolreader18:
```suggestion
            dict.get_item_option(name_str.as_object().clone(), self)?
```

vm/src/obj/objdict.rs
coolreader18: Not necessarily in this PR, but just something that crossed my mind.

vm/src/vm.rs
windelbouwman: Note that this will create a full PyObject, and be less efficient than the `String` implementation of the hashing.

vm/src/obj/objdict.rs
windelbouwman: Yes, thank you, I did this, and it worked nicely! Now we can also pass `&str` values to these functions!

windelbouwman:
Just did some benchmarking with the `benchmarks` folder. This change makes the mandelbrot 66 times slower than cpython instead of 179 times.

vm/src/vm.rs
coolreader18: Ah, never mind then.

Issue #1295, opened 2019-08-17
MacOS build fail due to getgroups


corona10:
```bash
error[E0425]: cannot find function `getgroups` in module `nix::unistd`
   --> vm/src/stdlib/os.rs:265:35
    |
265 |     let groups_ids = nix::unistd::getgroups()?;
    |                                   ^^^^^^^^^ not found in `nix::unistd`
help: possible candidate is found in another module, you can import it into scope
    |
1   | use libc::getgroups;
    |
```

corona10:
I am working on this issue.

PR #1296, opened 2019-08-17
use getgroups(2) for apple macos.

corona10:
- nix-rust/nix mentioned that this feature should be implemented
based on opendirectoryd. However, CPython implemented it based on getgroups(2)
for apple macOS.
So I ported the code from nix-rust/nix and it works well.

- Add apple macOS tests for travis CI.

ref0: https://github.com/nix-rust/nix/blob/500036a206188dc8f57bf95a7c61616db417038e/src/unistd.rs#L1320
ref1: https://github.com/python/cpython/blob/c4cacc8c5eab50db8da3140353596f38a01115ca/Modules/posixmodule.c#L6852

fix: #1295

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,22 @@ before_cache:
 matrix:
   fast_finish: true
   include:
-    - name: Run Rust tests
+    - name: Run Rust tests(linux)
       language: rust
+      os: linux
+      rust: stable
+      cache: cargo
+      script:
+        - cargo build --verbose --all
+        - cargo test --verbose --all
+      env:
+        # Prevention of cache corruption.
+        # See: https://docs.travis-ci.com/user/caching/#caches-and-build-matrices
+        - JOBCACHE=1
+
+    - name: Run Rust tests(osx)
+      language: rust
+      os: osx
       rust: stable
       cache: cargo
       script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ matrix:
       env:
         # Prevention of cache corruption.
         # See: https://docs.travis-ci.com/user/caching/#caches-and-build-matrices
-        - JOBCACHE=1
+        - JOBCACHE=11
 
     # To test the snippets, we use Travis' Python environment (because
     # installing rust ourselves is a lot easier than installing Python)
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -262,7 +262,7 @@ fn get_right_permission(
     let others_permissions = get_permissions(others_mode);
 
     let user_id = nix::unistd::getuid();
-    let groups_ids = nix::unistd::getgroups()?;
+    let groups_ids = getgroups()?;
 
     if file_owner == user_id {
         Ok(owner_permissions)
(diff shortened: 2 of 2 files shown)


vm/src/stdlib/os.rs
```diff
@@ -273,6 +273,35 @@ fn get_right_permission(
...
+            libc::getgroups(
+                groups.capacity() as c_int,
+                groups.as_mut_ptr() as *mut gid_t,
+            )
+        };
+
+        match Errno::result(ret) {
+            Ok(s) => {
+                unsafe { groups.set_len(s as usize) };
+                return Ok(groups);
```
mpajkowski:
```suggestion
                Ok(groups)
```

vm/src/stdlib/os.rs
```diff
@@ -273,6 +273,35 @@ fn get_right_permission(
...
+    let mut groups = Vec::<Gid>::with_capacity(Errno::result(ret)? as usize);
+    loop {
+        let ret = unsafe {
+            libc::getgroups(
+                groups.capacity() as c_int,
+                groups.as_mut_ptr() as *mut gid_t,
+            )
+        };
+
+        match Errno::result(ret) {
```
mpajkowski: Errno::result(err).map should also work here

vm/src/stdlib/os.rs
```diff
@@ -273,6 +273,35 @@ fn get_right_permission(
...
+                return Ok(groups);
+            }
+            Err(e) => return Err(e),
+        }
+    }
+}
+
+#[cfg(target_os = "linux")]
+fn getgroups() -> nix::Result<Vec<Gid>> {
+    return nix::unistd::getgroups();
```
mpajkowski:
```suggestion
    nix::unistd::getgroups()
```

vm/src/stdlib/os.rs
```diff
@@ -273,6 +273,35 @@ fn get_right_permission(
...
+            }
+            Err(e) => return Err(e),
+        }
+    }
+}
+
+#[cfg(target_os = "linux")]
+fn getgroups() -> nix::Result<Vec<Gid>> {
+    nix::unistd::getgroups()
+}
```
coolreader18: You could also just `use nix::unistd::getgroups` here

vm/src/stdlib/os.rs
corona10: I am not familiar with transformed it into the map.
Can you guide it, please?

coolreader18: You could do something like `return Errno::result(ret).map(|s| { unsafe { set_len... }; groups});`, because you're not transforming the err case at all, so there's no need to do that whole `match`.

coolreader18: You could do something like `return Errno::result(ret).map(|s| { unsafe { set_len... }; groups});`, because you're not transforming the err case at all, so there's no need to do that whole `match`.

corona10: @coolreader18 
Thanks alot!

corona10:
@coolreader18 
I've checked that on my local MacBook laptop success to run `tests/snippets/stdlib_os.py`
I will rollback travis CI script for osx if this is travis issue.

corona10:
@coolreader18 
Please take a look!

coolreader18:
The code looks good, but I'm not sure about adding macos testing on Travis. @windelbouwman what do you think?

corona10:
IMHO, CI for mac should be considered. During pycon sprint 99% of partipants were using apple macbook pro. This kinds of build failure can be occured uncomfortable experience for new comers

youknowone:
I agree to @corona10. macOS is one of the major platform for programmers, but due to lack of CI, tests in macOS was broken time by time. Is there any panelty to adding a CI platform?

coolreader18:
FWIW, I'm up for it, or at least checking that it builds on macos, I just wanted to see what others would think.

.travis.yml
```diff
@@ -12,8 +12,22 @@ before_cache:
...
+      os: linux
+      rust: stable
+      cache: cargo
+      script:
+        - cargo build --verbose --all
+        - cargo test --verbose --all
+      env:
+        # Prevention of cache corruption.
+        # See: https://docs.travis-ci.com/user/caching/#caches-and-build-matrices
+        - JOBCACHE=1
```
palaviv: Please choose a different `JOBCACHE`. You now have the same one for linux and osx.

corona10: @palaviv Thanks for the comment, I've updated job cache

windelbouwman:
We had discussions before about the amount of CI jobs. I'm in favor of having not too many to save CPU time, and time on pull request builds. However, I would like to have the macOS job, it seems more used than redox, so this might be more important to run.

.travis.yml
windelbouwman: @corona10 please take a look at all other jobs as well, we require unique cache values for each job, otherwise jobs will become corrupted.

corona10:
@windelbouwman @palaviv 
I've updated job cache for macOS as `JOBCACHE=11` which is unique.
Please take a look.

Issue #1302, opened 2019-08-18
Name resolving error for methods


youknowone:
## Feature
```python
hits = []

def _x(a, b):
    return 1

class A:

    def _x(self, b):
        assert(not hits)
        hits.append(1)
        return _x(self, b) + 1

a = A()
assert a._x(0) == 2
```

this is working for CPython but stack overflow(caught by `assert (not hits)`) in RustPython

coolreader18:
I think this is because we very simply treat a class as a scope, rather than as a store for attributes in the build class stage but not when you're using methods.

windelbouwman:
Agreed, this is a scoping thing which is obviously wrong.

windelbouwman:
@cthulahoops what is your view on this?

cthulahoops:
Yes, definitely a scoping bug, here's the simple example:

```
y = 7
class A:
    y = 8
    assert y == 8
    def __init__(self):
        assert y == 7 
A()
```

It's weird because we end up with two different scopes, the one captured by ```def``` and ```class```, and the one used by other code.

windelbouwman:
It appears the scope which is used during runtime is different then the scope used during construction of the class.

cthulahoops:
I think the fix is to tag the scope as an attribute scope and then the function/class/lambda def can check and remove it from the scope for the function.

I was wondering if we can have the compiler generate slightly different code for declarations inside classes and avoid the work at runtime, but I don't think that works:

```
>>> c = compile("f = lambda self: y", "", "exec")
>>> class A():
...     y = 5
...     exec(c)
```

windelbouwman:
I think the trick is to assign proper tags to scopes.

This documentation gives a hint already: https://docs.python.org/3/library/symtable.html?highlight=symtable#symtable.symtable

I will check it out further.

windelbouwman:
I tried to search through cpython, micropython and gpython, but I could not find where this behavior is implemented. Should it be implemented in the compiler, bytecode, during runtime?

windelbouwman:
Maybe the trick lies within the build_class builtin method?

cthulahoops:
I don't think the built in can do it alone. I think it has to be in the runtime handling of function def (etc.), but I'm not sure.

coolreader18:
What if we checked the type of the surrounding scope in the `MakeFunction`  instruction?

youknowone:
The given example codes are now working. If this is not finished yet, please reopen this issue.

PR #1301, opened 2019-08-18
Add isdst and slice support to time.struct_time

youknowone:

PR #1303, opened 2019-08-18
Datetime module

youknowone:
~#1301 is required for tests~ 

#1302 is a blocking issue for other tests

### DIFF
--- a/tests/snippets/testutils.py
+++ b/tests/snippets/testutils.py
@@ -20,7 +20,17 @@ def assert_raises(exc_type, expr, msg=None):
         assert False, failmsg
 
 
-class assertRaises:
+def assertRaises(expected, *args, **kw):
+    if not args:
+        assert not kw
+        return _assertRaises(expected)
+    else:
+        f, f_args = args[0], args[1:]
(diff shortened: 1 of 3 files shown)


windelbouwman:
@youknowone do you want to merge this work, or wait until #1302 is solved?

youknowone:
`datetime.py` itself is (partially) working without editing. So merging now will be fine. Bugs will be naturally fixed by catching up other issues.

tests/snippets/testutils.py
```diff
@@ -20,7 +20,17 @@ def assert_raises(exc_type, expr, msg=None):
         assert False, failmsg
 
 
-class assertRaises:
+def assertRaises(expected, *args, **kw):
```
windelbouwman: What does this function do? We already have another function` assert_raises`, is this what you were looking for? Note that this function might be tricky since it now becomes very versatile.

youknowone: About the tricky interface, it is just made for interface compatibility with unittest.assertRaises. Because the basic implementation of assertRaises was compatible with it, i thought it is intended to be compatible.

PR #1304, opened 2019-08-19
Add the array module

coolreader18:
Not all the methods for `array.array` are implemented, but the most relevant ones are.

### DIFF
--- a/vm/src/stdlib/array.rs
+++ b/vm/src/stdlib/array.rs
@@ -16,6 +16,7 @@ macro_rules! def_array_enum {
             $($n(Vec<$t>),)*
         }
 
+        #[allow(clippy::naive_bytecount, clippy::float_cmp)]
         impl ArrayContentType {
             fn from_char(c: char) -> Option<Self> {
                 match c {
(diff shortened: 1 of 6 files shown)


vm/src/stdlib/array.rs
```diff
@@ -0,0 +1,306 @@
...
+            fn getitem(&self, i: usize, vm: &VirtualMachine) -> Option<PyResult> {
+                match self {
+                    $(ArrayContentType::$n(v) => v.get(i).map(|x| x.into_pyobject(vm)),)*
+                }
+            }
+        }
+    };
+}
+
+def_array_enum!(
```
windelbouwman: Clever usage of macros!

vm/src/stdlib/array.rs
```diff
@@ -0,0 +1,306 @@
...
+            1 => spec.as_str().chars().next().unwrap(),
+            _ => {
+                return Err(vm.new_type_error(
+                    "array() argument 1 must be a unicode character, not str".to_owned(),
+                ))
+            }
+        };
+        let array = ArrayContentType::from_char(spec).ok_or_else(|| {
+            vm.new_value_error(
+                "bad typecode (must be b, B, u, h, H, i, I, l, L, q, Q, f or d)".to_owned(),
```
windelbouwman: It might be nice to generate this error message in the `from_char` method, and let it return this error message and use `map_err`.

windelbouwman:
It might be useful to add a snippet with some python code which demonstrates the array basic usecases, so that we have some test coverage of this code.

PR #1305, opened 2019-08-19
Repair try-except-else syntax. Fixes #1300.

### LINKED ISSUE #1300: try-else without except clause is illegal
## Feature

The code is SyntaxError in cpython, but legal in RustPython

```python
try:
    pass
else:
    pass
```

## Python Documentation

https://docs.python.org/3/tutorial/errors.html#handling-exceptions

```
The try … except statement has an *optional* else clause,
```

windelbouwman:


corona10:
@windelbouwman
````
 python
>>> try:
...    pass
...
  File "<stdin>", line 3

    ^
SyntaxError: invalid syntax
````

This should work as same as CPython.

windelbouwman:
@corona10  I'll add this right away!

PR #1311, opened 2019-08-19
Allow time.time() to work on WASM

coolreader18:


vm/src/vm.rs
```diff
@@ -100,6 +100,15 @@ pub struct PySettings {
 
     /// sys.argv
     pub argv: Vec<String>,
+
+    pub syscall_functions: PySyscallFunctions,
```
windelbouwman: I'm a big fan of the [dependency injection pattern](https://en.wikipedia.org/wiki/Dependency_injection), but, in this case I feel this is not the right choice.

Instead, I would solve this issue with `#cfg` macro's in the time module. Rationale for solving is like that is that I can imagine the `PySycallFunctions` struct grow to an unwieldy size.

vm/src/stdlib/time_module.rs
```diff
@@ -36,22 +36,24 @@ fn duration_to_f64(d: Duration) -> f64 {
...
-        Err(err) => panic!("Error: {:?}", err),
-    };
-    let value = vm.ctx.new_float(x);
-    Ok(value)
+        Err(err) => panic!("Time error: {:?}", err),
+    }
+}
+
+fn time_time(vm: &VirtualMachine) -> f64 {
+    let time_func = vm.settings.syscall_functions.time.unwrap_or(get_time);
```
windelbouwman: Instead of adding a `syscall_functions` member, create two `time_time` functions and select them with the `#cfg` structure. If you want to reduce duplicated code, you can create a helper function which you implement for each platform.

It might even make sense to create specific platform subfolders in the `stdlib` folder with the specifics for each platform.

vm/src/vm.rs
coolreader18: The issue is that that would require adding `wasm-bindgen` as a dependency of `rustpython-vm`, which I wasn't sure about doing.

windelbouwman: In this case I would recommend adding the wasm-bindgen as a dependency to the rustpython-vm crate. You can make this a wasm target only dependency with dependencies based upon configuration options.

At some point it might make sense to split the vm crate into two crate. One core vm crate, with the bare minimal modules, and a stdlib crate, which contains all the hairy details dealing with platform specifics.

PR #1313, opened 2019-08-20
Raise a KeyboardInterrupt on a SIGINT

coolreader18:
This is based on #1311 for the PySyscallFunctions struct.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -583,28 +583,26 @@ fn run_shell(vm: &VirtualMachine, scope: Scope) -> PyResult<()> {
     use std::io::{self, BufRead, Write};
 
     println!(
-        "Welcome to the magnificent Rust Python {} interpreter \u{1f631} \u{1f596}",
+        "Welcome to the magnificent Rust Python {} interpreter!",
         crate_version!()
     );
 
-    let stdin = io::stdin();
-    let stdout = io::stdout();
-    let mut stdout = stdout.lock();
-
-    let mut print_prompt = || {
+    fn print_prompt() {
         let prompt = get_prompt(vm, "ps1");
         let prompt = match prompt {
             Some(ref s) => s.as_str(),
             None => "",
         };
         print!("{}", prompt);
-        stdout.flush().expect("flush failed");
-    };
+        io::stdout().lock().flush().expect("flush failed");
+    }
+
+    let stdin = io::stdin();
 
     print_prompt();
     for line in stdin.lock().lines() {
         let mut line = line.expect("line failed");
-        line.push('\n');
+        line.push_str("\n");
         match shell_exec(vm, &line, scope.clone()) {
             ShellExecResult::Ok => {}
             ShellExecResult::Continue => println!("Unexpected EOF"),
--- a/src/main.rs
+++ b/src/main.rs
@@ -591,8 +591,17 @@ fn run_shell(vm: &VirtualMachine, scope: Scope) -> PyResult<()> {
     let stdout = io::stdout();
     let mut stdout = stdout.lock();
 
-    print!("{}", get_prompt(vm, "ps1"));
-    stdout.flush().expect("flush failed");
+    let mut print_prompt = || {
+        let prompt = get_prompt(vm, "ps1");
+        let prompt = match prompt {
+            Some(ref s) => s.as_str(),
+            None => "",
+        };
+        print!("{}", prompt);
+        stdout.flush().expect("flush failed");
+    };
+
+    print_prompt();
     for line in stdin.lock().lines() {
         let mut line = line.expect("line failed");
         line.push('\n');
@@ -601,8 +610,7 @@ fn run_shell(vm: &VirtualMachine, scope: Scope) -> PyResult<()> {
             ShellExecResult::Continue => println!("Unexpected EOF"),
             ShellExecResult::PyErr(exc) => print_exception(vm, &exc),
         }
-        print!("{}", get_prompt(vm, "ps1"));
-        stdout.flush().expect("flush failed");
+        print_prompt();
     }
 
     Ok(())
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,7 +9,7 @@ use rustpython_compiler::{compile, error::CompileError, error::CompileErrorType}
 use rustpython_parser::error::ParseErrorType;
 use rustpython_vm::{
     import,
-    obj::objstr,
+    obj::objstr::PyStringRef,
     print_exception,
     pyobject::{ItemProtocol, PyResult},
     scope::Scope,
@@ -486,12 +486,10 @@ fn get_history_path() -> PathBuf {
     xdg_dirs.place_cache_file("repl_history.txt").unwrap()
 }
 
-fn get_prompt(vm: &VirtualMachine, prompt_name: &str) -> String {
+fn get_prompt(vm: &VirtualMachine, prompt_name: &str) -> Option<PyStringRef> {
     vm.get_attribute(vm.sys_module.clone(), prompt_name)
+        .and_then(|prompt| vm.to_str(&prompt))
         .ok()
-        .as_ref()
-        .map(objstr::get_value)
-        .unwrap_or_else(String::new)
 }
 
 #[cfg(not(target_os = "redox"))]
@@ -521,7 +519,8 @@ fn run_shell(vm: &VirtualMachine, scope: Scope) -> PyResult<()> {
         } else {
             get_prompt(vm, "ps1")
         };
-        match repl.readline(&prompt) {
+        let prompt = prompt.as_ref().map(|s| s.as_str()).unwrap_or("");
+        match repl.readline(prompt) {
             Ok(line) => {
                 debug!("You entered {:?}", line);
                 input.push_str(&line);
@@ -542,7 +541,6 @@ fn run_shell(vm: &VirtualMachine, scope: Scope) -> PyResult<()> {
                         ..
                     }) => {
                         continuing = true;
-                        continue;
                     }
                     _ => {
                         input = String::new();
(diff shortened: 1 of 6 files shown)


vm/src/stdlib/signal.rs
```diff
@@ -23,51 +23,64 @@ const SIG_IGN: libc::sighandler_t = 1;
...
 
 // We cannot use the NSIG const in the arr macro. This will fail compilation if NSIG is different.
-static mut TRIGGERS: [AtomicBool; NSIG] = arr![AtomicBool::new(false); 64];
+static TRIGGERS: [AtomicBool; NSIG] = arr![AtomicBool::new(false); 64];
 
 extern "C" fn run_signal(signum: i32) {
-    unsafe {
-        TRIGGERS[signum as usize].store(true, Ordering::Relaxed);
-    }
+    TRIGGERS[signum as usize].store(true, Ordering::Relaxed);
```
windelbouwman: Nice one! The `AtomicBool` appears the be mutable itself!

vm/src/stdlib/signal.rs
```diff
@@ -23,51 +23,64 @@ const SIG_IGN: libc::sighandler_t = 1;
...
+    );
     Ok(old_handler)
 }
 
-fn getsignal(signalnum: PyIntRef, vm: &VirtualMachine) -> PyResult<Option<PyObjectRef>> {
-    let signalnum = signalnum.as_bigint().to_i32().unwrap();
-    Ok(vm.signal_handlers.borrow_mut().get(&signalnum).cloned())
+fn getsignal(signalnum: i32, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
+    assert_in_range(signalnum, vm)?;
+    Ok(vm.signal_handlers.borrow()[signalnum as usize].clone())
```
windelbouwman: We should probably check for signalnum boundaries in the array.

vm/src/stdlib/time_module.rs
```diff
@@ -22,36 +22,62 @@ use num_traits::cast::ToPrimitive;
...
+}
+
+fn time_sleep(seconds: PyFloatRef, vm: &VirtualMachine) -> PyResult<()> {
+    let seconds = seconds.to_f64();
     let secs: u64 = seconds.trunc() as u64;
     let nanos: u32 = (seconds.fract() * 1e9) as u32;
     let duration = Duration::new(secs, nanos);
-    thread::sleep(duration);
-    Ok(vm.get_none())
+    let sleep_func = vm.settings.syscall_functions.sleep.unwrap_or(do_sleep);
```
windelbouwman: See my other comment in PR #1311 about this construction. I think we should not add the `syscall_functions` member.

vm/src/stdlib/signal.rs
coolreader18: That's what `assert_in_range` does.

vm/src/stdlib/time_module.rs
coolreader18: I decoupled this PR from that change, we can look into that separately.

vm/src/stdlib/signal.rs
coolreader18: That's what `assert_in_range` does.

vm/src/vm.rs
```diff
@@ -166,7 +168,10 @@ impl VirtualMachine {
         let import_func = RefCell::new(ctx.none());
         let profile_func = RefCell::new(ctx.none());
         let trace_func = RefCell::new(ctx.none());
+
         let vm = VirtualMachine {
+            #[cfg(not(target_arch = "wasm32"))]
```
windelbouwman: We can have this signal handlers in wasm as well right? They do not hurt us there, so this `#[cfg]` could be removed.

src/main.rs
```diff
@@ -486,12 +486,10 @@ fn get_history_path() -> PathBuf {
     xdg_dirs.place_cache_file("repl_history.txt").unwrap()
 }
 
-fn get_prompt(vm: &VirtualMachine, prompt_name: &str) -> String {
```
windelbouwman: I would like to keep the old signature and return a `String` here. What is the benefit of returning the `Option` here?

vm/src/stdlib/signal.rs
windelbouwman: Btw, did you run benchmarks on this change? This change might trigger a performance boost :)

vm/src/stdlib/signal.rs
windelbouwman: Ah, I missed that. Thanks!

vm/src/vm.rs
coolreader18: It uses the `NSIG` constant from `signal.rs`; should I move that into `vm.rs`?

src/main.rs
coolreader18: It doesn't clone the string this way, `vm.to_str` returns a `PyStringRef` whose inner `String` is already held inside of a PyObjectRef on the heap, but `vm.to_pystr` clones that `String` when all `repl.readline()` needs is an `&str`. I know it's a really tiny micro-optimization, especially when 99% of the time the string's only going to be 5 characters long, but I just like doing them.

vm/src/vm.rs
windelbouwman: Yes, this is probably a good idea to keep the `NSIG` constant where the array is defined / created.

src/main.rs
```diff
@@ -542,18 +541,18 @@ fn run_shell(vm: &VirtualMachine, scope: Scope) -> PyResult<()> {
...
                     }
                 }
             }
             Err(ReadlineError::Interrupted) => {
-                // TODO: Raise a real KeyboardInterrupt exception
-                println!("^C");
+                let exc = vm
+                    .new_empty_exception(vm.ctx.exceptions.keyboard_interrupt.clone())
+                    .unwrap();
+                print_exception(vm, &exc);
```
windelbouwman: There is something which occurs strange to me when examining this piece of code. The exception is created and printed, but should it not be inserted into the virtualmachine in some way? Maybe we can inject it into the virtual machine in a similar manner in which the signals are inserted into the virtualmachine, with the check_signals function and some sort of a queue.

I would suggest to add a function to the virtualmachine, named something like `inject_exception`, which will put this exception object into a queue, or a slot, and the frame eval loop will pick it up.

windelbouwman:
@palaviv could you have a look at this change as well?

src/main.rs
coolreader18: The problem here is that the code I was converting from CPython here is actually from the builtin `input()` function, which then gets used by the `code` Python module to run the shell. Would it be okay if I left it like this for now but added `input()` and `code.py` in a separate PR?

vm/src/stdlib/signal.rs
```diff
@@ -23,51 +23,64 @@ const SIG_IGN: libc::sighandler_t = 1;
...
-    let sig_dfl = vm.get_attribute(signal_module.clone(), "SIG_DFL")?;
-    let sig_ign = vm.get_attribute(signal_module, "SIG_IGN")?;
-    let signalnum = signalnum.as_bigint().to_i32().unwrap();
-    check_signals(vm);
-    let sig_handler = if handler.is(&sig_dfl) {
-        SIG_DFL
-    } else if handler.is(&sig_ign) {
-        SIG_IGN
+fn assert_in_range(signum: i32, vm: &VirtualMachine) -> PyResult<()> {
+    if (1..=NSIG as i32).contains(&signum) {
```
palaviv: We should not allow `signum` to be equal to `NSIG`

vm/src/stdlib/signal.rs
```diff
@@ -81,50 +94,68 @@ fn alarm(time: PyIntRef, _vm: &VirtualMachine) -> u32 {
...
-            let handler = vm
-                .signal_handlers
-                .borrow()
-                .get(&(signum as i32))
-                .expect("Handler should be set")
-                .clone();
-            vm.invoke(&handler, vec![vm.new_int(signum), vm.get_none()])
-                .expect("Test");
+            let handler = &vm.signal_handlers.borrow()[signum];
+            if vm.is_callable(handler) {
```
palaviv: How can the handler be not callable? We check this when we set the signal.

coolreader18: If it's `SIG_DFL` or `SIG_IGN`, that still gets stored in the array and returned as the old handler when you call `signal()`.

src/main.rs
coolreader18: Actually no, it doesn't use `code.py`, that's just trying to emulate it. I'll look into this further.

coolreader18:
@windelbouwman I refactored the prompt a bit, and although that's sort of out of scope of the rest of the changes, I feel like it better represents the (data?)flow of the shell.

src/main.rs
```diff
@@ -582,14 +591,26 @@ fn run_shell(vm: &VirtualMachine, scope: Scope) -> PyResult<()> {
     let stdout = io::stdout();
     let mut stdout = stdout.lock();
 
-    print!("{}", get_prompt(vm, "ps1"));
-    stdout.flush().expect("flush failed");
+    let mut print_prompt = || {
```
windelbouwman: This does not need to be `mut`. You might also define a normal function like `fn print_prompt() {` here.

vm/src/vm.rs
```diff
@@ -166,6 +169,8 @@ impl VirtualMachine {
         let import_func = RefCell::new(ctx.none());
         let profile_func = RefCell::new(ctx.none());
         let trace_func = RefCell::new(ctx.none());
+        let signal_handlers = RefCell::new(arr![ctx.none(); 64]);
```
windelbouwman: This `64` can be the `NSIG` from above right?

windelbouwman: Ah, okay.

PR #1314, opened 2019-08-20
bool.real

youknowone:


cthulahoops:
This is an interesting case. The problem affects all subclasses of int, not just bool.

Given:

```
class X(int):
   pass
```

CPython:
```
>>> type(X(5).real)
<class 'int'>
```

RustPython:

```
>>>>> type(X(5).real)
<class 'X'>
```

I think the correct fix is for ```int_real``` to construct a new int rather than returning the object itself.

youknowone:
@cthulahoops changed int impl

coolreader18 (approved):

PR #1315, opened 2019-08-21
#1224 not implemented pattern (for now)

jmg7173:
Add error handling for not implemented patterns:

* `%#.#f`
* `%e` and `%E`
* `%g` and %G`

To implement this features, see this cpython implementation reference:
* [PyOS_double_to_string](https://github.com/python/cpython/blob/master/Python/pystrtod.c#L797~L943) for no short float 
* [PyOS_double_to_string](https://github.com/python/cpython/blob/master/Python/pystrtod.c#L1224~L1307) for short float 

I must also handle for `%r` expression but I can't put where to handle it.
Thank you :)


vm/src/obj/objstr.rs
```diff
@@ -1231,18 +1231,22 @@ fn do_cformat_specifier(
...
-                Err(vm.new_type_error(format!(
+                return Err(vm.new_type_error(format!(
                     "%{} format: {} is required, not {}",
                     format_spec.format_char,
                     required_type_string,
                     obj.class()
-                )))
+                )));
+            };
+            match result {
```
windelbouwman: Instead of this `match` statement, you might want to choose the `map_err` function like this:
```rust

if is_float {
    ...
} else if is_int {
    ...
} else {
    ...
}.map_err(|e| vm.new_not_implemented_error(e))
```

jmg7173: Do you mean that it should change like you say at Line 1234 to 1246?

Or not, really at Line 1247, I think it doesn't need to change like you say because `format_float` always returns `String` type `Result`.
```rust
pub fn format_float(&self, num: f64) -> Result<String, String>
```

So there is nothing to map like that.

Or, is there any good example for handling `Result` type returns?

Thank you for reviewing :)

windelbouwman: My comment is an idea to simplify the `match` statement at the end here. Let me try to clarify. You could change this:

```rust
match result {
    Ok(transformed) => Ok(transformed),
    Err(error) => Err(vm.new_not_implemented_error(error)),
}
```

Into this:

```rust
result.map_err(|e| vm.new_not_implemented_error(e))
```

The `map_err` can be used here to transform the type `Result<String, String>` into `PyResult<String>`.

I hope I could explain this idea.

jmg7173: I changed that part :)
I hope that's what you explained to me.

windelbouwman: Looks good to me! Nicely done!

PR #1318, opened 2019-08-25
ast module

youknowone:


vm/src/stdlib/ast.rs
```diff
@@ -638,7 +638,7 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
     let ctx = &vm.ctx;
```
coolreader18: I think you missed `"ast" -> "_ast"` in `create_node` and `AstNode::class` up above.

windelbouwman: Nice catch! See line 24. It might make sense to declare a global constant at the top of this file named `const module_name = "_ast";` and use it where appropriate.

Lib/ast.py
```diff
@@ -0,0 +1,332 @@
...
+
+
+    :copyright: Copyright 2008 by Armin Ronacher.
+    :license: Python License.
+"""
+from _ast import *
+
+
+# XXX: Rustpython _ast includes parse()
+#def parse(source, filename='<unknown>', mode='exec'):
```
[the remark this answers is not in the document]
windelbouwman: I would like to add the mode optional argument to the rust parse function. Note that the mode can be retrieve because mode implements `FromStr` now. Please refer to the builtin compile function how this is done.

parser/src/parser.rs
```diff
@@ -73,6 +73,50 @@ pub fn parse_expression(source: &str) -> Result<ast::Expression, ParseError> {
...
+        Mode::Eval => {
+            let statement = parse_statement(source)?;
+            ast::Top::Statement(statement)
+        }
+    })
+}
+
+// compile mode but defined here
+#[derive(Clone, Copy)]
+pub enum Mode {
```
windelbouwman: I'm not too sure that this `Mode` enum should be contained in the parser crate. I feel it has its place in the compiler crate (where it was). The new `parse` function can be place in near the compile function in the builtins.rs in the vm crate.

parser/src/parser.rs
```diff
@@ -73,6 +73,50 @@ pub fn parse_expression(source: &str) -> Result<ast::Expression, ParseError> {
     do_lalr_parsing!(source, Expression, StartExpression)
 }
 
+pub fn parse(source: &str, mode: Mode) -> Result<ast::Top, ParseError> {
```
windelbouwman: I suggest to move this function to the `builtins.rs` in the vm crate.

vm/src/stdlib/ast.rs
```diff
@@ -48,14 +50,27 @@ macro_rules! node {
     }
 }
 
-fn program_to_ast(vm: &VirtualMachine, program: &ast::Program) -> PyResult<AstNodeRef> {
-    let py_body = statements_to_ast(vm, &program.statements)?;
-    Ok(node!(vm, Module, { body => py_body }))
+pub(crate) fn parse(vm: &VirtualMachine, source: &str, mode: parser::Mode) -> PyResult {
```
windelbouwman: Probably move this into the `builtins.rs` file, where it is used.

parser/src/parser.rs
youknowone: It seems `Mode` should be defined out of `compile.rs` because `ast.parse` works without `#[cfg(feature = "rustpython-compiler")]`. Any idea?

youknowone: @windelbouwman any advice for proper place about this?

windelbouwman: ast.parse should not work without the compiler crate, since when you have the compiler crate, the parser crate is implied. Vice versa is not true, but I think it makes sense to excluse `ast.parse` when the compiler feature is disabled. So I think we should place `Mode` in `mode.rs` in the `rustpython-compiler` crate.

parser/src/parser.rs
windelbouwman: Yes, agreed, but in any case, this function should reside in the vm crate, since it uses a compilation mode.

PR #1320, opened 2019-08-26
Make bytecode::Label its own struct

coolreader18:
This helps to prevent accidentally using the `Label` type as a usize, as it was previously just a type alias and therefore could be used in any place a usize could, e.g. arithmetic operations. I actually caught a bug while I was making this change; the `fmt_dis` impls for `SetupExcept`, `SetupWith`, and `CleanupWith` printed the label value instead of the label target.

### DIFF
--- a/bytecode/src/bytecode.rs
+++ b/bytecode/src/bytecode.rs
@@ -56,7 +56,14 @@ bitflags! {
     }
 }
 
-pub type Label = usize;
+#[derive(Serialize, Debug, Deserialize, Clone, Copy, PartialEq, Eq, Hash)]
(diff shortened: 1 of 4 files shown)


windelbouwman:
This makes sense, to check that we actually use a `Label` and not just some random `usize`. Is this the only way to ensure this? I would have hoped the `Label` type to be strictly compatible with other `Label` types, and not `usize` values.

windelbouwman:
compatible is probably not the right word. `usize` was compatible with `Label`, but I would like to enforce only `Label` types to be usable where the `Label` type is specified. This is what you have done, and it is good.

PR #1321, opened 2019-08-27
Refactor block stack unwinding to enable finally execution.

windelbouwman:
This implements finally execution when an exception is raised and when code returns in a try statement.

- The unwind blocks function is cleaned up.

Note that the design deviates from cpython. In CPython, the reason for entering the finally clause is remembered on the value stack. I chose to use the blockstack instead.


windelbouwman:
This certainly was a rewarding refactor. The unwind functions took me some time to get a better structure.

PR #1322, opened 2019-08-28
fileio: Update closefd, closed when close method is called

corona10:


corona10:
@youknowone @windelbouwman Please take a look

windelbouwman:
Looks good to me! Thanks!

PR #1325, opened 2019-08-28
Add scope type and other symboltable properties.

windelbouwman:


compiler/src/symboltable.rs
```diff
@@ -273,9 +299,9 @@ impl SymbolTableBuilder {
         Ok(symbol_table)
     }
 
-    fn enter_block(&mut self, name: &str) {
+    fn enter_block(&mut self, name: &str, typ: SymbolTableType, line_number: usize) {
```
cthulahoops: Is enter_block correctly named? It's job is to create a new symbol table, but blocks don't create scopes in python. Should this be enter_scope, or push_table or something?

crawl_sourcecode.py
```diff
@@ -54,6 +54,10 @@ def print_node(node, indent=0):
 
 def print_table(table, indent=0):
```
youknowone: Suggestion:
```python
    indents = ' ' * indent;
    print(indents, 'table:', table.get_name())
    print(indents, '  name:', table.get_name())
    print(indents, '  type:', table.get_type())
    print(indents, '  line:', table.get_lineno())
    print(indents, '  identifiers:', table.get_identifiers())
```

compiler/src/symboltable.rs
windelbouwman: `enter_scope` would be a better name this function. I'll change it later on, I want to refactor this code more.

vm/src/stdlib/socket.rs
```diff
@@ -430,10 +430,10 @@ impl SocketRef {
                     self.timeout.borrow_mut().replace(Duration::from_secs(0));
                 }
                 if let Some(conn) = self.con.borrow_mut().as_mut() {
-                    return match conn.setblocking(value) {
+                    match conn.setblocking(value) {
```
pfalcon: > vm/src/stdlib/socket.rs

Nice change to shove in the "Add scope type and other symboltable properties." commit! But I'm missing the "conn.bind(); conn.listen(); conn.accept(); buf = conn.read(); eval(buf);" part. Is it in one of the other 4000 commits of this project? ;-)

youknowone:
#1302

PR #1326, opened 2019-08-28
Refactor snippets/testutil to use single assert_raises implementation

youknowone:
Due to enhancement of assertRaises, it seems these two methods doesn't need to be separated like now.

### DIFF
--- a/tests/snippets/builtin_all.py
+++ b/tests/snippets/builtin_all.py
@@ -6,5 +6,5 @@ assert not all([False])
 assert all([])
 assert not all([False, TestFailingBool()])
 
-assert_raises(RuntimeError, lambda: all(TestFailingIter()))
-assert_raises(RuntimeError, lambda: all([TestFailingBool()]))
+assert_raises(RuntimeError, all, TestFailingIter())
+assert_raises(RuntimeError, all, [TestFailingBool()])
--- a/tests/snippets/set.py
+++ b/tests/snippets/set.py
@@ -29,10 +29,10 @@ assert not set([1,3]) < set([1,2])
 
 assert (set() == []) is False
 assert set().__eq__([]) == NotImplemented
-assert_raises(TypeError, lambda: set() < [])  # '<' not supported between instances of 'set' and 'list'
-assert_raises(TypeError, lambda: set() <= [])  # '<=' not supported between instances of 'set' and 'list'
-assert_raises(TypeError, lambda: set() > [])  # '>' not supported between instances of 'set' and 'list'
-assert_raises(TypeError, lambda: set() >= [])  # '>=' not supported between instances of 'set' and 'list'
+assert_raises(TypeError, lambda: set() < [], _msg="'<' not supported between instances of 'set' and 'list'")
+assert_raises(TypeError, lambda: set() <= [], _msg="'<=' not supported between instances of 'set' and 'list'")
+assert_raises(TypeError, lambda: set() > [], _msg="'>' not supported between instances of 'set' and 'list'")
+assert_raises(TypeError, lambda: set() >= [], _msg="'>=' not supported between instances of 'set' and 'list'")
 assert set().issuperset([])
 assert set().issubset([])
 assert not set().issuperset([1, 2, 3])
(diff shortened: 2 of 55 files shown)


tests/snippets/builtin_all.py
```diff
@@ -6,5 +6,5 @@
 assert all([])
 assert not all([False, TestFailingBool()])
 
-assert_raises(RuntimeError, lambda: all(TestFailingIter()))
```
windelbouwman: We also could refactor this into:

```python
with assert_raises(RuntimeError):
    all(TestFailingIter())
```

What is the benefit of the lambda function type of assert_raises? In my opinion, the form with the `with` statement is clearer than the lambda-style.

What is your opinion on this? Should we go for one of the two styles?

tests/snippets/builtin_all.py
```diff
@@ -6,5 +6,5 @@
...
-assert_raises(RuntimeError, lambda: all([TestFailingBool()]))
+assert_raises(RuntimeError, all, TestFailingIter())
```
windelbouwman: Note that this behavior is slightly different than before. Before the TestFailingIter was create when the lambda function was invoked. Now, it is instantiated when the assert_raises function is called, so any exception in the constructor will not be catched.

tests/snippets/testutils.py
```diff
@@ -1,36 +1,14 @@
...
-    else:
-        failmsg = '{} was not raised'.format(exc_type.__name__)
-        if msg is not None:
-            failmsg += ': {}'.format(msg)
-        assert False, failmsg
-
-
-def assertRaises(expected, *args, **kw):
+def assert_raises(expected, *args, **kw):
     if not args:
```
windelbouwman: I usually use the rule: "whenever an if statement contains a not, I swap the arms of the if statement".

In this case, you could write:
```python
if args:
   ...
else:
   ...
```

In my opinion, this is easier to understand for readers.

tests/snippets/set.py
```diff
@@ -29,23 +29,23 @@
 
 assert (set() == []) is False
 assert set().__eq__([]) == NotImplemented
-assert_raises(TypeError, lambda: set() < [], "'<' not supported between instances of 'set' and 'list'")
-assert_raises(TypeError, lambda: set() <= [], "'<=' not supported between instances of 'set' and 'list'")
-assert_raises(TypeError, lambda: set() > [], "'>' not supported between instances of 'set' and 'list'")
-assert_raises(TypeError, lambda: set() >= [], "'>=' not supported between instances of 'set' and 'list'")
+assert_raises(TypeError, lambda: set() < [])  # '<' not supported between instances of 'set' and 'list'
```
windelbouwman: Would it be possible to retain the error message? This could be useful when the test fails.

windelbouwman:
Nice job! This seems like a lot of work!

tests/snippets/builtin_all.py
youknowone: Because this code is testing `all()`, i think it is fine. But it seems some other changes may bring a problem in same way.

Let me know if you want to  adapt more conservative replacement. I used regex replacement for most of the changes, so there can be some points i missed. (though passing all the tests implies there is no exceptions tests couldn't catch)

tests/snippets/set.py
youknowone:
```
  File "tests/snippets/set.py", line 32, in <module>
    assert_raises(TypeError, lambda: set() < set())  # '<' not supported between instances of 'set' and 'list'
  File "/Users/youknowone/Projects/RustPython/tests/snippets/testutils.py", line 8, in assert_raises
    f(*f_args, **kw)
  File "/Users/youknowone/Projects/RustPython/tests/snippets/testutils.py", line 22, in __exit__
    assert False, failmsg
AssertionError: TypeError was not raised
```

This is the result when it has error. The message is not printed at last line but still readable on the traceback as a part of code.

It will be still possible to keep the original behavior with a keyword argument. Do you prefer that way?

tests/snippets/builtin_all.py
youknowone: I agree that using lambda doesn't have much advantages for single case. I prefer to have single line test code to copy and edit many similar cases - which is common in test codes.

I don't have a nice answer about this, but maybe CPython code can give you some hints. I often can find this patterns in CPython test codes too. In this case, usually the lambda body is very short.

tests/snippets/testutils.py
youknowone: Fixed it

tests/snippets/set.py
windelbouwman: Having the error message along with the error, and not in the traceback is beneficial. It's not super important to me however.

youknowone: Please check the new commit is what you wanted

windelbouwman:
@youknowone I'm not 100% convinced refactoring the snippets is a good idea. In general I like the idea, but it also feels tricky to modify the snippets. 

@palaviv and @coolreader18 what is your opinion on this change?

youknowone:
@palaviv would you look into this PR? I will rebase and merge it

PR #1327, opened 2019-08-28
[WIP] Add stack size calculation to code object to pre-allocate value stack

windelbouwman:
This is still early work. Comments / ideas are welcome!

Idea is to determine the worst case stack size, and prevent a re-allocation of the value stack. Minor performance win, a lot of work :).

### DIFF
--- a/compiler/src/output_stream.rs
+++ b/compiler/src/output_stream.rs
@@ -1,5 +1,7 @@
 use rustpython_bytecode::bytecode::{CodeObject, Instruction, Label, Location};
 
+use crate::stack_effect::stack_effect;
+
 pub trait OutputStream: From<CodeObject> + Into<CodeObject> {
     /// Output an instruction
     fn emit(&mut self, instruction: Instruction, location: Location);
@@ -11,11 +13,17 @@ pub trait OutputStream: From<CodeObject> + Into<CodeObject> {
 
 pub struct CodeObjectStream {
     code: CodeObject,
+    max_stack: isize,
+    current_stack_size: isize,
 }
 
 impl From<CodeObject> for CodeObjectStream {
     fn from(code: CodeObject) -> Self {
-        CodeObjectStream { code }
+        CodeObjectStream {
+            code,
+            max_stack: 0,
+            current_stack_size: 0,
+        }
     }
 }
 impl From<CodeObjectStream> for CodeObject {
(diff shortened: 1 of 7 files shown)


cthulahoops:
This seems like it's going to add a bunch of code, for a possible speed up. Is it justified?

youknowone:
@cthulahoops Looking for this one?

vm/src/frame.rs
```diff
@@ -85,7 +85,7 @@ pub struct Frame {
     stack: RefCell<Vec<PyObjectRef>>, // The main data frame of the stack machine
     blocks: RefCell<Vec<Block>>,      // Block frames, for controlling loops and exceptions
     pub scope: Scope,                 // Variables
-    pub lasti: RefCell<usize>,        // index of last instruction ran
+    lasti: RefCell<usize>,            // index of last instruction ran
```
coolreader18: If you're going to touch this, do you want to make it a `Cell<usize>`? It really doesn't need to be a `RefCell`

vm/src/frame.rs
```diff
@@ -1244,6 +1261,7 @@ impl Frame {
 
     pub fn push_value(&self, obj: PyObjectRef) {
         self.stack.borrow_mut().push(obj);
+        // println!("Stack size: {}, capa: {}", self.stack.borrow().len(), self.stack.borrow().capacity());
```
coolreader18:
```suggestion
```

compiler/src/output_stream.rs
```diff
@@ -26,13 +34,21 @@ impl From<CodeObjectStream> for CodeObject {
 
 impl OutputStream for CodeObjectStream {
     fn emit(&mut self, instruction: Instruction, location: Location) {
+        let effect = stack_effect(&instruction);
+        self.current_stack_size += effect;
+        if self.current_stack_size > self.max_stack {
+            self.max_stack = self.current_stack_size;
+        }
```
coolreader18: You could get rid of the properties on `CodeObjectStream` and just mutate `max_stack_size` on `CodeObject`:

```rust
let max_stack_size = &mut self.code.max_stack_size;
*max_stack_size = std::cmp::max(*max_stack_size, *max_stack_size + effect);
```

windelbouwman: I understand. The thing is, I would like to gather all instructions in the stream, and in the end, create the immutable codeobject. This makes the code easier to understand in my opinion. That is the reason why I implemented the code like it is.

vm/src/frame.rs
windelbouwman: Agreed!

windelbouwman:
> This seems like it's going to add a bunch of code, for a possible speed up. Is it justified?

This change is not justified, but it is a copy of an idea from cpython and micropython.

compiler/src/stack_effect.rs
```diff
@@ -0,0 +1,88 @@
...
+        JumpIfTrueOrPop { .. } => -1,
+        Break => 0,
+        Continue => 0,
+        PopBlock => 0,
+        GetIter => 0,
+        ForIter { .. } => 1,
+        CallFunction { typ } => match typ {
+            CallType::Positional(amount) => -(*amount as isize),
+            CallType::Keyword(amount) => -1 - (*amount as isize),
+            CallType::Ex(has_kwargs) => {
```
whjpji, line 29: How about this?
```suggestion
            CallType::Ex(has_kwargs @ true) => -2,
            CallType::Ex(has_kwargs @ false) => -1,
```

PR #1328, opened 2019-08-29
Fix symtable scan_expression not to mark sequence as Load 

youknowone:


compiler/src/symboltable.rs
```diff
@@ -560,9 +560,17 @@ impl SymbolTableBuilder {
             }
             Bytes { .. } => {}
             Tuple { elements } | Set { elements } | List { elements } | Slice { elements } => {
-                self.scan_expressions(elements, &ExpressionContext::Load)?;
+                self.scan_expressions(elements, context)?;
             }
             Comprehension { kind, generators } => {
+                for generator in generators {
```
windelbouwman: Was there a bug that this block is moved? Btw, generators are implemented as functions, so it might make sense to define a new symbol table for generators. We could check with the crawl_source.py script on the test snippet you added here.

windelbouwman (approved):

compiler/src/symboltable.rs
youknowone: Reverted it because it didn't afffect the issue

youknowone:
@windelbouwman the crawl result is not very same https://gist.github.com/youknowone/def929a322d21535c9cf99117072f61a

windelbouwman:
> @windelbouwman the crawl result is not very same https://gist.github.com/youknowone/def929a322d21535c9cf99117072f61a
> 
> Child table here? https://gist.github.com/youknowone/def929a322d21535c9cf99117072f61a#file-out-diff-L309

We need to create symboltables for list/set/tuple/dict comprehensions as well. Comprehensions are implemented as functions, so they have their own symbol table.

The other issue is that the ast module does not create proper nodes for the elements in the comprehensions. This must be added as well.

Do you want to implement this change in this pull request? It can be quite useful!

windelbouwman:
@youknowone I implemented the comprehension scopes for generators, see #1336

PR #1330, opened 2019-08-31
Delay from_list tuple create in import

palaviv:
This change decreases the time when importing a module which is already in `sys.modules`. If you will run the following script:
```py
import time
start = time.time()
import io
mid = time.time()
import io
end = time.time()
print(mid - start)
print(end - mid)
```
Before the change you will get:
```
0.0029745101928710938
0.00000667572021484375
```
After the change:
```
0.002993345260620117
0.000006198883056640625
```

```
```


windelbouwman:
This looks good to me, tiny speed enhancement, but also a code simplification.

PR #1332, opened 2019-08-31
load_global shouldn't load outermost locals.

cthulahoops:
This leapt out at me while reviewing #1329. The code was was added recently.


cthulahoops:
Failing due #1333

PR #1339, opened 2019-09-04
Add IntoPyBool

palaviv:
The purpose of the change is to differentiate between function that accept `Boolean` arguments (`True`, `False`) and function that accept "bool like" arguments that can be converted to bool by using `__bool__` or `__len__`. For example the builtin function `all` can get any "bool like" objects in its iterator for example `all([True,[1] ])` while doing `[1,2,3].sort(reverse=[1])` will result in a `TypeError`.

If anyone has a better suggestion then `PyBoolLike` for the struct name I would love that as I could not think of a better name...

### DIFF
--- a/vm/src/obj/objbool.rs
+++ b/vm/src/obj/objbool.rs
@@ -202,13 +202,8 @@ pub struct IntoPyBool {
 }
 
 impl IntoPyBool {
-    pub fn get_false() -> Self {
-        IntoPyBool { value: false }
-    }
-
-    pub fn get_true() -> Self {
-        IntoPyBool { value: true }
-    }
+    pub const TRUE: IntoPyBool = IntoPyBool { value: true };
+    pub const FALSE: IntoPyBool = IntoPyBool { value: false };
 
     pub fn to_bool(self) -> bool {
         self.value
--- a/vm/src/obj/objbool.rs
+++ b/vm/src/obj/objbool.rs
@@ -202,6 +202,14 @@ pub struct PyBoolLike {
 }
 
 impl PyBoolLike {
+    pub fn get_false() -> Self {
+        PyBoolLike { value: false }
+    }
+
+    pub fn get_true() -> Self {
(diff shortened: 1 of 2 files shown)


coolreader18:
Maybe `PyBoolish`?

palaviv:
`PyBoolable`?

coolreader18:
Or maybe something that emphasizes that it's really just a wrapper around a bool rather than a PyObjectRef? Something simple like `PyBoolean`?

windelbouwman:
How about `IntoPyBool`?

palaviv:
> How about `IntoPyBool`?

I like this option the best

vm/src/obj/objbool.rs
```diff
@@ -18,7 +18,12 @@ impl IntoPyObject for bool {
 
 impl TryFromObject for bool {
     fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<bool> {
-        boolval(vm, obj)
+        if objtype::isinstance(&obj, &vm.ctx.int_type()) {
+            Ok(get_value(&obj))
+        } else {
+            let actual_type = vm.to_pystr(&obj.class())?;
```
coolreader18: You should be able to get `obj.class().name`, and avoid the possibility of an error

vm/src/obj/objbool.rs
```diff
@@ -199,3 +195,30 @@ fn bool_new(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
 }
+
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub struct IntoPyBool {
+    value: bool,
+}
+
+impl IntoPyBool {
+    pub fn get_false() -> Self {
+        IntoPyBool { value: false }
```
coolreader18: Instead of `get_{true,false}`, how about constants?
```rust
const TRUE: IntoPyBool = IntoPyBool { value: true };
```

PR #1341, opened 2019-09-05
Fix class vs method scopes

coolreader18:


youknowone:
Related to #1302 ?

bytecode/src/bytecode.rs
```diff
@@ -53,6 +53,7 @@ bitflags! {
         const HAS_DEFAULTS = 0x01;
         const HAS_KW_ONLY_DEFAULTS = 0x02;
         const HAS_ANNOTATIONS = 0x04;
+        const IS_CLASS = 0x08;
```
windelbouwman: Which flag in cpython does this correspond to?

See this list of flags:

https://github.com/python/cpython/blob/4a2edc34a405150d0b23ecfdcb401e7cf59f4650/Include/code.h#L71

It would be nice to keep this more or less similar.

youknowone:
I don't know well about scoping, but basically I think vm is not related with this problem.
Instead of it, I think resoving scope hierachy in compiler will solve it. Please check this idea fits for Python scoping.


```python
class A:  # parent(module) scope A
    a = 10  # child attr scope B 
    b = 20  # child attr scope B
    def f(self):  # child attr scope B
        x = 10   # function f scope C
        y = 10   # function f scope C
    def g(cls):  # child attr scope B
        z = 20  # function g scope D
        w = 20  # function g scope D
```

In this example, there are 4 scopes A, B, C and D.

For now, C and D seems they are having B as their parent like:
```
A (module)
ㄴ B (attributes of A)
  ㄴ C (A.f)
  ㄴ D (A.g)
```
which is 3-level.

I think B and (C or D) are same leveled scope. Because there is no way to access to anything in B from C or D. It always requires self, cls or A.

```
A (module)
ㄴ B (attributes of A)
ㄴ C (A.f)
ㄴ D (A.g)
````

So, will the problem be fixed by just moving C and D under A directly?

File: bytecode/src/bytecode.rs (hunk 1 above)
Comment (coolreader18) [reply bytecode/src/bytecode.rs]: I think it's somewhat equivalent to `CO_NEWLOCALS`, I'll look into it.

Comment (windelbouwman) [reply bytecode/src/bytecode.rs]: I'm still unaware how this issue is implemented in cpython

**Comment by cthulahoops** [comment]
> I don't know well about scoping, but basically I think vm is not related with this problem.
> Instead of it, I think resoving scope hierachy in compiler will solve it. Please check this idea fits for 
> Python scoping.

I don't think it can't be done by the compiler, because the compiler doesn't necessarily know the context the code will be executed in.

Consider:

```
c = compile("def f(self): assert x == 4", "", "exec")

x = 4
class A:
	x = 5
	exec(c)

A().f()
```

> So, will the problem be fixed by just moving C and D under A directly?

Agreed, but it does have to happen at function definition time, not at compile time.

File: vm/src/builtins.rs [hunk 2]
```diff
@@ -933,7 +935,20 @@ pub fn builtin_build_class_(
...
-    vm.invoke_with_locals(&function, cells.clone(), namespace.clone())?;
+    let PyFunction { code, scope, .. } = &*function;
+
+    let is_class = scope.is_class();
+
+    let mut scope = scope
+        .new_child_scope_with_locals(cells.clone())
+        .new_child_scope_with_locals(namespace.clone());
+
+    if is_class {
```
cthulahoops: I'm surprised by this ```if``` - aren't we always defining a class? (Or have I just not worked through this in enough detail.) I'd expect that we'd always mark the scope as a class here, and check it in function def. Do we need the compiler change?

coolreader18: I did this because you can just call `builtins.__build_class__` with a function of your own, and then it doesn't change the scope:

```python
>>> a = 1
>>> def f():
...   a = 2
...   def b(): print(a)
...   b()
... 
>>> __build_class__(f, "A")
2
<class '__main__.A'>
```

cthulahoops: Ouch. Python!

In that case can the magic go in the LoadBuildClass instruction?

ammaraskar:
Just chiming in from the Python discourse thread @coolreader18 posted: https://discuss.python.org/t/how-does-scoping-for-class-building-vs-methods-work/2272/2

As far I know, this behavior is entirely implemented within the compiler. I don't think any special run-time stuff is necessary for it. As @youknowone eluded to, this is a scoping issue and the correct way to scope this is as:
```
A (class)
ㄴ B (attributes of A)
ㄴ C (A.f)
ㄴ D (A.g)
```
As per https://docs.python.org/3/reference/executionmodel.html#resolution-of-names, "The scope of names defined in a class block is limited to the class block; it does not extend to the code blocks of methods". Thus, scopes C and D should not inherit from scope B.

----

As far as the example @cthulahoops posted
```python
c = compile("def f(self): assert x == 4", "", "exec")
x = 4
class A:
	x = 5
	exec(c)
A().f()
```
This is no different than the following:
```python
>>> def f():
...   return a + 1
...
>>> f()
  File "<stdin>", line 2, in f
NameError: name 'a' is not defined
>>> a = 1
>>> f()
2
``` 
unbound variables used in functions should be looked up in the globals anyway.

----

Finally, the key difference in

```python
>>> a = 1
>>> def f():
...   a = 2
...   def b(): print(a)
...   b()
... 
>>> __build_class__(f, "A")
2
<class '__main__.A'>
```
is the lack of the `class A` declaration, this is actually what changes the scoping rules.

cthulahoops:
> unbound variables used in functions should be looked up in the globals anyway.

Thanks! That's the bit I was needed. It does lead to really weird behaviours, like unevaluated expressions changing the scope another variable refers to.

```
c = compile("x", "", "eval")
x = "global"
def main():
    x = "main"
    def f():
        False and x
        assert "main" == eval(c)
    def g():
        assert "global" == eval(c)
    f()
    g()
main()
```

ammaraskar:
> like unevaluated expressions changing the scope another variable refers to.

Yup! Luckily code like this is relatively rare, though it is easy to come up with adversarial examples.

---

Also, for anyone else, that one is another head-scratcher. What happens is that when the symbol table goes over `f()`, it sees the used `x` and since it has an `x = "main"` from the nearest enclosing scope, it generates a closure to pass it into the function.

When it goes over `g()`, there is no usage of `x`, thus it never gets turned into a local or passed through a closure. When `eval(c)` runs, the `LOAD_NAME x` bytecode does a local lookup, which fails, and then it looks it up in the globals and finds `x = "global"`

coolreader18:
Thanks @ammaraskar, I think I finally got it. The key was changing this:
```rust
let found_in_outer_scope = self.tables.iter().skip(1).any(|t| {
    t.symbols.contains_key(&symbol.name)
});
```
to this:
```rust
let found_in_outer_scope = self.tables.iter().skip(1).any(|t| {
    t.typ != SymbolTableType::Class && t.symbols.contains_key(&symbol.name)
});
```
It looks to check if it's used in any outer scope, and if it is, it's established as free, but now it makes sure that it doesn't take it into account if it's a class scope.

tests/snippets/class.py
```diff
@@ -164,3 +164,25 @@ class T5(int):
...
+a = 1
+class A:
+    a = 2
+    def b():
+        assert a == 1
+    b()
+    assert a == 2
+A.b()
+
+# TODO: uncomment once free vars/cells are working
```
cthulahoops: Worth a comment about how/why this is currently broken?  Without testing, I'm guessing the symbol table spots the ```a = 2```, and marks a as being nonlocal, but then the load_name still walks the incorrect scope list and gets ```a = 3```.

coolreader18: Yep, exactly. I'll add a comment.

PR #1343, opened 2019-09-05
Add framework attribute in sys module

### LINKED ISSUE #1342: module 'sys' has no attribute '_framework' (mac)
## Feature
스크린샷 2019-09-05 오후 8 58 13
 
```
RustPython$ RUSTPYTHONPATH=`pwd`/Libs ./whats_left.sh

...

  File "/Users/vazrupe/workspace/RustPython/Lib/site.py", line 570, in <module>
    if not sys.flags.no_site:
  File "/Users/vazrupe/workspace/RustPython/Lib/site.py", line 556, in main
    ENABLE_USER_SITE = check_enableusersite()
  File "/Users/vazrupe/workspace/RustPython/Lib/site.py", line 312, in addusersitepackages
    user_site = getusersitepackages()
  File "/Users/vazrupe/workspace/RustPython/Lib/site.py", line 297, in getusersitepackages
    userbase = getuserbase() # this will also set USER_BASE
  File "/Users/vazrupe/workspace/RustPython/Lib/site.py", line 286, in getuserbase
    USER_BASE = _getuserbase()
  File "/Users/vazrupe/workspace/RustPython/Lib/site.py", line 257, in _getuserbase
    if sys.platform == "darwin" and sys._framework:
AttributeError: module 'sys' has no attribute '_framework'
```

RustPython and system Python have the following attributes:
```
RustPython$ cargo run -- -c 'import sys; print(sys._framework)'
AttributeError: module 'sys' has no attribute '_framework'
RustPython$ python3 --version
Python 3.7.4
RustPython$ python3 -c 'import sys; print(sys._framework)'
Python
```

## Python Documentation

This attribute represents the name of [the framework bundle](https://en.wikipedia.org/wiki/Bundle_(macOS)#macOS_framework_bundles) on mac.

[Here](https://docs.python.org/3/using/mac.html#getting-and-installing-macpython) you can see that the framework name of the Python binary on mac is `Python`.

Instructions on how to replace this attribute with a different value can be found in the [README in python/cpython](https://github.com/python/cpython#build-instructions). I could change to the following command:
```
$ git clone https://github.com/python/cpython
$ cd cpython
cpython$ ./configure --with-framework-name=example --enable-framework
cpython$ make frameworkinstall
cpython$ ls /Library/Frameworks/example.framework/Versions/Current/bin/
2to3			pip3			python3-config
2to3-3.9		pip3.9			python3.9
easy_install-3.9	pydoc3			python3.9-config
idle3			pydoc3.9
idle3.9			python3
cpython$ /Library/Frameworks/example.framework/Versions/Current/bin/python3
Python 3.9.0a0 (heads/master:29bb227a0c, Sep  5 2019, 23:11:17)
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
>>> import sys
>>> sys._framework
'example'
```

vazrupe:
Fixes #1342

### DIFF
--- a/tests/snippets/sysmod.py
+++ b/tests/snippets/sysmod.py
@@ -6,6 +6,9 @@ assert sys.argv[0].endswith('.py')
 
 assert sys.platform == "linux" or sys.platform == "darwin" or sys.platform == "win32" or sys.platform == "unknown"
 
+if hasattr(sys, "_framework"):
+    assert type(sys._framework) is str
+
 assert isinstance(sys.builtin_module_names, tuple)
 assert 'sys' in sys.builtin_module_names
 
--- a/vm/src/sysmodule.rs
+++ b/vm/src/sysmodule.rs
@@ -225,6 +225,8 @@ pub fn make_module(vm: &VirtualMachine, module: PyObjectRef, builtins: PyObjectR
         "unknown".to_string()
     };
 
+    let framework = "".to_string();
+
     // https://doc.rust-lang.org/reference/conditional-compilation.html#target_endian
     let bytorder = if cfg!(target_endian = "little") {
         "little".to_string()
@@ -347,6 +349,7 @@ settrace() -- set the global debug tracing function
       "modules" => modules.clone(),
       "warnoptions" => ctx.new_list(vec![]),
       "platform" => ctx.new_str(platform),
+      "_framework" => ctx.new_str(framework),
       "meta_path" => ctx.new_list(vec![]),
       "path_hooks" => ctx.new_list(vec![]),
       "path_importer_cache" => ctx.new_dict(),


vm/src/sysmodule.rs
```diff
@@ -225,6 +225,12 @@ pub fn make_module(vm: &VirtualMachine, module: PyObjectRef, builtins: PyObjectR
         "unknown".to_string()
     };
 
+    let framework = if cfg!(target_of = "macos") {
```
coolreader18:
```suggestion
    let framework = if cfg!(target_os = "macos") {
```

youknowone:
Thanks for the contribution.

Here are a few concerns about this patch (and the issue too).

- It seems it is not defined in document. https://docs.python.org/3/library/sys.html Please share if you know any use case of this attribute.
- It seems it must refers actual python framework it refers. My pyenv python build in macos have just empty string
```
➜  ~ python
Python 3.7.0 (default, Aug 24 2018, 13:00:59)
[Clang 9.1.0 (clang-902.0.39.2)] on darwin
>>> import sys
>>> sys.framework
AttributeError: module 'sys' has no attribute 'framework'
>>> sys._framework
''
```
- Test code is required
- This patch defines the attribute for all platforms. It seems it is not defiend out of macOS
```
➜  ~ python
Python 3.6.3 (default, Oct 12 2017, 21:12:06)
[GCC 5.4.0 20160609] on linux
p>>> import sys
>>> sys._framework
AttributeError: module 'sys' has no attribute '_framework'
>>>
```

vazrupe:
@youknowone 

Thank you for comment.

Currently this code only affects the `site` module. Where to use:

https://github.com/RustPython/RustPython/blob/1dceae9205ae3385cdafe368d8ef43bd753f2a34/Lib/site.py#L257-L259
https://github.com/RustPython/RustPython/blob/1dceae9205ae3385cdafe368d8ef43bd753f2a34/Lib/site.py#L271-L272

CPython also uses the [`sysconfig` module](https://github.com/python/cpython/blob/daf62627518ad97ce66a48c49496aa0573cf0731/Lib/sysconfig.py#L202-L204). This is the same code `as _getuserbase()` in the `site` module.
```
    if sys.platform == "darwin" and sys._framework:
        return joinuser("~", "Library", sys._framework,
                        "%d.%d" % sys.version_info[:2])
```

`site.getuserbase()` returns the [user base directory](https://docs.python.org/3/library/site.html#site.getuserbase).

This function works as follows on my system:
```
$ python3     # is system python
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
>>> import site
>>> site.getuserbase()
'/Users/vazrupe/Library/Python/3.7'
>>> import sys
>>> sys._framework
'Python'

$ pyenv shell 3.7.0
$ python
Python 3.7.0 (default, Sep  7 2019, 14:41:54)
[Clang 10.0.1 (clang-1001.0.46.4)] on darwin
>>> import site
>>> site.getuserbase()
'/Users/vazrupe/.local'
>>> import sys
>>> sys._framework
''
```

As you commented, this value seems to be an empty string. In such a case, the following value will always be returned.

And this attribute is defined on all platforms. However, it is only implemented in 3.7+ and up, so it cannot be used in lower versions.

For example, in my windows(anaconda 3.7.3) environment it would run:
```
PS> python
Python 3.7.3 (default, Apr 24 2019, 15:29:51) [MSC v.1915 64 bit (AMD64)] :: Anaconda, Inc. on win32

>>> import sys
>>> sys._framework
''
```

In a Mac(use pyenv 3.6.6) environment, it runs like this:
```
$ pyenv shell 3.6.6
$ python
Python 3.6.6 (default, Sep  7 2019, 14:35:29)
[GCC 4.2.1 Compatible Apple LLVM 10.0.1 (clang-1001.0.46.4)] on darwin
>>> import sys
>>> sys._framework
AttributeError: module 'sys' has no attribute '_framework'
```

Thanks for reading!

youknowone:
Thanks for detailed explanation. I am sorry to confuse it.

It seems CI fails because RustPython CI now tests only with CPython 3.6.
Instead of testing private attributes, how about testing site.py feature which uses `_framework` attribute?

vazrupe:
I apologize for writing the wrong test case first. For RustPython developers using mac, this test may fail even with Python 3.7+. Actually this test fails on my Mac.

This is because the system python on Mac is built with `Python.framework`. So test code should look like this:
```
assert sys._framework == "" or sys._framework == "Python"
```

I thought about testing in the `site` module, but I didn't think of a good way. 😢 Also, the method does not seem to be suitable for solving this problem.

There are a few things to think about for this test.

- We expect RustPython's `sys._framework` to be built with an empty string.
- If running on system python on Mac, `sys._framework` can be 'Python'.
- Since this attribute is implemented in python 3.7+, an earlier version will throw an AttributeError.

The best code I can think of is this:
```
try:
    assert sys._framework == ""
except AttributeError:
    assert "%d.%d" % sys.version_info[:2] < "3.7"
except AssertionError:
    assert sys.platform == "darwin" and sys._framework == "Python"
```

What do you think about this solution?

youknowone:
now, actually, the code itself looks totally fine. but not sure the test code is fit for the goal of snippet tests.
@coolreader18 how do you think about the test code?

tests/snippets/sysmod.py
```diff
@@ -6,6 +6,13 @@
 
 assert sys.platform == "linux" or sys.platform == "darwin" or sys.platform == "win32" or sys.platform == "unknown"
 
+try:
+    assert sys._framework == ""
```
coolreader18: Maybe something that just checks that if there is the `_framework` attr, that it is a `str`?

```py
if hasattr(sys, "_framework"):
    assert type(sys._framework) is str
```

PR #1347, opened 2019-09-06
Implement getenv function from os module

dvtkrlbs:
https://docs.python.org/3/library/os.html#os.getenv

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1041,7 +1041,9 @@ pub struct GetEnvArgs {
 
 pub fn os_getenv(args: GetEnvArgs, vm: &VirtualMachine) -> PyResult {
     // TODO: Change "_os" when _os module is changed to os
-    let os = vm.get_attribute(vm.sys_module.clone(), "modules")?.get_item("_os", &vm)?;
+    let os = vm
+        .get_attribute(vm.sys_module.clone(), "modules")?
+        .get_item("_os", &vm)?;
     let environ = vm.get_attribute(os, "environ")?;
     let env_var = environ.get_item(&args.key.value, &vm);
     match env_var {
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/os.rs
```diff
@@ -1030,6 +1030,26 @@ pub fn os_ttyname(fd: PyIntRef, vm: &VirtualMachine) -> PyResult {
...
+pub struct GetEnvArgs {
+    #[pyarg(positional_or_keyword, optional = false)]
+    key: PyStringRef,
+    #[pyarg(keyword_only, default = "None")]
+    default: Option<PyObjectRef>,
+}
+
+pub fn os_getenv(args: GetEnvArgs, vm: &VirtualMachine) -> PyObjectRef {
+    use std::env;
+    match env::var(&args.key.value) {
```
isidentical: i think you should use `os.environ` instead

dvtkrlbs:
Do we need to get these values from os.environ

isidentical:
> Do we need to get these values from os.environ

I believe we should (for cpython compability)
```py
>>> import os
>>> assert os.getenv("A") is None
>>> os.environ["A"] = "B"
>>> assert os.getenv("A") == "B"
>>> os.environ = type("Fake", (), {"get": lambda self, key, default: "blabla"})()
>>> os.getenv("y")
'blabla'
>>> os.getenv("yz")
'blabla'
>>> os.getenv("yzz")
'blabla'
```

dvtkrlbs:
Should be the final version getenv should use os.environ so i fetch os module from sys.modules and then call environ and get the appropriate key. (cpython writes this module in python so their code is more clean)

youknowone:
We have the function in same place: https://github.com/RustPython/RustPython/blob/master/Lib/os.py#L131

Do we need it in the rust module? Reimplementing python stdlibs in rust module sometimes have advantages. But in this case, due to the dynamic fetchings, it seems it doesn't have much advantage.

isidentical:
> Do we need it in the rust module?
@youknowone i dont think we do. CPython also implements `getenv` in python level

dvtkrlbs:
Yes that is right i am not very familiar with the codebase so i completely missed that

PR #1354, opened 2019-09-08
Improve __bool__

ChJR:
issue founded by @youknowone


vm/src/obj/objbool.rs
```diff
@@ -29,22 +30,30 @@ pub fn boolval(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<bool> {
...
         }
         None => match vm.get_method(obj.clone(), "__len__") {
             Some(method_or_err) => {
                 let method = method_or_err?;
                 let bool_obj = vm.invoke(&method, PyFuncArgs::default())?;
                 match bool_obj.payload::<PyInt>() {
-                    Some(int_obj) => !int_obj.as_bigint().is_zero(),
+                    Some(int_obj) => {
+                        let len_val = int_obj.as_bigint();
+                        if len_val.sign() == Sign::Minus {
```
youknowone: Do we prefer to have if-else block than this pattern?

```rust
if len_val.sign() == Sign::Minus {
   Err(
      vm.new_value_error("__len__() should return >= 0".to_string())
    )
} else {
    !len_val.is_zero()
}
```

coolreader18: It would still have to `return` the error, as the result of that expression is a bool and not a Result.

youknowone: Oh, it must escape from the function, i got it.

PR #1355, opened 2019-09-08
Fix global declaration error in the function

### LINKED ISSUE #1353: When call global twice for the same variable, SyntaxError occurs.
## Feature

```python
a = []
def f(b):
    global a
    global a  # Generate error only in RustPython
    a.append(b)
    print(a)

f(3)
print(a)
```
The code above generates the following error:
(when call global twice for the same variable, an error occurs.)
### Actual result
```bash
Hyeockz:RustPython hyeockjinkim$ cargo run e.py 
    Finished dev [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/rustpython e.py`
SyntaxError: name 'a' is used prior to global declaration at line 0 column 0
```
### Expected result
```bash
[3]
[3]
```

HyeockJinKim:
Fix Global symbol to prevent syntax error
when global declaration for same variable in function.

Fixes #1353

### DIFF
--- a/compiler/src/symboltable.rs
+++ b/compiler/src/symboltable.rs
@@ -746,10 +746,15 @@ impl SymbolTableBuilder {
             // Role already set..
             match role {
                 SymbolUsage::Global => {
-                    return Err(SymbolTableError {
-                        error: format!("name '{}' is used prior to global declaration", name),
-                        location,
-                    })
+                    let symbol = table.symbols.get(name).unwrap();
+                    if let SymbolScope::Global = symbol.scope {
+                        // Ok
+                    } else {
+                        return Err(SymbolTableError {
+                            error: format!("name '{}' is used prior to global declaration", name),
+                            location,
+                        });
+                    }
                 }
                 SymbolUsage::Nonlocal => {
                     return Err(SymbolTableError {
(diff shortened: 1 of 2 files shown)


HyeockJinKim:
Can I get a review?

compiler/src/symboltable.rs
```diff
@@ -747,10 +747,13 @@ impl SymbolTableBuilder {
             // Role already set..
             match role {
                 SymbolUsage::Global => {
-                    return Err(SymbolTableError {
-                        error: format!("name '{}' is used prior to global declaration", name),
-                        location,
-                    })
+                    let symbol = table.symbols.get(name).unwrap();
+                    if symbol.is_assigned {
```
windelbouwman: How about this case:
```python
def foo():
    print(x)
    global x
```

In this case `x` is not assigned, but global is not allowed anyways.

Instead of `symbol.is_assigned`, check if `symbol.scope` is set to `Global` before.

HyeockJinKim: I think it is important that the variable is used or referenced in same scope rather than global declaration of variable, so I used `symbol.is_assigned` and `symbol.is_referenced`. Is my solution wrong?

windelbouwman: I think your solution is wrong, and the only legal re-declaration of a global is when the variable was previously marked as global, so in that case, check for `symbol.is_global`, this should be true, otherwise we face an error.

HyeockJinKim: Can I get a review again?

Issue #1230, opened 2019-08-11
__new__ should be a static method


Lynskylate:
In RustPython, \_\_new\_\_ is bound method but it should be a static method.

```python
class A:
    def __new__(*args, **kwargs):
        print("Args Total: {}".format(len(args)+len(kwargs)))

class B(A):
    def __new__(cls, *args, **kwargs):
        return super().__new__(cls, *args, **kwargs)
"""
>>> B()
CPython output
Args Total: 1 
RustPython output
Args Total: 2
"""
```
 super().__new__(cls, *args, **kwargs) will pass two same ype class(args[0] is same as cls).

[object \_\_new\_\_ Python docuemtion](https://docs.python.org/3/reference/datamodel.html#basic-customization)

cthulahoops:
I think this is probably a duplicate of #778 . Just trying to get this stuff fresh in my mind so I can review the work here.

cthulahoops:
I've been considering this a problem with ```super``` as it works in the examples without ```super``` but finally got a simple example that shows the difference without using ```super```:

CPython:

```
>>> (1).__new__(int, 5)
5
```

RustPython:

```
>>>>> (1).__new__(int, 5)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Expected type <class 'type'>, not <class 'int'>
```

Lynskylate:
I try to fix RustPython/RustPython#1241, but I met some trouble. I simplified the code in enum.py that caused the error. 
```python
class EnumMeta(type):
    def __new__(metacls, cls, bases, classdict):
        enum_class = super().__new__(metacls, cls, bases, classdict)
        return enum_class

class Test(metaclass=EnumMeta):
    def __init__(self):
        pass
"""
>>> t = Test() 
# RustPython Output:
Traceback (most recent call last):
  File "/tmp/reproduce.py", line 17, in <module>
    class Test(metaclass=EnumMeta):
  File "/tmp/reproduce.py", line 14, in __new__
    enum_class = super().__new__(metacls, cls, bases, classdict)
TypeError: type() takes 1 or 3 arguments,
"""
```

Lynskylate:
Incompact example code for RustPython object's\ _\_new\_\_.
```python
type(object.__new__)
type(object().__new__)
```
RustPython:
```
<class 'builtin_function_or_method'>
<class 'method'>
```
CPython
```
<class 'builtin_function_or_method'>
<class 'builtin_function_or_method'>
```

PR #1357, opened 2019-09-09
Implement __init_subclass__

dvtkrlbs:
This pull request implements a portion of the [PEP 487](https://www.python.org/dev/peps/pep-0487/) this pep also defines  `__set_name__` dunder method but it is not implemented in this pr.

`__init_subclass__` method allows a superclass to initialize al subclasses it takes the subclass and any number of arguments

### DIFF
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -17,7 +17,7 @@ use crate::obj::objdict::PyDictRef;
 use crate::obj::objint::{self, PyIntRef};
 use crate::obj::objiter;
 use crate::obj::objstr::{PyString, PyStringRef};
-use crate::obj::objtype::{self, PyClassRef};
+use crate::obj::objtype::{self, PyClass, PyClassRef};
 #[cfg(feature = "rustpython-compiler")]
 use rustpython_compiler::compile;
 
@@ -944,5 +944,13 @@ pub fn builtin_build_class_(
         vec![name_obj, bases, namespace.into_object()],
     )?;
     cells.set_item("__class__", class.clone(), vm)?;
+
+    for base_class in class.clone().downcast::<PyClass>().unwrap().mro.clone() {
+        if let Ok(initter) = vm.get_attribute(base_class.into_object(), "__init_subclass__") {
+            let args: Args = Args::new(vec![class.clone()]);
+            vm.invoke(&initter, PyFuncArgs::from((&args, &kwargs)))?;
+            break;
+        }
+    }
     Ok(class)
 }
--- a/vm/src/function.rs
+++ b/vm/src/function.rs
@@ -290,6 +290,9 @@ impl<T> Args<T> {
     pub fn into_vec(self) -> Vec<T> {
         self.0
     }
+    pub fn new(args: Vec<T>) -> Self {
+        Self(args)
+    }
 }
 
 impl<T: PyValue> Args<PyRef<T>> {


youknowone:
Thanks for your contribution. Adding a test about this feature In `tests/snippets` will be helpful.

isidentical:
IMHO doing this calls on `builtins.__build_class__` doesn't cover all cases.
```console
(general) batuhan@x200-trisquel:~/cpython$ python
Python 3.9.0a0 (heads/master:b607d99, Jul  4 2019, 11:03:12) 
[GCC 5.4.0 20160609] on linux
>>> class X:           
...     def __init_subclass__(cls):     
...         print(f"initalizing {cls!r}")
... 
>>> Y = type("Y", (X,), {})
initalizing <class '__main__.Y'>
```

```console
(general) batuhan@x200-trisquel:~/RustPython$ git checkout dvt/init_subclasses 
(general) batuhan@x200-trisquel:~/RustPython$ cargo run
   Compiling rustpython-vm v0.1.0 (/home/batuhan/RustPython/vm)
   Compiling rustpython v0.1.0 (/home/batuhan/RustPython)
    Finished dev [unoptimized + debuginfo] target(s) in 55.62s
     Running `target/debug/rustpython`
Welcome to the magnificent Rust Python 0.1.0 interpreter 😱 🖖
>>>>> class X:
.....     def __init_subclass__(cls):
.....         print(f"initalizing {cls!r}")
..... 
>>>>> Y = type("Y", (X,), {})
>>>>> 
```

CPython does this notify operation on type object constructor (`__new__`). The reason 
@dvtkrlbs used `__build_class__` instead of `__new__` is unlike cpython rustpython's `__build_class__` doesn't pass it's keyword arguments (`class A(kw1=some_value, ..., metaclass=Z)`) to type constructor.

Also there is another thing that might be a problem with cpython compatibility. 
```c
    /* Special-case __init_subclass__ and __class_getitem__:
       if they are plain functions, make them classmethods */
    tmp = _PyDict_GetItemIdWithError(dict, &PyId___init_subclass__);
    if (tmp != NULL && PyFunction_Check(tmp)) {
        tmp = PyClassMethod_New(tmp);
        if (tmp == NULL)
            goto error;
        if (_PyDict_SetItemId(dict, &PyId___init_subclass__, tmp) < 0) {
            Py_DECREF(tmp);
            goto error;
        }
        Py_DECREF(tmp);
    }
    else if (tmp == NULL && PyErr_Occurred()) {
        goto error;
    }
```
CPython;
```py
>>> class X:
...     def __init_subclass__(cls):
...         pass
... 
>>> assert isinstance(X.__dict__["__init_subclass__"], classmethod)
>>> 
```
RustPython;
```py
>>>>> class X:
.....     def __init_subclass__(cls):
.....         pass
..... 
>>>>> assert isinstance(X.__dict__["__init_subclass__"], classmethod)
AssertionError
```
This is a special case like #1230

dvtkrlbs:
Also probably an error in my side but with this edit the parent class cant alter the subclass even though i pass the class as reference 
```py
class X:
    def __init_subclass__(cls):
        cls.initialized = True

class Y:
    pass

assert Y.initialized
```
this does not work even though it should work

isidentical:
> this does not work even though it should work
I am not familiar with rust (not even a little bit) so just asking. 
```rs
            let args: Args = Args::new(vec![class.clone()]);
```
is this `clone()` method related? In CPython
```c
    type = (PyTypeObject *)metatype->tp_alloc(metatype, nslots);
    ...
    if (init_subclass(type, kwds) < 0)
        goto error;    
...
`init_subclass` gets the real class object, not a clone.

dvtkrlbs:
Python objects in rustpython lives in Rc (reference counting) and they customize the behaviour. Clone just increments the counter. (Unlike normal objects)

isidentical:
> Python objects in rustpython lives in Rc (reference counting) and they customize the behaviour. Clone just increments the counter. (Unlike normal objects)

oh, i see.

dvtkrlbs:
Hmm you can access args from type_new function it actually passes them i will move the implementation to there

windelbouwman:
> this does not work even though it should work

This is probably caused because `__init_subclass__` is not a classmethod. The initialized attribute should be set on the class `X` in your example, and a lookup on `Y` will resort to using the attributes on `X`.

windelbouwman:
Could you add the python examples now mentioned as comments into the `tests/snippets` directory? Then we can run those snippets on cpython and rustpython to test for compatibility.

dralley:
@dvtkrlbs ?

dvtkrlbs:
Oh i completely forgot about this

PR #1359, opened 2019-09-10
Use the rustyline prompt on redox

coolreader18:


vm/src/stdlib/os.rs
```diff
@@ -1229,7 +1229,7 @@ fn extend_module_platform_specific(vm: &VirtualMachine, module: PyObjectRef) ->
         "openpty" => ctx.new_rustfunc(os_openpty),
     });
 
-    #[cfg(not(target_os = "macos"))]
+    #[cfg(not(any(target_os = "macos", target_os = "android", target_os = "redox")))]
```
youknowone: Is this a blacklist instead of a whitelist(maybe linux)?

coolreader18: It's actually [very specific](https://docs.rs/nix/0.15.0/src/nix/unistd.rs.html#962-987).

PR #1362, opened 2019-09-10
Only overwrite the locals in __build_class__ if it's a class

coolreader18:


compiler/src/compile.rs
```diff
@@ -975,7 +984,7 @@ impl<O: OutputStream> Compiler<O> {
 
         // Turn code object into function object:
         self.emit(Instruction::MakeFunction {
-            flags: bytecode::FunctionOpArg::empty(),
+            flags: bytecode::FunctionOpArg::NO_NEW_LOCALS,
```
windelbouwman: Instead of using the flag `NO_NEW_LOCALS`, it might be good to flag the inverse `NEW_LOCALS`. This is in line with CPython, plus it indicates the presence of new locals, instead of the absence of new locals. I find it usually prefer non-inverted booleans. This to prevent constructions like this:

```rust
if !no_new_locals {
  // double negation
  // this block is executed when we have new locals.
}
```

For example, I would turn this code:
```rust
if !x {
  // do a
} else {
  // do b
}
```

into this:

```rust
if x {
  // do b
} else {
  // do a
}

```

windelbouwman:
This looks good to me! I left a comment about the inverted boolean, but other than that, this change looks good to me!

compiler/src/compile.rs
coolreader18: I did this because it's a very specific special case where we don't want new locals, and every other time we emit a `MakeFunction` instruction we'd have to remember to include the `NEW_LOCALS` flag.

windelbouwman: Okay, I see. In this case, could be implement the `Default::default` trait to implement this sensible default on the bitflags?

PR #1364, opened 2019-09-11
nonlocal generate syntax error

HyeockJinKim:
If the scope depth is less than 2,
a syntax error occurs in nonlocal.


HyeockJinKim:
Can I get a review?

PR #1366, opened 2019-09-11
Fix unsupported operand type of str problem

### LINKED ISSUE #1365: unsupported operand type of str type make panic
## Feature
```python
a = ''
a['3']
```
unsupported operand type of str type make panic.
I think it should generate only TypeError, not panic.

### Expected Result
```bash
TypeError: string indices must be integers
```

### Actual Result
```bash
thread 'main' panicked at 'TypeError: indexing type "" with index [PyObj PyString { value: "3" }] is not supported (yet?)', vm/src/obj/objstr.rs:1575:9
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

HyeockJinKim:
Modify panic! to Err for unsupported operand type of str problem

Fixes #1365

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -1572,10 +1572,10 @@ pub fn subscript(vm: &VirtualMachine, value: &str, b: PyObjectRef) -> PyResult {
         let string = value.to_string().get_slice_items(vm, &b)?;
         Ok(vm.new_str(string))
     } else {
-        panic!(
-            "TypeError: indexing type {:?} with index {:?} is not supported (yet?)",
+        Err(vm.new_type_error(format!(
+            "indexing type {:?} with index {:?} is not supported",
             value, b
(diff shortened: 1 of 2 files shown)


HyeockJinKim:
Can I get a review?

vm/src/obj/objstr.rs
```diff
@@ -1572,10 +1572,10 @@ pub fn subscript(vm: &VirtualMachine, value: &str, b: PyObjectRef) -> PyResult {
         let string = value.to_string().get_slice_items(vm, &b)?;
         Ok(vm.new_str(string))
     } else {
-        panic!(
+        Err(vm.new_type_error(format!(
             "TypeError: indexing type {:?} with index {:?} is not supported (yet?)",
```
coolreader18: Maybe remove the "yet" here, as it's not really accurate anymore.

Issue #1361, opened 2019-09-10
Implement rest of itertools API


corona10:
## Feature
There are some APIs does not implement yet.
Let's implement them

## Python Documentation
https://docs.python.org/3/library/itertools.html

seeeturtle:
can i try it?

corona10:
@seeeturtle 
Sure go ahead.
But please leave a comment which API you will work with so that rest of APIs can be implemented by other Contributhon2019 participants.

corona10:
@coolreader18
 Can you assign this issue to @seeeturtle ? 
(There may be additional mentees who want to work with)

seeeturtle:
> Sure go ahead.
> But please leave a comment which API you will work with so that rest of APIs can be implemented by other Contributhon2019 participants.

sure, i'll try to implement `Combinatoric iterators`.

Thrimbda:
hello, can I try `itertools.groupby`?

coolreader18:
Yep, go ahead!

corona10:
I am working on `itertools.cycle` :)

pitachips:
I am working on `itertools.compress`

dralley:
What is "taken" and who it is assigned to.  I will try to keep this updated including what gets completed.

## Infinite Iterators

- [ ] ```itertools.count()```
- [x] ```itertools.cycle()``` @corona10 
- [x] ```itertools.repeat()``` @dralley

## Iterators terminating on the shortest input sequence

- [x] ```itertools.accumulate()``` @j30ng 
- [ ] ```itertools.chain()```
- [ ] ```itertools.chain.from_iterable()```
- [x] ```itertools.compress()``` @pitachips 
- [ ] ```itertools.dropwhile()```
- [ ] ```itertools.filterfalse()```
- [x] ```itertools.groupby()``` @Thrimbda 
- [ ] ```itertools.isslice()```
- [ ] ```itertools.starmap()```
- [x] ```itertools.takewhile()``` @dralley
- [x] ```itertools.tee()``` @j30ng 
- [ ] ```itertools.zip_longest()```

## Combinatoric iterators

- [x] ```itertools.product()``` @seeeturtle 
- [x] ```itertools.permutations()``` @seeeturtle 
- [x] ```itertools.combinations()``` @seeeturtle 
- [x] ```itertools.combinations_with_replacement()``` @seeeturtle

Space0726:
Hello, Can I try itertools.zip_longest()?

[...]

PR #1372, opened 2019-09-13
Implement itertools.accumulate

j30ng:
#1361 
This is an implementation of the Python standard library function `itertools.accumulate`. I also included a couple of very basic tests.

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -537,6 +537,66 @@ impl PyItertoolsFilterFalse {
     }
 }
 
+#[pyclass]
+#[derive(Debug)]
+struct PyItertoolsAccumulate {
+    iterable: PyObjectRef,
+    binop: PyObjectRef,
+    acc_value: RefCell<PyObjectRef>,
+}
+
+impl PyValue for PyItertoolsAccumulate {
+    fn class(vm: &VirtualMachine) -> PyClassRef {
+        vm.class("itertools", "accumulate")
+    }
+}
+
+#[pyimpl]
+impl PyItertoolsAccumulate {
+    #[pymethod(name = "__new__")]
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -537,6 +537,66 @@ impl PyItertoolsFilterFalse {
...
+        vm: &VirtualMachine,
+    ) -> PyResult {
+        let iter = get_iter(vm, &iterable)?;
+
+        Ok(PyItertoolsAccumulate {
+            iterable: iter,
+            binop: binop.unwrap_or_else(|| vm.get_none()),
+            acc_value: RefCell::from(vm.get_none()),
+        }
+        .into_ref(vm)
```
coolreader18: You should use `into_ref_with_type()` here and make the return type `PyResult<PyItertoolsAccumulate>`.

vm/src/stdlib/itertools.rs
```diff
@@ -537,6 +537,66 @@ impl PyItertoolsFilterFalse {
     }
 }
 
+#[pyclass]
+#[derive(Debug)]
+struct PyItertoolsAccumulate {
+    iterable: PyObjectRef,
+    binop: PyObjectRef,
+    acc_value: RefCell<PyObjectRef>,
```
coolreader18: Maybe have this as an `Option` rather than using `None` as "not yet started iterating", because the binop function could return `None`.

vm/src/stdlib/itertools.rs
j30ng: I see, thanks for pointing out!

vm/src/stdlib/itertools.rs
j30ng: Right, this wasn't carefully thought out.

PR #1380, opened 2019-09-15
os.scandir refinement

mpajkowski:
Added several changes to achieve better compatibility with CPython [[doc](https://docs.python.org/3/library/os.html#os.scandir)]:

* Make ```path``` an optional arg, set ```"."``` default argument,
* Implement missing methods:
    * context manager - ```__enter__```, ```__exit__```
    * ```close```.

### DIFF
--- a/tests/snippets/stdlib_os.py
+++ b/tests/snippets/stdlib_os.py
@@ -70,23 +70,16 @@ class TestWithTempDir():
             base_folder = "/tmp"
 
         name = os.path.join(base_folder, "rustpython_test_os_" + str(int(time.time())))
+
+        while os.path.isdir(name):
+            name = name + "_"
+
         os.mkdir(name)
         self.name = name
         return name
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        for root, dirs, files in os.walk(self.name, topdown=False):
-            for name in files:
-                os.remove(os.path.join(root, name))
-            for name in dirs:
-                to_remove = os.path.join(root, name)
-                if os.path.islink(to_remove):
-                    os.unlink(to_remove)
-                else:
-                    os.rmdir(to_remove)
-
-        os.rmdir(self.name)
-
+        pass
 
 class TestWithTempCurrentDir():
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/os.rs
```diff
@@ -528,6 +528,7 @@ impl DirEntryRef {
 #[derive(Debug)]
 struct ScandirIterator {
     entries: RefCell<fs::ReadDir>,
+    exhausted: RefCell<bool>,
```
coolreader18: This could just be a `Cell<bool>`

mpajkowski:
Snippet tests on Windows fails because of  missing ```nt``` module in shutil.

Should I remove ```shutil``` import from this test module or remove/comment out ```shutil.disk_usage``` function and create an issue for missing ```nt```?

palaviv:
I would recommend removing the `shutil` from this PR. We can fix `shutil` in a separate PR and then change the test to use it.

tests/snippets/stdlib_os.py
```diff
@@ -63,27 +63,37 @@
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        for root, dirs, files in os.walk(self.name, topdown=False):
+            for name in files:
+                os.remove(os.path.join(root, name))
+            for name in dirs:
+                to_remove = os.path.join(root, name)
+                if os.path.islink(to_remove):
+                    os.unlink(to_remove)
+                else:
+                    os.rmdir(to_remove)
```
palaviv: This will fail if the directory contains files in it. I think keeping the temporary folder for now is OK.

mpajkowski: The problem with keeping this directory is that the timestamp does not have enough time to get incremented during executing that tests - and simply tests fail.

I'll remove this ```__exit__``` code and try some manipulating with dirname.

windelbouwman:
@palaviv do you think this work is in good shape to merge?

palaviv:
Looks good

PR #1384, opened 2019-09-17
Fix __repr__ return value for -0.0

tsvankay:
Hi, I just noticed that the behavior of float.__repr __ is different between Python and RustPython. 

Python 3
```
>>> -0.0
-0.0
>>> float(-0.0)
-0.0
```

RustPython 
```
>>> -0.0
0.0
>>> float(-0.0)
0.0
```

On Python 3, __repr __ call on -0.0 returns -0.0, whereas on RustPython, it returns 0.0, without negative sign. I made a simple fix for that.


jeemyeong:
Thx

windelbouwman:
Nicely done! Thanks for contributing!

PR #1386, opened 2019-09-18
Enable Zero-padding with Format Strings

### LINKED ISSUE #1379: wrong str.format()
i=1
print(f'{i:04d}')

prints 1 instead of 0001

j30ng:
Hi! This patch is an attempt to fix #1379 

References
* https://www.python.org/dev/peps/pep-3101/
* https://github.com/python/cpython/blob/v3.7.4/Python/formatter_unicode.c#L219-L225

### DIFF
--- a/tests/snippets/builtin_format.py
+++ b/tests/snippets/builtin_format.py
@@ -12,3 +12,9 @@ class BadFormat:
     def __format__(self, spec):
         return 42
 assert_raises(TypeError, format, BadFormat())
+
+def test_zero_padding():
+    i = 1
+    assert f'{i:04d}' == '0001'
+
+test_zero_padding()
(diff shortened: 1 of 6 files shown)


windelbouwman:
This looks good to me, could you add the test case as mentioned in #1379 as a test snippet as well? You also might want to rebase upon the master branch to fix the windows build.

j30ng:
Rebasing is done, and I added the testcase, although the CI build seems to be failing due to some unrelated issues.

coolreader18:
Hmm, I'm guessing that's because the new Rust version that just got released comes with some changes to rustfmt/clippy. I know it's not related, but would you be able to upgrade your rustfmt version and fix the issues?

j30ng:
@coolreader18 Sure thing.

PR #1387, opened 2019-09-19
Add a hash field to PyString á la CPython, make PyString.value private

coolreader18:


vm/src/stdlib/socket.rs
```diff
@@ -502,9 +502,7 @@ impl TryFromObject for Address {
             Err(vm.new_type_error("Address tuple should have only 2 values".to_string()))
         } else {
             Ok(Address {
-                host: PyStringRef::try_from_object(vm, tuple.elements[0].clone())?
-                    .value
-                    .to_string(),
+                host: vm.to_pystr(&tuple.elements[0])?,
```
palaviv: This actually convert the first argument in the tuple to string without validation.

coolreader18: Ah, that's why the tests are failing.

coolreader18:
@palaviv I'm not sure if it improves variable lookup but it definitely should improve string-keyed dict lookup, given that it's the exact same string object. I think string interning is the next key feature to make this optimization really take effect.

PR #1392, opened 2019-09-20
Fix __eq__ of range

### LINKED ISSUE #1389: __eq__ of range problem
### Expected Result
```python
>>> range(0, 10, 3) == range(0, 11, 3)
True
>>> range(3, 1, 10) == range(5, 2, 1000)
True
```

### Actual Result
```python
>>>>> range(0, 10, 3) == range(0, 11, 3)
False
>>>>> range(3, 1, 10) == range(5, 2, 1000)
False
```

The __eq__ function of range is invalid. 
It is not enough just to compare start, stop, and step values.

HyeockJinKim:
__eq__ compare length, start, step of range

Fixes #1389


coolreader18:
Looks good, thanks!

Issue #1391, opened 2019-09-20
ValueError for strings containing whitespace in float()


rls1004:
### Expected Result
```
>>> float('   -12345\n')
-12345.0
```

### Actual Result
```
>>>>> float('   -12345\n')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: could not convert string to float: '   -12345
'
```

If the string contains `whitespace`, a ValueError occurs even though it can be converted to a float.

PR #1393, opened 2019-09-20
Fix the ValueError for whitespace in float()

rls1004:
Ignore whitespace

### Before
```
>>>>> float('   -12345\n')
ValueError: could not convert string to float: '   -12345
'
```
### After
```
>>>>> float('   -12345\n')
-12345.0
```

Fixed: #1391


coolreader18:
Looks good, thanks!

PR #1400, opened 2019-09-22
itertools.compress

pitachips:
implement itertools.compress

### DIFF
--- a/tests/snippets/stdlib_itertools.py
+++ b/tests/snippets/stdlib_itertools.py
@@ -235,3 +235,11 @@ assert 0 == next(it)
 assert 0 == next(it)
 with assert_raises(StopIteration):
     next(it)
+
+# itertools.compress
+assert list(itertools.compress("ABCDEF", [1,0,1,0,1,1])) == list("ACEF")
+assert list(itertools.compress("ABCDEF", [0,0,0,0,0,0])) == list("")
+assert list(itertools.compress("ABCDEF", [1,1,1,1,1,1])) == list("ABCDEF")
+assert list(itertools.compress("ABCDEF", [1,0,1])) == list("AC")
+assert list(itertools.compress("ABC", [0,1,1,1,1,1])) == list("BC")
+assert list(itertools.compress("ABCDEF", [True,False,"t","",1,9])) == list("ACEF")
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -73,6 +73,59 @@ impl PyItertoolsChain {
     }
 }
 
+#[pyclass(name = "compress")]
+#[derive(Debug)]
+struct PyItertoolsCompress {
+    data: PyObjectRef,
+    selector: PyObjectRef,
+}
+
+impl PyValue for PyItertoolsCompress {
+    fn class(vm: &VirtualMachine) -> PyClassRef {
+        vm.class("itertools", "compress")
+    }
+}
+
+#[pyimpl]
+impl PyItertoolsCompress {
+    #[pymethod(name = "__new__")]
+    #[allow(clippy::new_ret_no_self)]
+    fn new(
+        _cls: PyClassRef,
+        data: PyObjectRef,
+        selector: PyObjectRef,
+        vm: &VirtualMachine,
+    ) -> PyResult {
+        let data_iter = get_iter(vm, &data)?;
+        let selector_iter = get_iter(vm, &selector)?;
+
+        Ok(PyItertoolsCompress {
+            data: data_iter,
+            selector: selector_iter,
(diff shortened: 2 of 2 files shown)


windelbouwman:
Hi @pitachips , thanks for this nice addition! Please check the clippy build on travis. This build should be fixed. The azure build is broken on master, so that probably has a cause unrelated to your change.

vm/src/stdlib/itertools.rs
```diff
@@ -73,6 +73,66 @@ impl PyItertoolsChain {
...
+        let data_iter = get_iter(vm, &self.data)?;
+        let data_obj = call_next(vm, &data_iter);
+
+        if verdict {
+            match data_obj {
+                Ok(ok) => return Ok(ok),
+                Err(e) => return Err(e),
+            }
+        } else {
+            self.next(vm)
```
windelbouwman: Note that instead of recursion, you may want to prefer a loop here, to prevent hitting recursion limits.

vm/src/stdlib/itertools.rs
```diff
@@ -73,6 +73,66 @@ impl PyItertoolsChain {
...
+        let selector_iter = get_iter(vm, &self.selector)?;
+        let sel_obj = call_next(vm, &selector_iter)?;
+
+        let verdict = objbool::boolval(vm, sel_obj.clone())?;
+
+        let data_iter = get_iter(vm, &self.data)?;
+        let data_obj = call_next(vm, &data_iter);
+
+        if verdict {
+            match data_obj {
```
windelbouwman: This match could be change into a `return data_obj` statement, since both arms return from the function.

vm/src/stdlib/itertools.rs
```diff
@@ -73,6 +73,66 @@ impl PyItertoolsChain {
...
+
+    #[pymethod(name = "__next__")]
+    fn next(&self, vm: &VirtualMachine) -> PyResult {
+        let selector_iter = get_iter(vm, &self.selector)?;
+        let sel_obj = call_next(vm, &selector_iter)?;
+
+        let verdict = objbool::boolval(vm, sel_obj.clone())?;
+
+        let data_iter = get_iter(vm, &self.data)?;
+        let data_obj = call_next(vm, &data_iter);
```
windelbouwman: You may want to add the question-mark operator to the end of this line, like this:
```rust
let data_obj = call_next(vm, &data_iter)?;
```
What happens, is that in case of `Err` we return from the function, in case of `Ok`, we get the value out.

tests/snippets/stdlib_itertools.py
```diff
@@ -235,3 +235,12 @@ def assert_matches_seq(it, seq):
 assert 0 == next(it)
 with assert_raises(StopIteration):
     next(it)
+
+# itertools.compress
+assert list(itertools.compress("ABCDEF", [1,0,1,0,1,1])) == list("ACEF")
+assert list(itertools.compress("ABCDEF", [1,0,1,0,1,1])) == list("ACEF")
```
windelbouwman: This test looks the same as the line above it?

vm/src/stdlib/itertools.rs
pitachips: Sounds like a good idea. I will change it.

pitachips: @windelbouwman I made changes as suggested. Please have a look.

vm/src/stdlib/itertools.rs
```diff
@@ -73,6 +73,63 @@ impl PyItertoolsChain {
...
+            selector: selector_iter,
+        }
+        .into_ref(vm)
+        .into_object())
+    }
+
+    #[pymethod(name = "__next__")]
+    fn next(&self, vm: &VirtualMachine) -> PyResult {
+        loop {
+            let selector_iter = get_iter(vm, &self.selector)?;
```
windelbouwman: One more comment, this call to get_iter might be redundant. It will get the iterator from an iterator. The iterator was already go at line 100, so I think it is sufficient to call the call_next method.

vm/src/stdlib/itertools.rs
windelbouwman: This change looks good to me! I had one last comment, then I will merge this work.

dralley:
Might want to rebase to kick the CI pipeline and see if it passes.

dralley:
Leaving a reference to https://github.com/RustPython/RustPython/issues/1361 so that we can see from that thread when it is merged.

vm/src/stdlib/itertools.rs
pitachips: That's so true. Got rid of the lines. Thanks!

Issue #1405, opened 2019-09-23
int type casting with negative base value do not raise valueError


lntuition:
# Feature
``` python3
print(int('1', base=-1))
```

if base is negative value, it should be valueError

# Actual result
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OverflowError: Int value cannot fit into Rust u32
```

# Expected result
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: int() base must be >= 2 and <= 36, or 0
```

PR #1406, opened 2019-09-23
Fix int type casting error with negative base value

lntuition:
Change base type from u32 to i32

After result
```
>>>>> print(int(' 1 ', base=-1))
ValueError: int() base must be >= 2 and <= 36, or 0

```

Fixed: #1405

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -704,7 +704,7 @@ struct IntOptions {
     #[pyarg(positional_only, optional = true)]
     val_options: OptionalArg<PyObjectRef>,
     #[pyarg(positional_or_keyword, optional = true)]
-    base: OptionalArg<u32>,
+    base: OptionalArg<PyIntRef>,
 }
 
 impl IntOptions {
@@ -720,9 +720,9 @@ impl IntOptions {
                 }
                 base
             } else {
-                10
+                PyInt::new(10).into_ref(vm)
             };
-            to_int(vm, &val, base)
+            to_int(vm, &val, base.as_bigint())
         } else if let OptionalArg::Present(_) = self.base {
             Err(vm.new_type_error("int() missing string argument".to_string()))
         } else {
@@ -736,8 +736,14 @@ fn int_new(cls: PyClassRef, options: IntOptions, vm: &VirtualMachine) -> PyResul
 }
 
 // Casting function:
-pub fn to_int(vm: &VirtualMachine, obj: &PyObjectRef, base: u32) -> PyResult<BigInt> {
-    if base != 0 && (base < 2 || base > 36) {
+pub fn to_int(vm: &VirtualMachine, obj: &PyObjectRef, base: &BigInt) -> PyResult<BigInt> {
+    let base_u32 = match base.to_u32() {
+        Some(base_u32) => base_u32,
+        None => {
+            return Err(vm.new_value_error("int() base must be >= 2 and <= 36, or 0".to_string()))
+        }
+    };
+    if base_u32 != 0 && (base_u32 < 2 || base_u32 > 36) {
         return Err(vm.new_value_error("int() base must be >= 2 and <= 36, or 0".to_string()));
     }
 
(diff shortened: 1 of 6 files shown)


lntuition:
I want to test `ints.py` but my pytest does not working :(
does my test method is wrong? 

```
$ pytest ints.py 
======================================== test session starts ========================================
platform linux -- Python 3.6.8, pytest-5.1.3, py-1.8.0, pluggy-0.13.0
rootdir: /home/jsh/RustPython/tests/snippets
collected 0 items                                                                                   

======================================= no tests ran in 0.06s =======================================
```

coolreader18:
Try running it from `tests/` and passing `snippets/ints.py` as an argument.

lntuition:
I try with above method. but result is same.

```
$ pwd
/home/jsh/RustPython/tests
$ pytest snippets/ints.py 
============================= test session starts ==============================
platform linux -- Python 3.6.8, pytest-5.1.3, py-1.8.0, pluggy-0.13.0
rootdir: /home/jsh/RustPython/tests
collected 0 items                         
                                    
============================ no tests ran in 0.01s =============================
```

coolreader18:
Hmm, odd. And is just running `pytest` working?

vm/src/obj/objint.rs
```diff
@@ -704,7 +704,7 @@ struct IntOptions {
     #[pyarg(positional_only, optional = true)]
     val_options: OptionalArg<PyObjectRef>,
     #[pyarg(positional_or_keyword, optional = true)]
-    base: OptionalArg<u32>,
+    base: OptionalArg<i32>,
```
coolreader18: This doesn't completely fix the problem, it just shifts the range that the problem occurs, because if a PyInt can't fit into a `i32` it still overflows. I recommend changing `base` to be a PyIntRef and passing a `&BigInt` around instead of an `i32`.

lntuition:
Yeah. just running `pytest` is working, but it execute all test. How can i test only `ints.py`?

vm/src/obj/objint.rs
lntuition: Thank you for reviewing. I Change base to `PyIntRef`. And i also add test for minus base and Large value than max of i32.

coolreader18:
Try `pytest -k rustpython_ints`

coolreader18:
It's tricky because we're not actually running the test snippets with unittest/pytest, we have a custom loader `test_snippets.py` that loads all of the snippets as functions on a a test case class.

lntuition:
`pytest -k rustpython_ints` is working. thank you for your comment. 
However, I think other people can ask similar question. 
So, how about writing this method to `test/README.md`?

vm/src/stdlib/pystruct.rs
```diff
@@ -105,7 +105,7 @@ where
 }
 
 fn get_int(vm: &VirtualMachine, arg: &PyObjectRef) -> PyResult<BigInt> {
-    objint::to_int(vm, arg, 10)
+    objint::to_int(vm, arg, &BigInt::from(10))
```
windelbouwman: If you want to prevent using `BigInt` here, you can change the signature of `to_int` to accept an `Into<BigInt>` instead of bigint itself.

lntuition: Thank you for reviewing. I agree that change signature of `to_int` will make `to_int` function more general.

But I have some questions. When we use base type to `u32`, we can easily pass the argument because `u32` have copy trait. And when we use base type to `pyIntRef`, we can also pass reference of PyInt by `as_bigint` method.

But If i changed signature of `to_int` function. i cannot pass base to `to_int` function. Because `PyInt` does not have `copy` trait. Just simple modifying make this error.
```
error[E0507]: cannot move out of dereference of `pyobject::PyRef<obj::objint::PyInt>`
   --> vm/src/obj/objint.rs:721:34
    |
721 |                 to_int(vm, &val, base.value)
    |                                  ^^^^^^^^^^ move occurs because value has type `num_bigint::bigint::BigInt`, which does not implement the `Copy` trait

```

And also change base type to `BigInt` occurs other error. 
```
error[E0277]: the trait bound `num_bigint::bigint::BigInt: pyobject::TryFromObject` is not satisfied
   --> vm/src/obj/objint.rs:702:10
    |
702 | #[derive(FromArgs)]
    |          ^^^^^^^^ the trait `pyobject::TryFromObject` is not implemented for `num_bigint::bigint::BigInt`
    | 
   ::: vm/src/pyobject.rs:953:5
    |
953 |     fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self>;
    |     ---------------------------------------------------------------------------- required by `pyobject::TryFromObject::try_from_object`
```

Unfortunately, I'm not good at rust. so please give me an opinion. thx

windelbouwman: Ah, okay, the copy trait is important here. I'm not a rust wizard either, so I think the current situation is probably ok. Maybe other people could have a look at this?

coolreader18: I think that the current solution is fine as well, passing an `&BigInt` rather than an `Into<BigInt>` removes the need for cloning as well.

PR #1407, opened 2019-09-23
Add duplicate keyword argument error

### LINKED ISSUE #116: Duplicate keyword arguments should be a SyntaxError
We currently treat them as a TypeError, but CPython rejects them as invalid syntax.

(This issue will be introduced in #115, but is small enough for me to come back and fix it after that has landed.)

vazrupe:
Fixes #116

### DIFF
--- a/parser/src/function.rs
+++ b/parser/src/function.rs
@@ -1,3 +1,5 @@
+use std::collections::HashSet;
+
 use crate::ast;
 use crate::error::{LexicalError, LexicalErrorType};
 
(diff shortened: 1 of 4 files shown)


parser/src/function.rs
```diff
@@ -6,9 +6,19 @@ type FunctionArgument = (Option<Option<String>>, ast::Expression);
 pub fn parse_args(func_args: Vec<FunctionArgument>) -> Result<ast::ArgumentList, LexicalError> {
     let mut args = vec![];
     let mut keywords = vec![];
+
+    let mut keyword_names = vec![];
```
windelbouwman: Using a set like object instead of a vector would gain a speedup for larger amounts of arguments. The `contains` method for `Vec` is likely O(n) while the `contains` method (or equivalent method) on set like types is O(1).

windelbouwman:
The travis build is failing. Could you look into this? You also might want to rebase upon the master branch, since the azure build is fixed in there.

parser/src/function.rs
dralley: Might be worthwhile to use ```<collection>::with_capacity(func_args.len())``` to avoid having lots of small allocations as they get added to the collection.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d1eaab59dcae1ee483e992e19f819c91

windelbouwman: The vector re-allocation scheme is actually pretty good in rust. The amount of memory is roughly doubled, so capacity will grow something like this: 1, 2, 4, 8, 16 etc.. This results in few allocations.

Issue #1373, opened 2019-09-14
[::-1] returns wrong index for 'range' type


xarus01:
## Code
```
a = range(10)
a = a[::-1]
print(a)
```

## Expected
```
range(9, -1, -1)
```

## Result
```
range(0, 10, -1)
```

dralley:
Another slightly more complicated example:

RustPython

```
>>>>> range(0, 10, -1)[::-1]
range(0, 10)
```

CPython

```
>>> range(0, 10, -1)[::-1]
range(1, 1)
```

I'd love to work on this as a first issue, if others wouldn't mind.

vazrupe:
can I take this?

dralley:
I've already started but if I'll let you know if I give up on it.  I know where the problem is, though - ```PyRange::getitem```

dralley:
Making progress.  Found another range slicing bug which I will also try to fix.

RustPython

```
>>>>> range(0, -10)[::0]
range(0, -10, 0)
```

CPython

```
>>> range(0, -10)[::0]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: slice step cannot be zero
```

dralley:
@vazrupe Sorry!  But, I would love for you to review my PR :)

PR #1410, opened 2019-09-23
Fix range slicing behavior

dralley:
* Adds several new tests to ```tests/snippets/builtin_range.py```
* Fixes 2 bugs in range slicing behavior
* Does some refactoring

This implementation is basically [0] and [1] combined and translated to Rust.  There might be a better way to do it, but I went with just copying the CPython algorithm for the time being, since the way negative steps are handled with slices is counter-intuitive and really tricky.

[0]  https://github.com/python/cpython/blob/530f506ac91338b55cf2be71b1cdf50cb077512f/Objects/rangeobject.c#L297-L334

[1] https://github.com/python/cpython/blob/bed4817d52d7b5a383b1b61269c1337b61acc493/Objects/sliceobject.c#L368-L511

See: https://github.com/RustPython/RustPython/issues/1373

### DIFF
--- a/tests/snippets/builtin_range.py
+++ b/tests/snippets/builtin_range.py
@@ -35,6 +35,12 @@ assert range(10, 100, 3)[4:1000:5] == range(22, 100, 15)
 assert range(10)[:] == range(10)
 assert range(10, 0, -2)[0:5:2] == range(10, 0, -4)
 assert range(10)[10:11] == range(10,10)
+assert range(0, 10, -1)[::-1] == range(1, 1)
+assert range(0, 10)[::-1] == range(9, -1, -1)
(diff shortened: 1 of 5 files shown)


dralley:
Do I need to fix the formatting complaints?

coolreader18:
Yes, they're part of our CI pipeline; the builds will fail if rustfmt/clippy fails. You can run `cargo fmt --all` to fix the formatting and then just do what clippy says to fix clippy.

dralley:
Maybe it's just because I'm on nightly but clippy is complaining about stuff committed 3 months ago that I never touched.

dralley:
@coolreader18 Finally clear

coolreader18:
Thanks for contributing!

PR #1413, opened 2019-09-24
Fix exit()

ChJR:
After #1343, RustPython could not provide builtin exit() properly.

### DIFF
--- a/tests/snippets/exit.py
+++ b/tests/snippets/exit.py
@@ -1,0 +1,4 @@
+from testutils import assert_raises
+
+with assert_raises(SystemExit):
+    exit()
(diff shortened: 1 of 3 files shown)


tests/snippets/exit.py
```diff
@@ -0,0 +1,4 @@
+from testutils import assert_raises
```
youknowone: by convention, `tests/snippets/builtin_exit.py` looks better name of this test file.

ChJR: I think this file(exit.py) can hold sys.exit() test. Would be better if separating them?

youknowone:
I have no idea how this breaking is related to #1343 

But the patch looks legit to me

coolreader18:
This looks good, thanks! Could you edit `handle_exception` in `main.rs` to check if the exception is an instance of `SystemExit` and if so exit with the appropriate error code/print the args?

ChJR:
@coolreader18 

> I have no idea how this breaking is related to #1343
> 
> But the patch looks legit to me

ChJR:
I thought builtin exit() and sys.exit() need to placed in separate PRs. But they are related each other. I would handle them in this PR.

PR #1414, opened 2019-09-25
Fix weirdness with type.__new__()/type()/metaclass.__new__() in order to support enum

coolreader18:
This took me a long time to get working and ~~I'm not certain that this is actually a good solution, but~~ it does work.

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -20,6 +20,10 @@ enum ClassItem {
         py_name: String,
         setter: bool,
     },
+    Slot {
+        slot_ident: Ident,
+        item_ident: Ident,
+    },
 }
 
 fn meta_to_vec(meta: Meta) -> Result<Vec<NestedMeta>, Meta> {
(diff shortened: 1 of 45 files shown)


derive/src/pyclass.rs
```diff
@@ -20,6 +20,10 @@ enum ClassItem {
         py_name: String,
         setter: bool,
     },
+    Slot {
```
windelbouwman: What is the function of this new concept of slot? Could you try to explain a bit about it?

windelbouwman:
This is a fairly complex change. @cthulahoops could you have an eye on it as well? I'm not too familiar with this part of the codebase.

derive/src/pyclass.rs
coolreader18: It's something CPython uses to associate functions with a type without exposing them to Python code, usually for optimization. There's a lot more slots in CPython for operators, descriptors, getattr{ibute}, etc, but right now I only added `new` because the way that it works in conjunction with type_call and the [type_new wrapper](https://github.com/python/cpython/blob/master/Objects/typeobject.c#L6032) is very complicated, and it needed a separation of functionality from the type-checking and stuff.

cthulahoops:
I've had a look through to get the gist of it, but I'm going to need to go through it again more carefully to really understand it as it's been a while since I've looked at it.

Looks reasonable, I think. What's making you uncertain about this solution?

coolreader18:
The original implementation before I changed it in 43a03bbaf4d8945246479fb498db71881af02de2 tried to put the type_new wrapper on every type from the `extend_class` macro which felt very hacky. I've since realized it's unnecessary; it only needs to be on `type` and `object`.

vm/src/obj/objtype.rs
```diff
@@ -213,13 +228,15 @@ pub fn init(ctx: &PyContext) {
...
     extend_class!(&ctx, &ctx.types.type_type, {
+        "mro" => ctx.new_rustfunc(type_mro),
         "__call__" => ctx.new_rustfunc(type_call),
         "__dict__" =>
         PropertyBuilder::new(ctx)
                 .add_getter(type_dict)
                 .add_setter(type_dict_setter)
                 .create(),
-        "__new__" => ctx.new_rustfunc(type_new),
+        "__new__" => ctx.new_classmethod(type_new),
```
youknowone: Is this line nessessary? It seems `type` is a subclass of `object`

coolreader18: It is, because `type` isn't in the mro of object subclasses.

```python
>>> class A: pass
... 
>>> A.mro()
[<class '__main__.A'>, <class 'object'>]
```

If there wasn't anything on `object`, `class_get_attr` wouldn't encounter `__new__` in the mro.

Wait no never mind, I see what you're saying. It only needs to be on `object` because `type` has `object` in *its* mro.

coolreader18:
Is this good to merge now?

Issue #1415, opened 2019-09-25
int type casting which number has radix and front underscore is fail


lntuition:
# Feature
``` python3
print(int('0x_10', base=0))
```

if some number has radix and front underscore, type casting fail

# Actual result
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 0: '0x_10'
```

# Expected result
```
16
```

lntuition:
I think this issue should be fix same part as #1408. So I will make one PR for fix #1405 and #1408 both.

Issue #1408, opened 2019-09-23
int type casting with radix and larger base value is wrong


lntuition:
# Feature
``` python3
print(int('0b1', base=12))
```

If base is large value, radix should be considered value.
With above example, it is b1 with base 12(not 1 with base 2)

# Actual result
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: invalid literal for int() with base 12: '0b1'
```

# Expected result
```
133
```

PR #1420, opened 2019-09-26
Fix int type casting errors

lntuition:
this is fix of #1408 and #1415.

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -793,9 +793,18 @@ fn str_to_int(vm: &VirtualMachine, literal: &str, base: &BigInt) -> PyResult<Big
     if let Some(radix_candidate) = radix_candidate {
         if let Some(matched_radix) = detect_base(&radix_candidate) {
             if base_u32 == 0 || base_u32 == matched_radix {
+                /* If base is 0 or equal radix number, it means radix is validate
+                 * So change base to radix number and remove radix from literal
+                 */
                 base_u32 = matched_radix;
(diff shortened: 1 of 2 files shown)


vm/src/obj/objint.rs
```diff
@@ -792,13 +792,19 @@ fn str_to_int(vm: &VirtualMachine, literal: &str, base: &BigInt) -> PyResult<Big
...
+            if base_u32 == 0 || base_u32 == matched_radix {
                 base_u32 = matched_radix;
-            }
+                buf.drain(radix_range);
 
-            buf.drain(radix_range);
+                if buf.starts_with('_') {
+                    buf.remove(0);
+                }
+            } else if (matched_radix == 2 && base_u32 < 12)
```
windelbouwman: I tried to understand this code, but how does it work? Could you add a small comment to explain what is happening here? Do you check ascii codes?

lntuition: This is very simple. It is code to fix `'0x_1'` is unvalidate (issue : #1420)
First, we need to know that `'_1'` is invalid literal (Also it is filtered `validate_literal()`)
 
In current logic, After we detect validate radix, then remove the radix from literal. 
So if executed `int('0x_1', base=16)` and after doing `buf.drain(radix_range)`, literal will be considered `_1`. As we saw above it is invalid literal.

So i just remove the first underscore when `buf.drain(radix_range)` exectued, So literal will be considered `1`. I select this method Because it is easy to implement and do not hurt current logic.

PR #1423, opened 2019-09-26
Fix some clippy complaints

dralley:


vm/src/obj/objbytes.rs
```diff
@@ -509,8 +509,8 @@ impl PyBytesRef {
                                     }
                                 }
                                 None => {
-                                    if replacing_char.is_some() {
-                                        decode_content.push(replacing_char.unwrap())
+                                    if let Some(replacing_char) = replacing_char {
+                                        decode_content.push(replacing_char)
```
dralley: While the existing pattern is a little unidiomatic, I'm not sure I like how this is now the only place in the function where ```replacing_char``` is a different type.  I wanted to point it out in case you also think it might not be worthwhile.

coolreader18: This looks fine to me, but you could rename the `char` version of the variable something else if you feel that it's inconsistent.

dralley:
Feel free to merge, I wasn't planning on changing it unless someone else thought it should.

PR #1424, opened 2019-09-26
Allow float() to be called without an argument

### LINKED ISSUE #1421: No argument with float function do not return 0.0
# Feature
``` python3
print(float())
```

If no argument is given, float function have to return 0.0

# Actual result
```
TypeError: Expected at least 2 arguments (1 given)
```

# Expected result
```
0.0
```

dralley:
Copy the same pattern used by int()

closes #1421

### DIFF
--- a/vm/src/obj/objfloat.rs
+++ b/vm/src/obj/objfloat.rs
@@ -157,6 +157,19 @@ fn inner_gt_int(value: f64, other_int: &BigInt) -> bool {
 #[pyimpl]
 #[allow(clippy::trivially_copy_pass_by_ref)]
 impl PyFloat {
+    #[pymethod(name = "__new__")]
+    fn float_new(
+        cls: PyClassRef,
+        arg: OptionalArg<PyObjectRef>,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyFloatRef> {
+        let float_val = match arg {
+            OptionalArg::Present(val) => to_float(vm, &val),
+            OptionalArg::Missing => Ok(0f64),
+        };
+        PyFloat::from(float_val?).into_ref_with_type(vm, cls)
+    }
+
     #[pymethod(name = "__eq__")]
     fn eq(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
         let value = self.value;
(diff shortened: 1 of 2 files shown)


vm/src/obj/objfloat.rs
```diff
@@ -667,4 +686,7 @@ impl TryFromObject for IntoPyFloat {
 #[rustfmt::skip] // to avoid line splitting
 pub fn init(context: &PyContext) {
     PyFloat::extend_class(context, &context.types.float_type);
+    extend_class!(context, &context.types.float_type, {
+        "__new__" => context.new_rustfunc(float_new),
+    });
```
coolreader18: You should keep `float_new` in the `pyimpl`, it's a really simple function anyway.

vm/src/obj/objfloat.rs
```diff
@@ -574,6 +540,59 @@ impl PyFloat {
     }
 }
 
+#[derive(FromArgs)]
+struct FloatOptions {
+    #[pyarg(positional_only, optional = true)]
+    val_options: OptionalArg<PyObjectRef>,
+}
```
coolreader18: I don't think this separate options struct is necessary, it's positional by default when you just use `OptionalArg` as a function argument.

vm/src/obj/objfloat.rs
coolreader18: I do like the separation of `to_float` out into its own function though, nice work!

dralley:
@coolreader18 When I try this, it panics. Any idea what's going on?

```
>>>>> float()
thread 'main' panicked at 'unexpected payload for type', src/libcore/option.rs:1190:5
```

Commit pushed so you can see the changes.

vm/src/obj/objfloat.rs
```diff
@@ -342,6 +342,16 @@ impl PyFloat {
         -self.value
     }
 
+    #[pymethod(name = "__pow__")]
```
coolreader18: ^^^^

Should be `__new__`

dralley: I feel really, really dumb.  Thanks.

vm/src/obj/objfloat.rs
dralley: I actually had no idea you could do that, neat!

vm/src/obj/objfloat.rs
dralley: I actually had no idea you could do that, neat!

coolreader18:
Looks like the rustfmt job is failing.

PR #1426, opened 2019-09-27
impl __sizeof__ for bytearray, int, list and str

RobertBerglund:
work in progress

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -1,4 +1,5 @@
 use std::fmt;
+use std::mem::size_of;
 use std::str;
 
 use num_bigint::{BigInt, Sign};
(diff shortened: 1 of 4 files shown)


RobertBerglund:
I planned on implementing the remaining _sizeof_ methods as soon as I get some free time. 
Thanks for the feedback !

vm/src/obj/objstr.rs
```diff
@@ -279,6 +280,11 @@ impl PyString {
         self.value.chars().count()
     }
 
+    #[pymethod(name = "__sizeof__")]
+    fn sizeof(&self, _vm: &VirtualMachine) -> usize {
```
coolreader18: This should use `self.value.capacity()`

vm/src/obj/objint.rs
```diff
@@ -566,6 +567,11 @@ impl PyInt {
         !self.value.is_zero()
     }
 
+    #[pymethod(name = "__sizeof__")]
+    fn sizeof(&self, _vm: &VirtualMachine) -> usize {
```
coolreader18: Maybe use `self.value.bits()` and then round up to the nearest byte for this.

RobertBerglund: something like `(self.value.bits() as f64 / 8.).ceil() as usize` ?

windelbouwman:
@RobertBerglund I would suggest to merge this work so far, otherwise this branch will be in a more conflicted status.

Issue #1427, opened 2019-09-28
int.from_bytes is not working with argument keyword


lntuition:
# Feature
``` python3
print(int.from_bytes(b'\x00\x10', byteorder='big'))
```

int.from_bytes is not working with argument keyword

# Actual result
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Expected at least 2 arguments (1 given)
```

# Expected result
```
16
```

lntuition:
While resolving this issue, I also find `int.to_bytes` have similiar problem. So I will make another patch for resolving error in `int.to_bytes`

PR #1429, opened 2019-09-28
Fix error in int.from_bytes and int.to_bytes

lntuition:
Add new struct to fix int argument keyword error in `int.from_bytes` and `int.to_bytes`
Also Change ValueError to OverflowError like cPython

fixed: #1427

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -579,21 +579,21 @@ impl PyInt {
 
     #[pymethod]
     #[allow(clippy::match_bool)]
-    fn from_bytes(options: IntFromByteOptions, vm: &VirtualMachine) -> PyResult<BigInt> {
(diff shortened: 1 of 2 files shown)


vm/src/obj/objint.rs
```diff
@@ -735,6 +723,26 @@ fn int_new(cls: PyClassRef, options: IntOptions, vm: &VirtualMachine) -> PyResul
     PyInt::new(options.get_int_value(vm)?).into_ref_with_type(vm, cls)
 }
 
+#[derive(FromArgs)]
+struct IntFromByteOptions {
```
windelbouwman: Suggestion here to rename this struct into `IntFromByteArguments`. Likewise for the other struct below `IntToByteArguments`.

lntuition: Thank you for reviewing. I change struct name. And also change argument name from `option` to `args`

PR #1430, opened 2019-09-28
Generate SyntaxError at integer starting with 0

### LINKED ISSUE #1422: Integer problem starting with 0
## Feature
```python
0123
```

Integer starting with 0 should raise SyntaxError.
I think it should be fixed in lexer.rs

### Expected Result
```python
>>> 0123
  File "<stdin>", line 1
    0123
       ^
SyntaxError: invalid token
```
### Actual Result
```python
>>>>> 0123
123
```

HyeockJinKim:
Generate syntaxError at integer starting with 0,
not zero

Fixes #1422

### DIFF
--- a/parser/src/lexer.rs
+++ b/parser/src/lexer.rs
@@ -8,6 +8,7 @@ pub use super::token::Tok;
 use crate::error::{LexicalError, LexicalErrorType};
 use crate::location::Location;
 use num_bigint::BigInt;
+use num_traits::identities::Zero;
 use num_traits::Num;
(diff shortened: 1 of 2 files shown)


windelbouwman:
Nicely implemented! I'm still a bit clueless why a leading zero is not allowed. Do you have any ideas?

coolreader18:
I think it's ambiguous/not explicit to whether the person writing it intended it to be a decimal literal or octal literal, as some languages take a leading 0 to mean octal, so they decided to ban it altogether.

PR #1433, opened 2019-09-29
fix error message for __call__ for non-callable objects

### LINKED ISSUE #1416: Wrong error message when calling a non-callable object
CPython

```
>>> 'linux'()
TypeError: 'str' object is not callable
>>> 1()
TypeError: 'int' object is not callable
```

RustPython

```
>>>>> 'linux'()
TypeError: Unsupported method: __call__
>>>>> 1()
TypeError: Unsupported method: __call__
```

pitachips:
closes #1416.

Changed only `vm.rs` file, and the other files were just auto-formatted with `cargo fmt`

### DIFF
--- a/vm/src/vm.rs
+++ b/vm/src/vm.rs
@@ -603,15 +603,13 @@ impl VirtualMachine {
             self.invoke(&function, args.insert(object.clone()))
         } else if let Some(PyBuiltinFunction { ref value }) = func_ref.payload() {
             value(self, args)
+        } else if self.is_callable(&func_ref) {
+            self.call_method(&func_ref, "__call__", args)
         } else {
-            if self.is_callable(&func_ref) {
-                self.call_method(&func_ref, "__call__", args)
-            } else {
-                Err(self.new_type_error(format!(
-                    "'{}' object is not callable",
-                    func_ref.class().name
-                )))
-            }
+            Err(self.new_type_error(format!(
+                "'{}' object is not callable",
+                func_ref.class().name
+            )))
         }
     }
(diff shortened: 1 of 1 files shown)


coolreader18:
Could you upgrade your rust version to `1.38`? The reason that the other files were auto-formatted is because master is up to date with 1.38

pitachips:
Upgrading fixed the problem. Thanks for letting me know.

dralley:
Are there any test snippets for this?

vm/src/vm.rs
```diff
@@ -563,7 +563,7 @@ impl VirtualMachine {
                 let wrapped = self.call_get_descriptor(func, obj.clone())?;
                 self.invoke(&wrapped, args)
             }
-            None => Err(self.new_type_error(format!("Unsupported method: {}", method_name))),
+            None => Err(self.new_type_error(format!("{:?} object is not callable", cls.name))),
```
coolreader18: You should put the change in `vm.invoke` instead, maybe check if it's callable using `vm.is_callable` before calling `call_method`.

pitachips:
> Are there any test snippets for this?

I added tests that check not just Error type but also message, since the previous code also raised `TypeError`. However I am not sure if this error message comparison is a good choice, because I am afraid Python may change error messages in future versions. Any suggestions?

coolreader18:
I don't think any test snippets are necessary, as this is mainly just a UX change.

pitachips:
Alright, I'll just get rid of the added tests.

Issue #1437, opened 2019-09-29
int.from_byte do not working with array.array byte like object


lntuition:
# Feature
``` python3
int.from_bytes([255, 0, 0], byteorder='big')
```

int.from_bytes argument bytes must either be bytes-like object(bytes, bytearray, array.array)
but do not working at passing array.array byte like object

# Actual result
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: a bytes-like object is required, not list
```

# Expected result
```
16711680
```

PR #1438, opened 2019-09-29
Add logic to check array.array byte like object

lntuition:
Add logic to check array.array byte like object
After apply this patch, We can handle array.array byte like object

Fixed: #1437

### DIFF
--- a/vm/src/obj/objlist.rs
+++ b/vm/src/obj/objlist.rs
@@ -9,12 +9,13 @@ use num_traits::{One, Signed, ToPrimitive, Zero};
 use crate::function::OptionalArg;
 use crate::pyobject::{
     IdProtocol, PyClassImpl, PyContext, PyIterable, PyObjectRef, PyRef, PyResult, PyValue,
-    TryFromObject,
+    TryFromObject, TypeProtocol,
 };
 use crate::vm::{ReprGuard, VirtualMachine};
 
 use super::objbool;
-//use super::objint;
+use super::objbyteinner;
+use super::objint::PyIntRef;
 use super::objiter;
 use super::objsequence::{
     get_elements_list, get_item, seq_equal, seq_ge, seq_gt, seq_le, seq_lt, seq_mul, SequenceIndex,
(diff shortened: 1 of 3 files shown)


lntuition:
While i resolving this issue, i have a question. 

https://github.com/RustPython/RustPython/blob/328ad7a504eb936523867515d8f25880c675e566/vm/src/obj/objbyteinner.rs#L39-L53

If i change to 51 line something like this
``` Rust
l @ PyList => {
                 l.get_byte_inner(vm);
             },
``` 

it makes such errors while `cargo build`.
```
error: recursion limit reached while expanding the macro `$crate::match_class`
   --> vm/src/macros.rs:277:9
    |
277 |           $crate::match_class!(match ($($target)*) { $($inner)* })
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    | 
   ::: vm/src/obj/objbyteinner.rs:41:9
    |
41  | /         match_class!(match obj {
42  | |             i @ PyBytes => Ok(PyByteInner {
43  | |                 elements: i.get_value().to_vec()
44  | |             }),
...   |
57  | |             ))),
58  | |         })
    | |__________- in this macro invocation
    |
    = help: consider adding a `#![recursion_limit="256"]` attribute to your crate
```
I don't know why this error occurs. I think changing code as above method do not change result.
I hope someone solve my question. thx

coolreader18:
It's in the macro expansion, you can do what it says and bump the recursion limit to 256.

lntuition:
@coolreader18 Sorry, I didn't understand well. do you mean Add `#![recursion_limit="256"]` is valid solution? I also tried to add `#![recursion_limit="256"]` to `objbyteinner.rs` or `macro.rs`. But it still doesn't work

coolreader18:
Ah, sorry, it should be in `vm/src/lib.rs`, that's what it means by "your crate".

vm/src/obj/objlist.rs
```diff
@@ -96,6 +97,30 @@ impl PyList {
 
         start..stop
     }
+
+    pub fn get_byte_inner(&self, vm: &VirtualMachine) -> PyResult<objbyteinner::PyByteInner> {
+        let mut elements = Vec::<u8>::new();
```
youknowone: how about `Vec::<u8>::with_capacity(self.elements.borrow().len())`?

lntuition: Thank you for reviewing. Also there is implemented method `get_len`. in PyList. I will change to `Vec::<u8>::with_capacity(self.get_len())`

vm/src/obj/objlist.rs
```diff
@@ -96,6 +97,30 @@ impl PyList {
...
+        for elem in self.elements.borrow().iter() {
+            let method = vm.get_method_or_type_error(elem.clone(), "__int__", || {
+                format!(
+                    "'{}' object cannot be interpreted as an integer",
+                    elem.class().name
+                )
+            })?;
+            let result = vm.invoke(&method, PyFuncArgs::default())?;
+            if let Some(result) = result.payload::<PyInt>() {
+                match result.as_bigint().to_u8() {
```
coolreader18: It's ignored if `__int__` doesn't return an `int`, maybe use `PyIntRef::try_from_object`?

lntuition: Thank you for reviewing. I fixed my code. Also current code has bug. Because current code cannot handle `int.from_bytes([255.0, 0, 0], 'big')` as error. I fixed it, too.

PR #1441, opened 2019-10-01
Handle unicode indexing by character, not by grapheme

### LINKED ISSUE #1417: Unicode string indexing differences
Courtesy of the ["big list of naughty strings"](https://github.com/minimaxir/big-list-of-naughty-strings/):

CPython

```
>>> t = "Ṱ̺̺̕o͞ ̷i̲̬͇̪͙n̝̗͕v̟̜̘̦͟o̶̙̰̠kè͚̮̺̪̹̱̤ ̖t̝͕̳̣̻̪͞h̼͓̲̦̳̘̲e͇̣̰̦̬͎ ̢̼̻̱̘h͚͎͙̜̣̲ͅi̦̲̣̰̤v̻͍e̺̭̳̪̰-m̢iͅn̖̺̞̲̯̰d̵̼̟͙̩̼̘̳ ̞̥̱̳̭r̛̗̘e͙p͠r̼̞̻̭̗e̺̠̣͟s̘͇̳͍̝͉e͉̥̯̞̲͚̬͜ǹ̬͎͎̟̖͇̤t͍̬̤͓̼̭͘ͅi̪̱n͠g̴͉ ͏͉ͅc̬̟h͡a̫̻̯͘o̫̟̖͍̙̝͉s̗̦̲.̨̹͈̣"
>>> t[0]
'Ṱ'
>>> t[1]
'̺'
>>> t[2]
'̺'
>>> t[3]
'̕'
>>> t[4]
'o'
>>> t[5]
'͞'
>>> t[6]
' '
>>> t[7]
'̷'
>>> t[8]
'i'
```

RustPython

```
>>>>> t = "Ṱ̺̺̕o͞ ̷i̲̬͇̪͙n̝̗͕v̟̜̘̦͟o̶̙̰̠kè͚̮̺̪̹̱̤ ̖t̝͕̳̣̻̪͞h̼͓̲̦̳̘̲e͇̣̰̦̬͎ ̢̼̻̱̘h͚͎͙̜̣̲ͅi̦̲̣̰̤v̻͍e̺̭̳̪̰-m̢iͅn̖̺̞̲̯̰d̵̼̟͙̩̼̘̳ ̞̥̱̳̭r̛̗̘e͙p͠r̼̞̻̭̗e̺̠̣͟s̘͇̳͍̝͉e͉̥̯̞̲͚̬͜ǹ̬͎͎̟̖͇̤t͍̬̤͓̼̭͘ͅi̪̱n͠g̴͉ ͏͉ͅc̬̟h͡a̫̻̯͘o̫̟̖͍̙̝͉s̗̦̲.̨̹͈̣"
>>>>> t[0]
'Ṱ̺̺̕'
>>>>> t[1]
'o͞'
>>>>> t[2]
' ̷'
>>>>> t[3]
'i̲̬͇̪͙'
>>>>> t[4]
'n̝̗͕'
>>>>> t[5]
'v̟̜̘̦͟'
>>>>> t[6]
'o̶̙̰̠'
>>>>> 

```

dralley:
Python isn't grapheme aware, so for compatibility sake, we shouldn't be
either.

closes #1417


dralley:
I can try to make the implementations more efficient if required, or I can just file an issue for future improvement.  My main priority was correctness.

vm/src/obj/objstr.rs
```diff
@@ -1495,73 +1496,77 @@ impl PySliceableSequence for String {
...
+        // an inefficient but simple implementation
+        // TODO: don't check whether the range contains each and every index
+        // try to remove temp String
+       let slice = self.chars()
+            .enumerate()
+            .filter(|(idx, _)| range.contains(idx))
+            .map(|(_, ch)| ch)
+            .collect::<String>();
+
+        slice.chars().rev().collect()
```
dralley: For various reasons I couldn't do ```.rev()``` after ```.enumerate()``` so this was my workaround.  It can probably be done better

silmeth: I’d suggest something like this to use the `range` directly instead of enumerating and checking each index if it fits:

```rust
self.chars()
    .skip(range.start)
    .take(range.end - range.start)
    .collect::<String>()
    .chars()
    .rev()
    .collect();
```

Unfortunately I don’t think we can get rid of this collect, because `Skip` doesn’t implement `DoubleEndedIterator`.

silmeth: Alternatively, to get rid of the collected string, you could first count all the chars and use that information:

```rust
let count = self.chars().count();

self.chars()
    .rev()
    .skip(count - range.end)
    .take(range.end - range.start)
    .collect()
```

See [this in playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=ce8152952e5868ae553eb0679845a438).

dralley: Cool, thanks.  I'll work on this later today.

vm/src/stdlib/ast.rs
```diff
@@ -620,8 +620,12 @@ fn comprehension_to_ast(
 
 fn string_to_ast(vm: &VirtualMachine, string: &ast::StringGroup) -> PyResult<AstNodeRef> {
     let string = match string {
-        ast::StringGroup::Constant { value } => node!(vm, Str, { s => vm.ctx.new_str(value.clone()) }),
-        ast::StringGroup::FormattedValue { value, .. } => node!(vm, FormattedValue, { value => expression_to_ast(vm, value)? }),
+        ast::StringGroup::Constant { value } => {
```
dralley: Rustfmt

dralley:
@silmeth Updated

dralley:
@coolreader18 How does this look

Issue #1431, opened 2019-09-29
Implement slice method


HyeockJinKim:
## Feature
Currently, the `__repr__`, comparison, and` indices` functions are not implemented in slice.

### Expected result
```python
>>> slice(10)
slice(None, 10, None)
>>> slice(10).indices(0)
(0, 0, 1)
>>> slice(1) < slice(20)
True
>>> slice(10) == slice(10)
True
```

### Actual result
```python
>>>>> slice(10)
<slice object at 0x7fd391659260>
>>>>> slice(10).indices()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'slice' object has no attribute 'indices'
>>>>> slice(1) < slice(20)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Unsupported operand types for '<': 'slice' and 'slice'
>>>>> slice(10) == slice(10)
False
```

PR #1443, opened 2019-10-01
Implement __eq__, __ne__ of slice 

HyeockJinKim:
Compute `__eq__` and `__ne__` by comparing start,
stop and step of slice

Issue #1431

### DIFF
--- a/vm/src/obj/objslice.rs
+++ b/vm/src/obj/objslice.rs
@@ -2,10 +2,10 @@ use crate::function::{OptionalArg, PyFuncArgs};
 use crate::pyobject::{
     IdProtocol, PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
 };
-use crate::vm::VirtualMachine;
 
 use crate::obj::objint::PyInt;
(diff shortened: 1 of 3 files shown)


HyeockJinKim:
Can I get a review?

vm/src/obj/objslice.rs
```diff
@@ -141,6 +175,12 @@ fn to_index_value(vm: &VirtualMachine, obj: &PyObjectRef) -> PyResult<Option<Big
     }
 }
 
+fn _eq(a: PyObjectRef, b: PyObjectRef, vm: &VirtualMachine) -> PyResult<bool> {
+    let eq = vm._eq(a.clone(), b.clone())?;
+    let value = objbool::boolval(vm, eq)?;
+    Ok(value)
```
coolreader18: This pattern is used a lot in the codebase; if you'd like, you could add this as a method (`bool_eq`?) on `VirtualMachine` given that you've already separated it out into its own function here.

HyeockJinKim: I think that's a good idea. Fixed vm to have a `bool_eq` function.

PR #1445, opened 2019-10-01
Change sys.version* attributes to return the Python version

### LINKED ISSUE #1396: sys.version_info should return target CPython version, not RustPython version
Currently `sys.version_info` returns "0.1.0", which is a version of RustPython. This is incorrect, it should be `sys.implementation.version`, see https://docs.python.org/3/library/sys.html#sys.implementation.

Practically speaking, widely used `six` library uses `PY3 = sys.version_info[0] == 3` to test whether an implementation is Python 2 or 3. RustPython is currently detected as not being Python 3.

dralley:
Return the version of Python that we attempt to be compatible with, not
the version numbers of the RustPython interpreter itself.

closes #1396

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -454,7 +454,7 @@ pub fn impl_pystruct_sequence(attr: AttributeArgs, item: Item) -> Result<TokenSt
         #struc
         #class_def
         impl #ty {
-            fn into_struct_sequence(&self,
+            pub fn into_struct_sequence(&self,
                 vm: &::rustpython_vm::vm::VirtualMachine,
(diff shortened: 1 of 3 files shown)


dralley:
This is still a little flawed because it prints differently from the way CPython prints.

```
>>>>> sys.version_info
(3, 5, 0, 'alpha', 0)
```

```
>>> sys.version_info
sys.version_info(major=3, minor=7, micro=4, releaselevel='final', serial=0)
```

But I'm not sure where to look to fix that.  Any pointers?

coolreader18:
You could change the pystruct_sequence proc macro to add a repr method, but that's probably an issue for another PR.

dralley:
I can file a new issue for that

Issue #1442, opened 2019-10-01
Many objects do not implement `__ne__`.


HyeockJinKim:
## Feature
Many objects implement `__eq__` but do not implement `__ne__`.

### Expected result
```python
>>> ''.__ne__('')
False
>>> ''.__ne__('e')
True
>>> [].__ne__([])
False
>>> [].__ne__([32])
True
>>> (1,).__ne__((2,))
True
>>> (1,).__ne__((1,))
False
>>> {}.__ne__({})
False
>>> {}.__ne__({'a':'b'})
True
```

### Actual result
```python
>>>>> ''.__ne__('')
<NotImplementedType object at 0x7ff64ec12000>
>>>>> [].__ne__([])
<NotImplementedType object at 0x7ff64ec12000>
>>>>> (1,).__ne__((2,))
<NotImplementedType object at 0x7ff64ec12000>
>>>>> {}.__ne__({})
<NotImplementedType object at 0x7ff64ec12000>
```

makarchuk:
Should this be closed?

HyeockJinKim:
In addition to the objects mentioned above, there are some objects that __ne__ is not yet implemented. (such as float, complex, ...)

youknowone:
we'd better to have a checklist to make this done.

youknowone:
Once #1642 is done, implementing `__ne__` will be a lot easier than now.

felippemr:
👋 , is this issue still relevant?

youknowone:
I think this is now done. (whats_left.sh doesn't have any `__ne__` in the entry)
thanks!

PR #1446, opened 2019-10-01
Implement some __ne__ methods

makarchuk:
Currently involves some copy-paste from `__eq__`, which might be ok, but i think it can be improved, if you can advise me some way to apply `not` to `PyResult` object. So basically `__ne__` methods will be implemented as `not!(self.__eq__(other, vm)`. 

But the way it is now, it'll work just as well.

Issue: #1442


vm/src/obj/objlist.rs
```diff
@@ -519,6 +519,17 @@ impl PyListRef {
         }
     }
 
+    fn ne(self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
```
youknowone: Because `eq` and `ne` shares lots part of the implementations, creating a functoin `inner_eq` like other classes will be great.

vm/src/obj/objset.rs
```diff
@@ -358,6 +358,11 @@ impl PySet {
         try_set_inner!(vm, other, |other| self.inner.borrow().eq(other, vm))
     }
 
+    #[pymethod(name = "__ne__")]
+    fn ne(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        try_set_inner!(vm, other, |other| self.inner.borrow().eq(other, vm))
```
youknowone: this line looks just same as `eq`

vm/src/obj/objtuple.rs
```diff
@@ -152,6 +152,16 @@ impl PyTupleRef {
         }
     }
 
+    fn ne(self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
```
youknowone: `inner_eq` here too

makarchuk:
Sure. Will fix my code and add tests. Today, hopefully

vm/src/obj/objdict.rs
```diff
@@ -146,6 +146,15 @@ impl PyDictRef {
         }
     }
 
+    fn ne(self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        if let Some(other) = other.payload::<PyDict>() {
+            let eq = !self.inner_eq(other, vm)?;
```
silmeth: This local variable holds a not-equals result, thus calling it `eq` seems misleading to me. I’d either rename it to `neq` *or* change it to
```rust
let eq = self.inner_eeq(other, vm)?;
Ok(vm.ctx.new_bool(!eq))
```

makarchuk: Sure. Lazy copy-pasting happened here. Will fix.

makarchuk:
Ok, will implement for `FrozenSet`.

One more question. Currently `RustPython` and `cpython` have different implementations of `str.__eq__`.
`RustPython`:
```
"".__eq__(3))
>> False
```
`cpython`: 
```
In [1]: "".__eq__(10)                                                           
Out[1]: NotImplemented
```

I'll be happy to fix it (return `False` if `other` is `string`), if neccessary, but, not sure what some side effects of this change might be. Will be happy to hear your opinion.

youknowone:
Nice catch! I think compatibility is always better. Please share any test is broken by the change.

makarchuk:
I guess no regressions occured due to changes in `str.__eq__`, so, that's nice

PR #1447, opened 2019-10-01
Fix properties on `type`

coolreader18:
Also add `type.__bases__`


windelbouwman:
Nice! Could you add some test snippets which show the behavior of `__bases__`?

PR #1449, opened 2019-10-01
Rework the pyclass proc macro a bit

coolreader18:


windelbouwman:
Interesting stuff, I do not fully grasp this, but it looks okay.

PR #1452, opened 2019-10-02
Fix sequence comparison

### LINKED ISSUE #1450: Comparison of list does not return an error.
## Feature
Comparison of list does not return an error.

### Expected result
```python
>>> [0, 1, 2] < [0, 1, []]
TypeError: '<' not supported between instances of 'int' and 'list'
```

### Actual result
```python
>>>>> [0, 1, 2] < [0, 1, []]
False
```

https://github.com/RustPython/RustPython/blob/d9cdb32e49f0d18ef58ce0b94bdda1a9b6b9a4a3/vm/src/vm.rs#L1175-L1197

https://github.com/RustPython/RustPython/blob/d9cdb32e49f0d18ef58ce0b94bdda1a9b6b9a4a3/vm/src/obj/objsequence.rs#L305-L343

_ge, _gt, _le, and _lt return errors, but seq_ge, seq_gt, seq_le, and seq_lt do not check for errors. (_eq, _ne do not need to return an error.)

HyeockJinKim:
Fixed gt, lt comparison of sequence. Also, the values of
le and ge are calculated using the not operation of gt and lt

Fixes #1450

### DIFF
--- a/tests/snippets/list.py
+++ b/tests/snippets/list.py
@@ -488,3 +488,42 @@ del x[-1:-5:-1]
 assert x == [0, 1, 2, 3, 4, 5]
 x = list(range(10))
 del x[-5:-1:-1]
+
+# list gt, ge, lt, le
+assert_raises(TypeError, lambda: [0, []] < [0, 0])
+assert_raises(TypeError, lambda: [0, []] <= [0, 0])
+assert_raises(TypeError, lambda: [0, []] > [0, 0])
+assert_raises(TypeError, lambda: [0, []] >= [0, 0])
+
+assert_raises(TypeError, lambda: [0, 0] < [0, []])
+assert_raises(TypeError, lambda: [0, 0] <= [0, []])
+assert_raises(TypeError, lambda: [0, 0] > [0, []])
+assert_raises(TypeError, lambda: [0, 0] >= [0, []])
+
+assert [0, 0] < [1, -1]
+assert [0, 0] < [0, 0, 1]
+assert [0, 0] < [0, 0, -1]
+assert [0, 0] <= [0, 0, -1]
+assert not [0, 0, 0, 0] <= [0, -1]
+
+assert [0, 0] > [-1, 1]
+assert [0, 0] >= [-1, 1]
+assert [0, 0, 0] >= [-1, 1]
+
+assert [0, 0] <= [0, 1]
+assert [0, 0] <= [0, 0]
+assert [0, 0] <= [0, 0]
+assert not [0, 0] > [0, 0]
+assert not [0, 0] < [0, 0]
+
+assert not [float('nan'), float('nan')] <= [float('nan'), 1]
+assert not [float('nan'), float('nan')] <= [float('nan'), float('nan')]
+assert not [float('nan'), float('nan')] >= [float('nan'), float('nan')]
+assert not [float('nan'), float('nan')] < [float('nan'), float('nan')]
+assert not [float('nan'), float('nan')] > [float('nan'), float('nan')]
+
+assert [float('inf'), float('inf')] >= [float('inf'), 1]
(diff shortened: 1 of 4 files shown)


HyeockJinKim:
However, the current code remains an problem for returning an unsupported operand error for `<` instead of error for `<=` in `seq_le` and `seq_ge`.
I think that to solve this, we need to implement the code for `seq_le` and `seq_ge` respectively.

HyeockJinKim:
Can I get a review?

tests/snippets/list.py
```diff
@@ -488,3 +488,27 @@ def bad_iter_assign():
...
+assert [0, 0] < [0, 0, -1]
+assert [0, 0] <= [0, 0, -1]
+
+assert [0, 0] > [-1, 1]
+assert [0, 0] >= [-1, 1]
+assert [0, 0, 0] >= [-1, 1]
+
+assert [0, 0] <= [0, 1]
+assert [0, 0] <= [0, 0]
+assert [0, 0] >= [0, 0]
```
youknowone: Adding tests with inf, nan will be good
```
assert [nan] == [nan]
assert [nan] >= [nan]
...
```

HyeockJinKim: I added a test case and modified the `seq_ge` and` seq_le` codes.
Can I get a review again?

vm/src/obj/objsequence.rs
```diff
@@ -386,15 +347,39 @@ pub fn seq_ge(
...
+    for (a, b) in Iterator::zip(zelf.iter(), other.iter()) {
+        if vm.bool_gt(a.clone(), b.clone())? {
+            return Ok(true);
+        } else if !vm.bool_eq(a.clone(), b.clone())? {
+            return Ok(false);
+        }
+    }
+
+    if zelf.len() == other.len() {
+        Ok(true)
```
youknowone: gt and ge only have this value difference. any idea for refactoring?

HyeockJinKim:
```rust
fn seq_lt_le(
    vm: &VirtualMachine,
    zelf: &dyn SimpleSeq,
    other: &dyn SimpleSeq,
    eq: bool
) -> Result<bool, PyObjectRef> {
    for (a, b) in Iterator::zip(zelf.iter(), other.iter()) {
        if vm.bool_lt(a.clone(), b.clone())? {
            return Ok(true);
        } else if !vm.bool_eq(a.clone(), b.clone())? {
            return Ok(false);
        }
    }

    if zelf.len() == other.len() {
        Ok(eq)
    } else {
        Ok(zelf.len() < other.len())
    }
}

pub fn seq_lt(
    vm: &VirtualMachine,
    zelf: &dyn SimpleSeq,
    other: &dyn SimpleSeq,
) -> Result<bool, PyObjectRef> {
    seq_lt_le(vm, zelf, other, false)
}
pub fn seq_le(
    vm: &VirtualMachine,
    zelf: &dyn SimpleSeq,
    other: &dyn SimpleSeq,
) -> Result<bool, PyObjectRef> {
    seq_lt_le(vm, zelf, other, true)
}
```
What about the code above?

youknowone:
```rust
#[inline]
fn seq_lte(
    vm: &VirtualMachine,
    zelf: &dyn SimpleSeq,
    other: &dyn SimpleSeq,
    eq: bool
) -> Result<bool, PyObjectRef> {
   ...
}
```
Adding `#[inline]` will help the eq parameter to be evaluated in compile time.

Or maybe we even don't need 2 versions of them
```rust
#[inline]
fn seq_cmp(
    vm: &VirtualMachine,
    zelf: &dyn SimpleSeq,
    other: &dyn SimpleSeq,
    obj_cmp: Fn(PyObjectRef, PyObjectRef) -> PyResult<bool>
    len_cmp: Fn(usize, usize) -> bool) {
   ....
}
```

PR #1455, opened 2019-10-02
Add the _sre module

coolreader18:
So it turns out that [`_sre.py`](https://github.com/nikhaldi/_sre.py) actually is fairly compatible with Python 3's `re`. The `re` and `sre_*` python modules are taken from Python 3.0, but I may be able to use the Python 3.6 ones, I haven't tried yet. I kept the Rust `re` module, but I renamed it to `regex_crate`; I'm not sure if that's a good name or what we should do with it in general.

### DIFF
--- a/Lib/_sre.py
+++ b/Lib/_sre.py
@@ -1,0 +1,1284 @@
+# NOT_RPYTHON
+"""
+A pure Python reimplementation of the _sre module from CPython 2.4
+Copyright 2005 Nik Haldimann, licensed under the MIT license
+
+This code is based on material licensed under CNRI's Python 1.6 license and
+copyrighted by: Copyright (c) 1997-2001 by Secret Labs AB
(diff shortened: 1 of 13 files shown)


coolreader18:
Ok, so there are some issues using the up-to-date Python sre modules with `_sre.py`, but we just need to implement a few missing sre opcodes, which is doable. I'll split that off into a separate branch. (The 2 most recent commits are from my work on it).

coolreader18:
I think I'd eventually like to split this off into a separate repository so that other WIP Python implementations can easily get a pure-Python fully functioning `re` implementation.

coolreader18:
Never mind, I was able to use the CPython 3.6 versions, it was a tiny bug in another method that was causing it to fail.

windelbouwman:
Nice work!

PR #1459, opened 2019-10-03
Correctly handle a SystemExit unwound to the top level of execution

coolreader18:


src/main.rs
```diff
@@ -30,14 +32,57 @@ fn main() {
...
+                        }
+                        let s = vm.to_str(&args.elements[0]);
+                        let s = s
+                            .as_ref()
+                            .ok()
+                            .map_or("<element str() failed>", |s| s.as_str());
+                        println!("{}", s);
+                    }
+                }),
+                _ => {
```
youknowone: is this arm same to `print!("{}", vm.to_str(&args))`?

coolreader18: Yep, I think I was trying to emulate code from `exceptions.rs` but it's just the `str` of it I guess.

coolreader18: Actually, it's the repr of the args tuple.

PR #1460, opened 2019-10-04
Don't init() objtuple twice

dralley:


vm/src/types.rs
```diff
@@ -298,7 +298,6 @@ pub fn initialize_types(context: &PyContext) {
     objrange::init(&context);
     objslice::init(&context);
     objsuper::init(&context);
-    objtuple::init(&context);
```
dralley: Other instance on line 283

PR #1461, opened 2019-10-04
Use new style for PyZip impl

dralley:


vm/src/obj/objzip.rs
```diff
@@ -19,16 +19,18 @@ impl PyValue for PyZip {
...
-        .map(|iterable| objiter::get_iter(vm, &iterable))
-        .collect::<Result<Vec<_>, _>>()?;
-    PyZip { iterators }.into_ref_with_type(vm, cls)
-}
-
 #[pyimpl]
 impl PyZip {
+
+    #[pyslot(new)]
+    fn zip_new(cls: PyClassRef, iterables: Args, vm: &VirtualMachine) -> PyResult<PyZipRef> {
```
dralley: Can this be cleaned up further?  I noticed this note

```
/// `Args` optionally accepts a generic type parameter to allow type checks
```
and that a couple of other classes look look like they do this, but I'm not familiar enough with RustPython yet to construct that code.  I tried it and got stuck at the compile errors.

coolreader18: I think it's fine, you could have it be `Args<PyIterable>`, and then add a `iter_obj` method to `PyIterable` to get the object returned from `__iter__`, but I think this is fine if you don't want to do that.

coolreader18: The function name should be `tp_new`

coolreader18:
```suggestion
    fn tp_new(cls: PyClassRef, iterables: Args, vm: &VirtualMachine) -> PyResult<PyZipRef> {
```

dralley: I think I'll leave it for now -- but if you think that would be a worthwhile addition, I could make another PR later?

coolreader18: Yep, feel free if you'd like to.

PR #1469, opened 2019-10-04
Add tab autocompletion to the REPL

coolreader18:


coolreader18:
If anyone is reviewing this, I'd recommend trying out how the autocomplete feels in the repl, I'm not sure specifically about how it doesn't complete builtins on an empty line, only globals, but I liked the effect of it not completing anything on a freshly opened console. It might not be worth it for the inconsistency, though, and how it can complete for a global on an empty line then have more completions after:

```py
>>>>> abc = 1
>>>>> ade = 2
>>>>> <tab>
>>>>> a<tab>
abc ade abs all any ...
```

Maybe it shouldn't complete for builtins at all unless there was nothing found in globals? I'm not sure, I'd appreciate other people's input.

src/main.rs
```diff
@@ -486,17 +486,173 @@ fn shell_exec(vm: &VirtualMachine, source: &str, scope: Scope) -> ShellExecResul
     }
 }
 
+struct ShellHelper<'a> {
+    vm: &'a VirtualMachine,
+    scope: Scope,
+}
+
+impl ShellHelper<'_> {
+    fn complete_opt(&self, line: &str) -> Option<(usize, Vec<String>)> {
```
windelbouwman: I would suggest to reduce the size of this function. Try to create functions that can be displayed without having to scroll.

src/main.rs
```diff
@@ -486,17 +486,173 @@ fn shell_exec(vm: &VirtualMachine, source: &str, scope: Scope) -> ShellExecResul
...
+        let mut words = vec![String::new()];
+        fn revlastword(words: &mut Vec<String>) {
+            let word = words.last_mut().unwrap();
+            let revword = word.chars().rev().collect();
+            *word = revword;
+        }
+        let mut startpos = 0;
+        for (i, c) in line.chars().rev().enumerate() {
```
windelbouwman: This for loop is maybe a good candidate to put into a function called `split_line_on_dots` or something alike.

src/main.rs
```diff
@@ -486,17 +486,173 @@ fn shell_exec(vm: &VirtualMachine, source: &str, scope: Scope) -> ShellExecResul
...
+        let str_iter = |obj| {
+            PyIterable::<PyStringRef>::try_from_object(self.vm, obj)
+                .ok()?
+                .iter(self.vm)
+                .ok()
+        };
+
+        type StrIter<'a> = Box<dyn Iterator<Item = PyResult<PyStringRef>> + 'a>;
+
+        let (iter, prefix) = if let Some((last, parents)) = rest.split_last() {
```
windelbouwman: This could be the second function.

src/main.rs
```diff
@@ -486,17 +486,173 @@ fn shell_exec(vm: &VirtualMachine, source: &str, scope: Scope) -> ShellExecResul
...
+                    globals.chain(str_iter(
+                        self.vm
+                            .call_method(&self.vm.builtins, "__dir__", vec![])
+                            .ok()?,
+                    )?),
+                ) as StrIter
+            };
+            (iter, first.as_str())
+        };
+        let completions = iter
```
windelbouwman: This could be a third function named `sort_completions`

src/main.rs
```diff
@@ -486,17 +486,173 @@ fn shell_exec(vm: &VirtualMachine, source: &str, scope: Scope) -> ShellExecResul
...
+    ) -> rustyline::Result<(usize, Vec<String>)> {
+        if pos != line.len() {
+            return Ok((0, vec![]));
+        }
+        Ok(self.complete_opt(line).unwrap_or((0, vec![])))
+    }
+}
+
+impl rustyline::hint::Hinter for ShellHelper<'_> {}
+impl rustyline::highlight::Highlighter for ShellHelper<'_> {}
```
windelbouwman: Oh cool! Does this thing support syntax highlights?

windelbouwman:
Did you compare the behavior with [bpython](https://bpython-interpreter.org/)?

src/main.rs
coolreader18: Yep, it does! It'd probably be pretty easy to plug a Python [`syntect`](https://github.com/trishume/syntect) definition in for this.

coolreader18:
I tried out bpython, and I just realized why we wouldn't want any completions on an empty line; you wouldn't be able to use tab to index.

PR #1473, opened 2019-10-05
Add comparison of slice

HyeockJinKim:
Add gt, ge, lt, le for slice

Issue #1431


HyeockJinKim:
Can I get a review?

Issue #1374, opened 2019-09-14
Infinite recursion cause segfault, not RecursionError


xarus01:
## Code

```
def a():
    a()
a()
```

## Expected
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "<stdin>", line 2, in a
  File "<stdin>", line 2, in a
  File "<stdin>", line 2, in a
  [Previous line repeated 996 more times]
RecursionError: maximum recursion depth exceeded
```

## Result
```
[1]    36945 segmentation fault  RUSTPYTHONPATH=Lib cargo run
```

youknowone:
The given example now raises `RecursionError` (though the message is not compact as CPython)

PR #1477, opened 2019-10-05
Add RecursionError

lazylife7157:
Partial solution of #1374

I made a small change to check recursion depth before push the frame and then raise error instead of run until segmentation fault occurs.

But it seems there is still a problem. RustPython occurs segmentation fault on 700 (debug) ~ 2600 (release) recursive calls while cpython is fine until 15000. This is the reason that I set default limit lower than cpython.

If there is any inappropriate changes, please let me know. Thank you.

### DIFF
--- a/tests/snippets/sysmod.py
+++ b/tests/snippets/sysmod.py
@@ -1,5 +1,7 @@
 import sys
 
+from testutils import assert_raises
+
 print('python executable:', sys.executable)
 print(sys.argv)
 assert sys.argv[0].endswith('.py')
@@ -60,10 +62,5 @@ def recursive_call(n):
 sys.setrecursionlimit(200)
 assert sys.getrecursionlimit() == 200
 
-exc = None
-try:
+with assert_raises(RecursionError):
     recursive_call(300)
-except RecursionError as exc:
-    pass
-
-assert exc is not None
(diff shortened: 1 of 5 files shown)


tests/snippets/sysmod.py
```diff
@@ -49,3 +49,21 @@ def demo(x):
...
+        recursive_call(n - 1)
+
+sys.setrecursionlimit(200)
+assert sys.getrecursionlimit() == 200
+
+exc = None
+try:
+    recursive_call(300)
+except RecursionError as exc:
+    pass
```
coolreader18: You should use `with assert_raises(RecursionError): ...` here.

vm/src/vm.rs
```diff
@@ -65,6 +65,7 @@ pub struct VirtualMachine {
     pub use_tracing: RefCell<bool>,
     pub signal_handlers: RefCell<[PyObjectRef; NSIG]>,
     pub settings: PySettings,
+    pub recursion_limit: RefCell<usize>,
```
coolreader18: This can just be a `Cell<usize>`

lazylife7157: It's much better! Thanks.

PR #1483, opened 2019-10-06
Change something related to exit

ChJR:
Fixed builtin exit() method with SystemExit (can be tested without Lib).
Added os._exit() method and os.EX_* constants


ChJR:
This PR relates to #1459 and #1413.

PR #1487, opened 2019-10-07
Fixed __hash__ of range

### LINKED ISSUE #1482: hash of range object is not deterministic.
## Feature
Currently, hash of range is not deterministic. 
The range's `__hash__` is not implemented, so the object's `__hash__` is called. (it returned id of object)

### Expected result
```python
>>> hash(range(10))
-6299899980521991026
>>> hash(range(10))
-6299899980521991026
>>> a = {}
>>> a[range(10)] = 1
>>> a[range(10)] = 1
>>> a
{range(0, 10): 1}
```

### Actual result
```python
>>>>> hash(range(10))
140482322810896
>>>>> hash(range(10))
140482319976752
>>>>> a = {}
>>>>> a[range(10)] = 1
>>>>> a[range(10)] = 1
>>>>> a
{range(0, 10): 1, range(0, 10): 1}
```

HyeockJinKim:
Calculate hash value of range through hash of tuple

Fixes #1482

### DIFF
--- a/vm/src/obj/objrange.rs
+++ b/vm/src/obj/objrange.rs
@@ -15,7 +15,6 @@ use super::objint::{PyInt, PyIntRef};
 use super::objiter;
 use super::objslice::{PySlice, PySliceRef};
 use super::objtype::{self, PyClassRef};
-use crate::obj::objtuple::PyTuple;
 
 /// range(stop) -> range object
 /// range(start, stop[, step]) -> range object
@@ -77,18 +76,13 @@ impl PyRange {
     #[inline]
     pub fn get(&self, index: &BigInt) -> Option<BigInt> {
         let start = self.start.as_bigint();
-        let stop = self.stop.as_bigint();
         let step = self.step.as_bigint();
         let index = index.clone();
         if self.is_empty() {
             return None;
         }
 
-        let length: BigInt = if start < stop {
-            (stop - start - 1) / step + 1
-        } else {
-            (start - stop - 1) / (-step) + 1
-        };
+        let length = self.length();
 
         let index = if index.is_negative() {
             let new_index: BigInt = &length + &index;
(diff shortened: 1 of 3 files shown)


vm/src/obj/objrange.rs
```diff
@@ -402,6 +404,25 @@ impl PyRange {
...
+        let mut c = vm.get_none();
+        if !len.is_zero() {
+            b = zelf.start.clone().into_object();
+            if !len.is_one() {
+                c = zelf.step(vm).into_object();
+            }
+        };
+        let tuple = PyTuple {
+            elements: vec![a, b, c],
+        };
```
HyeockJinKim: I'm not sure if PyTuple creation is the way above.
Is there a better way?

HyeockJinKim:
Can I get a review?

vm/src/obj/objrange.rs
youknowone: looking for `vm.ctx.new_tuple`?

youknowone:
note for reviewer: cpython range hash https://github.com/python/cpython/blob/master/Objects/rangeobject.c#L462-L510

vm/src/obj/objrange.rs
```diff
@@ -402,6 +404,27 @@ impl PyRange {
         }
     }
 
+    #[pymethod(name = "__hash__")]
+    fn hash(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<pyhash::PyHash> {
+        let length = zelf.length();
```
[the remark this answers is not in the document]
HyeockJinKim: The code above looks better, so I modified it in the above way.

vm/src/obj/objrange.rs
coolreader18: It looks like PyTuple's hash just uses `pyhash::hash_iter`, so you could probably do the same.

vm/src/obj/objtuple.rs
```diff
@@ -179,7 +179,7 @@ impl PyTuple {
     }
 
     #[pymethod(name = "__hash__")]
-    fn hash(&self, vm: &VirtualMachine) -> PyResult<pyhash::PyHash> {
+    pub fn hash(&self, vm: &VirtualMachine) -> PyResult<pyhash::PyHash> {
```
coolreader18: Oh, and now this doesn't have to be `pub`.

windelbouwman:
Nice!

PR #1489, opened 2019-10-07
Implement .indices(len) of slice

### LINKED ISSUE #1431: Implement slice method

tcmal:
Fixes #1431.

CPython equivalent method: https://github.com/python/cpython/blob/master/Objects/sliceobject.c#L373

### DIFF
--- a/vm/src/obj/objrange.rs
+++ b/vm/src/obj/objrange.rs
@@ -339,77 +339,19 @@ impl PyRange {
     fn getitem(&self, subscript: RangeIndex, vm: &VirtualMachine) -> PyResult {
         match subscript {
             RangeIndex::Slice(slice) => {
-                let range_start = self.start.as_bigint();
-                let range_step = self.step.as_bigint();
-                let range_length = &self.length();
+                let (mut substart, mut substop, mut substep) =
+                    slice.inner_indices(&self.length(), vm)?;
+                let range_step = self.step(vm);
+                let range_start = self.start(vm);
 
-                let substep = if let Some(slice_step) = slice.step_index(vm)? {
-                    if slice_step.is_zero() {
-                        return Err(vm.new_value_error("slice step cannot be zero".to_string()));
-                    }
-                    slice_step
-                } else {
-                    BigInt::one()
-                };
-
-                let negative_step = substep.is_negative();
-                let lower_bound = if negative_step {
-                    -BigInt::one()
-                } else {
-                    BigInt::zero()
-                };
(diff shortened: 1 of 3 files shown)


dralley:
Since the code here uses the same algorithm, we should call this function from here as well.  https://github.com/RustPython/RustPython/blob/master/vm/src/obj/objrange.rs#L317-L383

vm/src/obj/objrange.rs
```diff
@@ -319,78 +319,12 @@ impl PyRange {
...
-                        upper_bound.clone()
-                    } else {
-                        slice_stop.clone()
-                    }
-                } else if negative_step {
-                    lower_bound.clone()
-                } else {
-                    upper_bound.clone()
-                };
-
```
dralley: The 3 lines below do need to remain here, though.

dralley: The function is equivalent to this CPython code (but the variables are named opposite) https://github.com/python/cpython/blob/master/Objects/rangeobject.c#L306-L322

windelbouwman:
@tcmal could you add a python test snippet with an example of the problem you fixed? This would help to prevent the issue from re-appearing.

tests/snippets/attr.py
```diff
@@ -4,12 +4,21 @@
 class A:
     pass
 
+class B:
+    x = 50
```
windelbouwman: This looks like code from another pull request, is it related to this change, or is it due to git merging?

tcmal: Yeah messed up the rebase

tcmal:
Added tests for slice, mostly based off [CPython's code](https://github.com/python/cpython/blob/master/Lib/test/test_slice.py).

dralley:
Is this waiting on anything?

dralley:
@windelbouwman

PR #1491, opened 2019-10-07
Compress bytecode using LZ4

windelbouwman:


windelbouwman:
This will achieve:
- python bytecode reduction (less disk space used)
- rustpython binary size (since it contains pre-compiled bytecode)
- Increase startup time (since the binary is smaller)
- Reduce wasm size (for web)

windelbouwman:
I did a comparison on the wasm build, using this command:

    $ npm run dist

Then I checked this file: `target/wasm32-unknown-unknown/release/rustpython_wasm.wasm`

- Master branch: 25708312 (25M)
- This branch: 16683050 (16M)

coolreader18:
Wow, that's a lot better!

PR #1492, opened 2019-10-07
WASI integration

syrusakbary:
This PRs adds support for WASI in the standalone `rustpython`.

Thanks to that, we were able to generate the module `rustpython.wasm` that can run standalone with [Wasmer](https://wasmer.io).

We also published the package into WAPM, so people can start using RustPython very easily anywhere (without having Cargo/Rust!).
https://wapm.io/package/rustpython

> Note: the `rustpython.wasm` module can be downloaded from the WAPM, page (in the "Browse modules" section)

If you want to test locally, you can do:

```shell
wapm install rustpython
wapm run rustpython # This might take ~20 seconds to compile and run
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -321,7 +321,11 @@ fn write_profile(matches: &ArgMatches) -> Result<(), Box<dyn std::error::Error>>
 }
 
 fn run_rustpython(vm: &VirtualMachine, matches: &ArgMatches) -> PyResult<()> {
+    #[cfg(not(target_os = "wasi"))]
     import::init_importlib(&vm, true)?;
+
+    #[cfg(target_os = "wasi")]
+    import::init_importlib(&vm, false)?;
 
     if let Some(paths) = option_env!("BUILDTIME_RUSTPYTHONPATH") {
         let sys_path = vm.get_attribute(vm.sys_module.clone(), "path")?;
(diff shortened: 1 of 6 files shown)


src/main.rs
```diff
@@ -321,8 +321,12 @@ fn write_profile(matches: &ArgMatches) -> Result<(), Box<dyn std::error::Error>>
 }
 
 fn run_rustpython(vm: &VirtualMachine, matches: &ArgMatches) -> PyResult<()> {
+    #[cfg(not(target_os = "wasi"))]
     import::init_importlib(&vm, true)?;
 
+    #[cfg(target_os = "wasi")]
+    import::init_importlib(&vm, false)?;
```
coolreader18: You can use the `cfg` macro instead of the attribute here: `import::init_importlib(&vm, cfg!(not(target_os = "wasi")))`

coolreader18:
Is there any way that we could manage the `rustpython` package on wapm ourselves?

syrusakbary:
Just addressed the feedback

> Is there any way that we could manage the rustpython package on wapm ourselves?
Of course! You can provide here your wapm usernames and I'll make you an admin of the package :)

coolreader18:
Alright, I just made an account, my username on wapm.io is also `coolreader18`.

syrusakbary:
Awesome, I just added you as admin of `coolreader18`, you should be able to publish new versions to wapm:

```
# First, let's build the binary
cargo build --release --target wasm32-wasi --features="freeze-stdlib"

# Then, publish to wapm
wapm publish
```

Issue #1404, opened 2019-09-22
An error occurs when deleting an attribute of a class.


rls1004:
### Expected Result
```
>>> class C:
...   x = 10
... 
>>> del C.x
>>> C.x
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: type object 'C' has no attribute 'x'
```

### Actual Result
```
>>>>> class C:
.....   x = 10
..... 
>>>>> del C.x
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'type' object has no attribute 'x'
>>>>> C.x
10
```
Can't delete attribute of class.

PR #1496, opened 2019-10-09
Implement delattr for type object

rls1004:
Implement delattr for type object

### Before
```
>>>>> class C:
.....   x = 10
..... 
>>>>> del C.x
AttributeError: 'type' object has no attribute 'x'
>>>>> C.x
10
```

### After
```
>>>>> class C:
.....   x = 10
.....
>>>>> del C.x
>>>>> C.x
AttributeError: C has no attribute 'x'
```

Fixed: #1404

### DIFF
--- a/tests/snippets/attr.py
+++ b/tests/snippets/attr.py
@@ -4,11 +4,20 @@ from testutils import assert_raises
 class A:
     pass
 
+class B:
+    x = 50
 
 a = A()
 a.b = 10
 assert hasattr(a, 'b')
 assert a.b == 10
+
+assert B.x == 50
+
+# test delete class attribute with del keyword
+del B.x
+with assert_raises(AttributeError):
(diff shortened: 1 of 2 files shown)


tests/snippets/attr.py
```diff
@@ -10,6 +10,15 @@ class A:
...
 
+a.c = 50
+assert hasattr(a, 'c')
+assert a.c == 50
+
+# test delete attribute with del keyword
+del a.c
+with assert_raises(AttributeError):
+    _ = a.c
+
```
windelbouwman: Could you add the snippet the from the explanation in the pull request as well? Class attributes probably take another code path as instance attributes.

I mean this snippet:
```python
class C:
    x = 10

...
```

rls1004: Thank you, I added it.

PR #1497, opened 2019-10-09
Add __ne__ methods

HyeockJinKim:
Add __ne__ for float, none and range

### DIFF
--- a/tests/snippets/floats.py
+++ b/tests/snippets/floats.py
@@ -229,3 +229,34 @@ assert float('nan').hex() == 'nan'
 # Test float exponent:
 assert 1 if 1else 0 == 1
 
+a = 3.
+assert a.__eq__(3) is True
+assert a.__eq__(3.) is True
+assert a.__eq__(3.00000) is True
+assert a.__eq__(3.01) is False
+
+pi = 3.14
+assert pi.__eq__(3.14) is True
+assert pi.__ne__(3.14) is False
+assert pi.__eq__(3) is False
+assert pi.__ne__(3) is True
+assert pi.__eq__('pi') is NotImplemented
+assert pi.__ne__('pi') is NotImplemented
+
+assert pi.__eq__(float('inf')) is False
+assert pi.__ne__(float('inf')) is True
+assert float('inf').__eq__(pi) is False
+assert float('inf').__ne__(pi) is True
+assert float('inf').__eq__(float('inf')) is True
+assert float('inf').__ne__(float('inf')) is False
+assert float('inf').__eq__(float('nan')) is False
+assert float('inf').__ne__(float('nan')) is True
+
+assert pi.__eq__(float('nan')) is False
+assert pi.__ne__(float('nan')) is True
+assert float('nan').__eq__(pi) is False
+assert float('nan').__ne__(pi) is True
+assert float('nan').__eq__(float('nan')) is False
+assert float('nan').__ne__(float('nan')) is True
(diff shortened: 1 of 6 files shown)


tests/snippets/floats.py
```diff
@@ -229,3 +229,10 @@
 # Test float exponent:
 assert 1 if 1else 0 == 1
 
+pi = 3.14
+assert pi.__eq__(3.14)
+assert pi.__ne__(3.14) == False
+assert pi.__eq__(3) == False
+assert pi.__ne__(3)
+assert pi.__eq__('pi') == NotImplemented
+assert pi.__ne__('pi') == NotImplemented
```
youknowone: Because float is not fully ordered, please add `nan` and `inf` comparisons to ensure its completeness

vm/src/obj/objfloat.rs
```diff
@@ -179,20 +179,39 @@ impl PyFloat {
...
 
-            if let (Some(self_int), Some(other_float)) = (value.to_bigint(), other_int.to_f64()) {
-                value == other_float && self_int == *other_int
-            } else {
-                false
-            }
+    #[pymethod(name = "__ne__")]
+    fn ne(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
+        let result = if objtype::isinstance(&other, &vm.ctx.float_type()) {
+            !self.float_eq(other)
```
youknowone: about the tests, i am suspicious here because `float.__ne__()` is not nessassarily `!float.__eq__()`

vm/src/obj/objnone.rs
```diff
@@ -111,6 +111,24 @@ impl PyNoneRef {
             Err(vm.new_attribute_error(format!("{} has no attribute '{}'", self.as_object(), name)))
         }
     }
+
+    #[pymethod(name = "__eq__")]
+    fn eq(self, rhs: PyObjectRef, vm: &VirtualMachine) -> PyResult {
```
youknowone: When returning value doesn't have `Err` case, return type simply can be `PyObjectRef` instead of `PyResult`

vm/src/obj/objrange.rs
```diff
@@ -239,30 +239,43 @@ impl PyRange {
         }
     }
 
-    #[pymethod(name = "__eq__")]
-    fn eq(&self, rhs: PyObjectRef, vm: &VirtualMachine) -> bool {
-        if objtype::isinstance(&rhs, &vm.ctx.range_type()) {
-            let rhs = get_value(&rhs);
+    fn inner_eq(&self, rhs: &PyRange) -> PyResult<bool> {
```
youknowone: Can the return type be just `bool`?

vm/src/obj/objrange.rs
```diff
@@ -239,30 +239,43 @@ impl PyRange {
...
-            let step = self.step.as_bigint();
-            if step.is_one() || step == rhs.step.as_bigint() {
-                return true;
-            }
+        Ok(false)
+    }
 
-            false
+    #[pymethod(name = "__eq__")]
+    fn eq(&self, rhs: PyObjectRef, vm: &VirtualMachine) -> PyResult {
```
youknowone: Adding tests about this change will be great

vm/src/obj/objfloat.rs
HyeockJinKim: I think `eq` and `ne`are different from `lt` or `gt`.
Even if the value is `nan` or` inf`, `ne` is `!eq`.

PR #1500, opened 2019-10-09
Refactor sequence item comparison

youknowone:


vm/src/obj/objsequence.rs
```diff
@@ -285,12 +284,11 @@ pub fn seq_equal(
...
     other: &dyn SimpleSeq,
-) -> Result<bool, PyObjectRef> {
+) -> PyResult<bool> {
     if zelf.len() == other.len() {
         for (a, b) in Iterator::zip(zelf.iter(), other.iter()) {
             if !a.is(b) {
-                let eq = vm._eq(a.clone(), b.clone())?;
-                let value = objbool::boolval(vm, eq)?;
+                let value = vm.bool_eq(a.clone(), b.clone())?;
                 if !value {
```
coolreader18: You could probably put `!vm.bool_eq...` right in the `if` condition.

PR #1501, opened 2019-10-10
introduce and adapt vm.identical_or_equal to test `is` or else `__eq__`

youknowone:


windelbouwman (approved):
This certainly makes a lot of sense! Good simplification.

vm/src/vm.rs
```diff
@@ -1270,11 +1269,19 @@ impl VirtualMachine {
     }
 
     pub fn bool_eq(&self, a: PyObjectRef, b: PyObjectRef) -> PyResult<bool> {
-        let eq = self._eq(a.clone(), b.clone())?;
+        let eq = self._eq(a, b)?;
         let value = objbool::boolval(self, eq)?;
         Ok(value)
     }
 
+    pub fn bool_equal(&self, a: &PyObjectRef, b: &PyObjectRef) -> PyResult<bool> {
```
windelbouwman: Minor note: this function is confusing with `bool_eq` above. The difference is only the extra check with `is`. Maybe rename into `identical_or_equal`?

youknowone: Fixed it

PR #1502, opened 2019-10-10
Small int cache like CPython

youknowone:


vm/src/pyobject.rs
```diff
@@ -364,7 +373,15 @@ impl PyContext {
         self.types.object_type.clone()
     }
 
+    #[inline]
     pub fn new_int<T: Into<BigInt>>(&self, i: T) -> PyObjectRef {
+        let i = i.into();
```
coolreader18: I was thinking that instead of immediately converting it, `T` could be `Into<BigInt> + PartialOrd<i32>`, to save on the overhead of allocation for a BigInt.

youknowone: Because `PartialOrd<i32>` is not implemented for unsigned ints, this way will be tricky

coolreader18: Maybe `T` could be `ToPrimitive`?

youknowone: This is great idea. I was struggling with TryInto and TryFrom but ToPrimitive is a lot easier to do this. Thanks!

PR #1503, opened 2019-10-10
Allow antigravity to be imported

coolreader18:


windelbouwman:
This truly is good news! :)

PR #1505, opened 2019-10-10
Defer BigInt::clone() for new_int

youknowone:
This change will reduce useless `BigInt::clone()` for a few cases.

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -384,6 +384,17 @@ impl PyContext {
         PyObject::new(PyInt::new(i), self.int_type(), None)
     }
 
+    #[inline]
+    pub fn new_bigint(&self, i: &BigInt) -> PyObjectRef {
+        if let Some(i) = i.to_i32() {
+            if i >= INT_CACHE_POOL_MIN && i <= INT_CACHE_POOL_MAX {
+                let inner_idx = (i - INT_CACHE_POOL_MIN) as usize;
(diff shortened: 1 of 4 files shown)


vm/src/pyobject.rs
```diff
@@ -384,6 +384,17 @@ impl PyContext {
         PyObject::new(PyInt::new(i), self.int_type(), None)
     }
 
+    #[inline]
+    pub fn new_bigint(&self, i: &BigInt) -> PyObjectRef {
```
youknowone: @coolreader18 Do you have any idea to integrate this function into new_int?

coolreader18:
Sorry for merging #1502 too early, I guess you were still working on it.

youknowone:
That is ok. Actually, i thought it was done too. I am now doing a little refactoring and found related issues.

vm/src/pyobject.rs
coolreader18: You could maybe declare your own trait and implement it for both `T where T: Into<BigInt> + ToPrimitive` and `&T where T: Into<BigInt> + ToPrimitive`. I think the real issue is that `ToPrimitive` should also be implemented by `num-traits` for `&T where T: ToPrimitive`.

youknowone: For now, I think I am not interested in working for trait puzzle.
Would you review this PR as is again please?

coolreader18 (approved):

PR #1506, opened 2019-10-10
Refactor function signatures

youknowone:
- Result<T, PyObjectRef> -> PyResult<T>
- Make pymethod return types as small as possible

### DIFF
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -9,8 +9,12 @@ use std::str;
 
 use num_bigint::Sign;
 use num_traits::{Signed, ToPrimitive, Zero};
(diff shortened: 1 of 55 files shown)


windelbouwman:
Wow, this is really nice, but also a lot of code. Did you do this manually or automated? Would it be possible to split this work into several pull requests?

youknowone:
Mostly manually. I will try to split it

windelbouwman:
If you can split it somehow, this would be good, otherwise no worries, I will look into it, but it will take me some time to view it all.

windelbouwman:
Super! Nice job!

youknowone:
Oh, it is merged without spliting, even more great :D

PR #1507, opened 2019-10-10
print_exception can print to places that aren't stdout

FauxFaux:
This should allow me to .. show users errors.

I've left `print_exception` ignoring the error, as this is what `println!` does. I'd be happy for it to return the error.

### DIFF
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -8,7 +8,7 @@ use crate::types::create_type;
 use crate::vm::VirtualMachine;
 use itertools::Itertools;
 use std::fs::File;
-use std::io::{BufRead, BufReader};
+use std::io::{self, BufRead, BufReader, Write};
 
 fn exception_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
     let exc_self = args.args[0].clone();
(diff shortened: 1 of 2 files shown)


vm/src/exceptions.rs
```diff
@@ -25,32 +25,50 @@ fn exception_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
 
 /// Print exception chain
 pub fn print_exception(vm: &VirtualMachine, exc: &PyObjectRef) {
+    let _ = write_exception(io::stdout(), vm, exc);
+}
+
+pub fn write_exception<W: Write>(
+    mut output: W,
+    vm: &VirtualMachine,
+    exc: &PyObjectRef,
```
coolreader18: This could be `io::Result<()>`

vm/src/exceptions.rs
```diff
@@ -25,32 +25,50 @@ fn exception_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
     }
-    print_exception_inner(vm, exc)
+    print_exception_inner(output, vm, exc)
 }
 
-fn print_source_line(filename: &str, lineno: usize) {
+fn print_source_line<W: Write>(
+    mut output: W,
+    filename: &str,
+    lineno: usize,
```
coolreader18: And here

vm/src/exceptions.rs
```diff
@@ -59,15 +77,21 @@ fn print_source_line(filename: &str, lineno: usize) {
...
     }
+
+    Ok(())
 }
 
 /// Print exception occurrence location from traceback element
-fn print_traceback_entry(vm: &VirtualMachine, tb_entry: &PyObjectRef) {
+fn print_traceback_entry<W: Write>(
+    mut output: W,
+    vm: &VirtualMachine,
```
coolreader18: And here, etc

FauxFaux:
I switched to the `io::Result` alias. I left it out on purpose (I don't like those Result typedefs), but we may as well be consistent!

coolreader18:
Thanks for contributing!

PR #1508, opened 2019-10-10
locals() return copy of locals dictionary

### LINKED ISSUE #1356: locals() dict share actual local variable
## Feature

```python
def a(d):
    locals()['d'] = 3
    b = 3 + d
    print(b)

a(21)
```
The code above generates the following error:
(locals dict share actual local variable)

### Actual Result
```bash
Finished dev [unoptimized + debuginfo] target(s) in 0.11s
Running `target/debug/rustpython c.py`
6
```
### Expected Result
```bash
24
```

HyeockJinKim:
Return copy of locals so that local variables
does not change even if locals() values is changed

Fixes #1356

### DIFF
--- a/tests/snippets/global_nonlocal.py
+++ b/tests/snippets/global_nonlocal.py
@@ -85,3 +85,27 @@ with assert_raises(SyntaxError):
 #     nonlocal c
 #     c = 2
 
+def a():
+    x = 0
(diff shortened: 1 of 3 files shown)


windelbouwman:
Nicely done!

PR #1514, opened 2019-10-12
`sys.std{in,out,err}` implemented in the most simple way

alanjds:
Opens the files on initialization and attaches to `sys.std{in,out,err}`
- It respects what already exists if any
- Otherwise it opens de UNIX standard files

Caveats:
- May be not on the proper place
- Is open to discussion on what codec, or if should be opened as bytes
- Can be also done in Rust side

### DIFF
--- a/Lib/_sitebuiltins.py
+++ b/Lib/_sitebuiltins.py
@@ -12,8 +12,8 @@ import os
 import sys
 
 sys.stdin = sys.__stdin__ = getattr(sys, '__stdin__', False) or os.fdopen(0, "r")
-sys.stdout = sys.__stdin__ = getattr(sys, '__stdout__', False) or os.fdopen(1, "w")
-sys.stderr = sys.__stdin__ = getattr(sys, '__stderr__', False) or os.fdopen(2, "w")
+sys.stdout = sys.__stdout__ = getattr(sys, '__stdout__', False) or os.fdopen(1, "w")
+sys.stderr = sys.__stderr__ = getattr(sys, '__stderr__', False) or os.fdopen(2, "w")
 
 
 class Quitter(object):
--- a/Lib/_sitebuiltins.py
+++ b/Lib/_sitebuiltins.py
@@ -10,6 +10,11 @@ The objects used by the site module to add custom builtins.
 
 import sys
 
+sys.stdin = getattr(sys, 'stdin', False) or open('/dev/stdin')
+sys.stdout = getattr(sys, 'stdout', False) or open('/dev/stdout', 'w')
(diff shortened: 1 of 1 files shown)


alanjds:
cc @janrg

youknowone:
Interesting. Is this working for both windows and unix?

coolreader18:
You could do something like this for cross-platform (I think):

```python
stdin = ... or os.fdopen(0)
stdout = ... or os.fdopen(1)
stderr = ... or os.fdopen(2)
```

coolreader18:
Actually, if it's that easy to do, it might be worth just putting it in `sys.rs`

windelbouwman:
We could also implement a wrapper type around this function: https://doc.rust-lang.org/std/io/fn.stderr.html

alanjds:
> Interesting. Is this working for both windows and unix?

I bet not. Unless under WSL.

alanjds:
> Actually, if it's that easy to do, it might be worth just putting it in sys.rs

Hum... If I do this, people will still be able to override in runtime? This ability is a need I guess

windelbouwman:
> Hum... If I do this, people will still be able to override in runtime? This ability is a need I guess

Yes, this will remain possible. You could define the `sys.stderr` attribute in `sysmodule.rs`. This would then be the initial value, and in python people still can set `sys.stderr` afterwards.

alanjds:
Changed to the `os.fdopen` strategy as suggested by @coolreader18.

> Actually, if it's that easy to do, it might be worth just putting it in `sys.rs`

Yes. For now I will let it be as is and go work on other issue (Am working on the import of `zipfile`).
I don't see a _need_ to move to `sys.rs` and yet am not against it.

Lib/_sitebuiltins.py
```diff
@@ -8,8 +8,14 @@
 # Note this means this module should also avoid keep things alive in its
 # globals.
 
+import os
 import sys
 
+sys.stdin = sys.__stdin__ = getattr(sys, '__stdin__', False) or os.fdopen(0, "r")
+sys.stdout = sys.__stdin__ = getattr(sys, '__stdout__', False) or os.fdopen(1, "w")
+sys.stderr = sys.__stdin__ = getattr(sys, '__stderr__', False) or os.fdopen(2, "w")
```
coolreader18: `sys.__stdin__` is being set in all of these, rather than `sys.__std{in,out,err}__`

coolreader18:
@alanjds are you still working on this? I think there's a typo, but otherwise it's good to merge.

alanjds:
Yeah, thanks for noticing it. I am on a trip and forgot to push the fix for the typo. Expect to push until tomorrow.

PR #1517, opened 2019-10-13
Add csv.reader options

kluid:
Added kwarg parameters of csv.reader 'delimiter' and 'quotechar'

### DIFF
--- a/tests/snippets/test_csv.py
+++ b/tests/snippets/test_csv.py
@@ -1,3 +1,5 @@
+from testutils import assert_raises
+
 import csv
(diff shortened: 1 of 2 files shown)


tests/snippets/test_csv.py
```diff
@@ -21,3 +23,23 @@ def f():
 	assert six == 'six'
 
 f()
+
+def test_delim():
+	iter = ['one|two|three', 'four|five|six']
```
youknowone: hard tab is used here

PR #1518, opened 2019-10-13
Add repeat count support to struct.pack()

### LINKED ISSUE #1397: struct.pack should support repeat count
`struct.pack('3B', 65, 66, 67)` should return `b'ABC'`. Currently it raises `ValueError: Illegal format code '3'`.

ChJR:
This PR will fix #1397.


vm/src/stdlib/pystruct.rs
```diff
@@ -88,10 +88,29 @@ where
     I: Sized + Iterator<Item = char>,
 {
     let mut codes = vec![];
+    let mut use_repeat = false;
```
windelbouwman: Idea:

Instead of two mutable variables, rewrite the loop as following:

```rust
while chars.peek().is_some() {
    // determine repeat operator:
    let repeat = match chars.peek() {
        '0' .. '9' => {
           let mut repeat = 0;
           while let Some('0' .. '9') = chars.peek()  {
               let current_digit = c.to_digit(10).unwrap();
               repeat = repeat * 10 + current_digit;
           }
           Some(repeat)
        },
        _ => None,
    };

    // determine format char:
    let c = chars.next();
    match c {
      Some('b' | 'B' | ... ) => {
          if let Some(repeat) = repeat {
             // emit format char repeated
          } else {
             // emit format char normally
          }
      },
      _ => return illegal format code
   }
}
```

ChJR: Nice approach. I applied this code to the PR.

vm/src/stdlib/pystruct.rs
```diff
@@ -88,13 +88,35 @@ where
...
+                }
+                Some(repeat)
             }
-            c => {
-                return Err(format!("Illegal format code {:?}", c));
+            _ => None,
+        };
+
+        // determine format char:
+        if let Some(c) = chars.next() {
```
windelbouwman: The else arm of this if should return an error.

windelbouwman: Does this work?
```rust
if let Some('b' | 'B' | ...) = chars.next() {
    ...
} else {
    // massive error
}
```

ChJR: There are syntax errors in your suggestion. But I got inspiration from that.

PR #1519, opened 2019-10-13
Change rustc version in markdown

ChJR:
~This PR is mutually exclusive with #1516.
That means if this PR is merged #1516 is needed to be dropped. On the other hand, if #1516 is merged this PR is needed to be dropped too.~

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -19,7 +19,7 @@ The contents of the Development Guide include:
 
 RustPython requires the following:
 
-- Rust 1.36 or higher
+- Rust latest stable version (at least 1.38.0)
     - To check Rust version: `rustc --version` 
     - If you have `rustup` on your system, enter to update to the latest
       stable version: `rustup update stable`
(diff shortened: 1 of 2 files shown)


DEVELOPMENT.md
```diff
@@ -19,7 +19,7 @@ The contents of the Development Guide include:
 
 RustPython requires the following:
 
-- Rust 1.36 or higher
+- Rust latest stable version (at least 1.38.0)
```
windelbouwman: We could say that we require at least the stable version, without mentioning the 1.38 version, since the stable version will move forward in time. What do you think?

youknowone: That will be the most correct description. But still my rule of thumb is that people understand the version easier with specific number. Maybe ‘(e.g 1.38.0 in oct 2019)’ can be a helpful comment

windelbouwman: Agreed, having an explicit version number is useful.

PR #1521, opened 2019-10-13
Implement round(number, digits) for digits != 0

janrg:
For most cases uses rust's round function to round via:
`round(number, decimals) = round(number * 10^decimals) / 10^decimals`
* skips rounding if rounding would not affect digits within double precision (also covers cases where the above implementation would overflow the f64 exponent)
* also handles arbitrarily negative values for `decimals`


youknowone:
Thanks for contributing! Will you add some test cases about new code? There are python test snippets about float in `tests/snippets/floats.py`.

janrg:
Will do. Might be a few days though as I won't be home before wednesday

janrg:
Found some time to write a few tests and also noticed a range of input values where it wasn't working properly yet. Fixed.

PR #1524, opened 2019-10-13
Add os.system

dan-fritchman:
* Add `os_system` and accompanying Python attribute `os.system` to `os.rs`
* Tested/ working on *Nix (MacOS) only.  Not tested on Windows. 
* Accepts Python `str` argument, but not `bytes or os.PathLike object`

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -943,6 +943,17 @@ fn os_chdir(path: PyStringRef, vm: &VirtualMachine) -> PyResult<()> {
 }
 
 #[cfg(unix)]
+fn os_system(command: PyStringRef, vm: &VirtualMachine) -> PyResult<i32> {
+    use libc::system;
+    use std::ffi::CString;
+
+    let rstr = command.as_str();
+    let cstr = CString::new(rstr).unwrap();
+    let x = unsafe { system(cstr.as_ptr()) };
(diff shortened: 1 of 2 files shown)


youknowone:
Thanks for contributing! Will you add some test cases about new code? There are python test snippets about os module in `tests/snippets/stdlib_os.py`.

vm/src/stdlib/os.rs
```diff
@@ -1292,6 +1303,7 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
 fn extend_module_platform_specific(vm: &VirtualMachine, module: PyObjectRef) -> PyObjectRef {
     let ctx = &vm.ctx;
     extend_module!(vm, module, {
+        "system" => ctx.new_rustfunc(os_system),
```
windelbouwman: I usually keep these type of lists on alphabetical order.

dan-fritchman:
Thanks for the notes.

* Adding basic unit tests
* Re-ordered the os-module export table, by type (function vs constant) first, and then (more or less) alphabetically
* Linting failed on the CI server, due to the unused `vm` argument.  (https://travis-ci.org/RustPython/RustPython/jobs/597262830)  I believe `os_system` will always be called with a `VirtualMachine` as the last argument, so this is just renamed to `_vm`.

PR #1526, opened 2019-10-13
Implement itertools.product

seeeturtle:
This implements `itertools.product` of standard library.

Related with #1361

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -1,15 +1,16 @@
 use std::cell::{Cell, RefCell};
 use std::cmp::Ordering;
+use std::iter;
 use std::ops::{AddAssign, SubAssign};
 use std::rc::Rc;
 
 use num_bigint::BigInt;
 use num_traits::ToPrimitive;
 
-use crate::function::{OptionalArg, PyFuncArgs};
+use crate::function::{Args, OptionalArg, PyFuncArgs};
 use crate::obj::objbool;
 use crate::obj::objint::{self, PyInt, PyIntRef};
-use crate::obj::objiter::{call_next, get_iter, new_stop_iteration};
+use crate::obj::objiter::{call_next, get_all, get_iter, new_stop_iteration};
 use crate::obj::objtuple::PyTuple;
 use crate::obj::objtype::{self, PyClassRef};
 use crate::pyobject::{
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -736,6 +737,140 @@ impl PyItertoolsTee {
...
+}
+
+#[derive(FromArgs)]
+struct ProductArgs {
+    #[pyarg(keyword_only, optional = true)]
+    repeat: OptionalArg<PyIntRef>,
+}
+
+#[pyimpl]
+impl PyIterToolsProduct {
```
coolreader18: This should be `#[pyslot(new)]` instead of a method `__new__`

vm/src/stdlib/itertools.rs
```diff
@@ -736,6 +737,140 @@ impl PyItertoolsTee {
...
+    #[pymethod(name="__new__")]
+    #[allow(clippy::new_ret_no_self)]
+    fn new(
+        _cls: PyClassRef,
+        iterables: Args<PyObjectRef>,
+        args: ProductArgs,
+        vm: &VirtualMachine,
```
coolreader18: You can do `PyResult<PyRef<Self>>`, call `.into_ref_with_type(vm, cls)` and then you don't need the clippy `allow`

vm/src/stdlib/itertools.rs
```diff
@@ -736,6 +737,143 @@ impl PyItertoolsTee {
...
+
+impl PyValue for PyIterToolsProduct {
+    fn class(vm: &VirtualMachine) -> PyClassRef {
+        vm.class("itertools", "product")
+    }
+}
+
+#[derive(FromArgs)]
+struct ProductArgs {
+    #[pyarg(keyword_only, optional = true)]
```
coolreader18: This can be an `OptionalArg<usize>` to avoid having to check for an overflow yourself.

vm/src/stdlib/itertools.rs
```diff
@@ -736,6 +737,143 @@ impl PyItertoolsTee {
     }
 }
 
+#[pyclass]
+#[derive(Debug)]
+struct PyIterToolsProduct {
+    pools: Vec<Vec<PyObjectRef>>,
+    idxs: RefCell<Vec<usize>>,
+    cur: RefCell<usize>,
```
coolreader18: Is this separate `sizes` Vec necessary? It seems like it's just the length of the Vec at the same index in `pools`.

vm/src/stdlib/itertools.rs
```diff
@@ -736,6 +737,143 @@ impl PyItertoolsTee {
     }
 }
 
+#[pyclass]
+#[derive(Debug)]
+struct PyIterToolsProduct {
+    pools: Vec<Vec<PyObjectRef>>,
+    idxs: RefCell<Vec<usize>>,
```
coolreader18: `cur` and `stop` can be `Cell<_>`s, as both their inner types are primitives.

vm/src/stdlib/itertools.rs
seeeturtle: Well, it can be implemented without separate `sizes` but i don't know what's the benefit of not separating. Is there special reason, @coolreader18 ?

vm/src/stdlib/itertools.rs
```diff
@@ -736,6 +737,123 @@ impl PyItertoolsTee {
...
+#[derive(FromArgs)]
+struct ProductArgs {
+    #[pyarg(keyword_only, optional = true)]
+    repeat: OptionalArg<usize>,
+}
+
+#[pyimpl]
+impl PyIterToolsProduct {
+    #[pyslot(new)]
+    fn new(
```
youknowone: `tp_new` is the conventional name of this method

vm/src/stdlib/itertools.rs
youknowone: Can we use `OptionalArg<usize>` directly into `product::tp_new`?

seeeturtle:
> Can we use OptionalArg<usize> directly into product::tp_new?
Well, it passes the snippet tests so I think yes.
@youknowone

windelbouwman:
I think this looks good now! Thanks!

PR #1528, opened 2019-10-14
Fix module attr error to use python `__name__` instead of rust struct `name`

youknowone:


vm/src/pyobject.rs
```diff
@@ -85,9 +84,6 @@ impl fmt::Display for PyObject<dyn PyObjectPayload> {
             }
         }
 
-        if let Some(PyModule { ref name, .. }) = self.payload::<PyModule>() {
```
youknowone: is there any code depending on here? it is not possible to keep compatibility here without name field.

coolreader18:
Why not just have a r/w property that gets/modifies the `name` struct field?

youknowone:
How about delete in that case? I think it is not compatible with CPython.

vm/src/pyobject.rs
windelbouwman: We should use payload as little as possible outside of the object itself, so this removal seems good to me!

youknowone: I confirmed current test code doesn't hit this line

PR #1533, opened 2019-10-14
Clean up the REPL and CodeFlags

coolreader18:
I changed `FunctionOpArg` to `CodeFlags`, which is basically the same but is stored inside `CodeObject`. This also means that we don't have to have `enter_function` return the flags for `MakeFunction`; it can just apply them to the `CodeObject` it pushes onto the stack.

### DIFF
--- a/bytecode/src/bytecode.rs
+++ b/bytecode/src/bytecode.rs
@@ -353,6 +353,7 @@ pub enum BlockType {
 */
 
 impl CodeObject {
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         flags: CodeFlags,
         arg_names: Vec<String>,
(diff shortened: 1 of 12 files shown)


bytecode/src/bytecode.rs
```diff
@@ -176,9 +177,7 @@ pub enum Instruction {
     JumpIfFalseOrPop {
         target: Label,
     },
-    MakeFunction {
-        flags: FunctionOpArg,
```
windelbouwman: This surprises me, don't we need a codeflags when creating the function? Or are the flags already packed in the codeobject itself before this happens?

coolreader18: They're already in the `CodeObject` that `MakeFunction` is operating on.

windelbouwman: Interesting, this looks good!

PR #1534, opened 2019-10-14
Fix the `yield from` expression

coolreader18:
Now that this is working, I'm planning on implementing `async def` coroutines :)


coolreader18:
Could I get a review from someone? I've already got async/await working locally, and I'd like to get this merged before opening a PR for that.

windelbouwman:
Now I'm curious to how you did the async/await thingy. Nice job!

PR #1537, opened 2019-10-15
Change os functions to new arg style

palaviv:


vm/src/stdlib/os.rs
```diff
@@ -83,17 +83,11 @@ fn make_path(_vm: &VirtualMachine, path: PyStringRef, dir_fd: &DirFd) -> PyStrin
     }
 }
 
-pub fn os_close(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
-    arg_check!(vm, args, required = [(fileno, Some(vm.ctx.int_type()))]);
-
-    let raw_fileno = objint::get_value(&fileno);
-
```
coolreader18: This could be an `i64` now

vm/src/stdlib/os.rs
```diff
@@ -83,17 +83,11 @@ fn make_path(_vm: &VirtualMachine, path: PyStringRef, dir_fd: &DirFd) -> PyStrin
...
+fn os_close(fileno: PyIntRef, _vm: &VirtualMachine) {
     //The File type automatically closes when it goes out of scope.
     //To enable us to close these file descriptors (and hence prevent leaks)
     //we seek to create the relevant File and simply let it pass out of scope!
-    rust_file(raw_fileno.to_i64().unwrap());
-
-    Ok(vm.get_none())
```
coolreader18: I know that this was here before the change, but using `drop(file)` is more explicit as to what's happening.

palaviv: Better :)

palaviv: I think that the drop don't work as expected on windows. I will try to see what happens when I revert it.

PR #1538, opened 2019-10-15
Convert platform to new arg style

palaviv:


windelbouwman:
Ahh, cleaning up, sweet code reductions!

Issue #304, opened 2019-02-04
Need to implement int.* 


veera83372:
Following methods/attributes need to be implemented on type `int`. Please feel to pick unchecked method/attributes
`int` is  implemented in `vm/src/obj/objint.rs`

- [x] int . \_\_bool__
- [x] int . \_\_ceil__
- [x] int . \_\_dir__
- [x] int . \_\_floor__
- [x] int . \_\_index__
- [ ] int . \_\_getnewargs__
- [ ] int . \_\_init_subclass__
- [x] int . \_\_int__
- [x] int . \_\_invert__
- [x] int . \_\_radd__
- [x] int . \_\_rand__
- [x] int . \_\_rdivmod__
- [x] int . \_\_reduce__
- [x] int . \_\_reduce_ex__
- [x] int . \_\_rfloordiv__
- [x] int . \_\_rlshift__
- [x] int . \_\_rmod__
- [x] int . \_\_rmul__
- [x] int . \_\_ror__
- [x] int . \_\_round__
- [x] int . \_\_rpow__
- [x] int . \_\_rrshift__
- [x] int . \_\_rsub__
- [x] int . \_\_rtruediv__
- [x] int . \_\_rxor__
- [x] int . \_\_setattr__
- [ ] int . \_\_sizeof__
- [x] int . \_\_subclasshook__
- [x] int . \_\_trunc__
- [x] int . denominator
- [x] int . from_bytes
- [x] int . imag
- [x] int . numerator
- [x] int . real
- [x] int . to_bytes

Tom1380:
Can someone point me towards where I can find the specifications for these methods?

lausek:
Sorry, `__int__` was merged in #326 aswell.

Mec-iS:
@Tom1380 The reference is CPython implementation -> https://github.com/python/cpython

dralley:
The current list is just 

```
int.__getnewargs__
int.__init_subclass__
int.__sizeof__
```

ashermancinelli:
I'd love to try at `int.__sizeof__`. Would anyone be willing to give some direction?

coolreader18:
[...]

PR #1541, opened 2019-10-15
`int.__sizeof__` from #1426

ashermancinelli:
Impl from @RobertBerglund 's comment, working on #304

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -576,7 +576,7 @@ impl PyInt {
 
     #[pymethod(name = "__sizeof__")]
     fn sizeof(&self, _vm: &VirtualMachine) -> usize {
-        size_of::<Self>() + ((self.value.bits() + 7) & !7) / 8)
+        size_of::<Self>() + ((self.value.bits() + 7) & !7) / 8
     }
 
     #[pymethod]
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -576,7 +576,7 @@ impl PyInt {
 
     #[pymethod(name = "__sizeof__")]
     fn sizeof(&self, _vm: &VirtualMachine) -> usize {
-        (size_of::<Self>() + ((self.value.bits() + 7) & !7)) / 8
+        size_of::<Self>() + ((self.value.bits() + 7) & !7) / 8)
     }
 
     #[pymethod]
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -576,7 +576,7 @@ impl PyInt {
 
     #[pymethod(name = "__sizeof__")]
     fn sizeof(&self, _vm: &VirtualMachine) -> usize {
-        (size_of::<Self>() + (self.value.bits() + 7) & !7) / 8
+        (size_of::<Self>() + ((self.value.bits() + 7) & !7)) / 8
     }
 
     #[pymethod]
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -576,7 +576,7 @@ impl PyInt {
 
     #[pymethod(name = "__sizeof__")]
     fn sizeof(&self, _vm: &VirtualMachine) -> usize {
-        size_of::<Self>()
+        (size_of::<Self>() + (self.value.bits() + 7) & !7) / 8
     }
 
     #[pymethod]


vm/src/obj/objint.rs
```diff
@@ -186,6 +186,11 @@ impl PyInt {
         PyInt::new(options.get_int_value(vm)?).into_ref_with_type(vm, cls)
     }
 
+    #[pymethod(name = "__sizeof__")]
+    fn sizeof(&self, _vm: &VirtualMachine) -> usize {
+        (self.value.bits() as f64 / 8.).ceil() as usize
```
ashermancinelli: Results in:
```python
>>>>> for i in range(0, 100, 10):
..... 	print((i).__sizeof__())
..... 
0
1
1
1
1
1
1
1
1
1
```

vm/src/obj/objint.rs
```diff
@@ -188,7 +189,7 @@ impl PyInt {
 
     #[pymethod(name = "__sizeof__")]
     fn sizeof(&self, _vm: &VirtualMachine) -> usize {
-        (self.value.bits() as f64 / 8.).ceil() as usize
+        mem::size_of::<Self>() as usize
```
ashermancinelli: Results in:
```python
>>>>> for i in range(0, 100, 10):
..... 	print((i).__sizeof__())
..... 
32
32
32
32
32
32
32
32
32
32
```

ashermancinelli:
@coolreader18 
```rust
mem::size_of::<Self>() as usize
```
This'll just always return 32 I believe, I'm not sure how useful this is. You're right that it goes deeper than I thought :sweat_smile:

coolreader18:
Yeah, it's not as simple as `mem::size_of::<T>()` because that just returns the size of `T` on the stack, which is not always necessarily a meaningful value. [This thread](https://users.rust-lang.org/t/solved-rust-round-usize-to-nearest-multiple-of-8/25549) from the rust forums might be useful; I'm not sure if there was some sort of bug with the f64 conversion.

coolreader18:
Although I don't think there was, it just didn't go above 1 byte because 100 can be contained with just 8 binary digits; you'd have to go above 256 to see a change.

vm/src/obj/objint.rs
```diff
@@ -189,7 +188,7 @@ impl PyInt {
 
     #[pymethod(name = "__sizeof__")]
     fn sizeof(&self, _vm: &VirtualMachine) -> usize {
-        mem::size_of::<Self>() as usize
+        self.value.bits() + 7 & !7
```
ashermancinelli:
```python
>>>>> (2**1).__sizeof__()
8
>>>>> (2**8).__sizeof__()
16
>>>>> (2**16).__sizeof__()
24
>>>>> (2**32).__sizeof__()
40
```

ashermancinelli: @coolreader18 I think you're right about just not going high enough, I'll test that out

youknowone: sizeof int is sum of size of int object itself + sum of its heap allocated memory.
So, roughly, `mem::sizeof<T> + 4 * self.value.data.len()` is expected.
If current code describes the latter, adding `mem::sizeof` will make the value correct

vm/src/obj/objint.rs
```diff
@@ -188,7 +189,7 @@ impl PyInt {
 
     #[pymethod(name = "__sizeof__")]
     fn sizeof(&self, _vm: &VirtualMachine) -> usize {
-        self.value.bits() + 7 & !7
+        mem::size_of::<Self>() + (self.value.bits() + 7) & !7
```
ashermancinelli:
```python
40
>>>>> (2**8).__sizeof__()
48
>>>>> (2**16).__sizeof__()
56
>>>>> (2**32).__sizeof__()
72
```

coolreader18: `__sizeof__` should actually return bytes, so you'd probably want to divide the bits you get by 8

vm/src/obj/objint.rs
```diff
@@ -576,7 +576,7 @@ impl PyInt {
 
     #[pymethod(name = "__sizeof__")]
     fn sizeof(&self, _vm: &VirtualMachine) -> usize {
-        size_of::<Self>()
+        (size_of::<Self>() + (self.value.bits() + 7) & !7) / 8
```
ashermancinelli: Divided by 8 to return bytes instead of bits.

coolreader18: `mem::size_of` actually does return bytes, it's the `BigInt::bits` method that returns bits.

vm/src/obj/objint.rs
```diff
@@ -576,7 +576,7 @@ impl PyInt {
 
     #[pymethod(name = "__sizeof__")]
     fn sizeof(&self, _vm: &VirtualMachine) -> usize {
-        size_of::<Self>()
+        (size_of::<Self>() + ((self.value.bits() + 7) & !7)) / 8
```
coolreader18:
```suggestion
        size_of::<Self>() + ((self.value.bits() + 7) & !7) / 8)
```

PR #1542, opened 2019-10-17
Make async def/await actually functional

coolreader18:
Here's what I've done so far with this and what's left to do:

- [x] `async def`, coroutines
- [x] `await`
- [x] `async with`
- [x] `async for`
- [ ] `async` generators

### DIFF
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -386,27 +386,72 @@ impl<O: OutputStream> Compiler<O> {
                 body,
             } => {
                 if *is_async {
-                    unimplemented!("async with");
-                } else {
-                    let end_label = self.new_label();
-                    for item in items {
-                        self.compile_expression(&item.context_expr)?;
-                        self.emit(Instruction::SetupWith { end: end_label });
-                        match &item.optional_vars {
-                            Some(var) => {
-                                self.compile_store(var)?;
+                    let end_labels = items
+                        .iter()
+                        .map(|item| {
+                            let end_label = self.new_label();
+                            self.compile_expression(&item.context_expr)?;
+                            self.emit(Instruction::BeforeAsyncWith);
+                            self.emit(Instruction::GetAwaitable);
+                            self.emit(Instruction::LoadConst {
+                                value: bytecode::Constant::None,
+                            });
+                            self.emit(Instruction::YieldFrom);
+                            self.emit(Instruction::SetupAsyncWith { end: end_label });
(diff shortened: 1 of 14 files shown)


coolreader18:
I think this is good for review now; async generators weren't added to CPython until a while after coroutines were, anyway.

compiler/src/compile.rs
```diff
@@ -797,11 +811,15 @@ impl<O: OutputStream> Compiler<O> {
...
         returns: &Option<ast::Expression>, // TODO: use type hint somehow..
+        is_async: bool,
     ) -> Result<(), CompileError> {
         // Create bytecode for this function:
         // remember to restore self.in_loop to the original after the function is compiled
         let was_in_loop = self.in_loop;
         let was_in_function_def = self.in_function_def;
+
+        let was_in_async_func = self.in_async_func;
+        self.in_async_func = is_async;
```
palaviv: Can we be `in_async_func` without `in_function_def` being `true`? Maybe these two flags should be combined to a enum named something like `func_mode`?

coolreader18: That would be good, some sort of Context enum

compiler/src/compile.rs
```diff
@@ -383,13 +385,26 @@ impl<O: OutputStream> Compiler<O> {
...
-                if *is_async {
-                    unimplemented!("async with");
-                } else {
-                    let end_label = self.new_label();
-                    for item in items {
+                let is_async = *is_async;
+
+                let end_labels = items
+                    .iter()
+                    .map(|item| {
```
windelbouwman: I'm still not a great fan of inline lambda functions like this. I would prefer a plain for loop and a mutable `end_labels` variable in this case.

Riey:
Will it can be work with Rust's Future?

windelbouwman:
@coolreader18 should this be merged already?

coolreader18:
Oh whoops, I forgot about this one. I'm going to add the context enum for compilation, and then it'll be good to merge.

coolreader18:
@Riey you could in theory write a library/module for RustPython to have an event loop based off of something like tokio, but I think we're probably just going to copy CPython's event loop for asyncio.

coolreader18:
Does the most recent commit look good? I put the context all into one struct so that it's easy to copy out/restore when necessary.

coolreader18:
Merging, I think these changes are ok. @windelbouwman if you think that the CompileContext isn't good let me know.

Issue #1147, opened 2019-07-14
Implement rest of range methods


dmitrijoseph:
## Feature
#293 
I will implement the rest of the range methods.
## Remaining Methods

range.\_\_init\_subclass\_\_
range.\_\_reduce\_\_
range.\_\_reduce_ex\_\_
range.\_\_sizeof\_\_
range.\_\_subclasshook\_\_

dmitrijoseph:
range.\_\_setattr\_\_ 
Bugged 
Current Behavior on Python 3.6
range(0,10).__setattr__("start",0) => AttributeError: readonly attribute
RustPython behavior
range(0,10).__setattr__("start",0) => AttributeError: "'range' object has no attribute 'start'"

isidentical:
for `range.__init_subclass__`, check #1346

tekknolagi:
Looks like `range` doesn't implement `__reduce_ex__` (which is correct) but the `object` implementation of it in RustPython is a little buggy:

```
>>>>> range(0, 1, 2).__reduce_ex__()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Expected at least 2 arguments (1 given)
```

youknowone:
It is done except for `__init__subclass__` (which will be followed up from #1346)

PR #1544, opened 2019-10-18
Implemented range.__reduce__

xarus01:
#1147


tests/snippets/builtin_range.py
```diff
@@ -108,6 +108,12 @@
 assert list(reversed(range(5, 0, -1))) == [1, 2, 3, 4, 5]
 assert list(reversed(range(1,10,5))) == [6, 1]
 
+# __reduce__
+assert range(10).__reduce__()[0] == range
+assert range(10).__reduce__()[1] == (0, 10, 1)
+assert range(10, 1, -2).__reduce__()[0] == range
+assert range(10, 1, -2).__reduce__()[1] == (10, 1, -2)
```
coolreader18: You could try testing using `copy.copy()`; it should use `__reduce__` and return a `range` object that is `==` but not `is` the original

Issue #1531, opened 2019-10-14
The deletion of the dict does not affect the size of the vector


HyeockJinKim:
## Feature
https://github.com/RustPython/RustPython/blob/d42d422566f64f48311cc7e7efb4d0ffcefb0297/vm/src/dictdatatype.rs#L21-L25

As you can see in the code above, Dict stores the values ​​in a Vector.
However, when deleting a value, dict simply decrease size, but the size of the vector continues to grow.

```python
>>>>> for i in range(100000):
.....     a[0] = 1
.....     del a[0]
..... 
```
If running the code above, the vector will continue to grow in size and take longer than python3.6.

So `dict` seems to need the `resize` function.

But I'm not sure when to call `resize` function.
Should it be the same as calling `dictresize` in` cpython`?

In cpython `dictresize`:
https://github.com/python/cpython/blob/8a6cbf8adb68aa2c43a48769af92e50d5ca1539d/Objects/dictobject.c#L1177

ashermancinelli:
https://github.com/RustPython/RustPython/blob/a1e3fe4e63cc21fb60093c0d43dd2cd8f56b09b6/vm/src/dictdatatype.rs#L69-L72
Could we free some memory here?

PR #1545, opened 2019-10-18
Reuse slots in `dictionary.entries` object

makarchuk:
Addresses #1531
Previously entries vector used to grow on insert/delete operations since slots were not being reused.
I've added a list of such indices and now they are being used on insert, preventing an `entries` list from growing

I considered two options for empty_slots:
- LinkedList
- Vector

With Vector could consume memory by itself and be not efficient in case when dictionary shrinks in size drammatically, so I chose LinkedList for now.

I tested it on my machine with script like this: 
```python
a = {0: 1}
for i in range(10000):
    del a[0]
    a[0] = 1
```
here are the results:
```
$ time ./target/release/rustpython-master test.py 

real    0m3,284s
user    0m3,280s
sys     0m0,005s

 $ time ./target/release/rustpython-list test.py 
real    0m1,608s
user    0m1,604s
sys     0m0,004s
```

Still, Dictionary resize operation would be reasonable, but now we can have an actual metrics on when to call it (some ratio between `empty_slots.len()` and `entries.len()` would be reasonable, I think), and anyway, reusing an indices in array seems like a reasonable thing to do.

### DIFF
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -7,7 +7,7 @@ use num_bigint::ToBigInt;
 /// Inspired by: https://morepypy.blogspot.com/2015/01/faster-more-memory-efficient-and-more.html
 /// And: https://www.youtube.com/watch?v=p33CVV29OG8
 /// And: http://code.activestate.com/recipes/578375/
-use std::collections::{hash_map::DefaultHasher, HashMap};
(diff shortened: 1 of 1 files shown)


makarchuk:
Closing in favour of resizing. This PR breaks contract of orderedness and doesn't address some problems

PR #1546, opened 2019-10-18
Implement dict resizing

makarchuk:
Dict resizing is working by removing all None entries, and filling up indices map from scratch
Frequency of resizing should be adjusted, I guess. Maybe it'll be useful to trigger it more often, it's hard to say without somewhat realistic benchmark

Addresses: #1531 

Benchmark: 
```python
a = {0: 1}
for i in range(10000):
    del a[0]
    a[0] = 1


```

Results: 
```
$ time ./target/release/rustpython-master test.py

real    0m3,316s
user    0m3,312s
sys     0m0,004s

$ time ./target/release/rustpython test.py

real    0m0,062s
user    0m0,055s
sys     0m0,008s

$ time python3 test.py

real    0m0,023s
user    0m0,015s
sys     0m0,008s

```

### DIFF
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -37,6 +37,7 @@ impl<T> Default for Dict<T> {
 
 #[derive(Clone)]
 struct DictEntry<T> {
+    hash_index: HashIndex,
     hash: HashValue,
     key: PyObjectRef,
     value: T,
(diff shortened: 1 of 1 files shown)


vm/src/dictdatatype.rs
```diff
@@ -79,6 +100,9 @@ impl<T: Clone> Dict<T> {
         key: K,
         value: T,
     ) -> PyResult<()> {
+        if self.indices.len() > 2 * self.size {
```
HyeockJinKim: It would be nice to check how it was implemented in cpython and implement a resize policy.
https://github.com/python/cpython/blob/master/Objects/dictobject.c

vm/src/dictdatatype.rs
```diff
@@ -37,6 +37,7 @@ impl<T> Default for Dict<T> {
 
 #[derive(Clone)]
 struct DictEntry<T> {
+    hash_index: HashIndex,
```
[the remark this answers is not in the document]
makarchuk: Oh, yes, you're correct. I was planning on using it in my first version and it stayed there by mistake.

vm/src/dictdatatype.rs
makarchuk: It seems like they have a very different reasons for resizing. E.g. they need to manually resize what we call `entries` (`ma_values` in cpython), C doesn't handle resizing for them, so they do it by calling `dictresize`.  
Overall they do it quite often, so I assume they don't have a problem we're trying to solve (`indices` map is growing big)

windelbouwman:
How is it possible that the benchmark is so much faster? Does this change also improve speed of other python code?

makarchuk:
> How is it possible that the benchmark is so much faster? Does this change also improve speed of other python code?

That's because reusing a key in a dictionary used to cause growth of `Dict.indices`. and since all of the keys have the same hash (as they are equal) each map insert used to be forced to go through hundreds of layers of hash collisions to actually find an empty slot.

Same goes for deletion.

Issue #1523, opened 2019-10-13
Objects inherit docstring from object if not overwritten


janrg:
in CPython, if an object doesn't have a docstring assigned, its `__doc__` is set to `None`, in RustPython, it inherits `the most base type` from `object`

seeeturtle:
I'll try it.

seeeturtle:
the same problem in function should be fixed too.

seeeturtle:
It seems the problem of function and the problem of class is a bit different.
Function doesn’t have doc attribute at all but class inherits the doc attribute of object

PR #1550, opened 2019-10-20
objtype: Set __doc__ as None if not included 

seeeturtle:
This set `__doc__` as None if it is not included in attribute
while making new instance of type.
In cpython if there is no docstring in class definition,
`__doc__` attribute will be None. So RustPython should follow it.

Partially solves #1523 

This only fixes class definition.

### DIFF
--- a/vm/src/obj/objtype.rs
+++ b/vm/src/obj/objtype.rs
@@ -321,7 +321,12 @@ fn type_new_slot(metatype: PyClassRef, args: PyFuncArgs, vm: &VirtualMachine) ->
         bases
     };
 
-    let attributes = dict.to_attributes();
+    let mut attributes = dict.to_attributes();
+
+    // insert __doc__ as None if it is not included in attributes
+    if !attributes.contains_key("__doc__") {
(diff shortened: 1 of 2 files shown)


vm/src/obj/objtype.rs
```diff
@@ -321,7 +321,12 @@ fn type_new_slot(metatype: PyClassRef, args: PyFuncArgs, vm: &VirtualMachine) ->
         bases
     };
 
-    let attributes = dict.to_attributes();
+    let mut attributes = dict.to_attributes();
+
+    // insert __doc__ as None if it is not included in attributes
+    if !attributes.contains_key("__doc__") {
```
HyeockJinKim:
```python
>>>>> def f():
.....  pass
..... 
>>>>> f.__doc__
'The most base type'
```
The current code does not check the code above.
The expected result is shown below.
```python
>>> def f():
...  pass
... 
>>> f.__doc__
>>> 
```

seeeturtle: yes, this doesn't fixes the same problem in function.
I'm planning to handle that in another PR.

>This only fixes class definition.

PR #1552, opened 2019-10-20
Improve  float.fromhex module of objfloat

Writtic:
- Python `float.fromhex` can handle string of float and hex without `.`
  - For example `0.0`, `-0.0`, `0x0p0`, `-0x0p0`
- Add test case mantioned

### DIFF
--- a/vm/src/obj/objfloat.rs
+++ b/vm/src/obj/objfloat.rs
@@ -576,7 +576,8 @@ impl PyFloat {
             "inf" => Ok(std::f64::INFINITY),
             "-inf" => Ok(std::f64::NEG_INFINITY),
             value => {
-                if value.contains("0x") {
+                let err_msg = "invalid hexadecimal floating-point string";
+                let res = if value.contains("0x") {
                     let mut hex = "".to_string();
                     for ch in value.chars() {
                         if ch == 'p' {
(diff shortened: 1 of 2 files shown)


vm/src/obj/objfloat.rs
```diff
@@ -575,7 +575,30 @@ impl PyFloat {
             "nan" => Ok(std::f64::NAN),
             "inf" => Ok(std::f64::INFINITY),
             "-inf" => Ok(std::f64::NEG_INFINITY),
-            _ => Err(vm.new_value_error("invalid hexadecimal floating-point string".to_string())),
+            value => {
+                let err_msg = "invalid hexadecimal floating-point string";
+                let res = if value.contains("0x") {
```
windelbouwman: Instead of assigning `res` first, and then using it as an expression, you could opt to not create a `res` variable and use the `if` as the value.

vm/src/obj/objfloat.rs
```diff
@@ -575,7 +575,30 @@ impl PyFloat {
...
+                        }
+                    }
+
+                    match hexf_parse::parse_hexf64(&hex, false) {
+                        Ok(f) => Ok(f),
+                        _e => Err(vm.new_value_error(err_msg.to_string())),
+                    }
+                } else {
+                    match value.parse::<f64>() {
+                        Ok(f) => Ok(f),
```
windelbouwman: You could maybe use the `map_err` function here.

Writtic: Thanks for your advice!
I changed the whole code on match statement to simplyfied one.

- Use `or_else` instead of value condition and `match` statement
  - For floating-point input
- Use `map_err` instead of `match ` statement
  - For hexadecimal floating-point input without dot

vm/src/obj/objfloat.rs
```diff
@@ -571,12 +571,28 @@ impl PyFloat {
...
+                    let mut hex = String::new();
+                    if value.contains("0x") {
+                        for ch in value.chars() {
+                            if ch == 'p' {
+                                hex.push_str(".p");
+                            } else {
+                                hex.push(ch);
+                            }
+                        }
+                    }
```
Writtic: Do I have to add the code below..?
```rust
} else {
    hex = value.to_string();
}
```

Writtic: Nevermind. I added it.

PR #1553, opened 2019-10-21
Add the select and selectors modules

coolreader18:


vm/src/stdlib/select.rs
```diff
@@ -0,0 +1,75 @@
...
+        .map(|to| TimeVal::nanoseconds((to * 1e9) as i64));
+
+    select::select(
+        None,
+        Some(&mut r),
+        Some(&mut w),
+        Some(&mut x),
+        timeout.as_mut(),
+    )
+    .map_err(|err| super::os::convert_nix_error(vm, err))?;
```
palaviv: We should check here if the select has failed due to an interrupt and if so we should check signals and continue. I would suggest just adding a `TODO` for now.

coolreader18:
I've addressed your suggestions @palaviv; I didn't add tests because I wasn't quite sure how I would do them, maybe subprocesses on separate threads? What do you think?

palaviv:
Looks good. I would suggest using sockets for testing. I think that will be the easiest way.

coolreader18:
I'm working on improving the `socket` module, I think I'll add tests based on that new version.

PR #1557, opened 2019-10-23
Implemented bytes.__sizeof__

xarus01:
#736


coolreader18:
Thanks for contributing!

Issue #1560, opened 2019-10-23
bug in replacement bytes with more than one byte


yanganto:
CPython can replace with bytes 
```
Python 3.7.4 (default, Oct  4 2019, 06:57:26) 
[GCC 9.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> b"rust  python".replace(b'  ', b'_')
b'rust_python'
>>> b"rust  python".replace(b' ', b'_')
b'rust__python'
```

RustPython
```
Welcome to the magnificent Rust Python 0.1.1 interpreter 😱 🖖
>>>>> b"rust  python".replace(b'  ', b'_')
b'rust_pytho'
>>>>> b"rust  python".replace(b' ', b'_')
b'rust__python'
```

PR #1561, opened 2019-10-23
bytes.replace: replacing bytes w/ diff len

yanganto:
fix the bug on replacing bytes with different length
issue #1560


coolreader18:
Thanks for contributing!

PR #1563, opened 2019-10-24
bytes.mod implement with CFormat

yanganto:
implement byte formatting with bytes or tuple of bytes #736

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -1269,6 +1269,7 @@ fn call_object_format(vm: &VirtualMachine, argument: PyObjectRef, format_spec: &
         Some(FormatPreconversor::Str) => vm.call_method(&argument, "__str__", vec![])?,
         Some(FormatPreconversor::Repr) => vm.call_method(&argument, "__repr__", vec![])?,
         Some(FormatPreconversor::Ascii) => vm.call_method(&argument, "__repr__", vec![])?,
+        Some(FormatPreconversor::Bytes) => vm.call_method(&argument, "decode", vec![])?,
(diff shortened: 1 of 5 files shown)


yanganto:
This PR is related to #1569

yanganto:
working on `%i` and `%f`

coolreader18:
I don't want to discourage you with all the work you've done so far, but printf-style bytes formatting [is pretty feature-equivalent](https://docs.python.org/3/library/stdtypes.html#printf-style-bytes-formatting) to the normal `str` version, e.g. it has named replacements, number formatting options, etc; so you might want to look in to modifying the existing code in `cformat.rs` to accept bytes.

yanganto:
@coolreader18 Thanks for point out.  A suitable implementation should use the existing functions properly.  I will figure out how `cformat.rs` works in `str`.

vm/src/obj/objstr.rs
```diff
@@ -1387,11 +1389,11 @@ fn try_update_quantity_from_tuple(
     }
 }
 
-fn do_cformat(
+pub fn do_cformat_string(
     vm: &VirtualMachine,
     mut format_string: CFormatString,
     values_obj: PyObjectRef,
-) -> PyResult {
+) -> Result<String, PyObjectRef> {
```
youknowone: This is same to `PyResult<String>`

yanganto: Thanks for the review.

yanganto:
Thanks for your time.  Thanks for the review.

coolreader18:
No problem, sorry it took this long.

PR #1565, opened 2019-10-24
test.not_impl_gen ignore py module w/ import error

yanganto:
Ignore python module which cannot be imported and print error message thereof

I am not sure the reason, but I have a problem with import `pyopenvdb`.

This does not really matter to block `whats_left.sh`, so I propose to ignore error and print error message as following.
```
!!! pyopenvdb skipped because ImportError: /usr/lib64/libjemalloc.so.2: cannot allocate memory in static TLS block
```


coolreader18:
Sorry for taking a while to get to this; this LGTM

PR #1570, opened 2019-10-26
Update chrono to 0.4.9

Riey:
chrono 0.4.7 doesn't support some features in wasm (e.g. `Utc::now()` `Local::now()` ...)

This PR can help such script can run on wasm

```python
import time

print(time.gmtime())
print(time.asctime())
```


coolreader18:
Thanks for contributing!

PR #1571, opened 2019-10-26
Implement "random" module (WIP)

malkoG:
Functions for integers, functions for sequences works. But, Real-valued distributions is still working on progress.

### DIFF
--- a/tests/snippets/stdlib_random.py
+++ b/tests/snippets/stdlib_random.py
@@ -1,0 +1,30 @@
+import random 
+
+random.seed(1234)
+
+# random.randint
+assert random.randint(1, 11) == 8
+
+# random.shuffle
+left = list(range(10))
+right = [2, 7, 3, 5, 8, 4, 6, 9, 0, 1]
+random.shuffle(left)
+
+assert all([l == r for l, r in zip(left, right) ])
+
+# random.choice
+assert random.choice(left) == 5
+
+# random.choices 
+expected = ['red', 'red', 'black', 'red', 'black', 'black']
+result = random.choices(['red', 'black', 'green'], [18, 18, 2], k=6)
+assert all([l == r for l, r in zip(expected, result) ])
+
+# random.sample
+sampled = [2, 1, 0]
+assert all([l == r for l, r in zip(random.sample(l, 3), sampled) ])
+
+# TODO : random.random(), random.uniform(), random.triangular(),
+#        random.betavariate, random.expovariate, random.gammavariate,
+#        random.gauss, random.lognormvariate, random.normalvariate,
+#        random.vonmisesvariate, random.paretovariate, random.weibullvariate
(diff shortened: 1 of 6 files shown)


vm/src/stdlib/random.rs
```diff
@@ -1,15 +1,82 @@
...
+impl PyRandom {
+    #[pyslot(new)]
+    fn new(cls: PyClassRef, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+        PyRandom {
+            rng: RefCell::new(SmallRng::from_entropy())
+        }.into_ref_with_type(vm, cls)
+    }
+
+    #[pymethod] 
+    fn seed(&self, n: Option<usize>, vm: &VirtualMachine) -> PyResult {
```
coolreader18: You can make `n` an `Option<u64>` to avoid the cast

tests/snippets/stdlib_random.py
```diff
@@ -0,0 +1,30 @@
...
+
+# random.randint
+assert random.randint(1, 11) == 8
+
+# random.shuffle
+left = list(range(10))
+right = [2, 7, 3, 5, 8, 4, 6, 9, 0, 1]
+random.shuffle(left)
+
+assert all([l == r for l, r in zip(left, right) ])
```
sanxiyn: Isn't this same as `assert left == right`? Same for other tests too.

vm/src/stdlib/random.rs
```diff
@@ -1,15 +1,82 @@
...
+use crate::function::OptionalArg;
+use crate::obj::objtype::PyClassRef;
+use crate::pyobject::{PyClassImpl, PyObjectRef, PyRef, PyValue, PyResult};
+
 use crate::vm::VirtualMachine;
 
+#[pyclass(name = "Random")]
+#[derive(Debug)]
+struct PyRandom {
+    rng: RefCell<SmallRng>
```
vks: I would strongly recommend to use `StdRng` as a default instead, it is a much safer choice, because it is not predictable.

vm/src/stdlib/random.rs
```diff
@@ -1,15 +1,82 @@
...
+        }.into_ref_with_type(vm, cls)
+    }
+
+    #[pymethod] 
+    fn seed(&self, n: Option<usize>, vm: &VirtualMachine) -> PyResult {
+        let rng = match n {
+            None => SmallRng::from_entropy(),
+            Some(n) => {
+                let seed = n as u64;
+                SmallRng::seed_from_u64(seed)
```
vks: It would be nice to support larger seeds (via bigints), but this requires more work and could be done later (likely in a backwards incompatible way).

vm/src/stdlib/random.rs
coolreader18: It would probably be possible to have an enum with both `StdRng` and `SmallRng` as variants, and only switch to `SmallRng` when `seed` is called.

vks: `SmallRng` is not the Mersenne Twister, so it does not make sense to switch to it for compatibility with CPython. Other than that, I agree!

coolreader18:
@malkoG I'd be willing to take over this PR, could I have your permission to use the code you've written as a base?

malkoG:
Sure

PR #1572, opened 2019-10-26
Update _socket to use the socket2 crate; similar to the C sockets api

coolreader18:
Also adds `socket.py` from CPython 3.6.0


palaviv:
@coolreader18 can you split this PR into smaller ones? It will be easier to review that way

coolreader18:
@palaviv #1574

vm/src/stdlib/socket.rs
```diff
@@ -1,493 +1,317 @@
...
-    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        match self {
-            Connection::TcpStream(con) => con.write(buf),
-            Connection::UdpSocket(con) => con.send(buf),
-            _ => Err(io::Error::new(io::ErrorKind::Other, "oh no!")),
-        }
-    }
-    fn flush(&mut self) -> io::Result<()> {
-        Ok(())
+        Ok(vm.ctx.new_tuple(vec![fd, addr_tuple]))
```
palaviv: You can return here `(fd, addr_tuple)`

vm/src/stdlib/socket.rs
```diff
@@ -1,493 +1,317 @@
...
-                    }
-                } else {
-                    Err(vm.new_type_error("".to_string()))
-                }
+    #[pymethod]
+    fn recvfrom(&self, bufsize: usize, vm: &VirtualMachine) -> PyResult {
+        let mut buffer = vec![0u8; bufsize];
+        let addr = match self.sock().recv_from(&mut buffer) {
+            Ok((_, addr)) => addr,
+            Err(ref e) if e.kind() == io::ErrorKind::TimedOut => {
```
palaviv: I would suggest adding a function name `convert_socket_error` that will check if this is a timeout and then call `convert_io_error`. That will remove a lot of duplication.

palaviv:
@coolreader18 can you look at #1580? The windows test of select failed there.

coolreader18:
I'm not sure, that's a really weird failure. I tried rerunning it but the checkout failed because the branch is deleted.

PR #1573, opened 2019-10-27
Traceback object

palaviv:
This PR change exception traceback to be with `traceback` object very similar to CPython.


vm/src/frame.rs
```diff
@@ -80,6 +81,7 @@ enum UnwindReason {
 }
 
 #[pyclass]
+#[derive(Clone)]
```
windelbouwman: This might be tricky, since the traceback will contain a copied frame, and not the actual frame. The solution might be to modify the run method to take a clonable `PyFrameRef`.

palaviv: Sure. I will merge this and we can do that in the future

PR #1576, opened 2019-10-30
Reorganize vm/Lib and frozen.rs

coolreader18:
I think I like the `vm/Lib` directory structure, but I'd like to get some feedback on if it seems like a good idea. Also, I changed `_warnings.warn` to not print the `{:?}` repr of a PyObjectRef, because that took up about 300 lines in my console while I was trying to debug something.


palaviv:
@coolreader18 can you explain the idea behind this change? Is `pylinks` something you invented or is this something that exist in CPython?

coolreader18:
It's something I came up with because symlinks don't work on windows, and I don't want to just copy the files over (if we need to make a change or update to a new version from CPython, it'd be easy to only change one of the versions). In retrospect, though, it might be better to parse a file `module_name.py` itself rather read a file path from a separate file.

palaviv:
I have to say I don't like the idea of the `pylink` files. From this [thread](https://stackoverflow.com/questions/5917249/git-symlinks-in-windows) in stackoverflow it seems that you can automatically convert the symlinks to windows symlink.

coolreader18:
I'm trying that now, with an updated version of the windows vm on azure pipelines, but another concern is that it might make it harder for newcomers to get started on windows: they might need a script to actually get it to build, or we could tell them to run git clone with admin privileges if they're not on a high enough version of windows, which might seem sketchy. If there is some way to configure git to just do it nicely, then I'd definitely prefer that, but otherwise I like a version where you can just clone it and it will work for all users.

palaviv:
I don't like the idea of adding our own concept of symlinks. I think that if we can print a good error with indicative instructions when this problem occurs that would be better. I do not see many issues or question in Gitter that are related to this problem.

coolreader18:
@palaviv I translated the script from the stackoverflow answer to Powershell, so pylink files aren't necessary. Does it look good to you?

Issue #1575, opened 2019-10-29
str.isupper and str.islower are not implemented properly


coolreader18:
## Feature

`str.isupper` should return `True` if all the characters that *can* be multiple cases in a string *are* uppercase. However, the current implementation just uses Rust's `char::is_uppercase`, which does not distinguish between 'cased' and 'non-cased' characters. The same goes for `str.islower`.

## Python Documentation

https://docs.python.org/3/library/stdtypes.html#str.isupper

## Implementation

This can be implemented by modifying the `isupper` and `islower` functions in `vm/src/obj/objstr.rs`.

Regexp96:
can I contribute this issue?
it's my first issue and I think I can do this

Regexp96:
that commit is my mistake sorry

coolreader18:
Yep, go ahead!

PR #1577, opened 2019-10-30
Fix issue #1575 distinguish between 'cased' and 'non-cased' characters

Regexp96:
#1575 distinguish between 'cased' and 'non-cased' characters

use unic::ucd::is_cased function to ignore non-cased character

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -753,6 +753,7 @@ impl PyString {
         }
         cased
     }
+    
     #[pymethod]
     fn isascii(&self, _vm: &VirtualMachine) -> bool {
         !self.value.is_empty() && self.value.chars().all(|c| c.is_ascii())
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -10,10 +10,10 @@ use std::str::FromStr;
 use std::string::ToString;
 
 use num_traits::ToPrimitive;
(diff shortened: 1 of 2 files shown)


coolreader18:
This looks good! Could you add some tests in `tests/snippets/strings.py` making sure that the new behavior is correct?

Regexp96:
ok I add some tests
thanks for reviewing

vm/src/obj/objstr.rs
```diff
@@ -725,25 +726,32 @@ impl PyString {
...
     #[pymethod]
     fn isupper(&self, _vm: &VirtualMachine) -> bool {
-        // TODO: assert not " ".isupper(), assert not "_".isupper()
-        !self.value.is_empty()
-            && self
-                .value
-                .chars()
-                .filter(|x| !x.is_ascii_whitespace() && *x != '_')
-                .all(char::is_uppercase)
+        let mut cased = false;
```
youknowone: how about this way?
```rust
   let mut cased = true;
   for c in self.value.chars() {
       if is_cased(c) && c.is_uppercase() {
            continue
        } else {
           cased = false;
           break;  // or `return true` as original
       }
   }
   cased
```

Regexp96: oh it's looks better!

Regexp96: this code return `false` when `"111a".islower()`
but in python it will return `true`

PR #1580, opened 2019-11-02
Set docstring of function as None if not declared

### LINKED ISSUE #1523: Objects inherit docstring from object if not overwritten

seeeturtle:
In cpython, if there is no docstring declared in function definition,
the `__doc__` attribute of the function is None.
So this implements the behavior.

Fix #1523

### DIFF
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -1015,22 +1015,23 @@ impl<O: OutputStream> Compiler<O> {
     }
 
     fn store_docstring(&mut self, doc_str: Option<String>) {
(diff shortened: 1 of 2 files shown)


palaviv:
Great work @seeeturtle. This make the change in #1550 redundant so I would suggest removing it in a different PR.

seeeturtle:
@palaviv what do you mean to remove in different PR?

coolreader18:
I think he's made the change in #1582

Issue #1584, opened 2019-11-06
math.floor and math.ceil don't return integer values


adolfogonzalez3:
## Feature

The math functions math.floor and math.ceil return floating point values which causes an error when trying to run:

```python
limit = 10.5
for i in range(math.ceil(limit)):
   pass
```

This works in python 3.7.

## Python Documentation

The documentation mentions the return type being Integral.
https://docs.python.org/3.5/library/math.html

PR #1585, opened 2019-11-06
Possible fix for math.ceil and math.floor

adolfogonzalez3:
Fixes the problem described in #1584 where rustpython's math.ceil and math.floor aren't consistent with cpython's.


vm/src/stdlib/math.rs
```diff
@@ -151,19 +151,45 @@ fn math_trunc(value: PyObjectRef, vm: &VirtualMachine) -> PyResult {
...
+/// # Remarks
+///
+/// Currently the code is applying ceil to the float using rust's builtin
+/// method and then converts to integral using python float's __int__ method.
+///
 fn math_ceil(value: PyObjectRef, vm: &VirtualMachine) -> PyResult {
     if objtype::isinstance(&value, &vm.ctx.float_type()) {
         let v = objfloat::get_value(&value);
-        Ok(vm.ctx.new_float(v.ceil()))
+        let v = vm.ctx.new_float(v.ceil());
```
silmeth: I’d suggest avoiding to go through Python float and calling magic `__int__` by doing the conversion in Rust and directly returning a Python int:
```rust
let v = v.ceil().to_i64().unwrap(); // this should never fail, since `ceil()` returns integer values
Ok(vm.ctx.new_int(v))
```
or something like that (and the same in `floor`).

adolfogonzalez3: I wanted to avoid that because there are situations where converting from f64 to i64 loses information. An example of something that works in cpython is:
```python
math.ceil(1e308)
```
cpython returns the correct value. But if f64 is unsafely converted into i64 directly in rust then it
returns an incorrect value of -9223372036854775808. Or it crashes the program with the safe conversion.

I think it would be better to convert the float directly into bigint but I couldn't find a way to pass that back as a result.

coolreader18: It looks like [`float.__int__`](https://github.com/RustPython/RustPython/blob/master/vm/src/obj/objfloat.rs#L530-L533) uses [`float.trunc`](https://github.com/RustPython/RustPython/blob/master/vm/src/obj/objfloat.rs#L467-L470) which uses [`objfloat::try_to_bigint`](https://github.com/RustPython/RustPython/blob/master/vm/src/obj/objfloat.rs#L92-L112). I'd recommend making that function public and directly calling it rather than the Python method.

silmeth: Then you can import `num::FromPrimitive` and do:

```rust
let v = BigInt::from_f64(v.ceil()).unwrap();
Ok(vm.ctx.new_int(v))
```

;-)

coolreader18: I think that would panic if `v` was `nan` or `inf`, though: [Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=68ab3dca3b31484e7d5c99ed1c881d1a)

silmeth: You’re right. I just wanted to suggest a standard way of converting to `BigInt` (and instead of unwrapping, proper error propagation could be used), but sure, if there’s already a standard way of `float` → `BigInt` conversion that generates proper Python error in RustPython, it’s better. I didn’t see your comment when writing mine.

adolfogonzalez3: Okay I updated the code based on coolreader18's suggestion.

adolfogonzalez3:
Should I remove the added test in the math_module?

tests/snippets/math_module.py
```diff
@@ -0,0 +1,120 @@
+import math
```
youknowone: The content of this file looks good. We are putting any stdlib test code into `stdlib_<name>.py` test file. So putting this test into `stdlib_math.py` will be better. https://github.com/RustPython/RustPython/blob/master/tests/snippets/stdlib_math.py

dralley: @adolfogonzales3 Are you continuing with the PR?

youknowone:
It seems something went wrong with merge.

when `RustPython/RustPython` is remote upstream, running this command will help to fix the branch
```sh
$ git fetch upstream
$ git rebase upstream/master
```

adolfogonzalez3:
Alright I ran the commands and fast forwarded the branch as @youknowone advised. Thank you for the advice.

coolreader18:
One last thing @adolfogonzalez3, could you put all the tests for this feature into `tests/snippets/stdlib_math.py` rather than `tests/snippets/math_module.py`? That's where we keep the tests for the `math` module.

adolfogonzalez3:
Okay I will move the test code into `tests/snippets/stdlib_math.py` and make another commit. It looks like `tests/snippets/math_module.py` was deleted from the master branch so should I remove the entire file or just remove the test?

coolreader18:
Removing the entire file would be great, thanks. I think it was renamed in #1530, which was probably after you started working on this.

coolreader18:
Thanks for contributing and seeing this through!

PR #1590, opened 2019-11-08
Impl input

### LINKED ISSUE #1586: Implement input()
## Feature

## Python Documentation

And the code from cpython

adolfogonzalez3:
Fixes #1586 

This a possible implementation of input. I've moved the rustyline repl into the vm. I would like feedback on improvements or whether input should be tackled differently. 

A few differences currently from cpython is that:

The interrupted signal is handled differently for cpython and rustpython.
Cpython treats the signal as a character and rustpython exits execution

When in interactive mode in rustpython and exit() is called in the repl, it seems the program immediately exits. This makes it difficult to save the history to a file. When Cpython has exit() called in its repl, the history is still saved.

For reference:
I use windows 10 and python 3.8


coolreader18:
I was thinking that you could move `shell.rs` into `vm/src`, and just create a new repl for `builtin_input` rather than store it in the VM.

adolfogonzalez3:
I had forgotten to rebase my repository. I plan on doing it over again on a new branch.

adolfogonzalez3:
Okay I will try as you suggested.

PR #1591, opened 2019-11-08
Change mem::uninitialized to mem::MaybeUninit

palaviv:
`mem::uninitialized` has been [deprecated](https://github.com/rust-lang/rust/blob/stable/RELEASES.md#compatibility-notes-3) in rust version 1.39. I have changed the usages to use `mem::MaybeUninit`.

### DIFF
--- a/vm/src/types.rs
+++ b/vm/src/types.rs
@@ -37,7 +37,8 @@ use crate::obj::objweakref;
 use crate::obj::objzip;
 use crate::pyobject::{PyAttributes, PyContext, PyObject, PyObjectRef};
 use std::cell::RefCell;
-use std::mem::MaybeUninit;
+use std::mem;
+use std::ptr;
 
 /// Holder of references to builtin types.
 #[derive(Debug)]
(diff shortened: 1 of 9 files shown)


vm/src/types.rs
```diff
@@ -232,7 +231,7 @@ fn init_type_hierarchy() -> (PyClassRef, PyClassRef) {
     // (and yes, this will never get dropped. TODO?)
     let (type_type, object_type) = unsafe {
         let object_type = PyObject {
-            typ: mem::uninitialized(), // !
```
coolreader18: The source of `mem::uninitialized` since `MaybeUninit` was added has just been `MaybeUninit::uninit().assume_init()`, so it might be easier to just do that than change the type of `typ` and special-case it throughout the codebase.

palaviv: That is better

silmeth: @palaviv, @coolreader18 Isn’t `assume_init()` technically undefined behaviour here, if called before actually writing to the location?

coolreader18: Yes, it is, but `mem::uninitialized` did that before as well. This is unsafe code, and we're taking a lot of care to not invoke UB by using `ptr::write` instead of `*a.b = c`, etc.

silmeth: @coolreader18 An uninitialized `PyRef` is created, which holds an uninitialized `Rc<_>`, which then holds an unitialized [`NonNull<RcBox<T>>`](https://doc.rust-lang.org/src/alloc/rc.rs.html#279) – which, by the virtue of being uninitialized, might actually be `null`. That is UB. And that is the reason why Rust team changed `mem::uninitialized()` into `mem:MaybeUninit` in the first place.

If it works now, it might not work in the future. To actually do it safely, you need to first do `prt::write` on `MaybeUninit`, and only then (when you know what bit pattern is at the given pointer), `assume_init()`. At least that’s my understanding.

PR #1594, opened 2019-11-12
Add asyncio stdlib module

coolreader18:
Most of the Lib files are taken from CPython 3.6. To see just the changes to the stdlib files (or just the changes I actually made) you can look [here](https://github.com/RustPython/RustPython/compare/87c08e6...coolreader18/asyncio); `asyncio.runners` and the changes to `asyncio.tasks` are taken from CPython 3.7.


palaviv:
This looks great @coolreader18. Can you split this PR into some smaller PRs?

Lib/selectors.py
```diff
@@ -320,7 +320,7 @@ def select(self, timeout=None):
         timeout = None if timeout is None else max(timeout, 0)
         ready = []
         try:
-            r, w, _ = self._select(self._readers, self._writers, [], timeout)
+            r, w, _ = type(self)._select(self._readers, self._writers, [], timeout)
```
palaviv: Why is this needed?

vm/src/stdlib/dis.rs
```diff
@@ -17,11 +18,25 @@ fn dis_disassemble(co: PyObjectRef, vm: &VirtualMachine) -> PyResult {
...
+}
+
 pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
     let ctx = &vm.ctx;
 
     py_module!(vm, "dis", {
         "dis" => ctx.new_rustfunc(dis_dis),
-        "disassemble" => ctx.new_rustfunc(dis_disassemble)
+        "disassemble" => ctx.new_rustfunc(dis_disassemble),
+        "COMPILER_FLAG_NAMES" => dis_compiler_flag_names(vm),
```
palaviv: Can you add test for this in dismod.py?

vm/src/obj/objobject.rs
```diff
@@ -210,20 +210,22 @@ fn object_class_setter(
...
 
-fn object_dict_setter(
-    _instance: PyObjectRef,
-    _value: PyObjectRef,
-    vm: &VirtualMachine,
-) -> PyResult {
-    Err(vm.new_not_implemented_error(
-        "Setting __dict__ attribute on an object isn't yet implemented".to_string(),
-    ))
+fn object_dict_setter(instance: PyObjectRef, value: PyDictRef, vm: &VirtualMachine) -> PyResult {
```
palaviv: Please add a snippet test for this.

vm/src/stdlib/socket.rs
```diff
@@ -27,7 +27,31 @@ type RawSocket = std::os::unix::io::RawFd;
 type RawSocket = std::os::windows::raw::SOCKET;
 
 #[cfg(unix)]
-use libc as c;
+mod c {
+    pub use libc::*;
+    // TODO: open a PR to add these constants to libc; then just use libc
```
palaviv: Sounds good :+1:

Lib/selectors.py
coolreader18: In CPython, built-in functions bind as methods to the module they're from, so `select.select` in CPython is a `method` bound to `select`. However, in RustPython, built-in functions don't do that (although that would be a good solution for accessing a function's module), so `self.select` binds `select.select` to `self` and then essentially calls `select.select(self, ...)`.

coolreader18: Nevermind, I did a hack in the VM so this isn't necessary.

vm/src/stdlib/dis.rs
coolreader18: What kind of test? It's essentially just a dict mapping ints to strings.

coolreader18:
This should be all good now @palaviv

PR #1598, opened 2019-11-16
Separate BaseException to __init__ and __new__

palaviv:
In CPython the BaseException initialize `__traceback__`, `__cause__`, `__context__` and `__suppress_context__` in `__new__` and not in `__init__`.
Do you think this attributes should be members of `PyBaseException`?


palaviv:
I decided to move everything into `PyBaseException`

PR #1603, opened 2019-11-21
Add itertools.combinations()

dralley:


vm/src/stdlib/itertools.rs
```diff
@@ -848,19 +848,133 @@ impl PyIterToolsProduct {
...
+    fn tp_new(
+        cls: PyClassRef,
+        iterable: PyObjectRef,
+        r: isize,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyRef<Self>> {
+        let iter = get_iter(vm, &iterable)?;
+        let pool = get_all(vm, &iter)?;
+
+        if r < 0 {
```
dralley: I would prefer to keep ```r``` as usize, but then I don't have control over ```ValueError``` being returned when a negative value is passed in, it comes out as ```OverflowError``` instead.  Is there a better way?

coolreader18: You could make a new type that implements `TryFromObject` that stores a `num_bigint::Sign` and a `usize`. I can see that being useful in other places as well like sequence indexing.

dralley: Ended up just copying what some of the other implementations are doing to keep it consistent.  If it's usable in many places it probably makes sense as a separate improvement.

PR #1604, opened 2019-11-22
Implement itertools.zip_longest

Space0726:
This pull request implements `itertools.zip_longest` in Python standard library, related with #1361.

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -998,11 +998,12 @@ impl PyItertoolsZiplongest {
             iterators,
             fillvalue,
             numactive,
-        }.into_ref_with_type(vm, cls)
+        }
+        .into_ref_with_type(vm, cls)
     }
 
     #[pymethod(name = "__next__")]
-    fn next(&self, vm:&VirtualMachine) -> PyResult {
+    fn next(&self, vm: &VirtualMachine) -> PyResult {
         if self.iterators.is_empty() {
             Err(new_stop_iteration(vm))
         } else {
@@ -1010,7 +1011,7 @@ impl PyItertoolsZiplongest {
             let mut result: Vec<PyObjectRef> = Vec::new();
             let mut numactive = self.numactive.clone().into_inner();
 
-            for idx in 0..self.iterators.len(){
+            for idx in 0..self.iterators.len() {
                 next_obj = match call_next(vm, &self.iterators[idx]) {
                     Ok(obj) => obj,
                     Err(_) => {
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -958,7 +958,7 @@ impl PyItertoolsCombinations {
 struct PyItertoolsZiplongest {
     iterators: Vec<PyObjectRef>,
     fillvalue: PyObjectRef,
-    numactive: RefCell<usize>,
+    numactive: Cell<usize>,
 }
 
 impl PyValue for PyItertoolsZiplongest {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -953,6 +953,87 @@ impl PyItertoolsCombinations {
...
+            Err(new_stop_iteration(vm))
+        } else {
+            let mut next_obj: PyObjectRef;
+            let mut result: Vec<PyObjectRef> = Vec::new();
+            let mut numactive = self.numactive.clone().into_inner();
+
+            for idx in 0..self.iterators.len() {
+                next_obj = match call_next(vm, &self.iterators[idx]) {
+                    Ok(obj) => obj,
+                    Err(_) => {
```
coolreader18: You might want to check if the error here is a `StopIteration` or not and handle it accordingly (if it isn't `StopIteration` we just want to pass it on):
```rust
Err(err) => {
    if !objtype::isinstance(&err, &vm.ctx.exceptions.stop_iteration) {
        return Err(err);
    }
    numactive -= 1;
    // ...
}
```

vm/src/stdlib/itertools.rs
```diff
@@ -953,6 +953,87 @@ impl PyItertoolsCombinations {
     }
 }
 
+#[pyclass]
+#[derive(Debug)]
+struct PyItertoolsZiplongest {
+    iterators: Vec<PyObjectRef>,
+    fillvalue: PyObjectRef,
+    numactive: RefCell<usize>,
```
coolreader18: This can just be a `Cell<usize>`. `Cell` is better for primitive data types that can just be copied. Then, later, you can just do `let numactive = self.numactive.get();` and `self.numactive.set(new_numactive);`

vm/src/stdlib/itertools.rs
```diff
@@ -953,6 +953,87 @@ impl PyItertoolsCombinations {
...
+    fn next(&self, vm: &VirtualMachine) -> PyResult {
+        if self.iterators.is_empty() {
+            Err(new_stop_iteration(vm))
+        } else {
+            let mut next_obj: PyObjectRef;
+            let mut result: Vec<PyObjectRef> = Vec::new();
+            let mut numactive = self.numactive.clone().into_inner();
+
+            for idx in 0..self.iterators.len() {
+                next_obj = match call_next(vm, &self.iterators[idx]) {
```
coolreader18: Do you need to have `next_obj` be mutable and declared outside the loop? I think you could just do `let next_obj = match call_next ...`.

Space0726:
@coolreader18 Thank you for your detailed review! I learned a lot from this review.

vm/src/stdlib/itertools.rs
```diff
@@ -953,6 +953,89 @@ impl PyItertoolsCombinations {
...
+        .into_ref_with_type(vm, cls)
+    }
+
+    #[pymethod(name = "__next__")]
+    fn next(&self, vm: &VirtualMachine) -> PyResult {
+        if self.iterators.is_empty() {
+            Err(new_stop_iteration(vm))
+        } else {
+            let mut result: Vec<PyObjectRef> = Vec::new();
+            let mut numactive = self.numactive.get();
```
coolreader18: Does `self.numactive` actually need to be a `Cell`? It doesn't look like you set it anywhere, but I'm not sure if it should be set at some point or if it doesn't need to be.

vm/src/stdlib/itertools.rs
```diff
@@ -953,6 +953,89 @@ impl PyItertoolsCombinations {
...
+
+            for idx in 0..self.iterators.len() {
```
coolreader18: You can make this `for iter in &self.iterators { ... }`

coolreader18:
Thanks for contributing!

PR #1606, opened 2019-11-23
Fix remove_importlib_frames

palaviv:
This will restore the functionality to before #1573.

### DIFF
--- a/vm/src/import.rs
+++ b/vm/src/import.rs
@@ -105,23 +105,37 @@ fn remove_importlib_frames_inner(
     vm: &VirtualMachine,
     tb: Option<PyTracebackRef>,
     always_trim: bool,
-) -> Option<PyTracebackRef> {
-    let traceback = tb.as_ref()?;
+) -> (Option<PyTracebackRef>, bool) {
+    if tb.is_none() {
+        return (None, false);
+    }
+    let traceback = tb.unwrap();
     let file_name = traceback.frame.code.source_path.to_string();
-    if (file_name == "_frozen_importlib" || file_name == "_frozen_importlib_external")
-        && (always_trim || traceback.frame.code.obj_name == "_call_with_frames_removed")
-    {
-        return remove_importlib_frames_inner(vm, traceback.next.as_ref().cloned(), always_trim);
+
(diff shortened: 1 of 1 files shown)


vm/src/import.rs
```diff
@@ -105,23 +105,37 @@ fn remove_importlib_frames_inner(
     vm: &VirtualMachine,
     tb: Option<PyTracebackRef>,
     always_trim: bool,
-) -> Option<PyTracebackRef> {
-    let traceback = tb.as_ref()?;
+) -> (Option<PyTracebackRef>, bool) {
+    if tb.is_none() {
+        return (None, false);
+    }
+    let traceback = tb.unwrap();
```
coolreader18: I'd say you could do something like `let traceback = if let Some(tb) = { tb } else { return (None, false) };`

vm/src/import.rs
```diff
@@ -105,23 +105,37 @@ fn remove_importlib_frames_inner(
...
     let file_name = traceback.frame.code.source_path.to_string();
-    if (file_name == "_frozen_importlib" || file_name == "_frozen_importlib_external")
-        && (always_trim || traceback.frame.code.obj_name == "_call_with_frames_removed")
-    {
-        return remove_importlib_frames_inner(vm, traceback.next.as_ref().cloned(), always_trim);
+
+    let (inner_tb, mut now_in_importlib) =
+        remove_importlib_frames_inner(vm, traceback.next.as_ref().cloned(), always_trim);
```
coolreader18: I think you can just do `traceback.next.clone()`

vm/src/import.rs
```diff
@@ -105,23 +105,37 @@ fn remove_importlib_frames_inner(
...
     tb: Option<PyTracebackRef>,
     always_trim: bool,
-) -> Option<PyTracebackRef> {
-    let traceback = tb.as_ref()?;
+) -> (Option<PyTracebackRef>, bool) {
+    if tb.is_none() {
+        return (None, false);
+    }
+    let traceback = tb.unwrap();
     let file_name = traceback.frame.code.source_path.to_string();
```
coolreader18: I think you can just borrow rather than cloning the string.

PR #1607, opened 2019-11-23
Add an actual implementation of _codecs from PyPy

coolreader18:
Given that we're taking a fair amount of stuff from PyPy, should we add its license to `Lib/`? It is just the MIT license for the most part, but still.


vm/src/vm.rs
```diff
@@ -1013,6 +1020,42 @@ impl VirtualMachine {
             })
     }
 
+    pub fn decode(
```
palaviv: There is a lot of code duplication between `decode` and `encode` how about creating some inner function?

vm/src/vm.rs
```diff
@@ -1013,6 +1020,42 @@ impl VirtualMachine {
...
+        &self,
+        obj: PyObjectRef,
+        encoding: Option<PyStringRef>,
+        errors: Option<PyStringRef>,
+    ) -> PyResult {
+        let codecsmodule = self.import("_codecs", &[], 0)?;
+        let decode = self.get_attribute(codecsmodule, "decode")?;
+        let mut args = vec![
+            obj,
+            encoding.map_or_else(|| self.get_none(), |s| s.into_object()),
```
palaviv: I think it would be more clear to do `errors.map_or_else(|| self.get_none(), |s| s.into_object())` as well instead of optionally to add them.

vm/src/obj/objstr.rs
```diff
@@ -179,17 +179,33 @@ impl PyString {
...
+        encoding: OptionalArg<PyStringRef>,
+        errors: OptionalArg<PyStringRef>,
         vm: &VirtualMachine,
     ) -> PyResult<PyStringRef> {
-        let string = match object {
-            OptionalArg::Present(ref input) => vm.to_str(input)?.into_object(),
-            OptionalArg::Missing => vm.new_str("".to_string()),
+        let string: PyStringRef = match object {
+            OptionalArg::Present(input) => {
+                if let OptionalArg::Present(enc) = encoding {
```
palaviv: This will not work in case someone do `str(..., errors=...)`

vm/src/exceptions.rs
```diff
@@ -538,24 +540,62 @@ fn import_error_init(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
...
 
-    extend_class!(context, &context.exceptions.stop_iteration, {
-        "value" => context.new_rustfunc(|obj: PyObjectRef, vm: &VirtualMachine| {
+    extend_class!(ctx, &excs.stop_iteration, {
+        "value" => ctx.new_rustfunc(|obj: PyObjectRef, vm: &VirtualMachine| {
             objiter::stop_iter_value(vm, &obj)
         }),
     });
+
+    extend_class!(ctx, &excs.unicode_decode_error, {
```
palaviv: Can we add some tests for this magic? Why is this done in Rust and not in Python?

coolreader18: What do you mean? It's properties on a built-in type, there's no reason to do it in Python. It's not really magic, just deduplicating getter code that would have been repetitive otherwise.

vm/src/vm.rs
coolreader18: If I did it like that it would throw an error because `errors` is an optional argument; it doesn't check that it's `None`, only whether or not it's present.

vm/src/function.rs
```diff
@@ -401,6 +401,13 @@ impl<T> OptionalArg<T> {
             Missing => default(),
         }
     }
+
+    pub fn opt_ref(&self) -> Option<&T> {
```
youknowone: maybe `as_opt_ref` is closer to rust naming convention?

coolreader18: I removed the function, I originally tried to implement `_codecs` in Rust and I was using it a lot more there, but now it was just 1 use so I just took it out.

coolreader18:
@youknowone here's an idea: compile `_codecs.py` to C using cython and then compile to Rust using c2rust 🙃

PR #1608, opened 2019-11-24
Travis CI: Matrix -> jobs and Python 3.6 to 3.8

cclauss:
https://docs.python.org/3/whatsnew/3.8.html#other-language-changes
> For integers, the three-argument form of the pow() function now permits the exponent to be negative in the case where the base is relatively prime to the modulus. It then computes a modular inverse to the base when the exponent is -1, and a suitable power of that inverse for other negative exponents.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_cache:
     fi
   - rm -rf ~/.cargo/registry/src
 
-matrix:
+jobs:
   fast_finish: true
   include:
     - name: Run Rust tests(linux)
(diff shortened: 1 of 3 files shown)


coolreader18:
Hi @cclauss, RustPython is actually targeting Python ~3.6, hence the CPython versions for our tests being at 3.6. We might change the target version in the future, but for now I think we're good at 3.6.

youknowone:
Once we start to target specific python version seriously, fixing a target version will be helpful to step over to goal.

For now, because RustPython is not spending any cost to keep 3.6 compatibility (comparing to 3.7+) at this stage, I think we can just move to later version of CPython if CI doesn't fail. It seems just a preference problem at this stage.

windelbouwman:
I think we should aim for 3.6 ~ 3.8 ish behavior of python. I agree that as long as CI does not fail, it is okay to move onwards.

Issue #1610, opened 2019-11-25
[WASM] default stdout dose not work as README


yanganto:
# Features
Browser devtools work as guided by the following instructions in the documentation.
> Please input your Python code below and click Run (or Ctrl+Enter), __or you can open up your browser's devtools and play with rp.pyEval('print("a")')__

## Expected Result
```
>> rp.pyEval('print("a")')
"a"
← undefined
```

## Actual Result
Firefox 70.0.1（64 位元）
```
>> rp.pyEval('print("a")')
← undefined
```

Chromium 78.0.3904.97 (正式版本) Arch Linux (64 位元)
```
>> rp.pyEval('print("a")')
← undefined
```

yanganto:
Update:
These methods work.
`rp.pyEval('print("a")',{stdout: (s) => console.log(s)})`
or 
`rp.pyEval('print("a")',{stdout: (s) => alert(s)})`

PR #1611, opened 2019-11-25
Set default stdout of wasm to `console.log`

yanganto:
Fix issue #1610 

The print function of wasm will be
- `console.log` by default or when giving undefined or "console"
- a dumb function when giving `null`
- the function when giving a function


yanganto:
The behaviors in firefox 70.0.1 are listed as following.

```
>> rp.pyEval('print("a")',{stdout: 'console'})
"a"
← undefined
>> rp.pyEval('print("a")',{stdout: null})
← undefined
>> rp.pyEval('print("a")')
"a"
← undefined
```

yanganto:
Thanks for your review.

Issue #1588, opened 2019-11-07
fetch snippet broken in wasm demo: '_browsers' is not a built-in module


silmeth:
On the [wasm demo](https://rustpython.github.io/demo/) the `fetch` snippet currently errors with the `Error: "'_browser' is not a built-in module"` message.

At least on Firefox 70.0.1 on Linux, but I guess it’s not a browser-related issue but rather some regression related to built-in modules changes.

yanganto:
There is a hacky way to work around on this issue.

Before import `browser`, fix the the tuple of builtin module first.
```
import sys
sys.builtin_module_names = ('_js', '_window', '_browser', *sys.builtin_module_names)
```

This may be close to the root cause, it will be fixed when the processes of injecting the browser module and the js module are corrected.

coolreader18:
Yeah, rustpython-wasm can definitely use a lot of work. I had plans to flesh out a lot of the Python/JS interop, but I never got around to it. There's still a lot more work to be done on the `JsValue` class (first and foremost that it should probably be called `JSValue`, looking back on it), including formalizing how security/sandboxing should work: right now, if you have access to just one `JSValue`, it's possible to get access to the `Object` prototype and define new methods on it, which is a security problem for someone who might want to use rustpython-wasm to run completely untrusted Python code:
```py
from _window import window
object_proto = window.new_object().get_prop('__proto__')
object_proto.set_prop('property', window.new_from_float(12.0))
```
and now in JS:
```js
({}).property // 12.0
```
Or, you could get access to the function constructor, which would allow you to get the window object and make a `fetch` request:
```py
from _window import window
# I know we already have `window` here, but imagine we just get some random object
js_obj = window.new_object()
del window
object_proto = js_obj.new_object().get_prop('__proto__')
Function = object_proto.get_prop("toString").get_prop("constructor")
window = Function.call(js_obj.new_from_str("return this")).call()
```

I also wanted to make a proxy class for `JSValue`s to make it easier to work with them, e.g. `__getattr__` would call `get_prop`, `__call__` would call `call`, etc.

If you @yanganto (or anyone else) wanted to work on any of this stuff, feel free to.

yanganto:
This may not easy to implement perfectly at once, but sounds interesting.  I am willing to try and may ask some questions on Gitter when running into problems. Thank @coolreader18 .

yanganto:
I got an idea of sandboxing.
Maybe we can construct an option to decide python code to access the javascript object

[...]

PR #1614, opened 2019-11-29
Update sys module when stdlib_inits is changed

yanganto:
Provide `update_sysmodule` in vm to update sys when update_sysmodule
changed

Fix issue #1588

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -7,7 +7,7 @@ extern crate log;
 use clap::{App, AppSettings, Arg, ArgMatches};
 use rustpython_compiler::compile;
 use rustpython_vm::{
-    import, match_class,
+    match_class,
     obj::{objint::PyInt, objtuple::PyTuple, objtype},
     print_exception,
     pyobject::{ItemProtocol, PyResult},
@@ -30,8 +30,9 @@ fn main() {
     let app = App::new("RustPython");
     let matches = parse_arguments(app);
     let settings = create_settings(&matches);
-    let vm = VirtualMachine::new(settings);
+    let mut uninitialized_vm = VirtualMachine::new(settings);
 
+    let vm = uninitialized_vm.initialize(true);
     let res = run_rustpython(&vm, &matches);
 
     #[cfg(feature = "flame-it")]
@@ -324,9 +325,6 @@ fn write_profile(matches: &ArgMatches) -> Result<(), Box<dyn std::error::Error>>
 }
 
 fn run_rustpython(vm: &VirtualMachine, matches: &ArgMatches) -> PyResult<()> {
-    // We only include the standard library bytecode in WASI
-    import::init_importlib(&vm, cfg!(not(target_os = "wasi")))?;
-
     if let Some(paths) = option_env!("BUILDTIME_RUSTPYTHONPATH") {
         let sys_path = vm.get_attribute(vm.sys_module.clone(), "path")?;
         for (i, path) in std::env::split_paths(paths).enumerate() {
@@ -446,7 +444,8 @@ fn run_script(vm: &VirtualMachine, scope: Scope, script_file: &str) -> PyResult<
 
 #[test]
 fn test_run_script() {
-    let vm: VirtualMachine = Default::default();
+    let mut uninitialized_vm: VirtualMachine = Default::default();
+    let vm = uninitialized_vm.initialize(true);
 
     // test file run
     let r = run_script(
--- a/src/main.rs
+++ b/src/main.rs
@@ -12,7 +12,7 @@ use rustpython_vm::{
     print_exception,
     pyobject::{ItemProtocol, PyResult},
     scope::Scope,
-    util, PySettings, VirtualMachine,
+    util, InitParameter, PySettings, VirtualMachine,
 };
 
 use std::convert::TryInto;
@@ -32,7 +32,9 @@ fn main() {
     let mut settings = create_settings(&matches);
 
     // We only include the standard library bytecode in WASI when initializing
-    settings.initialize_with_external_importer = Some(cfg!(not(target_os = "wasi")));
+    if cfg!(target_os = "wasi") {
+        settings.initialization_parameter = InitParameter::InitializeInternal;
+    }
 
     let vm = VirtualMachine::new(settings);
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -29,7 +29,11 @@ fn main() {
     env_logger::init();
     let app = App::new("RustPython");
     let matches = parse_arguments(app);
-    let settings = create_settings(&matches);
+    let mut settings = create_settings(&matches);
+
+    // We only include the standard library bytecode in WASI when initializing
+    settings.initialize_with_external_importer = Some(cfg!(not(target_os = "wasi")));
+
     let vm = VirtualMachine::new(settings);
 
     let res = run_rustpython(&vm, &matches);
@@ -324,9 +328,6 @@ fn write_profile(matches: &ArgMatches) -> Result<(), Box<dyn std::error::Error>>
 }
 
 fn run_rustpython(vm: &VirtualMachine, matches: &ArgMatches) -> PyResult<()> {
-    // We only include the standard library bytecode in WASI when initializing
-    vm.initialize(cfg!(not(target_os = "wasi")))?;
-
     if let Some(paths) = option_env!("BUILDTIME_RUSTPYTHONPATH") {
         let sys_path = vm.get_attribute(vm.sys_module.clone(), "path")?;
         for (i, path) in std::env::split_paths(paths).enumerate() {
--- a/src/main.rs
+++ b/src/main.rs
@@ -30,9 +30,8 @@ fn main() {
     let app = App::new("RustPython");
     let matches = parse_arguments(app);
     let settings = create_settings(&matches);
-    let mut uninitialized_vm = VirtualMachine::new(settings);
+    let vm = VirtualMachine::new(settings);
 
-    let vm = uninitialized_vm.initialize(true);
     let res = run_rustpython(&vm, &matches);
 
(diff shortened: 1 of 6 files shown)


vm/src/vm.rs
```diff
@@ -207,6 +207,12 @@ impl VirtualMachine {
         vm
     }
 
+    /// `update_sysmodule` should be called when the `vm.stdlib_inits` are changed,
+    /// and then the `builtin_module_names` of `sys` module will be updated.
+    pub fn update_sysmodule(&self) {
+        sysmodule::make_module(self, self.sys_module.clone(), self.builtins.clone());
```
coolreader18: I feel like this might have unintended side effects, it might be better to explicitly update `builtin_module_names` than call all of `sysmodule::make_module`

yanganto: Maybe we can define a stage `initialize` after VM created.
And put `init_importlib`, `sysmodule::make_module` and `builtins::make_module` in `initialize`.  Before `initialize` the `stdlib_inits` can be changed base on different environments.

coolreader18:
I'm not sure about allowing the VM be in an uninitialized state, but I suppose we already do have that in a sense with imports.

vm/src/vm.rs
```diff
@@ -198,13 +198,27 @@ impl VirtualMachine {
             vm.get_none(),
         );
 
-        builtins::make_module(&vm, builtins.clone());
-        sysmodule::make_module(&vm, sysmod, builtins);
+        vm
+    }
+
+    pub fn initialize(&mut self, external: bool) -> &Self {
```
coolreader18: I think this could return a `PyResult<()>`, since we already have enough of the VM to be able to print errors, and it is helpful occasionally to see an error that came out of `importlib` initialization or similar. Also, does it need to be `&mut self` instead of `&self`?

vm/src/vm.rs
```diff
@@ -198,13 +198,27 @@ impl VirtualMachine {
...
+        sysmodule::make_module(self, self.sys_module.clone(), self.builtins.clone());
 
         #[cfg(not(target_arch = "wasm32"))]
-        import::import_builtin(&vm, "signal").expect("Couldn't initialize signal module");
+        import::import_builtin(self, "signal").expect("Couldn't initialize signal module");
 
-        vm
+        if external {
+            // We only include the standard library bytecode in WASI
+            import::init_importlib(self, cfg!(not(target_os = "wasi")))
```
coolreader18: I think this `cfg` should be in `src/main.rs`, not here. Then you can just pass `external` to `init_importlib`

src/main.rs
```diff
@@ -444,7 +442,8 @@ fn run_script(vm: &VirtualMachine, scope: Scope, script_file: &str) -> PyResult<
 
 #[test]
 fn test_run_script() {
-    let vm: VirtualMachine = Default::default();
+    let mut uninitialized_vm: VirtualMachine = Default::default();
+    let vm = uninitialized_vm.initialize(true);
```
coolreader18: Maybe have `VirtualMachine::default()` initialize the VM and panic if it fails? That would fix small use cases like this.

src/main.rs
```diff
@@ -30,8 +30,9 @@ fn main() {
     let app = App::new("RustPython");
     let matches = parse_arguments(app);
     let settings = create_settings(&matches);
-    let vm = VirtualMachine::new(settings);
+    let mut uninitialized_vm = VirtualMachine::new(settings);
 
+    let vm = uninitialized_vm.initialize(true);
```
coolreader18: You can just do
```rust
let vm = VirtualMachine::new(settings);
vm.initialize(cfg!(not(target_os = "wasi")));
```

vm/src/eval.rs
```diff
@@ -22,7 +22,9 @@ mod tests {
     #[test]
     fn test_print_42() {
         let source = String::from("print('Hello world')");
-        let vm = VirtualMachine::default();
+        let mut uninitialized_vm = VirtualMachine::default();
+        let vm = uninitialized_vm.initialize(true);
```
windelbouwman: Instead of this construction, I would rather see that the `default` function will initialize the `VirtualMachine`. If you want a `VirtualMachine` without initialization, create a new function `VirtualMachine::without_external` which will give the virtual machine without a the external initialization.

src/main.rs
windelbouwman: You could add a setting to the `Settings` object, having a boolean to initialize the vm yes or no.

windelbouwman:
> I'm not sure about allowing the VM be in an uninitialized state, but I suppose we already do have that in a sense with imports.

Agreed, this sounds like a tricky plan.

src/main.rs
yanganto: Thanks for the review.  I put the `initialize` into `default`, and new a vm, setup stdlib_inits than initialize the vm in a customized way.

yanganto: The vm will be initialized when created as default.  And the vm will be initialized after stdlib_inits changed, by putting the flag into settings.

vm/src/eval.rs
yanganto: Thanks for the review.  The default will initialize the vm as normal, and the vm in wasm will be initialized by putting the flag into settings.

yanganto: I takes the suggestion and put the initialize parameters in the settings, such that the vm can also be created and initialized without external.

vm/src/vm.rs
```diff
@@ -197,14 +205,22 @@ impl VirtualMachine {
             vm.new_str("sys".to_owned()),
             vm.get_none(),
         );
+        if let Some(as_external) = initialize_parameter {
+            vm.initialize(as_external);
+        }
+        vm
+    }
+
+    pub fn initialize(&self, external: bool) {
```
windelbouwman: Could this method be private, to prevent double initialization from elsewhere?

vm/src/vm.rs
```diff
@@ -106,6 +106,12 @@ pub struct PySettings {
 
     /// sys.argv
     pub argv: Vec<String>,
+
+    /// Initialize VM after VM created
```
windelbouwman: This doc strings are definitely a good thing, but still I do not understand the logic of external initialization. Could you explain a bit more what this is and what it's effects are?

wasm/lib/src/vm_class.rs
```diff
@@ -27,7 +26,21 @@ pub(crate) struct StoredVirtualMachine {
 
 impl StoredVirtualMachine {
     fn new(id: String, inject_browser_module: bool) -> StoredVirtualMachine {
-        let mut vm: VirtualMachine = Default::default();
+        let mut vm: VirtualMachine = VirtualMachine::new(PySettings {
```
windelbouwman: In this case, I would rather construct a mutable default settings object, and flip the `initialize_as_external` value to false. This would ease the addition of extra settings in the future.

wasm/lib/src/vm_class.rs
```diff
@@ -44,7 +57,7 @@ impl StoredVirtualMachine {
             setup_browser_module(&vm);
         }
 
-        import::init_importlib(&vm, false).unwrap();
+        vm.initialize(false);
```
windelbouwman: This call can be removed. Set the setting `initialize_as_external` to `Some(false)`.

vm/src/vm.rs
yanganto: Thanks for the review.

In wasm, If the VM is initialized and then injected, these bugs will still there.
The VM should be injected before initialized, such that the `_browser`, `_js` can be correctly initialized.  so the `initialize` method is public.

The root cause is that the `builtin_module_names` is set in `make_module` according to the `stdlib_inits` in the timing of initializing.

wasm/lib/src/vm_class.rs
yanganto: The VM initialize needs to after then`js_module::setup_js_module` and `setup_browser_module`.  So this public function calls here and not in the timing of VM creating.

In wasm, the process is creating VM, injecting, and then initialize as external.

The `initialize(false)` is misleading I will change this to `initalize_as_external` latter.

vm/src/vm.rs
yanganto: The `initialized` flag is now used to prevent double initialization.  Thanks for the view.

vm/src/vm.rs
yanganto: The docstrings are modified and explain about initialization with the importers require an external file system or not.  Thanks for point out.  I am not truly understood the flow before, but it much clear now.

wasm/lib/src/vm_class.rs
yanganto: The `initialized_as_external` may be ambiguous, the `initialize_with_external_importer` is adopted for clearness.
Thanks.

vm/src/vm.rs
```diff
@@ -107,6 +108,18 @@ pub struct PySettings {
...
+    /// - Some(true):
+    ///     VM will be created and initialized with the importers, which require external
+    ///     filesystem access
+    /// - Some(false):
+    ///     VM will be created and initialized with the importers wihout external filesystem access
+    /// - None:
+    ///     VM will be created but not initialized, so it is ready to be injected with modules.
+    ///     After injection, the `initialize_as_external` or `initialize_without_external`
+    ///     should be called, such that the VM will be ready to use.
+    pub initialize_with_external_importer: Option<bool>,
```
coolreader18: Maybe this could be its own enum to make the option more clear?
```rust
enum Initialization { NoInitialize, InitializeInternal, InitializeExternal }
```

yanganto: Thanks for your advice.  Use enum is much clear in this case.

coolreader18:
@windelbouwman I think this is in a good state; should we merge?

yanganto:
Thanks for the review.

PR #1616, opened 2019-11-30
add some os stdlib functions

maybe-rachel:
Specifically: 
- os.pipe
- os.pipe2
- os.get_blocking
- os.set_blocking
- os.get_inheritable
- os.set_inheritable

Sorry for adding so many on my first PR. I was just planning on doing pipe at first, and kinda got dragged into the rest for testing.

Regarding os.pipe (~line 1000), is it okay to ignore close(fd) failure? That's what CPython does, and I'm not otherwise sure how to appropriately handle that. 

Also, flake8 complains about stdlib_os.py a lot. Not my code, and I didn't want to run a linter and make my first PR seem bigger.

### DIFF
--- a/tests/snippets/stdlib_os.py
+++ b/tests/snippets/stdlib_os.py
@@ -251,69 +251,6 @@ with TestWithTempDir() as tmpdir:
 	assert os.path.basename(fname) == FILE_NAME
 	assert os.path.dirname(fname) == tmpdir
 
-	# os.get_blocking, os.set_blocking
-	# TODO: windows support should be added for below functions
-	# os.pipe, 
-	# os.set_inheritable, os.get_inheritable, 
-	if os.name != "nt":
-		rfd, wfd = os.pipe()
-		try:
-			os.write(wfd, CONTENT2)
-			assert os.read(rfd, len(CONTENT2)) == CONTENT2
-			assert not os.get_inheritable(rfd)
-			assert not os.get_inheritable(wfd)
-			os.set_inheritable(rfd, True)
-			os.set_inheritable(wfd, True)
-			assert os.get_inheritable(rfd)
-			assert os.get_inheritable(wfd)
-			os.set_inheritable(rfd, True)
-			os.set_inheritable(wfd, True)
-			os.set_inheritable(rfd, True)
-			os.set_inheritable(wfd, True)
-			assert os.get_inheritable(rfd)
-			assert os.get_inheritable(wfd)
-
-			assert os.get_blocking(rfd)
-			assert os.get_blocking(wfd)
-			os.set_blocking(rfd, False)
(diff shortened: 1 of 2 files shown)


tests/snippets/stdlib_os.py
```diff
@@ -251,6 +251,69 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 	assert os.path.basename(fname) == FILE_NAME
 	assert os.path.dirname(fname) == tmpdir
 
+	# os.get_blocking, os.set_blocking
+	# TODO: windows support should be added for below functions
+	# os.pipe, 
+	# os.set_inheritable, os.get_inheritable, 
+	if os.name != "nt":
+		rfd, wfd = os.pipe()
```
coolreader18: I don't think these tests have to be in this `with` block, as they don't use `tmpdir` for anything

vm/src/stdlib/os.rs
```diff
@@ -925,6 +927,102 @@ fn os_chdir(path: PyStringRef, vm: &VirtualMachine) -> PyResult<()> {
...
+    };
+    _set_flag().or_else(|err| Err(convert_nix_error(vm, err)))
+}
+
+#[cfg(unix)]
+fn os_pipe(vm: &VirtualMachine) -> PyResult<(RawFd, RawFd)> {
+    use nix::unistd::close;
+    use nix::unistd::pipe;
+    match pipe() {
+        Err(err) => Err(convert_nix_error(vm, err)),
```
coolreader18: I think you could do something like `let (rfd, wfd) = pipe().map_err(|err| ...)?;`

vm/src/stdlib/os.rs
```diff
@@ -925,6 +927,102 @@ fn os_chdir(path: PyStringRef, vm: &VirtualMachine) -> PyResult<()> {
...
+        Ok(ok) => {
+            os_set_inheritable(ok.0, false, vm)
+                .and_then(|_| os_set_inheritable(ok.1, false, vm))
+                .or_else(|err| {
+                    close(ok.0)
```
coolreader18: You could do `let _ = close(...);` to ignore the result. I think ignoring the result is fine; Rust does the same thing when it calls `drop()` on a `std::fs::File`.

maybe-rachel:
> Thank you for contributing @maybe-rachel! I've left a few comments, but overall this looks very good.

Thanks for all your suggestions! I did all those. Also, I converted all tabs to spaces because tabs were breaking it, so the diff's a bit bigger than it should be :X

coolreader18:
Would you be able to find a way to not change all the indentation? It would just be nice for some git tools like blame.

coolreader18:
Thanks again for contributing!

PR #1617, opened 2019-11-30
Fix some various issues with `os` and `binascii`

coolreader18:
~~This might be a lot of miscellaneous changes, but I have a lot of other miscellaneous changes and I don't know the best way to split them all into PRs. If you're reviewing, I'd recommend going commit by commit.~~ I split it up

### DIFF
--- a/vm/src/obj/objbyteinner.rs
+++ b/vm/src/obj/objbyteinner.rs
@@ -1418,7 +1418,6 @@ fn split_slice_reverse<'a>(slice: &'a [u8], sep: &[u8], maxsplit: i32) -> Vec<&'
 pub enum PyBytesLike {
     Bytes(PyBytesRef),
     Bytearray(PyByteArrayRef),
-    Vec(Vec<u8>),
 }
 
 impl TryFromObject for PyBytesLike {
@@ -1426,8 +1425,6 @@ impl TryFromObject for PyBytesLike {
         match_class!(match obj {
             b @ PyBytes => Ok(PyBytesLike::Bytes(b)),
             b @ PyByteArray => Ok(PyBytesLike::Bytearray(b)),
-            m @ PyMemoryView => Ok(PyBytesLike::Vec(m.get_obj_value().unwrap())),
-            l @ PyList => Ok(PyBytesLike::Vec(l.get_byte_inner(vm)?.elements)),
             obj => Err(vm.new_type_error(format!(
                 "a bytes-like object is required, not {}",
                 obj.class()
@@ -1441,7 +1438,14 @@ impl PyBytesLike {
         match self {
             PyBytesLike::Bytes(b) => b.get_value().into(),
             PyBytesLike::Bytearray(b) => b.inner.borrow().elements.clone().into(),
-            PyBytesLike::Vec(b) => b.as_slice().into(),
+        }
+    }
+
+    #[inline]
+    pub fn with_ref<R>(&self, f: impl FnOnce(&[u8]) -> R) -> R {
+        match self {
+            PyBytesLike::Bytes(b) => f(b.get_value()),
+            PyBytesLike::Bytearray(b) => f(&b.inner.borrow().elements),
         }
     }
 }
--- a/vm/src/stdlib/binascii.rs
+++ b/vm/src/stdlib/binascii.rs
@@ -1,8 +1,52 @@
 use crate::function::OptionalArg;
-use crate::obj::objbytes::PyBytesRef;
-use crate::pyobject::{PyObjectRef, PyResult};
+use crate::obj::objbytearray::{PyByteArray, PyByteArrayRef};
+use crate::obj::objbyteinner::PyBytesLike;
+use crate::obj::objbytes::{PyBytes, PyBytesRef};
+use crate::obj::objstr::{PyString, PyStringRef};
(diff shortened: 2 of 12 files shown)


vm/src/obj/objbyteinner.rs
```diff
@@ -472,6 +472,140 @@ impl PyByteInner {
         }
     }
 
+    pub fn delitem(
```
coolreader18: This is mostly just copied from list.__delitem__

coolreader18:
@palaviv I fixed CI; could you review now?

vm/src/stdlib/binascii.rs
```diff
@@ -32,37 +76,45 @@ fn unhex_nibble(c: u8) -> Option<u8> {
...
 
-    Ok(vm.ctx.new_bytes(unhex))
+        Ok(unhex)
+    })
 }
 
-fn binascii_crc32(data: PyBytesRef, value: OptionalArg<u32>, vm: &VirtualMachine) -> PyResult {
-    let bytes = data.get_value();
-    let crc = value.unwrap_or(0u32);
+fn binascii_crc32(data: SerializedData, value: OptionalArg<u32>, vm: &VirtualMachine) -> PyResult {
```
palaviv: Could you add some tests for these function with `SerializedData`

vm/src/stdlib/binascii.rs
```diff
@@ -72,5 +124,7 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         "unhexlify" => ctx.new_rustfunc(binascii_unhexlify),
         "a2b_hex" => ctx.new_rustfunc(binascii_unhexlify),
         "crc32" => ctx.new_rustfunc(binascii_crc32),
+        "a2b_base64" => ctx.new_rustfunc(binascii_a2b_base64),
```
palaviv: Please add some tests to the new functions

vm/src/stdlib/os.rs
```diff
@@ -712,80 +671,58 @@ fn to_seconds_from_nanos(secs: i64, nanos: i64) -> f64 {
...
-
-        get_stats($path.as_str(), $follow_symlinks.follow_symlinks)
-            .map_err(|err| convert_io_error($vm, err))
-    }};
-}
-
-#[cfg(target_os = "linux")]
 fn os_stat(
-    path: PyStringRef,
+    file: Either<PyStringRef, i64>,
```
palaviv: I think we should create a `PathLike` object as this will probably be used in many places.

vm/src/stdlib/os.rs
```diff
@@ -347,17 +348,21 @@ fn os_fsync(fd: i64, vm: &VirtualMachine) -> PyResult<()> {
 fn os_read(fd: i64, n: usize, vm: &VirtualMachine) -> PyResult {
     let mut buffer = vec![0u8; n];
     let mut file = rust_file(fd);
-    file.read_exact(&mut buffer)
+    let n = file
+        .read(&mut buffer)
         .map_err(|err| convert_io_error(vm, err))?;
+    buffer.truncate(n);
```
palaviv: Nice

vm/src/stdlib/os.rs
```diff
@@ -1253,17 +1276,23 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         "cpu_count" => ctx.new_rustfunc(os_cpu_count),
         "_exit" => ctx.new_rustfunc(os_exit),
         "urandom" => ctx.new_rustfunc(os_urandom),
+        "isatty" => ctx.new_rustfunc(os_isatty),
```
palaviv: Please add some tests to the new functions

vm/src/stdlib/os.rs
```diff
@@ -1138,6 +1143,24 @@ fn os_urandom(size: usize, vm: &VirtualMachine) -> PyResult<Vec<u8>> {
     }
 }
 
+fn os_isatty(fd: i32, _vm: &VirtualMachine) -> bool {
+    unsafe { libc::isatty(fd) != 0 }
+}
+
+fn os_lseek(
+    fd: libc::c_int,
```
palaviv: Not sure about using the `libc` type as arguments. It is unclear to me what type is `libc::c_int` vs libc::c_long` Maybe we should do an explicit conversion?

vm/src/stdlib/os.rs
coolreader18: I really like that idea, but I think right now this construct is used in few enough places that it isn't worth it to extract it like that, and also for `os_stat` the code for `fd`s and paths is so different that it wouldn't really deduplicate that much.

coolreader18:
Ok, the error code I'm getting on Windows is a stack overflow, so I have no idea why that's happening...

vm/src/stdlib/os.rs
coolreader18: I think it's ok b/c even if they are the same type on a given platform, this function is just a thin wrapper over `lseek` from libc so it's easiest to use the types as they are defined in the `libc` signature.

coolreader18:
Alright it was from an invalid parameter error (CPython has `_Py_{BEGIN,END}_SUPPRESS_IPH` for this), so I made a macro to ignore it.

coolreader18:
Well that didn't work either so I just disabled the test on windows.

PR #1618, opened 2019-11-30
Some miscellaneous changes (print fix, bytes/bytearray fixes, mappingproxy iter)

coolreader18:


vm/src/obj/objbyteinner.rs
```diff
@@ -472,6 +472,140 @@ impl PyByteInner {
         }
     }
 
+    pub fn delitem(
```
coolreader18: Most of this code is taken from `list.__delitem__`, it could be deduplicated but I think this is a case of "duplicated code is better than the wrong abstraction", and atm we only have 2 copies of this basic code.

PR #1621, opened 2019-11-30
Update the io module with a bunch of methods

coolreader18:


vm/src/stdlib/io.rs
```diff
@@ -184,13 +208,15 @@ fn string_io_new(
...
         buffer: RefCell::new(BufferedIO::new(Cursor::new(raw_string.into_bytes()))),
+        closed: Cell::new(false),
     }
     .into_ref_with_type(vm, cls)
 }
 
 #[derive(Debug)]
 struct PyBytesIO {
     buffer: RefCell<BufferedIO>,
+    closed: Cell<bool>,
```
youknowone: `Option<BufferedIO>` for less effort of consistency of `buffer` and `closed`?

vm/src/stdlib/io.rs
```diff
@@ -115,27 +114,36 @@ impl PyValue for PyStringIO {
 }
 
 impl PyStringIORef {
+    fn buffer(&self, vm: &VirtualMachine) -> PyResult<RefMut<BufferedIO>> {
+        let buffer = self.buffer.borrow_mut();
+        if buffer.is_some() {
```
youknowone: is this not fit for `if let Some(buffer) = ...` pattern?

coolreader18: No, because `Ref::map` can only return an `&mut T` from the borrowed value. I think we have a similar pattern to this with `vm.current_frame()`.

PR #1625, opened 2019-12-13
Support expression in f-strings spec

palaviv:
This allows:
```py
spec = "0>+#10x"
f"{16:{spec}}"
```

### DIFF
--- a/parser/src/fstring.rs
+++ b/parser/src/fstring.rs
@@ -51,12 +51,11 @@ impl<'a> FStringParser<'a> {
                             '{' => {
                                 if in_nested {
                                     return Err(ExpressionNestedTooDeeply);
-                                } else {
-                                    in_nested = true;
-                                    nested = true;
-                                    self.chars.next();
-                                    continue;
                                 }
+                                in_nested = true;
+                                nested = true;
+                                self.chars.next();
+                                continue;
                             }
                             '}' => {
                                 if in_nested {
@@ -295,6 +294,9 @@ mod tests {
     fn test_parse_invalid_fstring() {
         assert_eq!(parse_fstring("{"), Err(UnclosedLbrace));
         assert_eq!(parse_fstring("}"), Err(UnopenedRbrace));
+        assert_eq!(parse_fstring("{a:{a:{b}}"), Err(ExpressionNestedTooDeeply));
+        assert_eq!(parse_fstring("{a:b}}"), Err(UnopenedRbrace));
+        assert_eq!(parse_fstring("{a:{b}"), Err(UnclosedLbrace));
 
         // TODO: check for InvalidExpression enum?
         assert!(parse_fstring("{class}").is_err());
--- a/parser/src/fstring.rs
+++ b/parser/src/fstring.rs
@@ -23,7 +23,7 @@ impl<'a> FStringParser<'a> {
 
     fn parse_formatted_value(&mut self) -> Result<StringGroup, FStringErrorType> {
         let mut expression = String::new();
-        let mut spec = String::new();
+        let mut spec = None;
         let mut delims = Vec::new();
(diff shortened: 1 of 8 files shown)


parser/src/fstring.rs
```diff
@@ -209,6 +245,42 @@ mod tests {
...
+        let parse_ast = parse_fstring(&source).unwrap();
+
+        assert_eq!(
+            parse_ast,
+            FormattedValue {
+                value: Box::new(mk_ident("foo", 1, 1)),
+                conversion: None,
+                spec: Some(Box::new(Constant {
+                    value: "spec".to_string(),
+                })),
```
coolreader18: Could you add some more tests for some weirder cases? e.g. `f"{val: {'#'} }"` interprets the spec as `' {\'#\'} '`; it only interpolates the value if the only characters for the spec are `:{expression}}`

parser/src/fstring.rs
```diff
@@ -43,13 +43,49 @@ impl<'a> FStringParser<'a> {
...
-                        if next != '}' {
-                            spec.push(next);
-                            self.chars.next();
-                        } else {
-                            break;
+                        match next {
+                            '{' => {
+                                if in_nested {
+                                    return Err(ExpressionNestedTooDeeply);
+                                } else {
```
coolreader18: You don't need this `else`, as you return in the first condition.

tests/snippets/fstrings.py
```diff
@@ -16,6 +16,9 @@
 assert f'{16:0>+#10x}' == '00000+0x10'
 assert f"{{{(lambda x: f'hello, {x}')('world}')}" == '{hello, world}'
 
+spec = "0>+#10x"
+assert f"{16:{spec}}{foo}" == '00000+0x10bar'
+
```
coolreader18: Could you add some TypeError tests here for formatting with `int`s as specs or similar?

palaviv: Could you be more specific?

parser/src/fstring.rs
palaviv: Done

parser/src/fstring.rs
palaviv: I know this is not perfect yet but we can improve on that later. I added a few more fail cases tests.

tests/snippets/fstrings.py
coolreader18: e.g. `assert_raises(TypeError, lambda: f"{123:{45}}")`

palaviv: > e.g. `assert_raises(TypeError, lambda: f"{123:{45}}")`

This is actually a valid format string.
I added a TODO in the tests for an invalid one. This currently does not fail but it is not related to this change.

doyshinda:
Hey, I have a PR for supporting `:f` string format codes in #1651 and I noticed that after rebasing with these changes, one of my new tests is now failing with `f-string` style formatting:
```Python
assert f'{5.1234: f}' == ' 5.123400'
```

When using the `f` style formatting, the `" "` prefix is missing:
```Python
Welcome to the magnificent Rust Python 0.1.1 interpreter 😱 🖖
>>>>> f'{5.1234: f}'
'5.123400'
>>>>> '{: f}'.format(5.1234)
' 5.123400'
>>>>> '{x: f}'.format_map({'x': 5.1234})
' 5.123400'
>>>>>
```

coolreader18:
It looks like that came from line 86 in `parser/sec/fstring.rs`:

```rust
spec = Some(Box::new(Constant {
    value: spec_expression.trim().to_string(),
}))
```

If you remove the `.trim()` in your branch, that should fix it.

doyshinda:
Confirmed, removing `trim()` fixed it, thanks!

PR #1626, opened 2019-12-14
Refacter objint

youknowone:
- `objint::get_float_value` -> `objfloat::from_int_value` because objint::get_float_value is only used in objfloat
- `PyInt::cmp`, `PyInt::int_op`, `PyInt::general_op` to make it simple

### DIFF
--- a/vm/src/obj/objbool.rs
+++ b/vm/src/obj/objbool.rs
@@ -7,7 +7,7 @@ use crate::pyobject::{
 };
 use crate::vm::VirtualMachine;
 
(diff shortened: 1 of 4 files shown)


youknowone:
Anything more to be merged?

coolreader18:
What do you mean?

youknowone:
Because it is only approved but not merged, so I wonder if you think any further work or review is required.

coolreader18:
Ah no, I think this is good to merge.

PR #1627, opened 2019-12-14
Modify os window check logic in test code

khg0712:
current os check logic in test code can't distinguish window and unix.
I solve this problem and make code more clearly


youknowone:
I am sorry to not to finish this on time. but this is now outdated too much

PR #1631, opened 2019-12-14
implement SyntaxError attributes

youknowone:


vm/src/vm.rs
```diff
@@ -399,7 +399,7 @@ impl VirtualMachine {
     }
 
     #[cfg(feature = "rustpython-compiler")]
-    pub fn new_syntax_error(&self, error: &CompileError) -> PyObjectRef {
+    pub fn new_syntax_error(&self, error: &CompileError, filename: Option<&str>) -> PyObjectRef {
```
windelbouwman: As an idea, add the field `filename: Option<String>` to `CompileError`. This save a lot of passing around the filename. Could this work?

youknowone: That sounds better. I will try it.

PR #1632, opened 2019-12-15
Implement built-in input

### LINKED ISSUE #1586: Implement input()

x86chi:
> close #1586 

Please read below commits


src/main.rs
```diff
@@ -7,12 +7,12 @@ extern crate log;
 use clap::{App, AppSettings, Arg, ArgMatches};
 use rustpython_compiler::compile;
 use rustpython_vm::{
-    import, match_class,
-    obj::{objint::PyInt, objtuple::PyTuple, objtype},
-    print_exception,
-    pyobject::{ItemProtocol, PyResult},
-    scope::Scope,
-    util, PySettings, VirtualMachine,
+	import, match_class,
```
youknowone: This formatting change doesn't look normal. Will you please update your rust by using `rustup` and run `cargo fmt --all` to format the code?

x86chi: I've change indent to 4 space

vm/src/vm.rs
```diff
@@ -185,6 +207,8 @@ impl VirtualMachine {
             settings,
             recursion_limit: Cell::new(512),
             codec_registry: RefCell::default(),
+            repl: RefCell::new(repl),
+            repl_history_path,
```
x86chi: Detail about force commit:
`repl_history_path: repl_history_path` to `repl_history_path`

coolreader18:
I made this comment on the previous PR:

> I was thinking that you could move shell.rs into vm/src , and just create a new repl for each call to builtin_input rather than storing it in the VM.

youknowone:
@x86chi will you rebase this PR to `upstream/master` and check the comment from @coolreader18 ?

vm/src/builtins.rs
```diff
@@ -345,7 +348,35 @@ fn builtin_id(obj: PyObjectRef, _vm: &VirtualMachine) -> usize {
     obj.get_id()
 }
 
-// builtin_input
+fn builtin_input(prompt_option: OptionalArg<PyObjectRef>, vm: &VirtualMachine) -> PyResult {
```
youknowone: `prompt` will be a good name

vm/src/builtins.rs
```diff
@@ -345,7 +348,35 @@ fn builtin_id(obj: PyObjectRef, _vm: &VirtualMachine) -> usize {
...
+    let prompt = if let OptionalArg::Present(prompt_object) = prompt_option {
```
youknowone: I think this value can be `&str` rather than `String`.

```rust
let prompt = prompt_option.map_or("", |obj| objstr::borrow_value(obj));
```

windelbouwman:
@x86chi thank you for contributing to this project!

Issue #1596, opened 2019-11-15
[RFC] Add CPython tests


palaviv:
## Summary

Now that we can run `unittest` ( #1587 ) I think we should start running the CPython tests (Lib/test) to test RustPython.

## Detailed Explanation

My suggestion is to do the following:

1. Copy all tests that can compile and run from CPython Lib/test to RustPython Lib/test.
2. Mark all the failed test with `@unittest.skip("TODO: fix RustPython")`.
3. Add a job for both linux and windows CI that will run the tests.

## Drawbacks, Rationale, and Alternatives

- There are many tests in CPython so we might want to consider choosing only a subsets of tests to work on at first.
- We should see how slow it would take to run all the tests. We might want to run the tests in parallel.

## Unresolved Questions

- Should we mark the tests that currently don't pass with `@unittest.expectedFailure` so we will know when something is fixed?
- What should happen to the snippets tests? Should we keep a duplicate set of tests?
- Would we add tests to CPython set of tests?

coolreader18:
I think this is a good idea; we should probably start with copying over only the test files we want and the required modules, so that we don't end up with a diff of 1000+ added files like in #1587. We also have to have a way to filter out the test files from the `py_compile_bytecode` macro, so that we don't include unwanted modules that are useless at Python runtime.

> Should we mark the tests that currently don't pass with @unittest.expectedFailure so we will know when something is fixed?

I think so; it would certainly be more of an alert to change it than the `# XXX RustPython TODO` comments we put in the stdlib now.

> What should happen to the snippets tests? Should we keep a duplicate set of tests?

I think the snippet tests will be good for a while, because to even run CPython's tests at all we have to have a **lot** of Python features working properly, so I think the snippet tests are a less fragile way of testing.

PR #1633, opened 2019-12-15
Lib/test/test_copy.py from CPython v3.8.1

youknowone:


palaviv:
@youknowone I have a put the test_bool.py unittests from CPython in `Lib/test`. I am working on running them in the CI. The plan on how to add them is specified in #1596.

youknowone:
commented out tests make crash instead of rust error

palaviv:
Are they failing when compiling or when you run the test? If that is when you run the test I would prefer adding `@unittest.skip("TODO: RUSTPYTHON")` as it would be shown clearly when running the tests.

PR #1636, opened 2019-12-17
Itertools.permutations implementation & combinations fixes

dralley:


vm/src/stdlib/itertools.rs
```diff
@@ -908,48 +908,179 @@ impl PyItertoolsCombinations {
...
 
-        Ok(res.into_ref(vm).into_object())
+        Ok(vm.ctx.new_tuple(
+            result
+                .as_ref()
+                .unwrap()
+                .iter()
+                .map(|&i| self.pool[i].clone())
+                .collect(),
+        ))
```
dralley: CPython tries to be more allocation-efficient by re-using the last-returned tuple (modifying it in-place) if the refcount == 1.  I'm not sure that it would be so straightforwards for us without making custom containers, unless I'm missing something?  Just wanted to point it out.

dralley:
Clippy says this: https://travis-ci.org/RustPython/RustPython/jobs/626275690#L665-L675

But when I make that change, rustc complains:

```error[E0499]: cannot borrow `indices` as mutable more than once at a time
    --> vm/src/stdlib/itertools.rs:1058:58
     |
1058 |                     std::mem::swap(&mut indices[i], &mut indices[n - j]);
     |                     --------------      -------          ^^^^^^^ second mutable borrow occurs here
     |                     |                   |
     |                     |                   first mutable borrow occurs here
     |                     first borrow later used by call

```

Is there a way to work around this, and/or should a clippy bug be filed?

coolreader18:
You can use `slice::swap` to swap the elements: https://doc.rust-lang.org/std/primitive.slice.html#method.swap

vm/src/stdlib/itertools.rs
coolreader18: I think you can use [`Rc::get_mut`](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.get_mut) for this functionality.

coolreader18: Or [`Rc::make_mut`](https://doc.rust-lang.org/std/rc/struct.Rc.html#method.make_mut) might be useful here as well?

coolreader18: Oh, or you might want to add those sorta methods to `PyRef`, as you'd have to use the `Rc` functions on a `PyObjectRef`.

coolreader18: Oh, or you might want to add those sorta methods to `PyRef`, as you'd have to use the `Rc` functions on a `PyObjectRef`.

dralley:
Awesome, that works.

vm/src/stdlib/itertools.rs
```diff
@@ -908,48 +908,177 @@ impl PyItertoolsCombinations {
...
+        vm.class("itertools", "permutations")
+    }
+}
+
+#[pyimpl]
+impl PyItertoolsPermutations {
+    #[pyslot(new)]
+    fn tp_new(
+        cls: PyClassRef,
+        iterable: PyObjectRef,
```
coolreader18: You could use `OptionalOption` here

vm/src/stdlib/itertools.rs
```diff
@@ -908,48 +908,177 @@ impl PyItertoolsCombinations {
...
+        r: OptionalArg<PyObjectRef>,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyRef<Self>> {
+        let iter = get_iter(vm, &iterable)?;
+        let pool = get_all(vm, &iter)?;
+
+        let n = pool.len();
+        // If r is not provided, r == n. If provided, r must be a positive integer, or None.
+        // If None, it behaves the same as if it was not provided.
```
coolreader18: And then `flat_option` here

vm/src/stdlib/itertools.rs
```diff
@@ -908,48 +908,177 @@ impl PyItertoolsCombinations {
...
-        if !sentinel {
+        .into_ref_with_type(vm, cls)
+    }
+
+    #[pymethod(name = "__iter__")]
+    fn iter(zelf: PyRef<Self>, _vm: &VirtualMachine) -> PyRef<Self> {
+        zelf
+    }
+
+    #[pymethod(name = "__next__")]
```
coolreader18: You could have the return type be `PyTuple`, and then you don't have to do `into_ref`

vm/src/stdlib/itertools.rs
```diff
@@ -908,48 +908,177 @@ impl PyItertoolsCombinations {
...
 
-        // Increment the current index which we know is not at its
-        // maximum.  Then move back to the right setting each index
-        // to its lowest possible value (one higher than the index
-        // to its left -- this maintains the sort order invariant).
-        indices[idx] += 1;
-        for j in idx + 1..r {
-            indices[j] = indices[j - 1] + 1;
+        let mut result = self.result.borrow_mut();
+
```
coolreader18: To avoid having to do `as_mut().unwrap()` later, you can do `if let Some(ref mut result) = result { ... } else { ... }` (or match if you want to keep the order it's in)

vm/src/stdlib/itertools.rs
```diff
@@ -908,48 +908,177 @@ impl PyItertoolsCombinations {
...
+        if result.is_none() {
+            // On the first pass, initialize result tuple using the indices
```
coolreader18: And this can just be `*result = ...`

coolreader18:
Sorry for not reviewing earlier; the implementation looks good.

dralley:
@coolreader18 Updated.  Re: earlier comments about reference counting, I struggled to make it work, and wasn't really able to, at least not right now.  I'll hold those changes in another branch until after I finish ```combinations_with_replacement()```

vm/src/stdlib/itertools.rs
dralley: Fixed

vm/src/stdlib/itertools.rs
dralley: Fixed

vm/src/stdlib/itertools.rs
dralley: Fixed

PR #1638, opened 2019-12-19
Add named expressions / assignment expressions

dralley:
Aka "walrus operator".
https://www.python.org/dev/peps/pep-0572/

### DIFF
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -1565,6 +1565,17 @@ impl<O: OutputStream> Compiler<O> {
         Ok(())
     }
 
+    fn compile_named_expression(
+            &mut self,
+            target: &ast::Expression,
+            value: &ast::Expression,
+        ) -> Result<(), CompileError> {
+
+            // evaluate value
+            // store into target
+        Ok(())
+    }
+
     fn compile_expression(&mut self, expression: &ast::Expression) -> Result<(), CompileError> {
         trace!("Compiling {:?}", expression);
         self.set_source_location(&expression.location);
@@ -1577,6 +1588,7 @@ impl<O: OutputStream> Compiler<O> {
                 keywords,
             } => self.compile_call(function, args, keywords)?,
             BoolOp { op, values } => self.compile_bool_op(op, values)?,
+            NamedExpression { target, value} => self.compile_named_expression(target, value)?,
             Binop { a, op, b } => {
                 self.compile_expression(a)?;
                 self.compile_expression(b)?;
@@ -1912,6 +1924,7 @@ impl<O: OutputStream> Compiler<O> {
 
         Ok(has_stars)
     }
+
 
     fn compile_comprehension(
         &mut self,
--- a/parser/src/ast.rs
+++ b/parser/src/ast.rs
@@ -186,6 +186,12 @@ pub enum ExpressionType {
         values: Vec<Expression>,
     },
 
+    /// A named expression aka. assignment expression aka. "Walrus operator"
+    NamedExpression {
+        target: Box<Expression>,
+        value: Box<Expression>,
+    },
+
     /// A binary operation on two operands.
     Binop {
         a: Box<Expression>,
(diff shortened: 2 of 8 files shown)


compiler/src/compile.rs
```diff
@@ -1567,6 +1567,17 @@ impl<O: OutputStream> Compiler<O> {
...
     }
 
+    fn compile_named_expression(
+            &mut self,
+            target: &ast::Expression,
+            value: &ast::Expression,
+        ) -> Result<(), CompileError> {
+
+            // evaluate value
+            // store into target
```
dralley: TODO

parser/src/python.lalrpop
```diff
@@ -787,6 +787,14 @@ MulOp: ast::Operator = {
     "@" => ast::Operator::MatMult,
 };
 
+NamedExpression: ast::Expression = {
+    <e1:Expression> <location:@L> ":=" <e2:Expression> => ast::Expression {
+        location,
+        node: ast::ExpressionType::NamedExpression { target: Box::new(e1), value: Box::new(e2) }
+    },
+    NamedExpression,
+};
```
dralley: Probably this is wrong.  I'm getting ```LalrpopError::UnrecognizedToken```->```ParseErrorType::UnrecognizedToken```

```
>>>>> if any(len(longline := line) >= 100 for line in lines):
SyntaxError: 
if any(len(longline := line) >= 100 for line in lines):
                    ↑
                    Got unexpected token ':='
```

parser/src/ast.rs
```diff
@@ -359,7 +366,7 @@ impl Expression {
             | String {
                 value: FormattedValue { .. },
             } => "f-string expression",
-            Identifier { .. } => "named expression",
```
dralley: This was a typo, right?  Or maybe it predates "named expressions" becoming an actual thing in Python?

parser/src/ast.rs
```diff
@@ -186,6 +186,12 @@ pub enum ExpressionType {
         values: Vec<Expression>,
     },
 
+    /// A named expression aka. assignment expression aka. "Walrus operator"
```
dralley: The fact that the PEP and the implementation use different terminology is somewhat irritating.

coolreader18:
I haven't reviewed this yet, but you can also refer to #968 for the walrus lexing implementation.

dralley:
Ah, I didn't see that PR, thank you.  It looks like it *only* added it to the lexer though, and my PR does the same thing in that respect (apart from slightly different token name).  It's the parsing and the compiling that I'm hung up on (first time dong anything like this).

Although re: the discussion on that PR, I can shelve this for now if the project doesn't want to add 3.8 features just yet.

parser/src/python.lalrpop
coolreader18: It looks like you might want to use `TestOrStarExpr`, as that's what the assignment statement does: https://github.com/RustPython/RustPython/blob/ed37c61df7eed1ada6e1a063720a20a0eb6ec58f/parser/src/python.lalrpop#L88

dralley: It didn't work unfortunately.  Might still be doing something wrong, idk.  I'll take another shot after New Years and do some deeper digging.

windelbouwman:
Since I opened that other pull request, I would like to see the walrus coming :).

parser/src/python.lalrpop
```diff
@@ -787,6 +787,14 @@ MulOp: ast::Operator = {
     "@" => ast::Operator::MatMult,
 };
 
+NamedExpression: ast::Expression = {
+    <e1:Expression> <location:@L> ":=" <e2:Expression> => ast::Expression {
```
windelbouwman: Please keep in mind here to check with this page: https://docs.python.org/3/reference/grammar.html?highlight=grammar

That's the reference I used when making the rest of this file. Instead of `<e1:Expression>` you might want to go for `<e1:Test>`.

parser/src/python.lalrpop
windelbouwman: Please check with the grammar on this page: https://docs.python.org/3/reference/grammar.html?highlight=grammar that contains useful tips on what to modify.

dralley:
Closed by https://github.com/RustPython/RustPython/pull/1934

PR #1639, opened 2019-12-20
Add str.format_map method

doyshinda:
Contributes to #190

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -651,6 +651,32 @@ impl PyString {
         }
     }
 
+    /// S.format_map(mapping) -> str
+    ///
+    /// Return a formatted version of S, using substitutions from mapping.
+    /// The substitutions are identified by braces ('{' and '}').
+    #[pymethod]
+    fn format_map(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
+        if args.args.len() != 2 {
+            return Err(vm.new_type_error(format!(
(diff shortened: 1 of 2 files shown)


vm/src/obj/objstr.rs
```diff
@@ -1590,6 +1616,36 @@ fn perform_format(
...
+    dict: &PyObjectRef,
+) -> PyResult {
+    let mut final_string = String::new();
+    for part in &format_string.format_parts {
+        let result_string: String = match part {
+            FormatPart::AutoSpec(_) => {
+                return Err(
+                    vm.new_value_error("Format string contains positional fields".to_string())
+                );
+            }
```
coolreader18: You can merge this match arm and the previous one by using the `|` character: `FormatPart::AutoSpec(_) | FormatPart::IndexSpec (_, _) => { ... }`

tests/snippets/builtin_format_map.py
```diff
@@ -0,0 +1,64 @@
+from testutils import AssertRaises
+
+
+def test_format_map():
+    # The following tests were performed in Python 3.7.5:
```
coolreader18: Could you move these tests to `tests/snippets/strings.py`? And you don't need the separate test function; our tests right now are a lot less structured than most Python tests 😁

vm/src/obj/objstr.rs
doyshinda: :+1:

coolreader18:
Thanks again for contributing!

PR #1640, opened 2019-12-20
Fix clippy warnings

youknowone:
Not all, but most of them

### DIFF
--- a/parser/src/lexer.rs
+++ b/parser/src/lexer.rs
@@ -33,26 +33,28 @@ impl IndentationLevel {
         // We only know for sure that we're smaller or bigger if tabs
         // and spaces both differ in the same direction. Otherwise we're
         // dependent on the size of tabs.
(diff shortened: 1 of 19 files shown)


parser/src/lexer.rs
```diff
@@ -33,26 +33,28 @@ impl IndentationLevel {
...
             }
-        } else if self.tabs > other.tabs {
-            if self.spaces >= other.spaces {
-                Ok(Ordering::Greater)
-            } else {
-                Err(LexicalError {
-                    location,
-                    error: LexicalErrorType::TabError,
-                })
+            std::cmp::Ordering::Greater => {
```
coolreader18: It looks like `Ordering` is already in scope here; you don't have to fully qualify this

youknowone: fixed it, thanks!

PR #1641, opened 2019-12-20
Refactor float, complex cmp/op , str, None

youknowone:
- float, complex same as #1641 
- str uses PyStringRef rather than PyObjectRef
- impl PyNoneRef -> impl PyNone

### DIFF
--- a/vm/src/obj/objcomplex.rs
+++ b/vm/src/obj/objcomplex.rs
@@ -3,7 +3,7 @@ use num_traits::Zero;
 use std::num::Wrapping;
 
 use super::objfloat::{self, IntoPyFloat};
-use super::objtype::{self, PyClassRef};
+use super::objtype::PyClassRef;
 use crate::function::OptionalArg;
 use crate::pyhash;
 use crate::pyobject::{
@@ -43,18 +43,15 @@ pub fn init(context: &PyContext) {
     PyComplex::extend_class(context, &context.types.complex_type);
 }
 
-pub fn get_value(obj: &PyObjectRef) -> Complex64 {
-    obj.payload::<PyComplex>().unwrap().value
-}
-
 fn try_complex(value: &PyObjectRef, vm: &VirtualMachine) -> PyResult<Option<Complex64>> {
-    Ok(if objtype::isinstance(&value, &vm.ctx.complex_type()) {
-        Some(get_value(&value))
+    let r = if let Some(complex) = value.payload_if_subclass::<PyComplex>(vm) {
+        Some(complex.value)
     } else if let Some(float) = objfloat::try_float(value, vm)? {
         Some(Complex64::new(float, 0.0))
     } else {
         None
-    })
+    };
+    Ok(r)
 }
 
 #[pyimpl]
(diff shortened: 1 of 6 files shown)


vm/src/obj/objcomplex.rs
```diff
@@ -75,33 +72,35 @@ impl PyComplex {
...
+        self.op(other, |a, b| (a - b).into_pyobject(vm), vm)
     }
 
     #[pymethod(name = "__rsub__")]
     fn rsub(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        try_complex(&other, vm)?.map_or_else(
-            || Ok(vm.ctx.not_implemented()),
-            |other| (other - self.value).into_pyobject(vm),
-        )
+        self.op(other, |a, b| (b - a).into_pyobject(vm), vm)
```
windelbouwman: You could move the part `.into_pyobject(vm)` into the op function, this cleans the code a bit.

vm/src/obj/objcomplex.rs
```diff
@@ -75,33 +72,35 @@ impl PyComplex {
         re.hypot(im)
     }
 
-    #[pymethod(name = "__add__")]
-    fn add(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+    #[inline]
+    fn op<F>(&self, other: PyObjectRef, op: F, vm: &VirtualMachine) -> PyResult
+    where
+        F: Fn(Complex64, Complex64) -> PyResult,
```
windelbouwman: Change this function signature into `Fn(Complex64, Complex64) -> Complex64` to reduce boiler plate code elsewhere.

vm/src/obj/objcomplex.rs
```diff
@@ -75,33 +72,35 @@ impl PyComplex {
         re.hypot(im)
     }
 
-    #[pymethod(name = "__add__")]
-    fn add(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+    #[inline]
+    fn op<F>(&self, other: PyObjectRef, op: F, vm: &VirtualMachine) -> PyResult
```
windelbouwman: Nice trick with this `op` function!

windelbouwman (approved):
I had one minor improvement, but for the rest this looks like a massive good cleanup!

vm/src/obj/objfloat.rs
```diff
@@ -179,132 +187,101 @@ impl PyFloat {
...
 
     #[pymethod(name = "__abs__")]
     fn abs(&self, _vm: &VirtualMachine) -> f64 {
         self.value.abs()
     }
 
-    #[pymethod(name = "__add__")]
-    fn add(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+    #[inline]
+    fn op<F>(&self, other: PyObjectRef, float_op: F, vm: &VirtualMachine) -> PyResult
```
windelbouwman: Instead of a single `op` function which works always, it might make sense to create several helper functions for specialized purposes, such as `binary_operator` for +,-,/ and *. The `divmod` operation requires the most generic form, `generic_binary_operation`.

youknowone: I made `simple_op` which returns `PyResult<f64>` and `tuple_op` which returns `PyResult<(f64, f64)>`. Please check if you like this change.

PR #1642, opened 2019-12-21
PyArithmaticValue to keep result type when it can be a special type like NotImplemented

youknowone:
For now, we take advantage of `PyResult<T>` when `T != PyObjectRef` but this is not always working because sometimes a method can return a `NotImplemented` value. When the actual result type is hidden into `PyObjectRef`, unwrapping the payload is painful and not safe.

By adding another option-like enum, it can be transform into a PyObjectRef at the last stage. Then we can reuse the value in other operation. Please check the second commit (https://github.com/RustPython/RustPython/commit/4b5479040e8d816faafef10b75aa6fdfc6b58508#diff-509d6ec78ee702db3d9c46957c6ecbe0R207-R209) about the example.

This is a Proof-of-concept PR. Let me know if there are any bad idea or any improvement suggestion. I don't think the names and the PyInt adoption is the best now. (Unlike PyFloat example, PyInt part doesn't give any value but just looking more verbose.)


windelbouwman:
Thanks for making this proof of concept!

A few of my thoughts about this change:
- The name `PyReturn` seems a bit odd to me. Something like `PyValue` would be better to me.
- How are `NotImplemented` values returned from python code handled? Are those transformed into the specific enum variant?

All in all I am not sure this change is a good idea. I see the point that we must check for the `NotImplemented` value, since this is an `Ok` value, not an error.

Maybe another idea is to create a new `Err` variant for returning either an exception, or a `NotImplemented` value?

```rust
type PyArithmaticResult<T> = Result<T, ArithmaticError>;

enum ArithmaticError {
    Exception(PyObjectRef),
    NotImplemented,
};

fn ne(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyArithmaticResult<bool> {
}
```

This has the benefit we can re-use the `.map` function of the `Result` type, so we do not require to re-implement this.

Btw, I see the benefits of having functions return `PyArithmaticResult<bool>` instead of `PyResult` without a type parameter, I think this is a good idea!

windelbouwman:
Btw, thanks for the detailed explanation of the goal of this change! Nicely done!

youknowone:
If you prefer to define `PyArithmaticResult` as `Result` type rather than a value, this declaration also will work.

```rust
enum PyArithmaticResult<T> {
    Ok(Option<T>),
    Err(PyObjectRef),
}
```

Arithmatic operations never(or rarely) return a `None` value, so I think `Ok(None)` can represent `NotImplemented`

windelbouwman:
Hmm, actually I'm in doubt about this, as you already mentioned, it might rarely occur that the result is indeed None, and then we are faced with an issue, since `None` rust values are magically converted into python `None` values, this might become tricky. Also, I think the best option is to create a new `Result` type, with a new error type, since we can then re-use the `map` function of the built-in `Result` type in rust.

This actually boils down to my initial proposal of the `PyArithmaticResult` type.

coolreader18:
I like the idea of an option-like enum that we use as the `Ok` type for `PyResult`. That way, we can still use `?` to propagate errors (although we could also do that for `ArithmaticError` if we implement `From<PyBaseExceptionRef>`), and we don't have to implement `Try` for a type like `PyReturn` ourselves. It would also make it very easy to work with thanks to the monadish interface of `map`, `map_or`, `unwrap_or_else`, etc that `Option`-like types usually have.

youknowone:
I renamed it to `PyArithmaticValue` and added a special common form `PyComparisonValue`.

If the `Result` approach only take advantage of `map` function, please check the revised code with macro to generate them. Then option-like and `Result` will have same start line. (When they are technically same, I also prefer vote for option-like one as it is an `Ok` value)

About python generated `NotImplemented`, I don't think we ever need to transform any `PyObject` into `PyArithmaticValue`. So it will not happen. In that case, we will have `PyObject` as same as now.

I think this is especially useful for `__ne__` functions not to have duplicated implementations in every class - though i don't see more usefulness that much yet.

windelbouwman (approved):
Looks good to me!

coolreader18 (approved):

PR #1644, opened 2019-12-24
Use PyBaseExceptionRef in places where we expect exceptions

coolreader18:
e.g. in `PyResult`, the exception stack, and any function that does an operation on an exception. I also noticed a bug where if the `from` argument to raise was missing, it would still overwrite the `__cause__` on the exception as if it had been raised `from None`.

### DIFF
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -3,10 +3,10 @@ use crate::obj::objnone::PyNone;
 use crate::obj::objstr::{PyString, PyStringRef};
 use crate::obj::objtraceback::PyTracebackRef;
 use crate::obj::objtuple::{PyTuple, PyTupleRef};
-use crate::obj::objtype::{self, PyClassRef};
+use crate::obj::objtype::{self, PyClass, PyClassRef};
 use crate::pyobject::{
-    Either, PyClassImpl, PyContext, PyIterable, PyObjectRef, PyRef, PyResult, PyValue,
-    TryFromObject, TypeProtocol,
+    PyClassImpl, PyContext, PyIterable, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject,
+    TypeProtocol,
 };
 use crate::types::create_type;
 use crate::vm::VirtualMachine;
@@ -299,6 +299,70 @@ fn exception_args_as_string(
     }
 }
 
+#[derive(Clone)]
+pub enum ExceptionCtor {
+    Class(PyClassRef),
+    Instance(PyBaseExceptionRef),
+}
+
+impl TryFromObject for ExceptionCtor {
+    fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
+        obj.downcast::<PyClass>()
+            .and_then(|cls| {
+                if objtype::issubclass(&cls, &vm.ctx.exceptions.base_exception_type) {
+                    Ok(Self::Class(cls))
+                } else {
+                    Err(cls.into_object())
+                }
+            })
+            .or_else(|obj| obj.downcast::<PyBaseException>().map(Self::Instance))
+            .map_err(|obj| {
+                vm.new_type_error(format!(
+                    "exceptions must be classes or instances deriving from BaseException, not {}",
+                    obj.class().name
+                ))
+            })
+    }
+}
+
+impl ExceptionCtor {
+    pub fn instantiate(self, vm: &VirtualMachine) -> PyResult<PyBaseExceptionRef> {
+        match self {
+            Self::Class(cls) => vm.new_empty_exception(cls),
+            Self::Instance(exc) => Ok(exc),
+        }
+    }
+
+    pub fn instantiate_value(
+        self,
+        value: PyObjectRef,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyBaseExceptionRef> {
+        let exc_inst = value.clone().downcast::<PyBaseException>().ok();
+        match (self, exc_inst) {
+            // both are instances; which would we choose?
+            (Self::Instance(_exc_a), Some(_exc_b)) => {
+                Err(vm
+                    .new_type_error("instance exception may not have a separate value".to_string()))
+            }
+            // if the "type" is an instance and the value isn't, use the "type"
+            (Self::Instance(exc), None) => Ok(exc),
+            // if the value is an instance of the type, use the instance value
+            (Self::Class(cls), Some(exc)) if objtype::isinstance(&exc, &cls) => Ok(exc),
+            // otherwise; construct an exception of the type using the value as args
+            (Self::Class(cls), _) => {
+                let args = match_class!(match value {
+                    PyNone => vec![],
+                    tup @ PyTuple => tup.elements.clone(),
+                    exc @ PyBaseException => exc.args().elements.clone(),
+                    obj => vec![obj],
+                });
+                vm.new_exception_obj(cls, args)
+            }
+        }
+    }
+}
+
 /// Similar to PyErr_NormalizeException in CPython
 pub fn normalize(
     exc_type: PyObjectRef,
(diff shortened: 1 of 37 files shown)


vm/src/exceptions.rs
```diff
@@ -84,29 +82,29 @@ impl PyBaseException {
     }
 
     #[pyproperty(name = "__traceback__", setter)]
-    fn set_traceback(&self, traceback: Option<PyTracebackRef>, vm: &VirtualMachine) -> PyResult {
+    fn setter_traceback(&self, traceback: Option<PyTracebackRef>, vm: &VirtualMachine) -> PyResult {
```
youknowone: it looks a little bit awkward that getter has `get_` prefixed name but setter has `setter_` prefixed names

vm/src/exceptions.rs
```diff
@@ -131,51 +129,91 @@ impl PyBaseException {
...
-            print_exception(vm, &cause);
+    let cause = exc.cause();
+    if let Some(cause) = &cause {
+        write_exception(output, vm, cause)?;
+        writeln!(
+            output,
+            "\nThe above exception was the direct cause of the following exception:\n"
+        )?;
+    }
+    if cause.is_none() {
```
windelbouwman: Instead of a new if test, you can use `else` here.

vm/src/frame.rs
```diff
@@ -176,32 +177,16 @@ impl Frame {
...
 
                     let new_traceback = PyTraceback::new(
                         next,
                         self.clone().into_ref(vm),
                         self.lasti.get(),
                         lineno.row(),
                     );
-                    vm.set_attr(&exception, "__traceback__", new_traceback.into_ref(vm))
-                        .unwrap();
+                    exception.set_traceback(Some(new_traceback.into_ref(vm)));
```
windelbouwman: Does this also work when a user defined exception implements a custom setter function for the `__traceback__` member?

vm/src/frame.rs
```diff
@@ -464,8 +445,8 @@ impl Frame {
                 let args = if let Some(exc) = exc {
                     let exc_type = exc.class().into_object();
                     let exc_val = exc.clone();
-                    let exc_tb = vm.ctx.none(); // TODO: retrieve traceback?
```
windelbouwman: Nice, another TODO removed :).

vm/src/obj/objtype.rs
```diff
@@ -288,7 +288,7 @@ fn _mro(cls: &PyClassRef) -> Vec<PyClassRef> {
 /// Determines if `obj` actually an instance of `cls`, this doesn't call __instancecheck__, so only
 /// use this if `cls` is known to have not overridden the base __instancecheck__ magic method.
 #[cfg_attr(feature = "flame-it", flame("objtype"))]
-pub fn isinstance(obj: &PyObjectRef, cls: &PyClassRef) -> bool {
+pub fn isinstance(obj: &impl TypeProtocol, cls: &PyClassRef) -> bool {
```
windelbouwman: Another option to use here might be a template like this:

```rust
pub fn isinstance<T>(obj: &T, cls: &PyClassRef) -> bool where T: TypeProtocol {
    issubclass(&obj.class(), &cls)
}
```

Did you consider both options? I think the templated version will result in slightly more code, but also faster code.

vm/src/pyobject.rs
```diff
@@ -64,7 +64,7 @@ pub type PyObjectRef = Rc<PyObject<dyn PyObjectPayload>>;
 /// Use this type for functions which return a python object or an exception.
 /// Both the python object and the python exception are `PyObjectRef` types
 /// since exceptions are also python objects.
-pub type PyResult<T = PyObjectRef> = Result<T, PyObjectRef>; // A valid value, or an exception
+pub type PyResult<T = PyObjectRef> = Result<T, PyBaseExceptionRef>; // A valid value, or an exception
```
windelbouwman: Nice change! I think it makes sense that exceptions cannot be just any object, but should be derived from `BaseException` in all cases.

vm/src/stdlib/os.rs
```diff
@@ -1055,18 +1056,18 @@ fn os_getegid(vm: &VirtualMachine) -> PyObjectRef {
 }
 
 #[cfg(unix)]
-fn os_getpgid(pid: u32, vm: &VirtualMachine) -> PyObjectRef {
+fn os_getpgid(pid: u32, vm: &VirtualMachine) -> PyResult {
     match unistd::getpgid(Some(Pid::from_raw(pid as i32))) {
-        Ok(pgid) => vm.new_int(pgid.as_raw()),
-        Err(err) => convert_nix_error(vm, err),
+        Ok(pgid) => Ok(vm.new_int(pgid.as_raw())),
+        Err(err) => Err(convert_nix_error(vm, err)),
```
windelbouwman: Nicely catched bug here!

vm/src/vm.rs
```diff
@@ -328,39 +329,99 @@ impl VirtualMachine {
         PyObject::new(PyModule {}, self.ctx.types.module_type.clone(), Some(dict))
     }
 
+    /// Similar to PyErr_NormalizeException in CPython
+    pub fn normalize_exception(
```
windelbouwman: Does it make sense to move this function into the file `exceptions.rs`?

vm/src/vm.rs
```diff
@@ -328,39 +329,99 @@ impl VirtualMachine {
...
+                vec![]
+            } else {
+                match_class!(match exc_val.clone() {
+                    tup @ PyTuple => tup.elements.clone(),
+                    exc @ PyBaseException => exc.args().elements.clone(),
+                    obj => vec![obj],
+                })
+            }
+        };
+        let exc_type = if let Ok(exc) = PyBaseExceptionRef::try_from_object(self, exc_type.clone())
```
windelbouwman: I would not create this `exc_type` variable, but do the condition checking in the code where the exc_type is used below. In fact it might make sense to create another helper function `figure_out_exception` which returns a `PyResult`, so we can do:

```rust
let exc = figure_out_exception(exc_type, exc_value)?;  // figure out a better name for this function ;)
```

vm/src/vm.rs
```diff
@@ -328,39 +329,99 @@ impl VirtualMachine {
...
+                }
+            }
+        } else {
+            match exc_type {
+                Either::A(exc) => exc,
+                Either::B(cls) => self.new_exception_obj(cls, args())?,
+            }
+        };
+        if let Some(tb) = TryFromObject::try_from_object(self, exc_tb)? {
+            exc.set_traceback(Some(tb));
```
windelbouwman: I like this API of having a `set_traceback` function. I'm only worried that it might not work out properly for custom derived user defined exception classes from `Exception`.

vm/src/exceptions.rs
coolreader18: Yeah I agree, but it was conflicting with the public setter methods so I just renamed them to something that seemed appropriate. I might go with `__traceback__/set____traceback__`, but that might be too many underscores : smiley:

vm/src/obj/objtype.rs
coolreader18: It is a template, I used `impl Trait` and not `dyn Trait`: https://doc.rust-lang.org/edition-guide/rust-2018/trait-system/impl-trait-for-returning-complex-types-with-ease.html

vm/src/frame.rs
coolreader18: It doesn't need to; CPython doesn't heed custom `__traceback__` setters either.

vm/src/exceptions.rs
youknowone: How about adapting #1647 for these methods? This will not fit for everytime but fortunately `set_traceback` and `setter_traceback` can share the implementation in this case.

coolreader18: That sounds good, it would be nice if we could omit the `&VirtualMachine` parameter too as it's not really necessary, but I'm not sure if that's doable.

vm/src/obj/objtype.rs
coolreader18: I do agree that it's more clear what's happening using explicit generics; `&impl Trait` is definitely confusing.

vm/src/exceptions.rs
youknowone: would you check #1649 for this?

vm/src/vm.rs
coolreader18: But that function *is* basically `normalize_exception`. I've reworked the code; I think it's a lot cleaner and easier to follow now.

coolreader18:
@windelbouwman would mind reviewing the latest 4 commits? I think I addressed your comments on `normalize`.

PR #1645, opened 2019-12-25
Add itertools.cycle() and itertools.chain.from_iterable()

dralley:


vm/src/stdlib/itertools.rs
```diff
@@ -177,6 +177,73 @@ impl PyItertoolsCount {
...
+            index: Cell::new(0),
+            first_pass: Cell::new(false),
+        }
+        .into_ref_with_type(vm, cls)
+    }
+
+    #[pymethod(name = "__next__")]
+    fn next(&self, vm: &VirtualMachine) -> PyResult {
+        let item = if let Some(item) = get_next_object(vm, &self.iter.borrow())? {
+            if self.first_pass.get() {
```
dralley: So I copied the CPython implementation verbatim basically, but I'm confused as to why this variable exists.

https://github.com/python/cpython/blob/master/Modules/itertoolsmodule.c#L1015-L1043

If you look at the code, it appears it is read from but never written to, except for ```__setstate__```.  It looks like a long long time ago it was actually used in the actual algorithm, but not anymore.

https://github.com/python/cpython/blob/a86f2c06fd08e8aebf3641c1ad2d234425b90c55/Modules/itertoolsmodule.c#L739-L768

Maybe it's something that has to be kept around for compatibility purposes with pickled code?  Or maybe it is completely vestigial?  I will bring this up in the python development IRC to understand what is going on here.

dralley:
I think the only thing left is ```combinations_with_replacement``` re: https://github.com/RustPython/RustPython/issues/1361

vm/src/stdlib/itertools.rs
coolreader18: That's really weird; I think it's fine for now but it'll definitely be interesting to see what the CPython team thinks.

coolreader18:
Nice work!

PR #1647, opened 2019-12-26
Allow empty return type for property setter

youknowone:


vm/src/obj/objproperty.rs
```diff
@@ -252,6 +252,11 @@ pub struct PropertyBuilder<'a> {
     setter: Option<PyObjectRef>,
 }
 
+pub trait PropertySetterResult {}
+
+impl PropertySetterResult for PyResult {}
```
coolreader18: Maybe this should just be `PyResult<()>`, as setters don't ever really return anything.

youknowone: That sounds better. I added another commit for that.

PR #1648, opened 2019-12-27
Support eval and exec mode in wasm

### LINKED ISSUE #1612: [WASM] pyEval does not return JsValue
# Description
pyEval should return the `JsValue`, as the run function of wasmvm proposed, 
```fn run(&self, source: &str, mode: compile::Mode) -> Result<JsValue, JsValue>```, 
but there is something missing such that the return value is always `undefined`.

## Expected Result
```
>> let a = rp.pyEval("1 + 1")
← undefined
>> a
← 2
```

## Actual Result
```
>> let a = rp.pyEval("1 + 1")
← undefined
>> a
← undefined
```

### PR DESCRIPTION
**Pull request by yanganto**
- Expose `pyEval`, `pyExec`, `pyExecSingle` in wasm module

This PR also fix #1612

### DIFF
--- a/wasm/lib/src/lib.rs
+++ b/wasm/lib/src/lib.rs
@@ -8,11 +8,13 @@ extern crate futures;
 extern crate js_sys;
 #[macro_use]
 extern crate rustpython_vm;
+extern crate rustpython_compiler;
 extern crate wasm_bindgen;
 extern crate wasm_bindgen_futures;
 extern crate web_sys;
 
 use js_sys::{Object, Reflect, TypeError};
+use rustpython_compiler::compile::Mode;
 use std::panic;
 use wasm_bindgen::prelude::*;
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by yanganto** [comment]
The behavior in the console list as following
```
>> rp.pyEval("1+1")
← 2
>> rp.pyExec("1+1")
← undefined
>> rp.pyExecSingle("1+1\n2+2")
2
4
← 4
```

File: wasm/tests/test_exec_mode.py [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+    assert driver.execute_script("return window.rp.pyEval('1+1')") == 2
+
+def test_exec_mode(driver):
+    assert driver.execute_script("return window.rp.pyExec('1+1')") is None
+
+def test_exec_single_mode(driver):
+    assert driver.execute_script("return window.rp.pyExecSingle('1+1')") == 2
+    assert driver.execute_script(
+        """
+        var output = "";
```
coolreader18: You might want to make `output` an array, just to make clearer that the outputs are separate.

wasm/lib/src/lib.rs
```diff
@@ -47,14 +49,39 @@ pub fn setup_console_error() {
...
+        } else {
+            return Err(TypeError::new("vars must be an object").into());
+        }
+    };
+
+    vm.set_stdout(Reflect::get(&options, &"stdout".into())?)?;
+
+    if let Some(js_vars) = js_vars {
+        vm.add_to_scope("js_vars".into(), js_vars.into())?;
+    }
```
coolreader18: This match is basically what `VMStore.run` does -- you might want to make that `pub(crate)` and just call it here.

yanganto:
Thanks for the review.

PR #1649, opened 2019-12-28
Allow to skip VM param for IntoPyNativeFunc functions

youknowone:
Suggestion from #1644

Not sure this is what we want - because I thought VM param is designated constraint not to confuse plain functions with python functions until now.

### DIFF
--- a/vm/src/function.rs
+++ b/vm/src/function.rs
@@ -498,11 +498,11 @@ pub type PyNativeFunc = Box<dyn Fn(&VirtualMachine, PyFuncArgs) -> PyResult + 's
 ///
 /// A bare `PyNativeFunc` also implements this trait, allowing the above to be
 /// done manually, for rare situations that don't fit into this model.
-pub trait IntoPyNativeFunc<T, R> {
+pub trait IntoPyNativeFunc<T, R, VM> {
     fn into_func(self) -> PyNativeFunc;
 }
 
-impl<F> IntoPyNativeFunc<PyFuncArgs, PyResult> for F
+impl<F> IntoPyNativeFunc<PyFuncArgs, PyResult, VirtualMachine> for F
 where
     F: Fn(&VirtualMachine, PyFuncArgs) -> PyResult + 'static,
 {
@@ -520,7 +520,7 @@ pub struct RefParam<T>(std::marker::PhantomData<T>);
 // Note that this could be done without a macro - it is simply to avoid repetition.
 macro_rules! into_py_native_func_tuple {
     ($(($n:tt, $T:ident)),*) => {
-        impl<F, $($T,)* R> IntoPyNativeFunc<($(OwnedParam<$T>,)*), R> for F
+        impl<F, $($T,)* R> IntoPyNativeFunc<($(OwnedParam<$T>,)*), R, VirtualMachine> for F
         where
             F: Fn($($T,)* &VirtualMachine) -> R + 'static,
             $($T: FromArgs,)*
@@ -535,7 +535,7 @@ macro_rules! into_py_native_func_tuple {
             }
         }
 
-        impl<F, S, $($T,)* R> IntoPyNativeFunc<(RefParam<S>, $(OwnedParam<$T>,)*), R> for F
+        impl<F, S, $($T,)* R> IntoPyNativeFunc<(RefParam<S>, $(OwnedParam<$T>,)*), R, VirtualMachine> for F
         where
             F: Fn(&S, $($T,)* &VirtualMachine) -> R + 'static,
             S: PyValue,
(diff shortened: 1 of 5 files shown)


coolreader18:
I personally like this change, but I think it'll require more discussion than would be necessary as a prerequisite for #1644. Maybe we can merge this after that?

I think this allows us a more unified interface for types that are exposed to both Python and Rust, like `PyBaseExceptionRef`. It's sorta like wasm-bindgen -- functions can be given the `#[wasm_bindgen]` attribute to expose them to JavaScript, but if the parameter and return types are ergonomic to work with then you can/should expose it to Rust code as well, all with the same function definition.

vm/src/obj/objproperty.rs
```diff
@@ -265,7 +265,7 @@ impl<'a> PropertyBuilder<'a> {
         }
     }
 
-    pub fn add_getter<I, V, F: IntoPyNativeFunc<I, V>>(self, func: F) -> Self {
+    pub fn add_getter<I, V, VM, F: IntoPyNativeFunc<I, V, VM>>(self, func: F) -> Self {
```
windelbouwman: Is the rust convention that template parameter names are a single character? In this case maybe we could use `M` instead of `VM`?

vm/src/function.rs
```diff
@@ -549,6 +549,36 @@ macro_rules! into_py_native_func_tuple {
                 })
             }
         }
+
+        impl<F, $($T,)* R> IntoPyNativeFunc<($(OwnedParam<$T>,)*), R, ()> for F
```
windelbouwman: I'm still struggeling to understand this part of the code. I'm unable to review this properly...

vm/src/obj/objproperty.rs
coolreader18: That's the most common case, but I think rustc's identifier lint rules view generics as type names, e.g. upper camel case.

vm/src/function.rs
```diff
@@ -549,6 +549,36 @@ macro_rules! into_py_native_func_tuple {
                 })
             }
```
coolreader18: Since we're essentially just using the generics on `IntoPyNativeFunc` as a tag to avoid conflicting generic impls, maybe we could unify the generic parameters into one FuncInfo parameter that's a tuple of `(Params, Return, VM)` and whatever else we might need later, to make it easier for functions like `new_rustfunc` that want to accept any sort of `IntoPyNativeFunc`. (Maybe even an [`HList`](https://docs.rs/frunk)? Probably not.)

coolreader18: Ooh, or maybe we could have a wrapper trait that's basically
```rust
impl<P, R, VM, T> IntoPyNativeFunc for T
where T: IntoPyNativeFuncInner
{
    type Param = P;
    type Returns = R;
    type VMPresent = VM;
    fn into_func(self) -> PyNativeFunc {
        IntoPyNativeFuncInner::into_func(self)
    }
}
```

That way we can leave out the generics in most cases, and use `IntoPyNativeFunc<Params=...>` when we do care about some of them.

vm/src/function.rs
youknowone: Would you tell me which part made the confusion? the last `()` part?

vm/src/obj/objproperty.rs
youknowone: Is it `VM` or `Vm` in this case? probably `Vm`? I thought it is ok because clippy didn't complain about this.

vm/src/function.rs
youknowone: To be honest, I didn't understand what you meant from this. will you please attach a commit into this PR?

coolreader18: Nevermind, I thought that basically this would be possible:
```rust
pub trait IntoPyNativeFunc {
    type Params;
    type Result;
    fn into_func(self) -> PyNativeFunc;
}

impl<T, R, VM, F: IntoPyNativeFuncInner<T, R, VM>> IntoPyNativeFunc for F {
    type Params = T;
    type Result = R;
    fn into_func(self) -> PyNativeFunc {
        IntoPyNativeFuncInner::into_func(self)
    }
}

trait IntoPyNativeFuncInner<T, R, VM> {
    fn into_func(self) -> PyNativeFunc;
}
// implementing IntoPyNativeFuncInner for Fn(...) ...

// then that can be used in e.g. new_rustfunc without caring about the generics:
fn new_rustfunc(ctx, f: impl IntoPyNativeFunc) -> PyObjectRef { ... }
```
But that also has the issue of "potentially" multiple impls of `IntoPyNativeFuncInner` for the same type, even though we know that doesn't happen. We might be able to use the [`#[fundamental]` attribute](https://github.com/rust-lang/rfcs/blob/master/text/1023-rebalancing-coherence.md#errors-from-cargo-and-the-fundamental-attribute) once it's stabilized, but I guess it's not possible right now.

youknowone:
Any suggestion about design policy or improvement is appreciated!

youknowone:
@windelbouwman Do you have things to comment? Let me know if you want more explanation about `VM` part or somewhere else.

@palaviv if this PR is acceptable, that sounds good for reusing functions

windelbouwman:
I have no comments on this, I would have to study the code more. The change as such looks okay I would say.

windelbouwman:
Do we already have a page with some internal documentation? To me it would make sense to have some documentation folder about the tricky parts of rustpython.

Potential tricky parts:
- how the stdlib is included with the bootstrap python code
- The ideas behind the derive crate
- How the lalrpop parser works, there are some tricky bits there

PR #1650, opened 2019-12-28
Run unittests

palaviv:
This change include many different changes that allows us to run the CPython test using `libregtest`. After the change you can do:
```
cargo run --release Lib/test/regrtest.py
== RustPython 3.5.0alpha (heads/run-unittests-dirty:cba70133, Dec 28 2019, 10:23:15) [rustc 1.40.0]
== cwd: /tmp/test_python_19746
== CPU count: 4
== encodings: locale=UTF-8, FS=utf-8
Run tests sequentially
0:00:00 [1/1] test_bool
test_bool passed

== Tests result: SUCCESS ==

1 test OK.

Total duration: 293 ms
Tests result: SUCCESS
<SystemExit object at 0x5571b65f8a10>
```

This change is based on the change in #1625.

### DIFF
--- a/Lib/test/libregrtest/__init__.py
+++ b/Lib/test/libregrtest/__init__.py
@@ -1,0 +1,5 @@
+# We import importlib *ASAP* in order to test #15386
+import importlib
+
+from test.libregrtest.cmdline import _parse_args, RESOURCE_NAMES, ALL_RESOURCES
+from test.libregrtest.main import main
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -896,38 +896,38 @@ if sys.platform == 'darwin':
     TESTFN_UNICODE = unicodedata.normalize('NFD', TESTFN_UNICODE)
 TESTFN_ENCODING = sys.getfilesystemencoding()
 
-# TESTFN_UNENCODABLE is a filename (str type) that should *not* be able to be
-# encoded by the filesystem encoding (in strict mode). It can be None if we
-# cannot generate such filename.
-TESTFN_UNENCODABLE = None
-if os.name == 'nt':
-    # skip win32s (0) or Windows 9x/ME (1)
-    if sys.getwindowsversion().platform >= 2:
-        # Different kinds of characters from various languages to minimize the
-        # probability that the whole name is encodable to MBCS (issue #9819)
-        TESTFN_UNENCODABLE = TESTFN + "-\u5171\u0141\u2661\u0363\uDC80"
-        try:
-            TESTFN_UNENCODABLE.encode(TESTFN_ENCODING)
-        except UnicodeEncodeError:
-            pass
-        else:
-            print('WARNING: The filename %r CAN be encoded by the filesystem encoding (%s). '
-                  'Unicode filename tests may not be effective'
-                  % (TESTFN_UNENCODABLE, TESTFN_ENCODING))
-            TESTFN_UNENCODABLE = None
-# Mac OS X denies unencodable filenames (invalid utf-8)
-elif sys.platform != 'darwin':
-    try:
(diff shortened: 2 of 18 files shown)


Lib/test/support/__init__.py
```diff
@@ -1624,21 +1624,23 @@ def gc_collect():
...
-    finally:
-        if have_gc:
-            gc.enable()
+    # have_gc = gc.isenabled()
+    # gc.disable()
+    # try:
+    #     yield
+    # finally:
+    #     if have_gc:
+    #         gc.enable()
```
coolreader18: You should probably still yield here, since it's a context manager. Unless `@contextlib.contextmanager` works with non-generators.

compiler/src/symboltable.rs
```diff
@@ -726,8 +726,15 @@ impl SymbolTableBuilder {
     fn scan_string_group(&mut self, group: &ast::StringGroup) -> SymbolTableResult {
         match group {
             ast::StringGroup::Constant { .. } => {}
-            ast::StringGroup::FormattedValue { value, .. } => {
+            ast::StringGroup::FormattedValue {
+                value,
+                conversion: _,
```
coolreader18: I think this is what's causing the failure, in both this and the other PR. You can just add `..` after the fields you actually use.

vm/src/stdlib/faulthandler.rs
```diff
@@ -0,0 +1,47 @@
...
+        frame.get_lineno().row(),
+        frame.code.obj_name
+    )
+}
+
+fn dump_traceback(_file: OptionalArg<i64>, _all_threads: OptionalArg<bool>, vm: &VirtualMachine) {
+    eprintln!("Stack (most recent call first):");
+
+    Ref::map(vm.frames.borrow(), |frames| {
+        &for frame in frames {
```
coolreader18: Why not just `for frame in vm.frames.borrow() { ... }` ?

vm/src/stdlib/time_module.rs
```diff
@@ -264,6 +264,7 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         "strptime" => ctx.new_rustfunc(time_strptime),
         "sleep" => ctx.new_rustfunc(time_sleep),
         "struct_time" => struct_time_type,
-        "time" => ctx.new_rustfunc(time_time)
+        "time" => ctx.new_rustfunc(time_time),
+        "perf_counter" => ctx.new_rustfunc(time_time),
```
coolreader18: Maybe a TODO here?

compiler/src/symboltable.rs
palaviv: I fixed that at #1625. You can ignore the changes there as I will rebase once it is in.

Lib/test/support/__init__.py
palaviv: Done

vm/src/stdlib/faulthandler.rs
palaviv: Not sure... Fixed

vm/src/stdlib/time_module.rs
palaviv: Done

compiler/src/symboltable.rs
coolreader18: Ah okay, I didn't realize that that was passing now; I merged it.

windelbouwman:
This is sincerely great news!

Lib/test/libregrtest/__init__.py
```diff
@@ -0,0 +1,5 @@
+# We import importlib *ASAP* in order to test #15386
+import importlib
+
```
windelbouwman: Would it make sense to place the `libregrtest` folder outside the `Lib` folder? Or has this discussion already happened?

Issue #1613, opened 2019-11-27
automatic integer to floating point conversion in format strings


lovasoa:
## Feature

Format codes such as 'e', 'f', and 'g' are supported in format strings for integers in CPython, but not in RustPython

### in cpython

```py
>>> f'{5:f}'
'5.000000'
```

### in RustPython

```py
>>>>> f'{5:f}'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Unknown format code 'f' for object of type 'int'
```

## Python Documentation

https://docs.python.org/3/library/string.html#formatspec

doyshinda:
Issue update:

`:f` is now supported for integers:
```Python
Welcome to the magnificent Rust Python 0.1.1 interpreter 😱 🖖
>>>>> f'{5:f}'
'5.000000'
>>>>> f'{5:.2f}'
'5.00'
```

Remaining work is to implement `e` and `g` format codes.

doyshinda:
Issue update:

`:e` is now supported for integers:
```Python
Welcome to the magnificent Rust Python 0.1.1 interpreter 😱 🖖
>>>>> f'{5:e}'
'5.000000e+00'
>>>>> f'{5:.2e}'
'5.00e+00'
```

youknowone:
not sure when its fixed, but now it works

```
>>>>> f'{5:f}'
'5.000000'
```

PR #1651, opened 2019-12-28
Add support for string formatting of floats with ":f" format code

doyshinda:
Also adds automatic conversion of int -> float when the ":f"
format code is specified.


doyshinda:
Contributes to #1613

vm/src/format.rs
```diff
@@ -281,14 +282,22 @@ impl FormatSpec {
         separator: char,
     ) -> String {
         let mut result = String::new();
-        let mut remaining: usize = magnitude_string.len();
-        for c in magnitude_string.chars() {
+
+        // Don't add separators to the floating decimal point of numbers
```
coolreader18: I'm not sure exactly what the functionality of this is supposed to be, but you might want to do `.splitn('.', 2)` so that the splitting sort of has right associativity. Then you don't have to `.collect()` it and you can do `let magnitude_integer_string = parts.next().unwrap();` and `if let Some(part) = parts.next() { ... }`

doyshinda: This is needed to prevent adding an incorrect separator to floats, as this is what it looked like before:
```Python
>>>>> '{:,.2f}'.format(1234567890.1234)
'1,234,567,890,.12'
```
With this change, it's only adding the separator to the integer portion of the number:
```Python
>>>>> '{:,.2f}'.format(1234567890.1234)
'1,234,567,890.12'
```

I'll change it to the way you suggested, as that is cleaner.

coolreader18:
Thanks for contributing!

Issue #1656, opened 2020-01-02
Implement remaining string format codes for floats


doyshinda:
## Feature

Implement the `g`, `G`, `e`, `E`, `n`, and `%` format codes for floats.

- [x] g
- [x] G
- [x] e
- [x] E
- [x] n
- [x] %

## A few examples from CPython:
```Python
>>> '{:g}'.format(10.0)
'10'
>>> '{:.1g}'.format(10.0)
'1e+01'
>>> '{:e}'.format(10.0)
'1.000000e+01'
>>> '{:.1e}'.format(10.0)
'1.0e+01'
>>> '{:n}'.format(999999.1234)
'999999'
>>> '{:n}'.format(1000000.1234)
'1e+06'
>>> '{:.8n}'.format(1000000.1234)
'1000000.1'
>>> '{:.11n}'.format(1000000.1234)
'1000000.1234'
>>> '{:%}'.format(10.0)
'1000.000000%'
```
## RustPython
```Python
>>>>> '{:g}'.format(10.0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Format code 'g' for object of type 'float' not implemented yet
>>>>> '{:G}'.format(10.0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Format code 'G' for object of type 'float' not implemented yet
>>>>> '{:e}'.format(10.0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Format code 'e' for object of type 'float' not implemented yet
>>>>> '{:E}'.format(10.0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Format code 'E' for object of type 'float' not implemented yet
>>>>> '{:n}'.format(10.0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ValueError: Format code 'n' for object of type 'float' not implemented yet
>>>>> '{:%}'.format(10.0)
'10.0'
```

## Python Documentation

See the section after `The available presentation types for floating point and decimal values are` in the docs found [here](https://docs.python.org/3/library/string.html#format-specification-mini-language).

merkrafter:
What is the exact current status of this issue? It looks like e.g. `:g` and `:G` are at least supported (though not yielding the same results as CPython does in every case in the master branch as of writing this).

JazzGlobal:
Im gonna go ahead and wrap up the 'n' case.

[...]

PR #1657, opened 2020-01-03
Add support for the `%` format code for floats.

doyshinda:
Contributes to #1656

### DIFF
--- a/tests/snippets/strings.py
+++ b/tests/snippets/strings.py
@@ -431,3 +431,17 @@ assert f'{1234567890.1234:,.2f}' == '1,234,567,890.12'
 assert f'{1234567890.1234:_.2f}' == '1_234_567_890.12'
 with AssertRaises(ValueError, msg="Unknown format code 'd' for object of type 'float'"):
     f'{5.0:04d}'
+
+# Test % formatting
+assert f'{10.0:%}' == '1000.000000%'
+assert f'{10.0:.2%}' == '1000.00%'
+assert f'{10.0:.8%}' == '1000.00000000%'
+assert f'{-10.0:%}' == '-1000.000000%'
+assert f'{-10.0:.2%}' == '-1000.00%'
+assert f'{-10.0:.8%}' == '-1000.00000000%'
+assert '{:%}'.format(float('nan')) == 'nan%'
+assert '{:.2%}'.format(float('nan')) == 'nan%'
+assert '{:%}'.format(float('inf')) == 'inf%'
+assert '{:.2%}'.format(float('inf')) == 'inf%'
+with AssertRaises(ValueError, msg='Invalid format specifier'):
+    f'{10.0:%3}'
--- a/vm/src/format.rs
+++ b/vm/src/format.rs
@@ -98,6 +98,7 @@ pub enum FormatType {
     GeneralFormatUpper,
     FixedPointLower,
     FixedPointUpper,
+    Percentage,
 }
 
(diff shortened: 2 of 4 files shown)


tests/snippets/strings.py
```diff
@@ -431,3 +431,21 @@ def try_mutate_str():
...
+# Test % formatting
+assert f'{10.0:%}' == '1000.000000%'
+assert f'{10.0:.2%}' == '1000.00%'
+assert f'{10.0:.8%}' == '1000.00000000%'
+assert f'{-10.0:%}' == '-1000.000000%'
+assert f'{-10.0:.2%}' == '-1000.00%'
+assert f'{-10.0:.8%}' == '-1000.00000000%'
+assert '{:%}'.format(float('nan')) == 'nan%'
+assert '{:%}'.format(float('NaN')) == 'nan%'
+assert '{:%}'.format(float('NAN')) == 'nan%'
```
coolreader18: The difference in casing is kinda redundant, as the formatting machinery isn't really involved; it's `float.__new__` that parses the `nan`s.

vm/src/obj/objfloat.rs
```diff
@@ -213,8 +213,11 @@ impl PyFloat {
 
     #[pymethod(name = "__format__")]
     fn format(&self, spec: PyStringRef, vm: &VirtualMachine) -> PyResult<String> {
-        let format_spec = FormatSpec::parse(spec.as_str());
-        match format_spec.format_float(self.value) {
+        let try_format = || {
+            let format_spec = FormatSpec::parse(spec.as_str())?;
+            format_spec.format_float(self.value)
+        };
+        match try_format() {
```
coolreader18: You could do something like
```rust
match FormatSpec::parse(spec.as_str()).and_then(|spec| spec.format_float(self.value)) {
    ...
}
```
to avoid the closure.

vm/src/format.rs
```diff
@@ -245,14 +247,17 @@ fn parse_format_spec(text: &str) -> FormatSpec {
     let (width, after_width) = parse_number(after_zero);
     let (grouping_option, after_grouping_option) = parse_grouping_option(after_width);
     let (precision, after_precision) = parse_precision(after_grouping_option);
-    let (format_type, _) = parse_format_type(after_precision);
+    let (format_type, after_format_type) = parse_format_type(after_precision);
+    if !after_format_type.is_empty() {
+        return Err("Invalid format spec");
+    }
```
coolreader18: Maybe add some tests that ensure that this properly fails with characters left after the format type.

tests/snippets/strings.py
doyshinda: Indeed it is, will remove :+1:

vm/src/format.rs
doyshinda: :+1:

vm/src/obj/objfloat.rs
doyshinda: Sure thing. Out of curiosity, what are the benefits (if any) of doing as shown above vs. using a closure? Or is it just Rust convention?

coolreader18: Since it's not a very complicated operation, it's simpler to use the combinator method than define a closure and immediately call it once. It's similar to `do` vs `>>=` if you've worked with Haskell ([do notation considered harmful](https://wiki.haskell.org/Do_notation_considered_harmful)).

coolreader18:
Thanks for contributing!

PR #1658, opened 2020-01-03
Update unicodedata with ucd_3_2_0

coolreader18:


vm/src/stdlib/unicodedata.rs
```diff
@@ -4,91 +4,171 @@
 
 use crate::function::OptionalArg;
 use crate::obj::objstr::PyStringRef;
-use crate::pyobject::{PyObjectRef, PyResult};
+use crate::obj::objtype::PyClassRef;
+use crate::pyobject::{PyClassImpl, PyObject, PyObjectRef, PyResult, PyValue};
 use crate::vm::VirtualMachine;
 
+use itertools::Itertools;
+
```
youknowone: it seems we usually only have inner/outer separation for `use`

PR #1664, opened 2020-01-07
Add htons, ntohl, and ntohs to socket.rs

metaphorshear:
For #1176.  Just added each of those functions to `vm/src/stdlib/socket.rs`, mostly by cheating off of the `htonl` implementation.

### DIFF
--- a/vm/src/stdlib/socket.rs
+++ b/vm/src/stdlib/socket.rs
@@ -449,6 +449,28 @@ fn socket_htonl(host: u32, vm: &VirtualMachine) -> PyResult {
     Ok(vm.new_int(host.to_be()))
 }
 
+fn socket_htons(host: u16, vm: &VirtualMachine) -> PyResult {
+    Ok(vm.new_int(host.to_be()))
+}
+
+fn socket_ntohl(network: u32, vm: &VirtualMachine) -> PyResult {
+    if cfg!(target_endian = "big") {
+        Ok(vm.new_int(network))
+    }
+    else {
+        Ok(vm.new_int(network.to_le()))
+    }
+}
+
+fn socket_ntohs(network: u16, vm: &VirtualMachine) -> PyResult {
+    if cfg!(target_endian = "big") {
+        Ok(vm.new_int(network))
+    }
+    else {
+        Ok(vm.new_int(network.to_le()))
+    }
+}
+
 #[derive(FromArgs)]
 struct GAIOptions {
     #[pyarg(positional_only)]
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/socket.rs
```diff
@@ -449,6 +449,28 @@ fn socket_htonl(host: u32, vm: &VirtualMachine) -> PyResult {
     Ok(vm.new_int(host.to_be()))
 }
 
+fn socket_htons(host: u16, vm: &VirtualMachine) -> PyResult {
```
youknowone: if the only difference is type, probably generic function will work
```rust
fn socket_hton<U>(host: U, _vm: &VirtualMachine) -> U {
    host.to_be()
}
```

and make rustfunc like this:
```rust
"htons" => ctx.new_rustfunc(socket_hton::<u16>),
```

coolreader18: I think you'd have to give the trait bound `U: num_traits::PrimInt` for that to work.

vm/src/stdlib/socket.rs
```diff
@@ -449,6 +449,28 @@ fn socket_htonl(host: u32, vm: &VirtualMachine) -> PyResult {
...
+    }
+    else {
+        Ok(vm.new_int(network.to_le()))
+    }
+}
+
+fn socket_ntohs(network: u16, vm: &VirtualMachine) -> PyResult {
+    if cfg!(target_endian = "big") {
+        Ok(vm.new_int(network))
+    }
```
coolreader18: These seem like no-ops -- if the machine is big endian, leave it as is, or else if it's little endian, convert to little endian, i.e. leave it as is. You might be looking for `u{16,32}::from_{le,be}`

vm/src/stdlib/socket.rs
coolreader18: You can have all these functions just return `u16` or `u32` rather than a PyResult -- `ctx.new_rustfunc` takes care of converting the return type.

vm/src/stdlib/socket.rs
metaphorshear: Ah.  Yeah, I couldn't find those before.

metaphorshear:
One sec, I might have pushed that a bit early.

metaphorshear:
Let me know if you want those rebased.

coolreader18:
I think it's fine; thanks for contributing! I'll merge this once the ci finishes.

coolreader18:
It looks like the rustfmt job failed @metaphorshear, could you run `cargo fmt` and push again?

PR #1670, opened 2020-01-10
Refactor bytes/bytearray/byteinner

youknowone:


vm/src/obj/objbyteinner.rs
```diff
@@ -47,15 +49,15 @@ impl TryFromObject for PyByteInner {
...
 #[derive(FromArgs)]
-pub struct ByteInnerNewOptions {
+pub(crate) struct ByteInnerNewOptions {
     #[pyarg(positional_only, optional = true)]
     val_option: OptionalArg<PyObjectRef>,
     #[pyarg(positional_or_keyword, optional = true)]
     encoding: OptionalArg<PyStringRef>,
 }
 
 impl ByteInnerNewOptions {
```
coolreader18: If the struct you're implementing a function for is `pub(crate)`, then just a `pub fn` will be effectively a `pub(crate) fn`.

vm/src/obj/objbyteinner.rs
```diff
@@ -1184,14 +1191,14 @@ impl PyByteInner {
...
+            new_value
         }
     }
 
-    pub fn irepeat(&mut self, n: isize, _vm: &VirtualMachine) -> PyResult<()> {
+    pub(crate) fn irepeat(&mut self, n: isize) {
         if self.elements.is_empty() {
             // We can multiple an empty vector by any integer, even if it doesn't fit in an isize.
-            return Ok(());
+            return;
```
coolreader18: I think this type is useful for other crates that might use RustPython -- why do you think this should be `pub(crate)`?

vm/src/obj/objbytes.rs
```diff
@@ -35,23 +37,29 @@ use std::str::FromStr;
 /// - an integer";
 #[pyclass(name = "bytes")]
 #[derive(Clone, Debug)]
-pub struct PyBytes {
```
coolreader18: Making `PyBytes` and `PyByteArray`  `pub(crate)` does seem like a good idea, providing that `PyBytesLike` remains `pub`, so that people can specify that they want a function to support any sort of bytes.

vm/src/obj/objbyteinner.rs
youknowone: Thanks for pointint it out. I was not thinking of rustpython crate users at all. Let me check `pub(crate)`s again

youknowone: I reverted most of them except for a few cases that private is obvious

PR #1675, opened 2020-01-12
Demo improvements

coolreader18:
- Add a new snippet, asyncbrowser - a mini `async/await` executor built on the browser event loop.
  - Maybe eventually turn it into a module?
- Combine `index.js` and `main.js` into one, to avoid the page being essentially without content while the wasm payload loads.
- Add `rp.pyExec` and `rp.pyExecSingle` to the "documentation".
- Improve the ci testing for wasm
  - Handle serving the demo and waiting inside python, so that you can just build the demo and run `pipenv run pytest` without having to mess with the npm commands.
  - Deduplicate the `driver` fixture, move it to `wdriver` in `conftest.py`.


wasm/demo/src/index.js
```diff
@@ -1,11 +1,156 @@
...
+
+function updateSnippet() {
+    const selected = snippets.value;
+
+    // the require here creates a webpack context; it's fine to use it
+    // dynamically.
+    // https://webpack.js.org/guides/dependency-management/
+    const snippet = require(`raw-loader!../snippets/${selected}.py`);
+
+    editor.setValue(snippet);
```
palaviv: When I run this locally I get the following error when trying to switch snippet:
```
Uncaught TypeError: string.split is not a function
    at splitLinesAuto (codemirror.js?56b3:680)
    at Doc.splitLines (codemirror.js?56b3:6452)
    at Doc.eval (codemirror.js?56b3:6111)
    at Doc.eval [as setValue] (codemirror.js?56b3:3941)
    at CodeMirror.eval [as setValue] (codemirror.js?56b3:9682)
    at HTMLSelectElement.updateSnippet (index.js?b635:83)
```
Changing the following works for me:
```js
editor.setValue(snippet.default);
```

Is this something only I get?

wasm/tests/conftest.py
```diff
@@ -0,0 +1,108 @@
...
+        except OSError as ex:
+            time.sleep(0.01)
+            if time.perf_counter() - start_time >= timeout:
+                raise TimeoutError(
+                    "Waited too long for the port {} on host {} to start accepting "
+                    "connections.".format(port, host)
+                ) from ex
+
+
+from selenium import webdriver
```
palaviv: How about splitting the conftest into two files. One for the server setup and one for the selenium?

wasm/tests/conftest.py
```diff
@@ -0,0 +1,108 @@
...
+import http.server
+http.server.test(HandlerClass=http.server.SimpleHTTPRequestHandler, port={PORT})
+"""
+
+demo_dist = os.path.join(os.path.dirname(os.path.realpath(__file__)), "../demo/dist/")
+
+server_proc = None
+
+
+def pytest_sessionstart(session):
```
palaviv: How about changing this to be a module scope fixture?

wasm/demo/src/index.js
coolreader18: `raw-loader` was probably updated, but I have a local `yarn.lock` that I have in my git exclude file. I'll look into it.

wasm/tests/conftest.py
coolreader18: Then the server would have to be started and stopped for every test, and given that a simple http server is stateless anyway, I don't think it makes much of a difference.

wasm/tests/conftest.py
coolreader18: I think it's good to have it in one file, because ideally the connection between the server and the `wdriver` fixture should be an "implementation detail" (whatever that means for test infrastructure) -- essentially, the http server is just a means to an end of loading the demo into the webdriver.

wasm/demo/src/index.js
coolreader18: It should be fixed now.

coolreader18: Should be fixed now.

PR #1666, opened 2020-01-08
Add __length_hint__ support for iterators

coolreader18:

PR #1676, opened 2020-01-12
Fix test_bool::test_sane_len

youknowone:


youknowone:
The result is wierd

```
test_sane_len (test.test_bool.BoolTest) ... unexpected success
```
This test is not marked as `expectedFailure` but doesn't show ok. It is success on my local machine so not easy to dig in.

youknowone:
@palaviv do you have any idea about this?

palaviv:
I am not sure what happens here. This test is flaky. I think marking the CPython test suite as allowed failed would be the best until we see this passing consistently.

coolreader18:
Oh, this might be my fault... I panicked when I pushed a commit to master instead of #1666 so I merged it to master right away, but #1666 had apparently broken a CPython test, so I pushed a commit to master last night to fix it.

palaviv:
@youknowone is this rebased in the latest master?

youknowone:
I thought so - but it didn't work. trying again

palaviv:
Seems like everything is OK now :smile:. Can we close this?

youknowone:
I found I fixed another test but didn't change the test decorator. yes, please merge and close this

PR #1679, opened 2020-01-12
math: Implement math.nextafter

corona10:
https://docs.python.org/3.9/library/math.html#math.nextafter is introduced yesterday :)

### DIFF
--- a/tests/snippets/stdlib_math.py
+++ b/tests/snippets/stdlib_math.py
@@ -240,6 +240,19 @@ assert math.factorial(10) == 3628800
 assert math.factorial(20) == 2432902008176640000
 assert_raises(ValueError, lambda: math.factorial(-1))
 
+if hasattr(math, 'nextafter'):
+    try:
+        assert math.nextafter(4503599627370496.0, -INF) == 4503599627370495.5
+        assert math.nextafter(4503599627370496.0, INF) == 4503599627370497.0
+        assert math.nextafter(9223372036854775808.0, 0.0) == 9223372036854774784.0
+        assert math.nextafter(-9223372036854775808.0, 0.0) == -9223372036854774784.0
+        assert math.nextafter(4503599627370496, -INF) == 4503599627370495.5
+        assert math.nextafter(2.0, 2.0) == 2.0
+        assert math.isnan(math.nextafter(NAN, 1.0))
+    except NotImplementedError:
+        # WASM
+        pass
+
 assert math.modf(1.25) == (0.25, 1.0)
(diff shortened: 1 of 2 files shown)


corona10:
@youknowone Can you please take a look?

tests/snippets/stdlib_math.py
```diff
@@ -240,6 +240,15 @@ def assertAllNotClose(examples, *args, **kwargs):
 assert math.factorial(20) == 2432902008176640000
 assert_raises(ValueError, lambda: math.factorial(-1))
 
+assert math.nextafter(4503599627370496.0, -INF) == 4503599627370495.5
```
youknowone: if this is 3.9 feature, CPython in CI will not run this test. Adding version check will work

corona10: Oh, thanks, this is 3.9 feature, I will add it

corona10:
@youknowon I 've updated the code only running on RustPython.

windelbouwman:
Nice one! I think we should follow CPython while we are still not on par with cpython. Allthough this is a python 3.9 thing, I think we could still add it.

tests/snippets/stdlib_itertools.py
```diff
@@ -1,4 +1,5 @@
 import itertools
+import platform
```
youknowone: This line looks like a mistake

vm/src/stdlib/math.rs
```diff
@@ -17,8 +17,15 @@ use crate::obj::objtype;
 use crate::pyobject::{PyObjectRef, PyResult, TypeProtocol};
 use crate::vm::VirtualMachine;
 
+use libc::c_double;
+
 use std::cmp::Ordering;
 
+#[link_name = "cmath"]
+extern "C" {
+    pub fn nextafter(x: c_double, y: c_double) -> c_double;
```
youknowone: is this function need to be a `pub`?

tests/snippets/stdlib_itertools.py
corona10: Thanks :)

vm/src/stdlib/math.rs
corona10: Not yet all

corona10:
@youknowone Thanks for the review, Updated!

corona10:
@youknowone @windelbouwman 
Finally, all CIs are passed.
- Change the unit test to be attribute-based. 
- raise NotImplementedError for WASM.

youknowone:
Not sure `NotImplementedError` fit for this use, but basically it looks good.

PR #1684, opened 2020-01-14
Refactor small things

youknowone:


vm/src/vm.rs
```diff
@@ -764,8 +758,7 @@ impl VirtualMachine {
         code_object: &bytecode::CodeObject,
```
coolreader18: Maybe we can move this to `objfunction.rs` so that we can make the fields all private?

youknowone: I did it. Will you please review newly added commits?

youknowone: I moved a few methods and now it looks better

coolreader18: Looks good!

PR #1685, opened 2020-01-14
pyslot guess slot name from function name

youknowone:


derive/src/pyclass.rs
```diff
@@ -207,15 +207,25 @@ impl Class {
...
-                let slot_ident = match nesteds.into_iter().next().unwrap() {
-                    NestedMeta::Meta(Meta::Word(ident)) => ident,
-                    bad => bail_span!(bad, "{}", pyslot_err),
+                let slot_ident = if nesteds.is_empty() {
+                    let ident_str = sig.ident.to_string();
+                    if ident_str.starts_with("tp_") {
+                        let slot_name = &ident_str[3..];
+                        proc_macro2::Ident::new(slot_name, sig.ident.span())
+                    } else {
+                        sig.ident.clone()
```
coolreader18: I think if it's empty and not `tp_*` it should throw an error. Otherwise `#[pyslot] fn new()` would be a valid new slot, and we want the convention to be `tp_new`

youknowone: for `new` and `tp_new`, unfortunately `new` is a conventional method name in rust. About other slots, it doesn't.
By reviewing other slot names, the name without `tp_` is more close to our common convention. See `hash`, `call`, `str`, `repr` and `init`. Otherwise we need to put `tp_` prefix to every this function.

We already will know they are slot because they will be attributed by `#[pyslot]`

```
 85     'tp_name', 
 86     'tp_basicsize', 
 87     'tp_itemsize', 
 88     'tp_dealloc', 
 89     'tp_print', 
 90     'tp_getattr', 
 91     'tp_setattr', 
 92     'tp_reserved', 
 93     'tp_repr', 
 94     'tp_as_number', 
 95     'tp_as_sequence', 
 96     'tp_as_mapping', 
 97     'tp_hash', 
 98     'tp_call', 
 99     'tp_str', 
100     'tp_getattro', 
101     'tp_setattro', 
102     'tp_as_buffer', 
103     'tp_flags', 
104     'tp_doc', 
105     'tp_traverse', 
106     'tp_clear', 
107     'tp_richcompare', 
108     'tp_weaklistoffset', 
109     'tp_iter', 
110     'iternextfunc', 
111     'tp_methods', 
112     'tp_members', 
113     'tp_getset', 
114     'tp_base', 
115     'tp_dict', 
116     'tp_descr_get', 
117     'tp_descr_set', 
118     'tp_dictoffset', 
119     'tp_init', 
120     'tp_alloc', 
121     'tp_new', 
122     'tp_free', 
123     'tp_is_gc', 
124     'tp_bases', 
125     'tp_mro', 
126     'tp_cache', 
127     'tp_subclasses', 
128     'tp_weaklist', 
129     'tp_del', 
130     'tp_version_tag', 
```

coolreader18: Alright, that makes sense.

Issue #1411, opened 2019-09-24
Replace uses of mem::uninitialized() with MaybeUninit<T>


dralley:
```mem::uninitialized()``` is being deprecated in Rust 1.38 for being able to cause undefined behavior

https://doc.rust-lang.org/std/mem/fn.uninitialized.html

coolreader18:
This is difficult, because we're using `mem::uninitialized` in a place where it can't easily be replaced with `MaybeUninit`, and the benefits we'd gain from maybe figuring out some way around it (a duplicate PyObject type but with `MaybeUninit` for fields?) wouldn't necessarily be worth it.

dralley:
Fixed by https://github.com/RustPython/RustPython/pull/1591

silmeth:
I still am very unsure about the fix, it effectively just replaced `mem::uninitialized()` with `MaybeUninit::uninit().assume_init()` which doesn’t remove any possible previous undefined behaviour.

As I commented on that PR:

> @coolreader18 An uninitialized `PyRef` is created, which holds an uninitialized `Rc<_>`, which then holds an unitialized [`NonNull<RcBox<T>>`](https://doc.rust-lang.org/src/alloc/rc.rs.html#279) – which, by the virtue of being uninitialized, might actually be `null`. That is UB (even if you don’t dereference it). And that is the reason why Rust team changed `mem::uninitialized()` into `mem::MaybeUninit` in the first place.
> 
> If it works now, it might not work in the future. To actually do it safely, you need to first do `prt::write` on `MaybeUninit`, and only then (when you know what bit pattern is at the given pointer), `assume_init()`. At least that’s my understanding.

coolreader18:
I've been thinking about this recently, and I agree with you @silmeth that this isn't ideal. Do you know if it would be undefined behavior if we initialized `Rc<MaybeUninit<PyObject>>`s, `ptr::write` the type and mro, and then casted them to `PyObjectRef`s?

silmeth:
@coolreader18 Not sure, when I was replying to that PR back then I was thinking about something like that:
1. create **`mut`** (?) `Rc<MeybeUninit<PyObject<…>>>`,
2. clone it as `cloned`,
3. create a `PyObject` with `typ` initialized to `PyRef { obj: cloned as Rc<PyObject<…>> }`,
4. `ptr_write` that py object as the content of the `mut Rc` from 1. and call it a day.

[...]

PR #1687, opened 2020-01-15
Purge all UB from init_type_hierarchy

coolreader18:
My plan I had in #1411 didn't work out because I failed to account for `PyClassRef` and `Rc<PyObject<PyClass>>` not being the same size due to the former being a trait object, so I went with using raw trait object manipulation.

cc @silmeth

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -642,7 +642,7 @@ impl<T: PyValue> PyRef<T> {
         }
     }
 
-    fn new_ref_unchecked(obj: PyObjectRef) -> Self {
+    pub(crate) fn new_ref_unchecked(obj: PyObjectRef) -> Self {
         PyRef {
             obj,
             _payload: PhantomData,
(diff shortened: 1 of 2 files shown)


silmeth:
Wow, took me a while to get the `mem::unititialized()` trick checking if all fields are present (you might want to change that to `MaybeUninit::uninit().assume_unit()` in line 252 to get rid of the usage of deprecated API though).

Also, I don’t really like the hack around the `PyClassRef` trait objects, but I don’t see any other way around it. So in general looks good to me. Would be worth to get some more advanced Rustacean to look at it though.

dralley:
The Rust Programming Language discord server has channels dedicated to asking for code review and discussing Rust "black magic" (advanced and/or unsafe code)

They might be willing to help. https://discord.gg/Qgs32H

coolreader18:
I'm not completely sure that this has no UB, but I'm going to merge because I think it's better than what we have now, and because the clippy builds are failing because of `invalid_value` in the current implementation.

PR #1688, opened 2020-01-15
Allow #[pyimpl] on traits

coolreader18:


vm/src/descriptor.rs
```diff
@@ -5,7 +5,8 @@ use crate::VirtualMachine;
 // TODO: pyimpl compose pyslot(descr_get) and pymethod(__get__) from this trait
```
youknowone: great! now we can remove this comment! :D

PR #1692, opened 2020-01-15
Improve test check for windows platform

alvinlindstam:
`'win' in sys.platform` returns true on MacOS, since it's platform string is [`"darwin"`](https://docs.python.org/3/library/sys.html#sys.platform). This was likely not intended, since the check seems to be testing for a windows platform, and was the only test error on a clean master build for me.

### DIFF
--- a/tests/snippets/stdlib_select.py
+++ b/tests/snippets/stdlib_select.py
@@ -20,7 +20,7 @@ assert_raises(TypeError, select.select, [Almost()], [], [])
 assert_raises(TypeError, select.select, [], [], [], "not a number")
 assert_raises(ValueError, select.select, [], [], [], -1)
 
-if "win" in sys.platform:
+if sys.platform in ["win32", "cygwin"]:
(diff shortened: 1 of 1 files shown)


alvinlindstam:
Ah, yes. I did not check for other places, as this was the only thing that failed for me on Mac OS. I wouldn't mind creating a function for this in test utils if desired. Alternatively, [`platform.system() == "Windows"`](https://docs.python.org/3/library/platform.html#platform.system) seems to be a clearer python pattern for detecting windows than using `sys.platform`. Another approach could be to get `platform.system()` implemented in rustpython and replace the checks with that (I'm new to the project, so I don't know how much work that is though).

youknowone:
Thanks for contributing.

Your suggestion make sense. Do you want to try this more?

alvinlindstam:
I might look into it, but not in a way where I'll be able to promise much success (noob to the project and to rust)

PR #1693, opened 2020-01-17
Added WAPM badge

syrusakbary:
We just created badges for WAPM packages, so projects that are published there can use them.

It will appear in the README like this:
[WAPM package](https://wapm.io/package/rustpython)

Hope you find it useful! :)


palaviv:
Looks great. Thanks for the contribution.

PR #1695, opened 2020-01-17
Add Cpython test_dict

palaviv:
I will take this opportunity to talk about the `gc` module. Should we remove it from our tests or should we create a stub?
I believe removing it from the tests is the correct way.


coolreader18:
I'm ambivalent regarding the `gc` module, I think removing it and its uses is good, but it might be less effort to have a stub module.

PR #1699, opened 2020-01-24
Support keepends in Str.splitlines

palaviv:


vm/src/obj/objstr.rs
```diff
@@ -778,11 +784,16 @@ impl PyString {
 
     // doesn't implement keep new line delimiter just yet
```
```suggestion
```

vm/src/obj/objstr.rs
```diff
@@ -778,11 +784,16 @@ impl PyString {
...
+        let end = if let OptionalArg::Present(true) = args.keepends {
+            "\n"
+        } else {
+            ""
+        };
         let elements = self
             .value
             .split('\n')
-            .map(|e| vm.ctx.new_str(e.to_string()))
+            .map(|e| vm.ctx.new_str(e.to_string() + end))
```
coolreader18: Looks like this is an issue; it shouldn't be appended for the last one. Maybe this, using itertools?

```rust
let elements = self
    .value
    .split('\n')
    .intersperse(end)
    .chunks(2)
    .map(|chunk| vm.ctx.new_str(chunk[0].to_string() + chunk.get(1).unwrap_or("")))
    .collect();
```

palaviv: I liked the suggestion but it did not solve `'hello\nhallo\nHallo\n'.splitlines(keepends=True)`. I went on the manual way. Seems to work...

PR #1700, opened 2020-01-25
GitHub actions

palaviv:
As suggested in the chat by @dralley this is an initial configuration of Github actions. My suggestion is to first run this in parallel to the travis CI and once this is stable remove travis.
The windows CI fails on the snippets test. I have not yet discovered why...

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,4 +1,4 @@
-on: [push]
+on: [push, pull_request]
 
 name: CI
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -4,11 +4,11 @@ name: CI
 
 jobs:
   rust_tests:
-    name: Run Rust tests
+    name: Run tests
     runs-on:  ${{ matrix.os }}
     strategy:
       matrix:
-        os: ["macos-latest", "ubuntu-latest", "windows-latest"]
+        os: [macos-latest, ubuntu-latest, windows-latest]
       fail-fast: false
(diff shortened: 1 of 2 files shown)


palaviv:
I managed to find the problem with the windows build and fixed it. Not sure why the github actions are not running on the PR. Maybe the change need to be in master for that to happen...

coolreader18:
Did you configure GitHub Actions to run in this repository? I'm not sure if that's still necessary (I think it was when they were more in beta), but it's worth a shot to see.

palaviv:
It worked automatically in my fork... I think once it will be merged it should work automatically.

abitrolly:
I don't like Travis or Cirrus CI running through GitHub actions, because it requires two more clicks to get to the failing build details,

dralley:
@abitrolly  What do you mean running "through" Github Actions?  It would run in parallel would it not?

abitrolly:
@dralley in parallel is ok. I just saw setups when Travis integrated as a GitHub action. For example, here is the integration of Cirrus CI through actions - https://github.com/containers/libpod/runs/344402331 - to get the full log you need to scroll down to the button and click the tiny gray link.

palaviv:
My suggestion is to keep the old CI alive for a few month. Then we should probably see what CI people like that most and keep only with that one.

coolreader18:
This looks fine, if it does run once merged than that's good, otherwise we can debug it from there.

coolreader18:
Yep, looks like it's working: https://github.com/RustPython/RustPython/actions

PR #1703, opened 2020-01-25
Fix Azure Pipelines build

youknowone:


azure-pipelines.yml
```diff
@@ -8,8 +8,8 @@ jobs:
     vmImage: 'vs2017-win2016'
   strategy:
     matrix:
-      Python36:
-        python.version: '3.6'
+      Python38:
```
palaviv: Would this make the snippets run with python 3.8? Our current goal is 3.6

youknowone: There was a discussion in #1608

palaviv: Missed that...

PR #1705, opened 2020-01-25
Bump up to python 3.7 for Azure Pipelines

youknowone:
Split from #1703

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -7,15 +7,12 @@ jobs:
   pool:
     vmImage: 'vs2017-win2016'
   strategy:
-    matrix:
-      Python36:
-        python.version: '3.6'
     maxParallel: 10
 
   steps:
   - task: UsePythonVersion@0
     inputs:
-      versionSpec: '$(python.version)'
+      versionSpec: '3.7'
       architecture: 'x64'
 
   - script: |


azure-pipelines.yml
```diff
@@ -8,8 +8,8 @@ jobs:
     vmImage: 'vs2017-win2016'
   strategy:
     matrix:
-      Python36:
-        python.version: '3.6'
+      Python38:
+        python.version: '3.8'
```
coolreader18: I think we can embed this directly in the `UsePythonVersion` task; I'm pretty sure this is what's causing the 2 different Azure jobs to run.

youknowone:
windows snippets test fail with python 3.8 but I don't have windows macine.

palaviv:
It seems like [os.path.realpath](https://docs.python.org/3/library/os.path.html#os.path.realpath) was changed in 3.8.

coolreader18:
The os.path.realpath implementation for 3.8 is [pretty complex](https://github.com/python/cpython/blob/3.8/Lib/ntpath.py#L537) and requires a fairly complex [`os` module function](https://github.com/python/cpython/blob/3.8/Modules/posixmodule.c#L3880), it might be best to comment it out for windows right now and put a TODO.

youknowone:
Changed to python3.7 for azure

PR #1706, opened 2020-01-26
Run test snippets for both mac/linux

youknowone:


.travis.yml
```diff
@@ -12,45 +12,40 @@ before_cache:
 jobs:
   fast_finish: true
   include:
-    - name: Run Rust tests(linux)
-      language: rust
+    - name: Run tests (Linux)
+      language: python
```
coolreader18: Have we ever tried using `language: rust` for these tests? Cause I think Python is already installed on the macos and linux environments, although it might not be the version we want.

youknowone:
hmm, it takes longer time than I expected

.travis.yml
youknowone: There is a comment: 
```
      # To test the snippets, we use Travis' Python environment (because
      # installing rust ourselves is a lot easier than installing Python)
```

coolreader18:
We have Github Actions running test snippets on all 3 major OS's right now; I know we're doing the "trying out different CIs" thing but travis is pretty overloaded as is, do you want to just use the GH actions CI?

youknowone:
i want to see how github actions work. I want to cover macOS better than now but also don't want to slow down travis more than now.

dralley:
So, here's one downside to Github Actions:

>Jobs that run on Windows and macOS runners that GitHub hosts consume minutes at 2 and 10 times the rate that jobs on Linux runners consume. For example, using 1,000 Windows minutes would consume 2,000 of the minutes included in your account. Using 1,000 macOS minutes, would consume 10,000 minutes included in a your account.

Open source projects using the free plan get 2000 minutes per month.  So just 200 minutes of combined CI time on MacOS would use up the entire allocation.  Which is honestly kinda BS, hopefully they get rid of that restriction.

https://help.github.com/en/github/setting-up-and-managing-billing-and-payments-on-github/about-billing-for-github-actions

coolreader18:
@dralley are you sure? It does say Github Free on that page, but on the [main Actions page](https://github.com/features/actions) it lists those pricing plans as for private repositories, and public repositories are free.

coolreader18:
Yeah, here's our billing page (https://github.com/organizations/RustPython/settings/billing); it specifically says minutes are for private repositories, and even though there have been like 3 builds so far on Actions, it lists no minutes used: GH Actions billing

dralley:
Great news then!  The fact that the pricing tier was literally called the "free tier" tripped me up.  A couple of other CI services (CircleCI was the one I had in mind) have minute restrictions in their "free" tier for open source projects.

palaviv:
> hmm, it takes longer time than I expected

We currently compile both in `release` and `debug` when running the snippets. I think we can make the time shorter by compiling in only one way throughout the whole build.

coolreader18:
I think either release mode or debug with `opt-level = 2` or something would be good. See also my [`no-azure` branch](https://github.com/RustPython/RustPython/compare/coolreader18/no-azure); I think my changes to `tests/test_snippets.py` resolve this.

coolreader18:
#1713

PR #1707, opened 2020-01-26
Add examples showcasing how to embed RustPython in Rust applications

cedric-h:
I'm sure examples like these were left out because you can just comb through the source, the documentation, pyckitup, etc. and figure out how everything works, and while that's definitely the case (I mean, I just did exactly that) doing so is quite time consuming, so I hope that I can leave these tutorials as breadcrumbs for anyone else who's interested in embedding RustPython into their applications.

I can say that personally I encountered RustPython a couple months back, and because there were no examples like these I simply assumed that embedding RustPython into Rust applications wasn't possible/feasible, and that it was meant simply to be a standalone application. Examples like these and the note in the README should make it much clearer to interested passerby that RustPython is not only embeddable into applications, but that it's also quite easy to do so.

### DIFF
--- a/examples/mini_repl.rs
+++ b/examples/mini_repl.rs
@@ -41,6 +41,7 @@ fn main() -> PyResult<()> {
             .map_err(|err| vm.new_syntax_error(&err))?;
 
         // this line also automatically prints the output
+        // (note that this is only the case when compile::Mode::Single is passed to vm.compile)
         let output = vm.run_code_obj(code_obj, scope.clone())?;
 
         // store the last value in the "last" variable
--- a/examples/mini_repl.rs
+++ b/examples/mini_repl.rs
@@ -3,50 +3,113 @@
 ///! This much smaller REPL is still a useful example because it showcases inserting
 ///! values and functions into the Python runtime's scope, and showcases use
 ///! of the compilation mode "Single".
-///! Note that in particular this REPL does a horrible job of showing users their errors
-///! (instead it simply crashes).
-use rustpython_compiler::compile;
-use rustpython_vm::{pyobject::PyResult, PySettings, VirtualMachine};
+use rustpython_compiler as compiler;
+use rustpython_vm as vm;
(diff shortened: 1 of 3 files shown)


examples/mini_repl.rs
```diff
@@ -0,0 +1,54 @@
...
+///! of the compilation mode "Single".
+///! Note that in particular this REPL does a horrible job of showing users their errors
+///! (instead it simply crashes).
+use rustpython_compiler::compile;
+use rustpython_vm::{pyobject::PyResult, PySettings, VirtualMachine};
+// this needs to be in scope in order to insert things into scope.globals
+use rustpython_vm::pyobject::ItemProtocol;
+
+fn main() -> PyResult<()> {
+    let mut on = true;
```
coolreader18: You can make this a `Rc<Cell<bool>>` to avoid the unsafe code later on; you could also put a comment here about why things like `Rc` and `Cell` and `RefCell` are necessary when interacting with RustPython (b/c Python doesn't respect Rust's borrow checker).

examples/mini_repl.rs
```diff
@@ -0,0 +1,54 @@
...
+
+    // typing `quit()` is too long, let's make `on(False)` work instead.
+    scope
+        .globals
+        .set_item(
+            "on",
+            vm.context().new_rustfunc({
+                let on: *mut bool = &mut on;
+                move |b: bool, _: &VirtualMachine| unsafe { *on = b }
+            }),
```
coolreader18: We actually renamed `new_rustfunc` to `new_function` recently; with the above change this can just be `vm.context().new_function(|b: bool| on.set(b))`.

cedric-h: I still needed a separate scope in which to clone the `Rc`, but thanks for the tip about `new_function`!

cedric-h:
We probably want more examples, especially on idiomatic Rust interop, freezing entire Python modules, etc.

These are definitely a good start though.

PR #1708, opened 2020-01-26
Utility ClassItemMeta for class item meta parsing

youknowone:
Parsing in for loop is painful.
Doing with a hashmap is easier.

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -42,6 +42,182 @@ enum ClassItem {
     },
 }
 
+struct ClassItemMeta<'a> {
+    sig: &'a Signature,
+    parent_type: &'static str,
+    meta: HashMap<String, Option<Lit>>,
+}
+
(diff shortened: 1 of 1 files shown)


derive/src/pyclass.rs
```diff
@@ -42,6 +42,182 @@ enum ClassItem {
...
 
+struct ClassItemMeta<'a> {
+    sig: &'a Signature,
+    parent_type: &'static str,
+    meta: HashMap<String, Option<Lit>>,
+}
+
+impl<'a> ClassItemMeta<'a> {
+    const METHOD_NAMES: [&'static str; 2] = ["name", "magic"];
+    const PROPERTY_NAMES: [&'static str; 3] = ["name", "magic", "setter"];
```
coolreader18: You can make these `const _NAMES: &'static [&'static str] = &[...];` so that you don't have to change the size in the type when you add a new one.

PR #1710, opened 2020-01-27
Remove wtf8 dependency

dralley:
```std::char``` can handle this

### CODE REVIEW

File: parser/src/lexer.rs [hunk 1]
```diff
@@ -442,23 +443,20 @@ where
...
                 },
-                None => return unicode_error,
+                None => return Err(unicode_error),
             }
         }
-        match wtf8::CodePoint::from_u32(p) {
-            Some(cp) => Ok(cp.to_char_lossy()),
-            None => unicode_error,
-        }
+        std::char::from_u32(p).ok_or(unicode_error)
```
coolreader18: This does seem right; cpython keeps characters like `'\uD801'` as `'\uD801'`, while `wtf8` would [replace it with the replacement character](https://docs.rs/wtf8/0.0.3/src/wtf8/lib.rs.html#108-127).

coolreader18: Oh no, based on the failing test, it seems CPython strings aren't actually utf-8; they allow surrogate code points (`'\uD800'.encode()` only fails on encode). That'll be questionable for compatibility... in any case, I'd say it would be best to basically embed the surrogate check right here; if `p` is in `D800..=DFFF`, use the REPLACEMENT CHARACTER; otherwise, `std::char::from_u32(p).ok_or(unicode_error)`

dralley: w/ code

```
        match p {
            '\u{D800}'..='\u{DFFF}' => Ok(std::char::REPLACEMENT_CHARACTER),
            _ =>std::char::from_u32(p).ok_or(unicode_error)
        }
```

I get a bunch of

```
   Compiling rustpython-parser v0.1.1 (/home/dalley/Devel/RustPython/parser)
error: invalid unicode character escape
   --> parser/src/lexer.rs:460:14
    |
460 |             '\u{D800}'..='\u{DFFF}' => Ok(std::char::REPLACEMENT_CHARACTER),
    |              ^^^^^^^^
    |
    = help: unicode escape must not be a surrogate
```

coolreader18: Well `p` is a u32, so you'd want to use integer literals instead of character literals.

dralley: Derp, my bad, I should've looked at the link you pasted.  It should be good now

PR #1711, opened 2020-01-27
Replace lazy_static! macro with once_cell

dralley:
```once_cell``` is generally considered to be superior to ```lazy_static``` nowadays since it doesn't require a macro and has some extra flexibility.  The ongoing proposal to add lazy initialization to the stdlib basically copies the ```once_cell``` API.

https://github.com/rust-lang-nursery/lazy-static.rs/issues/111
https://github.com/rust-lang/rfcs/pull/2788

### DIFF
--- a/vm/src/vm.rs
+++ b/vm/src/vm.rs
@@ -15,6 +15,7 @@ use std::sync::{Mutex, MutexGuard};
 use arr_macro::arr;
 use num_bigint::BigInt;
 use num_traits::ToPrimitive;
+use once_cell::sync::Lazy;
 #[cfg(feature = "rustpython-compiler")]
 use rustpython_compiler::{compile, error::CompileError};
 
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: vm/src/vm.rs [hunk 1]
```diff
@@ -1322,9 +1323,7 @@ impl Default for VirtualMachine {
     }
 }
 
-lazy_static! {
-    static ref REPR_GUARDS: Mutex<HashSet<usize>> = { Mutex::new(HashSet::new()) };
-}
+static REPR_GUARDS: Lazy<Mutex<HashSet<usize>>> = Lazy::new(|| Mutex::new(HashSet::new()));
```
coolreader18: Could this be `Lazy::default()`?

dralley: Probably!

dralley: Unfortunately I don't think so.  I don't see anything like that documented and throws a compile error

coolreader18: Really? It looks like `Default` is implemented for `Lazy<T: Default>` [here](https://docs.rs/once_cell/1.3.1/src/once_cell/lib.rs.html#930-935).

coolreader18: Oh, maybe it's not a `const fn`. Could do `Lazy::new(Mutex::default)`, though.

dralley: Yup, that's it.  But the latter strategy works

PR #1712, opened 2020-01-27
Add Py_TPFLAGS_BASETYPE support to restrict subclass

youknowone:


youknowone:
I tried bitflags but `|` operator is not const fn so the const part is still u64.
yes. `BASETYPE` is used by `best_base`. as `has_feature()` argument.

coolreader18:
You might have to update the `bitflags` dependency, but I think you could use `bitflags!` and generate something like `PyTpFlags::from_bits_truncate(PyTpFlags::DEFAULT.bits() | PyTpFlags::FLAG1.bits() | PyTpFlags::FLAG1.bits())` that would work in a `const` context.

vm/src/exceptions.rs
```diff
@@ -447,72 +447,84 @@ pub struct ExceptionZoo {
...
-        let module_not_found_error = create_type("ModuleNotFoundError", &type_type, &import_error);
-        let not_implemented_error = create_type("NotImplementedError", &type_type, &runtime_error);
-        let recursion_error = create_type("RecursionError", &type_type, &runtime_error);
-        let eof_error = create_type("EOFError", &type_type, &exception_type);
-        let indentation_error = create_type("IndentationError", &type_type, &syntax_error);
-        let tab_error = create_type("TabError", &type_type, &indentation_error);
-        let unicode_error = create_type("UnicodeError", &type_type, &value_error);
-        let unicode_decode_error = create_type("UnicodeDecodeError", &type_type, &unicode_error);
-        let unicode_encode_error = create_type("UnicodeEncodeError", &type_type, &unicode_error);
+        let base_exception_type = create_exception_type("BaseException", &type_type, &object_type);
```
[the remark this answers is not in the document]
coolreader18: I might just be not seeing a borrow rule, but do you have to pass `&type_type` to the closure? Why not just capture `type_type` in the closure environment and pass it to `create_type`?

vm/src/slots.rs
```diff
@@ -2,8 +2,28 @@ use crate::function::{OptionalArg, PyFuncArgs, PyNativeFunc};
...
+bitflags! {
+    pub struct PyTpFlags: u64 {
+        const DEFAULT = 0;
+        const BASETYPE = 1 << 10;
+    }
+}
+
+impl PyTpFlags {
+    pub fn has_feature(self, flag: Self) -> bool {
+        (self & flag).bits != 0
```
coolreader18: I think you could just use the `contains` method that is generated by bitflags instead of defining this method.

vm/src/pyobject.rs
```diff
@@ -1196,10 +1197,12 @@ where
 }
 
 pub trait PyClassImpl: PyClassDef {
+    const TP_FLAGS: u64;
```
coolreader18: You could give this a default value. so that you don't have to do it for `pystruct_sequence` or other pyclass derivatives we might have in the future

vm/src/slots.rs
```diff
@@ -2,8 +2,28 @@ use crate::function::{OptionalArg, PyFuncArgs, PyNativeFunc};
 use crate::pyobject::{IdProtocol, PyObjectRef, PyRef, PyResult, PyValue};
 use crate::VirtualMachine;
 
+bitflags! {
+    pub struct PyTpFlags: u64 {
+        const DEFAULT = 0;
```
coolreader18: `bitflags!` generates a "default" in the form of `FlagsType::empty()`, you might want to use that.

vm/src/exceptions.rs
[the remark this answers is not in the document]
youknowone: oh, that must be true. I was lazy to fix exiting code much.

vm/src/slots.rs
youknowone: this is 0 here because i only implemented BAESTYPE. but it can be a non-zero value in CPython. I will add a comment

PR #1713, opened 2020-01-28
Fix up some testing stuff; only build one target profile of rustpython when testing

coolreader18:


coolreader18:
I also just switched off the [build pushed branches](https://docs.travis-ci.com/user/web-ui/#build-pushed-branches) option in Travis' web settings, so now it shouldn't do both `pr` and `push` for every PR.

PR #1717, opened 2020-01-30
Add repr for dict_keys, dict_values and dict_items

palaviv:


vm/src/obj/objdict.rs
```diff
@@ -495,6 +495,22 @@ macro_rules! dict_iterator {
             fn len(&self, vm: &VirtualMachine) -> usize {
                 self.dict.clone().len(vm)
             }
+
+            #[pymethod(name = "__repr__")]
+            #[allow(clippy::redundant_closure_call)]
```
youknowone: which part raises this warning?

palaviv: It is the use of `result_fn` same as in `next`

PR #1720, opened 2020-01-30
Split rust test and snippets in github actions

palaviv:
This will decrease the CI time from ~25 min to ~18 min. The github actions [quotas](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-github-actions#usage-limits) are 20 concurrent builds (5 macOS) so I think this should be OK.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,7 +7,25 @@ name: CI
 
 jobs:
   rust_tests:
-    name: Run tests
+    name: Run rust tests
+    runs-on:  ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -7,7 +7,7 @@ name: CI
 
 jobs:
   rust_tests:
-    name: Run tests
+    name: Run rust tests
     runs-on:  ${{ matrix.os }}
```
coolreader18: Do we need to run Rust tests on all OSs? In theory we wouldn't be testing any platform specific behavior.

palaviv: I agree that we currently do not test anything OS specific but that can change...

PR #1721, opened 2020-01-30
Lib/test/test_types.py from CPython v3.8.1

youknowone:


Lib/test/test_types.py
```diff
@@ -553,25 +563,29 @@ def test(f, format_spec, result):
         test(12345.6, "1=20", '111111111111112345.6')
         test(12345.6, "*=20", '*************12345.6')
 
-    def test_format_spec_errors(self):
-        # int, float, and string all share the same format spec
-        # mini-language parser.
+    # TODO: RUSTPYTHON
+    # @unittest.expectedFailure
+    # def test_format_spec_errors(self):
```
palaviv: Why is this test commented out?

Lib/test/test_types.py
```diff
@@ -1150,22 +1194,23 @@ class N(type, metaclass=M):
 
 class SimpleNamespaceTests(unittest.TestCase):
 
-    def test_constructor(self):
-        ns1 = types.SimpleNamespace()
-        ns2 = types.SimpleNamespace(x=1, y=2)
-        ns3 = types.SimpleNamespace(**dict(x=1, y=2))
+    # TODO: RUSTPYTHON
+    # def test_constructor(self):
```
palaviv: Why is this test commented out?

youknowone: there is rust part bug to make crash and I found I can't fix it quickly.

palaviv: Thats fine but lets mark tests like that with `@unittest.skip("TODO: RUSTPYTHON")` instead of comment out. It will make them easier to find later.

youknowone: In that case, the test immediately stopped by crash so never run later tests.

youknowone: Oh, ok skip would be the correct solution

PR #1722, opened 2020-01-30
Change platform to CPython pure python platform

palaviv:
You can now use most of platform methods. Not sure about testing as this might be tricky when running on different OS's and both in RustPython and in CPython. Any suggestions?


coolreader18:
It looks like CPython has a test for `platform`, do you want to try that?

vm/src/stdlib/platform.rs
```diff
@@ -1,39 +0,0 @@
-use crate::pyobject::PyObjectRef;
-use crate::version;
-use crate::vm::VirtualMachine;
-
-pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
-    let ctx = &vm.ctx;
-    py_module!(vm, "platform", {
```
windelbouwman: Instead of deletion, I would prefer to keep this file and rename it to `_platform`. The python code of the platform module is a bit messy, so it might be good to selectively copy paste it from the cpython sourcetree. What do you think?

Lib/platform.py
```diff
@@ -0,0 +1,1260 @@
...
+    if sys_version is None:
+        sys_version = sys.version
+
+    # Try the cache first
+    result = _sys_version_cache.get(sys_version, None)
+    if result is not None:
+        return result
+
+    # Parse it
+    if 'IronPython' in sys_version:
```
windelbouwman: We could remove this ironpython specific code, right? What is the reason for keeping the code here? Is the idea to merge this file into the cpython source base and copy paste it regularly from there?

Lib/platform.py
```diff
@@ -0,0 +1,1260 @@
...
+    r'\(([\d.]+)\) on ([\w.]+ [\d.]+(?: \(\d+-bit\))?)\)'
+)
+
+_pypy_sys_version_parser = re.compile(
+    r'([\w.+]+)\s*'
+    r'\(#?([^,]+),\s*([\w ]+),\s*([\w :]+)\)\s*'
+    r'\[PyPy [^\]]+\]?')
+
+_rustpython_sys_version_parser = re.compile(
+    r'RustPython '
```
windelbouwman: Instead of this parsing, I would prefer the previous rust methods which return the information directly.

windelbouwman:
I think testing this module is hard in general. If there are CPython test cases, we could try them, but it might be too platform specific.

windelbouwman:
I have some doubts about the parsing of the version string, and the code supporting other python implementations with this parsing, but for the rest, this looks pretty good!

Lib/platform.py
coolreader18: I think that might be good to do; it seems like ironpython and jython and pypy are all supported in this module, I think adding `platform.py` support for parsing our own sys.version syntax is kind of the "correct" thing for a Python interpreter to do. It might also be worth it to eventually try to merge back these changes so that you could parse RustPython `sys_version`s from e.g. CPython. I'm not sure what the benefit of that is, but I'm sure somebody's used it for something.

youknowone: i don't remeber well but once i've read something that why pypy and ironpython is mentioned in this file but not pyston or any other newer interpreters. i wonder if there is any pep or something about non-cpython interpreters

palaviv:
I agree it looks nicer this way @windelbouwman.
I also added the tests from CPython. The platform has specific tests for Mac and Windows. I will change the CI to run the CPython tests suite on them as well in a different PR.

palaviv:
> I also added the tests from CPython. The platform has specific tests for Mac and Windows. I will change the CI to run the CPython tests suite on them as well in a different PR.

I have changed the CI here as I don't want to break the CPython tests for people working on Mac or Windows.

coolreader18:
@palaviv good idea, on ubuntu it only took 9 seconds to run the snippet tests; definitely not worth the whole other job to build it.

palaviv:
I am having hard time making the CPython test suite running on Windows. I opened a ticket to fix this later #1754.

PR #1724, opened 2020-01-31
Support timeout in Popen.communicate

palaviv:
The new version of rust-subprocess supports timeout hniksic/rust-subprocess#30


hniksic:
```rust
if let OptionalArg::Present(timeout) = args.timeout {
            communicator = communicator.limit_time(Duration::new(timeout, 0));
```

@palaviv I'm not acquainted with RustPython, but I think this could easily support fractional-second timeouts by retrieving the timeout as an `f64` and constructing the duration using [`from_secs_f64`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs_f64).

PR #1725, opened 2020-01-31
parse_number can raise error about length

youknowone:
And fixed crash when it is too long

### DIFF
--- a/vm/src/format.rs
+++ b/vm/src/format.rs
@@ -155,16 +155,17 @@ fn parse_fill_and_align(text: &str) -> (Option<char>, Option<FormatAlign>, &str)
     }
 }
 
-fn parse_number(text: &str) -> (Option<usize>, &str) {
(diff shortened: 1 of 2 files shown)


vm/src/format.rs
```diff
@@ -155,16 +155,19 @@ fn parse_fill_and_align(text: &str) -> (Option<char>, Option<FormatAlign>, &str)
...
+        return Ok((None, text));
     }
-    // This should never fail
-    (
+    if num_digits >= 20 {
+        // NOTE: this is different from CPython
+        return Err("Too many decimal digits in format string");
+    }
+    Ok((
         Some(text[..num_digits].parse::<usize>().unwrap()),
```
palaviv: Why not checking the return value of `parse` instead of checking for number of digits?

palaviv:
I think something went wrong with your rebase...

youknowone:
Thanks, I pushed it again

PR #1727, opened 2020-01-31
Fix int.from_bytes type

youknowone:


vm/src/obj/objint.rs
```diff
@@ -531,9 +531,13 @@ impl PyInt {
         zelf
     }
 
-    #[pymethod]
+    #[pyclassmethod]
     #[allow(clippy::match_bool)]
-    fn from_bytes(args: IntFromByteArgs, vm: &VirtualMachine) -> PyResult<BigInt> {
+    fn from_bytes(
+        _class: PyClassRef,
```
coolreader18: You might want to use this class and return `PyInt::from(x).into_ref_with_type(vm, cls)` at the end.

PR #1731, opened 2020-02-01
Implement _random.Random

### LINKED ISSUE #1269: Determinism and compatibility of random module
Should `random.random()` always be same when same `random.seed()` is applied? Should this be same across interpreter process? Should this be same across RustPython version? Should this be same with how CPython does?

Python document states this in https://docs.python.org/3/library/random.html#notes-on-reproducibility

> Sometimes it is useful to be able to reproduce the sequences given by a pseudo random number generator. By re-using a seed value, the same sequence should be reproducible from run to run as long as multiple threads are not running.
> 
> Most of the random module’s algorithms and seeding functions are subject to change across Python versions, but two aspects are guaranteed not to change:
> 
> - If a new seeding method is added, then a backward compatible seeder will be offered.
> 
> - The generator’s random() method will continue to produce the same sequence when the compatible seeder is given the same seed.

`random.seed()` takes `version` argument which can be 1 or 2.

> With version 2 (the default), a str, bytes, or bytearray object gets converted to an int and all of its bits are used.
> 
> With version 1 (provided for reproducing random sequences from older versions of Python), the algorithm for str and bytes generates a narrower range of seeds.

coolreader18:
Based off of @malkog's work in #1571. Closes #1571 and closes #1269.

cc @vks.

### DIFF
--- a/vm/src/stdlib/random.rs
+++ b/vm/src/stdlib/random.rs
@@ -77,7 +77,7 @@ impl PyRandom {
 
     #[pymethod]
     fn random(&self) -> f64 {
-        gen_res53(&mut *self.rng.borrow_mut())
+        mersenne::gen_res53(&mut *self.rng.borrow_mut())
     }
 
     #[pymethod]
@@ -127,12 +127,3 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         "Random" => PyRandom::make_class(ctx),
     })
 }
-
-// taken from the translated mersenne twister
-/* generates a random number on [0,1) with 53-bit resolution*/
-fn gen_res53<R: RngCore>(rng: &mut R) -> f64 {
-    let a = rng.next_u32() >> 5;
-    let b = rng.next_u32() >> 6;
-    (a as f64 * 67108864.0 + b as f64) * (1.0 / 9007199254740992.0)
-}
-/* These real versions are due to Isaku Wada, 2002/01/09 added */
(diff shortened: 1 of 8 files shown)


vm/src/stdlib/random.rs
```diff
@@ -1,46 +1,138 @@
...
+    #[pymethod]
+    fn random(&self) -> f64 {
+        gen_res53(&mut *self.rng.borrow_mut())
+    }
+
+    #[pymethod]
+    fn seed(&self, n: OptionalOption<PyIntRef>) {
+        let new_rng = match n.flat_option() {
+            None => PyRng::default(),
+            Some(n) => {
```
palaviv: I would like to suggest that instead of using `MT` when being called with seed we should keep using the `Std` with [SeedableRng](https://rust-num.github.io/num/rand/trait.SeedableRng.html). I think that users should receive the best available random as promised by `StdRng`.
I would suggest the following in order to allow using `MT`. Add the `version` parameter with a default of `3` and when this is called with `1` or `2` fallback to `MT`.

coolreader18: The version parameter is entirely managed in `random.py`; it only changes how strings and bytes are turned into seeds. The entire point of the mersenne twister implementation is to stay compatible with CPython's seeding, and if we don't do that then it's kind of pointless. `random` isn't meant to be cryptographically secure anyway, from the `random` docs: "The pseudo-random generators of this module should not be used for security purposes. For security or cryptographic uses, see the secrets module". I'd say that supplying actual random numbers when the seed is `None` is more than enough.

coolreader18:
I think that would be good to do; I'll look into it.

PR #1734, opened 2020-02-01
str::to_string -> str::to_owned

youknowone:
please reject this PR if it is out of our preference or rust community convention.


windelbouwman:
@youknowone could you explain a bit what your motivation is for this specific change? Reading [this blog post](https://blog.mgattozzi.dev/how-do-i-str-string/), it looks like there is no difference in the two methods.

youknowone:
There is a reddit link in the post you linked and I think this is my reason: https://www.reddit.com/r/rust/comments/4l71qw/how_do_i_convert_a_str_to_string_the_beginning_in/d3l23vf/

Derived pros:  For most of this call, we just want to create String from &str. But, because .to_string() is defined to Display trait, we accidently can put a ToString object instead of &str object - which is not mostly intended. For `to_owned()`, if any object not a &str called with this method, it mostly will not return String. Which means no mistake.

Probably `.into()` and `String::from()` can be a better choice for this reason. I think every other option except for `to_string()` is better then that.

This is a kind of human engineering. Here is another PR I made with similar motivation. https://github.com/RustPython/RustPython/pull/1654

windelbouwman:
@youknowone thank you for your detailed explanation!

PR #1738, opened 2020-02-02
Add getset_descriptor

youknowone:
- Add getset_descriptor
- Remove PyReadOnlyProperty
- #[pyproperty] now generates getset. PyProperty is only used for python code.

### DIFF
--- a/vm/src/obj/objgetset.rs
+++ b/vm/src/obj/objgetset.rs
@@ -1,0 +1,71 @@
+/*! Python `attribute` descriptor class. (PyGetSet)
+
+*/
+use super::objtype::PyClassRef;
+use crate::function::OptionalArg;
+use crate::pyobject::{PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue};
+use crate::slots::PyBuiltinDescriptor;
+use crate::vm::VirtualMachine;
+
+pub type PyGetter = dyn Fn(PyObjectRef, &VirtualMachine) -> PyResult;
+pub type PySetter = dyn Fn(PyObjectRef, PyObjectRef, &VirtualMachine) -> PyResult<()>;
+
+#[pyclass]
+pub struct PyGetSet {
+    // name: String,
+    getter: Box<PyGetter>,
+    setter: Box<PySetter>,
+    // doc: Option<String>,
+}
+
+impl std::fmt::Debug for PyGetSet {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "PyGetSet {{ getter: {:p}, setter: {:p} }}",
+            self.getter, self.setter
+        )
+    }
+}
+
+impl PyValue for PyGetSet {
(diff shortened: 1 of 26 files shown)


coolreader18:
How is PyGetSet different from PyProperty?

youknowone:
in CPython, PyGetSet corresponds to `getset_descriptor` while PyProperty corresponds to `property`. this is data specialized form of `property`. How about thinking like `PyBulitinMethod` and `PyBoundMethod` relationship?

Practically, it only corresponds with native functions with solid signature. So it must be related to performance. For compatibility point of view, the properties of `None` object matters too. Because descriptors get obj argument for `__get__` and checking its existance by not-equal-None, properties of `None` object is nessessarily broken. `getset_descriptor` doesn't check it but always call getter function.

coolreader18:
Maybe it would be good to replace all the property APIs (`new_property`, `PropertyBuilder`, etc) with these descriptors, if not in this PR then in another one. It seems like there's not really a benefit to using properties from native code, so in basically every situation you'd prefer getset over property (just not in, like, the `builtins.property()` decorator)

youknowone:
After grep CPython code, yes, that sounds true.

youknowone:
One problem left. I have no idea how CPython shows descriptor for following code while RustPython raises error.

```python
>>> Exception.args
<attribute 'args' of 'BaseException' objects>
```

vm/src/obj/objgetset.rs
```diff
@@ -0,0 +1,230 @@
...
+    }
+}
+
+pub type PyGetSetRef = PyRef<PyGetSet>;
+
+impl PyBuiltinDescriptor for PyGetSet {
+    fn get(
+        zelf: PyRef<Self>,
+        obj: PyObjectRef,
+        _cls: OptionalArg<PyObjectRef>,
```
coolreader18: Maybe to fix the exception.args case, you could have this check `cls` and if it's not there than just return `zelf`?

youknowone: I copied `descr_check` part of CPython. Now `descr_get` signature looks a little bit messy than before

youknowone:
@coolreader18 It seems this is finally done! will you review it please?

vm/src/stdlib/io.rs
```diff
@@ -573,7 +573,7 @@ mod fileio {
         let fileno = i64::try_from_object(vm, vm.get_attribute(instance.clone(), "__fileno")?)?;
         Ok(os::rust_file(fileno))
     }
-    fn fio_set_fileno(instance: &PyObjectRef, f: fs::File, vm: &VirtualMachine) -> PyResult<()> {
+    fn fio_set_fileno(instance: &PyObjectRef, f: fs::File, vm: &VirtualMachine) -> PySetResult {
```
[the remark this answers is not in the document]
youknowone: I changed the name `IntoPySetResult` to `IntoPyNoResult` and removed `PySetResult` type.

youknowone:
I attached #1736 to here because it is trivial derived work from this PR

youknowone:
Other getters of SyntaxError are actually not a getter by args. they are attributes set by `set_attr`

PR #1747, opened 2020-02-05
Fix a boolean test

arnavb:
I might be wrong, but the TypeError is for the example code. As far as I can tell, this test discards the exception before checking if it was raised, which seems wrong. Let me know what you think!

### DIFF
--- a/Lib/test/test_bool.py
+++ b/Lib/test/test_bool.py
@@ -6,19 +6,10 @@ from test import support
 import os
 
 class BoolTest(unittest.TestCase):
-
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_subclass(self):
-        try:
+        with self.assertRaises(TypeError):
             class C(bool):
                 pass
-        except TypeError:
-            pass
-        else:
-            self.fail("bool should not be subclassable")
-
-        self.assertRaises(TypeError, int.__new__, bool, 0)
 
     def test_print(self):
         try:


Lib/test/test_bool.py
```diff
@@ -6,19 +6,10 @@
...
-        try:
+        with self.assertRaises(TypeError):
             class C(bool):
                 pass
-        except TypeError:
-            pass
-        else:
-            self.fail("bool should not be subclassable")
-
-        self.assertRaises(TypeError, int.__new__, bool, 0)
```
coolreader18: I think the problem was in this line, not in the code above. Ideally we shouldn't have to change anything inside of the bodies of failing test functions, just remove the decorator.

The correct behavior here is to have `int.__new__(bool, 0)` raise `TypeError: int.__new__(bool) is not safe, use bool.__new__()`, and that's a bit more complex to fix. The relevant cpython code is [here](https://github.com/python/cpython/blob/46874c26ee1fc752e2e6930efa1d223b2351edb8/Objects/typeobject.c#L6086); essentially you have to check in `objtype.rs` (probably [here](https://github.com/RustPython/RustPython/blob/master/vm/src/obj/objtype.rs#L397)) that: the closest class down the `PyClass.base` chain starting from `cls` whose `new` slot isn't `None` is the same class as `zelf`, e.g. `staticbase.is(&zelf)`. It's somewhat complicated, and I totally get it if you don't want to try something so complex right now, but if you do want to try then I can help with any problems you might run in to.

arnavb: Cool, thanks for clarifying. It seems I misinterpreted the test function. Sorry!

PR #1749, opened 2020-02-06
Remove `_vm` parameter when it is not required

youknowone:


vm/src/exceptions.rs
```diff
@@ -90,51 +90,39 @@ impl PyBaseException {
     }
 
     #[pyproperty(name = "__cause__")]
-    fn get_cause(&self, _vm: &VirtualMachine) -> Option<PyBaseExceptionRef> {
```
coolreader18: Do you want to unify these setters and the ones for Rust code below?

vm/src/stdlib/socket.rs
```diff
@@ -445,11 +445,11 @@ fn socket_inet_ntoa(packed_ip: PyBytesRef, vm: &VirtualMachine) -> PyResult {
...
 
-fn socket_hton<U: num_traits::int::PrimInt>(host: U, _vm: &VirtualMachine) -> U {
+fn socket_hton<U: num_traits::int::PrimInt>(host: U) -> U {
     U::to_be(host)
 }
 
-fn socket_ntoh<U: num_traits::int::PrimInt>(network: U, _vm: &VirtualMachine) -> U {
+fn socket_ntoh<U: num_traits::int::PrimInt>(network: U) -> U {
     U::from_be(network)
 }
```
coolreader18: I think we could replace these functions with just `new_function(u{16,32}::{to,from}_be)`; I'm not sure if that's a good idea but we're giving the type arguments in the `make_module` function anyway.

PR #1757, opened 2020-02-10
Add more CPython tests

palaviv:
- test_baseexception
- test_int
- test_int_literal
- test_opcodes
- test_pow
- test_raise


youknowone:
I think adding cpython version you copied into commit message will be helpful in future when we want to update them

dralley:
Or as a comment at the top of each file

palaviv:
> I think adding cpython version you copied into commit message will be helpful in future when we want to update them

Done

PR #1759, opened 2020-02-10
Add as_integer_ratio to PyInt

minhnhdo:
This pull request adds support for `as_integer_ratio` to the VM's integer number type, `PyInt`.

## Current behavior on master
```python
>>>>> (10).as_integer_ratio()        
AttributeError: 'int' object has no attribute 'as_integer_ratio'
>>>>> 
```

## Behavior after this pull request
```python
>>>>> (10).as_integer_ratio()
(10, 1)
>>>>> 
```

### DIFF
--- a/tests/snippets/ints.py
+++ b/tests/snippets/ints.py
@@ -61,6 +61,12 @@ assert (6).__rmod__(10) == 4
 with assert_raises(ZeroDivisionError):
     (0).__rmod__(10)
 
+# as_integer_ratio
+# TODO uncomment the following tests once #1705 lands (or when the CPython version in test pipeline is upgraded to 3.8)
+# assert (42).as_integer_ratio() == (42, 1)
+# assert (-17).as_integer_ratio() == (-17, 1)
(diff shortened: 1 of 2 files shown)


minhnhdo:
It looks like the failed runs are due to `int.as_integer_ratio` not being available in CPython 3.6. Should I just delete the test?

youknowone:
Thanks for your contributing.
 Upgrading test to use 3.8 will be the best. If it is hard, add an if-statement for python version like:
```python
import sys
if sys.version_info >= (3, 8):
    # run 3.8 test here
```

minhnhdo:
I'll try upgrading the Python version in GitHub workflow first.

palaviv:
> Upgrading test to use 3.8 will be the best. If it is hard, add an if-statement for python version like:

@youknowone have already tried upgrading to 3.8 in #1705. I think that commenting out the tests and adding a `TODO` will be good enough.

minhnhdo:
@palaviv, sure, I'll modify this PR.

coolreader18:
Thank you for contributing!

PR #1761, opened 2020-02-13
Make _random work on wasm

coolreader18:
Adds the `wasm-bindgen` feature to rand/getrandom, and makes the `os` module optional for the `random` module.


Lib/random.py
```diff
@@ -41,7 +41,10 @@
 from types import MethodType as _MethodType, BuiltinMethodType as _BuiltinMethodType
 from math import log as _log, exp as _exp, pi as _pi, e as _e, ceil as _ceil
 from math import sqrt as _sqrt, acos as _acos, cos as _cos, sin as _sin
-from os import urandom as _urandom
+try:
```
youknowone: Adding a comment about wasm here will be helpful to prevent future confusion.

coolreader18:
I changed it to a function that raises a NotImplementedError, as the urandom documentation says that some OSs might do that as well.

PR #1762, opened 2020-02-14
Delete builtin_pow.py

youknowone:
Because this is practically identical as Lib/test/test_pow.py


palaviv:
I am not sure about removing the snippets yet. I would suggest that we will discuss this in an issue. My main points are:

- We do not run the CPython tests on Windows yet.
- I would like to try and run some of the snippets in the WASM tests. Running the CPython tests in WASM might prove more difficult.

windelbouwman:
I would like to keep the snippets. They might be useful for future python implementations, since the code is less tight coupled the python unittest module, which we did not have in the beginning.

PR #1763, opened 2020-02-14
Add missing exceptions

youknowone:


vm/src/exceptions.rs
```diff
@@ -361,6 +358,7 @@ pub struct ExceptionZoo {
     pub arithmetic_error: PyClassRef,
```
coolreader18: Do you want to reorder this too to be in line with exceptions.txt?

youknowone: I am neutral, but will do if you want

coolreader18: I think it would be good to have a consistent order for them across the 4 places we list them (ExceptionZoo struct definition, putting all of the exceptions into variables, constructing ExceptionZoo, adding them to `builtins`)

youknowone: fixed it

Issue #1659, opened 2020-01-03
MRO computation in RustPython doesn't seem to implement C3 correctly


cfbolz:
## Feature

I was experimenting with RustPython's implementation of multiple inheritance and it doesn't seem to implement the C3 method resolution order algorithm correctly.

Example 1:

```python
class A:
    pass

class B(A):
    pass

class C(A, B):
    pass
```

This should give an error, but doesn't in RustPython.

Example 2:

```python
class A:
    def f(self):
        print("A")

class B(A):
    def f(self):
        super().f()
        print("B")

class C(A):
    def f(self):
        super().f()
        print("C")

class D(B, C):
    def f(self):
        super().f()
        print("D")

D().f()
```

This should print "A, C, B, D" but prints instead "A, B, D", missing C completely in the mro.

## Python Documentation

There seems to be no great description of the C3 algorithm as part of the docs, but it's described as part of the 2.3 release notes here: https://www.python.org/download/releases/2.3/mro/

IdanDor:
This is now fixed and can be closed.

PR #1764, opened 2020-02-14
Fixes to MRO usage and issue #1659

IdanDor:
Fixed issue `MRO computation in RustPython doesn't seem to implement C3 correctly` RustPython#1659.

To my surprise the C3 algorithm was implemented, however I have found a few problems which cause it to appear like it isn't:
- Bad mro usage in super: starting from the current class mro instead of the original class at the index of the current class. This leads to skipping classes gotten from multiple inheritance.
- A minor error in mro c3 implementation - not keeping local precedence ordering.
- The mro c3 implementation also allowed duplicate base classes. (Not from the issue but I noticed this anyway).

To allow the fix in new class creation I added an optional vm argument.
This allows the class creation to correctly throw type errors at mro errors.
Due to classes being necessary to vm creation in the first place, this argument is optional (and should only be skipped during initial type creation) and if there is an error here, it will panic instead of returning it.

Surprisingly this pr fixes only a single test, which had a problem in its mocking attributes not being found when setting `__iter__` at line 1482.

I'm new to this project and rust in general. Any fixes/corrections/advice will be appreciated.


vm/src/obj/objsuper.rs
```diff
@@ -53,27 +54,30 @@ impl PySuper {
...
-                    if let Ok(item) = vm.get_attribute(class.as_object().clone(), name.clone()) {
-                        if item.payload_is::<PyBoundMethod>() {
-                            // This is a classmethod
-                            return Ok(item);
-                        }
-                        return vm.call_if_get_descriptor(item, inst.clone());
-                    }
+
+        // Super should be the next matching class in the object original class' mro after the current one.
+        let mut mro_iter = match self.obj_type.payload::<PyClass>() {
```
palaviv: You can use unwrap here instead of the `match` with panic.

vm/src/obj/objsuper.rs
```diff
@@ -53,27 +54,30 @@ impl PySuper {
...
+            Some(PyClass { ref mro, .. }) => mro.iter(),
+            _ => panic!("not Class"),
+        };
+        // The type itself is not in its mro, so skip finding the current class if its the type.
+        if !self.typ.is(&self.obj_type) {
```
palaviv: Can we move this validation to the creation of the super?

vm/src/obj/objtype.rs
```diff
@@ -462,19 +470,33 @@ fn take_next_base(mut bases: Vec<Vec<PyClassRef>>) -> Option<(PyClassRef, Vec<Ve
...
 }
 
 fn linearise_mro(mut bases: Vec<Vec<PyClassRef>>) -> Option<Vec<PyClassRef>> {
     vm_trace!("Linearising MRO: {:?}", bases);
+    // Python requires that the class direct bases are kept in the same order.
+    // This is called local precedence ordering.
+    // This means we must verify that for classes A(), B(A) we must reject C(A, B) even though this
+    // algorithm will allow the mro ordering of [C, B, A, object].
+    // To verify this, we make sure non of the direct bases are in the mro of bases after them.
+    for (i, base_mro) in bases.iter().enumerate() {
```
palaviv: Can you add some snippet test for this case?

vm/src/obj/objtype.rs
```diff
@@ -494,9 +516,31 @@ pub fn new(
     _base: PyClassRef,
     bases: Vec<PyClassRef>,
     dict: HashMap<String, PyObjectRef>,
+    optional_vm: Option<&VirtualMachine>,
```
palaviv: Not sure about this... What do you think about doing the unique bases check in the methods that call this instead of here? Maybe this function will return a `Result` instead of a `PyResult`?

vm/src/obj/objtype.rs
```diff
@@ -494,9 +516,31 @@ pub fn new(
...
 ) -> PyResult<PyClassRef> {
+    // Check for duplicates in bases.
+    let mut unique_bases = HashSet::new();
+    for base in bases.iter() {
```
palaviv: `bases.iter().all(move |x| unique_bases.insert(x))`

vm/src/obj/objtype.rs
```diff
@@ -462,19 +470,33 @@ fn take_next_base(mut bases: Vec<Vec<PyClassRef>>) -> Option<(PyClassRef, Vec<Ve
...
+        if item[0].is(&head) {
+            item.remove(0);
         }
-        return Some((head, bases));
     }
-    None
+    Some((head, bases))
 }
 
 fn linearise_mro(mut bases: Vec<Vec<PyClassRef>>) -> Option<Vec<PyClassRef>> {
```
palaviv: This should probably return `Result<Vec<PyClassRef>, String>` so the specific error can be propagated.

vm/src/obj/objtype.rs
```diff
@@ -494,9 +516,31 @@ pub fn new(
...
     dict: HashMap<String, PyObjectRef>,
+    optional_vm: Option<&VirtualMachine>,
 ) -> PyResult<PyClassRef> {
+    // Check for duplicates in bases.
+    let mut unique_bases = HashSet::new();
+    for base in bases.iter() {
+        if !unique_bases.insert(base.get_id()) {
+            let vm = optional_vm
+                .expect("Could not create a class made by internal code due to duplicate bases.");
+            return Err(vm.new_type_error(format!("duplicate base class {}", base.name)));
```
palaviv: Please add a snippet for this case.

IdanDor:
Good suggestions, I shall look into this.
Where should I add the tests? There are quite a few different places (including rust code, at different crates and also at least 2 different python test folders).

vm/src/obj/objsuper.rs
IdanDor: This is more than validation, it also shifts the mro iterator to the correct location ahead. It should probably also be checked in the constructor but it is part of the code here as well.

vm/src/obj/objtype.rs
IdanDor: Returning a result here and unwrapping on caller if he doesn't have a vm is much cleaner. Good idea.

vm/src/obj/objtype.rs
IdanDor: I need the failing class for the error string (so it matches CPython). I could do this twice, first to check if there are duplicates and the second time to find it, but that sounds worse.

vm/src/obj/objsuper.rs
IdanDor: I have decided to simply do the checks at the constructor, calculate everything there, and remember the resulting mro.

coolreader18:
@IdanDor #1766 was merged, do you want to rebase this on the current master?

coolreader18:
Oh, and test code for this would probably best fit in something like `tests/snippets/class.py`

IdanDor:
Yeah, I'll rebase and fix the small conflicts it caused.

coolreader18:
Thanks for contributing!

PR #1766, opened 2020-02-16
Add the typing standard library module, rework tp_new

coolreader18:


tests/snippets/types_snippet.py
```diff
@@ -50,13 +55,17 @@ class D(B, C): pass
...
-    import gc; gc.collect()
-except: # ...while RustPython doesn't have `gc` yet.
-    pass
-
-assert B.__subclasses__() == []
-assert C.__subclasses__() == []
+if gc:
+    # gc sweep is needed here for CPython...
+    gc.collect()
+    # ...while RustPython doesn't have `gc` yet.
```
palaviv: yet...?

coolreader18: Not sure, the comment was there already. I just moved it.

coolreader18:
I think this should be merged first, and I think #1764 implements a subset of the super changes I made, so the changes there might not be necessary? I'm not very knowledgeable on the C3 resolution algorithm, I just based mine off the CPython C code. To be honest, I didn't really know what changes that PR made when I started working on this one; sorry @IdanDor :confounded:

IdanDor:
It's alright, my PR is mostly bug fixes, I agree that this should be merged first.
Even if you missed one of my changes, it would be very easy to add to this code.

vm/src/vm.rs
```diff
@@ -234,8 +234,10 @@ impl VirtualMachine {
                 #[cfg(not(target_arch = "wasm32"))]
                 import::import_builtin(self, "signal").expect("Couldn't initialize signal module");
 
-                import::init_importlib(self, initialize_parameter)
-                    .expect("Initialize importlib fail");
+                self.expect_pyresult(
```
palaviv: This should never happen so I am not sure if we need to print the reason to the user. What would he do in that case?

vm/src/vm.rs
```diff
@@ -490,6 +492,20 @@ impl VirtualMachine {
         self.new_exception_msg(import_error, msg)
     }
 
+    // TODO: #[track_caller] when stabilized
+    pub fn unwrap_pyresult<T>(&self, result: PyResult<T>) -> T {
```
palaviv: Where is this used?

vm/src/vm.rs
coolreader18: It helped when I was breaking stuff and debugging; it's moreso printing a message to the developer.

vm/src/vm.rs
coolreader18: I figured it would be good as a public API, since `unwrap()`ing a `PyResult` doesn't actually show any information for the `Err`.

coolreader18:
@palaviv is this ok to merge?

vm/src/vm.rs
palaviv: Maybe this should be active only in verbose mode?

vm/src/vm.rs
palaviv: How about adding `#[cfg(build = "debug")]` and doing regular unwrap on `release`?

palaviv:
Yes. I added a few comment but they are not critical.

vm/src/vm.rs
coolreader18: Ooh, or maybe I could check for the `RUST_BACKTRACE` environment variable and only print out the Python backtrace there? I'm not sure about the cfg, because I usually test in release mode because it's so much faster.

PR #1772, opened 2020-02-22
Set fixes

palaviv:


vm/src/pyhash.rs
```diff
@@ -84,6 +84,19 @@ pub fn hash_iter<'a, I: std::iter::Iterator<Item = &'a PyObjectRef>>(
     Ok(hasher.finish() as PyHash)
 }
 
+pub fn hash_iter_no_order<I: std::iter::Iterator<Item = PyObjectRef>>(
```
youknowone: name suggestion: `hash_unordered_iter`

coolreader18: or `hash_iter_unordered`

PR #1774, opened 2020-02-24
Change the name of the _os module to posix/nt

coolreader18:


Lib/os.py
```diff
@@ -43,21 +43,50 @@ def _get_exports_list(module):
...
+    import ntpath as path
+
+    import nt
+    __all__.extend(_get_exports_list(nt))
+    del nt
+
+    try:
+        from nt import _have_functions
+    except ImportError:
+        pass
```
jamestwebber, line 86: Should this order be reversed, given that nt will be deleted?

Issue #1770, opened 2020-02-21
f-string interpolation expressions don't properly parse when there is a `!` in them


coolreader18:
## Feature

```py
f"1 =/= 2: {1 != 2}"
```
Expected:
```py
"1 =/= 2: True"
```
Actual:
```py
SyntaxError: 
f"1 =/= 2: {1 != 2}"
  ↑
  Got error in f-string: Invalid conversion flag
```

meain:
Related:

```
print(f"quoted: {!r 'cat'} non-quoted: {!s 'cat'}")
```

This fails in `cpython` with error as below, but works in `RustPython`.
```
  File "t.py", line 1
    print(f"quoted: {!r 'cat'} non-quoted: {!s 'cat'}")
         ^
SyntaxError: f-string: empty expression not allowed
```

PR #1776, opened 2020-02-26
Update f-string errors, handle != in f-string.

philippeitis:
This pull request addresses #1770.

It correctly handles != inside of f strings, as well as f"{!a}" not returning an EmptyExpression error. 

This also adds an error for conversion flags not being followed by brace characters, which would otherwise be a token error, unlike the Python implementation.

It also changes the print out for UnopenedRbrace / UnclosedLbrace to correctly print braces instead of brackets.

I've also included tests for expected brace characters and not equals operations inside of braces, as well as the expected EmptyExpression errors for f"{!a}".

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,13 @@
+FROM gitpod/workspace-full
+
+USER gitpod
+
+# Update Rust to the latest version
+RUN rm -rf ~/.rustup && ~/.cargo/bin/rustup update stable
+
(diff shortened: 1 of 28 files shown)


coolreader18:
It looks like something weird happened with the second to last commit 4bebc19e5b1c2f87d5906bf05120cc7ec3680d7f, do you think you'd be able to try undoing and redoing the changes so that git doesn't rewrite the entire file?

philippeitis:
I seem to have done something wrong - I'll open this pull request one more time.

PR #1777, opened 2020-02-27
Update f-string errors, handle != in f-string.

philippeitis:
Reopened the pull request due to unusual git behaviour. 

This pull request addresses #1770.

It correctly handles != inside of f strings, as well as f"{!a}" not returning an EmptyExpression error.

This also adds an error for conversion flags not being followed by brace characters, which would otherwise be a token error, unlike the Python implementation.

It also changes the print out for UnopenedRbrace / UnclosedLbrace to correctly print braces instead of brackets.

I've also included tests for expected brace characters and not equals operations inside of braces, as well as the expected EmptyExpression errors for f"{!a}".

### DIFF
--- a/parser/src/fstring.rs
+++ b/parser/src/fstring.rs
@@ -29,39 +29,25 @@ impl<'a> FStringParser<'a> {
 
         while let Some(ch) = self.chars.next() {
             match ch {
-                '!' if delims.is_empty() => {
-                    let x = self.chars.next();
-                    if let Some('=') = x {
-                        expression.push(ch);
-                        expression.push('=');
-                    } else {
-                        if expression.is_empty() {
-                            return Err(EmptyExpression);
-                        }
+                '!' if delims.is_empty() && self.chars.peek() != Some(&'=') => {
+                    if expression.trim().is_empty() {
+                        return Err(EmptyExpression);
+                    }
 
-                        conversion = Some(match x {
-                            Some('s') => ConversionFlag::Str,
-                            Some('a') => ConversionFlag::Ascii,
-                            Some('r') => ConversionFlag::Repr,
-                            Some(_) => {
-                                return Err(InvalidConversionFlag);
-                            }
-                            None => {
-                                break;
-                            }
-                        });
-
-                        match self.chars.peek() {
-                            Some('}') => {
-                                continue;
-                            }
-                            Some(_) => {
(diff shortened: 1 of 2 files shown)


parser/src/fstring.rs
```diff
@@ -30,17 +30,39 @@ impl<'a> FStringParser<'a> {
...
+                        conversion = Some(match x {
+                            Some('s') => ConversionFlag::Str,
+                            Some('a') => ConversionFlag::Ascii,
+                            Some('r') => ConversionFlag::Repr,
+                            Some(_) => {
+                                return Err(InvalidConversionFlag);
+                            }
+                            None => {
+                                break;
+                            }
```
coolreader18: It seems like `None` after the conversion character still yields an invalid conversion flag error:
```py
>>> f"{a!}"
  File "<stdin>", line 1
SyntaxError: f-string: invalid conversion character: expected 's', 'r', or 'a'
```
You could do `Some(_) | None => {...}`

parser/src/fstring.rs
```diff
@@ -30,17 +30,39 @@ impl<'a> FStringParser<'a> {
...
+                        match self.chars.peek() {
+                            Some('}') => {
+                                continue;
+                            }
+                            Some(_) => {
+                                return Err(ExpectedRbrace);
+                            }
+                            None => {
+                                break;
+                            }
```
coolreader18: Same kinda thing here with `None`:
```py
>>> f"{a!r"
  File "<stdin>", line 1
SyntaxError: f-string: expecting '}'
```

parser/src/fstring.rs
```diff
@@ -30,17 +30,39 @@ impl<'a> FStringParser<'a> {
         while let Some(ch) = self.chars.next() {
             match ch {
                 '!' if delims.is_empty() => {
```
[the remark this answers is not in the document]
philippeitis: Good suggestion - this helped make the code a lot cleaner.

parser/src/fstring.rs
philippeitis: This isn't an instance of the None case, which would be f"{a!", although I did add tests for both cases.

parser/src/fstring.rs
philippeitis: Turns out this could be simplified to just self.chars.peek() != Some(&'}' {}

PR #1778, opened 2020-02-27
non-int indexable base for int()

youknowone:


vm/src/obj/objint.rs
```diff
@@ -673,11 +673,16 @@ impl IntOptions {
                         "int() can't convert non-string with explicit base".to_owned(),
                     ));
                 }
-                base
+                if let Some(i) = base.payload_if_subclass::<PyInt>(vm) {
+                    i.as_bigint().clone()
+                } else {
+                    let base = vm.call_method(&base, "__index__", vec![])?;
+                    get_value(&base).clone()
+                }
```
coolreader18: Maybe this could be a function on `vm`? I think it's used in other places as well, at least it is in `list.__getitem__`.

youknowone: I created a small function but not sure what's the ideal form. Probably adding slot function of this magic method will be helpful to make things clear in future.

coolreader18: Oh, I was thinking that you could include the `payload_if_subclass::<PyInt>` inside the VM method, as a fast path

youknowone: Please check if this is ok. Any error message handling suggestion better than Option<PyResult> will be apprecidated.

coolreader18: Looks good!

Issue #1750, opened 2020-02-06
list panic when size > isize::MAX


andrew-ld:
## code
`[...] * (2 << 61)`

## result
`thread 'main' panicked at 'capacity overflow', src/liballoc/raw_vec.rs:777:5`

## expected result
an `MemoryError` exception

PR #1779, opened 2020-02-27
Handle MemoryError for list and strings.

philippeitis:
This aims to address #1750 by raising a Memory error for excessively large allocations. It adds a MAX_MEMORY_SIZE const to VM which is then used by objlist.rs and objstr.rs when doing imul or mul operations.

### DIFF
--- a/vm/src/obj/objlist.rs
+++ b/vm/src/obj/objlist.rs
@@ -19,7 +19,7 @@ use crate::pyobject::{
     PyObjectRef, PyRef, PyResult, PyValue, TryFromObject, TypeProtocol,
 };
 use crate::sequence::{self, SimpleSeq};
-use crate::vm::{ReprGuard, VirtualMachine};
+use crate::vm::{ReprGuard, VirtualMachine, MAX_MEMORY_SIZE};
 
 /// Built-in mutable sequence.
 ///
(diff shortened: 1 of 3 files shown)


vm/src/obj/objlist.rs
```diff
@@ -468,25 +470,31 @@ impl PyList {
...
     }
 
     #[pymethod(name = "__imul__")]
-    fn imul(zelf: PyRef<Self>, counter: isize) -> PyRef<Self> {
-        let new_elements = sequence::seq_mul(&zelf.borrow_sequence(), counter)
-            .cloned()
-            .collect();
-        zelf.elements.replace(new_elements);
-        zelf
+    fn imul(zelf: PyRef<Self>, counter: isize) -> PyResult<Self> {
```
philippeitis: I'm not sure how to handle this function, since it also needs to throw a MemoryError.

coolreader18: You can add a `vm: &VirtualMachine` parameter at the end of the signature to get the `vm`.

coolreader18:
I'm not sure if this is the correct approach; `2**61` bytes is 2.3 exabytes, which probably wouldn't get hit before an actual OOM error occurs in the process. [There is](https://github.com/rust-lang/rfcs/pull/2116) a RFC that allows for catching OOMs in a few different ways (`try_reserve -> Result<(), ...>` functions on `Vec` or `String`, `oom=panic` rustc option that will panic and unwind instead of just aborting) but none of them are stable yet. I'm not sure how to approach this problem, honestly.

philippeitis:
That is true, but this is also a reasonable sanity check, since a list with 8 byte pointers can only contain up to usize::MAX / 8 elements. Beyond that, allocating becomes the virtual machine's problem and will need to dealt with accordingly.

auvipy (changes requested):
it would be great if you could rebase it for further review.

Issue #1780, opened 2020-02-28
Panic on incorrect kwargs instead of TypeError


philippeitis:
Code:
```
print(**("123",))
```
Expected:
```
TypeError: print() argument after ** must be a mapping, not tuple
```
Got:
```
RuntimeError: panicked at 'Kwargs must be a dict.: [PyObj tuple]', src/libcore/result.rs:1165:5
```

BolunThompson:
With #1781 merged, shouldn't this be closed?

coolreader18:
Yep, thanks for noticing!

PR #1781, opened 2020-02-28
Raise TypeError when kwargs are not dict.

philippeitis:
This fixes a panic when a function receives kwargs which are not a dict: addresses #1780


coolreader18:
Thanks for contributing!

Issue #1760, opened 2020-02-12
[RFC] Need to be able to compile only rustpython's virtual machine to wasm


Joetib:
### RustPython in the web can be one of it's best use cases
The python community have been pushing to have python run in the browser.
Many such implementations have been brython, skulpt, Batavia etc.
Likened with batavia's approach to run ```python bytecode ``` in the browser, I have been wondering if rustpytgon can ship only it's virtual machine to do similar task

## Detailed Explanation
Rustpython's  wasm module currently is about 4.5mb. thus size is too big to be inculcated in most projects. 
However I believe if we could remove the parser, lexer and other compiler related stuff and compile only the virtual machine, the size will drastically reduce.
This approach will also require that apps get compiled on their maybe machines before the bytecode is shipped.

## Drawbacks, Rationale, and Alternative

# Rationale:
 It will take a long time for RustPython to get to compete with cpython natively(where cpython works well already) but the web is one place where competition is less. If this works, we can have web apps fully written in python(little to no javascript) 
Be able to run already existing software on the web.
This can also lead to the revolution of building front end livraries. Like implementing python alternatives to react, angular, Vue etc. 
Currently, the beeware team have been working hard to have the Toga(beewae's gui library) working in browsers but are limited by the size of Batavia.
I may certain everyone loved Anvil.works for their idea of creating web apps completely in python. 

# Possible Drawback: 
Since apps will need to be compiled before shipped and most people use cpython, rustpython's vm must be able to run cpython generated bytecode 
This may not even be an issue but I raise this because I'm unaware of the bytecode implementation.

coolreader18:
This is a really nice idea; we've experimented with excluding the parser and compiler for native executables of rustpython, but not yet for wasm. We don't currently run CPython bytecode; we have our own bytecode format and I think it would be less effort to have people install the rustpython compiler as part of whatever toolchain this would be than it would be to convert CPython's bytecode to RustPython's. Also, I could see bundling 2 different versions of rustpython into the `rustpython_wasm` npm package; one with the compiler (`import "rustpython_wasm"`) and one with only the vm (`import "rustpython_wasm/vm-only"`).

Joetib:
Well I think the proposed fix with regards to cpython bytecode (thus having to install RustPython for the toolchain) is great and won't be difficult implementing.

[...]

PR #1784, opened 2020-02-29
Add frozen executable example

palaviv:
This relates to the discussion in #1760. This adds an example of how to create a frozen executable without the complier and parser in the VM. @joetib I hope this will help you with the WASM.


freeze/src/main.rs
```diff
@@ -0,0 +1,24 @@
+use std::collections::HashMap;
+
+use rustpython_vm as vm;
+
+fn main() -> vm::pyobject::PyResult<()> {
+    let vm = vm::VirtualMachine::new(vm::PySettings::default());
+
```
coolreader18: Does this need to be its own crate? Can it be a file in `examples/`?

vm/src/builtins.rs
```diff
@@ -139,27 +143,28 @@ fn builtin_compile(args: CompileArgs, vm: &VirtualMachine) -> PyResult {
         .flags
         .map_or(Ok(0), |v| i32::try_from_object(vm, v.into_object()))?;
 
-    if (flags & ast::PY_COMPILE_FLAG_AST_ONLY).is_zero() {
-        #[cfg(feature = "rustpython-compiler")]
-        {
+    #[cfg(feature = "rustpython-parser")]
```
coolreader18: Wouldn't `feature = "rustpython-parser"` be guaranteed here, because of the cfg on the whole function?

freeze/src/main.rs
palaviv: The important part in the example is `default-features = false`. Not sure how to do that in the `examples/`...

vm/src/builtins.rs
palaviv: The question here is what happens if someone has a `compiler` but not the `parser`. Not sure if that is possible but maybe the error message is not good.

freeze/src/main.rs
coolreader18: What about moving the whole directory into `examples`? It feels weird to have an example crate as part of the workspace...

PR #1785, opened 2020-02-29
Implement input()

### LINKED ISSUE #1586: Implement input()

coolreader18:
Closes #1586, closes #1590, closes #1632

### DIFF
--- a/vm/src/readline.rs
+++ b/vm/src/readline.rs
@@ -1,8 +1,6 @@
 use std::io;
 use std::path::Path;
 
-use rustpython_vm::{scope::Scope, VirtualMachine};
-
 type OtherError = Box<dyn std::error::Error>;
 type OtherResult<T> = Result<T, OtherError>;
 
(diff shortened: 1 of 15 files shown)


vm/src/readline.rs
```diff
@@ -19,13 +17,16 @@ pub enum ReadlineResult {
 mod basic_readline {
     use super::*;
 
-    pub struct BasicReadline<'vm> {
-        vm: &'vm VirtualMachine,
+    pub trait Helper {}
```
palaviv: Maybe some more specific name then `Helper`?

coolreader18: This trait plays a similar role to `rustyline::Helper`, and would be referenced as `readline::Helper`, which I think is pretty specific. Something more specific might be `ReadlineHelper`, but that is kinda redundant because of the `readline::` module prefix.

PR #1786, opened 2020-02-29
Add more CPython tests

palaviv:
- test_builtin
- test_exceptions

### DIFF
--- a/Lib/test/test_builtin.py
+++ b/Lib/test/test_builtin.py
@@ -1,0 +1,1988 @@
+# Python test set -- built-in functions
+
+import ast
+import builtins
+import collections
+import decimal
+import fractions
+import io
+import locale
+import os
+import pickle
+import platform
+import random
+import re
+import sys
+import traceback
+import types
+import unittest
+import warnings
+from contextlib import ExitStack
+from operator import neg
+from test.support import (
+    EnvironmentVarGuard, TESTFN, check_warnings, swap_attr, unlink)
+from test.support.script_helper import assert_python_ok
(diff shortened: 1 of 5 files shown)


Lib/test/test_builtin.py
```diff
@@ -0,0 +1,2044 @@
+# Python test set -- built-in functions
+
+import ast
+import builtins
+import collections
+# import decimal XXX RustPython
```
coolreader18: Decimal was added recently in #1789, you could probably uncomment this if you want.

vm/src/obj/objcoroinner.rs
```diff
@@ -53,9 +72,9 @@ impl Coro {
             return Err(exceptions::normalize(exc_type, exc_val, exc_tb, vm)?);
         }
         vm.frames.borrow_mut().push(self.frame.clone());
+        let result =
+            self.run_with_context(|| self.frame.gen_throw(vm, exc_type, exc_val, exc_tb), vm);
         self.running.set(true);
```
coolreader18: You probably meant to remove this?
```suggestion
```

vm/src/obj/objcoroinner.rs
```diff
@@ -29,15 +31,32 @@ impl Coro {
         }
     }
 
+    fn run_with_context<F>(&self, func: F, vm: &VirtualMachine) -> PyResult<ExecutionResult>
+    where
+        F: FnOnce() -> PyResult<ExecutionResult>,
```
coolreader18: Good idea to reduce duplication!

Issue #1755, opened 2020-02-09
Positional only parameters PEP 570


youknowone:
## Feature

Positional-only parameters

## Python Documentation

https://www.python.org/dev/peps/pep-0570/

palaviv:
@youknowone I am not sure about the purpose of the issue? Do you wish to discuss if we should implement it? (I am pretty sure the answer is yes as it is part of python). Are you looking for help with implementing this?

youknowone:
I am sorry, it must be compatibility issue.

zetwhite:
Is there something left to do in this issue? 
I'm interested in it, but it looks almost everything is done.

youknowone:
@zetwhite by the comment on #1787, positional only error is not implemented. would you check `Lib/test/test_positional_only_arg.py`

zetwhite:
okay. I'll check it. 
thank you 😄

youknowone:
done by #3420

PR #1787, opened 2020-02-29
PEP570 - positional only parameter support

youknowone:
Partially add #1755 

This PR allows positional only parameter functions work but not disallow positional only parameter is passed by name.

### DIFF
--- a/Lib/test/test_grammar.py
+++ b/Lib/test/test_grammar.py
@@ -1,0 +1,1793 @@
+# Python test set -- part 1, grammar.
+# This just tests whether the parser accepts them all.
+
+from test.support import check_syntax_error, check_syntax_warning
+import inspect
+import unittest
+import sys
+import warnings
+# testing import *
+from sys import *
+
+# different import patterns to check that __annotations__ does not interfere
+# with import machinery
+import test.ann_module as ann_module
+import typing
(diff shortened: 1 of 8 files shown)


palaviv:
@youknowone it seems like there is some problem with the exceptions stack in generators. I think that maybe for now we should skip the tests that cause the raise failure.

palaviv:
I would prefer if you would mark the tests with the exception in the generator with `skip` instead of the `raise` tests as the error is actually there. Our problem is that generator exception state leak into the calling frame.

youknowone:
@palaviv I am sorry but I don't catch what you meant. would you help me to add a correct fix?

palaviv:
I think we should just keep what you did here. I tried doing what I suggested in #1786 and it did not work...

palaviv:
> I think we should just keep what you did here. I tried doing what I suggested in #1786 and it did not work...

I have pushed a fix to this problem to #1786. Please rebase this branch after #1786 get merged and that should fix the `raise` errors.

youknowone:
Oh, if i made confusion, I am not going to implement positional only error right now.

coolreader18:
Oh yeah, that's what I meant, if you'd like to implement the errors separately that sounds good to me.

PR #1788, opened 2020-03-01
Skip failing tests in CI

palaviv:
Should fix the CI failure in #1787 and #1786.


palaviv:
Closing after further investigation it seems that the exceptions stack was leaking.

PR #1792, opened 2020-03-03
Fix struct module and add test from unittest

youknowone:


vm/src/stdlib/pystruct.rs
```diff
@@ -102,20 +123,31 @@ impl FormatSpec {
...
+        let mut rdr = Cursor::new(data);
         let mut items = vec![];
         for code in &self.codes {
             debug!("unpack code: {:?}", code);
-            let item = match self.endianness {
-                Endianness::Little => unpack_code::<byteorder::LittleEndian>(vm, &code, &mut rdr)?,
-                Endianness::Big => unpack_code::<byteorder::BigEndian>(vm, &code, &mut rdr)?,
+            match self.endianness {
+                Endianness::Little => {
+                    unpack_code::<byteorder::LittleEndian>(vm, &code, &mut rdr, &mut items)?
```
coolreader18: Could you do the same thing here that you did below? :

```rust
let unpack = match thing { ... };
unpack(...)
```

vm/src/stdlib/pystruct.rs
```diff
@@ -252,171 +279,317 @@ fn pack_bool(vm: &VirtualMachine, arg: &PyObjectRef, data: &mut dyn Write) -> Py
...
+    Ok(())
 }
 
-fn struct_calcsize(fmt: PyStringRef, vm: &VirtualMachine) -> PyResult<usize> {
-    let fmt_str = fmt.as_str();
-    let format_spec = FormatSpec::parse(fmt_str).map_err(|e| vm.new_value_error(e))?;
+fn struct_calcsize(fmt: Either<PyStringRef, PyBytesRef>, vm: &VirtualMachine) -> PyResult<usize> {
+    let parsed = match fmt {
+        Either::A(string) => FormatSpec::parse(string.as_str()),
+        Either::B(bytes) => FormatSpec::parse(std::str::from_utf8(&bytes).unwrap()),
```
coolreader18: It looks like CPython primarily parses on bytes, and encodes `str`s to bytes using the ascii codec:
```python
>>> struct.calcsize(">BȔ")
UnicodeEncodeError: 'ascii' codec can't encode character '\u0214' in position 2: ordinal not in range(128)
```

youknowone: I added a comment about this, but didn't fix it at this time.

PR #1797, opened 2020-03-06
Add the msvcrt stdlib module

coolreader18:
`getpass` works on windows now.

### DIFF
--- a/vm/src/stdlib/msvcrt.rs
+++ b/vm/src/stdlib/msvcrt.rs
@@ -1,0 +1,58 @@
+use crate::obj::objbytes::PyBytesRef;
+use crate::obj::objstr::PyStringRef;
+use crate::pyobject::{PyObjectRef, PyResult};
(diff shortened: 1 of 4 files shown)


vm/src/stdlib/msvcrt.rs
```diff
@@ -0,0 +1,61 @@
+use crate::obj::objbytes::PyBytesRef;
+use crate::obj::objstr::PyStringRef;
+use crate::pyobject::{PyObjectRef, PyResult};
+use crate::VirtualMachine;
+// use ::suppress_iph;
```
palaviv: ?

coolreader18:
@palaviv I don't know if I'd be able to add tests, as all of the functions that I've added interact with the console that they're a part of; even the cpython tests for getpass don't test the win_getpass implementation: https://github.com/python/cpython/blob/master/Lib/test/test_getpass.py.

PR #1801, opened 2020-03-07
Implement async generators

coolreader18:


vm/src/obj/objasyncgenerator.rs
```diff
@@ -0,0 +1,401 @@
...
+
+        let res = self.ag.inner.throw(
+            exc_type,
+            exc_val.unwrap_or_else(|| vm.get_none()),
+            exc_tb.unwrap_or_else(|| vm.get_none()),
+            vm,
+        );
+        let res = PyAsyncGenWrappedValue::unbox(&self.ag, res, vm);
+        if res.is_err() {
+            self.state.set(AwaitableState::Closed);
```
palaviv: Maybe just call `self.close()` here?

coolreader18:
I'm not sure why travis is failing, my only guess might be that it has a smaller stack size than the other runners but I don't know why this change would trigger that.

coolreader18:
Hmmm, that worked, but I think the problem is really Travis for having such a small stack size. @palaviv what do you think about this?

palaviv:
I am in favor of removing Travis. It seems like Github actions is working great. This should probably be done in a different PR so we can see if there are any objections. In addition there are a few builds that have not moved to Github actions yet.

palaviv:
Can we merge this @coolreader18?

coolreader18:
Yup! I think this is good.

PR #1802, opened 2020-03-07
Fix str.{capitalize,split,rsplit} + test_unicode.py from CPython 3.8.2

youknowone:


vm/src/obj/objstr.rs
```diff
@@ -407,8 +407,16 @@ impl PyString {
...
     fn capitalize(&self) -> String {
-        let (first_part, lower_str) = self.value.split_at(1);
-        format!("{}{}", first_part.to_uppercase(), lower_str)
+        let mut char_indices = self.value.char_indices();
+        if let Some((_, first_char)) = char_indices.next() {
+            format!(
+                "{}{}",
+                first_char.to_uppercase(),
+                &self.value[first_char.len_utf8()..].to_lowercase(),
+            )
```
coolreader18: You could just use `chars()` for this:
```suggestion
        let mut chars = self.value.chars();
        if let Some(first_char) = chars.next() {
            format!(
                "{}{}",
                first_char.to_uppercase(),
                &chars.as_str().to_lowercase(),
            )
```

coolreader18 (approved):

PR #1803, opened 2020-03-08
Fix subprocess.wait to take None as a timeout argument

youknowone:


Lib/subprocess.py
```diff
@@ -230,7 +230,7 @@ def call(*popenargs, timeout=None, **kwargs):
     """
     with Popen(*popenargs, **kwargs) as p:
         try:
-            return p.wait(timeout=timeout)
+            return p.wait(timeout)
```
youknowone: is there an easy to add keyword argument support?

coolreader18: Other than a FromArgs derive on a struct, I don't think so.

vm/src/stdlib/subprocess.rs
```diff
@@ -142,8 +142,8 @@ impl PopenRef {
         self.process.borrow().exit_status()
     }
 
-    fn wait(self, timeout: OptionalArg<u64>, vm: &VirtualMachine) -> PyResult<()> {
-        let timeout = match timeout.into_option() {
+    fn wait(self, timeout: OptionalArg<Option<u64>>, vm: &VirtualMachine) -> PyResult<()> {
```
coolreader18: Could you add a `PopenWaitArgs` struct so that the `timeout=` kwarg works?

coolreader18 (changes requested):

vm/src/stdlib/subprocess.rs
youknowone: done

coolreader18 (approved):

PR #1806, opened 2020-03-08
Fix str.zfill

youknowone:


vm/src/obj/objstr.rs
```diff
@@ -993,7 +993,17 @@ impl PyString {
...
         } else {
-            format!("{}{}", "0".repeat(len - value.len()), value)
+            let first = value.bytes().next();
+            let (sign, s) = match first {
+                Some(b'+') | Some(b'-') => (
+                    std::str::from_utf8(&[first.unwrap(); 1])
+                        .unwrap()
+                        .to_owned(),
+                    &value[1..],
+                ),
```
```suggestion
            let mut chars = value.chars();
            let (sign, s) = match chars.next() {
                Some(sign @ '+') | Some(sign @ '-') => (sign.to_string(), chars.as_str()),
```

youknowone: Generally this changes looks a lot better. One question. If the sign always must be matched to `+` or `-`, isn't `chars()` an unnessessary utf-8 decode overhead?

coolreader18: Perhaps, but it might be optimized away anyway, and I think that it's better to err on the safe side of keeping everything UTF-8 rather than risk matching on the middle of a multi-byte sequence or something like that. This could be done with bytes though:
```suggestion
            let mut bytes = value.bytes();
            let (sign, s) = match bytes.next() {
                Some(sign @ b'+') | Some(sign @ b'-') => ((sign as char).to_string(), &value[1..]),
```
It just seems like a micro-optimization to me, but if you prefer bytes feel free to do it.

PR #1811, opened 2020-03-12
Make PyObject.typ a Rc<PyObject<PyClass>>

coolreader18:
The main motivator for this is to make the type hierarchy creation simpler and less error prone; this removes the trait object hack, including a few questionable transmutes.

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -571,7 +571,7 @@ impl PyContext {
 
     pub fn new_base_object(&self, class: PyClassRef, dict: Option<PyDictRef>) -> PyObjectRef {
         PyObject {
-            typ: class,
+            typ: class.into_generic_pyobj(),
             dict: dict.map(RefCell::new),
             payload: objobject::PyBaseObject,
         }
@@ -628,7 +628,7 @@ pub struct PyObject<T>
 where
     T: ?Sized + PyObjectPayload,
 {
-    pub typ: PyClassRef,
+    pub typ: Rc<PyObject<PyClass>>,
     pub dict: Option<RefCell<PyDictRef>>, // __dict__ member
     pub payload: T,
 }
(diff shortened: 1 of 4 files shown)


coolreader18:
Whoops, I forgot about this; I think it was actually fine when I first opened it.

vm/src/pyobject.rs
```diff
@@ -648,6 +648,21 @@ impl PyObject<dyn PyObjectPayload> {
             Err(self)
         }
     }
+
+    /// Dowcast this PyObjectRef to an `Rc<PyObject<T>>`. The [`downcast`](#method.downcast) method
```
youknowone, line 652:
```
    /// Downcast this PyObjectRef to an `Rc<PyObject<T>>`. The [`downcast`](#method.downcast) method
```

coolreader18: Whoops

coolreader18:
Yeah; `PyRef<T>` is really just a wrapper around a `PyObjectRef`, which has a trait object as a bound and is therefore a ["fat pointer"](https://doc.rust-lang.org/reference/dynamically-sized-types.html), with a vtable pointer. In order to properly initialize a `PyClassRef`, then, we need to provide it with a vtable pointer to the correct vtable, which is pretty unsafe and requires some hacks to get the vpointer in the first place. With `Rc<PyObject<PyClass>>` (instead of `Rc<PyObject<dyn PyObjectPayload>>`), the inner type is just a plain struct, `PyClass`, and so it doesn't require any vtable pointers or anything like that.

vm/src/pyobject.rs
```diff
@@ -648,6 +648,21 @@ impl PyObject<dyn PyObjectPayload> {
...
+    /// therefore can be used similarly to an `&T`.
+    pub fn downcast_generic<T: PyObjectPayload>(
+        self: Rc<Self>,
+    ) -> Result<Rc<PyObject<T>>, PyObjectRef> {
+        if self.payload_is::<T>() {
+            let ptr = Rc::into_raw(self) as *const PyObject<T>;
+            let ret = unsafe { Rc::from_raw(ptr) };
+            Ok(ret)
+        } else {
+            Err(self)
```
youknowone, line 663: is there benefits to return `self` for `Err`?

vm/src/pyobject.rs
```diff
@@ -705,6 +721,10 @@ impl<T: PyValue> PyRef<T> {
     pub fn typ(&self) -> PyClassRef {
         self.obj.class()
     }
+
+    pub fn into_generic_pyobj(self) -> Rc<PyObject<T>> {
```
youknowone, line 725: I don't have a good substitution but `generic_pyobj` gives me upcasting feeling to me.

vm/src/pyobject.rs
coolreader18: It's the same pattern as `PyObject::downcast()` and `Any::downcast()`; it allows the caller to get the object back if the downcasting failed.

vm/src/pyobject.rs
coolreader18: Ah, I see what you mean; "generic" as sort of general/untyped. Maybe `into_typed_pyobj()`?

youknowone: that sounds more clear to me

PR #1816, opened 2020-03-14
 Error on positional-only parameters passed as keywords

coolreader18:


vm/src/obj/objfunction.rs
```diff
@@ -17,6 +17,8 @@ use crate::scope::Scope;
 use crate::slots::{SlotCall, SlotDescriptor};
 use crate::vm::VirtualMachine;
```
youknowone, line 19: How about not making additional new line here? `rustfmt` is specially formatting self, super and crate so this is still the same order without the new line.

PR #1819, opened 2020-03-15
Remove Travis and Azure CI

coolreader18:
We've been using Github Actions for ~2 months now, and from my perspective at least it's been a much better experience than either Travis or Azure Pipelines. In #1700, the plan was that we make sure Actions works well, and than transition to it from Travis, and it also makes Azure's CI redundant, as Actions can run on Linux, Mac, and Windows. We still haven't transferred the cron job over that checks Redox compilation and uploads code coverage data to codecov, so I added it in this PR; a successful run of it be seen [here](https://github.com/RustPython/RustPython/actions/runs/56151068). In the codecov job, it needs to compile in debug mode, so that debug symbols don't get elided and data can be accurately collected, but that also meant that it was running into the same problem on the infinite recursion `RecursionError` test, so I lowered the recursion limit when `cfg!(debug_assertions)` is true.


coolreader18:
Yes, there isn't a limit for public, open source repositories.

PR #1821, opened 2020-03-16
Use _pyio.py as the base for our _io implementation

coolreader18:
Meant to open as draft, whoops...


palaviv:
#1868 should fix the deadlock problem but after fixing that I get into a infinite loop when enabling the trace.

coolreader18:
Actually, I don't think this is worth it, I'm going to look into just improving the rust _io module.

PR #1823, opened 2020-03-21
Add the ssl stdlib module

coolreader18:
This is all VERY flaky, with many uses of unsafe. I'm not sure if this is a good idea at this point, but it does work, ~~albeit not certificate verification~~. EDIT: certificate verification does work, just not with `google.com`. @palaviv what's the deal? :smile:  

```py
from urlllib.request import urlopen
print(urlopen("https://rustpython.github.io").read().decode())
```

### DIFF
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -101,10 +101,11 @@ libz-sys = "1.0"
 
 [target.'cfg(windows)'.dependencies]
 winreg = "0.7"
+schannel = "0.1"
 
 [target."cfg(windows)".dependencies.winapi]
 version = "0.3"
-features = ["winsock2", "handleapi", "ws2def", "std", "winbase"]
+features = ["winsock2", "handleapi", "ws2def", "std", "winbase", "wincrypt"]
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 wasm-bindgen = "0.2"
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -70,7 +70,7 @@ paste = "0.1"
 base64 = "0.11"
 is-macro = "0.1"
 result-like = "^0.2.1"
-foreign-types-shared = "0.1"
+foreign-types = "0.3"
 num_enum = "0.4"
 
 flame = { version = "0.2", optional = true }
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -70,6 +70,8 @@ paste = "0.1"
 base64 = "0.11"
 is-macro = "0.1"
 result-like = "^0.2.1"
+foreign-types-shared = "0.1"
+num_enum = "0.4"
 
 flame = { version = "0.2", optional = true }
 flamer = { version = "0.3", optional = true }
@@ -89,6 +91,8 @@ subprocess = "0.2.2"
 num_cpus = "1"
 socket2 = { version = "0.3", features = ["unix"] }
 rustyline = "6.0"
+openssl = "0.10"
+openssl-sys = "0.9"
 
 [target.'cfg(not(any(target_arch = "wasm32", target_os = "redox")))'.dependencies]
 dns-lookup = "1.0"
--- a/vm/src/obj/objset.rs
+++ b/vm/src/obj/objset.rs
@@ -631,6 +631,17 @@ macro_rules! multi_args_frozenset {
 
 #[pyimpl(flags(BASETYPE))]
 impl PyFrozenSet {
+    pub fn from_iter(
+        vm: &VirtualMachine,
+        it: impl IntoIterator<Item = PyObjectRef>,
+    ) -> PyResult<Self> {
+        let mut inner = PySetInner::default();
+        for elem in it {
+            inner.add(&elem, vm)?;
+        }
+        Ok(Self { inner })
+    }
+
     #[pyslot]
     fn tp_new(
         cls: PyClassRef,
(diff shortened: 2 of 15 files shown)


coolreader18:
Okay, I think this is in half-decent shape now. If anyone has experience with openssl, I'd love to hear any ideas about why certificate verification isn't working.

coolreader18:
Nevermind, certificate verification does work, just not with `google.com` for some reason.

Lib/test/test_bytes.py
```diff
@@ -253,6 +253,7 @@ def test_compare(self):
         self.assertFalse(b3 <  b2)
         self.assertFalse(b3 <= b2)
 
+    @unittest.expectedFailure
```
palaviv: Why is this failing now?

vm/Cargo.toml
```diff
@@ -89,6 +91,8 @@ subprocess = "0.2.2"
 num_cpus = "1"
 socket2 = { version = "0.3", features = ["unix"] }
 rustyline = "6.0"
+openssl = "0.10"
```
palaviv: Have we considered something pure rust? Maybe [RusTLS](https://github.com/ctz/rustls)?

vm/src/obj/objset.rs
```diff
@@ -67,11 +67,21 @@ impl PyValue for PyFrozenSet {
 }
 
 #[derive(Default, Clone)]
-struct PySetInner {
+pub struct PySetInner {
```
palaviv: Do we realy need to expose this? This is used only once to create a `frozenset` (unless I missed something)

vm/Cargo.toml
coolreader18: I looked into it, but `_ssl`'s API pretty directly corresponds to OpenSSL's API; I don't think it would be guaranteed that everything we needed would be present in rustls.

vm/src/obj/objset.rs
coolreader18: I'm not sure, I guess I could just have some sort of `from_iter` function for PyFrozenSet.

vm/src/stdlib/ssl.rs
```diff
@@ -0,0 +1,877 @@
...
+
+        PySslContext {
+            ctx: RefCell::new(builder),
+            check_hostname,
+        }
+        .into_ref_with_type(vm, cls)
+    }
+
+    #[pymethod]
+    fn set_ciphers(&self, cipherlist: CString, vm: &VirtualMachine) -> PyResult<()> {
```
youknowone: is there some advantage by using `CString` here instead of `String`?

coolreader18: Oh, it'll return a value error if there's a nul character in the string, but the openssl set_ciphers implementation internally makes a new cstring and panics if it's not valid. I'll add a comment about it.

vm/src/pyobject.rs
```diff
@@ -919,6 +919,13 @@ pub struct PyIterable<T = PyObjectRef> {
 }
 
 impl<T> PyIterable<T> {
+    pub fn from_method(method: PyObjectRef) -> Self {
```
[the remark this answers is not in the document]
coolreader18: I think so, this isn't the main way that `PyIterable` is intended to be created, and I think `from_method` is more descriptive than just `new: fn(PyObjectRef) -> Self`

Lib/test/test_bytes.py
coolreader18: Ah, it was because I allowed `PyByteInner::try_from_object` to accept an arbitrary iterator, and bytes.__eq__ used that. I modified `PyByteInner::cmp` to use `PyBytesLike` instead.

vm/src/stdlib/ssl.rs
coolreader18: I changed it to just manually check for a '\0' in a PyStringRef.

coolreader18:
Is this good to merge?

vm/src/stdlib/ssl.rs
```diff
@@ -0,0 +1,879 @@
...
+            ctx: RefCell::new(builder),
+            check_hostname,
+        }
+        .into_ref_with_type(vm, cls)
+    }
+
+    #[pymethod]
+    fn set_ciphers(&self, cipherlist: PyStringRef, vm: &VirtualMachine) -> PyResult<()> {
+        let ciphers = cipherlist.as_str();
+        if ciphers.contains('0') {
```
youknowone: prorably `ciphers.contains(0 as char)`?

coolreader18: 😳 yep, I meant `'\0'`

coolreader18:
Also, would it be best to vendor openssl by using the `vendored` feature for rust-openssl? It would increase binary size, but it would also get rid of a lot of issues with linking to openssl (I don't think any of the weird Windows stuff would be necessary). Maybe best to only vendor on windows, if possible?

vm/src/stdlib/ssl.rs
youknowone: having a test for this method call will be good to ensure this is correctly working

youknowone:
I have no idea about `rust-openssl` part but this PR looks good to me

coolreader18:
I think I'll look into vendoring-only-on-windows by contributing something to rust-openssl, I don't think it's possible otherwise, and it works for now.

palaviv:
> Also, would it be best to vendor openssl by using the `vendored` feature for rust-openssl? It would increase binary size, but it would also get rid of a lot of issues with linking to openssl (I don't think any of the weird Windows stuff would be necessary). Maybe best to only vendor on windows, if possible?

This fails on my PC without the `vendored` feature. I would suggest adding this by default to avoid problems with different systems and versions of OpenSSL.

PR #1824, opened 2020-03-21
Escape octet in string

palaviv:
This actually fixes the current failures in test_bytes. In addition I simplified lexer by merging `lex_byte` into `lex_string`.

### DIFF
--- a/parser/src/lexer.rs
+++ b/parser/src/lexer.rs
@@ -462,6 +462,19 @@ where
         }
     }
 
+    fn parse_octet(&mut self, first: char) -> char {
+        let mut octet_content = String::new();
+        octet_content.push(first);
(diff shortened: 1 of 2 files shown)


parser/src/lexer.rs
```diff
@@ -557,6 +571,11 @@ where
                                 error: LexicalErrorType::StringError,
                                 location: self.get_pos(),
                             });
+                        } else if is_bytes && !c.is_ascii() {
```
coolreader18: It looks like clippy wants you to merge this condition with the one above: `if c == '\n' && !triple_quoted || is_bytes && !c.is_ascii()`. (It'll probably ask you to make it single `else if` after that, too, rather than an `else { if ... { } }`

PR #1827, opened 2020-03-23
Adding readme under parser and parser/src

mireille-raad:
Adding readme files under parser and parser/src to make it faster to figure out the files and what each does.


coolreader18:
LGTM

parser/src/README.md
```diff
@@ -0,0 +1,34 @@
+# RustPython/parser/src
```
coolreader18: Maybe this could go in a section in `parser/README.md`? Similar to how we have the project directory structure in the root `README.md`.

mireille-raad: I wasn't sure which one made sense. I'll fix and update the PR in a bit.

PR #1832, opened 2020-03-30
pymodule derive macro

youknowone:
Creating an inner module for every stdlib looks not ideal but i have no idea how to do without them

### DIFF
--- a/derive/src/util.rs
+++ b/derive/src/util.rs
@@ -1,3 +1,223 @@
-pub fn path_eq(path: &syn::Path, s: &str) -> bool {
+use super::Diagnostic;
+use std::collections::HashMap;
+use syn::{AttributeArgs, Ident, Lit, Meta, NestedMeta, Path, Signature};
+
+pub fn path_eq(path: &Path, s: &str) -> bool {
     path.get_ident().map_or(false, |id| id == s)
+}
+
+pub fn def_to_name(
+    ident: &Ident,
+    attr_name: &'static str,
+    attr: AttributeArgs,
+) -> Result<String, Diagnostic> {
+    let mut name = None;
+    for attr in attr {
+        if let NestedMeta::Meta(meta) = attr {
+            if let Meta::NameValue(name_value) = meta {
+                if path_eq(&name_value.path, "name") {
+                    if let Lit::Str(s) = name_value.lit {
+                        name = Some(s.value());
+                    } else {
+                        bail_span!(
+                            name_value.lit,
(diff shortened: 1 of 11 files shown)


derive/src/util.rs
```diff
@@ -1,3 +1,216 @@
...
+    ident: &Ident,
+    attr_name: &'static str,
+    attr: AttributeArgs,
+) -> Result<String, Diagnostic> {
+    let mut name = None;
+    for attr in attr {
+        if let NestedMeta::Meta(meta) = attr {
+            if let Meta::NameValue(name_value) = meta {
+                if path_eq(&name_value.path, "name") {
+                    if let Lit::Str(s) = name_value.lit {
```
coolreader18, line 19: Should we do any `magic` name formatting here, or is that sorted out on the name collected here later on?

coolreader18: Oh, nevermind, this is just copied from `pyclass.rs`

coolreader18:
Maybe we could only do this for some modules until `custom_inner_attributes` is stabilized? I think rightward drift might be an issue, as would `git blame` results. It'd be great if we could make rustfmt format it like below, ignoring indentation for just that block, though I don't think that's possible:
```rust
#[pymodule]
mod decl {

#[pyfunction]
fn thing() {}

}
```

youknowone:
For the formatting, I couldn't find any way to change local indent hint if we don't want to disable the whole file formatting. Also derive macro seems still always require to match the brace pair.

Is it possible to add `#[pymodule]` to `use` part of `mod.rs` for same result? I don't know well but it doesn't look like - because item iteration looks only possible on actual module code.

youknowone:
The mentioned issue looks quite old and then it means it is not expected to be solved in near future. Just go by experiencing git blame barrier?

coolreader18:
Yeah, I guess so; I haven't gotten a ton of use out of git blame for this project, and I think the stdlib is generally simple/encapsulated enough that it wouldn't be used that much there.

palaviv:
@youknowone I think this change make developing harder. When I have errors the compiler does not show the line numbers and add suggestion:
```rust
error[E0308]: mismatched types
  |
  = note: expected reference `&std::sync::Arc<pyobject::PyObject<(dyn pyobject::PyObjectPayload + 'static)>>`
                found struct `std::sync::Arc<pyobject::PyObject<dyn pyobject::PyObjectPayload>>`
```
Is there a way to fix this?

youknowone:
I am sorry that discovering this comment so late. should we consider to revert this?

Issue #1671, opened 2020-01-10
Increase CPython test coverage


palaviv:
We are currently running part of the CPython test suite. This is a good place for people new to the project to contribute. Contribution here can come in two ways:

- Fix failing test - Look for tests marked as skipped or xfail and fix RustPython so that the test will pass.
- Add more tests suites - Copy test suite from CPython and mark failing tests with xfail or skip.

In order to run the test suite use `cargo run -- -m test -v`.
All tests are in `Lib/test`

abitrolly:
@palaviv is it possible to generate and online dashboard that shows which tests pass and which need more help?

palaviv:
I am sure it is possible... Would you like to create such a dashboard? I think you can use [testgrid](https://github.com/GoogleCloudPlatform/testgrid) for that.
My recommendation would be to run the test suite as described above and pick any of the `exprectedFailure` or skipped test marked as RustPython change. I am not sure a dashboard would be effective here.

abitrolly:
@palaviv is it possible to get the report in a file in machine readable format, such as https://testanything.org ?

palaviv:
The test use `test\libregtest` and run the underline unittests tests. I am not sure if there is a builtin support in either for the format you want...

mireille-raad:
Would it help having the dashboard somewhere on the current website?

If you are looking for a dashboard, we can do two pages:
- one for tests passed/failed status
- another for the "what is left" script.

The goal would be for people to know what they can help on in one place.

I can probably get the output of both scripts to push to github repos and then use that as part of jekyll page or just as something people can read.

coolreader18:
I think that would be cool, and it would probably require getting the test results in a machine readable format like TAP. It seems like [tappy](https://github.com/python-tap/tappy) works perfectly fine on RustPython, the only issue is getting regrtest to run a custom unittest test runner.

abitrolly:
@mireille-raad is it be possible to have single dashboard for both passed/failed and "what is left" by color coding them with green, red and yellow?

coolreader18:
I don't know if it makes sense to have them in the same dashboard, since the regrtests are tests that get run and are well structured, while the whats_left script just shows standard library items that aren't present in RustPython; they don't really show the same kind of data.

abitrolly:
@coolreader18 `regrtests` and `standard library items` are different test sets?

coolreader18:
[...]

PR #1834, opened 2020-04-02
 Add json output for CPython tests

coolreader18:
Related to #1671

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1534,6 +1534,7 @@ version = "0.1.1"
 dependencies = [
  "bincode",
  "bitflags",
+ "itertools",
  "lz4-compress",
  "num-bigint",
  "num-complex",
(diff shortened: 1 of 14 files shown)


ep12:
See #1838 (example section):
This pr implements the code object `co_varnames` attribute, but the value differs from the ones cpython and pypy present...

abitrolly:
Does it produce the result in real-time, or only after the processing is finished?

Is it a custom format, or something like https://chromium.googlesource.com/chromium/src.git/+/master/docs/testing/json_test_results_format.md was used for an inspiration?

coolreader18:
@abitrolly I'm not sure what you mean by your first question, but as to your second I think it's basically just the data that `unittest` outputs itself, converted to JSON. I didn't write the runner myself, I modified it from https://github.com/agramian/custom-text-test-runner.

coolreader18:
@ep12 I'll look into fixing co_varnames separately.

PR #1836, opened 2020-04-02
Don't allocate for PyNativeFunc when it's just a fn pointer

coolreader18:


youknowone:
This is an interesting crate.

Issue #1831, opened 2020-03-28
[RFC] Threading


palaviv:
## Summary

I think we have got to the stage we should support threading in RustPython.

## Detailed Explanation

There was already some discussion on threading but I wanted to start a dedicated issue as it is a big decision. I will try to break the issue in order to make some order:

### Definition of done
We need a user to be able to create a new thread in python and execute code on the same Python object from multiple threads:
```py
import threading

def worker(num):
    """thread worker function"""
    print 'Worker: %s' % num
    return

threads = []
for i in range(5):
    t = threading.Thread(target=worker, args=(i,))
    threads.append(t)
    t.start()
```

### GIL
I think that one of the biggest discussion is the [GIL](https://wiki.python.org/moin/GlobalInterpreterLock). Should we add a GIL to RustPython? Should we allow only one thread to use each object at a time?

### Suggested approach
I suggest the following changes in order to reach the ability to spawn a new thread:

- I suggest we will create a new VirtualMachine for each thread. Thus making VirtualMachine `!sync` and `!Send`.
- Use `Arc` instead of `Rc` in `PyObjectRef`. `pub type PyObjectRef = Arc<PyObject<dyn PyObjectPayload>>`
- `PyValue` and `PyObjectPayload` traits will implement `Sync` and `Send`. Thus forcing us to make all Py* structs sync as well.
- We will need to convert most of the code `Rc`, `Cell` and `RefCell` use to thread safe options. For example `Arc`, `Mutex` and `Atomic*`.
- When accessing data of an Object wrapped in `Mutex` we will lock the mutex for our use. This will require careful handling when using internal objects to avoid deadlocks when used between threads. 

A simple example of the `start_new_thread` method in `_thread` will look something like:
```rust
fn start_new_thread(func: PyFunctionRef, args: PyFuncArgs, vm: &VirtualMachine) -> u64 {
    let handle = thread::spawn(move || {
        let thread_vm = VirtualMachine::new(PySettings::default()); // Should get some params from original VM
        thread_vm.invoke(func.as_object(), args).unwrap();
    });
    get_id(handle.thread())
}
```
## Drawbacks, Rationale, and Alternatives

- I suggest that we will ignore the GIL and allow multiple threads to execute python code simultaneously. This can be a big advantage of RustPython on CPython. The user will be expected to make his own code thread safe. This might prove problematic to code that rely on the GIL.

- What about third party library's? We do not yet have an API but would we force them to implement `Sync` and `Send` as well?

[...]

PR #1837, opened 2020-04-02
Prep some things for threading

coolreader18:
Related to #1831

### DIFF
--- a/vm/src/obj/objiter.rs
+++ b/vm/src/obj/objiter.rs
@@ -2,8 +2,8 @@
  * Various types to support iteration.
  */
 
+use crossbeam_utils::atomic::AtomicCell;
 use num_traits::{Signed, ToPrimitive};
-use std::cell::Cell;
 
 use super::objint::PyInt;
 use super::objsequence;
@@ -28,12 +28,9 @@ pub fn get_iter(vm: &VirtualMachine, iter_target: &PyObjectRef) -> PyResult {
         vm.get_method_or_type_error(iter_target.clone(), "__getitem__", || {
             format!("Cannot iterate over {}", iter_target.class().name)
         })?;
-        let obj_iterator = PySequenceIterator {
-            position: Cell::new(0),
-            obj: iter_target.clone(),
-            reversed: false,
-        };
-        Ok(obj_iterator.into_ref(vm).into_object())
+        Ok(PySequenceIterator::new_forward(iter_target.clone())
+            .into_ref(vm)
+            .into_object())
     }
 }
 
@@ -140,7 +137,7 @@ pub fn length_hint(vm: &VirtualMachine, iter: PyObjectRef) -> PyResult<Option<us
 #[pyclass]
 #[derive(Debug)]
 pub struct PySequenceIterator {
-    pub position: Cell<isize>,
+    pub position: AtomicCell<isize>,
     pub obj: PyObjectRef,
     pub reversed: bool,
 }
@@ -153,14 +150,31 @@ impl PyValue for PySequenceIterator {
 
 #[pyimpl]
 impl PySequenceIterator {
+    pub fn new_forward(obj: PyObjectRef) -> Self {
+        Self {
+            position: AtomicCell::new(0),
+            obj,
+            reversed: false,
+        }
+    }
+
+    pub fn new_reversed(obj: PyObjectRef, len: isize) -> Self {
+        Self {
+            position: AtomicCell::new(len - 1),
+            obj,
+            reversed: true,
+        }
+    }
+
     #[pymethod(name = "__next__")]
     fn next(&self, vm: &VirtualMachine) -> PyResult {
-        if self.position.get() >= 0 {
+        let pos = self.position.load();
+        if pos >= 0 {
             let step: isize = if self.reversed { -1 } else { 1 };
-            let number = vm.ctx.new_int(self.position.get());
+            let number = vm.ctx.new_int(pos);
             match vm.call_method(&self.obj, "__getitem__", vec![number]) {
                 Ok(val) => {
-                    self.position.set(self.position.get() + step);
+                    self.position.store(pos + step);
                     Ok(val)
                 }
                 Err(ref e) if objtype::isinstance(&e, &vm.ctx.exceptions.index_error) => {
@@ -181,7 +195,7 @@ impl PySequenceIterator {
 
     #[pymethod(name = "__length_hint__")]
     fn length_hint(&self, vm: &VirtualMachine) -> PyResult<isize> {
-        let pos = self.position.get();
+        let pos = self.position.load();
         let hint = if self.reversed {
             pos + 1
         } else {
@@ -195,11 +209,7 @@ impl PySequenceIterator {
 }
 
 pub fn seq_iter_method(obj: PyObjectRef) -> PySequenceIterator {
-    PySequenceIterator {
-        position: Cell::new(0),
-        obj,
-        reversed: false,
-    }
+    PySequenceIterator::new_forward(obj)
 }
 
 #[pyclass(name = "callable_iterator")]
--- a/vm/src/obj/objiter.rs
+++ b/vm/src/obj/objiter.rs
@@ -168,20 +168,15 @@ impl PySequenceIterator {
 
     #[pymethod(name = "__next__")]
     fn next(&self, vm: &VirtualMachine) -> PyResult {
-        let pos = self.position.load();
+        let step: isize = if self.reversed { -1 } else { 1 };
+        let pos = self.position.fetch_add(step);
         if pos >= 0 {
-            let step: isize = if self.reversed { -1 } else { 1 };
-            let number = vm.ctx.new_int(pos);
-            match vm.call_method(&self.obj, "__getitem__", vec![number]) {
-                Ok(val) => {
-                    self.position.store(pos + step);
-                    Ok(val)
-                }
+            match vm.call_method(&self.obj, "__getitem__", vec![vm.new_int(pos)]) {
                 Err(ref e) if objtype::isinstance(&e, &vm.ctx.exceptions.index_error) => {
                     Err(new_stop_iteration(vm))
                 }
                 // also catches stop_iteration => stop_iteration
-                Err(e) => Err(e),
+                ret => ret,
             }
         } else {
             Err(new_stop_iteration(vm))
@@ -217,7 +212,7 @@ pub fn seq_iter_method(obj: PyObjectRef) -> PySequenceIterator {
 pub struct PyCallableIterator {
     callable: PyCallable,
     sentinel: PyObjectRef,
-    done: Cell<bool>,
+    done: AtomicCell<bool>,
 }
 
 impl PyValue for PyCallableIterator {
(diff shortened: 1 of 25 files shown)


coolreader18:
Looks like `crates.io` is acting up.

vm/src/obj/objiter.rs
```diff
@@ -139,7 +136,7 @@ pub fn length_hint(vm: &VirtualMachine, iter: PyObjectRef) -> PyResult<Option<us
 #[pyclass]
 #[derive(Debug)]
 pub struct PySequenceIterator {
-    pub position: Cell<isize>,
+    pub position: AtomicCell<isize>,
```
palaviv, line 139: Why do you prefer the `AtomicCell` and not using `std::sync::atomic::AtomicIsize`?

vm/src/obj/objiter.rs
```diff
@@ -152,14 +149,31 @@ impl PyValue for PySequenceIterator {
...
-        if self.position.get() >= 0 {
+        let pos = self.position.load();
+        if pos >= 0 {
             let step: isize = if self.reversed { -1 } else { 1 };
-            let number = vm.ctx.new_int(self.position.get());
+            let number = vm.ctx.new_int(pos);
             match vm.call_method(&self.obj, "__getitem__", vec![number]) {
                 Ok(val) => {
-                    self.position.set(self.position.get() + step);
+                    self.position.store(pos + step);
```
palaviv, line 176: You should not use the store here as this will not be atomic. Another thread can get the old value of `position` and the increment will be wrong. My suggestion is the change the `self.position.load` to `self.position.fetch_add(step)`

vm/src/obj/objlist.rs
```diff
@@ -858,10 +860,11 @@ impl PyValue for PyListIterator {
 impl PyListIterator {
     #[pymethod(name = "__next__")]
     fn next(&self, vm: &VirtualMachine) -> PyResult {
-        if self.position.get() < self.list.elements.borrow().len() {
-            let ret = self.list.elements.borrow()[self.position.get()].clone();
-            self.position.set(self.position.get() + 1);
-            Ok(ret)
+        let list = self.list.elements.borrow();
+        let pos = self.position.load();
```
palaviv, line 864: Again. Use `fetch_add`

vm/src/obj/objlist.rs
```diff
@@ -895,10 +900,12 @@ impl PyValue for PyListReverseIterator {
 impl PyListReverseIterator {
     #[pymethod(name = "__next__")]
     fn next(&self, vm: &VirtualMachine) -> PyResult {
-        if self.position.get() > 0 {
-            let position: usize = self.position.get() - 1;
-            let ret = self.list.elements.borrow()[position].clone();
-            self.position.set(position);
+        let pos = self.position.load();
```
palaviv, line 903: Again. Use `fetch_add`

vm/src/obj/objstr.rs
```diff
@@ -111,14 +111,13 @@ impl PyValue for PyStringIterator {
 impl PyStringIterator {
     #[pymethod(name = "__next__")]
     fn next(&self, vm: &VirtualMachine) -> PyResult {
-        let pos = self.byte_position.get();
+        let pos = self.byte_position.load();
```
palaviv, line 114: Again. Use `fetch_add`

vm/src/obj/objstr.rs
```diff
@@ -149,13 +148,12 @@ impl PyValue for PyStringReverseIterator {
 impl PyStringReverseIterator {
     #[pymethod(name = "__next__")]
     fn next(&self, vm: &VirtualMachine) -> PyResult {
-        if self.position.get() > 0 {
-            let position: usize = self.position.get() - 1;
+        let pos = self.position.load();
```
palaviv, line 151: Again. Use `fetch_add`

vm/src/obj/objstr.rs
```diff
@@ -312,11 +310,11 @@ impl PyString {
 
     #[pymethod(name = "__hash__")]
     fn hash(&self) -> pyhash::PyHash {
-        match self.hash.get() {
+        match self.hash.load() {
```
palaviv, line 313: Here you should probably use `compare_and_swap`

vm/src/obj/objiter.rs
coolreader18: The first one I did was objstr, which had a non-promitive type `Option<i64>`, and I kept using it because it seemed more consistent; you know that it has the same functionality for the most part because it's the same base type with a different generic? I'm not sure, if you're really opposed to it I can change it, but as far as I can tell it's a zero-cost wrapper over the standard Atomic types, except with generics instead of completely separate types.

vm/src/obj/objstr.rs
coolreader18: This is a reverse iterator, though, and fetch_sub (and fetch_add) wrap on over/underflow, and doing a fetch_op before the if means it's unconditional, so if you call `next(it)`  after it does a StopIteration than it would go to usize::max and be out of bounds.

vm/src/obj/objiter.rs
coolreader18: How about loading it first, and then doing a `compare_exchange` with `prev ± 1`?

vm/src/obj/objstr.rs
coolreader18: I think it's fine, since there's not really any stake to the value here; if 2 threads calculate and set it simultaneously than it'll be fine, and for `compare_and_swap` we have to provide a new value anyway, which defeats the purpose of a hash cache.

vm/src/obj/objiter.rs
coolreader18: @palaviv I think the change in the latest commit fixes this, could you take a look?

vm/src/obj/objiter.rs
palaviv: Sounds good to me. Just wanted to understand the reason.

Lib/logging/__init__.py
```diff
@@ -1410,7 +1410,9 @@ def error(self, msg, *args, **kwargs):
         logger.error("Houston, we have a %s", "major problem", exc_info=1)
         """
         if self.isEnabledFor(ERROR):
+            print("hmm")
```
palaviv, line 1413: ?

vm/src/obj/objiter.rs
```diff
@@ -152,25 +149,44 @@ impl PyValue for PySequenceIterator {
...
-                    Err(new_stop_iteration(vm))
+        let mut prev = self.position.load();
+        loop {
+            if prev < 0 {
+                break;
+            }
+            let step = if self.reversed { -1 } else { 1 };
+            match self.position.compare_exchange(prev, prev + step) {
+                Ok(new) => {
+                    match vm.call_method(&self.obj, "__getitem__", vec![vm.new_int(number)]) {
```
palaviv, line 178: Probably `vm.new_int(new)`

vm/src/obj/objiter.rs
```diff
@@ -152,25 +149,44 @@ impl PyValue for PySequenceIterator {
...
-            let step: isize = if self.reversed { -1 } else { 1 };
-            let number = vm.ctx.new_int(self.position.get());
-            match vm.call_method(&self.obj, "__getitem__", vec![number]) {
-                Ok(val) => {
-                    self.position.set(self.position.get() + step);
-                    Ok(val)
-                }
-                Err(ref e) if objtype::isinstance(&e, &vm.ctx.exceptions.index_error) => {
-                    Err(new_stop_iteration(vm))
+        let mut prev = self.position.load();
```
palaviv, line 170: I might be missing something but I think we can go for a simpler solution here:
```rust
let step = if self.reversed { -1 } else { 1 };
let curr_pos = self.position.fetch_add(step);
if curr_pos >= 0 {
...
```
then use the same code as before.

This will answer what we expect from an iterator. Each time next is called we will receive a unique item from the iterable. Once the iterator is exhausted all consecutive calls will get stop iteration.

vm/src/obj/objtuple.rs
```diff
@@ -257,10 +257,10 @@ impl PyValue for PyTupleIterator {
 impl PyTupleIterator {
     #[pymethod(name = "__next__")]
     fn next(&self, vm: &VirtualMachine) -> PyResult {
-        if self.position.get() < self.tuple.as_slice().len() {
-            let ret = self.tuple.as_slice()[self.position.get()].clone();
-            self.position.set(self.position.get() + 1);
-            Ok(ret)
+        let pos = self.position.load();
```
palaviv, line 260: Again lets use `fetch_add` here. In this solution we can have 2 thread getting the same value.

vm/src/util.rs
```diff
@@ -11,3 +12,20 @@ pub fn read_file(path: &Path) -> Result<String> {
 
     Ok(buffer)
 }
+
+pub(crate) fn atomic_iter_advance(
+    pos: &AtomicCell<usize>,
+    done: impl Fn(usize) -> bool,
+    next: impl Fn(usize) -> usize,
+) -> Option<usize> {
+    let mut prev = pos.load();
```
palaviv, line 21: Please see comment in `objiter`

Lib/logging/__init__.py
coolreader18: Whoops, leftover debugging.

vm/src/obj/objiter.rs
coolreader18: Ah, that probably would work since it's an isize and it wouldn't matter if it underflowed.

vm/src/util.rs
coolreader18: I made a comment on your last set of reviews; if it's a reverse iterator and it's going backward, then `fetch_sub` would underflow and the index would be `usize::MAX_VALUE`. What would you suggest for that?

coolreader18: Maybe a separate `done` flag?

palaviv: I do prefer to keep this simple as the current implementation is a little complex. I think we an go in a few directions here:
1. Change the position to `isize`.
2. Start the position at `len(iterator)` instead of `len(iterator) - 1` and go until 0.
3. Check if we are at 0 before doing the `fetch_sub`.

coolreader18: Oh, right, it'd be fine if we do a load and then a fetch_sub after we check that we're still in bounds, since it's still atomic. Okay, thanks for your feedback.

palaviv: Looking at this again I am not sure  2 or 3 will be simple at all. I think changing to `isize` will be the simplest. You can't actually do `load` and then `fetch_sub` as it will not be atomic.

coolreader18: Okay, I updated it to just use the `fetch_*` operations. I didn't really want to do that, but I think I was just being silly, because yes, if someone does this:
```py
it = iter([1,2])
try:
    next(it)
except:
    pass
```
Then it's eventually going to wrap around and start yielding elements again, but that's a very small edge case that doesn't really matter in the grand scheme of things.

vm/src/obj/objlist.rs
```diff
@@ -895,11 +899,10 @@ impl PyValue for PyListReverseIterator {
...
     #[pymethod(name = "__next__")]
     fn next(&self, vm: &VirtualMachine) -> PyResult {
-        if self.position.get() > 0 {
-            let position: usize = self.position.get() - 1;
-            let ret = self.list.elements.borrow()[position].clone();
-            self.position.set(position);
-            Ok(ret)
+        let list = self.list.elements.borrow();
+        let pos = self.position.fetch_sub(1) as usize - 1;
+        if let Some(ret) = list.get(pos as usize - 1) {
```
palaviv, line 904: Is `pos` not `usize` at this stage? Are you sure it should not be `list.get(pos)`?

coolreader18: Oh, whoops, I meant to just do that once.

vm/src/obj/objlist.rs
```diff
@@ -912,7 +915,7 @@ impl PyListReverseIterator {
 
     #[pymethod(name = "__length_hint__")]
     fn length_hint(&self) -> usize {
-        self.position.get()
+        self.position.load() as usize
```
palaviv, line 918: I think this need to be `self.position.load() as usize - 1`

PR #1839, opened 2020-04-03
Replace the json module with the Python implementation from CPython

coolreader18:
This would give us proper streaming parsing/generation support (`json.{dump,load}`), the `JSONEncoder` and `JSONDecoder` classes, and in general better compatibility with the full API of the `json` module.


vm/src/stdlib/json.rs
```diff
@@ -42,7 +42,7 @@ pub fn json_loads(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult {
         let module = vm
             .get_attribute(vm.sys_module.clone(), "modules")
             .unwrap()
-            .get_item("json", vm)
+            .get_item("_serde_json", vm)
```
youknowone, line 45: do we include this version into binary, but never using?

coolreader18: Yeah, I suppose there really isn't a point to keeping it around; most of the functionality is in `py_serde.rs` anyway.

youknowone: Not sure about the future - but If we are going to *incubate* rust native libraries, probably we can grow up the rust version of library from this one. Is this not related to `_json` in CPython?

coolreader18: No, `_json` has low-level json operations, e.g. "find where this string ends", "convert this string to JSON", or streaming encoding/lexing classes.

coolreader18: We could find another json library, but serde_json doesn't expose any lexing functionality.

PR #1840, opened 2020-04-04
Add Pathlib

palaviv:
This change add `pathlib` and changes many `os` method to support path like objects.
Many of the tests fail because Rust does not support `Cstring` with `nul`. Not sure how to bypass that. Would appreciate suggestion on how to solve this.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -36,7 +36,7 @@ use crate::obj::objdict::PyDictRef;
 use crate::obj::objint::PyIntRef;
 use crate::obj::objiter;
 use crate::obj::objset::PySet;
-use crate::obj::objstr::{self, PyString, PyStringRef};
+use crate::obj::objstr::{PyString, PyStringRef};
 use crate::obj::objtuple::PyTupleRef;
 use crate::obj::objtype::PyClassRef;
 use crate::pyobject::{
@@ -110,14 +110,14 @@ fn output_by_mode(val: String, mode: OutputMode, vm: &VirtualMachine) -> PyObjec
 }
 
 pub struct PyPathLike {
-    path: String,
+    path: ffi::OsString,
     mode: OutputMode,
 }
 
 impl PyPathLike {
     pub fn new_str(path: String) -> Self {
         PyPathLike {
-            path,
+            path: ffi::OsString::from(path),
             mode: OutputMode::String,
         }
     }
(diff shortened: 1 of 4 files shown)


coolreader18:
I did implement TryFromObject for CString recently, but you could also just have a nulcheck function that makes sure a string doesn't have a nul char in it, and returns a ValueError if it does, then just call `nulcheck(s, vm)?;`. (Or maybe a vm method?)

palaviv:
> I did implement TryFromObject for CString recently, but you could also just have a nulcheck function that makes sure a string doesn't have a nul char in it, and returns a ValueError if it does, then just call `nulcheck(s, vm)?;`. (Or maybe a vm method?)

The problem is that CPython support `nul` char while rust throws exception.
```py
os.stat("/dev/null\x00")
```
Should not throw an error.

coolreader18:
Ah, I see. Hmm, that is tricky.

vm/src/stdlib/os.rs
```diff
@@ -1087,21 +1109,16 @@ fn os_chmod(
...
-        || objtype::issubclass(&path.class(), &vm.ctx.bytes_type())
-    {
-        Ok(path)
-    } else {
-        Err(vm.new_type_error(format!(
-            "expected str or bytes object, not {}",
-            path.class()
-        )))
-    }
+fn os_fspath(path: PyPathLike, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 1112: is there `Err` path in this function?

palaviv:
> It seems only trailing `\x00` exists in tests. Trimming them before making `PyPathLike` would work?

This will actually give the opposite result. As `/dev/null` should be different then `/dev/null\x00`

 > This is not specifically related to this PR, but while reading, i found i have no idea why PyPathLike is not PyPath but Like

Thats how they are called in CPython.

youknowone:
I try to test this problem a bit, but I can't make null-included path working. Could you describe your working environment?

macOS, python 3.8.1, APFS
```
>>> open('abc\x00', 'w')
ValueError: embedded null byte
>>> import os
>>> os.stat('/dev/null\x00')
ValueError: embedded null byte
>>>
```

Linux,  Python 3.5.2, ext4
```
>>> open('abc\x00', 'w')
ValueError: embedded null byte
>>> import os
>>> os.stat('/dev/null\x00')
ValueError: embedded null byte
```
```

palaviv:
If you will look at the tests. For example https://github.com/RustPython/RustPython/pull/1840/files#diff-6bc3eb7fdef7557c945c66fb0b60a134R1921

palaviv:
@coolreader18 @youknowone can this be pushed and we will solve the `nul` later?

youknowone:
It seems `Vec<u8>` is the reasonble internal data of path types. And they may be used to create ffistring in future when they are actually used. Is it possible to fix now? or do you want to merge this for right now as it is?

palaviv:
> It seems `Vec<u8>` is the reasonble internal data of path types. And they may be used to create ffistring in future when they are actually used. Is it possible to fix now? or do you want to merge this for right now as it is?

I am not sure I understand... Would you like to change the `OsString` to `Vec<u8>`?

youknowone:
I guess so. I didn't check cpython implementation well yet, but it seems it is more close to `Vec<u8>` than `OsString`. Or `enum { Bytes(Vec<u8>), Str(String) }` if there are difference by str/bytes.

coolreader18:
@palaviv I think fixing it later would be fine, it's not too critical to pathlib, and I think it would rarely happen in practice.

palaviv:
I have merged this and we can do the proposed changes in the future

youknowone:
I checked about PyPathLike impl but it revealed enum is not fit for the use case.

PR #1841, opened 2020-04-04
Use vendored version of openssl

palaviv:
As discussed in #1823. I would suggest revisiting this once our testing matrix is larger.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1083,6 +1083,15 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


coolreader18:
Yeah, I think this is the best course of action for now. Do you want to try removing the openssl installation scripts from ci?

palaviv:
> Yeah, I think this is the best course of action for now. Do you want to try removing the openssl installation scripts from ci?

Can you do that in a different PR?

PR #1843, opened 2020-04-05
 Fix some clippy lints that were previously 'allow'ed 

coolreader18:


vm/src/vm.rs
```diff
@@ -768,8 +768,7 @@ impl VirtualMachine {
             self.trace_event(TraceEvent::Return)?;
             result
         } else if class.has_attr("__call__") {
-            let result = self.call_method(&callable, "__call__", args);
-            result
+            self.call_method(&callable, "__call__", args)
```
youknowone, line 771: These patterns were actually more debugger-friendly than now.
Do we prefer cleaner code than debugger friendly code?

coolreader18: Oh, like being able to see the value of `result` there? I'm not sure, I don't use traditional debuggers that let you step through, but I'd think that there might be some way to see the value of an expression after it executes? I'm not sure.

youknowone: Yes, exactly that use case. I just know this kind of "single expression for one line" or "single side effect for one line" is a traditional recommendation for debugger-friendly code. Because I am not a heavy user of debuggers, probably there must be experts who know better about this problem.

PR #1846, opened 2020-04-07
Direct print() to sys.stdout

coolreader18:
This allows e.g. `contextlib.redirect_stdout` to work with `print()`

### DIFF
--- a/vm/src/builtins.rs
+++ b/vm/src/builtins.rs
@@ -4,7 +4,6 @@
 
 use std::cell::Cell;
 use std::char;
-use std::io::{self, Write};
 use std::str;
 
 use num_bigint::Sign;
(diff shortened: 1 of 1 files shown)


vm/src/pyobject.rs
```diff
@@ -107,6 +107,7 @@ pub struct PyContext {
     pub exceptions: exceptions::ExceptionZoo,
     pub int_cache_pool: Vec<PyObjectRef>,
     tp_new_wrapper: PyObjectRef,
+    pub(crate) original_stdio: Option<(PyObjectRef, PyObjectRef, PyObjectRef)>,
```
youknowone, line 110: I think python expect to have `sys.__stdin__` and so on has original stdio objects. is this nessessary in spite of them?

youknowone:
I don't think I know enough about this topic. Could anyone review this PR? @palaviv @windelbouwman

vm/src/pyobject.rs
coolreader18: Not really, and this was all just in an effort to reduce the overhead of doing `sys.stdout.write("x")`, but I think it's actually fast enough that it doesn't really matter. I'll just do what CPython does and make file `sys.stdout` if the argument isn't present.

PR #1847, opened 2020-04-08
Fix bytes.replace

youknowone:


vm/src/obj/objbyteinner.rs
```diff
@@ -1096,41 +1096,165 @@ impl PyByteInner {
...
+
+    fn replace_delete(&self, from: PyByteInner, maxcount: Option<usize>) -> Vec<u8> {
+        let count = count_substring(self.elements.as_slice(), from.elements.as_slice(), maxcount);
+        if count == 0 {
+            // no matches
+            return self.elements.clone();
+        }
+
+        let result_len = self.len() - (count * from.len());
+        debug_assert!(self.len() >= count * from.len());
```
palaviv, line 1123: maybe remove this? I know is is only in debug but it does not provide additional data.

vm/src/obj/objbyteinner.rs
```diff
@@ -1096,41 +1096,165 @@ impl PyByteInner {
...
+        to: PyByteInner,
+        maxcount: Option<usize>,
+    ) -> Vec<u8> {
+        let len = from.len();
+        let mut iter = self.elements.find_iter(&from.elements);
+
+        let mut new = if let Some(offset) = iter.next() {
+            let mut new = self.elements.clone();
+            new[offset..offset + len].clone_from_slice(to.elements.as_slice());
+            if maxcount == Some(1) {
```
palaviv, line 1152: The logic here is a little confusing. Maybe we should do the rest of the function in the `else` of this?

vm/src/obj/objbyteinner.rs
```diff
@@ -1096,41 +1096,165 @@ impl PyByteInner {
...
+    ) -> PyResult<Vec<u8>> {
+        let count = count_substring(self.elements.as_slice(), from.elements.as_slice(), maxcount);
+        if count == 0 {
+            // no matches, return unchanged
+            return Ok(self.elements.clone());
+        }
+
+        // Check for overflow
+        //    result_len = self_len + count * (to_len-from_len)
+        debug_assert!(count > 0);
```
palaviv, line 1186: Do we need this?

vm/src/obj/objbyteinner.rs
```diff
@@ -1202,6 +1326,23 @@ pub fn try_as_byte(obj: &PyObjectRef) -> Option<Vec<u8>> {
     })
 }
 
+#[inline]
+fn count_substring(haystack: &[u8], needle: &[u8], maxcount: Option<usize>) -> usize {
+    let substrings = haystack.find_iter(needle);
+    if let Some(maxcount) = maxcount {
+        let mut count = 0;
```
palaviv, line 1333: How about `min(substring.count(), maxcount)`?

vm/src/obj/objbyteinner.rs
youknowone: This is line by line port of CPython implementation. Do you still want to remove this regardless of it?

vm/src/obj/objbyteinner.rs
youknowone: When `substring.count()` is big and `maxcount` is small, the iteration cost will be different by early break.

youknowone: Oh, I used it with `take()` and it works. Thanks!

vm/src/obj/objbyteinner.rs
palaviv: I am OK with you keeping this. Just not sure if that is necessary. Your call.

PR #1852, opened 2020-04-11
Fix {bytes|bytearray}.{|r|l}strip

youknowone:


vm/src/obj/objbyteinner.rs
```diff
@@ -1,9 +1,9 @@
-use std::convert::TryFrom;
-use std::ops::Range;
-
+use bstr::ByteSlice;
```
palaviv, line 1: Where are we using this?

youknowone: Every `trim` related methods are belong to here.
Example: https://github.com/RustPython/RustPython/pull/1852/files#diff-fced6f15bb0f9333ff5c2b5847a8be4aR945

PR #1861, opened 2020-04-11
Fix str.split bytes.split

youknowone:
~Failing test depends on #1855~

### DIFF
--- a/vm/src/obj/objbyteinner.rs
+++ b/vm/src/obj/objbyteinner.rs
@@ -15,6 +15,7 @@ use super::objsequence::{is_valid_slice_arg, PySliceableSequence};
 use super::objslice::PySliceRef;
 use super::objstr::{self, adjust_indices, PyString, PyStringRef, StringRange};
 use super::objtuple::PyTupleRef;
+use super::pystr::PyCommonString;
 use crate::function::{OptionalArg, OptionalOption};
 use crate::pyhash;
 use crate::pyobject::{
@@ -26,6 +27,12 @@ use crate::vm::VirtualMachine;
 #[derive(Debug, Default, Clone)]
 pub struct PyByteInner {
     pub elements: Vec<u8>,
+}
+
+impl From<Vec<u8>> for PyByteInner {
+    fn from(elements: Vec<u8>) -> PyByteInner {
+        Self { elements }
+    }
 }
 
 impl ThreadSafe for PyByteInner {}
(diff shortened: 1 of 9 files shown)


vm/src/obj/objbyteinner.rs
```diff
@@ -311,6 +309,8 @@ impl ByteInnerSplitlinesOptions {
 
 #[allow(clippy::len_without_is_empty)]
 impl PyByteInner {
+    const ASCII_WHITESPACES: [u8; 6] = [0x20, 0x09, 0x0a, 0x0c, 0x0d, 0x0b];
```
palaviv, line 312: Can we use `is_ascii_whitespace`?

vm/src/obj/objbyteinner.rs
```diff
@@ -995,21 +995,89 @@ impl PyByteInner {
...
+                    .collect()
+            }
         } else {
-            Ok(split_slice(&self.elements, &sep, maxsplit))
+            self.split_whitespace(maxsplit)
+        };
+        Ok(elements)
+    }
+
+    fn rsplit_whitespace(&self, maxsplit: isize) -> Vec<Vec<u8>> {
```
palaviv, line 1040: It seems like there is a lot of code duplication of `split_whitespace`. Maybe we can create some sort of macro?

vm/src/obj/objbyteinner.rs
```diff
@@ -995,21 +995,89 @@ impl PyByteInner {
...
+            }
+            haystack = &haystack[..offset];
+        }
+        if !haystack.is_empty() {
+            splited.push(haystack.to_vec());
         }
+        splited
+    }
+
+    pub fn rsplit(
```
palaviv, line 1060: Again maybe we can share code with `split`?

vm/src/obj/objbyteinner.rs
youknowone: find_byteset takes array, slice or vector instead of function. do you have any idea not to define this const?

vm/src/obj/objbyteinner.rs
youknowone: I think this is possible to share but `split_whitespace` looks to require too much effort

youknowone: @palaviv how do you think about this new revision?

vm/src/obj/objbyteinner.rs
youknowone: Unlike split and rsplit, I don't see a good solution for _whitespaces to achieve enough readability. They have similar structure but lots of signs and variables are different.

vm/src/obj/objbyteinner.rs
palaviv: Looks great

PR #1862, opened 2020-04-12
Update cargo dependencies; use new bindings in openssl

coolreader18:


Cargo.lock
```diff
@@ -1069,9 +1020,9 @@ checksum = "2839e79665f131bdb5782e51f2c6c9599c133c6098982a54c794358bf432529c"
 
 [[package]]
 name = "openssl"
-version = "0.10.28"
+version = "0.10.29"
```
youknowone, line 1023: Does new functions work with `0.10.29`+ or any version of `0.10.*`? Do we need to pin this specific version into Cargo.toml?

coolreader18: Only `0.10.29`+, but I think the presence of Cargo.lock should guarantee that it's at least the version specified, if not exactly the same.

PR #1865, opened 2020-04-12
make whats_left.sh generate a yaml file for dashboard

mireille-raad:
Make `whats_left.sh` generate a yaml file that we can use for a dashboard/page. Ref issue #1671

I used yaml format because it can be read by jekyll.
https://jekyllrb.com/docs/datafiles/

use becomes `whats_left.sh > whats_left.yaml`

### DIFF
--- a/whats_left.sh
+++ b/whats_left.sh
@@ -3,14 +3,10 @@ set -e
 
 ALL_SECTIONS=(methods modules)
 
-GREEN='[32m'
-BOLD='[1m'
-NC='(B[m'
-
 h() {
   # uppercase input
-  header_name=$(echo "$@" | tr "[:lower:]" "[:upper:]")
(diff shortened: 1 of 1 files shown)


whats_left.sh
```diff
@@ -3,14 +3,10 @@ set -e
...
 
-GREEN='[32m'
-BOLD='[1m'
-NC='(B[m'
-
 h() {
   # uppercase input
-  header_name=$(echo "$@" | tr "[:lower:]" "[:upper:]")
-  echo "$GREEN$BOLD===== $header_name =====$NC"
+  header_name=$(echo "$@")
```
coolreader18, line 8: Maybe we could have an option for `whats_left.sh` that makes it output yaml, rather than doing it by default?

mireille-raad: sure, i will add an option for -o filename

PR #1869, opened 2020-04-14
Remove unnecessary unic dependencies and add `\N{}` unicode name escapes

coolreader18:
Removing all of the unused unicode tables brought down my binary size from 24M to 18M !


parser/src/lexer.rs
```diff
@@ -475,6 +475,37 @@ where
         u8::from_str_radix(&octet_content, 8).unwrap() as char
     }
 
+    fn parse_unicode_name(&mut self) -> Result<char, LexicalError> {
```
palaviv, line 478: Maybe add a unit test for this?

PR #1871, opened 2020-04-15
Make PyList to ThreadSafe

palaviv:


vm/src/obj/objlist.rs
```diff
@@ -115,8 +113,8 @@ impl PyList {
         &self,
         vm: &VirtualMachine,
     ) -> PyResult<objbyteinner::PyByteInner> {
-        let mut elements = Vec::<u8>::with_capacity(self.get_len());
-        for elem in self.elements.borrow().iter() {
+        let mut elements = Vec::<u8>::with_capacity(self.borrow_elements().len());
+        for elem in self.borrow_elements().clone().iter() {
```
youknowone, line 117: is this cloning of Vec or lock?

palaviv: The `Vec`. `RwLock` does not implement `Clone`

youknowone: it seems additional clone is added to a few places. But this is using the iterator of the vector instead of the vector itself. Is there no way to avoid this kind of clone?

coolreader18: I don't think so, if there is python code running inside the loop body, then it could theoretically lock the pylist again and cause a panic or deadlock. Although that doesn't seem necessary for this one, @palaviv, none of the code inside the loop calls into the interpreter. Is there another reason you cloned the vec here?

vm/src/obj/objlist.rs
```diff
@@ -751,9 +777,9 @@ impl PyList {
         // replace list contents with [] for duration of sort.
         // this prevents keyfunc from messing with the list and makes it easy to
         // check if it tries to append elements to it.
-        let mut elements = self.elements.replace(vec![]);
+        let mut elements = PyList::_replace(self.borrow_elements_mut(), vec![]);
         do_sort(vm, &mut elements, options.key, options.reverse)?;
-        let temp_elements = self.elements.replace(elements);
+        let temp_elements = PyList::_replace(self.borrow_elements_mut(), elements);
```
coolreader18, line 782: I think if you borrow mutably here, you can just use `std::mem::take()` or `std::mem::replace()`. Something like:
```rust
let mut elements = self.borrow_elements_mut();
let mut temp_elements = std::mem::take(&mut elements);
do_sort(vm, &mut temp_elements, options.key, options.reverse)?;
std::mem::swap(&mut temp_elements, &mut elements);

File: vm/src/obj/objlist.rs [hunk 3]
```diff
@@ -484,17 +498,17 @@ impl PyList {
...
+            sequence::seq_mul(&elements, counter).cloned().collect();
+        PyList::_replace(elements, new_elements);
+        zelf.clone()
     }
 
     #[pymethod]
     fn count(&self, needle: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
         let mut count: usize = 0;
-        for element in self.elements.borrow().iter() {
+        for element in self.borrow_elements().clone().iter() {
```
coolreader18, line 511: It'd be really nice if we could somehow detect if something tries to lock while we already have one and then give it a clone, but I think any solution to that would violate rust's aliasing rules.

palaviv: The big question is how expensive is the `clone`. I hope that this is very quick because of the `Arc`.

vm/src/pyobject.rs
```diff
@@ -1222,6 +1222,10 @@ pub trait PyValue: fmt::Debug + Sized + 'static {
 
 // Temporary trait to follow the progress of threading conversion
 pub trait ThreadSafe: Send + Sync {}
+// Temporary definitions to help with converting object that contain PyObjectRef to ThreadSafe.
+// Should be removed before threading is allowed.
```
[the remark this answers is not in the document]
palaviv: I hope the new message is better. This should be a very temporary thing.

vm/src/obj/objlist.rs
palaviv: I removed the `clone` here. I tried to clone only where python code is called.

PR #1872, opened 2020-04-16
Enable the os and io modules on wasi

coolreader18:


coolreader18:
I think we could test in CI, I'll look into it separately, though.

PR #1875, opened 2020-04-18
Make Dict thread safe

palaviv:


vm/src/dictdatatype.rs
```diff
@@ -18,41 +19,84 @@ type HashIndex = pyhash::PyHash;
...
 type EntryIndex = usize;
 
-#[derive(Clone)]
 pub struct Dict<T = PyObjectRef> {
+    inner: RwLock<InnerDict<T>>,
+}
+
+impl ThreadSafe for Dict {}
+
+struct InnerDict<T> {
```
youknowone, line 28: I am a bit curious if PyDict and PySet can have `RwLock<Dict<T>>` instead of this InnerDict structure

coolreader18: I agree, for separation of concerns `Dict` seems like it would work best as a standard-ish data structure that requires a mutable reference to mutate.

palaviv: I have a actually started that way but it became very complex. One of my goals is to not hold the lock while "python" code is running. In the `Dict` lookup we call `__eq__` so we need to release that lock but in order to do that from `PyDict` and `PySet` we will need to split the implementation of `Dict` to smaller methods that do not call "python" code.
Do you guys think this is the way to go? That will make a lot of code duplication in `PyDict` and `PySet`.

coolreader18: Hmm, that is tricky. I don't really think I have a preference either way for mutable/interior mutable Dict.

palaviv: I will push this as other object are dependent on `PyDictRef`. We can always change the implementation in the future.

PR #1876, opened 2020-04-19
shared string strip implemenetation

youknowone:


youknowone:
I tried to use `fn()` but it is not easy to solve lifetime error.

PR #1883, opened 2020-04-24
Share str.splitlines and bytes.splitlines implementaion

youknowone:


vm/src/obj/objstr.rs
```diff
@@ -786,25 +786,33 @@ impl PyString {
...
     #[pymethod]
     fn splitlines(&self, args: pystr::SplitLinesArgs, vm: &VirtualMachine) -> PyObjectRef {
-        let mut elements = vec![];
-        let mut curr = "".to_owned();
-        let mut chars = self.value.chars().peekable();
-        while let Some(ch) = chars.next() {
-            if ch == '\n' || ch == '\r' {
-                if args.keepends {
-                    curr.push(ch);
+        let keep = if args.keepends { 1 } else { 0 };
```
palaviv, line 789: It seems like you have the same code duplicated. Maybe share it?

youknowone: I am struggling with how to share logics around str::chars() and [u8]::iter()

Do you have any idea to do it without macro?

palaviv: Maybe define a generic `get_iterator` in `PyCommonString`?

youknowone: In that case the type will be `dyn std::iter::Iterator` and i am afraid it would have additional runtime overhead

coolreader18: You could have an associated type on `PyCommonString` for the inner character type, either `u8` or `char`, and then another associated type with a bound `Iterator<Item = Self::Character>`.

youknowone: I tried this several days ago but It was not easy due to lifetime of `Chars`

coolreader18: You could just have a function in `pystr.rs` that takes a `Iterator` and does the operation, it doesn't need to be a method. Or just live with the code duplication for now, it's only 2 instances.

vm/src/obj/pystr.rs
```diff
@@ -297,4 +297,69 @@ pub trait PyCommonString<E> {
             &self
         }
     }
+
+    fn py_splitlines<U8, FW, W>(
+        &self,
+        bytes: impl std::iter::Iterator<Item = U8>,
```
youknowone, line 303: Unfortunately, `str::bytes()` is `std::iter::Itarator<Item=u8>` but `[u8]::iter()` is `std::iter::Iterator<Item=&u8>`. Can I do this without `CopyTo` trait?

coolreader18: You could probably just give it a byte slice using `str.as_bytes()`

youknowone: Thanks, it works. I am still curious how to write a generic function to take `T` and `&T` when T is a Copy. it is practically same but I couldn't find an easy way with std.

PR #1884, opened 2020-04-24
Disable the unpack optimization

### LINKED ISSUE #1878: Tuple unpacking the result of an if expression is miscompiled, leads to panic
## Feature

Minimal reproduction:
```py
a, b = (1, 2) if "s" else (3, 4)
```

coolreader18:
Fixes #1878


palaviv:
Let's add a test for this to avoid regression.

PR #1886, opened 2020-04-27
Add os.setreuid and os.setresuid methods

toan-n-phan:


youknowone:
Thanks for contributing! Generally It looks nice but macos doesn't seem to have `setresuid`

PR #1889, opened 2020-04-27
Add the _json module and _json.make_scanner

coolreader18:


vm/src/vm.rs
```diff
@@ -583,6 +583,13 @@ impl VirtualMachine {
     pub fn is_none(&self, obj: &PyObjectRef) -> bool {
         obj.is(&self.get_none())
     }
+    pub fn option_if_none(&self, obj: PyObjectRef) -> Option<PyObjectRef> {
```
youknowone, line 586: maybe `Option<T>::try_from_object` work as same

PR #1891, opened 2020-04-27
Fix a few codecs tests

coolreader18:


Lib/_codecs.py
```diff
@@ -115,7 +115,7 @@ def encode(v, encoding=None, errors='strict'):
     if not isinstance(errors, str):
         raise TypeError("Errors must be a string")
     codec = lookup(encoding)
-    res = codec.encode(v, errors)
+    res = codec[0](v, errors)
```
youknowone, line 118: are this fixes copy of certain version out of rustpython or original fix?

coolreader18: They're original fixes by me, since the original pypy version is old.

PR #1892, opened 2020-04-28
Make more objects ThreadSafe

palaviv:


palaviv:
I would love to get some other people opinion here. I have converted `PyNativeFunc` to require `Send + Sync`. The problem is that in 2 places in WASM we use closure in the function so we can call `js_sys::Function` from the context. `js_sys::Function` is `!Send + !Sync` but from looking into it threading is currently not implemented in WASM any way. Anyone has a suggestion how to solve this? @coolreader18?

palaviv:
> I would love to get some other people opinion here. I have converted `PyNativeFunc` to require `Send + Sync`. The problem is that in 2 places in WASM we use closure in the function so we can call `js_sys::Function` from the context. `js_sys::Function` is `!Send + !Sync` but from looking into it threading is currently not implemented in WASM any way. Anyone has a suggestion how to solve this? @coolreader18?

I have decided to use thread local storage to save the `js_sys::Function`. Currently WASM is single threaded so this should not be a problem.

wasm/lib/src/convert.rs
```diff
@@ -13,6 +15,9 @@ use rustpython_vm::{exceptions, py_serde};
 use crate::browser_module;
 use crate::vm_class::{stored_vm_from_wasm, WASMVirtualMachine};
 
+// Currently WASM do not support multithreading. We should change this once it is enabled.
+thread_local!(static JS_FUNC: RefCell<Option<js_sys::Function>> = RefCell::new(None));
```
coolreader18, line 19: Maybe this could be a vec instead, so multiple js functions can be passed to python? And then you could do something like:
```rust
let idx = JS_FUNC.with(|funcs| {
    let mut funcs = funcs.borrow_mut();
    funcs.push(func);
    funcs.len() - 1
});
```
and then capture `idx` in a closure.

palaviv: The problem is that we can't use closures because they are `!Send+!Sync`. Maybe I can create a new type of method that has a `idx` as well.

coolreader18: Are you sure? https://doc.rust-lang.org/reference/types/closure.html#other-traits

palaviv: > Are you sure? https://doc.rust-lang.org/reference/types/closure.html#other-traits

Sure I am wrong. Thanks :pray:

Issue #1714, opened 2020-01-29
Replace the remaining arg_check! calls with parameter-based arg checking


youknowone:

coolreader18:
I removed the RFC tag because this is more just some busywork that we've been chipping away at for a while, not necessarily something to decide on.

PR #1893, opened 2020-04-29
Remove arg_check! in tokenize

mrmiywj:
Remove the arg_check! in the tokenize function. #1714


youknowone:
Thank you for contributing!

PR #1894, opened 2020-04-29
Add zlib.decompressobj()

coolreader18:
This lets you run `rustpython -m zipfile` operations that decompress zip files!

### DIFF
--- a/vm/src/stdlib/zlib.rs
+++ b/vm/src/stdlib/zlib.rs
@@ -1,16 +1,20 @@
 use crate::exceptions::PyBaseExceptionRef;
 use crate::function::OptionalArg;
-use crate::obj::objbytes::PyBytesRef;
-use crate::pyobject::{ItemProtocol, PyObjectRef, PyResult};
+use crate::obj::objbyteinner::PyBytesLike;
+use crate::obj::objbytes::{PyBytes, PyBytesRef};
+use crate::obj::objtype::PyClassRef;
+use crate::pyobject::{PyClassImpl, PyObjectRef, PyResult, PyValue};
 use crate::types::create_type;
 use crate::vm::VirtualMachine;
 
 use adler32::RollingAdler32 as Adler32;
 use crc32fast::Hasher as Crc32;
+use crossbeam_utils::atomic::AtomicCell;
 use flate2::{write::ZlibEncoder, Compression, Decompress, FlushDecompress, Status};
 use libz_sys as libz;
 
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/zlib.rs
```diff
@@ -111,15 +118,187 @@ fn zlib_decompress(
...
+        decompress: Mutex::new(decompress),
+        eof: AtomicCell::new(false),
+        unused_data: Mutex::new(PyBytes::new(vec![]).into_ref(vm)),
+        unconsumed_tail: Mutex::new(PyBytes::new(vec![]).into_ref(vm)),
+    }
+}
+#[pyclass(name = "Decompress")]
+#[derive(Debug)]
+struct PyDecompress {
+    decompress: Mutex<Decompress>,
```
palaviv, line 141: Why not `RwLock`?

vm/src/stdlib/zlib.rs
```diff
@@ -111,15 +118,187 @@ fn zlib_decompress(
...
+    eof: AtomicCell<bool>,
+    unused_data: Mutex<PyBytesRef>,
```
palaviv, line 143: Why not using `Vec<u8>`?

vm/src/stdlib/zlib.rs
```diff
@@ -111,15 +118,187 @@ fn zlib_decompress(
...
+    PyDecompress {
+        decompress: Mutex::new(decompress),
+        eof: AtomicCell::new(false),
+        unused_data: Mutex::new(PyBytes::new(vec![]).into_ref(vm)),
+        unconsumed_tail: Mutex::new(PyBytes::new(vec![]).into_ref(vm)),
+    }
+}
+#[pyclass(name = "Decompress")]
+#[derive(Debug)]
+struct PyDecompress {
```
palaviv, line 140: Great work making this already thread safe :100:

vm/src/stdlib/zlib.rs
coolreader18: I figured there's not really a point, since every time we access it it's going to be to mutate it.

vm/src/stdlib/zlib.rs
coolreader18: Oh, there's supposed to be a getter for it, so it's that when somebody does `decompressor.unused_data` multiple times it doesn't clone the data, just the reference.

PR #1898, opened 2020-04-29
Cache str.len()

youknowone:
Rust str is utf-8, so it doesn't match to CPython str very well.
Probably we would reimplement PyString as same as CPython, but this is a
temporal fix for easy speedup.

Before:
    $ time cargo run --release Lib/test/test_unicode.py
    13.56s user 0.16s system 88% cpu 15.487 total

After:
    $ time cargo run --release Lib/test/test_unicode.py
    10.02s user 0.10s system 88% cpu 11.448 total

### DIFF
--- a/vm/src/obj/objstr.rs
+++ b/vm/src/obj/objstr.rs
@@ -50,6 +50,7 @@ use crate::vm::VirtualMachine;
 pub struct PyString {
     value: String,
     hash: AtomicCell<Option<pyhash::PyHash>>,
+    len: AtomicCell<Option<usize>>,
 }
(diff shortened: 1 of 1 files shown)


coolreader18:
Great idea!

coolreader18:
Is there anywhere that uses `.chars().count()` of a PyString to do some operation that we could replace with a call to `len()`?

youknowone:
Unfortunately, it is usually the opposite. We are using `len()` instead of `.chars().count()` so it sometimes make trouble with non-ascii strings.

PR #1899, opened 2020-04-29
The any key/p38 self documenting fstrings

TheAnyKey:
This PR contains the following changes for fstring:

### fix: so far it was not allowed that a conversion flag is followed by a format specifier. 
so far f"{'42'!s:10}" rose a parsing error as a '}' was expected after a conversion flag.This is fix now.

### implemented self documenting fstrings in rustpython
I implemented the self documenting fstrings in rustpython, as introduced with python 3.8. I.e., the following syntax is now allowed:

>>> foo="bar"
>>> f"{foo=}"
foo=bar
>>> f"{2*21=}"
2*21=42
...

Also combination with conversion flags and format specs are supported.

Open: the behaviour when the type of the expression is a string is so far not conform to cpython, thus those are temporarily modified or skipped. 

### Ported tests from cpython for self documenting fstrings (and some more for format spec handling)
Imported the tests for  selfdocumentign fstrings from cpython to rustpython.

### DIFF
--- a/parser/src/fstring.rs
+++ b/parser/src/fstring.rs
@@ -73,6 +73,11 @@ impl<'a> FStringParser<'a> {
                             return Err(ExpectedRbrace);
                         }
                     });
+
+                    let peek = self.chars.peek();
+                    if peek != Some(&'}') && peek != Some(&':') {
+                        return Err(ExpectedRbrace);
+                    }
                 }
 
                 // match a python 3.8 self documenting expression
@@ -383,6 +388,7 @@ mod tests {
     #[test]
     fn test_parse_invalid_fstring() {
         assert_eq!(parse_fstring("{5!a"), Err(ExpectedRbrace));
+
         assert_eq!(parse_fstring("{5!a1}"), Err(ExpectedRbrace));
         assert_eq!(parse_fstring("{5!"), Err(ExpectedRbrace));
 
--- a/parser/src/fstring.rs
+++ b/parser/src/fstring.rs
@@ -83,7 +83,6 @@ impl<'a> FStringParser<'a> {
                 // format '{' PYTHON_EXPRESSION '=' FORMAT_SPECIFIER? '}'
                 '=' if self.chars.peek() != Some(&'=') => { // check for delims empty?
                     pred_expression_text = expression.to_string(); // safe expression before = to print it
-                    //  pred_expression_text contains trailing spaces, which are trimmed somewhere later before printing, this needs to be prevented orovercome here
                 }
 
                 ':' if delims.is_empty() => {
--- a/parser/src/fstring.rs
+++ b/parser/src/fstring.rs
@@ -73,10 +73,6 @@ impl<'a> FStringParser<'a> {
                             return Err(ExpectedRbrace);
                         }
                     });
-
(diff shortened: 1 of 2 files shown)


parser/src/fstring.rs
```diff
@@ -139,6 +197,11 @@ impl<'a> FStringParser<'a> {
                         }
                     }
                 }
+
```
youknowone, line 200:
```suggestion
```

parser/src/fstring.rs
```diff
@@ -139,6 +197,11 @@ impl<'a> FStringParser<'a> {
...
+                ' ' if !pred_expression_text.is_empty() => {
+                    trailing_seq.push(ch);
+                }
+
```
youknowone, line 204:
```suggestion
```

parser/src/fstring.rs
```diff
@@ -298,9 +361,34 @@ mod tests {
...
+        let src = String::from("{user=:>10}");
+        let parse_ast = parse_fstring(&src);
+
+        assert!(parse_ast.is_ok());
+    }
+
     #[test]
     fn test_parse_invalid_fstring() {
         assert_eq!(parse_fstring("{5!a"), Err(ExpectedRbrace));
+
```
youknowone, line 391:
```suggestion
```

tests/snippets/fstrings.py
```diff
@@ -1,4 +1,20 @@
 from testutils import assert_raises
+
+#test only makes sense with python 3.8 or higher (or RustPython)
+import sys
+import platform
+if platform.python_implementation() == 'CPython':
+    assert sys.version_info.major == 3, 'Incompatible Python Version, expected CPython 3.8 or later'
+    assert sys.version_info.minor == 8, 'Incompatible Python Version, expected CPython 3.8 or later'
```
youknowone, line 8:
```suggestion
    assert sys.version_info >= (3, 8), 'Incompatible Python Version, expected CPython 3.8 or later'
```

parser/src/fstring.rs
```diff
@@ -26,9 +26,37 @@ impl<'a> FStringParser<'a> {
...
+        let mut pred_expression_text = String::new();
+        let mut trailing_seq = String::new();
 
         while let Some(ch) = self.chars.next() {
             match ch {
+                // can be integrated better with the remainign code, but as a starting point ok
+                // in general I would do here a tokenizing of the fstrings to omit this peeking.
+                '!' if self.chars.peek() == Some(&'=') => {
+                    expression.push('!');
+                    expression.push('=');
```
youknowone, line 38:
```suggestion
                    expression.push_str("!=");
```
And the same for below matches

parser/src/fstring.rs
```diff
@@ -46,10 +74,19 @@ impl<'a> FStringParser<'a> {
                         }
                     });
 
-                    if self.chars.peek() != Some(&'}') {
+                    let peek = self.chars.peek();
+                    if peek != Some(&'}') && peek != Some(&':') {
```
youknowone, line 78:
```suggestion
                    if let Some(peek) = self.chars.peek() {
                        if peek != '}' && peek != ':' {
```

PR #1902, opened 2020-05-03
Implement P3.9 style dict union (PEP584)

TheAnyKey:
This PR implements the dict union as defined in PEP 854 in RustPython and tests it. Compability tests with Cpython are skipped when tested with Python 3.8 or below (due to abvious reason - Githib doesnt support python 3.9 so far, changing ci configuration failed).

Now it is possible to write things like:
>>> a = {1:2, 2:3}
>>> b = {3:4, 5:6}
>>> c=a|b
>>> a|=b

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,7 +62,7 @@ jobs:
           args: --release --verbose --all
       - uses: actions/setup-python@v1
         with:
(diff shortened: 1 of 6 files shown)


coolreader18:
Further reviews/development can be directed to #1911, I believe.

PR #1903, opened 2020-05-03
Implement Python 3.9 style string functions: removeprefix and removesuffix PEP616

TheAnyKey:
This PR implements and tests the new string functions introduced by Python 3.9 to remove prefixes and suffixes. This change is defined in PEP616. Now it is possible to do:
>>> s='foobarfoo'
>>>t=s.removeprefix('foo')
>>>print(t.removesuffix('foo')
'bar'


TheAnyKey:
The list of commits is quite long for such a minimal change, as I messed up branches and later need to clean it up. But the final change is just objstr.rs and the related tests.

TheAnyKey:
pulled to main

youknowone:
Do you want to try to fix the commit list? when your remote `upstream` is rustpython official, `git rebase -i upstream/master` will give you editable commit list. remove everything except for the last commits.

If you don't have upstream remote, try to add it by `git remote add upstream https://github.com/rustpython/rustpython.git`

coolreader18:
Alternatively with `git rebase -i` you could try squashing the changes. It'll show you a list that looks something like this when you run `git rebase -i upstream/master`:

```
pick abcdef message
pick aafebf message 2
pick ffffffb third message
```
And if you change it to:
```
pick abcdef message
squash aafebf message 2
squash ffffffb third message
```
Then it will put all of the commits onto the first one, and let you change the commit message for it.

TheAnyKey:
Replaced by cleanedup PR #1915

PR #1907, opened 2020-05-04
Implement Py3.9 LCM and GCD

TheAnyKey:
This PR implements the bpo-39648 and bpo-39479, which are planned for CPython 3.9.

This change extends the math module to implement a lcm (least common multiple) function analog to the gcd. Further both functions are extended to take an arbitrary number of int arguments. 

The kind of args is in accordance to the 3.9.a6 cpython implementation, although this style of parameters is quite uncommon in the math module; in similar cases iterables are used as arguments. Maybe this changes before release.

Further, I imported the test_math.py module from CPython and skipped all failing tests (unluckily there are plenty of them)

Whats missing: cpython also accepts object arguments having __index__ implemented, this seems so far not be finalized in the reference. Thus I defer the implementation.

The sneaky improvement in this PR is the pdc.sh (pre-delivery check), i.e., a poor mans automation to run all builds and checks.


TheAnyKey:
replaced by #1916

PR #1908, opened 2020-05-05
Allow doctest to be imported and run

coolreader18:


src/shell.rs
```diff
@@ -19,19 +19,10 @@ enum ShellExecResult {
 
 fn shell_exec(vm: &VirtualMachine, source: &str, scope: Scope) -> ShellExecResult {
     match vm.compile(source, compile::Mode::Single, "<stdin>".to_owned()) {
-        Ok(code) => {
-            match vm.run_code_obj(code, scope.clone()) {
-                Ok(value) => {
-                    // Save non-None values as "_"
-                    if !vm.is_none(&value) {
-                        let key = "_";
```
windelbouwman, line 27: This is no longer required? This would break the interactive mode in which we can use the last result as the underscore right?

wasm/lib/src/wasm_builtins.rs
```diff
@@ -4,75 +4,37 @@
...
-use rustpython_vm::pyobject::{IdProtocol, PyResult, TypeProtocol};
+use rustpython_vm::obj::objstr::PyStringRef;
+use rustpython_vm::pyobject::{PyObjectRef, PyResult};
 use rustpython_vm::VirtualMachine;
 
 pub(crate) fn window() -> web_sys::Window {
     web_sys::window().expect("Window to be available")
 }
 
-pub fn format_print_args(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult<String> {
```
windelbouwman, line 19: More than happy to see this go.

src/shell.rs
windelbouwman: Ignore my comment, I continued reading further down the change.

PR #1912, opened 2020-05-08
add support for sync func in os module

mrmiywj:
As title, sync function is available in UNIX only.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1444,6 +1444,14 @@ fn os_utime(
     unimplemented!("utime")
 }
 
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/os.rs
```diff
@@ -1444,6 +1444,14 @@ fn os_utime(
     unimplemented!("utime")
 }
 
+#[cfg(unix)]
+fn os_sync(_vm: &VirtualMachine) -> PyResult<()> {
```
palaviv, line 1448: You can do `-> ()` Instead of wrapping in the `PyResult`.

PR #1914, opened 2020-05-08
add getpgrp support in os mod

mrmiywj:
As title, `getpgrp` support in os module in UNIX only


vm/src/stdlib/os.rs
```diff
@@ -1194,6 +1194,11 @@ fn os_getpgid(pid: u32, vm: &VirtualMachine) -> PyResult {
     }
 }
 
+#[cfg(unix)]
+fn os_getpgrp(vm: &VirtualMachine) -> PyResult {
```
palaviv, line 1198: Again you can just return `i32`.

PR #1915, opened 2020-05-08
Implement Python 3.9 style string functions: removeprefix and removesuffix PEP616

TheAnyKey:
This PR implements and tests the new string functions introduced by Python 3.9 to remove prefixes and suffixes. This change is defined in PEP616. Now it is possible to do:

            s='foobarfoo'
            t=s.removeprefix('foo')
            print(t.removesuffix('foo'))
            'bar'

Clean replacement of #1903

### DIFF
--- a/vm/src/obj/objbytearray.rs
+++ b/vm/src/obj/objbytearray.rs
@@ -389,24 +389,38 @@ impl PyByteArray {
         self.borrow_value().rstrip(chars).into()
     }
 
+    /// removeprefix($self, prefix, /)
+    ///
+    ///
+    /// Return a bytearray object with the given prefix string removed if present.
+    ///
+    /// If the bytearray starts with the prefix string, return string[len(prefix):]
+    /// Otherwise, return a copy of the original bytearray.
     #[pymethod(name = "removeprefix")]
     fn removeprefix(&self, prefix: PyByteInner) -> PyByteArray {
-        let value = self.borrow_value();
-        if value.elements.starts_with(&prefix.elements) {
-            return value.elements[prefix.elements.len()..].to_vec().into();
-        }
-        value.elements.to_vec().into()
+        self.borrow_value().elements[..]
+            .py_removeprefix(&prefix.elements, prefix.elements.len(), |s, p| {
+                s.starts_with(p)
+            })
+            .to_vec()
+            .into()
     }
 
+    /// removesuffix(self, prefix, /)
+    ///
+    ///
+    /// Return a bytearray object with the given suffix string removed if present.
+    ///
+    /// If the bytearray ends with the suffix string, return string[:len(suffix)]
+    /// Otherwise, return a copy of the original bytearray.
     #[pymethod(name = "removesuffix")]
     fn removesuffix(&self, suffix: PyByteInner) -> PyByteArray {
-        let value = self.borrow_value();
-        if value.elements.ends_with(&suffix.elements) {
-            return value.elements[..value.elements.len() - suffix.elements.len()]
-                .to_vec()
-                .into();
-        }
-        value.elements.to_vec().into()
+        self.borrow_value().elements[..]
+            .py_removesuffix(&suffix.elements, suffix.elements.len(), |s, p| {
+                s.ends_with(p)
+            })
+            .to_vec()
+            .into()
     }
 
     #[pymethod(name = "split")]
--- a/vm/src/obj/objbytearray.rs
+++ b/vm/src/obj/objbytearray.rs
@@ -389,6 +389,26 @@ impl PyByteArray {
         self.borrow_value().rstrip(chars).into()
     }
 
+    #[pymethod(name = "removeprefix")]
+    fn removeprefix(&self, prefix: PyByteInner) -> PyByteArray {
+        let value = self.borrow_value();
+        if value.elements.starts_with(&prefix.elements) {
+            return value.elements[prefix.elements.len()..].to_vec().into();
+        }
+        value.elements.to_vec().into()
+    }
+
+    #[pymethod(name = "removesuffix")]
+    fn removesuffix(&self, suffix: PyByteInner) -> PyByteArray {
+        let value = self.borrow_value();
+        if value.elements.ends_with(&suffix.elements) {
+            return value.elements[..value.elements.len() - suffix.elements.len()]
+                .to_vec()
+                .into();
+        }
+        value.elements.to_vec().into()
+    }
+
     #[pymethod(name = "split")]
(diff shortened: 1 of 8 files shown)


youknowone:
Nice work! Instead of opening a new PR, push-force to existing branch would update the PR too. Try it for other PRs.

youknowone:
The new commit message is not representing this change. Would you revise the commit message? `git commit --amend` will work for that.

TheAnyKey:
oh, when doing next time such a mess I ll use push -force. Now it should be clean.

vm/src/obj/objstr.rs
```diff
@@ -532,6 +532,22 @@ impl PyString {
...
+        if self.value.as_str().starts_with(&pref.value) {
+            return Ok(self.value[pref.len()..].to_string());
+        }
+        Ok(self.value.to_string())
+    }
+
+    #[pymethod]
+    fn removesuffix(&self, suff: PyStringRef) -> PyResult<String> {
+        if self.value.as_str().ends_with(&suff.value) {
+            return Ok(self.value[..self.value.len() - suff.len()].to_string());
```
youknowone, line 546: `self.value.len()` is a rust string length and `suff.len()` is a python string length. These numbers should not be mixed because they heve different value for non-ascii string. Will you please add non-ascii tests to the test code?

TheAnyKey: Thanks, pretty obvious fault.

TheAnyKey: in the py39  tests I found no such tests when creating it. maybe I  missed it or the test are also behind. anyhow as you are not interested at the moment I ll stop sening prs.

vm/src/obj/objbytearray.rs
```diff
@@ -389,6 +389,26 @@ impl PyByteArray {
         self.borrow_value().rstrip(chars).into()
     }
 
+    #[pymethod(name = "removeprefix")]
+    fn removeprefix(&self, prefix: PyByteInner) -> PyByteArray {
+        let value = self.borrow_value();
+        if value.elements.starts_with(&prefix.elements) {
+            return value.elements[prefix.elements.len()..].to_vec().into();
+        }
+        value.elements.to_vec().into()
```
youknowone, line 398: `bytearray` and `bytes` share the inner implementation. So this will work:

```suggestion
        self.borrow_value().removeprefix(prefix)
```

vm/src/obj/objbyteinner.rs
```diff
@@ -843,6 +843,22 @@ impl PyByteInner {
             .to_vec()
     }
 
+    // new in Python 3.9
+    pub fn removeprefix(&self, prefix: PyByteInner) -> Vec<u8> {
+        if self.elements.starts_with(&prefix.elements) {
+            return self.elements[prefix.elements.len()..].to_vec();
+        }
+        self.elements.to_vec()
```
youknowone, line 851: I feel like this line cause unnessessary data copying instead of increment of reference count. But I don't have simple fix suggestion of this.

vm/src/obj/objstr.rs
```diff
@@ -532,6 +532,22 @@ impl PyString {
         )
     }
 
+    #[pymethod]
+    fn removeprefix(&self, pref: PyStringRef) -> PyResult<String> {
+        if self.value.as_str().starts_with(&pref.value) {
```
youknowone, line 537: Basically this is the same as byteinner implementation. If you are interested in, there is a common string implementation file `pystr.rs`. If you don't want to try this change, it is good as it is.

TheAnyKey: Thanks for the hint, I had already realized this also the existance of pystr. All sotions I found here, would lead to very fine granular functions that need to be implemented in specializations of the trait. I have not found a way to a common trait or predicate that offers a starts_with or an equivalent function (at least not without significant overhead). One way would be using Pattern, but that is marked as unstable. Do you know a way to implement that?

youknowone: Because this is simple string match, I think python `str.startswith` is the the similar case. I think adding `py_removeaffix` will be very similar to this one.

- pystr part: https://github.com/RustPython/RustPython/blob/master/vm/src/obj/pystr.rs#L172-L205
- objstr part: https://github.com/RustPython/RustPython/blob/master/vm/src/obj/objstr.rs#L524-L533
- objbytes part: https://github.com/RustPython/RustPython/blob/master/vm/src/obj/objbytes.rs#L287-L300

Please let me know if I missed something you concern about.

TheAnyKey: In that functions, the unified implementation helps fa rmore as the logic is more tricky and there is some more error handling required. I found an acceptable trade off.

vm/src/obj/objbytearray.rs
```diff
@@ -389,6 +389,40 @@ impl PyByteArray {
...
+    /// If the bytearray starts with the prefix string, return string[len(prefix):]
+    /// Otherwise, return a copy of the original bytearray.
+    #[pymethod(name = "removeprefix")]
+    fn removeprefix(&self, prefix: PyByteInner) -> PyByteArray {
+        self.borrow_value().elements[..]
+            .py_removeprefix(&prefix.elements, prefix.elements.len(), |s, p| {
+                s.starts_with(p)
+            })
+            .to_vec()
+            .into()
```
youknowone, line 406: `self.borrow_value()` is `ByteInner` object. so `self.borrow_value().removeprefix(prefix)` will work

vm/src/obj/pystr.rs
```diff
@@ -264,4 +265,41 @@ pub trait PyCommonString<E> {
...
+    ) -> &Self
+    where
+        FC: Fn(&Self, &Self::Container) -> bool,
+    {
+        //if self.py_starts_with(prefix) {
+        if is_prefix(&self, &prefix) {
+            return self.get_bytes(Range {
+                start: prefix_len,
+                end: self.bytes_len(),
+            });
```
youknowone, line 283:
```suggestion
            return self.get_bytes(prefix_len..self.bytes_len())
```

vm/src/obj/objstr.rs
youknowone: great! I like this new implementation

vm/src/obj/pystr.rs
```diff
@@ -264,4 +265,41 @@ pub trait PyCommonString<E> {
...
+        }
+        &self
```
youknowone, line 285: I actually want to suggest remove `return`.

```suggestion
        if is_prefix(&self, &prefix) {
            self.get_bytes(prefix_len..self.bytes_len())
        } else {
            &self
        }
```

youknowone:
Nice work! Thank you for contributing.

PR #1916, opened 2020-05-08
Implement Py3.9 LCM and GCD

TheAnyKey:
This PR implements the bpo-39648 and bpo-39479, which are planned for CPython 3.9.

This change extends the math module to implement a lcm (least common multiple) function analog to the gcd. Further both functions are extended to take an arbitrary number of int arguments.

The kind of args is in accordance to the 3.9.a6 cpython implementation, although this style of parameters is quite uncommon in the math module; in similar cases iterables are used as arguments. Maybe this changes before release.

Further, I imported the test_math.py module from CPython and skipped all failing tests (unluckily there are plenty of them)

Whats missing: cpython also accepts object arguments having index implemented, this seems so far not be finalized in the reference. Thus I defer the implementation.

replace #1907

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -9,9 +9,9 @@ use statrs::function::gamma::{gamma, ln_gamma};
 use num_bigint::BigInt;
 use num_traits::{One, Zero};
 
-use crate::function::OptionalArg;
+use crate::function::{OptionalArg, PyFuncArgs};
 use crate::obj::objfloat::{self, IntoPyFloat, PyFloatRef};
-use crate::obj::objint::{self, PyIntRef};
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/math.rs
```diff
@@ -272,9 +272,55 @@ fn math_ldexp(
     Ok(value * (2_f64).powf(objint::try_float(i.as_bigint(), vm)?))
 }
 
-fn math_gcd(a: PyIntRef, b: PyIntRef) -> BigInt {
+fn math_perf_arb_len_int_op<F>(
+    args: PyFuncArgs,
```
palaviv, line 276: You can use here `Args<PyIntRef>`. Look at `PySetInner::update` for example. This will remove some of the argument parsing code.

PR #1917, opened 2020-05-08
Fix unpack error messages and add unittest test_unpack.py

youknowone:


vm/src/frame.rs
```diff
@@ -659,9 +659,30 @@ impl ExecutingFrame<'_> {
...
+                        vm.new_type_error(format!(
+                            "cannot unpack non-iterable {} object",
+                            value.class().name
+                        ))
+                    } else {
+                        e
+                    }
+                })?;
+                let ordering = elements.len().cmp(size);
+                if ordering != std::cmp::Ordering::Equal {
```
coolreader18, line 673: Maybe you could have the `for element in ...` be a match arm of `match ordering` to avoid the unsafe unreachable hint?

youknowone: Thanks, it looks better

PR #1918, opened 2020-05-08
Frame.f_trace, paving the way to support `pdb` debugging

alanjds:
f_trace is initialized with PyNone. I do not know if it would be better on memory/cpu to handle Option::None instead.

No tests for now, just wanting a feedback. Specially on the RwLock used.


coolreader18:
I think it would be a micro-optimization for `PyObjectRef` vs `Option<PyObjectRef>`, as they're the same size since `Option::<Rc>::None` is represented by a null pointer. `vm.is_none(&obj)` vs `opt.is_none()` is I think 1 asm instruction different, as it's basically `ptr == 0` vs `ptr == vm->none_address`

alanjds:
I see. The part intriguing me is that Option::None codepath is followed by `return vm.get_none()`

But storing PyNone lead to return `.unwrap().clone()` and I could not dodge this refcount cost on .clone()

alanjds:
I may be overthinking this all. Will keep pushing "just work" code until pdb.set_trace() works and then get back to optimize.

What about the rwlock? Is it ok? The rational was that I expect it to be checked on every single line change, so should be fast. But also light on memory as will be on every frame created.

alanjds:
...and thanks for the attention @coolreader18.

coolreader18:
Yeah, I think the RwLock is good, it makes sense since you might occasionally be reading the f_trace property concurrently.

alanjds:
- `pdb.runcall()` working, but not quite right. The 1st line stopped is into bdb yet :thinking:  
- `(pdb) args` works, but also behaves as if I had `(pdb) next` too :thinking: 
- Changed from RwLock to Mutex. On single-threaded is faster, on multiple it works IIUC.
- Optimized a bit the codepath of vm::trace_event for the case of no trace function set, the default.

alanjds:
And this is the end of PyCon US 2020 sprint:
- `(pdb) continue` works as expected, not breaking at some `del frame.f_trace`

Thanks @youknowone for the help on Gitter.
Had not found how do CPython implement `del frame.f_trace`. I implemented via `frame.__delattr__`. Yet I do not know how to call `super()` from Rust side. Will need help on this.

Now may need support for a setter on `frame.lineno` if wanting to have `(pdb) jump` to work. Maybe for `(pdb) next` too, as seems broken.

Good night.

vm/src/obj/objframe.rs
```diff
@@ -24,6 +24,15 @@ impl FrameRef {
         "<frame object at .. >".to_owned()
     }
 
+    #[pymethod(name = "__delattr__")]
+    fn delattr(self, value: PyObjectRef, vm: &VirtualMachine) {
+        println!("Called: FrameRef::f_trace over {:?}", value);
```
youknowone, line 29: nice workaround for now. putting an incompatibility comment will be helpful for future work.

youknowone:
Do you want to work more for this or want to merge part of the work?

alanjds:
Both 😬.

Better to merge the actual progress for this to not get rotten when I find the time to finish it.

youknowone:
Will you make a new PR with finished commits after rebase?

alanjds:
> Will you make a new PR with finished commits after rebase?

I am not used to rebase stuff, but can learn to.

Should I go cherry-picking over master and send a new PR? Is this the way?

youknowone:
Oh, yep. Because this one is very aged, cherry-pick could be easier than rebase. That sounds very reasonable.

coolreader18:
Oh, right -- looks like the errors are from `.lock().unwrap()`; we switched to parking_lot `Mutex`s since this PR was opened, so the `.unwrap()`s shouldn't be necessary.

coolreader18:
Is the `gc.py` module necessary if everything raises a NotImplementedError anyway? We've mostly just been commenting out `gc` imports and function calls in modules that have them, is there a reason that can't work for pdb?

alanjds:
> Is the `gc.py` module necessary if everything raises a NotImplementedError anyway? We've mostly just been commenting out `gc` imports and function calls in modules that have them, is there a reason that can't work for pdb?

bdb.py imports it to use somewhere, but I had not hit the case when it calls gc for real.

Anyway I do not feel bad by stubbing C-based stuff that we want not to implement _right now_.

(I need to say that sometimes I am missing the ability of calling rust code directly from python code. Maybe on future a custom importhook can do the trick.)

coolreader18:
@youknowone what do you think?

alanjds:
> We've mostly just been commenting out gc imports and function calls in modules that have them

My approach on this is to try to do the lowest number of changes possible on foreign code, if we want to get back to the original somewhere in future. That is why I am trying to provide a bogus gc instead of messing with bdb.py original code.

Less changes to track upstream on bdb.py and keep patching here until a fully compatible gc is ready.

alanjds:
https://github.com/RustPython/RustPython/pull/1918/checks?check_run_id=973313037#step:10:7685
```
AttributeError: module 'sys' has no attribute 'stdout'
```

What? I though that this was been fixed before :thinking:. I saw a PR about this!

alanjds:
Looks like sys.displayhook was natively implemented in meantime. Removing my hack from `site.py` and waiting for completion.

For some reason my local tests passes. I may have a local environment messed up.

alanjds:
Yay! Got green.

Now pdb.set_trace() "works" but stops on the wrong place.
Then works:
- a(rgs)
- c(ontinue)
- h(elp)
- l(ist)
- w(here)
- local evaluation also works

Yet broken:
- n(ext): Is ignored, showing the very same line forever 
- r(eturn): Is ignored, showing the very same line forever
- j(ump): Is ignored, and a follow c(ontinue) raises `AttributeError: attribute 'f_lineno' of 'frame' objects is not writable`

This PR is good to go. At least locals() and globals() variables inspection are working. Clojure variables inspection seemed broken.

Lets get this merged and open another to fix j(ump) and n(ext)

coolreader18:
I'm merging this manually to fix this weird lf->crlf change in frame.rs

coolreader18:
Alright, this is merged to master even though github doesn't show it.

PR #1919, opened 2020-05-09
Convert stdlib object to thread safe

palaviv:


Lib/test/test_array.py
```diff
@@ -341,8 +341,6 @@ def test_iterator_pickle(self):
             a.fromlist(data2)
             self.assertEqual(list(it), [])
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
```
coolreader18, line 345: It's interesting how this "overflow" behavior is actually correct with how iterators work in CPython

PR #1922, opened 2020-05-12
Fix unpack_ex error messages

youknowone:
`test_unpack_ex.py` 14/85 fails with this patch

### DIFF
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -1399,7 +1399,14 @@ impl<O: OutputStream> Compiler<O> {
                     }
                 }
             }
-            _ => return Err(self.error(CompileErrorType::Assign(target.name()))),
+            _ => {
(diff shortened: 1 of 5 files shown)


compiler/src/compile.rs
```diff
@@ -2031,21 +2034,34 @@ impl<O: OutputStream> Compiler<O> {
             }
         }
 
+        let mut compile_element = |element| {
+            self.compile_expression(element).map_err(|e| {
+                println!("ahahah {:?}", e);
```
coolreader18, line 2039:
```suggestion
```

coolreader18 (approved):

compiler/src/compile.rs
youknowone: Oops, thanks for checking this

PR #1929, opened 2020-05-16
EOF after `\\` raise EOF Error

### LINKED ISSUE #1928: Line continuation character does not work properly in python shell.
### Expected Result
```python
>>> a = \
... 
```
### Actual Result
```python
>>>>> a = \
SyntaxError: Got unexpected token Newline at line 2 column 1
```

https://github.com/RustPython/RustPython/blob/baf3ec16c024364ef87ee68d30dc1470b6eecca9/src/shell.rs#L20-L32

In the current code, if `vm.compile` return `ParseErrorType::EOF`, return `ShellExecResult::Continue` to receive the next line.
However, if the lexer returns EOF after `\\`, the parser raises `unexpected token error` and`ParseErrorType::EOF` does not occur, so `ShellExecResult :: Continue` is not returned.

https://github.com/python/cpython/blob/4a12d121860fb60d56cdcc212817577cac2356d0/Parser/tokenizer.c#L1753-L1771

In cpython code, an error occurred when EOF after `\\` in the lexer step.

HyeockJinKim:
Return EOF Error to get the next line after` \\ `in the shell.

Closes #1928

### DIFF
--- a/parser/src/error.rs
+++ b/parser/src/error.rs
@@ -28,6 +28,7 @@ pub enum LexicalErrorType {
     UnrecognizedToken { tok: char },
     FStringError(FStringErrorType),
     LineContinuationError,
+    EOF,
     OtherError(String),
 }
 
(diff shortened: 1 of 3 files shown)


coolreader18:
Nice!

PR #1933, opened 2020-05-17
Move some vm fields to an Arc-wrapped struct

coolreader18:
In preparation for threading; we want to keep some properties of the vm consistent across threads, so I'm moving the ones that should be into an `Arc`.

### DIFF
--- a/vm/src/vm.rs
+++ b/vm/src/vm.rs
@@ -4,7 +4,7 @@
 //!   https://github.com/ProgVal/pythonvm-rust/blob/master/src/processor/mod.rs
 //!
 
-use std::cell::{Cell, Ref, RefCell};
+use std::cell::{Ref, RefCell};
 use std::collections::hash_map::HashMap;
 use std::collections::hash_set::HashSet;
 use std::sync::{Arc, Mutex, MutexGuard};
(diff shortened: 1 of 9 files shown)


vm/src/vm.rs
```diff
@@ -56,23 +56,26 @@ use crate::sysmodule;
...
-    pub codec_registry: RefCell<Vec<PyObjectRef>>,
+    pub signal_handlers: Option<RefCell<[PyObjectRef; NSIG]>>,
+    pub state: Arc<PyGlobalState>,
     pub initialized: bool,
 }
 
+pub struct PyGlobalState {
+    pub settings: PySettings,
+    pub stdlib_inits: HashMap<String, stdlib::StdlibInitFunc>,
+    pub frozen: HashMap<String, bytecode::FrozenModule>,
```
palaviv, line 76: Maybe the frozen modules should be a global static? This should not change between VMs.

coolreader18: Maybe, but it's possible that someone would want to inject a frozen module into one VM to wrap their API or something, but want to use another vm for something different. I've done that myself in another project that uses RustPython.

PR #1934, opened 2020-05-17
Implement Py38 named expression (PEP 572)

TheAnyKey:
Python 3.8 instroduced named expressions or assignment expressions with PEP 572. Now it is possible 

>>> if f:=3>5: pass
or
>>> foo(e:=42)

Most of the CPython tests are already passing. There are some scoping issues with mutli level (e.g., when a comprehension is inside of a comprehension, the asignments from the inner one are not leaked to the scope outside of the outer comrpehension). Further, there are no error messages when (ab)using iterator variables in named expressions. Both are generic topicsin RustPython and require a bit of rework in the symboltable and some other parts. As the functionality is complete besides of the scope limitation I propose to accept this PR.


TheAnyKey:
Be carefull, when testing it in an existing workspace as the bytecode changed. Although, I set the new instruction to the end, such that it should have no interference with already compiled bytecode files, I cannot ensure it.

auvipy:
thanks for tackling this. codes look quite good

TheAnyKey:
Mostly done. Just need to add a few more checks, cleanup and merge. Some scoping issues will remain as they are an already existing RustPython problem. So we can fix that with a separate PR.

compiler/src/symboltable.rs
```diff
@@ -193,71 +200,137 @@ impl<'a> SymbolTableAnalyzer<'a> {
...
-                    if !symbols.contains_key(&symbol.name) || scope_depth < 2 {
+    fn analyze_symbol(
+        &mut self,
+        symbol: &mut Symbol,
+        curr_st_typ: SymbolTableType,
+    ) -> SymbolTableResult {
+        if symbol.is_assign_namedexpr_in_comprehension
+            && curr_st_typ == SymbolTableType::Comprehension
+        {
+            self.analyze_symbol_comprehension(symbol, 0)?
```
coolreader18, line 220: You could avoid putting the rest of the function in an indented block by just returning here, and leaving out the else:
```rust
if ... {
  return self.analyze_...()
}
match symbol.scope {...}
```

coolreader18:
@windelbouwman could you take a look at the symbol table changes?

windelbouwman:
Ahh, nice, the infamous walrus operator! :=

parser/src/python.lalrpop
```diff
@@ -335,7 +353,7 @@ IfStatement: ast::Statement = {
 };
 
 WhileStatement: ast::Statement = {
-    <location:@L> "while" <test:Test> ":" <body:Suite> <s2:("else" ":" Suite)?> => {
+    <location:@L> "while" <test:NamedExpression_Test> ":" <body:Suite> <s2:("else" ":" Suite)?> => {
```
windelbouwman, line 356: Thanks for sticking with the grammar as described here: https://docs.python.org/3/reference/grammar.html?highlight=grammar

compiler/src/symboltable.rs
```diff
@@ -99,6 +101,10 @@ pub struct Symbol {
     pub is_assigned: bool,
     pub is_parameter: bool,
     pub is_free: bool,
+
+    // indicates if the symbol gets a value assigned by a named expression in a comprehension
+    // this is required to correct the scope in the analysis.
+    pub is_assign_namedexpr_in_comprehension: bool,
```
windelbouwman, line 107: I guess this should be removed when the scoping is properly fixed? I cannot find the field here: https://docs.python.org/3/library/symtable.html?highlight=symboltable#symtable.SymbolTable

compiler/src/symboltable.rs
```diff
@@ -287,6 +361,7 @@ enum ExpressionContext {
     Load,
     Store,
     Delete,
+    NamedExprStore,
```
windelbouwman, line 364: Can't we use simply `Store` here? Why do we need an extra type?

Refer also here for the expression contexts: https://docs.python.org/3/library/ast.html?highlight=ast#module-ast

compiler/src/symboltable.rs
```diff
@@ -66,6 +66,7 @@ pub enum SymbolTableType {
     Module,
     Class,
     Function,
+    Comprehension,
```
windelbouwman, line 69: I would rather stick to the three options here:

https://docs.python.org/3/library/symtable.html?highlight=symboltable#symtable.SymbolTable.get_type

Still, it is weird that a comprehension is a function (which it is clearly not), but that's the way cpython did it ..

compiler/src/symboltable.rs
windelbouwman: While it would reduce the indentation, it also adds a return statement, and functions with multiple return statements are arguably harder to reason about.

My question is why the field `is_assign_namedexpr_in_comprehension` is required. I do not fully grasp this idea.

compiler/src/symboltable.rs
```diff
@@ -193,71 +200,137 @@ impl<'a> SymbolTableAnalyzer<'a> {
...
+            } else {
+                // If there are scopes above we can assume global.
+                symbol.scope = SymbolScope::Global;
             }
-            SymbolScope::Local => {
-                // all is well
+        }
+    }
+
+    fn analyze_symbol_comprehension(
```
windelbouwman, line 285: Could you explain the reasoning behind this function? I suspect it might be tricky to follow a different code path for symbols in named expressions, can the existing function `analyze_symbol` be modified to incorporate named expressions?

Could you maybe give an example python snippet why this function is required?

parser/src/python.lalrpop
TheAnyKey: When keeping Test and just extending it with named expressions, the grammar is too permisse and allows plentyof situations you need to find in a later analysis.  E.g., it is not allowed to use a named expression as expression statement, at least without braces. Thought, there  are other ways to achieve similarin the grammar, but for compatibility I would stick to that solution. 
For the the implementation, I consider [this grammar](https://github.com/python/cpython/blob/master/Grammar/python.gram) as more use helpful inspiration.
The naming sceme is violated.

compiler/src/symboltable.rs
TheAnyKey: There are some differences between the interface spec and the implementation. To achive compatiblity you need more context as the interface has. https://github.com/python/cpython/blob/7b78e7f9fd77bb3280ee39fb74b86772a7d46a70/Python/symtable.c#L1435 
There also some more flags that are not visible in the interface.

compiler/src/symboltable.rs
TheAnyKey: When you use a name the first time in an named expression in a comprehension you need to propagate the name to the next level outside of the outermost comprehension. Due to the generic multilevel scoping issue in rustpython it is currently not that helpfull (RustPytohn supports only the directly containing level, what would be be sufficient with that limitation - I added a [testcase for reference](https://github.com/TheAnyKey/RustPython/blob/e1d382d8df6423f2da9878d6b066e9fb579d9b82/Lib/test/test_named_expression.py#L399)). In that way the implementation is complete and will work immediately after solving this.
This is inspired by the extend scope function https://github.com/python/cpython/blob/7b78e7f9fd77bb3280ee39fb74b86772a7d46a70/Python/symtable.c#L1435
thought some aspects in the error handling are still missing. As this would require more context information, e.g., we distinguish iterator variables from others, what´s currently not implemented.

compiler/src/symboltable.rs
TheAnyKey: Thats unnecessary=>cleaned up

compiler/src/symboltable.rs
TheAnyKey: This is required for scope extension, see my other comments. There can more optimal solutions, but I guess this more reqork of the symtable. Some of the context information could be gathered directly but this would increase the runtime.

compiler/src/symboltable.rs
TheAnyKey: You can see and argument it in both directions, under pragmatic ocnsideration such as under dogmatic considerations. Usually have the tendency to use early returns in case of some pre checks and otherwise use nesting. I guess this still readable although the nesting level is high. And the most helpful optimization in terms of readablity is to move the whole else branch into a dedicated function.

compiler/src/symboltable.rs
windelbouwman: Okay, thank you for the explanation! Could we use the non-local in some way to fix this?

TheAnyKey: Hi Windel,
with the non-local stuff we can only fix the part bring in the variable from the outer scope for reading. Writing wont work, therefore also the scope.rs needs to be fixed. Currently when storing to a non-local it takes the directly containing scope and writes the variable to this. This is too simple and does not work in general. Instead you also need to search for the next scope that contains the name,  as when reading from a non-local. Doing this in the straight forward way is stronger than whats now implemented and should mostly be sufficient for the pre-3.8 stuff. For the named expressions, we need to take more context into account to implement the scope extension for so not used names in comprehensions

TheAnyKey: see #1951

PR #1936, opened 2020-05-17
Add _thread.start_new_thread

coolreader18:
~~Dependent on #1933 and #1935~~

### DIFF
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -71,7 +71,7 @@ smallbox = "0.8"
 bstr = "0.2.12"
 crossbeam-utils = "0.7"
 generational-arena = "0.2"
-parking_lot = { git = "https://github.com/Amanieu/parking_lot" }
+parking_lot = { git = "https://github.com/Amanieu/parking_lot" } # TODO: use published version
 
 ## unicode stuff
 unicode_names2 = "0.4"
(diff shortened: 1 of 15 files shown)


coolreader18:
`test_threading` does some of that already, but I think I could probably verify the ordering now/make sure that it's actually concurrent somehow.

coolreader18:
@palaviv I was actually able to get `test.test_thread` passing with not too much effort, so I think that should work for tests. I guess the `_thread` module really doesn't have that much API surface.

vm/Cargo.toml
```diff
@@ -71,7 +71,7 @@ smallbox = "0.8"
 bstr = "0.2.12"
 crossbeam-utils = "0.7"
 generational-arena = "0.2"
-parking_lot = { git = "https://github.com/Amanieu/parking_lot" }
+parking_lot = { git = "https://github.com/coolreader18/parking_lot", branch="raw-re-is_locked" }
```
palaviv, line 74: I was a bit iffy when we added the `parking_lot` from Github. For now at least add a TODO to change this.

coolreader18: That was just temporary, the PR for that is merged now. I'll still add a TODO for the official git version, though.

PR #1941, opened 2020-05-24
update rust version in readme

jamestwebber:
Had issues installing with 1.40.0 which the readme implies could work. Thanks @youknowone realized it was out of date, thought I'd update to reflect that.


coolreader18:
Looks good, thanks for making this more clear!

PR #1943, opened 2020-05-25
Switch to using parking_lot::Mutex instead of std::sync::Mutex

coolreader18:
All of the benefits of `parking_lot::Mutex` over the std one are listed in its
[readme](https://github.com/Amanieu/parking_lot#features), but the ones that apply
to us are copied here:

* Mutex and Once only require 1 byte of storage space, while Condvar and RwLock only require 1 word of storage space. On the other hand the standard library primitives require a dynamically allocated Box to hold OS-specific synchronization primitives. The small size of Mutex in particular encourages the use of fine-grained locks to increase parallelism.
  * less allocation is always good for performance, especially when all PyObjects are allocated and most require Mutexes
* Uncontended lock acquisition and release is done through fast inline paths which only require a single atomic operation.
  * good, since a lot of Python code will only be run single-threaded
* RwLock takes advantage of hardware lock elision on processors that support it, which can lead to huge performance wins with many readers.
  * requires nightly, but we could either build release binaries on nightly or wait until this is possible on stable
* RwLock uses a task-fair locking policy, which avoids reader and writer starvation, whereas the standard library version makes no guarantees.
  * not sure if this is super necessary but it would definitely benefit the GIL-less experiment we're doing here

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -275,7 +275,7 @@ fn extract_impl_items(mut items: Vec<ItemIdent>) -> Result<TokenStream2, Diagnos
                 #transform(Self::#item_ident)
             };
             Some(quote! {
-                (*class.slots.write().unwrap()).#slot_ident = Some(#into_func);
+                (*class.slots.write()).#slot_ident = Some(#into_func);
             })
         }
         _ => None,
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -3,6 +3,7 @@ use crate::pyhash;
 use crate::pyobject::{IdProtocol, IntoPyObject, PyObjectRef, PyResult};
 use crate::vm::VirtualMachine;
 use num_bigint::ToBigInt;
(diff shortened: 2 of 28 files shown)


palaviv:
I think we should do such a big change only after we are using a published stable release of `parking_lot`.

coolreader18:
That sounds good to me, I can put this on hold till then.

coolreader18:
This gave about an 11% speedup for me running the test suite (though it's like 8-9 minutes so I didn't run hyperfine or anything yet) and a 4%-8% speedup for the benchmarks in the benchmarks directory.

coolreader18:
Actually, I've been thinking about something for a while that this would be a good change to make with: we could make a wrapper type around `Arc/Mutex/RwLock` or `Rc/RefCell/RefCell`, to allow for if somebody wants the performance and doesn't want threading. Then we could do something like `PyRc`, `PyMutex`, and `PyRwLock` and have the inner types be just implementation details.

coolreader18:
@youknowone could you review this?

PR #1948, opened 2020-05-30
Add link to home page

palaviv:
I was trying to find the best place. Feel free to suggest other options.


coolreader18:
We could just edit the "Website" field for the github repository.

youknowone:
I think nothing bad to adding it, even if we could add it to github website field too.

coolreader18:
Alright, this LGTM then.

PR #1950, opened 2020-06-01
Implement bitcount for int (Py310 BPO-29882)

TheAnyKey:
Implement an efficient bitcount method directly for the type int. Bitcount sometimes refered as population count returns the number of 1 in an int of arbitrary length. In the implementation always the absolute value is takes.
This PR also includes the test from CPytohn for long.

https://bugs.python.org/issue29882

### DIFF
--- a/Lib/test/test_long.py
+++ b/Lib/test/test_long.py
@@ -313,7 +313,7 @@ class LongTest(unittest.TestCase):
             with self.subTest(got=got):
                 self.assertEqual(int(got, 0), x)
 
-    @unittest.expectedFailure # RustPython
+    @unittest.skip #Todo RustPython
     def test_format(self):
         for x in special:
             self.check_format_1(x)
--- a/Lib/test/test_long.py
+++ b/Lib/test/test_long.py
@@ -1,0 +1,1402 @@
+import unittest
+from test import support
+
+import sys
+
+import random
+import math
+import array
+
+
+# SHIFT should match the value in longintrepr.h for best testing.
+SHIFT = 32 #sys.int_info.bits_per_digit # TODO RustPython int_info not supported
+BASE = 2 ** SHIFT
+MASK = BASE - 1
+KARATSUBA_CUTOFF = 70   # from longobject.c
+
+# Max number of base BASE digits to use in test cases.  Doubling
+# this will more than double the runtime.
+MAXDIGITS = 15
+
+# build some special values
+special = [0, 1, 2, BASE, BASE >> 1, 0x5555555555555555, 0xaaaaaaaaaaaaaaaa]
+#  some solid strings of one bits
+p2 = 4  # 0 and 1 already added
+for i in range(2*SHIFT):
+    special.append(p2 - 1)
+    p2 = p2 << 1
+del p2
+# add complements & negations
+special += [~x for x in special] + [-x for x in special]
+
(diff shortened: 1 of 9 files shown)


youknowone:
This is nice work. For future tracking convenience, we are making a sole commit for importing cpython test with its version number. Will you seperate the cpython unittest commit? This PR will be a model work for that: https://github.com/RustPython/RustPython/pull/1949/commits

TheAnyKey:
Hi One,

whats the strategy for dealing with the different versions the tests originated from? Do we keep them on branches or in a parallel hierarchy for Lib/test? It think its worth to have them accessible all the time to see whats there and whats still missing in an automated manner.

youknowone:
There are no concrete policy about it. For now, I believe we are not making things strict.

For some test files, older cpython originated file is working with rustpython without huge effort but recent version one is not - due to lack of RustPython feature. Probably you've experienced similar things by working on compatibility issues.
So newer version is better - but older version is better than nothing.

Once we reach to good-enough state of compatibility, I hope to pin a specific cpython version of unittest.

vm/src/obj/objint.rs
```diff
@@ -653,6 +653,20 @@ impl PyInt {
...
+    fn bit_count(&self, vm: &VirtualMachine) -> PyResult {
+        Ok(vm.ctx.new_bigint(&BigInt::from(
+            self.value
+                .to_u32_digits()
+                .1
+                .iter()
+                .map(|n| n.count_ones())
+                .sum::<u32>(),
+        )))
+    }
```
coolreader18, line 669: You could just return a u32 here:

```suggestion
    fn bit_count(&self) -> u32 {
        self.value
            .to_u32_digits()
            .1
            .iter()
            .map(|n| n.count_ones())
            .sum()
    }
```

Lib/test/test_long.py
```diff
@@ -0,0 +1,1402 @@
+import unittest
```
coolreader18, line 1: What @youknowone was talking about was one commit that has only the copied file from CPython, and then another commit that modifies the file to add RustPython-specific changes. As an example, 68f5e2c156e55d38ead05ef9302d481723cf5f86 added some threading tests from cpython, and then 87a11fc0762a7eff3eeeb3d18cdc257f998fb78f added TODOs and skips to those tests.

coolreader18: This is probably fine for this PR -- we haven't been totally consistent in the past -- but try to do this in the future.

Lib/test/test_long.py
```diff
@@ -0,0 +1,1402 @@
...
+        # or not raise at all.
+        self.assertEqual(0 << sys.maxsize, 0)
+        self.assertEqual(0 << (sys.maxsize + 1), 0)
+
+    @support.cpython_only
+    @support.bigmemtest(sys.maxsize + 1000, memuse=2/15 * 2, dry_run=False)
+    def test_huge_lshift(self, size):
+        self.assertEqual(1 << (sys.maxsize + 1000), 1 << 1000 << sys.maxsize)
+
+    @unittest.expectedFailure # TODO RustPytohn
```
youknowone, line 959: typo of python

this is important to be searched by human for work. I will fix it for this time.

youknowone:
I rebased this PR before merging it

PR #1951, opened 2020-06-01
Fix scope issue for non-locals

TheAnyKey:
currently non-locals only support the direct outer scope. Whit that PR also the complete scopy hierarchy is supported.

### DIFF
--- a/Lib/test/test_named_expression.py
+++ b/Lib/test/test_named_expression.py
@@ -166,7 +166,7 @@ class NamedExpressionInvalidTest(unittest.TestCase):
                 with self.assertRaisesRegex(SyntaxError, msg):
                     exec(f"lambda: {code}", {}) # Function scope
 
-    @unittest.expectedFailure # TODO RustPython
+    @unittest.expectedFailure # TODO RustPythonskip
     def test_named_expression_invalid_comprehension_iterable_expression(self):
         cases = [
             ("Top level", "[i for i in (i := range(5))]"),
@@ -394,21 +394,21 @@ print(a)"""
         self.assertEqual(b, [1, 1])
         self.assertEqual(a, 1)
 
-    # the following test is not from CPyrgon and just as refernce for a common scoping problem of RustPython
-    @unittest.skip # needs skipping due to weired behaviour
+    # the following test is not from CPython and just as refernce for a common scoping problem of RustPython
     def test_named_expression_scop_10_rp_scope_prob(self):
         def foo():
-            rr=0
+            rr=36
             def foo0():
                 nonlocal rr
                 def foo1():
                     nonlocal rr
+                    self.assertEqual(rr,36)
                     rr+=42
                 foo1()
             foo0()
             return rr
 
-        self.assertEqual(foo(), 42)
+        self.assertEqual(foo(), 78)
         
 
     def test_named_expression_scope_11(self):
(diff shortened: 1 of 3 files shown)


Lib/test/test_named_expression.py
```diff
@@ -166,7 +166,7 @@ def test_named_expression_invalid_rebinding_comprehension_inner_loop(self):
                 with self.assertRaisesRegex(SyntaxError, msg):
                     exec(f"lambda: {code}", {}) # Function scope
 
-    @unittest.expectedFailure # TODO RustPython
+    @unittest.expectedFailure # TODO RustPythonskip
```
youknowone, line 169: is this a mistake?

Lib/test/test_named_expression.py
```diff
@@ -394,21 +394,21 @@ def test_named_expression_scope_10(self):
         self.assertEqual(b, [1, 1])
         self.assertEqual(a, 1)
 
-    # the following test is not from CPyrgon and just as refernce for a common scoping problem of RustPython
-    @unittest.skip # needs skipping due to weired behaviour
+    # the following test is not from CPython and just as refernce for a common scoping problem of RustPython
     def test_named_expression_scop_10_rp_scope_prob(self):
         def foo():
-            rr=0
+            rr=36
```
youknowone, line 400: is this changed by following original test code or original fix?

TheAnyKey: Thought, this is no CPython test, I added this test as reference for the existing scope issue (see comment to the test case). In the old verison it only tested, that data is propagated out of the inner scope. The new version tests also that data is also propagated to the inner scope. The snippets would maybe a better palce for own test cases, but as reference, until completion of the named expressions, I keep it here.

Lib/test/test_named_expression.py
TheAnyKey: That was not intended. I fix it when completing the stuff.

Lib/test/test_named_expression.py
youknowone: will you please move it to snippets if there is no reason to keep it in unittest directory?

youknowone:
nonlocal is a blocking issue for many unittests, so  i am really appreciated to see this PR!

pdc.sh
```diff
@@ -0,0 +1,27 @@
+RUSTPYTHONPATH=Lib
+export RUSTPYTHONPATH
+
```
coolreader18, line 3: What is this script? What does pdc stand for? This does seem useful, I'm just not sure exactly what it does.

TheAnyKey: PDC is an acronym for Pre-Delivery Check, as the intension is to run all checks before delivering (or raising a PR or commiting or what ever you want to use it for). 

Basically it executes all of the configured actions in the given order and reports which of them failed, without stopping in case of an fail. The actions in the given version are the RustPython CI steps plus a debug build in advance. So can run all of the checks in advance with one command.

Lib/test/test_named_expression.py
youknowone: Let me know when you complete things. Keeping it as unresolved will be helpful to not miss it before merging.

pdc.sh
youknowone: Will you put this comment on the top of this file as comment?

PR #1952, opened 2020-06-02
32 bit support

coolreader18:
I fixed a compile error and then a few tests that depended on oom-catching.

cc @aurelien-m

### DIFF
--- a/Lib/test/seq_tests.py
+++ b/Lib/test/seq_tests.py
@@ -321,6 +321,7 @@ class CommonTest(unittest.TestCase):
             self.assertEqual(self.type2test(s)*(-4), self.type2test([]))
             self.assertEqual(id(s), id(s*1))
 
+    @unittest.skip("TODO: RUSTPYTHON")
     def test_bigrepeat(self):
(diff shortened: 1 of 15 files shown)


coolreader18:
I was fixing one sequence issue, and I sorta just went all out on fixing other stuff that I'd been bothered by for a while `¯\_(ツ)_/¯`

Lib/test/seq_tests.py
```diff
@@ -321,6 +321,7 @@ def test_repeat(self):
             self.assertEqual(self.type2test(s)*(-4), self.type2test([]))
             self.assertEqual(id(s), id(s*1))
 
+    @unittest.skip("TODO: RUSTPYTHON")
```
youknowone, line 324: if this 32bit specific issue, adding 32bit note with RUSTPYTHON would be helpful

aurelien-m:
Thanks for looking into this :)
Seems to work perfectly on my Raspberry Pi running on Raspbian (or called Raspberry Pi OS now) 32bit!

PR #1953, opened 2020-06-02
Remove remaining usages of arg_check!

### LINKED ISSUE #1714: Replace the remaining arg_check! calls with parameter-based arg checking

limeburst:
This pull request removes the remaining usages of `arg_check!` and closes #1714.


limeburst:
I'm planning to look into parameter-based arg checking. Any pointers are appreciated!

limeburst:
I've reproduced the CI failure on my machine, will look into it!

vm/src/obj/objbool.rs
```diff
@@ -150,14 +150,17 @@ impl PyBool {
...
     #[pyslot]
-    fn tp_new(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
-        arg_check!(
-            vm,
-            args,
-            required = [(_zelf, Some(vm.ctx.type_type()))],
-            optional = [(val, None)]
-        );
-        let value = match val {
+    fn tp_new(vm: &VirtualMachine, mut args: PyFuncArgs) -> PyResult {
```
youknowone, line 153:
```suggestion
fn tp_new(zelf: PyObjectRef, value: PyObjectRef, vm: &VirtualMachine) -> PyResult
```

Unless there are keyword arguments, this form is more easy for most of case. In this way, you don't need to unpack zelf and value from args yourself.

limeburst: Is there another form I could use when I need to accept a zero-arguments case?

youknowone: `(args..., vm: &VirtualMachine) -> ReturnType` form work for everything.

These examples are actual RustPython code for zero-argument:
```rust
fn builtin_globals(vm: &VirtualMachine) -> PyResult<PyDictRef> { }
fn repr(self, vm: &VirtualMachine) -> PyResult<String> { }
```

If you are looking for optional arguments that would allow zero-arguments for users, check for `OptionalArg` for positional-only argument
```rust
fn sys_exit(code: OptionalArg<PyObjectRef>, vm: &VirtualMachine) -> PyResult
```
and search for `FromArgs` for keyword arguments.

limeburst: I tried using `OptionalArg` for `tp_new`, but it seems that it's not suitable for `tp_new` at the moment, since it allows keyword arguments to be passed as well. For example, when used in this form:

`fn tp_new(zelf: PyObjectRef, x: OptionalArg<PyObjectRef>, vm: &VirtualMachine)`

It fails the test `test_keyword_args` in `test_bool.py`.

coolreader18: Oh, that's fine, just remove the `unexpectedFailure` annotation from `test_keyword_args`.

limeburst: @coolreader18 I'm not sure if I used `OptionalArg` correctly, but I've added the related commits!

Lib/test/test_bool.py
```diff
@@ -171,7 +171,6 @@ def test_convert(self):
         self.assertIs(bool(), False)
 
     # TODO: RUSTPYTHON
```
coolreader18, line 173:
```suggestion
    # TODO: RUSTPYTHON
```

coolreader18:
Thanks for contributing!

vm/src/builtins.rs
```diff
@@ -716,11 +716,13 @@ fn builtin_setattr(
 // builtin_slice
 
 fn builtin_sorted(vm: &VirtualMachine, mut args: PyFuncArgs) -> PyResult {
-    arg_check!(vm, args, required = [(iterable, None)]);
-    let items = vm.extract_elements(iterable)?;
+    let iterable = args.take_positional();
+    if iterable.is_none() {
+        return Err(vm.new_type_error("sorted expected 1 arguments, got 0".to_string()));
+    }
+    let items = vm.extract_elements(&iterable.unwrap())?;
```
youknowone, line 723: Though it is already merge, we conventionally prefer to use this form to avoid unwrap when possible:
```suggestion
    let iterable = args.take_positional().unwrap_or_else(|| Err(vm.new_type_error("sorted expected 1 arguments, got 0".to_string())))?;
    let items = vm.extract_elements(&iterable)?;
```

PR #1954, opened 2020-06-02
Fix codecov collection

coolreader18:


coolreader18:
`panic=abort` breaks the `string_cache 0.7` build script, but once lalrpop updates it to `0.8` it doesn't have a build script anymore we could probably re-enable it.

PR #1957, opened 2020-06-03
Implement _json.encode_basestring{,_ascii}

coolreader18:


vm/src/vm.rs
```diff
@@ -1483,6 +1483,13 @@ impl VirtualMachine {
         attr_name: impl TryIntoRef<PyString>,
         attr_value: impl Into<PyObjectRef>,
     ) -> PyResult<()> {
+        // let attr_name = attr_name.try_into_ref(self)?;
```
palaviv, line 1486: Should this be here?

PR #1959, opened 2020-06-07
Share string constant data between str instances using Arc

coolreader18:
I know this is a lot of changes, I'd be willing to try to split into multiple PRs, though I'm not sure that would be possible.


palaviv:
Seems like this change should improve the performance. Can you add some numbers so we can see the improvment?

PR #1960, opened 2020-06-07
support setgroups in os module

mrmiywj:
As title, support setgroups
MACOS is not supported.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -36,8 +36,8 @@ use crate::obj::objstr::{PyString, PyStringRef};
 use crate::obj::objtuple::PyTupleRef;
 use crate::obj::objtype::PyClassRef;
 use crate::pyobject::{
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/os.rs
```diff
@@ -36,8 +36,8 @@ use crate::obj::objstr::{PyString, PyStringRef};
 use crate::obj::objtuple::PyTupleRef;
 use crate::obj::objtype::PyClassRef;
 use crate::pyobject::{
-    Either, ItemProtocol, PyClassImpl, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject,
-    TypeProtocol,
+    Either, ItemProtocol, PyClassImpl, PyIterable, PyObjectRef, PyRef, PyResult, PyValue,
```
youknowone, line 39: about the warnings, conditional import or use `crate::pyobject::PyIterable` below would remove it.

PR #1962, opened 2020-06-10
Implement subprocess using _posixsubprocess and _winapi

coolreader18:


vm/Cargo.toml
```diff
@@ -101,7 +102,6 @@ uname = "0.1.1"
 crc32fast = "1.2.0"
 adler32 = "1.0.3"
 gethostname = "0.2.0"
-subprocess = "0.2.2"
```
palaviv, line 104: I like the `subprocess` library. What are we missing in the library? The maintainer is very responsive so we should consider just asking him to add what we need.

youknowone: maybe due to platform compatibility like _posixsubprocess?

coolreader18: Yeah, it's generally more difficult to implement an API with a large surface area than it is to implement the low-level primitives it's based on. This also guarantees parity in the implementation, since there could be a subtle disparity between cpython's _posixsubprocess or windows subprocess implementation and the subprocess crate that causes some weird bug. This way we know that as long as the `fork_exec` or `CreateProcess` code (which are pretty small) are equivalent, then so should the `subprocess` module.

PR #1965, opened 2020-06-14
make "frame" module public

volodg:
I'd like to have access to local py script variables after execution embedded script like that:

```Rust
    let frame = Frame::new(code_obj, scope).into_ref(&vm);
    vm.frames.borrow_mut().push(frame.clone());
    let result = frame.run(&vm);
    //Access locals here !!!
    let locals = frame.scope.get_locals();
    vm.frames.borrow_mut().pop();
```

### DIFF
--- a/vm/src/lib.rs
+++ b/vm/src/lib.rs
@@ -62,7 +62,7 @@ mod dictdatatype;
 pub mod eval;
 pub mod exceptions;
 pub mod format;
-mod frame;
+pub mod frame;
 mod frozen;
 pub mod function;
 pub mod import;


volodg:
@coolreader18 @TheAnyKey Please review or give me advise is it feasible in any other way?

coolreader18:
There's a `get_locals()` method on vm that I think you can use to do what you're describing.

volodg:
@coolreader18 
```Rust
    let result = vm.run_code_obj(code_obj, scope)?;
    println!("locals2: {:?}", vm.get_locals());
```
gives
thread 'main' panicked at 'called current_scope but no frames on the stack', /Users/gorbv/.cargo/registry/src/github.com-1ecc6299db9ec823/rustpython-vm-0.1.1/src/vm.rs:255:21
stack backtrace:

volodg:
@coolreader18 Could you please with my this issue?

coolreader18:
Sorry @volodg; I think that's happening because run_code_object pops the frame off the stack after it runs. Looking at the existing code in `vm.rs`, there's a lot of pub methods that take or return a `Frame`, so I think it would be fine to expose it.

PR #1970, opened 2020-06-21
non-subprocess stuff from #1962

youknowone:
@coolreader18 is spliting non-subprocess part form #1962 ok?
@palaviv


coolreader18:
Yep, that's fine with me :)

PR #1972, opened 2020-06-21
Revert "Fix error with latest parking_lot version"

youknowone:
This reverts commit ca7d9b625a4c665b9c22f6d179d99365aa413c91.

@coolreader18 It seems this commit is breaking `cargo run`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -309,15 +309,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cloudabi"
-version = "0.1.0"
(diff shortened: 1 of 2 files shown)


coolreader18:
Oh, jeez, sorry, I figured since this was so simple I'd just push to master, but I forgot to put the `!` in `if !is_locked { return Err }`. Would it be alright if I just push an update fixing it?

coolreader18:
It should work now, sorry about this again. I think I'll just make a PR in the future, no matter what.

PR #1973, opened 2020-06-22
Test Annotation Framework for Imported CPython test

TheAnyKey:
When importing test currentlyfrom CPython,  they need to be annotated with comments like 
> \# TODO RustPython
or 
> \# TODO: RustPython

or how ever the author likes. Further the original file revision and the orignating python (language) revision is missing in most of the imported tests, what makes tracking nearly impossible. 
To improve the convention base "solution", I propose a python annotation framework using decorators to properly annotate the imported test cases, such that they can be tracked and evaluated in a comprehnesive and comfortable way.
Examples can be seen in Lib/test/test_rpt.py.

This the initial implementation. The core functionality is there and tested, but, e.g., the reporting is very rudimentary and more for demonstration purpoposes.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -75,8 +75,8 @@ When importing tests from CPython, please follow the following rules.
 1. Import complete files and document the version of the imported file and the major version this file relates to, e.g., 3.8.
 2. In any case, keep the name of the original tests and test cases unmodified, such that it can always be traced to the original one from CPython.
 3. Modify or substitute test cases, if required, to make them pass. We propose to
- - modify test cases only when threre is only a minor deviation from the original behaviour. E.g., when the text of an error message does not match.
- - substitute test cases when there are severe deviations from the original is required. E.g., when it is necessary to introduce new variables or change the logic of the test or split a test case in several test cases.
+ - modify test cases only when there is only a minor deviation from the original behaviour. E.g., when the text of an error message does not match.
+ - substitute test cases when severe deviations from the original is required. E.g., when it is necessary to introduce new variables or change the logic of the test or split a test case in several test cases.
 4. Add further test cases.
 5. Annotate each test case with:
  - Original and unmodified test case that is supposed to be passed: no annotation is required, but when you want to be more verbose you can use @rpt.imported.original
@@ -84,7 +84,7 @@ When importing tests from CPython, please follow the following rules.
  - Original and unmodified test case that is not executing, causes panic, breaks RustPython, etc.: @rpt.imported.skip
  - Original but modified test case: @rpt.imported.modified
  - Substituted test cases: @prt.imported.substituted (Note: this leads to skipping this test as it is considered as distirbung, by setting the parameter `run=True` it is executed but expected to fail.)
- - New test cases that substitue an original test case: @rpt.ours.subst and refer to the substituted original test cases.
+ - New test cases that substitute for an original test case: @rpt.ours.subst and refer to the substituted original test cases.
  - New test cases that extend an existing test case: @rpt.ours.extends and refers to the extended test case.
  - New test cases without direct relation to an original CPython test case: @rpt.ours.new
  
@@ -112,7 +112,6 @@ exists a raw html viewer which is currently broken, and we welcome a PR to fix i
 
 Understanding a new codebase takes time. Here's a brief view of the
 repository's structure:
-.
 - `bytecode/src`: python bytecode representation in rust structures
 - `compiler/src`: python compilation to bytecode
 - `derive/src`: Rust language extensions and macros specific to rustpython
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -46,7 +46,16 @@ Python code should follow the
 
 ## Testing
 
-To test RustPython's functionality, a collection of Python snippets is located
+RustPython uses three classes of tests. First, rust unit tests, second, our 
+own tests written in python, and last, tests imported from CPython. 
+
+Rust unit tests can be run with `cargo`:
+
+```shell
+$ cargo test --all
+```
+
+Our own collection of tests, called snippets, to test RustPython's functionality is located
 in the `tests/snippets` directory and can be run using `pytest`:
 
 ```shell
@@ -54,11 +63,32 @@ $ cd tests
 $ pytest -v
 ```
 
-Rust unit tests can be run with `cargo`:
+A set of annotated CPython tests are located in `Lib/tests`. Execute them by
 
 ```shell
-$ cargo test --all
+$ cargo run --release -- -m test
 ```
+
+
+### Creating tests
+When importing tests from CPython, please follow the following rules.
+1. Import complete files and document the version of the imported file and the major version this file relates to, e.g., 3.8.
+2. In any case, keep the name of the original tests and test cases unmodified, such that it can always be traced to the original one from CPython.
+3. Modify or substitute test cases, if required, to make them pass. We propose to
+ - modify test cases only when threre is only a minor deviation from the original behaviour. E.g., when the text of an error message does not match.
+ - substitute test cases when there are severe deviations from the original is required. E.g., when it is necessary to introduce new variables or change the logic of the test or split a test case in several test cases.
+4. Add further test cases.
+5. Annotate each test case with:
+ - Original and unmodified test case that is supposed to be passed: no annotation is required, but when you want to be more verbose you can use @rpt.imported.original
+ - Original and unmodified test case that is supposed to be failed: @rpt.imported.fail
+ - Original and unmodified test case that is not executing, causes panic, breaks RustPython, etc.: @rpt.imported.skip
+ - Original but modified test case: @rpt.imported.modified
+ - Substituted test cases: @prt.imported.substituted (Note: this leads to skipping this test as it is considered as distirbung, by setting the parameter `run=True` it is executed but expected to fail.)
+ - New test cases that substitue an original test case: @rpt.ours.subst and refer to the substituted original test cases.
+ - New test cases that extend an existing test case: @rpt.ours.extends and refers to the extended test case.
+ - New test cases without direct relation to an original CPython test case: @rpt.ours.new
+ 
+
 
 ## Profiling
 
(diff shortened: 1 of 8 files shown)


Lib/test/test_rpt.py
```diff
@@ -0,0 +1,151 @@
+
+
+import unittest
+import Lib.test.librptest as rpt
```
coolreader18, line 4:
```suggestion
import test.librptest as rpt
```

Lib/test/librptest/rptest.py
```diff
@@ -0,0 +1,51 @@
...
+url_regex = re.compile(
+        r'^(?:http|ftp)s?://' # http:// or https://
+        r'(?:(?:[A-Z0-9](?:[A-Z0-9-]{0,61}[A-Z0-9])?\.)+(?:[A-Z]{2,6}\.?|[A-Z0-9-]{2,}\.?)|' #domain...
+        r'localhost|' #localhost...
+        r'\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})' # ...or ip
+        r'(?::\d+)?' # optional port
+        r'(?:/?|[/?]\S+)$', re.IGNORECASE)
+
+
+def originated_from(major_version, minor_version, original_file_link):
```
coolreader18, line 21: Maybe this could take a patch version as well?

Lib/test/librptest/ours.py
```diff
@@ -0,0 +1,32 @@
+
+import types
+
+from .core import mark_test_case_or_reason, get_marked_test_cases, add_subst, get_substitutions as core_get_subs
+
+
+'''
+Front End for own test cases in imported test files.
+'''
```
coolreader18, line 9: Docstrings have to be the first statement of a file for it to register; could you move this to the top?

Lib/test/librptest/imported.py
```diff
@@ -0,0 +1,37 @@
...
+import unittest
+from .core import decorate_test_case_with_reason, mark_test_case_or_reason, get_marked_test_cases
+
+class OriginalMarker:pass
+class ModifiedMarker:pass
+
+
+'''
+Front end for RPT for imported test cases.
+'''
```
coolreader18, line 13: Same for this docstring

Lib/test/test_complex.py
```diff
@@ -0,0 +1,712 @@
+import unittest
+import Lib.test.librptest as rpt
```
coolreader18, line 2:
```suggestion
import test.librptest as rpt
```
We won't always be running from the root of the repo, which would put `Lib` under the pythonpath as a submodule, but `Lib` should already be in `RUSTPYTHONPATH`, so you don't need it here.

coolreader18 (commented):
I really like this idea! It would at least allow us to mark which tests should actually need work on our side in the test results webpage. @palaviv @youknowone what do you think?

Lib/test/librptest/rptest.py
TheAnyKey: The information here relates to the Python language version, the file itself can be from a micro CPython version. This is covered by the link. But the relevant information for the compatibility if it is 3.7 or 3.8 or 3.9.

Lib/test/test_complex.py
TheAnyKey: obviously...

palaviv:
I love this idea!

DEVELOPMENT.md
```diff
@@ -46,20 +46,50 @@ Python code should follow the
...
+ - modify test cases only when threre is only a minor deviation from the original behaviour. E.g., when the text of an error message does not match.
+ - substitute test cases when there are severe deviations from the original is required. E.g., when it is necessary to introduce new variables or change the logic of the test or split a test case in several test cases.
+4. Add further test cases.
+5. Annotate each test case with:
+ - Original and unmodified test case that is supposed to be passed: no annotation is required, but when you want to be more verbose you can use @rpt.imported.original
+ - Original and unmodified test case that is supposed to be failed: @rpt.imported.fail
+ - Original and unmodified test case that is not executing, causes panic, breaks RustPython, etc.: @rpt.imported.skip
+ - Original but modified test case: @rpt.imported.modified
+ - Substituted test cases: @prt.imported.substituted (Note: this leads to skipping this test as it is considered as distirbung, by setting the parameter `run=True` it is executed but expected to fail.)
+ - New test cases that substitue an original test case: @rpt.ours.subst and refer to the substituted original test cases.
```
jamestwebber, line 87:
```suggestion
 - New test cases that substitute for an original test case: @rpt.ours.subst and refer to the substituted original test cases.
```

DEVELOPMENT.md
```diff
@@ -46,20 +46,50 @@ Python code should follow the
...
+$ cargo run --release -- -m test
 ```
 
+
+### Creating tests
+When importing tests from CPython, please follow the following rules.
+1. Import complete files and document the version of the imported file and the major version this file relates to, e.g., 3.8.
+2. In any case, keep the name of the original tests and test cases unmodified, such that it can always be traced to the original one from CPython.
+3. Modify or substitute test cases, if required, to make them pass. We propose to
+ - modify test cases only when threre is only a minor deviation from the original behaviour. E.g., when the text of an error message does not match.
```
jamestwebber, line 78:
```suggestion
 - modify test cases only when there is only a minor deviation from the original behaviour. E.g., when the text of an error message does not match.
```

DEVELOPMENT.md
```diff
@@ -46,20 +46,50 @@ Python code should follow the
...
+ - substitute test cases when there are severe deviations from the original is required. E.g., when it is necessary to introduce new variables or change the logic of the test or split a test case in several test cases.
```
jamestwebber, line 79:
```suggestion
 - substitute test cases when severe deviations from the original is required. E.g., when it is necessary to introduce new variables or change the logic of the test or split a test case in several test cases.
```

DEVELOPMENT.md
```diff
@@ -46,20 +46,50 @@ Python code should follow the
...
+4. Add further test cases.
+5. Annotate each test case with:
+ - Original and unmodified test case that is supposed to be passed: no annotation is required, but when you want to be more verbose you can use @rpt.imported.original
+ - Original and unmodified test case that is supposed to be failed: @rpt.imported.fail
+ - Original and unmodified test case that is not executing, causes panic, breaks RustPython, etc.: @rpt.imported.skip
+ - Original but modified test case: @rpt.imported.modified
+ - Substituted test cases: @prt.imported.substituted (Note: this leads to skipping this test as it is considered as distirbung, by setting the parameter `run=True` it is executed but expected to fail.)
```
jamestwebber, line 86:
```suggestion
 - Substituted test cases: @prt.imported.substituted (Note: this leads to skipping this test as it is considered as distirbung, by setting the parameter `run=True` it is executed but expected to fail.)
```
Not sure what "distirbung" is meant to say

Lib/test/librptest/core.py
```diff
@@ -0,0 +1,76 @@
...
+def add_subst(fct, subst, reason=None):
+    if fct==None:
+        return
+    
+    if fct in substitutions:
+        substitutions[fct].add((subst, reason))
+    else:
+        substitutions[fct]=set([(subst, reason)])
+
+### Returns True if a fucntion is substituted otherwise False.
```
jamestwebber, line 22:
```suggestion
### Returns True if a function is substituted otherwise False.
```

Lib/test/librptest/core.py
```diff
@@ -0,0 +1,76 @@
...
+
+### Adds a function as a substitute of the original function fct.
+def add_subst(fct, subst, reason=None):
+    if fct==None:
+        return
+    
+    if fct in substitutions:
+        substitutions[fct].add((subst, reason))
+    else:
+        substitutions[fct]=set([(subst, reason)])
```
jamestwebber, line 20: Is there a reason not to use `defaultdict(set)`?

Lib/test/librptest/core.py
```diff
@@ -0,0 +1,76 @@
...
+
+### Returns True if a fucntion is substituted otherwise False.
+def is_subsituted(fct):
+    return fct in substitutions
+
+### Returns a list of functions that are beeing  used as substitutes of the
```
jamestwebber, line 26:
```suggestion
### Returns a list of functions that are being used as substitutes of the 
```

Lib/test/librptest/core.py
```diff
@@ -0,0 +1,76 @@
...
+    if fct==None:
+        return
+    
+    if fct in substitutions:
+        substitutions[fct].add((subst, reason))
+    else:
+        substitutions[fct]=set([(subst, reason)])
+
+### Returns True if a fucntion is substituted otherwise False.
+def is_subsituted(fct):
```
jamestwebber, line 23:
```suggestion
def is_substituted(fct):
```

Lib/test/librptest/core.py
```diff
@@ -0,0 +1,76 @@
...
+### original function fct. Does not fail and returns an empty set if the function
+### has no known substitutes.
+def get_substitutions(fct):
+    try:
+        return substitutions[fct]
+    except:
+        return set()
+
+### Add a function that is marked as beeing substituted 
+### When the function is alread known, nothing happens.
```
jamestwebber, line 36:
```suggestion
### Add a function that is marked as being substituted 
### When the function is already known, nothing happens.
```

jamestwebber (commented):
Wanted to read this to see the new scheme and found some typos

DEVELOPMENT.md
```diff
@@ -82,7 +112,7 @@ exists a raw html viewer which is currently broken, and we welcome a PR to fix i
 
 Understanding a new codebase takes time. Here's a brief view of the
 repository's structure:
-
+.
```
youknowone, line 115:
```suggestion
```

Lib/test/librptest/core.py
TheAnyKey: Yes, because in RustPython is crashes. I will raise a ticket for that...

youknowone:
I agree marking things more than comments is a good idea.

palaviv (approved):
Can we push this in @coolreader18, @youknowone?

coolreader18:
Yeah, lgtm. There might be more skips/expectedFailures to replace with this since this pr was opened, but we can do that as a follow-up

youknowone:
I strongly prefer this idea than current state.

Not sure librptest has to be placed in `Lib/test`. I basically regard keeping this directory same to CPython as possible as make easier to manage it in future. (Though it is ok if this is technical constraint)

I don't agree `rpt.originated_from` usage for class. We probably will not compose a test file with multiple class from different CPython versions. It is better to be placed for the file rather than the class.

coolreader18:
:+1: for the last paragraph @youknowone

PR #1975, opened 2020-06-30
Build dependencies with opt-level 3 always

youknowone:
https://doc.rust-lang.org/cargo/reference/profiles.html#overrides

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,9 @@ version = "0.2"
 name = "rustpython"
 path = "src/main.rs"
 
+[profile.dev.package."*"]
+opt-level = 3
+
 [patch.crates-io]
 # REDOX START, Uncommment when you want to compile/check with redoxer
 # # following patches are just waiting on a new version to be released to crates.io
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -29,7 +29,7 @@ jobs:
               ~/.cargo/registry
               ~/.cargo/git
               target
-          key: ${{ runner.os }}-debug-opt3-${{ hashFiles('Cargo.lock') }}
+          key: ${{ runner.os }}-debug_opt3-${{ hashFiles('**/Cargo.lock') }}
       - name: run rust tests
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -45,6 +45,9 @@ version = "0.2"
 name = "rustpython"
 path = "src/main.rs"
 
+[profile.dev.package."*"]
+opt-level = 3
```
palaviv, line 49: Wouldn't this slow down the `dev` build?

coolreader18: Speaking for myself, but I always build with `--release`. I think that dev builds (with the default opt-level 0, at least) are just too slow, especially with the reports of defaulting with dev binaries because maybe calls don't get inlined or something.

coolreader18: Though maybe opt-level 2 would be better for debug builds, since 3 makes there no difference between debug/release

youknowone: Yes, it will slow building dependencies (which rarely build again) but boost running speed.
For the building speed, dependencies are usually the same and we only rebuild rustpython part. So only the first build will be slowed down but we will always get the benefits of running speed.
Because we usually only need to debug RustPython part, I think this will only cause speedup of running speed of dependencies and (mostly) nothing to lose (Though this option must be edited when we also need to debug dependencies - which is not very ideal)

I don't see any reason not to use the same option for debug/release build. They are not the part of RustPython so they must be blackbox for usual use case. So using same option is actually better because users anyway will use them only with release build. But it is considerable If build speed of opt-level 2 is a lot better than 3.

coolreader18: Yeah, maybe we could benchmark it somehow -- like `touch */src/lib.rs && cargo build`, and see how big the difference is.

coolreader18:
It's about a 2.25 minute difference with a clean build; 5 minutes 30 seconds opt-level 2 vs 7 minutes 45 seconds opt-level 3. It's basically the same time for both running `touch */src/lib.rs && cargo build` (1 minute), but I think that might be because it knows it hasn't really changed at all.

youknowone:
I checked CI runtime for debug build and it is even faster than before
image

We will lose time for first build for cache but gain time for cached running. It looks like good deal now regardless of opt2.

@coolreader18 how do you think now?

coolreader18:
LGTM 😁

youknowone:
I changed workflow a bit. wasm build now run after rust_tests to reuse cache from it and python lint is merged to rustfmt and clippy job.

PR #1976, opened 2020-06-30
Config files for debugging and launch for VS Code and Gitpod

TheAnyKey:
This PR provides the configuration files and updates the existing ones to allow launching and debugging in VS Code and Gitpod/Theia.

### DIFF
--- a/.theia/settings.json
+++ b/.theia/settings.json
@@ -1,0 +1,12 @@
+{
+    "cpp.buildConfigurations": [
+        {
+            "name": "",
+            "directory": ""
+        },
+        {
+            "name": "",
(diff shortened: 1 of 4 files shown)


.theia/settings.json
```diff
@@ -0,0 +1,12 @@
+{
+    "cpp.buildConfigurations": [
+        {
+            "name": "",
+            "directory": ""
```
coolreader18, line 5: Are these necessary?

TheAnyKey: One of them are needed otherwise I was able to debug. I removed the second entry.

PR #1977, opened 2020-07-03
Fix str.isspace to support unicode space characters

skinny121:
This new behavior matches the Python documentation https://docs.python.org/3/library/stdtypes.html#str.isspace.


skinny121:
Thanks, I see that lots of work has been done :smile:.

Issue #1979, opened 2020-07-04
Payload conflicts


skinny121:
Currently you are allowed to have the following:
```python
class Weird(str, int):
    pass
```
But this causes issues using using instances of this class as both `str` and `int` expect the payload to be `PyString` and `PyInt` respectively, but the object can only have a single payload.
```python
>>>>> 2 + Weird()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: Unexpected payload for type "Weird"
```

CPython deals with this by raising an error during class construction time:
```python
>>> class Weird(str, int):
...     pass

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: multiple bases have instance lay-out conflict
```

I feel we should do a similar thing by detecting when bases have different payloads.

This also comes up in `__class__` assignment, the following is CPython behavior:
```python
>>> class A(str): pass
... 
>>> class B(int): pass
... 
>>> A().__class__ = B
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __class__ assignment: 'B' object layout differs from 'A'

```

coolreader18:
We could probably do something with a `TypeId` stored in `PyClass`

arihant2math:
This no longer occurs:
```
> python tmp.py
Traceback (most recent call last):
  File "D:\Documents\Programming\GitHub\RustPython\tmp.py", line 1, in <module>
    class Weird(str, int):
        pass
TypeError: multiple bases have instance lay-out conflict

> ./target/release/rustpython tmp.py
Traceback (most recent call last):
  File "tmp.py", line 1, in <module>
    class Weird(str, int):
TypeError: multiple bases have instance layout conflict
```

youknowone:
@arihant2math Thank you for checking!

Issue #1568, opened 2019-10-26
class conversion is not possible


HyeockJinKim:
## Feature
Converting __class__ in Python to change to another class is not currently supported in Rustpython.

### Actual Result
```python
>>>>> class A:
.....  pass
..... 
>>>>> class B:
.....  pass
..... 
>>>>> a = A()
>>>>> a.__class__ = B
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: can't change class of type '<class 'A'>'
```

### Expected result
```python
>>> class A:
...  pass
... 
>>> class B:
...  pass
... 
>>> a = A()
>>> a.__class__ = B
>>> a.__class__
<class '__main__.B'>
```

youknowone:
```
>>>>> class A: pass
>>>>> class B: pass
>>>>> a = A()
>>>>> a.__class__ = B
>>>>> a.__class__
<class 'B'>
```

PR #1981, opened 2020-07-04
Support assigning to __class__

skinny121:
Supports the following:
```python
a.__class__ = B
```
Note at the moment this incorrectly allows, but this is a wider issue(#1979 ):
```python
a.__class__ = int
```

I used https://docs.rs/arc-swap/0.4.7/arc_swap/, for both the `typ` and `dict` fields within `PyObject` as it seems like that would have a slightly lower overhead than a `Mutex`. A benchmark with just the `dict` mutex replaced with `ArcSwap` showed a small improvement. Even with that, this PR has caused a regression of around 25% :disappointed:, due to the `typ` field being accesed quite a lot(I think).


vm/src/pyobject.rs
```diff
@@ -642,9 +643,8 @@ pub struct PyObject<T>
 where
     T: ?Sized + PyObjectPayload,
 {
-    pub typ: Arc<PyObject<PyClass>>,
-    // TODO: make this RwLock once PyObjectRef is Send + Sync
-    pub(crate) dict: Option<Mutex<PyDictRef>>, // __dict__ member
+    pub(crate) typ: ArcSwap<PyObject<PyClass>>, // __class__ member
+    pub(crate) dict: Option<ArcSwap<PyObject<PyDict>>>, // __dict__ member
```
coolreader18, line 647: Very useful type! I was thinking about using a sort of `AtomicArcCell` for these fields, but I wasn't sure it would be able to be fully atomic; looks like it is!

coolreader18:
re: accessing `typ` a lot, maybe there could be `pub(crate)` method that returns an `arc_swap::Guard` so that internal method lookup doesn't have to go through the cost of `load_full()`

vm/src/pyobject.rs
skinny121: It suits out use case pretty well as it's optimised for reading, which is the operation that we really care about here.

coolreader18:
Or I suppose the field is already `pub(crate)` you could just do `obj.typ.load()`

skinny121:
I have been able to reduce the performance regression of this PR down to just under 10% by using `PyLease`(wrapper around `arc_swap::Guard`) in a lot of places. Is this an acceptable regression? Otherwise I will have to look again where I can improve performance around this.

coolreader18:
What's the performance regression like if you use `IndependentArcSwap`? That might be better, since it shouldn't have to deal with multiple locks on ALL `__class__` fields at once

skinny121:
That actually made it worse :cry:. The benchmark results for `bench_rustpy_mandelbrot` are as follows:

master:                        1,473,571,865 ns/iter (+/- 9,875,746)
ArcSwap:                     1,562,154,302 ns/iter (+/- 8,891,651)
IndependentArcSwap: 1,691,609,695 ns/iter (+/- 14,180,686)

I'm a little bit surprised by the amount these change are having on the overall performance.

coolreader18:
I guess 3+ atomic ops on every method lookup compared to just an offset of the pyobject does have a decent impact. Kinda just curious now, but how does this compare to a `parking_lot::RwLock`? If that's easy to do for you. Maybe these "optimizations" are just more work than locking an uncontended RwLock (which is just one atomic op, I think). Idk, if not, the regression is probably fine. Maybe we could have cargo features to disable obscure python features like this in order to get a performance speedup.

skinny121:
RwLock test bench_rustpy_mandelbrot 1,318,745,820 ns/iter (+/- 5,858,271) :tada:

The above includes the changes I have made within this PR that reduces the amount of clones that we perform, that is why it's faster than no locking. I'm surprised that the RwLock performs so well.

coolreader18:
Under the hood, it looks like arc-swap has their own sort of parking lot with shards and stuff, but that naive implementation with a bunch of work done on each lock + only a limited number of total locks can't beat parking_lot with its 1 atomic op for uncontended locking and parking lot for determining contention.

skinny121:
This is ready for review now.

vm/src/obj/objtype.rs
```diff
@@ -363,18 +372,40 @@ pub(crate) fn init(ctx: &PyContext) {
...
-    issubclass(&obj.class(), &cls)
+    issubclass(obj.lease_class(), &cls)
 }
 
 /// Determines if `subclass` is actually a subclass of `cls`, this doesn't call __subclasscheck__,
 /// so only use this if `cls` is known to have not overridden the base __subclasscheck__ magic
 /// method.
-pub fn issubclass(subclass: &PyClassRef, cls: &PyClassRef) -> bool {
-    subclass.iter_mro().any(|c| c.is(cls))
+pub fn issubclass<T: DerefToPyClass + IdProtocol, R: IdProtocol>(subclass: T, cls: R) -> bool {
```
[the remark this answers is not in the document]
skinny121: The `subclass` parameter needs to know it's identity for the case: `issubclass(A, A)`, which `&PyClass` doesn't know as that's just the payload.

youknowone:
#1568

PR #1982, opened 2020-07-04
Fix issues with wasm code missing _thread

coolreader18:


Lib/functools.py
```diff
@@ -18,7 +18,10 @@
 from collections import namedtuple
 # import types, weakref  # Deferred to single_dispatch()
 from reprlib import recursive_repr
-from _thread import RLock
+try:
+    from _thread import RLock
```
youknowone, line 22: Adding this kind of try-except import for everywhere doesn't look ideal.
If RUSTPYTHONPATH work with multiple items, how about adding semething like `WasmLib` with `_thread.py`?

coolreader18: Oh, or maybe even just a `_thread.py` in Lib? The built-in module takes precedence over a module in sys.path, but then when it's not there it's falls back to the dummy implementation.

youknowone: Loading dummy implementation for wasm sounds ok but it sounds like error out of wasm. For example, if we suggest module editing for embedding use case, its behavior will be weird.
It is ok if we can ensure `_thread` module existence is always guaranteed.

coolreader18: Actually, I just remembered the use case I actually had for this :facepalm: -- compiling to WASI. That doesn't have threading either, yet it just uses the standard rustpython executable, and it's more likely that code running in a posix-ish environment would assume that "`_thread` present == threading available". Ultimately, I don't think `functools` or `reprlib` really **need** threading, this is mostly just because some data structures they use aren't thread safe, and it was probably an oversight that they didn't fallback to `_dummy_thread` (since I don't think `no-threading` is very commonly used for CPython).

coolreader18: E.g. `tempfile` imports `_dummy_thread` as a fallback as well.

Issue #1175, opened 2019-07-24
Extend the os module


windelbouwman:
Currently, the python `os` module is missing many methods. Try adding a method to this module by editing the file `vm/src/stdlib/os.rs`.

This is an issue for new contributors, so please do not implement everything at once, so other people have a chance to go for it (or, actually, please implement all! >:)! ).

Note: Before working on platform-specific functions, check if they are members of `posix` or `nt`. Their members are reexported to `os`.

mpajkowski:
Ok, I'm working on get/set uid, guid etc. stuff

immersinn:
I'm going to take a stab at `chown` and `chroot`. 

Also, as a general question, is the current goal to match [CPython functionality here?](https://github.com/python/cpython/blob/3.8/Lib/os.py)

coolreader18:
Because Rust's cross-platform support is so much better than C's, we decided to implement a single `_os` module rather than the separate `posix` and `nt` modules that CPython has.

immersinn:
Ah, gotcha..so for `chown` and `chroot` methods, i'd reference the functionality and behavior in `posixmodule.c`

coolreader18:
Yep, and you can use the `#[cfg(unix)]` attribute to make sure they're only available on unix.

immersinn:
cool cool, thanks!

dan-fritchman:
PR #1524 has an implementation of `os.system`.

metaphorshear:
I was thinking of doing `wait` and `fork` if those are still available.

(Also, a question of etiquette: is it better to call dibs on things here, or just do them?)

CircleOnCircles:
I would like to do a quick summary of what is API still missing.
1. `os.ctermid`
2. `os. environb`
3. `os.fchdir`
4. `os.fsencode`
5. `os.fsdecode`
6. `os.PathLike`
7. `os.getenvb`
8. `os.get_exec_path`
9. `os.getgrouplist`
10. `os.getgroups` *partially*
11. `os.getlogin`
12. `os.getpgrp`
13. `os.getpriority`

... 

I just scroll thought 10% on the [os docs](https://docs.python.org/3/library/os.html). There are still so many interfaces left to implement. 😛

nicholasraphael:
Hi, new here. I was just wondering how to start working on one of these outstanding implementations without having any overlap with others? Should we open a new issue for what we want to tackle and then have it assigned to us?

Thanks!

[...]

PR #1984, opened 2020-07-05
Implementation of chown, lchown and fchown for the os module

LeBoucEtMistere:
Hello all,

I am trying to implement the chown function for the os module (https://docs.python.org/3/library/os.html#os.chown), because of issue #1175 .
I am a new contributor and I'd like to have feedback on this first take at the implementation. It is not finished yet but I have a few questions and I wanted to be sure that I am going in the good direction.

As of now, the chown function works properly given that you invoke rustpython with root permissions. I can successfully change uid and gid of a file, or use -1 to preserve them. I am having a few issue with the path to file, it does not seem yet to accept unix shorthands like ~ (FileNotFound error, might come from nix lib, need to investigate).
I have not tested dir_fd and _follow_symlink yet since I am not sure how to handle these arguments with defaults values.
More particularly, in line 1988, I am not sure to understand how to initialise a new SupportFunc, what it actually is and why it takes Options on bool rather than bool. I took example on other functions and used false in the Options but I think it should be true since this function supports use of file descriptors and following symlinks, unless I did not understand properly what the bool means here.

I am planning on implementing lchown and fchown after this by using to this first chown function.

I am planning to write proper unit tests but I have no clue on how to write them and what you are expecting as proper tests.

Finally, in the python doc, it says that these functions raises AuditingEvents. I am not familiar with this, is it supported in rustpython yet ? Should I care about this ?

Thanks for your feedback

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1088,6 +1088,20 @@ fn os_chown(
         .map_err(|err| convert_nix_error(vm, err))
 }
 
+#[cfg(all(unix, not(target_os = "redox")))]
+fn os_lchown(path: PyPathLike, uid: PyIntRef, gid: PyIntRef, vm: &VirtualMachine) -> PyResult<()> {
+    os_chown(
+        path,
+        uid,
+        gid,
+        DirFd { dir_fd: None },
+        FollowSymlinks {
+            follow_symlinks: false,
+        },
+        vm,
+    )
+}
+
 #[cfg(unix)]
 fn os_get_inheritable(fd: RawFd, vm: &VirtualMachine) -> PyResult<bool> {
     use nix::fcntl::fcntl;
@@ -2131,6 +2145,7 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         #[cfg(not(target_os = "redox"))]
         SupportFunc::new(vm, "chroot", os_chroot, Some(false), None, None),
         SupportFunc::new(vm, "chown", os_chown, None, Some(true), Some(true)),
+        SupportFunc::new(vm, "lchown", os_lchown, None, None, None),
         SupportFunc::new(vm, "umask", os_umask, Some(false), Some(false), Some(false)),
     ]);
     let supports_fd = PySet::default().into_ref(vm);
@@ -2280,6 +2295,7 @@ fn extend_module_platform_specific(vm: &VirtualMachine, module: &PyObjectRef) {
     #[cfg(not(target_os = "redox"))]
     extend_module!(vm, module, {
         "chown" => ctx.new_function(os_chown),
+        "lchown" => ctx.new_function(os_lchown),
         "chroot" => ctx.new_function(os_chroot),
         "getsid" => ctx.new_function(os_getsid),
         "setsid" => ctx.new_function(os_setsid),
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -2130,7 +2130,7 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         SupportFunc::new(vm, "chmod", os_chmod, Some(false), Some(false), Some(false)),
         #[cfg(not(target_os = "redox"))]
         SupportFunc::new(vm, "chroot", os_chroot, Some(false), None, None),
-        SupportFunc::new(vm, "chown", os_chown, Some(false), Some(false), Some(false)),
+        SupportFunc::new(vm, "chown", os_chown, None, Some(true), Some(true)),
         SupportFunc::new(vm, "umask", os_umask, Some(false), Some(false), Some(false)),
     ]);
     let supports_fd = PySet::default().into_ref(vm);
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1074,10 +1074,12 @@ fn os_chown(
         return Err(vm.new_os_error(String::from("Specified gid is not valid.")));
     };
 
-    let flag = match follow_symlinks.follow_symlinks {
-        true => nix::unistd::FchownatFlags::FollowSymlink,
-        false => nix::unistd::FchownatFlags::NoFollowSymlink,
+    let flag = if follow_symlinks.follow_symlinks {
+        nix::unistd::FchownatFlags::FollowSymlink
+    } else {
+        nix::unistd::FchownatFlags::NoFollowSymlink
     };
+
     let dir_fd: Option<std::os::unix::io::RawFd> = match dir_fd.dir_fd {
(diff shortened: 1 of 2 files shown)


LeBoucEtMistere:
Up ? Is any seasoned contributor available to answer my questions and help me get this PR in a mergeable state ? :)

vm/src/stdlib/os.rs
```diff
@@ -1027,6 +1027,47 @@ fn os_chroot(path: PyStringRef, vm: &VirtualMachine) -> PyResult<()> {
...
+    } else if gid == -1 {
+        None
+    } else {
+        return Err(vm.new_os_error(String::from("Specified gid is not valid.")));
+    };
+
+    let flag = match follow_symlinks.follow_symlinks {
+        true => nix::unistd::FchownatFlags::FollowSymlink,
+        false => nix::unistd::FchownatFlags::NoFollowSymlink,
+    };
```
coolreader18, line 1062: This should fix the error in CI.
```suggestion
    let flag = if follow_symlinks.follow_symlinks {
        nix::unistd::FchownatFlags::FollowSymlink
    } else {
        nix::unistd::FchownatFlags::NoFollowSymlink
    };
```

coolreader18:
`SupportFunc`s just determine what functions are in the `supports_{fd,dir_fd,follow_symlinks}` sets. For chown, it looks like it supports `dir_fd` and `follow_symlinks`, so you can put `None, Some(true), Some(true)`. I think the Options are because some functions are supposed to support one of them but they don't yet, so we put `None` for that as a sort of TODO. I think.

coolreader18:
Oh, and yeah, you can ignore the auditing events stuff.

LeBoucEtMistere:
I fixed the clippy lint.

> `SupportFunc`s just determine what functions are in the `supports_{fd,dir_fd,follow_symlinks}` sets. For chown, it looks like it supports `dir_fd` and `follow_symlinks`, so you can put `None, Some(true), Some(true)`. I think the Options are because some functions are supposed to support one of them but they don't yet, so we put `None` for that as a sort of TODO. I think.

I also fixed this, thanks. Do you think it could be usefull to refactor this weird Option<bool> into a cleaner enum then ? Something like 
```rust
enum SupportStatus {
    SUPPORTED,
    TODO,
    NOT_SUPPORTED,
}
```
I'll gladly open a PR for this if you think this is useful.

LeBoucEtMistere:
Finally, I'd feel more confident if I wrote at least unit tests for this piece of code. I can write some rust tests on the rust function but I think that it misses the point and I'd rather write python tests where I actually call the function os.chown from python code (liek integration testing)
There was nothing in the CONTRIBUTING.md file regarding the tests, what do you expect and where are these kind of tests located in the codebase ? I feel like it's important to write them as we add more functions to the python library to make sure there are no regressions whatsoever.

coolreader18:
Yeah, I think the 3-enum rather than the `Option<bool>` would definitely make it more clear.

re: unit testing; there is a `test_os` module in CPython that we could copy over, but it's ~4000 lines and with just a naive copy over there's 95 failing tests. You could write a few tests in `tests/snippets/stdlib_os.py`, though, maybe take inspiration from the [chown tests in CPython](https://github.com/python/cpython/blob/af4eda46d1538b1da700a86588bdb94b0a4d1ff2/Lib/test/test_os.py#L1515)?

LeBoucEtMistere:
Ok the CPython tests are a nice resource indeed ! I'll try to draw inspiration from these for sure, and maybe add a few more test cases.

LeBoucEtMistere:
Regarding the issue I mentioned in my original comment regarding the tilde expansions, it seems that neither Rust nor Python do it, so I won't add support for it.

coolreader18:
Ah, yeah, I think you have to manually opt in to that with `os.path.expandpath(p)`

LeBoucEtMistere:
Just wrote a first few tests (mostly copied from CPython test suite) and I have a few things to say:
- Overall the way testing is done for snippets (Lib) seems super messy and counterintuitive to me. I rather prefer the CPython test suite organisation with clear separation between function being tested, separation of tests cases, and so on. The strange custom runner you're using is not making a lot of sense to me atm so I'll gladly take any clarifications on this.
- Partly because of the first point, I have not a lot of trust in the tests I wrote since there is very little observability on what is running or not inside the tests files. Mainly on the fact that for chown the test suite needs to be started with and without sudo permissions to test both behaviours. Again, if you have any idea or clarification on that, I'll gladly hear them.

Regarding the CI failing, it seems that with rust stable bumping to 1.45 today, clippy is now more restrictive on certain things and code that was previously considered correct is no longer. It should eventually be fixed in another PR.

Regarding the sister functions of chown (lchown, fchown, ...), I will add them so please don't close this PR now.

Regarding the adding of an enum for SupportFuncs, I will move this to another PR to keep things clean.

As I am a newcomer on this project and that PR got me to take a look around a few parts of it (namely the standard lib and the integration tests), I will write down my take on this "experience" and on the entry barrier to new contributors **as I perceived it**. The architecture of the Lib reimplementation could be improved by creating rust module for each library module and creating a single file for each function in this module. That way, code would be cleaner, it would reduce conflicts, it would allow us to write clean rust unit tests in each file to test only a function, without cluttering an already too large file. (eg the os file is already more than 2300 loc and we are far from having everything implemented)
My second point refers back to the test suites and more particularly the snippets. I feel like everything is over-complex and sub-performant. The test files are a mess with no clear organisation and separation. It's like a gigantic plate of spaghettis.
Of course, this is an open source project with limited contributing workforce and time and I get these points are not a priority. But as  a newcomer, I faced them almost instantly and I understand it can discourage people from making polished and well-tested contributions. I would then happily discuss these points with maintainers and open the needed issues to try and improve the codebase if this makes sense as well as the developer's guidelines. I believe that properly testing the Lib as we add pieces to it is the best way to go, that it is fun to do, and that it is a good way to bring in new contributors with a myriad of self-contained PRs that would encourage best coding practices, mix rust and python and quickly obtain a nice and satisfactory finished new feature.
(All the paragraph above is my **personal** view on the subject, with the eyes of a newcomer, it might be totally wrong and subject to change)

coolreader18:
Yeah, the test snippets aren't very well organized, partly because we don't need them to be; the idea is that we just have a bunch of snippets that `assert` some basic functionality, and eventually we'll try to rely more on the CPython test suite to ensure that we have parity. I definitely do understand that it might be confusing for new contributors, though, we should explain that more in `tests/README.md`. Just know that we know that they're a mess, but our main focus w.r.t testing is getting the `Lib/test` suite to work.

coolreader18:
Also, the clippy lints should be fixed on master now.

LeBoucEtMistere:
I just added support for fd and sister functions fchown and lchown, I think that the PR is mergeable now. Let me know if there is anything to alter still.

PR #1985, opened 2020-07-06
Add annotation setup

### LINKED ISSUE #1980: Class attribute type annotations
## Feature

The ability to annotate a type on a class attribute. For example:

```python
class Employee:
    name: str
    id: int
```

This runs without error with CPython, but raises the following error with RustPython:

```
  File "test.py", line 3, in <module>
    id: int
  File "test.py", line 2, in Employee
    name: str
NameError: name '__annotations__' is not defined
```

## Python Documentation

https://docs.python.org/3/library/typing.html#typing.NamedTuple

Lynskylate:
Fix #1980 
Add setup annotation for compile class

### DIFF
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -984,15 +984,14 @@ impl<O: OutputStream> Compiler<O> {
         Ok(())
     }
 
-    fn option_stmt_to_bool(&mut self, stmts: &Option<ast::Suite>) -> bool {
-        match &stmts {
-            Some(stmts) => self.find_ann(stmts),
-            None => false,
-        }
-    }
-
-    fn find_ann(&mut self, body: &[ast::Statement]) -> bool {
+    fn find_ann(&self, body: &[ast::Statement]) -> bool {
         use ast::StatementType::*;
+        let option_stmt_to_bool = |suit: &Option<ast::Suite>| -> bool {
+            match suit {
+                Some(stmts) => self.find_ann(stmts),
+                None => false,
+            }
+        };
 
         for statement in body {
             let res = match &statement.node {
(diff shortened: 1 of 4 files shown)


compiler/src/compile.rs
```diff
@@ -984,6 +984,64 @@ impl<O: OutputStream> Compiler<O> {
...
     }
 
+    fn option_stmt_to_bool(&mut self, stmts: &Option<ast::Suite>) -> bool {
+        match &stmts {
+            Some(stmts) => self.find_ann(stmts),
+            None => false,
+        }
+    }
+
+    fn find_ann(&mut self, body: &[ast::Statement]) -> bool {
```
coolreader18, line 994: I don't think this needs to be `&mut self`:
```suggestion
    fn find_ann(&self, body: &[ast::Statement]) -> bool {
```

compiler/src/compile.rs
```diff
@@ -984,6 +984,64 @@ impl<O: OutputStream> Compiler<O> {
         Ok(())
     }
 
+    fn option_stmt_to_bool(&mut self, stmts: &Option<ast::Suite>) -> bool {
+        match &stmts {
+            Some(stmts) => self.find_ann(stmts),
+            None => false,
+        }
+    }
```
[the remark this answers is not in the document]
Lynskylate: Done, thank you.
Before, I mistakenly made self mut, which caused can't use closure function.

compiler/src/compile.rs
Lynskylate: Done, Thank you

vm/src/frame.rs
```diff
@@ -456,6 +456,13 @@ impl ExecutingFrame<'_> {
                 Ok(Some(ExecutionResult::Yield(value)))
             }
             bytecode::Instruction::YieldFrom => self.execute_yield_from(vm),
+            bytecode::Instruction::SetupAnnotation => {
+                let locals = self.scope.get_locals();
+                if let Err(_e) = locals.get_item("__annotations__", vm) {
```
coolreader18, line 461:
```suggestion
                if !locals.contains_key("__annotations__") {
```

coolreader18:
Thanks for contributing!

PR #1989, opened 2020-07-08
Set annotation for function args

Lynskylate:
Set annotation for other type function args.
```python        
def f(a, b: 1, c: 2, d, e: 3 = 4, f: int = 5, /, *g: 6, h: 7, i=8, j: 9 = 10,
              **k: 11) -> 12: pass
# before, we just set b,c,d,e,f in __annotations__
self.assertEqual(f.__annotations__,{'b': 1, 'c': 2, 'e': 3, 'f': int, 'g': 6, 'h': 7, 'j': 9, 'k': 11, 'return': 12})
```

### DIFF
--- a/Lib/test/test_grammar.py
+++ b/Lib/test/test_grammar.py
@@ -297,7 +297,6 @@ class GrammarTests(unittest.TestCase):
         # testlist ENDMARKER
         x = eval('1, 0 or 1')
 
-    @unittest.skip("TODO: RUSTPYTHON")
     def test_var_annot_basics(self):
         # all these should be allowed
         var1: int = 5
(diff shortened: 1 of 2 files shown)


Lib/test/test_grammar.py
```diff
@@ -297,7 +297,6 @@ def test_eval_input(self):
         # testlist ENDMARKER
         x = eval('1, 0 or 1')
 
-    @unittest.skip("TODO: RUSTPYTHON")
     def test_var_annot_basics(self):
         # all these should be allowed
         var1: int = 5
```
coolreader18, line 302: It looks like this is failing in CI

Lynskylate: Has anyway to check a annotation assign in a function scope?
 [Annotations are only evaluated in a module or class.](https://github.com/python/cpython/blob/master/Python/compile.c#L5182)

Lynskylate:
@coolreader18  Please merge it， Thank you.

coolreader18:
Whoops, sorry about that @Lynskylate, I think I probably wanted to wait for CI to be done just to be sure, and then it slipped my mind and there wasn't any more notifications about it.

Lynskylate:
> Whoops, sorry about that @Lynskylate, I think I probably wanted to wait for CI to be done just to be sure, and then it slipped my mind and there wasn't any more notifications about it.

It's fine. Thanks for your review and reply.

Issue #1983, opened 2020-07-05
Support pip


skinny121:
To support pip the following are required:

- [x]  `__class__` assignment(#1981)
- [ ]  `plistlib` which requires `xml.parsers.expat`
- [x]  Fix `shutil.get_terminal_size`
- [x]  `itertools.groupby`
- [x]  Fix `str.__add__` raising `TypeError` instead of returning `NotImplemented`
- [x]  Fix `"{x.attr}".format(...)`

If the above issues are worked/hacked around, then the pip help text is successfully displayed :tada::
```
ben@BenLinux:~/Hobby/RustPython$ RUSTPYTHONPATH=Lib:local_packages/ cargo run --release local_packages/pip 
    Finished release [optimized] target(s) in 0.12s
     Running `target/release/rustpython local_packages/pip`

Usage:   
  pip <command> [options]

Commands:
     install Install packages.
    download Download packages.
   uninstall Uninstall packages.
...
```

coolreader18:
The thing about `str.__add__` is that it's actually the slot `sq_concat`, that gets mapped to the wrapper method `__add__`. When we do `a + b`, what needs to happen is:

1. Check if a has the slot `nb_add`/magic method `__add__`
1. Check if b has `nb_add` (I think?) or `__radd__`
1.  Then, look at sq_concat for a (and maybe b if it doesn't have it, but I don't think so).

coolreader18:
Also, I've looked into binding to expat, and it might be easier to just skip importing plistlib for the time being, since I think it's used in a very niche way in pip.

jamestwebber:
Whoa is this all? I was checking the dependencies as described in [this post](https://doughellmann.com/blog/2019/06/29/dependencies-between-python-standard-library-modules/) and it seemed a lot hairier. I would guess that to actually `pip install` will require many more things working.

skinny121:
Yes that everything(I think), we have got most of core standard library importable at this point. I'm also expecting `pip install` to require more changes than this.

mireille-raad:
is this done?

infopensource:
How can i install this fucking pip!!!!!!!!!!

mireille-raad:
@infopensource  does this help? https://asciinema.org/a/3pAZSfrFeIbEUzpPyXtvYZCeN

PR #1990, opened 2020-07-09
Implement os.get_terminal_size

coolreader18:
Related to #1983


youknowone:
is there any test about this function?

coolreader18:
Maybe test_shutil? If pip uses it, it would probably be good to have it tested.

youknowone:
I made 2 more tests skipped for macOS

PR #1993, opened 2020-07-12
CompilerError always contains source_path

youknowone:


compiler/src/compile.rs
```diff
@@ -2377,16 +2372,24 @@ fn compile_conversion_flag(conversion_flag: ast::ConversionFlag) -> bytecode::Co
...
+    use super::{CompileOpts, Compiler, OutputStream};
     use crate::symboltable::make_symbol_table;
     use rustpython_bytecode::bytecode::Constant::*;
     use rustpython_bytecode::bytecode::Instruction::*;
     use rustpython_bytecode::bytecode::{CodeObject, Label};
     use rustpython_parser::parser;
 
+    impl<O> Default for Compiler<O>
+    where
+        O: OutputStream,
```
coolreader18, line 2384: Maybe just have this be a function, rather than only impl-ing Default when a cfg matches

PR #1994, opened 2020-07-13
Fix file reading failure in non-ascii path on `not_impl_gen.py`

BasixKOR:
```
===== MODULES =====
[2020-07-13T14:46:47Z ERROR rustpython] Failed reading file 'tests/snippets/whats_left_modules.py': InvalidData
```

Broken unicode characters in libdir

I discovered a issue when launching `./whats_left.sh` fails to read modules. Turns out it was because of non-ASCII path, so I created this small fix to solve this.

The issue is only confirmed on a Windows 10 machine with non-unicode codepage.

### DIFF
--- a/tests/not_impl_gen.py
+++ b/tests/not_impl_gen.py
@@ -104,7 +104,6 @@ def get_module_methods(name):
             return None
         except Exception as e:
             print("!!! {} skipped because {}: {}".format(name, type(e).__name__, str(e)))
-1
 
 def gen_modules(header, footer, output):
     output.write(header.read())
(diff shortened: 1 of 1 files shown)


tests/not_impl_gen.py
```diff
@@ -104,7 +104,7 @@ def get_module_methods(name):
             return None
         except Exception as e:
             print("!!! {} skipped because {}: {}".format(name, type(e).__name__, str(e)))
-
+1
```
coolreader18, line 107: ?
```suggestion

```

coolreader18:
Thanks for contributing!

PR #1995, opened 2020-07-14
Add fractions.py

skinny121:
Importing `fractions.py` required getting `divmod.__name__`. I achieved this by converting most of the `builtin` module to be declared with `#[pymodule]`, and then changed the proc macro to use `new_function_named`. As a consequence a lot more builtin functions now will have `__name__` and `__module__` correctly set.

### DIFF
--- a/derive/src/pymodule.rs
+++ b/derive/src/pymodule.rs
@@ -115,7 +115,10 @@ impl Module {
     }
 }
 
-fn extract_module_items(mut items: Vec<ItemIdent>) -> Result<TokenStream2, Diagnostic> {
+fn extract_module_items(
+    mut items: Vec<ItemIdent>,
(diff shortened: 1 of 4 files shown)


derive/src/pymodule.rs
```diff
@@ -132,7 +135,10 @@ fn extract_module_items(mut items: Vec<ItemIdent>) -> Result<TokenStream2, Diagn
             item_ident,
             py_name,
         } => {
-            let new_func = quote_spanned!(item_ident.span() => .new_function(#item_ident));
+            let new_func = quote_spanned!(item_ident.span() =>
+                                                      .new_function_named(#item_ident,
```
youknowone, line 139: :+1:

PR #1999, opened 2020-07-15
Add itertools.groupby

### LINKED ISSUE #1361: Implement rest of itertools API

skinny121:
The test `TestBasicOps.test_groupby` is still failing due to a general pickling bug.

For #1983
Closes #1361 (this is the last itertools function to be implemented)

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -5,9 +5,11 @@ mod decl {
     use crossbeam_utils::atomic::AtomicCell;
     use num_bigint::BigInt;
     use num_traits::{One, Signed, ToPrimitive, Zero};
+    use std::fmt;
     use std::iter;
 
-    use crate::common::cell::{PyRwLock, PyRwLockWriteGuard};
+    use crate::common::cell::{PyMutex, PyRwLock, PyRwLockWriteGuard};
+    use crate::common::rc::{PyRc, PyWeak};
     use crate::function::{Args, OptionalArg, OptionalOption, PyFuncArgs};
     use crate::obj::objbool;
     use crate::obj::objint::{self, PyInt, PyIntRef};
@@ -15,10 +17,10 @@ mod decl {
     use crate::obj::objtuple::PyTuple;
     use crate::obj::objtype::{self, PyClassRef};
     use crate::pyobject::{
-        IdProtocol, PyCallable, PyClassImpl, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
+        IdProtocol, PyCallable, PyClassImpl, PyObject, PyObjectRef, PyRef, PyResult, PyValue,
+        TypeProtocol,
     };
     use crate::vm::VirtualMachine;
-    use rustpython_common::rc::PyRc;
 
     #[pyclass(name = "chain")]
     #[derive(Debug)]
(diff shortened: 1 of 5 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -488,6 +490,196 @@ mod decl {
...
+                .field("current_value", &self.current_value)
+                .field("current_key", &self.current_key)
+                .field("next_group", &self.next_group)
+                .finish()
+        }
+    }
+
+    impl GroupByState {
+        fn is_current(&self, grouper: &PyItertoolsGrouperRef) -> bool {
+            if let Some(ref current_weak_grouper) = self.grouper {
```
youknowone, line 512: I personally prefer not to make simple value (`false` here) tailing chain with deep indent but no matter. (actually I was about to try to convince with below code but lost confidence after watching the result)
```rust
self.grouper.map_or(false, |ref weak| weak.upgrade().map_or(false, |current| grouper.is(&current))
```

vm/src/stdlib/itertools.rs
```diff
@@ -488,6 +490,196 @@ mod decl {
...
+    struct GroupByState {
+        current_value: Option<PyObjectRef>,
+        current_key: Option<PyObjectRef>,
+        next_group: bool,
+        grouper: Option<Weak<PyObject<PyItertoolsGrouper>>>,
+    }
+
+    impl fmt::Debug for GroupByState {
+        fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+            f.debug_struct("PyItertoolsGroupBy")
```
youknowone, line 502: is this intended naming or mistake of `GroupByState`?

vm/src/stdlib/itertools.rs
```diff
@@ -488,6 +490,196 @@ mod decl {
...
+                    Ok(value)
+                } else {
+                    let mut state = zelf.groupby.state.lock();
+                    state.current_value = Some(value);
+                    state.current_key = Some(key);
+                    state.next_group = true;
+                    state.grouper = None;
+                    Err(new_stop_iteration(vm))
+                }
+            }
```
youknowone, line 674: This is a question because I am not used to rust convention of lock lifetime.

I want to know which one more preferred of explicit lock dropping or scope control like:
```rust
            let old_key = {
                let mut state = zelf.groupby.state.lock();

                if !state.is_current(&zelf) {
                    return Err(new_stop_iteration(vm));
                }

                if let Some(val) = state.current_value.clone() {
                    state.current_value = None;
                    return Ok(val);
                }
                
                // Shouldn't mutex over call to advance as that executes arbitrary Python code
                state.current_key.clone().unwrap();
            };

            let (value, key) = zelf.groupby.advance(vm)?;
            if vm.bool_eq(key.clone(), old_key)? {
                Ok(value)
            } else {
                let mut state = zelf.groupby.state.lock();
                state.current_value = Some(value);
                state.current_key = Some(key);
                state.next_group = true;
                state.grouper = None;
                Err(new_stop_iteration(vm))
            }
```

vm/src/stdlib/itertools.rs
```diff
@@ -488,6 +490,196 @@ mod decl {
...
+        fn next(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult {
+            let mut state = zelf.groupby.state.lock();
+
+            if !state.is_current(&zelf) {
+                return Err(new_stop_iteration(vm));
+            }
+
+            if let Some(val) = state.current_value.clone() {
+                state.current_value = None;
+                Ok(val)
```
Kroisse, line 658: A redundant clone can be reduced like this:

```rust
if let Some(val) = state.current_value.take() {
    Ok(val)
} else {
```

youknowone:
This is not a serious question. I want to know this is a port of CPython impl or original implementation. It looked a bit different at a glance.

vm/src/stdlib/itertools.rs
skinny121: That does seem a lot nicer.

vm/src/stdlib/itertools.rs
```diff
@@ -488,6 +490,196 @@ mod decl {
         }
     }
 
+    struct GroupByState {
+        current_value: Option<PyObjectRef>,
+        current_key: Option<PyObjectRef>,
+        next_group: bool,
```
youknowone, line 496: Is this better than having key in grouper like CPython?

vm/src/stdlib/itertools.rs
```diff
@@ -488,6 +490,196 @@ mod decl {
...
+                false
+            }
+        }
+    }
+
+    #[pyclass(name = "groupby")]
+    struct PyItertoolsGroupBy {
+        iterable: PyObjectRef,
+        key_func: Option<PyObjectRef>,
+        state: Mutex<GroupByState>,
```
youknowone, line 528: 1. Is Mutex prefered to RwLock here? (I didn't check details well, so it could be a stupid question)
2. I thought about state design that checking initial state by optional `current_key` is a good idea or not. `Option<GroupByState>` probably express empty initial state the best, but it seems it is a duplicated option wrapping for most of use case. I want to listen your opinion about those designs.

vm/src/stdlib/itertools.rs
```diff
@@ -488,6 +490,196 @@ mod decl {
...
+
+            state.next_group = false;
+
+            let grouper = PyItertoolsGrouper {
+                groupby: zelf.clone(),
+            }
+            .into_ref(vm);
+
+            state.grouper = Some(Arc::downgrade(&grouper.clone().into_typed_pyobj()));
+            Ok((state.current_key.clone().unwrap(), grouper))
```
youknowone, line 613: This unwrapping implies curernt_key is always `Some`. In that case, I suggest to use `unwrap().clone()` rather than `clone().unwrap()` because the intention is cloning the key object rather than the `Option` itself.

skinny121:
I based this implementation off the implementation within the documentation: https://docs.python.org/3.8/library/itertools.html#itertools.groupby, but replaced `self.tgtkey` with `next_group` as for me that made more logical sense to model it as a state machine.  

This does produce the same result in normal circumstances, but does produce slightly different behaviour when next is called recursively, e.g.:
```python
class EvilIter(object):
    def __init__(self):
        self.gb = None
        self.i = 0

    def __iter__(self):
        return self

    def __next__(self):
        self.i += 1
        if self.i == 1:
            key, values = next(self.gb)
            return 5
        else:
            return self.i

it = EvilIter()
gb = itertools.groupby(it, lambda x: x // 10)

it.gb = gb
key, values = next(it)
```

I don't think CPython's behaviour is perfect as it can skips an element, but it probably better to be consistent with CPython. I will can the implementation to be more inline with CPython does it.

youknowone:
@skinny121 Do you want to fix it or merge now as it is?

skinny121:
@youknowone I will fix that, I'm just tossing up how much we want to be compatible with CPython when it comes to edge cases like the above. I tried to change it to replicate CPython but it meant that we had to acquire the mutexs in more places which didn't seem nice.

vm/src/stdlib/itertools.rs
skinny121: After thinking about it more, there are actually 3 or 4 different states that it can be in so maybe the following is a better way to model it:
```rust
enum GroupByState {
    Initial,
    InGroup {
        key: PyObjectRef,
        value: Option<PyObjectRef>,
        grouper: Option<Weak<PyObject<PyItertoolsGrouper>>>,
    },
    NextGroup {
        next_key: PyObjectRef,
        next_value: PyObjectRef,
    },
}
```

PR #2000, opened 2020-07-16
Fix compilation on wasi

coolreader18:


youknowone:
Is this testable with CI?

coolreader18:
Yeah, I could probably add it as part of the cron workflow that runs once a week.

PR #2001, opened 2020-07-16
Update the pwd module

coolreader18:
Removing the `pwd` dependency removes about 6 transitive dependencies, including failure-derive and synstructure. This should improve our compile time a fair bit.

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -124,7 +124,7 @@ fn inner_pow(int1: &BigInt, int2: &BigInt, vm: &VirtualMachine) -> PyResult {
         objfloat::float_pow(v1, v2, vm).into_pyobject(vm)
     } else {
         Ok(if let Some(v2) = int2.to_u64() {
-            vm.ctx.new_int(int1.pow(v2))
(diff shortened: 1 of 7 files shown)


vm/src/obj/objint.rs
```diff
@@ -124,7 +124,7 @@ fn inner_pow(int1: &BigInt, int2: &BigInt, vm: &VirtualMachine) -> PyResult {
         objfloat::float_pow(v1, v2, vm).into_pyobject(vm)
     } else {
         Ok(if let Some(v2) = int2.to_u64() {
-            vm.ctx.new_int(int1.pow(v2))
+            vm.ctx.new_int(Pow::pow(int1, v2))
```
youknowone, line 127: is this related to bignum 0.3?

coolreader18: Yeah, it adds a pow method directly on BigInt that takes a u32; I'm not sure why.

coolreader18:
I also implemented pwd.getpwall, which wasn't before, and made sure to an extent that there won't be race conditions with multiple threads, which the pwd crate didn't do.

PR #2002, opened 2020-07-16
Implement `sys.getwindowsversion` 

### LINKED ISSUE #1996: module 'sys' has no attribute 'getwindowsversion'
## Feature
`getwindowsversion` is Windows-specific Python function which returns the Windows version currently running.

### Note
```
  File ".\Lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File ".\Lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File ".\Lib\test\__main__.py", line 1, in <module>
    from test.libregrtest import main
  File ".\Lib\test\libregrtest\__init__.py", line 4, in <module>
    from test.libregrtest.cmdline import _parse_args, RESOURCE_NAMES, ALL_RESOURCES
  File ".\Lib\test\libregrtest\cmdline.py", line 4, in <module>
    from test import support
  File ".\Lib\test\support\__init__.py", line 942, in <module>
    if sys.getwindowsversion().platform >= 2:
AttributeError: module 'sys' has no attribute 'getwindowsversion'
```

This error was found while running the CPython unit tests via `cargo run --release -- -m test`.

## Python Documentation
https://docs.python.org/3.8/library/sys.html#sys.getwindowsversion

BasixKOR:
This PR fixes #1996.

# Note
>In Windows 8.1 and Windows 10, the GetVersion and GetVersionEx functions have been deprecated. In Windows 10, the VerifyVersionInfo function has also been deprecated. While you can still call the deprecated functions, if your application does not specifically target Windows 8.1 or Windows 10, the functions will return the Windows 8 version (6.2). - [Targeting your application for Windows](https://docs.microsoft.com/en-us/windows/win32/sysinfo/targeting-your-application-at-windows-8-1)

* Since Windows 8.1 and later versions return 6.2 for non-manifested application, for now, this function will report the Windows version as 6.2 on modern systems. This can be fixed by adding a manifest with `embed_resource` crate.
* `platform_version` returns the result from `GetVersionExW`, which is not accurate when running application in compatibility mode.

### DIFF
--- a/tests/snippets/stdlib_os.py
+++ b/tests/snippets/stdlib_os.py
@@ -491,3 +491,26 @@ if "win" not in sys.platform:
 
     for arg in [None, 1, 1.0, TabError]:
         assert_raises(TypeError, os.system, arg)
+
+if sys.platform.startswith("win"):
+	winver = sys.getwindowsversion()
+
+	# the biggest value of wSuiteMask (https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-osversioninfoexa#members).
+	all_masks = 0x00000004 | 0x00000400 | 0x00004000 | 0x00000080 | 0x00000002 | 0x00000040 | 0x00000200 | \
+		0x00000100 | 0x00000001 | 0x00000020 | 0x00002000 | 0x00000010 | 0x00008000 | 0x00020000
+
+	# We really can't test if the results are correct, so it just checks for meaningful value
+	assert winver.major > 0
+	assert winver.minor >= 0
+	assert winver.build > 0
+	assert winver.platform == 2
+	assert isinstance(winver.service_pack, str)
+	assert 0 <= winver.suite_mask <= all_masks
+	assert 1 <= winver.product_type <= 3
+
+	# XXX if platform_version is implemented correctly, this'll break on compatiblity mode or a build without manifest
+	assert winver.major == winver.platform_version[0]
+	assert winver.minor == winver.platform_version[1]
+	assert winver.build == winver.platform_version[2]
+
+
--- a/vm/src/sysmodule.rs
+++ b/vm/src/sysmodule.rs
@@ -5,6 +5,7 @@ use crate::builtins;
 use crate::frame::FrameRef;
 use crate::function::{Args, OptionalArg, PyFuncArgs};
 use crate::obj::objstr::PyStringRef;
+use crate::obj::objtuple::PyTupleRef;
 use crate::pyhash::PyHashInfo;
 use crate::pyobject::{
     IntoPyObject, ItemProtocol, PyClassImpl, PyContext, PyObjectRef, PyResult, TypeProtocol,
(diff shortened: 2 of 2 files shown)


vm/src/sysmodule.rs
```diff
@@ -222,6 +224,78 @@ fn sys_displayhook(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
...
+    service_pack: String,
+    service_pack_major: u16,
+    service_pack_minor: u16,
+    suite_mask: u16,
+    product_type: u8,
+    platform_version: (u32, u32, u32),
+}
+
+#[cfg(windows)]
+fn sys_getwindowsversion(vm: &VirtualMachine) -> PyResult<PyTupleRef> {
```
youknowone, line 244: `PyResult<crate::obj::objtuple::PyTupleRef>` is also possible to avoid conditional `use` statement. But current state is also good.

vm/src/sysmodule.rs
```diff
@@ -222,6 +224,78 @@ fn sys_displayhook(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
...
+    use std::ffi::OsString;
+    use std::os::windows::ffi::OsStringExt;
+    use winapi::um::{
+        sysinfoapi::GetVersionExW,
+        winnt::{LPOSVERSIONINFOEXW, LPOSVERSIONINFOW, OSVERSIONINFOEXW},
+    };
+
+    let mut version: OSVERSIONINFOEXW = Default::default();
```
youknowone, line 252: `let mut version = OSVERSIONINFOEXW::defalut()` also will work

vm/src/sysmodule.rs
```diff
@@ -434,6 +508,15 @@ settrace() -- set the global debug tracing function
       "__displayhook__" => ctx.new_function(sys_displayhook),
     });
 
+    #[cfg(windows)]
+    {
+        let getwindowsversion = WindowsVersion::make_class(ctx);
+        extend_module!(vm, module, {
+            "getwindowsversion" => ctx.new_function(sys_getwindowsversion),
+            "getwindowsversion_type" => getwindowsversion,
```
youknowone, line 516: Because this is not a documented or a CPython compitible attribute, naming to `_getwindowsversion_type` to mark it non-public and adding comment like '// XXX: This is not a python spec but required by current RustPython implementation` will be appreciated

BasixKOR: I just added some test cases and renamed the type. Could you check it again?

tests/snippets/stdlib_os.py
```diff
@@ -415,4 +415,25 @@ def __exit__(self, exc_type, exc_val, exc_tb):
     for arg in [None, 1, 1.0, TabError]:
         assert_raises(TypeError, os.system, arg)
 
+if "win" in sys.platform:
```
youknowone, line 418: Unfortunately, macos is `darwin` so it also inclues `win`. `sys.platform.startswith('win')` will work

vm/src/sysmodule.rs
```diff
@@ -222,6 +222,78 @@ fn sys_displayhook(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
     Ok(())
 }
 
+#[pystruct_sequence(name = "sys._getwindowsversion_type")]
```
youknowone, line 225: Actual name below is required to be aliased with `_` prefix, but this one is not. This name can be `sys.getwindowsversion` without any problem.

To make things clear, 
- The name here is the type name.
- The name below is the attrname of the sys module.

BasixKOR: Oh, thank you! I was bit confused about the differences between them.

vm/src/sysmodule.rs
```diff
@@ -222,6 +222,78 @@ fn sys_displayhook(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
     Ok(())
 }
 
+#[pystruct_sequence(name = "sys.getwindowsversion")]
```
youknowone, line 225: Would you please rebase this PR to upstream? This macro have a bit different signature due to #2020.

BasixKOR: Would you mind if I rearrange commit history a little bit while rebasing? I'd like to clean up commits related to formatting.

youknowone: Cleaning up local history by rebasing is always recommended, welcome

PR #2003, opened 2020-07-16
Implement lt, le, gt, ge method of array

qodot:
- Test case `BaseTest.test_cmp` in `Lib/test/test_array.py` is failing because there are no compare method of array.
- `BaseTest.test_cmp` is still failing after this PR because multiply methods of array are not implemented.

### DIFF
--- a/vm/src/obj/objbyteinner.rs
+++ b/vm/src/obj/objbyteinner.rs
@@ -115,11 +115,11 @@ impl ByteInnerNewOptions {
                             let bytes = vm.invoke(&bytes_method?, vec![])?;
                             return PyByteInner::try_from_object(vm, bytes);
                         }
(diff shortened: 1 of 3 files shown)


vm/src/obj/objbyteinner.rs
```diff
@@ -115,11 +115,11 @@ impl ByteInnerNewOptions {
                             let bytes = vm.invoke(&bytes_method?, vec![])?;
                             return PyByteInner::try_from_object(vm, bytes);
                         }
-                        let elements = vm.extract_elements(&obj).or_else(|_| {
-                            Err(vm.new_type_error(format!(
+                        let elements = vm.extract_elements(&obj).map_err(|_| {
```
youknowone, line 118: 👍

PR #2004, opened 2020-07-17
Fix clippy lints for Rust 1.45

coolreader18:


youknowone:
It looks good but is this duplicated work of #2003 ?

coolreader18:
Oh, yeah

PR #2006, opened 2020-07-17
Use loadname local replace loadname free

### LINKED ISSUE #2005: Should use (LoadName Local) instruction replace (LoadName Free)
Now we use [LoadName Free](https://github.com/RustPython/RustPython/blob/master/compiler/src/compile.rs#L301) to load local variable, maybe we should loadlocal for local variable.

e.g
```python
def f():
    sys.modules["test"] = "test";
```
In RustPython bytecode is 
```
                           0 LoadConst            (1)
                           1 LoadName             (test_sys, Free)
                           2 LoadAttr             (modules)
                           3 LoadConst            ("test")
                           4 StoreSubscript      
                           5 LoadConst            (None)
                           6 ReturnValue         
```

Lynskylate:
Fix #2005 . 

Use load local replace load free.


compiler/src/symboltable.rs
```diff
@@ -680,11 +682,11 @@ impl SymbolTableBuilder {
                 self.scan_expressions(vals, context)?;
             }
             Subscript { a, b } => {
-                self.scan_expression(a, context)?;
+                self.scan_expression(a, &ExpressionContext::Load)?;
```
Lynskylate, line 685: The result that symtable analyze assign ast may be error.
e.g: 
```python
def f():
    sys.modules["test"] = test
```
image
Identity get a assign context, and symtable think the variable scope is local.

compiler/src/symboltable.rs
```diff
@@ -276,6 +276,8 @@ impl<'a> SymbolTableAnalyzer<'a> {
     fn analyze_unknown_symbol(&self, symbol: &mut Symbol) {
         if symbol.is_assigned || symbol.is_parameter {
             symbol.scope = SymbolScope::Local;
+        } else if symbol.is_referenced {
```
Lynskylate, line 279: I am not sure about the correctness of this treatment.

coolreader18:
Looks like global scope names in nested closures aren't getting resolved correctly?

Lynskylate:
> Looks like global scope names in nested closures aren't getting resolved correctly?

Yes, the error is same as before.
Symtable thin object's context is assign when symtable analyze `registry[objcect] = func', and so object is a local variable

Lynskylate:
> Looks like global scope names in nested closures aren't getting resolved correctly?
```python
import symtable
content = open("Lib/functools.py").read()
table = symtable.symtable(content, "?", "exec")
// True in python
// False in RustPython
table.lookup("singledispatch").get_namespace().lookup("object").is_global()
"""

**Comment by Lynskylate** [comment]
RustPython‘s LoadLocal is not same as cpython's LoadFast.
```python
def f():
  ch = chr (100)
  if ch == 'd':
    chr = ch
  return chr

"""
Bytecode:
  2           0 LOAD_FAST                0 (chr)
              2 LOAD_CONST               1 (100)
              4 CALL_FUNCTION            1
              6 STORE_FAST               1 (ch)

  3           8 LOAD_FAST                1 (ch)
             10 LOAD_CONST               2 ('d')
             12 COMPARE_OP               2 (==)
             14 POP_JUMP_IF_FALSE       20

  4          16 LOAD_FAST                1 (ch)
             18 STORE_FAST               0 (chr)

  5     >>   20 LOAD_FAST                0 (chr)
             22 RETURN_VALUE
"""
```
Cpython analyze chr's scope is Local, and use LoadFast to load chr, but we can't load chr if we use LoadName local.
@coolreader18 Do you have any suggestions for handling this situation？

Lynskylate:
@coolreader18  Current _[codecss.py implemention](https://github.com/RustPython/RustPython/blob/master/Lib/_codecs.py#L1400)  is illegal in cpython. 
In PyUnicode_DecodeUnicodeEscape function,  chr referenced before assignment, and it will throw unbound error in cpython

coolreader18:
Hmm, that is tricky. Maybe we could do something in symtable to check if it's used global scope and then reassigned, to do a less optimized LoadName for it after? Or maybe we could just copy CPython's LoadFast, if feasible?

Lynskylate:
> RustPython‘s LoadLocal is not same as cpython's LoadFast.
> 
> ```python
> def f():
>   ch = chr (100)
>   if ch == 'd':
>     chr = ch
>   return chr
> 
> """
> Bytecode:
>   2           0 LOAD_FAST                0 (chr)
>               2 LOAD_CONST               1 (100)
>               4 CALL_FUNCTION            1
>               6 STORE_FAST               1 (ch)
> 
>   3           8 LOAD_FAST                1 (ch)
>              10 LOAD_CONST               2 ('d')
>              12 COMPARE_OP               2 (==)
>              14 POP_JUMP_IF_FALSE       20
> 
>   4          16 LOAD_FAST                1 (ch)
>              18 STORE_FAST               0 (chr)
> 
>   5     >>   20 LOAD_FAST                0 (chr)
>              22 RETURN_VALUE
> """
> ```
> 
> Cpython analyze chr's scope is Local, and use LoadFast to load chr, but we can't load chr if we use LoadName local.
> @coolreader18 Do you have any suggestions for handling this situation？

> RustPython‘s LoadLocal is not same as cpython's LoadFast.
> 
> ```python
> def f():
>   ch = chr (100)
>   if ch == 'd':
>     chr = ch
>   return chr
> 
> """
> Bytecode:
>   2           0 LOAD_FAST                0 (chr)
>               2 LOAD_CONST               1 (100)
>               4 CALL_FUNCTION            1
>               6 STORE_FAST               1 (ch)
> 
>   3           8 LOAD_FAST                1 (ch)
>              10 LOAD_CONST               2 ('d')
>              12 COMPARE_OP               2 (==)
>              14 POP_JUMP_IF_FALSE       20
> 
>   4          16 LOAD_FAST                1 (ch)
>              18 STORE_FAST               0 (chr)
> 
>   5     >>   20 LOAD_FAST                0 (chr)
>              22 RETURN_VALUE
> """
> ```
> 
> Cpython analyze chr's scope is Local, and use LoadFast to load chr, but we can't load chr if we use LoadName local.
> @coolreader18 Do you have any suggestions for handling this situation？
Sorry, the code throw unbound exception.
image

coolreader18:
Oh, I think we sent our comments at the same time, I didn't see yours before I sent mine. Yeah, I think the "reference before assignment" error + editing _codecs.py is the best course of action; the original version was from a PyPy branch compatible w/ python2 and there's been issues that pop up with it whenever a new module is added that calls into a part of _codecs that we haven't fixed yet.

PR #2007, opened 2020-07-17
test_zlib.py

youknowone:


Lib/test/test_zlib.py
```diff
@@ -0,0 +1,990 @@
...
+    @unittest.skip('TODO: RUSTPYTHON')
+    def test_flush_custom_length(self):
+        input = HAMLET_SCENE * 10
+        data = zlib.compress(input, 1)
+        dco = zlib.decompressobj()
+        dco.decompress(data, 1)
+        self.assertEqual(dco.flush(CustomInt()), input[1:])
+
+    @unittest.skip('TODO: RUSTPYTHON')
+    @requires_Compress_copy
```
coolreader18, line 696: You could probably just remove the `copy` method in `zlib.rs` and put a TODO comment instead; I wasn't aware that `copy` was optional/not guaranteed even in CPython.

PR #2012, opened 2020-07-19
Implement remove method of array

clemado1:
Implement remove method of array (vm/src/stdlib/array.rs)
I didn't uncommented TODO because there are still error with another problem.

#### Before
```
ERROR: test_remove (__main__.UnsignedShortTest)
  File "Lib/test/test_array.py", line 953, in test_remove
    self.assertRaises(ValueError, a.remove, self.outside)
AttributeError: 'array' object has no attribute 'remove'

```

#### After
```
ERROR: test_remove (__main__.UnsignedShortTest)
  File "Lib/test/test_array.py", line 955, in test_remove
    self.assertRaises(ValueError, a.remove, None)
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 746, in assertRaises
    context = None
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 743, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 743, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 181, in handle
    self = None
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 178, in handle
    callable_obj(*args, **kwargs)
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 178, in handle
    callable_obj(*args, **kwargs)
TypeError: Expected type <class 'int'>, not <class 'NoneType'>

```

### DIFF
--- a/vm/src/stdlib/array.rs
+++ b/vm/src/stdlib/array.rs
@@ -105,6 +105,20 @@ macro_rules! def_array_enum {
                 }
             }
 
+            fn remove(&mut self, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<()>{
+                match self {
+                    $(ArrayContentType::$n(v) => {
+                        let val = $t::try_from_object(vm, obj)?;
+                        if let Some(pos) = v.iter().position(|&a| a == val) {
+                            v.remove(pos);
+                        } else {
+                            return Err(vm.new_value_error("array.remove(x): x not in array".to_owned()));
+                        }
+                    })*
+                }
+                Ok(())
+            }
+
             fn frombytes(&mut self, b: &[u8]) {
                 match self {
                     $(ArrayContentType::$n(v) => {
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/array.rs
```diff
@@ -105,6 +105,20 @@ macro_rules! def_array_enum {
...
             }
 
+            fn remove(&mut self, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<()>{
+                match self {
+                    $(ArrayContentType::$n(v) => {
+                        let val = $t::try_from_object(vm, obj)?;
+                        if let Some(pos) = v.iter().position(|&a| a == val) {
+                            v.remove(pos);
+                        } else {
+                            return Err(vm.new_value_error("x not in array".to_owned()));
```
youknowone, line 115: Error message is different to CPython message. Will you check it please?

clemado1:
I modified Error Message to `array.remove(x): x not in array` and added the full error log. Thank you!
The error `TypeError: Expected type <class 'int'>, not <class 'NoneType'>` seems to occur too in array.index and array.count when the parameter is None.

Issue #356, opened 2019-02-06
Expected methods on object


veera83372:
The following methods need to implement in `src/vm/obj/objobject`.

- [ ] object . \_\_setattr__
- [ ] object . \_\_lt__
- [ ] object . \_\_le__
- [ ] object . \_\_gt__
- [ ] object . \_\_ge__
- [ ] object . \_\_reduce_ex__
- [ ] object . \_\_reduce__
- [ ] object . \_\_subclasshook__
- [ ] object . \_\_init_subclass__
- [ ] object . \_\_format__
- [ ] object . \_\_sizeof__
- [ ] object . \_\_dir__
- [ ] object . \_\_doc__

Tom1380:
@veera83372 you forgot to sign __doc__ as implemented.

Mec-iS:
A nice list of protocols with examples is [here](https://stephensugden.com/crash_into_python/Protocols.html)

dralley:
Current status is just

```
object.__init_subclass__
object.__reduce__
object.__reduce_ex__
object.__sizeof__
```

tekknolagi:
`__reduce__` and `__reduce_ex__` look implemented

dralley:
It's just ```object.__sizeof__`` missing now.

PR #2017, opened 2020-07-20
Implementation of __init_subclass__

### LINKED ISSUE #1346: PEP 487 (__init_subclass__, __set_name__)
## Feature

`__init_subclass__`, `__set_name__`

## Python Documentation

https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__
https://docs.python.org/3/reference/datamodel.html#object.__set_name__

BolunThompson:
This implements `__init_subclass__` from [PEP 487](https://www.python.org/dev/peps/pep-0487/).

Closes #1346. Partially fixes issue  #356.

### DIFF
--- a/vm/src/obj/objtype.rs
+++ b/vm/src/obj/objtype.rs
@@ -336,7 +336,11 @@ impl PyClassRef {
 
         if let Some(initter) = typ.get_super_attr("__init_subclass__") {
             let initter = vm
-                .call_get_descriptor_specific(initter.clone(), None, Some(typ.clone().into_object()))
+                .call_get_descriptor_specific(
+                    initter.clone(),
+                    None,
+                    Some(typ.clone().into_object()),
+                )
                 .unwrap_or(Ok(initter))?;
             vm.invoke(&initter, kwargs)?;
         };
--- a/vm/src/obj/objtype.rs
+++ b/vm/src/obj/objtype.rs
@@ -2,6 +2,7 @@ use parking_lot::RwLock;
 use std::collections::{HashMap, HashSet};
 use std::fmt;
 
+use super::objclassmethod::PyClassMethod;
 use super::objdict::PyDictRef;
 use super::objlist::PyList;
 use super::objmappingproxy::PyMappingProxy;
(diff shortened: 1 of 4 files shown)


vm/src/obj/objtype.rs
```diff
@@ -327,6 +334,17 @@ impl PyClassRef {
...
 
+        if let Some(initter) = typ.get_super_attr("__init_subclass__") {
+            let initter = PyClassMethod::descr_get(
+                vm,
+                initter,
+                None,
+                OptionalArg::Present(typ.clone().into_object()),
+            )?;
+            let init_args = PyFuncArgs::from(kwargs);
+            vm.invoke(&initter, init_args)?;
```
coolreader18, line 345: The second argument to `vm.invoke` can be any type that's `Into<PyFuncArgs>`:
```suggestion
            vm.invoke(&initter, kwargs)?;
```

vm/src/obj/objtype.rs
```diff
@@ -327,6 +334,17 @@ impl PyClassRef {
             }
         }
 
+        if let Some(initter) = typ.get_super_attr("__init_subclass__") {
+            let initter = PyClassMethod::descr_get(
+                vm,
+                initter,
+                None,
+                OptionalArg::Present(typ.clone().into_object()),
+            )?;
```
coolreader18, line 343: Maybe use `vm.call_descr_get_specific()` here? `initter` won't necessarily always be a PyClassMethod, even though that's not correct, but we should probably still call the appropriate `__get__` descriptor.

vm/src/obj/objtype.rs
```diff
@@ -335,14 +335,10 @@ impl PyClassRef {
...
-                vm,
-                initter,
-                None,
-                OptionalArg::Present(typ.clone().into_object()),
-            )?;
-            let init_args = PyFuncArgs::from(kwargs);
-            vm.invoke(&initter, init_args)?;
+            let initter = vm
+                .call_get_descriptor_specific(initter, None, Some(typ.clone().into_object()))
+                .unwrap()?;
```
coolreader18, line 340:
```suggestion
                .call_get_descriptor_specific(initter.clone(), None, Some(typ.clone().into_object()))
                .unwrap_or(Ok(initter))?;
```

coolreader18:
Oh, actually, would you be able to put the copied CPython test file in a separate commit? Something like

- Add test_subclassinit.py from CPython 3.x.x
  - Same exact file as in CPython
- Mark expected failures
  - edit the test file to put `expectedFailure` et al

BolunThompson:
I'm done with this.

coolreader18:
Thank you for contributing!

PR #2018, opened 2020-07-21
Move a few non-python specific functions into a 'common' crate, make threading optional

coolreader18:
Disabling threading (and not having mutexes around every object) gives about a 20% speedup, so I think it would be useful for an embedded scripting use case that doesn't need threading.


common/src/rc.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(not(feature = "threading"))]
+use std::rc::{Rc, Weak};
+#[cfg(feature = "threading")]
+use std::sync::{Arc, Weak};
+
+// type aliases instead of newtypes because you can't do `fn method(self: PyRc<Self>)` with a
+// newtype; requires the arbitrary_self_types unstable feature
+
+#[cfg(feature = "threading")]
+pub type PyRc<T> = Arc<T>;
```
youknowone, line 10: Do we still use `Py` prefix if when these code is regarded as non-python specific?

coolreader18: I did it moreso to recognize that it's part of rustpython; I think it would be weird to just call it `Rc` when it could actually be an `Arc` as well.

palaviv:
By the way not that I want to go there but this might make it easy to create a GIL mode. We just need all `PyMutex` and `PyRwLock` to share one mutex in that case...

coolreader18:
Is this good to merge do you think?

coolreader18:
I suppose that's fair; I was just sort of thinking that e.g. ufrexp doesn't have anything that means it has to be in `vm`, so we might as well move it to a separate crate so that others could use it if they wanted, but it's probably fine to just keep it in objfloat.

youknowone:
Oh, I am sorry to make confusion. I thought `eq_int` is taking `PyInt` not `BigInt`.

PR #2020, opened 2020-07-22
implemented __repr__ method in impl_pyclass_sequence

### LINKED ISSUE #1448: Implement a better __repr__ in impl_pystruct_sequence
The ```impl_pystruct_sequence``` proc macro needs to implement a custom ```__repr__``` to match CPython.

### Expected
```
>>> sys.version_info
sys.version_info(major=3, minor=7, micro=4, releaselevel='final', serial=0)
```

### Actual
```
>>>>> sys.version_info
(3, 5, 0, 'alpha', 0)
```

hatee-hatee-hatee-ho:
Just implemented 'rustpython_vm::version::VersionInfo::repr' method, and added as value of property 'sys.version_info'.
It's now works, but refactoring is required.

Fix #1448

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -523,13 +523,38 @@ pub fn impl_pystruct_sequence(attr: AttributeArgs, item: Item) -> Result<TokenSt
                 );
                 ::rustpython_vm::pyobject::PyValue::into_ref_with_type(tuple, vm, cls)
             }
+
+                        fn repr(zelf: rustpython_vm::pyobject::PyRef<rustpython_vm::obj::objtuple::PyTuple>, vm: &rustpython_vm::VirtualMachine) -> ::rustpython_vm::pyobject::PyResult<String> {
+                                let s = if let Some(_guard) = rustpython_vm::vm::ReprGuard::enter(zelf.as_object()) {
+                                        let field_names = vec![#(stringify!(#field_names)),*];
+                                        let mut fields = Vec::new();
+                                        for (value, field_name) in zelf.as_slice().iter().zip(field_names.iter()) {
+                                                let s = vm.to_repr(value)?;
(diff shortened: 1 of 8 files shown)


vm/src/version.rs
```diff
@@ -7,7 +7,7 @@ const MICRO: usize = 0;
 const RELEASELEVEL: &str = "alpha";
 const SERIAL: usize = 0;
 
-#[pystruct_sequence(name = "version_info")]
+#[pystruct_sequence(name = "sys.version_info")]
```
youknowone, line 10: The name is `version_info` and `sys` is module name.
```
>>> import sys
>>> type(sys.version_info).__name__
'version_info'
>>> type(sys.version_info).__module__
'sys'
```

hatee-hatee-hatee-ho: Thank you for advice. I'll fix it asap.
Yesterday, I found similar struct which has same problem.
https://github.com/RustPython/RustPython/blob/master/vm/src/pyhash.rs#L107-L118
Do I need create a new issue or only PR?

derive/src/pyclass.rs
```diff
@@ -523,13 +528,35 @@ pub fn impl_pystruct_sequence(attr: AttributeArgs, item: Item) -> Result<TokenSt
...
+                                fields.push(format!("{}: {}", field_name, s));
+                        }
+
+                        if fields.len() == 1 {
+                                format!("{}({},)", #py_class_name, fields[0])
+                        } else {
+                                format!("{}({})", #py_class_name, fields.join(", "))
+                        }
+                } else {
+                        concat!(#class_name, "(...)").to_string()
```
youknowone, line 547: Without checking, I feel like CPython may use `__name__` and `__module__` here. Though it will work because these kind of builtin types doesn't allow to set `__name__` and `__module__`

PR #2022, opened 2020-07-24
Add CPython 3.83 test test_cmd.py

minoring:
All tests are passed.


coolreader18:
Nice work, thanks for contributing!

PR #2025, opened 2020-07-25
fix a divmod bug

hyperbora:
Floating point modulo operation is different between python and rust.

### DIFF
--- a/vm/src/obj/objfloat.rs
+++ b/vm/src/obj/objfloat.rs
@@ -123,7 +123,13 @@ fn inner_floordiv(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<f64> {
 
 fn inner_divmod(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<(f64, f64)> {
     if v2 != 0.0 {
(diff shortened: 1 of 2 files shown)


vm/src/obj/objfloat.rs
```diff
@@ -123,7 +123,8 @@ fn inner_floordiv(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<f64> {
 
 fn inner_divmod(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<(f64, f64)> {
     if v2 != 0.0 {
-        Ok(((v1 / v2).floor(), v1 % v2))
+        let d = (v1 / v2).floor();
+        Ok((d, v1 - (v2 * d)))
```
youknowone, line 127: This is a bit different from CPython way. This implementation is ok, but look into it if you want to do. The corresponding CPython function is `float_divmod` in `floatobject.c` file. Please tell me if you want to fix it more or want to make this merged as it is.

hyperbora: Ok, I corrected it by referring to cpython.

PR #2026, opened 2020-07-25
Fixed bug where sys.executable returned an absolute path instead of a relative path

BolunThompson:
Before, sys.executable returned a relative path, unlike what the [documentation](https://docs.python.org/3/library/sys.html#sys.executable) says. This makes sys.executable return an absolute path.

### DIFF
--- a/vm/src/sysmodule.rs
+++ b/vm/src/sysmodule.rs
@@ -29,7 +29,7 @@ fn argv(vm: &VirtualMachine) -> PyObjectRef {
 
 fn executable(ctx: &PyContext) -> PyObjectRef {
     if let Ok(path) = env::current_exe() {
-        if let Some(path) = path.to_str() {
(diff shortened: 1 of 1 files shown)


vm/src/sysmodule.rs
```diff
@@ -28,11 +28,12 @@ fn argv(vm: &VirtualMachine) -> PyObjectRef {
 }
 
 fn executable(ctx: &PyContext) -> PyObjectRef {
-    if let Some(arg) = env::args().next() {
-        ctx.new_str(arg)
-    } else {
-        ctx.none()
+    if let Ok(path) = env::current_exe() {
+        if let Some(path) = path.to_str() {
```
coolreader18, line 32:
```suggestion
        if let Ok(path) = path.into_os_string().into_string() {
```
Avoids cloning

BolunThompson: Thanks!

Issue #1992, opened 2020-07-10
os.execv is missing, and other os methods like os.execl appear to try to use it


illicitonion:
```
$ cargo run -- -c 'import os ; os.execv("/bin/echo", ["/bin/echo", "Hello"])'
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/rustpython -c 'import os ; os.execv("/bin/echo", ["/bin/echo", "Hello"])'`
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: module 'os' has no attribute 'execv'
$ cargo run -- -c 'import os ; os.execl("/bin/echo", *["/bin/echo", "Hello"])'
    Finished dev [unoptimized + debuginfo] target(s) in 0.14s
     Running `target/debug/rustpython -c 'import os ; os.execl("/bin/echo", *["/bin/echo", "Hello"])'`
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/RustPython/Lib/os.py", line 437, in execl
    execv(file, args)
NameError: name 'execv' is not defined
```

## Python Documentation

https://docs.python.org/3/library/os.html#os.execv

```
$ python -c 'import os ; os.execv("/bin/echo", ["/bin/echo", "Hello"])'
Hello
$ python -c 'import os ; os.execl("/bin/echo", *["/bin/echo", "Hello"])'
Hello
```

youknowone:
Fixed by #2027

PR #2027, opened 2020-07-26
os.execv implementation for unix systems

ohgree:
This PR adds os.execv for unix systems which is a part of #1992
Uses nix::unistd::execv for implementation.
Would appreciate any help on line 1203 - temporary value referencing in closure while converting CString to &CStr

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1191,6 +1191,26 @@ fn os_chmod(
     Ok(())
 }
 
+#[cfg(unix)]
+fn os_execv(
+    path: PyStringRef,
+    argv_list: PyIterable<PyStringRef>,
+    vm: &VirtualMachine
+    ) {
+
+    let path = ffi::CString::new(path.as_str()).unwrap();
+
+    let argv: Vec<&ffi::CStr> = argv_list.iter(vm).unwrap().map(
+        |entry| ffi::CString::new(
+            entry
+            .unwrap()
+            .as_str()
+        ).unwrap().as_c_str()
+    ).collect();
+
+    unistd::execv(&path, &argv);
+}
+
 fn os_fspath(path: PyPathLike, vm: &VirtualMachine) -> PyResult {
     path.mode.process_path(path.path, vm)
 }
@@ -2089,6 +2109,7 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         #[cfg(not(target_os = "redox"))]
         SupportFunc::new(vm, "chroot", os_chroot, Some(false), None, None),
         SupportFunc::new(vm, "umask", os_umask, Some(false), Some(false), Some(false)),
+        SupportFunc::new(vm, "execv", os_execv, None, None, None)
     ]);
     let supports_fd = PySet::default().into_ref(vm);
     let supports_dir_fd = PySet::default().into_ref(vm);
@@ -2186,6 +2207,7 @@ fn extend_module_platform_specific(vm: &VirtualMachine, module: &PyObjectRef) {
 
     extend_module!(vm, module, {
         "chmod" => ctx.new_function(os_chmod),
+        "execv" => ctx.new_function(os_execv), // TODO: windows
         "get_inheritable" => ctx.new_function(os_get_inheritable), // TODO: windows
         "get_blocking" => ctx.new_function(os_get_blocking),
         "getppid" => ctx.new_function(os_getppid),
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1200,11 +1200,12 @@ fn os_execv(
     // TODO: argv_list to only accept tuples and lists
     // message: "TypeError: execv() arg 2 must be a tuple or list"
 
-    if path.to_string().contains("\0") {
-        return Err(vm.new_value_error("embedded null byte".to_owned()));
-    }
-
-    let path = ffi::CString::new(path.as_str()).unwrap();
+    let path = match ffi::CString::new(path.as_str()) {
+        Ok(s) => s,
+        Err(_) => {
+            return Err(vm.new_value_error("embedded null byte".to_owned()));
+        }
+    };
 
     if argv_list.iter(vm)?.count() == 0 {
         return Err(vm.new_value_error("execv() arg 2 must not be empty".to_owned()));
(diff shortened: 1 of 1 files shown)


ohgree:
Got my problem solved.
I'll mark this PR ready when all checks finish.

vm/src/stdlib/os.rs
```diff
@@ -1191,6 +1191,27 @@ fn os_chmod(
...
+fn os_execv(
+    path: PyStringRef,
+    argv_list: PyIterable<PyStringRef>,
+    vm: &VirtualMachine,
+) -> PyResult<()> {
+    let path = ffi::CString::new(path.as_str()).unwrap();
+
+    let argv_list: Vec<ffi::CString> = argv_list
+        .iter(vm)
+        .unwrap()
```
youknowone, line 1204: what's happened if `iter()` returns `Err`? `unwrap` will cause crash. We must relay out the exception. `.iter(vm)?` will do it.

vm/src/stdlib/os.rs
```diff
@@ -1191,6 +1191,27 @@ fn os_chmod(
     Ok(())
 }
 
+#[cfg(unix)]
+fn os_execv(
+    path: PyStringRef,
+    argv_list: PyIterable<PyStringRef>,
+    vm: &VirtualMachine,
+) -> PyResult<()> {
+    let path = ffi::CString::new(path.as_str()).unwrap();
```
youknowone, line 1200: how CPython work when path includes null character? Maybe there is some error then we'd better to raise here too.

ohgree: CPython raises ValueError for null-embeded path, and same exception is also raised for empty tuple/lists. These two exceptions should be taken care of.

vm/src/stdlib/os.rs
ohgree: Thx for the review. Expect a fix for this issue in the upcoming commit, which I’ll push asap after correct exception handling has been implemented

vm/src/stdlib/os.rs
```diff
@@ -1191,6 +1191,55 @@ fn os_chmod(
...
+        return Err(vm.new_value_error("execv() arg 2 must not be empty".to_owned()));
+    }
+
+    for arg in argv_list.iter(vm)? {
+        if arg?.as_str().contains("\0") {
+            return Err(vm.new_value_error("embedded null byte".to_owned()));
+        }
+    }
+
+    if argv_list
```
youknowone, line 1219: This condition will be easier with vec from `Either`. Otherwise, checking len of `argv_list: Vec<ffi::CString>` after the creation would be also easy.

vm/src/stdlib/os.rs
```diff
@@ -1191,6 +1191,55 @@ fn os_chmod(
...
+    vm: &VirtualMachine,
+) -> PyResult<()> {
+    // TODO: argv_list to only accept tuples and lists
+    // message: "TypeError: execv() arg 2 must be a tuple or list"
+
+    if path.to_string().contains("\0") {
+        return Err(vm.new_value_error("embedded null byte".to_owned()));
+    }
+
+    let path = ffi::CString::new(path.as_str()).unwrap();
```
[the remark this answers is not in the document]
ohgree: ? operator is used for Result enum in this function, so I'll probably use match instead of closure with map_err(). Thanks for the insight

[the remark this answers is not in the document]
youknowone: That's exactly what we want here. https://doc.rust-lang.org/std/result/enum.Result.html#method.map_err

`O` in `map_err` takes `E` returns `F`. In this case, it can take `NulByteError` and returns `PyObjectRef` which is used for error type in RustPython.

vm/src/stdlib/os.rs
ohgree: used Vec<ffi::CString>.

vm/src/stdlib/os.rs
```diff
@@ -1191,6 +1191,48 @@ fn os_chmod(
...
+) -> PyResult<()> {
+    // TODO: argv_list to only accept tuples and lists
+    // "TypeError: execv() arg 2 must be a tuple or list"
+
+    let path = match ffi::CString::new(path.as_str()) {
+        Ok(s) => s,
+        Err(_) => {
+            return Err(vm.new_value_error("embedded null byte".to_owned()));
+        }
+    };
```
youknowone, line 1208:
```suggestion
    let path = ffi::CString::new(path.as_str())
        .map_err(|_| vm.new_value_error("embedded null byte".to_owned()))?;
```

vm/src/stdlib/os.rs
```diff
@@ -1191,6 +1191,48 @@ fn os_chmod(
...
+    for arg in argv_list.iter(vm)? {
+        if arg?.as_str().contains('\0') {
+            return Err(vm.new_value_error("embedded null byte".to_owned()));
+        }
+    }
+
+    let argv_list: Vec<ffi::CString> = argv_list
+        .iter(vm)?
+        .map(|entry| ffi::CString::new(entry.unwrap().as_str()).unwrap())
+        .collect();
```
youknowone, line 1223: The upper null byte check can be done in below `CString::new()`

ohgree: Addressed.

vm/src/stdlib/os.rs
ohgree: Addressed at https://github.com/RustPython/RustPython/pull/2027/commits/f7b9c21fa43e6c4ec29d59ac963ef41cc38809d4

vm/src/stdlib/os.rs
```diff
@@ -1197,11 +1197,39 @@ fn os_execv(
     argv_list: PyIterable<PyStringRef>,
     vm: &VirtualMachine,
 ) -> PyResult<()> {
+    // TODO: argv_list to only accept tuples and lists
+    // message: "TypeError: execv() arg 2 must be a tuple or list"
```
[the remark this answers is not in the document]
ohgree: The change's been addressed at https://github.com/RustPython/RustPython/pull/2027/commits/f73bf8d6f56b076e99c696843097261ab802e7ab, along with a few minor tweaks on exception handling.

youknowone:
Fixed a bit in #2100

PR #2028, opened 2020-07-26
Implement __ne__ methods for array, bytearray

ugaemi:
Added `__ne__` methods to array and bytearray.

#1442

### DIFF
--- a/vm/src/stdlib/array.rs
+++ b/vm/src/stdlib/array.rs
@@ -456,6 +456,25 @@ impl PyArray {
         }
     }
 
+    #[pymethod(name = "__ne__")]
+    fn ne(lhs: PyObjectRef, rhs: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        let lhs = class_or_notimplemented!(vm, Self, lhs);
+        let rhs = class_or_notimplemented!(vm, Self, rhs);
+        let lhs = lhs.borrow_value();
+        let rhs = rhs.borrow_value();
+        if lhs.len() != rhs.len() {
+            Ok(vm.new_bool(true))
+        } else {
+            for (a, b) in lhs.iter(vm).zip(rhs.iter(vm)) {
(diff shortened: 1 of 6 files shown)


vm/src/obj/objbyteinner.rs
```diff
@@ -287,6 +287,10 @@ impl PyByteInner {
         self.cmp(other, |a, b| a == b, vm)
     }
 
+    pub fn ne(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyComparisonValue {
+        self.cmp(other, |a, b| a != b, vm)
```
youknowone, line 291: when `__eq__` is implemented with `PyComparisionValue` return type , `__ne__` is idiomatically implemented like:

```suggestion
        self.eq(other, vm).map(|v| !v)
```

This implementation is applicable regardless of `__eq__` implementation.

vm/src/stdlib/array.rs
```diff
@@ -431,6 +431,25 @@ impl PyArray {
         }
     }
 
+    #[pymethod(name = "__ne__")]
+    fn ne(lhs: PyObjectRef, rhs: PyObjectRef, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 435: Instead of copying `__eq__` implementation, refactoring `__eq__` to use `PyComparisonValue` like `bytes` will allow this method reuse the idiomatic `__ne__` implementation.

youknowone: Similar work: https://github.com/RustPython/RustPython/pull/2092/commits/feb1b895d7ce5fc8b4368c69b4ea65d0916037a8#diff-c9329f7f730f2866bea2eb2f3eebb95cR129-R145

ugaemi:
> LTGM. this PR now has conflict. would you rebase it to upstream?

Yes, I'll reflect it right away. Thank you for your feedback :)

PR #2029, opened 2020-07-26
Update dependencies

### LINKED ISSUE #1909: Depend on lz-fear instead of lz4-compress
lz4-compress is from RedoxOS and is unmaintained: https://gitlab.redox-os.org/redox-os/tfs/issues/89

lz-fear is a fork of lz4-compress that was modified significantly for better performance, and is better maintained: https://github.com/main--/rust-lz-fear

coolreader18:
Now there are only 2 crates with multiple versions in our dependency tree 

Closes #1909

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -100,12 +100,6 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "0.1.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


youknowone:
Is this related to #1909?

coolreader18:
Oh, yes, sort of. I didn't depend on lz-fear but a different fork of the redox crate, lz4-compress, which is maintained and has a more similar API to the original than lz-fear.

PR #2030, opened 2020-07-27
Fix #1986 which (-2) ** 0.5 yields NaN.

### LINKED ISSUE #1986: (-2) ** 0.5 yields NaN
## Feature

non-integer power of negative numbers.

## Python Documentation

* https://docs.python.org/3/library/stdtypes.html#numeric-types-int-float-complex

minoring:
Fix #1986 

**Before:**
- RustPython
`>>>>> float.__rpow__(0.5, -2)`
`NaN`

**After:**
- RustPython
`>>>float.__rpow__(0.5, -2)`
`(0.00000000000000008659560562354934+1.4142135623730951j)`

- CPython 3.85
`>>> float.__rpow__(0.5, -2)`
`(8.659560562354934e-17+1.4142135623730951j)`

### DIFF
--- a/vm/src/obj/objfloat.rs
+++ b/vm/src/obj/objfloat.rs
@@ -1,4 +1,5 @@
 use num_bigint::{BigInt, ToBigInt};
+use num_complex::Complex64;
 use num_rational::Ratio;
 use num_traits::{pow, ToPrimitive, Zero};
 
@@ -134,12 +135,16 @@ fn inner_divmod(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<(f64, f64)> {
     }
 }
 
-pub fn float_pow(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<f64> {
+pub fn float_pow(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<ComplexFloatValue> {
     if v1.is_zero() {
         let msg = format!("{} cannot be raised to a negative power", v1);
         Err(vm.new_zero_division_error(msg))
+    } else if (v1 < 0.0) & (v2.floor() != v2) {
+        let complex_v1 = Complex64::new(v1, 0.);
+        let complex_v2 = Complex64::new(v2, 0.);
+        Ok(ComplexFloatValue::Complex64(complex_v1.powc(complex_v2)))
     } else {
-        Ok(v1.powf(v2))
+        Ok(ComplexFloatValue::Float64(v1.powf(v2)))
     }
 }
 
@@ -259,6 +264,17 @@ impl PyFloat {
     }
 
     #[inline]
+    fn might_complex_op<F>(&self, other: PyObjectRef, op: F, vm: &VirtualMachine) -> PyResult
+    where
+        F: Fn(f64, f64) -> PyResult<ComplexFloatValue>,
+    {
+        try_float(&other, vm)?.map_or_else(
+            || Ok(vm.ctx.not_implemented()),
+            |other| op(self.value, other)?.into_pyobject(vm),
+        )
+    }
+
+    #[inline]
     fn tuple_op<F>(&self, other: PyObjectRef, op: F, vm: &VirtualMachine) -> PyResult
     where
         F: Fn(f64, f64) -> PyResult<(f64, f64)>,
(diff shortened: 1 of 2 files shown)


vm/src/obj/objfloat.rs
```diff
@@ -558,6 +600,20 @@ impl TryFromObject for IntoPyFloat {
     }
 }
 
+pub enum ComplexFloatValue {
```
youknowone, line 603: This is good trial for mixed return type. But in this case, I think this function is a simple combination of 2 functions:
- float ** positive float -> float
- float ** negative float -> complex

By keeping float_pow as it was and adding a new function (probably `float_negative_pow`?), it would be well-defined without this type.

`might_complex_op` can be a pow function which call 2 pow funcitons by v2. Roughly, it would look like:
```rust
fn _pow(v1: float, v2: float) -> PyResult {
   if v2 is positive {
       float_pow().into_pyobject()
   } else {
       float_negative_pow().into_pyobject()
   }
}
```

minoring: Thank you for your kind review :)
I updated PR with reference to your suggestion. Please take a look ;)
If you have any more advice, it would be welcomed!

vm/src/obj/objfloat.rs
```diff
@@ -134,6 +134,17 @@ fn inner_divmod(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<(f64, f64)> {
     }
 }
 
+fn inner_pow(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult {
+    if v1.is_zero() {
+        let msg = format!("{} cannot be raised to a negative power", v1);
+        Err(vm.new_zero_division_error(msg))
+    } else if (v1 < 0.0) & (v2.floor() != v2) {
```
youknowone, line 141: I am not sure `&` is inteded or not. Is this intended to be like below?
```suggestion
    } else if v1.is_negative() && v2.floor() != v2 {
```

vm/src/obj/objfloat.rs
```diff
@@ -134,6 +134,17 @@ fn inner_divmod(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<(f64, f64)> {
...
+        float_ops::float_negative_pow(v1, v2).into_pyobject(vm)
```
youknowone, line 142:
```suggestion
        let v1 = Complex64::new(v1, 0.);
        let v2 = Complex64::new(v2, 0.);
        v1.powc(v2).into_pyobject(vm)
```

vm/src/obj/objfloat.rs
```diff
@@ -134,6 +134,17 @@ fn inner_divmod(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<(f64, f64)> {
...
+    } else {
+        float_ops::float_pow(v1, v2).into_pyobject(vm)
```
youknowone, line 144:
```suggestion
        v1.powf(v2).into_pyobject(vm)
```

vm/src/obj/objfloat.rs
```diff
@@ -258,6 +269,17 @@ impl PyFloat {
         )
     }
 
+    #[inline]
+    fn might_complex_op<F>(&self, other: PyObjectRef, op: F, vm: &VirtualMachine) -> PyResult
```
youknowone, line 273: I think this name can be simply `complex_op`.
```suggestion
        fn complex_op<F>(&self, other: PyObjectRef, op: F, vm: &VirtualMachine) -> PyResult
```

vm/src/obj/objfloat.rs
```diff
@@ -134,12 +135,16 @@ fn inner_divmod(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<(f64, f64)> {
     }
 }
 
-pub fn float_pow(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult<f64> {
+pub fn float_pow(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult {
     if v1.is_zero() {
         let msg = format!("{} cannot be raised to a negative power", v1);
         Err(vm.new_zero_division_error(msg))
+    } else if v1.is_sign_negative() && v2.floor() != v2 {
```
youknowone, line 142: Check the clippy error please. In local, `cargo clippy --all` command works. We also can see the result in CI result too.

minoring: Updated PR. Thanks 👍

PR #2031, opened 2020-07-27
Fixed bug where sys.executable may follow symlinks

BolunThompson:
Currently, sys.executable calls [`std::env::current_exe`](https://doc.rust-lang.org/std/env/fn.current_exe.html). With current_exe, it is platform defined if it follows symlinks. However, from my testing sys.executable never follows symlinks in CPython. This behavior is important in certain cases, e.g [this](https://github.com/BolunThompson/RustPython/blob/51c9e8824a1305d1e670700aa0904825656c921e/Lib/test/test_venv.py#L327) test only passes if symlinks aren't followed. 

This changes sys.executable to not follow symlinks.

### DIFF
--- a/Lib/test/test_sys_executable.py
+++ b/Lib/test/test_sys_executable.py
@@ -1,0 +1,62 @@
+import unittest
+import subprocess
+import sys
+import os
+
+# These should go into a test_sys.py file, but given these are the only
+# ones being tested currently that seems unnecessary.
+
+
+class SysExecutableTest(unittest.TestCase):
+
+    # This is a copy of test.test_sys.SysModuleTest.test_executable from cpython.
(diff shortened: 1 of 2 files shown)


BolunThompson:
I just added those tests.

Lib/test/test_sys_executable.py
```diff
@@ -0,0 +1,62 @@
+import unittest
+import subprocess
+import sys
+import os
+
+# These should go into a test_sys.py file, but given these are the only
```
coolreader18, line 6: This is fine for now, I'll look into getting all of test_sys.py to work.

coolreader18:
Thanks for contributing!

PR #2032, opened 2020-07-27
os.rs - Minor typo fix for panic! macro

ohgree:
panic message has following typo: "Uknown file type"
fixed to "Unknown file type"


coolreader18:
Thanks!

Issue #935, opened 2019-05-07
should we change str/unicode implementation for PEP 383 ?


jgirardet:
when decoding/encoding some str/bytes you can (for example) :
'strict' :  raise an exception
'ignore' : wrong bytes are kicked out
'replace' :  wrong bytes are replaced by the UNICODE_REPLACEMENT_CHAR
'backslashreplace' : Replace with backslashed escape sequences

No problem with current implementation for these ones.

> This PEP proposes a means of dealing with [such] irregularities by embedding the bytes in  character strings in such a way that allows recreation of the original byte string.

'surrogateescape' : replace wrong bytes by surrogate 

> With this PEP, non-decodable bytes >= 128 will be represented as lone surrogate codes U+DC80..U+DCFF.

But the actual RustPython implementation stores `str` as valid unicode then : 
```
a.push('\u{dcc3}');
help: src/main.rs:25: unicode escape must not be a surrogate
```
storing surrogate is not allowed.
I tried to do it inside an unsafe block it's not possible either.

I'm note sure  to have a clear idea howto to do  it. I was thinking to a wrapper on a "homemade" unicode and the Rust unicode implementation to store strings,  but actually python is able to call some methods on invalid  utf8 sequence : 
```
a.capitalize()                                                         
>>>  '\udcc3\udca9'
```

I'd be happy to have some opinion about it.

youknowone:
Probably we need another str kind presented by `&[u32]`

youknowone:
we now have wtf8-backed str. 
```
>>>>> "A\udcdeB".encode('utf-8')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
UnicodeEncodeError: ('utf-8', 'A\udcdeB', 1, 2, 'surrogates not allowed')
```

PR #2034, opened 2020-07-28
Add cgi module

skinny121:
Added the cgi module(https://docs.python.org/3.8/library/cgi.html).

There are still three failing tests in `test_cgi.py`, one of which requires #935, and the other two needs 'w+' file mode support.

This also has updated the urllib stdlib package to match the CPython 3.8.3 version.

### DIFF
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -77,7 +77,6 @@ generational-arena = "0.2"
 parking_lot = "0.11"
 thread_local = "1.0"
 cfg-if = "0.1.10"
-memchr = "2.1.0"
 
 ## unicode stuff
 unicode_names2 = "0.4"
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -77,6 +77,7 @@ generational-arena = "0.2"
 parking_lot = "0.11"
 thread_local = "1.0"
 cfg-if = "0.1.10"
+memchr = "2.1.0"
 
 ## unicode stuff
 unicode_names2 = "0.4"
--- a/vm/src/stdlib/io.rs
+++ b/vm/src/stdlib/io.rs
@@ -1,9 +1,11 @@
 /*
  * I/O core tools.
  */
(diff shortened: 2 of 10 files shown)


vm/Cargo.toml
```diff
@@ -77,6 +77,7 @@ generational-arena = "0.2"
 parking_lot = "0.11"
 thread_local = "1.0"
 cfg-if = "0.1.10"
+memchr = "2.1.0"
```
youknowone, line 80: It seems `bstr::ByteSlice::find` is compatible with memchr.  In that case, because we already uses bstr, I think we don't need to add this crate

https://docs.rs/bstr/0.2.13/bstr/trait.ByteSlice.html#method.find

vm/src/stdlib/io.rs
```diff
@@ -109,12 +109,35 @@ impl BufferedIO {
...
+        let mut buf = Vec::new();
+        let mut left = size.flat_option().filter(|s| *s >= 0).unwrap_or(i64::MAX);
+        if left == 0 {
+            return Ok(buf);
+        }
+        loop {
+            let (done, used) = {
+                let mut available = self.cursor.fill_buf().map_err(|err| os_err(vm, err))?;
+                if left < available.len() as i64 {
+                    available = &available[..left as usize];
```
youknowone, line 122: It seems i64 to usize casting is not safe for 32bit platforms

youknowone:
@coolreader18 @palaviv @windelbouwman Do we have any reason to use `BufferedIO` instead of `std::io::BufRead`? https://doc.rust-lang.org/std/io/trait.BufRead.html

vm/src/stdlib/io.rs
skinny121: I think this was safe as `available.len()` is a `usize`, so `left` can always fit into 32bits here. I have changed it so that `left` is now also a `usize`, which removes a number of casts.

PR #2035, opened 2020-07-28
Refactor DictKey

youknowone:
This patch make DictKey used in more natural manner - except for `&str` case.

PR #2036, opened 2020-07-28
DictKey for str shares same hash with PyString

youknowone:
Creating a new PR from #2035 due to crash.

Do anyone know why changing str hash function causes crash?

### DIFF
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -1,16 +1,17 @@
-use crate::common::cell::{PyRwLock, PyRwLockReadGuard, PyRwLockWriteGuard};
-use crate::obj::objstr::{PyString, PyStringRef};
-use crate::pyobject::{IdProtocol, IntoPyObject, PyObjectRef, PyResult};
-use crate::vm::VirtualMachine;
-use num_bigint::ToBigInt;
-use rustpython_common::hash;
 /// Ordered dictionary implementation.
 /// Inspired by: https://morepypy.blogspot.com/2015/01/faster-more-memory-efficient-and-more.html
 /// And: https://www.youtube.com/watch?v=p33CVV29OG8
 /// And: http://code.activestate.com/recipes/578375/
-use std::collections::{hash_map::DefaultHasher, HashMap};
-use std::hash::{Hash, Hasher};
+use crate::common::cell::{PyRwLock, PyRwLockReadGuard, PyRwLockWriteGuard};
+use crate::obj::objstr::{PyString, PyStringRef};
+use crate::pyobject::{IdProtocol, IntoPyObject, PyObjectRef, PyResult};
+use crate::vm::VirtualMachine;
+use rustpython_common::hash;
+use std::collections::HashMap;
 use std::mem::size_of;
+
+// HashIndex is intended to be same size with hash::PyHash
+// but it doesn't mean the values are compatible with actual pyhash value
 
 /// hash value of an object returned by __hash__
 type HashValue = hash::PyHash;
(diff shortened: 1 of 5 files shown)


coolreader18:
Oh, I ran into something similar with my shared string payload branch; basically, hash(obj) calls the hash magic method and then hashes the BigInt, and you have to do the same thing to the string hash function for the DictKey impl if you want dict lookups (i.e. local variable lookups) to work correctly

coolreader18:
https://github.com/RustPython/RustPython/blob/coolreader18/constant-strings/vm/src/dictdatatype.rs#L436-L439

youknowone:
@coolreader18 thanks, I fixed it

vm/src/dictdatatype.rs
```diff
@@ -412,40 +413,43 @@ enum LookupResult {
...
 /// - str -> string reference used as key, this is often used internally
 pub trait DictKey: IntoPyObject {
-    fn do_hash(&self, vm: &VirtualMachine) -> PyResult<HashValue>;
-    fn do_is(&self, other: &PyObjectRef) -> bool;
-    fn do_eq(&self, vm: &VirtualMachine, other_key: &PyObjectRef) -> PyResult<bool>;
+    fn key_hash(&self, vm: &VirtualMachine) -> PyResult<HashValue> {
+        let raw_hash = self.key_object_hash(vm)?;
+        let mut hasher = DefaultHasher::new();
+        raw_hash.hash(&mut hasher);
+        Ok(hasher.finish() as HashValue)
```
youknowone, line 420: is this intentionally hashed again due to some reason like security?

coolreader18: Not sure, it does seem redundant. Does it work if you remove it?

vm/src/dictdatatype.rs
```diff
@@ -1,17 +1,18 @@
...
-/// Ordered dictionary implementation.
-/// Inspired by: https://morepypy.blogspot.com/2015/01/faster-more-memory-efficient-and-more.html
-/// And: https://www.youtube.com/watch?v=p33CVV29OG8
-/// And: http://code.activestate.com/recipes/578375/
 use std::collections::{hash_map::DefaultHasher, HashMap};
 use std::hash::{Hash, Hasher};
 use std::mem::size_of;
 
+// HashValue and HashIndex is intended to be same size with hash::PyHash
+// but it doesn't mean the values are compatible with actual pyhash value
```
coolreader18, line 15: Good explanation of this

vm/src/dictdatatype.rs
coolreader18: Oh, is this why you have the 2 different `key_hash` methods here/

youknowone: Yes, it work if I remove it. I kept it here because I wasn't sure it is intended or not.

And Yes, this is why we have duplicated hash. One for pure hash and another for wrapped key hash.

youknowone: would there be hashing quality problem? PyHash is basically 61bit integer.

coolreader18: I don't think so, you only get as much entropy/variation out as you put in, so if you put an integer in that you know will only have 2^61 possible values then the output will still only be one of 2^61 possible outputs.

youknowone: Reasonable. Do you think removing this part is better?

coolreader18: Yeah, I think it'd be best to do that; we already have mod_int for "normalizing" stuff to the same hash value.

PR #2037, opened 2020-07-28
Remove TryFromObject for std::ffi::CString

youknowone:
which is not used anymore


youknowone:
Oh.. ssl was disabled for my local build

PR #2038, opened 2020-07-28
flat_option -> flatten

youknowone:
by following std::option::Option

https://doc.rust-lang.org/std/option/enum.Option.html#method.flatten


coolreader18:
Oh, yikes, I merged this on my phone and I thought I wouldn't be able to if there were failing tests... :grimacing:

PR #2039, opened 2020-07-29
Allow None parameter for remove, index, count method of array

clemado1:
I changed return type to `Option` because when a parameter is None, `$t::try_from_object(vm, obj)` cause `TypeError` unlike CPython. But I'm not sure the way to compare value is good way.

### Before
```
ERROR: test_remove (__main__.UnsignedShortTest)
  File "Lib/test/test_array.py", line 955, in test_remove
    self.assertRaises(ValueError, a.remove, None)
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 746, in assertRaises
    context = None
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 743, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 743, in assertRaises
    return context.handle('assertRaises', args, kwargs)
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 181, in handle
    self = None
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 178, in handle
    callable_obj(*args, **kwargs)
  File "/home/clemado1/workspace/RustPython/Lib/unittest/case.py", line 178, in handle
    callable_obj(*args, **kwargs)
TypeError: Expected type <class 'int'>, not <class 'NoneType'>
```

### DIFF
--- a/vm/src/stdlib/array.rs
+++ b/vm/src/stdlib/array.rs
@@ -99,8 +99,9 @@ macro_rules! def_array_enum {
             fn count(&self, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
                 match self {
                     $(ArrayContentType::$n(v) => {
-                        let val = $t::try_from_object(vm, obj)?;
-                        Ok(v.iter().filter(|&&a| a == val).count())
+                        Ok(<Option<$t>>::try_from_object(vm, obj)?.map_or(0, |val| {
+                            v.iter().filter(|&&a| a == val).count()
+                        }))
                     })*
                 }
             }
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/array.rs
```diff
@@ -99,17 +99,17 @@ macro_rules! def_array_enum {
             fn count(&self, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
                 match self {
                     $(ArrayContentType::$n(v) => {
-                        let val = $t::try_from_object(vm, obj)?;
-                        Ok(v.iter().filter(|&&a| a == val).count())
+                        let val = <Option<$t>>::try_from_object(vm, obj)?;
+                        Ok(v.iter().filter(|&&a| Some(a) == val).count())
```
youknowone, line 103: I think this is the correct way to get the value of `val`.
Because any value in array cannot be `None`, I think the actual comparison is not required.
```suggestion
                        Ok(<Option<$t>>::try_from_object(vm, obj)?.map_or(0, |val| {
                            v.iter().filter(|&&a| a == val).count()
                        })
```

clemado1: Thank you for your advise 👍 I edit my commit :)

PR #2045, opened 2020-07-30
Add is_imported and is_annotation for symtable

Lynskylate:


compiler/src/symboltable.rs
```diff
@@ -97,10 +97,13 @@ pub struct Symbol {
     pub name: String,
     // pub table: SymbolTableRef,
     pub scope: SymbolScope,
+    // TODO(lynskylate@gmail.com): Use bitflags replace
```
youknowone, line 100: I don't want to go with TODO ownership in the code. How about creating and assigning an issue if you want to mark some issue for you?

compiler/src/symboltable.rs
```diff
@@ -460,7 +473,11 @@ impl SymbolTableBuilder {
     }
 
     fn scan_parameter(&mut self, parameter: &ast::Parameter) -> SymbolTableResult {
-        self.register_name(&parameter.arg, SymbolUsage::Parameter)
+        if let Some(_annotation) = &parameter.annotation {
+            self.register_name(&parameter.arg, SymbolUsage::AnnotationParameter)
+        } else {
+            self.register_name(&parameter.arg, SymbolUsage::Parameter)
+        }
```
youknowone, line 480:
```suggestion
        let usage = if parameter.annotation.is_some() {
            SymbolUsage::AnnotationParamater
        } else {
            SymbolUsage::Parameter
        };
        self.regsiter_name(&parameter.arg, usage)
```

vm/src/stdlib/symtable.rs
```diff
@@ -211,7 +249,32 @@ impl PySymbol {
 
     #[pymethod(name = "is_namespace")]
     fn is_namespace(&self) -> bool {
-        // TODO
-        false
+        self.namespaces.len() > 0
```
youknowone, line 252:
```suggestion
        self.namespaces.is_empty()
```

compiler/src/symboltable.rs
Lynskylate: Done.

compiler/src/symboltable.rs
Lynskylate: Done

youknowone:
@coolreader18 Will you review this PR? I am afread I don't know about symtable enough.

vm/src/stdlib/symtable.rs
```diff
@@ -129,12 +136,23 @@ impl PySymbolTable {
 
     #[pymethod(name = "get_symbols")]
     fn get_symbols(&self, vm: &VirtualMachine) -> PyResult {
-        let symbols = self
-            .symtable
-            .symbols
-            .values()
-            .map(|s| (PySymbol { symbol: s.clone() }).into_ref(vm).into_object())
-            .collect();
+        let mut symbols: Vec<PyObjectRef> = vec![];
```
coolreader18, line 139: Is there any reason this can't still be `self.symtable.symbols.values().map(...).collect()`? That's generally more idiomatic than `let mut v: Vec` and a for loop pushing values to it.

Lynskylate: Done, Thank you.
I'm just worried about whether too many chain calls will affect readability.

PR #2047, opened 2020-07-30
Windows build note for failing part

youknowone:
I don't use windows machine for RustPython, so please check if this is really helpful when you actually missed the step.


vm/src/frozen.rs
```diff
@@ -18,6 +18,7 @@ pub fn get_module_inits() -> HashMap<String, FrozenModule> {
     // Python modules that the vm calls into, but are not actually part of the stdlib. They could
     // in theory be implemented in Rust, but are easiest to do in Python for one reason or another.
     // Includes _importlib_bootstrap and _importlib_bootstrap_external
+    // For Windows: did you forget to run `powershell scripts\symlinks-to-hardlinks.ps1`?
```
coolreader18, line 21: Maybe put on the same line as the `pycompile_bytecode!()`? That way the error message will show the note.

youknowone: I want to put it somewhere and test how it shows but it is not eay without windows machine. Please put it to somewhere it need to be placed.

PR #2048, opened 2020-07-31
Implement os.waitpid on Windows

BasixKOR:
This PR implements os.waitpid using MSVCRT's [`_cwait`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/cwait) function.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1971,7 +1971,6 @@ mod posix {
         unsafe { libc::WEXITSTATUS(status) }
     }
 
-    // TODO: os.wait[pid] for windows
     #[pyfunction]
     fn waitpid(pid: libc::pid_t, opt: i32, vm: &VirtualMachine) -> PyResult<(libc::pid_t, i32)> {
         let mut status = 0;
@@ -2134,6 +2133,8 @@ mod nt {
     use super::*;
     pub(super) use std::os::windows::fs::OpenOptionsExt;
     use std::os::windows::io::RawHandle;
+    #[cfg(target_env = "msvc")]
+    use winapi::vc::vcruntime::intptr_t;
 
     pub(super) type OpenFlags = u32;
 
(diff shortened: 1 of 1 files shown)


BasixKOR:
This function has CPython test suite, although it's not possible to run it on Windows yet.

vm/src/stdlib/os.rs
```diff
@@ -2422,6 +2461,12 @@ fn extend_module_platform_specific(vm: &VirtualMachine, module: &PyObjectRef) {
         "_COPYFILE_DATA" => ctx.new_int(COPYFILE_DATA),
         "_fcopyfile" => ctx.new_function(os_fcopyfile),
     });
+
+    #[cfg(all(windows, target_env = "msvc"))]
+    extend_module!(vm, module, {
+        "wait" => ctx.new_function(os_wait),
+        "waitpid" => ctx.new_function(os_waitpid),
+    })
```
youknowone, line 2469: This is placed in unix `extend_module_platform_specific` for now. Moving this code a few lines below will be perfect.

vm/src/stdlib/os.rs
```diff
@@ -1996,6 +1995,46 @@ fn os_wait(vm: &VirtualMachine) -> PyResult<(libc::pid_t, i32)> {
...
+                "ECHILD: No spawned processes".to_owned(),
+            )),
+            EINVAL => Err(vm.new_exception_msg(
+                vm.ctx.exceptions.os_error.clone(),
+                "EINVAL: Invalid argument".to_owned(),
+            )),
+            _ => unreachable!(),
+        }
+    } else {
+        Ok((pid, status))
```
BasixKOR, line 2029: Oops, this should be lshifted by a byte.

vm/src/stdlib/os.rs
```diff
@@ -2290,6 +2291,46 @@ mod nt {
...
+            unsafe { _get_errno(&mut errno) };
+            match errno {
+                ECHILD => Err(vm.new_exception_msg(
+                    vm.ctx.exceptions.os_error.clone(),
+                    "ECHILD: No spawned processes".to_owned(),
+                )),
+                EINVAL => Err(vm.new_exception_msg(
+                    vm.ctx.exceptions.os_error.clone(),
+                    "EINVAL: Invalid argument".to_owned(),
+                )),
```
youknowone, line 2320: Following unix `convert_nix_error`, here could be `convert_nt_error`

BasixKOR: I'd love to use that kind of error handler, but it seems like in `_cwait` Microsoft uses some kind of POSIX-like errno which means it's neither WinAPI error nor POSIX nix error.

PR #2052, opened 2020-08-01
Add venv module

BolunThompson:
This is the venv module from CPython. If called without the --without-pip flag, it errors because the pip module isn't supported yet.


BolunThompson:
I took the venv module from Python 3.9 Beta 5. I changed the commit message accordingly.

coolreader18:
Thanks for contributing!

PR #2053, opened 2020-08-01
Add support for indexing and attribute access inside format strings

skinny121:
- Adds support for: `"{a.attr}".format(...)` and `"{l[0]}".format(...)`
- Also, adds the couple of underlying functions behind `string.Formatter`

For #1983

### DIFF
--- a/vm/src/format.rs
+++ b/vm/src/format.rs
@@ -856,7 +856,7 @@ impl FormatString {
                 auto_argument_index += 1;
                 arguments
                     .args
-                    .get(auto_argument_index)
+                    .get(auto_argument_index - 1)
                     .cloned()
                     .ok_or_else(|| vm.new_index_error("tuple index out of range".to_owned()))
             }
--- a/vm/src/format.rs
+++ b/vm/src/format.rs
@@ -2,8 +2,9 @@ use crate::function::PyFuncArgs;
 use crate::obj::objint::PyInt;
 use crate::obj::objstr::PyString;
 use crate::obj::{objstr, objtype};
-use crate::pyobject::{ItemProtocol, PyObjectRef, PyResult, TypeProtocol};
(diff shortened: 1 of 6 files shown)


vm/src/format.rs
```diff
@@ -724,86 +794,116 @@ impl FormatString {
...
+
+                    let mut argument = field_func(&field_type)?;
+
+                    for name_part in parts {
+                        match name_part {
+                            FieldNamePart::Attribute(attribute) => {
+                                argument = vm.get_attribute(argument, attribute.as_str())?;
+                            }
+                            FieldNamePart::Index(index) => {
+                                // TODO Implement DictKey for usize so we can pass index directly
```
youknowone, line 821: It will be easier on #2035

vm/src/obj/objlist.rs
```diff
@@ -47,6 +48,12 @@ impl From<Vec<PyObjectRef>> for PyList {
     }
 }
 
+impl FromIterator<PyObjectRef> for PyList {
+    fn from_iter<T: IntoIterator<Item = PyObjectRef>>(iter: T) -> Self {
+        Vec::from_iter(iter).into()
+    }
+}
```
youknowone, line 55: This is a really good idea.

vm/src/format.rs
```diff
@@ -567,32 +582,99 @@ impl FromStr for FormatSpec {
...
-impl FormatPart {
-    fn is_auto(&self) -> bool {
-        match self {
-            FormatPart::AutoSpec(_) => true,
-            _ => false,
+impl FieldNamePart {
+    fn parse_part(
+        chars: &mut impl PeekingNext<Item = char>,
+    ) -> Result<FieldNamePart, FormatParseError> {
+        let ch = chars.next().unwrap();
```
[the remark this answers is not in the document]
skinny121: I managed to avoid the need for this `unwrap` here by changing the return type to `Result<Option<FieldNamePart>, FormatParseError>`, which also made the use of this function more idiomatic.

PR #2057, opened 2020-08-02
Implement os.chmod on Windows

BasixKOR:
This fixes some broken test CPython cases on Windows.


vm/src/stdlib/os.rs
```diff
@@ -2291,6 +2291,27 @@ mod nt {
         get_stats().map_err(|e| convert_io_error(vm, e))
     }
 
+    #[pyfunction]
+    fn os_chmod(
```
youknowone, line 2295: `chmod` here. Otherwise `#[pyfunction(name = "chmod")]` is required.

PR #2061, opened 2020-08-02
Add target_is_directory parameter for symlink function

tobark:
`target_is_directory` parameter is required for `symlink` function.

https://docs.python.org/3/library/os.html#os.symlink


vm/src/stdlib/os.rs
```diff
@@ -156,6 +156,12 @@ pub fn errno_err(vm: &VirtualMachine) -> PyBaseExceptionRef {
     convert_io_error(vm, io::Error::last_os_error())
 }
 
+#[derive(FromArgs, Default)]
```
youknowone, line 159: clippy annotation wiill suppress the error

PR #2062, opened 2020-08-02
BufferedIO.read avoid clone

youknowone:


vm/src/stdlib/io.rs
```diff
@@ -83,28 +83,27 @@
...
         //for a defined number of bytes, i.e. bytes != -1
-        if bytes >= 0 {
-            let mut handle = self.cursor.clone().take(bytes as u64);
-            //read handle into buffer
-
-            if handle.read_to_end(&mut buffer).is_err() {
-                return None;
+        match bytes.and_then(|v| v.to_usize()) {
+            Some(bytes) => {
+                let mut buffer = Vec::with_capacity(bytes);
```
coolreader18, line 90: Are you sure this is safe? For fileio I think we just do a `vec![0; bufsize]`

youknowone: The length is set by next line unsafe `set_len` and `with_capacity` just help it a bit. If read_exact has any problem, regardless of zero-filled vector or garbage filled vector(with set_len), it causes problem.

coolreader18: Perhaps, though I'd prefer a vector with zeroes rather than UB, but if you think this is fine than feel free to use it.

youknowone: This is neither UB nor unsafe. Please tell me if I am missing something.

`Vec<T>::set_len` is a vector version of `MaybeUninit`, so it may cause UB by the element type `T` and capaticy. In this case, the type `T` is `u8`. so it does not cause any type related unsafe case. About the capacity, we set it in this line. So this is successfully avoiding both of the unsafe case. (https://doc.rust-lang.org/std/vec/struct.Vec.html#method.set_len)
By meeting 2 criteria of `set_len`, even in the worst case, the result is just garbage bytes in vector. Any wrong u8 value in safe vector is not related to UB.

The next steps are also guaranteed to be safe.
In document, the `Read` trait requires (though it is not forced, I believe `std` followed the requirements) the buffer must be read-only. (https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact)
So, we don't need to set it to any value because `read_exact` never use it. And after calling `read_exact` successfully, the vector will be full of new read data; otherwise the result will be `Err`.

Because `set_len` itself is not safe, I will put with_capacity into unsafe block and add a comment they must be coupled forever to help other people who will read this code next time.

PR #2063, opened 2020-08-03
Add test/test_sys.py

coolreader18:


vm/src/lib.rs
```diff
@@ -68,6 +68,7 @@ mod frozen;
 pub mod function;
 pub mod import;
 pub mod obj;
+mod py_io;
```
youknowone, line 71: py_io, py_serde but pyobject, pystr.
would there be a standard for the name?

vm/src/py_io.rs
```diff
@@ -0,0 +1,30 @@
...
+where
+    W: io::Write,
+{
+    type Error = io::Error;
+    fn write_fmt(&mut self, args: fmt::Arguments) -> io::Result<()> {
+        <W as io::Write>::write_fmt(self, args)
+    }
+}
+
+pub struct PyWriter<'vm>(pub PyObjectRef, pub &'vm VirtualMachine);
```
youknowone, line 21: nice idea

vm/src/lib.rs
coolreader18: Yeah, I noticed that too lol. I think `py_io`/`py_serde` make sense because it's a "py" version of io/serde, but then a "pyobject" is it's own sort of thing. Not too sure about pystr

youknowone: That make sense, but It seems not obvious so it would make confusion again in future.

PR #2064, opened 2020-08-03
BufferedIO OptionalSize, truncate and readline

youknowone:


vm/src/stdlib/io.rs
```diff
@@ -115,40 +128,50 @@ impl BufferedIO {
...
-            .flatten()
-            .filter(|s| *s >= 0)
-            .and_then(|s| s.try_into().ok())
-            .unwrap_or(usize::MAX);
-        if left == 0 {
-            return Ok(buf);
-        }
-        loop {
-            let (done, used) = {
-                let mut available = self.cursor.fill_buf().map_err(|err| os_err(vm, err))?;
```
youknowone, line 130: `fill_buf` read more data from reading source when the buffer is empty.
In this case, buffer is always the cursor itself and it doesn't have its background buffer source.
So we don't need to use loop here unless we are feeding new data into cursor ourselves.

skinny121: Ah true, it will return all of the remaining data. Might be good to add a comment to that affect.

vm/src/stdlib/io.rs
```diff
@@ -115,40 +128,50 @@ impl BufferedIO {
...
+        };
+        let buf = match available.find_byte(b'\n') {
+            Some(i) => (available[..=i].to_vec()),
+            _ => (available.to_vec()),
+        };
+        self.cursor.consume(buf.len());
+        Ok(buf)
+    }
+
+    fn truncate(&mut self, pos: Option<isize>, vm: &VirtualMachine) -> PyResult<()> {
```
skinny121, line 162: For consistency with other methods it's probably better to accept `Option<usize>` here and move the check for negative values into the caller, which is what is responsible to convert values from Python into this underlying interface.

vm/src/stdlib/io.rs
```diff
@@ -115,40 +128,50 @@ impl BufferedIO {
...
+        let pos = pos.map_or_else(
```
skinny121, line 163: Probably be nicer as a match.

vm/src/stdlib/io.rs
youknowone: Agree. I want to add a new method to `OptionalSize`, but I have no idea about the method name.
Do you have any suggestion not to be confused with `to_usize`?

skinny121: I'm also not sure amount the name, maybe `try_to_positive_usize`?

coolreader18: Well, usize is always positive anyway 🙃. But yeah, maybe just `to_size()` or something? It is a `[Optional]Size` anyway

youknowone: I used `try_usize`

PR #2067, opened 2020-08-03
Use pylib as a dependency of the rustpython binary in order to get a Lib path

coolreader18:
This should make sure that even if installed from crates.io, rustpython has access to the python standard library,
and it also allows access to the stdlib even when `-I` is passed, since the lib location is a `./configure` option
in CPython rather than it always assuming that PYTHONPATH will be set.


coolreader18:
I verified using `cargo package -l` that only the files we want should be packaged into the crate; just the python files not under a `test/` directory

Issue #1652, opened 2019-12-30
Issues with set iterators


dralley:
### Expected

```
>>> t = (1, 2, 3)
>>> l = [1, 2, 3]
>>> s = {1, 2, 3}
>>> d = {'a': 1, 'b': 2, 'c': 3}
>>> iter(t)
<tuple_iterator object at 0x7f8e1eedd0d0>
>>> iter(l)
<list_iterator object at 0x7f8e1edcecd0>
>>> iter(s)
<set_iterator object at 0x7f8e1edcf410>
>>> iter(d)
<dict_keyiterator object at 0x7f8e1ef4ce90>
>>> reversed(iter(l))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: 'list_iterator' object is not reversible
>>> i = iter(s)
>>> next(i)
1
>>> s.remove(2)
>>> next(i)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: Set changed size during iteration
>>> i = iter(d)
>>> next(i)
'a'
>>> del d['b']
>>> next(i)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
```

### Actual

```
>>>>> t = (1, 2, 3)
>>>>> l = [1, 2, 3]
>>>>> s = {1, 2, 3}
>>>>> s = {'a': 1, 'b': 2, 'c': 3}
>>>>> iter(s)
<list_iterator object at 0x55853bcaf510>       # <--- should be "set_iterator" object
>>>>> iter(t)
<tuple_iterator object at 0x55853bcaf510>
>>>>> iter(l)
<list_iterator object at 0x55853b881e10>
>>>>> iter(d)
<dict_keyiterator object at 0x56425a1aa9b0>
>>>>> reversed(iter(l))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: argument to reversed() must be a sequence    # <--- error can be improved
>>>>> s = {1, 2, 3}
>>>>> i = iter(s)
>>>>> next(i)
1
>>>>> s.remove(2)
>>>>> next(i)
2                               # <--- Should have returned RuntimeError
>>>>> i = iter(d)
>>>>> next(i)
'a'
>>>>> del d['b']
>>>>> next(i)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: dictionary changed size during iteration
```

[...]

PR #2068, opened 2020-08-04
Support pysetiterator.

hatee-hatee-hatee-ho:
related https://github.com/RustPython/RustPython/issues/1652

I just fixed @dralley 's implementaion a little.

### DIFF
--- a/vm/src/obj/objset.rs
+++ b/vm/src/obj/objset.rs
@@ -1,9 +1,10 @@
 /*
  * Builtin set type with a sequence of unique items.
  */
+use rustpython_common::rc::PyRc;
 use std::fmt;
 
-use super::objlist::PyListIterator;
+use super::objiter;
 use super::objtype::{self, PyClassRef};
 use crate::dictdatatype;
 use crate::function::{Args, OptionalArg};
@@ -66,7 +67,7 @@ impl PyValue for PyFrozenSet {
 
 #[derive(Default, Clone)]
 struct PySetInner {
-    content: SetContentType,
+    content: PyRc<SetContentType>,
 }
 
 impl PySetInner {
@@ -96,7 +97,7 @@ impl PySetInner {
 
     fn copy(&self) -> PySetInner {
         PySetInner {
-            content: self.content.clone(),
+            content: PyRc::new((*self.content).clone()),
         }
     }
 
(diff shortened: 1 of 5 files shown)


hatee-hatee-hatee-ho:
while fix this issue, I encountered a test failure.
https://github.com/RustPython/RustPython/blob/1249fab930b99e505099656f1a62be518df23303/Lib/test/test_set.py#L1808-L1819

In CPython's implementation, `set.update` seems not using `__iter__`.
https://github.com/python/cpython/blob/da4e09fff6b483fe858997da5599c25397107ca1/Objects/setobject.c#L891-L894

coolreader18:
I don't think that's exactly the problem; the code you linked to is specifically when the argument to update() is a dict, but a few lines below there's a call to `PyObject_GetIter()`. You could probably just mark that test as expectedFailure/skip for now; the fix for that is probably something in dictdatatype internals.

coolreader18:
Hmm, it is strange though that it's only failing now.

coolreader18:
Ohh, I see what you mean -- CPython skips iterating over the argument to update() when it's a set; it just merges them. Yeah, that would definitely be the fix to just add that optimization, if you'd like to.

vm/src/obj/objset.rs
```diff
@@ -2,8 +2,9 @@
  * Builtin set type with a sequence of unique items.
  */
 use std::fmt;
+use std::sync::Arc;
```
youknowone, line 5:
```suggestion
use rustpython_common::rc::PyRc;
```

skinny121:
That test case is for https://bugs.python.org/issue24583, which is a crash. So as long as it doesn't panic/hang, I think it's good.

vm/src/obj/objset.rs
hatee-hatee-hatee-ho: Thanks, I fixed it.

vm/src/obj/objset.rs
```diff
@@ -813,7 +818,68 @@ impl TryFromObject for SetIterable {
...
+impl fmt::Debug for PySetIterator {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        // TODO: implement more detailed, non-recursive Debug formatter
+        f.write_str("setiterator")
+    }
+}
+
+#[pyimpl]
+impl PySetIterator {
+    #[pymethod(name = "__next__")]
```
youknowone, line 842: `#[pymethod(magic)]` also works.
When funciton name is `f`, `magic` attribute turns the python attribute name into `__f__`.

vm/src/obj/objset.rs
```diff
@@ -813,7 +818,68 @@ impl TryFromObject for SetIterable {
     }
 }
 
+#[derive(Copy, Clone, Default)]
+struct SetSizeInfo {
+    size: Option<usize>,
```
youknowone, line 823: it seems this is intended to be None when this is invalidated (changed) but not used yet, right?

hatee-hatee-hatee-ho: Oh, I forgot to invalidate it... I just fixed it.

vm/src/obj/objset.rs
hatee-hatee-hatee-ho: Thanks, that looks great. I changed them to `#[pymethod(magic)]`.

youknowone:
Thanks for contributing!

PR #2069, opened 2020-08-04
Separate IntoPyObject and IntoPyResult

youknowone:
This separation removes a lot of unnessessary unwrapping.

### DIFF
--- a/vm/src/format.rs
+++ b/vm/src/format.rs
@@ -1,7 +1,7 @@
 use crate::exceptions::PyBaseExceptionRef;
 use crate::function::PyFuncArgs;
 use crate::obj::{objstr, objtype};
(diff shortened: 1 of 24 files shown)


coolreader18:
Oh, I've been thinking about this for so long but I never got around to it 🎉

vm/src/format.rs
```diff
@@ -823,7 +824,7 @@ impl FormatString {
                             }
                             FieldNamePart::Index(index) => {
                                 // TODO Implement DictKey for usize so we can pass index directly
-                                argument = argument.get_item(&index.into_pyobject(vm)?, vm)?;
+                                argument = argument.get_item(index, vm)?;
```
skinny121, line 827: The comment above can now be deleted.

youknowone:
will anyone review #2055 too?

PR #2070, opened 2020-08-04
Fix os module on wasi

coolreader18:


coolreader18:
Looks like macos is hanging; I'll merge and it should resolve on master

PR #2072, opened 2020-08-05
impl IntoPyException for FormatParseError

youknowone:


coolreader18:
Are you planning on impl-ing IntoPyResult for any `Result<T, E> where T: IntoPyObject, E: IntoPyException`? Or is `IntoPyException` just for organizing similar functions into a trait?

youknowone:
I am not going to make that impl. `Err(err.into_pyexception(vm))` will be enough for that use case (at least for a while).

PR #2077, opened 2020-08-07
Optimize int.__new__

youknowone:


vm/src/obj/objint.rs
```diff
@@ -750,38 +753,30 @@ struct IntToByteArgs {
...
-pub fn to_int(vm: &VirtualMachine, obj: &PyObjectRef, base: &BigInt) -> PyResult<BigInt> {
-    let base = match base.to_u32() {
-        Some(base) if base == 0 || (2..=36).contains(&base) => base,
-        _ => return Err(vm.new_value_error("int() base must be >= 2 and <= 36, or 0".to_owned())),
-    };
-
-    let bytes_to_int = |bytes: &[u8]| {
-        std::str::from_utf8(bytes)
-            .ok()
-            .and_then(|s| str_to_int(s, base))
```
coolreader18, line 762: This looks like it's basically the same function as to_int_radix, is there a reason you can just call that with base as 10?

youknowone: base 10 is the most common case, so it will be worth to optimize it.

2 technical reasons:
- By passing constant, bytes_to_int (hopefully) will by optimized for 10; otherwise we can manually work for it later.
- `to_int_radix` process only string, bytes and bytearray. it skips other types.

coolreader18: Hmm, ok. You could probably have to_int_radix return an actual python error in the `_` case, rather than checking in `int.__new__` and having an unreachable in to_int_radix.

youknowone: Oh, yep. Actually that was done seconds ago after checking priority of `val` and `base`

coolreader18: Oh, I see that lol

vm/src/obj/objint.rs
```diff
@@ -810,79 +786,147 @@ pub fn to_int(vm: &VirtualMachine, obj: &PyObjectRef, base: &BigInt) -> PyResult
...
-            if buf.chars().all(|c| matches!(c, '+' | '-' | '0' | '_')) {
-                return Some(BigInt::zero());
+    // remove zeroes
+    let mut last = *lit.first()?;
+    if last == b'0' {
+        let mut count = 0;
+        for pair in lit.windows(2) {
+            // get_unchecked below is safe because windows guarantees
+            let c1 = unsafe { pair.get_unchecked(0) };
+            let c2 = unsafe { pair.get_unchecked(1) };
```
coolreader18, line 876: not sure if this is necessary, it should be able to optimize away bounds checks for stuff like `windows()` and `chunks_exact()`:
```rust
fn foo(x: &[u8]) {
    for win in x.windows(2) {
        print(win[0]);
        print(win[1]);
    }
}
```
```asm
foo:
        push    r14
        push    rbx
        push    rax
        cmp     rsi, 2
        jb      .LBB0_3
        mov     r14, rsi
        mov     rbx, rdi
        add     rbx, 1
.LBB0_2:
        add     r14, -1
        movzx   edi, byte ptr [rbx - 1]
        call    example::print
        movzx   edi, byte ptr [rbx]
        call    example::print
        add     rbx, 1
        cmp     r14, 2
        jae     .LBB0_2
.LBB0_3:
        add     rsp, 8
        pop     rbx
        pop     r14
        ret
```
https://godbolt.org/z/YvxPrh

coolreader18: [playground link](https://play.rust-lang.org/?version=stable&mode=release&edition=2018), hit "asm" under the "..." next to "Build"; there's a bounds check for
```rust
        if lit.first() == Some(&b'_') {
            lit = &lit[1..]; // <<<<
        }
```
```rust
        // get_unchecked below is safe because count < lit.len() here
        let prefix_last = unsafe { *lit.get_unchecked(count) };
        lit = &lit[count + 1..]; // <<<<
```

youknowone: the link doesn't include code (share link?), but I got what you meant. that's unfortune.

coolreader18: Oh, lol, I'm an idiot. https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=d1eb4f68631167b6155c0b149f96825b

coolreader18: I figured out the line numbers by looking at the data sections in the bottom, and then the ones that referenced `L__unnamed_8` ("src/lib.rs") had a non-zero byte in the middle (`>` in `"\n\000\000\000\000\000\000\000>\000\000\000\024\000\000"`), which was the line number (`ord('>') == 62 and ord('Z') == 90`). Note that when I removed the get_unchecked on line 89, there was still just one bounds check because of the line after.

youknowone: I have no idea how did you read bound check from there. Does bound check included in "src/lib.rs"? In that case, I wish this is better:
https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=b89d48f636eb6989dcfd5b8b8ee8910c

coolreader18: There are a few labels (e.g. `.LBB0_65`) near the bottom that places in the rest of the code jump to when a bounds check fails; those reference data labels at the very bottom like `.L__unnamed_5`:
```asm
.L__unnamed_8:
	.ascii	"src/lib.rs"

.L__unnamed_1:
	.byte	48

.L__unnamed_2:
	.byte	95

.L__unnamed_6:
	.quad	.L__unnamed_8
	.asciz	"\n\000\000\000\000\000\000\000>\000\000\000\024\000\000"
;                                              ^ lineno     ^^^^ column, i think

.L__unnamed_5:
	.quad	.L__unnamed_8
	.asciz	"\n\000\000\000\000\000\000\000`\000\000\000\033\000\000"
;                                              ^ lineno     ^^^^ column, i think
```
I know that that panic location is in our code because they reference `.L__unnamed_8` (src/lib.rs), and I figured out the location from the byte in the middle of the data string.

vm/src/obj/objint.rs
```diff
@@ -810,79 +786,160 @@ pub fn to_int(vm: &VirtualMachine, obj: &PyObjectRef, base: &BigInt) -> PyResult
...
+    if has_radix {
+        lit = unsafe {
+            // safe by radix check
+            lit.get_unchecked(2..)
+        };
+        if lit.first() == Some(&b'_') {
+            lit = unsafe {
+                // safe by b'_' check
+                lit.get_unchecked(1..)
+            };
```
coolreader18, line 874:
```suggestion
        lit = &lit[2..];
        if lit.first()? == &b'_' {
            lit = &lit[1..];
```
This has no bounds checks; if there's nothing after `0x` it should be an error anyway

youknowone: nIce catch!

vm/src/obj/objint.rs
```diff
@@ -810,79 +786,154 @@ pub fn to_int(vm: &VirtualMachine, obj: &PyObjectRef, base: &BigInt) -> PyResult
...
+    let sign = match lit.first()? {
+        b'+' => Some(Sign::Plus),
+        b'-' => Some(Sign::Minus),
+        _ => None,
+    };
+    if sign.is_some() {
+        lit = unsafe {
+            // safe by sign check
+            lit.get_unchecked(1..)
+        };
```
coolreader18, line 833: This was fine as well as it was

youknowone: yes, it seems `lit.first()?` is covered well unlike `Some()` branch

youknowone: Actually, I think using `get_unchecked` is better when we are not sure it is properly optimized or not.
Still going to this way because I think this project is not on the performance optimization phase.

coolreader18:
I think that this is an optimization bug that will hopefully be fixed. This code:
```rust
const TOO_SMALL: usize = 0;

pub fn foo(foo: &[u8]) -> u8 {
    if foo.len() <= TOO_SMALL {
        return u8::MAX
    }
    let mut count = 0;
    for _ in &foo[1..] {
        count += 1;
    }
    foo[count]
}
```
has a bounds check in it. However, when TOO_SMALL is 1, it doesn't, even though it will never be out of bounds even when TOO_SMALL is 0. [playground](https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=a4a83d903be0510f95e233e1522f6a4f). I'll open an issue to the rust repo and hopefully this will be fixed at some point, but I think for now something like the following is fine for now.
```rust
    // remove zeroes
    let mut last = *lit.first()?;
    if last == b'0' {
        let mut count = 0;
        for &cur in &lit[1..] {
            if cur == b'_' {
                if last == b'_' {
                    return None;
                }
            } else if cur != b'0' {
                break
            };
            count += 1;
            last = cur;
        }
        // get_unchecked below is safe because count < lit.len() here
        let prefix_last = lit[count];
        lit = &lit[count + 1..];
        if lit.is_empty() && prefix_last == b'_' {
            return None;
        }
    }

    let mut last = *lit.first()?;
    // validate
    for c in lit.iter() {
// ...
```

PR #2078, opened 2020-08-07
vm.new_pystr and vm.ctx.new_str

youknowone:
- Remove duplicated vm.new_blah and add vm.new_object for generic usage
- Allow `&str` and `&String` for `vm.ctx.new_str`


coolreader18:
Not totally sure about the naming of `new_object()`, I feel like that would imply something like the default object constructor (`object()`). Maybe to_obj instead?

youknowone:
Will `new_obj` or `new_pyobj` work? Its role is exactly same as `vm.new_*` functions. I want to give a similar name for similar feature.

coolreader18:
Yeah, that make sense, I think `new_pyobj()` sounds better

PR #2079, opened 2020-08-07
small int cache for int.__new__

youknowone:


vm/src/obj/objint.rs
```diff
@@ -78,6 +78,23 @@ impl PyValue for PyInt {
     fn class(vm: &VirtualMachine) -> PyClassRef {
         vm.ctx.int_type()
     }
+
+    fn into_simple_object(self, vm: &VirtualMachine) -> PyObjectRef {
+        vm.ctx.new_bigint(&self.value)
```
coolreader18, line 83:
```suggestion
        vm.ctx.new_int(self.value)
```
I think `new_bigint` should probably be called `new_bigint_ref`, right? Since it's only for if you have a bigint reference that you might not want to clone if it fits into an int cache.

vm/src/obj/objint.rs
```diff
@@ -78,6 +78,23 @@ impl PyValue for PyInt {
...
+    }
+
+    fn into_object(
+        self,
+        vm: &VirtualMachine,
+        cls: PyClassRef,
+        dict: Option<PyDictRef>,
+    ) -> PyObjectRef {
+        if cls.is(&Self::class(vm)) {
+            vm.ctx.new_bigint(&self.value)
```
coolreader18, line 93:
```suggestion
            vm.ctx.new_int(self.value)
```

vm/src/obj/objint.rs
youknowone: Yes, exactly,

PR #2080, opened 2020-08-08
Implement the rest of the struct module

skinny121:


vm/src/byteslike.rs
```diff
@@ -35,6 +52,44 @@ impl PyBytesLike {
         match self {
             PyBytesLike::Bytes(b) => f(b.get_value()),
             PyBytesLike::Bytearray(b) => f(&b.borrow_value().elements),
+            PyBytesLike::Array(array) => f(&*array.get_bytes()),
+        }
+    }
+}
+
+pub enum PyBuffer {
```
youknowone, line 60: Oh, buffer protocol is splited from `PyBytesLike`, right?

vm/src/stdlib/array.rs
```diff
@@ -137,14 +140,24 @@ macro_rules! def_array_enum {
                 }
             }
 
-            fn tobytes(&self) -> Vec<u8> {
+            fn get_bytes(&self) -> &[u8] {
```
youknowone, line 143: :+1:

vm/src/stdlib/pystruct.rs
```diff
@@ -160,36 +190,39 @@ mod _struct {
 
     /// Parse endianness
     /// See also: https://docs.python.org/3/library/struct.html?highlight=struct#byte-order-size-and-alignment
-    fn parse_endiannes<I>(chars: &mut Peekable<I>) -> Endianness
+    fn parse_size_and_endiannes<I>(chars: &mut Peekable<I>) -> (bool, Endianness)
```
youknowone, line 193: Probably enum fits better than bool here.

vm/src/stdlib/pystruct.rs
```diff
@@ -72,26 +74,56 @@ mod _struct {
...
-    #[derive(Debug)]
+    #[derive(Debug, Clone)]
     struct FormatSpec {
         endianness: Endianness,
         codes: Vec<FormatCode>,
     }
 
     impl FormatSpec {
+        fn decode_and_parse(
+            vm: &VirtualMachine,
```
coolreader18, line 85: If you'd like, you could probably use `AsciiStr` from the `ascii` crate to simplify parsing/validating fmt strings.

coolreader18: That would be a lot though, probably best left to a separate PR

vm/src/byteslike.rs
skinny121: They are pretty much the same except `PyBytesLike` is read only. Looking at https://docs.python.org/3/glossary.html#term-bytes-like-object, this probably should be named `PyRWBytesLike` instead.

vm/src/stdlib/pystruct.rs
skinny121: Yeah will do it in a separate PR.

vm/src/stdlib/pystruct.rs
```diff
@@ -244,6 +286,58 @@ mod _struct {
...
+                let msg = format!(
+                    "{op} requires a buffer of at least {required} bytes for {op_action} {needed} \
+                    bytes at offset {offset} (actual buffer size is {buffer_len})",
+                    op = if is_pack { "pack_into" } else { "unpack_from" },
+                    op_action = if is_pack { "packing" } else { "unpacking" },
+                    required = offset + buffer_len as isize,
+                    needed = needed,
+                    offset = offset,
+                    buffer_len = buffer_len
+                );
```
youknowone, line 318: To clarify `op` and `op_action` is decided by same condition `is_pack`,
```rust
let (op, op_action) = if is_pack {
    ...
} else {
   ...
};
let msg = format!(“...”)
```

vm/src/stdlib/pystruct.rs
```diff
@@ -244,6 +286,58 @@ mod _struct {
...
+                    vm,
+                    format!(
+                        "offset {} out of range for {}-byte buffer",
+                        offset, buffer_len
+                    ),
+                ));
+            }
+            buffer_len - (-offset as usize)
+        } else {
+            if offset as usize >= buffer_len {
```
youknowone, line 308: Unlike then-block, offset is always `usize` in else-block.
```rust
let offset = offset as usize
```

PR #2081, opened 2020-08-08
Fix problem that cannot check overflow in len function.

hyperbora:


vm/src/obj/objsequence.rs
```diff
@@ -291,6 +291,11 @@ pub fn opt_len(obj: &PyObjectRef, vm: &VirtualMachine) -> Option<PyResult<usize>
         if len.is_negative() {
             return Err(vm.new_value_error("__len__() should return >= 0".to_owned()));
         }
+        if len.to_isize().is_none() {
+            return Err(
+                vm.new_overflow_error("cannot fit 'int' into an index-sized integer".to_owned())
+            );
+        }
```
youknowone, line 298: By using `to_isize`, we already paid narrowing cost. Reusing this value will be better for next.
```rust
let len = if let Some(len) = len.to_isize() {
    len
} else {
    return ...
};

hyperbora: OK, I fixed it. :)

PR #2082, opened 2020-08-08
Fix objtuple.tp_new so that return cloned when a parameter is tuple

clemado1:
Fix `objtuple.rs` `fn tp_new()` so that return cloned object when a parameter is tuple object.

```python
# test_tuple.py
def test_constructors(self):
    t0_3 = (0, 1, 2, 3)
    t0_3_bis = tuple(t0_3)
    self.assertTrue(t0_3 is t0_3_bis) # to pass this test
```

### DIFF
--- a/vm/src/obj/objtuple.rs
+++ b/vm/src/obj/objtuple.rs
@@ -6,9 +6,9 @@ use super::objsequence::get_item;
 use super::objtype::PyClassRef;
 use crate::function::OptionalArg;
 use crate::pyobject::{
-    self, BorrowValue, IntoPyObject,
+    self, BorrowValue, IdProtocol, IntoPyObject,
     PyArithmaticValue::{self, *},
-    PyClassImpl, PyComparisonValue, PyContext, PyObjectRef, PyRef, PyResult, PyValue,
+    PyClassImpl, PyComparisonValue, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
 };
 use crate::sequence::{self, SimpleSeq};
 use crate::vm::{ReprGuard, VirtualMachine};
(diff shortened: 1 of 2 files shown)


vm/src/obj/objtuple.rs
```diff
@@ -234,6 +235,10 @@ impl PyTuple {
         vm: &VirtualMachine,
     ) -> PyResult<PyTupleRef> {
         let elements = if let OptionalArg::Present(iterable) = iterable {
+            if cls.is(&vm.ctx.tuple_type()) && iterable.lease_class().is(&vm.ctx.tuple_type()) {
```
youknowone, line 238: Should both class be tuple type? I guess we just need to check `iterable.lease_class().is(cls)` because we can reuse the value when the type is same.

vm/src/obj/objtuple.rs
```diff
@@ -234,6 +235,10 @@ impl PyTuple {
...
+                return PyTupleRef::try_from_object(vm, iterable.clone());
```
youknowone, line 239: `try_from_object` checks type matches, but we already know iterable is compatible with return value.
Using `PyObject::into_pyref()` will remove the typecheck cost.
```suggestion
                return iterable.into_pyref();
```

coolreader18: That wouldn't work, that only works for `PyRc<PyObject<PyTuple>>`, not for any PyObjectRef. I think this is fine for now, we could maybe add a method downcast_exact to PyObjectRef if we want this.

youknowone: Ah, ok. In that case, I think changing return type to `PyObjectRef` would be the best. If it is too much, to track this issue in future, adding a comment including link to this issue will be great.

clemado1:
Thank you for reviews, I changed both of line. And in the second issues line, I changed it to `Ok(PyRef::from_obj_unchecked(iterable))' to avoid useless cost.

youknowone:
That was a nice catching

PR #2083, opened 2020-08-08
Unified value borrow interface for PyValue

youknowone:
This trait will allow `PyObjectRef::borrow_value()` which probably be useful to replace `obj*::get_value`.

```rust
    pub fn borrow_value<'a, B, T>(&'a self) -> Option<B> where
        T: PyObjectPayload + BorrowValue<'a, Borrowed=B>,
        B: 'a + Deref,
    {
        self.payload::<T>().map(BorrowValue::borrow_value)
    }
```

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -1337,6 +1337,11 @@ pub trait PyValue: fmt::Debug + PyThreadingConstraint + Sized + 'static {
     }
 }
 
+pub trait BorrowValue<'a>: PyValue {
+    type Borrowed: 'a + Deref;
+    fn borrow_value(&'a self) -> Self::Borrowed;
+}
+
 pub trait PyObjectPayload: Any + fmt::Debug + PyThreadingConstraint + 'static {
     fn as_any(&self) -> &dyn Any;
 }
--- a/vm/src/stdlib/binascii.rs
+++ b/vm/src/stdlib/binascii.rs
@@ -24,7 +24,7 @@ mod decl {
                 b @ PyBytes => Ok(SerializedData::Bytes(b)),
                 b @ PyByteArray => Ok(SerializedData::Buffer(b)),
                 a @ PyString => {
-                    if a.as_str().is_ascii() {
+                    if a.borrow_value().is_ascii() {
                         Ok(SerializedData::Ascii(a))
                     } else {
                         Err(vm.new_value_error(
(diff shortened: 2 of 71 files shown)


vm/src/pyobject.rs
```diff
@@ -1328,6 +1328,11 @@ pub trait PyValue: fmt::Debug + PyThreadingConstraint + Sized + 'static {
     }
 }
 
+pub trait BorrowValue<'a>: PyValue {
+    type Borrowed: 'a + Deref;
```
coolreader18, line 1332: Clever way of getting around a lack of GATs!

Should this look something like:
```rust
    type Target;
    type Borrowed: 'a + Deref<Target = Self::Target> = &'a Self::Target;
```
maybe?

vm/src/stdlib/binascii.rs
```diff
@@ -44,9 +44,9 @@ mod decl {
         #[inline]
         pub fn with_ref<R>(&self, f: impl FnOnce(&[u8]) -> R) -> R {
             match self {
-                SerializedData::Bytes(b) => f(b.get_value()),
```
coolreader18, line 47: You could implement BorrowValue for PyBytesLike, maybe? Maybe with `Borrowed` as an enum `LockOrRef` that looks like:

```rust
enum LockOrRef<'a, T> {
    Ref(&'a T),
    MuLock(PyMutexGuard<'a, T>),
    ReadLock(PyRwLockReadGuard<'a, T>),
}
impl Deref for LockOrRef { ... }
```
Maybe somewhere in rustpython_common?

coolreader18: Maybe BorrowValue could even go in rustpython_common? It is very generic.

vm/src/pyobject.rs
youknowone: I found `type Target` is redundant. It can be replaced with `<Self::Borrowed as Deref>::Target`

```rust
pub trait BorrowValue<'a>: PyValue {
    type Borrowed: 'a + Deref;
    fn borrow_value(&'a self) -> Self::Borrowed;
    fn clone_value(&'a self) -> <Self::Borrowed as Deref>::Target
    where
        <Self::Borrowed as Deref>::Target: Clone,
    {
        self.borrow_value().clone()
    }
}
```

vm/src/stdlib/binascii.rs
youknowone: It sounds nice. But I want to concentrate on simple replacement in this PR

coolreader18: Sounds good

PR #2084, opened 2020-08-09
Add PyAsciiBytesLike

skinny121:
This accepts either an string with only ascii characters or any bytes like object.

### DIFF
--- a/vm/src/byteslike.rs
+++ b/vm/src/byteslike.rs
@@ -1,5 +1,6 @@
 use crate::obj::objbytearray::{PyByteArray, PyByteArrayRef};
 use crate::obj::objbytes::{PyBytes, PyBytesRef};
+use crate::obj::objstr::{PyString, PyStringRef};
 use crate::pyobject::PyObjectRef;
 use crate::pyobject::{PyResult, TryFromObject, TypeProtocol};
 use crate::stdlib::array::{PyArray, PyArrayRef};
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/binascii.rs
```diff
@@ -2,55 +2,13 @@ pub(crate) use decl::make_module;
...
-                        ))
-                    }
-                }
-                obj => Err(vm.new_type_error(format!(
-                    "argument should be bytes, buffer or ASCII string, not '{}'",
-                    obj.class().name,
-                ))),
-            })
-        }
-    }
```
coolreader18, line 41: Maybe keep this? binascii does allow a bytearray, while struct doesn't:

```python
>>> binascii.a2b_hex(1)
TypeError: argument should be bytes, buffer or ASCII string, not 'int'
>>> struct.Struct(bytearray())
TypeError: Struct() argument 1 must be a str or bytes object, not bytearray
```

vm/src/byteslike.rs
```diff
@@ -93,3 +94,47 @@ impl PyRwBytesLike {
...
+            Ok(string) => {
+                if string.as_str().is_ascii() {
+                    Ok(PyAsciiBytesLike::String(string))
+                } else {
+                    Err(vm.new_value_error(
+                        "string argument should contain only ASCII characters".to_owned(),
+                    ))
+                }
+            }
+            Err(obj) => PyBytesLike::try_from_object(vm, obj).map(PyAsciiBytesLike::Buffer),
```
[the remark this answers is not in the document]
skinny121: We don't need to check if it's ascii, as all usages error if they encounter any characters that they don't recognise including bytes out of the ascii range. That also means we could get away with not checking the string is ascii as the bytes are the utf-8 encoding of it, so any bytes less than 128 match the ascii characters, and any bytes that are 128 and above are a part of an non-ascii character.

vm/src/stdlib/binascii.rs
skinny121: `PyAsciiBytesLike` aligns with what `binascii` accept more than this e.g. the following is allowed: `binascii.a2b_base64(array.array('B', b'1111'))`. I'm also not sure if `struct.Struct(bytearray())` should be disallowed or not. there is a comment about allowing it: https://github.com/python/cpython/blob/67acf74c4eaf64a860cc1bcda6efe6e9cb01f89b/Modules/_struct.c#L1466

vm/src/stdlib/pystruct.rs
```diff
@@ -50,19 +48,19 @@ mod _struct {
...
-                'q' | 'Q' | 'd' => 8,
-                'n' | 'N' | 'P' => std::mem::size_of::<usize>(),
+                b'x' | b'c' | b'b' | b'B' | b'?' | b's' | b'p' => 1,
+                b'h' | b'H' => 2,
+                b'i' | b'l' | b'I' | b'L' | b'f' => 4,
+                b'q' | b'Q' | b'd' => 8,
+                b'n' | b'N' | b'P' => std::mem::size_of::<usize>(),
                 c => {
-                    panic!("Unsupported format code {:?}", c);
+                    panic!("Unsupported format code {:?}", c as char);
```
whjpji, line 63: How about throw an exception (like `ValueError`) instead of panic?

PR #2085, opened 2020-08-09
Don't panic when a os function isn't supported on a platform

coolreader18:


vm/src/stdlib/os.rs
```diff
@@ -1073,8 +1082,8 @@ mod posix {
     use nix::unistd::getgroups;
 
     #[cfg(target_os = "redox")]
-    fn getgroups() -> nix::Result<Vec<Gid>> {
-        unimplemented!("redox getgroups")
+    fn getgroups(vm: &VirtualMachine) -> nix::Result<Vec<Gid>> {
+        os_unimpl("os.getgroups", vm)
```
youknowone, line 1086: `on_unimpl` returns `PyResult` but this function return `nix::Result`

coolreader18: Oops, nice catch

PR #2090, opened 2020-08-09
BorrowValue auto impl copy/clone helper from borrow

youknowone:
copy seems ok because `*v.borrow_value()` is annoying.

I am considering to revert owned and clone. It doesn't give that much values

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -1353,7 +1353,7 @@ pub trait BorrowValue<'a>: PyValue {
         *self.borrow_value()
     }
 
-    fn clone_value(&'a self) -> <Self::Borrowed as Deref>::Target
+    fn cloned_value(&'a self) -> <Self::Borrowed as Deref>::Target
     where
         <Self::Borrowed as Deref>::Target: Clone,
     {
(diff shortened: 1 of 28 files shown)


vm/src/pyobject.rs
```diff
@@ -1339,7 +1339,30 @@ pub trait PyValue: fmt::Debug + PyThreadingConstraint + Sized + 'static {
...
+    where
+        <Self::Borrowed as Deref>::Target: Clone,
+    {
+        self.borrow_value().clone()
+    }
+
+    fn owned_value<Owned>(&'a self) -> Owned
+    where
+        Owned: std::borrow::Borrow<<Self::Borrowed as Deref>::Target>,
+        <Self::Borrowed as Deref>::Target: ToOwned<Owned = Owned>,
```
coolreader18, line 1362: I think you don't need a generic for this, can you do `-> <<Self::Borrowed as Deref>::Target as ToOwned>::Owned where <Self:: Borrowed as Deref>::Target: ToOwned` ?

youknowone:
@skinny121 @coolreader18 how do you think about `owned_value` and `cloned_value`? keep or remove?

coolreader18:
Not sure about `owned_value`, I feel like it would be more clear to just do `.borrow_value().to_owned()`; I like `cloned_value()` more, though maybe it could be called `clone_value()`?

coolreader18:
Although maybe I just don't like owned_value because of the name, I feel like it doesn't communicate that you're converting the value to its owned form. Maybe like `borrow_to_owned()` or something?

PR #2093, opened 2020-08-09
Disable certain CPython tests on Windows

BasixKOR:
This patch allows running CPython tests fully on Windows. It doesn't fix failing tests yet.

### DIFF
--- a/Lib/test/test_exceptions.py
+++ b/Lib/test/test_exceptions.py
@@ -1190,7 +1190,8 @@ class ExceptionTests(unittest.TestCase):
         else:
             self.fail("RecursionError not raised")
         self.assertEqual(wr(), None)
-
+    
+    @unittest.skipIf(sys.platform == 'win32', 'error specific to cpython')
     def test_errno_ENOTDIR(self):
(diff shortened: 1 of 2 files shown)


Lib/test/test_exceptions.py
```diff
@@ -1187,7 +1190,8 @@ def inner():
         else:
             self.fail("RecursionError not raised")
         self.assertEqual(wr(), None)
-
+    
+    @unittest.skipIf(sys.platform == 'win32', 'error specific to cpython')
```
youknowone, line 1194: What do you mean by "error specific to cpython"?

BasixKOR: > Issue #12802: "not a directory" errors are ENOTDIR even on Windows 

It seems like CPython has a bug that errno returns `ENOTDIR` on Windows.

BasixKOR:
Yeah, sily mistake 😅 I'll fix it when I rebase it to master.

BasixKOR:
Converting to a draft for `test_os.py`

BasixKOR:
test_os got fixed by #2095

PR #2096, opened 2020-08-09
Enhance pyclass/pymodule

youknowone:
- Fix duplicated attribute checker for pyclass/pymodule
- `pyattr` for values from const and calcaulated (from function)
- py* macro doesn't allow mismatched item type
- pymodule inject `__module__` field of member class
- `pyattr` can explicitly create attribute for `pyclass` in `pymodule`
  - if function or method require similar work, it can be expanded.

- worked on a few modules with upper features

### DIFF
--- a/derive/src/pymodule.rs
+++ b/derive/src/pymodule.rs
@@ -1,17 +1,9 @@
 use super::Diagnostic;
-use crate::util::{def_to_name, ItemIdent, ItemMeta, ItemType};
+use crate::util::{def_to_name, meta_into_nesteds, ItemIdent, ItemMeta, ItemType};
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{quote, quote_spanned, ToTokens};
 use std::collections::HashMap;
 use syn::{parse_quote, spanned::Spanned, AttributeArgs, Ident, Item, Meta, NestedMeta};
-
-fn meta_to_vec(meta: Meta) -> Result<Vec<NestedMeta>, Meta> {
-    match meta {
-        Meta::Path(_) => Ok(Vec::new()),
-        Meta::List(list) => Ok(list.nested.into_iter().collect()),
-        Meta::NameValue(_) => Err(meta),
-    }
-}
 
 #[derive(Default)]
 struct Module {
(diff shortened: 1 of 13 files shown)


derive/src/pymodule.rs
```diff
@@ -1,50 +1,73 @@
...
-        Meta::NameValue(_) => Err(meta),
-    }
-}
+use std::collections::HashMap;
+use syn::{parse_quote, spanned::Spanned, AttributeArgs, Ident, Item, Meta, NestedMeta};
 
 #[derive(Default)]
 struct Module {
-    items: HashSet<(ModuleItem, Vec<Meta>)>,
+    items: HashMap<(String, Vec<Meta>), ModuleItem>,
```
coolreader18, line 10: Do we still have to have the `Vec<Meta>` in the key if we're distinguishing by name now?

derive/src/util.rs
```diff
@@ -59,23 +59,41 @@ pub fn optional_attribute_arg(
...
 }
 
-pub fn module_class_name(mod_name: Option<String>, class_name: &str) -> String {
+pub(crate) fn module_class_name(mod_name: Option<String>, class_name: &str) -> String {
     if let Some(mod_name) = mod_name {
         format!("{}.{}", mod_name, class_name)
     } else {
         class_name.into()
     }
 }
```
coolreader18, line 69: We can just use the `str.strip_prefix()` method, it was stabilized in 1.45

derive/src/pymodule.rs
youknowone: Yes, we still need it. `Vec<Meta>` is used to allow duplicated names when it has different cfgs. We use it to allow same named function exists one with `#[unix]` and the other one with `#[windows]`

derive/src/util.rs
youknowone: Sorry, I didn't get what you want to tell. It looks like optional concatnation rather than stripping.

coolreader18: Oh, whoops, meant to comment this on the `util::strip_prefix` function; the app I use isn't great about being precise like that :sweat_smile:

Issue #578, opened 2019-03-01
Current sorting algorithm isn't stable


adrian17:
Kinda missed that in #559. The algorithm needs to be changed to something stable.

evilpie:
I think Python famously uses [Timsort](https://en.wikipedia.org/wiki/Timsort), so that sort algorithm should probably be used.

OddCoincidence:
I don't think the particular sorting algorithm matters as long as it produces a stable sort and has roughly similar performance characteristics to python's. I wonder if we could get rust to add a `try_sort_by` method to slices so that we could just use the algo from `std`.

dralley:
Rust uses a sorting algorithm that is basically a variation of Timsort.

https://doc.rust-lang.org/std/primitive.slice.html#method.sort

youknowone:
#1072

PR #2097, opened 2020-08-10
Corrected the problem of mishandling optional kwarg in max function

hyperbora:


vm/src/builtins.rs
```diff
@@ -440,9 +449,17 @@ mod decl {
...
         if candidates.is_empty() {
-            let default = args.get_optional_kwarg("default");
             return default
                 .ok_or_else(|| vm.new_value_error("max() arg is an empty sequence".to_owned()));
+        } else {
+            let valid_keyword = vec!["default", "key"];
+            for (k, _) in args.kwargs.iter() {
+                if !valid_keyword.contains(&k.as_str()) {
+                    return Err(vm.new_type_error(
+                        format!("'{}' is an invalid keyword argument for max()", k).to_owned(),
```
youknowone, line 459: Because `format!` returns `String` object, this `.to_owned()` is redundant. That's what clippy says even though its message is `redundant clone`

vm/src/builtins.rs
```diff
@@ -430,8 +430,17 @@ mod decl {
 
     #[pyfunction]
     fn max(vm: &VirtualMachine, args: PyFuncArgs) -> PyResult {
+        let default = args.get_optional_kwarg("default");
```
youknowone, line 433: How about using `take_keyword` here? Then we can check invalid keyword just by testing it is empty instead of loop.

hyperbora:
[https://github.com/python/cpython/blob/master/Objects/listsort.txt](https://github.com/python/cpython/blob/master/Objects/listsort.txt)

```python
self.assertEqual(max(data, key=f),
                         sorted(reversed(data), key=f)[-1])
```

Cpython uses the timsort algorithm when sorting the list, but Rustpython uses the quicksort algorithm. In the end, the test fails. So I don't think I can fix it anymore. May I close the PR?

youknowone:
The sorting algorithm problem maybe related to #578.
Partial fix is better than no fix - you don't need to close the PR.
Even more, `test_max` is passing the test with your patch. The failing test is lint error(clippy) which is easy to fix. I commented about it.

youknowone (approved):

youknowone:
Thank you for contributing!

PR #2098, opened 2020-08-10
Fix whats_left.sh

youknowone:


tests/generator/not_impl_modules_footer.txt
```diff
@@ -2,39 +2,27 @@
...
-rustpymods = dict(map(
-    lambda mod: (
-        mod,
-        dir_of_mod_or_default(mod, not_imported)
-        if mod not in ("this", "antigravity") else None
-    ),
-    rustpymods
-))
+mod_names = [
+    name.decode() for name, ext in map(os.path.splitext, os.listdir(libdir))
```
youknowone, line 11: The bugfix is `name.decode()`. `os.listdir` returns list of bytes here but the code assumes it was str. Calling `decode()` will prevent future breaking by similar reason

PR #2101, opened 2020-08-10
Implement os.execve for unix system

ohgree:
# Implementation Details
Used `nix::unistd::execve` for execution.
Changes demonstrated at #2100 is applied for this execve as well.
I'm not quite sure whether the use of downcast() in the context is appropriate or not, but anyway it seems to work.
Any improvements or suggestions are welcome.

## Documentations
nix::unistd::execve 
https://docs.rs/nix/0.9.0/nix/unistd/fn.execve.html
Python3 os.execve 
https://docs.python.org/3/library/os.html#os.execve

## Test runs
```
cargo run -- -c 'import os;os.execve("/usr/bin/env", ["/usr/bin/env"], {"foo":"bar", "baz":"qux"})'
    Finished dev [unoptimized + debuginfo] target(s) in 0.40s
     Running `target/debug/rustpython -c 'import os;os.execve("/usr/bin/env", ["/usr/bin/env"], {"foo":"bar", "baz":"qux"})'`
foo=bar
baz=qux
```
Any null-embedded character throws ValueError

## Sidenote
Overlapping codes in os.execv and os.execve need to be refactored to improve modularity, but I believe it can wait until after this PR being merged.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1563,7 +1563,9 @@ mod posix {
             .ok_or_else(|| vm.new_value_error("execv() arg 2 must not be empty".to_owned()))?;
 
         if first.to_bytes().is_empty() {
-            return Err(vm.new_value_error("execv() arg2 first element cannot be empty".to_owned()));
+            return Err(
+                vm.new_value_error("execv() arg 2 first element cannot be empty".to_owned())
+            );
         }
 
         let env = env
@@ -1580,9 +1582,8 @@ mod posix {
 
         let env: Vec<&ffi::CStr> = env.iter().map(|entry| entry.as_c_str()).collect();
 
-        unistd::execve(&path, &args, &env)
-            .map(|_ok| ())
-            .map_err(|err| err.into_pyexception(vm))
+        unistd::execve(&path, &args, &env).map_err(|err| err.into_pyexception(vm))?;
+        Ok(())
     }
 
     #[pyfunction]
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1566,20 +1566,19 @@ mod posix {
             return Err(vm.new_value_error("execv() arg2 first element cannot be empty".to_owned()));
         }
 
-        let mut env_vec = vec![];
-        for (k, v) in env {
-            let key: PyStringRef = k.downcast().map_err(|_| {
-                vm.new_type_error("expected str, bytes or os.PathLike object".to_owned())
-            })?;
-            let value: PyStringRef = v.downcast().map_err(|_| {
-                vm.new_type_error("expected str, bytes or os.PathLike object".to_owned())
-            })?;
-            env_vec.push(
-                ffi::CString::new(format!("{}={}", key, value))
-                    .map_err(|_| vm.new_value_error("embedded null character".to_owned()))?,
-            );
-        }
-        let env: Vec<&ffi::CStr> = env_vec.iter().map(|entry| entry.as_c_str()).collect();
+        let env = env
+            .into_iter()
+            .map(|(k, v)| -> PyResult<_> {
+                let (key, value) = (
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/os.rs
```diff
@@ -1542,6 +1542,50 @@ mod posix {
...
+            .first()
+            .ok_or_else(|| vm.new_value_error("execv() arg 2 must not be empty".to_owned()))?;
+
+        if first.to_bytes().is_empty() {
+            return Err(vm.new_value_error("execv() arg2 first element cannot be empty".to_owned()));
+        }
+
+        let mut env_vec = vec![];
+        for (k, v) in env {
+            let key: PyStringRef = k.downcast().map_err(|_| {
```
youknowone, line 1571: `TryFromObject` is fitting here. Probably

```rust
let key = PyPathLike::try_from_object(k)?
```

ohgree: Thanks for the info. I'd gladly try that and make a commit

ohgree: fix is up. https://github.com/RustPython/RustPython/pull/2101/commits/b6621913921af0e7adfa762b5fa8a30cec9911df

vm/src/stdlib/os.rs
```diff
@@ -1542,6 +1542,49 @@ mod posix {
...
+            Either::B(tuple) => vm.extract_elements(tuple.as_object())?,
+        };
+        let args: Vec<&ffi::CStr> = args.iter().map(|entry| entry.as_c_str()).collect();
+
+        let first = args
+            .first()
+            .ok_or_else(|| vm.new_value_error("execv() arg 2 must not be empty".to_owned()))?;
+
+        if first.to_bytes().is_empty() {
+            return Err(vm.new_value_error("execv() arg2 first element cannot be empty".to_owned()));
```
youknowone, line 1566: formatting is not uniform
```suggestion
            return Err(vm.new_value_error("execv() arg 2 first element cannot be empty".to_owned()));
```

vm/src/stdlib/os.rs
```diff
@@ -1542,6 +1542,49 @@ mod posix {
...
+                ffi::CString::new(format!("{}={}", key.path.display(), value.path.display()))
+                    .map_err(|_| vm.new_value_error("embedded null character".to_owned()))
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let env: Vec<&ffi::CStr> = env.iter().map(|entry| entry.as_c_str()).collect();
+
+        unistd::execve(&path, &args, &env)
+            .map(|_ok| ())
+            .map_err(|err| err.into_pyexception(vm))
```
youknowone, line 1585:
```suggestion
        unistd::execve(&path, &args, &env)
            .map_err(|err| err.into_pyexception(vm))?
        Ok(())
```

ohgree: what difference does this make?

vm/src/stdlib/os.rs
ohgree: resolved https://github.com/RustPython/RustPython/pull/2101/commits/f1a7014183786035dd695d4248def2f433eda8ea

vm/src/stdlib/os.rs
ohgree: applied changes https://github.com/RustPython/RustPython/pull/2101/commits/f1a7014183786035dd695d4248def2f433eda8ea

youknowone: Just for human readability. Because the ok result is not interesting for human reader at that time. We can reduce one `map` by returning `Ok(())` at the bottom

ohgree: that is quite understandable, thanks

ohgree:
It seems to pass without any further effort. the commit is at https://github.com/RustPython/RustPython/pull/2101/commits/7ec074be53e04c803ae3ed2735bc267361ceb89b

PR #2102, opened 2020-08-10
Implement tp_name and derive support

youknowone:
I finished `pystruct_sequence` work but not about pyclass.

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -1,6 +1,7 @@
 use super::Diagnostic;
 use crate::util::{
-    attribute_arg, def_to_name, meta_into_nesteds, path_eq, ItemIdent, ItemMeta, ItemType,
+    attribute_arg, def_to_name, meta_into_nesteds, optional_attribute_arg, path_eq, ItemIdent,
+    ItemMeta, ItemType,
 };
 use proc_macro2::{Span, TokenStream as TokenStream2};
 use quote::{quote, quote_spanned, ToTokens};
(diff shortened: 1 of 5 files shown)


derive/src/pyclass.rs
```diff
@@ -420,6 +420,7 @@ pub fn impl_pyimpl(attr: AttributeArgs, item: Item) -> Result<TokenStream2, Diag
 fn generate_class_def(
     ident: &Ident,
     name: &str,
+    tp_name: &str,
```
coolreader18, line 423: Is there a similar thing to this in Python? A tp_name slot or something? If not, I'd recommend calling this `QUALNAME` or similar.

vm/src/pyobject.rs
```diff
@@ -1407,6 +1407,7 @@ where
 
 pub trait PyClassDef {
     const NAME: &'static str;
+    const TP_NAME: &'static str;
```
coolreader18, line 1410:
```suggestion
    const TP_NAME: &'static str = Self::NAME;
```

derive/src/pyclass.rs
youknowone: Yes, there is `tp_name`. Here is a test code about it:
```c
.tp_name = "custom.Custom",
```

vm/src/pyobject.rs
youknowone: I don't think this default value is good. `tp_name` in CPython includes module name.

coolreader18: Alright, this looks fine to me then.

PR #2103, opened 2020-08-11
Use a thread local variable for ReprGuard

coolreader18:


coolreader18:
Oh, true. Don't want to think about __repr__ implementations calling into sub-interpreters, but it isn't impossible

PR #2104, opened 2020-08-11
Make it a safety invariant that PyRef<T>.obj has a payload of T

coolreader18:
This make PyRef to be 0 cost, though I'm not sure how much that `expect()` in `PyRef::deref` was actually being optimized away.

### DIFF
--- a/vm/src/obj/objtuple.rs
+++ b/vm/src/obj/objtuple.rs
@@ -8,7 +8,7 @@ use crate::function::OptionalArg;
 use crate::pyobject::{
     self, BorrowValue, IdProtocol, IntoPyObject,
     PyArithmaticValue::{self, *},
-    PyClassImpl, PyComparisonValue, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
+    PyClassImpl, PyComparisonValue, PyContext, PyObjectRef, PyRef, PyResult, PyValue,
 };
 use crate::sequence::{self, SimpleSeq};
 use crate::vm::{ReprGuard, VirtualMachine};
@@ -238,11 +238,15 @@ impl PyTuple {
         vm: &VirtualMachine,
     ) -> PyResult<PyTupleRef> {
         let elements = if let OptionalArg::Present(iterable) = iterable {
-            if iterable.lease_class().is(&cls) {
-                return Ok(PyRef::from_obj_unchecked(iterable));
+            let exact_res = if cls.is(&vm.ctx.types.tuple_type) {
+                iterable.downcast_exact::<PyTuple>(vm)
+            } else {
+                Err(iterable)
+            };
+            match exact_res {
+                Ok(tup) => return Ok(tup),
+                Err(iterable) => vm.extract_elements(&iterable)?,
             }
-
-            vm.extract_elements(&iterable)?
         } else {
             vec![]
         };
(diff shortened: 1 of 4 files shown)


vm/src/pyobject.rs
```diff
@@ -753,7 +776,13 @@ where
     type Target = T;
 
     fn deref(&self) -> &T {
-        self.obj.payload().expect("unexpected payload for type")
+        self.obj.payload().unwrap_or_else(|| {
+            if cfg!(debug_assertions) {
+                panic!("PyRef.obj did not have a payload of T")
+            }
```
youknowone, line 782: I think `unreachable_unchecked()` still panic for opt 0. But to force check it in debug mode, this is also possible:
```
debug_assert!(false, "PyRef.obj did not have a payload of T");
```

vm/src/pyobject.rs
```diff
@@ -1308,7 +1339,7 @@ pub trait PyValue: fmt::Debug + PyThreadingConstraint + Sized + 'static {
     fn class(vm: &VirtualMachine) -> PyClassRef;
 
     fn into_simple_object(self, vm: &VirtualMachine) -> PyObjectRef {
```
youknowone, line 1341: will renaming to `into_exact_object` give more sense to fit to `downcast_exact`?

vm/src/obj/objtuple.rs
```diff
@@ -238,11 +238,15 @@ impl PyTuple {
...
-            if iterable.lease_class().is(&cls) {
-                return Ok(PyRef::from_obj_unchecked(iterable));
+            let exact_res = if cls.is(&vm.ctx.types.tuple_type) {
+                iterable.downcast_exact::<PyTuple>(vm)
+            } else {
+                Err(iterable)
+            };
+            match exact_res {
+                Ok(tup) => return Ok(tup),
+                Err(iterable) => vm.extract_elements(&iterable)?,
```
youknowone, line 248: This code seems confusing at the first glance maybe because the type of exact_res is not PyResult.
How about avoiding to bind non-PyResult to variable?
```suggestion
            let iterable = if cls.is(&vm.ctx.types.tuple_type) {
                match iterable.downcast_exact::<PyTuple>(vm) {
                    Ok(tuple) => return Ok(tuple),
                    Err(iterable) => iterable,
                }
            } else {
                iterable
            };
            vm.extract_elements(&iterable)?
```

vm/src/pyobject.rs
```diff
@@ -656,12 +656,31 @@ impl PyObject<dyn PyObjectPayload> {
...
+    /// Attempt to downcast this reference to the specific class that is associated `T`.
+    ///
+    /// If the downcast fails, the original ref is returned in as `Err` so
+    /// another downcast can be attempted without unnecessary cloning.
+    pub fn downcast_exact<T: PyObjectPayload + PyValue>(
+        self: PyRc<Self>,
+        vm: &VirtualMachine,
+    ) -> Result<PyRef<T>, PyObjectRef> {
+        if self.lease_class().is(&T::class(vm)) {
+            // TODO: is this always true?
```
youknowone, line 678: It will be always true once we implement CPython class assigning rule:
```
TypeError: __class__ assignment only supported for heap types or ModuleType subclasses
```

coolreader18: Alright, I guess I'll leave this here until we're sure, just as a safety guarantee.

vm/src/pyobject.rs
coolreader18: I think this is fine for now; I think if anything `into_object` should be renamed to `into_object_with_class` and `into_simple_object` should be renamed to `into_object`

PR #2106, opened 2020-08-12
unittest tests

coolreader18:


vm/src/obj/objtype.rs
```diff
@@ -308,16 +312,41 @@ impl PyClassRef {
...
+        }
+
+        let slots = PyClassSlots {
+            // TODO: how do we know if it should have a dict?
+            flags: base.slots.read().flags | PyTpFlags::HAS_DICT,
+            ..Default::default()
+        };
+
+        // TODO: is this correct behavior?
+        let cls_dict = if metatype.is(&vm.ctx.types.type_type) {
```
youknowone, line 333: is this related to metaclass support? I have no idea what's changed by this. It also seems this is decided by flags in pyvalue_to_ref

coolreader18: Yeah, this was to fix something in `enum` -- a method on EnumMeta was expecting that there was a __dict__ on the Enum type object. Againz not sure if this was actually correct or not but it fixed it lol

PR #2108, opened 2020-08-12
Add nested and optimized for symtable

Lynskylate:
Fix symtable is_nested test and optimized test.
Reference:
 [is_nested](https://github.com/python/cpython/blob/master/Python/symtable.c#L83)
[is_optimized](https://github.com/python/cpython/blob/3.8/Lib/symtable.py#L76)

### DIFF
--- a/vm/src/stdlib/symtable.rs
+++ b/vm/src/stdlib/symtable.rs
@@ -103,6 +103,16 @@ impl PySymbolTable {
         self.symtable.line_number
     }
 
+    #[pymethod(name = "is_nested")]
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/symtable.rs
```diff
@@ -103,6 +103,16 @@ impl PySymbolTable {
         self.symtable.line_number
     }
 
+    #[pymethod(name = "is_nested")]
```
youknowone, line 106: the name is omittable when it is same to the rust function name.

PR #2110, opened 2020-08-13
non-threading compatibility for imp.rs

youknowone:


vm/src/stdlib/imp.rs
```diff
@@ -5,46 +5,48 @@ use crate::obj::objmodule::PyModuleRef;
...
+#[cfg(feature = "threading")]
+mod lock {
+    use crate::pyobject::PyResult;
+    use crate::stdlib::thread::RawRMutex;
+    use crate::vm::VirtualMachine;
 
-fn imp_extension_suffixes(vm: &VirtualMachine) -> PyResult {
-    Ok(vm.ctx.new_list(vec![]))
-}
+    pub(super) static IMP_LOCK: RawRMutex = RawRMutex::INIT;
```
coolreader18, line 16: Hmm, this should probably be in the global state for the vm.

youknowone: agree

PR #2114, opened 2020-08-15
Allow complex numbers as arguments of complex()

edwardycl:
https://docs.python.org/3/library/functions.html#complex

Complex numbers can be arguments of complex()

### DIFF
--- a/vm/src/obj/objcomplex.rs
+++ b/vm/src/obj/objcomplex.rs
@@ -2,8 +2,7 @@ use num_complex::Complex64;
 use num_traits::Zero;
 use std::str::FromStr;
 
-use super::objfloat::{self, IntoPyFloat, PyFloat};
-use super::objint::{self, PyInt};
+use super::objfloat;
 use super::objstr::PyString;
 use super::objtype::PyClassRef;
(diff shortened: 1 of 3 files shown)


youknowone:
Suggestion:
```rust
let real = if let Some(comp) = try_complex(real) {
    comp
} else if let Some(s) = real.payload_if_subclass::<PyString>() {
   ...
} else {
   ...
};
// imag same as real
...

let result = Complex64::new(real.re - imag.im, real.im + imag.re);
...
```

vm/src/obj/objcomplex.rs
```diff
@@ -247,19 +247,15 @@ impl PyComplex {
...
-            OptionalArg::Missing => 0.0,
-            OptionalArg::Present(obj) => match_class!(match obj {
-                i @ PyInt => {
-                    objint::try_float(i.borrow_value(), vm)?
-                }
-                f @ PyFloat => {
-                    f.to_f64()
-                }
-                s @ PyString => {
+        let real: Complex64 = match real {
```
youknowone, line 253: redundant type annotation

youknowone:
Thanks for contributing!

PR #2116, opened 2020-08-15
Fix index and overflow error in itertools

clemado1:
1) fix index error (refer to [CPython](https://github.com/python/cpython/blob/fa5d7251987c70a9c5d58b59a0b36ac9287eaafa/Modules/itertoolsmodule.c#L3258))
```
test_permutations (__main__.TestBasicOps) ... thread 'main' panicked at 'index out of bounds: the len is 4 but the index is 4', vm/src/stdlib/itertools.rs:1420:42
```
2) fix overflow error in debug mode when r > n

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -1374,7 +1374,7 @@ mod decl {
             PyItertoolsPermutations {
                 pool,
                 indices: PyRwLock::new((0..n).collect()),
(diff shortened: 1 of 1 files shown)


coolreader18:
Is there a itertools test that we can remove the `unittest.skip()` decorator from now? Or did that panic only happen in debug mode?

youknowone:
As i've heard, it only occured in debug mode.

PR #2117, opened 2020-08-15
Allow `use` for `#[pyattr]`

youknowone:


derive/src/pymodule.rs
```diff
@@ -173,33 +173,22 @@ impl Module {
...
-                        }
-                        ItemType::Const => {
-                            items.push((
-                                Self::extract_const_attr(item.ident, into_nested()?)?,
-                                meta_span,
-                            ));
-                        }
-                        ItemType::Struct => {
+                    let item = match item.typ {
+                        ItemType::Fn => Self::extract_evaluated_attr(item.ident, into_nested()?)?,
```
coolreader18, line 177: You could probably just do `Const | Use => ...` here

derive/src/pymodule.rs
```diff
@@ -399,28 +388,46 @@ pub fn impl_pymodule(attr: AttributeArgs, item: Item) -> Result<TokenStream2, Di
...
+                    typ: ItemType::Const,
+                    attrs,
+                    ident,
+                },
+                Item::Use(syn::ItemUse { attrs, tree, .. }) => {
+                    let ident = match tree {
+                        UseTree::Path(path) => match &*path.tree {
+                            UseTree::Name(name) => &name.ident,
+                            UseTree::Rename(rename) => &rename.rename,
+                            _ => return None,
```
coolreader18, line 418: It'd probably be nice to support multiple imports `use foo::{BAR, BAZ, QUX};`, though that would probably require a bit of restructuring other code to do so

youknowone: I am thinking of overhaul of derive flow 😞

derive/src/pymodule.rs
youknowone: i will fix this one in the next PR

PR #2118, opened 2020-08-15
unexpected success on local

tobark:


youknowone:
@coolreader18 The weird thing here is, why it isn't catched by CI.

coolreader18:
Yeah, that is really weird. It wasn't failing on master before this?

youknowone:
yep, no failing. would there be any case of silience failure of unittest?

coolreader18:
Idk, maybe an unexpected success doesn't actually cause it to fail? Though I swear it has caused a ci failure before

PR #2119, opened 2020-08-16
check '=' character in environment variable name in os.execve to pass test_os.py

ohgree:
## About
after merging #2101, test_os.py just crashes without any message. By examining test_execve_invalid_env, it turned out that the initial implementation does not throw ValueError when an equal character `=` exists in the environment variable name. This PR fixes this issue.

## Result
```
> cargo run Lib/test/test_os.py
     Finished dev [unoptimized + debuginfo] target(s) in 21.27s
     Running `target/debug/rustpython Lib/test/test_os.py`
.sssssssssssssssss.s.s..xs..sxxx.xssxs..x.s...xsss.sxxssxxssssssssssssssssssss....x.Tested current directory length: 872
.x.xxxx..s.xxs...ss.Traceback (most recent call last):
OverflowError: Int value cannot fit into Rust u32
x.s......ssssssssssssssssssssssssssss...sssxss....ss.s.....x.......x..s.sssssssss.sssxsssssssssssssxsxs...sssssssssssssssssssssssssssssssssssssssssss
Ran 253 tests in 45.978s

OK (skipped=163, expected failures=27)
```

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1575,6 +1575,11 @@ mod posix {
                     PyPathLike::try_from_object(&vm, k)?,
                     PyPathLike::try_from_object(&vm, v)?,
                 );
+
+                if key.path.display().to_string().contains('=') {
+                    return Err(vm.new_value_error("illegal environment variable name".to_owned()));
+                }
+
                 ffi::CString::new(format!("{}={}", key.path.display(), value.path.display()))
                     .map_err(|_| vm.new_value_error("embedded null character".to_owned()))
             })


vm/src/stdlib/os.rs
```diff
@@ -1575,6 +1575,11 @@ mod posix {
                     PyPathLike::try_from_object(&vm, k)?,
                     PyPathLike::try_from_object(&vm, v)?,
                 );
+
+                if key.path.display().to_string().contains('=') {
```
youknowone, line 1579: is `display()` required here? I think any `Display` is `ToString`.
https://doc.rust-lang.org/std/string/trait.ToString.html#impl-ToString-3

ohgree: Isn't `display()` required to use `contains()`, which is implemented in `String`? AFAIK, Direct conversion from `PathBuf` to `String` creates more room for exceptions, due to how PathBuf and String handles string literals differently.

youknowone: Ah, you are right. I thought display is a trait method but it is a impl method https://doc.rust-lang.org/std/path/struct.Path.html#method.display

By checking a bit more, I now think Path -> OsStr -> CStr conversion is correct for this case. display() make it convert to `str` which implies utf-8, but Path and CStr actually doesn't need to be utf-8 guaranteed.
Because this is not a part of this PR but glitch from other changes before, It is ok for now as it is.

PR #2122, opened 2020-08-17
Allow keyword arguments in complex()

edwardycl:


vm/src/obj/objcomplex.rs
```diff
@@ -272,12 +266,12 @@ impl PyComplex {
...
         };
 
-        let imag = match imag {
-            OptionalArg::Missing => Complex64::new(0.0, 0.0),
-            OptionalArg::Present(obj) => {
+        let imag = match args.imag {
+            None => Complex64::new(0.0, 0.0),
+            Some(obj) => {
                 if let Some(c) = try_complex(&obj, vm)? {
                     c
```
coolreader18, line 273: You can say `&vm.ctx.types.str_type` instead of `&PyString::class()`

edwardycl: Fixed it!

PR #2123, opened 2020-08-17
Add FileIO.__del__

youknowone:


coolreader18:
Is there code in the stdlib that calls `f.__del__()`? Cause otherwise we don't have destructors yet; you'd have to implement FileIO as a pyclass and impl `Drop` for it

youknowone:
Thanks, I was trying to fix zipimport errors. I can't believe I didn't know that until now. I also found it is not even possible to implement with current structure: vm+rc

coolreader18:
Yeah, I've been thinking of ways to do it and it might be possible to still use Rc, if we have make PyObjectRef a wrapper struct around a ManuallyDrop PyRc, and in the Drop impl, if it has a __del__ method, then we push it to a queue of objects-to-be-destructed

youknowone:
I think i made a rough PoC of `__del__`. I will make a PR soon

PR #2124, opened 2020-08-17
Implement objcomplex::parse_str() to replace Complex64::from_str()

edwardycl:
`Complex64::from_str()` was used to parse a complex number from a string, but its implementation is quite different from how python's complex() should work.
For example, it cannot parse `"(1+2j)"`, `"J"`, or `"1_000"` - it throws `ValueError`.

The new function `objcomplex::parse_str()` handles parentheses correctly, allows capital `J`, and uses `float_ops::parse_str()` to parse float numbers.
I copied some tests from `test_complex.py` into snippets and made them work.

### DIFF
--- a/vm/src/obj/objcomplex.rs
+++ b/vm/src/obj/objcomplex.rs
@@ -1,4 +1,3 @@
-use itertools::Itertools;
 use num_complex::Complex64;
 use num_traits::Zero;
 
@@ -325,8 +324,8 @@ fn parse_str(s: &str) -> Option<Complex64> {
         Some(mut s) => {
             let mut real = 0.0;
             // Find the central +/- operator. If it exists, parse the real part.
-            for (i, (a, b)) in s.chars().tuple_windows().enumerate() {
-                if (b == '+' || b == '-') && !(a == 'e' || a == 'E') {
+            for (i, w) in s.as_bytes().windows(2).enumerate() {
+                if (w[1] == b'+' || w[1] == b'-') && !(w[0] == b'e' || w[0] == b'E') {
                     real = float_ops::parse_str(&s[..=i])?;
                     s = &s[i + 1..];
                 }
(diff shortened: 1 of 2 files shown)


vm/src/obj/objcomplex.rs
```diff
@@ -308,3 +309,39 @@ struct ComplexArgs {
...
+            Some(s) => s.trim(),
+        },
+    };
+
+    let value = match s.strip_suffix(|c| c == 'j' || c == 'J') {
+        None => Complex64::new(float_ops::parse_str(s)?, 0.0),
+        Some(mut s) => {
+            let mut real = 0.0;
+            // Find the central +/- operator. If it exists, parse the real part.
+            for (i, (a, b)) in s.chars().tuple_windows().enumerate() {
```
youknowone, line 328: Because every characters are ascii character, I think same algorithm will perfectly work as bytes. Will you try same version with `s.as_bytes()` isntead of `s.chars()`?

edwardycl: Fixed it! Thank you for the comment

vm/src/obj/objcomplex.rs
```diff
@@ -308,3 +308,39 @@ struct ComplexArgs {
...
+
+    let value = match s.strip_suffix(|c| c == 'j' || c == 'J') {
+        None => Complex64::new(float_ops::parse_str(s)?, 0.0),
+        Some(mut s) => {
+            let mut real = 0.0;
+            // Find the central +/- operator. If it exists, parse the real part.
+            for (i, w) in s.as_bytes().windows(2).enumerate() {
+                if (w[1] == b'+' || w[1] == b'-') && !(w[0] == b'e' || w[0] == b'E') {
+                    real = float_ops::parse_str(&s[..=i])?;
+                    s = &s[i + 1..];
```
youknowone, line 330: Maybe a `break` here?

PR #2127, opened 2020-08-17
Dict/set deletes existed element after returning borrow

youknowone:


vm/src/dictdatatype.rs
```diff
@@ -191,27 +197,40 @@ impl<T: Clone> Dict<T> {
...
     pub fn clear(&self) {
-        let mut inner = self.borrow_value_mut();
-        inner.entries.clear();
-        inner.indices.clear();
-        inner.size = 0
+        let mut swapped = Vec::new();
+        {
+            let mut inner = self.borrow_value_mut();
+            // defer dec rc
+            std::mem::swap(&mut inner.entries, &mut swapped);
```
coolreader18, line 204: I don't think you need the inner block, since it's at the end of the function anyway. Also, I think you could just do something like `let _old = std::mem::replace(entries, Vec::new());` to defer til the end of scope.

vm/src/obj/objlist.rs
```diff
@@ -521,13 +524,13 @@ impl PyList {
...
-            // TODO: Check if value was removed after lock released
-            self.borrow_value_mut().remove(index);
-            Ok(())
+            // defer delete out of borrow
+            Ok(self.borrow_value_mut().remove(index))
         } else {
             let needle_str = vm.to_str(&needle)?;
             Err(vm.new_value_error(format!("'{}' is not in list", needle_str.borrow_value())))
         }
+        .map(|_| ())
```
coolreader18, line 533:
```suggestion
        .map(drop)
```

vm/src/vm.rs
```diff
@@ -321,7 +321,9 @@ impl VirtualMachine {
         self.check_recursive_call("")?;
         self.frames.borrow_mut().push(frame.clone());
         let result = f(frame);
-        self.frames.borrow_mut().pop();
+        // defer dec frame
+        let popped = self.frames.borrow_mut().pop();
+        drop(popped);
```
coolreader18, line 326: I don't think this really defers anything

youknowone: It was popped while I am working on `__del__`, so it also worked. Will `let _ = self.frames.borrow_mut().pop()` work same?

coolreader18: No, it's really unintuitive but if it's `_` it doesn't store as a variable at all, but`_foo` holds onto it till the end of scope

youknowone: Thanks ,I will fix things again

Issue #1940, opened 2020-05-23
[RFC] What are the compatibility goals for RustPython?


jamestwebber:
## Summary

As a long-time Pythonista and short-time Rustacean I of course searched for this project and was happy to find it existed. Looks like a lot of fun, and if I get the itch I'll definitely try to contribute (with the caveat that like most of you, I won't be able to dedicate a ton of time).

I was wondering about the long-term vision for RustPython in terms of CPython interop (if not necessarily parity). I think this question is the core of issues like #1849, #1815, #1690, #1204 and #1192, and having read those discussions it seems like it's an ongoing discussion.

## Is RustPython a replacement for CPython, or a new interpreter to enable Python in new places (e.g. the web) but doesn't try to be compatible?

I think this is the big question, and answering it would define a roadmap for the future. The two answers define two different projects that are both exciting for different ways:

| Goal | Pros | Cons  |
|------|-----|-------|
| CPython replacement | Greenfield implementation of core code | Faithfully recreating stdlib is complicated |
| | Rust is simpler, less buggy than C | Tons of complex dynamic behavior to support |
| | Compatible with a _huge_ ecosystem | Limited ability to change/optimize any syntax or code behavior | 
| | Potential for major impact/user uptake | Potential for huge support surface |
| Novel interpreter / Wasm compiler | Can selectively implement libraries/syntax | More difficult for new users, less "pythonic" |
| | Can optimize the interpreter (like PyPy) | Libraries that aren't pure Python require active support or won't work |
| | Maybe more exciting to work on, for some? | Much less potential for broad impact IMO |

To me, RustPython is a cool idea in large part because Rust has good compatibility with C, and so RustPython could (maybe) work with the huge number of libraries that mix C with Python†. Most of the heavy hitters in the CPython space fall into this category–the whole data stack, for instance. I think e.g. PyPy suffers a lot from not being trivially compatible with such popular projects, even though it has lots of other advantages. Going this route means supporting all the complex meta-programming, introspection, etc that those projects rely on. That might be fun, if the goal is to learn about interpreters!

This approach doesn't mean _not_ compiling to Wasm or anything like that. But if the goal is _primarily_ web-native Python, I think that changes priorities and the direction of the project.

† not to mention Cython, C++, Fortran, etc: if it works with CPython then it could work with a RustPython.

[...]

PR #2128, opened 2020-08-18
CPython compatible Drop for PyObjectRc

youknowone:


youknowone:
Ok, I got what you meant.

About `__del__` timing, it is easy to avoid by one of below:
- Easy way: increase count, run `__del__`, decrease count
- Harder way: Create arc with count 2 and drop when it goes to count 2. decrease 1 more and drop arc

The thread_local problem:
I agree current way is not ideal. I didn't look up exact point to safely set the shared value. I think we at least need a thread local variable regardless it is a vm, an executing frame or a vec (for sweeping). I set it on frame because It was the safetest place for every choice, but it can be improved in future if the value is vm. things will be easier when a thread is always bound by a vm.

I don't know well about vm, gc and their performance that much.
But I am thinking about PyPy's pros and cons by its design.

I believe there is an opportunity of better performance for collecting and release.

I am thinking of making RustPython as a drop-in-replacement of CPython.
For this purpose, it requires a sort of 'hard compatibility' of CPython.
I don't think it is specific future of RustPython, but at lease we can consider a mode like that. Hard-compatibilty will give potential opportunity between CPython and PyPy.

One of the common problem is resource releasing timing. Some (wrong) python code is depending of CPython's drop timing, which only can simulated by calling `__del__` at exact timing to go to rc=0. So, I want to have this one as a possible memory management model, unless we can prove RustPython never can implements c/c++ extension interface.

vm/src/pyobjectrc.rs
```diff
@@ -0,0 +1,313 @@
...
+    PyRc<PyObject<T>>: AsPyObjectRef,
+{
+    pub fn into_raw(this: Self) -> *const PyObject<T> {
+        let ptr = PyRc::as_ptr(&this.inner);
+        std::mem::forget(this);
+        ptr
+    }
+
+    unsafe fn into_rc(this: Self) -> PyRc<PyObject<T>> {
+        std::mem::transmute(this)
```
coolreader18, line 50: Any reason you can't just do `this.inner` ?

youknowone: Then it will try to move inner as return value, which means `this` will have undefined inner field.

coolreader18: Huh, TIL [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=311b49495470d90fb0a052649c5a600c) is an error. Maybe something like the into_raw method, but you `ptr::read(&this.inner)` + `mem::forget(this)`?

youknowone: Ok, it seems `PyRc::from_raw(Self::into_raw(this))` works

vm/src/vm.rs
```diff
@@ -68,6 +68,13 @@ pub struct VirtualMachine {
     pub initialized: bool,
 }
 
+#[cfg(not(target_arch = "wasm32"))]
+thread_local! {
+    pub static THREAD_VM: Cell<*const VirtualMachine> = Cell::new(std::ptr::null());
+}
+#[cfg(target_arch = "wasm32")]
+pub static mut UNIQUE_VM: *const VirtualMachine = std::ptr::null();
```
youknowone, line 76: Unfortunately, `THREAD_VM.with()` always cause timeout in wasm

vm/src/vm.rs
```diff
@@ -68,6 +68,29 @@ pub struct VirtualMachine {
...
+    }
+
+    pub(crate) fn with_vm<F, R>(f: F) -> R
+    where
+        F: Fn(&VirtualMachine) -> R,
+    {
+        VM.with(|tvm| {
+            // vm is guaranteed to be existing unless it is not null
+            let pvm = tvm.get();
+            debug_assert!(!pvm.is_null());
```
coolreader18, line 87: I had an idea for how this could work, since as it is doesn't go particularly well with sub-interpreters: we make the entry point to RustPython something like an Interpreter struct, and you have to call `interpreter.enter(|vm| { })` to get a reference to the VM, and that's what sets the thread-local variable, calls f(), and then unsets it. Then, we could have a stack of thread-local vm pointers, and if you want to enter a sub-interpreter you have to call Interpreter::enter on that one, and if there's multiple vms on the stack then we find the one where `isinstance(obj, &vm.ctx.types.object_type)`

coolreader18: POC: https://github.com/RustPython/RustPython/tree/coolreader18/interpreter-enter

youknowone: Do you want to push your commits to this PR? As I know org admins can push to forked branches when it is opened as PR.

coolreader18: Sure, I'll merge it in.

coolreader18:
Ah, it's an issue with stuff being dropped while in `vm.initialize()`, I'll fix that

coolreader18 (approved):

coolreader18 (approved):

coolreader18:
@skinny121 what do you think?

vm/src/import.rs
```diff
@@ -14,45 +14,57 @@ use crate::vm::{InitParameter, VirtualMachine};
 #[cfg(feature = "rustpython-compiler")]
 use rustpython_compiler::compile;
 
-pub fn init_importlib(vm: &mut VirtualMachine, initialize_parameter: InitParameter) -> PyResult {
+pub fn init_importlib(
```
youknowone, line 17: this is receiving `vm` as an argument but requries to call `enter_vm`. this is only called in initialization step, so it will not make any problem. but this is making a bit uncomfortable feeling.

vm/src/vm.rs
```diff
@@ -1537,29 +1577,76 @@ impl<'vm> Drop for ReprGuard<'vm> {
     }
 }
 
+pub struct Interpreter {
```
youknowone, line 1580: does Interpreter matche to vm 1-to-1 relationship or is there any other usage of vm?
If it is 1-to-1, I don't feel like we really need a new type for that; we only need the new methods.

coolreader18: The separate struct is so that you can't access the VM except through putting it in the VM stack, otherwise, it wouldn't be guaranteed that the VM would be in stack while using it

vm/src/import.rs
coolreader18: Should I make init_importlib `pub(crate)`?

coolreader18: And yeah, that's definitely tricky

vm/src/vm.rs
youknowone: In that case, will there be any difference if we put interpreters in VM stack instead of VMs?

vm/src/import.rs
youknowone: i didn't think about `pub(crate)` at that time, but it looks like to be. Even if there is a need to manally init importlib, it must be a method of vm or interpreter rather than this kind of function. I think we'd better to list up pub apis. Sometimes internal stuffs are marked as pub just to be shared in the crate.

vm/src/vm.rs
coolreader18: I initially did that, but then it was tricky passing an Interpreter to `initialize(&mut interp.vm, &interp)`. Definitely could make it happen though, especially if we're making init_importlib crate-private

youknowone: ah, got it. interface visibility is a good reason.

skinny121:
Looks pretty good. `PyObjectRc`'s looks like they can still escape `crate::vm::thread::enter_vm` though, which means that this doesn't guarantee that a vm is always available in the drop. The way to ensure this is to add a lifetime parameter to `PyObjectRc`, though I'm unsure how feasible that would actually be though.

vm/src/vm.rs
skinny121: This actually feels like a 1-to-n relationship as each thread has it's own vm.

skinny121 (approved):

vm/src/vm.rs
youknowone: it seems each thread has its own vm and also its own interpreter. so still vm and interpreter is 1-to-1.

coolreader18: I was originally thinking that Interpreter would cover all vms across threads for that interpreter instance; like it could have a thread-local value to hold the vm, and it could be Send+Sync and whichever thread you called it from it would get that thread's vm, but I'm not sure how actually useful that would be.

coolreader18:
I added another commit to rework vm initialization a bit; not sure if it's totally on-topic but I think it makes more sense with how the `Interpreter` struct works now, and I think it's good to have the `InitParameter` be more explicit, since either way having an expectation for how it should work by default isn't great. If you expect it by default not to have access to external imports but it does, that could be a security risk, and if you expect it by default to have access to external imports but it doesn't, that might be confusing why something like `import os` fails.

vm/src/vm.rs
```diff
@@ -309,7 +341,8 @@ impl VirtualMachine {
             repr_guards: RefCell::default(),
             state: self.state.clone(),
             initialized: self.initialized,
-        }
+        };
+        Interpreter { vm }
```
skinny121, line 345: I think we can just return the vm itself here as this is only visible to this crate and the use of it function simply calls `enter`, thus it can be replaced by `crate::vm::thread::enter_vm`

coolreader18: I had it return vm just because the `Interpreter.vm` field was private, and it's the same functionality anyway. I suppose it would keep the distinction between interpreter/vm more concrete though, I'll change it

coolreader18:
Should we merge this?

youknowone:
If the design is good enough, let's go and fix remained problems later. this pr is growing bigger. bigger patch will cause more frequent conflicts for other PRs

youknowone:
I think the comments here is realted to #1940

PR #2130, opened 2020-08-19
io.FileIO into #[pyclass]

youknowone:


vm/src/stdlib/io.rs
```diff
@@ -643,6 +647,21 @@ mod fileio {
...
+        closefd: AtomicCell<bool>,
+    }
+
+    type FileIORef = PyRef<FileIO>;
+
+    impl PyValue for FileIO {
+        fn class(vm: &VirtualMachine) -> PyClassRef {
+            vm.class("_io", "FileIO")
+        }
+    }
```
coolreader18, line 663: `impl Drop for FileIO { self.close() }`  ?

youknowone: I am thinking of `ty_del` to support manual calling `__del__`

PR #2131, opened 2020-08-19
Move out read-only flags of slots.

youknowone:


coolreader18:
If you wanted to make everything immutable you could probably have an associated function for PyValue like `fn make_slots() -> PyClassSlots;`

coolreader18:
Or actually, maybe we could like split up extend_class and extend_slots:

```rust
trait PyClassImpl {
    fn impl_extend_class(ctx: &PyContext, class: &PyClassRef);
    fn extend_slots(slots: &mut PyClassSlots);
    fn make_slots() -> PyClassSlots {
        let mut slots = PyClassSlots::default();
        Self::extend_slots(&mut slots);
        slots
    }
}
```

And that way it's completely separated, and we can avoid having to ever mutate slots after the class is created.

youknowone:
extend_slot looks like a nice idea

youknowone:
@coolreader18 will you review this PR as is? I want to finish derive work first

PR #2133, opened 2020-08-19
Fix panic in itertools pruduct

clemado1:
I fixed panic caused by attempt to subtract with overflow.
But `test_permutations` and `test_product_pickling` still have error at `self.pickletest`.

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -1067,7 +1067,7 @@ mod decl {
             PyItertoolsProduct {
                 pools,
                 idxs: PyRwLock::new(vec![0; l]),
-                cur: AtomicCell::new(l - 1),
+                cur: AtomicCell::new(l.wrapping_sub(1)),
                 stop: AtomicCell::new(false),
             }
             .into_ref_with_type(vm, cls)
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -1067,7 +1067,7 @@ mod decl {
             PyItertoolsProduct {
                 pools,
                 idxs: PyRwLock::new(vec![0; l]),
-                cur: AtomicCell::new(l - 1),
+                cur: AtomicCell::new(l.saturating_sub(1)),
```
youknowone, line 1070: Does `wrapping_sub` make any problem here? otherwise i prefer to use wrapping_sub here.
Because we are using `fetch_sub` below, which implies `wrapping_sub`, we just can go the same way.

coolreader18: Isn't a wrapping subtraction what was causing the issue in the first place?

youknowone: It does, but this PR also added length check which prevent to hit self.cur when it is overflowed.

clemado1:
I changed `saturating_sub` to `wrapping_sub`.
Thank you for your advice :)

coolreader18:
Nice work :)

PR #2137, opened 2020-08-19
Trying revert for regression

youknowone:

PR #2139, opened 2020-08-20
Revert "Trying revert for regression"

coolreader18:
Reverts RustPython/RustPython#2137

Revert the reversion; I think I figured out what I did wrong and this should basically be a fixed version of #2106

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -1346,28 +1346,28 @@ pub trait PyValue: fmt::Debug + PyThreadingConstraint + Sized + 'static {
     }
 
     fn into_ref(self, vm: &VirtualMachine) -> PyRef<Self> {
-        pyvalue_into_ref(self, Self::class(vm), vm)
+        self.into_ref_with_type_unchecked(Self::class(vm), vm)
     }
 
     fn into_ref_with_type(self, vm: &VirtualMachine, cls: PyClassRef) -> PyResult<PyRef<Self>> {
         let class = Self::class(vm);
         if objtype::issubclass(&cls, &class) {
-            Ok(pyvalue_into_ref(self, cls, vm))
+            Ok(self.into_ref_with_type_unchecked(cls, vm))
         } else {
             let subtype = vm.to_str(&cls.obj)?;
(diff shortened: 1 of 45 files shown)


vm/src/pyobject.rs
```diff
@@ -1368,6 +1361,15 @@ pub trait PyValue: fmt::Debug + PyThreadingConstraint + Sized + 'static {
     }
 }
 
+fn pyvalue_into_ref<T: PyValue>(value: T, cls: PyClassRef, vm: &VirtualMachine) -> PyRef<T> {
```
youknowone, line 1364: can this function be a method of `PyValue`?

vm/src/slots.rs
```diff
@@ -4,14 +4,15 @@ use crate::VirtualMachine;
...
+        const HEAPTYPE = 1 << 9;
         const BASETYPE = 1 << 10;
         const HAS_DICT = 1 << 40;
     }
 }
 
 impl PyTpFlags {
     // CPython default: Py_TPFLAGS_HAVE_STACKLESS_EXTENSION | Py_TPFLAGS_HAVE_VERSION_TAG
-    pub const DEFAULT: Self = Self::from_bits_truncate(0);
+    pub const DEFAULT: Self = Self::HEAPTYPE;
```
youknowone, line 15: This is not a default flag of CPython by the upper comment. Can this value be injected when we are actually creating a heap type? Similar as `HAS_DICT` flag?

youknowone:
You found it quickly, great 👍

vm/src/slots.rs
coolreader18: The thing is, we don't actually have non-heap types yet; all of our type objects are allocated on the heap.

vm/src/pyobject.rs
coolreader18: Maybe, but I didn't put it because I just wanted to dedup some code; it wouldn't make sense to override it in an `impl PyValue`

coolreader18: Though maybe it would, idk. I'll just do it.

youknowone (approved):

PR #2142, opened 2020-08-22
Proof of concept JIT compiler

skinny121:
Proof of concept for a JIT(just in time) compiler, that compiles python functions down to native code.

This uses https://github.com/bytecodealliance/wasmtime/tree/main/cranelift to do the heavy lifting of generating the native code.

Currently this can successfully compile the following:
```python
def foo():
    a = 9_223
    return 10 + a
```

As a proof of concept, this has very limited functionality:
- only handles functions that take no arguments and return an integer
- only handles a very limited set of bytecodes
- aborts if operations overflow

### DIFF
--- a/jit/src/lib.rs
+++ b/jit/src/lib.rs
@@ -44,13 +44,13 @@ impl Error for JITCompileError {
     }
 }
 
-struct JIT {
+struct Jit {
     builder_context: FunctionBuilderContext,
     ctx: codegen::Context,
     module: Module<SimpleJITBackend>,
 }
 
-impl JIT {
+impl Jit {
     fn new() -> Self {
         let builder = SimpleJITBuilder::new(cranelift_module::default_libcall_names());
         let module = Module::new(builder);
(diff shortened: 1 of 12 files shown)


jit/src/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+impl Error for JITCompileError {
+    fn source(&self) -> Option<&(dyn Error + 'static)> {
+        match *self {
+            JITCompileError::CraneliftError(ref err) => Some(err),
+            _ => None,
+        }
+    }
+}
+
+struct JIT {
```
coolreader18, line 47: Rust type names are camel case, including for acronyms -- if you want, you could `s/JIT/Jit` for the type names

tests/test_snippets.py
```diff
@@ -136,7 +138,7 @@ def setUpClass(cls):
 
         # cargo stuff
         profile_args = [] if RUST_DEBUG else ["--release"]
-        subprocess.check_call(["cargo", "build", *profile_args])
+        subprocess.check_call(["cargo", "build", "--features", *RUSTPYTHON_FEATURES, *profile_args])
```
coolreader18, line 141:
```suggestion
        subprocess.check_call(["cargo", "build", "--features", ",".join(RUSTPYTHON_FEATURES), *profile_args])
```

coolreader18:
LGTM, feel free to merge

PR #2146, opened 2020-08-25
Add ftplib + test_ftplib

coolreader18:
I _tried_ to get ssl working, but it's just such a pain. I got it to an sort of useful error message after updating socket.py, but I was sorta burnt out by then, especially after I `git reset --hard` half my work accidentally. I just want this out of my face at this point :sweat_smile:


coolreader18:
Actually, I think I'll keep this a draft til alexcrichton/socket2-rs#97 is resolved/a new version is pushed

coolreader18:
Okay, this is the same socket error as #2179, so at least all the issues on the ftplib end are fixed

youknowone (approved):

PR #2147, opened 2020-08-25
Handle TypeErrors in bytes() & bytearray()

edwardycl:
Few changes were made to pass `test_constructor_type_errors`:

- Use the correct arguments: `[source[, encoding[, errors]]]`
- If either `encoding` or `errors` is present when `source` is NOT a string, raise `TypeError`
- If `encoding` is NOT present when `source` is a string, raise `TypeError`
- Do not allow iterable of non-ints
  - `bytes(["0"])` should raise TypeError
  - However, `objint::to_int` was used originally, allowing iterable of strings or floats as an argument
  - `vm.extract_elements::<PyIntRef>(&obj)` uses `TryFromObject` internally, so only `int` objects are allowed now

### DIFF
--- a/vm/src/bytesinner.rs
+++ b/vm/src/bytesinner.rs
@@ -140,16 +140,16 @@ impl ByteInnerNewOptions {
                                     //     ))
                                     // })?;
 
-                                    let mut data_bytes = vec![];
-                                    for elem in elements {
-                                        let i = elem.borrow_value().to_u8().ok_or_else(|| {
-                                            vm.new_value_error(
-                                                "bytes must be in range(0, 256)".to_owned(),
-                                            )
-                                        })?;
-                                        data_bytes.push(i);
-                                    }
-                                    Ok(data_bytes)
+                                    elements
+                                        .into_iter()
+                                        .map(|elem| {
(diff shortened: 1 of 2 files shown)


vm/src/bytesinner.rs
```diff
@@ -62,85 +62,101 @@ impl TryFromObject for PyBytesInner {
...
+
+                                    let mut data_bytes = vec![];
+                                    for elem in elements {
+                                        let i = elem.borrow_value().to_u8().ok_or_else(|| {
+                                            vm.new_value_error(
+                                                "bytes must be in range(0, 256)".to_owned(),
+                                            )
+                                        })?;
+                                        data_bytes.push(i);
+                                    }
```
youknowone, line 151: It is up to preference, but possible to be represented with `collect()`
```rust
                                    let data_bytes: Result<_, _> = elements.into_iter().map(|elem| {
                                        elem.borrow_value().to_u8().ok_or_else(|| {
                                            vm.new_value_error(
                                                "bytes must be in range(0, 256)".to_owned(),
                                            )
                                        })
                                    }).collect();
```

youknowone (approved):
nice work!

PR #2148, opened 2020-08-25
Windows CI

youknowone:


skinny121 (approved):

youknowone:
ok, it now works. there will be a lot problems to solve, but it is a good start point. The last pieces of windows unittest work was done by @BasixKOR, thanks!

PR #2150, opened 2020-08-25
Allow floats in jitted functions

coolreader18:


tests/snippets/jit.py
```diff
@@ -3,10 +3,18 @@ def foo():
...
+    assert foo() == 15
+    assert bar() == 2e5
+
+tests()
 
 if hasattr(foo, "__jit__"):
     print("Has jit")
     foo.__jit__()
-    assert foo() == 15
+    tests()
```
skinny121, line 20: You will need to call `bar.__jit__()` to actually compile `bar`.

skinny121 (approved):
LGTM

PR #2151, opened 2020-08-26
Unskip test_max

coolreader18:


youknowone:
Referring to @hyperbora 's comment in #2097, the test occasionally fails by sorting result

coolreader18:
Hmmmmm. Related to #1072, maybe?

PR #2154, opened 2020-08-27
Test lshift of zero

tobark:
It seems kinda rough code though, it passes the tests.
I'd like to get advice to improve this code.

### DIFF
--- a/vm/src/obj/objint.rs
+++ b/vm/src/obj/objint.rs
@@ -192,26 +192,26 @@ fn inner_divmod(int1: &BigInt, int2: &BigInt, vm: &VirtualMachine) -> PyResult {
 }
 
 fn inner_lshift(int1: &BigInt, int2: &BigInt, vm: &VirtualMachine) -> PyResult {
-    if int2.is_negative() {
-        return Err(vm.new_value_error("negative shift count".to_owned()));
-    } else if int1.is_zero() {
-        return Ok(vm.ctx.new_int(0));
-    } else if *int2 > BigInt::from(usize::max_value()) {
-        return Err(vm.new_overflow_error("the number is too large to convert to int".to_owned()))
-    } else {
-        return Ok(vm.ctx.new_int(int1 << int2.to_usize().expect("Failed converting {} to rust usize")));
-    }
+    inner_shift(int1, int2, |a, b| a << b, vm)
 }
 
 fn inner_rshift(int1: &BigInt, int2: &BigInt, vm: &VirtualMachine) -> PyResult {
+    inner_shift(int1, int2, |a, b| a >> b, vm)
+}
+
+fn inner_shift<F>(int1: &BigInt, int2: &BigInt, shift_op: F, vm: &VirtualMachine) -> PyResult
+where
+    F: Fn(&BigInt, usize) -> BigInt,
+{
     if int2.is_negative() {
-        return Err(vm.new_value_error("negative shift count".to_owned()));
+        Err(vm.new_value_error("negative shift count".to_owned()))
     } else if int1.is_zero() {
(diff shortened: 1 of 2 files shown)


vm/src/obj/objint.rs
```diff
@@ -192,13 +192,27 @@ fn inner_divmod(int1: &BigInt, int2: &BigInt, vm: &VirtualMachine) -> PyResult {
...
+    if int2.is_negative() {
+        return Err(vm.new_value_error("negative shift count".to_owned()));
+    } else if int1.is_zero() {
+        return Ok(vm.ctx.new_int(0));
+    } else if *int2 > BigInt::from(usize::max_value()) {
+        return Err(vm.new_overflow_error("the number is too large to convert to int".to_owned()))
+    } else {
+        return Ok(vm.ctx.new_int(int1 >> int2.to_usize().expect("Failed converting {} to rust usize")));
+    }
 }
```
youknowone, line 216: It seems only shift operation is different and the other parts are the same. If we introduce a closure, they can be combined into a single function

```suggestion
fn inner_shift<F>(int1: &BigInt, int2: &BigInt, shift_op: F, vm: &VirtualMachine) -> PyResult
where
    F: Fn(&BitInt, usize) -> BigInt
{
    if int2.is_negative() {
        Err(vm.new_value_error("negative shift count".to_owned()));
    } else if int1.is_zero() {
        Ok(vm.ctx.new_int(0));
    } else {
        let int2 = int2.to_usize().ok_or_else(|| {
            vm.new_overflow_error("the number is too large to convert to int".to_owned()))
        })?;
        Ok(vm.ctx.new_int(shift_op(int1, int2).expect("Failed converting {} to rust usize")));
    }
}
```

youknowone (changes requested):

vm/src/obj/objint.rs
```diff
@@ -192,13 +192,27 @@ fn inner_divmod(int1: &BigInt, int2: &BigInt, vm: &VirtualMachine) -> PyResult {
...
-    let n_bits = get_shift_amount(int2, vm)?;
-    Ok(vm.ctx.new_int(int1 << n_bits))
+    inner_shift(int1, int2, |a, b| a << b, vm)
 }
 
 fn inner_rshift(int1: &BigInt, int2: &BigInt, vm: &VirtualMachine) -> PyResult {
-    let n_bits = get_shift_amount(int2, vm)?;
-    Ok(vm.ctx.new_int(int1 >> n_bits))
+    inner_shift(int1, int2, |a, b| a >> b, vm)
+}
```
youknowone, line 200: we don't need one line `inner_lshift` and `inner_rshift` anymore. caling `inner_shift` in each shift funciton will be enough.

vm/src/obj/objint.rs
```diff
@@ -407,22 +399,22 @@ impl PyInt {
...
     #[pymethod(name = "__lshift__")]
     fn lshift(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        self.general_op(other, |a, b| inner_lshift(a, b, vm), vm)
+        self.general_op(other, |a, b| inner_shift(a, b, |a, b| a << b, vm), vm)
     }
 
     #[pymethod(name = "__rlshift__")]
     fn rlshift(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        self.general_op(other, |a, b| inner_lshift(b, a, vm), vm)
+        self.general_op(other, |a, b| inner_shift(a, b, |a, b| a << b, vm), vm)
```
youknowone, line 407: the parameters are swapped

youknowone (approved):

PR #2155, opened 2020-08-27
PyObjectRc: thin wrapper for PyObjectRef inner Rc

youknowone:
Split `__del__` unrelated part from #2128 

this patch will be base of any kind of drop implementation

### DIFF
--- a/vm/src/obj/objweakref.rs
+++ b/vm/src/obj/objweakref.rs
@@ -26,7 +26,7 @@ impl PyWeak {
     }
 
     pub fn upgrade(&self) -> Option<PyObjectRef> {
-        PyObjectRc::upgrade_weak(&self.referent)
+        self.referent.upgrade()
     }
 }
 
--- a/vm/src/obj/objweakref.rs
+++ b/vm/src/obj/objweakref.rs
@@ -1,33 +1,32 @@
 use super::objtype::PyClassRef;
+use crate::common::hash::PyHash;
 use crate::function::{OptionalArg, PyFuncArgs};
 use crate::pyobject::{
-    IdProtocol, PyClassImpl, PyContext, PyObject, PyObjectPayload, PyObjectRef, PyRef, PyResult,
-    PyValue, TypeProtocol,
+    IdProtocol, PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
 };
+use crate::pyobjectrc::{PyObjectRc, PyObjectWeak};
 use crate::slots::{Hashable, SlotCall};
 use crate::vm::VirtualMachine;
-use rustpython_common::hash::PyHash;
-use rustpython_common::rc::{PyRc, PyWeak as Weak};
 
 use crossbeam_utils::atomic::AtomicCell;
 
 #[pyclass(module = false, name = "weakref")]
 #[derive(Debug)]
 pub struct PyWeak {
-    referent: Weak<PyObject<dyn PyObjectPayload>>,
+    referent: PyObjectWeak,
     hash: AtomicCell<Option<PyHash>>,
 }
 
 impl PyWeak {
     pub fn downgrade(obj: &PyObjectRef) -> PyWeak {
         PyWeak {
-            referent: PyRc::downgrade(obj),
+            referent: PyObjectRc::downgrade(obj),
             hash: AtomicCell::new(None),
         }
     }
 
     pub fn upgrade(&self) -> Option<PyObjectRef> {
-        self.referent.upgrade()
+        PyObjectRc::upgrade_weak(&self.referent)
     }
 }
 
@@ -42,7 +41,7 @@ pub type PyWeakRef = PyRef<PyWeak>;
 impl SlotCall for PyWeak {
     fn call(&self, args: PyFuncArgs, vm: &VirtualMachine) -> PyResult {
         args.bind::<()>(vm)?;
-        Ok(self.referent.upgrade().unwrap_or_else(|| vm.get_none()))
+        Ok(self.upgrade().unwrap_or_else(|| vm.get_none()))
     }
 }
 
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -62,7 +62,7 @@ Basically reference counting, but then done by rust.
 /// this reference counting is accounted for by this type. Use the `.clone()`
 /// method to create a new reference and increment the amount of references
 /// to the python object by 1.
-pub type PyObjectRef = PyObjectRc;
+pub type PyObjectRef = PyObjectRc<dyn PyObjectPayload>;
 
 /// Use this type for functions which return a python object or an exception.
 /// Both the python object and the python exception are `PyObjectRef` types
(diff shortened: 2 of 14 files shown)


vm/src/obj/objweakref.rs
```diff
@@ -21,13 +21,13 @@ pub struct PyWeak {
 impl PyWeak {
     pub fn downgrade(obj: &PyObjectRef) -> PyWeak {
         PyWeak {
-            referent: PyRc::downgrade(obj),
+            referent: PyRc::downgrade(&obj.rc()),
```
skinny121, line 24: Maybe add `downgrade` method to `PyObjectRc` instead.

vm/src/pyobject.rs
```diff
@@ -509,11 +512,28 @@ impl PyObject<dyn PyObjectPayload> {
             let ret = unsafe { PyRc::from_raw(ptr) };
             Ok(ret)
         } else {
-            Err(self)
+            Err(self.into())
         }
     }
 }
 
+impl PyObjectRc {
```
skinny121, line 520: Moving this into pyobjectrc.rs, so we can access the inner value directly might be better.

vm/src/pyobject.rs
```diff
@@ -693,6 +713,12 @@ impl<T: ?Sized + PyObjectPayload> IdProtocol for PyObject<T> {
     }
 }
 
+impl IdProtocol for PyObjectRc {
```
skinny121, line 716: Same here if this was in pyobjectrc.rs then it could simply reference the inner value.

vm/src/stdlib/weakref.rs
```diff
@@ -5,12 +5,12 @@
...
 //!
 
+use crate::common::rc::PyRc;
 use crate::pyobject::PyObjectRef;
 use crate::vm::VirtualMachine;
-use rustpython_common::rc::PyRc;
 
 fn weakref_getweakrefcount(obj: PyObjectRef) -> usize {
-    PyRc::weak_count(&obj)
+    PyRc::weak_count(obj.rc())
```
skinny121, line 13: I think adding `weak_count` method on `PyObjectRc` would be good.

vm/src/sysmodule.rs
```diff
@@ -122,7 +122,7 @@ impl SysFlags {
 }
 
 fn sys_getrefcount(obj: PyObjectRef) -> usize {
-    PyRc::strong_count(&obj)
+    PyRc::strong_count(obj.rc())
```
skinny121, line 125: Same here from `strong_count`.

vm/src/pyobject.rs
```diff
@@ -472,9 +475,9 @@ impl PyObject<dyn PyObjectPayload> {
         self: PyRc<Self>,
```
skinny121, line 475: Can this be changed to take `self: PyObjectRc<Self>` instead?

vm/src/pyobject.rs
```diff
@@ -1078,7 +1104,7 @@ where
 
     // Move this object into a reference object, transferring ownership.
     pub fn into_ref(self) -> PyObjectRef {
-        PyRc::new(self)
+        PyRc::new(self).into()
```
skinny121, line 1107: Maybe add a `new` method to `PyObjectRc`.

skinny121 (changes requested):
Looks pretty good. I think we should be able to hide the fact that `PyObjectRc` is a thin wrapper around `Rc`/`Arc` pretty well, and have `PyObjectRc` offer all the interface that is needed instead of using `.rc()`/`PyObjectRc::into_rc()`.

vm/src/pyobject.rs
coolreader18: Nope -- blocked on the arbitrary_self_types feature

vm/src/pyobjectrc.rs
```diff
@@ -0,0 +1,136 @@
...
+
+impl PyObjectRc {
+    /// # Safety
+    /// if rc is dropped without wrapping again, drop will not be called
+    unsafe fn into_rc(this: Self) -> PyRc<DynPyObject> {
+        std::mem::transmute(this)
+    }
+
+    pub fn into_raw(this: Self) -> *const DynPyObject {
+        PyRc::into_raw(unsafe { Self::into_rc(this) })
```
skinny121, line 27: I think this can be done safely with: 
```rust
let ptr = PyRc::as_ptr(&this.inner);
mem::forget(this);
ptr
```
That how Rc and Arc implements this function.

youknowone:
this generic way seems not fit for drop implementation 🤦

coolreader18:
What if you make it generic over the payload for PyObject? e.g. `inner: PyRc<PyObject<T>>`

coolreader18:
I think that would make the name make more sense as well; PyObjectRc is a wrapper around Rc<PyObject>

youknowone:
@coolreader18 @skinny121 thanks for the reviews! I think this is finally done

vm/src/pyobjectrc.rs
```diff
@@ -0,0 +1,268 @@
...
+
+    pub fn downgrade(this: &Self) -> PyObjectWeak<T> {
+        PyRc::downgrade(&this.inner)
+    }
+
+    pub fn upgrade_weak(weak: &PyObjectWeak<T>) -> Option<Self> {
+        weak.upgrade().map(|inner| PyObjectRc { inner })
+    }
+
+    // pub unsafe fn new_uninit(value: T) -> PyObjectRc<>
```
skinny121, line 92: Remove?

skinny121 (approved):

youknowone:
after converting to embed PyObject<T>, it was possible to move some parts back to pyobject.rs

vm/src/pyobjectrc.rs
```diff
@@ -0,0 +1,199 @@
...
+    PyRc<PyObject<T>>: AsPyObjectRef,
+{
+}
+
+impl<T> Deref for PyObjectRc<T>
+where
+    T: ?Sized + PyObjectPayload,
+    PyRc<PyObject<T>>: AsPyObjectRef,
+{
+    type Target = PyRc<PyObject<T>>;
```
coolreader18, line 111: If we're making it unsafe to access the inner PyRc, we should probably make the target `PyObject<T>`

youknowone: I think trying to make target `PyObject<T>` is a good idea. But accessing to inner PyRc is safe. `into_rc` is unsafe because it skips dropping of PyObjectRc. This is borrowing, so there is no reason to be unsafe.

youknowone: oh, it can be a problem when it is cloned from the ref, got it.

vm/src/pyobjectrc.rs
```diff
@@ -0,0 +1,188 @@
...
+
+pub struct PyObjectRc<T = dyn PyObjectPayload>
+where
+    T: ?Sized + PyObjectPayload,
+    PyRc<PyObject<T>>: AsPyObjectRef,
+{
+    inner: PyRc<PyObject<T>>,
+}
+
+pub type PyObjectWeak<T = dyn PyObjectPayload> = PyWeak<PyObject<T>>;
```
coolreader18, line 15: Should probably be a newtype as well :grimacing: still the issue of downgrade->upgrade to Arc->drop original PyObjectRc->drop Arc without `__del__`. It would also let you add the upgrade() method to PyObjectWeak and have it return a PyObjectRc

coolreader18 (approved):

PR #2156, opened 2020-08-27
Fix FileIO.closefd always set when it dropped in __init__

youknowone:


skinny121:
What bug does this fix?

youknowone:
uh, nothing. thank you for checking

PR #2157, opened 2020-08-27
Implement array getitem by slice

qingshi163:


qingshi163:
It is my first PR, tell me if anything wrong. Tx

vm/src/stdlib/array.rs
```diff
@@ -194,14 +195,31 @@ macro_rules! def_array_enum {
...
+                    $(ArrayContentType::$n(v) => {
+                        let elements = v.get_slice_items(vm, &slice)?;
+                        let zelf = ArrayContentType::$n(elements);
+                        let obj = PyObject::new(
+                            PyArray {
+                                array: PyRwLock::new(zelf)
+                            },
+                            PyArray::class(vm),
+                            None
+                        );
```
youknowone, line 209: `zelf` is an alternative name of `self` when it is not allowed by rust languge restriction. In this code, it doesn't mean self but a sliced one. New variable name like `sliced` or something else would be better

youknowone (commented):
nice work, technically perfect! thanks for contributing

youknowone (approved):

PR #2158, opened 2020-08-27
Fix clippy warning

skinny121:
This is causing recent builds to fail. Rust 1.46.0 was just released, and the newer version of clippy linted against this case.

### DIFF
--- a/vm/src/stdlib/re.rs
+++ b/vm/src/stdlib/re.rs
@@ -237,8 +237,7 @@ fn do_split(
     // essentially Regex::split, but it outputs captures as well
     let mut output = Vec::new();
     let mut last = 0;
-    let mut n = 0;
-    for captures in pattern.regex.captures_iter(text) {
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/re.rs
```diff
@@ -237,16 +237,14 @@ fn do_split(
...
+    for (n, captures) in pattern.regex.captures_iter(text).enumerate() {
         let full = captures.get(0).unwrap();
         let matched = &text[last..full.start()];
         last = full.end();
         output.push(Some(matched));
         for m in captures.iter().skip(1) {
             output.push(m.map(|m| m.as_bytes()));
         }
-        n += 1;
         if maxsplit != 0 && n >= maxsplit {
```
youknowone, line 248: is this equivalent to before? For the first loop, `n` was 1 in the original code but now 0 in new code.

coolreader18: I think he did have n+1 >= maxsplit originally, so maybe that didn't actually work? @skinny121

PR #2160, opened 2020-08-28
Use siphash24 + PYTHONHASHSEED

coolreader18:
I didn't add test_hash b/c it expects strings are in the "unicodeobject" format, which is a union of `Vec<u8>|Vec<u16>|Vec<u32>`. Since that's not our actual internal string format, replicating it would require an allocation on every hash, which I felt was too heavy.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1503,6 +1503,7 @@ dependencies = [
  "once_cell",
  "parking_lot",
  "rand",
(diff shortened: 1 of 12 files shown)


youknowone (approved):

youknowone:
it seems importing only number parts from test_hash will be enough

PR #2161, opened 2020-08-28
chr accepts invalid unicode characters

youknowone:


coolreader18:
Nope, not that easy :stuck_out_tongue_winking_eye: 

Rust's string type is utf8, and its char type represents a scalar value -- all of the algorithms related to str/char make this assumption. Neither of those can represent a unicode surrogate value, which is a unicode code point but is **not** as unicode scalar value. That's what I'm sorta looking into with my `anicode` branch; using a string representation similar to what cpython has in order to store surrogates, so that stuff like surrogate_escape will work.

skinny121:
We could use https://docs.rs/wtf8/0.0.3/wtf8/, which is like utf-8 but allows encoding of surrogate code points. Rust itself uses an in tree version for OsString(https://github.com/rust-lang/rust/blob/master/library/std/src/sys_common/wtf8.rs).

youknowone:
Yep, not easy at all.
This is fixing a single case of test specifically using `chr`

coolreader18:
No, but it's UB to have a surrogate in a `String`. It's invalid utf8

youknowone:
uh, agree. it must be UB

PR #2163, opened 2020-08-28
Make class.slots completely immutable

coolreader18:


derive/src/pyclass.rs
```diff
@@ -174,7 +216,9 @@ pub(crate) fn impl_pystruct_sequence(
         )
     };
     let fake_ident = Ident::new("pystruct_sequence", struc.span());
-    let class_meta = ClassItemMeta::from_nested(struc.ident.clone(), fake_ident, attr.into_iter())?;
+    let meta = StructSeqItemMeta::from_nested(struc.ident.clone(), fake_ident, attr.into_iter())?;
+    let with_pyimpl = meta.inner()._bool("with_pyimpl")?;
+    let class_meta = ClassItemMeta(meta.0);
```
youknowone, line 221: Creating class_meta to share its method looks tricky. How about moving the methods into `impl ClassMetaInner` and use only `StructSeqItemMeta` here?

derive/src/pyclass.rs
```diff
@@ -206,6 +250,19 @@ pub(crate) fn impl_pystruct_sequence(
         }
     }
 
+    let (with_extend, with_slots) = if with_pyimpl {
```
youknowone, line 253: Will adding a new trait remove this conditional code? Probably `PyStructSequenceImpl` may include empty `__extend_py_class` just like `PyClassImpl` does

derive/src/pyclass.rs
```diff
@@ -691,18 +772,25 @@ fn extract_impl_attrs(
                     bail_span!(path, "Unknown pyimpl attribute")
                 }
             }
+            NestedMeta::Meta(Meta::Path(path)) if path_eq(&path, "structseq_impl") => {
```
youknowone, line 775: it seems `pyclass` always matches `pyimpl` and `pystruct_sequence(with_impl)` always matches `pyimpl(structseq_impl)`. If they can't share same `pyimpl`, it can use a new name like `pyimpl_sequence_struct` or `pystruct_sequence_impl`. Did I miss some advantages from this sharing?

youknowone (changes requested):

coolreader18:
Ooh, better idea: make `PyStructSequence` a trait that you add with `pyimpl(with())`. Then we make `pystruct_sequence` just a derive rather than a sort of duplicate pyclass

youknowone:
oh, that sounds great

derive/src/util.rs
```diff
@@ -223,7 +224,7 @@ impl ItemMeta for SimpleItemMeta {
     }
 }
 
-pub(crate) struct ClassItemMeta(ItemMetaInner);
+pub(crate) struct ClassItemMeta(pub ItemMetaInner);
```
coolreader18, line 227:
```suggestion
pub(crate) struct ClassItemMeta(ItemMetaInner);
```

coolreader18:
@youknowone can you review?

youknowone:
I will do it soon. I had a few busy days 😵

coolreader18:
No problem, take your time :)

vm/src/slots.rs
```diff
@@ -35,12 +36,22 @@ impl Default for PyTpFlags {
...
 pub struct PyClassSlots {
-    pub name: Option<String>, // tp_name, not class name
+    pub flags: PyTpFlags,
+    pub name: PyRwLock<Option<String>>, // tp_name, not class name
     pub new: Option<PyNativeFunc>,
     pub call: Option<PyNativeFunc>,
     pub descr_get: Option<PyDescrGetFunc>,
 }
 
+impl From<PyTpFlags> for PyClassSlots {
```
youknowone, line 46: not sure `From` trait fits here. it look like construct with flags rather than conversion from flags.

youknowone (approved):
Everything is pyclass, good!

derive/src/pyclass.rs
```diff
@@ -7,48 +7,95 @@ use proc_macro2::TokenStream;
...
+    let class_extensions = &context.class_extensions;
+    macro_rules! iter_chain {
+        ($($it:expr),*$(,)?) => {
+            std::iter::empty()
+                $(.chain(std::iter::once($it)))*
+        };
+    }
+    iter_chain![
+        parse_quote! {
+            fn __extend_py_class(
```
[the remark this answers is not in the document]
coolreader18: extend_class works by just calling the `#[extend_class]`-marked method in the generated extend_class method; for a PyClassImpl that's the impl_extend_class method and for a trait/interface that's this "magic" method; extend_class can't substitute for the actual generated method.

vm/src/macros.rs
```diff
@@ -38,30 +38,35 @@ macro_rules! py_class {
...
+    (@extract_slots($ctx:expr, $slots:expr, (slot $slot_name:ident), $value:expr)) => {
+        $slots.$slot_name = Some(
+            $crate::function::IntoPyNativeFunc::into_func($value)
+        );
+    };
+    (@extract_slots($ctx:expr, $class:expr, $name:expr, $value:expr)) => {};
+    (@extract_attrs($ctx:expr, $slots:expr, (slot $slot_name:ident), $value:expr)) => {};
+    (@extract_attrs($ctx:expr, $class:expr, $name:expr, $value:expr)) => {
+        $class.set_str_attr($name, $value);
+    };
```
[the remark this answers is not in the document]
coolreader18: Sort of; I know that `tokio` does that with [their macros](https://docs.rs/tokio/0.2.22/tokio/macro.select.html); I think it keeps the logic encapsulated so that when there's a `@extract_slots` inside you know that it's in the same macro definition.

PR #2164, opened 2020-08-29
Support arguments in jitted functions

skinny121:
To jit functions that have arguments, the type of the arguments needs to be known when it's compiled. this PR uses the annotations to determine the type of the arguments.

### DIFF
--- a/jit/src/lib.rs
+++ b/jit/src/lib.rs
@@ -1,4 +1,5 @@
 use std::fmt;
+use std::mem::MaybeUninit;
 
 use cranelift::prelude::*;
 use cranelift_module::{Backend, FuncId, Linkage, Module, ModuleError};
@@ -18,12 +19,6 @@ pub enum JitCompileError {
     BadBytecode,
     #[error("error while compiling to machine code: {0}")]
     CraneliftError(#[from] ModuleError),
-}
-
-#[derive(Debug, thiserror::Error)]
-pub enum JitExecutionError {
-    #[error("argument type differs from what is expected")]
-    ArgTypeMismatch,
 }
 
 struct Jit {
@@ -116,31 +111,23 @@ pub struct CompiledCode {
 }
 
 impl CompiledCode {
-    pub fn invoke(&self, args: &[AbiValue]) -> Result<Option<AbiValue>, JitExecutionError> {
-        let mut cif_args = Vec::new();
-        for (val, ty) in args.iter().zip(&self.sig.args) {
-            match (val, ty) {
-                (AbiValue::Int(ref val), JitType::Int) => {
-                    cif_args.push(libffi::middle::Arg::new(val))
-                }
-                (AbiValue::Float(ref val), JitType::Float) => {
-                    cif_args.push(libffi::middle::Arg::new(val))
-                }
-                _ => return Err(JitExecutionError::ArgTypeMismatch),
-            }
-        }
+    pub fn args_builder(&self) -> ArgsBuilder<'_> {
+        ArgsBuilder::new(self)
+    }
+
+    pub fn invoke<'a>(&self, args: &Args<'a>) -> Option<AbiValue> {
+        debug_assert_eq!(self as *const _, args.code as *const _);
         let cif = self.sig.to_cif();
         unsafe {
             let value = cif.call::<UnTypedAbiValue>(
                 libffi::middle::CodePtr::from_ptr(self.code as *const _),
-                &cif_args,
+                &args.cif_args,
             );
-            Ok(self.sig.ret.as_ref().map(|ty| value.to_typed(ty)))
+            self.sig.ret.as_ref().map(|ty| value.to_typed(ty))
         }
     }
 }
 
-#[derive(Default)]
 struct JitSig {
     args: Vec<JitType>,
     ret: Option<JitType>,
@@ -152,7 +139,7 @@ impl JitSig {
             Some(ref ty) => ty.to_libffi(),
             None => libffi::middle::Type::void(),
         };
-        libffi::middle::Cif::new(Vec::new(), ret)
+        libffi::middle::Cif::new(self.args.iter().map(JitType::to_libffi), ret)
     }
 }
 
@@ -178,12 +165,22 @@ impl JitType {
     }
 }
 
-#[derive(Clone)]
+#[derive(Copy, Clone)]
 pub enum AbiValue {
     Float(f64),
     Int(i64),
 }
 
+impl AbiValue {
+    fn to_untyped(self) -> UnTypedAbiValue {
+        match self {
+            AbiValue::Float(f) => UnTypedAbiValue { float: f },
+            AbiValue::Int(i) => UnTypedAbiValue { int: i },
+        }
+    }
+}
+
+#[derive(Copy, Clone)]
 union UnTypedAbiValue {
     float: f64,
     int: i64,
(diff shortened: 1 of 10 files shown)


jit/src/lib.rs
```diff
@@ -143,3 +211,54 @@ impl fmt::Debug for CompiledCode {
...
+    pub fn into_args(self) -> Option<Args<'a>> {
+        if self.initialised.iter().all(|x| *x) {
+            // SAFETY: we know all elements are initialised so it's safe to transmute away the
+            // MaybeUninit wrapper.
+            let values = unsafe {
+                std::mem::transmute::<Vec<MaybeUninit<UnTypedAbiValue>>, Vec<UnTypedAbiValue>>(
+                    self.values,
+                )
+            };
+            let cif_args = values.iter().map(|v| libffi::middle::Arg::new(v)).collect();
```
coolreader18, line 248: If you're going to recollect the vec anyway, you may as well just `v.assume_init()` in the mapping function and avoid the transmute.

vm/src/obj/objfunction.rs
```diff
@@ -1,5 +1,9 @@
 use super::objcode::PyCodeRef;
 use super::objdict::PyDictRef;
+#[cfg(feature = "jit")]
+use super::objfloat;
+#[cfg(feature = "jit")]
+use super::objint;
```
coolreader18, line 6: nit; you could condense this into one `#[cfg jit] use super::{objint, objfloat}`

vm/src/obj/objfunction.rs
```diff
@@ -313,10 +462,11 @@ impl PyFunction {
...
     #[cfg(feature = "jit")]
     #[pymethod(magic)]
-    fn jit(&self, vm: &VirtualMachine) -> PyResult<()> {
-        self.jitted_code
+    fn jit(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<()> {
+        zelf.jitted_code
             .get_or_try_init(|| {
-                rustpython_jit::compile(&self.code.code)
+                let arg_types = PyFunction::get_jit_arg_types(&zelf, vm)?;
+                rustpython_jit::compile(&zelf.code.code, &arg_types)
```
coolreader18, line 469: This is a lot of code in objfunction, maybe there could be a submodule `obj/objfunction/jitfunc.rs` or something to store all the logic for py<->jit conversion?

vm/src/obj/objfunction.rs
```diff
@@ -275,6 +285,145 @@ impl PyFunction {
...
+                Ok(JitType::Int)
+            } else if value.is(&vm.ctx.types.float_type) {
+                Ok(JitType::Float)
+            } else {
+                Err(vm.new_runtime_error(
+                    "Jit requires argument to be either int or float".to_owned(),
+                ))
+            }
+        } else {
+            Err(vm.new_runtime_error(format!("argument {} needs annotation", name)))
```
youknowone, line 302: maybe we need `new_jit_error`

jit/src/lib.rs
skinny121: I can't use `assume_init` as I need the reference into the `values` vector, `cif_args` is effectively a vector of pointers that are passed to `libffi`. This would be nicer if we had either https://github.com/rust-lang/rust/issues/63291 or https://github.com/rust-lang/rust/issues/63568.

coolreader18: Ah, gotcha

jit/src/lib.rs
```diff
@@ -96,23 +111,76 @@ pub struct CompiledCode {
...
+    fn to_cif(&self) -> libffi::middle::Cif {
+        let ret = match self.ret {
+            Some(ref ty) => ty.to_libffi(),
+            None => libffi::middle::Type::void(),
+        };
+        libffi::middle::Cif::new(self.args.iter().map(JitType::to_libffi), ret)
+    }
+}
+
+#[derive(Copy, Clone, PartialEq)]
```
coolreader18, line 146: The reason I didn't make this Copy is because I was thinking of eventually having composite types, like `List(Box<JitType>)` or `Tuple(Vec<JitType>)` or maybe some sort of product or sum type.

jit/src/lib.rs
```diff
@@ -143,3 +211,54 @@ impl fmt::Debug for CompiledCode {
...
+pub struct ArgsBuilder<'a> {
+    values: Vec<MaybeUninit<UnTypedAbiValue>>,
+    initialised: Vec<bool>,
+    code: &'a CompiledCode,
+}
+
+impl<'a> ArgsBuilder<'a> {
+    fn new(code: &'a CompiledCode) -> ArgsBuilder<'a> {
+        ArgsBuilder {
+            values: vec![MaybeUninit::uninit(); code.sig.args.len()],
```
coolreader18, line 224: Couldn't this be `Vec<Option<UntypedAbiArg>>`? and you could like do `self.values.iter().map(|val| val.as_ref().map(|val| middle::Abi::new(val)).collect::<Option<_>>()`

And you could just store the Vec<Option<>> in the Args struct, since it just matters that it's not dropped

ArniDagur:
An example like the following can cause UB, where the function is given incorrect input:

```python
def add_two(a: int, b: int) -> int:
    return a + b

add_two.__jit__()
result = add_two(2, 3.0)  # cpython would return 5.0 as a float
```

Here, I give a function that should take an `int` a `float` instead. CPython doesn't complain because it doesn't actually look at the type annotations. This branch understandably gives a nonsense result:
```
[arni][~/src/RustPython][jit-args]% target/release/rustpython
>>>>> def add_two(a: int, b: int) -> int:
..... 	return a + b
.....
>>>>> add_two.__jit__()
>>>>> add_two(2, 3.0)
4613937818241073154
>>>>> add_two(2, 3)
5
>>>>> def return_float() -> float:
..... 	return 3.0
.....
>>>>> add_two(2, return_float())
4613937818241073154
```

I was wondering if we're okay with sort of undefined behaviour when incorrect input is given to jitted functions.

If not, we could either

1. Do a runtime check of some sort; or
2. In the native code we output, insert calls to the same VM methods that would have been executed if it was bytecode. These can handle data of unknown type. It would still improve performance because we wouldn't have to match on the instruction every time, right? You'd only go through the match statements in https://github.com/RustPython/RustPython/blob/master/vm/src/frame.rs#L319 once. When working with data whose type we can determine statically, such as the output of in-built functions such as `range(...)` we could then output more specialized native code as we do in this PR. Perhaps this requires a change in the bytecode format to include type information when it's known?

I'm not sure if `2.` makes sense since I don't quite know how RustPython or interpreters in general work.

darleybarreto:
> I was wondering if we're okay with this sort of undefined behaviour when incorrect input is given to jitted functions?
> 
> If not, we could either
> 
> 1. Do a runtime check of some sort; or
> 2. In the native code we output, insert calls to the same VM methods that would have been executed if it was bytecode. These can handle data of unknown type. It would still improve performance because we wouldn't have to match on the instruction every time, right? You'd only go through the match statements in https://github.com/RustPython/RustPython/blob/master/vm/src/frame.rs#L319 once. When working with data whose type we can determine statically, such as the output of in-built functions such as `range(...)` we could then output more specialized native code as we do in this PR. Perhaps this requires a change in the bytecode format to include type information when it's known?

Another option is to use some sort of multiple dispatch, as Julia does. Which also implies introducing some sort of type inference system and other complex things... Perhaps it would be easier to just throw an error for now, and eventually introduce more complicated machinery.

coolreader18:
I think ideally we're trying to make a jitted function transparent -- it should be no different in behavior from a non-jitted function, and it shouldn't error because of a jit limitation (e.g. i64 ints instead of BigInts). So yeah, I think we should definitely type-check and then fall back to non-jit behavior to either error or get whatever dynamic behavior would happen with that.

coolreader18:
Doing something similar to Julia's multiple dispatch definitely sounds cool though -- with

```python
def add(a, b):
    return a + b
```

We could have like a hashmap of `Vec<JitType> -> CompiledCode`, and when it's called with e.g. `(int, int)`, we try compiling with `a` as an int and `b` as an int, and if that succeeds we populate the hashmap with that signature. The same thing could happen for `(float, float)`, or `(float, int)`, or `(int, float)`, etc

skinny121:
@ArniDagur Good catch, this PR did previously do a runtime check of the arguments types, and if it didn't match it would fallback to interpreting the bytecode instead, but a subsequent refactoring broke that.

@darleybarreto @coolreader18 I'm not familiar with Julia at all, but does sound like an interesting approach.

youknowone:
By looking PyPy practice, it make guards with typecheck and falling back to original code when the jit assumption failed.

PR #2165, opened 2020-08-29
add getcwdb function

hyperbora:


vm/src/stdlib/os.rs
```diff
@@ -682,6 +682,11 @@ mod _os {
             .to_owned())
     }
 
+    #[pyfunction]
+    fn getcwdb(vm: &VirtualMachine) -> PyResult<Vec<u8>> {
+        Ok(getcwd(vm)?.as_bytes().to_vec())
```
youknowone, line 687: to avoid redundant copy, `into_bytes()` is available in `String`

https://doc.rust-lang.org/std/string/struct.String.html#method.into_bytes

youknowone (changes requested):

vm/src/stdlib/os.rs
hyperbora: OK, i fixed it.

youknowone (approved):

PR #2167, opened 2020-08-30
add gethostbyname function

hyperbora:


youknowone:
Because it looks like a thin wrapper of C socket library gethostbyname, I am not sure this is the correct way to go. @coolreader18 how do you think?

youknowone (approved):

PR #2169, opened 2020-08-30
Troubleshooting displaying bytes in the wrong hexadecimal number in certain cases.

hyperbora:
Problems occur when byte values are 12, 14, and 15.


coolreader18:
Actually, I think you should be able to just fix the fallback case and remove the special cases for 0..=8 and the ones that you're adding. You can do `format!("\\x{:02x}", i)` to ensure that the hex form is padded to 2 digits if it isn't already 2 digits long

hyperbora:
OK, I fixed it.

youknowone (approved):

PR #2166, opened 2020-08-29
Implement array setitem by slice

qingshi163:

PR #2170, opened 2020-08-30
Implement array setitem by slice

qingshi163:
I just cleared out the git commits, sorry I am new of doing the PR.

Maybe I had miss something but as my throught, At PyArray setitem() we can not just simply call something like self.borrow_value_mut().setitem_by_slice(obj) because if [self] is referring to the same object as [obj] than we can not borrow obj anymore as it is borrowed mut by [self].

The alternate is we have to check if [self] and [obj] is refering to the same object before we doing the borrow, if it is we clone our self otherwise we pass a ref to the next function. But I did not get any idea to do so, the [self] is &PyArray and the [obj] is PyObjectRef.

### DIFF
--- a/vm/src/stdlib/array.rs
+++ b/vm/src/stdlib/array.rs
@@ -4,21 +4,21 @@ use crate::common::cell::{
 };
 use crate::function::OptionalArg;
 use crate::obj::objbytes::PyBytesRef;
-use crate::obj::objsequence::{PySliceableSequence, get_slice_range};
+use crate::obj::objsequence::{get_slice_range, PySliceableSequence};
 use crate::obj::objslice::PySliceRef;
 use crate::obj::objstr::PyStringRef;
 use crate::obj::objtype::PyClassRef;
 use crate::obj::{objbool, objiter};
 use crate::pyobject::{
     BorrowValue, Either, IntoPyObject, PyArithmaticValue, PyClassImpl, PyComparisonValue,
-    PyIterable, PyObject, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject, TypeProtocol
+    PyIterable, PyObject, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject, TypeProtocol,
 };
 use crate::VirtualMachine;
 use crossbeam_utils::atomic::AtomicCell;
 use itertools::Itertools;
-use std::fmt;
 use num_bigint::BigInt;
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/array.rs
```diff
@@ -440,7 +538,23 @@ impl PyArray {
         obj: PyObjectRef,
         vm: &VirtualMachine,
     ) -> PyResult<()> {
-        self.borrow_value_mut().setitem(needle, obj, vm)
+        match needle {
+            Either::A(i) => self.borrow_value_mut().setitem_by_idx(i, obj, vm),
+            Either::B(slice) => {
+                let items = match obj.payload::<PyArray>() {
```
coolreader18, line 544: If you want to check reference equality you can change `&self` to `zelf: PyRef<Self>`, and then you can do `zelf.is(&obj)` to check if they're the same

youknowone:
reopened as #2166

vm/src/stdlib/array.rs
```diff
@@ -435,12 +534,36 @@ impl PyArray {
...
         vm: &VirtualMachine,
     ) -> PyResult<()> {
-        self.borrow_value_mut().setitem(needle, obj, vm)
+        match needle {
+            Either::A(i) => zelf.borrow_value_mut().setitem_by_idx(i, obj, vm),
+            Either::B(slice) => {
+                let cloned;
+                let guard;
+                let items = if zelf.is(&obj) {
+                    cloned = zelf.borrow_value().clone();
```
youknowone, line 548: is `derive(Clone)` required by this one?

youknowone (approved):
Looks good!

vm/src/stdlib/array.rs
qingshi163: Yes, it clone out value from zelf, we need to modify zelf later so we need a copy, a ArrayContentType enum with a Vec. Is any reason we don't use derive(Clone)?

youknowone: Can we copy it *after* slicing the original? It seems it is wasting copy cost a little bit

qingshi163: We must use all of the items that supplied on the right side of operator, otherwise it is a err. If step is 1, we do replace and insert(splice). If step is not 1, the length of slice must equal to length of items. So here should have no waste unless the err happend.

youknowone: Oh, that's right.

PR #2172, opened 2020-08-31
Implement dict doc & reversed magic method

mode9:
Implement a document and reversed feature of dictionary type.
- Add dict_reversekeyiterator_type
- Add dict_reversevalueiterator_type
- Add dict_reverseitemiterator_type

Tests expect python 3.8 or higher (or RustPython)


tests/snippets/builtin_dict.py
```diff
@@ -1,3 +1,5 @@
+import pytest
```
youknowone, line 1: Unfortunately, pytest is running the test but not available from the test code. See `testutils.py`. it includes `assert_raises` function which almost working like pytest.raises

mode9: Could you tell me how to cPython test in local?

youknowone: README describes how to run test in local.
Roughly, running a single file is: `python tests/snippets/builtin_dict.py`

vm/src/obj/objdict.rs
```diff
@@ -650,36 +671,99 @@ macro_rules! dict_iterator {
...
+                if self.dict.entries.has_changed_size(&self.size) {
+                    return Err(
+                        vm.new_runtime_error("dictionary changed size during iteration".to_owned())
+                    );
+                }
+                let count = self.position.load() + 1;
+                match self.dict.len().checked_sub(count) {
+                    Some(mut pos) => {
+                        let (key, value) = self.dict.entries.next_entry(&mut pos).unwrap();
+                        self.position.store(count);
```
coolreader18, line 705: @palaviv can you check the atomicity of this?

youknowone:
Normally, contributors never use `merge` command (including `pull`) in github projects. Only github does it.

Will you try to rebase your work?
When `RustPython/RustPython` is remote `upstream`, (try `git remote add upstream https://github.com/RustPython/RustPython.git` if you don't have one yet)
`git rebase upstream/master` will remove merge commits from your patch and attach your commits onto top of `upstream/master`.

mode9:
> You chose a big work as a first contriibution ;) the PR generally looks good.
> 
> Normally, contributors never use `merge` command (including `pull`) in github projects. Only github does it.
> 
> Will you try to rebase your work?
> When `RustPython/RustPython` is remote `upstream`, (try `git remote add upstream https://github.com/RustPython/RustPython.git` if you don't have one yet)
> `git rebase upstream/master` will remove merge commits from your patch and attach your commits onto top of `upstream/master`.

Thank you for your advice. Thanks to `rebase`, Commit history got cleaned up.

vm/src/obj/objdict.rs
youknowone: I think @palaviv was busy for a while. Because I don't want to keep this PR open for too long time, here is a suggestion:

`count` looks like to be safely grow enough until overflow.
```rust
let count = self.position.fetch_add(1);
match ... {
    Some(..) => ... // we don't nees to store a new value here. it already was `fetch_add`ed. and this is the hot path
    None => { self.position.store(self.dict.len()+1); ... } // prevent overflow here
}
```

mode9: I fixed this issue, but don't know it's what you expected.

coolreader18: Yup

coolreader18 (approved):

PR #2173, opened 2020-08-31
Fix: array with float type should accept integer parameter

qingshi163:


vm/src/stdlib/array.rs
```diff
@@ -377,6 +385,44 @@ def_array_enum!(
     (Double, f64, 'd'),
 );
 
+trait TryIntoFromObject: Sized {
```
coolreader18, line 388: Maybe this could be named something like `ArrayElement`?

vm/src/stdlib/array.rs
```diff
@@ -105,23 +106,27 @@ macro_rules! def_array_enum {
             fn count(&self, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
                 match self {
                     $(ArrayContentType::$n(v) => {
-                        Ok(<Option<$t>>::try_from_object(vm, obj)?.map_or(0, |val| {
+                        Ok(if let Ok(val) = $t::try_into_from_object(vm, obj) {
```
coolreader18, line 109: I don't think this needs to return a result anymore, you can change it to just return `usize`

PR #2176, opened 2020-09-01
implement socket.inet_pton, socket.inet_ntop

hyperbora:


vm/src/stdlib/socket.rs
```diff
@@ -596,6 +598,56 @@ fn socket_gethostbyname(name: PyStringRef, vm: &VirtualMachine) -> PyResult<Stri
...
+        ip_string
+            .borrow_value()
+            .parse::<Ipv6Addr>()
+            .map(|ip_addr| vm.ctx.new_bytes(ip_addr.octets().to_vec()))
+            .map_err(|_| {
+                vm.new_os_error("illegal IP address string passed to inet_pton".to_owned())
+            })
+    } else {
+        return Err(vm.new_os_error("Address family not supported by protocol".to_owned()));
+    };
```
coolreader18, line 621: I think you should be able to do a match on this, something like:

```rust
fn socket_inet_pton(af_inet: i32, ...) -> PyResult {
    match af_inet {
        c::AF_INET => /* parse Ipv4Addr */,
        c::AF_INET6 => /* parse Ipv6Addr */,
        _ => Err(vm.new_os_error("...")),
    }
}
```

vm/src/stdlib/socket.rs
```diff
@@ -596,6 +598,56 @@ fn socket_gethostbyname(name: PyStringRef, vm: &VirtualMachine) -> PyResult<Stri
...
+    n
+}
+
+fn socket_inet_ntop(af_inet: PyIntRef, packed_ip: PyBytesRef, vm: &VirtualMachine) -> PyResult {
+    let af_inet = af_inet.borrow_value();
+    let p = if af_inet.eq(&BigInt::from(c::AF_INET)) {
```
coolreader18, line 627: You should be able to do a similar thing here with a `match`

hyperbora:
Thank you for your review. I revised it as you told me.

PR #2177, opened 2020-09-01
Fix test_calendar tests

coolreader18:


src/main.rs
```diff
@@ -36,6 +36,19 @@ fn main() {
         settings.initialization_parameter = InitParameter::InitializeInternal;
     }
 
+    // don't translate newlines (\r\n <=> \n)
```
skinny121, line 39: Does CPython do this? My concern is that as this will have process wide effects, this may have unintended consequences for programs that embed RustPython.

coolreader18: > this may have unintended consequences for programs that embed RustPython.

That's why I moved it from `vm::initialize` to main, it should only happen for the actual rustpython binary.

> Does CPython do this?

yup: https://github.com/python/cpython/blob/9b648a95ccb4c3b14f1e87158f5c9f5dbb2f62c0/Python/initconfig.c#L1812-L1817

coolreader18:
Ahh, I think I found the problem -- our TextIOWrapper doesn't translate newlines. Maybe. Like when the script process prints to stdout, it prints `'\n'`s, but the stdout TextIOWrapper doesn't translate that. I think I can make a quickfix just by changing the `conv()` function in `test_calendar`

skinny121 (approved):

PR #2179, opened 2020-09-03
test_support

coolreader18:


coolreader18:
I'm so confused... all the tests pass fine on my machine but in CI they all fail with "Address family not supported by protocol [family]". Is it something with the github actions environment?

skinny121:
They also pass on my machine. After a bit of searching I'm wondering if it's an IPv4/IPv6 issue, because it appears that they don't support it yet https://github.com/actions/virtual-environments/issues/668. https://stackoverflow.com/questions/2370388/socketexception-address-incompatible-with-requested-protocol is a stack overflow question about the error on windows which indicate an IPv4/IPv6 problem.I

vm/src/obj/objfunction.rs
```diff
@@ -379,6 +377,22 @@ impl PyBoundMethod {
...
+            self.function.is(&other.function) && self.object.is(&other.object),
+        ))
+    }
+
+    #[pymethod(magic)]
+    fn ne(&self, other: PyObjectRef) -> PyResult<PyComparisonValue> {
+        let other = class_or_notimplemented!(Self, other);
+        Ok(PyComparisonValue::Implemented(
+            !self.function.is(&other.function) || !self.object.is(&other.object),
+        ))
```
youknowone, line 394: when eq returns PyComparisonValue, we can add ne by using it
```suggestion
        Ok(self.eq(other)?.map(|v| !v))
```

it also looks like there is no path to return Err

coolreader18: Oh, yeah, ther PyResult is there because `class_or_notimplemented!()` returns an `Ok(NotImplemented)` when it fails.

youknowone (approved):
big progress! 👏

PR #2180, opened 2020-09-04
Refactor use trait to combine the slice operation

qingshi163:


vm/src/obj/objsequence.rs
```diff
@@ -12,6 +12,270 @@ use crate::function::OptionalArg;
 use crate::pyobject::{BorrowValue, PyObject, PyObjectRef, PyResult, TryFromObject, TypeProtocol};
 use crate::vm::VirtualMachine;
 
+pub trait PySliceableSequenceMut: PySliceableSequence {
```
youknowone, line 15: should this trait require PySliceableSequence? If `fn as_slice(&self) -> PySliceableSequence` is possible, we can skip `impl PySliceableSequence for Vec` which may causes unexpected cost

youknowone (commented):
impressive, Nice work!

youknowone (approved):
Nice work! Definitely a lot better than before

PR #2181, opened 2020-09-05
Remove PyValue::into_object & Rename PyValue::into_simple_object to PyValue::into_object

edwardycl:
`PyValue::into_object` was first introduced here: https://github.com/RustPython/RustPython/pull/2079

It was originally intended to be used in `into_ref_with_type`.
However, the definition of `PyValue` has changed, and `into_object` is not used anywhere.

I also renamed `into_simple_object` to `into_object`


coolreader18 (commented):
I think this is definitely better; the `into_object()` method with cls and dict was barely used and conflicts with how the presence of a dict for an object is a property of its type now. There should be a few instances of `.into_ref(vm).into_object()` that you should be able to replace with `.into_object(vm)` now as well

youknowone (approved):
It looks good. nice catch that PyValue::into_object is not used

PR #2182, opened 2020-09-05
Pass function refs to single_or_tuple_any

edwardycl:
The reference of `Fn` object also implements `Fn`.

Passing refs will solve the ownership problem with recursive calls, so the `Checker` struct can be removed.


coolreader18 (approved):
Oh, this is so much better. Thanks for contributing this!

PR #2186, opened 2020-09-05
Add macOS development requirements.

ohgree:
autoconf, automake, and libtool are required as an external dependencies on macOS.
This PR adds the guide at DEVELOPMENT.md

After Rust update, I've experienced errors regarding the compilation of [libffi-sys](https://crates.io/crates/libffi-sys), and after some trial and error, installing automake and libtool solved the problem for me.


youknowone:
Referring `libffi-sys` will be helpful for future maintanence why this is required and whether it still is required

youknowone (approved):

PR #2187, opened 2020-09-05
Add Jit integration tests

skinny121:
This add a test utility macro `jit_function`, which allows declaring jitted python functions inside rust test functions. Example:
```rust
let add = jit_function! { add(a:i64, b:i64) -> i64 => r##"
    def add(a: int, b: int):
        return a + b
"## };
```
The above define the python function `add`, which is compiled to bytecode at compile time. That bytecode is then jitted into native code at runtime(we could also do this a compile time in the future). The return value of the macro is a lambda with the signature `fn(i64, i64) -> i64`.

As part of that I refactored `py_compile_bytecode` into two different macros `py_compile` and `py_freeze`.

### DIFF
--- a/jit/tests/common.rs
+++ b/jit/tests/common.rs
@@ -1,0 +1,184 @@
+use std::collections::HashMap;
+
+use rustpython_bytecode::bytecode::{CodeObject, Constant, Instruction, NameScope};
+use rustpython_jit::{CompiledCode, JitType};
+
+#[derive(Debug, Clone)]
+pub struct Function {
+    code: Box<CodeObject>,
+    name: String,
+    annotations: HashMap<String, StackValue>,
+}
+
+impl Function {
+    pub fn compile(self) -> CompiledCode {
+        let mut arg_types = Vec::new();
+        for arg in self.code.arg_names.iter() {
+            let arg_type = match self.annotations.get(arg) {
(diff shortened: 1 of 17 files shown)


jit/tests/common.rs
```diff
@@ -0,0 +1,203 @@
...
+            Instruction::Duplicate => {
+                let value = self.stack.last().unwrap().clone();
+                self.stack.push(value);
+            }
+            Instruction::Rotate { amount } => {
+                let mut values = Vec::new();
+
+                // Pop all values from stack:
+                for _ in 0..amount {
+                    values.push(self.stack.pop().unwrap());
```
coolreader18, line 132: You might be able to do something like `values.extend(stack.drain(len-amount..).rev())` here

jit/tests/common.rs
```diff
@@ -0,0 +1,203 @@
...
+                }
+
+                // Push top of stack back first:
+                self.stack.push(values.remove(0));
+
+                // Push other value back in order:
+                for value in values.into_iter().rev() {
+                    self.stack.push(value);
```
coolreader18, line 140: Can use extend() here too

coolreader18 (approved):
Nice work! I really like the freeze/compile separation, and the tests for jit look good

PR #2188, opened 2020-09-05
implement socket.getprotobyname

hyperbora:


vm/src/stdlib/socket.rs
```diff
@@ -646,6 +646,17 @@ fn socket_inet_ntop(af_inet: i32, packed_ip: PyBytesRef, vm: &VirtualMachine) ->
     }
 }
 
+fn socket_getprotobyname(name: PyStringRef, vm: &VirtualMachine) -> PyResult {
+    use std::ffi::CString;
+    let cstr = CString::new(name.borrow_value()).unwrap();
```
youknowone, line 651: what happens if name includes a null byte?

youknowone (approved):

vm/src/stdlib/socket.rs
hyperbora: Thank you for your review. I revised it as you told me.

coolreader18 (approved):

Issue #2183, opened 2020-09-05
`nonlocal` raise `SyntaxError` when it is declared multiple times.


minoring:
**CPython 3.7.4**
```python
def test_nonlocal():
    x = 0
    y = 0
    def f():
        nonlocal x
        nonlocal x, y

def test_nonlocal2():
    x = 0
    def f():
        nonlocal x
        nonlocal x
```
are OK.

**RustPython** gives `SyntaxError: name 'x' is used prior to nonlocal declaration at line 0 column 0` for both of the examples.

minoring:
I would like to handle this problem.

PR #2190, opened 2020-09-05
Fix nonlocal incompatibility

### LINKED ISSUE #2189: nonlocal incompatibility
###  Here is the list of incompatibility of `nonlocal` between RustPython and CPython 

```python
  def f(a):
    nonlocal a
```
* CPython 3.8
  `SyntaxError: name 'a' is parameter and nonlocal`
* RustPython 
`SyntaxError: no binding for nonlocal 'a' found at line 0 column 0`

```python
def f():
  a = 3
  def g():
    print(a)
    nonlocal a
  g()
```
* CPython 3.8
`SyntaxError: name 'a' is used prior to nonlocal declaration`
* RustPython
 `No Error. f() prints 3`

```python
  def f():
    a: int
    nonlocal a
```
* CPython 3.8
`SyntaxError: annotated name 'a' can't be nonlocal`

* RustPython
`SyntaxError: no binding for nonlocal 'a' found at line 0 column 0`

```python
  def f():
    a = 3
    def g():
      a = 4
      nonlocal a
    g()
    print(a)
```
* CPython 3.8
`SyntaxError: name 'a' is assigned to before nonlocal declaration`
* RustPython
 `No error. f() prints 4`

minoring:
Fix #2189 and #2183

### DIFF
--- a/compiler/src/symboltable.rs
+++ b/compiler/src/symboltable.rs
@@ -954,10 +954,10 @@ impl SymbolTableBuilder {
         // Some checks:
         let containing = table.symbols.contains_key(name);
         if containing {
+            let symbol = table.symbols.get(name).unwrap();
             // Role already set..
             match role {
                 SymbolUsage::Global => {
-                    let symbol = table.symbols.get(name).unwrap();
                     if let SymbolScope::Global = symbol.scope {
                         // Ok
                     } else {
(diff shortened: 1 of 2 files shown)


compiler/src/symboltable.rs
```diff
@@ -954,10 +954,10 @@ impl SymbolTableBuilder {
         // Some checks:
         let containing = table.symbols.contains_key(name);
         if containing {
```
coolreader18, line 956: I think you can replace this with `if let Some(symbol) = table.symbols.get(name) {`

coolreader18 (approved):
Looks good!

compiler/src/symboltable.rs
minoring: Thank you for your tip :) 
I made refactoring to replace `containing` with `Some(symbol)`
PTAL!

coolreader18:
Nice work :)

PR #2191, opened 2020-09-06
Fix bool.from_bytes

hyperbora:


vm/src/obj/objint.rs
```diff
@@ -233,6 +233,12 @@ impl PyInt {
     {
         if cls.is(&vm.ctx.types.int_type) {
             Ok(vm.ctx.new_int(value).downcast().unwrap())
+        } else if cls.is(&vm.ctx.types.bool_type) {
+            if PyInt::from(value).value.is_zero() {
```
coolreader18, line 237: You can use `ctx.new_bool` here, and you don't need the `PyInt::from(value). value`, that's the same thing as `value`

hyperbora:
Thank you for your review. I revised it as you told me.

coolreader18 (approved):
Thanks for contributing!

vm/src/obj/objint.rs
```diff
@@ -233,6 +233,12 @@ impl PyInt {
     {
         if cls.is(&vm.ctx.types.int_type) {
             Ok(vm.ctx.new_int(value).downcast().unwrap())
+        } else if cls.is(&vm.ctx.types.bool_type) {
+            Ok(vm
+                .ctx
+                .new_bool(!value.into().eq(&BigInt::zero()))
```
youknowone, line 239: there is also `is_zero` method

https://docs.rs/num/0.3.0/num/trait.Zero.html#tymethod.is_zero

PR #2192, opened 2020-09-06
Fix and improve array compare operations

qingshi163:


coolreader18:
Weird, looks like there's a deadlock in an importlib test_deadlock function only on macos... I'll try rerunning the test and see if it was a fluke

qingshi163:
Is it all right?

vm/src/stdlib/array.rs
```diff
@@ -337,6 +339,18 @@ macro_rules! def_array_enum {
                     ret
                 })
             }
+
+            fn cmp(&self, other: &ArrayContentType) -> Result<Option<Ordering>, ()> {
+                match self {
+                    $(ArrayContentType::$n(v) => {
+                        if let ArrayContentType::$n(other) = other {
```
coolreader18, line 346: You could also do something like

```
match (self, other) {
     (ArrayContentType::$n(v), ArrayContentType::$n(other)) => cmp(),
     _ => Err(())
}
```

coolreader18 (approved):
This looks great! I have one nit, otherwise this is good to merge!

PR #2194, opened 2020-09-07
Mini changes pulled from #2146

coolreader18:


vm/src/stdlib/collections.rs
```diff
@@ -247,7 +247,32 @@ mod _collections {
             self.maxlen.store(maxlen);
         }
 
-        #[pymethod(name = "__repr__")]
+        #[pymethod(magic)]
+        fn getitem(&self, idx: isize, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 251: oh.. it even didn't have getitem

vm/src/obj/objint.rs
```diff
@@ -92,20 +93,24 @@ macro_rules! impl_into_pyobject_int {
 
 impl_into_pyobject_int!(isize i8 i16 i32 i64 usize u8 u16 u32 u64 BigInt);
 
+pub fn try_to_primitive<'a, I>(i: &'a BigInt, vm: &VirtualMachine) -> PyResult<I>
```
youknowone, line 96: 👍

vm/src/cformat.rs
```diff
@@ -420,25 +435,22 @@ impl CFormatString {
...
-                                // TODO: handle errors when truncating BigInt to usize
-                                *q = Some(CFormatQuantity::Amount(
-                                    objint::get_value(&width_obj).to_usize().unwrap(),
-                                ));
-                                Ok(tuple_index)
-                            }
+                Some(CFormatQuantity::FromValuesTuple) => match elements.next() {
+                    Some(width_obj) => {
+                        tuple_index += 1;
+                        if !objtype::isinstance(&width_obj, &vm.ctx.types.int_type) {
```
youknowone, line 441: swapping if and else would make else block a lot cleaner.
but it seems it is not part of this PR

youknowone (approved):

vm/src/stdlib/collections.rs
coolreader18: Yeah, I had the same reaction 😁

PR #2197, opened 2020-09-09
Implement array __copy__() and __deepcopy__()

qingshi163:


vm/src/stdlib/array.rs
```diff
@@ -571,6 +571,19 @@ impl PyArray {
         self.borrow_value_mut().reverse()
     }
 
+    #[pymethod(magic)]
+    fn copy(&self, vm: &VirtualMachine) -> PyObjectRef {
+        PyArray {
+            array: PyRwLock::new(self.array.read().clone()),
+        }
+        .into_object(vm)
```
coolreader18, line 579: You don't need the `into_object()` -- you can directly return a PyArray and some trait/macro magic will do the `into_object(vm)` conversion for you -- you don't even need to take a `&VirtualMachine` as an argument

coolreader18 (approved):
Nice work!

PR #2200, opened 2020-09-12
Small rework for IntoPyNativeFunc

coolreader18:


vm/src/function.rs
```diff
@@ -475,91 +476,111 @@ pub type PyNativeFunc = Box<py_dyn_fn!(dyn Fn(&VirtualMachine, PyFuncArgs) -> Py
 
 /// Implemented by types that are or can generate built-in functions.
 ///
-/// For example, any function that:
+/// This trait is implemented by any function that matches the pattern:
```
youknowone, line 479: :+1: this is a lot clear than before

youknowone (approved):
wow, i like this PR a lot

PR #2201, opened 2020-09-12
implement socket.getnameinfo and add constants

hyperbora:


vm/src/stdlib/socket.rs
```diff
@@ -658,6 +658,32 @@ fn socket_getprotobyname(name: PyStringRef, vm: &VirtualMachine) -> PyResult {
...
+        },
+        _ => {
+            let error_type = vm.class("_socket", "gaierror");
+            Err(vm.new_exception_msg(
+                error_type,
+                "nodename nor servname provided, or not known".to_owned(),
+            ))
+        }
+    }
+}
```
youknowone, line 685: Because they share the errors, creating error can be moved out of match
```suggestion
    let addr = get_addr(vm, address)?;
    let nameinfo = addr.as_std().and_then(|addr| dns_lookup::getnameinfo(&addr, flags).ok());
    nameinfo.ok_or_else(|| {
          let error_type = vm.class("_socket", "gaierror");
          Err(vm.new_exception_msg(
              error_type,
              "nodename nor servname provided, or not known".to_owned(),
          ))
     })
```

hyperbora: Thank you for your review. I revised it as you told me.

youknowone (approved):

PR #2202, opened 2020-09-12
Fix bytes setitem with Indexable value

qingshi163:


vm/src/bytesinner.rs
```diff
@@ -340,24 +340,29 @@ impl PyBytesInner {
...
 
+    fn value_try_from_object(vm: &VirtualMachine, object: PyObjectRef) -> PyResult<u8> {
+        let value = vm.to_index(&object).ok_or_else(|| {
+            vm.new_type_error(format!(
+                "'{}' object cannot be interpreted as an integer",
+                object.class().name
+            ))
+        })?;
+        // __index__ returned non-int type
+        let value = value?;
```
youknowone, line 359: is line 359 needed to be seperated?
```suggestion
        })??;
        // __index__ returned non-int type
```

youknowone (approved):
Looks good!

vm/src/bytesinner.rs
qingshi163: I'd like to spec what could return from the function. Or we just merge it together.
```suggestion
        })??;
```

PR #2203, opened 2020-09-12
Add rich comparison for dict_iterator

clemado1:
Modified inner_eq to inner_cmp so that dict_iterator can use and added rich comparison for dict_iterator.

### DIFF
--- a/vm/src/obj/objdict.rs
+++ b/vm/src/obj/objdict.rs
@@ -3,18 +3,21 @@ use std::fmt;
 use crossbeam_utils::atomic::AtomicCell;
 
 use super::objiter;
+use super::objset::PySet;
 use super::objstr;
 use super::objtype::{self, PyClassRef};
 use crate::dictdatatype::{self, DictKey};
 use crate::exceptions::PyBaseExceptionRef;
 use crate::function::{KwArgs, OptionalArg, PyFuncArgs};
 use crate::pyobject::{
-    BorrowValue, IdProtocol, IntoPyObject, ItemProtocol, PyAttributes, PyClassImpl, PyContext,
-    PyIterable, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject, TypeProtocol,
+    BorrowValue, IdProtocol, IntoPyObject, ItemProtocol, PyArithmaticValue, PyAttributes,
+    PyClassImpl, PyComparisonValue, PyContext, PyIterable, PyObjectRef, PyRef, PyResult, PyValue,
+    TryFromObject, TypeProtocol,
 };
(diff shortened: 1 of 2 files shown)


vm/src/obj/objdict.rs
```diff
@@ -146,22 +147,32 @@ impl PyDict {
...
+        zelf: PyRef<Self>,
+        other: PyDictRef,
+        size_func: fn(usize, usize) -> bool,
+        item: bool,
+        vm: &VirtualMachine,
+    ) -> PyResult<bool> {
+        if size_func(zelf.len(), other.len()) {
             return Ok(false);
         }
+        let other_cp = other.copy();
```
youknowone, line 160: I guess this is not nessessary if original code will also work for le or gt if we take smaller dict for loop instead of zelf. 
Then remained size of the other side of looped one will tell it is bigger than smaller one or not.

clemado1: Thank you for your advice. I revised it to iterate a smaller one.

vm/src/obj/objdict.rs
```diff
@@ -146,46 +149,65 @@ impl PyDict {
...
-        if other.entries.len() != zelf.entries.len() {
-            return Ok(false);
+    fn inner_cmp(
+        zelf: PyRef<Self>,
+        other: PyDictRef,
+        size_func: fn(usize, usize) -> bool,
+        item: bool,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyComparisonValue> {
+        if size_func(zelf.len(), other.len()) {
```
youknowone, line 159: For eq and ne, this test leads the loop comparison.
But for other comparisons like lt, gt, the size comparison doesn't guarantee subset.

Assume `a =  {'a': 1, 'b' 2}; b = {'a': 1, 'b': 3, 'c': 4}`,
`a.items() > b.items()` is false but also `b.items() > a.items()` is false.

clemado1: Thank you for your advice! but don't the cpython also return false and false? I put it at first because it seems cpython check a length at first. [link](https://github.com/python/cpython/blob/c75330605d4795850ec74fdc4d69aa5d92f76c00/Objects/dictobject.c#L4239) Should I change it as you said?

youknowone: Oh, I am sorry. I thought something totally wrong. Yes, it seems current behaviour is exactly as expected and also same as also what I described. Sorry for making confusion!

youknowone (approved):

vm/src/obj/objdict.rs
clemado1: No problem! I really want to thank you for your help.

PR #2205, opened 2020-09-12
Run tests/clippy on all crates

skinny121:
Fairly recently we removed passing `--all` when running clippy/tests on CI as the wasm crate can't handle be compiled with the threading feature, but that caused rust tests that aren't in the root crate to not be run on CI.

Split from #2187 

cc @coolreader18

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,6 +7,15 @@ name: CI
 
 env:
   CARGO_ARGS: --features "ssl jit"
+  NON_WASM_PACKAGES: >
+    -p rustpython-bytecode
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -7,6 +7,15 @@ name: CI
 
 env:
   CARGO_ARGS: --features "ssl jit"
+  NON_WASM_PACKAGES: |
+    -p rustpython-bytecode
```
coolreader18, line 11: Maybe use `>` instead of `|` to strip newlines

coolreader18 (approved):

PR #2206, opened 2020-09-13
Fix bytes function

qingshi163:


vm/src/bytesinner.rs
```diff
@@ -410,6 +416,48 @@ impl PyBytesInner {
...
+
+    pub fn append(&mut self, object: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        let value = Self::value_try_from_object(vm, object)?;
+        self.elements.push(value);
+        Ok(())
+    }
+
+    pub fn remove(&mut self, object: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        let value = Self::value_try_from_object(vm, object)?;
+        if let Some(index) = self.elements.iter().position(|&x| x == value) {
```
youknowone, line 447: `self.elements.find(value)` may work

vm/src/bytesinner.rs
```diff
@@ -1010,19 +1058,7 @@ impl PyBytesInner {
...
-            let n = n.to_usize().unwrap(); // always positive by outer if condition
-
-            let old = self.elements.clone();
-
-            self.elements.reserve((n - 1) * old.len());
-            for _ in 1..n {
-                self.elements.extend(&old);
-            }
-        }
+        self.elements = self.repeat(n);
```
youknowone, line 1061: unlike original code, this is replacing the existing characters again.
I feel like it can make difference when `*= 1` is called a lot.

youknowone (changes requested):

vm/src/bytesinner.rs
qingshi163: Vec.find return a reference to the item and we can't do deleting.

youknowone: Will you check bstr's find? It returns index instead of reference.
I think we use it to implement bytes.find too.

https://docs.rs/bstr/0.2.13/bstr/trait.ByteSlice.html#method.find

qingshi163: Thanks for the suggestion, I replaced it by bstr function.

youknowone (approved):

youknowone:
nice work! you are conquering all the arrays!

PR #2207, opened 2020-09-14
Add a tp_cmp slot for richcompare

coolreader18:


vm/src/slots.rs
```diff
@@ -148,10 +163,173 @@ pub trait SlotDescriptor: PyValue {
...
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum PyComparisonOp {
+    Lt,
+    Le,
+    Eq,
+    Ne,
+    Ge,
+    Gt,
+}
```
youknowone, line 252: is this mandatory for python spec or a utillity to allow tp_cmp? I thought this is CPython internal stuff.
If we can avoid enum, i prefer that way. We need to think more about inlining if we use enum for that.
(I didn't check it is possible - but maybe we can have 6 tp_ slots for comparison?)

Otherwise, appending `#[inline]` for `cmp` maybe helpful to inline it when we call them with constant enum for each operation for specific type not through tp_cmp.

coolreader18: I think this enum actually makes code a lot more readable; we were already doing a lot of

```rust
fn cmp() {}
fn lt() { self.cmp(|a, b| a < b) }
```

So this just makes it more standardized and less repetitive, rather than 10 types with 6 different comparison op functions that all call `self.cmp(|a, b| a OP b)`, we can deduplicate a lot of it and pass data around (`PyComparisonOp`) instead of functionality (`F: Fn(usize, usize) -> bool`). And yeah, I think it probably would be good to try `#[inline]`-ing the `cmp()` function to avoid branching in a loop, if possible.

vm/src/obj/objset.rs
```diff
@@ -106,76 +106,31 @@ impl PySetInner {
         self.content.contains(vm, needle)
     }
 
-    #[inline]
-    fn _compare_inner(
+    fn compare(
```
youknowone, line 109: `#[inline]`

vm/src/obj/objset.rs
```diff
@@ -106,76 +106,31 @@ impl PySetInner {
...
-        swap: bool,
+        op: PyComparisonOp,
         vm: &VirtualMachine,
     ) -> PyResult<bool> {
-        let (zelf, other) = if swap { (other, self) } else { (self, other) };
-
-        if size_func(zelf.len(), other.len()) {
+        if op == PyComparisonOp::Ne {
+            return self.compare(other, PyComparisonOp::Eq, vm).map(|eq| !eq);
+        }
```
youknowone, line 117: is this still required? It seems eq and ne can be handled by eval_ord-like way.

vm/src/obj/objset.rs
```diff
@@ -106,76 +106,31 @@ impl PySetInner {
...
+        if !op.eval_ord(self.len().cmp(&other.len())) {
             return Ok(false);
         }
-        for key in other.content.keys() {
-            if !zelf.contains(&key, vm)? {
+        let (superset, subset) = if matches!(op, PyComparisonOp::Lt | PyComparisonOp::Le) {
```
youknowone, line 121: maybe adding method like `op.has_less()`  will be helpful

vm/src/pyobject.rs
```diff
@@ -1402,6 +1411,26 @@ impl TryFromObject for std::time::Duration {
...
+    }
+}
+
+impl<T: TryFromObject> TryFromObject for PyArithmaticValue<T> {
+    fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
+        let arith = match PyArithmaticValue::from_object(vm, obj) {
+            PyArithmaticValue::Implemented(x) => Self::Implemented(T::try_from_object(vm, x)?),
+            PyArithmaticValue::NotImplemented => Self::NotImplemented,
+        };
+        Ok(arith)
```
youknowone, line 1430: PyArithmaticValue implements common option functions. So this is possible:
```suggestion
        PyArithmaticValue::from_object(vm, obj).map(|v| T::try_from_object(vm, x)).transpose()
```

vm/src/sequence.rs
```diff
@@ -39,34 +37,30 @@ pub(crate) fn eq(vm: &VirtualMachine, zelf: DynPyIter, other: DynPyIter) -> PyRe
     }
 }
 
-fn cmp<L, O>(zelf: DynPyIter, other: DynPyIter, len_cmp: L, obj_cmp: O) -> PyResult<bool>
-where
-    L: Fn(usize, usize) -> bool,
-    O: Fn(PyObjectRef, PyObjectRef) -> PyResult<Option<bool>>,
-{
-    let fallback = len_cmp(zelf.len(), other.len());
+pub fn cmp(
```
youknowone, line 40: #[inline]

vm/src/slots.rs
```diff
@@ -148,10 +163,176 @@ pub trait SlotDescriptor: PyValue {
...
+        zelf: PyObjectRef,
+        other: PyObjectRef,
+        op: PyComparisonOp,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyComparisonValue> {
+        let zelf = PyRef::try_from_object(vm, zelf)?;
+        Self::cmp(zelf, other, op, vm)
+    }
+
+    fn cmp(
```
youknowone, line 187: Adding comment to idiomatically add `#[inline]` for impls of this trait will be helpful. We actually only need to inline `cmp`s which is called inside of RustPython code, but that kind of tracking is not easy. Adding `#[inline]` for every impl is easy and hopefully will not be inlined if it is too bad to inline - if llvm is good enough.

youknowone (changes requested):
I thought about this PR and now I agree this is better approach. Most of cmp will be called from python code, which we can obviously inline by `#[inline(always)]` to `PyBaseObject::tp_cmp`. And then we need to take care about other internal calls of cmps in rust side, but it will be controllable size. Probably only arithmatic values matter. Sequences anyway cannot avoid to call through Python layer which never can be inlined.

vm/src/pyobject.rs
coolreader18: `transpose()` on PyArithmaticValue returns a `Result<Option<T>, E>` though, I guess I could do an into() on it but this seemed less verbose than all that

youknowone: By thinking of usage of transpose, I think that's bug of `ResultLike`. Let me fix it.

vm/src/obj/objset.rs
coolreader18: Adding this fixed an issue, I think it's kinda a weird case cause for `ne` and only `ne`, if a comparison that's predicated on equality (like `bool_seq_{lt,gt}`) fails, then it should return true instead of false. I'll try again to figure out how to make it work without this special-case, I agree that it seems like it should be possible.

vm/src/slots.rs
coolreader18: Actually, I don't think it would optimize very much, cause the `__op__` methods will only be called when explicitly called like `x.__gt__(other)` from Python; `x > other` will still call `(cls.slots.cmp)(a, b, PyComparisonOp::Gt, vm)`

vm/src/pyobject.rs
youknowone: result-like 0.3.0 will have fixed signature. I will do it later if you don't want to do it here.

vm/src/slots.rs
youknowone: ah, unfortune. maybe we can try multiple slot items for each cmp with proper benchmarks later

youknowone (approved):
feel free to merge it once it is done

coolreader18:
Should I merge now and try individual op slots later?

youknowone:
Sure. I think individual op slots are non-trivial optimization. We'd better to have benchmark before doing that.

PR #2208, opened 2020-09-15
new wasm example: notebook inspired by alpha.iodide.io

mireille-raad:
This PR is for a new example under the `wasm` directory. Inspired by iodide, a Mozilla Project https://alpha.iodide.io/ (now abandoned).

This example re-uses the code from `wasm/demo` and adds a basic import code functionality from `api.github.com`

Here is how it looks like:
Screenshot from 2020-09-14 21-51-25

Here is how the import works:
Screenshot from 2020-09-14 21-53-25

Here are the url for test imports (can be anything public as long as it uses the github api url):
https://api.github.com/repos/RustPython/RustPython/contents/wasm/demo/snippets/asyncbrowser.py
https://api.github.com/repos/RustPython/RustPython/contents/wasm/demo/snippets/fetch.py

Here is how the notebook looks like after fetching
Screenshot from 2020-09-14 21-53-42

Then code execution 
Screenshot from 2020-09-14 22-13-27

Changes:
- Removed xterm
- Changed the output from a text area to an HTML element (so later on, I can toy with adding markdown, graphs or interactive elements). If you print an html tag for example, it renders ;p 
- For faster load time, I disable the default snippet imports and execution
- Added the ability to import public python code from GitHub.
- Removed things in the UI

Submitting a PR and looking for feedback on:
- code
- UI tweaks
- features like should this only import from api.github.com or add others?
- Instead of this, should I try to do a notebook that looks like a Jupyter Lab Notebook instead?

### DIFF
--- a/wasm/notebook/package.json
+++ b/wasm/notebook/package.json
@@ -1,0 +1,39 @@
+{
+    "name": "app",
+    "version": "1.0.0",
+    "description": "Bindings to the RustPython library for WebAssembly",
+    "main": "index.js",
+    "dependencies": {
+        "codemirror": "^5.42.0",
+        "local-echo": "^0.2.0",
+        "xterm": "^3.8.0"
+    },
+    "devDependencies": {
+        "@wasm-tool/wasm-pack-plugin": "^1.1.0",
+        "clean-webpack-plugin": "^3.0.0",
+        "css-loader": "^3.4.1",
+        "html-webpack-plugin": "^3.2.0",
+        "mini-css-extract-plugin": "^0.9.0",
+        "raw-loader": "^4.0.0",
+        "serve": "^11.0.2",
+        "webpack": "^4.16.3",
+        "webpack-cli": "^3.1.0",
+        "webpack-dev-server": "^3.1.5"
+    },
+    "scripts": {
+        "dev": "webpack-dev-server -d",
+        "build": "webpack",
+        "dist": "webpack --mode production",
+        "test": "webpack --mode production && cd ../tests && pipenv run pytest"
+    },
+    "repository": {
+        "type": "git",
+        "url": "git+https://github.com/RustPython/RustPython.git"
+    },
+    "author": "Ryan Liddle",
+    "license": "MIT",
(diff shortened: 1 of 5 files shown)


coolreader18:
Is there a reason it can't just fetch from the raw github url? e.g. https://raw.githubusercontent.com/RustPython/RustPython/master/wasm/demo/snippets/asyncbrowser.py

mireille-raad:
The reason for using the GitHub API is:
- I was thinking we could use some of the github api features/integrations later on
- I could have done some javascript that figures out the github api url from the raw one, but I was a bit lazy. 

Now that I thought of it, it is better to just import raw files from everywhere, the github integration is not worth it or it is just a lock-in.

I updated the code which made it less complicated since there is no Base64 encoding/decoding of the content.

I tested by uploading the same code to a Netlify site:
https://reverent-fermat-2bdba5.netlify.app/test.txt 

The only thing that we need to give heads up about is that CORS configuration needs to be ` Access-Control-Allow-Origin = "*"`

coolreader18:
Do we want this to be at like `rustpython.github.io/demo/notebook`? Then we could upload to the same repo as the other demo, we wouldn't have to set up a new one with new deploy keys

coolreader18:
Oh, and it probably goes without saying, but this looks really great :))

mireille-raad:
Why not. It would be easier/better. And other demos can go there as well.

coolreader18 (approved):
LGTM; if you want to mess around with the github workflow feel free, otherwise I could merge this and edit it separately

mireille-raad:
If you can merge and edit separately, that would be better.

coolreader18:
Thanks for contributing!

youknowone:
@mireille-raad would you please check gitter messages when possible?

youknowone:
> Instead of this, should I try to do a notebook that looks like a Jupyter Lab Notebook instead?

I think both iodide and jupyter notebook style is useful. But which one will be more useful (or more impressive) for now? We must consider we now have limited features of python.

PR #2210, opened 2020-09-16
Implement pickle for bytes

qingshi163:


vm/src/obj/objbytearray.rs
```diff
@@ -563,6 +564,23 @@ impl PyByteArray {
...
         bytes_decode(zelf.into_object(), args, vm)
     }
+
+    #[pymethod(magic)]
+    fn reduce_ex(&self, _proto: usize, vm: &VirtualMachine) -> (PyClassRef, PyTuple) {
+        self.reduce(vm)
+    }
+
+    #[pymethod(magic)]
+    fn reduce(&self, vm: &VirtualMachine) -> (PyClassRef, PyTuple) {
```
qingshi163, line 574: I find that we have to impl __reduce_ex__ even we don't need to check the protocol, if I remove this function than __reduce__ will not be called also an error happens when pickling. Should we get it sorted? Any idea?

coolreader18:
It looks like `bytes` doesn't override its `__reduce__` function at all:

```py
>>> bytes.__reduce_ex__ is object.__reduce_ex__
True
>>> bytes.__reduce__ is object.__reduce__
True
```

And then [here's](https://github.com/python/cpython/blob/7219e27087baaa8a5693b5bef1b1357bddbffa53/Objects/bytearrayobject.c#L2114) the CPython implementation of `bytearray.__reduce__`, it looks like it actually does need to check the protocol.

vm/src/obj/objbytearray.rs
```diff
@@ -563,6 +565,17 @@ impl PyByteArray {
...
+    #[pymethod(magic)]
+    fn reduce_ex(zelf: PyRef<Self>, _proto: usize, vm: &VirtualMachine) -> (PyClassRef, PyTuple) {
+        Self::reduce(zelf, vm)
+    }
+
+    #[pymethod(magic)]
+    fn reduce(zelf: PyRef<Self>, vm: &VirtualMachine) -> (PyClassRef, PyTuple) {
+        let bytes = PyBytes::from(zelf.borrow_value().elements.clone()).into_pyobject(vm);
+        (Self::class(vm), PyTuple::from(vec![bytes]))
+    }
```
qingshi163, line 578: Unlike CPython, I find out that protocol does not need treat different than we get the right result. I think CPython use it for compatibility with python2.

extra_tests/snippets/bytearray.py
```diff
@@ -720,3 +721,15 @@
...
 b = bytearray(b'my bytearray')
 assert a.__ne__(b) is True
+
+# pickle
+a = bytearray(b'\xffab\x80\0\0\370\0\0')
+# TODO: Different output when pickle dumping a str if protocol is 0
+# CPython's output is like:
+# b'c__builtin__\nbytearray\np0\n(c_codecs\nencode\np1\n(V\xffab\x80\\u0000\\u0000\xf8\\u0000\\u0000\np2\nVlatin1\np3\ntp4\nRp5\ntp6\nRp7\n.'
+# We do have fully Interoperability between our output and CPython's
+assert pickle.dumps(a, 0) == b'c__builtin__\nbytearray\np0\n(c_codecs\nencode\np1\n(V\xffab\x80\x00\x00\xf8\x00\x00\np2\nVlatin1\np3\ntp4\nRp5\ntp6\nRp7\n.'
```
qingshi163, line 731: CPython use double backslash but our str use one for '\uxxxx', and for '\0' CPython use '\\u0000' but ours is '\x00'.
But interoperability seem works both side.

qingshi163:
Still working in progress, trying for compatibility...

qingshi163:
@coolreader18 I had check the implement of python lib pickle.py, the different output between our's and CPython under protocol version 0 is because the version of pickle.py.
Currently we are using pickle that from CPython 3.6.0 but testing with CPython 3.8.
The new version of pickle.py added the special function for bytearray
```
    def save_bytearray(self, obj):
        if self.proto < 5:
            if not obj:  # bytearray is empty
                self.save_reduce(bytearray, (), obj=obj)
            else:
                self.save_reduce(bytearray, (bytes(obj),), obj=obj)
            return
        n = len(obj)
        if n >= self.framer._FRAME_SIZE_TARGET:
            self._write_large_bytes(BYTEARRAY8 + pack("<Q", n), obj)
        else:
            self.write(BYTEARRAY8 + pack("<Q", n) + obj)
    dispatch[bytearray] = save_bytearray
```

and has different treating for str
```
    def save_str(self, obj):
        if self.bin:
            encoded = obj.encode('utf-8', 'surrogatepass')
            n = len(encoded)
            if n <= 0xff and self.proto >= 4:
                self.write(SHORT_BINUNICODE + pack("<B", n) + encoded)
            elif n > 0xffffffff and self.proto >= 4:
                self._write_large_bytes(BINUNICODE8 + pack("<Q", n), encoded)
            elif n >= self.framer._FRAME_SIZE_TARGET:
                self._write_large_bytes(BINUNICODE + pack("<I", n), encoded)
            else:
                self.write(BINUNICODE + pack("<I", n) + encoded)
        else:
            obj = obj.replace("\\", "\\u005c")
            obj = obj.replace("\0", "\\u0000")
            obj = obj.replace("\n", "\\u000a")
            obj = obj.replace("\r", "\\u000d")
            obj = obj.replace("\x1a", "\\u001a")  # EOF on DOS
            self.write(UNICODE + obj.encode('raw-unicode-escape') +
                       b'\n')
        self.memoize(obj)
    dispatch[str] = save_str
```

coolreader18:
Oh, nice catch! You can update it to the 3.8 version, just copy the file and put it in its own commit that mentions it's from v3.8.x.

vm/src/obj/objbytearray.rs
youknowone: I think python3 still support lower proto versions.

youknowone (approved):

coolreader18 (approved):
Nice work!

PR #2212, opened 2020-09-18
WASI: add a note for rust wasm target setup

pmp-p:
cf https://github.com/rustwasm/wasm-bindgen/issues/929

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ You can build the WebAssembly WASI file with:
 cargo build --release --target wasm32-wasi --features="freeze-stdlib"
 ```
 
-> Note: we use the `freeze-stdlib` to include the standard library inside the binary.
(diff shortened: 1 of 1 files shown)


coolreader18:
You could probably put the `rustup target add wasm32-wasi` inside the code block before the cargo build, with a shell comment before it like "# if you haven't built for wasi before:"

pmp-p:
As a total newcomer to rust, I'd think it's better to learn about what is a "target" and that it's possible to switch to CI toolchains.
 eg on emscripten sdk  the tot ("tip-of-tree") target was not documented but could have saved me time instead of strugging with faulty clang at the start  => always better to know your tools .

youknowone:
If this patch will not be revised, I think current patch is still better than original one.

youknowone (approved):

PR #2215, opened 2020-09-18
Rename str types

### LINKED ISSUE #2065: [RFC] Rename strings
## Summary

Rename python `str` related stuff to `str` instead of `string` in Rust side too.
Give the name `string` for abstract string type for places which require either `str` or `bytes`(-like).

## Detailed Explanation

For now, Python `str` is belong to `objstr` and the implementation is `PyString`. So it is occupying both the names.
Because both `bytes` and `str` are conceptionally a string In python, the boundary of them is not very clear.
Some functions only take `bytes`, other functions only take `str`, but the other functions take both `bytes` and `str`.

By unifying the name `str` for python `str` and giving the name `string` to the conceptional common strings, we will have clear names for every use case of string-like objects.

example code with the name:
```rust
type PyStringArg = Either<PyStr, PyBytes>;
type PyStringLikeArg = Either<PyStr, PyBytesLike>;

#[pyfunction]
fn system(command: PyStringArg) -> ... {
    unsafe { libc::system(command.as_bytes()) }
}
```

### Drawback
- `str` and `string` doesn't give a clear idea about what they do for new comers.
- `string` is still used in Python. The standard library `string` module will be an exception of this renaming.

### Alternatives
CPython uses `unicode` and `bytes` for its code. But this is based on its historical name from Python2, not a decision for the name. We may follow this policy.

Any other suggestion to make clear their roles are welcomed.

youknowone:
closes #2065

### DIFF
--- a/vm/src/stdlib/io.rs
+++ b/vm/src/stdlib/io.rs
@@ -26,7 +26,7 @@ mod _io {
     use crate::obj::objbytes::PyBytesRef;
     use crate::obj::objint;
     use crate::obj::objiter;
-    use crate::obj::objstr::{self, PyString, PyStringRef};
+    use crate::obj::objstr::{self, PyStr, PyStrRef};
     use crate::obj::objtype::{self, PyClassRef};
     use crate::pyobject::{
         BorrowValue, IntoPyObject, PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue,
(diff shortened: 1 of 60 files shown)


vm/src/stdlib/io.rs
```diff
@@ -178,20 +178,20 @@ impl BufferedIO {
 }
 
 #[derive(Debug)]
-struct PyStringIO {
```
skinny121, line 181: I'm wondering whether we should keep this as `PyStringIO` as it's called `StringIO` in Python.

skinny121 (approved):

youknowone:
@coolreader18 any comment about this change?

coolreader18:
LGTM

PR #2216, opened 2020-09-18
PyClass -> PyType

youknowone:
We referes this type everywhere `type` both in Python and internal code.

this seems giving confusion for new comers and doesn't seem giving any advantage.

Let me know if I am missing some benefits from this.


coolreader18:
I think we can keep PyClassImpl/Def as is, since it's a class definition, not a type definition

youknowone:
@skinny121 @coolreader18 will you review this change?

coolreader18 (approved):

skinny121 (approved):

PR #2220, opened 2020-09-19
Update dependencies

coolreader18:
Includes the wasi fix for the `instant` crate


vm/src/stdlib/os.rs
```diff
@@ -2080,27 +2080,27 @@ mod posix {
 
     #[pyfunction(name = "WIFSIGNALED")]
     fn wifsignaled(status: i32) -> bool {
-        unsafe { libc::WIFSIGNALED(status) }
+        libc::WIFSIGNALED(status)
```
youknowone, line 2083: what's happening here? is this safe in linux but unsafe in macos?

coolreader18: Looks like it, I guess I'll just put `#[allow(unused_unsafe)]`

youknowone: https://github.com/rust-lang/libc/issues/1888

youknowone: how about fixing libc as 0.2.76 for a while? It seems they will be all safe after fixing the issue

youknowone:
Does CI uses older version of Rust for windows? This is somewhat unexpected.

coolreader18:
I added a rust-toolchain file; I don't think we've ever explicitly talked about what versions we support but I feel like we've never shied away from using features in a version that just released, so I think "latest stable" is a good thing to keep to.

youknowone (approved):

PR #2223, opened 2020-09-20
slot optimization to skip attributes dict lookup

youknowone:


vm/src/slots.rs
```diff
@@ -40,7 +42,7 @@ pub struct PyClassSlots {
     pub flags: PyTpFlags,
     pub name: PyRwLock<Option<String>>, // tp_name, not class name
     pub new: Option<PyNativeFunc>,
-    pub call: Option<PyNativeFunc>,
+    pub call: AtomicCell<u128>,
```
youknowone, line 45: does anyone have idea using neither lock nor unsafe?

coolreader18: Just `fn(&VirtualMachine, PyFuncArgs) -> PyObjectRef`? Could generate it in pyslot like
```rust
slots.#slot_ident = Some(
    |vm: &::rustpython_vm::VirtualMachine, args: ::rustpython_vm::function::PyFuncArgs| -> ::...::PyResult {
        <::rustpython_vm::function::IntoPyNativeFunc<_>>::call(&Self::#ident, vm, args)
    } as _
);
```

youknowone: thank you! it seems working

vm/src/slots.rs
```diff
@@ -35,25 +38,65 @@ impl Default for PyTpFlags {
     }
 }
 
+type GenericMethod = fn(PyObjectRef, PyFuncArgs, &VirtualMachine) -> PyResult;
+type DescrGetFunc =
+    fn(&VirtualMachine, PyObjectRef, Option<PyObjectRef>, OptionalArg<PyObjectRef>) -> PyResult;
+type HashFunc = Box<py_dyn_fn!(dyn Fn(PyObjectRef, &VirtualMachine) -> PyResult<PyHash>)>;
```
coolreader18, line 44: tp_hash could probably be a plain fn() as well

vm/src/obj/objtype.rs
```diff
@@ -72,43 +72,70 @@ impl PyClassRef {
...
-impl PyClassRef {
     pub fn iter_mro(&self) -> impl Iterator<Item = &PyClassRef> + DoubleEndedIterator {
         std::iter::once(self).chain(self.mro.iter())
     }
 
+    pub(crate) fn first_in_mro<F, R>(&self, f: F) -> Option<R>
+    where
+        F: Fn(&Self) -> Option<R>,
+    {
+        f(self).or_else(|| self.mro.iter().filter_map(|cls| f(cls)).next())
```
coolreader18, line 84:
```suggestion
        self.iter_mro().find_map(|cls| f(cls))
```

vm/src/vm.rs
```diff
@@ -889,20 +884,25 @@ impl VirtualMachine {
...
+                    args: args.args.clone(),
+                    kwargs: args.kwargs.clone(),
+                };
+                self.trace_event(TraceEvent::Call)?;
+                let result = slot_call(callable.clone(), args, self);
+                self.trace_event(TraceEvent::Return)?;
+                result
+            })
+        });
+        invoked.unwrap_or_else(|| {
```
coolreader18, line 900: You could just have `first_in_mro(|cls| cls.slots.call.load())` and invoke it outside of the closure.

vm/src/obj/objtype.rs
youknowone: Due to the different expectation of the cost, I changed its form and added comments

vm/src/slots.rs
```diff
@@ -35,25 +39,78 @@ impl Default for PyTpFlags {
...
+}
+
+#[inline]
+fn get_class_magic(zelf: &PyObjectRef, name: &str) -> PyObjectRef {
+    zelf.get_class_attr(name).unwrap()
+
+    // TODO: we already looked up the matching class but lost the information here
+    // let cls = zelf.lease_class();
+    // let attrs = cls.attributes.read();
+    // attrs.get(name).unwrap().clone()
```
youknowone, line 113: we are double looking up the matching class. can slot refer its class? weak ref?

vm/src/slots.rs
```diff
@@ -35,25 +39,78 @@ impl Default for PyTpFlags {
...
     }
+
+    #[inline]
+    pub(crate) fn update_slot_func(&self, name: &str) {
+        match name {
+            "__call__" => {
+                let func: GenericMethod = |zelf, args, vm| {
+                    let magic = get_class_magic(&zelf, "__call__");
+                    let magic = vm.call_if_get_descriptor(magic, zelf.clone())?;
+                    args.insert(zelf);
```
coolreader18, line 72: `insert` returns a new PyFuncArgs

vm/src/slots.rs
coolreader18: Are you looking to sort of store the class that has the `__call__` method in the call slot, as if it's a closure, so that you can just do `class_call.attrs.get("__call__")`? not suggesting that we make it a closure, just trying to figure out what the optimization would be?

youknowone: Basically yes. We can find the class 2 ways.
1. In update_slot_func, the owner of `__call__` is always the type owning the slots. 
2. When finding Some() slot in first_in_mro, it is always the type owning `__call__`

vm/src/obj/objtype.rs
```diff
@@ -252,7 +261,7 @@ impl PyClass {
                 return Ok(());
             }
         }
-
+        zelf.slots.update_slot_func(attr_name.borrow_value());
```
coolreader18, line 264: Looks like CPython checks that the name is dunderscored before calling update_slot, probably a micro-op but :shrug: https://github.com/python/cpython/blob/3.8/Objects/typeobject.c#L3324-L3326

vm/src/slots.rs
youknowone: giving additional class parameter to slot funcs will work

vm/src/pyobject.rs
```diff
@@ -487,6 +487,15 @@ impl PyObjectRef {
             Err(self)
         }
     }
+
+    pub fn as_downcasted<T: PyObjectPayload + PyValue>(&self) -> Option<&PyRef<T>> {
```
coolreader18, line 491: `downcast_ref`, maybe? `Any` methods are like that; `downcast{,_{ref,mut}}`

coolreader18 (approved):
LGTM, I think it does make sense to continue developing this in a separate PR

qingshi163:
After I  merged this commit, the slot for bufferprotocol gives me a compile error

```
error[E0599]: no method named `store` found for enum `std::option::Option<std::boxed::Box<(dyn for<'r> std::ops::Fn(pyobjectrc::PyObjectRc, &'r vm::VirtualMachine) -> std::result::Result<std::boxed::Box<(dyn obj::objmemory::BufferProtocol + 'static)>, pyobject::PyRef<exceptions::PyBaseException>> + std::marker::Send + std::marker::Sync + 'static)>>` in the current scope
   --> vm/src/slots.rs:359:1
    |
359 | #[pyimpl]
    | ^^^^^^^^^ method not found in `std::option::Option<std::boxed::Box<(dyn for<'r> std::ops::Fn(pyobjectrc::PyObjectRc, &'r vm::VirtualMachine) -> std::result::Result<std::boxed::Box<(dyn obj::objmemory::BufferProtocol + 'static)>, pyobject::PyRef<exceptions::PyBaseException>> + std::marker::Send + std::marker::Sync + 'static)>>`
    |
```

I think the problem comes from the new code of here

```

impl<Item> ImplItem<Item> for SlotItem
where
    Item: ItemLike + ToTokens + GetIdent,
{
    fn gen_impl_item(&self, args: ImplItemArgs<'_, Item>) -> Result<()> {
        let ident = if args.item.is_function_or_method() {
            Ok(args.item.get_ident().unwrap())
        } else {
            Err(self.new_syn_error(args.item.span(), "can only be on a method"))
        }?;

        let item_attr = args.attrs.remove(self.index());
        let item_meta = SlotItemMeta::from_attr(ident.clone(), &item_attr)?;

        let slot_ident = item_meta.slot_name()?;
        let slot_name = slot_ident.to_string();
        let tokens = {
            if slot_name == "new" {
                let into_func = quote_spanned! {ident.span() =>
                    ::rustpython_vm::function::IntoPyNativeFunc::into_func(Self::#ident)
                };
                quote! {
                    slots.#slot_ident = Some(#into_func);
                }
            } else if slot_name == "cmp" {
                let into_func = quote_spanned! {ident.span() =>
                    ::std::boxed::Box::new(Self::#ident)
                };
                quote! {
                    slots.#slot_ident = Some(#into_func);
                }
            } else {
                let into_func = if slot_name == "call" {
                    quote_spanned! {ident.span() =>
                        |zelf: ::rustpython_vm::pyobject::PyObjectRef,  args: ::rustpython_vm::function::PyFuncArgs, vm: &::rustpython_vm::VirtualMachine| -> ::rustpython_vm::pyobject::PyResult {
                            use ::rustpython_vm::pyobject::TryFromObject;
                            let zelf = PyRef::<Self>::try_from_object(vm, zelf)?;
                            Self::#ident(zelf, args, vm)
                        }
                    }
                } else {
                    quote_spanned! {ident.span() =>
                        Self::#ident
                    }
                };
                quote! {
                    slots.#slot_ident.store(Some(#into_func as _))
                }
            }
        };

        args.context.extend_slots_items.add_item(
            format!("(slot {})", slot_name),
            args.cfgs.to_vec(),
            tokens,
        )?;

        Ok(())
    }
}
```

the errer message is totally unreadable and we should not expect for this error right?

coolreader18:
Oh, you should change the `Box<py_dyn_fn!()>` to be like the other Func type aliases, just a fn()

youknowone:
oh.. the PR would have lots of conflict. sorry for making that. probably #2231 will make more conflict but the situation will be a bit clearer. `new` is the only exception. other slots share same form.

qingshi163:
@coolreader18 I tried to change the function type but not work but I did work around treating as "cmp", just the proc-macro is difficult for me to read...
but I am thinking we could mark all the possible slot name on the pass so easy to tracing for the futher as the compiler gives so little infomation for the macro..

youknowone:
For now after #2231, `new` is the only exception and all other slots are sharing the `fn()` form. So I don't want to recommend to do like cmp in this PR but follow other slot functions. Will you push your work to the working PR regardless it compiles or not? let's reveal together.

qingshi163:
Sure

PR #2224, opened 2020-09-20
impl SlotCall for PyClass, PyFunction, JsonScanner

youknowone:


coolreader18 (approved):
LGTM, I wasn't sure whether this was supposed to be before or after #2223

youknowone:
When I work on big feature, I create small independent derived works as seperated PR. So the PRs have reversed order in this case, maybe confusing.
Reviewing this first will reduce the size of #2223 but you anyway finishied all of them, thanks!

Issue #2195, opened 2020-09-08
[RFC] Iterator for memoryview


darleybarreto:
## Summary

This [issue](https://bugs.python.org/issue41732) on bugs.python (and [impl](https://github.com/python/cpython/commit/b0ac5d75a59c356c44cfc9b25077da3049dd16e9)) reports a patch to iterate over a memoryview object. 

## Detailed Explanation
I am new to Rust and RustPython project itself so I do not know a fast rusty way of implementing this.

## Drawbacks, Rationale, and Alternatives
I am not providing an implementation, so I cannot argue alternatives. :(

## Unresolved Questions
I don't know.

qingshi163:
@youknowone this issue is fixed by #2226 , can we close it?

Issue #2125, opened 2020-08-17
`bytes(bytes)` is inefficient


edwardycl:
## Feature

`bytes` object is immutable, so constructing a new `bytes` object from another `bytes` object doesn't need copying.
However, `PyBytes` and `PyByteArray` share `PyBytesInner` implementation, causing an unnecessary copying of immutable data.

CPython:
```
>>> b = b'foo'
>>> bytes(b) is b
True
```

RustPython:
```
>>> b = b'foo'
>>> bytes(b) is b
False
```

## Python Documentation

https://docs.python.org/3/library/stdtypes.html#bytes-objects

awoimbee:
Fixed by #2284
@qingshi163 you should use github keywords (`Fixes #<issue number>`) ;)

PR #2226, opened 2020-09-21
Implement Buffer Protocol

qingshi163:
I add a slot for implement the buffer protocol, more or less like what CPython doing. I am not sure is the right way for RustPython. Also I am confusing about the slots seem do not inherit from the base? So how we control the behavour of the subclass?

### DIFF
--- a/Lib/test/test_memoryview.py
+++ b/Lib/test/test_memoryview.py
@@ -134,8 +134,6 @@ class AbstractMemoryTests:
             with self.assertRaises(TypeError):
                 del m[1:4]
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_tobytes(self):
         for tp in self._types:
             m = self._view(tp(self._source))
@@ -146,16 +144,12 @@ class AbstractMemoryTests:
             self.assertEqual(b, expected)
             self.assertIsInstance(b, bytes)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_tolist(self):
         for tp in self._types:
             m = self._view(tp(self._source))
             l = m.tolist()
             self.assertEqual(l, list(b"abcdef"))
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_compare(self):
         # memoryviews can compare for equality with other objects
         # having the buffer interface.
@@ -379,8 +373,6 @@ class AbstractMemoryTests:
             self.assertIs(wr(), None)
             self.assertIs(L[0], b)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_reversed(self):
         for tp in self._types:
             b = tp(self._source)
@@ -389,8 +381,6 @@ class AbstractMemoryTests:
             self.assertEqual(list(reversed(m)), aslist)
             self.assertEqual(list(reversed(m)), list(m[::-1]))
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_toreadonly(self):
         for tp in self._types:
             b = tp(self._source)
@@ -526,7 +516,6 @@ class BytesMemorySliceTest(unittest.TestCase,
     BaseMemorySliceTests, BaseBytesMemoryTests):
     pass
 
-@unittest.skip("TODO: RUSTPYTHON")
 class ArrayMemorySliceTest(unittest.TestCase,
     BaseMemorySliceTests, BaseArrayMemoryTests):
     pass
@@ -535,7 +524,6 @@ class BytesMemorySliceSliceTest(unittest.TestCase,
     BaseMemorySliceSliceTests, BaseBytesMemoryTests):
     pass
 
-@unittest.skip("TODO: RUSTPYTHON")
 class ArrayMemorySliceSliceTest(unittest.TestCase,
     BaseMemorySliceSliceTests, BaseArrayMemoryTests):
     pass
@@ -561,6 +549,8 @@ class OtherTest(unittest.TestCase):
                 m[2:] = memoryview(p6).cast(format)[2:]
                 self.assertEqual(d.value, 0.6)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_memoryview_hex(self):
         # Issue #9951: memoryview.hex() segfaults with non-contiguous buffers.
         x = b'0' * 200000
--- a/Lib/test/test_memoryview.py
+++ b/Lib/test/test_memoryview.py
@@ -71,8 +71,6 @@ class AbstractMemoryTests:
         m = None
         self.assertEqual(sys.getrefcount(b), oldrefcount)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_setitem_writable(self):
         if not self.rw_type:
             self.skipTest("no writable type to test")
@@ -114,11 +112,13 @@ class AbstractMemoryTests:
         self.assertRaises(TypeError, setitem, "a", b"a")
         # Not implemented: multidimensional slices
         slices = (slice(0,1,1), slice(0,1,2))
-        self.assertRaises(NotImplementedError, setitem, slices, b"a")
+        # TODO: RUSTPYTHON
+        # self.assertRaises(NotImplementedError, setitem, slices, b"a")
         # Trying to resize the memory object
         exc = ValueError if m.format == 'c' else TypeError
-        self.assertRaises(exc, setitem, 0, b"")
-        self.assertRaises(exc, setitem, 0, b"ab")
+        # TODO: RUSTPYTHON
+        # self.assertRaises(exc, setitem, 0, b"")
+        # self.assertRaises(exc, setitem, 0, b"ab")
         self.assertRaises(ValueError, setitem, slice(1,1), b"a")
         self.assertRaises(ValueError, setitem, slice(0,2), b"a")
 
@@ -272,11 +272,12 @@ class AbstractMemoryTests:
         with check: m.itemsize
         with check: m.ndim
         with check: m.readonly
-        with check: m.shape
-        with check: m.strides
-        with check:
-            with m:
-                pass
+        # TODO: RUSTPYTHON
+        # with check: m.shape
+        # with check: m.strides
+        # with check:
+        #     with m:
+        #         pass
         # str() and repr() still function
         self.assertIn("released memory", str(m))
         self.assertIn("released memory", repr(m))
@@ -298,8 +299,6 @@ class AbstractMemoryTests:
             with m:
                 m.release()
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_release(self):
         for tp in self._types:
             b = tp(self._source)
@@ -435,11 +434,9 @@ class BaseArrayMemoryTests(AbstractMemoryTests):
     itemsize = array.array('i').itemsize
     format = 'i'
 
-    @unittest.skip('XXX test should be adapted for non-byte buffers')
     def test_getbuffer(self):
         pass
 
-    @unittest.skip('XXX NotImplementedError: tolist() only supports byte views')
     def test_tolist(self):
         pass
 
@@ -503,7 +500,6 @@ class BytesMemoryviewTest(unittest.TestCase,
 class ArrayMemoryviewTest(unittest.TestCase,
     BaseMemoryviewTests, BaseArrayMemoryTests):
 
-    @unittest.skip("TODO: RUSTPYTHON")
     def test_array_assign(self):
         # Issue #4569: segfault when mutating a memoryview with itemsize != 1
         a = array.array('i', range(10))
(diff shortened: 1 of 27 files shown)


youknowone:
Slots is a data field of type (PyClass). For now, we typically iterate mro to find proper field of slots.

vm/src/obj/objmemory.rs
```diff
@@ -1,74 +1,212 @@
...
     ItemProtocol, PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
 };
-use crate::slots::Hashable;
-use crate::stdlib::array::PyArray;
+use crate::slots::{AsBuffer, Hashable};
 use crate::VirtualMachine;
+use crate::{bytesinner::try_as_bytes, pyobject::IntoPyObject};
+use crossbeam_utils::atomic::AtomicCell;
+
+pub trait BufferProtocol: Debug + Sync + Send {
```
coolreader18, line 16: Instead of `Send + Sync` here, you probably want to use `PyThreadingConstraint` from `pyobject`

youknowone:
related to #2125 and #2195, right?

qingshi163:
#2195 I had done it, I will push out when I passed the memoryview unittest.
#2125 I don't think it works with buffer, we have to do something inside objbytes.

youknowone:
@qingshi163 I attached a commit to change Box to plain fn https://github.com/RustPython/RustPython/pull/2226/commits/c0ecc35ea5b76fe37edcb0e8d5d170ef1c759a17

vm/src/slots.rs
```diff
@@ -376,3 +381,14 @@ impl PyComparisonOp {
         }
     }
 }
+
+#[pyimpl]
+pub trait Bufferable: PyValue {
```
youknowone, line 386: I think the original name `AsBuffer` you used was better. "hashable object" is a term in python(or cpython?) but bufferable is not.

Though "comparable object" is also nothing in python. @coolreader18 any idea about naming?

coolreader18: I think AsBuffer is good, "buffer" isn't a verb so it doesn't really make sense as `Verbable` and there's precedence for `AsNoun` as a trait name in Rust, e.g. `AsRawFd`, `AsRef`

qingshi163: AsBuffer is good, but the trait function name I changed it to get_buffer as same as CPython, also because it did return a new BufferProtocol. so we should go back to AsBuffer or something like GetBuffer? what we perfer?

vm/src/obj/objmemory.rs
```diff
@@ -1,74 +1,541 @@
...
+    }
+
+    fn as_contiguous_mut(&self) -> Option<BorrowedValueMut<[u8]>> {
+        let options = self.get_options();
+        if !options.contiguous {
+            return None;
+        }
+        Some(self.obj_bytes_mut())
+    }
+
```
qingshi163, line 47: the consumer should call as_contiguous if they want just a plain buffer. Should we also need to add a function here for consuming a non-contiguous buffer like a sliced memoryview? Like a iter() return the unpacked object?

vm/src/obj/objmemory.rs
```diff
@@ -1,74 +1,541 @@
...
+use num_traits::{One, Signed, ToPrimitive, Zero};
+use rustpython_common::borrow::{BorrowedValue, BorrowedValueMut};
+
+pub trait BufferProtocol: Debug + PyThreadingConstraint {
+    // TODO: return reference to avoid copy
+    fn get_options(&self) -> BufferOptions;
+    fn obj_bytes(&self) -> BorrowedValue<[u8]>;
+    fn obj_bytes_mut(&self) -> BorrowedValueMut<[u8]>;
+    fn release(&self);
+    fn is_resizable(&self) -> bool;
```
qingshi163, line 30: I renamed as_bytes to obj_bytes because it should always return the full memory range for the original object. as_bytes may easier to be miss using.

vm/src/stdlib/pystruct.rs
```diff
@@ -56,11 +56,12 @@ mod _struct {
...
+            // XXX: size of l L q Q is platform depended?
             match self.code {
                 'x' | 'c' | 'b' | 'B' | '?' | 's' | 'p' => 1,
                 'h' | 'H' => 2,
-                'i' | 'l' | 'I' | 'L' | 'f' => 4,
-                'q' | 'Q' | 'd' => 8,
+                'i' | 'I' | 'f' => 4,
+                'l' | 'L' | 'q' | 'Q' | 'd' => 8,
                 'n' | 'N' | 'P' => std::mem::size_of::<usize>(),
                 c => {
```
qingshi163, line 66: I have to change the unit size for 'l' and 'L' to 8, because that is how array.array doing. I don't know is any compatible problem, but in my machine CPython 'l' and 'L' is 8 bytes.

youknowone: isn't it architecture dependent value?

qingshi163:
```
    /* Integers */
    case 'h':
        intsize = sizeof(short);
        is_signed = 1;
        break;
    case 'H':
        intsize = sizeof(short);
        is_signed = 0;
        break;
    case 'i':
        intsize = sizeof(int);
        is_signed = 1;
        break;
    case 'I':
        intsize = sizeof(int);
        is_signed = 0;
        break;
    case 'l':
        intsize = sizeof(long);
        is_signed = 1;
        break;
    case 'L':
        intsize = sizeof(long);
        is_signed = 0;
        break;
    case 'q':
        intsize = sizeof(long long);
        is_signed = 1;
        break;
    case 'Q':
        intsize = sizeof(long long);
        is_signed = 0;
        break;
```
This is from CPython, should we put it somewhere together? because we have to make sure array.array and struct have the same size for the type.

youknowone: For common platforms, using std::mem::size_of::<isize> or cfg target_pointer_width work. It fits for common platforms but not perfect. using libc always will fit as it defined in CPython. Try `std::mem::size_of::<libc::c_int>`

vm/src/slots.rs
coolreader18: I personally think we can keep it as `AsBuffer`

vm/src/stdlib/pystruct.rs
qingshi163: I think we should do it in another PR and try unify the reflect from format code to size and the type.

common/src/borrow.rs
```diff
@@ -18,6 +18,31 @@ pub enum BorrowedValue<'a, T: ?Sized> {
...
+impl<'a, T: ?Sized> BorrowedValue<'a, T> {
+    pub fn map<U: ?Sized, F>(s: Self, f: F) -> BorrowedValue<'a, U>
+    where
+        F: FnOnce(&T) -> &U,
+    {
+        match s {
+            Self::Ref(r) => BorrowedValue::Ref(f(r)),
+            Self::MuLock(m) => BorrowedValue::MappedMuLock(PyMutexGuard::map(m, |x| unsafe {
+                #[allow(mutable_transmutes, clippy::transmute_ptr_to_ptr)]
+                std::mem::transmute(f(x))
```
coolreader18, line 30: This isn't sound -- somebody could match to get the mutex guard out of the enum, and then treat it as mutable. I ran into the same problem, and I'm not sure exactly what the solution is. Maybe try something in cell? ImmutableMappedMutexGuard or something?

coolreader18: I could try making that later today if you don't feel confident about it; I think it would be like `{ data: *const T, raw: &'a parking_lot::RawMutex, _marker: PhantomData<(&'a T, RawMutex::Guard market)> }`

qingshi163: If you can fix that will be perfect, I have no clue to get it soundless..

vm/src/slots.rs
youknowone: That sounds like the implementation trait name can be `Buffer` and the desriptor trait name can be `BufferProtocol`. Like,
- AsBuffer(Bufferable) -> BufferProtocol
- BufferProtocol -> Buffer

coolreader18: That makes sense, since a trait is inherently a "protocol", so there's no need to specify that stuff that a buffer can do is a protocol; it's redundant

common/src/borrow.rs
coolreader18: #2239

vm/src/slots.rs
qingshi163: That sounds good!

qingshi163:
But the obj and the buffer in the memoryview could be different, the obj should always point to the original data source and the buffer is the controller shows how can we visit the data. That is how I implement it as the memoryview build from a memoryview the obj is cloned from original object.

qingshi163:
I want to know if it is ok to merge, it is not fully completed but works.

coolreader18:
Oh, I suppose that's true. Nevermind

Lib/test/test_memoryview.py
```diff
@@ -114,11 +112,13 @@ def setitem(key, value):
         self.assertRaises(TypeError, setitem, "a", b"a")
         # Not implemented: multidimensional slices
         slices = (slice(0,1,1), slice(0,1,2))
-        self.assertRaises(NotImplementedError, setitem, slices, b"a")
+        # TODO: RUSTPYTHON
```
youknowone, line 115: I know this PR enabled a lot of parts from this test, but unless it is fully resolved, I prefer to keep the whole test as expectedFailure before merge. this is only trackable by code, but expectedFailure is trackable by test result. Once the feature is fully done, it will be detected if it is expectedFailiure but not if it is commented out.

extra_tests/snippets/memoryview.py
```diff
@@ -6,7 +6,7 @@
 a = memoryview(obj)
 assert a.obj == obj
 
-assert a[2:3] == b"c"
+# assert a[2:3] == b"c"
```
youknowone, line 9: is this intended?

vm/src/bytesinner.rs
```diff
@@ -263,6 +263,8 @@ impl PyBytesInner {
         op: PyComparisonOp,
         vm: &VirtualMachine,
     ) -> PyComparisonValue {
+        // TODO: bytes can compare with any object implemented buffer protocol
+        // but not memoryview, and not equal if compare with unicode str(PyStr)
         PyComparisonValue::from_option(
             try_bytes_like(vm, other, |other| {
                 op.eval_ord(self.elements.as_slice().cmp(other))
```
qingshi163, line 270: Now we have different behavour between (memoryview op bytes) and (bytes op memoryview), Is any solution better than check the type here?

common/src/borrow.rs
qingshi163: @coolreader18 will you merge #2239 before or after the pr?

vm/src/bytesinner.rs
```diff
@@ -38,17 +41,14 @@ impl From<Vec<u8>> for PyBytesInner {
...
-            }),
-            k @ PyMemoryView => Ok(PyBytesInner {
-                elements: k.try_bytes(|v| v.to_vec()).unwrap()
-            }),
+            // TODO: generic way from &[PyObjectRef]
             l @ PyList => l.to_byte_inner(vm),
+            t @ PyTuple => t.to_bytes_inner(vm),
             obj => {
                 let iter = vm.get_method_or_type_error(obj.clone(), "__iter__", || {
                     format!("a bytes-like object is required, not {}", obj.class())
```
qingshi163, line 54: I think we could have a trait for the objects that can be borrowed as &[PyObjectRef], so we can have a generic and efficiency way to iter.

vm/src/bytesinner.rs
coolreader18: Does memory view not implement the buffer protocol? I think that would maybe fix the issue(?)

qingshi163: But memoryview is implemented buffer protocol, we will have much more issue if not so.

coolreader18: Huh, that's strange, I think that's probably fine for now.

qingshi163:
What that random ci fail comes from?

coolreader18:
@qingshi163 I rebased this to master with the new mutex/`BorrowedValue::map` implementations; it put me as a committer for all the commits but hopefully you can just rebase it yourself and it would get rid of that.

qingshi163:
@coolreader18 Thanks, I think I need review for merge now.

coolreader18 (approved):
I think this overall looks really good!! I'm fine with merging it and fixing leftover issues later, since this is so big already. @youknowone what do you think?

coolreader18:
And don't worry about that CI failure; I can fix that when I merge.

youknowone:
oops, after a few days of resting, I totally forgot to merge this first before other PRs. I agree to merge this and fix other stuffs later. Let me fix the conflicts.

youknowone:
I fixed a few commit messages because fisrt two 'fix test' commits are not fixing test but disabling it.
I also reordered commits a bit to merge "update memoryview eq"s to be single commit and also "Implement buffer protocol" and "Rename to AsBuffer"

PR #2227, opened 2020-09-21
Rework native function creation a bit to give more functions __name__s

coolreader18:


vm/src/pyobject.rs
```diff
@@ -284,46 +283,34 @@ impl PyContext {
...
-        PyObject::new(
-            PyBuiltinFunction::from(f.into_func()),
-            self.types.builtin_function_or_method_type.clone(),
-            None,
-        )
+        PyFuncDef::from(f.into_func()).build_function(self)
     }
 
-    pub fn new_function_named<F, FKind>(&self, f: F, module: String, name: String) -> PyObjectRef
+    pub(crate) fn new_stringref(&self, s: String) -> objstr::PyStringRef {
```
youknowone, line 289: I think new_str can return PyStringRef and PyObjectRef can be returned by vm.new_pyobj

youknowone (approved):

vm/src/obj/objbuiltinfunc.rs
```diff
@@ -1,17 +1,64 @@
...
-use crate::pyobject::{PyClassImpl, PyContext, PyObjectRef, PyResult, PyValue, TypeProtocol};
+use crate::pyobject::{
+    PyClassImpl, PyContext, PyObject, PyObjectRef, PyResult, PyValue, TypeProtocol,
+};
 use crate::slots::{SlotCall, SlotDescriptor};
 use crate::vm::VirtualMachine;
 
+pub struct PyFuncDef {
+    pub func: PyNativeFunc,
+    pub name: Option<PyStringRef>,
```
youknowone, line 15: nice work, I wish we can see this type to be not an option soon

vm/src/pyobject.rs
coolreader18: I think I'll do that separately; there's a lot of code right now that assumes that new_str will return a PyObjectRef so it'd be better to change it in another PR

PR #2228, opened 2020-09-21
Move BorrowValue to rustpython-common, add BorrowedValue enum

coolreader18:
I think BorrowedValue would be really useful for #2226, for `BufferProtocol::as_bytes()` to return

I've had this for a while but the buffer protocol PR made me think this would be useful to merge now


common/src/borrow.rs
```diff
@@ -0,0 +1,66 @@
...
+};
+use std::ops::{Deref, DerefMut};
+
+pub trait BorrowValue<'a> {
+    type Borrowed: 'a + Deref;
+    fn borrow_value(&'a self) -> Self::Borrowed;
+}
+
+#[derive(Debug, derive_more::From)]
+pub enum BorrowedValue<'a, T: ?Sized> {
```
youknowone, line 13: is this better to be an enum rather than concrete type or boxed dyn?
I thought buffer protocol help to avoid this situation.

youknowone:
@qingshi163 will you please review this PR?

common/src/borrow.rs
coolreader18: I could probably add a Dyn variant that's just `Box<dyn Deref>` if eventually needed, but as is these are the most common forms of borrowing, and they all just have the underlying borrow stored as a reference/pointer, so no function calls are necessary to Deref it (unlike a dyn Deref)

qingshi163 (approved):
Looks nice!

common/src/borrow.rs
youknowone: Maybe I don't know rust pratice well, so this is trying discussion.
When I am seeing BorrowValue for specific type like `PyBytesLike`, it doesn't use all the enum fields but its borrowed type is BorrowedValue. I think this is giving unnessassary cost.
using enum for each case is ok - because it is what it need to be. But BorrowedValue itself as enum is hiding somewhat not expected to be hided.
my suggestion is, defining specific type for each borrowed types. `BorrowedBytesLike` or `BorrowedRwBytes`. And BorrowedValue itself as a trait (if it is needed). We can find better abstraction or we can use `dyn BorrowedValue` in future complicated scenarios.

youknowone:
I just remembered @skinny121 contributed to PyBytesLike a lot

common/src/borrow.rs
skinny121: I think this is good because with the buffer protocol `PyBytesLike` and `PyRwBytesLike` will become extensible and having a flexible return type from `borrow_value` will make it easier for implementer's of the buffer protocol. I don't this will actually have much cost as all variants probably will optimize down to a simply deref anyways.

skinny121 (approved):

youknowone (approved):

PR #2230, opened 2020-09-22
rustpython notebook, add markdown and math support

mireille-raad:
This PR is to add features onto the rustpython notebook.

You can still run pure python code there, but now there is the **option** to divide your content and add markdown and math blocks. There is a sample on `sample.txt` in this repo.

Screenshot from 2020-09-21 22-19-10

Here is a math block and how it renders.

Screenshot from 2020-09-21 22-28-02

Here is a sample python code ;p 

Screenshot from 2020-09-21 23-04-04

## Parsing
Code blocks are marked by `%%` + `keyword`. 

I re-used the parser code from iodide. It is on [this link](https://github.com/iodide-project/iodide/blob/master/src/editor/iomd-tools/iomd-parser.js).

The parser takes the content of the code editor and can look for any keyword. It returns an object with the different chunks, each chunk has these properties.

```json
{
  "chunkContent": "\nprint(\"Javascript called. It wants its DOM elements back.\")\n",
  "chunkType": "py",
  "chunkId": "-1267735422_0",
  "evalFlags": [],
  "startLine": 1,
  "endLine": 4
}
```

## Rendering

After I have the chunks, I added logic for rendering the md, py, math and math-inline chunkType.

- For markdown, I used MarkedJs (https://marked.js.org) - https://github.com/markedjs/marked
- For math, I used KaTex (https://katex.org/) - https://github.com/KaTeX/KaTeX

I tried to go with the smallest and fastest. I also chose things that have basic features but can be customized.

I didn't do the inline math rendering of iodide, because it felt that there is an endless amount of looping. You loop first through the code to get the code chunks, then you loop again to look for separators for math + markdown mixed together.

**Other slightly important code update:**

- renamed the main div,  `<div id="console">` to `<div id="rp-notebook">`. It stands for rust python notebook. console as a name sounded weird.
- add file-loader to webpack, so i can load the math related fonts

**Question**
I had one undecided question: is it worth it to add real time markdown rendering? does this open the door for having to add real time math rendering? this would means I have to re-parse the notebook on each key down and re-run the md blocks.

### DIFF
--- a/wasm/notebook/src/sample.txt
+++ b/wasm/notebook/src/sample.txt
@@ -1,0 +1,68 @@
+%%py 
+
+print("Javascript called. It wants its DOM elements back.")
+
+%%md 
+
+## RustPython Notebook with Markdown and Math
+
+Look! I can write markdown in the notebook. It messes up code highligting, but at least it works.
+
+1. i can make a list
+1. i can haz many lines and numbers
+1. I can paste emojis  🐍 😱 🤘
+
+In the notebook, you can write markdown, math and python. use %% then: 
+- md for markdown
+- py for python
+- math for large, centered math blocks
+- math-inline for an inline math block
+
+
+I **can** bold things and _italicize_ them. You got the gist, this is standard markdown syntax that you can [google](google.com).
+
+<br>
+
+## Math
+
+Supported TeX functions is [here](https://katex.org/docs/supported.html). For example: 
+
+%%math 
+
+\tau(u_f)*l_f
+
+ \newline
+
+H(t) \xrightarrow{write} \Big[A(t+1),\ H(t+1)\Big]
+
+\newline 
+
+
+\newline
+
+% \f is defined as #1f(#2) using the macro
+\f\relax{x} = \int_{-\infty}^\infty
+    \f\hat\xi,e^{2 \pi i \xi x}
+    \,d\xi
+
+\newline
+\begin{Bmatrix}
+   a & b \\
+   c & d
+\end{Bmatrix}
+
+\dbinom{n}{k}
+
+\sqrt[3]{x}
+
+
+%%py
+
+print("RustPython called back and said people need to have")
+
+# You can run regular python code here.
+# Here is a random number generator.
+import random 
+rnd = random.randint(1,5)
+
+print("nice " * rnd + "things.")
--- a/wasm/notebook/webpack.config.js
+++ b/wasm/notebook/webpack.config.js
@@ -31,22 +31,13 @@ module.exports = (env = {}) => {
                     use: [MiniCssExtractPlugin.loader, 'css-loader']
                 },
                 {
-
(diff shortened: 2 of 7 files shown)


paulromer149 (approved):
This looks great. 

Re rendering the markdown, I think it is most important for now to make sure that performance is ok. There will surely be lots of things that need to be added that will slow things down. So I'd avoid adding anything inessential now that would impose a performance hit, even a small one. We can always come back and add it later.

youknowone (approved):

wasm/notebook/webpack.config.js
```diff
@@ -29,7 +29,23 @@ module.exports = (env = {}) => {
                 {
                     test: /\.css$/,
                     use: [MiniCssExtractPlugin.loader, 'css-loader']
-                }
+                },
+                {
+
+                    test: /\.(woff(2)?|ttf|eot|svg)(\?v=\d+\.\d+\.\d+)?$/,
```
coolreader18, line 35: Does this do anything? It doesn't look like you're importing any fonts, just the ones from google fonts

wasm/notebook/src/sample.txt
```diff
@@ -0,0 +1,68 @@
+%%py 
+
+print("Javascript called. It wants its DOM elements back.")
+
+%%md 
+
+## RustPython Notebook with Markdown and Math
```
coolreader18, line 7: Do you want to have this as the default code? You could do something similar to the demo with the `defaultSnippet` in the html config

coolreader18 (approved):
Looks great!

wasm/notebook/src/sample.txt
mireille-raad: I will do a different example with default things. I want to add Javascript and css support first. So someone can write js to create like a mini UI. Whatever you select in that UI can then be passed to the python code to do the calculations or modeling. 
I know it is possible because you mentioned vars and js_vars in the previous demo. So, I kinda know how to do that. 
I will wait till I am done with that and then do the sample, I will probably have it be less silly and more boring with the demo all types of mark down elements.

wasm/notebook/webpack.config.js
mireille-raad: Will check and update

mireille-raad: updated to 
```
{
                    test: /\.(woff(2)?|ttf)$/,
                    use: {
                        loader:"file-loader",
                        options: { name: "fonts/[name].[ext]" }
                    },
                }
```

I need this to import fonts for katex. The files are `.woff`, `.woff2` and `.ttf` and go under `/fonts`

Screenshot from 2020-09-26 11-09-43

Is there a better way to do this in the webpack config?

I didn't know katex used 60 fonts for math... :open_mouth:  should I look into why and force maybe a smaller amount? (since this is an example, it might be ok -- but happy to look into it).

PR #2231, opened 2020-09-23
slot optimization 2

youknowone:
more work based on #2223

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -341,13 +341,6 @@ where
                 quote! {
                     slots.#slot_ident = Some(#into_func);
                 }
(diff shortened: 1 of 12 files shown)


coolreader18:
Oh, if you want to, you might want to try making tp_getattro a slot; it probably isn't amazing for performance that it does a method lookup/invoke()/FromArgs on every attribute lookup

youknowone:
nice idea. but i feel like i am going to be sick of slot works for now. maybe making an issue for slots to manage priority would be good. there are a lot of slots but there must be more important things than others like tp_getattro

coolreader18 (approved):
LGTM

PR #2233, opened 2020-09-24
_io uses pymodule

youknowone:


coolreader18:
Is there a reason that the StringIORef impl can't just be on StringIO?

youknowone:
I didnt check possible or not. I decided to keep it as it is to size down the diff. My goal is removing vm.class()

coolreader18 (approved):
LGTM

PR #2234, opened 2020-09-25
Make dict repr more efficient (50% faster!)

coolreader18:
```
❯ hyperfine -L X norm,opt './rustpython-{X} pystone.py 50000'    
Benchmark #1: ./rustpython-norm pystone.py 50000
  Time (mean ± σ):     14.349 s ±  0.561 s    [User: 14.295 s, System: 0.010 s]
  Range (min … max):   13.580 s … 14.982 s    10 runs
 
Benchmark #2: ./rustpython-opt pystone.py 50000
  Time (mean ± σ):     9.385 s ±  0.193 s    [User: 9.308 s, System: 0.016 s]
  Range (min … max):   8.986 s … 9.623 s    10 runs
 
Summary
  './rustpython-opt pystone.py 50000' ran
    1.53 ± 0.06 times faster than './rustpython-norm pystone.py 50000'
```
Based on this python version of the data structure (ran through lib2to3): https://code.activestate.com/recipes/578375/

I chose pystone to benchmark this because there aren't many stdlib or even builtins calls, it's just loops, `Load`/`StoreName`s, and arithmetic, so this is just the speedups for module/scope lookups, I think.

### DIFF
--- a/vm/src/obj/objdict.rs
+++ b/vm/src/obj/objdict.rs
@@ -299,14 +299,8 @@ impl PyDict {
         default: OptionalArg<PyObjectRef>,
         vm: &VirtualMachine,
     ) -> PyResult {
-        match self.entries.get(vm, &key)? {
-            Some(value) => Ok(value),
-            None => {
-                let set_value = default.unwrap_or_none(vm);
-                self.entries.insert(vm, key, set_value.clone())?;
-                Ok(set_value)
-            }
-        }
+        self.entries
+            .setdefault(vm, key, || default.unwrap_or_none(vm))
     }
 
     #[pymethod]
--- a/vm/src/obj/objdict.rs
+++ b/vm/src/obj/objdict.rs
@@ -376,7 +376,7 @@ impl PyDict {
 
     #[pymethod]
     fn popitem(&self, vm: &VirtualMachine) -> PyResult {
-        if let Some((key, value)) = self.entries.pop_front() {
+        if let Some((key, value)) = self.entries.pop_back() {
             Ok(vm.ctx.new_tuple(vec![key, value]))
         } else {
             let err_msg = vm.ctx.new_str("popitem(): dictionary is empty");
--- a/Lib/test/test_dict.py
+++ b/Lib/test/test_dict.py
@@ -368,8 +368,6 @@ class DictTest(unittest.TestCase):
         x.fail = True
         self.assertRaises(Exc, d.setdefault, x, [])
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_setdefault_atomic(self):
         # Issue #13521: setdefault() calls __hash__ and __eq__ only once.
         class Hashed(object):
--- a/common/src/cell.rs
+++ b/common/src/cell.rs
@@ -119,7 +119,31 @@ cfg_if::cfg_if! {
         type RwLockInner<T> = RefCell<T>;
         type RwLockReadInner<'a, T> = Ref<'a, T>;
         type MappedRwLockReadInner<'a, T> = Ref<'a, T>;
-        type RwLockWriteInner<'a, T> = RefMut<'a, T>;
+        #[derive(Debug)]
+        struct RwLockWriteInner<'a, T: ?Sized>(&'a RefCell<T>, RefMut<'a, T>);
+        impl<'a, T: ?Sized> RwLockWriteInner<'a, T> {
+            fn map<U: ?Sized, F>(s: Self, f: F) -> MappedRwLockWriteInner<'a, U>
+            where
+                F: FnOnce(&mut T) -> &mut U,
+            {
+                RefMut::map(s.1, f)
+            }
+            fn downgrade(s: Self) -> RwLockReadInner<'a, T> {
+                drop(s.1);
+                read_rwlock(s.0)
+            }
+        }
+        impl<'a, T: ?Sized> Deref for RwLockWriteInner<'a, T> {
+            type Target = T;
+            fn deref(&self) -> &T {
+                self.1.deref()
+            }
+        }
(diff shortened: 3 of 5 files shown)


youknowone:
it seems like we are reinventing hashmap

youknowone:
I think we are using plain hashmap for `__dict__`. how improving `dict` can be related to module/scope lookup?

coolreader18:
Which `__dict__`? For classes we're using a `HashMap<String, PyObjectRef>`, but in `crate::scope::Scope` it's a PyDictRef for globals and locals, and modules use `PyObject.dict: PyDictRef` as well

coolreader18:
And `obj.__dict__` is a `dict` as well, of course

coolreader18:
> it seems like we are reinventing hashmap

Yeah, sort of. I looked into repurposing hashbrown, but it's fundamentally designed for a different kind of use than python dicts; it expects all hash operations to be infallible or it may drop entries, the simd stuff makes it more complex, and it's got a lot of unsafe code that the simpler implementation doesn't, which makes it harder to hack on since you always have to be worrying about safety.

youknowone:
I got the other points but what about hash infallible? What kind of case be a problem in hashbrown but not in python?

vm/src/obj/objdict.rs
```diff
@@ -376,7 +370,7 @@ impl PyDict {
 
     #[pymethod]
     fn popitem(&self, vm: &VirtualMachine) -> PyResult {
-        if let Some((key, value)) = self.entries.pop_front() {
+        if let Some((key, value)) = self.entries.pop_back() {
```
youknowone, line 373: is this intended change?

youknowone (approved):
good as dict improvement, not sure for pystone benchmark. it doesn't benchmark dict, so it can be optimized in different way, maybe even better.

vm/src/obj/objdict.rs
coolreader18: Yeah, it was a lot easier to have it pop from the back rather than the front, and it's actually supposed to be like that for dict (though not for set):

```py
>>> d = {}
>>> d[1] = 2
>>> d[2] = 3
>>> d[3] = 4
>>> d.popitem()
(3, 4)
>>> d.popitem()
(2, 3)
>>> d.popitem()
(1, 2)
>>> d.popitem()
KeyError: 'popitem(): dictionary is empty'
>>> s = set()
>>> s.add(1)
>>> s.add(2)
>>> s.add(3)
>>> s.pop()
1
>>> s.pop()
2
>>> s.pop()
3
>>> s.pop()
KeyError: 'pop from an empty set'
```
Hopefully it's an ok tradeoff for now, since dict having a correct behavior would probably be more depended on then set, and it didn't trip (or even fix) any tests

coolreader18:
> I got the other points but what about hash infallible? What kind of case be a problem in hashbrown but not in python?

Specifically [here](https://docs.rs/hashbrown/0.9.0/src/hashbrown/raw/mod.rs.html#787-802) in hashbrown; when it rearranges/compacts the storage, it needs to know their hashes and if it can't get one that it sorta throws its hands up and deletes what it hasn't touched yet (I think, again, the code is complex). There's a few other places where it does that as well, you can grep for a call to `guard`

vm/src/obj/objdict.rs
youknowone: interesting, I didn't know there was any difference about this

youknowone:
I thought we would replace only hashmap to RawTable if we use hashbrown same as our previous design just like this PR re placing it into a vec. In that case only ints are placed in hashbrown bucket and then the values will be always hashable.

coolreader18:
Even with RawTable, there are still a few functions that take a hasher `Fn(&T) -> u64`, and those are the functions that are expected to be infallible; even if they were `Fn(&T) -> Result<u64, E>` there's no way to recover from an `Err` with the layout/strategy/whatever hashbrown uses

youknowone:
I still don't get what's the problem there. when the bucket of RawTable is `(isize, isize)` T in hasher will be also `isize`. It looks like infallible. Will you tell me more details of your design when you thought about hashbrown?

coolreader18:
Wait, you're saying just to make `indices` a RawTable? Cause a `RawTable<(isize, isize)>` shouldn't be any different than `HashMap<(isize, isize)>`; that's what HashMap is inside:
```rust
pub struct HashMap<K, V, S = DefaultHashBuilder> {
    pub(crate) hash_builder: S,
    pub(crate) table: RawTable<(K, V)>,
}
```

coolreader18:
And RawTable still has a vector of elements inside, but if we already know where to find the indices for a hash with perturb and stuff then it's more efficient to just have the `Vec<EntryIndex>` than all the machinery of a hashmap

coolreader18:
Cause barring hash collisions (or masked hash collisions?) it will basically be O(1), just arithmetic to the hashvalue and then index into `indices`, whereas hashbrown's gotta do a whole separate lookup which is still O(1), but just extra/unnecessary stuff when we could do it with just the index

coolreader18:
Idk, I'm not trying to be aggressive about this or anything, I'm just sorta working out why this works better for myself as well cause it does seem like a HashMap might be more efficient than a sparse array, but it turns out it's not

youknowone:
Thank you for explanation.

> Idk, I'm not trying to be aggressive about this or anything, I'm just sorta working out why this works better for myself as well cause it does seem like a HashMap might be more efficient than a sparse array, but it turns out it's not

For this, my view was the opposite. Because RawTable is just a wrapper for dynamic array, there will be not much cost comparing to custom vec. If we use custom hasher wiithout actual hashing, I expected the cost wiil be very similar to the custom vec. But simd optimization would be the free lunch in this case. I wonder why there is the difference.

PR #2235, opened 2020-09-25
getattro slot

coolreader18:
|:---|---:|---:|---:|---:|
| `./rustpython-norm pystone.py 50000` | 9.068 ± 0.422 | 8.567 | 10.075 | 1.14 ± 0.06 |
| `./rustpython-opt pystone.py 50000` | 7.934 ± 0.155 | 7.724 | 8.179 | 1.00 |

### DIFF
--- a/vm/src/slots.rs
+++ b/vm/src/slots.rs
@@ -3,6 +3,7 @@ use std::cmp::Ordering;
 use crate::common::cell::PyRwLock;
 use crate::common::hash::PyHash;
 use crate::function::{OptionalArg, PyFuncArgs, PyNativeFunc};
+use crate::obj::objstr::PyStrRef;
 use crate::pyobject::{
     Either, IdProtocol, PyComparisonValue, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject,
 };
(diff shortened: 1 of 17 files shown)


coolreader18:
Also taking suggestions for the name of the trait :upside_down_face: probably just `SlotGetattro`?

coolreader18:
Ok, I went with `SlotGetattro`, probably better than `Getattroble` :grin:

vm/src/slots.rs
```diff
@@ -51,6 +52,7 @@ pub(crate) type CmpFunc = fn(
     PyComparisonOp,
     &VirtualMachine,
 ) -> PyResult<Either<PyObjectRef, PyComparisonValue>>;
+pub(crate) type GetattroFunc = fn(PyObjectRef, PyStrRef, &VirtualMachine) -> PyResult;
```
youknowone, line 55: Can the perameters be ref like other slot funcs? Or maybe even &str for second one?

youknowone (approved):

vm/src/slots.rs
youknowone: Nevermind about second one. It will not  fit to string interning

coolreader18: Yeah, I'm planning on having it take a `&PyRef` but `vm.get_attribute()` takes an owned PyObjectRef, so I'll do that big diff separately

youknowone:
will you also share your pystone kit in somewhere benchmark or extra_tests?

vm/src/slots.rs
youknowone: I didn't work on descr_get because it always get PyObjectRef during get_class_attr but this change also will resolve this issue.

coolreader18:
Yeah, that's a good idea

PR #2237, opened 2020-09-26
string embedding

youknowone:


vm/src/obj/objstr.rs
```diff
@@ -30,6 +31,58 @@ use crate::slots::{Comparable, Hashable, PyComparisonOp};
 use crate::VirtualMachine;
 use rustpython_common::hash;
 
+#[cfg(target_pointer_width = "64")]
+type InnerIntRepr = u128;
+#[cfg(target_pointer_width = "32")]
+type InnerIntRepr = u64;
+
+#[derive(Debug)]
+struct StrInner(ManuallyDrop<Box<str>>);
```
coolreader18, line 40: Could put this in `rustpython_common::str`, maybe

coolreader18:
Ooh, here's an idea for embedding strings: proc macro that you pass a string literal `"foo"` to, generates code that looks like
```rust
{
    static STRING: StaticStr<[u8; len of strlit]> = StaticStr::new(*b"foo");
    let s: &StaticStr<[u8]> = &STRING;
    // SAFETY: s was generated from a utf8 string literal
    unsafe { &*(s as *const _ as *const StaticStr<str>) }
}
```

And `b"foo"` is the bytestring representation of the bytes of the string literal passed to the macro

coolreader18:
Oh, though maybe that wouldn't work, cause the hash value of str depends on the hash secret in vm

coolreader18:
Maybe we could have a global `OnceCell<HashSecret>`, and in vm init we set state.hash_secret to `*HASH_SECRET.get_or_init(|| make_secret())`, so that the whole process always has the same hash secret; pretty sure that's what CPython does anyway

youknowone:
that sounds like more to string interning for constant string, right? I thought vm can hold a vec of PyStringRef for that.

In this PR, I am trying short string embedding here. It will reuse fat pointer memory allocated for Box<str>.

coolreader18:
Ah, okay, that makes sense, I guess I misinterpreted what you meant by embedding

coolreader18:
Looking at the source for embed-str -- is there a reason you did `(1..=MAX_SIZE).contains(&s.len())` and not `s.is_empty()`?

coolreader18:
Also, when you `assume_init()` for EmbeddedStr, I don't think that's actually all initialized -- you only `copy_nonoverlapping()` len bytes, after that I think it's uninitialized. You might want to store EmbeddedStr/EmbeddingStr as a `MaybeUninit<[usize; 2]/[u8; SIZE]>` and read the mode by reading only the exact byte that actually holds the metadata. Have you tried running the tests with `cargo miri`? That would probably iron out UB

youknowone:
@Kroisse introduced https://docs.rs/string_cache/ for string interning

youknowone:
> Looking at the source for embed-str -- is there a reason you did (1..=MAX_SIZE).contains(&s.len()) and not s.is_empty()?

I think that's compact expression for `!s.is_empty() && s <= MAX_SIZE`

About mode, i'd better to use exact byte from byte array. But does it actually need to be stored as MaybeUninit? we can access only to initailized bytes and it doesn't have any drop behavior.

youknowone:
I think this PR requires benchmark before merging

coolreader18:
> I think that's compact expression for `!s.is_empty() && s <= MAX_SIZE`

Oh, lol, I completely missed the `const MAX_SIZE` before that, I thought MAX_SIZE was like `usize::MAX` or something

> But does it actually need to be stored as MaybeUninit? we can access only to initailized bytes and it doesn't have any drop behavior.

Yeah, unfortunately, it's still UB even if you're not accessing them, since you can't trust that the compiler won't optimize it to access some of the uninitialized memory -- there's a good article I read about it but I can't find it right now.

whjpji:
It seems that the `embed_str` crate does not check the invalid unicode points?

PR #2238, opened 2020-09-27
Trait for static type

youknowone:
otherwise heap type

### DIFF
--- a/common/src/static_cell.rs
+++ b/common/src/static_cell.rs
@@ -1,0 +1,76 @@
+#[cfg(not(feature = "threading"))]
+mod non_threading {
+    use crate::lock::OnceCell;
+    pub type StaticCell<T> = std::thread::LocalKey<OnceCell<&'static T>>;
+
+    #[macro_export]
+    macro_rules! static_cells {
+        // process multiple declarations
+        ($(#[$attr:meta])* $vis:vis static $name:ident: $t:ty;) => (
+            std::thread_local! {
+                $(#[$attr])* $vis static $name: $crate::lock::OnceCell<&'static $t> = $crate::lock::OnceCell::new();
+            }
+        );
+    }
+}
+#[cfg(not(feature = "threading"))]
+pub use non_threading::*;
+
+#[cfg(feature = "threading")]
+mod threading {
+    use crate::lock::OnceCell;
+
+    pub struct StaticKey<T> {
+        inner: T,
+    }
+
+    impl<T> StaticKey<T> {
+        pub const fn new(inner: T) -> Self {
+            Self { inner }
+        }
+
+        pub fn with<F, R>(&self, f: F) -> R
+        where
+            F: FnOnce(&T) -> R,
+        {
+            f(&self.inner)
+        }
+    }
+
+    pub type StaticCell<T> = StaticKey<OnceCell<&'static T>>;
+
+    #[macro_export]
+    macro_rules! static_cells {
+        // process multiple declarations
+        ($(#[$attr:meta])* $vis:vis static $name:ident: $t:ty;) => (
+            $(#[$attr])* $vis static $name: $crate::static_cell::StaticKey<$crate::lock::OnceCell<&'static $t>> = $crate::static_cell::StaticKey::new($crate::lock::OnceCell::new());
+        );
+    }
+}
+#[cfg(feature = "threading")]
+pub use threading::*;
+
+pub fn get<T>(cell: &'static StaticCell<T>) -> Option<&'static T> {
+    cell.with(|cell| cell.get().copied())
+}
+
+pub fn init_expect<T>(cell: &'static StaticCell<T>, value: T, msg: &'static str) -> &'static T {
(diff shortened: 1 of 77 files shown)


coolreader18:
Ah, that's what heap type or not means; I thought it was a type whose objects are allocated on the heap, but that doesn't make any sense cause every object is allocated on the heap

youknowone:
Static types are shared between subinterpreters and implemented as static variable in CPython. I don't think this implementation is really a static variable in C, but this PR will provide the semantic of static type.

youknowone:
Probably API's taking a type have to have signature with `&PyType` or `&PyTypeRef` rather than `PyTypeRef`. In that case, we can reduce RC cost for static types.

coolreader18:
Yeah, maybe we could make `get_id` return the ptr to the payload rather than the pyobject.

common/src/cell.rs
```diff
@@ -266,3 +266,43 @@ impl<'a, T: ?Sized> PyMappedRwLockWriteGuard<'a, T> {
...
+}
+
+// SAFETY: Inner value must be Send+Sync when threading is on; Otherwise single-thread.
+#[cfg(not(feature = "threading"))]
+unsafe impl<T> Send for StaticCellInner<T> {}
+#[cfg(not(feature = "threading"))]
+unsafe impl<T> Sync for StaticCellInner<T> {}
+
+pub struct StaticCell<T> {
+    inner: ::once_cell::sync::OnceCell<StaticCellInner<T>>,
```
coolreader18, line 281: I'll take a better look at this tomorrow, but maybe this could be like a `OnceCell<T>` when threading but a `&'static std::thread::LocalKey<T>` otherwise? I don't like the look of that `unsafe impl Send + Sync` block...

youknowone: That will be possible but sounds like to have different implementation for StaticCell for threading and non threading.

coolreader18:
How do you plan to handle multiple threads with `not(threading)`? Cause just cause `threading` is off doesn't mean people can't still use `std::thread`

youknowone:
That's true. Once it is initialized in main thread, they are read-only values. If we lock the initialization step not to allow multiple initialization in different threads, it will be safe.

coolreader18:
Are you sure? I'd think there still might be issues with non-atomic reference counting across threads; there might be a possibility that it could decref twice and free due to race conditions; if they're not atomic accesses/stores there's no guarantees that the visible effects will be coherent

youknowone:
hmm, it makes sense. Probably I mainly thought about wasm case for non-threadiing. how do we prevent that case before this change? I think std::thread to run same vm always can break non-threading mode regardless of the change.
The only difference from this PR is shared type ref, so it prevents to create and manage isolated enviroment by user. In this case, if we want to provide that kind of user guaranteed safety, I think there must be more complicated work.
Wouldn't cfg(threading) be forced to also std::thread to access vm? So, this is why you want to make it thread local?

youknowone:
not StaticCell is thread-local when `cfg(not(feature = "threading"))`

coolreader18:
I don't think you should need the unsafe to cast to a static reference, `Box::leak()` returns a `&'static mut T` that you should be able to just coerce to a `&'static T`

youknowone:
@coolreader18 any other thing to fix?

common/src/static_cell.rs
```diff
@@ -0,0 +1,76 @@
...
+
+        pub fn with<F, R>(&self, f: F) -> R
+        where
+            F: FnOnce(&T) -> R,
+        {
+            f(&self.inner)
+        }
+    }
+
+    pub type StaticCell<T> = StaticKey<OnceCell<&'static T>>;
```
coolreader18, line 40: I think you should be able to make this `OnceCell<T>` -- try modeling the non-threading version after the OnceCell API, rather than modeling the threading version off the LocalKey API, since OnceCell can provide a `&'static T` by itself, but LocalKey should still need the `Box::leak()` stuff.

youknowone: Yes, that should be possible. But this is easier by providing exact same API with `LocalKey`.
At the same time, the reference of immutable static variable is practically just same as permanently leaked reference.

So I agree that I was able to do it, but still don't have any reason to do it (yet).

Let me know if i am missing something.

coolreader18 (approved):
If you feel confident in this route towards static types, I think this is good; feel free to merge this.

vm/src/sysmodule.rs
```diff
@@ -710,7 +704,6 @@ settrace() -- set the global debug tracing function
         let getwindowsversion = WindowsVersion::make_class(ctx);
         extend_module!(vm, module, {
             "getwindowsversion" => named_function!(ctx, sys, getwindowsversion),
-            "_getwindowsversion_type" => getwindowsversion, // XXX: This is not a python spec but required by current RustPython implementation
```
youknowone, line 713: @BasixKOR I am going to remove this fake exposed type ;)

PR #2239, opened 2020-09-27
Unified lock types for rustpython_common

coolreader18:
For #2226 and specifically in order to implement `BorrowedValue::map()`; I realized that we can just make our own `RawMutex`/`RawRwLock` and use it to define the Mutex types. These should be faster, too, ~~since RefCell has more overhead (immutable accessor count mainly, `Cell<isize>` vs `Cell<bool>`)~~ since they're more direct with the use case -- e.g. just one writer vs RefCell supports multiple with `RefMut::map_split`. Turns out we can't actually make it a bool, since we subtly depend on PyRwLock having a recursive locking behavior, even though that's not guaranteed by parking_lot -- "[Note that attempts to recursively acquire a read lock on a `RwLock` when the current thread already holds one may result in a deadlock.](https://docs.rs/lock_api/0.4.1/lock_api/struct.RwLock.html#method.read)"


common/src/lock.rs
```diff
@@ -0,0 +1,117 @@
...
+cfg_if::cfg_if! {
+    if #[cfg(feature = "threading")] {
+        pub use parking_lot::{RawMutex, RawRwLock};
+
+        pub use once_cell::sync::{Lazy, OnceCell};
+    } else {
+        mod cell_lock;
+        pub use cell_lock::{RawCellMutex as RawMutex, RawCellRwLock as RawRwLock};
+
+        pub use once_cell::unsync::{Lazy, OnceCell};
```
youknowone, line 22: is this also belong to lock?

youknowone (approved):
I am not understanding this very well. Any other reviewer for this?

If possible, merging #2226 first maybe helpful. @qingshi163 experienced a lot of rebasing for that already

common/src/lock.rs
coolreader18: I think it makes sense there; `sync::OnceCell` at least has to lock when it's initializing

coolreader18: And it was there when this was `cell.rs` as well

coolreader18:
I suppose so, I'd just like to have them merged close together so that we don't run into issues with the soundness hole.

coolreader18:
I'm gonna merge this; I feel like `cell_lock` is fairly easy to maintain/understand if you just look at it in isolation.

PR #2250, opened 2020-09-30
PyFuncArgs from rust tuple

youknowone:
Now what we have:
- `FuncArgs::new` to replace `From((Args, KwArgs))`
- `vm.invoke(callable, (tuple,))`

Now what we don't have:
- FuncArgs `From((Args, KwArgs))`
- FuncArgs `From((&Args, &KwArgs))`
- Args `From(PyRef)` - this is simply replaced by single argument tuple.

### DIFF
--- a/vm/src/function.rs
+++ b/vm/src/function.rs
@@ -1,21 +1,52 @@
-use std::collections::HashMap;
-use std::marker::PhantomData;
-use std::mem;
-use std::ops::RangeInclusive;
-
-use indexmap::IndexMap;
-use result_like::impl_option_like;
-
+use self::OptionalArg::*;
 use crate::exceptions::PyBaseExceptionRef;
 use crate::obj::objtuple::PyTupleRef;
 use crate::obj::objtype::{isinstance, PyTypeRef};
 use crate::pyobject::{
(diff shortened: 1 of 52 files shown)


vm/src/function.rs
```diff
@@ -328,6 +332,34 @@ impl<T> From<Vec<T>> for Args<T> {
...
+    fn from(_args: ()) -> Self {
+        Args(Vec::new())
+    }
+}
+
+// A tuple of values that each implement `IntoPyObject` represents a sequence of
+// arguments that can be bound and passed to a built-in function.
+macro_rules! func_args_from_tuple {
+    ($(($n:tt, $T:ident)),*) => {
+        impl<$($T,)*> From<($($T,)* &VirtualMachine)> for Args<PyObjectRef>
```
coolreader18, line 345: What about just a `IntoArgs` trait? With a `fn into_args(self, vm)`; that way you wouldn't have to pass `vm` twice in a `invoke` call (as the `self` arg to invoke, and in the argument list)

youknowone: I created IntoFuncArgs trait

youknowone:
@coolreader18 will you also review this again?

coolreader18 (approved):

PR #2255, opened 2020-09-30
Document browser module availability in WASM Demo 

josalmi:
I couldn't find full documentation for the browser module anywhere where users could be pointed at but was able to scratch up some basic functionality by reading the related python and Rust code as well as using using `dir(browser)` in the RustPython demo.

Need for this came from the fact that I was running scripts that used the builtin `input()` function. Browsers don't have `stdin` that could be plugged in nicely to support `input()` command but I was able to monkey patch it in the end with:
```
import browser
input = browser.prompt
```

which worked just fine with my use cases. Just documenting the fact that the module exists can already save good amount of time from others.

### DIFF
--- a/wasm/demo/src/index.ejs
+++ b/wasm/demo/src/index.ejs
@@ -67,6 +67,13 @@
             </li>
         </ul>
 
+        <p>
+            Limited Interaction with browser is possible from Python by using
+            the <code>browser</code> module. Browser APIs such as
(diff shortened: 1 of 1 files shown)


coolreader18 (approved):
Good idea! As you said, just letting people know it exists enables discoverability. Thanks for contributing!

PR #2257, opened 2020-10-01
Don't panic on #[pyattr] use foo::bar::baz;

coolreader18:
I got this when I did:

```rust
#[pyattr]
use super::foo::BAR as BAZ;
```
So I think it makes sense not to panic here


coolreader18:
Hmm, missed that it was already merged. I only had a one line change.

youknowone (approved):
oh, i thought this case is Path->Rename but it actually is Path->Path->Rename maybe?

PR #2259, opened 2020-10-01
Optimize string iteartor 

youknowone:


vm/src/obj/objstr.rs
```diff
@@ -103,29 +103,44 @@ impl TryIntoRef<PyStr> for &str {
...
+                }
+            }
+            end.unwrap_or(start + 4)
+        };
 
-        if let Some(c) = self.string.value.chars().nth(pos) {
-            Ok(c.to_string())
-        } else {
-            Err(objiter::new_stop_iteration(vm))
+        let stored = self.position.swap(end);
```
coolreader18, line 137: You might want to use compare_exchange here

coolreader18 (approved):

PR #2260, opened 2020-10-01
Update README and powershell script to improve onboarding

coolreader18:


README.md
```diff
@@ -24,18 +24,19 @@ RustPython requires Rust latest stable version (e.g 1.43.0 at May 24th 2020).
 To check Rust version: `rustc --version` If you wish to update,
 `rustup update stable`.
 
-To test RustPython, do the following:
+To build RustPython locally, do the following:
 
     $ git clone https://github.com/RustPython/RustPython
     $ cd RustPython
-    # if you're on windows:
+      # if you're on windows:
```
youknowone, line 31: is this indent intended?

README.md
```diff
@@ -174,21 +175,6 @@ methods are often the simplest and easiest way to contribute.
...
-
-To do this, follow this method:
-
-```shell
-$ export RUSTPYTHONPATH=./Lib  # this is same as the default value
-$ cargo run -- -c 'import xdrlib'
-```
-
-You can play around with other standard libraries for python. For example, the
-[ouroboros library](https://github.com/pybee/ouroboros) or CPython Lib.
```
youknowone, line 190: I think this is still a good resource for development to add new standard library. Moving to development document?

youknowone (changes requested):

README.md
coolreader18: I don't know, I don't think there's much of a use case for this section anymore -- this has been here since before `__import__` with importlib was implemented, and all the modules we had were Rust native modules. Now, I think we're pretty settled on modified CPython Lib+some native modules, so this isn't very necessary for development anymore.

README.md
coolreader18: Yeah, I think it makes it more clear that it's not just a root prompt/differentiates it from the commands

youknowone (approved):

PR #2265, opened 2020-10-04
Implement _json.scanstring in Rust

coolreader18:


vm/src/stdlib/json.rs
```diff
@@ -3,11 +3,15 @@ mod machinery;
 
 #[pymodule]
 mod _json {
-    use crate::function::PyFuncArgs;
+    use super::*;
```
youknowone, line 6: We don't have much things in super. maybe `use super::machinery;`?

youknowone (approved):

vm/src/stdlib/json.rs
coolreader18: I suppose I could to remove the glob import, but I think it's sorta a convention to have `use super::*` at the top of an inline module, so you don't have to completely "enter" a new scope even though it's the same file.

PR #2266, opened 2020-10-05
Implement math.isqrt in Rust

ArniDagur:


Lib/test/test_math.py
```diff
@@ -957,19 +956,20 @@ def testIsqrt(self):
         self.assertIs(type(s), int)
         self.assertEqual(s, 0)
 
-        class IntegerLike(object):
-            def __init__(self, value):
-                self.value = value
+        # TODO: RUSTPYTHON
```
ArniDagur, line 959: Any idea how I can support this non-integer argument?

youknowone: It seems the function take not only int but also types with `__index__`. Search for `vm.to_index` in code. `bytesinner.rs` and `pystruct.rs` look like similar case.

ArniDagur: Thanks :)

youknowone (approved):

youknowone:
thank you for contributing!

PR #2269, opened 2020-10-06
Don't skip pathlib tests that actually succeed (on my machine)

ArniDagur:
There are a bunch of tests in `test_pathlib.py` that actually succeed when I run them on my machine. Perhaps something changed after they had been marked to be skipped? I'll let the CI be the judge of that.

### DIFF
--- a/Lib/test/test_pathlib.py
+++ b/Lib/test/test_pathlib.py
@@ -619,7 +619,6 @@ class _BasePurePathTest(object):
         self.assertRaises(ValueError, p.relative_to, '')
         self.assertRaises(ValueError, p.relative_to, P('a'))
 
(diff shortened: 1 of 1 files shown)


ArniDagur:
@youknowone When I run these tests like
```
[arni][~/src/RustPython][pathlib-tests]% python Lib/test/test_pathlib.py
..........................................s..........................................................................s..........................s.............................................................................................................................................ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss
Ran 398 tests in 0.161s

OK (skipped=115)
```
They all seem to succeed, am I doing something differently than the CI?

PR #2271, opened 2020-10-10
Add derive(Serialize, Deserialize) to FrozenModule

coolreader18:
This is so that a frozen module tree can be represented as a single bytestring, specifically for [`RustPython/pyckitup`](https://github.com/RustPython/pyckitup)


coolreader18:
Cancelled CI cause I don't want to waste 5*9 hours of electricity

youknowone (approved):

PR #2276, opened 2020-10-10
PyObjectRc::as_ref and PyLease::as_object

youknowone:


vm/src/pyobjectrc.rs
```diff
@@ -56,10 +56,15 @@ where
...
     pub fn into_ref(this: Self) -> PyObjectRc<dyn PyObjectPayload> {
         PyObjectRc::<dyn PyObjectPayload> {
-            inner: unsafe { Self::into_rc(this) }._as_ref(),
+            inner: unsafe { Self::into_rc(this) }._into_ref(),
         }
     }
 
+    #[allow(clippy::should_implement_trait)]
+    pub fn as_ref(this: &Self) -> &PyObjectRc<dyn PyObjectPayload> {
+        unsafe { &*(this as *const _ as *const _) }
```
youknowone, line 65: This casting looks safe, is it?

coolreader18: No, unfortunately; `PyObjectRc<T>` is usually pointer sized, but `PyObjectRc<dyn ...>` is double-pointer-sized; the other pointer is to the vtable for the trait. I think if we wanted to do something like this we could go the route of rc-borrow.

coolreader18: Actually, lemme try whipping something up for that, cause I've been thinking about it for a while.

youknowone: ah, I thought casting from narrow pointer to fat pointer without vtable information is not allowed but just tested &u8 even can be casted to the return type. I just realized the return type itself is narrow pointer and the pointer inside the rc is fat pointer

PR #2277, opened 2020-10-10
Implement bytes.hex() with optional sep

qingshi163:


vm/src/bytesinner.rs
```diff
@@ -24,6 +23,7 @@ use crate::{
     anystr::{self, AnyStr, AnyStrContainer, AnyStrWrapper},
     obj::objtuple::PyTuple,
 };
+use crate::{byteslike::try_bytes_like, obj::objbytes::PyBytesRef};
```
youknowone, line 26: `PyBytesRef` can be grouped with `PyBytes` in line 7

vm/src/bytesinner.rs
```diff
@@ -1235,3 +1237,95 @@ pub fn bytes_decode(
...
+pub fn bytes_to_hex(
+    bytes: &[u8],
+    sep: OptionalArg<Either<PyStrRef, PyBytesRef>>,
+    bytes_per_sep: OptionalArg<isize>,
+    vm: &VirtualMachine,
+) -> PyResult<String> {
+    if bytes.is_empty() {
+        return Ok("".to_owned());
+    }
+
```
youknowone, line 1256: I think seperating arguments processing and main code block would allow to dedent the main code and give visual seperation of the purpose for readers later.

```rust
    let s_guard;
    let b_guard;
    let (sep, bytes_per_sep) = match sep {
        OptionalArg::Present(sep) => {
            let bytes_per_sep = bytes_per_sep.unwrap_or(1);
            if bytes_per_sep == 0 {
                return Ok(hex_no_sep(bytes))
            } 
            let sep = match &sep {
                Either::A(s) => {
                    s_guard = s.borrow_value();
                    s_guard.as_bytes()
                }
                Either::B(bytes) => {
                    b_guard = bytes.borrow_value();
                    b_guard
                }
            };
            (sep, bytes_per_sep)
        }
        OptionalArg::Missing => return Ok(hex_no_sep(bytes)),
    };

    // now main code can be placed here without indent
```

vm/src/obj/objbytearray.rs
```diff
@@ -4,10 +4,10 @@ use crossbeam_utils::atomic::AtomicCell;
 use rustpython_common::borrow::{BorrowedValue, BorrowedValueMut};
 use std::mem::size_of;
 
-use super::objint::PyIntRef;
 use super::objiter;
 use super::objstr::PyStrRef;
 use super::objtype::PyTypeRef;
+use super::{objbytes::PyBytesRef, objint::PyIntRef};
```
youknowone, line 10: here also already is `PyBytes` below. I don't mind this kind of grouping when there is nothing in the file, but also want to keep the convention when other codes are already grouped by its own rule.

youknowone (changes requested):
Looks good in general.

vm/src/bytesinner.rs
qingshi163: It is rust-analyzer's auto import, forget to fix it.

youknowone: wow, i didn't know it has auto import feature

youknowone (approved):

PR #2278, opened 2020-10-11
Basic block support in JIT

evilpie:


evilpie:
I am not sure yet how well this approach will hold up, but it's a start.

The bytecode format has a number of things that make it a bit annoying to compile. 
- A `ReturnValue` instruction inside an if is still followed by a `Jump` that will never be executed
- I am not sure yet how we would even handle `JumpIfFalseOrPop`, we basically would need two different virtual stacks to emulate the maybe popping behavior.

evilpie:
Rebased. Please let me know if you need some changes before this can be merged. I don't want to fork too much.

ArniDagur:
I'm wondering: If the current bytecode format proves to be too limiting, would it not make sense to just... change it? (Easy for me to say since I wouldn't be the one  implementing those changes in the VM).

RustPython is relatively young so backwards compatibility with pre-compiled bytecode is probably not a concern. That's one thing RustPython can do but CPython can't. With that in mind, I think it would make sense to have a byte code format specifically designed to be easy to JIT compile.

Regarding JIT, perhaps some lessons can be [learned from Julia](https://www.bookstack.cn/read/julialang-1.3/5b7fcad6186f87af.md), especially in the use of type inference and optional type annotations to generate better code. We know the return values of all built-ins, so even just type inference without any other type annotations should be helpful. Furthermore, in Julia, if an `any`  (unannotated) value is bound to a type-annotated variable, a runtime type-check is made to assert that the value is compatible. Any further uses of that variable generate better code. Perhaps we could use the optional Python type annotations in a similar way? Would this require changes in the bytecode (to e.g. encode type information differently)?

evilpie:
> I'm wondering: If the current bytecode format proves to be too limiting, would it not make sense to just... change it? (Easy for me to say since I wouldn't be the one implementing those changes in the VM).
> 
I guess we will have to see. I haven't really thought about how to replace these conditional pop instructions yet. Considering how relatively simple the actual interpreter loop is, changing the bytecode might be the simplest way forward.
> RustPython is relatively young so backwards compatibility with pre-compiled bytecode is probably not a concern. That's one thing RustPython can do but CPython can't. With that in mind, I think it would make sense to have a byte code format specifically designed to be easy to JIT compile.
> 
:+1: 
> Regarding JIT, perhaps some lessons can be [learned from Julia](https://www.bookstack.cn/read/julialang-1.3/5b7fcad6186f87af.md), especially in the use of type inference and optional type annotations to generate better code. We know the return values of all built-ins, so even just type inference without any other type annotations should be helpful. Furthermore, in Julia, if an `any` (unannotated) value is bound to a type-annotated variable, a runtime type-check is made to assert that the value is compatible. Any further uses of that variable generate better code. Perhaps we could use the optional Python type annotations in a similar way? Would this require changes in the bytecode (to e.g. encode type information differently)?

As someone with many years of experience working on a JavaScript VM I am skeptical of type inference for dynamic languages. Of course the trade-offs are probably different and typical JS code is different from Python and again different from Julia as well. 
There is  a big difference between having most variables type annotated and untyped Python code. I think which one to optimize for is going to have a rather large impact.

Anyway I am probably not going to pursue any of these things deeply. I am mostly interested in experimenting with cranelift and typed annotated code is definitely a lot easier to work with.

evilpie:
I decided to add basic loop support. We really only needed to handle the case where multiple jump instructions point to the same target, as well as targets before the jump.

coolreader18 (approved):
I'm not the most well versed in the jit compiler, but this looks good to me!

coolreader18:
Thanks for contributing!

PR #2279, opened 2020-10-11
Rework pyobjectrc.rs and add PyObjectRcB

coolreader18:


coolreader18:
Not sure if we would want to do a mass find/replace in this PR or in a separate one, but I think it could be done here.

youknowone:
Either way sounds good. Separated commits will be enough to make easy to track change.

youknowone:
I am not understanding this PR very well. What's the benefit to have PyObjectRcB?

PR #2281, opened 2020-10-11
vm::obj -> vm::builtins and expose types right under builtins rather than each objxxx submodules

youknowone:


youknowone:
@coolreader18  I want quick review for this PR. this is a huge source of conflict. this is making `builtins` as a vm sub module which matches to the bulitins module concept in python. It also will make breaking change to RustPython API surface. Users should use `use rustpython_vm::builtins::PyInt` after the change.
`vm/src/builtins/mod.rs` is not clear enough but i will be going to improve it.

coolreader18 (approved):
I think overall this is a good idea, though I feel like one namespace for **everything** might not be great. Maybe we could have `pub mod foo;` and just export `PyFoo, PyFooRef` under `builtins/mod.rs` and then we could still keep utility functions related to foo/foo-like objects under `builtins::foo`. Specifically thinking about stuff like objiter, objsequence, though maybe those should go in something like CPython's `abstract.c`.

youknowone:
I think exporting PyFoo, PyFooRef is better than `*`. About utilities, I vote for the latter. I think objiter utilities are better to be moved out of builtins module. About objsequence, it even is not a builtins component.

youknowone:
Oh, one things I think good to have builtins are `isinstance` and `issubclass`.

PR #2282, opened 2020-10-11
Reimplement bytes fromhex

qingshi163:


vm/src/bytesinner.rs
```diff
@@ -557,35 +557,46 @@ impl PyBytesInner {
...
-        // Hex is evaluated on 2 digits
-        if stripped.len() % 2 != 0 {
-            return Err(vm.new_value_error(format!(
-                "non-hexadecimal number found in fromhex() arg at position {}",
-                stripped.len() - 1
-            )));
-        }
+            let (i, b) = match iter.next() {
+                Some(val) => val,
+                None => break i,
```
coolreader18, line 583:
```suggestion
                None => break i + 1,
```
Looks like `bytes.fromhex('7')` throws `ValueError: non-hexadecimal number found in fromhex() arg at position 1`

qingshi163: You are right.

coolreader18 (approved):
Thanks for contributing!

PR #2284, opened 2020-10-11
generic and efficiency way to consume iterable object

qingshi163:


vm/src/pyobject.rs
```diff
@@ -900,6 +899,34 @@ where
...
 
+pub fn try_iterable_map<F, T, R>(
+    vm: &VirtualMachine,
+    obj: &PyObjectRef,
+    mut f: F,
+) -> PyResult<Vec<R>>
+where
+    T: TryFromObject,
+    F: FnMut(T) -> PyResult<R>,
+{
```
qingshi163, line 910: better name?

coolreader18: try_map_iter?

coolreader18:
Looks good! I think it would make sense to put this as a method on `VirtualMachine`, and then you could also define `extract_elements` in terms of this

qingshi163:
I think I stacked here, CPython allows create the bytes from an iterable object like a list, and inside the list you could put the object that has __index__(), but when this __index__() been called it allows the list to be modified!

This cause the problem if we map the list so the list is borrowed, and when __index__() try to modity the list we are in deadlock.

I think we may have to go back to extract all the elements from the list before try to convert the items. (CPython is doing like lazy-evaluation).

Or any idea that help?

qingshi163:
Or we need to do the map manually, so weclone the item out,free the lock,do the convert.
but this did not sound thread safe, is ok to go this way?
@coolreader18

youknowone:
When there is difference, following CPython way is almost the best.
Copying here doesn't seem to solve the compatibility problem regardless threading. If `__index__` edits the content of the list, it will not affect the copied content.

qingshi163:
I moved the mutable function from bytesinner to bytearray for better control of the borrowing.

vm/src/vm.rs
```diff
@@ -987,6 +987,53 @@ impl VirtualMachine {
         }
     }
 
+    pub fn map_elements<F, T, R>(&self, obj: &PyObjectRef, mut f: F) -> PyResult<PyResult<Vec<R>>>
+    where
+        T: TryFromObject,
+        F: FnMut(T) -> PyResult<R>,
+    {
+        self.map_iterable_object(obj, |elem| f(T::try_from_object(self, elem)?))
+    }
```
qingshi163, line 996: This function not been use yet, maybe remove it?

youknowone: Generally, yes. having unused code is a sort of cost. Do you have plan to use it in other patch?

qingshi163:
#2125 fixed

coolreader18 (approved):
LGTM

vm/src/iterator.rs
```diff
@@ -78,6 +78,24 @@ pub fn get_all<T: TryFromObject>(vm: &VirtualMachine, iter_obj: &PyObjectRef) ->
     Ok(elements)
 }
 
+pub fn try_map<F, R>(vm: &VirtualMachine, iter_obj: &PyObjectRef, mut f: F) -> PyResult<Vec<R>>
+where
+    F: FnMut(PyObjectRef) -> PyResult<R>,
+{
+    let cap = length_hint(vm, iter_obj.clone())?.unwrap_or(0);
+    // TODO: fix extend to do this check (?), see test_extend in Lib/test/list_tests.py,
+    // https://github.com/python/cpython/blob/master/Objects/listobject.c#L934-L940
```
youknowone, line 87: link to cpython master will be broken one day. link to specific tag will be better.

vm/src/iterator.rs
```diff
@@ -78,6 +78,24 @@ pub fn get_all<T: TryFromObject>(vm: &VirtualMachine, iter_obj: &PyObjectRef) ->
...
+    if cap >= isize::max_value() as usize {
```
youknowone, line 88: I think this is `sys.maxsize` but just found there is no way to access to the const in rust side.

youknowone (approved):

PR #2285, opened 2020-10-12
Clean up vm::builtins

youknowone:
- `objtype::isinstance` is now `TypeProtocol::isinstance`
- `objtype::issubclass` is now `DerefToPyType::issubclass`. but `use DerefToPyType` looks like a pain api. I added `PyTypeRef::issubclass` and `PyLease<'_, PyType>::issubclass` for now. I am lookign for a better idea.
- `vm.new_stop_iteration` is added
- `vm::iterator` is splited from `vm::obj::objiter`
- `vm.ascii` is removed. it is replacible by `vm::builtlins::ascii`


vm/src/builtins/make_module.rs
```diff
@@ -72,8 +72,30 @@ mod decl {
         Ok(false)
     }
 
+    /// Convert a string to ascii compatible, escaping unicodes into escape
+    /// sequences.
+    pub fn to_ascii(value: &str) -> String {
```
coolreader18, line 77: Could be in `rustpython_common::str`?

coolreader18 (approved):
LGTM

PR #2287, opened 2020-10-14
Create To find GCD of two numbers

vinayrevankar14:
To find the GCD of Two numbers

### DIFF
--- a/To find GCD of two numbers
+++ b/To find GCD of two numbers
@@ -1,0 +1,15 @@
+def gcd(a,b):
+    gcd = 1
+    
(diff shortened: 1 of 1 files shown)


coolreader18:
Hi @vinayrevankar14, this isn't related to RustPython at all; contributing a GCD algorithm doesn't really make sense for an interpreter. I'm going to mark this as spam, if you feel differently I'm open to hearing it.

youknowone:
The standard library includes `math.gcd` already
https://docs.python.org/3/library/math.html#math.gcd

PR #2289, opened 2020-10-16
FuncArgs kwargs as Option

youknowone:


coolreader18:
LGTM and I'm not opposed to doing this, but do you have any benchmarks for this? It looks like creating an empty indexmap should be pretty much free, it just needs to call `std::collections::hash_map::RandomState::new()` which just gets+modifies a thread-local variable after the first call.

PR #2290, opened 2020-10-17
Update the browser module

coolreader18:
cc @phorward

Main relevant changes:

- Move `Promise` to the `_js` module
- Rename `JsValue`/`JsError` to `JSValue`/`JSError` to fit with Python naming conventions
- If an exception bubbles up to js and it's a `JSError`, throw the jsvalue inside it
- `browser.jsclosure -> JSClosure`
```py
class JSClosure:
    value: Optional[JSValue]
    destroyed: bool
    detached: bool
    destroy(self)
    detach(self) -> JSValue
```
`JSClosure.destroy()` decreases the refcount of the inner python function that gets called, and makes the js function uncallable. `JSValue.detach()` makes it so the inner function never gets dropped. If anyone has any API critiques, I'd love to hear them; I'm not really sure if having `destroyed` and `detached` completely orthogonal to each other is the best way to do things.

### DIFF
--- a/vm/src/builtins/mod.rs
+++ b/vm/src/builtins/mod.rs
@@ -29,6 +29,7 @@ pub use function::PyFunction;
 pub(crate) mod generator;
 pub use generator::PyGenerator;
 pub(crate) mod getset;
+pub use getset::PyGetSet;
 pub(crate) mod int;
 pub use int::{PyInt, PyIntRef};
 pub(crate) mod iter;
--- a/wasm/lib/src/browser_module.rs
+++ b/wasm/lib/src/browser_module.rs
@@ -1,20 +1,18 @@
 use js_sys::Promise;
-use std::future::Future;
 use wasm_bindgen::prelude::*;
 use wasm_bindgen::JsCast;
-use wasm_bindgen_futures::{future_to_promise, JsFuture};
+use wasm_bindgen_futures::JsFuture;
 
 use rustpython_vm::builtins::{PyDictRef, PyStrRef, PyTypeRef};
 use rustpython_vm::common::rc::PyRc;
 use rustpython_vm::function::OptionalArg;
 use rustpython_vm::import::import_file;
 use rustpython_vm::pyobject::{
-    BorrowValue, IntoPyObject, PyCallable, PyClassImpl, PyObject, PyObjectRef, PyRef, PyResult,
-    PyValue,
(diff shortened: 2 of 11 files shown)


vm/src/builtins/mod.rs
```diff
@@ -29,6 +29,7 @@ pub use function::PyFunction;
 pub(crate) mod generator;
 pub use generator::PyGenerator;
 pub(crate) mod getset;
+pub use getset::PyGetSet;
```
youknowone, line 32: oh, i thought this is internally used only but not in this case. maybe we'd better to expose builtinfunc also.

youknowone (approved):

wasm/lib/src/browser_module.rs
```diff
@@ -361,11 +263,10 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
 
     let element = Element::make_class(ctx);
 
-    py_module!(vm, "browser", {
+    py_module!(vm, "_browser", {
```
phorward, line 266: Ok this allows now for an `import _browser` in WASM.

But what are the differences between the browser (frozen) and the browser (built-in) module, they don't seem to be the same:
```
>>>> dir(browser)
['Document', 'Element', 'JSValue', 'Promise', '__annotations__', '__builtins__', '__doc__', '__loader__', '__name__', '__package__', '__spec__', '_alert', '_confirm', '_prompt', 'alert', 'cancel_animation_frame', 'confirm', 'document', 'fetch', 'jsclosure', 'jsstr', 'load_module', 'prompt', 'request_animation_frame', 'window']
>>>>> dir(_browser)
['Document', 'Element', '__doc__', '__loader__', '__name__', '__package__', '__spec__', 'cancel_animation_frame', 'document', 'fetch', 'load_module', 'request_animation_frame']
>>>>> browser
<module 'browser' (frozen)>
>>>>> _browser
<module '_browser' (built-in)>
```

coolreader18: It doesn't actually change anything but `import _browser; _browser.__name__`, if you look at the source of `browser.py` you can see that it imports `*` from `_browser`. `_browser` has stuff like fetch and document, `browser` wraps it and also provides some other functions that are easy to implement in Python. It's like the built-in `posix`/`nt` vs the Python `os`.

coolreader18: [For reference](https://github.com/RustPython/RustPython/blob/master/wasm/lib/src/browser.py)

PR #2294, opened 2020-10-18
Store PyObjectRef in CodeObject constants

coolreader18:
|:---|---:|---:|---:|---:|
| `./rustpython-norm benchmarks/pystone.py 50000` | 7.610 ± 0.181 | 7.369 | 7.927 | 1.09 ± 0.03 |
| `./rustpython-opt3 benchmarks/pystone.py 50000` | 6.999 ± 0.088 | 6.888 | 7.145 | 1.00 |

A bit faster, and should be in general better for memory usage. This also paves the way to string interning; `ctx.unwrap_constant()` can intern strings and that should also improve memory usage and performance (due to str hash caching and identity comparison in dicts).

I also removed the peephole optimizer; I was the one who originally wrote it and I didn't feel like rewriting it to work with the `constants` vec. I also feel that in general, it was adding too much complexity for not enough benefit. I think we can make a better optimizer, but it definitely has to be something that operates on an already compiled codeobject; the streaming optimizer thing we had with `OutputStream` just doesn't really work.

I also changed the compiler to use `insta` for snapshot testing of compilation results


coolreader18:
@youknowone could you take a look at this yourself?

youknowone (approved):

PR #2295, opened 2020-10-20
Expose sys module constants in rust level

youknowone:


vm/src/stdlib/sys.rs
```diff
@@ -16,6 +16,8 @@ use crate::{builtins, exceptions, py_io, version};
 /*
  * The magic sys module.
  */
+const MAXSIZE: usize = std::isize::MAX as usize;
```
youknowone, line 19: https://github.com/RustPython/RustPython/pull/2284#discussion_r507437781

coolreader18:
I'm not sure... I feel like it doesn't totally make sense to have it in the `stdlib` directory, since it's more a function of the python interpreter rather than the standard library, and that's why CPython has it as `Python/sysmodule.c`. Like, why not put the `builtins` module in `stdlib/`? I think it's the same sort of thing.

youknowone:
I removed the movnig part and remained only constant exposure part

PR #2297, opened 2020-10-20
Update pystruct + array

coolreader18:
This is sort of a big lump change, and there aren't necessarily tests to prove it, since there aren't any tests in `test_struct.py` that verify the native ordering -- even though there's padding now and correct alignment/size, the tests that I unskipped aren't related to that :disappointed: . Still, I think this is cleaner then it was before; there's an enum for the codes now, and finding the `pack()` function for a code is a bit cleaner now and respects endianness/architecture

### DIFF
--- a/vm/src/stdlib/pystruct.rs
+++ b/vm/src/stdlib/pystruct.rs
@@ -960,21 +960,28 @@ pub(crate) mod _struct {
     #[pyfunction]
     fn _clearcache() {}
 
+    rustpython_common::static_cell! {
+        pub(crate) static STRUCT_ERROR: PyTypeRef;
+    }
+
     fn new_struct_error(vm: &VirtualMachine, msg: String) -> PyBaseExceptionRef {
-        // _struct.error must exist
-        let class = vm.try_class("_struct", "error").unwrap();
-        vm.new_exception_msg(class, msg)
+        let class = STRUCT_ERROR.get().unwrap();
+        vm.new_exception_msg(class.clone(), msg)
     }
 }
 
 pub(crate) fn make_module(vm: &VirtualMachine) -> PyObjectRef {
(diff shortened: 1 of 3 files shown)


youknowone:
@qingshi163 would you review this PR?

qingshi163:
Seem like a problem if we want to pack a float but give a int.

```
import struct
struct.pack('f', 10)
```

vm/src/stdlib/pystruct.rs
```diff
@@ -346,101 +505,125 @@ pub(crate) mod _struct {
...
-    }
+    make_pack_with_endianess!(i16, get_int_or_index);
+    make_pack_with_endianess!(u16, get_int_or_index);
+    make_pack_with_endianess!(i32, get_int_or_index);
+    make_pack_with_endianess!(u32, get_int_or_index);
+    make_pack_with_endianess!(i64, get_int_or_index);
+    make_pack_with_endianess!(u64, get_int_or_index);
+    make_pack_with_endianess!(f32, TryFromObject::try_from_object);
+    make_pack_with_endianess!(f64, TryFromObject::try_from_object);
+
```
qingshi163, line 561: CPython allows pack float format with int, float or class with __float__() but not a float like string. Non of our function doing exactly the job. We may need one it also can be use by array.array

coolreader18: I think `IntoPyFloat::try_from_object` is correct; it accepts float, int, or anything with a `__float__` method

coolreader18: I modified it in #2309 to not accept a float string, since that's what `PyFloat_AsDouble` does

youknowone (approved):

PR #2298, opened 2020-10-21
Implement math.ulp

ArniDagur:


ArniDagur:
CI fails with
```
error[E0463]: can't find crate for `serde_derive` which `num_bigint` depends on
Error:  --> bytecode/src/bytecode.rs:6:5
  |
6 | use num_bigint::BigInt;
  |     ^^^^^^^^^^ can't find crate

```
Is this related to anything I'm doing? It seems unlikely to me.

coolreader18:
No, that's an existing problem related to cache stuff on macos; nothing to do with your changes.

Lib/test/test_math.py
```diff
@@ -1948,75 +1946,74 @@ def testComb(self):
...
+        self.assertEqual(math.nextafter(-INF, 0.0), -largest_normal)
+        self.assertEqual(math.nextafter(largest_normal, INF), INF)
+        self.assertEqual(math.nextafter(-largest_normal, -INF), -INF)
+
+        # NaN
+        self.assertIsNaN(math.nextafter(NAN, 1.0))
+        self.assertIsNaN(math.nextafter(1.0, NAN))
+        self.assertIsNaN(math.nextafter(NAN, NAN))
+
+    @requires_IEEE_754
```
coolreader18, line 1995: This won't actually be run because of the stub impl of `requires_IEEE_754` -- you could try changing it in `support/__init__.py` to `requires_IEEE_754 = unittest.skipIf(False, "RustPython always has IEEE 754 floating point numbers")`

ArniDagur: Silly me. I commented out the lines in the tests that failed. I both perform operations on the constant `sys.float_info.min * sys.float_info.epsilon`. I wonder if RustPython sets these constants incorrectly or something?

coolreader18: Oh, yeah, I think you're right. I think the `min: f64::MIN` in `sysmodule.rs` should actually be `min: f64::MIN_POSITIVE`

ArniDagur: Thanks. That got the test passing.

coolreader18:
Hmm, looks like some complex tests were enabled that were previously skipped by `requires_IEEE_754`. Do you want to add skips for those manually?

ArniDagur:
Done.

I think in general we should use `expectedFailure` instead of `skip`, since implementing feature A in one part of the codebase may cause tests for features B and C in other parts of the codebase to run successfully. If the tests are skipped, we may never know.

coolreader18:
Oh wow, I just realized that you can still have expectedFailure if the test is throwing an error. But yeah I definitely agree, we should mention in the development document that it's preferred to do expectedFailure, if the test isn't hanging/panicking

coolreader18 (approved):
Thanks for contributing!

PR #2299, opened 2020-10-22
Implement memoryview hash

qingshi163:


vm/src/builtins/memory.rs
```diff
@@ -118,6 +119,7 @@ pub struct PyMemoryView {
     step: isize,
     exports: AtomicCell<usize>,
     format_spec: FormatSpec,
+    hash: OnceCell<PyHash>,
```
youknowone, line 122: is this needed to be cached?

youknowone (approved):
This is ok if hash is needed to be cached inside of memoryview.
While reviewing this PR, I found memoryview internal looks like to be always contiguous. Do you know if it is actually always contiguous in CPython or that's only applied for this API?

https://docs.python.org/3/c-api/memoryview.html#c.PyMemoryView_GetContiguous

vm/src/builtins/memory.rs
qingshi163: CPython cache the hash value, also because CPython allows to get the hash value after the memoryview is released if it has been hashed before.

qingshi163: No, in memoryview it not always looks contiguous. Only this function GetContiguous will create a new PyBytes inside if the buffer is non-contiguous. If you create memoryview via FromBuffer, than the contiguous is depends on buffer.

PR #2301, opened 2020-10-22
Make StaticCell less complex to use

coolreader18:
Also made some tweaks to socket

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -111,63 +111,59 @@ impl PyContext {
     pub const INT_CACHE_POOL_MIN: i32 = -5;
     pub const INT_CACHE_POOL_MAX: i32 = 256;
 
+    fn init() -> Self {
+        flame_guard!("init PyContext");
+        let types = TypeZoo::init();
+        let exceptions = exceptions::ExceptionZoo::init();
+
+        fn create_object<T: PyObjectPayload + PyValue>(payload: T, cls: &PyTypeRef) -> PyRef<T> {
+            PyRef::new_ref(payload, cls.clone(), None)
+        }
+
+        let none = create_object(PyNone, PyNone::static_type());
+        let ellipsis = create_object(PyEllipsis, PyEllipsis::static_type());
+        let not_implemented = create_object(PyNotImplemented, PyNotImplemented::static_type());
+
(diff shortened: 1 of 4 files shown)


coolreader18:
Actually, not sure about using `static_cell!` for error types; we wouldn't want properties on the exception types to be carried across sub-interpreters.

coolreader18:
Although it looks like that's not restricted for builtin types either:
```
>>>>> type.a = 1
>>>>> 
```

vm/src/pyobject.rs
```diff
@@ -1242,23 +1238,25 @@ pub trait StaticType {
...
     fn static_type() -> &'static PyTypeRef {
-        static_cell::get(Self::static_cell()).unwrap_or_else(|| unsafe {
-            // SAFETY: object must be initialized by init_* method.
-            // So this is actually not safe as itself.
-            // But easy to find out when it happened in debug build
-            std::hint::unreachable_unchecked()
-        })
+        Self::static_cell()
+            .get()
+            .expect("static type has not been initialized")
```
youknowone, line 1243: how about keeping `std::hint::unreachable_unchecked` here? this will be frequently called for `PyValue::class` and `expect` will remain None check even it will not fail (unless it has a bug which is easy to be detected in debug mode)

youknowone (approved):

youknowone:
```
>>>>> type.a = 1
>>>>> 
```
this looks like a bug. does this type marked as `HAS_DICT`?

vm/src/pyobject.rs
coolreader18: I don't think it will be called all that often, since built-in types will implement `PyValue::class` by borrowing from `vm.ctx`, and this is unsound -- somebody could just do `#[pyclass] struct Foo; Foo::static_type()` and get UB. We could make it an unsafe function maybe? Idk, there's still not really a guarantee that an `init` function would be called before calling into Python code that calls into static_type. This is tricky though, cause we're getting into the soundness vs performance aspect of making Rust libraries 😟

coolreader18:
Well user classes can be edited like that:

```py
class A: pass
A.foo = 1
```

But I think we probably need a tpflag to prevent built-in types from being mutated

vm/src/pyobject.rs
youknowone: Ah, that's right. I was forgetting builtin type implementation for that.

PR #2304, opened 2020-10-24
socket: resolve addresses according to socket family (ipv4/6)

### LINKED ISSUE #2241: unittest test_support.py Failed
Env:
Manjaro 5.8.6-1
rustc 1.46.0

The unittest from test_support.py gives me the error on master branch.
Follow the test, I find out if we call socket.bind with the address 'localhost', than we fail, but CPython will treat it as 127.0.0.1.

```
ERROR: test_HOST (__main__.TestSupport)
  File "Lib/test/test_support.py", line 94, in test_HOST
    s = socket.create_server((support.HOST, 0))
  File "/home/skz/Projects/RustPython/vm/pylib-crate/Lib/socket.py", line 898, in create_server
    raise
  File "/home/skz/Projects/RustPython/vm/pylib-crate/Lib/socket.py", line 889, in create_server
    (err.strerror, address)
AttributeError: 'OSError' object has no attribute 'strerror'

ERROR: test_bind_port (__main__.TestSupport)
  File "Lib/test/test_support.py", line 104, in test_bind_port
    support.bind_port(s)
  File "/home/skz/Projects/RustPython/vm/pylib-crate/Lib/test/support/__init__.py", line 796, in bind_port
    sock.bind((host, 0))
OSError: Address family not supported by protocol (os error 97)

ERROR: test_find_unused_port (__main__.TestSupport)
  File "Lib/test/test_support.py", line 98, in test_find_unused_port
    port = support.find_unused_port()
  File "/home/skz/Projects/RustPython/vm/pylib-crate/Lib/test/support/__init__.py", line 759, in find_unused_port
    port = bind_port(tempsock)
  File "/home/skz/Projects/RustPython/vm/pylib-crate/Lib/test/support/__init__.py", line 759, in find_unused_port
    port = bind_port(tempsock)
  File "/home/skz/Projects/RustPython/vm/pylib-crate/Lib/test/support/__init__.py", line 796, in bind_port
    sock.bind((host, 0))
OSError: Address family not supported by protocol (os error 97)
```

awoimbee:
Fixes #2241

As I said on the issue:
> std::net::ToSocketAddrs does its job. host: "localhost" result: Ok(IntoIter([[::1]:5555, 127.0.0.1:5555]))
> Then get_addr() just reads the first one (which is IPv6 on my system).
> -> We then pass this IPv6 address to an IPv4 socket <-
> So get_addr() just need to be aware of PySocket::family and to filter incompatible IPs

I'm not a fan of `IpAddrFmt` but it's the cleanest way I found to do this.
I also removed `RUSTPYTHON_NO_IPV6` since it's no longer necessary :)

### DIFF
--- a/vm/src/stdlib/socket.rs
+++ b/vm/src/stdlib/socket.rs
@@ -5,7 +5,7 @@ use gethostname::gethostname;
 use nix::unistd::sethostname;
 use socket2::{Domain, Protocol, Socket, Type as SocketType};
 use std::io::{self, prelude::*};
-use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, Shutdown, SocketAddr, ToSocketAddrs};
+use std::net::{Ipv4Addr, Ipv6Addr, Shutdown, SocketAddr, ToSocketAddrs};
 use std::time::Duration;
 
 use crate::builtins::bytearray::PyByteArrayRef;
@@ -130,7 +130,7 @@ impl PySocket {
 
     #[pymethod]
     fn connect(&self, address: Address, vm: &VirtualMachine) -> PyResult<()> {
-        let sock_addr = get_addr(vm, address)?;
+        let sock_addr = get_addr(vm, address, self.family.load())?;
         let res = if let Some(duration) = self.sock().read_timeout().unwrap() {
             self.sock().connect_timeout(&sock_addr, duration)
         } else {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/socket.rs
```diff
@@ -691,32 +691,51 @@ fn socket_getnameinfo(
...
+        } else {
+            error!("Unknown IP family/domain: {:?}", i);
+            Self::Any()
         }
+    }
+}
+
+fn get_addr<F>(vm: &VirtualMachine, addr: impl ToSocketAddrs, fmt: F) -> PyResult<socket2::SockAddr>
+where
+    F: Into<IpAddrFmt>,
```
coolreader18, line 718: Maybe the "fmt" argument could just be something like `family: Option<i32>`? I think that's more direct (since this is only used in internal code), and it avoids needing `IpAddrFmt`

coolreader18 (approved):
Looks great! I'll merge once CI is green, thanks for contributing!

PR #2305, opened 2020-10-24
Update random.py with CPython 3.7 version.

ChJR:
This change makes random._os, found by whats_left.sh.

### DIFF
--- a/Lib/random.py
+++ b/Lib/random.py
@@ -41,17 +41,12 @@ from warnings import warn as _warn
 from types import MethodType as _MethodType, BuiltinMethodType as _BuiltinMethodType
 from math import log as _log, exp as _exp, pi as _pi, e as _e, ceil as _ceil
 from math import sqrt as _sqrt, acos as _acos, cos as _cos, sin as _sin
-try:
-    from os import urandom as _urandom
-except ImportError:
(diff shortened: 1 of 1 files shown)


Lib/random.py
```diff
@@ -41,17 +41,12 @@
 from types import MethodType as _MethodType, BuiltinMethodType as _BuiltinMethodType
 from math import log as _log, exp as _exp, pi as _pi, e as _e, ceil as _ceil
 from math import sqrt as _sqrt, acos as _acos, cos as _cos, sin as _sin
-try:
-    from os import urandom as _urandom
-except ImportError:
-    # On wasm, _random.Random.random() does give a proper random value, but
-    # we don't have the os module
```
youknowone, line 48: Doesn wasm now have urandom?

coolreader18: I don't think so; @chjr could you forward-port (?) this edit to the earlier random.py version? Maybe just put the os imports in the try block?

ChJR: I see. I didn't check on wasm. I'll try the method suggested by @coolreader18 .

coolreader18 (approved):
Thanks for contributing

PR #2306, opened 2020-10-24
Add format %e option of floating point numbers.

ChJR:


youknowone (commented):
The change generally looks good. 
will you please add a test for this?
Also please check the clippy errors

ChJR:
I see. I added some tests and fixed clippy error.

coolreader18:
Actually, we don't usually add our own tests to `Lib/test`, since that's a CPython library file that we might update at times like you did `random.py`. Could you move the test to somewhere in `extra_tests/snippets`? `strings.py` looks like it fits.

ChJR:
That's what I want to know. I'll move the test to extra_tests. Thanks for reviewing my code.

coolreader18 (approved):

coolreader18:
Thanks for contributing!

PR #2307, opened 2020-10-25
Fix default format(float) result when the number is too big

ChJR:
Before this change, format(1e16) returns '10000000000000000.0' in RustPython while CPython 3.7 returns '1e+16'.
After this change, format(1e16) returns '1e+16' in RustPython too.

### DIFF
--- a/vm/src/format.rs
+++ b/vm/src/format.rs
@@ -404,8 +404,6 @@ impl FormatSpec {
                 match magnitude {
                     magnitude if magnitude.is_nan() => Ok("nan".to_owned()),
                     magnitude if magnitude.is_infinite() => Ok("inf".to_owned()),
-                    // Using the Debug format here to prevent the automatic conversion of floats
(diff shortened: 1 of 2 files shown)


vm/src/format.rs
```diff
@@ -405,7 +406,7 @@ impl FormatSpec {
                     magnitude if magnitude.is_infinite() => Ok("inf".to_owned()),
                     // Using the Debug format here to prevent the automatic conversion of floats
                     // ending in .0 to their integer representation (e.g., 1.0 -> 1)
```
coolreader18, line 408: You could probably remove this explanation, since it doesn't apply anymore.

coolreader18 (approved):
Nice work!

vm/src/format.rs
ChJR: Good points!

coolreader18:
Thanks for contributing!

PR #2309, opened 2020-10-25
Implement nextafter in Rust

coolreader18:


youknowone (approved):
looks good, but clippy error need to be suppressed

PR #2311, opened 2020-10-26
Add socket sendfile function

rodrigocam:
Add socket function that sends all content of a file through the socket.

### DIFF
--- a/extra_tests/snippets/stdlib_os.py
+++ b/extra_tests/snippets/stdlib_os.py
@@ -17,6 +17,18 @@ assert_raises(FileNotFoundError,
 assert_raises(FileNotFoundError,
               lambda: os.rename('DOES_NOT_EXIST', 'DOES_NOT_EXIST 2'))
 
+src_fd = os.open('README.md', os.O_RDONLY)
+dest_fd = os.open('destination.md', os.O_RDWR | os.O_CREAT)
+src_len = os.stat('README.md').st_size
+
+bytes_sent = os.sendfile(dest_fd, src_fd, 0, src_len)
+assert src_len == bytes_sent
+
+os.lseek(dest_fd, 0, 0)
+assert os.read(src_fd, src_len) == os.read(dest_fd, bytes_sent)
+os.close(src_fd)
+os.close(dest_fd)
+
 try:
     os.open('DOES_NOT_EXIST', 0)
 except OSError as err:
--- a/extra_tests/snippets/stdlib_os.py
+++ b/extra_tests/snippets/stdlib_os.py
@@ -17,17 +17,18 @@ assert_raises(FileNotFoundError,
 assert_raises(FileNotFoundError,
               lambda: os.rename('DOES_NOT_EXIST', 'DOES_NOT_EXIST 2'))
 
-src_fd = os.open('README.md', os.O_RDONLY)
-dest_fd = os.open('destination.md', os.O_RDWR | os.O_CREAT)
-src_len = os.stat('README.md').st_size
+if hasattr(os, "sendfile"):
+    src_fd = os.open('README.md', os.O_RDONLY)
+    dest_fd = os.open('destination.md', os.O_RDWR | os.O_CREAT)
+    src_len = os.stat('README.md').st_size
 
-bytes_sent = os.sendfile(dest_fd, src_fd, 0, src_len)
-assert src_len == bytes_sent
+    bytes_sent = os.sendfile(dest_fd, src_fd, 0, src_len)
+    assert src_len == bytes_sent
 
-os.lseek(dest_fd, 0, 0)
-assert os.read(src_fd, src_len) == os.read(dest_fd, bytes_sent)
-os.close(src_fd)
-os.close(dest_fd)
+    os.lseek(dest_fd, 0, 0)
+    assert os.read(src_fd, src_len) == os.read(dest_fd, bytes_sent)
+    os.close(src_fd)
+    os.close(dest_fd)
 
 try:
     os.open('DOES_NOT_EXIST', 0)
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -357,25 +357,26 @@ mod _os {
     #[cfg(any(target_os = "macos"))]
     #[pyfunction]
     pub(crate) fn sendfile(
-        out_fd: i64,
-        in_fd: i64,
-        offset: u64,
+        out_fd: i32,
+        in_fd: i32,
+        offset: i64,
         count: u64,
         headers: OptionalArg<PyObjectRef>,
         trailers: OptionalArg<PyObjectRef>,
         flags: OptionalArg<i64>,
         vm: &VirtualMachine,
     ) -> PyResult {
-        let res = nix::sys::sendfile::sendfile(
+        let mut _count = count;
+        let (res, written) = nix::sys::sendfile::sendfile(
             in_fd,
             out_fd,
             offset,
-            Some(count as usize),
-            headers,
-            trailers,
+            Some(&mut _count),
+            headers.into_option(),
+            trailers.into_option(),
         )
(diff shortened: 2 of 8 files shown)


coolreader18:
I'm so sorry, I should've checked before but this isn't actually how the sendfile method is implemented. It's implemented in the Python `socket.py` module, and it calls `os.sendfile()`, which is a low-level function. If you'd like to take a shot at that you can try it ([cpython version](https://github.com/python/cpython/blob/c9bc290dd6e3994a4ead2a224178bcba86f0c0e4/Modules/posixmodule.c#L9275) [linux signature](https://docs.rs/libc/0.2.80/libc/fn.sendfile.html) [macos signature](https://docs.rs/libc/0.2.80/x86_64-apple-darwin/libc/fn.sendfile.html)) but otherwise this isn't really correct; hopefully you got a bit familiar with the codebase at least. Sorry again :persevere:

rodrigocam:
Oh, no problem. I will try to implement using low level functions. I saw that if the OS doesn't have `sendfile` it use `send` like I did, this will occur on Windows. So in our implementation, we will have a compilation macro that will choose linux signature or mac signature to `os.sendfile` and in `socket.sendfile` we will use it right? There is any example of doing this compilation macro?

coolreader18:
>So in our implementation, we will have a conditional compilation that will choose linux signature or mac signature to os.sendfile and in socket.sendfile we will use it right?

Yup, pretty much. We don't even need to edit socket.py since it already checks that `os.sendfile` exists and uses it to provide `socket.sendfile`:

https://github.com/RustPython/RustPython/blob/3c9ad2378bd18bb2febfd29eb6e4ff14a4c21c9d/Lib/socket.py#L343

For examples, you can look at other functions in `os.rs`, e.g. getgroups is implemented separately in `macos` than it is on `linux`. For sendfile, you'd probably want something like this:

```rust
#[cfg(any(target_os = "linux", target_os = "otheros..."))]
#[pyfunction]
fn sendfile(out_fd: i32, in_fd: i32, offset: ...) -> PyResult<...> {
   ...
}

#[cfg(any(target_os = "macos", target_os = "otheros..."))]
#[pyfunction]
fn sendfile(out_fd: i32, in_fd: i32, ..., headers: PyObjectRef, trailers: PyObjectRef, flags: i32) -> PyResult<...> {
    ...
}
```
You probably want to use the nix crate's sendfile wrappers, [linux](https://docs.rs/nix/0.19.0/nix/sys/sendfile/fn.sendfile.html) and [macos](https://docs.rs/nix/0.19.0/x86_64-apple-darwin/nix/sys/sendfile/fn.sendfile.html); that would mean you wouldn't have to deal with the raw libc APIs yourself.

rodrigocam:
Nice! I implemented the `os.sendfile` function, now I a have to use this function on `socket.sendfile` ?

extra_tests/snippets/stdlib_os.py
```diff
@@ -17,6 +17,18 @@
...
+src_len = os.stat('README.md').st_size
+
+bytes_sent = os.sendfile(dest_fd, src_fd, 0, src_len)
+assert src_len == bytes_sent
+
+os.lseek(dest_fd, 0, 0)
+assert os.read(src_fd, src_len) == os.read(dest_fd, bytes_sent)
+os.close(src_fd)
+os.close(dest_fd)
+
```
coolreader18, line 31: You probably want to put this in a `if hasattr(os, "sendfile"):` block

vm/src/stdlib/os.rs
```diff
@@ -337,6 +337,48 @@ mod _os {
...
+        vm: &VirtualMachine,
+    ) -> PyResult {
+        let mut _count = count;
+        let (res, written) = nix::sys::sendfile::sendfile(
+            in_fd,
+            out_fd,
+            offset,
+            Some(&mut _count),
+            headers.into_option(),
+            trailers.into_option(),
```
coolreader18, line 376: You have to process these into an array of buffers (`&[&[u8]]`) first -- something like
```rust
let headers = match headers.into_option() {
    Some(x) => Some(vm.extract_elements::<PyBytesLike>(&x)?),
    None => None,
};
let headers = headers.as_ref().map(|v| v.iter().map(|b| b.borrow_value()).collect::<Vec<_>>());
let headers = headers.as_ref().map(|v| v.iter().map(|borrowed| &**borrowed).collect::<Vec<_>>());
let headers = headers.as_deref();
```
And then the same thing for trailers. That's really inefficient, what with creating like 3 different vecs, but I think I could submit a PR to `nix` to allow sendfile to accept any `IntoIterator` over `AsRef<[u8]>`, and then we could make this more efficient later.

extra_tests/snippets/stdlib_os.py
rodrigocam: Done!

vm/src/stdlib/os.rs
rodrigocam: So basically with those lines, we are creating bytes from the PyObject? I put those lines but now the macos tests give me this error:

```shell
error[E0277]: the trait bound `for<'r> fn(i32, i32, i64, i64, function::OptionalArg, function::OptionalArg, function::OptionalArg<i64>, &'r vm::VirtualMachine) -> std::result::Result<pyobjectrc::PyObjectRc, pyobject::PyRef<exceptions::PyBaseException>> {stdlib::os::_os::sendfile}: function::sealed::PyNativeFuncInternal<_, _, _>` is not satisfied
```

coolreader18:
To fix the "IntoPyNativeFunc is not implemented" error, you can add another macro call after line 642 in `vm/src/function.rs`

vm/src/stdlib/os.rs
coolreader18: Oh, to fix that you can add another macro call (or 2?) after line 642 in `vm/src/function.rs`

coolreader18: And yep, `PyBytesLike` is anything that implements the buffer protocol (e.g. `bytes`, `bytearray`, `memoryview`) and `PyBytesLike.borrow_value()` gets a read lock to the data, and then the `&**` dereferences it to a `&[u8]`.

rodrigocam: Done! Now that `os` module has `sendfile` function, a python test is broken because use a `create_file` function that tries to open the file in `xb` mode, and there's no implementation for this mode.

coolreader18:
I added a commit to skip the test that's failing because of an unrelated issue with `os.open`, I think otherwise this is good to merge! (if CI doesn't fail for some reason)

rodrigocam:
Thanks! I loved to contribute to this project, what do you think are the next steps to keeping contributing to the project?

vm/src/stdlib/os.rs
```diff
@@ -337,6 +337,67 @@ mod _os {
...
+    #[cfg(any(target_os = "macos"))]
+    #[pyfunction]
+    pub(crate) fn sendfile(
+        out_fd: i32,
+        in_fd: i32,
+        offset: i64,
+        count: i64,
+        headers: OptionalArg<PyObjectRef>,
+        trailers: OptionalArg<PyObjectRef>,
+        flags: OptionalArg<i32>,
```
youknowone, line 366: Adding `_` prefix like `_flags` will remove this warning

youknowone (commented):
test_shutil also fails for now

coolreader18 (approved):

coolreader18:
Looks great, thanks for contributing! re: continue contributing, one way you can do it is to find tests that are expectedFailures/skipped and fix them, or else just try to do stuff in rustpython that you might do usually in python and see if it works (heads up, anything that uses numpy isn't going to happen for a while, since we don't have a c-api)

youknowone:
This article maybe helpful for test work: https://rustpython.github.io/guideline/2020/04/04/how-to-contribute-by-cpython-unittest.html

PR #2312, opened 2020-10-27
Fix array with float

qingshi163:


coolreader18 (approved):

coolreader18:
Any idea why windows snippets tests are failing?

qingshi163:
incompatible issue from CPython?

coolreader18:
Ah, probably, github might have updated the default CPython version

coolreader18:
@qingshi163 try rebasing onto master, that should fix that CI issue

PR #2313, opened 2020-10-28
wasm/notebook updates

mireille-raad:
Updates to wasm/notebook. Here is the main summary:
- added CSS and JS support
- Added ability to create python documents with "new tab"
- Added split view and linked doc for code editor
- Added snippets of code that show how to create a basic UI with JS, pass parameters to Python
- UI fixes, import code pop-up
- moved the files around to `editor.js`, `proces.js`,` tools.js`. Tried to clean up the code. emphasis on tried.
- added an overdue readme 😆 

Here is how the notebook like now:
screenshot

JS support and code structure need lot of improvements.


coolreader18 (approved):
Looks great!! I think there might be a way to make the communication between JS and Python more elegant, but I can look into that later; maybe something like a js function in scope like
```js
addPythonFunction("create_ui", (pycallback) => {
  // call callback when user click button...
  pycallback([...args], {...kwargs})
})
```

mireille-raad:
It would be neat if there was an elegant way to communicate between JS and Python. 

I was thinking of adding peer to peer collaborative editing on the notebook, so for a user, it would feel like google doc. I don't think I have seen such a feature anywhere else like with jupyter, iodide or others (I might have missed it). 

I think it will be neat to add collaborative editing and useful for things like learning python and helping someone fix a small piece of code or function by going to the same URL. Do you have opinions about how to implement that with peer to peer/without servers? I was doing research about OT, webRTC, CRDTs. I saw things like YJS and a few others... any recommendations or should I do my research? 

Once this PR is merged, i'll start on collaborative web editing :D

coolreader18:
I've heard good things about CRDTs (mainly from [this article](https://josephg.com/blog/crdts-are-the-future/)) but I haven't really done anything with real-time editing/syncing before, so I don't really have any opinions. Yjs does look pretty good though.

mireille-raad:
@coolreader18 that was such a good read... especially the "what's next" section. I'll read around for my own info but I'll probably end-up using Yjs for code mirror as in https://github.com/yjs/y-codemirror

PR #2316, opened 2020-10-31
Unify float formatting

ChJR:
This change will reduce duplications.


common/src/float_ops.rs
```diff
@@ -83,6 +83,23 @@ pub fn is_integer(v: f64) -> bool {
     (v - v.round()).abs() < std::f64::EPSILON
 }
 
+// Formats floats into Python style exponent notation, by first formatting in Rust style
+// exponent notation (`1.0000e0`), then convert to Python style (`1.0000e+00`).
+pub fn format_float_as_exponent(precision: usize, magnitude: f64) -> String {
```
coolreader18, line 88: Maybe you could have a `upper: bool` param, so you don't have to do `to_lowercase`/`to_uppercase` on the result?

youknowone: I think this change causes unnessessary cost for to_lowercase and to_uppercase for common finite numbers which maybe coolreader18 wanted to point out.
At the same time, I think bool paramter doesn't fit for function design because the parameter need to reperesent upper and lower rather than a bool value.
maybe we need `FormatType::Exponent::Upper` and `FormatType::Exponent::Lower` so that we just can pass the value from `parse` function.

common/src/float_ops.rs
```diff
@@ -83,6 +83,23 @@ pub fn is_integer(v: f64) -> bool {
...
+    if magnitude.is_nan() {
```
youknowone, line 89: I suggest to check `magnitude.is_finite()` first. Which is more common case but for now experiencing 2 branches for every formatting.

common/src/float_ops.rs
```diff
@@ -83,21 +83,45 @@ pub fn is_integer(v: f64) -> bool {
...
-    if magnitude.is_infinite() {
-        return "inf".to_string();
+pub fn format_float_as_exponent(precision: usize, magnitude: f64, case: FloatFormatCase) -> String {
+    match magnitude {
+        magnitude if magnitude.is_finite() => {
+            let r_exp = format!("{:.*e}", precision, magnitude);
+            let mut parts = r_exp.splitn(2, 'e');
+            let base = parts.next().unwrap();
+            let exponent = parts.next().unwrap().parse::<i64>().unwrap();
+            format!("{}e{:+#03}", base, exponent)
```
coolreader18, line 119: I think this `e` should be dependent on format case as well.

ChJR: My mistake.

ChJR:
I am working on %g now.

ChJR:
I found some bug in this PR. Please wait merging with new commits fixing that bug.

ChJR:
OK. I am done.

vm/src/builtins/pystr.rs
```diff
@@ -180,7 +180,11 @@ impl PyIter for PyStrReverseIterator {
                     break;
                 }
             }
-            start.unwrap_or(end - 4)
+            if end < 4 {
+                start.unwrap_or(0)
+            } else {
+                start.unwrap_or(end - 4)
+            }
```
coolreader18, line 187: This can be `start.saturating_sub(4)`

common/src/float_ops.rs
```diff
@@ -83,6 +83,100 @@ pub fn is_integer(v: f64) -> bool {
...
+    s
+}
+
+pub fn format_general(precision: usize, magnitude: f64, case: Case) -> String {
+    match magnitude {
+        magnitude if magnitude.is_finite() => {
+            let prec = match precision {
+                0 => 0,
+                _ => precision - 1,
+            };
```
coolreader18, line 155:
```suggestion
            let prec = precision.saturating_sub(1);
```

coolreader18 (approved):
Looks great!! I left a few comments, and you can rebase to master if you want to fix that CI failure, but otherwise this is good to go!

coolreader18:
Thanks for contributing!

Issue #2121, opened 2020-08-17
%s formatting of byte strings in byte strings difference with CPython


glandium:
The following python code:
```
b"%s" % b"foo"
```

returns
```
b'foo'
```
with python but

```
b'b'foo''
```
with rustpython.

(btw the formatting of the output string in the REPL is not valid python)

coolreader18:
Oh, wow, nice catch!

glandium:
It's not really weird, given that that's the right output if the format string is a string rather than a bytestring. Looking at the code, the problem is that bytestring format strings are not treated differently.

ChJR:
In CPython version 3.7
```
>>> b'%s' % 'a'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: %b requires a bytes-like object, or an object that implements __bytes__, not 'str'
>>> b'%b' % 'a'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: %b requires a bytes-like object, or an object that implements __bytes__, not 'str'
>>> b'%b' % b'a'
b'a'
>>> b'%s' % b'a'
b'a'
```

In RustPython commitID 249f1218
```
>>>>> b'%s' % 'a'
b'a'
>>>>> b'%b' % 'a'
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Unsupported method: decode
>>>>> b'%b' % b'a'
b'a'
>>>>> b'%s' % b'a'
b'b'a''
```

IMHO, making b'%s' to b'%b' will solve this issue.

KipHamiltons:
This issue appears to have a PR fixing it merged

PR #2317, opened 2020-10-31
Fix byte string formatting (#2121)

ChJR:


vm/src/anystr.rs
```diff
@@ -440,4 +441,13 @@ where
             .map_err(|err| vm.new_value_error(err.to_string()))?
             .format(vm, values)
     }
+
+    fn py_bytes_cformat(&self, values: PyObjectRef, vm: &VirtualMachine) -> PyResult<String> {
+        let regex = Regex::new(r"%(?P<spec>[<>=^]?[+\- ]?\d*,?\.?\d*)s").unwrap();
+        let format_string = self.as_utf8_str().unwrap();
+        let format_byte_string = regex.replace_all(format_string, "%${spec}b").to_string();
+        CFormatString::from_str(&format_byte_string)
```
coolreader18, line 449: Maybe instead of running a regex on the format string, you could have a parameter [here](https://github.com/RustPython/RustPython/blob/249f1218965472ad0c781010b65366c6d423eb4e/vm/src/cformat.rs#L319-L323) like `bytes_format: bool` that if true makes `CFormatPreconversor::Str` act like `CFormatPreconversor::Bytes`?

coolreader18 (changes requested):

vm/src/anystr.rs
ChJR: That's the way I first tried. But I found that approach needs to change all of CFormatSpec and CFormatString when I pass bytes_format as parameter. I don't want to change unrelated working code for this change.

I realized why a regex is not recommended in this code, I need other good approaches.

Do I need to use new traits for that structs? I am waiting your advices.

qingshi163: Sorry, I am working on CFormatSpec. It is nearly done but I did not push the PR yet.
I am going to push it now.
My apologize I should push it early.

coolreader18: #2318

ChJR:
After merged #2318, this PR will not needed.

PR #2318, opened 2020-11-01
Reimpl cformat to suit for bytes

qingshi163:


qingshi163:
Fixes #2121

vm/src/cformat.rs
```diff
@@ -412,6 +561,159 @@ impl CFormatPart {
     }
 }
 
+#[derive(Debug, PartialEq)]
+pub(crate) struct CFormatBytes {
+    parts: Vec<(usize, CFormatPart)>,
```
coolreader18, line 566: CFormatPart should probably be generic to allow `String` or `Vec<u8>`

qingshi163: I'm not sure that is necessary, I am using String as Vec<u8> here for bytes. As String stores char as utf-8 so I don't think any cost here.

vm/src/cformat.rs
```diff
@@ -122,10 +120,36 @@ struct CFormatSpec {
...
+        consume_length(iter);
+        let (format_type, format_char) = parse_format_type(iter)?;
+        let precision = if precision.is_some() {
+            precision
+        } else {
+            match format_type {
+                CFormatType::Float(_) => Some(CFormatQuantity::Amount(6)),
+                _ => None,
+            }
+        };
```
youknowone, line 141:
```rust
let precision = precision.or_else(|| match format_type {
   CFormatType::Float(_) => Some(CFormatQuantity::Amount(6)),
   _ => None,
});
```

vm/src/cformat.rs
youknowone: @coolreader18 do you have an example which distinguish bytes and str for the case?

coolreader18: Well bytes can have non-ascii and still be formatted, while a String wouldn't conserve that:

```py
>>> b'\xaa %b' % b'123'
b'\xaa 123'
```

Specifically, the `u8 as char` casting in this PR would map the non-ascii byte `\xaa` to the utf8 character sequence `'ª' == b'\xc2\xaa'`

coolreader18 (approved):
Looks great!!

PR #2320, opened 2020-11-02
Split the ast from the parser, remove compiler dep on parser

coolreader18:
This should improve build times, since now the compiler and the parser can build in parallel, and I think it also just separates concerns a bit more -- the core compiler *just* needs an AST to be able to do its job, and it's up to the user to provide that. I was thinking of making another crate that would essentially just be `vm::compile`, but I wasn't sure what to call it or if would even be necessary, since I'm not sure how many uses there would be for an external crate to use the compiler. Although, I gave pyckitup the ability to compile rustpython bytecode, so maybe it would be worth looking into to have that be another crate.


coolreader18:
Maybe something like `compiler/porcelain`? With the original compiler crate being renamed to `rustpython-compiler-core`, and the porcelain directory having a crate `rustpython-compiler` that has the wrapper stuff that's inside of `vm/src/compile.rs`

youknowone (approved):

PR #2323, opened 2020-11-06
Fully implement buffered IO

coolreader18:
Hoo boy, this took a while


coolreader18:
I *think* the issue with windows file stuff (here and with test_importlib) is that we don't use `_wopen`, we std's OpenOptions, which uses a winapi function. Maybe there is some way to fix the permissions while using OpenOptions, but I'm not sure what it is.

youknowone (approved):

PR #2324, opened 2020-11-08
Fix cell_lock, remove the RwLock<BufferOptions> from bytearray

coolreader18:
@qingshi163 could you take a look at this?

`rustpython` built without threading was panicking with `deadlock: tried to upgradably+sharedly lock a CellRwLock twice` in `ByteArray::get_options()`. The changes in memory.rs are because I thought it was a legit deadlock but it turned out to just be a type in `cell_lock.rs`, and I made the Buffer type for bytearray a separate type that stores the options directly in it, to avoid some overhead with the RwLock inside of `PyByteArray`

### DIFF
--- a/vm/src/builtins/bytearray.rs
+++ b/vm/src/builtins/bytearray.rs
@@ -697,8 +697,8 @@ impl Buffer for ByteArrayBuffer {
         self.bytearray.exports.fetch_sub(1);
     }
 
-    fn get_options(&self) -> BorrowedValue<BufferOptions> {
-        (&self.options).into()
+    fn get_options(&self) -> &BufferOptions {
+        &self.options
     }
 }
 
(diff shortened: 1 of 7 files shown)


vm/src/builtins/bytearray.rs
```diff
@@ -667,40 +663,42 @@ impl Comparable for PyByteArray {
...
-                ..Default::default()
-            }));
-            PyRwLockWriteGuard::downgrade(w)
-        } else {
-            PyRwLockUpgradableReadGuard::downgrade(guard)
-        };
-        PyRwLockReadGuard::map(guard, |x| x.as_ref().unwrap().as_ref()).into()
+        (&self.options).into()
     }
 }
```
qingshi163, line 703: I am thinking maybe simpler way is change the trait fn get_options to return a reference instead the borrow, than we need create the Buffer struct with options for every classes.

qingshi163:
This definitely the better apporach.

PR #2325, opened 2020-11-09
Add string interning

coolreader18:
|:---|---:|---:|---:|---:|
| `./rustpython-norm benchmarks/pystone.py 500000` | 78.248 ± 1.790 | 76.982 | 79.513 | 1.12 ± 0.05 |
| `./rustpython-opt benchmarks/pystone.py 500000` | 70.099 ± 2.543 | 68.301 | 71.897 | 1.00 |


coolreader18:
Ah, crap, I meant to make the rustpython_bytecode::bytecode flattening a separate commit, one sec

coolreader18:
Alright, ready for review now.

vm/src/builtins/code.rs
```diff
@@ -74,16 +75,71 @@ impl Constant for PyConstant {
...
+        let vm = self.0;
+        let ctx = &vm.ctx;
+        let obj = match constant {
+            bytecode::BorrowedConstant::Integer { value } => ctx.new_bigint(value),
+            bytecode::BorrowedConstant::Float { value } => ctx.new_float(value),
+            bytecode::BorrowedConstant::Complex { value } => ctx.new_complex(value),
+            bytecode::BorrowedConstant::Str { value } if value.len() <= 20 => {
+                vm.intern_string(value).into_object()
+            }
+            bytecode::BorrowedConstant::Str { value } => vm.ctx.new_str(value),
```
youknowone, line 120: Does constant interning also affect by string length?

youknowone (commented):
Will you explain a bit about PyStrExact?
I am curious how it works in CPython and how they make benefits in RustPython comparing to bare PyStr.

vm/src/builtins/code.rs
coolreader18: Oh, I think I confused it with [this](http://guilload.com/python-string-interning/#avoiding-large-pyc-files), I think it would be interned still.

coolreader18:
> Would you explain a bit about PyStrExact?

Yeah, so it's a lot easier to deal with interning if you know that everything in the dict is a strict `str`, cause subclasses' `__eq__` or `__hash__` implementations might raise an exception. I might actually make a `PyRefExact<T>` struct though, to make it more general over any type.

youknowone (approved):

PR #2326, opened 2020-11-11
Add ability to use `__trunc__` to `int()`

carbotaniuman:


Lib/test/test_fractions.py
```diff
@@ -459,7 +457,6 @@ def testLargeArithmetic(self):
         )
 
     # TODO: RUSTPYTHON
```
coolreader18, line 459:
```suggestion
```

coolreader18 (approved):
LGTM!

PR #2328, opened 2020-11-11
Fix min test case by doing mroe validation

carbotaniuman:


youknowone (approved):
Looks good as it is. How about `max`? If `min` and `max` has same rule for its arguments, can we have shared code for the argument process for the both function?

carbotaniuman:
I refactored both functions into a single one that changes depending on arguments.

vm/src/function.rs
```diff
@@ -196,12 +196,21 @@ impl FuncArgs {
                 T::arity().end(),
                 given_args,
             )))
-        } else if let Some(k) = self.kwargs.keys().next() {
-            Err(vm.new_type_error(format!("Unexpected keyword argument {}", k)))
         } else {
+            if let Some(err) = self.check_kwargs_empty(vm) {
+                return Err(err);
+            }
```
youknowone, line 202: is this moving related to `#![allow(clippy::suspicious_else_formatting)]`?

carbotaniuman:
Previous test failure looked unrelated, retrying now

coolreader18 (approved):
Great work!

PR #2331, opened 2020-11-16
Add iter, iternext slots

coolreader18:


Lib/test/test_types.py
```diff
@@ -1456,6 +1456,7 @@ def foo():
         self.assertIs(foo(), coro)
         self.assertIs(foo().__await__(), coro)
 
+    @unittest.skip("TODO: RUSTPYTHON, unittest.mock")
```
youknowone, line 1459: How did this work before?

youknowone (approved):

Lib/test/test_types.py
coolreader18: I'm not sure, I assume something to do with `mock` being able to easily mock methods, but us not supporting the machinery it uses to mock slots (or something like that).

PR #2333, opened 2020-11-17
Use lz-fear to get streaming decompression for bytecode

coolreader18:


youknowone:
This PR maybe better with explanation why this is better than before.

coolreader18:
Oh, yeah, this really improves startup times from decoding frozen bytecode:
```
Benchmark #1: ./rustpython-norm -c ""
  Time (mean ± σ):      68.8 ms ±  13.8 ms    [User: 61.4 ms, System: 7.0 ms]
  Range (min … max):    50.7 ms …  95.3 ms    31 runs
 
Benchmark #2: ./rustpython-opt -c ""
  Time (mean ± σ):      22.3 ms ±  11.1 ms    [User: 17.2 ms, System: 5.0 ms]
  Range (min … max):    11.0 ms …  44.7 ms    194 runs
 
Summary
  './rustpython-opt -c ""' ran
    3.08 ± 1.65 times faster than './rustpython-norm -c ""'
```

coolreader18:
It does change the bytecode format from `lz4_raw(bincode(codeobject))` to `lz4_framed(bincode(codeobject))`, but I think that should be fine.

youknowone (approved):

PR #2335, opened 2020-11-17
Return same object if exact type in int, float, complex and fix pow

carbotaniuman:


carbotaniuman:
I can't seem to fix the suspicious else formatting thing.

carbotaniuman:
I tried commenting out code and the if statements are just a normal if with linebreaks in between them. I guess I'll just allow the clippy warning.

coolreader18:
@carbotaniuman I took a look at it too, and I think it is a false positive; any way I could get the lint to stop would have changed the logic. I moved it to the specific function that the lint is from, but if this keeps happening it might make sense to put it on the whole crate, since we use rustfmt anyway

coolreader18 (approved):

PR #2336, opened 2020-11-17
Manually implement a vtable for PyObjectRef

coolreader18:
You can see a lot of the rationale for this in the comment at the top of `pyobjectrc.rs`. Here's the benchmark info:

```
Benchmark #1: ./rustpython-norm benchmarks/pystone.py 50000
  Time (mean ± σ):      6.282 s ±  0.198 s    [User: 6.247 s, System: 0.016 s]
  Range (min … max):    5.911 s …  6.542 s    10 runs
 
Benchmark #2: ./rustpython-opt benchmarks/pystone.py 50000
  Time (mean ± σ):      4.887 s ±  0.171 s    [User: 4.866 s, System: 0.008 s]
  Range (min … max):    4.687 s …  5.283 s    10 runs
 
Summary
  './rustpython-opt benchmarks/pystone.py 50000' ran
    1.29 ± 0.06 times faster than './rustpython-norm benchmarks/pystone.py 50000'
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -161,6 +161,21 @@ jobs:
         run: python -m pip install flake8
       - name: run lint
         run: flake8 . --count --exclude=./.*,./Lib,./vm/Lib  --select=E9,F63,F7,F82 --show-source --statistics
+  miri:
+    name: Run tests under miri
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - uses: actions-rs/toolchain@v1
+        with:
+            profile: minimal
+            toolchain: nightly
+            components: miri
+            override: true
+      - name: Run tests under miri
+        # miri-ignore-leaks because the type-object circular reference means that there will always be
+        # a memory leak, at least until we have proper cyclic gc
+        run: MIRIFLAGS='-Zmiri-ignore-leaks' cargo +nightly miri test -p rustpython-vm -- miri_test
 
   wasm:
     name: Check the WASM package and demo
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1894,6 +1894,7 @@ dependencies = [
  "sha2",
  "sha3",
  "socket2",
+ "static_assertions",
  "statrs",
  "thiserror",
  "thread_local",
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -79,6 +79,7 @@ cfg-if = "0.1.10"
 timsort = "0.1"
 thiserror = "1.0"
 atty = "0.2"
+static_assertions = "1.1"
 
 ## unicode stuff
 unicode_names2 = "0.4"
--- a/vm/src/builtins/code.rs
+++ b/vm/src/builtins/code.rs
@@ -198,7 +198,7 @@ impl PyCode {}
 #[pyimpl]
 impl PyCodeRef {
     #[pyslot]
-    fn tp_new(_cls: PyTypeRef, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+    fn tp_new(_cls: PyTypeRef, vm: &VirtualMachine) -> PyResult<Self> {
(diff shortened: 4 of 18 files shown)


coolreader18:
Wild... specifically when compiling to wasm32 (web or wasi) there's an index out of bounds error in `pytype::new`

coolreader18:
Maybe a miscompilation or something?

carbotaniuman:
Where is the index ooh found? Don't see it in GH Actions

carbotaniuman:
You also shouldn't need to go through ManuallyDrop when ptr cast - just go to MaybeUninit.

coolreader18:
@carbotaniuman it doesn't show up in the CI logs, but running with `cargo build --release --target wasm32-wasi --no-default-features && wasmer run target/wasm32-wasi/release/rustpython.wasm` shows an error. The web demo showed the error stack in the console, but it's specifically something in `pytype::new`, with iterating over `new_type.attributes.read().keys()`, even though `attributes` is empty. You can put debugging prints around it, it's *really weird*:
```rust
{
    let attrs = new_type.attributes.read();
    eprintln("made it here! len {}", attrs.len()); // len 0
    for attr_name in attrs.keys() {
        .update_slot...
    }
    eprintln!("won't get here");
}
```

coolreader18:
Okay I figured it out, I made some assumptions about layout that didn't hold up on wasm, so I flattened the `value` field of `PyInner` into the struct. Now `vtable`, `typeid`, `dict`, `typ`, and `payload` are all fields of the same struct.

coolreader18:
I think we need to reorganize the `pyobject` module, it's got too many types from too many different things and now it doesn't even have `struct PyObject` in it. Maybe we could extract some stuff out into an `abstract` module, or something like that?

coolreader18:
Here's the phenomenon that motivated this:
```rust
pub struct Foo<T: ?Sized> {
    x: u8,
    t: T,
}
pub trait Trait {}
pub fn concrete_offset(foo: &Foo<u16>) -> &u16 {
    &foo.t
}
pub fn dyn_offset(foo: &Foo<dyn Trait>) -> &dyn Trait {
    &foo.t
}
```
```asm
playground::concrete_offset:
	lea	rax, [rdi + 2]
	ret

playground::dyn_offset:
	mov	rdx, rsi
	mov	rcx, qword ptr [rsi + 16]
	mov	rax, rcx
	neg	rax
	and	rax, rcx
	add	rax, rdi
	ret
```
https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=87db3840556bac8f055bd28722d86950

carbotaniuman:
That by all accounts should work though. Does that break if you feed it through wasmer?

coolreader18:
Oh, no I figured that out. That was just a mistake wrt struct layout and alignment.

carbotaniuman:
So it looks like we're just doing tag dispatch across every PyObjectPayload. Wouldn't moving the TypeId back to the stable lead to the same vtable issues every access?

coolreader18:
No, because checking the vtable when casting is an expected operation, but previously just getting the type id required a) looking in the vtable for the offset of the payload, b) calling a vtable method on `PyObjectPayload` `as_any(&self) -> &dyn Any`, and c) calling the `downcast_ref(&self) -> Option<&T>` method which calls the vtable method of `Any` `type_id(&self) -> TypeId`. Now, it just requires a direct access into the object/object vtable to get a u64, compares that u64 with a constant, and returns a concrete offset of the object if that's true or None otherwise.

youknowone:
I am sorry for late review. I recently have other jobs with due, so it is not easy to spend time for this project. And I think these jobs will be finally resolved in early December. This PR is ambitious, so then it will take time to review.
Since you are the member who understand this the best, feel free to merge it first if you need it and don't block other PRs by this.
I will promise to cover this up again when I come back next month.

coolreader18:
@youknowone no worries! Definitely prioritize other stuff if you're busy :). That sounds good, I'm confident in this implementation so I guess I'll merge it now.

PR #2337, opened 2020-11-18
Fix os.stat().st_mode error in windows

owtotwo:
Use oct number like in CPython fileutils.c instead of decimal number leading zero. (e.g.: 0666 -> 0o666)


coolreader18 (approved):
Nice catch! That was kinda a silly mistake :grin: Are there any skipped tests that are fixed by this?

owtotwo:
Sorry, I am not quite sure about this. 😀
Because the master branch did not compile successfully in my development environment, I did not run the tests on my machine.

I just found that the results were inconsistent when using os.stat() on Windows. Then I found this problem after browsing the source code.

coolreader18:
No problem! This is a good fix :)

PR #2339, opened 2020-11-23
Replace FuncArgs.insert with prepend_arg, optimize check_signals

coolreader18:
|:---|---:|---:|---:|---:|
| `./rustpython-norm benchmarks/pystone.py 50000` | 6.816 ± 0.653 | 6.029 | 7.630 | 1.06 ± 0.11 |
| `./rustpython-opt benchmarks/pystone.py 50000` | 6.410 ± 0.294 | 5.937 | 6.853 | 1.00 |

### DIFF
--- a/vm/src/stdlib/signal.rs
+++ b/vm/src/stdlib/signal.rs
@@ -98,15 +98,22 @@ fn _signal_alarm(time: u32) -> u32 {
 }
 
 #[cfg_attr(feature = "flame-it", flame)]
+#[inline(always)]
 pub fn check_signals(vm: &VirtualMachine) -> PyResult<()> {
-    let signal_handlers = match vm.signal_handlers {
-        Some(ref h) => h.borrow(),
+    let signal_handlers = match &vm.signal_handlers {
+        Some(h) => h,
         None => return Ok(()),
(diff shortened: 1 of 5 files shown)


youknowone (approved):

vm/src/stdlib/signal.rs
```diff
@@ -98,15 +98,21 @@ fn _signal_alarm(time: u32) -> u32 {
...
         None => return Ok(()),
     };
 
     if !ANY_TRIGGERED.swap(false, Ordering::Relaxed) {
         return Ok(());
     }
+
+    trigger_signals(&signal_handlers.borrow(), vm)
+}
+#[inline(never)]
```
youknowone, line 114: How about `cold` attribute? this is a question instead of suggestion because i haven't seen any actual usage of `cold` yet. https://doc.rust-lang.org/reference/attributes/codegen.html

coolreader18: That sounds good, I was thinking that it really isn't *that* exceptional of a case, but given that it is (possibly) called for every instruction execution it would make sense to signal that it should very rarely ever happen.

PR #2340, opened 2020-11-23
Remove the need for a label_map in CodeObject

coolreader18:


compiler/src/compile.rs
```diff
@@ -18,11 +18,52 @@ use rustpython_bytecode::bytecode::{self, CallType, CodeObject, Instruction, Lab
...
+                SetupLoop { start, end } => {
+                    *start = label_map[start.0].expect("label never set");
+                    *end = label_map[end.0].expect("label never set");
+                    continue;
+                }
+                SetupFinally { handler } => handler,
+                SetupExcept { handler } => handler,
+                SetupWith { end } => end,
+                SetupAsyncWith { end } => end,
+                _ => continue,
```
youknowone, line 55: If the instruction set is expected to be not changed too frequently, I suggest to explicitly list them up here rather than matching `_`. It will force to read th ecomment up to this match.

youknowone (approved):

PR #2342, opened 2020-11-25
Make `type()` bases arg only take tuples

carbotaniuman:


coolreader18:
You can call `.borrow_value()` on the tuple to get a `&[PyObjectRef]`, and then use that instead of the PyIterable

coolreader18:
Oh I see, cause it has to be `Vec<PyTypeRef>`. Maybe just `.borrow_value().iter().map(typeref tryfromobject).collect::<PyResult<Vec<PyTypeRef>>()?`

coolreader18:
I recently added a `builtins::tuple::PyTupleTyped` type, you could try making the type of `bases: PyTupleTyped<PyTypeRef>`

carbotaniuman:
Updated. Is there a reason you added the `Transmute...` trait instead of just `TryFromObject`?

coolreader18:
`TryFromObject` is implemented for a ton of different types and doesn't have any invariants regarding representation, e.g. if it just required TryFromObject you could have `PyTupleTyped<u8>`, and that obviously wouldn't be safe to cast `PyObjectRef` to `u8`. `TransmuteFromObject` is an `unsafe trait`, so if anyone wants to implement it wrong they have to use `unsafe` to do so :upside_down_face:

coolreader18 (approved):
Thanks for contributing!

PR #2346, opened 2020-11-29
Fix crash on lexing octal escapes bigger than u8::MAX

isidentical:


parser/src/lexer.rs
```diff
@@ -457,7 +457,13 @@ where
                 break;
             }
         }
-        u8::from_str_radix(&octet_content, 8).unwrap() as char
+        match u32::from_str_radix(&octet_content, 8) {
```
coolreader18, line 460: I think you should be able to just `unwrap()` this, since we just verified that it's all octal digits

coolreader18 (approved):

PR #2348, opened 2020-11-29
PEP 498 compatibility: parenthesized expressions for f-strings

isidentical:
Since we are already adding parentheses, we don't need to trim(), this also improves location information for `f'{  b  }`!


coolreader18 (approved):
LGTM!

PR #2350, opened 2020-11-30
Handle raw yield expressions as RHS for augmented/annotated assignments

isidentical:


coolreader18 (approved):
I'm not super familiar with the parser, but this LGTM

PR #2353, opened 2020-12-01
Give the `rustpython` crate a library component

coolreader18:
A possible replacement for native extension modules for the moment.


youknowone (approved):
The broken test looks like irrelevant but need to be checked

PR #2354, opened 2020-12-02
Add injectPython function for notebook

coolreader18:
cc @mireille-raad


wasm/notebook/snippets/python-js-css/javascript.txt
```diff
@@ -1,45 +1,25 @@
...
+            resultDiv.innerHTML += output;
         });
-    
-}
\ No newline at end of file
+        pushNotebook(btn);
+
+        const resultDiv = document.createElement('div');
+        resultDiv.classList.add('result');
+        pushNotebook(resultDiv);
```
mireille-raad, line 23: oh wow. is is really neat and MUCH better than the hack around that I had.

mireille-raad:
oh wow those are some serious improvements, especially the parts for handling parameters, execution and displaying results.

1/ Thank you for creating the helpers for p, h1->h6. That is really neat.

2/ are `rp.vmStore.init();` and `pyvm.addToScope()`, `pyvm.exec(code)` something that you recently added?  is `rp.pyExec` still there?

3/ What do you use for editing and formatting code. I feel that we always have some spacing/formatting diff especially with `css` and `js` , it is probably because we have a different setup. do you use a code formatter? if yes, which one. I am happy to replicate your setup. if no, I  will stop being lazy and manually format code.

4/ My last comment, I am not sure about what I was thinking when I wrote the calculator example. The returned text is way too silly. 
Can you update the example code to just be a simple calculator? I think the idea is more straightforward... You can have two inputs and the output would be a simple addition. I think it drives home the point and the steps are included in markdown.
This might also be a neat way to demonstrate passing multiple arguments.

Otherwise, I am excited to have this merged.

wasm/notebook/snippets/python-js-css/javascript.txt
```diff
@@ -1,45 +1,25 @@
-// Javascript code go here
+// Javascript code goes here
 
-// Javascript code go here
+injectPython({
+    // injectPython functions take the positional arguments as
```
mireille-raad, line 4: thanks for adding a decent way to handle arguments. should we add a note about this in the README for the notebook?

mireille-raad (approved):

coolreader18:
2: no, they've been there for a while; they exist alongside rp.pyExec. We make use of them for the demo to have a different scope for the codemirror editor and for the terminal.

3: Oh, yeah, I use prettier, I think there should be a `.prettierrc` in the `wasm/` directory that the prettier cli or editor extensions should pick up.

4: yeah, for sure.

PR #2357, opened 2020-12-06
fix memoryview compare

hyperbora:


youknowone:
@qingshi163 I think you are one of the pro of memoryview. Could you review this PR please?

vm/src/builtins/bytes.rs
```diff
@@ -531,6 +531,13 @@ impl Comparable for PyBytes {
...
+        } else if other.isinstance(&vm.ctx.types.memoryview_type) {
+            return Err(vm.new_type_error(format!(
+                "'{}' not supported between instances of '{}' and '{}'",
+                op.operator_token(),
+                zelf.class().name,
+                other.class().name
+            )));
         } else {
             zelf.inner.cmp(other, op, vm)
         })
```
qingshi163, line 543: The CPython's logic is very strange here, memoryview could do == or != here with bytes, but raise type error when doing <, >, <=, >=. and bytearray just accept everything.
This code does not match the cpython's logic for the following compare:
```
bytes (==,!=) memoryview
memoryview (<,<=,>,>=)  bytearray
```

hyperbora:
Thank you for your review. I'm not sure if I edited it as you said.

qingshi163:
I don't know any reason CPython has a back door for ord compare between memoryview and bytearray(not bytes). Should we follow? @coolreader18 
Otherwise this code fix at lease some of the case.

coolreader18:
Huh, that's really weird behavior. I think if there's no test cases that test for that (or even if there are, honestly) it's fine to ignore that.

coolreader18 (approved):
Thank you for contributing!

PR #2359, opened 2020-12-06
Fix return types

youknowone:


vm/src/builtins/int.rs
```diff
@@ -570,11 +570,11 @@ impl PyInt {
     }
 
     #[pymethod(name = "as_integer_ratio")]
-    fn as_integer_ratio(&self, vm: &VirtualMachine) -> PyResult {
-        Ok(vm.ctx.new_tuple(vec![
+    fn as_integer_ratio(&self, vm: &VirtualMachine) -> PyObjectRef {
+        vm.ctx.new_tuple(vec![
             vm.ctx.new_bigint(&self.value),
```
coolreader18, line 575: This could maybe return `(PyIntRef, u8)`, just `(zelf, 1)`

coolreader18 (approved):

vm/src/builtins/int.rs
youknowone: it causes test_long error for bool type

PR #2363, opened 2020-12-09
wasm/notebook: mobile and tablet responsive layout

mireille-raad:
This PR adds responsive layout to `wasm/notebook` on phone and ipad/tablets.

The notebook looked decent/good on phone and iPad (especially when the iPad is flipped horizontally).  Since I think those devices would be mostly used for reading code, not writing it, imho the layout is good enough for now (It can be better on phones tho). 

Here are some screenshots from tests on real devices. There is a problem with Chrome on Android (will open an issue).

## iPad (safari)  

4031D3DD-5347-4ABE-A701-DDB95F7BFF9A

## Android (Firefox)

Screenshot_20201208_193255

## Android (Chrome)

needs fixing, will create an issue.

Screenshot_20201208_193819_com android chrome

### DIFF
--- a/wasm/notebook/src/index.ejs
+++ b/wasm/notebook/src/index.ejs
@@ -12,7 +12,7 @@
 <body>
 
 <div class="d-flex">
-    <div class="header item-center">RustPython 🐍 😱 🤘</div>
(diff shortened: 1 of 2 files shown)


coolreader18 (approved):
Looks great!

PR #2364, opened 2020-12-09
Ctypes implementation

rodrigocam:
@darleybarreto and I are implementing the `ctypes` module. This an initial implementation for review. At this stage, we are focusing on Linux platforms to in the future extend to other platforms like `ctypes` from `cpython` did.


coolreader18:
I just pushed a commit to get stuff compiling. Excited to review this!

darleybarreto:
One thing to note is that some things are directly translation from PyPy, others from CPython's. Also there's that difference I said on Gitter, regarding declaring a class like:
```python
class Custom(_SimpleCData):
   pass
... 
AttributeError: class must define a '_type_' attribute
```
which here will succeed.
Besides this, I've tried to stay as close as CPython's behavior as possible.

vm/src/stdlib/ctypes/basics.rs
```diff
@@ -0,0 +1,316 @@
...
+    }
+}
+
+// This Trait is the equivalent of PyCData_Type on tp_base for
+// Struct_Type, Union_Type, PyCPointer_Type
+// PyCArray_Type, PyCSimple_Type, PyCFuncPtr_Type
+#[pyclass(module = "_ctypes", name = "_CData")]
+pub struct PyCData {
+    _objects: AtomicCell<Vec<PyObjectRef>>,
+    _buffer: PyRwLock<Vec<u8>>,
```
coolreader18, line 275: I think the buffer is supposed to be a raw pointer and a length, so that you can mutate the original data it if you want to.

coolreader18 (commented):
Here's what I have so far; I think it would be better to use more of libffi's `middle` API in `ctypes::function`, just to avoid as much `unsafe` as possible for something like ctypes :upside_down_face: . Now that the functions are copied from cpython, it should be easier to refactor into something more idiomatic.

vm/src/stdlib/mod.rs
```diff
@@ -123,6 +124,9 @@ pub fn get_module_inits() -> HashMap<String, StdlibInitFunc> {
     #[cfg(any(unix, windows, target_os = "wasi"))]
     modules.insert(os::MODULE_NAME.to_owned(), Box::new(os::make_module));
 
+    #[cfg(any(unix, windows, target_os = "wasi"))]
+    modules.insert("_ctypes".to_owned(), Box::new(ctypes::make_module));
```
youknowone, line 128: instead of duplicating same cfg, it can be grouped like:
```rust
    #[cfg(any(unix, windows, target_os = "wasi"))]
    {
        modules.insert(os::MODULE_NAME.to_owned(), Box::new(os::make_module));
        modules.insert("_ctypes".to_owned(), Box::new(ctypes::make_module));
    }
```

coolreader18: Although I don't know if ctypes would be available on wasi, I don't think it has dll functionality yet

vm/src/stdlib/ctypes/basics.rs
```diff
@@ -0,0 +1,316 @@
...
+use crate::stdlib::ctypes::dll::dlsym;
+
+use crossbeam_utils::atomic::AtomicCell;
+
+pub fn compare_classes(
+    cls_a: &PyTypeRef,
+    cls_b: &PyTypeRef,
+    vm: &VirtualMachine,
+) -> PyResult<bool> {
+    Ok(vm.bool_eq(cls_a.as_object(), cls_b.as_object())?)
```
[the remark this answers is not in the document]
darleybarreto: No reason in particular, I wasn't aware of `a.is(b)`.

vm/src/stdlib/ctypes/basics.rs
darleybarreto: Something like
```rust
struct RawBuffer{
    inner: *mut u8,
    size: usize
}
```
?

vm/src/stdlib/mod.rs
darleybarreto: Would libffi work?

coolreader18: No, wasm doesn't really have any sort of support for loading another wasm module at runtime

vm/src/stdlib/ctypes/function.rs
```diff
@@ -88,49 +84,44 @@ pub fn str_to_type(ty: &str) -> *mut ffi_type {
...
-            &mut u8::try_from_object(vm, obj)? as *mut _ as *mut c_void
+            arg(&u8::try_from_object(vm, obj)?)
         }
         pointer => {
-            usize::try_from_object(vm, obj).unwrap() as *mut c_void
+            arg(&usize::try_from_object(vm, obj).unwrap())
         }
         void => {
-            ptr::null_mut()
+            arg(&ptr::null::<usize>())
```
darleybarreto, line 124: I've replace this for 
```rust
ptr::null::<c_void>()
```
which seemed more "correct".

vm/src/stdlib/ctypes/basics.rs
```diff
@@ -252,11 +242,17 @@ struct PyCDataBuffer {
 // This trait will be used by all types
 impl Buffer for PyCDataBuffer {
     fn obj_bytes(&self) -> BorrowedValue<[u8]> {
-        PyRwLockReadGuard::map(self.data.borrow_value(), |x| x.as_slice()).into()
+        PyRwLockReadGuard::map(self.data.borrow_value(), |x| unsafe {
+            slice::from_raw_parts(x.inner, x.size)
```
darleybarreto, line 246: I'm not sure if this is copying things

vm/src/stdlib/ctypes/basics.rs
```diff
@@ -265,14 +261,21 @@ impl Buffer for PyCDataBuffer {
         &self.options
     }
 }
+pub struct RawBuffer {
```
darleybarreto, line 264: @coolreader18 is this what you have in mind?

vm/src/stdlib/ctypes/function.rs
```diff
@@ -88,49 +84,44 @@ pub fn str_to_type(ty: &str) -> *mut ffi_type {
...
-            &mut f64::try_from_object(vm, obj)? as *mut _ as *mut c_void
+            arg(&f64::try_from_object(vm, obj)?)
         }
         c_uchar => {
-            &mut u8::try_from_object(vm, obj)? as *mut _ as *mut c_void
+            arg(&u8::try_from_object(vm, obj)?)
         }
         pointer => {
-            usize::try_from_object(vm, obj).unwrap() as *mut c_void
+            arg(&usize::try_from_object(vm, obj).unwrap())
```
darleybarreto, line 121: Also replace this with
```rust
arg(&(usize::try_from_object(vm, obj)? as *mut usize as *mut c_void))
```

vm/src/stdlib/ctypes/basics.rs
coolreader18: Yeah, it is.

vm/src/stdlib/ctypes/array.rs
```diff
@@ -20,3 +79,141 @@ impl PyValue for PyCArray {
...
+        let buffer = try_buffer_from_object(vm, &obj)?;
+
+        let res = if self._type_ == "u" {
+            vm.new_pyobj(
+                if cfg!(windows) {
+                    WideString::from_vec(
+                        buffer
+                            .obj_bytes()
+                            .chunks(4)
+                            .map(|c| NativeEndian::read_u32(c))
```
coolreader18, line 169: You can use `u32::from_ne_bytes` for this; also, there's already a byteorder dependency in vm/Cargo.toml

coolreader18: Also, what function is this supposed to be? CArray doesn't have a value property from what I can see

darleybarreto: > You can use `u32::from_ne_bytes` for this; also, there's already a byteorder dependency in vm/Cargo.toml

Oh boy, I missed that one, I've found the `bt`and `le` counterparts. I can simply ignore the whole `ByteOrder` crate, then.

> Also, what function is this supposed to be? CArray doesn't have a value property from what I can see

These properties come from it's metaclass (`PyCArrayType_Type`),  which are `CharArray_getsets` and `WCharArray_getsets`

vm/src/stdlib/ctypes/basics.rs
```diff
@@ -226,28 +252,25 @@ impl<'a> BorrowValue<'a> for PyCData {
...
 
 // This trait will be used by all types
-impl Buffer for PyCDataBuffer {
+impl<T> Buffer for PyCBuffer<T>
+where
+    for<'a> T: PyValue + fmt::Debug + BorrowValue<'a> + BorrowValueMut<'a>,
+{
     fn obj_bytes(&self) -> BorrowedValue<[u8]> {
+        // @TODO: This is broken
         PyRwLockReadGuard::map(self.data.borrow_value(), |x| unsafe {
```
darleybarreto, line 274: This need to be fixed, "casting" the associated type somehow

Pluriscient:
What are the major blockers for this PR? I think I want to give this a try :)

darleybarreto:
> What are the major blockers for this PR? I think I want to give this a try :)

Unfortunately there are several problems with the state of things and I haven't been able to continue working on this (and will not be able to do so for a while). The main problem is how it behaves differently from CPython's and PyPy's, a meta class should be added for Struct, CSimpleType, Pointer, ... (all but CFuntion), so that you can do `c_int * 10` and return a proper array. In the current impl `type(CSimpleType) == type`, which does not has a `__mul__` as required for a `c_int`, for instance.

darleybarreto:
@youknowone I tried to sync with master a while ago, but there was a problem and I got too busy to look at that again, so thanks for that! Are you planning into work on this? If so, I think there are some tests I ported from CPython that I altered just to see if the implementation was working-ish. 

To be fully compatible to CPython and PyPy, the types must have a metaclass that's not `PyType`. I tried to work something there, but I couldn't find a solution. This happens because a base type has to implement some ops at type level, eg, `c_int * 8` should call `__mul__` from its meta (a yet non-existing class `PrimitiveTypeMeta` which is a subclass of `PyType`).

I am happy to help with discussion and pointers, but I'm afraid I don't have much time to code :/

youknowone:
@darleybarreto Hi, thank you for the comment. I am not planning for working on ctypes at the moment, but looking for a way to merge it as it is for future contributors - because you already worked really a lot of it - if it doesn't break things too much. How do you think about it?

darleybarreto:
You mean merging as is?

youknowone:
yes, if we can manage it not to break CI

darleybarreto:
I think that we need to do a few things to help future contributions in this code before merging:

1. Refactor types structure.
2. Documentation.
3. A thorough review to improve code quality.

Undoubtedly `1` is the major blocker here. Since I don't know much of the internals of RustPython, I wasn't able to solve this problem. Perhaps you can give me pointers and I can work it out. PyPy does is basically this way:
```python
class _CDataMeta(type):
    ...
    def __mul__(self, length):
        return create_array_type(self, length)
    ...

class SimpleType(_CDataMeta):
    ...

class SimpleCData(..., metaclass=SimpleType):
    ...
```
I couldn't find a way to make `type(SimpleCData)` be `SimpleType` instead of `type`.

Point `2` is extremely important since this code is basically too experimental and not that much well written. I could gladly do it based on PyPy, CPython and documentations.
Point `3` is somewhat important because I am not a very experienced Rust programmer, so if anyone could help me writing idiomatic and **performant** Rust, I would appreciate!

youknowone:
Thank you for detailed explanation. The metaclass issue is recently solved. Here is the test:

```sh
$ cat t.py 
class _CDataMeta(type):
    def __mul__(self, length):
        return create_array_type(self, length)

class SimpleType(_CDataMeta):
    pass

class SimpleCData(object, metaclass=SimpleType):
    pass


print(type(SimpleCData))
assert type(SimpleCData) == SimpleType
$ cargo run t.py 
    Finished dev [unoptimized + debuginfo] target(s) in 0.06s
     Running `target/debug/rustpython t.py`
<class '__main__.SimpleType'>
```

I will look in the code for 3

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+    vm: &VirtualMachine,
+) -> PyResult<PyRef<PyCArray>> {
+    if let Some(outer_type) = cls.get_attr("_type_") {
+        let length = length as usize;
+        if let Ok(_type_) = vm.get_attribute(outer_type.clone(), "_type_") {
+            let itemsize = get_size(_type_.downcast::<PyStr>().unwrap().to_string().as_str());
+
+            if length.checked_mul(itemsize).is_none() {
+                Err(vm.new_overflow_error("Array size too big".to_string()))
+            } else {
```
youknowone, line 121:
```suggestion
            length.checked_mul(itemsize).unwrap_or_else(|| Err(vm.new_overflow_error("Array size too big".to_owned()))?;
```

then the whole block will use one-level less indent

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+                Ok(PyCArray {
+                    _type_: new_simple_type(Either::A(&outer_type), vm)?.into_ref(vm),
+                    _length_: length,
+                    _buffer: PyRwLock::new(RawBuffer {
+                        inner: Vec::with_capacity(length * itemsize).as_mut_ptr(),
```
youknowone, line 126: I don't think this line is safe. The new `Vec` will be removed right after this line and the given pointer will be a dangling pointer.

vm/src/stdlib/ctypes/basics.rs
youknowone: The safe expression of this type in Rust is `Box<[u8]>` which is exactly equivalent to this struct. Is this intended work to avoid somethiing like ownership check? I think there must be a better way.

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+
+    0
+}
+
+pub fn make_array_with_lenght(
+    cls: PyTypeRef,
+    length: usize,
+    vm: &VirtualMachine,
+) -> PyResult<PyRef<PyCArray>> {
+    if let Some(outer_type) = cls.get_attr("_type_") {
```
youknowone, line 114:
```suggestion
    let outer_type = cls.get_attr("_type_").unwrap_or_else(|| Err(vm.new_attribute_error("class must define a '_type_' attribute".to_string()))?;
```
and so on

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+) -> PyResult<()> {
+    if !obj.clone_class().issubclass(PyCData::static_type()) {
+        let value = PyCDataMethods::from_param(zelf._type_.clone(), obj, vm)?;
+
+        let v_buffer = try_buffer_from_object(vm, &value)?;
+        let v_buffer_bytes = v_buffer.obj_bytes_mut();
+
+        dst_buffer[idx..idx + offset].copy_from_slice(&v_buffer_bytes[0..]);
+        Ok(())
+    } else if vm.isinstance(&obj, &zelf._type_.clone_class())? {
```
youknowone, line 156:
```suggestion
    } else if vm.isinstance(&obj, zelf._type_.class())? {
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+
+fn set_array_value(
+    zelf: &PyRef<PyCArray>,
+    dst_buffer: &mut BorrowedValueMut<[u8]>,
+    idx: usize,
+    offset: usize,
+    obj: PyObjectRef,
+    vm: &VirtualMachine,
+) -> PyResult<()> {
+    if !obj.clone_class().issubclass(PyCData::static_type()) {
```
youknowone, line 148:
```suggestion
    if !obj.class().issubclass(PyCData::static_type()) {
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+        let value = PyCDataMethods::from_param(zelf._type_.clone(), obj, vm)?;
+
+        let v_buffer = try_buffer_from_object(vm, &value)?;
+        let v_buffer_bytes = v_buffer.obj_bytes_mut();
+
+        dst_buffer[idx..idx + offset].copy_from_slice(&v_buffer_bytes[0..]);
```
youknowone, line 154:
```suggestion
        dst_buffer[idx..idx + offset].copy_from_slice(&v_buffer_bytes[..]);
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+
+    assert!(step != 0);
+    assert!(step >= -isize::MAX);
+
+    let start = slice
+        .start
+        .clone()
+        .map_or(Ok(0), |o| isize::try_from_object(vm, o))?;
+
+    let stop = if slice.stop.is(&vm.ctx.none()) {
```
youknowone, line 193:
```suggestion
    let stop = if vm.is_none(slice.stop) {
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+
+impl PyCDataMethods for PyCArray {
+    fn from_param(
+        zelf: PyRef<Self>,
+        value: PyObjectRef,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyObjectRef> {
+        if vm.isinstance(&value, PyCArray::static_type())? {
+            return Ok(value);
+        };
```
youknowone, line 252:
```suggestion
        }
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+    flags(BASETYPE),
+    with(BufferProtocol, PyCDataFunctions, PyCDataMethods)
+)]
+impl PyCArray {
+    #[pyslot]
+    fn tp_new(cls: PyTypeRef, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+        let length = match vm.get_attribute(cls.as_object().to_owned(), "_length_") {
+            Ok(ref length_obj) => {
+                if let Ok(length_int) = length_obj.clone().downcast_exact::<PyInt>(vm) {
+                    if length_int.as_bigint().sign() == Sign::Minus {
```
youknowone, line 290:
```suggestion
                    if length_int.as_bigint().is_negative() {
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+}
+
+#[pyimpl(
+    flags(BASETYPE),
+    with(BufferProtocol, PyCDataFunctions, PyCDataMethods)
+)]
+impl PyCArray {
+    #[pyslot]
+    fn tp_new(cls: PyTypeRef, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+        let length = match vm.get_attribute(cls.as_object().to_owned(), "_length_") {
```
youknowone, line 287:
```suggestion
        let length = match vm.get_attribute(cls.as_object().to_owned(), "_length_").map_err(|e| vm
                        .new_type_error("The '_length_' attribute must be an integer".to_string()))? {
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+            }
+        }?;
+
+        make_array_with_lenght(cls, length, vm)
+    }
+
+    #[pymethod(magic)]
+    pub fn init(zelf: PyRef<Self>, value: OptionalArg, vm: &VirtualMachine) -> PyResult<()> {
+        value.map_or(Ok(()), |value| {
+            let value_lenght = vm.obj_len(&value)?;
```
youknowone, line 317: every lenght -> length

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+            let bytes = buffer.obj_bytes();
+
+            let bytes_inner = if let Some((last, elements)) = bytes.split_last() {
+                if *last == 0 {
+                    elements.to_vec()
+                } else {
+                    bytes.to_vec()
+                }
+            } else {
+                vec![0; 0]
```
youknowone, line 388:
```suggestion
                vec![]
```

or

```suggestion
                Vec::new()
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+        let new_value = try_buffer_from_object(vm, &value)?;
+        let new_size = new_value.get_options().len;
+
+        // byte string zelf._type_ == "c"
+        if new_size > my_size {
+            Err(vm.new_value_error("byte string too long".to_string()))
+        } else {
+            let mut borrowed_buffer = my_buffer.obj_bytes_mut();
+            let src = new_value.obj_bytes();
+            borrowed_buffer[0..new_size].copy_from_slice(&src);
```
youknowone, line 490:
```suggestion
            borrowed_buffer[..new_size].copy_from_slice(&src);
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+
+    #[pymethod(name = "__len__")]
+    fn len(&self) -> usize {
+        self._length_
+    }
+
+    #[pymethod(magic)]
+    fn getitem(
+        zelf: PyRef<Self>,
+        k_or_idx: Either<isize, PyRef<PySlice>>,
```
youknowone, line 503: `rustpython_vm::sliceable::SequenceInex` type fits here

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+                    }
+                    Ok(())
+                }
+            }
+        }
+    }
+}
+
+impl PyCDataFunctions for PyCArray {
+    fn size_of_instances(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
```
youknowone, line 594:
```suggestion
    fn size_of_instances(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<usize> {
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+
+    fn alignment_of_instances(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
+        PyCDataFunctions::alignment_of_instances(zelf._type_.clone(), vm)
+    }
+
+    fn ref_to(
+        zelf: PyRef<Self>,
+        offset: OptionalArg,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyObjectRef> {
```
youknowone, line 612:
```suggestion
    ) -> PyResult<usize> {
```

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+            Err(vm.new_index_error("offset out of bounds".to_string()))
+        } else {
+            let guard = zelf.borrow_value();
+            let ref_at: *mut u8 = unsafe { guard.inner.add(off_set) };
+
+            Ok(vm.new_pyobj(ref_at as *mut _ as *mut usize as usize))
+        }
+    }
+
+    fn address_of(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
```
youknowone, line 627:
```suggestion
    fn address_of(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<usize> {
```

vm/src/stdlib/ctypes/basics.rs
```diff
@@ -0,0 +1,507 @@
...
+                        Err(vm.new_value_error(format!(
+                            "Buffer size too small ({} instead of at least {} bytes)",
+                            cls_size,
+                            opts.len + (offset_int as usize)
+                        )))
+                    } else if let Some(mut buffer) = buffer.as_contiguous_mut() {
+                        // @TODO: Is this copying?
+
+                        let buffered = if copy {
+                            unsafe { slice::from_raw_parts_mut(buffer.as_mut_ptr(), buffer.len()) }
```
youknowone, line 136: this is not a copy

darleybarreto:
Oh, I'm sorry, I meant to do the same thing in the Rust side. CPython does it in the C side of things by manually filling the type:
```c
#define MOD_ADD_TYPE(TYPE_EXPR, TP_TYPE, TP_BASE) \
    do { \
        PyTypeObject *type = (TYPE_EXPR); \
        Py_SET_TYPE(type, TP_TYPE); \
        type->tp_base = TP_BASE; \
        if (PyModule_AddType(mod, type) < 0) { \
            return -1; \
        } \
    } while (0)

MOD_ADD_TYPE(&Simple_Type, &PyCSimpleType_Type, &PyCData_Type);
```
Here `PyCSimpleType_Type` is the meta and `PyCData_Type` is the base class.

vm/src/stdlib/ctypes/basics.rs
```diff
@@ -0,0 +1,507 @@
...
+            "expected {} instance instead of {}",
+            cls.name,
+            value.class().name
+        )))
+    }
+}
+#[pyimpl]
+pub trait PyCDataFunctions: PyValue {
+    #[pymethod]
+    fn size_of_instances(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<PyObjectRef>;
```
youknowone, line 186:
```suggestion
    fn size_of_instances(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult;
```

`PyResult` == `PyResult<PyObjectRef>`

vm/src/stdlib/ctypes/basics.rs
youknowone: no, it isn't. slice is a sort of view

vm/src/stdlib/ctypes/basics.rs
```diff
@@ -0,0 +1,507 @@
...
+}
+
+pub fn alignment(tp: Either<PyTypeRef, PyObjectRef>, vm: &VirtualMachine) -> PyResult {
+    match tp {
+        Either::A(type_) if type_.issubclass(PySimpleType::static_type()) => {
+            let zelf = new_simple_type(Either::B(&type_), vm)?;
+            PyCDataFunctions::alignment_of_instances(zelf.into_ref(vm), vm)
+        }
+        Either::B(obj) if obj.has_class_attr("alignment_of_instances") => {
+            let size_of = vm.get_attribute(obj, "alignment_of_instances").unwrap();
```
youknowone, line 471:
```suggestion
            let size_of = vm.get_attribute(obj, "alignment_of_instances")?;
```
is this `unwrap()` intended?

vm/src/stdlib/ctypes/function.rs
```diff
@@ -0,0 +1,438 @@
...
+
+    #[pyproperty(name = "_restype_")]
+    fn restype(&self, _vm: &VirtualMachine) -> PyObjectRef {
+        unsafe { &*self._restype_.as_ptr() }.clone()
+    }
+
+    #[pyproperty(name = "_argtypes_", setter)]
+    fn set_argtypes(&self, argtypes: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        if vm.isinstance(&argtypes, &vm.ctx.types.list_type).is_ok()
+            || vm.isinstance(&argtypes, &vm.ctx.types.tuple_type).is_ok()
```
youknowone, line 282:
```suggestion
        if vm.isinstance(&argtypes, &vm.ctx.types.list_type).and_then(|_| vm.isinstance(&argtypes, &vm.ctx.types.tuple_type)).map_err(|e| vm.new_type_error(format!(
                "_argtypes_ must be a sequence of types, {} found.",
                argtypes.to_string()
            )))? {
```

youknowone:
by watching `rustpython_vm::builtins::pytype::new`, I feel like that'd be simply done by replacing `typ` field in `PyObject`, but not sure it actually will be that easy.
Because I don't understand this patch that much, would you make specific place that you need those type stuff? Then I will fill the right code there.

darleybarreto:
> would you make specific place in this PR that you need those type stuff?

Let me get the CPython definitions as example
```c
MOD_ADD_TYPE(&Struct_Type, &PyCStructType_Type, &PyCData_Type);
MOD_ADD_TYPE(&Union_Type, &UnionType_Type, &PyCData_Type);
MOD_ADD_TYPE(&PyCPointer_Type, &PyCPointerType_Type, &PyCData_Type);
MOD_ADD_TYPE(&PyCArray_Type, &PyCArrayType_Type, &PyCData_Type);
MOD_ADD_TYPE(&Simple_Type, &PyCSimpleType_Type, &PyCData_Type);
MOD_ADD_TYPE(&PyCFuncPtr_Type, &PyCFuncPtrType_Type, &PyCData_Type);
```
Here we have `MOD_ADD_TYPE(&Class, &MetaClass, &BaseClass)` right? Looking at `Simple_Type` (in RustPython this is `PySimpleType` in `primitive.rs`), we see that its metaclass implements the `__mul__`, so we can have `c_int * 8` just fine, where `c_int` is subclass of `Simple_Type`. I used a trait `PyCDataSequenceMethods` in `basics.rs` to implement the `__mul__`, but  doing `impl PyCDataSequenceMethods for PySimpleTypeMeta {}` in `primitive.rs` doesn't work because `PySimpleTypeMeta` is not the metaclass of `PySimpleType`.

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+            "L" | "Q" => u64
+            "f" => f32
+            "d" | "g" => f64
+            "?" | "B" => u8
+            "P" | "z" | "Z" => usize
+        )
+    }
+}
+
+fn slice_adjust_size(length: isize, start: &mut isize, stop: &mut isize, step: isize) -> isize {
```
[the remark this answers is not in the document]
darleybarreto: They have some resemblance, but are different nonetheless.

vm/src/stdlib/ctypes/array.rs
```diff
@@ -0,0 +1,633 @@
...
+use crate::VirtualMachine;
+
+use crate::stdlib::ctypes::basics::{
+    default_from_param, generic_get_buffer, get_size, BorrowValue as BorrowValueCData,
+    BorrowValueMut, PyCData, PyCDataFunctions, PyCDataMethods, PyCDataSequenceMethods, RawBuffer,
+};
+use crate::stdlib::ctypes::pointer::PyCPointer;
+use crate::stdlib::ctypes::primitive::{new_simple_type, PySimpleType};
+
+macro_rules! byte_match_type {
```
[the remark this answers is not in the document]
darleybarreto: I'm not sure if I understand what you meant. Should I move the `byte_match_type` macro inside `byte_to_pyobj`?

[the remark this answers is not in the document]
youknowone: let me clean up this file a bit. that will be easier and you might see the diff.

darleybarreto:
I'm not sure why, but I cannot comment on the following reply
> The safe expression of this type in Rust is Box<[u8]> which is exactly equivalent to this struct. Is this intended work to avoid somethiing like ownership check? I think there must be a better way.

The idea of `RawBuffer` is to be a "view" of the external buffer, so when needed, we copy the bytes from that buffer. I think there's one case we need to copy data and another where we need a view.

vm/src/stdlib/ctypes/array.rs
darleybarreto: Hm, that's new right? I couldn't find it in my local repo.

darleybarreto: Like this, right (inside `sliceable.rs`)?
```rust
fn get_item(
    &self,
    vm: &VirtualMachine,
    needle: PyObjectRef,
    owner_type: &'static str,
) -> PyResult<Either<Self::Item, Self::Sliced>> {
    let needle = SequenceIndex::try_from_object_for(vm, needle, owner_type)?;
    match needle {
        SequenceIndex::Int(value) => {
            let pos_index = self.wrap_index(value).ok_or_else(|| {
                vm.new_index_error(format!("{} index out of range", owner_type))
            })?;
            Ok(Either::A(self.do_get(pos_index)))
        }
        SequenceIndex::Slice(slice) => Ok(Either::B(self.get_slice_items(vm, &slice)?)),
    }
}
```

vm/src/stdlib/ctypes/array.rs
darleybarreto: Will `usize` be converted to a Python `int`?

vm/src/stdlib/ctypes/basics.rs
darleybarreto: Here
```rust
let buffered = if copy {
    unsafe { slice::from_raw_parts_mut(buffer.as_mut_ptr(), buffer.len()) }
        .as_mut_ptr()
} else {
    buffer.as_mut_ptr()
};
```
The idea is to copy the bytes from the buffer if `copy`, or return a view otherwise. Should I use `to_contiguous` to copy it?

vm/src/stdlib/ctypes/array.rs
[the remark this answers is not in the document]
youknowone: here is: https://github.com/RustPython/RustPython/pull/2364/commits/d82ba1e9933b5f85c49ba095c8a52838e22d4180

vm/src/stdlib/ctypes/array.rs
youknowone: Yes, exactly

vm/src/stdlib/ctypes/array.rs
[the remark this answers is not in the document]
darleybarreto: Oh, I see it now. I wasn't aware of this. Thanks!

youknowone:
What happens when the buffer is destroyed? Does it deletes only the pointer(as a reference) or also deletes the data it contains (as an owner)? If it does both, then does it need a flag to distinguish them?

vm/src/stdlib/ctypes/array.rs
youknowone: Yes, when it eventually returned for `#[pyfunction]`, `#[pymethod]` or `#[pyproperty]`.

darleybarreto:
I forgot where these two functions are used, but based on the docs:
[`from_buffer_copy`](https://github.com/python/cpython/blob/2780df478129c4f1042df2868b6adf2bf507cd6e/Modules/_ctypes/_ctypes.c#L678): `C.from_buffer_copy(object, offset=0) -> C instance\ncreate a C instance from a readable buffer`
[`from_buffer`](https://github.com/python/cpython/blob/2780df478129c4f1042df2868b6adf2bf507cd6e/Modules/_ctypes/_ctypes.c#L598): `C.from_buffer(object, offset=0) -> C instance\ncreate a C instance from a writeable buffer`

youknowone:
I am not sure it exactly fit, but it sounds like a `Cow<[u8]>`

https://doc.rust-lang.org/std/borrow/enum.Cow.html

youknowone:
If it doesn't fit to rust lifetime, creating a similar enum will fit.

```rust
pub enum RawBuffer {
    Borrowed(*mut [u8]),
    Owned(Box<[u8]>),
}
```

darleybarreto:
Yeah, I think its good!

darleybarreto:
Yeah, I think it works!

darleybarreto:
Quick question, how can I add your suggestions all at once in a single commit?

youknowone:
I never used it myself before, but `Add suggetion to batch` button exist.

image

But I don't think it will work well. I suggested one side of code but didn't for the other side. So accepting suggestions will break the build.

darleybarreto:
Yeah, that button is not available for me 😞

youknowone:
@darleybarreto this must be what you are looking for, right?

```python
>>>>> import ctypes
>>>>> ctypes._SimpleCData
<class '_ctypes._SimpleCData'>
>>>>> type(ctypes._SimpleCData)
<class '_ctypes.CSimpleType'>
>>>>> 
```

darleybarreto:
YES! That's the very thing! I called `_SimpleTypeMeta` internally, but I will stick with the CPython naming so we can be similar too.

youknowone:
I just made a test code in commit https://github.com/RustPython/RustPython/pull/2364/commits/efabe9b1b460ff8e2055419a6f5c51856f7c42b7

You can adapt it to any type you actually need to.

And sorry for accidental push-force. that's not what I intended at this time but misoperated :|

darleybarreto:
Don't worry. Thanks for helping me on this :)

darleybarreto:
@youknowone Could you kindly help me with something else? How can I fetch your changes in this MR?

youknowone:
I pushed to your branch. So if you named your repository as `origin`, like when result of `git remote -v` is `origin       https://github.com/rodrigocam/RustPython.git`,

`git fetch origin && git reset --hard origin/ctypes` will work. Note that this command will remove your entire local changes.

darleybarreto:
> Note that this command will remove your entire local changes.

Yeah, there wasn't anything different anyways. Thanks!

youknowone:
I pushed to your repo directly since I thought you are not active for this PR anymore, but now I know you come back. So I will make PRs for suggestion from now.

Here is another refactoring suggestion: https://github.com/rodrigocam/RustPython/pull/4

darleybarreto:
Nice, thanks!

darleybarreto:
@youknowone So the idea is to do that for all files in `ctypes`? Using `my_filename::extend_module`?

youknowone:
Yes, that's possible to adapt to all files. It is up to you for convenience. `dll` module looked good to use it. But you don't need to use it if you don't feel like to do.

darleybarreto:
@youknowone I basically patched all your comments, you wrote:
> no, it isn't. slice is a sort of view

For this code
```rust
fn obj_bytes(&self) -> BorrowedValue<[u8]> {
    PyRwLockReadGuard::map(self.data.borrow_value(), |x| unsafe {
        slice::from_raw_parts(x.inner, x.size)
    })
    .into()
}
```
What did you meant there?

youknowone:
There was a question asking if it does copy. The comment is an answer about it. Creating an slice object doesn't copy anything.

darleybarreto:
I added a C file that should be compiled as a shared lib and bundled together with the interpreter. Its `_ctypes_test.c`. How would one make this happen? Would it be some sort of build script using `cc`?

youknowone:
`build.rs` file can contain any build script. Using [cc](https://docs.rs/cc/1.0.69/cc/) crate will be helpful.

darleybarreto:
@youknowone I haven't figured out a nice way to implement the `from_buffer`. `from_buffer_copy` is straight forward, you simply create a new instance (default args for both tp_new and init) and copy the bytes from the value to create an instance. For example:
```python
data = b'data'
ubyte = c_ubyte * len(data)
byteslike = ubyte.from_buffer_copy(data) # an instance of ubyte made from data by copying things
```
`from_buffer` does a similar thing, but it doesn't copy things, it points to the source. So if you change `byteslike`, you change `data`.

My implementation of `from_buffer_copy` is something in these lines
```rust
#[pyclassmethod]
fn from_buffer_copy(
    cls: PyRef<Self>,
    obj: PyObjectRef,
    offset: OptionalArg,
    vm: &VirtualMachine,
) -> PyResult {
    let buffer = try_buffer_from_object(vm, &obj)?;
    let opts = buffer.get_options().clone();

    let (size, offset) = Self::buffer_check(cls, opts, offset, vm); //ignore this for now

    let src_buffer = buffer.obj_bytes();
    let empty_instance = ... //creates a new empty instance
    let dst_buffer = empty_instance.obj_bytes_mut();

    dst_buffer.copy_from_slice(&src_buffer[offset..offset+size]);

    Ok(empty_instance.as_object().clone())
}
```
So the idea is simply get the bytes from `obj` (it should implement the buffer protocol) and point to them. How would I point to the data in `obj` instead of copying it? Should I use some refcounted type to avoid pointing to dangling things? `obj` should always be a buffer with write access.

youknowone:
I don't know well about your requirements, but I guess you can do same way as how `PyMemoryView` does. Or even just use it. it just holds a `PyBufferRef` and copy the data only when it needs to be. It also locks `PyBufferRef` when the lock is required.

youknowone:
I think this PR is (practically) not possible to rebase anymore in this state. Do you mind if we squash the commits into single commit for rebase?

darleybarreto:
> I think this PR is (practically) not possible to rebase anymore in this state. Do you mind if we squash the commits into single commit for rebase?

Not at all :)

I'm afraid I can't contribute in the following months. I can help others eventually, tho.

akbog:
Is this effort still ongoing / has this ctypes implementation been abandoned for another?

darleybarreto:
Hi folks, I think we should close this, it's too old for any merging attempt.
Happy to see @arihant2math tackling ctypes. I was actually look at this again (starting from scratch).
I'd like to help on designing the new implementation, but haven't got much spare time to code.

youknowone:
This is still a best trial of ctypes now. I wouldn't close it unless we actually merge any ctypes implementation.

For anyone who will try another ctypes implementation, I hope it can be gradually mergable, not to be suffered by huge size of rebase.

arihant2math:
Agreed, the rebase effort was slowed by refactors that have happened since this pr, I'm planning to start incremental work on this soon.

arihant2math:
Note that #5572 is now the tracking issue for further ctypes implementation.

PR #2366, opened 2020-12-10
replaced timing_safe_cmp implementation with Subtle

orf:
This PR runs `cargo upgrade` on the project and fixes associated errors.

I replaced `timing_safe_cmp` with Subtle rather than trying to upgrade the code to use the new version of `volatile`. I think Subtle is a good crate to rely on for this.

I also used the `default-members` feature of Cargo to remove the need for specific package flags in the CI tests, making getting up and running with RustPython simpler (for me at least!)


youknowone:
If we regard RustPython core modules as a  library, caret or tilde notation will give a bit of freedom for users
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

common/src/cmp.rs
```diff
@@ -1,52 +1,10 @@
...
-        let r = right.read();
-        right.write(r.wrapping_add(1));
-        // safety: the 0..length range will always be either:
-        // * as long as the length of both a and b, if len_a and len_b are equal
-        // * as long as b, and both `left` and `right` are b
-        result |= unsafe { l.read_volatile() ^ r.read_volatile() };
-    }
-
-    result == 0
+    a.ct_eq(b).into()
```
coolreader18, line 9: I think if we're just using subtle, we could just depend on subtle from `rustpython-vm` and remove the `cmp` module here.

coolreader18:
I think caret is the default:

> The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

This is just so that CI and stuff uses updated versions, if I understand correctly.

coolreader18:
Also, I'm not sure if default-members is the desirable behavior, `cargo build` now would build all the libraries rather than just building the binary

youknowone:
This PR was covering too much things in a single PR. I rebased it without outdated stuff.

PR #2372, opened 2020-12-17
Show functions' names

coolreader18:


youknowone (approved):
wow, finally this is done

PR #2373, opened 2020-12-18
Have a block-based IR for compiling, calculate max stack size

coolreader18:


coolreader18:
This would also be great for compiling to cranelift (especially since I modeled the block system off of their ir, a bit :upside_down_face: ), but I'm not sure if there's any way to preserve the block information all the way to the function object that you call `.__jit__()` on

coolreader18:
Best I can tell, there's little performance improvement and this may be very, *very* slightly slower (I get stuff like `norm 1.00 ± 0.01 faster than opt`), but I think there's a lot more room for optimization with this compilation model than what we had before.

carbotaniuman:
It'd be nice for a JIT where we just remove the interpreter loop and directly call the functions imo.

youknowone (approved):

youknowone:
it looks like arrayvec exists in crates.io but boxvec doesn't. maybe it would be useful as an independent crate

PR #2376, opened 2020-12-19
Flatten rustpython_bytecode::bytecode into the root of the crate

coolreader18:


vm/pylib-crate/src/lib.rs
```diff
@@ -2,16 +2,11 @@
...
 #[cfg(feature = "compiled-bytecode")]
-use {
-    rustpython_bytecode::bytecode::{self, FrozenModule},
-    std::collections::HashMap,
-};
+use {rustpython_bytecode::FrozenModule, std::collections::HashMap};
 #[cfg(feature = "compiled-bytecode")]
 pub fn frozen_stdlib() -> HashMap<String, FrozenModule> {
-    rustpython_derive::py_freeze!(dir = "Lib", crate_name = "rustpython_pylib")
+    rustpython_derive::py_freeze!(dir = "Lib", crate_name = "rustpython_bytecode")
```
youknowone, line 11: is this crate_name chage intended?

youknowone (approved):

vm/pylib-crate/src/lib.rs
coolreader18: Yes, I updated compile_bytecode to be a bit more lenient with it. I think it used to require stuff from `rustpython_vm::__exports` but it shouldn't anymore.

PR #2377, opened 2020-12-20
Add fsum to the math module

carbotaniuman:


Lib/test/test_math.py
```diff
@@ -1559,12 +1558,12 @@ def testIsinf(self):
     #     self.assertTrue(math.isnan(math.nan))
 
     # TODO: RUSTPYTHON
```
youknowone, line 1560: This `TODO` is now invalidated

vm/src/stdlib/math.rs
```diff
@@ -372,6 +372,100 @@ fn math_lcm(args: Args<PyIntRef>) -> BigInt {
...
+        let mut x = obj?.to_f64();
+
+        let xsave = x;
+        let mut j = 0;
+        // This inner loop applies `hi`/`lo` summation to each
+        // partial so that the list of partial sums remains exact.
+        for i in 0..partials.len() {
+            let mut y: f64 = partials[i];
+            if x.abs() < y.abs() {
+                let t = x;
```
youknowone, line 391:
```
std::mem::swap(&mut x, &mut y);
```

youknowone (commented):
would you check the clippy errors?

carbotaniuman:
Hmm, this is a straight port, and the strict equality is present in the C implementation.

coolreader18:
Yeah, that clippy lint isn't always that helpful. You could just `#[allow()]` it, it's fine.

youknowone:
For float point strict comparision, if it is intended, `#[allow()]` for that block with a comment will be helpful.
In that case, I think it must be verified if C code doesn't have a bug for that.

I think swap must be fixed because it is lack of C functionality.

`while -> for` is good to chage if possible. But if it is too much painful, ignoring is also ok.
For now, to be honest, I am not 100% trust that warning, because it had a bug to make false warning before

coolreader18 (approved):
Nice work!

PR #2387, opened 2020-12-30
use unsigned abs to avoid overflow

### LINKED ISSUE #2386: Overflow error in slicing
When building to debug (not release, so that overflows lead to panics), I have a bug which results in a crash:
If this is built using release then this crash doesn't happen.

```
% ./target/debug/rustpython     
>>>>> import sys
>>>>> bytearray()[-sys.maxsize - 1]
thread 'main' panicked at 'attempt to negate with overflow', /rustc/2987785df3d46d5ff144a5c67fbb8f5cca798d78/library/core/src/num/mod.rs:155:5
```

Note that `-sys.maxsize-2` or just `-sys.maxsize` do not result in crashes, but report the expected python error. This is built using current master, and current nightly rust on a mac.

I only discovered this project last night and I am actively trying to learn rust. I spent some time trying to track the bug down and the best I could come up with was this MWE.
`bytearray()` can also just be an empty list.

carbotaniuman:
Closes #2386

### DIFF
--- a/extra_tests/snippets/builtin_slice.py
+++ b/extra_tests/snippets/builtin_slice.py
@@ -1,5 +1,4 @@
 from testutils import assert_raises
-import sys
 
 a = []
 assert a[:] == []
@@ -130,5 +129,3 @@ def test_all_slices():
 
 
 test_all_slices()
-
-assert_raises(IndexError, lambda: b[::0], _msg='bytearray index out of range')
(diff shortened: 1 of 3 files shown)


carbotaniuman:
Closes #2386, I'll add a test soon.

coolreader18 (approved):

PR #2389, opened 2021-01-02
Delete no_kwargs! macro

carbotaniuman:


coolreader18 (approved):
Huh, forgot we had that

coolreader18:
Looks like the failure is from a new clippy lint for rust 1.49

PR #2392, opened 2021-01-03
Fix clippy lints for Rust 1.49

coolreader18:


qingshi163:
Are we plan to move to rust 1.5?

coolreader18:
Yeah, we generally support the most recent version or 2 of Rust. Rust 1.5 is about 5 years old, so you can definitely use `char::from_u32_unchecked`

youknowone (approved):

youknowone:
I guess @qingshi163 meant 1.50

Issue #2388, opened 2020-12-31
[RFC] AST Customization


pca006132:
## Summary

Provide means to customize the AST, for including additional data with each node.

## Detailed Explanation

Background: I'm currently developing a compiler for a subset of Python, currently planning to use the RustPython parser to get the AST. The project requires type inference and other processing with the AST. It would be helpful if we can attach data to each node in the AST. 

I think allowing users to store custom data in AST nodes would also help other developers to use the crate for other purposes, like static analysis, or provide better optimization for the RustPython runtime by trying to recover more type information (not sure if this is feasible though, I am not familiar with the project).

I think it would be possible to provide an additional field using generic with a default value. For example:
```rust
pub struct Located<T, U = ()> {
    pub location: Location,
    pub node: T,
    pub custom: Option<U>,
}
```
And add the `U` parameter to all references to `Located` (`Expression`, `Statement`, ... basically all structs in the AST and parser methods)

With the default parameter, I think it would not break the existing code as users can still use `Expression<U>` as `Expression`.
This may also break pattern matching, but that would just require the users to add `..` to the pattern so it should be simple.

## Drawbacks, Rationale, and Alternatives

Drawbacks:
- May break pattern matching in existing code.
- The default type `U = ()` would increase the size of the `Located` struct. One way to prevent the problem would be to use the never type `!`, but this requires nightly. The following code print 16 and 8 in the rust playground.
  ```rust
  #![feature(never_type)]
  struct Test<U = ()> {
      a: usize,
      b: Option<U>
  }
  
  fn main() {
      println!("{}", std::mem::size_of::<Test>());
      println!("{}", std::mem::size_of::<Test<!>>());
  }
  ```

An alternative would simply be not supporting the additional field and require users to store it elsewhere, maybe using HashMap or another tree. That would be harder to use and suffer from performance problems.

coolreader18:
The default could also be `std::convert::Infallible`, which is almost equivalent to the never type and will eventually become a type alias for `!` once the latter is stabilized.

coolreader18:
I think this definitely sounds cool! Would you be interested in implementing it?

pca006132:
> The default could also be `std::convert::Infallible`, which is almost equivalent to the never type and will eventually become a type alias for `!` once the latter is stabilized.

Yes, it seems that the resulting size is the same as `!`.

> I think this definitely sounds cool! Would you be interested in implementing it?

Yes.

coolreader18:
[...]

PR #2394, opened 2021-01-04
Implements AST Customization (#2388)

pca006132:
Tested with `cargo test` in `ast`, `parser` and `compiler` directory (as `cargo test --all` would fail to compile for wasm...)

### DIFF
--- a/parser/src/python.lalrpop
+++ b/parser/src/python.lalrpop
@@ -69,7 +69,7 @@ PassStatement: ast::Statement = {
     <location:@L> "pass" => {
         ast::Statement {
             location,
-            custom: Default::default(),
+            custom: (),
             node: ast::StatementType::Pass,
         }
     },
@@ -79,7 +79,7 @@ DelStatement: ast::Statement = {
     <location:@L> "del" <targets:ExpressionList2> => {
         ast::Statement {
             location,
-            custom: Default::default(),
+            custom: (),
             node: ast::StatementType::Delete { targets },
         }
     },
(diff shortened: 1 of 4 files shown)


coolreader18:
(The clippy error is a separate issue, addressed in #2392)

Is this ready to be merged? Or will you implement custom-type mapping inside rustpython-ast?

parser/src/python.lalrpop
```diff
@@ -69,6 +69,7 @@ PassStatement: ast::Statement = {
     <location:@L> "pass" => {
         ast::Statement {
             location,
+            custom: Default::default(),
```
coolreader18, line 72: Is there a reason these can't be `custom: ()` as well? If you were looking to eventually have the parser be generic, `custom: ()` can be easily find/replaced eventually.

pca006132: I was looking to have the parser be generic later, but `()` seems to be clearer. will fix.

pca006132:
> (The clippy error is a separate issue, addressed in #2392)
> 
> Is this ready to be merged? Or will you implement custom-type mapping inside rustpython-ast?

I think custom-type mapping can be done in a separate PR, as that would probably be pretty large, as the AST involves quite a lot of structs.

coolreader18:
Could you switch out all the `Default::default()`s with `()` ? It's just less noisy, IMO

coolreader18 (approved):

coolreader18:
Thanks for contributing!

PR #2399, opened 2021-01-10
Generate rustpython-ast based off of Python.asdl

### LINKED ISSUE #2388: [RFC] AST Customization

coolreader18:
cc @isidentical, I think you had mentioned being interested in this

Closes #2395, closes #2388


coolreader18:
:tada:

```
❯ rustpython
>>>>> import flask
>>>>> flask
<module 'flask' from '/home/coolreader18/rspy-tests/pallets/flask/__init__.py'>
```

coolreader18:
@pca006132 this also implements AST mapping, since it's pretty easy when everything's autogenerated :)

pca006132:
> @pca006132 this also implements AST mapping, since it's pretty easy when everything's autogenerated :)

I just have a glance on it, so the python scripts are used for parsing the python asdl file, and generate the rust code for AST definitions?

Considering the change would break `rustpython_parser' API, maybe a version bump would be needed?

coolreader18:
Yeah, I was planning to release all new versions next time we release, we make breaking changes often to all of our crates.

coolreader18:
@youknowone could you review this?

compiler/porcelain/src/lib.rs
```diff
@@ -75,29 +75,16 @@ pub fn compile(
...
-            compile::compile_program_single(ast, source_path, opts)
-        }
+    let mode = match mode {
+        compile::Mode::Exec => parser::Mode::Module,
+        compile::Mode::Eval => parser::Mode::Expression,
+        compile::Mode::Single => parser::Mode::Interactive,
+    };
+    let ast = match parser::parse(source, mode) {
+        Ok(x) => x,
+        Err(e) => return Err(CompileError::from_parse(e, source, source_path)),
```
youknowone, line 85:
```
let ast = parser::parse(source, mode).map_err(|e| CompileError::from_parse(e, source, source_path))?;
```

vm/src/builtins/make_module.rs
```diff
@@ -292,30 +321,17 @@ mod decl {
...
     fn getattr(
         obj: PyObjectRef,
         attr: PyStrRef,
         default: OptionalArg<PyObjectRef>,
         vm: &VirtualMachine,
     ) -> PyResult {
-        let ret = vm.get_attribute(obj, attr);
         if let OptionalArg::Present(default) = default {
-            ret.or_else(|ex| catch_attr_exception(ex, default, vm))
+            Ok(vm.get_attribute_opt(obj, attr)?.unwrap_or(default))
```
youknowone, line 332: is this better to be changed like this?

youknowone (approved):
I am sorry to be late 😵

compiler/porcelain/src/lib.rs
coolreader18: `source_path` is moved into `from_parse()`, and rust doesn't connect the closure to the early return, so I have to manually do the map_err.

ast/src/map_ast.rs
```diff
@@ -0,0 +1,66 @@
...
+
+pub trait MapAst<T, U>: Sized {
+    type Mapped;
+    fn try_map_ast<E, F: FnMut(T) -> Result<U, E>>(self, f: &mut F) -> Result<Self::Mapped, E>;
+    fn map_ast<F: FnMut(T) -> U>(self, mut f: F) -> Self::Mapped {
+        let result: Result<_, Infallible> = self.try_map_ast(&mut |u| Ok(f(u)));
+        match result {
+            Ok(mapped) => mapped,
+            Err(never) => match never {},
+        }
```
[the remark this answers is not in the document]
coolreader18: Maybe, but this guarantees that no panic machinery gets pulled in.

ast/src/constant.rs
```diff
@@ -0,0 +1,53 @@
...
+    Complex { real: f64, imag: f64 },
+    Ellipsis,
+}
+
+impl From<String> for Constant {
+    fn from(s: String) -> Constant {
+        Self::Str(s)
+    }
+}
+impl From<Vec<u8>> for Constant {
```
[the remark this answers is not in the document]
coolreader18: Yeah, I figured it doesn't make too much of a difference since it's not going to be "alive" for very long.

vm/src/builtins/make_module.rs
coolreader18: I think so, I added the get_attribute_opt function for use in _ast, but I realized that the logic was being duplicated here as well.

PR #2401, opened 2021-01-17
Implimenting os ctermid

AthulMuralidhar:
## Description
- the main issue is here: https://github.com/RustPython/RustPython/issues/1175
- I am trying to impliment `ctermid` first - so that i can try to understand the rust python compiler
- Please let me know if this one works

### DIFF
--- a/vm/src/stdlib/nt.rs
+++ b/vm/src/stdlib/nt.rs
@@ -356,6 +356,16 @@ pub(crate) mod module {
         raw_set_handle_inheritable(handle, inheritable).map_err(|e| e.to_pyexception(vm))
     }
 
+    fn ctermid(vm: &VirtualMachine) -> PyResult{
+        let ret = unsafe { libc::ctermid() };
+        if ret.is_null(){
+            Err(errno_err(vm))
+        } else {
+            let ret = unsafe { ffi::CStr::from_ptr(ret) }.to_str().unwrap();
+            Ok(vm.ctx.new_str(name))
+        }
+    }
+
     #[pyfunction]
     fn mkdir(
         path: PyPathLike,


AthulMuralidhar:
Hey @coolreader18  could you please let me know if my approach is right?

coolreader18:
It doesn't look like libc has ctermid -- rust-lang/libc#1928. However, you can just define it inline, as:
```
extern "C" {
    fn ctermid(s: *mut libc::c_char) -> *mut libc::c_char;
}
```
Also, you have to pass `std::ptr::null_mut()` as the argument, and you should put the function at the end of the `mod posix` instead of the `mod nt` as you've done.

AthulMuralidhar:
should I inline it within the function itself or is it ok as is?

vm/src/stdlib/os.rs
```diff
@@ -2466,6 +2466,32 @@ mod posix {
...
+            Ok(vm.ctx.new_str(returned))
+        }
+    }
+
+    #[pyfunction]
+    // fchdir is the same as os.chdir(file descriptor)
+    // 1. have a file desriptor that repsresents a directory
+    // 2. pass this fd to the fchdir function
+    // 3. check if the fd is a directory or not
+    // 4. if its a dirrectory, then proceed to chage current directory to the fd's path
```
Pluriscient, line 2488: Some small typos in the comment: desriptor -> descriptor. repsresents -> represents. dirrectory -> directory.

Is this supposed to be a doc comment? then it should have three slashes right?

AthulMuralidhar: no - its just some pseudo code / my thinking strategy of implementing fchdir - i will remove it once i work through the implementation

Pluriscient: Ah ignore my comment then, thought that it was meant to stay ;)

AthulMuralidhar: haha nooooo :P thanks for your (future) review. looking forward to it :)

vm/src/stdlib/os.rs
```diff
@@ -3115,6 +3115,27 @@ mod posix {
...
+        } else {
+            let returned = unsafe { ffi::CStr::from_ptr(returned) }.to_str().unwrap();
+            Ok(vm.ctx.new_str(returned))
+        }
+    }
+
+    #[pyfunction]
+    // fchdir is the same as os.chdir(file descriptor)
+    // also check out: https://docs.rs/libc/0.2.86/libc/fn.fchdir.html
+    fn fchchdir(path: PyPathLike, vm: &VirtualMachine) -> PyResult<()> {
```
youknowone, line 3135: comment says the function name is `fchdir`, but the written name is `fchchdir`.
By Python document, the function should take fd instead of PathLike.

youknowone:
I checked macos python 3.9 and linux 3.6 but cannot find `os._ctermid`
where is it expected to be defined?

vm/src/stdlib/os.rs
AthulMuralidhar: i will update the branch with the changes :) thanks again for the review

AthulMuralidhar: quick question: i do agree that the `fd` must be passed into the function rather than the pathlike above, so should i use an extern as i did with`ctermid` ?  or is the latest version ok?

AthulMuralidhar:
> where is it expected to be defined?

I dont really understand you comment :/  maybe please explain a little more? If you were referring to the function name, it was indeed named `_ctermid` now changed it back to `ctermid` with some refactoring

youknowone:
I expect this PR has just 1 commit. Any PR shouldn't contain a merge commit. Rebase or cherry-pick your work onto upstream/master please.

vm/src/stdlib/os.rs
youknowone: I don't know if `env::set_current_dir` take `fd` as its argument. If it does, it will be ok. If there is any test about this funciton in `test_os.py` or `test_poisix.py`, make them run to test if it works correctly. Otherwise adding a test in `extra_tests/snippets` will be helpful. If the new functions are not fit for the snippet testing, creating a comment how you tested your new functions will be helpful.

AthulMuralidhar: ill add the test cases as well - in this way i'll also get to learn how the testing is done :)

AthulMuralidhar: update: when i was running the `test_os.py` file, the existing test case seems to pass:
 ```bash
test_fchdir (__main__.TestInvalidFD) ... ok
```
so i am guessing i dont need to add anything here ?

AthulMuralidhar:
Note for the reviewers: i saw another pr(https://github.com/RustPython/RustPython/pull/3017) for fcdir so removed the code here

youknowone (commented):
no `#[pyfunction]` means never tested

youknowone:
I close this abandoned PR mainly because this code is never tested before submitting PR.

PR #2403, opened 2021-01-20
Serialize entire frozen hashmap to bytes

coolreader18:
So instead of `hashmap! { "a" => decode(b"....."), "b" => decode(b"......") }`, it's just `decode(b"<map>")`

release binary sizes:

|        | default | freeze-stdlib |
| :----- | ------: | ------------: |
| master | 18296   | 29908         |
| libmap | 18292   | 24832         |

My hunch on this was that putting all the data in the same buffer gives lz4 more opportunity to deduplicate, and I'm pretty sure that's what made the difference in binary size.

### DIFF
--- a/bytecode/src/lib.rs
+++ b/bytecode/src/lib.rs
@@ -1075,7 +1075,7 @@ impl<C: Constant> fmt::Debug for CodeObject<C> {
     }
 }
 
-#[derive(Serialize, Deserialize)]
+#[derive(Serialize, Deserialize, Debug)]
 pub struct FrozenModule<C: Constant = ConstantData> {
     #[serde(bound(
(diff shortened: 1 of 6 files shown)


bytecode/src/lib.rs
```diff
@@ -1084,3 +1084,132 @@ pub struct FrozenModule<C: Constant = ConstantData> {
     pub code: CodeObject<C>,
     pub package: bool,
 }
+
+pub mod frozen_lib {
+    use super::*;
+    use bincode::{options, Options};
+    use std::convert::TryInto;
+    use std::io::{self};
```
youknowone, line 1092: not just `std::io`?

youknowone (approved):

bytecode/src/lib.rs
coolreader18: Hah, nice catch :)

PR #2404, opened 2021-01-23
Fix whats_new.sh not finding libraries that are directories

verhovsky:
I was wondering why `whats_left` says the entire `collections` module isn't implemented yet. It's because both `libdir` and `name` are `bytes`, when you format them into a string, you get a invalid path with `b''` in it.


coolreader18 (approved):
Heh, nice catch!

coolreader18:
Thanks for contributing!

PR #2409, opened 2021-01-24
Fix BufferedReader over a SocketIO to properly read the right amount

coolreader18:
This gets basic `requests` functionality working:

```sh
$ mkdir deps
$ pip install -t deps requests
<snip>
$ cd deps
$ rustpython
```
```py
>>>>> import requests
>>>>> res = requests.get("http://httpbin.org/get")
>>>>> res.json()
{'args': {}, 'headers': {'Accept': '*/*', 'Accept-Encoding': 'gzip, deflate', 'Host': 'httpbin.org', 'User-Agent': 'python-requests/2.25.1', 'X-Amzn-Trace-Id': 'Root=1-600ce51d-187d27d01dae14c54f63f539'}, 'origin': '99.104.70.102', 'url': 'http://httpbin.org/get'}
```

I spent a lot of time debugging this, but I think the key thing that stopped it from hanging forever in `SocketIO.readinto` was fixing `_io` to use `as_contiguous` instead of `obj_bytes`, which wasn't taking the range of a memoryview into account, as well as changing a `>` to a `>=` in `_io`.

### DIFF
--- a/vm/src/stdlib/socket.rs
+++ b/vm/src/stdlib/socket.rs
@@ -59,6 +59,7 @@ pub struct PySocket {
     kind: AtomicCell<i32>,
     family: AtomicCell<i32>,
     proto: AtomicCell<i32>,
+    timeout: AtomicCell<f64>,
     sock: PyRwLock<Socket>,
 }
 
@@ -86,6 +87,7 @@ impl PySocket {
             kind: AtomicCell::default(),
             family: AtomicCell::default(),
             proto: AtomicCell::default(),
+            timeout: AtomicCell::new(-1.0),
             sock: PyRwLock::new(invalid_sock()),
         }
         .into_ref_with_type(vm, cls)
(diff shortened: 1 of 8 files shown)


vm/src/stdlib/socket.rs
```diff
@@ -234,10 +322,42 @@ impl PySocket {
...
+        while buf_offset < buf_len {
+            let interval = timeout.map(|dur| match deadline {
+                Some(d) => d
+                    .checked_duration_since(Instant::now())
+                    // past the deadline already
+                    .ok_or_else(|| timeout_error(vm)),
+                None => {
+                    let dl = Instant::now() + dur;
+                    deadline = Some(dl);
+                    Ok(dur)
```
youknowone, line 347: I am not understanding this part very well. `deadline` looks like to be None at the beginning and only assigned as Some here. If then, it looks like to decided out of loop. Am I missing somethinig?

coolreader18: I think CPython does it like this to avoid the cost of the monotonic clock when there's no deadline. I think it would simplify the code a lot just to always do it; or should I just add comments about why we do it like this?

coolreader18:
The re tests are fixed on master

PR #2411, opened 2021-01-24
Detect incompatible function signatures in whats_left

verhovsky:
And there's a framework for checking other differences, for example you could check if integer constants are different. 

It's WIP in the sense that the code is a mess but it does work. There's a bit of noise because of superficial differences (like `kwds` instead of CPython's `kwargs`) but I think requiring `inspect.signature` to produce results identical to CPython for all inputs is a reasonable requirement.


extra_tests/not_impl_gen.py
```diff
@@ -49,51 +67,50 @@ def gen_methods(header, footer, output):
...
 
-    output.write("}\n\n")
-    output.write(footer.read())
+    output = "expected_methods = {\n"
+    for name, (typ_code, extra, attrs) in methods.items():
+        output += f" '{name}': ({typ_code}, {extra!r}, [\n"
+        for attr, attr_extra in attrs:
+            output += f"    ({attr!r}, {attr_extra!r}),\n"
+        output += " ]),\n"
+        # TODO: why?
```
andrew-ld, line 109: to format the file correctly, it should not insert \n in the last iteration

coolreader18:
I think this looks good! @verhovsky, is this ready for review?

verhovsky:
@coolreader18 sure.

coolreader18 (approved):
Thanks for working on this!

PR #2416, opened 2021-01-26
DirEntry as os.PathLike, st_xtime_ns

coolreader18:


youknowone (approved):
a new unused import warning added. please check it before merging

PR #2417, opened 2021-01-26
Declare ourselves as Python 3.9

coolreader18:
Python 3.5 is pretty old at this point, and I think we generally try to keep up with the latest features (e.g. assignment expression, self-documenting fstrings), so I think it makes sense to say we're 3.9. I put 3.8 in the READMEs because I don't think we have full 3.9 compatibility yet and 3.8 is more used, but I'm not really sure if that makes a difference :shrug:

### DIFF
--- a/Lib/site.py
+++ b/Lib/site.py
@@ -265,13 +265,14 @@ def _getuserbase():
 def _get_path(userbase):
     version = sys.version_info
 
+    # XXX RUSTPYTHON: we replace pythonx.y with rustpythonx.y
     if os.name == 'nt':
-        return f'{userbase}\\Python{version[0]}{version[1]}\\site-packages'
+        return f'{userbase}\\RustPython{version[0]}{version[1]}\\site-packages'
 
     if sys.platform == 'darwin' and sys._framework:
-        return f'{userbase}/lib/python/site-packages'
+        return f'{userbase}/lib/rustpython/site-packages'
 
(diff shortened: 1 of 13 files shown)


youknowone (approved):

domdfcoding:
What was the rationale for changing the site-packages paths to rustpythonX.Y? pip seems to be hardcoded to install to pythonX.Y and PyPy uses pythonX.Y too.

coolreader18:
It seems like pip respects the rustpythonX.Y path in my experience, and I think my motivation was that we don't have anywhere close to full compatibility for many libraries (especially that require native C extensions) so I feel it makes sense to keep them separate, at least for now.

domdfcoding:
Makes sense. 

I think the pip behaviour I saw is because there's still a reference to the `pythonX.Y` form in `distutils.sysconfig`:

https://github.com/RustPython/RustPython/blob/2bcf7d4d5b0d16b5e0f0dab555ba222f42aaf796/Lib/distutils/sysconfig.py#L136-L138

and in `distutils.command.install`: 

https://github.com/RustPython/RustPython/blob/2bcf7d4d5b0d16b5e0f0dab555ba222f42aaf796/Lib/distutils/command/install.py#L30-L37

If I change those by hand pip installs to the correct location.

coolreader18:
Huh, interesting, it installs to the right place on my machine but feel free to make a PR if that fixes it in your environment.

PR #2421, opened 2021-01-28
Extract the sre interpreter into a separate repo

coolreader18:
cc @qingshi163 - I have a local repo that keeps the commit history on `sre/interp.rs` but moves it to its own crate, do I have your permission to push it to a `RustPython/sre-engine`? (since it's your IP, I think). Or do you think this is a good idea? I think it's good in general to try to keep rustpython-vm small, since it's already got a ton of stuff in it and takes a while to compile, and sre/interp is a really contained piece of logic that I think it makes sense to split out to its own crate.


qingshi163:
Sure it is better to move it to the crate as compiling is already so slow.. and yes you can push it.

PR #2422, opened 2021-01-28
Stub plistlib and xmlrpc

coolreader18:


youknowone (approved):
is there no test for this library?

coolreader18:
No, it wouldn't work anyway, cause we don't really have any xml support (we need bindings to expat). However, pip imports plistlib and xmlrpc, and though it doesn't really do much with them, I think effectively stubbing them is better than editing pip's code.

PR #2423, opened 2021-01-28
Update argparse

fanninpm:
This pull request will:

- Update the `argparse` module to the one found in [Python 3.9](https://github.com/python/cpython/tree/3.9)
- Fix a capitalization error that caused one of the newly-added `argparse` tests to fail


coolreader18 (approved):
Nice work!

PR #2424, opened 2021-01-29
add a github workflow to save whats left to the website repo

### LINKED ISSUE #2400: bash script / github workflow to generate csv out of whats_left.sh
In this previous PR https://github.com/RustPython/RustPython/pull/1865 , I made some tweaks to `whats_left.sh` to generate yaml files. The goal is to display some sort of dashboard on the website (rustpython.github.io). Eventually we didn't merge the PR because it meant that we needed to add some more flags or argument for the script (which I never got around to).

Here is a easier suggestion for a new bash script that we can run as a github workflow.   
The script:
- generates csv files
- cleans up some output from the whats_left script that we are not likely to display on the site
- creates two separate files (one for missing entire modules and one for missing functions and classes)

Here is what the script looks like:

``` bash
#!/bin/bash

# run whats_left and save the output to a temp file
# use grep to remove some test output like hello world and constant string
sh ./whats_left.sh | grep -v 'Constant String' | grep -v 'Hello world'  > whats_left.temp

# start a new csv file that has the modules
echo "module" > whats_left_modules.csv
# read the temp file and look for "entire module" 
cat whats_left.temp | grep "entire module"  >> whats_left_modules.csv

# start a new csv file that has just the missing functions and classes
echo "module, partial" > whats_left_partials.csv
# get the diff between the full list and the module list
# remove some keywords like "existing but not importable"
# replace the "." with "," which nicely creates a new column in csv
grep -v -f  whats_left_modules.csv whats_left.temp | grep -v 'existing but not importable' |  tr "." "," >> whats_left_partials.csv
```

I tested that jekyll can actually read and display the data with a simple loop, for example.
```
 {% for item in site.data.whats_left_modules %}
      <li>
        {{ item.module }}  
      </li>
    {% endfor %}
```
For the second file, jekyll layout needs to be more fancy and group by the module name.

Anyway.... this is the whole script that is needed and it re-uses `whats_left.sh`

- is this a better solution than my previous attempt at a PR?
- should I submit this bash script as a PR for a file here? I thought creating a new one like `whats_left_workflow.sh` is just weird and looks like clutter.
- adding those instructions in the github workflow is also weird

any thoughts?

mireille-raad:
This PR fixes #2400.

I added a new job that runs the `whats_left.sh` script 

```bash
 chmod +x ./whats_left.sh
./whats_left.sh > whats_left.temp
```

Then after cloning the website and cd-ing into it, it does:

``` bash
git clone git@github.com:RustPython/rustpython.github.io.git website
cd website
[ -f ./_data/whats_left.temp ] && cp ./_data/whats_left.temp ./_data/whats_left_lastrun.temp
cp ../whats_left.temp ./_data/whats_left.temp
git add ./_data/whats_left_lastrun.temp
git add ./_data/whats_left.temp
```

which translates to:
- check if website/_data/whats_left.temp exits (on the first run it won't)
- make a copy of that to a "lastrun" file
- the goal is to have the last run and the what's left now, so we can do a diff then show what has been recently finished (kind of a basic "tada"-log  🎉  )
- replace whatsleft
- commit and push

### DIFF
--- a/.github/workflows/cron-ci.yaml
+++ b/.github/workflows/cron-ci.yaml
@@ -86,3 +86,37 @@ jobs:
           git add ./_data/regrtests_results.json
           git -c user.name="Github Actions" -c user.email="actions@github.com" commit -m "Update regression test results" --author="$GITHUB_ACTOR"
           git push
+
+  whatsleft:
+    name: Collect what is left data
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: build rustpython
+        uses: actions-rs/cargo@v1
+        with:
+          command: build
+          args: --release --verbose
+      - name: collect what is left data
+        run: |
+          chmod +x ./whats_left.sh
+          ./whats_left.sh > whats_left.temp
+        env:
+          RUSTPYTHONPATH: ${{ github.workspace }}/Lib
+      - name: upload data to the website
+        env:
+          SSHKEY: ${{ secrets.ACTIONS_TESTS_DATA_DEPLOY_KEY }}
+          GITHUB_ACTOR: ${{ github.actor }}
+        run: |
+          echo "$SSHKEY" >~/github_key
+          chmod 600 ~/github_key
+          export GIT_SSH_COMMAND="ssh -i ~/github_key"
+
+          git clone git@github.com:RustPython/rustpython.github.io.git website
+          cd website
+          [ -f ./_data/whats_left.temp ] && cp ./_data/whats_left.temp ./_data/whats_left_lastrun.temp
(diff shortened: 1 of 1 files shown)


youknowone (approved):
Not sure if saving whatsleft result to git repository is the best idea, but I have no better idea for now

fanninpm:
Just as a heads up, the current wording ("fixes issue \#xxxx") won't [automatically close the referenced issue when this PR gets merged.](https://docs.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) If you *want* that behavior, you can delete the word "issue".

coolreader18 (approved):
LGTM!

PR #2429, opened 2021-02-01
Add secrets from CPython 3.8

verhovsky:


verhovsky:
`binascii` should be changed to use a Rust crate.

coolreader18 (approved):
Yeah, I think we'll eventually need to rewrite binascii to be specifically compatible with CPython; I've run into issues with Python's base64 allowing newlines in the middle of a b64 string but the base64 crate not.

PR #2430, opened 2021-02-01
Add telnetlib from CPython 3.8

verhovsky:


coolreader18 (approved):
Thanks for contributing! Now we can watch star wars in RustPython :)

PR #2431, opened 2021-02-01
Fix formatting of decimal powers below 1 million

merkrafter:
The previous implementation truncated all decimal powers of 10.0 below 1 million (by default) to 1, because it cut away any
combination of trailing 0's and .'s. 
(Powers of 10.0 beginning with 1000000.0 are formatted to an exponential format and thus not affected by the bug.)
My fix now checks whether the number string to truncate represents a floating point value at all and if that is the case, truncation stops at the decimal point.
(The check whether there is a decimal point in the string to truncate is necessary as `format!("{:.*}", precision, magnitude)` with `precision==0` does return strings without a decimal point; truncating those is fatal)

### DIFF
--- a/common/src/float_ops.rs
+++ b/common/src/float_ops.rs
@@ -137,12 +137,32 @@ pub fn format_exponent(precision: usize, magnitude: f64, case: Case) -> String {
     }
 }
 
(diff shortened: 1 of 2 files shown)


coolreader18 (approved):
Nice catch, thank you for contributing!

PR #2432, opened 2021-02-01
Add statistics from CPython 3.8

verhovsky:


vm/src/stdlib/random.rs
```diff
@@ -69,7 +69,13 @@ mod _random {
     #[pyimpl(flags(BASETYPE))]
     impl PyRandom {
         #[pyslot(new)]
-        fn new(cls: PyTypeRef, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+        #[allow(unused_variables)]
```
youknowone, line 72: rather than adding this allow, renaming `x` to  `_x` will work

youknowone (changes requested):

youknowone:
Any other part look good! Thank you.

coolreader18 (approved):

PR #2437, opened 2021-02-02
Add reason for skipped codeop test

fanninpm:


coolreader18 (approved):
Thanks for following up on this!

PR #2438, opened 2021-02-03
Handle splat args in bases/kwargs for __build_class__

### LINKED ISSUE #2435: Unable to use starred expressions in class bases
## Feature

in CPython you are able to do:
```py
class Foo(*(Bar,)):
    pass
```
but in RustPython you get:
```py
>>>>> class Foo(*(Bar,)):
.....     pass
.....
SyntaxError: can't use starred expression here at line 1 column 11
class Foo(*(Bar,)):

          ↑
>>>>>
```
## Python Documentation
*dont know what to put here as its part of the syntax rather than stdlib or etc.*

### LINKED ISSUE #2436: Metaclass kwargs are giving strange outputs
## Feature

when doing some metaclass wierdness, rustpython does not handle it correctly, annoyingly I can't pinpoint where the issue lies, so im hoping someone else can and make a fix for this.

code in questioning:
```py
class DictSetter(dict):
    def __init__(self, *args, func):
        self.func = func
        super().__init__(*args)

    def __getitem__(self, name):
        if name not in self:
            self[name] = self.func(name)
        return dict.__getitem__(self, name)

class Meta(type):
    def __init__(cls, name, bases, attrs, **_kwargs):
        super().__init__(name, bases, attrs)

    @classmethod
    def __prepare__(cls, _name, _bases, *, values):
        return DictSetter(func=values)

    def __new__(cls, name, bases, attrs, **_kwargs):
        return super().__new__(cls, name, bases, attrs)

class MagicAttrs(metaclass=Meta, values=lambda name: name):
    pass

class MyClass(MagicAttrs, values = lambda name: name):
    a
    b
    c
```
this works as intended in CPython (run `Myclass.a`, `Myclass.b` and `Myclass.c`), but in rustpython im getting an error:
```py
  File "magicattrs.py", line 24, in <module>
    class MagicAttrs(metaclass=Meta, values=lambda name: name):
TypeError: Missing required kw only argument: 'values'
```

## Python Documentation

https://docs.python.org/3/reference/datamodel.html#preparing-the-class-namespace
https://docs.python.org/3/reference/datamodel.html#creating-the-class-object
> Once the class namespace has been populated by executing the class body, the class object is created by calling `metaclass(name, bases, namespace, **kwds)` (the additional keywords passed here are the same as those passed to `__prepare__`).

coolreader18:
Fixes #2435, and also fixes #2436

@Zomatree, could you verify that your example works now?

### DIFF
--- a/Lib/test/test_subclassinit.py
+++ b/Lib/test/test_subclassinit.py
@@ -197,8 +197,6 @@ class Test(unittest.TestCase):
         self.assertIs(B.meta_owner, B)
         self.assertEqual(B.name, 'd')
 
-    #TODO: RUSTPYTHON
-    @unittest.skip("infinite loops")
     def test_set_name_modifying_dict(self):
         notified = []
         class Descriptor:
@@ -215,8 +213,6 @@ class Test(unittest.TestCase):
 
         self.assertCountEqual(notified, ['a', 'b', 'c', 'd', 'e'])
 
-    #TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_errors(self):
         class MyMeta(type):
             pass
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -1158,6 +1158,8 @@ impl Compiler {
     ) -> CompileResult<()> {
         self.prepare_decorators(decorator_list)?;
 
+        self.emit(Instruction::LoadBuildClass);
+
         let prev_ctx = self.ctx;
         self.ctx = CompileContext {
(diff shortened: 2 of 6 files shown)


Zomatree:
I can confirm both starred arguments and `__prepare__` work, however there is now an issue in the typings lib related to this
```py
  File "/home/zomatree/rustpythonpatch/vm/pylib-crate/Lib/typing.py", line 160, in <module>
    class _TypingBase(metaclass=TypingMeta, _root=True):
TypeError: Unexpected keyword argument _root
```

PR #2439, opened 2021-02-03
Add more functionality to _ssl

coolreader18:


coolreader18:
The windows failure looks like a fluke

vm/src/stdlib/ssl.rs
```diff
@@ -743,19 +783,29 @@ fn convert_ssl_error(
...
-                    vm.ctx.new_str(e.to_string()),
-                ],
+    let (cls, msg) = match e.code() {
+        ssl::ErrorCode::WANT_READ => (
+            vm.class("_ssl", "SSLWantReadError"),
+            "The operation did not complete (read)",
+        ),
+        ssl::ErrorCode::WANT_WRITE => (
+            vm.class("_ssl", "SSLWantWriteError"),
+            "The operation did not complete (read)",
```
youknowone, line 793: is this message also `read` instead of `write`?

coolreader18: Heh, nice catch

youknowone (approved):

PR #2440, opened 2021-02-05
import this

dbrgn:
A humble suggestion 😌 🦀


fanninpm:
RIIR requires `from __future__ import braces`.

youknowone:
Because `import this` starts with referring Tim Peter, this change will infringes [moral rights](https://en.wikipedia.org/wiki/Moral_rights)

dbrgn:
> Because `import this` starts with referring Tim Peter, this change will infringe [moral rights](https://en.wikipedia.org/wiki/Moral_rights) :p

Well, even Aristoteles famously stated that wrong attribution is not morally wrong if it results in more memory-safe software.

ShadowJonathan (changes requested):
```diff
- Have you considered rewriting it in Rust?
+ Also, have you considered rewriting it in Rust?
```

(I think that fits the flow of the Zen better)

Also, wouldn't an extra newline, or parentheses work better? To make it clear the extra addition is supplemental

PR #2441, opened 2021-02-06
Implement os.getlogin

dbrgn:
Implement `os.getlogin`.

This is still marked as draft, because right now in case of an error, an empty string is returned. However, when testing in a Docker container without a defined login name, I noticed that calling `os.getlogin()` on CPython returns an `OSError` instead:

> OSError: unable to determine login name

I'll fix that tomorrow.

Refs #1175.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -2329,6 +2329,30 @@ mod posix {
             SupportFunc::new(vm, "execv", execv, None, None, None),
         ]
     }
+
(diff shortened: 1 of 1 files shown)


dbrgn:
Fixed!

CPython:

2021-02-06-133103_920x203_scrot

RustPython:

2021-02-06-133112_839x159_scrot

Dealing with invalid UTF-8:

2021-02-06-133737_1259x185_scrot

coolreader18 (approved):
Thank you for working on this!

PR #2442, opened 2021-02-06
Allow importing PyPI packages on wasm

coolreader18:
cc @birkenfeld

image


birkenfeld:
That's nifty!  For my purposes the package doesn't come from PyPI though, since it needs to reflect the current master branch.

coolreader18:
Ah, it looked like in demo.js it was downloading from pypi. Regardless, this can be extended to fetch any wheel, so as long as a Pygments.whl is built for the master branch this can work with that easily.

jhoobergs:
How far would this feature eventually reach? Would it for example be possible to load numpy?

coolreader18:
@jhoobergs maybe, but we can't load numpy at all at the moment, so it would be dependent on how extension modules eventually work. The `_numpy` module might have to be pre-packaged into the binary, especially for something like wasm which doesn't have dynamically-loaded modules yet.

vm/src/stdlib/zlib.rs
```diff
@@ -89,86 +113,118 @@ mod decl {
...
+            }
+        }
+        fn compress(self, level: Compression) -> Compress {
+            match self {
+                #[cfg(feature = "zlib")]
+                Self::Standard { header, wbits } => {
+                    Compress::new_with_window_bits(level, header, wbits)
+                }
+                #[cfg(not(feature = "zlib"))]
+                Self::Standard { header } => Compress::new(level, header),
```
youknowone, line 144: if this one doesn't share condition with the others, is this needed to be placed between `feature  = "zlib"`?

youknowone (approved):

vm/src/stdlib/zlib.rs
coolreader18: I moved it above the `feature = "zlib"` branches.

PR #2443, opened 2021-02-09
Cleanup skipped tests

fanninpm:
I grep'd the codebase for this regular expression: `@unittest\.skip.[\"']TODO: RUSTPYTHON[^ :;,]`, and ran every test that was associated with it. If it errored, I put the error in the skip reason. Otherwise, I un-skipped it. (This was initially tested on Manjaro, so some breakage on Windows and macOS is to be expected. That's why I'm initially marking this as draft.)


youknowone (approved):
wow, this is a nice work. if you made a script for this, it also will be a nice contribution

youknowone:
one point: we skipped test when it crashed; we intended to mark `expectedFailure` when it just (safely) failed. So when it shows error messages, it looks like to be ready to be promoted to `expectedFailure` instead of `skip`

PR #2449, opened 2021-02-11
Implement os.getgrouplist

kenkoooo:
Implement `os.getgrouplist`.

Ref #1175 

### Question
Looks like we don't have [test_posix.py](https://github.com/python/cpython/blob/49926cf/Lib/test/test_posix.py) to check `os.getgrouplist`. Do we want to copy it from `CPython`?

### DIFF
--- a/Lib/test/test_posix.py
+++ b/Lib/test/test_posix.py
@@ -1,0 +1,2148 @@
+"Test posix functions"
+
+from test import support
+from test.support.script_helper import assert_python_ok
+
+# Skip these tests if there is no posix module.
+posix = support.import_module('posix')
+
+import errno
+import sys
+import signal
+import time
+import os
+import platform
+import pwd
+import stat
+import tempfile
+import unittest
+import warnings
+import textwrap
+
+_DUMMY_SYMLINK = os.path.join(tempfile.gettempdir(),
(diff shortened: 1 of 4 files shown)


kenkoooo:
Looks like we don't have [test_posix.py](https://github.com/python/cpython/blob/49926cf/Lib/test/test_posix.py) to check `os.getgrouplist`. Do we want to copy it from `CPython`?

coolreader18:
Yeah, I think that'd be good! You can follow [this blog post](https://rustpython.github.io/guideline/2020/04/04/how-to-contribute-by-cpython-unittest.html) to add the unittest file.

kenkoooo:
Thanks. Let me do that.

kenkoooo:
@coolreader18 I copied `test_posix.py` and edited it as the following:
- Put `expectedFailure` to unimplemented features.
- Put `skip` to 1 test, which causes `panic`.
- Put `skip` to 1 class, which is an unimplemented macOS feature.

And also I formatted `bytecode/src/lib.rs` and `compiler/src/ir.rs` to follow the latest `rustfmt`, but we still need to fix a lot of places to follow the latest `clippy`. I guess we want to fix them in another PR.

coolreader18 (approved):
Thank you for contributing!

PR #2452, opened 2021-02-13
Allow clippy::unnecessary_wraps lint, and rename "to_js" to "into_js"

fanninpm:
See rust-lang/rust-clippy#6726


coolreader18:
You can put `#![allow(clippy:: unnecessary_wraps)]` in lib.rs so that you don't have to duplicate it in the tank file

coolreader18 (approved):
Thanks for fixing this!

PR #2455, opened 2021-02-14
Update abc to CPython 3.9.1 and add test suite

fanninpm:


coolreader18:
Could you split the commit adding test_abc into 2? One for just copying it over, the next for editing to add the expectedFailures?

coolreader18 (approved):
Thank you for contributing!

PR #2456, opened 2021-02-14
Add __floor__ and __ceil__ to float w/ tests

fanninpm:


fanninpm:
Closing as this was not present in CPython 3.8

coolreader18:
Oh, it's fine to use stuff from CPython 3.9, if you want to reopen this. We have self-describing fstring specifiers and `:=` as well from 3.9.

coolreader18:
Oh, unless there's a issue with CI tests, it seems weird that the test snippets would be affected.

coolreader18:
I suppose you could probably comment out the AttributeError test in stdlib_math

PR #2457, opened 2021-02-15
Add ast tests from CPython 3.8.7

fanninpm:


fanninpm:
Some say [this run](https://github.com/RustPython/RustPython/runs/1900058979) is still running to this day.

coolreader18 (approved):

coolreader18:
Yeah, looks like there's a lot of stuff in compiler that panics instead of returning an error on ill-formed input, since until recently the only input it would get was from the parser.

PR #2460, opened 2021-02-15
Optimize builtins/range

pca006132:
This avoids unnecessary operations, such as clone, multiplication, and division.

For the following microbenchmark, the current master would take time `3.41s, 0.73s, 3.69s`, and it takes `2.27s, 0.50s, 2.60s` after applying this patch.
```python
import time

t0 = time.time()
for i in range(10000000):
    pass
t1 = time.time()
print(t1 - t0)

t0 = time.time()
for i in range(0, 10000000, 5):
    pass
t1 = time.time()
print(t1 - t0)

t0 = time.time()
for i in range(10000000, 0, -1):
    pass
t1 = time.time()
print(t1 - t0)
```

Previously, the `PyRange::get` function takes 30% of the time, and it takes only 10% of execution time after this optimization.

Currently, 34% of the time is spent on `PyDict::inner_itemset_fast` (a lot of time spent on dropping the old value), and about 10% is spent on `BigInt` conversion from `usize` and to `PyInt`.

### DIFF
--- a/vm/src/builtins/range.rs
+++ b/vm/src/builtins/range.rs
@@ -78,28 +78,36 @@ impl PyRange {
     pub fn get(&self, index: &BigInt) -> Option<BigInt> {
         let start = self.start.borrow_value();
         let step = self.step.borrow_value();
-        let index = index.clone();
+        let stop = self.stop.borrow_value();
         if self.is_empty() {
             return None;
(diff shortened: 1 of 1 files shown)


coolreader18 (approved):
LGTM

PR #2461, opened 2021-02-15
Optimize check_signals

pca006132:
`swap` seems to be pretty slow, it took 5.66% of the time in my microbenchmarks. After the optimization, `check_signals` would only take 0.29% of the time. (from flamegraph).

### DIFF
--- a/vm/src/stdlib/signal.rs
+++ b/vm/src/stdlib/signal.rs
@@ -111,9 +111,10 @@ pub fn check_signals(vm: &VirtualMachine) -> PyResult<()> {
         None => return Ok(()),
     };
 
-    if !ANY_TRIGGERED.swap(false, Ordering::Relaxed) {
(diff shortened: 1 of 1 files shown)


coolreader18:
Huh, nice! Do you think you could leave in the `if !swap(false) { return }` as a second sorta backup condition, just to ensure that's it's always atomic?

pca006132:
> Huh, nice! Do you think you could leave in the `if !swap(false) { return }` as a second sorta backup condition, just to ensure that's it's always fully atomic?

But I think signals can't be canceled? Even if it can, it would be a race condition as it depends on whether the canceling occurs between the two if statements or after the last if statement, and the current implementation should match the original one. So I'm not sure why we should ensure the `swap` operation is fully atomic.

coolreader18:
I suppose so; I guess a signal could even happen while we're in the signal handler. This LGTM then, thank you for contributing!

coolreader18 (approved):

PR #2463, opened 2021-02-15
Add test_base64 from CPython 3.8.7

fanninpm:


coolreader18:
I'm not sure if it makes sense to copy this test file yet; the majority of tests are failing or panicking and the underlying rust implementation is just using an external crate, not necessarily attempting to match CPython's behavior.

fanninpm:
It probably makes sense to defer this PR for the reasons you mentioned above. Perhaps we could revisit this test suite in the future.

youknowone:
I agree adding many `unittest.skip`ed tests doesn't give benefits as much as `unittest.expectedFailure` marked tests. But still it will show goals for newcomers.

youknowone (approved):

PR #2465, opened 2021-02-15
Add test_bufio from CPython 3.8.7

fanninpm:


coolreader18 (approved):
Yeah, there's some really weird stuff with windows permissions on CI, I'm pretty sure it's something to do with github actions rather than rustpython.

PR #2466, opened 2021-02-15
Add test_call from CPython 3.8.7

fanninpm:


Lib/test/test_call.py
```diff
@@ -492,6 +494,8 @@ def testfunction_kw(self, *, kw):
 
 class TestPEP590(unittest.TestCase):
```
coolreader18, line 495: I think you could probably skip this whole class, since we don't have support for anything like vectorcalls yet.

youknowone (approved):

PR #2470, opened 2021-02-16
Fix `math.floor` and `math.ceil` functions to match cpython behavior

zsilver1:
Previously the `math.floor` and `math.ceil` functions did not work on subclasses of `float` or on types that implemented the `__float__` method. This change brings the `floor` and `ceil` implementations closer to the [cpython](https://github.com/python/cpython/blob/0837f99d3367ecf200033bbddfa05d061ae9f483/Modules/mathmodule.c#L1194) [versions](https://github.com/python/cpython/blob/0837f99d3367ecf200033bbddfa05d061ae9f483/Modules/mathmodule.c#L1257).


coolreader18 (approved):

PR #2472, opened 2021-02-16
Don't use atomics as much for lasti

coolreader18:
```
Benchmark #1: ./rustpython-norm benchmarks/pystone.py
  Time (mean ± σ):      2.794 s ±  0.160 s    [User: 2.783 s, System: 0.006 s]
  Range (min … max):    2.597 s …  3.056 s    10 runs
 
Benchmark #2: ./rustpython-opt benchmarks/pystone.py
  Time (mean ± σ):      2.749 s ±  0.103 s    [User: 2.738 s, System: 0.005 s]
  Range (min … max):    2.577 s …  2.897 s    10 runs
 
Summary
  './rustpython-opt benchmarks/pystone.py' ran
    1.02 ± 0.07 times faster than './rustpython-norm benchmarks/pystone.py'
```
cc @pca006132

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -1,5 +1,6 @@
 use std::fmt;
-use std::sync::atomic::{AtomicU32, Ordering};
+#[cfg(feature = "threading")]
+use std::sync::atomic;
 
 use indexmap::IndexMap;
 use itertools::Itertools;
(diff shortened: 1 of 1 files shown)


pca006132:
Actually, I wonder if it would be possible to revise and see how can we make most of the code not multithreaded, and limit atomic/locks to places where they are absolutely needed. Making the structs `impl Sync` costs too much in terms of performance...

coolreader18:
I think we'd like to keep being GIL-less for now, which would involve Sync datatypes and mutexes - Larry Hasting's [Gilectomy talk](https://www.youtube.com/watch?v=P3AyI_u66Bw) is a really good overview of the troubles with doing so, but as a clean-room implementation we have the opportunity to experiment with it and we already have it working. For pure performance, I don't think that's very good, which is the case with Gilectomy as well, but if you need performance I think turning off `threading` is a good solution for now.

youknowone (approved):

PR #2473, opened 2021-02-16
Follow the latest `itertools` module

rickygao:
I make `itertools.accumulate` follow the latest signature `itertools.accumulate(iterable[, func, *, initial=None])`, and split the tests on `accumulate` into two parts.

Now, only `pickletest`s fail.

```
test_accumulate (__main__.TestBasicOps) ... ok
test_accumulate_pickle (__main__.TestBasicOps) ... expected failure
```

### DIFF
--- a/Lib/test/test_itertools.py
+++ b/Lib/test/test_itertools.py
@@ -120,8 +120,6 @@ class TestBasicOps(unittest.TestCase):
             c = expand(compare[took:])
             self.assertEqual(a, c);
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_accumulate(self):
         self.assertEqual(list(accumulate(range(10))),               # one positional arg
                           [0, 1, 3, 6, 10, 15, 21, 28, 36, 45])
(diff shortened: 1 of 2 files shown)


Lib/test/test_itertools.py
```diff
@@ -148,15 +146,19 @@ def test_accumulate(self):
...
-            self.pickletest(proto, accumulate(range(10), initial=7))
         self.assertEqual(list(accumulate([10, 5, 1], initial=None)), [10, 15, 16])
         self.assertEqual(list(accumulate([10, 5, 1], initial=100)), [100, 110, 115, 116])
         self.assertEqual(list(accumulate([], initial=100)), [100])
         with self.assertRaises(TypeError):
             list(accumulate([10, 20], 100))
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    def test_accumulate_pickle(self):
```
fanninpm, line 157: You might want to take this change up with the CPython devs.

coolreader18:
You might want to copy the latest test_itertools.py from CPython

rickygao:
> You might want to copy the latest test_itertools.py from CPython

I'd wonder whether I am expected not to collect the failed tests to another function even if a most part of them pass and the failed ones are all about the same theme.

Lib/test/test_itertools.py
rickygao: Refer to the new commits.

coolreader18 (approved):
Thank you for contributing!

PR #2485, opened 2021-02-19
Misc fixes to get aiohttp working

coolreader18:


fanninpm:
Curious that the `test_unique_composite` unexpectedly succeeds, but only on Windows.

Lib/test/test_enum.py
```diff
@@ -2950,9 +2950,7 @@ class Color(StrMixin, AllMixin, IntFlag):
         self.assertEqual(Color.ALL.value, 7)
         self.assertEqual(str(Color.BLUE), 'blue')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
-    @unittest.skipIf(sys.platform == "win32", "TODO: RUSTPYTHON, universal newlines")
+    @unittest.skip("TODO: RUSTPYTHON, nondeterministic on win32 (universal newlines) fails on other platforms")
```
fanninpm, line 2953: The test you're looking for is actually in a completely different class (around line 2520).

coolreader18: Ah, that's confusing, that would explain why nothing I was doing worked. Thanks :sweat_smile:

vm/build.rs
```diff
@@ -49,6 +49,11 @@ fn main() {
             println!("cargo:rustc-cfg=ossl111");
         }
     }
+    if let Ok(v) = env::var("DEP_OPENSSL_CONF") {
```
youknowone, line 52: updating build document for this?

vm/src/stdlib/json/machinery.rs
```diff
@@ -136,12 +136,12 @@ pub fn scanstring<'a>(
         chunks.push(chunk);
     };
     let unterminated_err = || DecodeError::new("Unterminated string starting at", end - 1);
-    let mut chunk_start = end;
     let mut chars = s.char_indices().enumerate().skip(end).peekable();
-    while let Some((char_i, (i, c))) = chars.next() {
+    let (_, (mut chunk_start, _)) = chars.peek().ok_or_else(unterminated_err)?;
+    while let Some((char_i, (byte_i, c))) = chars.next() {
```
youknowone, line 141: :+1:

youknowone (approved):

vm/build.rs
coolreader18: No, it's set by the openssl-sys crate: https://doc.rust-lang.org/cargo/reference/build-script-examples.html#using-another-sys-crate

PR #2487, opened 2021-02-19
Fix test_enum unexpected success on Windows

fanninpm:


coolreader18:
It might be best just to skip it, my guess is that it's non-deterministic based on hashmap state or something.

fanninpm:
I believe it previously failed due to the presence/absence of the `\r` character.

youknowone (approved):

PR #2491, opened 2021-02-21
Fix other test_enum unexpected success on Windows

fanninpm:
Should alleviate the CI failure seen in #2485.

### DIFF
--- a/Lib/test/test_enum.py
+++ b/Lib/test/test_enum.py
@@ -2518,6 +2518,7 @@ class TestFlag(unittest.TestCase):
 
     # TODO: RUSTPYTHON
     @unittest.expectedFailure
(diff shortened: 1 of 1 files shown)


Lib/test/test_enum.py
```diff
@@ -2518,6 +2518,7 @@ class Color(StrMixin, AllMixin, Flag):
 
     # TODO: RUSTPYTHON
     @unittest.expectedFailure
+    @unittest.skipIf(sys.platform == "win32", "TODO: RUSTPYTHON, universal newlines")
     @support.reap_threads
```
fanninpm, line 2522: The decorators here are applied in this order (at least in CPython):

1. `support.reap_threads()`
2. `unittest.skipIf()`
3. `unittest.expectedFailure()`

coolreader18 (approved):

PR #2501, opened 2021-02-22
Make IntoPyGetter/Setter/DeleterFunc Traits Public

zicklag:
Hey there, I am trying to use RustPython for a Python scripting adapter for a custom scripting system and I had the need to define dynamic getters and setters at runtime that could not be specified as a `Fn` closure or a `fn` function pointer, because they needed to be created with extra internal data. That requires that `IntoPyGetterFunc` and friends had to be public so that I could manually implement them for a custom struct.

For example, here's a super dumb but otherwise impossible to implement getter:

```rust
#[derive(Debug)]
struct ComponentGetter {
    key: String,
    value: String,
}

impl ComponentGetter {
    pub fn new(key: &str, value: &str) -> Self {
        Self {
            key: key.into(),
            value: value.into(),
        }
    }
}

impl IntoPyGetterFunc<Component> for ComponentGetter {
    fn get(&self, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult {
        println!("Getting key! — {}", self.key);

        Ok(vm.ctx.new_str(&self.value))
    }
}
```

Because I'm defining my classes dynamically at runtime, I have to be able to create getters and setters for attributes that I have discovered at runtime, and I couldn't use a closure or a function pointer because I need to be able to pass in extra data to the getter such as the `key` and `value` data in the dummy `ComponentGetter` example above.

Let me know if you have any questions/concerns!

### DIFF
--- a/vm/src/builtins/mod.rs
+++ b/vm/src/builtins/mod.rs
@@ -29,7 +29,7 @@ pub use function::PyFunction;
 pub(crate) mod generator;
 pub use generator::PyGenerator;
 pub(crate) mod getset;
-pub use getset::PyGetSet;
+pub use getset::{IntoPyDeleterFunc, IntoPyGetterFunc, IntoPySetterFunc, PyGetSet};
 pub(crate) mod int;
 pub use int::{PyInt, PyIntRef};
 pub(crate) mod iter;


coolreader18:
I'm not opposed to making the traits public, but I believe you should be able to do that with a closure - try

```rust
fn make_component_getter(name: &str, key: &str, value: &str, ctx: &PyContext) -> PyObjectRef {
    let key = key.to_owned();
    let value = value.to_owned();
    ctx.new_readonly_getset(name, move |obj: PyObjectRef, vm: &VirtualMachine| -> String {
        println!("getting key");
        key.clone()
    })
}
```
or even -
```rust
#[derive(Debug)]
struct ComponentGetter {
    key: String,
    value: String,
}

impl ComponentGetter {
    pub fn new(key: &str, value: &str) -> Self {
        Self {
            key: key.into(),
            value: value.into(),
        }
    }

    fn get(&self, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult {
        println!("Getting key! — {}", self.key);

        Ok(vm.ctx.new_str(&self.value))
    }

    fn into_getter(self) -> impl Fn(PyObjectRef, &VirtualMachine) -> PyResult { // i.e. -> impl IntoPyGetterFunc
        move |obj, vm| self.get(obj, vm)
    }
}
```

zicklag:
I tried the closure, and it didn't work because taking ownership of values from outside the closure makes the closure `FnOnce`. I didn't try it the other way, but I think it'd have the same problem.

Either way, I think that the most intuitive way from an API perspective would be to just to implement the trait, anyway.

coolreader18:
> taking ownership of values from outside the closure makes the closure `FnOnce`

That's not true, it's not FnOnce if you don't consume the values. Try it, it should work. Thinking about it more, I think I'd actually prefer not to publicize `IntoPyGetterFunc`/make it possible to implement, that Kind parameter is pretty ugly and it's supposed to represent a *function*, not an arbitrary type - just like the `Fn{,Once,Mut}` traits aren't possible to implement for that reason.

zicklag:
Ah, it did work! I tried to copy your code last night, and couldn't get it to compile, but it wasn't because of the closure issue, I was just confused. I'm not sure what I did, but now it works! And I learned something about closures.

That covers my use-case and makes this unnecessary. I agree that the type parameter is really confusing. I had actually misunderstood what it was for when I used it, it just worked so I thought that it was fine.

If I get the time I might open a PR with something like you example in a doc comment.

---

On that note, actually, what do you think of having a RustPython mdbook? I'm going to likely be spending some time with RustPython as I try to make a scripting solution for Rust programs and I might be able to find the time to write a user-guide on how to use RustPython ( likely in the context of embedding for now ) + maybe some API docs.

I found it a little difficult to figure out how to do certain things and felt like I could use some documentation, but now I'm getting a better understanding and I figure I might be able to write it. :stuck_out_tongue_winking_eye:

coolreader18:
Yeah, if you want to write some stuff for it that would be very much appreciated!

PR #2515, opened 2021-03-08
Implement dir_fd in os.open

deantvv:
Now pass `test_posix.py PosixTester.test_open_dir_fd`

Implement **dir_fd** in
`make_path` and `os.open(path, flags, mode=0o777, *, dir_fd=None)`

### DIFF
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -84,6 +84,7 @@ atty = "0.2"
 static_assertions = "1.1"
 half = "1.6"
 memchr = "2"
+filepath = "0.1.1"
 
 # RustPython crates implementing functionality based on CPython
 mt19937 = "2.0"
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -123,7 +123,6 @@ openssl = { version = "0.10.32", features = ["vendored"], optional = true }
 openssl-sys = { version = "0.9", optional = true }
 openssl-probe = { version = "0.1", optional = true }
 which = "4.0"
-filepath = "0.1.1"
 
 [target.'cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))'.dependencies]
 num_cpus = "1"
@@ -147,6 +146,9 @@ features = [
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 wasm-bindgen = "0.2"
+
+[target.'cfg(any(target_os = "linux", target_os = "macos", windows))'.dependencies]
+filepath = "0.1.1"
 
 [build-dependencies]
 itertools = "0.9"
--- a/vm/Cargo.toml
+++ b/vm/Cargo.toml
@@ -84,7 +84,6 @@ atty = "0.2"
 static_assertions = "1.1"
 half = "1.6"
 memchr = "2"
-filepath = "0.1.1"
 
 # RustPython crates implementing functionality based on CPython
 mt19937 = "2.0"
@@ -124,6 +123,7 @@ openssl = { version = "0.10.32", features = ["vendored"], optional = true }
 openssl-sys = { version = "0.9", optional = true }
 openssl-probe = { version = "0.1", optional = true }
 which = "4.0"
+filepath = "0.1.1"
 
 [target.'cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))'.dependencies]
 num_cpus = "1"
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -7,6 +7,7 @@ use std::time::{Duration, SystemTime};
 use std::{env, fs};
 
 use crossbeam_utils::atomic::AtomicCell;
+use filepath::FilePath;
 use num_bigint::BigInt;
 
 use super::errno::errors;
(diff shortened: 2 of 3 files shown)


vm/src/stdlib/os.rs
```diff
@@ -157,16 +159,24 @@ impl TryFromObject for PyPathLike {
...
-        Err(vm.new_os_error("dir_fd not supported yet".to_owned()))
-    } else {
-        Ok(path.path.as_os_str())
+fn make_path(vm: &VirtualMachine, path: &PyPathLike, dir_fd: &DirFd) -> PyResult<ffi::OsString> {
+    let path = path.path.to_path_buf();
+    if dir_fd.0.is_none() | path.is_absolute() {
+        return Ok(path.into_os_string());
+    }
+
+    if cfg!(target_os = "wasi") {
```
kenkoooo, line 168: I guess we can use `cfg_if!` here.

deantvv: Use `cfg_if!` in cdcc191

deantvv:
Related to #1671

deantvv:
@coolreader18 Sorry for pinging you out of nowhere but could you please take a look at this? Is there anything I could change to make this better? Thanks in advanced.

vm/Cargo.toml
```diff
@@ -123,6 +123,7 @@ openssl = { version = "0.10.32", features = ["vendored"], optional = true }
 openssl-sys = { version = "0.9", optional = true }
 openssl-probe = { version = "0.1", optional = true }
 which = "4.0"
+filepath = "0.1.1"
```
coolreader18, line 126: It looks like this only supports windows/mac/linux, so it might be good to put the feature gate behind `[target.'cfg(any(target_os = "linux", target_os = "macos", windows))'.dependencies]`

vm/src/stdlib/os.rs
```diff
@@ -157,15 +159,25 @@ impl TryFromObject for PyPathLike {
...
-fn make_path<'a>(
-    vm: &VirtualMachine,
-    path: &'a PyPathLike,
-    dir_fd: &DirFd,
-) -> PyResult<&'a ffi::OsStr> {
-    if dir_fd.0.is_some() {
-        Err(vm.new_os_error("dir_fd not supported yet".to_owned()))
-    } else {
-        Ok(path.path.as_os_str())
+fn make_path(vm: &VirtualMachine, path: &PyPathLike, dir_fd: &DirFd) -> PyResult<ffi::OsString> {
```
coolreader18, line 162: You should be able to return `std::borrow::Cow<ffi::OsStr>`, since unless dir_fd is `Some` you can just borrow from `path`

coolreader18 (changes requested):
I think this is good idea for a "quick and dirty" handling of dir_fd! I think eventually it'd be good to use stuff like `openat` that handle dir_fd natively, but for now this definitely addresses the lack of support.

vm/src/stdlib/os.rs
deantvv: Thanks for reviewing the code. It is fixed in commits below.

vm/Cargo.toml
deantvv: Thanks and fixed!

vm/src/stdlib/os.rs
```diff
@@ -161,11 +163,25 @@ fn make_path<'a>(
...
     dir_fd: &DirFd,
-) -> PyResult<&'a ffi::OsStr> {
-    if dir_fd.0.is_some() {
-        Err(vm.new_os_error("dir_fd not supported yet".to_owned()))
-    } else {
-        Ok(path.path.as_os_str())
+) -> PyResult<std::borrow::Cow<'a, ffi::OsStr>> {
+    let path = &path.path;
+    if dir_fd.0.is_none() | path.is_absolute() {
+        return Ok(std::borrow::Cow::Borrowed(path.as_os_str().into()));
```
coolreader18, line 169: It looks like the unnecessary `.into()` here is causing the CI failure

vm/src/stdlib/os.rs
```diff
@@ -161,11 +163,25 @@ fn make_path<'a>(
...
+    }
+
+    cfg_if::cfg_if! {
+        if #[cfg(target_os = "wasi")] {
```
coolreader18, line 173: Also, the `cfg`s you added here should match the cfg in Cargo.toml

vm/src/stdlib/os.rs
```diff
@@ -161,11 +163,25 @@ fn make_path<'a>(
...
     path: &'a PyPathLike,
     dir_fd: &DirFd,
-) -> PyResult<&'a ffi::OsStr> {
-    if dir_fd.0.is_some() {
-        Err(vm.new_os_error("dir_fd not supported yet".to_owned()))
-    } else {
-        Ok(path.path.as_os_str())
+) -> PyResult<std::borrow::Cow<'a, ffi::OsStr>> {
+    let path = &path.path;
+    if dir_fd.0.is_none() | path.is_absolute() {
```
coolreader18, line 168:
```suggestion
    if dir_fd.0.is_none() || path.is_absolute() {
```

coolreader18 (approved):

Issue #2406, opened 2021-01-23
Pasting multiple lines into the REPL doesn't split the lines into multiple statements


verhovsky:
Open the RustPython REPL, paste the following code (and press Enter):

```
print('hello')
with open('/tmp/file.txt', 'w') as f:
     f.write('just a test')
```

on CPython I get

```
>>> print('hello')
hello
>>> with open('/tmp/file.txt', 'w') as f:
...      f.write('just a test')
... 
11
>>> 
```

On RustPython I get

```
>>>>> print('hello')
with open('/tmp/file.txt', 'w') as f:
     f.write('just a test')

hello
>>>>> 
```

when I should get

```
>>>>> print('hello')
hello
>>>>> with open('/tmp/file.txt', 'w') as f:
     f.write('just a test')
11
>>>>> 
```

TheOnlyError:
Currently working on this! Opening a PR later today :).

verhovsky:
while you're at it, it would be nice to have `.....` on continuation lines, just like CPython has `...`.

coolreader18:
@verhovsky don't we already?

```python
❯ rustpython
Welcome to the magnificent Rust Python 0.1.2 interpreter 😱 🖖
>>>>> class A:
.....     def foo():
.....         pass
..... 
>>>>> 
```

coolreader18:
oh nevermind you mean specifically when pasting

PR #2519, opened 2021-03-19
2406 parse multiple lines REPL

TheOnlyError:
Fixes [the parsing of multiple lines into the REPL](https://github.com/RustPython/RustPython/issues/2406).

Updates RustyLine to version 8.0.0 and turns off [bracketed paste](https://docs.rs/rustyline/8.0.0/rustyline/config/struct.Builder.html#method.bracketed_paste) to do so. The respective test is currently marked as ignored due to an [issue ](https://github.com/kkawakam/rustyline/issues/504)with RustyLine.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -654,6 +654,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "dirs-next"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b98cf8ebf19c3d1b223e151f99a4f9f0690dca41414773390fc824184ac833e1"
+dependencies = [
+ "cfg-if 1.0.0",
+ "dirs-sys-next",
+]
+
+[[package]]
 name = "dirs-sys-next"
 version = "0.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


coolreader18:
I guess one way to do it would just be to spawn `rustpython` in a pty and check that the output is as you expect, but I don't know how to have the binary be a dependency of the tests.

coolreader18:
Maybe something like this: https://rust-cli.github.io/book/tutorial/testing.html#testing-cli-applications-by-running-them

Pluriscient:
> Maybe something like this: https://rust-cli.github.io/book/tutorial/testing.html#testing-cli-applications-by-running-them

These methods seem to work best when the package the test is in is the one being tested, at least `cargo_bin` doesn't seem to recognize the `rustpython` crate from within `vm`

coolreader18:
Well you could add a separate `tests` directory in the root of the repository, so that would be integration tests for `rustpython` itself

coolreader18 (approved):
I think this looks good, thanks for contributing

Pluriscient:
> 
> 
> I think this looks good, thanks for contributing! We can figure out how to add tests later, if that sounds good?

I agree, I've create #2526 as a separate issue for that 👍

coolreader18:
Yeah, that sounds good

Pluriscient (approved):

TheOnlyError:
Should be done now!

coolreader18:
Thanks!

PR #2530, opened 2021-03-26
remove noise from importing modules in whats_left

verhovsky:
some modules will print something to stdout when imported

### DIFF
--- a/extra_tests/not_impl_gen.py
+++ b/extra_tests/not_impl_gen.py
@@ -7,11 +7,13 @@
 # cpymods - a dictionary mapping module names to their contents
 # libdir - the location of RustPython's Lib/ directory.
 
-import re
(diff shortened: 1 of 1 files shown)


verhovsky:
If you look at https://github.com/RustPython/rustpython.github.io/blob/dac2a94d0efde5c4330ac08ff90b2fad2f59e472/_data/whats_left.temp it starts like this:

```
('Constant String', 2, None, 4)
Hello world!
bytearray.__alloc__
bytearray.__getattribute__ inherited
bytearray.__init__ inherited
```

This PR should remove the first 2 lines.

verhovsky:
The CI failures don't seem to be related to this PR.

coolreader18 (approved):
LGTM!

Issue #2522, opened 2021-03-21
[RFC] Documentation


Pluriscient:
# Documentation overview

## Summary

Documentation is in need of improvement. There are a number of components that have little to no documentation, and this makes the project harder to understand for newcomers. 

## Detailed Explanation

While #1677 exists, I believe it is better to create a more fleshed out issue for just general documentation as opposed to more detailed explanations which I believe that is a reference to.

As a checklist, these are the crates, though this can be fleshed out a lot:

- [ ] AST
- [ ] Bytecode 
- [ ] Common
- [ ] Compiler
- [ ] Derive
- [ ] Parser
- [ ] JIT
- [ ] VM

Which ones would be most helpful to start with? 

## Current State

I ran rustdoc with the nightly code-coverage on all the crates and logged the amount of warnings
`cargo rustdoc -- -Z unstable-options --show-coverage -W missing_docs`
<details>

<summary>All Crates</summary>

### Main

<details>

<summary>Table</summary>

| File | Documented | Percentage | Examples | Percentage |
| ---- | ---------- | ---------- | -------- | ---------- |
| src/lib.rs                          |          2 |     100.0% |          1 |      50.0% || Total                               |          2 |     100.0% |          1 |      50.0% |

</details>

No warnings about missing docs

### AST

<details>

<summary>Table</summary>

| File | Documented | Percentage | Examples | Percentage |
| ---- | ---------- | ---------- | -------- | ---------- |
| ast/src/ast_gen.rs                  |          0 |       0.0% |          0 |       0.0% |
| ast/src/constant.rs                 |          4 |      25.0% |          0 |       0.0% |
| ast/src/lib.rs                      |          0 |       0.0% |          0 |       0.0% |
| ast/src/location.rs                 |          2 |     100.0% |          0 |       0.0% |
| ast/src/map_ast.rs                  |          0 |       0.0% |          0 |       0.0% |
| Total                               |          6 |       2.0% |          0 |       0.0% |

</details>

299 warnings for missing documentation

### Benches

<details>

<summary>Table</summary>

| File | Documented | Percentage | Examples | Percentage |
| ---- | ---------- | ---------- | -------- | ---------- |
| src/lib.rs                          |          2 |     100.0% |          1 |      50.0% |
| Total                               |          2 |     100.0% |          1 |      50.0% |

</details>

### Bytecode
<details>

<summary>Table</summary>

| File | Documented | Percentage | Examples | Percentage |
[...]

PR #2532, opened 2021-03-27
Documentation of VM & Bytecode

Pluriscient:
As described in #2522, it's nice to get some documentation going.
This is my first crack at it, mostly focused on the bytecode crate. I'd really appreciate any feedback on it!

I've left some TODOs that can either be fixed in this PR or be left for future ones. In particular I believe concrete examples would help a lot but I don't think I can construct realistic example scenarios, maybe there are some suggestions?

### DIFF
--- a/bytecode/src/lib.rs
+++ b/bytecode/src/lib.rs
@@ -143,8 +143,8 @@ pub struct CodeObject<C: Constant = ConstantData> {
     pub cell2arg: Option<Box<[isize]>>,
     pub constants: Box<[C]>,
     #[serde(bound(
-    deserialize = "C::Name: serde::Deserialize<'de>",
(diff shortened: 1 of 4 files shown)


coolreader18:
@Pluriscient do you think this is good for the most part? I see there are still a couple TODO comments, but if you're not able to continue this, just removing the TODOs would be fine.

Pluriscient:
> 
> 
> @Pluriscient do you think this is good for the most part? I see there are still a couple TODO comments, but if you're not able to continue this, just removing the TODOs would be fine.

I've just done this as I'm in the middle of a busy week and won't be able to complete them now. I'll make a list of questions to help me resolve these TODOs when I am less overloaded :)

coolreader18 (approved):

Issue #2396, opened 2021-01-08
Automate generation of benchmarks and display them on website


mireille-raad:
This PR https://github.com/RustPython/RustPython/pull/2367 by @orf refactors the benchmark suite to use Criterion. It also adds micro-benchmarks.

It would be neat if we:
- [ ] select a set of benchmarks
- [ ] automate their generation (maybe once a month). 
- [ ] save the output (data or images) to https://github.com/RustPython/rustpython.github.io.
- [ ] create a webpage to show the results

nielsbauman:
Hi there, I'm trying to implement this together with @theangryhobbit. We're currently discussing which benchmarks we should select, so we'd like some input from others as well. Perhaps you (@mireille-raad), @coolreader18 and @youknowone could weigh in on how we should select a set of benchmarks. The ones that are currently generated/available are:

![image](https://user-images.githubusercontent.com/33722607/111687935-2e7f5480-882b-11eb-9589-de6e03467ae4.png)

I'd say all the 'violin' plots are useful as they provide direct comparisons with the CPython implementation. This is an example of one:

![image](https://user-images.githubusercontent.com/33722607/111689238-122fe780-882c-11eb-92fd-a2d6e5964b17.png)

Do we then just show _all_ violin plots (execution, microbenchmarks, parse_to_ast and pystone)?

We'd be interested to hear your take on this.

coolreader18:
I think it's not a bad idea to show the same metrics for all the benchmarks, they're all pretty varied in what specifically they show so it's good to be able to see the comparisons for all the different aspects of the interpreters.

mireille-raad:
Thank you @nielsbauman for looking at this. I have no strong opinion about this and would defer to what @coolreader18  and @youknowone would pick.

I will be more interested in the organization of the output and the format in which data can be saved. I want to make sure we can easily integrate outputs with the website (jekyll) and that over time it is easy to scale/add more things.

youknowone:
I think showing all plots will be good unless we collect a lot more benchmarks in future. Thank you for the work, it looks great.

PR #2533, opened 2021-03-28
Run benchmarks in CI

nielsbauman:
The first half of #2396:
- run benchmarks
- restructure the folders and rename the generated SVG's to simplify displaying them
- upload/commit the generated data to [RustPython/rustpython.github.io](https://github.com/RustPython/rustpython.github.io)

The second half (actually displaying the data) will be implemented in [RustPython/rustpython.github.io](https://github.com/RustPython/rustpython.github.io).

I considered moving the `restructure generated files` 'run' to a separate script file to add some comments, but I decided that that was not 'worth' the extra file. Be sure to let me know if people have a different opinion on this.

The for loop renames files as follows: `./execution/mandelbrot.py/report/violin.svg` -> `./execution/mandelbrot.py.svg`.

### DIFF
--- a/.github/workflows/cron-ci.yaml
+++ b/.github/workflows/cron-ci.yaml
@@ -119,3 +119,37 @@ jobs:
           git add -A
           git -c user.name="Github Actions" -c user.email="actions@github.com" commit -m "Update what is left results" --author="$GITHUB_ACTOR"
           git push
+
+  benchmark:
+    name: Collect benchmark data
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+      - name: collect execution benchmark data
+        run: cargo bench --bench execution
+      - name: collect microbenchmarks data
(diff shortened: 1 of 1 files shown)


theangryhobbit:
The second half (actually displaying the data) for [RustPython/rustpython.github.io](https://github.com/RustPython/rustpython.github.io) has just been implemented and [this PR](https://github.com/RustPython/rustpython.github.io/pull/43) has been opened to address this.

coolreader18 (approved):
I think this looks good! @nielsbauman do you think this is ready to merge or is there still work to do on it?

nielsbauman:
@coolreader18 great! I think it's ready to be merged already. I copied most of the data uploading part from other runs/actions, so we shouldn't run into too much trouble there. Also, I coordinated the file restructuring/renaming with @theangryhobbit so that should also already be working fine.

PR #2538, opened 2021-04-04
os: Add support for fd and dir_fd in `listdir`, `stat`, `fstat`

deantvv:
This PR would like to add support for `os.fwalk` which need 
`{open, stat} <= supports_dir_fd and {listdir, stat} <= supports_fd` 

So I add support for dir_fd and fd for each functions.

In listdir, I use same approach as make_path to add support_fd.

Noted that, test coverage doesn't increase due to the lack of 3rd argument in `os.symlink`.

### DIFF
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -3958,8 +3958,8 @@ class TestScandir(unittest.TestCase):
             self.assertIs(type(entry.name), bytes)
             self.assertIs(type(entry.path), bytes)
 
-    @unittest.skipUnless(os.listdir in os.supports_fd,
-                         'fd support for listdir required for this test.')
+    @unittest.skipUnless(os.scandir in os.supports_fd,
+                         'fd support for scandir required for this test.')
     def test_fd(self):
         self.assertIn(os.scandir, os.supports_fd)
         self.create_file('file.txt')
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -3958,6 +3958,7 @@ class TestScandir(unittest.TestCase):
             self.assertIs(type(entry.name), bytes)
             self.assertIs(type(entry.path), bytes)
 
+    # TODO: RUSTPYTHON (scandir needs to have supports_fd)
     @unittest.skipUnless(os.scandir in os.supports_fd,
                          'fd support for scandir required for this test.')
     def test_fd(self):
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -170,7 +170,7 @@ fn path_from_fd(raw_fd: i64) -> Result<PathBuf, String> {
             };
             Ok(path)
         } else {
-            Err(vm.new_os_error("fd not supported on wasi yet".to_owned()));
+            Err("fd not supported on wasi yet");
         }
     }
 }
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -159,6 +159,22 @@ impl TryFromObject for PyPathLike {
     }
 }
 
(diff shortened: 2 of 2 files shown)


vm/src/stdlib/os.rs
```diff
@@ -555,8 +555,30 @@ mod _os {
...
+                                return Err(vm.new_os_error(format!(
+                                    "Cannot determine path of fno: {:?}",
+                                    fno
+                                )));
+                            }
+                        }
+                    } else {
+                        return Err(vm.new_os_error("dir_fd not supported on wasi yet".to_owned()));
+                    }
+                }
```
coolreader18, line 577: I think you could just call `make_path` here - then for the other, do `path.path.into()` to convert it to a `Cow<Path>`

coolreader18, line 577: I think you could just call `make_path` here - then for the other, do `path.path.into()` to convert it to a `Cow<Path>`

coolreader18: Oh, wait, I see. Maybe just deduplicate this into another function that's called from both here and make_path?

deantvv: Thanks for reviewing. I extract the logic to one function as suggested.

vm/src/stdlib/os.rs
```diff
@@ -159,6 +159,22 @@ impl TryFromObject for PyPathLike {
...
+                Ok(path) => path,
+                Err(_) => {
+                    return Err(format!("Cannot determine path of fd: {:?}", raw_fd));
+                }
+            };
+            Ok(path)
+        } else {
+            Err(vm.new_os_error("fd not supported on wasi yet".to_owned()));
+        }
+    }
```
coolreader18, line 175: This won't compile on wasi I think - it tries to call something on vm and it returns a py exception instead of a string - but I think it's fine to just accept `vm` as an argument and return `PyResult<PathBuf>`

deantvv: I remove the `vm` in `path_from_fd` and let the caller decide what to do with the error (`String`).

vm/src/stdlib/os.rs
```diff
@@ -159,6 +159,24 @@ impl TryFromObject for PyPathLike {
...
+            let path = match file.path() {
+                Ok(path) => path,
+                Err(e) => {
+                    return Err(format!("{:?} Cannot determine path of fd: {:?}", e, raw_fd));
+                }
+            };
+            raw_file_number(file);  // Do not consume `raw_fd`
+            Ok(path)
+        } else {
+            Err("fd not supported on wasi yet");
```
coolreader18, line 175:
```suggestion
            Err("fd not supported on wasi yet".to_owned())
```

Lib/test/test_os.py
```diff
@@ -3958,8 +3958,8 @@ def test_bytes_like(self):
             self.assertIs(type(entry.name), bytes)
             self.assertIs(type(entry.path), bytes)
 
-    @unittest.skipUnless(os.listdir in os.supports_fd,
-                         'fd support for listdir required for this test.')
+    @unittest.skipUnless(os.scandir in os.supports_fd,
```
coolreader18, line 3961: Could you put a TODO comment here for changing it back to `listdir` once scandir has supports_fd?

vm/src/stdlib/os.rs
deantvv: Thanks for catching this.

Lib/test/test_os.py
deantvv: Yes, I should do this.

coolreader18 (approved):
Thank you for contributing!

PR #2539, opened 2021-04-04
Add os in detect requires environment script

deantvv:
Before this patch, you need to set PYTHONHOME to run the test suite.
Now it will automatically detect whether you need environment for your interpreter.


coolreader18 (approved):
Nice work, I think I've run into this before!

PR #2540, opened 2021-04-04
Add the ast::fold::Fold trait, replacing MapAst

coolreader18:
cc @pca006132 - you'd still be able to do what you wanted the custom field for, with this instead of MapAst, right?


pca006132:
> cc @pca006132 - you'd still be able to do what you wanted the custom field for, with this instead of MapAst, right?

Looks good to me, but haven't tried yet. I would try this later today.

ast/asdl_rs.py
```diff
@@ -243,84 +253,111 @@ def visitProduct(self, product, name, depth):
...
-            return
-
-
-        self.emit(f"impl<T, U> MapAst<T, U> for {enumname}{apply_t} {{", depth)
+        self.emit(f"impl<T, U> Foldable<T, U> for {enumname}{apply_t} {{", depth)
         self.emit(f"type Mapped = {enumname}{apply_u};", depth + 1)
-        self.emit("fn try_map_ast<E, F: FnMut(T) -> Result<U, E>>(self, f: &mut F) -> Result<Self::Mapped, E> {", depth + 1)
-        self.emit("match self {", depth + 2)
+        self.emit("fn fold<F: Fold<T, TargetU = U> + ?Sized>(self, folder: &mut F) -> Result<Self::Mapped, F::Error> {", depth + 1)
+        self.emit(f"fold_{name}(folder, self)", depth + 2)
```
pca006132, line 295: Shouldn't we use `folder.fold_{name}(self)` here? Otherwise, overriding the default implementation would not be useful as the default one would be called.

Also, I think we should add test cases for nested structure, for the constant propagation test. As the bug mentioned above is not revealed by current test case.

coolreader18: Oh, very good point, thanks.

pca006132:
Looks good to me now.

youknowone (approved):

pca006132:
Hi, just wondering when would the new version of rustpython-parser be released to crates.io? Not sure if my project should depend on the github version or the crates.io version for now.

PR #2544, opened 2021-04-07
Misc fixes related to IPython

coolreader18:


coolreader18:
Interesting - it looks like CPython does this by just setting the entry's key+value to NULL and just letting it sit there, and only on "resize" does it get rid of the unused entry and recompact. That might be something to try, if you want, or else this looks fine to me.

youknowone (approved):

youknowone:
the code looks good but I have no idea what "CPython does this" means from your comment

coolreader18:
Oh, man, I commented on completely the wrong issue there 😅

PR #2548, opened 2021-04-10
Fix os.symlink 3rd argument

deantvv:
From [python doc](https://docs.python.org/3/library/os.html?highlight=os%20symlink#os.symlink),
`os.symlink(src, dst, target_is_directory=False, *, dir_fd=None)`
which means that `target_is_directory` is a positional argument.

Related to #1175

### DIFF
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -1013,7 +1013,6 @@ from test import mapping_tests
 #         self._test_environ_iteration(os.environ.values())
 
 
-@unittest.skip("TODO: RUSTPYTHON, 3rd arg to symlink")
 class WalkTests(unittest.TestCase):
     """Tests for os.walk()."""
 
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/os.rs
```diff
@@ -260,7 +260,7 @@ pub fn errno_err(vm: &VirtualMachine) -> PyBaseExceptionRef {
 #[allow(dead_code)]
 #[derive(FromArgs, Default)]
 pub struct TargetIsDirectory {
-    #[pyarg(named, default = "false")]
+    #[pyarg(positional, default = "false")]
```
coolreader18, line 263: I think this should actually be `any` - it looks like it can accept `target_is_directory=` as a keyword argument too.

deantvv: Yes it should be `any` and now it is! Thanks for catching this.

Lib/test/test_os.py
```diff
@@ -1295,7 +1294,7 @@ def walk(self, top, **kwargs):
             bdirs[:] = list(map(os.fsencode, dirs))
             bfiles[:] = list(map(os.fsencode, files))
 
-@unittest.skipUnless(hasattr(os, 'fwalk'), "Test needs os.fwalk()")
+
```
coolreader18, line 1297: I don't think you need to remove this, since the `os` module *does* define fwalk on some unixes, but this won't necessarily work when `os` doesn't define fwalk (i.e. when `not ({open, stat} <= supports_dir_fd and {listdir, stat} <= supports_fd)`)

coolreader18:
Oh lol, good timing.

Lib/test/test_os.py
deantvv: Yes, this line should be kept as is.

coolreader18 (approved):

deantvv:
> 
> 
> LGTM, thank you for contributing!

Thank you for reviewing this!

PR #2549, opened 2021-04-11
Refactor sequence owner name into trait

fermian:
This pull request adds the following changes:

 - Refactors owner_type argument in PySliceableSequence's indexing functions into a separate PySliceableSequenceOwner trait
 - Makes setitem tests in list_tests.py, test_bytes.py files work

### DIFF
--- a/vm/src/builtins/bytearray.rs
+++ b/vm/src/builtins/bytearray.rs
@@ -26,8 +26,8 @@ use crate::slots::{
 use crate::utils::Either;
 use crate::vm::VirtualMachine;
 use crate::{
-    IdProtocol, IntoPyObject, PyClassImpl, PyComparisonValue, PyContext, PyIterable, PyObjectRef,
-    PyRef, PyResult, PyValue, TypeProtocol,
+    IdProtocol, IntoPyObject, PyClassDef, PyClassImpl, PyComparisonValue, PyContext, PyIterable,
+    PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
 };
 use bstr::ByteSlice;
 use crossbeam_utils::atomic::AtomicCell;
@@ -151,11 +151,11 @@ impl PyByteArray {
     #[pymethod(magic)]
     fn setitem(
         zelf: PyRef<Self>,
-        needle: SequenceIndex,
+        needle: PyObjectRef,
         value: PyObjectRef,
         vm: &VirtualMachine,
     ) -> PyResult<()> {
-        match needle {
+        match SequenceIndex::try_from_object_for(vm, needle, Self::NAME)? {
             SequenceIndex::Int(i) => {
                 let value = value_from_object(vm, &value)?;
                 let mut elements = zelf.borrow_buf_mut();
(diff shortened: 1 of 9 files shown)


vm/src/sliceable.rs
```diff
@@ -208,6 +208,22 @@ impl<T: Clone> PySliceableSequenceMut for Vec<T> {
     }
 }
 
+pub trait PySliceableSequenceOwner {
+    fn owner_type() -> &'static str;
```
coolreader18, line 212: You can make this an associated constant - `const OWNER_TYPE: &'static str;`. Then the macro can be `<$type as $crate::pyobject::PyClassDef>::NAME`

coolreader18:
Is it really necessary to have it be a trait? I think I understand why you did it, since it's less "stringly-typed"/error prone since you don't have to type the exact same thing, just pass `::<Self>`, but I think as it is now it's duplicating the code in the function for each generic even though in theory it just has to take a string argument.

fermian:
I'm honestly not sure, I thought it would be cleaner to express it with a trait, but I can see your point. Maybe Rust is smart enough to optimize the specialized template functions away, but if not, I'll push a patch that only fixes the test cases.

youknowone:
`cargo bloat` could be helpful to check if the code is duplicated or not. (though not sure it matters that much)

I think `PySliceableSequenceOwner` approach is useful to organize the data and could be repeated in a few more situations like this.
But generic `getitem` can make actual problems in similar situation. So if we care about the problem, trait looks ok but the generic functions are not.
Maybe just `getitem(Self::OWNER_TYPE_NAME, ...)` would be better.

vm/src/sliceable.rs
```diff
@@ -335,10 +348,10 @@ pub enum SequenceIndex {
 }
 
 impl SequenceIndex {
-    fn try_from_object_for(
+    pub fn try_from_object_for(
+        owner_type: &'static str,
```
youknowone, line 352: I think the parameter order of this function shouldn't be changed.
We conventionally keep `vm`'s position as the first or the last. @coolreader18 is it?

coolreader18: Yeah, I think it could probably be back to how it was before, `vm obj owner_type`

vm/src/builtins/bytearray.rs
```diff
@@ -21,7 +21,9 @@ use crate::pyobject::{
     BorrowValue, Either, IdProtocol, IntoPyObject, PyClassImpl, PyComparisonValue, PyContext,
     PyIterable, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
 };
-use crate::sliceable::{PySliceableSequence, PySliceableSequenceMut, SequenceIndex};
+use crate::sliceable::{
+    PySliceableSequence, PySliceableSequenceMut, PySliceableSequenceOwner, SequenceIndex,
```
coolreader18, line 25: I wonder if we could just import `PyClassDef` and instead of `Self::OWNER_TYPE` just use the `Self::NAME` that's already defined on `PyClassDef`

fermian: If that's the case, I probably should redo this patch, remove the trait and only fix the tests.

coolreader18: Yeah, I think that might be best. Sorry for leading you on a sort of wild goose chase :sweat_smile:

youknowone:
@fermian are you going to finish this PR? I hope we didn’t depress you for the reviews

youknowone (approved):

youknowone:
@fermian I finished it based on your work, thank you!

PR #2550, opened 2021-04-12
Retain order of keys in dict when deleting a key

fermian:
This patch makes dictionaries retain the order of elements after deleting a key (the correct behavior in CPython).

### DIFF
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -557,10 +557,13 @@ impl<T: Clone> Dict<T> {
         let removed = if entry_index == inner.used {
             inner.entries.pop().unwrap()
         } else {
-            let last_index = inner.entries.last().unwrap().index;
-            let removed = inner.entries.remove(entry_index);
-            inner.indices[last_index] = IndexEntry::FREE;
-            removed
+            let entries_rem = entry_index + 1;
+            let entries_len = inner.entries.len();
+            for move_entry_idx in entries_rem..entries_len {
+                let index_index = inner.entries[move_entry_idx].index;
+                inner.indices[index_index] = move_entry_idx as i64 - 1;
+            }
+            inner.entries.remove(entry_index)
         };
         Ok(Some(removed))
     }
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -558,8 +558,8 @@ impl<T: Clone> Dict<T> {
             inner.entries.pop().unwrap()
         } else {
             let last_index = inner.entries.last().unwrap().index;
-            let removed = inner.entries.swap_remove(entry_index);
-            inner.indices[last_index] = entry_index as i64;
+            let removed = inner.entries.remove(entry_index);
+            inner.indices[last_index] = IndexEntry::FREE;
             removed
         };
         Ok(Some(removed))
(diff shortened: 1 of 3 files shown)


coolreader18:
I don't think it's so simple, since there's still indices in `indices` pointing to the "old" positions of the entries. I think you'd have to recompute/shift over the `indices` vector as well.

coolreader18:
But then maybe that would conflict with something in the hash/bucket algorithm? I'm not sure, hashmaps are complicated :upside_down_face:

fermian:
> I don't think it's so simple, since there's still indices in `indices` pointing to the "old" positions of the entries. I think you'd have to recompute/shift over the `indices` vector as well.

I think I'll do that. I'm new to this code base so I'm not sure how the dict implementation works yet :P

I'm assuming the indices vector stores a mapping from the hash table's keys(?) to its corresponding index in the entries vector.

vm/src/dictdatatype.rs
```diff
@@ -557,10 +557,13 @@ impl<T: Clone> Dict<T> {
         let removed = if entry_index == inner.used {
             inner.entries.pop().unwrap()
         } else {
-            let last_index = inner.entries.last().unwrap().index;
-            let removed = inner.entries.swap_remove(entry_index);
```
youknowone, line 561: oh... I believed it is not swap_remove and we were doing it as how CPython does.

vm/src/dictdatatype.rs
```diff
@@ -557,10 +557,13 @@ impl<T: Clone> Dict<T> {
...
-            inner.indices[last_index] = entry_index as i64;
-            removed
+            let entries_rem = entry_index + 1;
+            let entries_len = inner.entries.len();
+            for move_entry_idx in entries_rem..entries_len {
+                let index_index = inner.entries[move_entry_idx].index;
+                inner.indices[index_index] = move_entry_idx as i64 - 1;
+            }
```
youknowone, line 565: I don't think CPython uses loop for this. What happens if the dict has very many items? Would you check `_PyDict_Pop_KnownHash` in `dictobject.c`? I think this is the corresponding funciton in CPython.

coolreader18:
@youknowone here is where I meant to put this comment, not the other PR...

Interesting - it looks like CPython does this by just setting the entry's key+value to NULL and just letting it sit there, and only on "resize" does it get rid of the unused entry and recompact. That might be something to try, if you want, or else this looks fine to me.

fermian:
@coolreader18 I think you can't represent unused entries with the current DictEntry struct. Maybe we could turn DictInner's entry type into `Vec<Option<DictEntry<T>>>` and rewrite all the dictionary functions?

coolreader18:
@fermian yeah, I think that would be good to try, if you feel confident in it.

vm/src/dictdatatype.rs
DimitrisJim: Yeah, this definitely won't be a loop, pretty sure that leads to `O(N)` deletes.

youknowone: @eldpswp99 is working on this again

DimitrisJim:
Hey @fermian do you have time to see this through or should I pick it up?

youknowone:
#2902

PR #2552, opened 2021-04-14
Add test_ipaddress from CPython 3.8

fanninpm:


coolreader18:
Could you copy over `ipaddress.py` as well? It looks like some of the errors are just caused by the module being older than its tests.

coolreader18:
And then if you want, I think the remaining errors could be fixed by adding a `#[pyfunction] fn index(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyIntRef> { vm.to_index(&obj) }` to `vm/src/stdlib/operator.rs`

fanninpm:
> And then if you want, I think the remaining errors could be fixed by adding a `#[pyfunction] fn index(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyIntRef> { vm.to_index(&obj) }` to `vm/src/stdlib/operator.rs`

The requested changes are coming shortly.

coolreader18:
Ah, alright, I can look into that later.

coolreader18 (approved):

PR #2554, opened 2021-04-15
Add test_index from CPython 3.8.7

fanninpm:


coolreader18:
I think the issue with the other indexing was that `_operator` isn't in a `#[pymodule] mod`; could you try putting the `fn index` without the `#[pyfunction]` and just adding it to the `py_module!` macro at the bottom of the file?

coolreader18 (approved):
Thanks!

PR #2555, opened 2021-04-15
os: add getloadavg

deantvv:
Add `os.getloadavg()` by using [systemstat](https://github.com/unrelentingtech/systemstat)

Related to #1175

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1228,6 +1228,22 @@ mod _os {
         Ok(())
     }
 
+    #[cfg(unix)]
+    #[pyfunction]
+    fn getloadavg(vm: &VirtualMachine) -> PyResult<(f64, f64, f64)> {
+        let mut loadavg = [0f64; 3];
+
+        // Safety: loadavg is on stack and only write by `getloadavg` and are freed
+        // after this function ends.
+        unsafe {
+            if libc::getloadavg(&mut loadavg[0] as *mut f64, 3) != 3 {
+                return Err(vm.new_os_error("Load averages are unobtainable".to_string()));
+            }
+        }
+
+        Ok((loadavg[0], loadavg[1], loadavg[2]))
+    }
+
     #[pyattr]
     #[pyclass(module = "os", name = "terminal_size")]
     #[derive(PyStructSequence)]


coolreader18:
It looks like nix has [a sort of version](https://docs.rs/nix/0.20.0/nix/sys/sysinfo/struct.SysInfo.html#method.load_average) of this, or there's the direct [libc::getloadavg](https://docs.rs/libc/0.2.93/libc/fn.getloadavg.html) function which should be pretty simple to use ([here's how CPython does it](https://github.com/python/cpython/blob/master/Modules/posixmodule.c#L12530-L12540)). I'd prefer not to add another dependency just for this, since it's got a lot of functionality in it and we just need one function.

deantvv:
@coolreader18 Yes, I should definitely used `libc::getloadavg` instead of adding another library. 
This is now changed in 5707a96.

vm/src/stdlib/os.rs
```diff
@@ -1228,6 +1228,24 @@ mod _os {
         Ok(())
     }
 
+    #[cfg(unix)]
+    #[pyfunction]
+    fn getloadavg(vm: &VirtualMachine) -> PyResult<PyTupleRef> {
```
coolreader18, line 1233: You can actually return a `(f64, f64, f64)` tuple in rust, it will get automatically converted to a tuple.

deantvv: This is so cool. It is now in the code! Thanks for reviewing!

vm/src/stdlib/os.rs
```diff
@@ -1228,6 +1228,22 @@ mod _os {
...
+
+        // Safety: loadavg is on stack and only write by `getloadavg` and are freed
+        // after this function ends.
+        unsafe {
+            if libc::getloadavg(&mut loadavg[0] as *mut f64, 3) != 3 {
+                return Err(vm.new_os_error("Load averages are unobtainable".to_string()));
+            }
+        }
+
+        return Ok((loadavg[0], loadavg[1], loadavg[2]));
```
coolreader18, line 1244: Oh, and I think this is what's failing the linter, you don't need the `return` here - you can just leave off the last semicolon and it returns the value from the function.

deantvv: Yes, I will setup clippy in my develop environment to avoid this kind error from happening again.

coolreader18 (approved):
Thanks for contributing!

PR #2557, opened 2021-04-15
Add test_keyword from CPython 3.8

fanninpm:


Lib/test/test_keyword.py
```diff
@@ -29,6 +31,8 @@ def test_async_and_await_are_keywords(self):
         self.assertIn("async", keyword.kwlist)
         self.assertIn("await", keyword.kwlist)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_keywords_are_sorted(self):
         self.assertListEqual(sorted(keyword.kwlist), keyword.kwlist)
```
coolreader18: If you wanted, you could fix this by calling `sort_by_key` on the vec in `keywords.rs` before passing it to `new_list`

fanninpm: See e2aca3e

coolreader18 (approved):

PR #2559, opened 2021-04-15
Add some os-related modules and improve stdlib_inits

coolreader18:
Now IPython fully works:

```
$ rustpython -m pip install ipython
$ rustpython -m IPython
/home/coolreader18/.local/lib/rustpython3.9/site-packages/IPython/core/history.py:226: UserWarning: IPython History requires SQLite, your history will not be saved
  warn("IPython History requires SQLite, your history will not be saved")
Python 3.9.0alpha (heads/coolreader18/moduless-dirty:3f79daf1, Apr 15 2021, 09:45:25) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.22.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: x = 1

In [2]: x
Out[2]: 1

In [3]: import color
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
    [... skipping hidden 1 frame]

<ipython-input-3-c312e97f5a4c> in <module>
----> 1 import color

ModuleNotFoundError: No module named 'color'

In [4]: import colorsys

In [5]: colorsys.rgb_to_hsv?
Signature: colorsys.rgb_to_hsv(r, g, b)
Docstring: <no docstring>
File:      ~/projects/rustpython/vm/pylib-crate/Lib/colorsys.py
Type:      function         
```

### DIFF
--- a/vm/src/byteslike.rs
+++ b/vm/src/byteslike.rs
@@ -1,4 +1,5 @@
 use crate::builtins::memory::{try_buffer_from_object, BufferRef};
+use crate::builtins::PyStrRef;
 use crate::common::borrow::{BorrowedValue, BorrowedValueMut};
 use crate::pyobject::{BorrowValue, PyObjectRef, PyResult, TryFromObject};
 use crate::vm::VirtualMachine;
@@ -122,5 +123,28 @@ impl<'a> BorrowValue<'a> for PyRwBytesLike {
     type Borrowed = BorrowedValueMut<'a, [u8]>;
     fn borrow_value(&'a self) -> Self::Borrowed {
         self.0.as_contiguous_mut().unwrap()
+    }
+}
+
+pub enum BufOrStr {
+    Buf(PyBytesLike),
+    Str(PyStrRef),
+}
+
+impl TryFromObject for BufOrStr {
+    fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
+        obj.downcast()
+            .map(Self::Str)
+            .or_else(|obj| PyBytesLike::try_from_object(vm, obj).map(Self::Buf))
+    }
+}
+
+impl<'a> BorrowValue<'a> for BufOrStr {
+    type Borrowed = BorrowedValue<'a, [u8]>;
+    fn borrow_value(&'a self) -> Self::Borrowed {
+        match self {
+            Self::Buf(b) => b.borrow_value(),
+            Self::Str(s) => s.borrow_value().as_bytes().into(),
+        }
     }
 }
--- a/vm/src/byteslike.rs
+++ b/vm/src/byteslike.rs
@@ -126,6 +126,7 @@ impl<'a> BorrowValue<'a> for PyRwBytesLike {
     }
 }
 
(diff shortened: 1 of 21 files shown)


vm/src/stdlib/mod.rs
```diff
@@ -71,108 +77,116 @@ mod winreg;
...
-            "_ast".to_owned(),
-            Box::new(ast::make_module) as StdlibInitFunc,
-        );
-    }
-
-    #[cfg(feature = "rustpython-parser")]
-    {
-        modules.insert("keyword".to_owned(), Box::new(keyword::make_module));
-    }
+pub type StdlibMap = HashMap<Cow<'static, str>, StdlibInitFunc, ahash::RandomState>;
```
youknowone, line 80: what's the benefit to have this as `Cow`? Does this not fit for `&'static str` but still Cow is better than String?

vm/src/byteslike.rs
```diff
@@ -124,3 +125,26 @@ impl<'a> BorrowValue<'a> for PyRwBytesLike {
         self.0.as_contiguous_mut().unwrap()
     }
 }
+
+pub enum BufOrStr {
```
youknowone, line 129: is there a concept for this type in CPython?

coolreader18: Yeah, it's like the `s*` format code for PyArg_Parse: https://docs.python.org/3/c-api/arg.html#strings-and-buffers

I could add a doc comment for that.

vm/src/stdlib/mod.rs
coolreader18: Yeah, cause we still want people to be able to input dynamic builtin modules (e.g. `_sysconfigdata_{abi}_{os}_{triple}`) but there's really no need to allocate a String for every key since *most* of them will be static.

vm/src/byteslike.rs
youknowone: got it. it named `Py_buffer` but we use it for different struct.

vm/src/stdlib/mod.rs
youknowone: A bit doubtful for the benefit. The number of modules will not explode to big numbers. So it will be less than a few kb, but Cow might have performance drop for every access. (I have no idea for that. Is it free for accessing?)
If it does have benefit, `abi`, `os` and `triple` are const, so `_sysconfigdata_{abi}_{os}_{triple}` looks like to possibly be a `&'str` (const_concat: https://github.com/Vurich/const-concat). Though users must make every module name `&'static str` in this case, it will not be a big deal with leaking.

coolreader18: `Cow<str>` is slightly bigger in size - 32 bytes vs String is 24 bytes - but that's offset by not needing to allocate strings separately. And access is pretty much free, and it wouldn't be hot anyway, since after stdlib_inits is consulted once the module is put into sys.modules.

coolreader18: We could also use `beef::Cow` which implements a `Cow<str>` in only 24 bytes.

youknowone (approved):

PR #2560, opened 2021-04-15
Always have fds represent the C RT's fds, even on windows

coolreader18:


vm/src/crt_fd.rs
```diff
@@ -0,0 +1,159 @@
...
+        #[cfg(target_os = "wasi")]
+        let file = {
+            let fd = self.0;
+            if fd < 0 {
+                return Err(io::Error::from_raw_os_error(libc::EBADF));
+            }
+            // SAFETY: as of now, File is a wrapper around WasiFd, which is a wrapper around
+            // wasi::Fd (u32). This isn't likely to change, and if it does change to a different
+            // sized integer, mem::transmute will fail.
+            unsafe { mem::transmute::<u32, fs::File>(fd as u32) }
```
youknowone, line 97: if it is same as WasiFd, is it possible to make a WasiFd value with fd here instead of transmute?

youknowone (approved):

vm/src/crt_fd.rs
coolreader18: This is just copied from the old `os::rust_file` function; WasiFd is a struct internally in the standard library and [`wasi::Fd`](https://docs.rs/wasi/0.10.2+wasi-snapshot-preview1/wasi/type.Fd.html) is a type alias.

PR #2563, opened 2021-04-17
os: Implement copy_file_range

deantvv:
Use `libc::copy_file_range`

Related to #1175

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1,5 +1,3 @@
-use std::convert::TryFrom;
-use std::convert::TryInto;
 use std::ffi;
 use std::fs::File;
 use std::fs::OpenOptions;
@@ -1197,6 +1195,7 @@ mod _os {
         }
     }
 
+    #[cfg(target_os = "linux")]
     #[derive(FromArgs)]
     struct CopyFileRangeArgs {
         #[pyarg(positional)]
@@ -1214,12 +1213,13 @@ mod _os {
     #[cfg(target_os = "linux")]
     #[pyfunction]
     fn copy_file_range(args: CopyFileRangeArgs, vm: &VirtualMachine) -> PyResult<usize> {
+        use std::convert::{TryFrom, TryInto};
         let p_offset_src = args.offset_src.as_ref().map_or_else(std::ptr::null, |x| x);
         let p_offset_dst = args.offset_dst.as_ref().map_or_else(std::ptr::null, |x| x);
         let count: usize = args
             .count
             .try_into()
-            .map_err(|_| vm.new_value_error("count should > 0".to_string()))?;
+            .map_err(|_| vm.new_value_error("count should >= 0".to_string()))?;
 
         // The flags argument is provided to allow
         // for future extensions and currently must be to 0.
@@ -1238,7 +1238,7 @@ mod _os {
             )
         };
 
-        usize::try_from(ret).map_err(|_| vm.new_os_error("Fail to copy_file_range".to_string()))
+        usize::try_from(ret).map_err(|_| errno_err(vm))
     }
 
     #[pyfunction]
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/os.rs
```diff
@@ -1,3 +1,5 @@
+use std::convert::TryFrom;
+use std::convert::TryInto;
```
coolreader18, line 2: Could you move these `use`s into the body of the function? They're generating a lot of "unused import" warnings since they're only used on `cfg(linux)`

vm/src/stdlib/os.rs
```diff
@@ -1195,6 +1197,50 @@ mod _os {
         }
     }
 
+    #[derive(FromArgs)]
+    struct CopyFileRangeArgs {
```
coolreader18, line 1201: Could you put the `cfg()` on here too? So there aren't `unused` warnings on other platforms?

vm/src/stdlib/os.rs
```diff
@@ -1195,6 +1197,50 @@ mod _os {
...
+
+    #[cfg(target_os = "linux")]
+    #[pyfunction]
+    fn copy_file_range(args: CopyFileRangeArgs, vm: &VirtualMachine) -> PyResult<usize> {
+        let p_offset_src = args.offset_src.as_ref().map_or_else(std::ptr::null, |x| x);
+        let p_offset_dst = args.offset_dst.as_ref().map_or_else(std::ptr::null, |x| x);
+        let count: usize = args
+            .count
+            .try_into()
+            .map_err(|_| vm.new_value_error("count should > 0".to_string()))?;
```
coolreader18, line 1222: `>= 0` I think, since it just can't be negative

vm/src/stdlib/os.rs
```diff
@@ -1195,6 +1197,50 @@ mod _os {
...
+                args.src,
+                p_offset_src as *mut i64,
+                args.dst,
+                p_offset_dst as *mut i64,
+                count,
+                flags,
+            )
+        };
+
+        usize::try_from(ret).map_err(|_| vm.new_os_error("Fail to copy_file_range".to_string()))
```
coolreader18, line 1241: You can do `errno_err(vm)` to fetch the current errno and turn it into an exception - that'll give information about what actually failed, hopefully.

coolreader18 (commented):
Looks good! A lot of warnings are being generated on non-linux platforms, so that + some other nits and this should be good to go!

vm/src/stdlib/os.rs
deantvv: Yes, this seems much better.

vm/src/stdlib/os.rs
deantvv: Maybe `UtimeArgs` need to move behind `cfg()` also?
```rust
#[derive(FromArgs)]
    struct UtimeArgs {
    }
```

vm/src/stdlib/os.rs
deantvv: Yeah, but `args.count` is type `isize` and `count` is type `usize` so the `try_into()` will actually get `>=0` or it will be error.

vm/src/stdlib/os.rs
deantvv: I actually wonder how can I get the equivalent of `errno` on libc but fail to this code.
Thanks for the catch.

vm/src/stdlib/os.rs
coolreader18: Right, I was saying the error message says `>` but it's actually `>=`

vm/src/stdlib/os.rs
coolreader18: Oh, sure, go for it if you want

vm/src/stdlib/os.rs
deantvv: Oh this is great catch! Will fix it.

coolreader18 (approved):

coolreader18:
Thanks for contributing!

PR #2567, opened 2021-04-18
Add test_marshal from CPython 3.8

fanninpm:


coolreader18:
I'm not sure about this - we basically only implement the marshal module in order to support dumping/loading bytecode, and I don't think CPython supports it for anything other than that either. The [docs](https://docs.python.org/3/library/marshal.html) say as much, I think. The tests seem to be internal consistency tests, and given we have a completely different codeobject format, I don't think these tests are really necessary.

PR #2569, opened 2021-04-18
Add test_mimetypes from CPython 3.8

fanninpm:


coolreader18:
Try updating `mimetypes.py`? Brings the error count down to 3 for me.

coolreader18 (approved):

PR #2570, opened 2021-04-18
Rename test_named_expression to name in CPython

fanninpm:
I don't know if you want different error messages than CPython. (See the 'wrong error message' annotations in the test file.)


coolreader18:
Hmm, yeah, that would probably be good/easy to fix. At least if it does get fixed now we'll notice by the unexepected success in CI.

coolreader18 (approved):

PR #2572, opened 2021-04-19
Fix remaining code coverage collection issues

coolreader18:


vm/src/stdlib/os.rs
```diff
@@ -1274,7 +1274,7 @@ mod _os {
         Ok(())
     }
 
-    #[cfg(unix)]
+    #[cfg(all(unix, not(any(target_os = "redox", target_os = "android"))))]
```
fanninpm, line 1277:
```suggestion
    #[cfg(all(unix, not(any(target_os in {"redox", "android"}))))]
```

fanninpm:
Are there any plans to add coverage testing to PRs?

coolreader18:
I think we used to have that a while ago, but since it's a 4 hour ci run now no, probably not

coolreader18:
Yooo, nice!

coolreader18:
Oooh wait, I've found a much better way to do code coverage

coolreader18:
WOW, that's so much better than the other method of collection profiling data. It actually lets the compiler optimize the code, so we can run with `--release` no problem. [Article about source-based code coverage](https://blog.rust-lang.org/inside-rust/2020/11/12/source-based-code-coverage.html)

youknowone (approved):

PR #2575, opened 2021-04-20
Add extra check in test_py_compile for Big Sur and/or Apple Silicon

fanninpm:
I don't know if this has to do with the operating system or the architecture. Does anyone have an Intel Mac running Big Sur that can test this patch's effectiveness?

### DIFF
--- a/Lib/test/test_py_compile.py
+++ b/Lib/test/test_py_compile.py
@@ -115,7 +115,7 @@ class PyCompileTestsBase:
 
     # TODO: RUSTPYTHON
     import platform
-    if sys.platform == "darwin" and platform.release() != "20.3.0":
+    if sys.platform == "darwin" and int(platform.release().split(".")[0]) < 20:
         test_relative_path = unittest.expectedFailure(test_relative_path)
 
     @unittest.skipIf(hasattr(os, 'geteuid') and os.geteuid() == 0,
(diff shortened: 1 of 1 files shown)


youknowone (commented):
I tested it on intel big sur. `test_relative_path` always success. So it looks like mac version check is better than platform check .

fanninpm:
> I tested it on intel big sur. `test_relative_path` always success. So it looks like mac version check is better than platform check .

Unfortunately, `platform.mac_ver()` errors because of missing XML parsers. Is the kernel version number the same for both architectures?

Lib/test/test_py_compile.py
```diff
@@ -114,7 +114,8 @@ def test_relative_path(self):
         self.assertFalse(os.path.exists(self.cache_path))
 
     # TODO: RUSTPYTHON
-    if sys.platform == "darwin":
+    import platform
+    if sys.platform == "darwin" and platform.release() != "20.3.0":
```
youknowone, line 118:
```python
tuple(int(p) for p in platform.release().split('.')) >= "20.3.0"
```

youknowone:
Yes. I also get `20.3.0`. No idea how much reliable it is.

Lib/test/test_py_compile.py
coolreader18: Wouldn't it be `>= (20, 3, 0)` ?

fanninpm: @youknowone Does 84f13e1 work on your architecture?

youknowone (approved):

Lib/test/test_py_compile.py
youknowone: Yes, it does work :+1:

PR #2580, opened 2021-04-21
Change data chunk passing in Rust zlib compression

ishigoya:
- Pass data chunks on a slice-by-slice basis to compression API
- decompress method: Amend error message

This is my first PR in Rust :grin:
Please let me know if I've made any mistakes / if there's anything I could improve on!

### DIFF
--- a/vm/src/stdlib/zlib.rs
+++ b/vm/src/stdlib/zlib.rs
@@ -246,7 +246,10 @@ mod decl {
                 if stream_end {
                     Ok(buf)
                 } else {
-                    Err(new_zlib_error("incomplete or truncated stream", vm))
+                    Err(new_zlib_error(
+                        "Error -5 while decompressing data: incomplete or truncated stream",
+                        vm,
+                    ))
                 }
             })
         })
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/zlib.rs
```diff
@@ -246,7 +246,10 @@ mod decl {
                 if stream_end {
                     Ok(buf)
                 } else {
-                    Err(new_zlib_error("incomplete or truncated stream", vm))
+                    Err(new_zlib_error(
+                        "Error -5 while decompressing data: incomplete or truncated stream",
+                        vm,
+                    ))
```
fanninpm, line 252: @coolreader18 @youknowone In the future, could we look at dynamically generating these error messages in `new_zlib_error`, like [how `zlib_error` is implemented in the CPython source](https://github.com/python/cpython/blob/3.8/Modules/zlibmodule.c)?

coolreader18: Hmm, that'd be nice, but flate2 doesn't expose any information about the specific error that occurred, so we'd need to change it there first.

vm/src/stdlib/zlib.rs
```diff
@@ -477,34 +480,36 @@ mod decl {
     const CHUNKSIZE: usize = u32::MAX as usize;
 
     impl CompressInner {
-        fn save_unconsumed_input(&mut self, data: &[u8], orig_in: u64) {
-            let leftover = &data[(self.compress.total_in() - orig_in) as usize..];
+        fn save_unconsumed_input(&mut self, data: &[u8], cur_in: usize) {
+            let leftover = &data[cur_in..];
             self.unconsumed.extend_from_slice(leftover);
         }
```
coolreader18, line 486: I don't think we need this method anymore, we can probably just inline it since it doesn't do anything all that special.

vm/src/stdlib/zlib.rs
```diff
@@ -477,34 +480,36 @@ mod decl {
...
-                loop {
+                while cur_in < chunk.len() {
                     buf.reserve(DEF_BUF_SIZE);
                     let status = self
                         .compress
-                        .compress_vec(chunk, &mut buf, FlushCompress::None)
+                        .compress_vec(&chunk[cur_in..], &mut buf, FlushCompress::None)
                         .map_err(|_| {
-                            self.save_unconsumed_input(data, orig_in);
+                            self.save_unconsumed_input(data, cur_in);
```
coolreader18, line 501: e.g. here this can just be one line as well, `self.unconsumed.extend_from_slice(&data[cur_in..]);`

coolreader18 (approved):
Very nice! I have one comment, but otherwise this is good to merge!

vm/src/stdlib/zlib.rs
fanninpm: It seems like [flate2's `Status` enum](https://docs.rs/flate2/1.0.20/flate2/enum.Status.html) could be expanded to account for more than just `BufError`, but I could be wrong.

coolreader18: I think it should be possible to fetch zlib's `msg` in flate2, it should give better error messages there as well. I'll look into contributing that.

coolreader18:
Thank you for contributing!

ishigoya:
Thanks!

PR #2584, opened 2021-04-22
Unskip more macOS tests in CI

fanninpm:
This PR will incrementally determine exactly which tests must be skipped on the macOS CI run. Submitting as draft until everything is narrowed down.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -111,7 +111,6 @@ jobs:
         name: run cpython tests (macOS lightweight)
         run:
           target/release/rustpython -m test -v -x
-            test_bytearray
             test_long
             test_unicode
             test_array
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -111,7 +111,6 @@ jobs:
         name: run cpython tests (macOS lightweight)
         run:
           target/release/rustpython -m test -v -x
-            test_array
             test_asyncgen
             test_list
             test_complex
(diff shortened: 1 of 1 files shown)


coolreader18:
Shoot... I probably should've mentioned this earlier but I wasn't entirely sure then, but I think the reason those were skipped is just because the macos runner is a lot slower -- ubuntu/windows take about 20 minutes, while the macos run for the latest commit took 40 minutes. It wasn't actually because they were failing on macos.

fanninpm:
> the macos runner is a lot slower -- ubuntu/windows take about 20 minutes, while the macos run for the latest commit took 40 minutes

If you want, you could defer this PR. To that end, I have noticed that the _build_ times on the macOS runner are about 3-4x slower than those on the Ubuntu or Windows runners. However, I have also noticed that the addition of the extra texts added little to the overall run time.

coolreader18:
Huh, I guess that's true, other PRs have macos taking like 35m. This is \~probably\~ fine then

coolreader18:
Ohhh right, I turned off caching for macos cause it was erroring. Do you want to see if it works now? I think you can just the `if: runner.os != 'macOS'` from the actions/cache step.

fanninpm:
I just found [this action](https://github.com/Swatinem/rust-cache), which claims to work around currently unresolved issues that corrupt the cache on macOS builds. I don't think it would be too hard to implement.

fanninpm:
@coolreader18 should I put that GitHub action in, or should I leave this as it is?

coolreader18:
@fanninpm I think that'd be good

coolreader18 (approved):
LGTM, thanks!

PR #2592, opened 2021-04-24
Type checking for __index__-defined integers

ishigoya:
- add a special_retrieve fudge function to the PyValue trait
- define the special_retrieve function for PyInt
- remove blocks on successful tests to reflect effect of changes

### DIFF
--- a/vm/src/builtins/int.rs
+++ b/vm/src/builtins/int.rs
@@ -79,8 +79,8 @@ impl PyValue for PyInt {
         vm.ctx.new_int(self.value)
     }
 
-    fn special_retrieve(vm: &VirtualMachine, obj: PyObjectRef) -> Option<PyResult<PyRef<Self>>> {
-        vm.to_index_opt(obj)
+    fn special_retrieve(vm: &VirtualMachine, obj: &PyObjectRef) -> Option<PyResult<PyRef<Self>>> {
+        Some(vm.to_index(obj))
     }
 }
 
--- a/vm/src/builtins/int.rs
+++ b/vm/src/builtins/int.rs
@@ -78,6 +78,18 @@ impl PyValue for PyInt {
     fn into_object(self, vm: &VirtualMachine) -> PyObjectRef {
         vm.ctx.new_int(self.value)
     }
+
+    fn special_retrieve(vm: &VirtualMachine, obj: PyObjectRef) -> Option<PyResult<PyRef<Self>>> {
+        vm.get_method(obj, "__index__").map(|index| {
+            // TODO: returning strict subclasses of int in __index__ is deprecated
+            vm.invoke(&index?, ())?.downcast().map_err(|bad| {
+                vm.new_type_error(format!(
+                    "__index__ returned non-int (type {})",
+                    bad.class().name
+                ))
+            })
+        })
+    }
 }
 
 macro_rules! impl_into_pyobject_int {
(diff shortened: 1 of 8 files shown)


vm/src/pyobject.rs
```diff
@@ -926,6 +927,12 @@ pub trait PyValue: fmt::Debug + PyThreadingConstraint + Sized + 'static {
         self.into_ref(vm).into_object()
     }
 
+    fn special_retrieve(vm: &VirtualMachine, obj: PyObjectRef) -> Option<PyResult<PyRef<Self>>> {
```
coolreader18, line 930: I think this could be `&PyObjectRef`, since `try_from_object` always calls `special_retrieve` with `obj.clone()`. Also I think you can put `#[inline(always)]`

vm/src/builtins/int.rs
```diff
@@ -78,6 +78,18 @@ impl PyValue for PyInt {
...
+    fn special_retrieve(vm: &VirtualMachine, obj: PyObjectRef) -> Option<PyResult<PyRef<Self>>> {
+        vm.get_method(obj, "__index__").map(|index| {
+            // TODO: returning strict subclasses of int in __index__ is deprecated
+            vm.invoke(&index?, ())?.downcast().map_err(|bad| {
+                vm.new_type_error(format!(
+                    "__index__ returned non-int (type {})",
+                    bad.class().name
+                ))
+            })
+        })
```
coolreader18, line 91: You can just call `vm.to_index()` here, we already have this logic implemented.

coolreader18 (approved):
Very nice solution!! I've got a couple small comments, but otherwise this is really clever!

vm/src/builtins/int.rs
ishigoya: I added it as `to_index_opt()` so I could return `None` in the default implementation, is that OK?

coolreader18: Well I think it makes sense to always return `Some`, since `to_index` will have a better error message.

ishigoya: Ah, I was using the `pyref_type_error` for all circumstances.

So, have the default implementation of `special_retrieve` return the current `pyref_type_error`, and pass `to_index` unchanged from the int `special_retrieve`?

ishigoya: Like this?

vm/src/pyobject.rs
```diff
@@ -926,6 +926,13 @@ pub trait PyValue: fmt::Debug + PyThreadingConstraint + Sized + 'static {
         self.into_ref(vm).into_object()
     }
 
+    #[inline(always)]
+    fn special_retrieve(vm: &VirtualMachine, obj: &PyObjectRef) -> Option<PyResult<PyRef<Self>>> {
+        let _ = vm;
+        let _ = obj;
+        None
+    }
```
youknowone, line 934:
```suggestion
    fn special_retrieve(_vm: &VirtualMachine, _obj: &PyObjectRef) -> Option<PyResult<PyRef<Self>>> {
        None
    }
```

youknowone (changes requested):

youknowone (approved):

youknowone:
Looks good to me. A question: why is it named `special_retrieve`?

ishigoya:
I tried to keep the name quite vague, as I couldn't predict how it might be used for other types. I'd welcome any other suggestions!

youknowone:
@coolreader18 do you have a suggestion?

coolreader18 (approved):
Oh, nope, LGTM

PR #2594, opened 2021-04-25
os: Fix `open` keyword arguments

deantvv:
Following code used to fail due to the way used to define
os.open in rust. To fix this, we need to use `FromArgs`
to define `struct OpenArgs {...}`.
```
f = os.open(path=__file__, flags=os.O_RDONLY, mode=0o777, dir_fd=None)
```

Noted that the original `open` is now rename to `rust_open`
and is also exported. The only usage of `rust_open` is only in `io.rs`.

Related to #1175

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -357,11 +357,29 @@ mod _os {
     }
 
     #[cfg(any(unix, windows, target_os = "wasi"))]
+    #[derive(FromArgs)]
+    struct OpenArgs {
+        #[pyarg(any)]
+        path: PyPathLike,
+        #[pyarg(any)]
+        flags: OpenFlags,
+        #[pyarg(any, default)]
+        mode: Option<PyIntRef>,
+        #[pyarg(flatten)]
+        dir_fd: DirFd,
+    }
+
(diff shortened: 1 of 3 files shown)


youknowone (approved):

vm/src/stdlib/os.rs
```diff
@@ -356,12 +356,30 @@ mod _os {
...
+
     #[cfg(any(unix, windows, target_os = "wasi"))]
     #[pyfunction]
-    pub(crate) fn open(
+    fn open(args: OpenArgs, vm: &VirtualMachine) -> PyResult<i64> {
+        rust_open(args.path, args.flags, args.mode, args.dir_fd, vm)
+    }
+
+    #[cfg(any(unix, windows, target_os = "wasi"))]
+    pub(crate) fn rust_open(
```
coolreader18, line 379: Maybe you could rename this to `os_open`, but export it below as `pub(crate) use _os::os_open as open;`? This sorta follows the convention of `io_open` in `io.rs`.

deantvv: I am actually wondering about the name `rust_open`, and I think your choice is better. It is now changed.

coolreader18:
Thanks!

PR #2596, opened 2021-04-25
Add test_peepholer from CPython 3.8

fanninpm:


coolreader18:
I don't think this is necessary, since it tests CPython's internal bytecode format, and we don't even use the same names - CPython uses `SCREAMING_SNAKE_CASE` for the instruction names for its `dis` implementation while we use `CamelCase`, so it'll never match any of the instructions.

PR #2606, opened 2021-04-29
os: symlink use `FromArgs` for arguments

deantvv:
Pass `FileTests.test_symlink_keywords` in `test_os.py`

Related to #1175

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1775,19 +1775,25 @@ mod posix {
 
     pub(super) const SYMLINK_DIR_FD: bool = cfg!(not(target_os = "redox"));
 
-    #[pyfunction]
-    pub(super) fn symlink(
+    #[derive(FromArgs)]
+    pub(super) struct SimlinkArgs {
+        #[pyarg(any)]
         src: PyPathLike,
+        #[pyarg(any)]
         dst: PyPathLike,
+        #[pyarg(flatten)]
         _target_is_directory: TargetIsDirectory,
+        #[pyarg(flatten)]
         dir_fd: DirFd<{ SYMLINK_DIR_FD as usize }>,
-        vm: &VirtualMachine,
-    ) -> PyResult<()> {
-        let src = src.into_cstring(vm)?;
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/os.rs
```diff
@@ -3122,14 +3135,20 @@ mod minor {
...
-        _src: PyPathLike,
-        _dst: PyPathLike,
+    #[derive(FromArgs)]
+    pub(super) struct SimlinkArgs {
+        #[pyarg(any)]
+        src: PyPathLike,
+        #[pyarg(any)]
+        dst: PyPathLike,
+        #[pyarg(flatten)]
         _target_is_directory: TargetIsDirectory,
```
youknowone, line 3145: `_` prefix in struct will affact keyword argument name.

deantvv: Since `_target_is_directory` is `#[pyarg(flatten)]`, `_` prefix will not affect keyword argument name.

```rust
#[derive(FromArgs, Default)]
pub struct TargetIsDirectory {
    #[pyarg(any, default = "false")]
    target_is_directory: bool,
}
```

vm/src/stdlib/os.rs
```diff
@@ -2777,25 +2783,32 @@ mod nt {
...
         src: PyPathLike,
+        #[pyarg(any)]
         dst: PyPathLike,
+        #[pyarg(flatten)]
         target_is_directory: TargetIsDirectory,
-        _dir_fd: DirFd<0>,
-        vm: &VirtualMachine,
-    ) -> PyResult<()> {
+        #[pyarg(flatten)]
+        dir_fd: DirFd<{ SYMLINK_DIR_FD as usize }>,
```
youknowone, line 2795: Thanks! That's true. Then, `_dir_fd` here looks also safe. Here is an unused field warning.

youknowone (approved):

vm/src/stdlib/os.rs
deantvv: @youknowone Done!

PR #2614, opened 2021-05-01
Add test_richcmp from CPython 3.8

fanninpm:


Lib/test/test_richcmp.py
```diff
@@ -253,6 +253,8 @@ def test_recursion(self):
         self.assertTrue(a != b)
         self.assertTrue(a < b)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
coolreader18, line 260: I think this should be pretty easy to fix, if you want - I think it's on line 631 in vm.rs, we currently differ from CPython's message format, which is `'{}' not supported between instances of '{}' and '{}'`

fanninpm: Resolved in 4c8e402.

fanninpm:
That WASM build failure is likely due to rustwasm/wasm-pack#997.

coolreader18 (approved):

PR #2617, opened 2021-05-02
os: Implement `inode()` for class `DirEntry`

deantvv:
Use `stat_inner` to get stat and get inode number.
Currently on windows platform inode will only contains
default value (0).


vm/src/stdlib/os.rs
```diff
@@ -793,6 +793,21 @@ mod _os {
...
+                PathOrFd::Path(PyPathLike {
+                    path: self.entry.path(),
+                    mode: OutputMode::String,
+                }),
+                DirFd::default(),
+                FollowSymlinks(false),
+            )
+            .map_err(|e| e.into_pyexception(vm))?
+            .ok_or_else(|| vm.new_value_error("embedded null character".to_owned()))
+            .map(|stat| stat.st_ino)
```
coolreader18, line 808: I think on unix, you can use [`std::os::unix::fs::DirEntryExt::ino`](https://doc.rust-lang.org/stable/std/os/unix/fs/trait.DirEntryExt.html#tymethod.ino), and on windows, you can cache the result of `stat_inner` with an `AtomicCell<Option<u64>>` on DirEntry

deantvv: The new version uses [`std::os::unix::fs::DirEntryExt::ino`](https://doc.rust-lang.org/stable/std/os/unix/fs/trait.DirEntryExt.html#tymethod.ino) for unix and use `stat_inner` for other platform. Thanks for the nice suggestion.

vm/src/stdlib/os.rs
```diff
@@ -793,6 +797,34 @@ mod _os {
...
+                        PathOrFd::Path(PyPathLike {
+                            path: self.entry.path(),
+                            mode: OutputMode::String,
+                        }),
+                        DirFd::default(),
+                        FollowSymlinks(false),
+                    )
+                    .map_err(|e| e.into_pyexception(vm))?
+                    .ok_or_else(|| vm.new_value_error("embedded null character".to_owned()))?;
+                    self.ino.swap(Some(stat.st_ino));
```
coolreader18, line 816: Could you do a compare_exchange here? That way it won't get overwritten if a new thread calculated it in the meantime.

coolreader18 (approved):

coolreader18 (approved):
Awesome, thank you for contributing!

coolreader18:
Huh, yeah, weird. Thanks for contributing!

PR #2619, opened 2021-05-02
Unskip more Windows tests in CI

fanninpm:
This PR will incrementally determine exactly which tests must be skipped on the Windows CI run. Submitting as draft until everything is narrowed down.

NOTE: This differs from #2584 in a couple of ways:

- I'm essentially flying blind, as I currently don't own a machine with Windows that I can tinker on.
- Apparently, the Windows CI behaves differently from local machines. I expect quite a few force-pushes.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -100,10 +100,30 @@ jobs:
         name: run cpython tests (windows partial - fixme)
         run:
           target/release/rustpython -m test -v -x
-            test_argparse test_json test_bytes test_long test_pwd test_bool test_cgi test_complex
-            test_exception_hierarchy test_glob test_iter test_list test_os test_pathlib
-            test_py_compile test_set test_shutil test_sys test_unicode test_unittest test_venv
-            test_zipimport test_importlib test_io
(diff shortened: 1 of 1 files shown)


coolreader18:
I don't think this is necessary - obviously if there are more io/os-related modules those should be unskipped if possible, but stuff like argparse, json, etc are just cpu-based, and were probably skipped due to the windows runner being much slower than the linux one. Same for macos, to some extent - even with caching, running the regrtests takes like 10 more minutes than on linux, it looks like. ([macos 23min](https://github.com/RustPython/RustPython/runs/2485303101) [linux 10min](https://github.com/RustPython/RustPython/runs/2485303108))

youknowone:
I think the CI file needs comments about this.

I removed these tests from macos and windows because they are not platform-dependent but affect CI performance.
```
test_argparse test_json test_bytes test_bytearray test_long test_unicode test_array
test_asyncgen test_list test_complex test_json test_set test_dis test_calendar
```

The other windows tests must be once unskipped in future.

fanninpm:
This effort will now be abandoned.

PR #2621, opened 2021-05-03
Add a bunch of stuff to the socket module

coolreader18:


fanninpm:
If need be, you can mark a test as an expected failure on a specific OS by placing this snippet below the test's definition (the first line is only needed if the test module does not already import `sys`):

```python
import sys # TODO: RUSTPYTHON
if sys.platform == "win32":
    test_function = unittest.expectedFailure(test_function)
```

coolreader18:
Thanks, I think I'll try to figure out what the issue is on windows though (just need to find some time to boot into it), since those tests were working before and I'm not sure how far-reaching the underlying issue is.

PR #2626, opened 2021-05-07
Add test_shlex from CPython 3.8

fanninpm:


fanninpm:
Unrelated Clippy failures are fixed in #2627.

coolreader18 (approved):

PR #2635, opened 2021-05-12
remove empty field `_priv`

youknowone:


coolreader18:
I'm not sure about this; I think I was probably the one to put the empty field for most of those, and it's so that it's not constructable outside of the module. Otherwise, it being a unit type would be part of the public API. Specifically for something like `ConstantOptimizer`, there are probably optimizations that could be made that would require adding a field to the struct, which would then be a breaking change.

PR #2636, opened 2021-05-12
module _scproxy

youknowone:
on the way to go to `--install-pip` for macos

### DIFF
--- a/vm/src/stdlib/scproxy.rs
+++ b/vm/src/stdlib/scproxy.rs
@@ -1,0 +1,147 @@
+pub(crate) use _scproxy::make_module;
+
+#[pymodule]
+mod _scproxy {
+    // straight-forward port of Modules/_scproxy.c
+
+    use crate::builtins::{PyDictRef, PyStr};
+    use crate::pyobject::{IntoPyObject, ItemProtocol, PyResult};
+    use crate::VirtualMachine;
+    use system_configuration::core_foundation::{
+        array::CFArray,
+        base::{CFType, FromVoid, TCFType},
+        dictionary::CFDictionary,
+        number::CFNumber,
+        string::{CFString, CFStringRef},
+    };
+    use system_configuration::sys::{
+        dynamic_store_copy_specific::SCDynamicStoreCopyProxies, schema_definitions::*,
+    };
+
+    fn proxy_dict() -> Option<CFDictionary<CFString, CFType>> {
+        // Py_BEGIN_ALLOW_THREADS
+        let proxy_dict = unsafe { SCDynamicStoreCopyProxies(std::ptr::null()) };
+        // Py_END_ALLOW_THREADS
+        if proxy_dict.is_null() {
+            None
+        } else {
+            Some(unsafe { CFDictionary::wrap_under_create_rule(proxy_dict) })
+        }
+    }
+
+    #[pyfunction]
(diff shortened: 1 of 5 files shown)


vm/src/stdlib/scproxy.rs
```diff
@@ -0,0 +1,140 @@
...
+        if proxy_dict_raw.is_null() {
+            return Ok(None);
+        }
+        let proxy_dict =
+            unsafe { CFDictionary::<_, CFNumber>::wrap_under_get_rule(proxy_dict_raw) };
+
+        let result = vm.ctx.new_dict();
+        let v = 0
+            != if let Some(a_num) =
+                unsafe { proxy_dict.find(kSCPropNetProxiesExcludeSimpleHostnames) }
```
coolreader18, line 36: I don't think this `unsafe` block is necessary? It looks like `CFDictionary::find` is safe to me from the docs: https://docs.rs/core-foundation/0.7.0/core_foundation/dictionary/struct.CFDictionary.html#method.find

vm/src/stdlib/scproxy.rs
```diff
@@ -0,0 +1,140 @@
...
+            {
+                let num: &CFNumber = a_num.deref();
+                num.to_i32().unwrap_or(0)
+            } else {
+                0
+            };
+        result.set_item("exclude_simple", vm.ctx.new_bool(v), vm)?;
+
+        let proxy_dict =
+            unsafe { CFDictionary::<_, CFArray<CFString>>::wrap_under_create_rule(proxy_dict_raw) };
```
coolreader18, line 46: I think you can just use the `proxy_dict` from before; the high-level [get_proxies](https://docs.rs/system-configuration/0.4.0/system_configuration/dynamic_store/struct.SCDynamicStore.html#method.get_proxies) function returns it as a `CFDictionary<CFString, CFType>`, and I think `find` still works regardless of the V type parameter for CFDictionary

vm/src/stdlib/scproxy.rs
```diff
@@ -0,0 +1,140 @@
...
+            result.set_item("exceptions", vm.ctx.new_tuple(v), vm)?;
+        }
+
+        Ok(Some(result))
+    }
+
+    #[pyfunction]
+    fn _get_proxies(vm: &VirtualMachine) -> PyResult<Option<PyDictRef>> {
+        // Py_BEGIN_ALLOW_THREADS
+        let proxy_dict_raw = unsafe { SCDynamicStoreCopyProxies(std::ptr::null()) };
```
coolreader18, line 68: Maybe this could be deduplicated/extracted out to a separate `fn() -> Option<CFDictionary<CFString, CFType>>` since we know it's always safe?

coolreader18 (approved):
Nice! I have a few nits about deduplicating code, but otherwise this looks good!

fanninpm:
In `.github/workflows/ci.yaml`, I had skipped `test_http_cookiejar` and `test_robotparser` because this module was missing. Now, when they are run, do they no longer emit a `ModuleNotFoundError` for `_scproxy`?

youknowone (commented):
@fanninpm unskipped the tests, thanks!

vm/src/stdlib/scproxy.rs
coolreader18: Ah, that makes sense.

PR #2649, opened 2021-05-15
os: Fix `DirEntry` is_file and is_dir function

deantvv:
Change `perform_on_metadata` to fit python document and tests.

From [`os.DirEntry.is_dir()`](https://docs.python.org/3/library/os.html?highlight=direntry#os.DirEntry.is_dir)
> Return True if this entry is a directory or a symbolic link pointing to a directory;
> return False if the entry is or points to any other kind of file, or if it doesn’t exist anymore.
> ...
> This method can raise OSError, such as PermissionError, but FileNotFoundError is caught and not raised.

Related to #1175


coolreader18 (approved):
Nice, thank you for contributing!

PR #2651, opened 2021-05-15
Correct documentation link

### LINKED ISSUE #2598: Link to online documentation is broken
On the README.md: 

>You can read the [online documentation](https://docs.rs/rustpython-vm) for the latest release.

Gives a 404.

rahul-kumi:
resolves #2598

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,8 +154,8 @@ make cool projects:
 Currently along with other areas of the project, documentation is still in an
 early phase.
 
-You can read the [online documentation](https://rustpython.github.io/docs/) for the
-latest release.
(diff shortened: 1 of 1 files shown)


coolreader18:
I'm not sure about this; the docs.rs links will be back up soon but also the docs on the website aren't really the same, those are more of a user guide rather than an API reference. Could you maybe add it as a separate link? "for the latest release, or [the user guide[(...)"

rahul-kumi:
@coolreader18, I have made the suggested changes in this commit 1dfc783

coolreader18:
@rahul-kumi Thanks! Although it looks like you put double parentheses around the link, so it won't render as a link properly.

PR #2652, opened 2021-05-15
Add test_sundry from CPython 3.8

fanninpm:


coolreader18:
Ah, I thought there might have been something in test_subprocess that broke with #2634. Looks like that just should be marked as a failure, since #2646 doesn't fix it either (need to implement the mbcs codec)

fanninpm:
I thought test_subprocess broke when #2645 was merged. I'm addressing that (and the test_os failure) in a separate PR (forthcoming).

coolreader18:
Huh, maybe. I guess I see the benefits of bors' PR batching now 🙃

fanninpm:
> Huh, maybe. I guess I see the benefits of bors' PR batching now 🙃

I think PR batching is especially helpful when there's a high volume of PRs. I also think the fact that bors stages each PR before it gets merged into master would be very beneficial for this project.

coolreader18 (approved):

PR #2654, opened 2021-05-15
Fix CI failures with test_subprocess and/or test_os

fanninpm:
As observed in #2652, various tests failed around the time when #2645 and #2647 were both merged into the master branch. Submitting this as draft because I can't reproduce the `test_os` failure on my machine.

PR #2655, opened 2021-05-16
Cache stat result in DirEntry to fix spurious CI failures

coolreader18:
Follow-up to #2654, cc @fanninpm


coolreader18:
Ok this should work now

PR #2661, opened 2021-05-18
Add tests file for the bisect module.

DimitrisJim:
Kept the same structure (as also described in [PEP 399](https://www.python.org/dev/peps/pep-0399/)) in order to easily be able to plug in a future rust implementation of bisect.

No changes other than removing statements dealing with the C accelerator module (and importing module directly for now). All tests locally pass.

### DIFF
--- a/Lib/test/test_bisect.py
+++ b/Lib/test/test_bisect.py
@@ -1,0 +1,328 @@
+import sys
+import unittest
+from test import support
+from collections import UserList
+
+py_bisect = support.import_fresh_module('bisect', blocked=['_bisect'])
+c_bisect = support.import_fresh_module('bisect', fresh=['_bisect'])
+
+class Range(object):
+    """A trivial range()-like object that has an insert() method."""
+    def __init__(self, start, stop):
+        self.start = start
+        self.stop = stop
+        self.last_insert = None
+
+    def __len__(self):
+        return self.stop - self.start
+
+    def __getitem__(self, idx):
+        n = self.stop - self.start
+        if idx < 0:
+            idx += n
+        if idx >= n:
+            raise IndexError(idx)
+        return self.start + idx
+
+    def insert(self, idx, item):
+        self.last_insert = idx, item
+
+
+class TestBisect:
+    def setUp(self):
+        self.precomputedCases = [
+            (self.module.bisect_right, [], 1, 0),
+            (self.module.bisect_right, [1], 0, 0),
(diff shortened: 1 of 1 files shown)


fanninpm:
@coolreader18 should this commit be split (as per [your comment on a separate issue from February 2021](https://github.com/RustPython/RustPython/pull/2455#issuecomment-778901045))?

coolreader18:
Oh, yeah, that would be good. @DimitrisJim could you have one commit that's *just* copying the file directly from CPython (Like "Add test.test_bisect from CPython x.y" for whatever CPython version you used), and another with the changes you make?

DimitrisJim:
@coolreader18 done. Wasn't aware of the etiquette on this (licensing related?). Hopefully everything is fine now.

coolreader18 (approved):
Yup, looks good! I'm not exactly sure about licensing, but it also helps with seeing the history of the changes we've made to test files; what needed to be skipped and so forth.

Lib/test/test_bisect.py
```diff
@@ -290,10 +282,7 @@ def test_arg_parsing(self):
             self.assertRaises(TypeError, f, 10)
 
 class TestErrorHandlingPython(TestErrorHandling, unittest.TestCase):
-    module = py_bisect
-
-class TestErrorHandlingC(TestErrorHandling, unittest.TestCase):
-    module = c_bisect
+    module = bisect
```
coolreader18, line 285: Another option for this is to leave the `py_bisect = `/`c_bisect =`/`module = py_bisect` as is, and just add like `@unittest.skip("TODO: RUSTPYTHON, _bisect module")` to the C tests.

DimitrisJim: Ah, yes. The reason for that is `test.support.import_fresh_module` isn't currently implemented. I could just change it to:

```python
...
import bisect
...

py_bisect = bisect
c_bisect = None
```

and then annotate the classes with `skip`. Adding skip is definitely a better option as it acts like a marker.

DimitrisJim:
@coolreader18 done! Thanks for the feedback.

DimitrisJim:
I noticed I was being daft, `support.import_fresh_module` exists just fine. (I guess it errored previously because `_bisect` doesn't exist and I must've misread the error.)

@coolreader18 do you think we can push this? (I'm waiting on it to push a `_bisect` implementation I'm working on).

coolreader18:
Oh, yeah, sorry about that!

PR #2662, opened 2021-05-20
Flatten pyobject module into the root of the crate, don't require BorrowValue everywhere

coolreader18:
I was gonna make an RFC about this, but I wasn't able to write that much about it so I figured I'd just make a PR. For the first change, I realized that `pyobject` is really just a "generic" module that has anything you might need for doing anything with rustpython, so I think it would make sense if that stuff (`PyObjectRef`, `PyResult`, etc) was easily accessible right from `rustpython_vm::`, and it's not like there's a lot of stuff there already.

For the second one, I think I basically have gotten tired of having to import `BorrowValue` everywhere, and it also doesn't communicate the cost of "borrowing" something - `foo.borrow_value()` could lock a mutex, read a rwlock, write a rwlock, or just immutably borrow, and you wouldn't know unless you know how the type is implemented. I think using `.as_foo` for immutable types (`PyStr.as_str()`, `PyInt.as_bigint()`) and `borrow_foo[_mut]()` for types that have rwlocks much more clearly communicates the actual cost of accessing the data. In addition, `BorrowValue` only supports one type of access because you can only impl it once, so mutable objects had to have a separate `borrow_value_mut()` that's just its own method. I didn't remove `BorrowValue` (it still exists and is still implemented for the types it was before) because I do think it's good to be able to abstract over "type that can give you some sort of access to the data it contains", but I don't actually think it's being used at all in the codebase. @youknowone you created it, so do you think it's worth keeping or should we get rid of it completely?


youknowone (approved):

vm/src/builtins/list.rs
```diff
@@ -55,16 +58,20 @@ impl PyValue for PyList {
...
 
     fn borrow_value(&'a self) -> Self::Borrowed {
-        self.elements.read()
+        self.borrow_list()
     }
 }
 
 impl PyList {
-    pub fn borrow_value_mut(&self) -> PyRwLockWriteGuard<'_, Vec<PyObjectRef>> {
+    pub fn borrow_list(&self) -> PyMappedRwLockReadGuard<'_, [PyObjectRef]> {
```
youknowone, line 70: Do we expose names of implementation or names of python interfaces? Comparing to `as_bigint` or `as_str`, I think `list` doesn't give any hint what it returns in Rust side. (slice for read and vec for write?)

youknowone:
I expected it to work like `Borrow`, `Deref` or `AsRef`, but that doesn't look successful. So voting for removing.
I still want to give shared interfaces for similar types like bytes group (PyBytes, PyByteArray and PyStr), but it doesn't seem to be achieved by current BorrowValue.

vm/src/builtins/list.rs
coolreader18: I sorta specifically didn't want to name it after a rust type, since the 2 methods return different types, and I wanted them to be the same name just with or without `_mut`. Like `borrow_slice_mut()` would be inaccurate cause it's a Vec, and `borrow_vec()` would be inaccurate cause it's a slice. I suppose `borrow_vec` could be ok, since vec is short for vector anyway, but Rust usually prefers slice over vector for that sort of thing.

coolreader18:
Yeah, that seems like it would be something like `AnyStr`, maybe

vm/src/builtins/list.rs
youknowone: As you said, Rust code conventionally returns slice rather than read-only vec because they are practically the same and read-only vec doesn't have much mroe thing compare to slice. For the reason, I think `borrow_vec` can be just ok name. (though I think `borrow_slice` and `borrow_vec_mut` pair is better than that)

coolreader18: I'll switch it to borrow_vec; I think if someone realizes they need to borrow mutably rather than immutably, it's good to be able to have a rule of "just add `_mut` at the end" (or vice versa) rather than having to remember which types have special cases.

PR #2663, opened 2021-05-21
Add test_sysconfig from CPython 3.8

fanninpm:
EDIT: This will also update `os.py` to CPython 3.8 and fix tests in the `test_os.py` file.


Lib/os.py
```diff
@@ -92,6 +94,97 @@ def _get_exports_list(module):
 from os.path import (curdir, pardir, sep, pathsep, defpath, extsep, altsep,
     devnull)
 
+del _names
+
+
+if _exists("_have_functions"):
```
coolreader18, line 100: We don't need this if block, we implement the `supports_foo` sets in the `posix`/`nt` module.

coolreader18:
The windows error seems like a pretty easy fix, you just have to move `UnameResult` to `mod _os` instead of the posix-specific module, even though `uname()` is only available on posix it probably shims it on non-posix

Lib/os.py
fanninpm: [I wonder what it would break if it were removed from the CPython sources?](https://github.com/python/cpython/blob/4844abdd700120120fc76c29d911bcb547700baf/Lib/os.py#L100)

coolreader18: Well CPython doesn't do it in the native module, it just exports the `HAVE_x` constants that autoconf generates from the `posix`/`nt` module. But since the `os` functions don't necessarily directly correspond to native functions under the hood in RustPython, it's a bit easier to just generate it on the Rust side rather than making up fake constants.

coolreader18:
@fanninpm you probably didn't notice my comment (↑↑) cause we posted around the same time; the windows tests should be pretty easy to fix.

vm/src/stdlib/os.rs
```diff
@@ -2304,31 +2330,6 @@ mod posix {
...
-    fn uname(vm: &VirtualMachine) -> PyResult<UnameResult> {
-        let info = uname::uname().map_err(|err| err.into_pyexception(vm))?;
-        Ok(UnameResult {
-            sysname: info.sysname,
-            nodename: info.nodename,
-            release: info.release,
-            version: info.version,
-            machine: info.machine,
-        })
-    }
```
coolreader18, line 2331: You can keep this function here, but just change it to return `super::_os::UnameResult`

fanninpm: Do I need to make the `UnameResult` struct public to do that?

coolreader18: Oh, yeah, probably `pub(super)`

coolreader18 (approved):
Great, thanks!

coolreader18:
Oh or actually could you re-remove that if block? It really doesn't do anything in our setup.

PR #2664, opened 2021-05-21
Add test_tarfile from CPython 3.8

fanninpm:
This PR also adds files that will be used by other tests (which I will later publish PRs for).


fanninpm:
@coolreader18 that Ubuntu test failure [seems to be a heisenbug.](https://github.com/RustPython/RustPython/runs/2588137490?check_suite_focus=true)

youknowone (approved):

coolreader18:
@fanninpm yeah :/ file times are finnicky, I guess. We could probably just skip it as non-deterministic

PR #2666, opened 2021-05-22
Add test_tempfile from CPython 3.8

fanninpm:


fanninpm:
@coolreader18 do I need to mess with `os.rs` at all to get this to work?

coolreader18:
Hmm, well it looks like libc doesn't have `O_TEMPORARY` so I submitted a PR rust-lang/libc#2194, but for now you could add a `#[pyattr] const O_TEMPORARY: i32 = 0x40;` to `mod nt { ... }`, that should fix hopefully most of the errors.

Lib/tempfile.py
```diff
@@ -43,12 +43,9 @@
...
 import errno as _errno
 from random import Random as _Random
+import sys as _sys
 import weakref as _weakref
-
-try:
-    import _thread
-except ImportError:
-    import _dummy_thread as _thread
+import _thread
```
coolreader18, line 48: Could you keep this in? We patched it to allow running without `--features threading`

fanninpm: It was in there from CPython 3.6, but it got removed in response to [this issue on bugs.python.org](https://bugs.python.org/issue31370).

coolreader18: Yeah, we're probably going to keep threadless as an option for the time being, just since we don't have a GIL option and can't enable threading without taking a performance hit everywhere. Plus we target wasm which doesn't have threads.

coolreader18 (approved):

PR #2667, opened 2021-05-22
Add test_textwrap from CPython 3.8

fanninpm:


coolreader18:
Try updating textwrap.py as well - copying both on my machine seems to pass all the tests.

coolreader18 (approved):

PR #2669, opened 2021-05-23
os: argument in listdir should be optional

deantvv:
From `listdir(path)` to `listdir(path='.')`


coolreader18:
It looks like `test_listdir_default` in `test_posix.py` can also be unskipped.

deantvv:
> It looks like `test_listdir_default` in `test_posix.py` can also be unskipped.

Sorry for missing that.

coolreader18:
No worries! LGTM

coolreader18 (approved):

PR #2672, opened 2021-05-25
Add test_timeit from CPython 3.8

fanninpm:


coolreader18:
I think that `KeyError` can be fixed by changing line 300 of `vm/src/builtins/object.rs` to catch a `KeyError` and turn it into an `AttributeError`.

coolreader18 (approved):
Nice, thanks!

PR #2674, opened 2021-05-26
os: Implement times

deantvv:


coolreader18 (approved):
Awesome, thanks!

vm/src/stdlib/os.rs
```diff
@@ -1455,6 +1455,76 @@ mod _os {
...
+        pub children_user: f64,
+        pub children_system: f64,
+        pub elapsed: f64,
+    }
+
+    #[pyimpl(with(PyStructSequence))]
+    impl TimesResult {}
+
+    #[pyfunction]
+    fn times(vm: &VirtualMachine) -> PyResult {
```
coolreader18, line 1473: This doesn't really matter, but note that you could have this just return `PyResult<TimesResult>` - anything that implements `IntoPyObject` can be returned from a function and it will get automatically converted to a PyObjectRef.

PR #2676, opened 2021-05-27
Add test_traceback from CPython 3.8

fanninpm:


fanninpm:
@coolreader18 That Windows failure seems to be flaky. Should it be skipped?

coolreader18:
Yeah, probably, threading in general can be pretty flaky I think

coolreader18 (approved):

PR #2678, opened 2021-05-28
Add test_urllib from CPython 3.8

fanninpm:


coolreader18:
Wow, surprisingly few failures! Do you wanna try updating urllib itself as well?, I feel like that quoting one might be from "userspace"

fanninpm:
@coolreader18 I added/amended a commit to the `test_traceback` PR that will fix that spurious CI failure.

coolreader18 (approved):
LGTM! That test_threading_local one might actually be able to be resolved just by adding `n-2` to the list of the number of threads it collects.

PR #2679, opened 2021-05-29
Add Rust accelerator module for bisect.

DimitrisJim:
Adds bisect functionality implemented in Rust.

### DIFF
--- a/Lib/bisect.py
+++ b/Lib/bisect.py
@@ -19,8 +19,6 @@ def insort_right(a, x, lo=0, hi=None):
         else: lo = mid+1
     a.insert(lo, x)
 
-insort = insort_right   # backward compatibility
-
 def bisect_right(a, x, lo=0, hi=None):
     """Return the index where to insert item x in list a, assuming a is sorted.
 
@@ -41,8 +39,6 @@ def bisect_right(a, x, lo=0, hi=None):
         if x < a[mid]: hi = mid
         else: lo = mid+1
     return lo
-
-bisect = bisect_right   # backward compatibility
 
 def insort_left(a, x, lo=0, hi=None):
     """Insert item x in list a, and keep it sorted assuming a is sorted.
@@ -85,8 +81,11 @@ def bisect_left(a, x, lo=0, hi=None):
         else: hi = mid
     return lo
 
-# Overwrite above definitions with a fast C implementation
+# Overwrite above definitions with a fast Rust implementation
 try:
     from _bisect import *
 except ImportError:
     pass
+
+bisect = bisect_right   # backward compatibility
+insort = insort_right   # backward compatibility
--- a/vm/src/stdlib/bisect.rs
+++ b/vm/src/stdlib/bisect.rs
@@ -114,16 +114,38 @@ mod _bisect {
         Ok(lo)
     }
 
+    /// Return the index where to insert item x in list a, assuming a is sorted.
+    ///
+    /// The return value i is such that all e in a[:i] have e < x, and all e in
+    /// a[i:] have e >= x.  So if x already appears in the list, a.insert(x) will
+    /// insert just before the leftmost x already there.
+    ///
+    /// Optional args lo (default 0) and hi (default len(a)) bound the
+    /// slice of a to be searched.
     #[pyfunction]
     fn bisect_left(args: BisectArgs, vm: &VirtualMachine) -> PyResult<usize> {
         bisect_left_impl(args, vm)
     }
 
+    /// Return the index where to insert item x in list a, assuming a is sorted.
+    ///
+    /// The return value i is such that all e in a[:i] have e <= x, and all e in
+    /// a[i:] have e > x.  So if x already appears in the list, a.insert(x) will
+    /// insert just after the rightmost x already there.
+    ///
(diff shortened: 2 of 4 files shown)


Lib/bisect.py
```diff
@@ -85,8 +81,11 @@ def bisect_left(a, x, lo=0, hi=None):
         else: hi = mid
     return lo
 
-# Overwrite above definitions with a fast C implementation
+# Overwrite above definitions with a fast Rust implementation
```
coolreader18, line 84: In general you don't have to reword `C -> Rust` - there's a *lot* of mentions of C in the source code, so it's easier to just read "C" as "implementation language" than it is to change *all* mentions of C to rust. Applies to the test file as well.

vm/src/stdlib/bisect.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    use crate::function::OptionalArg;
+    use crate::slots::PyComparisonOp::Lt;
+    use crate::vm::VirtualMachine;
+    use crate::{ItemProtocol, PyObjectRef, PyResult};
+    use num_traits::ToPrimitive;
+
+    #[derive(FromArgs)]
+    struct BisectArgs {
+        #[pyarg(any, name = "a")]
```
coolreader18, line 15: `#[pyarg(any)] foo: Bar` implies `name = "foo"`, you don't have to specify `name` for these.

vm/src/stdlib/bisect.rs
```diff
@@ -0,0 +1,182 @@
...
+        lo: OptionalArg<PyObjectRef>,
+        #[pyarg(any, optional, name = "hi")]
+        hi: OptionalArg<PyObjectRef>,
+    }
+
+    // Handles objects that implement __index__ and makes sure index fits in needed isize.
+    #[inline]
+    fn handle_default(
+        arg: OptionalArg<PyObjectRef>,
+        default: Option<isize>,
```
coolreader18, line 29: I think it's better to just return `None` in the `Missing` case, and then you can choose what you want to do with the option and avoid `.unwrap()`s - e.g. instead of passing `handle_default(Some(1)).unwrap()`, you can do `handle_default()?.unwrap_or(1)`

vm/src/stdlib/bisect.rs
```diff
@@ -0,0 +1,182 @@
...
+        vm: &VirtualMachine,
+    ) -> PyResult<(usize, usize)> {
+        // We only deal with positives for lo, try_from can't fail.
+        // Default is always a Some so we can safely unwrap.
+        let lo = handle_default(lo, Some(0), vm)?
+            .map(|value| {
+                if value < 0 {
+                    return Err(vm.new_value_error("lo must be non-negative".to_owned()));
+                }
+                Ok(usize::try_from(value).unwrap())
```
coolreader18, line 63: Instead of unwrapping the result here, you could do `usize::try_from(value).map_err(|_| vm.new_error...)` - `usize::try_from(isize)` only fails when the isize is negative.

vm/src/stdlib/bisect.rs
```diff
@@ -0,0 +1,182 @@
...
+            })
+            .unwrap()?;
+        let hi = handle_default(hi, None, vm)?
+            .map(|value| {
+                if value < 0 {
+                    0
+                } else {
+                    // Can't fail.
+                    usize::try_from(value).unwrap()
+                }
```
coolreader18, line 73: Same here - `try_from().unwrap_or(0)`

vm/src/stdlib/bisect.rs
```diff
@@ -0,0 +1,182 @@
...
+    /// Return the index where to insert item x in list a, assuming a is sorted.
+    ///
+    /// The return value i is such that all e in a[:i] have e < x, and all e in
+    /// a[i:] have e >= x.  So if x already appears in the list, a.insert(x) will
+    /// insert just before the leftmost x already there.
+    ///
+    /// Optional args lo (default 0) and hi (default len(a)) bound the
+    /// slice of a to be searched.
+    #[pyfunction]
+    fn bisect_left(args: BisectArgs, vm: &VirtualMachine) -> PyResult<usize> {
```
coolreader18, line 126: You could have `bisect_left_impl` just *be* the function with the `#[pyfunction]` - you can still call functions with `#[pyfunction]` from other Rust functions.

coolreader18:
Overall this looks good!

vm/src/stdlib/bisect.rs
DimitrisJim: @coolreader18 The only thing I was unsure of is how `#[inline]` affects a `#[pyfunction]`, that's why I separated them (expecting a reviewer to point things out :)). Does it make no difference? If so, your suggestion is clearly the best approach.

coolreader18: The `fn` item the macro's applied to *should* just be passed straight through (last time I looked at the macro code at least), so `#[inline]` and everything works like a normal function would.

DimitrisJim:
@coolreader18 Done, hope the changes addressed your review.

coolreader18 (approved):
Nice work, thanks for contributing!

PR #2681, opened 2021-05-30
Add test_heapq from CPython 3.8.

DimitrisJim:


Lib/test/test_heapq.py
```diff
@@ -0,0 +1,456 @@
...
+        target = sorted(data, reverse=True)
+        self.assertEqual(hsort(data, LT), target)
+        self.assertRaises(TypeError, data, LE)
+
+
+class TestHeapPython(TestHeap, TestCase):
+    module = py_heapq
+
+
+@unittest.skip("TODO: RUSTPYTHON, _heapq module")
```
DimitrisJim, line 281: Note: Could change these to `@skipUnless(c_heapq, 'requires _heapq')` as they were in the original source file.

coolreader18: Yeah, I think that'd be good, if it's possible to leave something as it is it's usually good to do so.

Lib/test/test_heapq.py
```diff
@@ -9,7 +9,7 @@
 from operator import itemgetter
 
 py_heapq = support.import_fresh_module('heapq', blocked=['_heapq'])
-c_heapq = support.import_fresh_module('heapq', fresh=['_heapq'])
+c_heapq = None
```
coolreader18, line 12: Is this necessary to change? I feel like I've seen these working in other test files, and a benefit of leaving the `import_fresh_module` is that the test file doesn't require any changes once the native module is implemented.

DimitrisJim: Indeed, this actually doesn't raise. Thanks for the feedback, will get to it now!

DimitrisJim:
@coolreader18 Done, thanks for the review!

coolreader18:
@DimitrisJim I just pushed a commit that adds `method.__self__`, I think that should fix some/most of the tests?

DimitrisJim:
@coolreader18 yes! All tests now pass just fine so no reason for the second commit.

coolreader18 (approved):
Great, thanks for contributing!

PR #2684, opened 2021-06-03
Add `builtin.__self__`.

DimitrisJim:
Note a discrepancy: in the [Reference manual](https://docs.python.org/3/reference/datamodel.html), it is stated that `__self__` for builtins should return `None`. This is what I've currently done.

On the other hand, `CPython` currently returns the module the function is defined in. 

If the module is to be returned, what's the best way to do it? Yank it from the `globals` using the module name?

### DIFF
--- a/vm/src/builtins/builtinfunc.rs
+++ b/vm/src/builtins/builtinfunc.rs
@@ -121,6 +121,10 @@ impl PyBuiltinFunction {
     fn doc(&self) -> Option<PyStrRef> {
         self.value.doc.clone()
     }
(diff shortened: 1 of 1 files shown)


coolreader18:
Hopefully it should be fine as-is; the reason CPython does it like that is that in CPython, *all* builtin functions have a self argument - for methods, that's the method receiver, and for module functions, it's the module it belongs to, which allows it to access the module state/other exports of the module. For now we don't need to change to fit that, afaict, so this works fine.

coolreader18 (approved):

PR #2688, opened 2021-06-04
Fix arguments passed to make_funcdef.

DimitrisJim:
Wrong values passed for `__name__` and `__module__` for functions built with `named_function`.


coolreader18 (approved):
Oh, whoops! Nice catch! :sweat_smile:

PR #2689, opened 2021-06-06
Handle NotImplemented return from length_hint.

DimitrisJim:
Essentially, by making `length_hint` take a `default` argument instead of returning an `Option`.

### DIFF
--- a/Lib/test/test_operator.py
+++ b/Lib/test/test_operator.py
@@ -516,11 +516,6 @@ class PyOperatorTestCase(OperatorTestCase, unittest.TestCase):
 class COperatorTestCase(OperatorTestCase, unittest.TestCase):
     module = c_operator
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
-    def test_length_hint(self):
-        super().test_length_hint()
-
 
 class OperatorPickleTestCase:
     def copy(self, obj, proto):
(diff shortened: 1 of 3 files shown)


coolreader18:
Sorry I'm late to this; do you think you could keep the Option return value? To me, I think Option is preferable in almost every case than a default argument, as with an option you can always tell unambiguously whether it was "some" or not, and you can still get the same behavior using `unwrap_or()`

DimitrisJim:
@coolreader18 No problem, many PR's for one person! My reasoning here was that we should never really expect to get a `None` out of it, it either returns a `usize` or it has `Err`or'ed.

I'll get to the changes tomorrow, thanks for the feedback.

DimitrisJim:
@coolreader18 done. Hope this addresses your review.

coolreader18:
The default arg is still there and it still returns `Some(default)` - you could probably revert back to how master was an just add the `not_implemented` check.

DimitrisJim:
@coolreader18 ah, I see now. `default` is required here since we need something to return if `length_hint` is `NotImplemented` (this isn't always zero, `operator.length_hint` can pass its own default (and this is what the test was failing on).

coolreader18:
Right, but that could be implemented just by doing `unwrap_or(passed_default)` in `operator.length_hint`. I think in C that default argument pattern is more common, but in Rust it almost always makes more sense to return an Option.

DimitrisJim:
@coolreader18 Your point finally got through to me! The latest changes should be what is needed.

coolreader18 (approved):
Heh, no worries. This looks good, thanks for fixing it!

PR #2690, opened 2021-06-06
Refactor unary ops.

DimitrisJim:
Add after binops in `vm.rs` and handle them in a similar fashion. Update frame and `abs` to forward calls to them.


coolreader18 (approved):
Nice refactoring!

PR #2691, opened 2021-06-06
Add dbm from CPython 3.8

verhovsky:
just the Python implementation. `_dbm` and `_gdbm` will have to be re-implemented in Rust


coolreader18 (approved):
Nice, all tests already passing!

PR #2694, opened 2021-06-07
os: Fix openpty test

deantvv:
Make openpty return non-inheritable file descriptors

Related to #1175

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -2371,6 +2371,9 @@ mod posix {
     #[pyfunction]
     fn openpty(vm: &VirtualMachine) -> PyResult {
         let r = nix::pty::openpty(None, None).map_err(|err| err.into_pyexception(vm))?;
+        for fd in &[r.master, r.slave] {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/os.rs
```diff
@@ -2371,6 +2372,10 @@ mod posix {
     #[pyfunction]
     fn openpty(vm: &VirtualMachine) -> PyResult {
         let r = nix::pty::openpty(None, None).map_err(|err| err.into_pyexception(vm))?;
+        for fd in &[r.master, r.slave] {
+            fcntl::fcntl(*fd, fcntl::F_SETFD(fcntl::FdFlag::FD_CLOEXEC))
```
DimitrisJim, line 2376: Not really versed in `os` stuff but, I see there's `raw_set_inheritable` over on line 2071. Wouldn't that do the trick here?

coolreader18: Yep, that'd be good to use for this @deantv

deantvv: Yes, I should use this. Fixed in new commit.

coolreader18 (approved):
Nice!

PR #2698, opened 2021-06-09
Use a Utf8size struct to represent the {bytes,chars} pairs in textio

coolreader18:
The name is a "pun" on `usize` - `s/u/Utf8`

### DIFF
--- a/vm/src/stdlib/io.rs
+++ b/vm/src/stdlib/io.rs
@@ -1946,6 +1946,58 @@ mod _io {
         }
     }
 
+    /// A length of or index into a UTF-8 string, measured in both chars and bytes
+    #[derive(Debug, Default, Copy, Clone)]
+    struct Utf8size {
+        bytes: usize,
+        chars: usize,
+    }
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/io.rs
```diff
@@ -2688,13 +2738,18 @@ mod _io {
                         PyStr::from(self.slice()).into_ref(vm)
                     }
                 }
+                fn utf8_len(&self) -> Utf8size {
```
youknowone, line 2741: i think this name is confusing due to `char::len_utf8()`

youknowone (approved):

vm/src/stdlib/io.rs
coolreader18: Yeah, but I think this a decent name - gets the length in units of utf8 codepoints and utf8 bytes.

coolreader18: At the very least I can't think of an alternative

PR #2705, opened 2021-06-13
os: Implement getgroups

deantvv:
`getgroups` is already implement in rust, just wrap the function to let it be used in python.


youknowone (approved):
nice work, thanks!

PR #2707, opened 2021-06-13
os: Implement (f)pathconf

deantvv:
Implement pathconf and fpathconf using libc::pathconf.

os.pathconf_names is not implemented.


youknowone (approved):
wow, PathconfVar is a lot more complicated than I expected

PR #2708, opened 2021-06-13
Add operator.rs

DimitrisJim:


vm/src/stdlib/operator.rs
```diff
@@ -1,62 +1,731 @@
...
+    fn xor(a: PyObjectRef, b: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
+        vm._xor(&a, &b)
+    }
+
+    // Sequence based operators
+
+    /// Return a + b for a and b sequences.
+    #[pyfunction]
+    fn concat(a: PyObjectRef, b: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
+        // Best attempt at checking that a is sequence-like.
```
DimitrisJim, line 214: Looking for tips on how to improve this (and `iconcat`) if anyone has ideas!

youknowone: Maybe PySequence_Check in CPython? https://docs.python.org/3/c-api/sequence.html

I have no idea in RustPython.

youknowone (approved):

vm/src/stdlib/operator.rs
DimitrisJim: Yeah, it's similar to the question I asked you the other day about a `PyNumber_Check` like method. We don't currently have the abstract interface.

PR #2709, opened 2021-06-14
Fix iteration issues, pickling by marking iterators as exhausted.

DimitrisJim:
An iterator should be able to know when it has been exhausted. Certain iterations through the iterator and pickling currently require this, a number of test cases could be fixed.

The approach uses an `AtomicCell` as was done in `PyCallableIterator`, it just packages the cases in an enum to make it publicly available. If any other approaches seem better, please share!

Minimal application of it, if the changes are agreeable, I can add PRs for the other types.

### DIFF
--- a/vm/src/builtins/list.rs
+++ b/vm/src/builtins/list.rs
@@ -5,8 +5,14 @@ use std::ops::DerefMut;
 
 use crossbeam_utils::atomic::AtomicCell;
 
+use super::int;
+use super::iter::{
+    IterStatus,
+    IterStatus::{Active, Exhausted},
+};
 use super::pytype::PyTypeRef;
 use super::slice::PySliceRef;
+use super::PyInt;
 use crate::common::lock::{
     PyMappedRwLockReadGuard, PyRwLock, PyRwLockReadGuard, PyRwLockWriteGuard,
 };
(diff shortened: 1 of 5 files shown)


vm/src/builtins/list.rs
```diff
@@ -458,6 +465,7 @@ fn do_sort(
 #[derive(Debug)]
 pub struct PyListIterator {
     pub position: AtomicCell<usize>,
+    status: AtomicCell<IterStatus>,
     pub list: PyListRef,
```
DimitrisJim, line 469: any reason why these are public currently?

DimitrisJim, line 469: any reason why these are public currently?

youknowone: I have no idea. @coolreader18 is this a part of public API?

vm/src/builtins/list.rs
```diff
@@ -471,19 +479,61 @@ impl PyValue for PyListIterator {
...
+            let position = std::cmp::min(
+                int::try_to_primitive(i.as_bigint(), vm).unwrap_or(0),
+                self.list.len(),
+            );
+            self.position.store(position);
+            Ok(vm.ctx.none())
+        } else {
+            Err(vm.new_type_error("an integer is required.".to_owned()))
+        }
+    }
```
youknowone, line 508: this function looks like always returning `Ok(vm.vtx.none())` when it returns `Ok`. By using `PyResult<()>`, `Ok(())` will work instead of returning `vm.ctx.none()` everytime.

youknowone (changes requested):
Thanks, this is a good improvement. I left a small suggestion about return value. Would you check it?

DimitrisJim:
@youknowone Thanks for the feedback. Done!

youknowone (approved):
Thank you. I like the part changing bool to enum.

PR #2710, opened 2021-06-14
Fix miscalculation when reversing a range.

DimitrisJim:
which caused the test failure.


youknowone (approved):
oh, nice catch

DimitrisJim:
I found more cases while playing around with other range stuff. I've been working on range objects since yesterday though and I'll fix `reverse` fully (I hope!) in another PR that adds a number of fixes for other tests in `range`.

PR #2711, opened 2021-06-15
Fix a number of range issues.

DimitrisJim:
In short:

1. Added fallback for when item isn't found via `index/count/contains`, iterate through the range and invoke `__eq__` with every element.
2. Contains has a fast lane only for exact ints. Subclasses should fall back to normal slow iteration described in 1.
3. Use similar logic as cpython/ironpython when calculating new start,stop,step in `__reverse__`. There's some corner cases current approach doesn't cover. (see `reversed(range(-2, -1, -3))` for an example)
4. `is_empty` just invokes `length`.

All in all, 5 more tests should now pass.

### DIFF
--- a/Lib/test/test_range.py
+++ b/Lib/test/test_range.py
@@ -255,8 +255,6 @@ class RangeTest(unittest.TestCase):
         self.assertRaises(TypeError, range, 0.0, 0.0, 1)
         self.assertRaises(TypeError, range, 0.0, 0.0, 1.0)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_index(self):
         u = range(2)
         self.assertEqual(u.index(0), 0)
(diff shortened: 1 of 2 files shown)


Lib/test/test_range.py
```diff
@@ -490,8 +484,10 @@ def test_empty(self):
         self.assertNotIn(-1, r)
         self.assertNotIn(1, r)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+    # TODO: RUSTPYTHON.
+    # NOTE: Test passes, fast iterators are required for this to not
+    #       take ~ 1m.
+    @unittest.skip
```
DimitrisJim, line 490: As the comment suggests, this takes time. Since the test suite is already lengthy, I've skipped it until fast iterators for ranges are implemented. Should I un-skip?

youknowone: I think this skip is ok. We are skipping a few tests like this.

youknowone (approved):
Nice, a big improvement! Thank you

PR #2715, opened 2021-06-17
Make `frozenset(empty_iterable)` return a singleton.

DimitrisJim:
This covers creating an empty frozenset through Python code. Modules implemented in rust could still use `from_iter` (as `ssl` does) and end up with an empty frozenset which isn't the singleton.

### DIFF
--- a/vm/src/builtins/set.rs
+++ b/vm/src/builtins/set.rs
@@ -5,7 +5,6 @@ use super::pytype::PyTypeRef;
 use crate::common::hash::PyHash;
 use crate::common::rc::PyRc;
 use crate::dictdatatype;
-use crate::function::OptionalArg::{Missing, Present};
 use crate::function::{Args, OptionalArg};
 use crate::slots::{Comparable, Hashable, Iterable, PyComparisonOp, PyIter, Unhashable};
 use crate::vm::{ReprGuard, VirtualMachine};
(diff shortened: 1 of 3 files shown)


DimitrisJim:
clippy appears to be complaining for something unrelated which I can't replicate (even after updating to `0.1.53` which was released earlied today).

vm/src/builtins/set.rs
```diff
@@ -573,23 +572,26 @@ impl PyFrozenSet {
         iterable: OptionalArg<PyObjectRef>,
         vm: &VirtualMachine,
     ) -> PyResult<PyRef<Self>> {
-        let iterable = if let Present(iterable) = iterable {
-            if cls.is(&vm.ctx.types.frozenset_type) {
-                match iterable.downcast_exact::<PyFrozenSet>(vm) {
-                    Ok(iter) => return Ok(iter),
-                    Err(iterable) => Present(PyIterable::try_from_object(vm, iterable)?),
+        let elements = if let OptionalArg::Present(iterable) = iterable {
+            let iterable = if cls.is(&vm.ctx.types.frozenset_type) {
```
DimitrisJim, line 576: Though I've seen this used in other `__new__`'s it seems superfluous. Wouldn't:

```rust
let iterable = match iterable.downcast_exact::<Self>(vm) {
    Ok(fs) => return Ok(fs),
    Err(iterable) => iterable,
};
```

suffice in only allowing `frozenset` instances (and not subclasses of `frozenset`!) to return immediately?

youknowone (approved):

youknowone (changes requested):
By this change, `#[allow(dead_code)]` of `from_iter`will be useless. Could you remove the annotation too?

DimitrisJim:
@youknowone Removed that annotation (left the comment though, it seems helpful to have it around).

youknowone (approved):
It looks great! Thank you!

PR #2719, opened 2021-06-22
Don't export stdlib modules.

DimitrisJim:
`io` and `signal` are used elsewhere in the crate so limit their visibility there.


youknowone:
@coolreader18 are those public api by design?

youknowone (approved):

PR #2720, opened 2021-06-24
Change `os.rs` to build on musl-libc

sisungo:
On `musl-libc`-based targets, `libc::copy_file_range` is not provided, but `libc::syscall` and `libc::SYS_copy_file_range` are supported. This pull request use them instead of `libc::copy_file_range`.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1562,7 +1562,8 @@ mod _os {
         // Safety: p_offset_src and p_offset_dst is a unique pointer for offset_src and offset_dst respectively,
         // and will only be freed after this function ends.
         let ret = unsafe {
-            libc::copy_file_range(
+            libc::syscall(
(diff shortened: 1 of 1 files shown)


youknowone (changes requested):
The change looks good to me, but I worry people who will look at this code will be confused why not using `copy_file_range`. Would you add a comment about musl just like you already described in the PR text?
Because we currently don't have a CI setup for musl-libc, keeping this change will not be easy without comment.

sisungo:
See #2726.

ShadowJonathan:
...why did you make another PR? You can commit to the existing branch to update the exact change you're merging into the repository.

youknowone:
Yes, exactly as @ShadowJonathan said. When you have another PR in future, please try to force-push to update your previous patch. This is not only a convention in RustPython, but widely shared one.

Thank you for the patch! This  is a good patch regardless how many PR is made for it.

PR #2722, opened 2021-07-07
Lib: Update operator.py to follow CPython's implementation

corona10:
There was an inconsistent implementation issue between C and Python versions from CPython.
The regression did not happen until the CPython user doesn't use _operator module.
But since RustPython only uses pure Python implementation, the implementation should be updated.

ref: https://bugs.python.org/issue44558


youknowone (approved):
Oh, nice catch. Thank you!

PR #2733, opened 2021-07-11
math: raise OverflowError when result is infinite

yossi-k:
Matches CPython behavior.

For example, `math.exp(1e6)` raises `OverflowError` in cpython, but the previous RustPython implementation just returned `inf`.

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -20,8 +20,10 @@ use std::cmp::Ordering;
 // Helper macro:
 macro_rules! make_math_func {
     ( $fname:ident, $fun:ident ) => {
-        fn $fname(value: IntoPyFloat) -> f64 {
-            value.to_f64().$fun()
+        fn $fname(value: IntoPyFloat, vm: &VirtualMachine) -> PyResult<f64> {
+            let value = value.to_f64();
+            let result = value.$fun();
+            result_or_overflow(value, result, vm)
         }
     };
 }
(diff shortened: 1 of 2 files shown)


DimitrisJim (approved):
Looks good to me.

DimitrisJim:
Shame the test for `ldexp` still fails. Some handling of `NaNs` seems to be required.

yossi-k:
@DimitrisJim I fixed `ldexp` as well, it only required handling non-finite cases differently like you said.

vm/src/stdlib/math.rs
```diff
@@ -18,10 +18,24 @@ use rustpython_common::float_ops;
...
+    ( $value:ident, $result:ident, $vm: ident ) => {
+        if !$result.is_finite() && $value.is_finite() {
+            // CPython doesn't return `inf` when called with finite
+            // values, it raises OverflowError instead.
+            Err($vm.new_overflow_error("math range error".to_owned()))
+        } else {
+            Ok($result)
+        }
+    }
+}
```
youknowone, line 31: Could this macro be replacable by an inline function? Unlike `make_math_func`, this macro looks like a funciton rather than a macro.

The code itself looks fine.

youknowone (changes requested):
@DimitrisJim Thank you for the review!

vm/src/stdlib/math.rs
yossi-k: Sounds good, does it look OK now?

youknowone (approved):
Yes, it looks good. Thank you!

youknowone:
And thanks a lot to @DimitrisJim for the review, it helps a lot!

PR #2743, opened 2021-07-15
Fix failure -- use PyStrRef to extract String.

DimitrisJim:
Most recent PR removed the `TryFromObject` implementation for `CString` which was used in another PR that was added before it. This uses `PyStrRef` to do the same thing and should fix the error.


youknowone (approved):
Nice catch! This was the build failure from MacOS

PR #2744, opened 2021-07-15
Initial implementation of pyexpat

yossi-k:
Many python libraries, for example `plistlib`, use `expat` to parse xml.

This is a very initial implementation for the `pyexpat` module using `xml-rs` as the actual xml parser. 
This fixes `plistlib`, which allows running `platform.mac_ver()` - now we can install pip on Macs!

I would like to get some feedback about this before I add more error handling and other missing features.

### DIFF
--- a/vm/src/stdlib/pyexpat.rs
+++ b/vm/src/stdlib/pyexpat.rs
@@ -1,0 +1,205 @@
+/* Pyexpat builtin module
+*
+*
+*/
+
+use crate::vm::VirtualMachine;
+use crate::PyObjectRef;
+
+pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
+    let module = _pyexpat::make_module(vm);
+
+    extend_module!(vm, module, {
+        "errors" => _errors::make_module(vm),
+        "model" => _model::make_module(vm)
+    });
+
+    module
+}
+
+#[pymodule(name = "pyexpat")]
+mod _pyexpat {
+    use crate::builtins::{PyStr, PyStrRef, PyTypeRef};
+    use crate::byteslike::PyBytesLike;
+    use crate::function::{IntoFuncArgs, OptionalArg};
+    use crate::pyobject::StaticType;
+    use crate::{
+        IntoPyObject, IntoPyRef, ItemProtocol, PyObjectRef, PyRef, PyResult, PyValue,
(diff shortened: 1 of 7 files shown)


vm/src/stdlib/pyexpat.rs
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    #[derive(FromArgs)]
+    struct ParserCreateArgs {
+        #[pyarg(any, optional)]
+        encoding: OptionalArg<PyStrRef>,
+        #[pyarg(any, optional)]
+        namespace_separator: OptionalArg<PyStrRef>,
+        #[pyarg(any, optional)]
+        intern: OptionalArg<PyStrRef>,
```
DimitrisJim, line 188: What's `intern`s intended use? I can't see it in the signature of `ParserCreate` in the docs.

yossi-k: I saw it in CPython source. I can remove it though, we're not using this argument now.

yossi-k:
What is missing to get this PR merged? (Obviously the tests aren't passing but that's waiting for #2743 to be merged).

Also, does `xml-rs` look OK? I took another look and saw that `quick-xml` is also quite similar and more actively developed, perhaps we should be using this instead?

youknowone:
Great! I cannot wait to run this PR. I merged the blocker PR. Could you please rebase this PR?

vm/src/stdlib/pyexpat.rs
```diff
@@ -0,0 +1,185 @@
...
+
+    #[pyimpl]
+    impl PyExpatLikeXmlParser {
+        fn new(vm: &VirtualMachine) -> PyResult<PyExpatLikeXmlParserRef> {
+            Ok(PyExpatLikeXmlParser {
+                start_element: MutableObject::new(vm.ctx.none()),
+                end_element: MutableObject::new(vm.ctx.none()),
+                character_data: MutableObject::new(vm.ctx.none()),
+                entity_decl: MutableObject::new(vm.ctx.none()),
+                buffer_text: MutableObject::new(vm.ctx.new_bool(false))
```
youknowone, line 82:
```suggestion
                buffer_text: MutableObject::new(vm.ctx.new_bool(false)),
```
Fix rustfmt error

youknowone (approved):

PR #2747, opened 2021-07-20
Fake gc

youknowone:


DimitrisJim:
ah, I'd actually like this now. I'm trying to iron out some issues `pyperf` has when used through RustPython and one of them is it uses `gc` if it is importable so a dummy `gc` would be needed.

Can we run the CI on this again to see where it fails?

PR #2749, opened 2021-07-24
os: fix fdopen invalid file descriptor

deantvv:
Related to #1175

### DIFF
--- a/vm/src/stdlib/io.rs
+++ b/vm/src/stdlib/io.rs
@@ -8,6 +8,9 @@ cfg_if::cfg_if! {
         type Offset = i64;
     }
 }
+
+#[cfg(unix)]
+use crate::stdlib::os::{errno_err, PathOrFd};
 use crate::VirtualMachine;
 use crate::{PyObjectRef, PyResult, TryFromObject};
 pub(crate) use _io::io_open as open;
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/io.rs
```diff
@@ -8,6 +8,8 @@ cfg_if::cfg_if! {
         type Offset = i64;
     }
 }
+use crate::stdlib::os::errno_err;
```
DimitrisJim, line 11: Place a `cfg` here and on the next line to get rid of the warnings for windows.

DimitrisJim:
I can't see why the test for wasm is error'ing. Could be a dud.

deantvv:
@youknowone Could you please restart the CI test for macos? It seems that it panic at some weird place which is not related to this PR. Thanks!

youknowone:
@deantvv I think there is an actual bug. Please check @DimitrisJim's review. It also affect WASM.

See Line 58 of the failing test.

```
error[E0432]: unresolved import `crate::stdlib::os`
  --> vm/src/stdlib/io.rs:11:20
   |
11 | use crate::stdlib::os::errno_err;
   |                    ^^ could not find `os` in `stdlib`
```

vm/src/stdlib/io.rs
deantvv: Thanks for catching this!

deantvv:
> @deantvv I think there is an actual bug. Please check @DimitrisJim's review. It also affect WASM.
> 
> See Line 58 of the failing test.
> 
> ```
> error[E0432]: unresolved import `crate::stdlib::os`
>   --> vm/src/stdlib/io.rs:11:20
>    |
> 11 | use crate::stdlib::os::errno_err;
>    |                    ^^ could not find `os` in `stdlib`
> ```
> 
> For local dev environment,
> 
> ```
> cd wasm/lib && cargo run
> ```

Sorry for this. I fixed this in my local environment but forgot to commit the change.

youknowone (approved):

PR #2750, opened 2021-07-24
os: implement fork

yossi-k:
This seems to work and tests pass (`test_posix:test_register_at_fork`).

I took a look at CPython's implementation and it seems they're doing more things before/after fork, namely acquiring the import lock, and re-initiating the interpreter in the child process (see here - https://github.com/python/cpython/blob/main/Modules/posixmodule.c#L593). Do we want to do the same thing here?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -57,9 +57,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.45"
+version = "1.0.59"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


youknowone (approved):
I personally support hard compatibility with CPython.
At least, leaving comments about CPython implementations for each places will be helpful for future workers.

youknowone:
what's happend to the ubuntu tests?

yossi-k:
I’ll take another look. The problem is that fork enables functions like os.spawn which in turn enable a whole bunch of new tests.

DimitrisJim:
thanks! This is superseded by https://github.com/RustPython/RustPython/pull/4877 if I'm not mistaken. Feel free to re-open if I'm mistaken.

PR #2751, opened 2021-07-25
socket test remove unnecessary skip

deantvv:


deantvv:
`os.sendfile` in macos need more fix. Closing.

PR #2753, opened 2021-07-25
Make set iteration linear

DimitrisJim:
Previously, `zelf.keys` was called for every invocation of `next`resulting in an unfortunate quadratic runtime. This pr just grabs keys and makes them part of the iterator instead. As a result, iterating through a set iterator is now linear.

This should speed up the test-suite by, I'm hoping, at least a minute or two. Specifically `test_itertools` has improved _a lot_ since it has many calls of the form `sorted(set_instance)` with large sets.


youknowone (approved):
Nice catch!

PR #2754, opened 2021-07-25
Add fixes for itertools.repeat

DimitrisJim:
Specifically:
 - an implementation of `repr`.
 - an implementation for `reduce`
 - a fix for the arguments, namely, allowing named arguments for `object` and `times`
 - fixing `length_hint` to return `TypeError` if `times` is missing, that way `operator.length_hint` returns a default.

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -260,7 +260,7 @@ mod decl {
     #[derive(Debug)]
     struct PyItertoolsRepeat {
         object: PyObjectRef,
-        times: Option<PyRwLock<BigInt>>,
+        times: Option<PyRwLock<usize>>,
     }
 
     impl PyValue for PyItertoolsRepeat {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -269,40 +269,81 @@ mod decl {
...
-            object: PyObjectRef,
-            times: OptionalArg<PyIntRef>,
+            PyRepeatNewArgs { object, times }: PyRepeatNewArgs,
             vm: &VirtualMachine,
         ) -> PyResult<PyRef<Self>> {
-            let times = times
-                .into_option()
-                .map(|int| PyRwLock::new(int.as_bigint().clone()));
-
+            let times = match times.into_option() {
```
DimitrisJim, line 288: `times` was changed from a `BigInt` to a `usize` with a max value of `isize::MAX`. This matches current implementation in `CPython` (which uses a `Py_ssize_t`) and does make its execution faster.

youknowone (approved):

PR #2756, opened 2021-07-26
Fix arguments for islice. 3 args are allowed.

DimitrisJim:
Fixes test for `islice` and also un-blocks doctest for `itertools` to finally be able to run. Also fixes acceptable values for `start`, `step` and `stop` to be in the range `0 <= value <= sys.maxsize` when they previously were in range `0 <= value <= usize::MAX`.

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -5,6 +5,7 @@ mod decl {
     use crossbeam_utils::atomic::AtomicCell;
     use num_bigint::BigInt;
     use num_traits::{One, Signed, ToPrimitive, Zero};
+    use std::convert::TryFrom;
     use std::fmt;
 
     use crate::builtins::int::{self, PyInt, PyIntRef};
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -682,13 +683,24 @@ mod decl {
         }
     }
 
-    fn pyobject_to_opt_usize(obj: PyObjectRef, vm: &VirtualMachine) -> Option<usize> {
+    // Restrict obj to ints with value 0 <= val <= sys.maxsize (isize::MAX).
+    // On failure (out of range, non-int object) a ValueError is raised.
+    fn pyobject_to_opt_usize(obj: PyObjectRef, name: &str, vm: &VirtualMachine) -> PyResult<usize> {
```
youknowone, line 688: Since `name` is always a const, `name: &'static str` would be better.

vm/src/stdlib/itertools.rs
```diff
@@ -682,13 +683,24 @@ mod decl {
...
         let is_int = obj.isinstance(&vm.ctx.types.int_type);
         if is_int {
-            int::get_value(&obj).to_usize()
-        } else {
-            None
+            let value = int::get_value(&obj).to_usize();
+            if let Some(value) = value {
+                // Only succeeds for values for which 0 <= value <= isize::MAX
+                if value <= usize::try_from(isize::MAX).unwrap() {
```
youknowone, line 694: Because `isize::MAX` is a const, I think `isize::MAX as usize` or `usize::try_from(isize::MAX).unwrap_or_else(|| unsafe { std::hint::unreachable_unchecked() })` is safe enough

youknowone (changes requested):

vm/src/stdlib/itertools.rs
DimitrisJim: Thanks for the feedback, I went with your first suggestion.

youknowone (approved):

PR #2758, opened 2021-07-28
Fix builtins.__name__

### LINKED ISSUE #2757: Value of builtins.__name__ is incorrect
From an interactive prompt or inside a Python script, in RustPython:
```python
__builtins__.__name__ == '__main__' # incorrect
__name__ == '__main__' # correct

import builtins
builtins.__name__ == '__main__' # incorrect
```
whereas in CPython:
```python
__builtins__.__name__ == 'builtins' # correct
__name__ == '__main__' # correct

import builtins
builtins.__name__ == 'builtins' # correct
```

As you can see, `builtins.__name__` is incorrectly set to `__main__` instead of `builtins`.

I think this issue was introduced by PR https://github.com/RustPython/RustPython/pull/149, which was intended to fix the global `__name__`.

`__name__` was fixed properly with PR https://github.com/RustPython/RustPython/pull/1223, so the fix in PR https://github.com/RustPython/RustPython/pull/149 (which breaks `builtins.__name__`) is no longer needed.

yaxollum:
Fixes #2757 

Can be tested with:
```
cargo run --release extra_tests/snippets/name.py
```

### DIFF
--- a/extra_tests/snippets/name.py
+++ b/extra_tests/snippets/name.py
@@ -7,7 +7,3 @@ from import_name import import_func
 import_func()
 
 assert __name__ == "__main__"
-
-assert __builtins__.__name__ == "builtins"
-import builtins
-assert builtins.__name__ == "builtins"
--- a/extra_tests/snippets/name.py
+++ b/extra_tests/snippets/name.py
@@ -7,3 +7,7 @@ from import_name import import_func
 import_func()
 
 assert __name__ == "__main__"
+
+assert __builtins__.__name__ == "builtins"
+import builtins
+assert builtins.__name__ == "builtins"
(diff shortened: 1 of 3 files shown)


extra_tests/snippets/name.py
```diff
@@ -7,3 +7,7 @@
 import_func()
 
 assert __name__ == "__main__"
+
+assert __builtins__.__name__ == "builtins"
+import builtins
+assert builtins.__name__ == "builtins"
```
youknowone, line 13: The test code is good, but I want to suggest to replace it to `builtins_module.py` instead of here.

youknowone (changes requested):

youknowone (approved):
Thank you for contributing!

Issue #2732, opened 2021-07-11
Some failures in whats_left.sh: method inheritance not as expected


jamestwebber:
## Feature

The `whats_left.sh` script generates a list of incompatibilities by comparing signatures with `inspect.signature()`. I think one source of mismatches is because of how builtin object inherit (or don't inherit) from `object`. I don't yet understand how RustPython is doing this so I'm documenting the issue here, but I could write a PR with some help.

### Description

As an example: `sys.__loader__` is listed by `whats_left` because `signature()` returns a ValueError. The CPython signature is `()`, which is the default signature of `object`.

In both CPython and RustPython `inspect.signature` is calling `inspect._signature_from_callable()` ([code](https://github.com/RustPython/RustPython/blob/master/Lib/inspect.py#L2206)), which eventually gets to [this block](https://github.com/RustPython/RustPython/blob/master/Lib/inspect.py#L2367-L2373).

The class is an instance of `_frozen_importlib.BuiltinImporter` which has no `__init__` or `__new__` method defined. In CPython, both of those methods are the same as the `object` methods (using the `is` test, e.g. the same memory address). In RustPython the `__new__` method is different. I would guess that something about the class definition is creating a separate `__new__` method even though there shouldn't be one.

youknowone:
@jamestwebber is this fixed by your PRs?

jamestwebber:
Unfortunately no, I just collapse the error into a single message.

In the case of `__loader__` the same message showed up many times because every built-in module had a `__loader__` instance that came from `_frozen_importlib`. So my PRs turn that into one message but the underlying cause is still there.

moreal:
I resolved #2869 so I believe this issue is also resolved as I understood. To prove this issue is resolved, I executed `./whats_left.sh` and looked up `sys.__loader__` but I couldn't. But I can't feel sure that this issue was resolved, could you check it? @jamestwebber

jamestwebber:
Yes, this is fixed! The messages were no longer output by `whats_left` because in #2760 I tweaked the code to remove items defined in other modules (in retrospect perhaps this was too broad, as there are no examples of `__loader__` left to check). But I verified the change in behavior which fixes this issue:

```python
>>>>> import sys
>>>>> sys.__loader__.__init__ is object.__init__
True
>>>>> sys.__loader__.__new__ is object.__new__
True  # was previously False
```

[...]

PR #2760, opened 2021-07-29
Use inspect.getmodule to prevent redundant output in whats_left.sh

jamestwebber:
As I (very slowly) progress toward making an actual contribution, I wanted to propose a tweak to `whats_left.sh`. In its current form, the script uses `dir` to crawl the modules and this means that there is some redundant output from items defined somewhere else. A big culprit is `__loader__` (see #2732) but there are others. In total this shaves 388 (redundant) lines from the output. The `missing_items` and `mismatched_items` go down but I think that was because they overcounted before.

Another thing I noticed in the output is that for some entries, RustPython is actually _better_ than the CPython implementation. Many of the entries in `operator` have a signature, where the CPython version returns `None`. I'm not sure it makes sense to include those in the output but that's a separate change.

### DIFF
--- a/extra_tests/not_impl_gen.py
+++ b/extra_tests/not_impl_gen.py
@@ -187,7 +187,10 @@ def dir_of_mod_or_error(module_name):
     result = {}
     for item_name in item_names:
         item = getattr(module, item_name)
-        result[item_name] = extra_info(item)
+        item_mod = inspect.getmodule(item)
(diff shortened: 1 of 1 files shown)


youknowone (approved):
Thanks! now it will look even better.

youknowone:
Do you have any suggestion about the signature check?

PR #2763, opened 2021-07-29
Fixes `type.__module__` issue, refs #2310

### LINKED ISSUE #2310: Builtin function's class object return wrong value for attribute __module__ __doc__
While I am working on the pickle for iterator object, I went into the unexpected throw from pickle.py. After reading the source I find out that because it need the module name from the pickled object's class but __module__ return a getset_descriptor instead a str.

Put it simple, the following code cause the issue:

```
import builtins
builtins.iter.__class__.__module__
builtins.iter.__class__.__doc__
```

sobolevn:
Previously `type.__module__` was returning `<getset descriptor>`.
It happend because this code:

```rust
#[pyproperty(magic)]
    fn module(&self, vm: &VirtualMachine) -> PyObjectRef {
        // TODO: Implement getting the actual module a builtin type is from
        self.attributes
            .read()
            .get("__module__")
            .cloned()
            .unwrap_or_else(|| vm.ctx.new_str("builtins"))
    }
```

was able to find `__module__` in `self.attributes`. But, it was the method `#[pyproperty(magic)] fn module` definition itself.
The same happened to `__qualname__` as well.

I've added a simple test to ensure that it is now correct.

Closes #2310

### DIFF
--- a/extra_tests/snippets/builtin_type.py
+++ b/extra_tests/snippets/builtin_type.py
@@ -3,5 +3,7 @@ assert type.__qualname__ == 'type'
 assert type.__name__ == 'type'
 assert isinstance(type.__doc__, str)
 
+# Regression to
+# https://github.com/RustPython/RustPython/issues/2310
 import builtins
 assert builtins.iter.__class__.__module__ == 'builtins'
--- a/extra_tests/snippets/builtin_type.py
+++ b/extra_tests/snippets/builtin_type.py
@@ -23,3 +23,11 @@ assert B.__qualname__ == 'BB'
 # https://github.com/RustPython/RustPython/issues/2310
 import builtins
 assert builtins.iter.__class__.__module__ == 'builtins'
+assert builtins.iter.__class__.__qualname__ == 'builtin_function_or_method'
+
+try:
+    builtins.iter.__class__.__docs__
+except AttributeError:
+    assert True
+else:
+    assert False, 'Attribute error was not raised'
--- a/extra_tests/snippets/builtin_type.py
+++ b/extra_tests/snippets/builtin_type.py
@@ -3,6 +3,22 @@ assert type.__qualname__ == 'type'
 assert type.__name__ == 'type'
 assert isinstance(type.__doc__, str)
 
+
+class A(type):
+    pass
+
+
+class B(type):
+    __module__ = 'b'
+    __qualname__ = 'BB'
+
+
+assert A.__module__ == '__main__'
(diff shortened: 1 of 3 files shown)


sobolevn:
Ok, turns out `type` and `builtins.iter.__class__` is not the same thing 😒

sobolevn:
Any ideas why this happens?

```
  File "/Users/runner/work/RustPython/RustPython/extra_tests/snippets/bytearray.py", line 728, in <module>
    assert pickle.dumps(a, 0) == b'c__builtin__\nbytearray\np0\n(c_codecs\nencode\np1\n(V\xffab\x80\\u0000\\u0000\xf8\\u0000\\u0000\np2\nVlatin1\np3\ntp4\nRp5\ntp6\nRp7\n.'
  File "/Users/runner/work/RustPython/RustPython/vm/pylib-crate/Lib/pickle.py", line 1744, in _dumps
    buffer_callback=buffer_callback).dump(obj)
  File "/Users/runner/work/RustPython/RustPython/vm/pylib-crate/Lib/pickle.py", line 487, in dump
    self.save(obj)
  File "/Users/runner/work/RustPython/RustPython/vm/pylib-crate/Lib/pickle.py", line 560, in save
    f(self, obj)  # Call unbound method with explicit self
  File "/Users/runner/work/RustPython/RustPython/vm/pylib-crate/Lib/pickle.py", line 814, in save_bytearray
    self.save_reduce(bytearray, (bytes(obj),), obj=obj)
  File "/Users/runner/work/RustPython/RustPython/vm/pylib-crate/Lib/pickle.py", line 691, in save_reduce
    save(func)
  File "/Users/runner/work/RustPython/RustPython/vm/pylib-crate/Lib/pickle.py", line 560, in save
    f(self, obj)  # Call unbound method with explicit self
  File "/Users/runner/work/RustPython/RustPython/vm/pylib-crate/Lib/pickle.py", line 1128, in save_type
    return self.save_global(obj)
  File "/Users/runner/work/RustPython/RustPython/vm/pylib-crate/Lib/pickle.py", line 1077, in save_global
    (obj, module_name, name))
PicklingError: Can't pickle <class 'bytearray'>: it's not the same object as builtins.type
=============================== warnings summary ===============================
```

sobolevn:
Local tests pass! 🎉 

The solution looks a bit ugly to me, but it is the easiest one to stop the recursion.

DimitrisJim (approved):
This does look like the best approach to me.

PR #2770, opened 2021-07-31
Adds `__qualname__` to `PyBuiltinMethod`

### LINKED ISSUE #2767: `method_descriptor` must have `__qualname__` attribute
## Feature

`CPython`:

```pycon
>>> str.replace
<method 'replace' of 'str' objects>

>>> str.replace.__qualname__
'str.replace'
```

`RustPython`:

```pycon
>>>>> str.replace
<method_descriptor object at 0x7fd669446a30>

>>>>> str.replace.__qualname__
AttributeError: 'method_descriptor' object has no attribute '__qualname__'
```

Related #2310

## Python Documentation

https://www.python.org/dev/peps/pep-3155/

sobolevn:
Closes #2767


vm/src/builtins/builtinfunc.rs
```diff
@@ -140,9 +139,23 @@ impl PyBuiltinFunction {
     }
 }
 
+// PyMethodDescrObject in
+// https://github.com/python/cpython/blob/main/Objects/descrobject.c
 #[pyclass(module = false, name = "method_descriptor")]
 pub struct PyBuiltinMethod {
     value: PyNativeFuncDef,
+    dtype: Option<PyTypeRef>,
```
sobolevn, line 147: I can make this `PyObjectRef` and store just the type name. 
But, the current implementation is closer to https://github.com/python/cpython/blob/main/Objects/descrobject.c#L43

sobolevn: And https://github.com/python/cpython/blob/302cf3550514b174b51b8c24d6e16cb47dc61755/Include/descrobject.h#L43

DimitrisJim: I think its good right now. Staying closer to CPython is generally preferred.

sobolevn:
I have changed `.new_method` signature. It compiles locally, but `wasm` is also using it. I am not quite familiar with `wasm`, so I have several questions:

1. This usage https://github.com/RustPython/RustPython/blob/4890f09c576fc2bb0654c67eecc1ca52a0353c06/wasm/lib/src/convert.rs#L219 looks like it should be changed to `ctx.api.new_function()`
2. Here we can create `py_class!` first, then use `extend_class!` to set methods: https://github.com/RustPython/RustPython/blob/4890f09c576fc2bb0654c67eecc1ca52a0353c06/wasm/lib/src/wasm_builtins.rs#L36

sobolevn:
How do I test the wasm build?

derive/src/pyclass.rs
```diff
@@ -317,18 +317,25 @@ where
...
 
         let py_name = item_meta.method_name()?;
-        let build_func = Ident::new(&format!("build_{}", &self.method_type), args.item.span());
         let tokens = {
             let doc = args.attrs.doc().map_or_else(
                 TokenStream::new,
                 |doc| quote!(.with_doc(#doc.to_owned(), ctx)),
             );
+            let build_func = match self.method_type.as_str() {
+                "method" => quote!(.build_method(ctx, class)),
```
sobolevn, line 326: My main motivation for this change was the fact that "Find usages" and `grep` failed to find this `.build_method` call site.

Which resulted in 1000s of errors in the project. It was quite hard to find the reason 🙂

vm/src/builtins/builtinfunc.rs
youknowone: I think the corresponding object for `PyBuiltinMethod` in CPython is `PyCMethodObject`.
https://github.com/python/cpython/blob/302cf3550514b174b51b8c24d6e16cb47dc61755/Include/cpython/methodobject.h#L32-L35

And I found a field named `mm_class` with same type. is this related? or is this different field?

vm/src/builtins/builtinfunc.rs
```diff
@@ -140,9 +139,26 @@ impl PyBuiltinFunction {
...
+// and
+// https://github.com/python/cpython/blob/main/Objects/descrobject.c
 #[pyclass(module = false, name = "method_descriptor")]
 pub struct PyBuiltinMethod {
     value: PyNativeFuncDef,
+    dtype: Option<PyTypeRef>,
+}
+
+impl PyBuiltinMethod {
+    pub fn with_dtype(mut self, dtype: PyTypeRef) -> Self {
```
youknowone, line 153: `pub(crate)` not to expose to users.

But, if all methods finally have `d_type` before fully initialized, I suggest to not to create this method but adding the parameter to each creation methods not to let user failure.

vm/src/builtins/builtinfunc.rs
sobolevn: Yeah, I think that it is more like `PyCMethodObject`, renamed `dtype` to `mmtype` to match the sctructure. 👍

vm/src/builtins/builtinfunc.rs
sobolevn: Yeap, totally! Now `mmtype` is just `PyTypeRef`

youknowone:
`cd wasm/lib && cargo build` will build wasm

vm/src/builtins/builtinfunc.rs
```diff
@@ -140,9 +143,12 @@ impl PyBuiltinFunction {
     }
 }
 
+// PyCMethodObject in
+// https://github.com/python/cpython/blob/main/Include/cpython/methodobject.h
 #[pyclass(module = false, name = "method_descriptor")]
 pub struct PyBuiltinMethod {
     value: PyNativeFuncDef,
+    mmtype: PyTypeRef,
```
youknowone, line 151: I suggest to use just `class` because this field is saving the class owning the method. we don't use those prefixes in RustPython.

sobolevn: 👍

vm/src/builtins/builtinfunc.rs
```diff
@@ -38,17 +38,20 @@ impl PyNativeFuncDef {
     pub fn build_function(self, ctx: &PyContext) -> PyObjectRef {
         self.into_function().build(ctx)
     }
-    pub fn build_method(self, ctx: &PyContext) -> PyObjectRef {
+    pub fn build_method(self, ctx: &PyContext, class: &PyTypeRef) -> PyObjectRef {
```
youknowone, line 41: `class: PyTypeRef`

when `class` is used by only clone, it doesn't need to be a reference.
(and also for callers)

youknowone (changes requested):
almost done!

youknowone (approved):

PR #2772, opened 2021-07-31
Fix redox and other stuff

coolreader18:
- Upgrade dependencies for redox
- Fix no fstatat/times on redox
- Fix building with openssl 1.1.0
- Remove derive_more dep
- Run cargo update
- Fix socket2::Socket niche optimization breaking stuff


Cargo.toml
```diff
@@ -46,7 +43,7 @@ flame = { version = "0.2", optional = true }
 flamescope = { version = "0.1", optional = true }
 
 [target.'cfg(not(target_os = "wasi"))'.dependencies]
-rustyline = "8.0"
+rustyline = "8.1"
 
 [dev-dependencies]
 cpython = "0.5.0"
```
DimitrisJim, line 49: Any chance we can get `cpython` and `python3-sys` bumped to `0.6`? I've been running into some issues running the benchmarks and I've only been getting it to work after bumping these to `0.6` and pointing it to a version of Python build with `GCC > 10`. (Otherwise, it picks up an anaconda version of Python that has been compiled with a gcc version < 10).

youknowone (approved):

PR #2774, opened 2021-08-01
Fixes clippy on macos

sobolevn:
For some reason I was having an error from `clippy` locally:
Снимок экрана 2021-08-01 в 12 05 46

It was quite annoying, so I have fixed it:
Снимок экрана 2021-08-01 в 12 07 14


youknowone (approved):

youknowone:
thank you!

PR #2784, opened 2021-08-02
added test_deque.py from CPython 3.8.7

jungsu-kwon:
hello I have added test_deque.py from CPython 3.8.7 

if there is any issue plz let me know

I will fix it asap

thanks in prior


DimitrisJim (approved):
Looks good. Thanks!

PR #2787, opened 2021-08-02
Add test_yield_from from CPython 3.8

fanninpm:


fanninpm:
@youknowone @DimitrisJim I can't reproduce this on my machine. Are you able to reproduce this locally?

DimitrisJim:
Nope I can't. I've re-run the jobs to see if they fail again.

youknowone:
same here, i can't. how about marking it as skip for now?

DimitrisJim (approved):

PR #2791, opened 2021-08-04
Adds `__qualname__` to `PyBoundMethod`

### LINKED ISSUE #2786: `.__new__.__qualname__` fails for some types
## Feature

CPython:

```python
>>> type.__new__.__qualname__
'type.__new__'

>>> ImportError.__new__.__qualname__
'Exception.__new__'
```

RustPython:

```python
>>>>> ImportError.__new__.__qualname__
AttributeError: 'builtin_function_or_method' object has no attribute '__qualname__'
>>>>> ImportError.__new__
<bound method __new__ of <class 'ImportError'>>

>>>>> type.__new__.__qualname__
AttributeError: 'builtin_function_or_method' object has no attribute '__qualname__'
>>>>> type.__new__
<bound method __new__ of <class 'type'>>
```

sobolevn:
Closes #2786

I am marking this PR as WIP, because it doesn't quite work right now.
The problem is with `__new__` bound method.

Right now:

```python
>>> type.__new__.__qualname__`
'__new__'
```

Should be:

```python
>>> type.__new__.__qualname__
'type.__new__'
```

Right now, `__new__` is not a method, it is an instance of `PyBuiltInFunction`. 
And that's why its `__qualname__` is just `__new__`.

I have several options:
1. Make this a special case in `PyBoundMethod`, looks quite ugly to me, but is simple.
2. Change `tp_new_wrapper` to be an instance of `PyBuiltinMethod` with `class: &ctx.type_type`, modify `class` field when `slot` is updated here: https://github.com/RustPython/RustPython/blob/62af4222ba7bbe9a520d9aa242d2d8af45874b08/vm/src/pyobject.rs#L383 this looks right to me, but can change the interpreter internals in a way I cannot predict.
3. Possibly some others I haven't consider

What should I do? 🙂


DimitrisJim:
Without having looked into it, have you peeked at how CPython goes about doing this?

sobolevn:
Looks like `__new__` is `PyMethodDef` (2): https://github.com/python/cpython/blob/cee67fa66129b5d1db5c8aa3884338f82f0da3de/Objects/typeobject.c#L7047-L7053

sobolevn:
Ok, I am using the first solution. Tests are passing 🎉

vm/src/builtins/function.rs
```diff
@@ -482,6 +482,28 @@ impl PyBoundMethod {
...
+            // Special case: we work with `__new__`, which is not really a method.
+            // It is a function, so its `__qualname__` is just `__new__`.
+            // We need to add object's part manually.
+            // Note: at the moment, `__new__` in the form of `tp_new_wrapper`
+            // is the only instance of `builtin_function_or_method_type`.
+            let obj_name = vm.get_attribute_opt(self.object.clone(), "__qualname__")?;
+            let obj_name: Option<PyStrRef> = obj_name.and_then(|o| o.downcast().ok());
+            return Ok(vm.ctx.new_str(format!(
+                "{}.__new__",
+                obj_name.as_ref().map_or("?", |s| s.as_str())
```
sobolevn, line 501: I was not sure about this `?` feature, but I've copied the `repr` implementation.
Maybe we should `match obj_name` and `return Err` on `None`?

youknowone: How CPython handle it? I didn't looked up the history but I guess that happened just to avoid errors at that time due to lack of implementations.

sobolevn: In CPython it is just passed to `self.func`: https://github.com/python/cpython/blob/ddf8ae31a0f371eff2db14c7f7a45976b86d56ea/Objects/classobject.c#L201

`repr` is also quite different: https://github.com/python/cpython/blob/ddf8ae31a0f371eff2db14c7f7a45976b86d56ea/Objects/classobject.c#L274-L300 It also has no `?`

youknowone: I can see `?` here:
https://github.com/python/cpython/blob/ddf8ae31a0f371eff2db14c7f7a45976b86d56ea/Objects/classobject.c#L280

sobolevn: Yes, indeed! I've missed that. 

Anyway, let's leave `repr` out of scope (as it is for this task). 
What should we do with `__qualname__`? 🙂 

I see several things:
1. Leave it as is, until someone report a case where this code is wrong
2. Switch `map_or` to `unwrap()`, because all object must have this property, this can cause bugs if some objects accidentally don't have it
3. ???

youknowone: I think this branch itself is not the CPython way anyway. So any approach looks acceptable. When we handle `__new__` correctly, the whole branch will be gone. isn't it?

sobolevn: Yes, it only executes for `__new__` on several builtin types, others are just proxied to `vm.get_attribute(self.function.clone(), "__qualname__")` like CPython does. I can add a `TODO` for future to fix this special case on a lower-level.

youknowone (approved):

sobolevn:
Rebased 👍

Issue #1782, opened 2020-02-28
Built in classes receive args from subclass despite empty __init__()


philippeitis:
This is related to #1672 - Counter inherits from dict and calls its base constructor, which receives all arguments, despite them not being passed through the `super(...)__init__()` constructor.

I've included two examples below:

```
class ListSubClass(list):
    def __init__(self, *args):
        super(ListSubClass, self).__init__()
        print(args)

x = ListSubClass(1)
```
Expected:
```
(1, )
```
Got:
```
TypeError: Cannot iterate over int
```

Additionally, for dict:
```
class DictSubClass(dict):
    def __init__(self, *args):
        super(DictSubClass, self).__init__()
        print(args)

x = DictSubClass("1")
```
Expected
```
('1', )
```
Got
```
TypeError: Iterator must have exactly two elements
```

I was not able to reproduce this for classes defined within Python itself, but was able to reproduce this with sets as well. However, this does not apply for the tuple type, as it also raises a TypeError in CPython.

This goes away when we override `__new__` to pass no arguments or keyword arguments, but is not required in CPython - eg:
```
def __new__(cls, *args, **kwargs):
      super(ClassName, cls).__new__(cls)
```

youknowone:
closed by #2792

PR #2792, opened 2021-08-04
Initialize with init for mutable builtins.

DimitrisJim:
This moves the initialization from the arguments into `__init__`. `dict` already did this, this just fixes the behavior for `list` and `set`.


youknowone (approved):

youknowone:
Fixed #1782

DimitrisJim:
There's still a couple more (`bytearray` as a builtin and `deque` and maybe others in `collections`) and I'll get to those in a later PR.

PR #2795, opened 2021-08-05
Adds reduce, length_hint and set_state to range iterators.

DimitrisJim:


vm/src/builtins/range.rs
```diff
@@ -481,10 +483,40 @@ impl PyValue for PyLongRangeIterator {
 }
 
 #[pyimpl(with(PyIter))]
-impl PyLongRangeIterator {}
+impl PyLongRangeIterator {
+    #[pymethod(name = "__length_hint__")]
```
youknowone, line 487: `#[pymethod(magic)]` will work.
`magic` adds dunder to rust function name.

vm/src/builtins/range.rs
```diff
@@ -1,3 +1,5 @@
+use std::cmp::max;
```
youknowone, line 1: The convention is mixed for now, but let's not make separated section of `use`. `rustfmt` will sort it.

vm/src/builtins/range.rs
```diff
@@ -514,20 +546,85 @@ impl PyValue for PyRangeIterator {
...
+        let index = zelf.index.fetch_add(1);
+        if index < zelf.length {
+            Ok(vm.ctx.new_int(zelf.start + (index as isize) * zelf.step))
         } else {
             Err(vm.new_stop_iteration())
         }
     }
 }
 
+fn iterrange_reduce(
```
youknowone, line 593:
```suggestion
fn range_iterator_reduce(
```

vm/src/builtins/range.rs
```diff
@@ -514,20 +546,85 @@ impl PyValue for PyRangeIterator {
...
+    };
+    Ok(vm.ctx.new_tuple(vec![
+        iter,
+        vm.ctx.new_tuple(vec![range.into_object(vm)]),
+        vm.ctx.new_int(index),
+    ]))
+}
+
+// Silently clips state (i.e index) in range [0, usize::MAX].
+fn get_state(length: &BigInt, state: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
```
youknowone, line 615: In rust, I think `get` usually means getter with small cost.

how about `make_state` or `range_state`?

vm/src/builtins/range.rs
DimitrisJim: Yeah, missed this (rest of the range class uses `name` for the dunder methods and it threw me off).

youknowone (approved):

PR #2798, opened 2021-08-07
socket: impl herror

deantvv:
There is one todo that is waiting for hstrerror to land in libc crate.
For now it shows the same error as cpython when hstrerror is not
available.


DimitrisJim (approved):
This looks good to me, I see you opened https://github.com/rust-lang/libc/pull/2323/ so I'm hoping you'll be able to fix the TODO when that lands.

PR #2803, opened 2021-08-07
Fix incorrect unwrapping

### LINKED ISSUE #2799: panic occurs when a key in the dictionary is unhashable.
## Feature
In cpython, if the key in the dictionary is unhashable, only an error is raised.
```python
>>> {{}:2}
TypeError: unhashable type: 'dict'
```
But in rustpython, a panic is occurring.
```python
>>>>> {{}:2}
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: [PyObj PyBaseException]', vm/src/frame.rs:1314:50
```

moreal:
This pull request resolves #2799.

Because the implementation of [`ItemProtocol<T>.set_item(…)` for `PyDictRef`][itemprotocol-setitem-for-pydictref] can return `Err` having `TypeError` when `key` implemented [`Unhashable`][unhashable], it should not call `unwrap`. So, this pull request removed calling `unwrap` and used `?` operator.

[itemprotocol-setitem-for-pydictref]: https://github.com/RustPython/RustPython/blob/53b8067acd1af79c2081e9eb56586fa8955749fd/vm/src/builtins/dict.rs#L527-L534
[unhashable]: https://github.com/RustPython/RustPython/blob/53b8067acd1af79c2081e9eb56586fa8955749fd/vm/src/slots.rs#L219-L228

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -1311,7 +1311,7 @@ impl ExecutingFrame<'_> {
             }
         } else {
             for (key, value) in self.pop_multiple(2 * size).tuples() {
-                map_obj.set_item(key, value, vm).unwrap();
+                map_obj.set_item(key, value, vm)?;
             }
         }
 


DimitrisJim:
This is good. I would request you also fix the unwrapping done on line 1309 of the same file and function in the same way. (see [here](https://github.com/RustPython/RustPython/blob/cc7269c68d7d888b24372541524a416f88133bc5/vm/src/frame.rs#L1309)). 

The issue is basically the same, it is currently unable to be triggered since we don't accept dict subclasses when unpacking is performed (but at some point in the future we will and that error would pop-up then).

youknowone (approved):

youknowone:
@DimitrisJim Though your comment is appropriate, I will merge this PR as current state. Because that's a separated and independent bug caused by different code, which is not belong to this PR. Instead of it, I will paste the link of your code to the original issue.

DimitrisJim:
Yup @youknowone, no prob with that.

PR #2805, opened 2021-08-08
Add math.comb function

voidsatisfaction:
I implemented `math.comb` function.

As I am new to Rust, if there is any other suggestions, please give me some advice :)

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -533,8 +533,8 @@ fn math_comb(n: PyIntRef, k: PyIntRef, vm: &VirtualMachine) -> PyResult<BigInt> 
         return Ok(one);
     }
 
-    let mut result = n - 0;
-    let mut factor = n - 0;
+    let mut result = n.clone();
+    let mut factor = n.clone();
     let mut current = one;
     while current < *k {
         factor -= 1;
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/math.rs
```diff
@@ -510,6 +510,43 @@ fn math_factorial(value: PyIntRef, vm: &VirtualMachine) -> PyResult<BigInt> {
...
+    let n = n.as_bigint();
+    let one = BigInt::one();
+    let zero = BigInt::zero();
+
+    if n.is_negative() || k.is_negative() {
+        return Err(vm.new_value_error("comb() not defined for negative values".to_owned()))
+    }
+
+    let temp = n - k;
+    if temp < zero {
```
youknowone, line 524:
```suggestion
    if temp.is_negative() {
```

vm/src/stdlib/math.rs
```diff
@@ -510,6 +510,43 @@ fn math_factorial(value: PyIntRef, vm: &VirtualMachine) -> PyResult<BigInt> {
...
+        return Ok(zero);
+    }
+
+    if temp < *k {
+        k = &temp
+    }
+
+    if *k == zero {
```
youknowone, line 532:
```suggestion
    if k.is_zero() {
```

youknowone (changes requested):

vm/src/stdlib/math.rs
voidsatisfaction: I fixed it. Thanks!

vm/src/stdlib/math.rs
voidsatisfaction: I fixed it. Thanks!

youknowone (approved):

PR #2807, opened 2021-08-08
Add __index__ operator overriding class support for RangeIndex

Snowapril:
This revision fix **RangeTestCase.test_range** test in [test_index.py](https://github.com/RustPython/RustPython/blob/master/Lib/test/test_index.py#L248)

`TryFromObject` implementation for `RangeIndex` enum did not cover classes
that override `__index__` operator which cpython implementation is currently providing
(https://docs.python.org/3.8/library/operator.html#operator.__index__)

I add codes that cover the above case using `vm.to_index` method.

##### Before
```rust
impl TryFromObject for RangeIndex {
        match_class!(match obj {
            i @ PyInt => Ok(RangeIndex::Int(i)),
            s @ PySlice => Ok(RangeIndex::Slice(s)),
            obj => Err(vm.new_type_error(format!(
                "sequence indices be integers or slices, not '{}'",
                obj.class().name,
            ))),
        })
    }
}
```

##### After
```rust
impl TryFromObject for RangeIndex {
        match_class!(match obj {
            i @ PyInt => Ok(RangeIndex::Int(i)),
            s @ PySlice => Ok(RangeIndex::Slice(s)),
            obj => match vm.to_index(&obj) {
                Ok(val) => Ok(RangeIndex::Int(val)),
                Err(_) => Err(vm.new_type_error(format!(
                    "sequence indices be integers or slices or classes that override __index__ operator, not '{}'",
                    obj.class().name
                ))),
            },
        })
    }
}
```

### DIFF
--- a/vm/src/builtins/range.rs
+++ b/vm/src/builtins/range.rs
@@ -634,10 +634,13 @@ impl TryFromObject for RangeIndex {
         match_class!(match obj {
             i @ PyInt => Ok(RangeIndex::Int(i)),
             s @ PySlice => Ok(RangeIndex::Slice(s)),
-            obj => Err(vm.new_type_error(format!(
-                "sequence indices be integers or slices, not '{}'",
-                obj.class().name,
-            ))),
+            obj => {
+                let val = vm.to_index(&obj).map_err(|_| vm.new_type_error(format!(
+                    "sequence indices be integers or slices or classes that override __index__ operator, not '{}'",
+                    obj.class().name
+                )))?;
+                Ok(RangeIndex::Int(val))
+            }
         })
     }
 }
--- a/Lib/test/test_index.py
+++ b/Lib/test/test_index.py
@@ -243,8 +243,6 @@ class NewSeqTestCase(SeqTestCase, unittest.TestCase):
 
 class RangeTestCase(unittest.TestCase):
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_range(self):
         n = newstyle()
         n.ind = 5


vm/src/builtins/range.rs
```diff
@@ -634,10 +634,14 @@ impl TryFromObject for RangeIndex {
...
+                Ok(val) => Ok(RangeIndex::Int(val)),
+                Err(_) => Err(vm.new_type_error(format!(
+                    "sequence indices be integers or slices or classes that override __index__ operator, not '{}'",
+                    obj.class().name
+                ))),
+            },
         })
     }
 }
+
```
youknowone, line 647: this line look like unnessessary change

vm/src/builtins/range.rs
```diff
@@ -634,10 +634,14 @@ impl TryFromObject for RangeIndex {
...
-                "sequence indices be integers or slices, not '{}'",
-                obj.class().name,
-            ))),
+            obj => match vm.to_index(&obj) {
+                Ok(val) => Ok(RangeIndex::Int(val)),
+                Err(_) => Err(vm.new_type_error(format!(
+                    "sequence indices be integers or slices or classes that override __index__ operator, not '{}'",
+                    obj.class().name
+                ))),
+            },
```
youknowone, line 643:
```suggestion
            obj => {
                let val = vm.to_index(&obj).map_err(|_| vm.new_type_error(format!(
                    "sequence indices be integers or slices or classes that override __index__ operator, not '{}'",
                    obj.class().name
                )));
                RangeIndex::Int(val)
            }
```

Your version also looks very good, but we also conventionally write it like this to show the success flow better.

youknowone (approved):

vm/src/builtins/range.rs
youknowone: this line also causes lint error: https://github.com/RustPython/RustPython/pull/2807/checks?check_run_id=3273128526

Snowapril: That line was added by rustfmt command but I also think it is unnecessary. I'll pick it out with below conversation.

vm/src/builtins/range.rs
Snowapril: Oh, you are right. it is better seems great. But in my environment, rust complained about type mismatching like below.

>    Compiling rustpython-vm v0.1.2 (/home/snowapril/RustPython/vm)
> error[E0308]: mismatched types
>    --> vm/src/builtins/range.rs:642:33
>     |
> 642 |                 RangeIndex::Int(val)
>     |                                 ^^^ expected struct `pyobjectrc::PyRef`, found enum `Result`
>     |
>     = note: expected struct `pyobjectrc::PyRef<_>`
>                  found enum `Result<pyobjectrc::PyRef<_>, pyobjectrc::PyRef<PyBaseException>>`
> 
> error[E0308]: mismatched types
>    --> vm/src/builtins/range.rs:642:17
>     |
> 633 |     fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
>     |                                                                  -------------- expected `Result<RangeIndex, pyobjectrc::PyRef<PyBaseException>>` because of return type
> ...
> 642 |                 RangeIndex::Int(val)
>     |                 ^^^^^^^^^^^^^^^^^^^^
>     |                 |
>     |                 expected enum `Result`, found enum `RangeIndex`
>     |                 help: try using a variant of the expected enum: `serde::__private::Ok(RangeIndex::Int(val))`
>     |
>     = note: expected enum `Result<RangeIndex, pyobjectrc::PyRef<PyBaseException>>`
>                found enum `RangeIndex`
> 
> error: aborting due to 2 previous errors
> 
> For more information about this error, try `rustc --explain E0308`.
> error: could not compile `rustpython-vm`
> 
> To learn more, run the command again with --verbose.

I pushed the second version of the code below. could you review it one more time? thanks.

vm/src/builtins/range.rs
Snowapril: I'm sorry. I misunderstood that change. I resend v3 code and lint seems good now.

youknowone:
thank you for contributing!

PR #2808, opened 2021-08-08
Adds `class` field to `PyGetSet` to fix `__qualname__`

### LINKED ISSUE #2790: `getset_descriptor` object has no attribute `__qualname__`
## Feature
CPython:

```python
>>> Exception.args.__qualname__
'BaseException.args'
```

RustPython:

```python
>>>>> Exception.args.__qualname__
AttributeError: 'getset_descriptor' object has no attribute '__qualname__'
```

sobolevn:
In the next PR I will also fix the `repr` of `getset` as well.

Changes:
- New `class` field on `PyGetSet`
- New arguments to `new_getset` and `new_readonly_getset`
- Fixed how `derive` is constructing types
- Fixes all `.new_getset` and `.new_readonly_getset` calls
- Also fixes recent `class` argument order in `new_method` from #2770 to be consistent: name, class, functions

CPython impl: https://github.com/python/cpython/blob/635bfe8162981332b36cc556bac78e869af579c2/Objects/descrobject.c#L572-L599

Closes #2790

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -563,7 +563,7 @@ impl ToTokens for GetSetNursery {
                     class.set_str_attr(
                         #name,
                         ::rustpython_vm::PyObject::new(
-                            ::rustpython_vm::builtins::PyGetSet::new(#name.into())
+                            ::rustpython_vm::builtins::PyGetSet::new(#name.into(), class.clone())
                                 .with_get(&Self::#getter)
                                 #setter #deleter,
                             ctx.types.getset_type.clone(), None)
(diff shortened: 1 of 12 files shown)


youknowone:
Thanks for the series of qualname implementations.
CPython seems having `PyDescrObject d_common` field for common descriptor fields. After reviewing the series, I feel like RustPython also need to have that kind of common structure for repeating features and field. How do you think about it?
Probably we also need to seperate PyGetSet and PyGetSetDescrObject in future. If we don't go that far at this moment, having a common descriptor object will be a first step for the future.

sobolevn:
> After reviewing the series, I feel like RustPython also need to have that kind of common structure for repeating features and field. How do you think about it?

Yes, I think so too. 

> Probably we also need to seperate PyGetSet and PyGetSetDescrObject in future

I would be happy to work on that soon 🙂

youknowone (approved):

PR #2814, opened 2021-08-09
implement Unhashable for collections.deque

### LINKED ISSUE #2813: Deque instances shouldn't be hashable
Currently, `deque` instances are able to be hashed:

```
>>>>> hash(deque())
94434894504112
```

Expected behavior:

```
>>>>> hash(list())
TypeError: unhashable type: 'deque'
```

jamestwebber:
Closes #2813.

I am assuming that the way this trait is structured (the struct must implement `Hashable`, which is then overruled by the implementation of `Unhashable`) is required by how Rust checks for trait implementations? Like the hash function wouldn't work otherwise.

### DIFF
--- a/Lib/test/test_deque.py
+++ b/Lib/test/test_deque.py
@@ -590,8 +590,6 @@ class TestBasic(unittest.TestCase):
         self.assertRaises(TypeError, deque, 'abc', 2, 3);
         self.assertRaises(TypeError, deque, 1);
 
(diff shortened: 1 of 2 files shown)


DimitrisJim:
Yup! Looks good to me, test passes but you forgot to remove the `unittest.expectedFailure` from `test_deque.test_hash`, as a result it fails with the unexpected success.

DimitrisJim (approved):

PR #2817, opened 2021-08-09
"testFactorialNonIntegers" should pass the test

leesungbin:
I think it is not "TODO".
Since python 3.9, accepting floats with integral values (like 5.0) is deprecated. [Link](https://docs.python.org/3/library/math.html#math.factorial)


youknowone (changes requested):
Thank you for contributing.
I am sorry but we don't update CPython unittest specifically one by one.

By looking at history of `Lib/test/test_math.py`, the file is based on CPython 3.6.x.
If you want to update the test file to newer version, please copy a new version test from newer CPython version, rather than updating specific function manually.

PR #2818, opened 2021-08-09
Add __index__ operator overriding class support for SequenceIndex

Snowapril:
This revision fix three test cases in [test_index.py](https://github.com/RustPython/RustPython/blob/master/Lib/test/test_index.py)
* `SeqTestCase.test_index`
* `SeqTestCase.test_wrappers`
* `ListTestCase.test_setdelitem`

All these three cases' failures were caused by not supporting classes
that override `__index__` operator when indexing sequence type.

This commit implement almost same routine with PR(#2807) which also did not
support classes that override `__index__` operator in range indexing.

### DIFF
--- a/Lib/test/test_index.py
+++ b/Lib/test/test_index.py
@@ -101,8 +101,6 @@ class SeqTestCase:
         self.o2 = newstyle()
         self.n2 = newstyle()
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_index(self):
         self.o.ind = -2
         self.n.ind = 2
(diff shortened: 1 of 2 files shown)


Snowapril:
@youknowone Below tests make current changes unacceptable. Is it better to fix error message `"{} indices be integers or slices or classes that override __index__ operator, not '{}'"` to `"{} indices must be integers or slices, not {}"`?

##### test_tuple.py
```python
    def test_getitem_error(self):
        t = ()
        msg = "tuple indices must be integers or slices"
        with self.assertRaisesRegex(TypeError, msg):
            t['a']
```

##### list_tests.py
```python
    def test_getitem_error(self):
        a = []
        msg = "list indices must be integers or slices"
        with self.assertRaisesRegex(TypeError, msg):
            a['a']
```

youknowone:
The test check whether the actual message include it instead of exact match.
I think you missed `must` in your error message.

Snowapril:
Ah! Thank you. I missed the actual work of `assertRaiseRegex` method. I sent v2 code of this PR.

youknowone (approved):
Thank you for contributing!

youknowone:
@Snowapril I attached one more commit to refactor `SequenceIndex::Int` cases

PR #2819, opened 2021-08-09
Add `math_perm` method

leesungbin:
Added `math_perm` method.

Unfortunately, test only fails when `k` is `None`.

### DIFF
--- a/Lib/test/test_math.py
+++ b/Lib/test/test_math.py
@@ -1814,8 +1814,6 @@ class MathTests(unittest.TestCase):
         self.assertEqual(type(prod([1, decimal.Decimal(2.0), 3, 4, 5, 6])),
                          decimal.Decimal)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def testPerm(self):
         perm = math.perm
         factorial = math.factorial
@@ -1839,7 +1837,7 @@ class MathTests(unittest.TestCase):
         # Test one argument form
         for n in range(20):
             self.assertEqual(perm(n), factorial(n))
-            self.assertEqual(perm(n, None), factorial(n))
+            # self.assertEqual(perm(n, None), factorial(n)) # TODO: RUSTPYTHON
 
         # Raises TypeError if any argument is non-integer or argument count is
         # not 1 or 2
--- a/Lib/test/test_math.py
+++ b/Lib/test/test_math.py
@@ -1837,7 +1837,7 @@ class MathTests(unittest.TestCase):
         # Test one argument form
         for n in range(20):
             self.assertEqual(perm(n), factorial(n))
-            # self.assertEqual(perm(n, None), factorial(n)) # TODO: RUSTPYTHON
+            self.assertEqual(perm(n, None), factorial(n))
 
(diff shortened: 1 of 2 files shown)


Lib/test/test_math.py
```diff
@@ -1839,7 +1837,7 @@ def testPerm(self):
         # Test one argument form
         for n in range(20):
             self.assertEqual(perm(n), factorial(n))
-            self.assertEqual(perm(n, None), factorial(n))
+            # self.assertEqual(perm(n, None), factorial(n)) # TODO: RUSTPYTHON
```
youknowone, line 1840: instead of editing test line by line, keep it back to the original version. we prefer to keep the test files to be detected in automated way rather than looking up the comments.
when the test is marked as `expectedFailure`, we will not miss it if a patch fix the problem. But when the test is commented out like now, CI cannot catch if it is fixed or not.

vm/src/stdlib/math.rs
```diff
@@ -510,6 +510,29 @@ fn math_factorial(value: PyIntRef, vm: &VirtualMachine) -> PyResult<BigInt> {
     Ok(product)
 }
 
+fn math_perm(n: PyIntRef, k: OptionalArg<PyIntRef>, vm: &VirtualMachine) -> PyResult<BigInt> {
```
youknowone, line 513:
```suggestion
fn math_perm(n: PyIntRef, k: OptionalArg<Option<PyIntRef>>, vm: &VirtualMachine) -> PyResult<BigInt> {
```
`Option<PyIntRef>` means `None` or `int`

youknowone (changes requested):

youknowone:
@voidsatisfaction would you review this PR please? you recently contributed to `math.comb`, so I think you fit as reviewer for this PR.

vm/src/stdlib/math.rs
leesungbin: Thanks!!

yeongyumin:
I think it looks good.

But I am very new to Rust, @youknowone you also should review whether it is good or not.

vm/src/stdlib/math.rs
```diff
@@ -510,6 +510,36 @@ fn math_factorial(value: PyIntRef, vm: &VirtualMachine) -> PyResult<BigInt> {
...
+    vm: &VirtualMachine,
+) -> PyResult<BigInt> {
+    let n = n.as_bigint();
+    let v: BigInt = match k {
+        OptionalArg::Missing => n.clone(),
+        OptionalArg::Present(x) => match x {
+            Some(xi) => xi.as_bigint().clone(),
+            None => n.clone(),
+        },
+    };
```
youknowone, line 525:
```suggestion
    let k_val;
    let v = match k.flatten() {
        Some(k) => {
            k_val = k.as_bigint().clone();
            &k_val
        }
        None => n,
    };
```
we can avoid a redundant clone of `k`

youknowone (changes requested):

yeongyumin (approved):

youknowone (approved):

PR #2820, opened 2021-08-09
Fixes infinite recursion on `raise e from e`

### LINKED ISSUE #2779: Raising a user-defined class from itself causes interpreter crashing
In the following example, we define a class error, and instance it. When we execute raise e from e, Rust Python crashes.

test.py
```
class Error(Exception):
	pass

e = Error()
raise e from e

```

Actual output:

> Segmentation fault (core dumped)

System:

> Ubuntu 16.04
> Rust Python 0.1.2 interpreter

sobolevn:
Closes #2779

### DIFF
--- a/extra_tests/snippets/builtin_exceptions.py
+++ b/extra_tests/snippets/builtin_exceptions.py
@@ -1,0 +1,154 @@
+import platform
+import sys
+
+
+# Regression to:
+# https://github.com/RustPython/RustPython/issues/2779
+
+class MyError(Exception):
+    pass
+
+
+e = MyError('message')
+
+try:
+    raise e from e
+except MyError as exc:
+    # It was a segmentation fault before, will print info to stdout:
+    sys.excepthook(type(exc), exc, exc.__traceback__)
+    assert isinstance(exc, MyError)
+    assert exc.__cause__ is e
+    assert exc.__context__ is None
+else:
+    assert False, 'exception not raised'
+
+try:
+    raise ValueError('test') from e
+except ValueError as exc:
+    sys.excepthook(type(exc), exc, exc.__traceback__)  # ok, will print two excs
+    assert isinstance(exc, ValueError)
+    assert exc.__cause__ is e
+    assert exc.__context__ is None
+else:
+    assert False, 'exception not raised'
+
+
+# New case:
+# potential recursion on `__context__` field
+
+e = MyError('message')
+
+try:
+    try:
+        raise e
+    except MyError as exc:
+        raise e
+    else:
+        assert False, 'exception not raised'
+except MyError as exc:
+    sys.excepthook(type(exc), exc, exc.__traceback__)
+    assert exc.__cause__ is None
+    assert exc.__context__ is None
+else:
+    assert False, 'exception not raised'
+
+e = MyError('message')
+
+try:
+    try:
+        raise e
+    except MyError as exc:
+        raise exc
+    else:
+        assert False, 'exception not raised'
+except MyError as exc:
+    sys.excepthook(type(exc), exc, exc.__traceback__)
+    assert exc.__cause__ is None
+    assert exc.__context__ is None
+else:
+    assert False, 'exception not raised'
+
+e = MyError('message')
+
+try:
+    try:
+        raise e
+    except MyError as exc:
+        raise e from e
+    else:
+        assert False, 'exception not raised'
+except MyError as exc:
+    sys.excepthook(type(exc), exc, exc.__traceback__)
+    assert exc.__cause__ is e
+    assert exc.__context__ is None
+else:
+    assert False, 'exception not raised'
+
+e = MyError('message')
+
+try:
+    try:
+        raise e
+    except MyError as exc:
+        raise exc from e
+    else:
+        assert False, 'exception not raised'
+except MyError as exc:
+    sys.excepthook(type(exc), exc, exc.__traceback__)
+    assert exc.__cause__ is e
+    assert exc.__context__ is None
+else:
+    assert False, 'exception not raised'
+
+
+# New case:
+# two exception in a recursion loop
+
+class SubError(MyError):
+    pass
+
+e = MyError('message')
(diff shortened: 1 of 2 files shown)


vm/src/exceptions.rs
```diff
@@ -195,11 +195,16 @@ pub fn write_exception<W: Write>(
...
     if let Some(cause) = exc.cause() {
-        write_exception(output, vm, &cause)?;
-        writeln!(
-            output,
-            "\nThe above exception was the direct cause of the following exception:\n"
-        )?;
+        // This can be a special case: `raise e from e`,
+        // we just ignore it and treat like `raise e`.
+        // So, doing nothing here.
+        if !cause.as_object().is(&exc) {
```
DimitrisJim, line 201: Though this handles `raise e from e` it still segfaults in other situations:

```
class E(Exception): pass
class D(E): pass

e, d = E(), D()
raise d from e
```

CPython uses a set to help stop the infinite recursion, constantly filling it up with exceptions until it sees them all. Something similar can be done here (a new private function that `write_exception` calls seems better as to not require callers of this public function to supply the set).

DimitrisJim (changes requested):

vm/src/exceptions.rs
sobolevn: Good catch! I've refactored this into a `write_exception_recursive` and added several new tests.
Please, take a look 🙂

vm/src/exceptions.rs
```diff
@@ -253,6 +241,39 @@ fn write_traceback_entry<W: Write>(
...
+            write_exception_recursive(output, vm, &cause, seen)?;
+            writeln!(
+                output,
+                "\nThe above exception was the direct cause of the following exception:\n"
+            )?;
+        } else {
+            seen.insert(cause.as_object().get_id());
+        }
+    } else if let Some(context) = exc.context() {
+        write_exception_recursive(output, vm, &context, seen)?;
```
DimitrisJim, line 267: Pretty sure this *should* spin if `e.__context__ = e`. Checking the set again before recursing, as done with `cause`, should do the trick.

DimitrisJim (changes requested):

vm/src/exceptions.rs
sobolevn: Yes, you are right. I was trying to somehow reproduce it, but I didn't think about direct manipulation 🙈 😆 
Fixed!

DimitrisJim: Yeah, Python allows some pretty wild and head-scratching manipulation. I wasn't even sure if a setter was available for `__context__` in RustPython.

extra_tests/snippets/builtin_exceptions.py
```diff
@@ -0,0 +1,151 @@
...
+
+# New case:
+# explicit `__context__` manipulation.
+
+e = MyError('message')
+e.__context__ = e
+
+try:
+    raise e
+except MyError as exc:
```
DimitrisJim, line 146: I think its actually CPython hanging on this (not sure why CPython runs these snippets) probably due to some ciruclar reference clean-up issues? Not really sure. Running this snippet of code with a build of rustpython passes. 

Changing this to `except Exception: pass` should make the check pass.

sobolevn: Yes, I can confirm that `CPython` hangs, while `RustPython` works fine.
I will research if we need to submit this bug to CPython team.

Is there any way to distinguish `RustPython` from `CPython` in runtime?

sobolevn: I see `python_implementation`

sobolevn: @DimitrisJim it is already reported https://bugs.python.org/issue25782

DimitrisJim: Good find. That issue is slightly different in scope as far as I can tell (not really involving printing the exceptions). Rustpython also hangs on some of the cycles they create but that should be tackled in a different issue.

DimitrisJim (approved):

DimitrisJim:
Looks good to me, can you rebase the commits to get a cleaner history?

sobolevn:
Sure!

But, you can press "Squash and merge" to merge a single commit into `master`: https://docs.github.com/en/github/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/about-pull-request-merges#squash-and-merge-your-pull-request-commits This way it won't even re-trigget the CI 👍

vm/src/exceptions.rs
```diff
@@ -253,6 +241,48 @@ fn write_traceback_entry<W: Write>(
...
+        if !seen.contains(&context.as_object().get_id()) {
+            write_exception_recursive(output, vm, &context, seen)?;
+            writeln!(
+                output,
+                "\nDuring handling of the above exception, another exception occurred:\n"
+            )?;
+        } else {
+            seen.insert(context.as_object().get_id());
+        }
+    }
```
youknowone, line 281:
```suggestion
    if let Some((cause_or_context, msg)) = if let Some(cause) = exc.cause() {
        // This can be a special case: `raise e from e`,
        // we just ignore it and treat like `raise e` without any extra steps.
        Some((
            cause,
            "\nThe above exception was the direct cause of the following exception:\n",
        ))
    } else if let Some(context) = exc.context() {
        // This can be a special case:
        //   e = ValueError('e')
        //   e.__context__ = e
        // In this case, we just ignore
        // `__context__` part from going into recursion.
        Some((
            context,
            "\nDuring handling of the above exception, another exception occurred:\n",
        ))
    } else {
        None
    } {
        if !seen.contains(&cause_or_context.as_object().get_id()) {
            write_exception_recursive(output, vm, &cause_or_context, seen)?;
            writeln!(output, "{}", msg)?;
        } else {
            seen.insert(cause_or_context.as_object().get_id());
        }
    }
```

youknowone (changes requested):

youknowone (approved):

PR #2823, opened 2021-08-09
Add missing `__annotations__`  variable into main module

### LINKED ISSUE #2822: Missing `__annotations__` variable
## Feature

Currently, there is no `__annotations__` variable in multiple environment (e.g. interactive, module, expression). Because of it, annotated assignment statements in interactive also doesn't work.

### Interactive

```python
# RustPython
>>>>> a: int = 1
NameError: name '__annotations__' is not defined
>>>>> __annotations__
NameError: name '__annotations__' is not defined

# CPython
>>> a: int = 1
>>> __annotations__
{'a': <class 'int'>}
```

### Execute file

```python
# test.py
print(__annotations__)

# RustPython
  File "test.py", line 1, in <module>
    print(__annotations__)
NameError: name '__annotations__' is not defined

# CPython
{}
```
### Execute expression (a.k.a. `-c` option)
```python
# with <python-binary> -c "print(__annotations__)"

# RustPython
NameError: name '__annotations__' is not defined

# CPython
{}
```

## Python Documentation (also references)

- https://www.python.org/dev/peps/pep-0526/
- https://docs.python.org/3/reference/simple_stmts.html#annotated-assignment-statements
- https://github.com/python/cpython/blob/fa919fdf2583bdfead1df00e842f24f30b2a34bf/Python/pylifecycle.c#L1620-L1625

moreal:
This pull request resolves #2822.

For testing, please see #2822 's testcases. 🙏🏻 

I tried to code like [CPython `add_main_module` does](https://github.com/python/cpython/blob/fa919fdf2583bdfead1df00e842f24f30b2a34bf/Python/pylifecycle.c#L1612-L1658).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -520,6 +520,17 @@ fn write_profile(matches: &ArgMatches) -> Result<(), Box<dyn std::error::Error>>
 fn run_rustpython(vm: &VirtualMachine, matches: &ArgMatches) -> PyResult<()> {
     let scope = vm.new_scope_with_builtins();
     let main_module = vm.new_module("__main__", scope.globals.clone());
+    let ann_dict = vm.ctx.new_dict();
+    let main_module_dict = main_module.dict();
+
+    if main_module_dict.is_none()
+        || main_module_dict
+            .unwrap()
+            .set_item("__annotations__", ann_dict.as_object().to_owned(), vm)
+            .is_err()
+    {
+        error!("Failed to initialize __main__.__annotations__")
+    }
 
     vm.get_attribute(vm.sys_module.clone(), "modules")?
         .set_item("__main__", main_module, vm)?;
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -520,6 +520,17 @@ fn write_profile(matches: &ArgMatches) -> Result<(), Box<dyn std::error::Error>>
...
+    let main_module_dict = main_module.dict();
+
+    if main_module_dict.is_none()
+        || main_module_dict
+            .unwrap()
+            .set_item("__annotations__", ann_dict.as_object().to_owned(), vm)
+            .is_err()
+    {
+        error!("Failed to initialize __main__.__annotations__")
+    }
```
youknowone, line 533: `and_then` is available when we want to run something only when an Option or Result is Some or Ok.

```suggestion
    let main_module_dict = main_module.dict();
    main_module_dict
        .and_then(|d| {
            d.set_item(
                "__annotations__",
                vm.ctx.new_dict().as_object().to_owned(),
                vm,
            )
            .ok()
        })
        .expect("Failed to initialize __main__.__annotations__");
```

youknowone (changes requested):
interesting issue, nice solution

src/lib.rs
moreal: Thanks to let me know it! awesome code 😀  🙏🏻

youknowone (approved):
thank you for contributing!

PR #2826, opened 2021-08-10
socket: use libc::hstrerror

deantvv:
Follow up for #2798


vm/src/stdlib/socket.rs
```diff
@@ -1713,8 +1712,8 @@ fn convert_socket_error(
                     std::str::from_utf8(s.to_bytes()).unwrap()
                 }
                 SocketError::HError => {
-                    // TODO: wait for libc hstrerror to land (https://github.com/rust-lang/libc/pull/2323)
-                    "host not found"
+                    let s = unsafe { ffi::CStr::from_ptr(libc::hstrerror(err.error_num())) };
+                    std::str::from_utf8(s.to_bytes()).unwrap()
```
youknowone, line 1716:
```suggestion
                    s.to_str().unwrap()
```
if `to_str` doesn't have a problem here

deantvv: Fix as suggested! Thanks!

youknowone (approved):

PR #2828, opened 2021-08-10
Add __floor__ and __ceil__ to float w/ tests

Tetramad:
Originally pull requested by @fanninpm [#2456]

### DIFF
--- a/vm/src/builtins/float.rs
+++ b/vm/src/builtins/float.rs
@@ -390,6 +390,11 @@ impl PyFloat {
     }
 
     #[pymethod(magic)]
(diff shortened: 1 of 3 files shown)


vm/src/builtins/float.rs
```diff
@@ -389,6 +389,16 @@ impl PyFloat {
         try_bigint(self.value, vm)
     }
 
+    #[pymethod(name = "__floor__")]
```
DimitrisJim, line 392: you can replace these with `pymethod(magic)` and the name is handled automatically. I.e:

```suggestion
    #[pymethod(magic)]
```

DimitrisJim (changes requested):

PR #2830, opened 2021-08-10
Add __alloc__ to bytearray and fix its __init__ issues.

DimitrisJim:
- `__alloc__` differs from CPython's `bytearray.__alloc__` due to the difference in the underlying storage used.
 - `__init__` now controls actual initialization instead of `__new__`. `__new__` now also accepts `*args, **kwargs` as required.

### DIFF
--- a/vm/src/bytesinner.rs
+++ b/vm/src/bytesinner.rs
@@ -294,6 +294,11 @@ impl PyBytesInner {
     }
 
     #[inline]
+    pub fn capacity(&self) -> usize {
+        self.elements.capacity()
+    }
+
+    #[inline]
     pub fn is_empty(&self) -> bool {
         self.elements.is_empty()
     }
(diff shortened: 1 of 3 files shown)


vm/src/bytesinner.rs
```diff
@@ -120,12 +120,8 @@ impl ByteInnerNewOptions {
         PyBytes::from(inner).into_ref_with_type(vm, cls)
     }
 
-    pub fn get_bytearray(
-        mut self,
-        cls: PyTypeRef,
-        vm: &VirtualMachine,
-    ) -> PyResult<PyByteArrayRef> {
-        let inner = if let OptionalArg::Present(source) = self.source.take() {
+    pub fn get_bytearray_inner(mut self, vm: &VirtualMachine) -> PyResult<PyBytesInner> {
```
youknowone, line 123: :+1:

DimitrisJim: There's quite an overlap between it and `get_bytes` but trying to refactor things seemed out of scope for now (plus, not sure how much better the end result would be).

youknowone (approved):

PR #2831, opened 2021-08-10
bump up reference CPython to 3.9

youknowone:
This PR is not intended to be merged.
Just to check the CI result - at least at the moment.

related issue for disabled test: #3212


DimitrisJim:
Since #2417 was merged I see no reason why the CI shouldn't be bumped to 3.9 too.

youknowone:
probably because python 3.9 was not stable yet at that time

PR #2841, opened 2021-08-10
Fixes that `id()` of `tuple` was changing on `* 1`

### LINKED ISSUE #2840: `id(a) != id(a * 1)` for `tuple`, str and bytes
## Feature

CPython:

```python
>>> a = (1, 2)
>>> id(a), id(a * 1)
(4379030656, 4379030656)
```

RustPython:

```
>>>>> a = (1, 2)
>>>>> id(a), id(a * 1)
(140352989490368, 140352989503168)
```

This is also covered by `seq_tests.test_repeat` which is now marked as `unittest.expectedFailure`.

sobolevn:
Refs #2840

Reference: https://github.com/python/cpython/blob/c0ab59f7de1906feee21c057ad433fad924d1e38/Objects/tupleobject.c#L560-L567

### DIFF
--- a/vm/src/builtins/tuple.rs
+++ b/vm/src/builtins/tuple.rs
@@ -172,7 +172,7 @@ impl PyTuple {
     fn mul(zelf: PyRef<Self>, counter: isize, vm: &VirtualMachine) -> PyRef<Self> {
         if zelf.elements.is_empty() || counter == 0 {
             vm.ctx.empty_tuple.clone()
-        } else if counter == 1 && zelf.clone_class().is(PyTuple::class(vm)) {
+        } else if counter == 1 && zelf.class().is(&vm.ctx.types.tuple_type) {
             // Special case: when some `tuple` is multiplied by `1`,
             // nothing really happens, we need to return an object itself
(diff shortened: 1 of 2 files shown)


vm/src/builtins/tuple.rs
```diff
@@ -169,11 +169,17 @@ impl PyTuple {
 
     #[pymethod(name = "__mul__")]
     #[pymethod(name = "__rmul__")]
-    fn mul(&self, counter: isize, vm: &VirtualMachine) -> PyRef<Self> {
-        if self.elements.is_empty() || counter == 0 {
+    fn mul(zelf: PyRef<Self>, counter: isize, vm: &VirtualMachine) -> PyRef<Self> {
+        if zelf.elements.is_empty() || counter == 0 {
             vm.ctx.empty_tuple.clone()
+        } else if counter == 1 && zelf.clone_class().is(PyTuple::class(vm)) {
```
DimitrisJim, line 175: No need to clone here:

```suggestion
        } else if counter == 1 && zelf.class().is(&vm.ctx.types.tuple_type) {
```

PR #2843, opened 2021-08-11
add deque __iadd__ method

### LINKED ISSUE #2838: `collections.deque` doesn't have `__iadd__` method.
## Feature

CPython:

```python
>>> d = deque()
>>> d += deque(range(3))
deque([0, 1, 2])
```

RustPython:

```python
>>>>> d = deque()
>>>>> d += deque(range(3))
TypeError: '+=' not supported between instances of 'deque' and 'deque'
```

eldpswp99:
resolve #2838 

add `deque.__iadd__` method

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -8,7 +8,7 @@ mod _collections {
     use crate::slots::{Comparable, Hashable, Iterable, PyComparisonOp, PyIter, Unhashable};
     use crate::vm::ReprGuard;
     use crate::VirtualMachine;
-    use crate::{sequence, sliceable};
+    use crate::{sequence, sliceable, IdProtocol, TryFromObject};
     use crate::{PyComparisonValue, PyIterable, PyObjectRef, PyRef, PyResult, PyValue, StaticType};
     use itertools::Itertools;
     use std::collections::VecDeque;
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/collections.rs
```diff
@@ -321,6 +322,23 @@ mod _collections {
...
         }
+
+        #[pymethod(magic)]
+        fn iadd(
+            zelf: PyRef<Self>,
+            other: PyObjectRef,
+            vm: &VirtualMachine,
+        ) -> PyResult<PyRef<Self>> {
+            let other: PyObjectRef = if zelf.is(&other) {
+                vm.new_pyobj(zelf.copy())
```
youknowone, line 333: by calling here `new_pyobj` and calling `try_from_object` later, this is making unnessessary `PyObjectRef`.

youknowone (changes requested):

eldpswp99:
the unccessary pyObject was created due to extend bug regard to recursive call , i'll fix extend method first and fix this later

youknowone (approved):
Ok, It looks more like `extend` problem not `__iadd__` one. I will merge this PR for now. We can fix it when `extend` is fixed later.

Thank you for contributing!

PR #2844, opened 2021-08-11
Adds `__rmul__` to `collections.deque`

### LINKED ISSUE #2834: `collections.deque` does not have `__rmul__` method
## Feature

CPython:

```python
>>> deque((1, 2)).__rmul__(2)
deque([1, 2, 1, 2])
```

RustPython:

```python
>>>>> deque((1, 2)).__rmul__(2)
AttributeError: 'deque' object has no attribute '__rmul__'
```

sobolevn:
Closes #2834

Note: at the moment almost all collections are missing `MemoryError` corner case, like here: https://github.com/python/cpython/blob/d1ae57027fc39ff60dcfc1b63881400e5ca3ce56/Modules/_collectionsmodule.c#L709-L711

I guess that this is out of scope of this task, related:
- https://github.com/RustPython/RustPython/issues/1750
- https://github.com/RustPython/RustPython/pull/1779

I think that some tests in `test_deque` might hang because of that.

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -307,10 +307,10 @@ mod _collections {
         fn mul(&self, n: isize) -> Self {
             let deque: SimpleSeqDeque = self.borrow_deque().into();
             let mul = sequence::seq_mul(&deque, n);
-            let skipped = if let Some(maxlen) = self.maxlen.load() {
-                mul.len() - maxlen
-            } else {
-                0
+            let mul_len = mul.len();
+            let skipped = match self.maxlen.load() {
+                Some(maxlen) if mul_len > maxlen => mul_len - maxlen,
+                _ => 0,
             };
             let deque = mul.skip(skipped).cloned().collect();
             PyDeque {
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -303,6 +303,7 @@ mod _collections {
         }
 
         #[pymethod(magic)]
+        #[pymethod(name = "__rmul__")]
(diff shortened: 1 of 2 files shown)


sobolevn:
Local test produces a very strange result:

```
» target/debug/rustpython 
>>>>> d = deque('abc', maxlen=5)
>>>>> d * 1 == deque('abc')
thread 'main' panicked at 'attempt to subtract with overflow', vm/src/stdlib/collections.rs:310:17
```

youknowone:
`d.len()` is `3`, `maxlen` is `5`. so `3-5` is overflowed.

DimitrisJim:
> I think that some tests in test_deque might hang because of that.

That's deadlocking, unfortunately. I've idly been trying to find a good way to tackle it. It is for another time, though.

vm/src/stdlib/collections.rs
```diff
@@ -303,13 +303,14 @@ mod _collections {
...
             let deque: SimpleSeqDeque = self.borrow_deque().into();
             let mul = sequence::seq_mul(&deque, n);
-            let skipped = if let Some(maxlen) = self.maxlen.load() {
-                mul.len() - maxlen
-            } else {
-                0
+            let mul_len = mul.len();
+            let skipped = match self.maxlen.load() {
+                Some(maxlen) if mul_len > maxlen => mul_len - maxlen,
+                _ => 0,
```
youknowone, line 313:
```rust
self.maxlen.load().and_then(|maxlen| mul_len.wrapping_sub(max_len)).unwrap_or(0)
```

DimitrisJim (approved):

PR #2845, opened 2021-08-11
vscode lldb configuration template

youknowone:
vscode's auto-generated lldb doesn't catch the exception.

I edited it a bit using `preLaunchTask`
this template will be helpful for people who want to use debugger

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,283 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
+    "configurations": [
+        {
+            "type": "lldb",
+            "request": "launch",
+            "name": "Debug executable 'rustpython'",
+            "preLaunchTask": "Build RustPython Debug",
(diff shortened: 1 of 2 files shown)


youknowone:
@coolreader18 @DimitrisJim any cons to adding these files?

DimitrisJim:
I can see how these would be helpful to new contributors who want to set up VSCode for learning RustPython. 

I'm not sure if these are better suited for inclusion in the codebase or if it would be better to create a Wiki/doc page on "How to set up VSCode for debugging RustPython."

youknowone:
If we don't have any cons about it, i think encouraging everybody to read docs and copy & paste configurations from it whenever they set a new environment (github is promoting codespaces) is not very productive way.

DimitrisJim:
I honestly haven't heard of codespaces :smile: If github supports using a repo as a vscode project then it sounds way cooler now.

youknowone:
image

youknowone:
I think we still need documents even with the patch. i will write one with this patch. let's go for it now. let me know if it will have any problem.

DimitrisJim:
@youknowone I'm actually looking forward to trying codespaces, so this pr looks like an excellent choice.

youknowone:

PR #2846, opened 2021-08-11
Fixed unwrapping in frame.rs

### LINKED ISSUE #2799: panic occurs when a key in the dictionary is unhashable.

Arnab1181412:
closes #2799

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -1306,7 +1306,7 @@ impl ExecutingFrame<'_> {
                             return Err(vm.new_type_error(msg));
                         }
                     }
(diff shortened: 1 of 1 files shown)


DimitrisJim (approved):

DimitrisJim:
Thanks!

PR #2849, opened 2021-08-11
Refactor Buffer Protocol (mostly namings)

youknowone:
- tp_buffer -> tp_as_buffer by following CPython name
- Buffer -> PyBuffer
- try_buffer_from_object -> PyBufferRef::try_from_object
- sorted type slots


DimitrisJim (approved):
Haven't really had a change to look very much into the type slots but from what I can understand it looks good to me.

PR #2851, opened 2021-08-11
Fix os sendfile

deantvv:
For #1175

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -517,34 +517,68 @@ mod _os {
         fd.map(|fd| fd.0).map_err(|e| e.into_pyexception(vm))
     }
 
+    #[cfg(any(target_os = "linux", target_os = "macos"))]
+    #[derive(FromArgs)]
+    struct SendFileArgs {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/os.rs
```diff
@@ -517,34 +517,79 @@ mod _os {
...
+        let headers = _extract_vec_bytes(args.headers, vm)?;
+        let count = headers
+            .as_ref()
+            .map(|v| {
+                let mut header_sz = 0;
+                for slice in v {
+                    header_sz += slice.len();
+                }
+                header_sz
+            })
```
youknowone, line 590:
```suggestion
            .map(|v| v.iter().map(|s| s.len()).sum())
```

youknowone (changes requested):

vm/src/stdlib/os.rs
deantvv: @youknowone When I was writing this code, I thought there should be a better way but I just can't figure it out. Thank you for sharing that!

coolreader18 (approved):

PR #2852, opened 2021-08-11
Fix to use existing variable in test_unicode.py

tony-jinwoo-ahn:


youknowone:
Thank you for contributing.
I am sorry but we don't update CPython unittest specifically one by one.

By looking at history of Lib/test/test_unicode.py, the file is based on CPython 3.8.2.
If you want to update the test file to newer version, please copy a new version test from newer CPython version, rather than updating specific function manually.

tony-jinwoo-ahn:
Thank you!

PR #2853, opened 2021-08-11
support deque.extend recursive call

eldpswp99:
before: RustPython hangs when

	d = deque('a')
	d.extend(d)

after: it extends to the copy of it

I'll fix the temporary change in deque.__iadd__ with another PR

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -137,17 +137,13 @@ mod _collections {
         fn extend(zelf: PyRef<Self>, iter: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
             // TODO: use length_hint here and for extendleft
             if zelf.is(&iter) {
-                let iter =
-                    PyIterable::try_from_object(vm, PyDeque::iter(zelf.copy().into_ref(vm), vm)?)?;
-                for elem in iter.iter(vm)? {
-                    zelf.append(elem?)
-                }
-                Ok(())
+                let copied: Vec<PyObjectRef> = vm.extract_elements(zelf.as_object())?;
+                zelf._extend(copied.into_iter().map(Ok))?;
             } else {
                 let iter = PyIterable::try_from_object(vm, iter)?;
-
-                zelf._extend(iter.iter(vm)?)
+                zelf._extend(iter.iter(vm)?)?;
             }
+            Ok(())
         }
 
         #[pymethod]
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -79,7 +79,7 @@ mod _collections {
                 maxlen: AtomicCell::new(maxlen),
             };
             if let OptionalArg::Present(iter) = iter {
-                py_deque._extend(iter, vm)?;
+                py_deque._extend(iter.iter(vm)?)?;
             }
             py_deque.into_ref_with_type(vm, cls)
         }
@@ -126,9 +126,8 @@ mod _collections {
             Ok(count)
         }
 
-        fn _extend(&self, iter: PyIterable, vm: &VirtualMachine) -> PyResult<()> {
-            // TODO: use length_hint here and for extendleft
-            for elem in iter.iter(vm)? {
+        fn _extend(&self, iter: impl std::iter::Iterator<Item = PyResult>) -> PyResult<()> {
+            for elem in iter {
                 self.append(elem?);
             }
             Ok(())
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/collections.rs
```diff
@@ -126,15 +126,31 @@ mod _collections {
...
             for elem in iter.iter(vm)? {
                 self.append(elem?);
             }
             Ok(())
         }
 
+        #[pymethod]
+        fn extend(zelf: PyRef<Self>, iter: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+            // TODO: use length_hint here and for extendleft
+            if zelf.is(&iter) {
```
DimitrisJim, line 140: Pretty sure you can follow what `list` does here. Extract the elements and then use `VecDeque`s `append`.

vm/src/stdlib/collections.rs
```diff
@@ -126,15 +126,31 @@ mod _collections {
...
 
-        #[pymethod]
-        fn extend(&self, iter: PyIterable, vm: &VirtualMachine) -> PyResult<()> {
+        fn _extend(&self, iter: PyIterable, vm: &VirtualMachine) -> PyResult<()> {
             // TODO: use length_hint here and for extendleft
             for elem in iter.iter(vm)? {
                 self.append(elem?);
             }
             Ok(())
         }
```
youknowone, line 135: My suggestion about this function is:
```suggestion
        fn _extend(&self, iter: impl std::iter::Iterator<Item=PyResult>) -> PyResult<()> {
            for elem in iter {
                self.append(elem?);
            }
            Ok(())
        }
```
Then it will take general rust iterator rather than a PyIterable - which also be able to accept the first branch of next function

vm/src/stdlib/collections.rs
youknowone: I think he is trying somewhat better way. It will be `VecDeque` -> rust `Vec` -> `VecDeque` transition. In this case, it will only depending on iterators - which is more cpython way. (they call iternext and append for every iteration)

DimitrisJim: Indeed, I'd argue there's precedent in the project for using `extract_elements` and it makes the function readable, plus, it *should* be relatively fast for the common case (`PyList`s). 

You've got more experience here so in the end of the day it's definitely your call.

youknowone: I agree that is faster in list. (for there, Vec -> Vec -> Vec)
I think our `extract_elements` is same as CPython's `PySequence_Fast_ITEMS`, right?
CPython uses PySequence_Fast_ITEMS for list, but not for deque. I think this is how also they handle the different situations.

youknowone: Oh, that sounds like another point.

DimitrisJim: It should also be possible to special case `iter` being another `PyDeque` and just cloning its `VecDeque`.

vm/src/stdlib/collections.rs
```diff
@@ -127,11 +128,22 @@ mod _collections {
...
+            let mut elements: Vec<PyObjectRef> = vm.extract_elements(&iter)?;
+            if let Some(max_len) = max_len {
+                let cut_len = max_len as isize - elements.len() as isize;
+                if cut_len > 0 {
+                    let mut deque = zelf.borrow_deque_mut();
+                    let drain_until = deque.len().saturating_sub(cut_len as usize);
+                    deque.drain(..drain_until);
+                } else {
+                    zelf.borrow_deque_mut().clear();
+                    elements.drain(..(-cut_len) as usize);
```
DimitrisJim, line 143: I'd just compare these:

```suggestion
                if max_len > elements.len() {
                    let mut deque = zelf.borrow_deque_mut();
                    let drain_until = deque.len().saturating_sub(max_len - elements.len());
                    deque.drain(..drain_until);
                } else {
                    zelf.borrow_deque_mut().clear();
                    elements.drain(..elements.len() - max_len);
```

youknowone: the new code has additional subtraction operations. idk if they will be optimized out, but why if we can avoid it explicitly?

DimitrisJim: Large values of `maxlen` (i.e `deque([1], maxlen = maxsize + 2)` cause panics. I just noticed that `maxsize` shouldn't be allowed to be `> isize::MAX`, though, and I'll fix that at another PR.

Either way, using a comparison here means we only operate on positive numbers, which means easier to follow and harder to make mistakes (personally, at least!)

DimitrisJim (approved):

PR #2855, opened 2021-08-11
Add  reversed  to deque

whgusdn321:


DimitrisJim:
Hey @whgusdn321, this issue is slightly trickier than the rest. If you want, I can find you a better issue to start with.

youknowone:
The function is called `reversed` but you didn't add any code related to reversion. This patch have to include `PyDequeReverseIterator` just like `PyListReverseIterator`.

whgusdn321:
> Hey @whgusdn321, this issue is slightly trickier than the rest. If you want, I can find you a better issue to start with.

yep it would be really thankful if you find be another issue to start with.

DimitrisJim:
Hey @whgusdn321, I just opened a new issue (#2877) which I believe would be a much better one to start with. If you want to tackle it, drop a comment on the issue so I can assign it to you!

PR #2856, opened 2021-08-11
give an own module for buffer protocol + TryFromBorrowedObject

### LINKED ISSUE #2854: Reference version of TryObjectFrom
TryObjectFrom now take `PyObjectRef` instead of `&PyObjectRef`.

From Python call, it must be PyObejctRef, but not in rust internal call.
We can add another trait to optimize in-rust usage of them.

youknowone:
TryFromBorrowedObject reduces unnessessary cloning.

PyBuffer is a trait for buffer protocol but hidden in memory view codes. Giving it a sole module will be a guideline for future protocols.

renamed `PyBytesLike`, `PyRwBytesLike` and `BufOrStr` into `ArgBytesLike`, `ArgMemoryBuffer` and `ArgStrOrBytesLike` by following the description of `getarg.c`. Let's keep `Py` prefixes only for actual python objects.

closes #2854

### DIFF
--- a/vm/src/buffer.rs
+++ b/vm/src/buffer.rs
@@ -1,0 +1,167 @@
+//! Buffer protocol
+
+use crate::common::borrow::{BorrowedValue, BorrowedValueMut};
+use crate::common::rc::PyRc;
+use crate::PyThreadingConstraint;
+use crate::{PyObjectRef, PyResult, TypeProtocol};
+use crate::{TryFromBorrowedObject, VirtualMachine};
+use std::{borrow::Cow, fmt::Debug, ops::Deref};
+
+pub trait PyBuffer: Debug + PyThreadingConstraint {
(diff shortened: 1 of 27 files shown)


DimitrisJim:
Ah, I don't feel I can give a good review for this (yet!) unfortunately. I'm not all that familiar with this part of the codebase.

coolreader18 (approved):
LGTM!

coolreader18:
Wondering if `TryFromBorrowedObject` should have a lifetime parameter (`TryFromBorrowedObject<'a>: 'a { fn(&'a PyObjectRef) -> PyResult<Self> }`) to allow e.g. `&str` to work as a function parameter, but that might make things more confusing

youknowone:
@coolreader18 I didn't try it, but maybe yes. Let's try it when we need it.

PR #2858, opened 2021-08-11
`id()` for `bytes` stays the same after `* 1`

sobolevn:
Refs #2840

Only `str` is left.

### DIFF
--- a/extra_tests/snippets/bytes.py
+++ b/extra_tests/snippets/bytes.py
@@ -624,3 +624,25 @@ assert bytes(A()) == b"bytess"
 # Issue #2125
 b = b'abc'
 assert bytes(b) is b
+
+
+# Regression to
+# https://github.com/RustPython/RustPython/issues/2840
+
+a = b'123abc!?'
+assert id(a) == id(a)
+assert id(a) != id(a * -1)
+assert id(a) != id(a * 0)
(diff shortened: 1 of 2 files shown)


DimitrisJim (approved):

youknowone:
I expected those are tested by CPython unittest suites 😕

sobolevn:
@youknowone me too, but I was not able to find any for `bytes`. Maybe it is something to contribute to CPython? 🙂

fanninpm:
The CPython test suites in *this* repository are currently incomplete.

youknowone:
@sobolevn yes, if we can cover tests with CPython one, we don't need to manage our own `extra_tests`. less code means less management cost.

sobolevn:
@fanninpm I was searching in CPython's repo as well: https://github.com/python/cpython/blob/main/Lib/test/test_bytes.py
Do you have any specific CPython's test to point me to? 🤔

fanninpm:
> @fanninpm I was searching in CPython's repo as well: https://github.com/python/cpython/blob/main/Lib/test/test_bytes.py
> Do you have any specific CPython's test to point me to?

After a cursory search, I couldn't find anything specific. I suggest asking the wiser folks on https://bugs.python.org.

PR #2862, opened 2021-08-11
Check compilation for Android and Redox in CI

coolreader18:


vm/src/stdlib/os.rs
```diff
@@ -262,27 +262,7 @@ impl IntoPyException for &'_ io::Error {
...
     fn into_pyexception(self, vm: &VirtualMachine) -> PyBaseExceptionRef {
-        match self {
-            nix::Error::InvalidPath => {
-                let exc_type = vm.ctx.exceptions.file_not_found_error.clone();
-                vm.new_exception_msg(exc_type, self.to_string())
-            }
-            nix::Error::InvalidUtf8 => {
-                let exc_type = vm.ctx.exceptions.unicode_error.clone();
-                vm.new_exception_msg(exc_type, self.to_string())
-            }
```
youknowone, line 273: is this branch covered by `io::Error::from`?

youknowone (commented):
this is a great news. 
do you want to remove warnings here or later?

vm/src/stdlib/os.rs
coolreader18: In 0.22, if something needs to validate utf8 it'll return something other than a nix::Result.

youknowone (approved):

PR #2865, opened 2021-08-12
support deque.insert when idx is gte to len

eldpswp99:
before: 

when idx >= len it doesn't insert the element to the end
it inserted to index len-1

after:  

when idx >= len it inserts the element to the end of the deque


youknowone (approved):
Nice catch, thank you for contributing!

PR #2866, opened 2021-08-12
remove improper setter of deque.maxlen

eldpswp99:
before: maxlen was writable
        d = deque()
        d.maxlen = 10 was allowed
after : make maxlen read-only
	the before code raise attributError


youknowone (commented):
When `maxlen` is a read-only field, I think its type also can be changed from `AtomicCell<Option<usize>>` to `Option<usize>`. Do you want to try this also?

eldpswp99:
@youknowone Then, I'll fix that on this PR!

PR #2871, opened 2021-08-12
Deque.__init__ handles initialization.

DimitrisJim:


vm/src/stdlib/collections.rs
```diff
@@ -3,24 +3,26 @@ pub(crate) use _collections::make_module;
...
     use crossbeam_utils::atomic::AtomicCell;
 
     #[pyattr]
     #[pyclass(name = "deque")]
-    #[derive(Debug)]
+    #[derive(Debug, Default)]
     struct PyDeque {
         deque: PyRwLock<VecDeque<PyObjectRef>>,
-        maxlen: Option<usize>,
+        maxlen: AtomicCell<Option<usize>>,
```
DimitrisJim, line 25: Missed the fact that its mutated in the last review, unfortunately.

DimitrisJim: I'd have to look into it because I'm not sure how to do it now. Wouldn't that require getting a lock for reading it, too?

If you have a way to do it now, just push to get the branch directly

youknowone: I pushed a commit. I also fixed a bug that `maxlen=None` is ignored

DimitrisJim: Good catch on `maxlen`. That's a pretty nifty trick there. I'll keep it in my in case this pops up someplace else.

youknowone (approved):

PR #2875, opened 2021-08-13
Retrying contains, remove, discard with a frozenset.

DimitrisJim:
If the item passed to a sets `{remove, discard, contains}` is a set/set subclass, retry the operation after converting it to a frozenset.

### DIFF
--- a/Lib/test/test_set.py
+++ b/Lib/test/test_set.py
@@ -379,8 +379,6 @@ class TestSet(TestJointOps, unittest.TestCase):
     thetype = set
     basetype = set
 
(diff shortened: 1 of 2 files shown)


youknowone (approved):
huh, this is a design totally unexpected.

DimitrisJim:
yup, I also wasn't aware of it until I took a closer look at the tests. I knew they do an optimization with set literals whereby they transform them into `frozenset`s at some cases (i.e `dis('0 in {1, 2}')` but wasn't aware of this behavior for the methods.

PR #2876, opened 2021-08-13
Adds .editorconfig file

sobolevn:
It was really bothering me that my `vscode` was not adding 4 spaces as the default ident.

What is `editorconfig`? It is a tool to standartize things like new lines, idents, and whitespaces across different editors and different OSes.
Docs: https://editorconfig.org/

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# Check http://editorconfig.org for more information
+# This is the main config file for this project:
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+trim_trailing_whitespace = true
+end_of_line = lf
+insert_final_newline = true
+indent_size = 2
+
+[*.py]
+indent_size = 4
+
+[*.rs]
+indent_size = 4


youknowone:
Do you experience it even after installing RLS(Rust Language Server)? We just follow the default rule of rustfmt, so I expect that must be configured well at least after installing the plugins.

sobolevn:
I use https://github.com/rust-analyzer/rust-analyzer
Without this file it is inconvenient in the formatting part 🙂

fanninpm:
No real need. Just make sure `"editor.formatOnSave": true` is in your workspace's `.vscode/settings.json`, and `rustfmt` will take care of the rest. (Remember to turn it off when you're working within the `Lib/` directory, though.)

youknowone:
I think `formatOnSave` is a convenient tool but not the proper solution of the wrong indentation for editing itself. I wonder why it happens. I tried a bare vscode but didn't have that problem. So now I think you have some global indent configuration in vscode.

To test a bare vscode, stroke `.` key in anywhere of https://github.com/RustPython/RustPython/ . it will show a web version of vscode without much things.

youknowone:
I think #2897 is a reasonable support for this PR, though it only can fix the indent problem but not formatting itself.

PR #2878, opened 2021-08-13
Add reduce for set, frozenset.

DimitrisJim:
Also fixes a minor issue with `set.__new__`'s signature.

I added an `elements` method for inner for this that just exposes the underlying `keys`.


coolreader18 (approved):
LGTM!

youknowone (approved):

PR #2879, opened 2021-08-14
Remove unnecessary slot wrapper addition

### LINKED ISSUE #2869: Added undefined `__new__` method to non-built-in types
## Feature

Currently, non-built-in types (e.g. user defined class) has its `__new__` method. So they are different with `object.__new__` and it cause error in `inspect._signature_from_callable()`.

### RustPython-side

```
>>>>> type.__new__
<bound method __new__ of <class 'type'>>
>>>>> object.__new__
<bound method __new__ of <class 'object'>>
>>>>> class A:
.....   ...
..... 
>>>>> A.__new__
<bound method __new__ of <class '__main__.A'>>
>>>>> A.__new__ == object.__new__
False
```

### CPython 3.8-side
```python
>>> type.__new__
<built-in method __new__ of type object at 0x7fc673a1c480>
>>> object.__new__
<built-in method __new__ of type object at 0x7fc673a1c620>
>>> class A:
...   ...
... 
>>> A.__new__
<built-in method __new__ of type object at 0x7fc673a1c620>
>>> A.__new__ == object.__new__
True
```

## Python Documentation

- https://github.com/python/cpython/blob/fa919fdf2583bdfead1df00e842f24f30b2a34bf/Objects/typeobject.c#L6065-L6083
- https://docs.python.org/3/reference/datamodel.html?highlight=__new__#object.__new__

moreal:
This pull request resolves #2869.

## Before (RustPython)

```
>>>>> class A: ...
>>>>> A.__new__
<bound method __new__ of <class '__main__.A'>>
>>>>> object.__new__
<bound method __new__ of <class 'object'>>
>>>>> A.__new__ == object.__new__
False
```

## After (RustPython)

```
>>>>> class A: ...
>>>>> A.__new__
<bound method __new__ of <class 'object'>>
>>>>> object.__new__
<bound method __new__ of <class 'object'>>
>>>>> A.__new__ == object.__new__
True
```

### DIFF
--- a/Lib/test/test_enum.py
+++ b/Lib/test/test_enum.py
@@ -1698,8 +1698,6 @@ class TestEnum(unittest.TestCase):
             test = 1
         self.assertEqual(Test.test.test, 'dynamic')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_no_duplicates(self):
         class UniqueEnum(Enum):
             def __init__(self, *args):
--- a/vm/src/builtins/pytype.rs
+++ b/vm/src/builtins/pytype.rs
@@ -467,8 +467,6 @@ impl PyType {
         let typ = new(metatype, name.as_str(), base, bases, attributes, slots)
             .map_err(|e| vm.new_type_error(e))?;
 
-        vm.ctx.add_slot_wrappers(&typ);
-
         // avoid deadlock
         let attributes = typ
             .attributes


youknowone:
Once this is successful, I think we also can remove the if statement in `add_slot_wrappers`

moreal:
> Once this is successful, I think we also can remove the if statement in `add_slot_wrappers`

I understood the code you mean is below line. Sadly, when I remove the line I met `segmentation fault (core dumped)` 😞 

https://github.com/RustPython/RustPython/blob/61bae9aee84e80605dd4a93c95d507d01f752d8f/vm/src/pyobject.rs#L392

youknowone:
Yes. actually I expected the same thing for this patch itself also ;)

I don't know well the detail - so this is a rough guess. I think that line is added as a lazy workaround about double calling the method, mostly not by intension. Because you already removed the unnessessary duplicated call of the function. Maybe there will be another way to remove another duplicated call of it. (note:  I am not meaning that work needs to be included in this patch)

youknowone (approved):

PR #2880, opened 2021-08-14
add ` time.time_ns()`  function

Regexp96:
I add `time.time_ns()` 
` time_ns`  is to get the time in nanoseconds since the epoch


vm/src/stdlib/time_module.rs
```diff
@@ -58,6 +58,18 @@ pub fn get_time() -> f64 {
     Date::now() / 1000.0
 }
 
+#[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))]
```
youknowone, line 61: i didn't know wasm has confusing discussion about it.
https://stackoverflow.com/questions/6002808/is-there-any-way-to-get-current-time-in-nanoseconds-using-javascript

vm/src/stdlib/time_module.rs
```diff
@@ -58,6 +58,18 @@ pub fn get_time() -> f64 {
...
+pub fn get_time_ns() -> u64 {
```
youknowone, line 62: In general, we never want the interpreter is crashed by python code. The correct signature of failible function is like this:
```suggestion
pub fn get_time_ns() -> PyResult<u64> {
```

youknowone (changes requested):

vm/src/stdlib/time_module.rs
```diff
@@ -58,6 +58,18 @@ pub fn get_time() -> f64 {
...
+    match SystemTime::now().duration_since(UNIX_EPOCH) {
+        Ok(v) => v.as_nanos() as u64,
+        Err(err) => panic!("Time error: {:?}", err),
```
youknowone, line 65:
```suggestion
        Err(err) => unsafe { std::hint::unreachable_unchecked() }, // guaranted to be not to be happen with now() + UNIX_EPOCH
```

this hint will remove the branch itself for release build

youknowone (changes requested):

vm/src/stdlib/time_module.rs
```diff
@@ -267,5 +279,6 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         "struct_time" => struct_time_type,
         "time" => named_function!(ctx, time, time),
         "perf_counter" => named_function!(ctx, time, time), // TODO: fix
+        "time_ns" => named_function!(ctx,time,time_ns),
```
youknowone, line 282: `extend_module!` macro will be helpful to quarantine this function from wasm

youknowone:
@Regexp96 it looks good, but there is a conflict by recent merge. would you rebase it once again?

youknowone (approved):

coolreader18:
Ah, should've commented earlier, but isn't it possible for `::now() < UNIX_EPOCH` if someone sets the system time to before 1970? I'd think it should just return a negative number in that case (which is definitely possible, you can just do `Err(e) => e.duration()`), though I haven't looked at what cpython does

Issue #2877, opened 2021-08-13
`frozenset.copy` shouldn't return a new object.


DimitrisJim:
## Feature

CPython:

```python
>>> f = frozenset(range(1000))
>>> f2 = f.copy()
>>> f is f2
True
```

RustPython:

```python
>>>>> f = frozenset(range(1000))
>>>>> f2 = f.copy()
>>>>> f is f2
False
```

Calling `copy` on a `frozenset` should return itself and not a new object.

DimitrisJim:
@whgusdn321 do you want to handle this?

Snowapril:
@whgusdn321 There is difference in `id` of frozenset copy between rustpython and cpython.
#### rustpython
``` python
>>>>> a = frozenset(range(100))
>>>>> b = a.copy()
>>>>> id(a)
94675967765472
>>>>> id(b)
94675967759248
>>>>> a is b
False
```

#### cpython
```python
>>> a = frozenset(range(100))
>>> b = a.copy()
>>> id(a)
140340461132704
>>> id(b)
140340461132704
>>> a is b
True
```

I think `copy` method work of PyFrozenSet need to be modified.

youknowone:
I think this issue doesn't fit for bare new comer. It requires understanding of `PyRef` which doesn't fit to common sense of rust programmer.

Snowapril:
I fixed this issue in #2883 by fixing `copy` method in `PyFrozenSet`

PR #2883, opened 2021-08-14
Fix frozenset copy to return self

Snowapril:
This revision fixes issue #2877

implementation referenced on `imul` method of `PyList`.

#### Before
```python
>>>>> f = frozenset(range(1000))
>>>>> f2 = f.copy()
>>>>> f is f2
False
```

#### After
```python
>>>>> f = frozenset(range(1000))
>>>>> f2 = f.copy()
>>>>> f is f2
True
```

### DIFF
--- a/Lib/test/test_set.py
+++ b/Lib/test/test_set.py
@@ -738,8 +738,6 @@ class TestFrozenSet(TestJointOps, unittest.TestCase):
             results.add(hash(self.thetype(seq)))
         self.assertEqual(len(results), 1)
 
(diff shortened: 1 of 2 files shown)


youknowone (approved):

youknowone:
@Snowapril will you remove `expectedFailure` from the related tests?

Snowapril:
I found difference in copy implementation between mine and cpython, I add some type checking routine and remove expected failure annotation.

PR #2886, opened 2021-08-14
Add math.prod function and enable its tests

voidsatisfaction:
I implemented math.prod function and enabled its tests.

However, I have a problems to deal with the test which I commented below.

Please suggest better solution for this if there is an nice idea :)

### DIFF
--- a/Lib/test/test_math.py
+++ b/Lib/test/test_math.py
@@ -1809,9 +1809,8 @@ class MathTests(unittest.TestCase):
         self.assertEqual(type(prod([1, 2.0, 3, 4, 5, 6])), float)
         self.assertEqual(type(prod(range(1, 10000))), int)
         self.assertEqual(type(prod(range(1, 10000), start=1.0)), float)
-        # TODO: RUSTPYTHON
-        # self.assertEqual(type(prod([1, decimal.Decimal(2.0), 3, 4, 5, 6])),
-                        #  decimal.Decimal)
+        self.assertEqual(type(prod([1, decimal.Decimal(2.0), 3, 4, 5, 6])),
+                         decimal.Decimal)
 
     def testPerm(self):
         perm = math.perm
--- a/Lib/test/test_math.py
+++ b/Lib/test/test_math.py
@@ -1726,8 +1726,6 @@ class MathTests(unittest.TestCase):
             self.fail('Failures in test_mtestfile:\n  ' +
                       '\n  '.join(failures))
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_prod(self):
         prod = math.prod
         self.assertEqual(prod([]), 1)
@@ -1811,8 +1809,9 @@ class MathTests(unittest.TestCase):
         self.assertEqual(type(prod([1, 2.0, 3, 4, 5, 6])), float)
         self.assertEqual(type(prod(range(1, 10000))), int)
         self.assertEqual(type(prod(range(1, 10000), start=1.0)), float)
-        self.assertEqual(type(prod([1, decimal.Decimal(2.0), 3, 4, 5, 6])),
-                         decimal.Decimal)
+        # TODO: RUSTPYTHON
+        # self.assertEqual(type(prod([1, decimal.Decimal(2.0), 3, 4, 5, 6])),
+                        #  decimal.Decimal)
 
     def testPerm(self):
         perm = math.perm
--- a/Lib/test/test_statistics.py
+++ b/Lib/test/test_statistics.py
@@ -2078,9 +2078,6 @@ class TestStdev(VarianceStdevMixin, NumericTestCase):
         self.assertEqual(self.func(data, xbar=2.0), 1.0)
 
 class TestGeometricMean(unittest.TestCase):
-
-    # TODO: RUSTPYTHON math.prod isn't implemented
-    @unittest.expectedFailure
     def test_basics(self):
         geometric_mean = statistics.geometric_mean
         self.assertAlmostEqual(geometric_mean([54, 24, 36]), 36.0)
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -667,10 +667,7 @@ struct ProdArgs {
 fn math_prod(args: ProdArgs, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
     let iter = args.iterable;
 
-    let mut result = match args.start {
-        OptionalArg::Present(start) => start,
-        OptionalArg::Missing => vm.new_pyobj(1),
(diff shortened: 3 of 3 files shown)


Lib/test/test_math.py
```diff
@@ -1811,8 +1809,9 @@ def _naive_prod(iterable, start=1):
         self.assertEqual(type(prod([1, 2.0, 3, 4, 5, 6])), float)
         self.assertEqual(type(prod(range(1, 10000))), int)
         self.assertEqual(type(prod(range(1, 10000), start=1.0)), float)
-        self.assertEqual(type(prod([1, decimal.Decimal(2.0), 3, 4, 5, 6])),
-                         decimal.Decimal)
+        # TODO: RUSTPYTHON
+        # self.assertEqual(type(prod([1, decimal.Decimal(2.0), 3, 4, 5, 6])),
+                        #  decimal.Decimal)
```
voidsatisfaction, line 1814: If I uncommented it, it raises `TypeError: unexpected type Decimal` error.

youknowone: Please do not edit test code itself. Rather than that, keep this function `expectedFailure` as it was. We can remove it when it finally pass all test one day.

vm/src/stdlib/math.rs
```diff
@@ -657,6 +657,64 @@ fn math_remainder(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResu
...
+    #[pyarg(positional)]
+    iterable: PyIterable<Either<PyFloatRef, PyIntRef>>,
+    #[pyarg(named, optional)]
+    start: OptionalArg<Either<PyFloatRef, PyIntRef>>,
+}
+fn math_prod(args: ProdArgs, vm: &VirtualMachine) -> PyResult<Either<BigInt, f64>> {
+    let iter = args.iterable;
+
+    let mut is_result_float = false;
+    let mut i_result: BigInt = BigInt::one();
```
youknowone, line 671:
```suggestion
    let mut i_result = BigInt::one();
```

Lib/test/test_math.py
DimitrisJim: Yes, `prod` accepts any number type. You've currently restricted it to `int`s and `float`s. Your iterable needs to hold `PyObject`s which you multiply. CPython specializes for `int`s and `float`s and for any other type falls back to generic number multiplication (`PyNumber_Multiply`) which is _roughly_ equivalent to `vm._mul`.

vm/src/stdlib/math.rs
```diff
@@ -657,6 +657,64 @@ fn math_remainder(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResu
...
+    let mut f_result: f64 = 1.0;
```
youknowone, line 672: we only need a single type of result, but for now, preparing both of them reagardless the type of input.

vm/src/stdlib/math.rs
```diff
@@ -657,6 +657,64 @@ fn math_remainder(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResu
...
+            Either::B(z) => {
+                if is_result_float {
+                    let z = z.as_bigint();
+                    f_result *= int::to_float(z, vm)?;
+                    continue;
+                }
+
+                let z = z.as_bigint();
+                i_result *= z.clone()
+            }
```
youknowone, line 707: let me think how can we do this better

youknowone (changes requested):

Lib/test/test_math.py
yeongyumin: Thank you for your advice! I will re-implement it

vm/src/stdlib/math.rs
```diff
@@ -657,6 +657,34 @@ fn math_remainder(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResu
...
+    let iter = args.iterable;
+
+    let mut result = match args.start {
+        OptionalArg::Present(start) => start,
+        OptionalArg::Missing => vm.new_pyobj(1),
+    };
+
+    for obj in iter.iter(vm)? {
+        let obj = obj?;
+        if let Ok(Some(_v)) = float::try_float_opt(&obj, vm) {
```
youknowone, line 677: `try_float_opt` is only used for floats. this functino shouldn't be called for general path.

vm/src/stdlib/math.rs
```diff
@@ -657,6 +657,38 @@ fn math_remainder(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResu
...
+        let obj = obj?;
+
+        result = vm._mul(&result, &obj).unwrap_or_else(|_| {
+            multiply_error_occured = true;
+            result
+        });
+
+        if multiply_error_occured {
+            return Err(vm.new_type_error("math type error".to_owned()));
+        }
```
youknowone, line 686:
```suggestion
        result = vm._mul(&result, &obj).map_err(|_| vm.new_type_error("math type error".to_owned()))?;
```

yeongyumin: Thank you for nice suggestion!

Lib/test/test_statistics.py
```diff
@@ -2078,9 +2078,6 @@ def test_center_not_at_mean(self):
         self.assertEqual(self.func(data, xbar=2.0), 1.0)
 
 class TestGeometricMean(unittest.TestCase):
-
-    # TODO: RUSTPYTHON math.prod isn't implemented
-    @unittest.expectedFailure
```
yeongyumin, line 2083: I removed it to escape unexpected success error

vm/src/stdlib/math.rs
```diff
@@ -657,6 +657,34 @@ fn math_remainder(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResu
...
+    #[pyarg(named, optional)]
+    start: OptionalArg<PyObjectRef>,
+}
+fn math_prod(args: ProdArgs, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
+    let iter = args.iterable;
+
+    let mut result = match args.start {
+        OptionalArg::Present(start) => start,
+        OptionalArg::Missing => vm.new_pyobj(1),
+    };
```
youknowone, line 673:
```suggestion
    let mut result = args.start.unwrap_or_else(|| v.new_pyobj(1));
```

voidsatisfaction: very clear advice! thanks a lot!

youknowone (approved):

youknowone:
thank you for contributing!

PR #2887, opened 2021-08-14
Add reverse deque iterator.

### LINKED ISSUE #2836: `collections.deque` doesn't have `__reversed__` method
## Feature

CPython:

```python
>>> reversed(deque(range(3))
<_collections._deque_reverse_iterator at 0x7fec6d2fff40>
```

RustPython:

```python
>>>>> reversed(deque(range(3))
<iter object at 0x5622a0d03340>
```
`reversed(deque)` currently falls back to a generic iterator instead of having a specific `deque_reverse_iterator` type.

DimitrisJim:
closes #2836

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -2,12 +2,18 @@ pub(crate) use _collections::make_module;
 
 #[pymodule]
 mod _collections {
-    use crate::builtins::{PyInt, PyTypeRef};
     use crate::common::lock::{PyRwLock, PyRwLockReadGuard, PyRwLockWriteGuard};
     use crate::function::{FuncArgs, OptionalArg};
     use crate::slots::{Comparable, Hashable, Iterable, PyComparisonOp, PyIter, Unhashable};
     use crate::vm::ReprGuard;
     use crate::VirtualMachine;
+    use crate::{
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/collections.rs
```diff
@@ -426,6 +446,8 @@ mod _collections {
     #[derive(Debug)]
     struct PyDequeIterator {
         position: AtomicCell<usize>,
+        status: AtomicCell<IterStatus>,
+        length: usize, // To track length immutability.
```
youknowone, line 450: by adding 2 more fields, it looks like to be better to have `fn new(len: usize, deque: PyRef<PyDeque>)` static method.

youknowone (approved):

vm/src/stdlib/collections.rs
DimitrisJim: could also have a top level `make_deque_iter` function that takes a `reversed` flag. It might be a good refactoring-first issue for somebody on your open source contributing team.

DimitrisJim: Merged this for now, if you can assign the refactoring to someone you're guiding that would be great, if not, tell me and fix it up.

PR #2888, opened 2021-08-15
Removing obsolete files from stdlib

jamestwebber:
This is a WIP. I just noticed that the copy of the stdlib was likely copied from 3.7 and some things were gradually updated. Some of the files in there are no longer present in later versions so I should I'd open a PR to just remove them. But some of these will require additional modifications (i.e. they were being imported in older versions of other modules) so I still need to check what is broken before making this a real PR.


coolreader18:
I know at the very least I'd like to keep `[_]dummy_thread[ing].py` - cpython has dropped the ability to disable threading, but we still target platforms like wasm where threading isn't available. Though maybe at some point it'd be easier to just stub out `_thread` on platforms where it's not supported.

jamestwebber:
> I know at the very least I'd like to keep `[_]dummy_thread[ing].py` - cpython has dropped the ability to disable threading, but we still target platforms like wasm where threading isn't available.

Hm interesting, that's a good point. And maybe this is an overly broad change anyway.

I had started by just checking for python modules that had changed between 3.7 and 3.8 / 3.9 / 3.10 and then noticed some of them are just gone at this point. As it seems like RustPython is still aiming at a moving target re: compatibility, I figured I'd remove these to start and see what happened.

jamestwebber:
> platforms like wasm where threading isn't available

Although: is this still true in this specific case? I'm not a wasm expert but it appears it has threads now?

fanninpm:
> Although: is this still true in this specific case? I'm not a wasm expert but it appears it has threads now?

It depends on how recent your browser is. This exact feature is not on caniuse.com, but [there's an open issue in the caniuse repo.](https://github.com/Fyrd/caniuse/issues/5200)

jamestwebber:
Coming up on a year anniversary, I think this is quite out of date. Closing but might revisit someday.

PR #2890, opened 2021-08-16
Allow enumerate as base class.

DimitrisJim:


DimitrisJim:
Error unrelated, seems like version for `bitflags` needs to be bumped.

youknowone (approved):

PR #2893, opened 2021-08-16
Fix extendleft method of deque

whgusdn321:


vm/src/stdlib/collections.rs
```diff
@@ -205,10 +205,24 @@ mod _collections {
...
+                    let mut deque = self.borrow_deque_mut();
+                    let drain_until = deque.len().saturating_sub(max_len - elements.len());
+                    deque.drain(..drain_until);
+                } else {
+                    self.borrow_deque_mut().clear();
+                    elements.drain(..(elements.len() - max_len));
+                }
             }
+            self.borrow_deque_mut().extend(elements);
+            self.borrow_deque_mut().rotate_right(elements_len);
```
youknowone, line 225: I don't have a concrete idea but `rotate_right` looks like a redundant operation. Probably something like this maybe possible.
```suggestion
            let created = VecDeque::from(elements);
            let borrowed = self.borrow_deque_mut();
            created.extend(borrowed.drain(..));
            std::mem::swap(&mut created, borrowed);
```

youknowone: i have no idea about VecDeque internal, so not confident even this is better. @DimitrisJim do you have what's the best for this?

But borrowing can be merged to reduce the locking cost.

```suggestion
            let mut deque = self.borrow_deque_mut()
            deque.extend(elements);
            deque.rotate_right(elements_len);
```

DimitrisJim: Replace `created.extend(borrowed.drain())` with `created.append(borrowed)` (since append does exactly that) and use that. Using `append` here is better than continuously swapping elements around with `rotate_right`.

vm/src/stdlib/collections.rs
```diff
@@ -205,10 +205,24 @@ mod _collections {
...
+            let max_len = self.maxlen;
+            let mut elements: Vec<PyObjectRef> = vm.extract_elements(&iter)?;
+            elements.reverse();
+            let elements_len = elements.len();
+
+            if let Some(max_len) = max_len {
+                if max_len > elements.len() {
+                    let mut deque = self.borrow_deque_mut();
+                    let drain_until = deque.len().saturating_sub(max_len - elements.len());
+                    deque.drain(..drain_until);
```
DimitrisJim, line 218: This should be a `truncate` call. Recall, its the opposite of `extend`, so you need to trim of elements from the end. A `truncate_until` similar to `drain_until` needs to be calculated that denotes how many elements must be removed *from the front* to make room for `elements`.

youknowone (approved):

vm/src/stdlib/collections.rs
```diff
@@ -205,10 +205,25 @@ mod _collections {
...
+            elements.reverse();
+
+            if let Some(max_len) = max_len {
+                if max_len > elements.len() {
+                    let mut deque = self.borrow_deque_mut();
+                    let truncate_until = max_len - elements.len();
+                    deque.truncate(truncate_until);
+                } else {
+                    self.borrow_deque_mut().clear();
+                    elements.drain(..(elements.len() - max_len));
```
DimitrisJim, line 220: I'm sorry I missed this last time (I thought I had written that both should be changed). This should be `elements.truncate(max_len)` and not a `drain`.

We've reversed the elements and have them in a correct order, if `max_len > elements.len()` we only need to keep the first `max_len` items in `elements` i.e `truncate(max_len)`.

DimitrisJim (changes requested):

DimitrisJim (approved):

PR #2894, opened 2021-08-16
Fix failed test pow in test_math

owljoa:
add two conditions to raise ValueError for input values

case 1: NaN, when both input values are finite, and base is negative and exponent is non-integer
case 2: divde by zero, when base is 0 exponent is negative value

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -131,8 +131,21 @@ fn math_log1p(x: IntoPyFloat) -> f64 {
 make_math_func!(math_log2, log2);
 make_math_func!(math_log10, log10);
 
-fn math_pow(x: IntoPyFloat, y: IntoPyFloat) -> f64 {
-    x.to_f64().powf(y.to_f64())
+fn math_pow(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResult<f64> {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/math.rs
```diff
@@ -131,8 +131,21 @@ fn math_log1p(x: IntoPyFloat) -> f64 {
...
+        return Err(vm.new_value_error("math domain error".to_owned()));
+    }
+
+    if x == 0.0 && y < 0.0 {
+        return Err(vm.new_value_error("math domain error".to_owned()));
+    }
+
+    let value = x.powf(y);
+
+    return Ok(value);
```
youknowone, line 148:
```suggestion
    Ok(value)
```
the last `return` statement in a block can be replaced by an expression of its value.

owljoa: Thank you for suggestion : )
Should I add a commit for this suggestion? or rewrite my commit?

youknowone: that's up to you. either amending + push force or commiting suggestion would be the same result when it will be merged by squash merging.

owljoa: I got it, thanks for explaining!

youknowone (approved):

youknowone:
thank you for contributing!

PR #2896, opened 2021-08-16
Support unicode array type.

### LINKED ISSUE #2895: String index not correct
## Feature

RustPython
```
>>>>> '日月abc'.index('a')
6
>>>>>
```

Cpython
```
>>> '日月abc'.index('a')
2
>>> 
```

It is probably because the `str.find` in Rust returns the index of byte, instead of the code point.

I'll try fix this.

## Python Documentation

https://docs.python.org/3.8/library/stdtypes.html?highlight=str#str.index

whjpji:
fix #2895

### DIFF
--- a/vm/src/builtins/memory.rs
+++ b/vm/src/builtins/memory.rs
@@ -54,8 +54,7 @@ type PyMemoryViewRef = PyRef<PyMemoryView>;
 #[pyimpl(with(Hashable, Comparable, AsBuffer))]
 impl PyMemoryView {
     fn parse_format(format: &str, vm: &VirtualMachine) -> PyResult<FormatSpec> {
-        FormatSpec::parse(format)
-            .map_err(|msg| vm.new_exception_msg(vm.ctx.types.memoryview_type.clone(), msg))
+        FormatSpec::parse(format, vm)
     }
 
     pub fn from_buffer(
--- a/vm/src/builtins/pystr.rs
+++ b/vm/src/builtins/pystr.rs
@@ -780,6 +780,12 @@ impl PyStr {
         self.value.py_join(iter)
     }
 
+    // FIXME: two traversals of str is expensive
+    #[inline]
+    fn _to_char_idx(r: &str, byte_idx: usize) -> usize {
+        r[..byte_idx].chars().count()
+    }
+
     #[inline]
     fn _find<F>(&self, args: FindArgs, find: F) -> Option<usize>
     where
@@ -791,25 +797,25 @@ impl PyStr {
 
     #[pymethod]
     fn find(&self, args: FindArgs) -> isize {
-        self._find(args, |r, s| r.find(s))
+        self._find(args, |r, s| Some(Self::_to_char_idx(r, r.find(s)?)))
             .map_or(-1, |v| v as isize)
     }
 
     #[pymethod]
     fn rfind(&self, args: FindArgs) -> isize {
-        self._find(args, |r, s| r.rfind(s))
+        self._find(args, |r, s| Some(Self::_to_char_idx(r, r.rfind(s)?)))
             .map_or(-1, |v| v as isize)
     }
 
     #[pymethod]
     fn index(&self, args: FindArgs, vm: &VirtualMachine) -> PyResult<usize> {
-        self._find(args, |r, s| r.find(s))
+        self._find(args, |r, s| Some(Self::_to_char_idx(r, r.find(s)?)))
             .ok_or_else(|| vm.new_value_error("substring not found".to_owned()))
     }
 
     #[pymethod]
     fn rindex(&self, args: FindArgs, vm: &VirtualMachine) -> PyResult<usize> {
-        self._find(args, |r, s| r.rfind(s))
+        self._find(args, |r, s| Some(Self::_to_char_idx(r, r.rfind(s)?)))
             .ok_or_else(|| vm.new_value_error("substring not found".to_owned()))
     }
 
--- a/vm/src/stdlib/array.rs
+++ b/vm/src/stdlib/array.rs
@@ -1,7 +1,7 @@
 use crate::buffer::{BufferOptions, PyBuffer, ResizeGuard};
 use crate::builtins::float::IntoPyFloat;
 use crate::builtins::list::{PyList, PyListRef};
-use crate::builtins::pystr::PyStrRef;
+use crate::builtins::pystr::{PyStr, PyStrRef};
 use crate::builtins::pytype::PyTypeRef;
 use crate::builtins::slice::PySliceRef;
 use crate::builtins::{PyByteArray, PyBytes};
@@ -15,28 +15,16 @@ use crate::function::OptionalArg;
 use crate::sliceable::{saturate_index, PySliceableSequence, PySliceableSequenceMut};
 use crate::slots::{AsBuffer, Comparable, Iterable, PyComparisonOp, PyIter};
 use crate::utils::Either;
-use crate::VirtualMachine;
 use crate::{
     IdProtocol, IntoPyObject, PyClassImpl, PyComparisonValue, PyIterable, PyObjectRef, PyRef,
     PyResult, PyValue, StaticType, TryFromObject, TypeProtocol,
 };
+use crate::{IntoPyResult, VirtualMachine};
 use crossbeam_utils::atomic::AtomicCell;
 use itertools::Itertools;
 use std::cmp::Ordering;
+use std::convert::TryFrom;
 use std::{fmt, os::raw};
-
-struct ArrayTypeSpecifierError {
-    _priv: (),
-}
-
-impl fmt::Display for ArrayTypeSpecifierError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(
-            f,
-            "bad typecode (must be b, B, u, h, H, i, I, l, L, q, Q, f or d)"
-        )
-    }
-}
 
 macro_rules! def_array_enum {
     ($(($n:ident, $t:ty, $c:literal, $scode:literal)),*$(,)?) => {
@@ -47,10 +35,10 @@ macro_rules! def_array_enum {
 
         #[allow(clippy::naive_bytecount, clippy::float_cmp)]
         impl ArrayContentType {
-            fn from_char(c: char) -> Result<Self, ArrayTypeSpecifierError> {
+            fn from_char(c: char) -> Result<Self, String> {
                 match c {
                     $($c => Ok(ArrayContentType::$n(Vec::new())),)*
-                    _ => Err(ArrayTypeSpecifierError { _priv: () }),
+                    _ => Err("bad typecode (must be b, B, u, h, H, i, I, l, L, q, Q, f or d)".into()),
                 }
             }
 
@@ -94,10 +82,10 @@ macro_rules! def_array_enum {
                 Ok(())
             }
 
-            fn pop(&mut self, i: usize, vm: &VirtualMachine) -> PyObjectRef {
+            fn pop(&mut self, i: usize, vm: &VirtualMachine) -> PyResult {
                 match self {
                     $(ArrayContentType::$n(v) => {
-                        v.remove(i).into_pyobject(vm)
+                        v.remove(i).into_pyresult(vm)
                     })*
                 }
             }
@@ -225,9 +213,11 @@ macro_rules! def_array_enum {
                 Ok(i)
             }
 
-            fn getitem_by_idx(&self, i: usize, vm: &VirtualMachine) -> Option<PyObjectRef> {
+            fn getitem_by_idx(&self, i: usize, vm: &VirtualMachine) -> PyResult<Option<PyObjectRef>> {
                 match self {
-                    $(ArrayContentType::$n(v) => v.get(i).map(|x| x.into_pyobject(vm)),)*
+                    $(ArrayContentType::$n(v) => {
+                        v.get(i).map(|x| x.into_pyresult(vm)).transpose()
+                    })*
                 }
             }
 
@@ -244,8 +234,8 @@ macro_rules! def_array_enum {
             fn getitem(&self, needle: Either<isize, PySliceRef>, vm: &VirtualMachine) -> PyResult {
                 match needle {
                     Either::A(i) => {
-                        self.idx(i, "array", vm).map(|i| {
-                            self.getitem_by_idx(i, vm).unwrap()
+                        self.idx(i, "array", vm).and_then(|i| {
+                            self.getitem_by_idx(i, vm).map(Option::unwrap)
                         })
                     }
                     Either::B(slice) => self.getitem_by_slice(slice, vm),
@@ -377,13 +367,8 @@ macro_rules! def_array_enum {
                 Ok(s)
             }
 
-            fn iter<'a>(&'a self, vm: &'a VirtualMachine) -> impl Iterator<Item = PyObjectRef> + 'a {
-                let mut i = 0;
-                std::iter::from_fn(move || {
-                    let ret = self.getitem_by_idx(i, vm);
-                    i += 1;
-                    ret
-                })
+            fn iter<'a, 'vm: 'a>(&'a self, vm: &'vm VirtualMachine) -> impl Iterator<Item = PyResult> + 'a {
+                (0..self.len()).map(move |i| self.getitem_by_idx(i, vm).map(Option::unwrap))
             }
 
             fn cmp(&self, other: &ArrayContentType) -> Result<Option<Ordering>, ()> {
@@ -404,7 +389,7 @@ macro_rules! def_array_enum {
 def_array_enum!(
     (SignedByte, i8, 'b', "b"),
     (UnsignedByte, u8, 'B', "B"),
-    // TODO: support unicode char
+    (PyUnicode, WideChar, 'u', "u"),
     (SignedShort, raw::c_short, 'h', "h"),
     (UnsignedShort, raw::c_ushort, 'H', "H"),
     (SignedInt, raw::c_int, 'i', "i"),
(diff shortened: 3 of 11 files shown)


vm/src/builtins/pystr.rs
```diff
@@ -767,6 +767,12 @@ impl PyStr {
         self.value.py_join(iter)
     }
 
+    // FIXME: two traversals of str is expensive
+    #[inline]
+    fn _to_char_idx(r: &str, byte_idx: usize) -> usize {
+        r.char_indices().take_while(|&(idx, _)| idx < byte_idx).count()
```
youknowone, line 773:
```suggestion
        r[..byte_idx].chars().count()
```
is this equivalent code?

vm/src/stdlib/array.rs
```diff
@@ -417,6 +418,9 @@ def_array_enum!(
     (Double, f64, 'd', "d"),
 );
 
+#[derive(Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Debug)]
+pub struct PyUnicode(libc::wchar_t);
```
youknowone, line 422: this is confusign name because `PyUnicode` indicates `unicode` in Python2 which is `str` in Python3.
This is just wchar. So `WideChar` will fit enough which as referred in CPython codes.

vm/src/stdlib/array.rs
```diff
@@ -464,6 +468,40 @@ fn f64_try_into_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<f
     IntoPyFloat::try_from_object(vm, obj).map(|x| x.to_f64())
 }
 
+impl ArrayElement for PyUnicode {
+    fn try_into_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
+        let ch = PyStrRef::try_from_object(vm, obj)?
+            .as_str()
+            .chars()
+            .nth(0)
```
youknowone, line 476:
```suggestion
            .next()
```
otherwise if it must be single character, `itertools::exactly_one`

youknowone (changes requested):
wow, finally we will get unicode array! great!

vm/src/stdlib/array.rs
whjpji: I got `PyUnicode` from [python3.8's doc](https://docs.python.org/3.8/library/array.html), where the C type is named `Py_UNICODE`. However, `Py_UNICODE` is deprecated and replaced by `wchar_t` in [python3.9](https://docs.python.org/3.9/library/array.html), so I agree with you to use `WideChar` making it clearer.

vm/src/builtins/pystr.rs
whjpji: Cool! Your suggestion looks much better!

vm/src/stdlib/array.rs
whjpji: Having this check (`exactly_one`) is nice, I raise a `ValueError` instead of panic.

whjpji: Tested with CPython:
```
>>> import array
>>> a = array.array('u', 'abcde')
>>> a.append('cd')
TypeError: array item must be unicode character
>>> 
```
It is better to throw a `TypeError`,

coolreader18:
One thing that's tricky about the `'u'` array is that it looks like it's encoded as either UTF-32 or UTF-16 depending on the size of wchar_t:

https://github.com/python/cpython/blob/main/Objects/unicodeobject.c#L3251-L3274

One code point could be split over 2 `WideChar`s in the array if its value is `> u16::MAX`, and as the code is now it just silently truncates the high bits of the value with the `ch as _`

whjpji:
> One thing that's tricky about the `'u'` array is that it looks like it's encoded as either UTF-32 or UTF-16 depending on the size of wchar_t: https://github.com/python/cpython/blob/main/Objects/unicodeobject.c#L3251-L3274
> 
> One code point could be split over 2 `WideChar`s in the array if its value is `> u16::MAX`, and as the code is now it just silently truncates the high bits of the value with the `ch as _`

From the CPython's code, I am a little confused. Should `SIZEOF_WCHAR_T == sizeof(wchar_t)` always be true?

My understanding is that `sizeof(wchar_t)` and `SIZEOF_WCHAR_T` can be either 2 or 4 depending on machines, and the cases are
```Rust
match (SIZEOF_WCHAR_T, sizeof(wchar_t)) {
    (2, 2) | (4, 4) =>  /* case [1] */,
    (4, 2) => /* case [2]  */
    (2, 4) => /* case [3] */
}
```

In RustPython, how can we deal with cases that `SIZEOF_WCHAR_T != sizeof(wchar_t)`? We do not have a macro `SIZEOF_WCHAR_T` and only have `sizeof(wchar_t)` from `libc`.

[1] https://github.com/python/cpython/blob/1512bc21d60f098a9e9f37b44a2f6a9b49a3fd4f/Objects/unicodeobject.c#L3230-L3241
```c
#if USE_UNICODE_WCHAR_CACHE
    const wchar_t *wstr = _PyUnicode_WSTR(unicode);
    if (wstr != NULL) {
        memcpy(w, wstr, size * sizeof(wchar_t));
        return;
    }
#else /* USE_UNICODE_WCHAR_CACHE */
    if (PyUnicode_KIND(unicode) == sizeof(wchar_t)) {
        memcpy(w, PyUnicode_DATA(unicode), size * sizeof(wchar_t));
        return;
    }
#endif /* USE_UNICODE_WCHAR_CACHE *
```
[2] https://github.com/python/cpython/blob/1512bc21d60f098a9e9f37b44a2f6a9b49a3fd4f/Objects/unicodeobject.c#L3252-L3256
```c
        assert(PyUnicode_KIND(unicode) == PyUnicode_2BYTE_KIND);
        const Py_UCS2 *s = PyUnicode_2BYTE_DATA(unicode);
        for (; size--; ++s, ++w) {
            *w = *s;
        }
```
[3] https://github.com/python/cpython/blob/1512bc21d60f098a9e9f37b44a2f6a9b49a3fd4f/Objects/unicodeobject.c#L3258-L3273
```c
        assert(PyUnicode_KIND(unicode) == PyUnicode_4BYTE_KIND);
        const Py_UCS4 *s = PyUnicode_4BYTE_DATA(unicode);
        for (; size--; ++s, ++w) {
            Py_UCS4 ch = *s;
            if (ch > 0xFFFF) {
                assert(ch <= MAX_UNICODE);
                /* encode surrogate pair in this case */
                *w++ = Py_UNICODE_HIGH_SURROGATE(ch);
                if (!size--)
                    break;
                *w = Py_UNICODE_LOW_SURROGATE(ch);
            }
            else {
                *w = ch;
            }
        }
```

coolreader18:
I don't think it does check `sizeof(wchar_t)`, and `SIZEOF_WCHAR_T` I believe should always be equivalent - maybe what you're confused on is a missing case, which was from the outer function which calls `unicode_copy_as_widechar`: https://github.com/python/cpython/blob/main/Objects/unicodeobject.c#L4235-L4267

CPython's string repr is kinda weird, so it's just checking whether the string is internally a `Vec<u8>`, `Vec<u16>`, or `Vec<u32>`. When they're converting to a wide string, it's obviously easy if the string is a `Vec<wchar_t>`, and a `Vec<u8>` can just be casted up from `u8 -> wchar_t`. The special case is converting from `u16 to wchar_t if wchar_t == u32` or `u32 to wchar_t if wchar_t == u16`. The first one is also just an upcasting, the second requires converting to UTF-16.

coolreader18:
Oh hey it'd actually be not too bad to do UTF-16 - [`char::decode_utf16`](https://doc.rust-lang.org/stable/core/char/fn.decode_utf16.html) and [`str::encode_utf16`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.encode_utf16)/[`char::encode_utf16`](https://doc.rust-lang.org/1.54.0/std/primitive.char.html#method.encode_utf16) seem like they'd make this pretty painless.

whjpji:
@coolreader18 Thanks for your explanation, and that's what I've learnt from the CPython's code:

In CPython, a string can be either encoded in UTF-8, USC-1, USC-2 or USC-4 depends on its content.

However, in RustPython, a string is always encoded in UTF-8. (as the `PyStr::value`'s type is `Box<str>`). So we need no conversions inside the unicode array (`Vec<wchar_t>`). We only need conversions during initialization (from `Box<str>`) or the explicit call of `fromunicode`/`tounicode` (where the arguments are `Box<str>`).

I have another question:

If I call `'a.insert(-1, '\U0001FF00')` where `a` is an array of unicode and `sizeof(wchar_t) == 2`, should we throw exception (and what kind of exception) or just truncate the value from the lower 2 bytes?

I have no such environment that `sizeof(wchar_t) == 2` so I cannot test how CPython will behavior. Do you have any ideas?

whjpji:
> Oh hey it'd actually be not too bad to do UTF-16 - [`char::decode_utf16`](https://doc.rust-lang.org/stable/core/char/fn.decode_utf16.html) and [`str::encode_utf16`](https://doc.rust-lang.org/stable/std/primitive.str.html#method.encode_utf16)/[`char::encode_utf16`](https://doc.rust-lang.org/1.54.0/std/primitive.char.html#method.encode_utf16) seem like they'd make this pretty painless.

Note that CPython splits a USC-4 to two USC-2s, which is different from the behavior of `char::encode_utf16`. We may not use this function.

Maybe we can add a function like this?
```Rust
fn to_usc16(ch: char) -> impl Iterator<Item = u16> {
    let ch = ch as u32;
    let mut usc16 = [0_u16; 2];
    if ch > u16::MAX as u32 {
        usc16[0] = (ch >> 16) as u16;
        usc16[1] = ch as u16;
        usc16.into_iter()
    } else {
        usc16[0] = ch as u16;
        usc16[..1].into_iter()
    }
}
```

whjpji:
Now all the test cases have passed :)

whjpji:
Fix clippy tips.

vm/src/exceptions.rs
```diff
@@ -488,6 +488,7 @@ pub struct ExceptionZoo {
     pub indentation_error: PyTypeRef,
     pub tab_error: PyTypeRef,
     pub system_error: PyTypeRef,
+    pub struct_error: PyTypeRef,
```
youknowone, line 491: is this same error as the `struct.struct_error`? Then `pystruct.rs` has `new_struct_error`.

If not, still we were keeping this place only for built-in exceptions until now. Maybe adding built-in modules exception here could be a nice idea. But I want to know if it could affect to compatibility somehow.

vm/src/stdlib/array.rs
```diff
@@ -464,6 +467,40 @@ fn f64_try_into_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<f
...
+    }
+
+    fn byteswap(self) -> Self {
+        Self(self.0.swap_bytes())
+    }
+}
+
+impl From<WideChar> for char {
+    fn from(ch: WideChar) -> Self {
+        // safe because every configuration of bytes for the types we support are valid
```
youknowone, line 487: I am not sure if it is actully safe. Unfortunately, our rust string implementation for python string has a few problems. One of them is this kind of thing.
Rust char only take valid unicode codepoints as its characters, while python is generally not. So if there isn't addiotional guarantee of validity.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2032ef87f8862e128d3f7150cb9d9365

vm/src/stdlib/array.rs
```diff
@@ -584,6 +621,71 @@ impl PyArray {
         }
     }
 
+    #[pymethod]
+    fn fromunicode(zelf: PyRef<Self>, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        let str = PyStrRef::try_from_object(vm, obj.clone()).map_err(|_| {
```
youknowone, line 626: how about avoiding `str` not to be confused with Rust `str`?

youknowone (changes requested):
Looks great in general. I left a few comments

vm/src/builtins/memory.rs
```diff
@@ -54,8 +54,7 @@ type PyMemoryViewRef = PyRef<PyMemoryView>;
 #[pyimpl(with(Hashable, Comparable, AsBuffer))]
 impl PyMemoryView {
     fn parse_format(format: &str, vm: &VirtualMachine) -> PyResult<FormatSpec> {
-        FormatSpec::parse(format)
-            .map_err(|msg| vm.new_exception_msg(vm.ctx.types.memoryview_type.clone(), msg))
+        FormatSpec::parse(format).map_err(|msg| vm.new_struct_error(msg))
```
whjpji, line 57: @youknowone the `StructError` is used here. I added this because I ran the UTs before adding any patches and got "all the exceptions thrown must be derived from 'BaseException` (I ccouldn't remember the exactly error info).
I tested it on CPython:
```
>>> import array
>>> a = array.array('u', 'abc')
>>> x = memoryview(a)
>>> x.format
'w'
>>> x[0]
NotImplementedError: memoryview: format w not supported
```
The behaviour is very different from RustPython.

Before I added `FormatType::WideChar = b'u'`, it just threw an exception when creating the `memoryview`.

youknowone: The original code looks really weird but it also doesn't look like struct error. If you are looking for `NotImplementedError`, it is `vm.new_not_implemented_error`

whjpji: I found where the format `'u'` come from:
https://github.com/python/cpython/blob/3240bc62f4e0afa09964f3afc845697f0a0806b9/Modules/arraymodule.c#L2595-L2599

From the commit message,
> Close #13072: Restore code before the PEP 393 for the array module
> 
> 'u' format of the array module uses again Py_UNICODE type for backward compatibility with Python 3.2.
> ...

I found this issue: https://bugs.python.org/issue13072

Should we support this compatibility? What if the macro `Py_UNICODE_WIDE` is not defined?

My opinion is to modify the type of `PyMemoryView::format_spec` to `PyResult<FormatSpec>`, and then we can throw the error when the `memoryview` is actually used (pack/unpack), instead of when it is created.

vm/src/stdlib/array.rs
whjpji: Thanks for carefully checking.

I tested with CPython, and here is the result:
```
>>> a = array.array('u', '\ud800')
>>> a[0]
'\ud800'
>>> print(a[0])
UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800' in position 0: surrogates not allowed
```
Maybe we should `impl TryFrom<WideChar> for char` instead of `From<WideChar>`, and then we can throw an `UnicodeEncodeError` if the `u32` value is not a valid unicode point.

whjpji: I found another bug with string:

CPython:
```
>>> a = '\ud800'
>>> a[0]
'\ud800'
>>> print(a[0])
UnicodeEncodeError: 'utf-8' codec can't encode character '\ud800' in position 0: surrogates not allowed
```

RustPython
```
>>>>> a = '\ud800'
>>>>> a[0]
'�'
>>>>> print(a[0])
�
```

vm/src/stdlib/array.rs
whjpji: No problem.

vm/src/stdlib/array.rs
youknowone: `TryFrom<WideChar>` sounds like a good idea. We'd better to overhaul string implementation one day...

vm/src/builtins/memory.rs
whjpji: The latest CPython's doc says that the `array('u')` type is
> Deprecated since version 3.3, will be removed in version 4.0.

See https://docs.python.org/3.11/library/array.html.

I think CPython maintains `array('u')` only for compatibility consideration, and they might not change its behavior because early or late it will be removed.

There are two possible solutions for us:
1. leave the memory view created from `array('u')` unaccepted and throws `NotImplementedError`, but it is not compatible with CPython,
2. implement the memory view format `'u'` ourselves, and let it exposes `array('u')` as a `[wchar_t]` slice.

Which one do you prefer?

youknowone: I vote for 2. And I think this PR is almost doing it :D

whjpji: Yes it is, because I didn't know that CPython does not support formatting of `array('u')`.

Now I have fixed the problems you mentioned before.

Lib/test/test_urllib2.py
```diff
@@ -1304,6 +1304,8 @@ def test_redirect_fragment(self):
         fp = o.open('http://www.example.com')
         self.assertEqual(fp.geturl(), redirected_url.strip())
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
whjpji, line 1308: There are two tests failed on my computer, so I marked them `expectedFailure`:
```
FAIL: test_redirect_encoding (test.test_urllib2.HandlerTests) [b'/spaced path/']
  File ".../vm/pylib-crate/Lib/test/test_urllib2.py", line 1358, in test_redirect_encoding
    self.assertTrue(request.startswith(expected), repr(request))
AssertionError: False is not true : b'GET http://example.com/spaced%20path/ HTTP/1.1\r\nAccept-Encoding: identity\r\nHost: example.com\r\nUser-Agent: Python-urllib/3.9\r\nConnection: close\r\n\r\n'

FAIL: test_redirect_encoding (test.test_urllib2.HandlerTests) [b'/?p\xc3\xa5-dansk']
  File ".../vm/pylib-crate/Lib/test/test_urllib2.py", line 1358, in test_redirect_encoding
    self.assertTrue(request.startswith(expected), repr(request))
AssertionError: False is not true : b'GET http://example.com/?p%C3%A5-dansk HTTP/1.1\r\nAccept-Encoding: identity\r\nHost: example.com\r\nUser-Agent: Python-urllib/3.9\r\nConnection: close\r\n\r\n'
```
But it seems to pass on the GitHub CI.

My OS is Ubuntu 20.04, and the CPU is 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz.

My rustc version is 1.53.0, and the test command is
```
cargo run --release -- -m test -j8 -v
```

Is there anyone who can reproduce this error?

vm/src/stdlib/array.rs
```diff
@@ -417,6 +402,9 @@ def_array_enum!(
     (Double, f64, 'd', "d"),
 );
 
+#[derive(Copy, Clone, Ord, PartialOrd, Eq, PartialEq, Debug)]
+pub struct WideChar(libc::wchar_t);
```
whjpji, line 406: WASM seems to have no `libc::wchar_t`.

Should I add a cfg like this?
```Rust
#[cfg(target_arch = "wasm32")]
pub struct WideChar(u32);
```

youknowone: It looks good

youknowone (approved):

youknowone:
Thank you for contributing!

PR #2897, opened 2021-08-16
Adds better exception macro

### LINKED ISSUE #2771: `Exception.__init__.__qualname__` is incorrectly set to `'BaseException.__init__'`
## Feature

CPython:

```pycon
>>> Exception.__init__.__qualname__
'Exception.__init__'
```

RustPython:

```
>>>>> Exception.__init__.__qualname__
'BaseException.__init__'
```

sobolevn:
Closes #2771

I've changed a single exception to show my progress. If this is ok, then I can change all of them.

### DIFF
--- a/derive/src/pymodule.rs
+++ b/derive/src/pymodule.rs
@@ -109,6 +109,7 @@ fn new_module_item(
             inner: ContentItemInner { index, attr_name },
             pyattrs: pyattrs.unwrap_or_else(Vec::new),
         }),
+        "pyexception" => unreachable!("#[pyexception] {:?}", pyattrs.unwrap_or_else(Vec::new)),
         other => unreachable!("#[pymodule] doesn't accept #[{}]", other),
     }
 }
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -526,16 +526,12 @@ macro_rules! extends_exception {
     // This case is like `SimpleExtendsException`:
     (
         $class_name: ident,
-        $class_name_literal: literal,
         $base_class: ident,
-        $base_class_literal: literal,
         $docs: tt
     ) => {
         extends_exception!(
             $class_name,
-            $class_name_literal,
             $base_class,
-            $base_class_literal,
             $docs,
             {}
         );
@@ -546,13 +542,11 @@ macro_rules! extends_exception {
     // This case is like `ComplexExtendsException`:
     (
         $class_name: ident,
-        $class_name_literal: literal,
         $base_class: ident,
-        $base_class_literal: literal,
         $docs: tt,
         { $($attr_name:expr => $attr_value:expr),* $(,)* }
     ) => {
-        #[pyclass(module = false, name = $class_name_literal, base = $base_class_literal)]
+        #[pyexception($class_name, $base_class)]
         #[derive(Debug)]
         struct $class_name {}
 
@@ -585,11 +579,9 @@ macro_rules! extends_exception {
 }
 
 extends_exception! {
-  PyKeyboardInterrupt,
-  "KeyboardInterrupt",
-  PyBaseException,
-  "PyBaseException",
-  "Program interrupted by user."
+    PyKeyboardInterrupt,
+    PyBaseException,
+    "Program interrupted by user."
 }
 
 impl ExceptionZoo {
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -531,7 +531,7 @@ macro_rules! extends_exception {
         $base_class_literal: literal,
         $docs: tt
     ) => {
-        simple_extends_exception!(
+        extends_exception!(
             $class_name,
             $class_name_literal,
             $base_class,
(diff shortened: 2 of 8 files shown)


vm/src/exceptions.rs
```diff
@@ -522,13 +522,84 @@ pub fn create_exception_type(name: &str, base: &PyTypeRef) -> PyTypeRef {
...
+                $(
+                    class.set_str_attr($attr_name, $attr_value);
+                )*
+            }
+        }
+    };
+}
+
+extends_exception! {
+    PyKeyboardInterrupt,
```
sobolevn, line 588: What I don't like here is that I have to repeat `ident` and `literal` names. Why?
- I cannot use just `ident` names with `#[pyclass(name = ..., base = ...)]`, because they explicitly need `literal` strings. I even cannot use `stringify!`, because it is not expanded in `#[pyclass]` attributes
- I cannot use just `literal`, because Rust cannot `unstrigify!` things, unless https://docs.rs/unstringify/0.1.1/unstringify/macro.unstringify.html is installed

vm/src/exceptions.rs
```diff
@@ -522,13 +522,84 @@ pub fn create_exception_type(name: &str, base: &PyTypeRef) -> PyTypeRef {
...
+}
+
 impl ExceptionZoo {
     pub(crate) fn init() -> Self {
         let base_exception_type = PyBaseException::init_bare_type().clone();
 
         // Sorted By Hierarchy then alphabetized.
         let system_exit = create_exception_type("SystemExit", &base_exception_type);
-        let keyboard_interrupt = create_exception_type("KeyboardInterrupt", &base_exception_type);
+        // let keyboard_interrupt = create_exception_type("KeyboardInterrupt", &base_exception_type);
```
sobolevn, line 601: Comment is here to show the diff.

vm/src/exceptions.rs
```diff
@@ -522,13 +522,84 @@ pub fn create_exception_type(name: &str, base: &PyTypeRef) -> PyTypeRef {
...
+                $(
+                    class.set_str_attr($attr_name, $attr_value);
+                )*
+            }
+        }
+    };
+}
+
+extends_exception! {
+  PyKeyboardInterrupt,
```
sobolevn, line 588: Btw, here's another example why I need #2876

vm/src/exceptions.rs
sobolevn: Friendly ping @youknowone @DimitrisJim

youknowone: I don't any a nice idea for this. This problem is known to be a tranditional rust macro problem. So it looks just ok to me for now. Will it be better if pyclass takes both ident and str for its macro arguments? @coolreader18

sobolevn: I think so, `#[pyclass(name = :ident | :literal, base = :ident | :literal)]` seems like a good solution 👍 
I will implement it in a separate PR.

sobolevn: Looks like that there's nothing I can do. It is a limitation of `pub type AttributeArgs = Vec<NestedMeta>;`, where `NestedMeta` can be:

```rust
    /// ## Path
    ///
    /// A meta path is like the `test` in `#[test]`.
    ///
    /// ## List
    ///
    /// A meta list is like the `derive(Copy)` in `#[derive(Copy)]`.
    ///
    /// ## NameValue
    ///
    /// A name-value meta is like the `path = "..."` in `#[path =
    /// "sys/windows.rs"]`.
```

And `NameValue` only allows `Lit` values 😞 
So, I guess that the only way is to pass `ident` and `literal` as two separate params.

sobolevn: https://github.com/dtolnay/syn/blob/5a9efb0b772b583af93cdca9453fa9fd4e522641/src/lit.rs#L745

sobolevn: Hm `#[pyexception($class_name, $base_class)]` seem to work, I will continue my experiments 👍

derive/src/pymodule.rs
```diff
@@ -109,6 +109,7 @@ fn new_module_item(
             inner: ContentItemInner { index, attr_name },
             pyattrs: pyattrs.unwrap_or_else(Vec::new),
         }),
+        "pyexception" => unreachable!("#[pyexception] {:?}", pyattrs.unwrap_or_else(Vec::new)),
```
sobolevn, line 112: Test that it is not counted in `module`, I can leave it or remove it.

sobolevn:
@youknowone looks like a nice solution to me, what do you think?

sobolevn:
Am I moving in the right direction? 🙂

sobolevn:
I am asking because I am also thinking about this alternative and easier solution from https://github.com/RustPython/RustPython/pull/2897/commits/a725e2b8cff2af9f57d8503bb3025c50c90e8745

It just enhances `extend_class!` macro: https://github.com/RustPython/RustPython/pull/2897/files#diff-d19484f26ba252a3b3ff27867db9682f948572dd37e83e811d6f073bfc2f07c2R646-R657

youknowone:
@coolreader18 could you review this PR?

coolreader18 (approved):
LGTM. I've been thinking for a while about what to do with our inheritance situation, and though we're not there yet with e.g. OSError having fields of its own as well as Exception's, this will make it easy to add once we get there.

sobolevn:
Oh no! My bad, I should've marked it as "WIP".
It is totally not ready to be mreged!

1. It has **two** competing solutions: [`extends_exception!`](https://github.com/RustPython/RustPython/pull/2897/files#diff-d19484f26ba252a3b3ff27867db9682f948572dd37e83e811d6f073bfc2f07c2R536) macro and [`extend_exception!`](https://github.com/RustPython/RustPython/pull/2897/files#diff-d19484f26ba252a3b3ff27867db9682f948572dd37e83e811d6f073bfc2f07c2R646) macro
2. It has some temporary code inside

Considering `1.` which is better in your opinion?

PR #2900, opened 2021-08-17
Add reversed iterator type.

DimitrisJim:
`PySequenceIterator` handles forward iteration while `PyReverseSequenceIterator`, located in `enumerate.rs` (not sure why) as in CPython, handles reverse iteration.


vm/src/builtins/enumerate.rs
```diff
@@ -60,6 +61,52 @@ impl PyIter for PyEnumerate {
     }
 }
 
+#[pyclass(module = false, name = "reversed")]
+#[derive(Debug)]
+pub struct PyReverseSequenceIterator {
+    pub position: AtomicCell<isize>,
```
youknowone, line 67: is this isize rather than a usize?

vm/src/builtins/enumerate.rs
```diff
@@ -60,6 +61,52 @@ impl PyIter for PyEnumerate {
...
+
+    #[pymethod(magic)]
+    fn length_hint(&self) -> PyResult<isize> {
+        Ok(self.position.load() + 1)
+    }
+}
+
+impl PyIter for PyReverseSequenceIterator {
+    fn next(zelf: &PyRef<Self>, vm: &VirtualMachine) -> PyResult {
+        let pos = zelf.position.fetch_sub(1);
```
youknowone, line 94: what happens if this is called forever until negatively overflow?

DimitrisJim: I'm fixing all these methods in another pr. Currently I've just split these appart for now to make the changes a little smaller in size.

vm/src/builtins/enumerate.rs
```diff
@@ -60,6 +61,52 @@ impl PyIter for PyEnumerate {
     }
 }
 
+#[pyclass(module = false, name = "reversed")]
+#[derive(Debug)]
+pub struct PyReverseSequenceIterator {
```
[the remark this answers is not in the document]
DimitrisJim: I know. I don't know why CPython has them in the same module but it does.

vm/src/builtins/enumerate.rs
youknowone: Will you make another huge PR with the whole diff for a reference? I will also check the final result.

DimitrisJim: I could push the changes on this branch if it would make it easier for you.

vm/src/builtins/enumerate.rs
DimitrisJim: Usually the reverse iterators keep an `isize` in order to check that its `> 0` while decreasing it and indexing the underlying object.

vm/src/builtins/enumerate.rs
youknowone: I think seperated small PR still have benefits. So another PR would be more convenient. It will be reusable after merging this one.

For this PR, I thought I can compare other parts also, but after checking your answers about it, it doesn't look that's nessessary for now.

youknowone (approved):

PR #2902, opened 2021-08-17
make dict remain order after delete

eldpswp99:
closes #2550

before:
       
- dict is insertion order in CPython, but in RustPython the order was broken when  deletion is occured

after:

- it follows the CPython way. when it is deleted, make the entry to None, and remove this None only if resize occurs.

- added nentries to optimize popitem. it is equal to the index for next insertion (= last valid entry index + 1)

### DIFF
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -70,6 +70,8 @@ struct DictInner<T> {
     filled: usize,
     indices: Vec<i64>,
     entries: Vec<Option<DictEntry<T>>>,
+    // index to new inserted element should be in entries
+    nentries: usize,
 }
 
 impl<T: Clone> Clone for Dict<T> {
@@ -88,6 +90,7 @@ impl<T> Default for Dict<T> {
                 filled: 0,
                 indices: vec![IndexEntry::FREE; 8],
                 entries: Vec::new(),
+                nentries: 0,
             }),
         }
     }
@@ -107,6 +110,7 @@ pub struct DictSize {
     entries_size: usize,
     used: usize,
     filled: usize,
+    nentires: usize,
 }
 
 struct GenIndexes {
@@ -163,6 +167,7 @@ impl<T> DictInner<T> {
             }
         }
         self.filled = self.used;
+        self.nentries = self.entries.len();
     }
 
     fn unchecked_push(
@@ -179,10 +184,15 @@ impl<T> DictInner<T> {
             value,
             index,
         };
-        let entry_index = self.entries.len();
-        self.entries.push(Some(entry));
+        let entry_index = self.nentries;
+        if self.entries.len() == entry_index {
+            self.entries.push(Some(entry));
+        } else {
+            self.entries[entry_index] = Some(entry);
+        }
         self.indices[index] = entry_index as i64;
         self.used += 1;
+        self.nentries += 1;
         if let IndexEntry::Free = index_entry {
             self.filled += 1;
             if let Some(new_size) = self.should_resize() {
@@ -197,6 +207,7 @@ impl<T> DictInner<T> {
             entries_size: self.entries.len(),
             used: self.used,
             filled: self.filled,
+            nentires: self.nentries,
         }
     }
 
@@ -313,6 +324,7 @@ impl<T: Clone> Dict<T> {
             inner.indices.resize(8, IndexEntry::FREE);
             inner.used = 0;
             inner.filled = 0;
+            inner.nentries = 0;
             // defer dec rc
             std::mem::take(&mut inner.entries)
         };
@@ -596,7 +608,7 @@ impl<T: Clone> Dict<T> {
 
     pub fn pop_back(&self, vm: &VirtualMachine) -> Option<(PyObjectRef, T)> {
         let inner = self.read();
-        let mut idx: i64 = inner.size().entries_size as i64 - 1;
+        let mut idx: i64 = inner.nentries as i64 - 1;
 
         while idx >= 0 && inner.entries.get(idx as usize).unwrap().is_none() {
             idx -= 1;
@@ -612,7 +624,11 @@ impl<T: Clone> Dict<T> {
             let removed_item = entry.value;
             if let Ok(lookup) = self.lookup(vm, &removed_key, entry.hash, Some(inner)) {
                 match self.pop_inner(lookup) {
-                    Ok(_) => Some((removed_key, removed_item)),
+                    Ok(_) => {
+                        let mut inner = self.write();
+                        inner.nentries = idx as usize;
+                        Some((removed_key, removed_item))
+                    }
                     Err(_) => None,
                 }
             } else {
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -25,6 +25,7 @@ type EntryIndex = usize;
 pub struct Dict<T = PyObjectRef> {
     inner: PyRwLock<DictInner<T>>,
 }
+
 impl<T> fmt::Debug for Dict<T> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("Debug").finish()
@@ -37,10 +38,12 @@ enum IndexEntry {
     Free,
     Index(usize),
 }
+
 impl IndexEntry {
     const FREE: i64 = -1;
     const DUMMY: i64 = -2;
 }
+
 impl From<i64> for IndexEntry {
     fn from(idx: i64) -> Self {
         match idx {
@@ -50,6 +53,7 @@ impl From<i64> for IndexEntry {
         }
     }
 }
+
 impl From<IndexEntry> for i64 {
     fn from(idx: IndexEntry) -> Self {
         match idx {
@@ -65,7 +69,7 @@ struct DictInner<T> {
     used: usize,
     filled: usize,
     indices: Vec<i64>,
-    entries: Vec<DictEntry<T>>,
+    entries: Vec<Option<DictEntry<T>>>,
 }
 
 impl<T: Clone> Clone for Dict<T> {
@@ -143,15 +147,19 @@ impl<T> DictInner<T> {
         self.indices = vec![IndexEntry::FREE; new_size];
         let mask = (new_size - 1) as i64;
         for (entry_idx, entry) in self.entries.iter_mut().enumerate() {
-            let mut idxs = GenIndexes::new(entry.hash, mask);
-            loop {
-                let index_index = idxs.next();
-                let idx = &mut self.indices[index_index];
-                if *idx == IndexEntry::FREE {
-                    *idx = entry_idx as i64;
-                    entry.index = index_index;
-                    break;
+            if let Some(entry) = entry {
+                let mut idxs = GenIndexes::new(entry.hash, mask);
+                loop {
+                    let index_index = idxs.next();
+                    let idx = &mut self.indices[index_index];
+                    if *idx == IndexEntry::FREE {
+                        *idx = entry_idx as i64;
+                        entry.index = index_index;
+                        break;
+                    }
                 }
+            } else {
+                //removed entry
             }
         }
         self.filled = self.used;
@@ -172,7 +180,7 @@ impl<T> DictInner<T> {
             index,
         };
         let entry_index = self.entries.len();
-        self.entries.push(entry);
+        self.entries.push(Some(entry));
         self.indices[index] = entry_index as i64;
         self.used += 1;
         if let IndexEntry::Free = index_entry {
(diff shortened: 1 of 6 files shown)


DimitrisJim:
Fix the passing test in `test_ordered_dict` (`test_del`) and I'll re-run the ci, not sure why the ubuntu check isn't finishing.

eldpswp99:
I think there might be an infinite loop in the ubuntu test. 
When i working on this commit, there was an infinite loop when implementing pop_back. when i have a inner with write lock and call lookup with lock None. but i fixed with making inner to read lock and passing it to lookup. the loop removed. but i'm not sure why the ci doesn't finished in ubuntu particularly.

DimitrisJim:
It's weird that it only occurs there. I'll try and look into it.

vm/src/dictdatatype.rs
```diff
@@ -453,10 +477,18 @@ impl<T: Clone> Dict<T> {
...
+        loop {
+            let entry = inner.entries.get(*position);
+            if let Some(entry) = entry {
+                *position += 1;
+                if let Some(entry) = entry {
+                    break Some((entry.key.clone(), entry.value.clone()));
+                }
+            } else {
+                break None;
+            }
```
youknowone, line 490:
```suggestion
            let entry = inner.entries.get(*position)?;
            *position += 1;
            if let Some(entry) = entry {
                break Some((entry.key.clone(), entry.value.clone()));
            }
```

vm/src/dictdatatype.rs
```diff
@@ -540,23 +577,19 @@ impl<T: Clone> Dict<T> {
             return Ok(None);
         };
         let mut inner = self.write();
-        if matches!(inner.entries.get(entry_index), Some(entry) if entry.index == index_index) {
+        if matches!(inner.entries.get(entry_index), Some(entry) if entry.as_ref().unwrap().index == index_index)
```
youknowone, line 580:
```suggestion
        if matches!(inner.entries.get(entry_index), Some(Some(entry)) if entry.index == index_index)
```

vm/src/dictdatatype.rs
```diff
@@ -540,23 +577,19 @@ impl<T: Clone> Dict<T> {
...
-            let last_index = inner.entries.last().unwrap().index;
-            let removed = inner.entries.swap_remove(entry_index);
-            inner.indices[last_index] = entry_index as i64;
-            removed
-        };
-        Ok(Some(removed))
+        let removed = inner.entries[entry_index].clone();
+        inner.entries[entry_index] = None;
+
+        Ok(Some(removed.unwrap()))
```
youknowone, line 592:
```suggestion
        let removed = std::mem::take(&mut inner.entries[entry_index]);
        Ok(removed)
```
`std::mem` has a few tools to avoid clone

vm/src/dictdatatype.rs
```diff
@@ -573,13 +606,35 @@ impl<T: Clone> Dict<T> {
...
-            inner.indices[entry.index] = IndexEntry::DUMMY;
-            (entry.key, entry.value)
-        })
+    pub fn pop_back(&self, vm: &VirtualMachine) -> Option<(PyObjectRef, T)> {
+        let inner = self.read();
+        let mut idx: i64 = inner.nentries as i64 - 1;
+
+        while idx >= 0 && inner.entries.get(idx as usize).unwrap().is_none() {
+            idx -= 1;
+        }
```
youknowone, line 615: is this nessessary if `nentries` is being keeped maintained well?

vm/src/dictdatatype.rs
```diff
@@ -573,13 +606,35 @@ impl<T: Clone> Dict<T> {
...
+                        let mut inner = self.write();
+                        inner.nentries = idx as usize;
+                        Some((removed_key, removed_item))
+                    }
+                    Err(_) => None,
+                }
+            } else {
+                None
+            }
+        }
```
youknowone, line 637: gramatical suggestion:
```suggestion
        let mut idx = inner.nentries.checked_sub(1)?;

        while inner.entries.get(idx as usize).unwrap().is_none() {
            idx = idx.checked_sub(1)?;
        }

        let entry = inner.entries.get(idx).unwrap().as_ref().unwrap().clone();

        let removed_key = entry.key;
        let removed_item = entry.value;
        let lookup = self.lookup(vm, &removed_key, entry.hash, Some(inner)).ok()?;
        self.pop_inner(lookup).ok()?;
        self.write().nentries = idx as usize;
        Some((removed_key, removed_item))
```

But this implementation looks a bit weird for me.
1. do we need to copy the entry?
2. do we need to lookup to pop_back?

youknowone (changes requested):

vm/src/dictdatatype.rs
```diff
@@ -266,6 +286,7 @@ impl<T: Clone> Dict<T> {
             if let IndexEntry::Index(index) = entry {
                 let inner = self.read();
                 if let Some(entry) = inner.entries.get(index) {
+                    let entry = entry.as_ref().unwrap();
```
youknowone, line 289: I found this is a very common pattern. A helper function will be useful.

vm/src/dictdatatype.rs
```diff
@@ -189,6 +207,7 @@ impl<T> DictInner<T> {
             entries_size: self.entries.len(),
             used: self.used,
             filled: self.filled,
+            nentires: self.nentries,
```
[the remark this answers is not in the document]
eldpswp99: this field was added to make popitem fast. if there is no nentries it costs O(N^2) for popping all elements of the dict with n entries. [..., None, None, None] kind of entries cause inefficient work without nentries.

if entry is like [..,valid entry, None, None, None] , then nentries points to the left None, popitem would pick the valid entry at once. also, if i insert one entry, this would make the entires to [..,valid entry, valid entry, None, None] which make popitem remains fast

vm/src/dictdatatype.rs
```diff
@@ -573,13 +606,35 @@ impl<T: Clone> Dict<T> {
...
 
-    pub fn pop_back(&self) -> Option<(PyObjectRef, T)> {
-        let mut inner = self.write();
-        inner.entries.pop().map(|entry| {
-            inner.used -= 1;
-            inner.indices[entry.index] = IndexEntry::DUMMY;
-            (entry.key, entry.value)
-        })
+    pub fn pop_back(&self, vm: &VirtualMachine) -> Option<(PyObjectRef, T)> {
+        let inner = self.read();
```
youknowone, line 610: this line is the cause of deadlock. this `read` lock is not removed while calling `pop_inner` which make `write` lock

youknowone (changes requested):

vm/src/dictdatatype.rs
eldpswp99: since pop(key) can make entry to None in the middle of the entries. it is necessary

vm/src/dictdatatype.rs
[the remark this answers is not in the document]
youknowone: I think you are talking about `DictInner::nentries`. I agree that's a clever way to do that. But this one is `DictSize::nentires`.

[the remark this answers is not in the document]
eldpswp99: ah, that's not useful now. but i thought that size function should give the size information about the dict and nentries is about the valid entries length, so i included on the size function. if it is better to remove the unused field, then i'll remove it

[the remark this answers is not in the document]
youknowone: Let's go without it for for now. I think we can add it when we actually can use it later. It will be straightforward because it doesn't have any trick but just copying the field.

vm/src/dictdatatype.rs
youknowone: So, what happens if we repeat to delete last item and add new item one by one? Does it infinitely grow instead of reusing the last entry?

eldpswp99: if we use pop(key) or delete to delete the last item, it is infinitely grow. but when inserting, the resize calls if the should_resize returns true. this will remove None entries and the infinite grow would not happen that much i think. I can make the delete function to detect whether the deleted item was the last entry, but this can also happen when deleting the first and insert again. I think it's a problem about resize function.

vm/src/dictdatatype.rs
eldpswp99: 1. only removed_key and removed_item should be cloned, but i copied entry to use clone once and I thought it is more clean to see the code. if it is better to clone only the two things, i'll fix that
2. It is totally redundant. i'll fix that

youknowone: For the clones, I don't think we need any clone for this operation in best solution.
(though I didn't looked in it that enough yet)

youknowone: I think the entry should be moved out rather than being cloned. Please check `std::mem::take` part of other comment.

vm/src/dictdatatype.rs
eldpswp99: @youknowone do i have to fix resize also? or left to it another issue?

vm/src/dictdatatype.rs
```diff
@@ -223,6 +240,10 @@ impl<T: Clone> Dict<T> {
                 let mut inner = self.write();
                 // Update existing key
                 if let Some(entry) = inner.entries.get_mut(index) {
+                    let entry = entry.as_mut().unwrap_or_else(|| unsafe {
+                        // SAFETY: `unwrap()` is guaranteed to success because `index` is guaranteed to refer existing entry
+                        std::hint::unreachable_unchecked()
```
coolreader18, line 245: I'm not sure this is guaranteed - couldn't something happen in between `lookup()` and `write()`?

vm/src/dictdatatype.rs
youknowone: I am not sure about the best solution. At the same time, that also look like not a strict purpose of this PR. So vote for leaving it for later.

vm/src/dictdatatype.rs
coolreader18: Though I suppose it could still happen before this PR too, but I'd just be hesitant about using `unreachable_unchecked` - there are already cases where dict panics when there's multiple threads accessing one.

youknowone: keeping these future problems as unwraps to fix them later make sense, but `unreachable_unchecked` still panic in Debug. I don't think we can keep everything bare `unwrap` for release mode forever.

vm/src/dictdatatype.rs
```diff
@@ -575,11 +602,19 @@ impl<T: Clone> Dict<T> {
...
+            idx = idx.checked_sub(1)?;
+        }
+
+        let entry = std::mem::take(&mut inner.entries[idx]).unwrap();
+        let removed_key = entry.key;
+        let removed_item = entry.value;
+
+        inner.used -= 1;
+        inner.indices[entry.index] = IndexEntry::DUMMY;
+        inner.nentries = idx as usize;
```
youknowone, line 616:
```suggestion
        inner.nentries = idx;
```

vm/src/dictdatatype.rs
eldpswp99: for this case, then rather than using unwrap_or_else, use unwrap to make panic? I'm not sure about the solution of resolving the multi thread issue. Can i remain it to another issue?

youknowone: Yes, @coolreader18 's comment is correct. Something like `expect("index is changed before retrieving")` will be even better.

eldpswp99: i chose the message "// The dict was changed since we did lookup. Let's try again." which is used for comment in code

vm/src/dictdatatype.rs
```diff
@@ -223,6 +240,9 @@ impl<T: Clone> Dict<T> {
                 let mut inner = self.write();
                 // Update existing key
                 if let Some(entry) = inner.entries.get_mut(index) {
+                    let entry = entry
+                        .as_mut()
+                        .expect("The dict was changed since we did lookup. Let's try again.");
```
youknowone, line 245: `expect` is a sort of `panic`. So it will not try again anything. Just `"The dict was changed since we did lookup."` will be enough

vm/src/dictdatatype.rs
```diff
@@ -65,7 +69,9 @@ struct DictInner<T> {
     used: usize,
     filled: usize,
     indices: Vec<i64>,
-    entries: Vec<DictEntry<T>>,
+    entries: Vec<Option<DictEntry<T>>>,
+    // index to new inserted element should be in entries
+    nentries: usize,
```
youknowone, line 74:
```suggestion
    next_new_entry_idx: usize,
```
I found `nentries` is used in different meaning from CPython.

eldpswp99: I thought it has similar work in pop_back. i would fix that.

youknowone (approved):

eldpswp99:
Thank you for all the reviews...!

Issue #2680, opened 2021-05-30
Built-in methods/classes/functions lack documentation.


DimitrisJim:
Most (?) of the built in functionality currently lacks corresponding documentation. The functionality to grab and attach the docs is there, the docs just haven't been added. 

This, I believe, could make a good first issue for many people, it just involves taking the documentation and adding it before (after works too but most current examples add it before) the `#[py(function/method/class)]` attribute. As an example, for `list.append` one could just add:

```rust
/// Append object to the end of the list.
#[pymethod]
pub(crate) fn append(&self, x: PyObjectRef) { ... }
```

The main built-in classes seem to have documentation for the classes but not for the methods. Builtin modules, I'd assume, probably also have many more cases.

AthulMuralidhar:
this seems pretty straightforward. Can i start working on this on? @youknowone @DimitrisJim ?

DimitrisJim:
Go ahead @AthulMuralidhar!

nandhinianandj:
https://github.com/RustPython/RustPython/pull/3435 -- very first commit.. just getting used to process here.

youknowone:
how about closing this issue by preference for #3101?

PR #2904, opened 2021-08-18
Introduce `__text_signature__` property

### LINKED ISSUE #2410: Builtin functions and methods lack a .__text_signature__ property
On CPython:

```py
>>> import inspect, struct
>>> inspect.signature(struct.unpack_from)
<Signature (format, /, buffer, offset=0)>
```

Running this under the debugger (`import pdb; pdb.runcall(inspect.signature, struct.unpack_from)`), you see that `inspect` does ` getattr(struct.unpack_from, "__text_signature__")` which is defined in CPython [here](https://github.com/python/cpython/blob/0ec57e25c918b859b9f8d464e34e0ac859c2f8b3/Objects/methodobject.c#L193), and that's not implemented in RustPython, so you get this instead:

```py
>>>>> import inspect, struct
>>>>> inspect.signature(struct.unpack_from)
  File "/home/boris/code/RustPython/vm/pylib-crate/Lib/inspect.py", line 3093, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/home/boris/code/RustPython/vm/pylib-crate/Lib/inspect.py", line 2843, in from_callable
    follow_wrapper_chains=follow_wrapped)
  File "/home/boris/code/RustPython/vm/pylib-crate/Lib/inspect.py", line 2297, in _signature_from_callable
    skip_bound_arg=skip_bound_arg)
  File "/home/boris/code/RustPython/vm/pylib-crate/Lib/inspect.py", line 2107, in _signature_from_builtin
    raise ValueError("no signature found for builtin {!r}".format(func))
ValueError: no signature found for builtin <builtin_function_or_method object at 0x55f59d0558d0>
```

moreal:
This pull request tries to resolve #2410 

This pull request does:
- Fill 'object' type's missing docstrings 0b52d0c 
- Introduce new `__text_signature__` property to `PyType`, `PyBuiltinMethod`, `PyBuiltinFunction`. 40ab1c4 

## Code to verify compatibility

```python
print(object.__str__.__text_signature__)  # expected `($self, /)`
print(object.__text_signature__)  # expected `()`
print(type.__text_signature__)  # expected `None` (no output)
```

## References

- parse `__text_signature__` logic: https://github.com/python/cpython/blob/fa919fdf2583bdfead1df00e842f24f30b2a34bf/Objects/typeobject.c#L183-L203

### DIFF
--- a/vm/src/builtins/builtinfunc.rs
+++ b/vm/src/builtins/builtinfunc.rs
@@ -147,6 +147,7 @@ impl PyBuiltinFunction {
     fn text_signature(&self) -> Option<String> {
         self.value.doc.as_ref().and_then(|doc| {
             pytype::get_text_signature_from_internal_doc(self.value.name.as_str(), doc.as_str())
+                .map(|signature| signature.to_string())
         })
     }
 }
@@ -218,6 +219,7 @@ impl PyBuiltinMethod {
     fn text_signature(&self) -> Option<String> {
         self.value.doc.as_ref().and_then(|doc| {
             pytype::get_text_signature_from_internal_doc(self.value.name.as_str(), doc.as_str())
+                .map(|signature| signature.to_string())
         })
     }
     #[pymethod(magic)]
--- a/vm/src/builtins/builtinfunc.rs
+++ b/vm/src/builtins/builtinfunc.rs
@@ -1,6 +1,7 @@
 use std::fmt;
 
 use super::classmethod::PyClassMethod;
+use super::pytype;
 use crate::builtins::pystr::PyStrRef;
 use crate::builtins::pytype::PyTypeRef;
 use crate::function::{FuncArgs, PyNativeFunc};
@@ -142,6 +143,16 @@ impl PyBuiltinFunction {
     fn repr(&self) -> String {
         format!("<built-in function {}>", self.value.name)
     }
+    #[pyproperty(magic)]
+    fn text_signature(&self, vm: &VirtualMachine) -> PyObjectRef {
+        match self.value.doc.as_ref()
+            .and_then(|doc| pytype::get_text_signature_from_internal_doc(
+                self.value.name.as_str(),
+                doc.as_str())) {
+            Some(signature) => vm.ctx.new_str(signature),
+            None => vm.ctx.none(),
+        }
+    }
 }
 
 // `PyBuiltinMethod` is similar to both `PyMethodDescrObject` in
@@ -206,6 +217,16 @@ impl PyBuiltinMethod {
     #[pyproperty(magic)]
     fn doc(&self) -> Option<PyStrRef> {
         self.value.doc.clone()
+    }
+    #[pyproperty(magic)]
+    fn text_signature(&self, vm: &VirtualMachine) -> PyObjectRef {
+        match self.value.doc.as_ref()
+            .and_then(|doc| pytype::get_text_signature_from_internal_doc(
+                self.value.name.as_str(),
+                doc.as_str())) {
+            Some(signature) => vm.ctx.new_str(signature),
+            None => vm.ctx.none(),
+        }
     }
     #[pymethod(magic)]
     fn repr(&self) -> String {
--- a/vm/src/builtins/builtinfunc.rs
+++ b/vm/src/builtins/builtinfunc.rs
@@ -144,13 +144,10 @@ impl PyBuiltinFunction {
         format!("<built-in function {}>", self.value.name)
     }
     #[pyproperty(magic)]
-    fn text_signature(&self, vm: &VirtualMachine) -> PyObjectRef {
-        match self.value.doc.as_ref().and_then(|doc| {
+    fn text_signature(&self) -> Option<String> {
+        self.value.doc.as_ref().and_then(|doc| {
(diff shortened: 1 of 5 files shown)


moreal:
This pull request seems to resolve https://github.com/RustPython/RustPython/issues/2680 in some part 👀

vm/src/builtins/builtinfunc.rs
```diff
@@ -142,6 +143,15 @@ impl PyBuiltinFunction {
     fn repr(&self) -> String {
         format!("<built-in function {}>", self.value.name)
     }
+    #[pyproperty(magic)]
+    fn text_signature(&self, vm: &VirtualMachine) -> PyObjectRef {
```
youknowone, line 147:
```suggestion
    fn text_signature(&self, vm: &VirtualMachine) -> Option<String> {
```
With this signature, `None` will turn into Python `None` and `Some(String)` will turn into Python `str`

youknowone (changes requested):

DimitrisJim:
This is nice! It does introduce an issue though, `__doc__` now contains that text signature in its contents, you'll need to tweak doc to ignore the signature at the start (similar to CPythons `_PyType_DocWithoutSignature`).

vm/src/builtins/builtinfunc.rs
moreal: I tried to apply it in d12d286 commit. Could you check it again?

moreal:
> This is nice! It does introduce an issue though, `__doc__` now contains that text signature in its contents, you'll need to tweak doc to ignore the signature at the start (similar to CPythons `_PyType_DocWithoutSignature`).

Yeah, I missed it to comment. 😅 I made it as an issue #2907.

vm/src/builtins/pytype.rs
```diff
@@ -517,6 +517,47 @@ impl PyType {
             "Setting __dict__ attribute on a type isn't yet implemented".to_owned(),
         ))
     }
+
+    #[pyproperty(magic)]
+    fn text_signature(&self) -> Option<String> {
+        let doc_string = self.get_attr("__doc__");
+        let doc_string: Option<PyStrRef> = doc_string.and_then(|o| o.downcast().ok());
+        doc_string.and_then(|doc| {
+            get_text_signature_from_internal_doc(self.name().as_str(), doc.as_str())
```
youknowone, line 526:
```suggestion
        let doc: PyStrRef = doc_string.and_then(|o| o.downcast().ok())?;
        get_text_signature_from_internal_doc(self.name().as_str(), doc.as_str())
```

youknowone (changes requested):
nice work, I left a few comments about code style.

vm/src/builtins/pytype.rs
```diff
@@ -517,6 +517,47 @@ impl PyType {
             "Setting __dict__ attribute on a type isn't yet implemented".to_owned(),
         ))
     }
+
+    #[pyproperty(magic)]
+    fn text_signature(&self) -> Option<String> {
+        let doc_string = self.get_attr("__doc__");
```
youknowone, line 523: If you are interested in, I feel like `tp_doc` which should be added in `PyTypeSlots` is related to improvement of this line.

vm/src/builtins/pytype.rs
```diff
@@ -517,6 +517,47 @@ impl PyType {
...
+        return None;
+    }
+
+    let doc = doc[name.len()..].to_owned();
+    if !doc.starts_with('(') {
+        None
+    } else {
+        Some(doc)
+    }
+}
```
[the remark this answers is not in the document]
moreal: > I feel like `find_signature` should return the first index of string of signature, but this function returns docs starting with signature paranthesis.
> 
> My suggestions:
> 
>     1. if this function is just once used by `get_text_signature_from_internal_doc`, it doesn't look like to be an independent function.
> 
>     2. If this function is planned to be used from somewhere, giving it a better name or comments will be a good idea. For now, I don't have a good name suggestion.
> 
> 
> by the way, here is my suggestion for the code itself.

Thanks for suggestions. I looked CPython implementation again and there is the usage of `find_signature` and `skip_signature` at `_PyType_DocWithoutSignature` (see also #2907). Current `skip_signature` implementation is incorrect and cannot be reused in `_PyType_DocWithoutSignature`. So I will implement the functions to return `index` not `str` like you said your feel 🙏🏻 .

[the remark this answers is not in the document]
youknowone: Oh, I didn't know the name was CPython referenced, I am sorry to confuse you. The CPython code looks like intending `Option<&str>`. So keep going with it.

I got the feeling from the convention of Rust library, but CPython has its own convention. So never mind about it.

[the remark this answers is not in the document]
moreal: As a note, I fixed `skip_signature` implementation but it is not used in this pull request so I removed it. And I implemented `get_signature` renamed from `skip_signature`. (89c8c6e, 34d5215)

vm/src/builtins/pytype.rs
moreal: I tried to apply your suggestion in 78ff936. Could you check that it was you said?

youknowone (approved):

youknowone:
it looks great, thank you for contributing!

PR #2905, opened 2021-08-18
Skip tests that cause RustPython to hang.

DimitrisJim:
Temporarily so that other PRs don't get stalled until its fixed. Note that these don't hang consistently, they do hang often, though.

### DIFF
--- a/Lib/test/lock_tests.py
+++ b/Lib/test/lock_tests.py
@@ -128,6 +128,7 @@ class BaseLockTests(BaseTestCase):
         self.assertFalse(result[0])
         lock.release()
 
+    @unittest.skip("TODO: RUSTPYTHON, sometimes hangs")
     def test_acquire_contended(self):
         lock = self.locktype()
         lock.acquire()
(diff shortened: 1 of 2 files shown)


DimitrisJim:
`test_deadlock` is only sometimes hanging. From a quick run I've done it hangs once every 10-15 runs. The other two are more common, hanging at 1 out of ~5 runs.

Lib/test/lock_tests.py
```diff
@@ -128,6 +128,7 @@ def f():
         self.assertFalse(result[0])
         lock.release()
 
+    @unittest.skip("TODO: RUSTPYTHON hangs")
```
youknowone, line 131:
```suggestion
    @unittest.skip("TODO: RUSTPYTHON occasionally hangs")
```

Probably the word `occasional` will be helpful to prevent removing skip again before fixing it correctly

youknowone (dismissed):

Lib/test/lock_tests.py
youknowone: ah, @fanninpm already did it in different PR. how about just pulling it?

fanninpm: @youknowone I'm okay with you and/or @DimitrisJim cherry-picking those commits.

youknowone (approved):

PR #2911, opened 2021-08-19
Implement methods for `PyReverseSequenceIterator`

Snowapril:
This revision includes:
* Implement `length_hint`, `reduce`, `setstate` and `next` for `PyReverseSequenceIterator`
* Remove `PyStrReverseIterator` to make `PyStr` type use general `PyReverseSequenceIterator`
* Removes `expectedFailure` annotations in [`test_enumerate.py`](https://github.com/RustPython/RustPython/blob/master/Lib/test/test_enumerate.py#L177), [`test_tuple.py`](https://github.com/RustPython/RustPython/blob/master/Lib/test/test_tuple.py#L366)

By these changes, `TestReversed.test_len`, `TestReversed.test_pickle`  and `TupleTest.test_reversed_pickle` tests became successful now.

Also, below python codes are now available by this PR
#### Before
```python
>>>>> a = reversed('hello')
>>>>> operator.length_hint(a)
0
>>>>> a.__reduce__()
(<function _reconstructor at 0x55e7c6094470>, (<class 'reversed'>, <class 'object'>, None))
>>>>> a.__setstate__
AttributeError: 'str_reverseiterator' object has no attribute '__setstate__'
```
#### After
```python
>>>>> a = reversed('hello')
>>>>> operator.length_hint(a)
5
>>>>> a.__reduce__()
(<built-in function reversed>, ('hello',), 2)
>>>>> a.__setstate__(2)
>>>>> a.__reduce__()
(<built-in function reversed>, ('hello',), 2)
```

#### Before
```python
>>>>> a = reversed(tuple('hello'))
>>>>> operator.length_hint(a)
5
>>>>> a.__setstate__
AttributeError: 'reversed' object has no attribute '__setstate__'
>>>>> a.__reduce__()
(<function _reconstructor at 0x55e7c6094470>, (<class 'reversed'>, <class 'object'>, None))
```
#### After
```python
>>>>> a = reversed(tuple('hello'))
>>>>> operator.length_hint(a)
5
>>>>> a.__reduce__()
(<built-in function reversed>, (('h', 'e', 'l', 'l', 'o'),), 4)
>>>>> a.__setstate__(2)
>>>>> a.__reduce__()
(<built-in function reversed>, (('h', 'e', 'l', 'l', 'o'),), 2)
```

Except above example codes, general object that implements required interface can use `PyReverseSequenceIterator` and its methods.


vm/src/builtins/enumerate.rs
```diff
@@ -76,32 +82,90 @@ impl PyValue for PyReverseSequenceIterator {
...
+                    tp_new(
+                        vm,
+                        FuncArgs {
+                            args: vec![obj_type.clone().into_object()],
+                            kwargs: IndexMap::new(),
+                        },
+                    )?
+                } else {
+                    vm.ctx.new_list(vec![])
+                };
```
DimitrisJim, line 142: An empty tuple is returned when the iterator is exhausted. See i.e:

```python
r = reversed("foo"); list(r); r.__reduce__()       # (reversed, ((),))
r = reversed((1, 2, 3)); list(r); r.__reduce__()   # (reversed, ((),))

class Foo:
    def __len__(self):
        return 10
    def __getitem__(self, i):
        if i < 10:
            return i
        raise StopIteration

r = reversed(Foo()); list(r); r.__reduce__()  # (reversed, ((),))
```

vm/src/builtins/enumerate.rs
```diff
@@ -76,32 +82,90 @@ impl PyValue for PyReverseSequenceIterator {
...
-                    Err(vm.new_stop_iteration())
-                }
-                // also catches stop_iteration => stop_iteration
-                ret => ret,
+        if pos == 0 {
+            zelf.status.store(Exhausted);
+        }
+        match zelf.obj.get_item(pos, vm) {
+            Err(ref e) if e.isinstance(&vm.ctx.exceptions.index_error) => {
+                Err(vm.new_stop_iteration())
```
DimitrisJim, line 165: set the status here too. If it returns `StopIteration` once it should continue doing so.

DimitrisJim (changes requested):

vm/src/builtins/enumerate.rs
Snowapril: Oh. you are right. I thought it must return empty object with given type as deque object in cpython work like below.
```python
>>> a = reversed(deque())
>>> list(a)
[]
>>> a.__reduce__()
(<class '_collections._deque_reverse_iterator'>, (deque([]), 0))
```
But there exist specialized methods for deque, not a general one. This is my miss. I sent v2 code.
Thanks!

vm/src/builtins/enumerate.rs
Snowapril: Thanks. I missed it. I applied it in v2 code 😊

DimitrisJim (approved):
@youknowone pinging you here because remember you had some feedback when I added this type back in #2900. Overall this lgtm.

youknowone:
@DimitrisJim yes, I found this PR is solving infinite sub problem, thanks!

PR #2915, opened 2021-08-19
adjust dict reverse iterator next to new dict entries structure

eldpswp99:
dict entries was changed due to remain order even when delete any element. #2902 

in this PR, i only adjust dict reverse iterator next function to this new dict entries structure.

however, there remains some issue about detecting the change while iteration.
before, the size change was equivalent to the change of the dictionary entries, so it was easy to detect it
however, making deletion not to erase entries, and insertion occurs on next_new_entry_idx, it is hard to detect change during iteration.
all iterators with dict may have this issue.

### DIFF
--- a/vm/src/builtins/dict.rs
+++ b/vm/src/builtins/dict.rs
@@ -753,7 +753,7 @@ macro_rules! dict_iterator {
         impl $reverse_iter_name {
             fn new(dict: PyDictRef) -> Self {
                 $reverse_iter_name {
-                    position: AtomicCell::new(1),
+                    position: AtomicCell::new(0),
                     size: dict.size(),
                     dict,
                     status: AtomicCell::new(IterStatus::Active),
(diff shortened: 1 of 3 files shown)


vm/src/builtins/dict.rs
```diff
@@ -782,10 +782,10 @@ macro_rules! dict_iterator {
...
                             ));
                         }
-                        let count = zelf.position.fetch_add(1);
-                        match zelf.dict.len().checked_sub(count) {
-                            Some(mut pos) => {
-                                let (key, value) = zelf.dict.entries.next_entry(&mut pos).unwrap();
+                        let mut position = zelf.position.load();
+                        match zelf.dict.entries.next_entry_reversed(&mut position) {
+                            Some((key, value)) => {
+                                zelf.position.store(position);
```
youknowone, line 788: `fetch_add` was atomic operation, but the combination of `load` and `store ` is not. Can we express this as consecutive `fetch_add`?

eldpswp99: i fixed this with only using fetch_add in next_entry_reversed. there is same problem in dict iterator, i would fix it in another PR since this PR is about reverse iterator.

youknowone (approved):

PR #2920, opened 2021-08-21
Add math.dist function and enabled its tests

voidsatisfaction:
I implemented `math.dist` function and enabled its tests.

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -232,7 +232,11 @@ fn vector_norm(v: &[f64], max: f64) -> f64 {
     max * f64::sqrt(csum - 1.0 + frac)
 }
 
-fn math_dist(p: PySequence<IntoPyFloat>, q: PySequence<IntoPyFloat>, vm: &VirtualMachine) -> PyResult<f64> {
+fn math_dist(
+    p: PySequence<IntoPyFloat>,
+    q: PySequence<IntoPyFloat>,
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/math.rs
```diff
@@ -232,6 +232,48 @@ fn vector_norm(v: &[f64], max: f64) -> f64 {
...
+    p: PySequence<IntoPyFloat>,
+    q: PySequence<IntoPyFloat>,
+    vm: &VirtualMachine,
+) -> PyResult<f64> {
+    let mut max = 0.0;
+    let mut has_nan = false;
+
+    let p = IntoPyFloat::vec_into_f64(p.into_vec());
+    let q = IntoPyFloat::vec_into_f64(q.into_vec());
+    let mut diffs = vec![];
```
DimitrisJim, line 245: pedantic but, as in CPython, this seems like it could be preallocated with capacity `max(p.len(), 16)` after the check for equal dimensions is made.

DimitrisJim (approved):

vm/src/stdlib/math.rs
youknowone:
```suggestion
    let mut diffs = Vec::with_capacity(std::cmp::max(p.len(), 16));
```
This code will be equivalent.

voidsatisfaction: Thank you for your kind comments!

PR #2922, opened 2021-08-21
Fixes typos in all other files

sobolevn:
Refs #2921


benches/benchmarks/pystone.py
```diff
@@ -16,7 +16,7 @@
 
 Version History:
 
-                Inofficial version 1.1.1 by Chris Arndt:
+                Unofficial version 1.1.1 by Chris Arndt:
```
youknowone, line 19: This file is not belong to RustPython project

youknowone (changes requested):

youknowone (approved):

Issue #2623, opened 2021-05-03
Missing dataclasses module


oroppas:
## Feature

[rich](https://github.com/willmcgugan/rich) depends on `dataclasses`

## Python Documentation

https://docs.python.org/3/library/dataclasses.html?highlight=dataclasses#module-dataclasses

moreal:
This issue is not able to solve until https://github.com/RustPython/RustPython/issues/2869 solved.

`dataclasses` module calls `inspect._signature_from_callable` and it compares `__new__` built-in method.
Though empty class doesn't have additional `__new__` method definition but RustPython implementation always bind new bound method for each types.

## References

https://github.com/RustPython/RustPython/blob/8cf3506f53bb3d7748d86bc6e20d6dd279458cce/vm/src/pyobject.rs#L391-L397

https://github.com/python/cpython/blob/fa919fdf2583bdfead1df00e842f24f30b2a34bf/Lib/inspect.py#L2367-L2368:
``` python
# _signature_from_callable
                if (obj.__init__ is object.__init__ and
                    obj.__new__ is object.__new__):
```

moreal:
Sadly, though #2869 was resolved, it doesn't work because there #2410 is not resolved  😿

```
Welcome to the magnificent Rust Python 0.1.2 interpreter 😱 🖖
>>>>> import dataclasses
>>>>> @dataclasses.dataclass
..... class A:
.....   s: str
..... 
Traceback (most recent call last):
  File "<stdin>", line 3, in <module>
  File "/home/moreal/github/RustPython/RustPython/vm/pylib-crate/Lib/dataclasses.py", line 1000, in dataclass
    return wrap(cls)
  File "/home/moreal/github/RustPython/RustPython/vm/pylib-crate/Lib/dataclasses.py", line 992, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash, frozen)
  File "/home/moreal/github/RustPython/RustPython/vm/pylib-crate/Lib/dataclasses.py", line 972, in _process_class
    str(inspect.signature(cls)).replace(' -> None', ''))
  File "/home/moreal/github/RustPython/RustPython/vm/pylib-crate/Lib/inspect.py", line 3093, in signature
    return Signature.from_callable(obj, follow_wrapped=follow_wrapped)
  File "/home/moreal/github/RustPython/RustPython/vm/pylib-crate/Lib/inspect.py", line 2843, in from_callable
    follow_wrapper_chains=follow_wrapped)
[...]

PR #2923, opened 2021-08-21
derive __text_signature__ from method definition

youknowone:
This PR is not perfectly solving the issue but showing the first step to go. I think we can keep the signature and actual function better from this way.

improving #2904 in automated way

cc @moreal

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -337,13 +337,51 @@ where
 
         let item_attr = args.attrs.remove(self.index());
         let item_meta = MethodItemMeta::from_attr(ident.clone(), &item_attr)?;
-
         let py_name = item_meta.method_name()?;
+
+        let sig_doc = args.item.function_or_method_impl().ok().map(|item| {
+            let sig = item.sig();
+            let args: Vec<_> = sig
+                .inputs
+                .iter()
+                .filter_map(|arg| {
+                    use syn::FnArg::*;
+                    let arg = match arg {
+                        Receiver(_) => return Some("$self".to_owned()),
+                        Typed(typed) => typed,
+                    };
+                    let ty = arg.ty.as_ref();
+                    let ty = quote!(#ty).to_string();
+                    if ty == "FuncArgs" {
+                        return Some("*args, **kwargs".to_owned());
+                    }
+                    if ty == "& VirtualMachine" {
+                        return None;
+                    }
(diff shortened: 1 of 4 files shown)


derive/src/pyclass.rs
```diff
@@ -337,13 +337,56 @@ where
...
+                .filter_map(|arg| {
+                    use syn::FnArg::*;
+                    let arg = match arg {
+                        Receiver(_) => return Some("$self".to_owned()),
+                        Typed(typed) => typed,
+                    };
+                    let ty = arg.ty.as_ref();
+                    let ty = quote!(#ty).to_string();
+                    if ty == "FuncArgs" {
+                        has_args = true;
```
DimitrisJim, line 357: `'/'` is used to separate positional-only parameters from positional-or-keyword ones (similarly `*` separates positional_or_keyword parameters from keyword only). I.e the `'/'` separator should be placed before the first `Optional` is encountered (or at the end if the function only has positionals). 

Since this is a first step, I say we remove `'/'` for now and fix it up when we cover more cases.

DimitrisJim:
In general I think this is a step in the right direction. Many builtins will be covered and those that aren't covered can just have the signature placed there manually (i.e this step would be nice if it was made *only* if a signature hasnt been specified already).

Another small detail on this: it now requires us to *try* and follow the names used in CPython more closely (though I'm sure most functions already do that).

youknowone:
Yes, that will be a new concern for argument names. Already generated `object.__eq__` uses `other` as the argument name but doc refers `value`.

vm/src/builtins/object.rs
```diff
@@ -14,9 +14,6 @@ use crate::{
     PyContext, PyObject, PyObjectRef, PyResult, PyValue, TryFromObject, TypeProtocol,
 };
 
-/// object()
-/// --
-///
```
moreal, line 19: As I understood, `ImplItem<Item> for MethodItem` will not affect at class documentation `ClassItem`. When I executed `object.__text_signature__`, it returns `None`. I guess these changes will revert some resolved problems again (https://github.com/RustPython/RustPython/issues/2732#issuecomment-898936219, https://github.com/RustPython/RustPython/issues/2623#issuecomment-898862717).

moreal (approved):

DimitrisJim (approved):

PR #2924, opened 2021-08-21
Fix __class_getitem__ slightly

coolreader18:
Because it was treating `__class_getitem__` as a special method, subscripting a type from `typing` would just always return `type` instead of some sort of `_GenericAlias`

### DIFF
--- a/Lib/test/test_genericclass.py
+++ b/Lib/test/test_genericclass.py
@@ -1,0 +1,286 @@
+import unittest
+from test import support
+
+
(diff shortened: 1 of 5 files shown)


youknowone:
cc @fanninpm

youknowone:
Do you think it is time to go to 3.9? It looks like the issue was also related to 3.8/3.9 problem

youknowone (approved):

PR #2927, opened 2021-08-22
make dict iterator use atomic operation

eldpswp99:
in #2915, there was a comment to use only fetch_add, not using load & store to maintain atomic operation.

since dict iterator has this load & store, change to use fetch_add.

### DIFF
--- a/vm/src/builtins/dict.rs
+++ b/vm/src/builtins/dict.rs
@@ -578,7 +578,7 @@ impl Iterator for DictIter {
     type Item = (PyObjectRef, PyObjectRef);
 
     fn next(&mut self) -> Option<Self::Item> {
-        self.dict.entries._next_entry(&mut self.position)
+        self.dict.entries.next_entry(&mut self.position)
     }
 
     fn size_hint(&self) -> (usize, Option<usize>) {
@@ -718,7 +718,7 @@ macro_rules! dict_iterator {
                                 "dictionary changed size during iteration".to_owned(),
                             ));
                         }
-                        match zelf.dict.entries.next_entry(&zelf.position) {
+                        match zelf.dict.entries.next_entry_atomic(&zelf.position) {
                             Some((key, value)) => Ok(($result_fn)(vm, key, value)),
                             None => {
                                 zelf.status.store(IterStatus::Exhausted);
@@ -778,7 +778,7 @@ macro_rules! dict_iterator {
                                 "dictionary changed size during iteration".to_owned(),
                             ));
                         }
-                        match zelf.dict.entries.next_entry_reversed(&zelf.position) {
+                        match zelf.dict.entries.next_entry_atomic_reversed(&zelf.position) {
                             Some((key, value)) => Ok(($result_fn)(vm, key, value)),
                             None => {
                                 zelf.status.store(IterStatus::Exhausted);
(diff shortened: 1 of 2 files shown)


vm/src/dictdatatype.rs
```diff
@@ -488,6 +488,17 @@ impl<T: Clone> Dict<T> {
         }
     }
 
+    pub fn next_entry(&self, position: &AtomicCell<usize>) -> Option<(PyObjectRef, T)> {
```
coolreader18, line 491: Maybe this could be named `next_entry_atomic`?

coolreader18 (approved):
Nice! This is definitely better.

vm/src/dictdatatype.rs
eldpswp99: Is it better to revert `_next_entry` to `next_entry` when I change the function name to `next_entry_atomic`?

vm/src/builtins/dict.rs
```diff
@@ -718,12 +718,8 @@ macro_rules! dict_iterator {
                                 "dictionary changed size during iteration".to_owned(),
                             ));
                         }
-                        let mut position = zelf.position.load();
-                        match zelf.dict.entries.next_entry(&mut position) {
-                            Some((key, value)) => {
-                                zelf.position.store(position);
-                                Ok(($result_fn)(vm, key, value))
-                            }
+                        match zelf.dict.entries.next_entry_atomic(&zelf.position) {
```
youknowone, line 721:
```suggestion
                        match zelf.dict.entries.next_entry(&zelf.position) {
```

vm/src/builtins/dict.rs
```diff
@@ -782,7 +778,7 @@ macro_rules! dict_iterator {
                                 "dictionary changed size during iteration".to_owned(),
                             ));
                         }
-                        match zelf.dict.entries.next_entry_reversed(&zelf.position) {
+                        match zelf.dict.entries.next_entry_atomic_reversed(&zelf.position) {
```
youknowone, line 781:
```suggestion
                        match zelf.dict.entries.next_entry_reversed(&zelf.position) {
```

vm/src/dictdatatype.rs
```diff
@@ -488,7 +488,21 @@ impl<T: Clone> Dict<T> {
         }
     }
 
-    pub fn next_entry_reversed(&self, position: &AtomicCell<usize>) -> Option<(PyObjectRef, T)> {
+    pub fn next_entry_atomic(&self, position: &AtomicCell<usize>) -> Option<(PyObjectRef, T)> {
```
youknowone, line 491:
```suggestion
    pub fn next_entry(&self, position: &AtomicCell<usize>) -> Option<(PyObjectRef, T)> {
```

vm/src/dictdatatype.rs
```diff
@@ -488,7 +488,21 @@ impl<T: Clone> Dict<T> {
...
+        loop {
+            let position_usize = position.fetch_add(1);
+            let entry = inner.entries.get(position_usize)?;
+            if let Some(entry) = entry {
+                break Some((entry.key.clone(), entry.value.clone()));
+            }
+        }
+    }
+
+    pub fn next_entry_atomic_reversed(
```
youknowone, line 502:
```suggestion
    pub fn next_entry_reversed(
```

Issue #2898, opened 2021-08-16
`dict_keys`, `dict_values`, and `dict_items` are missing several methods


sobolevn:
## Feature

CPython:

```python
>>> dir(dict().keys())
['__and__', '__class__', '__contains__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__iter__', '__le__', '__len__', '__lt__', '__ne__', '__new__', '__or__', '__rand__', '__reduce__', '__reduce_ex__', '__repr__', '__reversed__', '__ror__', '__rsub__', '__rxor__', '__setattr__', '__sizeof__', '__str__', '__sub__', '__subclasshook__', '__xor__', 'isdisjoint']
```

RustPython:

```python
>>>>> dir(dict().keys())
['__class__', '__delattr__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__iter__', '__le__', '__len__', '__lt__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__reversed__', '__setattr__', '__str__', '__subclasshook__']
```

Diff: 
- `'__contains__'`
-  `'__and__'`
- `'__rxor__'`
- `'isdisjoint'`
- `'__or__'`
- `'__rand__'` 
- `'__ror__'` 
- `'__xor__'`
- `'__rsub__'`
- `'__sub__'`
- `'__sizeof__'`

The same is for `dict_values` and `dict_items`

DimitrisJim:
Yup, I don't think a macro should be used to generate the types as is currently done. For their respective iterators, the macro seems like a good idea, for the types, not that much.

DimitrisJim:
Additionally, though `__contains__` currently works (falling back to iterating), it can be `O(1)` for `keys` and `items` views by keying the dict ref they hold.

youknowone:
is this done? can we list up what's done and left?

rvsampson94:
I think __sizeof__ is the last one that needs to be merged. After that it should be done.

nadavkluger:
Hey I want to start working on it but it seems like part of it is already implemented:
![image](https://user-images.githubusercontent.com/27895842/147653023-59f46d30-41ae-4b28-a82a-267c4fdb84a8.png)

[...]

PR #2929, opened 2021-08-22
prepare extension for each dict iterator types

youknowone:
The extensions for each type can be added into empty `#[pyimpl]` later

preparing #2898


DimitrisJim (approved):
This generally does lgtm, I personally couldn't think of a better way to get the `impl` out of the macro for `keys`, `items` and `values`.

PR #2930, opened 2021-08-22
Add missing `__str__` method in base types

Snowapril:
According to `whats_left.sh` script printing,
there is no `__str__` for base types(int, float, complex, bool)

Therefore, str call for these types go to `PyBaseObject.str`
`PyBaseObject.str` method implementation call `vm.to_repr(&zelf)`.
With this implementation, str call for these types will be slightly improved in performance.

#### Before
```python
>>>>> a = 10
>>>>> a.__str__
<bound method object.__str__ of 10>
>>>>> a = complex(10, 20)
>>>>> a.__str__
<bound method object.__str__ of (10+20j)>
>>>>> a = float(10)
>>>>> a.__str__
<bound method object.__str__ of 10.0>
```
#### After
```python
>>>>> a = 10
>>>>> a.__str__
<bound method int.__str__ of 10>
>>>>> a = complex(10, 20)
>>>>> a.__str__
<bound method object.__str__ of (10+20j)>
>>>>> a = 10.3
>>>>> a.__str__
<bound method float.__str__ of 10.3>
```

### DIFF
--- a/vm/src/builtins/complex.rs
+++ b/vm/src/builtins/complex.rs
@@ -220,6 +220,7 @@ impl PyComplex {
         -self.value
     }
 
+    #[pymethod(name = "__str__")]
     #[pymethod(magic)]
     fn repr(&self) -> String {
         let Complex64 { re, im } = self.value;
--- a/vm/src/builtins/float.rs
+++ b/vm/src/builtins/float.rs
@@ -359,6 +359,7 @@ impl PyFloat {
         self.simple_op(other, |a, b| Ok(b - a), vm)
     }
 
+    #[pymethod(name = "__str__")]
     #[pymethod(magic)]
     fn repr(&self) -> String {
         float_ops::to_string(self.value)
--- a/vm/src/builtins/int.rs
+++ b/vm/src/builtins/int.rs
@@ -567,6 +567,7 @@ impl PyInt {
         !(&self.value)
     }
 
+    #[pymethod(name = "__str__")]
     #[pymethod(magic)]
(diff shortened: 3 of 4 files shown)


vm/src/builtins/complex.rs
```diff
@@ -337,6 +337,11 @@ impl PyComplex {
         vm.ctx
             .new_tuple(vec![vm.ctx.new_float(re), vm.ctx.new_float(im)])
     }
+
+    #[pymethod(magic)]
+    fn str(&self, vm: &VirtualMachine) -> PyStrRef {
+        PyStr::from(self.repr()).into_ref(vm)
```
youknowone, line 343: `self.repr()` will be enough with return type `String`

youknowone (changes requested):

vm/src/builtins/complex.rs
youknowone: I think adding `#[pymethod(name = __str__)]` on `fn repr` will work

Snowapril: Thank you. I sent v2 code 🥳

Snowapril:
I found that missing `__str__` method in `PyBool` leads to [`fstrings.py`](https://github.com/RustPython/RustPython/blob/master/extra_tests/snippets/fstrings.py) test failed.
I also provide `__str__` support in `PyBool` too.

Snowapril:
It was old features in python 3.6. Sorry for giving confusion :(

PR #2931, opened 2021-08-22
add deque __bool__ method

### LINKED ISSUE #2837: `collections.deque` doesn't have a `__bool__` method.
## Feature

CPython:

```python
>>> deque().__bool__()
False
```

RustPython:

```python
>>>>> deque().__bool__()
AttributeError: 'deque' object has no attribute '__bool__'
```

`bool` for `deque`s currently falls back to using `__len__` when it should supply an implementation of `__bool__`.

OhBonsai:
closes #2837

### DIFF
--- a/Lib/test/test_deque.py
+++ b/Lib/test/test_deque.py
@@ -269,6 +269,14 @@ class TestBasic(unittest.TestCase):
         self.assertRaises(IndexError, d.__getitem__, 0)
         self.assertRaises(IndexError, d.__getitem__, -1)
 
+    def test_bool(self):
+        d = deque('abc')
+        self.assertEqual(d, True)
+
+        d = deque([])
+        self.assertEqual(d, False)
+
+
     @unittest.skip("TODO: RUSTPYTHON hangs")
     def test_index(self):
         for n in 1, 2, 30, 40, 200:
--- a/Lib/test/test_deque.py
+++ b/Lib/test/test_deque.py
@@ -276,7 +276,6 @@ class TestBasic(unittest.TestCase):
         d = deque([])
         self.assertEqual(d, False)
 
-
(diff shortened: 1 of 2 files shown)


Lib/test/test_deque.py
```diff
@@ -269,6 +269,13 @@ def test_getitem(self):
         self.assertRaises(IndexError, d.__getitem__, 0)
         self.assertRaises(IndexError, d.__getitem__, -1)
 
+    def test_bool(self):
+        d = deque('abc')
+        self.assertEqual(d, True)
+
+        d = deque([])
+        self.assertEqual(d, False)
+
```
youknowone, line 278: we don't add tests inside `Lib/test`. To add our original tests, please do it into `extra_tests/snippets`

youknowone (changes requested):

vm/src/stdlib/collections.rs
```diff
@@ -399,6 +399,11 @@ mod _collections {
             self.borrow_deque().len()
         }
 
+        #[pymethod(magic)]
+        fn bool(&self) -> bool {
+            self.borrow_deque().len() != 0
```
DimitrisJim, line 404: Instead of comparing with zero, just use[ `is_empty()`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.is_empty) here. Does the same thing!

fanninpm:
Just a heads up: if you word your PR description such that it contains [these magic words,](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) GitHub will automatically close the linked issue.

DimitrisJim (approved):

DimitrisJim:
Thanks for the contribution! :tada:

PR #2932, opened 2021-08-22
Find key from only its attributes

moreal:
# Overview

Currently, there is a bug to find a key from itself and the base types not only its attributes. If there was user defined class named `A`, `"__setattr__" in A.__dict__` returns `True` because `object` has `__setattr__` though `A` class doesn't.

## RustPython

```python
>>>>> "__setattr__" in type.__dict__
True
>>>>> "__setattr__" in object.__dict__
True
>>>>> class A: ...
>>>>> "__setattr__" in A.__dict__
True
```

## CPython 3.8

```python
>>>>> "__setattr__" in type.__dict__
True
>>>>> "__setattr__" in object.__dict__
True
>>>>> class A: ...
>>>>> "__setattr__" in A.__dict__
False
```

### DIFF
--- a/Lib/enum.py
+++ b/Lib/enum.py
@@ -10,31 +10,41 @@ __all__ = [
 
 
 def _is_descriptor(obj):
-    """Returns True if obj is a descriptor, False otherwise."""
+    """
+    Returns True if obj is a descriptor, False otherwise.
+    """
     return (
             hasattr(obj, '__get__') or
             hasattr(obj, '__set__') or
-            hasattr(obj, '__delete__'))
-
+            hasattr(obj, '__delete__')
+            )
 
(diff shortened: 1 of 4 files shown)


Lib/test/support/__init__.py
```diff
@@ -2842,14 +2842,23 @@ def test__all__(self):
 
     expected = set(extra)
 
+    print()
+    print(module)
+    print(dir(module))
+    print()
```
youknowone, line 2848: these prints looks like debug codes

moreal: Oops, sorry 😞 I amended it, thanks!

youknowone (approved):

youknowone (approved):

PR #2933, opened 2021-08-22
Fix list entering deadlock in several methods

### LINKED ISSUE #2919: Current lock mechanism in container types make rustpython hang
## Feature
Current container types(`PyList`, `PyDeque`, ..) which use `RWLock` for keeping mutual exclusion cause RustPython **hang** when there exists **write** try while reading elements.

Because of it, Below codes work differently in RustPython and CPython
```python
class test:
	def __eq__(self, other):
		self.d.pop()
		return True
t = test()
d = [1,2,t,3,4]
t.d = d
d.count(4)
```
#### RustPython
```python
d.count(4) #Rustpython hang here
```
#### CPython
```python
d.count(4) # return 1
```
I tracked call stack for debugging when hanging is occurred.
During `list::count` method is calling, `PyRWLock` is locked with read guard
```rust
fn count(&self, needle: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
    let mut count: usize = 0;
    for element in self.borrow_vec().to_vec().iter() {    // borrow_vec keep RWLockReadGuard
        ......
}
```
As above python code try pop() method call during count() method calling, it call borrow_vec_mut() which try to keep
RWLockWriteGuard and rustpython hang.
```rust
fn pop(&self, i: OptionalArg<isize>, vm: &VirtualMachine) -> PyResult {
    let mut i = i.into_option().unwrap_or(-1);
    let mut elements = self.borrow_vec_mut();
    .....
}
```

## Python Documentation
* https://docs.python.org/3.8/tutorial/datastructures.html

Snowapril:
This revision fix #2919 in some part.

Current `count`, `contains`, `remove` and `index` methods in `PyList` implementation does not unlock the read guard before entering loop.
If there is something try to lock the write guard while iterating loops, rustpython die.

With this fix, there is no more deadlock, but resulting values from that methods slightly incompatible with cpython.
```python
class test:
	def __eq__(self, other):
		self.d.pop()
		return True
t = test()
d = [1,2,t,3,4]
t.d = d
d.count(4) 
# Rustpython : 2
# CPython : 1
```

### DIFF
--- a/extra_tests/snippets/list.py
+++ b/extra_tests/snippets/list.py
@@ -635,3 +635,24 @@ assert it.__length_hint__() == 0
 
 a = [*[1, 2], 3, *[4, 5]]
 assert a == [1, 2, 3, 4, 5]
+
+# Test for list entering daedlock or not (https://github.com/RustPython/RustPython/pull/2933)
+class MutatingCompare:
+    def __eq__(self, other):
+        self.list.pop()
+        return True
+
+m = MutatingCompare()
+
+l = [1, 2, 3, m, 4]
+m.list = l
+l.count(4) # TODO: assert l.count(4) == 1
+
+l = [1, 2, 3, m, 4]
+m.list = l
+assert l.index(4) == 3
+
+l = [1, 2, 3, m, 4]
+m.list = l
+l.remove(4) 
+assert_raises(ValueError, lambda: l.index(4)) # element 4 must not be in the list
--- a/vm/src/builtins/list.rs
+++ b/vm/src/builtins/list.rs
@@ -260,9 +260,10 @@ impl PyList {
 
     #[pymethod]
     fn count(&self, needle: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
+        let elements = self.borrow_vec().to_vec();
         let mut count: usize = 0;
-        for element in self.borrow_vec().to_vec().iter() {
-            if vm.identical_or_equal(element, &needle)? {
+        for elem in elements.iter() {
+            if vm.identical_or_equal(elem, &needle)? {
                 count += 1;
             }
         }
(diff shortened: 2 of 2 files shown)


youknowone (approved):
Because there is no tests about it, future regression is not preventable. Will you add related tests in `extra_tests/snippets`?

DimitrisJim:
I'm still not sure if this is the best approach to this. These all now require `O(N)` space instead of `O(1)`. The runtime considerations aren't that back since `to_vec` is *really* fast in relation to iterating and calling `identical_or_equal`.

I'd suggest we add a couple of TODO's in the code so we can keep track of this and try and see if we can find a better approach in the future.

youknowone:
I think the only difference is the lifecycle of locking. The original codes also use `to_vec` already.

Snowapril:
I also agree. Current implementation(in original) using `.to_vec` method cause `copy` which decrease performance.
It need to be improved like `if there is write try, unlock current locked read guard, do stuff and lock again.` 
I have no idea right now..

DimitrisJim:
Yeah, sorry, I was thinking of deque where `to_vec` isn't currently used. Here the difference is really none since `to_vec` as indeed used. I think this is better than hanging for now but a TODO would be good in order for us to not forget it.

Snowapril:
Yes. I'll send v2 code with comment later. 
but unfortunately, there is also hang in deque implementation.
```python
from collections import deque
class MutatingCompare:
    def __eq__(self, other):
        self.d.pop()
        return True

m = MutatingCompare()
d = deque([1, 2, 3, m, 4])
m.d = d
d.count(4) # Hang here.
```
I think current lock & unlock routine must be modified.

youknowone:
The comments will be nice if you do, but that's not up to responsibility of this PR. So please feel free about it.

Snowapril:
I add extra_test for checking rustpython entering deadlock or not in [`list.py`](https://github.com/RustPython/RustPython/blob/master/extra_tests/snippets/list.py).
Also I insert comment in  `count`, `contains`, `remove` and `index` methods about `to_vec()` for future work.

youknowone (approved):

DimitrisJim (approved):

PR #2934, opened 2021-08-22
resolve deque deadlock using state

eldpswp99:
resolve deque deadlock while iteration

by using CPython method, added state property to PyDeque.

by using state, RustPython can detect mutation while iterating.

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -27,8 +27,7 @@ mod _collections {
     struct PyDeque {
         deque: PyRwLock<VecDeque<PyObjectRef>>,
         maxlen: Option<usize>,
-        /* incremented whenever the indices move */
-        state: AtomicCell<usize>,
+        state: AtomicCell<usize>, // incremented whenever the indices move
     }
 
     type PyDequeRef = PyRef<PyDeque>;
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/collections.rs
```diff
@@ -447,7 +493,8 @@ mod _collections {
     struct PyDequeIterator {
         position: AtomicCell<usize>,
         status: AtomicCell<IterStatus>,
-        length: usize, // To track length immutability.
+        length: usize,
+        // To track length immutability.
```
youknowone, line 497: this comment looks to be placed wrong

vm/src/stdlib/collections.rs
```diff
@@ -171,23 +177,37 @@ mod _collections {
             PyDeque {
                 deque: PyRwLock::new(self.borrow_deque().clone()),
                 maxlen: self.maxlen,
+                state: AtomicCell::new(self.state.load()),
             }
         }
 
+        fn get_cloned_vec_deque(
```
DimitrisJim, line 184: I don't think this function is really necessary, calling `clone` on the lock should get you your `VecDeque` just fine.

DimitrisJim (changes requested):

vm/src/stdlib/collections.rs
eldpswp99: I fixed with using `clone`. Thanks!

vm/src/stdlib/collections.rs
eldpswp99: I fixed this, and also place new comment about state using this placement. Thanks!

DimitrisJim (approved):

PR #2935, opened 2021-08-22
os: implement os.nice 

zetwhite:
this implements `os.nice` (which linked to a unix system call, changing process priority) function. 

with this commit, rustpyhton works as below. 
```python 
>>>>> import os
>>>>> os.nice(0)
0
>>>>> os.nice(-1)
-1
>>>>> os.nice(-10)
-11
>>>>> os.nice(10)
-1
>>>>> os.nice(20)
19
>>>>> os.nice(20)
19
>>>>> os.nice(-20)
-1
>>>>> os.nice(-12)
-13
>>>>> os.nice(-100)
-20

```

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -2356,6 +2356,18 @@ mod posix {
     }
 
     #[pyfunction]
+    fn nice(increment: i32, vm: &VirtualMachine) -> PyResult<i32> {
+        use nix::errno::{errno, Errno};
+        Errno::clear();
+        let res = unsafe { libc::nice(increment) };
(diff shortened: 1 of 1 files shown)


youknowone (changes requested):
`nice` function is defined in `posix` module rather than `os` itself. Please move this funciton to `posix` module in the same file. (Search for `mod posix {`)
Then os module will import it from posix automatically. Becasue `posix` module is defined only for unix, you don't need to add `#[cfg(unix)]` then.

zetwhite:
Thank you for your kind comment! 
I changed as you mentioned. please check it out.

youknowone (approved):

PR #2938, opened 2021-08-22
Fix deque repr hang if element __repr__ mutates.

DimitrisJim:
a small follow up to #2934

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -382,8 +382,8 @@ mod _collections {
         #[pymethod(magic)]
         fn repr(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<String> {
             let repr = if let Some(_guard) = ReprGuard::enter(vm, zelf.as_object()) {
-                let elements = zelf
-                    .borrow_deque()
+                let deque = zelf.borrow_deque().clone();
+                let elements = deque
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/collections.rs
```diff
@@ -382,8 +382,8 @@ mod _collections {
         #[pymethod(magic)]
         fn repr(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<String> {
             let repr = if let Some(_guard) = ReprGuard::enter(vm, zelf.as_object()) {
-                let elements = zelf
-                    .borrow_deque()
+                let deque = zelf.borrow_deque().clone();
```
DimitrisJim, line 385: Weirdly enough, this can't be

```rust
let elements = zelf.borrow_deque().clone()
    .iter()
    .map(|obj| vm.to_repr(obj))
    .collect::<Result<Vec<_>, _>>()?;
```

It _seems_ as if rust compiles this down to a loop that goes through the elements, clones them and calls map on them  one by one. Moving it on a separate statement avoids this.

youknowone (approved):

vm/src/stdlib/collections.rs
youknowone: isn't it because of the lifecycle of borrow?

PR #2942, opened 2021-08-23
fix bytes and bytearray fromhex to support subclass

qingshi163:


vm/src/builtins/bytearray.rs
```diff
@@ -373,9 +375,15 @@ impl PyByteArray {
...
-    fn fromhex(string: PyStrRef, vm: &VirtualMachine) -> PyResult<PyByteArray> {
-        Ok(PyBytesInner::fromhex(string.as_str(), vm)?.into())
+    #[pyclassmethod]
+    fn fromhex(cls: PyTypeRef, string: PyStrRef, vm: &VirtualMachine) -> PyResult {
+        let bytes = PyBytesInner::fromhex(string.as_str(), vm)?;
+        let bytes = vm.ctx.new_bytes(bytes);
+        let args = FuncArgs {
+            args: vec![bytes],
+            ..Default::default()
+        };
```
qingshi163, line 385: better way to get FuncArgs from Vec<u8>?

youknowone: I think `FuncArgs::from(vec![..])` works

vm/src/builtins/bytearray.rs
```diff
@@ -373,9 +375,12 @@ impl PyByteArray {
...
 
-    #[pymethod]
-    fn fromhex(string: PyStrRef, vm: &VirtualMachine) -> PyResult<PyByteArray> {
-        Ok(PyBytesInner::fromhex(string.as_str(), vm)?.into())
+    #[pyclassmethod]
+    fn fromhex(cls: PyTypeRef, string: PyStrRef, vm: &VirtualMachine) -> PyResult {
+        let bytes = PyBytesInner::fromhex(string.as_str(), vm)?;
+        let bytes = vm.ctx.new_bytes(bytes);
+        let args = FuncArgs::from(vec![bytes]);
+        Callable::call(&cls, args, vm)
```
youknowone, line 383:
```suggestion
        Callable::call(&cls, vec![bytes].into(), vm)
```
Oh, I didn't look in the usage very well. maybe this expression also works

vm/src/builtins/bytearray.rs
```diff
@@ -373,9 +375,12 @@ impl PyByteArray {
         self.inner().hex(sep, bytes_per_sep, vm)
     }
 
-    #[pymethod]
-    fn fromhex(string: PyStrRef, vm: &VirtualMachine) -> PyResult<PyByteArray> {
-        Ok(PyBytesInner::fromhex(string.as_str(), vm)?.into())
+    #[pyclassmethod]
+    fn fromhex(cls: PyTypeRef, string: PyStrRef, vm: &VirtualMachine) -> PyResult {
+        let bytes = PyBytesInner::fromhex(string.as_str(), vm)?;
+        let bytes = vm.ctx.new_bytes(bytes);
```
youknowone, line 381: to check if this is intended: this is `bytearray` but calling `new_bytes`, right?

youknowone (approved):

vm/src/builtins/bytearray.rs
qingshi163: yes, see the updated test code. the bytes will pass to __new__ or __init__ in python than back to tp_new.

PR #2944, opened 2021-08-23
support deuqe.index using start, end index

eldpswp99:
before: `deque.index` only work when start and end was not given. 

after: this function now can handle when start and end is given

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -19,6 +19,7 @@ mod _collections {
     use crossbeam_utils::atomic::AtomicCell;
     use itertools::Itertools;
     use num_traits::ToPrimitive;
+    use std::cmp::{max, min};
     use std::collections::VecDeque;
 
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/collections.rs
```diff
@@ -240,26 +241,44 @@ mod _collections {
...
+            }
+        }
+
         #[pymethod]
         fn index(
             &self,
             obj: PyObjectRef,
-            start: OptionalArg<usize>,
-            stop: OptionalArg<usize>,
+            start: OptionalArg<i32>,
```
DimitrisJim, line 256: Don't have time to do a full review yet but `i32`? Indices usually take values in the range supplied by `isize`.

eldpswp99: ah.. i didn't know it. i would fix that

DimitrisJim (approved):

PR #2949, opened 2021-08-23
added __add__ method to collections.deque

### LINKED ISSUE #2835: `deque.__add__`

stromberger:
resolve #2835

add `deque.__add__` method


CryptoGodfatherVA (approved):

youknowone:
@eldpswp99 would you review this PR?

eldpswp99 (changes requested):
since we have extend function of PyDeque, we can use this. This has length check logic, the code can be more clear.

vm/src/stdlib/collections.rs
```diff
@@ -472,6 +472,32 @@ mod _collections {
             self.borrow_deque().len()
         }
 
+        #[pymethod(magic)]
+        fn add(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult<Self> {
+            if other.payload_is::<PyDeque>() {
+                let mut deque = self.borrow_deque().clone();
```
eldpswp99, line 478: sorry for not anchoring. we can make a new deque instance and use self.extend function to it

eldpswp99 (changes requested):

vm/src/stdlib/collections.rs
eldpswp99: and to prevent the fmt&clippy ci, before commit the code run `cargo fmt --all` and `cargo clippy`

DimitrisJim: Consider the following though @eldpswp99, in `extend` we have to deal with any object as long as we can grab an iterator from it to extract its elements. Here we only deal with `deque`s, as such, we can directly clone the underlying `VecDeque` from `other` and get a significant speed-up (still O(N) but with significantly smaller lower-order terms).

vm/src/stdlib/collections.rs
```diff
@@ -472,6 +472,32 @@ mod _collections {
             self.borrow_deque().len()
         }
 
+        #[pymethod(magic)]
+        fn add(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult<Self> {
+            if other.payload_is::<PyDeque>() {
```
DimitrisJim, line 477: You can use `payload_if_subclass` to get a handle of `other` as a `PyDeque`. Then you can clone its `VecDeque` directly and add it to the elements of `self`'s `VecDeque`.

DimitrisJim (changes requested):

vm/src/stdlib/collections.rs
stromberger: Thanks for the hint! I refactored it.

DimitrisJim:
Thanks for updating this, could you resolve the conflicts that have come up?

DimitrisJim (approved):
lgtm. not sure why the option to run the actions isn't showing for me though.

fanninpm:
> not sure why the option to run the actions isn't showing for me though.

Is it because of the merge commit? Perhaps a rebase is needed.

PR #2950, opened 2021-08-24
Add contextvars from CPython 3.8

fanninpm:
This is currently scaffolding for the `contextvars` module introduced in CPython 3.7 (see PEP 567).

I'm welcome to any suggestions on how to flesh this out.

### DIFF
--- a/Lib/contextvars.py
+++ b/Lib/contextvars.py
@@ -1,0 +1,4 @@
+from _contextvars import Context, ContextVar, Token, copy_context
+
+
+__all__ = ('Context', 'ContextVar', 'Token', 'copy_context')
--- a/Lib/test/test_context.py
+++ b/Lib/test/test_context.py
@@ -1,0 +1,1070 @@
+import concurrent.futures
+import contextvars
+import functools
+import gc
+import random
+import time
+import unittest
+import weakref
+
+try:
+    from _testcapi import hamt
+except ImportError:
+    hamt = None
+
+
+def isolated_context(func):
+    """Needed to make reftracking test mode work."""
+    @functools.wraps(func)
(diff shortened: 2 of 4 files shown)


vm/src/stdlib/contextvars.rs
```diff
@@ -0,0 +1,63 @@
...
+        const TP_FLAGS: crate::slots::PyTpFlags = crate::slots::PyTpFlags::DEFAULT;
+
+        fn impl_extend_class(_ctx: &PyContext, _class: &PyTypeRef) {
+            // TODO: RUSTPYTHON
+        }
+
+        fn extend_slots(_slots: &mut PyTypeSlots) {
+            // TODO: RUSTPYTHON
+        }
+    }
```
youknowone, line 25: are you going to write these part manually instead of writing an `#[pyimpl]`?

fanninpm:
CPython references:

- https://github.com/python/cpython/blob/3.8/Include/context.h
- https://github.com/python/cpython/blob/3.8/Python/context.c
- https://github.com/python/cpython/blob/3.8/Python/clinic/context.c.h
- https://github.com/python/cpython/blob/3.8/Modules/_contextvarsmodule.c
- https://github.com/python/cpython/blob/3.8/Modules/clinic/_contextvarsmodule.c.h
- https://github.com/python/cpython/blob/3.8/Include/internal/pycore_context.h

DimitrisJim:
Bump, to you want to open an issue for the discussion on this? I think it would be better than having the discussion on the PR.

youknowone (approved):
I wasnt aware this was going. Looks good. Other people can work on here later.

Issue #2936, opened 2021-08-22
isinstance does not support generic types.


domdfcoding:
## Feature

CPython supports checking whether an object is an instance of one of the abstract types from the typing module, such as `Dict` or `Sequence`:

```python
>>> from typing import Sequence
>>> isinstance([], Sequence)
True
```

These "classes" are actually instances of `typing._GenericAlias` in Python 3.8 (`typing._SpecialGenericAlias` in 3.9).

In RustPython an exception is raised:

```python
>>>>> from typing import Sequence
>>>>> isinstance([], Sequence)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: isinstance() arg 2 must be a type or tuple of types, not _GenericAlias

```

-----

In CPython instance checks with subscripted generics fail:

```python

>>> isinstance([], Sequence[str])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/typing.py", line 769, in __instancecheck__
    return self.__subclasscheck__(type(obj))
  File "/usr/lib/python3.8/typing.py", line 777, in __subclasscheck__
    raise TypeError("Subscripted generics cannot be used with"
TypeError: Subscripted generics cannot be used with class and instance checks

```

-----

In CPython, `isinstance` will accept anything as the second argument as long as it has an ``__instancecheck__`` method:

```python
>>> class G:
...     def __instancecheck__(self, obj):
...             return True
...
>>> isinstance([], G())
True
```

It seems RustPython isn't checking for the presence of the method before raising the exception:

```python
>>>>> class G:
.....     def __instancecheck__(self, obj):
.....             return True
.....
>>>>> isinstance([], G())
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: isinstance() arg 2 must be a type or tuple of types, not G
```

In the typing module `_GenericAlias` isn't a subclass of `type` but has a ``__instancecheck__`` method.

## Python Documentation

The docs are a bit unclear, but relevant parts are:

* https://docs.python.org/3/library/functions.html#isinstance
* https://docs.python.org/3/reference/datamodel.html#class.__instancecheck__
* https://docs.python.org/3/glossary.html#term-abstract-base-class

and [PEP 3119](https://www.python.org/dev/peps/pep-3119/)

Snowapril:
Hi. I fixed this issue in #2952 . Could you check it one more time? 😊

[...]

Issue #2943, opened 2021-08-23
`issubclass` does not work on generic types


Snowapril:
## Feature

This issue is almost same as #2936 issue. 
Current rustpython implementation does not catch `__subclasscheck__` method defined in class object when call `issubclass` operation.

```python
class G:
	def __subclasscheck__(self, other):
		return True
issubclass([], G())
```
#### CPython
```python
True
```
cpython catch `__subclasscheck__` method defined in class G.

#### RustPython
```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: Expected type 'type', not 'list'
```
but current rustpython do not catch  `__subclasscheck__` method defined in class G.

## Python Documentation
* https://docs.python.org/3.8/library/functions.html?highlight=issubclass#issubclass
* https://docs.python.org/3.8/reference/datamodel.html?highlight=__subclasscheck__#class.__subclasscheck__
* https://www.python.org/dev/peps/pep-3119/

PR #2952, opened 2021-08-24
Fix `isinstance` to support Generic type object

Snowapril:
This revision fix issue #2936.

In original codes, there are problem in both `builtins::isinstance` and `vm.isinstance`.

`builtins::isinstance` use `single_or_tuple_any` function for checking
given object is `<class 'type'>` or if it is tuple, iterating over them
recursively.
But it did not check given object have `__instancecheck__` if it isn't `<class 'type'>`.

`vm.isinstance` get second argument as `PyTypeRef`. 
Because of it, `call_special_method` use `instancecheck` of `TypeProtocol`'s one and
never reach to user defined `__instancecheck__`

Therefore I fixed it with referenced [cpython implementation](https://github.com/python/cpython/blob/3.8/Objects/abstract.c#L2424).
And also add extra test about it.

#### Before
```python
>>>>> from typing import Sequence
>>>>> isinstance([], Sequence)
TypeError: isinstance() arg 2 must be a type or tuple of types, not _GenericAlias

>>>>> class G:
.....     def __instancecheck__(self, obj):
.....             return True
.....
>>>>> isinstance([], G())
TypeError: isinstance() arg 2 must be a type or tuple of types, not G
```

#### After
```python
>>>>> from typing import Sequence
>>>>> isinstance([], Sequence)
True

>>>>> class G:
.....   def __instancecheck__(self, other):
.....           return True
..... 
>>>>> isinstance([], G())
True
```

issue #2943 is almost same with this PR, but I want to know If I'm going right way. Thanks.


Snowapril:
I found `test_isinstance.py` in cpython repository, and current implementation(also original one) fail it.

vm/src/vm.rs
```diff
@@ -941,21 +941,45 @@ impl VirtualMachine {
...
-            let ret = self.call_special_method(
-                cls.as_object().clone(),
-                "__instancecheck__",
-                (obj.clone(),),
-            )?;
-            pybool::boolval(self, ret)
+            return Ok(true);
         }
+
+        if self.ctx.types.type_type.is(&cls.clone_class()) {
```
DimitrisJim, line 964:
```suggestion
        if self.ctx.types.type_type.is(&cls.clone_class()) 
```

Snowapril: Oh you are right. no need to use `clone_class`. Thanks!

youknowone:
```suggestion
        if cls.lease_class().is(self.ctx.types.type_type) {
```

DimitrisJim: I pasted the wrong thing :laughing: I wanted to suggest `cls.class().is(self.ctx.types.type_type)`. Any difference between that and `lease_class` @youknowone ?

Snowapril: I think his intend was your suggestion code :) I apply your suggestion and add expected failure to
failed test in `test_isinstance.py`.

fanninpm: @DimitrisJim `lease_class` doesn't exist any more thanks to a089c9b. I think @youknowone might have slipped up.

Snowapril:
@youknowone I think failed tests in newly added `test_isinstance.py` is little complex to fix in this PR and there is also `issubclass` test not `isinstance`. How about open failed tests as TODO for future contributors?

DimitrisJim:
@Snowapril yeah, you've done the right thing to mark them as `TODO's`. The rest of the test cases can be fixed in the future.

extra_tests/snippets/isinstance.py
```diff
@@ -53,4 +60,4 @@ class ReturnInt(metaclass=MCReturnInt):
 
 assert isinstance("a", ReturnInt) is True
 
-assert isinstance(1, ((int, float,), str))
+assert isinstance(1, ((int, float,), str))
```
youknowone, line 63: revert this line not to make whitespace change

vm/src/vm.rs
```diff
@@ -941,21 +941,45 @@ impl VirtualMachine {
...
+        }
+    }
+
     /// Determines if `obj` is an instance of `cls`, either directly, indirectly or virtually via
     /// the __instancecheck__ magic method.
-    pub fn isinstance(&self, obj: &PyObjectRef, cls: &PyTypeRef) -> PyResult<bool> {
+    pub fn isinstance(&self, obj: &PyObjectRef, cls: &PyObjectRef) -> PyResult<bool> {
         // cpython first does an exact check on the type, although documentation doesn't state that
-        // https://github.com/python/cpython/blob/a24107b04c1277e3c1105f98aff5bfa3a98b33a0/Objects/abstract.c#L2408
+        // https://github.com/python/cpytho(n/blob/a24107b04c1277e3c1105f98aff5bfa3a98b33a0/Objects/abstract.c#L2408
```
youknowone, line 959: typo?

youknowone (changes requested):

youknowone:
totally positive. you don't need to make everything working to add a test file.

vm/src/vm.rs
Snowapril: Thanks! I applied it 🥳

youknowone: Will you check it again? It seems something didnt' work.

fanninpm:
```suggestion
        // https://github.com/python/cpython/blob/a24107b04c1277e3c1105f98aff5bfa3a98b33a0/Objects/abstract.c#L2408
```

Snowapril: omg .. I had been thought I modified it. Sorry. I upload again.

extra_tests/snippets/isinstance.py
youknowone: would you also clear this change not to be made?

Snowapril: applied it ! Sorry for make you cumbersome 😂

PR #2955, opened 2021-08-24
Fixes panic when empty array is indexed.

DimitrisJim:
Fixes a panic when an empty array was indexed. Uses `PySliceableSequence::wrap_index` as other collections do.

### DIFF
--- a/vm/src/stdlib/array.rs
+++ b/vm/src/stdlib/array.rs
@@ -12,7 +12,9 @@ use crate::common::lock::{
     PyRwLockWriteGuard,
 };
 use crate::function::OptionalArg;
-use crate::sliceable::{saturate_index, PySliceableSequence, PySliceableSequenceMut};
+use crate::sliceable::{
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/array.rs
```diff
@@ -221,27 +207,26 @@ macro_rules! def_array_enum {
                 }
             }
 
-            fn getitem_by_slice(&self, slice: PySliceRef, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
+            fn getitem(&self, needle: Either<isize, PySliceRef>, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 210: enum `SequenceIndex` is recently added to replace `Either<isize, PySliceRef>`

youknowone (changes requested):

youknowone (approved):

vm/src/stdlib/array.rs
youknowone: Oh, it wasn't a change from this PR. 😅

PR #2959, opened 2021-08-24
Bump recursion limit to match CPython's

### LINKED ISSUE #2948: python-speed failing to run
This is a simple python benchmark:

https://github.com/vprelovac/python-speed

Fails in first test with recursion limit message.

DimitrisJim:
This was set in #1477 to 512 though 1000 appears to easily be supported as a default. This matches the value currently used by CPython and closes #2948


DimitrisJim:
Locally, this is fine. Seems like the test runners aren't able to handle it though. Closing for now since I'm pretty sure 1000 is considered an implementation detail at the end of the day.

youknowone:
https://docs.python.org/3/library/sys.html#sys.setrecursionlimit

I think default value can be 1000 and we can override it for CI

DimitrisJim:
But where would I set it? The only place I can think of is in the failing function.

youknowone:
I am thinking of `Lib/test/__init__.py` with environment variable check

Lib/test/test_threading.py
```diff
@@ -1062,6 +1062,13 @@ def test_releasing_unacquired_lock(self):
         self.assertRaises(RuntimeError, lock.release)
 
     def test_recursion_limit(self):
+        # TODO: RUSTPYTHON
```
DimitrisJim, line 1065: @youknowone I couldn't find an `__init__` file for `test`, plus I thought it would be good to keep the change localized (only this function). I've added here and hopefully the tests will pass now, if you think it should placed elsewhere (maybe `__main__`?) let me know.

DimitrisJim:
I cant diagnore the issue here. Trying to set the recursion limit doesn't seem to do the trick either.

PR #2960, opened 2021-08-24
Make OSError subclass instead of itself

### LINKED ISSUE #2914: OSError has invalid `__new__`
## Feature

Right now `OSError` has a standard `__new__`, which is not correct. Because it is a special case in CPython:

```python
>>> OSError()
OSError()
>>> OSError(1)
OSError(1)
>>> OSError(1, 2)
PermissionError(1, 2)
```

RustPython:

```python
>>>>> OSError()  # ok
OSError()
>>>>> OSError(1)  # ok
OSError(1)
>>>>> OSError(1, 2)  # error
OSError(1, 2)
```

## Python Documentation

Source code: https://github.com/python/cpython/blob/ecc3c8e4216958d85385bf2467441c975128f26c/Objects/exceptions.c#L991-L1003

Related #2897 
Related #2771

ChJR:
This PR will fix #2914.

### DIFF
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -22,6 +22,7 @@ use std::collections::HashSet;
 use std::fmt;
 use std::fs::File;
 use std::io::{self, BufRead, BufReader};
+use std::ops::Deref;
 
 #[pyclass(module = false, name = "BaseException")]
 pub struct PyBaseException {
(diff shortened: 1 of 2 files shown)


vm/src/exceptions.rs
```diff
@@ -1408,4 +1389,218 @@ pub(super) mod types {
...
+
+            error
+        } else {
+            None
+        }
+    }
+
+    fn base_exception_new(cls: PyTypeRef, args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+        PyBaseException::slot_new(cls, args, vm)
+    }
```
youknowone, line 1565: if these methods are not required to be moved, keep them on the original place. they will be put in `impl` later once we revise `pyexception` macro

ChJR: They are used in define_macro.

youknowone: I agree it is used. I don't see a good reason why it need to be moved to here.

vm/src/lib.rs
```diff
@@ -85,6 +85,8 @@ pub use rustpython_common as common;
 #[cfg(feature = "rustpython-compiler")]
 pub use rustpython_compiler as compile;
 
+use crate::stdlib::errno::errors;
```
youknowone, line 88: why is this module imported here? I expect this is imported from `os_error_optional_new`.

vm/src/exceptions.rs
youknowone: I agree it is used. I don't see a good reason why it need to be moved to here. I checked out the code and tried to compile it without moving this function. I couldn't see any problem.

youknowone (changes requested):

vm/src/lib.rs
ChJR: My mistake.

vm/src/exceptions.rs
ChJR: I see. I moved functions just for distinguishing functions and macros. For readability, if you want not moving the functions I can revert that.

PR #2961, opened 2021-08-24
fix pattern for construct the bytes and bytearray

qingshi163:


vm/src/bytesinner.rs
```diff
@@ -80,62 +78,114 @@ impl ByteInnerNewOptions {
...
+                        Ok(b) => {
+                            return Ok(b);
+                        }
+                        Err(obj) => obj,
                     }
+                } else {
+                    obj
+                };
+
+                let inner = if let Some(bytes_method) = vm.get_method(obj.clone(), "__bytes__") {
```
qingshi163, line 108: any way to check the method existing without clone?

DimitrisJim: don't think so.

vm/src/bytesinner.rs
```diff
@@ -80,62 +78,114 @@ impl ByteInnerNewOptions {
...
+                    // construct an exact bytes from __bytes__ slot.
+                    // if __bytes__ return a bytes, use the bytes object except we are the subclass of the bytes
+                    let bytes = vm.invoke(&bytes_method?, ())?;
+                    let bytes = if cls.is(PyBytes::class(vm)) {
+                        match bytes.downcast::<PyBytes>() {
```
DimitrisJim, line 113: Pretty sure you can just use `downcast_exact` here.

qingshi163: Don't think so, the cls here is the type of the object to be constructed, and the bytes is the object returned from \_\_bytes__ slot.

PR #2967, opened 2021-08-26
Avoid hanging when mutating __index__ is used.

DimitrisJim:
In short, a mutating `__index__` could be called that tried to lock the collection and cause a deadlock.

This separates the steps by first converting using `__index__` and then safely converting into a rust `Range` in the `PySliceableSequence(Mut)` traits.


youknowone:
@DimitrisJim If you need large work for update, how about just replacing the name `SeqSlice` to `SliceIndices` and merge it first?

DimitrisJim:
this is a bigger mess to rebase than to create a fresh PR so I'll close and open another.

PR #2984, opened 2021-08-28
implement `complex.__complex__`

Snowapril:
This revision implement `complex.__complex__` which will be added in cpython 3.11 for `typing` module.

### References
* https://docs.python.org/3.11/whatsnew/3.11.html#other-cpython-implementation-changes
* https://bugs.python.org/issue24234

### DIFF
--- a/extra_tests/snippets/complex.py
+++ b/extra_tests/snippets/complex.py
@@ -1,10 +1,15 @@
-# __complex__
-z = 3 + 4j
-assert z.__complex__() == z
-assert type(z.__complex__()) == complex
+import testutils
 
-class complex_subclass(complex):
-    pass
-z = complex_subclass(3 + 4j)
-assert z.__complex__() == 3 + 4j
-assert type(z.__complex__()) == complex
+# __complex__
+def test__complex__():
+    z = 3 + 4j
+    assert z.__complex__() == z
+    assert type(z.__complex__()) == complex
+    
+    class complex_subclass(complex):
+        pass
+    z = complex_subclass(3 + 4j)
+    assert z.__complex__() == 3 + 4j
+    assert type(z.__complex__()) == complex
+
+testutils.skip_if_unsupported(3,11,test__complex__)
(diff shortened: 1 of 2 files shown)


Lib/test/test_complex.py
```diff
@@ -412,6 +412,18 @@ def __complex__(self):
             self.assertEqual(complex(complex1(1j)), 2j)
         self.assertRaises(TypeError, complex, complex2(1j))
 
+    def test___complex__(self):
```
DimitrisJim, line 415: I feel like the whole test file needs to be updated and not just a single function since now we have a test file thats part Python 3.8 part Python 3.11. @youknowone what's do we do in these cases?

Snowapril:
I fixed some code according to suggestion in #2986

Lib/test/test_complex.py
youknowone: Yes, you are right. But 3.11 is not published and not even stablized yet. We added tests to `extra_tests/snippets` before in similar situation. But I don't have a concrete idea what's the best.

youknowone: In my opinion, generally we'd better to keep easy to update `Lib/` from CPython. But adding test test here will increase the chance for conflict in Future. So I think adding to `extra_tests` is better than adding here.

DimitrisJim: Yup that makes sense to me too. @Snowapril move this test into `extra_tests/snippets/complex.py` so we can push this through (similarly for bytes).

Snowapril: I applied it! (Also same for bytes). As `extra_tests/snippets/complex.py` doesn't exist I newly created it and pushed test.
Thanks :)

Snowapril: Uhm... As `__complex__` and `__bytes__` are not a spec of cpython 3.8, command `pipenv run pytest -v` in github action(Run snippets and cpython tests) go failed. How can I handle this case?

Snowapril: Ah! I found `skip_if_unsupported` in test utils can resolve this failure.

youknowone (approved):

PR #2986, opened 2021-08-28
Implement `bytes.__bytes__`

Snowapril:
This revision implement `bytes.__bytes__` in cpython 3.11 for typing module.

### References
* https://docs.python.org/3.11/whatsnew/3.11.html#other-cpython-implementation-changes
* https://bugs.python.org/issue24234

### DIFF
--- a/vm/src/builtins/bytes.rs
+++ b/vm/src/builtins/bytes.rs
@@ -129,6 +129,15 @@ impl PyBytes {
     }
 
     #[pymethod(magic)]
+    fn bytes(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyRef<Self> {
+        if zelf.is(&vm.ctx.types.bytes_type) {
+            zelf
+        } else {
+            PyBytes::from(zelf.inner.clone()).into_ref(vm)
+        }
+    }
+
+    #[pymethod(magic)]
     fn sizeof(&self) -> usize {
         size_of::<Self>() + self.inner.elements.len() * size_of::<u8>()
     }
--- a/extra_tests/snippets/bytes.py
+++ b/extra_tests/snippets/bytes.py
@@ -1,4 +1,4 @@
-from testutils import assert_raises
+from testutils import assert_raises, skip_if_unsupported
 
 # new
 assert bytes([1, 2, 3])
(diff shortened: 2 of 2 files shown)


vm/src/builtins/bytes.rs
```diff
@@ -128,6 +128,11 @@ impl PyBytes {
         &self.inner.elements
     }
 
+    #[pymethod(name = "__bytes__")]
+    pub fn bytes(&self) -> Vec<u8> {
+        self.inner.elements.to_vec()
+    }
+
```
qingshi163, line 135: If the object itself is an exact PyBytes, we should return the reference of the object instead creating a new object.

qingshi163 (changes requested):

vm/src/builtins/bytes.rs
Snowapril: Is this right? 
```rust
#[pymethod(magic)]
pub fn bytes(zelf: PyRef<Self>) -> PyRef<Self> {
    zelf
}
```

Snowapril: I suffered from exactly same problem in #2984. When I return PyBytes itself, below test go failed.
```python
class bytes_subclass(bytes):
    pass

bar = bytes_subclass(b'bar')
self.assertEqual(bar.__bytes__(), bar)
self.assertEqual(type(bar.__bytes__()), self.type2test)
```

Snowapril: Ah I'm sorry. I understood. Let me try one more time

Snowapril: @qingshi163 I applied it! Could you check one more time please? Thanks 😊

vm/src/builtins/bytes.rs
```diff
@@ -128,6 +128,15 @@ impl PyBytes {
...
     }
 
+    #[pymethod(name = "__bytes__")]
+    pub fn bytes(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyObjectRef {
+        if zelf.is(&vm.ctx.types.bytes_type) {
+            zelf.into_object()
+        } else {
+            PyBytes::from(zelf.inner.elements.to_vec()).into_object(vm)
+        }
+    }
```
DimitrisJim, line 138: Don't think you can avoid the clone here.

```suggestion
    #[pymethod(magic)]
    fn bytes(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyRef<PyBytes> {
        if zelf.is(&vm.ctx.types.bytes_type) {
            zelf
        } else {
            PyBytes::from(zelf.inner.clone()).into_ref(vm)
        }
    }
```

DimitrisJim (changes requested):

vm/src/builtins/bytes.rs
Snowapril: You are right. It seems extremely better than original. Thanks. I sent v3 code.

youknowone (approved):

PR #2987, opened 2021-08-28
add __new__ method to deque iterator and reverse iterator

eldpswp99:
add `__new__` function to deque iterator, and reverse deque iterator

this allows to create iterator instance using type instance. for example,
```
klass = type(iter(deque()))
a = klass(deque('abcd'))
```
creates a deque iterator instance

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -3,7 +3,7 @@ pub(crate) use _collections::make_module;
 #[pymodule]
 mod _collections {
     use crate::common::lock::{PyRwLock, PyRwLockReadGuard, PyRwLockWriteGuard};
-    use crate::function::{FuncArgs, OptionalArg};
+    use crate::function::{FuncArgs, KwArgs, OptionalArg};
     use crate::slots::{Comparable, Hashable, Iterable, PyComparisonOp, PyIter, Unhashable};
     use crate::vm::ReprGuard;
     use crate::VirtualMachine;
@@ -603,6 +603,7 @@ mod _collections {
         fn tp_new(
             cls: PyTypeRef,
             DequeIterArgs { deque, index }: DequeIterArgs,
+            _kwargs: KwArgs,
             vm: &VirtualMachine,
         ) -> PyResult<PyRef<Self>> {
             let len = deque.len();
@@ -693,6 +694,7 @@ mod _collections {
         fn tp_new(
             cls: PyTypeRef,
             DequeIterArgs { deque, index }: DequeIterArgs,
+            _kwargs: KwArgs,
             vm: &VirtualMachine,
         ) -> PyResult<PyRef<Self>> {
             let len = deque.len();
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -360,15 +360,14 @@ mod _collections {
         }
 
         #[pymethod(magic)]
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/collections.rs
```diff
@@ -591,6 +590,19 @@ mod _collections {
             }
         }
 
+        #[pyslot]
+        fn tp_new(cls: PyTypeRef, args: FuncArgs, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+            if args.args.len() <= 1 {
+                let deque = args.bind(vm)?;
+                PyDequeIterator::new(deque).into_ref_with_type(vm, cls)
+            } else {
+                let (deque, position) = args.bind(vm)?;
```
DimitrisJim, line 599: `index` (i.e `position` here) should be allowed as a positional and a keyword. Right now it's only positional. Other than that, it lgtm.

DimitrisJim (changes requested):

vm/src/stdlib/collections.rs
DimitrisJim: Hm, in addition to that, you'd need to check that the index passed in isn't negative and that it isn't `> deque.len`. On first thought, in the first case you should silently ignore it and in the second you should set the iterator to exhausted.

eldpswp99: > `index` (i.e `position` here) should be allowed as a positional and a keyword. Right now it's only positional. Other than that, it lgtm.

can you give more information about what "positional" and "keyword" means?

DimitrisJim: Essentially, the iterators should be able to be called in three ways:

```
rev_it = type(reversed(deque()))
rev_it(deque())                 # only deque specified
rev_it(deque(), 1)             # deque and index as a positional arg specified
rev_it(deque(), index=1)  # deque and index as a keyword arg specified
```

In short, it should be easier to specify `IterNewArgs` (see `PyDequeOptions` for insipation) with `index` being `optional`.

eldpswp99: Ah, i got it, thanks! i will send v2 code till tomorrow

eldpswp99: @DimitrisJim  I checked the keyword case using the code
```
from collections import deque

rev_it = type(reversed(deque()))
c = rev_it(deque('abcd'), index=1)

for i in c:
    print(i)
```
and the index key doesn't work. it prints d,c,b,a
also, in [CPython](https://github.com/python/cpython/blob/eb263f9a356f5c5f21b8d5ce20bac92f31c40cad/Modules/_collectionsmodule.c#L1775) it is using ParseTuple in `dequeiter_new`. So, I think only positional parameter is allowed in `__new__`

I added extra_test in snippet. in CPython 3.8, the test code works well also with the commented test. however, in RustPython, it detects the wrong keyword index and return error. So, I commented the keyword test in extra_test and remained it as TODO

youknowone: Add an unused parameter with type `rustpython_vm::function::KwArgs` like `_kwargs: KwArgs`. Then it will take any keyword arguments but ignore them

eldpswp99: I fixed with v3 code. Thank you for the advice!

DimitrisJim: Yup, it actually allows any number of kwargs but ignores them as you say. Can't see how this is the intended behavior, an issue about it should probably be opened.

DimitrisJim (approved):

PR #2988, opened 2021-08-28
Clean up edge cases in `isinstance`/`issubclass`.

DimitrisJim:


DimitrisJim:
@Snowapril wanna review these changes?

vm/src/vm.rs
```diff
@@ -944,14 +944,50 @@ impl VirtualMachine {
...
+            // Only mask AttributeErrors.
+            if e.class().is(&self.ctx.exceptions.attribute_error) {
+                self.new_type_error(msg)
+            } else {
+                e
+            }
+        })
+    }
+
     pub fn _isinstance(&self, obj: &PyObjectRef, cls: &PyObjectRef) -> PyResult<bool> {
```
Snowapril, line 960:
```suggestion
    fn recursive_isinstance(&self, obj: &PyObjectRef, cls: &PyObjectRef) -> PyResult<bool> {
```
Totally looks good to me 😊 How about rename `_isinstance` to `recursive_isinstance`? As cpython and `issubclass` use `recursive_*` naming for that case, I think it is a little more reasonable.

vm/src/vm.rs
```diff
@@ -944,14 +944,50 @@ impl VirtualMachine {
         }
     }
 
+    // Equivalent to check_class. Masks Attribute errors (into TypeErrors) and lets everything
+    // else go through.
+    fn _check_cls(&self, cls: &PyObjectRef, msg: String) -> PyResult {
```
youknowone, line 949: msg part need to be a closure to avoid unnecessary formatting

```suggestion
    fn _check_cls<F>(&self, cls: &PyObjectRef, msg: F) -> PyResult
    where
        F: Fn() -> String
    {
```

vm/src/vm.rs
```diff
@@ -944,14 +944,50 @@ impl VirtualMachine {
...
+        })
+    }
+
     pub fn _isinstance(&self, obj: &PyObjectRef, cls: &PyObjectRef) -> PyResult<bool> {
         if let Ok(typ) = PyTypeRef::try_from_object(self, cls.clone()) {
-            Ok(obj.isinstance(&typ))
+            if obj.class().issubclass(typ.clone()) {
+                Ok(true)
+            } else if let Ok(icls) =
+                PyTypeRef::try_from_object(self, self.get_attribute(obj.clone(), "__class__")?)
```
youknowone, line 965: is this line meaning `obj.class()` sometimes different from `obj.__class__`?

vm/src/vm.rs
DimitrisJim: Good idea!

vm/src/vm.rs
DimitrisJim: Yup, check out the test cases for examples where this is manipulated.

vm/src/vm.rs
DimitrisJim: This generally follows the pattern found in the `vm.new_exception_type(&self, msg: String)`. Is there any benefit in passing a closure that calls the `format!` macro itself?

youknowone: Yes, the error message is only used when conditionally error raised. But if we format the argument itself, the format is always executed regardless the condition it is required or not.

vm/src/vm.rs
youknowone: when `__class__` is replaced by python code, I think `obj.class()` need to be replaced rather than checked by here. I am good with this way for now, but could you add a FIXME comment here?

vm/src/vm.rs
DimitrisJim: Ah true, address that.

vm/src/vm.rs
DimitrisJim: Oh, sorry, I just saw the comment. Are you sure that's possible? I don't think `CPython` allows Python code modifying `tp_type` (and `obj.class()` grabs the equivalent of that if I'm not mistaken). It's the reason why on the next line we check:

```
if icls.is(&obj.class())
```

this way, if `__class__` is something different, we check if that is a subclass of `typ`.

Snowapril (approved):

vm/src/vm.rs
youknowone: Thanks, I think that's possible but honestly I didn't checked how CPython handle it.

youknowone (approved):

PR #2991, opened 2021-08-30
Improve set iteration, add __reduce__

DimitrisJim:
Uses the underlying `Dict` to perform the iteration similar to how the dictionary iterators do. This leads to a slight improvement since the elements Vec isn't constructed (which was constructed to temporarily avoid the previous `O(N**2)` implementation of set iteration).

### DIFF
--- a/extra_tests/snippets/set.py
+++ b/extra_tests/snippets/set.py
@@ -194,17 +194,12 @@ with assert_raises(TypeError):
 a = set([1, 2, 3])
 i = iter(a)
 a.add(4)
-a.remove(4)
-assert next(i) == 1
-
-a = set([1, 2, 3])
-i = iter(a)
-a.add(4)
 with assert_raises(RuntimeError):
     next(i)
 a.remove(4)
-with assert_raises(RuntimeError):
-    next(i)
+# TODO: Raises RuntimeError in CPython but we currently raise Runtime error.
+# with assert_raises(StopIteration):
+#     next(i)
 
 # frozen set
 
--- a/Lib/test/test_iterlen.py
+++ b/Lib/test/test_iterlen.py
@@ -170,13 +170,9 @@ class TestSet(TestTemporarilyImmutable, unittest.TestCase):
         self.it = iter(d)
         self.mutate = d.pop
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_immutable_during_iteration(self):
         super().test_immutable_during_iteration()
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_invariant(self):
         super().test_invariant()
 
--- a/Lib/test/test_set.py
+++ b/Lib/test/test_set.py
@@ -235,8 +235,6 @@ class TestJointOps:
                 dup = pickle.loads(p)
                 self.assertEqual(self.s.x, dup.x)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_iterator_pickling(self):
         for proto in range(pickle.HIGHEST_PROTOCOL + 1):
             itorg = iter(self.s)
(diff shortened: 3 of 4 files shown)


extra_tests/snippets/set.py
```diff
@@ -195,7 +195,8 @@ def __hash__(self):
 i = iter(a)
 a.add(4)
 a.remove(4)
-assert next(i) == 1
+with assert_raises(RuntimeError):
```
DimitrisJim, line 198: Note on this:

Set behaves the same as dictionaries and detects mutations even when they are balanced, i.e:

```
d = {1: 2, 3: 4}
dit = iter(d)
d[5] = 6
del d[6]
next(dit)  # RuntimeError
```

CPython doesn't detect any of these but openly states in [PEP 3106](https://www.python.org/dev/peps/pep-3106/#id5) that mutations while iterating is undefined behavior:

> As in Python 2.x, mutating a dict while iterating over it using an iterator has an undefined effect and will in most cases raise a `RuntimeError` exception. (This is similar to the guarantees made by the Java Collections Framework.) 

So I'm pretty sure that a `RuntimeError` is the correct approach here.

extra_tests/snippets/set.py
```diff
@@ -195,15 +195,16 @@ def __hash__(self):
...
+    next(i) == 1
 
 a = set([1, 2, 3])
 i = iter(a)
 a.add(4)
 with assert_raises(RuntimeError):
     next(i)
 a.remove(4)
-with assert_raises(RuntimeError):
+with assert_raises(StopIteration):
```
DimitrisJim, line 207: After raising it won't continue iterating, as expected.

extra_tests/snippets/set.py
```diff
@@ -195,16 +195,18 @@ def __hash__(self):
...
 
 a = set([1, 2, 3])
 i = iter(a)
 a.add(4)
 with assert_raises(RuntimeError):
     next(i)
 a.remove(4)
-with assert_raises(RuntimeError):
-    next(i)
+# TODO: Raises RuntimeError in CPython but we currently raise Runtime error.
```
DimitrisJim, line 207: Update: After checking CPython, seems like this continues raising `RuntimeError`. We currently have no easy way of encoding this information for dicts and sets, I'll look into it at a following PR.

extra_tests/snippets/set.py
```diff
@@ -191,20 +191,15 @@ def __hash__(self):
 with assert_raises(TypeError):
 	a ^= [1,2,3]
 
-a = set([1, 2, 3])
```
DimitrisJim, line 194: Note on this:

Set behaves the same as dictionaries and detects mutations even when they are balanced, i.e:

```
d = {1: 2, 3: 4}
dit = iter(d)
d[5] = 6
del d[6]
next(dit)  # RuntimeError
```

CPython doesn't detect any of these (neither for sets nor for dicts) but openly states in [PEP 3106](https://www.python.org/dev/peps/pep-3106/#id5) that mutations while iterating is undefined behavior:

> As in Python 2.x, mutating a dict while iterating over it using an iterator has an undefined effect and will in most cases raise a `RuntimeError` exception. (This is similar to the guarantees made by the Java Collections Framework.) 

So I'm pretty sure that a `RuntimeError` is the correct approach here.

DimitrisJim: The fact that we detect it while CPython doesn't does make me believe we're currently doing more book-keeping than them and possibly slowing things down. I'll try and look into that too.

extra_tests/snippets/set.py
youknowone: Do you need to work more on here?

youknowone (approved):

extra_tests/snippets/set.py
DimitrisJim: Yup, I'll try and match CPythons behavior here but I'm leaving it for another PR since the dict iterators also need to be tweaked (and the test for them needs to be added in `extra_tests`).

PR #2992, opened 2021-08-30
Correct `object.__repr__` format

moreal:
## Overview

Before this pull request, there was a bug where `object`'s repr message is not compatible with CPython implementation.

When you run the below code:

```python
class A:
    def a(self):
        class C:
            ...
        print(repr(C()))

print(repr(A()))
A().a()
```

CPython implementation prints the following lines:

```
<__main__.A object at 0x7f9411fca7f0>
<__main__.A.a.<locals>.C object at 0x7f9411fca790>
```

But RustPython implementation does:

```
<A object at 0x555d487e75c0>
<C object at 0x555d487bdc20>
```

## Reason

It was because `object.__repr__` used only `__name__` without `__module__`, `__qualname__`. 😢 

https://github.com/RustPython/RustPython/blob/f1e81c991e3a0eec3708110075e6bb5a3c280e72/vm/src/builtins/object.rs#L185-L189

## References

- [`object.__repr__` CPython implementation](https://github.com/python/cpython/blob/7e246a3a7b43762480ee4fe0cfb859e8e997a8c8/Objects/typeobject.c#L4517-L4544)

### DIFF
--- a/vm/src/builtins/object.rs
+++ b/vm/src/builtins/object.rs
@@ -185,7 +185,11 @@ impl PyBaseObject {
     /// Return repr(self).
     #[pymethod(magic)]
     fn repr(zelf: PyObjectRef) -> String {
-        format!("<{} object at {:#x}>", zelf.class().name, zelf.get_id())
+        format!(
+            "<{} object at {:#x}>",
+            zelf.class().tp_name(),
+            zelf.get_id()
+        )
     }
 
     #[pyclassmethod(magic)]
--- a/Lib/test/test_reprlib.py
+++ b/Lib/test/test_reprlib.py
@@ -318,8 +318,6 @@ class bar:
         # Module name may be prefixed with "test.", depending on how run.
         self.assertEqual(repr(bar.bar), "<class '%s.bar'>" % bar.__name__)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_instance(self):
         self._check_path_limitations('baz')
         write_file(os.path.join(self.subpkgname, 'baz.py'), '''\
@@ -332,8 +330,6 @@ class baz:
         self.assertTrue(repr(ibaz).startswith(
             "<%s.baz object at 0x" % baz.__name__))
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_method(self):
         self._check_path_limitations('qux')
         eq = self.assertEqual


vm/src/builtins/object.rs
```diff
@@ -184,8 +185,25 @@ impl PyBaseObject {
 
     /// Return repr(self).
     #[pymethod(magic)]
-    fn repr(zelf: PyObjectRef) -> String {
-        format!("<{} object at {:#x}>", zelf.class().name, zelf.get_id())
+    fn repr(zelf: PyObjectRef, vm: &VirtualMachine) -> String {
+        let class = zelf.class();
+        let module = class.module(vm);
+        let module = module.downcast_ref::<PyStr>().map(|m| m.as_str());
+        match module {
```
DimitrisJim, line 192: You can just combine these lines instead of re-binding to `module`:

```suggestion
        match class.module(vm).downcast_ref::<PyStr>().map(|m| m.as_str()) {
```

vm/src/builtins/pytype.rs
```diff
@@ -320,7 +320,7 @@ impl PyType {
     }
 
     #[pyproperty(magic)]
-    fn qualname(&self, vm: &VirtualMachine) -> PyObjectRef {
+    pub fn qualname(&self, vm: &VirtualMachine) -> PyObjectRef {
```
DimitrisJim, line 323: Better to have these `pub(crate)` for now and if required they can be bumped to `pub` later.

vm/src/builtins/object.rs
```diff
@@ -184,8 +185,23 @@ impl PyBaseObject {
...
+            Some(module) if module != "builtins" => {
+                format!(
+                    "<{}.{} object at {:#x}>",
+                    module,
+                    class
+                        .qualname(vm)
+                        .downcast_ref::<PyStr>()
+                        .map(|n| n.as_str())
+                        .unwrap_or_else(|| class.name.as_str()),
+                    zelf.get_id()
```
youknowone, line 200: is this also different from `PyType::tp_name()`? Maybe because builtins module.

vm/src/builtins/pytype.rs
```diff
@@ -337,7 +337,7 @@ impl PyType {
     }
 
     #[pyproperty(magic)]
-    fn module(&self, vm: &VirtualMachine) -> PyObjectRef {
+    pub fn module(&self, vm: &VirtualMachine) -> PyObjectRef {
```
DimitrisJim, line 340: I should've made this clearer in the last comment, this should also be `pub(crate)`.

vm/src/builtins/object.rs
moreal: I'm so sorry but I couldn't understand your comment in sure in my English skill. what points are different between those lines and `PyType::tp_name`? Do you mean that `PyType::tp_name` is also different from CPython `tp_name`?

youknowone: Sorry. I just ran test if it is worth to discuss or not.

`class.tp_name()` is not just name(`class.name()`) but `{module}.{name}` by itself. So If the fallback case is using `class.tp_name()`, I think we don't need to use match expression here.

vm/src/builtins/object.rs
```diff
@@ -184,8 +185,23 @@ impl PyBaseObject {
...
+                )
+            }
+            _ => format!("<{} object at {:#x}>", class.tp_name(), zelf.get_id()),
+        }
```
youknowone, line 204:
```suggestion
    format!("<{} object at {:#x}>", class.tp_name(), zelf.get_id())
```
So this code must be same as now

youknowone (changes requested):

vm/src/builtins/object.rs
moreal: Ah, perhaps, I understood well. Actually I suspected the `PyType::tp_name()`.

While researching this issue, I tried to change `format!("<{} object at {:#x}>", zelf.name.clone(), zelf.get_id())` into `format!("<{} object at {:#x}>", class.tp_name(), zelf.get_id())` but the `PyType::tp_name()` hasn't returned `{module}.{qualname}` (i.e. `{module}.{full_path}`) but `{module}.{class_name}`.

For instance, if there are codes like below and I changed `format!("<{} object at {:#x}>", zelf.name.clone(), zelf.get_id())` into `format!("<{} object at {:#x}>", class.tp_name(), zelf.get_id())`:

```python
class F:
    def f(self):
        class A: ...

        print(repr(A()))

F().f()
```

We expected `<__main__.F.f.<locals>.A object at 0x7fed36f58790>` but `<__main__.A object at 0x55f0665da830>`.

So I tried to correct `PyType::tp_name()` but I confused while comparing with CPython because CPython implementation stores `tp_name` and derives `name` from the `tp_name` while RustPython derives `tp_name` from `name`. Is it intended? I don't have opinion which it is bad but I wonder the reason 🤔

youknowone: I didn't investigate CPython code yet, but I am pretty sure `{module}.{class_name}` is a bug. If that was the reason, I think this code must simply use `tp_name`, and then we also need to fix tp_name later for compatibility. How do you think about it?

The difference between RustPython and CPython tp_name is not made by design. So it is not intended.

moreal: I understood and agree with your opinion that this code must use `tp_name` and we should fix `tp_name`.

I will remove all commits and add a commit to replace `zelf.class().name` with `zelf.class().tp_name()`. And I will continue this pull request with direction to fix the `tp_name`.

youknowone: Hou about finishing this PR by replacing it to tp_name and fix tp_name later in different patch?

moreal: Ok, it seems better 👌

moreal:
Through [above conversation thread](https://github.com/RustPython/RustPython/pull/2992#discussion_r699003031), I decided to finish this pull request by replace `zelf.class().name` with `zelf.class().tp_name()`. cc @DimitrisJim

youknowone:
@moreal there is a few unexpected successes

youknowone (approved):

PR #2994, opened 2021-08-31
Fix `PyMappingProxy::get`

moreal:
# Overview

When we run the Python code like below:

```python
class A: ...

print(A.__dict__.items())
print(A.__dict__.get("__hash__", "NONE"))
```

We can't find `__hash__` anywhere and will see `NONE` output because there is no `__hash__` method.

Let's see the output from the execution result of CPython implementation.

```
dict_items([('__module__', '__main__'), ('__dict__', <attribute '__dict__' of 'A' objects>), ('__weakref__', <attribute '__weakref__' of 'A' objects>), ('__doc__', None)])
NONE
```

But RustPython implementation fetches the `object.__hash__` method though it doesn't have.

```
dict_items([('__dict__', <getset_descriptor object at 0x556b113c8460>), ('__module__', '__main__'), ('__qualname__', 'A'), ('__doc__', None)])
<method '__hash__' of 'object' objects>
```

So this pull request fixed the bug, similar with #2932.

### DIFF
--- a/Lib/test/test_typechecks.py
+++ b/Lib/test/test_typechecks.py
@@ -50,8 +50,6 @@ class TypeChecksTest(unittest.TestCase):
         self.assertEqual(issubclass(Integer, Integer), True)
         self.assertEqual(issubclass(Integer, (Integer,)), True)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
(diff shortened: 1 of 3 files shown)


youknowone (approved):
nice catch!

PR #2997, opened 2021-09-01
os : implement os.sched_yield, os.sched_get_priority_min, os.sched_get_priority_max

zetwhite:
This PR contains implementation of `os.sched_yield`, `os.sched_get_priority_min` and `os.sched_get_priority_max`. 
You can check the interface of each function here : https://docs.python.org/3/library/os.html#interface-to-the-scheduler 

Inside each function, related system call is called through the libc Crate. 
with this code, rust python works as below. 
```python 
>>>>> import os 
>>>>> os.sched_get_priority_min(os.SCHED_RR)
1
>>>>> os.sched_get_priority_max(os.SCHED_RR)
99
>>>>> os.sched_get_priority_min(os.SCHED_OTHER)
0
>>>>> os.sched_get_priority_max(os.SCHED_OTHER)
0
>>>>> os.sched_yield()
```


DimitrisJim:
These are defined in the `posix` afaik and not `os`, `os` just imports them from `posix`. There should be a `mod posix` in that file where you can add them similarly to how `os.nice` was added in #2935

youknowone:
Is this unix api or linux api? if it is unix apis, we don't need linux check.

vm/src/stdlib/os.rs
```diff
@@ -2369,6 +2372,38 @@ mod posix {
...
+    }
+
+    #[cfg(target_os = "linux")]
+    #[pyfunction]
+    fn sched_yield(vm: &VirtualMachine) -> PyResult<()> {
+        let res = nix::sched::sched_yield();
+        match res {
+            Ok(_) => Ok(()),
+            Err(err) => Err(err.into_pyexception(vm)),
+        }
```
youknowone, line 2404:
```suggestion
        let _ = nix::sched::sched_yield().map_err(|e| e.into_pyexception(vm))?;
        Ok(())
```

youknowone:
huh, I have no idea why macos build fails on CI. Let's go back to linux only for now. I am sorry. 😅

zetwhite:
Okay. I understood. 
I will check how it works on other platforms through the libc documentation(https://rust-lang.github.io/libc/#platform-specific-documentation) and fix it! Thank you

zetwhite:
About `sched_get_priority_min` , `sched_get_priority_max`, as you can see, i just added `not(target_os = redox)`. 

About scheduling policies such as `SCHED_RR`, `SCHED_FIFO`, that can be much different depending on the OS.
In CPython, `ifdef` is used as below, but I have no idea how to do that in Rust. 
So, I just added each OS in `#[cfg(any(~)]` to make it works based on whitelist. 
```c
#ifdef SCHED_OTHER
    if (PyModule_AddIntMacro(m, SCHED_OTHER)) return -1;
#endif
#ifdef SCHED_FIFO
    if (PyModule_AddIntMacro(m, SCHED_FIFO)) return -1;
#endif
#ifdef SCHED_RR
    if (PyModule_AddIntMacro(m, SCHED_RR)) return -1;
#endif
```
https://github.com/python/cpython/blob/main/Modules/posixmodule.c#L15267-L15300

youknowone:
`SCHED_RR` in macos is defined in `<pthread.h>` but not in `libc`. I created a patch about it (https://github.com/rust-lang/libc/pull/2384) but it will take time.

youknowone (approved):

PR #2999, opened 2021-09-01
(WIP) implement async for function

HyeockJinKim:


HyeockJinKim:
Compilation for async for is done based on the current PR. However, panic is occurring in the part executing the generated bytecode.

fanninpm:
There are at least two tests in [`Lib/test/test_ast.py`](https://github.com/RustPython/RustPython/blob/main/Lib/test/test_ast.py) that can get un-skipped if this works.

PR #3001, opened 2021-09-02
SlotConstructor and uniform tp_new signature

eldpswp99:
closes #2985 by implement them to modules using tp_new which does not have signature with
`tp_new(PyTypeRef, FuncArgs, &VirtualMachine) -> PyResult`

but it panics with vm initaliztion. with `thread 'main' panicked at 'removal index (is 0) should be < len (is 0)'`
due to the panic, i open with draft and work on it.

I would be thankful for any advice!

### DIFF
--- a/vm/src/builtins/list.rs
+++ b/vm/src/builtins/list.rs
@@ -16,7 +16,7 @@ use super::PyInt;
 use crate::common::lock::{
     PyMappedRwLockReadGuard, PyRwLock, PyRwLockReadGuard, PyRwLockWriteGuard,
 };
-use crate::function::OptionalArg;
+use crate::function::{FuncArgs, OptionalArg};
 use crate::sequence::{self, SimpleSeq};
 use crate::sliceable::{PySliceableSequence, PySliceableSequenceMut, SequenceIndex};
 use crate::slots::{Comparable, Hashable, Iterable, PyComparisonOp, PyIter, Unhashable};
@@ -397,12 +397,8 @@ impl PyList {
     }
 
     #[pyslot]
-    fn tp_new(
-        cls: PyTypeRef,
-        _iterable: OptionalArg<PyObjectRef>,
-        vm: &VirtualMachine,
-    ) -> PyResult<PyRef<Self>> {
-        PyList::default().into_ref_with_type(vm, cls)
+    fn tp_new(cls: PyTypeRef, _args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+        PyList::default().into_pyresult_with_type(vm, cls)
     }
 
     #[pymethod(magic)]
--- a/vm/src/builtins/map.rs
+++ b/vm/src/builtins/map.rs
@@ -1,7 +1,7 @@
 use super::pytype::PyTypeRef;
 use crate::function::Args;
 use crate::iterator;
-use crate::slots::PyIter;
+use crate::slots::{PyIter, SlotConstructor};
 use crate::vm::VirtualMachine;
 use crate::{PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue};
 
@@ -22,15 +22,10 @@ impl PyValue for PyMap {
     }
 }
 
-#[pyimpl(with(PyIter), flags(BASETYPE))]
-impl PyMap {
-    #[pyslot]
-    fn tp_new(
-        cls: PyTypeRef,
-        function: PyObjectRef,
-        iterables: Args,
-        vm: &VirtualMachine,
-    ) -> PyResult<PyRef<Self>> {
+impl SlotConstructor for PyMap {
+    type Args = (PyObjectRef, Args<PyObjectRef>);
+
+    fn py_new(cls: PyTypeRef, (function, iterables): Self::Args, vm: &VirtualMachine) -> PyResult {
         let iterators = iterables
             .into_iter()
             .map(|iterable| iterator::get_iter(vm, iterable))
@@ -39,10 +34,12 @@ impl PyMap {
             mapper: function,
             iterators,
         }
-        .into_ref_with_type(vm, cls)
+        .into_pyresult_with_type(vm, cls)
     }
+}
 
-    /// Return an estimate of the number of items.
+#[pyimpl(with(PyIter, SlotConstructor), flags(BASETYPE))]
+impl PyMap {
     #[pymethod(magic)]
     fn length_hint(&self, vm: &VirtualMachine) -> PyResult<usize> {
         self.iterators.iter().try_fold(0, |prev, cur| {
--- a/vm/src/builtins/object.rs
+++ b/vm/src/builtins/object.rs
@@ -11,7 +11,7 @@ use crate::utils::Either;
 use crate::vm::VirtualMachine;
 use crate::{
     IdProtocol, ItemProtocol, PyArithmaticValue, PyAttributes, PyClassImpl, PyComparisonValue,
-    PyContext, PyObject, PyObjectRef, PyResult, PyValue, TryFromObject, TypeProtocol,
+    PyContext, PyObject, PyObjectRef, PyResult, PyValue, TypeProtocol,
 };
 
 /// object()
@@ -35,9 +35,8 @@ impl PyValue for PyBaseObject {
 impl PyBaseObject {
     /// Create and return a new object.  See help(type) for accurate signature.
     #[pyslot]
-    fn tp_new(mut args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+    fn tp_new(cls: PyTypeRef, _args: FuncArgs, vm: &VirtualMachine) -> PyResult {
         // more or less __new__ operator
-        let cls = PyTypeRef::try_from_object(vm, args.shift())?;
         let dict = if cls.is(&vm.ctx.types.object_type) {
             None
         } else {
--- a/vm/src/builtins/singletons.rs
+++ b/vm/src/builtins/singletons.rs
@@ -1,6 +1,9 @@
 use super::pytype::PyTypeRef;
+use crate::slots::SlotConstructor;
 use crate::vm::VirtualMachine;
-use crate::{IntoPyObject, PyClassImpl, PyContext, PyObjectRef, PyRef, PyValue, TypeProtocol};
+use crate::{
+    IntoPyObject, PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
+};
 
 #[pyclass(module = false, name = "NoneType")]
 #[derive(Debug)]
@@ -30,13 +33,16 @@ impl<T: IntoPyObject> IntoPyObject for Option<T> {
     }
 }
 
-#[pyimpl]
-impl PyNone {
-    #[pyslot]
-    fn tp_new(_: PyTypeRef, vm: &VirtualMachine) -> PyRef<Self> {
-        vm.ctx.none.clone()
-    }
+impl SlotConstructor for PyNone {
+    type Args = ();
 
+    fn py_new(_: PyTypeRef, _args: Self::Args, vm: &VirtualMachine) -> PyResult {
+        Ok(vm.ctx.none.clone().into_object())
+    }
+}
+
+#[pyimpl(with(SlotConstructor))]
+impl PyNone {
     #[pymethod(magic)]
     fn repr(&self) -> String {
         "None".to_owned()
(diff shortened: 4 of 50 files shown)


vm/src/builtins/object.rs
```diff
@@ -35,7 +35,7 @@ impl PyValue for PyBaseObject {
 impl PyBaseObject {
     /// Create and return a new object.  See help(type) for accurate signature.
     #[pyslot]
-    fn tp_new(mut args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+    fn tp_new(_cls: PyTypeRef, mut args: FuncArgs, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 38: because `cls` is provided outside, args decomposition need to be removed.

eldpswp99: i removed line 40, and changed _cls to cls to use cls provided from outside.
now the panic with the length gone, but still panics. The Error message is
Error in sys.excepthook:
```
lost sys.stderr
RuntimeError: lost sys.stderr
Original exception was:
lost sys.stderr
  File "_frozen_importlib", line 1094, in __import__
  File "_frozen_importlib", line 1015, in _gcd_import
  File "_frozen_importlib", line 992, in _find_and_load
  File "_frozen_importlib", line 992, in _find_and_load
  File "_frozen_importlib", line 976, in _find_and_load_unlocked
  File "_frozen_importlib", line 687, in _load_unlocked
  File "_frozen_importlib", line 685, in _load_unlocked
  File "_frozen_importlib", line 678, in _load_unlocked
  File "_frozen_importlib", line 672, in _load_unlocked
  File "_frozen_importlib_external", line 779, in exec_module
  File "_frozen_importlib_external", line 907, in get_code
  File "_frozen_importlib_external", line 883, in get_code
  File "_frozen_importlib_external", line 513, in _classify_pyc
BaseException: bad magic number in 'encodings': b'6afe'

RuntimeError: Could not import encodings. Is your RUSTPYTHONPATH set? If you don't have access to a consistent external environment (e.g. if you're embedding rustpython in another application), try enabling the freeze-stdlib feature
thread 'main' panicked at 'initialization failed; exception backtrace above', vm/src/vm.rs:808:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/panicking.rs:493:5
   1: std::panicking::begin_panic_fmt
             at /rustc/53cb7b09b00cbea8754ffb78e7e3cb521cb8af4b/library/std/src/panicking.rs:435:5
   2: rustpython_vm::vm::VirtualMachine::_py_panic_failed
             at ./vm/src/vm.rs:808:13
   3: rustpython_vm::vm::VirtualMachine::expect_pyresult::{{closure}}
             at ./vm/src/vm.rs:830:37
   4: core::result::Result<T,E>::unwrap_or_else
             at /home/eldpswp99/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/result.rs:821:23
   5: rustpython_vm::vm::VirtualMachine::expect_pyresult
             at ./vm/src/vm.rs:830:9
   6: rustpython_vm::vm::VirtualMachine::initialize
             at ./vm/src/vm.rs:380:9
   7: rustpython_vm::vm::Interpreter::new_with_init
             at ./vm/src/vm.rs:2122:9
   8: rustpython::run
             at ./src/lib.rs:95:18
   9: rustpython::main
             at ./src/main.rs:2:5
  10: core::ops::function::FnOnce::call_once
             at /home/eldpswp99/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ops/function.rs:227:5

```

vm/src/exceptions.rs
```diff
@@ -63,12 +63,8 @@ impl PyBaseException {
     }
 
     #[pyslot]
-    pub(crate) fn tp_new(
-        cls: PyTypeRef,
-        args: FuncArgs,
-        vm: &VirtualMachine,
-    ) -> PyResult<PyRef<Self>> {
-        PyBaseException::new(args.args, vm).into_ref_with_type(vm, cls)
```
youknowone, line 71: The error happened in ImportError. I think that's because `.into_ref_with_type(vm, cls)` is removed here so it lost its type .

vm/src/builtins/tuple.rs
```diff
@@ -254,36 +283,6 @@ impl PyTuple {
...
-            vec![]
-        };
-        // Return empty tuple only for exact tuple types if the iterable is empty.
-        if elements.is_empty() && cls.is(&vm.ctx.types.tuple_type) {
-            Ok(vm.ctx.empty_tuple.clone())
-        } else {
-            Self {
-                elements: elements.into_boxed_slice(),
-            }
-            .into_ref_with_type(vm, cls)
```
youknowone, line 284: I found `into_ref_with_type` is removed here and there. They must be keeped to provide subclass type information.
`.into_ref_with_type(vm, cls).into_pyresult(vm)` is expected

eldpswp99: actually, i removed all `into_ref_with_type` due to match the signature.

eldpswp99: but only making the two function to have `into_ref_with_type`, the problem fixed. Can you explain some detail of it?

youknowone: `into_ref_with_type` explicitly set the type of the instance while `into_ref` set the type by its payload. `into_pyresult` internally calls `into_pyobject` which also calls `into_ref` by needs. So `into_ref_with_type` is not called, any subclass of the types will lose its actual type while `into_ref` is implicitly called.

eldpswp99: i got it thanks! I think removing all `into_ref_with_type` caused so many problems.

vm/src/stdlib/itertools.rs
```diff
@@ -1028,23 +1114,17 @@ mod decl {
         }
     }
 
-    #[derive(FromArgs)]
-    struct ProductArgs {
-        #[pyarg(named, optional)]
-        repeat: OptionalArg<usize>,
-    }
```
youknowone, line 1035: to keep the original behaviour,
`type Args = (Args<PyObjectRef>, ProductArgs);`  will work.

eldpswp99: I would fix it! but putting iterable to Args which derive `FromArgs` cause Error due to Args does not implement `TryFromBorrowedObject` (maybe it does not have trait Sized?). To remain the original, I would use the first suggestion

youknowone: Not sure why it looked like related to TryFormBorrowedObject, but I think `vm::function::Args` can cause some problem.
The first approach will work.

```rust
    #[derive(FromArgs)]
    struct ProductArgs {
        #[pyarg(named, optional)]
        repeat: OptionalArg<usize>,
    }

    impl SlotConstructor for PyItertoolsProduct {
        type Args = (Args<PyObjectRef>, ProductArgs);
        fn py_new(cls: PyTypeRef, (iterables, args): Self::Args, vm: &VirtualMachine) -> PyResult {
            let repeat = args.repeat.unwrap_or(1);
            
```

vm/src/builtins/map.rs
```diff
@@ -24,13 +24,19 @@ impl PyValue for PyMap {
...
+    }
+
     #[pyslot]
-    fn tp_new(
-        cls: PyTypeRef,
-        function: PyObjectRef,
-        iterables: Args,
-        vm: &VirtualMachine,
-    ) -> PyResult<PyRef<Self>> {
+    fn tp_new(cls: PyTypeRef, mut args: FuncArgs, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 37: I think turning `FnucArgs`-less `tp_new`  to use `FuncArgs` is generally not a good idea. `SlotConstructor` can keep the original argument without code change by defining proper `SlotCosntructor::Args`, mostly just tuple of the types.

eldpswp99: I strongly agree with it. to be honest i didn't know the way using handles.

vm/src/pyobject.rs
```diff
@@ -1016,6 +1016,10 @@ pub trait PyValue: fmt::Debug + PyThreadingConstraint + Sized + 'static {
             )))
         }
     }
+
+    fn into_pyresult_ref_with_type(self, vm: &VirtualMachine, cls: PyTypeRef) -> PyResult {
```
youknowone, line 1020: `ref` in `into_ref` means `PyRef`. So `into_pyresult_with_type` will be enough here. `into_pyobject_with_type` would be also good because `into_pyobject` will be always successful

eldpswp99: i get the first one, `into_pyresult_with_type`! Thank you!

vm/src/builtins/singletons.rs
```diff
@@ -33,8 +36,8 @@ impl<T: IntoPyObject> IntoPyObject for Option<T> {
 #[pyimpl]
 impl PyNone {
     #[pyslot]
-    fn tp_new(_: PyTypeRef, vm: &VirtualMachine) -> PyRef<Self> {
-        vm.ctx.none.clone()
+    fn tp_new(_: PyTypeRef, _args: FuncArgs, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 39: Surprisingly, some types are too generous for `__new__` arguments, so they don't check them at all. But the other types check it. `None`, `Ellipsis` and `NotImplemented` are those types, so args type must be `()`

vm/src/stdlib/io.rs
```diff
@@ -3068,23 +3061,25 @@ mod _io {
...
+    struct StringIONewArgs {
+        #[pyarg(positional, optional)]
+        object: OptionalOption<PyStrRef>,
 
-        #[pyslot]
-        fn tp_new(
+        // TODO: use this
+        #[pyarg(any, default)]
+        #[allow(dead_code)]
+        _newline: Newlines,
```
youknowone, line 3072: the argument name must be `newline` without `_` because tests expect the exact keyword argument name

vm/src/builtins/list.rs
```diff
@@ -397,12 +397,8 @@ impl PyList {
     }
 
     #[pyslot]
-    fn tp_new(
-        cls: PyTypeRef,
-        _iterable: OptionalArg<PyObjectRef>,
-        vm: &VirtualMachine,
-    ) -> PyResult<PyRef<Self>> {
-        PyList::default().into_ref_with_type(vm, cls)
+    fn tp_new(cls: PyTypeRef, _iterable: FuncArgs, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 400:
```suggestion
    fn tp_new(cls: PyTypeRef, _args: FuncArgs, vm: &VirtualMachine) -> PyResult {
```
This change enhances compatibility :+1:, but the argument name doesn't fit to its actual value anymore

vm/src/builtins/weakref.rs
```diff
@@ -43,18 +51,24 @@ impl Callable for PyWeak {
...
-    ) -> PyResult<PyRef<Self>> {
-        PyWeak::downgrade(&referent).into_ref_with_type(vm, cls)
+    ) -> PyResult {
+        PyWeak::downgrade(&referent).into_pyresult_with_type(vm, cls)
     }
+}
+
+#[pyimpl(with(Callable, Hashable, Comparable, SlotConstructor), flags(BASETYPE))]
+impl PyWeak {
+    // TODO callbacks
```
youknowone, line 71: this comment was about `__new__`. So Line 65 looks the correct new place.

vm/src/stdlib/array.rs
```diff
@@ -656,7 +655,7 @@ mod array {
                 } else if let Some(utf8) = init.payload::<PyStr>() {
                     if spec == 'u' {
                         let bytes = Self::_unicode_to_wchar_bytes(utf8.as_str(), array.itemsize());
-                        array.frombytes_move(bytes);
+                        array.frombytes(&bytes);
```
youknowone, line 658: this line looks like unintended change.

vm/src/stdlib/itertools.rs
```diff
@@ -1258,9 +1350,12 @@ mod decl {
                 r: AtomicCell::new(r),
                 exhausted: AtomicCell::new(n == 0 && r > 0),
             }
-            .into_ref_with_type(vm, cls)
+            .into_pyresult_with_type(vm, cls)
         }
     }
+
+    #[pyimpl(with(PyIter, SlotConstructor))]
+    impl PyItertoolsCombinationsWithReplacement {}
```
youknowone, line 1358:
```suggestion
    impl PyItertoolsCombinationsWithReplacement {}

```

youknowone:
I think this is almost done! @eldpswp99 would you rebase this PR to remove the broken miri test and check latest comments?

youknowone (approved):

PR #3003, opened 2021-09-03
Add thread_time, process_time in `time` module

Snowapril:
This revision add four methods, `thread_time`, `thread_time_ns`, `process_time` and `process_time_ns`

### Documents
* https://docs.python.org/3.8/library/time.html#time.process_time
* https://docs.python.org/3.8/library/time.html#time.thread_time
* https://docs.rs/libc/0.2.92/libc/fn.clock_gettime.html
* https://docs.rs/libc/0.2.92/libc/constant.CLOCK_PROCESS_CPUTIME_ID.html
* https://docs.rs/libc/0.2.92/libc/constant.CLOCK_THREAD_CPUTIME_ID.html
* https://docs.rs/winapi/0.3.9/winapi/um/processthreadsapi/fn.GetProcessTimes.html
* https://docs.rs/winapi/0.3.9/winapi/um/processthreadsapi/fn.GetThreadTimes.html

### DIFF
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -14,6 +14,30 @@ use crate::utils::Either;
 use crate::vm::VirtualMachine;
 use crate::{PyClassImpl, PyObjectRef, PyResult, PyStructSequence, TryFromObject};
 
+#[cfg(windows)]
+use winapi::shared::{minwindef::FILETIME, ntdef::ULARGE_INTEGER};
+#[cfg(windows)]
+use winapi::um::processthreadsapi::{
+    GetCurrentProcess, GetCurrentThread, GetProcessTimes, GetThreadTimes,
+};
+
+#[allow(dead_code)]
+const SEC_TO_MS: i64 = 1000;
+#[allow(dead_code)]
+const MS_TO_US: i64 = 1000;
+#[allow(dead_code)]
+const SEC_TO_US: i64 = SEC_TO_MS * MS_TO_US;
+#[allow(dead_code)]
+const US_TO_NS: i64 = 1000;
+#[allow(dead_code)]
+const MS_TO_NS: i64 = MS_TO_US * US_TO_NS;
+#[allow(dead_code)]
+const SEC_TO_NS: i64 = SEC_TO_MS * MS_TO_NS;
+#[allow(dead_code)]
+const NS_TO_MS: i64 = 1000 * 1000;
+#[allow(dead_code)]
+const NS_TO_US: i64 = 1000;
+
 #[cfg(unix)]
 fn time_sleep(dur: Duration, vm: &VirtualMachine) -> PyResult<()> {
     // this is basically std::thread::sleep, but that catches interrupts and we don't want to;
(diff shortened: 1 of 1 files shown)


Snowapril:
I cant figure out getting process time and thread time in window platform without [`window-rs`](https://github.com/microsoft/windows-rs dependency),,, hmm..

Snowapril:
Uhm.. maybe #3005 must be preceeded 😂

vm/src/stdlib/time.rs
```diff
@@ -14,6 +14,16 @@ use crate::utils::Either;
...
 use crate::{PyClassImpl, PyObjectRef, PyResult, PyStructSequence, TryFromObject};
 
+#[cfg(windows)]
+use winapi::shared::{minwindef::FILETIME, ntdef::ULARGE_INTEGER};
+#[cfg(windows)]
+use winapi::um::processthreadsapi::{
+    GetCurrentProcess, GetCurrentThread, GetProcessTimes, GetThreadTimes,
+};
+
+#[cfg(not(windows))]
```
youknowone, line 24:
```suggestion
#[cfg(not(any(windows, target_arch = "wasm32")))]
```

youknowone (changes requested):
Yes, I think something broken by that. But wasm fail is not from that one. Would you fix this line and rebase it to upstream/main to see if it is fixed?

youknowone:
Exotic platforms CI must be also checked. `CLOCK_PROCESS_CPUTIME_ID`, `CLOCK_THREAD_CPUTIME_ID` doesnt' exist in redox.

Snowapril:
Yeah... I'm currently working on that issue. I found there are various cases where `clock_gettime` or `CLOCK_PROCESS_CPUTIME_ID` did not exist.

vm/src/stdlib/time.rs
```diff
@@ -173,6 +197,169 @@ fn time_strptime(
...
+    };
+    let u = match tv.tv_usec.checked_mul(US_TO_NS) {
+        Some(value) => value,
+        None => {
+            return Err(vm.new_overflow_error("timestamp too large to convert to i64".to_owned()));
+        }
+    };
+
+    t.checked_add(u)
+        .ok_or_else(|| vm.new_overflow_error("timestamp too large to convert to i64".to_owned()))
```
youknowone, line 316:
```suggestion
```suggestion
    (|tv| {
        let t = tv.tv_sec.checked_mul(SEC_TO_NS)?;
        let u = tv.tv_usec.checked_mul(US_TO_NS)?;
        t.checked_add(u)
    })(tv).ok_or_else(|| vm.new_overflow_error("timestamp too large to convert to i64".to_owned()))
```

I didnt' test the code but this kind of code will work

vm/src/stdlib/time.rs
```diff
@@ -173,6 +197,169 @@ fn time_strptime(
...
+fn get_process_time(vm: &VirtualMachine) -> PyResult<Duration> {
+    cfg_if::cfg_if! {
+        if #[cfg(any(
+            target_os = "wasi",
+            target_os = "redox",
+            target_os = "illumos",
+            target_os = "netbsd",
+            target_os = "solaris"
+        ))] {
+            let mut ru: libc::rusage = std::mem::zeroed();
```
youknowone, line 329:
```suggestion
            let mut ru: libc::rusage = unsafe { std::mem::zeroed() };
```
zeroed is unsafe

vm/src/stdlib/time.rs
```diff
@@ -173,6 +197,169 @@ fn time_strptime(
...
 
+fn get_thread_time(vm: &VirtualMachine) -> PyResult<Duration> {
+    cfg_if::cfg_if! {
+        if #[cfg(windows)] {
+            let mut creation_time = FILETIME::default();
+            let mut exit_time = FILETIME::default();
+            let mut kernel_time = FILETIME::default();
+            let mut user_time = FILETIME::default();
+
+            unsafe {
```
youknowone, line 208: this block looks like containing some safe codes

vm/src/stdlib/time.rs
```diff
@@ -173,6 +197,169 @@ fn time_strptime(
...
+        }
+    }
+}
+
+fn time_thread_time(vm: &VirtualMachine) -> f64 {
+    match get_thread_time(vm) {
+        Ok(v) => v.as_secs_f64(),
+        Err(err) => panic!("Time error: {:?}", err),
+    }
+}
```
youknowone, line 253: By looking inside `get_thread_time` the `Err` case looks able to be raised practically.

```suggestion
fn time_thread_time(vm: &VirtualMachine) -> PyResult<f64> {
    Ok(get_thread_time(vm)?.as_secs_f64())
}
```
And also other functions

vm/src/stdlib/time.rs
```diff
@@ -173,6 +197,169 @@ fn time_strptime(
...
+        }
+
+        let mut large: ULARGE_INTEGER = std::mem::zeroed();
+        large.u_mut().LowPart = kernel_time.dwLowDateTime;
+        large.u_mut().HighPart = kernel_time.dwHighDateTime;
+        let ktime = *large.QuadPart();
+
+        large.u_mut().LowPart = user_time.dwLowDateTime;
+        large.u_mut().HighPart = kernel_time.dwHighDateTime;
+        let utime = *large.QuadPart();
```
youknowone, line 288: it seems sharing some code with `get_thread_time`. how about adding a utility function for this case?

youknowone (changes requested):

youknowone:
because now we have lots of platform-specific code, defining a module can be convenient.

```rust
#[cfg(windows)]
mod windows {
   ...
}
```

vm/src/stdlib/time.rs
Snowapril: Oh great!! That way can avoid duplicates of error messages

vm/src/stdlib/time.rs
Snowapril: I missed that. Thanks. Is there any way to check grammar works on other platform?

vm/src/stdlib/time.rs
Snowapril: That way i can avoid panic call. Thanks

youknowone (approved):

vm/src/stdlib/time.rs
```diff
@@ -173,6 +197,179 @@ fn time_strptime(
...
+fn time_thread_time(vm: &VirtualMachine) -> PyResult<f64> {
+    Ok(get_thread_time(vm)?.as_secs_f64())
+}
+
+fn time_thread_time_ns(vm: &VirtualMachine) -> PyResult<u64> {
+    Ok(get_thread_time(vm)?.as_nanos() as u64)
+}
+
+#[cfg(windows)]
+unsafe fn time_from_filtime(time: FILETIME) -> u64 {
```
youknowone, line 249: I missed this part. `unsafe` keyword has different meaning for `unsafe fn` and `unsafe {}`. For `unsafe fn`, it means calling the fn is not safe. For `unsafe {}`, it mean the block is using unsafe operations. So in this case, each function calls are unsafe but the function itself is safe.

PR #3008, opened 2021-09-04
Derive `name` from `tp_name`

moreal:
# Overview

This pull request makes RustPython follow the relationship between `tp_name()` and `name()` in CPython implementation. In CPython, there is only `tp_name` slot and derives `name` from the `tp_name` slot. But RustPython had derived `tp_name` from `name` with `module` and `qualname`. This pull request fixes it.

# References

- https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_name (I'm not sure this documentation is correct... 🤔 )
- [`_PyType_Name`](https://github.com/python/cpython/blob/28264269de9ff88d9ee7110fc56ac2d2db275bec/Objects/typeobject.c#L474-L486) (i.e. `type_name`)

### DIFF
--- a/Lib/test/test_reprlib.py
+++ b/Lib/test/test_reprlib.py
@@ -318,6 +318,8 @@ class bar:
         # Module name may be prefixed with "test.", depending on how run.
         self.assertEqual(repr(bar.bar), "<class '%s.bar'>" % bar.__name__)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_instance(self):
         self._check_path_limitations('baz')
         write_file(os.path.join(self.subpkgname, 'baz.py'), '''\
@@ -330,6 +332,8 @@ class baz:
         self.assertTrue(repr(ibaz).startswith(
             "<%s.baz object at 0x" % baz.__name__))
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_method(self):
         self._check_path_limitations('qux')
         eq = self.assertEqual
--- a/vm/src/builtins/pytype.rs
+++ b/vm/src/builtins/pytype.rs
@@ -352,7 +352,6 @@ impl PyType {
 
     #[pyproperty(magic, setter)]
     fn set_module(&self, value: PyObjectRef) {
-        *self.slots.name.write() = None;
         self.attributes
             .write()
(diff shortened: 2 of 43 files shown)


vm/src/vm.rs
```diff
@@ -1011,7 +1021,8 @@ impl VirtualMachine {
             return self.abstract_isinstance(obj, cls);
         }
 
-        if let Ok(tuple) = PyTupleRef::try_from_object(self, cls.clone()) {
+        if cls.class().slots.flags.has_feature(PyTpFlags::TUPLE_SUBCLASS) {
+            let tuple = PyTupleRef::try_from_object(self, cls.clone()).unwrap();
```
DimitrisJim, line 1025: is this related with the change?

moreal: No, it is not. it was a thing should be resolved by 10eb3d0.

vm/src/builtins/pytype.rs
```diff
@@ -61,23 +60,8 @@ impl PyValue for PyType {
...
-                format!("{}.{}", module, &self.name)
-            } else {
-                self.name.clone()
-            };
-            *opt_name = Some(new_name);
-            PyRwLockWriteGuard::downgrade(opt_name)
-        };
-        PyRwLockReadGuard::map(read_guard, |opt| opt.as_deref().unwrap())
+    pub fn tp_name(&self) -> String {
+        self.slots.name.read().as_ref().unwrap().to_string()
```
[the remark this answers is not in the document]
DimitrisJim: yup. seems like there's a couple of other places that still use it though (from a quick `grep`). Maybe leave it as a follow up PR to get this merged now?

Lib/test/test_reprlib.py
```diff
@@ -318,6 +318,8 @@ class bar:
         # Module name may be prefixed with "test.", depending on how run.
         self.assertEqual(repr(bar.bar), "<class '%s.bar'>" % bar.__name__)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
youknowone, line 322: will this regressions be fixed in this PR? module name looks like not to be properly saved.

youknowone: Thank you for detailed explanation!

PR #3010, opened 2021-09-05
add precommit which runs  clippy & fmt

eldpswp99:
add precommit to RustPython
it runs `cargo clippy` / `cargo fmt --all` /  `git add -u` before commit

by running `cargo clean && cargo build`, it adds git hooks to the project.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1506,6 +1506,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ac74c624d6b2d21f425f752262f42188365d7b8ff1aff74c82e45136510a4857"
 
 [[package]]
+name = "pre-commit"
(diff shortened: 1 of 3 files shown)


youknowone (approved):
clippy looks heavy for pre-commit hook, but generally looks good.

@coolreader18 @DimitrisJim how do you think?

DimitrisJim:
I wouldn't really like clippy running on each commit I do (fmt seems fine since its instantaneous). I prefer to run clippy right before I push things and if I'm feeling like I made a relatively big change.

I think it would be perfect if it was optional so people who have a different workflow can opt-out of it but I doubt that's possible.

eldpswp99:
then how about if there is only fmt and git add -u statement?

youknowone:
That sounds better to me too

PR #3011, opened 2021-09-05
better error message when pymethod/pyproperty signature is wrong

youknowone:
Make signature errors show the funciton signature instead of macro attribute

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -1,6 +1,6 @@
 use super::Diagnostic;
 use crate::util::{
-    get_sig, path_eq, pyclass_ident_and_attrs, ClassItemMeta, ContentItem, ContentItemInner,
+    path_eq, pyclass_ident_and_attrs, text_signature, ClassItemMeta, ContentItem, ContentItemInner,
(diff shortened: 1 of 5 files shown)


derive/src/pyclass.rs
```diff
@@ -359,19 +359,19 @@ where
...
+        let ident = &func.sig().ident;
 
         let item_attr = args.attrs.remove(self.index());
         let item_meta = MethodItemMeta::from_attr(ident.clone(), &item_attr)?;
         let py_name = item_meta.method_name()?;
 
-        let sig_doc = args.item.function_or_method_impl().ok().map(|item| {
-            let sig = item.sig();
-            let args: Vec<_> = sig
+        let sig_doc = {
```
DimitrisJim, line 372: this seems likes its based on an old revision. With these new changes though, `get_sig` might be better suited taking a `Signature` immediately (and `gen_module_items` for `FunctionItem` can easily be changed to do pass the sig).

youknowone: thanks! I changed it to take &Signature during rebase

DimitrisJim (approved):

PR #3009, opened 2021-09-04
pymodule for time module

youknowone:
cc @Snowapril 

review link: https://github.com/RustPython/RustPython/pull/3009/files?w=1

PR #3012, opened 2021-09-05
Add time test file

Regexp96:
import CPython's time test 

Reference:
https://github.com/python/cpython/blob/main/Lib/test/test_time.py

### DIFF
--- a/Lib/test/test_time.py
+++ b/Lib/test/test_time.py
@@ -1,0 +1,1045 @@
+from test import support
+import decimal
+import enum
+import locale
+import math
+import platform
+import sys
+import sysconfig
+import time
+import threading
+import unittest
+try:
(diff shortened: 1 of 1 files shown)


DimitrisJim:
One issue: specify the CPython version in your commit message. That way we can keep track of which files might need updating in the future. (Usually checking out the 3.8 or 3.9 branch on CPython and copying things over)

Regexp96:
Cpython version is 3.8.11 
now it's added to commit message

DimitrisJim:
Great. I triggered the actions again since I just merged #3009 that made some big changes to the `time` module. Now we will hopefully get a better picture of what errors appear.

youknowone (approved):

PR #3014, opened 2021-09-05
Solve some lexical error with underscore

leesungbin:
This commit is to pass the test `test_underscore_literals` in `test_grammar.py`.

### DIFF
--- a/parser/src/lexer.rs
+++ b/parser/src/lexer.rs
@@ -315,10 +315,21 @@ where
 
             // 1e6 for example:
             if self.chr0 == Some('e') || self.chr0 == Some('E') {
+                if self.chr1 == Some('_') {
+                    return Err(LexicalError {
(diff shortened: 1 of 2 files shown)


parser/src/lexer.rs
```diff
@@ -315,10 +315,21 @@ where
 
             // 1e6 for example:
             if self.chr0 == Some('e') || self.chr0 == Some('E') {
+                if self.chr1 == Some('_') {
```
DimitrisJim, line 318: These are identical from what I can tell. Probably a good idea to refactor into a function (`check_invalid_underscore`?) that returns the error for you (which for tokenizing a number has the message `'invalid decimal literal'` if I'm not mistaken).

youknowone: @leesungbin will you check @DimitrisJim 's suggestion?

PR #3015, opened 2021-09-05
Remove unnecessary linebreak

leesungbin:
I think the error message has unnecessary linebreak.

When we execute `eval("1.0e_1")`
* in CPython
```bash
>>> eval("1.0e_1")
  File "<string>", line 1
    1.0e_1
       ^
SyntaxError: unexpected EOF while parsing
```

* in RustPython
```bash
>>>>> eval("1.0e_1")
SyntaxError: Invalid Syntax at line 1 column 4
1.0e_1
                 
   ↑
```

### DIFF
--- a/ast/src/location.rs
+++ b/ast/src/location.rs
@@ -30,7 +30,7 @@ impl Location {
             fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
                 write!(
                     f,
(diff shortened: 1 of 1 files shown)


leesungbin:
In `compiler/porcelain/src/lib.rs`, `get_statement` always return statement with linebreak.

DimitrisJim (approved):

PR #3017, opened 2021-09-05
Add os.fchdir builtin function

ChJR:


vm/src/stdlib/os.rs
```diff
@@ -1268,6 +1270,20 @@ mod _os {
         env::set_current_dir(&path.path).map_err(|err| err.into_pyexception(vm))
     }
 
+    #[cfg(not(windows))]
+    #[pyfunction]
+    fn fchdir(fd: PyIntRef, vm: &VirtualMachine) -> PyResult<()> {
```
youknowone, line 1275: this function need to be place in `posix` module. (try `import posix; posix.fchdir` in CPython`)
Then it will be automatically imported in os module.
Then `#[cfg(not(windows))]` is also not required anymore.

youknowone (changes requested):

vm/src/stdlib/os.rs
ChJR: That's a great point! I'll push the refined one.

youknowone (approved):

PR #3018, opened 2021-09-06
Add test_import from CPython 3.8

fanninpm:


fanninpm:
I can't reproduce the CI failures locally.

DimitrisJim:
Hm, I'm actually getting errors for `os.py`. Are things still passing locally for you?

fanninpm:
> Hm, I'm actually getting errors for `test_os.py`. Are things still passing locally for you?

I'm not getting any failures, but I am getting a few instances of `OverflowError: Python int too large to convert to Rust u32`. These seem to happen with `test_setregid_neg1` and `test_setreuid_neg1`.

DimitrisJim:
Bump, I ran these again and indeed I locally cannot repro, something weird was going on when I tried to run tests last time. 

It looks like `test_import` is failing on Windows while `test_unittest` is failing on the others. If these failures are just one or two functions we could probably just skip for CI as we usually do.

PR #3020, opened 2021-09-06
Add deque __imul__ method

### LINKED ISSUE #2839: `collections.deque` doesn't have `__imul__` method.
## Feature

CPython:

```python
>>> d = deque()
>>> d.__imul__(3)
deque([1, 1, 1, 1, 1])
```

RustPython:

```python
>>>>> d = deque()
>>>>> d.__imul__(5)
AttributeError: 'deque' object has no attribute '__imul__'
```

ConnorTroy:
Closes #2839

Add `deque.__imul__()` method

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -59,9 +59,9 @@ mod _collections {
         }
     }
 
-    struct SimpleSeqDeque<'a>(PyRwLockReadGuard<'a, VecDeque<PyObjectRef>>);
+    struct SimpleSeqDeque(VecDeque<PyObjectRef>);
 
-    impl sequence::SimpleSeq for SimpleSeqDeque<'_> {
+    impl sequence::SimpleSeq for SimpleSeqDeque {
         fn len(&self) -> usize {
             self.0.len()
         }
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/collections.rs
```diff
@@ -59,9 +59,9 @@ mod _collections {
         }
     }
 
-    struct SimpleSeqDeque<'a>(PyRwLockReadGuard<'a, VecDeque<PyObjectRef>>);
+    struct SimpleSeqDeque(VecDeque<PyObjectRef>);
```
youknowone, line 62: This is designed to borrow VecDeque but after this change, it will be cloned to make this struct.

ConnorTroy: I switched to clone because it seemed like the easiest way to implement for the write guard as well. If you want I can switch to `
    struct SimpleSeqDeque<'a>(&'a VecDeque<PyObjectRef>);` -- or do you know if there is a trait that `PyRwLockReadGuard` and `PyRwLockWriteGuard` implement which I can use?
```suggestion
    struct SimpleSeqDeque<'a>(&'a VecDeque<PyObjectRef>);
```

vm/src/stdlib/collections.rs
```diff
@@ -471,6 +471,23 @@ mod _collections {
             })
         }
 
+        #[pymethod(magic)]
+        fn imul(zelf: PyRef<Self>, value: isize, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+            let mut new_deque: VecDeque<PyObjectRef> = {
```
DimitrisJim, line 476: No reason to repeat ourselves here though, right? Either `imul` calls `mul` or `mul` calls `imul`, both can work. CPython has `mul` call into `imul` after doing a copy. For now, this can just call into `mul` and then swap the deques. We can time it at some later point to see if the opposite yields any better results (though I'm guessing the difference would be negligible).

DimitrisJim (approved):
lgtm! thanks!

PR #3023, opened 2021-09-08
Implement `syslog` module for unix

Snowapril:
This revision implement [`syslog`](https://docs.python.org/3.8/library/syslog.html) module.

```python
import syslog
syslog.syslog(syslog.LOG_INFO, "Hello syslog")
```
```bash
>> cat /var/log/syslog
...
Sep  8 22:02:25 DESKTOP-8N***B3 rustpython: Hello syslog
```

### Documentation
* https://docs.python.org/3.8/library/syslog.html
* https://docs.rs/libc/0.2.101/libc/fn.syslog.html
* https://pubs.opengroup.org/onlinepubs/007908799/xsh/syslog.h.html
---

But, before progressing, I cant figure out how to solve below problem
```python
import syslog
syslog.openlog("test-syslog", syslog.LOG_CONS, syslog.LOG_USER)
syslog.syslog(syslog.LOG_INFO, "Hello syslog")
syslog.closelog()
```
```bash
>> cat /var/log/syslog
...
Sep  8 22:06:51 DESKTOP-8N***B3 test-syslog: Hello syslog // Must printed like this
Sep  8 22:06:51 DESKTOP-8N***B3 ���FV: FV: Hello syslog // But something like this corrupted one
```

I use same routine for querying string for `openlog` & `syslog` but It's behaviour is seriously different.
Is there any idea please?

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -2048,6 +2048,7 @@ mod posix {
     use super::*;
 
     use crate::builtins::list::PyListRef;
+    use crate::utils::ToCString;
     use bitflags::bitflags;
     use nix::unistd::{self, Gid, Pid, Uid};
     #[allow(unused_imports)] // TODO: use will be unnecessary in edition 2021
--- a/vm/src/stdlib/syslog.rs
+++ b/vm/src/stdlib/syslog.rs
@@ -1,0 +1,141 @@
+pub(crate) use syslog::make_module;
+
+#[pymodule(name = "syslog")]
+mod syslog {
+    use crate::builtins::pystr::{PyStr, PyStrRef};
+    use crate::common::lock::PyRwLock;
+    use crate::function::{OptionalArg, OptionalOption};
+    use crate::pyobject::PyValue;
+    use crate::utils::ToCString;
+    use crate::vm::VirtualMachine;
+    use crate::{PyObjectRef, PyResult, TryFromObject};
+    use std::{ffi::CStr, os::raw::c_char};
+
+    #[pyattr]
+    use libc::{
+        LOG_ALERT, LOG_AUTH, LOG_CONS, LOG_CRIT, LOG_DAEMON, LOG_DEBUG, LOG_EMERG, LOG_ERR,
+        LOG_INFO, LOG_KERN, LOG_LOCAL0, LOG_LOCAL1, LOG_LOCAL2, LOG_LOCAL3, LOG_LOCAL4, LOG_LOCAL5,
+        LOG_LOCAL6, LOG_LOCAL7, LOG_LPR, LOG_MAIL, LOG_NDELAY, LOG_NEWS, LOG_NOTICE, LOG_NOWAIT,
+        LOG_ODELAY, LOG_PID, LOG_SYSLOG, LOG_USER, LOG_UUCP, LOG_WARNING,
+    };
+
+    #[cfg(not(target_os = "redox"))]
+    #[pyattr]
+    use libc::{LOG_AUTHPRIV, LOG_CRON, LOG_PERROR};
+
+    fn get_argv(vm: &VirtualMachine) -> Option<PyStrRef> {
+        if let Some(argv) = vm.state.settings.argv.first() {
+            if !argv.is_empty() {
+                return Some(
+                    PyStr::from(match argv.find('\\') {
+                        Some(value) => &argv[value..],
+                        None => argv,
+                    })
+                    .into_ref(vm),
+                );
+            }
+        }
+        None
+    }
+
+    #[derive(Debug)]
+    enum GlobalIdent {
+        Explicit(Box<CStr>),
+        Implicit,
+    }
+
+    impl GlobalIdent {
+        fn as_ptr(&self) -> *const c_char {
+            match self {
+                GlobalIdent::Explicit(ref cstr) => cstr.as_ptr(),
+                GlobalIdent::Implicit => std::ptr::null(),
+            }
+        }
+    }
+
+    fn global_ident() -> &'static PyRwLock<Option<GlobalIdent>> {
+        rustpython_common::static_cell! {
+            static IDENT: PyRwLock<Option<GlobalIdent>>;
+        };
+        IDENT.get_or_init(|| PyRwLock::new(None))
+    }
+
+    #[derive(Default, FromArgs)]
+    struct OpenLogArgs {
+        #[pyarg(any, optional)]
+        ident: OptionalOption<PyStrRef>,
+        #[pyarg(any, optional)]
+        logoption: OptionalArg<i32>,
+        #[pyarg(any, optional)]
+        facility: OptionalArg<i32>,
+    }
+
+    #[pyfunction]
+    fn openlog(args: OpenLogArgs, vm: &VirtualMachine) -> PyResult<()> {
+        let logoption = args.logoption.unwrap_or(0);
+        let facility = args.facility.unwrap_or(LOG_USER);
+        let ident = match args.ident.flatten() {
+            Some(args) => Some(args.to_cstring(vm)?),
+            None => get_argv(vm).map(|argv| argv.to_cstring(vm)).transpose()?,
+        }
+        .map(|ident| ident.into_boxed_c_str());
+
+        let ident = match ident {
+            Some(ident) => GlobalIdent::Explicit(ident),
+            None => GlobalIdent::Implicit,
+        };
+
+        unsafe { libc::openlog(ident.as_ptr(), logoption, facility) };
+        *global_ident().write() = Some(ident);
+        Ok(())
+    }
+
+    #[derive(FromArgs)]
+    struct SysLogArgs {
+        #[pyarg(positional)]
+        priority: PyObjectRef,
+        #[pyarg(positional, optional)]
+        message_object: OptionalOption<PyStrRef>,
+    }
+
+    #[pyfunction]
+    fn syslog(args: SysLogArgs, vm: &VirtualMachine) -> PyResult<()> {
+        let (priority, msg) = match args.message_object.flatten() {
+            Some(s) => (i32::try_from_object(vm, args.priority)?, s),
+            None => (LOG_INFO, PyStrRef::try_from_object(vm, args.priority)?),
+        };
+
+        if global_ident().read().is_none() {
+            openlog(OpenLogArgs::default(), vm)?;
+        }
+
+        let (cformat, cmsg) = ("%s".to_cstring(vm)?, msg.to_cstring(vm)?);
+        unsafe { libc::syslog(priority, cformat.as_ptr(), cmsg.as_ptr()) };
+        Ok(())
+    }
+
+    #[pyfunction]
+    fn closelog() {
+        if global_ident().read().is_some() {
+            unsafe { libc::closelog() };
+            *global_ident().write() = None;
+        }
+    }
+
+    #[pyfunction]
+    fn setlogmask(maskpri: i32) -> i32 {
+        unsafe { libc::setlogmask(maskpri) }
+    }
+
+    #[inline]
+    #[pyfunction(name = "LOG_MASK")]
+    fn log_mask(pri: i32) -> i32 {
+        pri << 1
+    }
+
+    #[inline]
+    #[pyfunction(name = "LOG_UPTO")]
+    fn log_upto(pri: i32) -> i32 {
+        (1 << (pri + 1)) - 1
+    }
+}
--- a/vm/src/stdlib/syslog.rs
+++ b/vm/src/stdlib/syslog.rs
@@ -85,8 +85,11 @@ mod syslog {
             None => GlobalIdent::Implicit,
         };
 
-        unsafe { libc::openlog(ident.as_ptr(), logoption, facility) };
-        *global_ident().write() = Some(ident);
+        {
+            let mut locked_ident = global_ident().write();
+            unsafe { libc::openlog(ident.as_ptr(), logoption, facility) };
+            *locked_ident = Some(ident);
+        }
         Ok(())
     }
 
@@ -117,8 +120,9 @@ mod syslog {
     #[pyfunction]
     fn closelog() {
         if global_ident().read().is_some() {
+            let mut locked_ident = global_ident().write();
             unsafe { libc::closelog() };
-            *global_ident().write() = None;
+            *locked_ident = None;
         }
     }
 
--- a/vm/src/utils.rs
+++ b/vm/src/utils.rs
@@ -1,4 +1,5 @@
-use crate::builtins::PyFloat;
+use crate::builtins::{pystr::PyStr, PyFloat};
+use crate::exceptions::IntoPyException;
 use crate::{
     IntoPyObject, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject, TypeProtocol,
     VirtualMachine,
@@ -107,5 +108,21 @@ impl TryFromObject for std::time::Duration {
 impl IntoPyObject for std::convert::Infallible {
     fn into_pyobject(self, _vm: &VirtualMachine) -> PyObjectRef {
         match self {}
+    }
+}
+
+pub trait ToCString {
+    fn to_cstring(&self, vm: &VirtualMachine) -> PyResult<std::ffi::CString>;
+}
+
+impl ToCString for &str {
+    fn to_cstring(&self, vm: &VirtualMachine) -> PyResult<std::ffi::CString> {
+        std::ffi::CString::new(*self).map_err(|err| err.into_pyexception(vm))
+    }
+}
+
+impl ToCString for PyStr {
+    fn to_cstring(&self, vm: &VirtualMachine) -> PyResult<std::ffi::CString> {
+        std::ffi::CString::new(self.as_ref()).map_err(|err| err.into_pyexception(vm))
     }
 }
--- a/Lib/test/test_syslog.py
+++ b/Lib/test/test_syslog.py
@@ -1,0 +1,40 @@
+
+from test import support
+syslog = support.import_module("syslog") #skip if not supported
+import unittest
+
+# XXX(nnorwitz): This test sucks.  I don't know of a platform independent way
+# to verify that the messages were really logged.
+# The only purpose of this test is to verify the code doesn't crash or leak.
+
+class Test(unittest.TestCase):
+
+    def test_openlog(self):
+        syslog.openlog('python')
+        # Issue #6697.
+        self.assertRaises(UnicodeEncodeError, syslog.openlog, '\uD800')
+
+    def test_syslog(self):
+        syslog.openlog('python')
+        syslog.syslog('test message from python test_syslog')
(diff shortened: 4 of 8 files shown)


vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,181 @@
...
+    #[pyattr]
+    const LOG_PID: i32 = libc::LOG_PID;
+    #[pyattr]
+    const LOG_SYSLOG: i32 = libc::LOG_SYSLOG;
+    #[pyattr]
+    const LOG_USER: i32 = libc::LOG_USER;
+    #[pyattr]
+    const LOG_UUCP: i32 = libc::LOG_UUCP;
+    #[pyattr]
+    const LOG_WARNING: i32 = libc::LOG_WARNING;
```
youknowone, line 80: if type casting is not required, these definitions can be simpler.

```rust
#[pyattr]
use libc::{
    LOG_ALERT, LOG_AUTH, LOG_CONS, ...
};
```

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,181 @@
...
+
+    fn get_argv(vm: &VirtualMachine) -> PyResult {
+        let argv = sysmodule::argv(vm);
+        let scriptobj = PyStrRef::try_from_object(vm, argv.get_item(0.into_pyobject(vm), vm)?)?;
```
youknowone, line 84: scriptobj -> script_obj
scriptlen -> script_len

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,181 @@
...
+        let scriptobj = scriptobj.as_ref();
+
+        let scriptlen = scriptobj.len();
+        if scriptlen == 0 {
+            return Err(vm.new_value_error("empty script was given".to_owned()));
+        }
+
+        let retstr = match scriptobj.find('\\') {
+            Some(value) => scriptobj[value..scriptlen].to_string(),
```
youknowone, line 93:
```suggestion
            Some(value) => scriptobj[value..].to_string(),
```
when slicing to end, end position is not required

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,181 @@
...
+    const LOG_SYSLOG: i32 = libc::LOG_SYSLOG;
+    #[pyattr]
+    const LOG_USER: i32 = libc::LOG_USER;
+    #[pyattr]
+    const LOG_UUCP: i32 = libc::LOG_UUCP;
+    #[pyattr]
+    const LOG_WARNING: i32 = libc::LOG_WARNING;
+
+    fn get_argv(vm: &VirtualMachine) -> PyResult {
+        let argv = sysmodule::argv(vm);
```
youknowone, line 83: The code looks like expecting string rather than `PyStr`. Then maybe exposing it through vm method could be easier. The type of `vm.settings.args` is `Vec<String>`

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,181 @@
...
+            Some(value) => scriptobj[value..scriptlen].to_string(),
+            None => scriptobj.to_string(),
+        };
+
+        Ok(retstr.into_pyobject(vm))
+    }
+
+    fn convert_cstring(s: &str, vm: &VirtualMachine) -> PyResult<std::ffi::CString> {
+        std::ffi::CString::new(s).map_err(|err| err.into_pyexception(vm))
+    }
```
youknowone, line 102: we already have `pystr::to_cstring` for `PyStr`, but nothing for `&str` itself. Creating a common utility for both usage in `vm/src/utils.rs` and sharing it would be great!

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,181 @@
...
+    fn openlog(args: OpenLogArgs, vm: &VirtualMachine) -> PyResult<()> {
+        let logoption = args.logoption.unwrap_or(0);
+        let facility = args.facility.unwrap_or(LOG_USER);
+        let ident = match args.ident.flatten() {
+            Some(s) => s,
+            None => {
+                let argv = get_argv(vm)?;
+                PyStrRef::try_from_object(vm, argv)?
+            }
+        };
```
youknowone, line 124:
```suggestion
        let ident = args.ident.flatten().unwrap_or_else(|| {
                let argv = get_argv(vm)?;
                PyStrRef::try_from_object(vm, argv)?
        });
```

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    #[pyfunction]
+    fn syslog(args: SysLogArgs, vm: &VirtualMachine) -> PyResult<()> {
+        let (priority, msg) = match args.message_object.flatten() {
+            Some(s) => (
+                PyIntRef::try_from_object(vm, args.priority)?
+                    .as_bigint()
+                    .to_i32()
+                    .unwrap(),
```
youknowone, line 146: This is a user input, so I think user can input anything out of range value here.

```suggestion
                i32::try_from_object(vm, args.priority)?,
```

vm/src/stdlib/syslog.rs
Snowapril: I didn't know that. It is definitely better. Thanks. I applied it 👍

vm/src/stdlib/syslog.rs
Snowapril: @youknowone Thanks! I want to implement their shared trait as below code but something go wrong..
```rust
pub trait IntoCString<T: 'static + ?Sized> : AsRef<T> {
    fn to_cstring(&self, vm: &VirtualMachine) -> PyResult<std::ffi::CString> {
        std::ffi::CString::new(self.as_ref()).map_err(|err| err.into_pyexception(vm))
    }
}

impl IntoCString<str> for &str {}
impl IntoCString<str> for PyStr {}
```
Is there any idea for making this codes work?

vm/src/utils.rs
```diff
@@ -109,3 +110,19 @@ impl IntoPyObject for std::convert::Infallible {
...
+
+impl IntoCString for &str {
+    fn to_cstring(&self, vm: &VirtualMachine) -> PyResult<std::ffi::CString> {
+        std::ffi::CString::new(*self).map_err(|err| err.into_pyexception(vm))
+    }
+}
+
+impl IntoCString for PyStr {
+    fn to_cstring(&self, vm: &VirtualMachine) -> PyResult<std::ffi::CString> {
+        std::ffi::CString::new(self.as_ref()).map_err(|err| err.into_pyexception(vm))
```
youknowone, line 126: The naming convention for error-returning traits are `TryFrom` and `TryInto` in Rust. In Rust, `to` prefix and `into` prefix have different meaning, so the method name must use `into` prefix. (create vs move)

I think `self.as_ref().try_into_cstring(vm)` will work here.

I originally thought just adding a free function will be enough, but this version also look great.

Snowapril: Ok Thanks! I applied it 😊

By the way, I am considering the way for removing duplicated code in both PyStr and &str.
```rust
std::ffi::CString::new(self.as_ref()).map_err(|err| err.into_pyexception(vm)) // PyStr impl
std::ffi::CString::new(*self).map_err(|err| err.into_pyexception(vm)) // &str impl
```

I found that PyStr and str both implementing [`AsRef`](https://doc.rust-lang.org/std/convert/trait.AsRef.html) trait and I want to utilize it something like below.
```rust
pub trait ToCString<T: 'static + ?Sized> : AsRef<T> {
    fn to_cstring(&self, vm: &VirtualMachine) -> PyResult<std::ffi::CString> {
        std::ffi::CString::new(self.as_ref()).map_err(|err| err.into_pyexception(vm))
    }
}

impl ToCString<str> for str {}
impl ToCString<str> for PyStr {}
```

But I cant resolve compile error.. Could you give me any hint for solving this? Thanks!

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,102 @@
...
+        LOG_INFO, LOG_KERN, LOG_LOCAL0, LOG_LOCAL1, LOG_LOCAL2, LOG_LOCAL3, LOG_LOCAL4, LOG_LOCAL5,
+        LOG_LOCAL6, LOG_LOCAL7, LOG_LPR, LOG_MAIL, LOG_NDELAY, LOG_NEWS, LOG_NOTICE, LOG_NOWAIT,
+        LOG_ODELAY, LOG_PID, LOG_SYSLOG, LOG_USER, LOG_UUCP, LOG_WARNING,
+    };
+
+    #[cfg(not(target_os = "redox"))]
+    #[pyattr]
+    use libc::{LOG_AUTHPRIV, LOG_CRON, LOG_PERROR};
+
+    fn get_argv(vm: &VirtualMachine) -> PyResult<PyStrRef> {
```
DimitrisJim, line 24: This should ignore any errors and return an `Option`. If openlog is passed a `NULL` pointer for `ident` it uses `argv[0]` (i.e `"rustpython"`, see [here](https://www.gnu.org/software/libc/manual/html_node/openlog.html)).

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    #[pyfunction]
+    fn syslog(args: SysLogArgs, vm: &VirtualMachine) -> PyResult<()> {
+        let (priority, msg) = match args.message_object.flatten() {
+            Some(s) => (i32::try_from_object(vm, args.priority)?, s),
+            None => (LOG_INFO, PyStrRef::try_from_object(vm, args.priority)?),
+        };
+
+        let (cformat, cmsg) = ("%s".to_cstring(vm)?, msg.to_cstring(vm)?);
+        unsafe { libc::syslog(priority, cformat.as_ptr(), cmsg.as_ptr()) };
```
DimitrisJim, line 77: I haven't been able to find a clear reason why the problem you have exists but it may be because you just call `syslog` here. In `CPython`, `syslog` calls `openlog` if it hasn't already done so (see `S_log_open`) and that might have something to do with it (I *think* `syslog` will try and call `openlog` again in some cases but I'm not feeling like looking into the implementation of `syslog` in `libc`).

Try following along with what they do and see if this solves it. 

Note: this could be done in a following PR so as to not stall the whole module being included. The other comment I left should be addressed in this PR, though (and it should be easier to address).

DimitrisJim (dismissed):

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    #[pyfunction]
+    fn openlog(args: OpenLogArgs, vm: &VirtualMachine) -> PyResult<()> {
+        let logoption = args.logoption.unwrap_or(0);
+        let facility = args.facility.unwrap_or(LOG_USER);
+        let ident = args
+            .ident
+            .flatten()
+            .unwrap_or(get_argv(vm)?)
```
youknowone, line 54:
```suggestion
            .unwrap_or_else(|| get_argv(vm)?)
```
`unwrap_or_else` prevent unnessessary running cost of `get_argv`

vm/src/stdlib/syslog.rs
Snowapril: Oh you are right. I was thinking that the way how passing nullptr to `libc::openlog` is to use `PyResult` 😂
I'll switch it with `Option` and pass `std::ptr::null` to `libc::openlog` if it returns `None` later. Thanks!

vm/src/stdlib/syslog.rs
Snowapril: Thanks! I'm continuously learning advanced rust features by your reviews. How grateful!
I'll fix it later. Thanks.

vm/src/stdlib/syslog.rs
Snowapril: There also exists problem when I called `syslog` after make connection to syslogger by calling `openlog`. As global variable `S_log_open` in cpython is only for make guarantee there exists connection to syslogger, I think it will not resolve this issue 😢
By the way, a mechanism something like global variable in cpython should be implemented.

Snowapril:
sorry for late.. I'm quite hard to work these days. 
I opened this issue, but there still exists work to do, such as corrupted text, syslogger connection check with global variable.
But I can't solve them in clearways.

vm/src/stdlib/syslog.rs
youknowone: one of OnceCell, lazy_static or LocalKey will be helpful for static variable

vm/src/stdlib/os.rs
```diff
@@ -8,6 +8,7 @@ use std::time::{Duration, SystemTime};
 use std::{env, fs};
 
 use crate::crt_fd::Fd;
+use crate::utils::ToCString;
```
youknowone, line 11: This unused warning is not resolved

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,106 @@
...
+    #[pyattr]
+    use libc::{LOG_AUTHPRIV, LOG_CRON, LOG_PERROR};
+
+    fn get_argv(vm: &VirtualMachine) -> Option<PyStrRef> {
+        if let Some(argv) = vm.state.settings.argv.first() {
+            if !argv.is_empty() {
+                return Some(
+                    PyStr::from(match argv.find('\\') {
+                        Some(value) => argv[value..].to_owned(),
+                        None => argv.to_owned(),
```
youknowone, line 30:
```suggestion
                        Some(value) => argv[value..].to_owned(),
                        None => argv.to_owned(),
```

PyStr::from works for `&str` too

youknowone (changes requested):

vm/src/stdlib/syslog.rs
DimitrisJim: @Snowapril  really not sure. Maybe it has something to do with how `syslog` is implemented, a double call to `openlog` might mess things up. Either way, I think this is a relatively minor bug for now so it can be addressed at some later point (I would like a `TODO:` comment though so we don't forget about it).

vm/src/stdlib/os.rs
Snowapril: Thanks! I did not check clippy in windows. I resolved it 😊

Snowapril:
I found the cause of the corrupted ident header message in [gnu document](https://www.gnu.org/software/libc/manual/html_node/openlog.html).

As document said, We must keep the `ident` string from deallocation while using `syslog` module.
> Please note that the string pointer ident will be retained internally by the Syslog routines.
> You must not free the memory that ident points to.

Declare two static `AtomicCell` for ident string and boolean which represent connection status.

If `closelog` is called and if there is connection between syslogger service, we must disconnect and deallocate `ident` string to prevent leak.

As Option and Box both do not provide Copy trait, I add one more static variable `SYSLOG_IS_OPEN` .

With this fix, now the problem I commented now work normal.
```python
import syslog
syslog.openlog("test-syslog", syslog.LOG_CONS, syslog.LOG_USER)
syslog.syslog(syslog.LOG_INFO, "Hello syslog")
syslog.closelog()
```
```bash
Sep 15 14:22:09 DESKTOP-******* test-syslog: Hello syslog
```

But [the codes](https://github.com/RustPython/RustPython/pull/3023/commits/f251fca1aec60c2231c562b999e7d8fe11c29838) for storing boxed `Cstr` to `AtomicCell` in `openlog` is little bit dirty.
I could not make these codes better. How can I do?

DimitrisJim:
Great find! I'll let @youknowone check the usage of for the global state here but it lgtm.

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    #[pyfunction]
+    fn syslog(args: SysLogArgs, vm: &VirtualMachine) -> PyResult<()> {
+        let (priority, msg) = match args.message_object.flatten() {
+            Some(s) => (i32::try_from_object(vm, args.priority)?, s),
+            None => (LOG_INFO, PyStrRef::try_from_object(vm, args.priority)?),
+        };
+
+        if !SYSLOG_IS_OPEN.load() {
```
youknowone, line 95: I am sorry. I messaged you it look like able to be managed with AtomicCell, but it doesn't. `closelog` can be called betwen line 95-96 and then SYSLOG_IDENT will be corrupted. Would you change it to RwLock?

youknowone: That's the point for lazy_static or OnceCell.

```rust
static SYSLOG_IDENT: OnceCell<PyRwLock<Option<Box<std::ffi::CStr>>>> = OnceCell::new();
```

https://docs.rs/once_cell/1.8.0/once_cell/#lazy-initialized-global-data

Creating a helper function will be convenient.

```rust
fn syslog_global_ident() -> &'static PyRwLock<Option<Box<CStr>>> {
    static SYSLOG_IDENT: OnceCell<PyRwLock<Option<Box<std::ffi::CStr>>>> = OnceCell::new();
    SYSLOG_INIT.get_ot_init(|| ...)
}
```

Then both open and close function can access to the shared global ident through the function like `syslog_global_ident().write()` or ``syslog_global_ident().read()`

Snowapril: Oh my gosh. this makes me shock 😂. I applied your suggestion.

Snowapril: uhm... There was no error in my environment. I'll solve CI error before sleeping 😢

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,132 @@
...
+                cstr_ptr
+            }
+            None => match get_argv(vm) {
+                Some(argv) => {
+                    let cstr = argv.to_cstring(vm)?.into_boxed_c_str();
+                    let cstr_ptr = cstr.as_ptr();
+                    *syslog_global_ident().write() = Some(cstr);
+                    cstr_ptr
+                }
+                None => std::ptr::null(),
```
youknowone, line 76: no update for `syslog_global_ident()` if this is None? what if openlog is called twice first with args but second without it?

vm/src/stdlib/syslog.rs
```diff
@@ -0,0 +1,132 @@
...
+            },
+        };
```
youknowone, line 78:
```suggestion
        let ident = match args.ident.flatten() {
            Some(args) => Some(args.to_cstring(vm)?),
            None => get_argv(vm).map(|argv| argv.to_cstring(vm)?),
        }.map(|ident| ident.into_boxed_c_str());
        *syslog_global_ident().write() = ident;
```

vm/src/stdlib/syslog.rs
DimitrisJim: Though [the docs](https://www.gnu.org/software/libc/manual/html_node/openlog.html) seem to _indicate_ this should be fine, [CPython sets it to `NULL`](https://github.com/python/cpython/blob/80d9ff16483b6c1898bcdcc811b5450b57a5e573/Modules/syslogmodule.c#L135) so I think we should too.

youknowone (approved):

youknowone:
@Snowapril I added a commit to sync opened ident and the variable

PR #3024, opened 2021-09-08
Correct `object.__repr__`

moreal:
This pull request fixes `object.__repr__` implementation.

## References

- [`object_repr` implementation](https://github.com/python/cpython/blob/f9242d50b18572ef0d584a1c815ed08d1a38e4f4/Objects/typeobject.c#L4518-L4544)

### DIFF
--- a/vm/src/builtins/object.rs
+++ b/vm/src/builtins/object.rs
@@ -1,7 +1,7 @@
 use super::dict::{PyDict, PyDictRef};
 use super::list::PyList;
 use super::pybool;
-use super::pystr::PyStrRef;
+use super::pystr::{PyStr, PyStrRef};
 use super::pytype::PyTypeRef;
 use crate::builtins::pytype::PyType;
 use crate::common::hash::PyHash;
@@ -183,12 +183,29 @@ impl PyBaseObject {
 
     /// Return repr(self).
     #[pymethod(magic)]
-    fn repr(zelf: PyObjectRef) -> String {
-        format!(
-            "<{} object at {:#x}>",
-            zelf.class().tp_name(),
-            zelf.get_id()
-        )
+    fn repr(zelf: PyObjectRef, vm: &VirtualMachine) -> Option<String> {
+        let class = zelf.class();
+
+        match (
+            class
+                .qualname(vm)
+                .downcast_ref::<PyStr>()
+                .map(|n| n.as_str()),
+            class.module(vm).downcast_ref::<PyStr>().map(|m| m.as_str()),
+        ) {
+            (None, _) => None,
+            (Some(qualname), Some(module)) if module != "builtins" => Some(format!(
+                "<{}.{} object at {:#x}>",
+                module,
+                qualname,
+                zelf.get_id()
+            )),
+            _ => Some(format!(
+                "<{} object at {:#x}>",
+                class.tp_name(),
+                zelf.get_id()
+            )),
+        }
     }
 
     #[pyclassmethod(magic)]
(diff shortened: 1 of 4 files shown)


vm/src/builtins/object.rs
```diff
@@ -4,6 +4,7 @@ use super::pybool;
 use super::pystr::PyStrRef;
 use super::pytype::PyTypeRef;
 use crate::builtins::pytype::PyType;
+use crate::builtins::PyStr;
```
youknowone, line 7: this `use` can be added to line 4

vm/src/builtins/object.rs
```diff
@@ -183,12 +183,24 @@ impl PyBaseObject {
...
+    fn repr(zelf: PyObjectRef, vm: &VirtualMachine) -> String {
+        let class = zelf.class();
+        match class.module(vm).downcast_ref::<PyStr>().map(|m| m.as_str()) {
+            Some(module) if module != "builtins" => {
+                let class_name = class.name();
+                format!(
+                    "<{}.{} object at {:#x}>",
+                    module,
+                    class
+                        .qualname(vm)
```
youknowone, line 195: by the document referenced in the previous PR (https://docs.python.org/3/c-api/typeobj.html?highlight=tp_name#c.PyTypeObject.tp_name)

it is `__module__` + `__name__` or just `tp_name`.  Then isn't this `.name()` not `qualname()`?

moreal: Yes, you're right. I believed myself at the past but I didn't 😅  I turned the return type into `Option<String>` and made it return `None` if the `qualname` is `None`.

moreal: Hmm, as I understood, you seem to say `tp_name` is `__module__` + `__name__` so `.name()` should be placed instead of `qualname` with the below section:

> For dynamically allocated type objects, this should just be the type name, and the module name explicitly stored in the type dict as the value for key`'__module__`.
> 
> For statically allocated type objects, the tp_name field should contain a dot. Everything before the last dot is made accessible as the `__module__` attribute, and everything after the last dot is made accessible as the `__name__` attribute.

At first, I implemented by following CPython implementation, the below code:

```c
static PyObject *
object_repr(PyObject *self)
{
    PyTypeObject *type;
    PyObject *mod, *name, *rtn;

    type = Py_TYPE(self);
    mod = type_module(type, NULL);
    if (mod == NULL)
        PyErr_Clear();
    else if (!PyUnicode_Check(mod)) {
        Py_DECREF(mod);
        mod = NULL;
    }
    name = type_qualname(type, NULL);
```

And the expression with `module` + `qualname` and another expression with `tp_name`, expect different result. The one with `tp_name` will be used in only `builtins` module.

The reason that we must use `qualname` and can't replace `qualname` with `name` for compatibility, is that it will make result different. The `module` doesn't treat function as module. Let's see with example. I prepared the example code:

```python
def f():
  class A: ...
  print(A.__module__)
  print(A.__name__)
```

There is a function `f` and the `f` function creates dynamic type `A` for each call. The `A.__module__` will have `__main__` and `A.__name__` will have only `A`. And `A.__qualname__` will be `f.<locals>.A`.

Review with Python documentation:

> For dynamically allocated type objects, this should just be the type name,

The `A.__name__` will be just type name, `A`.

> and the module name explicitly stored in the type dict as the value for key '__module__'.

The `A.__module__` will be just `__main__`.

----

P.S. though I already wrote long lines, if I didn't understand your question well, please let me know 😅

youknowone: I am really sorry to make these confusions. I'd better to see CPython code earlier. Thanks a lot.

youknowone (approved):

PR #3025, opened 2021-09-08
replace a few zeroed to MaybeUninit

youknowone:
this replacements don't have much benefits but as an example to use MaybeUninit.

### DIFF
--- a/vm/src/stdlib/winapi.rs
+++ b/vm/src/stdlib/winapi.rs
@@ -189,9 +189,9 @@ fn _winapi_CreateProcess(
         .as_mut()
         .map_or_else(null_mut, |w| w.as_mut_ptr());
 
-    let mut procinfo = unsafe { std::mem::zeroed() };
-    let ret = unsafe {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/winapi.rs
```diff
@@ -189,28 +189,30 @@ fn _winapi_CreateProcess(
...
-                | winbase::EXTENDED_STARTUPINFO_PRESENT
-                | winbase::CREATE_UNICODE_ENVIRONMENT,
-            env as _,
-            current_dir,
-            &mut si as *mut winbase::STARTUPINFOEXW as _,
-            &mut procinfo,
-        )
+    let procinfo = unsafe {
+        let mut procinfo = std::mem::MaybeUninit::uninit();
+        let ret = unsafe {
```
DimitrisJim, line 194: Ping, warning came up which you probably didn't see since you aren't on windows.

DimitrisJim (approved):
lgtm, other than the warning for the superfluous unsafe.

PR #3027, opened 2021-09-09
os: scandir add argument support

deantvv:
Add support for class `bytearray` and `memoryview`

Noted that clippy is not happy with the `Vec::from` part (warning: redundant closure). 
However when I change the code as clippy suggested, it fail to build with [error: implementation of `FnOnce` is not general enough].
It could be an issue in clippy implementation.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -14,6 +14,7 @@ use num_bigint::BigInt;
 use strum_macros::EnumString;
 
 use super::errno::errors;
+use crate::buffer::PyBufferRef;
 use crate::builtins::bytes::{PyBytes, PyBytesRef};
 use crate::builtins::dict::PyDictRef;
 use crate::builtins::int;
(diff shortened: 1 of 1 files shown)


Lib/test/test_os.py
```diff
@@ -3930,15 +3930,17 @@ def test_bytes(self):
...
         self.create_file("file.txt")
 
         for cls in bytearray, memoryview:
             path_bytes = cls(os.fsencode(self.path))
-            with self.assertWarns(DeprecationWarning):
-                entries = list(os.scandir(path_bytes))
+
+            # TODO: RUSTPYTHON (should emit DeprecationWarning)
+            # with self.assertWarns(DeprecationWarning):
+            #     entries = list(os.scandir(path_bytes))
```
youknowone, line 3941: please do not comment out tests in unittest but just keep the original `expectedFailure` until everything is fixed. They will be tracked automatically but these commented out codes are not.

vm/src/stdlib/os.rs
```diff
@@ -210,6 +210,16 @@ pub(crate) fn fspath(
...
 impl TryFromObject for PyPathLike {
     fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
+        let obj = match_class!(match obj {
+            a @ PyByteArray => {
+                a.borrow_buf_mut().clone().into_pyobject(vm)
+            }
+            m @ PyMemoryView => {
+                let b: PyBytes = m.try_bytes(vm, |x| Vec::from(x))?.into();
+                b.into_pyobject(vm)
+            }
```
youknowone, line 220: Rather than checking PyMemoryView directly, CPython is checking if it is a buffer or not. Try to getting buffer from PyObjectRef is like this:

```rust
let buffer = PyBufferRef::TryFromObject(obj)
```

Lib/test/test_os.py
deantvv: Good point. Will change PR in suggested way. Thanks!

vm/src/stdlib/os.rs
deantvv: Changed as suggested!

youknowone (approved):

PR #3030, opened 2021-09-10
posix : implement sched_param class

zetwhite:
This PR contains implementation of `os.sched_param` class.

You can check the interface of `sched_param` here :  
https://docs.python.org/3/library/os.html#os.sched_param 
 

With this code, RustPython works as below : 
```python 
>>>>> import os
>>>>> os.sched_param(1)
posix.sched_param(sched_priority = 1)
>>>>> os.sched_param('hi')
posix.sched_param(sched_priority = 'hi')
``` 
```python 
>>>>> import os 
>>>>> a = os.sched_param(1)
>>>>> a.sched_priority
1
>>>>> b = os.sched_param('hello')
>>>>> b.sched_priority
'hello'
```

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -2048,6 +2048,7 @@ mod posix {
     use super::*;
 
     use crate::builtins::list::PyListRef;
+    use crate::slots::SlotConstructor;
     use crate::utils::ToCString;
     use bitflags::bitflags;
     use nix::unistd::{self, Gid, Pid, Uid};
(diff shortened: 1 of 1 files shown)


zetwhite:
I am also working on `os.sched_setscheduler` and `os.sched_getscheduler`. 
But before that, I'm not sure that this class is fine.
So, I wanted to check it first by committing this first.
Please let me know if there is any problem.

vm/src/stdlib/os.rs
```diff
@@ -2436,6 +2436,53 @@ mod posix {
...
+                sched_priority: obj,
+            })
+        }
+    }
+
+    #[pyimpl]
+    impl SchedParam {
+        #[pyslot]
+        fn tp_new(cls: PyTypeRef, args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+            let obj: PyObjectRef = args.bind(vm)?;
```
youknowone, line 2464: Recently `SlotConstructor` is added. I recommend to use it for new code rather than `bind`

youknowone: Because you already added `TryFromObject for SchedParam`,
```rust
        fn tp_new(cls: PyTypeRef, args: FuncArgs, vm: &VirtualMachine) -> PyResult {
            let obj: SchedParam = args.bind(vm)?;
```
will work.

vm/src/stdlib/os.rs
```diff
@@ -2436,6 +2437,52 @@ mod posix {
...
+            let sched_priority_repr = vm.to_repr(&self.sched_priority)?;
+            Ok(format!(
+                "posix.sched_param(sched_priority = {})",
+                sched_priority_repr.as_str()
+            ))
+        }
+    }
+
+    impl SlotConstructor for SchedParam {
+        type Args = PyObjectRef;
```
youknowone, line 2479: Because `SchedParam` implements `TryFromObject`,
I think this conversion also works
```suggestion
        type Args = SchedParam;
```

vm/src/stdlib/os.rs
```diff
@@ -26,6 +26,7 @@ use crate::common::lock::PyRwLock;
 use crate::exceptions::{IntoPyException, PyBaseExceptionRef};
 use crate::function::{ArgumentError, FromArgs, FuncArgs, OptionalArg};
 use crate::slots::PyIter;
+use crate::slots::SlotConstructor;
```
youknowone, line 29: this line can be merged with line 28

youknowone (changes requested):

youknowone (approved):

vm/src/stdlib/os.rs
```diff
@@ -26,7 +26,7 @@ use crate::byteslike::ArgBytesLike;
 use crate::common::lock::PyRwLock;
 use crate::exceptions::{IntoPyException, PyBaseExceptionRef};
 use crate::function::{ArgumentError, FromArgs, FuncArgs, OptionalArg};
-use crate::slots::PyIter;
+use crate::slots::{PyIter, SlotConstructor};
```
youknowone, line 29: `use SlotConstructor` seems to be moved to `posix` module.

youknowone (changes requested):

youknowone (approved):

PR #3031, opened 2021-09-10
Add cmath module.

DimitrisJim:
The minimum structure needed for the tests to be able to execute. The functions that remain can be good first issues for anyone interested.

I've purposely not used macros similar to `math.rs` since we can't capture docs for them (though that could be an easy fix) and we don't generate the signature for them in they aren't marked as `pyfunction`s.

### DIFF
--- a/vm/src/builtins/complex.rs
+++ b/vm/src/builtins/complex.rs
@@ -7,11 +7,11 @@ use super::pystr::PyStr;
 use super::pytype::PyTypeRef;
 use crate::function::{OptionalArg, OptionalOption};
 use crate::slots::{Comparable, Hashable, PyComparisonOp, SlotConstructor};
-use crate::VirtualMachine;
 use crate::{
     IdProtocol, IntoPyObject,
     PyArithmaticValue::{self, *},
-    PyClassImpl, PyComparisonValue, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
+    PyClassImpl, PyComparisonValue, PyContext, PyObjectRef, PyRef, PyResult, PyValue,
+    TryFromObject, TypeProtocol, VirtualMachine,
 };
 use rustpython_common::{float_ops, hash};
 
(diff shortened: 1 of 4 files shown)


vm/src/builtins/complex.rs
```diff
@@ -451,3 +451,25 @@ fn try_complex(obj: &PyObjectRef, vm: &VirtualMachine) -> PyResult<Option<(Compl
...
+    pub fn to_complex(self) -> Complex64 {
+        self.value
+    }
+}
+
+impl TryFromObject for IntoPyComplex {
+    fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
+        // We do not care if it was already a complex.
+        let (value, _) = try_complex(&obj, vm)?.ok_or_else(|| {
+            vm.new_type_error(format!("must be real number, not {}", obj.class().name()))
```
youknowone, line 471: real -> complex?

vm/src/stdlib/cmath.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// -inf, inf and NaN behave similarly to the IEEE 754 Standard. That is, NaN is
+    /// not close to anything, even itself. inf and -inf are only close to themselves.
+    #[pyfunction]
+    fn isclose(args: IsCloseArgs, vm: &VirtualMachine) -> PyResult<bool> {
+        let a = args.a.to_complex();
+        let b = args.b.to_complex();
+        let rel_tol = match args.rel_tol {
+            OptionalArg::Missing => 1e-09,
+            OptionalArg::Present(ref value) => value.to_f64(),
+        };
```
youknowone, line 111: `OptionalArg` supports `Option` methods.
```suggestion
        let rel_tol = args.rel_tol.map_or(1e-09, |value| value.to_f64());
```

youknowone (changes requested):

vm/src/builtins/complex.rs
DimitrisJim: Technically I believe you're right since `math` is the one that requires reals (the message for cmath should probably be 'complex or real'`. I used 'real' since that's what they currently use in CPython, though, for consistency reasons.

youknowone (approved):

PR #3033, opened 2021-09-10
Apply path_converter structure

youknowone:
Apply CPython code structure for #3027

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -168,6 +168,8 @@ pub(crate) fn fspath(
     check_for_nul: bool,
     vm: &VirtualMachine,
 ) -> PyResult<FsPath> {
+    // path_converter
+    // unicode -> bytes -> buffer -> index -> fspath(unicode/bytes) -> TypeError
     let check_nul = |b: &[u8]| {
         if !check_for_nul || memchr::memchr(b'\0', b).is_none() {
             Ok(())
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/os.rs
```diff
@@ -168,6 +168,8 @@ pub(crate) fn fspath(
     check_for_nul: bool,
     vm: &VirtualMachine,
 ) -> PyResult<FsPath> {
+    // path_converter
+    // unicode -> bytes -> buffer -> index -> fspath(unicode/bytes) -> TypeError
```
DimitrisJim, line 172: I'm not versed with `os` stuff but I see this as a pretty common utility function used in `CPython` for many of the functions in `os` (in the argument clinic generated files). So wouldn't it make more sense in `TryFromObject`?

vm/src/stdlib/os.rs
```diff
@@ -191,7 +193,15 @@ pub(crate) fn fspath(
     };
     let obj = match match1(obj)? {
         Ok(pathlike) => return Ok(pathlike),
-        Err(obj) => obj,
+        Err(obj) => match PyBufferRef::try_from_borrowed_object(vm, &obj) {
```
deantvv, line 196: The reason why I try to implement this part of code inside trait `TryFromObject` in `PyPathLike` is because in CPython `os.fspath` doesn't support class `bytearray` or `memoryview`.

Maybe we could add a flag like `allow_buffer_like` in the function and reject buffer like object in `os.fspath` to match CPython behavior?

youknowone: I didn't know it very well but fspath and PathLike seem different. Thank you!

vm/src/stdlib/os.rs
youknowone: I agree

PR #3037, opened 2021-09-11
Fix failed test remainder in test_math

owljoa:
add conditions for below rules (specified in test codes)

1. remainder(x, inf) is x, for non-NaN non-infinite x
2. According to IEEE 754-2008 7.2(f)
 2-1. remainder(x, 0) for non-NaN x is invalid operation
 2-2. remainder(infinity, x) for non-NaN x is invalid operation

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -678,7 +678,7 @@ fn math_fmod(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResult<f6
 fn math_remainder(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResult<f64> {
     let x = x.to_f64();
     let y = y.to_f64();
-    
+
     if x.is_finite() && y.is_finite() {
         if y == 0.0 {
             return Err(vm.new_value_error("math domain error".to_owned()));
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/math.rs
```diff
@@ -696,6 +697,15 @@ fn math_remainder(x: IntoPyFloat, y: IntoPyFloat, vm: &VirtualMachine) -> PyResu
 
         return Ok(1.0_f64.copysign(x) * r);
     }
+    if x.is_finite() && y.is_infinite() {
```
DimitrisJim, line 700: There's currently a lot of duplicate checks going on here and in the following lines until the end of the function. Condensing it down to:

```rust
    if x.is_infinite() && !y.is_nan() {
        return Err(vm.new_value_error("math domain error".to_owned()));
    }
    if x.is_nan() || y.is_nan() {
        return Ok(std::f64::NAN);
    }
    if y.is_infinite() {
        Ok(x)
    } else {
        Err(vm.new_value_error("math domain error".to_owned()))
    }
```

seems to handle the cases fine (at least those tested).

DimitrisJim (changes requested):

vm/src/stdlib/math.rs
owljoa: Oh,, you're right. (I should have thought that..)
I fixed it thanks to your comment 👍

DimitrisJim (approved):
lgtm now. 

I'm under the impression a similar set of checks is done for other functions too (looking through the CPython code `FUNC_2` wraps many `(double, double) -> double` funcs).  Don't know what the other functions in `math.rs` do currently but I'm guessing there's lots of duplicated code checking these same edge cases.

Not requesting this now, it just seems like a good refactoring opportunity if you'd be interested.

youknowone:
I think we need a utility for "math domain error". it appears everywhere.

PR #3038, opened 2021-09-11
Use associated constants instead of module constants.

DimitrisJim:
Replace occurrences of module constants with associated constants since the latter will be deprecated at some point as per their docs. Only change not made is in `math_remainder` since #3037 will take care of that.


youknowone (approved):

youknowone:
I always wondered why this is not attached to the types, but never knew there was deprecation plan about it.

DimitrisJim:
Neither did I. I'd seen it before but had completely forgotten about it. Either way, thought it was best to just move to the associated constants just in case.

PR #3040, opened 2021-09-11
Clean up more instances of matching on Optional.

DimitrisJim:
Cleans up a couple more instances of matching on Optional by using `Option` methods.

### DIFF
--- a/vm/src/builtins/enumerate.rs
+++ b/vm/src/builtins/enumerate.rs
@@ -41,11 +41,9 @@ impl SlotConstructor for PyEnumerate {
     type Args = EnumerateArgs;
 
     fn py_new(cls: PyTypeRef, args: Self::Args, vm: &VirtualMachine) -> PyResult {
-        let counter = match args.start {
-            OptionalArg::Present(start) => start.as_bigint().clone(),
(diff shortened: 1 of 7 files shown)


vm/src/builtins/enumerate.rs
```diff
@@ -41,11 +41,9 @@ impl SlotConstructor for PyEnumerate {
...
 
     fn py_new(cls: PyTypeRef, args: Self::Args, vm: &VirtualMachine) -> PyResult {
-        let counter = match args.start {
-            OptionalArg::Present(start) => start.as_bigint().clone(),
-            OptionalArg::Missing => BigInt::zero(),
-        };
-
+        let counter = args
+            .start
+            .map_or(BigInt::zero(), |start| start.as_bigint().clone());
```
youknowone, line 46:
```suggestion
            .map_or_else(|| BigInt::zero(), |start| start.as_bigint().clone());
```
To make equivalent code, `map_or_else` fits better

youknowone (approved):

vm/src/builtins/enumerate.rs
DimitrisJim: ah yes, good catch!

PR #3042, opened 2021-09-12
Add uuencoding/decoding code to binascii.rs

fanninpm:
This code is the beginnings of proper uuencoding/decoding, but it can be further refined.

### DIFF
--- a/vm/src/stdlib/binascii.rs
+++ b/vm/src/stdlib/binascii.rs
@@ -204,6 +204,7 @@ mod decl {
 
             // Allocate the buffer
             let mut res = Vec::<u8>::with_capacity(length);
+            let trailing_garbage_error = || Err(vm.new_value_error("Trailing garbage".to_string()));
 
             for chunk in b.get(1..).unwrap_or_default().chunks(4) {
                 let char_a = chunk.get(0).map_or(Ok(0), |x| uu_a2b_read(x, vm))?;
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/binascii.rs
```diff
@@ -177,31 +177,113 @@ mod decl {
...
     }
 
+    #[inline]
+    fn uu_a2b_read(s: &[u8], index: usize, vm: &VirtualMachine) -> PyResult<u8> {
+        if let Some(c) = s.get(index) {
+            // Check the character for legality
+            // The 64 instead of the expected 63 is because
+            // there are a few uuencodes out there that use
+            // '`' as zero instead of space.
+            if *c < 32 || *c > 96 {
```
youknowone, line 187:
```suggestion
            if !(0x20..=0x60).contains(*c) {
```

Hexadecimal literals are used below but not here

vm/src/stdlib/binascii.rs
```diff
@@ -177,31 +177,113 @@ mod decl {
...
+                if *c == 10 || *c == 13 {
```
youknowone, line 188:
```suggestion
                if [b’\r’, b’\n’].contains(*c) {
```

vm/src/stdlib/binascii.rs
```diff
@@ -177,31 +177,113 @@ mod decl {
...
-                b
+                ((b[0] - 0x20) & 0x3f) as usize
             };
-            // TODO: RUSTPYTHON, implement actual uuencoding code
-            base64::decode(b)
+
+            // Allocate the buffer
+            let mut res = Vec::<u8>::with_capacity(length);
+
+            for i in (1..b.len()).step_by(4) {
```
youknowone, line 212: Rather than manually manipulating index and chunks, How about this way?
```suggestion
let (chunks, remainder) =  b.as_chunks::<4>();
for (i, chunk) in chunks.enumerate() {
```

youknowone (approved):

PR #3045, opened 2021-09-12
Clean up bare occurunces of `[pyarg(any)]`

DimitrisJim:
Since it implies them being handled as `positional_or_keyword`. Alternatively, an argument without a `pyarg` attribute should be disallowed.


youknowone:
@coolreader18 could you check if the design decision was making `pyarg(any)` mandatory or allowing empty attribute?

coolreader18:
Honestly I'm fine with leaving out `#[pyarg(...)]`, in python it's pretty implicit that by default arguments are positional_or_keyword, so it makes sense for it to be like that in Rust too.

youknowone (approved):

PR #3046, opened 2021-09-13
Implement latin_1 in Rust

fanninpm:
~~Based on the PyPy implementation. The encoding function needs some work with respect to error handling.~~ **EDIT**: Now based on @coolreader18's `ascii` codec.

Requesting @coolreader18 as a potential reviewer, as they wrote the architecture for the `encodings` module and the error handling function(s).

### DIFF
--- a/common/src/encodings.rs
+++ b/common/src/encodings.rs
@@ -189,39 +189,50 @@ pub mod latin_1 {
             match data
                 .char_indices()
                 .enumerate()
-                .find(|(_, (_, c))| (*c as u32) > 255)
+                .find(|(_, (_, c))| !c.is_ascii())
             {
                 None => {
                     out.extend_from_slice(data.as_bytes());
                     break;
                 }
-                Some((char_i, (byte_i, _))) => {
+                Some((char_i, (byte_i, ch))) => {
                     out.extend_from_slice(&data.as_bytes()[..byte_i]);
                     let char_start = char_data_index + char_i;
-                    // number of non-latin_1 chars between the first non-latin_1 char and the next latin_1 char
-                    let non_latin_1_run_length = data[byte_i..]
-                        .chars()
-                        .take_while(|c| (*c as u32) > 255)
-                        .count();
-                    let char_range = char_start..char_start + non_latin_1_run_length;
-                    let (replace, char_restart) =
-                        errors.handle_encode_error(full_data, char_range.clone(), ERR_REASON)?;
-                    match replace {
-                        EncodeReplace::Str(s) => {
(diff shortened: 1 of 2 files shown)


coolreader18 (commented):
Oh yea, so far this looks good! That is the correct way to add a new encoding; if that's what you were looking for. I'm not sure if `unicode_encode_ucs1` is reusable for other encodings, since anything `>ucs1` would be some way of splitting the ucs2/ucs4 into parts, at which point it's just a different encoding.

fanninpm:
> I'm not sure if `unicode_encode_ucs1` is reusable for other encodings, since anything `>ucs1` would be some way of splitting the ucs2/ucs4 into parts, at which point it's just a different encoding.

`unicode_encode_ucs1` is applicable to `latin_1` and `ascii`, so that's why I put that function outside the module.

> That is the correct way to add a new encoding; if that's what you were looking for.

While I was implementing this, I came across something that stymied me. In PyPy, the `unicode_encode_ucs1` function references an error handler that is [not yet implemented here](https://github.com/RustPython/RustPython/blob/3406daf0cc676acd3ee361b665df411159e969c1/vm/src/stdlib/codecs.rs#L108-L116). How would that error handler be implemented in Rust?

coolreader18:
Originally posted this in #3118 but it makes more sense here

re: the ucs1 helper function, I think it makes more sense to abstract over utf8/ascii than ascii/latin1, since we use a utf8 str type (so we can just validate + extend_from_slice for utf8/ascii) whereas CPython has UCS1 as a buffer (so they can just validate + memcpy for ascii/latin1). Finishing up #3118 made me remember why I did that; in our implementation of codecs, ascii and utf8 just naturally have a lot of shared code where the alternative is true in CPython.

common/src/encodings.rs
```diff
@@ -172,6 +172,70 @@ pub mod utf8 {
...
+        let mut char_data_index = 0;
+        let mut out = Vec::<u8>::new();
+        loop {
+            match data
+                .char_indices()
+                .enumerate()
+                .find(|(_, (_, c))| (*c as u32) > 255)
+            {
+                None => {
+                    out.extend_from_slice(data.as_bytes());
```
coolreader18, line 195: This is incorrect - data is utf8 formatted, so anything in `0x80..=0xff` will actually be 2 bytes wide in memory.

fanninpm: Resolved in a100178

youknowone (approved):
looks good to me. @coolreader18  could you review this again?

Issue #3039, opened 2021-09-11
Implement missing `cmath` functions


DimitrisJim:
The following functions are currently missing:

- [ ] `acos`
- [ ] `acosh`
- [ ] `asin`
- [ ] `asinh`
- [ ] `atan`
- [ ] `atanh`
- [ ] `cos`
- [ ] `cosh`
- [ ] `log`
- [ ] `log10`
- [ ] `sin`
- [ ] `sinh`
- [ ] `sqrt`
- [ ] `tan`
- [ ] `tanh` 

Inspiration for how their implementation might look can be found in `Modules/cmathmodule.c` (and `Modules/clinic/cmathmodule.c.h` for argument parsing). `Lib/src/stdlib/math.rs` equivalent functions should also be roughly similar.

---

If anyone wants to pick a function or two up, leave a comment naming the functions so other people can see who's working on what. Hopefully that will minimize conflicts.

CTZxVULKAN:
Hey i am very new but i would love to contribute could you tell me which files do i have to edit ?

DimitrisJim:
Hey @CTZxVULKAN the module is located at `vm/sec/stdlib/cmath.rs`,  you should probably look at other modules (like `math.rs` in the same directory) to get a feel for the type of constructs needed for defining a new built-in function.

After making your changes don't forget to test things out by running the Python tests with `cargo run --release -- -m test` (add `-j8` to the end of that for a faster multicore execution).

CTZxVULKAN:
nvm this is too complex for me rn. 
i will comebacc later when i am more confident :(

DimitrisJim:
You could drop me a message on gitter and I'll walk through most of it with you if you want. Everything is complex until you start doing mistakes with it and learning from them.

hk1997:
@DimitrisJim 

Hi I would like to pick up acos and acosh,
Though I'm very new to Rust so might require some guidance on the same.
In the math.rs module I see all the functions already implemented, for example:

```
// Trigonometric functions:
fn math_acos(x: IntoPyFloat, vm: &VirtualMachine) -> PyResult<f64> {
    let x = x.to_f64();
    if x.is_nan() || (-1.0_f64..=1.0_f64).contains(&x) {
        Ok(x.acos())
    } else {
        Err(vm.new_value_error("math domain error".to_owned()))
    }
}
```

On going through x.acos(), the implementation is:
```
#[must_use = "method returns a new number and does not mutate the original value"]
[...]

PR #3047, opened 2021-09-13
Implemented sqrt for cmath

Codemonk-adi:
This is my first time working with rust, pls tell me if my code is wrong, i ran the test but couldnt unserstand the output

### DIFF
--- a/Lib/test/test_cmath.py
+++ b/Lib/test/test_cmath.py
@@ -55,10 +55,7 @@ class CMathTests(unittest.TestCase):
     # commented out to allow incremented addition of functions.
     #
     # list of all functions in cmath
-    # test_functions = [getattr(cmath, fname) for fname in [
-    #        'acos', 'acosh', 'asin', 'asinh', 'atan', 'atanh',
-    #        'cos', 'cosh', 'exp', 'log', 'log10', 'sin', 'sinh',
-    #        'sqrt', 'tan', 'tanh']]
+    test_functions = [getattr(cmath, fname) for fname in ['sqrt']]
     # test first and second arguments independently for 2-argument log
     # test_functions.append(lambda x : cmath.log(x, 1729. + 0j))
     # test_functions.append(lambda x : cmath.log(14.-27j, x))
@@ -179,8 +176,6 @@ class CMathTests(unittest.TestCase):
         self.assertEqual(repr(cmath.nan), "nan")
         self.assertEqual(repr(cmath.nanj), "nanj")
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_user_object(self):
         # Test automatic calling of __complex__ and __float__ by cmath
         # functions
@@ -279,8 +274,6 @@ class CMathTests(unittest.TestCase):
             self.assertRaises(SomeException, f, MyComplexException())
             self.assertRaises(SomeException, f, MyComplexExceptionOS())
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_input_type(self):
         # ints should be acceptable inputs to all cmath
         # functions, by virtue of providing a __float__ method
@@ -326,11 +319,16 @@ class CMathTests(unittest.TestCase):
             'sinh' : real_line,
             'sqrt' : nonnegative,
             'tan' : real_line,
-            'tanh' : real_line}
+            'tanh' : real_line
+        }
 
         for fn, values in test_functions.items():
             float_fn = getattr(math, fn)
-            complex_fn = getattr(cmath, fn)
+            # TODO: RUSTPYTHON replace with 'complex_fn = getattr(cmath, fn)'
+            # when all functions have been added.
+            complex_fn = getattr(cmath, fn, None)
+            if complex_fn is None:
+                continue
             for v in values:
                 z = complex_fn(v)
                 self.rAssertAlmostEqual(float_fn(v), z.real)
--- a/vm/src/stdlib/cmath.rs
+++ b/vm/src/stdlib/cmath.rs
@@ -72,7 +72,11 @@ mod cmath {
         let z = z.to_complex();
         result_or_overflow(z, z.exp(), vm)
     }
-
+    /// Calculates the square Root of the complex number
+    #[pyfunction]
+    fn sqrt(z: IntoPyComplex) -> Complex64 {
+        z.to_complex().sqrt()
+    }
     #[derive(FromArgs)]
     struct IsCloseArgs {
         #[pyarg(positional)]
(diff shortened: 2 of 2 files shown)


vm/src/stdlib/cmath.rs
```diff
@@ -73,6 +73,12 @@ mod cmath {
         result_or_overflow(z, z.exp(), vm)
     }
 
+    //caluclates square Root of the complex number
```
youknowone, line 76: We don't need this comment. The method name and module obviously show what it is. Instead of it, we need doc-comment about this function to add document.
Doc-comment starts with 3 slashes like `///`. The contents can be found from CPython. (try cmath.sqrt.__doc__ in python)

youknowone (changes requested):
Thank you for contributing! The added code looks great but I have a few comments.

1. Please do not add `Cargo.lock` changes here. I have no idea why it is changed without any `Cargo.toml` changes but I think this is not what you intended.
2. Please run rustfmt. `cargo fmt --all` is usually enough.
3. You may forgot to correctly set user.name and user.email. Search for `git config user.email` how to do it and revise your commit metadata by `git commit --amend --reset-author`

fanninpm:
> i ran the test but couldnt unserstand the output

I don't believe we're currently at the point where we can confidently test your code just by running the CPython test suite (see [these lines of `test_cmath.py`](https://github.com/RustPython/RustPython/blob/a3bee931e74547b3ceeec13c57037132c0edbc6f/Lib/test/test_cmath.py#L53-L64)). Perhaps, in the interim, we could put some temporary tests in the `extra_tests/snippets/` directory.

youknowone:
@fanninpm  In this case, I think we'd better to edit the test file to run actual test rather than adding our own test. The running is suppressed but there is already tests

vm/src/stdlib/cmath.rs
youknowone: We want actual python document, not the original version.

youknowone:
a tip for new open source contributors: never use `git merge` or `git pull`. we only need `git fetch` and `git rebase`. (or `git pull --rebase`, but too easy to make mistake)

vm/src/stdlib/cmath.rs
Codemonk-adi: You mean just this one function?
Thankyou for being so helpful!

vm/src/stdlib/cmath.rs
```diff
@@ -72,7 +72,11 @@ mod cmath {
         let z = z.to_complex();
         result_or_overflow(z, z.exp(), vm)
     }
-
+    /// Calculates the square Root of the complex number
```
DimitrisJim, line 75:
```suggestion
    /// Return the square root of z.
```

DimitrisJim:
Fixed up the tests a bit to allow `sqrt` to run, I'll take care of the other PRs too and (when they all hopefully finish) bring `test_cmath` to its original form.

Lib/test/test_cmath.py
```diff
@@ -179,8 +176,6 @@ def test_infinity_and_nan_constants(self):
         self.assertEqual(repr(cmath.nan), "nan")
         self.assertEqual(repr(cmath.nanj), "nanj")
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
```
hk1997, line 183: Is there any specific reason to remove this line

Lib/test/test_cmath.py
```diff
@@ -55,10 +55,7 @@ class CMathTests(unittest.TestCase):
     # commented out to allow incremented addition of functions.
     #
     # list of all functions in cmath
-    # test_functions = [getattr(cmath, fname) for fname in [
-    #        'acos', 'acosh', 'asin', 'asinh', 'atan', 'atanh',
-    #        'cos', 'cosh', 'exp', 'log', 'log10', 'sin', 'sinh',
-    #        'sqrt', 'tan', 'tanh']]
+    test_functions = [getattr(cmath, fname) for fname in ['sqrt']]
```
hk1997, line 58: I was hoping let's keep the above commented lines intact in the comment itself and create a new variable.
Since I'm also working on this, we can easily compare what all is supposed to be implemented and what is already implemented.

Lib/test/test_cmath.py
```diff
@@ -314,19 +307,20 @@ def test_cmath_matches_math(self):
...
-            'log10' : positive,
-            'sin' : real_line,
-            'sinh' : real_line,
+            # 'acos' : unit_interval,
+            # 'asin' : unit_interval,
+            # 'atan' : real_line,
+            # 'cos' : real_line,
+            # 'cosh' : real_line,
+            # 'exp' : real_line,
+            # 'log' : positive,
```
hk1997, line 316: Why are you commenting this?

hk1997:
> Fixed up the tests a bit to allow `sqrt` to run, I'll take care of the other PRs too and (when they all hopefully finish) bring `test_cmath` to its original form.

Hi @DimitrisJim , I'm about to raise a PR for acos and acosh implementation, were these comments in the test file suggested by you to @Codemonk-adi ? 

Should I wait for his PR to be merged?

Lib/test/test_cmath.py
fanninpm: This line is removed when the test passes.

Lib/test/test_cmath.py
DimitrisJim: > we can easily compare what all is supposed to be implemented and what is already implemented

That's basically what the issue is there for. If something isn't implemented, you'd see it in the issue (instead of browsing to a test file).

Lib/test/test_cmath.py
DimitrisJim: Because the names are looked up further down in a `getattr`.

DimitrisJim:
> Should I wait for his PR to be merged?

Yup. Lets wait and see how we'll be running the tests. I'm personally open to either having a new variable or doing it as I've currently done. Since I'll try and follow most PRs for `cmath.rs` I think I can keep track of things whatever the case.

@youknowone opinions here?

youknowone:
My suggestion is replcing line 331 like below:
```rust
            complex_fn = getattr(cmath, fn, None)
            if complex_fn is None:
                continue
```

Then newly added function always will be tested but no error will be raised for not added one.
We can finally remove it when every functions are finally added.

DimitrisJim:
Yup, that was a better idea. (Ref #3039 to keep track of the PRs.)

DimitrisJim (approved):
Looks good to me, thanks!

youknowone:
@Codemonk-adi `user.email` doesn't look correctly set

Codemonk-adi:
@youknowone
I did git config --global user.name "Codemonk-adi"
and then pushed, can you pls tell me what should I do?

Snowapril:
@Codemonk-adi You should also set `git config --global user.email "your email"` then problems will be disappeared 😊

PR #3048, opened 2021-09-13
revert git-add from precommit

youknowone:
cc @eldpswp99 

I found this is breaking partial commit workflow at all.
`cargo clean` and rerunning will regenerate pre-commit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,6 @@ criterion = "0.3"
 
 [package.metadata.precommit]
 fmt = "cargo fmt --all"
-git-add = "git add -u"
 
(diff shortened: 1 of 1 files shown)


eldpswp99:
can you give me some detail about this? Is `git add -u` caused some problem?

youknowone:
When we edited multiples files at same time (e.g. file a.rs, b.rs, c.rs) and partially commiting them (e.g. just a.rs),
The script automatically adds b.rs and c.rs during the pre-commit stage.

eldpswp99:
Ah, that's right. Thank you for checking!

PR #2974, opened 2021-08-26
All exceptions are now modified with `extend_exception!` macro

sobolevn:
So, I've decided to go with the simplest `extend_exception!` I can think of.

Why?
1. It solves the problem
2. It is flexible enough to write any code we want
2. We already had something similar in `pub fn extend(ctx: &PyContext)`
3. We don't actually need exception structs
4. We cannot write fancy things with module level declarations, because we don't have `ctx` yet

---

**Update**: it was really hard to make it work with subclasses ad custom `__new__` and `__init__` in mind.
So, I've switched to a more complex `struct`-based macro.

Closes #2973 
Refs #2960
Refs #2897

PR #3057, opened 2021-09-14
make magic method new to follow cpython way

eldpswp99:
this issue was first mentioned in #2974 
This makes tp_new to be called by `object.__new__`

for example, with `tuple.__new__`

before: `<bound method __new__ of <class 'tuple'>>`
after: `<bound method object.__new__ of <class 'tuple'>>`

well, i have problem with controlling the new with pysuper, i remain this as draft now
I would be thankful for any advice!

### DIFF
--- a/vm/src/builtins/object.rs
+++ b/vm/src/builtins/object.rs
@@ -3,6 +3,7 @@ use super::list::PyList;
 use super::pybool;
 use super::pystr::{PyStr, PyStrRef};
 use super::pytype::PyTypeRef;
+use crate::builtins::pytype;
 use crate::builtins::pytype::PyType;
 use crate::common::hash::PyHash;
 use crate::function::FuncArgs;
@@ -33,6 +34,11 @@ impl PyValue for PyBaseObject {
 
 #[pyimpl(flags(BASETYPE))]
 impl PyBaseObject {
+    #[pyclassmethod(magic)]
+    fn new(cls: PyTypeRef, args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+        let (subtype, args): (PyTypeRef, FuncArgs) = args.bind(vm)?;
+        pytype::call_tp_new(cls, subtype, args, vm)
+    }
     /// Create and return a new object.  See help(type) for accurate signature.
     #[pyslot]
     fn tp_new(cls: PyTypeRef, _args: FuncArgs, vm: &VirtualMachine) -> PyResult {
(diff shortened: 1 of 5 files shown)


vm/src/builtins/object.rs
```diff
@@ -33,6 +34,11 @@ impl PyValue for PyBaseObject {
 
 #[pyimpl(flags(BASETYPE))]
 impl PyBaseObject {
+    #[pyclassmethod(magic)]
+    fn new(cls: PyTypeRef, args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+        let (subtype, args): (PyTypeRef, FuncArgs) = args.bind(vm)?;
```
youknowone, line 39: `SlotConstructor` will be preferred for newly added `__new__` with binding.

eldpswp99: I thought there may be collision with object.tp_new. is it okay to put both?

youknowone: oh! I am sorry. you are right. this is not `tp_new` but just `__new__`.

vm/src/pyobject.rs
```diff
@@ -144,7 +136,6 @@ impl PyContext {
             exceptions,
             int_cache_pool,
             string_cache,
-            tp_new_wrapper,
```
youknowone, line 147: we need this field yet in this implementation

youknowone:
`__new__` will not be looked up from tp_new anymore and the wrapper function `__new__` will be added to classes only when tp_new slot is defined.

youknowone:
@coolreader18 @DimitrisJim could you review this PR?

vm/src/builtins/pytype.rs
```diff
@@ -239,6 +239,19 @@ impl PyTypeRef {
 
 #[pyimpl(with(SlotGetattro, SlotSetattro, Callable), flags(BASETYPE))]
 impl PyType {
+    // bound method for every type
+    pub(crate) fn __new__(zelf: PyRef<Self>, args: FuncArgs, vm: &VirtualMachine) -> PyResult {
```
DimitrisJim, line 243: `__new__` is sort of confusing to me, I'd prefer if this kept `tp_new_wrapper` so we can easily see which function this maps to in CPython.

DimitrisJim:
this looks just fine to me but I've not really looked into it much so I'll leave the approval for Noa.

vm/src/builtins/pytype.rs
youknowone: This is a body of bound method `__new__` just like `mro`. The only difference of them is `__new__` is a bound method for every type including `type` but `mro` is an unbound method for `type` and bound as boundmethod for other types.
I think we'd better to keep this name as same as other methods when it works like other methods.

coolreader18 (approved):
LGTM, I think :sweat_smile: 

This whole part of the interpreter is pretty complex and I don't really have an intuition for how it works, so given that this seems simpler and still works it's got my approval :+1:

PR #3058, opened 2021-09-14
revise PySuper::getattro same as CPython super_getattro

youknowone:
possibly related to #3057 but didn't solve the problem itself


youknowone:
@eldpswp99 could you review this PR? I think you also reviewed cpython super_getattro

eldpswp99 (approved):
LGTM! I think this revision follows CPython more than before.

fanninpm:
Does this affect the results of `test_pysuper.py`?

youknowone:
rebased it onto test_super.py but nothing changed.

DimitrisJim (approved):
making the names follow the ones used in CPython made this easy to review. p.s the soundforge bug report (SF ID #743627) is actually a dead-end, probably a relic left over in the CPython code.

youknowone:
I replaced it to alive link

PR #3060, opened 2021-09-14
Change posix.*chmod

ChJR:
Added fchmod and lchmod to posix module.
Changed posix.chmod to handle file descriptor.


vm/src/stdlib/os.rs
```diff
@@ -2539,21 +2539,52 @@ mod posix {
...
+        match path {
+            PathOrFd::Path(path) => {
+                let body = move || {
+                    use std::os::unix::fs::PermissionsExt;
+                    let meta = fs_metadata(&path, follow_symlinks.0)?;
+                    let mut permissions = meta.permissions();
+                    permissions.set_mode(mode);
+                    fs::set_permissions(&path, permissions)
+                };
+                body().map_err(|err| err.into_pyexception(vm))
```
youknowone, line 2558:
```suggestion
                body()
```

vm/src/stdlib/os.rs
```diff
@@ -2539,21 +2539,52 @@ mod posix {
...
+            }
+            PathOrFd::Fd(fd) => nix::sys::stat::fchmod(
+                fd,
+                nix::sys::stat::Mode::from_bits(mode as libc::mode_t).unwrap(),
+            )
+            .map_err(|err| err.into_pyexception(vm)),
+        }
```
youknowone, line 2565:
```suggestion
        }
        .map_err(|err| err.into_pyexception(vm)),
```

vm/src/stdlib/os.rs
```diff
@@ -2539,21 +2539,52 @@ mod posix {
...
+                    permissions.set_mode(mode);
+                    fs::set_permissions(&path, permissions)
+                };
+                body().map_err(|err| err.into_pyexception(vm))
+            }
+            PathOrFd::Fd(fd) => nix::sys::stat::fchmod(
+                fd,
+                nix::sys::stat::Mode::from_bits(mode as libc::mode_t).unwrap(),
+            )
+            .map_err(|err| err.into_pyexception(vm)),
```
youknowone, line 2564: is this same as calling `fchmod`?

youknowone (changes requested):

vm/src/stdlib/os.rs
ChJR: Yes. In https://docs.python.org/3/library/os.html#os.fchmod , 

> As of Python 3.3, this is equivalent to os.chmod(fd, mode).

ChJR:
I tried to apply those changes, but it shows compile error due to type mismatch.

vm/src/stdlib/os.rs
youknowone: my fault, `into_pyexception` cannot be merged

youknowone:
I am sorry, I thought Err types were PyExceptionRef but `into_pyexception` means they are not.

vm/src/stdlib/os.rs
youknowone: Then I expect this is just calling fchmod.

youknowone: I also want lchmod and chmod share their parts if they do same function.

ChJR: OK. I will make them integrate with chmod.

vm/src/stdlib/posix.rs
```diff
@@ -687,21 +687,54 @@ pub mod module {
...
+                    fs::set_permissions(&path, permissions)
+                };
+                body().map_err(|err| err.into_pyexception(vm))
+            }
+            PathOrFd::Fd(fd) => nix::sys::stat::fchmod(
+                fd,
+                nix::sys::stat::Mode::from_bits(mode as libc::mode_t).unwrap(),
+            )
+            .map_err(|err| err.into_pyexception(vm)),
+        }
```
youknowone, line 713: Because `chmod` is superset of `fchmod` and `lchmod`,  I think `chmod` calling `fchmod` and `lchmod` is more natural way.
It seems you decided to do this way because `chmod` is not fully consistable with python `fchmod` and `lchmod` because they are not the full feature due to lack of `FollowSymlinks`. In such case, I still think adding `_fchmod` and `_lchmod` with `FollowSymlinks` and using them for all `chmod`, `fchmod` and `lchmod` is better.

Even though I think this not a common design, I still can accept this approach because it have a reasonable excuse of its reversed design flow. But because this is counterintuitive decision, I think the decision itself must be documented enough not to confuse future contributors to think this is a common approach in this project.

youknowone (approved):

vm/src/stdlib/posix.rs
ChJR: OK. I changed these.

vm/src/stdlib/posix.rs
```diff
@@ -704,6 +703,40 @@ pub mod module {
...
+        follow_symlinks: FollowSymlinks,
+        vm: &VirtualMachine,
+    ) -> PyResult<()> {
+        match path {
+            PathOrFd::Path(path) => _chmod(path, dir_fd, mode, follow_symlinks, vm),
+            PathOrFd::Fd(fd) => _fchmod(fd, mode, vm),
+        }
+    }
+
+    #[cfg(not(target_os = "redox"))]
```
youknowone, line 728: cfg is not applied to utility functions

ChJR: My mistake.

youknowone (approved):

PR #3062, opened 2021-09-14
better error for wrong py* attributes in pyclass

youknowone:


derive/src/pyclass.rs
```diff
@@ -890,12 +890,16 @@ fn extract_impl_attrs(attr: AttributeArgs) -> std::result::Result<ExtractedImplA
...
-fn new_impl_item<Item>(index: usize, attr_name: String) -> Box<dyn ImplItem<Item>>
+fn new_impl_item<Item>(
+    attr: &Attribute,
+    index: usize,
+    attr_name: String,
+) -> Result<Box<dyn ImplItem<Item>>>
 where
     Item: ItemLike + ToTokens + GetIdent,
 {
     assert!(ALL_ALLOWED_NAMES.contains(&attr_name.as_str()));
```
DimitrisJim, line 901: is this assert even necessary here?

DimitrisJim (approved):

derive/src/pyclass.rs
youknowone: I don't remember what it exactly check but generally this assertions are not for vm programmers but for derive programmers.

DimitrisJim: As far as I can tell, the match statement should cover what it tries to cover (only allow certain names) but I'll let you decide if you want to keep it or remove it.

youknowone: I don't think i will investigate it for now to decide whether remove it or not. I'd better to add comments what it does when I added it. It was added when adding new subattributes like pyslot, so I think it was useful for that kind of jobs. Maybe decidable when adding new subattribute.

PR #3063, opened 2021-09-14
Add $module for function text signatures.

DimitrisJim:
I.e just how methods have `$self` at the start of the sig, functions have `$module`.

### DIFF
--- a/derive/src/util.rs
+++ b/derive/src/util.rs
@@ -505,17 +505,29 @@ where
     Ok(())
 }
 
-// Best effort attempt to create a template from which a
+// Best effort attempt to generate a template from which a
 // __text_signature__ can be created.
-pub(crate) fn get_func_sig(sig: &Signature, name: &str) -> String {
-    let args: Vec<_> = sig
-        .inputs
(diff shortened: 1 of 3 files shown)


derive/src/util.rs
```diff
@@ -528,6 +540,7 @@ pub(crate) fn get_func_sig(sig: &Signature, name: &str) -> String {
             let ident = match arg.pat.as_ref() {
                 syn::Pat::Ident(p) => p.ident.to_string(),
                 // FIXME: other => unreachable!("function arg pattern must be ident but found `{}`", quote!(fn #ident(.. #other ..))),
+                // FIXME: Pat::PatStruct should be captured though and added to sig. (Common destructuring with custom Args)
```
youknowone, line 543: I think FromArgs trait need to have something like `FromArgs::text_signature()` to describe args inside of it. PatStruct will not cover when FromArgs struct is used as a single argument, but trait static method will cover both cases.

DimitrisJim: I'll look into that, dropped the comment for now.

youknowone (approved):

PR #3064, opened 2021-09-15
Implement __reduce__ and __reduce_ex__ for array

qingshi163:


vm/src/stdlib/array.rs
```diff
@@ -1037,6 +1059,47 @@ mod array {
...
+                let bytes = vm.ctx.new_bytes(array.get_bytes().to_vec());
+                let code = MachineFormatCode::from_typecode(array.typecode()).unwrap();
+                let code = PyInt::from(u8::from(code)).into_object(vm);
+                let func = _array_reconstructor.into_func();
+                let func =
+                    PyNativeFuncDef::new(func, PyStr::from("_array_reconstructor").into_ref(vm));
+                let module = zelf.class().module(vm);
+                let func = func.into_function().with_module(module).build(&vm.ctx);
+                (
+                    func,
```
qingshi163, line 1089: How could I get the module function _array_reconstructor here?

vm/src/stdlib/array.rs
```diff
@@ -545,6 +560,13 @@ mod array {
         fn byteswap(self) -> Self {
             Self(self.0.swap_bytes())
         }
+        fn to_object(self, vm: &VirtualMachine) -> PyObjectRef {
+            vm.ctx.new_str(
+                char::from_u32(self.0 as u32)
+                    .unwrap_or_default()
+                    .to_string(),
+            )
```
qingshi163, line 568: is the convert safe?

vm/src/stdlib/array.rs
DimitrisJim: `vm.import` followed by a `get_attribute` should also work.

vm/src/builtins/float.rs
```diff
@@ -51,6 +51,14 @@ impl From<f64> for PyFloat {
     }
 }
 
+impl From<f32> for PyFloat {
+    fn from(value: f32) -> Self {
+        PyFloat {
+            value: value as f64,
+        }
+    }
+}
```
youknowone, line 60: Since Rust is strict for f32 to f64 conversion, I wish we also do same policy for RustPython codes. `impl_array_element` can take `PyFloat::from` instead of `PyFloat`. Then `f32` can be covered by `pyfloat_from_f32` instead of `PyFloat::from`

vm/src/stdlib/array.rs
youknowone: WideChar can contain invalid utf-8 character like surrogate. So I don't think so. To be safe, It must be decoded.

youknowone (changes requested):

vm/src/stdlib/array.rs
youknowone: I wish we can access them through VirtualMachine in future.

vm/src/stdlib/array.rs
qingshi163: Are we assuming the data in the array is decoded?

youknowone: I am not sure. I guess this is corresponding to WChar unicode object in CPython - which is deprecated same as this type array.
Even if it is regarded as a decoded string, because this is array, it still can contain surrogate character, which is not a valid utf8 string.
So my last comment was wrong. Regardless it is decoded or not, it can contains invalid utf8 character.

qingshi163:
@youknowone Can you check what went wrong in windows tests and fix it?

extra_tests/snippets/stdlib_array.py
```diff
@@ -96,4 +97,10 @@ def test_array_frombytes():
...
     a[0] = 42
 with assert_raises(IndexError):
-    del a[42]
\ No newline at end of file
+    del a[42]
+
+test_str = '🌉abc🌐def🌉🌐'
+u = array('u', test_str)
+assert u.__reduce_ex__(1)[1][1] == list(test_str)
+assert str(loads(dumps(u, 1))) == f"array('u', '{test_str}')"
```
youknowone, line 105:
```suggestion
assert str(loads(dumps(u, 1))) == f"array('u', '{test_str}')", str(loads(dumps(u, 1)))
```

then it will shows the value on assertion failure

youknowone (commented):
I am really sorry. I don't have an accessible windows machine for now and until early october.

can anyone help this?

extra_tests/snippets/stdlib_array.py
fanninpm: (_jedi mind trick wave_) [That was not the line you're looking for…](https://github.com/RustPython/RustPython/pull/3064/checks?check_run_id=3621875306#step:12:418)

fanninpm:
@coolreader18 might be able to diagnose the Windows problem

youknowone:
@qingshi163 could you rebase this PR? I will check windows problem during this week

stdlib/src/array.rs
```diff
@@ -1077,6 +1109,52 @@ mod array {
...
+        fn reduce(
+            zelf: PyRef<Self>,
+            vm: &VirtualMachine,
+        ) -> PyResult<(PyObjectRef, PyObjectRef, Option<PyDictRef>)> {
+            let array = zelf.read();
+            let cls = zelf.as_object().clone_class().into_object();
+            let typecode = vm.ctx.new_utf8_str(array.typecode_str());
+            let values = if array.typecode() == 'u' {
+                let s = Self::_wchar_bytes_to_string(array.get_bytes(), array.itemsize(), vm)?;
+                s.chars().map(|x| x.into_pyobject(vm)).collect()
```
youknowone, line 1147: In windows, I tried this code with CPython:
```python
test_str = '🌉abc🌐def🌉🌐'
u = array('u', test_str)
print(u.__reduce_ex__(1))
```
then
```
  File ".\extra_tests\snippets\stdlib_array.py", line 104, in <module>
    assert u.__reduce_ex__(1)[1][1] == list(test_str), (u.__reduce_ex__(1)[1][1], list(test_str))
AssertionError: (['\ud83c', '\udf09', 'a', 'b', 'c', '\ud83c', '\udf10', 'd', 'e', 'f', '\ud83c', '\udf09', '\ud83c', '\udf10'], ['🌉', 'a', 'b', 'c', '🌐', 'd', 'e', 'f', '🌉', '🌐'])
```
The values are wchar encoded in CPython but doesn't look like that here

qingshi163: Do you mean CPython did not treat it as same in linux macos?

youknowone: I don't mind if you want to just skip testing for 16bit character environments.

PR #3066, opened 2021-09-15
PyStrKind to optimize ascii strings

youknowone:
And this structure also can be used for invalid unicode strings and ucs16/ucs32 representations in CPython in future.

### DIFF
--- a/vm/src/builtins/pystr.rs
+++ b/vm/src/builtins/pystr.rs
@@ -34,11 +34,45 @@ use unic_ucd_ident::{is_xid_continue, is_xid_start};
 use unicode_casing::CharExt;
 
 /// Utf8 + state.ascii (+ PyUnicode_Kind in future)
-#[derive(Debug)]
+#[derive(Debug, Copy, Clone, PartialEq)]
 pub(crate) enum PyStrKind {
+    Ascii,
+    Utf8,
+}
+
+impl std::ops::BitOr for PyStrKind {
+    type Output = Self;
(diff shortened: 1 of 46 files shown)


vm/src/builtins/pystr.rs
```diff
@@ -294,44 +339,52 @@ impl PyStr {
     }
 
     pub fn as_str(&self) -> &str {
-        &self.value
```
DimitrisJim, line 297: Safety comment? I'm haven't seen a specific rule throughout the project but I'm guessing comments describing safety should be required.

Also, wouldn't this benefit from being marked as `#[inlined(always)]?

youknowone: I added safey comments for every unsafe. About inline, I added one but I also tihnk these kind of tiny functions are easily marked as inline by compiler

DimitrisJim: [Reading a bit about this](https://internals.rust-lang.org/t/inlining-policy-for-functions-in-std/14189/10), for private functions you're probably right. Cross crate public ones don't automatically get inlined. It's probably not harmful to add in this case.

DimitrisJim (approved):

coolreader18:
This is unsound - you can pass a non-ascii string to new_ascii_str, and a non-utf8 string to `::from((v, Kind::Utf8))` that'll get passed without validation to `unsafe { str::from_utf8_unchecked }`

PR #3068, opened 2021-09-15
PyStr::char_all

youknowone:


DimitrisJim (approved):
looks good, shame `is_printable` ruins things by returning `True` for empty strings.

PR #3072, opened 2021-09-16
Remove HEAPTYPE flag from builtins.

DimitrisJim:
`HEAPTYPE` isn't set for built-in types, instead only used for dynamically created types. (see `<bltin_type>.__flags__ & (1 << 9)`)

The errors that appear now are a result of this, `copyreg` doesn't know how to `__reduce__` types that don't have the `HEAPTYPE` flag so any builtins that lack it require an implementation of `__reduce__` (currently, `Exception` types and types defined in `_ast` it seems).

### DIFF
--- a/Lib/test/test_ast.py
+++ b/Lib/test/test_ast.py
@@ -565,6 +565,8 @@ class AST_Tests(unittest.TestCase):
         x = ast.Sub()
         self.assertEqual(x._fields, ())
 
+    # TODO: RUSTPYTHON _ast classes should be HEAPTYPES (except for _ast.AST)
+    @unittest.expectedFailure
     def test_pickling(self):
         import pickle
         mods = [pickle]
--- a/vm/src/builtins/pytype.rs
+++ b/vm/src/builtins/pytype.rs
@@ -467,6 +467,10 @@ impl PyType {
             }
         }
 
+        // TODO: how do we know if it should have a dict?
+        // All *classes* should have a dict. Exceptions are *instances* of
+        // classes that define __slots__ and instances of built-in classes
+        // (with exceptions, e.g function)
         if !attributes.contains_key("__dict__") {
             attributes.insert(
                 "__dict__".to_owned(),
@@ -479,9 +483,10 @@ impl PyType {
             );
         }
 
-        // TODO: how do we know if it should have a dict?
-        let flags = base.slots.flags | PyTpFlags::HAS_DICT;
-
+        // TODO: Flags is currently initialized with HAS_DICT. Should be
+        // updated when __slots__ are supported (toggling the flag off if
+        // a class has __slots__ defined).
+        let flags = PyTpFlags::heap_type_flags() | PyTpFlags::HAS_DICT;
         let slots = PyTypeSlots::from_flags(flags);
 
         let typ = new(metatype, name.as_str(), base, bases, attributes, slots)
--- a/vm/src/builtins/pytype.rs
+++ b/vm/src/builtins/pytype.rs
@@ -467,7 +467,6 @@ impl PyType {
             }
         }
 
-        // TODO: how do we know if it should have a dict?
         // All *classes* should have a dict. Exceptions are *instances* of
         // classes that define __slots__ and instances of built-in classes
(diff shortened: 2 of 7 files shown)


vm/src/builtins/pytype.rs
```diff
@@ -480,8 +480,10 @@ impl PyType {
         }
 
         // TODO: how do we know if it should have a dict?
-        let flags = base.slots.flags | PyTpFlags::HAS_DICT;
-
+        // All *classes* should have a dict. Exceptions are *instances* of
```
DimitrisJim, line 483: revised this. `HAS_DICT` can be used (temporarily?) as `tp_dictoffset` is used in CPython.

Lib/test/test_ast.py
```diff
@@ -565,6 +565,12 @@ def test_no_fields(self):
         x = ast.Sub()
         self.assertEqual(x._fields, ())
 
+    # TODO: RUSTPYTHON
+    # In CPython classes in _ast with the exception of `AST` are created dynamically
+    # (by calling PyType_Type, see line 790 in asdl_c.py). As a result, they have the
+    # HEAPTYPE flag set and copyreg pickles them. We currently create them without the
+    # flag, apparently.
```
youknowone, line 572: This part looks more like discussion for the issue. Could you open an issue about this and start the thread with this comment? Then other contributors who find something about this can post their opnions easier than comments.

vm/src/builtins/pytype.rs
```diff
@@ -480,8 +480,10 @@ impl PyType {
         }
 
         // TODO: how do we know if it should have a dict?
```
youknowone, line 482: I think this comment is here because the next line is including `HAS_DICT`. Now it is hidden into `py_flags()`. Then the comment also need to be moved.
Otherwise, you seem to find the answer of it. Then we don't need it anymore.

vm/src/slots.rs
```diff
@@ -25,8 +25,16 @@ bitflags! {
...
 impl PyTpFlags {
+    // Default used for both built-in and normal classes: empty, for now.
     // CPython default: Py_TPFLAGS_HAVE_STACKLESS_EXTENSION | Py_TPFLAGS_HAVE_VERSION_TAG
-    pub const DEFAULT: Self = Self::HEAPTYPE;
+    pub const DEFAULT: Self = Self::empty();
+
+    // CPython: See initialization of flags in type_new.
+    /// Used for types created in Python. Subclassable, have a dictionary and are a
+    /// heaptype.
+    pub fn py_flags() -> Self {
```
youknowone, line 35: by the comment you left, I think `heap_type_flags()` fits better for its name.

vm/src/builtins/pytype.rs
DimitrisJim: Yeah. This isn't the place for this, it should actually be placed exactly above when the `__dict__` is inserted. It's still a `TODO` though because we still don't have slots.

youknowone: Then I think `Self::HAS_DICT` is not belong to `heap_type_flags()`. In that case, I prefer `heap_type_flags() | PyTypeFlags::HAS_DICT` with the comment, to show it is always added for now but that's not because it is the default flag for heap type but we are not implementing type without `__dict__` yet.

Lib/test/test_ast.py
DimitrisJim: Yup. I'll try and open an issue for this at some point.

vm/src/builtins/pytype.rs
```diff
@@ -467,6 +467,10 @@ impl PyType {
             }
         }
 
+        // TODO: how do we know if it should have a dict?
```
youknowone, line 470:
```suggestion
```

youknowone (approved):
Looks good, will you check about the TODO you answered? I think it is not a TODO anymore. because it is about the class dict, not the instance dict.

vm/src/builtins/pytype.rs
DimitrisJim: ah, missed it! (btw for small changes like these, feel free to commit the change directly)

PR #3074, opened 2021-09-16
FIxed NaN issue

Codemonk-adi:


Snowapril:
maybe you need to drop modification in [`example/freeze/main.rs`](https://github.com/RustPython/RustPython/pull/3074/files#diff-4c12deb96b10eaa2f8526c09b9627e279f6e7b404e6831e8522dd4590d19f4c0) and rebasing commit😊

DimitrisJim:
The newest change is good but your branch has some old commits on it (from your previous change). I believe you'd just need to rebase it and bring it up to speed with the main RustPython branch.

common/src/float_ops.rs
```diff
@@ -226,7 +226,7 @@ pub fn to_string(value: f64) -> String {
             format!("{}e{:+#03}", significand, exponent)
         }
     } else {
-        value.to_string()
+        value.to_string().to_ascii_lowercase()
```
coolreader18, line 229: You can do this in place without allocating a second string; `let mut  s = value.to_string(); s.make_ascii_lowercase(); s`

PR #3077, opened 2021-09-17
Fix compilation on unix platforms with i32 c_long

jam1garner:
Targets such as i686-unknown-linux-gnu have a [c_long definition of i32](https://docs.rs/libc/0.2/i686-unknown-linux-gnu/libc/type.c_long.html), so assuming that [`libc::pathconf`](https://docs.rs/libc/0.2.102/i686-unknown-linux-gnu/libc/fn.pathconf.html) returns an i64 will cause a compilation error on such platforms.

This patch fixes that by adding a cast to i64, which is valid as `c_long` is only ever i32 or i64, meaning this is either a no-op or a sign extension.

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -1932,7 +1932,7 @@ mod _os {
                 Err(io::Error::from(Errno::last()).into_pyexception(vm))
             }
         } else {
-            Ok(Some(raw))
+            Ok(Some(raw as i64))
         }
     }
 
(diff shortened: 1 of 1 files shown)


youknowone:
how about changing the return type to `libc::c_long` to follow its definition?

jam1garner:
> @youknowone how about changing the return type to `libc::c_long` to follow its definition?

Done, I didn't modify that originally as I wasn't sure the effects of it but there isn't much reliance on return type :+1:

youknowone (approved):

PR #3083, opened 2021-09-18
Introduce `types.GenericAlias` type

### LINKED ISSUE #3079: Implement `types.GenericAlias` type
## Feature

Since CPython 3.9, it started to support [*Type Hinting Generics in Standard Collections*](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections). So the code like below should be supported.

```python
list[int]
```

The type of `list[int]` evaluation result is `types.GenericAlias`. This should be implemented.

### Memo

- `genericaliasobject.c#PyGenericAlias` type is equal with `type(list[int])` (i.e. `types.GenericAlias`) because `list.__class_getitem__` returns the instance of the type.

## Python Documentation

- https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections
- CPython source: https://github.com/python/cpython/blob/v3.9.7/Objects/genericaliasobject.c

moreal:
## Overview

- Introduce `types.GenericAlias` type.  [7790ecb]
- Add `list.__class_getitem__` method.  [057a748]
- Bump `types` module to CPython 3.9.7 [ea53510, b9cf020]

## Description

This pull request will resolve #3079.

This pull request introduces `types.GenericAlias` type, introduced in CPython 3.9, and is the type of `list[int]` or `dict[str, int]`. It can be accessed from `types` module. And there are some features not been implemented yet.

### About tests

As more description, this `types.GenericAlias` type has a test named `test.test_genericalias` but the test requires `contextvars`:

```python
  File "Lib/test/test_genericalias.py", line 12, in <module>
    from contextvars import ContextVar, Token
ModuleNotFoundError: No module named 'contextvars'
```

And `contextvars` requires `_contextvars`:

```python
  File "Lib/test/test_genericalias.py", line 12, in <module>
    from contextvars import ContextVar, Token
  File "/home/moreal/github/RustPython/RustPython/vm/pylib-crate/Lib/contextvars.py", line 1, in <module>
    from _contextvars import Context, ContextVar, Token, copy_context
ModuleNotFoundError: No module named '_contextvars'
```

The test should be added after the `_contextvars` module is introduced.

## References

- https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections
  - PEP 585: https://www.python.org/dev/peps/pep-0585/
- https://docs.python.org/3/library/stdtypes.html?highlight=genericalias#generic-alias-type
- CPython v3.9.7 implementation: https://github.com/python/cpython/blob/v3.9.7/Objects/genericaliasobject.c

### DIFF
--- a/vm/src/builtins/genericalias.rs
+++ b/vm/src/builtins/genericalias.rs
@@ -1,4 +1,4 @@
-use crate::builtins::{PyStr, PyTuple, PyTupleRef, PyTypeRef};
+use crate::builtins::{PyStr, PyTupleRef, PyTypeRef};
 use crate::common::hash;
 use crate::slots::{Hashable, SlotConstructor};
 use crate::{
@@ -13,8 +13,6 @@ pub struct PyGenericAlias {
     args: PyTupleRef,
     parameters: PyTupleRef,
 }
-
-pub type PyGenericAliasRef = PyRef<PyGenericAlias>;
 
 impl fmt::Debug for PyGenericAlias {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
@@ -50,17 +48,14 @@ impl PyGenericAlias {
         let args: PyTupleRef = if let Ok(tuple) = PyTupleRef::try_from_object(vm, args.clone()) {
             tuple
         } else {
-            vm.ctx
-                .new_tuple(vec![args])
-                .downcast::<PyTuple>()
-                .ok()
-                .unwrap()
+            PyTupleRef::with_elements(vec![args], &vm.ctx)
         };
 
+        let parameters = make_parameters(&args, vm);
         Self {
             origin,
-            args: args.clone(),
-            parameters: make_parameters(args, vm),
+            args,
+            parameters,
         }
     }
 
@@ -123,16 +118,13 @@ impl PyGenericAlias {
 fn is_typevar(obj: PyObjectRef) -> bool {
     let class = obj.class();
     class.tp_name() == "TypeVar"
-        && match class.get_attr("__module__") {
-            Some(o) => o
-                .downcast_ref::<PyStr>()
-                .map(|s| s.as_str() == "typing")
-                .unwrap_or(false),
-            None => false,
-        }
+        && class
+            .get_attr("__module__")
+            .and_then(|o| o.downcast_ref::<PyStr>().map(|s| s.as_str() == "typing"))
+            .unwrap_or(false)
 }
 
-fn make_parameters(args: PyTupleRef, vm: &VirtualMachine) -> PyTupleRef {
+fn make_parameters(args: &PyTupleRef, vm: &VirtualMachine) -> PyTupleRef {
     let mut parameters: Vec<PyObjectRef> = vec![];
     for arg in args.as_slice() {
         if is_typevar(arg.clone()) {
@@ -147,11 +139,7 @@ fn make_parameters(args: PyTupleRef, vm: &VirtualMachine) -> PyTupleRef {
         }
     }
 
-    vm.ctx
-        .new_tuple(parameters)
-        .downcast::<PyTuple>()
-        .ok()
-        .unwrap()
+    PyTupleRef::with_elements(parameters, &vm.ctx)
 }
 
 impl Hashable for PyGenericAlias {
--- a/vm/src/builtins/genericalias.rs
+++ b/vm/src/builtins/genericalias.rs
@@ -1,0 +1,166 @@
+use crate::builtins::{PyStr, PyTuple, PyTupleRef, PyTypeRef};
+use crate::common::hash;
+use crate::slots::{Hashable, SlotConstructor};
+use crate::{
+    IdProtocol, PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject,
+    TypeProtocol, VirtualMachine,
+};
+use std::fmt;
+
+#[pyclass(module = "types", name = "GenericAlias")]
+pub struct PyGenericAlias {
+    origin: PyTypeRef,
+    args: PyTupleRef,
+    parameters: PyTupleRef,
+}
+
+pub type PyGenericAliasRef = PyRef<PyGenericAlias>;
+
+impl fmt::Debug for PyGenericAlias {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        f.write_str("GenericAlias")
+    }
+}
+
+impl PyValue for PyGenericAlias {
+    fn class(vm: &VirtualMachine) -> &PyTypeRef {
+        &vm.ctx.types.generic_alias_type
+    }
+}
+
+#[derive(FromArgs)]
+pub struct GenericAliasArgs {
+    #[pyarg(any)]
+    origin: PyTypeRef,
+    #[pyarg(any)]
+    arguments: PyObjectRef,
+}
+
+impl SlotConstructor for PyGenericAlias {
+    type Args = GenericAliasArgs;
+
+    fn py_new(cls: PyTypeRef, args: Self::Args, vm: &VirtualMachine) -> PyResult {
+        PyGenericAlias::new(args.origin, args.arguments, vm).into_pyresult_with_type(vm, cls)
+    }
+}
+
+#[pyimpl(with(Hashable), flags(BASETYPE))]
+impl PyGenericAlias {
+    pub fn new(origin: PyTypeRef, args: PyObjectRef, vm: &VirtualMachine) -> Self {
+        let args: PyTupleRef = if let Ok(tuple) = PyTupleRef::try_from_object(vm, args.clone()) {
+            tuple
+        } else {
+            vm.ctx
+                .new_tuple(vec![args])
+                .downcast::<PyTuple>()
+                .ok()
+                .unwrap()
+        };
+
+        Self {
+            origin,
+            args: args.clone(),
+            parameters: make_parameters(args, vm),
+        }
+    }
+
+    #[pymethod(magic)]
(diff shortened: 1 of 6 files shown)


vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,136 @@
+use crate::builtins::PyStr;
+use crate::builtins::PyTuple;
+use crate::builtins::PyTypeRef;
```
youknowone, line 3:
```suggestion
use crate::builtins::{PyStr, PyTuple, PyTypeRef};
```

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,136 @@
...
+use crate::common::hash;
+use std::fmt;
+
+use crate::slots::{Hashable, SlotConstructor};
+use crate::{
+    IdProtocol, PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue, VirtualMachine,
+};
+
+#[pyclass(module = false, name = "types.GenericAlias")]
```
youknowone, line 12:
```suggestion
#[pyclass(module = "types", name = "GenericAlias")]
```

I also wish you try to use `#[pymodule]` for new modules.

vm/src/builtins/object.rs
```diff
@@ -308,7 +308,7 @@ impl PyBaseObject {
 
     /// Return hash(self).
     #[pymethod(magic)]
-    fn hash(zelf: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyHash> {
+    pub fn hash(zelf: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyHash> {
```
youknowone, line 311: where is this used? I think `vm.hash()` is commonly used but not about this one. So I want to know where is this used.

moreal: Originally, I tried to use it for the `Hashable for PyGenericAlias` implementation but it is using `vm._hash` as you said. I'll drop the commit. Thanks!

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,136 @@
+use crate::builtins::PyStr;
+use crate::builtins::PyTuple;
+use crate::builtins::PyTypeRef;
+use crate::common::hash;
+use std::fmt;
+
```
youknowone, line 6:
```suggestion
```

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub fn new(origin: PyTypeRef, args: PyObjectRef, vm: &VirtualMachine) -> Self {
+        let args = if args.downcast_ref::<PyTuple>().is_some() {
+            args
+        } else {
+            vm.ctx.new_tuple(vec![args])
+        };
+
+        Self {
+            origin,
+            args: args.clone(),
```
youknowone, line 60: if this is always downcasted to PyObjectRef, I think its type must be `PyTupleRef` instead of `PyObjectRef`

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    #[pyproperty(magic)]
+    fn parameters(&self) -> PyObjectRef {
+        self.parameters.clone()
+    }
+
+    #[pyproperty(magic)]
+    fn args(&self) -> PyObjectRef {
+        self.args.clone()
```
youknowone, line 88: when `args` is `PyTuple`, this line can be `self.args.as_object().clone()`

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    #[pyproperty(magic)]
+    fn origin(&self) -> PyObjectRef {
+        self.origin.as_object().clone()
+    }
+}
+
+fn repr_item(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<String> {
```
youknowone, line 97: if this is a utility for `repr`, let's move this into `repr` function

vm/src/builtins/genericalias.rs
moreal: I fixed what you suggested. But I wonder it needs to declare a new `module`. There is already a `types` module written in Python (`types.py`). For example, `cmath` module was declared through `#[pymodule]` because there is no `cmath.py` Python file.

youknowone: pymodules can be merged. See `stdlib/time.rs` for other examples.
```rust
types::extend_module(vm, &module);
```

youknowone: I think I misunderstood something. You defined this type in `builtins::genericalias` like builtin types. But this is not a bulitin type but defined in `_testcapi` module.

youknowone: This is weird. I also found `_GenericAlias` is defined in `typing.py`. How this is working?

moreal: As I understood:
- `typing._GenericAlias` and `types.GenericAlias` are different
- CPython doesn't expose `types.GenericAlias` builtin direct type but the `types.GenericAlias` is exposed by `type(list[int])` indirectly. About this, I didn't understand its context well 😞 . `typing._GenericAlias` has a similar naming but independent class which acts differently.

youknowone: Thank you! I got it.
```
Python 3.9.5 (default, May 26 2021, 13:13:06) 
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
>>> import _testcapi
>>> _testcapi.GenericAlias
<class 'GenericAlias'>
>>> _testcapi.GenericAlias.__qualname__
'GenericAlias'
>>> _testcapi.GenericAlias.__module__
'builtins'
```

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,166 @@
...
+impl PyGenericAlias {
+    pub fn new(origin: PyTypeRef, args: PyObjectRef, vm: &VirtualMachine) -> Self {
+        let args: PyTupleRef = if let Ok(tuple) = PyTupleRef::try_from_object(vm, args.clone()) {
+            tuple
+        } else {
+            vm.ctx
+                .new_tuple(vec![args])
+                .downcast::<PyTuple>()
+                .ok()
+                .unwrap()
```
youknowone, line 57:
```suggestion
            PyTupleRef::with_elements(vec![args], &vm.ctx)
```

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,166 @@
...
+        };
+
+        Self {
+            origin,
+            args: args.clone(),
+            parameters: make_parameters(args, vm),
+        }
```
youknowone, line 64: because `make_parameters` doesn't consume args inside, `args.clone()` is redundant.

```suggestion
        let parameters = make_parameters(&args, vm);
        Self {
            origin,
            args,
            parameters,
        }
```

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,166 @@
...
+fn is_typevar(obj: PyObjectRef) -> bool {
+    let class = obj.class();
+    class.tp_name() == "TypeVar"
+        && match class.get_attr("__module__") {
+            Some(o) => o
+                .downcast_ref::<PyStr>()
+                .map(|s| s.as_str() == "typing")
+                .unwrap_or(false),
+            None => false,
+        }
```
youknowone, line 132:
```suggestion
        && class.get_attr("__module__").and_then(|o| {
            o.downcast_ref::<PyStr>().map(|s| s.as_str() == "typing")
        }).unwrap_or(false)
```

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,166 @@
...
+}
+
+fn make_parameters(args: PyTupleRef, vm: &VirtualMachine) -> PyTupleRef {
```
youknowone, line 135:
```suggestion
fn make_parameters(args: &PyTupleRef, vm: &VirtualMachine) -> PyTupleRef {
```

Because it doesn't use any tuple feature, taking slice is also an option

```suggestion
fn make_parameters(args: &[PyObjectRef], vm: &VirtualMachine) -> PyTupleRef {
```

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,166 @@
...
+                parameters.push(subparam.clone());
+            }
+        }
+    }
+
+    vm.ctx
+        .new_tuple(parameters)
+        .downcast::<PyTuple>()
+        .ok()
+        .unwrap()
```
youknowone, line 154:
```suggestion
    PyTupleRef::with_elements(parameters, &vm.ctx)
```

vm/src/builtins/genericalias.rs
```diff
@@ -0,0 +1,166 @@
...
+use std::fmt;
+
+#[pyclass(module = "types", name = "GenericAlias")]
+pub struct PyGenericAlias {
+    origin: PyTypeRef,
+    args: PyTupleRef,
+    parameters: PyTupleRef,
+}
+
+pub type PyGenericAliasRef = PyRef<PyGenericAlias>;
```
youknowone, line 17: I don't think we will have many usage of this dedicated ref type. How about just using `PyRef<PyGenericAlias>`?

youknowone (changes requested):

vm/src/builtins/genericalias.rs
moreal: I don't have an opinion about it but I just tried to create a type like the other type does. I will remove the type `PyGenericAliasRef` as you suggested. 👌🏻

youknowone (approved):

PR #3087, opened 2021-09-18
CI for Apple silicon (aarch64-apple-darwin)

youknowone:


DimitrisJim (approved):
This looks good too (tracked in rust-lang/rust/issues/73908 so we should keep an eye on that).

ShadowJonathan:
I know this is a random PR in my notifications, but y'all helped me learn the fact that apple silicon support is in GHA now, thanks :eyes:

PR #3088, opened 2021-09-18
os: fix stat_result test

deantvv:
Implement tp_new for `stat_result`

To unpack args, this PR implement a `flatten_args` closure to
unpack args like
```
args = (1, 2, 3, 4, 5, 6)
args = ((1, 2, 3, 4, 5, 6))
args = (((1, 2, 3, 4, 5, 6)))  # from pickle load
```

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -9,6 +9,7 @@ use std::{env, fs};
 
 use crate::crt_fd::Fd;
 use crossbeam_utils::atomic::AtomicCell;
+use itertools::Itertools;
 use num_bigint::BigInt;
 #[cfg(unix)]
 use strum_macros::EnumString;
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/os.rs
```diff
@@ -1102,6 +1118,34 @@ mod _os {
...
         }
+
+        #[pyslot]
+        fn tp_new(_cls: PyTypeRef, args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+            let flatten_args = |r: &[PyObjectRef]| {
+                let mut vec_args = Vec::from(r);
+                loop {
+                    if vec_args.len() == 1 {
+                        // SAFETY: len is check in previous line
+                        let obj = vec_args.first().unwrap();
```
youknowone, line 1129:
```suggestion
                    if let Some(obj) = vec_args.iter().exactly_one() {
```

deantvv: @youknowone Thanks for the suggestion! Will change this as suggested.

youknowone (approved):

youknowone (approved):

PR #3091, opened 2021-09-19
CI for x86 32bit (i686-unknown-linux-gnu)

youknowone:


.github/workflows/ci.yaml
```diff
@@ -50,10 +50,25 @@ jobs:
...
+    name: Ensure compilation on various targets
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+
+      - uses: actions-rs/toolchain@v1
+        with:
+          target: i686-unknown-linux-gnu
+
+      - uses: Swatinem/rust-cache@v1
```
DimitrisJim, line 62: I see the same action being used further down, shouldn't only one exist? (Preferably the one on this line since we need the cache to be present before we run `check`).

Looking[ at the run ](https://github.com/RustPython/RustPython/blob/main/.github/workflows/ci.yaml#L40)for `rust_tests` it _seems_ like one time should be enough.

youknowone: uh, this is weird. I read the project's document: https://github.com/Swatinem/rust-cache#example-usage
So I thought they will cache proper items by `action-rs/toolchain`. But the cache key prefix is `v0-rust-exotic_targets-1.55.0-x86_64-unknown-linux-gnu`. I think cache is not working for cross targets. I will remove them.

youknowone: https://github.com/Swatinem/rust-cache/issues/33

DimitrisJim: Gotcha. It looks fine to me for now. We can add the caching at a later point.

DimitrisJim (approved):

PR #3092, opened 2021-09-19
Fix PyStr operations to be safe

youknowone:
ref:
https://github.com/RustPython/RustPython/pull/3066#issuecomment-922493702

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -77,6 +77,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "23b62fc65de8e4e7f52534fb52b0f3ed04746ae267519eef2a83941e8085068b"
 
 [[package]]
+name = "ascii"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbf56136a5198c7b01a49e3afcbef6cf84597273d298f54432926024107b0109"
+
+[[package]]
 name = "ascii-canvas"
 version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1894,6 +1900,7 @@ version = "0.1.2"
 dependencies = [
  "adler32",
  "ahash",
+ "ascii",
  "atty",
  "base64",
  "bitflags",
--- a/benches/microbenchmarks.rs
+++ b/benches/microbenchmarks.rs
@@ -132,7 +132,11 @@ fn bench_rustpy_code(group: &mut BenchmarkGroup<WallTime>, bench: &MicroBenchmar
             if let Some(idx) = iterations {
                 scope
                     .locals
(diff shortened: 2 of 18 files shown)


coolreader18 (approved):
LGTM. Another option for `new_ascii_str` is to depend on https://docs.rs/ascii and take a `&ascii::AsciiStr`, and maybe make some macro that takes a string literal and ensures it's ascii. Or maybe the `assert!(s.is_ascii())` part of `new_ascii_str` could be inlined and resolve by llvm.

youknowone:
I looked in `ascii::AsciiStr` but I couldn't find a compact way to support both `&[u8]` and `&AsciiStr` as `new_ascii_str` input. Instead of it, I renamed `new_ascii_str` to `new_ascii_literal` and changed it to take `&'static [u8]` to inlining it easily can optimize the assertion.

coolreader18:
> I couldn't find a compact way to support both &[u8] and &AsciiStr as new_ascii_str input
I think the key is that we *don't* want to take an arbitrary `&[u8]` - typestate programming and all that, `new_ascii_str` should take an actual, for-sure ascii string so that we don't get any invalid data. For literals we can either do a macro or `.as_ascii_str().unwrap()` at the callsite, which would also get inlined.

coolreader18:
Actually, it wouldn't even need to be a proc macro, it can just use some const fn stuff:
```rust
const fn slice_is_ascii(x: &[u8]) -> bool {
    let mut i = 0;
    while i < x.len() {
        if !x[i].is_ascii() {
            return false
        }
        i += 1;
    }
    true
}

macro_rules! ascii {
    ($x:literal) => {{
        const _: () = {
            ["not ascii"][!slice_is_ascii($x) as usize];
        };
        $x // unsafe { AsciiStr::from_ascii_unchecked($x) }
    }}
}

pub fn foo() -> &'static [u8] {
    ascii!(b"\xff")
}
```
```
error: any use of this value will cause an error
  --> src/lib.rs:15:13
   |
14 | /         const _: () = {
15 | |             ["not ascii"][!slice_is_ascii($x) as usize];
   | |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ index out of bounds: the length is 1 but the index is 1
16 | |         };
   | |__________-
...
22 |       ascii!(b"\xff")
   |       --------------- in this macro invocation
   |
```

youknowone:
`new_ascii_literal` is only used for ascii literals. Do you think we need to use `vm.new_ascii_literal(ascii!("something"))` instead of `vm.new_ascii_literal("something")`?

Probably we can make `vm.new_str<T: IntoPyStr>(s: T)` with `impl AsciiString for IntoPyStr` and `impl String for IntoPyStr` later for more convenient `new_str`, but they will be more useful for variables rather than literals.

coolreader18:
> `vm.new_ascii_literal(ascii!("something"))`

Yeah, I think this is the best option.

PR #3094, opened 2021-09-19
General logic for positional iterator

qingshi163:
This should also garentee the multi-thread iteration as long as the iter object is not changed.

### DIFF
--- a/vm/src/builtins/iter.rs
+++ b/vm/src/builtins/iter.rs
@@ -10,7 +10,7 @@ use crate::{
     ItemProtocol, PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TypeProtocol,
     VirtualMachine,
 };
-use rustpython_common::lock::{PyRwLock, PyRwLockUpgradableReadGuard};
+use rustpython_common::lock::{PyMutex, PyRwLock, PyRwLockUpgradableReadGuard};
 
 /// Marks status of iterator.
 #[derive(Debug, Clone)]
@@ -170,7 +170,7 @@ impl<T> PositionIterInternal<T> {
 #[pyclass(module = false, name = "iterator")]
 #[derive(Debug)]
 pub struct PySequenceIterator {
-    internal: PyRwLock<PositionIterInternal<PyObjectRef>>,
+    internal: PyMutex<PositionIterInternal<PyObjectRef>>,
 }
 
 impl PyValue for PySequenceIterator {
@@ -183,13 +183,13 @@ impl PyValue for PySequenceIterator {
 impl PySequenceIterator {
     pub fn new(obj: PyObjectRef) -> Self {
         Self {
-            internal: PyRwLock::new(PositionIterInternal::new(obj, 0)),
+            internal: PyMutex::new(PositionIterInternal::new(obj, 0)),
         }
     }
 
     #[pymethod(magic)]
     fn length_hint(&self, vm: &VirtualMachine) -> PyObjectRef {
-        let internal = self.internal.read();
+        let internal = self.internal.lock();
         if let IterStatus::Active(obj) = &internal.status {
             vm.obj_len(obj)
                 .map(|x| PyInt::from(x).into_object(vm))
@@ -201,12 +201,12 @@ impl PySequenceIterator {
 
     #[pymethod(magic)]
     fn reduce(&self, vm: &VirtualMachine) -> PyObjectRef {
-        self.internal.read().builtin_iter_reduce(|x| x.clone(), vm)
+        self.internal.lock().builtin_iter_reduce(|x| x.clone(), vm)
     }
 
     #[pymethod(magic)]
     fn setstate(&self, state: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
-        self.internal.write().set_state(state, vm)
+        self.internal.lock().set_state(state, vm)
     }
 }
 
@@ -214,7 +214,7 @@ impl IteratorIterable for PySequenceIterator {}
 impl SlotIterator for PySequenceIterator {
     fn next(zelf: &PyRef<Self>, vm: &VirtualMachine) -> PyResult {
         zelf.internal
-            .write()
+            .lock()
             .next(|obj, pos| obj.get_item(pos, vm), vm)
     }
 }
--- a/vm/src/builtins/iter.rs
+++ b/vm/src/builtins/iter.rs
@@ -218,7 +218,7 @@ impl SlotIterator for PySequenceIterator {
 #[derive(Debug)]
 pub struct PyCallableIterator {
     sentinel: PyObjectRef,
-    status: PyRwLock<IterStatus<PyCallable>>,
+    status: PyRwLock<IterStatus<ArgCallable>>,
 }
 
 impl PyValue for PyCallableIterator {
--- a/vm/src/builtins/iter.rs
+++ b/vm/src/builtins/iter.rs
@@ -12,30 +12,34 @@ use crate::{
 };
 use crossbeam_utils::atomic::AtomicCell;
 
+/// Marks status of iterator.
+#[derive(Debug, Clone)]
+pub enum IterStatus {
+    /// Iterator hasn't raised StopIteration.
+    Active(PyObjectRef),
+    /// Iterator has raised StopIteration.
+    Exhausted,
+}
+
 #[derive(Debug)]
 pub struct PositionIterInternal {
-    pub position: AtomicCell<usize>,
-    /// object or PyNone if exhausted
-    pub obj: PyRwLock<PyObjectRef>,
+    pub status: IterStatus,
+    pub position: usize,
 }
 
 impl PositionIterInternal {
-    pub fn new(obj: PyObjectRef) -> Self {
+    pub fn new(obj: PyObjectRef, position: usize) -> Self {
         Self {
-            position: AtomicCell::new(0),
-            obj: PyRwLock::new(obj),
+            status: IterStatus::Active(obj),
+            position,
         }
     }
 
-    pub fn is_active(&self, vm: &VirtualMachine) -> bool {
-        !vm.is_none(&self.obj.read())
-    }
-
-    pub fn set_state(&self, state: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
-        if self.is_active(vm) {
+    pub fn set_state(&mut self, state: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        if let IterStatus::Active(_) = &self.status {
             if let Some(i) = state.payload::<PyInt>() {
                 let i = int::try_to_primitive(i.as_bigint(), vm).unwrap_or(0);
-                self.position.store(i);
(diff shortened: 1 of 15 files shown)


vm/src/builtins/iter.rs
```diff
@@ -13,6 +14,88 @@ use crate::{
     TypeProtocol,
 };
 
+#[derive(Debug)]
+pub struct PositionIterInternal {
+    pub position: AtomicCell<usize>,
+    /// object or PyNone if exhausted
+    pub obj: PyRwLock<PyObjectRef>,
```
qingshi163, line 21: Replace with PyRwLock<Option<PyObjectRef>> or this is fine?

coolreader18: I think it would be beneficial for it to be generic over `<T: PyValue>` so that you don't need to downcast, and then it would make sense for it to be Option

vm/src/builtins/iter.rs
```diff
@@ -13,6 +14,91 @@ use crate::{
     TypeProtocol,
 };
 
+#[derive(Debug)]
+pub struct PositionIterInternal {
+    pub position: AtomicCell<usize>,
```
coolreader18, line 19: This should be `PyAtomic<usize>`, I think there's some examples of its use if you grep for it.

vm/src/builtins/iter.rs
coolreader18: Or maybe something like `AsRef<PyObjectRef>` to allow `PyRef<T>` or `PyObjectRef`. I think that *would* require a couple AsRef impls though.

vm/src/builtins/iter.rs
```diff
@@ -13,6 +14,91 @@ use crate::{
...
+
+impl PositionIterInternal {
+    pub fn new(obj: PyObjectRef) -> Self {
+        Self {
+            position: AtomicCell::new(0),
+            obj: PyRwLock::new(obj),
+        }
+    }
+
+    pub fn is_active(&self, vm: &VirtualMachine) -> bool {
```
DimitrisJim, line 32: I'm pretty sure that for non-lto builds these will all benefit from being marked as inline but I'd like to time that real quick.

vm/src/builtins/iter.rs
```diff
@@ -13,6 +14,91 @@ use crate::{
...
+        !vm.is_none(&self.obj.read())
+    }
+
+    pub fn set_state(&self, state: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        if self.is_active(vm) {
+            if let Some(i) = state.payload::<PyInt>() {
+                let i = int::try_to_primitive(i.as_bigint(), vm).unwrap_or(0);
```
DimitrisJim, line 39: Constrict this in `[0, len(obj)]` to follow with CPython. I don't think this leads to bad behavior in any of the other methods, it's only a visible difference when you directly call `__reduce__` from an iterator and see its output.

vm/src/builtins/iter.rs
```diff
@@ -13,6 +14,91 @@ use crate::{
...
+    pub fn reduce(&self, func: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
+        if self.is_active(vm) {
+            vm.ctx.new_tuple(vec![
+                func,
+                vm.ctx.new_tuple(vec![self.obj.read().clone()]),
+                vm.ctx.new_int(self.position.load()),
+            ])
+        } else {
+            vm.ctx
+                .new_tuple(vec![func, vm.ctx.new_tuple(vec![vm.ctx.new_list(vec![])])])
```
DimitrisJim, line 59: What `__reduce__` returns here is a weird situation. `deque` is very opinionated about always returning a `deque` instance (and the deque iter type accepts nothing else) while others (types edited in this PR) are more open about this (since `iter` doesn't really care what its initialized with).

I think an empty tuple would be the more sensible default for now, even though `list.__reduce__` returns an empty list in CPython.

qingshi163: I had check that in CPython, list, set and dict are using empty list for the bulitin iter to construct the empty iterator. As you said, I also think this is not matter, because an exhursted iter should do nothing.

youknowone:
Can't this be achieved by 2 AtomicCells like before? Was there problems?

qingshi163:
First, CPython release the reference of the object after the iterator been exhausted in the most case(but not garentee).
Second, the two atomic looks like not safe in multi-threads iteration, one could setting the iterator to exhaust while the others already get the position, so the index will be over called.

Maybe I am misunderstanding, better way to do or we can ignore that conflict?

youknowone (changes requested):
looks good in general. This work will be a good base for actual sequence protocol implementation. Would you also check Jim's comment here? https://github.com/RustPython/RustPython/pull/3094#discussion_r712867591

youknowone:
cc @eldpswp99 for dict iterator changes
cc @Snowapril who was interested in sequence protocol

vm/src/builtins/iter.rs
```diff
@@ -9,23 +9,206 @@ use crate::{
...
+        if let IterStatus::Active(obj) = &self.status {
+            if self.position <= f(obj) {
+                return self.position + 1;
+            }
+        }
+        0
+    }
+}
+
+pub fn get_builtin_attribute_iter(vm: &VirtualMachine) -> &PyObjectRef {
```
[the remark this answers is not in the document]
youknowone: Another suggestion. Instead of this static cache, we can consider put this in `PyContext`, where we currently typically store these values.

DimitrisJim:
> Second, the two atomic looks like not safe in multi-threads iteration

Yeah I've been thinking about that too and reading about memory orderings is a real pain, especially when dealing with the interaction of two atomic variables. I'm pretty certain these should work, definitely with `SeqCst` but, even with `AcqRel` as `AtomicCell` currently uses. As far as I understand the [memory orders](https://en.cppreference.com/w/cpp/atomic/memory_order) `AcqRel` does tell us this:

> All writes in other threads that release the same atomic variable are visible before the modification and the modification is visible in other threads that acquire the same atomic variable. 

I'm still not 100% comfortable with memory sequencing but, as I read it, I get the impression that writing `Exhausted` to `status` from one thread should be visible to another when it `load`s it. Even without having that guarantee, things probably work right now because we (in most cases?) `.get` on the underlying collection and, in effect, only allow the last thread with a valid index to return an item (all the rest would raise `StopIteration`).

I'm not too comfortable yet with atomics and the orderings they imply to say we should keep them. Going with a `Mutex` for now, which is easy to reason about, and trying to go lock-free in the future seems like a sensible idea.

vm/src/builtins/iter.rs
qingshi163: I had add another function set_state_saturated() for the classes that should saturate the value.

vm/src/builtins/iter.rs
```diff
@@ -9,23 +9,206 @@ use crate::{
...
+                vm.ctx.new_tuple(vec![f(obj)]),
+                vm.ctx.new_int(self.position),
+            ])
+        } else {
+            vm.ctx
+                .new_tuple(vec![func, vm.ctx.new_tuple(vec![vm.ctx.new_list(vec![])])])
+        }
+    }
+
+    pub fn builtin_iter_reduce<F>(&self, f: F, vm: &VirtualMachine) -> PyObjectRef
```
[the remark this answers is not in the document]
qingshi163: Sure, like deque is using the own iterator. PositionIterInternal works like the helper class, you can ignore the methods that not suit. also like str and dict have their own implement of the next() slot.

vm/src/builtins/iter.rs
DimitrisJim: Ah I see. Isn't `set_state` _only_ used in the generic sequence iterator? Couldn't we just remove `set_state_saturated`, use the callable arg in `set_state` and pass `|_, pos| pos` for `PySequenceIterator` and `|obj, pos| pos.min(obj.len())` for the others?

vm/src/builtins/iter.rs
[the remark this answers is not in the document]
qingshi163: As we need to lookup the attribute from builtins, so we need vm in the current implement, so I do not think it suit to put under PyContext.

qingshi163:
> > Second, the two atomic looks like not safe in multi-threads iteration
> 
> Yeah I've been thinking about that too and reading about memory orderings is a real pain, especially when dealing with the interaction of two atomic variables. I'm pretty certain these should work, definitely with `SeqCst` but, even with `AcqRel` as `AtomicCell` currently uses. As far as I understand the [memory orders](https://en.cppreference.com/w/cpp/atomic/memory_order) `AcqRel` does tell us this:
> 
> > All writes in other threads that release the same atomic variable are visible before the modification and the modification is visible in other threads that acquire the same atomic variable.
> 
> I'm still not 100% comfortable with memory sequencing but, as I read it, I get the impression that writing `Exhausted` to `status` from one thread should be visible to another when it `load`s it. Even without having that guarantee, things probably work right now because we (in most cases?) `.get` on the underlying collection and, in effect, only allow the last thread with a valid index to return an item (all the rest would raise `StopIteration`).
> 
> I'm not too comfortable yet with atomics and the orderings they imply to say we should keep them. Going with a `Mutex` for now, which is easy to reason about, and trying to go lock-free in the future seems like a sensible idea.

As my understanding, the memory order is not going to save us when we using two atomic values.

I also make a test as:
```
import threading

class C1:
    def __init__(self) -> None:
        self.exhausted = False
    def __getitem__(self, pos):
        assert not self.exhausted
        if pos == 1000:
            self.exhausted = True
            raise IndexError()
        return 1


for _ in range(100):
    c = C1()
    rets = {}
    i = iter(c)
    def retrieve(n):
        s = 0
        for x in i:
            s += x
        rets[n] = s
    threads = [threading.Thread(target=retrieve, args=(n,)) for n in range(3)]

    for t in threads:
        t.start()
    for t in threads:
        t.join()

    print(rets)
```

The two atomic version is failing on my machine.

Also if we do not have this garenteed, we will have the problem even for the internal classes like bytearray and list, etc. because the python code maybe will catch the stop iteration exception than doing something tricky to the object like append more values.
but this is rare, I don't know should we limit the multi-thread iteration safe only if the object is not been modified, otherwise it is difficult to rid the lock.

DimitrisJim:
> The two atomic version is failing on my machine. 

not at all surprised, reasoning about atomics is a tricky thing, thanks for testing it out. 

> otherwise it is difficult to rid the lock.

Yup, I think this will be challenging, I'd be interested to see how people trying to remove the GIL from CPython actually tried to do this (the most recent attempt was a couple of years ago, i.e the Gilectomy). Pretty sure they were using atomics in many places, not sure if they did so with the iterators.

A Mutex definitely seems like the best way forward for now.

youknowone:
@qingshi163 could you rebase this onto main? I think then the failing CI will be gone

youknowone (approved):

youknowone:
all of the failures are about `test_cmd_line_script`

youknowone:
Thank you! it was a huge work!

PR #3098, opened 2021-09-20
CPython docs database for module items

youknowone:
no more copy & paste


extra_tests/docs_gen.py
```diff
@@ -0,0 +1,101 @@
+import sys
+import warnings
+from pydoc import ModuleScanner
+
+
+def scan_modules():
```
DimitrisJim, line 6: Since this only concerns the builtins, why not use `sys.builtin_module_names` here?

extra_tests/docs_gen.py
```diff
@@ -0,0 +1,101 @@
...
+
+def is_child(module, item):
+    import inspect
+    item_mod = inspect.getmodule(item)
+    return item_mod is module
+
+
+def traverse(module, names, item):
+    import inspect
+    is_builtin = inspect.ismodule(item) or inspect.isbuiltin(item)
```
DimitrisJim, line 52: `inspect.isbuiltin` only deals with `methods/functions`. This won't return docstrings for classes, unfortunately. Probably a `has_doc` variable that checks if its also a class would be better.

extra_tests/docs_gen.py
youknowone: I found `sys.bulitin_module_names` is not covering all the c modules. I didn't investigate well but I think it is more like pre-loaded module before import.

DimitrisJim: huh, maybe that in combination with `sys.stdlib_module_names` then (but that landed in `3.10` I believe.

derive/src/pyclass.rs
```diff
@@ -199,14 +206,14 @@ fn generate_class_def(
...
+        Some(quote! {
             fn static_metaclass() -> &'static ::rustpython_vm::builtins::PyTypeRef {
                 use rustpython_vm::StaticType;
                 #typ::static_type()
             }
-        }
+        })
     } else {
-        quote!()
+        None
```
coolreader18, line 216: I guess this could just become a `metaclass.map()` now.

DimitrisJim (approved):
lgtm. module and method docs aren't currently attached as they are for functions/classes but those could be a separate PR, you can choose if you want to merge this now or add them on.

PR #3099, opened 2021-09-21
impl `#[derive(PyValue)]` for static_type

deantvv:
This seems a big change but actually it is a simple macro that will produce the following implementation

```rust
impl ::rustpython_vm::PyValue for #ty {
    fn class(_vm: &VirtualMachine) -> &PyTypeRef {
        Self::static_type()
    }
}
```

### DIFF
--- a/wasm/lib/src/js_module.rs
+++ b/wasm/lib/src/js_module.rs
@@ -51,17 +51,11 @@ extern "C" {
 }
 
 #[pyclass(module = "_js", name = "JSValue")]
-#[derive(Debug)]
+#[derive(Debug, PyValue)]
 pub struct PyJsValue {
     pub(crate) value: JsValue,
 }
 type PyJsValueRef = PyRef<PyJsValue>;
-
-impl PyValue for PyJsValue {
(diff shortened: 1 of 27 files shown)


youknowone (approved):

wasm/lib/src/js_module.rs
```diff
@@ -386,20 +375,14 @@ pub struct PyPromise {
 }
 pub type PyPromiseRef = PyRef<PyPromise>;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PyValue)]
```
youknowone, line 378: PyValue is expected to be moved to PyPromise but to PromiseKind

deantvv: Yes. I will fix in new commit.
Thanks!

fanninpm:
@youknowone can the [relevant section in the wiki](https://github.com/RustPython/RustPython/wiki/Python-Attributes#impl-pyvalue-for-type) be updated to reflect this?

DimitrisJim:
done!

PR #3103, opened 2021-09-21
Fixed Nan issue

Codemonk-adi:


Codemonk-adi:
@DimitrisJim  I redid the commit cause I did something wrong and cmath.rs got deleted, it should be fine now

DimitrisJim (approved):

PR #3050, opened 2021-09-13
Run cargo update + upgrade

coolreader18:
`cargo upgrade` changeset:
<pre><font color="#55FF55"><b>    Updating</b></font> crates.io index
<font color="#55FF55"><b>    Updating</b></font> approx v0.4.0 -&gt; v0.5.0
<font color="#FF5555"><b>    Removing</b></font> arrayvec v0.5.2
<font color="#FF5555"><b>    Removing</b></font> build_const v0.2.2
<font color="#FF5555"><b>    Removing</b></font> byteorder v1.4.3
<font color="#55FF55"><b>    Updating</b></font> cranelift v0.72.0 -&gt; v0.76.0
<font color="#55FF55"><b>    Updating</b></font> cranelift-bforest v0.72.0 -&gt; v0.76.0
<font color="#55FF55"><b>    Updating</b></font> cranelift-codegen v0.72.0 -&gt; v0.76.0
<font color="#55FF55"><b>    Updating</b></font> cranelift-codegen-meta v0.72.0 -&gt; v0.76.0
<font color="#55FF55"><b>    Updating</b></font> cranelift-codegen-shared v0.72.0 -&gt; v0.76.0
<font color="#55FF55"><b>    Updating</b></font> cranelift-entity v0.72.0 -&gt; v0.76.0
<font color="#55FF55"><b>    Updating</b></font> cranelift-frontend v0.72.0 -&gt; v0.76.0
<font color="#55FF55"><b>    Updating</b></font> cranelift-jit v0.72.0 -&gt; v0.76.0
<font color="#55FF55"><b>    Updating</b></font> cranelift-module v0.72.0 -&gt; v0.76.0
<font color="#55FF55"><b>    Updating</b></font> cranelift-native v0.72.0 -&gt; v0.76.0
<font color="#FF5555"><b>    Removing</b></font> crc v1.8.1
<font color="#55FF55"><b>    Updating</b></font> env_logger v0.8.4 -&gt; v0.9.0
<font color="#FF5555"><b>    Removing</b></font> errno v0.2.8
<font color="#FF5555"><b>    Removing</b></font> errno-dragonfly v0.1.2
<font color="#55FF55"><b>    Updating</b></font> hexf-parse v0.1.0 -&gt; v0.2.1
<font color="#FF5555"><b>    Removing</b></font> lexical-core v0.7.6
<font color="#55FFFF"><b>      Adding</b></font> lexical-parse-float v0.8.2
[...]

PR #3105, opened 2021-09-21
extract wchar_t definition to rustpython-common

youknowone:
split from #3104 

not to struct/array modules refer each other


common/Cargo.toml
```diff
@@ -21,3 +21,4 @@ siphasher = "0.3"
 rand = "0.8"
 volatile = "0.3"
 radium = "0.6"
+libc = "0.2.101"
```
fanninpm, line 24: The latest version of `libc` is `0.2.102` as of when I'm writing this comment. (See also #3050.)

coolreader18 (approved):

PR #3108, opened 2021-09-21
Rename Py{Callable, Iterable} to Arg{..}

youknowone:
which were not a python object but just argument helper

### DIFF
--- a/src/shell/helper.rs
+++ b/src/shell/helper.rs
@@ -1,6 +1,6 @@
 use rustpython_vm::builtins::{PyDictRef, PyStrRef};
 use rustpython_vm::VirtualMachine;
-use rustpython_vm::{ArgIterable, PyResult, TryFromObject};
+use rustpython_vm::{function::ArgIterable, PyResult, TryFromObject};
 
 pub struct ShellHelper<'vm> {
     vm: &'vm VirtualMachine,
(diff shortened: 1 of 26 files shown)


DimitrisJim:
+1 but naming convention seems to better favor `CallableArg` and `IterableArg` imho. We already have `PosArg` and `KwArg` so following that convention makes sense (it also reads better, with `Callable/Iterable/Pos/Kw` acting as qualifiers for the `Arg`.

youknowone:
Hmm, I didn't thought about that way but it also sounds reasonable.

In this PR, I simply followed `ArgBytesLike`, `ArgMemoryBuffer` and `ArgStrOrBytesLike` which I renamed from `Py*` in other patch.
When I wrote the previous patch, I preferred prefix way because they give better sort for `use` statements.
I didn't think about args/kwargs at that time because I always regarded them as python special terms. So this is a new point of view.

I don't have strong opinion for prefix/suffix. But if we want to change them to suffix, I want to to it in separate job because byteslike types also need to be changed.

DimitrisJim:
> I don't have strong opinion for prefix/suffix. But if we want to change them to suffix, I want to to it in separate job because byteslike types also need to be changed.

Yup, agreed. I think a common pattern for naming these is warranted, would make the intent of the types much more clear.

DimitrisJim (approved):

PR #3110, opened 2021-09-22
Group use for vm stuff to prevent reorder for rustpython_stdlib

youknowone:


youknowone:
I am going to merge this PR without review because this is basically just regroupig use statements and doesn't change neither compiled code nor the meaning of code

PR #3111, opened 2021-09-22
split `posix` and `nt` files from `os`

youknowone:


youknowone:
this PR is mostly just relocating codes

PR #3112, opened 2021-09-22
move IntoPy{Object, Exception, Result} to vm::function

youknowone:
which are used to define function arguments - those functions are mostly placed in the crate


DimitrisJim (approved):
looks okay to me.

PR #3113, opened 2021-09-22
Solve clippy warnings in window

Snowapril:
Modify some codes as clippy advised.

### DIFF
--- a/vm/src/stdlib/multiprocessing.rs
+++ b/vm/src/stdlib/multiprocessing.rs
@@ -18,7 +18,7 @@ mod _multiprocessing {
 
     #[pyfunction]
     fn recv(socket: usize, size: usize, vm: &VirtualMachine) -> PyResult<libc::c_int> {
-        let mut buf = vec![0 as libc::c_char; size];
+        let mut buf = vec![0; size];
         let nread =
             unsafe { winsock2::recv(socket as SOCKET, buf.as_mut_ptr() as *mut _, size as i32, 0) };
         if nread < 0 {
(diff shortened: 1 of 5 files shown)


vm/src/stdlib/multiprocessing.rs
```diff
@@ -18,7 +18,7 @@ mod _multiprocessing {
 
     #[pyfunction]
     fn recv(socket: usize, size: usize, vm: &VirtualMachine) -> PyResult<libc::c_int> {
-        let mut buf = vec![0 as libc::c_char; size];
+        let mut buf = vec![0_i8; size];
```
Snowapril, line 21: As `libc::c_char` is always `i8` in all os, this can resolve clippy warning

youknowone: No, it doesn't. `libc` is following C type definitions. And C doesn't specified char type as signed or unsigned. So it can be either i8 or u8.

Snowapril: Oh you are right. I found both `i8` and `u8` exists in [`libc documents`](https://rust-lang.github.io/libc/aarch64-linux-android/doc/libc/type.c_char.html).
Then how could we resolve this clippy warning? should we add `#[allow(clippy::unnecessary_cast)]` in line 21?

youknowone: I think just `0` without casting is enough here. The type will be inferenced by `winsock2::recv` in line 23.
In the same reason, I think `as *mut _` in line 23 is also redundant.

Snowapril: Good! kinda simple solution 😊
I applied it

youknowone (approved):

PR #3116, opened 2021-09-22
Universal mechanism for Iterator slots

youknowone:
Adaptable for any kind of iterable/iterator with new machanism.
So now `io::_IOBase` also can implement `Iterable` and `PyIter`

This is the same strategy with `Hashable` and `Unhashable`


coolreader18 (approved):
LGTM

PR #3117, opened 2021-09-22
PyIter protocol object

youknowone:
~based on #3115 and #3116~

will be part of #3104 because `PyIter` is required for stdlib modules

### DIFF
--- a/vm/src/iterator.rs
+++ b/vm/src/iterator.rs
@@ -2,65 +2,24 @@
  * utilities to support iteration.
  */
 
-use crate::builtins::int::{self, PyInt};
-use crate::builtins::iter::PySequenceIterator;
-use crate::builtins::PyBaseExceptionRef;
-use crate::vm::VirtualMachine;
-use crate::{IdProtocol, PyObjectRef, PyResult, PyValue, TypeProtocol};
+use crate::{
+    builtins::{int, PyBaseExceptionRef, PyInt},
+    protocol::PyIter,
+    IdProtocol, PyObjectRef, PyResult, TypeProtocol, VirtualMachine,
+};
 use num_traits::Signed;
-
-/*
- * This helper function is called at multiple places. First, it is called
- * in the vm when a for loop is entered. Next, it is used when the builtin
- * function 'iter' is called.
- */
-pub fn get_iter(vm: &VirtualMachine, iter_target: PyObjectRef) -> PyResult {
-    let getiter = {
-        let cls = iter_target.class();
-        cls.mro_find_map(|x| x.slots.iter.load())
-    };
-    if let Some(getiter) = getiter {
-        let iter = getiter(iter_target, vm)?;
-        let cls = iter.class();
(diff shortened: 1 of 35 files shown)


youknowone:
@DimitrisJim could you review this PR?

vm/src/iterator.rs
```diff
@@ -2,65 +2,24 @@
...
-            })?
-    };
-    iternext(iter_obj, vm)
-}
-
 /*
  * Helper function to retrieve the next object (or none) from an iterator.
  */
-pub fn get_next_object(
+pub fn get_next_object<T>(
```
DimitrisJim, line 15: Is my understanding correct that the remaining functions in `iterator` will (eventually) not be part of the public API? If so, is there a reason keeping them here instead of having all iterator related functions in `protoco/iter.rs`?

youknowone: I think these functions will be part of public API. The major reason I didn't do it in this PR is to confirm the design by review before going further. I will do more work in future with new PRs.

In CPython, The public API `PyIter_Next` returns (if we can say so) `PyResult<Option<PyObjectRef>>` as its primitive. But in RustPython, `PyIter::next` returns `PyResult<PyObejctRef>` as its primitive. So there will be more decision works not just about `PyIter` itself.

This is related to #3124

vm/src/stdlib/operator.rs
```diff
@@ -217,9 +218,9 @@ mod _operator {
 
     /// Return the number of occurrences of b in a.
     #[pyfunction(name = "countOf")]
-    fn count_of(a: PyObjectRef, b: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
+    fn count_of(a: PyIter, b: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
         let mut count: usize = 0;
-        let iter = iterator::get_iter(vm, a)?;
+        let iter = a;
```
DimitrisJim, line 223: seems (here and in `index_of`) like these aren't necessary anymore (the type gives enough info on what it is already)

vm/src/protocol/iter.rs
```diff
@@ -0,0 +1,118 @@
...
+where
+    T: Borrow<PyObjectRef>;
+
+impl PyIter<PyObjectRef> {
+    pub fn into_object(self) -> PyObjectRef {
+        self.0
+    }
+}
+
+impl<T> PyIter<T>
```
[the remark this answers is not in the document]
youknowone: I honestly have no idea for Check for now. In CPython, it simply check `tp_iternext`. In RustPython, it checks something more than `tp_iternext` includign `__getitem__`. I want to investigate more about it later whether it happened because we are not correctly implementing iterator or mapping protocol or current `TryFromObject for PyIter` is implemented wrong and covering something more than PyIter itself.

DimitrisJim (commented):
lgtm

DimitrisJim (approved):

PR #3119, opened 2021-09-22
Refactor math.rs into module.

DimitrisJim:
diff is an unfortunate royal mess. moves stuff into a `pymodule`.

### DIFF
--- a/vm/src/stdlib/math.rs
+++ b/vm/src/stdlib/math.rs
@@ -1,817 +1,811 @@
-/* Math builtin module
- *
- *
- */
+pub(crate) use math::make_module;
 
-use crate::{
-    builtins::{try_bigint_to_f64, try_f64_to_bigint, IntoPyFloat, PyFloatRef, PyInt, PyIntRef},
-    function::{ArgIterable, Args, OptionalArg},
-    utils::Either,
-    PyObjectRef, PyResult, PySequence, TypeProtocol, VirtualMachine,
-};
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/math.rs
```diff
@@ -1,817 +1,820 @@
...
+pub(crate) fn make_module(vm: &VirtualMachine) -> PyObjectRef {
+    let module = math::make_module(vm);
+    let ctx = &vm.ctx;
+    extend_module!(vm, module, {
+        // Constants:
+        "pi" => ctx.new_float(std::f64::consts::PI), // 3.14159...
+        "e" => ctx.new_float(std::f64::consts::E), // 2.71..
+        "tau" => ctx.new_float(2.0 * std::f64::consts::PI),
+        "inf" => ctx.new_float(f64::INFINITY),
+        "nan" => ctx.new_float(f64::NAN)
```
coolreader18, line 12: these can be `#[pyattr] use f64::{FOO as foo};`, I'm pretty sure. Also I know this is old code but there's now a `std::f64::consts::TAU` constant.

DimitrisJim: Ah, that's even better. You can even use multiple rebindings and things work just fine, didn't condense these down to a single though, seemed too much.

youknowone (approved):

youknowone:
https://github.com/RustPython/RustPython/pull/3119/files?w=1

PR #3121, opened 2021-09-23
Implement Mapping Protocol

Snowapril:
This revision implements [`mapping protocol`](https://docs.python.org/3/c-api/mapping.html)

`get_item`, `set_item`, `del_item` in the `ItemProtocol` will be improved by using fast-path which is provided by mapping protocol.

With mapping protocol, this new issue can be fixed with `PyMapping::check`
According to [`docs`](https://docs.python.org/3/library/types.html?highlight=mappingproxy#types.MappingProxyType) and [`cpython implementation`](https://github.com/python/cpython/blob/3.9/Objects/descrobject.c#L1190) mapping proxy type must created only with mapping type.
```python
>>>>> mappingproxy = type(type.__dict__)
>>>>> mappingproxy({"ASCII": 256})
<mappingproxy object at 0x55bb16b37560>
>>>>> mappingproxy([1,2,3])
TypeError: mappingproxy() argument must be a mapping, not list
```

---
#### Documents
* https://docs.python.org/3/c-api/mapping.html
* https://docs.python.org/3/library/types.html?highlight=mappingproxy#types.MappingProxyType


youknowone:
could we merge `AsMapping` and `PyMappingMethods`

youknowone:
check for `update_slot` in `pytype.rs`. it is mandatory for protocols using slots

Snowapril:
I feel a little bit confused with using `update_slot` currently.
with `update_slot` we can update one slot that have exactly same signature with the given new method defined in python-side.
but `as_mapping` have three sub-slot(function) and if `__getitem__` is defined in python-side, I need to update `as_mapping` to return PyMapping struct with only `subscript` field modified. How can I achieve this?..

youknowone:
you need only single function `slots::AsMappingFunc`. Maybe you will get idea from richcompare.

roughly:

```rust
"__getitem__" | "__setitem__" | "__delitem__" => {
    let func: slots::AsMappingFunc = |zelf, vm|  PyMapping {
        get: vm.get_class_attr("__getitem__").unwrap_or(getitem_fallback),
        set: vm....
        del:...
    };
}
```
```rust
fn getitem_fallback(zelf: PyObjectRef, key: PyObjectRef) -> PyResult {
    super().getitem(key)  // something like this
}
```

Snowapril:
Ah! I got it. Thanks 😊

Snowapril:
Now `update_slot` and mappingproxy trait works as normal
```python
>>>>> mappingproxy = type(type.__dict__)
>>>>> mappingproxy({"ASCII": 256})
<mappingproxy object at 0x55bb16b37560>
>>>>> mappingproxy([1,2,3])
TypeError: mappingproxy() argument must be a mapping, not list
```
but update_slot routine for slot which have multiple  sub_slots is somewhat horrible now.
I think there need something like `#[pyslotstruct]` for a struct which have multiple `#[pyslot]`

Another slots in current rustpython has only one function that corresponded to its slot.
But `mapping protocol`, `sequence protocol` and `number protocol` have multiple slots..

youknowone:
looks great!
I think mapping protocol has only one slot `as_mapping` but having multiple python magic methods. (probably you wanted to point this out)
Maybe this is time to check how CPython handle this.

vm/src/slots.rs
```diff
@@ -530,6 +531,50 @@ pub trait AsBuffer: PyValue {
...
 
+#[pyimpl]
+pub trait AsMapping: PyValue {
+    #[pyslot]
+    fn tp_as_mapping(zelf: &PyObjectRef, vm: &VirtualMachine) -> PyResult<PyMapping> {
+        let zelf = zelf
+            .downcast_ref()
+            .ok_or_else(|| vm.new_type_error("unexpected payload for as_mapping".to_owned()))?;
+        Self::as_mapping(zelf, vm)
+    }
```
youknowone, line 542: `as_mapping` looks like not requiring any argument. Maybe this will be ok.
```suggestion
    #[pyslot]
    fn as_mapping() -> PyMapping {
            PyMapping {
                length: Some(Self::length),
                subscript: Some(Self::subscript),
                ass_subscript: Some(Self::ass_subscript),
            }
    }
```

vm/src/stdlib/posix.rs
```diff
@@ -1069,7 +1069,7 @@ pub mod module {
         #[pyarg(positional)]
         args: crate::function::ArgIterable<PyPathLike>,
         #[pyarg(positional)]
-        env: crate::builtins::dict::PyMapping,
+        env: crate::builtins::dict::PyDictRef,
```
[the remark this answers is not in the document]
Snowapril: with only `PyMapping` and its fields, I cannot get its both items and keys together.
As cpython implementation, adding `items`, `keys` and `values` method in `PyMapping` and use `PyObjectRef` for `env` is more appropriate. how do you think?

vm/src/slots.rs
Snowapril: I cannot declare empty parameter function as slot according to below error message 😢 I think function signature for pyslot must be a `fn name(&PyObjectRef, &VirtualMachine)`
```rust
non-primitive cast: `fn() -> std::result::Result<builtins::dict::PyMapping, 
pyobjectrc::PyRef<exceptions::types::PyBaseException>> {<Self as slots::AsMapping>::slot_as_mapping}` as `for<'r, 's> 
fn(&'r pyobjectrc::PyObjectRef, &'s vm::VirtualMachine) -> std::result::Result<builtins::dict::PyMapping, 
pyobjectrc::PyRef<exceptions::types::PyBaseException>>`
```

Although this is little bit dirty, how about this one?
```rust
#[pyslot]
fn slot_as_mapping(_zelf: &PyObjectRef, _vm: &VirtualMachine) -> PyResult<PyMapping> {
    Self::as_mapping()
}
fn as_mapping() -> PyResult<PyMapping>;
```

vm/src/stdlib/posix.rs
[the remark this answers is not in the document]
youknowone: uh, it looks more complicated than I thought before. For current MyMapping Implementation, you are right.

It looks like `PyMapping` itself is a wrapper of `PyObejctRef`, how I handled `PyIter` in #3117. (https://github.com/RustPython/RustPython/pull/3117/files#diff-688a17d340aa55627e0e307b8cef31b99d478c8b41664cd92f5c2572b6688097R13)

And current `PyMapping` looks like `PyMappingMethods` in CPython - probably the way you originally approached.
How do you think about this?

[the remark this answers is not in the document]
Snowapril: I dont know that I understand correctly though, did you mean switching current `PyMapping` to `PyMappingMethods` and creating`PyMapping` as a wrapper of `PyObjectRef` with its related methods right? 
It seems more appropriate than I thought 😊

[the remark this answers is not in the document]
youknowone: Yes, exactly

vm/src/builtins/pytype.rs
```diff
@@ -270,6 +271,54 @@ impl PyType {
...
+                            |zelf, needle, value, vm| match value {
+                                Some(value) => vm
+                                    .call_special_method(zelf, "__setitem__", (needle, value),)
+                                    .map(|_| Ok(()))?,
+                                None => vm
+                                    .call_special_method(zelf, "__delitem__", (needle,))
+                                    .map(|_| Ok(()))?,
+                            }
+                        ),
+                    })
```
youknowone, line 317: maybe we need `_PyObject_NextNotImplemented` later

vm/src/pyobject.rs
```diff
@@ -639,6 +640,12 @@ where
     T: IntoPyObject,
 {
     fn get_item(&self, key: T, vm: &VirtualMachine) -> PyResult {
+        if let Ok(map) = PyMappingMethods::try_from_borrowed_object(vm, self) {
+            if let Some(getitem) = map.subscript {
+                return getitem(self.clone(), key.into_pyobject(vm), vm);
+            }
+        }
+
```
youknowone, line 648: do we still need `__getitem__` call under this line? I think PyMapping covers the whole cases

Snowapril: Yes. there exists extra steps (sequence protocol check & method lookup) as [cpython implementation](https://github.com/python/cpython/blob/main/Objects/abstract.c#L180)

vm/src/stdlib/posix.rs
```diff
@@ -1049,8 +1050,21 @@ pub mod module {
     }
 
     #[cfg(any(target_os = "linux", target_os = "freebsd", target_os = "macos"))]
-    fn envp_from_dict(dict: PyDictRef, vm: &VirtualMachine) -> PyResult<Vec<CString>> {
-        dict.into_iter()
+    fn envp_from_dict(env: PyMapping, vm: &VirtualMachine) -> PyResult<Vec<CString>> {
```
youknowone, line 1053:
```suggestion
    fn envp_from_dict(env: crate::protocol::PyMapping, vm: &VirtualMachine) -> PyResult<Vec<CString>> {
```
to solve unused import warning

vm/src/protocol/mapping.rs
```diff
@@ -0,0 +1,142 @@
...
+        self.0
+    }
+
+    pub fn check(obj: &PyObjectRef, vm: &VirtualMachine) -> bool {
+        if let Ok(mapping) = PyMappingMethods::try_from_borrowed_object(vm, obj) {
+            mapping.subscript.is_some()
+        } else {
+            false
+        }
+    }
```
youknowone, line 53: `impl TryFromBorrowedObject for PyMappingMethods` can be placed here.

```rust
    pub fn (&self, vm: &VirtualMachine) -> PyMappingMethods {
        let obj_cls = self.0.class();
        ...
    }
```

vm/src/builtins/mappingproxy.rs
```diff
@@ -127,6 +138,37 @@ impl PyMappingProxy {
...
+    fn length(zelf: PyObjectRef, _vm: &VirtualMachine) -> PyResult<usize> {
+        unreachable!("length not implemented for {}", zelf.class())
+    }
+
+    #[inline]
+    fn subscript(zelf: PyObjectRef, needle: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        Self::downcast_ref(&zelf, vm).map(|zelf| zelf.getitem(needle, vm))?
+    }
+
+    #[inline]
```
youknowone, line 161:
```suggestion
    #[cold]
```

vm/src/builtins/bytes.rs
```diff
@@ -540,6 +540,36 @@ impl BufferInternal for PyRef<PyBytes> {
...
+    fn length(zelf: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
+        Self::downcast_ref(&zelf, vm).map(|zelf| Ok(zelf.len()))?
+    }
+
+    #[inline]
+    fn subscript(zelf: PyObjectRef, needle: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        Self::downcast_ref(&zelf, vm).map(|zelf| zelf.getitem(needle, vm))?
+    }
+
+    #[inline]
```
youknowone, line 562:
```suggestion
    #[cold]
```

Snowapril: It is cool 😊 Thanks

youknowone (approved):

PR #3132, opened 2021-09-25
vm::sysmodule -> vm::stdlib::sys

youknowone:
sys was the only inconsistent module not inside `stdlib` except for builtins.


vm/src/stdlib/sys.rs
```diff
@@ -0,0 +1,727 @@
+use num_traits::ToPrimitive;
```
DimitrisJim, line 1: I'm sort of confused. Isn't this supposed to replace `vm/src/sysmodule.rs`. Are you planning on another PR that completely removes sysmodule.rs?

youknowone: I am sorry. I will remove sysmodule.rs

PR #3135, opened 2021-09-25
tp_ -> slot_

youknowone:
`tp_` prefix is originated from CPython code, which conventionally use those kind of type prefixes for member fields.

RustPython is not following the naming policy, so it doesn't give any meaning to readers.
Renaming it will give more clear meanings about the type `PyTypeSlots` and slot implementation methods name.

### DIFF
--- a/vm/src/slots.rs
+++ b/vm/src/slots.rs
@@ -13,7 +13,7 @@ use crossbeam_utils::atomic::AtomicCell;
 use std::cmp::Ordering;
 
 bitflags! {
-    pub struct PyTpFlags: u64 {
+    pub struct PyTypeFlags: u64 {
         const HEAPTYPE = 1 << 9;
         const BASETYPE = 1 << 10;
         const METHOD_DESCR = 1 << 17;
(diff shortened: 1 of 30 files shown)


DimitrisJim:
> so it doesn't give any meaning to readers

I'd say it helps to be able to cross-reference with CPython. Then again, I really don't mind the rename, I always mentally think of the 'slot' whenever I see `tp_`. You work with many new time contributors so they might be better suited to say how much similar names help them.

youknowone:
Yes, I missed that point.
Still, CPython has member prefix convention across the whole project. I think keeping just this prefix while removing every other prefixes doesn't make that much sense.
Probably we'd better to get the cross-reference advantage from comments.

vm/src/slots.rs
```diff
@@ -75,6 +75,9 @@ pub(crate) type IterNextFunc = fn(&PyObjectRef, &VirtualMachine) -> PyResult;
 
 #[derive(Default)]
 pub struct PyTypeSlots {
+    // The corresponding field in CPython is `tp_` prefixed.
```
youknowone, line 78: @DimitrisJim I added short comment inspired by your comment

PR #3136, opened 2021-09-25
ResizeGuard -> BufferResizeGuard

youknowone:
to distinguish which type implements it by name


qingshi163:
As I remember ResizeGuard is meant to guarding the resize operation only. The writing operation that not change the size still can be done silently.

I will check it by the week so maybe we can have a better name.

youknowone:
Thanks, `BufferWritable` sounds like buffer itself is writable instead of the original object is resizable. I changed it to `BufferResizeGuard` by just putting the prefix

PR #3137, opened 2021-09-25
Fix array repr

youknowone:


DimitrisJim (approved):
iirc there's other classes that need a similar change for showing subclass name. I'll open an issue for them for new contributors.

PR #3142, opened 2021-09-25
int::try_to_primitive -> PyInt::try_to_primitive

youknowone:


vm/src/builtins/int.rs
```diff
@@ -80,7 +80,7 @@ macro_rules! impl_into_pyobject_int {
 
 impl_into_pyobject_int!(isize i8 i16 i32 i64 usize u8 u16 u32 u64 BigInt);
 
-pub fn try_to_primitive<'a, I>(i: &'a BigInt, vm: &VirtualMachine) -> PyResult<I>
+pub(crate) fn try_to_primitive<'a, I>(i: &'a BigInt, vm: &VirtualMachine) -> PyResult<I>
```
DimitrisJim, line 83: is this used elsewhere in the crate after this change? If not, why not make it private?

youknowone: `vm/src/iterator.rs` uses it

DimitrisJim: seems like that place can easily be changed to accommodate this change, either the case, this lgtm, I'll leave the choice of if you want to change it up to you.

DimitrisJim (approved):
looks

vm/src/builtins/int.rs
youknowone: I think this is still related to stdlib work. let me clean it up with it later.

PR #3144, opened 2021-09-26
pytype::new -> PyType::new

youknowone:
#3104


DimitrisJim (approved):
lgtm, never did understand why it was a standalone function.

Issue #3101, opened 2021-09-21
Bring documents from CPython


youknowone:
collecting script and basic adoption: #3098 
related issue: #2680

youknowone:
I will leave module and class item for other contributors.
Module will be straight-forward.
Class item can be more complex because the module name or a closure to retain proper document must be delivered to class items.

moreal:
I'll try to apply the module doc first.

youknowone:
all done

PR #3145, opened 2021-09-26
Sync some modules' doc with database

moreal:
This tries to resolve a part of #3101.

This pull request does:

 - Use document from the database (i.e. `derive/docs.json`) for modules derived with `rustpython_vm::macros::pymodule` (i.e. `rustpython_derive::pymodule::impl_pymodule`).
 
But this pull request doesn't achieve the purpose for modules derived with `rustpython_vm::macros::py_module` (though it seems equal with `rustpython_derive::pymodule::impl_pymodule`, maybe it seems left by historical(?) reason 🤔 . You can check this through `_csv` module.

### DIFF
--- a/derive/src/pymodule.rs
+++ b/derive/src/pymodule.rs
@@ -20,8 +20,8 @@ pub fn impl_pymodule(
     attr: AttributeArgs,
     module_item: Item,
 ) -> std::result::Result<TokenStream, Diagnostic> {
-    let mut module_item = match module_item {
-        Item::Mod(m) => m,
+    let (doc, mut module_item) = match module_item {
+        Item::Mod(m) => (m.attrs.doc(), m),
         other => bail_span!(other, "#[pymodule] can only be on a full module"),
     };
     let fake_ident = Ident::new("pymodule", module_item.span());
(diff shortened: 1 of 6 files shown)


youknowone:
The CPython doc database is not a part of vm. Only `rustpython-derive` has the database.
`py_module!` macro is defined in vm. There is no way to access to the database.
I think we'd better to deprecate `py_module!` in future for this doc machanism.

derive/src/pymodule.rs
```diff
@@ -58,10 +58,19 @@ pub fn impl_pymodule(
     // append additional items
     let module_name = context.name.as_str();
     let module_extend_items = context.module_extend_items;
+    let doc = crate::doc::try_read(module_name).ok().flatten();
+    let doc = if let Some(doc) = doc {
+        quote!(Some(#doc))
+    } else {
+        quote!(None)
+    };
```
youknowone, line 66: For the previously implemented items, they try to read user defined document first and CPython database document later. Probably because `pymodule` didn't have feature to load user defined document, this is only reading CPython document. I think this behavior is confused for its consistency.

Could you add either user defined document implementation or a `TODO:` comment for prioritized reading user defined document?

youknowone (changes requested):
other parts looks good

derive/src/pymodule.rs
moreal: I made it try to read user-defined documents first in dd5e55d. Please review again 🙏🏻

youknowone (approved):

derive/src/pymodule.rs
youknowone: looks good, thank  you!

PR #3147, opened 2021-09-26
Refactor ssl module

youknowone:
I moved `io::Read` and `io::Write` for PySocketRef to ssl. If they are implemented for `PySocket`, it would be member of socket. But the `PyRef<PySocket>` requirement is decided by `ssl::SslStream`. not by socket itself. So I moved it to ssl to make them specific to ssl requirement.

part of #3104

### DIFF
--- a/vm/src/stdlib/socket.rs
+++ b/vm/src/stdlib/socket.rs
@@ -5,7 +5,7 @@ use crate::{
     exceptions::{IntoPyException, PyBaseExceptionRef},
     function::{FuncArgs, OptionalArg, OptionalOption},
     utils::{Either, ToCString},
-    IntoPyObject, PyClassImpl, PyObjectRef, PyRef, PyResult, PyValue, TryFromBorrowedObject,
+    IntoPyObject, PyClassImpl, PyObjectRef, PyResult, PyValue, TryFromBorrowedObject,
     TryFromObject, TypeProtocol, VirtualMachine,
 };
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/socket.rs
```diff
@@ -170,7 +168,7 @@ impl PySocket {
         PyRwLockReadGuard::try_map(self.sock.read(), |sock| sock.get()).ok()
     }
 
-    fn sock_io(&self) -> io::Result<PyMappedRwLockReadGuard<'_, Socket>> {
+    pub(crate) fn sock_io(&self) -> io::Result<PyMappedRwLockReadGuard<'_, Socket>> {
```
coolreader18, line 171: Maybe it would make sense to expose a pub `send(&[u8]) -> io::Result<usize>` and `recv(&mut [u8]) -> io::Result<usize>` ~~for once the stdlib gets moved out of `rustpython-vm`~~ oh I guess `socket` is also in stdlib. But the above still seems like a better abstraction than exposing `ReadGuard`.

coolreader18: (and flush() for sockets is a no-op)

coolreader18 (approved):

vm/src/stdlib/socket.rs
youknowone: thanks. I made pub read, write and flush in PySocket

Issue #3122, opened 2021-09-23
cargo install --git https://github.com/RustPython/RustPython no longer works


andrew-ld:
```
andrew@arch ~> cargo install --git https://github.com/RustPython/RustPython
    Updating git repository `https://github.com/RustPython/RustPython`
  Installing rustpython v0.1.2 (https://github.com/RustPython/RustPython#0bb0946a)

error: failed to run custom build command for `pre-commit v0.5.2`

Caused by:
  process didn't exit successfully: `/tmp/cargo-installZNx3lq/release/build/pre-commit-0412e87eab42628d/build-script-build` (exit status: 101)
  --- stdout
  looking for root in "/tmp/cargo-installZNx3lq/release/build/pre-commit-171e0bad32e1578e/out"
  looking for root in "/tmp/cargo-installZNx3lq/release/build/pre-commit-171e0bad32e1578e"
  looking for root in "/tmp/cargo-installZNx3lq/release/build"
  looking for root in "/tmp/cargo-installZNx3lq/release"
  looking for root in "/tmp/cargo-installZNx3lq"
  looking for root in "/tmp"
  looking for root in "/"

  --- stderr
  thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Custom { kind: NotFound, error: "Crate root not found" }', /home/andrew/.cargo/registry/src/github.com-1ecc6299db9ec823/pre-commit-0.5.2/build.rs:9:17
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: failed to compile `rustpython v0.1.2 (https://github.com/RustPython/RustPython#0bb0946a)`, intermediate artifacts can be found at `/tmp/cargo-installZNx3lq`

Caused by:
  build failed

```

RustPython commit: 0bb0946a
Rust version: 1.55.0 (c8dfcfe04 2021-09-06)
Os: archlinux with linux 5.14

coolreader18:
@youknowone would you be opposed to getting rid of the pre-commit? I've just `rm`'d the hook whenever it gets re-added to the `.git/hooks` directory cause I don't like how it makes `git commit` take multiple seconds to run, and I was nervous about an issue like this popping up when I took a look at `pre-commit`'s source.

youknowone:
I think the idea adding precommit hook for fmt is good. It takes only a few seconds but ensure every commit is formatted - once if it is correctly set to add changed files which is not done yet.

[...]

PR #3149, opened 2021-09-27
remove precommit

eldpswp99:
I added precommit to reduce fmt ci failure
but this pre-commit tool raised some problems such as(#3122). so i removed the precommit.


DimitrisJim:
Shouldn't build.rs also be removed? As far as I can see it was added in #3010.

youknowone:
cc @coolreader18

DimitrisJim (approved):

PR #3150, opened 2021-09-27
OSError with filename and filename2

deantvv:
Implement OSError members `filename` and `filename2` by adding a new
struct `IOErrorWithFilename` will implement `IntoPyException`.

This PR will *NOT* pass more CPython test(*) but will allow user to
access `attr:filename` in OSError exception which IMO is a important
feacture.

The reason why CPython test failed before is `AttributeError: 'FileNotFoundError' object has no attribute 'filename'`
After this PR the reason will become `AssertionError: ...`
This is because CPython check the identity of `err.filename` to equal the input `filename` which means the only to pass this test is to store the `filename` object directly into `PyPathLike` or `PathOrFd` and return *the* `filename` if error which is how CPython deal with this with `path_t->object`.

### DIFF
--- a/Lib/test/test_fileio.py
+++ b/Lib/test/test_fileio.py
@@ -219,9 +219,10 @@ class AutoFileTests:
         self.assertRaises(TypeError, self.f.writelines)
         self.assertRaises(ValueError, self.f.writelines, b'')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def testOpendir(self):
+        # TODO: RUSTPYTHON
+        if sys.platform == "win32":
+            testOpendir = unittest.expectedFailure(testOpendir)
         # Issue 3703: opening a directory should fill the errno
         # Windows always returns "[Errno 13]: Permission denied
         # Unix uses fstat and returns "[Errno 21]: Is a directory"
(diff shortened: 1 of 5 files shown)


fanninpm:
> This PR will _NOT_ pass more CPython test(*)

CI is reporting two unexpected successes in `test_urllib`.

deantvv:
> 
> 
> > This PR will _NOT_ pass more CPython test(*)
> 
> CI is reporting two unexpected successes in `test_urllib`.

I missed that. I was too focus on `OSErrorTests::test_oserror_filename`.

I pushed a new commit that made `IOErrorWithFilename` a little bit more useful for letting it to be constructed with `PathOrFd`.

youknowone:
another unexpected success:
```
testOpendir (test.test_fileio.CAutoFileTests) ... unexpected success
```

youknowone (approved):

fanninpm:
> another unexpected success:
> 
> ```
> testOpendir (test.test_fileio.CAutoFileTests) ... unexpected success
> ```

Only on Windows. Ubuntu and macOS are reporting that as an expected failure. The following snippet may be inserted below the test in question:

```python
# TODO: RUSTPYTHON
if sys.platform == "win32":
    testOpendir = unittest.expectedFailure(testOpendir)
```

deantvv:
Changed as suggested. Hope this time will have a green light.

PR #3155, opened 2021-09-27
vm::utils::ascii! -> common::ascii!

youknowone:


common/src/str.rs
```diff
@@ -82,6 +82,28 @@ pub fn to_ascii(value: &str) -> String {
     ascii
 }
 
+pub const fn bytes_is_ascii(x: &str) -> bool {
```
coolreader18, line 85: This could be `#[doc(hidden)]`, since it's an implementation detail

coolreader18 (approved):

common/src/str.rs
```diff
@@ -82,6 +82,29 @@ pub fn to_ascii(value: &str) -> String {
...
+    true
+}
+
+#[macro_export]
+macro_rules! ascii {
+    ($x:literal) => {{
+        const _: () = {
+            ["not ascii"][!::rustpython_common::str::bytes_is_ascii($x) as usize];
+        };
+        unsafe { ::ascii::AsciiStr::from_ascii_unchecked($x.as_bytes()) }
```
coolreader18, line 104: Oh wait sorry this should probably also be a `$crate::` reference and the `ascii` crate should be `pub use`'d somewhere

common/src/str.rs
```diff
@@ -82,6 +82,29 @@ pub fn to_ascii(value: &str) -> String {
...
+        i += 1;
+    }
+    true
+}
+
+#[macro_export]
+macro_rules! ascii {
+    ($x:literal) => {{
+        const _: () = {
+            ["not ascii"][!::rustpython_common::str::bytes_is_ascii($x) as usize];
```
coolreader18, line 102: This can be `$crate::str::...`

common/src/str.rs
youknowone: I created `common::vendored::ascii`

PR #3156, opened 2021-09-27
remove exception_slots

youknowone:


DimitrisJim (approved):
apparently not, lgtm after warning is addressed.

PR #3158, opened 2021-09-27
Miscellaneous pip fixes

coolreader18:
Fix pip warnings about inconsistency between distutils' sysconfig and the actual sysconfig; fix error when installing package it would say something like `__env_pip__.zip is not a file`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -66,6 +66,10 @@ path = "src/main.rs"
 [profile.dev.package."*"]
 opt-level = 3
 
+[profile.test]
+opt-level = 3
+lto = "thin"
+
 [profile.bench]
 lto = true
 codegen-units = 1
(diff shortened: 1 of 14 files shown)


Cargo.toml
```diff
@@ -66,6 +66,10 @@ path = "src/main.rs"
 [profile.dev.package."*"]
 opt-level = 3
 
+[profile.test]
+opt-level = 3
```
youknowone, line 70: isn't this obfuscate backtrace and debug information? 
how do you think about running tests with release build job if this is desired?

youknowone (approved):

Cargo.toml
coolreader18: It should still keep debug info I'm pretty sure - the default for test enables debuginfo and I'm pretty sure it's orthogonal to opt-level: https://doc.rust-lang.org/nightly/cargo/reference/profiles.html#test

youknowone: When opt-level is not 0, the inlined callstacks are omitted, the progress is not linear to the code and some breakpoint doesn't work.

youknowone: I think the default development environment is better to be friendly to programmers unless there are unavoidable reasons

fanninpm:
13 unexpected successes from a test that wasn't in the codebase when you opened this — sounds like another reason for [bors](https://bors.tech/).

PR #3129, opened 2021-09-25
rustpython-stdlib crate

youknowone:
#3102

PR #3161, opened 2021-09-28
move init_winsock to nt

youknowone:
`socket` is stdlib but `signal` in vm requires this init. For now, we don't have any option to disable signal. So this might be ok.
Any better idea?


youknowone:
merged as part of #3129

PR #3163, opened 2021-09-28
android SSL support system certs

afpro:
Build with
cargo ndk -t arm64-v8a -p 23 build --release --features freeze-stdlib,ssl,ssl-vendor

Then push rustpython to /data/local/tmp/python

run /data/local/tmp/python with code:
import urllib.request
print(urllib.request.urlopen('https://httpbin.org/ip').read().decode('ascii'))

check if https connection successful

### DIFF
--- a/vm/src/stdlib/ssl.rs
+++ b/vm/src/stdlib/ssl.rs
@@ -285,6 +285,9 @@ impl SlotConstructor for PySslContext {
         let mut builder =
             SslContextBuilder::new(method).map_err(|e| convert_openssl_error(vm, e))?;
 
+        #[cfg(target_os = "android")]
+        droid::load_client_ca_list(vm, &mut builder)?;
+
         let check_hostname = proto == SslVersion::TlsClient;
         builder.set_verify(if check_hostname {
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/ssl.rs
```diff
@@ -1257,3 +1260,83 @@ impl Write for SocketStream {
         socket.flush()
     }
 }
+
+#[cfg(target_os = "android")]
+mod droid {
```
youknowone, line 1265: is there any reason this is not `android` but `droid`? are there more droids?

youknowone:
patch to cleanup: https://github.com/RustPython/RustPython/pull/3163/commits/728af615ff96efb3403979b6b0384722d1e133ac

vm/src/stdlib/ssl.rs
afpro: sorry about my bad naming style, i'll fix this soon.

youknowone (approved):

vm/src/stdlib/ssl.rs
youknowone: don't worry. I was actually wondered how it is named because I don't know about them well.

youknowone:
Thank you for contributing!

PR #3164, opened 2021-09-28
WIP: Fix `PyStr::startswith`

### LINKED ISSUE #3162: slice multi-byte characters panicing
- [slicing-strings doc](https://doc.rust-lang.org/book/ch08-02-strings.html#slicing-strings)
```python
from html.parser import HTMLParser

class TitleParser(HTMLParser):

    def __init__(self):
        HTMLParser.__init__(self)
        self.match = False
        self.title = None

    def handle_starttag(self, tag, attrs):
        self.match = tag == 'title'

    def handle_data(self, data):
        if self.match:
            self.title = data
            self.match = False


response_text = "<title>登录</title>"
title_parser = TitleParser()
title_parser.feed(response_text)
title = title_parser.title
title_parser.close()
print(title)
```
- output
```bash
rustpython Test/test_web_title.py

  File "Test/test_web_title.py", line 23, in <module>
    title_parser.feed(response_text)
  File "html.parser", line 111, in feed
  File "html.parser", line 173, in goahead
  File "html.parser", line 388, in parse_endtag
AssertionError: unexpected call to parse_endtag
```
- parser.py
```python
    def parse_endtag(self, i):
        rawdata = self.rawdata
        assert rawdata[i:i+2] == "</", "unexpected call to parse_endtag"
```
- `[i:i+2]`,Unicode is twice as long as `i`
```bash
thread '<unnamed>' panicked at 'byte index 220 is not a char boundary; it is ins
ide '登' (bytes 218..221) of `<meta http-equiv="X-UA-Compatible" content="IE=edg
e">
        <meta http-equiv="X-Permitted-Cross-Domain-Policies" content="master-onl
y">
    <meta name="viewport" content="width=device-width, initial-scale=1">
    <title>登录</title>
        <meta http-equiv="key`[...]', vm\src\builtins\pystr.rs:1474:10
```

moreal:
# WIP: I opened this pull request to run tests using CI.

This pull request tries to fix #3162.


youknowone:
Hmm, this is intended not to use expensive `char_len()`. Because our str internal is always a valid utf-8, when it actually `startswith` or `endswith` another string, their byte representation must be always same. Did you find the specific point that byte len makes problem? I want to fix it without char_len() if possible

youknowone:
Ah, I got it. `start`, `end` in adjust_indices are char index but len is byte index. Could it be solved by processing `start` and `end` first as char operations and `len` later to keep the startswith itself as a byte operation?

```rust
impl StartsEndsWithArgs {
    fn get_value(self, len: usize) -> (PyObjectRef, std::ops::Range<usize>) {
        let range = adjust_indices(self.start, self.end, len);
        (self.affix, range)
    }
}
```

moreal:
@youknowone I tried to rewrite codes in my best but feel free to suggest your better idea 🙇🏻

PR #3165, opened 2021-09-28
suggestion: corelib only expose pymodules

youknowone:
This is a sugestion about module public api.

What users should see in our module public api?
My suggestion is the same as Python.

So, if there is `time.time()` in Python, `corelib::time::time()` in Rust.

To achieve this, we need to expose inner pymodule to public api, not the outer one currently in `vm::stdlib`


coolreader18:
An alternative is I guess to do `pub use module::*;` at the top of the file

PR #3168, opened 2021-09-29
move builtins module to vm::stdlib::builtlins for consistency

youknowone:


youknowone:
merged as part of #3129

PR #3169, opened 2021-09-29
pub(crate) for type modules

youknowone:
there are no pub free functions exposed in this way.


DimitrisJim (approved):
Looks fine, have you investigated why android complains for the check?

youknowone:
No I didn't. I think it is just simply because android is not using PyMapping somewhere. Those stuff will be completely removed by #3121.  So I will not be a problem anymore. I will merge this PR once after #3121 is done to avoid warnings.

PR #3170, opened 2021-09-29
Expose subtle public interfaces

youknowone:
split from #3104


youknowone:
merged as part of #3129

PR #3173, opened 2021-09-30
OSError with filename2

deantvv:
Add OSError with filaname(2) support for `os.rename`, `os.readlink`, `os.link` and `os.chroot`

Follow up for #3150

### DIFF
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -223,6 +223,15 @@ impl From<PyPathLike> for PathOrFd {
     }
 }
 
+impl PathOrFd {
+    pub fn filename(&self, vm: &VirtualMachine) -> PyObjectRef {
+        match self {
+            PathOrFd::Path(path) => path.filename(vm).unwrap_or_else(|_| vm.ctx.none()),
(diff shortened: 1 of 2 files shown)


youknowone (changes requested):
`IOErrorWithFilename` is not a real error but a builder for IOErrror. It doesn't matter that much only with filename, but now it looks like to be better to follow the common builder pattern.

```rust
pub struct IOErrorBuilder {
    error: io::Error,
    filename: Option<PathOrFd>,
    filename2: Option<PathOrFd>,
}

impl IOErrorBuilder {
    fn new(error: io::Error) -> Self {
        Self { error, filename: None, filename2: None }
    }
    fn filename(self, filename: PathOrFd) -> Self {
        self.filename.replace(filename)
        self
    }
    fn filename2(self, filename: PathOrFd) -> Self {
        self.filename2.replace(filename)
        self
    }
}

impl IntoPyException for IOErrorBuilder {
    fn into_pyexception(self, vm: &VirtualMachine) -> PyBaseExceptionRef {
        ...
    }
}
```

deantvv:
Builder pattern makes this code better. Thanks for the suggestion.

youknowone (approved):

PR #3178, opened 2021-09-30
Fixes for wasi

coolreader18:


vm/src/stdlib/mod.rs
```diff
@@ -34,10 +34,8 @@ pub mod nt;
 pub mod posix;
 #[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))]
 #[cfg(not(any(unix, windows)))]
-pub(crate) mod posix_compat;
-#[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))]
-#[cfg(not(any(unix, windows)))]
-pub use posix_compat as posix;
+#[path = "posix_compat.rs"]
```
youknowone, line 37: :+1:

src/lib.rs
```diff
@@ -477,6 +470,15 @@ fn get_paths(env_variable_name: &str) -> impl Iterator<Item = String> + '_ {
                 .collect::<Vec<_>>()
         })
 }
+#[cfg(not(target_os = "wasi"))]
+use env::split_paths;
```
youknowone, line 474:
```suggestion
use std::env::split_paths;
```
I think indirect use here doesn't have any advantage.

src/lib.rs
```diff
@@ -477,6 +470,15 @@ fn get_paths(env_variable_name: &str) -> impl Iterator<Item = String> + '_ {
...
+#[cfg(target_os = "wasi")]
+fn split_paths<T: AsRef<std::ffi::OsStr> + ?Sized>(s: &T) -> impl Iterator<Item = PathBuf> + '_ {
+    use std::os::wasi::ffi::OsStrExt;
+    let s = s.as_ref().as_bytes();
+    s.split(|b| *b == b':')
+        .map(|x| std::ffi::OsStr::from_bytes(x).to_owned().into())
+}
```
youknowone, line 481: i think we need collection of platform utility somewhere

youknowone (approved):

PR #3183, opened 2021-10-01
 use PyIterReturn for gen internal

youknowone:


vm/src/coroutine.rs
```diff
@@ -1,55 +1,60 @@
...
+            }
         }
     }
 }
 
 #[derive(Debug)]
 pub struct Coro {
     frame: FrameRef,
-    pub closed: AtomicCell<bool>,
+    pub closed: AtomicCell<bool>, // redudant?
```
DimitrisJim, line 31: If frame was made into an option `closed` could be removed. That's closer to what CPython does.

DimitrisJim:
merged #3185 first so other PRs that might come before this is merged won't get stuck on the failing tests.

vm/src/coroutine.rs
youknowone: I replaced this comment to link to this comment

youknowone:
merged by #3202

Issue #3182, opened 2021-10-01
Fix complex.__repr__ handling of nan.


DimitrisJim:
Specifically two related issues:

 - Nan's should be lowercased (similar to #3034)
 - Signs for nan should be included (always `+`), an issue really if the real part of the complex number is != 0.

For example, in Python:

```python
>>> complex(0, float('nan'))
nanj
>>> complex(float('nan'), float('nan'))
(nan+nanj)
>>> complex(float('nan'), -float('nan'))
(nan+nanj)
>>> complex(-float('nan'), float('nan'))
(nan+nanj)
```

while in RustPython:

```python
>>>>> complex(0, float('nan'))
NaNj
>>>>> complex(float('nan'), float('nan'))
(NaNNaNj)
>>>>> complex(float('nan'), -float('nan'))
(NaNNaNj)
>>>>> complex(-float('nan'), float('nan'))
(NaNNaNj)
```

mdegis:
I think I can work on this 👍

PR #3184, opened 2021-10-01
Fix complex.__repr__ handling of nan

mdegis:
This PR will fix https://github.com/RustPython/RustPython/issues/3182

### Tested Scenarios 
```python
# default examples taken from issue:
>>>>> complex(0, float('nan'))
nanj
>>>>> complex(float('nan'), float('nan'))
(nan+nanj)
>>>>> complex(float('nan'), -float('nan'))
(nan+nanj)
>>>>> complex(-float('nan'), float('nan'))
(nan+nanj)
# some random complex numbers to also check it out manually
>>>>> 10j
10j
>>>>> 1-10j
(1-10j)
>>>>> complex(10, float('10'))
(10+10j)
>>>>> complex(-10, float('10'))
(-10+10j)
complex(10, float('nan'))
(10+nanj)
>>>>> complex(float('nan'), 10)
(nan+10j)
>>>>> complex(float('nan'), 1)
(nan+1j)
```

I am really new to Rust so feel free to give any constructive criticism 🥊

### DIFF
--- a/vm/src/builtins/complex.rs
+++ b/vm/src/builtins/complex.rs
@@ -305,10 +305,26 @@ impl PyComplex {
     #[pymethod(magic)]
     fn repr(&self) -> String {
         let Complex64 { re, im } = self.value;
-        if re == 0.0 {
+        let re_part = if re.is_nan() {
+            "nan".to_string()
+        } else {
+            format!("{}", re)
+        };
+        let im_part = if im.is_nan() {
+            if re == 0.0 {
+                "nanj".to_string()
+            } else {
+                "+nanj".to_string()
+            }
+        } else if re == 0.0 {
             format!("{}j", im)
         } else {
-            format!("({}{:+}j)", re, im)
+            format!("{:+}j", im)
+        };
+        if re == 0.0 {
+            format!("{}", im_part)
+        } else {
+            format!("({}{})", re_part, im_part)
         }
     }
 


vm/src/builtins/complex.rs
```diff
@@ -305,10 +305,26 @@ impl PyComplex {
     #[pymethod(magic)]
     fn repr(&self) -> String {
         let Complex64 { re, im } = self.value;
-        if re == 0.0 {
+        let re_part = if re.is_nan() {
```
DimitrisJim, line 308: So, here's a another idea. We can re-use float formatting to get the job done (and get long float values for the `real` and `imag` parts to print out in scientific notation).

Inside `float_ops` we have `to_string` that takes an `f64` and returns it nicely formatted (it also includes handling of the `NaN`). The main issue is the sign we add, if its a `nan` for `imag` we must always prefix with a `"+"` (granted that a `real` part exists).

mdegis: That's a really great suggestion, sorry I did not check `float_ops` before 😞 
Changing this right now.

mdegis:
> Inside float_ops we have to_string that takes an f64 and returns it nicely formatted (it also includes handling of the NaN). The main issue is the sign we add, if its a nan for imag we must always prefix with a "+" (granted that a real part exists).

So I've made the change you suggested as:
```rust
if re == 0.0 {
    format!("{}j", float_ops::to_string(im))
} else if im.is_nan() {
    format!("({}+nanj)", float_ops::to_string(re))
} else {
    format!("({}{:+}j)", float_ops::to_string(re), float_ops::to_string(im))
}
```
It covered examples given in the issue:

```python
>>>>> complex(0, float('nan'))
nanj
>>>>> complex(float('nan'), float('nan'))
(nan+nanj)
>>>>> complex(float('nan'), -float('nan'))
(nan+nanj)
>>>>> complex(-float('nan'), float('nan'))
(nan+nanj)
```

However, when I tested it with some random complex numbers, it does not go as planned:

```python
>>>>> 10j
10.0j
>>>>> 10j
10.0j
>>>>> 1-10j
(1.0-10.0j)
>>>>> complex(10, float('10'))
(10.010.0j)
>>>>> complex(-10, float('10'))
(-10.010.0j)
```

I know that `{:+}` does not work because given value is not numerical, it can be fixed with checking if the `im` part is greater than 0 and add plus sign but it does not feel right.

Second to that, `float_ops::to_string` serialize it as float always. For example; in current -messy- implementation `10j` will be `10j` but new one will print it as `10.0j`.

Right now, I am little bit confused how to proceed (maybe I misunderstood something?) . Open to any suggestion 😿

DimitrisJim:
You can forward the `f64` to `floats_ops::to_string` only if it has a fractional part (`f64::fract`) that is `!= 0.0`.

PR #3187, opened 2021-10-02
Use PyInt::try_to_primitive.

DimitrisJim:
clean up after #3142

### DIFF
--- a/vm/src/iterator.rs
+++ b/vm/src/iterator.rs
@@ -3,11 +3,10 @@
  */
 
 use crate::{
-    builtins::{int, PyBaseExceptionRef, PyInt},
+    builtins::{PyBaseExceptionRef, PyInt},
     protocol::{PyIter, PyIterReturn},
     IdProtocol, PyObjectRef, PyResult, TypeProtocol, VirtualMachine,
 };
-use num_traits::Signed;
 
 pub fn try_map<F, R>(vm: &VirtualMachine, iter: &PyIter, cap: usize, mut f: F) -> PyResult<Vec<R>>
 where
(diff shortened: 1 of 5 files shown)


vm/src/iterator.rs
```diff
@@ -66,20 +65,19 @@ pub fn length_hint(vm: &VirtualMachine, iter: PyObjectRef) -> PyResult<Option<us
...
                 result.class().name()
             ))
         })?
-        .as_bigint();
-    if result.is_negative() {
-        return Err(vm.new_value_error("__length_hint__() should return >= 0".to_owned()));
+        .try_to_primitive(vm)
+    {
+        Ok(hint) => Ok(Some(hint)),
+        Err(_) => Err(vm.new_value_error("__length_hint__() should return >= 0".to_owned())),
```
youknowone, line 79: The original intend looks like distinguising overflow error and negative value error

youknowone:
An idea. Sometimes, there will be some features we cannot move from BigInt to PyInt. Would it be better if I didn't move try_to_primitive into PyInt but adding `trait PyIntExt` and impl it for BigInt?

vm/src/iterator.rs
DimitrisJim: ah, good point. I'll address that.

DimitrisJim:
@youknowone are there places that would need that now? `try_from_primitive` was the last (as far as I can see) in `int.rs` that was a free-standing public function that accepted a `BigInt`.

youknowone (approved):

youknowone:
Probably `try_f64_to_bigint` and `try_bigint_to_f64` are one of those example for same relationship. And then I also think try_from_primitive is also one of the member of the group

PR #3082, opened 2021-09-18
increase stack size for windows

youknowone:
fix #2881

could anyone using windows test if this change works?

PR #3192, opened 2021-10-02
Fix windows stack size again

youknowone:
Fix #3082


youknowone:
@oroppas I am sorry. I am going to fix it again here 😅

oroppas:
Thanks for the fix!

PR #3194, opened 2021-10-02
underscore issue on parse_str

eunsung-lim:
When parssing string to float, strings with underscores need to raise error
ex. "1_e1"


youknowone (changes requested):
By this change, you turned a few tests to pass. Removing `unittest.expectedFailure` from those tests will make your PR pass tests

youknowone:
could you run fmt and push it again? `cargo fmt --all` will work

DimitrisJim:
Thanks, not quite sure unfortunately when, but this currently behaves as expected (code currently resides in the Parser repository). Thanks for the PR though!

PR #3198, opened 2021-10-02
Added inverse sine and inverse cosine

rishit-khandelwal:


stdlib/src/cmath.rs
```diff
@@ -78,12 +78,23 @@ mod cmath {
         z.to_complex().sin()
     }
 
+    /// Return the inverse sine of z
```
youknowone, line 81: We don't write our original document here. They are expected to be same as CPython.

Actually we don't need document anyomre. #3101 is recently brought up, so you simply can remove comment here to let it follow the default CPython document database.

youknowone (changes requested):

stdlib/src/cmath.rs
rishit-khandelwal: Ok. Should i also remove all the comments which were there previously?

egemengol:
I think updating the tests may be worthwhile, you could refer to #3200 if you want to

rishit-khandelwal:
ok!

rishit-khandelwal:
i have made the changes in test_cmath.py

DimitrisJim:
You'll need to rebase, there were recent changes in `cmath` that you need to include. Running the tests will currently fail since `asin` and `acos` have some slight rounding errors as discussed #3039. For now, include the additional changes in `test_cmath` as done in #3200, i.e:

```python
# TODO: RUSTPYTHON
# This line currently fails for acos and asin
# cmath.asin(0.2) should produce a real number,
# but imaginary part is 1.1102230246251565e-16 for both.
if fn != "asin" and fn != "acos":
    self.assertEqual(0., z.imag)
```

@youknowone I'll handle this PR and then open another one to bring `test_cmath` as close as possible to its original form.

Lib/test/test_cmath.py
```diff
@@ -2,7 +2,8 @@
 from test.test_math import parse_testfile, test_file
 import test.test_math as test_math
 import unittest
-import cmath, math
+import cmath
```
DimitrisJim, line 5: did you run a formatter on the file? We follow the same exact structure as CPython's test files (except for cases where an error justifies the change). If you can, revert to the previous formatting.

rishit-khandelwal: ok i will revert back to old formatting

DimitrisJim:
Great! You'll just need to rebase now, there's still conflicts present.

youknowone:
rebase is required, not merge(or pull)

check this article: https://www.atlassian.com/git/tutorials/merging-vs-rebasing

rishit-khandelwal:
I am very sorry for making this mess, how can i fix this? Sorry for the trouble 😔

youknowone:
I would do squash(using rebase -i to last common commit) everything to a single commit and cherry-pick it onto the upstream main.

rishit-khandelwal:
[Other PR](https://github.com/RustPython/RustPython/pull/3227)

Sorry for troubling you, but I couldnt figure out how to fix this 😔

PR #3200, opened 2021-10-02
Implement asin and acos in cmath

egemengol:
These functions show a weird rounding error like this:
```python
>>>>> import cmath
>>>>> cmath.acos(0.2)
(1.369438406004566+0.00000000000000011102230246251565j)
>>>>> cmath.asin(0.2)
(0.20135792079033082+0.00000000000000011102230246251565j)
```

For which CPython returns:
```python
Python 3.9.7 (default, Oct  1 2021, 17:26:30) 
[GCC 9.3.0] on linux
>>> import cmath
>>> cmath.acos(0.2)
(1.369438406004566-0j)
>>> cmath.asin(0.2)   
(0.20135792079033082+0j)
```

The tests are ignored for the moment, see [here](https://github.com/RustPython/RustPython/issues/3039#issuecomment-932422414)

Docstrings are copied from CPython docs.

### DIFF
--- a/Lib/test/test_cmath.py
+++ b/Lib/test/test_cmath.py
@@ -56,9 +56,15 @@ class CMathTests(unittest.TestCase):
     #
     # list of all functions in cmath
     test_functions = [getattr(cmath, fname) for fname in [
-        'sin','cos','log','log10','sqrt','acosh','tan','tanh','asinh', 'atan', 'atanh', 'sinh', 'cosh', 'exp'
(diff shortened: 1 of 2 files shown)


egemengol:
Please note that the result of `acos` should be negative zero but it is positive something instead.

Snowapril:
Ah.. I think duplicated #3198 here.

egemengol:
It seems so, please close at will :)

PR #3202, opened 2021-10-02
Remove not very useful iterator utilty functions

youknowone:
based on #3183


DimitrisJim:
Right, was going to ask, does this supersede #3183 or are you still planning on merging that too?

youknowone:
either way ok by review convenience

DimitrisJim (approved):
looks great to me, quick timings show me there's also a _slight_ performance gain too.

youknowone:
yes, though it is just a little bit

Issue #2979, opened 2021-08-27
storing to `__debug__` should be rejected by compiler.


HyeockJinKim:
## Feature

In Cpython, raise syntax error when saving a value to `__debug__`.
```python
>>> __debug__ = 3
  File "<stdin>", line 1
SyntaxError: cannot assign to __debug__
```

In RustPython, value can be stored to `__debug__`.
```python
>>>>> __debug__ = 3
>>>>> 
```

DimitrisJim:
The additional cases not covered all appear to be related to function definitions/calls:

```python
>>> def foo(__debug__): pass
...
  File "<stdin>", line 1
SyntaxError: cannot assign to __debug__
>>> def foo(x): pass
...
>>> foo(__debug__=30)
  File "<stdin>", line 1
SyntaxError: cannot assign to __debug__
>>> def foo(__debug__=3): pass
...
  File "<stdin>", line 1
SyntaxError: cannot assign to __debug__
```

The other un-assignables (`True, False`) all handle this case correctly.

muscar:
Hi there 👋🏻 I think https://github.com/RustPython/RustPython/pull/3203 should cover all the cases. It looks like the GH workflow needs approval before it can run.

DimitrisJim:
Additional cases (some present in `test_syntax.py` and `test_compile.py`):

 - Calling a function accepting arbitrary kwargs: `def foo(**kwargs): pass; foo(__debug__='ok')`
 - Setting `__debug__` on `builtins` (tested in `test_compile.py`):

    ```python
    import builtins
    setattr(builtins, "__debug__", "ok")
    __debug__ == "ok"  # True
   ```

 - Wrong error message (implementation probably checks if args match before checking if arg name is ok):

    ```python
    def f(a, *args, **kwargs): pass
    f(__debug__ = 30)  # missing 1 pos arg, should be syntax error.
   ```
 - Attribute assignment. Currently not handled.
    
    ```python
    class Foo: pass
    f = Foo()
    f.__debug__ = 30
    ```
    A second case of this for an object that hasn't been defined yet should also be handled. CPython raises an error for the
    `__debug__` assignment while RustPython (currently) complains about an undefined name:

    ```python
     # CPython
     obj.__debug__ = 30   # SyntaxError
    # RustPython
    obj.__debug__ = 30   # NameError

youknowone:
The given cases on this issue and comments are fully fixed

DimitrisJim:
[...]

PR #3203, opened 2021-10-02
fix: forbid stores to the __debug__ builtin

muscar:
Addresses https://github.com/RustPython/RustPython/issues/2979.

According to [the docs][1], `__debug__` is treated as a constant by the interpreter.

This patch adds some checks to bring RustPython's behaviour in line with cpython's.

[1]: https://docs.python.org/3/library/constants.html#built-in-constants

### DIFF
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -396,19 +396,12 @@ impl Compiler {
     }
 
     fn check_forbidden_name(&self, name: &str, usage: NameUsage) -> CompileResult<()> {
-        if usage == NameUsage::Store && is_forbidden_name(name) {
-            return Err(self.error(CompileErrorType::SyntaxError(format!(
-                "cannot assign to {}",
-                name
-            ))));
-        }
-        if usage == NameUsage::Delete && is_forbidden_name(name) {
-            return Err(self.error(CompileErrorType::SyntaxError(format!(
-                "cannot delete {}",
-                name
-            ))));
-        }
-        Ok(())
+        let msg = match usage {
+            NameUsage::Store if is_forbidden_name(name) => "cannot assign to",
+            NameUsage::Delete if is_forbidden_name(name) => "cannot delete",
+            _ => return Ok(()),
+        };
+        Err(self.error(CompileErrorType::SyntaxError(format!("{} {}", msg, name))))
     }
 
     fn compile_name(&mut self, name: &str, usage: NameUsage) -> CompileResult<()> {
--- a/compiler/src/compile.rs
+++ b/compiler/src/compile.rs
@@ -22,6 +22,7 @@ use std::borrow::Cow;
 
 type CompileResult<T> = Result<T, CompileError>;
 
+#[derive(PartialEq, Eq, Clone, Copy)]
 enum NameUsage {
     Load,
     Store,
@@ -49,6 +50,13 @@ impl CallType {
             CallType::Ex { has_kwargs } => Instruction::CallMethodEx { has_kwargs },
         }
     }
+}
+
+fn is_forbidden_name(name: &str) -> bool {
+    // See https://docs.python.org/3/library/constants.html#built-in-constants
+    const BUILTIN_CONSTANTS: &[&str] = &["__debug__"];
+
+    BUILTIN_CONSTANTS.contains(&name)
 }
 
 /// Main structure holding the state of compilation.
@@ -375,26 +383,39 @@ impl Compiler {
         Ok(())
     }
 
-    fn load_name(&mut self, name: &str) {
+    fn load_name(&mut self, name: &str) -> CompileResult<()> {
         self.compile_name(name, NameUsage::Load)
     }
 
     fn store_name(&mut self, name: &str) -> CompileResult<()> {
-        if name.eq("__debug__") {
-            return Err(self.error(CompileErrorType::SyntaxError(
-                "cannot assign to __debug__".to_owned(),
-            )));
-        }
-        self.compile_name(name, NameUsage::Store);
-        Ok(())
+        self.compile_name(name, NameUsage::Store)
     }
 
     fn mangle<'a>(&self, name: &'a str) -> Cow<'a, str> {
         symboltable::mangle_name(self.class_name.as_deref(), name)
     }
 
-    fn compile_name(&mut self, name: &str, usage: NameUsage) {
+    fn check_forbidden_name(&self, name: &str, usage: NameUsage) -> CompileResult<()> {
+        if usage == NameUsage::Store && is_forbidden_name(name) {
+            return Err(self.error(CompileErrorType::SyntaxError(format!(
+                "cannot assign to {}",
+                name
+            ))));
+        }
+        if usage == NameUsage::Delete && is_forbidden_name(name) {
+            return Err(self.error(CompileErrorType::SyntaxError(format!(
+                "cannot delete {}",
+                name
+            ))));
+        }
+        Ok(())
+    }
+
+    fn compile_name(&mut self, name: &str, usage: NameUsage) -> CompileResult<()> {
         let name = self.mangle(name);
+
+        self.check_forbidden_name(&name, usage)?;
+
         let symbol_table = self.symbol_table_stack.last().unwrap();
         let symbol = symbol_table.lookup(name.as_ref()).expect(
             "The symbol must be present in the symbol table, even when it is undefined in python.",
@@ -461,6 +482,8 @@ impl Compiler {
             },
         };
         self.emit(op(idx as u32));
+
+        Ok(())
     }
 
     fn compile_statement(&mut self, statement: &ast::Stmt) -> CompileResult<()> {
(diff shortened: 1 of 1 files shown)


youknowone:
@HyeockJinKim could you review this PR?

compiler/src/compile.rs
```diff
@@ -51,6 +52,15 @@ impl CallType {
     }
 }
 
+// See https://docs.python.org/3/library/constants.html#built-in-constants
+static BUILTIN_CONSTANTS: phf::Set<&'static str> = phf::phf_set! {
```
DimitrisJim, line 56: A constant set seems like its too much here. There's only a couple of constants (Python has 4, `None, True, False, __debug__`, and usually skips checking all of them), a slice would probably do the trick without needing an additional crate.

DimitrisJim:
I'd also add a couple of checks in `extra_tests/snippets` if the run doesn't produce any unexpected successes. (not sure if the test suite currently checks these).

fanninpm:
> I'd also add a couple of checks in `extra_tests/snippets` if the run doesn't produce any unexpected successes. (not sure if the test suite currently checks these).

One check is in `test_syntax.py`, but it's commented out (because doctests behave weirdly). Another check is in `test_compile.py`, which we currently don't have.

DimitrisJim:
@fanninpm mind adding `test_compile`? would be good to test this PR against it.

muscar:
Thanks both. `test_compile` and `test_syntax` pass.

```alex@tekton:~/src/hf/RustPython> cargo run --release -- -m test test_syntax -j 4
    Finished release [optimized] target(s) in 0.09s
     Running `target/release/rustpython -m test test_syntax -j 4`
Run tests in parallel using 4 child processes
0:00:00 load avg: 0.47 [1/1] test_syntax passed

== Tests result: SUCCESS ==

1 test OK.

Total duration: 241 ms
Tests result: SUCCESS
alex@tekton:~/src/hf/RustPython> cargo run --release -- -m test test_compile -j 4
    Finished release [optimized] target(s) in 0.09s
     Running `target/release/rustpython -m test test_compile -j 4`
Run tests in parallel using 4 child processes
0:00:01 load avg: 0.43 [1/1] test_compile passed

== Tests result: SUCCESS ==

1 test OK.

Total duration: 1 sec 156 ms
Tests result: SUCCESS
alex@tekton:~/src/hf/RustPython> cargo run --release -- -m test test_syntax -j 4
    Finished release [optimized] target(s) in 0.09s
     Running `target/release/rustpython -m test test_syntax -j 4`
Run tests in parallel using 4 child processes
0:00:00 load avg: 0.47 [1/1] test_syntax passed

== Tests result: SUCCESS ==

1 test OK.

Total duration: 241 ms
Tests result: SUCCESS
alex@tekton:~/src/hf/RustPython> cargo run --release -- -m test test_compile -j 4
    Finished release [optimized] target(s) in 0.09s
     Running `target/release/rustpython -m test test_compile -j 4`
Run tests in parallel using 4 child processes
0:00:01 load avg: 0.43 [1/1] test_compile passed

== Tests result: SUCCESS ==

1 test OK.

Total duration: 1 sec 156 ms
Tests result: SUCCESS
```

compiler/src/compile.rs
```diff
@@ -51,6 +52,13 @@ impl CallType {
     }
 }
 
+// See https://docs.python.org/3/library/constants.html#built-in-constants
+static BUILTIN_CONSTANTS: [&str; 1] = ["__debug__"];
+
+fn is_builtin_constant(name: &str) -> bool {
+    BUILTIN_CONSTANTS.contains(&name)
+}
```
youknowone, line 60:
```suggestion
fn is_builtin_constant(name: &str) -> bool {
    // See https://docs.python.org/3/library/constants.html#built-in-constants
    const BUILTIN_CONSTANTS: &[&str] = &["__debug__"];

    BUILTIN_CONSTANTS.contains(&name)
}
```

youknowone (dismissed):

DimitrisJim:
There's still a couple of things missing (see `test_debug_assignment` in `test_compile.py` and some of the doctests in `test_syntax`).

Additionally, supplying `__debug__` as a kwarg is still not handled (`def foo(**_): pass; foo(__debug__='yup')`). I say we merge this now since its in the right direction (and only needs extending to handle these other cases), I'll post this comment to the issue so you (or anyone else) can try and tackle the remaining cases.

fanninpm (changes requested):

Lib/test/test_compile.py
```diff
@@ -61,7 +61,6 @@ def test_none_keyword_arg(self):
     def test_duplicate_global_local(self):
         self.assertRaises(SyntaxError, exec, 'def f(a): global a; a = 1')
 
-    # TODO: RUSTPYTHON
     @unittest.expectedFailure
```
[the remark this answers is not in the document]
DimitrisJim: huh, guess not or else the CI would fail with the unexpected success.

muscar:
@DimitrisJim which part of `test_debug_assignment`? Using `setattr` to se the value of `__debug__` works in python:
```python
Python 3.9.7 (default, Sep  3 2021, 12:37:55)
[Clang 12.0.5 (clang-1205.0.22.9)] on darwin
>>> import builtins
>>> setattr(builtins, '__debug__', 5)
>>> builtins.__debug__
5
```

And which of the cases in `test_syntax` do you mean?

I'll have a look at using `__debug__` as a kwarg when calling a function, which doesn't seem to be covered by the PR.

DimitrisJim:
> Using setattr to se the value of `__debug__` works in python:

The `setattr` works but the value of `__debug__` isn't changed (i.e not the value of `builtins.__debug__`, that _is_ changed).

```python
setattr(builtins, "__debug__", "whatever")
__debug__  # still is True
```

This might be more involved as a change though.

> And which of the cases in test_syntax do you mean

It's a couple of the docstrings, you'll need to uncomment the line at the end of `test_syntax` that runs them in order to see which of them fail.

muscar:
@DimitrisJim updated so that the `setattr` on `__debug__` works as in cpython. Also uncommented the relevant tests in `test_compile`. Still not running the docstrings in `test_syntax`, but I _think_ that the extra cases should be covered:

```python
>>>>> __debug__ = 1
SyntaxError: cannot assign to __debug__ at line 1 column 13
__debug__ = 1
            ↑
>>>>> __debug__ += 1
SyntaxError: cannot assign to __debug__ at line 1 column 14
__debug__ += 1
             ↑
>>>>> def foo(__debug__): pass
SyntaxError: cannot assign to __debug__ at line 1 column 1
def foo(__debug__): pass
↑
>>>>> def foo(__debug__=1): pass
SyntaxError: cannot assign to __debug__ at line 1 column 19
def foo(__debug__=1): pass
                  ↑
>>>>> def foo(**kwargs): pass
>>>>> foo(__debug__=1)
SyntaxError: cannot assign to __debug__ at line 1 column 1
foo(__debug__=1)
↑
>>>>> import builtins
>>>>> builtins.__debug__ = 1
SyntaxError: cannot assign to __debug__ at line 1 column 22
builtins.__debug__ = 1
                     ↑
>>>>> setattr(builtins, '__debug__', 'hi!')
>>>>> __debug__
True
```

vm/src/builtins/object.rs
```diff
@@ -323,6 +323,19 @@ pub(crate) fn setattr(
 ) -> PyResult<()> {
     vm_trace!("object.__setattr__({:?}, {}, {:?})", obj, attr_name, value);
 
+    if attr_name.as_str() == "__debug__" {
```
DimitrisJim, line 326: I'd revert these changes. This adds a check for every single look-up and doesn't match behavior expected (the `setattr` is supposed to succeed in changing the value). 

Looking at the disassembly of looking up `__debug__` it seems as if `True` is immediately loaded, indicating no dynamic look-up is performed for `__debug__`. CPython probably detects `__debug__` being looked-up in their peephole optimizer and replaces it with `True` (that's why I said this change is probably a bit involved).

DimitrisJim (changes requested):

vm/src/builtins/object.rs
muscar: Fair enough. Better left for another PR.

compiler/src/compile.rs
```diff
@@ -375,26 +383,39 @@ impl Compiler {
...
+                name
+            ))));
+        }
+        if usage == NameUsage::Delete && is_forbidden_name(name) {
+            return Err(self.error(CompileErrorType::SyntaxError(format!(
+                "cannot delete {}",
+                name
+            ))));
+        }
+        Ok(())
```
youknowone, line 411:
```suggestion
        if usage == NameUsage::Store && is_forbidden_name(name) {
            return Err(self.error(CompileErrorType::SyntaxError(format!(
                "cannot assign to {}",
                name
            ))));
        }
        if usage == NameUsage::Delete && is_forbidden_name(name) {
            return Err(self.error(CompileErrorType::SyntaxError(format!(
                "cannot delete {}",
                name
            ))));
        }
        Ok(())
```

youknowone: @muscar oh, I thought I changed it quickly enough but it wasn't. could you check the newer version of this comment? I am sorry to make confusion.

DimitrisJim:
`test_debug_assignment` should be skipped again (as it previously was), that's causing the CI failure.

DimitrisJim (approved):

PR #3204, opened 2021-10-02
Clean up some skipped tests

fanninpm:
Some skipped tests don't need to be skipped anymore. Submitting as draft until I've combed through our entire CPython test suite.

PR #3208, opened 2021-10-03
Update http library to CPython 3.8

fanninpm:
Companion to #3204.

**NOTE**: I have added `test_nntplib` because an execution path requires it. However, per [PEP 594](https://www.python.org/dev/peps/pep-0594/#nntplib), this module is deprecated as of CPython 3.8 (although the CPython docs are silent about this matter, as far as I've seen). Dropping it would require rewriting the `test.support` module.

### DIFF
--- a/Lib/http/__init__.py
+++ b/Lib/http/__init__.py
@@ -15,6 +15,8 @@ class HTTPStatus(IntEnum):
         * RFC 7238: Permanent Redirect
         * RFC 2295: Transparent Content Negotiation in HTTP
         * RFC 2774: An HTTP Extension Framework
+        * RFC 7725: An HTTP Status Code to Report Legal Obstacles
+        * RFC 7540: Hypertext Transfer Protocol Version 2 (HTTP/2)
     """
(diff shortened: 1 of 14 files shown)


jamestwebber:
> **NOTE**: I have added `test_nntplib` because an execution path requires it. However, per [PEP 594](https://www.python.org/dev/peps/pep-0594/#nntplib), this module is deprecated as of CPython 3.8 (although the CPython docs are silent about this matter, as far as I've seen)

PEP 594 was written a while ago and hasn't been approved...I think it will be eventually but they've been bike-shedding/yak-shaving/etc for years. I think the 3.8 deprecation was a suggestion in that PEP, not a decision that had been made. If the PEP is approved they'll change the schedule for everything.

DimitrisJim:
@fanninpm is this PR relevant at this point (considering it targets 3.8) or should we close it?

fanninpm:
I think I should re-make this PR for CPython 3.11.

(Side note: `nntplib` is slated to be deprecated in CPython 3.13)

PR #3210, opened 2021-10-03
Fix windows ssl

coolreader18:


stdlib/src/ssl.rs
```diff
@@ -119,9 +136,16 @@ fn obj2txt(obj: &Asn1ObjectRef, no_name: bool) -> Option<String> {
...
+        let buflen = buflen as usize;
+        let mut buf = Vec::<u8>::with_capacity(buflen + 1);
+        let ret = sys::OBJ_obj2txt(
+            buf.as_mut_ptr() as *mut libc::c_char,
+            buf.capacity() as _,
+            ptr,
+            no_name,
+        );
         assert!(ret >= 0);
+        buf.set_len(buflen);
```
youknowone, line 148:
```suggestion
        // SAFETY: set_len is safe when capacity is enoguh and all values are already initialized
        buf.set_len(buflen);
```

stdlib/src/ssl.rs
```diff
@@ -992,6 +1021,33 @@ fn convert_ssl_error(
     vm.new_exception_msg(cls, msg.to_owned())
 }
 
+fn x509_stack_from_der(der: &[u8]) -> Result<Vec<X509>, ErrorStack> {
+    unsafe {
+        openssl::init();
```
youknowone, line 1026: does `openssl::init()` is designed to be called whenever this function is called?

coolreader18: Most of rust-openssl's functions call `init()` at the start of the function, e.g. `X509::stack_from_pem`'s first line calls it, since it's invalid to use the openssl api without calling it. Obviously we wouldn't get to the point since we call it in `make_module` but I figured since it's a more generic function it would be good to follow their convention.

youknowone (approved):

Issue #3217, opened 2021-10-04
Fix collection(s) repr to correctly handle subclasses


DimitrisJim:
In CPython:

```python
>>> from collections import deque
>>> class MyDeque(deque): pass
>>> MyDeque()
'MyDeque([])'
```

While in RustPython:

```python
>>>>> from collections import deque
>>>>> class MyDeque(deque): pass
>>>>> MyDeque()
deque([])
```

This needs to be applied for `set`, `frozenset` and `bytearray` too.

Prior similar work: #3137

Snowapril:
currently, rustpython's `repr` is hardcoded with its name. maybe fix it with `slots.name` can be work.

muscar:
I think this addresses the issue: https://github.com/RustPython/RustPython/pull/3219/files.

DimitrisJim:
Please request to work on issues and then someone can assign you. #3218 and #3219 are in conflict. @muscar if possible, alter your PR to only fix the classes other than deque. This way both PRs can be merged!

Please do request things though, it makes it easier for everyone involved :smile:

youknowone:
well, claims will be respected but generally it is not necessary. (at least, we were not using that workflow before.)

If there is no claim, first PR will be preferred to be merged without special reasons.

DimitrisJim:
> it is not necessary

Oh sure, I'm not saying otherwise. Requesting things just makes it easier and doesn't result in one of the PRs being closed (and possibly disheartening new contributors).

muscar:
Fair enough :) Will as on any future ones.

PR #3218, opened 2021-10-04
Fix deque `repr` to print correct format

lijm1358:
This fixes `PyDeque.repr` in #3217 

**before**
```python
>>>>> class MyDeque(deque): pass
>>>>> MyDeque()
deque([])
```

**after**
```python
>>>>> class MyDeque(deque): pass
>>>>> MyDeque()
MyDeque([])
```

### DIFF
--- a/vm/src/stdlib/collections.rs
+++ b/vm/src/stdlib/collections.rs
@@ -420,7 +420,12 @@ mod _collections {
                     .maxlen
                     .map(|maxlen| format!(", maxlen={}", maxlen))
                     .unwrap_or_default();
-                format!("deque([{}]{})", elements.into_iter().format(", "), maxlen)
+                format!(
+                    "{}([{}]{})",
+                    zelf.class().name(),
+                    elements.into_iter().format(", "),
+                    maxlen
+                )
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/collections.rs
```diff
@@ -420,7 +420,12 @@ mod _collections {
                     .maxlen
                     .map(|maxlen| format!(", maxlen={}", maxlen))
                     .unwrap_or_default();
-                format!("deque([{}]{})", elements.into_iter().format(", "), maxlen)
+                format!(
+                    "{}([{}]{})",
+                    zelf.class().slot_name(),
```
youknowone, line 425: is this not `zelf.class().name()` but `zelf.class().slot_name()`?

lijm1358: First thought `name()` and `slot_name()` showed the same value, but `name()` returning `self.slot_name().rsplit('.').next().unwrap().to_string()` looks more good to me now. Thanks for reviewing!

youknowone (approved):
thank you for contributing!

PR #3219, opened 2021-10-04
fix: __repr__ use collection subclass name

muscar:
Addresses: https://github.com/RustPython/RustPython/issues/3217

This patch changes the `set`, `frozenset`, `deque` and `bytearray`
collections to use the subclass name in `__repr__`, if they are
subclassed.

### DIFF
--- a/vm/src/builtins/set.rs
+++ b/vm/src/builtins/set.rs
@@ -198,14 +198,17 @@ impl PySetInner {
         }
     }
 
-    fn repr(&self, vm: &VirtualMachine) -> PyResult<String> {
+    fn repr(&self, class_name: Option<&str>, vm: &VirtualMachine) -> PyResult<String> {
         let mut str_parts = Vec::with_capacity(self.content.len());
         for key in self.elements() {
             let part = vm.to_repr(&key)?;
             str_parts.push(part.as_str().to_owned());
         }
-
-        Ok(format!("{{{}}}", str_parts.join(", ")))
+        let inner_repr = format!("{{{}}}", str_parts.join(", "));
+        if let Some(name) = class_name {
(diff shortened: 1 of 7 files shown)


vm/src/builtins/set.rs
```diff
@@ -473,12 +473,18 @@ impl PySet {
 
     #[pymethod(magic)]
     fn repr(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult {
+        let class_name = zelf.class().name();
         let s = if zelf.inner.len() == 0 {
-            "set()".to_owned()
+            format!("{}()", class_name)
         } else if let Some(_guard) = ReprGuard::enter(vm, zelf.as_object()) {
-            zelf.inner.repr(vm)?
+            let inner_repr = zelf.inner.repr(vm)?;
```
youknowone, line 480: inner_repr can be very long. in that case, using it as argument for another formatting will cost a lot. Could you change `PySetInner::repr` to take class name like below?

```rust
fn repr(&self, class_name: Option<&str>, vm: &VirtualMachine) -> PyResult<String>
```

This approach can be used for other reprs too.

vm/src/builtins/set.rs
```diff
@@ -198,14 +198,17 @@ impl PySetInner {
...
             let part = vm.to_repr(&key)?;
             str_parts.push(part.as_str().to_owned());
         }
-
-        Ok(format!("{{{}}}", str_parts.join(", ")))
+        let inner_repr = format!("{{{}}}", str_parts.join(", "));
+        if let Some(name) = class_name {
+            return Ok(format!("{}({})", name, inner_repr));
+        };
+        Ok(inner_repr)
```
youknowone, line 211: still double formatting when class_name is Some

PR #3223, opened 2021-10-05
Fix test_chr by raising a ValueError on chr(-1) instead of OverflowError

pheki:
`test_chr` used to fail on `chr(-1)` because it would fail to convert the -1 to u32 in `try_to_primitive`, raising an OverflowError.

By changing the received value to a `BigInt`, we're able to match [chr's docs](https://docs.python.org/3.8/library/functions.html#chr) and always raise a `ValueError` if the value is out of range. Note that CPython in turn may raise an `OverflowError` if the argument is too high (platform dependent as it depends on int size), even if it's not documented anywhere.

Another option would be to change the received value to `i32`.

### DIFF
--- a/vm/src/stdlib/builtins.rs
+++ b/vm/src/stdlib/builtins.rs
@@ -35,7 +35,7 @@ mod builtins {
         IdProtocol, ItemProtocol, PyArithmeticValue, PyClassImpl, PyObjectRef, PyRef, PyResult,
         PyValue, TryFromObject, TypeProtocol, VirtualMachine,
     };
-    use num_traits::{Signed, Zero};
+    use num_traits::{Signed, ToPrimitive, Zero};
 
     #[pyfunction]
     fn abs(x: PyObjectRef, vm: &VirtualMachine) -> PyResult {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/builtins.rs
```diff
@@ -87,8 +89,8 @@ mod builtins {
     }
 
     #[pyfunction]
-    fn chr(i: u32, vm: &VirtualMachine) -> PyResult<String> {
-        match std::char::from_u32(i) {
+    fn chr(i: BigInt, vm: &VirtualMachine) -> PyResult<String> {
```
DimitrisJim, line 92: A `PyInt` probably better matches what CPython does.

youknowone: Oh, I didn't recognize this is added. `TryFromObject for BigInt` cannot avoid clone. So it will mislead cost. `PyInt` must be used.
I will remove TryFromObject for BigInt.

youknowone: #3224

pheki: Done! Changed to a PyIntRef

PR #3233, opened 2021-10-06
Clean up overflow cases.

DimitrisJim:
by using `try_to_primitive`. There's probably other cases where the error is explicitly handled yet.

### DIFF
--- a/stdlib/src/bisect.rs
+++ b/stdlib/src/bisect.rs
@@ -6,7 +6,6 @@ mod _bisect {
         function::OptionalArg, slots::PyComparisonOp::Lt, ItemProtocol, PyObjectRef, PyResult,
         VirtualMachine,
     };
-    use num_traits::ToPrimitive;
     use std::convert::TryFrom;
 
     #[derive(FromArgs)]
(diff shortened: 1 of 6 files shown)


stdlib/src/bisect.rs
```diff
@@ -26,11 +25,7 @@ mod _bisect {
         vm: &VirtualMachine,
     ) -> PyResult<Option<isize>> {
         Ok(match arg {
-            OptionalArg::Present(v) => {
-                Some(vm.to_index(&v)?.as_bigint().to_isize().ok_or_else(|| {
-                    vm.new_index_error("cannot fit 'int' into an index-sized integer".to_owned())
-                })?)
-            }
+            OptionalArg::Present(v) => Some(vm.to_index(&v)?.try_to_primitive(vm)?),
```
DimitrisJim, line 28: was actually `OverflowError` and not `IndexError`.

vm/src/stdlib/collections.rs
```diff
@@ -90,19 +88,15 @@ mod _collections {
             // need to move that function elsewhere and refactor usages.
             let maxlen = if let Some(obj) = maxlen.into_option() {
                 if !vm.is_none(&obj) {
-                    let value = obj.payload::<PyInt>().ok_or_else(|| {
-                        vm.new_value_error("maxlen must be non-negative.".to_owned())
```
DimitrisJim, line 94: Not related but fixed this slight mistake too.

PR #3235, opened 2021-10-06
Add length check, `reduce`, `setstate` on zip from cpython 3.10

Snowapril:
This revision implements optional length-checking to `zip` and its related new methods (`__reduce__`, `__setstate__`).

This is new features in cpython 3.10 stable-release.
If `strict=True` option is given to `zip`, all the iterators to be zipped must have same length.
```python
>>>>> a = [1,2,3]
>>>>> b = (4,5)
>>>>> for k in zip(a,b,strict=True):
.....   print(k)
..... 
(1, 4)
(2, 5)
ValueError: zip() argument 2 is shorter than argument 1

>>>>> for k in zip(b,a,strict=True):
.....   print(k)
..... 
(4, 1)
(5, 2)
ValueError: zip() argument 2 is longer than argument 1
```

also implement `__setstate__` and `__reduce__` for the `strict` parameter in `zipobject`.
```python
>>>>> z = zip([1,2,3], (4,5,6), strict=True)
>>>>> z.__reduce__()
(<class 'zip'>, (<list_iterator object at 0x557d82f1b9f0>, <tuple_iterator object at 0x557d82df3a30>), True)
>>>>> z.__setstate__(False)
>>>>> z.__reduce__()
(<class 'zip'>, (<list_iterator object at 0x557d82f1b9f0>, <tuple_iterator object at 0x557d82df3a30>))
```
---
### Documents
* https://www.python.org/dev/peps/pep-0618/
* https://github.com/python/cpython/pull/20921

### DIFF
--- a/vm/src/builtins/zip.rs
+++ b/vm/src/builtins/zip.rs
@@ -1,15 +1,17 @@
 use super::PyTypeRef;
 use crate::{
-    function::PosArgs,
+    function::{OptionalArg, PosArgs},
     protocol::{PyIter, PyIterReturn},
     slots::{IteratorIterable, SlotConstructor, SlotIterator},
     PyClassImpl, PyContext, PyRef, PyResult, PyValue, VirtualMachine,
 };
+use rustpython_common::atomic::{self, PyAtomic, Radium};
 
 #[pyclass(module = false, name = "zip")]
 #[derive(Debug)]
 pub struct PyZip {
     iterators: Vec<PyIter>,
+    strict: PyAtomic<bool>,
 }
 
 impl PyValue for PyZip {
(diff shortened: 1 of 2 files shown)


vm/src/builtins/zip.rs
```diff
@@ -1,15 +1,19 @@
...
+    IntoPyObject, PyClassImpl, PyContext, PyObjectRef, PyRef, PyResult, PyValue, TryFromObject,
+    TypeProtocol, VirtualMachine,
 };
+use crossbeam_utils::atomic::AtomicCell;
 
 #[pyclass(module = false, name = "zip")]
 #[derive(Debug)]
 pub struct PyZip {
     iterators: Vec<PyIter>,
+    strict: AtomicCell<bool>,
```
DimitrisJim, line 16: Use a `PyAtomic<bool>` here. To get the same semantics as `AtomicCell` use `Acquire` ordering on loads and `Release` on stores.

DimitrisJim: Doesn't use atomics when threading is disabled.

PR #3237, opened 2021-10-06
Restore test_cmath to original form.

DimitrisJim:


stdlib/src/cmath.rs
```diff
@@ -98,10 +98,15 @@ mod cmath {
     ///
     /// If the base not specified, returns the natural logarithm (base e) of z.
     #[pyfunction]
-    fn log(z: IntoPyComplex, base: OptionalArg<IntoPyFloat>) -> Complex64 {
+    fn log(z: IntoPyComplex, base: OptionalArg<IntoPyComplex>) -> Complex64 {
+        // TODO: Complex64.log with a negative base yields wrong results.
```
DimitrisJim, line 102: I'll probably open an issue on `num_complex` for this. Until then this could _probably_ be resolved (and some issues with `PyComplex` too) if division is implemented manually (the algorithm for it didn't seem too difficult) without relying on `Complex64.fdiv`.

youknowone (approved):

PR #3240, opened 2021-10-07
Add numlike for handling of numlike arguments

DimitrisJim:
Idea is similar to other branches for byteslike/argument, i.e aggregate function argument handling in the `function` folder and normalize names (using `Arg<Type>Like`). Essentially these types all signify that some magic method gets invoked on a Python object in order to create the type (`__iter__`, `__bytes__`, `__float__`, `__complex__`, `__bool__`, etc)

@youknowone you've pushed the original work for function/argument/byteslike, so I'll wait for the green light from you before I add the rest of the types.

### DIFF
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -17,7 +17,7 @@ mod array {
         },
         class_or_notimplemented,
         function::{
-            ArgBytesLike, ArgFloatLike, ArgIterable, IntoPyObject, IntoPyResult, OptionalArg,
+            ArgBytesLike, ArgIntoFloat, ArgIterable, IntoPyObject, IntoPyResult, OptionalArg,
         },
         protocol::{
             BufferInternal, BufferOptions, BufferResizeGuard, PyBuffer, PyIterReturn,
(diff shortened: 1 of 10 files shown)


youknowone:
The changes looks great, but I have a question about the name. Does -`Like` suffix fit? When we handle `ArgBytesLike`, it is not strictly a `bytes` because it hold its original data type on its `PyBuffer` backend. For these numbers, it doesn't hold its original type but fully transformed into a value. I feel like the original `Into` naming shows its mechanism better - except for `Py` part because it turns into just number not a `PyValue`.
How about `ArgIntoBool`, `ArgIntoComplex` etc?

DimitrisJim:
Yeah, that makes sense to me.

DimitrisJim:
I'm leaving `PyInt` from now since I'm still thinking how best it would be transformed (it could either target `BigInt`s and/or `usize/isize`). Maybe two distinct types would be better `IntoInt (BigInt)` and `IntoIndex (isize)` but I'm still uncertain.

PR #3241, opened 2021-10-07
Refactor list avoid duplicate the vec

qingshi163:
This is like another appaorch to replace #2933 

This will avoid duplicate the vec before the loop.

This will have compatible result when changing the list while looping.

This may have more overhead as we lock and unlock the list for each comparison operation.

Possible optimization:
- keeping the lock for internal classes if we can identify it.

Please advise if this appoarch is acceptable or do we have better idea?
@youknowone

### DIFF
--- a/vm/src/builtins/list.rs
+++ b/vm/src/builtins/list.rs
@@ -2,16 +2,15 @@ use super::{PositionIterInternal, PyGenericAlias, PySliceRef, PyTupleRef, PyType
 use crate::common::lock::{
     PyMappedRwLockReadGuard, PyMutex, PyRwLock, PyRwLockReadGuard, PyRwLockWriteGuard,
 };
-use crate::sliceable::saturate_index;
 use crate::{
     function::{ArgIterable, FuncArgs, IntoPyObject, OptionalArg},
     protocol::{PyIterReturn, PyMappingMethods},
     sequence::{self, SimpleSeq},
-    sliceable::{PySliceableSequence, PySliceableSequenceMut, SequenceIndex},
+    sliceable::{saturate_index, PySliceableSequence, PySliceableSequenceMut, SequenceIndex},
     stdlib::sys,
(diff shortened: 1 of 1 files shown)


youknowone:
When @DimitrisJim benchmarked similar stuff, the result showed collecting to vec is a lot cheaper than `n` lock-unlock

DimitrisJim:
Slightly better for bigger lists but slightly worse for smaller ones, which makes sense considering a copy of the vector is made and size plays a big role in how fast that can be done.

qingshi163:
If we could efficiently check does python class customize the eq slot, we could optimize to keep the lock.

qingshi163:
Whatever the case we need to check the slot to call. So if we can identify the function in the slot is rust code, we could avoid one lock unlock, one rc clone for the object for each loop.

youknowone:
do we need `member_descriptor`?

DimitrisJim:
Microbenchmarking shows this executing slightly faster for builtins. With a custom type the old code is faster, checking for a custom `__eq__` would probably help with that.

Overall, this looks ok to me, it seems like it could also generalize for other collections like `deque`s and `array`s.

youknowone:
the benchmark result sounds like we need to change the compare code but not about the locking

qingshi163:
> the benchmark result sounds like we need to change the compare code but not about the locking

optimize the compare code make a big improvment.

on the other hand, the original clone the vec or this pr both have acceptable performence. this one is a little better for builtins but a little worse for custom types.

also as we are checking via list.count that have to go throw the whole list, the other function like list.remove, list.contains will in average return earlier. this appoarch should be slightly better than clone the whole vec.

or maybe we just ignore the compatibility, let who write the code like that just go to the hell :thinking:

youknowone:
ah, got it. I didn't notice `count()` compatibility

PR #3249, opened 2021-10-07
Add rand, ror, rsub operators in dict_items

joohongpark:
Added rand, ror, rsub functions for dict_items object


youknowone:
ref #3248

joohongpark:
Added rand, ror, rsub functions for dict_items object and removed test annotation

Issue #2796, opened 2021-08-05
Recursions in rust code trigger segmentation faults.


DimitrisJim:
Case in point, issues #2783 #2782 #2781 #2780 and #2778.

All these fail due to the same underlying reason, the rust call stack grows until a stack overflow is triggered. RustPython should be able to handle these and raise `RecursionError` instead, as CPython does.

pheki:
Sorry I didn't see this issue!

I can imagine 3 "simple" solutions:

1. Add a `recursion_depth` private property to `VirtualMachine` and increment/decrement it manually on `with_frame` and `with_recursion`. This is the way CPython does it and would require a new `with_recursion` method which would need to be manually called by all potentially recursive rust functions.
2. Add an `external_recursion_depth` private property and increment it only on `with_recursion`, using `vm.frames.len() + vm.external_recursion_depth` to calculate the current recursion depth. This has the same caveat as 1 of needing `with_recursion` to be manually called.
3. Measure the current stack depth by actually looking at the stack, possibly using an external library such as [backtrace](https://docs.rs/backtrace/0.3.61/backtrace/fn.trace.html) (see [this stackoverflow issue](https://stackoverflow.com/a/69142325)). This has the downside of being platform dependent and less deterministic.

I personally believe 1 is the way to go right now.

DimitrisJim:
The complexity here appears when you hit the limit since you'd like to handle the `RecursionError` gracefully when you've possibly run out of stack space. I think there's a crate called `stacker` that might help with that (but this can be a follow up issue after basic handling of recursion is added).

The first solution is definitely the best approach, an attribute macro that basically does what `Py_EnterRecursiveCall`/`Py_LeaveRecursiveCall` would be cool.

changjoon-park:
Filed #7655 with current (2026) reproducers and evidence that this is parser-layer: `compile(src, '<s>', 'exec')` (which does NOT go through `_ast::ast_to_object`) and `ast.parse()` crash at the identical depth threshold (~4000), showing the shared component is `rustpython-ruff_python_parser` itself.

Also note: #4857 (closed as fixed in 2023) still reproduces on current main — its original `ast.parse('+chr(33)' * 1000000)` SIGSEGVs.

Prior fixes #7630 (`_ast::ast_from_object`) and #7632 (`json.loads`) addressed downstream symptoms; the parser's own recursive descent remains unguarded.

Issue #2778, opened 2021-08-02
Rust Python gets crashing when rebinding __repr__ as __str__


xiaxinmeng:
Rust Python interpreter crashes when rebinding __repr__ as __str__. 

```
class Foo(object):
    pass

Foo.__repr__ = Foo.__str__

foo = Foo()
str(foo)
```

Actual output:

> Segmentation fault (core dumped)

System:

> Ubuntu 16.04
> Rust Python 0.1.2 interpreter

PR #3252, opened 2021-10-08
Make it possible for rust functions to increase recursion depth

pheki:
Use a new recursion_depth property on VirtualMachine to make it possible to detect deep recursion (`RecursionError`) on rust code, as discussed in #2796.

The `#[recursive]` attribute should be added to recursive functions. A `VirtualMachine` argument must be present in the function, if it's not named `vm`, then the name of the vm argument should be specified as an input to the macro, such as `#[recursive(vm = "self")]` in the `VirtualMachine` methods (probably the only place where it'll be needed).

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,0 +1,8 @@
+#
+# This list is used by git-shortlog to aggregate contributions.  It is
+# necessary when either the author's full name is not always written
+# the same way, and/or the same author contributes from different
+# email addresses.
+#
+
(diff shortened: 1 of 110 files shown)


DimitrisJim:
It appears that #2778 isn't tested anywhere (but it behaves correctly after this pr). Could you add a `recursion.py` file in `extra_tests/snippets` that exercises this case and verifies a `RecursionError` is raised?

pheki:
I believe I've commited all that's needed to be done ;)

PR #3256, opened 2021-10-08
Add `_statistics` module containing `_normal_dist_inv_cd`

### LINKED ISSUE #3251: Add `_statistics` module containing `_normal_dist_inv_cdf`
This should be _relatively_ easy. `_statistics` needs to be defined as a built-in module and `_normal_dist_inv_cdf` should be added as its sole function.

`_normal_dist_inv_cdf` is pretty straight-forward to implement with the [C version](https://github.com/python/cpython/blob/main/Modules/_statisticsmodule.c#L29) being a pretty direct translation from the [Python version](https://github.com/RustPython/RustPython/blob/main/Lib/statistics.py#L825).

aDotInTheVoid:
Closes #3251

Not sure if this needs a test, or if you run the python ones in `Lib/test/test_statistics.py` is enough

### DIFF
--- a/stdlib/src/lib.rs
+++ b/stdlib/src/lib.rs
@@ -20,11 +20,11 @@ mod platform;
 mod pyexpat;
 mod pystruct;
 mod random;
+mod statistics;
 // TODO: maybe make this an extension module, if we ever get those
 // mod re;
 #[cfg(not(target_arch = "wasm32"))]
 pub mod socket;
-mod statistics;
 #[cfg(unix)]
 mod syslog;
 mod unicodedata;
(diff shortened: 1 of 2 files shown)


aDotInTheVoid:
> In addition to that, no need to define function outside and then use super here.

I thing this is good as one function handles the stats, and the other handles the arg conversion and error.

The tests now pass

youknowone:
@DimitrisJim how this one has going? is this need to be rebased?

fanninpm:
Sorry, I forgot that this PR existed when I trampled on it with #3552.

youknowone:
This version is better structured than current version. I rebased this version on the main branch.

youknowone:
@aDotInTheVoid I am sorry for making you wait for this for so long time. I didn't know this is untracked.

PR #3257, opened 2021-10-09
Add rxor, rand, ror and rsub operators in dict_keys

Dolpheyn:
ref #3249

Added `PyDictKeys::{xor, and, or, sub}` and removed expected failure test annotation from `test_dictview_set_operations_on_keys`.


Dolpheyn:
> looks great, thank you for contributing!

You're welcome!

Issue #3232, opened 2021-10-06
make PyType::name return type `&str`


youknowone:
`PyType::name()` is a commonly used internal api.
Even though it can return `&str`, it returns a `String` for now.

Snowapril:
As it is annotated with `pyproperty`, it cant return &str directly. Is it better to make new method, such as `_name`, which return &str and use it in the other codes?

youknowone:
Yes, but I think `name()` for `&str` and `__name__` for `#[pyproperty]` is better.

Snowapril:
Oh gosh. it is cool 😊 Could I work on this issue?

DimitrisJim:
closed by #3261

PR #3261, opened 2021-10-10
Modify `PyType::name` method to return `BorrowedValue<str>`

Snowapril:
This will close issue #3232


youknowone:
Thank you for contributing!

PR #3263, opened 2021-10-10
Implement keyword suggestion on `PyAttributeError` and `PyNameError` from cpython 3.10

Snowapril:
This revision implements keyword suggestion on `PyAttributeError` and `PyNameError`

#### RustPython
```python
>>>>> collections.namedtoplo
  File "/home/snowapril/RustPython/vm/pylib-crate/Lib/collections/__init__.py", line 1, in __getattr__
    '''This module implements specialized container datatypes providing
namedtuple
AttributeError: module 'collections' has no attribute 'namedtoplo'. Did you mean: 'namedtuple'?

>>>>> schwarzschild_black_hole = None
>>>>> schwarschild_black_hole
schwarzschild_black_hole
NameError: name 'schwarschild_black_hole' is not defined. Did you mean: 'schwarzschild_black_hole'?
```
This is new feature in cpython 3.10 ([link](https://docs.python.org/3/whatsnew/3.10.html#attributeerrors))

### Documents
* https://bugs.python.org/issue38530
* https://docs.python.org/3/whatsnew/3.10.html#attributeerrors
* https://docs.python.org/3/whatsnew/3.10.html#nameerrors

### DIFF
--- a/vm/src/pyobject.rs
+++ b/vm/src/pyobject.rs
@@ -1047,7 +1047,7 @@ impl PyMethod {
         let getattro = cls.mro_find_map(|cls| cls.slots.getattro.load()).unwrap();
         if getattro as usize != object::PyBaseObject::getattro as usize {
             drop(cls);
-            return getattro(obj, name, vm).map(Self::Attribute);
+            return obj.get_attr(name, vm).map(Self::Attribute);
         }
 
         let mut is_method = false;


youknowone:
`rustpython_common::levenshtein` would be better place because it is not a common string operation.
Any problem with `MOVE_COST` const? It still can be a const and exported via `use rustpython_common::levenshtein::MOVE_COST`.

Snowapril:
@youknowone how about make `levenshtein` mod in `str` like `rustpython_common::str::levenshtein`. As `levenshtein` itself is not much common algorithm and it is actually a string operation.

coolreader18:
@Snowapril that makes sense to me.

youknowone:
Great, thank you for long time effort!

PR #3272, opened 2021-10-11
added gethostbyname_ex function

varun-krishnan:
Implements gethostbyname_ex function 

Issue: https://github.com/RustPython/RustPython/issues/1176

Function overview:
https://pythontic.com/modules/socket/gethostbyname_ex


varun-krishnan:
I see that there is a test in test_socket.py that uses this function, not sure, if i need to add one?

PR #3273, opened 2021-10-11
migrate TryIntoRef to IntoPyStrRef 

youknowone:
For downcasting from `PyObjectRef`, `PyObjectRef::try_into_value` covers it. So it is duplication of features.

For converting to `PyValue`, implenting `Into` works well. Maybe we could add `IntoPyRef` for same usage later.

In spite of these reasons, `IntoPyRef<PyStr>` is widely used for convenient accessing to python maps. So I am keeping it for now.


qingshi163 (approved):
I think it is reasonable to keep IntoPyStrRef as in rust this is normal.

coolreader18 (approved):

PR #3290, opened 2021-10-13
Modify range in pytype.rs

tony-jinwoo-ahn:
For readability, replace exclusive range with inclusive range.


DimitrisJim:
Thanks! Please run `cargo fmt --all` to fix the formatting.

Issue #3244, opened 2021-10-07
Implement protocols (Abstract obejcts layer)


youknowone:
- [x] Object protocol
  - https://docs.python.org/3/c-api/object.html
  - This protocol will not be a type or trait. `impl PyObjectRef` in `vm/src/protocol/object.rs` will be enuogh
  - [x] Py_RETURN_NOTIMPLEMENTED (PyArithmeticValue::NotImplemented)
  - [ ] PyObject_Print
  - [x] PyObject_HasAttr / PyObject_HasAttrString (has_attr)
  - [x] PyObject_GetAttr / PyObject_GetAttrString (get_attr)
  - [x] PyObject_GenericGetAttr
  - [x] PyObject_SetAttr / PyObject_SetAttrString (set_attr)
  - [x] PyObject_GenericSetAttr
  - [x] PyObject_DelAttr / PyObject_DelAttrString (del_attr)
  - [ ] PyObject_GenericGetDict
  - [ ] PyObject_GenericSetDict
  - [x] PyObject_RichCompare (rich_compare)
  - [x] PyObject_RichCompareBool (rich_compare_bool)
  - [x] PyObject_Repr (repr)
  - [x] PyObject_ASCII (ascii)
  - [x] PyObject_Str (str)
  - [x] PyObject_Bytes (bytes)
  - [x] PyObject_IsSubclass (is_subclass)
  - [x] PyObject_IsInstance (is_instance)
  - [x] PyObject_Hash (hash)
    - [x] PyObject_HashNotImplemented  #4640
  - [x] PyObject_IsTrue (is_true)
  - [x] PyObject_Not (not)
  - [x] PyObject_Type #5091
  - [x] PyObject_TypeCheck #5091
  - [x] PyObject_Size / PyObject_Length (length)
  - [x] PyObject_LengthHint (length_hint)
  - [x] PyObject_GetItem (get_item)
  - [x] PyObject_SetItem (set_item)
  - [x] PyObject_DelItem (del_item)
  - [ ] PyObject_Dir
  - [x] PyObject_GetIter (get_iter)
  - [x] PyObject_GetAIter #5090
- [ ] Call protocol
  - https://docs.python.org/3/c-api/call.html
  - [x] `Callable` trait to provide type slot `call`. 
  - [ ] `PyCallable` protocol object to provide interface (at least check) #4654
  - [ ] Vectercall protocol
- [ ] Number protocol
  - https://docs.python.org/3/c-api/number.html
[...]

PR #3306, opened 2021-10-14
Object protocol

youknowone:
I also suggest to replace the paired vm methods and use PyObjectRef methods

ref #3244

### DIFF
--- a/vm/src/bytesinner.rs
+++ b/vm/src/bytesinner.rs
@@ -38,11 +38,11 @@ impl TryFromBorrowedObject for PyBytesInner {
 #[derive(FromArgs)]
 pub struct ByteInnerNewOptions {
     #[pyarg(any, optional)]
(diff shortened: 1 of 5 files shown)


DimitrisJim:
should `ItemProtocol` be moved here too?

youknowone:
Probably. I doub't we really need that trait now. I think relocating `impl ItemProtocol for PyObjectRef` to here, `impl ItemProtocol for PyDictRef` to its `impl` and removing the trait could be better than now.

DimitrisJim:
The `delitem/setitem/getitem` in `PyDict` would need to also be made public so objects that are known to be dicts to be able to call directly into them.

youknowone:
@DimitrisJim could you review it? I want to open issue for relocating functino bodies and refactoring around them.

PR #3307, opened 2021-10-14
PyObjectPtr: no RC even through function pointers

youknowone:
it will enhance performance for type slots by removing double dereference, hopefully.

it also can be used like deferred rc in limited cases

### DIFF
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -28,8 +28,8 @@ mod array {
             AsBuffer, AsMapping, Comparable, Constructor, IterNext, IterNextIterable, Iterable,
             PyComparisonOp,
         },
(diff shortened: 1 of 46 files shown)


coolreader18:
I'd prefer a lifetime to be attached to the type, to avoid use after free. Is there any reason there can't be one?

coolreader18:
Just `PyObjectRef::from_raw`. Like `to_owned` for it could look like `let m = ManuallyDrop::new(PyObjectRef::from_raw(self as *const _ as *const PyObjectRaw)); m.clone()`. Should I make a branch of what I was thinking?

youknowone:
@coolreader18 could you help it please?

youknowone:
`incref` is not a rust concept. So removed it and replaced it with `to_owned`. The name `Py` is too much generic. Any suggestion? It must be used only as reference.

youknowone:
@DimitrisJim I agree they must be documented better.

PR #3310, opened 2021-10-14
as_mapping never fails

youknowone:
@Snowapril could you review this PR? I think `as_mapping` never fails because it doesn't call any user function but just retuning function pointers.


Snowapril (approved):
This is great! At initial thought, I was meaning that with `PyResult` I can represent whether implementing `as_mapping` or not. But `slot` itself is enough to that stuff. Looks good to me 😊

DimitrisJim (approved):

Issue #158, opened 2018-10-15
How to deal with C-API extensions?


windelbouwman:
Create a prototype on how to interact with a python library with a C-API extension.

See also:
https://pythoncapi.readthedocs.io/

There are some questions regarding C extensions. Primary question: how to handle this? Maybe in a similar way to cpyext?

windelbouwman:
Just wanted to bump this issue with a link to handlepy: https://github.com/pyhandle/hpy this might be the way to go forwards.

ShadowJonathan:
hpy is definitely the way, it's getting increasing support and attention from the python community

arihant2math:
I've started with an initial implementation with #5604. Currently I haven't found a way to properly expose `PyObject`, but I don't see that being a massive issue for now.

bschoenmaeckers:
> I've started with an initial implementation with [#5604](https://github.com/RustPython/RustPython/pull/5604). Currently I haven't found a way to properly expose `PyObject`, but I don't see that being a massive issue for now.

[Pep 803](https://peps.python.org/pep-0803/) makes `PyObject` an opaque pointer. So this might help improve things.

encukou:
Rust wasn't the primary motivation for PEP 803, but I definitely had alternate implementations in mind :)

arihant2math:
This seems interesting, if I have time, I'll take another look at a C-API.

PR #3327, opened 2021-10-16
First version of C-API 

youknowone:
ref #158

The combination of thead vm stack and PyObjectPtr brought me somewhere interesting spot.


Lynskylate:
@youknowone Has any new developments or progress in c-api? After providing basic c-api, maybe we can start to support [hpy](https://github.com/hpyproject/hpy).

PR #3329, opened 2021-10-16
Cleanup lists of tests in ci.yaml

fanninpm:
I deduplicated the list of platform-independent tests and sequestered it to an environment variable for reusability. I then sorted the resulting two lists (platform-independent and Windows-only) and took out duplicate entries of `test_json` and `test_set`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -177,7 +177,6 @@ jobs:
             test_pathlib
             test_pwd
             test_py_compile
-            test_set
             test_shutil
             test_sys
             test_unittest


fanninpm:
Question for the maintainers: would you rather leave the lists as separate lines or fold those lines up into fewer lines?

CryptoGodfatherVA:
Just so everyone knows once again the iphone 7 has been sold so any changes made last night other than on the iphone XS is false and another attempt on hacking my information and accounts . Also if any fx card or coinbase transactions or any transactions come across they are not mines. I have not recieved any progress because of these same issues since day one . Vedarius Russell of Youtac RS/ VARs

PR #3337, opened 2021-10-18
os: fix putenv

deantvv:
Check key and value before calling `env::set_var`

From doc of `env::set_var`,
```
This function may panic if `key` is empty, contains an ASCII equals sign
`'='` or the NUL character `'\0'`, or when the value contains the NUL
character.
```


youknowone:
great!

Issue #3324, opened 2021-10-16
Relocate object protocol implementation from vm 


youknowone:
For now, Object protocol methods are (mostly) thin wrapper of matching vm functions. Because we don't need to have duplicated functions for same features, they need to be moved from vm to object protocol functions.

- `vm` path: `vm/src/vm.rs`
- object protocol path: `vm/src/protocol/object.rs`

The matching vm functions are easy to find because object methods are calling it.

- [x] get_attr #3331
- [x] set_attr #3338
- [x] del_attr #3341
- [x] rich_compare #3347
- [x] repr #3364
- [x] str #3349 
- [x] is_subclass #3345
- [x] is_instance #3344 
- [x] hash #3342
- [x] length #3343

Becasue relocating each function will cause many code editing, every new PR for each function is recommended for easier review.

Reference:
- #3306: The original patch to add object protocol

AP2008:
So, that function can be called from `vm/src/protocol/object.rs` as `vm.call_special_method` ?

youknowone:
this is done! thank you everyone who contributed in this issue!

PR #3345, opened 2021-10-19
Relocate vm.issubclass to obj.is_subclass

ChJR:
This PR fixes Task 7 of #3324.


DimitrisJim:
Yeah, I agree with @Snowapril but I'd say we leave it to another PR.

@ChJR can you rebase this?

youknowone:
I think relocating abstract_issubclass and recursive_issubclass is part of this PR. they are internal functions and not expected to be called outside, but visibility is changed to `pub` for now.

Snowapril:
`isinstance` in #3344 also need to be handled later😁

PR #3346, opened 2021-10-19
set repr push to buffer directly

youknowone:
So, everything was wrong.
Allocating parts vector is useless.
Performance regression came from parts.
`format!`s and `join` looks somehow working well.

the best manually written version with parts vector didn't have any difference from original version


youknowone:
Yes, this is fine on both my machines.

DimitrisJim (approved):

PR #3381, opened 2021-10-27
Time mod additions

DimitrisJim:
This adds some needed functionality for running perf scripts on unix based platforms, windows is unfortunately pending since I haven't got a machine.

### DIFF
--- a/Lib/test/test_time.py
+++ b/Lib/test/test_time.py
@@ -60,12 +60,16 @@ class TimeTestCase(unittest.TestCase):
         time.timezone
         time.tzname
 
+    # TODO: RUSTPYTHON
+    @unittest.skipIf(sys.platform == "win32", "Implement get_clock_info for Windows.")
     def test_time(self):
         time.time()
         info = time.get_clock_info('time')
         self.assertFalse(info.monotonic)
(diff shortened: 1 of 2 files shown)


youknowone (approved):

coolreader18:
Not sure if this is still relevant, but my idea for implementing monotonic time was to have a `OnceCell<Instant>` that gets initialized with Instant::now at first use and then it acts roughly as processtime. Just since Instant has already had a bunch of effort put into making sure it's actually monotonic

DimitrisJim:
Looking at the docs it appears Instant calls the same clock for unices, `clock_gettime(CLOCK.MONOTONIC)`. For Windows, Instant calls `QueryPerformanceCounter` (which CPython uses for `perf_counter` iirc) instead of `GetTickCount64`.

Admittedly, I wasn't aware of Instant but, since we expose all clock constants and `clock_gettime`, is there any point to using Instant instead of forwarding the call ourselves?

coolreader18:
> is there any point to using Instant instead of forwarding the call ourselves?

Nah, probably not, we'll just need to implement clock_gettime for wasm in browsers

PR #3383, opened 2021-10-27
Fix warnings for wasi check.

DimitrisJim:


youknowone (approved):

youknowone:
ignoring clippy warnings for wasi would be acceptable

PR #3387, opened 2021-10-28
Get (most of) test_type_name passing

OddBloke:
See individual commits for details.

### DIFF
--- a/Lib/test/test_builtin.py
+++ b/Lib/test/test_builtin.py
@@ -2052,8 +2052,6 @@ class TestType(unittest.TestCase):
         with self.assertRaises(TypeError):
             type('a', (), dict={})
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_type_name(self):
         for name in 'A', '\xc4', '\U0001f40d', 'B.A', '42', '':
             with self.subTest(name=name):
@@ -2063,8 +2061,10 @@ class TestType(unittest.TestCase):
                 self.assertEqual(A.__module__, __name__)
         with self.assertRaises(ValueError):
             type('A\x00B', (), {})
-        with self.assertRaises(ValueError):
-            type('A\udcdcB', (), {})
+        # TODO: RUSTPYTHON (https://github.com/RustPython/RustPython/issues/935)
+        with self.assertRaises(AssertionError):
+            with self.assertRaises(ValueError):
+                type('A\udcdcB', (), {})
         with self.assertRaises(TypeError):
             type(b'A', (), {})
 
@@ -2080,9 +2080,13 @@ class TestType(unittest.TestCase):
         with self.assertRaises(ValueError):
             A.__name__ = 'A\x00B'
         self.assertEqual(A.__name__, 'C')
-        with self.assertRaises(ValueError):
-            A.__name__ = 'A\udcdcB'
-        self.assertEqual(A.__name__, 'C')
+        # TODO: RUSTPYTHON (https://github.com/RustPython/RustPython/issues/935)
+        with self.assertRaises(AssertionError):
+            with self.assertRaises(ValueError):
+                A.__name__ = 'A\udcdcB'
+            self.assertEqual(A.__name__, 'C')
+        # TODO: RUSTPYTHON: the previous __name__ set should fail but doesn't: reset it
+        A.__name__ = 'C'
         with self.assertRaises(TypeError):
             A.__name__ = b'A'
         self.assertEqual(A.__name__, 'C')
--- a/vm/src/builtins/pytype.rs
+++ b/vm/src/builtins/pytype.rs
@@ -491,6 +491,10 @@ impl PyType {
             }
         }
 
+        attributes
+            .entry("__qualname__".to_string())
+            .or_insert_with(|| vm.ctx.new_str(name.as_str()).into());
+
         // All *classes* should have a dict. Exceptions are *instances* of
         // classes that define __slots__ and instances of built-in classes
         // (with exceptions, e.g function)
--- a/vm/src/builtins/pytype.rs
+++ b/vm/src/builtins/pytype.rs
@@ -336,7 +336,6 @@ impl PyType {
 
     #[pyproperty(magic)]
     pub fn module(&self, vm: &VirtualMachine) -> PyObjectRef {
-        // TODO: Implement getting the actual module a builtin type is from
         self.attributes
             .read()
             .get("__module__")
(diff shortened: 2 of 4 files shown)


Lib/test/test_builtin.py
```diff
@@ -2063,8 +2061,9 @@ def test_type_name(self):
                 self.assertEqual(A.__module__, __name__)
         with self.assertRaises(ValueError):
             type('A\x00B', (), {})
-        with self.assertRaises(ValueError):
-            type('A\udcdcB', (), {})
+        # TODO: RUSTPYTHON (https://github.com/RustPython/RustPython/issues/935)
+        # with self.assertRaises(ValueError):
+        #    type('A\udcdcB', (), {})
```
youknowone, line 2066: We always prefer `expectedFailure` decorator than commenting out. (Guess what happens when this is fixed later)

youknowone:
@moreal could you please review this PR? the `slot_name` looks different than your idea before. So I want to listen your opinion about it.

vm/src/builtins/pytype.rs
```diff
@@ -282,11 +282,20 @@ impl PyType {
 
     pub fn name(&self) -> BorrowedValue<str> {
         PyRwLockReadGuard::map(self.slots.name.read(), |slot_name| {
-            slot_name.as_ref().unwrap().rsplit('.').next().unwrap()
+            slot_name.as_ref().unwrap().as_str()
```
moreal, line 285: In my thought, this change should be reverted because it makes a break between RustPython and CPython.

There is an example to check it. You can run `binascii.Incomplete.__name__` in two environments.

```python
# CPython
>>> import binascii; binascii.Incomplete.__name__
'Incomplete'

# RustPython
>>>>> import binascii; binascii.Incomplete.__name__
'binascii.Incomplete'
```

You can see the `name` CPython implementation in:

https://github.com/python/cpython/blob/8a310dd5f4242b5d28013c1905c6573477e3b957/Objects/typeobject.c#L489

moreal (changes requested):

Lib/test/test_builtin.py
OddBloke: Yeah, that's a very fair point. In this case, I've fixed all but these failures, which are due to a much broader bug than I can reasonably fix alongside all these other failures. With a decorator, none of these assertions would run, so regressions could occur in all of the ones that I've just fixed. Should I move these assertions to a separate test to decorate? Or perhaps I can use it as a context manager? What do you think?

vm/src/builtins/pytype.rs
OddBloke: Great point, thanks! Looking at that typeobject.c implementation, I think the existing implementation is implementing one of these branches, and I have (incorrectly) *replaced* that implementation with an implementation of the other branch. Each is correct for different cases, which is why the tests pass with my implementation, but your counterexample is broken by it. I'm on my phone at the moment, so I can't dig into it further but: I agree, there's no point landing this change as-is. I'll look into an implementation of *both* branches.

Lib/test/test_builtin.py
OddBloke: OK, I've just pushed up a fix: use `self.assertRaises(AssertionError)` as a context manager around the expected failures: once they're passing, that will fail the test.

vm/src/builtins/pytype.rs
OddBloke: Hmm, I actually think my initial analysis was incorrect here (at least for the specific `binascii` example). In CPython, `Incomplete` is created by calling `PyErr_NewException("binascii.Incomplete", ...)` (https://github.com/python/cpython/blob/main/Modules/binascii.c#L1257). `PyErr_NewException` _splits_ the given string on that `.`: the first half is set as `__module__`, the second half is set as `__name__`.

So CPython's `binascii.Incomplete` object only has `Incomplete` as its name: this would be compatible with my implementation. So I think this is an example of a class of issue we'll have: RustPython exceptions will currently store `module.name` as `__name__`, which isn't how CPython Exceptions are constructed.

I'll dig into this some today.

OddBloke: Oh, https://github.com/python/cpython/blob/main/Python/errors.c#L1082-L1087 is the relevant `PyErr_NewException` code.

OddBloke: https://github.com/RustPython/RustPython/pull/3396 is what I've come up with to fix up the `binascii.Incomplete` case (and those like it). Please do let me know if you think that's sufficient to unblock landing this, or if there's more you'd like to see!

Lib/test/test_builtin.py
youknowone: We just left it or added a test in `extra_tests`, but this is an interesting approach.

vm/src/builtins/pytype.rs
moreal: I have researched after I saw the commit message of 07cfd112917d0438bda551875fdd5e2253fbfe65, which I missed in the before review.

First, [the fixup](#3396) looks good. 👍🏻 

But, in my current thought, this line should be reverted and we should add other lines to resolve the problem. 
CPython implementation seems to have two variables related to its name, `PyHeapTypeObject.ht_name` and `PyTypeObject.tp_name`. When we see the `type_name` function implementation:

```c
static PyObject *
type_name(PyTypeObject *type, void *context)
{
    puts("type_name"); // Added by @moreal for debugging
    if (type->tp_flags & Py_TPFLAGS_HEAPTYPE) {
        puts("is heaptype"); // Added by @moreal for debugging
        puts(type->tp_name); // Added by @moreal for debugging
        PyHeapTypeObject* et = (PyHeapTypeObject*)type;

        Py_INCREF(et->ht_name);
        return et->ht_name;
    }
    else {
        puts("is not heaptype"); // Added by @moreal for debugging
        return PyUnicode_FromString(_PyType_Name(type));
    }
}
```

Then, let's execute `import _ssl; _ssl.SSLError.__name__` and you will see the below outputs:

```
type_name
is heaptype
ssl.SSLError
'SSLError'
```

It branches into `type->tp_flags & Py_TPFLAGS_HEAPTYPE` and it returns `ht_name`.

Then, in case the commit message of 07cfd112917d0438bda551875fdd5e2253fbfe65:

```python
>>> class C: ...
... 
>>> C.__name__
type_name
is heaptype
C
'C'
>>> C.__name__ = "A.B"
>>> C.__name__
type_name
is heaptype
A.B
'A.B'
>>> 
```

The CPython implementation updates the `ht_name` field at [Objects/typeobject.c#L541](https://github.com/python/cpython/blob/14a4fce457033412278ca9a056787db424310dc3/Objects/typeobject.c#L541), in `type_set_name` (i.e. `PyType::set_name`), and at [Objects/typeobject.c#L3426](https://github.com/python/cpython/blob/b494f5935c92951e75597bfe1c8b1f3112fec270/Objects/typeobject.c#L3426), in `PyType_FromModuleAndSpec` (like `ssl.SSLError`).

Maybe, we should add new property like `ht_name` and separate the role.

OddBloke: Thanks for this comment, it's really helpful! I should have some time over the weekend to get my head back into this.

youknowone: I found this is resolved by d531e90319

youknowone (approved):
@OddBloke thanks!

PR #3389, opened 2021-10-29
underscore issue on parse_str (fmt done)

eunsung-lim:
I sent a pull request #3194 without formatting. 
Now, I've done with `cargo fmt --all` command. 
I think I need to do some works on #3194 but cannot find the way how to send a pull request continuously with my previous PR. If you know about it, please let me know...


youknowone:
Could you please do rebase rather than merge? Also, when you select `Files Changed` tap, nothing more than your contribution showed up.

youknowone:
Another thing. You don't need to open a new PR for every revision. `git push --force` to the original remote branch will replace codes from the original PR.

youknowone:
looking like outdated.
thank you for the effort!

PR #3392, opened 2021-10-30
Implement __str__ for OSError

deantvv:
In CPython, `__str__` and `__repr__` for `OSError` show different results.
This PR try to match this behavior but without the part after the
message (in following example is "'123' -> '456'").

```
In : exc.__repr__()
Out: "FileNotFoundError(2, 'No such file or directory')"

In : exc.__str__()
Out: "[Errno 2] No such file or directory: '123' -> '456'"

In : exc.args
Out: (2, 'No such file or directory')
```

### DIFF
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -745,6 +745,7 @@ impl ExceptionZoo {
             "filename" => ctx.none(),
             // second exception filename
             "filename2" => ctx.none(),
+            "__str__" => ctx.new_method("__str__", excs.os_error.clone(), os_error_str),
         });
         // TODO: this isn't really accurate
         #[cfg(windows)]
@@ -856,6 +857,21 @@ fn key_error_str(exc: PyBaseExceptionRef, vm: &VirtualMachine) -> PyStrRef {
             .unwrap()
     } else {
         exc.str(vm)
+    }
+}
+
+fn os_error_str(exc: PyBaseExceptionRef, vm: &VirtualMachine) -> PyResult<PyStrRef> {
+    let args = exc.args();
+    if args.as_slice().len() == 2 {
+        // SAFETY: len() == 2 is checked so get_arg 1 or 2 won't panic
+        let s = format!(
+            "[Errno {}] {}",
+            exc.get_arg(0).unwrap().str(vm)?,
+            exc.get_arg(1).unwrap().str(vm)?
+        );
+        Ok(vm.ctx.new_str(s))
+    } else {
+        Ok(exc.str(vm))
     }
 }
 


vm/src/exceptions.rs
```diff
@@ -859,6 +861,24 @@ fn key_error_str(exc: PyBaseExceptionRef, vm: &VirtualMachine) -> PyStrRef {
...
+    if args.as_slice().len() == 2 {
+        // SAFETY: len() == 2 is checked so get_arg 1 or 2 won't panic
+        let s = match (
+            exc.get_arg(0).unwrap().downcast_exact::<PyInt>(vm),
+            exc.get_arg(1).unwrap().downcast_exact::<PyStr>(vm),
+        ) {
+            (Ok(errno), Ok(msg)) => format!("[Errno {}] {}", errno, msg),
+            (Ok(errno), Err(_)) => format!("[Errno {}]", errno),
+            (Err(_), _) => "Error".to_owned(),
+        };
```
DimitrisJim, line 875: Seems like CPython just calls `__str__` on whatever is in these args and formats them:

```python
>>> class Foo:
...     def __str__(self):
...             return "__str__"
...     def __repr__(self):
...             return "__repr__"
... 
>>> OSError(Foo(), Foo()).__str__()
'[Errno __str__] __str__'
```

so this seems like it could be simplified to directly calling `str` whatever the args are as long as two of them exist.

vm/src/exceptions.rs
```diff
@@ -859,6 +860,24 @@ fn key_error_str(exc: PyBaseExceptionRef, vm: &VirtualMachine) -> PyStrRef {
...
+    if args.as_slice().len() == 2 {
+        // SAFETY: len() == 2 is checked so get_arg 1 or 2 won't panic
+        let s = match (
+            exc.get_arg(0).unwrap().str(vm),
+            exc.get_arg(1).unwrap().str(vm),
+        ) {
+            (Ok(errno), Ok(msg)) => format!("[Errno {}] {}", errno, msg),
+            (Ok(errno), Err(_)) => format!("[Errno {}]", errno),
+            (Err(_), _) => "Error".to_owned(),
+        };
```
DimitrisJim, line 874: Ah, this now handles the error if `__str__` fails, which I don't believe should happen, i.e:

```python
class Foo:
    def __str__(self):
        raise TypeError("for whatever reason")

OSError(Foo(), Foo()).__str__()  # doesn't raise
```

The match isn't needed, just use `?` on `str` and use the result directly, something like:

```rust
let s = format!("[Error {}] {}", exc.get_arg(0).unwrap().str(vm)?, exc.get_arg(1).unwrap().str(vm)?;
```

DimitrisJim (changes requested):

vm/src/exceptions.rs
deantvv: Interesting. Maybe `str` can return `PyStrRef` instead of `PyResult<PyStrRef>`.
I updated the code as suggested.

DimitrisJim (approved):

PR #3394, opened 2021-10-31
Fix `GenericAlias` parameter chaining

Snowapril:
This revision fix some misworking functions
* Fix original `make_parameters` & `subs_tvars`
* `test_parameters` and `test_parameter_chaining` tests now going success

I removed `extra_tests/snippets/test_typing.py` which I added in #3374, that I had been thought there are no tests for
using `TypeVar` with `GenericAlias`. It is actually covered by `test_parameter_chaining` in `test_genericalias.py`

@moreal Could you review this [commit](https://github.com/RustPython/RustPython/commit/59a2ade1cb3f3677e6c4c50b287f70548f76b333)? As you originally added this commit, I'd be a pleasure if you review this one 😊
In cpython, only they push parameters if they are not already exist. I fixed that mismatch between rustpython.

### DIFF
--- a/Lib/test/test_genericalias.py
+++ b/Lib/test/test_genericalias.py
@@ -173,8 +173,6 @@ class BaseTest(unittest.TestCase):
         self.assertEqual(a.__args__, (int,))
         self.assertEqual(a.__parameters__, ())
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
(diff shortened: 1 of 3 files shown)


youknowone (approved):

moreal (approved):
👍🏻 🔨

PR #3395, opened 2021-10-31
Time mod additions for Window

### LINKED ISSUE #3382: Time module functions for Windows
See #3381 for the PR for Unices.

The `time` module for Windows is currently missing the following (among other things):

 - [ ] `get_clock_info`
 - [ ] `monotonic`/`monotonic_ns`
 - [ ] `perf_counter`/`perf_counter_ns`

(note that the last two set of functions are there but they aren't implemented correctly)

Not sure how easy/hard these might be, but, I'd be willing to help as best as I can through gitter (though I don't have a Windows machine).

Snowapril:
Closes #3382 

This revision implements some function of time module for windows
* `get_clock_info`
* `monotonic` / `monotonic_ns`
* `perf_counter` / `perf_counter_ns`

### DIFF
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -90,28 +90,28 @@ mod time {
         Ok(Date::now() / 1000.0)
     }
 
-    #[cfg(any(windows, target_os = "wasi"))]
+    #[cfg(target_os = "wasi")]
     #[pyfunction]
     fn monotonic(vm: &VirtualMachine) -> PyResult<f64> {
         // TODO: implement proper monotonic time for other platforms.
         Ok(duration_since_system_now(vm)?.as_secs_f64())
     }
 
-    #[cfg(any(windows, target_os = "wasi"))]
+    #[cfg(target_os = "wasi")]
     #[pyfunction]
     fn monotonic_ns(vm: &VirtualMachine) -> PyResult<u128> {
         // TODO: implement proper monotonic time for other platforms.
         Ok(duration_since_system_now(vm)?.as_nanos())
     }
 
-    #[cfg(any(windows, target_os = "wasi"))]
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/time.rs
```diff
@@ -704,6 +710,136 @@ mod windows {
         }
     }
 
+    fn win_perf_counter_frequency(vm: &VirtualMachine) -> PyResult<i64> {
+        let freq = unsafe {
+            let mut frequency = std::mem::MaybeUninit::uninit();
+            if QueryPerformanceFrequency(frequency.as_mut_ptr()) == 0 {
+                // TODO: modify error message more properly
+                return Err(vm.new_os_error("Failed to get performance frequency".to_owned()));
```
DimitrisJim, line 718: Should also probably use `os::errno_err` which is the closest we currently have to `PyErr_SetFromWindowsErr` I believe.

vm/src/stdlib/time.rs
```diff
@@ -704,6 +710,136 @@ mod windows {
...
+
+        let frequency = unsafe { freq.QuadPart() };
+        if *frequency < 1 {
+            return Err(vm.new_runtime_error("invalid QueryPerformanceFrequency".to_owned()));
+        }
+
+        // TODO: is this overflow check essential?
+        if *frequency > i64::MAX / SEC_TO_NS {
+            return Err(vm.new_overflow_error("QueryPerformanceFrequency is too large".to_owned()));
+        }
```
DimitrisJim, line 731: I don't think so (the comment in the equivalent check in CPython notes that this shouldn't happen with 64bit values) but I guess its best if we still keep it.

DimitrisJim (approved):
Thanks! Can't test this locally but since CI isn't complaining, looks fine.

PR #3396, opened 2021-10-31
correctly set __module__ and __name__ for builtin exceptions

OddBloke:
In CPython's PyErr_NewException[0], the dotted "module.name" passed in is
split on that dot: the content before the dot is set as __module__ in
the created exception class, and the content after it is set as
__name__.

This commit mirrors that behaviour, by introducing a `module` argument
to `PyContext.new_class`: if Some("module_name") is passed, it will be
set as `__module__`.  All exception-creating uses of `new_class` are
updated to pass in their module and class names separately.

[0] https://github.com/python/cpython/blob/main/Python/errors.c#L1082-L1087

### DIFF
--- a/extra_tests/snippets/stdlib_binascii.py
+++ b/extra_tests/snippets/stdlib_binascii.py
@@ -50,3 +50,9 @@ assert_equal(
     "☢🐣  ᖇ𝓤𝕊тⓟ𝕐𝕥卄σ𝔫  ♬👣".encode(),
 )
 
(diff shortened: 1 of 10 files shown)


youknowone (approved):
looks good. is there any test related here or could you add a test under `extra_tests`?

youknowone:
cc @moreal

moreal (approved):
It looks good to me also! 👍🏻

PR #3398, opened 2021-11-01
os: implement os-error-str according to CPython

deantvv:
As was: "FileNotFoundError: [Errno 2] No such file or directory (os error 2)"
To be: "FileNotFoundError: [Errno 2] No such file or directory (os error 2): 'filename' -> 'filename2'"

Reference CPython implementation
```
if (self->filename) {
    if (self->filename2) {
        return PyUnicode_FromFormat("[Errno %S] %S: %R -> %R",
                                    OR_NONE(self->myerrno),
                                    OR_NONE(self->strerror),
                                    self->filename,
                                    self->filename2);
    } else {
        return PyUnicode_FromFormat("[Errno %S] %S: %R",
                                    OR_NONE(self->myerrno),
                                    OR_NONE(self->strerror),
                                    self->filename);
    }
}
if (self->myerrno && self->strerror)
    return PyUnicode_FromFormat("[Errno %S] %S",
                                self->myerrno, self->strerror);
```

### DIFF
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -862,13 +862,28 @@ fn key_error_str(exc: PyBaseExceptionRef, vm: &VirtualMachine) -> PyStrRef {
 
 fn os_error_str(exc: PyBaseExceptionRef, vm: &VirtualMachine) -> PyResult<PyStrRef> {
     let args = exc.args();
+    let obj = exc.as_object().to_owned();
+
     if args.as_slice().len() == 2 {
         // SAFETY: len() == 2 is checked so get_arg 1 or 2 won't panic
-        let s = format!(
-            "[Errno {}] {}",
-            exc.get_arg(0).unwrap().str(vm)?,
-            exc.get_arg(1).unwrap().str(vm)?
-        );
+        let errno = exc.get_arg(0).unwrap().str(vm)?;
+        let msg = exc.get_arg(1).unwrap().str(vm)?;
+
+        let s = match obj.clone().get_attr("filename", vm) {
+            Ok(filename) => match obj.get_attr("filename2", vm) {
+                Ok(filename2) => format!(
+                    "[Errno {}] {}: '{}' -> '{}'",
+                    errno,
+                    msg,
+                    filename.str(vm)?,
+                    filename2.str(vm)?
+                ),
+                Err(_) => format!("[Errno {}] {}: '{}'", errno, msg, filename.str(vm)?),
+            },
+            Err(_) => {
+                format!("[Errno {}] {}", errno, msg)
+            }
+        };
         Ok(vm.ctx.new_str(s))
     } else {
         Ok(exc.str(vm))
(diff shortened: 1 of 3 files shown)


vm/src/exceptions.rs
```diff
@@ -862,13 +862,30 @@ fn key_error_str(exc: PyBaseExceptionRef, vm: &VirtualMachine) -> PyStrRef {
 
 fn os_error_str(exc: PyBaseExceptionRef, vm: &VirtualMachine) -> PyResult<PyStrRef> {
     let args = exc.args();
+    let obj = exc.as_object().to_owned();
```
youknowone, line 865:
```suggestion
    let obj = exc.into();
```
otherwise if obj type is not inferrable,

```suggestion
    let obj: PyObjectRef = exc.into();
```

youknowone (approved):

vm/src/exceptions.rs
deantvv: This doesn't work since it will take the ownership of `exc`.

```
error[E0382]: borrow of moved value: `exc`
    |
869 |         let errno = exc.get_arg(0).unwrap().str(vm)?;
    |                     ^^^^^^^^^^^^^^ value borrowed here after move
```

vm/src/exceptions.rs
```diff
@@ -862,13 +862,30 @@ fn key_error_str(exc: PyBaseExceptionRef, vm: &VirtualMachine) -> PyStrRef {
...
+                    "[Errno {}] {}: '{}' -> '{}'",
+                    errno,
+                    msg,
+                    filename.str(vm)?,
+                    filename2.str(vm)?
+                )
+            }
+            (Ok(filename), Err(_)) => format!("[Errno {}] {}: '{}'", errno, msg, filename.str(vm)?),
+            (Err(_), _) => format!("[Errno {}] {}", errno, msg),
+        };
```
DimitrisJim, line 887: Pretty sure we can avoid one of the `get_attr` lookups here, i.e:

```suggestion
        let s = match obj.clone().get_attr("filename", vm) {
            Ok(filename) => match obj.get_attr("filename2", vm) {
                Ok(filename2) => format!(
                    "[Errno {}] {}: '{}' -> '{}'",
                    errno,
                    msg,
                    filename.str(vm)?,
                    filename2.str(vm)?
                ),
                _ => format!("[Errno {}] {}: '{}'", errno, msg, filename.str(vm)?),
            },
            _ => format!("[Errno {}] {}", errno, msg),
        };
```

DimitrisJim: Its unfortunate we need to `getattr` here, though. CPython stores these in the struct so its super cheap to look-up. Either way, thats an issue for another PR.

deantvv: > 
> 
> Its unfortunate we need to `getattr` here, though. CPython stores these in the struct so its super cheap to look-up. Either way, thats an issue for another PR.

Totally agree with that. In the first place, I thought I can avoid the `get_attr` but later found out it is impossible.

PR #3399, opened 2021-11-02
Allow sending custom signals to a vm

### LINKED ISSUE #3379: Stop the interpreter
Hi, I would like to know if we can stop the vm on runtime and if so how to stop it ? Thanks

coolreader18:
Resolves #3379

### DIFF
--- a/vm/src/vm.rs
+++ b/vm/src/vm.rs
@@ -24,8 +24,7 @@ use crate::{
     import,
     protocol::{PyIterIter, PyIterReturn, PyMapping},
     scope::Scope,
-    signal::NSIG,
-    stdlib,
+    signal, stdlib,
     types::PyComparisonOp,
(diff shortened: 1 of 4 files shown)


vm/src/vm.rs
```diff
@@ -407,6 +412,11 @@ impl VirtualMachine {
         state.frozen.extend(frozen);
     }
 
+    /// Set the custom signal channel for the interpreter
+    pub fn set_user_signal_channel(&mut self, signal_rx: signal::UserSignalReceiver) {
```
youknowone, line 416: is this never set to None once it is set?

youknowone (approved):

BannAgainVA (approved):

vm/src/vm.rs
coolreader18: Nope, or at least I don't see a point in doing so? It's just a configuration thing, so if someone doesn't want it they just don't call it in the first place.

PR #3397, opened 2021-11-01
Correct typo of types manually created

moreal:
This pull request fixes typo of types manually created. You can check these changes are valid fixups through the below lines:

```python
# test.py
import _csv, zlib, _ssl
print(_csv.Error, zlib.error, _ssl.SSLError, _ssl.SSLCertVerificationError, _ssl.SSLZeroReturnError, _ssl.SSLWantReadError, _ssl.SSLWantWriteError, _ssl.SSLSyscallError, _ssl.SSLEOFError)
```

## CPython (3.9.7)

`python3 test.py`

```
<class '_csv.Error'> <class 'zlib.error'> <class 'ssl.SSLError'> <class 'ssl.SSLCertVerificationError'> <class 'ssl.SSLZeroReturnError'> <class 'ssl.SSLWantReadError'> <class 'ssl.SSLWantWriteError'> <class 'ssl.SSLSyscallError'> <class 'ssl.SSLEOFError'>
```

## RustPython ([current main](https://github.com/RustPython/RustPython/tree/5f158f2c6f92d785865569659e90beb72008d306))

`cargo run --features=ssl -- test.py`

```
<class 'Error'> <class 'error'> <class 'SSLError'> <class 'SSLCertVerificationError'> <class 'SSLZeroReturnError'> <class 'SSLWantReadError'> <class 'SSLWantWriteError'> <class 'SSLSyscallError'> <class 'SSLEOFError'>
```

## RustPython (current pull request)

`cargo run --features=ssl -- test.py`

```
<class '_csv.Error'> <class 'zlib.error'> <class 'ssl.SSLError'> <class 'ssl.SSLCertVerificationError'> <class 'ssl.SSLZeroReturnError'> <class 'ssl.SSLWantReadError'> <class 'ssl.SSLWantWriteError'> <class 'ssl.SSLSyscallError'> <class 'ssl.SSLEOFError'>
```

PR #3402, opened 2021-11-02
Refactor duplicated codes in exception creations

Snowapril:
This revision implements below
*  with procedural-macro, auto-wrapping `#[pyattr(once)]` attached functions with static_cell! codes
*  refactor duplicated codes in several functions with `#[pyattr]` attached.
*  add `new_exception_type` method for creating module-level exceptions.

### DIFF
--- a/derive/src/pymodule.rs
+++ b/derive/src/pymodule.rs
@@ -1,6 +1,6 @@
 use crate::error::Diagnostic;
 use crate::util::{
-    iter_use_idents, pyclass_ident_and_attrs, text_signature, AttributeExt, ClassItemMeta,
+    iter_use_idents, path_eq, pyclass_ident_and_attrs, text_signature, AttributeExt, ClassItemMeta,
     ContentItem, ContentItemInner, ErrorVec, ItemMeta, ItemNursery, SimpleItemMeta,
     ALL_ALLOWED_NAMES,
 };
@@ -37,6 +37,41 @@ pub fn impl_pymodule(attr: AttributeArgs, module_item: Item) -> Result<TokenStre
     // collect to context
     for item in items.iter_mut() {
         let r = item.try_split_attr_mut(|attrs, item| {
+            // If attribute is #[pyattr] and item is ItemFn, then
+            // wrapping it with static_cell for preventing it from using it as function
+            if attrs.iter().any(|attr| {
+                path_eq(&attr.path, "pyattr")
+                    && if let Ok(syn::Meta::List(l)) = attr.parse_meta() {
+                        l.nested
+                            .into_iter()
+                            .any(|n| n.get_ident().map_or(false, |p| p == "once"))
+                    } else {
+                        false
+                    }
+            }) {
+                if let Item::Fn(syn::ItemFn { sig, block, .. }) = item {
+                    let stmts = &block.stmts;
+                    let return_type = match &sig.output {
+                        syn::ReturnType::Default => {
+                            unreachable!("#[pyattr] attached function must have return type.")
+                        }
+                        syn::ReturnType::Type(_, ty) => ty,
+                    };
+                    let stmt: syn::Stmt = parse_quote! {
+                        {
+                            rustpython_common::static_cell! {
+                                static ERROR: #return_type;
+                            }
+                            ERROR
+                                .get_or_init(|| {
+                                    #(#stmts)*
+                                })
+                                .clone()
+                        }
+                    };
+                    block.stmts = vec![stmt];
+                }
+            }
             let (pyitems, cfgs) = attrs_to_module_items(attrs, new_module_item)?;
             for pyitem in pyitems.iter().rev() {
                 let r = pyitem.gen_module_item(ModuleItemArgs {
--- a/derive/src/pymodule.rs
+++ b/derive/src/pymodule.rs
@@ -1,7 +1,7 @@
 use crate::error::Diagnostic;
 use crate::util::{
-    iter_use_idents, path_eq, pyclass_ident_and_attrs, text_signature, AttributeExt, ClassItemMeta,
-    ContentItem, ContentItemInner, ErrorVec, ItemMeta, ItemNursery, SimpleItemMeta,
+    iter_use_idents, pyclass_ident_and_attrs, text_signature, AttrItemMeta, AttributeExt,
+    ClassItemMeta, ContentItem, ContentItemInner, ErrorVec, ItemMeta, ItemNursery, SimpleItemMeta,
     ALL_ALLOWED_NAMES,
 };
 use proc_macro2::TokenStream;
@@ -37,41 +37,6 @@ pub fn impl_pymodule(attr: AttributeArgs, module_item: Item) -> Result<TokenStre
     // collect to context
     for item in items.iter_mut() {
         let r = item.try_split_attr_mut(|attrs, item| {
-            // If attribute is #[pyattr] and item is ItemFn, then
-            // wrapping it with static_cell for preventing it from using it as function
-            if attrs.iter().any(|attr| {
-                path_eq(&attr.path, "pyattr")
-                    && if let Ok(syn::Meta::List(l)) = attr.parse_meta() {
-                        l.nested
-                            .into_iter()
-                            .any(|n| n.get_ident().map_or(false, |p| p == "once"))
-                    } else {
-                        false
-                    }
-            }) {
-                if let Item::Fn(syn::ItemFn { sig, block, .. }) = item {
-                    let stmts = &block.stmts;
-                    let return_type = match &sig.output {
-                        syn::ReturnType::Default => {
-                            unreachable!("#[pyattr] attached function must have return type.")
-                        }
-                        syn::ReturnType::Type(_, ty) => ty,
-                    };
-                    let stmt: syn::Stmt = parse_quote! {
-                        {
-                            rustpython_common::static_cell! {
-                                static ERROR: #return_type;
-                            }
-                            ERROR
-                                .get_or_init(|| {
-                                    #(#stmts)*
-                                })
-                                .clone()
-                        }
-                    };
-                    block.stmts = vec![stmt];
-                }
-            }
             let (pyitems, cfgs) = attrs_to_module_items(attrs, new_module_item)?;
             for pyitem in pyitems.iter().rev() {
                 let r = pyitem.gen_module_item(ModuleItemArgs {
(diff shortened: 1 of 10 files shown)


stdlib/src/ssl.rs
```diff
@@ -174,90 +175,75 @@ mod _ssl {
...
+            rustpython_common::static_cell! {
+                static ERROR: PyTypeRef;
+            }
+            ERROR
+                .get_or_init(|| {
+                    PyType::new_ref(
+                        concat!("ssl.", $exc_name),
+                        vec![$($base), +],
+                        Default::default(),
+                        PyBaseException::make_slots(),
```
Snowapril, line 189: Although original `new_ref` for `SSLError` use `PyBaseException::make_slots()` and `new_simple_ref` for the other errors use `Default::default()`, I replaced them as just single `PyBaseException::make_slots()` as I think there are no difference between them in cpython.

DimitrisJim: I'm thinking this should apply to more errors, not just `ssl` (i.e the places where `PyErr_NewException` is used). Using the same macro for errors in the other modules #3397 didn't work?

Snowapril: Hmm.. I got it though, it seems a lot complex than I think. 
It seems `zlib.error` can be replaced by current macro, though `_csv` use `PyType_FromModuleAndSpec` with `PyExc_Exception` as bases arguments. I dont know what is the difference between them yet..

Snowapril: And current module-level exceptions have wrong `__module__` as its attribute
#### rustpython
```python
>>>>> csv.Error.__module__
>>>>> SSLCertVerificationError.__module__
'builtins'
>>>>> zlib.error.__module__
'builtins'
```

#### cpython
```python
>>> csv.Error.__module__
'_csv'
>>> _ssl.SSLCertVerificationError.__module__
'ssl'
>>> zlib.error.__module__
'zlib'
```

Snowapril: I think there need to `PyErr_NewException` which create exception with given bases and also parse module name from given exception separated by dot('.').

Snowapril: What is current `vm.new_exception` exactly for? I wonder if it is somewhat originally intended to use in there cases,.

DimitrisJim: I think #3396 might be what we need here, when that lands the macro could just invoke `vm.new_class` it seems.

Snowapril: Oh. It seems so 😊. After #3396 merged, I'll modify the macro to use it and apply it to other errors.

DimitrisJim: Merged!

DimitrisJim: Ok, so, I don't think this needs a macro tbh. It would probably be better to just create a new function (in ctx), name it `new_exception_type` and duplicate what `new_class` is doing. 

Looking (quite back!) into why `new_class` was added, it appears it was meant as a way to create exceptions ([ref](https://github.com/RustPython/RustPython/commit/f76472c00b7fc32ec2bf821539e3b321c3bfd3aa)). I think it is misnamed and currently confusing.

`vm.new_exception` -> Create a new exception instance.
`vm.ctx.new_exception_type` -> Create a new exception (specifically for module exceptions, mirroring what `PyErr_NewException` does).

It should definitely take a name, a module, and an optional docstring (`_ssl` errors seem to have docs attached to them). It should then forward everything to `PyType::new_ref`). I *think* this way it would make it usable for all module exceptions we currently need to create.

Snowapril: Sorry for being late! Thanks for the detailed advice!  I was meant to modify macro itself to something other (sorry for bad English 😂). but your idea seems much better than what I originally planned to do. 

I would ping you after implementation (I have no time to work on it today, sorry .😢)

Snowapril: @DimitrisJim Long time no see! I add some procedural macro code for wrapping each function which have `#[pyattr]` in [b761e69](https://github.com/RustPython/RustPython/pull/3402/commits/b761e6991d3ac6f12005aa4cf15912deae5d8f5f) and [ae9f85c](https://github.com/RustPython/RustPython/pull/3402/commits/ae9f85c2b9bcc3dd2627d9a32c3bb3bcc3050f2c).

We planned to pass docstring to `new_exception_type` like cpython, though I think automatic document parsing with comment prefix with `///` would be better. 

But it is currently not working as we expected. This maybe handled in separate PR.

DimitrisJim: > I add some procedural macro code for wrapping each function which have #[pyattr]

that's an interesting approach! It should also work fine for non-error `#[pyattr] fn`'s I'm supposing.

> We planned to pass docstring to new_exception_type like cpython, though I think automatic document parsing with comment prefix with /// in rustpython would be better.

Right, I don't remember if we actually try and grab docstrings for `pyattr`s, but that is definitely a separate PR.

Snowapril:
Uhm... I have no idea why rusttests have failed 😂

youknowone:
@Snowapril  Look upside to find frozen lib errors

```
 RuntimeError: Could not import encodings. Is your RUSTPYTHONPATH set? If you don't have access to a consistent external environment (e.g. if you're embedding rustpython in another application), try enabling the freeze-stdlib feature
test dictdatatype::tests::test_abc ... FAILED
test dictdatatype::tests::test_insert ... FAILED
RuntimeError: lost sys.stderr
  File "_frozen_importlib_external", line 1439, in find_spec
AttributeError: 'FileFinder' object has no attribute 'path'

During handling of the above exception, another exception occurred:

  File "_frozen_importlib", line 1094, in __import__
  File "_frozen_importlib", line 1015, in _gcd_import
  File "_frozen_importlib", line 992, in _find_and_load
  File "_frozen_importlib", line 992, in _find_and_load
  File "_frozen_importlib", line 972, in _find_and_load_unlocked
  File "_frozen_importlib", line 915, in _find_spec
  File "_frozen_importlib", line 915, in _find_spec
  File "_frozen_importlib_external", line 1342, in find_spec
  File "_frozen_importlib_external", line 1314, in _get_spec
  File "_frozen_importlib_external", line 1440, in find_spec
NameError: name 'OSError' is not defined. Did you mean: 'EOFError'?

RuntimeError: Could not import encodings. Is your RUSTPYTHONPATH set? If you don't have access to a consistent external environment (e.g. if you're embedding rustpython in another application), try enabling the freeze-stdlib feature
test dictdatatype::tests::test_x ... FAILED
test format::tests::test_all ... ok
test format::tests::test_fill_and_align ... ok
test format::tests::test_fill_and_width ... ok
test format::tests::test_format_int ... ok
test format::tests::test_format_invalid_specification ... ok
test format::tests::test_format_parse ... ok
test format::tests::test_format_parse_escape ... ok
test format::tests::test_format_parse_fail ... ok
test format::tests::test_parse_field_name ... ok
test format::tests::test_width_only ... ok
test function::tests::test_intonativefunc_noalloc ... ok
test pyobjectrc::tests::miri_test_type_initialization ... ok
test stdlib::io::_io::tests::test_buffered_read ... ok
test stdlib::io::_io::tests::test_buffered_seek ... ok
test stdlib::io::_io::tests::test_buffered_value ... ok
RuntimeError: lost sys.stderr
  File "_frozen_importlib_external", line 1439, in find_spec
AttributeError: 'FileFinder' object has no attribute 'path'

During handling of the above exception, another exception occurred:

  File "_frozen_importlib", line 1094, in __import__
  File "_frozen_importlib", line 1015, in _gcd_import
  File "_frozen_importlib", line 992, in _find_and_load
  File "_frozen_importlib", line 992, in _find_and_load
  File "_frozen_importlib", line 972, in _find_and_load_unlocked
  File "_frozen_importlib", line 915, in _find_spec
  File "_frozen_importlib", line 915, in _find_spec
  File "_frozen_importlib_external", line 1342, in find_spec
  File "_frozen_importlib_external", line 1314, in _get_spec
  File "_frozen_importlib_external", line 1440, in find_spec
NameError: name 'OSError' is not defined. Did you mean: 'EOFError'?

```

derive/src/pymodule.rs
```diff
@@ -37,6 +37,36 @@ pub fn impl_pymodule(attr: AttributeArgs, module_item: Item) -> Result<TokenStre
     // collect to context
     for item in items.iter_mut() {
         let r = item.try_split_attr_mut(|attrs, item| {
+            // If attribute is #[pyattr] and item is ItemFn, then
```
DimitrisJim, line 40: this is causing issues someplace but I can't track it down. Removing it from here and moving the static_cell to `new_exception_type` should be a viable alternative though.

Snowapril: Moving `static_cell` to `new_exception_type` may solve that test failures. But we need to declare it as macro as each functions need its own `static_cell`. Approach with procedural-macro is best I think as it also solve other stuffs (#[pyattr] attached function in module which is not exception-type)

After more trying, I would mention you. sorry 😢

DimitrisJim: Ah, right. Adding an `error` argument to `pyattr` then? Have it only applied on attributes that are meant to denote errors.

Snowapril: Oh . I missed your comment. I add "once" keyword to #[pyattr]  like #[pyattr(name = "blahblah", once)] for that exceptions.

youknowone: I am not sure this is fit for here, but adding a field to `ALLOWED_NAMES` allows to read it from `ItemMeta`.
(with a new `AttrItemMeta`)
If you are interested in, check how `ClassItemMeta` handles `metaclass`. If this is not fit for here, still `AttrItemMeta` is preferred not to allow other simple items to have `once`.

Snowapril: Sorry for begin late 😂 I moved static_cell wrapping code to `get_module_item` of `AttributeItem`.
Use `AttrItemMeta::from_attr` for `ItemFn` and replace original once keyword checking code to `attr_meta.inner()._bool("once")?` . For the other cases, `Item::Const` and `Item::Use`, just use `SimpleItemMeta::from_attr` as they do not have `once` keyword ever. Please take a look this commit [7f2fd97](https://github.com/RustPython/RustPython/pull/3402/commits/7f2fd97278905bf9f3e7e8bd834b89bfb7fda1e8). Thanks 😊

youknowone (approved):

PR #3405, opened 2021-11-02
cformat.rs: fix remaining test_format.test_common_format failures

OddBloke:
With one exception, which I'll comment on inline. Otherwise, commits are atomic.

### DIFF
--- a/Lib/test/test_format.py
+++ b/Lib/test/test_format.py
@@ -94,8 +94,6 @@ def test_exc_common(formatstr, args, exception, excmsg):
 
 class FormatTest(unittest.TestCase):
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_common_format(self):
         # test the format identifiers that work the same across
         # str, bytes, and bytearrays (integer, float, oct, hex)
--- a/Lib/test/test_types.py
+++ b/Lib/test/test_types.py
@@ -86,8 +86,6 @@ class TypesTests(unittest.TestCase):
         if float(1) == 1.0 and float(-1) == -1.0 and float(0) == 0.0: pass
         else: self.fail('float() does not work properly')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_float_to_string(self):
         def test(f, result):
             self.assertEqual(f.__format__('e'), result)
--- a/common/src/float_ops.rs
+++ b/common/src/float_ops.rs
@@ -157,9 +157,9 @@ pub fn format_exponent(precision: usize, magnitude: f64, case: Case) -> String {
 /// If s represents a floating point value, trailing zeros and a possibly trailing
 /// decimal point will be removed.
 /// This function does NOT work with decimal commas.
-fn remove_trailing_redundant_chars(s: String) -> String {
-    if s.contains('.') {
-        // only truncate floating point values
+fn maybe_remove_trailing_redundant_chars(s: String, alternate_form: bool) -> String {
+    if !alternate_form && s.contains('.') {
+        // only truncate floating point values when not in alternate form
         let s = remove_trailing_zeros(s);
         remove_trailing_decimal_point(s)
     } else {
@@ -183,7 +183,12 @@ fn remove_trailing_decimal_point(s: String) -> String {
     s
 }
 
-pub fn format_general(precision: usize, magnitude: f64, case: Case) -> String {
+pub fn format_general(
+    precision: usize,
+    magnitude: f64,
+    case: Case,
+    alternate_form: bool,
+) -> String {
     match magnitude {
         magnitude if magnitude.is_finite() => {
             let r_exp = format!("{:.*e}", precision.saturating_sub(1), magnitude);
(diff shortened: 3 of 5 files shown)


Lib/test/test_format.py
```diff
@@ -94,14 +94,13 @@ def test_exc_common(formatstr, args, exception, excmsg):
...
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_common_format(self):
         # test the format identifiers that work the same across
         # str, bytes, and bytearrays (integer, float, oct, hex)
         testcommon("%%", (), "%")
         testcommon("%.1d", (1,), "1")
-        testcommon("%.*d", (sys.maxsize,1), overflowok=True)  # expect overflow
+        # TODO: RUSTPYTHON (overflows kill the process immediately in Rust)
+        #testcommon("%.*d", (sys.maxsize,1), overflowok=True)  # expect overflow
```
OddBloke, line 103: I know we don't like doing this, but AIUI running this line doesn't result in an exception at all, or even a panic, but kills the process immediately: I'm not sure what else to do, and this feels like a lesser evil than copying the rest of this test elsewhere (where it won't get updated for new CPython changes to it in future).

DimitrisJim: I think this might be a relatively easy error to get. Seems like CPython tries to transform this to an `int` at some point and this won't work for `sys.maxsize`. We probably store precision as `usize` or `isize`, while it should probably be an `i32`. Not sure where exactly this is done but it shouldn't be a hard fix (I believe)

OddBloke: OK, I'll dig into what's causing this. This line should return a string equivalent to `"0" * sys.maxsize + "1"`, so my assumption has been that it's the allocation of that string which is failing, but looking at this again I think that would be a `MemoryError`.

DimitrisJim: I doubt we even handle that yet but I'm not sure tbh. Not sure if RustPython is able to handles the max CPython currently does without crashing.

Looking into the source, `2**31 - 4`, assuming 32bit ints is the max `precision` could be (see `_PyUnicode_FormatLong` in `unicodeobject.c`). Not sure why yet, but it does result in slightly different reporting of the error:

```python
 >>> "%.*d" % (2**31, 1)
OverflowError: Python int too large to convert to C int
>>> "%.*d" % (2**31-1, 1)
OverflowError: precision too large
```

OddBloke: OK, so it looks like this is actually a boundary issue:

```py
>>>>> import sys
>>>>> "%.*d" % (sys.maxsize + 1, 2)
OverflowError: Python int too large to convert to Rust isize
>>>>> "%.*d" % (sys.maxsize, 2)
memory allocation of 9223372036854775806 bytes failed
[1]    3949127 abort (core dumped)  cargo run
```

CPython rejects exactly `sys.max_size` whereas RustPython accepts `sys.max_size` (but _does_ raise `OverflowError`s, starting at `sys.max_size + 1`).

I don't think the _exact_ behaviour matters too much here, `sys.max_size - 20000` does this:

```py
>>>>> "%.*d" % (sys.maxsize - 20000, 2)
memory allocation of 9223372036854755806 bytes failed
[1]    3949314 abort (core dumped)  cargo run
```

and raises MemoryError on CPython: so regardless of the boundary behaviour, it's going to fail.

I think I'll tweak this to test with `sys.max_size + 1`, so we do trigger the overflow. LMK if you'd prefer a different resolution!

DimitrisJim: Changing this to match the test case is relatively easy, in `try_update_quantity_from_tuple`, `try_to_primitive` should cast to an `i32` instead of an `isize`. To match the same behavior when the precision is supplied in the string, `parse_quantity` can use `as i32` instead of `isize` for `num`.

I think these cover both precision and width, right? Either way, I think those casts should be enough.

OddBloke: `sys.maxsize` is 64-bit on 64-bit platforms: on my laptop I can run:

```
$ python3 -c "print(len('%*.f' % (2**32 + 1, 12)))"
4294967297
```

and, indeed (after closing my browser for a minute :grin:):

```
$ cargo run --release -- -c "print(len('%*.f' % (2**32 + 1, 12)))"
    Finished release [optimized] target(s) in 1.26s
     Running `target/release/rustpython -c 'print(len('\''%*.f'\'' % (2**32 + 1, 12)))'`
4294967297
```

The RustPython version rejects this with `i32` (and even with `u32`). So I do think `isize` is right?

DimitrisJim: Ah, so the width can be a `ssize_t` while the precision can't. That seems like such a weird disparity. Either way, since we don't actually separate the conversion of width/precision, I'd say keeping it an `i32` is probably the best approach for now.

OddBloke: Aha, oops, I hadn't intended to switch from one to the other in my examples! I hadn't realised they were different either, just from the opposite angle: I can look into replicating that more closely in a follow-up.

(Did you mean "keeping it an `isize`"?)

DimitrisJim: Sorry, changing it into an `i32`! I think its a better way to approach this for now until (and if!) we handle width/precision separately.

OddBloke: Hmm, I'm not sure I agree.

With `isize`, we are more permissive than CPython: we will accept precisions larger than CPython does, but Python code using large widths will behave the same on both CPython and RustPython. With `i32`, there's Python code which CPython runs which we would reject.

My feeling is that accepting larger precisions than CPython isn't much of a problem: given that the canonical Python implementation rejects them, there's unlikely to be any such Python code out there (and some quick testing locally suggests that such values don't break RustPython).

Am I missing an issue that accepting larger precisions introduces?

DimitrisJim: We agree with the test. This might seem pedantic but when tests get updated (and most tests currently do need that!) it's far more likely the whole test will be skipped. The updating process is usually just: copy things over and re-mark as failing/skip the appropriate tests. A small change like this is very likely to be missed by both future contributor and future reviewer (Especially in bulk-updates).

OddBloke: :facepalm: I had once again thought that the test was for width, not precison. Fixing now!

OddBloke: Done!

DimitrisJim: Annoyingly enough, another test case for precision further down in `test_format` (which is currently skipped altogether) *actually uses* `sys.maxsize + 1` (and also doesn't lead to overflow, just a value error, another slight discrepancy, now between `format` and `%` formatting).

DimitrisJim (approved):

OddBloke:
Thanks for the patient review, @DimitrisJim!

PR #3408, opened 2021-11-03
Fix panic in debug mode

coolreader18:
The nix change is necessary in order to upgrade bitflags to 1.3 - there was a whole mess of dependency stuff where nix 0.22.0 requires just bitflags ^1.1, but 0.22.{1,2} requires bitflags <1.3 - rustyline only requires nix 0.22.0, so upgrading our nix requirement lets cargo figure out a solution.


coolreader18:
Oh, and the specific panic I was getting was an assertion failure on `is_created_with_flags()`, since I think it's possible for rustpython-derive to be compiled not in debug mode but the vm to be compiled in debug mode, so moving the `cfg()` inside `quote!{}` fixes that.

DimitrisJim (approved):

PR #3409, opened 2021-11-03
Add wasi check to ci.

DimitrisJim:


vm/src/exceptions.rs
```diff
@@ -998,6 +998,7 @@ pub(crate) fn raw_os_error_to_exc_type(errno: i32, vm: &VirtualMachine) -> Optio
         errors::EALREADY => Some(excs.blocking_io_error.clone()),
         errors::EINPROGRESS => Some(excs.blocking_io_error.clone()),
         errors::EPIPE => Some(excs.broken_pipe_error.clone()),
+        #[cfg(not(target_os = "wasi"))]
```
DimitrisJim, line 1001: this appears to be in `wasi-libc` but not in `libc`, I'll probably add it soon.

youknowone (approved):

PR #3410, opened 2021-11-04
fix failures in test_types' test_int__format__

OddBloke:


OddBloke:
Hmm, this isn't close to fixing all of that test yet, must have accidentally run without removing the expectedFailure decorator! Will resubmit once it's ready.

PR #3412, opened 2021-11-04
Allow any object as an argument for random.seed

### LINKED ISSUE #3411: random.seed does not accept floats
On CPython/PyPy, this works:

```python
>>>>> import random
>>>>> random.seed(0.5)
  File "/.../RustPython/vm/pylib-crate/Lib/random.py", line 133, in seed
    super().seed(a)
TypeError: 'float' object cannot be interpreted as an integer
```

DimitrisJim:
Closes #3411

### DIFF
--- a/Lib/test/test_random.py
+++ b/Lib/test/test_random.py
@@ -28,8 +28,6 @@ class TestBasicOps:
         state2 = self.gen.getstate()
         self.assertNotEqual(state1, state2)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_saverestore(self):
         N = 1000
         self.gen.seed()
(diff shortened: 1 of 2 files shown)


Lib/test/test_random.py
```diff
@@ -899,7 +883,7 @@ def test_gammavariate_alpha_greater_one(self, random_mock):
         returned_value = random.gammavariate(1.1, 2.3)
         self.assertAlmostEqual(returned_value, 2.53)
 
-    # TODO: RUSTPYTHON
+    # TODO: RUSTPYTHON
```
fanninpm, line 886: Was this just a failure?

fanninpm (commented):
Sometimes you added only spaces to the ends of comments.

Lib/test/test_random.py
DimitrisJim: I think both these errors had to do with precision issues (an assert failed) so I was slightly hesitant to add the output.

DimitrisJim: went ahead and just added that the assert fails in these cases.

youknowone (approved):

PR #3413, opened 2021-11-04
Remove slot_new for longrange_iterator.

DimitrisJim:
`Unconstructible` handles forbidding instantiation.


youknowone (approved):
I am surprised having both `Constructor` and `slot_new` didn't raise any error

PR #3416, opened 2021-11-06
os: fix posix_spawn exception

deantvv:


vm/src/stdlib/posix.rs
```diff
@@ -1278,7 +1278,7 @@ pub mod module {
         fn spawn(self, spawnp: bool, vm: &VirtualMachine) -> PyResult<libc::pid_t> {
             use crate::TryFromBorrowedObject;
 
-            let path = CString::new(self.path.into_bytes())
+            let path = CString::new(self.path.clone().into_bytes())
```
youknowone, line 1281:
```suggestion
            let path = CString::new(self.path.as_bytes())
```
is this working?

youknowone (approved):

vm/src/stdlib/posix.rs
deantvv: No, we don't have `as_bytes` for `PyPathLike` but IMO it makes sense to implement it.

youknowone: Probably not. `Path` doesn't look like to be borrowable as bytes.

youknowone: But I found another utility to do this job
```suggestion
            let path = self.path.into_cstring()?;
```

deantvv: We need to clone for this to work.

PR #3421, opened 2021-11-06
Implement `object.__sizeof__`

moreal:
This pull request may close https://github.com/RustPython/RustPython/issues/356.

This pull request implements `object.__sizeof__` fully, except for setting all types' `ob_size`. It introduces `PyVarObject.ob_size` field and `tp_itemsize` slot. Also it provides a way to setup `itemsize` slot with `pyclass` macro.

`ob_size` field is used to calculate variable object's size with `itemsize` (`ob_size * itemsize + basicsize`).

Because this pull request modified the core part like (`PyInner` and `PyObjectRef`) and I felt the next work can be more bigger, I cut this part and request a review for this pull request.

----

Related codes:

- CPython `object.__sizeof__` impl: https://github.com/python/cpython/blob/e5d8dbdd304935dbd0631ee9605efb501332f792/Objects/typeobject.c#L5520-L5533
- CPython `PyVarObject`: https://github.com/python/cpython/blob/fa2b8b75eb2b8a0193d587e02b488a73579118fc/Include/object.h#L111-L114

-----

The next work will be to call `set_ob_size`.

### DIFF
--- a/vm/src/builtins/object.rs
+++ b/vm/src/builtins/object.rs
@@ -4,7 +4,7 @@ use crate::{
     class::PyClassImpl,
     function::Either,
     function::{FuncArgs, PyArithmeticValue, PyComparisonValue},
-    types::PyComparisonOp,
+    types::{PyComparisonOp, SizeOf},
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyResult, VirtualMachine,
 };
 
@@ -25,7 +25,7 @@ impl PyPayload for PyBaseObject {
     }
 }
 
-#[pyimpl(flags(BASETYPE))]
+#[pyimpl(flags(BASETYPE), with(SizeOf))]
 impl PyBaseObject {
     /// Create and return a new object.  See help(type) for accurate signature.
     #[pyslot]
@@ -331,10 +331,11 @@ impl PyBaseObject {
     fn hash(zelf: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyHash> {
         Self::slot_hash(&zelf, vm)
     }
+}
 
-    #[pymethod(magic)]
-    fn sizeof(zelf: PyObjectRef) -> usize {
-        zelf.class().slots.basicsize
+impl SizeOf for PyBaseObject {
+    fn sizeof(zelf: &Py<Self>) -> PyResult<usize> {
+        Ok(zelf.class().slots.basicsize)
     }
 }
 
--- a/vm/src/types/slot.rs
+++ b/vm/src/types/slot.rs
@@ -52,6 +52,8 @@ pub struct PyTypeSlots {
     pub iter: AtomicCell<Option<IterFunc>>,
     pub iternext: AtomicCell<Option<IterNextFunc>>,
 
+    pub sizeof: AtomicCell<Option<SizeOfFunc>>,
+
     // Flags to define presence of optional/expanded features
     pub flags: PyTypeFlags,
 
@@ -162,6 +164,7 @@ pub(crate) type NewFunc = fn(PyTypeRef, FuncArgs, &VirtualMachine) -> PyResult;
 pub(crate) type InitFunc = fn(PyObjectRef, FuncArgs, &VirtualMachine) -> PyResult<()>;
 pub(crate) type DelFunc = fn(&PyObject, &VirtualMachine) -> PyResult<()>;
 pub(crate) type AsSequenceFunc = fn(&PyObject, &VirtualMachine) -> &'static PySequenceMethods;
+pub(crate) type SizeOfFunc = fn(&PyObject, &VirtualMachine) -> PyResult<usize>;
 
 // slot_sq_length
 pub(crate) fn slot_length(obj: &PyObject, vm: &VirtualMachine) -> PyResult<usize> {
@@ -943,5 +946,25 @@ where
     #[cold]
     fn iter(_zelf: PyRef<Self>, _vm: &VirtualMachine) -> PyResult {
         unreachable!("slot_iter is implemented");
+    }
+}
+
+#[pyimpl]
+pub trait SizeOf: PyPayload {
+    #[pyslot]
+    fn slot_sizeof(zelf: &PyObject, vm: &VirtualMachine) -> PyResult<usize> {
+        if let Some(zelf) = zelf.downcast_ref() {
+            Self::sizeof(zelf)
+        } else {
+            Err(vm.new_type_error("unexpected payload for __sizeof__".to_owned()))
+        }
+    }
+
+    fn sizeof(zelf: &Py<Self>) -> PyResult<usize>;
+
+    #[inline]
+    #[pymethod]
+    fn __sizeof__(zelf: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        Self::slot_sizeof(&zelf, vm).to_pyresult(vm)
     }
 }


vm/src/builtins/dict.rs
```diff
@@ -989,6 +989,11 @@ impl PyDictKeys {
     fn contains(zelf: PyRef<Self>, key: PyObjectRef, vm: &VirtualMachine) -> PyResult<bool> {
         zelf.dict().contains(key, vm)
     }
+
+    #[pymethod(magic)]
+    fn sizeof(zelf: PyRef<Self>) -> usize {
+        zelf.dict().sizeof()
```
youknowone, line 995: `dict` has data but `dict_keys` is just a view of dict. So it would be not the same

DimitrisJim: I think this could just be `size_of::<Self>` and could be implemented directly on `DictView`.

youknowone:
@moreal is there anything more work for this PR?

moreal:
When I was working on this pull request, I realized there is no `__sizeof__` method for `dict_keys` and `dict_values`, `dict_items` but it is from `object.__sizeof__`. So the correct solution is to implement `object.__sizeof__` with additional slots.

So I'll implement `object.__sizeof__` with additional slots.

P.S. The reason I didn't any action for this pull request, is just I forgot. Thanks!

vm/src/builtins/object.rs
```diff
@@ -305,6 +305,21 @@ impl PyBaseObject {
...
+
+    #[pymethod(magic)]
+    fn sizeof(zelf: PyObjectRef) -> usize {
+        let mut size: usize = 0;
+        let isize = zelf.class().slots.itemsize;
+        if isize > 0 {
+            // TODO: CPython implementation has `PyVarObject.ob_size` field for each objects.
+            // Its default value seems 0 but it is changed by dynamically.
+            let object_size = 0;
+            size = object_size * isize;
```
moreal, line 317: About these comments, I couldn't find an API to get `PyObject`'s size yet. 😞 

You can see the original implementation at https://github.com/python/cpython/blob/6d430ef5ab62158a200b94dff31b89524a9576bb/Objects/typeobject.c#L5481-L5494.

Snowapril:
```c
PyTypeObject PyBaseObject_Type = {
    PyVarObject_HEAD_INIT(&PyType_Type, 0)
    "object",        
...
```

I think that second argument in `PyVarObject_HEAD_INIT` is `ob_size` and that `PyVarObject_HEAD_INIT` assigned to `ob_base` in each types

rng-dynamics:
Regarding `tp_basicsize`, `tp_itemsize`, and `ob_size` in CPython: `tp_itemsize` and `ob_size` are for variable-size objects only. At https://docs.python.org/3/extending/newtypes_tutorial.html#defining-new-types we can see the intended use of these fields.

AFAIU RustPython does not try to replicate this interface exactly. For example, in RustPython `__sizeof__` is already implemented for list, bytearray, ... without `tp_itemsize` or `ob_size`. So, I am wondering: can't we just return `std::mem::size_of::<PyBaseObject>() + std::mem::size_of::<PyObject>()` in `object.__sizeof__`?

If you want I can make a separate pull request, since this issue is already inactive since more than four months.

moreal:
I stopped to work on this pull request since I met I couldn't find the way to get RustPython object's memory size.
As I understood, if it doesn't need to be same between RustPython's `__sizeof__` and CPython's `__sizeof__`, https://github.com/RustPython/RustPython/pull/3728 will be merged and this pull request should be closed.

youknowone:
I merged #3728. If this PR has more enhancement, please keep go ahead.

moreal:
This comment is a note to record the progress to solve this issue.

---------

At [current commit](https://github.com/RustPython/RustPython/commit/4a0be5a6d0e2325f88a03abe3714395912734bbb), when we run the below code, it seems like unnatural.

```python
# RustPython
>>>>> object.__sizeof__(1223456789812391231291231231231212312312312312312312321321321321312312321123123123199129391239219394923912949213021949302194942130123949203912430392402139210492139123012940219394923942395943856228368385)
32

# CPython
>>> object.__sizeof__(1223456789812391231291231231231212312312312312312312321321321321312312321123123123199129391239219394923912949213021949302194942130123949203912430392402139210492139123012940219394923942395943856228368385)
116
```

When I build the CPython with debugging log in local environment and I ran it, the `object.__sizeof__` doesn't call `int.__sizeof__`. So I decided to focus on the `int` type. 

The `object.__sizeof__` requires:

- `tp_basicsize`: `int` uses `offsetof(PyLongObject, ob_digit)`
- `tp_itemsize`: `int` uses `sizeof(digit)`
- `ob_size`: `int` initializes the size at `_PyLong_New` -> `_PyObject_InitVar` -> `Py_SET_SIZE` -> `_Py_SET_SIZE` -> `ob->ob_size = size;` after `PyObject_Malloc` manually.

`bytes` also does it.

Then It will be the solution to introduce `ob_size` field and to provide method to initialize it. Where should the field be added? Now, I guess it'll be `PyInner` 🤔 

This comment will be continued...

vm/src/builtins/object.rs
moreal: Thank you! It was helpful to analyze this issue again. 🙏🏻

youknowone:
Does `int.__sizeof__` correctly work? If that correct, How about adding `slot_sizeof` and `SizeOf` trait? 
Then `object.__sizeof__` will call correct variants of each types

vm/src/builtins/int.rs
```diff
@@ -33,7 +33,7 @@ use std::fmt;
 /// Base 0 means to interpret the base from the string as an integer literal.
 /// >>> int('0b100', base=0)
 /// 4
-#[pyclass(module = false, name = "int")]
+#[pyclass(module = false, name = "int", itemsize = 4)]
```
youknowone, line 36: I wish we have a better way than adding manual `itemsize` to type definition.

vm/src/builtins/object.rs
```diff
@@ -329,7 +329,14 @@ impl PyBaseObject {
 
     #[pymethod(magic)]
     fn sizeof(zelf: PyObjectRef) -> usize {
-        zelf.class().slots.basicsize
+        let mut res: usize = 0;
+        let isize: usize = zelf.class().slots.itemsize;
```
youknowone, line 333:
```suggestion
        let itemsize: usize = zelf.class().slots.itemsize;
```
to avoid reusing primitive type name `isize`

vm/src/object/core.rs
```diff
@@ -114,9 +114,23 @@ struct PyInner<T> {
...
     weak_list: WeakRefList,
 
+    size: PyRwLock<usize>, // PyVarObject.ob_size
+
     payload: T,
 }
 
+impl<T> PyInner<T> {
+    #[inline(always)]
+    pub fn ob_size(&self) -> usize {
```
youknowone, line 124: In RustPython, we don't use the prefix `ob_`. 

```suggestion
    pub fn size(&self) -> usize {
```

vm/src/object/core.rs
```diff
@@ -567,6 +582,16 @@ impl PyObjectRef {
...
         }
     }
+
+    #[inline(always)]
+    pub fn ob_size(&self) -> usize {
+        self.0.ob_size()
+    }
+
+    #[inline(always)]
+    pub fn set_ob_size(&self, size: usize) {
```
youknowone, line 592: where is this being called?

moreal:
> Does `int.__sizeof__` correctly work? If that correct, How about adding `slot_sizeof` and `SizeOf` trait? Then `object.__sizeof__` will call correct variants of each types

I agree about this suggestion because it doesn't need to implement same struct model with CPython. I'll implement the idea again by referring to other traits. If I implement it, some methods or fields will be removed or replaced. (e.g. `set_ob_size`).

moreal:
I wanted RustPython's implementations return same value with CPython implementation.
But I realized `object.__sizeof__` should return different value with other types like `str`. (`object.__sizeof__(_) != str.__sizeof__(_)`).

Of course, it isn't clean to set `ob_size` manually. I'll close this pull request but it may be because of my passion and my code skill.

youknowone:
oh.. I thought this is going to the right way. what was not fitting?

PR #3422, opened 2021-11-06
fix freebsd build wrong imports

devnexen:


stdlib/src/resource.rs
```diff
@@ -39,7 +39,7 @@ mod resource {
     #[pyattr]
     use libc::{RLIMIT_NPTS, RLIMIT_SBSIZE, RLIMIT_SWAP};
 
-    #[cfg(any(target_os = "freebsd", target_os = "solaris", target_os = "illumos"))]
+    #[cfg(any(target_os = "solaris", target_os = "illumos"))]
```
DimitrisJim, line 42: I was meaning to do this, thanks! Seems like it could be improved with the previous `cfg` though. I.e:

```rust
#[cfg(any(target_os = "freebsd", target_os = "netbsd", target_os = "solaris", target_os = "illumos"))]
#[pyattr]
use libc::{RLIMIT_NPTS, RLIMIT_SBSIZE, RLIMIT_SWAP};

#[cfg(any(target_os = "solaris", target_os = "illumos"))]
#[pyattr]
use libc::RLIMIT_VMEM;
```

DimitrisJim (approved):

PR #3424, opened 2021-11-07
os: update test `PidTests`

deantvv:
Follow up test for #3423

### DIFF
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -2701,7 +2701,7 @@ class PidTests(unittest.TestCase):
         # We are the parent of our subprocess
         self.assertEqual(int(stdout), os.getpid())
 
-    def check_waitpid(self, code, exitcode):
+    def check_waitpid(self, code, exitcode, callback=None):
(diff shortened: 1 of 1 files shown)


DimitrisJim:
Is this a partial update to the test file?

deantvv:
> Is this a partial update to the test file?

Yes. I was thinking about updating the whole file but it would change `Lib/test/support/` a lot and inevitably will be quite a large PR. I can do that if you prefer it that way.

This PR add one more passing test to windows but it doesn't add any more passing tests in unix since it requires `os.spawnv`.

DimitrisJim:
> Yes. I was thinking about updating the whole file but it would change Lib/test/support/ a lot and inevitably will be quite a large PR. I can do that if you prefer it that way.

That's generally preferable, if you're willing to update them! (If its too much of a hassle I'll take a stab at it when I find time for it)

deantvv:
Yes, I can update them. I will try to find some time next week for it. 
Closing for now, will open another PR once it is ready.

PR #3425, opened 2021-11-07
os: implement getrandom

deantvv:
`os.urandom` use the crate [getrandom](https://crates.io/crates/getrandom) to generate random but unfortunately that crate doesn't provide the needed `flags` according to [python doc](https://docs.python.org/3/library/os.html?highlight=os%20getrandom#os.getrandom). This PR implement `os.getrandom` by using `libc` binding.

### DIFF
--- a/vm/src/stdlib/posix.rs
+++ b/vm/src/stdlib/posix.rs
@@ -68,6 +68,10 @@ pub mod module {
     #[pyattr]
     use libc::{RTLD_GLOBAL, RTLD_LAZY, RTLD_LOCAL, RTLD_NOW};
 
+    #[cfg(target_os = "linux")]
+    #[pyattr]
+    use libc::{GRND_NONBLOCK, GRND_RANDOM};
+
     #[pyattr]
     const EX_OK: i8 = exitcode::OK as i8;
     #[pyattr]
@@ -1974,5 +1978,24 @@ pub mod module {
         );
         res.map_err(|err| err.into_pyexception(vm))?;
         Ok(vm.ctx.new_int(written as u64).into())
+    }
+
+    #[cfg(target_os = "linux")]
+    #[pyfunction]
+    fn getrandom(size: isize, flags: OptionalArg<u32>, vm: &VirtualMachine) -> PyResult<Vec<u8>> {
+        let size = usize::try_from(size)
+            .map_err(|_| vm.new_os_error(format!("Invalid argument for size: {}", size)))?;
+        let mut buf = Vec::with_capacity(size);
+        unsafe {
+            let len = libc::getrandom(
+                buf.as_mut_ptr() as *mut libc::c_void,
+                size,
+                flags.unwrap_or(0),
+            )
+            .try_into()
+            .map_err(|_| errno_err(vm))?;
+            buf.set_len(len);
+        }
+        Ok(buf)
     }
 }


vm/src/stdlib/posix.rs
```diff
@@ -1975,4 +1979,21 @@ pub mod module {
         res.map_err(|err| err.into_pyexception(vm))?;
         Ok(vm.ctx.new_int(written as u64).into())
     }
+
+    #[cfg(target_os = "linux")]
+    #[pyfunction]
+    fn getrandom(size: usize, flags: OptionalArg<u32>, vm: &VirtualMachine) -> PyResult<Vec<u8>> {
```
DimitrisJim, line 1985: CPython takes an (essentially) `isize` here and fails with `EINVAL`. Not sure if we can currently set `errno` to what we want but we could take an `isize`, try to convert it to a `usize` and then bail on failure with an `OSError` (We currently raise `OverflowError`).

deantvv: Fixed in new commit.

vm/src/stdlib/posix.rs
```diff
@@ -1975,4 +1979,23 @@ pub mod module {
...
+    #[pyfunction]
+    fn getrandom(size: isize, flags: OptionalArg<u32>, vm: &VirtualMachine) -> PyResult<Vec<u8>> {
+        let size = usize::try_from(size)
+            .map_err(|_| vm.new_os_error(format!("Invalid argument for size: {}", size)))?;
+        let mut buf = vec![0u8; size];
+        let ret = unsafe {
+            libc::getrandom(
+                buf.as_mut_ptr() as *mut libc::c_void,
+                buf.len(),
+                flags.unwrap_or_else(|| 0),
```
youknowone, line 1993: unwrap_or_else with const is same as just unwrap_or.
```suggestion
                flags.unwrap_or(0),
```

vm/src/stdlib/posix.rs
```diff
@@ -1975,4 +1979,23 @@ pub mod module {
...
+            )
+        };
+        if ret == -1 {
```
youknowone, line 1996: `ret` must be used as size because `GRND_RANDOM` flag allows it to be less then size.

vm/src/stdlib/posix.rs
```diff
@@ -1975,4 +1979,23 @@ pub mod module {
...
+            return Err(errno_err(vm));
+        }
```
youknowone, line 1998:
```suggestion
        let mut buf = Vec::with_capacity(size);
        unsafe {
            let len = libc::getrandom(
                buf.as_mut_ptr() as *mut libc::c_void,
                buf.len(),
                flags.unwrap(0),
            );
            if len == -1 {
                return Err(errno_err(vm));
            }
            buf.set_len(len as usize);
        }
        buf
```

youknowone (changes requested):

vm/src/stdlib/posix.rs
deantvv: Didn't know about this. Will update as suggested.

vm/src/stdlib/posix.rs
```diff
@@ -1975,4 +1979,24 @@ pub mod module {
...
+            libc::getrandom(
+                buf.as_mut_ptr() as *mut libc::c_void,
+                buf.len(),
+                flags.unwrap_or(0),
+            )
+        };
+        if len == -1 {
+            return Err(errno_err(vm));
+        }
+        buf.set_len(len as usize);
```
youknowone, line 1999: these lines also need to be placed in unsafe block

deantvv: Planed to use `shrink_to` which is safe and new (cool!)

youknowone: `buf.len()` here is `0`. `shrink_to` changes the capacity not the length.

deantvv: I end up using `truncate` which is safe and deals with dropping the rest of the elements.

youknowone:
I still want to recommend https://github.com/RustPython/RustPython/pull/3425#discussion_r744535023.
It will prevent unnecessary zero-initialization and safe unless `libc::getrandom` returns correct length. See the safety section of the method: https://doc.rust-lang.org/std/vec/struct.Vec.html#safety-2

deantvv:
@youknowone I changed the code as suggested in your comment. The rust document is quite helpful. Please take a look again once CI finished. Thanks a lot!

youknowone (approved):

PR #3426, opened 2021-11-07
netbsd/openbsd build fix

devnexen:


stdlib/src/socket.rs
```diff
@@ -61,9 +61,9 @@ mod _socket {
     #[pyattr]
     use c::{AF_UNIX, SO_REUSEPORT};
 
-    #[cfg(not(any(target_os = "freebsd", target_os = "netbsd", target_os = "openbsd")))]
+    #[cfg(not(any(target_os = "freebsd")))]
```
DimitrisJim, line 64: I think I added these to libc recently.

stdlib/src/resource.rs
```diff
@@ -41,8 +41,11 @@ mod resource {
         target_os = "solaris",
         target_os = "illumos"
     ))]
+    use libc::RLIMIT_SBSIZE;
```
DimitrisJim, line 44:
```suggestion
    #[pyattr]
    use libc::RLIMIT_SBSIZE;
```

DimitrisJim (approved):
lgtm, thanks!

PR #3427, opened 2021-11-08
[WIP] Introduce Structural Pattern Matching (Part1)

Snowapril:
This revision introduce structural pattern matching from cpython 3.10 which introduced in pep 634, 635, 636.

##### Simple example
(Syntax highlighting for it on github is not available yet 😂)
```python
# point is an (x, y) tuple
match point:
    case (0, 0):
        print("Origin")
    case (0, y):
        print(f"Y={y}")
    case (x, 0):
        print(f"X={x}")
    case (x, y):
        print(f"X={x}, Y={y}")
    case _:
        raise ValueError("Not a point")
```

If possible, I want to separate PR to two PRs (syntax, semantic).

### Python Documentation
*  https://www.python.org/dev/peps/pep-0634/
*  https://www.python.org/dev/peps/pep-0635/
*  https://www.python.org/dev/peps/pep-0636/

### DIFF
--- a/Lib/ast.py
+++ b/Lib/ast.py
@@ -47,8 +47,8 @@ def parse(source, filename='<unknown>', mode='exec', *,
     elif feature_version is None:
         feature_version = -1
     # Else it should be an int giving the minor version for 3.x.
-    return compile(source, filename, mode, flags)
-                   # _feature_version=feature_version)
+    return compile(source, filename, mode, flags,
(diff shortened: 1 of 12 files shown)


youknowone:
This is blocker of https://github.com/charliermarsh/ruff/issues/282
I think adding an actual match implementation will not be an easy job,
but adding match expr to parser would be enough with this PR.

PR #3428, opened 2021-11-08
fixed set.__ror__ operator to behave according to type

joohongpark:
I will also add sub, xor and and operations later.

### DIFF
--- a/vm/src/builtins/dict.rs
+++ b/vm/src/builtins/dict.rs
@@ -16,7 +16,7 @@ use crate::{
     IdProtocol, ItemProtocol,
     PyArithmeticValue::*,
     PyAttributes, PyClassDef, PyClassImpl, PyComparisonValue, PyContext, PyObject, PyObjectRef,
-    PyObjectView, PyRef, PyResult, PyValue, TypeProtocol,
+    PyObjectView, PyRef, PyResult, PyValue, TryFromObject, TypeProtocol,
 };
 use rustpython_common::lock::PyMutex;
 use std::fmt;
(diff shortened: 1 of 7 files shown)


vm/src/builtins/set.rs
```diff
@@ -485,30 +485,60 @@ impl PySet {
...
-        self.union(other.iterable, vm)
+    fn or(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyArithmeticValue<Self>> {
+        if other.isinstance(&vm.ctx.types.set_type)
+            || other.isinstance(&vm.ctx.types.frozenset_type)
+        {
+            let val = PosArgs::new(vec![ArgIterable::try_from_object(vm, other)?]);
+            Ok(PyArithmeticValue::Implemented(self.union(val, vm)?))
+        } else {
+            Ok(PyArithmeticValue::NotImplemented)
+        }
```
DimitrisJim, line 496: `SetIterable` does most of this, it just doesn't return the right things. I think taking inspiration from what `PyInt` does (with `int_op`) would be the best thing here (a function that tries to create a `SetIterable` from the `PyObject` ref and then forwards the call to the appropriate set/frozenset method).

joohongpark: thank you! I will refer to the information you provided and try to correct it! 🙏🏼

joohongpark: I have a question. As you said, I referred to how int_op works in PyInt, which means that PySet should add functions such as set_op to PySet to implement similarly? Or does it mean adding another function to the SetItable structure?

vm/src/builtins/dict.rs
```diff
@@ -946,14 +946,23 @@ trait ViewSetOps: DictView {
         Ok(PySet { inner })
     }
 
-    #[pymethod(name = "__rsub__")]
+    #[pymethod(name = "__sub__")]
```
youknowone, line 949: `__sub__` is handled by below `#[pymethod(magic)]`. Remove this line if you are going to add separated rsub

vm/src/builtins/dict.rs
```diff
@@ -946,14 +946,23 @@ trait ViewSetOps: DictView {
...
     #[pymethod(magic)]
     fn sub(zelf: PyRef<Self>, other: ArgIterable, vm: &VirtualMachine) -> PyResult<PySet> {
         let zelf = Self::to_set(zelf, vm)?;
         let inner = zelf.difference(other, vm)?;
         Ok(PySet { inner })
     }
 
+    #[pymethod(name = "__rsub__")]
```
youknowone, line 957: same here

youknowone (changes requested):

youknowone (approved):

youknowone:
looks good. please use rebase rather than merge next time.

PR #3433, opened 2021-11-13
Update test for `support/os_helper.py`

deantvv:
Plan to update all code under `Lib/test/support`.
This PR only cover the change for `Lib/test/support/os_helper.py` and reflect the change necessary for other test to use.
It is already a big one so I decide to make a PR for this and plan for others (import_helper, socket_helper...) to go for another PR.
Mostly of the part is `from test.support import TESTFN` --> `from test.support.os_helper import TESTFN` etc..

### DIFF
--- a/Lib/test/libregrtest/cmdline.py
+++ b/Lib/test/libregrtest/cmdline.py
@@ -2,6 +2,7 @@ import argparse
 import os
 import sys
 from test import support
+from test.support import os_helper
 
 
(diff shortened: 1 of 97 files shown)


deantvv:
The test failure seems not related to this PR and also failed on master CI test.

youknowone:
I thought the failing test is only about broken CI on main branch but more windows tests looks failing

PR #3436, opened 2021-11-14
pymethod documentations for set functions

nandhinianandj:
Address the documentation for set object part of #2680


Snowapril:
Seems good though you need rebasing. Rather than merge main into your branch, rebasing is preferred 😊.
And it seems only last commit is desired one, so dropping other commits except [4c3d3b3](https://github.com/RustPython/RustPython/pull/3436/commits/4c3d3b3fc07c58d455b52676389f72b6d47fe8c0) will be good.

PR #3316, opened 2021-10-15
Impl PySequence Protocol

qingshi163:
#3244

PR #3437, opened 2021-11-15
Dict keys isdisjoint

rvsampson94:
Addresses part of issue #2898.
After this and pull request #3421 are merged, issue #2898 will be resolved.

### DIFF
--- a/Lib/test/test_dictviews.py
+++ b/Lib/test/test_dictviews.py
@@ -90,8 +90,6 @@ class DictSetTest(unittest.TestCase):
         self.assertTrue(r == "dict_values(['ABC', 10])" or
                         r == "dict_values([10, 'ABC'])")
 
-    # TODO: RUSTPYTHON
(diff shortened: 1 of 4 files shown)


Snowapril:
This seems work properly as `isdisjoint` method at now.
though, `to_set` is quite expensive method as it makes a redundant copy. 
After #3316 merged, we need to modify it with `PySequence_Contains` like cpython implementation 😊

youknowone (dismissed):

rvsampson94:
I will get the tests passing and resubmit. @Snowapril would you like me to add a TODO to come back and update this implementation after #3316 is merged?

Snowapril:
@rvsampson94 Oh yes of course. but rather than that, you could add TODO comment in `isdisjoint`. maybe it is better

youknowone (approved):
looks great, thank you!

PR #3438, opened 2021-11-15
documentation for set method

nandhinianandj:
PR for documentation of set method for issue #2680


DimitrisJim:
thanks for this! We've changed how we handle documentation now to generate it automatically instead. Closing in preference to #3101

PR #3439, opened 2021-11-15
relocate wrongly placed sys test in snippets/stdlib_os.py

youknowone:


youknowone:
this is also fixing(hiding) CI error

PR #3445, opened 2021-11-18
Refactor and new sequence traits, generic sequence operation

qingshi163:


qingshi163:
#3316

vm/src/stdlib/collections.rs
```diff
@@ -431,32 +384,32 @@ mod _collections {
...
         }
 
-        #[pymethod(magic)]
-        #[pymethod(name = "__rmul__")]
-        fn mul(&self, value: isize, vm: &VirtualMachine) -> PyResult<Self> {
-            let deque: SimpleSeqDeque = self.borrow_deque().into();
-            let mul = sequence::seq_mul(vm, &deque, value)?;
+        fn _mul(&self, n: isize, vm: &VirtualMachine) -> PyResult<VecDeque<PyObjectRef>> {
+            let deque = self.borrow_deque();
+            let n = vm.check_repeat_or_memory_error(deque.len(), n)?;
```
qingshi163, line 397: bit confuse, should it return overflow error rather than memory error?

qingshi163: after reading the issue 45044 from cpython, I think we should replace all the memory error with overflow error because currently we did not catch the malloc failing.
What do you think? @DimitrisJim @youknowone

youknowone: that sounds like reasonable

vm/src/sequence.rs
```diff
@@ -1,107 +1,298 @@
...
+    fn mut_index(&'a self, vm: &VirtualMachine, needle: &PyObject) -> PyResult<Optioned<usize>> {
+        self.mut_index_range(vm, needle, 0..isize::MAX as usize)
+    }
+
+    fn mut_contains(&'a self, vm: &VirtualMachine, needle: &PyObject) -> PyResult<bool> {
+        self.mut_index(vm, needle).map(|x| x.is_some())
+    }
+
+    #[inline]
+    fn _mut_iter_equal_skeleton<F, const SHORT: bool>(
```
Snowapril, line 103: This method seems somewhat complex despite it declared as `inline`. How do you think?

qingshi163: This one is the skeleton for other related function, only few place is calling it. Also as it is generic, so it should be duplicate for each generic parameters. I mark it inline to try to hint the compiler to optimize out the const bool condition check and the closure it called. I don't know what exactly code the compiler produce but inline is what I expected.

youknowone (changes requested):

youknowone (approved):

PR #3446, opened 2021-11-20
pystr.rs isdecimal() checks not only ASCII but also Unicode decimals

BongYang:


youknowone (approved):
looks good to me, thanks!

PR #3447, opened 2021-11-20
arrow crash in windows

waroad:
in windows, arrow is not shown
and crahsed.
changed to another arrow that's
compatible


youknowone:
@coolreader18 I am going to accept this PR because plain arrow is better than a crash. If we want to go back to the original fancy arrow, I think there must be a windows platform and charset check

youknowone (approved):
Thank you! This change looks reasonable

PR #3448, opened 2021-11-20
Update test_math.py to cpython 3.10

YYun-D:
* update `test_math.py` to cpython 3.10
* fix failed test, `testFactorialNonIntegers`


youknowone (approved):
Thank you!

PR #3450, opened 2021-11-20
implement `reduce` method to `PySlice`

YYun-D:
* implement `reduce` method to `PySlice`
* fix `test_pickle` in `test_slice`


youknowone (approved):
looks good! thank you!

PR #3451, opened 2021-11-20
add reduce in enumerate

waroad:
fixed test_enumerate


youknowone (approved):
looks good, thank you!

PR #3452, opened 2021-11-20
Remove staled comments

moreal:
https://github.com/RustPython/RustPython/issues/2410 was resolved by #2904. So this pull request removes the staled lines from `extra_tests/not_impl_gen.py`


youknowone (approved):
nice catch!

PR #3459, opened 2021-11-27
Fix maketrans not thrown ValueError when length of the arguments are …

TeamTamoad:
…not the same


youknowone (approved):

PR #3460, opened 2021-11-27
Fix compatibility of Tuple's hash function

TeamTamoad:
Change hash function implementation of Tuple to compatible with CPython's implementation.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated tuple hashing to use a new algorithm that produces platform-specific hash values, improving consistency with expected hash outputs.
  * Adjusted tests so that tuple hash tests are now required to pass, reflecting the improved hashing implementation.

### DIFF
--- a/Lib/test/test_tuple.py
+++ b/Lib/test/test_tuple.py
@@ -77,8 +77,6 @@ class TupleTest(seq_tests.CommonTest):
     # We expect tuples whose base components have deterministic hashes to
     # have deterministic hashes too - and, indeed, the same hashes across
     # platforms with hash codes of the same bit width.
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_hash_exact(self):
         def check_one_exact(t, e32, e64):
             got = hash(t)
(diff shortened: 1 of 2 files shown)


youknowone (changes requested):
The change looks great, but by looking at the CI result, `range()` hash must follow the reuslt of `tuple`. Could you also update `range.__hash__` in the same way?

You can extract hash function like `fn tuple_hash(elements: &[PyObjectRef], vm: &VirtualMachine)` and call it form `PyTuple::hash` like `tuple_hash(zelf.as_slice())`

youknowone:
The new bug is bug of `exec`. Somehow when `__hash__` is added by `exec`, it doesn't run `update_slot`.
As the result, the upper test passes but the lower test fails

```
import dataclasses

@dataclasses.dataclass(unsafe_hash=True)
class X:
    pass


a = X().__hash__()
b = ().__hash__()
assert a == b, '{} {}'.format(a, b)

a = hash(X())
b = hash(())
assert a == b, '{} {}'.format(a, b)
```

youknowone:
@coolreader18 do you have any idea?

youknowone (approved):
Great, I have no idea which certain point made the changes, but it now works!

PR #3462, opened 2021-11-27
Refactor Mapping Protocol and Item Protocol

qingshi163:
#3316

### DIFF
--- a/vm/src/function/argument.rs
+++ b/vm/src/function/argument.rs
@@ -1,7 +1,10 @@
-use super::IntoFuncArgs;
+use super::{IntoFuncArgs, IntoPyObject};
 use crate::{
-    builtins::iter::PySequenceIterator, protocol::PyIter, protocol::PyIterIter, PyObject,
-    PyObjectRef, PyObjectWrap, PyResult, PyValue, TryFromObject, TypeProtocol, VirtualMachine,
+    builtins::{iter::PySequenceIterator, PyDict, PyDictRef},
+    protocol::PyIter,
+    protocol::{PyIterIter, PyMapping, PyMappingMethods},
+    PyObject, PyObjectRef, PyObjectWrap, PyResult, PyValue, TryFromObject, TypeProtocol,
+    VirtualMachine,
 };
 use std::marker::PhantomData;
 
@@ -83,6 +86,54 @@ where
             iterable: obj,
             iterfn,
             _item: PhantomData,
+        })
+    }
+}
+
+#[derive(Clone)]
+pub struct ArgMapping {
+    obj: PyObjectRef,
+    mapping_methods: PyMappingMethods,
+}
+
+impl ArgMapping {
+    pub fn from_dict_exact(dict: PyDictRef) -> Self {
+        Self {
+            obj: dict.into(),
+            mapping_methods: PyDict::MAPPING_METHODS,
+        }
+    }
+
+    pub fn mapping(&self) -> PyMapping {
+        PyMapping::with_methods(&self.obj, self.mapping_methods)
+    }
+}
+
+impl AsRef<PyObject> for ArgMapping {
+    fn as_ref(&self) -> &PyObject {
+        &self.obj
+    }
+}
+
+impl PyObjectWrap for ArgMapping {
+    fn into_object(self) -> PyObjectRef {
+        self.obj
+    }
+}
+
+impl IntoPyObject for ArgMapping {
+    fn into_pyobject(self, _vm: &VirtualMachine) -> PyObjectRef {
+        self.obj
+    }
+}
+
+impl TryFromObject for ArgMapping {
+    fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
+        let mapping = PyMapping::try_protocol(&obj, vm)?;
+        let mapping_methods = *mapping.methods(vm);
+        Ok(Self {
+            obj,
+            mapping_methods,
         })
     }
 }
--- a/vm/src/protocol/mapping.rs
+++ b/vm/src/protocol/mapping.rs
@@ -1,82 +1,116 @@
 use crate::{
     builtins::{
-        dict::{PyDictKeys, PyDictValues},
-        PyDictRef, PyList,
+        dict::{PyDictItems, PyDictKeys, PyDictValues},
+        PyDict,
     },
-    function::IntoPyObject,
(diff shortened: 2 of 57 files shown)


Snowapril:
Oh yes. `ItemProtocol` is one of stuffs that should be moved in object protocol. Refactored mapping protocol also seems better than before.

youknowone:

youknowone (changes requested):

youknowone:
does PyMapping and ArgMapping separated to keep PyObjectRef in ArgMapping?

vm/src/protocol/mapping.rs
```diff
@@ -1,126 +1,167 @@
...
-{
-    pub fn new(obj: T) -> Self {
-        Self(obj)
+impl PyMapping<'_> {
+    // PyMapping::Check
+    pub fn has_protocol(&self, vm: &VirtualMachine) -> bool {
+        self.methods(vm).subscript.is_some()
+    }
+
+    pub fn obj_as<T: PyObjectPayload>(&self) -> &PyObjectView<T> {
```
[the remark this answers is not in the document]
qingshi163: This helper function is used to downcast the object to the type who provide the methods and should not used by any case if the type is not certain. maybe add the comment will be better than mark it unsafe because it is used in each methods to get the instance of Self?

vm/src/function/argument.rs
```diff
@@ -86,3 +89,51 @@ where
...
+    mapping_methods: PyMappingMethods,
+}
+
+impl ArgMapping {
+    pub fn from_dict_exact(dict: PyDictRef) -> Self {
+        Self {
+            obj: dict.into(),
+            mapping_methods: PyDict::MAPPING_METHODS,
+        }
+    }
```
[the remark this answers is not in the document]
qingshi163: is PyDictRef means exact PyDict not the subclass?

qingshi163:
> does PyMapping and ArgMapping separated to keep PyObjectRef in ArgMapping?

PyMapping do not take the ownership, it bind the lifetime to &PyObject. ArgMapping keeps PyObjectRef and the cached methods to avoid mro search.

vm/src/function/argument.rs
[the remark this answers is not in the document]
youknowone: oh, i am sorry. exact meant that exact.

vm/src/protocol/mapping.rs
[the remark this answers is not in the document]
youknowone: That sounds like this method is strictly required to be call like `obj_as::<Self>` always. In that case, the function is unsafe beacuse the guarantee of `Self` came from user, not from automatic constraints.
If you want to have a safe version of the function, I think a trait can wrap it safely. Maybe `AsMapping` can be the place.

```rust
trait AsMapping {
    fn mapping_as(mapping: ...) -> &PyObjectView<Self> {
        unsafe { mapping.obj_as::<Self>() }
    }
   ...
}
```

vm/src/protocol/mapping.rs
```diff
@@ -1,126 +1,167 @@
...
-where
-    T: Borrow<PyObject>,
-{
-    pub fn new(obj: T) -> Self {
-        Self(obj)
+impl PyMapping<'_> {
+    // PyMapping::Check
+    pub fn has_protocol(&self, vm: &VirtualMachine) -> bool {
+        self.methods(vm).subscript.is_some()
+    }
```
Snowapril, line 66: How about just using `check`? PyIter also adopt `check` for `PyIter::Check`

vm/src/protocol/object.rs
```diff
@@ -411,4 +415,92 @@ impl PyObject {
...
+        vm: &VirtualMachine,
+    ) -> PyResult<()> {
+        if let Some(dict) = self.downcast_ref_if_exact::<PyDict>(vm) {
+            return dict.set_item(needle, value, vm);
+        }
+
+        let needle = needle.into_pyobject(vm);
+
+        if let Ok(mapping) = PyMapping::try_protocol(self, vm) {
+            mapping.ass_subscript(&needle, Some(value), vm)
```
Snowapril, line 468: According to [cpython code](https://github.com/python/cpython/blob/main/Objects/abstract.c#L208-L213), they check `ass_subscript` exist then call it, otherwise pass. For now, It seems we just check only mapping & `subscript` and call `ass_subscript`

vm/src/protocol/object.rs
```diff
@@ -411,4 +415,92 @@ impl PyObject {
...
+    ) -> PyResult<()> {
+        if let Some(dict) = self.downcast_ref_if_exact::<PyDict>(vm) {
+            return dict.del_item(needle, vm);
+        }
+
+        let needle = needle.into_pyobject(vm);
+
+        if let Ok(mapping) = PyMapping::try_protocol(self, vm) {
+            mapping.ass_subscript(&needle, None, vm)
+        } else {
```
Snowapril, line 496: Same with `set_item`. We must check `ass_subscript` exist before calling it

vm/src/protocol/mapping.rs
qingshi163: make sense, but I think I'd like to keep fn try_protocol() as it is more rust style.

vm/src/types/slot.rs
```diff
@@ -163,31 +163,37 @@ fn as_mapping_wrapper(zelf: &PyObject, _vm: &VirtualMachine) -> PyMappingMethods
...
+                            })
                         })
-                    })
-                    .unwrap()
-            })?
+                        .unwrap()
+                })?
+        }),
+        subscript: then_some_closure!(zelf.has_class_attr("__getitem__"), |mapping, needle, vm| {
+            vm.call_special_method(mapping.obj.to_owned(), "__getitem__", (needle.to_owned(),))
```
Snowapril, line 179: Although it is not directly relevant with this PR, as we update all the mapping methods it became incompatibility between cpython 3.10.

```python
>>> class TestDict(dict):
...     def __setitem__(self, key, value):
...             pass

>>> test1 = dict([[1,2],[3,4]])
>>> test2 = TestDict([[1,2],[3,4]])
```
##### RustPython
```python
>>>>> id(test1)
4821477152
>>>>> id(test2)
4318054400
```
##### cpython
```python
>>> id(test1)
4308420304
>>> id(test2)
4308420304
```
As we update `__setitem__` only, `__getitem__` must be same with base class.

I don't know what this difference makes problem yet. but just for mind up 😊

youknowone (approved):

PR #3465, opened 2021-11-27
Support tuple type also for memoryview shape arg

24seconds:
## Description

This pr is to fix [test_base64.py](https://github.com/24seconds/RustPython/blob/dd773b3413017955db8639968b0a286adfa05e61/Lib/test/test_base64.py) test failure. Turned out most of them are related to `memoryview`'s `cast` function.

In the Cpython,  `cast` function's argument `shape` seems to support `list` and `tuple` type.
https://github.com/python/cpython/blob/af9ee57b96cb872df6574e36027cc753417605f9/Objects/memoryobject.c#L1367-L1418  

But RustPython currently supports `list` only. 
So this pr did those below
- Change `CastArgs` shape field type to `OptionalArg<PyObjectRef>`
- Check shape type is one of list or tuple
- convert shape properly

Because many `test_base64.py` tests are related to this problem, most of them are resolved after fix.

For the check, I ran these. If there is another thing that I should do, please give me guide :bow: 
```
cargo check
cargo fmt
cargo clippy
```

### DIFF
--- a/vm/src/builtins/memory.rs
+++ b/vm/src/builtins/memory.rs
@@ -10,6 +10,7 @@ use crate::{
     bytesinner::bytes_to_hex,
     function::{FuncArgs, IntoPyObject, OptionalArg},
     protocol::{BufferDescriptor, BufferMethods, PyBuffer, PyMappingMethods, VecBuffer},
+    sequence::SequenceOp,
     sliceable::wrap_index,
     stdlib::pystruct::FormatSpec,
     types::{AsBuffer, AsMapping, Comparable, Constructor, Hashable, PyComparisonOp},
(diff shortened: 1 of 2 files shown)


vm/src/builtins/memory.rs
```diff
@@ -865,7 +876,7 @@ struct CastArgs {
     #[pyarg(any)]
     format: PyStrRef,
     #[pyarg(any, optional)]
-    shape: OptionalArg<PyListRef>,
+    shape: OptionalArg<PyObjectRef>,
```
youknowone, line 879: If they take only `list` and `tuple`, `Either` works.
```suggestion
    shape: OptionalArg<Either<PyTupleRef, PyListRef>>,
```

If it takes any iterable, `ArgIterable` will fit.

youknowone (commented):
We usually use types to narrow down to specific types when it is possible. Could you update it using `Either`?

vm/src/builtins/memory.rs
24seconds: Wow good. Let me try `Either`

24seconds (commented):
65d9e2f, NICE!

youknowone (approved):
Nice work, thank you!

PR #3470, opened 2021-11-28
enable lto for release build

### LINKED ISSUE #1978: Support and enable LTO by default
In order to reduce size and maybe increase performance, link-time-optimization could be enabled.
I tried these commands:
set RUSTFLAGS=-Clto
cargo install rustpython

I got this error:
error: cannot prefer dynamic linking when performing LTO

note: only 'staticlib', 'bin', and 'cdylib' outputs are supported with LTO

error: could not compile `proc-macro-hack`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: failed to compile `rustpython v0.1.2`, intermediate artifacts can be found at `C:\Users\Keith\AppData\Local\Temp\cargo-installbvSdKc`

Caused by:
  build failed
I'm on windows. This problem seems to be with both theh Mingw and MSVC toolchains.

youknowone:
fix #1978

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -75,9 +75,12 @@ opt-level = 3
 # lto = "thin"
 
 [profile.bench]
-lto = true
+lto = "thin"
 codegen-units = 1
 opt-level = 3
+
+[profile.release]
+lto = "thin"
 
 [patch.crates-io]
 # REDOX START, Uncommment when you want to compile/check with redoxer


fanninpm:
I think it's also time to revert part of #3516, as the solution to https://github.com/rust-lang/rust/issues/92869 has been on stable for several months.

youknowone:
That sounds like a good idea

PR #3471, opened 2021-11-28
refactor: add a repr helper function for the collections

TeamTamoad:
Implement an helper functions to for repr function of the collections including list, tuple, set and deque.

### DIFF
--- a/vm/src/utils.rs
+++ b/vm/src/utils.rs
@@ -127,3 +127,41 @@ impl ToCString for PyStr {
         std::ffi::CString::new(self.as_ref()).map_err(|err| err.into_pyexception(vm))
     }
 }
+
+pub(crate) fn collection_repr<'a, I>(
+    class_name: Option<&str>,
+    prefix: &str,
+    suffix: &str,
+    iter: I,
+    vm: &VirtualMachine,
+) -> PyResult<String>
+where
+    I: std::iter::Iterator<Item = &'a PyObjectRef>,
+{
+    let mut repr = String::new();
+    if let Some(name) = class_name {
+        repr.push_str(name);
+        repr.push('(');
+    }
+    repr.push_str(prefix);
+    {
+        let mut parts_iter = iter.map(|o| o.repr(vm));
+        repr.push_str(
+            parts_iter
+                .next()
+                .transpose()?
+                .expect("this is not called for empty collection")
+                .as_str(),
+        );
+        for part in parts_iter {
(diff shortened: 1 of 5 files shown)


vm/src/utils.rs
```diff
@@ -127,3 +127,40 @@ impl ToCString for PyStr {
         std::ffi::CString::new(self.as_ref()).map_err(|err| err.into_pyexception(vm))
     }
 }
+
+pub(crate) fn repr_helper<'a, I>(
+    class_name: Option<&str>,
+    brackets: [char; 2],
```
youknowone, line 133: This variable is always used only for each char, but never as array. If we change this to `prefix: &str, suffix: &str`, we can handle more cases. For example, `deque` is now formatted again using maxlen. But if we just pass suffix string, we can pass right bracket and maxlen at same time.

youknowone (changes requested):

TeamTamoad:
Those changes are really good ideas. I'll do that.
Btw, do you have any idea why my last changes don't pass the test in time ?

youknowone:
I have no idea yet. But becasue it failed again exceeding the time limit, it seems this changes made something.

fanninpm:
> Btw, do you have any idea why my last changes don't pass the test in time ?

Try running the following command:

```text
cargo run --release --features ssl,jit -- -m test -v
```

fanninpm:
[It was hanging in the snippets.](https://github.com/RustPython/RustPython/runs/4857239926) To diagnose,

1. Build RustPython in release mode (with the features normally used for testing): `cargo build --release --features ssl,jit`
2. Navigate to the `extra_tests` directory and install `pytest` into a virtual environment. (e.g., `cd extra_tests && pipenv install --python 3.9`)
3. Run pytest on the `extra_tests` directory. (e.g., `pipenv run pytest -v`)

pwnorbitals:
how far are we from completion here @TeamTamoad ? :)

TeamTamoad:
Sorry for an late update. I'm super busy with life and college right now.
The hanging problem is cause by this snippet in file `extra_tests/snippets/stdlib_collections.py`
```python
# Test that calling an evil __repr__ can't hang deque
class BadRepr:
    def __repr__(self):
        self.d.pop()
        return ''

b = BadRepr()
d = deque([1, b, 2])
b.d = d
repr(d)
```
It should be fixed now

youknowone (approved):
great! thank you for the long time effort!

Issue #2889, opened 2021-08-16
`./rustpython -m dis`: No code object available for dis


whjpji:
## Feature

`dis`: Disassembler for Python bytecode.

I want to should compiled bytecode for a `.py` script, and I got this error:
```
$ ./rustpython -m dis xxx.py
./rustpython: No code object available for dis
```

In cpython, most of the std module will be pre-compiled during build.

## Python Documentation

https://docs.python.org/3.8/library/dis.html

fanninpm:
See also #367

coolreader18:
This is because dis is implemented as a native module, not using cpython's dis.py that's specific to their bytecode format.

youknowone:
I think we can move our native dis module to `_dis` and create our own `dis.py` with `_test` in CPython `dis.py`

youknowone:
I expect `cargo run -- -m dis x.py` works like `python -m dis x.py` after resolving this issue

tai271828:
@youknowone thanks for #3469 , for me it is working like

```
(rustpython-session) [^_^]─[~/work-my-projects/pyconapac-2021-sprint/rustpython-session/RustPython] tai271828@syakaro: 35 file 300Kb
$ cargo run -- -m dis ./demo.py 
   Compiling rustpython-pylib v0.1.0 (/home/tai271828/work-my-projects/pyconapac-2021-sprint/rustpython-session/RustPython/vm/pylib-crate)
   Compiling rustpython-stdlib v0.1.2 (/home/tai271828/work-my-projects/pyconapac-2021-sprint/rustpython-session/RustPython/stdlib)
   Compiling rustpython-vm v0.1.2 (/home/tai271828/work-my-projects/pyconapac-2021-sprint/rustpython-session/RustPython/vm)
   Compiling rustpython v0.1.2 (/home/tai271828/work-my-projects/pyconapac-2021-sprint/rustpython-session/RustPython)
    Finished dev [unoptimized + debuginfo] target(s) in 1m 05s
     Running `target/debug/rustpython -m dis ./demo.py`
(rustpython-session) [^_^]─[~/work-my-projects/pyconapac-2021-sprint/rustpython-session/RustPython] tai271828@syakaro: 35 file 300Kb
$ 
```

Issue #367, opened 2019-02-06
Add support for dis module


cthulahoops:
Lack of this makes compiler debugging more painful than it should be.

Will need us to make the code object work fully.

ksunden:
`dis` module is pure python and appears to work, at least as far as my simple test goes

```
>>>>> def fib(a):
.....     return a if a < 2 else fib(a-1) + fib(a-2)
>>>>> for i in range(10):
.....    print(fib(i))
0
1
1
2
3
5
8
13
21
34
>>>>> dis.dis(fib)
                 0 LoadName             (a)
                 1 LoadConst            (2)
                 2 CompareOperation     (Less)
                 3 Duplicate
                 4 JumpIfFalse          (8)
                 5 Pop
                 6 LoadName             (a)
                 7 Jump                 (19)
          >>     8 LoadName             (fib)
                 9 LoadName             (a)
                10 LoadConst            (1)
                11 BinaryOperation      (Subtract, false)
                12 CallFunction         (Positional(1))
                13 LoadName             (fib)
                14 LoadName             (a)
                15 LoadConst            (2)
                16 BinaryOperation      (Subtract, false)
                17 CallFunction         (Positional(1))
                18 BinaryOperation      (Add, false)
          >>    19 ReturnValue
                20 LoadConst            (None)
                21 ReturnValue
```

(I have `PYTHONPATH=/usr/lib/python3.6/:/usr/lib/python3.6/site-packages`, just my distro provided python 3.6 standard lib)

cthulahoops:
I really should have linked to this issue with my pull requests:

#512 and #555 add the beginnings of a rust implementation of the dis module, which is what you're seeing there.

Only dis and disassemble are implemented. (And dis doesn't support everything it should yet.)

As discussed in #445, the internals of the VM are too different to make compatibility with the python implementation of dis sensible at the moment.

This is still open as I'm still planning on fleshing out the dis module a bit more.

youknowone:
is this still open?

youknowone:
basic one is implemented

PR #3473, opened 2021-11-28
feat(Lib): print the disassembled info

tai271828:
This pull resolves Issue: #2889 

**Steps to Test This Pull Request**
```
$ target/debug/rustpython -m dis ./demo.py

                 0 LoadNameAny          (0, print)
                 1 LoadConst            ("Hello, RustPython!")
                 2 CallFunctionPositional (1)
                 3 Pop
                 4 LoadConst            (None)
                 5 ReturnValue
```

### DIFF
--- a/Lib/dis.py
+++ b/Lib/dis.py
@@ -1,1 +1,18 @@
 from _dis import *
+
+
+# Disassembling a file by following cpython Lib/dis.py
+def _test():
+    """Simple test program to disassemble a file."""
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument('infile', type=argparse.FileType('rb'), nargs='?', default='-')
+    args = parser.parse_args()
+    with args.infile as infile:
+        source = infile.read()
+    code = compile(source, args.infile.name, "exec")
+    dis(code)
+
+if __name__ == "__main__":
+    _test()


tai271828:
Maybe also related to issue #367

youknowone (approved):
thank you for contributing!

youknowone:
Hmm, something is broken

Lib/dis.py
```diff
@@ -1 +1,12 @@
...
+
+
+# Disassembling a file by following cpython Lib/dis.py
+parser = argparse.ArgumentParser()
+parser.add_argument('infile', type=argparse.FileType('rb'), nargs='?', default='-')
+args = parser.parse_args()
+with args.infile as infile:
+    source = infile.read()
+code = compile(source, args.infile.name, "exec")
+_dis.dis(code)
```
youknowone, line 12: This file is not only used for the command `rustpyhon -m dis` but also used as an independent module like `import dis` in common python code.
So this new codes must be isolated by guarding `if __name__ == "__main__":` just like CPython handle it.

youknowone (changes requested):

Lib/dis.py
```diff
@@ -1 +1,19 @@
-from _dis import *
```
coolreader18, line 1: I think you should keep this line, and then simply call `dis()` in `_test`

Lib/dis.py
```diff
@@ -1 +1,19 @@
...
+import argparse
```
coolreader18, line 1: Since there's also an `import argparse` inside the function, this can be removed

tai271828: Oooops, right. Thanks for the good catch!

tai271828:
Thanks @coolreader18 's review. Re-based and re-submitted. This pull request is ready to review again.

coolreader18 (approved):

Issue #3467, opened 2021-11-28
Add `usedforsecurity` optional keyword argument to hashlib functions


youknowone:
https://docs.python.org/3/library/hashlib.html

> Changed in version 3.9: All hashlib constructors take a keyword-only argument usedforsecurity with default value True. A false value allows the use of insecure and blocked hashing algorithms in restricted environments. False indicates that the hashing algorithm is not used in a security context, e.g. as a non-cryptographic one-way compression function.

We can handle this in 2-steps.

1. Adding an optional parameter `usedforsecurity` everywhere.
2. Actually use the value.

TeamTamoad:
I'll take this issue

TeamTamoad:
The first step is completed and merged. #3474 
> 1. Adding an optional parameter `usedforsecurity` everywhere.

Anyone can take this issue and continue working on the second step
> 2. Actually use the value.

PR #3474, opened 2021-11-28
feat(hashlib): add usedforsecurity kwarg to hash functions

TeamTamoad:
This PR is related to #3467 
I add "usedforsecurity" as a optional keyword argument to hash functions in hashlib.
I also copy test_hashlib.py from CPython v3.9.9 and mark some tests expectedFailure

### DIFF
--- a/Lib/test/test_hashlib.py
+++ b/Lib/test/test_hashlib.py
@@ -1,0 +1,1121 @@
+# Test hashlib module
+#
+# $Id$
+#
+#  Copyright (C) 2005-2010   Gregory P. Smith (greg@krypto.org)
+#  Licensed to PSF under a Contributor Agreement.
+#
+
+import array
+from binascii import unhexlify
+import hashlib
+import importlib
+import itertools
+import os
+import sys
+import sysconfig
+import threading
+import unittest
+import warnings
+from test import support
+from test.support import _4G, bigmemtest, import_fresh_module
+from http.client import HTTPException
+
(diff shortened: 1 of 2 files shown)


youknowone (changes requested):
The changed code looks perfect. But we are commiting CPython-originated files separatedly with its source version name. Could you split and message the commits like this PR?

https://github.com/RustPython/RustPython/pull/3207/commits

TeamTamoad:
Do I need to split this PR into 2 PRs ? or I can just split this PR into 2 commits?

youknowone:
single PR with 2 commits will be the best

youknowone (approved):
Looks good, thank you!

youknowone:
@TeamTamoad if you are interested in further contribution, `uuid.py` probably now can be updated to Python 3.9 since this was a blocker of them.

TeamTamoad:
> @TeamTamoad if you are interested in further contribution, `uuid.py` probably now can be updated to Python 3.9 since this was a blocker of them.

OK I'll work on that. Can I just copy `uuid.py` from CPython and update the `test_uuid.py`? Do I need to do anything in particular?

youknowone:
`test_uuid.py` is already very updated. I think updating `uuid.py` matches to `test_uuid.py` would be great. Just like you handled test_hashlib.py here, we make separated commit with version message for CPython originated files.

TeamTamoad:
Done #3489

PR #3476, opened 2021-11-28
Update test for support/os_helper.py (second attempt)

deantvv:
Second attempt to update `test/support/os_helper.py` (original #3433)

Plan to update all code under Lib/test/support.
This PR only cover the change for Lib/test/support/os_helper.py and reflect the changes necessary for other tests to use.
It is already a big one so I decide to make a PR for this and plan for others (import_helper, socket_helper...) to go for another PR.
Mostly of the part is from test.support import TESTFN --> from test.support.os_helper import TESTFN etc..

Follow up for #3424

### DIFF
--- a/Lib/test/libregrtest/cmdline.py
+++ b/Lib/test/libregrtest/cmdline.py
@@ -2,6 +2,7 @@ import argparse
 import os
 import sys
 from test import support
+from test.support import os_helper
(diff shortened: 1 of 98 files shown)


youknowone (approved):
great! I was waiting for this

PR #3477, opened 2021-11-28
Simplify android ssl

coolreader18:
Still not sure about unconditionally using android system certs - on e.g. termux the ssl library provides its own, and I'd think so should android's libssl if it's dynamically linked in. If not dynamically linked, the path should be added to openssl-probe and it can take care of it. I'm on my phone, or else I'd find the person who added it originally and ask the use case.

Sent from my Moto z3

### DIFF
--- a/stdlib/src/ssl.rs
+++ b/stdlib/src/ssl.rs
@@ -1,11 +1,14 @@
 use crate::vm::{PyObjectRef, VirtualMachine};
 
 pub(crate) fn make_module(vm: &VirtualMachine) -> PyObjectRef {
+    #[cfg(target_os = "android")]
+    std::env::set_var("SSL_CERT_DIR", "/system/etc/security/cacerts");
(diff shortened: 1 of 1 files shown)


coolreader18:
Ah, yeah, #3163. `/system/etc/security/cacerts` should be added to openssl-probe then. @afpro do you think that would also be a solution for your use case?

youknowone (approved):
how are they magically not required anynore? they do it by watching to envvar?

afpro:
> Ah, yeah, #3163. `/system/etc/security/cacerts` should be added to openssl-probe then. @afpro do you think that would also be a solution for your use case?

i'm not familiar with 'SSL_CERT_DIR' envirtonment variable. this would be more maintainability if set 'SSL_CERT_DIR' could work as expectedly

afpro:
just tried 'SSL_CERT_DIR', not working.

PR #3479, opened 2021-11-28
Fix __isabstractmethod__ for class/static methods and properties

chrismoradi:
Includes a minor departure from CPython implementation of `inspect.isabstract` because TPFLAGS_IS_ABSTRACT is not set in `obj.__flags__`. Not sure if this is a problem but happy to revert this change if it is.

### DIFF
--- a/Lib/inspect.py
+++ b/Lib/inspect.py
@@ -300,16 +300,16 @@ def isroutine(object):
 
 def isabstract(object):
     """Return true if the object is an abstract base class (ABC)."""
+    # TODO: RUSTPYTHON
+    # TPFLAGS_IS_ABSTRACT is not being set for abstract classes, so this implementation differs from CPython.
     if not isinstance(object, type):
         return False
-    if object.__flags__ & TPFLAGS_IS_ABSTRACT:
-        return True
     if not issubclass(type(object), abc.ABCMeta):
         return False
     if hasattr(object, '__abstractmethods__'):
         # It looks like ABCMeta.__new__ has finished running;
         # TPFLAGS_IS_ABSTRACT should have been accurate.
-        return False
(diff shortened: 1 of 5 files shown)


Lib/inspect.py
```diff
@@ -300,16 +300,16 @@ def isroutine(object):
 
 def isabstract(object):
     """Return true if the object is an abstract base class (ABC)."""
+    # TODO: RUSTPYTHON
```
youknowone, line 303: How about leaving the original code as commented-out and put this comments on line 305? So future contributor can notice what's changed in the source code.

vm/src/builtins/property.rs
```diff
@@ -217,6 +217,32 @@ impl PyProperty {
         }
         .into_ref_with_type(vm, TypeProtocol::clone_class(&zelf))
     }
+
+    #[pyproperty(magic)]
+    fn isabstractmethod(&self, vm: &VirtualMachine) -> PyObjectRef {
+        let getter_abstract = match self.getter.read().to_owned() {
```
youknowone, line 223: @coolreader18 is `to_owned` the best way to handle this?

Lib/inspect.py
chrismoradi: Sure! Fixed this.

youknowone (approved):

vm/src/builtins/property.rs
youknowone: @coolreader18 do you have any idea about this? or i will merge this PR

chrismoradi: @youknowone & @coolreader18, Just wanted to check in on this PR. Should I work on trying to remove the `to_owned()`?

youknowone:
I am sorry to keep this unmerged for long time!

PR #3480, opened 2021-11-29
Test: move import_helper out

deantvv:
This PR only cover the change for Lib/test/support/import_helper.py and reflect the changes necessary for other tests to use.
Mostly of the part is `from test.support import xxx` --> `from test.support.import_helper import xxx` etc.

`import_fresh_module` and `CleanImport` still use old implementation because it need `_imp._override_frozen_modules_for_tests` and it is currently not there.


youknowone (approved):

youknowone:
great, thank you!

PR #3481, opened 2021-11-29
Register `True`, `False`, `None` to builtins

Snowapril:
According to `whats_left.sh` printing, that `True`, `False`, `None` constants must be registered at builtins


coolreader18 (approved):
Nice catch!

PR #3484, opened 2021-11-30
Add missing `__getattribute__` on several types

Snowapril:
This revision implements follow things
*  Implement default code for `getattro` in `GetAttro` trait with `PyObject_GenericGetAttr`
*  Fix some incompatibility in `repr` of `PyBoundMethod`
*  Add missing `__getattribute__` to below types
    * BaseException
    * bytearray & bytearray_iterator
    * bytes & bytes_iterator
    * complex
    * dict, dict_keyiterator, dict_keys, dict_reversekeyiterator, dict_values, dict_valueiterator, dict_reversevalueiterator, dict_items, dict_itemsiterator, dict_reverseitemiterator
    * enumerate
    * filter
    * float
    * int
    * list, list_iterator, list_reverseiterator
    * map
    * memoryview
    * property
    * str & str_iterator
    * range, range_iterator, longrange_iterator
    * set, set_iteartor, frozenset
    * slice
    * tuple & tuple_iterator
    * zip

##### Before
```python
>>>>> str.__getattribute__
<method '__getattribute__' of 'object' objects>
>>>>> str("assdf").__getattribute__
<bound method object.__getattribute__ of 'assdf'>
```
##### After
```python
>>>>> str("assdf").__getattribute__
<method-wrapper str.__getattribute__ of str object at 0x55f52b946570>
>>>>> str.__getattribute__
<method '__getattribute__' of 'str' objects>
```


youknowone:
if all of them just share the same implementation, what's the benefit to have them for each type?

Snowapril:
These are incompatibility stuff according to `./whats_left.sh`. 
I think there are no more benefits except that `__getattribute__` of several types are different among them although their work is fallback to the same one.

##### RustPython (Before this PR)
```python
>>>>> str.__getattribute__ == object.__getattribute__
True
```
##### CPython 3.10
```python
>>> str.__getattribute__ == object.__getattribute__
False
```

youknowone:
I want to check what kind of benefit CPython is gaining from this implementation

youknowone:
I checked CPython source code and found all of the getattro slots are filled with the same function `PyObject_GenericGetAttr` , which is the same implementation as `PyBaseObject::getattr` but not `GetAttr::getattr` from our code base. If the `repr` result is different, I think it means only the wrapper object is not shared.

youknowone:
For additional information, the way using same wrapper function but pointing same internal native function is same as type_new wrapper

youknowone:
Now we have fully revised slot implementation, so this is going to be closed. Thank you so much for making many improvements for those slot systems! You made the starts.

PR #3488, opened 2021-12-01
import_helper: refactor `frozen_modules`

deantvv:
`frozen_modules` requires non-trivial changes in import (`imp.rs`)
and from the doc it is deprecated since 3.3 which makes this feacture
unlikely to be support on this project.

So instead of implementing this feature, I make `frozen_modules` to
raise exception if `frozen` is required. Other than that, all implementation
is just like the upstream version (except one TODO comment to remind
us there is something not fully supported)

### DIFF
--- a/Lib/test/support/import_helper.py
+++ b/Lib/test/support/import_helper.py
@@ -91,204 +91,105 @@ def _save_and_remove_modules(names):
     return orig_modules
 
 
-# XXX RUSTPYTHON: need _imp._override_frozen_modules_for_tests
-# @contextlib.contextmanager
-# def frozen_modules(enabled=True):
-#     """Force frozen modules to be used (or not).
+# TODO RUSTPYTHON: need _imp._override_frozen_modules_for_tests
+# The following implementation is NOT correct and only raise
+# exception when it needs enabled=True
+@contextlib.contextmanager
+def frozen_modules(enabled=True):
+    """Force frozen modules to be used (or not).
 
-#     This only applies to modules that haven't been imported yet.
-#     Also, some essential modules will always be imported frozen.
-#     """
-#     _imp._override_frozen_modules_for_tests(1 if enabled else -1)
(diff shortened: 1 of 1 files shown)


coolreader18:
`_imp` is actually only deprecated for user code, it's still heavily used in cpython internals. I think it'd probably be good to implement it in `_imp` (at least at some point)

deantvv:
@coolreader18 
The part that is needed in `frozen_module` is `_imp._override_frozen_modules_for_tests`. 

From grepping this function (`_override_frozen_modules_for_test`) in cpython repo, it is only used in `frozen_modules` in `test/support/import_helper.py`.

Instead of implementing `_override_frozen_modules_for_test` right now, I chosed to raise Exception if we actually use this function. And from test result, we didn't use this function at all. Following are the implementation from cpython and this PR.

[cpython version](https://github.com/python/cpython/blob/main/Lib/test/support/import_helper.py#L93-L104)
```python
@contextlib.contextmanager
def frozen_modules(enabled=True):
    _imp._override_frozen_modules_for_tests(1 if enabled else -1)
    try:
        yield
    finally:
        _imp._override_frozen_modules_for_tests(0)
```

This PR
```python
@contextlib.contextmanager
def frozen_modules(enabled=True):
    if enabled:
        raise NotImplemented("frozen_modules is not implemented on RustPython")

    yield
```

Compared with previous commit, this PR also doesn't change the behavior since the old commit just ignore all function called of `frozen_modules`. Now we add `frozen_modules` and check if we need `_imp.override_frozen_modules_for_tests`. If yes, we raise exception. If not (seems like all tests we're running now doesn't need this), it exhibit the same behavior as cpython.

youknowone:
it sounds reasonable to me

coolreader18:
Oh yes, me as well, sorry I just didn't have the time to fully review the added code when I first noticed this. LGTM

PR #3492, opened 2021-12-02
Add test_bdb from CPython 3.9

fanninpm:
Requesting @deantvv as a reviewer to help out with the `test.support` stuff.


deantvv:
@fanninpm There is still some weird error. For the `test.support` part, it looks good to me.

Issue #3493, opened 2021-12-03
MemoryError


youknowone:
## Feature

Rust 1.57 added Fallible allocation.
It will be easy to implement MemoryError than before

coolreader18:
Well I'm working on this now I suppose :grin: I was gonna just test some stuff out for a bit but now I've been doing it for an hour and I'm on the way to an implementation lol

coolreader18:
This could probably serve as a tracking issue of things to switch over to fallible allocations, with checkboxes and stuff (at some point), since this should be a fairly easy thing for people to work on as a first contribution (hopefully).

PR #3494, opened 2021-12-04
First impl of fallible allocations

coolreader18:
This PR doesn't change a _ton_ of things to use fallible allocations, but it does set up a framework for switching over to fallible allocations inside RustPython.

Related to #3493

### DIFF
--- a/common/src/safe_alloc.rs
+++ b/common/src/safe_alloc.rs
@@ -1,0 +1,194 @@
+use std::ptr;
+
+pub use std::collections::TryReserveError;
+pub type TryReserveResult<T> = Result<T, TryReserveError>;
+
+pub trait VecExt: Sized {
+    type T;
+    fn try_with_capacity(n: usize) -> TryReserveResult<Self>;
+    fn try_push(&mut self, x: Self::T) -> TryReserveResult<()>;
+    fn try_extend_from_slice_copy(&mut self, other: &[Self::T]) -> TryReserveResult<()>
+    where
+        Self::T: Copy;
+    fn try_extend_from_slice_clone(&mut self, other: &[Self::T]) -> TryReserveResult<()>
+    where
+        Self::T: Clone;
+    fn try_append(&mut self, other: &mut Self) -> TryReserveResult<()>;
+}
+
+impl<T> VecExt for Vec<T> {
+    type T = T;
+    fn try_with_capacity(n: usize) -> TryReserveResult<Self> {
+        let mut v = Vec::new();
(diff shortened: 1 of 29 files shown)


common/src/safe_alloc.rs
```diff
@@ -0,0 +1,194 @@
...
+            local_len: vec.len(),
+            vec,
+        }
+    }
+
+    #[inline]
+    fn increment_len(&mut self, increment: usize) {
+        self.local_len += increment;
+    }
+}
```
youknowone, line 92: This idea looks good. I think it can have similar interface like `MaybeUninit` because they have similar concept.

```rust
struct VecBuilder {
    vec: Vec<T>,
    len: usize,
}

impl VecUninit {
    fn try_with_len(len: usize) -> TryReserveResult {
        Self::try_with_capacity(len, len)
    }
    /// SAFETY: len <= capacity
    unsafe fn try_with_capacity(capacity: usize, len: usize) -> TryReserveResult {
        debug_assert!(len <= capacity);
        Self {
            vec: Vec::try_with_capacity(...)
            len,
        }
    }
    unsafe fn assume_init(self) -> Vec {
        self.vec.set_len(self.len);
        self.vec
    }
}

```

vm/src/function.rs
```diff
@@ -25,14 +25,61 @@ pub trait IntoPyObject {
...
+    fn map_pyerr(self, vm: &VirtualMachine) -> PyResult<Self::Ok>;
+}
+
+impl<T, E: IntoPyException> PyErrResultExt for Result<T, E> {
+    type Ok = T;
+    #[inline]
+    fn map_pyerr(self, vm: &VirtualMachine) -> PyResult<T> {
+        self.map_err(|e| e.into_pyexception(vm))
+    }
+}
```
youknowone, line 81: 👍

youknowone (approved):

common/src/safe_alloc.rs
coolreader18: Oh.. this is just copied from the standard library. I should add a comment.

vm/src/function.rs
coolreader18: Glad you like it! I've been thinking about something like this for a while, to get rid of ".map_err(|e| e.into_pyexception(vm))` __everywhere__

youknowone: I think this is different approach but solving a part of problems I tried to handle in  #3326

DimitrisJim:
bumping here, any chance we could rebase this and get it in?

PR #3499, opened 2021-12-11
Expose `types.GenericAlias.__getitem__`

moreal:
## Description

While making #3497 work, I found `types.GenericAlias.__getitem__` wasn't exposed. This pull request just makes it 

## Check

You can run the below lines to test.

```
import typing
type(list[int]).__getitem__(list[typing.T], ...)  # expect list[...]
```

Before this pull request, you can see:

```
>>>>> import typing
>>>>> type(list[int]).__getitem__(list[typing.T], ...)
AttributeError: type object 'GenericAlias' has no attribute '__getitem__'. Did you mean: '__setattr__'?
```

### DIFF
--- a/vm/src/builtins/genericalias.rs
+++ b/vm/src/builtins/genericalias.rs
@@ -127,6 +127,58 @@ impl PyGenericAlias {
     }
 
     #[pymethod(magic)]
+    fn getitem(&self, needle: PyObjectRef, vm: &VirtualMachine) -> PyResult {
+        let num_params = self.parameters.len();
+        if num_params == 0 {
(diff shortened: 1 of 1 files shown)


youknowone (approved):
@qingshi163 @Snowapril can this be wriiten with mapping protocol?

Snowapril (approved):
@youknowone  ooh yeah, It can. Actually it must be slot-wrapper (wrapper_descriptor) not a bounded method. At current step, as all the other types that implementing mapping protocol have them as bounded method, seems reasonable 😊

youknowone (approved):

PR #3503, opened 2021-12-14
os: implement device_encoding

deantvv:


vm/src/stdlib/os.rs
```diff
@@ -1697,6 +1697,42 @@ pub(super) mod _os {
...
+    fn device_encoding(fd: i32, _vm: &VirtualMachine) -> PyResult<Option<String>> {
+        if !isatty(fd) {
+            return Ok(None);
+        }
+
+        cfg_if::cfg_if! {
+            if #[cfg(target_os = "android")] {
+                Ok(Some("UTF-8".to_owned()))
+            } else if #[cfg(windows)] {
+                let cp = if fd == 0 {
```
youknowone, line 1712: Here is warnings. I think `match` fits here better becaues all of the condition expression is about `fd`

deantvv: Changed as requested in new commit.

youknowone (approved):
great, thank  you!

PR #3507, opened 2021-12-20
Implement Number Protocol

qingshi163:
number protocol of #3244

### DIFF
--- a/vm/src/builtins/float.rs
+++ b/vm/src/builtins/float.rs
@@ -1,5 +1,3 @@
-use std::borrow::Cow;
-
 use super::{
     try_bigint_to_f64, PyByteArray, PyBytes, PyInt, PyIntRef, PyStr, PyStrRef, PyType, PyTypeRef,
 };
@@ -153,9 +151,7 @@ impl Constructor for PyFloat {
         let float_val = match arg {
             OptionalArg::Missing => 0.0,
             OptionalArg::Present(val) => {
-                if cls.is(vm.ctx.types.float_type) && val.class().is(PyFloat::class(vm)) {
-                    unsafe { val.downcast_unchecked::<PyFloat>().value }
-                } else if let Some(f) = val.try_float_opt(vm)? {
+                if let Some(f) = val.try_float_opt(vm)? {
                     f.value
                 } else {
                     float_from_string(val, vm)?
@@ -539,13 +535,40 @@ impl Hashable for PyFloat {
 }
 
 impl AsNumber for PyFloat {
-    fn as_number(_zelf: &crate::Py<Self>, _vm: &VirtualMachine) -> Cow<'static, PyNumberMethods> {
-        Cow::Borrowed(&Self::NUMBER_METHODS)
-    }
+    const AS_NUMBER: PyNumberMethods = PyNumberMethods {
+        add: Some(|number, other, vm| Self::number_float_op(number, other, |a, b| a + b, vm)),
+        subtract: Some(|number, other, vm| Self::number_float_op(number, other, |a, b| a - b, vm)),
+        multiply: Some(|number, other, vm| Self::number_float_op(number, other, |a, b| a * b, vm)),
+        remainder: Some(|number, other, vm| Self::number_general_op(number, other, inner_mod, vm)),
+        divmod: Some(|number, other, vm| Self::number_general_op(number, other, inner_divmod, vm)),
+        power: Some(|number, other, vm| Self::number_general_op(number, other, float_pow, vm)),
+        negative: Some(|number, vm| {
+            let value = Self::number_downcast(number).value;
+            (-value).to_pyresult(vm)
+        }),
+        positive: Some(|number, vm| Self::number_float(number, vm).to_pyresult(vm)),
+        absolute: Some(|number, vm| {
+            let value = Self::number_downcast(number).value;
+            value.abs().to_pyresult(vm)
+        }),
+        boolean: Some(|number, _vm| Ok(Self::number_downcast(number).value.is_zero())),
+        int: Some(|number, vm| {
+            let value = Self::number_downcast(number).value;
+            try_to_bigint(value, vm).map(|x| vm.ctx.new_int(x))
+        }),
+        float: Some(|number, vm| Ok(Self::number_float(number, vm))),
+        floor_divide: Some(|number, other, vm| {
+            Self::number_general_op(number, other, inner_floordiv, vm)
+        }),
+        true_divide: Some(|number, other, vm| {
+            Self::number_general_op(number, other, inner_div, vm)
+        }),
+        ..PyNumberMethods::NOT_IMPLEMENTED
+    };
 }
 
 impl PyFloat {
-    fn np_general_op<F, R>(
+    fn number_general_op<F, R>(
         number: &PyNumber,
         other: &PyObject,
         op: F,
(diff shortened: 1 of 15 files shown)


youknowone:
@qingshi163 what have to be done to finish this PR?

youknowone:
awesome!

youknowone (commented):
I have questions about cow and OnceCell part, but it looks working!
This is the way to go and we can fix other stuff later.

Do you have more planning to do on this PR?

vm/src/types/slot.rs
```diff
@@ -900,6 +932,40 @@ pub trait AsSequence: PyPayload {
     }
 }
 
+#[pyimpl]
+pub trait AsNumber: PyPayload {
+    #[inline]
+    #[pyslot]
+    fn slot_as_number(zelf: &PyObject, vm: &VirtualMachine) -> Cow<'static, PyNumberMethods> {
```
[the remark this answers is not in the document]
qingshi163: I think for heap object we need to use Cow::Owned witch implemented in as_number_wrapper

qingshi163:
There is quite a lot of things not done yet, i think all the mathematic operations should be implemented via number protocol

vm/src/types/slot.rs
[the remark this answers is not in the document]
youknowone: They are not decided by object, but decided by type. Then I think the type can hold a cloned `PyNumberMethods` with edited values. instead of `Cow` for every object when `as_number` is called.

youknowone:
How about finishing this PR as current state as core implementation of AsNumber trait, and working on other types in following other PRs?
Keeping a huge PR will be a burden to maintain (to rebase everything after other merge)

vm/src/types/slot.rs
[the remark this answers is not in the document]
qingshi163: Yes, it should be stored in the type struct, but we can fix it later

youknowone:
@qingshi163 are you going to make this ready for review? or do you want to do more work on it?

qingshi163:
I am still busy with it

vm/src/builtins/float.rs
```diff
@@ -170,17 +153,10 @@ impl Constructor for PyFloat {
...
-                        Err(val) => val,
-                    }
-                } else {
-                    val
-                };
-
-                if let Some(f) = val.try_to_f64(vm)? {
-                    f
+                if cls.is(vm.ctx.types.float_type) && val.class().is(PyFloat::class(vm)) {
+                    unsafe { val.downcast_unchecked::<PyFloat>().value }
```
youknowone, line 155: here is a regression. val object will not be reused even for same type.

vm/src/builtins/float.rs
```diff
@@ -562,6 +536,71 @@ impl Hashable for PyFloat {
...
+        floor_divide: Some(|number, other, vm| {
+            Self::np_general_op(number, other, inner_floordiv, vm)
+        }),
+        true_divide: Some(|number, other, vm| Self::np_general_op(number, other, inner_div, vm)),
+        ..PyNumberMethods::NOT_IMPLEMENTED
+    };
+}
+
+impl PyFloat {
+    fn np_general_op<F, R>(
```
youknowone, line 571: we are trying not to use c-style prefix like `np_`

vm/src/builtins/int.rs
```diff
@@ -751,6 +755,78 @@ impl Hashable for PyInt {
...
+        add: Some(|number, other, vm| Self::np_int_op(number, other, |a, b| a + b, vm)),
+        subtract: Some(|number, other, vm| Self::np_int_op(number, other, |a, b| a - b, vm)),
+        multiply: Some(|number, other, vm| Self::np_int_op(number, other, |a, b| a * b, vm)),
+        remainder: Some(|number, other, vm| Self::np_general_op(number, other, inner_mod, vm)),
+        divmod: Some(|number, other, vm| Self::np_general_op(number, other, inner_divmod, vm)),
+        power: Some(|number, other, vm| Self::np_general_op(number, other, inner_pow, vm)),
+        negative: Some(|number, vm| {
+            Self::number_downcast(number)
+                .value
+                .clone()
```
youknowone, line 769: https://docs.rs/num/latest/num/struct.BigInt.html#impl-Neg-1

Neg is also implemented for `&BigInt`. This clone is redundant.

vm/src/builtins/int.rs
```diff
@@ -751,6 +755,78 @@ impl Hashable for PyInt {
...
+                .neg()
+                .to_pyresult(vm)
+        }),
+        positive: Some(|number, vm| Ok(Self::np_int(number, vm).into())),
+        absolute: Some(|number, vm| Self::number_downcast(number).value.abs().to_pyresult(vm)),
+        boolean: Some(|number, _vm| Ok(Self::number_downcast(number).value.is_zero())),
+        invert: Some(|number, vm| {
+            let value = Self::number_downcast(number).value.clone();
```
youknowone, line 777: `Not` also implemented for reference.

https://docs.rs/num/latest/num/struct.BigInt.html#impl-Not-1

vm/src/protocol/number.rs
```diff
@@ -0,0 +1,291 @@
...
+    stdlib::warnings,
+    AsObject, PyObject, PyPayload, PyRef, PyResult, TryFromBorrowedObject, VirtualMachine,
+};
+
+#[allow(clippy::type_complexity)]
+#[derive(Clone)]
+pub struct PyNumberMethods {
+    /* Number implementations must check *both*
+    arguments for proper type and implement the necessary conversions
+    in the slot functions themselves. */
```
youknowone, line 16: Does this mean we need to check the first argument(&PyNumber) is a correct PyNumber object?
In this case, I feel like the type will fit better in `&PyObject` rather than `&PyNumber`

vm/src/stdlib/warnings.rs
```diff
@@ -1,5 +1,25 @@
...
+pub fn warn(
+    category: &Py<PyType>,
+    message: String,
+    stack_level: usize,
+    vm: &VirtualMachine,
+) -> PyResult<()> {
+    // let module = vm.import("warnings", None, 0)?;
+    // let func = module.get_attr("warn", vm)?;
+    // vm.invoke(&func, (message, category, stack_level))?;
+    // TODO
```
youknowone, line 14: unuseful comments

vm/src/protocol/number.rs
```diff
@@ -0,0 +1,291 @@
...
+                None => Err(vm.new_value_error(format!(
+                    "invalid literal for int() with base {}: {}",
+                    base,
+                    obj.repr(vm)?,
+                ))),
+            }
+        }
+
+        if self.obj.class().is(PyInt::class(vm)) {
+            Ok(unsafe { self.obj.to_owned().downcast_unchecked::<PyInt>() })
```
youknowone, line 158: this part can be replaced with `payload_if_exact` to delegate unsafe operation to single spot.

vm/src/protocol/number.rs
```diff
@@ -0,0 +1,291 @@
...
+                    format!(
+                        "__int__ returned non-int (type {}).  \
+                The ability to return an instance of a strict subclass of int \
+                is deprecated, and may be removed in a future version of Python.",
+                        ret.class()
+                    ),
+                    1,
+                    vm,
+                )?;
+                Ok(vm.ctx.new_int(ret.as_bigint().clone()))
```
youknowone, line 173:
```suggestion
                Ok(vm.ctx.new_bigint(ret.as_bigint()))
```

vm/src/protocol/number.rs
```diff
@@ -0,0 +1,291 @@
...
+                self.obj.class()
+            )))
+        }
+    }
+
+    pub fn index_opt(&self, vm: &VirtualMachine) -> PyResult<Option<PyIntRef>> {
+        if self.obj.class().is(PyInt::class(vm)) {
+            Ok(Some(unsafe {
+                self.obj.to_owned().downcast_unchecked::<PyInt>()
+            }))
```
youknowone, line 217: payload_if_exact

vm/src/protocol/number.rs
```diff
@@ -0,0 +1,291 @@
...
+                    format!(
+                        "__index__ returned non-int (type {}).  \
+                The ability to return an instance of a strict subclass of int \
+                is deprecated, and may be removed in a future version of Python.",
+                        ret.class()
+                    ),
+                    1,
+                    vm,
+                )?;
+                Ok(Some(vm.ctx.new_int(ret.as_bigint().clone())))
```
youknowone, line 234:
```suggestion
                Ok(Some(vm.ctx.new_bigint(ret.as_bigint())))
```

vm/src/protocol/number.rs
```diff
@@ -0,0 +1,291 @@
...
+                self.obj.class()
+            ))
+        })
+    }
+
+    pub fn float_opt(&self, vm: &VirtualMachine) -> PyResult<Option<PyRef<PyFloat>>> {
+        if self.obj.class().is(PyFloat::class(vm)) {
+            Ok(Some(unsafe {
+                self.obj.to_owned().downcast_unchecked::<PyFloat>()
+            }))
```
youknowone, line 256: payload_if_exact

youknowone (changes requested):
I think #3750 changes also be able to be adapt here.
They can be done later, but I attached a commit simplifying `AsNumber` interfaces.

youknowone:
```
+-----------+----------------------------------+-----------------------------------+----------------------------------------+
| Benchmark | benches/out/cpython38/nbody.json | benches/out/rustpython/nbody.json | benches/out/rustpython-3507/nbody.json |
+===========+==================================+===================================+========================================+
| nbody     | 103 ms                           | 1.79 sec: 17.39x slower           | 1.82 sec: 17.70x slower                |
+-----------+----------------------------------+-----------------------------------+----------------------------------------+
```

vm/src/protocol/number.rs
qingshi163: The comment is from cpython but as the source shows that cpython only does the check for binary operations which could be called invertly but assume the type is correct for other functions

youknowone:
at least, no regression after using references
```
+-----------+----------------------------------+-----------------------------------+----------------------------------------+--------------------------------------------+
| Benchmark | benches/out/cpython38/nbody.json | benches/out/rustpython/nbody.json | benches/out/rustpython-3507/nbody.json | benches/out/rustpython-3507/nbody-ref.json |
+===========+==================================+===================================+========================================+============================================+
| nbody     | 103 ms                           | 1.79 sec: 17.39x slower           | 1.82 sec: 17.70x slower                | 1.79 sec: 17.38x slower                    |
+-----------+----------------------------------+-----------------------------------+----------------------------------------+--------------------------------------------+
```

youknowone (approved):

PR #3508, opened 2021-12-21
add #[non_exhaustive]

youknowone:


coolreader18:
I think it's not really necessary for private types, since it's only outside of the crate that it's actually non-exhaustive. But otherwise, looks good!

coolreader18 (approved):

PR #3510, opened 2021-12-30
fix doc-test failure

youknowone:
did `#` handling changed recently?


youknowone:
this is not the reason

PR #3512, opened 2022-01-04
Upgrade clap to v3 (and other dep upgrades)

coolreader18:


youknowone:
@coolreader18 will you rebase this PR?

key262yek:
When we upgrade clap from v2.34 to v3, we cannot parse anymore multiple argument when the first argument is attached to short flag. 
It make test_quopri.py fail. 

I asked it to clap devlopers. 
[Multiple value doesn't work when the first argument attached to short flag #3903](https://github.com/clap-rs/clap/issues/3903)

PR #3516, opened 2022-01-09
Fix doctests in ci

coolreader18:
- Fix doc comment for Interpreter
- Fix doctests in ci

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -71,7 +71,8 @@ opt-level = 3
 
 [profile.test]
 opt-level = 3
-lto = "thin"
+# https://github.com/rust-lang/rust/issues/92869
+# lto = "thin"
 
 [profile.bench]
 lto = true
--- a/common/src/boxvec.rs
+++ b/common/src/boxvec.rs
@@ -465,7 +465,7 @@ pub struct Drain<'a, T> {
 }
 
 unsafe impl<'a, T: Sync> Sync for Drain<'a, T> {}
-unsafe impl<'a, T: Send> Send for Drain<'a, T> {}
+unsafe impl<'a, T: Sync> Send for Drain<'a, T> {}
 
 impl<T> Iterator for Drain<'_, T> {
     type Item = T;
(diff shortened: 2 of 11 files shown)


Cargo.toml
```diff
@@ -73,6 +73,11 @@ opt-level = 3
 opt-level = 3
 lto = "thin"
 
+# ugly hack since doctests link against debug version of rustpython-vm even
+# when profile.test is optimized
```
youknowone, line 77: Though we usually use optimization and debug option exclusively, optimization is about optimization itself and debug is about debug information. So I think giving optimization option to test doesn't change it is a debug or not.

coolreader18: Right, and it does work fine unoptimized on our machines apparently, but I think there's some timeout in ci that kills the test process if it takes too long? And afaict it was linking against release in Rust 1.56, the update broke it a bit, this just brings it back to previous behavior

coolreader18: ach wait the tests aren't even fixed, dang it

youknowone: yep, I think it doesn't fix anything unless it is related to cache

coolreader18: ugh I don't know why I didn't consider that it was a cache issue

fanninpm: The cache action is reporting a cache *miss*, so I doubt that it's actually a cache issue.

fanninpm: Could https://github.com/rust-lang/rust/issues/92869 be related?

coolreader18: Ahh, probably

coolreader18:
WOOOooooooo....! Passing!

coolreader18:
Merging this so main turns green.

Cargo.toml
```diff
@@ -71,7 +71,8 @@ opt-level = 3
 
 [profile.test]
 opt-level = 3
-lto = "thin"
+# https://github.com/rust-lang/rust/issues/92869
```
youknowone, line 74: this is totally unexpected effect

Issue #3519, opened 2022-01-15
Error on wasm build: `there is no argument named meta_indent`


FoxNerdSaysMoo:
```
   Compiling rustpython-derive v0.1.2 (/home//Desktop/Pydom/RustPython/derive)
error: there is no argument named `meta_ident`
   --> derive/src/util.rs:105:32
    |
105 | ...                   "#[{meta_ident}({name})] is not one of allowed attributes [{}]",
    |                          ^^^^^^^^^^^^

error: there is no argument named `name`
   --> derive/src/util.rs:105:45
    |
105 | ...                   "#[{meta_ident}({name})] is not one of allowed attributes [{}]",
    |                                       ^^^^^^

error: there is no argument named `meta_ident`
   --> derive/src/util.rs:117:28
    |
117 |                 format!("#[{meta_ident}(..)] cannot contain literal"),
    |                            ^^^^^^^^^^^^

error: could not compile `rustpython-derive` due to 3 previous errors
warning: build failed, waiting for other jobs to finish...
error: build failed

```
Occurred when running
```
cargo build --release --target wasm32-wasi --features="freeze-stdlib"
```
(I already ran `rustup target add wasm32-wasi`)

Platform information:
```
$ uname -a
Linux computer 5.15.11-arch2-1 #1 SMP PREEMPT Wed, 22 Dec 2021 09:23:54 +0000 x86_64 GNU/Linux
```
```
$ rustc --version
rustc 1.57.0 (f1edd0429 2021-11-29)
```
This is using the latest version of RustPython.

tyoc213:
Same here

```
➜  RustPython git:(main) uname -a
Linux u 5.11.0-46-generic #51~20.04.1-Ubuntu SMP Fri Jan 7 06:51:40 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
➜  RustPython git:(main) rustc --version
rustc 1.57.0 (f1edd0429 2021-11-29)
```

besok:
Actually, The named parameters work properly on the latest version of rust. Try to update compiler:
```
❯ rustup update stable
...
 stable-x86_64-pc-windows-msvc updated - rustc 1.58.0 (02072b482 2022-01-11) (from rustc 1.57.0 (f1edd0429 2021-11-29))
```

tyoc213:
Well on main page it says this 

```
RustPython requires Rust latest stable version (e.g 1.43.0 at May 24th 2020)
```

[...]

PR #3520, opened 2022-01-15
Adding named parameters to strings

tyoc213:
This should let compile #3519 3519

### DIFF
--- a/derive/src/util.rs
+++ b/derive/src/util.rs
@@ -103,7 +103,7 @@ impl ItemMetaInner {
                         ident,
                         format!(
                             "#[{meta_ident}({name})] is not one of allowed attributes [{}]",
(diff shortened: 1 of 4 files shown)


fanninpm:
@coolreader18 something something "bleeding edge"… something something "backwards compatibility"…

tyoc213:
Im using `rustc 1.57.0 (f1edd0429 2021-11-29)` also on main page

```
RustPython requires Rust latest stable version (e.g 1.43.0 at May 24th 2020)
```

youknowone:
Thank you for contributing. But we concentrate on supporting only the latest stable Rust compilers for now. As you commented, it will run on 1.58 the most recent version which fit for our expectation.

PR #3523, opened 2022-01-22
Add staticmethod in macros and apply it to maketrans

hackerwins:
**What this PR does / why we need it:**

This PR added `staticmethod` in macros. And apply it to `maketrans` and changed `staticmethod` to be callable.

- Change `staticmethod` to be callable as in 3.10.x 
- Apply `staticmethod` to maketrans 
- Remove unexpected successes in `test_shlex`

See
- https://docs.python.org/3.10/library/stdtypes.html?highlight=maketrans#str.maketrans
- https://docs.python.org/3/whatsnew/3.10.html#other-language-changes
  
AS-IS
```python
$ target/debug/rustpython
>>>>> def abc():
.....   print('hello')
..... 
>>>>> staticmethod(abc)()
TypeError: 'staticmethod' object is not callable
```

TO-BE

```python
$ target/debug/rustpython
>>>>> def abc():
.....   print('hello')
..... 
>>>>> staticmethod(abc)()
hello
>>>>> staticmethod(abc)
<staticmethod object at 0x7ff7e7c55490>
>>>>> staticmethod(abc)()
hello
```

### DIFF
--- a/Lib/test/test_decorators.py
+++ b/Lib/test/test_decorators.py
@@ -76,6 +76,7 @@ class TestDecorators(unittest.TestCase):
         self.assertEqual(C.foo(), 42)
         self.assertEqual(C().foo(), 42)
 
+    @unittest.skip("TODO: 3.10.x changed staticmethod to be callable")
     def test_staticmethod_function(self):
         @staticmethod
         def notamethod(x):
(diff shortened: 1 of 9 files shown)


youknowone (approved):
thank you for contributing!

PR #3525, opened 2022-01-23
fcntl module update proposal apple and freebsd specific flags.

devnexen:


stdlib/src/fcntl.rs
```diff
@@ -21,6 +21,12 @@ mod fcntl {
     #[pyattr]
     use libc::{F_GETOWN, F_RDLCK, F_SETOWN, F_UNLCK, F_WRLCK};
 
+    #[cfg(target_vendor = "apple")]
+    use libc::F_NOCACHE;
+
+    #[cfg(target_os = "freebsd")]
+    use libc::{F_DUP2FD, F_DUP2FD_CLOEXEC};
+
```
fanninpm, line 29: Did you check and make sure these were actually _accessible_ when you run the RustPython executable?

devnexen: with last libc release they are.

fanninpm: That wasn't my point. Can you get to the things you added _while you're running the RustPython executable?_

devnexen: ah right sorry

fanninpm:
The CI failure is unrelated to your changes, so it's bogus.

PR #3527, opened 2022-01-28
Allow heap types' `__name__` setter

moreal:
In CPython implementation, immutable types (usually bulletins types) doesn't allow `type.__name__` setter but heap types allow `<TYPE>.__name__` setter. For example, there is `typing.io`. You can see also the 2733202 commit.

So this pull request does:

- Allow heap types' `__name__` setter.
- Don't split the name if it is heap type.

### CPython 3.10.0

```python
>>> type.__name__ = "foo"
TypeError: cannot set '__name__' attribute of immutable type 'type'
>>> class io: ...
... 
>>> io.__name__ = "typing.io"
>>> io.__name__
'typing.io'
```

### RustPython

```python
>>>>> type.__name__ = "foo"
TypeError: cannot set '__name__' attribute of immutable type 'type'
>>>>> class io: ...
>>>>> io.__name__ = "typing.io"
>>>>> io.__name__
'typing.io'
```

### DIFF
--- a/vm/src/builtins/pytype.rs
+++ b/vm/src/builtins/pytype.rs
@@ -545,6 +545,29 @@ impl PyType {
         ))
     }
 
+    #[pyproperty(magic, setter)]
+    fn set_name(&self, value: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        if !self.slots.flags.has_feature(PyTypeFlags::HEAPTYPE) {
+            Err(vm.new_type_error(format!(
+                "cannot set '{}' attribute of immutable type '{}'",
+                "__name__",
+                self.name()
+            )))
+        } else {
+            match value.downcast_ref::<PyStr>().map(|m| m.as_str()) {
+                Some(name) => {
+                    *self.slots.name.write() = Some(name.to_string());
+                    Ok(())
+                }
+                None => Err(vm.new_type_error(format!(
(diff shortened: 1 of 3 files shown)


vm/src/builtins/pytype.rs
```diff
@@ -545,6 +550,29 @@ impl PyType {
...
+                    *self.slots.name.write() = Some(name.to_string());
+                    Ok(())
+                }
+                None => Err(vm.new_type_error(format!(
+                    "can only assign string to {}.__name__, not '{}'",
+                    self.name(),
+                    value.class().name()
+                ))),
+            }
+        }
```
youknowone, line 573:
```suggestion
        if !self.slots.flags.has_feature(PyTypeFlags::HEAPTYPE) {
            return Err(vm.new_type_error(format!(
                "cannot set '{}' attribute of immutable type '{}'",
                "__name__",
                self.name()
            )));
        }
        let name = value
            .downcast_ref::<PyStr>()
            .map(|m| m.as_str())
            .ok_or_else(|| {
                vm.new_type_error(format!(
                    "can only assign string to {}.__name__, not '{}'",
                    self.name(),
                    value.class().name()
                ))
            })?;
        *self.slots.name.write() = Some(name.to_string());
        Ok(())
```

youknowone (approved):

Issue #2059, opened 2020-08-02
mmap is not implemented


tobark:
## Feature

Currently, I'm working on adding `test_os.py` test file. Then I met this error.

```
Traceback (most recent call last):
  File "Lib/test/test_os.py", line 15, in <module>
    import mmap
ModuleNotFoundError: No module named 'mmap'
```
[mmap](https://github.com/python/cpython/blob/master/Modules/mmapmodule.c) module in CPython is not implemented on RustPython.

## Python Documentation

https://docs.python.org/3/library/mmap.html

coolreader18:
Hmm, yeah, that would be tricky. For now you can probably just comment out that import line and skip any tests that use `mmap`

darleybarreto:
Perhaps translating [fmmap](https://github.com/friedelwolff/fmmap) and combining with [libffi-rs](https://github.com/tov/libffi-rs) would do the trick for now.

youknowone:
nothing known progress

youknowone:
not working, but #3564

killme2008:
@youknowone  may i try this issue?

youknowone:
@killme2008 sure! you can work on #3564 if you want. If you want to start from zero base, still it will be helpful to understand how you can add a new module.

killme2008:
@youknowone  Great, thank you for the suggestion. I will read it.

fanninpm:
After a significant amount of effort, #3755 has been merged.

PR #3532, opened 2022-02-02
Fix f-string format spec

### LINKED ISSUE #3531: Bad f-string, causing CI issues
#### Code
```python
foo = "bar"
part_spec = ">+#10x"
print(f"{16:0{part_spec}}{foo}")
```

#### CPython
```text
00000+0x10bar
```

#### RustPython
```text
SyntaxError: Got error in f-string: Invalid expression: invalid syntax. Got unexpected token 'part_spec' at line 1 column 9
print(f"{16:0{part_spec}}{foo}")
        ^
```

#### Code
```python
import decimal
width = 10
precision = 4
value = decimal.Decimal('12.34567')
print(f'result: {value:{width}.{precision}}')
```

#### CPython
```text
result:      12.35
```

#### RustPython
```text
SyntaxError: Got error in f-string: Invalid expression: invalid syntax. Got unexpected token '{' at line 1 column 9
print(f'result: {value:{width}.{precision}}')
        ^
```

## Python Documentation

https://docs.python.org/3/reference/lexical_analysis.html#grammar-token-python-grammar-format_spec

@coolreader18 this has been plaguing the CI for the past day or so (because of the `--install-pip` step).

fanninpm:
I was using CPython 3.9's `Lib/test/test_fstring.py` as an implementation guide to track my progress, but there's still a long way to go, particularly w/r/t escape sequences.

Should fix #3531.

### DIFF
--- a/extra_tests/snippets/fstrings.py
+++ b/extra_tests/snippets/fstrings.py
@@ -47,6 +47,15 @@ assert f'{num=:>10}' == 'num=        42'
 spec = "0>+#10x"
 assert f"{16:{spec}}{foo}" == '00000+0x10bar'
 
+part_spec = ">+#10x"
+assert f"{16:0{part_spec}}{foo}" == '00000+0x10bar'
+
+# TODO: RUSTPYTHON, delete the next block once `test_fstring.py` can successfully parse
+assert f'{10:#{1}0x}' == '       0xa'
+assert f'{10:{"#"}1{0}{"x"}}' == '       0xa'
+assert f'{-10:-{"#"}1{0}x}' == '      -0xa'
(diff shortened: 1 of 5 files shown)


fanninpm:
Now, CI is blocked by `mmap` not being present (c.f. #2059)

youknowone:
is this happened because `get-pip.py` is changed?

fanninpm:
I think that it's because `pip` itself is vendoring dependencies that we currently don't support.

youknowone:
great, thank you!

youknowone (approved):

youknowone (approved):
looks good to me, but I am not very familiar with this part. so I want to let a few days for @coolreader18's review.

PR #3225, opened 2021-10-05
(TEST) Cpython 3.10

youknowone:

PR #3538, opened 2022-02-05
[TEST] Use CPython 3.10 as a CI reference

fanninpm:
Based on #3225.

If merged, this will close #3225.


fanninpm:
@youknowone this should be good to go.

youknowone (approved):

PR #3540, opened 2022-02-06
Add test/support/hashlib_helper from CPython 3.10

fanninpm:
Requesting @deantvv as a reviewer.

### DIFF
--- a/Lib/test/test_urllib2_localnet.py
+++ b/Lib/test/test_urllib2_localnet.py
@@ -9,6 +9,7 @@ import unittest
 import hashlib
 
 from test import support
+from test.support import hashlib_helper
 
 try:
     import ssl
@@ -315,6 +316,7 @@ class BasicAuthTests(unittest.TestCase):
         self.assertRaises(urllib.error.HTTPError, urllib.request.urlopen, self.server_url)
 
 
+@hashlib_helper.requires_hashdigest("md5", openssl=True)
 class ProxyAuthTests(unittest.TestCase):
     URL = "http://localhost"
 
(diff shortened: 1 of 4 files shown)


Lib/test/test_urllib2_localnet.py
```diff
@@ -315,14 +316,14 @@ def test_basic_auth_httperror(self):
         self.assertRaises(urllib.error.HTTPError, urllib.request.urlopen, self.server_url)
 
 
+@hashlib_helper.requires_hashdigest("md5", openssl=True)
```
deantvv, line 319: CPython doesn't specify `openssl=True`, is there a reason why this is needed?

deantvv (changes requested):
LGTM, only one difference with CPython may need some clarification.

Lib/test/test_urllib2_localnet.py
fanninpm: Which CPython version are you referencing? [`openssl=True` has been there since CPython 3.9.](https://github.com/python/cpython/blob/3.9/Lib/test/test_urllib2_localnet.py#L319)

deantvv: You're right. I accidentally compared it with CPython master branch.

deantvv (approved):
LTGM

Lib/test/test_urllib2_localnet.py
fanninpm: If it's removed in CPython's `main` branch, then we'll deal with it when CPython 3.11 gets released.

fanninpm:
@youknowone the CI failure here seems to be one of those spurious ones again.

youknowone (approved):
thanks, the fragile CI is stressful 😵 

@deantvv thank you for the review!

PR #3547, opened 2022-02-11
Align test/support/os_helper with CPython 3.10

fanninpm:
Requesting @deantvv as a reviewer.


deantvv (changes requested):
The PR looks great!
Just one nitpick. If we are trying to align with CPython 3.10, we could also remove `copy` (L626-628) part since it is only in CPython main branch.

youknowone (approved):

youknowone (approved):

PR #3555, opened 2022-02-18
Remove several special method of complex class

Snowapril:
This revision removes below special method from `complex` class
* `__int__`
* `__float__`
* `__floordiv__`
* `__mod__`
* `__divmod__`
* `__rfloordiv__`
* `__rmod__`
* `__rdivmod__`

According to [Cpython 3.10 changes](https://docs.python.org/3.10/whatsnew/3.10.html#removed), these special methods of complex class are removed as **they always raise `TypeError`**.
reference : https://bugs.python.org/issue41974

### DIFF
--- a/Lib/test/test_complex.py
+++ b/Lib/test/test_complex.py
@@ -1,6 +1,6 @@
 import unittest
+import sys
 from test import support
-from test.support import os_helper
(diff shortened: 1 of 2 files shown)


Snowapril:
Oh I should have removed related test cases. I'll applying it later

youknowone (approved):
thank you!

PR #3560, opened 2022-02-24
os: add `device_encoding` to wasi

deantvv:
Update `libc` to 0.2.119 to include libc [PR#2683](https://github.com/rust-lang/libc/pull/2683) which added support `device_encoding` on wasi.


coolreader18 (approved):
Oh neat, thanks!

PR #3561, opened 2022-02-24
Add __format__ type check in f-string

tkmikan:
RustPython doesn't check the return type of `__format__` in f-string, and will panic when `__format__` not returning `str`

```python
class A:
    def __format__(self, format_spec):
        return 0

a = A()
print(f'{a}')
# thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', vm/src/frame.rs:623:59
```

It should raise `TypeError: __format__ must return a str, not int` instead, as what `format(a, '')` and `'{}'.format(a)` are doing.

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -1076,8 +1076,16 @@ impl ExecutingFrame<'_> {
                 };
 
                 let spec = self.pop_value();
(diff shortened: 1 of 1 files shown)


youknowone (approved):
Thank you for contributing, and I am sorry to miss this PR from review for long time.

PR #3562, opened 2022-02-25
Fix Clippy lints introduced in Rust 1.59

fanninpm:
Clippy lints fixed:
- [`let_underscore_lock`](https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_lock)
- [`unnecessary_lazy_evaluations`](https://rust-lang.github.io/rust-clippy/master/index.html#unnecessary_lazy_evaluations)
- [`unwrap_or_else_default`](https://rust-lang.github.io/rust-clippy/master/index.html#unwrap_or_else_default)
- [`needless_late_init`](https://rust-lang.github.io/rust-clippy/master/index.html#needless_late_init)

### DIFF
--- a/Lib/test/test_threading.py
+++ b/Lib/test/test_threading.py
@@ -1061,6 +1061,7 @@ class ThreadingExceptionTests(BaseTestCase):
         lock = threading.Lock()
         self.assertRaises(RuntimeError, lock.release)
 
+    @unittest.skip("TODO: RUSTPYTHON, flaky test")
     def test_recursion_limit(self):
         # Issue 9670
         # test that excessive recursion within a non-main thread causes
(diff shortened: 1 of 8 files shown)


fanninpm:
@coolreader18 if you could take a look at that last commit and see if it's a false positive,

fanninpm:
Also, @qingshi163, you may want to check the `Fix let_underscore_lock Clippy error` commit just to be safe.

coolreader18:
LGTM, that's the same effect. Only thing I'd say is that drop is in the prelude, so you don't need the std::mem::

fanninpm:
Re: the `Fix let_underscore_lock Clippy error` commit, I didn't know if the intention was to drop the lock immediately or to hang onto it until the end of the scope. (c.f. [the documentation for that lint](https://rust-lang.github.io/rust-clippy/master/index.html#let_underscore_lock))

coolreader18 (approved):

Issue #2362, opened 2020-12-08
Can I use RustPython in Microcontrollers


jackieke724:
Hello all,

I wonder if I can use RustPython in a microcontroller which uses the [TockOS](https://github.com/tock/tock) kernel developed in Rust. Tock currently just has example user applications in Rust, as explained [here](https://github.com/tock/tock/blob/master/doc/Userland.md).

I figured I might be able to use RustPython in two ways:
1. add the RustPython dependency to my Rust applications and embed the python codes
2. directly load the bytecode generated by RustPython

The issue with the first option is that Tock applications are run in `#![no_std]` configurations. I could not build the RustPython dependency. For the second option, I am not sure if it is feasible at all. 

I saw @windelbouwman mentioned in an issue that running in microcontrollers was a stretch goal. What is the status now? 
How can I resolve the issue with the standard library? How much work would that be?

Thank you!

coolreader18:
There's not really any work on this yet, but it shouldn't be too hard! I whipped up [this](https://github.com/RustPython/RustPython/tree/no_std) branch for converting the parser, compiler etc to no_std, but I imagine the vm would be the main hurdle.

coolreader18:
Looks like `bincode` would be the main blocker: servo/bincode#265

windelbouwman:
This item is still a stretch goal, but I would love to see this happen!

Things to note / to be done:
- Be able to use rustpython in a no_std env (like @coolreader18 mentioned)
- Performance / code size: we need to reduce code size and improve performance overall. This has many benefits: faster runtime on desktop use, faster usage on microcontrollers, less power usage, less memory usage. Good thing all the way!
- The standard library. This boils down to being able to run rustpython with or without the standard library, and as such being able to run it on a microcontroller where there might be a flash memory restriction on amount of python files you can embed.
- The VM is where most of the action happens, and is I guess the hardest part to get going. I would start by porting the parser crate to no_std and see if this can be used on a microcontroller. Next up would be the compiler, and final step (and most difficult) the vm crate. I have no clue about code size at this point, but I'm guessing the vm is using the biggest amount of memory.

jackieke724:
Thank you for starting the no_std development again! My team is building a system which uses a RISC-V microcontroller, with a Tock kernel and a hardware accelerator, for accelerating machine learning applications. For now, we might proceed with Rust applications, but our end goal is to support Python applications. Hopefully, RustPython can help us achieve that goal.

mental32:
[...]

PR #3563, opened 2022-03-01
Switch to bincode 2 beta and make serde optional

coolreader18:
I think this'll be a draft til bincode 2 releases (which should hopefully be soonish?) but this is exactly what I wanted from it :blush: related to #2362


coolreader18:
Also related is bincode-org/bincode#511

PR #3566, opened 2022-03-10
Fix augassign

### LINKED ISSUE #3565: Augmented assignment loads the assignment target twice
The following code prints "here" twice. CPython only prints it once.
```python
def get_list():
    print("here")
    return [0]


get_list()[0] += 1
```

Relevant code:
https://github.com/RustPython/RustPython/blob/48fa217c4303993ef9cb00acb655d7b7eabaedad/compiler/src/compile.rs#L766-L773

coolreader18:
Resolves #3565


Snowapril (approved):
Oh.. I give a try this issue last night but failed to build 😂
Great work!!

youknowone (approved):

PR #3573, opened 2022-03-13
Update test_uuid

hrchu:
Most cases are passed now since we've updated uuid.py to 3.9.
There is still one failing case due to lack of os.fork().


youknowone (approved):
great, thank you!

PR #3574, opened 2022-03-13
U32 backed PyStr for non-unicode data

youknowone:
#935


youknowone:
replaced by #5587, which is a lot better idea

PR #3576, opened 2022-03-13
Sys unraisablehook

youknowone:


Snowapril:
There are many test cases related to `sys.unraisablehook` with below comments.
* # TODO: RUSTPYTHON, AttributeError: module 'sys' has no attribute 'unraisablehook'
* # TODO: RUSTPYTHON, sys.unraisablehook

youknowone:
@Snowapril  could you review this PR?

vm/src/stdlib/sys.rs
```diff
@@ -442,6 +442,52 @@ mod sys {
...
+            let str = match repr {
+                Ok(v) => format!("{v}"),
+                Err(_) => format!(
+                    "<object {} repr() failed>",
+                    unraisable.object.class().name()
+                ),
+            };
+            writeln!(stderr, "{str}");
+        } else if !vm.is_none(&unraisable.err_msg) {
+            writeln!(stderr, "{}:", unraisable.err_msg.repr(vm)?);
```
Snowapril, line 470: In cpython, use `__str__` only for this line of code. 
https://github.com/python/cpython/blob/main/Python/errors.c#L1316-L1323
```c
else if (err_msg != NULL && err_msg != Py_None) {
    if (PyFile_WriteObject(err_msg, file, Py_PRINT_RAW) < 0) {
        return -1;
    }
    if (PyFile_WriteString(":\n", file) < 0) {
        return -1;
    }
}
```
https://github.com/python/cpython/blob/main/Objects/fileobject.c#L128-L130
```c
if (flags & Py_PRINT_RAW) {
    value = PyObject_Str(v);
}
else
    value = PyObject_Repr(v);
```

vm/src/stdlib/sys.rs
```diff
@@ -442,6 +442,52 @@ mod sys {
...
+    fn __unraisablehook__(
+        unraisable: PyRef<UnraisableHookArgs>,
+        vm: &VirtualMachine,
+    ) -> PyResult<()> {
+        use super::PyStderr;
+
+        let stderr = PyStderr(vm);
+        if !vm.is_none(&unraisable.object) {
+            if !vm.is_none(&unraisable.err_msg) {
+                write!(stderr, "{}: ", unraisable.err_msg.repr(vm)?);
```
Snowapril, line 455: Oh here also should be `str`
https://github.com/python/cpython/blob/main/Python/errors.c#L1292-L1299
```c
if (err_msg != NULL && err_msg != Py_None) {
    if (PyFile_WriteObject(err_msg, file, Py_PRINT_RAW) < 0) {
    	return -1;
    }
    if (PyFile_WriteString(": ", file) < 0) {
    	return -1;
    }
}
```

vm/src/stdlib/sys.rs
```diff
@@ -442,6 +442,52 @@ mod sys {
...
+        } else if !vm.is_none(&unraisable.err_msg) {
+            writeln!(stderr, "{}:", unraisable.err_msg.repr(vm)?);
+        }
+
+        // TODO: print received traceback
+        let tb_module = vm.import("traceback", None, 0)?;
+        let print_stack = tb_module.get_attr("print_stack", vm)?;
+        vm.invoke(&print_stack, ())?;
+
+        if let Ok(repr) = unraisable.exc_value.repr(vm) {
```
Snowapril, line 478: Sorry. also here
https://github.com/python/cpython/blob/main/Python/errors.c#L1379-L1393
```c
if (exc_value && exc_value != Py_None) {
    if (PyFile_WriteString(": ", file) < 0) {
        return -1;
    }
    if (PyFile_WriteObject(exc_value, file, Py_PRINT_RAW) < 0) {
        _PyErr_Clear(tstate);
        if (PyFile_WriteString("<exception str() failed>", file) < 0) {
            return -1;
        }
    }
}
```

Snowapril:
Isn't the print for `__module__` and `__qualname__` from `unraisable.exc_type` necessary for this PR at now?
If it need to be implemented later, maybe it is good to add comment about it 😊

https://github.com/python/cpython/blob/main/Python/errors.c#L1336-L1377
```c
assert(PyExceptionClass_Check(exc_type));
PyObject *modulename = PyObject_GetAttr(exc_type, &_Py_ID(__module__));
if (modulename == NULL || !PyUnicode_Check(modulename)) {
    Py_XDECREF(modulename);
    _PyErr_Clear(tstate);
    if (PyFile_WriteString("<unknown>", file) < 0) {
        return -1;
    }
}
else {
    if (!_PyUnicode_Equal(modulename, &_Py_ID(builtins)) &&
        !_PyUnicode_Equal(modulename, &_Py_ID(__main__))) {
        if (PyFile_WriteObject(modulename, file, Py_PRINT_RAW) < 0) {
            Py_DECREF(modulename);
            return -1;
        }
        Py_DECREF(modulename);
        if (PyFile_WriteString(".", file) < 0) {
            return -1;
        }
    }
    else {
        Py_DECREF(modulename);
    }
}
PyObject *qualname = PyType_GetQualName((PyTypeObject *)exc_type);
if (qualname == NULL || !PyUnicode_Check(qualname)) {
    Py_XDECREF(qualname);
    _PyErr_Clear(tstate);
    if (PyFile_WriteString("<unknown>", file) < 0) {
        return -1;
    }
}
else {
    if (PyFile_WriteObject(qualname, file, Py_PRINT_RAW) < 0) {
        Py_DECREF(qualname);
        return -1;
    }
    Py_DECREF(qualname);
}
```

youknowone:
thanks, I fixed it based on the code

derive/src/pystructseq.rs
```diff
@@ -36,8 +36,11 @@ pub(crate) fn impl_pystruct_sequence(input: DeriveInput) -> Result<TokenStream>
                 ::rustpython_vm::builtins::PyTuple::new_unchecked(items.into_boxed_slice())
             }
         }
-        impl ::rustpython_vm::function::IntoPyObject for #ty {
-            fn into_pyobject(self, vm: &::rustpython_vm::VirtualMachine) -> ::rustpython_vm::PyObjectRef {
+        impl ::rustpython_vm::PyValue for #ty {
+            fn class(_vm: &::rustpython_vm::VirtualMachine) -> &rustpython_vm::builtins::PyTypeRef {
+                <Self as ::rustpython_vm::StaticType>::static_type()
+            }
+            fn into_object(self, vm: &::rustpython_vm::VirtualMachine) -> ::rustpython_vm::PyObjectRef {
```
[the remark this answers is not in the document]
coolreader18: Like, PyValue implies PyObjectPayload, and for structseqs, the struct definitions that the derive() is on *aren't payloads*, PyTuple is the payload. And ideally we'd be able to have it do `#[subclass = PyTuple] struct PyStructTime { super: PyTuple }` but we're not there yet, I still don't have a good way to solve it.

vm/src/stdlib/sys.rs
```diff
@@ -442,6 +442,69 @@ mod sys {
...
+        if !vm.is_none(&unraisable.object) {
+            if !vm.is_none(&unraisable.err_msg) {
+                write!(stderr, "{}: ", unraisable.err_msg.str(vm)?);
+            } else {
+                write!(stderr, "Exception ignored in: ");
+            }
+            // exception in del will be ignored but printed
+            let repr = &unraisable.object.repr(vm);
+            let str = match repr {
+                Ok(v) => format!("{v}"),
```
coolreader18, line 462: I think `v.to_string()` is more idiomatic here.

vm/src/stdlib/sys.rs
```diff
@@ -442,6 +442,69 @@ mod sys {
         .into_struct_sequence(vm))
     }
 
+    #[pyfunction]
+    fn __unraisablehook__(
+        unraisable: PyRef<UnraisableHookArgs>,
```
coolreader18, line 447: e.g. for this (instead of the PyValue thing) you could write a custom TryFromObject for `UnraisableHookArgs` that just checks if `x.isinstance(Self::static_type())` and then extracts the tuple from there.

coolreader18 (commented):
Some comments on structseq and other nits

vm/src/stdlib/sys.rs
```diff
@@ -442,6 +442,69 @@ mod sys {
...
+            }
+        }
+        writeln!(stderr, "");
+        // TODO: call file.flush()
+
+        Ok(())
+    }
+
+    #[pyfunction]
+    fn unraisablehook(unraisable: PyRef<UnraisableHookArgs>, vm: &VirtualMachine) {
```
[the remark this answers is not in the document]
youknowone: Still `sys.unraisablehook` is not `sys.__unraisablehook__`, but I regard this result as bug of `#[pymodule]` macro for current version

derive/src/pystructseq.rs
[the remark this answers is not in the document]
youknowone: reverted

youknowone:
@coolreader18 I splited macro work to  #3613 and rebased this PR on it.

vm/src/stdlib/sys.rs
```diff
@@ -442,6 +443,66 @@ mod sys {
...
+                write!(stderr, "<exception str() failed>");
+            }
+        }
+        writeln!(stderr);
+        // TODO: call file.flush()
+
+        Ok(())
+    }
+
+    #[pyattr]
```
youknowone, line 498: @coolreader18 Now they will be the same object

PR #3578, opened 2022-03-13
Fix bool un/marshaling

dieterplex:
Test: import marshal; assert marshal.loads(marshal.dumps(True)) == True
Test: import marshal; assert marshal.loads(marshal.dumps(False))== False


youknowone (approved):
looks good, thank you!
If you want to find more targets, please feel free to let me know

PR #3585, opened 2022-03-28
Fix dict.fromkeys to return correct type

dieterplex:


Lib/test/test_dict.py
```diff
@@ -213,7 +213,6 @@ def __next__(self):
         self.assertRaises(ValueError, {}.update, [(1, 2, 3)])
 
     # TODO: RUSTPYTHON
-    @unittest.expectedFailure
```
fanninpm, line 216: When you remove this line, please also remove the `# TODO: RUSTPYTHON` line above it.

dieterplex:
Now `UserDict` related assertions in `test_fromkeys()` can be passed.
But still unable to pass `baddict#` checks.
Also observed that `__init__` and `__setitem__` magics seems not invoked when dict creation.

vm/src/builtins/dict.rs
```diff
@@ -150,14 +150,25 @@ impl PyDict {
         iterable: ArgIterable,
         value: OptionalArg<PyObjectRef>,
         vm: &VirtualMachine,
-    ) -> PyResult<PyRef<Self>> {
+    ) -> PyResult {
+        if !class.is(PyDict::class(vm)) {
+            if let Some(ref slot_new) = class.slots.new.load() {
+                let slot_new_cls = slot_new(class, FuncArgs::default(), vm)?;
```
youknowone, line 156: The casue of CI error is lack of calling `__init__`. I found `PyType::call` from `Callable` trait calls both of them.

```suggestion
            let slot_new_cls = PyType::call(&class, ().into(), vm)?;
```

Rather than empty tuple, we can pass `KwArgs` to init dict. 
```suggestion
            let value = value.unwrap_or_none(vm);
            let mut items = IndexMap::<PyObjectRef, PyObjectRef>::new();
            // TODO: fill key-value to items 
            let slot_new_cls = PyType::call(&class, KwArgs::new().into(), vm)?;
```

dieterplex: Do you mean kwargs for callable or when invoking fromkeys?

BTW, without load check or some way to distinguish the object, it would recursive call `fromkeys` method till stack overflow… Is there a way to prevent this? Especially when using vm_trace! I run into many times.

youknowone: I think you shouldn't call `fromkeys` inside there. It will be always recursive call in that case.

I checked same funciton in CPython and found my comments about `kwargs` part is wrong. Forget about it.
ref: https://github.com/python/cpython/blob/main/Objects/dictobject.c#L2260-L2351

- It calls the type itself (`_PyObject_CallNoArg(cls)` is `PyType::call(&cls, ().into(), vm)` in RustPython
- If the created object's type is exactly `dict`, it follows the optimized path. (For now, let's regard it same as current implementation)
- Otherwise, it calls setitem over the iterable.

You set items by calling `fromkeys`, but it seems calling setitem for every item is the way how it does.

dieterplex: It make sense. Thanks for your advise!

fanninpm:
@youknowone Windows failure is because Chocolatey couldn't download openssl. Can we rerun just that job?

dieterplex:
Update commit message to trigger rerun.

youknowone (approved):
looks great, thank you!

PR #3597, opened 2022-04-09
Add `Reduce` method in reverse_iter

jeon-yongjun:
add `reduce` method to reverse_iter and obliterated the failure that was foreseen by it

### DIFF
--- a/vm/src/builtins/dict.rs
+++ b/vm/src/builtins/dict.rs
@@ -837,6 +837,8 @@ macro_rules! dict_view {
                     internal: PyMutex::new(PositionIterInternal::new(dict, position)),
                 }
             }
+
+            #[allow(clippy::redundant_closure_call)]
             #[pymethod(magic)]
             fn reduce(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyTupleRef {
                 let iter = builtins_reversed(vm).to_owned();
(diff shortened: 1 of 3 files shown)


vm/src/builtins/dict.rs
```diff
@@ -837,6 +837,20 @@ macro_rules! dict_view {
...
             }
+            #[pymethod(magic)]
+            fn reduce(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyTupleRef {
+                let iter = builtins_reversed(vm).to_owned();
+                let internal = zelf.internal.lock();
+                // TODO: entries must be reversed too
+                let entries = match &internal.status {
+                    IterStatus::Active(dict) => dict
+                        .into_iter()
+                        .map(|(key, value)| ($result_fn)(vm, key, value))
```
youknowone, line 848: here is a clippy error.

youknowone (changes requested):
This patch also fixed a test  `test_enum::test_unique_composite`. Please remove expectedFailure decorator from the test.

Snowapril:
> This patch also fixed a test `test_enum::test_unique_composite`. Please remove expectedFailure decorator from the test.

It seems temporary error. we cant replicate our environment(window, ubuntu).
We add `expectedFailure` again and it works now 😊

youknowone (approved):

PR #3598, opened 2022-04-09
Modify timeout print message

hwi-ya:
Change the run result msg of 'socket.timeout'

`socket.timeout` ->`TimeoutError`

### DIFF
--- a/stdlib/src/socket.rs
+++ b/stdlib/src/socket.rs
@@ -661,14 +661,11 @@ mod _socket {
         vm.ctx.exceptions.os_error.to_owned()
     }
 
-    #[pyattr(once)]
+    #[pyattr]
     fn timeout(vm: &VirtualMachine) -> PyTypeRef {
-        vm.ctx.new_exception_type(
-            "socket",
-            "timeout",
-            Some(vec![vm.ctx.exceptions.os_error.to_owned()]),
-        )
+        vm.ctx.exceptions.timeout_error.to_owned()
     }
+
     #[pyattr(once)]
     fn herror(vm: &VirtualMachine) -> PyTypeRef {
         vm.ctx.new_exception_type(


stdlib/src/socket.rs
```diff
@@ -80,14 +80,11 @@ mod _socket {
         vm.ctx.exceptions.os_error.clone()
     }
 
-    #[pyattr(once)]
+    #[pyattr(name = "timeout", once)]
```
youknowone, line 83: when `name` is missing, it follows the following identifier's name. Because the function name is `timeout`, this is redundant.

stdlib/src/socket.rs
```diff
@@ -80,14 +80,11 @@ mod _socket {
...
     fn timeout(vm: &VirtualMachine) -> PyTypeRef {
-        vm.ctx.new_exception_type(
-            "socket",
-            "timeout",
-            Some(vec![vm.ctx.exceptions.os_error.clone()]),
-        )
+        PyType::new_simple_ref("TimeoutError", &error(vm)).unwrap()
```
youknowone, line 85: This is not OSError anymore after this change. Is this not an OSError?

youknowone (changes requested):

stdlib/src/socket.rs
Snowapril: Oh yes. It is just TimeoutError which already exist in the RustPython.
https://github.com/python/cpython/blob/3.10/Modules/socketmodule.c#L7130
But actually this will not be act as I expected. I'll tell him to fix it 😊

hwi-ya: > This is not OSError anymore after this change. Is this not an OSError?

i modify

stdlib/src/socket.rs
```diff
@@ -82,12 +82,9 @@ mod _socket {
 
     #[pyattr(once)]
```
youknowone, line 83: When the function is guaranteed to return same object everytime, we don't need `once`.

```suggestion
    #[pyattr]
```

youknowone (changes requested):

youknowone (approved):

PR #3599, opened 2022-04-10
Add select.error

himmelmin:
`builtins::PyTypeRef` added
Added `socket error` to `select`


youknowone (approved):
Thank you for contributing!

PR #3601, opened 2022-04-10
add thisclass to the original typ

simbajisu:
add thisclass to the original typ

### DIFF
--- a/vm/src/builtins/pysuper.rs
+++ b/vm/src/builtins/pysuper.rs
@@ -107,6 +107,11 @@ impl PySuper {
         Ok(Self { typ, obj })
     }
 
+    #[pyproperty(magic)]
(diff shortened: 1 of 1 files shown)


vm/src/builtins/pysuper.rs
```diff
@@ -107,6 +107,11 @@ impl PySuper {
         Ok(Self { typ, obj })
     }
 
+    #[pyproperty(magic)]
+    fn thisclass(&self) -> PyObjectRef {
+        self.typ.as_object().to_owned()
+    }
```
youknowone, line 113:
```suggestion
    fn thisclass(&self) -> PyTypeRef {
        self.typ.clone()
    }
```

youknowone (changes requested):

youknowone (approved):
Thank you for contributing!

PR #3603, opened 2022-04-12
test: update test.support, test_io and test_os to 3.10

deantvv:
Update several tests to python 3.10, especially `test.support.warnings_helper` duplicate


youknowone (approved):
great, you always make the huge useful updates, thank you!

PR #3606, opened 2022-04-13
split pyobject.rs to multiple files

youknowone:
- Reorganize `pyobject` as a internal module importing all public items from `pyobject.rs` and `pyobjectrc.rs`.
- Split `TryFromObject`, `TransmuteFromObject`, `TryFromBorrowedObject` into `vm::convert`. They will still be exposed to vm crate root.
- Split `PyArithmaticValue`, `PyComparisonValue` to `vm::function` through `vm::function::arithmetic`. `IntoPyRef` through  `vm::function`. They will not be exposed to crate too.
- Split `PyClassDef`, `PyClassImpl`, `StaticType` into `vm::pyclass` as the same level as `vm::pyobject`. They will not be exposed to crate.
- Hide uncommon types `PyAttributse`, `PyLease`, `PyThreadingConstraint` from crate root.


youknowone:
@coolreader18 I am going to merge this to avoid massive conflict. The changes itself is not easy to revert, but the UX can be reverted easily by just exposing a few types to root again

coolreader18:
Oh, nice, looks good!

PR #3613, opened 2022-04-14
derive(TryIntoPyStructSequence)

youknowone:


vm/src/types/structseq.rs
```diff
@@ -0,0 +1,90 @@
...
+    builtins::{PyTuple, PyTupleRef, PyTypeRef},
+    pyclass::{PyClassImpl, StaticType},
+    pyobject::PyContext,
+    AsObject, PyObjectRef, PyRef, PyResult, PyValue, VirtualMachine,
+};
+
+#[pyimpl]
+pub trait PyStructSequence: StaticType + PyClassImpl + Sized + 'static {
+    const FIELD_LEN: usize;
+    const FIELD_NAMES: &'static [&'static str];
```
coolreader18, line 11: This is kinda redundant; is there a reason not to just do `FIELD_NAMES.len()` as before?

vm/src/types/structseq.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    fn into_tuple(self, vm: &VirtualMachine) -> PyTuple;
+
+    fn into_struct_sequence(self, vm: &VirtualMachine) -> PyTupleRef {
+        self.into_tuple(vm)
+            .into_ref_with_type(vm, Self::static_type().clone())
+            .unwrap()
+    }
+
+    fn try_elements_from(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<Vec<PyObjectRef>> {
```
coolreader18, line 21: Ideally this could return `Box<[PyObjectRef; Self::FIELD_LEN]>` but that's a generic parameter in a const expression :/

derive/src/pystructseq.rs
```diff
@@ -38,7 +46,24 @@ pub(crate) fn impl_pystruct_sequence(input: DeriveInput) -> Result<TokenStream>
...
+}
+
+pub(crate) fn impl_pystruct_sequence_try_from_object(input: DeriveInput) -> Result<TokenStream> {
+    let field_names = field_names(&input)?;
+    let ty = &input.ident;
+    let ret = quote! {
+        impl ::rustpython_vm::TryFromObject for #ty {
+            fn try_from_object(vm: &::rustpython_vm::VirtualMachine, seq: ::rustpython_vm::PyObjectRef) -> ::rustpython_vm::PyResult<Self> {
+                let seq = Self::try_elements_from(seq, vm)?;
+                let mut iter = seq.into_iter();
```
coolreader18, line 63: Yeah, this could probably be `let seq: [PyObjectRef; <#ty as ::rustpython_vm::types::PyStructSeq>::FIELD_NAMES.len()] = seq.try_into().unwrap();`

coolreader18 (approved):
LGTM!

vm/src/types/structseq.rs
youknowone: we usually have solution with generics 😎  https://github.com/RustPython/RustPython/pull/3613/files#diff-c134527406501aa4d7eebd8c9dcd3cfbc8e273cc81cac85e4cba7bf57ddcf707R20-R23

derive/src/pystructseq.rs
youknowone: no idea how to handle index in `quote!()`

PR #3616, opened 2022-04-15
Fix CI

youknowone:


youknowone:
The remaining false positive one will be fixed in clippy side.

PR #3618, opened 2022-04-15
Move string intern to PyContext

youknowone:
so that we can intern strings before vm or even without vm for preparation step.

### DIFF
--- a/vm/src/intern.rs
+++ b/vm/src/intern.rs
@@ -1,0 +1,113 @@
+use crate::{
+    builtins::{PyStr, PyTypeRef},
+    common::lock::PyRwLock,
+    convert::ToPyObject,
+    PyRef, PyRefExact,
+};
+use std::ops::Deref;
+
+#[derive(Debug)]
+pub struct StringPool {
+    inner: PyRwLock<std::collections::HashSet<CachedPyStrRef, ahash::RandomState>>,
+}
+
+impl Default for StringPool {
+    fn default() -> Self {
+        Self {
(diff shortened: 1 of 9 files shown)


vm/src/intern.rs
```diff
@@ -0,0 +1,121 @@
...
+
+#[derive(Debug)]
+pub struct CachedPyStrRef {
+    inner: PyRefExact<PyStr>,
+}
+
+impl Clone for CachedPyStrRef {
+    fn clone(&self) -> Self {
+        Self {
+            inner: unsafe { PyRefExact::new_unchecked(self.inner.clone()) },
```
coolreader18, line 52: Would probably be good to impl Clone for PyRefExact and then you can just derive

vm/src/intern.rs
```diff
@@ -0,0 +1,121 @@
...
+        }
+    }
+}
+
+impl std::hash::Hash for CachedPyStrRef {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        self.inner.as_str().hash(state)
```
coolreader18, line 59: Hmmm.. this should probably use the existing hash but idk how to do that but use a custom hash function for get(). Maybe hashbrown could do it

coolreader18: Yeah, maybe with the raw_entry api you could do it

coolreader18: Oh, nevermind actually, this is fine since it never needs to rehash the keys anyway I think.

PR #3622, opened 2022-04-16
gen ast on build phase

youknowone:


youknowone:
@coolreader18 removed python dependent but still generate it by default

ast/build.rs
```diff
@@ -0,0 +1,130 @@
...
+
+fn read_file_content(path: impl AsRef<Path>) -> std::io::Result<String> {
+    let path = path.as_ref();
+    let mut file = File::open(path)?;
+    let mut content = String::new();
+    file.read_to_string(&mut content)?;
+    Ok(content)
+}
+
+fn file_changed(path1: impl AsRef<Path>, path2: impl AsRef<Path>) -> std::io::Result<bool> {
```
coolreader18, line 122: mtime isn't as reliable as something like a checksum; I'm pretty sure git checkout touches the mtime and ideally this would only be run when necessary (i.e. in development, not when building as a dependency or something)

ast/build.rs
```diff
@@ -0,0 +1,130 @@
...
+    let mod_path = {
+        let mut path = build_dir;
+        path.push("ast_mod.rs");
+        path
+    }
+    .to_str()
+    .expect("mod path is not str representable")
+    .to_owned();
+
+    // // TODO: do this when pyo3 is fixed on apple silicon
```
coolreader18, line 38: I think shelling out is better than depending on pyo3; it's ideally an uncommon situation and adding a dependency would increase build time for a code path that's rarely used.

coolreader18 (changes requested):
Oh wait I thought this was for auto-updating the checked in file like in #3689. No, I disagree with this for the same reasons as the other one, python shouldn't be a build dependency of rustpython. Plus, the current setup wouldn't work for both the stdlib module and the ast crate - you'd have to have 2 different build scripts; `$AST_CARGO_MANIFEST_DIR/../stdlib` isn't always guaranteed to be the stdlib directory, like if we have this on crates.io.

ast/build.rs
youknowone: this function does not use mtime. it actually read the file content and compare.

PR #3630, opened 2022-04-17
Rename PyObjectView -> Py and PyValue -> PyPayload

### LINKED ISSUE #3631: [RFC] Give proper names to types and traits
## Summary

By adding and removing types ad-hoc, we now have many types and traits related to describing Python objects. Giving proper names for them will reduce confusions.

## Detailed Explanation

- preview: #3630

### Types related to managed references

As-is 

 |       Management         |    Typed     |    Untyped    |
 | ------------------------ | ------------ | --------------- | 
 | Interpreter-independent  | `PyObjectView<T>`        | PyObject      |
 | Reference-counted          | `PyRef<T>`                      | PyObjectRef   |
 | Weak                                  | `PyWeakRef<T>`            | PyObjectWeak |

Giving names to show their relationship will be great. It will be mostly cleared by renaming `PyObjectView` as `Py`, by following @coolreader18 's original suggestion when they worked about it.

I additionally suggest to remove PyObjectWeak and just go ahead by its internal type `PyRef<PyWeak>`, because `PyObjectWeak` doesn't look like `PyRef` at all. We mostly handle `PyRef` in the code, so it will be confusing.

To-be v1

 |       Management        |    Typed     |    Untyped    |
 |-------------------------|--------------|---------------|
 | Interpreter-independent | `Py<T>`        | `PyObject `     |
 | Reference-counted       | `PyRef<T> `    | `PyObjectRef `  |
 | Weak                    | `PyWeakRef<T>` | `PyRef<PyWeak>` |

Unfortunately, ur common naming convention implies `PyWeakRef` is `PyRef<PyWeak>`. So the left and right side of weak type is not distinguishable by the name. `PyRefWeak` or `WeakPyRef` will not be confusing.

To-be v2

 |       Management        |    Typed     |    Untyped    |
 |-------------------------|--------------|---------------|
 | Interpreter-independent | `Py<T>`        | `PyObject `     |
 | Reference-counted       | `PyRef<T> `    | `PyObjectRef `  |
 | Weak                    | `PyRefWeak<T>` | `PyRef<PyWeak>` |

## PyValue -> PyPayload

By looking back the history, there was `PyObjectPayload` and `PyObjectPayload2`  but the latter one is renamed to `PyValue`. For now, it looks like describing a payload as its past name. We use functions like `fn payload()`, then `fn payload_is()` which indirectly based on `PyValue` through (still alive) PyObjectPayload.

We actually don't have concrete concept of "value". 

## Unresolved Questions

Is `PyObjectWeak` intentionally wrapped as different object to prevent to call PyRef methods?

youknowone:
As I remember, the name `Py` was @coolreader18's idea, right?

After reviewing their relationships, I agree this is a consistent and not confusing name than now, by showing their relationships in the name.

closes #3631

### DIFF
--- a/common/src/rc.rs
+++ b/common/src/rc.rs
@@ -1,7 +1,7 @@
 #[cfg(not(feature = "threading"))]
-use std::rc::{Rc, Weak};
+use std::rc::Rc;
 #[cfg(feature = "threading")]
-use std::sync::{Arc, Weak};
+use std::sync::Arc;
 
 // type aliases instead of newtypes because you can't do `fn method(self: PyRc<Self>)` with a
 // newtype; requires the arbitrary_self_types unstable feature
@@ -10,5 +10,3 @@ use std::sync::{Arc, Weak};
 pub type PyRc<T> = Arc<T>;
 #[cfg(not(feature = "threading"))]
 pub type PyRc<T> = Rc<T>;
-
-pub type PyWeak<T> = Weak<T>;
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -17,8 +17,8 @@ mod doc;
 mod from_args;
 mod pyclass;
 mod pymodule;
+mod pypayload;
 mod pystructseq;
-mod pyvalue;
 
 use error::{extract_spans, Diagnostic};
 use proc_macro2::TokenStream;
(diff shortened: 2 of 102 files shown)


coolreader18 (approved):
LGTM!

youknowone:
@coolreader18 what's your idea about `PyObjectWeak`? Should it be a wrapper of `PyRef<PyWeak>` or PyRef itself?

PR #3641, opened 2022-04-19
Small fixes

coolreader18:


vm/src/stdlib/time.rs
```diff
@@ -378,22 +378,19 @@ mod time {
...
             let seq: Vec<_> = seq.try_to_value(vm)?;
-            if seq.len() != 9 {
-                return Err(
-                    vm.new_type_error("time.struct_time() takes a sequence of length 9".to_owned())
-                );
-            }
-            let mut i = seq.into_iter();
+            let v: [PyObjectRef; 9] = seq.try_into().map_err(|_| {
+                vm.new_type_error("time.struct_time() takes a sequence of length 9".to_owned())
+            })?;
```
youknowone, line 383: Will you check #3613 for this? The idea using array looks good. Maybe we can put it in the general way.

stdlib/src/socket.rs
```diff
@@ -1667,7 +1657,9 @@ mod _socket {
                     netioapi::ConvertInterfaceLuidToNameW(luid, buf.as_mut_ptr(), buf.len())
                 };
                 if ret == 0 {
-                    Ok(widestring::WideCString::from_vec_with_nul(&buf[..]).unwrap())
+                    Ok(widestring::WideCString::from_ustr_truncate(
+                        widestring::WideStr::new(&buf[..]),
```
youknowone, line 1661: here is build failure

Issue #3609, opened 2022-04-14
Add `aiter` and `anext` builtin function


youknowone:
## Feature

- [ ] `builtins.aiter`: https://docs.python.org/3/library/functions.html#aiter
- [ ] `builtins.anext`: https://docs.python.org/3/library/functions.html#anext

Snowapril:
`builtins.aiter` can be implemented quite easy with just a few codes.
but for `builtins.anext`, it seems `anext_awaitable` must be implemented first.
```c
static PyObject *
builtin_anext_impl(PyObject *module, PyObject *aiterator,
                   PyObject *default_value)
{
    // ...skip
    awaitable = (*t->tp_as_async->am_anext)(aiterator);
    if (default_value == NULL) {
        return awaitable;
    }

    PyObject* new_awaitable = PyAnextAwaitable_New(
            awaitable, default_value);
    Py_DECREF(awaitable);
    return new_awaitable;
}
```

sum12:
Thanks for the "good for issue". I am probably misunderstanding things. 
the slots for `tp_as_async->am_aiter` is missing. According to cpython that would be the correct way to check if the object has `__aiter__` implementaed, Is this the correct way to proceed, probably an outline of how this could be done would be great

youknowone:
@sum12 you are right. that's the right way. But instead of the right way, you can implement it by calling `__aiter__` directly.

If you are interested in the right way, You can do it by looking at other slots, like `IterNext` or `Callable`. The right way would not be the 'good first issue'.

youknowone:
Thanks @sum12 and @YashSuthar983 ! this is done!

Thanks @Snowapril for advices

PR #3646, opened 2022-04-19
implement aiter

sum12:
the caller is expected to make sure that the passed in object does in
fact supports the protocol

relates #3609

### DIFF
--- a/Lib/test/test_asyncgen.py
+++ b/Lib/test/test_asyncgen.py
@@ -543,8 +543,6 @@ class AsyncGenAsyncioTest(unittest.TestCase):
         self.loop.run_until_complete(consume())
         self.assertEqual(results, [1, 2])
 
-    # TODO: RUSTPYTHON, NameError: name 'aiter' is not defined
-    @unittest.expectedFailure
     def test_aiter_idempotent(self):
         async def gen():
             yield 1
@@ -766,8 +764,6 @@ class AsyncGenAsyncioTest(unittest.TestCase):
         run_test(test5)
         run_test(test6)
 
-    # TODO: RUSTPYTHON, NameError: name 'aiter' is not defined
-    @unittest.expectedFailure
     def test_aiter_bad_args(self):
         async def gen():
             yield 1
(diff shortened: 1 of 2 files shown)


youknowone:
Will you add a simple test if it actually calls `__aiter__` from the object?
When possible, we prefer to add tests from CPython's test, but if not possible.
Otherwise, adding a file to `extra_tests/snipepts` will be enough.
Then our CI will ensure their behavior is same in both CPython and RustPython.

fanninpm:
> Will you add a simple test if it actually calls `__aiter__` from the object?
> When possible, we prefer to add tests from CPython's test, but if not possible.
> Otherwise, adding a file to `extra_tests/snipepts` will be enough.
> Then our CI will ensure their behavior is same in both CPython and RustPython.

I think we can just update our copy of [`test_asyncgen.py`](https://github.com/RustPython/RustPython/blob/787d9d0bbb59da79fa061b870c234314877ff9b0/Lib/test/test_asyncgen.py) to the version from [CPython 3.10](https://github.com/python/cpython/blob/3.10/Lib/test/test_asyncgen.py).

youknowone:
I think you already added tests from the CPython, but here is a guideline if you want to know the regular way: https://github.com/RustPython/RustPython/wiki/How-to-update-test-files

youknowone (changes requested):
looks good for codes. I have one more request for the commit message. For the commit for test update (https://github.com/RustPython/RustPython/pull/3646/commits/64cb090f96a130a4e18b6ffb0ac8ec9ccdea80e1), please leave the updated source of the CPython version. That will be very useful for future contributor of the file. Because `git log Lib/test/test_asyncgen.py` will show how it is changed by the history just by looking for the log.

youknowone:
Because #3647 updated the test file, rebase will be enough

youknowone (approved):
looks good, thank you!

youknowone:
@sum12 If you want, converting it to use slots would be also an approachable task

PR #3650, opened 2022-04-22
Stop depending on serde_json

coolreader18:
Should speed up compilation like, slightly


fanninpm:
I almost forgot that `cargo --timings` was a thing.

youknowone (approved):

PR #3658, opened 2022-04-24
Move crt_fd to common

coolreader18:


common/src/crt_fd.rs
```diff
@@ -21,10 +20,31 @@ pub type Offset = libc::off_t;
 #[cfg(windows)]
 pub type Offset = libc::c_longlong;
 
+// copied from stdlib::os
```
youknowone, line 23: how about moving rather than copying? it seems stdlib::os can call this function also from common.

coolreader18: fair, I was being lazy :P

youknowone (approved):

PR #3662, opened 2022-04-26
Relocate `PyObject_Generic*` into object protocol

Snowapril:
This revision does the below things.
* Relocate `generic_setattr` from `builtins::object` into `protocol::object`
* Relocate `generic_getattr`, `generic_getattr_opt` from `vm` into object protocol
* Remove duplicate `generic_getattr` in `builtins::object`

ref #3244 
https://docs.python.org/3/c-api/object.html


youknowone (approved):
looks great!

Issue #3611, opened 2022-04-14
Implement missing `__reduce__` and `__reduce_ex__` for several types


Snowapril:
## Feature
Many existing test cases check that `pickle.dumps` and `pickle.loads` of a given type return the exact same value. Almost failed pickle test in RustPython, lack `__reduce__` method.

- [x] `arrayiterator.__reduce__` (test_array::test_iterator_pickle) — #3868
- [x] `filter.__reduce__` (test_builtin::test_filter_pickle) — #3732
- [x] `map.__reduce__` (test_builtin::test_map_pickle) — #3734
- [x] `ImportError.__reduce__` (test_exceptions::test_copy_pickle) — #4973
- [x] `OSError.__reduce__` — #4191 
- [x] `BaseException.__reduce__` — #3665
- [x] `method.__reduce__` — #4237 
- [ ] `groupby.__reduce__`
- [ ] `_grouper.__reduce__`
- [ ] `tee.__reduce__`
- [ ] `cycle.__reduce__`
- [x] `dropwhile.__reduce__` — #4201 
- [x] `takewhile.__reduce__` — #4190
- [x] `islice.__reduce__` — #4390
- [x] `starmap.__reduce__` — #3988 
- [x] `chain.__reduce__` — #4232
- [x] `product.__reduce__` #5089 
- [x] `combinations.__reduce__` — #3931, #4272
- [ ] `combinations_with_replacement.__reduce__`
- [x] `permutations.__reduce__` #5029
- [x] `accumulate.__reduce__` #4434
- [x] `compress.__reduce__` — #3791
- [x] `filterfalse.__reduce__` — #3745
- [x] `count.__reduce__` — #3737
- [x] `zip_longest.__reduce__` #4235
- [x] `NotImplemented.__reduce__` — #3738
- [x] `types.SimpleNamespace.__reduce__` #4975 

These listed elements are almost similar and quite simple to implement.
Please take a look at many PR and existing codes for `__reduce__`

youknowone:
itertools pickle is going to be deprecated https://github.com/python/cpython/issues/101588

youknowone:
Since all of remaining checklist is about itertools, I am going to close this issue.
Thank you so much everyone who worked on this issue!

PR #3665, opened 2022-04-28
Add BaseException.__reduce__

fanninpm:
Addresses three items in #3611.

cc @Snowapril

### DIFF
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -526,11 +526,7 @@ impl PyBaseException {
         zelf: PyRef<Self>,
         _vm: &VirtualMachine,
     ) -> (PyTypeRef, PyTupleRef, Option<PyDictRef>) {
-        (
-            zelf.class().clone(),
-            zelf.args(),
-            zelf.as_object().dict(),
-        )
+        (zelf.class().clone(), zelf.args(), zelf.as_object().dict())
     }
 }
 
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -2,8 +2,7 @@ use self::types::{PyBaseException, PyBaseExceptionRef};
 use crate::common::lock::PyRwLock;
 use crate::{
     builtins::{
-        traceback::PyTracebackRef, PyDictRef, PyNone, PyStr, PyStrRef, PyTuple, PyTupleRef, PyType,
-        PyTypeRef,
+        traceback::PyTracebackRef, PyNone, PyStr, PyStrRef, PyTuple, PyTupleRef, PyType, PyTypeRef,
     },
     class::{PyClassImpl, StaticType},
     convert::{ToPyException, ToPyObject},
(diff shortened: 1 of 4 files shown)


Snowapril:
Looks good! This PR can be a good specimen for the other remaining works. Does "three items" mean `OSError`, `ImportError`, `BaseException`?

fanninpm:
> Does "three items" mean `OSError`, `ImportError`, `BaseException`?

Yes. I only had to add things to `BaseException`, since `ImportError` and `OSError` inherit from `BaseException`.

vm/src/exceptions.rs
```diff
@@ -519,6 +520,18 @@ impl PyBaseException {
...
+    #[pymethod(magic)]
+    fn reduce(
+        zelf: PyRef<Self>,
+        _vm: &VirtualMachine,
+    ) -> (PyTypeRef, PyTupleRef, Option<PyDictRef>) {
+        (
+            zelf.class().clone(),
+            zelf.args(),
+            zelf.as_object().dict(),
+        )
```
Snowapril: Last field must not be included in tuple if it is `None`. This leads to print empty set like this
```python
>>>>> BaseException("test").__reduce__()
(<class 'BaseException'>, ('test',), {})
```
It must be printed like this
```python
>>>>> BaseException("test").__reduce__()
(<class 'BaseException'>, ('test',))
```
It is good to refer this cpython code 😊
```c
static PyObject *
BaseException_reduce(PyBaseExceptionObject *self, PyObject *Py_UNUSED(ignored))
{
    if (self->args && self->dict)
        return PyTuple_Pack(3, Py_TYPE(self), self->args, self->dict);
    else
        return PyTuple_Pack(2, Py_TYPE(self), self->args);
}
```

Snowapril:
Unfortunately, OSError's and ImportError's reduce act different than BaseException's 😭
* BaseException(https://github.com/python/cpython/blob/3.10/Objects/exceptions.c#L132-L140)
* OSError(https://github.com/python/cpython/blob/3.10/Objects/exceptions.c#L1147-L1194)
* ImportError(https://github.com/python/cpython/blob/3.10/Objects/exceptions.c#L742-L758)

fanninpm:
> Unfortunately, OSError's and ImportError's reduce act different than BaseException's 😭

I would direct future contributors to look at the `extend_exception!` macro.

Snowapril:
Ok good 😊. Except for the issue about the empty set that I commented on, everything seems fine.

vm/src/exceptions.rs
fanninpm: Resolved in e2af440.

Snowapril (approved):

vm/src/exceptions.rs
```diff
@@ -519,6 +519,15 @@ impl PyBaseException {
         let cls = zelf.class();
         format!("{}({})", cls.name(), repr_args.iter().format(", "))
     }
+
+    #[pymethod(magic)]
+    fn reduce(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyTupleRef {
```
youknowone, line 524: If it doesn't require type checking, this may be:
```suggestion
    fn reduce(zelf: PyObjectRef, vm: &VirtualMachine) -> PyTupleRef {
```

youknowone (approved):
great, thank you! @Snowapril and thank you for the review!

vm/src/exceptions.rs
fanninpm:
```text
no method named `args` found for struct `core::PyObjectRef` in the current scope
method not found in `core::PyObjectRef`
```

PR #3668, opened 2022-04-29
Remove Interpreter::default to alert better there is no stdlib

youknowone:


coolreader18 (approved):
LGTM! I think we should probably make some sort of `InterpreterBuilder` struct at some point, that would probably make configuring this a lot easier.

youknowone:
That sounds like good idea, but not sure how it will look like because most of Settings struct came from command-line

Issue #3610, opened 2022-04-14
Add `__self__` and `__self_class__` property accessor for `super`


Snowapril:
## Feature
Add `__self__` and `__self_class__` for `PySuper`. Please refer this PR #3601 
For now, `obj` field in RustPython have both for `obj` and `obj_type` in cpython. See below link

## Python Documentation
https://github.com/python/cpython/blob/main/Objects/typeobject.c#L8755-L8772

youknowone:
@djrmarques if you are looking for a good starting point to learn how RustPython works, this is a good starting point.

PR #3683, opened 2022-05-03
Added self and self_class to PySuper

djrmarques:
Hello. 

Submitting this PR as a draft because I would like to get some feedback about  #3610. 
Do I have the right idea here or is this completely wrong? 

Thank you!

### DIFF
--- a/Lib/test/test_fcntl.py
+++ b/Lib/test/test_fcntl.py
@@ -141,7 +141,7 @@ class TestFcntl(unittest.TestCase):
         finally:
             os.close(fd)
 
-    # TODO: RUSTPYTHON, AttributeError: module 'fcntl' has no attribute 'flock'
+    # TODO: RUSTPYTHON, TypeError: 'BufferedRandom' object cannot be interpreted as an integer
     @unittest.expectedFailure
     def test_flock(self):
         # Solaris needs readable file for shared lock
@@ -157,7 +157,7 @@ class TestFcntl(unittest.TestCase):
         self.assertRaises(ValueError, fcntl.flock, -1, fcntl.LOCK_SH)
         self.assertRaises(TypeError, fcntl.flock, 'spam', fcntl.LOCK_SH)
 
-    # TODO: RUSTPYTHON, AttributeError: module 'fcntl' has no attribute 'lockf'
+    # TODO: RUSTPYTHON, TypeError: 'BufferedRandom' object cannot be interpreted as an integer
     @unittest.expectedFailure
     @unittest.skipIf(platform.system() == "AIX", "AIX returns PermissionError")
     def test_lockf_exclusive(self):
@@ -170,7 +170,7 @@ class TestFcntl(unittest.TestCase):
         fcntl.lockf(self.f, fcntl.LOCK_UN)
         self.assertEqual(p.exitcode, 0)
 
-    # TODO: RUSTPYTHON, AttributeError: module 'fcntl' has no attribute 'lockf'
+    # TODO: RUSTPYTHON, TypeError: 'BufferedRandom' object cannot be interpreted as an integer
     @unittest.expectedFailure
     @unittest.skipIf(platform.system() == "AIX", "AIX returns PermissionError")
     def test_lockf_share(self):
--- a/stdlib/src/fcntl.rs
+++ b/stdlib/src/fcntl.rs
@@ -139,4 +139,74 @@ mod fcntl {
             }
         }
     }
+
+    // XXX: at the time of writing, wasi and redox don't have the necessary constants/function
+    #[cfg(not(any(target_os = "wasi", target_os = "redox")))]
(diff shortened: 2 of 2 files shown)


djrmarques:
@youknowone any tips on this? :)

youknowone (changes requested):
Please add tests to `extra_tests/snippets/builtin_super.py` to ensure they are working same in CPython and RustPython.

Otherwise, you can find related tests from `Lib/test/test_super.py` and remove skip/unexpectedFailure decorator there. But I don't think these features tests are in the current version of test file.

youknowone:
@djrmarques did you set your git config user.email correctly?
@Snowapril could you review this patch?

vm/src/builtins/super.rs
```diff
@@ -110,10 +110,26 @@ impl PySuper {
     }
 
     #[pyproperty(magic)]
-    fn thisclass(&self) -> PyTypeRef {
+    fn __thisclass__(&self) -> PyTypeRef {
```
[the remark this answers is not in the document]
djrmarques: But the problem that I was having was that I can't have one of the functions named *self*, that's why I changed the names.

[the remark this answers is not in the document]
youknowone: Thank you for sharing your motivation. Now I understand what's going on.
But still, if you had problem with `__self__`, why did you change this function?

Use `magic` as much as possible. But do not use `magic` if it doesn't fit. Search for `fn __set__` and `fn __name__` to check other similar cases as examples.

vm/src/builtins/super.rs
```diff
@@ -110,10 +110,26 @@ impl PySuper {
...
         self.typ.clone()
     }
 
+    #[pyproperty(magic)]
+    fn __self_class__(&self) -> Option<PyTypeRef> {
```
Snowapril, line 118: As youknowone said, `magic` in `#[pyproperty]` attach two underscores at the left-right sides automatically. This code leads to the below result.
```python
>>>>> super(list).__self_class__
AttributeError: 'super' object has no attribute '__self_class__'. Did you mean: '____self_class____'?
```

You can use `#[pyproperty(name = "__self__")]` instead of `magic`. 
The reserved keyword `self` could be replaced with `zelf`.

vm/src/builtins/super.rs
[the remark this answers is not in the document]
djrmarques: Thank you for your feedback, will keep working on this :)

vm/src/builtins/super.rs
```diff
@@ -110,10 +110,26 @@ impl PySuper {
...
+        match &self.obj {
+            Some((_, obj_type)) => Some(obj_type.clone()),
+            _ => None,
+        }
```
Snowapril, line 122: This code is also considerable.
```rust
self.obj.as_ref().map(|(_, obj_type)| obj_type.clone())
```
Same code can be used for `__self__`

youknowone: I prefer this form:
```suggestion
        Some(self.obj.as_ref()?.1.clone())
```

vm/src/builtins/super.rs
```diff
@@ -114,6 +114,16 @@ impl PySuper {
         self.typ.clone()
     }
 
+    #[pyproperty(magic)]
+    fn selfclass(&self) -> Option<PyTypeRef> {
+        Some(self.obj.as_ref()?.1.clone())
+    }
+
+    #[pyproperty(name = "__self__")]
+    fn zelf(&self) -> Option<PyObjectRef> {
```
youknowone, line 123:
```suggestion
    #[pyproperty]
    fn __self__(&self) -> Option<PyObjectRef> {
```
For this case, naming it `__self__` looks better because `zelf` is not a thing in python (though we use it for argument names)

youknowone (changes requested):
The lint failure doesn't look related to this PR. (#3690)

youknowone:
Like most of upstreams, we only merge from upstream, so expecting every PR doesn't include any merge commit inside.
This is a simple PR, so we can squash it for this one. But learning rebase will be helpful if you want to make more PRs to RustPython or any other projects

Check this document if you are interested in: https://docs.gitlab.com/ee/topics/git/git_rebase.html

youknowone (approved):
Looks good now.
Please change the status 'Ready for review` once you feel it's done

djrmarques:
Thanks for the support on this. Probably going to just add some tests and possibly also figure out the rebase thing and then change it to ready

djrmarques:
Will make another PR for the same issue, this time without any merge commits and with relevant tests added. I am not even sure why this got closed, possibly because I did a force push on my fork branch?

youknowone:
it seems you pushed the same commit of upstream/master to your patch branch. In that case, github automatically closes the PR. (I don't like the behavior)

PR #3685, opened 2022-05-03
Changed pyclass name of PyArray and PyArrayIter #3612

### LINKED ISSUE #3612: Array iterator name is different from CPython
## Feature
The name of `array.array` iterator has to be `array_iterator` but `arrayiterator` in RustPython.

## Python Documentation
https://github.com/python/cpython/blob/3.8/Modules/arraymodule.c#L2993

rebunto:
Resolves #3612 

The issue is

> The name of array.array iterator has to be array_iterator but arrayiterator in RustPython.

but it is "The name of array.array iterator has to be **arrayiterator** but **array_iterator** in RustPython", right?

and I also changed also PyArray's name, 
because it looks different from [CPython's](https://github.com/python/cpython/blob/3.8/Modules/arraymodule.c#L2838).

### DIFF
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -1280,7 +1280,7 @@ mod array {
     }
 
     #[pyattr]
-    #[pyclass(name = "array_iterator")]
+    #[pyclass(name = "arrayiterator")]
     #[derive(Debug, PyPayload)]
     pub struct PyArrayIter {
         position: AtomicUsize,
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -581,7 +581,7 @@ mod array {
     }
 
     #[pyattr]
-    #[pyclass(name = "array.array")]
+    #[pyclass(name = "array")]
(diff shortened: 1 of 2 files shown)


stdlib/src/array.rs
```diff
@@ -581,7 +581,7 @@ mod array {
     }
 
     #[pyattr]
-    #[pyclass(name = "array")]
+    #[pyclass(name = "array.array")]
```
youknowone, line 584: we typically put only `array` here. the other array came from the module name.

youknowone (commented):
Could you please add a related test to `extra_tests/snippets/stdlib_array.py` to ensure this behavior is now same for CPython and RustPython?

youknowone:
please check your git config user.email matches to your account. It is referring different account.

stdlib/src/array.rs
rebunto: Understood! I fixed it and added tests.

youknowone (approved):

PR #3686, opened 2022-05-03
Add flock and lockf to fcntl module

fanninpm:


stdlib/src/fcntl.rs
```diff
@@ -2,6 +2,8 @@ pub(crate) use fcntl::make_module;
 
 #[pymodule]
 mod fcntl {
+    use std::mem;
+
```
youknowone, line 6:
```suggestion
```

fanninpm: Accepted with revisions in 2c8588d.

stdlib/src/fcntl.rs
```diff
@@ -8,6 +8,7 @@ mod fcntl {
         stdlib::{io, os},
         PyResult, VirtualMachine,
     };
+    use std::mem;
```
youknowone, line 11: unused import warning here. To avoid cfg, I left another suggestion

stdlib/src/fcntl.rs
```diff
@@ -139,4 +140,74 @@ mod fcntl {
...
+    #[pyfunction]
+    fn lockf(
+        fd: i32,
+        cmd: i32,
+        len: OptionalArg<PyIntRef>,
+        start: OptionalArg<PyIntRef>,
+        whence: OptionalArg<i32>,
+        vm: &VirtualMachine,
+    ) -> PyResult {
+        let mut l: libc::flock = unsafe { mem::zeroed() };
```
youknowone, line 167:
```suggestion
        let mut l: libc::flock = unsafe { std::mem::zeroed() };
```

youknowone (approved):

Issue #3693, opened 2022-05-08
optional `key` argument for `_bisect` functions


youknowone:
## Feature

`key` argument is added in CPython 3.10

## Python Documentation

https://docs.python.org/3/library/bisect.html

supreetsingh10:
I would like to work on this. I am going through the code base, will reach out on gitter if I have questions.

youknowone:
@supreetsingh10 This issue is already taken by @rebunto in #3695

supreetsingh10:
> 

No problem, I will study the codebase and pick another issue soon. I will contact you on Gitter, if I need some guidance. Thanks

PR #3695, opened 2022-05-09
Optional key argument for bisect functions

rebunto:
#3693
I modified `stdlib\src\bisect.rs`, 
and updated `Lib/test/test_bisect.py` and `Lib/bisect.py` to the version 3.10.

https://github.com/python/cpython/blob/3.10/Lib/bisect.py
https://github.com/python/cpython/blob/3.10/Lib/test/test_bisect.py

### DIFF
--- a/stdlib/src/bisect.rs
+++ b/stdlib/src/bisect.rs
@@ -14,6 +14,8 @@ mod _bisect {
         lo: OptionalArg<PyObjectRef>,
         #[pyarg(any, optional)]
         hi: OptionalArg<PyObjectRef>,
+        #[pyarg(named, default)]
+        key: Option<PyObjectRef>,
     }
 
     // Handles objects that implement __index__ and makes sure index fits in needed isize.
@@ -66,7 +68,7 @@ mod _bisect {
     #[inline]
     #[pyfunction]
     fn bisect_left(
-        BisectArgs { a, x, lo, hi }: BisectArgs,
+        BisectArgs { a, x, lo, hi, key }: BisectArgs,
         vm: &VirtualMachine,
     ) -> PyResult<usize> {
         let (mut lo, mut hi) = as_usize(lo, hi, a.length(vm)?, vm)?;
@@ -74,9 +76,13 @@ mod _bisect {
         while lo < hi {
             // Handles issue 13496.
             let mid = (lo + hi) / 2;
-            if a.get_item(&mid, vm)?
-                .rich_compare_bool(&x, PyComparisonOp::Lt, vm)?
-            {
+            let a_mid = a.get_item(&mid, vm)?;
+            let comp = if let Some(ref key) = key {
+                vm.invoke(key, (a_mid,))?
+            } else {
+                a_mid
+            };
+            if comp.rich_compare_bool(&x, PyComparisonOp::Lt, vm)? {
                 lo = mid + 1;
             } else {
                 hi = mid;
@@ -96,7 +102,7 @@ mod _bisect {
     #[inline]
     #[pyfunction]
     fn bisect_right(
-        BisectArgs { a, x, lo, hi }: BisectArgs,
+        BisectArgs { a, x, lo, hi, key }: BisectArgs,
         vm: &VirtualMachine,
     ) -> PyResult<usize> {
         let (mut lo, mut hi) = as_usize(lo, hi, a.length(vm)?, vm)?;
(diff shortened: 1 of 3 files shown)


stdlib/src/bisect.rs
```diff
@@ -66,20 +68,35 @@ mod _bisect {
...
+
+        } else {
+            while lo < hi {
+                // Handles issue 13496.
+                let mid = (lo + hi) / 2;
+                if a.get_item(mid, vm)?.rich_compare_bool(&x, Lt, vm)? {
+                    lo = mid + 1;
+                } else {
+                    hi = mid;
+                }
```
Snowapril, line 96: It seems `bisect_left` and `bisect_right` implementation share some duplicated codes except the order of codes `lo = mid = 1;` and `hi = mid;`. How do you think?

youknowone: I think that is regarded as repeated by coincidence. we named everything binary search, but they have a few variants. we don't have much gain by combining them.

youknowone: We don't have much gain by combining them.

stdlib/src/bisect.rs
```diff
@@ -94,20 +111,36 @@ mod _bisect {
...
+                let mid = (lo + hi) / 2;
+                if x.rich_compare_bool(&*a.get_item(mid, vm)?, Lt, vm)? {
+                    hi = mid;
+                } else {
+                    lo = mid + 1;
+                }
             }
         }
+
+
```
youknowone, line 143:
```suggestion
```

stdlib/src/bisect.rs
```diff
@@ -66,20 +68,35 @@ mod _bisect {
...
+                // Handles issue 13496.
+                let mid = (lo + hi) / 2;
+                if a.get_item(mid, vm)?.rich_compare_bool(&x, Lt, vm)? {
+                    lo = mid + 1;
+                } else {
+                    hi = mid;
+                }
             }
         }
+
```
youknowone, line 99:
```suggestion
```

stdlib/src/bisect.rs
```diff
@@ -94,20 +111,36 @@ mod _bisect {
...
+        } else {
+            while lo < hi {
+                // Handles issue 13496.
+                let mid = (lo + hi) / 2;
+                if x.rich_compare_bool(&*a.get_item(mid, vm)?, Lt, vm)? {
+                    hi = mid;
+                } else {
+                    lo = mid + 1;
+                }
```
youknowone, line 139: this part can be merged with `Some(key)` part. 
they share everything except for key invoke.

none checking is very cheap comparing to `get_item`.

youknowone (changes requested):
Nice work for updating library! We want to know which version is used for the files by looking `git log`. Please fix the commit messages include the specific CPython version you brought the library from. (using `reword` of `git rebase -i` is helpful)

stdlib/src/bisect.rs
rebunto: Thank you for your review!
In my opinion, the duplication is a coincidence.
I think solving value `i` where all `e` in `a[:i]` have `e < x` and solving value `i` where all `e` in `a[:i]` have `e <= x` are different tasks.
Generally speaking, unnecessary dependency may increase maintenance costs.
So, I thought it is better to leave the duplication, but I'm not sure😖

youknowone:
would you try [rebase](https://docs.gitlab.com/ee/topics/git/git_rebase.html) instead of merge? when we use git, only upstream(RustPython) make merge commit. You can remove your merge commits by rebasing on `upstream/main`

rebunto:
I resolved the conflict on github, but it was merged into main. 
For removing my merge commits, do I need to pull this repository?
I usually work with my fork, but I should not rebase there, right?
Sorry for my lack of knowledge.

youknowone:
Don't worry! I will show you my way. I am a CLI user, so if you are using GUI tool, look for your way based on this comment.

when I tried `git remote -v`, it looks like:
```
origin  git@github.com:youknowone/RustPython.git (fetch)
origin  git@github.com:youknowone/RustPython.git (push)
upstream        https://github.com/RustPython/RustPython.git (fetch)
upstream        https://github.com/RustPython/RustPython.git (push)
```
You can make your remote looks similar by using `git remote rename` and `git remote add` (sometimes, also `git remote rm`)

Then I usually do:
```sh
$ git fetch upstream  # to fetch recent updates
$ git rebase --interactive upstream/main  # to remove useless commits. you can omit `--interactive` if you don't need to remove any commit
```

This is enough for most of cases. Tell me if you meet more problems.

rebunto:
Thank you for the detailed explanation!

I did:
```
git rebase --interactive upstream/main
```

It opened `git-rebase-todo`.
I edited it as below:
```
pick 9900ab5a1 Try to invoke lalrpop from parser build script
drop fc515a130 update test_bisect.py to support CPython 3.10.4
drop 9cfb59fe9 update bisect.py and bisect.rs
pick 46a93d790 Added self and self_class to super (#3691)
pick 871992aa1 Add document for wasi
drop cfeef3c1e clean bisect.rs
drop b0c39eefd update bisect.rs
```
Is it OK to save this?
and then should I do just `git push`?

youknowone:
you are trying to doing them in opposite way.
pick your changes and drop others' changes. check rebase documents how it works.

I will do like this:
```
pick fc515a130 update test_bisect.py to support CPython 3.10.4
pick 9cfb59fe9 update bisect.py and bisect.rs
pick cfeef3c1e clean bisect.rs
pick b0c39eefd update bisect.rs
```

and yes, push force them after rebase

youknowone (approved):

youknowone:
resolved #3693

PR #3696, opened 2022-05-11
Remove DictInner::next_new_entry_idx and IndexEntry as a shallow wrapper of i64

youknowone:


vm/src/dictdatatype.rs
```diff
@@ -550,22 +540,34 @@ impl<T: Clone> Dict<T> {
...
+                                // DUMMY and FREE are already handled above.
+                                // i is always valid and entry always exists.
+                                let i = idx
+                                    .index()
+                                    .unwrap_or_else(|| std::hint::unreachable_unchecked());
+                                inner
+                                    .entries
+                                    .get_unchecked(i)
+                                    .as_ref()
+                                    .unwrap_or_else(|| std::hint::unreachable_unchecked())
```
coolreader18, line 568: Honestly, I'd rather just have just `idx.0` or `idx.as_i64()` or something, the risk (which I know, this is totally fine) of unreachable_unchecked UB seems totally unnecessary here. alternatively, `index()` could return a `Result<usize, bool>` ? basically the same tristate enum as before but passed around less, I guess.

vm/src/dictdatatype.rs
```diff
@@ -644,14 +646,14 @@ impl<T: Clone> Dict<T> {
...
-            .iter_mut()
-            .enumerate()
-            .rev()
-            .find_map(|(i, entry)| entry.take().map(|e| (i, e)))?;
+        let entry = loop {
+            let entry = inner.entries.pop()?;
+            if let Some(entry) = entry {
+                break entry;
+            }
+        };
```
coolreader18, line 654: Haven't touched dict in a while so I don't remember exactly how it all works, but this is different behavior from before. I suppose there are probably thousands of invocations of this while running the test suite, so if it was an issue it would show up, but does this leave dangling indices in the indices vector or something? I'm not seeing obvious changes where that change would be handled or something so just wondering.

vm/src/dictdatatype.rs
youknowone: unreachable_unchecked is just same unsafe as `idx.0 as usize`. But `idx.0 as usize` hides it can be unsafe value without proper preprocess, while `.index()` shows we are doing unsafe operation by trusting the preprocess. Even more, `unreachable_unchecked` doesn't work without debug mode. This is safer than naive casting.

vm/src/dictdatatype.rs
youknowone: `next_new_entry_idx` was a redundant value of `len()` before. this function was the only exception.
So using `next_new_entry_idx` outside of this function is same as using the `len()`.

I thought that  means I only need to concern about this function's behavior. That doesn't look have much chance to be changed due to the lock of the first line of the function.

youknowone: @eldpswp99 do you have any advice about this?

eldpswp99: I used `next_new_entry_idx` to maintain the dict deletion system - deleting makes index element to DUMMY on `pop_back`. 
And since `should_resize` function and `resize` function will handle about danling indices, I think this change would remain the same behaviour.

vm/src/dictdatatype.rs
coolreader18: Well, you can at least use `option.unwrap_unchecked()` I guess.

youknowone: Oh, I didn't know that's stablized

vm/src/dictdatatype.rs
```diff
@@ -204,16 +194,11 @@ impl<T> DictInner<T> {
...
-        let entry_index = self.next_new_entry_idx;
-        if self.entries.len() == entry_index {
-            self.entries.push(Some(entry));
-        } else {
-            self.entries[entry_index] = Some(entry);
-        }
-        self.indices[index] = entry_index as i64;
+        let entry_index = self.entries.len();
+        self.entries.push(Some(entry));
+        self.indices[index] = unsafe { IndexEntry::from_index_unchecked(entry_index) };
```
coolreader18, line 199: This should either take a IndexEntry or the function should be unsafe probably

coolreader18 (commented):
probably need to more extensively document the unsafety invariants and whatnot at some point but lgtm for now

coolreader18 (approved):

youknowone:
@coolreader18 I added additional comments for them why they are safe. please let me know if there is any unclear point. including anything out of this PR

PR #3697, opened 2022-05-11
drain with open range

youknowone:


vm/src/frame.rs
```diff
@@ -1804,7 +1804,7 @@ impl ExecutingFrame<'_> {
 
     fn pop_multiple(&mut self, count: usize) -> crate::common::boxvec::Drain<PyObjectRef> {
         let stack_len = self.state.stack.len();
-        self.state.stack.drain(stack_len - count..stack_len)
+        self.state.stack.drain(stack_len - count..)
```
youknowone, line 1807: this change enhance pystone behchmark 0.8%

coolreader18 (approved):

PR #3698, opened 2022-05-11
Improve comparison ops performance

youknowone:
8% enhancement for `microbenchmarks/cmp.py`.

Because most of the running time is consumed by `execute_instruction`, `PyMapping::_subscript` and a few more vm fundamentals, the comparison ops itself is a lot more improved.

The major part of improvement came from new `PyComparisonOp` representation - probably because it removed unnecessary branches.
A minor part of improvement came from spliting compare instruction and test instructions + inlining `rich_compare`.


coolreader18 (approved):
LGTM! If you wanna go all the way, it looks like newer versions of cpython's bytecode separate out `IsOperator { invert: bool }` and `InOperator { invert: bool }` as well. Oh, and then it seems `ExceptionMatch` becomes its own `JumpIfNotExceptionMatch { target: Label }`

coolreader18:
Oh, also, if we don't wanna have a newtype for `PyComparisonOp` we could just make them the same discriminant values and then a `match a { LessThan => Lt, ... }` should still be 0 cost.

youknowone:
CPython is changing its bytecode a lot these days because they decided to make tiered optimization. whatever we do more, it will not match soon.

Still `IsOperator` and `InOperator` with `invert` sounds very reasonable. Creating a good first issue about them will make chance to new contributors to access our bytecode processing.

PR #3700, opened 2022-05-12
PyStrInterned for PyAttributes/Constants

youknowone:
2.5% performance gain for pystone

- first bit of `RefCount` is now used to mark leaked objects
- `identifier!()` macro to refer static identifier
- `PyStrInterned` impl `DictKey`


youknowone:
Do you mean changing `interned` macro to `PyIdentifier`? This is making sense.
Or something like `_Py_IDENTIFIER` to define them? This is not working because we don't have global variables. (Any clever idea?)

youknowone:
@coolreader18 I am sorry, but I think I wasn't getting what you meant correctly. What do you mean by "to call these identifiers"?

coolreader18:
Oh, just in the names, not like change to be the same functionality as the `_Py_IDENTIFIER`

youknowone:
not easy to solve static type problems

youknowone:
@coolreader18 do you have a good idea?

youknowone:
@coolreader18 does `identifier!` look good? `py_identifier!`?

youknowone:
@coolreader18 done

PR #3715, opened 2022-05-17
Fixed int.real, int.numerator etc. returns wrong value when it's int type.

### LINKED ISSUE #3687: int.real returns self for `int` type without subclassing
when a is int,
```python
assert a.real is a
```

when a is subclass of int,
```python
assert a.real is not a
assert type(a.real) is int
```

killme2008:
Try to fix #3687

### DIFF
--- a/vm/src/builtins/int.rs
+++ b/vm/src/builtins/int.rs
@@ -562,17 +562,17 @@ impl PyInt {
 
     #[pymethod(magic)]
     fn trunc(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyRef<Self> {
-        Self::long_long(zelf, vm)
+        Self::clone_if_subclass(zelf, vm)
     }
 
     #[pymethod(magic)]
     fn floor(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyRef<Self> {
-        Self::long_long(zelf, vm)
+        Self::clone_if_subclass(zelf, vm)
     }
 
     #[pymethod(magic)]
     fn ceil(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyRef<Self> {
-        Self::long_long(zelf, vm)
+        Self::clone_if_subclass(zelf, vm)
     }
 
     #[pymethod(magic)]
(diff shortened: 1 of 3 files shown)


Snowapril:
Test issue3687 must be in extra_tests/snippets/builtin_int.py

killme2008:
@Snowapril Got it, any other comments?

Snowapril (approved):
Looks good to me 😊

vm/src/builtins/int.rs
```diff
@@ -688,10 +688,18 @@ impl PyInt {
...
     }
 
+    #[inline]
+    fn long_long(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyRef<Self> {
+        if zelf.class().is(&vm.ctx.types.int_type) {
+            return zelf;
+        }
+
+        vm.ctx.new_bigint(&zelf.value)
+    }
```
Snowapril, line 698: Maybe it is trivial, below one seems better
```rust
if zelf.class().is(&vm.ctx.types.int_type) {
    zelf
} else {
    vm.ctx.new_bigint(&zelf.value)
}
```

vm/src/builtins/int.rs
```diff
@@ -688,10 +688,18 @@ impl PyInt {
         Ok(bytes.into())
     }
 
+    #[inline]
+    fn long_long(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyRef<Self> {
+        if zelf.class().is(&vm.ctx.types.int_type) {
+            return zelf;
+        }
+
+        vm.ctx.new_bigint(&zelf.value)
```
Snowapril, line 697:
```suggestion
        if zelf.class().is(&vm.ctx.types.int_type) {
            zelf
        } else {
            vm.ctx.new_bigint(&zelf.value)
        }
```

youknowone (changes requested):
looks good! thank you!
I left a minor comment about the name. could you please check it?

@Snowapril thank you for the review!

vm/src/builtins/int.rs
```diff
@@ -688,10 +688,18 @@ impl PyInt {
         Ok(bytes.into())
     }
 
+    #[inline]
+    fn long_long(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyRef<Self> {
```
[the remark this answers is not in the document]
killme2008: Yep, `clone_if_subclass` is a better name. `long_long` is from cpython.

youknowone (approved):
looks good, thank you!

PR #3716, opened 2022-05-17
PyResult<T, E> instead of PyResult<T>

youknowone:


vm/src/object/ext.rs
```diff
@@ -27,7 +27,7 @@ Basically reference counting, but then done by rust.
 /// Use this type for functions which return a python object or an exception.
 /// Both the python object and the python exception are `PyObjectRef` types
 /// since exceptions are also python objects.
-pub type PyResult<T = PyObjectRef> = Result<T, PyBaseExceptionRef>; // A valid value, or an exception
+pub type PyResult<T = PyObjectRef, E = PyBaseExceptionRef> = Result<T, E>; // A valid value, or an exception
```
coolreader18, line 30: wait why is this necessary? Why not just use `Result<>`, the main reason type aliases would have a default E type param is because if you bring it into scope then you can't call the std Result by its name you'd have to do std::result::Result, but that's not an issue because PyResult doesn't shadow std Result

youknowone: Oh, you are right.

PR #3724, opened 2022-05-19
ensure bytecode::Instruction never grows more than 8bytes

youknowone:


coolreader18:
Not that I'm against the dependency but you can also do the same thing by just `const _: () = assert!(size_of::<Instruction>() == 8);`

youknowone:
Do you want to replace it not to use the crate? It is already used by core. It will not affect build time or binary size that much.

coolreader18 (approved):
nah, it's fine either way actually

PR #3728, opened 2022-05-20
Add `object.__sizeof__`

rng-dynamics:
I had a look at `object.__sizeof__`. @moreal opened a pull request (#3421) on it last year, but it never got finished.

CPython uses `tp_basicsize`, `tp_itemsize`, and `ob_size` in `__sizeof__`. The intended use of `tp_basicsize` and `tp_itemsize` can be seen, e.g., at  https://docs.python.org/3/extending/newtypes_tutorial.html#defining-new-types. Since new types are implemented different in RustPython we don't need `tp_itemsize` (and `ob_size`).

If merged, this should be the last point for #2898.

Can you give me some feedback? Thanks!

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -210,6 +210,7 @@ fn generate_class_def(
         Some(v) => quote!(Some(#v) ),
         None => quote!(None),
     };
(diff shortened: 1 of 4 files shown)


youknowone (approved):

youknowone:
thank you! it looks great!

PR #3731, opened 2022-05-21
Add weak proxy str

CHOUMnote:
After a  weakReference and str() call, the two values were different in comparison, resulting in an error
I found it, "weakproxy object" is not have "\_\_str__()" so calling by "parent object" in "\_\_str__()" 
So I added in "weakproxy. rs"  made it and "str()"method is work normally

### DIFF
--- a/Lib/test/test_deque.py
+++ b/Lib/test/test_deque.py
@@ -889,8 +889,6 @@ class TestSubclass(unittest.TestCase):
             for d in DequeWithBadIter('abc'), DequeWithBadIter('abc', 2):
                 self.assertRaises(TypeError, pickle.dumps, d, proto)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_weakref(self):
         d = deque('gallahad')
         p = weakref.proxy(d)
--- a/vm/src/builtins/weakproxy.rs
+++ b/vm/src/builtins/weakproxy.rs
@@ -70,6 +70,16 @@ impl PyWeakProxy {
         })?;
         obj.get_attr(attr_name, vm)
     }
+    #[pymethod(magic)]
+    fn str(&self, vm: &VirtualMachine) -> PyResult<PyStrRef> {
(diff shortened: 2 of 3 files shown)


Lib/test/test_deque.py
```diff
@@ -889,8 +889,7 @@ def test_pickle_recursive(self):
             for d in DequeWithBadIter('abc'), DequeWithBadIter('abc', 2):
                 self.assertRaises(TypeError, pickle.dumps, d, proto)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+
```
youknowone, line 892:
```suggestion
```

vm/src/builtins/weakproxy.rs
```diff
@@ -1,6 +1,7 @@
 use super::{PyStrRef, PyTypeRef, PyWeak};
 use crate::{
     class::PyClassImpl,
+    convert::ToPyObject,
```
youknowone, line 4:
```suggestion
```

unused import

vm/src/builtins/weakproxy.rs
```diff
@@ -70,6 +71,16 @@ impl PyWeakProxy {
...
         obj.get_attr(attr_name, vm)
     }
+    #[pymethod(magic)]
+    fn str(&self, vm: &VirtualMachine) -> PyResult<PyStrRef> {
+        match self.weak.upgrade() {
+            Some(obj) => obj.str(vm),
+            None => Err(vm.new_exception_msg(
+                vm.ctx.exceptions.reference_error.clone(),
+                "weakly-referenced object no longer exists".to_owned(),
+            )),
```
youknowone, line 81: this part is an exact repeat of error in `getattr()`.

please create a function `new_reference_error()` not to manually repeat the exception creation.

youknowone (changes requested):

youknowone (approved):

PR #3735, opened 2022-05-21
Zlib ver

kangwhon:
add ZLIB_RUNTIME_VERSION, ZLIB_VERSION function

### DIFF
--- a/Lib/test/test_zlib.py
+++ b/Lib/test/test_zlib.py
@@ -51,7 +51,6 @@ skip_on_s390x = unittest.skipIf(hasattr(os, 'uname') and os.uname().machine == '
 
 class VersionTestCase(unittest.TestCase):
 
-
     def test_library_version(self):
         # Test that the major version of the actual library in use matches the
         # major version that we were compiled against. We can't guarantee that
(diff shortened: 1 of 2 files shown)


Lib/test/test_zlib.py
```diff
@@ -19,8 +19,7 @@
 
 class VersionTestCase(unittest.TestCase):
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+
```
youknowone, line 22:
```suggestion
```

stdlib/src/zlib.rs
```diff
@@ -53,6 +54,26 @@ mod zlib {
...
+    #[cfg(feature = "zlib")]
+    #[pyattr(name = "ZLIB_VERSION", once)]
+    fn zlib_version(_vm: &VirtualMachine) -> String {
+        unsafe {
+            CStr::from_ptr(libz_sys::zlibVersion())
+                .to_string_lossy()
+                .into_owned()
+        }
+    }
+
```
youknowone, line 76:
```suggestion
```

By looking [zlib.h](https://github.com/madler/zlib/blob/master/zlib.h), `ZLIB_VERSION` is the version from header and `zlibVersion` is the version from library.

youknowone (changes requested):

youknowone:
we now don't link zlib anymore

PR #3739, opened 2022-05-22
Fixed `bytes_from_object`

Lithium07z:
Issue: `int.from_bytes("", 'big')` It have to return `TypeError` but it did return `AssertionError: TypeError not raised by from_bytes` so fixed it `bytes_from_object` function that it returns Error


youknowone (approved):
thank you for contributing!

PR #3740, opened 2022-05-22
add __delitem__() funtion

CHOUMnote:
when call "__delitem__()" error message did not match, So add "__delitem__()" function and matched it with cPython

### DIFF
--- a/extra_tests/snippets/builtin_memoryview.py
+++ b/extra_tests/snippets/builtin_memoryview.py
@@ -69,3 +69,17 @@ def test_resizable():
     b.append(7)
 
 test_resizable()
+
+def test_delitem():
+    a = b'abc'
+    b = memoryview(a)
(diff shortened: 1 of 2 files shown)


fanninpm:
Interestingly, I can't find this functionality covered anywhere within CPython's unit test suite. In that case, @youknowone will probably want you to add a test to the `extra_tests` folder instead.

youknowone:
@fanninpm exactly, thank you!

@CHOUMnote could you add test to `extra_tests/snippets/builtin_memoryview.py`?

youknowone:
@fanninpm there was a test but it only tests if it is TypeError. So it was undistinguishable to missing method case. https://github.com/RustPython/RustPython/blob/main/Lib/test/test_memoryview.py#L127-L134

youknowone (approved):

youknowone:
thank you for contributing!

PR #3746, opened 2022-05-25
Slot inheritance

youknowone:


youknowone:
Long rally including #6622 replaced this

PR #3750, opened 2022-05-27
use static reference for AsMapping / AsSequence

youknowone:
3% from pystone

3.5% from fannkuch
```
+-----------+-----------------+------------------------+------------------------+
| Benchmark | rustpython_orig | mapping                | mapping+sequence        |
+===========+=================+========================+========================+
| fannkuch  | 3.79 sec        | 3.67 sec: 1.03x faster | 3.65 sec: 1.04x faster |
+-----------+-----------------+------------------------+------------------------+
```

```
+-----------+---------+-------------------------+
| Benchmark | cpython | rustpython-final        |
+===========+=========+=========================+
| fannkuch  | 354 ms  | 3.64 sec: 10.26x slower |
+-----------+---------+-------------------------+
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1518,7 +1518,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn",
- "syn-ext",
+ "syn-ext 0.3.1",
 ]
 
 [[package]]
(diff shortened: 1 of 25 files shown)


youknowone:
@Snowapril would you review this PR?

Snowapril (approved):
It seems better to use static ref than before 😊

PR #3753, opened 2022-05-27
Make the correctness of the bytecode in PyCode a safety invariant

coolreader18:
A cherrypick of an olllld WIP branch. Sep 2020! Figured if we're gonna be eliding bounds checks in Frame now I may as well bring this back.

### DIFF
--- a/compiler/codegen/src/compile.rs
+++ b/compiler/codegen/src/compile.rs
@@ -16,8 +16,8 @@ use num_complex::Complex64;
 use num_traits::ToPrimitive;
 use rustpython_ast as ast;
 use rustpython_compiler_core::{
-    self as bytecode, Arg as OpArgMarker, CodeObject, ConstantData, Instruction, Location, NameIdx,
-    OpArg, OpArgType,
+    self as bytecode, Arg as OpArgMarker, CodeObject, CodeObjectInner, ConstantData, Instruction,
+    Location, NameIdx, OpArg, OpArgType,
 };
 use std::borrow::Cow;
 
@@ -1219,7 +1219,7 @@ impl Compiler {
         self.store_name(name)
     }
 
-    fn build_closure(&mut self, code: &CodeObject) -> bool {
+    fn build_closure(&mut self, code: &CodeObjectInner) -> bool {
         if code.freevars.is_empty() {
             return false;
         }
--- a/compiler/codegen/src/ir.rs
+++ b/compiler/codegen/src/ir.rs
@@ -2,8 +2,8 @@ use std::ops;
 
 use crate::IndexSet;
 use rustpython_compiler_core::{
-    CodeFlags, CodeObject, CodeUnit, ConstantData, InstrDisplayContext, Instruction, Label,
-    Location, OpArg,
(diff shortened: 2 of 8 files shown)


youknowone:
is there a way to create invalid code object?

coolreader18:
Sure, assuming we fully modify Frame to assume that bytecode is correct you can do unsafe-less UB in pure rust code from just passing `CodeObject { instructions: Box::new([Instruction::Pop]), ... }` to a vm eval function. This separates `CodeObject` as a type-to-build-up-in-the-compiler from `CodeObject` as a type-to-consume-in-the-vm.

youknowone:
For now, only valid code seemed to be handled. is there a plan to support invalid code object? otherwise, I think we can keep the name CodeObject for valid objects.
I don't have a concrete idea, but we don't call a valid utf8 string type like `ValidString`, but it is just `String`. The responsibility to make sure the data valid is up to its creation function(checked) or the user(unchecked).
For the same point of view, If CodeObject requires validation, marking the creation functions as `unsafe` would be enough, instead of creating a wrapper type ValidCodeObject.

coolreader18:
Maybe I'll rename what's now CodeObject to CodeObjectInner or something then, and ValidCodeObject to just CodeObject. Cause it's really bad ux to have a like 20-arg constructor, and also wouldn't be able to stop them from mutating the inside without changing everything to accessor methods and that would be a whole big code change.

youknowone (approved):
While this is looking good enough, I have another suggestion for the relationship of the types by following MaybeUninit convention.

Currently CodeObject accesses its data via Deref. But it a safe and normal state while CodeObjectInner is not.
I think we can keep CodeObject as before, and add another wrapper `CodeObjectBuilder` for building object stage.
```rust
struct CodeObjectBuilder {
    code: CodeObject
}
```
Then adding any editing methods to CodeObjectBuilder will not affect to CodeObject because we don't need to have Deref anymore.

coolreader18:
I know, but I don't want to add any editing methods. If CodeObject is both the direct container of all the attributes (instructions, locations, etc) AND the valid-code-object type, then those attributes can't be public, and adding accessors would involve changing a bunch of code. Having a CodeObjectBuilder that has mutable accessors would similarly involve changing a lot of code, and code objects aren't even mutated, really - they're just constructed but with a ton of arguments.

youknowone:
I also don't think we actually need to implement any actual builder (unless we need it)
By naming them inside-out, we will use CodeObject with the most simple layout without having inner and Deref.
And maybe CodeObjectOuter or CodeObjectSomewhat as unsafe wrapper.
I know this is not an actual advantage. It is mostly a matter of feeling. That's why I thought this is good enough.

compiler/codegen/src/compile.rs
```diff
@@ -1219,7 +1219,7 @@ impl Compiler {
         self.store_name(name)
     }
 
-    fn build_closure(&mut self, code: &CodeObject) -> bool {
+    fn build_closure(&mut self, code: &CodeObjectInner) -> bool {
```
youknowone, line 1222: is this getting CodeObjectInner instead of CodeObject?

youknowone (commented):
I left another suggestion of implementing this purpose.

https://github.com/RustPython/RustPython/pull/4547/files

PR #3767, opened 2022-06-07
Upgrade result-like

coolreader18:


youknowone (approved):

Cargo.lock
```diff
@@ -7,10 +7,6 @@ name = "Inflector"
 version = "0.11.4"
```
youknowone, line 7: `is-macro` is still using Inflector

coolreader18:
Oh, great, miri's enforcing strict provenance now. I'll look into that, mayyybe actually I might not have the time but I'll try

coolreader18:
Ah shoot AtomicCell is unsound under strict provenance

PR #3769, opened 2022-06-10
wrap_index without abs

youknowone:
used by #3755

@coolreader18 I am worrying if you had a reason to choose that form. Could you check it?


DimitrisJim (approved):
this lgtm, not sure about why taking the absolute value was used initially.

qingshi163 (approved):
LGTM

PR #3771, opened 2022-06-11
fix(binascii): a2b_base64() raises on non padded input

isac322:


youknowone (approved):
thank you for contributing!

PR #3776, opened 2022-06-11
feat(sys): add `sys.is_finalizing()`

isac322:
Added `sys.is_finalizing()` but could not enable test_sys::SysModuleTest.test_is_finalizing. RustPython have to implement module destruction to enable the TC.


youknowone (approved):
looks great, thank you for contributing!

PR #3777, opened 2022-06-11
add Binascii crc hqx

yangbeom:
add fn binascii.crc_hqx

test result
```
/Users/yangbeom/github/RustPython` ❯❯❯ cargo run --release Lib/test/test_binascii.py
    Finished release [optimized] target(s) in 0.08s
     Running `target/release/rustpython Lib/test/test_binascii.py`
.xx...x.xxxxxxxxxx.xx...x.xxxxxxxxxx.xx...x.xxxxxxxxxx.xx...x.xxxxxxxxxx
Ran 72 tests in 0.015s

OK (expected failures=52)
```


stdlib/src/binascii.rs
```diff
@@ -87,6 +87,47 @@ mod decl {
...
+        let mut crc = init.as_u32_mask() & 0xffff;
+
+        data.with_ref(|buf| {
+            for byte in buf {
+                crc =
+                    ((crc << 8) & 0xFF00) ^ CRCTAB_HQX[((crc >> 8) as u8 ^ (byte)) as usize] as u32;
+            }
+        });
+
+        return crc;
```
youknowone, line 128: Please check this clippy advice

youknowone (changes requested):

stdlib/src/binascii.rs
yangbeom: fixed this error 
check use `cargo +nightly clippy` command

youknowone:
For now, I have no idea why adding crc_hqx leads test failures.
If you want `a2b_qp` to be merged first, please create a new PR only with `a2b_qp`.

Snowapril:
Hmm.. As `crc_hqx` does not exist before, all the uses of `BinHex.{_write, _writecrc, _read, _checkcrc]` in `binhex.py` caused `AttributeError` such as "binascci has no attribute crc_hqx". 
Because of the existence of `expectedFailure`, test_binhex test cases were passed. 
`crc_hqx` implementation is now available, there is no failure in the middle of test cases and it leads to tearDown calling. 
We need to focus on them. But I think they need to be addressed in a separate PR 😊

yangbeom:
removed not neccesery code about a2b_qp

youknowone:
This is happened by a potential bug of `binhex`. But `binhex` is deprecated since Python 3.9. I don't think it is worth to fix it. 
Let's remove useless `test_binhex.py`

PR #3779, opened 2022-06-11
Add itertools.chain.__setstate__

nonzerofloat:


youknowone:
@nonzerofloat you may forgot to set `git config user.email`

nonzerofloat:
Generally git do not allow commit without email registered. That's perhaps the commit's email is not consistent with my account's email. I have fixed.

Correct lint errors.

vm/src/stdlib/itertools.rs
```diff
@@ -61,18 +61,49 @@ mod decl {
...
+            {
+                return Err(vm.new_type_error(String::from("Arguments must be iterators.")));
+            }
+            *self.iterables.write() = source.try_to_value(vm)?;
+            self.cur_idx.store(0);
+            *self.cached_iter.write() = if let Some(active) = active {
+                Some(active.clone().get_iter(vm)?)
+            } else {
+                None
+            };
```
youknowone, line 89:
```suggestion
            *self.cached_iter.write() = active.map(|active| active.clone().get_iter(vm)).transpose()?;
```

vm/src/stdlib/itertools.rs
```diff
@@ -61,18 +61,49 @@ mod decl {
...
+            Ok(())
+        }
     }
+
     impl IterNextIterable for PyItertoolsChain {}
     impl IterNext for PyItertoolsChain {
         fn next(zelf: &Py<Self>, vm: &VirtualMachine) -> PyResult<PyIterReturn> {
+            let iterables = zelf.iterables.read();
```
youknowone, line 97: This change may cause dead-lock

nonzerofloat: Can you give me more detailed explanation for deadlock? I suspect the place is suitable. Deadlock is unexpected.

youknowone: Guess we have a chain with self-editing iterable on `__iter__`.  Previously, we locked iterable only before and after `get_iter`, it is safe.
After change, this line will hold the lock for entire loop. Any write operation in get_iter will be permanently blocked.

vm/src/stdlib/itertools.rs
nonzerofloat: Accepted

rng-dynamics:
The implementation of `itertools.chain` has changed with #3788. I guess we should rebase this MR onto main (or something similar).

youknowone:
Oh, this is done in #4232 

Thanks @dannasman !

vm/src/stdlib/itertools.rs
```diff
@@ -58,7 +58,36 @@ mod decl {
         fn class_getitem(cls: PyTypeRef, args: PyObjectRef, vm: &VirtualMachine) -> PyGenericAlias {
             PyGenericAlias::new(cls, args, vm)
         }
+
+        #[pymethod(magic)]
+        fn setstate(&self, state: PyTupleRef, vm: &VirtualMachine) -> PyResult<()> {
+            let args = state.as_slice();
+            if args.is_empty() {
+                let msg = String::from("function takes at leat 1 arguments (0 given)");
```
Josh015, line 66: `leat` should be `least`.

PR #3781, opened 2022-06-11
[WIP] refactor number protocol store in pytype

qingshi163:


qingshi163:
I am having the problem with this appoarch, the pyobject has a pytyperef witch is protected via rw lock but pytype's fields also need a lock to support interior mutability, that will make me difficult to return the borrowed reference for methods stored in pytype.

I am thinking maybe we could have a lock for whole pytype and cancel the lock in the pyobject as I don't see any chance pyobject could change it's type and pytype itself is rarely changed.

youknowone:
I met the same conclusion when I tried the same thing. I think we don't need to lock it. PyObjectRef is already Arc. Locking the type every time only to prevent the type-changing is too much. Instead of it, we can keep the rule to use only single access to `PyObject::class()` for consecutive operations.

youknowone:
replaced by #3819

PR #3784, opened 2022-06-13
Add itertools count repr

oow214:


vm/src/stdlib/itertools.rs
```diff
@@ -202,6 +202,13 @@ mod decl {
         fn reduce(zelf: PyRef<Self>) -> (PyTypeRef, (BigInt,)) {
             (zelf.class().clone(), (zelf.cur.read().clone(),))
         }
+
+        #[pymethod(magic)]
+        fn repr(&self, _vm: &VirtualMachine) -> PyResult<String> {
+            let fmt = &self.cur.read().to_string();
```
youknowone, line 208:
```suggestion
            let cur = self.cur.read();
```
Because `format!` will format the number, this is redundant.

Use `.as_bigint()` to refer actual int data from `PyIntRef`

youknowone (approved):
Looks good! thank you for contributing!

PR #3788, opened 2022-06-14
Fix itertools chain

### LINKED ISSUE #3787: itertools.chain should evaluate lazily
CPython:
```
>>> from itertools import *
>>> a = chain.from_iterable(repeat(range(2)))
>>> next(a)
0
>>> next(a)
1
>>> next(a)
0
```

RustPython does not evaluate lazily and the call to `from_iterable` never returns:
```
>>>>> from itertools import *
>>>>> a = chain.from_iterable(repeat(range(2)))
```

rng-dynamics:
Fix #3787

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -7,7 +7,7 @@ mod decl {
         rc::PyRc,
     };
     use crate::{
-        builtins::{int, PyGenericAlias, PyInt, PyIntRef, PyTuple, PyTupleRef, PyTypeRef},
+        builtins::{int, PyGenericAlias, PyInt, PyIntRef, PyList, PyTuple, PyTupleRef, PyTypeRef},
         convert::ToPyObject,
         function::{ArgCallable, FuncArgs, OptionalArg, OptionalOption, PosArgs},
         identifier,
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -65,37 +63,45 @@ mod decl {
...
+                                Ok(PyIterReturn::StopIteration(_)) => {
+                                    *zelf.active.write() = None;
+                                }
+                                Err(err) => {
+                                    return Err(err);
+                                }
+                            },
+                        }
+                    },
                 }
```
fanninpm, line 99: 📎 "_It looks like you're `match`ing variants of an `Option`. Would you like help?_" 📎

```rust
if let Some(source) = zelf.source.read().clone() {
    loop {
        if let Some(active) = zelf.active.read().clone() {
            match active.next(vm) {
                Ok(PyIterReturn::Return(ok)) => {
                    return Ok(PyIterReturn::Return(ok));
                }
                Ok(PyIterReturn::StopIteration(_)) => {
                    *zelf.active.write() = None;
                }
                Err(err) => {
                    return Err(err);
                }
            }
        } else {
            match source.next(vm) {
                Ok(PyIterReturn::Return(ok)) => {
                    *zelf.active.write() = Some(ok.get_iter(vm)?);
                }
                Ok(PyIterReturn::StopIteration(_)) => {
                    return Ok(PyIterReturn::StopIteration(None));
                }
                Err(err) => {
                    return Err(err);
                }
            }
        }
    }
} else {
    Ok(PyIterReturn::StopIteration(None))
}
```

youknowone (changes requested):
Thank you! The changes looks good.
Do you have more works to do? I wonder what made this marked as draft.

rng-dynamics:
Thanks for the suggestions. I tried to incorporate them and I tried to reduce the nesting. I also removed the closure which I didn't like that much. Let me know what you think. If you have more suggestions, I'd be happy to hear them.

> Thank you! The changes looks good. Do you have more works to do? I wonder what made this marked as draft.

The main reason was the clippy test which failed.

youknowone (approved):
Looks good, thank you!

youknowone:
It fixed the ridiculously slow running of `test_itertools.py`, nice catch and fix!
Now the only slow running is `tee`

PR #3789, opened 2022-06-15
Update traceback to CPython 3.10

fanninpm:


fanninpm:
Largely superseded by #3870.

youknowone:
oh, you already did almost same work before

youknowone:
I figured out only `expectedFailure` marks was missed for `test_locals` in this PR comparing to recent works. 🥲

PR #3790, opened 2022-06-15
Fix broken module_names in compile_dir for sub-directories

xforce:
Assume we have a directory structure like this:
```
test
  nested_test_1
    nested_test_2
```

Expected module names are:
```
nested_test_1
nested_test-1.nested_test-2
```

However in the actual result was:
```
nested_test_1
nested_test-1nested_test-2
```
the '.' separator missing

### DIFF
--- a/derive/src/compile_bytecode.rs
+++ b/derive/src/compile_bytecode.rs
@@ -143,7 +143,11 @@ impl CompilationSource {
             if path.is_dir() {
                 code_map.extend(self.compile_dir(
                     &path,
-                    format!("{}{}", parent, file_name),
+                    if parent.is_empty() {
+                        file_name.to_string()
+                    } else {
+                        format!("{}.{}", parent, file_name)
+                    },
                     mode,
                 )?);
             } else if file_name.ends_with(".py") {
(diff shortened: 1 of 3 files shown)


derive/src/compile_bytecode.rs
```diff
@@ -143,7 +143,11 @@ impl CompilationSource {
             if path.is_dir() {
                 code_map.extend(self.compile_dir(
                     &path,
-                    format!("{}{}", parent, file_name),
+                    if parent.is_empty() {
+                        format!("{}{}", parent, file_name)
```
fanninpm, line 147:
```suggestion
                        format!("{}", file_name)
```

xforce: Yes. Though. Might as well just go with `file_name.to_string()` at that point. Will update the PR.

xforce:
> oh, wow. unexpected. how can I test its wrong behavior? I added a test but the test also success for main branch. I want to add a related test for this fix.

https://github.com/xforce/rust-python-repo/blob/incorrect-module-name/src/main.rs

that code will print
```
  File "<embedded>", line 1, in <module>
ModuleNotFoundError: No module named 'nested1.nested2'
```
if you change the `from nested1.nested2` to `from nested1nested2` it works without error.
this is what this PR fixes.

xforce:
I am at this time unfortunately not familiar enough with this codebase to provide a better test as I haven't fully explored if that particular code path is even involved in the setup that's used for tests

youknowone:
@xforce Thank you! I added a new test based on your example

youknowone (approved):
Thank you for contributing!

PR #3801, opened 2022-06-19
fix panic when parsing invalid decimal, close #3796

### LINKED ISSUE #3796: `0.E` causes panic
Expectation:
return error properly instead of panic

Actual
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseFloatError { kind: Invalid }', /home/szx5097/.cargo/git/checkouts/rustpython-f8ef4d934ac33cd8/ff50213/parser/src/lexer.rs:303:52
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: rustpython_parser::lexer::Lexer<T>::lex_normal_number
   4: rustpython_parser::lexer::Lexer<T>::inner_next
   5: <rustpython_parser::lexer::Lexer<T> as core::iter::traits::iterator::Iterator>::next
   6: lalrpop_util::state_machine::Parser<D,I>::drive
   7: rustpython_parser::parser::parse
   8: rustpython_compiler::compile
   9: rustpython_vm::vm::compile::<impl rustpython_vm::vm::VirtualMachine>::compile
  10: <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once
  11: std::thread::local::LocalKey<T>::with
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Reproduce:

```rust
use std::env;
use rustpython_vm as vm;

    let args: Vec<String> = env::args().collect();

    let filepath = &args[1];
    let data = std::fs::read(filepath).unwrap();
        let code = String::from_utf8_lossy(&data).into_owned();

        let code_obj = vm
            .compile(&code, vm::compile::Mode::Exec, "<embedded>".to_owned())
            .map_err(|err| vm.new_syntax_error(&err))?;


        let res: vm::PyResult<()> = Ok(());
        return res;
    });
    Ok(())
}
```

Cargo.toml
```
[dependencies]
rustpython-vm = { git = "https://github.com/RustPython/RustPython.git", rev = "ff502134dbb9dd3bfc33331308de255ebefcabf2"}
```

Input:
[in.zip](https://github.com/RustPython/RustPython/files/8921417/in.zip)

killme2008:
Try to fix #3796 

In cpython:

```
>>> 0.E
  File "<stdin>", line 1
    0.E
     ^
SyntaxError: invalid decimal literal
```

After this patch , RustPython:

```
>>>>> 0.E
SyntaxError: Invalid decimal literal at line 1 column 4
0.E
   ^
```

### DIFF
--- a/parser/src/lexer.rs
+++ b/parser/src/lexer.rs
@@ -300,7 +300,10 @@ where
                 value_text.push_str(&self.radix_run(10));
             }
 
-            let value = f64::from_str(&value_text).unwrap();
+            let value = f64::from_str(&value_text).map_err(|_| LexicalError {
+                error: LexicalErrorType::OtherError("Invalid decimal literal".to_owned()),
+                location: self.get_pos(),
+            })?;
             // Parse trailing 'j':
             if self.chr0 == Some('j') || self.chr0 == Some('J') {
                 self.next_char();
--- a/extra_tests/snippets/syntax_decimal.py
+++ b/extra_tests/snippets/syntax_decimal.py
@@ -1,0 +1,6 @@
+try:
+    eval("0.E")
+except SyntaxError:
+   pass
+else:
+  assert False


parser/src/lexer.rs
```diff
@@ -300,7 +300,15 @@ where
...
-            let value = f64::from_str(&value_text).unwrap();
+            let value = match f64::from_str(&value_text) {
+                Ok(v) => v,
+                Err(_e) => {
+                    return Err(LexicalError {
+                        error: LexicalErrorType::OtherError("Invalid decimal literal".to_owned()),
+                        location: self.get_pos(),
+                    });
+                }
+            };
```
youknowone, line 311:
```suggestion
            let value = match f64::from_str(&value_text).map_err(|e| LexicalError {
                        error: LexicalErrorType::OtherError("Invalid decimal literal".to_owned()),
                        location: self.get_pos(),
                    })?;
```

youknowone (commented):
Good! exception is always better than crash. would you consider suggestion for error handling?

parser/src/lexer.rs
killme2008: Thank you , `map_err` is better, i should remember this way.

killme2008: The `e` in closure must be ignored to prevent compiling warning, so i don't commit  your suggestion directly.

youknowone: yep, I suggest on the web but you write code in a full dev environment. don't worry about it. I usually trust your additional revision than my suggestion.

youknowone (approved):
Prevent crash from `0.E`

PR #3803, opened 2022-06-19
[WIP] Remove Lock for Type in PyObject and impl Heap PyType Structure

qingshi163:
Do you agree with this appoarch?
@youknowone

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -12,6 +12,7 @@ use crate::{
     class::{PyClassImpl, StaticType},
     function::{FuncArgs, KwArgs, OptionalArg},
     identifier,
+    protocol::PyNumberMethods,
     types::{Callable, GetAttr, PyTypeFlags, PyTypeSlots, SetAttr},
     AsObject, Context, Py, PyObjectRef, PyPayload, PyRef, PyResult, VirtualMachine,
 };
@@ -22,6 +23,7 @@ use std::{borrow::Borrow, collections::HashSet, fmt, ops::Deref};
 /// type(object_or_name, bases, dict)
 /// type(object) -> the object's type
 /// type(name, bases, dict) -> a new type
+#[repr(C)]
 #[pyclass(module = false, name = "type")]
 pub struct PyType {
     pub base: Option<PyTypeRef>,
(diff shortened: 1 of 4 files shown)


vm/src/builtins/type.rs
```diff
@@ -32,6 +34,28 @@ pub struct PyType {
...
 
+#[repr(C)]
+pub struct PyHeapType {
+    pub typ: PyType,
+    pub number_methods: PyRwLock<PyNumberMethods>,
+}
+
+impl From<PyHeapType> for PyType {
+    fn from(heaptype: PyHeapType) -> Self {
+        //TODO: SAFYTY
```
fanninpm, line 45: minor nitpick: spelling

```suggestion
        //TODO: SAFETY
```

youknowone:
uuuh, i realized we cannot avoid lock due to our gil-less design. let me think about this.

vm/src/builtins/type.rs
```diff
@@ -32,6 +34,28 @@ pub struct PyType {
...
+        unsafe { std::mem::transmute(heaptype) }
```
youknowone, line 46: This is not possible. I think this `From` impl never can exist. Their references can be casted, but not about the object body itself.

vm/src/object/core.rs
```diff
@@ -110,7 +110,9 @@ struct PyInner<T> {
     typeid: TypeId,
     vtable: &'static PyObjVTable,
 
-    typ: PyRwLock<PyTypeRef>, // __class__ member
+    typ: PyTypeRef,
+    type_override: OnceCell<PyRwLock<PyTypeRef>>,
```
youknowone, line 114: I agree this is the only way we can practically accept for now, but growing `PyObject` body is not very pleasure.

vm/src/object/core.rs
```diff
@@ -640,9 +643,29 @@ impl PyObject {
         }
     }
 
-    #[inline(always)]
-    pub(crate) fn class_lock(&self) -> &PyRwLock<PyTypeRef> {
-        &self.0.typ
+    pub(crate) fn get_class(&self) -> PyTypeRef {
```
youknowone, line 646:
```suggestion
    pub(crate) fn clone_class(&self) -> PyTypeRef {
```

PR #3806, opened 2022-06-21
Update install rustup

ellie-ys:
update rustup install

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ A Python-3 (CPython >= 3.10.0) Interpreter written in Rust :snake: :scream:
 
 #### Check out our [online demo](https://rustpython.github.io/demo/) running on WebAssembly.
 
-RustPython requires Rust latest stable version (e.g 1.43.0 at May 24th 2020). 
+RustPython requires Rust latest stable version (e.g 1.43.0 at May 24th 2020).
 To check Rust version: `rustc --version` If you wish to update,
 `rustup update stable`.
 
@@ -29,6 +29,8 @@ To build RustPython locally, do the following:
     $ git clone https://github.com/RustPython/RustPython
     $ cd RustPython
       # --release is needed (at least on windows) to prevent stack overflow
+
+    $ curl https://sh.rustup.rs -sSf | sh -s -- -y
     $ cargo run --release demo.py
     Hello, RustPython!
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,6 +29,8 @@ To build RustPython locally, do the following:
     $ git clone https://github.com/RustPython/RustPython
     $ cd RustPython
       # --release is needed (at least on windows) to prevent stack overflow
+
+    $ curl https://sh.rustup.rs -sSf | sh -s -- -y
```
fanninpm, line 33: Installing `rustup` is only applicable for those who have not already installed it. Perhaps this should be moved to before the `git clone` step with an applicable comment.

youknowone: I also want to link https://rustup.rs website instead of putting the script here. The script is various depending on running platforms

README.md
```diff
@@ -29,6 +29,8 @@ To build RustPython locally, do the following:
...
     $ cargo run --release demo.py
```
bonjune, line 34: `demo.py` should be changed to `demo_closures.py`. There is no `demo.py` in this repository root.

README.md
```diff
@@ -18,36 +18,49 @@ A Python-3 (CPython >= 3.11.0) Interpreter written in Rust :snake: :scream:
 
 ## Usage
 
-#### Check out our [online demo](https://rustpython.github.io/demo/) running on WebAssembly.
+**Check out our [online demo](https://rustpython.github.io/demo/) running on WebAssembly.**
```
DimitrisJim, line 21: some of these changes are made after warnigns from markdown lint.

README.md
```diff
@@ -96,25 +115,27 @@ $ wapm run rustpython
...
+```bash
 cargo build --release --target wasm32-wasi --features="freeze-stdlib"
 ```
 
 > Note: we use the `freeze-stdlib` to include the standard library inside the binary. You also have to run once `rustup target add wasm32-wasi`.
 
 ### JIT (Just in time) compiler
 
-RustPython has a **very** experimental JIT compiler that compile python functions into native code. 
+RustPython has an **very** experimental JIT compiler that compile python functions into native code.
```
youknowone, line 126:
```suggestion
RustPython has a **very** experimental JIT compiler that compile python functions into native code.
```

youknowone (approved):

PR #3807, opened 2022-06-22
Run test suites like how CPython runs theirs

fanninpm:
The extra flags come from the `buildbottest` rule in CPython's Makefile.

This will enable checking the tests we've been skipping by default (such as network-based tests). This will also report the top 10 slowest tests and will mark a test as failed if it altered the execution environment.


fanninpm:
I can investigate the Windows failures tomorrow evening.

youknowone (approved):
great!!

PR #3813, opened 2022-06-23
format_int precision error handling

gilteunchoi:
Because precision is not allowed in integer format specifier:

In this revision, 
* Precision error handling has been done in `format_int` function. 

Before:
```
>>>>> print('{:.2}'.format(200))
200
```
After:
```
>>>>> print('{:.2}'.format(200))
ValueError: Precision not allowed in integer format specifier
```

### DIFF
--- a/vm/src/format.rs
+++ b/vm/src/format.rs
@@ -428,7 +428,7 @@ impl FormatSpec {
     }
 
     #[inline]
-    fn format_int_internal(&self, magnitude: BigInt, radix: u32) -> Result<String, &'static str> {
+    fn format_int_radix(&self, magnitude: BigInt, radix: u32) -> Result<String, &'static str> {
         match self.precision {
             Some(_) => Err("Precision not allowed in integer format specifier"),
             None => Ok(magnitude.to_str_radix(radix)),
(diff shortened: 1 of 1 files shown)


vm/src/format.rs
```diff
@@ -428,6 +428,10 @@ impl FormatSpec {
     }
 
     pub(crate) fn format_int(&self, num: &BigInt) -> Result<String, &'static str> {
+        if self.precision.is_some() {
+            return Err("Precision not allowed in integer format specifier");
+        }
+
```
Snowapril, line 434: This check should be conducted only if the format type is a kind of integer. That is, the checks should not be executed when the format type is not a kind of integer.

youknowone:
@Snowapril could you review this again?

vm/src/format.rs
```diff
@@ -428,6 +428,28 @@ impl FormatSpec {
...
+            | Some(FormatType::FixedPointLower)
+            | Some(FormatType::GeneralFormatUpper)
+            | Some(FormatType::GeneralFormatLower)
+            | Some(FormatType::Percentage) => false,
+            None => true,
+        };
+        if precision_error_case && self.precision.is_some() {
+            return Err("Precision not allowed in integer format specifier");
+        }
+
```
Snowapril, line 452: How about add simple method and replace original `Ok(magnitude.to_str_radix(~~))`. Like this
```rust
#[inline]
fn format_int_internal(&self, magnitude: BigInt, radix: u32) -> Result<String, &'static str>  {
    match self.precision {
        Some(_) => Err("Precision not allowed in integer format specifier"),
        None => Ok(magnitude.to_str_radix(radix))
    }
}
```
And 
```rust
let raw_magnitude_string_result: Result<String, &'static str> = match self.format_type {
            Some(FormatType::Binary) => format_int_internal(magnitude, 2),
            Some(FormatType::Decimal) => format_int_internal(magnitude, 10),
            Some(FormatType::Octal) => format_int_internal(magnitude, 8),
            Some(FormatType::HexLower) => format_int_internal(magnitude, 16),
            Some(FormatType::HexUpper) => {
                let mut result = format_int_internal(magnitude, 16)?;
                result.make_ascii_uppercase();
                Ok(result)
            }
```

Snowapril (approved):
Looks good to me!

vm/src/format.rs
```diff
@@ -427,6 +427,14 @@ impl FormatSpec {
         self.format_sign_and_align(&magnitude_string, sign_str)
     }
 
+    #[inline]
+    fn format_int_internal(&self, magnitude: BigInt, radix: u32) -> Result<String, &'static str> {
```
youknowone, line 431:
```suggestion
    fn format_int_radix(&self, magnitude: BigInt, radix: u32) -> Result<String, &'static str> {
```
when we can name it specific name, I like to avoid too common names like `internal`

youknowone (changes requested):
Thank you! I left a small comment about the function naming. Could you check it?

youknowone (approved):
thank you for contributing!

PR #3814, opened 2022-06-23
add copyright property in sre module

sm136599:
* add copyright property in sre module
-------------
result
```
>>>>> import _sre
>>>>> _sre.copyright
' SRE 2.2.2 Copyright (c) 1997-2002 by Secret Labs AB '
```

### DIFF
--- a/vm/src/stdlib/sre.rs
+++ b/vm/src/stdlib/sre.rs
@@ -28,12 +28,8 @@ mod _sre {
     #[pyattr]
     pub use sre_engine::{constants::SRE_MAGIC as MAGIC, CODESIZE, MAXGROUPS, MAXREPEAT};
 
-    #[pyattr]
-    fn copyright(_vm: &VirtualMachine) -> &str {
-        let copy_right = " SRE 2.2.2 Copyright (c) 1997-2002 by Secret Labs AB ";
-        copy_right
-    }
-
+    #[pyattr(name = "copyright")]
+    const COPYRIGHT : &str = " SRE 2.2.2 Copyright (c) 1997-2002 by Secret Labs AB ";
     #[pyfunction]
     fn getcodesize() -> usize {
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/sre.rs
```diff
@@ -28,6 +28,12 @@ mod _sre {
     #[pyattr]
     pub use sre_engine::{constants::SRE_MAGIC as MAGIC, CODESIZE, MAXGROUPS, MAXREPEAT};
 
+    #[pyattr]
+    fn copyright(_vm: &VirtualMachine) -> &str {
+        let copy_right = " SRE 2.2.2 Copyright (c) 1997-2002 by Secret Labs AB ";
+        copy_right
+    }
+
```
Snowapril, line 36:
```suggestion
    #[pyattr(name = "copyright")]
    const COPYRIGHT : &str = " SRE 2.2.2 Copyright (c) 1997-2002 by Secret Labs AB ";
```

Snowapril: Oh, I missed that... But It seems also true because most of the things offered by `sre.rs` and `_sre.c` are the same.

youknowone:
@qingshi163 as i understand, this is mostly your work, right?

vm/src/stdlib/sre.rs
youknowone: That means the API is the same. It doesn't necessarily mean the code is directly translation from one source to the other one. Though we now have unclear points due to google vs oracle case, basically copyright came from copy of code, not from copy of signatures.

qingshi163:
> @qingshi163 as i understand, this is mostly your work, right?

Yes, the implements is heavily inspired by cpython as we use the same frontend, but I don't think there is direct translation.

youknowone:
thank you for checking

PR #3815, opened 2022-06-23
Fix OverflowError at collections.rs

dldks321:
Fix OverflowError when an integer value larger than Rust's isize is entered in the index method of deque.
and add test case for checking this error

before
```
>>>>> deq = deque()
>>>>> deq.index(92233720368547758090,0,11111111111111111111111111111111111111)
OverflowError: Python int too large to convert to Rust isize
```

after
```
>>>>> deq = deque()
>>>>> deq.index(92233720368547758090,0,11111111111111111111111111111111111111)
ValueError: 92233720368547758090 is not in deque
```

### DIFF
--- a/Lib/test/seq_tests.py
+++ b/Lib/test/seq_tests.py
@@ -362,8 +362,6 @@ class CommonTest(unittest.TestCase):
 
         self.assertRaises(BadExc, a.count, BadCmp())
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_index(self):
         u = self.type2test([0, 1])
         self.assertEqual(u.index(0), 0)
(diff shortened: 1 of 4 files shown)


Lib/test/seq_tests.py
```diff
@@ -362,8 +362,6 @@ def __eq__(self, other):
 
         self.assertRaises(BadExc, a.count, BadCmp())
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
```
youknowone, line 366: This is breaking test_tuple because tuple also has similar bug

Snowapril: It seems `tuple.index` could be solved in a similar way as we did in `deque.index` and `list.index`.
So, I think we have two selections, making `test_tuple.test_index` as `expectedFailure` or fixing `tuple.index`.

dldks321: I'm fixing ```tuple.index```
It is really similar to ```list.index``` and ```deque.index```  : )

but using deq[10000000000000],
overflow error is ouccr.
because may be same bug exist at item.

PR #3816, opened 2022-06-23
Add `binascii` `a2b_qp`

chang-57:


stdlib/src/binascii.rs
```diff
@@ -181,6 +181,74 @@ mod decl {
...
     }
+    #[derive(FromArgs)]
+    struct A2bqpArgues {
+        #[pyarg(any)]
+        data: ArgAsciiBuffer,
+        #[pyarg(named, default = "false")]
+        header: bool,
+    }
+    #[pyfunction]
+    fn a2b_qp(args: A2bqpArgues) -> PyResult<Vec<u8>> {
```
fanninpm, line 192: `Argues` is a very peculiar way to write "args" or "arguments". Perhaps this struct should look like this:

```suggestion
    struct A2bQpArgs {
        #[pyarg(any)]
        data: ArgAsciiBuffer,
        #[pyarg(named, default = "false")]
        header: bool,
    }
    #[pyfunction]
    fn a2b_qp(args: A2bQpArgs) -> PyResult<Vec<u8>> {
```

@youknowone I think we should take a look at every `struct` that derives `FromArgs` so we can agree on a consistent naming scheme.

Snowapril: I felt creating a new struct outside of function for everything which supports optional, named, etc is somewhat cumbersome to me. I think there's a better way for us to deal with this..?

youknowone:
please try to rebase instead of merge https://docs.gitlab.com/ee/topics/git/git_rebase.html

Snowapril:
@youknowone This PR seems ready for review. Could you run CI workflow? 😊

stdlib/src/binascii.rs
```diff
@@ -181,6 +181,299 @@ mod decl {
...
+                    {
+                        linelen = 0;
+                        if (inidx != 0) && ((buf[inidx - 1] == 0x20) || (buf[inidx - 1] == 0x09)) {
+                            delta += 2;
+                        }
+                        if crlf {
+                            delta += 2;
+                        } else {
+                            delta += 2;
+                        }
```
youknowone, line 335: maybe one of the number is wrong

Snowapril:
```suggestion
                        if crlf {
                            delta += 2;
                        } else {
                            delta += 1;
                        }
```

Snowapril:
please also push `cargo fmt --all` 😊

Snowapril:
It seems `test_qp` in `test_binascii.py` is fixed by your change. Good! 
Please remove `# TODO: RUSTPYTHON` and `@unittest.expectedFailure` lines 😊

youknowone (approved):

PR #3817, opened 2022-06-23
Fix nightly clippy warnings

youknowone:


fanninpm:
We can fix these, but I need to remind myself to switch Clippy to the stable channel next Thursday (when Rust 1.62 is released).

Regarding that failure in test_strtod, I think the offending test should be just outright skipped.

Lib/test/test_strtod.py
```diff
@@ -219,9 +219,8 @@ def test_bigcomp(self):
                 s = '{}e{}'.format(digits, exponent)
                 self.check_strtod(s)
 
-    @unittest.skipIf(sys.platform == "win32", "TODO: RUSTPYTHON, possibly flaky test on Windows?")
     # TODO: RUSTPYTHON, Incorrectly rounded str->float conversion for -07e-321
-    @unittest.expectedFailure
+    @unittest.skip("TODO: RUSTPYTHON, possibly flaky test on Windows?")
```
fanninpm, line 223: It's not just on Windows. (I wonder if it's related to why `test_halfway_cases` is flaky…)

```suggestion
    @unittest.skip("TODO: RUSTPYTHON, possibly flaky test?")
```

PR #3819, opened 2022-06-25
HeapTypeExt and Protocol Methods Pointer

qingshi163:


qingshi163:
@youknowone can you run the performance check? I think we may use week order read after replace AtomicCell to AtomicPtr

youknowone:
I am running tests. I used benchmark tests from https://github.com/python/pyperformance/tree/main/pyperformance/data-files/benchmarks
It has `pyperf` dependency. I manually installed it to my `RUSTPYTHONPATH`.

running command was:
```
./rustpython_main fannkuch.py -o reports/fannkuch_main.json
```

comparing command was:
```
python -m pyperf compare_to --table reports/*nbody*
```

```
$ python3 -m pyperf compare_to --table reports/*nbody*
+-----------+-------------------+----------------------+-------------------------+-------------------------+
| Benchmark | nbody_cpython3.10 | nbody_cpython3.9     | nbody_rustpython_3819   | nbody_rustpython_main   |
+===========+===================+======================+=========================+=========================+
| nbody     | 104 ms            | 101 ms: 1.03x faster | 1.78 sec: 17.16x slower | 1.80 sec: 17.36x slower |
+-----------+-------------------+----------------------+-------------------------+-------------------------+
$ python3 -m pyperf compare_to --table reports/*fannkuch*
+-----------+----------------------+----------------------+--------------------------+--------------------------+
| Benchmark | fannkuch_cpython3.10 | fannkuch_cpython3.9  | fannkuch_rustpython_3819 | fannkuch_rustpython_main |
+===========+======================+======================+==========================+==========================+
| fannkuch  | 352 ms               | 354 ms: 1.01x slower | 3.63 sec: 10.31x slower  | 3.66 sec: 10.41x slower  |
+-----------+----------------------+----------------------+--------------------------+--------------------------+
$ python3 -m pyperf compare_to --table reports/*scimark* 
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| Benchmark               | scimark_cpython3.10 | scimark_cpython3.9    | scimark_rustpython_3819 | scimark_rustpython_main |
+=========================+=====================+=======================+=========================+=========================+
| scimark_fft             | 291 ms              | 241 ms: 1.21x faster  | 3.82 sec: 13.11x slower | 3.87 sec: 13.28x slower |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| scimark_lu              | 120 ms              | 115 ms: 1.05x faster  | 1.14 sec: 9.45x slower  | 1.14 sec: 9.49x slower  |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| scimark_monte_carlo     | 75.0 ms             | 86.9 ms: 1.16x slower | 802 ms: 10.69x slower   | 799 ms: 10.65x slower   |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| scimark_sor             | 143 ms              | 165 ms: 1.15x slower  | 1.69 sec: 11.83x slower | 1.71 sec: 11.92x slower |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| scimark_sparse_mat_mult | 4.06 ms             | 3.47 ms: 1.17x faster | 44.0 ms: 10.84x slower  | 44.0 ms: 10.84x slower  |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| Geometric mean          | (ref)               | 1.02x faster          | 11.12x slower           | 11.17x slower           |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
```

vm/src/builtins/float.rs
```diff
@@ -552,39 +553,49 @@ impl Hashable for PyFloat {
...
-        }),
-        float: Some(|number, vm| Ok(Self::number_float(number, vm))),
-        floor_divide: Some(|number, other, vm| {
-            Self::number_general_op(number, other, inner_floordiv, vm)
-        }),
-        true_divide: Some(|number, other, vm| {
-            Self::number_general_op(number, other, inner_div, vm)
-        }),
-        ..PyNumberMethods::NOT_IMPLEMENTED
+macro_rules! atomic_func {
```
youknowone, line 556: putting in `type/slots.rs` will be better to share the macro over all protocol implementations

vm/src/types/slot.rs
```diff
@@ -880,13 +932,8 @@ pub trait AsSequence: PyPayload {
 
 #[pyimpl]
 pub trait AsNumber: PyPayload {
-    const AS_NUMBER: PyNumberMethods;
-
-    #[inline]
     #[pyslot]
-    fn as_number(_zelf: &PyObject, _vm: &VirtualMachine) -> &'static PyNumberMethods {
-        &Self::AS_NUMBER
-    }
```
youknowone, line 889: is this need to be changed?

vm/src/builtins/float.rs
```diff
@@ -552,39 +553,49 @@ impl Hashable for PyFloat {
...
     };
 }
 
+impl AsNumber for PyFloat {
+    fn as_number() -> &'static PyNumberMethods {
+        &AS_NUMBER
+    }
+}
+
+static AS_NUMBER: PyNumberMethods = PyNumberMethods {
```
youknowone, line 568: even if we make this out of trait, it still can be defined in `fn as_number()` method unless we need it as a pub

vm/src/builtins/type.rs
```diff
@@ -15,9 +10,17 @@ use crate::{
     types::{Callable, GetAttr, PyTypeFlags, PyTypeSlots, SetAttr},
     AsObject, Context, Py, PyObjectRef, PyPayload, PyRef, PyResult, VirtualMachine,
 };
+use crate::{
```
youknowone, line 13:
```suggestion
```

vm/src/protocol/number.rs
```diff
@@ -1,189 +1,155 @@
+use crossbeam_utils::atomic::AtomicCell;
+use once_cell::sync::OnceCell;
+
```
youknowone, line 3:
```suggestion
```

vm/src/builtins/type.rs
```diff
@@ -30,10 +33,56 @@ pub struct PyType {
...
     pub slots: PyTypeSlots,
+    pub heaptype_ext: Option<Pin<Box<HeapTypeExt>>>,
+}
+
+#[derive(Default)]
+pub struct HeapTypeExt {
+    pub number_methods: PyNumberMethods,
+}
+
+pub struct PointerSlot<T>(NonNull<T>);
```
youknowone, line 44: What's the benefit of this type comparing to `&'static`?
They seem to guarantee same things.

vm/src/protocol/number.rs
```diff
@@ -1,189 +1,155 @@
...
     pub obj: &'a PyObject,
     // some fast path do not need methods, so we do lazy initialize
-    pub methods: Option<&'static PyNumberMethods>,
+    methods: OnceCell<PointerSlot<PyNumberMethods>>,
 }
 
-impl<'a> PyNumber<'a> {
-    pub fn new(obj: &'a PyObject, vm: &VirtualMachine) -> Self {
+impl<'a> From<&'a PyObject> for PyNumber<'a> {
+    fn from(obj: &'a PyObject) -> Self {
```
youknowone, line 116: This change means having a `PyNumber` object doesn't guarantee it is an actual number object.
Do you intend to change it?
In that case, we'd better to carefully comment about it because it is counter-intutitive to rust sense.

I still think `OnceCell` for `methods` is an overkill because it can be safely managed only with properly designed types.

qingshi163: I think PyNumber shouldn't garrentee anything as it could be int, float, subclass, complex or class with __int__, __float__ etc, so what the caller expect is depend on what method been called.

The reason I use OnceCell is because we always use the type id first to avoid method call and I think that is more common case like get int from PyInt

youknowone: We will obtain an `PyNumber` object only when we need `PyNumber` operations, isn't it? The method can be obtained when we create `PyNumber`, not when retrieving methods from it. `PyNumber` itself is a `Once`. `OnceCell` in `PyNumber` is duplicated encoding of data.

qingshi163: so maybe a function that will create PyNumber only if it is not exact type could be better?

vm/src/types/slot.rs
qingshi163: we need static variable, const variable is not globally static, it creates instance wherever is used.

vm/src/protocol/number.rs
youknowone: I am sorry. I didn't understand well what you meant by "create PyNumber only if it is not exact type". could you help me to understand it?

I think we need to create a `PyNumber` type object only when we need to call `methods()` of it. In that case, we don't need `OnceCell` for PyNumber - because we will create a `PyNumber` object always with methods. If we don't need method, we can handle `PyObjectRef` instead of `PyNumber`.

youknowone:
oh,  I learned about those performance tools from @corona10, thank you!

vm/src/protocol/number.rs
youknowone: because recent patches are adding number protocol implementations, quickly merging this one will be less painful in future.
I want to know how do you think about this OnceCell problem for now. You know about it the best.

youknowone: @qingshi163 do you mind if I try to edit this PR to work this way? I want to listen your decision because you are the best professional for PyNumber.

qingshi163: I am thinking we should remove OnceCell from PyNumber but keep From<&PyObject> trait.
Than we modify or create function under VirtualMachine(to_int, to_index, to_float, etc) for all the function that could short cut with type id and do the check there. so we will always call number methods if we created PyNumber.

We should always use functions under VirtualMachine instead PyNumber if we do not sure the type.

I'd like to keep From<&PyObject> is because PyNumber is not garuentee anything about the type. we should not check the type when creating it.

qingshi163: Sure you can modify the pr, I am quite busy and I want to fix the sre-engine first.

youknowone:
```
$ python3 -m pyperf compare_to --table reports/*nbody*
+-----------+-------------------+----------------------+-------------------------+-------------------------+
| Benchmark | nbody_cpython3.10 | nbody_cpython3.9     | nbody_rustpython_3819   | nbody_rustpython_main   |
+===========+===================+======================+=========================+=========================+
| nbody     | 103 ms            | 101 ms: 1.02x faster | 1.78 sec: 17.27x slower | 1.77 sec: 17.17x slower |
+-----------+-------------------+----------------------+-------------------------+-------------------------+

$ python3 -m pyperf compare_to --table reports/*fannkuch*
+-----------+----------------------+----------------------+--------------------------+--------------------------+
| Benchmark | fannkuch_cpython3.10 | fannkuch_cpython3.9  | fannkuch_rustpython_3819 | fannkuch_rustpython_main |
+===========+======================+======================+==========================+==========================+
| fannkuch  | 349 ms               | 348 ms: 1.00x faster | 3.59 sec: 10.31x slower  | 3.66 sec: 10.48x slower  |
+-----------+----------------------+----------------------+--------------------------+--------------------------+

$ python3 -m pyperf compare_to --table reports/*scimark* 
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| Benchmark               | scimark_cpython3.10 | scimark_cpython3.9    | scimark_rustpython_3819 | scimark_rustpython_main |
+=========================+=====================+=======================+=========================+=========================+
| scimark_fft             | 290 ms              | 239 ms: 1.21x faster  | 3.82 sec: 13.17x slower | 3.84 sec: 13.24x slower |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| scimark_lu              | 120 ms              | 114 ms: 1.06x faster  | 1.13 sec: 9.41x slower  | 1.14 sec: 9.46x slower  |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| scimark_monte_carlo     | 75.6 ms             | 86.1 ms: 1.14x slower | 790 ms: 10.45x slower   | 803 ms: 10.62x slower   |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| scimark_sor             | 142 ms              | 164 ms: 1.16x slower  | 1.69 sec: 11.89x slower | 1.68 sec: 11.87x slower |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| scimark_sparse_mat_mult | 4.04 ms             | 3.43 ms: 1.18x faster | 44.1 ms: 10.91x slower  | 44.2 ms: 10.95x slower  |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+
| Geometric mean          | (ref)               | 1.03x faster          | 11.10x slower           | 11.16x slower           |
+-------------------------+---------------------+-----------------------+-------------------------+-------------------------+

$ python3 -m pyperf compare_to --table reports/*int*    
+-----------+-----------------+-----------------------+----------------------+----------------------+
| Benchmark | int_cpython3.10 | int_cpython3.9        | int_rustpython_3819  | int_rustpython_main  |
+===========+=================+=======================+======================+======================+
| int       | 47.8 ns         | 41.8 ns: 1.14x faster | 389 ns: 8.13x slower | 392 ns: 8.19x slower |
+-----------+-----------------+-----------------------+----------------------+----------------------+

$ cat int.py
import pyperf

def bench_int(loops):
    range_it = range(loops)
    t0 = pyperf.perf_counter()

    for _ in range_it:
        int(1.1)

    return pyperf.perf_counter() - t0

if __name__ == '__main__':
    DEFAULT_ITERATIONS=1000
    runner = pyperf.Runner()
    runner.metadata['description'] = "n-body benchmark"

    args = runner.parse_args()
    runner.bench_time_func('int', bench_int)
```

youknowone:
@qingshi163 could you review changes?

qingshi163 (commented):
Great!

youknowone (approved):

PR #3821, opened 2022-06-26
add <code>IMMUTABLE</code> flag and set non-heaptype as immutable

lijm1358:
* as <code>PyType_Ready()</code> sets <code>Py_TPFLAGS_IMMUTABLETYPE</code> flag if <code>Py_TPFLAGS_HEAPTYPE</code> not set in CPython, add IMMUTABLETYPE flag and set it to non-heaptype.
* add error message when setting attribution of immutable types
```
>>>>> int.a = 1
TypeError: cannot set 'a' attribute of immutable type 'int'
```
--------------
* but it's not working correctly; some types might be/not be heaptype in CPython
```
>>>>> import io
  File "D:\RustPython\Lib\io.py", line 83, in <module>
    UnsupportedOperation.__module__ = "io"
TypeError: cannot set '__module__' attribute of immutable type 'UnsupportedOperation'
```

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -104,6 +104,10 @@ impl PyType {
 
         *slots.name.get_mut() = Some(String::from(name));
 
+        if !slots.flags.has_feature(PyTypeFlags::HEAPTYPE) {
+            slots.flags |= PyTypeFlags::IMMUTABLETYPE
+        }
+
         let new_type = PyRef::new_ref(
             PyType {
                 base: Some(base),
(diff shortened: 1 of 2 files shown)


Snowapril:
RustPython has more static type than cpython. Making all static types immutable will leads to such incompatibility in the case of the type which is static type in rustpython and heap type in cpython.

youknowone:
@Snowapril do you know which type is different? Probably we'd better to consider turn them to heap types.

Snowapril:
It seems built-in types are compatible with CPython's.
With a simple python script, I found the below types are different from CPython's.

| items | rustpython | cpython3.10 |
|--------| ----- | ----- |
| _thread.LockType | static | heap |
| _thread.RLock | static | heap |
| _thread._local | static | heap |
| _io.UnsupportedOperation | static | heap |
| posix.DirEntry | static | heap |
| posix.sched_param | static | heap |
| posix.stat_result | static | heap |
| posix.terminal_size | static | heap |
| posix.times_result | static | heap |
| posix.uname_result | static | heap |
| os.path.DirEntry | static | heap |
| os.path.sched_param | static | heap |
| os.path.stat_result | static | heap |
| os.path.terminal_size | static | heap |
| os.path.times_result | static | heap |
| os.path.uname_result | static | heap |
| os.DirEntry | static | heap |
| os.sched_param | static | heap |
| os.stat_result | static | heap |
| os.terminal_size | static | heap |
| os.times_result | static | heap |
| os.uname_result | static | heap |

youknowone:
Because we recently revised type implementation a lot, this is implemented in different way.
Thank you for your contributing.

PR #3822, opened 2022-06-26
Fix itertools.count to take PyNumber instead of PyInt

oow214:


vm/src/stdlib/itertools.rs
```diff
@@ -174,14 +174,13 @@ mod decl {
...
-        cur: PyRwLock<BigInt>,
-        step: BigInt,
+        cur: PyRwLock<PyObjectRef>,
+        step: Option<PyIntRef>,
     }
 
     #[derive(FromArgs)]
     struct CountNewArgs {
-        #[pyarg(positional, optional)]
-        start: OptionalArg<PyIntRef>,
```
youknowone, line 184: `#[pyarg]` and `OptinoalArg` is not expected to be removed.
(See the test results how this change broke tests)

Use `OptionalArg<PyObjectRef>`

vm/src/stdlib/itertools.rs
```diff
@@ -235,7 +237,9 @@ mod decl {
         fn next(zelf: &Py<Self>, vm: &VirtualMachine) -> PyResult<PyIterReturn> {
             let mut cur = zelf.cur.write();
             let result = cur.clone();
-            *cur += &zelf.step;
+            if let Some(step) = &zelf.step {
+                *cur = vm._iadd(&*cur, step.as_object())?;
+            }
```
youknowone, line 242: Doesn't `1` have to be added when step is `None`?

vm/src/stdlib/itertools.rs
```diff
@@ -11,14 +11,14 @@ mod decl {
...
-        protocol::{PyIter, PyIterReturn},
+        protocol::{PyIter, PyIterReturn, PyNumber},
         stdlib::sys,
         types::{Constructor, IterNext, IterNextIterable},
         AsObject, Py, PyObjectRef, PyPayload, PyRef, PyResult, PyWeakRef, VirtualMachine,
     };
     use crossbeam_utils::atomic::AtomicCell;
-    use num_bigint::BigInt;
-    use num_traits::{One, Signed, ToPrimitive, Zero};
+    // use num_bigint::BigInt;
```
Snowapril, line 20:
```suggestion
```

vm/src/stdlib/itertools.rs
```diff
@@ -174,14 +173,14 @@ mod decl {
     #[pyclass(name = "count")]
     #[derive(Debug, PyPayload)]
     struct PyItertoolsCount {
-        cur: PyRwLock<BigInt>,
-        step: BigInt,
+        cur: PyRwLock<PyObjectRef>,
+        step: Option<PyIntRef>,
```
Snowapril, line 177: Can I ask you is there any reason for making `step` as `Option`?
It seems `step` also need to be checked via `PyNumber::check` in cpython which means that any type that implements `number protocol` can be fitted.

vm/src/stdlib/itertools.rs
```diff
@@ -196,13 +195,19 @@ mod decl {
             vm: &VirtualMachine,
         ) -> PyResult {
             let start = match start.into_option() {
-                Some(int) => int.as_bigint().clone(),
-                None => BigInt::zero(),
+                Some(num) => vm.new_pyobj(num),
+                None => vm.new_pyobj(0),
             };
```
Snowapril, line 200:
```rust
let start: PyObjectRef = start.into_option().unwrap_or_else(|| vm.new_pyobj(0));
```

vm/src/stdlib/itertools.rs
```diff
@@ -196,13 +195,19 @@ mod decl {
...
             let step = match step.into_option() {
-                Some(int) => int.as_bigint().clone(),
-                None => BigInt::one(),
+                Some(int) => {
+                    let val: isize = int.try_to_primitive(vm)?;
+                    Some(vm.new_pyref(val.to_usize().unwrap_or(0)))
+                }
+                None => None,
             };
```
Snowapril, line 207:
```rust
let step: PyObjectRef = step.into_option().unwrap_or_else(|| vm.new_pyobj(1));
```

vm/src/stdlib/itertools.rs
```diff
@@ -196,13 +195,19 @@ mod decl {
...
+            if !PyNumber::check(&start, vm) {
```
Snowapril, line 208:
```suggestion
            if !PyNumber::check(&start, vm) || !PyNumber::check(&step, vm) {
```
It seems `step` also need to be checked via `PyNumber::check`.
https://github.com/python/cpython/blob/3.10/Modules/itertoolsmodule.c#L4178-L4182

vm/src/stdlib/itertools.rs
```diff
@@ -235,7 +240,12 @@ mod decl {
         fn next(zelf: &Py<Self>, vm: &VirtualMachine) -> PyResult<PyIterReturn> {
             let mut cur = zelf.cur.write();
             let result = cur.clone();
-            *cur += &zelf.step;
+            if let Some(step) = &zelf.step {
+                *cur = vm._iadd(&*cur, step.as_object())?;
+            } else {
+                let one = vm.new_pyobj(1);
+                *cur = vm._add(&*cur, &one)?;
+            }
```
Snowapril, line 248:
```suggestion
            *cur = vm._add(&*cur, &zelf.step)?;
```
If you have interest, this line could be replaced with `PyNumber::add` which is not yet implemented in rustpython

youknowone: that will not be easy before #3819

vm/src/stdlib/itertools.rs
Snowapril:
```suggestion
        step: PyObjectRef,
```

oow214: Because step is an optional value as the second argument of count, set it as Option.
And when processing, there is something similar to repeat, so I considered it.
But if I need to set it up differently, I will.
I checked all reviews. thank. I'll fix.

vm/src/stdlib/itertools.rs
oow214: I didn't because it's a PyIntRef. I'll fix it.

vm/src/stdlib/itertools.rs
youknowone: What's different between saving it as a `PyIntRef` and `Option<PyIntRef>`?
e.g. Does it affect to `repr` result? If they act same for everything, we don't need to distinguish them.

vm/src/stdlib/itertools.rs
youknowone: PyIntRef is always a PyNumber.

vm/src/stdlib/itertools.rs
oow214: Sorry. I was wrong. I'll fix it. Thanks for the suggestion.

PR #3824, opened 2022-06-28
Fix itertools count repr

oow214:


fanninpm:
Feel free to ignore those Clippy failures. @youknowone likely missed them in #3817.

fanninpm:
In `Lib/test/test_itertools.py`, if you comment out the `@unittest.expectedFailure` decorator above `def test_count(self):` (about line 537 or so), what happens? I suspect that the `__reduce__()` method of `itertools.count` is causing the pickle portion of that test to fail.

youknowone:
please do not use `git merge`. would you try to rebase it?

https://docs.gitlab.com/ee/topics/git/git_rebase.html

vm/src/stdlib/itertools.rs
```diff
@@ -220,9 +220,13 @@ mod decl {
         }
 
         #[pymethod(magic)]
-        fn repr(&self) -> PyResult<String> {
-            let cur = self.cur.read();
-
+        fn repr(&self, vm: &VirtualMachine) -> PyResult<String> {
+            let mut cur = format!("{}", self.cur.read().clone().repr(vm)?);
+            let step = format!("{}", self.step.clone());
+            if step != "1" {
```
youknowone, line 226: I expected comparing the value instead of comparing the result.
Do you know what's the original way?

Snowapril: cpython is also comparing the value (https://github.com/python/cpython/blob/3.10/Modules/itertoolsmodule.c#L4305-L4310).

vm/src/stdlib/itertools.rs
```diff
@@ -221,13 +221,12 @@ mod decl {
 
         #[pymethod(magic)]
         fn repr(&self, vm: &VirtualMachine) -> PyResult<String> {
-            let mut cur = format!("{}", self.cur.read().clone().repr(vm)?);
-            let step = format!("{}", self.step.clone());
-            if step != "1" {
-                cur.push_str(", ");
-                cur.push_str(&step);
+            let cur = format!("{}", self.cur.read().clone().repr(vm)?);
+            let step = self.step.as_u32_mask().clone();
```
Snowapril, line 225:
```suggestion
            let step = self.step.as_bigint();
```

Snowapril:
After #3830 merged, the clippy error also will be passed.

youknowone (approved):
Thank you for contributing!

PR #3828, opened 2022-06-29
Introduce PyAtomicRef and impl for PyObject.typ

qingshi163:
@youknowone correctness and performance check?


vm/src/object/ext.rs
```diff
@@ -323,6 +317,46 @@ where
...
+            ptr: AtomicCell::new(ptr),
+        }
+    }
+}
+
+impl<T: PyObjectPayload> Deref for PyAtomicRef<T> {
+    type Target = Py<T>;
+
+    fn deref(&self) -> &Self::Target {
+        unsafe { self.ptr.load().as_ref() }
```
youknowone, line 347: This is looking unsafe.

While thread 1 called `deref()` to obtain `&Py<T>`, `inner` can be swapped out in thread 2.
When the swapped out object is dropped, thread 1 will have a dangling pointer.

qingshi163:
#3970

qingshi163:
Is this approach safe as we manaully require the reference to the pytype cannot across between the frames. Most of the case we use the pytype is either using pytyperef or using the reference immediately.
@youknowone

youknowone:
Oh, I missed your last push. I will check the changes soon.

vm/src/builtins/type.rs
```diff
@@ -266,20 +266,20 @@ impl PyType {
     }
 }
 
-impl PyTypeRef {
+impl Py<PyType> {
```
youknowone, line 269: This is a great idea.

vm/src/frame.rs
```diff
@@ -111,6 +111,7 @@ pub struct Frame {
     /// tracer function for this frame (usually is None)
     pub trace: PyMutex<PyObjectRef>,
     state: PyMutex<FrameState>,
+    pub temperary_refs: PyMutex<Vec<PyObjectRef>>,
```
youknowone, line 114: Maybe we can reuse this for `del`.
I didn't know many fields in Frame are mutex. I thought they are single-thread accessible.

typo
```suggestion
    pub temporary_refs: PyMutex<Vec<PyObjectRef>>,
```

vm/src/object/core.rs
```diff
@@ -641,8 +642,15 @@ impl PyObject {
...
-        &self.0.typ
+    pub fn class(&self) -> &Py<PyType> {
+        self.0.typ.deref()
+    }
+
+    pub fn set_class(&self, typ: PyTypeRef, vm: &VirtualMachine) {
+        let old = unsafe { self.0.typ.swap(typ) };
+        if let Some(frame) = vm.current_frame() {
+            frame.temperary_refs.lock().push(old.into());
+        }
```
youknowone, line 653: If the swapped object is always expected to put in temporary refs,(I guess so)
frame or atomicref is better to have a safe interface to do the job.

vm/src/object/ext.rs
```diff
@@ -1,3 +1,5 @@
+use crossbeam_utils::atomic::AtomicCell;
+
```
youknowone, line 2:
```suggestion
```
let's not do this. I found everybody has different standard to make order of import sections

vm/src/object/ext.rs
```diff
@@ -224,6 +225,42 @@ impl<T: PyPayload> ToPyObject for PyRefExact<T> {
...
+}
+
+impl<T: PyObjectPayload> PyAtomicRef<T> {
+    /// # Safety
+    /// The caller is responsible to keep the returned PyRef alive
+    /// until no more reference can be used via PyAtomicRef::deref()
+    #[must_use]
+    pub unsafe fn swap(&self, pyref: PyRef<T>) -> PyRef<T> {
+        let py = PyRef::leak(pyref);
+        let old = self.0.swap(NonNull::from(py));
```
youknowone, line 259:
```suggestion
        let old = self.0.swap(NonNull::new_unchecked(py));
```
we already know py is not a null.

I am not sure possible but retaining `&'static mut T` will be good when available.

youknowone (approved):
Logically looking good to me.

vm/src/object/ext.rs
qingshi163: these two are equal. no check will be made as NonNull::from only accept valid value and it is calling new_unchecked inside the function

vm/src/frame.rs
qingshi163: If frame itself is executed single-threaded, we could replace mutex with an atomic increasing vec, do we have any crates doing that?

vm/src/object/ext.rs
qingshi163: Oh, that is done by rust-analyzer :)

vm/src/object/ext.rs
qingshi163: what do you mean about &'static mut T

vm/src/frame.rs
youknowone: I don't think we have those crate. When we create safety by human, `UnsafeCell` is the common tool

vm/src/object/ext.rs
```diff
@@ -212,6 +212,49 @@ impl<T: PyPayload> ToPyObject for PyRefExact<T> {
     }
 }
 
+pub struct PyAtomicRef<T: PyObjectPayload>(AtomicCell<NonNull<Py<T>>>);
```
youknowone, line 215: Will using `AtomicCell<NonNull>` spend same cost as `AtomicPtr`?

vm/src/object/ext.rs
youknowone: oh, got it. `py` is a reference, not a pointer.

I was thinking if this is possible.
```
pub struct PyAtomicRef<T: PyObjectPayload>(AtomicCell<&'static mut Py<T>>);
```

coolreader18: I think I've found over the past while, mainly from looking at crossbeam's issue list, that atomics aren't actually able to be hidden away enough such that AtomicCell would be a good abstraction. I'd honestly prefer a `PyAtomic<*mut Py<T>>`, which is a type alias that resolves to `std::sync::atomic::AtomicPtr` when threading is enabled, and `Cell<*mut _>` when it's not. I meant to go and replace all the AtomicCell uses with PyAtomic after I added PyAtomic, but I never got around to it.

vm/src/object/ext.rs
```diff
@@ -212,6 +212,49 @@ impl<T: PyPayload> ToPyObject for PyRefExact<T> {
...
+    pub unsafe fn swap(&self, pyref: PyRef<T>) -> PyRef<T> {
+        let py = PyRef::leak(pyref);
+        let old = self.0.swap(NonNull::from(py));
+        PyRef::from_raw(old.as_ptr())
+    }
+
+    pub fn swap_to_temporary_refs(&self, pyref: PyRef<T>, vm: &VirtualMachine) {
+        let old = unsafe { self.swap(pyref) };
+        if let Some(frame) = vm.current_frame() {
+            frame.temporary_refs.lock().push(old.into());
```
coolreader18, line 253: Does temporary_refs ever get cleared? Like I guess it doesn't need to be, necessarily, but I don't like the idea of
```py
class A: pass
class B: pass
B().__class__ = A
```
meaning that A will never ever ever be deallocated.

What if instead `class()` returned a guard.... no that wouldn't work. I'll try and think of something by the morning, maybe.

youknowone: I guess when the frame is popped

vm/src/object/ext.rs
qingshi163: [AtomicCell::<T>::is_lock_free()](https://docs.rs/crossbeam/latest/crossbeam/atomic/struct.AtomicCell.html#method.is_lock_free) should be true for the platforms that support AtomicPtr. But I am not sure about any overhead that AtomicCell does.

vm/src/object/ext.rs
coolreader18: Ohhh, I missed that its per-frame, yeah I guess that does make sense. Nevermind!

coolreader18:
(using PyAtomic instead of AtomicCell would fix the provenance errors in miri)

vm/src/object/ext.rs
qingshi163: > I guess when the frame is popped

Yes, we can clear the refs for each instruction, but I don't think it is worth.

qingshi163:
> (using PyAtomic instead of AtomicCell would fix the provenance errors in miri)

let me try

coolreader18:
Also, `vm.frames` should definitely be made private now, since it's now a safety thing that a frame cannot be popped off the frame stack + dropped before it's fully done running.

qingshi163:
> > (using PyAtomic instead of AtomicCell would fix the provenance errors in miri)
> 
> let me try

Don't know why but seem like failing the compile when threading is off, any mistake about how to use PyAtomic?
@coolreader18

qingshi163:
> > > (using PyAtomic instead of AtomicCell would fix the provenance errors in miri)
> > 
> > 
> > let me try
> 
> Don't know why but seem like failing the compile when threading is off, any mistake about how to use PyAtomic? @coolreader18

nevermind, it just the name conflict

qingshi163:
@coolreader18 @youknowone review?

youknowone (approved):
@qingshi163 looks good to me.
If @coolreader18 is too busy to review this PR, please feel free to merge it yourself.

Thank you for the awesome work!

Issue #3831, opened 2022-07-01
Fix beta clippy warnings


youknowone:
It will be stable warnings in 6 weeks

PR #3832, opened 2022-07-01
Fix rustfmt and clippy warnings

GoGim1:
related to #3831


youknowone (approved):
Thank you, this is missed due to broken CI.
I edited text body not to close the issue - because the issue is not about stable clippy.

cc @oow214

PR #3833, opened 2022-07-01
support cancel previous CI workflow

GoGim1:
Cancel previous workflows if they are the same workflow on the same ref (branch/tags)
with the same event (push/pull_request) even if they are in progress.
So that we don't waste time to verify a previous commit - what we're interested in is the latest view.


youknowone (approved):
Great, thank you!

PR #3838, opened 2022-07-03
Implement Number protocol for PyComplex

fanninpm:
Implementation based on #3507.

### DIFF
--- a/vm/src/builtins/complex.rs
+++ b/vm/src/builtins/complex.rs
@@ -429,8 +429,6 @@ impl AsNumber for PyComplex {
         multiply: Some(|number, other, vm| {
             Self::number_complex_op(number, other, |a, b| a * b, vm)
         }),
-        remainder: None,
-        divmod: None,
         power: Some(|number, other, vm| Self::number_general_op(number, other, inner_pow, vm)),
         negative: Some(|number, vm| {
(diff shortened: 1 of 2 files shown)


vm/src/builtins/complex.rs
```diff
@@ -419,6 +420,98 @@ impl Hashable for PyComplex {
...
+        inplace_remainder: None,
+        inplace_divmod: None,
+        inplace_power: None,
+        inplace_lshift: None,
+        inplace_rshift: None,
+        inplace_and: None,
+        inplace_xor: None,
+        inplace_or: None,
+        floor_divide: None,
+        true_divide: Some(|number, other, vm| {
```
qingshi163, line 465: all of the function witch is None could be replaced by
```
..PyNumberMethods::NOT_IMPLEMENTED
```
at the end of the block

qingshi163 (changes requested):
Looks great for the implement

Snowapril:
I think Implementing missing `PySequenceMethods`, `PyMappingMethods`, `PyNumberMethods` for some types can be good first issue for first-comer. How do you think?

youknowone (approved):

qingshi163 (approved):
Great!

PR #3845, opened 2022-07-08
Fix `str.join` (#3841)

### LINKED ISSUE #3841: `str.join` should not return new object if given argument is `iterable` with only one exact `str` type.
### CPython3.10
```python
>>> a1 = "abc"
>>> a2 = "test".join([a1])
>>> a1 is a2
True
```

### RustPython
```python
>>>>> a1 = "abc"
>>>>> a2 = "test".join([a1])
>>>>> a1 is a2
False
```

If the given `iterable` argument has only one item and its type is the exact `str` type, then `str.join` should return that one right away.

### Reference
https://github.com/python/cpython/blob/3.10/Objects/unicodeobject.c#L10383-L10577

jopemachine:
Fix #3841.

### DIFF
--- a/vm/src/builtins/str.rs
+++ b/vm/src/builtins/str.rs
@@ -928,9 +928,13 @@ impl PyStr {
     }
 
     #[pymethod]
-    fn join(&self, iterable: ArgIterable<PyStrRef>, vm: &VirtualMachine) -> PyResult<String> {
-        let iter = iterable.iter(vm)?;
-        self.as_str().py_join(iter)
+    fn join(&self, iterable: ArgIterable<PyStrRef>, vm: &VirtualMachine) -> PyResult<PyStrRef> {
+        let mut iter = iterable.iter(vm)?;
+
+        match iter.size_hint().0 {
+          1 => Ok(iter.next().unwrap().unwrap()),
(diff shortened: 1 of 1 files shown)


vm/src/builtins/str.rs
```diff
@@ -928,9 +928,13 @@ impl PyStr {
     }
 
     #[pymethod]
-    fn join(&self, iterable: ArgIterable<PyStrRef>, vm: &VirtualMachine) -> PyResult<String> {
-        let iter = iterable.iter(vm)?;
-        self.as_str().py_join(iter)
+    fn join(&self, iterable: ArgIterable<PyStrRef>, vm: &VirtualMachine) -> PyResult<PyStrRef> {
+        let mut iter = iterable.iter(vm)?;
+
+        match iter.size_hint().0 {
```
youknowone, line 934: `size_hint` is not a length itself. I think retrieving all items to a `Vec` and checking the size is the correct way.

youknowone (changes requested):

Snowapril:
You can implement this by using **Sequence Protocol**. With use of `length` for getting length, `extract` for getting items could be way. Please refer on `vm/src/protocol/sequence.rs`

youknowone (approved):
thank you for contributing!

PR #3855, opened 2022-07-10
Update traceback

youknowone:


youknowone:
dependency of #3862

fanninpm (approved):

fanninpm:
I think we should wait for CI to finish before merging.

___

On a tangential note, do you want to add [Bors](https://github.com/bors-ng/bors-ng) to this repository?

youknowone:
huh... the result of running `cargo run Lib/test/test_traceback.py` and `cargo run -- -m test test_traceback` are different.

fanninpm:
> huh... the result of running `cargo run Lib/test/test_traceback.py` and `cargo run -- -m test test_traceback` are different.

Hence why I made #3807.

youknowone:
I though both of running should pass. Does former one also fail in CPython?

PR #3863, opened 2022-07-10
Add modulefinder from CPython v3.10.5

youknowone:


youknowone:
not worth to keep open

PR #3864, opened 2022-07-10
Fix union args

TwoPair:
I'm focusing on fixing `test_types::UnionTests::test_union_args`.
Modifying or Adding operator or(|) function for python's `Lib/typing.py`.

It fails, however, when the very last check procedures proceed.
The problem is `NoneType` has no `__args__` function.
I tried making that function, but I cannot guess how to do it
cuz I completely don't know how `PyNone` struct works.

Anyway I got pass at least for the previous check routines.
Thx.

### DIFF
--- a/vm/src/builtins/singletons.rs
+++ b/vm/src/builtins/singletons.rs
@@ -1,5 +1,6 @@
-use super::{PyType, PyTypeRef};
+use super::{PyType, PyTypeRef, union_};
 use crate::{
+    builtins::PyTuple,
     class::PyClassImpl, convert::ToPyObject, types::Constructor, Context, Py, PyObjectRef,
     PyPayload, PyResult, VirtualMachine,
 };
@@ -46,9 +47,29 @@ impl PyNone {
         "None".to_owned()
     }
 
+    // NoneType has no __args__ function,
+    // so it fails that an union type includes 'None'
+    // in "test_types::UnionTests::test_union_args".
+
+    // #[pyproperty(magic)]
+    // fn args(&self) -> PyObjectRef {
+    //     self.args.clone().into()
+    // }
+
     #[pymethod(magic)]
     fn bool(&self) -> bool {
         false
+    }
+
+    #[pymethod(name = "__ror__")]
+    #[pymethod(magic)]
+    fn or(zelf: PyObjectRef, other: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
+        if !union_::is_unionable(zelf.clone(), vm) || !union_::is_unionable(other.clone(), vm) {
+            return vm.ctx.not_implemented();
+        }
+
+        let tuple = PyTuple::new_ref(vec![zelf, other], &vm.ctx);
+        union_::make_union(tuple, vm)
     }
 }
 
--- a/vm/src/builtins/singletons.rs
+++ b/vm/src/builtins/singletons.rs
@@ -1,6 +1,5 @@
-use super::{PyType, PyTypeRef, union_};
+use super::{PyType, PyTypeRef};
 use crate::{
-    builtins::PyTuple,
     class::PyClassImpl, convert::ToPyObject, types::Constructor, Context, Py, PyObjectRef,
     PyPayload, PyResult, VirtualMachine,
 };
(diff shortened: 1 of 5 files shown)


vm/src/builtins/singletons.rs
```diff
@@ -46,10 +47,30 @@ impl PyNone {
...
+    #[pymethod(name = "__ror__")]
+    #[pymethod(magic)]
+    fn or(zelf: PyObjectRef, other: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
+        if !union_::is_unionable(zelf.clone(), vm) || !union_::is_unionable(other.clone(), vm) {
+            return vm.ctx.not_implemented();
+        }
+
+        let tuple = PyTuple::new_ref(vec![zelf, other], &vm.ctx);
+        union_::make_union(tuple, vm)
+    }
```
youknowone, line 73: `None` doesn't implement `__or__` or `__ror__`.
The other types (`union` in this case) have to implement `__ror__` to handle `None` arguement.

youknowone (commented):
We shouldn't add anything to `None` type to solve this problem.
Before working on `None` type, try to check in interactive shell if `None` have those methods.

I didn't look in this issue deep enough, but I am pretty sure you can achieve `None` handling by only editing `type` and `union`'s `__ror__`.
(and `__or__` for union to union operation)

If you are not understanding the relationship between `__or__` and `__ror__`, check this article: https://stackoverflow.com/a/25211911

Snowapril:
As `is_unionable` makes decision on `type.__or__`, `type.__ror__`, we need to focus on it.
We need to make `is_unionable` to check whether the given arg is `None`, not `none_type` as none_type will be passed by `obj.payload_if_subclass::<PyType>(vm).is_some()` (checks `vm.ctx.none`.)
In the flattening procedure, while iterating the args we need to check the given arg is None and convert it to none_type (checks `vm.ctx.types.none_type`.

Snowapril:
Please do not use git merge. Would you try to rebase it?

TwoPair:
Complete to rebase, and complete to resolve `test_types::UnionTests::test_union_args`!

Snowapril:
It seems `test_types::test_union_args` is passed by your change.

vm/src/builtins/union.rs
```diff
@@ -100,11 +100,22 @@ impl PyUnion {
...
+    #[pymethod(name = "__ror__")]
+    #[pymethod(magic)]
+    fn or(zelf: PyObjectRef, other: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
+        if !is_unionable(zelf.clone(), vm) || !is_unionable(other.clone(), vm) {
+            return vm.ctx.not_implemented();
+        }
+
+        let tuple = PyTuple::new_ref(vec![zelf, other], &vm.ctx);
+        make_union(tuple, vm)
+    }
```
Snowapril, line 113: It seems `union`, `genericalias` and `type` have duplicated implementation for `__or__` and `__ror__`.

vm/src/builtins/union.rs
```diff
@@ -156,6 +167,8 @@ fn flatten_args(args: PyTupleRef, vm: &VirtualMachine) -> PyTupleRef {
     for arg in &args {
         if let Some(pyref) = arg.downcast_ref::<PyUnion>() {
             flattened_args.extend(pyref.args.iter().cloned());
+        } else if arg.payload_if_subclass::<PyNone>(vm).is_some() {
```
youknowone, line 170:
```suggestion
        } else if vm.is_none(arg) {
```
`None` is singleton in python. we check its id instead of its type.

youknowone (changes requested):

vm/src/builtins/type.rs
```diff
@@ -403,15 +403,14 @@ impl PyType {
...
 
-    #[pymethod(name = "__ror__")]
     #[pymethod(magic)]
-    pub fn or(zelf: PyObjectRef, other: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
-        if !union_::is_unionable(zelf.clone(), vm) || !union_::is_unionable(other.clone(), vm) {
-            return vm.ctx.not_implemented();
-        }
+    pub fn ror(zelf: PyObjectRef, other: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
+        _or(other, zelf, vm)
+    }
```
youknowone, line 409: when ror and or share same implementation, please keep the single function share both method like original code.

vm/src/builtins/type.rs
```diff
@@ -826,6 +825,15 @@ pub(crate) fn call_slot_new(
     unreachable!("Should be able to find a new slot somewhere in the mro")
 }
 
+pub(super) fn _or(zelf: PyObjectRef, other: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
```
youknowone, line 828: `_` prefix means unused variable. please remove it.

youknowone (changes requested):
because you fixed the bug, the marker `expectedFailure` on the test also must be removed. check the test report with result `unexpected success`

vm/src/builtins/type.rs
TwoPair: If or, and ror in type.rs were placed in a same function, the validation of union's order couldn't be guaranteed.
I've already tested and checked the test was failed where they're combined.
That's why I make them just like that.

youknowone: Oh, you are right! I didn't catch that point.

vm/src/builtins/union.rs
```diff
@@ -156,6 +162,8 @@ fn flatten_args(args: PyTupleRef, vm: &VirtualMachine) -> PyTupleRef {
     for arg in &args {
         if let Some(pyref) = arg.downcast_ref::<PyUnion>() {
             flattened_args.extend(pyref.args.iter().cloned());
+        } else if vm.is_none(arg) {
+            flattened_args.push(vm.ctx.types.none_type.as_ref().to_pyobject(vm).clone());
```
youknowone, line 166:
```suggestion
            flattened_args.push(vm.ctx.types.none_type.to_owned().into());
```
this is the last piece

youknowone (approved):
thank you for contributing!

PR #3867, opened 2022-07-11
Add attribute to SyntaxError

key262yek:
Add attributes 'end_lineno' and 'end_offset' to SyntaxError

### DIFF
--- a/Lib/test/test_traceback.py
+++ b/Lib/test/test_traceback.py
@@ -4,10 +4,11 @@ from collections import namedtuple
 from io import StringIO
 import linecache
 import sys
+import inspect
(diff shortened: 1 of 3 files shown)


fanninpm:
Is this superseded by #3870?

key262yek:
I'm sorry, I should close this pr.
I tried to fix error which occurs while traceback update, and the first thing I did is add attribute to syntax error.
But, when I check it deeply, I realized the new attributes are related to highlight. 
So I roll-back traceback update and change only about syntax error in #3870.

I didn't realize that the #3870 is merged. 
Because highlight syntax update has barely done.

PR #3868, opened 2022-07-11
add arrayiter.__reduce__

maong0927:


stdlib/src/array.rs
```diff
@@ -1258,9 +1258,11 @@ mod array {
 
     impl Iterable for PyArray {
         fn iter(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult {
+            let zelf_ = zelf.clone();
```
youknowone, line 1261: `_` prefix usually means unused variable in rust code.

stdlib/src/array.rs
```diff
@@ -1258,9 +1258,11 @@ mod array {
...
             Ok(PyArrayIter {
                 position: AtomicUsize::new(0),
                 array: zelf,
```
youknowone, line 1264:
```suggestion
                array: zelf.clone(),
```
change here to allow to use zelf later.

stdlib/src/array.rs
Snowapril, line 1261:
```suggestion
```

youknowone (changes requested):

stdlib/src/array.rs
```diff
@@ -1258,9 +1258,11 @@ mod array {
...
+                internal: PyMutex::new(PositionIterInternal::new(zelf_, 0)),
```
Snowapril, line 1265:
```suggestion
                internal: PyMutex::new(PositionIterInternal::new(zelf.clone(), 0)),
```
Is `PositionIterInternal` necessary for implementing `__reduce__`?

stdlib/src/array.rs
fanninpm: The `_` _suffix_ usually indicates that you're dodging the use of a reserved keyword. From the [Rust style guide](https://github.com/rust-dev-tools/fmt-rfcs/blob/master/guide/advice.md):

> When a name is forbidden because it is a reserved word (e.g., `crate`), use a trailing underscore to make the name legal (e.g., `crate_`), or use raw identifiers if possible.

Snowapril:
Please do not use git merge. Would you try to rebase it? 
https://docs.gitlab.com/ee/topics/git/git_rebase.html

stdlib/src/array.rs
```diff
@@ -1281,10 +1282,22 @@ mod array {
     pub struct PyArrayIter {
         position: AtomicUsize,
         array: PyArrayRef,
+        internal: PyMutex<PositionIterInternal<PyArrayRef>>,
```
youknowone, line 1285: do you really need both `position` and `internal`? Though I didn't look in details, those fields look like duplication.

Yaminyam: The PositionIterInternal was used by referring to the implementation of the list iter.

stdlib/src/array.rs
youknowone: I think @Snowapril got confused here. because `zelf` is already moved in upper line, it can't be clone here.

fanninpm:
> Please do not use git merge. Would you try to rebase it?  
> https://docs.gitlab.com/ee/topics/git/git_rebase.html

The following commands should get you started:

```text
git switch main
git pull upstream main
git checkout arrayiter_reduce
git rebase -i main
```

stdlib/src/array.rs
Yaminyam: Is it because `internal` contains position because it is `PositionIterInternal` as its name suggests?

youknowone:
because people sometimes use `main` as a working branch, I like to recommend to use `upstream/main` directly

```
git fetch upstream
git rebase -i upstream/main
```

maong0927:
I'll use rebase from now on

stdlib/src/array.rs
youknowone: Check the `PositionIterInternal` type. it contains `position` field. By this change, the code have duplicated position fields but most of code uses position and new code uses internal. That can't work correctly.

Snowapril: With comparison with other types' iterator, we really don't need position and array field because `PyMutex<PositionIterInternal<PyArrayRef>>` will do their roles. Because of `__next__` and `__setstate__`, we need to modify `position` and `array` fields. It leads to the need for `PyMutex`.

Snowapril: If we replace `position` and `array` fields with `PyMutex<PositionIterInternal<PyArrayRef>>`, we can implement other methods as usual we did (`__next__` and `__reduce__`)

youknowone: Does arrayiterator require `closed` field? I thought it only need `position`.
`closed` is useful when it include another iterator in it.

stdlib/src/array.rs
```diff
@@ -1279,37 +1277,31 @@ mod array {
...
-                PyIterReturn::StopIteration(None)
-            };
-            Ok(r)
+			zelf.internal.lock().next(|array, pos| {
+            	let r = if let Some(item) = array.read().get(pos, vm) {
+                	PyIterReturn::Return(item?)
+            	} else {
+                	PyIterReturn::StopIteration(None)
+            	};
+            	Ok(r)
```
Snowapril, line 1303:
```suggestion
            	Ok(match array.read().get(pos, vm) {
                    Some(item) => PyIterReturn::Return(item?),
                    None => PyIterReturn::StopIteration(None),
                })
```

youknowone (approved):
thank you for contributing!

PR #3869, opened 2022-07-11
Fix define_exception! of PyStopIteration to set value of value attribute

tgsong827:


youknowone (changes requested):
Hi! Thank you for contributing!
I think you made the correct change for StopIteration itself.

The one you might miss is the other side - the user of StopIteration. Check this link:
https://github.com/RustPython/RustPython/blob/main/vm/src/protocol/iter.rs#L153-L176
It is retrieving the value with `.get_arg(0)`, the first argument, instead of `value`.

Could you also fix it to get value from `value` attribute? Then everything will work!

youknowone:
I found this change doesn't fit to our `new_stop_exception` implementation. Could you please also fix that part?

https://github.com/RustPython/RustPython/blob/main/vm/src/vm/vm_new.rs#L267-L274

For now, `new_stop_exception` only call `BaseException.__new__` but we also need to call `__init__` to run your new `value` design.

vm/src/vm/vm_new.rs
```diff
@@ -270,7 +270,11 @@ impl VirtualMachine {
         } else {
             Vec::new()
         };
-        self.new_exception(self.ctx.exceptions.stop_iteration.to_owned(), args)
+        let exc = self.new_exception(self.ctx.exceptions.stop_iteration.to_owned(), args.clone());
+        exc.as_object()
+            .set_attr("value", self.unwrap_or_none(args.get(0).cloned()), self)
+            .unwrap();
```
youknowone, line 276: We dont use unwrap for python errors. Try ? operator please

vm/src/vm/vm_new.rs
```diff
@@ -270,7 +270,11 @@ impl VirtualMachine {
         } else {
             Vec::new()
         };
-        self.new_exception(self.ctx.exceptions.stop_iteration.to_owned(), args)
+        let exc = self.new_exception(self.ctx.exceptions.stop_iteration.to_owned(), args.clone());
+        exc.as_object()
```
youknowone, line 274: I think this is conceptionally calling __init__. I like this approach for performance, but please leave a comment this is intending calling init.
Actually, creating a shared function and calling it would be perfect

youknowone (changes requested):

vm/src/vm/vm_new.rs
tgsong827: Thank you for comments!
I tried to use `?` operator behind `set_attr()`, but it didn't work.
I got an error message. 
"The return type is not implemented for `core::PyRef<exceptions::types::PyBaseException>`"

vm/src/vm/vm_new.rs
tgsong827: Thank you for comments!
I created a `set_exception_attr()` function and made `new_stop_iteration` to call it to initialize `value` attribute.

vm/src/vm/vm_new.rs
Snowapril:
```rust
match name_error.as_object().set_attr("name", name, self) {
    Ok(_) => name_error,
    Err(e) => e,
}
```
We can avoid `unwrap` like this because `Err` contains PyRef<PyBaseException>`.
But it seems there are several methods use `unwrap`. Maybe we can make some inline method for this kind of actions

tgsong827: Thank you!
I updated the code and It works.

vm/src/vm/vm_new.rs
```diff
@@ -312,4 +314,19 @@ impl VirtualMachine {
...
+        attr_value: Option<PyObjectRef>,
+    ) -> PyBaseExceptionRef {
+        match exc
+            .as_object()
+            .set_attr(attr_name, self.unwrap_or_none(attr_value), self)
+        {
+            Ok(_) => exc,
+            Err(e) => e,
+        }
+    }
```
youknowone, line 331: This function is looking useful in `exceptions.rs`, but I didn't intended this way.

vm/src/vm/vm_new.rs
```diff
@@ -270,7 +270,9 @@ impl VirtualMachine {
         } else {
             Vec::new()
         };
-        self.new_exception(self.ctx.exceptions.stop_iteration.to_owned(), args)
+        let exc = self.new_exception(self.ctx.exceptions.stop_iteration.to_owned(), args.clone());
+        // To initialize the `value` attribute to first argument
+        self.set_exception_attr(exc, "value", args.get(0).cloned())
```
youknowone, line 275: I wanted this code is showing it is same to `StopIteration.__init__`, but it still doesn't look like it.

youknowone (commented):
I will attach fix

youknowone (approved):
I got a little bit more idea about new_stop_iteration while working on it. I would not be able to find it myself without this PR.

Thank you for contributing!

tgsong827:
> 

Maybe I misunderstood what you said.
i'm trying to understand the fix that you attached.

Thank you for helping me in this PR.

> I got a little bit more idea about new_stop_iteration while working on it. I would not be able to find it myself without this PR.
> 
> Thank you for contributing!

PR #3879, opened 2022-07-14
Fix failure test_assign_del() in Lib/test/test_syntax.py 

### LINKED ISSUE #3878: cannot delete list directly `del [x]`
## Feature

In CPython, they can delete a list in following manner

```python3
>>> x = 1
>>> del [x]
```

But, in RustPython we cannot.
```python3
>>> del [x]
SyntaxError: cannot delete list at line 1 column 1
del [x]
^
```

Related test is in `Lib/test/test_syntax.py`.
```python3
def test_assign_del(self):
        # ...
        self._check_error("del [*x,]", "cannot delete starred")    # Not "cannot delete list"
```

## Python Documentation

Cannot find documentation exactly related this difference.

### LINKED ISSUE #3882: Statement parsing algorithm should be change
## Feature

In CPython grammar (Grammar/python.gram) line 695, there are group expression type exists.
```gram
atom[expr_ty]:
    ...
     | &'(' (tuple | group | genexp)
```

The definition of group is at line 709 of same file
```gram
group[expr_ty]:
    | '(' a=(yield_expr | named_expression) ')' { a }
    | invalid_group
```
and of invalid group is at line 947
```gram
invalid_group:
    | '(' a=starred_expression ')' {
        RAISE_SYNTAX_ERROR_KNOWN_LOCATION(a, "cannot use starred expression here") }
    | '(' a='**' expression ')' {
        RAISE_SYNTAX_ERROR_KNOWN_LOCATION(a, "cannot use double starred expression here") }
```

Here, the first pattern in invalid_group is related to SyntaxError of `del (*x)` (pr #3879)
It should print **cannot use starred expression here**, but without group expression it makes **cannot delete starred**.

In parser/src/python.lalrpop at line 956, only tuple, yield, genexp are parsed. 
Even, parenthesized-form is missing. 

I think, we need to make group expression to parse parenthesized-form and handle error about starred epression,
 or add equivalent parsing grammar into atom parser

## Python Documentation

Maybe those are related things. 
[atom](https://docs.python.org/3/reference/expressions.html#atoms)
[parenthesized-forms](https://docs.python.org/3/reference/expressions.html#parenthesized-forms)

key262yek:
1. SyntaxError msg of del statement is changed from `can't delete` to `cannot delete`
2. del operator now can handle list directly `del [x]`
3. raise syntax error for cases `(*x)` and `(**x)` before pass them into del statement
4. change error message to "cannot delete expression" for binary operators and unary operators. 

fix #3882 
fix #3878


key262yek:
Until the commit 532de56, commits are related to pr #3873

fanninpm:
It seems that bb2aa27 ("Copy Lib/test/test_cmd_line_script.py from CPython v3.10.5") and 3f7fdf4 ("Update Lib/test/test_cmd_line_script.py to be runable") cancel each other out (with the exception of extra blank lines). The next time you rebase this PR, would you be open to dropping those two commits?

fanninpm:
Did you do a `cargo build` before you committed the change to the LALRPOP file?

key262yek:
@fanninpm I missed that point. After this commit, I will upload result of build too.

youknowone:
The commit message from third commit (https://github.com/RustPython/RustPython/pull/3879/commits/02f51d450cb86227d42b9ce090ace52051300917) includes unexisting commit id. Actually it seems not a revert commit. Could you fix the mssage?

key262yek:
fix #3878

key262yek:
@youknowone  I changed the message.

PR #3880, opened 2022-07-14
Implement Number protocol for PyNone

hy-kiera:
Implemented based on #3838

### DIFF
--- a/vm/src/builtins/singletons.rs
+++ b/vm/src/builtins/singletons.rs
@@ -1,7 +1,7 @@
 use super::{PyType, PyTypeRef};
 use crate::{
-    class::PyClassImpl, convert::ToPyObject, types::Constructor, Context, Py, PyObjectRef,
-    PyPayload, PyResult, VirtualMachine,
+    class::PyClassImpl, convert::ToPyObject, protocol::PyNumberMethods, types::AsNumber,
+    types::Constructor, Context, Py, PyObjectRef, PyPayload, PyResult, VirtualMachine,
 };
 
 #[pyclass(module = false, name = "NoneType")]
@@ -39,7 +39,7 @@ impl Constructor for PyNone {
     }
 }
 
-#[pyimpl(with(Constructor))]
+#[pyimpl(with(Constructor, AsNumber))]
 impl PyNone {
     #[pymethod(magic)]
     fn repr(&self) -> String {
@@ -50,6 +50,13 @@ impl PyNone {
     fn bool(&self) -> bool {
         false
     }
+}
+
+impl AsNumber for PyNone {
+    const AS_NUMBER: PyNumberMethods = PyNumberMethods {
+        boolean: Some(|number, _vm| Ok(false)),
+        ..PyNumberMethods::NOT_IMPLEMENTED
+    };
 }
 
 #[pyclass(module = false, name = "NotImplementedType")]
--- a/vm/src/builtins/singletons.rs
+++ b/vm/src/builtins/singletons.rs
@@ -54,7 +54,7 @@ impl PyNone {
 
 impl AsNumber for PyNone {
     const AS_NUMBER: PyNumberMethods = PyNumberMethods {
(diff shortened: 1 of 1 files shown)


vm/src/builtins/singletons.rs
```diff
@@ -52,6 +54,13 @@ impl PyNone {
     }
 }
 
+impl AsNumber for PyNone {
+    const AS_NUMBER: PyNumberMethods = PyNumberMethods {
+        boolean: Some(|number, _vm| Ok(self::number_downcast(number).value.is_zero()))
```
youknowone, line 59: is_zero? the value looking swapped

hy-kiera: `non_bool()`, which is of `none_as_number`, returns 0 in cpython `Objects/object.c`
```
static int
none_bool(PyObject *v)
{
    return 0;
}
```

Snowapril: return zero means false in this case. As our `bool` returns false as you know, we just need to return what `bool` returns exactly.

youknowone: oh, sorry. I was confused with bool. 
as you said, it returns false(0). But this code is not simply returning false. make this line to do return false same to the code you shared.

hy-kiera: Hi. I fixed my code. Could you please review the new commit?

vm/src/builtins/singletons.rs
```diff
@@ -39,7 +41,7 @@ impl Constructor for PyNone {
     }
 }
 
-#[pyimpl(with(Constructor))]
+#[pyimpl(flags(BASETYPE), with(Constructor, AsNumber))]
```
Snowapril, line 44:
```suggestion
#[pyimpl(with(Constructor, AsNumber))]
```
It seems PyNone does not have `BASETYPE` flags in cpython. Is there any reason for adding it?

Snowapril:
Please run `cargo fmt --all` 😊

hy-kiera:
> Please run `cargo fmt --all` 😊

Okay. I'll run it.

vm/src/builtins/singletons.rs
```diff
@@ -52,6 +52,13 @@ impl PyNone {
     }
 }
 
+impl AsNumber for PyNone {
+    const AS_NUMBER: PyNumberMethods = PyNumberMethods {
+        boolean: Some(|number, _vm| Ok(Self::number_downcast(number).bool())),
```
youknowone, line 57:
```suggestion
        boolean: Some(|number, _vm| Ok(false)),
```
None is singleton. I am pretty sure you can skip actual object value.

hy-kiera: Good idea. BTW, what is the object's functions such as `repr()` and `bool()` used be for?

Snowapril: They are exposed as python method by #[pymethod] macro attached. You can call them by below way.
```python
>>>>> None.__bool__()
False
>>>>> None.__repr__()
'None'
```

youknowone: And those functions @Snowapril referred are used by facade functions `bool(None)` and `repr(None)`

vm/src/builtins/singletons.rs
```diff
@@ -1,7 +1,7 @@
 use super::{PyType, PyTypeRef};
 use crate::{
-    class::PyClassImpl, convert::ToPyObject, types::Constructor, Context, Py, PyObjectRef,
-    PyPayload, PyResult, VirtualMachine,
+    class::PyClassImpl, convert::ToPyObject, protocol::PyNumberMethods, types::AsNumber,
+    types::Constructor, Context, Py, PyObjectRef, PyPayload, PyResult, VirtualMachine,
```
youknowone, line 4: you can also group type traits
```rust
types::{AsNumber, Constructor}
```

qingshi163 (approved):
Looks great

PR #3884, opened 2022-07-14
fix: non-utf8 source causes panic, close #3798

### LINKED ISSUE #3798: non-utf8 source causes panic
Expectation:
return error properly instead of panic

Actual
```
thread 'main' panicked at 'range end index 206 out of range for slice of length 89', /home/szx5097/snap/rustup/common/rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:73:5
stack backtrace:
   0:     0x55c425c92d9f - std::backtrace_rs::backtrace::trace_unsynchronized::h7fefb4efbec818a2
   1:     0x55c425d52cd7 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::hd72840bfc06c2d3b
   2:     0x55c425dac297 - core::fmt::write::hdc2a0f6fbc4b6426
   3:     0x55c425c9ef55 - std::io::Write::write_fmt::hca7f516bdcc75b19
   4:     0x55c425d52578 - std::sys_common::backtrace::print::h72da513265a2a259
   5:     0x55c425cd28c6 - std::panicking::default_hook::{{closure}}::hda7ab18d70dbbc8e
   6:     0x55c425cd20bc - std::panicking::default_hook::hb7acfdbd1612d90a
   7:     0x55c425cd4460 - std::panicking::rust_panic_with_hook::hf266b347c72dd399
   8:     0x55c425d5456e - std::panicking::begin_panic_handler::{{closure}}::hbb6db86789726dc9
   9:     0x55c425d53ac9 - std::sys_common::backtrace::__rust_end_short_backtrace::h9266edc5c7e7f332
  10:     0x55c425cd399c - rust_begin_unwind
  11:     0x55c4245a77d2 - core::panicking::panic_fmt::h45f58e17be61496b
  12:     0x55c425dc7de1 - core::slice::index::slice_end_index_len_fail_rt::h62e8e3323f29297a
  13:     0x55c425db22bb - core::ops::function::FnOnce::call_once::h43377be9ffec5c4e
  14:     0x55c425db359a - core::intrinsics::const_eval_select::h4067aa1a3bb1eab2
  15:     0x55c4245a824a - core::slice::index::slice_end_index_len_fail::h6da3fc8c55fffadd
  16:     0x55c425a55b76 - unicode_names2::character::h6eec0819f2821f13
  17:     0x55c42560030a - rustpython_parser::lexer::Lexer<T>::lex_string::hbf5da7abb263896b
  18:     0x55c4256092be - rustpython_parser::lexer::Lexer<T>::inner_next::h21b0fee2d80ffc79
  19:     0x55c4256125b8 - <rustpython_parser::lexer::Lexer<T> as core::iter::traits::iterator::Iterator>::next::ha771fcc27ad382ff
  20:     0x55c425a48f6f - lalrpop_util::state_machine::Parser<D,I>::drive::h5b523e0deba625cb
  21:     0x55c425657c88 - rustpython_parser::python::__parse__Top::TopParser::parse::h82d49a421b6830e0
  22:     0x55c4255efdcd - rustpython_parser::parser::parse::he1051afea0331df3
  23:     0x55c4255a40e9 - rustpython_compiler::compile::h42b38299785169e4
  24:     0x55c4248578b1 - rustpython_vm::vm::compile::<impl rustpython_vm::vm::VirtualMachine>::compile::h0d9f03d2fced5b4d
  25:     0x55c4246522e2 - <core::panic::unwind_safe::AssertUnwindSafe<F> as core::ops::function::FnOnce<()>>::call_once::hf4c5d08b4be7fd08
  26:     0x55c424646e37 - std::panicking::try::do_call::h56b7810a1617f35f
  27:     0x55c42464882d - __rust_try
  28:     0x55c424646c8f - std::panicking::try::h279451504198543e
  29:     0x55c424649890 - std::thread::local::LocalKey<T>::with::hc96cce0e84a74de0
  30:     0x55c4246469c8 - rustpython_vm::vm::thread::enter_vm::h24984e19905c70e5
  31:     0x55c42464850c - rustpythonbug::main::h3537c39100935ca5
  32:     0x55c424646a53 - std::sys_common::backtrace::__rust_begin_short_backtrace::h2e161759387fb3ae
  33:     0x55c4246494d9 - std::rt::lang_start::{{closure}}::hf8f913bc7047be2b
  45:                0x0 - <unknown>

```

Reproduce:

```rust
use std::env;
use rustpython_vm as vm;

    let args: Vec<String> = env::args().collect();

    let filepath = &args[1];
    let data = std::fs::read(filepath).unwrap();
        let code = String::from_utf8_lossy(&data).into_owned();

        let code_obj = vm
            .compile(&code, vm::compile::Mode::Exec, "<embedded>".to_owned())
            .map_err(|err| vm.new_syntax_error(&err))?;


        let res: vm::PyResult<()> = Ok(());
        return res;
    });
    Ok(())
}
```

Cargo.toml
```
[dependencies]
rustpython-vm = { git = "https://github.com/RustPython/RustPython.git", rev = "ff502134dbb9dd3bfc33331308de255ebefcabf2"}
```

Input:
[in.zip](https://github.com/RustPython/RustPython/files/8931027/in.zip)

killme2008:
Try to fix #3798 .

I found that it's panic at `unicode_name2:character` : https://github.com/progval/unicode_names2/blob/master/src/lib.rs#L323

```rust
  let mut buf = [0u8; MAX_NAME_LENGTH + 1];
    for (place, byte) in buf.iter_mut().zip(name.bytes()) {
        *place = ASCII_UPPER_MAP[byte as usize]
    }
    let search_name = &buf[..name.len()];
```

The unicode name's length exceeds `MAX_NAME_LENGTH`:

```rust
pub const MAX_NAME_LENGTH: usize = 88;
```

The patch just added a checking before calling `unicode_name2:character`. But `MAX_NAME_LENGTH` is not public in `unicode_name2`, so i added a const to `lexer.rs`:

```
const MAX_UNICODE_NAME: usize = 88;
```


killme2008:
The `extra_tests/snippets/non_utf8.txt` is in fact a text file taken from the issue example but github recognize it as binary.

```
$ cat extra_tests/snippets/non_utf8.txt
" ÄˇÄ%\N{\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\222{{˘"Äˇ›"Ûˇˇ$"ˇÄ%\N{222+-. ÄˇÄ%\N{222{{˘"Äˇ›
```

extra_tests/snippets/syntax_non_utf8.py
```diff
@@ -0,0 +1,11 @@
...
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+try:
+    with open(os.path.join(dir_path , "non_utf8.txt")) as f:
+        eval(f.read())
+except:
+    # TODO: RUSTPYTHON, rustpython raise SyntaxError but cpython raise ValueError here.
+    pass
+else:
+  assert False
```
fanninpm, line 11:
```suggestion
# TODO: RUSTPYTHON, RustPython raises a SyntaxError here
with assert_raises(ValueError):
    with open(os.path.join(dir_path , "non_utf8.txt")) as f:
        eval(f.read())
```

extra_tests/snippets/syntax_non_utf8.py
```diff
@@ -0,0 +1,11 @@
+import os
+dir_path = os.path.dirname(os.path.realpath(__file__))
```
fanninpm, line 2:
```suggestion
import os

from testutils import assert_raises

dir_path = os.path.dirname(os.path.realpath(__file__))
```

fanninpm (commented):
We can use the `assert_raises` context manager from the `testutils` module[^1]. (My suggestion will cause RustPython to fail the test because it raises the wrong exception.)

[^1]: https://github.com/RustPython/RustPython/blob/main/extra_tests/snippets/testutils.py

parser/src/lexer.rs
```diff
@@ -145,6 +145,8 @@ where
     }
 }
 
+const MAX_UNICODE_NAME: usize = 88;
```
fanninpm, line 148: In the future, someone browsing the codebase might come across this line and wonder why it's there. Perhaps a comment can help to clarify things. Something like:

```rust
// unicode_name2 does not expose `MAX_UNICODE_NAME`, so we replicate that constant here
```

extra_tests/snippets/syntax_non_utf8.py
killme2008: It's not working, becase cpython raise a `ValueError` but rustpython raise a `SyntaxError`, then the test will fail.

The snippet tests will run both on rustpython and cpython.

killme2008: I tried another fix by checking python platform implementation:

```python
import os
import platform

from testutils import assert_raises

dir_path = os.path.dirname(os.path.realpath(__file__))

# TODO: RUSTPYTHON, RustPython raises a SyntaxError here, but cpython raise a ValueError
error = SyntaxError if platform.python_implementation() == 'RustPython' else ValueError
with assert_raises(error):
    with open(os.path.join(dir_path , "non_utf8.txt")) as f:
        eval(f.read())

```

youknowone (approved):
nice catch and fix, thank you!

PR #3889, opened 2022-07-15
Fix testEnterAttributeError2 in test_with.py

daeun503:


vm/src/frame.rs
```diff
@@ -775,6 +775,7 @@ impl ExecutingFrame<'_> {
             }
             bytecode::Instruction::SetupWith { end } => {
                 let context_manager = self.pop_value();
+                let _enter = context_manager.get_attr(identifier!(vm, __enter__), vm)?;
```
Snowapril, line 778: It seems `get_attr` for `__enter__` is duplicated with below `vm.call_special_method(context_manager, identifier!(vm, __enter__), ())?;`. You can replace `call_special_method` with `get_attr` and `invoke` method. Like this
```rust
let enter = context_manager.get_attr(identifier!(vm, __enter__), vm)?;
let enter_res = vm.invoke(&enter, ())?;
```

Snowapril: But it seems just place `let enter_res =
                    vm.call_special_method(context_manager, identifier!(vm, __enter__), ())?;` to here is sufficient.

daeun503: I tried `let enter_res = vm.call_special_method(context_manager, identifier!(vm, __enter__), ())?;`
But the code doesn't work because  the ownership of  `context_manager` goes out of scope.
`&context_manager` is also same.. so... I'll try to fix `get_attr` & `invoke`

daeun503: I'll try according to your advice. Thank you !

daeun503:
Other `extra_tests` files often used `from testsutils import asset_rises`, but the module did not have pattern matching. 
I needed pattern matching, so I used `unittest` in the `syntax_async.py`, is that okay?

youknowone:
I think this is good

> Other `extra_tests` files often used `from testsutils import asset_rises`, but the module did not have pattern matching. I needed pattern matching, so I used `unittest` in the `syntax_async.py`, is that okay?

youknowone (approved):
looks good, thank you!

PR #3902, opened 2022-07-16
Fix byte indices must be integers or slices, error message

kimnanhee:
`b = b'python'`
`b['a']`

TypeError: bytes indices must be integers or slices or classes that override __index__ operator, not 'str'
=> TypeError: byte indices must be integers or slices or classes that override __index__ operator, not 'str'


youknowone (approved):
looks good, thank you!

PR #3903, opened 2022-07-16
Implement number protocol for `PyBytes`

TwoPair:


vm/src/builtins/bytes.rs
```diff
@@ -606,6 +607,21 @@ impl AsSequence for PyBytes {
...
+impl AsNumber for PyBytes {
+    const AS_NUMBER: PyNumberMethods = PyNumberMethods {
+        remainder: Some(|number, other, vm| {
+            let formatted = number
+                .obj
+                .downcast_ref::<PyBytes>()
+                .unwrap()
+                .inner
+                .cformat(other.to_owned(), vm)?;
+            Ok(formatted.to_pyobject(vm))
```
Snowapril, line 619:
```suggestion
            Ok(Self::number_downcast(number)
                .inner
                .cformat(other.to_pyobject(vm), vm)?
                .to_pyobject(vm))
```

Snowapril: `number_downcast` will do what we want 😊

TwoPair: @Snowapril Thanks for recognizing to me 😄

vm/src/builtins/bytes.rs
```diff
@@ -606,6 +607,18 @@ impl AsSequence for PyBytes {
     };
 }
 
+impl AsNumber for PyBytes {
+    const AS_NUMBER: PyNumberMethods = PyNumberMethods {
+        remainder: Some(|number, other, vm| {
+            Ok(Self::number_downcast(number)
+                .inner
+                .cformat(other.to_pyobject(vm), vm)?
```
youknowone, line 615:
```suggestion
                .mod_(other, vm)?
```
reusing `mod_` will be better

youknowone (changes requested):

vm/src/builtins/bytes.rs
```diff
@@ -606,6 +607,17 @@ impl AsSequence for PyBytes {
     };
 }
 
+impl AsNumber for PyBytes {
+    const AS_NUMBER: PyNumberMethods = PyNumberMethods {
+        remainder: Some(|number, other, vm| {
+            Self::number_downcast(number)
+                .mod_(other.to_pyobject(vm), vm)
```
youknowone, line 614:
```suggestion
                .mod_(other.to_owned(), vm)
```
we use to_pyobject to create PyObjectRef from a concrete type.
Because we already have `PyObject`, `to_owned()` is enough.

youknowone (changes requested):

youknowone:
The  commit history looking weird.
while upstream is RustPython/RustPython, please try `git rebase -i upstream/main` and remove commits you didn't directly authored.

Issue #3899, opened 2022-07-16
getset doesn't have deleter


youknowone:
`del` calls `setattr` with None in getset.

CPython type: `PyGetSetDef`
RustPython type: `PyGetSet`

#3897 is related

moreal:
I'm not sure, but the `PyGetSet.deleter` may resolve this issue?

youknowone:
Removing `PyGetSet.deleter` will resolve the issue.

youknowone:
closed by #3904

PR #3904, opened 2022-07-16
Remove `PyGetSet.deleter`

### LINKED ISSUE #3897: f_trace of frame(PyFrameRef) is deletable

moreal:
This pull request tries to resolve #3897, #3899. You can run the below code in the main branch (before this pull request) and you'll see the result message like below:

```
>>>>> import sys
>>>>> frame = sys._getframe()
>>>>> del frame.f_trace
AttributeError: attribute 'f_trace' of 'frame' objects is not writable
```

But after this pull request, it became deletable.

```
>>>>> import sys
>>>>> frame = sys._getframe()
>>>>> del frame.f_trace
>>>>> frame.f_trace = "a"
>>>>> frame.f_trace
'a'
>>>>> del frame.f_trace
>>>>> frame.f_trace
```

### DIFF
--- a/vm/src/builtins/function.rs
+++ b/vm/src/builtins/function.rs
@@ -620,10 +620,6 @@ impl PyCell {
     fn set_cell_contents(&self, x: PyObjectRef) {
         self.set(Some(x))
     }
-    #[pyproperty(deleter)]
-    fn del_cell_contents(&self) {
-        self.set(None)
-    }
 }
 
 pub fn init(context: &Context) {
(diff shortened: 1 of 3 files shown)


youknowone:
The deleter in PyGetSet is a redundant one. Removing the whole deleter and replacing with setter(None) is preferred

fanninpm:
> The deleter in PyGetSet is a redundant one. Removing the whole deleter and replacing with setter(None) is preferred

See also #3899.

moreal:
I'll open this pull request when I understand the relationship between getset descripter and property 🤔 Or it seems okay to merge when you look, it can be opened and merged.

youknowone:
property is used for python code. getset is used for rust code (c code for cpython).
they are very similar but a little bit different.
(e.g. null == None for  property but different in getset. property has deleter while getset doesn't)

it is confusing, but the name `pyproperty` is used to create getset. we don't have a rust way to create property yet.

moreal:
While looking the `derive/src/pyclass,rs` code, should the `GetSet` in the `AttrName` enums become `Property`? 🤔 There are `PyGetSet` and `PyProperty`, and they are different so isn't it better that the enum should name `Property` instead `GetSet`?

youknowone:
In `derive/src/pyclass.rs`, All they are referring getset. If I unify the names, I vote for renaming everything to getset instead of property.

youknowone (approved):
thank you!

PR #3908, opened 2022-07-17
Add MappingProxy's get method implementation handling missing key

### LINKED ISSUE #3906: Add `MappingProxy`'s `get` method implementation handling missing key
## Summary 

Add `MappingProxy`'s `get` method implementation handling missing key

jopemachine:
Fixes #3906.

### DIFF
--- a/vm/src/builtins/mappingproxy.rs
+++ b/vm/src/builtins/mappingproxy.rs
@@ -83,9 +83,12 @@ impl PyMappingProxy {
         default: OptionalArg,
         vm: &VirtualMachine,
     ) -> PyResult<Option<PyObjectRef>> {
-        let default = default.into_option();
-        let value = self.get_inner(key, vm)?.or(default);
-        Ok(value)
+        let obj = self.to_object(vm)?;
+        Ok(Some(vm.call_method(
+            &obj,
+            "get",
+            (key, default.unwrap_or_none(vm)),
+        )?))
     }
 
     #[pymethod(magic)]
--- a/Lib/test/test_types.py
+++ b/Lib/test/test_types.py
@@ -1029,8 +1029,6 @@ class MappingProxyTests(unittest.TestCase):
         self.assertIsNone(view.get('xxx'))
         self.assertEqual(view.get('xxx', 42), 42)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_missing(self):
         class dictmissing(dict):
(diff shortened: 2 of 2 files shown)


vm/src/builtins/mappingproxy.rs
```diff
@@ -83,9 +83,12 @@ impl PyMappingProxy {
         default: OptionalArg,
         vm: &VirtualMachine,
     ) -> PyResult<Option<PyObjectRef>> {
-        let default = default.into_option();
-        let value = self.get_inner(key, vm)?.or(default);
-        Ok(value)
+        let obj = self.to_object(vm)?;
+        Ok(Some(vm.call_method(
+            &obj,
```
youknowone, line 88: how is this handled when inner type is class? just attribute error about `get`?

jopemachine: Below code is some codes I wrote for the test.

```rust
...mappingproxy.rs

#[pymethod]
fn get(
    &self,
    key: PyObjectRef,
    default: OptionalArg,
    vm: &VirtualMachine,
) -> PyResult<Option<PyObjectRef>> {
    match &self.mapping {
        MappingProxyInner::Class(_) => {
            println!("MappingProxyInner::Class get called");
            let obj = self.to_object(vm)?;
            Ok(Some(vm.call_method(
                &obj,
                "get",
                (key, default.unwrap_or_none(vm)),
            )?))
        },
        MappingProxyInner::Mapping(_) => {
            ...
        }
    }
}

...dict.rs

#[pymethod]
fn get(
    &self,
    key: PyObjectRef,
    default: OptionalArg<PyObjectRef>,
    vm: &VirtualMachine,
) -> PyResult {
    println!("dict.get called");

    match self.entries.get(vm, &*key)? {
        Some(value) => Ok(value),
        None => Ok(default.unwrap_or_none(vm)),
    }
}
```

When MappingProxy::Class's `get` is called like the below code,

```
Exception.__dict__.get('__init__')
```

The code execution result is as follows.

```
>>>>> Exception.__dict__.get('__init__')
MappingProxyInner::Class get called
dict.get called
<method '__init__' of 'Exception' objects>
```

In the result, `dict.rs`'s get method is called in the same way when the inner type is class.

So, I think this code works in the same way with`MappingProxy::Class` too.

youknowone: great, thank you!

youknowone (approved):

PR #3909, opened 2022-07-17
Add weakproxy richcompare

oow214:


vm/src/builtins/weakproxy.rs
```diff
@@ -136,6 +141,25 @@ impl SetAttr for PyWeakProxy {
...
 
+impl Comparable for PyWeakProxy {
+    fn cmp(
+        zelf: &crate::Py<Self>,
+        other: &PyObject,
+        op: PyComparisonOp,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyComparisonValue> {
+        println!("richcompare");
+        op.eq_only(|| {
```
youknowone, line 152: Does weakproxy only support eq operation?
What happens if the proxy object is `int` and used for `>` operation?

youknowone: I checked cpython source code and it simply calls `PyObject_RichCompare`, which is `PyObjectRef::rich_compare` in RustPython.

youknowone: I checked cpython source code and it simply calls `PyObject_RichCompare`, which is `PyObjectRef::rich_compare` in RustPython.

youknowone (changes requested):

youknowone (approved):
looks good, thank you!

PR #3910, opened 2022-07-17
Update multiple container-related unittests from CPython 3.10.5

youknowone:


Lib/test/test_array.py
```diff
@@ -1174,6 +1249,8 @@ def test_iadd(self):
     def test_imul(self):
         super().test_imul()
```
fanninpm, line 1250: Can the ones around here that no longer have the `@unittest.expectedFailure` decorators on them be removed? @frank-king left these behind without realizing why I put them there.

Lib/test/test_deque.py
```diff
@@ -949,6 +929,21 @@ def test_free_after_iterating(self):
...
+
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    def test_count(self):
+        super().test_count()
+
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    def test_index(self):
+        super().test_index()
```
fanninpm, line 945: When I add things like this, I'd like to direct contributors against leaving the last two lines of this pattern behind when they un-mark these tests.

Lib/test/test_list.py
```diff
@@ -167,5 +198,48 @@ def test_preallocation(self):
...
+
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    def test_count(self):
+        super().test_count()
+
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    def test_contains_fake(self):
+        super().test_contains_fake()
```
fanninpm, line 242: When I add things like this, I'd like to direct contributors against leaving the last two lines of this pattern behind when they un-mark these tests.

Lib/test/test_userlist.py
```diff
@@ -11,6 +11,16 @@ class UserListTest(list_tests.CommonTest):
...
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    def test_contains_fake(self):
+        super().test_contains_fake()
+
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    def test_count(self):
+        super().test_count()
```
fanninpm, line 22: When I add things like this, I'd like to direct contributors against leaving the last two lines of this pattern behind when they un-mark these tests.

Lib/test/test_array.py
youknowone: @fanninpm you don't need to mention someone unless we actually need their attention. In this case, I think referring commit or pr url can be helpful, but not about the author.

youknowone:
@fanninpm I added additional comments for new methods from rustpython and added 4 more commits

Lib/test/list_tests.py
```diff
@@ -363,6 +365,8 @@ def __eq__(self, other):
             # verify that original order and values are retained.
             self.assertIs(x, y)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_index(self):
         super().test_index()
```
fanninpm, line 371: In the future, classes dependent on this test may start unexpectedly passing. Will they unexpectedly pass _all at once_ or _one at a time_?

youknowone: Tests are supportive code. They will be handled when they actually be solved.

The answer is I don't know. Sometimes they passes at once, sometimes not. I didn't look in the details.

Lib/test/test_userlist.py
```diff
@@ -11,6 +11,16 @@ class UserListTest(list_tests.CommonTest):
     def test_init(self):
         super().test_init()
```
fanninpm, line 12: The `test_init` method in `list_tests.CommonTest` is no longer marked. I don't think this needs to be here.

fanninpm (commented):
Just a few minor cleanup notes.

Lib/test/test_userlist.py
youknowone: @fanninpm  we expect code review is about changed code, instead of any code around changed code.

I think that change is independent from this PR. If you want me to fix it, I will do it in separated PR.

PR #3911, opened 2022-07-17
Unify elements referring to getset to `GetSet`

moreal:
This pull request is from https://github.com/RustPython/RustPython/pull/3904#issuecomment-1186551766. It tries to unify naming of elements referring to getset to `GetSet`.

It was a reply of my question and it's one vote of one maintainer. It may need discussion between maintainers 🤔 
Feel free to discuss it even if this pull request closed or become to stay while long time. 🙏🏻 .

### DIFF
--- a/derive/src/pyclass.rs
+++ b/derive/src/pyclass.rs
@@ -404,7 +404,7 @@ struct MethodItem {
 }
 
 /// #[pyproperty]
-struct PropertyItem {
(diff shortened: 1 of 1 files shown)


youknowone (approved):

youknowone:
I don't think we currently have active members to discuss it.
`derive` code is hidden from user side. I think this is not a big deal.

Issue #3881, opened 2022-07-14
`set` `intersection_update`, `difference_update`, `symmetric_difference_update` should check self is equal to given arg


Snowapril:
## Feature
This is also relevant to inplace operations( `iand`, `isub`, `ixor` ). 
CPython checks whether given argument and selfobject are equal and then returns.

* `intersection_update` : If self == arg, returns self.copy()
* `difference_update` : If self == arg, returns self.clear(); Ok(())
* `symmetric_difference_update` : If self == arg, returns self.clear(); Ok(())

At the current implementation, they get a parameter as `SetIterable` which is converted from PyObjectRef before entering each method.
For checking `self == arg`, we should convert it after checking.

## Python Documentation
* https://github.com/python/cpython/blob/3.10/Objects/setobject.c#L1627-L1630
* https://github.com/python/cpython/blob/3.10/Objects/setobject.c#L1410-L1412
* https://github.com/python/cpython/blob/3.10/Objects/setobject.c#L1186-L1189

## Relevant Test Case
* `test_set.py` `test_inplace_on_self`

kth496:
@youknowone Could you assign me to this issue? 😀

PR #3912, opened 2022-07-18
Fix some set operation to compare self with args

kth496:
## Intro
Fix https://github.com/RustPython/RustPython/issues/3881

### 1. `iand`
**Before**
```
>>>>> import operator
>>>>> a = set((1, 2, 3))
>>>>> b = set((2, 3, 4))
>>>>> operator.iand(a, b)
{2, 3}
>>>>> a
{2, 3}
>>>>> operator.iand(a, a)
set()
>>>>> a
set()
>>>>>
```

**After**
```
>>>>> import operator
>>>>> a = set((1, 2, 3))
>>>>> b = set((2, 3, 4))
>>>>> operator.iand(a, b)
{2, 3}
>>>>> a
{2, 3}
>>>>> operator.iand(a, a)
{2, 3}
>>>>> a
{2, 3}
>>>>>
```
## What's left
- [x] iand
- [ ] isub
- [ ] ixor
- [ ] intersection_update
- [ ] difference_update
- [ ] symmetric_difference_update


youknowone:
It is up to you, but you don't need to put everything in single PR. You can submit splat small PRs for each functions.

vm/src/builtins/set.rs
```diff
@@ -658,7 +658,11 @@ impl PySet {
     }
 
     #[pymethod(magic)]
-    fn iand(zelf: PyRef<Self>, set: AnySet, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+    fn iand(zelf: PyRef<Self>, set: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+        if zelf.get_id() == set.get_id() {
+            return Ok(zelf);
+        }
```
Snowapril, line 664: `intersection_update` also needs to have a check like this. As `iand` and `intersection_update` share the same `PySetInner::intersection_update`, we can do better 😊

kth496:
@youknowone @Snowapril 
All bugs described in the issue have been fixed. But I think there's another way to do this better. I'm not good at Rust. Could you please review the code?

I tried to remove duplicated codes. But there are some different constraints between inplace operation methods. (Please See NOTE below)

NOTE:
1. `iand`
    - Only one argument with AnySet type. Need to check the type of argument.

2. `intersection_update`
    - Variable arguments are allowed. (even 0)
    - All Iterable Types are allowed. (include AnySet)

Same issue on `isub` `difference_update` `ixor` ...

vm/src/builtins/set.rs
```diff
@@ -649,49 +652,103 @@ impl PySet {
...
+        zelf.inner.intersection_update(arguments.into_iter(), vm)?;
         Ok(())
     }
 
     #[pymethod(magic)]
-    fn iand(zelf: PyRef<Self>, set: AnySet, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
-        zelf.inner
-            .intersection_update(std::iter::once(set.into_iterable(vm)?), vm)?;
+    fn iand(zelf: PyRef<Self>, set: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+        if !set.class().is(vm.ctx.types.set_type) {
```
youknowone, line 675: how about frozenset? if frozenset is allowed, maybe AnySet check?

vm/src/builtins/set.rs
```diff
@@ -649,49 +652,103 @@ impl PySet {
...
         vm: &VirtualMachine,
     ) -> PyResult<()> {
-        self.inner.intersection_update(others.into_iter(), vm)?;
+        let arguments = others.into_vec();
+        if arguments.len() == 1 {
+            if let Some(iterable) = arguments.first() {
+                if zelf.is(iterable) {
+                    return Ok(());
+                }
+            }
```
youknowone, line 665: I think this check is not only adaptable for `len() == 1` case but also do for every argument.

For example, how about `s = set(['a']); s.intersection_update(s, s, s, s, s, s)`?

youknowone (changes requested):
About your concern, I think your approach is reasonable enough.
Maybe there can be more room to do better, but current versions looks good enough to handle them in proper cost.

vm/src/builtins/set.rs
```diff
@@ -649,49 +652,103 @@ impl PySet {
...
         Ok(())
     }
 
     #[pymethod(magic)]
-    fn iand(zelf: PyRef<Self>, set: AnySet, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
-        zelf.inner
-            .intersection_update(std::iter::once(set.into_iterable(vm)?), vm)?;
+    fn iand(zelf: PyRef<Self>, set: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
+        if !set.class().is(vm.ctx.types.set_type) {
+            return Err(vm.new_type_error("Type error".to_owned()));
```
DimitrisJim, line 676: Should be `unsupported operand type(s) for +=: 'type1' and 'type2'"`, similarly for the rest of the type errors.

PR #3915, opened 2022-07-18
implement TextIOBase encoding attribute

### LINKED ISSUE #3896: io.StringIO doesn't have encoding attribute
`_TextIOBase` is the expected owner of the attribute.

Adding a dummy property `encoding` with value `utf-8` will be a good first issue

hy-kiera:
fix #3896 

- python3
```
>>> import io
>>> io.TextIOBase.encoding
<attribute 'encoding' of '_io._TextIOBase' objects>
```
- RustPython
```
>>>>> import io
>>>>> io.TextIOBase.encoding
AttributeError: 'TextIOBase' object has no attribute 'encoding'
```

### DIFF
--- a/vm/src/stdlib/io.rs
+++ b/vm/src/stdlib/io.rs
@@ -704,7 +704,12 @@ mod _io {
     struct _TextIOBase;
 
     #[pyimpl(flags(BASETYPE))]
-    impl _TextIOBase {}
+    impl _TextIOBase {
+        #[pyattr]
+        fn encoding(vm: &VirtualMachine) -> PyObjectRef {
+            vm.ctx.none()
+        }
+    }
 
     #[derive(FromArgs, Clone)]
     struct BufferSize {
--- a/vm/src/stdlib/io.rs
+++ b/vm/src/stdlib/io.rs
@@ -701,12 +701,13 @@ mod _io {
     // TextIO Base has no public constructor
     #[pyattr]
     #[pyclass(name = "_TextIOBase", base = "_IOBase")]
+    #[derive(Debug, PyPayload)]
     struct _TextIOBase;
 
     #[pyimpl(flags(BASETYPE))]
     impl _TextIOBase {
-        #[pyattr]
-        fn encoding(vm: &VirtualMachine) -> PyObjectRef {
+        #[pyproperty]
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/io.rs
```diff
@@ -701,10 +701,16 @@ mod _io {
...
+    #[derive(Debug, PyPayload)]
     struct _TextIOBase;
 
     #[pyimpl(flags(BASETYPE))]
-    impl _TextIOBase {}
+    impl _TextIOBase {
+        #[pyattr]
+        fn encoding(ctx: &Context) -> PyObjectRef {
+            ctx.none()
+        }
```
Snowapril, line 712:
```suggestion
        #[pyproperty]
        fn encoding(&self, vm: &VirtualMachine) -> PyObjectRef {
            vm.ctx.none()
        }
```
For adding `encoding` as getset_descriptor, we need to use `#[pyproperty]`

hy-kiera: When I recomplied your suggestion, I cannot get `None` value. I'm confusing `pyproperty` and `pyattr`.
- python3
```
>>> print(io.TextIOBase().encoding)
None
>>> print(io.TextIOBase.encoding)
<attribute 'encoding' of '_io._TextIOBase' objects>
```
- rustpython
```
>>>>> print(io.TextIOBase().encoding)
RuntimeError: Expected payload '_TextIOBase' but 'TextIOBase' found
>>>>> print(io.TextIOBase.encoding)
<getset_descriptor object at 0x60000341f480>
```

youknowone: `pyproperty` creates a getset field, which is very similar to `property` in python.
`pyattr` creates a simple attribute, same to `self.encoding = None`

youknowone: @Snowapril is it a getset instead of attribute?

Snowapril: Yes. It should be `getset`. `getset_repr` will print "<attribute '%V' of '%s' objects>".
Current rustpython getset does not implement `repr`.

youknowone (approved):
looks good, thank you!

PR #3898, opened 2022-07-16
Implement missing methods to `MappingProxy`

jopemachine:
Fixes #3875.

PR #3916, opened 2022-07-18
Fix outdated test_userlist

youknowone:
requested by @fanninpm 

https://github.com/RustPython/RustPython/pull/3910#discussion_r923347986


youknowone:
that was an invisible conflict to #3898

PR #3931, opened 2022-07-19
Add itertool.combinations.__reduce__ method

seryoungshim17:
```bash
>>>>> from itertools import combinations
>>>>> c = combinations([1,2,3,4], 2)
>>>>> c.__reduce__()
(<class 'itertools.combinations'>, ((1, 2, 3, 4), 2))
```

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -1308,6 +1308,7 @@ mod decl {
     struct PyItertoolsCombinations {
         pool: Vec<PyObjectRef>,
         indices: PyRwLock<Vec<usize>>,
+        result: PyRwLock<Option<Vec<usize>>>,
         r: AtomicCell<usize>,
         exhausted: AtomicCell<bool>,
     }
@@ -1341,6 +1342,7 @@ mod decl {
             PyItertoolsCombinations {
                 pool,
                 indices: PyRwLock::new((0..r).collect()),
+                result: PyRwLock::new(None),
                 r: AtomicCell::new(r),
                 exhausted: AtomicCell::new(r > n),
             }
@@ -1350,7 +1352,39 @@ mod decl {
     }
 
     #[pyclass(with(IterNext, Constructor))]
-    impl PyItertoolsCombinations {}
+    impl PyItertoolsCombinations {
+        #[pymethod(magic)]
+        fn reduce(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyTupleRef {
+            let result = zelf.result.read();
+            if let Some(result) = &*result {
+                if zelf.exhausted.load() {
+                    vm.new_tuple((
+                        zelf.class().to_owned(),
+                        vm.new_tuple((vm.new_tuple(()), vm.ctx.new_int(zelf.r.load()))),
+                    ))
+                } else {
+                    vm.new_tuple((
+                        zelf.class().to_owned(),
+                        vm.new_tuple((
+                            vm.new_tuple(zelf.pool.clone()),
+                            vm.ctx.new_int(zelf.r.load()),
+                        )),
+                        vm.ctx
+                            .new_tuple(result.iter().map(|&i| zelf.pool[i].clone()).collect()),
+                    ))
+                }
+            } else {
+                vm.new_tuple((
+                    zelf.class().to_owned(),
+                    vm.new_tuple((
+                        vm.new_tuple(zelf.pool.clone()),
+                        vm.ctx.new_int(zelf.r.load()),
+                    )),
+                ))
+            }
+        }
+    }
+
     impl IterNextIterable for PyItertoolsCombinations {}
     impl IterNext for PyItertoolsCombinations {
         fn next(zelf: &Py<Self>, vm: &VirtualMachine) -> PyResult<PyIterReturn> {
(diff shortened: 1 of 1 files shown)


Snowapril:
Looks good! But, did you check how `itertools.combinations.__reduce__` work in cpython? They act differently according to several conditions. Please check the below codes.
### CPython
```python
>>> a = itertools.combinations(range(4), 3)
>>> a.__reduce__()
(<class 'itertools.combinations'>, ((0, 1, 2, 3), 3))
>>> next(a); a.__reduce__()
(0, 1, 2)
(<class 'itertools.combinations'>, ((0, 1, 2, 3), 3), (0, 1, 2))
>>> next(a); a.__reduce__()
(0, 1, 3)
(<class 'itertools.combinations'>, ((0, 1, 2, 3), 3), (0, 1, 3))
>>> next(a); a.__reduce__()
(0, 2, 3)
(<class 'itertools.combinations'>, ((0, 1, 2, 3), 3), (0, 2, 3))
>>> next(a); a.__reduce__()
(1, 2, 3)
(<class 'itertools.combinations'>, ((0, 1, 2, 3), 3), (1, 2, 3))
>>> next(a)
StopIteration
>>> a.__reduce__()
(<class 'itertools.combinations'>, ((), 3))
```

But in current PR always returns `(<class 'itertools.combinations'>, ((0, 1, 2, 3), 3))` in this case.
Please check this [cpython impl](https://github.com/python/cpython/blob/3.10/Modules/itertoolsmodule.c#L2760-L2787)

vm/src/stdlib/itertools.rs
```diff
@@ -1264,7 +1267,18 @@ mod decl {
     }
 
     #[pyimpl(with(IterNext, Constructor))]
-    impl PyItertoolsCombinations {}
+    impl PyItertoolsCombinations {
+        #[pymethod(magic)]
+        fn reduce(zelf: PyRef<Self>, vm: &VirtualMachine) -> (PyTypeRef, (PyTupleRef, PyIntRef)) {
```
fanninpm, line 1272: You may want to change this function's signature so that the function returns `PyTupleRef`, a more flexible return type. Something like…

```rust
fn reduce(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyTupleRef {
    if zelf.result.load().is_none() {
        return vm.new_tuple((IntoPyTuple::into_pytuple(zelf.pool.clone(), vm), vm.ctx.new_int(zelf.r.load())));
    } else if zelf.exhausted.load() {
        return vm.new_tuple(((), vm.ctx.new_int(zelf.r.load())));
    } else {
        ...
        return vm.new_tuple((IntoPyTuple::into_pytuple(zelf.pool.clone(), vm), vm.ctx.new_int(zelf.r.load()), ...));
    }
}
```

vm/src/stdlib/itertools.rs
```diff
@@ -1264,7 +1266,42 @@ mod decl {
...
+                        zelf.result
+                            .read()
+                            .as_ref()
+                            .unwrap()
+                            .iter()
+                            .map(|&i| zelf.pool[i].clone())
+                            .collect(),
+                    ),
+                ));
+            }
```
youknowone, line 1302:
```suggestion
            if zelf.result.read().is_none() {
                vm.new_tuple((
                    zelf.class().clone(),
                    vm.new_tuple((
                        vm.new_tuple(zelf.pool.clone()),
                        vm.ctx.new_int(zelf.r.load()),
                    )),
                ))
            } else if zelf.exhausted.load() {
                vm.new_tuple((
                    zelf.class().clone(),
                    vm.new_tuple((vm.new_tuple(()), vm.ctx.new_int(zelf.r.load()))),
                ))
            } else {
                vm.new_tuple((
                    zelf.class().clone(),
                    vm.new_tuple((
                        vm.new_tuple(zelf.pool.clone()),
                        vm.ctx.new_int(zelf.r.load()),
                    )),
                    vm.ctx.new_tuple(
                        zelf.result
                            .read()
                            .as_ref()
                            .unwrap()
                            .iter()
                            .map(|&i| zelf.pool[i].clone())
                            .collect(),
                    ),
                ))
            }
```

Rust programmers prefer to not to use `return` for last expression of the function block

vm/src/stdlib/itertools.rs
```diff
@@ -1264,7 +1266,42 @@ mod decl {
     }
 
     #[pyimpl(with(IterNext, Constructor))]
-    impl PyItertoolsCombinations {}
+    impl PyItertoolsCombinations {
+        #[pymethod(magic)]
+        fn reduce(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyTupleRef {
+            if zelf.result.read().is_none() {
```
youknowone, line 1272: zelf.result.read() is repeated here and below.

Because here is no other locking chance, reusing will be better.

```rust
if let Some(result) = zelf.result.read() {
    ...  // is_some case (second and third block) here
} else {
    ...  // is_none case (first block) here
}
```

youknowone (changes requested):
I need to look in details but it seems next algorithm is changed. did it have a bug?
I also want to know if this is actually fixing reduce test cases. Which test in test_itertools is related and how much is it fixed?

fanninpm:
> Which test in test_itertools is related and how much is it fixed?

`TestBasicOps.test_combinations` is the test in question. (Unfortunately, I think the test functions in `test_itertools.py` are a bit too over-broad in scope. Ideally, a test function should be as minimal as possible.)

youknowone:
Yes. the tests unit are sometime too big. I think checking a few lines of them is fixed would be enough, practically.

youknowone:
@fanninpm could you review this PR again?

fanninpm (commented):
In lieu of removing a `@unittest.expectedFailure` decorator from the CPython test suite, can you add something to the `extra_tests` directory to ensure that this change matches CPython's behavior?

youknowone:
@fanninpm  If the test cases are included in unittest as part of TestBasicOps.test_combinations, I don't think that's necessary. The test will eventually pass and I don't want make contributors pay too much cost for temporal test quality.

seryoungshim17:
221~224 line in `TestBasicOps.test_combinations` has been fixed.

fanninpm:
Could you please rebase this branch on a fresh copy of `main`?

fanninpm:
@seryoungshim17 Are you still interested in pursuing this?

PR #3935, opened 2022-07-20
Fix set test_merge_and_mutate

### LINKED ISSUE #3866: `set.update` should not use `IterNext`
## Feature
Our `set.update` uses `IterNext` for iterating given args and update the set contents (makes sense).
But this makes incompatibility with cpython.
cpython's implementation is converting given args to tuple and uses `get_size` and `get_item`. 
As these method does not go through the `iternext`, there is no exception if the set is mutated during `update`. 

## Python Documentation
https://github.com/python/cpython/blob/3.10/Objects/setobject.c#L930-L941

## Related test cases
* `test_set.py` `test_merge_and_mutate`

tgsong827:
fix #3866

- [X] make function update_internal()
- [X] check AnySet Type
- [x] check Dict Type

### DIFF
--- a/vm/src/builtins/set.rs
+++ b/vm/src/builtins/set.rs
@@ -2,8 +2,8 @@
  * Builtin set type with a sequence of unique items.
  */
 use super::{
-    builtins_iter, IterStatus, PositionIterInternal, PyDictRef, PyGenericAlias, PyTupleRef, PyType,
-    PyTypeRef,
+    builtins_iter, IterStatus, PositionIterInternal, PyDict, PyDictRef, PyGenericAlias, PyTupleRef,
+    PyType, PyTypeRef,
 };
 use crate::common::{ascii, hash::PyHash, lock::PyMutex, rc::PyRc};
 use crate::{
@@ -353,6 +353,36 @@ impl PySetInner {
         Ok(())
     }
 
+    fn update_internal(&self, iterable: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        // check AnySet
+        if let Ok(any_set) = AnySet::try_from_object(vm, iterable.to_owned()) {
+            self.merge_set(any_set, vm)
+        // check Dict
+        } else if let Ok(dict) = iterable.to_owned().downcast_exact::<PyDict>(vm) {
+            self.merge_dict(dict, vm)
+        } else {
+            // add iterable that is not AnySet or Dict
+            for item in iterable.try_into_value::<ArgIterable>(vm)?.iter(vm)? {
+                self.add(item?, vm)?;
+            }
+            Ok(())
+        }
+    }
+
+    fn merge_set(&self, any_set: AnySet, vm: &VirtualMachine) -> PyResult<()> {
+        for item in any_set.as_inner().elements() {
+            self.add(item, vm)?;
+        }
+        Ok(())
+    }
+
+    fn merge_dict(&self, dict: PyDictRef, vm: &VirtualMachine) -> PyResult<()> {
+        for (key, _value) in dict {
+            self.add(key, vm)?;
+        }
+        Ok(())
+    }
+
     fn intersection_update(
         &self,
         others: impl std::iter::Iterator<Item = ArgIterable>,
@@ -642,8 +672,10 @@ impl PySet {
     }
 
     #[pymethod]
-    fn update(&self, others: PosArgs<ArgIterable>, vm: &VirtualMachine) -> PyResult<()> {
-        self.inner.update(others.into_iter(), vm)?;
+    fn update(&self, others: PosArgs<PyObjectRef>, vm: &VirtualMachine) -> PyResult<()> {
+        for iterable in others {
+            self.inner.update_internal(iterable, vm)?;
+        }
         Ok(())
     }
 
@@ -718,7 +750,7 @@ impl Constructor for PySet {
 }
 
 impl Initializer for PySet {
-    type Args = OptionalArg<ArgIterable>;
+    type Args = OptionalArg<PyObjectRef>;
 
     fn init(zelf: PyRef<Self>, iterable: Self::Args, vm: &VirtualMachine) -> PyResult<()> {
         if zelf.len() > 0 {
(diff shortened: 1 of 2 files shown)


vm/src/builtins/set.rs
```diff
@@ -346,11 +347,41 @@ impl PySetInner {
         vm: &VirtualMachine,
     ) -> PyResult<()> {
         for iterable in others {
+            self.update_internal(iterable, vm)?;
+        }
+        Ok(())
+    }
+
+    fn update_internal(&self, iterable: ArgIterable, vm: &VirtualMachine) -> PyResult<()> {
```
youknowone, line 355:
```suggestion
    fn update_internal(&self, iterable: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
```
Creating ArgIterable can be deferred because any ArgIterable feature is not used before this function call.

vm/src/builtins/set.rs
```diff
@@ -346,11 +347,41 @@ impl PySetInner {
...
+
+            // add items that is not AnySet
             for item in iterable.iter(vm)? {
                 self.add(item?, vm)?;
             }
+            Ok(())
+        }
+    }
+
+    fn set_merge(&self, any_set: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
```
youknowone, line 372:
```suggestion
    fn set_merge(&self, any_set: AnySet, vm: &VirtualMachine) -> PyResult<()> {
```
Once we checked the type is AnySet, keeping the type is better to encode the information into the code.
We might add `fn as_inner()` to `AnySet` for better readability.

vm/src/function/protocol.rs
```diff
@@ -100,6 +101,13 @@ where
     }
 }
 
+impl From<ArgIterable> for PyObjectRef {
```
youknowone, line 104: I don't think this is necessary for your approach.
Instead of this one, maybe `impl From<AnySet> for PyObjectRef` could be useful.

youknowone (changes requested):

vm/src/function/protocol.rs
tgsong827: Thank you for your advice.
I'll give it a try!

vm/src/builtins/set.rs
```diff
@@ -637,13 +656,15 @@ impl PySet {
 
     #[pymethod(magic)]
     fn ior(zelf: PyRef<Self>, set: AnySet, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
```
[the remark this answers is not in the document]
tgsong827: Thanks for your good advice!
I fix that `ior` can use the origin update().
Could you please review what I fixed?

vm/src/builtins/set.rs
```diff
@@ -637,13 +679,18 @@ impl PySet {
...
+        // fn ior(zelf: PyRef<Self>, set: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
         zelf.inner.update(set.into_iterable_iter(vm)?, vm)?;
+        // zelf.inner.update_internal(set.into_object(), vm)?;
+        // zelf.inner.update_internal(set, vm)?;
```
Snowapril, line 685: If you have a mind to fix `ior`, if `PyObjectRef` cannot be converted into `AnySet` we should return `PyArithmeticValue::NotImplemented`

vm/src/builtins/set.rs
```diff
@@ -353,6 +354,47 @@ impl PySetInner {
...
+        } else {
+            // check Dict
+            if let Ok(dict) = iterable.to_owned().downcast_exact::<PyDict>(vm) {
+                self.merge_dict(dict, vm)
+            } else {
+                // add iterable that is not AnySet or Dict
+                for item in iterable.try_into_value::<ArgIterable>(vm)?.iter(vm)? {
+                    self.add(item?, vm)?;
+                }
+                Ok(())
```
Snowapril, line 370:
```suggestion
        } else if let Ok(dict) = iterable.to_owned().downcast_exact::<PyDict>(vm) {
            self.merge_dict(dict, vm)
        } else {
            // add iterable that is not AnySet or Dict
            for item in iterable.try_into_value::<ArgIterable>(vm)?.iter(vm)? {
                self.add(item?, vm)?;
            }
            Ok(())
```

tgsong827: Thanks for your comment!

vm/src/builtins/set.rs
```diff
@@ -353,6 +354,45 @@ impl PySetInner {
...
+        match extract_set(obj.as_object()) {
+            Some(other_inner) => {
+                for item in other_inner.elements() {
+                    self.add(item, vm)?;
+                }
+                Ok(())
+            }
+            None => {
+                Err(vm.new_type_error(format!("{} is not a subtype of set or frozenset", class)))
+            }
```
youknowone, line 385: this is impossible. `extract_set` always succeed. obj is always AnySet here. Try to add `as_inner()` to `AnySet`. I think `extract_set` doesn't fit to AnySet.

youknowone (changes requested):
looks good in general. I left a comment about impossible branch

vm/src/builtins/set.rs
```diff
@@ -997,6 +1029,14 @@ impl AnySet {
     ) -> PyResult<impl std::iter::Iterator<Item = ArgIterable>> {
         Ok(std::iter::once(self.into_iterable(vm)?))
     }
+
+    fn as_inner(&self, _vm: &VirtualMachine) -> PySetInner {
+        match_class!(match self.object.as_object() {
+            ref set @ PySet => set.inner.clone(),
+            ref frozen @ PyFrozenSet => frozen.inner.clone(),
```
youknowone, line 1036: cloning will result unexpected behaviors when writing.
is it impossible to borrow `&PySetInner` from `self.object`? 
in that case, probably AnySet need to be changed to enum like
```
enum AnySet {
    Set(PyRef<PySet>),
    FrozenSet(PyRef<PyFrozenSet>),
}
```

vm/src/builtins/set.rs
```diff
@@ -997,6 +1029,14 @@ impl AnySet {
     ) -> PyResult<impl std::iter::Iterator<Item = ArgIterable>> {
         Ok(std::iter::once(self.into_iterable(vm)?))
     }
+
+    fn as_inner(&self, _vm: &VirtualMachine) -> PySetInner {
```
youknowone, line 1033:
```suggestion
    fn as_inner(&self) -> &PySetInner {
```
The return type should be fixed to `&PySetInner` for this method.

youknowone (approved):
Great! Thank you for the long time journey to reach here!

PR #3939, opened 2022-07-20
update array.index for optional range

### LINKED ISSUE #3938: support array `index` range
## Feature

Support array `index` method's the optional arguments, start and stop

## Python Documentation

https://docs.python.org/3/library/array.html?highlight=array#array.array.index

## Relevant Test Case
https://github.com/RustPython/RustPython/pull/3910/files#diff-9171e2213953c56e609d6fae2f15892982341fd9f076fe69c4a73c3e52d7dfb9R933-R934

daeun503:
Fix #3938 
update array.index for optional range, start and stop

### DIFF
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -230,12 +230,18 @@ mod array {
                     }
                 }
 
-                fn index(&self, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
+                fn index(
+                    &self,
+                    obj: PyObjectRef,
+                    start: usize,
+                    stop: usize,
+                    vm: &VirtualMachine
+                ) -> PyResult<usize> {
                     match self {
                         $(ArrayContentType::$n(v) => {
                             if let Ok(val) = <$t>::try_into_from_object(vm, obj) {
-                                if let Some(pos) = v.iter().position(|&a| a == val) {
-                                    return Ok(pos);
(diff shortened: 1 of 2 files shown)


youknowone:
`MutObjectSequenceOp` is desinged for dynamic-length data structures. `array` operation will be more simpler like `tuple`.

stdlib/src/array.rs
```diff
@@ -230,12 +230,20 @@ mod array {
...
+                ) -> PyResult<usize> {
                     match self {
                         $(ArrayContentType::$n(v) => {
                             if let Ok(val) = <$t>::try_into_from_object(vm, obj) {
-                                if let Some(pos) = v.iter().position(|&a| a == val) {
-                                    return Ok(pos);
+                                for (index, element) in v.iter().enumerate().take(stop as usize).skip(start as usize) {
+                                    if element == &val {
+                                        return Ok(index);
+                                    }
```
youknowone, line 246: Can we use `position` again instead of manual comparision?

```suggestion
                                if let Some(pos) = v.iter().take(stop as _).skip(start as _).position(|&elem| elem == val) {
                                    return Ok(pos + start);
```

daeun503: I think your suggestion is very good
I'll try as you suggest. Thank you !

youknowone (approved):
looks good, thank you!

PR #3940, opened 2022-07-20
MutObjectSequenceOp find using Iteartor trait

youknowone:
to isolate control flow code from next() operation


qingshi163 (approved):
Looks Great, I assume no effect of the performance?

youknowone:
I didn't test but rust is quite good at optimizing iterators.
Do you want performance check?

qingshi163:
I don't think any noticeable performance difference, but have a check is aways better :)

qingshi163 (approved):

PR #3942, opened 2022-07-21
Change to raise SyntaxError for the generator without parenthsized

### LINKED ISSUE #3937: Generator expression without parenthesis
## Feature

In CPython, generator expression without parenthesis raises following error

```python3
>>> foo(x, z for z in range(10), t, w)
  File "<stdin>", line 1
    foo(x, z for z in range(10), t, w)
           ^^^^^^^^^^^^^^^^^^^^
SyntaxError: Generator expression must be parenthesized
```

However, In RustPython, it does not matter
```python3
>>>>> foo(x, z for z in range(10), t, w)
NameError: name 'foo' is not defined
```

key262yek:
Fix #3937


DimitrisJim:
I don't think the conflicts are going to be easy to resolve here (some recent prs to the grammar and the parser cover some of these changes). The parenthesized genexprs aren't as far as I can tell (the test in main is still marked as failing) so there's still something to be solved here.

DimitrisJim:
Should be moved to Parser, `test_genexps` is still failing so this PR is still useful.

youknowone:
Thank you for working on this, @key262yek! This issue has since been fixed in the current codebase — unparenthesized generator expressions now correctly raise SyntaxError. Closing as resolved.

PR #3946, opened 2022-07-21
Enable ssl by default for linux/macos

youknowone:
I believe openssl is always available on macos and major linux.


DimitrisJim:
ping, is this PR still relevant after the recent macOs shenanigans?

youknowone:
I am sorry. I have no idea. I just thought auto-enabling ssl is troublesome only for windows.

youknowone:
now we have rustls as a default ssl implementation.

PR #3948, opened 2022-07-22
Add unittest main in list_tests.py

TwoPair:


youknowone:
I don't think we will accept this patch.
First, this code didn't came from CPython code base - we only update `Lib/test` by following CPython.
Second, the file is not designed to be run itself. use `test_list.py` instead.

TwoPair:
Thank you for telling the reasons!

PR #3950, opened 2022-07-22
Fix list remove by referring to CPython

TwoPair:


TwoPair:
We are trying to resolve the test generated from CI.

TwoPair:
Incorrect access with problem

PR #3951, opened 2022-07-23
Fix `float` and `complex`'s incorrect hash value when using with `nan`

### LINKED ISSUE #3949: hash of `nan` return 0 instead of hash value
## Expected

```py
>>>>> value = float('nan')
>>>>> value
nan
>>>>> hash(value)
140454771093600 (some hash value)
```

## Actual

```py
>>>>> value = float('nan')
>>>>> value
nan
>>>>> hash(value)
0
```

## Failed test

https://github.com/RustPython/RustPython/pull/3910/files#diff-101957dfb9ef9c3a34ad39d7ad89a2d7d69290980553f9425acda0d5633ad453R585-R586

jopemachine:
Fixes #3949.

### DIFF
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -194,7 +194,7 @@ pub fn lcg_urandom(mut x: u32, buf: &mut [u8]) {
     }
 }
 
-pub fn hash_pointer_raw(p: *const libc::c_void) -> PyHash {
+pub fn hash_pointer_raw(p: *const std::ffi::c_void) -> PyHash {
     // TODO: Use commented logic when below issue resolved.
     // Ref: https://github.com/RustPython/RustPython/pull/3951#issuecomment-1193108966
 
@@ -206,6 +206,6 @@ pub fn hash_pointer_raw(p: *const libc::c_void) -> PyHash {
     p as PyHash
 }
 
-pub fn hash_pointer(p: *const libc::c_void) -> PyHash {
+pub fn hash_pointer(p: *const std::ffi::c_void) -> PyHash {
     fix_sentinel(hash_pointer_raw(p))
 }
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -1,10 +1,8 @@
 use num_bigint::BigInt;
-use num_complex::Complex64;
 use num_traits::ToPrimitive;
 use siphasher::sip::SipHasher24;
 use std::{
     hash::{BuildHasher, Hash, Hasher},
-    num::Wrapping,
 };
 
 pub type PyHash = i64;
@@ -138,24 +136,6 @@ pub fn hash_float(value: f64) -> Option<PyHash> {
     x = ((x << e) & MODULUS) | x >> (BITS32 - e);
 
     Some(fix_sentinel(x as PyHash * value.signum() as PyHash))
-}
-
-#[inline]
-pub fn hash_complex(value: &Complex64) -> PyHash {
-    let re_hash = (match hash_float(value.re) {
-        Some(value) => Some(value),
-        None => Some(hash_pointer(value as *const _ as *const std::ffi::c_void)),
-    })
-    .unwrap();
-
-    let im_hash = (match hash_float(value.im) {
-        Some(value) => Some(value),
-        None => Some(hash_pointer(value as *const _ as *const std::ffi::c_void)),
-    })
-    .unwrap();
-
-    let Wrapping(ret) = Wrapping(re_hash) + Wrapping(im_hash) * Wrapping(IMAG);
-    fix_sentinel(ret)
 }
 
 pub fn hash_iter_unordered<'a, T: 'a, I, F, E>(iter: I, hashf: F) -> Result<PyHash, E>
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -94,17 +94,18 @@ impl HashSecret {
     }
 }
 
-pub fn hash_float(value: f64) -> PyHash {
+#[inline]
+pub fn hash_float(value: f64) -> Option<PyHash> {
     // cpython _Py_HashDouble
     if !value.is_finite() {
         return if value.is_infinite() {
             if value > 0.0 {
-                INF
+                Some(INF)
             } else {
-                -INF
+                Some(-INF)
             }
         } else {
-            NAN
+            None
         };
     }
 
@@ -136,12 +137,12 @@ pub fn hash_float(value: f64) -> PyHash {
     };
     x = ((x << e) & MODULUS) | x >> (BITS32 - e);
 
-    fix_sentinel(x as PyHash * value.signum() as PyHash)
+    Some(fix_sentinel(x as PyHash * value.signum() as PyHash))
 }
 
 pub fn hash_complex(value: &Complex64) -> PyHash {
-    let re_hash = hash_float(value.re);
-    let im_hash = hash_float(value.im);
+    let re_hash = hash_float(value.re).unwrap();
+    let im_hash = hash_float(value.im).unwrap();
     let Wrapping(ret) = Wrapping(re_hash) + Wrapping(im_hash) * Wrapping(IMAG);
     fix_sentinel(ret)
 }
@@ -191,4 +192,20 @@ pub fn lcg_urandom(mut x: u32, buf: &mut [u8]) {
         x = x.wrapping_add(2531011);
         *b = ((x >> 16) & 0xff) as u8;
     }
+}
+
+pub fn hash_pointer_raw(p: *const libc::c_void) -> PyHash {
+    // TODO: Use commented logic when below issue resolved.
+    // Ref: https://github.com/RustPython/RustPython/pull/3951#issuecomment-1193108966
+
+    // let mut y = p as usize;
+    // /* bottom 3 or 4 bits are likely to be 0; rotate y by 4 to avoid
+    //    excessive hash collisions for dicts and sets */
+    // y = (y >> 4) | (y << (8 * std::mem::size_of::<usize>() - 4));
+    // y as PyHash
+    p as PyHash
+}
+
+pub fn hash_pointer(p: *const libc::c_void) -> PyHash {
+    fix_sentinel(hash_pointer_raw(p))
 }
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -140,6 +140,7 @@ pub fn hash_float(value: f64) -> Option<PyHash> {
     Some(fix_sentinel(x as PyHash * value.signum() as PyHash))
 }
 
+#[inline]
 pub fn hash_complex(value: &Complex64) -> Option<PyHash> {
     let re_hash = hash_float(value.re)?;
     let im_hash = hash_float(value.im)?;
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -95,11 +95,7 @@ pub fn hash_float(value: f64) -> Option<PyHash> {
     // cpython _Py_HashDouble
     if !value.is_finite() {
         return if value.is_infinite() {
-            if value > 0.0 {
-                Some(INF)
-            } else {
-                Some(-INF)
-            }
+            Some(if value > 0.0 { INF } else { -INF })
         } else {
             None
         };
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -1,9 +1,7 @@
 use num_bigint::BigInt;
 use num_traits::ToPrimitive;
 use siphasher::sip::SipHasher24;
-use std::{
-    hash::{BuildHasher, Hash, Hasher},
-};
+use std::hash::{BuildHasher, Hash, Hasher};
 
 pub type PyHash = i64;
 pub type PyUHash = u64;
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -140,11 +140,11 @@ pub fn hash_float(value: f64) -> Option<PyHash> {
     Some(fix_sentinel(x as PyHash * value.signum() as PyHash))
 }
 
-pub fn hash_complex(value: &Complex64) -> PyHash {
-    let re_hash = hash_float(value.re).unwrap();
-    let im_hash = hash_float(value.im).unwrap();
+pub fn hash_complex(value: &Complex64) -> Option<PyHash> {
+    let re_hash = hash_float(value.re)?;
+    let im_hash = hash_float(value.im)?;
     let Wrapping(ret) = Wrapping(re_hash) + Wrapping(im_hash) * Wrapping(IMAG);
-    fix_sentinel(ret)
+    Some(fix_sentinel(ret))
 }
 
 pub fn hash_iter_unordered<'a, T: 'a, I, F, E>(iter: I, hashf: F) -> Result<PyHash, E>
--- a/common/src/hash.rs
+++ b/common/src/hash.rs
@@ -185,7 +185,7 @@ pub fn hash_object_id_raw(p: usize) -> PyHash {
     // Ref: https://github.com/RustPython/RustPython/pull/3951#issuecomment-1193108966
 
     /* bottom 3 or 4 bits are likely to be 0; rotate y by 4 to avoid
-       excessive hash collisions for dicts and sets */
+    excessive hash collisions for dicts and sets */
     // p.rotate_right(4) as PyHash
     p as PyHash
 }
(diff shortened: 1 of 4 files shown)


jopemachine:
I'm wondering if I have to edit or remove the below line since I think the line will not be used.

https://github.com/RustPython/RustPython/blob/main/common/src/hash.rs#L107

Snowapril:
@jopemachine How about fixing `hash_float` which is in the link you attached. As `hash_complex` function use the same `hash_float` for its real number and imaginary number so we can fix them at the same time. We should return hash of pointer address if it is `nan`.

youknowone:
using `unreachable!()` macro is preferred way

jopemachine:
> How about fixing `hash_float` which is in the link you attached.

I'm trying to fix `hash_float`'s signature like `pub fn hash_float(arg: &Py<PyFloat>) -> PyHash`, but failed since the source code is different crate. Is there some good way to fix `hash_float` without changing the signature?

> We should return hash of pointer address if it is nan.

I'm trying to migrate the `hash_pointer` function from `cpython`.
It is like below code.
(I tried to use `libc::c_void` type because it seems rust does not have generic pointer type of c language.)

```rust
// common/src/hash.rs
pub fn hash_pointer_raw(p: *const libc::c_void) -> PyHash {
    let mut y = p as usize;
    /* bottom 3 or 4 bits are likely to be 0; rotate y by 4 to avoid
       excessive hash collisions for dicts and sets */
    y = (y >> 4) | (y << (8 * std::mem::size_of::<usize>() - 4));
    return y as PyHash;
}

pub fn hash_pointer(p: *const libc::c_void) -> PyHash {
    fix_sentinel(hash_pointer_raw(p))
}
...

// src/builitins/float.rs
#[inline]
fn hash(zelf: &crate::Py<Self>, _vm: &VirtualMachine) -> PyResult<hash::PyHash> {
    if zelf.to_f64().is_nan() {
        Ok(hash::hash_pointer(zelf as *const _ as *const libc::c_void))
    } else {
        Ok(hash::hash_float(zelf.to_f64()))
    }
}
```

but these code failed test because `object.__hash__` returns object id which maybe means reference address.
As mentioned in cpython's comment, `hash_pointer_raw` rotate 4 digits from the reference address.

```
>>>>> a = float('nan')
>>>>> object.__hash__(a)
140603489638352 // 11111111110000011001101000011011101011111010000
>>>>> hash(a)
8787718102397 // 1111111111000001100110100001101110101111101
>>>>> 
```

I'm not sure how to handle this.

Should I just use reference address as hash value instead of rotating one?
Could it cause some kind of compatibility issue with `cpython`?

youknowone:
About the crate problem, yes, we will not expose python types to the `common` crate.
My suggestion is turning `hash_float`'s return value to `Option<PyHash>` and making hashing `nan` returns `None`. Adding `#[inline]` is preferred to optimize this immediate Option type.
Then we can handle the None case later in python crates.

The value passed to hash_pointer is same to object id, obtainable by `AsObject::get_id()`. I didn't investigate how much `_Py_HashPointer` is used, but it can take `usize` for now. We can change it to more proper pointer type once we need later.

common/src/hash.rs
```diff
@@ -136,12 +137,12 @@ pub fn hash_float(value: f64) -> PyHash {
...
 
-    fix_sentinel(x as PyHash * value.signum() as PyHash)
+    Some(fix_sentinel(x as PyHash * value.signum() as PyHash))
 }
 
 pub fn hash_complex(value: &Complex64) -> PyHash {
-    let re_hash = hash_float(value.re);
-    let im_hash = hash_float(value.im);
+    let re_hash = hash_float(value.re).unwrap();
+    let im_hash = hash_float(value.im).unwrap();
```
Snowapril, line 145: This will cause panic if `complex(float('nan'), float('nan'))` is passed to builtin hash function.
```python
>>>>> hash(complex(float('nan'), float('nan')))
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', common/src/hash.rs:144:40
```
`re_hash` and `im_hash` also need to have `hash_pointer` value if the given value is `nan`.

common/src/hash.rs
```diff
@@ -136,14 +137,15 @@ pub fn hash_float(value: f64) -> PyHash {
...
+    Some(fix_sentinel(x as PyHash * value.signum() as PyHash))
 }
 
-pub fn hash_complex(value: &Complex64) -> PyHash {
-    let re_hash = hash_float(value.re);
-    let im_hash = hash_float(value.im);
+#[inline]
+pub fn hash_complex(value: &Complex64) -> Option<PyHash> {
+    let re_hash = hash_float(value.re)?;
+    let im_hash = hash_float(value.im)?;
```
Snowapril, line 146: `re_hash` and `im_hash` need not be propagated. They need to have a value of hash_pointer.

jopemachine: I'm sorry for my misunderstanding. Could you check new implementation is correct again?

jopemachine:
CI named `CI / Run tests under miri` is failing.
It says `IndexMap`'s generic argument length is not correct.
I don't have any clue about this... Could I get some advice about this?
Below is the error log. 

----

```
error[E0107]: this struct takes 3 generic arguments but 2 generic arguments were supplied
  --> vm/src/function/argument.rs:64:17
   |
64 |     pub kwargs: IndexMap<String, PyObjectRef>,
   |                 ^^^^^^^^ ------  ----------- supplied 2 generic arguments
   |                 |
   |                 expected 3 generic arguments
   |
note: struct defined here, with 3 generic parameters: `K`, `V`, `S`
  --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/indexmap-1.8.1/src/map.rs:76:12
   |
76 | pub struct IndexMap<K, V, S> {
   |            ^^^^^^^^ -  -  -
help: add missing generic argument
   |
64 |     pub kwargs: IndexMap<String, PyObjectRef, S>,
   |                                             +++
error[E0107]: this struct takes 3 generic arguments but 2 generic arguments were supplied
   --> vm/src/function/argument.rs:322:36
    |
322 | pub struct KwArgs<T = PyObjectRef>(IndexMap<String, T>);
    |                                    ^^^^^^^^ ------  - supplied 2 generic arguments
    |                                    |
    |                                    expected 3 generic arguments
    |
note: struct defined here, with 3 generic parameters: `K`, `V`, `S`
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/indexmap-1.8.1/src/map.rs:76:12
    |
76  | pub struct IndexMap<K, V, S> {
    |            ^^^^^^^^ -  -  -
help: add missing generic argument
    |
322 | pub struct KwArgs<T = PyObjectRef>(IndexMap<String, T, S>);
    |                                                      +++
error[E0107]: this struct takes 3 generic arguments but 2 generic arguments were supplied
   --> vm/src/function/argument.rs:325:21
    |
325 |     pub fn new(map: IndexMap<String, T>) -> Self {
    |                     ^^^^^^^^ ------  - supplied 2 generic arguments
    |                     |
    |                     expected 3 generic arguments
    |
note: struct defined here, with 3 generic parameters: `K`, `V`, `S`
   --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/indexmap-1.8.1/src/map.rs:76:12
    |
76  | pub struct IndexMap<K, V, S> {
    |            ^^^^^^^^ -  -  -
help: add missing generic argument
    |
325 |     pub fn new(map: IndexMap<String, T, S>) -> Self {
    |                                       +++
For more information about this error, try `rustc --explain E0107`.
error: could not compile `rustpython-vm` due to 3 previous errors
```

common/src/hash.rs
```diff
@@ -94,17 +94,18 @@ impl HashSecret {
...
     // cpython _Py_HashDouble
     if !value.is_finite() {
         return if value.is_infinite() {
             if value > 0.0 {
-                INF
+                Some(INF)
             } else {
-                -INF
+                Some(-INF)
             }
```
youknowone, line 106:
```rust
Some(if value > 0.0 {
    INF
} else {
    -INF
}
```

common/src/hash.rs
```diff
@@ -136,12 +137,23 @@ pub fn hash_float(value: f64) -> PyHash {
...
+    Some(fix_sentinel(x as PyHash * value.signum() as PyHash))
 }
 
+#[inline]
 pub fn hash_complex(value: &Complex64) -> PyHash {
-    let re_hash = hash_float(value.re);
-    let im_hash = hash_float(value.im);
+    let re_hash = (match hash_float(value.re) {
+        Some(value) => Some(value),
+        None => Some(hash_pointer(value as *const _ as *const std::ffi::c_void)),
```
youknowone, line 147: I thought the pointer to pass `hash_pointer` is `&PyObject`, not `&Complex64`

jopemachine: Thanks for pointing it out.
I think the easiest way is to pass `Py<PyComplex>`.
But the function also cannot use python type in the signature.
Should I move `hash_complex` implementation from `common/src` to `complex.rs`?
Or passing address itself as a second argument is acceptable like below?

```rust
pub fn hash_complex(value: &Complex64, ptr: *const std::ffi::c_void) -> PyHash
```

vm/src/builtins/complex.rs
```diff
@@ -417,7 +418,26 @@ impl Comparable for PyComplex {
...
-        Ok(hash::hash_complex(&zelf.value))
+        let value = zelf.value;
+
+        let re_hash = (match hash::hash_float(value.re) {
+            Some(value) => Some(value),
+            None => Some(hash::hash_pointer(
+                zelf as *const _ as *const std::ffi::c_void,
+            )),
+        })
+        .unwrap();
```
Snowapril, line 429:
```suggestion
        let re_hash = hash::hash_float(value.re)
            .unwrap_or_else(|| hash::hash_pointer(zelf as *const _ as *const std::ffi::c_void));
```

vm/src/builtins/complex.rs
```diff
@@ -417,7 +418,26 @@ impl Comparable for PyComplex {
...
+
+        let im_hash = (match hash::hash_float(value.im) {
+            Some(value) => Some(value),
+            None => Some(hash::hash_pointer(
+                zelf as *const _ as *const std::ffi::c_void,
+            )),
+        })
+        .unwrap();
```
Snowapril, line 437:
```suggestion
        let im_hash = hash::hash_float(value.im)
            .unwrap_or_else(|| hash::hash_pointer(zelf as *const _ as *const std::ffi::c_void));
```

vm/src/builtins/float.rs
```diff
@@ -539,7 +539,12 @@ impl Comparable for PyFloat {
 impl Hashable for PyFloat {
     #[inline]
     fn hash(zelf: &crate::Py<Self>, _vm: &VirtualMachine) -> PyResult<hash::PyHash> {
-        Ok(hash::hash_float(zelf.to_f64()))
+        match hash::hash_float(zelf.to_f64()) {
+            Some(value) => Ok(value),
+            None => Ok(hash::hash_pointer(
+                zelf as *const _ as *const std::ffi::c_void,
+            )),
+        }
```
Snowapril, line 547:
```suggestion
        Ok(hash::hash_float(zelf.to_f64())
            .unwrap_or_else(|| hash::hash_pointer(zelf as *const _ as *const libc::c_void)))
```

vm/src/builtins/float.rs
```diff
@@ -539,7 +539,8 @@ impl Comparable for PyFloat {
 impl Hashable for PyFloat {
     #[inline]
     fn hash(zelf: &crate::Py<Self>, _vm: &VirtualMachine) -> PyResult<hash::PyHash> {
-        Ok(hash::hash_float(zelf.to_f64()))
+        Ok(hash::hash_float(zelf.to_f64())
+            .unwrap_or_else(|| hash::hash_pointer(zelf as *const _ as *const libc::c_void)))
```
youknowone, line 543: Casting to raw pointer is somewhat unpleasure thing in Rust code.
If possible, we want to avoid it.
I checked all of the callers of _Py_HashPointer in cpython. Most of them were `&PyObject` value in RustPython except for method and contextvar. In this case, In this case, I prefer to have a function getting `&PyObject` instead of doing raw pointer casting here and there.
And, we already have the pointer value of PyObject. `AsObject::get_id` is the function to get it. (I left a comment about this too)

youknowone (changes requested):

vm/src/builtins/float.rs
jopemachine: > Casting to raw pointer is somewhat unpleasure thing in Rust code. If possible, we want to avoid it. I checked all of the callers of _Py_HashPointer in cpython. Most of them were `&PyObject` value in RustPython except for method and contextvar. In this case, In this case, I prefer to have a function getting `&PyObject` instead of doing raw pointer casting here and there. And, we already have the pointer value of PyObject. `AsObject::get_id` is the function to get it. (I left a comment about this too)

In cpython, `float('nan')`'s object id is not equals to the object's hashing value due to the bit rotating logic.

```
>>> import math
>>> hash(math.nan)
276386259  // 10000011110010101000111010011
>>> id(math.nan)
4422180144 // 100000111100101010001110100110000
>>> 
```

So, I think it would be good to have `hash_pointer` function somewhere for future compatibility.

Since `&PyObject` type is not available in `common/src` crate, how about moving the two functions under the `vm` crate to avoid the type issue?

youknowone: sure, that's because we didn't implement pointer hash compatibility.
In object.rs, hash is implemented like this
```
    fn slot_hash(zelf: &PyObject, _vm: &VirtualMachine) -> PyResult<PyHash> {
        Ok(zelf.get_id() as _)
    }
```
which is hash_pointer in CPython.

Of course we need new hash function for that, but I mean, not taking raw pointer but object id.
We don't need pointer for shift operation. That's an integer operation. In this PR, we need hash_object_id, which can take `usize`, instead of hash_pointer. We may add hash_pointer to take something other than `usize` for methods and contextvar, but they also will never take a c pointer - because that's not a rust type.

jopemachine: I appreciate for your detailed explanation :)

```rust
    fn slot_hash(zelf: &PyObject, _vm: &VirtualMachine) -> PyResult<PyHash> {
        Ok(hash_object_id(zelf.get_id() as _))
    }
```

Would it be good to apply this change in this PR?

common/src/hash.rs
```diff
@@ -192,3 +178,19 @@ pub fn lcg_urandom(mut x: u32, buf: &mut [u8]) {
...
+
+pub fn hash_object_id_raw(p: usize) -> PyHash {
+    // TODO: Use commented logic when below issue resolved.
+    // Ref: https://github.com/RustPython/RustPython/pull/3951#issuecomment-1193108966
+
+    // let mut y = p;
+    // /* bottom 3 or 4 bits are likely to be 0; rotate y by 4 to avoid
+    //    excessive hash collisions for dicts and sets */
+    // y = (y >> 4) | (y << (8 * std::mem::size_of::<usize>() - 4));
+    // y as PyHash
```
youknowone, line 190: Rust have [rotate_right](https://doc.rust-lang.org/std/primitive.u64.html#method.rotate_right) for this function

youknowone (commented):
it looks going well. I think fixing and adapting hash_object_id will be out of this PR's purpose.

common/src/hash.rs
```diff
@@ -192,3 +178,19 @@ pub fn lcg_urandom(mut x: u32, buf: &mut [u8]) {
         *b = ((x >> 16) & 0xff) as u8;
     }
 }
+
```
youknowone, line 181:
```
#[inline]
```
we need inline marker here

common/src/hash.rs
```diff
@@ -192,3 +178,19 @@ pub fn lcg_urandom(mut x: u32, buf: &mut [u8]) {
...
+    // Ref: https://github.com/RustPython/RustPython/pull/3951#issuecomment-1193108966
+
+    // let mut y = p;
+    // /* bottom 3 or 4 bits are likely to be 0; rotate y by 4 to avoid
+    //    excessive hash collisions for dicts and sets */
+    // y = (y >> 4) | (y << (8 * std::mem::size_of::<usize>() - 4));
+    // y as PyHash
+    p as PyHash
+}
+
```
youknowone, line 193:
```
#[inline]
```
we need inline marker here

youknowone (changes requested):
probably done with those small fixes

youknowone (approved):
Thank you for contributing and your long time effort for this feature

PR #3953, opened 2022-07-23
Fix: `range` beyond the isize range

### LINKED ISSUE #3936: `range` iterator pickling
## Feature

Implement a range's iter so that it can be dumped into the pickle module and reloaded

## Relevant Test Case
https://github.com/RustPython/RustPython/pull/3910/files#diff-34ede011e3fd5660af3db245be7bd645d4c15fac138b477ac4be9be57ae5d30cR378-R379

Yaminyam:
* Check if start + step is greater than sys.maxsize

Close #3936 
An attempt was made to implement the pickle-related function of the range iter, but the implementation was already completed.
The real problem is the difference in behavior with cpython.
In cpython, even if only start + step of range exceeds the range of int, longrange_iter is returned, but in rustpython, only start, stop, and step exceed the range of int.

### DIFF
--- a/Lib/test/test_range.py
+++ b/Lib/test/test_range.py
@@ -375,8 +375,6 @@ class RangeTest(unittest.TestCase):
                     self.assertEqual(list(pickle.loads(pickle.dumps(r, proto))),
                                      list(r))
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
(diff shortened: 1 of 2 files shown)


youknowone (approved):
nice catch! I now wonder what will be potential problem when start+step exceed the range

Issue #3726, opened 2022-05-20
Replace std::process::exit with ExitCode or Termination


youknowone:
- ExitCode: https://doc.rust-lang.org/1.61.0/std/process/struct.ExitCode.html
- Termination: https://doc.rust-lang.org/std/process/trait.Termination.html

Snowapril:
* `std::process::exit` : https://doc.rust-lang.org/std/process/fn.exit.html

PR #3956, opened 2022-07-24
replace exit with Exitcode

joohongpark:
issue #3726

### DIFF
--- a/examples/generator.rs
+++ b/examples/generator.rs
@@ -1,4 +1,5 @@
 use rustpython_vm as vm;
+use std::process::ExitCode;
 use vm::{
     builtins::PyIntRef,
     protocol::{PyIter, PyIterReturn},
@@ -40,10 +41,10 @@ gen()
     Ok(())
 }
 
-fn main() {
+fn main() -> ExitCode {
     let interp = vm::Interpreter::with_init(Default::default(), |vm| {
         vm.add_native_modules(rustpython_stdlib::get_module_inits());
     });
     let result = py_main(&interp);
-    std::process::exit(interp.run(|_vm| result));
+    ExitCode::from(interp.run(|_vm| result).to_be_bytes()[0])
 }
--- a/examples/generator.rs
+++ b/examples/generator.rs
@@ -46,5 +46,5 @@ fn main() -> ExitCode {
         vm.add_native_modules(rustpython_stdlib::get_module_inits());
     });
     let result = py_main(&interp);
-    ExitCode::from(interp.run(|_vm| result).to_be_bytes()[0])
+    ExitCode::from(interp.run(|_vm| result))
 }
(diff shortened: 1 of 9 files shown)


examples/generator.rs
```diff
@@ -40,10 +41,10 @@ gen()
...
 }
 
-fn main() {
+fn main() -> ExitCode {
     let interp = vm::Interpreter::with_init(Default::default(), |vm| {
         vm.add_native_modules(rustpython_stdlib::get_module_inits());
     });
     let result = py_main(&interp);
-    std::process::exit(interp.run(|_vm| result));
+    ExitCode::from(interp.run(|_vm| result).to_be_bytes()[0])
```
youknowone, line 49: I don't think we need `to_be_bytes()`. the exitcode was already int.

joohongpark:
The work that is difficult to reflect immediately has a TODO comment.

parser/build.rs
```diff
@@ -2,15 +2,19 @@ use std::fmt::Write as _;
...
+    let result = check_lalrpop("src/python.lalrpop", "src/python.rs");
+    if let Some(result) = result {
+        return result;
+    }
     gen_phf();
+    ExitCode::SUCCESS
 }
 
-fn check_lalrpop(source: &str, generated: &str) {
+fn check_lalrpop(source: &str, generated: &str) -> Option<ExitCode> {
```
youknowone, line 17: I think this function is expected to return Result instead of Option<ExitCode>

stdlib/src/posixsubprocess.rs
```diff
@@ -92,6 +92,7 @@ fn exec(args: &ForkExecArgs, procargs: ProcArgs) -> ! {
             let _ = write!(cur, "OSError:{}:", e as i32);
             let pos = cur.position();
             let _ = unistd::write(args.errpipe_write, &buf[..pos as usize]);
+            // TODO: Need to change to ExitCode or Termination
```
youknowone, line 95: I don't think this is expected to be ExitCode because it doesn't raise SystemExit

youknowone (changes requested):
I left a few comments, but other parts looks great!

parser/build.rs
joohongpark: There is a control statement that returns `None` in the same function, so isn't it possible to reach that code?

youknowone: Now, error case is encoded as `Some(ExitCode)` and successful case is encoded as `None`. This is typically encoded as Result type in Rust code like `Ok(())` and `Err(_)`. Give please give a proper error to `Err` case.

youknowone: Edit: because error type is not what this PR covers, `Result<(), ExitCode>` will be good enough.

youknowone (approved):
looks good, thank you!

PR #3967, opened 2022-07-25
Feat: grp module

Yaminyam:
add grp module and test

### DIFF
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -1999,6 +1999,7 @@ class POSIXProcessTestCase(BaseTestCase):
         with self.assertRaises(ValueError):
             subprocess.check_call(ZERO_RETURN_CMD, user=65535)
 
+    @unittest.expectedFailure
     @unittest.skipUnless(hasattr(os, 'setregid'), 'no setregid() on platform')
     def test_group(self):
         gid = os.getegid()
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -2048,7 +2048,7 @@ class POSIXProcessTestCase(BaseTestCase):
         with self.assertRaises(ValueError):
             subprocess.check_call(ZERO_RETURN_CMD, group=65535)
 
-    # TODO: RUSTPYTHON, SpamError: not enough spam
+    # TODO: RUSTPYTHON, observed gids do not match expected gids
     @unittest.expectedFailure
     @unittest.skipUnless(hasattr(os, 'setgroups'), 'no setgroups() on platform')
     def test_extra_groups(self):
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -2048,6 +2048,8 @@ class POSIXProcessTestCase(BaseTestCase):
         with self.assertRaises(ValueError):
             subprocess.check_call(ZERO_RETURN_CMD, group=65535)
 
+    # TODO: RUSTPYTHON, SpamError: not enough spam
+    @unittest.expectedFailure
     @unittest.skipUnless(hasattr(os, 'setgroups'), 'no setgroups() on platform')
     def test_extra_groups(self):
         gid = os.getegid()
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -1999,6 +1999,7 @@ class POSIXProcessTestCase(BaseTestCase):
         with self.assertRaises(ValueError):
             subprocess.check_call(ZERO_RETURN_CMD, user=65535)
 
+    # TODO: RUSTPYTHON, SpamError: not enough spam
     @unittest.expectedFailure
     @unittest.skipUnless(hasattr(os, 'setregid'), 'no setregid() on platform')
     def test_group(self):
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -1999,7 +1999,7 @@ class POSIXProcessTestCase(BaseTestCase):
         with self.assertRaises(ValueError):
             subprocess.check_call(ZERO_RETURN_CMD, user=65535)
 
-    # TODO: RUSTPYTHON, SpamError: not enough spam
+    # TODO: RUSTPYTHON, observed gids do not match expected gids
     @unittest.expectedFailure
     @unittest.skipUnless(hasattr(os, 'setregid'), 'no setregid() on platform')
     def test_group(self):
--- a/vm/src/stdlib/grp.rs
+++ b/vm/src/stdlib/grp.rs
@@ -47,16 +47,10 @@ mod grp {
             Ok(gid) => unistd::Group::from_gid(gid).map_err(|err| err.into_pyexception(vm))?,
             Err(_) => None,
         };
-        match group {
-            Some(group) => Ok(Group::from_unistd_group(group, vm)),
-            None => {
-                let message = vm
-                    .ctx
-                    .new_str(format!("getgrgid: group id {} not found", gid.as_bigint()))
-                    .into();
-                Err(vm.new_key_error(message))
-            }
-        }
+        let group = group.ok_or_else(|| {
+            vm.new_key_error(format!("getgrgid: group id {} not found", gid.as_bigint()))
+        })?;
+        Group::from_unistd_group(group, vm)
     }
 
     #[pyfunction]
--- a/vm/src/stdlib/grp.rs
+++ b/vm/src/stdlib/grp.rs
@@ -40,22 +40,6 @@ mod grp {
         }
     }
 
-    // impl From<unistd::Group> for Group {
-    //     fn from(group: unistd::Group) -> Self {
-    //         // this is just a pain...
-    //         let cstr_lossy = |s: std::ffi::CString| {
-    //             s.into_string()
-    //                 .unwrap_or_else(|e| e.into_cstring().to_string_lossy().into_owned())
-    //         };
-    //         Group {
-    //             gr_name: group.name,
-    //             gr_passwd: cstr_lossy(group.passwd),
-    //             gr_gid: group.gid.as_raw(),
-    //             gr_mem: group.mem,
-    //         }
-    //     }
-    // }
-
     #[pyfunction]
     fn getgrgid(gid: PyIntRef, vm: &VirtualMachine) -> PyResult<Group> {
         let gid_t = libc::gid_t::try_from(gid.as_bigint()).map(unistd::Gid::from_raw);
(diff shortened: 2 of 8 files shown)


Lib/test/test_subprocess.py
```diff
@@ -1999,6 +1999,7 @@ def test_user_error(self):
         with self.assertRaises(ValueError):
             subprocess.check_call(ZERO_RETURN_CMD, user=65535)
 
+    @unittest.expectedFailure
```
fanninpm, line 2002:
```suggestion
    # TODO: RUSTPYTHON, SpamError: not enough spam
    @unittest.expectedFailure
```

Lib/test/test_subprocess.py
```diff
@@ -2048,6 +2048,8 @@ def test_group_error(self):
         with self.assertRaises(ValueError):
             subprocess.check_call(ZERO_RETURN_CMD, group=65535)
 
+    # TODO: RUSTPYTHON, SpamError: not enough spam
+    @unittest.expectedFailure
```
fanninpm, line 2052: I put `SpamError: not enough spam` as a placeholder for you to provide the actual failure reason.

fanninpm: Is this an accurate summary? Feel free to amend it.

```suggestion
    # TODO: RUSTPYTHON, observed gids do not match expected gids
    @unittest.expectedFailure
```

Yaminyam: As far as I currently understand, yes
Thanks for clearing it up!

vm/src/stdlib/grp.rs
```diff
@@ -0,0 +1,113 @@
...
+    //         };
+    //         Group {
+    //             gr_name: group.name,
+    //             gr_passwd: cstr_lossy(group.passwd),
+    //             gr_gid: group.gid.as_raw(),
+    //             gr_mem: group.mem,
+    //         }
+    //     }
+    // }
+
```
youknowone, line 58:
```suggestion
```

vm/src/stdlib/grp.rs
```diff
@@ -0,0 +1,113 @@
...
+    #[pyfunction]
+    fn getgrgid(gid: PyIntRef, vm: &VirtualMachine) -> PyResult<Group> {
+        let gid_t = libc::gid_t::try_from(gid.as_bigint()).map(unistd::Gid::from_raw);
+        let group = match gid_t {
+            Ok(gid) => unistd::Group::from_gid(gid).map_err(|err| err.into_pyexception(vm))?,
+            Err(_) => None,
+        };
```
youknowone, line 65:
```suggestion
        let gid = libc::gid_t::try_from(gid.as_bigint()).map(unistd::Gid::from_raw);
        let group = gid.ok().map(|gid| unistd::Group::from_gid(gid)).transpose()?;
```

vm/src/stdlib/grp.rs
```diff
@@ -0,0 +1,113 @@
...
+        match group {
+            Some(group) => Ok(Group::from_unistd_group(group, vm)),
+            None => {
+                let message = vm
+                    .ctx
+                    .new_str(format!("getgrgid: group id {} not found", gid.as_bigint()))
+                    .into();
+                Err(vm.new_key_error(message))
+            }
+        }
```
youknowone, line 75:
```suggestion
        let group = group.ok_or_else(|| {
            vm.new_key_error(format!("getgrgid: group id {} not found", gid.as_bigint()))
        })?;
        Group::from_unistd_group(group, vm)
```

vm/src/stdlib/grp.rs
```diff
@@ -0,0 +1,113 @@
...
+        let group = match gid_t {
+            Ok(gid) => unistd::Group::from_gid(gid).map_err(|err| err.into_pyexception(vm))?,
+            Err(_) => None,
+        };
+        match group {
+            Some(group) => Ok(Group::from_unistd_group(group, vm)),
+            None => {
+                let message = vm
+                    .ctx
+                    .new_str(format!("getgrgid: group id {} not found", gid.as_bigint()))
```
youknowone, line 71: this PyStr creation is not necessary

vm/src/stdlib/grp.rs
```diff
@@ -0,0 +1,113 @@
...
+    #[pyfunction]
+    fn getgrall(vm: &VirtualMachine) -> PyResult<Vec<PyObjectRef>> {
+        // setgrent, getgrent, etc are not thread safe. Could use fgetgrent_r, but this is easier
+        static GETGRALL: parking_lot::Mutex<()> = parking_lot::const_mutex(());
+        let _guard = GETGRALL.lock();
+        let mut list = Vec::new();
+
+        unsafe { libc::setgrent() };
+        while let Some(ptr) = NonNull::new(unsafe { libc::getgrent() }) {
+            let _group = unistd::Group::from(unsafe { ptr.as_ref() });
```
youknowone, line 105:
```suggestion
            let group = unistd::Group::from(unsafe { ptr.as_ref() });
```
`_` prefixed variable means unused one in Rust.

youknowone (changes requested):

vm/src/stdlib/grp.rs
Yaminyam: I forgot I didn't delete this, thank you

vm/src/stdlib/grp.rs
Yaminyam: This function cannot be called because from_unistd_group receives a group that is not optional<group> as an argument.

vm/src/stdlib/grp.rs
Yaminyam: If the following is performed, the type of the result is Option<Option<Group> instead of Option<Group>.

vm/src/stdlib/grp.rs
Yaminyam: new_key_error receives String type as an argument

vm/src/stdlib/grp.rs
```diff
@@ -0,0 +1,88 @@
...
+            return Err(exceptions::cstring_error(vm));
+        }
+        match unistd::Group::from_name(name.as_str()).map_err(|err| err.into_pyexception(vm))? {
+            Some(group) => Ok(Group::from_unistd_group(group, vm)),
+            None => {
+                let name_repr = name.as_object().repr(vm)?;
+                let message = vm
+                    .ctx
+                    .new_str(format!("getgrnam(): name not found: {}", name_repr))
+                    .into();
```
youknowone, line 65: this new_str is also not necessary like getgrgid

youknowone (changes requested):

fanninpm:
Are you sure this belongs in the `stdlib` crate (rather than the `vm` crate)?

youknowone:
yes, `grp` is not a builtin module.

```
>>> import sys
>>> sys.builtin_module_names
('_abc', '_ast', '_codecs', '_collections', '_functools', '_imp', '_io', '_locale', '_operator', '_peg_parser', '_signal', '_sre', '_stat', '_string', '_symtable', '_thread', '_tracemalloc', '_warnings', '_weakref', 'atexit', 'builtins', 'errno', 'faulthandler', 'gc', 'itertools', 'marshal', 'posix', 'pwd', 'sys', 'time', 'xxsubtype')
>>> 'grp' in sys.builtin_module_names
False
>>> 
```

youknowone:
@Yaminyam oh, no. the communication was poor. grp is not a builtin module. So it belongs to `stdlib` by following other library convention. You didn't need to move it to vm. Actually, it is better to be placed in `stdlib` as previous one.

vm/src/stdlib/grp.rs
Yaminyam: Thank you, I have read the material you sent.
and_then() seems to receive a function as an argument, but shouldn't it be implemented so that option<option<string>> can be made into option<string> in that function?

fanninpm: I think @MrGreenTea meant `flatten`.

https://doc.rust-lang.org/std/option/enum.Option.html#method.flatten

Yaminyam: I'm not sure how, but why not just do the unwrap() twice?
`let group = gid.ok().map(|gid| unistd::Group::from_gid(gid)).transpose().unwrap().unwrap();`

youknowone: Oh, I didn't catch git_t was used later. `gid_t` is a type name. Could you give it another proper name?

Always avoid unwrap unless it never happens. We are making interpreter. Our users want to see Python exceptions instead of Rust error when they failed but `unwrap()` panics in Rust stack.

Yaminyam: Since gid should be used later and gid_t is created only for .ok, I think it would be better to use it without making it a variable.

So, how to safely unwrap an option<option<string>> object into an option<string>?
`let group = gid_t.ok().map(|gid| unistd::Group::from_gid(gid)).transpose()?.flatten();`
If you use something like this, the problem with ?

MrGreenTea: I believe it could be

```rs
gid_t.ok().and_then(|gid| unistd::Group::from_gid(gid)).transpose()?
```

youknowone: Well, I didn't look in the code carefully when I left last comment. `gid_t` can be safely replaced by `gid` because it is never used after its last usage to create `git`. You can reuse same variable names again and again if you don't need the previous names later.

Line 50 expect `group` to be `Option`. `transpose()?.flatten()` will work.

youknowone: @MrGreenTea That doesn't looking work because one is Option and the other one is Result.

fanninpm: If in doubt, follow the types.

```rust
let group = gid_t                                         // Result<T, E>
                 .ok()                                    // Option<T>
                 .map(|gid| unistd::Group::from_gid(gid)) // Option<Result<Option<Group>>>
                 .transpose()                             // Result<Option<Option<Group>>, nix::errno::Errno>
                 ?                                        // Option<Option<Group>>
                 .flatten();                              // Option<Group>
```

Does RustPython have a way to convert `nix::errno::Errno` into some sort of Exception?

Yaminyam: I understand what you said.
However, it is not possible to use gid as a variable name because it has a gid in the argument in the same form as getgrgid(gid, vm) and needs to be used later.

And since gid_t will not be used later, I think it will be okay to use it without declaring it as a variable.

Yaminyam: @fanninpm In the code you mentioned, the following error occurs in `?`
`grp.rs(44, 56): expected `core::PyRef<exceptions::types::PyBaseException>` because of this`

fanninpm: You would need a call to `map_err()` between `transpose()` and `?` to convert `nix::errno::Errno` into something compatible with `PyBaseException`.

Yaminyam: @fanninpm I was wondering if the code related to the exception disappeared as the code was changed, but this is the problem! Thanks it has been fixed thanks!

youknowone (approved):
looks good, thank you for long time effort!

PR #3968, opened 2022-07-25
Update `_mut_iter_equal_skeleton` method

TwoPair:
This code is tested based on `list_tests.py::CommonTest::test_remove`,
except for lines `352~358` that seem related to `str` test.

### DIFF
--- a/Lib/test/test_list.py
+++ b/Lib/test/test_list.py
@@ -231,15 +231,5 @@ class ListTest(list_tests.CommonTest):
         lst = [X(), X()]
         X() in lst
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
-    def test_count(self):  # XXX: RUSTPYTHON; the method also need to be removed when done
-        super().test_count()
-
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
-    def test_contains_fake(self):  # XXX: RUSTPYTHON; the method also need to be removed when done
-        super().test_contains_fake()
-
 if __name__ == "__main__":
     unittest.main()
--- a/vm/src/sequence.rs
+++ b/vm/src/sequence.rs
@@ -1,10 +1,6 @@
 use crate::{
-    builtins::PyIntRef,
-    function::{Either, OptionalArg, PyComparisonValue},
-    sliceable::SequenceIndexOp,
-    types::{richcompare_wrapper, PyComparisonOp, RichCompareFunc},
-    vm::VirtualMachine,
-    AsObject, PyObject, PyObjectRef, PyResult,
+    builtins::PyIntRef, function::OptionalArg, sliceable::SequenceIndexOp, types::PyComparisonOp,
+    vm::VirtualMachine, AsObject, PyObject, PyObjectRef, PyResult,
 };
 use optional::Optioned;
 use std::ops::Range;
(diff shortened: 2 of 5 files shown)


youknowone:
line 352 is not related to string. try to print its type like `print(type(self.type2test))` that's list type. e.g. `list`.
`list('abcdefghcij')` is `['a', 'b', 'c', 'd', 'e', 'f', 'g', 'h', 'c', 'i', 'j']`

youknowone:
By watching CI, it is causing dead-lock. it didn't finish during 4 hours.

TwoPair:
Now, two list method `index`, `remove` works normally by fixing `_mut_iter_equal_skeleton` method.

vm/src/sequence.rs
```diff
@@ -81,91 +72,10 @@ pub trait MutObjectSequenceOp<'a> {
...
-                            }
-                        }
-                    }
+                let eq = {
+                    let elem = elem.clone();
+                    drop(guard);
+                    elem.rich_compare_bool(needle, PyComparisonOp::Eq, vm)?
                 };
 
                 if eq {
```
Snowapril, line 81:
```suggestion
                let elem = elem.clone();
                drop(guard);
                
                if elem.rich_compare_bool(needle, PyComparisonOp::Eq, vm)? {
```

Snowapril:
Nice catch! The changed codes seem totally fine to me. Please check unexpected success in `test_list`, `test_deque` tests.

Lib/test/test_list.py
```diff
@@ -231,13 +231,13 @@ def __eq__(self, other):
         lst = [X(), X()]
         X() in lst
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+    # # TODO: RUSTPYTHON
+    # @unittest.expectedFailure
     def test_count(self):  # XXX: RUSTPYTHON; the method also need to be removed when done
         super().test_count()
```
fanninpm:
```suggestion
```

Lib/test/test_list.py
```diff
@@ -231,13 +231,13 @@ def __eq__(self, other):
...
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+    # # TODO: RUSTPYTHON
+    # @unittest.expectedFailure
     def test_contains_fake(self):  # XXX: RUSTPYTHON; the method also need to be removed when done
         super().test_contains_fake()
```
fanninpm:
```suggestion
```

fanninpm (changes requested):
Please don't leave these things behind in the code. They're not there in CPython.

fanninpm (commented):
LGTM.

youknowone:
@qingshi163 could you review this PR?

qingshi163:
As I remember, the guard or borrower is to optimize the compare by check if the class is not the heap type, so we can reuse the compare function in each loop. Also to use rich_compare_bool we cannot reuse the PyTypeRef for the compared object but this will be no more the issue if we can fetch the PyType atomicly in the future.

Yes, the correctness is more important than performance, but if you can solve the problem without drawback the performance could be amazing. if not please advise the issue or the failing point of the current implement, so we could see why this is not going to work.

@youknowone 

Thanks

youknowone (approved):
@qingshi163 I created an issue to refer this pr #4008
@TwoPair Thank you!

PR #3971, opened 2022-07-25
Fix sre-engine wrong repeat context used when multiple max_until recusion

### LINKED ISSUE #2793: pip support in macos
```sh
$ rustpython --install-pip
downloading get-pip.py...
running get-pip.py...
Collecting pip
  Using cached pip-21.2.2-py3-none-any.whl (1.6 MB)
Collecting setuptools
  Using cached setuptools-57.4.0-py3-none-any.whl (819 kB)
Collecting wheel
  Using cached wheel-0.36.2-py2.py3-none-any.whl (35 kB)
Installing collected packages: wheel, setuptools, pip
Successfully installed pip-21.2.2 setuptools-57.4.0 wheel-0.36.2
$ rustpython -m pip
~/.cargo/bin/rustpython: No module named pip
```

it looks like installed successfully, but still `no module named pip` happens

### LINKED ISSUE #3544: Pip installation issue on MacOS 12.0.2
The error message is quite long, so here's the last bit:

```
SyntaxError: Got error in f-string: Invalid expression: invalid syntax. Got unexpected token ',' at line 408 column 27
        completed_str = f"{completed_ratio:,.{precision}f}"
```

This happens whether I build a version with `ssl` or run `get-pip.py` normally.

### LINKED ISSUE #3829: [EPIC] PIP IS BROKEN 🛬💥😭
_Author's note: This was originally meant to be compiled in a blog post once pip was working again, but since people have repeatedly submitted GitHub issues complaining about pip not working, I decided to write and pin this issue._

# Background

On 29 January 2022, [pip v22.0 was released](https://github.com/pypa/pip/blob/main/NEWS.rst#220-2022-01-29). This broke our CI workflow because pip vendors dependencies that we don't currently support. (In that same update, pip dropped support for Python 3.6, which can introduce more incompatibilities, as there are parts of our codebase that haven't been updated since Python 3.6, such as `asyncio`.)

# Progress so far

- In #3531, I reported a problem in the way that RustPython parsed "format specifier expressions" in f-strings. In CPython's f-strings, it's possible to format numbers with a width and a precision. However, RustPython's f-strings wouldn't let you use variables in the context of those width and precision values. [With the help of VSCode and LLDB](https://github.com/RustPython/RustPython/wiki/Debugger-with-VSCode), I engineered a fix and submitted #3532.
- While I was working on the previous problem, I noticed that pip needed the `mmap` module in order to function. In #3564, @youknowone contributed a (nonfunctional) skeleton for the `mmap` module, which was abandoned shortly thereafter. (I suppose this effort stalled because we needed a way to abstract over both read-only and read-write memory maps, and `memmap2` doesn't have a trait to extract over both the `Mmap` and the `MmapMut` structs. However, I could be wrong.) Thankfully, @killme2008 contributed #3755, taking a considerable amount of effort and time to make the `mmap` module functional (on macOS and Linux)
- Next, it was noticed that `get-pip.py` was taking too long, so @youknowone identified and fixed a deadlock in #3792.
- Finally, the `get-pip.py` script was working, but it couldn't find a `pip` to download, and it wasn't leaving much of a trail to trace the root cause. (Even when pip was installed with `ensurepip`, pip would still fail to install a package from PyPI.) Thankfully, @dieterplex narrowed the problem down to [the HTML parser](https://github.com/RustPython/RustPython/pull/3795#issuecomment-1162563665). A little later, the HTML parser's problem was [isolated to within the `re` module](https://github.com/RustPython/RustPython/issues/3808#issuecomment-1164395303). **As of 29 June 2022, this is the extent of the current progress.**

If you were mentioned in the previous bullet points, thank you very much for helping this along.

# How you can help

- Please consider leaving an emoji reaction (or a comment) on this issue instead of filing a new issue. Filing new issues tends to drown out other issues that other people might find more approachable.
- A good way to get started contributing code is to update a Python module (and its corresponding test suite, if possible) to the latest stable version of CPython (v3.10 as of when I'm writing this). You can let a maintainer know if you're running into problems.
- From what I have seen, Windows doesn't get nearly as much attention as Linux or macOS. If you are a Windows user, I invite you to help RustPython expand its Windows support.

(_Maintainers, please feel free to edit this issue as you see fit, especially if I misrepresented something._)

qingshi163:
fix #3829 
fix #3544 
fix #2793

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2066,8 +2066,9 @@ dependencies = [
 
 [[package]]
 name = "sre-engine"
-version = "0.1.2"
-source = "git+https://github.com/qingshi163/sre-engine?branch=refactor#82675ae5bd541139163c33312f84e421dddcdd1e"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "55e283f0ec6488739d0b972e3c17b70a8698b33c298a169430387f871af51a03"
 dependencies = [
  "bitflags",
  "num_enum",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2067,8 +2067,7 @@ dependencies = [
 [[package]]
 name = "sre-engine"
 version = "0.1.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e5872399287c284fed4bc773cb7f6041623ac88213774f5e11e89e2131681fc1"
+source = "git+https://github.com/qingshi163/sre-engine?branch=refactor#82675ae5bd541139163c33312f84e421dddcdd1e"
 dependencies = [
  "bitflags",
  "num_enum",
--- a/Lib/test/test_htmlparser.py
+++ b/Lib/test/test_htmlparser.py
@@ -112,8 +112,6 @@ class HTMLParserTestCase(TestCaseBase):
             ("pi", "processing instruction ?"),
             ])
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_simple_html(self):
         self._run_check("""
 <!DOCTYPE html PUBLIC 'foo'>
@@ -258,8 +256,6 @@ text
             ("endtag", "p"),
             ])
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_get_starttag_text(self):
         s = """<foo:bar   \n   one="1"\ttwo=2   >"""
         self._run_check_extra(s, [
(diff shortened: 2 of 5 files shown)


youknowone:
pip is fixed with this PR

qingshi163:
@youknowone we need publish the new release on sre-engine to merge this pr

youknowone:
awesome, I published it. That's really weird you don't have permission to the crates.

youknowone (approved):

dieterplex:
Just a reminder that since https://github.com/RustPython/sre-engine/pull/10 got merged and published, should it be tagged with new version in sre-engine repo?

youknowone:
Thank you for reminding! I pushed tag 0.2.0

PR #3974, opened 2022-07-26
Add: _uuid module

rimi0108:
- Add _uuid module
- Add _uuid function generate_time_safe()
- uuid v1 = generate_time_safe
- Return mac address to node_id if you have mac address or return random generated bytes to node_id if you do not have mac address

### DIFF
--- a/stdlib/Cargo.toml
+++ b/stdlib/Cargo.toml
@@ -65,7 +65,7 @@ num_enum = "0.5.7"
 ascii = "1.0.0"
 
 # uuid
-[target.'cfg(not(all(target_os = "ios", target_os = "android")))'.dependencies]
+[target.'cfg(not(any(target_os = "ios", target_os = "android")))'.dependencies]
 mac_address = "1.1.3"
 once_cell = "1.13.0"
 uuid = { version = "1.1.2", features = ["v1", "fast-rng", "macro-diagnostics"] }
--- a/stdlib/Cargo.toml
+++ b/stdlib/Cargo.toml
@@ -63,6 +63,8 @@ ahash = "0.7.6"
 libz-sys = { version = "1.1.5", optional = true }
 num_enum = "0.5.7"
 ascii = "1.0.0"
+once_cell = "1.13.0"
+mac_address = "1.1.3"
 
 # mmap
 [target.'cfg(all(unix, not(target_arch = "wasm32")))'.dependencies]
@@ -111,3 +113,11 @@ features = [
 
 [target.'cfg(target_os = "macos")'.dependencies]
 system-configuration = "0.5.0"
+
+[dependencies.uuid]
+version = "1.1.2"
+features = [
+    "v1",                # Lets you generate random UUIDs
+    "fast-rng",          # Use a faster (but still sufficiently random) RNG
+    "macro-diagnostics", # Enable better diagnostics for compile-time UUIDs
+]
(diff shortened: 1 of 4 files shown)


stdlib/Cargo.toml
```diff
@@ -63,6 +63,8 @@ ahash = "0.7.6"
 libz-sys = { version = "1.1.5", optional = true }
 num_enum = "0.5.7"
 ascii = "1.0.0"
+once_cell = "1.13.0"
+mac_address = "1.1.3"
```
youknowone, line 67: Adding this line instead of `dependencies.uuid` section
will be more easy to read

```
uuid = { version = "1.1.2", features = ["v1", "fast-rng", "macro-diagnostics"] }
```

stdlib/src/lib.rs
```diff
@@ -51,6 +51,8 @@ mod select;
 mod ssl;
 #[cfg(all(unix, not(target_os = "redox")))]
 mod termios;
+#[cfg(target_os = "macos")]
```
youknowone, line 54: what's the blocker for  linux and windows?

stdlib/src/uuid.rs
```diff
@@ -0,0 +1,53 @@
+pub(crate) use _uuid::make_module;
+
+#[pymodule]
+mod _uuid {
+
```
youknowone, line 5:
```suggestion
```

stdlib/src/uuid.rs
```diff
@@ -0,0 +1,53 @@
...
+            }
+            Ok(None) => {
+                let node_id = NODE_ID.get_or_init(|| rand::thread_rng().gen::<[u8; 6]>());
+                node_id
+            }
+            Err(_e) => {
+                let node_id = NODE_ID.get_or_init(|| rand::thread_rng().gen::<[u8; 6]>());
+                node_id
+            }
+        }
```
youknowone, line 32: this is checking `get_mac_address` every time. That probably is not expectation for node id. `NODE_ID.get_or_init` must be the outmost wrapper of this function

youknowone (changes requested):

youknowone:
`_uuid.has_uuid_generate_time_safe`  seems required to let the module run

youknowone (changes requested):

stdlib/src/uuid.rs
```diff
@@ -0,0 +1,62 @@
...
+
+        (
+            Uuid::new_v1(ts, &unique_node_id).as_bytes().to_vec(),
+            PyNone,
+        )
+    }
+
+    #[pyattr]
+    fn has_uuid_generate_time_safe(_vm: &VirtualMachine) -> PyInt {
+        PyInt::from(0)
```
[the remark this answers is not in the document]
rimi0108: I wrote the code like that because it was printed below the cpython
`>>> _uuid.has_uuid_generate_time_safe
0`

stdlib/Cargo.toml
youknowone: `mac_address` needs more cfg guard for more OS.
See `mmap` case. By referring CI failure, we (at least) need to exclude iOS and android.

youknowone (changes requested):
You seem didn't set git email correctly. Check `git log` and if it matches your actual email (and github email)

youknowone (approved):
looks good, thank you for long time effort

PR #3975, opened 2022-07-27
Precision bug fix when `FormatType` is `None`

gilteunchoi:
In this revision, 
* Bugs about precision handling malfunction are fixed.

---
- [X] 1) Precision handing is not working when the `FormatType` is `None`.

Before:
```
>>>>> print('{:.2}'.format(math.pi))
3.141592653589793
```
After:
```
>>>>> print('{:.2}'.format(math.pi))
3.1
```

---
- [x] 2) Must includes at least one digit past the decimal point. 

RUSTPython:
```
print('{}'.format(60.0)) // 60 
print('{}'.format(120.0)) // 120 
```
CPython:
```
print('{}'.format(60.0)) // 60.0
print('{}'.format(120.0)) // 120.0
```

---
- [x] 3) `self.precision.unwrap_or(6);` should not apply when `FormatType` is `None`.

RUSTPython:
```
print('{}'.format(math.pi)) // 3.14159
```
CPython:
```
print('{}'.format(math.pi)) // 3.141592653589793
```

---
- [x] 4)  Must use scientific notation even if the rounded number does not have fractional part.

RUSTPython:
```
print('{:.2}'.format(math.pi)) // 3.1
print('{:.2}'.format(math.pi * 10)) // 31
print('{:.2}'.format(math.pi * 100)) // 3.1e+02
```
CPython:
```
print('{:.2}'.format(math.pi)) // 3.1
print('{:.2}'.format(math.pi * 10)) // 3.1e+01
print('{:.2}'.format(math.pi * 100)) // 3.1e+02
```

### DIFF
--- a/common/src/float_ops.rs
+++ b/common/src/float_ops.rs
@@ -188,6 +188,7 @@ pub fn format_general(
     magnitude: f64,
     case: Case,
     alternate_form: bool,
+    always_shows_fract: bool,
 ) -> String {
     match magnitude {
         magnitude if magnitude.is_finite() => {
@@ -195,7 +196,7 @@ pub fn format_general(
             let mut parts = r_exp.splitn(2, 'e');
             let base = parts.next().unwrap();
             let exponent = parts.next().unwrap().parse::<i64>().unwrap();
-            if exponent < -4 || exponent >= (precision as i64) {
+            if exponent < -4 || exponent + (always_shows_fract as i64) >= (precision as i64) {
                 let e = match case {
                     Case::Lower => 'e',
                     Case::Upper => 'E',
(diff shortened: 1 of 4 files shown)


common/src/float_ops.rs
```diff
@@ -188,14 +188,17 @@ pub fn format_general(
     magnitude: f64,
     case: Case,
     alternate_form: bool,
+    is_format_type_none: Option<bool>,
```
youknowone, line 191: by concept, this function doesn't recognize about format type. adding a parameter using this name doesn't look well fit and breaking abstraction layer.
And `Some(false)` and `None` doesn't have difference for this function.

```suggestion
    always_shows_fract: bool,
```

youknowone (changes requested):

youknowone (approved):
Thank you!

Issue #3820, opened 2022-06-26
Regular expressions implementation is slow on large strings


DCNick3:
Trying to run [pytube](https://github.com/pytube/pytube) inside rust-python seems to work, but is very slow (I think just retrieving a list of video streams takes > 10 minutes when it's about 4 seconds with CPython).

Tapping into the process with a debugger I was able to figure out that most of the time is spent in regular expressions code that is implemented in `sre_engine` crate. The expressions are relatively simple, like `ytInitialPlayerResponse\s*=\s*` (though some are more complex like `(?:\b|[^a-zA-Z0-9$])(?P<sig>[a-zA-Z0-9$]{2})\s*=\s*function\(\s*a\s*\)\s*{\s*a\s*=\s*a\.split\(\s*""\s*\)`), but are being run on the whole html page, which is usually 0.5-2 MiB in size.

It takes more than 30 seconds to perform one match like this and it's very slow compared to what CPython seems to be doing.

DCNick3:
Tried playing around with string being just a repetition of "a"'s and regex being `.*blabla`. The growth in execution time with growth of input string seems to be quadratic. Works fine with everything <8KiB, but later the search time starts growing rapidly

youknowone:
is it better than before?

DCNick3:
It is much better now: 3.48 vs 12.35 seconds

Not ideal, but at least it's usable now

PR #3976, opened 2022-07-27
Test Bump sre-engine to 0.2.1

qingshi163:


qingshi163:
@youknowone we need publish the sre-engine v0.2.1 for merge this pr.

youknowone:
is this fixing #3820 ?

qingshi163:
> is this fixing #3820 ?

I didn't check to run pytube, that huge time spend looks more like a bug instead just performance.

for performance I am trying to catch up but two things cannot compete with cpython.

goto vs function pointer call
fixed width char vs utf-8

beside that rust should do well with much more clear and readbility code.

this pr should improve performance slightly.

BTW coroutines should be tried when stable.

youknowone (approved):

PR #3979, opened 2022-07-28
Fix `PyFloat::py_new` always returning new float object issue

### LINKED ISSUE #3978: `PyFloat::py_new` always return new float object even if the argument's oid is same.
# Summary

`PyFloat::py_new` always return new float object even if the argument's oid is same.

## Expected

```
>>> import math
>>> a = math.nan
>>> id(float(a))
4358871344
>>> a= math.nan
>>> id(float(a))
4358871344
>>> id(float(a))
4358871344
>>> id(float(a))
4358871344
>>> id(float(a))
4358871344
```

## Actual

```
>>> import math
>>> a = math.nan
>>>>> id(float(a))
140470641517808
>>>>> id(float(a))
140470639935376
>>>>> id(float(a))
140470629849984
>>>>> id(float(a))
140470642896176
>>>>> id(float(a))
140470639935376
```

jopemachine:
Fixes #3978.

```
>>>>> import math
>>>>> a = float(math.nan)
>>>>> b = float(math.nan)
>>>>> a is b
>>>>> true
```

## Reference

- https://github.com/python/cpython/blob/main/Objects/abstract.c#L1633-L1635

- https://github.com/RustPython/RustPython/blob/main/vm/src/builtins/int.rs#L258-L264

### DIFF
--- a/extra_tests/snippets/builtin_float.py
+++ b/extra_tests/snippets/builtin_float.py
@@ -441,6 +441,8 @@ assert float('inf').hex() == 'inf'
 assert float('-inf').hex() == '-inf'
 assert float('nan').hex() == 'nan'
 
+assert float(math.nan) is float(math.nan)
+
 # Test float exponent:
 assert 1 if 1else 0 == 1
 
--- a/vm/src/builtins/float.rs
+++ b/vm/src/builtins/float.rs
@@ -142,8 +142,7 @@ impl Constructor for PyFloat {
         let float_val = match arg {
             OptionalArg::Missing => 0.0,
             OptionalArg::Present(val) => {
-                if cls.is(vm.ctx.types.float_type) && val.payload_if_exact::<PyFloat>(vm).is_some()
-                {
+                if cls.is(vm.ctx.types.float_type) && val.class().is(vm.ctx.types.float_type) {
                     return Ok(val);
                 }
 
--- a/vm/src/builtins/float.rs
+++ b/vm/src/builtins/float.rs
@@ -142,7 +142,7 @@ impl Constructor for PyFloat {
         let float_val = match arg {
             OptionalArg::Missing => 0.0,
             OptionalArg::Present(val) => {
-                if let Some(_) = val.payload_if_exact::<PyFloat>(vm) {
+                if val.payload_if_exact::<PyFloat>(vm).is_some() {
                     return Ok(val);
                 } else if let Some(f) = val.try_float_opt(vm)? {
                     f.value
--- a/vm/src/builtins/float.rs
+++ b/vm/src/builtins/float.rs
@@ -142,9 +142,14 @@ impl Constructor for PyFloat {
         let float_val = match arg {
             OptionalArg::Missing => 0.0,
             OptionalArg::Present(val) => {
-                if val.payload_if_exact::<PyFloat>(vm).is_some() {
-                    return Ok(val);
-                } else if let Some(f) = val.try_float_opt(vm)? {
+                if cls.is(vm.ctx.types.float_type) {
+                    return match val.downcast_exact::<PyFloat>(vm) {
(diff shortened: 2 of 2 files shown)


Snowapril:
It seems below incompatibility make extra_test failed.
```python
class F(float):
    def __int__(self):
        return 3
# rustpython
>>>>> type(F(1.2))
<class 'float'>
# cpython
>>> type(F(1.2))
<class '__main__.F'>
```

Snowapril:
We need to move the exact float type check into `PyNumber::float` and return its `payload` with ref. Please refer to what `payload_if_exact` return and use `into_ref` 😊

jopemachine:
> We need to move the exact float type check into `PyNumber::float` and return its `payload` with ref. Please refer to what `payload_if_exact` return and use `into_ref` 😊

`number::float` is not called when the argument is `math.nan` because `try_float_opt` [early returned in this line](https://github.com/RustPython/RustPython/blob/main/vm/src/protocol/number.rs#L102).

So I tried to edit `try_float_opt` instead like below, but it seems it is not working as expected. (`id(float(math.nan))` prints different oid each time)

```rust
pub fn try_float_opt(&self, vm: &VirtualMachine) -> PyResult<Option<PyRef<PyFloat>>> {
	let value = if let Some(float) = self.payload_if_exact::<PyFloat>(vm) {
			Some(float.into_ref(vm))
	} else {
			let number = self.to_number();
			#[allow(clippy::manual_map)]
			if let Some(f) = number.float(vm)? {
					Some(f)
			} else if let Some(i) = self.try_index_opt(vm) {
					let value = int::try_to_float(i?.as_bigint(), vm)?;
					Some(vm.ctx.new_float(value))
			} else if let Some(value) = self.downcast_ref::<PyFloat>() {
					Some(vm.ctx.new_float(value.to_f64()))
			} else {
					None
			}
	};
	Ok(value)
}
``` 

I guess this is because below line returned new float instance instead of existing one.

https://github.com/RustPython/RustPython/blob/main/vm/src/builtins/float.rs#L152-L154

So, I think maybe we need to update the `py_new` function somehow.

jopemachine:
> We need to move the exact float type check into `PyNumber::float` and return its `payload` with ref. Please refer to what `payload_if_exact` return and use `into_ref` 😊

`number::float` is not called when the argument is `math.nan` because `try_float_opt` [early returned in this line](https://github.com/RustPython/RustPython/blob/main/vm/src/protocol/number.rs#L102).

So I tried to edit `try_float_opt` instead like below, but it seems it is not working as expected. (`id(float(math.nan))` prints different oid each time)

```rust
pub fn try_float_opt(&self, vm: &VirtualMachine) -> PyResult<Option<PyRef<PyFloat>>> {
	let value = if let Some(float) = self.payload_if_exact::<PyFloat>(vm) {
			Some(float.into_ref(vm))
	} else {
			let number = self.to_number();
			#[allow(clippy::manual_map)]
			if let Some(f) = number.float(vm)? {
					Some(f)
			} else if let Some(i) = self.try_index_opt(vm) {
					let value = int::try_to_float(i?.as_bigint(), vm)?;
					Some(vm.ctx.new_float(value))
			} else if let Some(value) = self.downcast_ref::<PyFloat>() {
					Some(vm.ctx.new_float(value.to_f64()))
			} else {
					None
			}
	};
	Ok(value)
}
``` 

I guess this is because below line returned new float instance instead of existing one.

https://github.com/RustPython/RustPython/blob/main/vm/src/builtins/float.rs#L152-L154

So, I think maybe we need to update the `py_new` function somehow.

youknowone:
you are right. you will find PyInt have a similar logic you are looking for

youknowone:
looks good, would you add a related test to extra_tests?

youknowone (approved):
Thank you!

PR #3980, opened 2022-07-28
Fix: check if name contains null byte

Yaminyam:
In `getpwnam()`, a function used to find user information based on name, name cannot contain null characters.

However, this has not been verified in tests.


youknowone (approved):
nice catch!

fanninpm:
> However, this has not been verified in tests.

Remind me to add something to the `extra_tests`, as I can't find this case in the CPython test suite (from a cursory look).

Yaminyam:
@fanninpm So I asked cpython's test to add that
https://github.com/python/cpython/pull/95394

PR #3981, opened 2022-07-28
Add correct method for ImportError.__str__()

fanninpm:


fanninpm:
I added this as a warmup for adapting `ImportError.__reduce__()`.

youknowone:
by checking CPython's `ImportError_methods`, it only has `__reduce__` as methods. Do we need it because other RustPython side glitches?

fanninpm:
I forgot to check the current behavior of `str(ImportError("msg"))`.

fanninpm:
As it turns out, this seems to be completely unnecessary.

PR #3984, opened 2022-07-29
Fix `instancecheck`, `subclasscheck` of `UnionType`

TwoPair:
Resolve `test_types::UnionTests::test_instancecheck, test_bad_instancecheck,test_bad_subclasscheck`

### DIFF
--- a/vm/src/builtins/union.rs
+++ b/vm/src/builtins/union.rs
@@ -90,15 +90,33 @@ impl PyUnion {
     }
 
     #[pymethod(magic)]
-    fn instancecheck(_zelf: PyRef<Self>, _obj: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        Err(vm
-            .new_type_error("isinstance() argument 2 cannot be a parameterized generic".to_owned()))
+    fn instancecheck(zelf: PyRef<Self>, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<bool> {
+        if zelf
+            .args
+            .iter()
+            .any(|x| x.class().is(vm.ctx.types.generic_alias_type))
+        {
+            Err(vm.new_type_error(
+                "isinstance() argument 2 cannot be a parameterized generic".to_owned(),
+            ))
+        } else {
+            obj.is_instance(zelf.args().as_object(), vm)
+        }
     }
 
     #[pymethod(magic)]
-    fn subclasscheck(_zelf: PyRef<Self>, _obj: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        Err(vm
-            .new_type_error("issubclass() argument 2 cannot be a parameterized generic".to_owned()))
+    fn subclasscheck(zelf: PyRef<Self>, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<bool> {
+        if zelf
+            .args
+            .iter()
+            .any(|x| x.class().is(vm.ctx.types.generic_alias_type))
+        {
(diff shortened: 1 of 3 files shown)


youknowone:

vm/src/builtins/union.rs
```diff
@@ -90,15 +90,13 @@ impl PyUnion {
...
+        obj.is_instance(zelf.args().as_object(), vm)
     }
 
     #[pymethod(magic)]
-    fn subclasscheck(_zelf: PyRef<Self>, _obj: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        Err(vm
-            .new_type_error("issubclass() argument 2 cannot be a parameterized generic".to_owned()))
+    fn subclasscheck(zelf: PyRef<Self>, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<bool> {
+        obj.is_subclass(zelf.args().as_object(), vm)
     }
```
Snowapril, line 100: We need to check whether `GenericAlias` object exist in args and then return `TypeError` with message "isinstance() argument 2 cannot contain a parameterized generic"

vm/src/builtins/union.rs
```diff
@@ -90,15 +90,13 @@ impl PyUnion {
...
+    fn instancecheck(zelf: PyRef<Self>, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<bool> {
+        obj.is_instance(zelf.args().as_object(), vm)
     }
 
     #[pymethod(magic)]
-    fn subclasscheck(_zelf: PyRef<Self>, _obj: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        Err(vm
-            .new_type_error("issubclass() argument 2 cannot be a parameterized generic".to_owned()))
+    fn subclasscheck(zelf: PyRef<Self>, obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<bool> {
+        obj.is_subclass(zelf.args().as_object(), vm)
```
Snowapril, line 99: We need to check whether `GenericAlias` object exist in args and then return `TypeError` with message "isinstance() argument 2 cannot contain a parameterized generic"

```python
>>>>> isinstance(1, int | list[str])
# rustpython
True
# cpython
TypeError: isinstance() argument 2 cannot contain a parameterized generic
```

Snowapril: Hmm... As `is_instance` and `is_subclass` in the `object.rs` are already checking each argument in the tuple, so we don't have to iterate args and call `is_instance` or `is_subclass`. I also agree with your implmenetation.

But in cpython, they switched to the implementation that iterates all the args in `union` and call `is_instance` or `is_subclass` for each one. (https://github.com/python/cpython/pull/27120)

I don't know the exact reason yet, haha.. 😂

TwoPair: Thx for recognizing me!

TwoPair:
I made `test_types.py` update by referring to [CPython 3.10's one](https://github.com/python/cpython/blob/0579a187df9ca5d1bd928eec710c29aa40acd826/Lib/test/test_types.py#L704-L759) for more precise checks.

Snowapril (approved):

youknowone:
The changes looks great, thank you! Just one more request. Could you rebase this PR to 2 commits - one for test changes and the other one for non-test code changes?
When I'd going to squash all commits to single commit, I usually can do it my self with github UI, but not when I wants more than 2 commits.

youknowone (approved):
Great! Thank you!

PR #3986, opened 2022-07-30
Fix metaclass check classcell

oow214:
- [x] Fix `__build_class__` check `__class__` cell
- [x] Fix `type.__new__` set classcell

When a class object is created with metaclass, the following operation is performed.
[https://docs.python.org/3/reference/datamodel.html?highlight=__classcell__#:~:text=CPython implementation detail%3A In,a RuntimeError in Python 3.8](https://docs.python.org/3/reference/datamodel.html?highlight=__classcell__#:~:text=CPython%20implementation%20detail%3A%20In,a%20RuntimeError%20in%20Python%203.8).

**CPython implementation detail:** In CPython 3.6 and later, the `__class__` cell is passed to the metaclass as a `__classcell__` entry in the class namespace. If present, this must be propagated up to the `type.__new__` call in order for the class to be initialised correctly. Failing to do so will result in a [RuntimeError] in Python 3.8.
1. Error handling when `__class__` (the class it belongs to) and metaclass are different ([bltinmodule.c](https://github.com/python/cpython/blob/3.10/Python/bltinmodule.c#L225-L242))
2. If the result in step 1 is the same, set class in `type.__new__` ([typeobject.c](https://github.com/python/cpython/blob/3.10/Objects/typeobject.c#L3030-L3056))

But in Rust Python it sets classcell without comparing `__class__`.
```rust
if let Some(ref classcell) = classcell {
	  classcell.set(Some(class.clone()));
}
```

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -624,21 +624,15 @@ impl PyType {
         let typ = Self::new_verbose_ref(name.as_str(), base, bases, attributes, slots, metatype)
             .map_err(|e| vm.new_type_error(e))?;
 
-        typ.attributes
-            .write()
-            .get(identifier!(vm, __classcell__))
-            .map(|cell| {
-                if let Ok(cell) = PyCellRef::try_from_object(vm, cell.clone()) {
-                    cell.set(Some(typ.clone().to_pyobject(vm)));
-                    Ok(())
-                } else {
-                    Err(vm.new_type_error(format!(
-                        "__classcell__ must be a nonlocal cell, not {}",
-                        cell.class()
-                    )))
-                }
-            })
-            .transpose()?;
+        if let Some(cell) = typ.attributes.write().get(identifier!(vm, __classcell__)) {
+            let cell = PyCellRef::try_from_object(vm, cell.clone()).map_err(|_| {
+                vm.new_type_error(format!(
+                    "__classcell__ must be a nonlocal cell, not {}",
+                    cell.class().name()
+                ))
+            })?;
+            cell.set(Some(typ.clone().to_pyobject(vm)));
+        };
 
         // avoid deadlock
         let attributes = typ
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -8,13 +8,15 @@ use crate::common::{
     lock::{PyRwLock, PyRwLockReadGuard},
 };
 use crate::{
+    builtins::function::PyCellRef,
     builtins::PyBaseExceptionRef,
     class::{PyClassImpl, StaticType},
+    convert::ToPyObject,
     function::{FuncArgs, KwArgs, OptionalArg, PySetterValue},
     identifier,
     protocol::PyNumberMethods,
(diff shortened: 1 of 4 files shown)


Snowapril:
Unless we locate the code for setting type to classcell in `type.__new__`, build failure will not be disappeared 😂

vm/src/builtins/type.rs
```diff
@@ -594,6 +596,19 @@ impl PyType {
...
+        let cell = typ
+            .attributes
+            .write()
+            .get(__classcell__)
+            .map(|cell| PyCellRef::try_from_object(vm, cell.clone()));
+
+        if let Some(Ok(cell)) = cell {
+            cell.set(Some(typ.clone().to_pyobject(vm)));
+        }
+
```
Snowapril, line 611:
```suggestion
        typ.attributes
            .write()
            .get(identifier!(vm, __classcell__))
            .map(|cell| {
                if let Some(cell) = PyCellRef::try_from_object(vm, cell.clone()) {
                    cell.set(Some(typ.clone().to_pyobject(vm)));
                    Ok(())
                } else {
                    Err(vm.new_type_error(format!(
                        "__classcell__ must be a nonlocal cell, not {}",
                        cell.class()
                    )))
                }
            })
            .transpose()?;

```
It seems that we miss to raise type error when the given `__classcell__` is actually not a cell type

Snowapril (approved):
BTW, without that change request, looks great !! 😊

vm/src/builtins/type.rs
```diff
@@ -594,6 +596,21 @@ impl PyType {
...
+            .write()
+            .get(identifier!(vm, __classcell__))
+            .map(|cell| {
+                let cell = PyCellRef::try_from_object(vm, cell.clone()).map_err(|_| {
+                    vm.new_type_error(format!(
+                        "__classcell__ must be a nonlocal cell, not {}",
+                        cell.class().name()
+                    ))
+                })?;
+                cell.set(Some(typ.clone().to_pyobject(vm)));
```
youknowone, line 609: This line has side-effect and the closure doesn't return anything. Using if-let statement would be better than map

youknowone (commented):
The logic looks good, but I changed error-handling style in https://github.com/RustPython/RustPython/pull/3986/commits/edb4a9614bac4399246d1fa91bfb451ed41f2c28

`PyObject` is not expected to be used as Display object because it generates not very helpful error messages.
This was on TODO list for long time but forgotten. Once #4045 is merged, please rebase this PR and fix error formatting arguments.

Thank you for fixing this hair-pulling problem! Now I am try to getting how it is working.

vm/src/builtins/type.rs
youknowone: I think previous version is better. `cell.set` has side effect and the result of `map` is not used. This is more like control flow than data stream. Using `if-let` will show the flow better and will not hide side-effect of `cell.set`.

Snowapril: Oh I agree. Then we only need to add type error handling

youknowone (approved):

PR #3987, opened 2022-07-31
Fix `cmath` to handle some edge cases in complex number calculation

jirheee:
Make  `test_user_object` and `test_abs_overflows` pass for `test_cmath.py`

Related to #1671: Increase CPython test coverage


jirheee:
I realized I had made a mistake, so I fixed it!

vm/src/builtins/complex.rs
```diff
@@ -228,9 +228,15 @@ impl PyComplex {
     }
 
     #[pymethod(magic)]
-    fn abs(&self) -> f64 {
+    fn abs(&self, vm: &VirtualMachine) -> PyResult<PyArithmeticValue<f64>> {
```
youknowone, line 231:
```suggestion
    fn abs(&self, vm: &VirtualMachine) -> PyResult<f64> {
```
Because `PyArithmeticValue::NotImplemented` is not used, `PyResult<f64>` will be enough.

youknowone (changes requested):

Lib/test/test_cmath.py
```diff
@@ -536,9 +534,6 @@ def test_abs(self):
         self.assertEqual(abs(complex(INF, NAN)), INF)
         self.assertTrue(math.isnan(abs(complex(NAN, NAN))))
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
-    @requires_IEEE_754
```
youknowone, line 541: The upper 2 lines are our temporary change for test, but the last line is not.
Please put `@requires_IEEE_754` back 😄

jirheee: Got it! I restored the `@requires_IEEE_754 ` decorator :)

vm/src/builtins/complex.rs
jirheee: Oh yeah, you're right. I just fixed and pushed it again 👍

youknowone (approved):
Thank you for contributing!

PR #3989, opened 2022-08-01
Update cmp of `PyUnion`

TwoPair:
Resolve `test_types::UnionTests::test_or_types_operator`.


TwoPair:
This PR could be continued when #3990 is merged.
Cuz the comparison way of set type including every dictdatatype has a problem of itself arising from union hashing process.

youknowone (approved):
Great! Thank you!

PR #3990, opened 2022-08-02
Correct union hash

### LINKED ISSUE #3851: `PyUnionType` `hash` should be consistent regardless of the order of arguments.
### CPython3.10
```python
>>> (str|int).__args__
(<class 'str'>, <class 'int'>)
>>> (int|str).__args__
(<class 'int'>, <class 'str'>)
>>> hash(str|int)
-2443872088029664091
>>> hash(int|str)
-2443872088029664091
```
### Python3.10
```python
>>>>> (str|int).__args__
(<class 'str'>, <class 'int'>)
>>>>> (int|str).__args__
(<class 'int'>, <class 'str'>)
>>>>> hash(str | int)
93930046793440
>>>>> hash(int | str)
93930046797200
```
This is related with `test_types::UnionTests::test_hash`.

moreal:
This pull request resolves #3851.

### DIFF
--- a/Lib/test/test_types.py
+++ b/Lib/test/test_types.py
@@ -703,8 +703,6 @@ class UnionTests(unittest.TestCase):
         y.__args__ = [str, int]
         self.assertEqual(x, y)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_hash(self):
         self.assertEqual(hash(int | str), hash(str | int))
         self.assertEqual(hash(int | str), hash(typing.Union[int, str]))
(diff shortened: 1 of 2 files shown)


youknowone (approved):
Nice! what behavior is changed? I can't guess how the bug is fixed 😵

moreal:
> Nice! what behavior is changed? I can't guess how the bug is fixed 😵

Though it was already merged, as a reply, the `PyTuple::iter` returns `PyResult` and it is `Result`. The `Result` implements `IntoIterator` so the `PyFrozenSet::from_iter` doesn't work well. It may expect `PyFrozenset::from_iter(vm, [1, 2])` but it was `PyFronzenSet::from_iter(vm, [[1, 2]])`.

PR #3991, opened 2022-08-02
Refactor: pwd module

Yaminyam:
The refactored contents while developing the grp module are also reflected in the pwd module.

### DIFF
--- a/vm/src/stdlib/pwd.rs
+++ b/vm/src/stdlib/pwd.rs
@@ -56,36 +56,36 @@ mod pwd {
         if name.as_str().contains('\0') {
             return Err(exceptions::cstring_error(vm));
         }
-        match User::from_name(name.as_str()).map_err(|err| err.into_pyexception(vm))? {
-            Some(user) => Ok(Passwd::from(user)),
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/pwd.rs
```diff
@@ -56,36 +56,36 @@ mod pwd {
...
             return Err(exceptions::cstring_error(vm));
         }
-        match User::from_name(name.as_str()).map_err(|err| err.into_pyexception(vm))? {
-            Some(user) => Ok(Passwd::from(user)),
-            None => {
-                let name_repr = name.as_object().repr(vm)?;
-                let message = vm
-                    .ctx
+        let user = User::from_name(name.as_str()).map_err(|err| err.into_pyexception(vm))?;
+        let name_repr = name.as_object().repr(vm)?;
```
youknowone, line 60: `name_repr` is only used for `user.is_none()` case. this change will make redundant repr call

youknowone (changes requested):

youknowone (approved):
thank you!

PR #4003, opened 2022-08-06
Implemente type __qualname__ setter

### LINKED ISSUE #3998: Implemente type `__qualname__` setter
## Feature

Implemente type `__qualname__` setter

```python
class test: 
    ...

test.__qualname__ = "test"
```

## Python Documentation

https://docs.python.org/3/library/stdtypes.html#special-attributes

daeun503:
Fix #3998

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -405,6 +405,34 @@ impl PyType {
             .unwrap_or_else(|| vm.ctx.new_str(self.name().deref()).into())
     }
 
+    #[pyproperty(magic, setter)]
+    fn set_qualname(&self, value: PySetterValue, vm: &VirtualMachine) -> PyResult<()> {
+        // TODO: we should replace heaptype flag check to immutable flag check
+        if !self.slots.flags.has_feature(PyTypeFlags::HEAPTYPE) {
+            return Err(vm.new_type_error(format!(
+                "cannot set '__qualname__' attribute of immutable type '{}'",
+                self.name()
+            )));
+        };
+        let value = value.ok_or_else(|| {
+            vm.new_type_error(format!(
+                "cannot delete '__qualname__' attribute of immutable type '{}'",
+                self.name()
+            ))
+        })?;
+        if !value.class().fast_issubclass(vm.ctx.types.str_type) {
+            return Err(vm.new_type_error(format!(
+                "can only assign string to {}.__qualname__, not '{}'",
+                self.name(),
+                value.class().name()
+            )));
+        }
+        self.attributes
+            .write()
+            .insert(identifier!(vm, __qualname__), value);
+        Ok(())
+    }
+
     #[pyproperty(magic)]
     pub fn module(&self, vm: &VirtualMachine) -> PyObjectRef {
         self.attributes
--- a/Lib/test/test_pickle.py
+++ b/Lib/test/test_pickle.py
@@ -146,8 +146,6 @@ class PyPicklerTests(AbstractPickleTests, unittest.TestCase):
     def test_py_methods(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_py_methods() # TODO: RUSTPYTHON, remove when this passes
 
-    # TODO: RUSTPYTHON, AttributeError: attribute '__qualname__' of 'type' objects is not writable
-    @unittest.expectedFailure
     def test_recursive_nested_names(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_recursive_nested_names() # TODO: RUSTPYTHON, remove when this passes
 
@@ -251,8 +249,6 @@ class InMemoryPickleTests(AbstractPickleTests, AbstractUnpickleTests,
     def test_py_methods(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_py_methods() # TODO: RUSTPYTHON, remove when this passes
 
-    # TODO: RUSTPYTHON, AttributeError: attribute '__qualname__' of 'type' objects is not writable
-    @unittest.expectedFailure
     def test_recursive_nested_names(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_recursive_nested_names() # TODO: RUSTPYTHON, remove when this passes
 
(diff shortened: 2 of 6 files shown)


vm/src/builtins/type.rs
```diff
@@ -405,6 +405,35 @@ impl PyType {
...
+    #[pyproperty(magic, setter)]
+    fn set_qualname(&self, value: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        if value.is(&vm.ctx.none()) {
+            return Err(vm.new_type_error(format!(
+                "cannot delete '{}' attribute of immutable type '{}'",
+                "__qualname__",
+                self.name()
+            )));
+        };
+        if !self.slots.flags.has_feature(PyTypeFlags::HEAPTYPE) {
```
Snowapril, line 417: This sound to me at now as all static types have IMMUTABLE flag.
But i think we need to add comment about it. We need to replace this HEAPTYPE flag check to IMMUTABLE flag check in future.

vm/src/builtins/type.rs
```diff
@@ -405,6 +405,35 @@ impl PyType {
...
 
+    #[pyproperty(magic, setter)]
+    fn set_qualname(&self, value: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        if value.is(&vm.ctx.none()) {
+            return Err(vm.new_type_error(format!(
+                "cannot delete '{}' attribute of immutable type '{}'",
+                "__qualname__",
+                self.name()
+            )));
+        };
```
Snowapril, line 416:
```suggestion
    fn set_qualname(&self, value: Option<PyObjectRef>, vm: &VirtualMachine) -> PyResult<()> {
        let value = value.ok_or_else(|| vm.new_type_error(format!(
            "cannot delete '{}' attribute of immutable type '{}'",
            "__qualname__",
            self.name()
        )))?;
```
We can also pass `Option<PyObjectRef>` for separating whether it is a setter call or a delete call.

vm/src/builtins/type.rs
```diff
@@ -405,6 +405,35 @@ impl PyType {
...
+        value.downcast_ref::<PyStr>().ok_or_else(|| {
+            vm.new_type_error(format!(
+                "can only assign string to {}.__qualname__, not '{}'",
+                self.name(),
+                value.class().name()
+            ))
+        })?;
```
Snowapril, line 430:
```suggestion
        if value.payload_if_subclass::<PyStr>(vm).is_none() {
            return Err(vm.new_type_error(format!(
                "can only assign string to {}.__qualname__, not '{}'",
                self.name(),
                value.class().name()
            )));
        }
```
It might be a useless temporary return that occurred in `downcast_ref`. As we need to check is subclass of `PyStr`, I think this is enough 😊

vm/src/builtins/type.rs
```diff
@@ -405,6 +405,35 @@ impl PyType {
...
+            return Err(vm.new_type_error(format!(
+                "cannot delete '{}' attribute of immutable type '{}'",
+                "__qualname__",
+                self.name()
+            )));
+        };
+        if !self.slots.flags.has_feature(PyTypeFlags::HEAPTYPE) {
+            return Err(vm.new_type_error(format!(
+                "cannot set '{}' attribute of immutable type '{}'",
+                "__qualname__",
```
Snowapril, line 420:
```suggestion
                "cannot set '__qualname__' attribute of immutable type '{}'",
```

vm/src/builtins/type.rs
```diff
@@ -405,6 +405,35 @@ impl PyType {
...
+                self.name()
+            )));
+        };
```
Snowapril, line 423: Maybe it is trivial, It seems this `HEAPTYPE` flag check should precede a `None` check according to cpython's work.
```python
# cpython
>>> del int.__qualname__
TypeError: cannot set '__qualname__' attribute of immutable type 'int'
# rustpython
>>>>> del int.__qualname__
TypeError: cannot delete '__qualname__' attribute of immutable type 'int'
```

vm/src/builtins/type.rs
youknowone: Oh, I think we don't have proper `PySetterFunc` to handle deleter case

youknowone: To distinguish setting None and delete, #4010 is required.

Snowapril: Oh I missed that setting None also possible 😂

daeun503: Oh.... #4010 has not been merged yet, so I will correct it when it is merged. Thank you.

vm/src/builtins/type.rs
daeun503: Thank you. I'll fix it !

vm/src/builtins/type.rs
daeun503: That's a better idea. Thank you.

vm/src/builtins/type.rs
daeun503: Oh, I missed it. I will check `HEAPTYPE` before `None`. Thank you

vm/src/builtins/type.rs
daeun503: I added comment as you advised. Thank you

vm/src/builtins/type.rs
youknowone: #4010 is done

youknowone:
Please also remove this lines which must be fixed by this PR 😄 

https://github.com/RustPython/RustPython/blob/0156c097d1a96b3cd9cf2d1d59d58ce6b8187f86/Lib/typing.py#L2286-L2289

vm/src/builtins/type.rs
daeun503: I checked it. Thank you !

daeun503:
> Please also remove this lines which must be fixed by this PR 😄

I deleted the line from `typing.py`. Thank you !

vm/src/builtins/type.rs
```diff
@@ -405,6 +405,34 @@ impl PyType {
...
+        let value = value.ok_or_else(|| {
+            vm.new_type_error(format!(
+                "cannot delete '__qualname__' attribute of immutable type '{}'",
+                self.name()
+            ))
+        })?;
+        if value.payload_if_subclass::<PyStr>(vm).is_none() {
```
youknowone, line 423: if you don't need payload, try to use `fast_issubclass` instead of this function

daeun503: Oh.... I will correct it according to your advice. Thank you.

Snowapril (approved):
Looks great!! 😊

youknowone (approved):
Looks great! You finally got it!

PR #4012, opened 2022-08-07
disable LTO for CI

youknowone:
https://github.com/RustPython/RustPython/issues/1978#issuecomment-1207262094


DimitrisJim:
for compile time improvements? Do you remember?

youknowone:
Yes, it was. But it needs measurements if it actually reduce CI time or not.

PR #4019, opened 2022-08-08
Update test_urllib* from CPython 3.10.5

youknowone:


Lib/test/test_urllib.py
```diff
@@ -1537,9 +1537,10 @@ def test_quoting(self):
                          "url2pathname() failed; %s != %s" %
                          (expect, result))
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
fanninpm, line 1541: Why is this test erroring/failing?

youknowone: I don't know. But it was expected failure before updating. So it will fails on windows without it. See 1542 line before patch.

PR #4035, opened 2022-08-10
Fix making union parameters

TwoPair:
Resolve `test_types::UnionTests::test_union_parameter_chaining`

### DIFF
--- a/vm/src/builtins/union.rs
+++ b/vm/src/builtins/union.rs
@@ -1,6 +1,6 @@
 use super::{genericalias, type_};
 use crate::{
-    builtins::{PyFrozenSet, PyStr, PyStrRef, PyTuple, PyTupleRef, PyType, PyTypeRef},
+    builtins::{PyFrozenSet, PyStr, PyStrRef, PyTuple, PyTupleRef, PyType},
     class::PyClassImpl,
     common::hash,
     convert::ToPyObject,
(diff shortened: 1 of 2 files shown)


vm/src/builtins/union.rs
```diff
@@ -196,21 +196,10 @@ fn dedup_and_flatten_args(args: PyTupleRef, vm: &VirtualMachine) -> PyTupleRef {
...
-                        .rich_compare_bool(arg, PyComparisonOp::Eq, vm)
-                        .ok()
-                        .unwrap()
-                }
-                _ => param.is(arg),
-            }
+            param
+                .rich_compare_bool(arg, PyComparisonOp::Eq, vm)
+                .ok()
+                .unwrap()
```
youknowone, line 202: For `Result` type, `.ok().unwrap()` equals to `unwrap()`.
But why unwrap? If there is no special guarantee `unwrap` is safe, propagate it. Otherwise use expect with message about the guarantee.

youknowone (changes requested):

vm/src/builtins/union.rs
youknowone: Oh, the unwrap part looks like a bug, but it already existed in previous version.

youknowone (approved):

youknowone:
Thank you!

vm/src/builtins/union.rs
TwoPair: I tested again with your feedback, and figured out `ok()` isn't neccessary.
So I made pr #4036 for this.

PR #4036, opened 2022-08-10
Refactor making union parameters 

TwoPair:
#4035


youknowone:
would you also replace unwrap to `?` please? that will cause crash which usually we don't want

TwoPair:
> would you also replace unwrap to `?` please? that will cause crash which usually we don't want

If I did that, I shoud do refactoring every parts using `dedup_and_flatten_args` function.
Is that really what you intend?

youknowone:
ah, ok, maybe that's not expected to be failed. `.unwrap()` -> `.expect("types are always comparable")` will be enough

youknowone (approved):
Thank you!

PR #4048, opened 2022-08-12
Feature: sys.orig_argv

TwoPair:
I did like [this](https://github.com/RustPython/RustPython/pull/4048/files#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R392-R396) for `orig_argv`, but I'm not sure this implementation is right.
Plz, let me know if you have any feedback for me.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -389,7 +389,6 @@ fn create_settings(matches: &ArgMatches) -> Settings {
     });
 
     settings.argv = argv;
-    settings.orig_argv = env::args().collect();
 
     settings
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -389,6 +389,7 @@ fn create_settings(matches: &ArgMatches) -> Settings {
     });
 
     settings.argv = argv;
+    settings.orig_argv = env::args().collect();
 
     settings
 }
--- a/vm/src/stdlib/sys.rs
+++ b/vm/src/stdlib/sys.rs
@@ -218,6 +218,16 @@ mod sys {
     }
 
     #[pyattr]
+    fn orig_argv(vm: &VirtualMachine) -> Vec<PyObjectRef> {
+        vm.state
+            .settings
+            .orig_argv
+            .iter()
+            .map(|arg| vm.ctx.new_str(arg.clone()).into())
+            .collect()
+    }
+
+    #[pyattr]
     fn path(vm: &VirtualMachine) -> Vec<PyObjectRef> {
         vm.state
             .settings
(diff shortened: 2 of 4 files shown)


src/lib.rs
```diff
@@ -389,6 +389,11 @@ fn create_settings(matches: &ArgMatches) -> Settings {
     });
 
     settings.argv = argv;
+    settings.orig_argv = vec![env::current_exe()
+        .unwrap_or(PathBuf::new())
+        .to_str()
+        .unwrap()
+        .to_string()];
```
youknowone, line 396: The code looks putting only executable names not the full argv.
The value also need to be fetched from actual argv rather than 
directly processing executable name.

```python
import sys

print(sys.orig_argv)
```

```
$ python f.py -a -b -c
['~/.pyenv/versions/RustPython/bin/python', 'f.py', '-a', '-b', '-c']
```

TwoPair: I fixed it to get the correct result, but still don't know how to get the executable name indirectly...

src/lib.rs
```diff
@@ -388,7 +388,13 @@ fn create_settings(matches: &ArgMatches) -> Settings {
...
     });
 
-    settings.argv = argv;
+    settings.argv = argv.clone();
+    settings.orig_argv = vec![env::current_exe()
+        .unwrap_or_default()
+        .to_str()
+        .unwrap()
+        .to_string()];
+    settings.orig_argv.append(&mut argv);
```
youknowone, line 397: This is post-processed by clap. That doesn't look like 'orig' enough. Using [std::env::args](https://doc.rust-lang.org/std/env/fn.args.html) will fit better.

youknowone (changes requested):

vm/src/stdlib/sys.rs
```diff
@@ -217,6 +217,16 @@ mod sys {
         Vec::new()
     }
 
+    #[pyattr]
+    fn orig_argv(vm: &VirtualMachine) -> Vec<PyObjectRef> {
+        vm.state
+            .settings
+            .orig_argv
```
youknowone, line 224: does `orig_argv` only used for this function? Then I think settings.orig_argv is not necessary.

```suggestion
        std::env::args().
```

TwoPair: I agree with you. I checked the python doc and the `orig_argv` is not called anywhere else.
I'll remake this then.

youknowone (approved):
looks good, thank you!

PR #4050, opened 2022-08-12
Partially fix `global` behavior on inner class

### LINKED ISSUE #3993: Using `global` on inner class has different outcome
The behavior for `global Subclass` is not the same in this situation.

CPython:
image

RustPython:
image

jeffwang0516:
Trying to fix #3993 
---
Using global on inner Subclass now have the same behavior.
image

but still not correct for the Nested class
image

Seems that all nested classes need to be handled when applying `global`, need advise on how to fix this in a better way!
Thanks!

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -559,6 +559,16 @@ impl ExecutingFrame<'_> {
             }
             bytecode::Instruction::StoreGlobal(idx) => {
                 let value = self.pop_value();
+                if value.class().is(vm.ctx.types.type_type) {
+                    let qualified_name = value
+                        .get_attr(identifier!(vm, __qualname__), vm)?
+                        .downcast::<PyStr>()
+                        .expect("qualified name to be a string");
+                    let name = qualified_name.as_str().split('.').next_back().unwrap();
+                    value.set_attr(identifier!(vm, __qualname__), vm.new_pyobj(name), vm)?;
+
+                    // TODO: qualname for type_type attrs needs to be set as well
+                }
                 self.globals
                     .set_item(self.code.names[*idx as usize], value, vm)?;
                 Ok(None)


vm/src/frame.rs
```diff
@@ -559,6 +559,16 @@ impl ExecutingFrame<'_> {
...
+                if value.class().is(vm.ctx.types.type_type) {
+                    let qualified_name = value
+                        .get_attr(identifier!(vm, __qualname__), vm)?
+                        .downcast::<PyStr>()
+                        .expect("qualified name to be a string");
+                    let name = qualified_name.as_str().split('.').next_back().unwrap();
+                    value.set_attr(identifier!(vm, __qualname__), vm.new_pyobj(name), vm)?;
+
+                    // TODO: qualname for type_type attrs needs to be set as well
+                }
```
youknowone, line 571: If this is an experiment, it is ok. But we cannot actually add this code here because checking type for every StoreGlobal will be very expensive comparing to its own job.

youknowone (commented):
Is this rebased on #3998 which fixed `__qualname__` behavior?

youknowone:
To share text result, you can use 3 back-quote(`) like this:

\`\`\`
some code here
\`\`\`

```
some code here
```

I can't use text search for the image.
What's the expected result? and what's the actual result?

jeffwang0516:
> Is this rebased on #3998 which fixed `__qualname__` behavior?

Yes. 


> I can't use text search for the image. What's the expected result? and what's the actual result?

Take the following snippet as example:
```python
def test_c_methods():
    global Subclass
    class Subclass(tuple):
        class Nested(str):
            pass

test_c_methods()
```

The expected result is:

```
>>> Subclass.__qualname__
'Subclass'
>>>
>>> Subclass.Nested.__qualname__
'Subclass.Nested'
```

The actual result before applying my experiment:

```
>>>>> Subclass.__qualname__
'test_c_methods.<locals>.Subclass'
>>>>>
>>>>> Subclass.Nested.__qualname__
'test_c_methods.<locals>.Subclass.Nested'
```

The result after applying my experiment:
```
>>>>> Subclass.__qualname__
'Subclass'
>>>>>
>>>>> Subclass.Nested.__qualname__
'test_c_methods.<locals>.Subclass.Nested'
```

jeffwang0516:
Thanks for your advice!
I'll close this pr first and reopen a new one.

PR #4051, opened 2022-08-12
Update CI toolchain to stable

moreal:
Currently, the `cargo clippy` or the CI `Check Rust code with rustfmt and clippy` job failed.
It warns `clippy::derive_partial_eq_without_eq` as unknown lints but it exists in 1.63.0 release. You can see it through https://rust-lang.github.io/rust-clippy/master/.

At the #4046, it updated the toolchain's version to `1.62.1` from `stable` and the `stable` version was `1.63.0`.

But I see the pull request description while writing this pull request's description:

> I need time to check if the failure is our bug or rust regression.

I don't know its context so it may be intended and if then, this pull request should be closed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -263,7 +263,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: dtolnay/rust-toolchain@1.62.1
+      - uses: dtolnay/rust-toolchain@stable
         with:
             components: rustfmt, clippy
       - name: run rustfmt


youknowone:
hmm, why is this not fixing the error?

moreal:
> hmm, why is this not fixing the error?

Hmm.. I thought it is because the GitHub Actions follows the main branch but it seems not the situtation I thought...
I'll look more. 🙏🏻

.github/workflows/ci.yaml
```diff
@@ -263,11 +263,13 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: dtolnay/rust-toolchain@1.62.1
+      - uses: dtolnay/rust-toolchain@1.63.0
```
youknowone, line 266:
```suggestion
      - uses: dtolnay/rust-toolchain@stable
```
using stable is our common way

moreal:
When I saw the debug log, the `clippy` command's version was not 1.63.0 🤔 I think it's a bug of `dtolnay/rust-toolchain` or `rustup` 🤔 .
image

I'll push with the commit to use `stable` version.

.github/workflows/ci.yaml
moreal: I applied the suggestion by amending [d60baae](https://github.com/RustPython/RustPython/pull/4051/commits/d60baaeab5a1aa0b44e8f2b9741a441a42a6f35c) commit.

youknowone (approved):
Thank you! the difference between 1.63.0 and stable are confusing

PR #4052, opened 2022-08-12
Raise `ValueError` if `\x00` character exists for `eval` argument

moreal:
It isn't a full perfect solution for `eval` implementation but it fixes:

- `eval` became to receive `bytes` also.
- `eval` raises `ValueError` when there is `null` (`\x00`) character before compiling.

You can see also:

- [`_Py_SourceAsString`](https://github.com/python/cpython/blob/b5e3ea286289fcad12be78480daf3756e350f69f/Python/pythonrun.c#L1820-L1867)
- https://github.com/python/cpython/blob/b5e3ea286289fcad12be78480daf3756e350f69f/Python/bltinmodule.c#L961
```
    str = _Py_SourceAsString(source, "eval", "string, bytes or code", &cf, &source_copy);
```

p.s. The message says like it should receive `string` or `bytes` or `code` but it also receives `bytearray` well. 🤔

### DIFF
--- a/extra_tests/snippets/syntax_non_utf8.py
+++ b/extra_tests/snippets/syntax_non_utf8.py
@@ -5,7 +5,6 @@ from testutils import assert_raises
 
 dir_path = os.path.dirname(os.path.realpath(__file__))
 
-error = ValueError
-with assert_raises(error):
+with assert_raises(ValueError):
     with open(os.path.join(dir_path , "non_utf8.txt")) as f:
         eval(f.read())
--- a/extra_tests/snippets/syntax_non_utf8.py
+++ b/extra_tests/snippets/syntax_non_utf8.py
@@ -5,8 +5,7 @@ from testutils import assert_raises
 
 dir_path = os.path.dirname(os.path.realpath(__file__))
 
-# TODO: RUSTPYTHON, RustPython raises a SyntaxError here, but cpython raise a ValueError
-error = SyntaxError if platform.python_implementation() == 'RustPython' else ValueError
+error = ValueError
 with assert_raises(error):
     with open(os.path.join(dir_path , "non_utf8.txt")) as f:
         eval(f.read())
--- a/vm/src/stdlib/builtins.rs
+++ b/vm/src/stdlib/builtins.rs
@@ -24,8 +24,8 @@ mod builtins {
         format::call_object_format,
         function::Either,
         function::{
-            ArgBytesLike, ArgCallable, ArgIntoBool, ArgIterable, ArgMapping, FuncArgs, KwArgs,
-            OptionalArg, OptionalOption, PosArgs, PyArithmeticValue,
+            ArgBytesLike, ArgCallable, ArgIntoBool, ArgIterable, ArgMapping, ArgStrOrBytesLike,
+            FuncArgs, KwArgs, OptionalArg, OptionalOption, PosArgs, PyArithmeticValue,
         },
         protocol::{PyIter, PyIterReturn},
         py_io,
@@ -248,20 +248,14 @@ mod builtins {
     #[cfg(feature = "rustpython-compiler")]
     #[pyfunction]
     fn eval(
-        source: Either<
-            Either<PyStrRef, crate::builtins::PyBytesRef>,
-            PyRef<crate::builtins::PyCode>,
-        >,
+        source: Either<ArgStrOrBytesLike, PyRef<crate::builtins::PyCode>>,
         scope: ScopeArgs,
         vm: &VirtualMachine,
     ) -> PyResult {
         // source as string
         let code = match source {
             Either::A(either) => {
-                let source: &[u8] = match either {
-                    Either::A(ref a) => a.as_str().as_bytes(),
-                    Either::B(ref b) => b.as_bytes(),
-                };
+                let source: &[u8] = &either.borrow_bytes();
                 for x in source {
                     if *x == 0 {
                         return Err(vm.new_value_error(
--- a/vm/src/stdlib/builtins.rs
+++ b/vm/src/stdlib/builtins.rs
@@ -262,17 +262,16 @@ mod builtins {
                     ));
                 }
 
-                match std::str::from_utf8(source) {
-                    Ok(s) => Ok(Either::A(vm.ctx.new_str(s))),
-                    Err(err) => {
-                        let msg = format!(
-                            "(unicode error) 'utf-8' codec can't decode byte 0x{:x?} in position {}: invalid start byte",
-                            source[err.valid_up_to()],
-                            err.valid_up_to()
-                        );
-                        Err(vm.new_exception_msg(vm.ctx.exceptions.syntax_error.to_owned(), msg))
-                    }
-                }
+                let source = std::str::from_utf8(source).map_err(|err| {
+                    let msg = format!(
+                        "(unicode error) 'utf-8' codec can't decode byte 0x{:x?} in position {}: invalid start byte",
+                        source[err.valid_up_to()],
+                        err.valid_up_to()
+                    );
+
+                    vm.new_exception_msg(vm.ctx.exceptions.syntax_error.to_owned(), msg)
+                })?;
+                Ok(Either::A(vm.ctx.new_str(source)))
             }
             Either::B(code) => Ok(Either::B(code)),
         }?;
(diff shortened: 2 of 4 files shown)


extra_tests/snippets/syntax_non_utf8.py
```diff
@@ -5,8 +5,7 @@
 
 dir_path = os.path.dirname(os.path.realpath(__file__))
 
-# TODO: RUSTPYTHON, RustPython raises a SyntaxError here, but cpython raise a ValueError
-error = SyntaxError if platform.python_implementation() == 'RustPython' else ValueError
+error = ValueError
 with assert_raises(error):
```
fanninpm, line 9:
```suggestion
with assert_raises(ValueError):
```

moreal:
About the `CI / Check Rust code with rustfmt and clippy (pull_request)` is broken, you can see #4051

extra_tests/snippets/syntax_non_utf8.py
moreal: I applied your suggestion in [7e88c8f](https://github.com/RustPython/RustPython/pull/4052/commits/7e88c8ff55017f0c11b807d14e97398a650aaad5). Thank you! 🙏🏻

vm/src/stdlib/builtins.rs
```diff
@@ -248,11 +248,35 @@ mod builtins {
     #[cfg(feature = "rustpython-compiler")]
     #[pyfunction]
     fn eval(
-        source: Either<PyStrRef, PyRef<crate::builtins::PyCode>>,
+        source: Either<
+            Either<PyStrRef, crate::builtins::PyBytesRef>,
+            PyRef<crate::builtins::PyCode>,
+        >,
```
youknowone, line 254: is this exactly `str` or `bytes` or `ArgStrOrBytesLike`?
`ArgStrOrBytesLike` accepts more types like `bytearray`

vm/src/stdlib/builtins.rs
```diff
@@ -248,11 +248,35 @@ mod builtins {
...
+                for x in source {
+                    if *x == 0 {
+                        return Err(vm.new_value_error(
+                            "source code string cannot contain null bytes".to_owned(),
+                        ));
+                    }
+                }
+
+                Ok(Either::A(
+                    vm.ctx.new_str(std::str::from_utf8(source).unwrap()),
```
youknowone, line 274: what happens if `bytes` include non-ascii bytes?
what happens if `bytes` include invalid utf8 character?

vm/src/stdlib/builtins.rs
moreal: When I run rustpython with `ArgStrOrBytesLike`, it seems very suit for this case. `string, bytes or code` is just error message and the `eval` also receives `bytearray` `Buffer` 🤔 . So I'll push a commit apply `ArgStrOrBytesLike`. Thanks! 🙏🏻

vm/src/stdlib/builtins.rs
moreal: > what happens if bytes include invalid utf8 character?

It causes `panic`. I tested with the below python code:

```python
eval(b'\xff')
```

And it prints:


```text
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Utf8Error { valid_up_to: 0, error_len: Some(1) }', vm/src/stdlib/builtins.rs:268:64
```

Because the `C` language `const char *` type for string and also byte array, it can be represented as a single type with a compiler option. So I'll see more the compiler, parser section. If you have useful documentations to recommend, please leave them as comments 🙇🏻‍♂️ 

```c
// _PyPegen_run_parser_from_string
    if (flags != NULL && flags->cf_flags & PyCF_IGNORE_COOKIE) {
        tok = PyTokenizer_FromUTF8(str, exec_input);
    } else {
        tok = PyTokenizer_FromString(str, exec_input);
    }
```

youknowone: What I thought was `.unwrap()` need to be avoided because CPython would not panic but raise  a `SyntaxError`.
The major difference between FromUTF8 and FromString is using decoder or not. The decode error seems to be wrapped by SyntaxError somewhere.

moreal: Ah, I agree with your thought. 🙏🏻  I'll work for wrapping the decode error.

moreal: When you execute `eval(b'\xff')` in CPython implementation, you will see the below output:

```
>>> eval(b"\xff")
  File "<string>", line 1
    �
    ^
SyntaxError: (unicode error) 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```

I set `SyntaxError: (unicode error) 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte` message as format and I tried to apply it in [714ce4d](https://github.com/RustPython/RustPython/pull/4052/commits/714ce4d07d7c900e7ceb2dab9eae725f15f749e4).

vm/src/stdlib/builtins.rs
```diff
@@ -248,11 +248,37 @@ mod builtins {
...
+        let code = match source {
+            Either::A(either) => {
+                let source: &[u8] = &either.borrow_bytes();
+                for x in source {
+                    if *x == 0 {
+                        return Err(vm.new_value_error(
+                            "source code string cannot contain null bytes".to_owned(),
+                        ));
+                    }
+                }
```
youknowone, line 265:
```suggestion
                if source.iter().any(|&b| b == 0) {
                    return Err(vm.new_value_error(
                        "source code string cannot contain null bytes".to_owned(),
                    ));
                }
```

vm/src/stdlib/builtins.rs
```diff
@@ -248,11 +248,37 @@ mod builtins {
...
+                match std::str::from_utf8(source) {
+                    Ok(s) => Ok(Either::A(vm.ctx.new_str(s))),
+                    Err(err) => {
+                        let msg = format!(
+                            "(unicode error) 'utf-8' codec can't decode byte 0x{:x?} in position {}: invalid start byte",
+                            source[err.valid_up_to()],
+                            err.valid_up_to()
+                        );
+                        Err(vm.new_exception_msg(vm.ctx.exceptions.syntax_error.to_owned(), msg))
+                    }
```
youknowone, line 276: I prefer to handle error first not to disturb to read main control flow
```suggestion
                let source = std::str::from_utf8(source).map_err(|err| {
                    let msg = format!(
                            "(unicode error) 'utf-8' codec can't decode byte 0x{:x?} in position {}: invalid start byte",
                            source[err.valid_up_to()],
                            err.valid_up_to()
                        );
                        Err(vm.new_exception_msg(vm.ctx.exceptions.syntax_error.to_owned(), msg))
                })?;
                Either::A(vm.ctx.new_str(s))),

```

youknowone (commented):
looks great! I left a few minor change requests

vm/src/stdlib/builtins.rs
moreal: https://doc.rust-lang.org/nightly/core/iter/trait.Iterator.html#method.any

> any() is short-circuiting; in other words, it will stop processing as soon as it finds a true, given that no matter what else happens, the result will also be true.

It's simple but powerful 🚀

vm/src/stdlib/builtins.rs
```diff
@@ -248,11 +248,35 @@ mod builtins {
...
+        source: Either<ArgStrOrBytesLike, PyRef<crate::builtins::PyCode>>,
         scope: ScopeArgs,
         vm: &VirtualMachine,
     ) -> PyResult {
-        run_code(vm, source, scope, compile::Mode::Eval, "eval")
+        // source as string
+        let code = match source {
+            Either::A(either) => {
+                let source: &[u8] = &either.borrow_bytes();
+                if source.iter().any(|&b| b == 0) {
```
youknowone, line 259:
```suggestion
                if source.contains(&0) {
```
I am sorry, I was forgetting about this  but slice has contains method

vm/src/stdlib/builtins.rs
moreal: I applied it in [c672d8f](https://github.com/RustPython/RustPython/pull/4052/commits/c672d8f8b2217d41b680eeeb7dbe38b4f9a1119c).

vm/src/stdlib/builtins.rs
moreal: I applied it in [879ed6e](https://github.com/RustPython/RustPython/pull/4052/commits/879ed6e005cb98188051b13148d9599f26974f69)

youknowone (approved):
Thank you!

Issue #4000, opened 2022-08-05
ssl.SSLError.reason


youknowone:
## Feature
ssl.SSLError.reason

## Python Documentation
https://docs.python.org/3/library/ssl.html#ssl.SSLError.reason

tgsong827:
can you assign this to me?

tgsong827:
Is this issue related with the sslError format made by `fn convert_openssl_error()` ?

I found that RustPython's sslError format is different with CPython's.

- RustPython
```
ssl.SSLError: [Errno 136] [x509 certificate routines] no certificate or crl found (ssl.rs:691)
```

- CPython
```
ssl.SSLError: [X509: NO_CERTIFICATE_OR_CRL_FOUND] no certificate or crl found (_ssl.c:4293)
```

CPython is printing lib name and reason name.
RustPython is printing errCode and lib str.

CPython is using `_ssl_data.h` to get the lib name and reason name.
and ``

so I wonder this issue is about it.

youknowone:
The issue is not about messages but absense of the attribute `reason`.
In CPython, `fill_and_set_sslerror` function adds `reason` attribute, while RustPython doesn't do in `convert_openssl_error`.

Maybe having same messages would be better, but we don't count them as incompatibility.

PR #4058, opened 2022-08-13
Add 'library', 'reason' Attribute to ssl.SSLError

tgsong827:
This PR associated with #4000.

I added `libaray` and `reason` attribute to ssl.SSLError like CPython.


youknowone (approved):
Thank you! I had a case triggering this problem, but couldn't find one.

PR #4081, opened 2022-08-16
Feat: locale-module

Yaminyam:


stdlib/src/locale.rs
```diff
@@ -0,0 +1,55 @@
...
+    fn error_type(vm: &VirtualMachine) -> PyTypeRef {
+        vm.ctx.new_exception_type(
+            "locale",
+            "error",
+            Some(vec![vm.ctx.exceptions.value_error.to_owned()]),
+        )
+    }
+
+    #[pyfunction]
+    fn setlocale(category: PyIntRef, locale: Option<PyStr>, vm: &VirtualMachine) -> PyResult<*mut i8> {
```
youknowone, line 33:
```suggestion
    fn setlocale(category: PyIntRef, locale: Option<PyStrRef>, vm: &VirtualMachine) -> PyResult<Vec<u8>> {
```

youknowone: no, they are not compatible.

`Vec<u8>` -> `PyBytes`
`String` -> `PyStr`

stdlib/src/locale.rs
```diff
@@ -0,0 +1,86 @@
...
+                        libc::strlen(result),
+                    )
+                })
+            }
+        }
+    }
+
+    #[pyfunction]
+    fn localeconv(vm: &VirtualMachine) -> PyResult<PyStrRef> {
+        let result = unsafe { libc::localeconv() };
```
fanninpm, line 80: In the current Windows version of the `libc` crate, this function and its return type are both missing.

fanninpm:
I don't think this particular module belongs in the `stdlib` crate:

```text
>>> import _locale
>>> _locale.__file__
AttributeError: module '_locale' has no attribute '__file__'. Did you mean: '__name__'?
```

I think it should belong in the `vm` crate instead. What do you think, @youknowone?

youknowone:
@fanninpm Interesting. Does `__file__` attribute existence different by it is builtin or not?

I usually check `sys.builtin_module_names`
```
>>> import sys
>>> sys.builtin_module_names
('_abc', '_ast', '_codecs', '_collections', '_functools', '_imp', '_io', '_locale', '_operator', '_signal', '_sre', '_stat', '_string', '_symtable', '_thread', '_tracemalloc', '_warnings', '_weakref', 'atexit', 'builtins', 'errno', 'faulthandler', 'gc', 'itertools', 'marshal', 'posix', 'pwd', 'sys', 'time', 'xxsubtype')
>>> '_locale' in sys.builtin_module_names
True
```
Yes, so this belongs in `vm` crate.

fanninpm:
> Does `__file__` attribute existence different by it is builtin or not?

Let's see:

```text
>>> import argparse # pure-python module
>>> argparse.__file__
'/usr/lib/python3.10/argparse.py'
>>> import mmap # dynamically-linked c module
>>> mmap.__file__
'/usr/lib/python3.10/lib-dynload/mmap.cpython-310-x86_64-linux-gnu.so'
>>> import gc # statically-linked c module
>>> gc.__file__
AttributeError: module 'gc' has no attribute '__file__'. Did you mean: '__name__'?
```

[The Python docs](https://docs.python.org/3/reference/datamodel.html#the-standard-type-hierarchy) say that the `__file__` attribute is
> The pathname of the file from which the module was loaded, if it was loaded from a file. The [`__file__`](https://docs.python.org/3/reference/import.html#file__) attribute may be missing for certain types of modules, such as C modules that are statically linked into the interpreter. For extension modules loaded dynamically from a shared library, it’s the pathname of the shared library file.

youknowone:
Thanks! I didn't know that.

youknowone:
Thank you for contributing. This is not implemented by #4558

Issue #3922, opened 2022-07-18
_thread.interrupt_main and PyErr_SetInterruptEx


youknowone:
## Feature

See #3921 

## Python Documentation

https://docs.python.org/ko/3/library/_thread.html#thread.interrupt_main

PR #4082, opened 2022-08-16
Feature: _thread.interrupt_main

TwoPair:
Additionally I ran the test below, and it was successful.

```py
# SuperFastPython.com
# example of main thread cannot be interrupting while sleeping
from time import sleep
from threading import Thread
from _thread import interrupt_main
import sys

# task executed in a new thread
def task():
    # block for a moment
    sleep(3)
    # interrupt the main thread
    print('Interrupting main thread now')
    interrupt_main()

# start the new thread
thread = Thread(target=task)
thread.start()
# handle being interrupted
try:
    # block for a long time
    print('Main thread waiting...')
    sleep(7)
except KeyboardInterrupt:
    # terminate main thread
    print('Main interrupted! Exiting.')
    sys.exit()
``` 

Close: #3922

### DIFF
--- a/vm/src/stdlib/signal.rs
+++ b/vm/src/stdlib/signal.rs
@@ -33,23 +33,23 @@ pub(crate) mod _signal {
     }
 
     #[cfg(unix)]
-    use nix::unistd::alarm as sig_alarm;
+    pub use nix::unistd::alarm as sig_alarm;
 
     #[cfg(not(windows))]
-    use libc::SIG_ERR;
+    pub use libc::SIG_ERR;
 
     #[cfg(not(windows))]
     #[pyattr]
-    use libc::{SIG_DFL, SIG_IGN};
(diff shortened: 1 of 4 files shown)


vm/src/stdlib/signal.rs
```diff
@@ -279,4 +279,21 @@ pub(crate) mod _signal {
...
+
+        match signum as usize {
+            SIG_DFL | SIG_IGN => Ok(()),
+            _ => {
+                // interrupt the main thread with given signal number
+                run_signal(signum);
+                Ok(())
+            }
+        }
+    }
```
youknowone, line 298: we put internal signal tools to `vm/src/signal.rs` and python functions to `vm/src/stdlib/signal.rs`. This is better to be pu in `vm.src/signal.rs`

youknowone (approved):
looks good, thank you!

PR #4087, opened 2022-08-17
Update `traceback.tb_next` to be writable.

### LINKED ISSUE #3857: traceback.tb_next is not writable
See `PyTraceback` type.

It is writable at lease since Python 3.10
(I didn't track it down if it is changed or not)

#3862 requires this feature

tgsong827:
Resolves #3857


youknowone (approved):
looks good, thank you!

PR #4090, opened 2022-08-17
generated parser cache out of repository

youknowone:


fanninpm:
It might make more sense to undo #3652.

youknowone:
@coolreader18 didn't want to add lalrpop dependency for main CI path.

I'd like to find a way to meet the requirement.

youknowone:
@fanninpm you know both the previous work and this change. you had a coverage concern. I want to know how you think about removing it again.

@coolreader18 I tried best to meet your requirements. please check if I missed something when you come back later.

fanninpm:
> you know both the previous work and this change. you had a coverage concern. I want to know how you think about removing it again.

I welcome this sort of change because I think it should ease the burden placed upon new/novice contributors. (See, for example, https://github.com/RustPython/RustPython/pull/3879#issuecomment-1188048140.) I'm not too concerned about test coverage for the code that LALRPOP generates, especially because I don't see a clear way to represent test coverage of a LALRPOP source file.

youknowone:
Thank you, lets try

coolreader18:
For this, wouldn't someone need to always either have `lalrpop` installed or pass the feature? Also, the goal was to reduce compile times in general, but maybe the situation overall has gotten better such that lalrpop is fine.

youknowone:
Oh, welcome back. Yes. it will require lalrpop for first build.
The lalrpop build is comparably small part of first build time. (by referring Amdahl's law, i don't think we need any special handling for lalrpop. the clean build requires many minutes while lalrpop takes only small portion of it)
On the other hands, it still reduce compile times in general because it is still using hash checker you made.

About the lalrpop running time, not the build time, it was already reasonable fast on my machines when #3652 made. So it is hard to be able to tell about it. I don't think the running time is enhanced much.

not-my-profile:
This change probably broke the docs.rs build for rustpython-parser, see #4436. But I am afraid I understand too little of the build system to know how we can resolve this.

Issue #4095, opened 2022-08-17
Implicit string concatenation produces different AST than CPython


charliermarsh:
The AST produced by RustPython differs from CPython when parsing implicitly concatenated strings.

For example, given this source:

```py
import ast

expr = """(
    'Hello, '
    'world!'
)"""

print(ast.dump(ast.parse(expr), include_attributes=True))
```

CPython produces a single `Constant`, with a value equal to the concatenated strings:

```
Expr(value=Constant(value='Hello, world!', kind=None, lineno=2, col_offset=4, end_lineno=3, end_col_offset=12), lineno=1, col_offset=0, end_lineno=4, end_col_offset=1)
```

Whereas RustPython produces a `JoinedStr` consisting of two `Constant` sub-expressions:

```
Expr { value: Located { location: Location { row: 2, column: 6 }, custom: (), node: JoinedStr { values: [Located { location: Location { row: 2, column: 6 }, custom: (), node: Constant { value: Str("Hello, "), kind: None } }, Located { location: Location { row: 3, column: 6 }, custom: (), node: Constant { value: Str("world!"), kind: None } }] } } }
```

There are some other differences too, e.g.:

- CPython seems to propagate the string kind of the first substring. So if the first substring is `u'Hello, '`, then the whole thing will be treated as Unicode. If the second substring is `u'world!'`, CPython drops the `u` entirely. (I found this surprising.)
- If any of the substrings are f-strings, CPython returns a `JoinedStr`, but exhibits the same constant-collapsing as in the provided example.

As far as I can tell, the rules are (roughly):

- If every substring is a `Constant`, return a single `Constant` with the substrings concatened.
- If any substring is an f-string, return a `JoinedStr`, then flatten out the values into that top-level `JoinedStr`, collapsing any adjacent `Constant` expressions, so that you're left with `Constant` expressions interspersed with `FormattedValue` expressions.
- If the first substring has a "kind", apply that "kind" to any `Constant` expressions.

(I haven't been able to find this in any documentation, only from testing out CPython 3.8.12 itself.)

Happy to attempt a PR, unless this behavior is intentional.

(By the way: very impressive project!)

charliermarsh:
I _think_ this can be fixed by modifying the `<location:@L> <s:(@L string)+> =>?` case in the LALRPOP.

charliermarsh:
Made an attempt with #4097.

PR #4097, opened 2022-08-18
Implement CPython-like implicit string concatenation in parser

charliermarsh:
# Summary

This PR attempts to address #4095 by modifying the parser's handling of implicit string concatenation to better match CPython's behavior. Specifically, rather than _always_ returning a `JoinedStr` in the presence of multiple substrings, we return a single `Constant` if all the substrings are constant, and a single `JoinedStr` if _any_ of the substrings are `JoinedStr`, but with the sub-expressions flattened as much as possible.


charliermarsh:
(High- and low-level feedback welcome!)

fanninpm:
@youknowone does this conflict with #4090?

parser/src/string.rs
```diff
@@ -0,0 +1,182 @@
...
+            None
+        };
+
+        // Preserve the initial location.
+        let location = values[0].location;
+
+        if values
+            .iter()
+            .all(|value| matches!(value.node, ExprKind::Constant { .. }))
+        {
```
youknowone, line 19: this condition seems to be handled by compiler instead of parser

parser/src/string.rs
```diff
@@ -0,0 +1,182 @@
...
+    #[test]
+    fn test_parse_f_string_concat_3() {
+        let source = String::from("'Hello ' f'world{\"!\"}'");
+        let parse_ast = parse_program(&source).unwrap();
+        insta::assert_debug_snapshot!(parse_ast);
+    }
+
+    #[test]
+    fn test_parse_u_f_string_concat_2() {
+        let source = String::from("u'Hello ' f'world'");
```
youknowone, line 178: another 3-parts test will be useful to check how f-string cut concatenation

```
u'Hello ' f'world' '!'
```

youknowone (commented):
Nice work! The approach seems convincing. I left a few comments about high-level decisions.

@fanninpm conflicts happen everyday and we always do rebase. It will not be a big deal.

parser/src/string.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::error::{LexicalError, LexicalErrorType};
+
+pub fn parse_implicit_concatenation(values: Vec<Expr>) -> Result<Expr, LexicalError> {
+    Ok(if values.len() > 1 {
+        // As in CPython, use the kind of the first Expression.
+        let kind = if let ExprKind::Constant { kind, .. } = &values[0].node {
+            kind.clone()
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
charliermarsh: Are you saying that the _behavior_ is different from what I've implemented, or my description of the behavior is inaccurate?

parser/src/string.rs
```diff
@@ -0,0 +1,182 @@
...
+        values.into_iter().next().unwrap()
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::parser::parse_program;
+
+    #[test]
+    fn test_parse_string_concat() {
```
charliermarsh, line 135: These arguably belong in `parser.rs` since they depend on the parser (which depends on this file, rather than the other way around).

parser/src/string.rs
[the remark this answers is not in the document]
youknowone: Oh, I am sorry, I thought the kind is pointing string or f-string. But it was str or bytes.

youknowone:
@fanninpm oh, if you wanted to asked to check if #4042 has string changes to conflict, the answer is still no. only `!` conversions changed in the update.

parser/src/string.rs
```diff
@@ -0,0 +1,189 @@
...
+            )
+        } else {
+            // Otherwise, we have at least one JoinedStr, so return a single JointedStr with as few
+            // Constants as possible by merging adjacent Constants.
+            let values: Vec<Expr> = values
+                .into_iter()
+                .map(|value| match value.node {
+                    ExprKind::JoinedStr { values } => Ok(values),
+                    ExprKind::Constant { .. } => Ok(vec![value]),
+                    _ => Err(LexicalError {
```
charliermarsh, line 58: Considering structuring this code such that I iterate once upfront to ensure that I only have these expression types (maybe even cast them to a custom Enum that's distinct from `ExprKind`?), so that I don't have to handle these mismatch errors multiple times.

parser/src/string.rs
```diff
@@ -0,0 +1,189 @@
+use crate::ast::{Constant, Expr, ExprKind};
+use crate::error::{LexicalError, LexicalErrorType};
+
+pub fn parse_implicit_concatenation(values: Vec<Expr>) -> Result<Expr, LexicalError> {
```
youknowone, line 4: The new signature will look like:
```suggestion
pub fn parse_implicit_concatenation(values: Vec<(Location, (String, StringKind))>) -> Result<Expr, LexicalError> {
```

parser/src/string.rs
```diff
@@ -0,0 +1,189 @@
...
+        } else {
+            None
+        };
+
+        // Preserve the initial location.
+        let location = values[0].location;
+
+        if values
+            .iter()
+            .all(|value| matches!(value.node, ExprKind::Constant { .. }))
```
youknowone, line 18:
```suggestion
        let has_fstring = values
            .iter()
            .any(|value| !matches!(value.node, ExprKind::Constant { .. }));
```

parser/src/string.rs
```diff
@@ -0,0 +1,189 @@
...
+                    location,
+                    ExprKind::Constant {
+                        value: Constant::Str(current.join("")),
+                        kind,
+                    },
+                ));
+                current.clear();
+            }
+
+            Expr::new(location, ExprKind::JoinedStr { values: deduped })
```
youknowone, line 123: The new JoinedStr creation will look like:
```suggestion
            if has_fstring {
                Expr::new(location, ExprKind::JoinedStr { values: deduped })
            } else {
                deduped.into_iter().exactly_one().expect("string must be concatenated to single element")
            }
```

youknowone (changes requested):
Thank you! I learned more from your comments.

Here is 2 choices of suggestions. Minimal fix or more refactoring.

Because I think the [string-only branch](https://github.com/RustPython/RustPython/pull/4097/files#diff-23fa8df3292fbdfa484a18bd5316eb8f276854829034ba0f4e78fa63188fe4d8R16-R18) is duplicated logic, I want to remove it.

Suggestion 1. If you want to finish this PR quickly, please remove the string-only branch and use `has_fstring` flag I suggested to decide to create JoinedStr. That will be good enough. And we can refactor it later.
Suggestion 2. If you want to fully refactor the string processing, please move the whole string expr handling to the new function and use the StringKind instead of ast::ExprKind matches.

parser/src/string.rs
youknowone: I suggested a way to use StringKind instead of Expr. That will be helpful if you want to go.

charliermarsh:
Awesome, thanks tons @youknowone, I'm learning a lot from your comments. Will do another pass on this and figure out which of the two suggestions I want to pursue.

charliermarsh:
@youknowone - I went for Suggestion 2, attempted to refactor to use `StringKind`, etc.

parser/src/string.rs
charliermarsh: I think I still need this `has_fstring` flag and logic, even after the refactor, since I need this at the end:

```rust
if has_fstring {
    Expr::new(initial_location, ExprKind::JoinedStr { values: deduped })
} else {
    deduped
        .into_iter()
        .exactly_one()
        .expect("String must be concatenated to a single element.")
}
```

...but correct me if I'm wrong.

parser/Cargo.toml
```diff
@@ -23,6 +23,7 @@ unic-ucd-ident  = "0.9.0"
 unicode_names2 = "0.5.0"
 phf = "0.10.1"
 ahash = "0.7.6"
+itertools = "0.10.3"
```
charliermarsh, line 26: (Had to add this to the `parser`, assume that's ok? Also, unclear if you want to keep these sorted.)

parser/src/string.rs
charliermarsh: (In theory, if `values` at the end is just a single `Expr`, and it's a constant, we _could_ just return that -- but CPython seems to return a `JoinedStr` with one constant child, so we'd be deviating from their behavior IIUC.)

parser/src/string.rs
```diff
@@ -0,0 +1,170 @@
...
+        }
+
+        if has_fstring {
+            Expr::new(initial_location, ExprKind::JoinedStr { values: deduped })
+        } else {
+            deduped
+                .into_iter()
+                .exactly_one()
+                .expect("String must be concatenated to a single element.")
+        }
```
youknowone, line 107: To reduce indent depth, Moving `Ok()` to here will be better

parser/src/string.rs
```diff
@@ -0,0 +1,170 @@
...
+                Ok(values) => values.into_iter().map(Ok).collect(),
+                Err(e) => vec![Err(e)],
+            })
+            .collect::<Result<Vec<Expr>, LexicalError>>()?;
+
+        // De-duplicate adjacent constants.
+        let mut deduped: Vec<Expr> = vec![];
+        let mut current: Vec<String> = vec![];
+        for value in values {
+            match value.node {
```
youknowone, line 57: match here is actually matching `StringKind::F` or not.
The entire upper match can be moved here.

youknowone (commented):
would you also rebase it to our main branch? parser is recently updated to 3.10 compliant version.

parser/src/string.rs
youknowone: Oh, yep. Sorry for poor explanation. we need has_fstring, but not by iteration for its own. we can manage a mutable variable to update when it hits fstring branch later.

parser/Cargo.toml
youknowone: sorting will be good. this is also ok.

parser/src/string.rs
charliermarsh: I was able to consolidate but I believe I still need an inner match?

An example would be something like:

```
(
  "foo"
  f"bar{1}baz"
)
```

In this case, we want to produce `JoinedStr([Constant("foobar"), FormattedValue(1), Constant("baz")])`. So I have to iterate over the values of f-string within the `StringKind` match, IIUC.

parser/Cargo.toml
```diff
@@ -9,22 +9,23 @@ license = "MIT"
...
 log = "0.4.16"
 num-bigint = "0.4.3"
 num-traits = "0.2.14"
+phf = "0.10.1"
+rustpython-ast = { path = "../ast" }
 unic-emoji-char = "0.9.0"
 unic-ucd-ident  = "0.9.0"
 unicode_names2 = "0.5.0"
-phf = "0.10.1"
-ahash = "0.7.6"
```
charliermarsh, line 27: (Sort, and add `itertools`.)

parser/src/string.rs
```diff
@@ -0,0 +1,155 @@
...
+
+    // De-duplicate adjacent constants.
+    let mut deduped: Vec<Expr> = vec![];
+    let mut current: Vec<String> = vec![];
+    for (location, (string, string_kind)) in values {
+        match string_kind {
+            StringKind::Normal => current.push(string),
+            StringKind::U => current.push(string),
+            StringKind::F => {
+                if let ExprKind::JoinedStr { values } = parse_located_fstring(&string, location)
```
charliermarsh, line 29: Arguably `parse_located_fstring` could now return `Vec<Expr>`?

charliermarsh:
(Rebased!)

parser/src/string.rs
youknowone: oh, that makes sense! we never use the returned JoinedStr

parser/src/string.rs
```diff
@@ -0,0 +1,155 @@
+use crate::ast::{Constant, Expr, ExprKind, Location};
+use crate::error::LexicalErrorType::FStringError;
+use crate::error::{LexicalError, LexicalErrorType};
+use crate::fstring::parse_located_fstring;
+use crate::token::StringKind;
+use itertools::Itertools;
+
+pub fn parse_implicit_concatenation(
```
youknowone, line 8: The function is not only handling concatenation but strings itself now
```suggestion
pub fn parse_strings(
```

parser/src/string.rs
charliermarsh: I can do this in a separate PR.

parser/src/string.rs
```diff
@@ -1,86 +1,74 @@
...
+    error::{LexicalError, LexicalErrorType},
+    fstring::parse_located_fstring,
+    token::StringKind,
+};
 use itertools::Itertools;
 
-pub fn parse_strings(
-    values: Vec<(Location, (String, StringKind))>,
-) -> Result<Expr, LexicalError> {
+pub fn parse_strings(values: Vec<(Location, (String, StringKind))>) -> Result<Expr, LexicalError> {
```
charliermarsh, line 9: (Gah, thank you, forgot to `cargo fmt`.)

charliermarsh:
Good changes, thanks @youknowone!

youknowone (approved):
Thank you! I made a few minor changes.

PR #4106, opened 2022-08-19
Implement `__slots__` with `member_descriptor`

moreal:
This pull request tries to implement `__slots__` feature with `member_descriptor`.

This pull request doesn't contain `mebmer_descriptor.slots.descr_get` yet and it'll be implemented in the next pull request because I worried that this pull request becomes too big to review.

### DIFF
--- a/vm/src/builtins/descriptor.rs
+++ b/vm/src/builtins/descriptor.rs
@@ -1,6 +1,7 @@
 use rustpython_common::lock::PyRwLock;
 
-use crate::{Context, Py, PyRef, PyResult, VirtualMachine};
+use crate::types::GetDescriptor;
+use crate::{Context, Py, PyObjectRef, PyRef, PyResult, VirtualMachine};
 
 use super::{PyStr, PyType, PyTypeRef};
 use crate::class::PyClassImpl;
@@ -18,11 +19,21 @@ pub enum MemberKind {
     ObjectEx = 16,
 }
 
-#[derive(Debug)]
 pub struct MemberDef {
     pub name: String,
     pub kind: MemberKind,
+    pub getter: fn(PyObjectRef, &VirtualMachine) -> PyResult,
     pub doc: Option<String>,
+}
+
+impl std::fmt::Debug for MemberDef {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("MemberDef")
+            .field("name", &self.name)
+            .field("kind", &self.kind)
+            .field("doc", &self.doc)
+            .finish()
+    }
 }
 
 #[pyclass(name = "member_descriptor", module = false)]
@@ -51,7 +62,7 @@ fn calculate_qualname(descr: &DescrObject, vm: &VirtualMachine) -> PyResult<Opti
     }
 }
 
-#[pyclass(flags(BASETYPE))]
+#[pyclass(with(GetDescriptor), flags(BASETYPE))]
 impl MemberDescrObject {
     #[pymethod(magic)]
     fn repr(zelf: PyRef<Self>) -> String {
@@ -74,6 +85,23 @@ impl MemberDescrObject {
         }
 
         Ok(self.common.qualname.read().to_owned())
+    }
+}
+
+impl GetDescriptor for MemberDescrObject {
+    fn descr_get(
+        zelf: PyObjectRef,
+        obj: Option<PyObjectRef>,
+        _cls: Option<PyObjectRef>,
+        vm: &VirtualMachine,
+    ) -> PyResult {
+        Ok(match obj {
+            Some(x) => {
+                let zelf = Self::_zelf(zelf, vm)?;
+                (zelf.member.getter)(x, vm)?
+            }
+            None => zelf,
+        })
     }
 }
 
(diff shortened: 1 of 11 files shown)


vm/src/builtins/descriptor.rs
```diff
@@ -0,0 +1,134 @@
+use rustpython_common::lock::PyRwLock;
+
```
youknowone, line 2:
```suggestion
```

vm/src/builtins/descriptor.rs
```diff
@@ -0,0 +1,134 @@
...
+use crate::function::Either;
+use crate::types::{Constructor, GetDescriptor, Unconstructible};
+use crate::{AsObject, Context, Py, PyObjectRef, PyPayload, PyRef, PyResult, VirtualMachine};
+
+#[derive(Debug)]
+pub struct DescrObject {
+    pub typ: PyTypeRef,
+    pub name: String,
+    pub qualname: PyRwLock<Option<String>>,
+}
```
youknowone, line 14: 👍

vm/src/builtins/descriptor.rs
```diff
@@ -0,0 +1,134 @@
...
+
+#[derive(Debug)]
+pub enum MemberKind {
+    ObjectEx = 16,
+}
+
+pub struct MemberDef {
+    pub name: String,
+    pub kind: MemberKind,
+    pub getter_or_offset: Either<fn(PyObjectRef, &VirtualMachine) -> PyResult, usize>,
```
youknowone, line 24: Using a dedicated enum type will be better to adding utility functions.

```rust
enum MemberGetter {
    Getter(fn(&VirtualMachine, PyObjectRef) -> PyResult),
    Offset(usize),
}
```
Swapping parameter orders in Getter will be helpful to have same convention with `PyGetterFnuc`.

vm/src/builtins/descriptor.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+impl PyPayload for MemberDescrObject {
+    fn class(vm: &VirtualMachine) -> &'static Py<PyType> {
+        vm.ctx.types.member_descriptor_type
+    }
+}
+
+fn calculate_qualname(descr: &DescrObject, vm: &VirtualMachine) -> PyResult<Option<String>> {
+    let type_qualname = vm.get_attribute_opt(descr.typ.to_owned().into(), "__qualname__")?;
```
youknowone, line 52:
```suggestion
    let type_qualname = if let Some(qualname) = vm.get_attribute_opt(descr.typ.to_owned().into(), "__qualname__")? {
        qualname
    } else {
        return Ok(None);
    };
```

vm/src/builtins/descriptor.rs
```diff
@@ -0,0 +1,134 @@
...
+    match type_qualname {
+        None => Ok(None),
+        Some(obj) => match obj.downcast::<PyStr>() {
+            Ok(str) => Ok(Some(format!("{}.{}", str, descr.name))),
+            Err(_) => Err(vm.new_type_error(
+                "<descriptor>.__objclass__.__qualname__ is not a unicode object".to_owned(),
+            )),
+        },
+    }
```
youknowone, line 61:
```suggestion
    let str = obj.downcast::<PyStr>().map_err(|_| vm.new_type_error(
                "<descriptor>.__objclass__.__qualname__ is not a unicode object".to_owned(),
    ))?;
    Ok(Some(format!("{}.{}", str, descr.name))))
```

vm/src/builtins/descriptor.rs
```diff
@@ -0,0 +1,134 @@
...
+        zelf.member.doc.to_owned()
+    }
+
+    #[pyproperty(magic)]
+    fn qualname(&self, vm: &VirtualMachine) -> PyResult<Option<String>> {
+        if self.common.qualname.read().is_none() {
+            *self.common.qualname.write() = calculate_qualname(&self.common, vm)?;
+        }
+
+        Ok(self.common.qualname.read().to_owned())
```
youknowone, line 86: reduce lock once when it is some
```suggestion
        let qualname = self.common.qualname.read();
        Ok(if qualname.is_none() {
            drop(qualname);
            let calculated = calculate_qualname(&self.common, vm)?;
            *self.common.qualname.write() = calculated;
            calculated
        } else {
            qualname.to_owned()
        })
```

vm/src/builtins/descriptor.rs
```diff
@@ -0,0 +1,134 @@
...
+        _cls: Option<PyObjectRef>,
+        vm: &VirtualMachine,
+    ) -> PyResult {
+        match obj {
+            Some(x) => {
+                let zelf = Self::_zelf(zelf, vm)?;
+                match zelf.member.getter_or_offset {
+                    Either::A(getter) => (getter)(x, vm),
+                    Either::B(offset) => get_slot_from_object(x, offset, &zelf.member, vm),
+                }
```
youknowone, line 124: When `zelf.member.getter` is `MemberGetter`,

```suggestion
                zelf.member.getter.get(x, vm)
```

moreal:
Currently, this pull request add 381 lines with 12 commits and it needs to implement setter also. And I saw passing CI with success before applying reviews so I expect it will pass now too. So I thought it can stop working on this pull request in current step and should implement the setter in the next PR. 🤔 

I'll make this pull request as `Ready for review`

youknowone (approved):
`PyType::slot_new` is now huge chunk of side effects. I want to break them down but it doesn't look easy.

PR #4107, opened 2022-08-19
Build `lalrpop` without `lalrpop binary`

moreal:
When I started to work on RustPython project today, I met compile error about `python.rs`. I guess it occurred because the `python.rs`, generated lalrpop file, was removed and it became to generate the file every compile. And the error message guided me to install `lalrpop` by using `cargo install lalrpop` command.

This pull request makes the `parser/build.rs` uses `lalrpop` crates instead of `lalrpop` binary.

You can test with the below commands:

```
# Remove lalrpop if it was installed already
cargo uninstall lalrpop

# Remove generated python.rs
rm -rf ./target

cargo build
```


youknowone:
maybe i'd better to find a better way to do it.

moreal:
> maybe i'd better to find a better way to do it.

@youknowone With more research, I know it can run the `lalrpop` in Rust code directly, instead of binary. I made it always use `lalrpop` in Rust code directly. Is there a case to require using lalrpop binary, that I don't know?

youknowone:
No, actually I recently added it as an optional dependency. @coolreader18 wanted to avoid lalrpop dependency from CI side.

moreal:
> No, actually I recently added it as an optional dependency. @coolreader18 wanted to avoid lalrpop dependency from CI side.

Ok, I see. Then I think it'll be the best if `lalrpop` is used in default and the CI excludes the `lalrpop` with some args.

https://github.com/rust-lang/cargo/issues/3126 seems able to implement my thought but it seems not resolved yet.

At now, I'll make this pull request as draft because I'm not sure what is the solution 🤔 

Current my thought is:

- Add `rustpython-parser/lalrpop` feature into `/Cargo.toml` `default` feature.
- Make CI to use an option to exclude the feature like `--features=-rustpython-parser/lalrpop`. (https://github.com/rust-lang/cargo/issues/3126)

moreal:
I opened again because @youknowone pushed [ca256d3](https://github.com/RustPython/RustPython/pull/4107/commits/ca256d329204b1ca624ecbecf53fac5cb8fda03b) commit and the CI seems passed well.

youknowone (approved):
thank you for good idea

Issue #4108, opened 2022-08-20
Syntax error with named expressions in generators


charliermarsh:
# Summary

This yields a syntax error in `RustPython` (`invalid syntax. Got unexpected token 'for'`):

```
a = 1
b = [1, 2, 3, 4]
genexp = (c := i + a for i in b)
```

CPython returns this AST:

```
Module(body=[Assign(targets=[Name(id='genexp', ctx=Store())], value=GeneratorExp(elt=NamedExpr(target=Name(id='c', ctx=Store()), value=BinOp(left=Name(id='i', ctx=Load()), op=Add(), right=Name(id='a', ctx=Load()))), generators=[comprehension(target=Name(id='i', ctx=Store()), iter=Name(id='b', ctx=Load()), ifs=[], is_async=0)]), type_comment=None)], type_ignores=[])
```

PR #4109, opened 2022-08-20
Allow named expressions in generator comprehensions

charliermarsh:
# Summary

See: #4108.

I noticed this when trying to parse the CPython codebase itself, which has this test (on which `RustPython` throws a syntax error):

```
def test_named_expression_scope_in_genexp(self):
    a = 1
    b = [1, 2, 3, 4]
    genexp = (c := i + a for i in b)

    self.assertNotIn("c", locals())
    for idx, elem in enumerate(genexp):
        self.assertEqual(elem, b[idx] + a)
```

### DIFF
--- a/extra_tests/snippets/syntax_generator.py
+++ b/extra_tests/snippets/syntax_generator.py
@@ -17,6 +17,9 @@ assert r == [1, 2, 42, 3]
 r = list(x for x in [1, 2, 3])
 assert r == [1, 2, 3]
 
+r = list(y := x + 1 for x in [1, 2, 3])
+assert r == [2, 3, 4]
+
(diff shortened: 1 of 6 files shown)


youknowone (commented):
Looks great! Because we don't have related cpython unittest for this grammer,
Adding a test under `extra_tests/snippets/syntax_generator.py` will ensure both CPython and RustPython have same behavior

charliermarsh:
Done!

```
...
test_snippets.py::SampleTestCase::test_rustpython_syntax_generator PASSED
...
```

youknowone (approved):
Thank you!

PR #4110, opened 2022-08-20
Implements 6 os/posix constants missing for macos

ecpost:
Implements 6 os/posix constants missing for macos:
- SEEK_DATA and SEEK_HOLE already exist in posix.rs for a few other target_os's, but they are re-exported from libc, which does not currently include them for macos.  So the the consts in this PR are hard-coded.
- O_EVTONLY, O_FSYNC, O_SYMLINK, and O_NOFOLLOW_ANY did not yet exist in posix.rs.  Per the CPython docs, these are only available in macOS.  2 of the 4 are not available in libc.  For consistency, 4 all are hard-coded in this PR.

I suppose it would be better to send a PR to libc to have these added (and I'm happy to do that as well), rather than hard-coding them here.  But I don't know how long it would take for that change (in whatever libc version) to make its way back here.  So I did them as constants that can be switched to libc re-exports if/when they're available in the future.

Ref: #1175


youknowone:
Thank you! libc is very responsive and they have frequent release schedules. Once you submit a proper patch, it will be available in 2 weeks in high chance.

vm/src/stdlib/posix.rs
```diff
@@ -60,6 +60,26 @@ pub mod module {
...
+    const SEEK_DATA: i32 = 4;
+    #[cfg(target_os = "macos")]
+    #[pyattr]
+    const SEEK_HOLE: i32 = 3;
+    #[cfg(target_os = "macos")]
+    #[pyattr]
+    const O_EVTONLY: i32 = 0x8000;
+    #[cfg(target_os = "macos")]
+    #[pyattr]
+    const O_FSYNC: i32 = 0x80;
```
youknowone, line 75: https://docs.rs/libc/latest/x86_64-apple-darwin/libc/constant.O_FSYNC.html

vm/src/stdlib/posix.rs
```diff
@@ -60,6 +60,26 @@ pub mod module {
...
+    #[cfg(target_os = "macos")]
+    #[pyattr]
+    const O_SYMLINK: i32 = 0x200000;
```
youknowone, line 78: https://docs.rs/libc/latest/x86_64-apple-darwin/libc/constant.O_SYMLINK.html

youknowone (changes requested):

youknowone:
libc update: https://github.com/rust-lang/libc/pull/2886

ecpost:
Thanks for submitting the PR to libc!  I saw your reply last night before I went to bed and was going to do that myself in the morning -- I should've let you know last night, sorry.

For the 2 requested changes, did you just want me to go ahead and use libc for the 2 it has already?  (I had purposely done them all as constants for consistency.  But I can fix those 2.  Just confirming that's what you meant.)

youknowone:
I want to keep Python-independent things as much as possible outside of RustPython.

To be honest, if you are not planning to work on another issue depending on this PR, I want to make everything referring libc before merging - unless they take too long time.
If you want to merge it immediately, yes, I want to use libc as much as possible. I want to use libc for those 2 values.

ecpost:
Yeah, that makes sense.  Let's leave it open until libc has them all.  (No rush on my side.  Just wanted to do an easy first PR on the project.)

youknowone:
Thank you for understanding! I will notify you when it done.

youknowone:
The PR is merged. It will be released in the next version (probably [0.2.133](https://github.com/rust-lang/libc/tree/0.2.133))

youknowone:
@ecpost libc is updated! it took a lot longer time than expected.

youknowone (approved):
Thank you!

PR #4116, opened 2022-08-21
Avoid creating unused JoinedStr in FStringParser

charliermarsh:
# Summary

Per discussion in #4097, we can now avoid creating a `JoinedStr` entirely when parsing f-strings, and instead return the vector of values.

### DIFF
--- a/compiler/parser/src/fstring.rs
+++ b/compiler/parser/src/fstring.rs
@@ -4,6 +4,7 @@ use crate::{
     error::{FStringError, FStringErrorType, ParseError},
     parser::parse_expression,
 };
+use itertools::Itertools;
 use std::{iter, mem, str};
 
 struct FStringParser<'a> {
@@ -105,13 +106,16 @@ impl<'a> FStringParser<'a> {
                                 nested -= 1;
                                 if nested == 0 {
                                     formatted_value_piece.push(next);
-                                    spec_constructor.extend(
-                                        FStringParser::new(
-                                            &format!("{{{}}}", formatted_value_piece),
-                                            Location::default(),
-                                            &self.recurse_lvl + 1,
-                                        )
-                                        .parse()?,
+                                    let values = FStringParser::new(
+                                        &formatted_value_piece,
+                                        Location::default(),
+                                        &self.recurse_lvl + 1,
+                                    )
+                                    .parse()?;
+                                    spec_constructor.push(values
+                                        .into_iter()
+                                        .exactly_one()
+                                        .expect("Expected formatted value to produce exactly one expression.")
                                     );
                                     formatted_value_piece.clear();
                                 } else {
--- a/compiler/parser/src/fstring.rs
+++ b/compiler/parser/src/fstring.rs
@@ -105,19 +105,13 @@ impl<'a> FStringParser<'a> {
                                 nested -= 1;
                                 if nested == 0 {
                                     formatted_value_piece.push(next);
-                                    spec_constructor.push(
-                                        self.expr(ExprKind::FormattedValue {
-                                            value: Box::new(
-                                                FStringParser::new(
-                                                    &formatted_value_piece,
-                                                    Location::default(),
-                                                    &self.recurse_lvl + 1,
-                                                )
-                                                .parse()?,
-                                            ),
-                                            conversion: ConversionFlag::None as _,
-                                            format_spec: None,
-                                        }),
+                                    spec_constructor.extend(
+                                        FStringParser::new(
+                                            &format!("{{{}}}", formatted_value_piece),
+                                            Location::default(),
+                                            &self.recurse_lvl + 1,
+                                        )
+                                        .parse()?,
                                     );
                                     formatted_value_piece.clear();
                                 } else {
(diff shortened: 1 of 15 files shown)


parser/src/fstring.rs
```diff
@@ -130,11 +124,13 @@ impl<'a> FStringParser<'a> {
...
-                                    kind: None,
-                                }));
-                                constant_piece.clear();
+                                if !constant_piece.is_empty() {
+                                    spec_constructor.push(self.expr(ExprKind::Constant {
+                                        value: constant_piece.to_owned().into(),
+                                        kind: None,
+                                    }));
+                                    constant_piece.clear();
+                                }
```
charliermarsh, line 133: This was another difference I noticed.

In CPython, you get:

```py
>>> import ast
>>> print(ast.dump(ast.parse('f"{foo:{spec}}"')))
Module(body=[Expr(value=JoinedStr(values=[FormattedValue(value=Name(id='foo', ctx=Load()), conversion=-1, format_spec=JoinedStr(values=[FormattedValue(value=Name(id='spec', ctx=Load()), conversion=-1)]))]))], type_ignores=[])
```

Prior to this change, the RustPython AST was similar, except that the inner `FormattedValue` was padded by an empty `Constant` on both sides.

youknowone:
parser/src/fstring.rs conflicts to #4101
could you rebase it to see CI result?

Snapshots will make annoying conflicts. My suggestion is to remove them from the current working commit and rebase it only with fstring.rs
(`git checkout upstream/main -- parser/src/snapshots/* && git add parser/src/snapshots/ && git commit --ame`)

charliermarsh:
Yup, on it.

charliermarsh:
Looking into a few of the test failures.

parser/src/fstring.rs
```diff
@@ -105,19 +105,13 @@ impl<'a> FStringParser<'a> {
...
-                                                    &self.recurse_lvl + 1,
-                                                )
-                                                .parse()?,
-                                            ),
-                                            conversion: ConversionFlag::None as _,
-                                            format_spec: None,
-                                        }),
+                                    spec_constructor.extend(
+                                        FStringParser::new(
+                                            &format!("{{{}}}", formatted_value_piece),
```
youknowone, line 110: the parsing string is regressed

charliermarsh: Thanks, bad rebase -- working on it.

parser/src/fstring.rs
```diff
@@ -107,14 +107,14 @@ impl<'a> FStringParser<'a> {
...
+                                            value: Box::new(self.expr(ExprKind::JoinedStr {
+                                                values: FStringParser::new(
                                                     &formatted_value_piece,
                                                     Location::default(),
                                                     &self.recurse_lvl + 1,
                                                 )
-                                                .parse()?,
-                                            ),
+                                                    .parse()?
+                                            })),
```
charliermarsh, line 117: This yields a slightly different AST than CPython, we're returning a `FormattedValue(JoinedStr(FormattedValue(` whereas CPython just returns a `FormattedValue`.

youknowone:
I am really sorry but I just finished massive relocation from #4119. parser directory is moved under compiler/parser.
Could you please rebase once again like before without snapshots?
These kinds of big changes will not happen again for a while - I promise.

charliermarsh:
Yeah no worries!

parser/src/fstring.rs
youknowone: We thought the return type can be changed, but maybe actually not? I thought The returned JoinedStr is always a vector of strings, but it seems it had a nested structure.

charliermarsh: Yeah, although the returned `JoinedStr` can either contain `FormattedValue`, `Constant`, or `JoinedStr`, right? I think the code as-written does work, it just includes some additional nesting (but still fewer levels than before IIUC).

compiler/parser/src/fstring.rs
```diff
@@ -105,19 +106,16 @@ impl<'a> FStringParser<'a> {
...
-                                        }),
+                                    let values = FStringParser::new(
+                                        &formatted_value_piece,
+                                        Location::default(),
+                                        &self.recurse_lvl + 1,
+                                    )
+                                    .parse()?;
+                                    spec_constructor.push(values
+                                        .into_iter()
+                                        .exactly_one()
```
youknowone, line 117: I worried if this is possible to be broken, but `ExpressionNestedTooDeeply` seems to prevent it successfully.

youknowone (commented):
I am ok with ast incompatibility if its behavior is the same.

@coolreader18 how do you think about it?

parser/src/fstring.rs
charliermarsh: As written, this does now return `format_spec: Some(JoinedStr(FormattedValue(`, just like CPython. (The `exactly_one` enforcement was necessary, AFAICT.)

fanninpm:
As an aside, how far does RustPython get through https://github.com/python/cpython/blob/3.10/Lib/test/test_fstring.py without raising a `SyntaxError` and bailing out?

youknowone (approved):
looks great. I misunderstood the changed test. So the ast seems fully compatible now.

PR #4122, opened 2022-08-22
Add built-in `breakpoint` function

jopemachine:
## Outline

- Add built-in `breakpoint` function

## Result

```
>>>>> breakpoint()
--Return--
> /Users/igyubong/Desktop/RustPython/Lib/bdb.py(336)set_trace()->None
-> sys.settrace(self.trace_dispatch)
(Pdb) 
```

## References

- [Python breakpoint documentation](https://docs.python.org/3.7/library/functions.html#breakpoint)
- [cpython breakpoint implementation](https://github.com/python/cpython/blob/main/Python/sysmodule.c#L499-L582)

### DIFF
--- a/vm/src/stdlib/builtins.rs
+++ b/vm/src/stdlib/builtins.rs
@@ -365,8 +365,8 @@ mod builtins {
         obj.hash(vm)
     }
 
-    #[pyfunction(name = "breakpoint")]
-    fn builtin_breakpoint(args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+    #[pyfunction]
+    fn breakpoint(args: FuncArgs, vm: &VirtualMachine) -> PyResult {
         let hook =
             sys::get_object("breakpointhook".to_owned(), vm).expect("lost sys.breakpointhook");
 
--- a/vm/src/stdlib/builtins.rs
+++ b/vm/src/stdlib/builtins.rs
@@ -367,10 +367,13 @@ mod builtins {
 
     #[pyfunction]
     fn breakpoint(args: FuncArgs, vm: &VirtualMachine) -> PyResult {
-        let hook =
-            sys::get_object("breakpointhook".to_owned(), vm).expect("lost sys.breakpointhook");
-
-        vm.invoke(hook.as_ref(), args)
+        match vm
+            .sys_module
+            .get_attr(vm.ctx.intern_str("breakpointhook"), vm)
+        {
+            Ok(hook) => vm.invoke(hook.as_ref(), args),
+            Err(_) => Err(vm.new_runtime_error("lost sys.breakpointhook".to_owned())),
+        }
     }
 
     #[pyfunction]
--- a/vm/src/stdlib/builtins.rs
+++ b/vm/src/stdlib/builtins.rs
@@ -78,8 +78,6 @@ mod builtins {
         }
     }
 
-    // builtin_breakpoint
-
     #[pyfunction]
     fn callable(obj: PyObjectRef, vm: &VirtualMachine) -> bool {
         vm.is_callable(&obj)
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/builtins.rs
```diff
@@ -367,7 +365,13 @@ mod builtins {
         obj.hash(vm)
     }
 
-    // builtin_help
+    #[pyfunction(name = "breakpoint")]
+    fn builtin_breakpoint(args: FuncArgs, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 369:
```suggestion
    #[pyfunction]
    fn breakpoint(args: FuncArgs, vm: &VirtualMachine) -> PyResult {
```
if name doesn't conflict, we can use the python function name here

vm/src/stdlib/sys.rs
```diff
@@ -316,6 +317,57 @@ mod sys {
...
+                    (
+                        envar[..(envar.len() - last.len() - 1)].to_owned(),
+                        last.to_owned(),
+                    )
+                }
+            }
+            None => {
+                return print_unimportable_module_warn();
+            }
+        };
```
youknowone, line 358: let's reduce indent depth.

```suggestion
        let last = match envar.split('.').last().ok_or_else(print_unimportable_module_warn)?;
        let (modulepath, attrname) = if last.eq(&envar) {
            ("builtins".to_owned(), envar.to_owned())
        } else {
            (
                envar[..(envar.len() - last.len() - 1)].to_owned(),
                 last.to_owned(),
            )
        };
```

Please check the next block too.

youknowone (changes requested):
Looks good. In `test_builtin.py` `TestBreakpoint` is marked as skip. will you remove the marker and see how it is going?

vm/src/stdlib/sys.rs
jopemachine: I think the error handling logic here is printing warning log and returning `vm.ctx.none`, so I think maybe error propagating can not be used here.

I'm trying to find good way to reduce indents, how about one?

```rust
        let last = match envar.split('.').last() {
            Some(last) => last,
            None => {
                return print_unimportable_module_warn();
            }
        };

        let (modulepath, attrname) = if last.eq(&envar) {
            ("builtins".to_owned(), envar.to_owned())
        } else {
            (
                envar[..(envar.len() - last.len() - 1)].to_owned(),
                last.to_owned(),
            )
        };

        let module = match vm.import(modulepath, None, 0) {
            Ok(module) => module,
            Err(_) => {
                return print_unimportable_module_warn();
            }
        };

        match vm.get_attribute_opt(module, attrname) {
            Ok(Some(hook)) => vm.invoke(hook.as_ref(), args),
            _ => print_unimportable_module_warn(),
        }
```

youknowone: Oh, I am sorry. you are right

vm/src/stdlib/sys.rs
```diff
@@ -878,6 +943,11 @@ pub fn get_stderr(vm: &VirtualMachine) -> PyResult {
         .map_err(|_| vm.new_runtime_error("lost sys.stderr".to_owned()))
 }
 
+pub fn get_object(name: String, vm: &VirtualMachine) -> PyResult {
+    let dict = vm.sys_module.dict();
+    dict.get_item(&name, vm)
+}
+
```
youknowone, line 950:
```suggestion
```

youknowone: access to module dict is not an expected way

vm/src/stdlib/builtins.rs
```diff
@@ -367,7 +365,13 @@ mod builtins {
         obj.hash(vm)
     }
 
-    // builtin_help
+    #[pyfunction]
+    fn breakpoint(args: FuncArgs, vm: &VirtualMachine) -> PyResult {
+        let hook =
+            sys::get_object("breakpointhook".to_owned(), vm).expect("lost sys.breakpointhook");
```
youknowone, line 371: To obtain an attribute from module,
```rust
        let breakpointhook = vm.ctx.intern_str("breakpointhook");
        vm.sys_module.get_attribute(breakpointhook, vm)
```

youknowone, line 371: The `lost sys.breakpointhook` error is not a panic but a `RuntimeError` in CPython.

youknowone (changes requested):
Thank you! I left a few comments about error handling

youknowone:
The missing test for `del sys.breakpointhook`: https://github.com/python/cpython/issues/96197

PR #4125, opened 2022-08-23
Rework marshal.rs to use an error enum

coolreader18:
- Rework frozenset construction
- Rework marshal.rs to use an error enum

### DIFF
--- a/vm/src/stdlib/marshal.rs
+++ b/vm/src/stdlib/marshal.rs
@@ -8,7 +8,7 @@ mod decl {
             PyList, PySet, PyStr, PyTuple,
         },
         bytecode,
-        convert::ToPyObject,
+        convert::{IntoPyException, ToPyObject, ToPyResult},
         function::{ArgBytesLike, OptionalArg},
         object::AsObject,
         protocol::PyBuffer,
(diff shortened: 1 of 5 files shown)


vm/src/stdlib/marshal.rs
```diff
@@ -237,158 +196,172 @@ mod decl {
...
-            Type::Float => {
-                if buf.len() < 8 {
-                    return Err(too_short_error(vm));
-                }
-                let (bytes, buf) = buf.split_at(8);
-                let number = f64::from_le_bytes(bytes.try_into().unwrap());
-                (vm.ctx.new_float(number).into(), buf)
+                let len = i32::from_le_bytes(*buf.read_n_const::<4>()?);
+                let sign = if len < 0 { Sign::Minus } else { Sign::Plus };
+                let len = len.unsigned_abs() as usize;
```
youknowone, line 300: Not a big deal but this is changed from single comparison to double comparison (symentically)

```suggestion
                let (sign, len) = if len < 0 {
                    (Sign::Minus, (-len) as usize)
                } else {
                    (Sign::Plus, len as usize)
                };
```

fanninpm: I actually prefer what @coolreader18 wrote here because its intention is clearer. However, I do have one improvement for the first line:

```suggestion
                let sign = if len.is_negative() { Sign::Minus } else { Sign::Plus };
                let len = len.unsigned_abs() as usize;
```

Issue #4088, opened 2022-08-17
function __qualname__ is not getset


youknowone:
## Feature

CPython
```
>>> def f(): pass
... 
>>> f.__qualname__
'f'
>>> del f.__qualname__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: __qualname__ must be set to a string object
>>> 
```
See func_set_qualname

RustPython
```
Welcome to the magnificent Rust Python 0.1.2 interpreter 😱 🖖
>>>>> def f(): pass
>>>>> 
>>>>> f.__qualname__
'f'
>>>>> f.__qualname__ = 'x'
>>>>> f.__qualname__
'x'
>>>>> del f.__qualname__
>>>>> f.__qualname__
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
AttributeError: 'function' object has no attribute '__qualname__'
```

youknowone:
generator(PyGenerator), coroutine, and asyncgenerator also have similar issues.

oow214:
@youknowone Could you assign me to this issue?

PR #4126, opened 2022-08-23
__qualname__ getset edit

Lithium07z:
I`m editing issue [ function __qualname__ is not getset #4088 ], but It doesn`t work on now 
I don`t know that how can i fill up the [ let qualname  = ~~~ ]


youknowone:
Thank you for contributing. We revised __qualname__ multiple times and now it is working good.

Issue #4013, opened 2022-08-07
implement warnings more


youknowone:
## Feature

- base: #3736

related source code: https://github.com/python/cpython/blob/v3.10.6/Python/_warnings.c#L940-L956

not all features needed to be implemented at once. step by step progress will be easier.

oow214:
@youknowone Could you assign me to this issue?

youknowone:
we now have better warnings implementation

PR #4130, opened 2022-08-23
Update warning setup_context

oow214:
Update warning module https://github.com/RustPython/RustPython/issues/4013
- [x] is_internal_frame
- [x] next_external_frame
- [x] setup_context

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -272,17 +272,20 @@ impl FrameRef {
         filename.as_str().contains("importlib") && filename.as_str().contains("_bootstrap")
     }
 
-    pub fn next_external_frame(&self, vm: &VirtualMachine) -> FrameRef {
-        let mut frame = self.clone();
+    pub fn next_external_frame(&self, vm: &VirtualMachine) -> Option<FrameRef> {
+        let mut frame = self.clone().f_back(vm);
 
-        while let Some(f) = frame.clone().f_back(vm) {
-            if !f.is_internal_frame() {
-                break;
+        loop {
+            if let Some(back) = frame.clone() {
+                if back.is_internal_frame() {
+                    break Some(back);
+                } else {
+                    frame = back.f_back(vm);
+                }
+            } else {
+                break frame;
             }
-            frame = f;
         }
-
-        frame
     }
 }
 
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -273,19 +273,17 @@ impl FrameRef {
     }
 
     pub fn next_external_frame(&self, vm: &VirtualMachine) -> Option<FrameRef> {
-        let mut frame = self.clone().f_back(vm);
-
-        loop {
-            if let Some(back) = frame.clone() {
-                if back.is_internal_frame() {
-                    break Some(back);
-                } else {
-                    frame = back.f_back(vm);
-                }
+        self.clone().f_back(vm).map(|mut back| loop {
+            back = if let Some(back) = back.to_owned().f_back(vm) {
+                back
             } else {
-                break frame;
+                break back;
+            };
+
+            if !back.is_internal_frame() {
+                break back;
             }
-        }
+        })
     }
 }
 
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -272,7 +272,7 @@ impl FrameRef {
         filename.as_str().contains("importlib") && filename.as_str().contains("_bootstrap")
     }
 
-    pub fn next_external_frame(&self, vm: &VirtualMachine) -> Option<FrameRef> {
+    pub fn next_external_frame(&self, vm: &VirtualMachine) -> FrameRef {
         let mut frame = self.clone();
 
         while let Some(f) = frame.clone().f_back(vm) {
@@ -282,7 +282,7 @@ impl FrameRef {
             frame = f;
         }
 
-        Some(frame)
+        frame
     }
 }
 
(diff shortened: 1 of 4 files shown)


vm/src/warn.rs
```diff
@@ -72,32 +72,39 @@ fn setup_context(
...
+                while let Some(frame) = f.clone() {
+                    stack_level -= 1;
+                    if stack_level > 0 {
+                        break;
+                    }
+                    f = frame.next_external_frame(vm);
+                }
+            }
+        }
+    }
```
Snowapril, line 104:
```suggestion
    if stack_level <= 0 || f.clone().map(|frame| frame.is_internal_frame()).unwrap_or_default() {
        loop {
            stack_level -= 1;
            if stack_level <= 0 || f.is_none() {
                break;
            }
            f = f.unwrap().f_back(vm);
        }
    } else {
        loop {
            stack_level -= 1;
            if stack_level <= 0 || f.is_none() {
                break;
            }
            f = f.unwrap().next_external_frame(vm);
        }
    }
```

youknowone: I don't support `f.is_none()` and `f.unwrap()` combination when it doesn't give a significant advantage. It looks ok for now, but that works only when human carefully keep the pair in the correct order. We can write the same logic with `if let` matches which prevent mistakes from future logic changes.

vm/src/warn.rs
```diff
@@ -72,32 +72,34 @@ fn setup_context(
...
+        || f.clone()
+            .map(|frame| frame.is_internal_frame())
+            .unwrap_or_default()
+    {
+        loop {
+            stack_level -= 1;
+            if stack_level <= 0 || f.is_none() {
+                break;
+            }
+            f = f.unwrap().f_back(vm);
```
youknowone, line 89:
```suggestion
            if stack_level <= 0 {
                break;
            }
            if let Some(f) = f {
                f = f.f_back(vm);
            } else {
                break;
            }
```

vm/src/frame.rs
```diff
@@ -264,6 +264,23 @@ impl FrameRef {
...
+        filename.as_str().contains("importlib") && filename.as_str().contains("_bootstrap")
+    }
+
+    pub fn next_external_frame(&self, vm: &VirtualMachine) -> Option<FrameRef> {
+        self.clone().f_back(vm).map(|back| {
+            if back.is_internal_frame() {
+                back
+            } else {
+                back.f_back(vm).unwrap()
+            }
```
youknowone, line 281: The function name and its behavior looks confusing.
The name is `next_external_frame` but the condition is `back.is_internal_frame()` than return the value.
If this is a correct implementation, we'd better to add comment how this is working.

vm/src/warn.rs
```diff
@@ -72,32 +72,41 @@ fn setup_context(
...
-    //         while (--stack_level > 0 && f != NULL) {
-    //             f = next_external_frame(f);
-    //         }
-    //     }
+    // Stack level comparisons to Python code is off by one as there is no
+    // warnings-related stack level to avoid.
+    if stack_level <= 0
+        || f.clone()
+            .map(|frame| frame.is_internal_frame())
+            .unwrap_or_default()
```
youknowone, line 82: maybe redundant clone?
```suggestion
        || f.map_or(false, |frame| frame.is_internal_frame())
```

vm/src/warn.rs
```diff
@@ -72,32 +72,41 @@ fn setup_context(
...
+    {
+        loop {
+            stack_level -= 1;
+            if stack_level <= 0 || f.is_none() {
+                break;
+            }
```
youknowone, line 88: redundant check with line 89 and 92

vm/src/warn.rs
```diff
@@ -72,32 +72,41 @@ fn setup_context(
...
+            if let Some(tmp) = f {
+                f = tmp.f_back(vm);
+            } else {
+                break;
+            }
+        }
+    } else {
+        loop {
+            stack_level -= 1;
+            if stack_level <= 0 || f.is_none() {
```
youknowone, line 101: the `is_none()` and `unwrap()` pair here also can be eliminated by using line 89 and 92 patterns

youknowone (commented):
looks good in general. I have a question about next_external_frame and a few more small change requests.

vm/src/frame.rs
```diff
@@ -264,6 +264,29 @@ impl FrameRef {
...
+    // If the back frame is an internal frame, it is judged as a new external frame and the loop is stopped.
+    pub fn next_external_frame(&self, vm: &VirtualMachine) -> Option<FrameRef> {
+        self.clone().f_back(vm).map(|mut back| loop {
+            back = if let Some(back) = back.to_owned().f_back(vm) {
+                back
+            } else {
+                break back;
+            };
+
+            if back.is_internal_frame() {
```
youknowone, line 285: maybe
```suggestion
            if !back.is_internal_frame() {
```

vm/src/warn.rs
```diff
@@ -72,32 +72,38 @@ fn setup_context(
...
-    //         while (--stack_level > 0 && f != NULL) {
-    //             f = next_external_frame(f);
-    //         }
-    //     }
+    // Stack level comparisons to Python code is off by one as there is no
+    // warnings-related stack level to avoid.
+    if stack_level <= 0 || f.as_ref().map_or(false, |frame| frame.is_internal_frame()) {
+        loop {
+            stack_level -= 1;
+            if stack_level <= 0 || f.is_none() {
```
youknowone, line 82:
```suggestion
            if stack_level <= 0 {
```

youknowone (changes requested):

youknowone (approved):
Looks good, thank you!

PR #4132, opened 2022-08-23
Only run lalrpop generation on ubuntu-latest

coolreader18:
Unless I'm misunderstanding something in the gh actions config, running lalrpop on windows and linux doesn't actually do anything - the cache key doesn't have the os in it, so it'll just overwrite each other and it's doing more work than it needs to.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -104,7 +104,6 @@ jobs:
     env:
       RUST_BACKTRACE: full
     name: Run rust tests
-    needs: lalrpop
     runs-on:  ${{ matrix.os }}
     strategy:
       matrix:
@@ -114,11 +113,9 @@ jobs:
       # for windows, always checkout files with their original lf line endings
       - run: git config --global core.autocrlf false; git config --global core.eol lf
       - uses: actions/checkout@v2
-      - name: Cache generated parser
-        uses: actions/cache@v3
-        with:
-          path: compiler/parser/python.rs
-          key: lalrpop-${{ hashFiles('compiler/parser/python.lalrpop') }}
+      - uses: actions/cache@v3
+        with: { path: "~/.cargo/bin/lalrpop\n~/.cargo/.crates.toml", key: "lalrpop-${{ runner.os }}" }
+      - run: cargo install lalrpop@0.19.8
       - uses: dtolnay/rust-toolchain@stable
       - name: Set up the Windows environment
         shell: bash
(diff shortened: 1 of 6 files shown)


youknowone:
I wish if it were true. Mysteriously, they generate different file hash. Check hash values from any other builds: https://github.com/RustPython/RustPython/runs/7983393352?check_suite_focus=true
You will find this version of CI fails if you change python.lalrpop.

coolreader18:
crlf vs lf issue (hopefully) :)

coolreader18:
Well that's strange....

windows:
Cache not found for input keys: lalrpop-299a3f1010a7175d7b8df916d49cbc7d2121b53be5ae36fbd6a90aac9a15a97f

ubuntu
Cache restored from key: lalrpop-299a3f1010a7175d7b8df916d49cbc7d2121b53be5ae36fbd6a90aac9a15a97f

coolreader18:
ohohoho, it was that AND actions/cache#888. updated to actions/cache@v3 to fix it.

coolreader18:
I give up, I think it's easier to cache the binary for each platform.

coolreader18:
Okay, looks like it doesn't actually cache it yet but I'm okay with this for now - I've requested lalrpop be added to [cargo-quickinstall](https://github.com/alsuren/cargo-quickinstall)'s binary package cache so in the future we can use that.

coolreader18:
I know I've not been around for a couple months, but I really do still think that just checking the generated file into the repository is the easiest way to do this - it cuts down on compile times for everyone, and it *just works* for first contributors (and ci, hahahaaa): no having to remember to pass the correct feature, or install an external tool. Just `cargo build` and you have an interpreter.

coolreader18:
(ok now it does really cache the binary in the gh actions cache)

youknowone:
previously, lalrpop job was only checking hash and used cached crate install for lalrpop. Check the previous CI results. running cost for lalrpop was very low unless lalrpop was actually required. If you care about CI cost, this changed version increases the CI cost.

> no having to remember to pass the correct feature, or install an external tool

#4129 also do it. please check it.

youknowone:
Oh, I found you made your own cache of lalrpop binary.

youknowone:
optional lalrpop feature of parser revealed not working well. going back to this way would be better

DimitrisJim:
Is this issue relevant now that Parser lives in a separate repo? (Should it be moved there?)

youknowone:
The generated parser is bundled again in the new repository

PR #4133, opened 2022-08-24
Split pylib from vm

youknowone:
now vm/pylib builds parallel, which enhances (re)build time a lot when freeze-stdlib given


src/lib.rs
```diff
@@ -248,6 +248,38 @@ fn add_stdlib(vm: &mut VirtualMachine) {
...
+    #[cfg(not(feature = "freeze-stdlib"))]
+    {
+        use rustpython_vm::common::rc::PyRc;
+        let state = PyRc::get_mut(&mut vm.state).unwrap();
+
+        #[allow(clippy::needless_collect)] // false positive
+        let path_list: Vec<_> = state.settings.path_list.drain(..).collect();
+
+        // add the current directory to sys.path
+        state.settings.path_list.push("".to_owned());
```
youknowone, line 265: this is not stdlib

Issue #4111, opened 2022-08-20
platform.python_implementation returning "CPython"


ecpost:
Python's platform.python_implementation function is now returning "CPython" instead of "RustPython".  I believe it is because Lib/platform.py was refreshed recently (2daddd913dc2bf0dbc046159f28d897508199815) with the latest from CPython, which contains the python_implementation function.  Previously, that function (among other things) had been removed from Lib/platform.py back in 2020 (bbb04c3dedc6af62f2567b5af4292bff97f8780f) so that the Rust version of that function in stdlib/src/platform.rs would handle the implementation.

I noticed this because it breaks whats_left.py.  (There's a check to make sure the sub-process isn't running CPython, and it now thinks it is, even though it's from "cargo run...".)

I can work on a PR, but I wasn't sure if there's more to this, like other related changes in the works or whatever.  If not, I could bascially mimic what was done in bbb04c3dedc6af62f2567b5af4292bff97f8780f.  Let me know if I can help.

youknowone:
Oh, thank you! it was a mistake. It will be great if you can try.

@palaviv do you have advice?

ecpost:
Comparing the last version of platform.py (prior to the 3.10.6 update) to the very first version (which matches the CPython version from late 2019), there are 3 areas with changes:
- The libc_ver function has a TODO comment and immediate return of the defaults at the beginning of the function.
- The _syscmd_uname function has a similar change.
- A large section of code is removed (compiled regexes, a cache, and all the python_* functions that were redone in Rust) and replaced with `from _platform import *`.

That should be easy enough for me to do to the 3.10.6 version.  I can wait a bit to see if palaviv has any comments first.

Regarding the old TODO's in there, I could take a closer look at those too and see if I can take care of them.  I'm thinking it should be a separate PR so the "CPython" platform issue is fixed sooner rather than later.

youknowone:
palaviv is not very active these days. You can go ahead as you want. Your analysis sounds very clear

ecpost:
Ok, thanks.  I have it ready.  I'm not able to build at the moment, I think because of #4117?  I'll submit the PR later...

youknowone:
That must be not a problem when you build from project root. How it fails?

youknowone:
Deleting code can mislead future update if the CPython code is updated - exactly what I thought.
The typical python flow to check native module is checking import error.

```python
try:
    from _platform import *
except ModuleNotFound:
    # Add python function here
```
[...]

PR #4135, opened 2022-08-25
Modify the new 3.10.6 version of platform.py to work with platform.rs

ecpost:
This addresses #4111

@youknowone This is the original change I had planned to do.  See my 3rd comment [here](https://github.com/RustPython/RustPython/issues/4111#issuecomment-1224256627) regarding your try/except comment.  Let me know if you'd like me to do something like that instead.  (Although I'm not sure how the import of _platform would ever fail without other things falling apart first.)


youknowone:
`import _platform` never fails. But if we delete code, it will be unconciously recovered whenever we update platform.py from recent CPython version. deleted code is not easy to track.
I'd like to keep the original code as much as possible not to mislead [library updating process](https://github.com/RustPython/RustPython/wiki/How-to-update-test-files).

I now think `try: import _platform` is not necessary. Instead of it, `from _platform import *` at the bottom of the platform.py will work by overriding the python attributes.

ecpost:
Funny, I was thinking that too, earlier, and gave it a try.  (It worked fine.)  But I figured there was a reason/preference for not doing it that way in the earlier versions.  Anyway, I'll go ahead with that.

(This will also take care of those tests that failed, that I forgot to check first -- oops!  I'm curious though how the pre-3.10.6 version got through with the same issue.  There's a cache variable that was removed, that the test tries to clear.)

BTW, there are those two TODO items in the prior version that short-circuited a couple things.  Let me take a closer look at those to see if they're important.  I'll submit the change after I check those.

youknowone:
> But I figured there was a reason/preference for not doing it that way in the earlier versions.
There could be an historical issue we couldn't do. The change made long time ago and we had a huge progress on compatibility. Maybe we just couldn't make the best decision for every time.

ecpost:
I confirmed, the old TODO's weren't necessary.

However...  The test now fails because they're doing some "interesting" things in test_platform.py, setting sys.version and sys.platform to fake out the platform.* functions.  We could probably support this in platform.rs (have it check sys.version and sys.platform).  But what they're doing isn't really a documented feature, and the rust code will get pretty convoluted.  But what do you think?  Should I just modify the test to handle our rust overrides better?  (That [link](../wiki/How-to-update-test-files) you sent earlier about updating tests was very informative.  Thanks.)

(I have other ideas that would involve changes to platform.py.  But since we're trying to minimize changes there...)

ecpost:
Latest commit assumes that the simplified approach (just the import at the end) and an `expectedFailure` on that specific test is fine.  Happy to make adjustments if necessary.

youknowone:
I am really sorry, but after reviewing the failing test, I think `_platform` was introduced only for historical reason and now we are ready to leave it to past. This is not what I expected when I first reviewed this issue.
I had a busy week but now I can spend more time for RustPython. If you want to look for an issue to step into RustPython, I will try to find one. Please leave a comment.

youknowone (approved):

ecpost:
Understood, no problem.  (Regarding other issues to work on, I was going to do one of the number protocol items when I get a chance later.)

PR #4137, opened 2022-08-26
Refactor Sequence Protocol

qingshi163:


youknowone:
@Snowapril would you review this pr?

vm/src/builtins/type.rs
```diff
@@ -194,7 +204,7 @@ impl PyType {
             None,
         );
 
-        for attr_name in new_type.attributes.read().keys() {
+        for attr_name in attr_name_set {
             if attr_name.as_str().starts_with("__") && attr_name.as_str().ends_with("__") {
```
youknowone, line 208: these check seems to be able to be done when inserting to attr_name_set.

youknowone (approved):

Snowapril:
I also agree to put all sequence methods into function as static local. It'll reduce inessential memory residency. It seems good to me.

vm/src/protocol/mapping.rs
```diff
@@ -45,7 +45,7 @@ impl PyMappingMethods {
         ];
 
         fn length(mapping: &PyMapping, vm: &VirtualMachine) -> PyResult<usize> {
-            crate::types::slot_length(mapping.obj, vm)
+            crate::types::len_wrapper(mapping.obj, vm)
```
Snowapril, line 48: Could this be called through the protocol interface? It seems it'll go through the mro resolution inside python side, doesn't it?

Snowapril: Oh is it for generic cases? Ok I see

PR #4138, opened 2022-08-28
Add `memory_iterator`

jopemachine:
Added `memory_iterator` type implemented in python 3.10

## Refereneces

> [bpo-41732](https://bugs.python.org/issue41732): Added an [iterator](https://docs.python.org/release/3.10.0/glossary.html#term-iterator) to [memoryview](https://docs.python.org/release/3.10.0/library/stdtypes.html#memoryview).

- https://docs.python.org/release/3.10.0/whatsnew/changelog.html#python-3-10-0-final


vm/src/builtins/memory.rs
```diff
@@ -52,6 +57,8 @@ pub struct PyMemoryView {
     // instead it relay on the buffer it viewing to maintain the count
 }
 
+pub type PyMemoryViewRef = PyRef<PyMemoryView>;
```
youknowone, line 60: we are creating these shortcut type alias for only very common types. using `PyRef<PyMemoryView>` will be better.

whats_left.py
```diff
@@ -200,6 +200,7 @@ def gen_methods():
         "type(range(0).__iter__())",
         "type(str().__iter__())",
         "type(tuple().__iter__())",
+        "type(memoryview(bytearray('0', 'utf-8')).__iter__())",
```
youknowone, line 203:
```suggestion
        "type(memoryview(bytearray(b'0')).__iter__())",
```

vm/src/builtins/memory.rs
```diff
@@ -1082,3 +1096,52 @@ fn is_equiv_format(a: &BufferDescriptor, b: &BufferDescriptor) -> bool {
...
+            .builtins_iter_reduce(|x| x.clone().into(), vm)
+    }
+}
+impl Unconstructible for PyMemoryViewIterator {}
+
+impl IterNextIterable for PyMemoryViewIterator {}
+impl IterNext for PyMemoryViewIterator {
+    fn next(zelf: &crate::Py<Self>, vm: &VirtualMachine) -> PyResult<PyIterReturn> {
+        zelf.internal.lock().next(|memoryview, pos| {
+            let mv = memoryview;
```
youknowone, line 1136:
```suggestion
        zelf.internal.lock().next(|mv, pos| {
```

vm/src/builtins/memory.rs
```diff
@@ -1082,3 +1096,52 @@ fn is_equiv_format(a: &BufferDescriptor, b: &BufferDescriptor) -> bool {
...
+            let len = mv.len(vm)?;
+            if pos >= len {
+                Ok(PyIterReturn::StopIteration(None))
+            } else {
+                Ok(PyIterReturn::Return(
+                    mv.getitem_by_idx(pos.try_into().unwrap(), vm)?,
+                ))
+            }
```
youknowone, line 1144:
```suggestion
            Ok(if pos >= len {
                PyIterReturn::StopIteration(None)
            } else {
                PyIterReturn::Return(
                    mv.getitem_by_idx(pos.try_into().unwrap(), vm)?,
                )
            })
```

youknowone (approved):
I left a few style change requests. Other parts looks great, thank you!

PR #4139, opened 2022-08-28
Improve: binaryops with number protocol

TwoPair:
# Environments

- Use `rustpython` binary generated after using `cargo run --release` for faster runtime than debug one.

# Tasks
- [x] 1. Apply only for `multiply` for test
- [ ] 2. Apply for every number methods
   * Sub-tasks
     * ~`PyNumberBinaryOp` implementation~
     * extra_tests/snippets/builtin_complex.py: fix multiply
     * `or` => `type_::or_()` problem
     * `and` => `set` type doesn't have number protocol
     * `mod(remainder)` => `str` module - implant number protocol, compatible with bytes.rs's remainder method
- [ ] 3. Refactor relative codes being useless, and so on.

# Profiling

## Benchmark source

Let me know if you have any good idea for improved benchmark!

```py
# benches/microbenchmarks/multiply.py

import pyperf
import sys

ITERATIONS = 1000000

def multiply():
    integer = 1
    floating = 1.0
    byte_str = b'1'

    for i in range(ITERATIONS):
        integer = (integer * i) % sys.maxsize
        floating = (floating * float(i)) % sys.maxsize
        byte_str = byte_str * (i % 1000)
        # a = i % sys.maxsize
        # complexing = complex(a, a) * a

# multiply()
if __name__ == "__main__":
    runner = pyperf.Runner()
    runner.bench_func('multiply', multiply)
```

## flamegraph

**Before**
flame_before

**After**
flame_after

You can check the call-stack *has decreased* comparing with before!

## pyperf

This represents binaryops with number protocol can improve RustPython's performance just by applying only for multiplying operation.

```
+-----------+-----------------------+---------------------------+----------------------------+
| Benchmark | multiply_cpython_3_10 | multiply_rustpython_after | multiply_rustpython_before |
+===========+=======================+===========================+============================+
| multiply  | 340 ms                | 5.57 sec: 16.36x slower   | 6.14 sec: 18.05x slower    |
+-----------+-----------------------+---------------------------+----------------------------+
```

### DIFF
--- a/vm/src/protocol/number.rs
+++ b/vm/src/protocol/number.rs
@@ -8,6 +8,7 @@ use crate::{
     VirtualMachine,
 };
 use crossbeam_utils::atomic::AtomicCell;
+use std::ptr;
 
 type UnaryFunc<R = PyObjectRef> = AtomicCell<Option<fn(PyNumber, &VirtualMachine) -> PyResult<R>>>;
 type BinaryFunc<R = PyObjectRef> =
@@ -109,6 +110,7 @@ impl PyObject {
 }
 
 #[derive(Default)]
+// #[repr(C)]
 pub struct PyNumberMethods {
     /* Number implementations must check *both*
     arguments for proper type and implement the necessary conversions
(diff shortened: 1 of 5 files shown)


youknowone:
I didn't look into deep, but maybe a bug from PyBytes::mul?

TwoPair:
> I didn't look into deep, but maybe a bug from PyBytes::mul?

Yes, I already recognize where the problems come from. The basic mechanism of multiplying works normally anyway, so I'll fix it bc it doesn't affect the performance.

vm/src/vm/vm_ops.rs
```diff
@@ -219,13 +268,53 @@ impl VirtualMachine {
...
+                self.ctx.not_implemented.as_object(),
+                PyComparisonOp::Eq,
+                self,
+            )? {
+                let seq_a = PySequence::new(a, self);
+                let seq_b = PySequence::new(b, self);
+
+                // TODO: remove the duplicated conditions
+                // The reason I don't put the below sepecial condition into `PyBytes::AsSequence::repeat` is
+                // we cannot use `to_owned()` in `PyBytes::repeat`.
```
youknowone, line 285: you can fix `PyBytes::repeat` to take `PyRef<Self>` or `&Py<Self>` instead of `&self` to allow `to_owned()`

vm/src/protocol/number.rs
```diff
@@ -216,6 +218,97 @@ impl PyNumberMethods {
...
+    InplaceMatrixMultiply,
+}
+
+impl PyNumberMethodsOffset {
+    pub fn method(&self, methods: &PyNumberMethods, vm: &VirtualMachine) -> PyResult<&BinaryFunc> {
+        use PyNumberMethodsOffset::*;
+        unsafe {
+            match self {
+                // BinaryFunc
+                Add => ptr::addr_of!(methods.add),
```
youknowone, line 266: Isn't this working?
```suggestion
                Add => &methods.add,
```

vm/src/protocol/number.rs
```diff
@@ -216,6 +218,97 @@ impl PyNumberMethods {
...
+                InplaceMatrixMultiply => ptr::addr_of!(methods.inplace_matrix_multiply),
+                // UnaryFunc
+                Negative => ptr::null(),
+                Positive => ptr::null(),
+                Absolute => ptr::null(),
+                Boolean => ptr::null(),
+                Invert => ptr::null(),
+                Int => ptr::null(),
+                Float => ptr::null(),
+                Index => ptr::null(),
```
youknowone, line 302: this is user error and we never can reach here only with python code. panic using `unreachable!` macro will be preferred.

vm/src/protocol/number.rs
```diff
@@ -237,8 +330,12 @@ impl PyNumber<'_> {
         obj.class().mro_find_map(|x| x.slots.as_number.load())
     }
 
-    pub fn methods(&self) -> &PyNumberMethods {
-        self.methods
+    pub fn methods<'a>(
```
youknowone, line 333:
```suggestion
    pub fn binary_op<'a>(
```

vm/src/protocol/number.rs
```diff
@@ -216,6 +218,97 @@ impl PyNumberMethods {
     };
 }
 
+pub enum PyNumberMethodsOffset {
```
youknowone, line 221: Name suggestion:
```suggestion
pub enum PyNumberBinaryOp {
```

I don't think we need a single linear enum to represent both binary and unary ops.

vm/src/vm/vm_ops.rs
```diff
@@ -174,160 +177,346 @@ impl VirtualMachine {
...
-                identifier!(self, __mul__),
-                identifier!(self, __rmul__),
-                |vm, a, b| Err(vm.new_unsupported_binop_error(a, b, "*=")),
-            )
+    pub fn _mul(&self, a: &PyObject, b: &PyObject) -> PyResult {
+        self.binary_op(a, b, PyNumberMethodsOffset::Multiply, "*", |vm, a, b| {
+            let seq_a = PySequence::new(a, vm);
+            let seq_b = PySequence::new(b, vm);
+
+            // TODO: I think converting to isize process should be handled in repeat function.
```
youknowone, line 389: who is "I"? the future reader will be confused.

If you think this topic requires discussion rather than leaving concrete TODO lists,
please consider opening an issue and leave here the issue number or link.

youknowone: If allowing arbitary n is not necessary for cpython compatibility, I think `repeat` is better to take `isize` or `usize` because we can skip range check when we already know repeating `n` is possible.
If we are going to remove repetitive pattern, adding `try_repeat` with more inclusive `n` will be possible.

youknowone: another idea is this will be a temporary code.
once all sequences implement number protocol correctly, they will handle it in their own number protocol.

vm/src/vm/vm_ops.rs
```diff
@@ -174,160 +177,348 @@ impl VirtualMachine {
...
+    {
+        let num_a = a.to_number();
+        let num_b = b.to_number();
+
+        let slot_a = num_a.methods(&op_slot, self)?.load();
+        let slot_b = num_b.methods(&op_slot, self)?.load();
+
+        if let Some(slot_a) = slot_a {
+            if let Some(slot_b) = slot_b {
+                // Check if `a` is subclass of `b`
```
xiaozhiyan, line 198: @youknowone 

According to CPython implementation https://github.com/python/cpython/blob/main/Objects/abstract.c#L842 , we may need to ensure `b` is a "strict subclass" of `a` in order to use `slot_b` for operation.

`fast_isinstance` method seems including the situation that `a` & `b` belong to the same class, https://github.com/RustPython/RustPython/blob/main/vm/src/object/ext.rs#L459 .
Is there a method or proper approach to check "strict subclass" please?

youknowone: I don't think we have a function to do that.
But how about `!a.is(b) && a.fast_issubclass(b)`?
when a and b are identical type, `a.is(b)` will be true

vm/src/protocol/number.rs
```diff
@@ -220,8 +314,12 @@ impl PyNumber<'_> {
         obj.class().mro_find_map(|x| x.slots.as_number.load())
     }
 
-    pub fn methods(&self) -> &PyNumberMethods {
-        self.methods
+    pub fn methods<'a>(
+        &'a self,
+        op_slot: &'a PyNumberMethodsOffset,
+        vm: &VirtualMachine,
+    ) -> PyResult<&BinaryFunc> {
```
youknowone, line 321: if we keep the name `binary_op`, the return type will be something other than `BinaryFunc`

vm/src/vm/vm_ops.rs
```diff
@@ -174,160 +177,348 @@ impl VirtualMachine {
...
+                if b.fast_isinstance(a.class()) {
+                    let ret = slot_b(num_a, b, self)?;
+                    if ret.rich_compare_bool(
```
xiaozhiyan, line 201: @youknowone 
Is `rich_compare_bool` slow since it calls `_cmp` which is not a trivial function?
Is it the suggested way to check whether `ret` is `self.ctx.not_implemented`?

xiaozhiyan: How about using `ret.fast_isinstance(PyNotImplemented::class(self))` to check `self.ctx.not_implemented`?

youknowone: I think this check can be fine:
```rust
ret.is(&self.ctx.not_implemented)
```
But not 100% sure because some type can override operators in weird way and rich_compare_bool will respect it. I have no idea about how CPython handle it.

youknowone: Oh, I missed the second comment. In this case, `fast_isinstance` doesn't have any advantage to simple `is` calling. Because `fast_isinstance` doesn't check `__isinstance__`  python-implemented weird things will not be covered. But because `NotImplemented` is always guaranteed to be singleton, `fast_isinstance` doesn't check more thing. So `fast_isinstance` is not strong as `rich_compare` but same weak as simple `is`.

youknowone:
merged through #4615

Issue #4070, opened 2022-08-14
fcntl.ioctl first argument is not only fd. 


youknowone:
## Feature

Adding `test_ioctl.py` will show the missing features

tgsong827:
can you assign this issue to me?

youknowone:
resolved by #4140

PR #4140, opened 2022-08-29
update ioctl first arg

tgsong827:
#4070 

- add `test_ioctl.py`.
- update `fn ioctl` can take others having `fileno` method not only Integer.

### DIFF
--- a/Lib/test/test_ioctl.py
+++ b/Lib/test/test_ioctl.py
@@ -1,0 +1,92 @@
+import array
+import unittest
+from test.support import get_attribute
+from test.support.import_helper import import_module
+import os, struct
+fcntl = import_module('fcntl')
+termios = import_module('termios')
+get_attribute(termios, 'TIOCGPGRP') #Can't run tests without this feature
+
+try:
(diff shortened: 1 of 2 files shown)


Lib/test/test_ioctl.py
```diff
@@ -0,0 +1,94 @@
...
+        # Issue #9758: a mutable buffer of exactly 1024 bytes wouldn't be
+        # copied back after the system call.
+        self._check_ioctl_mutate_len(1024)
+
+    def test_ioctl_mutate_2048(self):
+        # Test with a larger buffer, just for the record.
+        self._check_ioctl_mutate_len(2048)
+
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
youknowone, line 70: We usually create a separated commit for cpython libraries and commits. Please create commits again like this: https://github.com/RustPython/RustPython/pull/4029/commits

youknowone (changes requested):
Rust code changes are looking good.

youknowone (approved):
Thank you!

PR #4143, opened 2022-08-29
Add expression context parsing

charliermarsh:
# Summary

It looks like the parser is _almost_ always setting `ast::ExprContext::Load`. This PR attempts to set `ast::ExprContext::Store` and `ast::ExprContext::Del` in the right places, based on playing around with CPython's own behavior.

I'm not familiar enough with LALRPOP to know if there's a way to do this within the parser itself -- so, instead, I had to "rewrite" the expression tree in the right places.

### DIFF
--- a/compiler/parser/python.lalrpop
+++ b/compiler/parser/python.lalrpop
@@ -8,6 +8,7 @@ use crate::{
     error::{LexicalError, LexicalErrorType},
     function::{ArgumentList, parse_args, parse_params},
     lexer,
+    context::set_context,
     string::parse_strings,
     token::StringKind
 };
@@ -82,7 +83,7 @@ DelStatement: ast::Stmt = {
             location,
             end_location: Some(end_location),
             custom: (),
-            node: ast::StmtKind::Delete { targets },
+            node: ast::StmtKind::Delete { targets: targets.into_iter().map(|expr| set_context(expr, ast::ExprContext::Del)).collect() },
         }
     },
 };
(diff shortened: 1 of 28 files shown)


youknowone:
@charliermarsh maybe you forgot to add `compiler/parser/src/snapshots/rustpython_parser__context__tests__assign_set_comp.snap` file

charliermarsh:
Fixed, thanks!

charliermarsh:
(This code has been part of ruff for a while now and has worked well for me.)

compiler/ast/src/ast_gen.rs
```diff
@@ -281,7 +281,7 @@ pub enum ExprKind<U = ()> {
 }
 pub type Expr<U = ()> = Located<ExprKind<U>, U>;
 
-#[derive(Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
```
youknowone, line 284:
```suggestion
#[derive(Debug, PartialEq)]
```
changed without code generation fix

youknowone (commented):
@coolreader18 @qingshi163 @fanninpm does anyone have a better idea?

charliermarsh:
@fanninpm - Yeah I'm very interested in that. \cc @isidentical who I discussed this with yesterday.

compiler/ast/src/ast_gen.rs
```diff
@@ -283,7 +283,7 @@ pub enum ExprKind<U = ()> {
 }
 pub type Expr<U = ()> = Located<ExprKind<U>, U>;
 
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq)]
```
youknowone, line 286: This change doesn't match to current version of generation script

charliermarsh: 🤦

youknowone (approved):

PR #4149, opened 2022-09-04
Add PartialOrd to Location

charliermarsh:
I'm finding this useful when using RustPython as a library, though if you don't it in RustPython I can just use my own wrapper type.


youknowone (approved):
Thank you!

PR #4150, opened 2022-09-08
Add `math.exp2` implemented in cpython 3.11

jopemachine:
`math.exp2` return 2 raised to the power x.

## Related links

* Python codes are copy-pasted from CPython.

- https://github.com/python/cpython/pull/29829

- [Python documentation](https://docs.python.org/3.11/library/math.html#math.exp2)

### DIFF
--- a/Lib/test/test_math.py
+++ b/Lib/test/test_math.py
@@ -1,7 +1,7 @@
 # Python test set -- math module
 # XXXX Should not do tests around zero only
 
-from test.support import run_unittest, verbose, requires_IEEE_754
+from test.support import verbose, requires_IEEE_754
 from test import support
 import unittest
 import itertools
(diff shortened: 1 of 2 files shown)


youknowone:
The rust part looks great. Could you also check if you followed this guideline?
https://github.com/RustPython/RustPython/wiki/How-to-update-test-files

jopemachine:
> The rust part looks great. Could you also check if you followed this guideline? [RustPython/RustPython/wiki/How-to-update-test-files](https://github.com/RustPython/RustPython/wiki/How-to-update-test-files)

Thanks for letting me know!
I just made a fixing commit.

I didn't include unrelated diffs with `math.exp2` because maybe including these diffs might be out of this PR's scope.

youknowone:
We usually 'sync' our version to a specific CPython version, instead of updating each function. The bulk update is easy by following the guideline.

jopemachine:
> We usually 'sync' our version to a specific CPython version, instead of updating each function. The bulk update is easy by following the guideline.

Understood. I attached `@unittest.expectedFailure` to the expected failed tests.

This could be fixed by other 2 PRs that I made.

youknowone (approved):
Great, thank you!

PR #4152, opened 2022-09-09
Fix edge case of `math.pow(0.0, -inf)`, `math.pow(-0.0, -inf)`

jopemachine:
Python documentation says `The special cases pow(0.0, -inf) and pow(-0.0, -inf) were changed to return inf instead of raising ValueError, for consistency with IEEE 754`

## Related links

- [Python documentation](https://docs.python.org/ko/3.11/library/math.html#math.pow)

- [Python 3.11 change log](https://github.com/python/cpython/blob/main/Doc/whatsnew/3.11.rst#math)

- https://github.com/python/cpython/pull/26606


youknowone (approved):
nice catch!

PR #4154, opened 2022-09-09
Add `datetime.UTC` alias for `datetime.timezone.utc`

jopemachine:
Added `UTC` alias added in cpython 3.11

By the way, It seems maybe the `Lib/test/datetimetester.py` tests are missing, I'm wondering if this is intended.

## Manual test 

```
>>>>> datetime.UTC
datetime.timezone.utc
```

## Related links

* Python codes are copy-pasted from CPython.

- https://github.com/python/cpython/pull/91973

- [Python change log](https://github.com/python/cpython/blob/main/Doc/whatsnew/3.11.rst#datetime)

### DIFF
--- a/Lib/datetime.py
+++ b/Lib/datetime.py
@@ -5,7 +5,7 @@ time zone and DST data sources.
 """
 
 __all__ = ("date", "datetime", "time", "timedelta", "timezone", "tzinfo",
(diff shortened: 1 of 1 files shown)


youknowone (approved):
we usually don't partially update python library. but datetime looks hard to work on right now.

PR #4156, opened 2022-09-09
Update `functools` lib, tests and Add `singledispatch`

jopemachine:
- Update `functools` lib, tests
- Add `singledispatch` implemented in cpython 3.11

# Related links

- [Related PR](https://github.com/python/cpython/pull/30017)
- [Python documentation](https://docs.python.org/ko/3.11/library/functools.html#functools.singledispatch)


youknowone:
When you update python files from CPython, could you leave version message to commit log? It is helpful when we update the files to future version later. Otherwise we need to check file content to check its version.

Here is guideline: https://rustpython.github.io/blog/2020/04/04/how-to-contribute-by-cpython-unittest.html
Here is an example: https://github.com/RustPython/RustPython/pull/4025

jopemachine:
I'm waiting for this issue resolved before going further. https://github.com/RustPython/RustPython/issues/4157

youknowone:
we'd better to step on 3.11 before 3.12. 3.12 is very early version of next cpython.

youknowone:
functools is updated by #5078. threading is not.

PR #4159, opened 2022-09-13
Fix `bool_eq` argument order of `_membership_iter_search` for compatibility with cpython

jopemachine:
## Related

- [Related CPython implementation](https://github.com/python/cpython/blob/main/Objects/abstract.c#L2173-L2254)


youknowone (approved):
nice catch!

PR #4160, opened 2022-09-13
Implement `MemberDescrObject` setter

moreal:
This pull request implements `MemberDescrObject`'s setter.

---

I'll implement the setter macro in the next PR with `Frame.f_trace_lines` member.

- Support `T_BOOL` member type.
- Implement `Frame.f_trace_lines` member.
- Implement `#[pymember(setter)` macro.

### DIFF
--- a/vm/src/object/core.rs
+++ b/vm/src/object/core.rs
@@ -25,6 +25,7 @@ use crate::{
     builtins::{PyDictRef, PyTypeRef},
     vm::VirtualMachine,
 };
+use itertools::Itertools;
 use std::{
     any::TypeId,
     borrow::Borrow,
@@ -115,7 +116,7 @@ struct PyInner<T> {
     weak_list: WeakRefList,
 
     payload: T,
-    slots: Vec<Option<PyObjectRef>>,
+    slots: Box<[PyRwLock<Option<PyObjectRef>>]>,
 }
 
 impl<T: fmt::Debug> fmt::Debug for PyInner<T> {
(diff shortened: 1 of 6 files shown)


vm/src/object/core.rs
```diff
@@ -115,7 +115,7 @@ struct PyInner<T> {
     weak_list: WeakRefList,
 
     payload: T,
-    slots: Vec<Option<PyObjectRef>>,
+    slots: PyRwLock<Vec<Option<PyObjectRef>>>,
```
youknowone, line 118: Are slots editable after initialization? Maybe we only need `Box<[Option<PyObjectRef>]>`. I am not sure this is related to the problem.

moreal: Sorry, the above comment is incorrect. I made the `slots`' type as `Box<[PyRwLock<Option<PyObjectRef>>]>` and I saw the situation the slots' length is too big though it was correct when creating the slice. And I tried `Vec<T>` also but it also does. I researched more and I guess it's from 

In `_HashedSeq` case, it has `slots: vec{len: 1, cap: 1}`. But after `Into::into` (`From<PyRef<T>> for PyObjectRef`) it became to have `slots: vec{len: 0, cap: 8}` (random). And in the `PyObject.set_slot`, it becomes to have `slots: vec{len: 4, cap: <SOME_BIG_INT e.g. 4396189216>}`. 😭

moreal: When I tried to see the memory view with vscode codelldb extension, that's same with the screenshot.
image

moreal: It was a problem related to struct fields order (`repr(C)`). I resolved in [f140794](https://github.com/RustPython/RustPython/pull/4160/commits/f1407944c6edd29c33141ba5485e446c674a6e21)

youknowone (approved):

youknowone:
it looks great, thank you!

youknowone (approved):

PR #4161, opened 2022-09-14
Fix `hash`'s error message when `hash` method is `None`

jopemachine:
## CPython (Expected result)

```
>>> class NoHash(object):
...   __hash__ = None
... 
>>> hash(NoHash())
TypeError: unhashable type: 'NoHash'
```

## RustPython (Actual result)

```
>>>>> class NoHash(object):
.....   __hash__ = None
..... 
>>>>> hash(NoHash())
TypeError: 'NoneType' object is not callable
```

## Result

This PR fixes RustPython's error message when `hash` method is `None`

```
>>>>> class NoHash(object):
.....   __hash__ = None
..... 
>>>>> hash(None)
TypeError: unhashable type: 'NoHash'
>>>>>

### DIFF
--- a/vm/src/protocol/object.rs
+++ b/vm/src/protocol/object.rs
@@ -514,9 +514,8 @@ impl PyObject {
     }
 
     pub fn hash(&self, vm: &VirtualMachine) -> PyResult<PyHash> {
-        // hash always exist
         let hash = self.get_class_attr(identifier!(vm, __hash__)).unwrap();
-        if hash.is(&vm.ctx.none()) {
+        if vm.is_none(&hash) {
             return Err(vm.new_exception_msg(
                 vm.ctx.exceptions.type_error.to_owned(),
                 format!("unhashable type: '{}'", self.class().name()),
--- a/vm/src/protocol/object.rs
+++ b/vm/src/protocol/object.rs
@@ -514,10 +514,20 @@ impl PyObject {
     }
 
     pub fn hash(&self, vm: &VirtualMachine) -> PyResult<PyHash> {
+        // hash always exist
+        let hash = self.get_class_attr(identifier!(vm, __hash__)).unwrap();
+        if hash.is(&vm.ctx.none()) {
+            return Err(vm.new_exception_msg(
+                vm.ctx.exceptions.type_error.to_owned(),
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: vm/src/protocol/object.rs [hunk 1]
```diff
@@ -514,10 +514,20 @@ impl PyObject {
     }
 
     pub fn hash(&self, vm: &VirtualMachine) -> PyResult<PyHash> {
+        // hash always exist
+        let hash = self.get_class_attr(identifier!(vm, __hash__)).unwrap();
```
youknowone, line 518: This line triggers python code and will increase runtime cost.
When hash is None, `cls.slots.hash` is pointing a default implementation raising error. Comparing pointers will be able to check the same thing.

jopemachine: > This line triggers python code and will increase runtime cost. When hash is None, `cls.slots.hash` is pointing a default implementation raising error. Comparing pointers will be able to check the same thing.

May I ask how can I acquire the pointer pointing to the default error implementation?
I'm looking into the codes, but I have trouble finding the pointer.

youknowone: Oh, I found we cannot obtain function pointer of default implementation.

We don't seem to have a single pointer of `Unhashable::slot_hash`. Do you have a good idea?

Snowapril: I think we need to check update_slot and hash_wrapper. If given one is None (singleton object), then use Unhashable::slot_hash (https://github.com/python/cpython/blob/main/Objects/typeobject.c#L8826-L8935)

vm/src/protocol/object.rs
```diff
@@ -514,10 +514,20 @@ impl PyObject {
...
+        if hash.is(&vm.ctx.none()) {
```
youknowone, line 519:
```suggestion
        if vm.is_none(hash) {
```

vm/src/protocol/object.rs
youknowone: I am sorry, this comment was wrong.

youknowone (approved):
I am really sorry. I misunderstood this is related to the fast path, but this was related to the python path.
Your approach is reasonable enough for current situation.

youknowone (approved):

PR #4165, opened 2022-09-16
Fix `property.__set_name__` error message when arguments are invalid

jopemachine:


vm/src/builtins/property.rs
```diff
@@ -154,6 +155,23 @@ impl PyProperty {
...
+    fn set_name(&self, args: PosArgs, vm: &VirtualMachine) -> PyResult<()> {
+        let arg_len = args.into_vec().len();
+
+        if arg_len != 2 {
+            Err(vm.new_exception_msg(
+                vm.ctx.exceptions.type_error.to_owned(),
+                format!(
+                    "__set_name__() takes 2 positional arguments but {} were given",
+                    arg_len
+                ),
```
youknowone, line 168: This message is common python type error which we are not handling well. We'd better to find enhance the general argument error handling instead of this way.

youknowone: I am sorry, the given type is `PosArgs`. So no way to do it in shared code.

youknowone:
@jopemachine is it still a draft or done?

DimitrisJim:
this seems like its good to go, no?

youknowone (approved):

PR #4170, opened 2022-09-18
pyproperty -> pygetset

youknowone:
Though it actually create a getset descriptor, we didn't change the name because its concept was a property for users. Now we have `pymember`, another property-like descriptor. So naming them under same level would be less confusing instead of telling everybody "it is a getset but we call it pyproperty"

cc @moreal


moreal (approved):
I checked there are only changes about `pyproperty` -> `pygetset` by replacing `pygetset` -> `pyproperty` and checking with `git diff main`.

PR #4172, opened 2022-09-19
ci: Support for ci skip through tag

Yaminyam:
You can skip unnecessary ci actions through the `skip:ci` tag.
The default pull_request action trigger does not include labeled or unlabeled, so when the label is modified, the trigger is modified to reactivate the action.


youknowone (approved):
Looks working great! Thank you!

Issue #4166, opened 2022-09-16
python.rs file is not generated locally


lastmjs:
If I want to fork RustPython and access it locally in a Rust project, it doesn't currently work because `RustPython/compiler/parser/python.rs` is not generated. I haven't been able to figure out how to generate it, so I just copied the file from my `.cargo` folder.

fanninpm:
@youknowone does this have to do with LALRPOP not being installed?

fanninpm:
@lastmjs Try `cargo clean` followed by `cargo build`. Does that work?

youknowone:
@fanninpm that's not intended

lastmjs:
I would expect it to just work after I've cargo installed RustPython

lastmjs:
I think my issue might be a duplicate of this: https://github.com/RustPython/RustPython/issues/4117

lastmjs:
I also have all features turned off, should I have `rustpython-parser/lalrpop` as a feature?

lastmjs:
I believe this has been resolved with https://github.com/RustPython/RustPython/pull/4176

Issue #4117, opened 2022-08-21
Unable to read python.rs when installing RustPython from source


charliermarsh:
# Summary

In my Cargo.toml, I have:

```
[dependencies]
rustpython-parser = { git = "https://github.com/RustPython/RustPython.git", rev = "f1c7aebfad707526602c8162c8094a67129b6731" }
```

Now, running `cargo build` on my project yields:

```
∴ cargo build
   Compiling rustpython-parser v0.1.2 (https://github.com/RustPython/RustPython.git?rev=f1c7aebfad707526602c8162c8094a67129b6731#f1c7aebf)
error: couldn't read /Users/crmarsh/.cargo/git/checkouts/rustpython-f8ef4d934ac33cd8/f1c7aeb/parser/src/../python.rs: No such file or directory (os error 2)
 --> /Users/crmarsh/.cargo/git/checkouts/rustpython-f8ef4d934ac33cd8/f1c7aeb/parser/src/python.rs:3:1
  |
3 | include!("../python.rs");
  | ^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: this error originates in the macro `include` (in Nightly builds, run with -Z macro-backtrace for more info)

The following warnings were emitted during compilation:

warning: python.rs doesn't exist. regenerate.

error: could not compile `rustpython-parser` due to previous error
```

Is this a user error? I know this is a recent change as my project built without error on `dff916d45c5d13074d21ad329a5ab68a6499426a`.

youknowone:
I am sorry, this is my fault. I will fix it soon.

youknowone:
I want to make it easier, not it will take time. Could you try this for now?

```
[dependencies]
rustpython-parser = { features=["lalrpop"], git = "https://github.com/RustPython/RustPython.git", rev = "f1c7aebfad707526602c8162c8094a67129b6731" }
```

charliermarsh:
That worked! Thank you.

charliermarsh:
(Feel free to close if you'd like.)

ecpost:
As of about 2 days ago, I'm also having problems compiling rustpython-parser / python.rs.  If I do a fresh `git clone https://github.com/RustPython/RustPython.git` and in it do `cargo build --release`, after compiling most of the crates I eventually get:

```
The following warnings were emitted during compilation:

warning: python.rs doesn't exist. regenerate.

[...]

PR #4176, opened 2022-09-21
temporary fix of lalrpop build

youknowone:
maybe related to #4166 and #4117


lastmjs:
I believe this has resolved https://github.com/RustPython/RustPython/issues/4166, all of our tests are passing with our fork of RustPython which is on the latest RustPython commit.

PR #4177, opened 2022-09-21
Add `co_names` to `PyCode`

jopemachine:
# Outline

Add `co_names` to `PyCode`

CPython document says `co_names` is a *tuple of names other than arguments and function locals*

## Reference

https://github.com/python/cpython/blob/main/Doc/library/inspect.rst#types-and-members

## Result

```
>>>>> a = 3
>>>>> b = 5
>>>>> c = 7
>>>>> def f(k, l, m):
.....   d = 9
.....   return a + k
..... 
>>>>> f.__code__.co_names
1
('a',)
>>>>> 
```

### DIFF
--- a/Lib/opcode.py
+++ b/Lib/opcode.py
@@ -4,9 +4,9 @@ opcode module - potentially shared between dis and other modules which
 operate on bytecodes (e.g. peephole optimizers).
 """
 
-__all__ = ["cmp_op", "hasconst", "hasname", "hasjrel", "hasjabs",
-           "haslocal", "hascompare", "hasfree", "opname", "opmap",
-           "HAVE_ARGUMENT", "EXTENDED_ARG", "hasnargs"]
+__all__ = ["cmp_op", "hasarg", "hasconst", "hasname", "hasjrel", "hasjabs",
+           "haslocal", "hascompare", "hasfree", "hasexc", "opname", "opmap",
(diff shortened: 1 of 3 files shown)


youknowone:
if there is no related unittest in `Lib/tests`, could you add the test you used in this issue to somewhere in `extra_tests/snippets`?

Snowapril:
Oh, It seems we missed `Lib/code.py` from cpython. We'd better use it
https://github.com/python/cpython/blob/main/Lib/code.py

youknowone:
@Snowapril is it related to this changes?

Snowapril:
I'm sorry. I uploaded a different link 😂 Please take a look again @jopemachine

jopemachine:
I updated the `test_code.py`, but there are some points that I'd like to mention.

* Updated `opcode.py` for updating `test_code.py`. but I didn't update test_opcode.py because test code is not changed.
* There are some `dump` function calls in the above comment of the `test_code.py`. Because this raised test failure, I commented it out and added `TODO: RUSTPYTHON` comment.
* I checked the added `co_names` manually in the dump call, the first dump call's actual result is different from the expected result. The actual result is `('__doc__')` while the expected result is empty tuple, the remaining tests result are correct.

youknowone (approved):
Great! thank you!

Issue #4158, opened 2022-09-13
[RFC] GC: Both Stop-the-World and On-the-fly


discord9:
## Summary
I propose we implement a cycle collector for our rustPython RefCount garbage collection system based on Bacon etc.'s work in this two paper:
- [Concurrent Cycle Collection in Reference Counted Systems](http://link.springer.com/10.1007/3-540-45337-7_12), this paper come up with a sync(stop the world) cycle collector and a concurrent version of it, although I suggest only impl the sync version of it, for concurrent version is harder to implement and still limited in preformance which is not as good as the second paper we are introducing
- [An efficient on-the-fly cycle collection](https://dl.acm.org/doi/10.1145/1255450.1255453), this paper based on the sync cycle collector in previous paper, adding a "sliding view" algorithm to it, which is like a "copy on write" for Object so during tracing the GC can see a fixed version of object dependence graph while mutator can still freely mutate object whatever they want.

## Detailed Explanation
The synchronous cycle collector is easier to impl, I suggest we implement a synchronous verison of cycle collector for proof of Concept, then modfiy it to a On-the-Fly version.
### Synchronous Cycle Collection
The algo in "Concurrent Cycle Collection in Reference Counted Systems" is O(N + E) wrost-case time for collection at most.

For an object T in its' header there need fields denoted buffered(T), color(T),and RC(T). Where:
- buffer is used to place potential roots of cyclic garbage, buffered(T) denote whether a object is in the buffer.
- color is explain in this table(Red and Orange is only used in concurrent version):

| Color       | Meaning                     |
| ----------- | -----------                 |
| Black       | In use or free              |
| Gray        | Possible member of cycle    |
| White       | Member of garbage cycle     |
| Purple      | Possible root of cycle      |
| Green       | Acyclic                     |
| Red         | Candidate cycle undergoing $\Sigma$-computation |
| Orange      | Candidate cycle awaiting epoch boundary |

- RC is just a reference count(which doesn't even need to be atomic for a sync version).
#### Pseudocode and Explanation
In this section I just ref to paper because it explain much better than me, the basic idea is to first scan for possible cycles, then check for actual cycles:
> Our synchronous algorithm is similar to Lins’ algorithm: when reference counts are decremented, we place potential roots of cyclic garbage into abuffer called Roots. Periodically,we process this buffer and look for cycles by subtracting internal reference counts. 
> 
[...]

PR #4180, opened 2022-09-22
(READY FOR REVIEW)Garbage collect: A stop-the-world cycle collector

discord9:
That might be extend to be a on-the-fly version. Based on #4158
# Summary of GC Algorithm
the algorithm store a buffer of object thats' "possible roots" of garbage cycles(i.e. if a object is decrement but not to zero, its a possible root) then, during gc, algorithm have three main phase:
1. iterate over this buffer, free objects of rc==0.
2. then in next phase tracing other object(whose rc>0) and do a "trial decrement" to all of its childrens, so to remove all the "internal reference" in the graph, so whats left is external reference, so by now,object with rc==0 is safe to be collected
3. restore everyone elses rc, only drop garbage found from previous phase, note in here we first do a drop in place for every garbage then dealloc everything to prevent UB from deref to deallocated objects

# Progress

- [x] basic cycle collector algorithm(stop the world version)
- [x] necessary trait and cond comp flag and feature
- [x] impl GcTrace for all builtin types(Some type's trace() still need tweak, see below, still have some not impl)
- [x] rebase off the newest main(for rustc 1.64.0 update)
- [x] export `gc()` as a python function(in the stdlib), now invoke `gc()` in python will force a cycle collect to happen, return number of cyclic garbage collected(and freed)
- [x] test(especially for threading)
- [x] bug: I use wrong drop in collect_white, should drop `PyInner<T>`, but I drop `PyInner<Erased>`, need correct that.(Fixed)
- [x] bug: PyInner is deallocated early before PyWeak release
- [x] bug: gc in drop() sometimes seems to cause deadlock(Fix by putting `gc()` to instruction exec loop)
- [x] use `PyObject` directly instead of dyn dispatch in gc
- [x] move `gc()` to its' own module in stdlib
- [x] find out why `stdlib::io::_io::BufferedReader`(and other types in _io module) keeps double free
- [x] rewrite commits 
- [x] drop `header` later to prevent UB
- [x] run `__del__`, drop, dealloc separately to prevent UB
- [x] change back to use PyMutex instead of PyRwLock(which is unacceptly slow)
- [x] use proc macro to impl trace for datatype
- [x] compact `header`'s most field into one byte
- [x] bug with weak ref

## future features for GC
-  use [MemBalancer](http://arxiv.org/abs/2204.10455) to determine when to gc(actually tried on a new branch, but I am having difficulty sampling memory usage&memory deallocate rate so not having a very good effect)
-  found out how to make optional trait bound without using typeid&macro(which make it hard to extend)

## GC condition
gc will be triggered if the internal buffer(`roots`) have exceed 10007(Just a random reasonably large number) element and at least 100ms after last gc, that means after 10007 different objects get decref after last gc, a new gc will be trigger, `gc()` is only called in the loop of executing instruction, but also reexport a `gc()` to stdlib so one can call `gc()` from Python side(which will force a gc regardless of above conditon).
stop-the-world is impl by try to blocking on a pausing `Mutex` owned by gc whenever try to deref PyObjectRef/PyRef<T>/Py<T>
# Basic ideas
cycle collector algorithm:
1. stop the world, and remove internal reference
2. found out those with zero ref(meaning they are cyclic garbage), mark them for later free, restore others' ref count.
3. free those marked object(also unpause the world), also for those marked object form a cycle, some check is done to make sure no double drop or deallocate happen
# Data struct
```rust
#[repr(C)]
struct PyInner<T> {
    #[cfg(not(feature = "gc"))]
    ref_count: RefCount,
    #[cfg(feature = "gc")]
    header: GcHeader,
    ....
```
```rust
/// Garbage collect header, containing ref count and other info, using repr(C) to stay consistent with PyInner 's repr
#[repr(C)]
pub struct GcHeader {
    ref_cnt: PyAtomic<usize>,
    color: PyMutex<Color>,
    buffered: PyMutex<bool>,
    pub exclusive: PyMutex<()>,
    pub gc: PyRc<CcSync>
}
```

# Example
for code as below:
```python
tot = 0
runs = 16000
for i in range(runs):
    a = [0]
    b = [a]
    a[0] = b
    del a
    del b
    a = {}
    b = {}
    a[0] = b
    b[0] = a
    del a
    del b
```
With and without gc, the memory usage is(run on a win10(19044.2006) with rustc 1.64.0, might vary between different runs):
| loop | mem with gc(MB) | without gc(MB)|
| --- | --- | --- |
| 0 |   13.29 | 8.82    |
| 1000  | 13.49 | 10.36 |
| 2000  | 13.76 | 11.76 |
| 3000  | 14.29 | 13.32 |
| 4000  | 14.29 | 14.45 |
| 5000  | 14.31 | 16.32 |
| 6000  | 14.31 | 17.24 |
| 7000  | 14.31 | 18.32 |
| 8000  | 14.31 | 19.45 |
| 9000  | 15.36 | 21.38 |
| 10000 | 15.36 | 22.45 |
| 11000 | 15.26 | 23.71 |
| 12000 | 15.26 | 24.63 |
| 13000 | 15.26 | 25.97 |
| 14000 | 15.26 | 27.36 |
| 15000 | 15.26 | 28.36 |

# currently `GcTrace`d types are:
```rust
// builtin types
// PyRange, PyStr is acyclic
PyBoundMethod,
PyDict,
PyEnumerate,
PyFilter,
PyFunction,
PyList,
PyMappingProxy,
PyProperty,
PySet,
PySlice,
PyStaticMethod,
PySuper,
PyTraceback,
PyTuple,
// PyType, (strange bug, see PyType's trace() for detail)
PyZip,
// misc
PyCell,
// iter in iter.rs
PySequenceIterator,
PyCallableIterator,
// iter on types
// PyList's iter
PyListIterator,
PyListReverseIterator,
// PyTuple's iter
PyTupleIterator,
// PyEnumerate's iter
PyReverseSequenceIterator,
// PyMemory's iter
PyMemoryViewIterator,
// function/Arg protocol
ArgCallable,
ArgIterable,
ArgMapping,
ArgSequence,
// protocol
// struct like
PyBuffer,
PyIter,
// FIXME(discord9): confirm this is ok to do
PyIterIter<T>,
PyIterReturn,
PyMapping,
PyNumber,
PySequence
```

### DIFF
--- a/Lib/test/test_ordered_dict.py
+++ b/Lib/test/test_ordered_dict.py
@@ -656,8 +656,6 @@ class OrderedDictTests:
         dict.update(od, [('spam', 1)])
         self.assertNotIn('NULL', repr(od))
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_reference_loop(self):
         # Issue 25935
         OrderedDict = self.OrderedDict
@@ -667,6 +665,8 @@ class OrderedDictTests:
         r = weakref.ref(A)
         del A
         gc.collect()
+        # TODO: RustPython, Need to fix this: somehow after del A, it takes two call to `gc.collect()`
+        # for gc to realize a loop is there and to be collected
         self.assertIsNone(r())
 
     # TODO: RUSTPYTHON
--- a/Lib/test/test_ordered_dict.py
+++ b/Lib/test/test_ordered_dict.py
@@ -665,8 +665,6 @@ class OrderedDictTests:
         r = weakref.ref(A)
         del A
         gc.collect()
-        # TODO: RUSTPYTHON, Need to fix this: somehow after del A, it takes two call to `gc.collect()`
-        # for gc to realize a loop is there and to be collected
         self.assertIsNone(r())
 
     # TODO: RUSTPYTHON
--- a/Lib/test/test_ordered_dict.py
+++ b/Lib/test/test_ordered_dict.py
@@ -665,7 +665,7 @@ class OrderedDictTests:
         r = weakref.ref(A)
         del A
         gc.collect()
-        # TODO: RustPython, Need to fix this: somehow after del A, it takes two call to `gc.collect()`
+        # TODO: RUSTPYTHON, Need to fix this: somehow after del A, it takes two call to `gc.collect()`
         # for gc to realize a loop is there and to be collected
         self.assertIsNone(r())
 
--- a/Lib/test/test_weakref.py
+++ b/Lib/test/test_weakref.py
@@ -1912,12 +1912,7 @@ class MappingTestCase(TestBase):
     def test_threaded_weak_valued_pop(self):
         d = weakref.WeakValueDictionary()
         with collect_in_thread():
-            print("")
             for i in range(100000):
-                if i%1000==0:
-                    print("\rLoop:"+str(i)+"/100000     ", end="")
-                    # TODO: RUSTPYTHON: so in log file the progress can be update in time
-                    sys.stdout.flush()
                 d[10] = RefCycle()
                 x = d.pop(10, 10)
                 self.assertIsNot(x, None)  # we never put None in there!
@@ -1927,12 +1922,7 @@ class MappingTestCase(TestBase):
         # WeakValueDictionary when collecting from another thread.
         d = weakref.WeakValueDictionary()
         with collect_in_thread():
-            print("")
             for i in range(200000):
-                if i%1000==0:
-                    print("\rLoop:"+str(i)+"/200000     ", end="")
-                    # TODO: RUSTPYTHON: so in log file the progress can be update in time
-                    sys.stdout.flush()
                 o = RefCycle()
                 d[10] = o
                 # o is still alive, so the dict can't be empty
--- a/Lib/test/test_weakref.py
+++ b/Lib/test/test_weakref.py
@@ -75,9 +75,10 @@ class TestBase(unittest.TestCase):
     def callback(self, ref):
         self.cbcalled += 1
 
-
+# TODO: RUSTPYTHON, cpython's period is 0.0001, but at least now using such a small gc period is too slow
+# so change to 0.001 for now
 @contextlib.contextmanager
-def collect_in_thread(period=0.0001):
+def collect_in_thread(period=0.001):
     """
     Ensure GC collections happen in a different thread, at a high frequency.
     """
@@ -1911,7 +1912,12 @@ class MappingTestCase(TestBase):
     def test_threaded_weak_valued_pop(self):
         d = weakref.WeakValueDictionary()
         with collect_in_thread():
+            print("")
             for i in range(100000):
+                if i%1000==0:
+                    print("\rLoop:"+str(i)+"/100000     ", end="")
+                    # TODO: RUSTPYTHON: so in log file the progress can be update in time
+                    sys.stdout.flush()
                 d[10] = RefCycle()
                 x = d.pop(10, 10)
                 self.assertIsNot(x, None)  # we never put None in there!
@@ -1921,7 +1927,12 @@ class MappingTestCase(TestBase):
         # WeakValueDictionary when collecting from another thread.
         d = weakref.WeakValueDictionary()
         with collect_in_thread():
+            print("")
             for i in range(200000):
+                if i%1000==0:
+                    print("\rLoop:"+str(i)+"/200000     ", end="")
+                    # TODO: RUSTPYTHON: so in log file the progress can be update in time
+                    sys.stdout.flush()
                 o = RefCycle()
                 d[10] = o
                 # o is still alive, so the dict can't be empty
--- a/derive/src/lib.rs
+++ b/derive/src/lib.rs
@@ -91,3 +91,20 @@ pub fn pypayload(input: TokenStream) -> TokenStream {
     let input = parse_macro_input!(input);
     derive_impl::pypayload(input).into()
 }
+
+/// use on struct with named fields like `struct A{x:i32, y:i32}` to impl `Trace` for datatype
+///
+/// use `#[notrace]` on fields you wish not to trace
+///
+/// add `trace` attr to `#[pyclass]` to make it
+/// traceable(Even from type-erased PyObject)(i.e. write `#[pyclass(trace)]`)
+/// better to place after `#[pyclass]` so pyclass know `pytrace`'s existance and impl a MaybeTrace calling Trace
+#[proc_macro_attribute]
+pub fn pytrace(
+    attr: proc_macro::TokenStream,
+    item: proc_macro::TokenStream,
+) -> proc_macro::TokenStream {
+    let attr = parse_macro_input!(attr);
+    let item = parse_macro_input!(item);
+    derive_impl::pytrace(attr, item).into()
+}
--- a/stdlib/src/gc.rs
+++ b/stdlib/src/gc.rs
@@ -8,7 +8,7 @@ mod gc {
     fn collect(_args: FuncArgs, _vm: &VirtualMachine) -> i32 {
         #[cfg(feature = "gc_bacon")]
         {
-            usize::from(rustpython_vm::object::collect()) as i32
+            usize::from(rustpython_vm::object::gc::collect()) as i32
         }
         #[cfg(not(feature = "gc_bacon"))]
         {
@@ -20,7 +20,7 @@ mod gc {
     fn isenabled(_args: FuncArgs, _vm: &VirtualMachine) -> bool {
         #[cfg(feature = "gc_bacon")]
         {
-            rustpython_vm::object::isenabled()
+            rustpython_vm::object::gc::isenabled()
         }
         #[cfg(not(feature = "gc_bacon"))]
         {
@@ -32,7 +32,7 @@ mod gc {
     fn enable(_args: FuncArgs, vm: &VirtualMachine) -> PyResult {
         #[cfg(feature = "gc_bacon")]
         {
-            rustpython_vm::object::enable();
+            rustpython_vm::object::gc::enable();
             Ok(vm.new_pyobj(true))
         }
         #[cfg(not(feature = "gc_bacon"))]
@@ -45,7 +45,7 @@ mod gc {
     fn disable(_args: FuncArgs, vm: &VirtualMachine) -> PyResult {
         #[cfg(feature = "gc_bacon")]
         {
-            rustpython_vm::object::disable();
+            rustpython_vm::object::gc::disable();
             Ok(vm.new_pyobj(true))
         }
         #[cfg(not(feature = "gc_bacon"))]
--- a/stdlib/src/gc.rs
+++ b/stdlib/src/gc.rs
@@ -6,22 +6,52 @@ mod gc {
 
     #[pyfunction]
     fn collect(_args: FuncArgs, _vm: &VirtualMachine) -> i32 {
-        0
+        #[cfg(feature = "gc_bacon")]
+        {
+            usize::from(rustpython_vm::object::collect()) as i32
+        }
+        #[cfg(not(feature = "gc_bacon"))]
+        {
+            0
+        }
     }
 
     #[pyfunction]
     fn isenabled(_args: FuncArgs, _vm: &VirtualMachine) -> bool {
-        false
+        #[cfg(feature = "gc_bacon")]
+        {
+            rustpython_vm::object::isenabled()
+        }
+        #[cfg(not(feature = "gc_bacon"))]
+        {
+            false
+        }
     }
 
     #[pyfunction]
     fn enable(_args: FuncArgs, vm: &VirtualMachine) -> PyResult {
-        Err(vm.new_not_implemented_error("".to_owned()))
+        #[cfg(feature = "gc_bacon")]
+        {
+            rustpython_vm::object::enable();
+            Ok(vm.new_pyobj(true))
+        }
+        #[cfg(not(feature = "gc_bacon"))]
+        {
+            Err(vm.new_not_implemented_error("".to_owned()))
+        }
     }
 
     #[pyfunction]
     fn disable(_args: FuncArgs, vm: &VirtualMachine) -> PyResult {
-        Err(vm.new_not_implemented_error("".to_owned()))
+        #[cfg(feature = "gc_bacon")]
+        {
+            rustpython_vm::object::disable();
+            Ok(vm.new_pyobj(true))
+        }
+        #[cfg(not(feature = "gc_bacon"))]
+        {
+            Err(vm.new_not_implemented_error("".to_owned()))
+        }
     }
 
     #[pyfunction]
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -349,15 +349,16 @@ impl ExecutingFrame<'_> {
         // Execute until return or exception:
         let instrs = &self.code.instructions;
         let mut arg_state = bytecode::OpArgState::default();
+        #[allow(dead_code)]
         let mut gc_count = 0;
         loop {
-            gc_count += 1;
-            if gc_count > 1000 {
-                #[cfg(feature = "gc_bacon")]
-                {
+            #[cfg(feature = "gc_bacon")]
+            {
+                gc_count += 1;
+                if gc_count > 1000 {
                     crate::object::try_gc();
+                    gc_count = 0;
                 }
-                gc_count = 0;
             }
             let idx = self.lasti() as usize;
             self.update_lasti(|i| *i += 1);
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -349,7 +349,16 @@ impl ExecutingFrame<'_> {
         // Execute until return or exception:
         let instrs = &self.code.instructions;
         let mut arg_state = bytecode::OpArgState::default();
+        let mut gc_count = 0;
         loop {
+            gc_count += 1;
+            if gc_count > 1000 {
+                #[cfg(feature = "gc_bacon")]
+                {
+                    crate::object::try_gc();
+                }
+                gc_count = 0;
+            }
             let idx = self.lasti() as usize;
             self.update_lasti(|i| *i += 1);
             let bytecode::CodeUnit { op, arg } = instrs[idx];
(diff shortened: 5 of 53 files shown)


discord9:
A simple example of collecting cycle ref of PyList(note the line of collect cyclic garbage composed of two object(`a` and `b`):
```
a= [0]
b = [a]
a[0] = b
```
gc

fanninpm:
Rust 1.64 came out a few hours ago, so there are some new compilation warnings and Clippy lints to deal with in a separate PR (#4182). Once that PR is merged, it might be a good idea to rebase this branch on top of a fresh copy of `main`.

vm/src/object/gc/trace.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// 
+    /// if some field is not called, the worse results is memory leak, but if some field is called repeatly, panic and deadlock can happen.
+    /// 
+    /// Note that Two `PyObjectRef` to the Same `PyObject` still count as two Ref, and should be called twice(once for each one) in this case.
+    ///
+    /// ```
+    /// for ch in childs:
+    ///     tracer_fn(ch)
+    /// ```
+    /// *DO NOT* clone a `PyObjectRef`, use `as_ptr()` instead and operate on NonNull
```
fanninpm, line 42: Is there any reason for `PyObjectRef` to be `Clone`?

discord9: clone PyObjectRef in trace() will mess up ref cnt cycle collector, but it is very tempting to do so, like in PySet.keys() give vec of PyObjectRef(which invoke clone())

> Is there any reason for `PyObjectRef` to be `Clone`?

fanninpm: In `vm/src/object/core.rs`, have you tried deleting `PyObjectRef`'s implementation of the `Clone` trait (or gating it behind the _absence_ of the `gc` feature)? If so, what consequences are there?

https://github.com/RustPython/RustPython/blob/c8c9c7b5f9815b376c584177b6b91a779026e4c3/vm/src/object/core.rs#L610-L615

discord9: > In `vm/src/object/core.rs`, have you tried deleting `PyObjectRef`'s implementation of the `Clone` trait (or gating it behind the _absence_ of the `gc` feature)? If so, what consequences are there?
> 
> https://github.com/RustPython/RustPython/blob/c8c9c7b5f9815b376c584177b6b91a779026e4c3/vm/src/object/core.rs#L610-L615

unfortunatly, I have to rely on `PyObjectRef`'s clone() in everywhere else(except in trace()) to make the ref count system work properly, so no, I can't delete Clone for PyObjectRef, but I think we can make trace unsafe so people take those code seriously?

discord9:
Just realize I still need make some change to make gc adopt to threading, need to think about it

discord9:
bug: might need to store the number of PyWeak ref count to prevent premature dealloc in heap and access violation. Also PyWeak itself is storing a PyObjectRef which keep the object from drop, some change need to be made to change that

discord9:
sorry for the messy commit logs, will reword them later

fanninpm:
When happens when you compile this with the `--no-default-features` flag? I suspect there may be a few compilation errors with `stdlib/src/gc.rs`.

discord9:
> 

Yes, just realize that, gonna fix it, also still debugging on the dead lock&double dealloc issue, my code is still very faulty

discord9:
Just found out I can't trace on `PyFunction`'s `closure` field(which is of type `PyTupleTyped`), when tracing closure, during object graph traverse, it always call trace on closure's PyCell second time(which shouldn't be, because traverse should mark a visited object and not call trace(therefore deadlock on it) (and not trace closure seems to fix that, is there something special about `PyFunction`'s closure field

youknowone:
@coolreader18 do you have advices?

discord9:
Might found a bug related to the old ref count system: https://github.com/RustPython/RustPython/issues/4238 ,still trying to find a fix

discord9:
I am gonna to close this PR and break it into two new PR:
- first pr that add `Trace` trait on all Py*
- second pr that add main gc algorithm
any advices?

discord9:
Using PyRwLock is indeed very slow, might need to found out a way to trace PyMutex properly

DimitrisJim:
@discord9 ping me whenever you might need the workflow executed.

discord9:
I think current version can pass most test for now...except for the following:
related to WeakRef, seems to access deallocated object in some rare case(one time in several million times):
- test_threaded_weak_valued_pop
- test_threaded_weak_valued_consistency

confusing error, seems to related with Frame:
- test_sys_setprofile

```
  File "pylib/Lib/test/test_sys_setprofile.py", line 72, in trace_return
    self.stack.pop()
IndexError: pop from empty list
```

test_enum's `test_unique_composite` can have inconsistent results when run test in local sometime(I'm using a wsl ubuntu 20.04 on win10)
@DimitrisJim Though I think the CI test can be run now?

discord9:
@DimitrisJim Sorry but it seems this time test_weakref trigger a panic on gc thread, and cause the entire program to halt(presumably poisoned) could you please help me terminate this workflow?

discord9:
@DimitrisJim I think the weakref bug should be fixed by now,and this version should be ready for CI test, although the `test_weakref` alone can take more than fifty minutes to complete because of high freq GC

discord9:
oh once again I forgot about conditional compilation flag, hey but at least the major cpython test are oked!

DimitrisJim:
>I wonder if we can lower the gc freq in test_weakref?

In general we do have instances where we've tweak the test code due to some implementation detail. We generally add a `TODO: RustPython` along with a reason for doing so. We'll get to adding that when the PR is getting reviewed.

discord9:
Also, any idead what's wrong with this wasm target? I found the error message with `cargo check --target wasm32-unknown-unknown --no-default-features` to be rather confusing

discord9:
The wasm test can't pass, I guess rebase off the newest main branch might solve it?

DimitrisJim:
Congrats on the tests passing!

So, I guess the biggest hurdle now is figuring out how to make it *not* stop the world this much (runners are taking around an hour to complete). I'll try to allocate some time each day to get up to speed but we'll definitely need reviews from @youknowone or @coolreader18 too to get it through.

discord9:
> 

My guess is that the main overhead is checking if need to pause(is checked on every deref), going to test it out

Lib/test/test_weakref.py
```diff
@@ -1919,6 +1919,7 @@ def test_threaded_weak_valued_pop(self):
             for i in range(100000):
                 if i%1000==0:
                     print("\rLoop:"+str(i)+"/100000     ", end="")
+                    sys.stdout.flush()
```
fanninpm, line 1922: Did you add these lines to the CPython test? If so, please add a `# TODO: RUSTPYTHON` annotation that explains your reasoning.

Note to self (and others interested): `Lib/test/support/threading_helper.py` and this file need updates to CPython 3.11 (there's a new `@threading_helper.requires_working_threading` annotation that tests for WASI and emscripten).

discord9: Added explanatory comment for using flush here

> Did you add these lines to the CPython test? If so, please add a `# TODO: RUSTPYTHON` annotation that explains your reasoning.
> 
> Note to self (and others interested): `Lib/test/support/threading_helper.py` and this file need updates to CPython 3.11 (there's a new `@threading_helper.requires_working_threading` annotation that tests for WASI and emscripten).

DimitrisJim:
I've slightly taken a peak at the paper this is based on, I'm guessing the two improvements it mentions have been implemented, right? I.e, using the buffered flag to not re-add it to the collector list and checking the current items in the collector as a single graph and not as distinct components.

discord9:
> I've slightly taken a peak at the paper this is based on, I'm guessing the two improvements it mentions have been implemented, right? I.e, using the buffered flag to not re-add it to the collector list and checking the current items in the collector as a single graph and not as distinct components.

Yes, this is the main improvement here, I write a buffered field for every object and there is a roots vec, actually the name of methods is corresponding between paper and the actual code

vm/src/protocol/iter.rs
```diff
@@ -14,6 +14,13 @@ pub struct PyIter<O = PyObjectRef>(O)
 where
     O: Borrow<PyObject>;
 
+#[cfg(feature = "gc")]
+unsafe impl<O: crate::object::Trace + Borrow<PyObject>> crate::object::Trace for PyIter<O> {
+    fn trace(&self, tracer_fn: &mut crate::object::TracerFn) {
+        self.0.trace(tracer_fn);
```
coolreader18, line 20: Shouldn't this just `self.0.borrow().trace()`? Shouldn't need the Trace bound I think.

coolreader18 (commented):
Thank you so much for working on this! I started this a week or 2 ago and then it slipped my mind after I took a break about halfway through looking at stuff. I haven't looked at the actualy gc code, and though I'd really like to, to be honest I'm not sure I'll have the bandwidth to do so. If no other maintainers feel able to either, worst case we mark it as a sort of experimental feature, but the test suite passing fine gives me confidence that it's working more or less as it should. It's not like there haven't already been multithreaded gc issues even without the tracing gc, lol.

vm/src/object/core.rs
```diff
@@ -121,6 +204,49 @@ struct PyInner<T> {
...
+            ($($TY: ty),*$(,)?) => {
+                $(
+                    if TypeId::of::<$TY>() == self.typeid{
+                        // Safety: because typeid said so!
+                        let inner: &PyInner<$TY> = unsafe { &*(self as *const PyInner<Erased> as *const PyInner<$TY>) };
+                        inner.payload.trace(tracer_fn);
+                    }
+                )else*
+            };
+        }
```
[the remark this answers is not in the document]
discord9: > Hmm... what if instead of this, `PyObjectPayload` required Trace (and stuck it in the PyObjVtable), but there's a DummyTrace trait that's safe and marks a object that isn't actually Trace. like:
> 
> ```rust
> unsafe trait Trace {
>     const IS_TRACE: bool = true;
>     fn trace(&self, ...);
> }
> trait DummyTrace {} // or NoTrace or something
> unsafe impl<T: DummyTrace> Trace for T {
>     const IS_TRACE: bool = false;
>     fn trace(&self, ...) {}
> }
> ```
> 
> And `#[pyclass]` could by default add a DummyTrace impl, unless specifically asked to pytrace it.
> 
> Oh and instead of `TraceHelper::is_traceable` the vtable has a `trace: Option<...>` that's None if IS_TRACE is false. Might be faster too, rather than checking 70 different TypeIds it's just one virtual function call.

Interesting idea, will try refactor into that

stdlib/src/gc.rs
```diff
@@ -6,22 +6,52 @@ mod gc {
...
+            Err(vm.new_not_implemented_error("".to_owned()))
+        }
     }
 
     #[pyfunction]
     fn disable(_args: FuncArgs, vm: &VirtualMachine) -> PyResult {
-        Err(vm.new_not_implemented_error("".to_owned()))
+        #[cfg(feature = "gc")]
+        {
+            rustpython_vm::object::disable();
```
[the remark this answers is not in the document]
fanninpm: > Also the `gc` module could probably get moved to rustpython-vm proper.

Indeed, in CPython, the `gc` module is compiled into the interpreter (rather than being dynamically linked):

```text
Python 3.11.0 (main, Oct 25 2022, 13:57:33) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import gc
>>> gc.__file__
AttributeError: module 'gc' has no attribute '__file__'. Did you mean: '__name__'?
```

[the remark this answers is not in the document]
discord9: > It would be good for these to have more descriptive names, either a `gc_` prefix or a separate module. Also the _gc module could probably get moved to rustpython-vm proper.

how about "gc_bacon" as feature name(as bacon is the papers' author)?

vm/src/object/core.rs
```diff
@@ -77,11 +91,72 @@ struct Erased;
...
+/// move drop `header` here to prevent UB
+unsafe fn dealloc_only_obj<T: PyObjectPayload>(x: *mut PyObject) {
+    #[cfg(feature = "gc")]
+    {
+        let obj = x.cast::<PyInner<T>>().as_ref().expect("Non-Null Pointer");
+        partially_drop!(obj.header, vtable, weak_list);
+    }
+    std::alloc::dealloc(
+        x.cast(),
+        std::alloc::Layout::for_value(x.cast::<PyInner<T>>().as_ref().unwrap()),
```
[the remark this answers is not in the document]
discord9: > Maybe also could move this out of the function call and into a variable, just to be safe wrt stacked borrows (not sure if the reference is still "live" while dealloc() runs)

Will do, yes I also have doubt on some of my code's soundness, especially with dealloc&drop

vm/src/object/core.rs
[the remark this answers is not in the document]
discord9: > 

Might rename those to `MaybeTrace` `try_trace` and make `is_trace` a const function

derive/src/lib.rs
```diff
@@ -119,3 +120,16 @@ pub fn pypayload(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let input = parse_macro_input!(input as DeriveInput);
     result_to_tokens(pypayload::impl_pypayload(input))
 }
+
+/// use on struct with named fields like `struct A{x:i32, y:i32}` to impl `Trace` for datatype
+///
+/// use `#[notrace]` on fields you wish not to trace
+#[proc_macro_attribute]
```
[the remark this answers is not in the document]
discord9: > This could probably be a normal derive() macro, and then you wouldn't have to deal with removing the notrace attr

Yes, `#[notrace]` is only used four time in th following, still considering whether or not to remove it and hand write Trace:
```
#[pyclass(module = false, name = "enumerate")]
#[derive(Debug)]
#[pytrace]
pub struct PyEnumerate {
    #[notrace]
    counter: PyRwLock<BigInt>,
    iterator: PyIter,
}
```
```
#[pyclass(module = false, name = "iterator")]
#[derive(Debug)]
#[pytrace]
pub struct PySequenceIterator {
    // cached sequence methods
    #[notrace]
    seq_methods: &'static PySequenceMethods,
    internal: PyMutex<PositionIterInternal<PyObjectRef>>,
}
```

```
#[pyclass(module = false, name = "zip")]
#[derive(Debug)]
#[pytrace]
pub struct PyZip {
    iterators: Vec<PyIter>,
    #[notrace]
    strict: PyAtomic<bool>,
}
```

```
#[derive(Debug, Clone)]
#[pytrace]
pub struct ArgMapping {
    obj: PyObjectRef,
    #[notrace]
    methods: &'static PyMappingMethods,
}
```

DimitrisJim:
Maybe we can create a separate CI check that runs the GC with a select few python test files? The idea is to at least have it in and running relevant tests on it while not slowing down the total runtime.

killme2008:
Is this PR still in review? @youknowone @discord9

youknowone (commented):
I am not understanding how this gc is working - and I seem to not have enough time to get it before finish review - but I'd like to merge it as an optional feature and a good starting point to enhance gc better in future.
But before merging this change, I hope to minimize changes of CPython-originated code because changed CPython code will be easily forgotten from future developers and it even might be broken when the future enhancement is correct but not compatible with this change.

discord9:
> I am really sorry for late review. I am not understanding how this gc is working - and I seem to not have enough time to get it before finish review - but I'd like to merge it as an optional feature and a good starting point to enhance gc better in future. But before merging this change, I hope to minimize changes of CPython-originated code because changed CPython code will be easily forgotten from future developers and it even might be broken when the future enhancement is correct but not compatible with this change.

I am thinking about update document on how this GC works, perhaps even add it to documents of RustPython?

youknowone:
That will be great in future. I think separating gc feature and reduce effect to non-gc build as it was will be good enough for now.

youknowone:
i rebased it on main branch

discord9:
> i rebased it on main branch

I will fix the CI

Lib/test/test_ordered_dict.py
```diff
@@ -651,6 +649,8 @@ class A:
         r = weakref.ref(A)
         del A
         gc.collect()
+        # TODO: RUSTPYTHON, Need to fix this: somehow after del A, it takes two call to `gc.collect()`
+        # for gc to realize a loop is there and to be collected
```
youknowone, line 653: Is this still a valid comment for this test?

vm/src/frame.rs
```diff
@@ -349,7 +349,16 @@ impl ExecutingFrame<'_> {
...
+        let mut gc_count = 0;
         loop {
+            gc_count += 1;
+            if gc_count > 1000 {
+                #[cfg(feature = "gc_bacon")]
+                {
+                    crate::object::try_gc();
+                }
+                gc_count = 0;
+            }
```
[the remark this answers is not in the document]
discord9: > Idk well about gc, so I'd like to learn about the reasoning. is this the preferred place to put try_gc? and how the number 1000 is decided? isn't it a potential cause the reason gc is too slow because it is too small?

My bad, I should mimic CPython(https://github.com/python/cpython/blob/7703def37e4fa7d25c3d23756de8f527daa4e165/Python/ceval.c#L833), here in handlePending it does necessary check whether to gc or not, I should rewrite this part

tisonkun:
@discord9 @youknowone do we have some blockers here? Or it's waiting for a rebase and we generally agree on its merge?

discord9:
> @discord9 @youknowone do we have some blockers here? Or it's waiting for a rebase and we generally agree on its merge?

Currently, an old bug stemming from code predating the implementation of garbage collection is causing memory corruption because the garbage collector is indeed collecting that memory. I will open a new pull request once I have fixed those bugs and can reliably pass continuous integration.

youknowone:
@discord9 I am sorry miss the comment. I wish we finally could merge it. If there is any PyTraverse update, could we merge it first? Regardless how GC designed later, it will be useful.

killme2008:
> @discord9 I am sorry miss the comment. I wish we finally could merge it. If there is any PyTraverse update, could we merge it first? Regardless how GC designed later, it will be useful.

+1. It's a first step and it can be improved in future. cc @discord9

PR #4183, opened 2022-09-24
Update function `__qualname__` and `repr`

oow214:
Fix https://github.com/RustPython/RustPython/issues/4088 and refactor function `repr()`


youknowone (approved):
looks good, thank you!

PR #4185, opened 2022-09-27
Use of deprecated function `itertools::zip` use `std::iter::zip` instead

oow214:
When I built RustPython, I got the following warning:
```
warning: use of deprecated function `itertools::zip`: Use [std::iter::zip](https://doc.rust-lang.org/std/iter/fn.zip.html) instead
   --> vm/src/frame.rs:190:39
    |
190 |             for (&k, v) in itertools::zip(&map[..j], &**fastlocals) {
    |                                       ^^^
    |
    = note: `#[warn(deprecated)]` on by default

warning: use of deprecated function `itertools::zip`: Use [std::iter::zip](https://doc.rust-lang.org/std/iter/fn.zip.html) instead
   --> vm/src/frame.rs:200:43
    |
200 |                 for (&k, v) in itertools::zip(keys, values) {
    |     
```
And I edited it to fit the warning.

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -187,7 +187,7 @@ impl FrameRef {
         let j = std::cmp::min(map.len(), code.varnames.len());
         if !code.varnames.is_empty() {
             let fastlocals = self.fastlocals.lock();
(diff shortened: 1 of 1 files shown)


oow214:
Sorry. My version was weird.

Issue #4063, opened 2022-08-14
__match_args__ for structseq


youknowone:
## Feature

See `test_structseq::test_match_args`.

dvermd:
I've made some progress on this one but I need help on how 'n_fields', 'n_sequence_fields' and 'n_unnamed_fields' should be filled.

youknowone:
This is looking done. Thanks @dvermd !

PR #4187, opened 2022-09-30
match args without n_*_fields

dvermd:
This PR treat __match_args__ without n_*_fields

refs: #4063

### DIFF
--- a/vm/src/types/structseq.rs
+++ b/vm/src/types/structseq.rs
@@ -86,5 +86,16 @@ pub trait PyStructSequence: StaticType + PyClassImpl + Sized + 'static {
                 .into(),
             );
         }
+
+        class.set_attr(
+            identifier!(ctx, __match_args__),
+            ctx.new_tuple(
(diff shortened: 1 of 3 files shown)


vm/src/types/structseq.rs
```diff
@@ -86,5 +86,16 @@ pub trait PyStructSequence: StaticType + PyClassImpl + Sized + 'static {
                 .into(),
             );
         }
+
+        class.set_attr(
+            ctx.intern_str("__match_args__"),
```
youknowone, line 91: Because `__match_args__` is a core language attribute, adding it to `vm/src/vm/context.rs` would be a good idea.
Then we can refer interned keyword directly.
```suggestion
            identifier!(vm, __match_args__),
```

youknowone (changes requested):
Thank you for contributing! The change looks good. I left a minor request about interning.

youknowone (approved):

Issue #4102, opened 2022-08-18
Refactor f-string parser


youknowone:
F-string parser is not efficient enough.
Especially not easy to maintain the brace pairs.

First step:
FStringParser need to take &str and consume any amount of matching characters and returns &str again. The remaining bytes will be fed to parser again.

Second steps
- parse function for outer brace and inner expr need to be separated.
- location need to be correctly passed to inner expressions

fanninpm:
Before making the f-string parser more efficient, it might be better to make it _compliant_ with https://github.com/python/cpython/blob/3.10/Lib/test/test_fstring.py (without crashing or throwing unnecessary `SyntaxError`s).

youknowone:
The brace matching is really painful for current structure. I think the first step is also related to implement future features.

dvermd:
I'd like to work on this one.
Could you elaborate a little more on what would be the first step ?

I understand the location passing to the inner expressions as `Location::default()` is used when creating sub-parsers.
Splitting `FStringParser::parse` also seems like a good idea.

youknowone:
`FStringParser::parse` currently consume its inner iterator and returns only its parsed result. Due to its design, we need to detect the end of f-string first and pass the substring iterator to FStringParser.
If we can remove the char iterator from FStringParser and users can pass &str to `FStringParser::parse` and take &str from its result, we will not need to find the end of f-string before calling FStringParser::parse.

dvermd:
Can you explain why the parser accepts stacked '{}' for the format specifier ? I don't manage to forge a working example on cpython with these. I must be missing something...

fanninpm:
@dvermd have you tried taking a look at https://github.com/python/cpython/blob/3.10/Lib/test/test_fstring.py?

charliermarsh:
@youknowone - Heads up that there's a new PEP focused on formalizing the f-string grammar, which is now open for comment: https://discuss.python.org/t/pep-701-syntactic-formalization-of-f-strings/22046

PR #4188, opened 2022-09-30
Refactor fstrings

dvermd:
There's still a problem with this rework where 
```
arg_list = 'hits, misses, maxsize, currsize'
f'lambda _cls, {arg_list}: _tuple_new(_cls, ({arg_list}))
```
returns 
`lambda _cls, {'hits, misses, maxsize, currsize'}: _tuple_new(_cls, ({'hits, misses, maxsize, currsize'}))`
instead of 
`lambda _cls, hits, misses, maxsize, currsize: _tuple_new(_cls, (hits, misses, maxsize, currsize))`

closes: #4102

### DIFF
--- a/compiler/parser/src/fstring.rs
+++ b/compiler/parser/src/fstring.rs
@@ -4,22 +4,15 @@ use crate::{
     error::{FStringError, FStringErrorType, ParseError},
     parser::parse_expression,
 };
-use itertools::Itertools;
 use std::{iter, mem, str};
 
-struct FStringParser<'a> {
-    chars: iter::Peekable<str::Chars<'a>>,
+struct FStringParser {
     str_location: Location,
-    recurse_lvl: u8,
 }
 
-impl<'a> FStringParser<'a> {
-    fn new(source: &'a str, str_location: Location, recurse_lvl: u8) -> Self {
-        Self {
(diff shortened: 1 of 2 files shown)


fanninpm:
> Do you think the test https://github.com/RustPython/RustPython/actions/runs/3222166028/jobs/5270956252#step:8:509 could/should be updated to reflect the new returned value ?

Try playing around with `cargo-insta` to update the relevant snapshot.

dvermd:
To validate this refactor, I took the test_fstring.py from CPython and commented every assert not passing with main branch and checked that it also passes after the refactor.

compiler/parser/src/fstring.rs
```diff
@@ -236,48 +184,105 @@ impl<'a> FStringParser<'a> {
...
                     }
+                    if !content.is_empty() {
+                        values.push(self.expr(ExprKind::Constant {
+                            value: mem::take(&mut content).into(),
+                            kind: None,
+                        }));
+                    }
+
+                    let (parsed_values, remaining_chars) =
+                        self.parse_formatted_value(chars, nested)?;
```
youknowone, line 267: 👍

youknowone (approved):
great! thank you!

PR #4191, opened 2022-10-01
Add OSError.__reduce__

dvermd:
Part of #3611


vm/src/exceptions.rs
```diff
@@ -764,6 +765,7 @@ impl ExceptionZoo {
             // second exception filename
             "filename2" => ctx.none(),
             "__str__" => ctx.new_method("__str__", excs.os_error, os_error_str),
+            "__reduce__" => ctx.new_method("__str__", excs.os_error, os_error_reduce),
```
youknowone, line 768:
```suggestion
            "__reduce__" => ctx.new_method("__reduce__", excs.os_error, os_error_reduce),
```

vm/src/exceptions.rs
```diff
@@ -1297,9 +1335,19 @@ pub(super) mod types {
...
+                zelf.set_attr("filename2", args.args[args.args.len() - 1].clone(), vm)?;
+            }
+        }
 
-    fn base_exception_init(zelf: PyObjectRef, args: FuncArgs, vm: &VirtualMachine) -> PyResult<()> {
-        PyBaseException::init(zelf, args, vm)
+        let mut new_args = args;
+        if new_args.args.len() >= 2 {
+            new_args.args.truncate(2);
+        }
```
youknowone, line 1349: This is a question because I don't know the spec very well. Exceptions usually takes infinite args and store it. What happens for more args than 4 in this case?

Snowapril: It seems to work that no matter how we put arguments to its parenthesis (`work` means it does not throw exceptions).
But according to code and behavior. it works **as expected** when only the given argument length is between 2 and 5.
```python
>>> OSError(1,2).errno
1
>>> OSError(1,2).strerror
2
>>> OSError(1)
OSError(1)
>>> OSError(1).errno
```
https://github.com/python/cpython/blob/3.10/Objects/exceptions.c#L820-L876

dvermd: Let's discuss this point in #4194 ...

PR #4192, opened 2022-10-02
Add end locations to all nodes

charliermarsh:
# Summary

I'm finding myself needing end locations for AST nodes (similar to in the Python AST). Is there interest in enabling this? If so, happy to clean up and test further.


youknowone:
Please feel free to let me know if you need any help

charliermarsh:
@youknowone - Just rebased. I've been using this code successfully in Ruff for a while now, would love to get it merged if you're open to it! Let me know what needs fixing / addressing :)

youknowone:
@charliermarsh I didn't check the details yet, but could you check the failing tests?

charliermarsh:
@youknowone - Absolutely, will do, thanks.

charliermarsh:
@youknowone - Tests passing!

youknowone:
Because we are trying to leave our own footprints to CPython codes as less as possible, I reverted the last commit but tried to change end_location to be optional.

charliermarsh:
I can change it to `Optional` (but populated everywhere, as it is now) if you'd prefer?

charliermarsh:
Or are you suggesting something else? Using `Optional` is less convenient for me as a "client" of the parser but I know the spec marks it as optional :)

charliermarsh:
I can try to make that change today.

charliermarsh:
In practice, does CPython ever omit `end_lineno` and `end_col_offset` when you go through `ast.parse`? Or are they just optional when creating AST nodes "manually"?

youknowone:
@charliermarsh because I made incomplete commit last push, you can work on the new one as a base.

I am sorry, I don't know well about CPython ast that much.
cc @fanninpm @coolreader18

charliermarsh:
@youknowone - Hoping I didn't override your work. Here's the first part.

Lib/test/test_ast.py
```diff
@@ -2307,8 +2307,6 @@ class C:
         cdef = ast.parse(s).body[0]
         self.assertEqual(ast.get_source_segment(s, cdef.body[0], padded=True), s_method)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
```
charliermarsh, line 2311: @youknowone - All other test changes were now able to be reverted.

compiler/ast/src/ast_gen.rs
```diff
@@ -10,14 +10,16 @@ type Ident = String;
...
 pub struct Located<T, U = ()> {
     pub location: Location,
+    pub end_location: Option<Location>,
     pub custom: U,
     pub node: T,
 }
 
 impl<T> Located<T> {
-    pub fn new(location: Location, node: T) -> Self {
+    pub fn new(location: Location, end_location: Location, node: T) -> Self {
```
charliermarsh, line 19: I kept `Location` as a required field in `::new` for now. (We can omit it in special cases by going through the Struct directly, as in `ast.rs`.)

charliermarsh: I could see arguments either way here. On the one hand, it's odd that the field you pass in differs from what's stored on the struct. On the other hand, this encourages good behavior _and_ reduces the number of code changes required.

youknowone: I agree this approach and the idea encouraging good behavior.

compiler/parser/src/context.rs
```diff
@@ -0,0 +1,173 @@
+use rustpython_ast::{Expr, ExprContext, ExprKind};
+
+pub fn set_context(expr: Expr, ctx: ExprContext) -> Expr {
```
youknowone, line 3: is this necessary part of end_location?

youknowone (approved):

compiler/parser/src/context.rs
charliermarsh: Oh no, sorry, I need to revert this. Not sure how it got included.

charliermarsh: (Removed.)

youknowone:
Awesome! Thank you!

Issue #4193, opened 2022-10-02
OSError attributes


dvermd:
## Feature

CPython fills attributes for OSError
```
>>> q = OSError('foo', 'bar', 'baz', None, 'quux')
>>> q.args
('foo', 'bar')
>>> q.errno
'foo'
>>> q.strerror
'bar'
>>> q.filename
'baz'
>>> q.filename2
'quux'
>>> 
```

RustPython doesn't yet:
```
>>>>> p = OSError('foo', 'bar', 'baz')
>>>>> p.args
('foo', 'bar', 'baz')
>>>>> p.errno
'foo'
>>>>> p.strerror
'bar'
>>>>> p.filename
>>>>> p.filename2
>>>>> 
```

## Python Documentation

Here's the doc for OSError : https://docs.python.org/3/library/exceptions.html?highlight=oserror#OSError

PR #4194, opened 2022-10-02
Fill OSError attributes

dvermd:
Tested by removing expected failure and commenting everything but OSError in Lib/test/test_exceptions.py::testAttributes::exceptionList, WindowsError and pickling part of the test.

fixes : #4193

### DIFF
--- a/Lib/test/test_exception_hierarchy.py
+++ b/Lib/test/test_exception_hierarchy.py
@@ -5,6 +5,7 @@ import socket
 import unittest
 import errno
 from errno import EEXIST
+import sys
 
 
 class SubOSError(OSError):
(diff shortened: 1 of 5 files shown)


dvermd:
From the discussion in #4191, there seems to be 2 different behaviors of the OSError constructor:

- the one which takes from 2 to 5 arguments and that fills errno, strerror, winerror, filename and filename2
- the other one which takes 1 or more than 5 arguments, that just behaves like a BaseException without filling any of the OSError fields

Lib/test/test_exceptions.py
```diff
@@ -410,6 +410,32 @@ def test_WindowsError(self):
...
+            self.assertEqual(w.filename2, None)
+            self.assertEqual(str(w), "foo")
+
+            w = OSError('a', 'b', 'c', 'd', 'e', 'f')
+            self.assertEqual(w.errno, None)
+            self.assertEqual(w.winerror, None)
+            self.assertEqual(w.strerror, None)
+            self.assertEqual(w.filename, None)
+            self.assertEqual(w.filename2, None)
+            self.assertEqual(str(w), "('a', 'b', 'c', 'd', 'e', 'f')")
```
youknowone, line 438: if this our own test or copied from cpython? we usually don't write our own tests here but in `extra_tests/snippets/`

dvermd: Moved the new tests to `extra_tests/snippets/builtin_exceptions.py`

youknowone (approved):
thank you!

PR #4195, opened 2022-10-04
Fix compile error when using `vm-tracing-logging` is on

jopemachine:
This PR fixes compile error when using `vm-tracing-logging` feature is on

## Ref

- https://github.com/RustPython/RustPython/discussions/4162


jopemachine:
By the way, I think it would be good to set `RUST_LOG` environment variable to `trace` in RustPython when the `vm-tracing-logging` feature is on to avoid confusion.

I think some users like me could be a little embarrassed if any console log is not printed although the log feature is enabled.

youknowone (approved):

PR #4196, opened 2022-10-04
Update `pprint` lib and tests from CPython v3.12.0a0

jopemachine:
This is a change of cpython 3.10
Python codes are copy-pasted from CPython.

`pprint.pprint` function displays integer with underscores.

* https://github.com/python/cpython/pull/24864


youknowone (approved):
Thank you!

PR #4197, opened 2022-10-04
Update `unittest/runner.py` from CPython v3.12.0a0

jopemachine:
I think updating unittest runner could be useful because it shows unexpected success info without the `--verbose` flag.


jopemachine:
Test failure seems to be related to #4198.

youknowone (approved):

PR #4200, opened 2022-10-04
Add `vm-tracing-logging` feature

jopemachine:
- This PR set env_logger's log level to `trace` when "vm-tracing-logging" feature is enabled.
- This PR enables to pass the "vm-tracing-logging" feature when to build rustpython like below command to trace detail of `rustpython-vm`

```
cargo run --features "vm-tracing-logging"
```


src/lib.rs
```diff
@@ -57,7 +57,13 @@ pub use settings::{opts_with_clap, RunMode};
 /// The main cli of the `rustpython` interpreter. This function will return with `std::process::ExitCode`
 /// based on the return code of the python code ran through the cli.
 pub fn run(init: impl FnOnce(&mut VirtualMachine) + 'static) -> ExitCode {
-    env_logger::init();
+    if cfg!(feature = "vm-tracing-logging") {
```
DimitrisJim, line 60: isn't this typically done with some ENV var instead of a feature? I'm legitimately curious here since I'm not sure but it feels as if a feature is a big of a heavy handed approach to this.

PR #4201, opened 2022-10-04
Add dropwhile.__reduce__

dvermd:
The test `def test_dropwhile(self):` of Lib/test/test_itertools.py is not activated because the line `self.assertRaises(TypeError, next, dropwhile(10, [(4,5)]))` fails.

The bytes against which the pickling is compared in stdlib_itertools.py are generated with CPython

This is part of #3611


youknowone (approved):
looks, great thank you!

Issue #4022, opened 2022-08-08
<code object>::replace()


youknowone:
## Feature

```python
>>> def f(): pass
...
>>> f.__code__.replace()

```

expected: silent
actual: AttributeError

PR #4207, opened 2022-10-09
<code object>::replace()

dannasman:
Hello there!

I started working on issue #4022  and was just wondering whether this is the right place to implement the replace() function.

### DIFF
--- a/vm/src/builtins/code.rs
+++ b/vm/src/builtins/code.rs
@@ -333,7 +333,7 @@ impl PyRef<PyCode> {
                 max_stackdepth: self.code.max_stackdepth,
                 instructions: self.code.instructions.clone(),
                 locations: self.code.locations.clone(),
-                constants: constants.into_iter().map(|o| Literal(o)).collect(),
+                constants: constants.into_iter().map(Literal).collect(),
                 names: names
                     .into_iter()
                     .map(|o| o.as_interned_str(vm).unwrap())
--- a/vm/src/builtins/code.rs
+++ b/vm/src/builtins/code.rs
@@ -5,7 +5,7 @@
 use super::{PyStrRef, PyTupleRef, PyType, PyTypeRef};
 use crate::{
     builtins::PyStrInterned,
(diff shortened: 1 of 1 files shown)


youknowone:
Yes, exactly!

dannasman:
Should I wrap the function parameters with `OptionalArg`? There is quite a lot of optional parameters.  Also what should the parameter types be. For example if you want to replace `flags` field of the code object should the parameter passed to `replace()` be `OptionalArg<u16>` or `OptionalArg<CodeFlags>`?

youknowone:
You'd better to check `#[derive(FromArgs)]`. `OptionalArg` in function signature only supports optional positional args, but you need optional keyword args.
I prefer `CodeFlags` to `u16` when possible. RustPython automatically can make additional check during its conversion.

dannasman:
I managed to implement the method so that it passes `test_replace` in `test_code.py` for all 10 attributes that have a `#[pygetset]` function in `code.rs` (rest I commented out for the duration of testing).

vm/src/builtins/code.rs
```diff
@@ -5,15 +5,39 @@
...
-    function::FuncArgs,
+    function::{FuncArgs, OptionalArg},
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyRef, PyResult, VirtualMachine,
 };
 use num_traits::Zero;
 use std::{borrow::Borrow, fmt, ops::Deref};
 
+#[derive(FromArgs)]
+pub struct ReplaceArgs {
+    #[pyarg(any, optional)]
```
youknowone, line 19:
```suggestion
    #[pyarg(named, optional)]
```
for every lines. because it doesn't take positional argument

youknowone (changes requested):
other parts looks good, thank you!

youknowone (approved):
Great! Thank you for contributing!

dannasman:
Thank you for the guidance!

PR #4211, opened 2022-10-14
Make rustpython-vm compatible with non-js wasm32-unknown & add tests

coolreader18:
Ideally the test crate shouldn't take long to compile, since it's the same target with only a few crates configured differently.

### DIFF
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -102,7 +102,7 @@ mod decl {
         Ok(duration_since_system_now(vm)?.as_secs_f64())
     }
 
-    #[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
+    #[cfg(all(target_arch = "wasm32", feature = "js", not(target_os = "wasi")))]
     fn _time(_vm: &VirtualMachine) -> PyResult<f64> {
         use wasm_bindgen::prelude::*;
         #[wasm_bindgen]
@@ -113,6 +113,11 @@ mod decl {
         }
         // Date.now returns unix time in milliseconds, we want it in seconds
         Ok(Date::now() / 1000.0)
+    }
+
+    #[cfg(all(target_arch = "wasm32", not(feature = "js"), not(target_os = "wasi")))]
+    fn _time(vm: &VirtualMachine) -> PyResult<f64> {
+        Err(vm.new_not_implemented_error("time.time".to_owned()))
     }
 
     #[pyfunction]
--- a/vm/src/vm/vm_object.rs
+++ b/vm/src/vm/vm_object.rs
@@ -19,7 +19,7 @@ impl VirtualMachine {
             self.flush_std();
             panic!("{msg}")
         }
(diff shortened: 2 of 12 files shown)


coolreader18:
Hmm, is there a reason we need to have wasm32-unknown-unknown checked in exotic_targets when we have the wasm job? Is that a holdover from before we had the wasm job?

youknowone:
I recently added it there without specific reason. Do you think it belongs to wasm job? That makes sense.

wasm/wasm-unknown-test/README.md
```diff
@@ -0,0 +1 @@
+A test crate to ensure that `rustpython-vm` compiles on `wasm32-unknown-unknown` without a JS host.
```
youknowone, line 1: :+1:

youknowone:
I added the build to prevent the target's build error.

wasm build job was not fully covering the top-most crate rustpython build
See #3634

coolreader18:
Oh, I see. But that's not supported anyway, right? You can't compile a cli application to wasm w/ js-interop, there's no syscalls or argv or anything.

youknowone:
there was wasm32-unknown use case without js interop.
the top crate is now more useful than vm because user must manually construct interpreter using vm and stdlib since we seprated stdlib crate. `rustpython` crate suggests a convenient way to construct one without knowing the details.

youknowone:
cc @lastmjs

youknowone (approved):
not sure if skipping `cargo check --target wasm32-unknown-unknown --no-default-features` is safe.
other parts looks great

vm/src/stdlib/time.rs
```diff
@@ -102,6 +102,11 @@ mod time {
         Ok(Date::now() / 1000.0)
     }
 
+    #[cfg(all(target_arch = "wasm32", not(feature = "js"), not(target_os = "wasi")))]
+    fn _time(vm: &VirtualMachine) -> PyResult<f64> {
+        Err(vm.new_not_implemented_error("time.time".to_owned()))
```
lastmjs, line 107: I will need a way to implement time, I have the APIs in my environment, but I will need a way to override this

vm/src/vm/vm_object.rs
```diff
@@ -53,6 +53,11 @@ impl VirtualMachine {
             error(&s);
             panic!("{}; exception backtrace above", msg)
         }
+        #[cfg(all(target_arch = "wasm32", not(feature = "js"), not(target_os = "wasi")))]
```
lastmjs, line 56: Wouldn't it be useful to still see information from the exception? Can't you do something like:

```rust
let err_string: String = exc.to_pyobject(vm).repr(vm).unwrap().to_string();
```

Or does the `msg` have everything we need?

lastmjs (changes requested):

vm/src/vm/vm_object.rs
youknowone: It will be useful

PR #4213, opened 2022-10-15
add recursion check in __call__

joohongpark:


youknowone (approved):

vm/src/types/slot.rs
```diff
@@ -254,7 +254,9 @@ fn hash_wrapper(zelf: &PyObject, vm: &VirtualMachine) -> PyResult<PyHash> {
 }
 
 fn call_wrapper(zelf: &PyObject, args: FuncArgs, vm: &VirtualMachine) -> PyResult {
-    vm.call_special_method(zelf.to_owned(), identifier!(vm, __call__), args)
+    vm.with_recursion("while calling a Python object", || {
+        vm.call_special_method(zelf.to_owned(), identifier!(vm, __call__), args)
+    })
```
Snowapril, line 259: Actually, in cpython, the recursion check what we inserted in this PR exist inside of `_PyObject_Call`. It is corresponded to `invoke` in our code. I'm a little bit afraid that `invoke` is currently used anywhere in our codes. It can make a little bit slow-down.

PR #4217, opened 2022-10-15
Add annotations for type

oow214:
Fix test_default_annotations_exist in test_opcodes

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -524,6 +524,32 @@ impl PyType {
     }
 
     #[pygetset(magic)]
+    fn annotations(&self, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
+        if !self.slots.flags.has_feature(PyTypeFlags::HEAPTYPE) {
+            return Err(vm.new_attribute_error(format!(
+                "type object '{}' has no attribute '__annotations__'",
+                self.name()
+            )));
+        }
+
+        let __annotations__ = identifier!(vm, __annotations__);
+        let annotations = self.attributes.read().get(__annotations__).cloned();
+
+        let annotations = if let Some(annotations) = annotations {
+            annotations
+        } else {
+            let annotations: PyObjectRef = vm.ctx.new_dict().into();
+            let removed = self
+                .attributes
+                .write()
+                .insert(__annotations__, annotations.clone());
+            debug_assert!(removed.is_none());
+            annotations
+        };
+        Ok(annotations)
+    }
+
+    #[pygetset(magic)]
     pub fn module(&self, vm: &VirtualMachine) -> PyObjectRef {
         self.attributes
             .read()
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -549,6 +549,34 @@ impl PyType {
         Ok(annotations)
     }
 
+    #[pygetset(magic, setter)]
+    fn set_annotations(&self, value: Option<PyObjectRef>, vm: &VirtualMachine) -> PyResult<()> {
+        if self.slots.flags.has_feature(PyTypeFlags::IMMUTABLETYPE) {
+            return Err(vm.new_type_error(format!(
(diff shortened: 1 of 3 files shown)


vm/src/builtins/type.rs
```diff
@@ -523,6 +523,22 @@ impl PyType {
...
+            .get(identifier!(vm, __annotations__))
+            .cloned()
+            .and_then(|found| {
+                if found.fast_isinstance(vm.ctx.types.dict_type) {
+                    None
+                } else {
+                    Some(found)
+                }
+            })
+            .unwrap_or_else(|| vm.ctx.new_dict().into())
```
youknowone, line 539: I am not really getting how this is going. Is it creating a new dict when the found `__annotations__` is a dict?

Snowapril: Oh. If there is no `__annotations__` inside the dict, we need to create `__annotations__` into the dict and return that created one.

youknowone:
@Snowapril could you review this one?

vm/src/builtins/type.rs
```diff
@@ -523,6 +523,57 @@ impl PyType {
...
+        Ok(self
+            .attributes
+            .read()
+            .get(identifier!(vm, __annotations__))
+            .cloned()
+            .unwrap_or_else(|| {
+                self.attributes
+                    .write()
+                    .insert(identifier!(vm, __annotations__), vm.ctx.new_dict().into())
+                    .unwrap_or_else(|| vm.ctx.new_dict().into())
```
youknowone, line 544: The result of `insert` is not the newly inserted one but the old removed one.
The result of insert must be always `None` here.

vm/src/builtins/type.rs
```diff
@@ -523,6 +523,57 @@ impl PyType {
...
+            }))
```
youknowone, line 545: Because `entry()` doesn't fit to rwlock very well, I suggest this form.
```suggestion
        let __annotations__ = identifier!(vm, __annotations__);
        let annotations = if let Some(annotations) = self.attributes.read().get(__annotations__) {
            annotations.clone()
        } else {
            let annotations: PyObjectRef = vm.ctx.new_dict().into();
            let removed = self
                .attributes
                .write()
                .insert(__annotations__, annotations.clone());
            debug_assert!(removed.is_none());
            annotations
        };
        Ok(annotations)
```

vm/src/builtins/type.rs
```diff
@@ -523,6 +523,57 @@ impl PyType {
...
+                .read()
+                .get(__annotations__)
+                .cloned()
+                .ok_or_else(|| {
+                    vm.new_attribute_error(format!(
+                        "'{}' object has no attribute '__annotations__'",
+                        self.name()
+                    ))
+                })?;
+            self.attributes.write().remove(__annotations__);
```
youknowone, line 571: maybe a single write lock? double-deleting `__annotations__` will not be a common case.

vm/src/builtins/type.rs
```diff
@@ -523,6 +523,56 @@ impl PyType {
...
+        let __annotations__ = identifier!(vm, __annotations__);
+        let annotations = self.attributes.read().get(__annotations__).cloned();
+
+        let annotations = annotations.unwrap_or_else(|| {
+            let annotations: PyObjectRef = vm.ctx.new_dict().into();
+            self.attributes
+                .write()
+                .insert(__annotations__, annotations.clone());
+            annotations
+        });
```
youknowone, line 544: I'd like to keep my original suggestion for a few (weak) reasons.
- `debug_assert!` for insert result will be useful to prevent logic bug.
- I'd like not to put side-effect in functional styled codes. I expect `unwrap_or_else` is immutable operation unless it panics. (though this is a sort of preference)

vm/src/builtins/type.rs
oow214: Can I use `write()` in `else` while using `read()` in `if let`? I tried it, but an error seems to occur in this part, so I edited this part.

youknowone (approved):
Thank you!

PR #4220, opened 2022-10-15
Start string location at kind or quote prefix

### LINKED ISSUE #4219: Parsed string start location differs from CPython's AST parser
If you parse `x = u"""abc"""`, RustPython reports that the `Constant` starts at one-indexed column 7 (just after the first `"`):

```
[
    Located {
        location: Location {
            row: 1,
            column: 1,
        },
        end_location: Location {
            row: 1,
            column: 15,
        },
        custom: (),
        node: Assign {
            targets: [
                Located {
                    location: Location {
                        row: 1,
                        column: 1,
                    },
                    end_location: Location {
                        row: 1,
                        column: 2,
                    },
                    custom: (),
                    node: Name {
                        id: "x",
                        ctx: Store,
                    },
                },
            ],
            value: Located {
                location: Location {
                    row: 1,
                    column: 7,
                },
                end_location: Location {
                    row: 1,
                    column: 15,
                },
                custom: (),
                node: Constant {
                    value: Str(
                        "abc",
                    ),
                    kind: Some(
                        "u",
                    ),
                },
            },
            type_comment: None,
        },
    },
]
```

CPython reports that the `Constant` starts at one-indexed column 5 (at the `u`).

charliermarsh:
Resolves #4219.

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -205,7 +205,9 @@ where
 
             // Check if we have a string:
             if self.chr0 == Some('"') || self.chr0 == Some('\'') {
-                return self.lex_string(saw_b, saw_r, saw_u, saw_f);
+                return self
+                    .lex_string(saw_b, saw_r, saw_u, saw_f)
+                    .map(|(_, tok, end_pos)| (start_pos, tok, end_pos));
             }
         }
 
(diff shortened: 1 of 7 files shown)


compiler/parser/src/lexer.rs
```diff
@@ -205,7 +205,9 @@ where
 
             // Check if we have a string:
             if self.chr0 == Some('"') || self.chr0 == Some('\'') {
-                return self.lex_string(saw_b, saw_r, saw_u, saw_f);
+                return self
+                    .lex_string(saw_b, saw_r, saw_u, saw_f)
+                    .map(|(_, tok, end_pos)| (start_pos, tok, end_pos));
```
charliermarsh, line 210: We want to use the `u` or `r` or first quote prefix as the "start". Right now, RustPython uses the location just _after_ the first quote.

youknowone (approved):
nice catch! thank you!

PR #4222, opened 2022-10-15
Refactor Mapping Protocol

qingshi163:


youknowone:
@Snowapril could you review this pr?

Snowapril (approved):
Looks good to me! The part of this change is actually what I intended when I implemented mapping protocol at the beginning. By this change, the only field that needs to be updated is changed by a new wrapper. Looks good.

vm/src/types/slot.rs
```diff
@@ -939,16 +941,10 @@ pub trait AsBuffer: PyPayload {
...
-    const AS_MAPPING: PyMappingMethods;
-
-    #[inline]
     #[pyslot]
-    fn as_mapping(_zelf: &PyObject, _vm: &VirtualMachine) -> &'static PyMappingMethods {
-        &Self::AS_MAPPING
-    }
+    fn as_mapping() -> &'static PyMappingMethods;
 
-    #[inline]
```
youknowone, line 950: maybe accidentally removed?

youknowone (approved):

PR #4225, opened 2022-10-16
Fix class method of a float subclass

joohongpark:
Fixed return type for fromhex class method of float subclass.

### DIFF
--- a/vm/src/builtins/float.rs
+++ b/vm/src/builtins/float.rs
@@ -13,7 +13,7 @@ use crate::{
         PyComparisonValue,
     },
     protocol::{PyNumber, PyNumberMethods},
-    types::{AsNumber, Comparable, Constructor, Hashable, PyComparisonOp},
+    types::{AsNumber, Callable, Comparable, Constructor, Hashable, PyComparisonOp},
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyRef, PyResult,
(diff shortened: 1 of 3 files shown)


vm/src/builtins/float.rs
```diff
@@ -481,11 +481,13 @@ impl PyFloat {
...
-        float_ops::from_hex(repr.as_str().trim()).ok_or_else(|| {
-            vm.new_value_error("invalid hexadecimal floating-point string".to_owned())
-        })
+    #[pyclassmethod]
+    fn fromhex(cls: PyTypeRef, string: PyStrRef, vm: &VirtualMachine) -> PyResult {
+        let result = float_ops::from_hex(string.as_str().trim());
+        match result {
+            Some(value) => PyType::call(&cls, vec![vm.ctx.new_float(value).into()].into(), vm),
+            None => Err(vm.new_value_error("invalid hexadecimal floating-point string".to_owned())),
+        }
```
Snowapril, line 490:
```suggestion
        let result = float_ops::from_hex(string.as_str().trim()).ok_or_else(|| {
            vm.new_value_error("invalid hexadecimal floating-point string".to_owned())
        })?;
        PyType::call(&cls, vec![vm.ctx.new_float(result).into()].into(), vm)
```

youknowone (approved):

youknowone:
Thank you!

PR #4227, opened 2022-10-16
Start simple string at quote mark

charliermarsh:
The same as #4220, but for simple strings. E.g., right now `"foo"` starts at column 2, instead of column 1.


charliermarsh:
@youknowone - Sorry, I missed this the first time around.

youknowone (approved):

PR #4229, opened 2022-10-17
Expose a method to parse AST from tokens directly

charliermarsh:
I'm using this in [Ruff](https://github.com/charliermarsh/ruff) where it's useful to be able to generate the token stream once upfront, use it in a few places, then pass it to the parser without re-tokenizing the entire stream.

### DIFF
--- a/compiler/parser/src/parser.rs
+++ b/compiler/parser/src/parser.rs
@@ -5,9 +5,11 @@
 //! parse a whole program, a single statement, or a single
 //! expression.
 
+use std::iter;
+
+use crate::lexer::LexResult;
 pub use crate::mode::Mode;
 use crate::{ast, error::ParseError, lexer, python};
-use std::iter;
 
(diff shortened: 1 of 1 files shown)


compiler/parser/src/parser.rs
```diff
@@ -23,6 +25,17 @@ pub fn parse_program(source: &str, source_path: &str) -> Result<ast::Suite, Pars
     })
 }
 
+/// Parse the token stream for a full python program.
+pub fn parse_program_tokens(
```
youknowone, line 29: I understand we'd better to add `parse_tokens` for some use case. But `parse_program_tokens` too?
This is a very specific use case which can be also covered by `parse_program_tokens`.

charliermarsh: Removed.

youknowone (approved):

PR #4231, opened 2022-10-17
Improve string fstring coverage

dvermd:
The test_fstring.py file comes from CPython

This improve coverage as of #1671.

### DIFF
--- a/Lib/test/test_fstring.py
+++ b/Lib/test/test_fstring.py
@@ -1,0 +1,1342 @@
+# -*- coding: utf-8 -*-
+# There are tests here with unicode string literals and
+# identifiers. There's a code in ast.c that was added because of a
+# failure with a non-ascii-only expression.  So, I have tests for
+# that.  There are workarounds that would let me run tests for that
+# code without unicode identifiers and strings, but just using them
+# directly seems like the easiest and therefore safest thing to do.
+# Unicode identifiers in tests is allowed by PEP 3131.
+
+import ast
+import os
+import re
+import types
+import decimal
+import unittest
+from test.support.os_helper import temp_cwd
+from test.support.script_helper import assert_python_failure
+
+a_global = 'global variable'
+
+# You could argue that I'm too strict in looking for specific error
+#  values with assertRaisesRegex, but without it it's way too easy to
+#  make a syntax error in the test strings. Especially with all of the
+#  triple quotes, raw strings, backslashes, etc. I think it's a
+#  worthwhile tradeoff. When I switched to this method, I found many
+#  examples where I wasn't testing what I thought I was.
+
+class TestCase(unittest.TestCase):
+    def assertAllRaise(self, exception_type, regex, error_strings):
+        for str in error_strings:
+            with self.subTest(str=str):
+                with self.assertRaisesRegex(exception_type, regex):
+                    eval(str)
+
+    def test__format__lookup(self):
+        # Make sure __format__ is looked up on the type, not the instance.
+        class X:
+            def __format__(self, spec):
+                return 'class'
+
+        x = X()
+
+        # Add a bound __format__ method to the 'y' instance, but not
+        #  the 'x' instance.
+        y = X()
+        y.__format__ = types.MethodType(lambda self, spec: 'instance', y)
+
+        self.assertEqual(f'{y}', format(y))
+        self.assertEqual(f'{y}', 'class')
+        self.assertEqual(format(x), format(y))
+
+        # __format__ is not called this way, but still make sure it
+        #  returns what we expect (so we can make sure we're bypassing
+        #  it).
+        self.assertEqual(x.__format__(''), 'class')
+        self.assertEqual(y.__format__(''), 'instance')
+
+        # This is how __format__ is actually called.
+        self.assertEqual(type(x).__format__(x, ''), 'class')
+        self.assertEqual(type(y).__format__(y, ''), 'class')
+
+    def test_ast(self):
+        # Inspired by http://bugs.python.org/issue24975
+        class X:
+            def __init__(self):
+                self.called = False
+            def __call__(self):
+                self.called = True
+                return 4
+        x = X()
+        expr = """
+a = 10
+f'{a * x()}'"""
+        t = ast.parse(expr)
+        c = compile(t, '', 'exec')
+
+        # Make sure x was not called.
+        self.assertFalse(x.called)
+
+        # Actually run the code.
+        exec(c)
+
+        # Make sure x was called.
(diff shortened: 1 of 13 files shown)


youknowone:
@charliermarsh could you review this PR when it is ready?

fanninpm:
There are currently a bunch of unexpectedly passing tests in `test_ast.py` and `test_tokenize.py`.

dvermd:
The change on python.lalrpop might be extended to all binop. What do you think ?

Lib/test/test_global.py
```diff
@@ -25,15 +25,15 @@ def wrong1():
     global a
     global b
 """
-        check_syntax_error(self, prog_text_1, lineno=4, offset=5)
+        check_syntax_error(self, prog_text_1, lineno=4, offset=4)
```
youknowone, line 28: We mostly try not to make changes in cpython-originated code.
Please keep this as original code and add `@unittest.expectedFailure` if you cannot keep this compatibility with the patch.

compiler/core/src/location.rs
```diff
@@ -43,7 +43,7 @@ impl Location {
 
     pub fn reset(&mut self) {
         self.row = 1;
-        self.column = 1;
+        self.column = 0;
```
youknowone, line 46: maybe this is the reason `test_global` is broken.

compiler/parser/src/fstring.rs
```diff
@@ -67,10 +67,18 @@ impl FStringParser {
                         Some('a') => ConversionFlag::Ascii,
                         Some('r') => ConversionFlag::Repr,
                         Some(_) => {
-                            return Err(InvalidConversionFlag);
+                            if expression[1..].trim().is_empty() {
+                                return Err(EmptyExpression);
+                            } else {
+                                return Err(InvalidConversionFlag);
+                            }
```
youknowone, line 74:
```suggestion
                            return Err(if expression[1..].trim().is_empty() {
                                EmptyExpression
                            } else {
                                InvalidConversionFlag
                            };
```
This is also possible

youknowone (commented):
Thank you for contributing. I left a few comments about policy and styles.

@charliermarsh Could you please do the technical review?

compiler/core/src/location.rs
dvermd: That change make so many other test pass for example in `test_ast` that I don't think the `test_global` error comes from it.
There may have been 2 problems hiding/cancelling each other for `test_global` to be OK.

Lib/test/test_global.py
dvermd: fixed

fanninpm:
Can you split d34d361 into two separate commits: one commit adding the test file, and one commit adding the RustPython changes?

charliermarsh:
(I promise to review when I can, if others get to it first no prob.)

dvermd:
> Can you split [d34d361](https://github.com/RustPython/RustPython/commit/d34d361cefc643537c9c2e51eccda07a64b9181c) into two separate commits: one commit adding the test file, and one commit adding the RustPython changes?

Done

dvermd:
Looks like some more unexpected success appeared

Lib/test/test_fstring.py
```diff
@@ -0,0 +1,1392 @@
...
+        self.assertFalse(x.called)
+
+        # Actually run the code.
+        exec(c)
+
+        # Make sure x was called.
+        self.assertTrue(x.called)
+
+    # TODO: RUSTPYTHON - binop lineno
+    @unittest.expectedFailure
```
fanninpm, line 87: These two lines, and pairs of lines like this, should live in a separate commit.

dvermd: commit split and branch rebased on main

Lib/test/test_global.py
dvermd: I think this change might be handle the same way this CPython commit handles it [025eb98dc0](https://github.com/python/cpython/commit/025eb98dc0c1dc27404df6c544fc2944e0fa9f3a): make all syntax error 1-based by incrementing location in "constructor" by 1

youknowone: Oh, thank you! good to know.

youknowone:
which version of CPython did you use for https://github.com/RustPython/RustPython/pull/4231/commits/d48d998e298a75c0f8ea9d629d0a77835f084219 ?

compiler/core/src/location.rs
charliermarsh: I guess that's consistent with the CPython `ast` module, but it seems like a fairly significant change.

compiler/parser/src/fstring.rs
```diff
@@ -245,16 +285,22 @@ impl FStringParser {
...
                     if nested == 0 {
-                        if let Some('{') = chars.peek() {
-                            chars.next();
-                            content.push('{');
-                            continue;
+                        match chars.peek() {
+                            Some('{') => {
+                                chars.next();
+                                content.push('{');
+                                // brace_opened = true;
```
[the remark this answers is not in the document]
dvermd: This is fixed now

dvermd:
> which version of CPython did you use for [d48d998](https://github.com/RustPython/RustPython/commit/d48d998e298a75c0f8ea9d629d0a77835f084219) ?

I took it from [CPython 3.10.8](https://github.com/python/cpython/blob/v3.10.8/Lib/test/test_fstring.py)

fanninpm:
> > which version of CPython did you use for [d48d998](https://github.com/RustPython/RustPython/commit/d48d998e298a75c0f8ea9d629d0a77835f084219) ?
> 
> I took it from [CPython 3.10.8](https://github.com/python/cpython/blob/v3.10.8/Lib/test/test_fstring.py)

I think @youknowone wants you to amend the commit name to mention that version.

dvermd:
> Is it intended that all locations are now 0-indexed rather than 1-indexed?

@charliermarsh It looks like it's how CPython handles it regarding how it improve coverage in test_ast.py. 

Also @youknowone already merged this  part of the PR.

dvermd:
> > > which version of CPython did you use for [d48d998](https://github.com/RustPython/RustPython/commit/d48d998e298a75c0f8ea9d629d0a77835f084219) ?
> > 
> > 
> > I took it from [CPython 3.10.8](https://github.com/python/cpython/blob/v3.10.8/Lib/test/test_fstring.py)
> 
> I think @youknowone wants you to amend the commit name to mention that version.

commit amended with the CPython version

youknowone (approved):
@dvermd Thank you! This is a big progress!

@charliermarsh @fanninpm Thank you for the reviews!

PR #4232, opened 2022-10-18
Chain reduce

dannasman:
Hello!

I tried to implement the `reduce` method for `PyItertoolsChain` (issue #3611) but got stuck. I used the [CPython implementation](https://github.com/python/cpython/blob/9608bef84afd797ba6d16ec97439909f2f0d1095/Modules/itertoolsmodule.c#L2276) as a reference. The implementation seems to work just fine with simple arrays etc. but fails the `test_chain_reducible` test. Any suggestions on why?

The error message:
```
ERROR: test_chain_reducible (test.test_itertools.TestBasicOps)
  File "/home/dan/Projects/RustPython/pylib/Lib/test/test_itertools.py", line 187, in test_chain_reducible
    self.assertEqual(list(oper(it)), list('abcdef'))
  File "/home/dan/Projects/RustPython/pylib/Lib/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/home/dan/Projects/RustPython/pylib/Lib/copy.py", line 279, in _reconstruct
    y.__dict__.update(state)
AttributeError: 'chain' object has no attribute '__dict__'

```

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -43,7 +43,7 @@ mod decl {
             .map(Into::into)
         }
 
-        #[pyclassmethod]    
+        #[pyclassmethod]
         fn from_iterable(
             cls: PyTypeRef,
             source: PyObjectRef,
@@ -63,19 +63,17 @@ mod decl {
 
         #[pymethod(magic)]
         fn reduce(zelf: PyRef<Self>, vm: &VirtualMachine) -> PyResult<PyTupleRef> {
-            let source =  zelf.source.read().clone();
+            let source = zelf.source.read().clone();
             let active = zelf.active.read().clone();
             let cls = zelf.class().to_owned();
             match source {
-                Some(source) => {
-                    match active {
-                        Some(active) => {
-                            Ok(vm.new_tuple((cls, vm.ctx.empty_tuple.clone(), (source, active))))
-                        },
-                        None => Ok(vm.new_tuple((cls, vm.ctx.empty_tuple.clone(), (source, ))))
+                Some(source) => match active {
+                    Some(active) => {
+                        Ok(vm.new_tuple((cls, vm.ctx.empty_tuple.clone(), (source, active))))
                     }
+                    None => Ok(vm.new_tuple((cls, vm.ctx.empty_tuple.clone(), (source,)))),
                 },
-                None => Ok(vm.new_tuple((cls, vm.ctx.empty_tuple.clone())))
+                None => Ok(vm.new_tuple((cls, vm.ctx.empty_tuple.clone()))),
             }
         }
     }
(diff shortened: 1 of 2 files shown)


youknowone:
Search for `HAS_DICT` in the same file and add it to `Chain`'s flags annotation.

dannasman:
I added `HAS_DICT` and now the problem seems to be that when a `deepcopy` operation is performed on a chain object it appears empty. I was thinking that it might iterate through the values somewhere where it shouldn't and at the same time "clears" them.

How it works with Python 3:
```
>>> import copy
>>> from itertools import chain
>>> oper = copy.deepcopy
>>> it = chain('abc', 'def')
>>> res = oper(it)
>>> list(res)
['a', 'b', 'c', 'd', 'e', 'f']
>>> 
```
How it works with this implementation:
```
>>>>> import copy
>>>>> from itertools import chain
>>>>> oper = copy.deepcopy
>>>>> it = chain('abc', 'def')
>>>>> res = oper(it)
>>>>> list(res)
[]
>>>>> 

```

dannasman:
The solution was to implement `setstate` for chain. This will also solve #3779 I think,

fanninpm:
@Snowapril you may be interested in this

vm/src/stdlib/itertools.rs
```diff
@@ -59,6 +59,51 @@ mod decl {
...
+            let cls = zelf.class().to_owned();
+            match source {
+                Some(source) => match active {
+                    Some(active) => {
+                        Ok(vm.new_tuple((cls, vm.ctx.empty_tuple.clone(), (source, active))))
+                    }
+                    None => Ok(vm.new_tuple((cls, vm.ctx.empty_tuple.clone(), (source,)))),
+                },
+                None => Ok(vm.new_tuple((cls, vm.ctx.empty_tuple.clone()))),
+            }
```
youknowone, line 76:
```suggestion
            let empty_tuple = vm.ctx.empty_tuple.clone();
            let reduced = match source {
                Some(source) => match active {
                    Some(active) => vm.new_tuple((cls, empty_tuple, (source, active))),
                    None => vm.new_tuple((cls, empty_tuple, (source,))),
                },
                None => vm.new_tuple((cls, empty_tuple)),
            };
            Ok(reduced)
```

vm/src/stdlib/itertools.rs
```diff
@@ -59,6 +59,51 @@ mod decl {
...
+                *zelf.source.write() = source.to_owned().try_into_value(vm)?;
+                return Ok(());
+            }
+            let active = &args[1];
+
+            if !PyIter::check(source.as_ref()) || !PyIter::check(active.as_ref()) {
+                return Err(vm.new_type_error(String::from("Arguments must be iterators.")));
+            }
+            *zelf.source.write() = source.to_owned().try_into_value(vm)?;
+            *zelf.active.write() = active.to_owned().try_into_value(vm)?;
```
youknowone, line 104: To avoid unpleasure broken synchronization, locking both and writing will be better.
```suggestion
            let source_lock = zelf.source.write();
            let active_lock = zelf.active.write();
            *source_lock = source.to_owned().try_into_value(vm)?;
            *active_lock = active.to_owned().try_into_value(vm)?;
```

youknowone (changes requested):
Looks good! I left a few suggestions about style and locks.

youknowone (approved):

youknowone:
Thank you for contributing!

PR #4234, opened 2022-10-19
Fix `global` behavior on class

### LINKED ISSUE #3993: Using `global` on inner class has different outcome

jeffwang0516:
Fix #3993 

This pr tries to fix the compile time issue for `global` declaration.

Since lots of the tests under `Lib/test` used `global SomeClass` inside their test methods, this fix might help in the future.
e.g. Test `test_dataclasses_pickleable` has passed because the global classes are pickleable after this fix.

---

Take the following code snippet as example:
```python
def test_c_methods():
    global Subclass
    class Subclass(tuple):
        class Nested(str):
            pass

test_c_methods()
```
The result before this pr:

```
>>>>> Subclass.__qualname__
'test_c_methods.<locals>.Subclass'
>>>>>
>>>>> Subclass.Nested.__qualname__
'test_c_methods.<locals>.Subclass.Nested'
```

The result after fix:
```
>>> Subclass.__qualname__
'Subclass'
>>>
>>> Subclass.Nested.__qualname__
'Subclass.Nested'
```

### DIFF
--- a/Lib/test/test_pickle.py
+++ b/Lib/test/test_pickle.py
@@ -110,10 +110,8 @@ class PyPicklerTests(AbstractPickleTests, unittest.TestCase):
     def test_in_band_buffers(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_in_band_buffers() # TODO: RUSTPYTHON, remove when this passes
 
-    # TODO: RUSTPYTHON, pickle.PicklingError
-    @unittest.expectedFailure
-    def test_nested_names(self): # TODO: RUSTPYTHON, remove when this passes
-        super().test_nested_names() # TODO: RUSTPYTHON, remove when this passes
+    def test_nested_names(self):
+        super().test_nested_names()
 
     # TODO: RUSTPYTHON, AttributeError: module 'pickle' has no attribute 'PickleBuffer'
     @unittest.expectedFailure
@@ -201,10 +199,8 @@ class InMemoryPickleTests(AbstractPickleTests, AbstractUnpickleTests,
     def test_load_python2_str_as_bytes(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_load_python2_str_as_bytes() # TODO: RUSTPYTHON, remove when this passes
 
-    # TODO: RUSTPYTHON, pickle.PicklingError
-    @unittest.expectedFailure
-    def test_nested_names(self): # TODO: RUSTPYTHON, remove when this passes
-        super().test_nested_names() # TODO: RUSTPYTHON, remove when this passes
+    def test_nested_names(self):
+        super().test_nested_names()
 
     # TODO: RUSTPYTHON, AttributeError: module 'pickle' has no attribute 'PickleBuffer'
     @unittest.expectedFailure
--- a/Lib/test/test_pickle.py
+++ b/Lib/test/test_pickle.py
@@ -110,9 +110,6 @@ class PyPicklerTests(AbstractPickleTests, unittest.TestCase):
     def test_in_band_buffers(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_in_band_buffers() # TODO: RUSTPYTHON, remove when this passes
 
-    def test_nested_names(self):
-        super().test_nested_names()
-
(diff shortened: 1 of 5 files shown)


Lib/test/test_pickle.py
```diff
@@ -110,10 +110,8 @@ def test_complex_newobj_ex(self): # TODO: RUSTPYTHON, remove when this passes
     def test_in_band_buffers(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_in_band_buffers() # TODO: RUSTPYTHON, remove when this passes
 
-    # TODO: RUSTPYTHON, pickle.PicklingError
-    @unittest.expectedFailure
-    def test_nested_names(self): # TODO: RUSTPYTHON, remove when this passes
-        super().test_nested_names() # TODO: RUSTPYTHON, remove when this passes
+    def test_nested_names(self):
+        super().test_nested_names()
```
youknowone, line 114:
```suggestion
```

Lib/test/test_pickle.py
```diff
@@ -201,10 +199,8 @@ def test_in_band_buffers(self): # TODO: RUSTPYTHON, remove when this passes
     def test_load_python2_str_as_bytes(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_load_python2_str_as_bytes() # TODO: RUSTPYTHON, remove when this passes
 
-    # TODO: RUSTPYTHON, pickle.PicklingError
-    @unittest.expectedFailure
-    def test_nested_names(self): # TODO: RUSTPYTHON, remove when this passes
-        super().test_nested_names() # TODO: RUSTPYTHON, remove when this passes
+    def test_nested_names(self):
+        super().test_nested_names()
```
youknowone, line 203:
```suggestion
```

Lib/test/test_pickletools.py
```diff
@@ -35,9 +35,7 @@ def test_complex_newobj_ex(self): # TODO: RUSTPYTHON, remove when this passes
     def test_in_band_buffers(self): # TODO: RUSTPYTHON, remove when this passes
         super().test_in_band_buffers()
 
-    # TODO: RUSTPYTHON, pickle.PicklingError
-    @unittest.expectedFailure
-    def test_nested_names(self): # TODO: RUSTPYTHON, remove when this passes
+    def test_nested_names(self):
```
youknowone, line 38: the remove comment means we can remove the whole test function when done

compiler/codegen/src/compile.rs
```diff
@@ -1275,6 +1275,17 @@ impl Compiler {
...
+        let symbol_table = self.symbol_table_stack.last().unwrap();
+        let symbol = symbol_table.lookup(name.as_ref()).expect(
+            "The symbol must be present in the symbol table, even when it is undefined in python.",
+        );
+        let mut prefix_path = vec![];
+        if symbol.scope == SymbolScope::GlobalExplicit {
+            while let Some(value) = self.qualified_path.pop() {
+                prefix_path.insert(0, value);
+            }
+        }
```
youknowone, line 1288: Because `prefix_path` is only used for global, I'd like to add it to variable names.
I also suggest changing the while loop to extend and drain.
```suggestion
        let mut global_path_prefix = Vec::new();
        if symbol.scope == SymbolScope::GlobalExplicit {
            global_path_prefix.extend(self.qualified_path.drain(..));
        }
```

youknowone (changes requested):
Awesome! You must have done a deep research about it. Thanks a lot. I left a few suggestion comments.

compiler/codegen/src/compile.rs
jeffwang0516: extend & drain is new to me! Thanks

jeffwang0516: I found that clippy suggests using `append`

https://rust-lang.github.io/rust-clippy/master/index.html#extend_with_drain

youknowone: that looks better

youknowone (approved):

youknowone:
Thank you for contributing!

PR #4235, opened 2022-10-19
Add reduce and setstate for zip_longest

dannasman:
Hi!

Here is a solution proposal for `reduce` and `setstate` functions for `zip_longest` (#3611).


youknowone (approved):
Thank you for contributing!

PR #4239, opened 2022-10-22
Add `name` field and `__set_name__` method in property type

oow214:
Add `name` field and `__set_name__` method in property type


oow214:
Sorry, this is already working on another PR 😓

PR #4240, opened 2022-10-22
Add `__doc__` for type

oow214:
Added `__doc__` and `get_doc_from_internal_doc` for type

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -914,11 +914,43 @@ impl PyType {
     }
 
     #[pygetset(magic)]
+    fn doc(&self, vm: &VirtualMachine) -> Option<PyObjectRef> {
+        if !self.slots.flags.has_feature(PyTypeFlags::HEAPTYPE) {
+            return self
+                .slots
+                .doc
+                .write()
+                .and_then(|doc| get_doc_from_internal_doc(&self.name(), doc))
+                .map(|signature| signature.to_pyobject(vm));
+        }
+
+        self.attributes
+            .read()
+            .get(identifier!(vm, __doc__))
+            .cloned()
+    }
+
+    #[pygetset(magic)]
     fn text_signature(&self) -> Option<String> {
         self.slots
             .doc
+            .write()
             .and_then(|doc| get_text_signature_from_internal_doc(&self.name(), doc))
             .map(|signature| signature.to_string())
+    }
+
+    #[pygetset(magic, setter)]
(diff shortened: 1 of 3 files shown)


vm/src/builtins/type.rs
```diff
@@ -913,13 +913,41 @@ impl PyType {
...
     #[pygetset(magic)]
     fn text_signature(&self) -> Option<String> {
         self.slots
             .doc
             .and_then(|doc| get_text_signature_from_internal_doc(&self.name(), doc))
             .map(|signature| signature.to_string())
     }
+
+    #[pygetset(magic, setter)]
+    fn set_doc(&mut self, value: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
```
youknowone, line 941:
```suggestion
    fn set_doc(&self, value: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
```
`mut` here is not allowed. `self.slots.doc` needs a lock.

vm/src/builtins/type.rs
```diff
@@ -913,13 +913,45 @@ impl PyType {
...
+            .write()
             .and_then(|doc| get_text_signature_from_internal_doc(&self.name(), doc))
             .map(|signature| signature.to_string())
     }
+
+    #[pygetset(magic, setter)]
+    fn set_doc(&self, value: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        let doc = value.as_interned_str(vm).ok_or_else(|| {
+            vm.new_type_error(format!(
+                "can only assign string to {}.__doc__, not '{}'",
```
DimitrisJim, line 946: do we see this somewhere in CPython?

DimitrisJim:
I remember there's some weird issue with `__doc__`, at some point during class creation (I think, can't remember exactly), we overwrite it (i.e `__doc__` isn't actually set as the getset descriptor). 

Though the implementation look generally fine, I don't believe this is ever invoked. I *think* there was an issue for this someplace but can't recall now.

oow214:
https://github.com/python/cpython/blob/main/Objects/typeobject.c#L1054-L1091

If RustPython doesn't need this PR, I'll close it. 👀

youknowone:
@oow214 would you add a test fixed by this PR under `extra_tests/snippets`? that will be enough to validate the patch and helpful to prevent regression.

oow214:
Oh, sorry. I have confirmed that `__doc__` works fine without this PR.

PR #4241, opened 2022-10-22
Allow log2() to work on large integers

devonhollowood:
Part of #1671

My plan is to extend this to log() and log10() in the next PR — I wanted to see how you felt about the approach, first. Let me know what you think.

### DIFF
--- a/stdlib/src/math.rs
+++ b/stdlib/src/math.rs
@@ -157,7 +157,7 @@ mod math {
     }
 
     /// Generates the base-2 logarithm of a BigInt `x`
-    fn ilog2(x: &BigInt) -> f64 {
+    fn int_log2(x: &BigInt) -> f64 {
         // log2(x) = log2(2^n * 2^-n * x) = n + log2(x/2^n)
         // If we set 2^n to be the greatest power of 2 below x, then x/2^n is in [1, 2), and can
         // thus be converted into a float.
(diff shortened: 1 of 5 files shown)


devonhollowood:
It looks like the error is in `test_statistics.py`:

```
UNEXPECTED SUCCESS: test_regression_25177 (__main__.TestMean)
```

However, I'm not sure why this is suddenly passing, so I'm a little hesitant to remove the `@unittest.expectedFailure`. What do you think?

stdlib/src/math.rs
```diff
@@ -155,13 +156,35 @@ mod math {
         }
     }
 
+    /// Generates the base-2 logarithm of a BigInt `x`
+    fn ilog2(x: &BigInt) -> f64 {
```
youknowone, line 160:
```suggestion
    fn int_log2(x: &BigInt) -> f64 {
```

youknowone (changes requested):
Great! I left a naming style request.

Yes, removing the decorator is correct way. Maybe it was using log2 somewhere and you caught it.

stdlib/src/math.rs
devonhollowood: Done and done!

youknowone (approved):
Thank you for contributing!

Issue #4238, opened 2022-10-22
bug: Double Free On `BufferedReader`&`FileIO`?


discord9:
# Bug behavior
Added a `is_drop` field in PyInner. and found out both `BufferedReader`&`FileIO` gets double drop? Did some search on the code, no clue where misses a `clone()` to PyRef or PyObjectRef
# Possible root of bug
a obscure memcpy on a struct containing a PyRef/PyObjectRef
# Necessity of fixing the bug
Useful for writing garbage collecting, also prevent UB
# Part of the log
the remaining of log file just basically repeats those lines. The command is just `cargo run`, and alright it did run, but with all those double drop: 
[log_buf.log](https://github.com/RustPython/RustPython/files/9844712/log_buf.log)
<details>

```
[ERROR rustpython_vm::object::core] Double drop on PyRef, type="rustpython_vm::stdlib::io::_io::BufferedReader"
[ERROR rustpython_vm::object::core] Double drop on PyRef, type="rustpython_vm::stdlib::io::fileio::FileIO"
[ERROR rustpython_vm::object::core] Double drop on PyObjectRef, typeid=TypeId { t: 14576741135102236696 }, type=Some("rustpython_vm::stdlib::io::fileio::FileIO")
[ERROR rustpython_vm::object::core] Double drop on PyRef, type="rustpython_vm::stdlib::io::_io::BufferedReader"
                    backtrace=   0: <rustpython_vm::object::core::PyRef<T> as core::ops::drop::Drop>::drop
                 at vm/src/object/core.rs:979:17
       1: core::ptr::drop_in_place<rustpython_vm::object::core::PyRef<rustpython_vm::stdlib::io::_io::BufferedReader>>
                 at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/ptr/mod.rs:486:1
       2: rustpython_vm::stdlib::io::_io::BufferedMixin::close
                 at vm/src/stdlib/io.rs:1575:9
       3: core::ops::function::Fn::call
                 at /rustc/e092d0b6b43f2de967af0887873151bb1c0b18d3/library/core/src/ops/function.rs:77:5
       4: rustpython_vm::function::builtin::<impl rustpython_vm::function::builtin::sealed::PyNativeFuncInternal<(rustpython_vm::function::builtin::OwnedParam<T1>,),R,rustpython_vm::vm::VirtualMachine> for F>::call_
                 at vm/src/function/builtin.rs:79:17
[...]

PR #4242, opened 2022-10-23
feature: Double drop check in debug mode

discord9:
Add a double drop check field for PyInner, which:
- guaranteed not to have False positives，but might have false negative because after drop other things might alter `is_drop` field back to false, also might cause seg fault if others write even more exotic data in it after drop
- only print `[ERROR]` message, but doesn't panic
- only activate in debug mode
- print type name of corrsponding typeid(assuming `TypeId`'s SipHash impl doesn't collide with so few types)

### DIFF
--- a/vm/src/object/core.rs
+++ b/vm/src/object/core.rs
@@ -42,8 +42,8 @@ use std::{
 use once_cell::sync::Lazy;
 
 #[cfg(debug_assertions)]
-pub static ID2TYPE: Lazy<PyMutex<std::collections::HashMap<TypeId, String>>> =
-    Lazy::new(|| PyMutex::new(std::collections::HashMap::new()));
+pub static ID2TYPE: Lazy<std::sync::Mutex<std::collections::HashMap<TypeId, String>>> =
+    Lazy::new(|| std::sync::Mutex::new(std::collections::HashMap::new()));
 
 // so, PyObjectRef is basically equivalent to `PyRc<PyInner<dyn PyObjectPayload>>`, except it's
 // only one pointer in width rather than 2. We do that by manually creating a vtable, and putting
@@ -866,8 +866,12 @@ impl Drop for PyObjectRef {
             error!(
                 "Double drop on PyObjectRef with typeid={:?}(Type={:?})",
                 self.0.typeid,
-                ID2TYPE.lock().get(&self.0.typeid),
+                ID2TYPE
+                    .lock()
+                    .expect("can't read ID2TYPE")
+                    .get(&self.0.typeid),
             );
+            return;
         }
         if self.0.ref_count.dec() {
             #[cfg(debug_assertions)]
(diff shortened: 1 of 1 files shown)


vm/src/object/core.rs
```diff
@@ -849,7 +861,23 @@ impl<'a, T: PyObjectPayload> From<&'a Py<T>> for &'a PyObject {
 impl Drop for PyObjectRef {
     #[inline]
     fn drop(&mut self) {
```
youknowone, line 863: Because `PyObjectRef::drop` is very small, I'd like to suggest to add a whole debug drop , rather than adding a lot bigger debug features inside of the functions.

```rust
#[cfg(debug_assertions)]
fn drop(&mut self) {
    // debug drop here
}

#[cfg(not(debug_assertions))
#[inline]
fn drop(&mut self) {
    // production drop here
}
```

youknowone (changes requested):

youknowone (approved):

vm/src/object/core.rs
discord9: > Because `PyObjectRef::drop` is very small, I'd like to suggest to add a whole debug drop , rather than adding a lot bigger debug features inside of the functions.
> 
> ```rust
> #[cfg(debug_assertions)]
> fn drop(&mut self) {
>     // debug drop here
> }
> 
> #[cfg(not(debug_assertions))
> #[inline]
> fn drop(&mut self) {
>     // production drop here
> }
> ```

Doing the refactor right now, after I done refactor, should I amend&force push or create a new commit?

discord9: > Because `PyObjectRef::drop` is very small, I'd like to suggest to add a whole debug drop , rather than adding a lot bigger debug features inside of the functions.
> 
> ```rust
> #[cfg(debug_assertions)]
> fn drop(&mut self) {
>     // debug drop here
> }
> 
> #[cfg(not(debug_assertions))
> #[inline]
> fn drop(&mut self) {
>     // production drop here
> }
> ```

Also I refactor GC PR on top of this drop check, which, you can see, make drop() a much larger function, should it be spilted the same way?
```rust
impl Drop for PyObjectRef {
    #[inline]
    fn drop(&mut self) {
        #[cfg(debug_assertions)]
        if *self.0.is_drop.lock() {
            error!(
                "Double drop on PyObjectRef with typeid={:?}(Type={:?})",
                self.0.typeid,
                ID2TYPE
                    .lock()
                    .expect("can't read ID2TYPE")
                    .get(&self.0.typeid),
            );
            return;
        }
        #[cfg(feature = "gc")]
        {
            let stat = self.decrement();
            match stat {
                GcStatus::ShouldDrop => unsafe {
                    *self.0.is_drop.lock() = true;
                    PyObject::drop_slow(self.ptr);
                },
                GcStatus::ShouldDropOnly | GcStatus::BufferedDrop => unsafe {
                    *self.0.is_drop.lock() = true;
                    PyObject::drop_only(self.ptr);
                },
                GcStatus::ShouldKeep | GcStatus::DoNothing => (),
            }
        }
        #[cfg(not(feature = "gc"))]
        if self.0.ref_count.dec() {
            #[cfg(debug_assertions)]
            {
                *self.0.is_drop.lock() = true;
            }
            unsafe { PyObject::drop_slow(self.ptr) }
        }
    }
}
```

discord9:
I would suggest we first fix #4238 before merge this PR, or the error message can be rather annoying for every cargo run pop up dozens of those error now.

discord9:
Gonna to integrate drop check with gc, so closing this for now

PR #4248, opened 2022-10-25
Use improved log2() implementation for log10()

devonhollowood:
This also fixes a bug in the `log2()` implementation, where very large negative integer arguments would not raise an exception (and instead would turn into NaNs).

Follow-up to #4241. Part of #1671.

### DIFF
--- a/stdlib/src/math.rs
+++ b/stdlib/src/math.rs
@@ -132,18 +132,9 @@ mod math {
     }
 
     #[pyfunction]
-    fn log(x: ArgIntoFloat, base: OptionalArg<ArgIntoFloat>, vm: &VirtualMachine) -> PyResult<f64> {
-        let x = *x;
-        base.map_or_else(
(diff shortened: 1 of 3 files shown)


stdlib/src/math.rs
```diff
@@ -179,7 +170,12 @@ mod math {
             }
             Err(float_err) => {
                 if let Ok(x) = x.try_int(vm) {
-                    Ok(int_log2(x.as_bigint()))
+                    let x = x.as_bigint();
+                    if x > &BigInt::zero() {
```
youknowone, line 174:
```suggestion
                    if x.is_positive() {
```

youknowone (changes requested):
Great! I left a subtle change request about bigint.

DimitrisJim (approved):

DimitrisJim:
thanks!

PR #4249, opened 2022-10-25
Add `StopIteration` type to marshal

oow214:
Added to marshal to support StopIteration type and remove expectedFailure


DimitrisJim (approved):
lgtm, thanks!

PR #4251, opened 2022-10-26
Clean up a few obsolete TODOs in test_math.py

devonhollowood:
It looks like these were accidentally left in the codebase after #2773.


youknowone (approved):
Thank you!

PR #4252, opened 2022-10-27
Implement a more accurate hypot()

devonhollowood:
This uses the improved-accuracy hypotenuse algorithm from Borges 2019, see https://arxiv.org/abs/1904.09481.

Part of #1671.

### DIFF
--- a/stdlib/src/math.rs
+++ b/stdlib/src/math.rs
@@ -343,10 +343,11 @@ mod math {
             let correction = v
                 .iter()
                 .copied()
-                .map(|x| (x / scale).pow(2))
+                .map(|x| (x / scale).powi(2))
                 .chain(std::iter::once(-norm * norm))
+                // Pairwise summation of floats gives less rounding error than a naive sum.
                 .tree_fold1(std::ops::Add::add)
-                .unwrap();
+                .expect("expected at least 1 element");
             norm = norm + correction / (2.0 * norm);
         }
         norm * scale
--- a/stdlib/src/math.rs
+++ b/stdlib/src/math.rs
@@ -7,6 +7,7 @@ mod math {
         function::{ArgIntoFloat, ArgIterable, Either, OptionalArg, PosArgs},
         identifier, PyObject, PyObjectRef, PyRef, PyResult, VirtualMachine,
     };
+    use itertools::Itertools;
     use num_bigint::BigInt;
     use num_rational::Ratio;
     use num_traits::{One, Signed, ToPrimitive, Zero};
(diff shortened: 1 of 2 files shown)


stdlib/src/math.rs
```diff
@@ -283,24 +284,73 @@ mod math {
...
+            // needed, similar to that in `accurate_hypot()`.
+            // First, we estimate [sum of squares - norm^2], then we add the first-order
+            // approximation of the square root of that to `norm`.
+            let correction = v
+                .iter()
+                .cloned()
+                .map(|x| x / scale)
+                .map(|x| x * x)
+                .chain(std::iter::once(-norm * norm))
+                .tree_fold1(std::ops::Add::add)
```
DimitrisJim, line 349: Pretty sure this can just be a `reduce` since addition is associative.

DimitrisJim: Hm, scratch that, doesn't hold for floats. Is the tree-like reduce here required?

stdlib/src/math.rs
```diff
@@ -283,24 +284,73 @@ mod math {
...
+            // ^ This can be an `else if`, because if the max is near f64::MAX and the min is near
+            // f64::MIN_POSITIVE, then the min is relatively unimportant and will be effectively
+            // ignored.
+            min
+        } else {
+            1.0
+        };
+        let mut norm = v
+            .iter()
+            .cloned()
```
youknowone, line 334:
```suggestion
            .copied()
```

stdlib/src/math.rs
```diff
@@ -283,24 +284,73 @@ mod math {
...
+            .reduce(accurate_hypot)
+            .unwrap_or_default();
+        if v.len() > 2 {
+            // For larger lists of numbers, we can accumulate a rounding error, so a correction is
+            // needed, similar to that in `accurate_hypot()`.
+            // First, we estimate [sum of squares - norm^2], then we add the first-order
+            // approximation of the square root of that to `norm`.
+            let correction = v
+                .iter()
+                .cloned()
```
youknowone, line 345:
```suggestion
                .copied()
```

stdlib/src/math.rs
```diff
@@ -283,24 +284,73 @@ mod math {
...
+                .map(|x| x / scale)
+                .map(|x| x * x)
```
youknowone, line 347:
```suggestion
                .map(|x| (x / scale).pow(2))
```

stdlib/src/math.rs
```diff
@@ -283,24 +284,73 @@ mod math {
...
+                .chain(std::iter::once(-norm * norm))
+                .tree_fold1(std::ops::Add::add)
+                .unwrap();
```
youknowone, line 350: Is this infallible? than `expect()` with reason will better for debugging.

youknowone (changes requested):
Great! i left a few comments.

stdlib/src/math.rs
devonhollowood: Done; switched to `expect()`. This is infallible, because `tree_fold1()` fails if the iterator is empty, but this iterates over every element of `v` (which we have checked has a length of at least 2) plus an additional element for `-norm * norm`. So the minimum number of elements iterated over is 3.

stdlib/src/math.rs
devonhollowood: I just checked, and this actually passes all of the tests with just a simple `reduce()`. However, I think [pairwise summation](https://en.wikipedia.org/wiki/Pairwise_summation) is a little safer in general, since it typically results in less error for floats, and it has very little overhead. I added a comment to this effect.

That said, I'm happy to change this to a `reduce()` if you want, or to implement [Kahan summation](https://en.wikipedia.org/wiki/Kahan_summation_algorithm), which has even less error than pairwise summation. Let me know what you think.

DimitrisJim: I don't think the accuracy is checked for the `> 2` arg case in `testHypotAccuracy` (though the functionality is checked via `dist`).

No reason to change, completely satisfied with your explanation.

youknowone (approved):

PR #4253, opened 2022-10-28
Fix `repr` for union

oow214:


vm/src/builtins/union.rs
```diff
@@ -64,7 +64,9 @@ impl PyUnion {
                     .and_then(|o| o.downcast_ref::<PyStr>().map(|m| m.as_str().to_string())),
             ) {
                 (None, _) | (_, None) => Ok(obj.repr(vm)?.as_str().to_string()),
-                (Some(qualname), Some(module)) => Ok(if module == "builtins" {
+                (Some(qualname), Some(module)) => Ok(if qualname.eq("NoneType") {
```
DimitrisJim, line 67: glad to see you found where the actual bug was! Issue is, this isn't *exactly* the place :smile:. If you look at line `46` here, you'll notice it uses `vm.is_none(&obj)` which checks if the object is None (essentially `obj is None`).

Since `Union` holds the types though, this won't pass (`obj` will be `NoneType` and so `NoneType is None`) will be false.  Replacing it with `obj.is(vm.ctx.types.none_type)` should handle that case correctly.

DimitrisJim (approved):
Great! Thanks for the fix!

PR #4255, opened 2022-10-31
Fix memoryview iter in whats_left.py

tekknolagi:
`memoryview` has no `__iter__`

### DIFF
--- a/whats_left.py
+++ b/whats_left.py
@@ -200,7 +200,7 @@ def gen_methods():
         "type(range(0).__iter__())",
         "type(str().__iter__())",
         "type(tuple().__iter__())",
-        "type(memoryview(bytearray(b'0')).__iter__())",
+        "type(iter(memoryview(bytearray(b'0'))))",
     ]
 
     methods = {}


tekknolagi:
Hold on, I might have completely misread this and the error is something different.

DimitrisJim:
Hi @tekknolagi and thanks for the PR. You probably ran this with Python 3.9 which did not have a dedicated iterator. `memoryview.__iter__` was added in 3.10 (https://github.com/python/cpython/pull/22119) which we are currently following. 

It might make sense to add a check in `whats_left.py` to inform people that >=3.10 is expected (similarly to the two other checks for the implementation and the `-I` flag).

tekknolagi:
Oh! I can pivot to doing that instead.

fanninpm:
@DimitrisJim should we be advertising for Python 3.11, or is it too early to do that?

DimitrisJim:
Hey! Couldn't say with certainty, after we update some libs to 3.11 it might make sense.

DimitrisJim:
I see what you were trying to do here (fix the cron job for whatsleft.py?). 

In that case, this could be solved easily by specifying that python 3.10 be used for the job (see the `with` used in the equivalent job for generating the cov data).

PR #4257, opened 2022-11-02
Re-organizing lexer struct

bonjune:
I slightly made some changes to the lexer for better organization of the struct

- Changed `Vec<IndentationLevel>` into a new struct `Indentations` and limited its methods just for `pop`, `push`, `is_empty`, and `current`: Checking empty semantics is different from normal vectors in that `is_empty()` should return when the length is one, rather than zero because of the default indentation level.
- Made a struct for `chr0, chr1, chr2`, which is `CharWindow<const N: usize>`
- Changed condition expression in if statements. For example `if x == Some(1) || x == Some(2) { ... }` changed into `if matches!(x, Some(1 | 2))`
- Shortcircuited an if statement to reduce indentation level.
- Add a new test for `123_45_67_890` which was missing in `lexer.rs` tests module

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -147,11 +147,13 @@ where
 
 pub struct Lexer<T: Iterator<Item = char>> {
     chars: T,
+    window: CharWindow<3>,
+
     at_begin_of_line: bool,
     nesting: usize, // Amount of parenthesis
     indentations: Indentations,
+
     pending: Vec<Spanned>,
-    window: CharWindow<3>,
     location: Location,
 }
 
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -426,7 +426,7 @@ where
         loop {
             if let Some(c) = self.take_number(radix) {
                 value_text.push(c);
-            } else if self.window[1] == Some('_')
+            } else if self.window[0] == Some('_')
                 && Lexer::<T>::is_digit_of_radix(self.window[1], radix)
             {
                 self.next_char();
@@ -1392,7 +1392,7 @@ mod tests {
 
     #[test]
     fn test_numbers() {
-        let source = "0x2f 0b1101 0 123 0.2 2j 2.2j";
+        let source = "0x2f 0b1101 0 123 123_45_67_890 0.2 2j 2.2j";
         let tokens = lex_source(source);
         assert_eq!(
             tokens,
@@ -1408,6 +1408,9 @@ mod tests {
                 },
                 Tok::Int {
                     value: BigInt::from(123),
+                },
+                Tok::Int {
+                    value: BigInt::from(1234567890),
                 },
                 Tok::Float { value: 0.2 },
                 Tok::Complex {
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -111,11 +111,8 @@ struct CharWindow<const N: usize>([Option<char>; N]);
 
 impl<const N: usize> CharWindow<N> {
     fn slide(&mut self, next_char: Option<char>) {
-        let len = self.0.len();
-        for i in 0..(len - 1) {
-            self[i] = self[i + 1];
-        }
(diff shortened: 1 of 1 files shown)


compiler/parser/src/lexer.rs
```diff
@@ -56,15 +71,89 @@ impl IndentationLevel {
     }
 }
 
+#[derive(Debug)]
+struct Indentations {
+    indent_stack: Vec<IndentationLevel>,
+}
+
+impl Indentations {
+    pub fn is_empty(&self) -> bool {
```
DimitrisJim, line 80: don't think there's much point in pub functions on a private struct.

compiler/parser/src/lexer.rs
```diff
@@ -56,15 +71,89 @@ impl IndentationLevel {
...
+
+impl Default for Indentations {
+    fn default() -> Self {
+        Self {
+            indent_stack: vec![IndentationLevel::default()],
+        }
+    }
+}
+
+struct CharWindow<const N: usize>(pub [Option<char>; N]);
```
DimitrisJim, line 110: similarly for this `pub` here.

bonjune: Thx, I've changed those you pointed out!

compiler/parser/src/lexer.rs
```diff
@@ -56,15 +71,89 @@ impl IndentationLevel {
...
+}
+
+impl<const N: usize, Idx> IndexMut<Idx> for CharWindow<N>
+where
+    Idx: SliceIndex<[Option<char>], Output = Option<char>>,
+{
+    fn index_mut(&mut self, index: Idx) -> &mut Self::Output {
+        self.0.index_mut(index)
+    }
+}
```
youknowone, line 146:
```suggestion
impl<const N: usize> Default for CharWindow<N> {
    fn default() -> Self {
        Self([None; N])
    }
}
```
We will not need `IndexMut` by adding a proper method to CharWindow

compiler/parser/src/lexer.rs
```diff
@@ -11,10 +11,25 @@ use num_traits::identities::Zero;
...
+    make_tokenizer_located(source, Location::new(0, 0))
+}
+
+pub fn make_tokenizer_located(
+    source: &str,
+    start_location: Location,
+) -> impl Iterator<Item = LexResult> + '_ {
+    let nlh = NewlineHandler::new(source.chars());
+    Lexer::new(nlh, start_location)
+}
```
youknowone, line 31: Could you keep the position of these functions to keep the diff smaller?

youknowone (changes requested):
Thank you for contributing! I left a few comments.

bonjune:
Removing `IndexMut`, I added a `CharWindow::swap_first` because we need mutation for collapsing EOL characters.

bonjune:
I think there is still room for enhancement.

Whenever a struct uses `CharWindow`, it should implement its working mechanism with `input`(in `Lexer` struct) or `source` (in `NewlineHandler` struct), such as
* initializing the CharWindow and
* what value for `CharWindow` to slide with.

I think those mechanisms should be provided by `CharWindow` wholly. It means `CharWindow` should own `T: impl Iterator<Item = char>`.

bonjune:
I've made the changes I mentioned above. Sorry for requesting extra reviews.

youknowone (approved):
looks good, thank you!

youknowone:
oh, it has conflict. could you rebase it? 

@charliermarsh do you have any advice about this?

fanninpm:
@bonjune Could you please rebase this branch on a fresh copy of `main`?

charliermarsh:
I haven't reviewed this thoroughly, but I'm guessing the conflict is from #4266. The main changes therein were to ensure that we emit `Tok::Comment` when lexing comments.

bonjune:
I've just rebased. Works fine on my machine!

youknowone:
@charliermarsh oh, i'd like to ask you code review, but that was also a good feedback, thank you!

PR #4259, opened 2022-11-04
Fix errors from update to rust 1.65

DimitrisJim:


vm/src/stdlib/winreg.rs
```diff
@@ -191,7 +191,8 @@ mod winreg {
         key.with_key(|k| k.get_raw_value(subkey))
             .map_err(|e| e.to_pyexception(vm))
             .and_then(|regval| {
-                let ty = regval.vtype as usize;
+                #[allow(clippy::redundant_clone)]
+                let ty = regval.vtype.clone() as usize;
```
fanninpm, line 195: This smells like a bug somewhere else.

youknowone (approved):
The redundant clone part looks weird but go for it now

vm/src/stdlib/winreg.rs
DimitrisJim: seems like a clippy false positive tbh, I don't have windows so this was a pain to go through

PR #4260, opened 2022-11-04
Implement Sqlite3 Module

qingshi163:


qingshi163:
@youknowone Can you help with that compilation fail with android?

youknowone:
that looks like a toolchain problem. rather than fighting with it, how about just disabling sqlite3 for android?

stdlib/src/sqlite.rs
```diff
@@ -0,0 +1,2963 @@
...
+            if guard.is_some() {
+                Ok(PyMutexGuard::map(guard, |x| unsafe {
+                    x.as_mut().unwrap_unchecked()
+                }))
+            } else {
+                Err(new_programming_error(
+                    vm,
+                    "Cannot operate on a closed cursor.".to_owned(),
+                ))
+            }
```
youknowone, line 1375:
```suggestion
            let Some(guard) = self.inner.lock() else {
                return Err(new_programming_error(
                    vm,
                    "Cannot operate on a closed cursor.".to_owned(),
                ));
            };
            Ok(PyMutexGuard::map(guard, |x| unsafe {
                x.as_mut().unwrap_unchecked()
            }))
```

stdlib/src/sqlite.rs
```diff
@@ -0,0 +1,2963 @@
...
+            zelf
+        }
+
+        #[pymethod(magic)]
+        fn exit(&self, _args: FuncArgs) {
+            self.close()
+        }
+
+        fn inner(&self, vm: &VirtualMachine) -> PyResult<PyMappedMutexGuard<BlobInner>> {
+            let guard = self.inner.lock();
```
youknowone, line 2032: let-else for guard here

youknowone (approved):
Great! You made a huge work!

stdlib/src/sqlite.rs
```diff
@@ -0,0 +1,2963 @@
...
+            if guard.is_some() {
+                Ok(PyMutexGuard::map(guard, |x| unsafe {
+                    x.as_mut().unwrap_unchecked()
+                }))
+            } else {
+                Err(new_programming_error(
+                    vm,
+                    "Cannot operate on a closed database.".to_owned(),
+                ))
+            }
```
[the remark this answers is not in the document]
qingshi163: This is not going to work as the variable guard is MutexGuard<Option<_>>

stdlib/src/sqlite.rs
fanninpm: Can you do let-else on a `MutexGuard`?

PR #4261, opened 2022-11-04
Set comparator start location to beginning of comparison

charliermarsh:
Right now, `1 == 2` has a start location at the first `=`. This PR changes the parser such that start location is equivalent to the start of the `left` item (i.e., the `1`).


charliermarsh:
\cc @harupy

youknowone (approved):

PR #4262, opened 2022-11-05
Add support for printing to JIT

evilpie:
With thanks to https://zmedley.com/calling-rust.html for showing me how to do this.


jit/src/instructions.rs
```diff
@@ -403,7 +410,62 @@ impl<'a, 'b> FunctionCompiler<'a, 'b> {
...
+            Instruction::CallFunctionPositional { nargs } => {
+                if nargs != &1 {
+                    return Err(JitCompileError::NotSupported);
+                }
+
+                let arg1 = self.stack.pop().ok_or(JitCompileError::BadBytecode)?;
+                let function = self.stack.pop().ok_or(JitCompileError::BadBytecode)?;
+
+                match (arg1.ty, function.ty) {
+                    (JitType::Int, JitType::PrintFunction) => {
```
DimitrisJim, line 434: I generally haven't touched on the `jit` folder and have no real experience with cranelift. Is my understanding correct that what this PR allows is a subcase of the Python print, i.e: `print(some_int_value)`?

evilpie: Exactly. `print` with exactly one int argument works.

afonso360:
Hey, I've been thinking about adding something similar recently (function calls), but had some different ideas. 

My initial idea would be to add a `vmctx` arg that we pass through all JIT functions, that allows us to restore control back to the VM and call functions that way. This also means that we can load globals from the global list in the VM and return `PyRef` references back to the JIT'ed code that we can use later.

What do you think about that?

(The obvious downside is that this is WAYYYY slower than a direct function call as you are doing here, although we can probably specialize later)

jit/src/lib.rs
```diff
@@ -165,6 +180,7 @@ pub enum JitType {
     Int,
     Float,
     Bool,
+    PrintFunction,
```
afonso360, line 183: Would it make sense to represent this as `Function(enum JITFunction)` or something along those lines?

evilpie: I think that would make sense when adding a second builtin function. See also #4263.

jit/src/lib.rs
```diff
@@ -33,10 +34,19 @@ struct Jit {
     module: JITModule,
 }
 
+// TODO: print should return None
+pub fn print_fun(a: i64) -> i64 {
```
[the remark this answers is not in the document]
evilpie: Good catch. I think I might just put them in a different file/mod?

jit/src/lib.rs
afonso360: Yeah, sounds good 👍

evilpie:
> 

Definitely sounds like a good long term goal. I don't understand enough about how memory management in cranelift works to do this. Realistically most JITs also have both: 1. Fast native calls 2. Slower calls into the VM.

afonso360:
> Definitely sounds like a good long term goal. I don't understand enough about how memory management in cranelift works to do this. Realistically most JITs also have both: 1. Fast native calls 2. Slower calls into the VM

Right, just wanted to know if you had any thoughts on how to expand from here! I think this is a good start on getting external function calls.

If you need any help working on cranelift feel free to @ me, I'll try my best to help out. (Or ask in the wasmtime zulip!)

DimitrisJim:
Hey both! 

This sounds like something that could warrant an issue to hammer the details out. I have no familiarity with cranelift (yet!) but I'd follow your discussion to get a grip on things. It would be nice to see how other JITs tackle these things (built with or without cranelift) and what tradeoffs exist with each approach.

afonso360:
Yeah we should definitely do that! I'm really only familiar with how Wasmtime works, but I don't mind looking into a few more JIT's and writing a summary.

jit/src/lib.rs
```diff
@@ -33,10 +34,19 @@ struct Jit {
...
+    println!("PRINT: {a}");
```
youknowone, line 39: I am confused  here. would `print(1)` be compiled to print `PRINT: {a}` instead of `{a}`? Then it will not be compatible to pure python behavior.

afonso360:
> This sounds like something that could warrant an issue to hammer the details out. I have no familiarity with cranelift (yet!) but I'd follow your discussion to get a grip on things. It would be nice to see how other JITs tackle these things (built with or without cranelift) and what tradeoffs exist with each approach.

I've opened #4271 which turned out way bigger than I thought it was going to be.

Hopefully if we can pin that down we are then able to make a decision about this!

youknowone:
I am sorry. I tried to rebase this PR but not easy because many things changed.

PR #4263, opened 2022-11-05
Make JitValue an enum

evilpie:
The duplication with JitType is a bit unfortunate, but this is a big
ergonomics improvement when working on the compiler itself.

This will also allows us to define types that are internal to the
compiler and shouldn't leak to other code via the JitType

### DIFF
--- a/jit/src/instructions.rs
+++ b/jit/src/instructions.rs
@@ -21,14 +21,33 @@ struct Local {
 }
 
 #[derive(Debug)]
-struct JitValue {
-    val: Value,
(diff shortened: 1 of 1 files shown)


youknowone:
I'd like to approve the "Make JitValue an enum" commit, but not about printing support yet. If you rebase this PR only to include the latter commit, I will merge it. Otherwise once when #4262 is done.

evilpie:
Thanks, rebased it on top of main.

youknowone (approved):
Thank you!

Issue #4265, opened 2022-11-06
Lexer omits Python comments


charliermarsh:
CPython's `tokenize` module includes comments in the tokenized representation:

```py
>>> list(tokenize.generate_tokens(io.StringIO('x == y # foo').readline))
[TokenInfo(type=1 (NAME), string='x', start=(1, 0), end=(1, 1), line='x == y # foo'), TokenInfo(type=54 (OP), string='==', start=(1, 2), end=(1, 4), line='x == y # foo'), TokenInfo(type=1 (NAME), string='y', start=(1, 5), end=(1, 6), line='x == y # foo'), TokenInfo(type=61 (COMMENT), string='# foo', start=(1, 7), end=(1, 12), line='x == y # foo'), TokenInfo(type=4 (NEWLINE), string='', start=(1, 12), end=(1, 13), line=''), TokenInfo(type=0 (ENDMARKER), string='', start=(2, 0), end=(2, 0), line='')]
```

It would be really useful for Ruff if the RustPython token stream included comments.

I could help with this if there's agreement that it's desirable.

charliermarsh:
\cc @youknowone

PR #4266, opened 2022-11-07
Implement Tok::Comment

charliermarsh:
Resolves: #4265.


charliermarsh:
(Fixing tests.)

youknowone (dismissed):

charliermarsh:
There's something I don't understand in the parsing / compilation flow that's causing this to error. Need to spend a bit more time digging into it.

charliermarsh:
Good to go from my perspective.

DimitrisJim (approved):
lgtm too, no idea why the runner for mac hanged for a previous run there.

PR #4269, opened 2022-11-07
Implement a few more ops in the JIT

afonso360:
👋 Hey,

This PR implements a few more ops in the JIT.

I noticed that shifting by negative numbers raises an exception, I've modeled this as a trap, with the hopes that some day we'll catch that custom trap code and do proper exception handling.

Are there any other cases of special values in these ops that python treats differently?


youknowone (approved):
Great, this is looking a good base to extend jit part

youknowone:
@afonso360 Could this trap can be helpful to handle #2293 better?

PR #4273, opened 2022-11-10
Allow retry if `as_mut` fail when insert Dict

discord9:
This as_mut doesn't fail in most condition, although if you actually impl a garbage collect, there is a very small chances(three in 200,000, edited: more like 20 in 200000) that this following as_mut can fail due to change in dict:
https://github.com/RustPython/RustPython/blob/1a92399349704d2a49ae57539f308c5f91a1d436/vm/src/dictdatatype.rs#L252
```rust
/// Store a key
pub fn insert<K>(&self, vm: &VirtualMachine, key: &K, value: T) -> PyResult<()>
where K: DictKey + ?Sized,
{
    let hash = key.key_hash(vm)?;
    let _removed = loop {
        let (entry_index, index_index) = self.lookup(vm, key, hash, None)?;
        let mut inner = self.write();
        if let Some(index) = entry_index.index() {
            // Update existing key
            if let Some(entry) = inner.entries.get_mut(index) {
                // this `as_mut` might fail
                let entry = entry
                    .as_mut()
                    .expect("The dict was changed since we did lookup.");
                if entry.index == index_index {
                    let removed = std::mem::replace(&mut entry.value, value);
                    // defer dec RC
                    break Some(removed);
                } else {
                    // stuff shifted around, let's try again
                }
            } else {
                // The dict was changed since we did lookup. Let's try again.
            }
        } else {
            // New key:
            inner.unchecked_push(index_index, hash, key.to_pyobject(vm), value, entry_index);
            break None;
        }
    };
    Ok(())
}
```
so change to allow continue to loop and try to acquire a mut ref:
```rust
let entry = if let Some(entry) = entry.as_mut() {
    entry
} else {
    // The dict was changed since we did lookup. Let's try again.
    // this is very rare to happen
    // (and seems only happen with very high freq gc, and about three to four time in 200000 iters)
    // but still possible
    // so logging a warn is acceptable in here
    warn!("The dict was changed since we did lookup. Let's try again.");
    continue;
};
```

### DIFF
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -248,9 +248,17 @@ impl<T: Clone> Dict<T> {
             if let Some(index) = entry_index.index() {
                 // Update existing key
                 if let Some(entry) = inner.entries.get_mut(index) {
-                    let entry = entry
-                        .as_mut()
-                        .expect("The dict was changed since we did lookup.");
+                    let entry = if let Some(entry) = entry.as_mut() {
+                        entry
+                    } else {
+                        // The dict was changed since we did lookup. Let's try again.
+                        // this is very rare to happen
+                        // (and seems only happen with very high freq gc, and about three to four time in 200000 iters)
+                        // but still possible
+                        // so logging a warn is acceptable in here
+                        warn!("The dict was changed since we did lookup. Let's try again.");
+                        continue;
+                    };
                     if entry.index == index_index {
                         let removed = std::mem::replace(&mut entry.value, value);
                         // defer dec RC
--- a/vm/src/dictdatatype.rs
+++ b/vm/src/dictdatatype.rs
@@ -253,10 +253,8 @@ impl<T: Clone> Dict<T> {
                     } else {
                         // The dict was changed since we did lookup. Let's try again.
                         // this is very rare to happen
-                        // (and seems only happen with very high freq gc, and about three to four time in 200000 iters)
+                        // (and seems only happen with very high freq gc, and about one time in 10000 iters)
                         // but still possible
-                        // so logging a warn is acceptable in here
-                        warn!("The dict was changed since we did lookup. Let's try again.");
                         continue;
                     };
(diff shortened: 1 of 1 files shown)


discord9:
@DimitrisJim any advices?(Found this bug when trying to debug #4180)

DimitrisJim:
Unless someone else views this first, I'll have to look into it over the weekend when I have time. I don't think it is unreasonable to re-try here though.

discord9:
Just realize the rationale behind why it didn't occur when gc is not impled, for gc introduce more pause, hence more time to change the dict between lookup and actually insert

vm/src/dictdatatype.rs
```diff
@@ -248,9 +248,17 @@ impl<T: Clone> Dict<T> {
...
+                        entry
+                    } else {
+                        // The dict was changed since we did lookup. Let's try again.
+                        // this is very rare to happen
+                        // (and seems only happen with very high freq gc, and about three to four time in 200000 iters)
+                        // but still possible
+                        // so logging a warn is acceptable in here
+                        warn!("The dict was changed since we did lookup. Let's try again.");
+                        continue;
+                    };
```
DimitrisJim, line 261:
```suggestion
                    let Some(entry) = entry.as_mut() {
                        // The dict was changed since we did lookup. Let's try again.
                        // this is very rare to happen
                        // (and seems only happen with very high freq gc, and about three to four time in 200000 iters)
                        // but still possible
                        continue;
                    };
```

I don't think a warning is really needed here since we're just going to continue (person receiving warning can't do much about it). Same thing happens in all other cases where something changes the dict.

Do you need the warn for debugging purposes with the gc?

DimitrisJim (changes requested):

vm/src/dictdatatype.rs
discord9: Not really, I guess we could remove the warning here, doing it now.

DimitrisJim (approved):

Issue #4124, opened 2022-08-22
[RFC] 


Gelox:
## Summary

Provide an example for how to provide a rust object as an argument to python from rust and how to call a rust function with a python object argument from python.

## Detailed Explanation

Currently it is unclear to me how one would
1) Call a python function and provide a rust object as an argument to it except for creating a `String` representation of the Python object and compiling it, running some code that initializes it as a local variable and then calls the function with the local variable right after. If this is the idiomatic way then an example would be nice to clarify that.
2) Call a rust function from python, especially when using complicated types like channels in order to communicate with rust.

Hopefully this is possible, I'd be willing to create a PR for the examples if someone tells me understand.

fanninpm:
Interoperability with PyO3 is something we should look at in the future. See also #3016.

youknowone:
I think PyO3 is not related to the question.

Our wasm crate is a sort of demonstration of what you asked. But we don't have small example code for that. we need to make one.

Gelox:
Hi, I'm trying to find the demonstration of this in the wasm crate but I'm not finding it.

In `/wasm/demo/src/` there is one `.rs` file which contains 0 bytes.

I might be misunderstanding what I'm looking for but I'm thinking a really basic example of 1) a rust function calling into a python function with a rust type as an argument and 2) a python function calling into a rust function with a python type as an argument.

If someone could direct me to somewhere I could ask how to do this, or if you guys want to give me a small example then I could make a PR with this small example.

Gelox:
@youknowone Not sure if you are busy but could you point me in the right direction?

youknowone:
Will you check `#[pymodule]` in `wasm/lib/js_module.rs`?
I want to write a small example to write their own `#[pymodule]` and bidirectional communication under `exmaples` soon.

Gelox:
Cool, this looks interesting. I will take a look and see if I can understand it and if I can I will make a PR for an example.

youknowone:
we now have `example_projects` directory

PR #4276, opened 2022-11-14
Add example for calling between rust and python

Gelox:
Add an example which illustrates how to call between rust and python. Most importantly there no examples which illustrate how to call rust from python and it is not obvious how to do this.

I never figured out if there was a way to nicely call a constructor of a rust struct since all methods in python require the self parameter I assumed that this was not possible and that construction should be done via a standalone function.
If it is encouraged or idiomatic to use a python constructor from rust for some reason then perhaps that should be added to the example.

Previous opened issue: https://github.com/RustPython/RustPython/issues/4124

### DIFF
--- a/examples/call_between_rust_and_python.rs
+++ b/examples/call_between_rust_and_python.rs
@@ -1,7 +1,5 @@
 use rustpython_vm::{
-    builtins::PyStr,
-    function::{FuncArgs, KwArgs, PosArgs},
-    pyclass, pymodule, PyObject, PyObjectRef, PyPayload, PyResult, TryFromBorrowedObject,
+    pyclass, pymodule, PyObject, PyPayload, PyResult, TryFromBorrowedObject,
     VirtualMachine,
 };
 
@@ -19,19 +17,10 @@ pub fn main() {
 
         let module = vm.import("call_between_rust_and_python", None, 0).unwrap();
         let init_fn = module.get_attr("python_callback", vm).unwrap();
-
         vm.invoke(&init_fn, ()).unwrap();
 
-        let pystr = PyObjectRef::from(PyStr::new_ref(
-            unsafe {
-                PyStr::new_ascii_unchecked(String::from("Rust string sent to python").into_bytes())
-            },
-            vm.as_ref(),
-        ));
-        let take_string_args = FuncArgs::new(PosArgs::new(vec![pystr]), KwArgs::default());
         let take_string_fn = module.get_attr("take_string", vm).unwrap();
-
-        vm.invoke(&take_string_fn, take_string_args).unwrap();
+        vm.invoke(&take_string_fn, (String::from("Rust string sent to python"),)).unwrap();
     })
 }
 
@@ -43,8 +32,8 @@ mod rust_py_module {
     fn rust_function(
         num: i32,
         s: String,
-        python_person: PyObjectRef,
-        vm: &VirtualMachine,
+        python_person: PythonPerson,
+        _vm: &VirtualMachine,
     ) -> PyResult<RustStruct> {
         println!(
             "Calling standalone rust function from python passing args:
@@ -53,7 +42,7 @@ string: {},
 python_person.name: {}",
             num,
             s,
-            python_person.try_into_value::<PythonPerson>(vm).unwrap().name
+            python_person.name
         );
         Ok(RustStruct)
     }
@@ -78,10 +67,8 @@ python_person.name: {}",
     impl TryFromBorrowedObject for PythonPerson {
         fn try_from_borrowed_object(vm: &VirtualMachine, obj: &PyObject) -> PyResult<Self> {
             let name = obj
-                .get_attr("name", vm)
-                .unwrap()
-                .try_into_value::<String>(vm)
-                .unwrap();
+                .get_attr("name", vm)?
+                .try_into_value::<String>(vm)?;
             Ok(PythonPerson { name })
         }
     }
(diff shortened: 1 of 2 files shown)


examples/call_between_rust_and_python.rs
```diff
@@ -0,0 +1,88 @@
...
+
+#[pymodule]
+mod rust_py_module {
+    use super::*;
+
+    #[pyfunction]
+    fn rust_function(
+        num: i32,
+        s: String,
+        python_person: PyObjectRef,
```
DimitrisJim, line 46: could just be `PythonPerson`, the conversion will happen automatically.

examples/call_between_rust_and_python.rs
```diff
@@ -0,0 +1,88 @@
...
+
+        vm.invoke(&init_fn, ()).unwrap();
+
+        let pystr = PyObjectRef::from(PyStr::new_ref(
+            unsafe {
+                PyStr::new_ascii_unchecked(String::from("Rust string sent to python").into_bytes())
+            },
+            vm.as_ref(),
+        ));
+        let take_string_args = FuncArgs::new(PosArgs::new(vec![pystr]), KwArgs::default());
```
DimitrisJim, line 31: you can remove these and just use `vm.invoke(&take_string_fn, (pystr,))` where `pystr` is just the `String`,

examples/call_between_rust_and_python.rs
```diff
@@ -0,0 +1,88 @@
...
+        name: String,
+    }
+
+    impl TryFromBorrowedObject for PythonPerson {
+        fn try_from_borrowed_object(vm: &VirtualMachine, obj: &PyObject) -> PyResult<Self> {
+            let name = obj
+                .get_attr("name", vm)
+                .unwrap()
+                .try_into_value::<String>(vm)
+                .unwrap();
```
DimitrisJim, line 84:
```suggestion
            let name = obj.get_attr("name", vm)?.try_into_value(vm)?;
```

Gelox:
Is there a way that you can have a `String` with python code instead of having a separate file in the OS? How would I import this instead of the file?

examples/call_between_rust_and_python.rs
Gelox: Done.

examples/call_between_rust_and_python.rs
Gelox: Is this possible with every primitive and rust struct marked as a `pyclass`? Just out of curiosity why is `(pystr)` disallowed but `(pystr,)` allowed?

Done either way.

examples/call_between_rust_and_python.rs
Gelox: Done.

examples/call_between_rust_and_python.rs
fanninpm: > Just out of curiosity why is `(pystr)` disallowed but `(pystr,)` allowed?

[This can explain the difference better than I can.](https://dtolnay.github.io/rust-quiz/29)

youknowone:
You might set git email wrong (not same as your github id email)

youknowone:
to use `import`, this is the best way. consider `vm.compile` and `vm.run_code_obj` to avoid external file, but this is not 100% compliant with `import`.

Gelox:
I fixed the formatting.

> You might set git email wrong (not same as your github id email)

Yeah, I'm intentionally setting the email to a random one because I don't like having the other one associated with my account. However git refuses to not use the global config which causes me to forget and have to amend the email with a force push.

I hope it's not important that the email is set the same as my github email?

> to use import, this is the best way. consider vm.compile and vm.run_code_obj to avoid external file, but this is not 100% compliant with import.

Alright, then I'll leave the import business be as it is.

youknowone:
It will be ok if the email is real, but I think this is not ideal

```
commit f0faccfe46852552027bd1e43e3c95b2e763aa85 (HEAD -> Gelox/main)
Author: q <q@q.q>
Date:   Wed Nov 16 12:06:23 2022 +0100

    Format
```

You can set local name and email which doesn't affect global config using `git config user.email <emailaddress>`

Gelox:
Alright, I will update the email address to a real one.

I would like to put one more thing into the example which I realize is currently missing, which is to be able to use rust struct fields from inside of python.

Something like:
rust:
```rust
#[pyattr]
#[pyclass(module = "rust_py_module", name = "RustStruct")]
struct RustStruct {
    pub number: f32
}
```

python:
```python
from rust_py_module import RustStruct

def python_func():
    rust_struct = rust_function()
    print(rust_struct.number)
```

Currently this is not working and I'm not sure how to accomplish it or if it is even possible, but I don't see any obvious reason for why it wouldn't be possible.

youknowone:
That's not enough. Adding getset descriptor is necessary.

```rust
impl RustStruct {
    #[pygetset]
    fn number(&self) -> f32 {
        self.number
    }
}
```

we'd like to support member descriptor like below one day:
```
#[pyattr]
#[pyclass(module = "rust_py_module", name = "RustStruct")]
struct RustStruct {
    #[pymember]
    pub number: f32
}
```
but not yet.

cc @moreal

Gelox:
The direct member access is included and I've corrected the email to an existing one that I own. Let me know if there are any further changes you want (I had to leave that `Vec::clone()` in there, couldn't figure out how to get around it).

youknowone:
awesome! thank you!

DimitrisJim (approved):

DimitrisJim:
Thanks @Gelox!

m172021:
I am running into similar issues. I am looking at:

https://github.com/RustPython/RustPython/blob/main/examples/call_between_rust_and_python.rs#L65-L83

Here is my question: on the Rust side, I have a RustStruct. Question: how do I bind this in the vm so python code can access this RustStruct ?

Do I add it via VM, goobal scope? .... ?

PR #4277, opened 2022-11-17
doc: add greptimedb's use case to README

discord9:
GreptimeDB use RsutPython for simple embedded scripting, so I think we can add this use case here?: https://docs.greptime.com/developer-guide/datanode/python-coprocessor


DimitrisJim (approved):
seems like a good idea to me.

PR #4281, opened 2022-11-19
change itertools::zip to std::iter::zip because deprecated

MisileLab:


DimitrisJim (approved):
thanks! we probably need to update itertools since this was a `0.10.4` deprecation.

PR #4283, opened 2022-11-20
Implement some minor performance optimizations

charliermarsh:
I tend to benchmark the parser by parsing all of CPython in [Ruff](https://github.com/charliermarsh/ruff). This involves (1) disabling Ruff's linting pass, then (2) running `cargo build --release && hyperfine --warmup 10 --runs 50 -i "./target/release/ruff resources/test/cpython --no-cache"`.

Prior to this change:

```
Benchmark 1: ./target/release/ruff resources/test/cpython --no-cache
  Time (mean ± σ):     265.2 ms ±   4.6 ms    [User: 2260.7 ms, System: 83.1 ms]
  Range (min … max):   259.0 ms … 281.1 ms    50 runs

Benchmark 1: ./target/release/ruff resources/test/cpython --no-cache
  Time (mean ± σ):     268.4 ms ±   4.2 ms    [User: 2268.1 ms, System: 83.5 ms]
  Range (min … max):   262.5 ms … 283.6 ms    50 runs

Benchmark 1: ./target/release/ruff resources/test/cpython --no-cache
  Time (mean ± σ):     269.7 ms ±   5.5 ms    [User: 2277.9 ms, System: 83.5 ms]
  Range (min … max):   260.1 ms … 280.7 ms    50 runs
```

After this change:

```
Benchmark 1: ./target/release/ruff resources/test/cpython --no-cache
  Time (mean ± σ):     255.6 ms ±   5.1 ms    [User: 2156.0 ms, System: 83.4 ms]
  Range (min … max):   247.2 ms … 271.8 ms    50 runs

Benchmark 1: ./target/release/ruff resources/test/cpython --no-cache
  Time (mean ± σ):     255.4 ms ±   4.8 ms    [User: 2152.8 ms, System: 83.0 ms]
  Range (min … max):   246.9 ms … 269.7 ms    50 runs

Benchmark 1: ./target/release/ruff resources/test/cpython --no-cache
  Time (mean ± σ):     256.3 ms ±   4.8 ms    [User: 2156.3 ms, System: 82.8 ms]
  Range (min … max):   248.5 ms … 270.6 ms    50 runs
```

So it's a ~4.8% improvement, if you take the median results.

Note that there's about ~58ms of overhead associated with this benchmark (e.g., to locate all the Python files on-disk, read their contents, etc.), so that's really a lower-bound on improvement.

### DIFF
--- a/compiler/parser/src/function.rs
+++ b/compiler/parser/src/function.rs
@@ -1,7 +1,6 @@
 use crate::ast;
 use crate::error::{LexicalError, LexicalErrorType};
-use ahash::RandomState;
-use std::collections::HashSet;
+use fnv::FnvHashSet;
 
 pub struct ArgumentList {
     pub args: Vec<ast::Expr>,
@@ -54,7 +53,8 @@ pub fn parse_args(func_args: Vec<FunctionArgument>) -> Result<ArgumentList, Lexi
     let mut args = vec![];
     let mut keywords = vec![];
 
-    let mut keyword_names = HashSet::with_capacity_and_hasher(func_args.len(), RandomState::new());
+    let mut keyword_names =
+        FnvHashSet::with_capacity_and_hasher(func_args.len(), Default::default());
     for (name, value) in func_args {
         match name {
             Some((start, end, name)) => {
--- a/compiler/parser/src/function.rs
+++ b/compiler/parser/src/function.rs
@@ -1,6 +1,6 @@
 use crate::ast;
 use crate::error::{LexicalError, LexicalErrorType};
-use fnv::FnvHashSet;
+use rustc_hash::FxHashSet;
 
 pub struct ArgumentList {
     pub args: Vec<ast::Expr>,
@@ -54,7 +54,7 @@ pub fn parse_args(func_args: Vec<FunctionArgument>) -> Result<ArgumentList, Lexi
     let mut keywords = vec![];
 
     let mut keyword_names =
-        FnvHashSet::with_capacity_and_hasher(func_args.len(), Default::default());
+        FxHashSet::with_capacity_and_hasher(func_args.len(), Default::default());
     for (name, value) in func_args {
         match name {
             Some((start, end, name)) => {
--- a/compiler/parser/src/string.rs
+++ b/compiler/parser/src/string.rs
@@ -15,7 +15,7 @@ pub fn parse_strings(
     let initial_kind = (values[0].1 .1 == StringKind::U).then(|| "u".to_owned());
 
     // Optimization: fast-track the common case of a single string.
-    if values.len() == 1 && matches!(&values[0].1 .1, StringKind::Normal | StringKind::U) {
+    if matches!(&*values, [(_, (_, StringKind::Normal | StringKind::U), _)]) {
         let value = values.into_iter().last().unwrap().1 .0;
         return Ok(Expr::new(
             initial_start,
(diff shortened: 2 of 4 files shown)


compiler/parser/src/function.rs
```diff
@@ -54,7 +53,8 @@ pub fn parse_args(func_args: Vec<FunctionArgument>) -> Result<ArgumentList, Lexi
     let mut args = vec![];
     let mut keywords = vec![];
 
-    let mut keyword_names = HashSet::with_capacity_and_hasher(func_args.len(), RandomState::new());
+    let mut keyword_names =
+        FnvHashSet::with_capacity_and_hasher(func_args.len(), Default::default());
```
DimitrisJim, line 57: does using `fnv` make such a difference? Reading on it the benefit seems to be small inputs (of which keyword args commonly are probably). As an aside, [rustc-hash](https://crates.io/crates/rustc-hash) is another option (and supposedly faster).

charliermarsh: This change alone (that is, excluding the changes in `string.rs`) drops execution time to ~263ms from ~268ms. So the two changes contribute close to equally, at least in my benchmark.

`rustc-hash` seems to be a tiny bit faster, even:

```
Benchmark 1: ./target/release/ruff resources/test/cpython --no-cache
  Time (mean ± σ):     254.2 ms ±   3.8 ms    [User: 2149.0 ms, System: 82.6 ms]
  Range (min … max):   248.8 ms … 266.1 ms    50 runs

Benchmark 1: ./target/release/ruff resources/test/cpython --no-cache
  Time (mean ± σ):     254.9 ms ±   4.8 ms    [User: 2164.6 ms, System: 82.4 ms]
  Range (min … max):   245.5 ms … 266.9 ms    50 runs

Benchmark 1: ./target/release/ruff resources/test/cpython --no-cache
  Time (mean ± σ):     254.8 ms ±   4.6 ms    [User: 2159.1 ms, System: 83.1 ms]
  Range (min … max):   246.5 ms … 269.4 ms    50 runs
```

compiler/parser/src/string.rs
```diff
@@ -14,6 +14,19 @@ pub fn parse_strings(
     let initial_end = values[0].2;
     let initial_kind = (values[0].1 .1 == StringKind::U).then(|| "u".to_owned());
 
+    // Optimization: fast-track the common case of a single string.
+    if values.len() == 1 && matches!(&values[0].1 .1, StringKind::Normal | StringKind::U) {
```
coolreader18, line 18: If you wanted, you could do `matches!(&*values, [(_, (_, StringKind::Normal | StringKind::U), _)])` for this whole thing (or the equivalent with if-let). I find that using patterns is generally clearer for understanding the structures of tuples over using indices.

coolreader18 (approved):
This looks good, thank you! I had no idea ruff used rustpython-parser, that's so cool!!

compiler/parser/src/string.rs
charliermarsh: Nice suggestion, thanks! (I still can't figure out how to take ownership of the `String` itself without the weird `values.into_iter()`, so I couldn't `if-let`.)

charliermarsh:
@coolreader18 - Definitely! We've used it from the very beginnings of Ruff. I've tried to contribute back as many compatibility improvements as I can :)

youknowone (approved):

PR #4286, opened 2022-11-22
Add Supporting for PyAtomicRef<Option<_>> and PyAtomicRef<PyObject>

qingshi163:


vm/src/object/ext.rs
```diff
@@ -214,7 +214,10 @@ impl<T: PyPayload> ToPyObject for PyRefExact<T> {
     }
 }
 
-pub struct PyAtomicRef<T: PyObjectPayload>(PyAtomic<*mut Py<T>>);
+pub struct PyAtomicRef<T> {
+    inner: PyAtomic<*mut i8>,
```
youknowone, line 218:
```suggestion
    inner: PyAtomic<*mut u8>,
```
it doesn't look important, but i've usually seen other code use u8 for this case. (to express `void*`)

vm/src/object/ext.rs
```diff
@@ -257,6 +269,88 @@ impl<T: PyObjectPayload> PyAtomicRef<T> {
...
+    }
+
+    pub fn swap_to_temporary_refs(&self, opt_ref: Option<PyRef<T>>, vm: &VirtualMachine) {
+        let old = unsafe { self.swap(opt_ref) };
+        if let Some(old) = old {
+            if let Some(frame) = vm.current_frame() {
+                frame.temporary_refs.lock().push(old.into());
+            }
+        }
+    }
```
youknowone, line 308: a tiny style suggestion 
```suggestion
        let Some(old) = unsafe { self.swap(opt_ref) } else {
            return;
        };
        if let Some(frame) = vm.current_frame() {
            frame.temporary_refs.lock().push(old.into());
        }
    }
```

youknowone (approved):

PR #4288, opened 2022-11-22
Refactor lexer functions

yt2b:
I refactored following lexer functions.
* `lex_number`
* ~~`take_number`~~
* `at_exponent`
* `is_identifier_continuation`

What I have done are as follows.
* using match expression instead of if expression
* reducing the nesting of match expression

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -438,9 +438,10 @@ where
     fn take_number(&mut self, radix: u32) -> Option<char> {
         let take_char = Lexer::<T>::is_digit_of_radix(self.window[0], radix);
 
-        match take_char {
-            true => Some(self.next_char().unwrap()),
-            _ => None,
+        if take_char {
+            Some(self.next_char().unwrap())
+        } else {
+            None
         }
     }
 
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -293,27 +293,26 @@ where
     /// Numeric lexing. The feast can start!
     fn lex_number(&mut self) -> LexResult {
         let start_pos = self.get_pos();
-        if self.window[0] == Some('0') {
-            if matches!(self.window[1], Some('x' | 'X')) {
+        match (self.window[0], self.window[1]) {
+            (Some('0'), Some('x' | 'X')) => {
                 // Hex! (0xdeadbeef)
                 self.next_char();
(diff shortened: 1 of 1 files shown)


youknowone:
@bonjune could you review this PR?

bonjune:
> @bonjune could you please review this PR?

I'll look into it today.

compiler/parser/src/lexer.rs
```diff
@@ -439,10 +438,9 @@ where
     fn take_number(&mut self, radix: u32) -> Option<char> {
         let take_char = Lexer::<T>::is_digit_of_radix(self.window[0], radix);
 
-        if take_char {
-            Some(self.next_char().unwrap())
-        } else {
-            None
+        match take_char {
+            true => Some(self.next_char().unwrap()),
+            _ => None,
```
bonjune, line 443: I like the idea of pattern matching, but if expression should be preferred to match expression when we are dealing with a boolean value.

bonjune (changes requested):
I like how the pattern matching here improves readability.
But, as I left a comment, if expression should be preferred for boolean values.

compiler/parser/src/lexer.rs
yt2b: I'll restore if expression in `take_number`

yt2b:
@bonjune 
Code fix completed.

bonjune (approved):
Good!
Have a good day. Thank you.

compiler/parser/src/lexer.rs
DimitrisJim: note that this could probably use [`bool.then`](https://doc.rust-lang.org/std/primitive.bool.html#method.then)

yt2b: Thank you for the adivse.

yt2b:
@bonjune 
Could you please review again?

bonjune (approved):
Looks good.

P.S.

https://doc.rust-lang.org/std/primitive.bool.html#method.then_some

I read bool's two method implementation: `bool::then` and `bool::then_some`, whose behaviors are subtly different. The key thing is here: the former lazily evaluated the passed closure while the latter one evaluates passed value eagerly. In this case, `bool::then` is a proper choice because we don't to evaluate the function only when `take_char` is true. (And you made the right decision, I think)

cc @youknowone

youknowone (approved):
Thank you!

Issue #4289, opened 2022-11-23
[RFC] Remove or revise outdated profiling document


youknowone:
## Summary

I never could run this: https://github.com/RustPython/RustPython/blob/main/DEVELOPMENT.md#profiling
It looked very slow and seemed never ended up.

PR #4290, opened 2022-11-23
Fix issues with flame-it.

DimitrisJim:
Running flame it works now as described in the profiling section:

```bash
echo "print('hello!')" >> script.py
cargo run --release --features flame-it script.py
```

or:

```bash
cargo run --release --features flame-it -- -c "print('hello world')"
```

to generate the flamescope (by default) which can be uploaded to speedscope.

ref #4289

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -43,6 +43,9 @@ extern crate env_logger;
 #[macro_use]
 extern crate log;
 
+#[cfg(feature = "flame-it")]
+use vm::Settings;
+
 mod interpreter;
 mod settings;
 mod shell;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -244,6 +193,57 @@ fn run_rustpython(vm: &VirtualMachine, run_mode: RunMode) -> PyResult<()> {
...
+        main_guard.end();
+        if let Err(e) = write_profile(&vm.state.as_ref().settings) {
+            error!("Error writing profile information: {}", e);
+        }
+    }
+    Ok(())
+}
+
+#[cfg(feature = "flame-it")]
+fn write_profile(settings: &Settings) -> Result<(), Box<dyn std::error::Error>> {
```
DimitrisJim, line 207: Moved this further down, right above the tests so it doesn't obstruct the rest of the code.

youknowone (approved):

PR #4293, opened 2022-11-24
depend: Update crossbeam-utils to "0.8.9" instead of "=0.8.9"

discord9:
Because it require `Lazy` for static (essentially global variable) and after 0.8.9, `crossbeam-utils` 's `AtomicCell` can't be use in static anymore, so refactor to wrap them in a `Lazy`


qingshi163:
This is adding a overhead for each visit witch is not neccesasy.
Is crossbeam denied to use AtomicCell with static by reason? should we create a pr to crossbeam?

discord9:
> This is adding a overhead for each visit witch is not neccesasy. Is crossbeam denied to use AtomicCell with static by reason? should we create a pr to crossbeam?

According to their comment(https://docs.rs/crossbeam-utils/0.8.10/src/crossbeam_utils/atomic/atomic_cell.rs.html#33-47)  in 0.8.10 of `crossbeam-utils`, yes it seems to have a reason
```
    /// Using MaybeUninit to prevent code outside the cell from observing partially initialized state:
    /// <https://github.com/crossbeam-rs/crossbeam/issues/833>
```
They add a MaybeUninit starting after 0.8.9, this is link to 0.8.9 for comparsion:
https://docs.rs/crossbeam-utils/0.8.9/src/crossbeam_utils/atomic/atomic_cell.rs.html#33-40

qingshi163:
Ok I see, due to MaybeUninit now crossbeam::AtomicCell has to implement Drop.

Can you create a RFC so we can optimize it in the future?

discord9:
> https://docs.rs/crossbeam-utils/0.8.10/src/crossbeam_utils/atomic/atomic_cell.rs.html#33-47

How about this RFC?
https://github.com/RustPython/RustPython/issues/4295

Issue #4297, opened 2022-11-27
Use bitflags in `Symbol`


yt2b:
https://github.com/RustPython/RustPython/blob/ec3648981dbf165321f17eb2cee64d5fcb7b16fd/compiler/codegen/src/symboltable.rs#L104-L107

As written in TODO comment, it's better way to use bitflags.
Following source code is also needed to fix.

* `codegen/src/compile.rs`
* `vm/src/stdlib/symtable.rs`

PR #4298, opened 2022-11-28
Use bitflags in `Symbol`

yt2b:
#4297
What I have done are as follows.

* Adding bitflags in rustpython-codgen crate.
* Replacing bool variable with bitflags in Symbol.

### DIFF
--- a/compiler/codegen/src/symboltable.rs
+++ b/compiler/codegen/src/symboltable.rs
@@ -1178,38 +1178,33 @@ impl SymbolTableBuilder {
         let symbol = if let Some(symbol) = table.symbols.get_mut(name.as_ref()) {
             // Role already set..
             match role {
-                SymbolUsage::Global => {
-                    if !symbol.is_global() {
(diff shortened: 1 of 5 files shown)


compiler/codegen/src/symboltable.rs
```diff
@@ -158,7 +154,7 @@ impl Symbol {
     }
 
     pub fn is_bound(&self) -> bool {
-        self.is_assigned || self.is_parameter || self.is_imported || self.is_iter
+        !(self.flags & SymbolFlags::BOUND).is_empty()
```
DimitrisJim, line 157:
```suggestion
        self.flags.intersects(SymbolFlags::BOUND)
```

is easier to read than the negation on empty here.

yt2b: Code fix completed.

DimitrisJim (approved):
lgtm, thanks for the find and PR!

PR #4300, opened 2022-11-28
rustpython unclared crate error so fixed it

MisileLab:


fanninpm:
What happens when you run `cargo clean`?

MisileLab:
> What happens when you run `cargo clean`?

it works thank you!

PR #4309, opened 2022-12-05
Bump openssl

DimitrisJim:
The test runner for ubuntu was recently bumped from 20.04 to 22.04, this appears to have caused some linking errors in openssl. Bumping it to the most recent version appears to clear things up.


DimitrisJim:
Okay, so it appears OpenSSL v3.02 which comes with Ubuntu 22.04 isn't exactly compatible with v1.1.1 which the `ssl` module requires. Either the action should pin to Ubuntu `20.04` or we should install OpenSSL 1.1.1 on Ubuntu `22.04`.

PR #4310, opened 2022-12-05
Fix nightly clippy warnings

youknowone:


youknowone:
false fix suggestion is reported here: https://github.com/rust-lang/rust-clippy/issues/9893

PR #4323, opened 2022-12-10
pattern matching parser support 

youknowone:
cherry-picked from #3427 

cc @Snowapril @charliermarsh

### DIFF
--- a/Lib/ast.py
+++ b/Lib/ast.py
@@ -53,10 +53,12 @@ def parse(source, filename='<unknown>', mode='exec', *,
 
 def literal_eval(node_or_string):
     """
-    Safely evaluate an expression node or a string containing a Python
+    Evaluate an expression node or a string containing only a Python
     expression.  The string or node provided may only consist of the following
     Python literal structures: strings, bytes, numbers, tuples, lists, dicts,
     sets, booleans, and None.
+
+    Caution: A complex expression can overflow the C stack and cause a crash.
     """
     if isinstance(node_or_string, str):
         node_or_string = parse(node_or_string.lstrip(" \t"), mode='eval')
@@ -234,6 +236,12 @@ def increment_lineno(node, n=1):
     location in a file.
     """
     for child in walk(node):
+        # TypeIgnore is a special case where lineno is not an attribute
+        # but rather a field of the node itself.
+        if isinstance(child, TypeIgnore):
+            child.lineno = getattr(child, 'lineno', 0) + n
+            continue
+
         if 'lineno' in child._attributes:
             child.lineno = getattr(child, 'lineno', 0) + n
         if (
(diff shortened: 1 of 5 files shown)


charliermarsh:
This would be _huge_ for Ruff.

youknowone:
ah, its parser part was not that much ready and I need to find how to add a soft-keyword concept to parser.

pauloxnet:
I hope this can help Ruff to support pattern matching and close the issue https://github.com/charliermarsh/ruff/issues/282

charliermarsh:
The dream...

charliermarsh:
Sorry to double-post, but copying over my most recent comment from the [related issue in Ruff](https://github.com/charliermarsh/ruff/issues/282):

> I'd love to sponsor someone (financially, that is) to work on this issue (in RustPython). If you're interested, shoot me a DM on [Twitter](https://twitter.com/charliermarsh) or email me at charlie.r.marsh@gmail.com :)

youknowone:
To be honest, after watching @andersk's work, I thought his way would be much better than my hacky local work.
I was going to add a mode to lexer to optionally generate a match token - which is not a good idea for either design or performance.
@andersk Do you have a plan to work on it? I think you are the one who is most close to the goal.

charliermarsh:
Another possibility is that if @andersk has any guidance to share on how he thought this might be implemented (per prior comments), I could give it a try. I've put off going deep on LALRPOP :)

andersk:
I’m still slowly working on this. I have a rough plan for splitting the grammar that ought to work with LR(1) theoretically, but I took a detour to investigate what seem to be LALRPOP bugs. One fix I wrote is lalrpop/lalrpop#697 (I wonder if LALRPOP still has maintainers who feel empowered to merge it…), but there are other issues I haven’t finished diagnosing.

I’m not entirely sure at this point whether I’m going down this path because I think it will result in the best parser or just because it seems like an interesting challenge. Switching to a more powerful parser algorithm may be the best long-term plan, especially if Python continues adding more tricky constructs to the grammar.

flying-sheep:
> Switching to a more powerful parser algorithm may be the best long-term plan, especially if Python continues adding more tricky constructs to the grammar.

“May be” might be an understatement. A parser library that can fundamentally handle the complexity that Python’s grammar has is definitely better than using one that doesn’t and hacking around its limitations, right? Guido wouldn’t have replaced Python’s own parser unless that was the case. ([The PEP](https://peps.python.org/pep-0617/) mentions these hacks in its introduction paragraph)

charliermarsh:
Totally agree that we should look to upgrade the parser if we believe LALRPOP will be insufficient in the long-term. From a pragmatic perspective, though, I still think there's a lot of value in supporting structural pattern matching via the current parser if possible, even if it requires some hacks.

flying-sheep:
I wasn’t suggesting that LALRPOP wasn’t enough! Python’s old parser was LL(1), i.e. fixed-length lookahead only. I think that’s where the hacks came in. LALRPOP more powerful than that and I’m think that that’s probably sufficiently powerful to parse the whole grammar without hacks.

PR #4324, opened 2022-12-10
Fix the end location of an implicitly-concatenated string

harupy:
## Before the fix:

```
"hello" "world"
      ^
      end_location
```

## After the fix:

```
"hello" "world"
              ^
              end_location
```


DimitrisJim:
another test passing: `UNEXPECTED SUCCESS: test_continued_str (test.test_ast.EndPositionTests)`, just like the last time remove the deco from it and the failures should go.

harupy:
@DimitrisJim Thanks for the comment! Removed the decorator.

DimitrisJim (approved):
thanks for the fix @harupy

Issue #4315, opened 2022-12-08
Removing list with index 0 causes panicked.


xiaxinmeng:
In the following example, we define a list1 with a class instance, then we remove the list1, rustpython is panicked. 

test.py:

```

class rewrite_list_eq(list):
	pass

class poc:
    def __eq__(self, other):
        list1.clear()
        return self
list1 = rewrite_list_eq([poc()])
list1.remove(list1)

```

Behavior on rustpython(rustpython 0.1.2):
```
thread 'main' panicked at 'removal index (is 0) should be < len (is 0)', vm/src/builtins/list.rs:303:38
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Expected behavior(behavior on CPython 3.9.1):
work well on the Python interpreter. 

Step to reproduce:
type "cargo run --release test.py" on the console of Ubuntu 18.04 and MacOS Big Sur 11.7.1
rust version: rustc 1.65.0

DimitrisJim:
Note to anyone wanting to tackle this, test similar to this exist in [`test_equal_operator_modifying_operand`](https://github.com/RustPython/RustPython/blob/77e5d0c8eda0c94700c48bd239021094c7fca964/Lib/test/test_list.py#L168) and [`test_count_index_remove_crashes`](https://github.com/RustPython/RustPython/blob/77e5d0c8eda0c94700c48bd239021094c7fca964/Lib/test/test_list.py#L202) of `test_list`.

yt2b:
I investigated the source code.
https://github.com/RustPython/RustPython/blob/0723271a3687856d2a961a5218eefd7b1ada2a3e/vm/src/builtins/list.rs#L298-L303
After calling `self.mut_index`, an elements of list has been deleted.
then `self.borrow_vec_mut().remove` is called.
But list is empty, So panic occured.

Checking list size is needed After calling `self.mut_index`.
I'll fix `PyList::remove` and add test in `test_list.py`.

PR #4326, opened 2022-12-10
Fix `PyList.remove`

yt2b:
#4315
I added comparison between index and list size in `PyList.remove`.

### DIFF
--- a/vm/src/builtins/list.rs
+++ b/vm/src/builtins/list.rs
@@ -300,8 +300,8 @@ impl PyList {
 
         if let Some(index) = index.into() {
             // defer delete out of borrow
-            let mut elements = self.borrow_vec_mut();
-            Ok((index < elements.len()).then(move || elements.remove(index)))
+            let is_inside_range = index < self.borrow_vec().len();
+            Ok(is_inside_range.then(|| self.borrow_vec_mut().remove(index)))
         } else {
             Err(vm.new_value_error(format!("'{}' is not in list", needle.str(vm)?)))
         }
--- a/vm/src/builtins/list.rs
+++ b/vm/src/builtins/list.rs
@@ -300,7 +300,8 @@ impl PyList {
 
         if let Some(index) = index.into() {
             // defer delete out of borrow
-            Ok(self.borrow_vec_mut().remove(index))
+            let mut elements = self.borrow_vec_mut();
+            Ok((index < elements.len()).then(move || elements.remove(index)))
         } else {
             Err(vm.new_value_error(format!("'{}' is not in list", needle.str(vm)?)))
         }
--- a/Lib/test/test_list.py
+++ b/Lib/test/test_list.py
@@ -198,7 +198,6 @@ class ListTest(list_tests.CommonTest):
         self.assertEqual(iter_size, sys.getsizeof(list([0] * 10)))
         self.assertEqual(iter_size, sys.getsizeof(list(range(10))))
 
-    @unittest.skip("TODO: RUSTPYTHON; hang")
     def test_count_index_remove_crashes(self):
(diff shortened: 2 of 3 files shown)


DimitrisJim:
also, this does actually fix one of the tests I mentioned in the issue, if you remove the skip decorator from `test_count_index_remove_crashes` it should pass just fine.

yt2b:
@DimitrisJim 
Code fix completed.

vm/src/builtins/list.rs
```diff
@@ -300,7 +300,8 @@ impl PyList {
 
         if let Some(index) = index.into() {
             // defer delete out of borrow
-            Ok(self.borrow_vec_mut().remove(index))
+            let mut elements = self.borrow_vec_mut();
+            Ok((index < elements.len()).then(move || elements.remove(index)))
         } else {
             Err(vm.new_value_error(format!("'{}' is not in list", needle.str(vm)?)))
         }
```
youknowone, line 307: This is incorrect fix. If size differs, it will silently skip remove operation - which is obvious error.

I suggest to lock both index and remove operation using `self.borrow_vec_mut()` if possible.

yt2b: I implemented it this way because `list.remove` returns None in CPython.

```
>>> class rewrite_list_eq(list):
...     pass
... 
>>> 
>>> class poc:
...     def __eq__(self, other):
...         list1.clear()
...         return self
... 
>>> 
>>> list1 = rewrite_list_eq([poc()])
>>> print(list1.remove(list1))
None
```

But as you said, this is hard-to-recognize error.
Should I fix this so that `list.remove` returns `ValueError`?

youknowone: The bug is broken sync between `index` and `remove`. out of index is one of the visible example of it. e.g. when l = [0, 1, 2, 3] and tried `l.remove(3)`, `del l[0]` can be done between `index` and `remove`. Then it is the case you described. When user ran `l.remove(3)`, they expect `3` is gone. This is the reason I said hard-to-recognize.
But what if `l.insert(0, 0)` happened instead of `del l[0]`? Then it will remove `2` instead of `3`. This is the original bug. Your fix is about only the case `l.remove(3)`, but not applicable to other cases.
Instead of it, we'd better to synchronize `index` and `remove`.

DimitrisJim: > Then it will remove 2 instead of 3.

Unless I'm misunderstanding your comment, this is also how CPython handles it, a user doing weird things in `__eq__` can't really expect sensible behavior:

```
>>> class BadEq:
...     def __eq__(self, other):
...             if isinstance(self, type(other)):
...                     l.insert(0, 0)
...                     return True
...             return NotImplemented
... 
>>> l = [10, 20, BadEq()]
>>> l.remove(BadEq())
>>> l
[0, 10, <__main__.BadEq object at 0x7f450686e070>]
```

As far as I can tell from [`list_ass_slice`](https://github.com/python/cpython/blob/2e279e85fece187b6058718ac7e82d1692461e26/Objects/listobject.c#L660) they check/force that the value for the index (`ilow` and `ihigh`) is in bounds for the list and then perform the operation.

youknowone: Ah, thank you. And sorry for wrong comment. I needed to check how cpython handle it first.

So is this version compatible to CPython @DimitrisJim ?

DimitrisJim: It definitely gets us closer, we definitely shouldn't panic when the list changes size underneath our feet.

yt2b:
@DimitrisJim @youknowone 
I changed to lock in both index and remove operation.
Could you please review again?

DimitrisJim (approved):
sorry, forgot about this, looks good to me!

PR #4327, opened 2022-12-11
Fix end location of compound statements

harupy:
For example, `FunctionDef` currently ends at where `DEDENT` occurs:

```python
def f1():
    pass



def f2():
^ here
    pass
```

but it should end at the final statement in the function body:

```python
def f1():
    pass
       ^ should end here



def f2():
    pass
```

### DIFF
--- a/compiler/parser/python.lalrpop
+++ b/compiler/parser/python.lalrpop
@@ -364,11 +364,11 @@ IfStatement: ast::Stmt = {
         // Determine last else:
         let mut last = s3.map(|s| s.2).unwrap_or_default();
         let end_location = last
-	            .last()
-	            .or_else(|| s2.last().and_then(|last| last.4.last()))
-	            .or_else(|| body.last())
-	            .unwrap()
-	            .end_location;
+            .last()
+            .or_else(|| s2.last().and_then(|last| last.4.last()))
+            .or_else(|| body.last())
+            .unwrap()
+            .end_location;
         // handle elif:
         for i in s2.into_iter().rev() {
             let x = ast::Stmt {
--- a/compiler/parser/python.lalrpop
+++ b/compiler/parser/python.lalrpop
@@ -360,15 +360,11 @@ IfStatement: ast::Stmt = {
     <location:@L> "if" <test:NamedExpressionTest> ":" <body:Suite> <s2:(@L "elif" NamedExpressionTest ":" Suite)*> <s3:("else" ":" Suite)?> => {
         // Determine last else:
         let mut last = s3.map(|s| s.2).unwrap_or_default();
-        let end_location = if let Some(last) = last.last() {
-            last.end_location
-        } else {
-            if let Some(last) = s2.last() {
-                last.4.last().unwrap().end_location
-            } else {
-                body.last().unwrap().end_location
-            }
-        };
+        let end_location = last
+            .last()
+            .map(|last| last.end_location)
+            .or_else(|| s2.last().map(|last| last.4.last().unwrap().end_location))
+            .unwrap_or_else(|| body.last().unwrap().end_location);
         // handle elif:
         for i in s2.into_iter().rev() {
             let x = ast::Stmt {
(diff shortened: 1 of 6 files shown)


compiler/parser/python.lalrpop
```diff
@@ -646,15 +647,16 @@ KwargParameter<ArgType>: Option<Box<ast::Arg>> = {
 };
 
 ClassDef: ast::Stmt = {
-    <decorator_list:Decorator*> <location:@L> "class" <name:Identifier> <a:("(" ArgumentList ")")?> ":" <body:Suite> <end_location:@R> => {
+    <decorator_list:Decorator*> <location:@L> "class" <name:Identifier> <a:("(" ArgumentList ")")?> ":" <body:Suite> => {
```
harupy, line 650: Just modified `ClassDef` and `FuncDef` as a test.

harupy: @DimitrisJim Thanks for the comment. Yes, I need to apply the same change to other compound statements. I'm working on that now.

DimitrisJim: ah yes, just saw it's still a draft! :smile:

harupy: @DimitrisJim Updated other compound statements.

compiler/parser/python.lalrpop
```diff
@@ -416,10 +434,19 @@ TryStatement: ast::Stmt = {
...
         let finalbody = finally.map(|s| s.2).unwrap_or_default();
+        let end_location = if let Some(last) = finalbody.last() {
+            last.end_location
+        } else {
+            if let Some(last) = orelse.last() {
+                last.end_location
+            } else {
+                handlers.last().unwrap().end_location
+            }
+        };
```
harupy, line 445: Asked ChatGPT if we can simplify this code:

image

compiler/parser/python.lalrpop
```diff
@@ -375,19 +379,23 @@ IfStatement: ast::Stmt = {
...
     },
 };
 
 WhileStatement: ast::Stmt = {
-    <location:@L> "while" <test:NamedExpressionTest> ":" <body:Suite> <s2:("else" ":" Suite)?> <end_location:@R> => {
+    <location:@L> "while" <test:NamedExpressionTest> ":" <body:Suite> <s2:("else" ":" Suite)?> => {
+        let end_location = s2
+            .as_ref()
+            .map(|s| s.2.last().unwrap().end_location)
+            .unwrap_or_else(|| body.last().unwrap().end_location);
```
DimitrisJim, line 393: there's the `orelse` declared in the next line (line `394`), move that to the front and use the same logic as the other places, i.e:

```suggestion
        let orelse = s2.map(|s| s.2).unwrap_or_default();
        let end_location = orelse
            .last()
            .or_else(body.last())
            .unwrap()
            .end_location;
```

compiler/parser/python.lalrpop
```diff
@@ -398,7 +406,12 @@ WhileStatement: ast::Stmt = {
 };
 
 ForStatement: ast::Stmt = {
-    <location:@L> <is_async:"async"?> "for" <target:ExpressionList> "in" <iter:TestList> ":" <body:Suite> <s2:("else" ":" Suite)?> <end_location:@R> => {
+    <location:@L> <is_async:"async"?> "for" <target:ExpressionList> "in" <iter:TestList> ":" <body:Suite> <s2:("else" ":" Suite)?> => {
+        let end_location = s2
```
DimitrisJim, line 410: similar case here as in the `While` case.

DimitrisJim (commented):
thanks! I've left a couple of comments on this

harupy:
@DimitrisJim Thanks for the comments! Applied your suggestions.

DimitrisJim (approved):
looks good to me, thanks!

harupy:
cc @charliermarsh

PR #4328, opened 2022-12-12
Remove unnecessary boxing of ASDL product children

andersk:
It seems unnecessary to box the children of an ASDL product type for the same reason that it’s unnecessary to box the elements of a `Vec`.

(Refs #2399, where the ASDL converter was introduced.)


youknowone:
Maybe this is boxed to reduce the type size. Do we need benchmark about it?

youknowone:
@andersk I am sorry, but could you rebase it? I will merge if @charliermarsh doesn't decline it.

coolreader18 (approved):
LGTM!

PR #4329, opened 2022-12-12
Parse Python 3.9+ parenthesized context managers

### LINKED ISSUE #4145: Unable to parse parenthesized context managers
In RustPython, this fails to parse:

```
from contextlib import contextmanager

@contextmanager
def ctx():
    yield

with (ctx() as foo):
    ...
```

...with:

```
---- parser::tests::test_parse_parenthesized stdout ----
thread 'parser::tests::test_parse_parenthesized' panicked at 'called `Result::unwrap()` on an `Err` value: BaseError { error: UnrecognizedToken(As, None), location: Location { row: 7, column: 13 }, source_path: "<test>" }', compiler/parser/src/parser.rs:209:58
```

andersk:
Since the upstream grammar for this is not LR(1), we abuse LALRPOP macros and the `Into`/`TryInto` traits to build a cover grammar that converts to either tuples or `with` items after additional validation. It’s annoying and ugly, but something like this is basically our only option short of switching to a more powerful parser algorithm.

Fixes #4145.

(Cc @charliermarsh, who I know is working on a more powerful parser algorithm in charliermarsh/ruff#286. But this may be a useful stopgap.)

### DIFF
--- a/Lib/test/test_ensurepip.py
+++ b/Lib/test/test_ensurepip.py
@@ -16,8 +16,6 @@ class TestPackages(unittest.TestCase):
         fullname = os.path.join(directory, filename)
         open(fullname, "wb").close()
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_version(self):
         # Test version()
         with tempfile.TemporaryDirectory() as tmpdir:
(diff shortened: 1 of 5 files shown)


youknowone (approved):
Awesome, thank you. Do you have a good idea about `match` too?

charliermarsh:
Amazing work!

PR #4330, opened 2022-12-12
Set `ExprContext::Store` on parenthesized with expressions

charliermarsh:
Small follow-up to #4329.

I noticed that in the `with (0 as a, 1 as b): pass` case, `a` and `b` had `ExprContext::Load`, which was causing Ruff to treat them as unbound variables in the body.


charliermarsh:
\cc @youknowone

andersk (approved):
Good catch, looks right to me.

DimitrisJim (approved):

PR #4333, opened 2022-12-15
Fix bug in Sliceable and rename del_item* to delitem

qingshi163:


fanninpm:
Rust 1.66 was released yesterday, which might explain a few of the CI failures here.

fanninpm:
I merged #4334 an hour ago, so feel free to rebase this branch on a fresh copy of `main`.

youknowone (approved):
which bug does it fix? is there a related test case?

qingshi163:
> which bug does it fix? is there a related test case?

It was logic bug that always fallback if step == -1

PR #4341, opened 2022-12-18
Support variable exchange (aka basic tuple support)

evilpie:
I mostly just wrote this limited implementation to test out the performance of a simple iterative fibonacci implementation that I found. https://pythonistaplanet.com/fibonacci-sequence-iterative/

I modified the code to avoid calling print. For testing I also started at a higher number so the call overhead from python to the jitted code is reduced.

| no jit | jit |
| -------|-----|
| f(79) = 14472334024676221 | f(79) = 14472334024676221 |
| time: 0.0044 s |  time: 0.00048 s |

<details>
<summary>code</summary>

```python
import time

def f(n: int):
  a = 0
  b = 1
  c = 0
  while (n - 1):
    c = a + b
    a,b = b,c
    n = n - 1
  return c


start = time.time()
for i in range(50, 80):
  print(f"f({i}) = {f(i)}")
print("time: {:.2} s".format(time.time() - start))

print("__jit__()", f.__jit__())

start = time.time()
for i in range(50, 80):
  print(f"f({i}) = {f(i)}")
print("time: {:.2} s".format(time.time() - start))
```
</details>

### DIFF
--- a/jit/src/instructions.rs
+++ b/jit/src/instructions.rs
@@ -26,6 +26,7 @@ enum JitValue {
     Float(Value),
     Bool(Value),
     None,
+    Tuple(Vec<JitValue>),
 }
 
 impl JitValue {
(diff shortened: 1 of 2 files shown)


youknowone (approved):
looks great, thank you!

PR #4342, opened 2022-12-18
Store Current VirtualMachine Pointer in LocalKey

qingshi163:
If I understand it correctly, the current vm is not required to be the same vm that the pyobject is created except the __del__ slot.
The current vm should match the vm provided by function argument that for now always getting from enter_vm function.

Is this the right appoarch?
@youknowone @coolreader18


coolreader18:
Ah, that's a good point, VM_STACK is overkill for the simple case. And actually, you can just use the `scoped-tls` crate for this, this is the exact sort of thing it's meant for.

youknowone (approved):
yes, this is right. can we even drop the whole VM_STACK?

youknowone (approved):

PR #4343, opened 2022-12-19
Fix the location of `FormattedValue`

harupy:
This is an attempt to fix the location of `FormattedValue`.

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -274,8 +274,9 @@ where
 
             // Check if we have a string:
             if matches!(self.window[0], Some('"' | '\'')) {
+                let prefix = if name.is_empty() { None } else { Some(name) };
                 return self
-                    .lex_string(saw_b, saw_r, saw_u, saw_f)
+                    .lex_string(saw_b, saw_r, saw_u, saw_f, prefix)
                     .map(|(_, tok, end_pos)| (start_pos, tok, end_pos));
             }
         }
@@ -559,6 +560,7 @@ where
         is_raw: bool,
         is_unicode: bool,
         is_fstring: bool,
+        prefix: Option<String>,
     ) -> LexResult {
         let start_pos = self.get_pos();
         let quote_char = self.next_char().unwrap();
@@ -685,21 +687,6 @@ where
             } else {
                 StringKind::Normal
             };
-
-            // Reference for string literal prefixes:
-            // https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
-            let prefix = if is_raw && is_fstring {
-                Some("rf".to_string())
-            } else if is_raw {
-                Some('r'.to_string())
-            } else if is_fstring {
-                Some('f'.to_string())
-            } else if is_unicode {
-                Some('u'.to_string())
-            } else {
-                None
-            };
-
             Tok::String {
                 value: string_content,
                 kind,
@@ -924,7 +911,7 @@ where
                 self.emit(comment);
             }
             '"' | '\'' => {
-                let string = self.lex_string(false, false, false, false)?;
+                let string = self.lex_string(false, false, false, false, None)?;
                 self.emit(string);
             }
             '=' => {
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -154,7 +154,7 @@ pub type LexResult = Result<Spanned, LexicalError>;
 
 #[inline]
 pub fn make_tokenizer(source: &str) -> impl Iterator<Item = LexResult> + '_ {
-    make_tokenizer_located(source, Location::new(0, 0))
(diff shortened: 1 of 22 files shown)


compiler/parser/src/lexer.rs
```diff
@@ -1366,6 +1379,8 @@ mod tests {
         Tok::String {
             value: s.to_owned(),
             kind: StringKind::Normal,
+            prefix: None,
+            triple_quoted: false,
         }
     }
```
fanninpm, line 1385: It might be a good idea to change the definition of this function to accept non-default values for `prefix` and `triple_quoted`.

harupy: @fanninpm Thanks for the comment. Yeah, I'll add `prefix` and `triple_quoted`.

```rust
    fn stok(s: &str, prefix: Option<String>, triple_quoted: bool) -> Tok {
        Tok::String {
            value: s.to_owned(),
            kind: StringKind::Normal,
            prefix,
            triple_quoted,
        }
    }
```

harupy: Fixed.

compiler/parser/src/lexer.rs
```diff
@@ -683,9 +681,24 @@ where
...
+                Some("rf".to_string())
+            } else if is_raw {
+                Some('r'.to_string())
+            } else if is_fstring {
+                Some('f'.to_string())
+            } else if is_unicode {
+                Some('u'.to_string())
+            } else {
+                None
+            };
```
harupy, line 695: Ref: https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals

compiler/parser/src/token.rs
```diff
@@ -6,12 +6,28 @@ use std::fmt::{self, Write};
...
+    },
+    Complex {
+        real: f64,
+        imag: f64,
+    },
+    String {
+        value: String,
+        kind: StringKind,
+        triple_quoted: bool,
+        prefix: Option<String>,
```
harupy, line 26: Added `triple_quoted` and `prefix` to get the number of characters preceding the f-string content.

```
  fr"""abc"""
  -----
    What is the number of characters preceding the f-string content?
```

compiler/parser/src/lexer.rs
youknowone: Would you put the reference to the source code comments?

youknowone (approved):

compiler/parser/src/lexer.rs
harupy: @youknowone Added!

compiler/parser/src/lexer.rs
```diff
@@ -683,9 +681,26 @@ where
             } else {
                 StringKind::Normal
             };
+
+            // Reference for string literal prefixes:
+            // https://docs.python.org/3/reference/lexical_analysis.html#string-and-bytes-literals
```
harupy, line 686: @youknowone Feel free to fix this if this is not what you meant by https://github.com/RustPython/RustPython/pull/4343#discussion_r1056854560

compiler/parser/src/lexer.rs
```diff
@@ -272,8 +270,9 @@ where
 
             // Check if we have a string:
             if matches!(self.window[0], Some('"' | '\'')) {
+                let prefix = if name.is_empty() { None } else { Some(name) };
```
harupy, line 273: If we encounter `"` or `'`, `name` should represent the prefix.

harupy:
Found the location of `x` is incorrect in the following code:

```
f"\\n {x}"
      ^   
```

compiler/parser/src/lexer.rs
```diff
@@ -1366,18 +1368,27 @@ mod tests {
         assert_eq!(vec!['b', '\\', '\n'], x);
     }
 
-    fn stok(s: &str) -> Tok {
+    fn stok(s: &str, prefix: Option<String>, triple_quoted: bool) -> Tok {
         Tok::String {
             value: s.to_owned(),
             kind: StringKind::Normal,
+            prefix,
```
fanninpm, line 1375: If you prefer not to repeat yourself (using, e.g, `'r'.to_string()` in function calls), consider trying something like this:

```suggestion
    fn stok<T: ToString>(s: &str, prefix: Option<T>, triple_quoted: bool) -> Tok {
        Tok::String {
            value: s.to_owned(),
            kind: StringKind::Normal,
            prefix: prefix.map(|x| x.to_string()),
```

harupy: @fanninpm Thanks!

harupy:
The current implementation doesn't work. Let's say we have code like this:

```python
f"\\{x}"
```

The lexer converts this to:

```python
f"\{x}"
```

If we parse this, the location of `x` is shited left by one character.

Issue #4336, opened 2022-12-17
`base64.b64decode(...)` function compatibility


moreal:
## Feature

There is `b64decode()` function in the `base64` module. And it will return `b'\xdd'` when you run the below code in the CPython implementation.

```python
import base64
base64.b64decode(b'3d==')
```

```python
>>> import base64
>>> base64.b64decode(b'3d==')
b'\xdd'
```

But I saw the error when running the code in RustPython implementation

```python
>>>>> import base64
>>>>> base64.b64decode(b'3d==')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/path/to/RustPython/pylib/Lib/base64.py", line 87, in b64decode
    return binascii.a2b_base64(s)
Error: error decoding base64: Invalid last symbol 100, offset 1.
```

## Research

When I saw the valid characters in base64, `'d'` (`100`) seems valid.

When I tried to run `decode(b"3d==")`, in the [marshallpierce/rust-base64](https://github.com/marshallpierce/rust-base64/) project, it was failed too. I guess the implementation is wrong or my knowledge is not perfect.

--------

- HEAD commit: ee4a51c512953eb16fb52fc061934ce2f5458595

PR #4344, opened 2022-12-19
Allow trailing bits while decoding base64

moreal:
## Description

This pull request resolves https://github.com/RustPython/RustPython/issues/4336.

Those lines to throw the `base64::DecodeError::InvalidLastSymbol{ .. }` are the below lines.

```rust
// https://github.com/marshallpierce/rust-base64/blob/10b73f67c4b19dc7fc1f4c4ae250558cba90c8d8/src/engine/fast_portable/decode_suffix.rs#L141-L147
    if !decode_allow_trailing_bits && (leftover_bits & mask) != 0 {
        // last morsel is at `morsels_in_leftover` - 1
        return Err(DecodeError::InvalidLastSymbol(
            start_of_leftovers + morsels_in_leftover - 1,
            last_symbol,
        ));
    }
```

And it provides an option, `decode_allow_trailing_bits`, to pass the check. This pull request uses the config option.

## Test

You can copy the below line and run the RustPython build on the `main` branch and on this pull request.

```python
import base64; base64.b64decode(b'3d==')
```

### DIFF
--- a/Lib/test/test_binascii.py
+++ b/Lib/test/test_binascii.py
@@ -32,8 +32,6 @@ class BinASCIITest(unittest.TestCase):
         self.assertTrue(issubclass(binascii.Error, Exception))
         self.assertTrue(issubclass(binascii.Incomplete, Exception))
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_functions(self):
         # Check presence of all functions
         for name in all_functions:
@@ -386,8 +384,6 @@ class BinASCIITest(unittest.TestCase):
         self.assertEqual(b2a_qp(type2test(b'a.\n')), b'a.\n')
         self.assertEqual(b2a_qp(type2test(b'.a')[:-1]), b'=2E')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_empty_string(self):
         # A test for SF bug #1022953.  Make sure SystemError is not raised.
         empty = self.type2test(b'')
--- a/Lib/test/test_binascii.py
+++ b/Lib/test/test_binascii.py
@@ -4,12 +4,14 @@ import unittest
 import binascii
 import array
 import re
+from test.support import bigmemtest, _1G, _4G, warnings_helper
+
 
 # Note: "*_hex" functions are aliases for "(un)hexlify"
-b2a_functions = ['b2a_base64', 'b2a_hex', 'b2a_hqx', 'b2a_qp', 'b2a_uu',
-                 'hexlify', 'rlecode_hqx']
-a2b_functions = ['a2b_base64', 'a2b_hex', 'a2b_hqx', 'a2b_qp', 'a2b_uu',
-                 'unhexlify', 'rledecode_hqx']
+b2a_functions = ['b2a_base64', 'b2a_hex', 'b2a_qp', 'b2a_uu',
+                 'hexlify']
+a2b_functions = ['a2b_base64', 'a2b_hex', 'a2b_qp', 'a2b_uu',
+                 'unhexlify']
 all_functions = a2b_functions + b2a_functions + ['crc32', 'crc_hqx']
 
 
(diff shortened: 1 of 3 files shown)


DimitrisJim:
Surprised nothing tested for this, we should add a test for this in `extra_tests/snippets/stdlib_binascii`.

moreal:
> Surprised nothing tested for this, we should add a test for this in `extra_tests/snippets/stdlib_binascii`.

I added the example (i.e., `3d==`) in `extra_tests/snippets/stdlib_binascii` at [d5b9664](https://github.com/RustPython/RustPython/pull/4344/commits/d5b96642717575f3bfe6a135cc71f64e72843f1d) commit. Is it enough?

fanninpm:
> Surprised nothing tested for this

It might be a good idea to update `test_binascii.py`.

moreal:
> It might be a good idea to update `test_binascii.py`.

To check I understood well, do you mean:

 - Add testcase to `test_binascii.py` in python/cpython
 - Add testcase to `test_binascii.py` in RustPython/RustPython
 - Update `test_binascii.py` to latest (i.e., 3.11 `test_binascii.py`)

fanninpm:
I'm sorry if I wasn't clear. I meant "update `Lib/test/test_binascii.py` to CPython 3.11". That version has the test case that covers what you're adding here (specifically, `test_base64_strict_mode()`).

moreal:
I updated `Lib/test/test_binascii.py`. Is there someone to review this pull request?

Lib/test/test_binascii.py
```diff
@@ -117,6 +114,51 @@ def addnoise(line):
         # empty strings. TBD: shouldn't it raise an exception instead ?
         self.assertEqual(binascii.a2b_base64(self.type2test(fillers)), b'')
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
fanninpm, line 118: Please break these lines out into a separate commit.

Lib/test/test_binascii.py
```diff
@@ -368,8 +384,6 @@ def test_qp(self):
         self.assertEqual(b2a_qp(type2test(b'a.\n')), b'a.\n')
         self.assertEqual(b2a_qp(type2test(b'.a')[:-1]), b'=2E')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
```
fanninpm, line 372: Please break these lines out into a separate commit.

fanninpm (changes requested):
Just a few little things, and then this will be good to go. Generally, we have one commit to update the CPython source, and an _additional_ commit for changes specific to RustPython.

Lib/test/test_binascii.py
moreal: I separated these lines in https://github.com/RustPython/RustPython/pull/4344/commits/04537b1993035f78aaab94b12edc31e03d10882e commit. 🙏🏻

Lib/test/test_binascii.py
moreal: I separated these lines in https://github.com/RustPython/RustPython/pull/4344/commits/dc9e958bb3172416e6d3c7747f32778c02436e13 commit. 🙏🏻

fanninpm:
Out of curiosity, what part of `test_base64_strict_mode()` is causing it to fail?

moreal:
> Out of curiosity, what part of `test_base64_strict_mode()` is causing it to fail?

Since CPython 3.11, A new keyword parameter `strict_mode` was added to `binascii.a2b_base64`. [^1] And we should implement it in `stdlib/src/binascii.rs#decl::a2b_binascii`. But I opened this pull request to resolve https://github.com/RustPython/RustPython/issues/4336 so I marked it with `unittest.expectedFailure`.

[^1]: https://docs.python.org/3/library/binascii.html?highlight=binascii#binascii.a2b_base64

moreal:
I opened a new issue https://github.com/RustPython/RustPython/issues/4374 for the `strict_mode` feature task.

fanninpm (approved):

fanninpm:
@DimitrisJim LGTM

PR #4346, opened 2022-12-20
Switch from 64-bit instruction enum to out-of-line arg values

coolreader18:
This gives a slight speedup and I have to imagine much better memory usage, but I don't think we have a way to profile the latter.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2002,8 +2002,8 @@ dependencies = [
  "lz4_flex",
  "num-bigint",
  "num-complex",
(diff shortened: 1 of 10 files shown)


youknowone (approved):

fanninpm:
> This gives a slight speedup and I have to imagine much better memory usage, but I don't think we have a way to profile the latter.

This method may only give anecdotal results, but is there any way of tracking how much memory the RustPython microbenchmarks use?

PR #4347, opened 2022-12-20
Refactor Number Protocol &PyNumber -> PyNumber

qingshi163:


vm/src/protocol/number.rs
```diff
@@ -57,71 +54,59 @@ impl PyObject {
 
         if let Some(i) = self.downcast_ref_if_exact::<PyInt>(vm) {
             Ok(i.to_owned())
+        } else if let Some(i) = self.to_number().int(vm) {
+            i
+        } else if let Some(i) = self.try_index_opt(vm) {
+            i
```
fanninpm, line 60: I feel like there's a simplification here. Would something like this work?

```suggestion
        } else if let Some(i) = self.to_number().int(vm).or_else(|| self.try_index_opt(vm)) {
            i
```

youknowone (approved):
Great! I'd like to work on this issue for long time.

Issue #4295, opened 2022-11-25
[RFC] Make PyXXXXMethod static


discord9:
## Summary
Releated static is 
- PyNumberMethod
- PyMappingMethod
- PySequenceMethod 
for now, they are using AtomicCell for storing methods, however, `crossbeam-utils` 0.8.10 make AtomicCell must impl Drop(Because internally it's become a MaybeUninit), so PyXXXXMethod can't be made static

## Detailed Explanation
The current method #4293 
Using `Lazy` will cause overhead for every visit. But according to this: https://docs.rs/crossbeam-utils/0.8.10/src/crossbeam_utils/atomic/atomic_cell.rs.html#33-47. using MaybeUninit can prevent some unsound, so I have no idea what we can do here to remove this overhead

## Drawbacks, Rationale, and Alternatives
Using `Lazy` for static have overhead, which might be unnecessary.

## Unresolved Questions
I have no idea what we can do here to remove this overhead of using `once_cell::sync::Lazy`

PR #4349, opened 2022-12-22
[Nightly] Atomic Structure with Const Initialize

qingshi163:
This appoarch need the nightly feature const trait impl, don't know if there is a workaround.

Ref: https://github.com/rust-lang/rust/issues/67792


qingshi163:
#4295

PR #4352, opened 2022-12-22
Simplify parenthesized context manager parsing with LALRPOP conditions

andersk:
LALRPOP is a bit more powerful than I realized. Its macros have a condition feature that let us easily create a subset of the expression grammar that doesn’t overlap with the `with` items grammar. Careful use of this feature lets us remove the cover grammar I added in #4329.

Now that the `with.rs` module is empty other than tests, maybe it makes sense to move the tests elsewhere? But I’ve left them as is for now.

(I’m feeling more confident that I understand LALRPOP’s limits, and I may have the beginning of a plan for correctly parsing `match` as a conditional keyword. I’ll try to get that working next.)

Cc @charliermarsh @youknowone


charliermarsh:
Amazing. `match` support would be _incredible_!

youknowone (approved):
awesome! I am learning from your patches. yeah, supporting match statement would be really great

PR #4355, opened 2022-12-24
make `test_globals` pass

yt2b:
To make `test_globals` in `test_symtable.py` pass, I've done following fix about `PySymbol`.

### Fix `is_global` function
I referenced the following CPython code.
https://github.com/python/cpython/blob/main/Lib/symtable.py#L247-L251

### Add `is_decleared_global` function
I referenced the following CPython code.
https://github.com/python/cpython/blob/main/Lib/symtable.py#L257-L260


youknowone (approved):
Thank you!

PR #4356, opened 2022-12-24
Fix parsing of tuple with named expression as context manager

andersk:
Because the `with` item grammar disallows named expressions, CPython parses `with (a := 0, b := 1):` as a tuple rather than two `with` items.


youknowone (approved):
👍

Issue #4353, opened 2022-12-23
Handling of BOM (byte order mark) in source files


squiddy:
## Feature

[PEP 263](https://peps.python.org/pep-0263/) specifies that 

> To aid with platforms such as Windows, which add Unicode BOM marks to the beginning of Unicode files, the UTF-8 signature \xef\xbb\xbf will be interpreted as ‘utf-8’ encoding as well (even if no magic encoding comment is given).

However, RustPython currently fails on files that include a BOM.

```console
$ printf '\xEF\xBB\xBF' > f.py
$ cargo run f.py
SyntaxError: Got unexpected token ﻿ at line 1 column 1

^
```

## Python Documentation

Not familiar with CPython source code, but some pointers:

https://github.com/python/cpython/blob/3.11/Lib/tokenize.py#L299 (`detect_encoding`)
https://github.com/python/cpython/blob/3.11/Lib/test/test_importlib/source/test_source_encoding.py

fanninpm:
I remember encountering something like this before, in https://github.com/RustPython/RustPython/pull/3919#discussion_r923493051.

yt2b:
Checking whether BOM exists is needed.

But in Rust, UTF-8 BOM(`EF BB BF`) is recognized as `FE FF` when converted from bytes to String.

```rust
let utf8_bom = vec![0xef, 0xbb, 0xbf];
println!("{:?}", String::from_utf8(utf8_bom)); // Ok("\u{feff}")
```

DimitrisJim:
Thanks for reporting this @squiddy! #4359 fixed the erroneous `SyntaxError`, handling of the mismatch between it and an encoding declaration that differs is not handled (and silently ignored atm) so keep that in mind.

PR #4359, opened 2022-12-25
Add BOM check

yt2b:
#4353 
I've added BOM check in `Lexer`.

With this fix, the following tests now pass.
So I've removed the decorators.
* `test_bom`
* `test_bom_and_utf_8`

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -236,7 +236,7 @@ where
         lxr.window.slide();
         lxr.window.slide();
         lxr.window.slide();
-        // Check if BOM exists
+        // TODO: Handle possible mismatch between BOM and explicit encoding declaration.
         if let Some('\u{feff}') = lxr.window[0] {
             lxr.window.slide();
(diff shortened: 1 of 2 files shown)


compiler/parser/src/lexer.rs
```diff
@@ -236,6 +236,10 @@ where
         lxr.window.slide();
         lxr.window.slide();
         lxr.window.slide();
+        // Check if BOM exists
+        if let Some('\u{feff}') = lxr.window[0] {
```
DimitrisJim, line 240: add a `TODO:` here since we still need to handle any conflicts between the BOM and the declared source encoding via the encoding declaration. I.e:

```suggestion
        // TODO: Handle possible mismatch between BOM and explicit encoding declaration.
        if let Some('\u{feff}') = lxr.window[0] {
```

yt2b: Code fix completed.

youknowone (approved):
good for now, thank you

PR #4361, opened 2022-12-25
Fix f-string padding

yt2b:
If multibyte character exists in string, f-string padding doesn't work correctly.
For example:
```py
f'{"🐍":4}'
```

On CPython, The result of this code is below.

```py
'🐍   '
```

But on RustPython, The Result is different.

```py
'🐍'
```

I've fixed the character counting method.
With this fix, the following tests now pass.

* `test_non_ascii_identifiers`
* `test_unicode`

### DIFF
--- a/Lib/test/test_tokenize.py
+++ b/Lib/test/test_tokenize.py
@@ -647,8 +647,6 @@ def"', """\
     DEDENT     ''            (4, 0) (4, 0)
     """)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_non_ascii_identifiers(self):
         # Non-ascii identifiers
         self.check_tokenize("Örter = 'places'\ngrün = 'green'", """\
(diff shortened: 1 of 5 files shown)


vm/src/format.rs
```diff
@@ -535,8 +535,7 @@ impl FormatSpec {
     ) -> Result<String, &'static str> {
         let align = self.align.unwrap_or(default_align);
 
-        // Use the byte length as the string length since we're in ascii
-        let num_chars = magnitude_string.len();
+        let num_chars = magnitude_string.chars().count();
```
youknowone, line 538: ah, this is not the best. probably we'd better to pass StringKind here to not to count characters for ascii strings.

yt2b: Code fix completed.

youknowone:
I added a commit. I am thinking adding a struct including `(&str, PyStrKindData)` and use them to propagate the data. this is not the version but proof of concept how it will work.

youknowone (approved):

PR #4362, opened 2022-12-26
Fix unwrap issue causing panic

### LINKED ISSUE #4312: Checking unmatched "raiseregex" leads to rustpython panicked
In the following example, we instance a object of unittest, then the unmatched error is collected by the assertRaisesRegex, rustpython get panicked.

test.py:

```
import unittest

class uTest(unittest.TestCase):
	pass

uTest = uTest()

with uTest.assertRaisesRegex(Exception, 'aaa'):
     aab
```

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', vm/src/suggestion.rs:55:22
```

Expected behavior(behavior on CPython 3.9.1):
`NameError: name 'aab' is not defined`

Step to reproduce:
type "cargo run --release test.py" on the console of Ubuntu 18.04 and MacOS Big Sur 11.7.1
rust version: rustc 1.65.0

rileysu:
This fixes #4312.

### DIFF
--- a/Lib/test/test_exceptions.py
+++ b/Lib/test/test_exceptions.py
@@ -1962,38 +1962,36 @@ class NameErrorTests(unittest.TestCase):
 
         self.assertNotIn("something", err.getvalue())
 
-    # TODO: RUSTPYTHON
-    # def test_issue45826(self):
-    #     # regression test for bpo-45826
-    #     def f():
-    #         with self.assertRaisesRegex(NameError, 'aaa'):
-    #             aab
+    def test_issue45826(self):
+        # regression test for bpo-45826
+        def f():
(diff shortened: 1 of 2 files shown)


youknowone:
@Snowapril could you review this change?

Snowapril (approved):
LGTM! I cannot remember why did I use `unwrap` there haha..

DimitrisJim (changes requested):
Great! This also appears to fix two tests (`test_issue45826` and `test_issue45826_focused`) in `test_exception.py`, un-comment both of them (and remove the `TODO: RUSTPYTHON` comment preceeding them in order for them to now run.

DimitrisJim (approved):
great, thanks!

PR #4366, opened 2022-12-27
Fix eval doesn't work correctly

### LINKED ISSUE #4364: `eval` doesn't work correctly.
On CPython, `eval` works even if space exists at the beginning of the line.

```
>>> eval("   1")
1
```

But on RustPython, IndentationError occurs.

```
>>>>> eval("   1")
IndentationError: unexpected indent at line 1 column 4
   1
   ^
```

branai:
Solved by removing preceding whitespaces, tabs, and newlines for evals only. Fixes #4364

### DIFF
--- a/vm/src/stdlib/builtins.rs
+++ b/vm/src/stdlib/builtins.rs
@@ -256,6 +256,15 @@ mod builtins {
                     ));
                 }
 
+                // Check for newlines, tabs, or whitespaces at start
+                let wsp_arr = [9, 10, 32];
+                let mut i = 0;
+                while i < source.len() && wsp_arr.contains(&source[i]) {
+                    i = i + 1;
+                }
+                // Slice off whitespaces
+                let source = &source[i..source.len()];
+
                 let source = std::str::from_utf8(source).map_err(|err| {
                     let msg = format!(
                         "(unicode error) 'utf-8' codec can't decode byte 0x{:x?} in position {}: invalid start byte",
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/builtins.rs
```diff
@@ -256,6 +256,15 @@ mod builtins {
                     ));
                 }
 
+                // Check for newlines, tabs, or whitespaces at start
```
DimitrisJim, line 259: Thanks! Though there's not really a need to manually remove them, you can use `trim_start()` on source before we create the `PyStr` from it, i.e `vm.ctx.new_str(source.trim_start())`.

Note that CPython currently ignores tabs and spaces only, don't think this is really an issue since the lexer ignores those either way.

branai:
@DimitrisJim I changed it to what you suggested, thank you for your help!

DimitrisJim:
There's a couple of tests passing in `test_string_literals.py` that you need to tweak. `test_eval_bytes_normal`, `test_eval_bytes_raw`, `test_eval_str_normal`, `test_eval_str_raw`, `test_eval_str_u` all now succeed. You should open that file and remove the decorators on them, i.e remove the lines:

```python
# TODO: RUSTPYTHON
@unittest.expectedFailure
```

and then the tests should pass just fine.

DimitrisJim (approved):
Great, thanks!

PR #4367, opened 2022-12-27
Prohibit starred arguments after double-starred arguments

andersk:
CPython prohibits `f(**kwargs, *args)`; we should too.

### DIFF
--- a/compiler/parser/src/function.rs
+++ b/compiler/parser/src/function.rs
@@ -55,6 +55,7 @@ pub fn parse_args(func_args: Vec<FunctionArgument>) -> Result<ArgumentList, Lexi
 
     let mut keyword_names =
         FxHashSet::with_capacity_and_hasher(func_args.len(), Default::default());
+    let mut double_starred = false;
     for (name, value) in func_args {
         match name {
             Some((start, end, name)) => {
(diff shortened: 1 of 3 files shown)


compiler/parser/src/function.rs
```diff
@@ -76,8 +79,9 @@ pub fn parse_args(func_args: Vec<FunctionArgument>) -> Result<ArgumentList, Lexi
                 ));
             }
             None => {
-                // Allow starred args after keyword arguments.
-                if !keywords.is_empty() && !is_starred(&value) {
+                // Allow starred arguments after keyword arguments but
+                // not after double-starred arguments.
```
DimitrisJim, line 83: Good catch, this does require a new error type though, CPython distinguishes between these:

```python
>>> foo(**{}, *[])
  File "<stdin>", line 1
    foo(**{}, *[])
              ^
SyntaxError: iterable argument unpacking follows keyword argument unpacking
```

fanninpm: In CPython, this is tested in `test_kwargs_last3()` in `test_syntax.py`.

fanninpm:

```bash
```

DimitrisJim (approved):

PR #4372, opened 2022-12-27
Refactor `FStringParser.parse_formatted_value`

harupy:
A minor refactoring I found in #4343 that can simplify the code. `expression` in `FStringParser.parse_formatted_value` currently contains `{}`, but it doesn't have to.

### Current

```python
f"a {b} c"
    ^^^
    expression contains { and }, requires slicing when parsed
```

### Fixed

```python
f"a {b} c"
     ^
     expression doesn't contain { and }, doesn't require slicing when parsed
```

### DIFF
--- a/compiler/parser/src/fstring.rs
+++ b/compiler/parser/src/fstring.rs
@@ -26,7 +26,7 @@ impl FStringParser {
         mut chars: iter::Peekable<str::Chars<'a>>,
         nested: u8,
     ) -> Result<(Vec<Expr>, iter::Peekable<str::Chars<'a>>), FStringErrorType> {
-        let mut expression = String::from("{");
+        let mut expression = String::new();
         let mut spec = None;
         let mut delims = Vec::new();
         let mut conversion = ConversionFlag::None;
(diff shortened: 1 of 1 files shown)


compiler/parser/src/fstring.rs
```diff
@@ -67,14 +67,14 @@ impl FStringParser {
                         Some('a') => ConversionFlag::Ascii,
                         Some('r') => ConversionFlag::Repr,
                         Some(_) => {
-                            return Err(if expression[1..].trim().is_empty() {
+                            return Err(if expression.trim().is_empty() {
```
DimitrisJim, line 70: without being sure, is all the `trim`ing required? When we match a space we push it into `trailing_seq` so would `expression` ever have starting/trailing whitespace?

harupy: I think it's required:

```python
>>> f"{    }"
  File "<stdin>", line 1
SyntaxError: f-string: empty expression not allowed
>>> f"{}"
  File "<stdin>", line 1
SyntaxError: f-string: empty expression not allowed
```

DimitrisJim: Ah right, for some reason I completely missed that match on the space character is guarded by `self.documenting`. Still have a hunch these can be refactored away but that's another issue.

DimitrisJim (approved):

Issue #4375, opened 2022-12-28
`eval` doesn't raise correct exception


yt2b:
```py
class X:
    def __getitem__(self, key):
        raise ValueError

eval("foo", {}, X())
```

On CPython, The output of this code is below.

```
Traceback (most recent call last):
  File "/Users/yusuke/Programs/Rust/Github/RustPython/imp.py", line 9, in <module>
    eval("foo", {}, X())
  File "<string>", line 1, in <module>
  File "/Users/yusuke/Programs/Rust/Github/RustPython/imp.py", line 6, in __getitem__
    raise ValueError
ValueError
```

But on RustPython, The output is different.

```
Traceback (most recent call last):
  File "imp.py", line 9, in <module>
    eval("foo", {}, X())
  File "<string>", line 1, in <module>
NameError: name 'foo' is not defined
```

I'll look into this.

yt2b:
This bug occures from `LoadNameAny` opecode.

https://github.com/RustPython/RustPython/blob/b1db1be031cfa69604d05dde187932ab94edb503/vm/src/frame.rs#L539-L547

`self.locals.mapping().subscript(name, vm)` returns value of the name.
If the name is not found in locals, it returns `KeyError`.
But because of calling `ok` function, returned exceptions is ignored.

If the name is not found in local, `self.load_global_or_builtin(name, vm)` is called.
This function returns value of the name.
If the name is not found in global, it returns `NameError`.

Returned exception from local is ignored, So `NameError` always occures if name is not found.

PR #4378, opened 2022-12-28
`eval` doesn't raise correct exception

yt2b:
#4375
I've fixed `LoadNameAny` opecode to return exception occured in local.
* test_eval

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -541,7 +541,7 @@ impl ExecutingFrame<'_> {
                 let result = self.locals.mapping().subscript(name, vm);
                 match result {
                     Ok(x) => self.push_value(x),
-                    Err(e) if e.class().is(vm.ctx.exceptions.key_error) => {
+                    Err(e) if e.fast_isinstance(vm.ctx.exceptions.key_error) => {
                         self.push_value(self.load_global_or_builtin(name, vm)?);
                     }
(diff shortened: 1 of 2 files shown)


vm/src/frame.rs
```diff
@@ -538,11 +538,14 @@ impl ExecutingFrame<'_> {
...
                 let name = self.code.names[*idx as usize];
-                let value = self.locals.mapping().subscript(name, vm).ok();
-                self.push_value(match value {
-                    Some(x) => x,
-                    None => self.load_global_or_builtin(name, vm)?,
-                });
+                let result = self.locals.mapping().subscript(name, vm);
+                match result {
+                    Ok(x) => self.push_value(x),
+                    Err(e) if e.class().is(vm.ctx.exceptions.key_error) => {
```
DimitrisJim, line 544: Tiny issue, we also need to also accept subclasses of `KeyError`:

```suggestion
                    Err(e) if e.fast_isinstance(vm.ctx.exceptions.key_error) => {
```

DimitrisJim (changes requested):

vm/src/frame.rs
yt2b: Code fix completed.

DimitrisJim (approved):
looks good, thanks!

PR #4379, opened 2022-12-28
Refactor `FStringParser`

harupy:
Refactors `FStringParser` to simplify https://github.com/RustPython/RustPython/pull/4343.

### DIFF
--- a/compiler/parser/src/fstring.rs
+++ b/compiler/parser/src/fstring.rs
@@ -7,7 +7,7 @@ use crate::{
 use std::{iter, mem, str};
 
 struct FStringParser<'a> {
-    source: iter::Peekable<str::Chars<'a>>,
+    chars: iter::Peekable<str::Chars<'a>>,
     str_start: Location,
     str_end: Location,
 }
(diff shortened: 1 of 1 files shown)


compiler/parser/src/fstring.rs
```diff
@@ -194,21 +202,21 @@ impl FStringParser {
...
+                    loop {
+                        match self.next_char() {
+                            Some(c) => {
+                                expression.push(c);
+                                if c == ch {
+                                    break;
+                                }
+                            }
+                            None => return Err(UnterminatedString),
                         }
```
burrbull, line 218:
```suggestion
                        let Some(c) = self.next_char() else {
                            return Err(UnterminatedString);
                        };
                        expression.push(c);
                        if c == ch {
                            break;
                        }
```

fanninpm: I keep forgetting that let-else is a thing now.

harupy: @burrbull Thanks for the suggestion!

burrbull (approved):

compiler/parser/src/fstring.rs
```diff
@@ -6,63 +6,72 @@ use crate::{
...
+        }
+    }
+
+    fn next_char(&mut self) -> Option<char> {
+        self.chars.next()
+    }
+
+    fn peek(&mut self) -> Option<&char> {
+        self.chars.peek()
     }
```
fanninpm, line 30: @DimitrisJim @youknowone Probably not necessary, but should these two functions be `#[inline]`?

DimitrisJim: I'd say so, afaik the compiler usually inlines small functions like these but that isn't guaranteed, hinting it with `inline` is helps.

youknowone (approved):

compiler/parser/src/fstring.rs
youknowone: I'd like to have inline for tiny functions.

youknowone (approved):

Issue #4381, opened 2022-12-29
Report argument name when a duplicate keyword argument is present


DimitrisJim:
Currently we don't mention the keyword name that was detected as a duplicate, i.e:

```python
>>> def foo(**kwargs): pass
>>> foo(k=10, k=20)
```

Results in:

```
SyntaxError: keyword argument repeated at line 1 column 8
```

While in CPython, the same call:

```
SyntaxError: keyword argument repeated: k
```

Ignoring our addition of the line/col numbers (which need some fixing), the same message should exist, i.e:

```
SyntaxError: keyword argument repeated: k at line 1 column 8
```

should be the wanted result.

fanninpm:
Have you tried creating a `.py` file with that example?

DimitrisJim:
yup, am I missing something?

fanninpm:
I was thinking that the line/column information may be redundant in some cases (e.g. running a file), but not others (e.g. running interactively).

anilbey:
Hello, I'm new to Rust. I come from a Python background. 
I thought contributing to RustPython can be a very good start for me :)

I just tried to address this issue in #4382. Does it look good to you?

The output with the #4382 change:
```
>>>>> def foo(**kwargs): pass
>>>>> foo(k=10, k=20)
SyntaxError: keyword argument 'k' repeated at line 1 column 10
foo(k=10, k=20)
```

PR #4382, opened 2022-12-29
arg name to be written upon duplicate kwargs error #4381

anilbey:
Output before:

```
>>>>> def foo(**kwargs): pass
>>>>> foo(k=10, k=20)
SyntaxError: keyword argument repeated at line 1 column 10
foo(k=10, k=20)
         ^
```

Output after:

```
>>>>> def foo(**kwargs): pass
>>>>> foo(k=10, k=20)
SyntaxError: keyword argument 'k' repeated at line 1 column 10
foo(k=10, k=20)
         ^
```

### DIFF
--- a/compiler/parser/src/error.rs
+++ b/compiler/parser/src/error.rs
@@ -55,7 +55,7 @@ impl fmt::Display for LexicalErrorType {
                 write!(f, "duplicate argument '{arg_name}' in function definition")
             }
             LexicalErrorType::DuplicateKeywordArgumentError(arg_name) => {
-                write!(f, "keyword argument '{arg_name}' repeated")
+                write!(f, "keyword argument repeated: {arg_name}")
             }
             LexicalErrorType::PositionalArgumentError => {
                 write!(f, "positional argument follows keyword argument")
--- a/compiler/parser/src/error.rs
+++ b/compiler/parser/src/error.rs
@@ -24,7 +24,7 @@ pub enum LexicalErrorType {
     DuplicateArgumentError(String),
     PositionalArgumentError,
     UnpackedArgumentError,
-    DuplicateKeywordArgumentError,
+    DuplicateKeywordArgumentError(String),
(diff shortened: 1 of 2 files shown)


compiler/parser/src/error.rs
```diff
@@ -54,8 +54,8 @@ impl fmt::Display for LexicalErrorType {
             LexicalErrorType::DuplicateArgumentError(arg_name) => {
                 write!(f, "duplicate argument '{arg_name}' in function definition")
             }
-            LexicalErrorType::DuplicateKeywordArgumentError => {
-                write!(f, "keyword argument repeated")
+            LexicalErrorType::DuplicateKeywordArgumentError(arg_name) => {
+                write!(f, "keyword argument '{arg_name}' repeated")
```
fanninpm, line 58: Here is the correct form, according to `test_syntax.py`:

```suggestion
                write!(f, "keyword argument repeated: {arg_name}")
```

fanninpm:

```shell
```

anilbey:
Thanks for the info. Now I can reproduce the error locally :).

anilbey:
On my ubuntu local 3 tests are failing. I looked at their logs and I think the issue is due to my openssl installation.

```
274 tests OK.
3 tests failed:
    test_ioctl test_socket test_urllib2

```

fanninpm:
> On my ubuntu local 3 tests are failing. I looked at their logs and I think the issue is due to my openssl installation.

Feel free to ignore _those_ failures, as we don't get a chance to test them that often. (Did you use the "test_whatever" variant from earlier?)

anilbey:
> Feel free to ignore those failures, as we don't get a chance to test them that often. (Did you use the "test_whatever" variant from earlier?)

Yes I used it on `test_syntax.py`, before committing your suggestion.

DimitrisJim (approved):
lgtm, thanks for the PR @anilbey!

PR #4384, opened 2022-12-30
Fix the location of `FormattedValue`

harupy:
This PR fixes the location of `FormattedValue` (initial attempt: https://github.com/RustPython/RustPython/pull/4343). In the following screenshot, the yellow waving underlines mark the location of `xyz`.

image

### Major changes:

- Preserve the original string during tokenization.


compiler/parser/src/fstring.rs
```diff
@@ -1,3 +1,4 @@
+// We no longer need this file
```
harupy, line 1: I'll remove this file.

youknowone (approved):

compiler/parser/src/fstring.rs
youknowone: Great!

harupy:
@youknowone Thanks for merging! I filed https://github.com/RustPython/RustPython/pull/4389 to address https://github.com/RustPython/RustPython/pull/4384#discussion_r1059305831 and https://github.com/RustPython/RustPython/pull/4379#discussion_r1059566557. Would you mind taking a look? Thanks in advance!

PR #4386, opened 2022-12-30
Fix character format

yt2b:
In CPython, integer can be converted to character.

```
>>> f"{65:c}"
'A'
```

But in RustPython, ValueError is raised.

```
>>>>> f"{65:c}"
ValueError: Unknown format code 'c' for object of type 'int'
```

I've fixed to convert an integer to character.
I referenced the following source code.
https://github.com/python/cpython/blob/79c10b7da84f52999dc483fc62c8e758ad3eff23/Python/formatter_unicode.c#L864

### DIFF
--- a/extra_tests/snippets/builtin_format.py
+++ b/extra_tests/snippets/builtin_format.py
@@ -33,5 +33,3 @@ assert f"{65:c}" == "A"
 assert f"{0x1f5a5:c}" == "🖥"
 assert_raises(ValueError, "{:+c}".format, 1, _msg="Sign not allowed with integer format specifier 'c'")
 assert_raises(ValueError, "{:#c}".format, 1, _msg="Alternate form (#) not allowed with integer format specifier 'c'")
-# TODO: raise OverflowError
-assert_raises(ValueError, "{:c}".format, -1, _msg="%c arg not in range(0x110000)")
--- a/extra_tests/snippets/builtin_format.py
+++ b/extra_tests/snippets/builtin_format.py
@@ -29,3 +29,9 @@ assert f'{65536:,}' == '65,536'
 assert f'{4294967296:,}' == '4,294,967,296'
 assert 'F' == "{0:{base}}".format(15, base="X")
 assert f'{255:#X}' == "0XFF"
+assert f"{65:c}" == "A"
+assert f"{0x1f5a5:c}" == "🖥"
+assert_raises(ValueError, "{:+c}".format, 1, _msg="Sign not allowed with integer format specifier 'c'")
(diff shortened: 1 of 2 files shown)


yt2b:
CI failed at the code I added in `builtin_format.py`.
On my local environment, `ValueError` is raised.
But on CI, `OverflowError` is raised.

vm/src/format.rs
```diff
@@ -477,7 +477,17 @@ impl FormatSpec {
...
             Some(FormatType::String) => Err("Unknown format code 's' for object of type 'int'"),
-            Some(FormatType::Character) => Err("Unknown format code 'c' for object of type 'int'"),
+            Some(FormatType::Character) => match (self.sign, self.alternate_form) {
+                (Some(_), _) => Err("Sign not allowed with integer format specifier 'c'"),
+                (_, true) => {
+                    Err("Alternate form (#) not allowed with integer format specifier 'c'")
+                }
+                (_, _) => match num.to_u32() {
+                    Some(n) if n <= 0x10ffff => Ok(std::char::from_u32(n).unwrap().to_string()),
+                    // TODO: raise OverflowError
```
DimitrisJim, line 487: Open an issue for this (and if you're working on it, tell me and I'll assign you). Leaving it in code as a TODO will probably mean it gets forgotten for a while.

yt2b: I've opened the issue. #4387
I'll tell you when I work on it.

extra_tests/snippets/builtin_format.py
```diff
@@ -29,3 +29,7 @@ def test_zero_padding():
 assert f'{4294967296:,}' == '4,294,967,296'
 assert 'F' == "{0:{base}}".format(15, base="X")
 assert f'{255:#X}' == "0XFF"
+assert f"{65:c}" == "A"
+assert f"{0x1f5a5:c}" == "🖥"
+assert_raises(ValueError, "{:+c}".format, 1, _msg="Sign not allowed with integer format specifier 'c'")
```
youknowone, line 34: This is added here because CPython doesn't have those tests, right?

youknowone (approved):

extra_tests/snippets/builtin_format.py
yt2b: Similar test is in `test_types.py`.
https://github.com/RustPython/RustPython/blob/cb4bd51489188cb37010db26902b42b8948e7213/Lib/test/test_types.py#L360
But `test_int__format__` is docorated with `@unittest.expectedFailure`.
So I've added it in `builtin_format.py`.

PR #4388, opened 2022-12-31
Fix be silent even with no file using stdin

### LINKED ISSUE #4363: Be silent even with no file using stdin
https://github.com/RustPython/RustPython/blob/3660ac20cff4152af3281040b5949eef9667493f/src/lib.rs#L187

```sh
echo -e "import json\njson.dumps(['foo', {'bar': ('baz', None, 1.0, 2)}])" | rustpython -q
```

### Actual

```
'["foo", {"bar": ["baz", null, 1.0, 2]}]'
```

### Expected

```
'["foo", {"bar": ["baz", null, 1.0, 2]}]'
```

branai:
Fixes #4363

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1897,6 +1897,7 @@ dependencies = [
 name = "rustpython"
 version = "0.1.2"
 dependencies = [
+ "atty",
  "cfg-if",
  "clap",
  "cpython",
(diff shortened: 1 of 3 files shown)


youknowone (approved):
👍

youknowone:
Thank you!

PR #4389, opened 2022-12-31
Follow-up for #4384

harupy:
A follow-up PR for #4384. Makes the following changes:

1. Move tests in `compiler/parser/src/fstring.rs` to `compiler/parser/src/fstring.rs/string_parser.rs` and remove it.
2. Add `#[inline]` to tiny methods in `StringParser`


compiler/parser/src/fstring.rs
```diff
@@ -1,479 +0,0 @@
-// We no longer need this file
```
harupy, line 1: `StringParser` is a superset of `FStringParser`. We can remove this file.

compiler/parser/src/string.rs
```diff
@@ -124,160 +124,3 @@ pub fn parse_strings(
             .expect("String must be concatenated to a single element.")
     })
 }
-
```
harupy, line 127: I'll revert this change

compiler/parser/src/string_parser.rs
```diff
@@ -36,6 +36,7 @@ impl<'a> StringParser<'a> {
         }
     }
 
+    #[inline]
```
harupy, line 39: Do we need this `inline`?

compiler/parser/src/string_parser.rs
```diff
@@ -36,6 +36,7 @@ impl<'a> StringParser<'a> {
...
     fn next_char(&mut self) -> Option<char> {
```
harupy, line 40: Do we need this `inline` on `next_char`?

compiler/parser/src/string_parser.rs
```diff
@@ -560,3 +563,136 @@ pub fn parse_string(
...
+    use super::*;
+
+    fn parse_fstring(source: &str) -> Result<Vec<Expr>, FStringErrorType> {
+        StringParser::new(
+            source,
+            StringKind::FString,
+            false,
+            Location::new(1, 0),
+            Location::new(1, source.len() + 3), // 3 for prefix and quotes
+        )
```
harupy, line 578: Assuming we're parsing a simple f-string that looks like this:

```
f"a {b} c"
```

youknowone (approved):

compiler/parser/src/string_parser.rs
DimitrisJim: its very likely this will get inlined either way (small-ish and private), annotating it doesn't hurt though.

harupy: @DimitrisJim Got it, thanks!

DimitrisJim:
thanks for all your recent PRs on this @harupy!

PR #4392, opened 2022-12-31
Prefix format with padding doesn't work correctly.

yt2b:
Prefixed format with padding doesn't work correctly.

```
>>>>> f"{256:#010x}"
'000000x100'
```

the output should be `0x00000100`
I've fixed format function.


youknowone (approved):
great!

PR #4394, opened 2023-01-01
Fix the markdown URL in the sample notebook

YDX-2147483647:
`[google](google.com)` → `[google](https://google.com)`

URLs without `https://` are relative.

If you run it in [the demo notebook](https://rustpython.github.io/demo/notebook/), then `google.com` becomes `https://rustpython.github.io/demo/notebook/google.com`.


DimitrisJim (approved):
thanks!

PR #4396, opened 2023-01-01
Remove unreachable code in `compiler/parser/src/string.rs`

harupy:
Removes unreachable code in `compiler/parser/src/string.rs`.

### DIFF
--- a/compiler/parser/src/string.rs
+++ b/compiler/parser/src/string.rs
@@ -111,18 +111,11 @@ pub fn parse_strings(
         deduped.push(take_current(&mut current));
     }
 
-    Ok(if has_fstring {
-        Expr::new(
-            initial_start,
-            last_end,
-            ExprKind::JoinedStr { values: deduped },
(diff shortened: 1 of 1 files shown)


compiler/parser/src/string.rs
```diff
@@ -111,18 +111,11 @@ pub fn parse_strings(
...
-        deduped
-            .into_iter()
-            .exactly_one()
-            .expect("String must be concatenated to a single element.")
-    })
+    Ok(Expr::new(
+        initial_start,
+        last_end,
+        ExprKind::JoinedStr { values: deduped },
+    ))
```
harupy, line 118: At line 55, there is an early return for the non-fstring:

https://github.com/RustPython/RustPython/blob/663edd22dea4f74c53547952d25e637f1e2a6f42/compiler/parser/src/string.rs#L55-L68

DimitrisJim (approved):

PR #4397, opened 2023-01-01
Fix separated format

yt2b:
Separated format with padding doesn't work correctly.

```
>>>>> f"{123:07,}"
'0000123'
```
The output should be `000,123`.
I've fixed format function.

* `test_format` in `test_ipaddress.py`


youknowone (approved):

youknowone:
Thank you!

PR #4399, opened 2023-01-02
Fix IndentationError works differently with cpython in interective shell

### LINKED ISSUE #3892: IndentationError works differently with cpython in interective shell.
cpython returns `IndentationError` when it gets ‘\n’ after statements like `for`, `if`
-  cpython
```
>>> for i in range(1,4):
... 
  File "<stdin>", line 2
    
    ^
IndentationError: expected an indented block
```
but rustpython keeps waiting until some string come in.
- rustpython
```
>>>>> for i in range(1,4):
..... 
..... 
..... d
..... 
IndentationError: expected an indented block at line 4 column 1
```

branai:
Fixes #3892  #4400 

Removed the automatic continuing logic that was in place for code that needs multiple lines in the shell to parse correctly. Now we check if full_code can be parsed after each line of input given, but ignoring specific EOF and IndentationLevel errors.

I am working on a way to do this without modifying errors.rs, if I should not be modifying that.

Edit: I think my edits to errors.rs may have improved EOF error handling.

### DIFF
--- a/compiler/parser/src/error.rs
+++ b/compiler/parser/src/error.rs
@@ -221,7 +221,7 @@ pub(crate) fn parse_error_from_lalrpop(
                     source_path,
                 }
             }
-        },
+        }
     }
 }
 
--- a/compiler/parser/src/error.rs
+++ b/compiler/parser/src/error.rs
@@ -205,10 +205,22 @@ pub(crate) fn parse_error_from_lalrpop(
                 source_path,
             }
         }
-        LalrpopError::UnrecognizedEOF { location, .. } => ParseError {
-            error: ParseErrorType::Eof,
-            location,
-            source_path,
+        LalrpopError::UnrecognizedEOF { location, expected } => {
+            // This could be an initial indentation error that we should ignore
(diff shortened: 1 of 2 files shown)


youknowone (approved):
Thank you!

DimitrisJim:
I think this PR might have been a bit more complex than needed. As far as I can tell, the main issue here could of been solved by just tweaking the `ShellExecResult::Continue` case to check if an empty line was the input (and only an empty line).

There's also a couple of little issues here, the Indentation Error isn't exactly the right one and now any line containing a comment or space/tabs results in an indent error.

The changes to error seem right though, but it can also be slightly tweaked to simplify (matching against it in `shell_exec` as the others do. For this we probably could use an IndentError with a wrong `token` value that we only use for the `match`.

branai:
@DimitrisJim I agree this fix was too complex, I simplified it in #4406 . The improved fix does exactly what you said with checking for an empty line. There are also cases in continue mode when a non-empty line is given where an error should be thrown, so I accounted for those too.

I'll look into what you mentioned about the indentationerror

PR #4402, opened 2023-01-02
Remove unreachable code in `compiler/parser/src/string_parser.rs`

harupy:
This PR removes unreachable code in `compiler/parser/src/string_parser.rs` to simplify the code.

### DIFF
--- a/compiler/parser/src/string_parser.rs
+++ b/compiler/parser/src/string_parser.rs
@@ -226,12 +226,12 @@ impl<'a> StringParser<'a> {
                         }
                     };
 
-                    if let Some(&peek) = self.peek() {
-                        if peek != '}' && peek != ':' {
-                            return Err(UnclosedLbrace.to_lexical_error(self.get_pos()));
+                    match self.peek() {
+                        Some('}' | ':') => {}
+                        Some(_) => {
+                            return Err(InvalidConversionFlag.to_lexical_error(self.get_pos()))
                         }
-                    } else {
-                        return Err(UnclosedLbrace.to_lexical_error(self.get_pos()));
+                        None => return Err(UnclosedLbrace.to_lexical_error(self.get_pos())),
                     }
                 }
 
--- a/compiler/parser/src/string_parser.rs
+++ b/compiler/parser/src/string_parser.rs
@@ -219,35 +219,19 @@ impl<'a> StringParser<'a> {
                         Some('a') => ConversionFlag::Ascii,
                         Some('r') => ConversionFlag::Repr,
                         Some(_) => {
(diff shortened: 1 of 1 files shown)


compiler/parser/src/string_parser.rs
```diff
@@ -219,35 +219,19 @@ impl<'a> StringParser<'a> {
                         Some('a') => ConversionFlag::Ascii,
                         Some('r') => ConversionFlag::Repr,
                         Some(_) => {
-                            return Err(if expression.trim().is_empty() {
```
harupy, line 222: We have the following code at line 213:

```rust
                    if expression.trim().is_empty() {
                        return Err(EmptyExpression.to_lexical_error(self.get_pos()));
                    }
```

After this if block, `expression.trim().is_empty()` is always false.

compiler/parser/src/string_parser.rs
```diff
@@ -219,35 +219,19 @@ impl<'a> StringParser<'a> {
...
-                                EmptyExpression.to_lexical_error(self.get_pos())
-                            } else {
-                                UnclosedLbrace.to_lexical_error(self.get_pos())
-                            });
+                            return Err(UnclosedLbrace.to_lexical_error(self.get_pos()));
                         }
                     };
 
                     if let Some(&peek) = self.peek() {
```
DimitrisJim, line 229: Pretty sure this can be simplified to:

```rust
match self.peek() {
    Some("}" | ":") => {},
    Some(_) | None => return ...
```

harupy: Great suggestion!

compiler/parser/src/string_parser.rs
```diff
@@ -219,35 +219,19 @@ impl<'a> StringParser<'a> {
...
-                        if peek != '}' && peek != ':' {
-                            return Err(if expression.trim().is_empty() {
-                                EmptyExpression.to_lexical_error(self.get_pos())
-                            } else {
-                                UnclosedLbrace.to_lexical_error(self.get_pos())
-                            });
+                    match self.peek() {
+                        Some('}' | ':') => {}
+                        Some(_) => {
+                            return Err(InvalidConversionFlag.to_lexical_error(self.get_pos()))
```
DimitrisJim, line 232: seems like this should still complain about the unclosed brace, i.e `f"{20!ao}"` results in `SyntaxError: f-string: expecting '}'`

harupy: Looks like I copy & pasted a wrong error. I'll fix this.

harupy: Fixed.

DimitrisJim (approved):

PR #4403, opened 2023-01-02
Fix exponential format

yt2b:
Exponential format raises ValueError.

```
>>>>> f"{1234:.3g}"
ValueError: Unknown format code 'g' for object of type 'int'
>>>>> f"{1234567:.6G}"
ValueError: Unknown format code 'G' for object of type 'int'
```

I've fixed format function.


youknowone (approved):
Thank you!

PR #4411, opened 2023-01-03
Introduce `ImplicitConcat`

harupy:
**I don't think this proposal is acceptable** because the proposed node doesn't exist in Python. I just wanted to hear thoughts on it. This PR adds a new AST node `ImplicitConcat` that represents implicit concatenation (e.g. `"a" "b"`). Currently, implicitly-concatenated strings are merged into a single `Constant` or `JoinedStr` node. This makes it difficult to analyze each string component. Here's an example:

```python
(
  f"a"    # useless f-string
  ""      # useless empty string
  f"{x}"
)
```

This code is parsed as:

```
JoinedStr {
  values: [
    Constant,
    FormattedValue
  ]
}
```

This is is equivalent to `f"a{x}"`, which is a valid f-string.

If the python code above was parsed as:

```
ImplicitConcat {
  values: [
    JoinedStr { values: [Constant] },
    Constnat,
    JoinedStr { values: [FormattedValue] },
  ]
}
```

that would allow a linter such as Ruff to analyze each token and provide more helpful errors.

The following video is a quick demo that utilizes `ImplicitConcat`:

https://user-images.githubusercontent.com/17039389/210311906-3043c61d-b562-44aa-b88f-e752df1f2925.mov

### DIFF
--- a/compiler/ast/Cargo.toml
+++ b/compiler/ast/Cargo.toml
@@ -8,7 +8,7 @@ edition = "2021"
 default = ["constant-optimization", "fold"]
 constant-optimization = ["fold"]
 fold = []
-implicit-concatenation = []
+implicit-concat = []
 unparse = ["rustpython-common"]
 
 [dependencies]
--- a/compiler/ast/Cargo.toml
+++ b/compiler/ast/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2021"
 default = ["constant-optimization", "fold"]
 constant-optimization = ["fold"]
 fold = []
+implicit-concatenation = []
 unparse = ["rustpython-common"]
 
 [dependencies]
--- a/compiler/ast/src/ast_gen.rs
+++ b/compiler/ast/src/ast_gen.rs
@@ -245,8 +245,8 @@ pub enum ExprKind<U = ()> {
     JoinedStr {
         values: Vec<Expr<U>>,
     },
-    #[cfg(feature = "implicit-concatenation")]
-    ImplicitConcatenation {
+    #[cfg(feature = "implicit-concat")]
+    ImplicitConcat {
         values: Vec<Expr<U>>,
     },
     Constant {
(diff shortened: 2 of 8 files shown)


harupy:
Actually, it's possible to break down an implicitly-concatenated string into each component by re-tokenizing and re-parsing `Constant` or `JoinedStr`

DimitrisJim:
I can see the idea here. How do other linters deal with this? Have you looked up if this has been proposed on python-dev or python-ideas?

harupy:
@DimitrisJim Thanks for the comment.

> How do other linters deal with this?

I think other linters just analyze a merged string.

> Have you looked up if this has been proposed on python-dev or python-ideas?

I haven't yet.

harupy:
Looks like pylint re-tokenizes code:

https://github.com/PyCQA/pylint/blob/a57edea63fa7a1edfcb63074f9e2a373ebabb7cd/pylint/utils/utils.py#L179

harupy:
[pyright](https://github.com/microsoft/pyright) seems to have a node for an implicitly-concatenated string (`StringList`): https://github.com/microsoft/pyright/search?q=StringList

harupy:
It's actually possible to break down an implicitly-concatenated string, but it requires re-tokenization and re-parsing, which can degrade the performance.

```python
a = "a" "b"
    ^^^^^^^

# 1. By tokenizing the ^^^ range, we can get two string tokens.
# 2. By parsing each string token, we can get two constant nodes.
```

DimitrisJim:
> which can degrade the performance.

yeah I thought that would be the issue. Not sure why the implicit joining is done before the AST phase, there could be some reason behind it which I'm unaware of. In general, I'd probably ask around other projects to see if something like this would be a feature many would find useful, then a proposal might be made to CPython. 

If that doesn't lead anywhere, sealing it behind a feature _might_ be acceptable but @youknowone and @coolreader18 would have to give the ok for it.

youknowone:
To change ast, as @DimitrisJim said, I think a proposal for CPython would be the best option. Now language developers understand tooling support is really important for the language ecosystem. It will be worth.

In short term, I am open to add features to support toolings. But I also think we have to keep the update cost as small as possible. I'd like to accept it if the patch can make ast updating easy enough.
e.g. we might need to do something like #4042 in future again.

harupy:
@youknowone Thanks for the comment!


> But I also think we have to keep the update cost as small as possible. I'd like to accept it if the patch can make ast updating easy enough.

Yeah, this is what I'm concerned the most. I don't want this node to be a blocker for RustPython's development.

PR #4412, opened 2023-01-03
Implement integer format validation

yt2b:
Integer Format validation is not implemented.

```
>>>>> f"{4096:,o}"
'1,0000'  # Should be raised `ValueError`
>>>>> f"{123456:_n}"
'123_456' # Should be raised `ValueError`
```

I've implemented format validation.
With this fix, following tests now pass.
* `test_format_spec_errors` in `test_types.py`
* `test_int__format__` in `test_types.py`


youknowone (approved):

youknowone:
Thank you!

PR #4414, opened 2023-01-03
Correct exitcode of KeyboardInterrupt

moreal:
This pull requests fixes exitcode of `KeyboardInterrupt`. You can run the below code with CPython and RustPython(before/after). (in unix target os)

```
# RUSTPYTHON
cargo run -- -c "raise KeyboardInterrupt"
echo $?  # before = 1 / after = 130

# CPYTHON
python3 -c "raise KeyboardInterrupt"
echo $?  # 130
```

## Tasks

- [x] macOS
- [x] ubuntu
- ~~windows~~
  - https://docs.rs/winapi/latest/winapi/um/winnt/constant.STATUS_CONTROL_C_EXIT.html
- ~~wasm~~

## Related links

 - https://github.com/python/cpython/blob/8b1f1251215651c4ef988622345c5cb134e54d69/Modules/main.c#L666

### DIFF
--- a/vm/src/vm/mod.rs
+++ b/vm/src/vm/mod.rs
@@ -757,6 +757,19 @@ impl VirtualMachine {
                 writeln!(stderr, "{msg}");
             }
             1
+        } else if exc.fast_isinstance(self.ctx.exceptions.keyboard_interrupt) {
+            #[allow(clippy::if_same_then_else)]
+            {
+                self.print_exception(exc);
+                #[cfg(unix)]
+                {
+                    (libc::SIGINT as u8) + 128u8
+                }
+                #[cfg(not(unix))]
+                {
+                    1
+                }
+            }
         } else {
             self.print_exception(exc);
             1


vm/src/vm/mod.rs
```diff
@@ -757,6 +757,16 @@ impl VirtualMachine {
                 writeln!(stderr, "{msg}");
             }
             1
+        } else if exc.fast_isinstance(self.ctx.exceptions.keyboard_interrupt) {
+            self.print_exception(exc);
+            #[cfg(windows)]
+            {
+                winapi::winnt::STATUS_CONTROL_C_EXIT
```
youknowone, line 764: maybe adding a feature related to winnt will be helpful

moreal:
I tried to implement for targets (not unix) but there were some issues:

 - `std::process::ExitCode` requires exit code be `u8`.
   - It seems there is an extension for Windows case 🤔  https://doc.rust-lang.org/std/os/windows/process/trait.ExitCodeExt.html. Maybe `handle_exit_exception` should be return `ExitCode`.
 - What is `KeyboardInterrupt`'s exit code in WASM(WASI) environment? 🤔 

So this pull request became to implement only for unix now. Other pull request will resolve the problem in windows and wasm instead.

youknowone (approved):
Thank you!

PR #4415, opened 2023-01-03
Remove incorrect `EmptyExpression` in `StringParser.parse_formatted_value`

harupy:
Removes incorrect `EmptyExpression` in `StringParser.parse_formatted_value`.

### DIFF
--- a/compiler/parser/src/error.rs
+++ b/compiler/parser/src/error.rs
@@ -42,7 +42,7 @@ impl fmt::Display for LexicalErrorType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             LexicalErrorType::StringError => write!(f, "Got unexpected string"),
-            LexicalErrorType::FStringError(error) => write!(f, "Got error in f-string: {error}"),
+            LexicalErrorType::FStringError(error) => write!(f, "f-string: {error}"),
             LexicalErrorType::UnicodeError => write!(f, "Got unexpected unicode"),
             LexicalErrorType::NestingError => write!(f, "Got unexpected nesting"),
             LexicalErrorType::IndentationError => {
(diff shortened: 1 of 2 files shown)


compiler/parser/src/string_parser.rs
```diff
@@ -358,11 +358,7 @@ impl<'a> StringParser<'a> {
                 }
             }
         }
-        Err(if expression.trim().is_empty() {
-            FStringError::new(EmptyExpression, self.get_pos()).into()
-        } else {
-            FStringError::new(UnclosedLbrace, self.get_pos()).into()
-        })
+        Err(FStringError::new(UnclosedLbrace, self.get_pos()).into())
```
harupy, line 361: If we reach here, that means we didn't encounter `}`. That means `{` is not closed.

compiler/parser/src/error.rs
```diff
@@ -42,7 +42,7 @@ impl fmt::Display for LexicalErrorType {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
             LexicalErrorType::StringError => write!(f, "Got unexpected string"),
-            LexicalErrorType::FStringError(error) => write!(f, "Got error in f-string: {error}"),
+            LexicalErrorType::FStringError(error) => write!(f, "f-string: {error}"),
```
harupy, line 45: We don't need `Got error in`.

```python
> python
>>> f"{"
  File "<stdin>", line 1
    f"{"
        ^
SyntaxError: f-string: expecting '}'
```

compiler/parser/src/string_parser.rs
harupy: In cpython:

```
>>> f"{   "
  File "<stdin>", line 1
SyntaxError: f-string: expecting '}'
```

DimitrisJim (approved):

PR #4418, opened 2023-01-04
Clean up string check in `Lexer`

harupy:
Cleans up string check in `Lexer`.

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -244,55 +244,37 @@ where
         let mut name = String::new();
         let start_pos = self.get_pos();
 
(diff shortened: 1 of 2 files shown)


compiler/parser/src/lexer.rs
```diff
@@ -244,55 +244,37 @@ where
...
+                    self.next_char().unwrap();
+                    self.next_char().unwrap();
+                    return self
+                        .lex_string(format!("{c1}{c2}").try_into().unwrap())
+                        .map(|(_, tok, end_pos)| (start_pos, tok, end_pos));
+                }
+                _ => {}
+            },
+            _ => {}
+        };
```
harupy, line 277: We know the pattern a string should start with. The new code just checks it.

PR #4419, opened 2023-01-04
Implement float format validation

yt2b:
Float Format validation is not implemented.

```
>>>>> f"{1.0:,o}"
ValueError: Unknown format code 'o' for object of type 'float'
```
The output should be raised `ValueError: Cannot specify ',' with 'o'.`
I've implemented format validation.


DimitrisJim (approved):
lgtm!

PR #4420, opened 2023-01-04
Simplify string check in `Lexer.lex_identifier`

harupy:
Simplifies string check in `Lexer.lex_identifier`.

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -241,59 +241,23 @@ where
 
     // Lexer helper functions:
     fn lex_identifier(&mut self) -> LexResult {
-        let mut name = String::new();
-        let start_pos = self.get_pos();
-
         // Detect potential string like rb'' b'' f'' u'' r''
-        let mut saw_b = false;
-        let mut saw_r = false;
-        let mut saw_u = false;
-        let mut saw_f = false;
-        loop {
-            // Detect r"", f"", b"" and u""
-            if !(saw_b || saw_u || saw_f) && matches!(self.window[0], Some('b' | 'B')) {
-                saw_b = true;
-            } else if !(saw_b || saw_r || saw_u || saw_f)
-                && matches!(self.window[0], Some('u' | 'U'))
-            {
-                saw_u = true;
-            } else if !(saw_r || saw_u) && matches!(self.window[0], Some('r' | 'R')) {
-                saw_r = true;
-            } else if !(saw_b || saw_u || saw_f) && matches!(self.window[0], Some('f' | 'F')) {
-                saw_f = true;
-            } else {
-                break;
+        match self.window[..3] {
+            [Some(c), Some('"' | '\''), ..] => {
+                if let Ok(kind) = c.to_string().try_into() {
+                    return self.lex_string(kind);
+                }
             }
-
-            // Take up char into name:
-            name.push(self.next_char().unwrap());
-
-            // Check if we have a string:
-            if matches!(self.window[0], Some('"' | '\'')) {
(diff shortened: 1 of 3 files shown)


compiler/parser/src/lexer.rs
```diff
@@ -241,59 +241,23 @@ where
...
     // Lexer helper functions:
     fn lex_identifier(&mut self) -> LexResult {
-        let mut name = String::new();
-        let start_pos = self.get_pos();
-
         // Detect potential string like rb'' b'' f'' u'' r''
-        let mut saw_b = false;
-        let mut saw_r = false;
-        let mut saw_u = false;
-        let mut saw_f = false;
```
harupy, line 251: I think the current logic is a bit difficult to verify if it's correct.

compiler/parser/src/token.rs
```diff
@@ -240,6 +240,23 @@ pub enum StringKind {
     Unicode,
 }
 
+impl TryFrom<String> for StringKind {
+    type Error = String;
+
+    fn try_from(value: String) -> Result<Self, String> {
+        match value.as_str() {
+            "" => Ok(StringKind::String),
```
DimitrisJim, line 248: does this ever get hit? `lex_identifier` is called when the starting char is some character that starts identifiers so the cases for `r`, `u`, `b`, `f` etc to pass through.

compiler/parser/src/token.rs
```diff
@@ -240,6 +240,23 @@ pub enum StringKind {
     Unicode,
 }
 
+impl TryFrom<String> for StringKind {
```
DimitrisJim, line 243: instead of requiring an allocation here, why not `TryFrom` for `char` and `[char; 2]`? This is slightly more code but should avoid the `to_string` calls in `lex_identifier`.

compiler/parser/src/lexer.rs
```diff
@@ -495,6 +459,9 @@ where
 
     fn lex_string(&mut self, kind: StringKind) -> LexResult {
         let start_pos = self.get_pos();
+        for _ in 0..kind.to_string().len() {
```
DimitrisJim, line 462: we can also avoid this allocation by adding a `len`-like method on `StringKind` that returns the correct amount based on the kind.

compiler/parser/src/lexer.rs
DimitrisJim: oh definitely, the change makes it much better.

compiler/parser/src/token.rs
DimitrisJim: @youknowone any better ideas?

compiler/parser/src/lexer.rs
fanninpm: I think this originally came from the [tokenizer code](https://github.com/python/cpython/blob/3.11/Parser/tokenizer.c).

compiler/parser/src/token.rs
harupy: > why not TryFrom for char and [char; 2]?

Sounds good. I'll make that change.

compiler/parser/src/token.rs
harupy: You're right. It never hits. I'll remove it.

compiler/parser/src/lexer.rs
harupy: Agreed. I'll add a `len` method.

harupy: Added a `prefix_len` method.

DimitrisJim (approved):
great! looks much better, thanks!

Issue #3992, opened 2022-08-02
`&=` and `-=` operator between Set types do not work correctly.


tgsong827:
## Feature

I'm trying to fix `@unittest.expectedFailure` `test_inplace_on_self` in file `test_set.py`.
```
def test_inplace_on_self(self):
    t = self.s.copy()
    t |= t
    self.assertEqual(t, self.s)
    t &= t
    self.assertEqual(t, self.s).      -> FAIL
    t -= t
    self.assertEqual(t, self.thetype()).     -> FAIL
    t = self.s.copy()
    t ^= t
    self.assertEqual(t, self.thetype())
```

I found the two operator fail.
`&=` and `-=` is not work correctly.

### [ `&=` ]
- RustPython
```
>>>>> s = set({'a', 'b', 'c'})
>>>>> s &= s
>>>>> s
set()
```
- CPython
```
>>> s = set({'a', 'b', 'c'})
>>> s &= s
>>> s
{'b', 'c', 'a'}
```

### [ `-=` ]
- RustPython
```
>>>>> s = set({'a', 'b', 'c'})
>>>>> s -= s
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: set changed size during iteration
```

### Maybe the points to start
- `&=`
```
#[pymethod(magic)]
fn iand(zelf: PyRef<Self>, set: AnySet, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
    zelf.inner
        .intersection_update(std::iter::once(set.into_iterable(vm)?), vm)?;
    Ok(zelf)
}
```
- `-=`
```
#[pymethod(magic)]
fn isub(zelf: PyRef<Self>, set: AnySet, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
    zelf.inner
        .difference_update(set.into_iterable_iter(vm)?, vm)?;
    Ok(zelf)
}
```

### CPython Method
- `&=`
https://github.com/python/cpython/blob/75a6441718dcbc65d993c9544e67e25bef120e82/Objects/setobject.c#L1181-L1261

- `-=`
https://github.com/python/cpython/blob/75a6441718dcbc65d993c9544e67e25bef120e82/Objects/setobject.c#L1406-L1461

tgsong827:
This may be related with #3881 #3912.
Can I try it after #3912 is merged?

arihant2math:
`&=` now works, `-=` still fails

PR #4424, opened 2023-01-06
Fix set bugs with `&=` and `-=`

### LINKED ISSUE #3881: `set` `intersection_update`, `difference_update`, `symmetric_difference_update` should check self is equal to given arg

branai:
I changed the logic for `intersection_update` and `difference_update` to fix some bugs with set operations that call these functions.

Fixes #3881 #3992

### DIFF
--- a/vm/src/builtins/set.rs
+++ b/vm/src/builtins/set.rs
@@ -383,16 +383,15 @@ impl PySetInner {
         others: impl std::iter::Iterator<Item = ArgIterable>,
         vm: &VirtualMachine,
     ) -> PyResult<()> {
-        let mut temp_inner = self.copy();
-        self.clear();
-        for iterable in others {
-            for item in iterable.iter(vm)? {
-                let obj = item?;
-                if temp_inner.contains(&obj, vm)? {
-                    self.add(obj, vm)?;
+        // For each set in other
+        //  remove all elements from LHS (self) that are not in set
+        for other_set_iter in others {
+            let other_set = Self::from_iter(other_set_iter.iter(vm)?, vm)?;
+            for item in self.elements().into_iter() {
+                if !other_set.contains(&item, vm)? {
+                    self.remove(item, vm)?;
                 }
             }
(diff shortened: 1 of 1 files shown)


fanninpm:

```shell
```

branai:
If I am failing tests with
```
# TODO: RUSTPYTHON
    @unittest.expectedFailure
```
should I remove them?

DimitrisJim:
So the main idea here is that we want to skip iterations when we intersect/difference/symmetric_difference a set with itself. Though the PR fixes faulty assumptions in `intersect_update` and `difference_update` (the assumption that the set(s) we supply as an argument isn't the same set as `self`) it doesn't do the short-circuiting, i.e `s = {*range(20)}; s.intersection_update(s, s, s)` would iterate over every element in the arguments when it could just skip them all since `s &= s` is `s`.

The best approach is along the lines of #3912, we accept objects as arguments for `intersection_update/difference_update` (`AnySet` works fine for `iand/isub``) and compare these objects with `is`, this is fast `O(1)` and allows us to ignore sets or clear them depending on the function.

PR #4426, opened 2023-01-06
Include comment text in token

harupy:
Include comment text in the token for a linter like Ruff which checks comments to ignore errors.


DimitrisJim:
You use the comment text in some way in Ruff? IIRC CPython doesn't keep the text around but I could be remembering wrong.

fanninpm:
I'm assuming @charliermarsh wants something like the `# noqa: E404` annotations in Flake8. It turns out that Flake8 grabs the physical line and regexes it (which is rather inefficient).

harupy:
@DimitrisJim 

> CPython doesn't keep the text around but I could be remembering wrong.

```
> cat a.py
# foo
print(1)

> python -m tokenize a.py
0,0-0,0:            ENCODING       'utf-8'        
1,0-1,5:            COMMENT        '# foo'        
1,5-1,6:            NL             '\n'           
2,0-2,5:            NAME           'print'        
2,5-2,6:            OP             '('            
2,6-2,7:            NUMBER         '1'            
2,7-2,8:            OP             ')'            
2,8-2,9:            NEWLINE        ''             
3,0-3,0:            ENDMARKER      ''
```

harupy:
@DimitrisJim 

https://github.com/python/cpython/blob/15c44789bb125b93e96815a336ec73423c47508e/Parser/tokenizer.c#L1603

```
    /* Skip comment, unless it's a type comment */
    if (c == '#') {
        const char *prefix, *p, *type_start;
        int current_starting_col_offset;

        while (c != EOF && c != '\n') {
            c = tok_nextc(tok);
        }

        if (tok->type_comments) {
            p = tok->start;
            current_starting_col_offset = tok->starting_col_offset;
            prefix = type_comment_prefix;
            while (*prefix && p < tok->cur) {
                if (*prefix == ' ') {
                    while (*p == ' ' || *p == '\t') {
                        p++;
                        current_starting_col_offset++;
                    }
                } else if (*prefix == *p) {
                    p++;
                    current_starting_col_offset++;
                } else {
                    break;
                }

                prefix++;
            }

```

charliermarsh:
Right now, we use the locations on the token and extract the comments from the source code. It works fine. It'd be convenient to have the comment text directly, and likely more performant for Ruff, but if it's a CPython incompatibility and the RustPython team would prefer not to include it, I'd understand that too.

harupy:
@charliermarsh Thanks for the comment!

harupy:
https://github.com/python/cpython/blob/72263f2a20002ceff443e3a231c713f2e14fe3fe/Lib/tokenize.py#L17

```
It is designed to match the working of the Python tokenizer exactly, except
that it produces COMMENT tokens for comments and gives type OP for all
operators.  Additionally, all token lists start with an ENCODING token
which tells you which encoding was used to decode the bytes stream.
```

charliermarsh:
@harupy - Always grateful for all the work you're doing to improve Ruff and RustPython :)

DimitrisJim:
> Right now, we use the locations on the token and extract the comments from the source code.

Oh, that's rough, pun intended. I think its fine if the comment contents are caught since it doesn't change the semantics in any way, its a relatively small change that's easily maintainable and it helps the Ruff team simplify things.

harupy:
@DimitrisJim Awesome! I love the pun.

DimitrisJim (approved):
lgtm, let's let @youknowone also give the ok and merge it.

compiler/parser/python.lalrpop
```diff
@@ -1463,6 +1463,6 @@ extern {
         name => lexer::Tok::Name { name: <String> },
         "\n" => lexer::Tok::Newline,
         ";" => lexer::Tok::Semi,
-        "#" => lexer::Tok::Comment,
+        "#" => lexer::Tok::Comment(_),
```
harupy, line 1466: We never hit this arm because we only parse non-comment tokens.

charliermarsh:
Thanks @DimitrisJim, grateful for all the collaboration that's happening between these two projects!

youknowone (approved):
This incompatibility is not a big deal. Let's go.

PR #4427, opened 2023-01-07
Fix integer format

yt2b:
Panic occured when formatted from integer to multibyte characater.

```
>>>>> f"{0x1f5a5:c}"
thread 'main' panicked at 'assertion failed: s.is_ascii()', common/src/str.rs:75:9
```

I've implemented the check whether string contains multibyte character.


youknowone (approved):
Thank you!

Issue #4387, opened 2022-12-31
`PyInt.format` only raises `ValueError`


yt2b:
In the current implementation, `PyInt.format` only raises `ValueError`.

https://github.com/RustPython/RustPython/blob/2c8af125e396f9dee73bd426a0bc088b60c30969/vm/src/builtins/int.rs#L569-L574

But in CPython, Other error can also be raised.

```
>>> f"{-1:c}"
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OverflowError: %c arg not in range(0x110000)
```

yt2b:
@DimitrisJim 
I'll work on this issue.

DimitrisJim:
great, thanks for your recent work on the formatting @yt2b!

PR #4428, opened 2023-01-07
`PyInt.format` only raises `ValueError`

yt2b:
#4387
I've fixed to raise `OverflowError`.

```
>>>>> f"{-1:c}"
OverflowError: %c arg not in range(0x110000)
```

### DIFF
--- a/common/src/format.rs
+++ b/common/src/format.rs
@@ -223,7 +223,7 @@ fn parse_fill_and_align(text: &str) -> (Option<char>, Option<FormatAlign>, &str)
     }
 }
 
-fn parse_number(text: &str) -> Result<(Option<usize>, &str), &'static str> {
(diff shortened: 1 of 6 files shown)


common/src/format.rs
```diff
@@ -271,6 +271,8 @@ fn parse_precision(text: &str) -> Result<(Option<usize>, &str), &'static str> {
 }
 
 impl FormatSpec {
+    pub const NOT_IN_RANGE: &str = "%c arg not in range(0x110000)";
```
youknowone, line 274: We'd better to define a new enum for this type.

```rust
enum FormatSpecError {
     ...
}

impl IntoPyException for FormatSpecError {
   ...
}
```

This is also a possible fix of the problem, but I'd more like to avoid string comparison to decide the error type from python side.

yt2b: Code fix completed.

youknowone (approved):

PR #4430, opened 2023-01-08
Implement strict mode for binascii a2b_base64()

### LINKED ISSUE #4374: Implement `strict_mode` keyword parameter in `binascii.a2b_base64()`
Since Python 3.11, `binascii.a2b_base64()` function became to provide `strict_mode` keyword parameter. [^1]

[^1]: https://docs.python.org/3/library/binascii.html?highlight=binascii#binascii.a2b_base64

## Related links

 - Python docs: https://docs.python.org/3/library/binascii.html?highlight=binascii#binascii.a2b_base64
 - RustPython implementation: https://github.com/RustPython/RustPython/blob/1d8269fb729c91fc56064e975172d3a11bd62d07/stdlib/src/binascii.rs#L147
 - CPython implementation: https://github.com/python/cpython/blob/deaf509e8fc6e0363bd6f26d52ad42f976ec42f2/Modules/binascii.c#L388

evanrittenhouse:
There's one TODO that I had - I'm not sure how to return different errors. It seems like the compiler always expects `DecodeError::InvalidLength` errors, so submitting this in a semi-draft state for some guidance. I'm also not super sure how the tests work, but I ran `pytest -v` in `extra_tests/` and they all passed. Please let me know if there's something else that I should do!

Should fix #4374 when finished.

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -1075,10 +1075,13 @@ where
                 self.next_char();
                 let tok_end = self.get_pos();
 
-                // Depending on the nesting level, we emit newline or not:
+                // Depending on the nesting level, we emit a logical or
+                // non-logical newline:
                 if self.nesting == 0 {
                     self.at_begin_of_line = true;
                     self.emit((tok_start, Tok::Newline, tok_end));
+                } else {
+                    self.emit((tok_start, Tok::NonLogicalNewline, tok_end));
                 }
             }
             ' ' | '\t' | '\x0C' => {
@@ -1464,7 +1467,16 @@ mod tests {
         $(
             #[test]
             fn $name() {
-                let source = format!("x = [{}    1,2{}]{}", $eol, $eol, $eol);
+                let source = r"x = [
+
+    1,2
+,(3,
+4,
+), {
+5,
+6,\
+7}]
+".replace("\n", $eol);
                 let tokens = lex_source(&source);
                 assert_eq!(
                     tokens,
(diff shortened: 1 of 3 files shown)


moreal:
> I'm also not super sure how the tests work, but I ran pytest -v in extra_tests/ and they all passed.

You can manually run `cargo run -- Lib/test/test_binascii.py` command to test `binascii` module instead of running `pytest`. In `Lib/test` directory, there are unittests to test modules.

DimitrisJim:
>  I'm not sure how to return different errors. It seems like the compiler always expects DecodeError::InvalidLength errors

It might be because its early but what is the compiler saying? As long as they are variants of the `DecodeError` enum there shouldn't be an issue, they are formatted in the `binascii` error returned at the end.

evanrittenhouse:
Not sure why this got closed as I'm still working on it. May be a bit as I've got a lot to learn.

DimitrisJim:
> Not sure why this got closed as I'm still working on it

It shows me that you closed it? Seems like this is probably automatic after your force push to 91a897d (to where `main` points), it sees that no files are changed and just closes it (maybe you made a small mix up with the merge?)

evanrittenhouse:
Ah gotcha. Okay, makes sense. I'm curious - do we want to preserve the current error messages? It seems like `base64::DecodeError` has its own error messages which don't line up with the current CPython implementation's error messages. 

If we want to change them, how do you recommend doing so? You're right in that the compiler requires variants of `DecodeError`. I was thinking of maybe checking the `DecodeError` variant when formatting the message and printing that, but not sure if there's something else you'd like to do!

evanrittenhouse:
@DimitrisJim I've also got a question about a test case like [this](https://github.com/RustPython/RustPython/blob/main/Lib/test/test_binascii.py#L156). `b'='` should also throw an `InvalidLength` - do we want to surface that `InvalidLength` issue or the `LeadingPadding` one? What about other cases where several errors are present?

DimitrisJim:
hey @evanrittenhouse, sorry for the delay, I'll look into your new PR over the weekend.

> do we want to surface that InvalidLength issue or the LeadingPadding one? What about other cases where several errors are present?

Whatever CPython does, if that means we ignore some errors or convert between errors (if that might be the only option), so be it.

Issue #4179, opened 2022-09-21
Publishing a `rustpython-parser` fork to crates.io?


charliermarsh:
I'm using RustPython as the underlying parser for [ruff](https://github.com/charliermarsh/ruff), a Python linter written in Rust. ruff uses `rustpython-parser`, but not the interpreter / runtime.

ruff is installable via PyPI, but I'd like to distribute it via cargo (on crates.io) as well. The Git dependency on RustPython (and on my [RustPython fork](https://github.com/charliermarsh/RustPython)) makes that impossible right now. (Installing via GIt has also led to flakiness in CI / builds.)

I'd like to publish my RustPython fork to crates.io to enable installing ruff via cargo, but I wanted to check here before doing so as I'm a big fan of this project (as well as a contributor :)) and wouldn't want to cause any issues. Do the maintainers project have any objection? My plan was to publish as `ruff-rustpython-parser` or ` ruff-parser` or something along those lines though am of course happy to publish under any name.

charliermarsh:
FTR, I initially tried to vendor my fork within ruff (then set the `path` in Cargo.toml) but that's _also_ unsupported by Cargo.

lastmjs:
It would be great to get RustPython back up on crates.io!

youknowone:
seriously, I want to do this soon.

fanninpm:
> seriously, I want to do this soon.

That's why I assigned @coolreader18.

youknowone:
Let me work on it during this coming winter. Recently RustPython is looking like to be used from more serious projects than before. I feel like a nice packaging will be more worth than a few more features now.

mendelsshop:
Hi there, sorry to budge in,

I have a crate [git function history](https://github.com/mendelsshop/git_function_history), which only supported rust until now, but now I will be supporting Python, and a few other languages. I am using RustPython to parse python files, I need to get the end locations of `Located`<`StmtKind`> by getting the `end_location` field on `Located`, which requires using the git dependency of RustPython instead of the version from crate.io. Now I want to publish my new version to crates.io, which as mentioned before is not possible using git dependencies or via path.

not-my-profile:
There already are [6 rustpython crates](https://crates.io/search?q=rustpython) published on crates.io, only that they have not been updated in over 2 years. I think it would make more sense to update these crates rather than forking the project.

Why exactly haven't these crates been updated? What are the exact steps that are needed to publish new versions of them?

fanninpm:
[...]

PR #4431, opened 2023-01-09
Make version 0.2.0 to prepare a new release

youknowone:
ref #4179


coolreader18:
Oh, I didn't see that this was a PR, sorry. This is superceded by the commit I just pushed to `main`, I guess.

youknowone:
@coolreader18 your commit didn't add versions to the crates which didn't have versions before. would you merge the diff if we need it?

PR #4434, opened 2023-01-09
Implement itertools.accumulate.__reduce__ and __setstate__

lgyanf:
Add accumulate.__reduce__ implementation for #3611.

Publishing as a draft to collect initial feedback. Behaviour seems to be identical to CPython's, but I'm still testing.

# Relevant links
- https://github.com/python/cpython/blob/main/Modules/itertoolsmodule.c#L3858-L3894


fanninpm:
Is this related to #3611?

youknowone:
Because you fixed related bug from the test suites, you can remove `@unittest.expectedFailure` decorator from `test_accumulate_reducible` test

fanninpm:
cc @Snowapril

youknowone (approved):

PR #4442, opened 2023-01-12
lexer: Add `NonLogicalNewline` token

### LINKED ISSUE #4385: Tokenizer is missing `NL` token
The CPython tokenizer has both a `NEWLINE` and `NL` token.

Given this code:

```python
x = (
  "a"
  "b"
)

x = "a" \
  "b"
```

RustPython currently outputs the same tokenization for these two snippets (modulo the left and right parens -- notice that in both cases, we have `String` followed by `String`):

```
Name {
    name: "x",
}
Equal
Lpar
String {
    value: "a",
    kind: Normal,
}
String {
    value: "b",
    kind: Normal,
}
Rpar
Newline
Name {
    name: "x",
}
Equal
String {
    value: "a",
    kind: Normal,
}
String {
    value: "b",
    kind: Normal,
}
Newline
```

Whereas in CPython, there are `NL` tokens in the first case that aren't present for continuations:

```
❯ python -m tokenize test.py
0,0-0,0:            ENCODING       'utf-8'
1,0-1,1:            NAME           'x'
1,2-1,3:            OP             '='
1,4-1,5:            OP             '('
1,5-1,6:            NL             '\n'
2,2-2,5:            STRING         '"a"'
2,5-2,6:            NL             '\n'
3,2-3,5:            STRING         '"b"'
3,5-3,6:            NL             '\n'
4,0-4,1:            OP             ')'
4,1-4,2:            NEWLINE        '\n'
5,0-5,1:            NL             '\n'
6,0-6,1:            NAME           'x'
6,2-6,3:            OP             '='
6,4-6,7:            STRING         '"a"'
7,2-7,5:            STRING         '"b"'
7,5-7,6:            NEWLINE        '\n'
8,0-8,0:            ENDMARKER      ''
```

bluetech:
This token is completely ignored by the parser, but it's useful for other users of the lexer, such as the Ruff linter. For example, the token is helpful for a "trailing comma" lint.

The same idea exists in Python's `tokenize` module - there is a NEWLINE token (logical newline), and a NL token (non-logical newline).

Fixes #4385.

### DIFF
--- a/compiler/parser/src/token.rs
+++ b/compiler/parser/src/token.rs
@@ -25,6 +25,7 @@ pub enum Tok {
         triple_quoted: bool,
     },
     Newline,
+    NonLogicalNewline,
     Indent,
     Dedent,
     StartModule,
@@ -136,6 +137,7 @@ impl fmt::Display for Tok {
                 write!(f, "{kind}{quotes}{value}{quotes}")
             }
             Newline => f.write_str("Newline"),
+            NonLogicalNewline => f.write_str("NonLogicalNewline"),
             Indent => f.write_str("Indent"),
             Dedent => f.write_str("Dedent"),
             StartModule => f.write_str("StartProgram"),
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -1075,10 +1075,13 @@ where
                 self.next_char();
                 let tok_end = self.get_pos();
 
-                // Depending on the nesting level, we emit newline or not:
+                // Depending on the nesting level, we emit a logical or
+                // non-logical newline:
                 if self.nesting == 0 {
                     self.at_begin_of_line = true;
                     self.emit((tok_start, Tok::Newline, tok_end));
+                } else {
+                    self.emit((tok_start, Tok::NonLogicalNewline, tok_end));
                 }
             }
             ' ' | '\t' | '\x0C' => {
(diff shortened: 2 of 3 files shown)


compiler/parser/src/token.rs
```diff
@@ -25,6 +25,7 @@ pub enum Tok {
         triple_quoted: bool,
     },
     Newline,
+    NonLogicalNewline,
```
DimitrisJim, line 28: I think `PhysicalNewLine` is better tbh, better suited when someone reads the [Lexical Analysis](https://docs.python.org/3/reference/lexical_analysis.html) section of the reference. Logical line there is basically defined as `NewLine` so we're good.

bluetech: I was also initially thinking of calling it `PhysicalNewLine`, but the concept doesn't match exactly. Technically, every logical new line is also a physical new line, by the Language Reference definition. If we were issuing a `PhysicalNewLine` before or after every `Newline`, it would match, but I chose to go with tokenize.py which doesn't do that.

DimitrisJim: Sounds reasonable enough to me, I don't think there's much point in bike-shedding something like this.

DimitrisJim (approved):
thanks for this!

charliermarsh:
Oh, awesome! Thanks!

PR #4443, opened 2023-01-12
Fixup parse_tokens after "Add NonLogicalNewline token"

bluetech:
Fixup for commit e64faa781c1faa9052a9d57c00d29830d2f11850. I only updated `parse()` to ignore `NonLogicalNewline`, didn't notice it's also needed in `parse_tokens()`. Now I've ran the full Ruff test suite, and it passes (except for a couple intended fixes for the new token).

### DIFF
--- a/compiler/parser/src/parser.rs
+++ b/compiler/parser/src/parser.rs
@@ -93,14 +93,7 @@ pub fn parse_located(
     location: Location,
 ) -> Result<ast::Mod, ParseError> {
     let lxr = lexer::make_tokenizer_located(source, location);
-    let marker_token = (Default::default(), mode.to_marker(), Default::default());
-    let tokenizer = iter::once(Ok(marker_token))
-        .chain(lxr)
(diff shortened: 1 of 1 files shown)


compiler/parser/src/parser.rs
```diff
@@ -112,7 +112,7 @@ pub fn parse_tokens(
     let marker_token = (Default::default(), mode.to_marker(), Default::default());
     let tokenizer = iter::once(Ok(marker_token))
         .chain(lxr)
-        .filter_ok(|(_, tok, _)| !matches!(tok, Tok::Comment(_)));
+        .filter_ok(|(_, tok, _)| !matches!(tok, Tok::Comment { .. } | Tok::NonLogicalNewline));
```
DimitrisJim, line 115: completely missed this too. could you just move the filtering into a different function instead? I doubt there's going to be many tokens that are filtered but its best if we keep it in a single function to avoid forgetting it in the future.

bluetech: Good idea; updated so that `parse_located` just calls `parse_tokens`, removing the entire duplication.

DimitrisJim (approved):

PR #4444, opened 2023-01-13
Fix panic when parsing a format spec with multi-byte character

messense:
[String::split_at](https://doc.rust-lang.org/std/string/struct.String.html#method.split_at) panics if `mid` is not on a UTF-8 code point boundary, changed to use [`String::char_indices`](https://doc.rust-lang.org/std/string/struct.String.html#method.char_indices) instead of `String::chars().enumerate()` to fix it.

See also https://github.com/charliermarsh/ruff/issues/1845

### DIFF
--- a/common/src/format.rs
+++ b/common/src/format.rs
@@ -869,7 +869,7 @@ impl FormatString {
         let mut left = String::new();
 
         // There may be one layer nesting brackets in spec
-        for (idx, c) in text.chars().enumerate() {
+        for (idx, c) in text.char_indices() {
             if idx == 0 {
                 if c != '{' {
                     return Err(FormatParseError::MissingStartBracket);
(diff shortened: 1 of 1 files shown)


not-my-profile:
(ruff will need to upgrade its rustpython dependency for that issue to be fixed so you may want to change "fixes" in the PR description to "See also" so that the ruff issue isn't closed too early)

messense:
```bash
git grep 'chars().enumerate()'
common/src/cformat.rs:        let mut chars = text.chars().enumerate().peekable();
common/src/cformat.rs:        let mut iter = text.chars().enumerate().peekable();
common/src/float_ops.rs:            for (index, ch) in value.chars().enumerate() {
```

BTW, someone should also audit the other similar cases.

youknowone (approved):
Thank you!

Issue #4441, opened 2023-01-11
Alternate float format doesn't work correctly.


yt2b:
```
>>>>> f"{1.0:#.0f}"
'1'
```

The output should be `'1.'`.
I'll look into this.

PR #4446, opened 2023-01-13
Alternate float format doesn't work correctly.

yt2b:
#4441 

I've implemented Alternate float format.
* `test_format_testfile` in `test_float.py`
* `test_float__format__` in `test_types.py`


common/src/float_ops.rs
```diff
@@ -124,9 +124,19 @@ fn format_inf(case: Case) -> String {
     inf.to_string()
 }
 
-pub fn format_fixed(precision: usize, magnitude: f64, case: Case) -> String {
+pub fn get_dot(precision: usize, alternate_form: bool) -> &'static str {
```
youknowone, line 127:
```suggestion
pub fn decimal_point_or_empty(precision: usize, alternate_form: bool) -> &'static str {
```

youknowone (approved):
Thanks a lot. I left a name suggestion for more descriptive function name. Please check it and if it fits in the purpose.

common/src/float_ops.rs
yt2b: Code fix completed.

PR #4449, opened 2023-01-15
Fix AST generated from a dict literal containing dict unpacking

harupy:
Related to https://github.com/charliermarsh/ruff/issues/1879.

`{"foo": 1, **{"bar": 1}}` is parsed into the AST below in CPython and RustPython. In CPython, `Dict.keys` contains two keys (`Constant` and `None`), but in RustPython, `Dict.keys` only contains one key. This PR fixes it.

### CPython:

<details>
<summary>Expand</summary>

```
Module(
  body=[
    Expr(
      value=Dict(
        👇👇👇 Contains two keys, but the second key is None
        keys=[
          Constant(
            value='foo',
            lineno=1,
            col_offset=1,
            end_lineno=1,
            end_col_offset=6),
          None],
        values=[
          Constant(
            value=1,
            lineno=1,
            col_offset=8,
            end_lineno=1,
            end_col_offset=9),
          Dict(
            keys=[
              Constant(
                value='bar',
                lineno=1,
                col_offset=14,
                end_lineno=1,
                end_col_offset=19)],
            values=[
              Constant(
                value=1,
                lineno=1,
                col_offset=21,
                end_lineno=1,
                end_col_offset=22)],
            lineno=1,
            col_offset=13,
            end_lineno=1,
            end_col_offset=23)],
        lineno=1,
        col_offset=0,
        end_lineno=1,
        end_col_offset=24),
      lineno=1,
      col_offset=0,
      end_lineno=1,
      end_col_offset=24)],
  type_ignores=[])
```
</details>

### RustPython

<details>
<summary>Expand</summary>

```
[
    Located {
        location: Location {
            row: 1,
            column: 0,
        },
        end_location: Some(
            Location {
                row: 1,
                column: 24,
            },
        ),
        custom: (),
        node: Expr {
            value: Located {
                location: Location {
                    row: 1,
                    column: 0,
                },
                end_location: Some(
                    Location {
                        row: 1,
                        column: 24,
                    },
                ),
                custom: (),
                node: Dict {
                    👇👇👇 contains one key
                    keys: [
                        Located {
                            location: Location {
                                row: 1,
                                column: 1,
                            },
                            end_location: Some(
                                Location {
                                    row: 1,
                                    column: 6,
                                },
                            ),
                            custom: (),
                            node: Constant {
                                value: Str(
                                    "foo",
                                ),
                                kind: None,
                            },
                        },
                    ],
                    👇👇👇 values look correct?
                    values: [
                        Located {
                            location: Location {
                                row: 1,
                                column: 8,
                            },
                            end_location: Some(
                                Location {
                                    row: 1,
                                    column: 9,
                                },
                            ),
                            custom: (),
                            node: Constant {
                                value: Int(
                                    1,
                                ),
                                kind: None,
                            },
                        },
                        Located {
                            location: Location {
                                row: 1,
                                column: 13,
                            },
                            end_location: Some(
                                Location {
                                    row: 1,
                                    column: 23,
                                },
                            ),
                            custom: (),
                            node: Dict {
                                keys: [
                                    Located {
                                        location: Location {
                                            row: 1,
                                            column: 14,
                                        },
                                        end_location: Some(
                                            Location {
                                                row: 1,
                                                column: 19,
                                            },
                                        ),
                                        custom: (),
                                        node: Constant {
                                            value: Str(
                                                "bar",
                                            ),
                                            kind: None,
                                        },
                                    },
                                ],
                                values: [
                                    Located {
                                        location: Location {
                                            row: 1,
                                            column: 21,
                                        },
                                        end_location: Some(
                                            Location {
                                                row: 1,
                                                column: 22,
                                            },
                                        ),
                                        custom: (),
                                        node: Constant {
                                            value: Int(
                                                1,
                                            ),
                                            kind: None,
                                        },
                                    },
                                ],
                            },
                        },
                    ],
                },
            },
        },
    },
]
```

</details>


compiler/parser/python.lalrpop
```diff
@@ -1138,42 +1138,42 @@ Atom<Goal>: ast::Expr = {
...
+        //             }
+        //             values.push(value);
+        //         }
+        //         if !items.is_empty() {
+        //             values.push(build_map(&mut items));
+        //         }
+        //         (keys, values)
+        //     },
+        //     None => pairs.into_iter().map(|(k, v)| (k, v)).unzip()
+        // };
```
harupy, line 1175:
```suggestion
```

compiler/parser/src/snapshots/rustpython_parser__parser__tests__dict_containing_spread.snap
```diff
@@ -0,0 +1,81 @@
...
+                ),
+                custom: (),
+                node: Name {
+                    id: "d",
+                    ctx: Load,
+                },
+            },
+        ],
+    },
+}
```
harupy, line 81: CPython's AST:

```
Module(
  body=[
    Expr(
      value=Dict(
        keys=[
          Constant(
            value='k',
            lineno=1,
            col_offset=1,
            end_lineno=1,
            end_col_offset=4),
          None],
        values=[
          Constant(
            value='v',
            lineno=1,
            col_offset=6,
            end_lineno=1,
            end_col_offset=9),
          Name(
            id='d',
            ctx=Load(),
            lineno=1,
            col_offset=13,
            end_lineno=1,
            end_col_offset=14)],
        lineno=1,
        col_offset=0,
        end_lineno=1,
        end_col_offset=15),
      lineno=1,
      col_offset=0,
      end_lineno=1,
      end_col_offset=15)],
  type_ignores=[])
```

compiler/parser/python.lalrpop
harupy, line 1175: I see, what this block does is:

```python
{
  "a": "b",
  **c,
  "d": "e", <- treat this as dict.
}
```

harupy: If so, `compile_dict` needs more fixes

compiler/codegen/src/compile.rs
```diff
@@ -1986,18 +1986,24 @@ impl Compiler {
...
+            .partition(|(k, _)| k.is_some());
+        for (key, value) in packed {
+            self.compile_expression(key.as_ref().unwrap())?;
             self.compile_expression(value)?;
             size += 1;
         }
         emit!(self, Instruction::BuildMap { size });
 
-        for value in unpacked_values {
+        for (_, value) in unpacked {
```
harupy, line 2006: Could this be simplified to:

```rust
        emit!(self, Instruction::BuildMap { size: 0 });
        for (key, value) in keys.iter().zip(values.iter()) {
            if let Some(key) = key {
                self.compile_expression(key)?;
                self.compile_expression(value)?;
                emit!(self, Instruction::MapAdd { i: 0 });
            } else {
                self.compile_expression(value)?;
                emit!(self, Instruction::DictUpdate);
            }
        }
```

?

harupy:
```
> target/release/rustpython
Welcome to the magnificent Rust Python 0.2.0 interpreter 😱 🖖
>>>>> x = {"a": 0}
>>>>> x
{'a': 0}
>>>>> y = {"b": 1, **x}
>>>>> y
{'b': 1, 'a': 0}
```

It seems to work.

compiler/codegen/src/compile.rs
```diff
@@ -1986,22 +1986,22 @@ impl Compiler {
...
-
-        for value in unpacked_values {
-            self.compile_expression(value)?;
-            emit!(self, Instruction::DictUpdate);
+    fn compile_dict(
+        &mut self,
+        keys: &[Option<ast::Expr>],
+        values: &[ast::Expr],
+    ) -> CompileResult<()> {
+        emit!(self, Instruction::BuildMap { size: 0 });
```
youknowone, line 1994: This implementation looks like to make performance regression

harupy: Got it. I'll revert it.

harupy:
@youknowone @coolreader18 Would you mind taking a look?

compiler/ast/src/ast_gen.rs
```diff
@@ -195,7 +195,7 @@ pub enum ExprKind<U = ()> {
         orelse: Box<Expr<U>>,
     },
     Dict {
-        keys: Vec<Expr<U>>,
+        keys: Vec<Option<Expr<U>>>,
```
youknowone, line 198: I found I changed it to the previous form to keep cpython compatibility in #4042.
I'd like to ask `Option<Expr<U>>` is fit in cpython way.
I thought this is the new cpython way when i was working on #4042 which is changed in 3.10

harupy: @youknowone Why did you change it to the previous form?

youknowone: I can't remember well so looking in it again.
I thought CPython's Python.asdl actually changed it to not optional here.
https://github.com/RustPython/RustPython/pull/4042/files#diff-242a44b9d53b6c6212a752e0c97d6cf2f87794f579700aa8015bb99cef63925cR63

But I still can see ast.parse generate None for empty keys in Python 3.10. I am confused.
Do you know what the '?' means in `Python.asdl`?

harupy: I think `?` means optional.

compiler/ast/src/ast_gen.rs
```diff
@@ -195,7 +195,7 @@ pub enum ExprKind<U = ()> {
         orelse: Box<Expr<U>>,
     },
     Dict {
-        keys: Vec<Expr<U>>,
+        keys: Vec<Option<Box<Expr<U>>>>,
```
harupy, line 198:
```suggestion
        keys: Vec<Option<Expr<U>>>,
```

@youknowone Can we remove `Box` or not?

compiler/ast/src/ast_gen.rs
youknowone: Yes, that was the reason I thought CPython changed it from optional key to non-optional one in 3.10. How they still have None in parse.dump?

compiler/ast/Python.asdl
```diff
@@ -60,7 +60,7 @@ module Python
          | UnaryOp(unaryop op, expr operand)
          | Lambda(arguments args, expr body)
          | IfExp(expr test, expr body, expr orelse)
-         | Dict(expr* keys, expr* values)
+         | Dict(expr?* keys, expr* values)
```
youknowone, line 63: CPython doesn't have `?` in keys anymore.

https://github.com/python/cpython/blob/main/Parser/Python.asdl#L64

harupy: @youknowone True, but `dict.keys` does contain `None`.

harupy: https://github.com/python/cpython/blob/main/Parser/Python.asdl#L64 used to look like `| Dict(expr?* keys, expr* values)?`

harupy: It looks like this line hasn't changed for 11 years: https://github.com/python/cpython/blame/3325f054e33b318aa56b74472f76a56b8afc0510/Parser/Python.asdl#L64

harupy: https://github.com/python/cpython/blob/3325f054e33b318aa56b74472f76a56b8afc0510/Grammar/python.gram#L904

```
double_starred_kvpair[KeyValuePair*]:
    | '**' a=bitwise_or { _PyPegen_key_value_pair(p, NULL, a) }
                                                     ^^^^
    | kvpair
```

harupy: Posted https://github.com/python/cpython/issues/101211

youknowone: Oh, then `expr?* keys` must have been our local patch. I didn't know that. Thank you.

harupy: @youknowone Can we move forward with `expr?* keys`?

youknowone:
Sure, i am sorry for that.I will add a comment that part is our local patch not to break it again in future2023. 1. 21. 12:16, Harutaka Kawamura ***@***.***> 작성:﻿
@harupy commented on this pull request.



In compiler/ast/Python.asdl:
> @@ -60,7 +60,7 @@ module Python
          | UnaryOp(unaryop op, expr operand)
          | Lambda(arguments args, expr body)
          | IfExp(expr test, expr body, expr orelse)
-         | Dict(expr* keys, expr* values)
+         | Dict(expr?* keys, expr* values)

@youknowone Can we move forward with expr?* keys?

harupy:
> I will add a comment that part is our local patch not to break it again in future

Makes sense! Feel free to push a commit to this PR if necessary :)

compiler/ast/Python.asdl
DimitrisJim: I don't think `expr?*` is the right thing here. Not only is it _probably_ not valid ASDL syntax (see [here](https://github.com/eliben/asdl_parser/blob/master/docs/ASDL%20Zephyr%20-%20Wang.pdf), the modifiers are in brackets as opposed to braces, though this might not really be an issue), but it is also kinda ambiguous, is it an Optional list or a list of optionals?

Can't we get the same behavior by modifying `asdl_rs` and keeping `Python.asdl` the same as CPythons?

harupy: > Not only is it probably not valid ASDL syntax (see [here](https://github.com/eliben/asdl_parser/blob/master/docs/ASDL%20Zephyr%20-%20Wang.pdf), the modifiers are in brackets as opposed to braces, though this might not really be an issue), but it is also kinda ambiguous, is it an Optional list or a list of optional?

Agreed!

> Can't we get the same behavior by modifying asdl_rs and keeping Python.asdl the same as CPythons?

We can. I'm happy to revert the last commit.

@youknowone What do you think?

youknowone: If we could do, I prefer to keep Python.asdl as same as CPython.

harupy: Yes, updated the code.

harupy:
@youknowone @DimitrisJim Updated the code! Would you mind taking another look?

youknowone (approved):
Thank you! And I am really sorry for confusing review for this PR 😅

harupy:
Not a problem at all!

PR #4450, opened 2023-01-15
Refactor `FormatType`

yt2b:
I've fxied to hold `float_ops::Case` in the following `FormatType` value.
* Hex
* Exponent
* GeneralFormat
* FixedPoint


youknowone (approved):
Thank you!

Issue #4247, opened 2022-10-25
Update libraries to Python 3.11


youknowone:
3.11 is released today.

fanninpm:
@jopemachine from what I've seen, you seem to be interested in updating the code in the `Lib/` directory. Do you want to take care of this upgrade? If so, how much do you want to tackle?

jopemachine:
> @jopemachine from what I've seen, you seem to be interested in updating the code in the `Lib/` directory. Do you want to take care of this upgrade? If so, how much do you want to tackle?

Thanks for letting me know.
I'm interested in participating in RustPython's various issues including python the Lib update,
but unfortunately I'm not sure how much time I can spend on this for now because I have to process several personal works.

lastmjs:
I'm excited for the upgrade, our project needs this issue to be resolved which has been resolved in Python 3.11: https://github.com/python/cpython/issues/89026

lastmjs:
Sorry  I wasn't asking to be assigned, just saying I'm excited and that this is important to our project.

fanninpm:
@lastmjs I took a look at the PR you linked, and it seems easy enough to update the files mentioned in that PR. Would you like to give it a try and send us a PR?

lastmjs:
No thanks, I am happy waiting

youknowone:
replaced by #4564
I can't believe I made the same issue just a few weeks ago.

PR #4451, opened 2023-01-15
add Lib/ctypes from cpython 3.11

notJoon:
## Description
I newly added `Lib/ctypes` because it was missing. imported from CPython 3.11 branch.

## Relate
#4247


youknowone (approved):
Thank you!

PR #4453, opened 2023-01-15
Fix error message

yt2b:
I've fixed the error message of ValueError.
* `test_error_message` in `test_float.py`


youknowone (approved):
Thank you!

Issue #3692, opened 2022-05-08
__new__ doesn't raise TypeError


rebunto:
## Feature

`<built-in type>.__new__(<other built-in type>)` raise `TypeError`.
(Maybe 'built-in type' is not a exact definition because `object.__new__(_frozen_importlib.BuiltinImporter)` is valid in CPython.)

CPython:

```python
>>> a = object.__new__(dict)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: object.__new__(dict) is not safe, use dict.__new__()
```

RustPython:

```python
>>>>> a = object.__new__(dict) # This doen't raise TypeError
>>>>> repr(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: Expected payload 'dict' but 'dict' found
>>>>> str(a)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
RuntimeError: Expected payload 'dict' but 'dict' found
```
 
Most of cases are above,
but when `bool` is passed as argument, it is panicked.

```python
>>>>> a = object.__new__(bool)
>>>>> a
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', vm\src\builtins\bool.rs:188:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `C:\Users\user\RustPython\target\release\rustpython.exe` (exit code: 101)
```

## Python Documentation

https://github.com/python/cpython/blob/85354ed78c0edb6d81a2bd53cabc85e547b8b26e/Objects/typeobject.c#L7120

Relevant code in RustPython:
https://github.com/RustPython/RustPython/blob/996250d959134382d563f3ca5961cad02bfd43a0/vm/src/builtins/type.rs#L227

youknowone:
This is also related to #1979

PR #4455, opened 2023-01-15
Raise TypeError when calling __new__ unsafely

garychia:
Related to #3692
This implementation is based on the following code from CPython.
https://github.com/python/cpython/blob/206f05a46b426eb374f724f8e7cd42f2f9643bb8/Objects/typeobject.c#L7685-L7700
image

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -1120,6 +1120,24 @@ pub(crate) fn call_slot_new(
     args: FuncArgs,
     vm: &VirtualMachine,
 ) -> PyResult {
+    let mut staticbase = Some(subtype.clone());
+    while let Some(ref basetype) = staticbase {
+        if (basetype.flags() & PyTypeFlags::HEAPTYPE.bits()) != 0 {
+            staticbase = basetype.base().clone();
+        } else {
+            break;
+        }
+    }
+    if let Some(ref basetype) = staticbase {
+        if !PyType::subclasscheck(basetype.to_owned(), typ.to_owned()) {
+            return Err(vm.new_type_error(format!(
+                "{}.__new__({}) is not safe, use {}.__new__()",
+                typ.name(),
+                subtype.name(),
+                basetype.name()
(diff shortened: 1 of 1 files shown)


youknowone (changes requested):
CI Failure came from successful test.
```
test_subclass (test.test_bool.BoolTest) ... unexpected success
```
Because you fixed the bug, you can remove `@unittest.expectedFailure` decorator from the test.

Thank you!

garychia:
I found the following issue in test_ast.py.

image

But I'm sure if it is related to my code. It seems that `typ` in `call_slot_new` is set to `object` instead of `Constant` for some reason.

vm/src/builtins/type.rs
```diff
@@ -1120,6 +1120,24 @@ pub(crate) fn call_slot_new(
...
+    let mut staticbase = Some(subtype.clone());
+    while let Some(ref basetype) = staticbase {
+        if (basetype.flags() & PyTypeFlags::HEAPTYPE.bits()) != 0 {
+            staticbase = basetype.base().clone();
+        } else {
+            break;
+        }
+    }
+    if let Some(ref basetype) = staticbase {
+        if !PyType::subclasscheck(basetype.to_owned(), typ.to_owned()) {
```
youknowone, line 1132: `PyType_IsSubtype` in CPython is same to `fast_issubclass` in RustPython

DimitrisJim:
Hey @garychia do you still have time/want to work on this?

garychia:
Hi, @DimitrisJim! I was struggling to solve the problem that happens when running the `test_ast` test as I mentioned before. Based on my observation, RustPython tried to construct an `N` object by using `object.__new__`, which is not safe. But I have no idea why it behaved this way and whether it relates to my code.
This piece of code triggered the problem:
https://github.com/RustPython/RustPython/blob/bbd8fb23420ba13c007f180a77254d5ee966cb47/Lib/test/test_ast.py#L521-L528

PR #4460, opened 2023-01-19
Implement `strict_mode` keyword for `binascii.a2b_base64()`

evanrittenhouse:
New PR for #4374. Still working on figuring out how to get the error messages to match, but wanted to get this up to at least run against CI and see what else I'm missing. I think the general direction should be okay.

### DIFF
--- a/stdlib/src/binascii.rs
+++ b/stdlib/src/binascii.rs
@@ -203,7 +203,7 @@ mod decl {
                     if quad_pos >= 2 && quad_pos + pads >= 4 {
                         if strict_mode && i + 1 < b.len() {
                             // Represents excess data after padding error
-                            return Err(base64::DecodeError::InvalidLastSymbol(i, el));
+                            return Err(base64::DecodeError::InvalidLastSymbol(i, PAD));
                         }
 
                         return Ok(decoded);
@@ -246,18 +246,11 @@ mod decl {
                 }
             }
 
-            if quad_pos == 1 {
-                // Ensure that a PAD never gets passed, since that'd mistakenly cause an excess
-                // data after padding error
-                return Err(base64::DecodeError::InvalidLastSymbol(
-                    decoded.len() / 3 * 4 + 1,
-                    0,
-                ));
-            } else if quad_pos > 1 {
-                return Err(base64::DecodeError::InvalidLength);
-            }
-
-            Ok(decoded)
+            return match quad_pos {
+                0 => Ok(decoded),
+                1 => Err(base64::DecodeError::InvalidLastSymbol(decoded.len() / 3 * 4 + 1, 0)),
+                _ => Err(base64::DecodeError::InvalidLength)
+            };
         })
         .map_err(|err| {
             let python_error = match err {
@@ -270,10 +263,13 @@ mod decl {
                 base64::DecodeError::InvalidByte(_, _) => {
                     String::from("Only base64 data is allowed")
                 }
-                base64::DecodeError::InvalidLastSymbol(_, _) => {
+                base64::DecodeError::InvalidLastSymbol(_, PAD) => {
                     String::from("Excess data after padding")
                 }
-                base64::DecodeError::InvalidLength => String::from("Not implemented (yet)"),
+                base64::DecodeError::InvalidLastSymbol(length, _) => {
+                    format!("Invalid base64-encoded string: number of data characters {} cannot be 1 more than a multiple of 4", length)
+                }
+                base64::DecodeError::InvalidLength => String::from("Incorrect padding"),
             };
 
             new_binascii_error(format!("error decoding base64: {python_error}"), vm)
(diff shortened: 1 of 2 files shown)


stdlib/src/binascii.rs
```diff
@@ -171,25 +179,125 @@ mod decl {
...
-            let decoded = if b.len() % 4 == 0 {
-                decode(b)
+            println!(
+                "{} {:?}: {:?}",
+                strict_mode,
+                b,
+                is_valid(&b.to_vec(), strict_mode)
+            );
+
+            let bytes_are_valid = is_valid(&b.to_vec(), strict_mode);
```
DimitrisJim, line 254: A small amount of quick feedback here, this will iterate over every char and then we `iter()` through it again. You can break it apart and `for`-loop through the bytes and move everything from `is_valid` in the for loop body.

evanrittenhouse: Oh hm. Are you to saying break apart the `b` vector and loop through that? I was thinking about doing that, but didn't really understand the point of filtering out the `-1` bytes only after we've decided that `decoded` was an error.

evanrittenhouse: Yeah totally! It may take me a few more days before I can work on it, work has been crazy. Appreciate your patience.

evanrittenhouse: @DimitrisJim at the risk of reinventing the wheel, could we just filter out all `-1` bits (similar to how we do now) in `is_valid` if strict_mode is true, then pass that filtered vector to `base64::decode`? Happy to go byte by byte, was just looking at `base64::decode`'s implementation last night and there are a lot of optimizations for speed there

DimitrisJim: Totally, whatever you're most comfortable with at the end of the day. I mentioned the CPython implementation since its usually easier working against a reference implementation.

evanrittenhouse: Ended up copying the CPython implementation. Thanks for the tip!

evanrittenhouse:
@DimitrisJim was wondering if you could help me out. `cargo run -- Lib/test/test_binascii.py` passes locally, but it seems like some of the `binascii` tests are failing here. Is there a difference? I'm also unsure why `test_ioctl`, `test_urllib`, etc. would be failing if I haven't done any work on them. 

Thanks for all your help so far!

fanninpm:

```shell
```

DimitrisJim:
Seems like `test_urllib` has a [test](https://github.com/RustPython/RustPython/blob/e5735cde67b86699bd68c8959167ecf2649a6f2d/Lib/test/test_urllib.py#L682) that indirectly uses the changes. (`test_ioctl` seems to just be skipped after glancing through the ci logs so you can ignore that).

As @fanninpm mentioned, you should use `-- -m test` to run these (running with `cargo run -- Lib/test/test_binascii.py` doesn't actually invoke the test runner, it just runs it as a normal Python file).

I'll try and make some time to check through the code to see if I can spot where any issues might be and thanks for working on it!

evanrittenhouse:
Looks like `test_urllib` is failing because we're not checking for `Incorrect padding`. It's marked as an expected failure in `test_binascii.py`, so that's what I'm working on now

evanrittenhouse:
@DimitrisJim I think (fingers crossed) this is done?

stdlib/src/binascii.rs
```diff
@@ -171,25 +177,107 @@ mod decl {
...
+                // data after padding error
+                return Err(base64::DecodeError::InvalidLastSymbol(decoded.len() / 3 * 4 + 1, 0));
+            } else if quad_pos > 1 {
+                return Err(base64::DecodeError::InvalidLength)
+            }
+
+            Ok(decoded)
+        })
+        .map_err(|err| {
+            let python_error = match err {
```
youknowone, line 260: `impl ToPyException for base64::DecodeError` is possible to replace this block

evanrittenhouse: Sorry, pretty new to Rust. Could you please provide an example? Not super sure what you mean :grimacing:

DimitrisJim: `io.rs` in `vm/src/stdlib` contains an implementation of `ToPyException` for `Io::Error`which can definitely serve as a good example. You basically just move the `match` in there and use `to_pyexception` here.

youknowone: We have a trait called `ToPyException` which used to convert any error to python exception. It is used to convert `Io::Error` to python exception as @DimitrisJim explained.
Once you implement it, this code will look like:

```rust
        .map_err(|err| err.to_pyexception(vm))
```

If you stuck with it, please feel free to ask anything again or tell us to take over that part.

The patch is really nice and this suggestion is totally optional. You don't need to finish to implement ToPyException to finish this PR.

Thank you!

evanrittenhouse: Thanks for the kind words! I'm really sorry - it looks like work is getting pretty busy, so I'm not sure that I'll be able to implement this. Possible for you to take it on? @youknowone

youknowone (approved):
Thank you for contributing! I appended a commit with a little of changes

PR #4461, opened 2023-01-19
Add basic number formatting

notJoon:
## Description
Only basic functionalities have been implemented so that formatting can be applied and output even when a 'float' type number comes in.

However, not all functions are fully functional(i.e, decimal rounding or `inf` value handling.), so it seems that additional functions will need to be supplemented in the future.

## Example
```python
## before
print('{:5n}'.format(1234567.1))
>>> Traceback (most recent call last):
    File "<wasm>", line 1, in <module>
    ValueError: Format code 'n' for object of type 'float' not implemented yet

## after
>>> 1.23457e+06
```
```python
a = '{:.8n}'.format(12345.12345)

type(a)
>>> <class 'str'>

print(a)
>>> 12345.123
```

## Related Issue
#1656

### DIFF
--- a/common/src/format.rs
+++ b/common/src/format.rs
@@ -3,7 +3,10 @@ use float_ops::Case;
 use itertools::{Itertools, PeekingNext};
 use num_bigint::{BigInt, Sign};
 use num_traits::{cast::ToPrimitive, Signed};
-use std::{cmp, str::FromStr};
+use std::{
+    cmp::{self, Ordering},
+    str::FromStr,
+};
 
 trait FormatParse {
     fn parse(text: &str) -> (Option<Self>, &str)
@@ -116,6 +119,19 @@ impl FormatParse for FormatGrouping {
             Some(',') => (Some(Self::Comma), chars.as_str()),
             _ => (None, text),
         }
+    }
+}
+
+#[derive(Debug)]
+#[repr(u8)]
+enum PaddingStyle {
+    Whitespace = b' ',
+    Zero = b'0',
+}
+
+impl From<PaddingStyle> for char {
+    fn from(from: PaddingStyle) -> char {
+        from as u8 as char
     }
 }
 
@@ -312,13 +328,15 @@ impl FormatSpec {
         inter: i32,
         sep: char,
         disp_digit_cnt: i32,
+        padding: PaddingStyle,
     ) -> String {
         // Don't add separators to the floating decimal point of numbers
         let mut parts = magnitude_str.splitn(2, '.');
         let magnitude_int_str = parts.next().unwrap().to_string();
         let dec_digit_cnt = magnitude_str.len() as i32 - magnitude_int_str.len() as i32;
         let int_digit_cnt = disp_digit_cnt - dec_digit_cnt;
-        let mut result = FormatSpec::separate_integer(magnitude_int_str, inter, sep, int_digit_cnt);
+        let mut result =
+            FormatSpec::separate_integer(magnitude_int_str, inter, sep, int_digit_cnt, padding);
         if let Some(part) = parts.next() {
             result.push_str(&format!(".{part}"))
         }
@@ -330,6 +348,7 @@ impl FormatSpec {
         inter: i32,
         sep: char,
         disp_digit_cnt: i32,
+        padding: PaddingStyle,
     ) -> String {
         let magnitude_len = magnitude_str.len() as i32;
         let offset = (disp_digit_cnt % (inter + 1) == 0) as i32;
@@ -338,7 +357,8 @@ impl FormatSpec {
         if pad_cnt > 0 {
             // separate with 0 padding
             let sep_cnt = disp_digit_cnt / (inter + 1);
-            let padding = "0".repeat((pad_cnt - sep_cnt) as usize);
+            let padding_char = padding.into();
+            let padding = FormatSpec::compute_fill_string(padding_char, pad_cnt - sep_cnt);
             let padded_num = format!("{padding}{magnitude_str}");
             FormatSpec::insert_separator(padded_num, inter, sep, sep_cnt)
         } else {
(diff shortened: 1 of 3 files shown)


common/src/format.rs
```diff
@@ -338,7 +346,10 @@ impl FormatSpec {
         if pad_cnt > 0 {
             // separate with 0 padding
             let sep_cnt = disp_digit_cnt / (inter + 1);
-            let padding = "0".repeat((pad_cnt - sep_cnt) as usize);
+            let padding = match use_whitespace_padding {
+                true => " ".repeat((pad_cnt - sep_cnt) as usize),
+                false => "0".repeat((pad_cnt - sep_cnt) as usize),
+            };
```
fanninpm, line 352: `match`ing on a bool is less readable than using `if`-statements

```suggestion
            let padding = if use_whitespace_padding {
                " ".repeat((pad_cnt - sep_cnt) as usize)
            } else {
                "0".repeat((pad_cnt - sep_cnt) as usize)
            };
```

fanninpm (commented):
A few nitpicks regarding the usage of `match` expressions vs. `if` expressions. If you decide to apply these suggestions, please remember to run them through `rustfmt`.

common/src/format.rs
youknowone:
```suggestion
            let padding_char = if use_whitespace_padding {
                ' '
            } else {
                '0'
            };
            let padding = padding_char.repeat((pad_cnt - sep_cnt) as usize);
```

notJoon:
@fanninpm Thanks for your advice! It's definitely better than before 👍👍

common/src/format.rs
notJoon: > 

I found a similar code implemented before(`FormatSpec::compute_fill_string`) and changed it to that instead. Thank you.

common/src/format.rs
```diff
@@ -338,7 +349,8 @@ impl FormatSpec {
         if pad_cnt > 0 {
             // separate with 0 padding
             let sep_cnt = disp_digit_cnt / (inter + 1);
-            let padding = "0".repeat((pad_cnt - sep_cnt) as usize);
+            let padding_char = if use_whitespace_padding { ' ' } else { '0' };
+            let padding = FormatSpec::compute_fill_string(padding_char, pad_cnt - sep_cnt);
```
DimitrisJim, line 353: `use_whitespace_padding` could alternatively be an enum that `into`'s a `char`. A nice side-effect is that code would definitely read better (seeing `false` being supplied as an arg is mysterious as opposed to e.g `Padding::Zero` or `Padding::WhiteSpace`).

youknowone:
The newly added test seems fail in CPython 3.10

```
  File "/home/runner/work/RustPython/RustPython/extra_tests/snippets/builtin_str.py", line 619, in <module>
    assert '{:5n}'.format(1e+3) == " 1000"
AssertionError
```

common/src/format.rs
notJoon:
```rust
#[derive(Debug)]
#[repr(u8)]
enum PaddingStyle {
    Null = b'\0',
    Whitespace = b' ',
}

impl Into<char> for PaddingStyle {
    fn into(self) -> char {
        self as u8 as char
    }
}
```
hard-coded padding has been extracted with `PaddingStyle` enum type. thanks for your advice :)

notJoon: Based on your review, extracted the hard-coded padding as a 'PaddingStyle' enum type
```rust
#[derive(Debug)]
#[repr(u8)]
enum PaddingStyle {
    Whitespace = b' ',
    Zero = b'0',
}

impl From<PaddingStyle> for char {
    fn from(from: PaddingStyle) -> char {
        from as u8 as char
    }
}
```

youknowone (changes requested):
`test.test_format.FormatTest` is now unexpected success because you correctly fixed it!
We now can remove `@unittest.expectedFailure` from the test.

common/src/format.rs
DimitrisJim: ah, I should clarify a bit more now that I see your change. The idea behind the enum is to also change the function signatures to accept it, i.e `use_whitespace_padding` should just just be a `PaddingStyle` argument, so, for example:

```rust
fn separate_integer(
    magnitude_str: String,
    inter: i32,
    sep: char,
    disp_digit_cnt: i32,
    use_whitespace_padding: bool,
) -> String {
```

becomes:

```rust
fn separate_integer(
    magnitude_str: String,
    inter: i32,
    sep: char,
    disp_digit_cnt: i32,
    padding: PaddingStyle,
) -> String {
```

and call sites are changed from:

```rust
self.add_magnitude_separators(raw_magnitude_str?, sign_str, false);
```

into:

```rust
self.add_magnitude_separators(raw_magnitude_str?, sign_str, PaddingStyle::Zero);
```

In addition to these, the `if` statement that creates the `padding_char` can just be:

```rust
let padding_char = padding.into();
```

notJoon: ah I got it. I wasn't familiar with using Enum yet, so I couldn't handle it neatly. thanks for the feedback

notJoon:
> 

I don't know if this is right way, but I removed `@unittest.expectedFailure` from the `test.test_format.FormatTest`.

DimitrisJim:
@notJoon it's the right way! You should also remove the `# TODO` comments before the decorators though since they aren't needed anymore

common/src/format.rs
```diff
@@ -433,7 +460,42 @@ impl FormatSpec {
...
+                    }
+                    self.check_number_formatting_cases(magnitude, width, precision, &mut inter)
+                };
+
+                Ok(FormatSpec::add_magnitude_separators_for_char(
+                    magnitude_str,
+                    inter as i32,
+                    ZERO_SEP,
+                    disp_digit_cnt as i32,
+                    PaddingStyle::Zero,
```
DimitrisJim, line 496: Here's where you should use `PaddingStyle::WhiteSpace`.

notJoon: Oh why didn't I see this?? edited again

common/src/format.rs
```diff
@@ -384,14 +404,20 @@ impl FormatSpec {
 
     fn get_separator_interval(&self) -> usize {
         match self.format_type {
+            Some(FormatType::Number) => 9999,
```
youknowone, line 407: Is this meaning something or just intended a big number? I think bigint can have more digits than 9999.

youknowone:
I thought this is done except for test decorators but seems still on working, right?

common/src/format.rs
```diff
@@ -312,13 +334,15 @@ impl FormatSpec {
         inter: i32,
         sep: char,
         disp_digit_cnt: i32,
+        padding: PaddingStyle,
```
youknowone, line 337: I think we don't need this `padding`. It only allows whitespace or `0`, but whitespace padding doesn't add seprator.

Issue #4316, opened 2022-12-09
Extra argument in compile() leads to panicked.


xiaxinmeng:
The following test assign the compile() four arguments, then rustpython thread 'main' get panicked.

test.py:
```
import sys
PyCF_ONLY_AST = 1024
obj = compile('1', '', 'exec', PyCF_ONLY_AST)
```

Behavior on rustpython(rustpython 0.1.2):
```
thread 'main' panicked at 'static type has not been initialized', vm/src/class.rs:22:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Behavior on CPython 3.9.1(Expected):
work well

Step to reproduce:
type "cargo run --release test.py" on the console of Ubuntu 18.04 or MacOS Big Sur 11.7.1
rust version: rustc 1.65.0

DimitrisJim:
Nice find. `ast` is not initialized here (i.e, if you `import ast` before running this, it works fine). Seems we also need to initialize `ast` earlier (it exists in `sys.modules` for cpython but not for rustpython).

PR #4462, opened 2023-01-19
Initialize ast when calling compile function

garychia:
Related to #4316
This pull request allows the `compile` function to initialize the `ast` module before the compilation.
I am not sure if this is the correct way to initialize the module. But it seems to solve the problem.

### DIFF
--- a/vm/src/vm/mod.rs
+++ b/vm/src/vm/mod.rs
@@ -242,6 +242,7 @@ impl VirtualMachine {
         let mut essential_init = || -> PyResult {
             #[cfg(not(target_arch = "wasm32"))]
             import::import_builtin(self, "_signal")?;
+            import::import_builtin(self, "_ast")?;
             #[cfg(not(feature = "threading"))]
             import::import_frozen(self, "_thread")?;
             let importlib = import::init_importlib_base(self)?;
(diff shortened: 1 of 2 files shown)


youknowone (approved):
The change looks good. I cannot think of any better way to do it for now.

Your email address of this commit looks like not the same one you used before. Could you check if the email is correct and registered in your github account?

DimitrisJim:
I haven't checked where CPython does it but, wouldn't it make sense to initialize `ast` earlier during interpreter start-up? Do you remember how costly `make_module` is to call @youknowone ? (because we'll call it every time we call `compile`)

garychia:
Thanks for telling me. The email address is correct but I forgot to add it to my GitHub account. There should be no problem now.
Is there a way to initialize the module only if it is not initialized?

youknowone:
@DimitrisJim  I thought `make_module` has a duplicated running check, but it didn't.

@garychia The most high-level one is `vm.import`, which corresponds to python code `import ast`.

garychia:
Thanks for the advice from both of you. I changed my solution based on what @DimitrisJim suggested.

vm/src/vm/mod.rs
```diff
@@ -242,6 +242,7 @@ impl VirtualMachine {
         let mut essential_init = || -> PyResult {
             #[cfg(not(target_arch = "wasm32"))]
             import::import_builtin(self, "_signal")?;
+            import::import_builtin(self, "_ast")?;
```
youknowone, line 245:
```suggestion
            #[cfg(any(feature = "parser", feature = "compiler"))]
            import::import_builtin(self, "_ast")?;
```
if it is only required by compile function, feature checking may be helpful.

youknowone (approved):
Thank you!

PR #4463, opened 2023-01-20
Implement lone-dot semantics for %-style format strings

charliermarsh:
This PR extends `cformat.rs` to handle lone-dot cases, like `"%.f" % 123.456`. Previously, RustPython used six decimal places for this case, treating it equivalently to `"%f" % 123.456`. This differs from CPython, which treats `"%.f"` as "remove the decimal point". Similarly, CPython treats `"%.s"` as a full string truncation.

### DIFF
--- a/common/src/cformat.rs
+++ b/common/src/cformat.rs
@@ -34,7 +34,7 @@ impl fmt::Display for CFormatError {
         use CFormatErrorType::*;
         match self.typ {
             UnmatchedKeyParentheses => write!(f, "incomplete format key"),
-            CFormatErrorType::IncompleteFormat => write!(f, "incomplete format"),
+            IncompleteFormat => write!(f, "incomplete format"),
             UnsupportedFormatChar(c) => write!(
                 f,
                 "unsupported format character '{}' ({:#x}) at index {}",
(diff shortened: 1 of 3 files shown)


common/src/cformat.rs
```diff
@@ -203,19 +202,47 @@ impl CFormatSpec {
         }
     }
 
+    fn fill_string_with_precision(&self, string: String, fill_char: char) -> String {
```
charliermarsh, line 205: I split this into a separate function. Previously, it was coupled to `fill_string` with a boolean flag. It's a bit subjective, but I find it easier to reason about them as two separate functions.

charliermarsh:
(Taking a look at failed tests, etc.)

fanninpm:
Two tests in `Lib/test/test_locale.py` are now passing.

charliermarsh:
Fixed, thx.

youknowone (approved):

charliermarsh:
@youknowone - No rush, but anything else I can do to help here?

youknowone:
It looks good. I'd just like to edit style a little bit. Thank you!

charliermarsh:
No prob, thanks friend!

PR #4464, opened 2023-01-20
Bump phf to 0.11 series

akx:
string_cache is still using phf_shared 0.10.0 though.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1567,30 +1567,30 @@ dependencies = [
 
 [[package]]
 name = "phf"
-version = "0.10.1"
+version = "0.11.1"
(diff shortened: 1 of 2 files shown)


DimitrisJim:
Last time I remember seeing an error like this, it was caused by an update to one of the runners github uses. This time though, it seems like only this PR causes the failure since others seem to go through without a hitch. Not sure what's causing it here.

akx:
@DimitrisJim Hm, strange! I gave this a rebase, so the workflow needs to be reapproved...

DimitrisJim (approved):
scratch my previous comment, failure is some openssl headache, looks fine to me

PR #4465, opened 2023-01-22
Fix float mod

yt2b:
I've fixed float modulo operation.
* `test_float_mod` in `test_float.py`


youknowone (approved):
Thank you!

PR #4467, opened 2023-01-22
temporary skip rust tests for macOS to avoid CI failure

youknowone:


DimitrisJim:
maybe we should start thinking about using `rustls` instead of trying to link against openssl on the given system.

youknowone:
I cannot find the reason. It perfectly works on my machine. But my machine is macos-13 and CI is macos-12. Let's try it later on macos-13 image once ready.

DimitrisJim (approved):

DimitrisJim:
According to https://github.com/actions/runner-images/issues/6426#issuecomment-1341047365 the next version should be available in a couple of months.

PR #4469, opened 2023-01-23
Register `array.array` to `collections.abc.MutableSequence`

moreal:
This pull request registers `array.array` to `collections.abc.MutableSequence._abc_registry` and it makes `isinstance(array.array("<type>"), collections.abc.MutableSequence)` as `true`.

There is an issue that it cannot load `MutableSequence` from `collections.abc`, so I bypassed it by loading `MutableSequence` from `_collections_abc` directly. But it should be fixed or note comments about specific reason.

You can run the below code to test these changes:

```python
import array, collections.abc; print(isinstance(array.array('B'), collections.abc.MutableSequence))
```

### Related links

 - https://github.com/python/cpython/blob/bd7903967cd2a19ebc842dd1cce75f60a18aef02/Modules/arraymodule.c#L3075-L3088

### DIFF
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -1,4 +1,31 @@
-pub(crate) use array::make_module;
+use rustpython_vm::{PyObjectRef, VirtualMachine};
+
+pub(crate) fn make_module(vm: &VirtualMachine) -> PyObjectRef {
+    let module = array::make_module(vm);
+
+    let array = module
+        .get_attr("array", vm)
+        .expect("Expect array has arrat type.");
+
+    // TODO: RUSTPYTHON
+    // FIXME: it should import 'collections.abc' instead of '_collection_abc'
+    let collections_abc = vm
+        .import("_collections_abc", None, 0)
+        .expect("Expect collections.abc exist.");
+    let mutable_sequence = collections_abc
+        .get_attr("MutableSequence", vm)
+        .expect("Expect collections.abc has MutableSequence type.");
+
+    vm.invoke(
+        &mutable_sequence
+            .get_attr("register", vm)
+            .expect("Expect collections.abc.MutableSequence has register method."),
+        (array,),
+    )
+    .expect("Expect collections.abc.MutableSequence.register(array.array) not fail.");
+
+    module
+}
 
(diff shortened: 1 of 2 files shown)


stdlib/src/array.rs
```diff
@@ -1,4 +1,31 @@
-pub(crate) use array::make_module;
+use rustpython_vm::{PyObjectRef, VirtualMachine};
+
+pub(crate) fn make_module(vm: &VirtualMachine) -> PyObjectRef {
+    let module = array::make_module(vm);
+
+    let array = module
+        .get_attr("array", vm)
+        .expect("Expect array has arrat type.");
```
DimitrisJim, line 8:
```suggestion
        .expect("Expect array has array type.");
```

stdlib/src/array.rs
```diff
@@ -1,4 +1,31 @@
...
+
+    // TODO: RUSTPYTHON
+    // FIXME: it should import 'collections.abc' instead of '_collection_abc'
+    let collections_abc = vm
+        .import("_collections_abc", None, 0)
+        .expect("Expect collections.abc exist.");
+    let mutable_sequence = collections_abc
+        .get_attr("MutableSequence", vm)
+        .expect("Expect collections.abc has MutableSequence type.");
```
DimitrisJim, line 17: the equivalent for us would probably be `getattr` on `collections_abc` to get the `abc` submodule, our `import` doesn't exactly work like `PyImport_ImportModule` (which, as [its docs](https://docs.python.org/3/c-api/import.html#c.PyImport_ImportModule) state, returns the named module after the `.`), so, something like:

```suggestion
    let collections_abc = vm
        .import("collections.abc", None, 0)
        .expect("Expect collections exist.");
    let abc = collections_abc.get_attr("abc", vm).expect("Expect collections.abc has abc submodule.");
    let mutable_sequence = abc
        .get_attr("MutableSequence", vm)
        .expect("Expect collections.abc has MutableSequence type.");
```

would probably be closest to `PyImport_ImportModule("collections.abc")`.

DimitrisJim (approved):
lgtm, thanks moreal!

PR #4474, opened 2023-01-27
Fix #4157: handle error in case of unsupported format

### LINKED ISSUE #4157: `time.strftime(arg)` fails when arg is not valid format string
## Expected Result

In cpython, when arg is not a valid format string, `strftime` returns arg itself.

```
>>> import time
>>> time.strftime("%Y")
'2022'
>>> time.strftime("%4Y")
'4Y'
```

## Actual Result

~In RustPython, `strftime` raise panic when arg is not a valid format string~

```
>>>>> import time
>>>>> time.strftime("%Y")
'2022'
>>>>> time.strftime("%4Y")
thread 'main' panicked at 'a Display implementation returned an error unexpectedly: Error', /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/alloc/src/string.rs:2490:14
```

itsankitkp:
Chrono panics in case of unsupported formats, this patch handles such cases and returns supplied format as a result.


Lib/test/test_time.py
```diff
@@ -171,7 +170,12 @@ def test_strftime(self):
 
         self.assertRaises(TypeError, time.strftime, b'%S', tt)
         # embedded null character
-        self.assertRaises(ValueError, time.strftime, '%S\0', tt)
+        # XXX this test case fails, this should have been handled in
+        # in in layers above fn strftime.
+        # self.assertRaises(ValueError, time.strftime, '%S\0', tt)
+
+        # check if unsupported arg in strftime returns arg itself
+        self.assertEqual(time.strftime("%4Y"), "%4Y")
```
fanninpm, line 178: Is this in CPython?

fanninpm (changes requested):

Lib/test/test_time.py
itsankitkp: No, CPython returns '2023'. But rust chrono doesn't support it yet

fanninpm: We prefer to keep the tests in the `Lib/test` directory as close as possible to what's in CPython.

itsankitkp: Okay, should I keep it in extra_tests/snippets/stdlib_datetime.py ?

fanninpm: If the `"%4Y"` case no longer panics, then you could change the decorator to this:

```python
# TODO: RUSTPYTHON
@unittest.expectedFailure
def test_function():
    # ...
```

vm/src/stdlib/time.rs
```diff
@@ -207,8 +207,22 @@ mod time {
...
+        /*
+         * chrono doesn't support all formats and it
+         * raised an error if unsupported format is supplied.
+         * If error happens, we set result as input arg
+         */
+        let result = write!(&mut formatted_time, "{}", instant.format(format.as_str()));
+        match result {
+            Ok(r) => r,
+            Err(_) => formatted_time = format.to_string(),
+        };
```
fanninpm, line 225: Does this work?

```suggestion
        let formatted_time = String::try_from(instant.format(format.as_str())).unwrap_or(|| format.to_string());
```

fanninpm:
It's fine to have that `"%4Y"` case in the `extra_tests/`.

Lib/test/test_time.py
itsankitkp: > If the `"%4Y"` case no longer panics, then you could change the decorator to this:
> 
> ```python
> # TODO: RUSTPYTHON
> @unittest.expectedFailure
> def test_function():
>     # ...
> ```

I have done this, test ran fine.

vm/src/stdlib/time.rs
itsankitkp: I am getting following error

`the trait `std::convert::From<DelayedFormat<StrftimeItems<'_>>>` is not implemented for `std::string::String``

itsankitkp: But I do see what you are trying to do here. I will give another pass at this.

itsankitkp:
> It's fine to have that `"%4Y"` case in the `extra_tests/`.

Okay, I tried that but there is an issue
If I write assert_equal(_time.strftime("%4Y"), "%4Y"),
it runs first arg against cpython (which returns "2023") and 2nd part against rustpython (which returns "%4Y")
What I require is to run both first and second argument against rustpython only. Am I looking at wrong direction here?

> Assertion Failure: <class 'str'>(2023) == <class 'str'>(%4Y)

vm/src/stdlib/time.rs
itsankitkp: I have refactored code, it looks lot better now

itsankitkp:
> > It's fine to have that `"%4Y"` case in the `extra_tests/`.
> 
> Okay, I tried that but there is an issue If I write assert_equal(_time.strftime("%4Y"), "%4Y"), it runs first arg against cpython (which returns "2023") and 2nd part against rustpython (which returns "%4Y") What I require is to run both first and second argument against rustpython only. Am I looking at wrong direction here?
> 
> > Assertion Failure: <class 'str'>(2023) == <class 'str'>(%4Y)

I have fixed this for now by using unsupported format which returns same result in both python and rustpython.

youknowone:
@itsankitkp it passes tests on ubuntu, but not on macos and windows. could you check it?

```
----------------------------- Captured stdout call -----------------------------
Assertion Failure: <class 'str'>(?) == <class 'str'>(%?)
----------------------------- Captured stderr call -----------------------------
  File "/Users/runner/work/RustPython/RustPython/extra_tests/snippets/stdlib_datetime.py", line 135, in <module>
    assert_equal(_time.strftime("%?"), "%?")
  File "/Users/runner/work/RustPython/RustPython/extra_tests/snippets/testutils.py", line 58, in assert_equal
    _assert_print(lambda: a == b, [_typed(a), '==', _typed(b)])
  File "/Users/runner/work/RustPython/RustPython/extra_tests/snippets/testutils.py", line 47, in _assert_print
    assert f()
AssertionError
```

youknowone:
rebase seems messed up. could you try:
```
git fetch upstream
git rebase upstream/main
```
when upstream is `github.com/RustPython/RustPython`

PR #4475, opened 2023-01-28
feat: allow not to set SIGINT handler

discord9:
# Allow not to set SIGINT handler
which is similiar to pyo3's `prepare_freethreaded_python()`, will not raise the `KeyboardInterrupt` exception, this will be benefit when embedding RustPython in other application(Which might want to set their own Interrupt)

### DIFF
--- a/vm/src/stdlib/signal.rs
+++ b/vm/src/stdlib/signal.rs
@@ -103,7 +103,9 @@ pub(crate) mod _signal {
             .clone()
             .get_attr("default_int_handler", vm)
             .expect("_signal does not have this attr?");
-        signal(libc::SIGINT, int_handler, vm).expect("Failed to set sigint handler");
+        if !vm.state.settings.no_sig_int {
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/signal.rs
```diff
@@ -103,7 +103,9 @@ pub(crate) mod _signal {
             .clone()
             .get_attr("default_int_handler", vm)
             .expect("_signal does not have this attr?");
-        signal(libc::SIGINT, int_handler, vm).expect("Failed to set sigint handler");
+        if !vm.state.settings.no_sig_int {
```
DimitrisJim, line 106: think this should be an implementation specific option, i.e, passed with `-X`.

discord9: > think this should be an implementation specific option, i.e, passed with `-X`.

Do you mean like a clap arg `-X` and add codes in `setting_from` function?

DimitrisJim (approved):
thanks, this option seems sensible to me

Issue #4473, opened 2023-01-27
Qualname should be used in function's TypeErrors


nicku12345:
## Feature

Running the two code snippets below in CPython and in RustPython give different results

**Example 1**
```python3
def f(a, /, b=0): pass
f.__qualname__ = "g"
f(a=1, b=0)
```

In CPython:
```
TypeError: g() got some positional-only arguments passed as keyword arguments: 'a'
```

In RustPython:
```
TypeError: f() got some positional-only arguments passed as keyword arguments: 'a'
```

**Example 2**
```python3
class A:
    def f(self, a, /, b=0): pass

A().f(a=1, b=0)
```

In CPython:
```
TypeError: A.f() got some positional-only arguments passed as keyword arguments: 'a'
```
In RustPython:
```
TypeError: f() got some positional-only arguments passed as keyword arguments: 'a'
```

youknowone:
Thank you!

PR #4476, opened 2023-01-28
use qualname in `TypeErrors` for functions

nicku12345:
fix https://github.com/RustPython/RustPython/issues/4473


DimitrisJim:
It seems like clippy is complaining about a couple of things, if you could address those.

fanninpm:
> It seems like clippy is complaining about a couple of things, if you could address those.

Those might be due to Rust 1.67.0 coming out on 26 Jan.

youknowone (approved):
Thank you!

PR #4480, opened 2023-01-31
Minor lexer optimizations

DimitrisJim:
Mainly by matching on `'a'..='z'` and `'A'..='Z'` which are the most common identifier characters (avoiding the calls to `is_xid_start` and `is_xid_continue`). Apart from that makes `IndentationLevel` slightly smaller and adds some sensible capacities.

Testing locally, this results in tokenizing the stdlib around ~20% faster (mainly, the `match` change yields that), I'll probably add that benchmark and one for the parser since its nice to have lying around).

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -534,12 +534,15 @@ where
     }
 
     fn is_identifier_start(&self, c: char) -> bool {
-        c == '_' || is_xid_start(c)
+        match c {
+            'a'..='z' | 'A'..='Z' | '_' => true,
+            _ => is_xid_start(c),
+        }
     }
 
     fn is_identifier_continuation(&self) -> bool {
         match self.window[0] {
-            Some('_' | '0'..='9') => true,
+            Some('a'..='z' | 'A'..='Z' | '_' | '0'..='9') => true,
             Some(c) => is_xid_continue(c),
             _ => false,
         }
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -690,12 +690,12 @@ where
     fn consume_normal(&mut self) -> Result<(), LexicalError> {
         // Check if we have some character:
         if let Some(c) = self.window[0] {
-               if self.is_identifier_start(c) {
-                    let identifier = self.lex_identifier()?;
-                    self.emit(identifier);
-               } else {
-                    self.consume_character(c)?;
-               }
+            if self.is_identifier_start(c) {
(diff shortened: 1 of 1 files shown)


compiler/parser/src/lexer.rs
```diff
@@ -18,8 +18,8 @@ use unic_ucd_ident::{is_xid_continue, is_xid_start};
 
 #[derive(Clone, Copy, PartialEq, Debug, Default)]
 struct IndentationLevel {
-    tabs: usize,
-    spaces: usize,
+    tabs: u32,
+    spaces: u32,
```
fanninpm, line 22: I think there should be some sort of achievement for attempting to use 4,294,967,296 tabs or spaces for indentation, resulting in a script that's over 4 GB in size.

compiler/parser/src/lexer.rs
```diff
@@ -1124,7 +1125,7 @@ where
 
     fn eat_single_char(&mut self, ty: Tok) {
         let tok_start = self.get_pos();
-        self.next_char().unwrap();
```
youknowone, line 1127: is this safe to skip?

youknowone (approved):
nice idea!

compiler/parser/src/lexer.rs
DimitrisJim: just removed the `unwrap` call here since it wasn't needed

youknowone: I think this `unwrap` guarantees it never fails. Changing to this one will be zero cost when optimized but check if it is not an error when debugging.

```rust
self.next_char().unwrap_or_else(|| std::hint::unreachable_unchecked());
```

DimitrisJim: this should always be a `Some` since it gets called after a `Some(char)` has been found in `comsume_normal`, I'll add the suggestion though since it definitely makes the intent much clearer.

charliermarsh:
> Testing locally, this results in tokenizing the stdlib around ~20% faster

Rock on, that's awesome!

charliermarsh:
FWIW this made a noticeable difference in Ruff.

Before:

Screen Shot 2023-02-03 at 2 10 32 PM

After:

Screen Shot 2023-02-03 at 2 08 06 PM

DimitrisJim:
That's so nice to see (also a good second bench to check things on!). I've got a couple more ideas so hopefully we can cut down the time a bit more.

charliermarsh:
Awesome :) Always happy to benchmark on Ruff if helpful. We greatly appreciate these improvements!

Issue #4468, opened 2023-01-22
Store byte indices on tokens and nodes


charliermarsh:
In Ruff, we often need to go from `Location` to byte index (e.g., if we know a statement starts on row 31, column 4, we need a way to map that to a byte index in the source code, so that we can do a slice in the appropriate spot).

Today, we solve this problem by [indexing the source code in advance](https://github.com/charliermarsh/ruff/blob/a7ce8621a9c43bae200bdea60aa6803ef497ede1/src/source_code/locator.rs#L1) -- so, we iterate over all characters, and create a map from (row, column) to byte index.

It would be nice if this information was already included in the parser's representation, since in theory, we have the byte indices during lexing.

Though I haven't dug into the details, I assume it's enough to extend `Location` to include a byte index during lexxing -- then the parsing step would "just work"? (This does create some oddities whereby methods like `go_right` no longer make sense on `Location`, since that byte offset isn't well-defined in isolation, so perhaps it needs to be a separate piece of data, independent from `Location`.)

Is RustPython open to supporting something like this?

DimitrisJim:
> I assume it's enough to extend Location to include a byte index during lexxing

If I'm understanding the idea correctly, this would be slightly tricky mainly due to the handling of new lines on Windows. Our `NewLineHandler` simply slides over the `'\r'` character so any count based on what the `Lexer` sees would be slightly off.

If that is tackled then I can see how lexer might hold onto a running count of the byte index (incrementing it with every `next_char` call) and insert that into a `Location` when `get_pos` is called on it (AFAIK, nothing in the next stages after the lexer care what `Location`s members are, so we could add an index in there).

charliermarsh:
I'm happy to work on this when I have time. I think my main hesitation is around how it changes the semantics of `Location`. As an example: in Ruff, I sometimes want to take a location, and "jump to the next line". So I just increment the row by one, and set the column to zero. If `Location` has a `byte_index` field, though, I can't really do that anymore -- I have to figure out the byte index of that next line in order to create a `Location`. So, maybe that field becomes `Option`? Or maybe there's a separate struct, like `ContextualizedLocation`? Or maybe it's passed in as separate fields on `Located`?

DimitrisJim:
> So I just increment the row by one, and set the column to zero.

[...]

PR #4484, opened 2023-02-08
Remove NewLineHandler, move handling inline.

DimitrisJim:
Removes the new line handler window and moves the checks inside the lexer. This is slightly more complex but it does mean we don't check each source character twice (once during the NewLineHandler and once during the proper lexing phase. This does result in an 7-8% speed-up on my machine.

I've also gone ahead and documented most I could in the file.

### DIFF
--- a/compiler/parser/src/lexer.rs
+++ b/compiler/parser/src/lexer.rs
@@ -115,10 +115,6 @@ where
         *self.window.last_mut().expect("never empty") = next;
         next
     }
-
-    fn change_first(&mut self, ch: char) {
-        *self.window.first_mut().expect("never empty") = Some(ch);
-    }
 }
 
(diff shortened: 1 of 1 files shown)


DimitrisJim:
@charliermarsh if you could run ruff against this before we decide on merging it would be great.

charliermarsh:
Before:

Screen Shot 2023-02-08 at 10 50 42 AM

After:

Screen Shot 2023-02-08 at 10 56 14 AM

charliermarsh:
A nice improvement :)

charliermarsh:
(All tests pass too.)

DimitrisJim:
Great, this change would also make it easier to support #4468 (though having a clean way to do it would still need some thought) since the lexer would see every single character of the source.

youknowone (approved):

PR #4488, opened 2023-02-10
Use entire range for generators-as-arguments

charliermarsh:
Given the following:

```py
" ".join(
    sql
    for sql in (
        "LIMIT %d" % limit if limit else None,
        ("OFFSET %d" % offset) if offset else None,
    )
)
```

The `GeneratorExp` was using the range of the `sql` token on the second line, as opposed to the entire range of the actual generator.

### DIFF
--- a/compiler/parser/src/parser.rs
+++ b/compiler/parser/src/parser.rs
@@ -311,6 +311,19 @@ with (0 as a, 1 as b,): pass
     }
 
     #[test]
+    fn test_generator_expression_argument() {
+        let source = r#"' '.join(
+    sql
+    for sql in (
+        "LIMIT %d" % limit if limit else None,
+        ("OFFSET %d" % offset) if offset else None,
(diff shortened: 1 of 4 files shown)


compiler/parser/src/parser.rs
```diff
@@ -310,6 +310,19 @@ with (0 as a, 1 as b,): pass
...
+        let source = r#"' '.join(
+    sql
+    for sql in (
+        "LIMIT %d" % limit if limit else None,
+        ("OFFSET %d" % offset) if offset else None,
+    )
+)"#;
+        let parse_ast = parse_expression(source, "<test>").unwrap();
+        insta::assert_debug_snapshot!(parse_ast);
+    }
```
charliermarsh, line 324: Thank you, I was lazy 🤦

youknowone (approved):

PR #4489, opened 2023-02-11
Refactor: Join string and string_parser.

DimitrisJim:
Didn't see a reason to keep these separate, they both deal with parsing strings.


DimitrisJim:
@harupy do you by any chance remember if there was some rationale for keeping them separate, I couldn't find anything in the pr (https://github.com/RustPython/RustPython/pull/4384) that introduced `string_parser`.

PR #4490, opened 2023-02-11
Add tests, some comments, to function.rs.

DimitrisJim:


compiler/parser/src/function.rs
```diff
@@ -1,16 +1,21 @@
+// Contains functions that perform validation and parsing of arguments and parameters.
+// Checks apply both to functions and to lambdas.
```
youknowone, line 2: Did you intend doc-comment for module?

```suggestion
//! Contains functions that perform validation and parsing of arguments and parameters.
//! Checks apply both to functions and to lambdas.
```

youknowone (approved):

compiler/parser/src/function.rs
DimitrisJim: yeah since it isn't exposed atm and I don't think we'll ever expose it (since its an internal parser detail).

PR #4491, opened 2023-02-12
fix the typos

howjmay:


vm/src/builtins/float.rs
```diff
@@ -471,7 +471,7 @@ impl PyFloat {
         }
 
         let ratio = Ratio::from_float(value).unwrap();
-        let numer = vm.ctx.new_bigint(ratio.numer());
+        let number = vm.ctx.new_bigint(ratio.numer());
```
DimitrisJim, line 474: should be `numer` here, it denotes the numerator :smile:

Lib/distutils/ccompiler.py
```diff
@@ -392,7 +392,7 @@ def _fix_compile_args(self, output_dir, macros, include_dirs):
         return output_dir, macros, include_dirs
 
     def _prep_compile(self, sources, output_dir, depends=None):
-        """Decide which souce files must be recompiled.
+        """Decide which source files must be recompiled.
```
DimitrisJim, line 395: We generally dont touch the Python code since that's usually copied directly over from the CPython codebase.

Lib/distutils/unixccompiler.py
```diff
@@ -155,7 +155,7 @@ def link(self, target_desc, objects,
                                         runtime_library_dirs)
         libraries, library_dirs, runtime_library_dirs = fixed_args
 
-        # filter out standard library paths, which are not explicitely needed
+        # filter out standard library paths, which are not explicitly needed
```
DimitrisJim, line 158: similarly here

Lib/test/test_wsgiref.py
```diff
@@ -581,7 +581,7 @@ def testEnviron(self):
         # Test handler.environ as a dict
         expected = {}
         setup_testing_defaults(expected)
-        # Handler inherits os_environ variables which are not overriden
+        # Handler inherits os_environ variables which are not overridden
```
DimitrisJim, line 584: and here

Lib/test/cfgparser.2
```diff
@@ -338,7 +338,7 @@
 
 [homes]
    comment = Home Directories
-   browseable = no
+   browsable = no
```
DimitrisJim, line 341: and the changes here

DimitrisJim (changes requested):
thanks! I've left some comments for you.

Lib/distutils/ccompiler.py
youknowone: You can send the same fix to cpython project. Then we will bring it from their next release.

https://github.com/python/cpython

howjmay: Thank you! Let me fix them there too!

DimitrisJim (approved):

howjmay:
Thanks

PR #4496, opened 2023-02-13
optimize str.(l|r)strip

youknowone:


Lib/test/test_bigmem.py
```diff
@@ -765,16 +765,6 @@ def test_translate(self, size):
         self.assertEqual(s.count(_('!')), repeats * 2)
         self.assertEqual(s.count(_('z')), repeats * 3)
 
-    # TODO: RUSTPYTHON
```
DimitrisJim, line 768: did we add these in the past?

youknowone: I think so. This is a pattern to mark superclass'es test expected failure in subclass.

DimitrisJim (approved):

Issue #4497, opened 2023-02-13
Optimize bytes-like (l|r)strip


youknowone:
See #4496 for similar work

PR #4500, opened 2023-02-14
Optimize bytes-like (l|r)strip

dannasman:
Hi!

Here is a solution proposal to #4497. I was not quite sure what the optimization part of #4496 was (I assumed it was the if-else part).

### DIFF
--- a/vm/src/builtins/bytes.rs
+++ b/vm/src/builtins/bytes.rs
@@ -372,10 +372,10 @@ impl PyBytes {
         vm: &VirtualMachine,
     ) -> PyRef<Self> {
         let stripped = zelf.inner.lstrip(chars);
-        if stripped == zelf.as_bytes().to_vec() {
+        if stripped == zelf.as_bytes() {
             zelf
         } else {
-            vm.ctx.new_bytes(stripped)
+            vm.ctx.new_bytes(stripped.to_vec())
         }
     }
 
@@ -389,7 +389,7 @@ impl PyBytes {
         if stripped == zelf.as_bytes().to_vec() {
             zelf
         } else {
-            vm.ctx.new_bytes(stripped)
+            vm.ctx.new_bytes(stripped.to_vec())
         }
     }
 
--- a/vm/src/builtins/bytes.rs
+++ b/vm/src/builtins/bytes.rs
@@ -366,7 +366,11 @@ impl PyBytes {
     }
 
     #[pymethod]
-    fn lstrip(zelf: PyRef<Self>, chars: OptionalOption<PyBytesInner>, vm: &VirtualMachine) -> PyRef<Self> {
+    fn lstrip(
+        zelf: PyRef<Self>,
+        chars: OptionalOption<PyBytesInner>,
+        vm: &VirtualMachine,
+    ) -> PyRef<Self> {
         let stripped = zelf.inner.lstrip(chars);
         if stripped == zelf.as_bytes().to_vec() {
             zelf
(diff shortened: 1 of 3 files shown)


vm/src/builtins/bytes.rs
```diff
@@ -366,13 +366,31 @@ impl PyBytes {
     }
 
     #[pymethod]
-    fn lstrip(&self, chars: OptionalOption<PyBytesInner>) -> Self {
-        self.inner.lstrip(chars).into()
+    fn lstrip(
+        zelf: PyRef<Self>,
+        chars: OptionalOption<PyBytesInner>,
+        vm: &VirtualMachine,
+    ) -> PyRef<Self> {
```
youknowone, line 373: I found bytes.lstrip has different issue to str.lstrip. zelf.inner.lstrip already returns `Vec<u8>`, which means unnecessary copy happened.
I think this duplication check need to be done inside `zelf.inner.lstip`. otherwise zelf.inner.lstip also could return `&[u8]` instead of `Vec<u8>`, but it may not be easy due to lifetime.

dannasman: I see. If the `zelf.inner.lstrip` returns `&[u8]` is it possible to turn `&[u8]` to `Vec<u8>` without copying the values or is there a implementation for turning `&[u8]` directly to `PyBytesInner`? Because otherwise `zelf.inner.elements` is unnecessarily copied in the case of duplication.

dannasman: Could wrapping the return value of `zelf.inner.lstrip` in `Option` be an option? In that case it would return None on duplication and `zelf` would be returned.

vm/src/builtins/bytes.rs
```diff
@@ -366,13 +366,31 @@ impl PyBytes {
...
+        let stripped = zelf.inner.lstrip(chars);
+        if stripped == zelf.as_bytes().to_vec() {
```
fanninpm, line 375: Does this work?

```suggestion
        if stripped == zelf.as_bytes() {
```

vm/src/builtins/bytes.rs
```diff
@@ -366,13 +366,31 @@ impl PyBytes {
...
     #[pymethod]
-    fn rstrip(&self, chars: OptionalOption<PyBytesInner>) -> Self {
-        self.inner.rstrip(chars).into()
+    fn rstrip(
+        zelf: PyRef<Self>,
+        chars: OptionalOption<PyBytesInner>,
+        vm: &VirtualMachine,
+    ) -> PyRef<Self> {
+        let stripped = zelf.inner.rstrip(chars);
+        if stripped == zelf.as_bytes().to_vec() {
```
fanninpm, line 389:
```suggestion
        if stripped == zelf.as_bytes() {
```

fanninpm (commented):
Does this work? I don't think we need the explicit conversion in these two places, as `PartialEq<&[U]>` is implemented for `Vec<T, A>`.

vm/src/builtins/bytes.rs
youknowone: I think `&stripped == zelf.as_bytes()` works, but still stripped is `Vec<u8>`.

vm/src/builtins/bytes.rs
fanninpm: `PyBytesInner` is a wrapper around `Vec<u8>`, so allocation is inevitable. Fortunately, converting a `Vec<u8>` into a `PyBytesInner` only takes ownership of that `Vec<u8>`.

youknowone (approved):
Great, Thank you!

PR #4504, opened 2023-02-14
PyObjectRef::downcast_exact returns PyRefExact

youknowone:


DimitrisJim:
this looks fine to me but I'm mostly basing it on the tests passing fine. In general, I feel like we need to better document these core traits we have since there's many and the differences between them aren't always so obvious.

vm/src/object/ext.rs
```diff
@@ -107,6 +107,7 @@ impl<T: PyPayload> std::borrow::ToOwned for PyExact<T> {
     }
 }
 
+/// PyRef but guaranteed not to be a subtype instance
```
youknowone, line 110: @DimitrisJim I don't think this single line fulfill your concern, but it is mostly it is.

Other types are roughly tabled in https://github.com/RustPython/RustPython/blob/main/vm/src/object/core.rs#L1-L12

Other types in core and ext are mostly hidden - hopefully

DimitrisJim: Thanks! At some point in the future (after I'm hopefully done documenting the compiler), I'll start documenting as much of the vm as possible. We should set up a call then to discuss all these so I can get the docs right.

PR #4507, opened 2023-02-15
Add co_cellvars to code object

### LINKED ISSUE #4501: Add `co_cellvars` to code object
Opening this separately from #1838 since it seems like a good first issue.

Currently, code objects are missing the `co_cellvars` attribute:

```python
def foo(a, b):
    def bar(c, d):
        print(a, b, c, d)
```

In RustPython:

```python
>>>>> foo.__code__.co_cellvars
AttributeError: 'code' object has no attribute 'co_cellvars'
```

While in Python:

```python
>>> foo.__code__.co_cellvars
('a', 'b')
```

looking at the other methods on [`PyCode`](https://github.com/RustPython/RustPython/blob/1fd557d549cd72e7da83608b99e5464273abd4b3/vm/src/builtins/code.rs#L188) should help, the implementation should be similar to others there that return a tuple.

howjmay:
Closes #4501

### DIFF
--- a/vm/src/builtins/code.rs
+++ b/vm/src/builtins/code.rs
@@ -219,6 +219,18 @@ impl PyRef<PyCode> {
     }
 
     #[pygetset]
+    pub fn co_cellvars(self, vm: &VirtualMachine) -> PyTupleRef {
+        let cellvars = self
+            .code
+            .cellvars
+            .deref()
+            .iter()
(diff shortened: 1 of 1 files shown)


DimitrisJim (approved):
looks good, thanks @howjmay!

howjmay:
@DimitrisJim Thank you too

PR #4513, opened 2023-02-16
Optimize Py<PyDict>::to_attributes() to reuse PyRefExact

### LINKED ISSUE #4503: Optimize Py<PyDict>::to_attributes() to reuse PyRefExact
`PyAttributes` only uses `PyStrInterned` as its key.
`PyStrInterned` is always compatible to `PyRefExact<PyStr>`

in `Py<PyDict>::to_attributes()` it creates interned string using `vm.ctx.intern_str(key.as_str())` which always create a new PyObjectRef using `&str`(from key.as_str()).
If we downcast it to `PyRefExact<PyStr>` instead of `PyStrRef` and put it directly to `vm.ctx.intern_str`, it will reuse the given object unlike taking `&str` like now.

This is assigned to @xiaozhiyan

xiaozhiyan:
Closes #4503

closes #4504

### DIFF
--- a/vm/src/builtins/complex.rs
+++ b/vm/src/builtins/complex.rs
@@ -128,7 +128,7 @@ impl Constructor for PyComplex {
                 let val = if cls.is(vm.ctx.types.complex_type) && imag_missing {
                     match val.downcast_exact::<PyComplex>(vm) {
                         Ok(c) => {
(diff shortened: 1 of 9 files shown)


youknowone (approved):
Thank you!

PR #4514, opened 2023-02-17
narrow publicity of BytesInner::elements

youknowone:


DimitrisJim (approved):
+1 we need to think about narrowing as much as possible, I think there's many places we leak the implementation details for stuff

PR #4518, opened 2023-02-17
Remove unnecessary to_vec() in bytes.rs

dannasman:
One unnecessary `to_vec()` was not removed in #4500.


fanninpm (approved):
Good catch.

Issue #2759, opened 2021-07-29
Recent changes to CPython's grammar and PEG parser, PEP-617


sobolevn:
## Feature
Python3.9+ support some new grammar rules, like `()` around `with` expression:

```pycon
» python3.9 --version
Python 3.9.4
            
» python3.9
Python 3.9.4 (default, Apr  5 2021, 01:47:16) 
[Clang 11.0.0 (clang-1100.0.33.17)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> with (open('a') as f):
...     print(f)
... 
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: 'a'
```

But, RustPython does not support it:
<img width="961" alt="Снимок экрана 2021-07-29 в 12 54 24" src="https://user-images.githubusercontent.com/4660275/127471814-00c35e98-ebeb-473e-a219-1a8d73814eba.png">

I think that the root of this problem is that new PEG parser is not supported: https://github.com/RustPython/RustPython/blob/master/parser/src/python.lalrpop

## Python Documentation

https://www.python.org/dev/peps/pep-0617/

DimitrisJim:
Yup, the project is currently targeting 3.8 so this shouldn't come as a surprise. Hopefully, at some point in the future, 3.9 will be the new goal and using a PEG parser will probably be needed.

sobolevn:
Then, I guess we can change this line in the README, because `3.9>=3.8` 🙂 
<img width="501" alt="Снимок экрана 2021-07-29 в 19 09 31" src="https://user-images.githubusercontent.com/4660275/127527041-337577e4-00eb-44d4-94ef-16e57d16390d.png">

sobolevn:
Also: https://github.com/RustPython/RustPython/pull/2417

youknowone:
I think we are practically not pinning a target. Any patch for 3.9+ features will be appreciated.

fanninpm:
I believe 3.8 is the absolute minimum because it's referenced in `.github/workflows/ci.yaml`.

youknowone:
we just can change it to 3.9 once we feel it is more convenient. that happened a few times.

youknowone:
#2831

sobolevn:
Related https://github.com/zsol/LibCST/blob/parser/native/libcst_parser/src/grammar.rs

youknowone:
[...]

PR #4519, opened 2023-02-18
Add support for match statements to parser

charliermarsh:
# Summary

This is a work-in-progress, but it does work thus far for the implemented cases. I'm putting it up now to (1) signal that I'm working on it, and (2) get feedback from anyone else that has explored this and can see problems ahead.

closes #4323
closes #3427


andersk:
The main challenge here is that `match` and `case` are [soft keywords](https://docs.python.org/3/reference/lexical_analysis.html#soft-keywords). Naïvely adding them as keyword tokens will break parsing of code like

```python
match = re.match(pattern, str)
if match is not None:
    print(match[1])
```

that uses `match` as an identifier.

We typically don’t know whether `match` should be treated as a keyword or an identifier until many tokens later. Some stupid examples I’ve collected where this results in interesting parse tree differences:

```python
match * a + b, c   # ((match * a) + b), c
match * a + b, c:  # match ((*(a + b)), c):

match -a * b + c   # (match - (a * b)) + c
match -a * b + c:  # match ((-a) * b) + c):

match (a).b   # (match(a)).b
match (a).b:  # match ((a).b):

match [a].b   # (match[a]).b
match [a].b:  # match ([a].b):

match [a,].b   # (match[(a,)]).b   (not (match[a]).b)
match [a,].b:  # match ([a].b):    (not match ([(a,)].b):)
```

My rough plan was going to be to start parsing `match * …`, `match(…`, `match[EXPR]…` under the assumption that it’s an expression with a `match` identifier, but start parsing `match + …`, `match - …`, `match[EXPR, …` under the assumption that it’s a `match` statement, so that in either case there should be enough information to convert to the other form later if we find the initial assumption was wrong. (This specific initial classification is important: there’s not enough information to translate in the other direction in either case.)

charliermarsh:
Thanks, this is extremely helpful. (I kind of assumed I was missing something important given how smoothly this went.)

If you’re further along and want to keep pushing on it, I can close and defer to you. Or if you want to collaborate on a branch that works too. Or I can keep trying here based on some of the pointers above. Whatever you prefer.

charliermarsh:
Some discussion from Parso that may or may not end up being relevant here: https://github.com/davidhalter/parso/issues/138#issuecomment-817378192

charliermarsh:
To what degree would getting those LALRPOP changes merge help here? Do we need to consider publishing a fork for now?

charliermarsh:
I also found this interesting on how Python 3.5 and earlier handled async and await as soft keywords via lexer tricks, but I doubt they’d help us here: https://benjam.info/blog/posts/2019-09-18-python-deep-dive-tokenizer/#async-and-await

charliermarsh:
Although… could we have the lexer look ahead? Is that crazy? Isn’t it a match statement IFF the line starts with match and ends in a colon? (Haven’t thought this through, I assume that won’t work.)

andersk:
No I don’t think a lexer trick like that can work:

```python
match()[a:
    b]
match()[a:
    b]:
```

charliermarsh:
Well you’d need to skip over bracketed content, no? I.e., look ahead until you hit a logical newline (not a physical newline)? At which point you’d either have a semicolon or not.

charliermarsh:
Put differently, what I’m asking is whether we could backtrack or lookahead in the lexer (where we have more control) rather than complicating the parser. E.g. I think LALRPOP does some backtracking in its own lexer (https://github.com/lalrpop/lalrpop/blob/b777a6a884052d6283971f71b3ba453aa6d417ab/lalrpop/src/tok/mod.rs#L138), but I’d have to try this myself to have a real opinion on it.

compiler/parser/python.lalrpop
```diff
@@ -357,6 +358,336 @@ CompoundStatement: ast::Stmt = {
...
+    }
+}
+
+Pattern: ast::Pattern = {
+    <pattern:AsPattern> => pattern,
+    <pattern:OrPattern> => pattern,
+}
+
+AsPattern: ast::Pattern = {
+   <location:@L> <name:Identifier> <end_location:@R> => ast::Pattern {
```
DimitrisJim, line 397: isn't `'as'` allowed here? I.e, probably two alternatives where either a single ident is allowed (as you've currently allowed) and a `<pattern:OrPattern> "as" <alias:Identifier>`.

DimitrisJim: though I haven't tried to add this yet so lalrpop might not like it (though if it parses the current grammar it should parse that)

charliermarsh: Yes, oops.

charliermarsh: (It's just one case here, the no-`as` case is a case on `ClosedPattern`. Have it working locally.)

charliermarsh:
This is able to parse all of [`test_patma.py`](https://github.com/python/cpython/blob/60bbed7f174e481d3fc69984430a4667951678b3/Lib/test/test_patma.py#L133).

charliermarsh:
From what I can tell, performance is the ~same or within 1% (for parsing CPython via Ruff).

compiler/parser/python.lalrpop
```diff
@@ -357,6 +358,561 @@ CompoundStatement: ast::Stmt = {
...
+    },
+    <location:@L> <e:MatchNameOrAttr> "(" ")" <end_location:@R> => {
+        ast::PatternKind::MatchClass {
+            cls: Box::new(e),
+            patterns: vec![],
+            kwd_attrs: vec![],
+            kwd_patterns: vec![],
+        }
+    },
+}
```
charliermarsh, line 914: I suspect some of this could be made more concise.

youknowone:
@andersk could you review this PR?

compiler/parser/python.lalrpop
```diff
@@ -357,6 +358,561 @@ CompoundStatement: ast::Stmt = {
...
+                operand: Box::new(ast::Expr {
+                    location,
+                    end_location: Some(end_location),
+                    custom: (),
+                    node: ast::ExprKind::Constant { value, kind: None }
+                })
+            }
+        })
+    },
+    <location:@L> <left:Constant> <op:AddOp> <right:Constant> <end_location:@R> => ast::PatternKind::MatchValue {
```
youknowone, line 590: I am worrying how many variants can be placed here and the next pattern

youknowone (approved):
not sure this design fits for full-featured match statement support, but it looks good as starting of it.

DimitrisJim (approved):

compiler/parser/src/soft_keywords.rs
```diff
@@ -0,0 +1,118 @@
...
+                            Tok::Lpar => par_count += 1,
+                            Tok::Rpar => par_count -= 1,
+                            Tok::Lsqb => sqb_count += 1,
+                            Tok::Rsqb => sqb_count -= 1,
+                            Tok::Lbrace => brace_count += 1,
+                            Tok::Rbrace => brace_count -= 1,
+                            _ => {}
+                        }
+                        first = false;
+                    }
```
DimitrisJim, line 83: Under the impression the following should work equally well since we don't really care about what the delimiter token is, just if we are in or not in one:

```suggestion
                    let mut nesting = 0;
                    let mut first = true;
                    let mut seen_colon = false;
                    while let Some(Ok((_, tok, _))) = self.underlying.peek() {
                        match tok {
                            Tok::Newline => break,
                            Tok::Colon if nesting == 0 => {
                                if !first {
                                    seen_colon = true;
                                }
                            }
                            Tok::Lpar | Tok::Lsqb | Tok::Lbrace => nesting += 1,
                            Tok::Rpar | Tok::Rsqb | Tok::Rbrace => nesting -= 1,
                            _ => {}
                        }
                        first = false;
                    }
```

charliermarsh: Oh, we may also need to handle unbalanced parens, since we're not guaranteed to get a valid stream here, right? Do unbalanced parens get handled by the lexer, or fail to parse?

DimitrisJim: we do handle them being unbalanced from what I remember, they just might not match (and its up to the parser then to reject, i.e `print("hello"}` is lexed just fine).

DimitrisJim: (pretty sure CPython should do the same, they use a single `level` counter to track the nesting)

compiler/parser/python.lalrpop
charliermarsh: Do you mean, due to the complexity of reading / understanding it? Or performance? Or something else?

youknowone:
Totally fine if andersk is busy. I forgot the specific branch, but I found @charliermarsh already tested it with ruff somewhere.

youknowone:
unpinning #2759

charliermarsh:
Yeah I've run that branch over the following codebases and, in each case, verified that the lint violation counts are identical before and after that change:

- Hatch (198 violations)
- Pandas (2766 violations)
- Zulip (151 violations)
- scikit-build-core (9 violations)
- Airflow (11 violations)
- CPython (18367 before, 18533 after; however, one fewer parse error in `test_patma.py`, so more violations are _expected_, since we're now actually analyzing that file)

compiler/parser/python.lalrpop
youknowone: I thought this is arithmetic expressions in patterns and manually covering those. Not sure how it is working, but you say it covers everything, right?

charliermarsh: Yeah, but only a few specific patterns are allowed: a signed number on its own (so "4" or "-4", but not "+4"), or an addition or subtraction where the second value is complex (so "-4 + 4j" or "4 + 4j" or "-4 - 4j" or "-4 - 4j"). Arbitrary arithmetic is not allowed.

youknowone: Got it. I was misunderstanding the pattern spec.

charliermarsh: I do think we could DRY up some of the branches here, probably.

DimitrisJim:
Merging so we don't unnecessarily hold up https://github.com/charliermarsh/ruff/issues/282 too, really big thanks for this @charliermarsh!

Issue #4511, opened 2023-02-16
OverflowError handling in rustpython_common::bytes::repr_with


youknowone:
`rustpython_common::bytes::repr_with` can raise `OverflowError` but it panics. (the `unwrap()` part)

https://github.com/RustPython/RustPython/blob/3f1fb9d29daad302f4fe1857c74f3917fad710a2/common/src/bytes.rs#L26-L27

Because `rustpython_common` crate doesn't know about Python exceptions, `repr_with` have to return temporary type (maybe `ReprError`). To convert it to python exception, `impl ToPyException for ReprError` will be helpful.

assigned to @minhrongcon2000

PR #4520, opened 2023-02-18
Make common::repr throw error instead of panic

minhrongcon2000:
A solution for #4511

### DIFF
--- a/common/src/bytes.rs
+++ b/common/src/bytes.rs
@@ -25,11 +25,7 @@ pub fn repr_with(b: &[u8], prefixes: &[&str], suffix: &str) -> Result<String, Re
             0x20..=0x7e => 1,
             _ => 4, // \xHH
         };
-        // TODO: OverflowError
-        out_len = match out_len.checked_add(incr) {
-            Some(valid) => valid,
-            None => return Err(ReprOverflowError),
-        };
+        out_len = out_len.checked_add(incr).ok_or(ReprOverflowError)?;
     }
 
     let (quote, num_escaped_quotes) = crate::str::choose_quotes_for_repr(squote, dquote);
--- a/common/src/bytes.rs
+++ b/common/src/bytes.rs
@@ -1,8 +1,10 @@
-pub fn repr(b: &[u8]) -> String {
+use crate::str::ReprOverflowError;
+
+pub fn repr(b: &[u8]) -> Result<String, ReprOverflowError> {
     repr_with(b, &[], "")
 }
 
-pub fn repr_with(b: &[u8], prefixes: &[&str], suffix: &str) -> String {
+pub fn repr_with(b: &[u8], prefixes: &[&str], suffix: &str) -> Result<String, ReprOverflowError> {
     use std::fmt::Write;
 
     let mut out_len = 0usize;
@@ -24,7 +26,10 @@ pub fn repr_with(b: &[u8], prefixes: &[&str], suffix: &str) -> String {
             _ => 4, // \xHH
         };
         // TODO: OverflowError
-        out_len = out_len.checked_add(incr).unwrap();
+        out_len = match out_len.checked_add(incr) {
+            Some(valid) => valid,
+            None => return Err(ReprOverflowError),
+        };
     }
 
     let (quote, num_escaped_quotes) = crate::str::choose_quotes_for_repr(squote, dquote);
@@ -57,5 +62,5 @@ pub fn repr_with(b: &[u8], prefixes: &[&str], suffix: &str) -> String {
     res.push(quote);
     res.push_str(suffix);
 
-    res
+    Ok(res)
 }
--- a/compiler/ast/src/constant.rs
+++ b/compiler/ast/src/constant.rs
@@ -1,3 +1,5 @@
+use std::fmt::Error;
+
 use num_bigint::BigInt;
 pub use rustpython_compiler_core::ConversionFlag;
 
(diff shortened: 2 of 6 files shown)


vm/src/exceptions.rs
```diff
@@ -1621,3 +1622,9 @@ pub(super) mod types {
         "Base class for warnings about encodings."
     }
 }
+
+impl ToPyException for ReprOverflowError {
+    fn to_pyexception(&self, vm: &VirtualMachine) -> PyBaseExceptionRef {
+        vm.new_value_error(String::from("This object has overflow repr"))
```
DimitrisJim, line 1628: use an `Overflow` error instead here (unless Python explicitly raises this ValueError, I haven't checked)

common/src/bytes.rs
```diff
@@ -24,7 +26,10 @@ pub fn repr_with(b: &[u8], prefixes: &[&str], suffix: &str) -> String {
             _ => 4, // \xHH
         };
         // TODO: OverflowError
```
DimitrisJim, line 28: can remove this `TODO` now :smile:

vm/src/builtins/bytearray.rs
```diff
@@ -126,8 +126,8 @@ impl PyByteArray {
     fn repr(zelf: PyRef<Self>, _vm: &VirtualMachine) -> PyResult<String> {
         let class = zelf.class();
         let class_name = class.name();
-        let s = zelf.inner().repr(Some(&class_name));
-        Ok(s)
+        let s = zelf.inner().repr(Some(&class_name), _vm);
```
DimitrisJim, line 129: since we now use the `vm` passed in we can rename it (since variables prefixed with `_` usually indicate it is unused). I.e rename `_vm` to `vm`.

common/src/bytes.rs
```diff
@@ -24,7 +26,10 @@ pub fn repr_with(b: &[u8], prefixes: &[&str], suffix: &str) -> String {
...
-        out_len = out_len.checked_add(incr).unwrap();
+        out_len = match out_len.checked_add(incr) {
+            Some(valid) => valid,
+            None => return Err(ReprOverflowError),
+        };
```
DimitrisJim, line 32: alternatively:

```suggestion
        out_len = out_len.checked_add(incr).ok_or(ReprOverflowError)?;
```

DimitrisJim (commented):
thanks, I've left some comments

vm/src/exceptions.rs
```diff
@@ -1621,3 +1622,9 @@ pub(super) mod types {
         "Base class for warnings about encodings."
     }
 }
+
+impl ToPyException for ReprOverflowError {
+    fn to_pyexception(&self, vm: &VirtualMachine) -> PyBaseExceptionRef {
+        vm.new_overflow_error(String::from("This object has overflow repr"))
```
youknowone, line 1628: `PyStr` do this like `vm.new_overflow_error(err.to_string())`. Using same code will be better.

youknowone (changes requested):
Thank you! I left a change request about message. Other parts looks good.

vm/src/builtins/bytearray.rs
```diff
@@ -126,8 +126,8 @@ impl PyByteArray {
     fn repr(zelf: PyRef<Self>, _vm: &VirtualMachine) -> PyResult<String> {
```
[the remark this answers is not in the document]
minhrongcon2000: I think I have fixed it in the recent commit :D

vm/src/builtins/bytearray.rs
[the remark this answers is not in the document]
minhrongcon2000: This is also fixed as well :))

vm/src/exceptions.rs
minhrongcon2000: I just fixed the new code to follow this convention. However, this leaves `impl ToPyException for ReprOverflowError` unused, but I won't remove this in case of further usage.

youknowone: We can keep your original version in bytesinner and change this to_pyexception inside to reuse err.to_string().

vm/src/bytesinner.rs
```diff
@@ -246,11 +246,13 @@ impl PyBytesInner {
...
 
-    pub fn repr(&self, class_name: Option<&str>) -> String {
+    pub fn repr(&self, class_name: Option<&str>, vm: &VirtualMachine) -> PyResult<String> {
         if let Some(class_name) = class_name {
             rustpython_common::bytes::repr_with(&self.elements, &[class_name, "("], ")")
+                .map_err(|err| vm.new_overflow_error(err.to_string()))
         } else {
             rustpython_common::bytes::repr(&self.elements)
+                .map_err(|err| vm.new_overflow_error(err.to_string()))
         }
```
youknowone, line 256:
```suggestion
        let repr = if let Some(class_name) = class_name {
            rustpython_common::bytes::repr_with(&self.elements, &[class_name, "("], ")")
        } else {
            rustpython_common::bytes::repr(&self.elements)
        };
        repr.map_err(|err| err.to_pyexception(vm))
```

compiler/ast/src/constant.rs
```diff
@@ -42,7 +44,10 @@ impl std::fmt::Display for Constant {
             Constant::None => f.pad("None"),
             Constant::Bool(b) => f.pad(if *b { "True" } else { "False" }),
             Constant::Str(s) => rustpython_common::str::repr(s).fmt(f),
-            Constant::Bytes(b) => f.pad(&rustpython_common::bytes::repr(b)),
+            Constant::Bytes(b) => f.pad(match &rustpython_common::bytes::repr(b) {
+                Ok(repr) => repr,
+                Err(_e) => return Err(Error),
+            }),
```
youknowone, line 50:
```suggestion
            Constant::Bytes(b) => f.pad(match &rustpython_common::bytes::repr(b)?,
```

youknowone (approved):
looks good, Thank you!

Issue #4522, opened 2023-02-18
Missing Python function `unicodedata.east_asian_width`


xiaozhiyan:
## Feature

Python function `unicodedata.east_asian_width` along with `unicodedata.ucd_3_2_0.east_asian_width` is missing.

## Python Documentation

https://docs.python.org/3/library/unicodedata.html#unicodedata.east_asian_width

PR #4523, opened 2023-02-18
Add unicodedata.east_asian_width

xiaozhiyan:
Closes https://github.com/RustPython/RustPython/issues/4522

~~This PR is a draft for further discussion.~~
Update: In this PR, we've adopted `ucd` crate to implement `unicodedata.east_asian_width` using database v9.0.0, while leaving `unicodedata.ucd_3_2_0.unidata_version` referring to `unicodedata.east_asian_width` as is for now.

# Python Document

https://docs.python.org/3/library/unicodedata.html#unicodedata.east_asian_width

# Expectation (as in CPython 3.11)
```
>>> import unicodedata
>>> unicodedata.unidata_version
'14.0.0'
>>> unicodedata.east_asian_width('\u231a')
'W'
>>> unicodedata.ucd_3_2_0.unidata_version
'3.2.0'
>>> unicodedata.ucd_3_2_0.east_asian_width('\u231a')
'N'
```

# Status
- `unicodedata.east_asian_width` is supported by using `ucd` crate
- `unicodedata.ucd_3_2_0.east_asian_width` is **not working** due to 9.0 change as in test https://github.com/RustPython/RustPython/blob/main/Lib/test/test_unicodedata.py#L264

# Reason
- `ucd` crate uses Unicode Character Database (v9.0.0)
- `self.assertEqual(self.db.ucd_3_2_0.east_asian_width('\u231a'), 'N')`
- https://github.com/sourtin/libucd/blob/master/src/tables/misc.rs#L768 => `W`

# Extra Information
- `unic` crate uses v10.0.0 https://github.com/open-i18n/rust-unic/blob/master/unic/ucd/age/tables/unicode_version.rsv#L3
- Thus existing functions `unicodedata.ucd_3_2_0.*` seem using v10.0.0 instead of v3.2.0 https://github.com/RustPython/RustPython/blob/main/stdlib/src/unicodedata.rs#L68

# Possible Solution
- We may (generate and) load Unicode Character Database according to `unidata_version`, including v3.2.0 and another major version (say, v14.0.0).
- `icu` crate may be a good option to load dynamic database https://github.com/unicode-org/icu4x/blob/main/components/properties/src/maps.rs#L376
- That means we may need to change all functions in `unicodedata` module to use database determined by `unidata_version`.

# Example Commits
- `ucd` crate: https://github.com/RustPython/RustPython/pull/4523/commits/420240c402e3cf4cc4baeec61a36659e2d1de8b1
- `icu` crate: https://github.com/RustPython/RustPython/pull/4523/commits/38eaa75c7e1fae44bc1a08add8a2d2ffee1f79c8

# Related Tests
- `cargo run --release -- -m test test_unicodedata -v`
- `cargo run --release -- extra_tests/snippets/builtin_str_unicode.py`

### DIFF
--- a/stdlib/src/unicodedata.rs
+++ b/stdlib/src/unicodedata.rs
@@ -10,9 +10,16 @@ pub fn make_module(vm: &VirtualMachine) -> PyObjectRef {
         .into_ref(vm)
         .into();
 
-    for attr in ["category", "lookup", "name", "bidirectional", "normalize"]
-        .iter()
-        .copied()
+    for attr in [
+        "category",
+        "lookup",
+        "name",
+        "bidirectional",
+        "east_asian_width",
+        "normalize",
+    ]
+    .iter()
+    .copied()
     {
         crate::vm::extend_module!(vm, &module, {
             attr => ucd.get_attr(attr, vm).unwrap(),
@@ -29,6 +36,7 @@ mod unicodedata {
         VirtualMachine,
     };
     use itertools::Itertools;
+    use ucd::{Codepoint, EastAsianWidth};
     use unic_char_property::EnumeratedCharProperty;
     use unic_normal::StrNormalForm;
     use unic_ucd_age::{Age, UnicodeVersion, UNICODE_VERSION};
@@ -113,6 +121,16 @@ mod unicodedata {
             Ok(bidi.to_owned())
         }
 
+        /// NOTE: This function uses 9.0.0 database instead of 3.2.0
+        #[pymethod]
+        fn east_asian_width(&self, character: PyStrRef, vm: &VirtualMachine) -> PyResult<String> {
+            Ok(self
+                .extract_char(character, vm)?
+                .map_or(EastAsianWidth::Neutral, |c| c.east_asian_width())
+                .abbr_name()
+                .to_owned())
+        }
+
         #[pymethod]
         fn normalize(
             &self,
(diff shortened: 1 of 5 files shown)


stdlib/src/unicodedata.rs
```diff
@@ -138,6 +157,24 @@ mod unicodedata {
         }
     }
 
+    trait EastAsianWidthAbbrName {
+        fn abbr_name<'a>(self) -> &'a str;
+    }
+
+    impl EastAsianWidthAbbrName for EastAsianWidth {
+        fn abbr_name<'a>(self) -> &'a str {
```
youknowone, line 165:
```suggestion
        fn abbr_name(self) -> &'static str {
```

stdlib/src/unicodedata.rs
```diff
@@ -113,6 +121,17 @@ mod unicodedata {
             Ok(bidi.to_owned())
         }
 
+        #[pymethod]
```
youknowone, line 124:
```suggestion
        /// NOTE: This function uses 10.0.0 database instead of 3.2.0
        #[pymethod]
```

Maybe this is enough for the version mismatch for now.

youknowone (approved):
Looks great. Thank you. Do you have more concerns not to make this ready?

stdlib/src/unicodedata.rs
xiaozhiyan: Let me change this.

stdlib/src/unicodedata.rs
xiaozhiyan: I see.
Current `unic` functions use v10.0.0, and in this PR we can introduce `ucd` crate for `unicodedata.east_asian_width` using v9.0.0.
Then we just add the above comment and leave `unicodedata.ucd_3_2_0.east_asian_width` as is for now.
Let me remove the example commit for `icu` and update the PR.

xiaozhiyan: PR is updated as in commit https://github.com/RustPython/RustPython/commit/7661fab743fb06baab3c86f60d6dedbd07b6b72b .

xiaozhiyan:
> Looks great. Thank you. Do you have more concerns not to make this ready?

I've updated the PR (as in commit https://github.com/RustPython/RustPython/commit/7661fab743fb06baab3c86f60d6dedbd07b6b72b) and made it ready. Please help to merge it.

Issue #4481, opened 2023-02-06
[RFC] Replace memchr dependency with std::slice::fill


youknowone:
## Summary

memchr can be replaced with [std::slice::fill](https://doc.rust-lang.org/std/primitive.slice.html#method.fill)

Jhowww:
Hi @youknowone, I can give it a shot.

Jhowww:
Hi @youknowone and @DimitrisJim, I have looked at the code that uses memchr, but I have no direction how to use [std::slice::fill](https://doc.rust-lang.org/std/primitive.slice.html#method.fill) to replace memchr?
I think that memchr is find a element in the slice and std::slice::fill is change the in the slice to some value.
Seems like I should use position(https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.position) or find

, like https://github.com/RustPython/RustPython/blob/61b48f108982d865524f86624a9d5bc2ae3bccef/vm/src/stdlib/io.rs#L1905-L1915
Seem like I should use find instead of [std::slice::fill](https://doc.rust-lang.org/std/primitive.slice.html#method.fill)?

youknowone:
Oh, I am sorry. It's my fault. You are right.

youknowone:
I am sorry. I made the wrong issue. memchr is highly optimized so we can't replace it with std::iter.
I thought it was replacable using newly added std::slice::fill, but it wasn't.

PR #4525, opened 2023-02-19
Replace memchr dependency with std::iter

Jhowww:
Closes https://github.com/RustPython/RustPython/issues/4481

### DIFF
--- a/Lib/test/test_unicode.py
+++ b/Lib/test/test_unicode.py
@@ -208,7 +208,7 @@ class UnicodeTest(string_tests.CommonTest,
 
     def test_find(self):
         string_tests.CommonTest.test_find(self)
-        # test implementation details of the memchr fast path
+        # test implementation details of the std::iter fast path
         self.checkequal(100, 'a' * 100 + '\u0102', 'find', '\u0102')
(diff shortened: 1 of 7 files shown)


Lib/test/test_unicode.py
```diff
@@ -208,7 +208,7 @@ def test_count(self):
 
     def test_find(self):
         string_tests.CommonTest.test_find(self)
-        # test implementation details of the memchr fast path
+        # test implementation details of the std::iter fast path
```
fanninpm, line 211: Is this in the CPython source?

DimitrisJim: yeah, its okay if we don't change this to say we use `std::iter`.

youknowone:
I am sorry. I made the wrong issue. memchr is highly optimized so we can't replace it with std::iter.
I thought it was replacable using newly added std::slice::fill, but it wasn't.
I am really sorry to make you work on it.
If you don't mind, can I share you another (real) good first issue for you?

Jhowww:
It's okay, we can move on to a new issue.
Please assign me that new one, thanks.

PR #4526, opened 2023-02-20
Update hmac.py to CPython 3.11 and integrate test_hmac.py tests

haxelion:
Hi!

This is a bit of a rabbit hole but:

- I noticed the function `digest()` was missing from the `hmac` module so I updated `hmac.py` to the version from CPython 3.11
- I noticed `test_hmac.py` was missing from `Lib/test` so I also imported `test_hmac.py` from CPython 3.11
- I saw some test were failing because the `update` method of the `PyHasher` struct was taking the wrong argument type so I fixed that.
- I saw some other tests were failing because the `PyHasher` struct was not implementing the python `copy` method so I fixed that as well.
- Finally running `test_hmac.py` I noticed that two tests (`test_sha384_rfc4231` and `test_sha512_rfc4231`) are returning incorrect output so I marked them as expected failure.

The incorrect output probably need some more digging to see why that hmac implementation is wrong.

### DIFF
--- a/stdlib/src/hashlib.rs
+++ b/stdlib/src/hashlib.rs
@@ -4,12 +4,13 @@ pub(crate) use hashlib::make_module;
 mod hashlib {
     use crate::common::lock::{PyRwLock, PyRwLockReadGuard, PyRwLockWriteGuard};
     use crate::vm::{
-        builtins::{PyBytes, PyBytesRef, PyStrRef, PyTypeRef},
-        function::{FuncArgs, OptionalArg},
+        builtins::{PyBytes, PyStrRef, PyTypeRef},
+        function::{ArgBytesLike, FuncArgs, OptionalArg},
         PyPayload, PyResult, VirtualMachine,
     };
     use blake2::{Blake2b512, Blake2s256};
     use digest::DynDigest;
+    use dyn_clone::{clone_trait_object, DynClone};
     use md5::Md5;
     use sha1::Sha1;
     use sha2::{Sha224, Sha256, Sha384, Sha512};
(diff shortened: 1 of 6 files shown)


DimitrisJim:
There's a test passing (test_file (test.test_urllib2.HandlerTests) ... unexpected success) but only on Mac for some reason, you could just mark it as ok only for that case.

haxelion:
@DimitrisJim It unfortunately fails on my mac ... so I'm unsure what's going on there ...

DimitrisJim:
The runner uses a macOS 12.6.3 so it could just be a difference due to that (I doubt you'd want to jump into this rabbit hole too).

If you conditionally skip it for the rest and just add a comment that mentions this oddity we should be fine with it. (and if it blows up when the runner is updated and that caused the issue, we'll know).

haxelion:
@DimitrisJim I'm also running macOS 12.6.3. Maybe it's a difference with the urllib2 installed on my system. Let me check again.

stdlib/src/hashlib.rs
```diff
@@ -260,8 +265,24 @@ mod hashlib {
...
+                    fn box_clone(&self) -> Box<dyn ThreadSafeDynDigest> {
+                        Box::new(self.clone())
+                    }
+                }
+            )+
+        };
+    }
+
+    impl_thread_safe_dyn_digest!({Md5, Sha1, Sha224, Sha256, Sha384, Sha512,
+        Sha3_224, Sha3_256, Sha3_384, Sha3_512, Blake2b512, Blake2s256});
```
youknowone, line 285: maybe you wanted to use https://crates.io/crates/dyn-clone

youknowone (commented):
Welcome. The rabbit hole process was our standard development paradigm for a while. I am happy to share it with you 😉

youknowone:
The failing test is weird and I don't think investigating it is coverage of this test.
It also expected failure on my machine.

Just changing the decorator from `@unittest.expectedFailure` to `@unittest.skip` will be fine.

stdlib/src/hashlib.rs
haxelion: Yes indeed!
I updated the code to use the dyn-clone crate instead!

youknowone (approved):
Looks great, thank you!

Issue #4515, opened 2023-02-17
Recently added `_feature_version` parameter in `compile()` is not supported yet


youknowone:
## Feature

it prints error messages:
```
TODO: compile() got `_feature_version` but ignored
```

dannasman:
Hi!

How and where exactly should the `_feature_version` be used? I looked at CPython and to my understanding it is used to compute parser flags there.

youknowone:
I didn't investigate what it is. But it is used in https://github.com/python/cpython/blob/3.11/Lib/ast.py#L50-L51
If it is harmless, just safely ignoring without error message will be fine.

DimitrisJim:
yeah, I don't know how much sense it makes to support it. `feature_version` allows parsing code as previous versions of Python. I think we can ignore it and add a TODO instead (so it doesn't spam the output).

PR #4528, opened 2023-02-20
Switch `eprintln!` to `TODO` when handling `feature_version`

dannasman:
Switch `eprintln!` to `TODO` when handling `feature_version` in `builtins.rs`. Solves issue #4515.


dannasman:
Could the failed Windows-specific test be caused by [this CPython issue](https://github.com/python/cpython/issues/75720) ?

youknowone (approved):
Maybe. I recently reduced the time but have to increase it to prevent it fails

youknowone:
Thank you!

Issue #4479, opened 2023-01-31
Two CPython 3.11 syntax changes involving * operator are unsupported


ned2:
## Feature

In Python 3.11, the `except*` was added for working with `ExceptionGroup`s. The availability of `ExceptionGroup` itself will depend on #4247, but in the meantime, the following snippet is valid CPython 3.11 that currently throws a `SyntaxError` in RustPython:

```python
try:
    print("yay")
except* Exception as e:
    print("oh no")
```

I discovered this while digging into some code that [ruff cannot handle](https://github.com/charliermarsh/ruff/issues/2321), which on the surface looks like a slightly different issue. The following snippet is valid CPython 3.11 but also throws a `SyntaxError` in RustPython:

```python
my_dict = {}
my_dict[*"ab"] = 1
```

The expected behaviour is that the string should be unpacked and the value of the dict with key `('a','b')` set to 1. In case that feels a little contrived, here's what I was doing at the time:

```python
import typing
values = ["dog", "cat"]
Pets = typing.Literal[*values]
```

## Python Documentation

The first incompatibility issue is clearly documented [here](https://docs.python.org/3/reference/compound_stmts.html#except-star), and called out under _New syntax features_ in [What’s New In Python 3.11](https://docs.python.org/3/whatsnew/3.11.html).

However the second incompatibility is not documented in the changelog, and I did confirm that this was new in CPython 3.11, as the snippet throws a syntax error in 3.10.

I wonder if this was a (nice?) side-effect of the changes that were introduced to make `except*` work. There is a [seemingly related precedent for this](https://github.com/python/cpython/issues/90881), where using `*` to unpack without parentheses in `for` statements was enabled via the PEG parser in CPython 3.9, but was not noticed then and only documented in 3.11.

DimitrisJim:
Unpacking in indexes was introduced with PEP 646 (specifically, see [here]). I guess it isn't documented in that section because it isn't new syntax per se, more of a change in where `*` is allowed (it probably should be in the [Other Language Changes](https://peps.python.org/pep-0646/#change-1-star-expressions-in-indexes) though, which also mentions another small syntax change).

ned2:
ahhh nice yes, that's exactly it! turns out that is a hard document to find with Google. I certainly looked for it. 

yeah agreed, would have been good to see that in the Other Language Changes section

charliermarsh:
Fixed :)

youknowone:
resolved by @charliermarsh , thanks!

[...]

PR #4531, opened 2023-02-20
Implement except* syntax

charliermarsh:
This PR implements the `TryStar` spec documented [here](https://docs.python.org/3/reference/compound_stmts.html#except-star).

Note that this PR merely adds support to the parser and AST representation, rather than the runtime itself.

See: #4479.

### DIFF
--- a/compiler/codegen/src/compile.rs
+++ b/compiler/codegen/src/compile.rs
@@ -709,6 +709,12 @@ impl Compiler {
                 orelse,
                 finalbody,
             } => self.compile_try_statement(body, handlers, orelse, finalbody)?,
+            TryStar {
+                body,
+                handlers,
+                orelse,
+                finalbody,
+            } => self.compile_try_star_statement(body, handlers, orelse, finalbody)?,
             FunctionDef {
                 name,
                 args,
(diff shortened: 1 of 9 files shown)


DimitrisJim:
> That one error may be transient?

there's some flakiness there, a re-run should get it passing (just triggered it)

compiler/codegen/src/compile.rs
```diff
@@ -709,6 +709,7 @@ impl Compiler {
                 orelse,
                 finalbody,
             } => self.compile_try_statement(body, handlers, orelse, finalbody)?,
+            TryStar { .. } => unimplemented!("except*"),
```
DimitrisJim, line 712: if you could change this into a simple call to `compile_try_star` which unconditionally raises an exception (basically a copy paste of `compile_match`. This way the interpreter won't be brought down if it encounters it someplace (e.g a test file we update).

charliermarsh: On it

compiler/parser/python.lalrpop
```diff
@@ -451,6 +451,27 @@ TryStatement: ast::Stmt = {
...
     },
+    <location:@L> "try" ":" <body:Suite> <handlers:ExceptStarClause+> <else_suite:("else" ":" Suite)?> <finally:("finally" ":" Suite)?> <end_location:@R> => {
+        let orelse = else_suite.map(|s| s.2).unwrap_or_default();
+        let finalbody = finally.map(|s| s.2).unwrap_or_default();
+        let end_location = finalbody
+            .last()
+            .map(|last| last.end_location)
+            .or_else(|| orelse.last().map(|last| last.end_location))
+            .or_else(|| handlers.last().map(|last| last.end_location))
+            .unwrap();
```
youknowone, line 462:
```suggestion
        let end_location = finalbody
            .last()
            .or_else(|| orelse.last())
            .map(|last| last.end_location)
            .or_else(|| handlers.last().map(|last| last.end_location))
            .unwrap();
```

youknowone (approved):

PR #4532, opened 2023-02-20
Allow starred expressions in subscripts

charliermarsh:
This PR implements the behavior defined in [PEP 646](https://peps.python.org/pep-0646/#change-1-star-expressions-in-indexes).

See: #4479.

### DIFF
--- a/compiler/parser/python.lalrpop
+++ b/compiler/parser/python.lalrpop
@@ -1037,7 +1037,7 @@ SubscriptList: ast::Expr = {
 };
 
 Subscript: ast::Expr = {
-    NamedExpressionTest,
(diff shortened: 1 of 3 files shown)


compiler/parser/python.lalrpop
```diff
@@ -1037,7 +1037,7 @@ SubscriptList: ast::Expr = {
 };
 
 Subscript: ast::Expr = {
-    NamedExpressionTest,
+    TestOrStarNamedExpr,
```
charliermarsh, line 1040: This seems to... just work?

youknowone (approved):
Thank you!

PR #4533, opened 2023-02-21
Implemented the block_size attribute for PyHasher

haxelion:
Following PR #4526, I decided to investigate the two failing tests.

The `PyHasher` implementation was missing the `block_size` attribute which is used by the `hmac.py` code. This code was generating a warning and defaulting to an arbitrary block size: https://github.com/RustPython/RustPython/blob/ef873b4b606f0a58e3640b6186416631fdeead26/Lib/hmac.py#L84-L95
This caused the `sha2_384` and `sha2_512` HMAC to use  an incorrect block size and generate incorrect outputs.

In the process of implementing the block size property, I also simplified the `HashWrapper` constructor to use a generic type parameter.

### DIFF
--- a/stdlib/src/hashlib.rs
+++ b/stdlib/src/hashlib.rs
@@ -9,7 +9,7 @@ mod hashlib {
         PyPayload, PyResult, VirtualMachine,
     };
     use blake2::{Blake2b512, Blake2s256};
-    use digest::DynDigest;
+    use digest::{core_api::BlockSizeUser, DynDigest};
     use dyn_clone::{clone_trait_object, DynClone};
     use md5::Md5;
     use sha1::Sha1;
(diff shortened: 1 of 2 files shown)


stdlib/src/hashlib.rs
```diff
@@ -274,78 +300,29 @@ mod hashlib {
     /// Generic wrapper patching around the hashing libraries.
     #[derive(Clone)]
     struct HashWrapper {
+        block_size: usize,
         inner: Box<dyn ThreadSafeDynDigest>,
     }
 
     impl HashWrapper {
-        fn new<D: 'static>(d: D) -> Self
+        fn new<D>() -> Self
```
youknowone, line 308: I like this version of HashWrapper. Thank you!

youknowone (approved):

PR #4534, opened 2023-02-21
Copy graphlib module from cpython

seungha-kim:
This PR is no-brainer copy of graphlib from cpython v3.11.2

### DIFF
--- a/Lib/graphlib.py
+++ b/Lib/graphlib.py
@@ -1,0 +1,250 @@
+from types import GenericAlias
+
+__all__ = ["TopologicalSorter", "CycleError"]
+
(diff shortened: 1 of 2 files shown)


DimitrisJim:
Thanks! Could you change your commit message to mention the CPython version this was copied from? Also note that, if some tests fail, they just need to be annotated with:

```python
# TODO: RUSTPYTHON
@unittest.expectedFailure
```

if the test fails and:

```python
# TODO: RUSTPYTHON
@unittest.skip()
```

if the test brings down the interpreter.

youknowone (approved):
Looks good, Thank you!

PR #4536, opened 2023-02-21
less flaky syntax_async

youknowone:


DimitrisJim:
still some flakiness (#4551) here unfortunately.

PR #4542, opened 2023-02-22
Always wrap token streams in `SoftKeywordTransformer`

charliermarsh:
In Ruff, we often generate token streams outside of the parse step. Right now, that's causing failures for lint rules in certain cases that rely on the token stream rather than the AST. This PR always wraps the lexer in the soft keyword transformer. The primary downside is that we now have to thread through the `Mode` to all lexer calls.


charliermarsh:
Maybe it's better though to have `make_tokenizer` return this.

youknowone (approved):

PR #4543, opened 2023-02-22
Flatten parser interface

youknowone:
Externally:
- parser api is flattened except for `ast` and `lexer`. anything related to parsing will be directly accessible right under crate. `use rustpython_parser as parser` can replace `use rustpython_parser::parser` in most of cases.
- renamed `make_tokenizer` to `lex`. tokenizer sometimes can means a little bit different thing form lexer. but in our case, it returns lexed result instead of either string token, tokenizer or even a lexer - it returns the lexed result.

Internally:
- broke down error sub module. error submodule is expected to hold common errors of the super module. but it was containing 3 very independent errors fitting better with their own owners.

compiler/core is previously worked on in same motivation: https://github.com/RustPython/RustPython/blob/main/compiler/core/src/lib.rs

@charliermarsh could you review the changes?


DimitrisJim:
similarly, `codegen` currently doesn't use a flat structure.

I personally think `rustpython_parser::parser::parse` was _too_ verbose and prefer a flat structure. Lets indeed see what Charlie thinks since ruff is the biggest consumer of this API.

compiler/parser/src/lib.rs
```diff
@@ -124,14 +121,21 @@
 extern crate log;
 pub use rustpython_ast as ast;
 
-pub mod error;
```
DimitrisJim, line 127: this used to hold the `FStringErrorType` which does make sense to export now.

compiler/parser/src/parser.rs
```diff
@@ -189,12 +187,129 @@ pub fn parse_tokens(
...
     let marker_token = (Default::default(), mode.to_marker(), Default::default());
-    let tokenizer = iter::once(Ok(marker_token))
+    let lexer = iter::once(Ok(marker_token))
         .chain(lxr)
         .filter_ok(|(_, tok, _)| !matches!(tok, Tok::Comment { .. } | Tok::NonLogicalNewline));
     python::TopParser::new()
-        .parse(tokenizer.into_iter())
-        .map_err(|e| crate::error::parse_error_from_lalrpop(e, source_path))
+        .parse(lexer.into_iter())
+        .map_err(|e| crate::parser::parse_error_from_lalrpop(e, source_path))
```
DimitrisJim, line 195:
```suggestion
        .map_err(|e| parse_error_from_lalrpop(e, source_path))
```

compiler/parser/src/parser.rs
```diff
@@ -189,12 +187,129 @@ pub fn parse_tokens(
...
+}
+
+/// Represents represent errors that occur during parsing and are
+/// returned by the `parse_*` functions in the [parser] module.
+///
+/// [parser]: crate::parser
```
DimitrisJim, line 201: this is private now so it should be removed, i.e:

```suggestion
/// Represents represent errors that occur during parsing and are
/// returned by the `parse_*` functions.
```

there's a couple of others similar to this, `cargo doc --no-deps -p rustpython-parser` should show them so you can remove them.

youknowone:
Thanks! I resolved them.

DimitrisJim (approved):

charliermarsh:
This seems reasonable to me. So `use rustpython_parser::lexer` and `ast` are unchanged, but `use rustpython_parser::parser` is flattened out?

youknowone:
Yes, lexer and ast will remain. But other things are flattened to the root. Please feel free to suggest any enhancement you think.

PR #4552, opened 2023-02-22
Limit match range to end of last statement

charliermarsh:
Similar to other compound statements, this PR ends the match statement at the end of the last statement within the body.


charliermarsh:
Fixes this issue in Ruff: https://github.com/charliermarsh/ruff/issues/3126

youknowone (approved):

youknowone (approved):

Issue #3850, opened 2022-07-08
locale support including _locale and time.tzname


youknowone:
- [x] `_locale` module #4558
- [ ] time.tzname

bonjune:
#4054 

If we are talking about holding states such as `tzname` (which is initialized at module load time), this can be related.

youknowone:
> The doc says tzname is determined at module load time. Does "module load time" mean those values to be initialized in the function below?

It may not be exactly, but you can regard it is.

youknowone:
And yes. #4054 is generally related to module attributes and hidden data like Info types in sys module.

JazzGlobal:
Noticed quite a few functions in [CPython Locale](https://github.com/python/cpython/blob/3.11/Modules/_localemodule.c) are not implemented in locale.rs. I don't have much experience with this type of work but would the general idea be to write a wrapper for say: [GetLocaleIInfoA](https://learn.microsoft.com/en-us/windows/win32/api/winnls/nf-winnls-getlocaleinfoa) since it is used throughout CPython's implementation of `getdefaultlocale` so it can be used in our implementation?

DimitrisJim:
You are on Windows right? I haven't looked into how CPython does it (am out atm) but if this is calling some os provided function then yes, we typically wrap it and call into it after converting the args.

The function to be wrapped should be in windows.rs?

JazzGlobal:
That's right, and gotcha. Looks like libc has bindings for `setlocale` but that's all im seeing

arihant2math:
`time.tzname` is now supported on all platforms

youknowone:
Great, both of the list done

PR #3840, opened 2022-07-04
Add `_socket` constants from CPython 3.10

fanninpm:
The following constants may be missing from either rust's `libc` or the `winapi` crate (as of 4 July 2022):

- `AF_AAL5`
- `NETLINK_SKIP`
- `NETLINK_W1`
- `NETLINK_TCPDIAG`
- `NETLINK_ARPD`
- `NETLINK_ROUTE6`
- `NETLINK_TAPBASE`
- `AF_QIPCRTR`
- `BLUETOOTH_PROTO_L2CAP`
- `BLUETOOTH_PROTO_HCI`
- `SOL_HCI_RAW`
- `SO_HCI_RAW_FILTER`
- `HCI_TIME_STAMP`
- `HCI_DATA_DIR`
- `BLUETOOTH_PROTO_RFCOMM`
- `BTPROTO_SCO`
- `PACKET_HOST`
- `PACKET_BROADCAST`
- `PACKET_MULTICAST`
- `PACKET_OTHERHOST`
- `PACKET_OUTGOING`
- `PACKET_LOOPBACK`
- `PACKET_FASTROUTE`
- `TIPC_ADDR_NAMESEQ`
- `TIPC_ADDR_NAME`
- `TIPC_ADDR_ID`
- `TIPC_ZONE_SCOPE`
- `TIPC_CLUSTER_SCOPE`
- `TIPC_NODE_SCOPE`
- `TIPC_IMPORTANCE`
- `TIPC_SRC_DROPPABLE`
- `TIPC_DEST_DROPPABLE`
- `TIPC_CONN_TIMEOUT`
- `TIPC_LOW_IMPORTANCE`
- `TIPC_MEDIUM_IMPORTANCE`
- `TIPC_HIGH_IMPORTANCE`
- `TIPC_CRITICAL_IMPORTANCE`
- `TIPC_SUB_PORTS`
- `TIPC_SUB_SERVICE`
- `TIPC_SUB_CANCEL`
- `TIPC_WAIT_FOREVER`
- `TIPC_PUBLISHED`
- `TIPC_WITHDRAWN`
- `TIPC_SUBSCR_TIMEOUT`
- `TIPC_CFG_SRV`
- `TIPC_TOP_SRV`
- `ALG_SET_PUBKEY`
- `ALG_OP_SIGN`
- `ALG_OP_VERIFY`
- `MSG_BTAG`
- `MSG_ETAG`
- `TX_SETUP`
- `TX_DELETE`
- `TX_READ`
[...]

PR #4558, opened 2023-02-24
Add localeconv function to locale module

minhrongcon2000:
This PR is to add localeconv function to locale module.

#3850

### DIFF
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -6,7 +6,7 @@ import locale
 import sys
 import codecs
 
-@unittest.SkipTest(reason="Library not completed")
+@unittest.SkipTest(reason="TODO: RUSTPYTHON, Library not completed")
 class BaseLocalizedTest(unittest.TestCase):
     #
     # Base class for tests using a real locale
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -565,7 +565,6 @@ class TestMiscellaneous(unittest.TestCase):
             # If encoding non-empty, make sure it is valid
             codecs.lookup(enc)
 
-    @unittest.skip("TODO: RustPython strcoll has not been implemented")
     def test_strcoll_3303(self):
         # test crasher from bug #3303
         self.assertRaises(TypeError, locale.strcoll, "a", None)
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -506,8 +506,6 @@ class NormalizeTest(unittest.TestCase):
 
 
 class TestMiscellaneous(unittest.TestCase):
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_defaults_UTF8(self):
         # Issue #18378: on (at least) macOS setting LC_CTYPE to "UTF-8" is
         # valid. Furthermore LC_CTYPE=UTF is used by the UTF-8 locale coercing
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -375,10 +375,22 @@ class TestEnUSCollation(BaseLocalizedTest, TestCollation):
             raise unittest.SkipTest('wcscoll/wcsxfrm have known bugs')
         BaseLocalizedTest.setUp(self)
 
+    @unittest.skipIf(sys.platform.startswith('aix'),
+                     'bpo-29972: broken test on AIX')
+    @unittest.skipIf(
+        is_emscripten or is_wasi,
+        "musl libc issue on Emscripten/WASI, bpo-46390"
+    )
     @unittest.skip("TODO: RUSTPYTHON")
     def test_strcoll_with_diacritic(self):
         self.assertLess(locale.strcoll('à', 'b'), 0)
 
+    @unittest.skipIf(sys.platform.startswith('aix'),
+                     'bpo-29972: broken test on AIX')
+    @unittest.skipIf(
+        is_emscripten or is_wasi,
+        "musl libc issue on Emscripten/WASI, bpo-46390"
+    )
     @unittest.skip("TODO: RUSTPYTHON")
     def test_strxfrm_with_diacritic(self):
         self.assertLess(locale.strxfrm('à'), locale.strxfrm('b'))
@@ -500,8 +512,6 @@ class TestMiscellaneous(unittest.TestCase):
         # Issue #18378: on (at least) macOS setting LC_CTYPE to "UTF-8" is
         # valid. Furthermore LC_CTYPE=UTF is used by the UTF-8 locale coercing
         # during interpreter startup (on macOS).
-        if hasattr(sys, 'getwindowsversion'):
-            self.skipTest("TODO: _locale has not been implemented for window")
             
         import _locale
         import os
@@ -555,6 +565,7 @@ class TestMiscellaneous(unittest.TestCase):
             # If encoding non-empty, make sure it is valid
             codecs.lookup(enc)
 
+    @unittest.skip("TODO: RustPython strcoll has not been implemented")
     def test_strcoll_3303(self):
         # test crasher from bug #3303
         self.assertRaises(TypeError, locale.strcoll, "a", None)
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -512,6 +512,9 @@ class TestMiscellaneous(unittest.TestCase):
         # Issue #18378: on (at least) macOS setting LC_CTYPE to "UTF-8" is
         # valid. Furthermore LC_CTYPE=UTF is used by the UTF-8 locale coercing
         # during interpreter startup (on macOS).
+        if hasattr(sys, 'getwindowsversion'):
+            self.skipTest("TODO: _locale has not been implemented for window")
+            
         import _locale
         import os
 
@@ -566,6 +569,9 @@ class TestMiscellaneous(unittest.TestCase):
         self.assertRaises(TypeError, locale.strcoll, b"a", None)
 
     def test_setlocale_category(self):
+        if hasattr(sys, "getwindowsversion"):
+            self.skipTest("TODO: locale has not been implemented for window")
+        
         locale.setlocale(locale.LC_ALL)
         locale.setlocale(locale.LC_TIME)
         locale.setlocale(locale.LC_CTYPE)
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -375,21 +375,23 @@ class TestEnUSCollation(BaseLocalizedTest, TestCollation):
             raise unittest.SkipTest('wcscoll/wcsxfrm have known bugs')
         BaseLocalizedTest.setUp(self)
 
-    @unittest.skipIf(sys.platform.startswith('aix'),
-                     'bpo-29972: broken test on AIX')
-    @unittest.skipIf(
-        is_emscripten or is_wasi,
-        "musl libc issue on Emscripten/WASI, bpo-46390"
-    )
+    # @unittest.skipIf(sys.platform.startswith('aix'),
+    #                  'bpo-29972: broken test on AIX')
+    # @unittest.skipIf(
+    #     is_emscripten or is_wasi,
+    #     "musl libc issue on Emscripten/WASI, bpo-46390"
+    # )
+    @unittest.SkipTest("TODO: strcoll hasn't been implemented") 
     def test_strcoll_with_diacritic(self):
         self.assertLess(locale.strcoll('à', 'b'), 0)
 
-    @unittest.skipIf(sys.platform.startswith('aix'),
-                     'bpo-29972: broken test on AIX')
-    @unittest.skipIf(
-        is_emscripten or is_wasi,
-        "musl libc issue on Emscripten/WASI, bpo-46390"
-    )
+    # @unittest.skipIf(sys.platform.startswith('aix'),
+    #                  'bpo-29972: broken test on AIX')
+    # @unittest.skipIf(
+    #     is_emscripten or is_wasi,
+    #     "musl libc issue on Emscripten/WASI, bpo-46390"
+    # )
+    @unittest.SkipTest("TODO: strcoll hasn't been implemented")
     def test_strxfrm_with_diacritic(self):
         self.assertLess(locale.strxfrm('à'), locale.strxfrm('b'))
 
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -563,8 +563,6 @@ class TestMiscellaneous(unittest.TestCase):
         self.assertRaises(TypeError, locale.strcoll, "a", None)
         self.assertRaises(TypeError, locale.strcoll, b"a", None)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_setlocale_category(self):
         locale.setlocale(locale.LC_ALL)
         locale.setlocale(locale.LC_TIME)
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -381,8 +381,8 @@ class TestEnUSCollation(BaseLocalizedTest, TestCollation):
     #     is_emscripten or is_wasi,
     #     "musl libc issue on Emscripten/WASI, bpo-46390"
     # )
-    @unittest.SkipTest("TODO: strcoll hasn't been implemented") 
     def test_strcoll_with_diacritic(self):
+        self.skipTest("TODO: strcoll hasn't been implemented")
         self.assertLess(locale.strcoll('à', 'b'), 0)
 
     # @unittest.skipIf(sys.platform.startswith('aix'),
@@ -391,8 +391,8 @@ class TestEnUSCollation(BaseLocalizedTest, TestCollation):
     #     is_emscripten or is_wasi,
     #     "musl libc issue on Emscripten/WASI, bpo-46390"
     # )
-    @unittest.SkipTest("TODO: strcoll hasn't been implemented")
     def test_strxfrm_with_diacritic(self):
+        self.skipTest("TODO: strxfrm hasn't been implemented")
         self.assertLess(locale.strxfrm('à'), locale.strxfrm('b'))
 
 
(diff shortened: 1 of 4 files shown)


stdlib/src/locale.rs
```diff
@@ -0,0 +1,117 @@
...
+            ptr = ptr.offset(1);
+        }
+        vm.ctx.new_list(group_vec)
+    }
+
+    unsafe fn _parse_ptr_to_str(vm: &VirtualMachine, raw_ptr: *mut c_char) -> PyResult {
+        let slice = unsafe { CStr::from_ptr(raw_ptr) };
+        let cstr = slice
+            .to_str()
+            .map(|s| s.to_owned())
```
youknowone, line 47:
```suggestion
```
This is unnecessary for `vm.new_pyobj`

stdlib/src/locale.rs
```diff
@@ -0,0 +1,117 @@
...
+            .map_err(|e| vm.new_unicode_decode_error(format!("unable to decode: {e}")))?;
```
youknowone, line 48:
```suggestion
            .expect("localeconv always return decodable string");
```

stdlib/src/locale.rs
```diff
@@ -0,0 +1,117 @@
...
+
+        Ok(vm.new_pyobj(cstr))
+    }
+
+    #[pyfunction]
+    fn localeconv(vm: &VirtualMachine) -> PyResult<PyDictRef> {
+        let result = vm.ctx.new_dict();
+        macro_rules! set_string_field {
+            ($field:expr) => {{
+                result.set_item(stringify!($field), _parse_ptr_to_str(vm, $field)?, vm)?
```
youknowone, line 58:
```suggestion
                result.set_item(stringify!($field), _parse_ptr_to_str(vm, (*lc).$field)?, vm)?
```
Then you can skip 
```
let n_sign_posn = (*lc).n_sign_posn;
```

youknowone (changes requested):
Thanks! I left a few comments about styles.

stdlib/src/locale.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[pyattr(name = "CHAR_MAX")]
+    fn char_max(vm: &VirtualMachine) -> PyIntRef {
+        vm.ctx.new_int(libc::c_char::MAX)
+    }
+
+    unsafe fn copy_grouping(group: *mut c_char, vm: &VirtualMachine) -> PyListRef {
+        let mut group_vec: Vec<PyObjectRef> = Vec::new();
+        let mut ptr = group;
+
+        while ![0 as i8, c_char::MAX].contains(&*ptr) {
```
fanninpm, line 34: @youknowone Clippy wants this:

```suggestion
        while ![0_i8, c_char::MAX].contains(&*ptr) {
```

Lib/test/test_locale.py
```diff
@@ -6,7 +6,7 @@
 import sys
 import codecs
 
-
+@unittest.SkipTest(reason="Library not completed")
```
fanninpm, line 9:
```suggestion
@unittest.SkipTest("TODO: RUSTPYTHON, Library not completed")
```

stdlib/src/locale.rs
```diff
@@ -0,0 +1,101 @@
...
+    #[pyattr]
+    use libc::{
+        ABDAY_1, ABDAY_2, ABDAY_3, ABDAY_4, ABDAY_5, ABDAY_6, ABDAY_7, ABMON_1, ABMON_10, ABMON_11,
+        ABMON_12, ABMON_2, ABMON_3, ABMON_4, ABMON_5, ABMON_6, ABMON_7, ABMON_8, ABMON_9,
+        ALT_DIGITS, AM_STR, CODESET, CRNCYSTR, DAY_1, DAY_2, DAY_3, DAY_4, DAY_5, DAY_6, DAY_7,
+        D_FMT, D_T_FMT, ERA, ERA_D_FMT, ERA_D_T_FMT, ERA_T_FMT, LC_ALL, LC_COLLATE, LC_CTYPE,
+        LC_MESSAGES, LC_MONETARY, LC_NUMERIC, LC_TIME, MON_1, MON_10, MON_11, MON_12, MON_2, MON_3,
+        MON_4, MON_5, MON_6, MON_7, MON_8, MON_9, NOEXPR, PM_STR, RADIXCHAR, THOUSEP, T_FMT,
+        T_FMT_AMPM, YESEXPR,
+    };
```
fanninpm, line 21: Not all of these constants are available on every platform. When I encountered this sort of problem in the past, I looked up each item individually in the `libc` repository to see which platforms support that item. (See #3840, for example.)

I realized back then (and I still maintain) that this process would prove to be tedious and error-prone. Perhaps I should look into using the `autocfg` crate instead, but that's ultimately beyond the scope of this PR.

minhrongcon2000: I tried with Window locale.h this afternoon and found out that Window API uses different naming convention (prefix with W, if I remember). Hence, if we would like to fix it, I'm afraid that we need to implement our own locale interaction. I will do it in another PR once this module takes shape.

Lib/test/test_locale.py
youknowone: I think we cannot merge this PR as this way. because `setlocale` is a mendatory part of _locale.
If you want to work on setlocale on different PR, could you change `Lib/locale.py` from
```python
try:
    from _locale import *
except ImportError:
```
to only cherry-picking `_locale.localeconv` and not skipping the python fallback part?

minhrongcon2000: @youknowone I just commit the implementation for setlocale function! Should I allow the test for locale module after this commit?

youknowone: Sure, that will be perfect

stdlib/src/lib.rs
```diff
@@ -159,5 +160,9 @@ pub fn get_module_inits() -> impl Iterator<Item = (Cow<'static, str>, StdlibInit
         {
             "_uuid" => uuid::make_module,
         }
+        #[cfg(all(unix, not(any(target_os = "ios", target_os = "android"))))]
```
youknowone, line 163: :+1:

stdlib/src/locale.rs
```diff
@@ -0,0 +1,154 @@
...
+        unsafe {
+            let result = match args.locale {
+                OptionalArg::Missing => {
+                    let null_ptr: *const i8 = ptr::null();
+                    libc::setlocale(args.category, null_ptr)
+                }
+                OptionalArg::Present(locale) => match locale {
+                    None => {
+                        let null_ptr: *const i8 = ptr::null();
+                        libc::setlocale(args.category, null_ptr)
```
youknowone, line 135:
```suggestion
                        libc::setlocale(args.category, ptr::null())
```

stdlib/src/locale.rs
```diff
@@ -0,0 +1,154 @@
...
+    fn setlocale(args: LocaleArgs, vm: &VirtualMachine) -> PyResult {
+        unsafe {
+            let result = match args.locale {
+                OptionalArg::Missing => {
+                    let null_ptr: *const i8 = ptr::null();
+                    libc::setlocale(args.category, null_ptr)
+                }
+                OptionalArg::Present(locale) => match locale {
+                    None => {
+                        let null_ptr: *const i8 = ptr::null();
```
youknowone, line 134: we don't know if this is i8 or u8. we have to use `libc::c_char` to avoid those problem.
But i'd like remove this line.

stdlib/src/locale.rs
```diff
@@ -0,0 +1,154 @@
...
+                        libc::setlocale(args.category, null_ptr)
+                    }
+                    Some(l) => {
+                        let mut l_str = l.to_string();
+                        if l_str.is_empty() {
+                            l_str.push('\0');
+                        }
+                        let l_ptr = l_str.as_ptr() as *const i8;
```
youknowone, line 142: we prefer to use `std::ffi::CString::new(l)` and `CStr::as_ptr` to obtain the pointer.

stdlib/src/locale.rs
```diff
@@ -0,0 +1,154 @@
...
+        #[pyarg(any)]
+        category: i32,
+        #[pyarg(any, optional)]
+        locale: OptionalArg<Option<PyStrRef>>,
+    }
+
+    #[pyfunction]
+    fn setlocale(args: LocaleArgs, vm: &VirtualMachine) -> PyResult {
+        unsafe {
+            let result = match args.locale {
```
youknowone, line 127: this is handling both missing argument and `None` given same way, right?
using `flatten()` will merge two case and makes match simpler.
```suggestion
            let result = match args.locale.flatten() {
```

youknowone (changes requested):

stdlib/src/locale.rs
```diff
@@ -0,0 +1,154 @@
...
+            ptr = ptr.offset(1);
+        }
+        // https://github.com/python/cpython/blob/677320348728ce058fa3579017e985af74a236d4/Modules/_localemodule.c#L80
+        if !group_vec.is_empty() {
+            group_vec.push(vm.ctx.new_int(0i32).into());
+        }
+        vm.ctx.new_list(group_vec)
+    }
+
+    unsafe fn _parse_ptr_to_str(vm: &VirtualMachine, raw_ptr: *mut libc::c_char) -> PyResult {
```
youknowone, line 47:
```suggestion
    unsafe fn _parse_ptr_to_str(vm: &VirtualMachine, raw_ptr: *const libc::c_char) -> PyResult {
```
This function seems using `raw_ptr` only as `*const _`

youknowone:
`test_setlocale_category` is also successful!

youknowone:
The windows failure is fine. But test_type failure on linux is... what's happened

minhrongcon2000:
https://github.com/RustPython/RustPython/actions/runs/4271112901/jobs/7435359720#step:12:12102
@youknowone It seems like the `format()` function is not implemented with locale, so it returns an unformatted string, resulting in failed test case...

minhrongcon2000:
@youknowone it turns out the locale code is inconsistent between platform, so after some fixes, it's normal now!

Lib/test/test_locale.py
```diff
@@ -382,7 +382,7 @@ def setUp(self):
         "musl libc issue on Emscripten/WASI, bpo-46390"
     )
     def test_strcoll_with_diacritic(self):
-        self.assertLess(locale.strcoll('à', 'b'), 0)
+        self.assertGreater(locale.strcoll('à', 'b'), 0) # tested on CPython
```
youknowone, line 385: we normally don't edit CPython originated code without sufficient reason.
If it fails, we usually prefer to keep it as `@unittest.expectedFailure`

Lib/test/test_types.py
```diff
@@ -403,13 +403,15 @@ def test_float__format__locale(self):
...
-    @run_with_locale('LC_NUMERIC', 'en_US.UTF8')
+    @run_with_locale('LC_NUMERIC', 'en_US.UTF-8')
     def test_int__format__locale(self):
         # test locale support for __format__ code 'n' for integers
 
-        x = 123456789012345678901234567890
-        for i in range(0, 30):
-            self.assertEqual(locale.format_string('%d', x, grouping=True), format(x, 'n'))
+        # for some reason, setting x to be 123456789012345678901234567890
+        # cause vector overflow, so I set it to small number.
```
youknowone, line 411: Do you want to try debugger?

minhrongcon2000: Sure :D

Lib/test/test_locale.py
```diff
@@ -375,22 +375,24 @@ def setUp(self):
...
-    @unittest.skipIf(
-        is_emscripten or is_wasi,
-        "musl libc issue on Emscripten/WASI, bpo-46390"
-    )
+    # @unittest.skipIf(sys.platform.startswith('aix'),
+    #                  'bpo-29972: broken test on AIX')
+    # @unittest.skipIf(
+    #     is_emscripten or is_wasi,
+    #     "musl libc issue on Emscripten/WASI, bpo-46390"
+    # )
```
youknowone, line 383: rather than commenting,
We usually add `@unittest.skip("TODO: RUSTPYTHON")` to make updating tests easier in future

Lib/test/test_locale.py
```diff
@@ -565,9 +568,10 @@ def test_strcoll_3303(self):
         self.assertRaises(TypeError, locale.strcoll, "a", None)
         self.assertRaises(TypeError, locale.strcoll, b"a", None)
```
youknowone, line 569: Instead of editing tests inside, we usually do:
```rust
    # TODO: RUSTPYTHON
    if sys.platform == "win32":
        test_defaults_UTF8 = unittest.expectedFailure(test_defaults_UTF8)
```
Searching for `unittest.expectedFailure(` will show examples

youknowone (changes requested):

Lib/test/test_locale.py
```diff
@@ -375,21 +375,11 @@ def setUp(self):
             raise unittest.SkipTest('wcscoll/wcsxfrm have known bugs')
         BaseLocalizedTest.setUp(self)
 
-    @unittest.skipIf(sys.platform.startswith('aix'),
-                     'bpo-29972: broken test on AIX')
-    @unittest.skipIf(
-        is_emscripten or is_wasi,
-        "musl libc issue on Emscripten/WASI, bpo-46390"
-    )
```
youknowone, line 383: please do not remove the original code but only add
`@unittest.skip()`

Lib/test/test_locale.py
```diff
@@ -375,21 +375,11 @@ def setUp(self):
...
+    @unittest.skip("TODO: RUSTPYTHON")
     def test_strcoll_with_diacritic(self):
         self.assertLess(locale.strcoll('à', 'b'), 0)
 
-    @unittest.skipIf(sys.platform.startswith('aix'),
-                     'bpo-29972: broken test on AIX')
-    @unittest.skipIf(
-        is_emscripten or is_wasi,
-        "musl libc issue on Emscripten/WASI, bpo-46390"
-    )
```
youknowone, line 392: here too

Lib/test/test_locale.py
```diff
@@ -506,12 +496,13 @@ def test_japanese(self):
...
 class TestMiscellaneous(unittest.TestCase):
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_defaults_UTF8(self):
         # Issue #18378: on (at least) macOS setting LC_CTYPE to "UTF-8" is
         # valid. Furthermore LC_CTYPE=UTF is used by the UTF-8 locale coercing
         # during interpreter startup (on macOS).
+        if hasattr(sys, 'getwindowsversion'):
+            self.skipTest("TODO: _locale has not been implemented for window")
+
```
youknowone, line 505: due to later `expectedFailure`, we don't need this line.

Lib/test/test_locale.py
```diff
@@ -577,6 +570,10 @@ def test_setlocale_category(self):
 
         # crasher from bug #7419
         self.assertRaises(locale.Error, locale.setlocale, 12345)
+        
+    # TODO: RUSTPYTHON
+    if sys.platform == "win32":
+        test_setlocale_category = unittest.expectedFailure(test_setlocale_category)
```
youknowone, line 576: 👍

Lib/test/test_types.py
```diff
@@ -405,6 +405,7 @@ def test_float__format__locale(self):
 
     @run_with_locale('LC_NUMERIC', 'en_US.UTF8')
     def test_int__format__locale(self):
+        self.skipTest("TODO: format with type 'n' is not integrated with locale")
```
youknowone, line 408: instead of  skip here, skip on 405

youknowone (changes requested):

Lib/test/test_locale.py
```diff
@@ -555,6 +565,7 @@ def test_getpreferredencoding(self):
             # If encoding non-empty, make sure it is valid
             codecs.lookup(enc)
 
+    @unittest.skip("TODO: RustPython strcoll has not been implemented")
```
fanninpm, line 568: Does this work?

```suggestion
    # TODO: RUSTPYTHON, strcoll has not been implemented
    @unittest.expectedFailure
```

minhrongcon2000: Sorry! I accidentally skip it since I haven't implemented the strcoll but forgot that it will raise error nonetheless :((

fanninpm: That particular test is expecting to raise a `TypeError` there. If it doesn't, it fails.

Lib/test/test_locale.py
```diff
@@ -381,6 +381,7 @@ def setUp(self):
         is_emscripten or is_wasi,
         "musl libc issue on Emscripten/WASI, bpo-46390"
     )
+    @unittest.skip("TODO: RUSTPYTHON")
```
fanninpm, line 384:
```suggestion
    # TODO: RUSTPYTHON, strcoll has not been implemented
    @unittest.expectedFailure
```

Lib/test/test_locale.py
```diff
@@ -390,6 +391,7 @@ def test_strcoll_with_diacritic(self):
...
```
fanninpm: Does this work?
```suggestion
    # TODO: RUSTPYTHON
    @unittest.expectedFailure
```

minhrongcon2000: I fixed in https://github.com/RustPython/RustPython/pull/4558/commits/9e5be11c43806d2a4faa42a9348b47cbaa2d03c6

stdlib/src/locale.rs
```diff
@@ -0,0 +1,145 @@
...
+        locale: OptionalArg<Option<PyStrRef>>,
+    }
+
+    #[pyfunction]
+    fn setlocale(args: LocaleArgs, vm: &VirtualMachine) -> PyResult {
+        unsafe {
+            let result = match args.locale.flatten() {
+                None => libc::setlocale(args.category, ptr::null()),
+                Some(l) => {
+                    let l_str = CString::new(l.to_string()).expect("expect to be always converted");
```
youknowone, line 133: this is not always successful
```
>>> _locale.setlocale(0, '\0')
ValueError: embedded null character
```

stdlib/src/locale.rs
```diff
@@ -0,0 +1,145 @@
...
+                    let l_ptr = CStr::as_ptr(&l_str);
```
youknowone, line 134: because CString implements `std::ops::Deref<Target=CStr>`, this is redundant.

```
libc::setlocale(args.category, lstr.as_ptr())
```
works exactly same way.

youknowone (approved):
Thank you for hard working for this tricky issue. It will be a great base of the future locale works.

Maintaining note: Though a few tests regressed, I think this is the way to go forward with true locale support.
squash is desired when merging.

PR #4562, opened 2023-02-24
double windows sleep for syntax_async until it success

youknowone:
this is very annoying...


DimitrisJim (approved):
let's hope this finally fixed the issue :smile:

PR #4568, opened 2023-02-25
add sys.flags and sys.int_info

youknowone:


youknowone:
Because this patch temporary break changes, lets merge it via #4569.
Seperation is possible, but not worth to

PR #4570, opened 2023-02-25
Update __hello__  from cpython

youknowone:


youknowone:
dropped phello because it fixes too many test_importlib tests which are already fixed in test_import PR

Issue #4566, opened 2023-02-24
Some valid named unicode characters cause syntax errors


charliermarsh:
`"\N{BACKSPACE}another cool trick"` yields a syntax error in RustPython, but runs without error on CPython.

See: https://github.com/charliermarsh/ruff/issues/3207.

charliermarsh:
For reference, Python does throw errors on obviously-invalid named unicode characters like `"\N{AAA} another cool trick"`.

DimitrisJim:
Seems like this is coming from `unicode_names2`. `unicode_names2::character("BACKSPACE")` doesn't return a character.

fanninpm:
Related question (out of curiosity): What does `unicode_names2::name('\x08')` return?

DimitrisJim:
It's also a `None`, it doesn't support any of the control characters (so `'\N{LINE FEED}'`, `\N{NUL}'` etc, all raise SyntaxErrors).

DimitrisJim:
After a bit of searching it appears it doesn't support [name aliases](http://www.unicode.org/Public/12.1.0/ucd/NameAliases.txt) (which all these control characters fall into). I personally don't know of any libs that support it but we could just grab that file, parse it and have `unicodedata` and the string parser fall back to it.

youknowone:
[unic-ucd-name_aliases](https://docs.rs/unic-ucd-name_aliases/) supports getting alias from the name but not the other way.

youknowone:
https://github.com/progval/unicode_names2/issues/11

youknowone:
unicode_names2 will take patch
https://github.com/progval/unicode_names2/issues/11#issuecomment-1445037320

PR #4572, opened 2023-02-25
Generate aliases for use by string parser.

DimitrisJim:
A temporary fix for #4566 so as to not hold back the parser. (This only covers the string parser for now, not `unicodedata`).


DimitrisJim:
I can open another issue to keep track of the permanent fix for this (and for unicodedata)

common/Cargo.toml
```diff
@@ -4,12 +4,17 @@ version = "0.2.0"
 description = "General python functions and algorithms for use in RustPython"
 authors = ["RustPython Team"]
 edition = "2021"
+build = "build.rs"
```
youknowone, line 7:
```suggestion
```
This is the default value

youknowone (commented):
is this doing the same thing?
https://github.com/RustPython/RustPython/commit/49adb5cd88bc5052f70dd362916940ada572dfb5

DimitrisJim:
> is this doing the same thing? [49adb5c](https://github.com/RustPython/RustPython/commit/49adb5cd88bc5052f70dd362916940ada572dfb5)

does that compile? that's basically what I was starting with but there's some characters that need to be escaped or else `phf` can't really generate the map correctly for them.

youknowone:
Yes, that does compile and working well at least for BACKSPACE. 
`include_str!` obtains the whole file content as a valid str. And I didn't put them in source code but handled only with string operations.

DimitrisJim:
> Yes, that does compile and working well at least for BACKSPACE.

Ah then push it through to here, I have no issue.

> I think something similar like this version can be filed as patch to unicode_names2

I still am not sure for this since they already have the data in [UnicodeData.txt](https://github.com/progval/unicode_names2/blob/master/data/UnicodeData.txt) and they use their own implementation of `phf`. But if you wanna close this PR and go with the patch to `unicode_names2` I'd be fine with that too.

youknowone:
wow, i didn't know their `phf` is their own's.

youknowone:
I tried to add alias, but I found unicode_names2 cannot regenerate master branch's output using the same version of generator 😂 
directly putting your version seems more realistic

DimitrisJim:
Close in favor of #4582

youknowone:
We still need the tests from here, right?

PR #4579, opened 2023-02-26
Satisfy safety conditions for std::slice::from_raw_parts. Fixes #4575.

### LINKED ISSUE #4575: array._array_reconstructor panics on an empty array with debug stdlib
I wanted to fix this myself, but I think it needs some TLC from @youknowone or similar to find the best fix.

To reproduce:

```
rustup override set nightly
cargo run -Zbuild-std --target=x86_64-unknown-linux-gnu -- -m test -v test_array
```

This uses a debug version of the standard library, which exposes the fact that array._array_reconstructor calls array::frombytes (array.rs line 1561) and array::frombytes calls as_ptr(). Unfortunately, as_ptr() does not give a useful answer when the array is empty, and in fact will provide an unaligned value. This then trips the safety condition in std::slice::from_raw_parts.

```
unsafe precondition(s) violated: slice::from_raw_parts requires the pointer to be aligned and non-null, and the total size of the slice not to exceed `isize::MAX`
```

There are probably many ways to fix this but I'd prefer someone who knows RustPython better than I do to pick the best way!

threeifbyair:
Thanks to @youknowone 's suggestion, I was able to fix the bug I reported. Thank you so much!

### DIFF
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -215,8 +215,10 @@ mod array {
                             // support are valid
                             let ptr = b.as_ptr() as *const $t;
                             let ptr_len = b.len() / std::mem::size_of::<$t>();
-                            let slice = unsafe { std::slice::from_raw_parts(ptr, ptr_len) };
-                            v.extend_from_slice(slice);
+                            if ptr_len > 0 {
+                                let slice = unsafe { std::slice::from_raw_parts(ptr, ptr_len) };
+                                v.extend_from_slice(slice);
+                            }
                         })*
                     }
                 }
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -213,9 +213,9 @@ mod array {
                         $(ArrayContentType::$n(v) => {
                             // safe because every configuration of bytes for the types we
                             // support are valid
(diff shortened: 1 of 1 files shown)


stdlib/src/array.rs
```diff
@@ -215,8 +215,10 @@ mod array {
                             // support are valid
                             let ptr = b.as_ptr() as *const $t;
                             let ptr_len = b.len() / std::mem::size_of::<$t>();
-                            let slice = unsafe { std::slice::from_raw_parts(ptr, ptr_len) };
-                            v.extend_from_slice(slice);
+                            if ptr_len > 0 {
```
youknowone, line 218:
```suggestion
                                if b.len() > 0 {
                                    let ptr = b.as_ptr() as *const $t;
                                    let ptr_len = b.len() / std::mem::size_of::<$t>();
                                    let slice = unsafe { std::slice::from_raw_parts(ptr, ptr_len) };
```

This is a little bit weird though. Because extend_from_slice have not to copy anything from the slice if its length is 0.

youknowone (changes requested):

youknowone (approved):
Thank you for contributing!

PR #4581, opened 2023-02-26
Fix py_freeze! to run library with __main__ without __init__

youknowone:
This fix enables wasi runs `-m test`


coolreader18:
It seems like CPython just has a dummy file as `Lib/test/__init__.py`? It seems like `__init__`-less packages are deprecated, or at least the code enabling them is marked as a backwards-compatibility hack

youknowone:
oh, i missed that. Thank you!

PR #4582, opened 2023-02-26
Another temporary fix for #4566 

### LINKED ISSUE #4566: Some valid named unicode characters cause syntax errors

youknowone:
see also #4572

we can temporarily go with this because unicode_names2 will work with this interface in future.

closes #4566

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2933,9 +2933,11 @@ checksum = "f962df74c8c05a667b5ee8bcf162993134c104e96440b663c8daa176dc772d8c"
 
 [[package]]
 name = "unicode_names2"
-version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


DimitrisJim (approved):

youknowone:
Oh, no. The UnicodeData.txt does include aliases but not abbrs.

DimitrisJim:
ah, that's true, I hadn't noticed that either tbh (and really annoying unfortunately since you now need to handle both files.)

DimitrisJim:
does your branch of unicode_names2 currently handle the abbreviations? If so, lets merge it.

youknowone:
Yes, finally.

PR #4584, opened 2023-02-27
Use proper locations for sub-expressions in constant matches

charliermarsh:
This was an oversight in the initial implementation: I used the `MatchValue` location for _all_ sub-expressions (so, e.g., if you had `case -0j`, we used the range of `-0j` for the sub-expressions `0j`, which is incorrect). Now, we use the correct ranges for every sub-expression in the match.

### DIFF
--- a/compiler/parser/python.lalrpop
+++ b/compiler/parser/python.lalrpop
@@ -576,6 +576,41 @@ StarPattern: ast::PatternKind = {
     },
 }
 
+ConstantAtom: ast::Expr = {
+    <location:@L> <value:Constant> <end_location:@R> => ast::Expr {
+        location,
+        end_location: Some(end_location),
+        custom: (),
+        node: ast::ExprKind::Constant { value, kind: None }
+    },
+}
+
+ConstantExpr: ast::Expr = {
+    ConstantAtom,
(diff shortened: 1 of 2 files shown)


compiler/parser/python.lalrpop
```diff
@@ -740,76 +708,7 @@ MappingKey: ast::Expr = {
...
-            }),
-            op,
-            right: Box::new(ast::Expr {
-                location,
-                end_location: Some(end_location),
-                custom: (),
-                node: ast::ExprKind::Constant { value: right, kind: None }
-            }),
-        }
-    },
```
charliermarsh, line 810: (I was also able to DRY this up quite a bit.)

compiler/parser/src/snapshots/rustpython_parser__parser__tests__patma.snap
```diff
@@ -65,7 +65,7 @@ expression: parse_ast
                                     operand: Located {
                                         location: Location {
                                             row: 5,
-                                            column: 9,
+                                            column: 10,
```
charliermarsh, line 68: GitHub hides the diff here but it's pretty minimal -- just column adjustments.

youknowone (approved):

PR #4597, opened 2023-02-28
Add unittset.expectedFailureIf for RustPython 

youknowone:
We really need conditional expectedFailure in the name of human right and pain killer.


fanninpm (approved):
We should take care to remove these once all of their usages are gone.

PR #4602, opened 2023-03-01
Update test_syntax from CPython 3.11.0

DimitrisJim:
And also allow the doctests to run for checking the edgecases.

### DIFF
--- a/Lib/test/test_syntax.py
+++ b/Lib/test/test_syntax.py
@@ -555,7 +555,8 @@ SyntaxError: expected default value expression
 Traceback (most recent call last):
 SyntaxError: expected default value expression
 
->>> import ast; ast.parse('''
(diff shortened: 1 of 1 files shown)


Lib/test/test_syntax.py
```diff
@@ -351,7 +351,212 @@
...
+>>> lambda a=1,d=,c: None
+Traceback (most recent call last):
+SyntaxError: expected default value expression
+
+>>> lambda a,d=,c: None
+Traceback (most recent call last):
+SyntaxError: expected default value expression
+
+# TODO: RUSTPYTHON NameError: name 'PyCF_TYPE_COMMENTS' is not defined
+>>> import ast; ast.parse('''  # doctest: +SKIP
```
youknowone, line 559: oh, i didn't know this is possible

youknowone (approved):

PR #4603, opened 2023-03-01
Add strxfrm and strcoll to locale mod

minhrongcon2000:
This PR is made to add `strcoll` and `strxfrm` function to `locale` module

### DIFF
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -342,9 +342,6 @@ class TestFrFRNumberFormatting(FrFRCookedTest, BaseFormattingTest):
 
 class TestCollation(unittest.TestCase):
     # Test string collation functions
-
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_strcoll(self):
         self.assertLess(locale.strcoll('a', 'b'), 0)
         self.assertEqual(locale.strcoll('a', 'a'), 0)
@@ -353,8 +350,7 @@ class TestCollation(unittest.TestCase):
         self.assertRaises(ValueError, locale.strcoll, 'a\0', 'a')
         self.assertRaises(ValueError, locale.strcoll, 'a', 'a\0')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+
     def test_strxfrm(self):
         self.assertLess(locale.strxfrm('a'), locale.strxfrm('b'))
         # embedded null character
@@ -381,8 +377,6 @@ class TestEnUSCollation(BaseLocalizedTest, TestCollation):
         is_emscripten or is_wasi,
         "musl libc issue on Emscripten/WASI, bpo-46390"
     )
-    # TODO: RUSTPYTHON", strcoll has not been implemented
-    @unittest.expectedFailure
     def test_strcoll_with_diacritic(self):
         self.assertLess(locale.strcoll('à', 'b'), 0)
 
@@ -392,8 +386,6 @@ class TestEnUSCollation(BaseLocalizedTest, TestCollation):
         is_emscripten or is_wasi,
         "musl libc issue on Emscripten/WASI, bpo-46390"
     )
-    # TODO: RUSTPYTHON", strxfrm has not been implemented
-    @unittest.expectedFailure
     def test_strxfrm_with_diacritic(self):
         self.assertLess(locale.strxfrm('à'), locale.strxfrm('b'))
 
--- a/Lib/test/test_locale.py
+++ b/Lib/test/test_locale.py
@@ -342,6 +342,7 @@ class TestFrFRNumberFormatting(FrFRCookedTest, BaseFormattingTest):
 
 class TestCollation(unittest.TestCase):
     # Test string collation functions
+    @unittest.expectedFailureIfWindows("TODO: RUSTPYTHON")
     def test_strcoll(self):
         self.assertLess(locale.strcoll('a', 'b'), 0)
         self.assertEqual(locale.strcoll('a', 'a'), 0)
@@ -349,17 +350,12 @@ class TestCollation(unittest.TestCase):
         # embedded null character
         self.assertRaises(ValueError, locale.strcoll, 'a\0', 'a')
         self.assertRaises(ValueError, locale.strcoll, 'a', 'a\0')
-    
-    if sys.platform == "win32":
-        test_strcoll = unittest.expectedFailure(test_strcoll)
 
+    @unittest.expectedFailureIfWindows("TODO: RUSTPYTHON")
     def test_strxfrm(self):
         self.assertLess(locale.strxfrm('a'), locale.strxfrm('b'))
         # embedded null character
         self.assertRaises(ValueError, locale.strxfrm, 'a\0')
-        
-    if sys.platform == "win32":
-        test_strxfrm = unittest.expectedFailure(test_strxfrm)
 
 
 class TestEnUSCollation(BaseLocalizedTest, TestCollation):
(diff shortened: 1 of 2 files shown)


youknowone (changes requested):
Thank you! I left style a suggestion and concern about strxfrm implementation. Could you please check it?

stdlib/src/locale.rs
```diff
@@ -83,9 +81,8 @@ mod _locale {
         unsafe {
             let string_cstr = CString::new(string.as_str()).map_err(|e| e.to_pyexception(vm))?;
             let string_ptr = CStr::as_ptr(&string_cstr);
-            let empty_str = CString::new("").unwrap();
-            let result_ptr = CStr::as_ptr(&empty_str) as *mut c_char;
-            libc::strxfrm(result_ptr, string_ptr, libc::strlen(string_ptr) + 1);
+            let result_ptr = ptr::null_mut();
+            libc::strxfrm(result_ptr, string_ptr, string.char_len() + 1);
```
youknowone, line 85: the return value is length. you can create buffer with the given length and call strxfrm again.

minhrongcon2000: Wait! So strxfrm is called 2 times?

stdlib/src/locale.rs
```diff
@@ -67,6 +67,44 @@ mod _locale {
...
+            let string2_cstr = CString::new(string2.as_str()).map_err(|e| e.to_pyexception(vm))?;
+            Ok(vm.new_pyobj(libc::strcoll(string1_cstr.as_ptr(), string2_cstr.as_ptr())))
+        }
+    }
+
+    #[pyfunction]
+    fn strxfrm(string: PyStrRef, vm: &VirtualMachine) -> PyResult {
+        // https://github.com/python/cpython/blob/eaae563b6878aa050b4ad406b67728b6b066220e/Modules/_localemodule.c#L390-L442
+        unsafe {
+            let n1 = string.char_len() + 1;
```
youknowone, line 83: because we are using strxfrom not wcsxfrom, I guess this len is `string.len()` instead of char_len

stdlib/src/locale.rs
```diff
@@ -67,6 +67,44 @@ mod _locale {
...
+            let mut buff = vec![0; n1];
+            let buff_ptr = buff.as_mut_ptr();
+
+            let string_cstr = CString::new(string.as_str()).map_err(|e| e.to_pyexception(vm))?;
+            let string_ptr = CStr::as_ptr(&string_cstr);
+            let n2 = libc::strxfrm(buff_ptr, string_ptr, n1) + 1;
+
+            if n2 >= n1 {
```
youknowone, line 91:
```suggestion
            if n2 > n1 {
```
you added 1 before comparison unlike cpython version

stdlib/src/locale.rs
```diff
@@ -67,6 +67,44 @@ mod _locale {
...
+                let mut new_buff = vec![0; n2];
+                // this part corresponds to re-allocate in CPython
+                for (i, item) in buff.iter().enumerate() {
+                    if *item != 0 {
+                        new_buff[i] = *item;
+                    }
+                }
```
youknowone, line 98: I think we don't need to copy content

stdlib/src/locale.rs
```diff
@@ -67,6 +67,44 @@ mod _locale {
...
+            let n1 = string.char_len() + 1;
+            let mut buff = vec![0; n1];
+            let buff_ptr = buff.as_mut_ptr();
+
+            let string_cstr = CString::new(string.as_str()).map_err(|e| e.to_pyexception(vm))?;
+            let string_ptr = CStr::as_ptr(&string_cstr);
+            let n2 = libc::strxfrm(buff_ptr, string_ptr, n1) + 1;
+
+            if n2 >= n1 {
+                let mut new_buff = vec![0; n2];
```
youknowone, line 92: because we will not going to use buff anymore, this will be fine.

```suggestion
buff = vec![0; n2];
```

stdlib/src/locale.rs
```diff
@@ -67,6 +67,44 @@ mod _locale {
...
+                // this part corresponds to re-allocate in CPython
+                for (i, item) in buff.iter().enumerate() {
+                    if *item != 0 {
+                        new_buff[i] = *item;
+                    }
+                }
+
+                let new_buff_ptr = new_buff.as_mut_ptr();
+                libc::strxfrm(new_buff_ptr, string_ptr, n2);
```
youknowone, line 101: A trick to reuse the given length.
```suggestion
                let n2 = libc::strxfrm(buff.as_mut_ptr(), string_ptr, n2);
                buff.trancate(n2);
                vm.new_pyobj(String::from_utf8(buff).expect("strxfrm returned invalid utf-8 string"))
```

stdlib/src/locale.rs
```diff
@@ -67,6 +68,33 @@ mod _locale {
...
+            Ok(vm.new_pyobj(libc::strcoll(string1_cstr.as_ptr(), string2_cstr.as_ptr())))
+        }
+    }
+
+    #[pyfunction]
+    fn strxfrm(string: PyStrRef, vm: &VirtualMachine) -> PyResult {
+        // https://github.com/python/cpython/blob/eaae563b6878aa050b4ad406b67728b6b066220e/Modules/_localemodule.c#L390-L442
+        unsafe {
+            let n1 = string.byte_len() + 1;
+            let mut buff = vec![0; n1];
```
youknowone, line 85:
```suggestion
            let mut buff: Vec<u8> = vec![0; n1];
```
If you are going to use transmute, I prefer to set `Vec<u8>` here to avoid it.

stdlib/src/locale.rs
```diff
@@ -67,6 +68,33 @@ mod _locale {
...
+            let buff_ptr = buff.as_mut_ptr();
+
+            let string_cstr = CString::new(string.as_str()).map_err(|e| e.to_pyexception(vm))?;
+            let string_ptr = string_cstr.as_ptr();
+            let n2 = libc::strxfrm(buff_ptr, string_ptr, n1) + 1;
```
youknowone, line 90:
```suggestion
            let n2 = libc::strxfrm(buff.as_mut_ptr() as _, string_ptr, n1) + 1;
```
Then this call can be rewritten using raw pointer casting.

Lib/test/test_locale.py
```diff
@@ -343,8 +343,7 @@ def test_currency(self):
 class TestCollation(unittest.TestCase):
     # Test string collation functions
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+    @unittest.expectedFailureIfWindows("TODO: RUSTPYTHON")
```
youknowone, line 346: we recently introduced a new tool to do this better

stdlib/src/locale.rs
```diff
@@ -69,27 +69,21 @@ mod _locale {
...
     #[pyfunction]
     fn strcoll(string1: PyStrRef, string2: PyStrRef, vm: &VirtualMachine) -> PyResult {
-        unsafe {
-            let string1_cstr = CString::new(string1.as_str()).map_err(|e| e.to_pyexception(vm))?;
-            let string2_cstr = CString::new(string2.as_str()).map_err(|e| e.to_pyexception(vm))?;
-            Ok(vm.new_pyobj(libc::strcoll(string1_cstr.as_ptr(), string2_cstr.as_ptr())))
-        }
+        let cstr1 = CString::new(string1.as_str()).map_err(|e| e.to_pyexception(vm))?;
+        let cstr2 = CString::new(string2.as_str()).map_err(|e| e.to_pyexception(vm))?;
+        Ok(vm.new_pyobj(unsafe { libc::strcoll(cstr1.as_ptr(), cstr2.as_ptr()) }))
```
youknowone, line 74: the actually unsafe part is only calling libc function here.

youknowone (approved):

PR #4604, opened 2023-03-01
Fix sys.getsizeof

qingshi163:


Lib/test/test_ioctl.py
```diff
@@ -66,8 +66,6 @@ def test_ioctl_mutate_2048(self):
         # Test with a larger buffer, just for the record.
         self._check_ioctl_mutate_len(2048)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
```
DimitrisJim, line 70: fyi, I think this test is flaky, it passes locally for me too but I don't think it passes on CI

vm/src/stdlib/sys.rs
```diff
@@ -419,10 +419,30 @@ mod sys {
...
+        match sizeof() {
+            Ok(res) => Ok(res),
+            Err(err) => {
+                if let Some(default) = args.default {
+                    Ok(default)
+                } else {
+                    Err(err)
+                }
+            }
+        }
```
youknowone, line 444:
```suggestion
        sizeof().map_err(|err| args.default.ok_or(err))
```

youknowone (approved):

Lib/test/test_ioctl.py
qingshi163: surprise! it passed!!! I don't know why either

youknowone:
👍

PR #4606, opened 2023-03-02
usize::to_u32 for compiler internal conversion

youknowone:


coolreader18 (approved):
You could stick a `#![deny(clippy::cast_possible_truncation)]` to fully enforce it

PR #4608, opened 2023-03-02
Rework frozen modules and directly deserialize to CodeObject<Literal>

coolreader18:
Now frozen modules are stored as lz4-compressed in static memory until they are imported, which should help with memory usage in a freeze-stdlib environment (and the size increase from less efficient(?) compression due to potentially less dict-sharing(?? i have no clue how compression algorithms actually work) should hopefully not be too bad)

### DIFF
--- a/compiler/core/src/bytecode.rs
+++ b/compiler/core/src/bytecode.rs
@@ -1,7 +1,6 @@
 //! Implement python as a virtual machine with bytecodes. This module
 //! implements bytecode structure.
 
-use crate::marshal::MarshalError;
 use crate::{marshal, Location};
 use bitflags::bitflags;
 use itertools::Itertools;
@@ -44,6 +43,19 @@ pub trait ConstantBag: Sized + Copy {
     fn make_tuple(&self, elements: impl Iterator<Item = Self::Constant>) -> Self::Constant;
     fn make_code(&self, code: CodeObject<Self::Constant>) -> Self::Constant;
     fn make_name(&self, name: &str) -> <Self::Constant as Constant>::Name;
+}
+
+pub trait AsBag {
+    type Bag: ConstantBag;
+    #[allow(clippy::wrong_self_convention)]
+    fn as_bag(self) -> Self::Bag;
+}
+
+impl<Bag: ConstantBag> AsBag for Bag {
+    type Bag = Self;
+    fn as_bag(self) -> Self {
+        self
+    }
 }
 
 #[derive(Clone, Copy)]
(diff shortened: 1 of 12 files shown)


derive-impl/src/compile_bytecode.rs
```diff
@@ -369,12 +374,11 @@ pub fn impl_py_compile(
...
 
-    let bytes = code.to_bytes();
-    let bytes = LitByteStr::new(&bytes, Span::call_site());
+    let frozen = frozen_lib::FrozenCodeObject::encode(&code);
+    let bytes = LitByteStr::new(&frozen.bytes, Span::call_site());
 
     let output = quote! {
-        #crate_name::CodeObject::from_bytes(#bytes)
-            .expect("Deserializing CodeObject failed")
+        #crate_name::frozen_lib::FrozenCodeObject { bytes: &#bytes[..] }
```
youknowone, line 381: I think the name `py_compile` doesn't imply its result will be compressed but getting code object directly.
Renaming or making a new macro?

compiler/core/src/bytecode.rs
```diff
@@ -1483,32 +1474,83 @@ impl<C: Constant> fmt::Debug for CodeObject<C> {
...
-/// A frozen module. Holds a code object and whether it is part of a package
-#[derive(Debug)]
-pub struct FrozenModule {
-    pub code: CodeObject<ConstantData>,
-    pub package: bool,
-}
-
 pub mod frozen_lib {
+    use crate::marshal::ReadBorrowed;
+
```
youknowone, line 1479:
```suggestion
```

compiler/core/src/bytecode.rs
```diff
@@ -1483,32 +1474,83 @@ impl<C: Constant> fmt::Debug for CodeObject<C> {
...
 }
 
-/// A frozen module. Holds a code object and whether it is part of a package
-#[derive(Debug)]
-pub struct FrozenModule {
-    pub code: CodeObject<ConstantData>,
-    pub package: bool,
-}
-
 pub mod frozen_lib {
```
youknowone, line 1477: this is going bigger. maybe `frozen_lib.rs` later

derive-impl/src/compile_bytecode.rs
```diff
@@ -44,6 +44,11 @@ enum CompilationSourceKind {
     Dir(PathBuf),
 }
 
+struct FrozenModule {
```
youknowone, line 47: This struct is sharing the name `FrozenModule` with frozen_lib::FrozenModule but its code object is different. giving another name will be good to reduce later confusion.

youknowone (changes requested):
I really like the changes.

derive-impl/src/compile_bytecode.rs
coolreader18: Perhaps, yeah, but because `vm.ctx.new_code()` still accepts it, it won't actually make a difference for users.

coolreader18: They already were - the old CodeObject::from_bytes also did lz4 decompression, and py_compile!() just generated CodeObject::from_bytes(b"..."). I don't think `compile` returning a frozen code object is necessarily unintuitive (they're expecting it to compile it at build time and embed it in the binary; what's that if not freezing?) and again, since it's still accepted in the same places as before it doesn't actually make much of a difference to the user

youknowone (approved):

PR #4610, opened 2023-03-02
Implementing shake_128 and shake_256 in the hashlib module

haxelion:
I implemented support for hashlib shake_128 and shake_256.
Going through the code I made some refactoring to better match CPython behavior.

There is one test I had to disable however. The test was enabled but didn't execute any code because shake implementation were missing: https://github.com/RustPython/RustPython/blob/a7c985685146ca401011eb0c25c0b2805a767ce7/Lib/test/test_hashlib.py#L311-L312

But also I'm not sure we want to match openssl length size limitations in our implementation.

### DIFF
--- a/Lib/test/test_hashlib.py
+++ b/Lib/test/test_hashlib.py
@@ -102,8 +102,7 @@ class HashLibTestCase(unittest.TestCase):
                              'sha384', 'SHA384', 'sha512', 'SHA512',
                              'blake2b', 'blake2s',
                              'sha3_224', 'sha3_256', 'sha3_384', 'sha3_512',
-                            #  TODO: RUSTPYTHON
-                            #  'shake_128', 'shake_256'
+                             'shake_128', 'shake_256'
                             )
 
     shakes = {'shake_128', 'shake_256'}
@@ -260,8 +259,6 @@ class HashLibTestCase(unittest.TestCase):
         self.assertRaises(ValueError, hashlib.new, 'spam spam spam spam spam')
         self.assertRaises(TypeError, hashlib.new, 1)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_new_upper_to_lower(self):
         self.assertEqual(hashlib.new("SHA256").name, "sha256")
 
@@ -303,6 +300,8 @@ class HashLibTestCase(unittest.TestCase):
                 self.assertIsInstance(h.digest(), bytes)
                 self.assertEqual(hexstr(h.digest()), h.hexdigest())
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_digest_length_overflow(self):
         # See issue #34922
         large_sizes = (2**29, 2**32-10, 2**32+10, 2**61, 2**64-10, 2**64+10)
@@ -404,8 +403,6 @@ class HashLibTestCase(unittest.TestCase):
         self.check_no_unicode('blake2b')
         self.check_no_unicode('blake2s')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     @requires_sha3
     def test_no_unicode_sha3(self):
         self.check_no_unicode('sha3_224')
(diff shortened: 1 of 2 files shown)


Lib/test/test_hashlib.py
```diff
@@ -303,6 +300,8 @@ def test_hexdigest(self):
                 self.assertIsInstance(h.digest(), bytes)
                 self.assertEqual(hexstr(h.digest()), h.hexdigest())
 
+    # TODO: RUSTPYTHON
+    @unittest.skip
```
DimitrisJim, line 304: does this crash the interpreter or could it just be an expected failure?

stdlib/src/hashlib.rs
```diff
@@ -45,17 +55,33 @@ mod hashlib {
...
+            }
+        }
+    }
+
+    #[derive(FromArgs)]
+    #[allow(unused)]
+    struct XofDigestArgs {
+        #[pyarg(positional)]
+        length: usize,
+    }
```
youknowone, line 72:
```suggestion
        length: isize,
    }

    impl XofDigestArgs {
        fn length(&self, vm: &VirtualMachine) -> PyResult<usize> {
            if length < 0 {
                return vm.new_value_error("length must be non-negative".to_owned());
            }
            Ok(length as usize)
        }
    }
```
the skipped test may be fixed with this

Lib/test/test_hashlib.py
haxelion: Looking at this test, I think I know how to fix it ... wait a second 🙂

haxelion: Well, it's possible but probably worth a separate PR.
The test doesn't crash the interpreter so I'll mark it as expectedFailure.

stdlib/src/hashlib.rs
haxelion: Ok let me test this later 👍

youknowone:
Looks great. Fixing the isize part will make the patch perfect.

haxelion:
@youknowone So I tested it and it does pass the check at line 317:
https://github.com/RustPython/RustPython/blob/a7c985685146ca401011eb0c25c0b2805a767ce7/Lib/test/test_hashlib.py#L317

The issue is that instead we will fail on these lines (after a lengthy computation): https://github.com/RustPython/RustPython/blob/a7c985685146ca401011eb0c25c0b2805a767ce7/Lib/test/test_hashlib.py#L319-L320

Because we do support those sizes:
https://github.com/RustPython/RustPython/blob/a7c985685146ca401011eb0c25c0b2805a767ce7/Lib/test/test_hashlib.py#L308

So if we go back to CPython, they have two implementations:

- A fallback implementation which does not support those larger sizes: https://github.com/python/cpython/blob/568fc0dee42a353f327b059a48f97c911de904b3/Modules/_sha3/sha3module.c#L455
- An default  (if the library is found on the system) OpenSSL implementation which does support those sizes and skip the test with this check: https://github.com/RustPython/RustPython/blob/a7c985685146ca401011eb0c25c0b2805a767ce7/Lib/test/test_hashlib.py#L313-L315

So there are basically two solutions:

1. Adhere to the fallback implementation and only accept an `unsigned long`.
2. Copy the module structure of the CPython OpenSSL implementation so that we can skip the test.

I'm unsure which solution you would prefer.

youknowone:
The second one is preferred, but it doesn't sound like a part to cover in this PR anymore. 
Would you finish them after only changing the `usize` to `isize`?

youknowone:
I am not sure but second approach also will not be that hard.
If you'd like to try,

adding
```rust
#[pyclass]
struct HASH;
```
and changing `PyHashser` to
```rust
#[pyclass(module = "hashlib", name = "hasher", base = "HASH")]
```
will be a trial

haxelion:
@youknowone So I already tried this change but it's not enough. The classes need to be in a separate `_hashlib` module for this to work ... I tried the change but it became a much larger PR so I gave up.

I will push the change with `isize`.

youknowone (approved):
Thank you!

youknowone:
I got what you meant. both CPython's `hashilb` and `test_hashlib` heavily suppose `_hashlib` is openssl implementation.

Issue #4541, opened 2023-02-21
Command line parameter -P


youknowone:
https://docs.python.org/3/using/cmdline.html#cmdoption-P

https://github.com/RustPython/RustPython/pull/4537/files#diff-cb42fd26146ec61cb2242243494a9f314667a9a39153ad17e37465ff5a8a8000R536

naonus:
I'll take this issue

AgentK9:
@youknowone - just FYI, tasks 2 and 3 are also completed - https://github.com/RustPython/RustPython/blob/main/src/settings.rs#L245-L249

After diving into the code, I don't see where it is implemented, but it seems like it is after testing it out.

Could you please close this issue for future people like me?

Thanks

Issue #4564, opened 2023-02-24
Update python libraries and tests from CPython 3.11


youknowone:
Though it will be sometimes blocked by rust side issues, don't worry!
Please just create PR about the broken test and and create issues related to the new bugs.

If you are new to updating cpython libraries, this guide will be useful:
https://github.com/RustPython/RustPython/wiki/How-to-update-test-files

When updating module Lib/X, please don't forget to also update Lib/test/test_X

A few libraries doesn't need to be updated because CPython also kept it same as previous version. In that case, please leave a comment to let us mark it as done. Thank you!

Tips
- To get specific version of CPython source code, try `git checkout v3.11.2` from CPython repository.
- Please check #4064 for example. Not the entire changes but the each commits. One of the middle size example is https://github.com/RustPython/RustPython/commit/17e12dea1ec11b871a9e5d3a4a26e2f9ad83fde2 . Sometimes it can be really simple like https://github.com/RustPython/RustPython/pull/4064/commits/9571a68c01284191d4b18229694c303dad8992ff
- Please create one PR for one library+test pair whenever possible. Once you stuck by dependencies, go to one of the dependency first and finish it before restarting the stuck one unless they become simpler to be updated at the same time. Normally adding more libraries make passing tests (a lot) harder.
- To run tests, try `cargo run --features ssl -- -m test -v <test_name>`. e.g. If you changed `test_unicode.py`, `cargo run --features ssl -- -m test -v test_unicode`
  - If it runs too slow, try to add `--release` like `cargo run --release --features ssl -- -m test -v test_unicode`

# List of libraries
- [ ] `__future__.py`
  - [ ] maybe related to #4060?
- [x] `__hello__.py` #4570
- [x] `__phello__` #4570
- [x] `_collections_abc.py`https://github.com/RustPython/RustPython/pull/5016
- [ ] `_compat_pickle.py`
- [x] `_compression.py` #4688 
- [x] `_markupbase.py` #4703
- [x] ~~`_osx_support.py`~~ unchanged
  - [x] #4670
- [ ] `_py_abc.py`
- [ ] `_pycodecs.py`
- [ ] `_pydecimal.py`
- [ ] `_pyio.py`
- [ ] `_sitebuiltins.py`
- [ ] `_threading_local.py`
- [ ] `_weakrefset.py`
- [x] `abc.py` #4985 
- [x] `aifc.py`
- [x] `antigravity.py` — #4591 
- [x] `argparse.py` — #4592 
[...]

PR #4611, opened 2023-03-02
Add command line parameter -P

naonus:
It's not passed test case with errors below

AssertionError: '[]' != "['-P']"
- []
+ ['-P']

#4541

### DIFF
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -309,6 +309,7 @@ def _args_from_interpreter_flags():
         'verbose': 'v',
         'bytes_warning': 'b',
         'quiet': 'q',
+        'safe_path': 'P'
         # -O is handled in _optim_args_from_interpreter_flags()
     }
     args = _optim_args_from_interpreter_flags()
--- a/Lib/subprocess.py
+++ b/Lib/subprocess.py
@@ -309,7 +309,6 @@ def _args_from_interpreter_flags():
         'verbose': 'v',
         'bytes_warning': 'b',
         'quiet': 'q',
-        'safe_path': 'P'
         # -O is handled in _optim_args_from_interpreter_flags()
     }
     args = _optim_args_from_interpreter_flags()
--- a/Lib/test/test_cmd_line.py
+++ b/Lib/test/test_cmd_line.py
@@ -1008,6 +1008,7 @@ class IgnoreEnvironmentTest(unittest.TestCase):
             PYTHONSAFEPATH="1",
         )
 
+
 class SyntaxErrorTests(unittest.TestCase):
     def check_string(self, code):
         proc = subprocess.run([sys.executable, "-"], input=code,
--- a/Lib/test/test_cmd_line.py
+++ b/Lib/test/test_cmd_line.py
@@ -1008,26 +1008,6 @@ class IgnoreEnvironmentTest(unittest.TestCase):
             PYTHONSAFEPATH="1",
         )
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
-    def test_pythonsafepath_env(self):
-        # Test the PYTHONSAFEPATH environment variable
-        code = "import sys; print(sys.flags.safe_path)"
-        env = dict(os.environ)
-        env.pop('PYTHONSAFEPATH', None)
-        args = (sys.executable, '-P', code)
-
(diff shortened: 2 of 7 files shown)


src/settings.rs
```diff
@@ -228,6 +233,12 @@ fn settings_from(matches: &ArgMatches) -> (Settings, RunMode) {
         settings.dont_write_bytecode = true;
     }
 
+    if matches.is_present("safe-path")
+        || (!ignore_environment && env::var_os("PYTHONSAFEPATH").is_some())
+    {
+        settings.safe_path = true;
```
youknowone, line 239: interpreter setting is done here, but safe_path in Flags is not set yet in `Flags::from_settings`.
The test is running
```python
from test.support import args_from_interpreter_flags
args_from_interpreter_flags()
```

naonus: fixed safe_path in Flags is to be set. but it still not passed test cases

Lib/test/test_cmd_line.py
```diff
@@ -884,6 +884,25 @@ def test_sys_flags_not_set(self):
...
+        proc = subprocess.run(args, stdout=subprocess.PIPE,
+                              universal_newlines=True, env=env)
+        self.assertEqual(proc.stdout.rstrip(), 'False')
+        self.assertEqual(proc.returncode, 0, proc)
+
+        env['PYTHONSAFEPATH'] = '1'
+        proc = subprocess.run(args, stdout=subprocess.PIPE,
+                              universal_newlines=True, env=env)
+        self.assertEqual(proc.stdout.rstrip(), 'True')
+        self.assertEqual(proc.returncode, 0, proc)
```
fanninpm, line 905: Is this from CPython?

naonus: No, I just added it. I'll check CPython.

naonus: This test is not existed in CPython. Is it better deleting this code ?

Lib/subprocess.py
```diff
@@ -288,6 +288,7 @@ def _args_from_interpreter_flags():
         'verbose': 'v',
         'bytes_warning': 'b',
         'quiet': 'q',
+        'safe_path': 'P'
```
youknowone, line 291: if subprocess.py is too old, rather than editing it yourself, could you try to update it by following the library update guideline on #4564?

src/settings.rs
youknowone: at least the error is changed. something happened!

Lib/test/test_support.py
```diff
@@ -516,7 +516,7 @@ def check_options(self, args, func, expected=None):
         self.assertEqual(proc.returncode, 0)
 
     # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+    # @unittest.expectedFailure
```
DimitrisJim, line 519: nit: remember to completely remove these comments

DimitrisJim:
hey @naonus do you have time to follow along with this PR? If not, we can pick it up.

Lib/test/test_cmd_line.py
```diff
@@ -884,6 +884,25 @@ def test_sys_flags_not_set(self):
...
         )
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    def test_pythonsafepath_env(self):
+        # Test the PYTHONSAFEPATH environment variable
+        code = "import sys; print(sys.flags.safe_path)"
+        env = dict(os.environ)
+        env.pop('PYTHONSAFEPATH', None)
+        args = (sys.executable, '-P', code)
```
youknowone, line 894:
```suggestion
        args = (sys.executable, '-P', '-c', code)
```

Lib/test/test_cmd_line.py
```diff
@@ -884,6 +884,25 @@ def test_sys_flags_not_set(self):
...
+
+        proc = subprocess.run(args, stdout=subprocess.PIPE,
+                              universal_newlines=True, env=env)
+        self.assertEqual(proc.stdout.rstrip(), 'False')
```
youknowone, line 898:
```suggestion
        self.assertEqual(proc.stdout.rstrip(), 'True')
```

youknowone (changes requested):

youknowone (approved):
Thank you for contributing!

PR #4612, opened 2023-03-02
Add vm.format to call __format__ like PyObject_Format

youknowone:


vm/src/builtins/object.rs
```diff
@@ -242,15 +242,18 @@ impl PyBaseObject {
...
 
     #[pymethod(magic)]
-    fn format(obj: PyObjectRef, format_spec: PyStrRef, vm: &VirtualMachine) -> PyResult<PyStrRef> {
-        if format_spec.as_str().is_empty() {
-            obj.str(vm)
-        } else {
-            Err(vm.new_type_error(format!(
+    fn format(
+        obj: PyObjectRef,
+        format_spec: Option<PyStrRef>,
```
youknowone, line 247: this is optional

vm/src/builtins/str.rs
```diff
@@ -745,10 +745,24 @@ impl PyStr {
...
+        vm: &VirtualMachine,
+    ) -> PyResult<PyStrRef> {
+        let spec = spec.as_ref().map_or("", |spec| spec.as_str());
+        if spec.is_empty() {
+            return if zelf.class().is(vm.ctx.types.str_type) {
+                Ok(zelf.to_owned())
+            } else {
+                zelf.as_object().str(vm)
+            };
+        }
```
youknowone, line 760: str optimization when no spec

vm/src/protocol/object.rs
```diff
@@ -330,12 +330,22 @@ impl PyObject {
...
-            s.try_into_value(vm)
-        }
+        let obj = match self.to_owned().downcast_exact::<PyStr>(vm) {
+            Ok(s) => return Ok(s.into_pyref()),
+            Err(obj) => obj,
+        };
+        // TODO: replace to obj.class().slots.str
+        let str_method = match vm.get_special_method(obj, identifier!(vm, __str__))? {
+            Ok(str_method) => str_method,
+            Err(obj) => return obj.repr(vm),
```
youknowone, line 340: repr is called here. it means we may have duplicated repr call somewhere.

vm/src/builtins/object.rs
```diff
@@ -183,6 +183,7 @@ impl PyBaseObject {
     /// Return str(self).
     #[pymethod(magic)]
     fn str(zelf: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyStrRef> {
+        // FIXME: try tp_repr first and fallback to object.__repr__
```
youknowone, line 186: ah, found the redundant place. but no way to fix it before tp_repr

DimitrisJim (approved):
I only quickly glanced through the CPython equivalent but this seems good to me

Issue #4614, opened 2023-03-02
Improve: binary ops with Number Protocol


xiaozhiyan:
Related to https://github.com/RustPython/RustPython/issues/3244

fanninpm:
cc @qingshi163

PR #4615, opened 2023-03-02
Improve: binary ops with Number Protocol

xiaozhiyan:
Closes https://github.com/RustPython/RustPython/issues/4614
Based on https://github.com/RustPython/RustPython/pull/4139
Related to https://github.com/RustPython/RustPython/issues/3244

This PR will not pass some tests at this moment, since Number Protocol is missing for some types.
Let me create separate PRs to implement (part of) missing Number Protocols, and then rebase this PR once those are merged.

closes #4639
closes #4672


youknowone:
when rustpython crashes on initialization step, debugger is very useful

youknowone:
The next step is update_slot in slot.rs.
When operator is overriden from Python side, we have to set a function to the python function from rust side.

youknowone:
To support simple binary operators, looking for `index_wrapper` or `len_wrapper` will be useful. One wrapper matches to one slot.
To support self-mutating unary ops like `__iadd__` and `__add__`, looking for `richcompare_wrapper` will be useful. When user defined `__add__` on python side, both `__add__` and `__iadd__` slots have to be set.(1 python function to n slots) `richcompare_wrapper` is doing similar work - but the opposite way. (n python funciton to 1 slot)

vm/src/protocol/number.rs
```diff
@@ -197,8 +195,73 @@ impl PyNumberMethods {
...
+    InplaceAnd,
+    InplaceXor,
+    InplaceOr,
+    FloorDivide,
+    TrueDivide,
+    InplaceFloorDivide,
+    InplaceTrueDivide,
+    MatrixMultiply,
+    InplaceMatrixMultiply,
+}
```
qingshi163, line 262: This is ok but let the op function to take a lambda should also work

qingshi163: Oh, I mean closure in rust. Instead having a matching table from enum to methods(get_binary_op), we could make binary_ops generic that would take a FnOnce to specifiy which op will be use.

youknowone: It sounds like function pointer vs pointer offset design decision. I think either design works well enough to run the features. 
Probably we can discuss better once this is fully done and glance at the design overview.

youknowone:
your works seems based on #4139, right? By respecting the original contributor's work, I'd like to keep the original commit as part of this PR.

vm/src/types/slot.rs
```diff
@@ -214,6 +214,18 @@ fn float_wrapper(num: PyNumber, vm: &VirtualMachine) -> PyResult<PyRef<PyFloat>>
...
+    ($name:ident) => {
+        |num, other, vm| {
+            vm.call_special_method(
+                num.obj.to_owned(),
+                identifier!(vm, $name),
+                (other.to_owned(),),
+            )
+        }
+    };
+}
```
youknowone, line 227:
```suggestion
macro_rules! number_binary_op_wrapper {
    ($op:ident, $rop:ident) => {
        |num, other, vm| {
            if let Ok(op) = vm.get_special_method(num.obj.to_owned(), identifier!(vm, $op))? {
                op.invoke((other.to_owned(),), vm)
            } else if let Ok(op) =
                vm.get_special_method(other.to_owned(), identifier!(vm, $rop))?
            {
                op.invoke((num.obj.to_owned(),), vm)
            } else {
                unreachable!("invalid wrapper")
            }
        }
    };
}
```
This change will be helpful to resolve builtin_pow error.

vm/src/types/slot.rs
```diff
@@ -469,6 +481,194 @@ impl PyType {
                 toggle_ext_func!(number_methods, float, float_wrapper);
                 update_pointer_slot!(as_number, number_methods);
             }
+            _ if name == identifier!(ctx, __add__) => {
+                toggle_ext_func!(number_methods, add, number_binary_op_wrapper!(__add__));
+                update_pointer_slot!(as_number, number_methods);
+            }
```
youknowone, line 487: And then this part will look like:
```suggestion
            _ if name == identifier!(ctx, __add__)  || name == identifier!(ctx, __radd__)=> {
                toggle_ext_func!(number_methods, add, number_binary_op_wrapper!(__add__, __radd__)); // no toggle but ok for a while
                update_pointer_slot!(as_number, number_methods);
            }
```

xiaozhiyan:
The following is the behavior in CPython
```
>>> class C:
...     def __add__(self, x):
...             return x + 100
...     def __radd__(self, x):
...             return x + 200
... 
>>> c = C()
>>> c + 5
105
>>> 5 + c
205
```

vm/src/vm/vm_ops.rs
```diff
@@ -103,371 +132,260 @@ impl VirtualMachine {
...
+            &PyNumberBinaryOpSlot::RightMultiply,
+        )?;
+        if !result.is(&self.ctx.not_implemented) {
+            return Ok(result);
+        }
+        if let Ok(seq_a) = PySequence::try_protocol(a, self) {
+            let n =
+                b.try_int(self)?.as_bigint().to_isize().ok_or_else(|| {
+                    self.new_overflow_error("repeated bytes are too long".to_owned())
+                })?;
```
xiaozhiyan, line 349: Currently float number will also be converted to `isize` and used to repeat seq, which is unexpected.
```
>>>>> b'a' * 3.14
b'aaa'
```
The above would fail in CPython with the following error message.
```
TypeError: can't multiply sequence by non-int of type 'float'
```
@youknowone What is the proper way to restrict non-int numbers here please?

youknowone: instead of `try_int`, `try_index` or `try_index_opt` will be helpful

youknowone:
I finally understand the trick. instead of `number_binary_op_wrapper`, adding `number_binary_op_left_wrapper` and `number_binary_op_right_wrapper` allows to share single PyNumber operation. And they can be separately overriden from python side. We need another `PyNumberSlots` struct.

vm/src/vm/vm_ops.rs
xiaozhiyan: `try_int` has been changed to `try_index` to disable repeating sequences by float number.

youknowone:
@xiaozhiyan I made a rough sketch of the cpython way
https://github.com/RustPython/RustPython/commit/10c1929eb00953b04e53f33c7ffac600f0487225

youknowone:
the difference between my version and cpython is slot implementation.
I used different slot function for wrapper and number protocol implementation. But cpython use a single function for both cases
https://github.com/python/cpython/blob/6716254e71eeb4666fd6d1a13857832caad7b19f/Objects/typeobject.c#L7792-L7833

youknowone:
Because the patch is really huge, I’d like to suggest to split it down to multiple partial fixes. 

Currently `vm.binary_{i}op{1}` with `#[allow(dead_code)]` seems to be a chunk.
Once finished, `update_slot!` stuff can be a chunk too. Even if it is not very correct at the moment, it will be fine unless we actually call them from `vm._add` etc

@qingshi163 how do you think?

qingshi163:
I am bit confuse about the right_ops witch did not exist in cpython, why can't we just swap the direction as what we doing before?

xiaozhiyan:
> I am bit confuse about the right_ops witch did not exist in cpython, why can't we just swap the direction as what we doing before?

As in the example https://github.com/RustPython/RustPython/pull/4615#issuecomment-1455131249 , left and binary ops and be defined separately. So we need to store them separately. Please check the latest commit https://github.com/RustPython/RustPython/pull/4615/commits/cdac5cfb69ee5bc976164b1da05d2b0ebac2ca1e to create `PyNumberSlots` to save both left and right binary ops.

qingshi163:
> > I am bit confuse about the right_ops witch did not exist in cpython, why can't we just swap the direction as what we doing before?
> 
> As in the example [#4615 (comment)](https://github.com/RustPython/RustPython/pull/4615#issuecomment-1455131249) , left and binary ops and be defined separately. So we need to store them separately. Please check the latest commit [cdac5cf](https://github.com/RustPython/RustPython/commit/cdac5cfb69ee5bc976164b1da05d2b0ebac2ca1e) using `PyNumberSlots` to save both left and right binary ops.

Oh, I see the trick.

vm/src/builtins/float.rs
```diff
@@ -546,50 +545,29 @@ impl Hashable for PyFloat {
 impl AsNumber for PyFloat {
     fn as_number() -> &'static PyNumberMethods {
         static AS_NUMBER: Lazy<PyNumberMethods> = Lazy::new(|| PyNumberMethods {
```
qingshi163, line 547: as PyNumberMethods is fully immutable, we should able to remove the Lazy<> things

xiaozhiyan:
I temporarily implemented Number Protocol for `PyBool` in commit https://github.com/RustPython/RustPython/pull/4615/commits/22fca357c587ff9c0315ff11b14c03c6e1ab7885  to see whether this PR can pass all the tests.

- As discussed in https://github.com/RustPython/RustPython/pull/4639 , Number Protocol of `PyBool` will be implemented by inheriting slots of `PyInt`, which depends on this PR. 
- This PR also requires Number Protocol of `PyBool` to pass tests.

So let me deprecate the PR https://github.com/RustPython/RustPython/pull/4639 and implement it within this PR.

xiaozhiyan:
As in https://github.com/RustPython/RustPython/actions/runs/4357591970/jobs/7617136283 , 
```
2 tests failed:
    test_array test_bytes
```

```
TypeError: '*' not supported between instances of 'int' and 'array'
```

```
RUST_BACKTRACE=1 cargo run Lib/test/test_bytes.py

thread 'main' panicked at 'assertion failed: self.payload_is::<T>()', vm/src/object/core.rs:730:9
```

youknowone:
`mod` operator of `bytes` and `array` is not communicative like numeric types. their mod and rmod are different.

vm/src/vm/vm_ops.rs
```diff
@@ -103,371 +125,217 @@ impl VirtualMachine {
...
+    ) -> PyResult {
+        let result = self.binary_iop1(a, b, iop_slot, op_slot)?;
+        if !result.is(&self.ctx.not_implemented) {
+            return Ok(result);
+        }
+        Err(self.new_unsupported_binop_error(a, b, op))
+    }
+
+    binary_func!(_sub, Subtract, "-");
+    binary_func!(_mod, Remainder, "%");
```
youknowone, line 234: So.. maybe cpython have special way to handle this? or it came from the different implementation of binary op wrapper?

youknowone:
I found why array doesn't work. It missed update_slot. I can fix it tomrorow.

xiaozhiyan:
Since related changes in this PR seem to increase along the way, now it may be a good timing to move the following loosely coupled part to another PR as prerequisite of the current one.

- Refactor all existing Number Protocol methods (actually only some of them need big changes)
  - so that they accept arguments of type similar to (&PyObject, &PyObject, &VirtualMachine)
  - within this method, we don't need to consider whether the caller comes from left or right binary op, and the method just needs to handle two `PyObject` arguments and return the value (or not_implemented) as expected

- Then these methods can be used by both left and right binary ops in the current PR

@youknowone If you also agree with the approach, I will open a new PR to work on that as prerequisite of the current PR.

xiaozhiyan:
I pushed latest PR at https://github.com/RustPython/RustPython/commit/a2da2d20ae0c49d9ac3e8e7bb6a8c419816e8e58 , which contains 3 `Temp` commits for testing purpose to see whether there're any other test failures in CI.

xiaozhiyan:
From latest CI after the temp fixes,

- there's a minor clippy warning, and I've fixed it locally

- the following tests failed
```
test_collections test_dict test_dictviews test_types
```

- `test_collections`
  - methods defined in `ChainMap` like `__or__` not taking effect

- `test_dict`
  - need to implement Number Protocol for `PyDictItems`
```
'^' not supported between instances of 'dict_items' and 'dict_items'
```

- `test_dictviews`
  - need to implement Number Protocol for `PyDictKeys`
```
'&' not supported between instances of 'dict_keys' and 'list_iterator'
```

- `test_types`
```
'|' not supported between instances of 'GenericAlias' and 'GenericAlias'
```

xiaozhiyan:
After (temporarily) implementing Number Protocol for `PyDictKeys` and `PyDictItems`, now `test_dict` and `test_dictviews` succeeded, with the following two testing errors remained.
https://github.com/RustPython/RustPython/actions/runs/4371776446/jobs/7647999208

```
test_collections test_types
```

youknowone:
I agree we need to remove PyNumber from PyNumberMethods.
But it will last on PyNumberSlots.

I will take a look how we can handle it smart way.

@qingshi163 because this PR is already huge, I'd like to merge it unless it has obvious wrong design or bugs. could you review it?

xiaozhiyan:
Now that the PR has past the CI, except for the WASM error inherited from `main` branch.
I will proceed to refactor it a little bit, and add necessary `TODO` comments where needed.

xiaozhiyan:
I've rearranged the commits and pushed to the PR.
Please help to review based on this version.
Some commits are marked as `Temp` intentionally, as they may be further improved in separate PRs.

Lib/test/test_mmap.py
```diff
@@ -721,6 +721,8 @@ def test_resize_past_pos(self):
         self.assertRaises(ValueError, m.write_byte, 42)
         self.assertRaises(ValueError, m.write, b'abc')
 
+    # TODO: RUSTPYTHON
+    @unittest.skip
```
fanninpm, line 725:
```suggestion
    @unittest.skip("TODO: RUSTPYTHON, WhateverError: this thingy doesn't work")
```

…but please replace the `WhateverError: this thingy doesn't work` part with something more useful than that.

xiaozhiyan: Let me change it to the following
```
@unittest.skip('TODO: RUSTPYTHON, Number Protocol is not properly implemented yet')
```

xiaozhiyan:
@youknowone 
The latest CI failed with Clippy error, but that part of source has not been changed for a long time. Why suddenly CI treats it as an error and makes the CI fail?
https://github.com/RustPython/RustPython/actions/runs/4375751891/jobs/7656913583

vm/src/protocol/number.rs
```diff
@@ -238,12 +306,12 @@ impl PyNumber<'_> {
 
     // PyIndex_Check
     pub fn is_index(&self) -> bool {
-        self.methods().index.load().is_some()
+        self.obj.class().slots.number.index.load().is_some()
```
youknowone, line 309: binary ops operands can be swapped, but is this also need to access obj.class() again?

vm/src/vm/vm_ops.rs
```diff
@@ -103,371 +125,217 @@ impl VirtualMachine {
...
+        a: &PyObject,
+        b: &PyObject,
+        iop_slot: &PyNumberBinaryOpSlot,
+        op_slot: &PyNumberBinaryOpSlot,
     ) -> PyResult {
-        if rhs.fast_isinstance(lhs.class()) {
-            let lop = lhs.get_class_attr(reflection);
-            let rop = rhs.get_class_attr(reflection);
-            if let Some((lop, rop)) = lop.zip(rop) {
-                if !lop.is(&rop) {
```
youknowone, line 153: we may lost this check. it seems to related to test_collections failure

vm/src/protocol/number.rs
youknowone: ah, got it. we didn't handle unary operator yet.

vm/src/vm/vm_ops.rs
```diff
@@ -131,7 +131,12 @@ impl VirtualMachine {
     ///   b.rop(b,a)[*], a.op(a,b), b.rop(b,a)
     ///
     /// [*] only when Py_TYPE(a) != Py_TYPE(b) && Py_TYPE(b) is a subclass of Py_TYPE(a)
-    fn binary_op1(&self, a: &PyObject, b: &PyObject, op_slot: &PyNumberBinaryOpSlot) -> PyResult {
+    pub fn binary_op1(
+        &self,
+        a: &PyObject,
+        b: &PyObject,
+        op_slot: &PyNumberBinaryOpSlot,
+    ) -> PyResult {
```
xiaozhiyan: Which format config should we use please? When I run `cargo fmt`, it will just change back to one line as before.

fanninpm:
> The latest CI failed with Clippy error, but that part of source has not been changed for a long time. Why suddenly CI treats it as an error and makes the CI fail?
> https://github.com/RustPython/RustPython/actions/runs/4375751891/jobs/7656913583

Perhaps it has to do with Rust 1.68.0 being released.

vm/src/vm/vm_ops.rs
xiaozhiyan: I think the purpose of the old code is to follow the CPython logic to check whether `lop` and `rop` are referring to the same `op`, in which case `rop` should not be used.

The new implementation uses the following part when loading `rop`.
```
        let mut slot_b = if b.class().is(a.class()) {
            None
        } else {
            match b.class().slots.number.get_right_binary_op(op_slot)? {
                Some(slot_b)
                    if slot_b as usize == slot_a.map(|s| s as usize).unwrap_or_default() =>
                {
                    None
                }
                slot_b => slot_b,
            }
        };
```

youknowone (approved):
let me handle 1.68 things

xiaozhiyan:
@youknowone 
> let me handle 1.68 things
Please help to check my comments in https://github.com/RustPython/RustPython/pull/4673 in case they're useful.

youknowone:
Thanks!

qingshi163:
Sorry for late review, as long as the ci passed, we can accept this pr.

But there is something I want to clarify:

1. Heap type is the type which created dynamicly like class in python etc.
2. Heap type will have HeapTypeExt witch is the extra space only allocate for heap type.
3. PyTypeSlots::as_number, as_sequence, as_mapping etc is a pointer to tell the type where to look for the function and it has only two possible place to pointing witch is either a static function from AsNumber::as_number or a self-referenced HeapTypeExt::number_methods etc.
4. The type slots for number, sequence, mapping is not necessary but for speedup, witch means we can always fallback to a wrapper to call in python.
5. For right_* functions I think we can have a generic way to do it, to put it into the type slots is not ideal, we do not need to store the __r*__ functions, we just need to fallback the type.
6. The PyNumber type it self is loosy, when convert from PyObject to PyNumber it will just give a empty implement if the object is not supporting number protocol.
7. Cpython is inherit the slots but this makes the complex while you updateing the parent class's slot than need to recusively fallback all the subclasses's slots, the way we doing to search via mro maybe will slow down a little bit(I don't think will be big as most internal type should reach the protocol in the first loop) but much easier for now.

youknowone:
@qingshi163 All other parts sounds good, but for No. 5, for each binary operator, having 2 slots and 1 number method is the CPython design.

qingshi163:
> @qingshi163 All other parts sounds good, but for No. 5, for each binary operator, having 2 slots and 1 number method is the CPython design.

https://github.com/python/cpython/blob/main/Include/cpython/object.h
https://github.com/python/cpython/blob/ca066bdbed85094a9c4d9930823ce3587807db48/Objects/typeobject.c#L7221-L7242

cpython do not store the right_* functions in the type slots, it fallback the slots and than call it opposite direction

youknowone:
we have a little bit different design now, so the wrapper looks different. but isn't `wrap_binaryfunc_r` the wrapper used by `RBINSLOT` as one of the 2 binaryfunc entries?

xiaozhiyan:
Measure performance enhancement of this PR

- Download https://github.com/python/pyperformance/blob/main/pyperformance/data-files/benchmarks/bm_nbody/run_benchmark.py to `benches/benchmarks/bm_nbody.py`

- Before patch
```
% git log -n1
commit 1fceeab0fcc77c11dfe94092b27b2164b94d62e4 (HEAD)
Merge: 51020cd63 4d05077ec
Author: Jeong YunWon <69878+youknowone@users.noreply.github.com>
Date:   Thu Mar 9 21:19:47 2023 +0900

    Merge pull request #4674 from carlosmiei/copy-update
    
    Update copy.py from CPython 3.11

% cargo run --release benches/benchmarks/bm_nbody.py -o base.json
.....................
nbody: Mean +- std dev: 4.01 sec +- 0.17 sec

% cargo run --release benches/benchmarks/bm_nbody.py -o base2.json
.....................
nbody: Mean +- std dev: 3.91 sec +- 0.14 sec
```

- After patch
```
% git log -n1
commit ea665cb743ef1f2a1a2b6030e8384d7ab814b305 (HEAD)
Merge: 1fceeab0f 228772020
Author: Jeong YunWon <69878+youknowone@users.noreply.github.com>
Date:   Fri Mar 10 02:47:01 2023 +0900

    Merge pull request #4615 from xiaozhiyan/binaryops-with-number-protocol
    
    Improve: binary ops with Number Protocol

% cargo run --release benches/benchmarks/bm_nbody.py -o patch.json
.....................
nbody: Mean +- std dev: 1.24 sec +- 0.08 sec

% cargo run --release benches/benchmarks/bm_nbody.py -o patch2.json
.....................
nbody: Mean +- std dev: 1.28 sec +- 0.12 sec
```

- CPython 3.11
```
% python -V
Python 3.11.0

% python benches/benchmarks/bm_nbody.py -o cpython.json
.....................
nbody: Mean +- std dev: 101 ms +- 2 ms

% python benches/benchmarks/bm_nbody.py -o cpython2.json
.....................
nbody: Mean +- std dev: 104 ms +- 7 ms
```

- Performance comparison
```
% pyperf compare_to cpython.json patch.json base.json --table 
+-----------+---------+-------------------------+-------------------------+
| Benchmark | cpython | patch                   | base                    |
+===========+=========+=========================+=========================+
| nbody     | 101 ms  | 1.24 sec: 12.28x slower | 4.01 sec: 39.79x slower |
+-----------+---------+-------------------------+-------------------------+

% pyperf compare_to cpython2.json patch2.json base2.json --table
+-----------+----------+-------------------------+-------------------------+
| Benchmark | cpython2 | patch2                  | base2                   |
+===========+==========+=========================+=========================+
| nbody     | 104 ms   | 1.28 sec: 12.34x slower | 3.91 sec: 37.54x slower |
+-----------+----------+-------------------------+-------------------------+
```

vm/src/types/slot.rs
```diff
@@ -85,6 +84,7 @@ pub struct PyTypeSlots {
 
     // The count of tp_members.
     pub member_count: usize,
+    pub number: PyNumberSlots,
```
Snowapril, line 87: @xiaozhiyan Sorry for late comment 😂Could I ask you the reason why not to use `Option` for `PyNumberSlots`? If we want to know which type implements the Number protocol, shouldn't we verify all the fields in the Number methods?

xiaozhiyan: For a type to implement Number Protocol, basically it means the type implements a subset of Number Protocol methods.
I'm not sure what the scenario is to check whether a type has implemented "at least one Number Protocol method", instead of checking whether a type has implemented "certain specified method".
If it's really needed in such scenarios, I think we may be able to change this implementation to use `Option` without side effect. Otherwise, it may be better to keep it as is to avoid an additional layer of `Option` unwrapping.

youknowone: due to difference of AsNumber/PyNumberSlots implementation strategy, i guess PyNumberSlots always exists while AsNumber is not.

xiaozhiyan: According to current implementation,

- for embedded types, if they don't implement `AsNumber` trait, `extend_slots` will not be executed, and their `PyNumberSlots` will contain default method implementation, i.e. `None`

- for user defined types, if they implement any Number Protocol method, say, `__radd__`, the function will be saved in  corresponding field of `PyNumberSlots` for the type

- when trying to call certain Number Protocol method for a type, `PyNumberSlots` will be looked up, without considering whether the type has implemented `AsNumber` or not

Snowapril: Ah I see. Thanks for detail explanation 😊

PR #4617, opened 2023-03-02
Pin `youknowone/unicode_names2` to a specific revision

charliermarsh:
This avoids the `Updating git repository `https://github.com/youknowone/unicode_names2.git`` on every Cargo command.


youknowone (approved):
Thanks! I was forgetting about this

konstin:
see also https://github.com/rust-lang/cargo/issues/11792

PR #4621, opened 2023-03-03
Update test_atexit from cpython 3.11.2

dalinaum:


DimitrisJim (approved):
thanks for this @dalinaum!

PR #4623, opened 2023-03-04
Treat match and case as soft keywords in lambda assignments

charliermarsh:
Will close https://github.com/charliermarsh/ruff/issues/3337 in Ruff.


charliermarsh:
(I had to do this somewhere else in Ruff too, I think it's correct to special-case it.)

youknowone (approved):

PR #4627, opened 2023-03-04
Update Lib/test/exception_hierarchy.txt from CPython 3.11.2

dalinaum:
> @fanninpm(#4624) One thing: since this references a file named exception_hierarchy.txt, it would make sense if it could also be brought up to date.

I didn't realize this file needed to be updated.


youknowone (approved):
Oh, they changed formatting.

If you are interested in, we manage our list of exceptions by following this file in
`vm/src/stdlib/builtins.rs`

PR #4636, opened 2023-03-05
Fix AsNumber for PyComplex bugs

youknowone:
It was not converting int/float to complex.

I found it during testing #4615 , thanks @xiaozhiyan


youknowone:
going to merge this first to unblock #4615

Issue #4638, opened 2023-03-05
Implement Number Protocol for `PyBool`


xiaozhiyan:
Related to https://github.com/RustPython/RustPython/issues/3244

PR #4639, opened 2023-03-05
Implement Number Protocol for `PyBool`

xiaozhiyan:
Closes https://github.com/RustPython/RustPython/issues/4638

In CPython, `bool` is a subclass of `int`, and its `as_number` method overrides `and`, `xor` and `or`.

In RustPython, ~as `PyBool` is not a subclass of `PyInt`~, the `as_number` "overriding" is done by including all other methods of `PyInt`.

### DIFF
--- a/vm/src/builtins/bool.rs
+++ b/vm/src/builtins/bool.rs
@@ -1,11 +1,19 @@
 use super::{PyInt, PyStrRef, PyType, PyTypeRef};
 use crate::{
-    class::PyClassImpl, convert::ToPyObject, function::OptionalArg, identifier, types::Constructor,
+    atomic_func,
+    class::PyClassImpl,
+    convert::{ToPyObject, ToPyResult},
+    function::OptionalArg,
+    identifier,
+    protocol::PyNumberMethods,
+    types::{AsNumber, Constructor},
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyResult, TryFromBorrowedObject,
     VirtualMachine,
 };
+use crossbeam_utils::atomic::AtomicCell;
 use num_bigint::Sign;
 use num_traits::Zero;
+use once_cell::sync::Lazy;
 use std::fmt::{Debug, Formatter};
 
 impl ToPyObject for bool {
@@ -102,7 +110,7 @@ impl Constructor for PyBool {
     }
 }
 
-#[pyclass(with(Constructor))]
+#[pyclass(with(Constructor, AsNumber))]
 impl PyBool {
     #[pymethod(magic)]
     fn repr(zelf: bool, vm: &VirtualMachine) -> PyStrRef {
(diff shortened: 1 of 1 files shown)


vm/src/builtins/bool.rs
```diff
@@ -166,6 +174,48 @@ impl PyBool {
...
+            remainder: int_method!(remainder),
+            divmod: int_method!(divmod),
+            power: int_method!(power),
+            negative: int_method!(negative),
+            positive: int_method!(positive),
+            absolute: int_method!(absolute),
+            boolean: int_method!(boolean),
+            invert: int_method!(invert),
+            lshift: int_method!(lshift),
+            rshift: int_method!(rshift),
```
youknowone, line 198: Not sure yet, but I think those fields in bool have to be `None`.
In my opinion, If that's not working with None, any inherited object of numbers can make similar problem.

qingshi163:
In RustPython PyBool is a subclass of PyInt
```
>>>>> isinstance(True, int)
True
```

vm/src/builtins/bool.rs
```diff
@@ -166,6 +174,48 @@ impl PyBool {
...
+            and: atomic_func!(|number, other, vm| {
+                PyBool::and(number.obj.to_owned(), other.to_owned(), vm).to_pyresult(vm)
+            }),
+            xor: atomic_func!(|number, other, vm| {
+                PyBool::xor(number.obj.to_owned(), other.to_owned(), vm).to_pyresult(vm)
+            }),
+            or: atomic_func!(|number, other, vm| {
+                PyBool::or(number.obj.to_owned(), other.to_owned(), vm).to_pyresult(vm)
+            }),
```
youknowone, line 207: Here is more room to optimize because and, xor, or actually doesn't require to increase reference count.
But it looks working fine.

To fully optimize it, `PyBool::and` can be changed like this and AsNumber can call `inner_and`
```rust
    #[pymethod(name = "__rand__")]
    #[pymethod(magic)]
    fn and(lhs: PyObjectRef, rhs: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
        Self::inner_and(&lhs, &rhs, vm)
    }
    fn inner_and(lhs: &PyObject, rhs: &PyObject, vm: &VirtualMachine) -> PyObjectRef {
        if lhs.fast_isinstance(vm.ctx.types.bool_type)
            && rhs.fast_isinstance(vm.ctx.types.bool_type)
        {
            let lhs = get_value(&lhs);
            let rhs = get_value(&rhs);
            (lhs && rhs).to_pyobject(vm)
        } else {
            get_py_int(&lhs).and(rhs, vm).to_pyobject(vm)
        }
    }
```

youknowone:
The inheritance is set like this:
```
#[pyclass(name = "bool", module = false, base = "PyInt")]  // this `PyInt` part
struct PyBool;
```

xiaozhiyan:
> In RustPython PyBool is a subclass of PyInt
> 
> ```
> >>>>> isinstance(True, int)
> True
> ```

I'm not sure, but does that mean `PyBool` will inherit existing methods of `PyInt` please?

youknowone:
Yes, `int.to_bytes` is a method of `int`.
```python
>>>>> int.to_bytes == bool.to_bytes
True
```

vm/src/builtins/bool.rs
xiaozhiyan: It seems `int` operations are needed for `bool` in CPython.
```
>>> (True + True) ** True
2
```

xiaozhiyan: Since `PyBool` is a subclass of `PyInt` and inherits its methods, so `(True + True) ** True` should somehow work without the above overriding in `as_number` I think.

youknowone: I am sorry to make confusion. I agree bool need int operations.
But I'd like to say manually filling `PyNumberMethods` is not the way to do. 
We can do that job only for builtin types written in Rust.
Making them `None` and finding a way to make it work will be helpful to solve the problem in general.

qingshi163: Did not test but I think PyBool is working with int number methods via mro search in PyNumber::find_methods().

But Maybe we can override it in PyBool to have some optimize

youknowone: In my opinion, because we will not lookup PyNumberMethods via mro, we need to fill it as @xiaozhiyan suggested.
But I'd like to avoid to manually copy them like this. This is better to be handled during inheritance procedure.

vm/src/builtins/bool.rs
xiaozhiyan: This approach seems not working directly, since `get_py_int(&lhs).and(rhs, vm)` requires `rhs` to be an `PyObjectRef` as defined in `int.rs`.

youknowone: maybe `rhs.to_owned()` fornow. we can optimize int.and also in same way later

xiaozhiyan: I'm not sure but would `rhs.to_owned()` be an extra burden? If it's ok (for now), I will adopt the above practice.

youknowone: a proper inheritance procedure would be possible using #4645

vm/src/builtins/bool.rs
youknowone: a proper inheritance procedure would be possible using #4645

youknowone: No, it wasn't. Because that is a static variable. We have no way to edit it later.

youknowone: @qingshi163 do you have any suggestion?

qingshi163: I think we can inherit slots when initializing the subclass

qingshi163: https://github.com/python/cpython/blob/71db5dbcd714b2e1297c43538188dd69715feb9a/Objects/typeobject.c#L8849-L8902

slots will be inherit if subclass does not override it.

when there is a change of slots on the type, cpython recusively update slots for all the subclasses.

youknowone: > slots will be inherit if subclass does not override it.

It is true in cpython but not here yet. That's the reason we always call `mro_find_map()` to find slots.

PR #4640, opened 2023-03-05
Update to pass test for unhashable collections

seungha-kim:
In CPython:

- If a type is marked as unhashable on the native side (i.e. `PyObject_HashNotImplemented` is assigned to hash slot), then `__hash__` attribute of the type dict is set to None so that the type is considered as unhashable on Python side too.
- If a type is created at runtime and `"__hash__": None` is given as the type dict (like `type('C', (object,), {'__hash__': None})`), then the hash slot is filled with PyObject_HashNotImplemented.

In RustPython, Unhashable trait exists to mark a type as unhashble. But there is no way to check if a type implements Unhashable trait at runtime, so it is hard to follow the CPython's mechanism.

So I've added `unhashable_wrapper` function similar to PyObject_HashNotImplemented, and used that as a marker for unhashable on the native side.

---

Thanks for the advice from @youknowone !

### DIFF
--- a/vm/src/builtins/bytearray.rs
+++ b/vm/src/builtins/bytearray.rs
@@ -30,9 +30,8 @@ use crate::{
     },
     sliceable::{SequenceIndex, SliceableSequenceMutOp, SliceableSequenceOp},
     types::{
-        hash_not_implemented, AsBuffer, AsMapping, AsNumber, AsSequence, Callable, Comparable,
-        Constructor, Initializer, IterNext, IterNextIterable, Iterable, PyComparisonOp,
-        Unconstructible,
+        AsBuffer, AsMapping, AsNumber, AsSequence, Callable, Comparable, Constructor, Initializer,
+        IterNext, IterNextIterable, Iterable, PyComparisonOp, Unconstructible,
     },
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyRef, PyResult, TryFromObject,
     VirtualMachine,
--- a/vm/src/builtins/bytearray.rs
+++ b/vm/src/builtins/bytearray.rs
@@ -41,7 +41,7 @@ use bstr::ByteSlice;
 use once_cell::sync::Lazy;
 use std::mem::size_of;
 
-#[pyclass(module = false, name = "bytearray")]
+#[pyclass(module = false, name = "bytearray", unhashable = true)]
 #[derive(Debug, Default)]
 pub struct PyByteArray {
     inner: PyRwLock<PyBytesInner>,
@@ -91,13 +91,6 @@ impl PyPayload for PyByteArray {
 
 /// Fill bytearray class methods dictionary.
 pub(crate) fn init(context: &Context) {
-    context
-        .types
-        .bytearray_type
-        .slots
-        .hash
-        .store(Some(unhashable_wrapper));
-
     PyByteArray::extend_class(context, context.types.bytearray_type);
     PyByteArrayIterator::extend_class(context, context.types.bytearray_iterator_type);
 }
--- a/vm/src/builtins/bytearray.rs
+++ b/vm/src/builtins/bytearray.rs
@@ -30,7 +30,7 @@ use crate::{
     },
     sliceable::{SequenceIndex, SliceableSequenceMutOp, SliceableSequenceOp},
     types::{
-        unhashable_wrapper, AsBuffer, AsMapping, AsNumber, AsSequence, Callable, Comparable,
+        hash_not_implemented, AsBuffer, AsMapping, AsNumber, AsSequence, Callable, Comparable,
         Constructor, Initializer, IterNext, IterNextIterable, Iterable, PyComparisonOp,
         Unconstructible,
     },
--- a/vm/src/builtins/bytearray.rs
+++ b/vm/src/builtins/bytearray.rs
@@ -30,9 +30,9 @@ use crate::{
     },
     sliceable::{SequenceIndex, SliceableSequenceMutOp, SliceableSequenceOp},
     types::{
-        AsBuffer, AsMapping, AsNumber, AsSequence, Callable, Comparable, Constructor, Hashable,
-        Initializer, IterNext, IterNextIterable, Iterable, PyComparisonOp, Unconstructible,
-        Unhashable,
+        unhashable_wrapper, AsBuffer, AsMapping, AsNumber, AsSequence, Callable, Comparable,
+        Constructor, Initializer, IterNext, IterNextIterable, Iterable, PyComparisonOp,
+        Unconstructible,
     },
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyRef, PyResult, TryFromObject,
     VirtualMachine,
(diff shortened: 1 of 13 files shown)


vm/src/builtins/list.rs
```diff
@@ -612,6 +602,13 @@ impl IterNext for PyListReverseIterator {
 }
 
 pub fn init(context: &Context) {
+    context
+        .types
+        .list_type
+        .slots
+        .hash
+        .store(Some(unhashable_wrapper));
```
seungha-kim, line 610: Maybe there is a more elegant way to do this...?

youknowone:
I will look in the details.
The single failing test is this one:
```
FAIL: test_hash (test.test_deque.TestBasic.test_hash)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/test_deque.py", line 537, in test_hash
    self.assertRaises(TypeError, hash, deque('abc'))
AssertionError: TypeError not raised by hash
```

vm/src/builtins/list.rs
seungha-kim: And I missed adding unhashable_wrapper thing for deque, so test_deque fails. Deque implementation resides in collections.rs, but there is no type initialization function similar to this `init` function in `collections.rs`. Where is a good place to put this code?

vm/src/builtins/bytearray.rs
```diff
@@ -91,6 +91,13 @@ impl PyPayload for PyByteArray {
 
 /// Fill bytearray class methods dictionary.
 pub(crate) fn init(context: &Context) {
+    context
+        .types
+        .bytearray_type
+        .slots
+        .hash
+        .store(Some(unhashable_wrapper));
```
youknowone, line 99: I have a suggestion about the location of this initialization.

We put these information to `PyClassDef` for builtin types.
adding `HASHABLE` const with default value true will be good.

Then we use the information during making class. that's `PyClassImpl::make_slots`.

vm/src/class.rs
```diff
@@ -112,6 +112,16 @@ pub trait PyClassImpl: PyClassDef {
...
         }
+
+        if class
+            .slots
+            .hash
+            .load()
+            .map(|h| h as usize == unhashable_wrapper as usize)
+            .unwrap_or(false)
+        {
+            class.set_attr(ctx.names.__hash__, ctx.none.clone().into());
```
youknowone, line 123: are you going to delete `__hash__` or setting `None` value to it?

```rust
class.set_attr(ctx.names.__hash__, ctx.none.clone().into());  // this is A.__class__.__hash__ = None
class.set_attr(ctx.names.__hash__, None);  // this is del A.__class__.__hash__

```

seungha-kim: Setting Python-side `None` value.

If you delete `__hash__`, the class will be considered as Hashable (i.e. `isinstance(list(), Hashable) is True`) which is what we don't want.

vm/src/builtins/bytearray.rs
seungha-kim: Thanks, I'll try.

vm/src/class.rs
youknowone: Thanks, I found even my comment to delete is wrong too 😂

vm/src/builtins/bytearray.rs
youknowone: maybe default value is false, not true. hashable is common in builtin type world, but not outside.

vm/src/class.rs
```diff
@@ -112,6 +112,16 @@ pub trait PyClassImpl: PyClassDef {
                     .into();
             class.set_attr(identifier!(ctx, __new__), bound);
         }
+
+        if class
+            .slots
+            .hash
+            .load()
+            .map(|h| h as usize == unhashable_wrapper as usize)
+            .unwrap_or(false)
```
youknowone, line 121:
```suggestion
            .map_or(0, |h| h as usize) == unhashable_wrapper as usize
```
the cost of `map_or(0, |h| h as usize)` is zero here due to its layout.

vm/src/class.rs
```diff
@@ -112,6 +112,16 @@ pub trait PyClassImpl: PyClassDef {
                     .into();
             class.set_attr(identifier!(ctx, __new__), bound);
         }
+
+        if class
```
youknowone, line 116: I'd like to check if this is related to #3460 later

vm/src/types/slot.rs
```diff
@@ -414,7 +419,17 @@ impl PyType {
                 update_pointer_slot!(as_mapping, mapping_methods);
             }
             _ if name == identifier!(ctx, __hash__) => {
-                toggle_slot!(hash, hash_wrapper);
+                let is_unhashable = self
+                    .attributes
+                    .read()
+                    .get(identifier!(ctx, __hash__))
+                    .map(|a| a.is(&ctx.none()))
+                    .unwrap_or(false);
```
youknowone, line 427:
```suggestion
                    .map_or(false, |a| a.is(&ctx.none));
```

`ctx.none()` increase reference count of `None`. We don't need to do.

vm/src/types/slot.rs
```diff
@@ -414,7 +419,17 @@ impl PyType {
...
+                if is_unhashable {
+                    toggle_slot!(hash, unhashable_wrapper);
+                } else {
+                    toggle_slot!(hash, hash_wrapper);
+                }
```
youknowone, line 432:
```suggestion
                let wrapper = if is_unhashable {
                    unhashable_wrapper
                } else {
                    hash_wrapper
                };
                toggle_slot!(wrapper)
```

youknowone (changes requested):
I added a few more style suggestions.

vm/src/class.rs
```diff
@@ -60,6 +60,7 @@ pub trait PyClassDef {
     const TP_NAME: &'static str;
     const DOC: Option<&'static str> = None;
     const BASICSIZE: usize;
+    const UNHASHABLE: bool;
```
youknowone, line 63:
```suggestion
    const UNHASHABLE: bool = false;
```
default value can be placed here

vm/src/class.rs
```diff
@@ -71,6 +72,7 @@ where
     const TP_NAME: &'static str = T::TP_NAME;
     const DOC: Option<&'static str> = T::DOC;
     const BASICSIZE: usize = T::BASICSIZE;
+    const UNHASHABLE: bool = false;
```
youknowone, line 75: `PyRef<T>` must follows `T`

```suggestion
    const UNHASHABLE: bool = T::UNHASHABLE;
```

youknowone (commented):
Great! it now passes deque too.

vm/src/types/slot.rs
```diff
@@ -243,6 +243,11 @@ fn hash_wrapper(zelf: &PyObject, vm: &VirtualMachine) -> PyResult<PyHash> {
     }
 }
 
+/// Marks a type as unhashable. Similar to PyObject_HashNotImplemented in CPython
+pub fn unhashable_wrapper(zelf: &PyObject, vm: &VirtualMachine) -> PyResult<PyHash> {
```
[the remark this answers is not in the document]
youknowone: Oh, probably `hash_not_implemented` by following cpython naming

youknowone (approved):
Looks great, thank you for digging in this tricky issue.

Issue #4641, opened 2023-03-05
Fix `rsub` of `PySet` and `PyFrozenSet`


xiaozhiyan:
In the following example, `__rsub__` incorrectly returns the same output of `__sub__`, which should be `{3}` instead.
```
>>>>> set([1, 2]).__sub__(set([2, 3]))
{1}
>>>>> set([1, 2]).__rsub__(set([2, 3]))
{1}
```

PR #4642, opened 2023-03-05
Fix `rsub` of `PySet` and `PyFrozenSet`

xiaozhiyan:
Closes https://github.com/RustPython/RustPython/issues/4641


xiaozhiyan:
@youknowone The implementation in this PR seems not correct. Could you help to have a look please?

vm/src/builtins/set.rs
```diff
@@ -605,7 +605,13 @@ impl PySet {
...
     #[pymethod(magic)]
     fn rsub(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult<PyArithmeticValue<Self>> {
-        self.sub(other, vm)
+        PySet::number_downcast(other.to_number()).sub(
+            PySet {
+                inner: self.inner.to_owned(),
+            }
+            .into_pyobject(vm),
+            vm,
+        )
```
youknowone, line 614: sub and rsub shares many things. the implementation will be similar.

```suggestion
        if let Ok(other) = AnySet::try_from_object(vm, other) {
            Ok(PyArithmeticValue::Implemented(other.op(
                self,
                PySetInner::difference,
                vm,
            )?))
        } else {
            Ok(PyArithmeticValue::NotImplemented)
        }
```

youknowone: Set implementation looks not ideal. and there was working unfinished PRs also.

youknowone: Otherwise, would this work?
```suggestion
        if let Ok(other) = AnySet::try_from_object(vm, other) {
            Ok(PyArithmeticValue::Implemented(other.inner.difference(self, vm))?))
        } else {
            Ok(PyArithmeticValue::NotImplemented)
        }
```

youknowone:
Could you also add the test in issue to `extra_tests/snippets/builtin_set.py`?

vm/src/builtins/set.rs
youknowone: This version will work
```rust
    #[pymethod(magic)]
    fn rsub(
        zelf: PyRef<Self>,
        other: PyObjectRef,
        vm: &VirtualMachine,
    ) -> PyResult<PyArithmeticValue<Self>> {
        Ok(if let Ok(other) = AnySet::try_from_object(vm, other) {
            PyArithmeticValue::Implemented(Self {
                inner: other
                    .as_inner()
                    .difference(ArgIterable::try_from_object(vm, zelf.into())?, vm)?,
            })
        } else {
            PyArithmeticValue::NotImplemented
        })
    }
```

youknowone:
I made a report about missing test https://github.com/python/cpython/issues/102463

vm/src/builtins/set.rs
xiaozhiyan: The above style has been adopted in the PR.

youknowone (approved):

PR #4646, opened 2023-03-06
Add sys.winver

### LINKED ISSUE #4550: Add sys.winver
Python documentation: https://docs.python.org/3/library/sys.html#sys.winver

It is DLL version of Python.
Though we don't have DLL library, it is required to load `site` correctly.

By the document, it is same as python `<major>.<minor>` version in most of cases.

we are defining them under `vm/src/versions.rs`

dalinaum:
Resolve #4550

With this patch, I got the same string as with CPython's `sys.winver`.

### DIFF
--- a/vm/src/stdlib/sys.rs
+++ b/vm/src/stdlib/sys.rs
@@ -262,6 +262,13 @@ mod sys {
         version::get_version()
     }
 
+    #[cfg(windows)]
+    #[pyattr]
+    fn winver(_vm: &VirtualMachine) -> String {
+        // Note: This is Python DLL version in CPython, but we arbitrary fill it for compatibility
+        version::get_winver_number()
+    }
+
     #[pyattr]
     fn _xoptions(vm: &VirtualMachine) -> PyDictRef {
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/sys.rs
```diff
@@ -262,6 +262,12 @@ mod sys {
         version::get_version()
     }
 
+    #[cfg(windows)]
+    #[pyattr]
+    fn winver(_vm: &VirtualMachine) -> String {
+        version::get_winver_number()
```
youknowone, line 268:
```suggestion
        // Note: This is Python DLL version in CPython, but we arbitrary fill it for compatibility
        version::get_winver_number()
```
Because this is not the implementation by specification, adding a note will be helpful.

youknowone (changes requested):

vm/src/stdlib/sys.rs
dalinaum: Done

youknowone (approved):
Thank you for contributing!

PR #4648, opened 2023-03-06
Update test/test_with.py from CPython 3.11.2

dalinaum:
#4564
Tests related to python/cpython#26809 have been marked as failing tests.


fanninpm (approved):
LGTM.

@youknowone we should probably follow up on this change in error type.

Issue #4629, opened 2023-03-05
Remove PyPayload::special_retrieve and replace to `ArgIndex`


youknowone:
`PyPayload::special_retrieve` is a wrong design because we can't distinguish real `int` and int created by `__index__`.

See `ArgIntoBool`, `ArgIntoFloat` and `ArgIntoComplex` for arguments calling type conversion magic method.
The name will be `ArgIndex` or `ArgIntoIndex`. I am not sure about the names.

The final step will be dividing original `PyIntRef` to `PyIntRef` and `ArgIndex`. This step will take time.

PR #4651, opened 2023-03-07
Add ArgIndex according to #4629

minhrongcon2000:
First step to solve #4629


vm/src/function/number.rs
```diff
@@ -125,3 +126,32 @@ impl TryFromObject for ArgIntoBool {
...
+    }
+}
+
+impl TryFromObject for ArgIndex {
+    fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
+        Ok(Self {
+            value: obj.try_index(vm)?.as_bigint().clone(),
+        })
+    }
+}
```
youknowone, line 157: ...
using `TryFromBorrowedObject`. Either way we can avoid an additional clone.
```rust
impl TryFromBorrowedObject for ArgIndex {
    ...
```

youknowone (changes requested):

youknowone (approved):
Looks good, thank you!

PR #4652, opened 2023-03-07
Add locale implementation for windows

minhrongcon2000:
Implement locale for windows


stdlib/Include/liblocale/liblocale.c
```diff
@@ -0,0 +1 @@
+#include<locale.h>
```
youknowone, line 1: I am trying to understand how are you going to solve it.
If this file only includes `<locale.h>`, what does cmake do here?
what kind of symbols will `liblocale` obtains?

youknowone: If you only need the single function, you can manually define an extern function.
https://doc.rust-lang.org/std/keyword.extern.html

If you need many symbols, using [bindgen](https://github.com/rust-lang/rust-bindgen) is another way to generate it.

stdlib/Cargo.toml
```diff
@@ -115,4 +115,4 @@ features = [
 ]
 
 [target.'cfg(target_os = "macos")'.dependencies]
-system-configuration = "0.5.0"
+system-configuration = "0.5.0"
```
youknowone, line 118: unnecessary changes

```suggestion
system-configuration = "0.5.0"
```

stdlib/src/locale.rs
```diff
@@ -15,17 +23,64 @@ mod _locale {
...
+        D_FMT, D_T_FMT, ERA, ERA_D_FMT, ERA_D_T_FMT, ERA_T_FMT, LC_MESSAGES, MON_1, MON_10, MON_11,
+        MON_12, MON_2, MON_3, MON_4, MON_5, MON_6, MON_7, MON_8, MON_9, NOEXPR, PM_STR, RADIXCHAR,
+        THOUSEP, T_FMT, T_FMT_AMPM, YESEXPR,
     };
 
+    #[cfg(any(
+        unix,
+        windows,
+        not(any(target_os = "ios", target_os = "android", target_arch = "wasm32"))
+    ))]
```
youknowone, line 75: The module already seems to have this cfg
```suggestion
```

stdlib/src/locale.rs
```diff
@@ -67,76 +127,144 @@ mod _locale {
...
     #[pyfunction]
     fn strxfrm(string: PyStrRef, vm: &VirtualMachine) -> PyResult {
         // https://github.com/python/cpython/blob/eaae563b6878aa050b4ad406b67728b6b066220e/Modules/_localemodule.c#L390-L442
         let n1 = string.byte_len() + 1;
-        let mut buff: Vec<u8> = vec![0; n1];
+        let mut buff = vec![0u8; n1];
```
youknowone, line 151: 👍

stdlib/src/locale.rs
```diff
@@ -67,76 +127,144 @@ mod _locale {
...
+            libc::strxfrm(buff.as_mut_ptr() as _, cstr.as_ptr(), n2 + 1);
+        }
         Ok(vm.new_pyobj(String::from_utf8(buff).expect("strxfrm returned invalid utf-8 string")))
     }
 
-    #[pyfunction]
-    fn localeconv(vm: &VirtualMachine) -> PyResult<PyDictRef> {
+    #[cfg(all(windows, not(target_arch = "wasm32")))]
+    #[pyfunction(name = "localeconv")]
+    fn _localeconv_w(vm: &VirtualMachine) -> PyResult<PyDictRef> {
```
youknowone, line 164: `localeconv_w` and `localeconv_u` seems same function except for the i. is it? in that case, can we share a single `localeconv` for both case?

stdlib/src/locale.rs
```diff
@@ -15,17 +23,64 @@ mod _locale {
...
+        int_p_sep_by_space: libc::c_char,
+        int_n_sep_by_space: libc::c_char,
+        int_p_sign_posn: libc::c_char,
+        int_n_sign_posn: libc::c_char,
+    }
+
+    #[cfg(windows)]
+    extern "C" {
+        fn localeconv() -> *mut lconv;
+    }
```
youknowone, line 58: i'd like to put this function out of this module. Then it can be called like
```rust
unsafe { super::localeconv() }
```
below

youknowone (changes requested):

stdlib/src/locale.rs
minhrongcon2000: so you mean this function is in a `common` module so that it can be used in other module?

youknowone (approved):
Thank you!

PR #4654, opened 2023-03-07
Add protocol object PyCallable

youknowone:
#3244

### DIFF
--- a/vm/src/protocol/callable.rs
+++ b/vm/src/protocol/callable.rs
@@ -1,5 +1,5 @@
 use crate::{
-    function::IntoFuncArgs,
+    function::{FuncArgs, IntoFuncArgs},
     types::GenericMethod,
     {AsObject, PyObject, PyResult, VirtualMachine},
 };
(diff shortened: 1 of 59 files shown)


vm/src/protocol/callable.rs
```diff
@@ -0,0 +1,97 @@
...
+};
+
+impl PyObject {
+    #[inline]
+    pub fn to_callable(&self) -> Option<PyCallable<'_>> {
+        PyCallable::new(self)
+    }
+
+    #[inline]
+    pub fn is_callable(&self) -> bool {
```
youknowone, line 14: The new `is_callable` doesn't require vm argument anymore

vm/src/protocol/callable.rs
```diff
@@ -0,0 +1,97 @@
...
+        vm.trace_event(TraceEvent::Return)?;
+        result
+    }
+}
+
+/// Trace events for sys.settrace and sys.setprofile.
+enum TraceEvent {
+    Call,
+    Return,
+}
```
youknowone, line 42: relocated `TraceEvent` because it is use only to call functions now

qingshi163 (approved):
Great

Issue #4650, opened 2023-03-07
Investigate the `with` statement changes related to context manager


youknowone:
introduced from #4648 and #4649

dalinaum:
I took a quick look, made the following patch, and it worked for `test_with`, but not for `test_contextlib`. I'll look into it a bit more.
dalinaum@f4b51e299f4dcaef3d72c3a16e7b46f2eacad9fd

youknowone:
done by @dalinaum on #4656 and #4659

PR #4656, opened 2023-03-07
Change error type for bad objects in "with" and "async with"

dalinaum:
This patch does not cover all cases of `test_contextlib.py`

`bytecode::Instruction::SetupWith` and `BeforeAsyncWith` raise a TypeError instead of an AttributeError.

It relates to #4650

### DIFF
--- a/vm/src/frame.rs
+++ b/vm/src/frame.rs
@@ -840,12 +840,24 @@ impl ExecutingFrame<'_> {
             }
             bytecode::Instruction::SetupWith { end } => {
                 let context_manager = self.pop_value();
-                let enter_res = vm.call_special_method(
-                    context_manager.clone(),
-                    identifier!(vm, __enter__),
-                    (),
-                )?;
-                let exit = context_manager.get_attr(identifier!(vm, __exit__), vm)?;
+                let error_string = || -> String {
+                    format!(
+                        "'{:.200}' object does not support the context manager protocol",
+                        context_manager.class().name(),
+                    )
+                };
+                let enter_res = vm
+                    .get_special_method(context_manager.clone(), identifier!(vm, __enter__))?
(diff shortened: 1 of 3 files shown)


vm/src/frame.rs
```diff
@@ -840,12 +840,25 @@ impl ExecutingFrame<'_> {
...
-                let enter_res = vm.call_special_method(
-                    context_manager.clone(),
-                    identifier!(vm, __enter__),
-                    (),
-                )?;
-                let exit = context_manager.get_attr(identifier!(vm, __exit__), vm)?;
+                let error_string = || -> String {
+                    format!(
+                        "'{:.*}' object does not support the context manager protocol",
+                        200,
```
dalinaum, line 846: I just realized that CPython tends to use `%.200s`  in many strings. Would it be okay to replace `{:.*}` with `{}`?

youknowone: Your versions looks better.

vm/src/frame.rs
```diff
@@ -840,12 +840,25 @@ impl ExecutingFrame<'_> {
...
+                        context_manager.class().name(),
+                    )
```
youknowone, line 848:
```suggestion
                    format!(
                        "'{:200}' object does not support the context manager protocol",
                        context_manager.class().name(),
                    )
```
is this the same thing?

youknowone (approved):
contextlib seemed to related to contextlib.py version.
so this patch is fully fixing the with changes in 3.11. right?

dalinaum:
@youknowone #4659 fixed 3 tests in text_conextlib.py and this patch fixed 2 tests.

However, there are still `expectedFailure`s that were added during updating test_contextlib.py and contextlib.py.

youknowone:
at least, they are not regressions ;)

youknowone:
Thank you for contributing!

PR #4666, opened 2023-03-08
Update test/test_userlist.py and test_userdict.py from CPython 3.11.2

dalinaum:
In the case of test_userdict.py file only white space is different.
#4564


youknowone:
let's see if the windows test passes

youknowone (approved):
👍

Issue #4655, opened 2023-03-07
array subclasses doesn't allow kwargs


youknowone:
Related test: `test_array::BaseTest::test_subclass_with_kwargs`

As I understand,

`array.__new__`  allows kwargs
`array.__init__ ` prohibit kwargs

But we only implements `Constructor` for `PyArray` which corresponds to `__new__`.
Adding `Initializer` for `__init__` will be a starting point to solve this issue.

minhrongcon2000:
I will work on this issue!

JazzGlobal:
Is this resolved?

youknowone:
Yes, this is done! Thank you for reminding!

PR #4667, opened 2023-03-08
Add init for array.rs

minhrongcon2000:
First step for #4655

### DIFF
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -47,7 +47,9 @@ mod array {
             },
             class_or_notimplemented,
             convert::{ToPyObject, ToPyResult, TryFromObject},
-            function::{ArgBytesLike, ArgIntoFloat, ArgIterable, OptionalArg, PyComparisonValue},
+            function::{
+                ArgBytesLike, ArgIntoFloat, ArgIterable, KwArgs, OptionalArg, PyComparisonValue,
+            },
             protocol::{
                 BufferDescriptor, BufferMethods, BufferResizeGuard, PyBuffer, PyIterReturn,
                 PyMappingMethods, PySequenceMethods,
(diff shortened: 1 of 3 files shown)


stdlib/src/array.rs
```diff
@@ -649,16 +649,35 @@ mod array {
 
         fn py_new(
             cls: PyTypeRef,
-            Self::Args { spec, init }: Self::Args,
+            Self::Args { spec, .. }: Self::Args,
```
youknowone, line 652: you can try `FuncArgs`, `PosArgs` and `KwArgs` to make compatible interface.

stdlib/src/array.rs
```diff
@@ -645,20 +647,48 @@ mod array {
...
+                })?;
+            let array = ArrayContentType::from_char(spec).map_err(|err| vm.new_value_error(err))?;
+
+            let zelf = Self::from(array).into_ref_with_type(vm, cls)?;
+            Ok(zelf.into())
+        }
+    }
+
+    impl Initializer for PyArray {
         type Args = ArrayNewArgs;
```
youknowone, line 683:
```suggestion
        type Args = ArrayInitArgs;
```

stdlib/src/array.rs
```diff
@@ -645,20 +647,48 @@ mod array {
     }
 
     impl Constructor for PyArray {
+        type Args = FuncArgs;
```
youknowone, line 650:
```suggestion
        type Args = (PyStrRef, FuncArgs);
```
maybe this will work.

looking for other `fn py_new` may give more idea.

minhrongcon2000: Thank you! This is a really good suggestion :D

youknowone:
Did you remove Initializer?

I think this case is now broken

```
>>> from array import array
>>> array.__new__(array, 'i', [], kwarg='something')
TypeError: array.array() takes no keyword arguments
```

youknowone:
oh.. it is weird. maybe we need exact type checking from Constructor

```python
>>> from array import array
>>> array.__new__(array, 'i', [], kwarg='something')
TypeError: array.array() takes no keyword arguments
>>> class A(array): pass
... 
>>> array.__new__(A, 'i', [], kwarg='something')
A('i')
>>> 
```

youknowone:
how about trying `type Args = (ArrayNewArgs, FuncArgs);` and ignore FuncArgs part only when the type is exactly array?

youknowone:
oh, probably `type Args = (ArrayNewArgs, KwArgs);`

youknowone (approved):
Looks good, thank you!

PR #4673, opened 2023-03-09
Pin our toolchain to Rust version 1.67.1

youknowone:
it will be better than only showing it on README and getting continuous same issues.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,6 +6,7 @@ name = "rustpython"
 version = "0.2.0"
 authors = ["RustPython Team"]
 edition = "2021"
+rust-version = "1.67.1"
 description = "A python interpreter written in rust."
 repository = "https://github.com/RustPython/RustPython"
 license = "MIT"


xiaozhiyan:
@youknowone 
Is the CI error in WASM related to the current PR?
https://github.com/RustPython/RustPython/actions/runs/4373906862/jobs/7652708971

Should we specify Rust version in `rust-toolchain` instead of `Cargo.toml`?
The current config may only apply to the main package, while not affecting other packages in the workspace.

xiaozhiyan:
I see. Then I'll ignore the CI error for WASM.

How about the location of setting Rust version?
> Should we specify Rust version in rust-toolchain instead of Cargo.toml?
The current config may only apply to the main package, while not affecting other packages in the workspace.

xiaozhiyan:
I think https://github.com/RustPython/RustPython/blob/main/rust-toolchain may be the convenient place to specify Rust version. Otherwise we'd better remove the `rust-toolchain` file to avoid version conflict in the future.

xiaozhiyan:
I think `rust-toolchain.toml` may be a suggested way to specify Rust version of a project (which may include multiple packages), which is respected by Cargo and is not specific to particular CI tool or script.
https://rust-lang.github.io/rustup/overrides.html#the-toolchain-file

Btw, Rust 1.68.0 was just officially released, but maybe it would not cause compatibility issue with current config.

DimitrisJim:
Ah, it does get respected but it doesn't fail. Instead it automatically uses the appropriate version if it already exists or does a download of the more recent version automatically. 

I'd personally prefer to let the user make that decision on their own after getting a build error but I could also see how this would work.

PR #4679, opened 2023-03-09
[Draft] Biased Reference Count

qingshi163:


fanninpm:
I'm curious about why you're going with bit shifting when there's a risk of "underflow".

qingshi163:
In my local benchmark, this pr is 10% slower than the original atomic reference counting. something I was missing?

PR #4683, opened 2023-03-12
Update test/test_types from CPython 3.11.2

tdub0:
[#4564](https://github.com/RustPython/RustPython/issues/4564)
Changes from CPython v3.11.2 produce a single failure on a new test `test_dunder_get_signature`. There seems to be no signature found in builtin method `__get__` of 'method_descriptor' objects, so it raises a ValueError. I added the `expectedFailure` decorator to that function.

### DIFF
--- a/Lib/test/test_types.py
+++ b/Lib/test/test_types.py
@@ -602,6 +602,8 @@ class TypesTests(unittest.TestCase):
         self.assertIsInstance(object.__lt__, types.WrapperDescriptorType)
         self.assertIsInstance(int.__lt__, types.WrapperDescriptorType)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_dunder_get_signature(self):
         sig = inspect.signature(object.__init__.__get__)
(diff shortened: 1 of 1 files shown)


Lib/test/test_types.py
```diff
@@ -604,6 +602,15 @@ def test_slot_wrapper_types(self):
         self.assertIsInstance(object.__lt__, types.WrapperDescriptorType)
         self.assertIsInstance(int.__lt__, types.WrapperDescriptorType)
 
+    # TODO: No signature found in builtin method __get__ of 'method_descriptor'
+    # objects so ValueError is raised
```
DimitrisJim, line 606: Nit, we append `RUSTPYTHON` here to be able to differentiate if this is our `TODO` or not. 

```suggestion
    # TODO: RUSTPYTHON No signature found in builtin method __get__ of 'method_descriptor' objects.
```

tdub0: Thanks for reviewing this. I will fix that.

DimitrisJim (approved):
thanks for this!

PR #4686, opened 2023-03-13
[Draft] Make builtin method type same as that of builtin function

### LINKED ISSUE #3100: `BuiltinFunctionType` and `BuiltinMethodType` are different
## Feature

`types.BuiltinFunctionType` must be equal to `types.BuiltinMethodType`.
In other word, types returned by builtin function and types returned by builtin method must be of the same type.

### In CPython
```py
>>> import types
>>> types.BuiltinFunctionType == types.BuiltinMethodType
True
>>> type(len) == type([].append)
True
```

### In RustPython
```py
>>>>> import types
>>>>> types.BuiltinFunctionType == types.BuiltinMethodType
False
>>>>> type(len) == type([].append)
False
```

## Python Documentation

[BuiltinFunctionType](https://docs.python.org/3/library/types.html#types.BuiltinFunctionType) and [BuiltinMethodType](https://docs.python.org/3/library/types.html#types.BuiltinMethodType) have same descriptions.
Also, commented at [source code](https://github.com/python/cpython/blob/050d1035957379d70e8601e6f5636637716a264b/Lib/types.py#L44).

seungha-kim:
Fixes #3100

### DIFF
--- a/vm/src/builtins/builtin_func.rs
+++ b/vm/src/builtins/builtin_func.rs
@@ -1,6 +1,5 @@
 use super::{type_, PyClassMethod, PyStaticMethod, PyStr, PyStrInterned, PyStrRef, PyType};
 use crate::{
-    builtins::PyBoundMethod,
     class::PyClassImpl,
     function::{FuncArgs, IntoPyNativeFunc, PyNativeFunc},
     types::{Callable, Constructor, GetDescriptor, Representable, Unconstructible},
@@ -8,14 +7,15 @@ use crate::{
 };
 use std::fmt;
 
+#[derive(Clone)]
 pub struct PyNativeFuncDef {
-    pub func: PyNativeFunc,
+    pub func: &'static PyNativeFunc,
     pub name: &'static PyStrInterned,
     pub doc: Option<PyStrRef>,
 }
 
 impl PyNativeFuncDef {
-    pub fn new(func: PyNativeFunc, name: &'static PyStrInterned) -> Self {
+    pub fn new(func: &'static PyNativeFunc, name: &'static PyStrInterned) -> Self {
         Self {
             func,
             name,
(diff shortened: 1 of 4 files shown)


vm/src/function/builtin.rs
```diff
@@ -3,9 +3,10 @@ use crate::{
     convert::ToPyResult, object::PyThreadingConstraint, PyPayload, PyRef, PyResult, VirtualMachine,
 };
 use std::marker::PhantomData;
+use std::sync::Arc;
 
 /// A built-in Python function.
-pub type PyNativeFunc = Box<py_dyn_fn!(dyn Fn(&VirtualMachine, FuncArgs) -> PyResult)>;
+pub type PyNativeFunc = Arc<py_dyn_fn!(dyn Fn(&VirtualMachine, FuncArgs) -> PyResult)>;
```
seungha-kim, line 9: I had to do this to make PyNativeFuncDef clonable, but actually it is a very wide change and can cause performance issues. (Arc::clone is called whenever calling native methods.)

youknowone: what if we keep PyFunction owns `Box<>` but PyBoundMethod holds `&'static`?

seungha-kim: Oh, I got what you meant. Maybe I can use Box::leak?

seungha-kim:
I missed that PyClassMethod must be considered too. Turning this PR to draft now and will look into PyClassMethod.

vm/src/function/builtin.rs
youknowone: not 100% sure, but it can be one of the way. do we always use static functions, never clousure? In that case, `leak` is one of the way.
I was thinking giving unbound method (e.g. `object.__repr__`) Box but bound method (`object().__repr__` reference)

youknowone: ah, i was totally not getting how this is going.

youknowone: @qingshi163 @DimitrisJim @coolreader18 any idea about this?

vm/src/builtins/builtin_func.rs
```diff
@@ -29,14 +29,36 @@ impl PyNativeFuncDef {
     }
 
     pub fn into_function(self) -> PyBuiltinFunction {
-        self.into()
+        PyBuiltinFunction {
+            zelf: None,
+            value: self,
+            module: None,
+            is_classmethod: false,
```
youknowone, line 36: I am not sure about this field

youknowone: ah, i got it. we don't have `PyMethodDef`. So we cannot avoid it now.

youknowone:
I made #4873 including this patch.
It took almost a month 😵

PR #4687, opened 2023-03-13
Macro stuff

coolreader18:
This is sorta a work in progress that's ready for review, mainly since I'm still not sure whether parallelizing `py_freeze!` is actually a good idea. My original idea was to just output `py_compile!` invocations from `freeze` and let rustc parallelize it for us, but it turns out it doesn't actually run proc macros in parallel, so.


youknowone:
@coolreader18 I like this patch so much even without parallel parts. Could you split this PR to 2 chunks one for refactoring and the other one for parallel compilation?

coolreader18:
Ough now I wanna use `syn` 2's meta parsing api instead but I'd have to update like `syn-ext` and stuff too

youknowone:
now we use syn2. will this patch be revived?

PR #4688, opened 2023-03-13
Update _compression, gzip, and test_gzip for CPython v3.11.2

tdub0:
Updated `_compression.py` to match  CPython v3.11.2. I didn't see any immediate unit test for this, so it seemed fit to test using a library that relies on `_compression.py`. I updated `gzip.py` / `test_gzip.py` and accounted for the unexpected failures on `test_gzip.py`. 

The failures were primarily wbit errors tracing to `stdlib/src/zlib.rs`, but I am not familiar enough with this module in rust to correct the issues.

### DIFF
--- a/Lib/_compression.py
+++ b/Lib/_compression.py
@@ -1,7 +1,7 @@
 """Internal classes used by the gzip, lzma and bz2 modules"""
 
 import io
-
+import sys
 
 BUFFER_SIZE = io.DEFAULT_BUFFER_SIZE  # Compressed data read chunk size
 
@@ -109,6 +109,16 @@ class DecompressReader(io.RawIOBase):
             return b""
         self._pos += len(data)
         return data
+
+    def readall(self):
+        chunks = []
+        # sys.maxsize means the max length of output buffer is unlimited,
+        # so that the whole input buffer can be decompressed within one
+        # .decompress() call.
+        while data := self.read(sys.maxsize):
+            chunks.append(data)
+
+        return b"".join(chunks)
 
     # Rewind the file to the beginning of the data stream.
     def _rewind(self):
--- a/Lib/gzip.py
+++ b/Lib/gzip.py
@@ -399,6 +399,59 @@ class GzipFile(_compression.BaseStream):
         return self._buffer.readline(size)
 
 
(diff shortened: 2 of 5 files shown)


tdub0:
See [#4564](https://github.com/RustPython/RustPython/issues/4564).

Lib/_compression.py
```diff
@@ -1,7 +1,7 @@
 """Internal classes used by the gzip, lzma and bz2 modules"""
```
youknowone, line 1: by looking the header, bz2 also seems to depending on here. could you also update it?

Lib/_compression.py
```diff
@@ -110,6 +110,16 @@ def read(self, size=-1):
         self._pos += len(data)
         return data
 
+    def readall(self):
+        chunks = []
+        # sys.maxsize means the max length of output buffer is unlimited,
+        # so that the whole input buffer can be decompressed within one
+        # .decompress() call.
+        while data := self.read(sys.maxsize):
```
youknowone, line 118: maybe we don't support this operator yet

youknowone (commented):
That's really good decision 😄
thank you!

Lib/_compression.py
fanninpm: We have no `bz2` rust library, unfortunately.

tdub0: Yeah, I hit the `SkipTest: No module named ' bz2'` when I copied over and ran the `test_bz2.py` as there were no changes in the `Lib/bz2.py`. Would this be connected to the [bz2.rs](https://github.com/RustPython/RustPython/blob/main/stdlib/src/bz2.rs) in `stdlib/src/`

Lib/_compression.py
tdub0: I did see a reference on [L84](https://github.com/RustPython/RustPython/blob/main/compiler/ast/src/unparse.rs#L84) of `compiler/ast/src/unparse.rs` but it might not be handling it as the assignment operator described in [PEP 572](https://peps.python.org/pep-0572/).

youknowone (approved):
looks good, thank you!

Issue #4653, opened 2023-03-07
Implement tp_repr and tp_str


youknowone:
part of #3244 

The slot field to start: https://github.com/RustPython/RustPython/blob/7744e95aea79179f3dc31f9c3ab46e22fd3b5cbf/vm/src/types/slot.rs#L47

In Python side, `str(1)` hits this line :
https://github.com/RustPython/RustPython/blob/7744e95aea79179f3dc31f9c3ab46e22fd3b5cbf/vm/src/builtins/str.rs#L281

`input.str(vm)` in the line is calling rust side rust side `PyObjectRef::str` which is:
https://github.com/RustPython/RustPython/blob/7744e95aea79179f3dc31f9c3ab46e22fd3b5cbf/vm/src/protocol/object.rs#L331-L349

And `object.__str__`, which every other object's `__str__` inherits as default is:
https://github.com/RustPython/RustPython/blob/7744e95aea79179f3dc31f9c3ab46e22fd3b5cbf/vm/src/builtins/object.rs#L184-L188

Once this is done, a wrapper need to be added to `update_slot!` to connect python side defined magic methods to rust type slots.
https://github.com/RustPython/RustPython/blob/7744e95aea79179f3dc31f9c3ab46e22fd3b5cbf/vm/src/types/slot.rs#L349

Once those are finished, we can actually replace the implementation of other types' `__str__` to this new implementation
We typically did it using trait. 
One of the good example is Hashable: https://github.com/RustPython/RustPython/blob/7744e95aea79179f3dc31f9c3ab46e22fd3b5cbf/vm/src/types/slot.rs#L645-L664

And the adaption is looking like this: 
https://github.com/RustPython/RustPython/blob/7744e95aea79179f3dc31f9c3ab46e22fd3b5cbf/vm/src/builtins/complex.rs#L439-L453

haxelion:
@youknowone Please assign me 🙂

arihant2math:
Completed, and superseded by #3244.

PR #4689, opened 2023-03-14
Implemented repr slot

haxelion:
With respect to issue #4653, this PR contains the implementation of the slot for `repr`.

### DIFF
--- a/vm/src/protocol/object.rs
+++ b/vm/src/protocol/object.rs
@@ -321,7 +321,7 @@ impl PyObject {
                 Some(slot) => slot(self, vm),
                 None => vm
                     .call_special_method(self.to_owned(), identifier!(vm, __repr__), ())?
-                    .try_into_value(vm),
+                    .try_into_value(vm), // TODO: remove magic method call once __repr__ is fully ported to slot
             }
         })
     }
--- a/vm/src/protocol/object.rs
+++ b/vm/src/protocol/object.rs
@@ -317,8 +317,12 @@ impl PyObject {
 
     pub fn repr(&self, vm: &VirtualMachine) -> PyResult<PyStrRef> {
         vm.with_recursion("while getting the repr of an object", || {
-            let repr = vm.call_special_method(self.to_owned(), identifier!(vm, __repr__), ())?;
-            repr.try_into_value(vm)
+            match self.class().slots.repr.load() {
+                Some(slot) => slot(self, vm),
+                None => vm
+                    .call_special_method(self.to_owned(), identifier!(vm, __repr__), ())?
+                    .try_into_value(vm),
+            }
         })
     }
 
--- a/vm/src/types/slot.rs
+++ b/vm/src/types/slot.rs
@@ -1,5 +1,7 @@
 use crate::{
-    builtins::{type_::PointerSlot, PyFloat, PyInt, PyStrInterned, PyStrRef, PyType, PyTypeRef},
+    builtins::{
+        type_::PointerSlot, PyFloat, PyInt, PyStr, PyStrInterned, PyStrRef, PyType, PyTypeRef,
+    },
     bytecode::ComparisonOperator,
     common::{hash::PyHash, lock::PyRwLock},
     convert::{ToPyObject, ToPyResult},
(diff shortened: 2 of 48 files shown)


vm/src/protocol/object.rs
```diff
@@ -317,8 +317,12 @@ impl PyObject {
...
     pub fn repr(&self, vm: &VirtualMachine) -> PyResult<PyStrRef> {
         vm.with_recursion("while getting the repr of an object", || {
-            let repr = vm.call_special_method(self.to_owned(), identifier!(vm, __repr__), ())?;
-            repr.try_into_value(vm)
+            match self.class().slots.repr.load() {
+                Some(slot) => slot(self, vm),
+                None => vm
+                    .call_special_method(self.to_owned(), identifier!(vm, __repr__), ())?
+                    .try_into_value(vm),
+            }
```
Snowapril, line 325: If we replace all `#[pymethod] repr` with slot implementation, I think we do not need to look up mro again here. We can just use `repr` of `"object"` ( object_repr in cpython )

vm/src/types/slot.rs
```diff
@@ -43,7 +45,7 @@ pub struct PyTypeSlots {
     // More standard operations (here for binary compatibility)
     pub hash: AtomicCell<Option<HashFunc>>,
     pub call: AtomicCell<Option<GenericMethod>>,
-    // tp_str
+    pub repr: AtomicCell<Option<ReprFunc>>,
```
Snowapril, line 48: As `repr` use always object `repr` if there is no `tp_repr` slot, do we need to use `AtomicCell<Option< .. >>` here? Is there any other way like using default implemtation and overriding

haxelion: I'm really unsure about that part 😅
I agree that it doesn't make sense for an object not to have a `repr` slot defined but I don't know how to wire the default object slot here.

Snowapril: Ah.. let me take-back what I commented before. Im not sure exactly, we need to use AtomicCell<Option<..>> at now for updating slot from python-side code. I'm sorry.

vm/src/protocol/object.rs
```diff
@@ -317,8 +317,12 @@ impl PyObject {
 
     pub fn repr(&self, vm: &VirtualMachine) -> PyResult<PyStrRef> {
         vm.with_recursion("while getting the repr of an object", || {
-            let repr = vm.call_special_method(self.to_owned(), identifier!(vm, __repr__), ())?;
-            repr.try_into_value(vm)
+            match self.class().slots.repr.load() {
+                Some(slot) => slot(self, vm),
+                None => vm
+                    .call_special_method(self.to_owned(), identifier!(vm, __repr__), ())?
```
youknowone, line 323: If builtin types fully implemented `__repr__` through `Representable` trait, this is not necessary. `slot` always exists and it will call `__repr__`.

vm/src/types/slot.rs
```diff
@@ -43,7 +45,7 @@ pub struct PyTypeSlots {
     // More standard operations (here for binary compatibility)
     pub hash: AtomicCell<Option<HashFunc>>,
     pub call: AtomicCell<Option<GenericMethod>>,
-    // tp_str
```
youknowone, line 46: still tp_str is not implemented

youknowone (changes requested):
looks good in general. The only concern is about calling unnecesasry `__repr__` from `protocol/object.rs`. Other comments are style suggestions.

vm/src/types/slot.rs
haxelion: Ah no problem, I'm also itching to remove that superfluous option check 🙂

vm/src/protocol/object.rs
haxelion: Yes I guess I just need to continue my implementation work for all the stdlib as well 😅

youknowone: Oh, please don't do it now. I thought it is fully transited. but it is ok to keep it until we finish to migrate them later.

haxelion: I thought I also needed that `vm.call_special_method` for pure python class but it turns out I really only need it for the stdlib so I'll finish cleanly that PR 🙂 There's only 24 methods left to patch so it is doable.

youknowone:
Let me simplify this a bit

youknowone (approved):
👍

Issue #4691, opened 2023-03-15
Encoding and functools ModuleError


basicfunc:
## Feature

While running program, 
```
from functools import lru_cache as lc

@lc
def fib(s):
    if s <= 1:
        return s

    return fib(s-2) + fib(s-1)


for i in range(5_000_000):
    print(f"Fib for {i}: {fib(i)}")
```
I received, 
```
encodings initialization failed. Only utf-8 encoding will be supported.
ModuleNotFoundError: No module named 'encodings'

The above exception was the direct cause of the following exception:

RuntimeError: Could not import encodings. Is your RUSTPYTHONPATH set? If you don't have access to a consistent external environment (e.g. if you're embedding rustpython in another application), try enabling the freeze-stdlib feature
Traceback (most recent call last):
  File "C:\Users\Rahul\tests\fib.py", line 1, in <module>
    from functools import lru_cache as lc
ModuleNotFoundError: No module named 'functools'
```

![image](https://user-images.githubusercontent.com/84739019/225354814-a3f0e560-48a9-4584-aeda-de9e8feee2a8.png)

which is rather weird, I understand if `functools` are not implemented for current implementation of `RustPython`, but why this encoding errors?

Whats more surprising is that it works on playground,
![image](https://user-images.githubusercontent.com/84739019/225355738-8fa993a2-cced-4c95-b775-7255ec1fd20f.png)

Did I missed something while compiling?
I compiled just by `cargo b --release -j10`

## Python Documentation

Here is current implementation of [`functools`](https://github.com/python/cpython/blob/main/Lib/functools.py)

basicfunc:
It appears to be I compiled `RustPython` without `stdlib`, while trying to compile `RustPython` on Windows with the `freeze-stdlib` feature. I noticed that the error occurred due to a difference in how Windows and Linux handle symbolic links.

Upon inspecting the source code, I found that the directory being referred to in the code was a symbolic link, which is interpreted differently on Windows than on Linux. As a result, the compilation failed with the following error message:

![image](https://user-images.githubusercontent.com/84739019/225375302-29aff0de-8499-49cd-84c3-e67f401828ae.png)

While digging deep, I was able to find bug and resolve it where I just hardcoded the `/Lib` directory path in `\pylib\src\lib.rs` as follows:

![image](https://user-images.githubusercontent.com/84739019/225377287-0704be11-47ec-40a8-b330-35ba369f8ad4.png)

[...]

PR #4695, opened 2023-03-15
Update lib.rs

basicfunc:
Resolve the issue listed on [ Encoding and functools ModuleError #4691 ](https://github.com/RustPython/RustPython/issues/4691)

### DIFF
--- a/pylib/src/lib.rs
+++ b/pylib/src/lib.rs
@@ -6,9 +6,9 @@
 // so build.rs sets this env var
 pub const LIB_PATH: &str = match option_env!("win_lib_path") {
     Some(s) => s,
-    None => concat!(env!("CARGO_MANIFEST_DIR"), "/Lib"),
+    None => concat!(env!("CARGO_MANIFEST_DIR"), "../Lib"),
 };
 
 #[cfg(feature = "freeze-stdlib")]
 pub const FROZEN_STDLIB: &rustpython_compiler_core::frozen_lib::FrozenLib =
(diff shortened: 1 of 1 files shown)


pylib/src/lib.rs
```diff
@@ -6,9 +6,9 @@
 // so build.rs sets this env var
 pub const LIB_PATH: &str = match option_env!("win_lib_path") {
     Some(s) => s,
-    None => concat!(env!("CARGO_MANIFEST_DIR"), "/Lib"),
+    None => concat!(env!("CARGO_MANIFEST_DIR"), "../Lib"),
```
youknowone, line 9: during building `rustpython-pylib`, `Lib/` is copied to under `pylib/Lib/`. So this change is not necessary.
somehow in your environemnt, `Lib/` looks not copied under `pylib`. Could you check why it didn't?

basicfunc: It's not copied, it appears to be a symlink, which itself interpreted as a plain text file in windows if it is a unix symbolic link.

youknowone: I don't remember well about windows build, but because our CI is running also on windows environment, processing that part looks having bugs in some cases.
Could you look more in there why it fails on your machine?

basicfunc: OKay, will let you know

basicfunc: Hey, I able to compile without errors when using `RUSTPYTHONPATH` to correct directory

PR #4696, opened 2023-03-15
Test statistics

bearney74:
fix 2 out of 3 TODO python tests in test_statistics.py


Lib/test/test_statistics.py
```diff
@@ -2914,10 +2912,8 @@ def setUp(self):
     def tearDown(self):
         sys.modules['statistics'] = statistics
 
-    # TODO: RUSTPYTHON, ValueError: math domain error
-    @unittest.expectedFailure
-    def test_inv_cdf(self): # TODO: RUSTPYTHON, remove when this passes
-        super().test_inv_cdf() # TODO: RUSTPYTHON, remove when this passes
+    def test_inv_cdf(self):
+        super().test_inv_cdf()
```
fanninpm, line 2916: The `# TODO: RUSTPYTHON, remove when this passes` comments mean that this should be removed **completely**.

youknowone, line 2916:
```suggestion
```

@fanninpm making a suggestion will be easier

youknowone (changes requested):
Looks good. could you check @fanninpm's comment and the CI failure? it is formatting error. You can fix it using `cargo fmt --all`

stdlib/src/math.rs
```diff
@@ -211,7 +211,9 @@ mod math {
     #[pyfunction]
     fn sqrt(value: ArgIntoFloat, vm: &VirtualMachine) -> PyResult<f64> {
         let value = *value;
-        if value.is_nan() { return Ok(value); }   //return nan if value is nan
+        if value.is_nan() {
+            return Ok(value);
+        } //return nan if value is nan
```
DimitrisJim, line 216: Don't think this really requires a comment here, can just remove it:

```suggestion
        if value.is_nan() {
            return Ok(value);
        }
```

fanninpm:
Also, I think @youknowone prefers _rebasing_ to using merge commits.

youknowone (approved):
Thank you for contributing!

@fanninpm When I can keep the commit history, rebase is preferred. In this case, patch is small enough to squash. So it is ok.

PR #4703, opened 2023-03-17
Update _markupbase, html/parser for CPython v3.11.2

tdub0:
See https://github.com/RustPython/RustPython/issues/4564. Updated `_markupbase.py` and `html/parser.py` to match CPython v3.11.2. 

Like `_compression.py`, there were no direct tests, but there were import references in `html/parser.py` and `test/test_htmlparser.py`. Copying from the CPython `Lib`: `_markupbase.py`, `html/`, `test_html.py`, and `test_htmlparser.py` showed only minor changes in `_markupbase.py` and `html/parser.py`. Thus, the tests are unchanged. `test_html.py` and `test_htmlparser.py` ran without error locally following the update.


DimitrisJim:
Thanks! Could you modify your commit to mention the Python version these libraries now correspond to?

youknowone (approved):
Thank you!

Issue #4665, opened 2023-03-07
Add at least one architecture `libc::c_char == u8` build to CI


youknowone:
to prevent breaking #4661

youknowone:
`aarch64-unknown-linux-gnu` looks good. The only Tier 1, so the toolchain would be stable.

yugborana:
hii, I would like to contribute, if you could tell more about the process to be done in this issue

DimitrisJim:
@yugborana great! You basically need to update the workflow file to add the target. If you take a look at the current workflow, there's already some checks made [for other targets](https://github.com/RustPython/RustPython/blob/9b55505cdf514f591cc913e615693ff92f58ff71/.github/workflows/ci.yaml#L192) which you could use as inspiration.

yugborana:
well, I am very new to this, what new updates I have to make in workflow file? Sorry for disturbance

youknowone:
I'd not to recommend this issue if you are very new to github actions. Would you look in #4564 ?

yugborana:
yeah ok

youknowone:
closed by #4704

PR #4704, opened 2023-03-17
Add an architecture with libc::c_char == u8 build to CI

patrickzbhe:
Attempting to add #4665, but I don't know if this is right.

Needed to install `gcc-aarch64-linux-gnu` to get this to check locally.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -194,6 +194,13 @@ jobs:
 
       - uses: dtolnay/rust-toolchain@stable
         with:
+          target: aarch64-unknown-linux-gnu
+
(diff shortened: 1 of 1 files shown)


youknowone:
To check if it is correct, changing one of lib::c_char to i8 and watching it success on x86 but fails on the new platform will work.
isn't gcc target not stable yet? `cargo check --target=aarch64-unknown-linux-gnu` will be fine to check them

patrickzbhe:
> To check if it is correct, changing one of lib::c_char to i8 and watching it success on x86 but fails on the new platform will work.

@youknowone Yup, lib::c_char as i8 works with other architectures but fails with mismatched type on gcc-aarch64-linux-gnu

PR #4705, opened 2023-03-17
pyclass macro to recognize Py/PyRef pattern

youknowone:


vm/src/class.rs
```diff
@@ -63,18 +63,6 @@ pub trait PyClassDef {
     const UNHASHABLE: bool = false;
 }
 
-impl<T> PyClassDef for PyRef<T>
```
youknowone, line 66: This part was never used but added to support `impl PyRef<T>`

PR #4706, opened 2023-03-17
#[pyclass] impl Py<PyType> 

youknowone:
on the top of 
close #4692
close #4705


qingshi163 (approved):
This is GREAT!
We probably can change most of the pymethods witch with PyRef to &Py.

youknowone:
@qingshi163 I am sorry, i pushed commits after your review. and yes, indeed.

PR #4707, opened 2023-03-17
get &= -= working for sets

bearney74:
current set implementation of &= and -= does not work correctly for the following test cases:
s=set([1,2,3])
s &= s
s -= s

This new implementation solves these issues and the test_inplace_on_self test passes.


bearney74:
This is not ready so removing pull request for now.

PR #4708, opened 2023-03-17
test_inplace_on_self now works in test_set and test_weakset

bearney74:
updated &= and -= so that it produces the right result when you do an operation on self:
s=set()
s &= s
s -= s

### DIFF
--- a/Lib/test/test_set.py
+++ b/Lib/test/test_set.py
@@ -582,8 +582,6 @@ class TestSet(TestJointOps, unittest.TestCase):
             else:
                 self.assertNotIn(c, self.s)
 
-    # TODO: RUSTPYTHON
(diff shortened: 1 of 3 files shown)


fanninpm (approved):
LGTM.

___

The parts that say
```rust
for item in temp_inner.elements() {
    self.add(item, vm)?;
}
```
seem inefficient to me, but I'm not sure if RustPython provides a more efficient way than this O(n) approach.

DimitrisJim:
Yeah I think we should merge this but open an issue for the fact that we can skip many iterations knowing that the argument is indeed `self`. See my comment [here](https://github.com/RustPython/RustPython/pull/4424#issuecomment-1373633600) for a bit on this.

PR #4709, opened 2023-03-18
Fix str.splitlines to recognise unicode line terminators

r3m0t:


vm/src/anystr.rs
```diff
@@ -376,7 +376,7 @@ pub trait AnyStr {
         }
     }
 
-    fn py_splitlines<FW, W>(&self, options: SplitLinesArgs, into_wrapper: FW) -> Vec<W>
+    fn py_bytes_splitlines<FW, W>(&self, options: SplitLinesArgs, into_wrapper: FW) -> Vec<W>
```
fanninpm, line 379: Did you mean to make this function so that it's no longer generic over both `str`s and bytes? If so, then it probably doesn't belong in `anystring.rs` anymore.

fanninpm (changes requested):
I have a concern that probably should be addressed.

vm/src/anystr.rs
youknowone, line 379:
```suggestion
    # TODO: remove this function from anystr.
    # See https://github.com/RustPython/RustPython/pull/4709/files#r1141013993
    fn py_bytes_splitlines<FW, W>(&self, options: SplitLinesArgs, into_wrapper: FW) -> Vec<W>
```

youknowone (approved):

vm/src/anystr.rs
r3m0t: @youknowone wrong comment format 😄

r3m0t: I guess, it doesn't have any uses in the `str` context any more.

PR #4711, opened 2023-03-18
Fix panic from test_int__format__locale

### LINKED ISSUE #4588: Format panic in test_int__format__locale
## Issues

In `Lib/test/test_types.py``test_int__format__locale`, if you change from `format(x, 'n')` to `format(x, ',')` and run the test, RustPython will panic! It doesn't crash in CPython!

See #4593 for test case

minhrongcon2000:
Fix #4588

closes #4593


youknowone:
would you remove `skip` from the test and add test case commit from #4593 to ensure it is fixed?

minhrongcon2000:
It will still fail because I didn't integrate locale into format integer with numeric option... But if you want to remove it to test whether it is panic, I will do it!

youknowone:
you can replace `skip` to `expectedFailure` in that case. At least we can test if it is crash or not.

Lib/test/test_format.py
```diff
@@ -492,6 +492,14 @@ def test_precision_c_limits(self):
...
         with self.assertRaises(ValueError) as cm:
             format(c, ".%sf" % (INT_MAX + 1))
+            
+    def test_issue_4588_rustpython(self):
+        x = 123456789012345678901234567890
+        for i in range(0, 30):
+            format(x, ',')
+
+            # move to the next integer to test
+            x = x // 10
```
youknowone, line 502: we don't add our original tests under `Lib/`. Could you move this test under `extra_tests/snippets/`?

Lib/test/test_format.py
```diff
@@ -493,6 +493,5 @@ def test_precision_c_limits(self):
         with self.assertRaises(ValueError) as cm:
             format(c, ".%sf" % (INT_MAX + 1))
 
-
```
youknowone, line 496: this is not intended change, right?

minhrongcon2000: Yeah! This is not intended

extra_tests/snippets/test_issue_4588.py
```diff
@@ -0,0 +1,5 @@
+x = 123456789012345678901234567890
+for i in range(0, 30):
+    format(x, ',')
```
youknowone, line 3: this is test for `format`. `builtin_format.py` covers tests about it.

Lib/test/test_format.py
youknowone: please also revert this line too

youknowone (changes requested):

youknowone (approved):
Thank you!

youknowone:
rerunning CI with #4593

youknowone:
Could you check the test result? new code seems generating malformatted string.

```
---- format::tests::test_format_int_sep stdout ----
thread 'format::tests::test_format_int_sep' panicked at 'assertion failed: `(left == right)`
  left: `Ok("0000001234567,890,123,456,789,012,345,678")`,
 right: `Ok("1,234,567,890,123,456,789,012,345,678")`', common/src/format.rs:1035:9
```

youknowone (approved):
Thanks!

PR #4715, opened 2023-03-18
Fix test_bool test_subclass

minhrongcon2000:
Fix unsafe `__new__` for int function

### DIFF
--- a/Lib/test/test_bool.py
+++ b/Lib/test/test_bool.py
@@ -6,9 +6,6 @@ from test.support import os_helper
 import os
 
 class BoolTest(unittest.TestCase):
-
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_subclass(self):
         try:
             class C(bool):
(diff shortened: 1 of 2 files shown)


youknowone (commented):
Thank you for contributing, but this fix doesn't seem to correct implementation.

Could you check corresponding cpython implementation?
https://github.com/python/cpython/blob/main/Objects/typeobject.c#L7694-L7701

minhrongcon2000:
Sure! I will check it when I have time :D

minhrongcon2000:
@youknowone The block of code that you show me in the previous comment is to check whether `cls` is a subtype of the current class that called `__new__`. In CPython, however, `bool` is a subtype of int and in CPython, int is not allowed to call `__new__` with subtype as well. See https://github.com/python/cpython/blob/87be8d95228ee95de9045cf2952311d20dc5de45/Objects/typeobject.c#L7711-L7718

youknowone:
Oh, i am sorry. I left wrong link

Snowapril:
Oh .... It seems duplicate PR exist here #4455 😭 (cc. @youknowone )

DimitrisJim:
yeah and that PR goes about things in the approach CPython has by modifying `type.__new__` to catch this.

minhrongcon2000:
Oh! So I will close this PR now :D

DimitrisJim:
we should just coordinate with garychia and see if they still want to work on it, I'll leave a comment there to check. If they don't have time, you can take inspiration from their approach :smile:

PR #4721, opened 2023-03-18
type.__name__ & getattr optimization

youknowone:
I was not able to split changes to multiple commits.
I will leave comments to each key changes.


vm/src/types/slot.rs
```diff
@@ -1215,19 +1219,19 @@ impl PyComparisonOp {
 #[pyclass]
 pub trait GetAttr: PyPayload {
     #[pyslot]
-    fn slot_getattro(obj: &PyObject, name: PyStrRef, vm: &VirtualMachine) -> PyResult {
+    fn slot_getattro(obj: &PyObject, name: &Py<PyStr>, vm: &VirtualMachine) -> PyResult {
```
youknowone, line 1222: `getattro` slot now takes `&Py<Str>` to avoid ref count manipulation.

vm/src/builtins/str.rs
```diff
@@ -189,6 +189,42 @@ impl IntoPyStrRef for &'static PyStrInterned {
     }
 }
 
+pub trait AsPyStr<'a>
```
youknowone, line 192: new trait to generalize `&Py<PyStr>` input

vm/src/builtins/str.rs
```diff
@@ -189,6 +189,42 @@ impl IntoPyStrRef for &'static PyStrInterned {
...
+}
+
+impl<'a> AsPyStr<'a> for &'a PyStrRef {
+    #[inline]
+    fn as_pystr(self, _ctx: &Context) -> &'a Py<PyStr> {
+        self
+    }
+}
+
+impl AsPyStr<'static> for &'static str {
```
youknowone, line 214: `&'static str` is a constant or interned string. either way they will be interned

vm/src/builtins/type.rs
```diff
@@ -39,8 +40,8 @@ pub struct PyType {
     pub heaptype_ext: Option<Pin<Box<HeapTypeExt>>>,
 }
 
-#[derive(Default)]
 pub struct HeapTypeExt {
+    pub name: PyRwLock<PyStrRef>,
```
youknowone, line 44: `ht_name` in CPython

vm/src/builtins/type.rs
```diff
@@ -133,29 +134,44 @@ impl PyType {
             ctx,
         )
     }
-    pub fn new_ref(
+    pub fn new_heap(
```
youknowone, line 137: renamed to show this is creating a heap type.

vm/src/builtins/type.rs
```diff
@@ -133,29 +134,44 @@ impl PyType {
...
+            heaptype_ext,
             metaclass,
             ctx,
         )
     }
 
     #[allow(clippy::too_many_arguments)]
-    fn new_verbose_ref(
-        name: &str,
+    fn new_heap_inner(
```
youknowone, line 174: `_inner` because not desired to be called from outside.

vm/src/builtins/type.rs
```diff
@@ -133,29 +134,44 @@ impl PyType {
...
-            HeapTypeExt::default(),
+            heaptype_ext,
             metaclass,
             ctx,
         )
     }
 
     #[allow(clippy::too_many_arguments)]
-    fn new_verbose_ref(
-        name: &str,
```
youknowone, line 158: because slots is holding name, the caller have to correctly set name.

vm/src/builtins/type.rs
```diff
@@ -213,8 +227,7 @@ impl PyType {
         Ok(new_type)
     }
 
-    pub fn new_bare_ref(
-        name: &str,
+    pub fn new_static(
```
youknowone, line 230: new_static to distinguish from new_heap

vm/src/builtins/type.rs
```diff
@@ -425,8 +439,15 @@ impl PyType {
     }
 
     #[pygetset]
-    fn __name__(&self) -> String {
-        self.name().to_string()
+    pub fn __name__(&self, vm: &VirtualMachine) -> PyStrRef {
```
youknowone, line 442: `__name__` now returns object. interned str for static and ht_name for heaptype

vm/src/types/slot.rs
```diff
@@ -30,7 +30,10 @@ macro_rules! atomic_func {
 #[derive(Default)]
 #[non_exhaustive]
 pub struct PyTypeSlots {
-    pub name: PyRwLock<Option<String>>, // tp_name, not class name
+    /// # Safety
+    /// For static types, always safe.
+    /// For heap types, `__name__` must alive
+    pub(crate) name: AtomicCell<&'static str>, // tp_name with <module>.<class> for print, not class name
```
youknowone, line 36: `slots.name` is now atomic cell. And the value itself is unsafe for heaptypes. instead of direct accessing, `slot_name()` will be safe.
So this field is not pub anymore.

In fact, this field even does not need to be `AtomicCell` anymore. UnsafeCell?

youknowone:
@qingshi163 could you review this changes?

qingshi163 (approved):
LGTM

vm/src/builtins/type.rs
```diff
@@ -837,26 +876,43 @@ impl PyType {
...
+        if name.as_str().as_bytes().contains(&0) {
             return Err(vm.new_value_error("type name must not contain null characters".to_owned()));
         }
-        *self.slots.name.write() = Some(name.as_str().to_string());
+
+        unsafe {
+            // # Safety
+            // changing slots.name while holding __name__ lock is safe
+            let mut name_lock = self.heaptype_ext.as_ref().unwrap().name.write();
+            self.slots.name.store(&*(name.as_str() as *const _));
```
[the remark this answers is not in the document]
youknowone: ah, good catch. I think cpython use it as fast-path but we anyway do lock for that.

Issue #4724, opened 2023-03-19
errors when installing RustPython with ssl features on Windows


rust-kotlin:
Openssl-x64 prepared using vcpkg on Windows11. 
After run this command, I receive this error message
```bash
cargo install --git https://github.com/RustPython/RustPython --features ssl
```
```rust
error[E0308]: mismatched types
   --> stdlib\src\ssl.rs:542:13
    |
541 |         fn options(&self) -> libc::c_ulong {
    |                              ------------- expected `u32` because of return type
542 |             self.ctx.read().options().bits()
    |             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `u32`, found `u64`
    |
help: you can convert a `u64` to a `u32` and panic if the converted value doesn't fit
    |
542 |             self.ctx.read().options().bits().try_into().unwrap()
    |                                             ++++++++++++++++++++

error[E0308]: mismatched types
   --> stdlib\src\ssl.rs:547:61
    |
547 |                 .set_options(SslOptions::from_bits_truncate(opts));
    |                              ------------------------------ ^^^^ expected `u64`, found `u32`
    |                              |
    |                              arguments to this function are incorrect
    |
note: associated function defined here
   --> C:\Users\tom\.cargo\registry\src\github.com-1ecc6299db9ec823\openssl-0.10.46\src\ssl\mod.rs:144:1
    |
144 | / bitflags! {
145 | |     /// Options controlling the behavior of an `SslContext`.
146 | |     pub struct SslOptions: SslOptionsRepr {
147 | |         /// Disables a countermeasure against an SSLv3/TLSv1.0 vulnerability affecting CBC ciphers.
...   |
279 | |     }
280 | | }
    | |_^
    = note: this error originates in the macro `__impl_bitflags` which comes from the expansion of the macro `bitflags` (in Nightly builds, run with -Z macro-backtrace for more info)
help: you can convert a `u32` to a `u64`
    |
547 |                 .set_options(SslOptions::from_bits_truncate(opts.into()));
    |                                                                 +++++++
```
I changed the sourced code following the suggestions given by the compiler, and then I succeeded in installing RustPython with ssl features.

PR #4725, opened 2023-03-19
Solve the problem of unmatched types

rust-kotlin:
Solve the issue #4724

### DIFF
--- a/stdlib/src/ssl.rs
+++ b/stdlib/src/ssl.rs
@@ -539,7 +539,7 @@ mod _ssl {
 
         #[pygetset]
         fn options(&self) -> libc::c_ulong {
-            self.ctx.read().options().bits().try_into().unwrap()
+            self.ctx.read().options().bits() as _
         }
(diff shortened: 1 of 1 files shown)


rust-kotlin:
I guess the problem I met with is because I am using the msvc-stable rust toolchain, which might call the VC++ compiler to compile the openssl, resulting in a program whose arch is x86 by default. Then I should consider how to avoid the clippy error message: useless type conversion...

stdlib/src/ssl.rs
```diff
@@ -539,12 +539,12 @@ mod _ssl {
 
         #[pygetset]
         fn options(&self) -> libc::c_ulong {
-            self.ctx.read().options().bits()
+            self.ctx.read().options().bits().try_into().unwrap()
```
youknowone, line 542:
```suggestion
            self.ctx.read().options().bits() as _
```
maybe this will be fine

youknowone (approved):
Thank you for contributing!

Issue #4470, opened 2023-01-24
Comparision between `str` and `str` including surrogate characters doesn't work


moreal:
## Feature

Comparison between `str` and `str` doesn't work when there are unicode surrogates[^1]. Then it became `REPLACEMENT CHARACTER, 0xefbfbd` and their comparison result became `Equal`.

You can see the `test_unicode.UnicodeTest.test_comparison` test.

```python
self.assertTrue('\ud800\udc02' < '\ud84d\udc56')
```

[^1]: https://unicodebook.readthedocs.io/unicode_encodings.html#utf-16-surrogate-pairs

## Solution sketch

 - Find a way to handle unicode surrogates (utf16) also.
 - Write custom comparison logic between `PyStr` and `PyStr` and use it in `impl Compare for PyStr`.
 
```rust
// I didn't check this code works.
fn cmp(zelf: PyStr, other: PyStr): Op {
  for i in 0..zelf.bytes.len() {
    if zelf.bytes[i] > other.bytes[i] {
      return Greater;
    }
    else if zelf.bytes[i] < other.bytes[i] {
      return Less;
    }
  }
  Equal
}
```

youknowone:
#3574 maybe related

youknowone:
now we have wtf8

PR #4726, opened 2023-03-19
Allow unicode surrogates to be stored in str

garychia:
Related to #4470 .
This pull request attempts to allow RustPython to store unicode surrogates in a `str`.

### DIFF
--- a/Lib/test/test_unicode.py
+++ b/Lib/test/test_unicode.py
@@ -541,8 +541,6 @@ class UnicodeTest(string_tests.CommonTest,
             self.assertEqual('abc' == bytearray(b'abc'), False)
             self.assertEqual('abc' != bytearray(b'abc'), True)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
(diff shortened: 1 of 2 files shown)


youknowone:
This is potentially not safe. Rust will fill the unused `char` ranges with other values when it can do. (if not yet, it can do those optimization in future)
To correctly fix this problem, #3574 need to be finished.
are you interested in it?

garychia:
I'm still looking for a better solution for this. Sure. I'll take a look at #3574. Thanks.

youknowone:
@garychia did you get any idea about it?

garychia:
Currently the parser uses Rust chars and strings to handle Python string literals. This does not work with Unicode surrogate characters. We may need to change the way we store Python string literals. My suggestion is to replace Rust strings with our custom class (such as PyStr, I guess). But I guess this may require a lot of work to be done.

fanninpm:
Perhaps the [`bstr`](https://crates.io/crates/bstr) crate can be useful.

PR #4731, opened 2023-03-20
PyPayload::class takes ctx instead of vm

youknowone:


derive-impl/src/pyclass.rs
```diff
@@ -482,8 +482,8 @@ pub(crate) fn impl_define_exception(exc_def: PyExceptionDef) -> Result<TokenStre
 
         // We need this to make extend mechanism work:
         impl ::rustpython_vm::PyPayload for #class_name {
-            fn class(vm: &::rustpython_vm::VirtualMachine) -> &'static ::rustpython_vm::Py<::rustpython_vm::builtins::PyType> {
-                vm.ctx.exceptions.#ctx_name
+            fn class(ctx: &::rustpython_vm::vm::Context) -> &'static ::rustpython_vm::Py<::rustpython_vm::builtins::PyType> {
```
DimitrisJim, line 485: is there a benefit to doing it this was as opposed to passing the whole vm struct?

youknowone: This is related to further PRs. Context exists before vm initialization. To use `PyPayload::class()` during interpreter & vm initialization step, this change is necessary.

DimitrisJim (approved):

PR #4732, opened 2023-03-20
Fix str.__str__ and PyRef::into_exact_or for int() or str() pattern

youknowone:
disallowing subclass is a repeatitive pattern

### DIFF
--- a/vm/src/builtins/int.rs
+++ b/vm/src/builtins/int.rs
@@ -15,7 +15,7 @@ use crate::{
     },
     protocol::{PyNumber, PyNumberMethods},
     types::{AsNumber, Comparable, Constructor, Hashable, PyComparisonOp, Representable},
-    AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyRef, PyResult,
+    AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyRef, PyRefExact, PyResult,
     TryFromBorrowedObject, VirtualMachine,
 };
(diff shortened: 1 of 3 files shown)


vm/src/builtins/int.rs
```diff
@@ -701,6 +687,17 @@ impl PyInt {
     }
 }
 
+#[pyclass]
+impl PyRef<PyInt> {
+    #[pymethod(magic)]
+    fn int(self, vm: &VirtualMachine) -> PyRefExact<PyInt> {
+        self.into_exact_or(&vm.ctx, |zelf| unsafe {
+            // TODO: this is actually safe. we need better interface
```
DimitrisJim, line 695: we should definitely start documenting all the unsafe better at some point.

DimitrisJim (approved):
this lgtm

vm/src/builtins/int.rs
fanninpm: There's the allow-by-default [`undocumented_unsafe_blocks`](https://rust-lang.github.io/rust-clippy/master/index.html#undocumented_unsafe_blocks) lint, in case we want to activate that.

youknowone: That clippy option looks very good for us because we use a lot of unsafe blocks.

For this function, I'd like to make the basic PyRef creation functions to return PyRefExact first and turn them to PyRef later. In that case, everything will be safe by both semantics and rust grammars.

PR #4733, opened 2023-03-20
memory optimization using leaked PyRef<PyStr>

youknowone:
in my environment, repl runtime consumes:

before: 6.6MB
after: 6.0MB

the difference is almost same as our doc db.

closes #4731
closes #4732
as included dependency


youknowone:
i need to find a way to convince miri in safe way

youknowone:
@qingshi163 could you review this PR?

youknowone:
uh, failing again. i will look in it.

PR #4738, opened 2023-03-21
Refactor Number Protocol

qingshi163:
- PyNumberBinaryFunc take both &PyObject instead PyNumber
- Remove PyNumberMethods from HeatTypeExt
- Rename slots.number to slots.as_number, the original as_number for reference is no more needed
- Initalize slots.as_number from PyNumberMethods if static type
- Initalize slots.as_number via magic attribute if heap type
- now PyNumber is a thin wrapper for &PyObject


youknowone (approved):
👍

PR #4742, opened 2023-03-23
Improve Comparable take zelf as reference

qingshi163:


vm/src/types/slot.rs
```diff
@@ -992,58 +992,35 @@ pub trait Comparable: PyPayload {
         vm: &VirtualMachine,
     ) -> PyResult<PyComparisonValue>;
 
+    #[inline]
```
fanninpm, line 995: This seems unrelated.

youknowone: other functions anyway have it

youknowone (approved):

PR #4745, opened 2023-03-25
Update test_array.py from Cpython v3.11.2

ilp-sys:
#4564 update test_array.py from Cpython


youknowone (approved):
Thank you for contributing!

PR #4747, opened 2023-03-25
update test from test_pwd cpython 3.11.2

JaeyoungAhn:
#4564 Update test from  `test_pwd` cpython 3.11.2


youknowone (approved):
Thank you for contributing!

PR #4748, opened 2023-03-25
Fix int.from_bytes and Update random.py and test/test_random.py from CPython v3.11.2

### LINKED ISSUE #4750: int.from_bytes can omit byte_order parameter
## Feature

Changed in version 3.11: Added default argument value for byteorder.
Remove unnecessary parameter in "random.py"
```python
a = int.from_bytes(a + _sha512(a).digest())
```

TypeError: Expected at least 1 arguments (2 given)
## Python Documentation

documentation: [CPython v3.11.2 from_bytes issue](https://docs.python.org/3.11/library/stdtypes.html?highlight=from_bytes#int.from_bytes) 
source code: [ref](https://github.com/RustPython/RustPython/pull/4748/commits/822c764d6d8754d738daa4ba2608ac80fafe98e1#diff-851cf2d8d99b6d8dee26acc31319431dce51dbe26e3e56494d9b4c45edd8e11fL170)

jyj0816:
#4564

Fix #4750

### DIFF
--- a/Lib/random.py
+++ b/Lib/random.py
@@ -167,15 +167,11 @@ class Random(_random.Random):
         elif version == 2 and isinstance(a, (str, bytes, bytearray)):
             if isinstance(a, str):
                 a = a.encode()
-            a = int.from_bytes(a + _sha512(a).digest(), 'big')
+            a = int.from_bytes(a + _sha512(a).digest())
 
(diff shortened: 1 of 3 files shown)


youknowone (approved):
Thank you for contributing!

PR #4749, opened 2023-03-25
updated test_sort from cpython 3.11.2

JaeyoungAhn:
#4564 marked failed case expected failure


youknowone (approved):
Thank you for contributing!

PR #4752, opened 2023-03-25
update test_getopt.py from cpython 3.11.2

jinnarajin:
update test_getopt.py from cpython 3.11.2


youknowone (approved):
Thank you for contributing!

PR #4753, opened 2023-03-25
Add test_eof.py from Cpython v3.11.2

ilp-sys:
#4564 Add test_eof.py from Cpython v3.11.2 with expectedFailure

### DIFF
--- a/Lib/test/test_eof.py
+++ b/Lib/test/test_eof.py
@@ -1,0 +1,77 @@
+"""test script for a few new invalid token catches"""
+
+import sys
+from test import support
+from test.support import os_helper
+from test.support import script_helper
+import unittest
+
+# TODO: RUSTPYTHON
+@unittest.expectedFailure
+class EOFTestCase(unittest.TestCase):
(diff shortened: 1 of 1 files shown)


youknowone (approved):

Lib/test/test_eof.py
```diff
@@ -0,0 +1,77 @@
+"""test script for a few new invalid token catches"""
+
+import sys
+from test import support
+from test.support import os_helper
+from test.support import script_helper
+import unittest
+
+# TODO: RUSTPYTHON
+@unittest.expectedFailure
```
fanninpm, line 10: @youknowone I don't think that decorator _can_ be applied on a class, but I could be wrong.

youknowone: It works as we expect. Removing it results all test fail.
On the other hand, applying them on function looks more useful to support issue hunters though.

fanninpm: Either way, these lines should have been broken out into a separate commit.

PR #4754, opened 2023-03-25
update test_math.py from cpython 3.11.2

Brain-organizer:
update test_math.py from python 3.11.2

#4564


youknowone (approved):
Thank you for contributing!

PR #4757, opened 2023-03-25
update test_listcomps.py from cpython 3.11.2

Brain-organizer:
update test_listcomps.py from cpython 3.11.2

#4564


DimitrisJim (approved):
lgtm, thanks!

PR #4746, opened 2023-03-25
Update test_bytes.py from CPython v3.11.2

MegasKomnenos:
https://github.com/RustPython/RustPython/issues/4564

PR #4760, opened 2023-03-25
Fix issue with cformat converting bytes to string for bytesinner

MegasKomnenos:
This is a follow-up of the following PR
https://github.com/RustPython/RustPython/pull/4746

It addresses the failing test case for test_mod, where,

```
b'...%(foo)b...' % {b'foo':b"abc"}
```

caused invalid key error for 'foo'

The cause of this issue was that individual tokens in bytesinner were converted to string before being used for string formatting, but the token being addressed was expected to be in bytes.

It was addressed by making sure that it is treated as bytes.


youknowone (approved):
looks good, thank you!

PR #4761, opened 2023-03-25
updated test_deque.py from CPython v3.11.2

JaeyoungAhn:
#4564 updated `test_deque.py` from CPython v3.11.2

### DIFF
--- a/Lib/test/test_deque.py
+++ b/Lib/test/test_deque.py
@@ -814,6 +814,8 @@ class TestSubclass(unittest.TestCase):
         d.clear()
         self.assertEqual(len(d), 0)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_copy_pickle(self):
         for cls in Deque, DequeWithSlots:
             for d in cls('abc'), cls('abcde', maxlen=4):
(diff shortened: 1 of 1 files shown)


Lib/test/test_deque.py
```diff
@@ -809,41 +814,31 @@ def test_basics(self):
         d.clear()
         self.assertEqual(len(d), 0)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
DimitrisJim, line 818: Interesting. Do you happen to remember what the error was here? We should open an issue because of the regression.

JaeyoungAhn: ERROR: test_copy_pickle (__main__.TestSubclass.test_copy_pickle)
  File "./test_deque.py", line 837, in test_copy_pickle
    self.assertEqual(e.x, d.x)
AttributeError: 'Deque' object has no attribute 'x'

DimitrisJim: thanks! I'll open the issue after merging.

DimitrisJim (approved):

Lib/test/test_deque.py
fanninpm: @DimitrisJim I believe this should have been broken out into a separate commit.

Issue #4762, opened 2023-03-25
cformat star asterisk should behave differently when given negative sign


MegasKomnenos:
In string formatting, a print-f style format specifier can have its amount and precision formatted by variable value through use of star asterik(%*s).

In cformat_bytes and cformat_string, through try_update_quantity_from_tuple, try_update_quantity_from_element is used to update quantity if possible, but it makes the assumption that its behavior should be the same regardless of the sign value of the given element.

```python
if let Some(i) = width_obj.payload::<PyInt>() {
    let i = i.try_to_primitive::<i32>(vm)?.unsigned_abs();
    Ok(CFormatQuantity::Amount(i as usize))
}
```

However, that assumption is false, as an amount can be preceded by a negative sign, and it should make the resulting string left adjusted, according to format specifier rule.
https://docs.python.org/3/library/stdtypes.html#old-string-formatting

```python
assert('%*s' % (-5, 'abc') == 'abc  ')
```

This issue makes this test to fail.

I think this issue can be resolved by checking the sign of the element and adding the minus sign accordingly. 

I'm working on the fix at the moment.

DimitrisJim:
hah, first time I've heard of the asterisk tbh.

MegasKomnenos:
> hah, first time I've heard of the asterisk tbh.

Same here lol. I wouldn't have known it if there wasn't a failing test that used it, haha.

PR #4766, opened 2023-03-25
Allow try_update_quantity_from_tuple to set left alignment when needed

MegasKomnenos:
More details can be found here
https://github.com/RustPython/RustPython/issues/4762

It was the last issue that made test_mod to fail, so I also removed the expectedFailure flag from that test.

### DIFF
--- a/common/src/cformat.rs
+++ b/common/src/cformat.rs
@@ -754,6 +754,12 @@ impl CFormatString {
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq)]
+pub enum CFormatAlign {
+    Left,
+    Right,
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/common/src/cformat.rs
+++ b/common/src/cformat.rs
@@ -754,12 +754,6 @@ impl CFormatString {
     }
 }
 
-#[derive(Debug, Copy, Clone, PartialEq)]
-pub enum CFormatAlign {
-    Left,
-    Right,
-}
-
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/vm/src/cformat.rs
+++ b/vm/src/cformat.rs
@@ -204,12 +204,19 @@ fn spec_format_string(
 fn try_update_quantity_from_element(
     vm: &VirtualMachine,
     element: Option<&PyObjectRef>,
-) -> PyResult<CFormatQuantity> {
+) -> PyResult<(CFormatQuantity, CFormatAlign)> {
     match element {
         Some(width_obj) => {
             if let Some(i) = width_obj.payload::<PyInt>() {
-                let i = i.try_to_primitive::<i32>(vm)?.unsigned_abs();
-                Ok(CFormatQuantity::Amount(i as usize))
+                let i = i.try_to_primitive::<i32>(vm)?;
+                let align: CFormatAlign;
+                if i < 0 {
+                    align = CFormatAlign::Left;
+                } else {
+                    align = CFormatAlign::Right;
+                }
+                let i = i.unsigned_abs();
+                Ok((CFormatQuantity::Amount(i as usize), align))
             } else {
                 Err(vm.new_type_error("* wants int".to_owned()))
             }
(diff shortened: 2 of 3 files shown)


vm/src/cformat.rs
```diff
@@ -204,12 +204,18 @@ fn spec_format_string(
 fn try_update_quantity_from_element(
```
youknowone, line 204: Now quantity and precision has different spec, but they are handled by same function.
I don't think we necessarily need to split it to 2 functions, but adding a comment to `try_update_quantity_from_element` will be helpful to prevent confusion.

my suggestion is:
```suggestion
# This functions is used to parse both quantity and precision. Please consider to split this function once they doesn't fit in single function anymore.
fn try_update_quantity_from_element(
```

common/src/cformat.rs
```diff
@@ -754,6 +754,12 @@ impl CFormatString {
     }
 }
 
+#[derive(Debug, Copy, Clone, PartialEq)]
```
DimitrisJim, line 757: along with YunWon's suggestion, I'd document this. We currently don't do this as much as we'd probably like but it's relatively straight-forward here.

vm/src/cformat.rs
```diff
@@ -239,7 +249,7 @@ fn try_update_precision_from_tuple<'a, I: Iterator<Item = &'a PyObjectRef>>(
     let Some(CFormatPrecision::Quantity(CFormatQuantity::FromValuesTuple)) = p else {
         return Ok(());
     };
-    let quantity = try_update_quantity_from_element(vm, elements.next())?;
+    let (quantity, _) = try_update_quantity_from_element(vm, elements.next())?;
```
DimitrisJim, line 252: this also lends to the fact that `try_update_quantity_from_element` might be doing two jobs that aren't always needed.

vm/src/cformat.rs
MegasKomnenos: Thanks for your feedback! Yeah, I think that would help ease the potential confusion.

vm/src/cformat.rs
MegasKomnenos: That is a fair point. I think it could be possible to add a new, separate function that only checks the sign of the given element and set the flag directly, without messing with return types. 

I did what I did yesterday mainly because I originally did not know that I could just set left adjustment to spec flags directly, so I wanted to have some means of storing the alignment data to later reference when I'm actually formatting the string. However, I think I forgot to adjust later when I discovered what I can do with CConversionFlags. 

Let me work on the fix real quick.

vm/src/cformat.rs
```diff
@@ -218,15 +218,39 @@ fn try_update_quantity_from_element(
     }
 }
 
+fn try_update_adjust_from_tuple(
```
youknowone, line 221:
```suggestion
fn try_conversion_flag_from_tuple(
```

vm/src/cformat.rs
```diff
@@ -218,15 +218,39 @@ fn try_update_quantity_from_element(
...
+    vm: &VirtualMachine,
+    element: Option<&PyObjectRef>,
+    f: &mut CConversionFlags,
+) -> PyResult<()> {
```
youknowone, line 225:
```suggestion
) -> PyResult<CConversionFlags> {
```
I prefer to have immutable parameters when possible

vm/src/cformat.rs
```diff
@@ -218,15 +218,39 @@ fn try_update_quantity_from_element(
...
+    match element {
+        Some(width_obj) => {
+            if let Some(i) = width_obj.payload::<PyInt>() {
+                let i = i.try_to_primitive::<i32>(vm)?;
+                if i < 0 {
+                    f.insert(CConversionFlags::LEFT_ADJUST)
+                }
+                Ok(())
```
youknowone, line 233:
```suggestion
                let flags = if i < 0 {
                    CConversionFlags::LEFT_ADJUST
                } else {
                    CConversionFlags::from_bits(0).unwrap()
                };
                Ok(flags)
```

vm/src/cformat.rs
```diff
@@ -218,15 +218,39 @@ fn try_update_quantity_from_element(
...
     elements: &mut I,
     q: &mut Option<CFormatQuantity>,
+    f: &mut CConversionFlags,
 ) -> PyResult<()> {
     let Some(CFormatQuantity::FromValuesTuple) = q else {
         return Ok(());
     };
-    let quantity = try_update_quantity_from_element(vm, elements.next())?;
+    let element = elements.next();
+    try_update_adjust_from_tuple(vm, element, f)?;
```
youknowone, line 252:
```suggestion
    f.insert(try_conversion_flag_from_tuple(vm, element)?);
```

youknowone (changes requested):
style suggestion for immutable arguments

vm/src/cformat.rs
MegasKomnenos: That is fair. I think I was trying to mimic what other functions nearby were doing, but that is a fair suggestion.

vm/src/cformat.rs
MegasKomnenos: All implemented!

youknowone (approved):
looks good, thank you!

PR #4772, opened 2023-03-26
update string_tests and solve failing test

Brain-organizer:
I update string_tests.py from python 3.11.2. And then I found that the test case in string_tests was failing because rustpython returns b'' when executing the statement b''.replace('', 'A', 100), whereas cpython returns b'A'. 
After investigating, I discovered that there was no handling for cases where both the self and the first argument are empty strings, so I fixed it.

### DIFF
--- a/vm/src/builtins/str.rs
+++ b/vm/src/builtins/str.rs
@@ -826,9 +826,11 @@ impl PyStr {
         let s = self.as_str();
         match count {
             OptionalArg::Present(max_count) if max_count >= 0 => {
-                if max_count == 0 || s.is_empty() {
+                if max_count == 0 || (s.is_empty() && !old.is_empty()) {
                     // nothing to do; return the original bytes
-                    s.into()
+                    s.to_owned()
+                } else if s.is_empty() && old.is_empty() {
+                    new.as_str().to_owned()
                 } else {
                     s.replacen(old.as_str(), new.as_str(), max_count as usize)
                 }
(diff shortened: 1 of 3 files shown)


youknowone:
Could you also check `test_unicode`? `string_tests` is not a test endpoint but a shared test library from `test_unicode`, `test_bytes` and `test_userstring`

youknowone:
I'd like to suggest make a PR with [bugfix](https://github.com/RustPython/RustPython/pull/4772/commits/72f9f5e150525699decbd1d46f0b0c0e0ba499b8)
and the other one with tests file update including `test_unicode` if possible

Brain-organizer:
thank you for your review! I included test_unicode!

vm/src/builtins/str.rs
```diff
@@ -826,10 +826,12 @@ impl PyStr {
         let s = self.as_str();
         match count {
             OptionalArg::Present(max_count) if max_count >= 0 => {
-                if max_count == 0 || s.is_empty() {
+                if max_count == 0 || (s.is_empty() && !old.is_empty()) {
                     // nothing to do; return the original bytes
-                    s.into()
-                } else {
+                    s.to_owned()
+                }else if s.is_empty() && old.is_empty(){
```
youknowone, line 832: running `cargo fmt --all` will fix this format warning

Brain-organizer: Thanks for your advice! I modify that part by running `cargo fmt --all`

fanninpm:
Does `test_bytes.py` import `string_tests.py`?

youknowone (approved):
Looks good, thank you!

youknowone:
@fanninpm test_bytes is already updated by other PR

PR #4773, opened 2023-03-26
Fix test_path_like_objects

ilp-sys:
add support for FsPath

### DIFF
--- a/Lib/test/test_compile.py
+++ b/Lib/test/test_compile.py
@@ -757,8 +757,6 @@ if 1:
         self.assertTrue(f1(0))
         self.assertTrue(f2(0.0))
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_path_like_objects(self):
         # An implicit test for PyUnicode_FSDecoder().
         compile("42", FakePath("test_compile_pathlike"), "single")
--- a/vm/src/stdlib/os.rs
+++ b/vm/src/stdlib/os.rs
@@ -5,6 +5,7 @@ use crate::{
     function::{ArgumentError, FromArgs, FsPath, FuncArgs},
     AsObject, Py, PyObjectRef, PyPayload, PyResult, TryFromObject, VirtualMachine,
 };
+
 use std::{
     ffi, fs, io,
     path::{Path, PathBuf},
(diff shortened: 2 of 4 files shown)


Lib/test/test_compile.py
```diff
@@ -710,7 +710,7 @@ def check_different_constants(const1, const2):
         self.assertTrue(f2(0.0))
 
     # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+    # @unittest.expectedFailure
```
youknowone, line 713: Please remove the entire comments. We don't need them anymore.

youknowone (changes requested):

vm/src/stdlib/os.rs
```diff
@@ -184,6 +185,19 @@ impl FsPath {
             FsPath::Bytes(b) => b.as_bytes(),
         }
     }
+
+    pub fn as_str(&self) -> &str {
+        match self {
+            FsPath::Bytes(b) => std::str::from_utf8(b).unwrap(),
+            FsPath::Str(s) => s.as_str(),
+        }
+    }
```
youknowone, line 194: Since `FsPath` is intended to obtain `OsString` in Rust side, I am opposed to add `as_str` like this signature here.

My suggestion is adding `fn to_os_string(&self) -> std::os::OsString` and use it from `fn compile`.

Not this case, but even if we need to implement `as_str()`, it will look like:
```rust
pub fn as_str(&self) -> Result<&str, Utf8Error>
```

youknowone (changes requested):

youknowone:
Oh, `FsPath` is placed in `os.rs` but it is not compiled for `wasm`.
relocating them to a new file `vm/src/function/fspath.rs` or `vm/src/function/mod.rs` will be helpful.

youknowone:
@ilp-sys Hi, is this accidently closed? could you open it again?

ilp-sys:
I must have done something wrong with the git cli, reopened it .

youknowone:
I found refactoring FsPath is more complex issue than it looks like. Let me look in more for a few more days.

youknowone:
@ilp-sys #4782 is finally done! Could you please rebase this patch? I think removing the last commit will make the other 2 commits be rebase without conflict. And not that much changes will be needed.

youknowone:
`test_basic_script_with_path_object` is also succsesful

youknowone (approved):
👍

PR #4774, opened 2023-03-26
Update bisect.py and test_bisect.py from CPython v3.11.2

MegasKomnenos:


youknowone (approved):
Thank you!

PR #4776, opened 2023-03-26
changed builtins.rs to resolve test case error in test_builtin.py

JaeyoungAhn:
fn delattr, fn getattr, fn setattr, fn hasattr in `builtins.rs` was changed to resolve test case error in test_delattr, test_getattr, test_hasattr, test_setattr in `test_builtin.py`

Refactoring is required for more general code.


vm/src/stdlib/builtins.rs
```diff
@@ -338,7 +351,13 @@ mod builtins {
     }
 
     #[pyfunction]
-    fn hasattr(obj: PyObjectRef, attr: PyStrRef, vm: &VirtualMachine) -> PyResult<bool> {
+    fn hasattr(obj: PyObjectRef, attr: PyObjectRef, vm: &VirtualMachine) -> PyResult<bool> {
+        let attr = PyStrRef::try_from_object(vm, attr.clone()).map_err(|_e| {
```
youknowone, line 355:
```suggestion
        let attr = PyStrRef::try_from_object(vm, attr).map_err(|_e| {
```
attr can be moved

JaeyoungAhn: error occurs when .clone() is removed.

youknowone: Oh, i am sorry. It is used again for type error below

vm/src/stdlib/builtins.rs
```diff
@@ -338,7 +351,13 @@ mod builtins {
...
+            vm.new_type_error(format!(
+                "attribute name must be string, not '{}'",
+                attr.class().name()
+            ))
+        })?;
```
youknowone, line 360:
```suggestion
        let attr = attr.try_to_ref::<PyStr>(vm)?.map_err(|_e| {
            vm.new_type_error(format!(
                "attribute name must be string, not '{}'",
                attr.class().name()
            ))
        })?;
```
Another suggestion to remove unnecessary retain operation.

JaeyoungAhn: I'm not really sure but the initial code I commited returns `&PyRef<PyStr>` but the suggestion seems to return `&Py<PyStr>`, which leads to an error.

Could I ask how to solve it? Thank you.

youknowone: Yes, they are different.
`PyStrRef::try_from_object` moves argument's value to its return value while `PyObject::try_to_ref` borrows `&Py<T>` from its argument.

Because `attr` is now a borrow, return part also need to be `Ok(vm.get_attribute_opt(obj, attr)?.is_some())`
without `&`.

If you still experience problem, i will look in it again. pasting error message will be helpful.

JaeyoungAhn: Just changed the code based on your suggestion. I appreciate your help.

youknowone (approved):
Looks great, thank you for contributing!

PR #4777, opened 2023-03-26
make int.to_bytes arguments optional and update base64.py and test_base64.py from CPython v3.11.2

MegasKomnenos:

I also added default parameters to to_bytes to fix tests that failed.


youknowone (approved):
Thank you!

Issue #4662, opened 2023-03-07
io.TextIOWrapper.__init__ encoding="locale" support


youknowone:
## Feature

https://docs.python.org/3/library/io.html#io.TextIOWrapper

## Python Documentation

documentation: https://docs.python.org/3/library/io.html#io.TextIOWrapper
source code: https://github.com/python/cpython/blob/4a3ea1fdd890e5e2ec26540dc3c958a52fba6556/Modules/_io/textio.c#L1120-L1130

## Tips
- search for `TextIOWrapperArgs` in RustPython source code to look for `encoding` parameter and init function.
- ` _Py_GetLocaleEncodingObject` is same as `locale.getencoding`

PR #4781, opened 2023-03-26
Add encoding="locale" support for TextIOWrapper

MegasKomnenos:
https://github.com/RustPython/RustPython/issues/4662

### DIFF
--- a/vm/src/stdlib/io.rs
+++ b/vm/src/stdlib/io.rs
@@ -2193,10 +2193,14 @@ mod _io {
             *data = None;
 
             let encoding = match args.encoding {
-                Some(enc) => enc,
-                None => {
-                    // TODO: try os.device_encoding(fileno) and then locale.getpreferredencoding()
-                    PyStr::from(crate::codecs::DEFAULT_ENCODING).into_ref(&vm.ctx)
+                None if vm.state.settings.utf8_mode > 0 => PyStr::from("utf-8").into_ref(&vm.ctx),
+                Some(enc) if enc.as_str() != "locale" => enc,
+                _ => {
+                    // None without utf8_mode or "locale" encoding
+                    vm.import("locale", None, 0)?
+                        .get_attr("getencoding", vm)?
+                        .call((), vm)?
+                        .try_into_value(vm)?
                 }
             };
 
--- a/vm/src/stdlib/sys.rs
+++ b/vm/src/stdlib/sys.rs
@@ -728,7 +728,7 @@ mod sys {
                 hash_randomization: settings.hash_seed.is_none() as u8,
                 isolated: settings.isolated as u8,
                 dev_mode: settings.dev_mode,
-                utf8_mode: 1,
+                utf8_mode: settings.utf8_mode,
                 int_max_str_digits: -1,
                 safe_path: false,
                 warn_default_encoding: settings.warn_default_encoding as u8,
--- a/vm/src/vm/setting.rs
+++ b/vm/src/vm/setting.rs
@@ -74,6 +74,8 @@ pub struct Settings {
     /// false for wasm. Not a command-line option
     pub allow_external_library: bool,
 
+    pub utf8_mode: u8,
+
     #[cfg(feature = "flame-it")]
     pub profile_output: Option<OsString>,
     #[cfg(feature = "flame-it")]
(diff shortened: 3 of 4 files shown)


vm/src/stdlib/sys.rs
```diff
@@ -701,7 +701,7 @@ mod sys {
         /// -X dev
         dev_mode: bool,
         /// -X utf8
-        utf8_mode: u8,
+        utf8_mode: bool,
```
youknowone, line 704: I am sorry for wrong guide. This is `u8` on this struct

vm/src/vm/setting.rs
```diff
@@ -74,6 +74,8 @@ pub struct Settings {
     /// false for wasm. Not a command-line option
     pub allow_external_library: bool,
 
+    pub utf8_mode: bool,
```
youknowone, line 77: I have no idea about this variable, but keeping it as same type as flag looks reasonable. So probably `u8` here too?

youknowone (changes requested):
And you fixed 2 more tests by this change!

```
UNEXPECTED SUCCESS: test__getitem___deprecation (test.test_fileinput.FileInputTests.test__getitem___deprecation)
UNEXPECTED SUCCESS: test_gz_ext_fake (test.test_fileinput.Test_hook_compressed.test_gz_ext_fake)
```
Please remove `unittest.expectedFailure` from those tests

vm/src/stdlib/sys.rs
MegasKomnenos: Hehe, no problemo.

vm/src/vm/setting.rs
MegasKomnenos: Yeah, I think that'd make sense.

MegasKomnenos: Done!

vm/src/stdlib/io.rs
```diff
@@ -2202,10 +2202,26 @@ mod _io {
...
-                    PyStr::from(crate::codecs::DEFAULT_ENCODING).into_ref(&vm.ctx)
+                    if vm.state.settings.utf8_mode == 1 {
+                        // Used to be crate::codecs::DEFAULT_ENCODING, which was "utf-8"
+                        PyStr::from("utf-8").into_ref(&vm.ctx)
+                    } else {
+                        vm.import("locale", None, 0)?
+                            .get_attr("getencoding", vm)?
+                            .call((), vm)?
+                            .try_into_value(vm)?
+                    }
```
youknowone, line 2224: conditional match arms will be useful.
Also, we can directly use `crate::codecs::DEFAULT_ENCODING` rather than copying the value.
```suggestion
                None if vm.state.settings.utf8_mode == 1 => {
                    PyStr::from(crate::codecs::DEFAULT_ENCODING).into_ref(&vm.ctx)
                }
                Some(enc) if enc.as_str() != "locale" => enc,
                _ => {
                    // None without utf8_mode or "locale" encoding
                    vm.import("locale", None, 0)?
                        .get_attr("getencoding", vm)?
                        .call((), vm)?
                        .try_into_value(vm)?
                }
```

youknowone (changes requested):

vm/src/stdlib/io.rs
MegasKomnenos: Yeah, I decided to put "utf-8" instead of default encoding because the corresponding cpython code explicitly called for utf-8, not default encoding. Those two are same atm for rustpython, but I figured that kind of implicit assumption are better removed.

Ah, I forgot that I could use match guards to further refine match cases. 

I think that is a very nice way to put it. At first I did not get why you used a default case rather than a case without a match guard, but then I realized that that would again lead to code duplication, which was the very issue that you tried to solve.

Pushed!

vm/src/stdlib/io.rs
```diff
@@ -2202,10 +2202,14 @@ mod _io {
             *data = None;
 
             let encoding = match args.encoding {
-                Some(enc) => enc,
-                None => {
-                    // TODO: try os.device_encoding(fileno) and then locale.getpreferredencoding()
-                    PyStr::from(crate::codecs::DEFAULT_ENCODING).into_ref(&vm.ctx)
+                None if vm.state.settings.utf8_mode == 1 => PyStr::from("utf-8").into_ref(&vm.ctx),
```
youknowone, line 2205:
```suggestion
                None if vm.state.settings.utf8_mode > 0 => PyStr::from("utf-8").into_ref(&vm.ctx),
```

PR #4782, opened 2023-03-26
relocate FsPath from stdlib::os and rename PyPathLike to OsPath

youknowone:
FsPath work to support #4773 

cc @ilp-sys 

`PyPathLike` is defined in `posixmodule` as `path_t`.
But FsPath is more common concept.

`PyPathLike` also is not a python object. We don't prefix non-python object with `Py` anymore without good reasons

### DIFF
--- a/common/src/os.rs
+++ b/common/src/os.rs
@@ -1,6 +1,6 @@
 // TODO: we can move more os-specific bindings/interfaces from stdlib::{os, posix, nt} to here
 
-use std::io;
+use std::{io, str::Utf8Error};
 
 #[cfg(windows)]
 pub fn errno() -> io::Error {
@@ -20,4 +20,15 @@ pub fn errno() -> io::Error {
 #[cfg(not(windows))]
 pub fn errno() -> io::Error {
     io::Error::last_os_error()
+}
+
+#[cfg(unix)]
+pub fn bytes_as_osstr(b: &[u8]) -> Result<&std::ffi::OsStr, Utf8Error> {
+    use std::os::unix::ffi::OsStrExt;
+    Ok(std::ffi::OsStr::from_bytes(b))
+}
+
+#[cfg(not(unix))]
+pub fn bytes_as_osstr(b: &[u8]) -> Result<&std::ffi::OsStr, Utf8Error> {
+    Ok(std::str::from_utf8(b)?.as_ref())
 }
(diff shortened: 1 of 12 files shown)


vm/src/function/fspath.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    #[cfg(unix)]
+    pub fn to_cstring(&self, vm: &VirtualMachine) -> PyResult<std::ffi::CString> {
+        std::ffi::CString::new(self.as_bytes().to_vec()).map_err(|e| e.into_pyexception(vm))
+    }
+
+    #[cfg(windows)]
+    pub fn to_widecstring(&self, vm: &VirtualMachine) -> PyResult<widestring::WideCString> {
+        widestring::WideCString::from_os_str(&self.as_os_str()?)
```
fanninpm, line 99:
```suggestion
        widestring::WideCString::from_os_str(&self.as_os_str(vm)?)
```

youknowone:
@fanninpm could you review it again?

common/src/os.rs
```diff
@@ -21,3 +21,19 @@ pub fn errno() -> io::Error {
...
+#[cfg(unix)]
+pub fn bytes_as_osstr(b: &[u8]) -> Result<&std::ffi::OsStr, Utf8Error> {
+    use std::os::unix::ffi::OsStrExt;
+    Ok(std::ffi::OsStr::from_bytes(b))
+}
+
+#[cfg(not(unix))]
+pub fn bytes_as_osstr(b: &[u8]) -> Result<&std::ffi::OsStr, Utf8Error> {
+    Ok(std::str::from_utf8(b)?.as_ref())
+}
```
fanninpm, line 34: Does this work?
```suggestion
pub fn bytes_as_osstr(b: &[u8]) -> Result<&std::ffi::OsStr, Utf8Error> {
    #[cfg(unix)]
    {
        use std::os::unix::ffi::OsStrExt;
        Ok(std::ffi::OsStr::from_bytes(b))
    }
    #[cfg(not(unix))]
    {
        Ok(std::str::from_utf8(b)?.as_ref())
    }
}
```

DimitrisJim (approved):

common/src/os.rs
youknowone: I prefer to keep separated functions. Those functions are small enough. But when the logic goes longer, single function commonly leads to `#[cfg]` spaghetti here and their while separated functions leads to create a new shared function.

PR #4783, opened 2023-03-27
Add test_descr from cpython v3.11.2

Snowapril:
While working on unary & ternary function behaviour in number protocol, I found there is `test_descr` in cpython. 
There are a lot of tc which makes failure or internal crash. It will be helpful for our slot & descriptor works

### DIFF
--- a/Lib/test/test_descr.py
+++ b/Lib/test/test_descr.py
@@ -1,0 +1,5825 @@
+import builtins
+import copyreg
+import gc
+import itertools
+import math
+import pickle
+import random
+import string
+import sys
+import types
+import unittest
+import warnings
+import weakref
+
+from copy import deepcopy
+from contextlib import redirect_stdout
+from test import support
+
+try:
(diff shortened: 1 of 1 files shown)


Lib/test/test_descr.py
```diff
@@ -1826,7 +1848,8 @@ def __init__(self, foo):
         object.__init__(A(3))
         self.assertRaises(TypeError, object.__init__, A(3), 5)
 
-    @unittest.expectedFailure
+    # RUSTPYTHON: this is added from cpython3.11.2. below test case is success in rustpython
```
fanninpm, line 1851: Minor clarification:
```suggestion
    # TODO: RUSTPYTHON, CPython 3.5 and above expect this test case to fail, but in RustPython this currently passes.
    # See https://github.com/python/cpython/issues/49572 for more details.
```

youknowone (approved):
Thank you!

Issue #3846, opened 2022-07-08
Add missing dis featues for modulefinder


youknowone:
It is using a few missing features in `_dis` module.

Maybe adding them will not be enough for modulefinder due to bytecode imcompatibility.But adding `dis` items will be a good issue to start looking at RustPython project.

To add a new python module, check this article: https://rustpython.github.io/blog/2020/04/04/how-to-contribute-by-cpython-unittest.html

itsankitkp:
Can I pick up this issue? Also, can you suggest a good place to start?

youknowone:
try to checkout #3863 and check what kind of functions are not implemented. 
reading related cpython code is also helpful.

In RustPython,
`_dis` module is here: https://github.com/RustPython/RustPython/blob/main/stdlib/src/dis.rs
`dis` module is here: https://github.com/RustPython/RustPython/blob/main/Lib/dis.py

itsankitkp:
Thanks @youknowone, I will take a look at PR and specified files.

itsankitkp:
My observations:
1. dis.opmap['LOAD_CONST']
cpython: 100
rustpython: AttributeError: module 'dis' has no attribute 'opmap'
2. dis.EXTENDED_ARG
cpython: 144
rustpython: AttributeError: module 'dis' has no attribute 'EXTENDED_ARG'

3. dis._unpack_opargs(100)
cpython: <generator object _unpack_opargs at 0x7f465322fde0>
rustpython: AttributeError: module 'dis' has no attribute '_unpack_opargs'

These are only relevant to [#3863](https://github.com/RustPython/RustPython/pull/3863)

Additionally:
dir(dis) in cpython
[...]

PR #4798, opened 2023-04-01
Update dis module from Python 3.12

itsankitkp:
This PR add functions and method from Python 3.12 for dis module and also updates test_dis.py.
test_dis.py mostly gets skipped due to lack of _testinternalcapi (`test_dis skipped -- _testinternalcapi required`) but that would be separate work (to make _testinternalcapi functional). 
For this change, I have removed one OPCODE `RETURN_CONST` and commented out some test cases with TODO: RustPython  as they required features which are not yet available.

This also addresses https://github.com/RustPython/RustPython/issues/3846

EDIT 1:
Failures: extra_tests/snippets/stdlib_dis.py
```
import dis

dis.dis(compile("5 + x + 5 or 2", "", "eval"))
```
fails with error
  File "/home/ankit/RustPython/pylib/Lib/dis.py", line 113, in dis
    type(x).__name__)
TypeError: don't know how to disassemble code objects

on cpython, `compile("5 + x + 5 or 2", "", "eval")` object has `co_code` attribute which leads to it to branch to `co_code` handler.
```
    elif hasattr(x, 'co_code'): # Code object
        _disassemble_recursive(x, file=file, depth=depth, show_caches=show_caches, adaptive=adaptive)
```
On RustPython, `co_code` attribute is not there and hence none of if..else branch in dis function gets matched leading to TypeError fail getting triggered (as mentioned above).
Looking into how it used to work before and how can this be fixed.

EDIT 2:
probably need co_code


itsankitkp:
Closing this, will revisit again with better approach.

PR #4799, opened 2023-04-02
Update test_charmapcodec.py from Cpython v3.11.2

Masorubka1:


fanninpm:
I'd recommend not giving up on this one quite yet.

PR #4807, opened 2023-04-02
Update test_crashers from Cpython v3.11.2

Masorubka1:


DimitrisJim:
Not sure if this one should be added actually since it mostly should be testing some very cpython specific stuff.

PR #4809, opened 2023-04-02
Update test_dict_version.py from Cpython v3.11.2

Masorubka1:


Lib/test/test_dict_version.py
```diff
@@ -0,0 +1,191 @@
+"""
+Test implementation of the PEP 509: dictionary versioning.
+"""
+import unittest
+from test.support import import_helper
+
+# PEP 509 is implemented in CPython but other Python implementations
```
DimitrisJim, line 7: I don't believe we implement it, unsure if this test file actually gets executed tbh.

DimitrisJim:
Thanks, closing this for now since we do not implement this. I also see PEP 509 has been superseded (don't remember by what) so it is very likely we won't be supporting this (though I haven't looked into it).

PR #4815, opened 2023-04-02
Update test_embed.py from Cpython v3.11.2

Masorubka1:


DimitrisJim:
thanks! This also seems to be CPython specific (testing embedding of CPython). Sorry for not keeping track of this in the issue list, I'll try and update it.

PR #4817, opened 2023-04-02
Add test_exception_group.py from Cpython v3.11.2

Masorubka1:


youknowone (approved):
we don't have exception group yet. but this test file will be a good starting point to implement it

PR #4822, opened 2023-04-02
Add test_fileutils.py from Cpython v3.11.2

Masorubka1:


DimitrisJim:
Thanks for all these PRs @Masorubka1, I'll be closing the CPython specific ones, sorry for not keeping track of that in the list to save you the trouble!

PR #4824, opened 2023-04-02
Add test_flufl.py from Cpython v3.11.2

Masorubka1:


fanninpm:
NOTE: This was an April Fools Day prank back in the day.

DimitrisJim:
Yup, unsure if this should be added too, doubt we'll ever special case that import.

PR #4827, opened 2023-04-03
Remove merge group.

### LINKED ISSUE #4826: [RFC] Removing merge group
## Summary

Let's remove merge group for now until they leave public beta and hopefully become more feature complete.

## Detailed Explanation

Currently having the merge group removes the ability to squash, rebase, set a commit message when merging a PR. As a result, PRs with a lot of noise due to feedback etc can seriously pollute the commit history. Though you can set it up to squash, this isn't always the best approach and might remove information that could be relevant.

Though having it does seem like a good approach, we should probably reconsider in the future when [these issues](https://github.com/orgs/community/discussions/46757) might hopefully have been addressed.

DimitrisJim:
closes #4826 if agreed.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -3,7 +3,6 @@ on:
     branches: [main, release]
   pull_request:
     types: [unlabeled, opened, synchronize, reopened]
-  merge_group:
 
 name: CI
 


youknowone:
This is enabling CI running for merge queue branches. Removing this line causes not disabling merge queue but it makes merge queue runs without CI check.

youknowone:
Instead of this patch, I disabled merge queue. This line is harmless. Unless we enable it again later, it will do nothiing.

DimitrisJim:
nice, sounds good to me!

PR #4835, opened 2023-04-04
Update test_hashlib.py from Cpython v3.11.2 & refactor hashlib

Masorubka1:
1) split hashlib and _hashlib
2) add libraries: _md5, _sha1, _sha3, _sha256, _blake2
3) add tests for hashlib

### DIFF
--- a/Lib/test/test_enum.py
+++ b/Lib/test/test_enum.py
@@ -2889,9 +2889,7 @@ class TestIntFlag(unittest.TestCase):
         self.assertEqual(Color.ALL.value, 7)
         self.assertEqual(str(Color.BLUE), 'blue')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
-    @unittest.skipIf(sys.platform == "win32", "TODO: RUSTPYTHON, inconsistent test result on Windows due to threading")
+    @unittest.skipIf(sys.platform == "win32" or sys.platform.startswith("linux"), "TODO: RUSTPYTHON, inconsistent test result on Windows due to threading")
     @threading_helper.reap_threads
     def test_unique_composite(self):
         # override __eq__ to be identity only
--- a/Lib/test/test_enum.py
+++ b/Lib/test/test_enum.py
@@ -2889,7 +2889,7 @@ class TestIntFlag(unittest.TestCase):
         self.assertEqual(Color.ALL.value, 7)
         self.assertEqual(str(Color.BLUE), 'blue')
 
-    @unittest.skipIf(sys.platform == "win32" or sys.platform.startswith("linux"), "TODO: RUSTPYTHON, inconsistent test result on Windows due to threading")
+    @unittest.skip("TODO: RUSTPYTHON, inconsistent test result due to threading")
     @threading_helper.reap_threads
     def test_unique_composite(self):
         # override __eq__ to be identity only
--- a/stdlib/Cargo.toml
+++ b/stdlib/Cargo.toml
@@ -55,6 +55,7 @@ sha-1 = "0.10.0"
 sha2 = "0.10.2"
 sha3 = "0.10.1"
 blake2 = "0.10.4"
+hmac = "0.12.1"
 
 ## unicode stuff
 unicode_names2 = { workspace = true }
--- a/stdlib/src/hashlib.rs
+++ b/stdlib/src/hashlib.rs
@@ -7,7 +7,9 @@ pub mod _hashlib {
     use crate::common::lock::PyRwLock;
     use crate::vm::{
         builtins::{PyBytes, PyStrRef, PyTypeRef},
-        function::{ArgBytesLike, FuncArgs, OptionalArg},
+        convert::ToPyObject,
+        function::{ArgBytesLike, ArgStrOrBytesLike, FuncArgs, OptionalArg},
+        protocol::PyBuffer,
         PyObjectRef, PyPayload, PyResult, VirtualMachine,
     };
     use blake2::{Blake2b512, Blake2s256};
@@ -19,14 +21,13 @@ pub mod _hashlib {
     use sha2::{Sha224, Sha256, Sha384, Sha512};
     use sha3::{Sha3_224, Sha3_256, Sha3_384, Sha3_512, Shake128, Shake256};
 
-    #[derive(FromArgs, Traverse)]
+    #[derive(FromArgs, Debug)]
     #[allow(unused)]
     struct NewHashArgs {
         #[pyarg(positional)]
         name: PyStrRef,
         #[pyarg(any, optional)]
         data: OptionalArg<ArgBytesLike>,
-        #[pytraverse(skip)]
         #[pyarg(named, default = "true")]
         usedforsecurity: bool,
     }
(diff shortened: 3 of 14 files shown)


fanninpm:
Hmm… I think we need to split up [`stdlib/src/hashlib.rs`](https://github.com/RustPython/RustPython/blob/main/stdlib/src/hashlib.rs) into different modules and turn `hashlib` into a pure-Python module. What do you think, @youknowone?

youknowone:
`hashlib.py` strongly supposes `_hashlib` is cpython-like openssl implementation. It will require efforts.
In my opinion, that's more like test-purpose compatibility  rather than actual compatiblity. But I will happily take patch if anyone want to work on it.

Masorubka1:
@youknowone Could you review this pr?

stdlib/Cargo.toml
```diff
@@ -55,6 +55,7 @@ sha-1 = "0.10.0"
 sha2 = "0.10.2"
 sha3 = "0.10.1"
 blake2 = "0.10.4"
+hmac = {version = "0.12.1"}
```
fanninpm, line 58: Simplification:
```suggestion
hmac = "0.12.1"
```

stdlib/src/md5.rs
```diff
@@ -0,0 +1,13 @@
+pub(crate) use _md5::make_module;
+
+#[pymodule]
+mod _md5 {
+    use crate::hashlib::_hashlib::{HashArgs, HashWrapper, PyHasher};
+    use crate::vm::{PyObjectRef, PyPayload, PyResult, VirtualMachine};
+    use md5::Md5;
+
+    #[pyfunction(name = "md5")]
+    fn md5(args: HashArgs, vm: &VirtualMachine) -> PyResult<PyObjectRef> {
```
youknowone, line 10:
```suggestion
    fn md5(args: HashArgs, vm: &VirtualMachine) -> PyResult {
```
PyResult = PyResult<PyObjectRef>

youknowone (changes requested):

stdlib/src/md5.rs
```diff
@@ -0,0 +1,13 @@
...
+        Ok(PyHasher::new("md5", HashWrapper::new::<Md5>(args.string)).into_pyobject(vm))
```
youknowone, line 11:
```suggestion
        Ok(crate::hashlib::md5(args).into_pyobject(vm))
```
Is this possible?

stdlib/src/hashlib.rs
```diff
@@ -229,90 +231,142 @@ mod hashlib {
...
+        };
+
+        let b_hash = match b.class().is(_py_str) {
+            true => b.try_to_value::<&str>(vm)?.as_bytes().to_vec(),
+            _ => b.try_to_value::<PyBytes>(vm)?.as_bytes().to_vec(),
+        };
+
+        println!("\na: {:?}", a_hash);
+        println!("b: {:?}", b_hash);
+        println!("equal: {}", (a_hash == b_hash));
```
youknowone, line 336: debug code remained

stdlib/src/hashlib.rs
```diff
@@ -229,90 +231,142 @@ mod hashlib {
...
+    #[pyfunction(name = "openssl_blake2s")]
     fn blake2s(args: BlakeHashArgs) -> PyHasher {
         PyHasher::new("blake2s", HashWrapper::new::<Blake2s256>(args.data))
     }
 
-    trait ThreadSafeDynDigest: DynClone + DynDigest + Sync + Send {}
+    #[pyfunction]
+    fn compare_digest(
+        a: PyObjectRef,
+        b: PyObjectRef,
```
youknowone, line 307:
```suggestion
        a: ArgStrOrBytesLike,
        b: ArgStrOrBytesLike,
```
This type will be helpful

youknowone (changes requested):
Wow, You made it 👍 
Looks mostly good. I left a few comments.

Lib/test/test_enum.py
```diff
@@ -2889,9 +2889,7 @@ class Color(StrMixin, AllMixin, IntFlag):
         self.assertEqual(Color.ALL.value, 7)
         self.assertEqual(str(Color.BLUE), 'blue')
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
-    @unittest.skipIf(sys.platform == "win32", "TODO: RUSTPYTHON, inconsistent test result on Windows due to threading")
+    @unittest.skipIf(sys.platform == "win32" or sys.platform.startswith("linux"), "TODO: RUSTPYTHON, inconsistent test result on Windows due to threading")
```
fanninpm, line 2892: If we also observe this on macOS…

```suggestion
    @unittest.skip("TODO: RUSTPYTHON, inconsistent test result due to threading")
```

Lib/test/test_secrets.py
```diff
@@ -44,6 +44,7 @@ def test_bad_types(self):
 
     def test_bool(self):
         # Test that compare_digest returns a bool.
+        print(secrets.compare_digest("abc", "abc"))
```
youknowone, line 47:
```suggestion
```
Unless the print came from CPython source code

vm/src/function/buffer.rs
```diff
@@ -141,6 +141,20 @@ impl ArgStrOrBytesLike {
...
+            Self::Str(_) => "str",
+        }
+    }
+
+    pub fn get_enum_int(&self) -> u8 {
+        match self {
+            Self::Buf(_) => 1,
+            Self::Str(_) => 2,
+        }
+    }
```
youknowone, line 157: This is too much ad-hoc

stdlib/src/hashlib.rs
```diff
@@ -211,108 +213,144 @@ mod hashlib {
...
+        b: ArgStrOrBytesLike,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyObjectRef> {
+        if a.get_enum_int() != b.get_enum_int() {
+            return Err(vm.new_type_error(format!(
+                "different types in compare_digest {} and {} ",
+                a.get_enum(),
+                b.get_enum()
+            )));
+        }
```
youknowone, line 316: The error message is different.
I am not going to ask to match the message matching, but I disagree about adding `get_enum` and `get_enum_int` to make incompatible error message.
```
Python 3.11.1 (main, Feb  4 2023, 11:11:18) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> _hashlib.compare_digest('', bytearray(b''))
TypeError: a bytes-like object is required, not 'str'
```

```suggestion
        fn is_str(arg: &ArgStrOrBytesLike) -> {
           matches!(arg, ArgStrOrBytesLike::String(_))
        }
        if is_str(&a) != is_str(&b) {
            return Err(...);
        }
```

youknowone (changes requested):

stdlib/src/hashlib.rs
Masorubka1: @youknowone are you suggesting comparing class names as strings character by character? It takes a lot of time on performance.

Masorubka1: I agree that method that i used is disgusting, but i didn't found any examples by comparing instances enums.

youknowone: I left a code suggestion using matches! macro

youknowone (approved):
Thank you! This is great, I like it very much!

PR #4839, opened 2023-04-04
Update test_list.py from Cpython v3.11.2

Masorubka1:


Lib/test/test_list.py
```diff
@@ -21,7 +21,7 @@ def test_basic(self):
                          [1, 3, 5, 7, 9])
 
         # XXX RUSTPYTHON TODO: catch ooms
-        if sys.maxsize == 0x7fffffff and False:
+        if sys.maxsize == 0x7fffffff:
```
DimitrisJim, line 24: hm, I'd need to check the blame here. Maybe we added this `False` because this triggered a segfault. Seems like there's no issue now so the preceding comment might be able to be removed.

fanninpm: I'd recommend testing this on an older-model (32-bit) Raspberry Pi, if you have one.

DimitrisJim: Yup. As @fanninpm mentioned, this is a guard for 32bit platforms. Let's revert this change until we get a chance to properly check it.

Masorubka1: @fanninpm I tested on my old laptop 32 and there was no problem

youknowone (approved):
Thank you!

PR #4847, opened 2023-04-04
Don't run whats_left with older CPython versions

verhovsky:
I added a check that whats_left is being run with CPython 3.11 or newer, as suggested here https://github.com/RustPython/RustPython/pull/4255#issuecomment-1298014225 and also put one signature under the other to make them easier to compare

### DIFF
--- a/whats_left.py
+++ b/whats_left.py
@@ -34,8 +34,9 @@ GENERATED_FILE = "extra_tests/not_impl.py"
 
 implementation = platform.python_implementation()
 if implementation != "CPython":
-    sys.exit("whats_left.py must be run under CPython, got {implementation} instead")
-
+    sys.exit(f"whats_left.py must be run under CPython, got {implementation} instead")
+if sys.version_info[:2] < (3, 11):
+    sys.exit(f"whats_left.py must be run under CPython 3.11 or newer, got {implementation} {sys.version} instead")
 
 def parse_args():
     parser = argparse.ArgumentParser(description="Process some integers.")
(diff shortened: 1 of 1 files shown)


whats_left.py
```diff
@@ -483,11 +484,17 @@ def remove_one_indent(s):
...
+            if cpython_value and cpython_value.startswith("ValueError("):
+                continue  # these items will never match
+            if rustpy_value is None or rustpy_value.startswith("ValueError("):
+                rustpy_value = f" {rustpy_value}"
+            print(f"{item}{rustpy_value}")
+            if cpython_value is None or cpython_value.startswith("ValueError("):
+                cpython_value = f" {cpython_value}"
+            print(f"{' ' * len(item)}{cpython_value}")
+            if i < len(mismatched) - 1:
+                print()
```
DimitrisJim, line 497: Thanks! Can you add a comment explaining your logic here? Are we filtering the same things here? (I.e do we ever have `ValueError("something else")` which is now being ignored?

verhovsky: I'm not at my computer right now, you can remove the if statement and run the code to see the error that I'm filtering. The point is really that if we can't get the signature for a function in CPython then there's no point in checking it signature in RustPython.

whats_left.py
```diff
@@ -483,11 +484,17 @@ def remove_one_indent(s):
...
-                continue # these items will never match
-
-            print(f"{item} {rustpy_value} != {cpython_value}")
+        for i, (item, rustpy_value, cpython_value) in enumerate(mismatched):
+            if cpython_value and cpython_value.startswith("ValueError("):
+                continue  # these items will never match
+            if rustpy_value is None or rustpy_value.startswith("ValueError("):
+                rustpy_value = f" {rustpy_value}"
+            print(f"{item}{rustpy_value}")
+            if cpython_value is None or cpython_value.startswith("ValueError("):
```
youknowone, line 493:
```suggestion
            if cpython_value is None:
```
Will this be the same logic? because `ValueError` is already filtered in line 488

whats_left.py
```diff
@@ -483,11 +484,17 @@ def remove_one_indent(s):
...
+                cpython_value = f" {cpython_value}"
+            print(f"{' ' * len(item)}{cpython_value}")
```
DimitrisJim, line 495: also, minor nit here, but we don't account for the potential space added in the `rustpy_value` on line 491 so the alignment might be off by one.

DimitrisJim (approved):
ran this locally and looks good to me, like the alignment of signatures! Apart from youknowone's suggestion, should be gtg. Thanks!

youknowone:
Thank you for contributing!

Issue #4797, opened 2023-03-31
[RFC] Faster BigInt Implement via Malachite?


qingshi163:
## Summary
Currently we are using a num-bigint, a pure rust implement for BigInt support, but at lease now the implement is bit naive especially do not have small int optimize witch is a big deal.

The fastest BigInt implement I know is gmplib witch part of GNU, there is a high-level wrapper https://docs.rs/rug/latest/rug/ witch looks really nice.

## Drawbacks, Rationale, and Alternatives
I am worried about the platform supporting mostly for webassembly, I don't see gmplib has offical support for that. and compiling on Windows seem like depends on MingW.

Otherwise we probably need to have our own implement.

DimitrisJim:
Yes! The small int optimization would be really cool to have.

In general, I think we should prefer options that don't require additional software to be present on the host machine, that's why I was also looking into malachite too. 

@qingshi163 are there any other benefits to using `rug` over something like `malachite`? I can see that `GMP` would be battle hardened and probably be faster than malachite with the downside that it might be troublesome to set up.

qingshi163:
malachite sounds great!

PR #4850, opened 2023-04-04
[WIP] Faster BigInt Implement via Malachite

qingshi163:
#4797


common/src/int.rs
```diff
@@ -1,120 +1,631 @@
...
+        // log2(x) = log2(2^n * 2^-n * x) = n + log2(x/2^n)
+        // If we set 2^n to be the greatest power of 2 below x, then x/2^n is in [1, 2), and can
+        // thus be converted into a float.
+        let n = self.bits() as u32 - 1;
+        let frac = Rational::from_integers_ref(self.inner(), Self::from(2).pow(n.into()).inner());
+        let float: f64 = frac.rounding_into(RoundingMode::Down);
+        f64::from(n) + float.log2()
+    }
+
+    pub fn getrandbits<F>(mut k: usize, mut f: F) -> Self
```
youknowone, line 388: is this part of BigInt?

common/src/int.rs
```diff
@@ -1,120 +1,631 @@
...
+    }
+
+    pub fn to_bytes_le(&self) -> (Ordering, Vec<u8>) {
+        (
+            self.sign(),
+            self.inner().unsigned_abs_ref().to_power_of_2_digits_asc(8),
+        )
+    }
+
+    pub fn from_bytes_be(sign: bool, digits: &[u8]) -> Self {
```
youknowone, line 501:
```suggestion
    pub fn from_bytes_be(sign: Sign, digits: &[u8]) -> Self {
```
BigInt will be a part of public API of RustPython.
I wish we avoid boolean for non-boolean-meaning values for public interface as documenting purpose.

common/src/int.rs
```diff
@@ -1,120 +1,631 @@
...
+    From,
+    Shr,
+    Shl,
+    Sum,
+    Product,
+    ShlAssign,
+    ShrAssign,
+)]
+#[from(forward)]
+pub struct BigInt(Integer);
```
youknowone, line 53: It seems you are almost making alternative num-bigint implementation. 
I think making a crate like `malachite-bigint` as a drop-in replacement of `num-bigint` might be worth if you are interested in.

qingshi163: malachite-bigint is a good idea

qingshi163:
> Can we close in preference to #4952?

sure

Issue #4863, opened 2023-04-06
Panicked if type(self)() in __del__ methods 


xiaxinmeng:
This code creates an instance of the Dummy class and then immediately deletes it, which triggers the __del__ method to create a new instance of the Dummy class. 

```
class Dummy:
   def __del__(self):
      type(self)()

d = Dummy()
del d
```

Crash message:
```
Exception ignored in: <object function repr() failed>
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: PyBaseException', vm/src/stdlib/sys.rs:593:59
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Environment:
Ubuntu 18.04
rustpython v0.2.0

itsankitkp:
It gets stuck in cpython (3.12a)

DimitrisJim:
we should probably hang too? This will continuously try and call `del` since it's creating a new `d` and then immediately discarding it, triggering a loop.

The call to `unwrap` definitely needs fixing though.

youknowone:
In general, Panic is better than hang. Both are useless but hanging is silent.

fanninpm:
@xiaxinmeng What is the intended result in CPython? This seems like you want the interpreter to hang.

DimitrisJim:
Yeah, the correct behaviour here is to hang due to this continuously calling `__del__`.

youknowone:
reported as fixed by @changjoon-park

PR #4864, opened 2023-04-06
handle panic if str is not defined in unraisablehook

itsankitkp:
This PR aims to handle panic in `unraisablehook`
and thus fixes panic issue as mentioned in https://github.com/RustPython/RustPython/issues/4863
Note: It doesn't fixes underlying issue, as I am not sure what is expected in given scenario (cpython hangs if inputs as per as #4863 ).
Also, I am not sure what to put as default in unwrap_or_else.

### DIFF
--- a/vm/src/stdlib/sys.rs
+++ b/vm/src/stdlib/sys.rs
@@ -590,7 +590,7 @@ mod sys {
     fn unraisablehook(unraisable: UnraisableHookArgs, vm: &VirtualMachine) {
         if let Err(e) = _unraisablehook(unraisable, vm) {
             let stderr = super::PyStderr(vm);
-            let default = vm.ctx.intern_str("{}").to_owned();
+            let default = vm.ctx.intern_str("").to_owned();
             writeln!(
                 stderr,
                 "{}",
--- a/vm/src/stdlib/sys.rs
+++ b/vm/src/stdlib/sys.rs
@@ -593,7 +593,10 @@ mod sys {
             writeln!(
                 stderr,
                 "{}",
-                e.as_object().repr(vm).unwrap_or_else(|_| vm.ctx.empty_str.to_owned()).as_str()
+                e.as_object()
+                    .repr(vm)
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/sys.rs
```diff
@@ -590,7 +590,12 @@ mod sys {
     fn unraisablehook(unraisable: UnraisableHookArgs, vm: &VirtualMachine) {
         if let Err(e) = _unraisablehook(unraisable, vm) {
             let stderr = super::PyStderr(vm);
-            writeln!(stderr, "{}", e.as_object().repr(vm).unwrap().as_str());
+            let default = vm.ctx.intern_str("").to_owned();
+            writeln!(
+                stderr,
+                "{}",
+                e.as_object().repr(vm).unwrap_or(default).as_str()
```
DimitrisJim, line 597: it might make sense to try and find out what CPython uses here, it might very well be `""` if a repr can't be found.

youknowone, line 597:
```suggestion
                e.as_object().repr(vm).unwrap_or_else(|| vm.ctx.empty_str.to_owned()).as_str()
```
We have a shared empty_str

youknowone: I didn't check cpython side, but if it hangs on CPython, would those kind of thing exist?

DimitrisJim: I was thinking more of a general "what if repr somehow can't be found" case but dunno if that ever happens. I guess `""` would be a sensible default here.

youknowone: I also guess this is just fine for unraisable context - or maybe something repr-error notification.

itsankitkp: Fixed

DimitrisJim (approved):
a `cargo fmt --all` and should be gtg, thanks!

itsankitkp:
> cargo fmt --all

Done

PR #4609, opened 2023-03-02
baseline to start locale support for formatting :n

youknowone:
was this ubuntu-only failure? i have no idea. but macos is fine.

PR #4865, opened 2023-04-07
'n' support for float format

JazzGlobal:
Contributes to #1656

I've only recently sprouted my Rustacean legs so let me know if there are better ways to achieve this. I'm sure there will be!

```py
'{:n}'.format(999999.1234)
'{:n}'.format(9999.1234)
'{:n}'.format(1000000.1234)
'{:n}'.format(-1000000.1234)
'{:.8n}'.format(1000000.1234)
'{:.10n}'.format(1000000.1234)
'{:.11n}'.format(1000000.1234)
'{:.11n}'.format(-1000000.1234)
'{:0n}'.format(-1000000.1234)
'{:n}'.format(-1000000.1234)
'{:-1n}'.format(-1000000.1234)

'{:-1N}'.format(-1000000.1234)
'{:.1N}'.format(1000000.1234)
'{:0N}'.format(-1000000.1234)
'{:N}'.format(999999.1234)

```
yields

```
999999
9999.12
1e+06
-1e+06
1000000.1
1000000.123
1000000.1234
-1000000.1234
-1e+06
-1e+06
-1e+06
Unknown format code 'N' for object of type 'float'
Unknown format code 'N' for object of type 'float'
Unknown format code 'N' for object of type 'float'
Unknown format code 'N' for object of type 'float'
```


JazzGlobal:
This will probably need tests before it gets accepted, but I could not successfully run the tests in Lib/test/test_format.py ... 

I used `cargo run` using the absolute path and also `cargo run --release` using the absolute path. Am I going about this incorrectly?

fanninpm:

```shell
```

fanninpm:
To format the Rust code, try running `cargo fmt --all`.

JazzGlobal:
Something's not quite right with either the tests I added for the N case or the action. I'm not getting local test failures at least.

JazzGlobal:
> Something's not quite right with either the tests I added for the N case or the action. I'm not getting local test failures at least.

I think I got everything figured out. Thanks for your help. A lot of my issues were because of my dev environment but I got them sorted out. 

The issue with the test was caused by my assuming Python 3.11 supported the N float formatter code but it doesn't. Removed the functionality for that and the test should yield the same Error Py 3.11 does when N is used.

DimitrisJim:
there's a passing test now in `test_enum.py` so :tada: : 

```bash
UNEXPECTED SUCCESS: test_format_enum_float (test.test_enum.TestEnum.test_format_enum_float)
```

there's also a panic in the formatting, looking at [the runner logs](https://github.com/RustPython/RustPython/actions/runs/4648587868/jobs/8228793225?pr=4865#step:12:7798) it seems to be due to something being off during a slicing operation.

This test doesn't run on Windows (hence why you did not probably see this error). It blows up in `test_locale` of `test_format` and my guess would be `format(1234.5, "n")`. Try running that locally in the REPL and see what it does.

JazzGlobal:
> there's a passing test now in `test_enum.py` so 🎉 :
> 
> ```shell
> UNEXPECTED SUCCESS: test_format_enum_float (test.test_enum.TestEnum.test_format_enum_float)
> ```
> 
> there's also a panic in the formatting, looking at [the runner logs](https://github.com/RustPython/RustPython/actions/runs/4648587868/jobs/8228793225?pr=4865#step:12:7798) it seems to be due to something being off during a slicing operation.
> 
> This test doesn't run on Windows (hence why you did not probably see this error). It blows up in `test_locale` of `test_format` and my guess would be `format(1234.5, "n")`. Try running that locally in the REPL and see what it does.

Thank you for the feedback! I think I fixed it and I also unmarked the unexpected success

youknowone:
You also fixed more test
```
UNEXPECTED SUCCESS: test_locale (test.test_format.FormatTest.test_locale)
```

and more failure
```
FAIL: test_int__format__ (test.test_types.TypesTests.test_int__format__)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/test_types.py", line 373, in test_int__format__
    self.assertRaises(ValueError, 0 .__format__, format_spec)
AssertionError: ValueError not raised by __format__
```

DimitrisJim:
> Can't run the tests on my dev environment

Generally good idea to push here and wait for results since other platforms might have issues you can't catch, you only need to be able to parse the logs (which is straight-forward after doing it once or twice but can get slightly tedious).

JazzGlobal:
> > Can't run the tests on my dev environment
> 
> Generally good idea to push here and wait for results since other platforms might have issues you can't catch, you only need to be able to parse the logs (which is straight-forward after doing it once or twice but can get slightly tedious).

Glad to hear that, hopefully I won't hit your GitHub action minute limit haha. Can't recall if there's only limitations in private repos or not. 

Forgot to run `cargo fmt` so the workflow will have to run at least once more.

common/src/format.rs
```diff
@@ -430,11 +432,17 @@ impl FormatSpec {
...
             | Some(FormatType::Hex(_))
             | Some(FormatType::String)
-            | Some(FormatType::Character) => {
+            | Some(FormatType::Character)
+            | Some(FormatType::Number(Case::Upper)) => {
                 let ch = char::from(self.format_type.as_ref().unwrap());
                 Err(FormatSpecError::UnknownFormatCode(ch, "float"))
             }
-            Some(FormatType::Number) => Err(FormatSpecError::NotImplemented('n', "float")),
+            Some(FormatType::Number(Case::Lower)) => {
```
youknowone, line 440: By document, `n` seems to same as `g` except for seperators.
Can we reuse `g` implementation? 

https://docs.python.org/3/library/string.html#formatspec

Otherwise, if it is a sort of scientific notation, can we reuse `:e` format in Rust?

JazzGlobal: Reused g's implementation and everything lines up for me locally but I'm not 100% sure how to test the locale based separators. Gonna see what the workflow thinks when it runs the tests.

DimitrisJim: triggered the run now, locale based separators could be a future addition (i.e no need to do everything in one PR :smile:)

fanninpm: > i.e no need to do everything in one PR 😄

Rome wasn't built in a day.

youknowone (approved):
Thank you for contributing!

youknowone:
@JazzGlobal If you are interested in supporting locale, you can start with #4609

JazzGlobal:
> @JazzGlobal If you are interested in supporting locale, you can start with #4609

Thank you, and thank you all for guiding me through my first Rust contribution 😁

PR #4870, opened 2023-04-10
Add bash script to autotest cpython lib

MegasKomnenos:

The following PR contains a bash script to automate the initial testing workflow for updating python library components from cpython. 

The workflow includes selecting and updating an individual library component, then running the related test to search for bugs. I realized that this work can be automated and be run for every library components, to screen out ones that fail. I wrote a bash script that does that, and made this PR.


scripts/clib/readme.txt
```diff
@@ -0,0 +1,11 @@
+To use, write down your local cpython path in clib_path.txt, and run,
+bash clib_test_one.sh
+
+and check the result in clib_out.txt
+
+The script will try to test every component in clib_list.txt, where,
+
+clib_test_one.sh : test each component individually
+clib_test_all.sh : test every components simultaneously
```
fanninpm, line 9: Grammar nitpick (adjective-noun agreement):
```suggestion
clib_test_all.sh : test all components simultaneously
```
or:
```suggestion
clib_test_all.sh : test every component simultaneously
```

scripts/clib/clib_test_all.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
```
youknowone, line 1: shell is good enough, but if possible, can this script be a python script? Practically, we usually struggle with windows tests because its different from unix.
But running bash script on windows is not easy.

DimitrisJim: plus maintainability. Sure we could get issues fixed but if possible we should limit the # of languages we use (as much as possible).

scripts/clib/clib_path.txt
fanninpm, line 1: I'm not sure this file should be checked into the repository, as it could vary from user to user.

scripts/clib/readme.txt
```diff
@@ -0,0 +1,11 @@
+To use, write down your local cpython path in clib_path.txt, and run,
```
DimitrisJim, line 1: could better yet be an `.md` for auto rendering by github.

scripts/clib/readme.txt
MegasKomnenos: Hehe, thanks! Grammar has never been my strong suit :D

scripts/clib/clib_test_all.sh
MegasKomnenos: I think that could be doable, as python supports file IO system calls, and I think it should also be possible to use os.system() to run cargo for building crates.

One question, would that python script be run in RustPython, or just python?

scripts/clib/clib_list.txt
```diff
@@ -0,0 +1,156 @@
...
+warnings
+weakref
+webbrowser
+xdrlib
+zipapp
+zipfile
+zipimport
+
+# Components listed below are excluded from being tested
+signal
```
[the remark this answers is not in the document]
MegasKomnenos: As in, something like this?

shutil
# signal
threading

scripts/clib/clib_path.txt
MegasKomnenos: That is a fair point, it is indeed something that would vary from user to user. I thought I'd leave it there as a sort of guideline for someone trying to use it, but I guess that could also go into readme.

scripts/clib/readme.txt
MegasKomnenos: :+1:

scripts/clib/clib_list.txt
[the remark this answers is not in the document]
fanninpm: Yes, exactly.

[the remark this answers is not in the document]
MegasKomnenos: I can see why that would be advantageous. I implemented it like this mainly for ease of implementation, but that is a better way. Let me work on the fix real quick.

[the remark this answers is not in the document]
MegasKomnenos: Done!

scripts/clib/clib_path.txt
youknowone: If you are going to use python, argparse is a good tool to manage command line arguments

youknowone: I agree about it. Adding a default argument will be a good option.

youknowone: Or just printing a sample command is another good option. User can learn from error messages quickly.

MegasKomnenos: I think both can work together, where there's a default argument value and an error message that shows up when cpython cannot be found in a provided path

scripts/clib/clib_test_all.sh
youknowone: Just python will be ok.

MegasKomnenos: > Just python will be ok. We have to fix rustpython to run any python script, eventually.

Done!

youknowone:
Thank you! let me test it a bit.

youknowone (changes requested):

youknowone:
Any suggestion about the script name? `clib` doesn't give good sense what the script does.
We may be able to turn this script to auto-updating trivial cpython library easily.

cc @fanninpm @DimitrisJim

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+import argparse
+import shutil
+import subprocess
+
+def main():
+    parser = argparse.ArgumentParser(description="Test cpython library")
+    parser.add_argument("-cp", nargs=1, default="../../../cpython", required=False, help="Local cpython path.")
+    parser.add_argument("-rp", nargs=1, default="../..", required=False, help="Local RustPython path.")
+    args = vars(parser.parse_args())
+    CPYTHONPATH = args["cp"]
```
youknowone, line 11:
```suggestion
    CPYTHON_PATH = args["cp"]
```

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+    RPYTHONPATH = args["rp"]
```
youknowone, line 12:
```suggestion
    RPYTHON_PATH = args["rp"]
```

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
+import os
+import argparse
+import shutil
+import subprocess
+
+def main():
+    parser = argparse.ArgumentParser(description="Test cpython library")
+    parser.add_argument("-cp", nargs=1, default="../../../cpython", required=False, help="Local cpython path.")
```
youknowone, line 8:
```suggestion
    parser.add_argument("--cpython", nargs=1, default="../../../cpython", required=False, help="Local cpython path.")
```

unix commonly uses `--` for long argument names.

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+    parser.add_argument("-rp", nargs=1, default="../..", required=False, help="Local RustPython path.")
```
youknowone, line 9:
```suggestion
    parser.add_argument("--rustpython", nargs=1, default="../..", required=False, help="Local RustPython path.")
```

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+    if isinstance(CPYTHONPATH, list):
+        CPYTHONPATH = CPYTHONPATH[0]
+    if isinstance(RPYTHONPATH, list):
+        RPYTHONPATH = RPYTHONPATH[0]
+
+    liblst = list()
+
+    with open("clib_list.txt", 'r') as f:
+        for line in f.readlines():
```
youknowone, line 22:
```suggestion
        for line in f:
```
This is just same

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+            line = line[:line.find('#')]
```
youknowone, line 23:
```suggestion
            line = line.split('#')[0]
```

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+            line = line.strip()
+            if line:
+                liblst.append(line)
+
+    open("clib_out.txt", 'w').close()
```
youknowone, line 28: instead of opening outfile, printing to stdout will be ok.
we can pipe it to write it to file

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+    for lib in liblst:
+        lib_exist = False
+        test_exist = False
+        test_do = True
+        message = [f"{lib}:"]
+
+        if os.path.isfile(f"{RPYTHONPATH}/Lib/{lib}.py"):
+            lib_exist = True
+            os.rename(f"{RPYTHONPATH}/Lib/{lib}.py", f"{RPYTHONPATH}/Lib/{lib}_tmp_mv.py")
```
youknowone, line 38: Rather than renaming original library files, redefining `RUSTPYTHONPATH` will be fine. 

This is the logic I think.

1. cp -r Lib LibTest  (but in python, not shell)
1. cp ../cpython/Lib/x LibTest/x
1. RUSTPYTHONPATH=LibTest rustpython Lib/test/test_...

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+            message.append(f"No cpython/Lib/test/test_{lib}.py")
+        
+        if test_do:
+            result = subprocess.run(["cargo", "run", "-q", f"{RPYTHONPATH}/Lib/test/test_{lib}.py"], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+            result = result.stdout.decode("utf-8")
+
+            if "OK" in result:
+                message.append("OK")
+            else:
+                message.append("Failed")
```
youknowone, line 62: maybe you want to define a dataclass and format the result later.

```python
from dataclasses import dataclass

@dataclass
class Item:
    lib_exist: bool
    test_exist: bool
    ...

   def to_string(self):
        return ...
```

scripts/clib/clib_test.py
MegasKomnenos: Huh. I did not know that. Thanks!

scripts/clib/clib_test.py
MegasKomnenos: Ah, that's a fair point. Yeah, that seems to be a more versatile design.

scripts/clib/clib_test.py
MegasKomnenos: Sure, that would be easy. But I am not fully grasping why that would be a better design. Is it because messing with library files via script may open room for possible human mistakes?

scripts/clib/readme.md
```diff
@@ -0,0 +1,9 @@
+To use, open cmd and run clib_test.py with args through python
+and check the result in clib_out.txt
+
+The script assumes that the script is being run from RustPython/scripts/clib,
+and that both RustPython and cpython project directories are located under a same parent directory, aka that they are siblings
+
+If either of those assumptions are false, then you must provide a correct path when running clib_test.py
+
+The script will try to test every component in clib_list.txt
```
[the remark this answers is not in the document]
MegasKomnenos: I copy-pasted the content of the issue page into a txt file, then used regex to filter out its contents

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
+import os
+import argparse
+import shutil
+import subprocess
+
```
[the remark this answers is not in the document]
MegasKomnenos: Sure, that would indeed be a necessary improvement.

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+    args = vars(parser.parse_args())
+    CPYTHONPATH = args["cp"]
+    RPYTHONPATH = args["rp"]
+
+    if isinstance(CPYTHONPATH, list):
+        CPYTHONPATH = CPYTHONPATH[0]
+    if isinstance(RPYTHONPATH, list):
+        RPYTHONPATH = RPYTHONPATH[0]
+
+    liblst = list()
```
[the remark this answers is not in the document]
MegasKomnenos: Ah, thanks for catching it. It's a bad habit of mine that I picked up from CP, where we try to shorten variable name as much as possible to the detriment of readability. I try to be self-conscious about it, but it often gets the better of me still.

scripts/clib/clib_test.py
```diff
@@ -0,0 +1,81 @@
...
+            message.append(f"No cpython/Lib/{lib}.py")
+
+        if os.path.isfile(f"{CPYTHONPATH}/Lib/test/test_{lib}.py"):
+            shutil.copyfile(f"{CPYTHONPATH}/Lib/test/test_{lib}.py", f"{RPYTHONPATH}/Lib/test/test_{lib}.py")
+        else:
+            test_do = False
+            message.append(f"No cpython/Lib/test/test_{lib}.py")
+        
+        if test_do:
+            result = subprocess.run(["cargo", "run", "-q", f"{RPYTHONPATH}/Lib/test/test_{lib}.py"], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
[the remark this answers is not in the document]
MegasKomnenos: But wouldn't it be necessary to do cargo build each time as we're changing its library files? Or maybe the interpreter functions independent of its library python files and simply references whatever is written there when it needs to? I'm not familiar with the structure of the project and I assumed the former to be true, but maybe it's not.

scripts/clib/clib_test.py
MegasKomnenos: I've tried to use RUSTPYTHONPATH environment variable to control which library files are referenced by rustpython interpreter but I seem to be doing something wrong, as it's working as if it's still referencing files in Lib, not LibTest. Could you guide me through how I could do what you suggested in python?

scripts/clib/clib_test.py
MegasKomnenos: That's a better way to implement comment. Thanks for letting me know!

scripts/clib/clib_test.py
MegasKomnenos: I tried to set environment variable through os.environ. Let me try again with a new method and see how it turns out.

scripts/clib/clib_test.py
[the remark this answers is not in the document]
MegasKomnenos: Huh. I'm seeing an odd change in behavior.  I'm comparing the previous result stored in clib_out.txt with the new result that I'm getting  from this change, and it seems to be slightly different? Some tests that failed before now  are now passing. Odd.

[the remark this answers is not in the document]
MegasKomnenos: [clib_out_new.txt](https://github.com/RustPython/RustPython/files/11348301/clib_out_new.txt)
[clib_out.txt](https://github.com/RustPython/RustPython/files/11348304/clib_out.txt)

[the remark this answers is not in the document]
MegasKomnenos: I believe that this command is not working as intended.

result = subprocess.run(
                [RUSTEXECPATH, "-q", self.path_tpython_test],
                stdout=subprocess.PIPE, 
                stderr=subprocess.STDOUT,
                env={"RUSTPYTHONPATH": "/home/kjh/20231/rust/RustPython/LibTest"}
            )

When I run it with the Lib directory removed, it returns ModuleNotFoundError

DimitrisJim:
do we have any updates on the state of this PR? It's been sorta stale for 3 weeks now and we should try and get your work through soon if possible.

youknowone:
My bad, I forgot about this after getting in the parser rally.
The whole idea is awesome. I need to clean up its interface more fit in our other scripts and naming so on.

youknowone:
@DimitrisJim do you have a good name suggestion instead of clib? it is not a C something.

PR #4872, opened 2023-04-12
feature: `PyTraverse` derive macro for traverse object's childrens(like CPython's `tp_traverse`)

discord9:
added a `PyTrace` derive macro for traverse an object's childrens, this is from CPython's similar function: https://docs.python.org/3/c-api/typeobj.html#c.PyTypeObject.tp_traverse
use on struct with named fields like `struct A{x:PyRef<B>, y:PyRef<C>}` to impl `Trace` for datatype.
use `#[notrace]` on fields you wish not to trace
add `trace` attr to `#[pyclass]` to make it impl `MaybeTrace` that will call `Trace`'s `trace` method so make it
traceable(Even from type-erased PyObject)(i.e. write `#[pyclass(trace)]`).
# Example
```rust
#[pyclass(module = false, trace)]
#[derive(Default, PyTrace)]
pub struct PyList {
    elements: PyRwLock<Vec<PyObjectRef>>,
    #[notrace]
    len: AtomicCell<usize>,
}
```
This create 
- `MaybeTrace` that call `Trace`'s `trace` method and 
- `PyTrace` that impl `Trace` for `PyList` which call elements' `trace` method and ignore `len` field.


discord9:
> does it use both derive(PyTrace) and pyclass(trace)? I expected only `derive(PyTrace)` could fully cover it.

Yes, this is for supporting a Optional trait bound, like all PyPayload type impl `MaybeTrace` but may or may not impl `Trace`(the actual tracing trait), for that, `pyclass` macro need to be able to dispatch call to actual `trace` if `Trace` is impl (Whether it's manually or through a `#derive(PyTrace)`) and do nothing if `Trace` is not present, and since macro can't ask type system whether a certain type impl a trait, this is the method I choose to do it, however I am open to other more elegant solution

youknowone:
I am not sure this is best, but a quick idea is using 2 derive macros.
The default implementation of MaybeTrace can be derived by `#[derive(PyTrace)]`.
And the blank implementation can be derived by `#[derive(NoTrace)]`.

discord9:
> I am not sure this is best, but a quick idea is using 2 derive macros. The default implementation of MaybeTrace can be derived by `#[derive(PyTrace)]`. And the blank implementation can be derived by `#[derive(NoTrace)]`.

This current design covers with mininal changes to codes perhaps?
|         | pyclass | not pyclass|
|------|------|------|
| Trace| `pyclass(trace)`+`#[derive(PyTrace)]` | #[derive(PyTrace)] |
|Not Trace| no modify | no modify |

However with `#[derive(NoTrace)]` it's like, and it seems there is a awful amount of `pyclass` that don't need trace(about ~650 struct with pyclass that don't own a PyRef hence no need to trace):
|         | pyclass | not pyclass|
|------|------|------|
| Trace| #[derive(PyTrace)] | #[derive(PyTrace)] |
|Not Trace| #[derive(NoTrace)] | no modify |

stdlib/src/array.rs
```diff
@@ -1424,12 +1424,13 @@ mod array {
...
     }
 
-    #[derive(FromArgs)]
+    #[derive(FromArgs, PyTraverse)]
     struct ReconstructorArgs {
         #[pyarg(positional)]
         arraytype: PyTypeRef,
         #[pyarg(positional)]
         typecode: PyStrRef,
+        #[notrace]
```
youknowone, line 1433:
```suggestion
        #[pytraverse = false]
```
I feel like this attr can be enhanced. Instead of using `PyTypeRef`, we can use the underlying type `PyRef<PyType>`. Because only `PyRef<T>` and `PyObjectRef` are ref-counted types, recognizing them and automatically marking `#[pytraverse = true]` maybe useful.
(Though still Arg* needs to be manually marked)

discord9: Types that can't contain PyRef/PyObjectRef don't need to be trace as them can't form a cycle by their own, this should reduce a lot of work, on the other hand wouldn't `#[pytraverse = false]` be a litte too verbose?


> I feel like this attr can be enhanced. Instead of using `PyTypeRef`, we can use the underlying type `PyRef<PyType>`. Because only `PyRef<T>` and `PyObjectRef` are ref-counted types, recognizing them and automatically marking `#[pytraverse = true]` maybe useful. (Though still Arg* needs to be manually marked)

derive-impl/src/pyclass.rs
```diff
@@ -413,8 +413,65 @@ pub(crate) fn impl_pyclass(attr: AttributeArgs, item: Item) -> Result<TokenStrea
...
+            quote! {
+                impl ::rustpython_vm::object::gc::MaybeTraverse for #ident {
+                    const IS_TRACE: bool = true;
+                    fn try_traverse(&self, tracer_fn: &mut ::rustpython_vm::object::gc::TraverseFn) {
+                        ::rustpython_vm::object::gc::Traverse::traverse(self, tracer_fn);
+                    }
+                }
+            },
+            quote! {#[derive(PyTraverse)]},
+        ),
```
youknowone, line 454:
```suggestion
        Some(value) => {
            let derive_trace = match value {
                Some(s) => {
                    if !ALLOWED_TRAVERSE_OPTS.contains(&s.as_str()) {
                        bail_span!(
                            item,
                            "trace attribute only accept {ALLOWED_TRAVERSE_OPTS:?} as value or no value at all",
                        );
                    }
                    quote! {}
                }
                None => quote! {#[derive(PyTraverse)]},
            };
            (
                quote! {
                    impl ::rustpython_vm::object::gc::MaybeTraverse for #ident {}
                },
                derive_trace
            )
        }
```

can MaybeTraverse have traverse as its default implementation? It will reduce generated code footprint

derive-impl/src/util.rs
```diff
@@ -178,6 +178,27 @@ impl ItemMetaInner {
...
+                    lit: syn::Lit::Str(lit),
+                    ..
+                }) => Some(lit.value()),
+                Meta::Path(_) => None,
+                _ => bail_span!(meta, r#"Expect `trace` or `trace = "manual""#),
+            };
+            Ok(Some(inner))
+        } else {
+            Ok(None)
+        }
```
youknowone, line 199: The new syntax help to reduce indent.
btw, I think this helper is not general to reuse from other place.
combination of `_has_key` and `_optional_str` will be more useful and easy to understand.
(`Option<Option<T>>` is not very human friendly and it doesn't give more information than `_has_key` and `_optional_str`)

```suggestion
        let Some((_, meta)) = self.meta_map.get(key) else {
            return Ok(None);
        };
        let inner = match meta {
            Meta::NameValue(syn::MetaNameValue {
                lit: syn::Lit::Str(lit),
                ..
            }) => Some(lit.value()),
            Meta::Path(_) => None,
            _ => bail_span!(meta, r#"Expect `trace` or `trace = "manual""#),
        };
        Ok(Some(inner))
```

derive-impl/src/util.rs
```diff
@@ -178,6 +178,27 @@ impl ItemMetaInner {
...
+    /// key doesn't exist -> Ok(None)
+    pub fn _optional_key_with_optional_str(&self, key: &str) -> Result<Option<Option<String>>> {
+        if let Some((_, meta)) = self.meta_map.get(key) {
+            let inner = match meta {
+                Meta::NameValue(syn::MetaNameValue {
+                    lit: syn::Lit::Str(lit),
+                    ..
+                }) => Some(lit.value()),
+                Meta::Path(_) => None,
+                _ => bail_span!(meta, r#"Expect `trace` or `trace = "manual""#),
```
youknowone, line 194: this is ItemMetaInner but error message is about trace.

stdlib/src/array.rs
youknowone: Cancelling about other parts, but still vote for `#[pytraverse = false]` or `#[pytraverse(false)]`.
keeping using same name `pytraverse` or `Traverse` will be helpful to recognize what it is.

derive/src/lib.rs
```diff
@@ -91,3 +91,27 @@ pub fn pypayload(input: TokenStream) -> TokenStream {
...
+/// #[derive(Default, PyTraverse)]
+/// pub struct PyList {
+///     elements: PyRwLock<Vec<PyObjectRef>>,
+///     #[notrace]
+///     len: AtomicCell<usize>,
+/// }
+/// ```
+/// This create both `MaybeTraverse` that call `Traverse`'s `traverse` method and `PyTraverse` that impl `Traverse`
+/// for `PyList` which call elements' `traverse` method and ignore `len` field.
+#[proc_macro_derive(PyTraverse, attributes(notrace))]
```
youknowone, line 113:
```suggestion
#[proc_macro_derive(Traverse, attributes(pytraverse))]
```
we don't do `PyFromArgs` but `FromArgs`. `Traverse` will be enough.

vm/src/object/gc/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod object;
+mod traverse;
```
youknowone, line 2: though traverse is only used for gc, I think traverse itself is a bit out of gc's implementation detail.
I'd like to suggest `vm/src/object/traverse.rs`.
it will be also helpful to enable traverse code even when disabling gc

vm/src/object/gc/traverse.rs
```diff
@@ -0,0 +1,231 @@
...
+    (C, 2),
+    (D, 3),
+    (E, 4),
+    (F, 5),
+    (G, 6),
+    (H, 7),
+    (I, 8),
+    (J, 9),
+    (K, 10),
+    (L, 11)
```
youknowone, line 230: wow.. do we really have this long tuple? lol

youknowone (changes requested):

derive-impl/src/pyclass.rs
fanninpm: @youknowone would something like this work (rewriting the inner `match` to be an `if let ... else`)?

```rust
Some(value) => {
    let derive_trace = if let Some(s) = value {
        if !ALLOWED_TRAVERSE_OPTS.contains(&s.as_str()) {
            bail_span!(
                item,
                "trace attribute only accept {ALLOWED_TRAVERSE_OPTS:?} as value or no value at all",
            );
        }
        quote! {}
    } else {
        quote! {#[derive(PyTraverse)]}
    };
}
```

youknowone: sure, that looks better.

stdlib/src/array.rs
discord9: > Cancelling about other parts, but still vote for `#[pytraverse = false]` or `#[pytraverse(false)]`. keeping using same name `pytraverse` or `Traverse` will be helpful to recognize what it is.

How about `#[pytraverse(no)]` or `#[pytraverse(skip)]` on fields of struct? Feeling `#[pytraverse = false]` maybe not easily extendable for future

discord9:
Please take a look? I think most problems mentioned above should be fixed now?

youknowone (commented):
Looking really good.
I feel like this almost final review.

vm/src/builtins/iter.rs
```diff
@@ -24,12 +25,27 @@ pub enum IterStatus<T> {
...
 pub struct PositionIterInternal<T> {
     pub status: IterStatus<T>,
     pub position: usize,
 }
 
+unsafe impl<T: Traverse> Traverse for PositionIterInternal<T> {
+    fn traverse(&self, tracer_fn: &mut TraverseFn) {
+        self.status.traverse(tracer_fn)
+    }
+}
```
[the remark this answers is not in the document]
discord9: Currently this macro didn't deal with generic arguement so using`#[derive(Traverse)]` would generate code without generic parameters

> IterStatus is enum but this is struct. is this not able to derive?

vm/src/function/argument.rs
```diff
@@ -58,20 +58,17 @@ into_func_args_from_tuple!((v1, T1), (v2, T2), (v3, T3), (v4, T4), (v5, T5));
...
+#[derive(Debug, Default, Clone, Traverse)]
 pub struct FuncArgs {
     pub args: Vec<PyObjectRef>,
     // sorted map, according to https://www.python.org/dev/peps/pep-0468/
     pub kwargs: IndexMap<String, PyObjectRef>,
 }
 
-unsafe impl Traverse for FuncArgs {
+unsafe impl<K, V: Traverse, S> Traverse for IndexMap<K, V, S> {
+    /// FIXME(discord9): what about `K: Traverse`?
```
youknowone, line 69: If it is worried, implementing Traverse only for `PyAttributes`(=`IndexMap<&'static PyStrInterned, PyObjectRef>`) and `IndexMap<String, PyObjectRef>` will prevent it. And those are the only types traverse might can step in.

vm/src/builtins/type.rs
```diff
@@ -114,6 +115,12 @@ cfg_if::cfg_if! {
 /// faster and only supports strings as keys.
 pub type PyAttributes = IndexMap<&'static PyStrInterned, PyObjectRef, ahash::RandomState>;
 
+unsafe impl Traverse for IndexMap<&'static PyStrInterned, PyObjectRef, ahash::RandomState> {
```
youknowone, line 118:
```suggestion
unsafe impl Traverse for PyAttributes {
```

derive-impl/src/pyclass.rs
discord9: > can MaybeTraverse have traverse as its default implementation? It will reduce generated code footprint

That might be not possible until negative impl or optional impl is available? Like T: Traverse for MaybeTraverse & T: !Traverse for MaybeTrace?

youknowone (approved):
Thank you so much for the long time effort!
I wished it could be reviewed by more reviewers, but they seem busy.

youknowone:
Ah, I found *Args are implementing `Traverse`. They are only used for argument, so it will never have change to be called from GC cycles.

PR #4874, opened 2023-04-14
Fork RustPyton and change `Location` to `TextSize`

MichaReiser:
This PR replaces RustPython's `Location` with `TextSize`. 

See https://github.com/charliermarsh/ruff/pull/3931 for the motivation


Cargo.toml
```diff
@@ -1,19 +1,18 @@
...
-name = "rustpython"
-version = "0.2.0"
-authors = ["RustPython Team"]
-edition = "2021"
-rust-version = "1.67.1"
-description = "A python interpreter written in rust."
-repository = "https://github.com/RustPython/RustPython"
-license = "MIT"
-include = ["LICENSE", "Cargo.toml", "src/**/*.rs"]
+#[package]
```
MichaReiser, line 1: I removed all unused crates from the workspace setup rather than deleting them. This should make it easier for us to sync with upstream (rather than dealing with delete merge conflicts all the time).

compiler/ast/Cargo.toml
```diff
@@ -11,10 +11,8 @@ license = "MIT"
 default = ["constant-optimization", "fold"]
 constant-optimization = ["fold"]
 fold = []
-unparse = ["rustpython-common"]
 
 [dependencies]
-rustpython-compiler-core = { path = "../core", version = "0.2.0" }
-rustpython-common = { path = "../../common", version = "0.2.0", optional = true }
+ruff_text_size =  { path = "../../ruff_text_size" }
```
MichaReiser, line 16: I decided to move `ruff_text_size` into this repository because:

* I want to avoid git submodules in our main repository
* `ruff_text_size` should rarely change

MichaReiser:
Whoops, wrong target repository. I'm sorry

youknowone:
I didn't check how `TextRange` looks like, but the traditional design to achieve it is using a record(struct) with 3 fields - row, column and byte offset. It allows to access both raw text data and human friendly location in constant time.
If you are interested in making patch again, we are very open to those changes.

MichaReiser:
> I didn't check how `TextRange` looks like, but the traditional design to achieve it is using a record(struct) with 3 fields - row, column and byte offset. It allows to access both raw text data and human friendly location in constant time. If you are interested in making patch again, we are very open to those changes.

We aim to remove the row and column indices entirely and replace them with byte offsets only because our benchmarks show that this improves the parser performance (less data that needs to be read/written). This has the downside that we have to recompute the row/column numbers when showing diagnostics but most projects have no diagnostics, which is why this seems a worthwhile trade. 

I initially wanted to add this as a feature flag to `RustPython` but it breaks too many upstream crates and e.g. wouldn't allow us to also change `Located::end` from ` Option<Location>` to `Location`

Issue #4320, opened 2022-12-09
Support deployment to wasm32-unknown-unknown without wasm-bindgen


lastmjs:
wasm32-unknown-unknown != wasm-bindgen support.

This is one of the main problems we have with RustPython, as the environment we are deploying to is a non-browser strict wasm32-unknown-unknown environment without wasm-bindgen support.

RustPython by default expects wasm32-unknown-unknown to support wasm-bindgen, thus in our fork of RustPython we have to make the following changes that can be seen in this pull request: https://github.com/demergent-labs/RustPython/pull/1/files

## 1

In `vm/Cargo.toml`:

We need to change this:

```
getrandom = { version = "0.2.6", features = ["js"] }
```

To this:

```
getrandom = { version = "0.2.6", features = ["custom"] }
```

The `js` feature assumes wasm-bindgen support.

## 2

In `vm/Cargo.toml`:

We need to change this:

```
chrono = { version = "0.4.19", features = ["wasmbind"] }
```

To this, or to a newer version without the `wasmbind` feature:

```
chrono = { version = "=0.4.19" }
```

## 3

In `vm/Cargo.toml` we need to comment out the following:

```
[target.'cfg(target_arch = "wasm32")'.dependencies]
wasm-bindgen = "0.2.80"
```

## 4

In `vm/src/stdlib/time.rs` we need to comment out everything from line 94-102, basically get rid of all of the `wasm_bindgen` stuff. We also need a way to provide our own time functionality, which comes from a crate called `ic_cdk`, which has an API `ic_cdk::api::time` that will give us system time for our environment.

## 5

In `vm/src/vm/vm_object` we need to comment out all `wasm_bindgen` functionality, from lines 45-55.

Assuming `wasm-bindgen` support requires us to maintain a fork with these many changes, it would be great to upstream the ability to turn off wasm-bindgen so that our fork can be much smaller or even disappear.

Thanks!

youknowone:
We'd better to add "wasm-bindgen" feature and disable them without the features.
On the other hand, we need to package the dependency reversed way to allow outer project can inject their own implementation. And then wasm-bindgen impl can be one of them.

lastmjs:
We're going to attempt a PR for this

PR #4875, opened 2023-04-14
add wasmbind feature

bdemann:
Resolves https://github.com/RustPython/RustPython/issues/4320

### DIFF
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -89,21 +89,17 @@ mod time {
         Ok(duration_since_system_now(vm)?.as_secs_f64())
     }
 
-    #[cfg(all(target_arch = "wasm32", not(target_os = "wasi")))]
+    #[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind"))]
     fn _time(_vm: &VirtualMachine) -> PyResult<f64> {
-        #[cfg(feature = "wasmbind")]
-        {
-            use wasm_bindgen::prelude::*;
-            #[wasm_bindgen]
-            extern "C" {
-                type Date;
-                #[wasm_bindgen(static_method_of = Date)]
-                fn now() -> f64;
-            }
-            // Date.now returns unix time in milliseconds, we want it in seconds
-            return Ok(Date::now() / 1000.0);
+        use wasm_bindgen::prelude::*;
+        #[wasm_bindgen]
+        extern "C" {
+            type Date;
+            #[wasm_bindgen(static_method_of = Date)]
+            fn now() -> f64;
         }
-        panic!("Date not available")
+        // Date.now returns unix time in milliseconds, we want it in seconds
+        return Ok(Date::now() / 1000.0);
     }
 
     #[pyfunction]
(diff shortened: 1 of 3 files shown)


vm/src/stdlib/time.rs
```diff
@@ -91,15 +91,19 @@ mod time {
...
+                type Date;
+                #[wasm_bindgen(static_method_of = Date)]
+                fn now() -> f64;
+            }
+            // Date.now returns unix time in milliseconds, we want it in seconds
+            return Ok(Date::now() / 1000.0);
         }
-        // Date.now returns unix time in milliseconds, we want it in seconds
-        Ok(Date::now() / 1000.0)
+        panic!("Date not available")
```
fanninpm, line 106: Why panic?

bdemann: If you are compiling wasm32-unknown-unknown and not wasi then std time doesn't work, so there is no way to get a date generally as far as I know for for wasm32-unknown-unknown.

lastmjs: In our project, which must compile and run in a wasm32-unknown-unknown environment, std time doesn't work, and we have our own time implementation that we import from the host environment. We would love for a solution to this problem as well, but this PR is scoped for general use.

fanninpm: When the interpreter panics, the interpreter process comes to a complete emergency stop. From there, you have to start it again. Is this what you want?

vm/src/vm/vm_object.rs
```diff
@@ -26,15 +26,18 @@ impl VirtualMachine {
...
+                    fn error(s: &str);
+                }
+                let mut s = String::new();
+                self.write_exception(&mut s, &exc).unwrap();
+                error(&s);
             }
-            let mut s = String::new();
-            self.write_exception(&mut s, &exc).unwrap();
-            error(&s);
             panic!("{}; exception backtrace above", msg)
```
fanninpm, line 41: This will fail to compile if the `wasmbind` feature is not enabled.

vm/src/stdlib/time.rs
DimitrisJim: Typically, unsupported platforms in Python just don't expose the module/function/constant etc (the `os` module is an example of this). If the `time` module doesn't work for the target it would make more sense to just not have it be available.

Baring that, an exception would be the second best option.

youknowone: I also vote for hiding the function `time.time` when not available.
Then user can set it like `time.time = custom_time`

youknowone: Ah, I found `process_time` and `process_time_ns` raises error. `NowImplementedError` also will be ok.

youknowone:
oh, no. this one is not merged yet.

youknowone (commented):
additionally, rustpython_wasm will require the new feature to work.

I am really sorry for late review.
Since the patch is created from unwritable branch, I will finish it on a new PR #5390

youknowone:
Any idea to add a test not to break this again?

PR #4876, opened 2023-04-15
Update collections from Python 3.11.2

Phosphorescentt:
Also added an extra couple of lines to the test just to make sure everyting's working as intended.

Let me know if this looks good or if any changes are needed :)


Lib/test/test_ordered_dict.py
```diff
@@ -983,19 +983,19 @@ def test_popitem(self):
...
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_pop(self):
         c = self.type2test(3)
         for i in range(1, 4):
             c[i] = i
         self.assertEqual(c.counts, {'get': 0, 'set': 3, 'del': 0})
         self.assertEqual(c.pop(2), 2)
-        self.assertEqual(c.counts, {'get': 0, 'set': 3, 'del': 0})
+        self.assertEqual(c.counts, {'get': 1, 'set': 3, 'del': 1})
```
youknowone, line 992: we need to fix implementation, not the test.

youknowone (changes requested):

Lib/test/test_ordered_dict.py
DimitrisJim: Might be easy to miss if you're a first time contributor but: the tests are pretty much copied from the CPython implementation and only tweaked to signify a test failing for RustPython. The tests basically test 'do we do the same thing'.

So, when something fails, we need to tweak the implementation to get in line as YunWon says.

Phosphorescentt: Hey! Thanks for the feedback - definitely something I wouldn't have noticed without the inupt :)

All totally makese sense, but clearly I'm not understanding what the implementation is supposed to be doing? Is it that `c.pop()` is not meant to update the number of `get`s and `del`s? In other words are the `get` and `del` fields only meant to be updated if we're explicitly using either the `.get("field")` or `["field"]` and `del c["field"]`?

Either way happy to make the changes just need some pointing in the right direction :)

Lib/collections/__init__.py
```diff
@@ -240,11 +240,18 @@ def pop(self, key, default=__marker):
...
+            link = self.__map.pop(key)
+            link_prev = link.prev
+            link_next = link.next
+            link_prev.next = link_next
+            link_next.prev = link_prev
+            link.prev = None
+            link.next = None
             return result
-        if default is self.__marker:
+        if default is marker:
```
fanninpm, line 254: Is this present in CPython?

Phosphorescentt: Yep https://github.com/python/cpython/blob/main/Lib/collections/__init__.py#L246

Phosphorescentt:
Checked this implementation against the old one and it makes the test pass without causing any other ones to fail

Lib/collections/__init__.py
DimitrisJim: ah, so this bug surfaced because we update the test lib but not the `collections`? Pretty nice if so.

DimitrisJim:
Thanks! Do you mind updating all the files in `collections`? As far as I can tell, this just update the `__init__.py` file, right? It would be best if the PR updated them all so we can have all files be in sync. (See https://github.com/RustPython/RustPython/issues/4564 for reference here)

Phosphorescentt:
As far as I can tell there's nothing to be changed in either `collections/abc.py` or `collections/defaultdict.py` so I think we're good? Unless I've made another blunder which is very possible 😅

DimitrisJim:
Copying `__init__.py` over from the `3.11` branch (specifically `3.11.2`) shows a couple more changes that have been made to it (some in `OrderedDict`s `__reduce__`, some shifting of methods in `Counter` and some tiny comment changes). Maybe you've copied from an older version?

So, I think the best approach now is to include those changes from `3.11` here, rename the PR to `Update collections from Python 3.11.2` and we can see if the test suite picks anything else up.

Thanks again for opening, feel free to ask if anything gets confusing.

Phosphorescentt:
Made the suggested changes. Thanks for all the help :) 

Didn't realise so much of the ethos of this project is about staying in line with the CPython implementations (which of course reflecting on it makes *total* sense)!

Phosphorescentt:
Just checking the tests now and it seems like actually these changes fix another test that was failing! 

```
___________ PurePythonOrderedDictWithSlotsCopyingTests.test_copying ____________
Unexpected success
```

If we're happy that this test should be passing then I'll go ahead and take away the decorator :)

DimitrisJim:
I'm going to guess it most likely isn't happy with 8e55919cf which seems kinda odd considering your output. Since the runner's thirst for success must be appeased we should rollback that commit and give it a second try to make sure.

Phosphorescentt:
I think that's done? Still working on my git-fu

Phosphorescentt:
Had a dig through the CI logs and it seems like the following tests are failing in a number of test classes:
- `test_copying`
- `test_pickle_recursive`
- `test_yaml_linkage`

DimitrisJim:
Okay, failures do make sense, they all are a result of the new `__reduce__` for `OrderedDict` (which calls `__setstate__` which is nowhere to be found). This sometimes happens, library is updated, uses a feature we might not have, tests that worked fine with the old implementation now fail.

You can now mark these as expected failures and I should probably open an issue on that `__setstate__` which we've somehow missed.

Phosphorescentt:
Before I add those decorators, would it be worth removing all of the decorators and rerunning the workflows just so everything's fully up to daTE? 

Also seems like an issue with `__getstate__` too? Not sure why the tests are only failing on ubuntu and not on mac os though? The code for `__reduce__` makes a call to `__getstate__()` immediately, but, like you, I can't find a definition for `__getstate__()` on anything dictionary related yet the tests still pass.

DimitrisJim:
> Not sure why the tests are only failing on ubuntu and not on mac os though? 

Don't think they are executed on Mac. These are platform independent (don't really test any platform specific things) so having them executed on Ubuntu is a solution to cut down on unnecessary execution time.

Phosphorescentt:
> > Not sure why the tests are only failing on ubuntu and not on mac os though?
> 
> Don't think they are executed on Mac. These are platform independent (don't really test any platform specific things) so having them executed only on Ubuntu (fastest runner) is a solution to cut down on unnecessary execution time.

Fair enough - makes sense. Still not sure why this test passes on my laptop though.

> Before I add those decorators, would it be worth removing all of the decorators and rerunning the workflows just so everything's fully up to date and we can see which tests are going through according to the CI?

Opinions on this idea? Probably worth making sure that we know which tests are now passing and/or failing?

fanninpm:
> > Before I add those decorators, would it be worth removing all of the decorators and rerunning the workflows just so everything's fully up to date and we can see which tests are going through according to the CI?
> 
> Opinions on this idea? Probably worth making sure that we know which tests are now passing and/or failing?

It would be a good idea to remove RustPython-specific `@unittest.skip()` decorators to see if they still need to be skipped. The `@unittest.expectedFailure` decorators _don't_ need to be removed.

fanninpm:
> I can't seem to replicate the errors with `test_pickle_recursive`, `test_yaml_linkage`, `test_reduce_not_too_fat`

```shell
```

Phosphorescentt:
Ahhhh thank you for this!

Just updated the decorators so now all the tests should pass the CI.

Lib/test/test_ordered_dict.py
```diff
@@ -281,6 +281,7 @@ def test_equality(self):
         # different length implied inequality
         self.assertNotEqual(od1, OrderedDict(pairs[:-1]))
 
+    @unittest.expectedFailure
```
fanninpm, line 284:
```suggestion
    # TODO: RUSTPYTHON, SpamError: Have you got anything without spam?
    @unittest.expectedFailure
```

Lib/test/test_ordered_dict.py
```diff
@@ -324,6 +325,7 @@ def check(dup):
         check(update_test)
         check(OrderedDict(od))
 
+    @unittest.expectedFailure
```
fanninpm, line 328:
```suggestion
    # TODO: RUSTPYTHON, SpamError: Well, there's spam egg sausage and spam, that's not got much spam in it.
    @unittest.expectedFailure
```

Lib/test/test_ordered_dict.py
```diff
@@ -334,6 +336,7 @@ def test_yaml_linkage(self):
         # '!!python/object/apply:__main__.OrderedDict\n- - [a, 1]\n  - [b, 2]\n'
         self.assertTrue(all(type(pair)==list for pair in od.__reduce__()[1]))
 
+    @unittest.expectedFailure
```
fanninpm, line 339:
```suggestion
    # TODO: RUSTPYTHON, SpamError: I don't want ANY spam!
    @unittest.expectedFailure
```

Lib/test/test_ordered_dict.py
```diff
@@ -345,6 +348,7 @@ def test_reduce_not_too_fat(self):
         self.assertEqual(od.__dict__['x'], 10)
         self.assertEqual(od.__reduce__()[2], {'x': 10})
 
+    @unittest.expectedFailure
```
fanninpm, line 351:
```suggestion
    # TODO: RUSTPYTHON, SpamError: Why can't she have egg bacon spam and sausage?
    @unittest.expectedFailure
```

fanninpm (commented):
These `@unittest.expectedFailure` decorators aren't in the CPython source, so we should mark them as RustPython-specific.

youknowone (approved):
I found collections/__init__ and test_ordered_dict is not fully updated without comments.
Attaching a commit to resolve them.

DimitrisJim:
re-run of ci won't fly well with the recent windows ci changes it seems @Phosphorescentt can you rebase?

youknowone:
Thank you!

PR #4878, opened 2023-04-15
This is a draft to show our remaining reasons for having a fork of RustPython

lastmjs:
The wasm-bindgen stuff is being addressed here: https://github.com/RustPython/RustPython/pull/4875

Also in that PR we are talking about time, but the PR doesn't address a solution to that.


compiler/parser/src/python.rs
```diff
@@ -1,3 +1,3 @@
 #![allow(clippy::all)]
 #![allow(unused)]
-include!(concat!(env!("OUT_DIR"), "/python.rs"));
+include!("../python.rs");
```
DimitrisJim, line 3: could you give more context behind this?

vm/src/vm/interpreter.rs
```diff
@@ -22,7 +22,7 @@ use std::sync::atomic::Ordering;
 /// });
 /// ```
 pub struct Interpreter {
-    vm: VirtualMachine,
+    pub vm: VirtualMachine,
```
DimitrisJim, line 25: ditto here. What do you use from `vm` that requires it be public? Can we add an additional method/function to cover the usecase?

compiler/parser/src/python.rs
lastmjs: In [our environment](https://internetcomputer.org/) which is a decentralized Wasm cloud environment, there are function complexity checks that are performed before a Wasm binary is allowed to be deployed. The complexity limit is currently `15_000`, but the generated lolrpop python.rs file has a function called `__reduce` that is ~9000 lines long and has a complexity of `16_722`. So for now we need to modify the generated python.rs file and split up the `__reduce` function. We have asked the protocol engineers of the Wasm environment to increase the complexity limit, but if you know of how to get lolrpop to generate less complex functions that would be great too.

DimitrisJim: > but if you know of how to get lolrpop to generate less complex functions that would be great too.

Doesn't seem likely https://github.com/lalrpop/lalrpop/issues/304. Hopefully at some point we should get a PEG parser which shouldn't result in such a behemoth of a function.

youknowone: Idk if it is possible now, but if you can take input as `*.pyc` instead of `*.py`, skipping parsing from the embedded machine will be possible (maybe with little bit fix of rustphython).

vm/src/stdlib/time.rs
```diff
@@ -91,15 +91,19 @@ mod time {
...
+                #[wasm_bindgen(static_method_of = Date)]
+                fn now() -> f64;
+            }
+            // Date.now returns unix time in milliseconds, we want it in seconds
+            return Ok(Date::now() / 1000.0);
         }
-        // Date.now returns unix time in milliseconds, we want it in seconds
-        Ok(Date::now() / 1000.0)
+        panic!{"No date available"}
     }
```
fanninpm, line 107: This should be
```rust
#[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind"))]
fn _time(_vm: &VirtualMachine) -> PyResult<f64> {
    // ..
}
```

vm/src/vm/vm_object.rs
```diff
@@ -26,15 +26,18 @@ impl VirtualMachine {
...
+                }
+                let mut s = String::new();
+                self.write_exception(&mut s, &exc).unwrap();
+                error(&s);
             }
-            let mut s = String::new();
-            self.write_exception(&mut s, &exc).unwrap();
-            error(&s);
             panic!("{}; exception backtrace above", msg)
         }
```
fanninpm, line 42: This should be
```rust
#[cfg(all(target_arch = "wasm32", not(target_os = "wasi"), feature = "wasmbind"))]
{
    // ...
}
```

fanninpm (changes requested):
Generally, it's better to make a function disappear than to crash the interpreter.

@lastmjs Does your use case involve running in an interactive shell?

lastmjs:
> Generally, it's better to make a function disappear than to crash the interpreter.
> 
> @lastmjs Does your use case involve running in an interactive shell?

No we are not running in an interactive shell

vm/src/stdlib/time.rs
bdemann: Addressed in https://github.com/RustPython/RustPython/pull/4875

PR #4888, opened 2023-04-18
Add test_poll.py from Cpython v3.11.2

Masorubka1:


Lib/test/test_poll.py
```diff
@@ -0,0 +1,241 @@
...
+        pollster = select.poll()
+        pollster.register(1)
+
+        # Issues #15989, #17919
+        self.assertRaises(OverflowError, pollster.register, 0, USHRT_MAX + 1)
+        self.assertRaises(OverflowError, pollster.modify, 1, USHRT_MAX + 1)
+        self.assertRaises(OverflowError, pollster.poll, INT_MAX + 1)
+        self.assertRaises(OverflowError, pollster.poll, UINT_MAX + 1)
+
+    @unittest.skip("TODO: RUSTPYTHON")
```
fanninpm, line 186:
```suggestion
    @unittest.skip("TODO: RUSTPYTHON, I forget why we're skipping this. @Masorubka1 do you know why?")
```

Masorubka1: because when i run this test, program stacks in infinite execution. I think that the problem in negative values of timeout

Masorubka1: because rustpython fails filedescriptor reallocation
```python
# trigger ufds array reallocation
            for fd in rfds:
                pollster.unregister(fd)
```

fanninpm:
```suggestion
    @unittest.skip("TODO: RUSTPYTHON, fd reallocation")
```

youknowone: @fanninpm  adding reason will be nice, but I don't want to make them as burden for contributors

Masorubka1:
@fanninpm well, there is flap test(as i understand)[or some invalid pointers].
изображение

youknowone (approved):
Thank you!

PR #4890, opened 2023-04-18
Update test_posix.py from Cpython v3.11.2

Masorubka1:


Lib/test/test_posix.py
```diff
@@ -1490,7 +1514,7 @@ def test_unlink_dir_fd(self):
                 raise
 
     @unittest.skip("TODO: RUSTPYTHON; no os.mkfifo")
-    # @unittest.skipUnless(os.mkfifo in os.supports_dir_fd, "test needs dir_fd support in os.mkfifo()")
+    # @unittest.skipUnless(hasattr(os, 'mkfifo') and os.mkfifo in os.supports_dir_fd, "test needs dir_fd support in os.mkfifo()")
```
fanninpm, line 1517:
```suggestion
    @unittest.skipUnless(hasattr(os, 'mkfifo') and os.mkfifo in os.supports_dir_fd, "test needs dir_fd support in os.mkfifo()")
```

…and we don't need the previous line anymore!

Lib/test/test_posix.py
```diff
@@ -1490,7 +1514,7 @@ def test_unlink_dir_fd(self):
                 raise
 
     @unittest.skip("TODO: RUSTPYTHON; no os.mkfifo")
```
fanninpm, line 1516:
```suggestion
```

fanninpm (approved):

Issue #4067, opened 2022-08-14
support property name


youknowone:
## Feature

PyProperty can handle an optional name.

Adding `name: Option<PyStrRef>` will be a good starting point.

seungha-kim:
I'll take a look.

youknowone:
related to test `test_property_setname_on_property_subclass`

youknowone:
The test test_property_setname_on_property_subclass is fixed

PR #4892, opened 2023-04-18
Update test_property.py from Cpython v3.11.2

Masorubka1:


fanninpm:
See also #4067

youknowone (approved):

youknowone:
If you are interested in further contribution, fixing the getter/setter error message seems like good one.
searching the message from cpython source code will be helpful.

PR #4897, opened 2023-04-20
Reduce runtime of stdlib_collections_deque

youknowone:


fanninpm:
Has this been known to slow down the test suite appreciably? In the long term, it might be better to improve our implementation.

youknowone:
I am using pretty good machine but it takes 73 seconds on debug build.

youknowone:
I didn't check what it actually tests. Maybe we don't need it if we have same thing on cpython test suite

DimitrisJim:
there's probably some clean-up that can be done in `extra_tests` for sure.

youknowone:
Agree

PR #4912, opened 2023-04-21
Add more issue templates

youknowone:


fanninpm:
Can we convert these to [issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms)? I feel like using issue forms would encourage more focused reporting.

youknowone:
@fanninpm I didn't know about that. Could you help to do it? I thought current only 2 suggestions are confusing. So adding an empty form would be helpful for people who cannot choose one template.

DimitrisJim (approved):
I guess we can move these to issue forms altogether in another PR if needed. Until then, I think its good to add these for now.

DimitrisJim:
Let's go with these for now and tweak them later (if required!)

PR #4922, opened 2023-04-21
Fix wasm ci failure due to dns

itsankitkp:


.github/workflows/ci.yaml
```diff
@@ -473,6 +473,7 @@ jobs:
       - uses: Swatinem/rust-cache@v2
       - name: Setup Wasmer
         uses: wasmerio/setup-wasmer@v2
+        security: strict
```
fanninpm, line 476:
```suggestion
        with:
          security: strict
```

itsankitkp: Fixed

itsankitkp:
> I'd recommend opening an issue upstream to https://github.com/wasmerio/setup-wasmer.

Yes, issue lies on their side

itsankitkp:
Closing this PR as it is not something we can fix here

PR #4923, opened 2023-04-21
sqlite: Throw `TypeError` when `callable` is not callable

moreal:
## RustPython (Before this pull request, main)

```python
>>>>> import sqlite3; sqlite3.connect(':memory:').create_collation('X', 42)
```

(no output)

## CPython (3.11.3)

```python
>>> import sqlite3; sqlite3.connect(':memory:').create_collation('X', 42)
TypeError: parameter must be callable
```

## RustPython (After this pull request)

```python
>>>>> import sqlite3; sqlite3.connect(':memory:').create_collation('X', 42)
TypeError: parameter must be callable
```

## Links

 - https://github.com/python/cpython/blob/dc328d398d8f65ec6d2fa493e16ceee75f6d774a/Modules/_sqlite/connection.c#L2078-L2081

### DIFF
--- a/stdlib/src/sqlite.rs
+++ b/stdlib/src/sqlite.rs
@@ -1107,13 +1107,17 @@ mod _sqlite {
         ) -> PyResult<()> {
             let name = name.to_cstring(vm)?;
             let db = self.db_lock(vm)?;
-            let Some(data )= CallbackData::new(callable, vm) else {
+            let Some(data )= CallbackData::new(callable.to_owned(), vm) else {
                 unsafe {
                     sqlite3_create_collation_v2(db.db, name.as_ptr(), SQLITE_UTF8, null_mut(), None, None);
                 }
                 return Ok(());
             };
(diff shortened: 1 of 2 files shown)


stdlib/src/sqlite.rs
```diff
@@ -1107,14 +1107,18 @@ mod _sqlite {
         ) -> PyResult<()> {
             let name = name.to_cstring(vm)?;
             let db = self.db_lock(vm)?;
-            let Some(data )= CallbackData::new(callable, vm) else {
+            let Some(data )= CallbackData::new(callable.to_owned(), vm) else {
```
youknowone, line 1110:
```suggestion
            let Some(data) = CallbackData::new(callable.clone(), vm) else {
```
`to_owned` came from `&PyObject`.
We are going to call `PyRc::clone(callable)` by the meaning.

youknowone (approved):
Thank you!

PR #4926, opened 2023-04-23
Remove Traverse from FromArgs

youknowone:
@discord9 could you check this is safe?

In my opinion, *Args object never be referenced by any object but they instantly removed after a function call.
They will not have chance to be called by traverse.

### DIFF
--- a/stdlib/src/array.rs
+++ b/stdlib/src/array.rs
@@ -1434,13 +1434,12 @@ mod array {
         }
     }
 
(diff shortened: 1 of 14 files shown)


discord9:
Yes, if a certain type can't be reference by any object, or isn't a field of a certain `pyclass` type, then it's `traverse` couldn't be called anyway, this should make sense

discord9 (approved):

DimitrisJim (approved):

PR #4927, opened 2023-04-24
Rename method/member types

youknowone:
- PyMemberDef: Matches to CPython name and similar naming as `PyMethodDef`. `PyMehtodDef` is a python object in CPython
- PyMemberDescrObject -> PyMemberDescriptor
   - we don't use -Object suffix
   - we avoid abbr as much as possible

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -5,7 +5,7 @@ use super::{
 use crate::{
     builtins::{
         descriptor::{
-            DescrObject, MemberDescrObject, MemberGetter, MemberKind, MemberSetter, PyMemberDef,
+            DescrObject, MemberGetter, MemberKind, MemberSetter, PyMemberDef, PyMemberDescriptor,
         },
         function::PyCellRef,
         tuple::{IntoPyTuple, PyTupleTyped},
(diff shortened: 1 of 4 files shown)


youknowone:
cc @moreal

vm/src/builtins/type.rs
```diff
@@ -5,7 +5,7 @@ use super::{
 use crate::{
     builtins::{
         descriptor::{
-            DescrObject, MemberDef, MemberDescrObject, MemberGetter, MemberKind, MemberSetter,
+            DescrObject, MemberGetter, MemberKind, MemberSetter, PyMemberDef, PyMemberDescriptor,
```
fanninpm, line 8: Would it make sense to also have `PyMemberGetter`, `PyMemberKind`, and `PyMemberSetter`?

youknowone: Though I am not confident of this logic,

Those types are RustPython's internal while PyMemberDef is a part of public python C-API starting with `PyMember_`

We don't use Py prefix for our internal but use it for exposed Python types. But not sure about the exact boundary yet.

DimitrisJim (approved):
lgtm. do we have more naming refactoring we need to do? We should definitely document the naming convention someplace to have it as a point of reference.

youknowone:
Maybe we'd better to study more on how pyo3 names them.

PR #4928, opened 2023-04-26
Use `Located::new` over struct initializer

MichaReiser:
This PR changes the `python.lalrpop` to use `Located::new` over the struct initializer syntax. 

This change is motivated to ease upstream contribution from Astral's RustPython fork. Astral's fork uses `Located::end_location: Location` over `Located::end_location: Option<Location>` [[PR](https://github.com/astral-sh/RustPython/pull/4)].

Using `Located::new` has the added benefit that it requires slightly less code because the `custom` field can be omitted, and `end_location` doesn't need to be wrapped in `Some`

### DIFF
--- a/compiler/parser/python.lalrpop
+++ b/compiler/parser/python.lalrpop
@@ -67,23 +67,21 @@ SmallStatement: ast::Stmt = {
 
 PassStatement: ast::Stmt = {
     <location:@L> "pass" <end_location:@R> => {
-        ast::Stmt {
+        ast::Stmt::new(
             location,
-            end_location: Some(end_location),
-            custom: (),
-            node: ast::StmtKind::Pass,
-        }
+            end_location,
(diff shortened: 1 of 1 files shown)


youknowone (approved):
Thank you!

youknowone:
I wish we can entirely remove Option from end_location, but not sure if it is possible while keeping cpython compatibiltiy.

MichaReiser:
Thanks for merging this so quickly. 

> I wish we can entirely remove Option from end_location, but not sure if it is possible while keeping cpython compatibiltiy.

I'm unfamiliar with CPython. Do you know what's the reason in CPython for `end_location` being optional?

youknowone:
I don't remember well, but one of them was python-side created ast objects issues.
I think a workaround using `__dict__` of the python object will work.
I will look in what more issues exist.

PR #4929, opened 2023-04-26
Add `Located::start`, `Located::end` and impl `Deref`

MichaReiser:
This is a follow-up to #4192 and is motivated by the discussions in https://github.com/astral-sh/RustPython/pull/4

The PR adds the`start` and `end` methods to `Located` that return a node's start or end location. The `end` method falls back to returning `location` if `end_location` is `None`. 

The PR further implements `Deref` for `Located` so that the node's fields and methods can be accessed directly: You can write `located.field` instead of `located.node.field`

### DIFF
--- a/compiler/ast/asdl_rs.py
+++ b/compiler/ast/asdl_rs.py
@@ -671,6 +671,24 @@ def write_ast_def(mod, typeinfo, f):
             pub fn new(location: Location, end_location: Location, node: T) -> Self {
                 Self { location, end_location: Some(end_location), custom: (), node }
             }
+            
+            pub const fn start(&self) -> Location {
+                self.location
+            }
+            
+            /// Returns the node's [`end_location`](Located::end_location) or [`location`](Located::start) if 
+            /// [`end_location`](Located::end_location) is `None`.
+            pub fn end(&self) -> Location {
+                self.end_location.unwrap_or(self.location)
+            }      
+        }
+        
+        impl<T, U> std::ops::Deref for Located<T, U> {
+            type Target = T;
+            
+            fn deref(&self) -> &Self::Target {
+                &self.node
+            }
         }
         \n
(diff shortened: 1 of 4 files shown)


youknowone (approved):
Thank you!

MichaReiser:
> Thank you!

Thank you for chiming in on the PR, suggesting to upstream these changes and merging them so quickly. It's great fun contributing to RustPython thanks to you :)

DimitrisJim:
I'd slightly push back on Deref here, docs boldly state it should only be implemented on smart pointer types but we might already be doing it in other places so :smile:

MichaReiser:
> I'd slightly push back on Deref here, docs boldly state it should only be implemented on smart pointer types but we might already be doing it in other places so 😄

We can undo the change if you feel strongly. 

The term smart-pointer isn't well defined and there's ongoing discussion to re-phrase or even remove the section https://github.com/rust-lang/rust/issues/91004. 

The way I think about it is that `Located` is a thin-wrapper around `Node` that adds source locations to it. You could probably even argue that it's, to some extend, similar to an `Rc` because it stores the inner data and some metadata about it (reference count / source locations)

DimitrisJim:
I definitely do not feel that strongly!

youknowone:
Maybe it can be anti-pattern, but we already depends on a lot for Deref for thin wrappers. And it is abused a lot in many projects 😂 . Replacing them to traits will be possible but will be painful. I hope rust have a similar features allows method call for wrappers but disallow `*` operator to prevent confusion between reference and wrapper.

@MichaReiser Thank you so much for paying attention to keep unified code base. You are most welcome!

jmaargh:
FYI: Some better guidance on when and when not to `Deref` is currently in review: https://github.com/rust-lang/rust/pull/110340

Issue #4894, opened 2023-04-19
[RFC] Add `visitor` capabilities to ASTs


thejcannon:
## Summary

I'm porting an AST visitor from Python to RustPython and would love to mirror the code which is using `ast.NodeVisitor`, but in Rust.

## Detailed Explanation

A rough equivalence to being able to provide functions for certain nodes, while not worrying about other nodes or traversal.

## Drawbacks, Rationale, and Alternatives

Rolling our own AST visitor.

## Unresolved Questions

N/A

youknowone:
could you check if #4872 covers it?

youknowone:
If you could share more detailed use case, it will be a lot helpful for us to understand it better!

thejcannon:
I couldn't grok the trace PR, I'll try again later.

As for the use case, we parse Python files too collect import statements anywhere in the AST. Using a visitor means:
- we don't have to worry about recursing on nodes with statements in them
- if we upgrade to newer RustPython with newer syntax support, we can't forget to support new nodes (like `TryStar` for instance) because the base visitor handles it.

thejcannon:
Yeah I'll whip up the trait and make a PR against this repo

youknowone:
Thank you!

youknowone:
https://github.com/RustPython/Parser/pull/12

PR #4930, opened 2023-04-26
Generate a visitor trait to `ast_gen.rs`

thejcannon:
Fixes #https://github.com/RustPython/RustPython/issues/4894 by generating a `Visitor` trait hidden behind a `visitor` feature.

(The Python code isn't the best, but certainly gets the job done.)
At a high level:
- Generates structs for each node type so we can pass a located object to the methods.
- Rolls a `visit_{name}` which forwards to `generic_visit_{name}` which itself recurses for children.
- This matches `ast.NodeVisitor` in Python itself, meaning users can easily port python code to Rust


fanninpm:
I find it better to use some sort of `git rebase` command.

thejcannon:
I prefer merge, as I find it leads to less conflicts in the long run and doesn't require force-pushing (which deleted history)

fanninpm:
@youknowone Which is preferred, `git rebase` or not `git rebase`, and why?

thejcannon:
Even if `rebase` is preferred, I don't feel comfortable force-pushing to branches out in the wild.

youknowone:
I believe there must be well-explained articles somewhere, but here is my brief.

## Rebase is better
It reduces bisect cost when we have regressions. Merge itself doesn't have problem. But bidirectional merging between 2 branches makes problem. The history will look like diamond rather than a tree.
How to create unidirectional merge? There can be multiple ways, but because GitHub basically do merge by merge button, allowing merge only for upstream is natural in GitHub.
Upstream is RustPython:main here. This branch is expected to merge other branches. I means this PR is not expected to include merge commit to `upstream/main`. (Merging to other downstream if fine - but it is not this case)

### Certain merges are acceptable
Since we have squash merge. If merge to upstream is included in a patch, I don't create a merge commit. Rather than, I squash merge to avoid bidirectional merge history.

The concept is same as running:
```
git reset <the last base commit you didn't write>
git add <every changes>
git commit -m "Make a new squashed commit"
```
This is worse way to keep your history. But it is acceptable for small patches.

## Force-push is safe for downstream
The upstream history must be reserved to prevent accidentally breaking references. Users depends on `RustPython:main`. So we have to keep the history of this branch. By point of view of git command, the upstream branch and temporary-created branches of downstream are not distinguishable. So git always provide best safety option for upstream branch.
When contributors creates a branch to create a Pull Request, the branch is going to be deleted after merge. Nothing bad happens.

This is not users fault. It happens just because git lacks features to distinguish upstream and downstream. Maybe because linux kernel is multi-tier unlike most of GitHub projects. And git didn't have motivation to be designed for GitHub, of course.


I usually share this article. https://docs.gitlab.com/ee/topics/git/git_rebase.html

thejcannon:
Also, on the force-push issue. It does remove some context from the PR as links to code pre-push are now dead. I get many GitHub emails from PR commits with dead links.
It also means I can't review "what's changed since last time" easily.

thejcannon:
I should add I've started porting my [Python-based NodeVisitor](https://github.com/pantsbuild/pants/blob/main/src/python/pants/backend/python/dependency_inference/scripts/general_dependency_visitor.py) to use the new visitor from my branch and not only does it work, it's been very easy to human-transpile it :tada:

youknowone:
That's fair point. I agree merge + github squash has advantages. Maybe encouraging rebase only when merge commit is needed could be better.

youknowone (commented):
This is looking working, but it shows me rustpython-ast need refactoring.
The patch is creating new types for each `StmtKind::*`.
Is making new types out of the trait Visitor and using it for both visitor and StmtKind makes sense?

youknowone:
@thejcannon Do you use rustpython-parser? Then could you also also look in this issue? https://github.com/RustPython/RustPython/issues/4941

youknowone:
@coolreader18 I know you are recently busy, but you must know this better than me.

thejcannon:
#4941 would be ideal. Ideally I'd love to reuse my visitor when parsing python 2 as well.

thejcannon:
> This is looking working, but it shows me rustpython-ast need refactoring.
> The patch is creating new types for each `StmtKind::*`.
> Is making new types out of the trait Visitor and using it for both visitor and StmtKind makes sense?

I started down that path, but that would break the API, and didn't want to do that just for the visitors sake.

If you're ok with an API breakage, that's doable.

youknowone:
> If you're ok with an API breakage, that's doable.

We'd like to have better API in general. If you are okay, let's do it 😄 
I believe the ast crate users use pinned commit id. It will be ok.

#4941 and this PR doesn't have conflict. If we do split it, it will be just creating a repo and move source code from here.
Please feel free to fix anything regardless when it will happen

thejcannon:
Closing. See https://github.com/RustPython/RustPython/issues/4941 Will re-open in new repo/crate

PR #4943, opened 2023-04-28
Use ruff for CI

youknowone:


DimitrisJim (approved):
shame on us we weren't using ruff already :smile:

charliermarsh (approved):
Yay :)

PR #4944, opened 2023-04-28
Property name

Masorubka1:
#4067


vm/src/builtins/property.rs
```diff
@@ -125,7 +128,8 @@ impl PyProperty {
 
     #[pymethod(magic)]
     fn set_name(&self, args: PosArgs, vm: &VirtualMachine) -> PyResult<()> {
```
youknowone, line 130: Previously, we didn't use the value. But to use the value, we typically declare args with more specific types.
```suggestion
    fn set_name(&self, owner: PyObjectRef, name: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
```

Masorubka1: What have i to do with this test? tests/test_propetry.py
```python
def test_property_set_name_incorrect_args(self):
        p = property()

        for i in (0, 1, 3):
            with self.assertRaisesRegex(
                TypeError,
                fr'^__set_name__\(\) takes 2 positional arguments but {i} were given$'
            ):
                p.__set_name__(*([0] * i))
```

Maybe would be better to do macros/pre_calling_function checks amout args?

youknowone: Ah, I am sorry. That's RustPython's design caveat. We have to fix the fundamental cause to inject function name.

if you don't want to mark test fails, I will take this patch as it is.
Otherwise, this will be also working:

```rust
fn set_name(&self, args: PosArgs, vm: &VirtualMachine) -> PyResult<()> {
    let func_args = args.into_args();
    let (owner, name): (PyObjectRef, PyObjectRef) = func_args.bind(vm).unwrap_or_else(|e| <the name error>)?;
```

Masorubka1: It will be ok, if i implement std::ops::Try for PyBaseExceptionRef, because as i understand, in your example i will get in |e| PyBaseExceptionRef, and it does`t support ?. 
```rust
let (owner, name): (PyObjectRef, PyObjectRef) = func_args.bind(vm).unwrap_or_else(|e| vm.new_exception_msg(
            vm.ctx.exceptions.type_error.to_owned(),
            format!(
                "__set_name__() takes 2 positional arguments but {} were given",
                func_args.args.len()
            ),
        ))?;
```

youknowone: I am sorry, it must be `map_err`
```
func_args.bind(vm).map_err(|e| <the name error>)?;
```

Your code looks good. You will only need to replace unwrap_or_else to map_err

youknowone: You also can use `vm.new_type_error` if you want to make it simpler.

vm/src/builtins/property.rs
```diff
@@ -1,23 +1,24 @@
...
+use crate::function::{IntoFuncArgs, PosArgs};
 use crate::{
     class::PyClassImpl,
     function::{FuncArgs, PySetterValue},
     types::{Constructor, GetDescriptor, Initializer},
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyRef, PyResult, VirtualMachine,
 };
 
-#[pyclass(module = false, name = "property", traverse)]
+#[pyclass(module = false, name = "property")]
```
youknowone, line 14:
```suggestion
#[pyclass(module = false, name = "property", traverse)]
```
`traverse` need to be kept for GC

youknowone (approved):
Thank you!

PR #4946, opened 2023-04-29
Update typing from CPython 3.11.2

youknowone:
#4564

related to TypedDict

cc @lastmjs


lastmjs:
Our goal is to be able to use generics with TypedDicts, like this:

```python
class Result(TypedDict, Generic[T, K], total=False):
    Ok: T
    Err: K
```

PR #4948, opened 2023-04-29
Format d bool support

### LINKED ISSUE #4939: {:d} format should support bool value.
{:d} format is supported to pass a boolean value. See the following test case. CPython can execute it and return expected result, while RustPython report a TypeError.

```
for i in range(12):
    print('{:d}, '.format(chr(i) in "ss"), end='')
    if i % 16 == 1:
        print()
```

Behaviors on CPython:
```
0, 0, 
0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 
```

Behaviors on RustPython
```
  File "/home/xxm/Desktop/RustPython/test.py", line 169, in <module>
    print('{:d}, '.format(chr(i) in "ss"), end='')
TypeError: unsupported format string passed to bool.__format__
```
Ubuntu 18.04
RustPython 0.2.0 [0284059](https://github.com/RustPython/RustPython/commit/02840593bc56ae416ba2646166628f1712d6cf43)
CPython 3.8.0, 3.9.0, 3.11.3

JazzGlobal:
Closes #4939

Probably more to do but im gonna let the action runner help me.

### DIFF
--- a/common/src/format.rs
+++ b/common/src/format.rs
@@ -423,7 +423,7 @@ impl FormatSpec {
         };
         let result: Result<String, FormatSpecError> = match &self.format_type {
             Some(FormatType::Decimal) => Ok(x.to_string()),
-            default => Err(FormatSpecError::InvalidFormatSpecifier),
+            _ => Err(FormatSpecError::InvalidFormatSpecifier),
         };
         result
     }
--- a/common/src/format.rs
+++ b/common/src/format.rs
@@ -417,12 +417,14 @@ impl FormatSpec {
     }
 
     pub fn format_bool(&self, input: bool) -> Result<String, FormatSpecError> {
-        let x = match input {
-            true => 1,
-            false => 0,
-        };
+        dbg!(&self.format_type);
+        let x = u8::from(input);
         let result: Result<String, FormatSpecError> = match &self.format_type {
             Some(FormatType::Decimal) => Ok(x.to_string()),
+            None => {
+                let first_letter = (input.to_string().as_bytes()[0] as char).to_uppercase();
+                Ok(first_letter.collect::<String>() + &input.to_string()[1..])
+            }
             _ => Err(FormatSpecError::InvalidFormatSpecifier),
         };
         result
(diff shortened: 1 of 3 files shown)


test.py
fanninpm, line 1: There might be a test in the CPython test suite that checks for this.

common/src/format.rs
```diff
@@ -416,6 +416,18 @@ impl FormatSpec {
         }
     }
 
+    pub fn format_bool(&self, input: bool) -> Result<String, FormatSpecError> {
+        let x = match input {
+            true => 1,
+            false => 0,
+        };
```
fanninpm, line 423: You could probably do something simpler like this:

```suggestion
        let x = u8::from(input);
```

(Just make sure Clippy is okay with it.)

vm/src/builtins/bool.rs
```diff
@@ -110,12 +111,11 @@ impl Constructor for PyBool {
...
 impl PyBool {
     #[pymethod(magic)]
-    fn format(obj: PyObjectRef, format_spec: PyStrRef, vm: &VirtualMachine) -> PyResult<PyStrRef> {
-        if format_spec.is_empty() {
-            obj.str(vm)
-        } else {
-            Err(vm.new_type_error("unsupported format string passed to bool.__format__".to_owned()))
-        }
+    fn format(obj: PyObjectRef, spec: PyStrRef, vm: &VirtualMachine) -> PyResult<String> {
+        let new_bool = obj.try_to_bool(vm).unwrap();
```
fanninpm, line 115: Does this work?
```suggestion
        let new_bool = obj.try_to_bool(vm)?;
```

JazzGlobal: Yes, still not totally used to all the Rust shorthands :D

common/src/format.rs
JazzGlobal: Looks good, clippy didnt complain locally at least.

test.py
JazzGlobal: Speaking of tests, looks like I broke a handful. Executing `f'{1 != 2}'` throws an error.

youknowone (approved):
Thank you for contributing!

Issue #4941, opened 2023-04-28
[RFC] Splitting parser out of RustPython repository


youknowone:
## Summary

Recently, rustpython-parser has definitely a lot more users than RustPython itself.
I suggest splitting parser out from RustPython repository.

## Detailed Explanation

When upstream is comparably smaller than downstream, downstream goes to make fragmentation rather than contributing to upstream.

The problem here is the user expectation and the project structure has opposite ratio.

In the project RustPython, parser is comparably small module - mostly due to huge stdlib of Python.
For rustpython-parser users, only rustpyhon-parser is useful but other parts is useless burden.

I'd like to suggest splitting parser out from RustPython repository.
it will be one of the good way to keep rust-implemented python parser in single repository with minimum fragmentation.

## Drawbacks, Rationale, and Alternatives

RustPython will use the separated parser with git tag. It will increase cost of maintenance of other parts of RustPython.

## Unresolved Questions

Not to restrict the parser's potent to just RustPython, the name even can be something different. `python-parser` or `python3-parser`.

thejcannon:
I'd put in a slight request to make the crate also support Python 2. Bonus points if it can map onto the same `Visitor` trait.

We have a use case for parsing Python 2 and Python 3.5+ scraping the same information from either, so one crate handling both languages in an identical (maybe even the exact same) way would be amazing.

youknowone:
@thejcannon Thank you for the comment! Your demand seems showing better why we need a new repository. Aggregating more different use case to it will make the module more organized.

thejcannon:
OK, I'd love to contribute to this, but I don't think I'm the best person to own it.
So how about:
- The new repo is added to this org: RustPython
- It is a strict port of the code inside RustPython. That way RustPython can pin the dependency to the commit containing the port
- I'll make a PR in that repo which breaks API to generate the per-kind structs and update the `FooKind` enums to use those.
- Then make a follow-up PR to generate the visitor (non API-breaking)
- (At some point in the future add Py2 capabilities)

thejcannon:
Oh and for naming, I suggest `python-parser` so we don't close ourselves out of Python2

youknowone:
I think it will be started with a hard-copy or git-subtree.
It will be created under https://github.com/RustPython/parser
It will include roughly 4 crates - rustpython-common, rustpython-parser, rustpython-ast and rustpython-compiler-core not to create cross-dependency between crates.
[...]

PR #4949, opened 2023-04-30
Replace row/column based Location with byte-offsets.

youknowone:
Try https://github.com/astral-sh/RustPython/pull/4


youknowone:
Many utilities are required to be imported from ruff.  #4941 will help the work to be done easier.

youknowone:
done by #4961

PR #4951, opened 2023-04-30
Port configurable repr quote from Ruff and refactoring

youknowone:
Logic was ported, but kept the default `repr` behaviors to use single quote.
I think the term `repr` - not representation or debug - means `repr()` in Python.
Instead of it, I added other configurable methods for each str and bytes.

Rather than the name repr, I found `unicode_escape` codec is doing the same thing.
I will try to organize them more under 'escape' later.

I realized str and bytes are doing almost same thing but have so much different interface.

---

Unrelated note: I am thinking the new crate name for these stuffs `rustpython-literal` including str/bytes/float repr


common/src/str.rs
```diff
@@ -345,10 +351,7 @@ macro_rules! ascii {
 /// Get a Display-able type that formats to the python `repr()` of the string value
 #[inline]
 pub fn repr(s: &str) -> Repr<'_> {
```
charliermarsh, line 353: Should there be a `repr_with_quote` here to mirror the bytes API? Or is the intent that we construct a `Repr` directly?

youknowone: The latter, but the current state seems ugly. Is it better to make advance now?

The function `repr` seems not doing much thing. Probably entirely removing it will be better. I now think doing this right now will be better.

In future, After renaming `Repr` to `Escape`, `Escape::new(quote, prefix, suffix)`, `Escape::with_single()`, `Escape::with_quote()` etc seems better way than having many functions. Could you share your expectation?

charliermarsh: Yeah I think something like what you're describing sounds reasonable. But I'm fine with anything in the interim, as long as the functionality is exposed in some sort of public interface.

common/src/escape.rs
```diff
@@ -0,0 +1,246 @@
...
+pub struct UnicodeEscape<'a> {
+    source: &'a str,
+    layout: EscapeLayout,
+}
+
+impl<'a> UnicodeEscape<'a> {
+    pub fn with_forced_quote(source: &'a str, quote: Quote) -> Self {
+        let layout = EscapeLayout { quote, len: None };
+        Self { source, layout }
+    }
```
youknowone, line 112: Python requires length calculation due to allocating string object, but formatters doesn't.

will this be helpful for formatter?

```rust
let escape = if source.len() < TOO_LONG_TO_FORMAT {
    UnicodeEscape::output_layout(source, Quote::Double)
} else {
    UnicodeEscape::with_forced_quote(source, Quote::Double)
};

```

PR #4952, opened 2023-05-02
Replace num-bigint with malachite-bigint

qingshi163:


Cargo.toml
```diff
@@ -36,10 +36,11 @@ itertools = "0.10.3"
 libc = "0.2.133"
 log = "0.4.16"
 nix = "0.26"
+malachite-bigint = { version = "0.1.0", git = "https://github.com/qingshi163/malachite-bigint.git" }
```
youknowone, line 39: 👍 you got it 😄

DimitrisJim: note that we can totally keep this under the RustPython github org to ease with maintainability. whatever you'd prefer.

qingshi163: I don't mind to move it under RustPython org.

DimitrisJim: totally, guess you can move it whenever you want. might be easier to finish up with it first in your local repo and then move.

youknowone: Either way will be ok up to @qingshi163's preference as the sole author.
Unlike sre-engine - though I don't know why this is out of RustPython repository, it is not necessarily a part of RustPython. But it will be good idea if you want to share maintenance burden.

qingshi163: repr moved under RustPython org.

youknowone:
`BigUint::parse_bytes` seems rejecting `_` in input, which is legal on num-bigint.

qingshi163:
Don't change bytes_to_int, malachite-bigint should handle it. I think it is a bug involved on the last commit

youknowone:
@qingshi163 Can we run a benchmark about this? A benchmark with many small int binary operations will fit.

corona10:
@youknowone @qingshi163 
FYI for the Fibonacci benchmark
````

Mean +- std dev: [base] 60.0 us +- 2.0 us -> [pr_4952] 55.4 us +- 1.1 us: 1.08x faster
```` 

`````python

import pyperf

runner = pyperf.Runner()
runner.timeit(name="bench fib",
              stmt="""
_ = fib(10)
""",
              setup = """
def fib(n):
    if n == 0:
        return 0
    if n == 1 or n == 2:
        return 1
    return fib(n-1) + fib(n-2)
""")

`````

youknowone:
hmm, this is unexpected. what's changed?

youknowone:
I can't locally reproduce it.

youknowone:
ah, main is already broken

Issue #4938, opened 2023-04-28
time.mktime((88,)*9) reports a ValueError


xiaxinmeng:
time.mktime() takes (88,)*9 as arguments causes a ValueError on RustPython. It can work well on CPython.

test.py
```
>>> import time
>>> time.mktime((88,)*9)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
OverflowError: mktime argument out of range
```

Behaviors on CPython:
work normally

Behaviors on RustPython:
```
Traceback (most recent call last):
  File "/home/xxm/Desktop/RustPython/test.py", line 381, in <module>
    time.mktime((88,)*9)
ValueError: invalid struct_time parameter
```

Ubuntu 18.04
RustPython 0.2.0 [0284059](https://github.com/RustPython/RustPython/commit/02840593bc56ae416ba2646166628f1712d6cf43)
CPython 3.8.0, 3.9.0, 3.11.3

DimitrisJim:
> Behaviors on CPython:
work normally

this is an overflow in CPython and should similarly be one in RustPython. The change should be relatively straight-forward: when trying to convert the arguments into a `NaiveDateTime` we hit the `ok_or_else` handling due to the values not being in a valid range:

https://github.com/RustPython/RustPython/blob/fa790558211ec690541299258f44d7453501958a/vm/src/stdlib/time.rs#L359-L364

where invalid is defined a couple of lines before:

https://github.com/RustPython/RustPython/blob/fa790558211ec690541299258f44d7453501958a/vm/src/stdlib/time.rs#L353

Invalid returning an `OverflowError` in this case with a message of `"mktime argument out of range"` would do the trick.

youknowone:
Half fixed by @bjoroen in #4959

By looking cpython, mktime always seems to raise OverflowError.

By fully changing it to OverflowError, one other test fails which expect ValueError
https://github.com/RustPython/RustPython/pull/4959/files#diff-d6e06231082cee70ab1dbfd3845ae4f634b70f79c0f2b25b5f08cbfa0e90d73bR355

i didn't check why it happens, but the failing feature was not mktime and maybe the function is specifically requiring ValueError.

bjoroen:
I think I've found the reason for the error.

It comes from the difference in behavior between Cpython and the rust crate Chrono which is used to create the Date.

[...]

PR #4959, opened 2023-05-07
Report overflow error

bjoroen:
#4938

It now creates an overflow error with the message described in the issue. 

Output: 
```
Welcome to the magnificent Rust Python 0.2.0 interpreter
>>>>> import time
>>>>> time.mktime((88,)*9)
OverflowError: mktime argument out of range

```

### DIFF
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -350,7 +350,7 @@ mod time {
         }
 
         fn to_date_time(&self, vm: &VirtualMachine) -> PyResult<NaiveDateTime> {
-            let invalid = || vm.new_value_error("invalid struct_time parameter".to_owned());
+            let invalid = || vm.new_overflow_error("invalid struct_time parameter".to_owned());
             macro_rules! field {
                 ($field:ident) => {
                     self.$field.clone().try_into_value(vm)?
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -350,7 +350,7 @@ mod time {
         }
 
         fn to_date_time(&self, vm: &VirtualMachine) -> PyResult<NaiveDateTime> {
-            let invalid = || vm.new_overflow_error("invalid struct_time parameter".to_owned());
+            let invalid = || vm.new_overflow_error("mktime argument out of range".to_owned());
             macro_rules! field {
                 ($field:ident) => {
                     self.$field.clone().try_into_value(vm)?
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -350,7 +350,10 @@ mod time {
         }
 
         fn to_date_time(&self, vm: &VirtualMachine) -> PyResult<NaiveDateTime> {
-            let invalid = || vm.new_overflow_error("mktime argument out of range".to_owned());
+            let invalid_overflow =
+                || vm.new_overflow_error("mktime argument out of range".to_owned());
+            let invalid_value = || vm.new_value_error("invalid struct_time parameter".to_owned());
+
             macro_rules! field {
                 ($field:ident) => {
                     self.$field.clone().try_into_value(vm)?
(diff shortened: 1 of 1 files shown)


youknowone (approved):

youknowone (changes requested):
By looking the test result, that function looks to have both types of errors.
Could you distinguish those errors?

`year=10000` seems to expect ValueError
```
test test_time crashed -- Traceback (most recent call last):
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/libregrtest/runtest.py", line 256, in _runtest_inner
    refleak = _runtest_inner2(ns, test_name)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/libregrtest/runtest.py", line 207, in _runtest_inner2
    the_module = importlib.import_module(abstest)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "_frozen_importlib", line 1207, in _gcd_import
  File "_frozen_importlib", line 1176, in _find_and_load
  File "_frozen_importlib", line 1179, in _find_and_load
  File "_frozen_importlib", line 1153, in _find_and_load_unlocked
  File "_frozen_importlib", line 1150, in _find_and_load_unlocked
  File "_frozen_importlib", line 1150, in _find_and_load_unlocked
  File "_frozen_importlib", line 706, in _load_unlocked
  File "_frozen_importlib", line 704, in _load_unlocked
  File "_frozen_importlib", line 697, in _load_unlocked
  File "_frozen_importlib", line 691, in _load_unlocked
  File "_frozen_importlib_external", line 940, in exec_module
  File "_frozen_importlib", line 241, in _call_with_frames_removed
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/test_time.py", line 674, in <module>
    del skip_if_not_supported
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/test_time.py", line 666, in _TestStrftimeYear
    @skip_if_not_supported(10000)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/test_time.py", line 661, in skip_if_not_supported
    cond = False
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/test_time.py", line 659, in skip_if_not_supported
    time.strftime('%Y', (y,) + (0,) * 8)
OverflowError: mktime argument out of range
```

bjoroen:
I sure can.
But I'm unsure what needs to be done,  should I return both types of errors? or only return Value error if the year is 10000? 
Some pointers in the right direction would be greatly appreciated.

bjoroen:
I pushed new changes, is that what you had in mind? The test that failed now is green. Let me know if you meant something different and I can fix it.

bjoroen:
No problem. 
Thank you for the review, I appreciate it.

bjoroen:
I would love to contribute more, so ill look for other things I can do to help.

vm/src/stdlib/time.rs
```diff
@@ -350,17 +350,20 @@ mod time {
...
+
             macro_rules! field {
                 ($field:ident) => {
                     self.$field.clone().try_into_value(vm)?
                 };
             }
             let dt = NaiveDateTime::new(
                 NaiveDate::from_ymd_opt(field!(tm_year), field!(tm_mon), field!(tm_mday))
-                    .ok_or_else(invalid)?,
+                    .ok_or_else(invalid_value)?,
```
DimitrisJim, line 364: I think this should raise an `Overflow` error too. The documentation for CPython states:

> If the input value cannot be represented as a valid time, either [OverflowError](https://docs.python.org/3/library/exceptions.html#OverflowError) or [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError) will be raised (which depends on whether the invalid value is caught by Python or the underlying C libraries).

The value being caught by Python probably means the conversion from a CPython object into a CPython longobject. I.e, if you pass in a list, this will obviously fail with a `ValueError` because it will be expecting an integer.

The underlying C library is basically equivalent to the `NaiveDate` calls here. I'll try and look into it further on the weekend to see where the value error YunWon mentions happens.

DimitrisJim: That sounds great to me! Do reply in the issue so we can keep as much of the context in one place.

Issue #4906, opened 2023-04-21
RustPython becomes silent if raising an error in trace() function


xiaxinmeng:
RustPython becomes silent if an error is raised in trace() function. The following example can normally print things on CPython while it fails on RustPython. 

```
import sys
def trace(frame, event, arg):
    raise NameError
    return trace

sys.settrace(trace)
print("fail here")
try:
    a = 1/0
except:
    print("not allowed")
```

Output on RustPython --- silent

Output on CPython (Expected):
fail here
not allowed

Environment 
Ubuntu 18.04
RustPython 0.2.0
CPython 3.11.3

DimitrisJim:
I *think* this should be an easy fix. The [docs](https://docs.python.org/3/library/sys.html#sys.settrace) for `settrace` state:

> If there is any error occurred in the trace function, it will be unset, just like settrace(None) is called.

looking at the way CPython currently behaves, it seems like the Exception is ignored, the trace function is set to nothing and execution continues as it would. In our case, every time a call is made (not only printing, _any_ call) the `trace` function is called:

https://github.com/RustPython/RustPython/blob/fa790558211ec690541299258f44d7453501958a/vm/src/protocol/callable.rs#L48-L54

and this will eventually call the trace function which will return an `Err` and end execution early:

https://github.com/RustPython/RustPython/blob/fa790558211ec690541299258f44d7453501958a/vm/src/protocol/callable.rs#L101-L106

so, the `res` over here shouldn't be propagated, instead, it should be matched on and if it is an `Err` the trace function should be set to `None`. Similarly for the `profile_func` (I'd assume, I haven't checked the docs).

matthew-hagemann:
I see in the docs for profile it does mention:
> Error in the profile function will cause itself unset.

I'll add in the match and create a PR

PR #4963, opened 2023-05-10
Set trace_func / profile_func to none on error

matthew-hagemann:
#4906

trace_func and profile_func now set to none if errors are encountered

Working with the example code:
```
import sys
def trace(frame, event, arg):
    raise NameError
    return trace

sys.settrace(trace)
print("fail here")
try:
    a = 1/0
except:
    print("not allowed")
```
RustPython now correctly will print:
fail here
not allowed

### DIFF
--- a/Lib/test/test_trace.py
+++ b/Lib/test/test_trace.py
@@ -298,8 +298,6 @@ class TestFuncs(unittest.TestCase):
         }
         self.assertEqual(self.tracer.results().calledfuncs, expected)
 
-    # TODO: RUSTPYTHON, gc
-    @unittest.expectedFailure
     def test_arg_errors(self):
         res = self.tracer.runfunc(traced_capturer, 1, 2, self=3, func=4)
         self.assertEqual(res, ((1, 2), {'self': 3, 'func': 4}))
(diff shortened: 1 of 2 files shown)


DimitrisJim:
thanks! there's a couple of tests that appear to now succeed:

```bash
UNEXPECTED SUCCESS: test_run_as_module (test.test_trace.TestCommandLine.test_run_as_module)
UNEXPECTED SUCCESS: test_sys_argv_list (test.test_trace.TestCommandLine.test_sys_argv_list)
UNEXPECTED SUCCESS: test_arg_errors (test.test_trace.TestFuncs.test_arg_errors)
```

if you haven't done it previously. You can open the relevant file (`Lib/test/test_trace.py`) and remove the decorator and that currently states that the test is expected to fail. For example for `test_arg_errors` the change should be:

```python
# TODO: RUSTPYTHON, gc
@unittest.expectedFailure
def test_arg_errors(self):
    # Function body
```

into:

```python
def test_arg_errors(self):
    # Function body
```

matthew-hagemann:
Thanks for the help, I have updated those tests 👍

fanninpm:
@xiaxinmeng Does this fix the issue you were having?

youknowone (approved):
Thank you so much!

PR #4965, opened 2023-05-11
Allow uuid on wasm32 and wasi

youknowone:
cc @lastmjs

 https://github.com/python/cpython/pull/104397


lastmjs:
I wonder if `_signal` will be a dependency of other modules though, since `signal` depends on `_signal`. This just came up when including `uuid`.

lastmjs:
It worked for uuid though!

youknowone:
It will be eventually need to be implemented.
`platform` unnecessarily requires `subprocess`. and `subprocess` requires `signal`.
If you find any problem with platform, It can be also quickly patched.
signal module will take time

Issue #4971, opened 2023-05-14
Add fork support functions


itsankitkp:
os.fork needs certain support functions to work properly, namely `_at_fork_reinit`
Basic idea is that whenever child is forked from parent, it shouldn't own locks held by its parent (parent's problem is not children's problem right?).
So whenever child gets initialized, lock object should get reallocated. 
cpython code 
```
int
_PyThread_at_fork_reinit(PyThread_type_lock *lock)
{
    PyThread_type_lock new_lock = PyThread_allocate_lock();
    if (new_lock == NULL) {
        return -1;
    }

    /* bpo-6721, bpo-40089: The old lock can be in an inconsistent state.
       fork() can be called in the middle of an operation on the lock done by
       another thread. So don't call PyThread_free_lock(*lock).

       Leak memory on purpose. Don't release the memory either since the
       address of a mutex is relevant. Putting two mutexes at the same address
       can lead to problems. */

    *lock = new_lock;
    return 0;
}
```
Notice, it doesn't bother cleanup up old lock object.

cpython's [Add _at_fork_reinit() method to locks](https://bugs.python.org/issue40089) gives relevant details

Unknowns: LockType object is alias for  Lock, but there is file called `Lib/_dummy_thread.py` in rustpython which I don't see in cpython.

Implementation strategy:
1. Add _at_fork_reinit in Lock and RLock in vm/src/stdlib/thread.rs
2. Realloc self.mu.
For this, I can not possibly do this
```
        #[pymethod]
        fn _at_fork_reinit(&mut self, vm: &VirtualMachine) -> PyResult<()> {
            self.mu = RawMutex::INIT;

            Ok(())
        }

```
It complains about
```
error[E0277]: the trait bound `for<'a, 'b> fn(&'a mut RLock, &'b vm::VirtualMachine) -> Result<(), object::core::PyRef<exceptions::types::PyBaseException>> {RLock::_at_fork_reinit}: PyNativeFnInternal<_, _, _>` is not satisfied
   --> vm/src/stdlib/thread.rs:186:5
    |
186 |     #[pyclass(with(Representable))]
[...]

PR #4972, opened 2023-05-14
Add fork support functions

itsankitkp:
Refer: https://github.com/RustPython/RustPython/issues/4971

### DIFF
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -3071,12 +3071,12 @@ class PidTests(unittest.TestCase):
         self.assertEqual(pid2, pid)
 
     # TODO: RUSTPYTHON (AttributeError: module 'os' has no attribute 'spawnv')
-    @unittest.skipUnless(hasattr(threading.Lock(), '_at_fork_reinit'), 'TODO: RUSTPYTHON, test needs lock._at_fork_reinit')
+    @unittest.skipUnless(hasattr(threading.Lock(), '_at_fork_reinit'))
     def test_waitpid(self):
         self.check_waitpid(code='pass', exitcode=0)
 
     # TODO: RUSTPYTHON (AttributeError: module 'os' has no attribute 'spawnv')
-    @unittest.skipUnless(hasattr(threading.Lock(), '_at_fork_reinit'), 'TODO: RUSTPYTHON, test needs lock._at_fork_reinit')
+    @unittest.skipUnless(hasattr(threading.Lock(), '_at_fork_reinit'))
     def test_waitstatus_to_exitcode(self):
         exitcode = 23
         code = f'import sys; sys.exit({exitcode})'
@@ -3108,7 +3108,7 @@ class PidTests(unittest.TestCase):
         with self.assertRaises(OverflowError):
             os.waitstatus_to_exitcode(-1)
             
-    @unittest.skipUnless(hasattr(threading.Lock(), '_at_fork_reinit'), 'TODO: RUSTPYTHON, test needs lock._at_fork_reinit')
+    @unittest.skipUnless(hasattr(threading.Lock(), '_at_fork_reinit'))
     # TODO: RUSTPYTHON (AttributeError: module 'os' has no attribute 'spawnv')
     # Skip the test on Windows
     @unittest.skipUnless(hasattr(signal, 'SIGKILL'), 'need signal.SIGKILL')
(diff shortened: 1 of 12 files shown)


itsankitkp:
```
s it about _PyThread_at_fork_reinit or lock__at_fork_reinit?
_PyThread_at_fork_reinit seems like internal API. (not exposed to Python API)
But lock__at_fork_reinit does.
```
Yes, it is about it. I am getting used to making distinction between python api and internal api.

```
I am doubtful if Rust type system regards fork. This will be limited by system API of platforms.
```
Yes, I agree to this. I will check this  if forked child has parent's locks or not. If child process doesn't has it, shouldn't this function be dummy then?

```
My suggestion is casting &self.mu from &RawMutex to &AtomicCell<RawMutex> using unsafe. And replace mutex under atomic cell context.
It will be as safe as the CPython one.
```
Done this change although doing it inside unsafe throws warnings that unsafe is un-needed. Should I suppress this?

@youknowone

vm/src/stdlib/thread.rs
```diff
@@ -137,6 +138,26 @@ pub(crate) mod _thread {
...
+                };
+            }
+            // Casting to AtomicCell is as unsafe as CPython code.
+            // Using AtomicCell will prevent compiler optimizer move it to somewhere later unsafe place.
+            // It will be not under the cell anymore after init call.
+
+            let new_mut = RawMutex::INIT;
+            unsafe {
+                let old_mutex: AtomicCell<&RawMutex> = AtomicCell::new(&self.mu);
+                old_mutex.swap(&new_mut);
```
youknowone, line 155: Isn't this swap only the reference which `old_mutex` holds?

Is this working?
```suggestion
                let old_mutex: &AtomicCell<RawMutex> = std::mem::transmute(&self.mu);
                old_mutex.swap(new_mut);
```

itsankitkp: This complies without any issue so looks like its working.
Just to reconfirm, 
i.e. if mu is mutex and mu_ptr is reference to mu
then in original code:
old_mu_ptr = copy(mu_ptr)
old_mu_ptr = new_mut

after above changes:
old_mu_ptr = mu_ptr
old_mu_ptr = new_mut

Am I getting it right?

Lib/test/test_os.py
```diff
@@ -3072,13 +3072,11 @@ def check_waitpid(self, code, exitcode, callback=None):
 
     # TODO: RUSTPYTHON (AttributeError: module 'os' has no attribute 'spawnv')
```
fanninpm: Is this still true?

itsankitkp: Well no, this comment is not valid anymore. Cleaned up (although is might not be correct PR for this but is on similar context).

itsankitkp:
Couple of things on test cases:
1. Linux tests are green because all these fork related functions are based on Unix but,
2. Mac is showing failures in fcntl.
3. UNEXPECTED SUCCESS in windows for test_httpservers test cases (which fails elsewhere).

Now current change (including fork and related functions) is based on Unix system so I didn't expect anything to change for windows but it did because some tests which were skipped earlier, are now being run (as fork re-init is not platform specific even though fork is). Maybe we need to add cfg[unix] in fork-reinit? Or let it be and look for fix in now visible functions (which couldn't work before due to missing dependency) ?

vm/src/stdlib/thread.rs
youknowone: `AtomicCell::swap` replaces `T`. `T` is `&RawMutex` here. `self.mu` is `RawMutex`. It can't be changed.

How do you know this is working?

youknowone:
Though I didn't check the actual reason of window failure yet, in general, if it was `AttributeError` and never actually used, it can be fixed by adding it.

vm/src/stdlib/thread.rs
itsankitkp: > How do you know this is working?

I tried something like this
```python
import os, time
import threading
from threading import Lock

l = Lock()
l.acquire()

def fork_in_thread():
    print("In fork in thread")

    pid = os.fork()
    if pid:
        # parent
        time.sleep(5)
        os._exit(0)

    # child process
    l = Lock()
    print("Child acquiring lock") 
    l.acquire() # lock from main thread shouldn't block this
    print(f'Child: lock acquired: {l.locked()}') 


t= threading.Thread(target=fork_in_thread, args=() )
t.run()
# block the main thread: fork_in_thread() exit the process
time.sleep(10)
```
This is based on https://bugs.python.org/issue6721
This gave results in cpython and current branch of rustpython
```
In fork in thread
Child acquiring lock
Child: lock acquired: True
```

For main branch of rustpython, it gave this:
```
In fork in thread
Exception ignored in: <function _after_fork at 0x5630acf32170>
  File "/home/ankit/tt.py", line 25, in <module>
    t.run()
  File "/home/ankit/RustPython/pylib/Lib/threading.py", line 953, in run
    self._target(*self._args, **self._kwargs)
  File "/home/ankit/tt.py", line 11, in fork_in_thread
    pid = os.fork()
AttributeError: 'lock' object has no attribute '_at_fork_reinit'
Child acquiring lock
Child: lock acquired: True

```
It actually work fine in main branch sans missing function definition `_at_fork_reinit` which doesn't adds anything.

itsankitkp:
> Though I didn't check the actual reason of window failure yet, in general, if it was `AttributeError` and never actually used, it can be fixed by adding it.

Issue is that test cases in `test_httpservers` is actually working on windows (who has no business in these fork functions)

youknowone (approved):
Thank you so much! I thought there were very tricky problems, but you made it!

PR #4700, opened 2023-03-16
Activate cspell lint

youknowone:

PR #4980, opened 2023-05-19
Update DEVELOPMENT.md

DimitrisJim:
add a note on the fact we're using cspell, we can then allow it during ci ref: https://github.com/RustPython/RustPython/pull/4700


youknowone (approved):
👍

PR #4985, opened 2023-05-22
Update abc.py from Python 3.11

patrickzbhe:
Was poking around abc and noticed it wasn't updated!

abc.py [from cpython](https://github.com/python/cpython/blob/3.11/Lib/abc.py)
test_abc.py [from cpython](https://github.com/python/cpython/blob/3.11/Lib/test/test_abc.py)

#4564


youknowone (approved):
Thank you!

PR #4987, opened 2023-05-23
Fix abc error message

patrickzbhe:
Print the names of abc and the abstract methods, instead of a generic message when trying to instantiate abc. 

Allows remaining test_abc.py tests to pass.

Also noticed a snippet in extra_tests that I think was just forgotten.

### DIFF
--- a/vm/src/builtins/object.rs
+++ b/vm/src/builtins/object.rs
@@ -6,6 +6,7 @@ use crate::{
     types::{Constructor, PyComparisonOp},
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyResult, VirtualMachine,
 };
+use itertools::Itertools;
 
 /// object()
 /// --
@@ -39,16 +40,15 @@ impl Constructor for PyBaseObject {
         if let Some(abs_methods) = cls.get_attr(identifier!(vm, __abstractmethods__)) {
             if let Some(unimplemented_abstract_method_count) = abs_methods.length_opt(vm) {
                 let methods: Vec<PyStrRef> = abs_methods.try_to_value(vm)?;
-                let methods: Vec<String> = methods
-                    .into_iter()
-                    .map(|name| name.as_str().to_owned())
-                    .collect();
-                let methods = methods.join(", ");
+                let methods: String =
+                    Itertools::intersperse(methods.iter().map(|name| name.as_str()), ", ")
+                        .collect();
 
                 let unimplemented_abstract_method_count = unimplemented_abstract_method_count?;
                 let name = cls.name().to_string();
 
                 match unimplemented_abstract_method_count {
+                    0 => {}
                     1 => {
                         return Err(vm.new_type_error(format!(
                             "Can't instantiate abstract class {} with abstract method {}",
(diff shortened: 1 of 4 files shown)


vm/src/builtins/object.rs
```diff
@@ -38,10 +38,26 @@ impl Constructor for PyBaseObject {
...
-                if unimplemented_abstract_method_count? > 0 {
-                    return Err(
-                        vm.new_type_error("You must implement the abstract methods".to_owned())
-                    );
+                let methods: Vec<PyStrRef> = abs_methods.try_to_value(vm)?;
+                let methods: Vec<String> = methods
+                    .into_iter()
+                    .map(|name| name.as_str().to_owned())
+                    .collect();
+                let methods = methods.join(", ");
```
patrickzbhe, line 46: Originally I avoided allocating a vec by turning this into a tuple, then into iter, then joining with ", " using IterTools. 

```rust
let methods = abs_methods
.to_sequence(vm)
.tuple(vm)
.unwrap()
.into_iter()
.map(|s| {
    s.str(vm)
        .unwrap()
        .to_string()
})
.join(", ");
```

Afterwards, I saw in another part of the code base it was written this way. Is the vec one more preferrable?

vm/src/builtins/object.rs
```diff
@@ -38,10 +38,30 @@ impl Constructor for PyBaseObject {
...
+                            name, methods
+                        )));
+                    }
+                    2.. => {
+                        return Err(vm.new_type_error(format!(
+                            "Can't instantiate abstract class {} with abstract methods {}",
+                            name, methods
+                        )));
+                    }
+                    _ => {}
```
youknowone, line 64: Is this not able to enter or safe case?
If this is safe, I prefer to use 0

```suggestion
                    0 => {}  // fully implemented
```
Otherwise unreachable! will be good

```suggestion
                    _ => unreachable!("unimplemented_abstract_method_count is always positive")
```

youknowone (changes requested):
Thank you so much! Looks good. I left a few nits

vm/src/builtins/object.rs
```diff
@@ -38,10 +38,30 @@ impl Constructor for PyBaseObject {
...
-                if unimplemented_abstract_method_count? > 0 {
-                    return Err(
-                        vm.new_type_error("You must implement the abstract methods".to_owned())
-                    );
+                let methods: Vec<PyStrRef> = abs_methods.try_to_value(vm)?;
+                let methods: Vec<String> = methods
+                    .into_iter()
+                    .map(|name| name.as_str().to_owned())
+                    .collect();
+                let methods = methods.join(", ");
```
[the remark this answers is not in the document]
patrickzbhe: Sounds good! I used Itertools::intersperse since rust was complaining about .intersperse.

vm/src/builtins/object.rs
patrickzbhe: Yeah couldn't just do 0,1,2.. , I think its not supported yet. Added in unreachable!.

youknowone (approved):
Thank you!

Issue #4988, opened 2023-05-23
RUSTPYTHONPATH is set in WINDOWS to the correct path, but there's still an error about missing encodings.


euanmacinnes:
## Summary

I've manually set the RUSTPYTHONPATH in the system environment variable, Rust sees the correct folder location prior to calling the 
vm::Interpreter::without_stdlib(Default::default()).enter(|vm| { ..

and yet, the following error appears:

encodings initialization failed. Only utf-8 encoding will be supported.
ModuleNotFoundError: No module named 'encodings'

The above exception was the direct cause of the following exception:

RuntimeError: Could not import encodings. Is your RUSTPYTHONPATH set? If you don't have access to a consistent external environment (e.g. if you're embedding rustpython in another application), try enabling the freeze-stdlib feature

## Expected

No message

## Actual

encodings initialization failed. Only utf-8 encoding will be supported.
ModuleNotFoundError: No module named 'encodings'

The above exception was the direct cause of the following exception:

RuntimeError: Could not import encodings. Is your RUSTPYTHONPATH set? If you don't have access to a consistent external environment (e.g. if you're embedding rustpython in another application), try enabling the freeze-stdlib feature

fanninpm:
Does the problem persist after running `cargo clean`?

fanninpm:
You probably need to depend on the `rustpython-vm` crate.

euanmacinnes:
Aren't the crates yanked?

fanninpm:
Try this:

```toml
rustpython-vm = { git = "https://github.com/RustPython/RustPython", features = ['jit'] }
```

fanninpm:
@youknowone Does this have anything to do with the parser migration?

euanmacinnes:
The -vm is in the same git repo, so it didn't change anything adding it as
a dependency. Cargo didn't even compile anythimg differently.

fanninpm:
One last troubleshooting step: does the problem persist after deleting `Cargo.lock`?

youknowone:
@euanmacinnes Are you using rustpython as dependency of your project? or directly using RustPython repository?

youknowone:
> The -vm is in the same git repo, 

I think using RustPython repository by that part. Then Cargo.lock will not help.

euanmacinnes:
It's a dependency, as I'm looking to embed dynamically generated python
functions into Rust.

youknowone:
Oh, wow. It was totally broken for `Setting::default()`.
[...]

PR #4994, opened 2023-06-07
Added constructor for Settings called with_path

LucaSforza:
Part of the solution for #4988


youknowone:
Thank you for contributing!
The error is in `fn import_encodings` in `vm/src/vm/mod.rs`

DimitrisJim (approved):
lgtm

Issue #4995, opened 2023-06-08
Fix broken mac build


youknowone:
First broken build: https://github.com/RustPython/RustPython/commit/7a3f965396b725e7fa16c7a10cd06b239f92f167

DimitrisJim:
do we set `RUST_BACKTRACE` in the tests? Might be useful since locally you can't repro so the logs are the only thing we're going with.

youknowone:
By testing from #5008, it came from CI changes, not the code changes.

DimitrisJim:
This still bugs me, the tests on the PR passed, when merged to main they started failing. What CI changes caused this?

youknowone:
No idea at all. I also experienced locally unreproducible error in #4733
I hope to know how can I have exactly same environment to CI macOS

PR #4997, opened 2023-06-10
Temporarily skip configparser.

DimitrisJim:
Temporarily skip testing the configparser for Mac until #4995 is resolved.


DimitrisJim:
we have an additional crash of socket server.

PR #5004, opened 2023-06-11
fix warning about mutable

MisileLab:


DimitrisJim (approved):
nice. Not sure why clippy doesn't flag this.

PR #5013, opened 2023-06-17
Return SyntaxError with null byte input in eval.

DimitrisJim:
Can't find anything in Python's changelogs but it appears eval now returns a SyntaxError in this case, lets see what the tests think.


DimitrisJim:
lol, mac uses 3.11 which doesn't have the change. I'll just make the test check for either a value error or a syntax error and add a todo.

Issue #5010, opened 2023-06-13
Add `get_int_max_str_digits` and `set_int_max_str_digits` in `sys`.


DimitrisJim:
## Feature

These were newly added in 3.11 and should be relatively straightforward to add, reference: https://docs.python.org/3/library/stdtypes.html#int-max-str-digits

Additions to `int_info` already exist so just the get/set function should be added. (See `PyIntInfo` in [`sys.rs`](https://github.com/RustPython/RustPython/blob/main/vm/src/stdlib/sys.rs)).

bobby-palmer:
It seems like this issue can be closed. Is there more to add with this or is it finished?

DimitrisJim:
Yeah, thanks for noticing! There's some more things to be done for it since it can be set via a command line arg but that should be a separate issue.

PR #5014, opened 2023-06-17
Add `get_int_max_str_digits` and `set_int_max_str_digits` in `sys`.

dannasman:
Hi!

Here is a solution proposal for issue #5010.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -245,7 +245,7 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - uses: actions/setup-python@v4
         with:
-          python-version: "3.11"
+          python-version: "3.11.4"
       - name: Set up the Windows environment
         shell: bash
         run: |
--- a/extra_tests/snippets/stdlib_sys.py
+++ b/extra_tests/snippets/stdlib_sys.py
@@ -92,3 +92,17 @@ if sys.platform.startswith("win"):
     # assert winver.major == winver.platform_version[0]
     # assert winver.minor == winver.platform_version[1]
     # assert winver.build == winver.platform_version[2]
+
+# test int_max_str_digits getter and setter
+
+assert sys.get_int_max_str_digits() == 4300
+sys.set_int_max_str_digits(640)
+assert sys.get_int_max_str_digits() == 640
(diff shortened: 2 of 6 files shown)


youknowone:
@DimitrisJim do you know if it has related tests?

DimitrisJim:
oh it definitely did. shame I didn't remember mentioning that in the issue, I'll look into it. What on earth is going on with this Mac runner though :face_exhaling:

youknowone:
Don't worry about CI failure. I fixed it here https://github.com/RustPython/RustPython/pull/5012/files#diff-575a2149c1bf435cd04009918dec66fd8fa2be92cf7d5fdcdb48ceb61d3ad4aaR11

DimitrisJim:
Thanks for this @dannasman. These are currently tested indirectly in `test_cmd_line` (this is basically an implementation `-X` flag that can be passed when invoking Python and we don't currently have that part implemented iirc so tests fail).

Could you by any chance add an additional test in `extra_tests/snippets/stdlib_sys.py` that plays with these `get/set` functions on a basic level? Something along the lines of get its value, make sure it equals the default, set it to a bad value, error is raised

dannasman:
Sure :+1:

dannasman:
I am having trouble running `pytest -v` with RustPython instead of my default CPython that I have installed. Do you have any tips on how to run pytest with RustPython?

Also, Is this the correct way of testing whether `set_int_max_str_digits` raises the correct error (`ValueError`)?
```python
    sys.set_int_max_str_digits(1)
```

DimitrisJim:
Hey, so yeah if you run it with `pytest -v` you should be fine, it calls the RustPython binary itself (see `test_snippets.py`). Pushing directly here is also an option.

> Also, is this the correct way of testing whether set_int_max_str_digits raises the correct error (ValueError)?

Yup! That looks right to me.

DimitrisJim:
Jesus Christ Mac started using `3.11.4` now. I'll push a commit here if you dont mind @dannasman, the runners using different version of Python to run the snippets has been a pain, sorry for the back and forth.

dannasman:
No problem!

DimitrisJim (approved):
thanks so much @dannasman, lgtm

PR #5015, opened 2023-06-17
Pin Python version to 3.11.4 in CI runs.

DimitrisJim:
It seems that the Mac CI sometimes targets 3.11.3, sometimes 3.11.4 and this is becoming very annoying. Pin the Python version to a specific major.minor.micro. Ideally, we'd like to target major.minor in the future.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -105,7 +105,7 @@ env:
     test_weakref
     test_yield_from
   # Python version targeted by the CI.
-  PYTHON_VERSION: 3.11.4
+  PYTHON_VERSION: "3.11.4"
 
 jobs:
   rust_tests:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -104,6 +104,8 @@ env:
     test_unpack
     test_weakref
     test_yield_from
+  # Python version targeted by the CI.
+  PYTHON_VERSION: 3.11.4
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -104,6 +104,8 @@ env:
     test_unpack
     test_weakref
     test_yield_from
+  # Python version targeted by the CI.
+  PYTHON_VERSION: 3.11.4
```
fanninpm, line 108: When we revert this to major.minor, this should most likely be surrounded by quotation marks. Otherwise, we'd be eventually actually targeting Python 3.2 when we meant to target Python 3.20.

```suggestion
  PYTHON_VERSION: "3.11.4"
```

.github/workflows/cron-ci.yaml
```diff
@@ -7,6 +7,7 @@ name: Periodic checks/tasks
 
 env:
   CARGO_ARGS: --no-default-features --features stdlib,zlib,importlib,encodings,ssl,jit
+  PYTHON_VERSION: 3.11.4
```
fanninpm, line 10: Same here.

```suggestion
  PYTHON_VERSION: "3.11.4"
```

.github/workflows/ci.yaml
DimitrisJim: good point and I was recently reading about this too.

PR #5016, opened 2023-06-18
updated _collections_abc.py and test_collections.py to CPython 3.11.2

bobby-palmer:
#4564 updating _collections_abc.py to 3.11.2 for Python

### DIFF
--- a/Lib/_collections_abc.py
+++ b/Lib/_collections_abc.py
@@ -430,24 +430,12 @@ class _CallableGenericAlias(GenericAlias):
             raise TypeError(
                 "Callable must be used as Callable[[arg, ...], result].")
         t_args, t_result = args
-        if isinstance(t_args, list):
+        if isinstance(t_args, (tuple, list)):
             args = (*t_args, t_result)
         elif not _is_param_expr(t_args):
             raise TypeError(f"Expected a list of types, an ellipsis, "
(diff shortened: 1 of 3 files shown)


DimitrisJim:
thanks @bobby-palmer, this might be a tricky one :smile: there's two test cases now failing in `Lib/test/test_typing.py`:

```
test_errors (test.test_typing.CollectionsCallableTests.test_errors)
test_valid_uses (test.test_typing.ParamSpecTests.test_valid_uses)
```

You can open that file and decorate them with `unittest.expectedFailure` to mark them as so. For `test_errors` you can add this function to `CollectionsCallableTests`:

```python
    # TODO: RUSTPYTHON
    @unittest.expectedFailure
    def test_errors(self): # TODO: RUSTPYTHON, remove when this passes
        super().test_errors() # TODO: RUSTPYTHON, remove when this passes
```

while for `test_valid_uses` just mark the function directly since it isn't brought in via inheritance:

```python
    # TODO: RUSTPYTHON
    @unittest.expectedFailure
    def test_valid_uses(self):
        ....  # function body unchanged
```

bobby-palmer:
@DimitrisJim thanks for the feedback! those tests should be fixed now

DimitrisJim (approved):
nice, guess this wasn't tricky after all. thanks!

Issue #4982, opened 2023-05-20
datetime().astimezone() is broken


xiaxinmeng:
It seems that RustPython supports datetime().astimezone() while the implementation of this function is broken and attribute  'tm_gmtoff' is missing. See the following example.

Example.py
```
from datetime import datetime
datetime(3001, 1, 19, 7, 59, 59, 999999).astimezone() 
```

Output on RustPython:
```
Traceback (most recent call last):
  File "/home/xxm/Desktop/RustPython/example.py", line 199, in <module>
    datetime(3001, 1, 19, 7, 59, 59, 999999).astimezone() 
  File "/home/xxm/.cargo/git/checkouts/rustpython-f8ef4d934ac33cd8/59df92d/pylib/Lib/datetime.py", line 1862, in astimezone
    tz = self._local_timezone()
  File "/home/xxm/.cargo/git/checkouts/rustpython-f8ef4d934ac33cd8/59df92d/pylib/Lib/datetime.py", line 1856, in _local_timezone
    gmtoff = localtm.tm_gmtoff
AttributeError: 'struct_time' object has no attribute 'tm_gmtoff'
```

Behavior on CPython: works normally

Environment:
RustPython: v0.2.0 (https://github.com/RustPython/RustPython#59df92d7)
CPython: 3.9.0
Ubuntu 18.04

DimitrisJim:
both `tm_gmtoff` and `tm_zone` are missing from the `struct_time` struct:

https://github.com/RustPython/RustPython/blob/59df92d700fdf479b7294d44ff3f340ec9a0769c/vm/src/stdlib/time.rs#L318-L328

reference, Python docs on these: https://docs.python.org/3/library/time.html#time.struct_time

 Might be easy to add them.

PR #5017, opened 2023-06-19
added tm_gmtoff and tm_zone attributes to PyStructTime

bobby-palmer:
This was attempting to solve #4982 but I'm currently stuck with trying to get the timezone abbreviation for the system. any help would be appreciated

### DIFF
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -6,6 +6,7 @@ pub use time::*;
 
 #[pymodule(name = "time", with(platform))]
 mod time {
+
     use crate::{
         builtins::{PyStrRef, PyTypeRef},
         function::{Either, FuncArgs, OptionalArg},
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/time.rs
```diff
@@ -325,6 +326,8 @@ mod time {
         tm_wday: PyObjectRef,
         tm_yday: PyObjectRef,
         tm_isdst: PyObjectRef,
+        tm_gmtoff: PyObjectRef,
```
DimitrisJim, line 329: okay, now I see what you were talking about in the other pr, I missed this one, sorry! 

Yeah, these should be objects here, not options. The distinction between naive/aware is done based on their value (it being `None` or not), that's what I've understood by quickly looking into this. I'd need some more time to get a better view of this.

youknowone:
Thank you for contributing. it was not easy to be finished due to lack of pystruct macro support.
This is done by #5391

PR #5018, opened 2023-06-20
Update datetime.py and test_datetime.py to CPYTHON v3.11.2

bobby-palmer:
I also had to add _strptime.py but test_datetime.py fails due to incorrect datetime struct implementation in time.rs
Contributing to issue : `#4564`

### DIFF
--- a/Lib/test/test_datetime.py
+++ b/Lib/test/test_datetime.py
@@ -1,0 +1,57 @@
+import unittest
+import sys
+
+from test.support.import_helper import import_fresh_module
+
+
+TESTS = 'test.datetimetester'
+
+def load_tests(loader, tests, pattern):
+    try:
+        pure_tests = import_fresh_module(TESTS, fresh=['datetime', '_strptime'],
+                                        blocked=['_datetime'])
+        fast_tests = import_fresh_module(TESTS, fresh=['datetime',
+                                                    '_datetime', '_strptime'])
+    finally:
+        # XXX: import_fresh_module() is supposed to leave sys.module cache untouched,
+        # XXX: but it does not, so we have to cleanup ourselves.
+        for modname in ['datetime', '_datetime', '_strptime']:
+            sys.modules.pop(modname, None)
+
+    test_modules = [pure_tests, fast_tests]
+    test_suffixes = ["_Pure", "_Fast"]
+    # XXX(gb) First run all the _Pure tests, then all the _Fast tests.  You might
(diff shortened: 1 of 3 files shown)


Lib/test/test_datetime.py
```diff
@@ -0,0 +1,58 @@
+import unittest
+import sys
+
+from test.support.import_helper import import_fresh_module
+
+
+TESTS = 'test.datetimetester'
+
+@unittest.skip("TODO: RUSTPYTHON")
```
fanninpm, line 9:
```suggestion
@unittest.skip("TODO: RUSTPYTHON, incorrect datetime struct implementation in time.rs")
```

bobby-palmer:
Thanks for the feedback. Do you know why the tests are failing?

Lib/test/test_datetime.py
```diff
@@ -0,0 +1,58 @@
+import unittest
+import sys
+
+from test.support.import_helper import import_fresh_module
+
+
+TESTS = 'test.datetimetester'
+
+@unittest.skip("TODO: RUSTPYTHON, incorrect datetime struct implementation in time.rs")
+def load_tests(loader, tests, pattern):
```
DimitrisJim, line 10: I don't believe you can skip this. This function is part of the `load_tests` protocol and generates the test suite to be executed. 

If you unskip this what is the error?

DimitrisJim: I can take a look later since am on mobile now. There's probably a reason this file wasn't already added.

bobby-palmer: when it loads _strptime.py I get an error of `no attribute : tzinfo` so I guess ill have to fix that before I can add these. On that topic, could you point me in the right direction for creating a field for the time_struct that may of may not be present for a given instance (naive vs timezone aware time struct)

DimitrisJim: aha, gotcha. Makes sense to remove this file for now and see how the updates play with the rest of the test suite.

> could you point me in the right direction for creating a field for the time_struct that may of may not be present for a given instance

It being an option seems like the way to go, the docs state it as optional and looking through Python code I'm seeing it being guarded by if `tzinfo is None` all over the place.

bobby-palmer: Do you know how I would go about making it so that when the option is none the python object no longer has that attribute? I still don't fully understand how pystructsequence works.

youknowone:
Thank you for contributing! we now have 3.12 datetime

PR #5021, opened 2023-06-23
add int_max_str_digits input from command line

### LINKED ISSUE #5019: Add `-X int_max_str_digits` and `PYTHONINTMAXSTRDIGITS` environment variable
## Feature

Follow up to #5010 

These just set the value of `int_max_str_digits` via an implementation specific command line flag (`-X`) and/or the environment variable. These need to be accepted in `settings.rs` and stored in the `Settings` struct. `Flag::from_settings` should be updated to pass the value along.

`test_int_max_str_digits` in `test_cmd_line.py` is a good place to evaluate changes, hopefully (but not necessarily) the test might  will succeed after this change. You can run it locally with:

```bash
cargo run --release --features ssl,jit -- -m test -u all --slowest --fail-env-changed -v test_cmd_line`
```

to get some early feedback.

bobby-palmer:
resolves #5019. Welcoming feedback on best practices because I was unsure how to correctly do error handling and default values

### DIFF
--- a/Lib/test/test_cmd_line.py
+++ b/Lib/test/test_cmd_line.py
@@ -933,8 +933,6 @@ class CmdLineTest(unittest.TestCase):
         self.assertTrue(proc.stderr.startswith(err_msg), proc.stderr)
         self.assertNotEqual(proc.returncode, 0)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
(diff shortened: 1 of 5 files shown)


youknowone (approved):
I am sorry for late review.
Thank you so much for contributing. This is really great!

Issue #4996, opened 2023-06-09
PEP 695 – Type Parameter Syntax


tjf801:
## Feature

Seeing as how Python 3.12 is now in beta, I feel like RustPython should also work on being able to parse the new syntax introduced in [PEP 695](https://peps.python.org/pep-0695/), which makes it much easier to type-hint generic functions and types in the language. I ran into this while trying to start writing a custom type checker, and wanted to use `rustpython_ast` to parse Python 3.12 code that uses this new syntax. Some examples include:
```python
def identity[T](x: T) -> T:
    return x

type Point = tuple[float, float]

type MappingOrSequence[T] = Mapping[int, T] | Sequence[T]

class Container[T](Sequence[T]): ...
```

## Python Documentation or reference to CPython source code

 - [The `type` statement](https://docs.python.org/3.12/reference/simple_stmts.html#type)
 - [Generic functions](https://docs.python.org/3.12/reference/compound_stmts.html#generic-functions)
 - [Generic classes](https://docs.python.org/3.12/reference/compound_stmts.html#generic-classes)
 - [Generic type aliases](https://docs.python.org/3.12/reference/compound_stmts.html#generic-type-aliases)

fanninpm:
I would also recommend posting this issue to https://github.com/RustPython/Parser. One part of the problem is being able to parse the new syntax (that's what the Parser repo is for); the other part of the problem is having the virtual machine make sense of the new syntax (that's what this repo is for).

tjf801:
Oh! I meant to use this issue template, but forgot to actually move the issue to that repository (oops). I'll move this there.

fanninpm:
Like I said, being able to parse the new syntax is not the only problem here.

PR #5026, opened 2023-07-17
Add parser support for PEP 695

zanieb:
Part of https://github.com/RustPython/RustPython/issues/4996

- Adds generated content using the latest Python ASDL
- Bumps RustPython/Parser to the latest version
- Adds stubs for handling generic type parameters and type alias statements

I know more will need to be done to actually _use_ PEP 695, I'd like to address these in separate pull requests if feasible.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1953,7 +1953,7 @@ dependencies = [
 [[package]]
 name = "ruff_source_location"
 version = "0.0.0"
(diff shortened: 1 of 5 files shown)


youknowone:
Thank you so much. Same as the way you said, we also prefer to implement features step by step. You don't need to put too much things in a single PR.

youknowone (approved):
👍

PR #5040, opened 2023-08-16
Add ability to initialize cwd from PWD when targeting WASI

d3lm:
When compiling to WASI the the current directory is set to `/` on boot. That's a bit unfortunate and it means that if you'd list files in `.` it would read the root directory. This PR allows to initialize the current working directory from the `PWD` environment variable. I have added a `target` check to only do this for WASI.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,6 +63,7 @@ pub use settings::{opts_with_clap, RunMode};
 pub fn run(init: impl FnOnce(&mut VirtualMachine) + 'static) -> ExitCode {
     env_logger::init();
 
+    // NOTE: This is not a WASI convention. But it will be convenient since POSIX shell always defines it.
     #[cfg(target_os = "wasi")]
     {
         if let Ok(pwd) = env::var("PWD") {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,6 +63,13 @@ pub use settings::{opts_with_clap, RunMode};
 pub fn run(init: impl FnOnce(&mut VirtualMachine) + 'static) -> ExitCode {
     env_logger::init();
 
+    #[cfg(target_os = "wasi")]
+    {
+        if let Ok(pwd) = env::var("PWD") {
+            let _ = env::set_current_dir(pwd);
+        };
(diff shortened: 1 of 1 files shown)


fanninpm:
cc @lastmjs

youknowone:
Thank you for contributing!
Because I am not good at WASI, I'd like to understand if this is an expected (either de jure or de facto) standard for WASI apps. Could you confirm if it is or not?

d3lm:
The way it's implemented is not a de facto solution and WASI doesn't have a notion of a cwd and any paths are always resolved via pre-opened directories. However, IMO it's reasonable to allow to set the cwd to something other than `/` and it wouldn't have any other implications other than `.` or `..` etc. would be resolved properly to the cwd that gets set. Without this it's not possible to e.g. list files in a directory for both `.` **and** `/` because if the cwd is `/` then `.` resolves to `/` even if the user is running python from some sub-directory.

src/lib.rs
```diff
@@ -63,6 +63,13 @@ pub use settings::{opts_with_clap, RunMode};
 pub fn run(init: impl FnOnce(&mut VirtualMachine) + 'static) -> ExitCode {
     env_logger::init();
 
+    #[cfg(target_os = "wasi")]
+    {
```
youknowone, line 67: Thank you for explanation. I think adding this feature is reasonable. Because this is neither a WASI feature not a Python feature, can we just put a note that this is our own convenience feature?

```suggestion
    {
        // NOTE: This is not a WASI convention. But it will be convenient since POSIX shell always defines it.
```

d3lm: Oh yea sure. I can add a note about this. Thanks for being open to this 🙏 I agree with you that it's a convenience feature from RustPython in this case.

d3lm: Done.

youknowone (approved):
Thank you so much! Please let us know if you find anything lack of WASI support.

d3lm:
Will do! Thanks a ton for getting this in, and also for all your work on this! This is really huge 🙏

PR #5049, opened 2023-08-27
Fix `encodings` related error messages to be less confusing

youknowone:
- Don't try `import encodings` when `Settings::path_list` is empty
  - And adding `.` is now part of `rustpython` crate. Users can skip it when they don't want to add cwd to path_list when they create their own interpreter.
- More information from encoding failure error messages for both interpreter developers/users and rustpython-vm users.


vm/src/vm/mod.rs
```diff
@@ -223,18 +223,30 @@ impl VirtualMachine {
...
+            let pythonpath_env = std::env::var("PYTHONPATH").ok();
+            let env_set = rustpythonpath_env.as_ref().is_some() || pythonpath_env.as_ref().is_some();
+            let path_contains_env = self.state.settings.path_list.iter().any(|s| {
+                Some(s.as_str()) == rustpythonpath_env.as_deref() || Some(s.as_str()) == pythonpath_env.as_deref()
+            });
+
+            let guide_message = if !env_set {
+                "Neither RUSTPYTHONPATH nor PYTHONPATH is set. Try setting one of them to the stdlib directory."
+            } else if path_contains_env {
+                "RUSTPYTHONPATH or PYTHONPATH is set, but it doesn't contain encodings library. If you are going to customize RustPython vm/interpreter, try adding it to the path. If you are developing RustPython interpreter, it might be a bug during development."
```
fanninpm, line 236: grammar nitpick:

```suggestion
                "RUSTPYTHONPATH or PYTHONPATH is set, but it doesn't contain the encodings library. If you are customizing the RustPython vm/interpreter, try adding the stdlib directory to the path. If you are developing the RustPython interpreter, it might be a bug during development."
```

vm/src/vm/mod.rs
```diff
@@ -223,18 +223,30 @@ impl VirtualMachine {
...
             } else {
-                "Could not import encodings. Try adding your path to Setting struct's path_list field. If you don't have \
-                    access to a consistent external environment (e.g. if you're embedding \
-                    rustpython in another application), try enabling the freeze-stdlib feature"
-                    .to_owned()
+                "RUSTPYTHONPATH or PYTHONPATH is set, but it wasn't loaded to `Settings::path_list`. If you are going to customize RustPython vm/interpreter, those environment variables are not loaded to Settings by default. Please try it yourself. If you are developing RustPython interpreter, it might be a bug during development."
```
fanninpm, line 238: grammar nitpick:

```suggestion
                "RUSTPYTHONPATH or PYTHONPATH is set, but it wasn't loaded to `Settings::path_list`. If you are going to customize the RustPython vm/interpreter, those environment variables are not loaded in the Settings struct by default. Please try creating a customized instance of the Settings struct. If you are developing the RustPython interpreter, it might be a bug during development."
```

vm/src/vm/mod.rs
```diff
@@ -223,18 +223,30 @@ impl VirtualMachine {
...
             };
 
+            let msg = format!(
+                "Could not import encodings. It usually means something goes wrong. Please carefully read messages and follow the steps.\n\
+                \n\
+                {guide_message}\n\
+                If you don't have access to a consistent external environment (e.g. targeting wasm, embedding \
+                    rustpython in another application), try enabling the freeze-stdlib feature.\n\
+                If this is intended and you want to exclude encodings library from your interpreter, please remove `encodings` feature from `rustpython-vm` crate."
```
fanninpm, line 247: grammar nitpick/editing for clarity:

```suggestion
                "RustPython could not import the encodings module. It usually means something went wrong. Please carefully read the following messages and follow the steps.\n\
                \n\
                {guide_message}\n\
                If you don't have access to a consistent external environment (e.g. targeting wasm, embedding \
                    rustpython in another application), try enabling the `freeze-stdlib` feature.\n\
                If this is intended and you want to exclude the encodings module from your interpreter, please remove the `encodings` feature from `rustpython-vm` crate."
```

fanninpm (commented):
A few nitpicks to help with grammar/clarity.

youknowone:
@fanninpm Thank you! Is this now looking good?

vm/src/vm/mod.rs
```diff
@@ -325,9 +334,21 @@ impl VirtualMachine {
...
+                );
+                self.print_exception(e);
+            }
+        } else {
+            // Here may not be the best place to give general `path_list` advice,
+            // but bare rustpython_vm::VirtualMachine users skipped proper settings must hit here while properly setup vm never enters here.
+            eprintln!(
+                "feature `encodings` is enabled but `settings.path_list` is empty. This is likely a user-side bug. \
+                Please add library path to `settings.path_list`. If you intended to disable stdlib including `encodings`, please also disable `encodings` feature.\n\
+                Tip: You may also want to add `\"\"` to `settings.path_list` to enable importing from current working directory."
```
youknowone, line 350: @fanninpm I am sorry to add a commit during review. I found silently skipping this branch can lead users be confused. So this fix will be good to put here. Could you also check this message too?

fanninpm: In my opinion, telling the user about "user error" seems a little bit off-putting. Other than that, I have a couple of minor changes:

```suggestion
            // Here may not be the best place to give general `path_list` advice,
            // but bare rustpython_vm::VirtualMachine users skipped proper settings must hit here while properly setup vm never enters here.
            eprintln!(
                "feature `encodings` is enabled but `settings.path_list` is empty. \
                Please add the library path to `settings.path_list`. If you intended to disable the entire standard library (including the `encodings` feature), please also make sure to disable the `encodings` feature.\n\
                Tip: You may also want to add `\"\"` to `settings.path_list` in order to enable importing from the current working directory."
```

PR #5054, opened 2023-08-30
update imp.py from CPython 3.11

Reid00:
related: https://github.com/RustPython/RustPython/issues/4564

### DIFF
--- a/Lib/hashlib.py
+++ b/Lib/hashlib.py
@@ -114,8 +114,8 @@ def __get_builtin_constructor(name):
             cache['shake_128'] = _sha3.shake_128
             cache['shake_256'] = _sha3.shake_256
     except ImportError:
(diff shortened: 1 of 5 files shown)


youknowone:
This PR  contains previously merged commit about test_hashlib.
You probably want to use `git rebase` or `git reset --hard` to clear the upstream context before making a patch.
Using patch branches rather than main will also be helpful to make those work easier.

youknowone:
Would you check if your git email configuration matches to the github id?

Checking current config:
```
git config user.email
```

You can change either the local config or GitHub config to make them aligned

Reid00:
Thank you for your advice. I changed the local user.email and will use patch branches/

youknowone (approved):
Thank you for contributing!

PR #5057, opened 2023-08-30
Print version and info text when executing the shell

d3lm:
This PR prints the same initial messages that you'd expect in the CPython shell, at least to some extend.

image


youknowone (approved):
Thank you!

PR #5064, opened 2023-09-13
simplify rustfmt invocation in CI

calebcartwright:
We added the `--check` flag directly to cargo fmt some time back, so when I stumble across cases of the older more verbose versions I open these types of PRs to increase awareness.

Additionally, when `cargo fmt` is invoked in the root of a workspace it will automatically format all the workspace members (you can see that `cargo fmt --all --check -- --verbose` will enumerate the same files as `cargo fmt --check -- --verbose`).

Our `--all` flag is a bit dated compared to other tools, and is only necessary with `cargo fmt` in some more edge type cases (e.g. wanting to format packages that aren't explicit workspace members, or when invoking from certain subdirectories)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -308,7 +308,7 @@ jobs:
         with:
             components: rustfmt, clippy
       - name: run rustfmt
-        run: cargo fmt --all -- --check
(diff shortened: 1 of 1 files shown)


youknowone (approved):
Thank you!

PR #5081, opened 2023-10-05
Implemented compare operation for boolean types in JIT engine

MannarAmuthan:
In JIT, Implemented more comparison operations for boolean types, and added tests

1. Verified with CPython implementation
2. Added necessary Tests

### DIFF
--- a/jit/src/instructions.rs
+++ b/jit/src/instructions.rs
@@ -348,6 +348,22 @@ impl<'a, 'b> FunctionCompiler<'a, 'b> {
                             .push(JitValue::Bool(self.builder.ins().bint(types::I8, val)));
                         Ok(())
                     }
+                    (JitValue::Bool(a), JitValue::Bool(b)) => {
+                        let cond = match op {
+                            ComparisonOperator::Equal => IntCC::Equal,
+                            ComparisonOperator::NotEqual => IntCC::NotEqual,
+                            ComparisonOperator::Less => IntCC::UnsignedLessThan,
+                            ComparisonOperator::LessOrEqual => IntCC::UnsignedLessThanOrEqual,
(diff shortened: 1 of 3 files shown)


jit/src/instructions.rs
```diff
@@ -348,6 +348,22 @@ impl<'a, 'b> FunctionCompiler<'a, 'b> {
                             .push(JitValue::Bool(self.builder.ins().bint(types::I8, val)));
                         Ok(())
                     }
+                    (JitValue::Bool(a), JitValue::Bool(b)) => {
```
youknowone, line 351: python `bool` is a subclass of `int` and they can be mixed in operations. This pattern looks like matching only for (bool, bool) operands. Isn't this better to be combined to (int, int) case? Then it will allows also (bool, int) or (int, bool) operations.

MannarAmuthan: Thanks for response!

Yeah! makes sense , even in cpython that's how it is.
However the comparison methods for integers are not working for booleans.
Especially those signed versions. That's why I put it in different block.

I will make changes as per suggestion , and update it.

MannarAmuthan: Yeah! Makes sense , will update it. Thanks for response !

MannarAmuthan:
This is done, Now both int and bool types can be compared.

youknowone (approved):
Thank you!

youknowone:
oh, could you run `cargo fmt --all` and rebase it on current main branch?

MannarAmuthan:
Sure

MannarAmuthan:
Hi , This is rebased with current main branch

youknowone:
you may have forgot fetching upstream. rebased again.

PR #5086, opened 2023-10-07
[JIT]  Binary operations for integers and floating-point numbers allowing mixed type calculations

MannarAmuthan:
Implemented Binary operations (ADD, SUB, MUL, DIV for now) for allowing mixed type calculations with integers and floating point numbers

1. I assume that division by zero is yet to be handled in current JIT implementation.
2. Added necessary tests


youknowone (approved):
Thank you!

PR #5088, opened 2023-10-08
implement PyObject_Dir for ByObjectRef

dannasman:
Hi!

Here is a solution proposal regarding the addition of PyObject_Dir to ByObjectRef (#3244).


youknowone (approved):
Thank you!

PR #5090, opened 2023-10-13
implement PyObject_GetAIter for PyObject

dannasman:
Hi!

Here is a solution proposal regarding the addition of PyObject_GetAIter to PyObject (#3244).


youknowone (approved):
Thank you!

PR #5091, opened 2023-10-14
implement PyObject_Type and PyObject_TypeCheck

dannasman:
Hi!

Here are proposed implementations for PyObject_Type and PyObject_TypeCheck related to issue (#3244 ).


youknowone (approved):
I felt like `obj.type()` already is `PyObject_Type`. Maybe not due to return type.

Thank you!

PR #5094, opened 2023-10-17
Add missing 'sysinfoapi' feature for winapi

zaddach:
stdlib doesn't compile on Windows MSVC x64 otherwise.


fanninpm:
I believe we're in the process of transitioning away from the `winapi` crate towards the `windows` crate.

cc @youknowone

youknowone (approved):
@zaddach you seem to set git email wrong or at least not same as your github account. Could you check it?

@fanninpm Yes, but I think that's not related to this issue. If we are not going to do it right now, we will want to fix build bug.

zaddach:
@youknowone I've updated the email address in the commits to the one on Github.

youknowone:
Thank you!

PR #5095, opened 2023-10-18
Add type_check to _check()

dannasman:
Hi!

After implementing `type_check` (#5091) I tried to add the `type_check` call in `_check` function by following the [CPython implementation](https://github.com/python/cpython/blob/cb1bf89c4066f30c80f7d1193b586a2ff8c40579/Objects/descrobject.c#L77). I also modified `descr_get` function of `PyGetSet` based on its [CPython counterpart](https://github.com/python/cpython/blob/cb1bf89c4066f30c80f7d1193b586a2ff8c40579/Objects/descrobject.c#L179). These modifications cause the code to panic and I am not quite sure why. Any thoughts?

### DIFF
--- a/vm/src/builtins/getset.rs
+++ b/vm/src/builtins/getset.rs
@@ -51,18 +51,19 @@ impl GetDescriptor for PyGetSet {
         _cls: Option<PyObjectRef>,
         vm: &VirtualMachine,
     ) -> PyResult {
-        let (zelf, obj) = match Self::_check(&zelf, obj, vm) {
-            Some(obj) => obj,
-            None => return Ok(zelf),
-        };
-        if let Some(ref f) = zelf.getter {
-            f(vm, obj)
+        if let Some(obj) = obj {
+            let (zelf, obj) = Self::_check(&zelf, obj, vm)?;
+            if let Some(ref f) = zelf.getter {
+                f(vm, obj)
+            } else {
+                Err(vm.new_attribute_error(format!(
+                    "attribute '{}' of '{}' objects is not readable",
+                    zelf.name,
+                    Self::class(&vm.ctx).name()
+                )))
+            }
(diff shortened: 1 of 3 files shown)


youknowone:
Welcome to the common RustPython development rabbit hole. Well, I only have general debug advices. It can be either a bug here or combination of other incompatabilty issues there - because getset is shown in error message, I suspect our incomplete getset_descriptor implementation may be releated.
General idea is breaking it down enough and see how it is going. When an exception is raised during initialization(VirtualMachine::initialize) step, you will want to set breakpoint and add some print_exception or other debugging helpers in `VirtualMachine::initialize` function.

I am currently on vacation. I can try a quick look, but not able to look deep down stuff for a while. (To be honest, usually not very helpful for these kind of problems even if I do that)

dannasman:
Alright, I'll keep digging deeper. Have a nice vacation!

youknowone:
@dannasman Do you have a specific goal to achieve? Otherwise looking around similar issues may be helpful. I also have many abandoned patches when I got stucked https://github.com/RustPython/RustPython/pulls/youknowone

dannasman:
I was searching for places where PyObject_TypeCheck should be used in the code and found the `_check()` function. The original goal was to add `type_check` into it just like in CPython. Further investigation showed that the implementations between RustPython and CPython have some differences when it comes to descriptors. To my current knowledge In RustPython `generic_getattr_opt()` function descriptor objects seem to always get a type `getset_descriptor` whereas in corresponding CPython function `_PyObject_GenericGetAttrWithDict` the descriptor objects are assigned the same type (or subtype) as the `self` object (and this is later checked with `PyObject_TypeCheck` to make sure this is the case). So it might be the case that `_check()` does not actually need `type_check()`.

PR #5096, opened 2023-10-18
Fix examples/package_embed

youknowone:


olivierlemasle (approved):
Fixes `cargo run --example package_embed`, thanks!

PR #5097, opened 2023-10-19
fix type_check

dannasman:
Hi!

I noticed that there was an error in my earlier pull request (#5091 ). `fast_isinstance()` should be called by `self` instead of `self.class()` in `type_check`. Sorry for the inconvenience.


youknowone (approved):
:+1:

Issue #5105, opened 2023-10-24
get-pip fails on ubuntu


iHnR:
## Summary

The command `rustpython --install-pip get-pip` fails with an error I haven't found elsewhere. I installed rustpyton with `cargo install --git https://github.com/RustPython/RustPython --features ssl`  
The rest of rustpython seems to work fine.

## Expected

pip installs

## Actual

This is the error message:
```
downloading get-pip.py...
running get-pip.py...
Traceback (most recent call last):
  File "get-pip.py", line 32992, in <module>
    main()
  File "get-pip.py", line 139, in main
    shutil.rmtree(tmpdir, ignore_errors=True)
  File "get-pip.py", line 135, in main
    bootstrap(tmpdir=tmpdir)
  File "get-pip.py", line 135, in main
    bootstrap(tmpdir=tmpdir)
  File "get-pip.py", line 111, in bootstrap
    monkeypatch_for_cert(tmpdir)
  File "get-pip.py", line 92, in monkeypatch_for_cert
    from pip._internal.commands.install import InstallCommand
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_internal/commands/__init__.py", line 9, in <module>
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_internal/cli/base_command.py", line 15, in <module>
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_internal/cli/cmdoptions.py", line 24, in <module>
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_internal/cli/parser.py", line 12, in <module>
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_internal/configuration.py", line 20, in <module>
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_internal/exceptions.py", line 18, in <module>
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_vendor/requests/__init__.py", line 43, in <module>
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_vendor/urllib3/__init__.py", line 13, in <module>
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_vendor/urllib3/connectionpool.py", line 39, in <module>
  File "/tmp/tmp_7cqaye7/pip.zip/pip/_vendor/urllib3/request.py", line 191, in <module>
TypeError: __class__ assignment only supported for types without a payload
```

rominf:
[...]

PR #5108, opened 2023-11-02
Update ensurepip to Python 3.12

youknowone:
related to #5105, not tested to fix it or not


youknowone:
#5228 fix this PR too

PR #5110, opened 2023-11-03
Add Lib/test/test___all__.py

NakanoMiku39:
Add a new test file Lib/test/test___all__.py 

I don't see test___all__.py file so I copied this from cpython version 3.12


youknowone (approved):
The patch looks perfect. Thank you for contributing!

PR #5111, opened 2023-11-06
Remove outdated info about yanked crates

olivierlemasle:
The `rustpython-*` crates have been published since this message was added.


youknowone (approved):
Thank you!

PR #5114, opened 2023-11-08
Update test__locale.py and test_atexit.py from CPython v3.12

NakanoMiku39:
I don't know if the update of Cargo.lock is important so I included it.


youknowone (approved):
Thanks! And yes, we update Cargo.lock to guarantee reproducible build of our binary

PR #5115, opened 2023-11-09
Update test_opcodes.py, test_codeop.py and codeop.py from CPython v3.12

NakanoMiku39:
There is no test_opcodes.py so I copied it from cpython
This file is dependent on Lib/test/typinganndata folder so I copied it too
Then I removed test_opcode.py
I also updated both codeop.py and test_codeop.py


youknowone (approved):
Thank you! 👍

PR #5117, opened 2023-11-13
Make PyMethodDef construction const

coolreader18:


youknowone (approved):
👍

coolreader18:
Oh shoot I pushed it to `RustPython` instead of my fork, didn't I. whoops

coolreader18:
Ideally this might improve startup performance a tiny bit but I think that's pretty negligible already

Issue #5104, opened 2023-10-23
Update Python libraries and tests from CPython 3.12


youknowone:
Though it will be sometimes blocked by rust side issues, don't worry!
Please just create PR about the broken test and and create issues related to the new bugs.

If you are new to updating cpython libraries, this guide will be useful:
https://github.com/RustPython/RustPython/wiki/How-to-update-test-files

When updating module Lib/X, please don't forget to also update Lib/test/test_X

A few libraries doesn't need to be updated because CPython also kept it same as previous version. In that case, please leave a comment to let us mark it as done. Thank you!

Tips
- To get specific version of CPython source code, e.g. try `git checkout v3.12.0` from CPython repository.
- Please check #4064 for example. Not the entire changes but the each commits. One of the middle size example is https://github.com/RustPython/RustPython/commit/17e12dea1ec11b871a9e5d3a4a26e2f9ad83fde2 . Sometimes it can be really simple like https://github.com/RustPython/RustPython/pull/4064/commits/9571a68c01284191d4b18229694c303dad8992ff
- Please create one PR for one library+test pair whenever possible. Once you stuck by dependencies, go to one of the dependency first and finish it before restarting the stuck one unless they become simpler to be updated at the same time. Normally adding more libraries make passing tests (a lot) harder.
- To run tests, try `cargo run --features ssl -- -m test -v <test_name>`. e.g. If you changed `test_unicode.py`, `cargo run --features ssl -- -m test -v test_unicode`
  - If it runs too slow, try to add `--release` like `cargo run --release --features ssl -- -m test -v test_unicode`

# List of libraries
- [x] `__future__.py` #5257
  - Related test is  `test_future_stmt`
  - [ ] maybe related to #4060?
- [x] `__hello__.py` unchanged
- [x] `__phello__` unchanged
- [x] `_compression.py` unchanged
- [x] `_markupbase.py` unchanged
- [x] `_osx_support.py` unchanged
- [ ] `_threading_local.py`
- [ ] `_weakrefset.py`
- [x] `abc.py` #5134 
  - `_collections_abc.py`
  - `_py_abc.py`
- [x] `aifc.py` removed in 3.13
- [x] `antigravity.py` unchanged
- [x] `argparse.py` #5134 
- [x] `ast.py` #5121
- [x] `asynchat.py` #5134  removed in 3.13
- [x] `asyncio` #3858
  - [x] #3859
  - requires `_contextvars`
  - requires `_overlapped` for windows
[...]

PR #5121, opened 2023-11-22
Update ast, test_ast from CPython 3.12.0

cygao90:
One of testcases contains`match-case` statement which is not supported so I commented it out.
#5104


youknowone (approved):
Thank you!

Issue #5101, opened 2023-10-23
Python 3.12 added slice hash 


youknowone:
Related test: https://github.com/RustPython/RustPython/pull/5078/commits/cfda063fb4bc1aecb0a726893928912eab01b5c7

JazzGlobal:
Should this issue be closed?

youknowone:
@JazzGlobal Thank you for confirming!

PR #5123, opened 2023-11-24
Support slice hash

cygao90:
related to #5101.
This PR implemented slice hash based on [sliceobject.c](https://github.com/python/cpython/blob/main/Objects/sliceobject.c#L653) and updated cooresponding tests.

It seems that the hash results for many objects are different from CPython's, e.g. `hash(None)`.

### DIFF
--- a/extra_tests/snippets/builtin_slice.py
+++ b/extra_tests/snippets/builtin_slice.py
@@ -82,7 +82,7 @@ assert_raises(TypeError, lambda: slice(0) > 3)
 assert_raises(TypeError, lambda: slice(0) <= 3)
 assert_raises(TypeError, lambda: slice(0) >= 3)
 
-# TODO: slice is hashable in CPython 3.12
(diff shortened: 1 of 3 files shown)


extra_tests/snippets/builtin_slice.py
```diff
@@ -82,7 +82,7 @@
 assert_raises(TypeError, lambda: slice(0) <= 3)
 assert_raises(TypeError, lambda: slice(0) >= 3)
 
-# TODO: slice is hashable in CPython 3.12
+# slice is hashable in CPython 3.12
 # assert_raises(TypeError, hash, slice(0))
 # assert_raises(TypeError, hash, slice(None))
 #
```
youknowone, line 88: Thank you! These lines were tests for < 3.11 versions. Could you also remove those lines?

youknowone (approved):
Thank you so much! The changes looks good.
I left a comment for a minor change.

PR #5124, opened 2023-11-25
None.__ne__(None) should be NotImplemented

### LINKED ISSUE #5103: None.__ne__(None) behavior is changed in 3.12
Related test:  https://github.com/youknowone/RustPython/blob/6e699d93fe9ed1f16921ef3ab4a83fc0a72c5473/extra_tests/snippets/builtin_none.py#L25

evanrittenhouse:
This was changed in Python 3.12. Closes #5103

### DIFF
--- a/vm/src/builtins/singletons.rs
+++ b/vm/src/builtins/singletons.rs
@@ -2,10 +2,12 @@ use super::{PyStrRef, PyType, PyTypeRef};
 use crate::{
     class::PyClassImpl,
     convert::ToPyObject,
+    function::{PyArithmeticValue, PyComparisonValue},
     protocol::PyNumberMethods,
-    types::{AsNumber, Constructor, Representable},
-    Context, Py, PyObjectRef, PyPayload, PyResult, VirtualMachine,
+    types::{AsNumber, Comparable, Constructor, PyComparisonOp, Representable},
+    Context, Py, PyObject, PyObjectRef, PyPayload, PyResult, VirtualMachine,
 };
+use rustpython_vm::AsObject;
 
 #[pyclass(module = false, name = "NoneType")]
 #[derive(Debug)]
@@ -42,7 +44,7 @@ impl Constructor for PyNone {
     }
 }
 
-#[pyclass(with(Constructor, AsNumber, Representable))]
+#[pyclass(with(Constructor, Comparable, AsNumber, Representable))]
 impl PyNone {
     #[pymethod(magic)]
     fn bool(&self) -> bool {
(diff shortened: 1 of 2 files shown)


vm/src/builtins/singletons.rs
```diff
@@ -118,6 +120,22 @@ impl Representable for PyNotImplemented {
...
 }
 
+impl Comparable for PyNone {
+    #[allow(unused_variables)]
+    fn cmp(
+        zelf: &Py<Self>,
+        other: &PyObject,
+        op: PyComparisonOp,
+        vm: &VirtualMachine,
+    ) -> PyResult<PyComparisonValue> {
```
youknowone, line 130: you didn't check `op` which holds the kind of operator (eq, ne, gt, ...)

evanrittenhouse: Yeah, I've got it as in progress as I'm using CI to test. Even with PyEnv shelled at 3.12, pytest was still picking up 3.11. Will work on it some more tonight, though I expect there'll be more to do outside of implementing the operator check

King44447:
EnoN .

vm/src/builtins/singletons.rs
evanrittenhouse: @youknowone any reason why [this block](https://github.com/RustPython/RustPython/blob/0fab6e606379cfda346c82fa2f3960d0449b40e9/vm/src/builtins/object.rs#L105-L111) uses `PyComparison::Eq` then negates it instead of `PyComparisonOp::Ne`?

youknowone: That's a default behavior of general object. It helps a user-defined class gives peroper `!=` result when it doesn't implement `__ne__`. We can overwrite anything in subclass. Though I am not sure about it because I didn't check what's changed in recent python versions, I believe we will not need to change this part in high chance.

evanrittenhouse:
@youknowone this is ready now, but I'm not sure why tests failed on wasm-wasi.

youknowone (approved):
Thank you for contributing! It looks almost perfect. I left a few nitpicks.

evanrittenhouse:
Done - thanks @youknowone!

youknowone:
Could you also check clippy errors? To check it, see the 'Details' of ubuntu-latest build.
Rebasing the patch on RustPython/RustPython main will also fix other CI failures unrelated to this patch.

I am really sorry for late response during you are writing this patch. And thank you so much for contributing.

evanrittenhouse:
Weird, now it's failing on a manual hash implementation? I've rebased against `main` though

youknowone:
That's due to recent Rust version upgrade. I fixed it on main branch. rebasing on main again will fix it.

evanrittenhouse:
Sorry, must've missed your comment - rebased

youknowone (approved):

Issue #5130, opened 2023-12-13
Clarify license


h7kanna:
## Summary

RustPython uses https://github.com/mhogrefe/malachite which is licensed as LGPL.
This does impact the licensing of projects(non-open-source) that embed RustPython.
Do you think this should be mentioned in the documentation?
Just the [LICENSE](https://github.com/RustPython/RustPython/blob/main/LICENSE) is not clear enough.

youknowone:
Thank you, that's a good point.

I think we still can support num-bigint instead of malachite-bigint without code change. (except for import)
We need to maintain feature flag for that and documentation.

h7kanna:
Oh, I could work on adding the feature flag for it then. 
Just started looking at this project. Will figure out the 'num-bigint' dep part.

h7kanna:
It seems like the Parser's format crate makes the `malachite` a hard dependency
https://github.com/RustPython/Parser/blob/main/format/Cargo.toml

Also embedding examples [not working](https://github.com/RustPython/RustPython/issues/5120) with tag `0.3.0` can be related.

In fact on my machine MacOS, with the latest Rust(1.74) the Parser doesn't event compile
```
   Compiling rustpython-ast v0.3.1 (/Users/h7kanna/Repos/RustPython/Parser/ast)
   Compiling rustpython-format v0.3.1 (/Users/h7kanna/Repos/RustPython/Parser/format)
error[E0252]: the name `bigint` is defined multiple times
  --> ast/src/lib.rs:28:9
   |
26 | pub use malachite_bigint as bigint;
   |         -------------------------- previous import of the module `bigint` here
27 | #[cfg(feature = "num-bigint")]
28 | pub use num_bigint as bigint;
   |         ^^^^^^^^^^^^^^^^^^^^ `bigint` reimported here
   |
   = note: `bigint` must be defined only once in the type namespace of this module
help: you can use `as` to change the binding name of the import
   |
28 | pub use num_bigint as other_bigint;
   |         ~~~~~~~~~~~~~~~~~~~~~~~~~~

For more information about this error, try `rustc --explain E0252`.
error: could not compile `rustpython-ast` (lib) due to previous error
warning: build failed, waiting for other jobs to finish...
error: could not compile `rustpython-ast` (lib test) due to previous error
```

JamesClarke7283:
Also these packages are under the Business Source License:
BSL-1.0 (2): clipboard-win, error-code

[...]

PR #5131, opened 2023-12-16
Added num/malachite-bigint features

h7kanna:
As per the comment, https://github.com/RustPython/RustPython/issues/5130#issuecomment-1853251581 added support for both `num-bigint` and `malachite-bigint` to fix https://github.com/RustPython/RustPython/issues/5130

Changes are tested on Mac and Ubuntu. 
Some tests fail, but they also fail without these changes. So I assume these are flaky.

**Note**: 

As the `num-bigint` and `malachite-bigint` are mutually exclusive features, care should be taken when changing things(deps) around. Most likely, adding 'default-features = false' will suffice.
This should be worthwhile as the malachite dependency is LGPL and it complicates licensing of any projects that embed RustPython.

Example usage:

```
rustpython-vm = { git = "https://github.com/h7kanna/rustpython", branch = "num-malachite-bigint-features", default-features = false, features = ["compiler", "serde", "num-bigint"]  }
rustpython = { git = "https://github.com/h7kanna/rustpython", branch = "num-malachite-bigint-features", default-features = false, features = ["num-bigint", "stdlib","threading"]}
```

Tests when `num-bigint` enabled
```
cargo test --workspace --exclude rustpython_wasm --no-default-features --features num-bigint,stdlib,threading
```

About CI:

Again as the features are mutually exclusive, CI workflow should be run with both enabled separately potentially doubling the run time. I added the 'malachite-bigint' in the workflow as it is the default and the default tests are successful

6 tests are failing when the 'num-bigint' is enabled, they need to be checked/fixed. That can be done in a separate PR if these changes are okay.

Also, we can refactor the code to use a separate crate 'rustpython-bigint' similar to https://github.com/RustPython/Parser/blob/main/format/src/bigint.rs

### DIFF
--- a/vm/src/builtins/float.rs
+++ b/vm/src/builtins/float.rs
@@ -17,9 +17,15 @@ use crate::{
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyRef, PyResult,
     TryFromBorrowedObject, TryFromObject, VirtualMachine,
 };
+#[cfg(feature = "malachite-bigint")]
 use malachite_bigint::{BigInt, ToBigInt};
+#[cfg(feature = "num-bigint")]
+use num_bigint::{BigInt, ToBigInt};
 use num_complex::Complex64;
+#[cfg(feature = "num-bigint")]
+use num_rational::Ratio;
 use num_traits::{Signed, ToPrimitive, Zero};
+#[cfg(feature = "malachite-bigint")]
 use rustpython_common::int::float_to_ratio;
 use rustpython_format::FormatSpec;
 
@@ -455,6 +461,7 @@ impl PyFloat {
         crate::literal::float::is_integer(self.value)
     }
 
+    #[cfg(feature = "malachite-bigint")]
     #[pymethod]
     fn as_integer_ratio(&self, vm: &VirtualMachine) -> PyResult<(PyIntRef, PyIntRef)> {
         let value = self.value;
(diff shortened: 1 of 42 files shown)


h7kanna:
I am working to simplify this by abstracting `bigint ' into a separate crate.
I will update the PR if I can do it.

h7kanna:
6 tests fail compared to malachite-bigint feature. Need to fix

```sh
287 tests OK.

6 tests failed:
    test_grp test_pathlib test_posixpath test_pwd test_shutil
    test_socket

10 tests skipped:
    test_compile test_ctypes test_devpoll test_epoll test_kqueue
    test_socketserver test_univnewlines test_urllib2net test_urllibnet
    test_zipfile64

```
num-bigint

```sh
281 tests OK.

12 tests failed:
    test_grp test_gzip test_long test_pathlib test_posixpath test_pwd
    test_shutil test_socket test_tarfile test_xmlrpc test_zipfile
    test_zlib

10 tests skipped:
    test_compile test_ctypes test_devpoll test_epoll test_kqueue
    test_socketserver test_univnewlines test_urllib2net test_urllibnet
    test_zipfile64
```

youknowone:
Does this patch closes #5130?

h7kanna:
Hi @youknowone, this is ready for review. 
Next CI should pass all the default 'malachite-bigint' tests.
These are many changes, but rather mechanical. Mostly reverting this [PR](https://github.com/RustPython/RustPython/pull/4952) and adding features.

youknowone:
umm, so many cfg and features only for bigint.
@qingshi163 do you any idea we can do this better?

vm/src/builtins/float.rs
```diff
@@ -472,6 +479,26 @@ impl PyFloat {
             })
     }
 
+    #[cfg(feature = "num-bigint")]
+    #[pymethod]
+    fn as_integer_ratio(&self, vm: &VirtualMachine) -> PyResult<(PyIntRef, PyIntRef)> {
```
youknowone, line 484: what's happened for this function? was there incompatibility?

youknowone (commented):
I can't take this much complexity for bigint. My suggestion is creating `rustpython-bigint` in parser repository and make every crate to use it. Then each crate need to depending on it by feature. I believe we don't need to pass the bigint features to other sub crates in this case - because enabling both `rustpython-bigint/malachite` and `rustpython-bigint/num` will raise error.

h7kanna:
Yes, I agree. It is complex. I will make changes to create 'rustpython-bigint' crate.
This is just to make it work as I learn the codebase.

vm/src/builtins/float.rs
h7kanna: Yes, took this code from the reverted PR.
https://github.com/RustPython/RustPython/pull/4952/files#diff-d030ca0d2309f322c97032ab7179319e19b3806317cac7194ce0d0ef6c55ff1aL461

PR #5140, opened 2023-12-28
Fix abc error messages

### LINKED ISSUE #5137: Mismatching abstractmethod error messages
Python 3.12 changed error messages.

https://github.com/RustPython/RustPython/pull/5134/files#diff-e020d5ddd536ec48b0414753af7d5240750a05262f32dad378b4c79e6cc3743dR152-R160 (#5134)

Related source code in RustPython is in `vm/src/builtins/object.rs`.

NakanoMiku39:
Fix #5137

### DIFF
--- a/Lib/test/test_abc.py
+++ b/Lib/test/test_abc.py
@@ -157,8 +157,6 @@ def test_factory(abc_ABCMeta, abc_get_cache_token):
             msg = r"class C without an implementation for abstract method 'method_one'"
             self.assertRaisesRegex(TypeError, msg, C)
 
-        # TODO: RUSTPYTHON
-        ##@unittest.expectedFailure
         def test_object_new_with_many_abstractmethods(self):
             class C(metaclass=abc_ABCMeta):
                 @abc.abstractmethod
--- a/Lib/test/test_abc.py
+++ b/Lib/test/test_abc.py
@@ -149,8 +149,6 @@ def test_factory(abc_ABCMeta, abc_get_cache_token):
             self.assertEqual(D.foo(), 4)
             self.assertEqual(D().foo(), 4)
 
-        # TODO: RUSTPYTHON
-        @unittest.expectedFailure
         def test_object_new_with_one_abstractmethod(self):
             class C(metaclass=abc_ABCMeta):
                 @abc.abstractmethod
@@ -160,7 +158,7 @@ def test_factory(abc_ABCMeta, abc_get_cache_token):
             self.assertRaisesRegex(TypeError, msg, C)
 
         # TODO: RUSTPYTHON
-        @unittest.expectedFailure
+        ##@unittest.expectedFailure
         def test_object_new_with_many_abstractmethods(self):
             class C(metaclass=abc_ABCMeta):
                 @abc.abstractmethod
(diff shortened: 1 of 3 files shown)


youknowone:
oh, `test_dataclasses` seems to be related the message changes too

NakanoMiku39:
> oh, `test_dataclasses` seems to be related the message changes too

The test_dataclasses is now a test folder instead of a single test file, and in what way should it be updated?

youknowone:
We have 2 choices.
1. Update test_dataclasses first and finish this PR later. The existing single file will be replaced by folder with this patch.
2. Mark `@unittest.expectedFalure` to current dataclasses test. Then they will be fixed when we update the test later.

NakanoMiku39:
> We have 2 choices.
> 
> 1. Update test_dataclasses first and finish this PR later. The existing single file will be replaced by folder with this patch.
> 2. Mark `@unittest.expectedFalure` to current dataclasses test. Then they will be fixed when we update the test later.

The test files in the folder seems to be unrelated with the test_dataclasses.py though

youknowone:
I checked cpython repository. `test_dataclasses.py` looks still a file in Python 3.12. https://github.com/python/cpython/blob/v3.12.0/Lib/test/test_dataclasses.py
Could you share what's the folder?

NakanoMiku39:
> I checked cpython repository. `test_dataclasses.py` looks still a file in Python 3.12. https://github.com/python/cpython/blob/v3.12.0/Lib/test/test_dataclasses.py Could you share what's the folder?

Oh I was checking out the 3.12 branch [https://github.com/python/cpython/tree/3.12/Lib/test/test_dataclasses](url)

youknowone:
Got it. I think the test is moved to https://github.com/python/cpython/blob/3.12/Lib/test/test_dataclasses/__init__.py

If you run test through `-m test`, both file and folder will work in same way.
`cargo run  -- -m test test_dataclasses`

NakanoMiku39:
> Got it. I think the test is moved to https://github.com/python/cpython/blob/3.12/Lib/test/test_dataclasses/__init__.py
> 
> If you run test through `-m test`, both file and folder will work in same way. `cargo run -- -m test test_dataclasses`

Oh, but we are still following the main branch right? Then I will update the test file and ignore the folder

youknowone:
I prefer to follow recent distribution's tag. I'd like recommend v3.12.1 tag at this time. https://github.com/python/cpython/tree/v3.12.1

NakanoMiku39:
> I prefer to follow recent distribution's tag. I'd like recommend v3.12.1 tag at this time. https://github.com/python/cpython/tree/v3.12.1

I found that test_dataclasses needs support from dataclasses.py, which the latter would went into the `match` statement support issue

youknowone:
Marking test_dataclasses expectedFailure seems reasonable in that case.

NakanoMiku39:
> Marking test_dataclasses expectedFailure seems reasonable in that case.

No it's not working. Things get worse when comment out the parts cuz it is not in a test function
image
image

youknowone:
> Marking test_dataclasses expectedFailure seems reasonable in that case.

I made this temporary change here: https://github.com/RustPython/RustPython/pull/5140/files#diff-f0436842533448f8f02a19081d7c9cc8372b6685756caf32c83dfecebbee68e7R3679

youknowone (approved):
Thank you!

PR #5147, opened 2024-01-02
Use RPITIT where applicable

coolreader18:
Woohoo Rust 1.75!! There's also a slight change to `static_func()` again, but that's sorta because of the change for RPITIT.

### DIFF
--- a/vm/src/function/builtin.rs
+++ b/vm/src/function/builtin.rs
@@ -7,7 +7,14 @@ use std::marker::PhantomData;
 
 /// A built-in Python function.
 // PyCFunction in CPython
-pub type PyNativeFn = py_dyn_fn!(dyn Fn(&VirtualMachine, FuncArgs) -> PyResult);
+pub trait PyNativeFn:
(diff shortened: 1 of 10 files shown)


vm/src/function/builtin.rs
```diff
@@ -7,7 +7,14 @@ use std::marker::PhantomData;
 
 /// A built-in Python function.
 // PyCFunction in CPython
-pub type PyNativeFn = py_dyn_fn!(dyn Fn(&VirtualMachine, FuncArgs) -> PyResult);
+pub trait PyNativeFn:
+    Fn(&VirtualMachine, FuncArgs) -> PyResult + PyThreadingConstraint + 'static
+{
+}
+impl<F: Fn(&VirtualMachine, FuncArgs) -> PyResult + PyThreadingConstraint + 'static> PyNativeFn
+    for F
```
youknowone, line 15: interesting

youknowone (approved):

PR #5150, opened 2024-01-20
test/os: ns_to_sec rounding

deantvv:
In cpython, they use `_PyTime_ROUND_FLOOR` to read time. But in RustPython, we use `[Duration::from_secs_f64](https://doc.rust-lang.org/std/time/struct.Duration.html#method.try_from_secs_f64)` to read time.

Therefore, RustPython isn't affected by the rounding issue in the way that cpython does. We can safely ignore the `0.5*1e-9` bit in `ns_to_sec` function.

Or should RustPython implement cpython `_PyTime_round_t` stuff to make RustPython does exactly as cpython does?


youknowone:
In principle, it's good for RustPython to have behavior that is compatible with CPython. But this part looking like not worth to do it. is it?

youknowone (approved):
👍

Issue #5138, opened 2023-12-28
Complex division by zero is not handled


youknowone:
## Summary

Test introduced from: https://github.com/RustPython/RustPython/pull/5127/files#diff-3d5c8c4c000b7f6f9cdbedd36358b5f61686d39ff942be840e81f36c81da0ce2R248

## Expected

```
>>> pow(0j, -1000)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
ZeroDivisionError: 0.0 to a negative or complex power
>>> 
```

## Actual

```
>>>>> pow(0j, -1000)
0j
```

## Note

Related RustPython code is https://github.com/RustPython/RustPython/blob/main/vm/src/builtins/complex.rs#L104-L122

kennysliding:
Can I pick this up?

youknowone:
Sure, thank you for joining!

youknowone:
closed by #5143

PR #5143, opened 2023-12-28
Raise error on power with negative number

kennysliding:
Addressing issue #5138, this PR adds another check on the `inner_pow` function to raise error when the real number part of power number is negative

PR #5151, opened 2024-01-21
power complex division by zero

alinioan:
Fixed case where pow for complex numbers doesn't result in division by zero. (https://github.com/RustPython/RustPython/issues/5138)
pow(0j, -1) now results in a division by zero error.

Added unit test for this condition.

### DIFF
--- a/Lib/test/test_complex.py
+++ b/Lib/test/test_complex.py
@@ -241,6 +241,13 @@ class ComplexTest(unittest.TestCase):
         else:
             self.fail("should fail 0.0 to negative or complex power")
 
+        try:
+            0j ** (-1)
+        except ZeroDivisionError:
+            pass
+        else:
+            self.fail("should fail 0.0 to negative or complex power")
+
         # The following is used to exercise certain code paths
         self.assertEqual(a ** 105, a ** 105)
         self.assertEqual(a ** -105, a ** -105)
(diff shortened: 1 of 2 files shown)


Lib/test/test_complex.py
```diff
@@ -241,6 +241,13 @@ def test_pow(self):
         else:
             self.fail("should fail 0.0 to negative or complex power")
 
+        try:
+            0j ** (-1)
+        except ZeroDivisionError:
+            pass
+        else:
+            self.fail("should fail 0.0 to negative or complex power")
```
youknowone, line 249: Could you revert this change? Since the test file is originated from CPython project, we usually don't edit it ourselves.

vm/src/builtins/complex.rs
```diff
@@ -103,7 +103,7 @@ fn inner_div(v1: Complex64, v2: Complex64, vm: &VirtualMachine) -> PyResult<Comp
 
 fn inner_pow(v1: Complex64, v2: Complex64, vm: &VirtualMachine) -> PyResult<Complex64> {
     if v1.is_zero() {
-        return if v2.im != 0.0 {
+        return if v2.im != 0.0 || (v1.im.is_zero() && v2.re().is_negative()) {
```
youknowone, line 106: The patch seems not to be based on the recent upstream.

#5143 has similar changes

youknowone (changes requested):

vm/src/builtins/complex.rs
alinioan: Yes, sorry about this. When i first fixed this error it was from [this issue](https://github.com/RustPython/RustPython/issues/5138) which wasn't closed.

PR #5158, opened 2024-02-07
Fix integer rounding

dchiquito:
The [docs](https://docs.python.org/3/library/functions.html#round) specify that calling `round` with a negative precision removes significant digits, so that `round(12345, -2) == 12300`. The implementation was simply returning the original integer.

Additionally, `round(a, b)` is implemented as `(a / 10^b) * 10^b`, using half-even rounding during the division.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,14 +22,15 @@ checksum = "aae1277d39aeec15cb388266ecc24b11c80469deae6067e17a1a7aa9e5c1f234"
 
 [[package]]
 name = "ahash"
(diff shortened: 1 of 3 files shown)


dchiquito:
It seems to me like CI is failing on Windows when trying to link some OpenSSL function. I don't have a Windows machine available so I can't really do any investigation, but it doesn't seem related to this PR.

youknowone (approved):
Windows SSL is recently an endless pain point. 😵

PR #5160, opened 2024-02-09
Fix bug in `binascii` `uu` encoding. Pass more related unit tests.

kingiler:
- Fix logic bug in both `binascii.a2b_uu` and `binascii.b2a_uu`.
- Use correct error type by `new_binascii_error` instead of `new_value_error`.
- Pass three more unit tests: `test_uu`, `test_returned_value` and `test_unicode_a2b`.

[Edit]
- Pass more related unit tests in `test_uu` and `test_codecs`.

### DIFF
--- a/Lib/test/test_binascii.py
+++ b/Lib/test/test_binascii.py
@@ -38,8 +38,6 @@ class BinASCIITest(unittest.TestCase):
             self.assertTrue(hasattr(getattr(binascii, name), '__call__'))
             self.assertRaises(TypeError, getattr(binascii, name))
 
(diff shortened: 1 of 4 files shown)


kingiler:
About CI:
It seems related to compilation of openssl on Windows but I do not have Windows machine accessible. Can anyone help to lookup the issue?

youknowone (approved):
CI failure does not seem to related to this patch. I need to investigate what's happened, but currently not accessible to a windows machine either. Let's ignore it for now.

Thank you for contributing!

PR #5163, opened 2024-02-09
Fix benchmarks by replacing rust-cpython with pyo3

### LINKED ISSUE #3498: Benchmarks use `PyParser_ASTFromStringObject`, which is removed from Python 3.10
Benchmarks fail to compile on a system with Python 3.10, in my case, Fedora 35.

```
[dalley@localhost RustPython]$ cargo bench
   Compiling rustpython v0.1.2 (/home/dalley/Devel/rustpython-dev/RustPython)
error[E0425]: cannot find function, tuple struct or tuple variant `PyParser_ASTFromStringObject` in crate `python3_sys`
  --> benches/execution.rs:82:32
   |
82 |         let ret = python3_sys::PyParser_ASTFromStringObject(
   |                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `python3_sys`

For more information about this error, try `rustc --explain E0425`.
error: could not compile `rustpython` due to previous error
```

https://github.com/python/cpython/pull/25001

https://github.com/dgrunwald/rust-cpython/commit/7bcf2d305851b5bd22f20c070e0dd75e71ebd388#diff-3a5720e55119ee77dd2db4a0e191e513099d227760a8c40891a9f78872b46982R91

dchiquito:
The benchmarks seem to have been broken since Python 3.10 deprecated the API
they were using to parse and execute CPython baselines. Since then
rust-cpython has been deprecated in favor of pyo3.

- Replace `cpython` and `python3-sys` dependencies with `pyo3`.
- Add `html_report` feature to `criterion`, it will be required in a
  future release.
- Remove `anyhow`. It was unused and cargo cleaned it up automatically.
- Use `compile` and `exec` from within the CPython context to
compile and execute benchmarks. There's probably more overhead to that than
the internal API had, but that overhead should be consistent per
benchmark.

If anyone cares about hyperoptimizing benchmarks then they can optimize
the harness as well.

Fixes #3498

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -329,18 +329,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "cpython"
-version = "0.7.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3052106c29da7390237bc2310c1928335733b286287754ea85e6093d2495280e"
-dependencies = [
- "libc",
- "num-traits",
- "paste",
- "python3-sys",
-]
-
-[[package]]
 name = "cranelift"
 version = "0.88.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -977,6 +965,12 @@ dependencies = [
  "autocfg",
  "hashbrown",
 ]
+
+[[package]]
+name = "indoc"
+version = "2.0.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1e186cfbae8084e513daff4240b4797e342f988cecda4fb6c939150f96315fd8"
 
 [[package]]
 name = "insta"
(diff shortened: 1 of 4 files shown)


fanninpm:
> Use compile and exec from within the CPython context to
compile and execute benchmarks.  There's probably more overhead to that than the internal API had, but that overhead should be consistent per benchmark.

I have a minor question: How do we know that this won't inadvertently give RustPython an advantage? As far as I know, the API we used (from before CPython 3.10) ensured that compilation was kept separate from execution.

fanninpm:
Also, if I recall correctly, the RustPython benchmarks will eat up a lot of memory because RustPython presently has no garbage collection more advanced than simple reference-counting and needs circular references in order to set up simple things like `object` and `type`.

dchiquito:
When I originally tried to run the benchmarks they just failed to build because of features deprecated in Python 3.10, so my understanding was that they just haven't worked since the project moved past that point. My objective with this PR was to get something working so that I could benchmark some changes I was making.

From my cursory investigation, I couldn't find that pyo3 exposed a similar API to what was there previously. I think the difference between calling the CPython `exec` and hooking into whatever function it uses internally is small enough for now. If maintaining parity is important I can dig deeper.

I'm bummed about garbage collection too, but I don't see that it should block improving the benchmarking code.

fanninpm:
The reason I warned about GC is so that you're not surprised if the benchmarks crash due to your system not having enough RAM. (From what I recall, RustPython chews through about 16 GB of RAM.)

dchiquito:
Thanks for the warning! That is probably why I crashed my computer a few times testing this PR lol

youknowone (approved):
Thank you! it was broken for years and finally going to be fixed

PR #5165, opened 2024-02-10
Removed unused __bool__ methods

dchiquito:
Python does not define `list().__bool__`, `dict().__bool__`, and `str().__bool__`, and some tests were failing because they were defined.
I could not find any references to them and deleting them doesn't seem to break anything.


youknowone (approved):
Nice catch! We needed those implementations as fallback before correctly implemented `PySequenceMethods`, but not anymore as you pointed out.
Thank you for contributing!

Issue #5166, opened 2024-02-10
Fix Windows CI


youknowone:
Failing due to openssl issue

derlih:
Looks like this link problem came from OpenSSL version update (3.2.0 ->3.2.1) in Chocolatey

derlih:
This issue can be closed. Now Windows build is green.

youknowone:
closed by #5168, thank you!

PR #5168, opened 2024-02-16
Fix Windows CI

derlih:
Attempt to address #5166

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -125,8 +125,14 @@ jobs:
           components: clippy
       - name: Set up the Windows environment
         run: |
-          cargo install cargo-vcpkg
-          cargo vcpkg build
+          vcpkg install openssl
+          # choco install llvm --no-progress
+          # choco install openssl --version=3.2.0 --no-progress
+          # if [[ -d "C:\Program Files\OpenSSL-Win64" ]]; then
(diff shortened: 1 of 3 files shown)


youknowone:
Thank you so much! miri failing doesn't like related.

derlih:
Let me polish it a bit. =)

derlih:
@youknowone  can you clean up the [build cache](https://github.com/RustPython/RustPython/actions/caches) from this job?
It doesn't contain SSL build and cargo-vcpkg, but I can't invalidate it in an easy way.

youknowone:
Triggered running again after deleting cache

youknowone:
The result looks fine

derlih:
But not perfect =)
It uses cached `cargo-vcpkg` but still compiles OpenSSL from sources.
It takes extra 5 minutes of the Windows build. I though `vcpkg` will use cached OpenSSL build.

youknowone (approved):
Thank you so much. Working CI is always better than broken one, so let's merge it first.

How about additonally adding vcpkg directory to `actions/cache`? would it fix it?

derlih:
The problem is that `cargo-vcpkg`, `vcpkg` itself and `OpenSSL` build are cached by rust-cache action. But I guess after the cache restoration the timestamps from files don't match and that's why it needs to build `OpenSSL` on each build.

youknowone:
Thank you for explanation

PR #5169, opened 2024-02-18
Update `calendar.py` and `test_calendar.py`  from CPython v3.12

Blues-star:
tests passed


Blues-star:
#5104

youknowone (approved):
Looks good, thank you!

PR #5175, opened 2024-02-20
Update `copy.py` and `test_copy.py` from CPython v3.12 

Blues-star:
```log
...........x.............................x..................................
Ran 76 tests in 0.223s

OK (expected failures=2)
```

#5104


youknowone (approved):
👍

PR #5176, opened 2024-02-20
【Need Update csv.rs】Update csv.py and test_csv.py from CPython v3.12 

Blues-star:
I've updated `csv.py` and `test_csv.py`, but version 3.12 has changed the import options, and `__version__` now needs to be imported from `_csv.c` (which is now `csv.rs`). It seems that `csv.rs` has not yet aligned with the behavior of `cpython` upstream. Should we consider updating `csv.rs` or postponing this commit? Due to this reason, I'm still unable to test `test_csv`.

### DIFF
--- a/Lib/test/test_csv.py
+++ b/Lib/test/test_csv.py
@@ -1,0 +1,1441 @@
+# Copyright (C) 2001,2002 Python Software Foundation
+# csv package unit tests
+
+import copy
+import sys
+import unittest
+from io import StringIO
+from tempfile import TemporaryFile
+import csv
+import gc
+import pickle
+from test import support
+from test.support import warnings_helper, import_helper, check_disallow_instantiation
+from itertools import permutations
+from textwrap import dedent
+from collections import OrderedDict
+
+
+class BadIterable:
+    def __iter__(self):
+        raise OSError
+
+
+class Test_Csv(unittest.TestCase):
+    """
+    Test the underlying C csv parser in ways that are not appropriate
+    from the high level interface. Further tests of this nature are done
+    in TestDialectRegistry.
+    """
+    def _test_arg_valid(self, ctor, arg):
+        self.assertRaises(TypeError, ctor)
+        self.assertRaises(TypeError, ctor, None)
+        self.assertRaises(TypeError, ctor, arg, bad_attr = 0)
+        self.assertRaises(TypeError, ctor, arg, delimiter = 0)
+        self.assertRaises(TypeError, ctor, arg, delimiter = 'XX')
+        self.assertRaises(csv.Error, ctor, arg, 'foo')
+        self.assertRaises(TypeError, ctor, arg, delimiter=None)
+        self.assertRaises(TypeError, ctor, arg, delimiter=1)
+        self.assertRaises(TypeError, ctor, arg, quotechar=1)
+        self.assertRaises(TypeError, ctor, arg, lineterminator=None)
+        self.assertRaises(TypeError, ctor, arg, lineterminator=1)
+        self.assertRaises(TypeError, ctor, arg, quoting=None)
+        self.assertRaises(TypeError, ctor, arg,
+                          quoting=csv.QUOTE_ALL, quotechar='')
+        self.assertRaises(TypeError, ctor, arg,
+                          quoting=csv.QUOTE_ALL, quotechar=None)
+        self.assertRaises(TypeError, ctor, arg,
+                          quoting=csv.QUOTE_NONE, quotechar='')
+
+    def test_reader_arg_valid(self):
+        self._test_arg_valid(csv.reader, [])
+        self.assertRaises(OSError, csv.reader, BadIterable())
+
+    def test_writer_arg_valid(self):
+        self._test_arg_valid(csv.writer, StringIO())
+        class BadWriter:
+            @property
+            def write(self):
+                raise OSError
+        self.assertRaises(OSError, csv.writer, BadWriter())
+
+    def _test_default_attrs(self, ctor, *args):
+        obj = ctor(*args)
+        # Check defaults
+        self.assertEqual(obj.dialect.delimiter, ',')
+        self.assertIs(obj.dialect.doublequote, True)
+        self.assertEqual(obj.dialect.escapechar, None)
+        self.assertEqual(obj.dialect.lineterminator, "\r\n")
+        self.assertEqual(obj.dialect.quotechar, '"')
+        self.assertEqual(obj.dialect.quoting, csv.QUOTE_MINIMAL)
+        self.assertIs(obj.dialect.skipinitialspace, False)
+        self.assertIs(obj.dialect.strict, False)
+        # Try deleting or changing attributes (they are read-only)
+        self.assertRaises(AttributeError, delattr, obj.dialect, 'delimiter')
+        self.assertRaises(AttributeError, setattr, obj.dialect, 'delimiter', ':')
+        self.assertRaises(AttributeError, delattr, obj.dialect, 'quoting')
+        self.assertRaises(AttributeError, setattr, obj.dialect,
+                          'quoting', None)
+
+    def test_reader_attrs(self):
+        self._test_default_attrs(csv.reader, [])
+
+    def test_writer_attrs(self):
+        self._test_default_attrs(csv.writer, StringIO())
+
+    def _test_kw_attrs(self, ctor, *args):
+        # Now try with alternate options
+        kwargs = dict(delimiter=':', doublequote=False, escapechar='\\',
+                      lineterminator='\r', quotechar='*',
+                      quoting=csv.QUOTE_NONE, skipinitialspace=True,
+                      strict=True)
+        obj = ctor(*args, **kwargs)
+        self.assertEqual(obj.dialect.delimiter, ':')
+        self.assertIs(obj.dialect.doublequote, False)
+        self.assertEqual(obj.dialect.escapechar, '\\')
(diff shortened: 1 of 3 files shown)


Blues-star:
Does the current csv implementation not incorporate the dialect feature?

youknowone:
I don't remember the exact state. Since test_csv didn't exist, I guess our csv implementation was very immature.

Blues-star:
I have done some work to make the csv module compatible with test_csv.py, but due to my unfamiliarity with the API, the current version contains too much template code and redundant operations. It looks like the implementation is a bit messy. I will try to optimize the code and submit a new commit. I will close the current commit because I have a few more commits to make. Here are the test results.
```log
# cargo run -- -m unittest -v Lib.test.test_csv
Ran 123 tests in 2.669s

FAILED (failures=4, errors=10, skipped=6, expected failures=11)
```

Blues-star:
Currently, I have adapted most of the behaviors specified by the Python CSV module, but there are still 18 test cases failing, which I have marked as 'failed.' The functionality of the module should be working as expected at the moment. By the way, I have used modules from std in the CSV module, such as Hashmap and Mutex, and initialized using Onecell. I'm not sure if this is feasible, or if I should switch to using PyMutex instead?

Blues-star:
@youknowone have a review？

stdlib/src/csv.rs
```diff
@@ -32,21 +43,381 @@ mod _csv {
...
+        }
+        #[pygetset]
+        fn skipinitialspace(&self, vm: &VirtualMachine) -> PyRef<PyInt> {
+            vm.ctx.new_bool(self.skipinitialspace).to_owned()
+        }
+        #[pygetset]
+        fn lineterminator(&self, vm: &VirtualMachine) -> PyRef<PyStr> {
+            match self.lineterminator {
+                Terminator::CRLF => vm.ctx.intern_str("\r\n".to_string()).to_owned(),
+                Terminator::Any(t) => vm.ctx.intern_str(format!("{}", t as char)).to_owned(),
```
youknowone, line 103: is this `t` static value or dynamic value?

`intern_str` [interns](https://en.wikipedia.org/wiki/String_interning) static strings. If this is a dynamic value, 

```suggestion
                Terminator::Any(t) => vm.ctx.new_str(format!("{}", t as char)).to_owned(),
```

stdlib/src/csv.rs
```diff
@@ -32,21 +43,381 @@ mod _csv {
...
+            }
+        })
+    }
+    fn prase_lineterminator_from_obj(
+        vm: &VirtualMachine,
+        obj: &PyObjectRef,
+    ) -> PyResult<Terminator> {
+        match_class!(match obj.get_attr("lineterminator", vm)? {
+            s @ PyStr => {
+                Ok(if s.as_str().bytes().eq(b"\r\n".iter().copied()) {
```
youknowone, line 213:
```suggestion
                Ok(if s.as_str().as_bytes().eq(b"\r\n") {
```

stdlib/src/csv.rs
```diff
@@ -32,21 +43,381 @@ mod _csv {
...
+                    csv_core::Terminator::CRLF
+                } else if let Some(t) = s.as_str().bytes().next() {
```
youknowone, line 215:
```suggestion
                } else if let Some(t) = s.as_str().as_bytes().first() {
```
Does this check require to ensure s.len() == 1?

stdlib/src/csv.rs
```diff
@@ -176,27 +1004,55 @@ mod _csv {
...
+                    format!(
                 "iterator should return strings, not {} (the file should be opened in text mode)",
                 obj.class().name()
-            ))
+            ),
+                )
             })?;
             let input = string.as_str().as_bytes();
-
+            if input.is_empty() || input.starts_with(&[b'\n']) {
```
youknowone, line 1016:
```suggestion
            if input.is_empty() || input.starts_with(b"\n") {
```

youknowone (changes requested):
Thank you for many efforts to add a new test and implementing a many features in rust native module!

> I'm not sure if this is feasible, or if I should switch to using PyMutex instead?
PyMutex is single/multi thread support helpers. You can easilty turns parking_lot::Mutex to PyMutex. (but not std Mutex) I think `PyMutex` fits here. Otherwise we prefer to use `parking_lot::PyMutex` rather than std Mutex

The most of changes looks really good, but I have a bit of concerns about using `intern_str`. Could you check if they are intended to be actually interned or not?

stdlib/src/csv.rs
Blues-star: Haha, I didn't notice this API when I was looking through the API documentation. I was still wondering why there wasn't a basic API for creating dynamic strings.

stdlib/src/csv.rs
Blues-star: I think this modification may be unnecessary. When s.len()<1, this function will throw an error, propagating the error to the upper layer, which conforms to the expected behavior in Python.

stdlib/src/csv.rs
Blues-star: Thank you for your review

stdlib/src/csv.rs
Blues-star: Here I need to get ownership of the first character, and because it is of type u8 I believe that copying it is inexpensive.

Blues-star:
> Thank you for many efforts to add a new test and implementing a many features in rust native module!
> 
> > I'm not sure if this is feasible, or if I should switch to using PyMutex instead?
> > PyMutex is single/multi thread support helpers. You can easilty turns parking_lot::Mutex to PyMutex. (but not std Mutex) I think `PyMutex` fits here. Otherwise we prefer to use `parking_lot::PyMutex` rather than std Mutex
> 
> The most of changes looks really good, but I have a bit of concerns about using `intern_str`. Could you check if they are intended to be actually interned or not?

Haha, I didn't notice this API when I was looking through the API documentation. I was still wondering why there wasn't a basic API for creating strings. 

I have modified part of the code, keeping the old code for multiple type returns and operations on one of the bytes. 

Now I have another question: is `std::hashmap` feasible here, or do I need to replace it with another `no-std` library such as `hashbrown`?

stdlib/src/csv.rs
Blues-star: > 

done

Blues-star:
PyMutex doesn't seem to implement sync, so I switched back to parking_lot.

Blues-star:
@youknowone

stdlib/src/csv.rs
```diff
@@ -32,21 +43,381 @@ mod _csv {
...
+                    .intern_str(format!("{}", t as char))
+                    .to_owned()
+                    .into(),
+                None => vm.ctx.none(),
+            }
+        }
+        #[pygetset]
+        fn doublequote(&self, vm: &VirtualMachine) -> PyRef<PyInt> {
+            vm.ctx.new_bool(self.doublequote).to_owned()
+        }
```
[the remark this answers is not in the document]
Blues-star: done

youknowone:
> Now I have another question: is std::hashmap feasible here, or do I need to replace it with another no-std library such as hashbrown?

Currently we don't support no-std. So that is ok while preparing it is still a good idea.

stdlib/src/csv.rs
youknowone: Oh, I am sorry to make confusion. The suggestion and question were not related. I thought creating an iterator was not necessary here (for suggestion), and worried if it requires to raise error when s.len() > 1 but it is missed (comment).

Blues-star: > Oh, I am sorry to make confusion. The suggestion and question were not related. I thought creating an iterator was not necessary here (for suggestion), and worried if it requires to raise error when s.len() > 1 but it is missed (comment). `*t` will be a copy for Copy types.

Oh, I see. Due to limitations in the current implementation within csv_core, the support for multiple characters in lineterminator is not complete. 

Therefore, I intentionally ignored the case where s.len() > 1. I will add comments to explain this and thank you for your suggestion. I have already optimized the iterator part.

Blues-star:
Are there any requested changes that I need to address?

youknowone (approved):
Thank you, I attached a commit with minor fixes

Lib/test/test_csv.py
```diff
@@ -0,0 +1,1469 @@
...
+    def compare_dialect_123(self, expected, *writeargs, **kwwriteargs):
+
+        with TemporaryFile("w+", newline='', encoding="utf-8") as fileobj:
+
+            writer = csv.writer(fileobj, *writeargs, **kwwriteargs)
+            writer.writerow([1,2,3])
+            fileobj.seek(0)
+            self.assertEqual(fileobj.read(), expected)
+
+    def test_dialect_apply(self):
```
youknowone, line 533:
```suggestion
    # TODO: RUSTPYTHON
    @unittest.expectedFailure
    def test_dialect_apply(self):
```
Rather than directly commenting out inner test code, we prefer to keep test expectedFailure in general - not when allowing partial test gives significant benefits.

Blues-star: Okay, I will check it out.

Blues-star:
> Thank you, I attached a commit with minor fixes

I accidentally made changes to `csv.py` earlier, but now I have reverted it back to the version for `Python 3.12`.

youknowone (approved):

youknowone:
Thank you so much!

youknowone:
We probably chose it as harvesting low-hanging fruit. The first csv module writers might not know about those limitation. Any reasonable suggestion will be appreciated.

Issue #5179, opened 2024-02-24
Add `__buffer__` and `__release_buffer__` methods to builtin types


dchiquito:
## Feature

The buffer protocol is mostly intended to provide provide a C API to access the underlying buffers of data structures like `bytes`, `array`, and `memoryview`. In Python 3.12 two new methods, `__buffer__` and `__release_buffer__`, were added to allow subclasses and custom classes to interface with the C API.

As far as I can tell RustPython does not support the C API, but the appropriate builtin types should still have these slots defined.

## Python Documentation or reference to CPython source code

https://docs.python.org/3.12/reference/datamodel.html#emulating-buffer-types
https://peps.python.org/pep-0688/

PR #5180, opened 2024-02-24
Update `_collections_abc.py` and `test_collections.py` to 3.12.2

dchiquito:
Update `_collections_abc.py` and `test_collections.py` to 3.12.2

`test_Buffer` is disabled pending #5179

This should check off `_collections_abc.py` and `test/test_ordered_dict.py`. from #5104. I'm unsure what else would be required to check off `collections`.


youknowone (approved):
Thank you!

PR #5187, opened 2024-03-08
fix some typos

wellweek:


Lib/test/test_unpack.py
```diff
@@ -162,7 +162,7 @@ def test_extended_oparg_not_ignored(self):
         ns = {}
         exec(code, ns)
         unpack_400 = ns["unpack_400"]
-        # Warm up the the function for quickening (PEP 659)
+        # Warm up the function for quickening (PEP 659)
```
youknowone, line 165: This file is originated from CPython repository. Updating this file from CPython repository is preferred.
The file is fixed in CPython repository.

See also #5104

benches/benchmarks/pystone.py
```diff
@@ -16,7 +16,7 @@
 
 Version History:
 
-                Inofficial version 1.1.1 by Chris Arndt:
```
youknowone, line 19: Originated from here: https://gist.github.com/SpotlightKid/3bd430a75b84a08e54f9

youknowone (approved):

Issue #5190, opened 2024-03-10
Fix failing unit tests


dchiquito:
## Summary
There are a number of tests currently failing on `main`:
* `test_unittest`
* `test_urllib`
* `test_userdict`
* `test_uuid`
* `test_webbrowser`
* `test_xmlrpc`
* `test_yield_from`
* `test_zipfile`

 When these tests are run individually, they pass, which is why I assume they were not fixed sooner. When the entire test suite is run, they fail. Some experimentation reveals that there are a small number of apparently unrelated tests which "poison" these failing tests:
* `test_import` and `test_importlib`
* `test_with`
* `test_typing`
* `test_mmap` and `test_os`

Some failing tests are impacted by multiple "poisonous" tests.

### `test_import` and `test_importlib`
These tests cause `test_unittest` to fail. If either is omitted or the order is changed, it will not fail:
```bash
RUSTPYTHONPATH=Lib cargo run --release -- Lib/test test_import test_importlib test_unittest
```
Stack trace:
```
test test_unittest failed -- Traceback (most recent call last):
  File "/home/daniel/git/RustPython/pylib/Lib/unittest/test/test_discovery.py", line 836, in test_discovery_failed_discovery
    with unittest.mock.patch('builtins.__import__', _import):
  File "/home/daniel/git/RustPython/pylib/Lib/unittest/test/test_discovery.py", line 840, in test_discovery_failed_discovery
    with test.test_importlib.util.uncache('package'):
AttributeError: module 'test.test_importlib' has no attribute 'util'
```

### `test_with`
`test_with` causes `test_unittest`, `test_xmlrpc`, and `test_yield_from` to fail:
```bash
RUSTPYTHONPATH=Lib cargo run --release -- Lib/test test_with test_unittest test_xmlrpc test_yield_from 
```
Stack trace for `test_unittest`:
```
test test_unittest failed -- Traceback (most recent call last):
  File "/home/daniel/git/RustPython/pylib/Lib/test/test_with.py", line 546, in failAsBool
    self.fail("Should NOT see this")
AssertionError: Should NOT see this

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/daniel/git/RustPython/pylib/Lib/unittest/test/test_result.py", line 275, in test_addFailure_filter_traceback_frames_context
    self.assertEqual(len(dropped), 1)
AssertionError: 0 != 1

test_unittest failed
```
Stack trace for `test_xmlrpc`:
```
[...]

PR #5191, opened 2024-03-10
Fix convenient failing tests

dchiquito:
Fix some easy test failures.

There are more failing tests, but they don't fail when run individually, and so are much harder to debug. See #5190


youknowone (approved):
Nice catch! Thank you so much!

PR #5192, opened 2024-03-11
Properly unload modules between tests

dchiquito:
There seems to have been a bug in the libregrtest code which unloaded modules between tests. The previous state was calculated using `sys.modules.keys()`, which is actually a mutable object that is updated as the underlying `sys.modules` is updated. The result was that modules were not unloaded between tests, which is the root cause for `test_unittest` failing when run after `test_import` and `test_importlib`.

This code is copied from 3.12. Ideally all of `libregrtest` should probably be updated as it seems wildly out of date, but that's a lot more work.

### DIFF
--- a/Lib/test/libregrtest/main.py
+++ b/Lib/test/libregrtest/main.py
@@ -373,7 +373,7 @@ class Regrtest:
             import trace
             self.tracer = trace.Trace(trace=False, count=True)
 
-        save_modules = sys.modules.keys()
+        save_modules = set(sys.modules)
 
         print("Run tests sequentially")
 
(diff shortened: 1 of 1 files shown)


youknowone:
Thank you!

Since the library is originated from https://github.com/python/cpython/tree/main/Lib/test/libregrtest

I have questions.

- Would updating it to recent version fix the bug? Then updating it will be better than making original fix. It will be easily lost when we update the library from CPython.
- Is it a still bug in cpython version too even if we update it? Then submitting the patch also to CPython project will be highly recommended to prevent regression.
- Is this RustPython-specific bug? Then leaving comment starting with `# XXX: RustPython; <note>` will be helpful to keep it when updating the library.

dchiquito:
* I briefly looked in to updating libregrtest, but it seemed like a bit of a rabbit hole. The latest version uses `match` statements which are currently unsupported, and also some as yet unimplemented methods on `TextIOWrapper` IIRC. There are probably more dependencies that would also need to be updated before libregrtest, but I gave up at that point. This fix copies only the relevant lines from the updated libregrtest, since updating the whole thing is a lot more work.
* I'm not sure of the history of the CPython code. I assume that either this was a CPython bug, or that in some past version of CPython `dict.keys()` returned an immutable object. I copied the changes here directly from 3.12, so CPython is definitely aware of the situation and has already made the requisite change.
* The "bug" being fixed here is that libregrtest is out of date and has slight incompatibilities with the current Python implementation. The current version is quite different, so I imagine someone will need to basically wipe out what we have and replace it with the up to date version. I don't see that a comment would be relevant or helpful for that, but I can add one if you think it would.

Ultimately, this is just a stopgap to get all the tests passing. Personally I find that having to filter out weird test results is quite disruptive when running unit tests frequently. Even if there is a better solution, it's a lot more work. IMHO having accurate test results justifies a bit of duct tape. If you'd rather wait for the complete libregrest update I can close this PR and leave a note on #5190.

youknowone (approved):
Thank you so much for detailed explanation! I understand how it is going. No worry if it is already fixed in CPython side. This change will be replaced by the new code, and both the code will be correct code anyway!

PR #5195, opened 2024-03-12
Update ``Lib/pprint.py`` and ``Lib/test/test_pprint.py`` to 3.12

Eclips4:
This PR for a ``pprint.py`` module
* ``pprint.py`` needs an update
* ``test_pprint.py`` needs an update


youknowone (approved):
👍

PR #5197, opened 2024-03-13
Update ``Lib/cgitb.py`` to 3.12

Eclips4:
This PR for a ``cgitb.py`` module

* ``cgitb.py`` needs an update
* ``test_cgitb.py`` doesn't need to be updated


youknowone (approved):
👍

PR #5199, opened 2024-03-13
Update ``Lib/compileall.py`` and ``Lib/test/test_compileall.py``

Eclips4:
This PR for a ``compileall.py`` module

* ``Lib/compileall.py`` needs an update
* ``Lib/test/test_compileall.py`` needs to be created :)


Eclips4:
Seems that CI/CD is failing. Let me check what's wrong

youknowone:
maybe there is few test causing hanging

Eclips4:
> maybe there is few test causing hanging

Now everything should be fine.

youknowone:
That will be mostly because of poor RustPython implementation for multiprocessing. Finding candidates and mark them as `@unittest.skip` will be totally fine

PR #5204, opened 2024-03-20
Skip TestScandir.test_uninstantiable

dchiquito:
This test was marked as an expected failure. Because the garbage collector is missing, that meant that the `os.scandir` object went unclosed. This object was squatting on the file descriptors of all the files contained in the test directory, which was breaking test_zipfile.

### DIFF
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -4249,7 +4249,8 @@ class TestScandir(unittest.TestCase):
             self.assertEqual(stat1, stat2)
 
     # TODO: RUSTPPYTHON (AssertionError: TypeError not raised by ScandirIter)
-    @unittest.expectedFailure
+    # TODO: See https://github.com/RustPython/RustPython/issues/5190 for skip rationale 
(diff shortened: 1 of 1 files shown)


youknowone (approved):
Thank you!

Lib/test/test_os.py
```diff
@@ -4249,7 +4249,8 @@ def assert_stat_equal(self, stat1, stat2, skip_fields):
             self.assertEqual(stat1, stat2)
 
     # TODO: RUSTPPYTHON (AssertionError: TypeError not raised by ScandirIter)
-    @unittest.expectedFailure
+    # TODO: See https://github.com/RustPython/RustPython/issues/5190 for skip rationale 
+    @unittest.skip("skipping to avoid the unclosed scandir from squatting on file descriptors")
```
fanninpm, line 4253: The current change will show up as "not skipped by us" on the website. I suggest this:

```suggestion
@unittest.skip("TODO: RUSTPYTHON, avoid the unclosed scandir from squatting on file descriptors")
```

PR #5206, opened 2024-03-21
Increase threshold for zipfile test_many_opens 

dchiquito:
See https://github.com/RustPython/RustPython/pull/5204#discussion_r1534119326
Thanks @fanninpm , did not know that

I wound up finding at least three other tests that were also leaving unclosed file descriptors lying around and causing this test to fail. Rather than continue hunting them all down, I think it's easier to just increase the threshold so that they don't impact this test.

I suspect this is not an issue for CPython because it has a working garbage collector, and it has very few tests that are expected to fail.

### DIFF
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -4248,9 +4248,8 @@ class TestScandir(unittest.TestCase):
         else:
             self.assertEqual(stat1, stat2)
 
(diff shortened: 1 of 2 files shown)


youknowone (approved):
Thank you! The change looks good. I wish we finally can revert file limit changes by implementing GC.
The CI fail mark attracted me not to look in it very well. That looks ok because the change doesn't affect the Rust code.

PR #5207, opened 2024-03-21
Skip typing test which causes other failures

dchiquito:
See https://github.com/RustPython/RustPython/issues/5190#issuecomment-2010535802

I suspect that upgrading typing to 3.12 will resolve this, but I am not sure, and it would be a lot of work. For now, disabling this test should fix the related test failures.


youknowone (approved):
👍

PR #5208, opened 2024-03-21
Implement new clippy lints

dchiquito:
clippy was just updated and has a few minor issues with the code base. This should get CI passing again.


youknowone (approved):
👍

PR #5209, opened 2024-03-28
Add itertools.batched Support

Poppro:
Adds the itertools.batched function new in python 3.12

https://docs.python.org/3/library/itertools.html#itertools.batched

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -1986,10 +1986,7 @@ mod decl {
                 return Err(vm.new_value_error("n must be at least one".to_owned()));
             }
             let n = n.to_usize().unwrap();
-            let iterable = match iterable_ref.get_iter(vm) {
-                Ok(it) => it,
-                Err(e) => return Err(e),
-            };
+            let iterable = iterable_ref.get_iter(vm)?;
 
             Self {
                 iterable,
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -1985,9 +1985,9 @@ mod decl {
             if n.lt(&BigInt::one()) {
                 return Err(vm.new_value_error("n must be at least one".to_owned()));
             }
-            let n = n
-                .to_usize()
-                .ok_or(vm.new_value_error("Python int too large to convert to usize".to_owned()))?;
+            let n = n.to_usize().ok_or(
+                vm.new_overflow_error("Python int too large to convert to usize".to_owned()),
+            )?;
             let iterable = iterable_ref.get_iter(vm)?;
 
             Self {
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -1985,7 +1985,9 @@ mod decl {
             if n.lt(&BigInt::one()) {
                 return Err(vm.new_value_error("n must be at least one".to_owned()));
             }
-            let n = n.to_usize().unwrap();
+            let n = n
+                .to_usize()
+                .ok_or(vm.new_value_error("Python int too large to convert to usize".to_owned()))?;
             let iterable = iterable_ref.get_iter(vm)?;
 
             Self {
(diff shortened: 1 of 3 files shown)


youknowone (commented):
Thank you so much! Could you also update test_itertools from CPython 3.12 to ensure this patch is compliant its spec? The related issue is #5104

Poppro:
@youknowone 
> Thank you so much! Could you also update test_itertools from CPython 3.12 to ensure this patch is compliant its spec? The related issue is #5104

I went ahead and added the 3.12 itertools_test update to this PR.

Notes:
  1. This caught an out-of-order exception throw in the batched implementation. Corrected now.
  2. I had to mark a number of previously passing tests as ignored, since they now expect DepreciatedWarnings to be thrown.

vm/src/stdlib/itertools.rs
```diff
@@ -1952,4 +1955,79 @@ mod decl {
...
+        ) -> PyResult {
+            let n = n.as_bigint();
+            if n.lt(&BigInt::one()) {
+                return Err(vm.new_value_error("n must be at least one".to_owned()));
+            }
+            let n = n.to_usize().unwrap();
+            let iterable = match iterable_ref.get_iter(vm) {
+                Ok(it) => it,
+                Err(e) => return Err(e),
+            };
```
youknowone, line 1992:
```suggestion
            let iterable = iterable_ref.get_iter(vm)?;
```

vm/src/stdlib/itertools.rs
```diff
@@ -1952,4 +1955,79 @@ mod decl {
...
+        fn py_new(
+            cls: PyTypeRef,
+            Self::Args { iterable_ref, n }: Self::Args,
+            vm: &VirtualMachine,
+        ) -> PyResult {
+            let n = n.as_bigint();
+            if n.lt(&BigInt::one()) {
+                return Err(vm.new_value_error("n must be at least one".to_owned()));
+            }
+            let n = n.to_usize().unwrap();
```
youknowone, line 1988: What happens if n is a negative number? That will be a python error rather than panic (by unwrap).
By the error type and message, you might want to change the type of `BatchedNewArgs::n` to `usize` or any unsigned type. Otherwise making an `ok_or_else` chain will be a good choice.

youknowone (changes requested):
That's a good point of tests 😄  Thank you!

vm/src/stdlib/itertools.rs
Poppro: Doesn't the conditional on line 1985 ensure that n will not be a number less than 1?

youknowone: That's right, thank you! How about n > usize::MAX?
Since unwrap is one of main source of uncontrolled panic, we prefer to unsure they don't have edge cases and a bit picky about it.
If that's logically not able to be triggered, using `expect()` with reasoning  rather than `unwrap()` will be very helpful to provide the reasoning and a debug hint for future regression.

Poppro: Great point! CPython throws the following error in such case: `OverflowError: Python int too large to convert to C ssize_t`.

I've modified the unwrap to instead use ok_or and return an overflow error in this case.

Poppro:
Build failures seem unrelated?

fanninpm:
> Build failures seem unrelated?

The macOS failure in `test_re` seems unrelated. However, can you please run `cargo fmt` to see if that fixes the other failure? (And, for good measure, consider running `cargo clippy`?)

Poppro:
> > Build failures seem unrelated?
> 
> The macOS failure in `test_re` seems unrelated. However, can you please run `cargo fmt` to see if that fixes the other failure? (And, for good measure, consider running `cargo clippy`?)

Screenshot 2024-04-05 at 6 47 57 PM

Here's the results - no changes or issues generated

youknowone (approved):
Thank you so much!

PR #5226, opened 2024-04-14
Fix `None.__ne__` bug

moreal:
## Description

This pull request is related with https://github.com/python/cpython/issues/112125, https://github.com/RustPython/RustPython/issues/5103, https://github.com/RustPython/RustPython/pull/5124.

In Python 3.12, `None.__ne__` behavior was changed but it was a bug. According to the GitHub issue, the bug was fixed in *3.12.1* version and deployed. When I run `None.__ne__(None)` in `3.12.2` CPython, it returns `False`.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -105,7 +105,7 @@ env:
     test_weakref
     test_yield_from
   # Python version targeted by the CI.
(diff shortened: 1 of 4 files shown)


youknowone (approved):
wow.. didn't know it was a bug.

PR #5227, opened 2024-04-14
Check `value` is `PyType` first

moreal:
This pull request continues from #5225 PR. In that PR, I missed to downcast `value` to `PyType` type when evaluating `both_module` value. So `cargo run -- --install-pip get-pip` was still failed in the same reason.

This pull request downcasts `value` to `PyRef<PyType>` when `value` is module type. Also it checks `value` is `PyType` first, once.

This pull request doesn't resolve an issue, *get-pip doesn't work*, completely. After this pull request, it should bump `Lib/shutil` to Python 3.12 version's.


youknowone (approved):
Thank you!

Issue #4964, opened 2023-05-11
_signal does not exist for wasm32


lastmjs:
When compiling to the `wasm32` architecture, `_signal` is conditionally not compiled. This will break any module that depends on `signal`, which is included when compiling to `wasm32`. I was able to polyfill this for now by adding my own custom `_signal.py` with the following:

```python
def signal():
    raise NotImplementedError

def getsignal():
    raise NotImplementedError
```

It would be great if `wasm32` could at least have a compatible `_signal`.

P.S. this was discovered while trying to get `uuid` to work for `wasm32-wasi`

PR #5282, opened 2024-04-25
Add const-only signals for wasm32

### LINKED ISSUE #5050: Missing `_signal` module on WASI
## Summary

If you try to import `platform` you get the following error:

```
>>>>> import platform
  File "platform", line 119, in <module>
  File "subprocess", line 49, in <module>
  File "signal", line 1, in <module>
ModuleNotFoundError: No module named '_signal'
```

It seems that the `_signal` module is missing. Any idea what this could be or how we could resolve this?

youknowone:
partially resolve #5050 #4964


d3lm:
Yay, thanks!

PR #5286, opened 2024-04-28
Implement `DISALLOW_INSTANTIATION` PyTypeFlags

Snowapril:
This revision implement `DISALLOW_INSTANTIATION` PyTypeFlags to prevent several type from instantiating which is not intended.


Snowapril:
#5324 must be solved for passing test in this PR.

PR #5300, opened 2024-05-06
contextvars

### LINKED ISSUE #3188: [RFC] Implement PEP 567 (`contextvars`)
## Summary

The `contextvars` module, from PEP 567, is necessary to upgrade `asyncio` (and `decimal`) to CPython 3.8. I started #2950 to implement `contextvars`, but the PR has since stalled. (@DimitrisJim urged me to open this issue.)

## Detailed Explanation and Rationale

[PEP 567](https://www.python.org/dev/peps/pep-0567/), implemented in [CPython 3.7](https://docs.python.org/3/whatsnew/3.7.html#contextvars), introduced the `contextvars` module and context variables, which are conceptually similar to thread-local variables. However, unlike thread-local variables, context variables correctly support asynchronous code. Following that implementation, the `asyncio` and `decimal` modules were re-written to take advantage of this new capability. As part of our endeavors to get as much of the codebase covered by the CPython test suite as possible (see #1671), and out of a desire support CPython 3.8 or later (see the README), I think it's necessary to add the `test_asyncio` module to the CPython test suite. The addition of the CPython 3.8 version of `test_asyncio` would benefit from upgrading the `asyncio` module itself to that same version. (@coolreader18 and I have noticed that test failures were often alleviated by upgrading the relevant Python module.) However, `asyncio` has a dependency — `contextvars` — that we don't have yet. Accordingly, I have started a draft PR (#2950) to implement this module here, but I got stymied when I was drafting the Rust implementation because I am unfamiliar with how RustPython deals with thread safety. For context, here are the parts of the C implementation that I think are relevant:
- https://github.com/python/cpython/blob/3.8/Include/context.h
- https://github.com/python/cpython/blob/3.8/Python/context.c
- https://github.com/python/cpython/blob/3.8/Python/clinic/context.c.h
- https://github.com/python/cpython/blob/3.8/Modules/_contextvarsmodule.c
- https://github.com/python/cpython/blob/3.8/Modules/clinic/_contextvarsmodule.c.h
- https://github.com/python/cpython/blob/3.8/Include/internal/pycore_context.h

## Unresolved Questions

- What is the meaning of `PyContext` (defined in [`pyobject.rs`](https://github.com/RustPython/RustPython/blob/main/vm/src/pyobject.rs)), and does it have any bearing on the scope of PEP 567?
- The scaffolding is present, but how should it be fleshed out?
- How would each `Context` object relate to a thread in RustPython?

youknowone:
close #3188


stdlib/src/contextvars.rs
```diff
@@ -1,49 +1,217 @@
...
+    };
+    use crossbeam_utils::atomic::AtomicCell;
+    use indexmap::IndexMap;
+    use once_cell::sync::Lazy;
+    use std::{
+        cell::{Cell, RefCell, UnsafeCell},
+        sync::atomic::Ordering,
     };
 
+    // TODO: Real hamt implementation
```
fanninpm, line 43: I'm trying to remember why they chose to go with a HAMT.

Issue #5252, opened 2024-04-22
Update typing to 3.12


dchiquito:
## Feature
The `typing` module needs to be updated to 3.12. It seems that it was last updated to 3.9, so it's missing several major versions worth of features, like [Unpack](https://github.com/RustPython/RustPython/issues/5185), generics, and type aliases using the `type` keyword.

Not having these typing features impedes the process of updating modules that use, which presumably will be most new stuff going forward.

I estimate that updating will require:
* Moving some abstract base classes from `typing.py` into a new native module `typing.rs`, specifically `TypeVar`, `ParamSpec`, `TypeVarTuple`, `ParamSpecArgs`, `ParamSpecKwards`, `TypeAliasType`, and `Generic`. 
* ~Adding new syntax like `type` type aliases and generics to the parser.~ It seems like this is already included in https://github.com/RustPython/Parser/blob/main/parser/src/python.lalrpop
* Adding new instructions which generate/consume type information to `compiler.rs` and `frame.rs`. For example, `def foo[T](x: T): ...` must create a new `TypeVar(T)` and inject it into the correct scope so that it is available as an argument annotation. This requires a new instruction defined in `bytecode.rs` which is emitted by `compiler.rs` and consumed by `frame.rs`.
* Copying `typing.py` and `test_typing.py` from the CPython code base.

## Python Documentation or reference to CPython source code
https://docs.python.org/3/library/typing.html

dchiquito:
I've been chewing on this for a bit, but it's pretty slow going since I'm also learning about a lot of compilation internals. As soon as I have something that is mostly compatible with the 3.12 `typing.py` I will submit a partial PR to get some incremental feedback.

youknowone:
That's awesome. Please feel free to ask anything if you stuck on something.

About typing module, if it helps, mixing Rust and Python implementation is limited but still possible. #5254 is doing it for ExceptionGroup.

hydrogen602:
I don't know if this is already known, but I just wanted to add that if not running in release, syntax like `def foo[T](): pass` causes a panic in the parser at `core/src/source_code.rs:139:9`

dchiquito:
That problem should be resolved as of #5319 / https://github.com/RustPython/Parser/pull/120. Thanks for reporting!

hbina:
Hi, what is the progress on this? Is there a way I can help? Trying to get started with RustPython myself :)

youknowone:
[...]

PR #5302, opened 2024-05-07
Add 3.12 typing features to the compiler

dchiquito:
This is the first step towards implementing 3.12 compliant typing(#5252 ) . It's taken me a long time to get this far so I wanted to check in before spending more time building on this foundation.

Changes made:
* Add the builtin `_typing` stdlib module. This will eventually replace the fundamental typing primitives that are currently implemented in `typing.py`. This is still very much a work in progress, but it works well enough to exercise the new compiler features.
* Compile and evaluate function generics (`def foo[X](x: X): ...`)
* Compile and evaluate class generics (`class C[X]: def __init__(x: X): ...`)
* Compile type alias statements (`type X = list[int]`)

Some questions:
* The `push_symbol_table` and `pop_symbol_table` are my attempt to simulate the behavior of the new [Annotation Scopes](https://docs.python.org/3/reference/executionmodel.html#annotation-scopes). I'm not sure if I got this 100% right.
* At some point, CPython introduced the compiler-level concept of [intrinsics](https://github.com/python/cpython/blob/v3.12.2/Python/intrinsics.c#L212). There are a handful of instructions like `CALL_INTRINSIC_1` and `CALL_INTRINSIC_2` that do weird fundamental python things like printing, importing, or defining types. It seems to me that this is mostly a way of organizing the code than a fundamental design requirement, so I ignored this and added the intrinsics I needed as instructions. Perhaps this is not correct?
* I made a small change to how the bookkeeping is done for unmarshalling bytecode. Adding a new instruction caused me some very odd intermittent errors that took me a while to debug, so my hope is to prevent headaches in the future.


vm/src/frame.rs
```diff
@@ -1659,6 +1699,14 @@ impl ExecutingFrame<'_> {
             vm.ctx.new_dict().into()
         };
 
+        let type_params: PyTupleRef = if flags.contains(bytecode::MakeFunctionFlags::TYPE_PARAMS) {
+            self.pop_value()
+                .downcast()
+                .map_err(|_| vm.new_type_error("Type params must be a tuple.".to_owned()))?
+        } else {
+            vm.ctx.new_tuple(vec![])
```
youknowone, line 1707:
```suggestion
            vm.ctx.empty_tuple.clone()
```
empty_tuple shares a singleton object

youknowone (commented):
Awesome! I like the changes.

> The push_symbol_table and pop_symbol_table are my attempt to simulate the behavior of the new [Annotation Scopes](https://docs.python.org/3/reference/executionmodel.html#annotation-scopes). I'm not sure if I got this 100% right.

To be honest, I don't know well about either Annotation Scopes or our symbol table. If you are not sure about the design, leaving a comment with your concern will be fine.
@coolreader18 Do you have any advice?

> At some point, CPython introduced the compiler-level concept of [intrinsics](https://github.com/python/cpython/blob/v3.12.2/Python/intrinsics.c#L212). There are a handful of instructions like CALL_INTRINSIC_1 and CALL_INTRINSIC_2 that do weird fundamental python things like printing, importing, or defining types. It seems to me that this is mostly a way of organizing the code than a fundamental design requirement, so I ignored this and added the intrinsics I needed as instructions. Perhaps this is not correct?

Yes, CPython is recently very quickly changing its instruction design. Our instructions design is staying around 3.6~3.8. Your approach seems correct and fits in traditional Python instruction design well. We will be able to introduce those new concepts later. Don't worry!

> I made a small change to how the bookkeeping is done for unmarshalling bytecode. Adding a new instruction caused me some very odd intermittent errors that took me a while to debug, so my hope is to prevent headaches in the future.

I am sorry to make you to be confused. How about placing a comment to the last instruction like "Please add instruction in the middle of the enum; otherwise `LAST_INSTRUCTION` need to be updated"

dchiquito:
Thanks for the feedback! Would you prefer keeping this branch separate until typing is finished or merging it now?

youknowone:
I prefer to merge small patches, frequently. I will review this patch soon, thank you so much!

youknowone (approved):
Awesome, finally we have `_typing`!

I also tried #4946 on it. `_typing` look like almost close to working.

Thank you for making a big progress 😄

PR #5303, opened 2024-05-08
Set magic methods slot wrapper only for heaptypes

youknowone:


youknowone:
included in a recent series of slots

PR #5304, opened 2024-05-08
Update audiotests from CPython 3.12.3

youknowone:
#4881 with stable version fix


youknowone:
Thanks @moonlightaria

PR #5320, opened 2024-05-13
Adding generator_stop to compile_future_features

### LINKED ISSUE #4060: allow `from __future__ import generator_stop`
## Feature

code:
```python
from __future__ import generator_stop
```

Expected: Nothing

Actual: SyntaxError: future feature generator_stop is not defined

## TODO

Add it to `compile_future_features` function

matheusfgoncalves95:
As well as some mild formatting. 

Fix #4060

### DIFF
--- a/compiler/codegen/src/compile.rs
+++ b/compiler/codegen/src/compile.rs
@@ -2903,15 +2903,8 @@ impl Compiler {
         for feature in features {
             match feature.name.as_str() {
                 // Python 3 features; we've already implemented them by default
-                "nested_scopes"
-                | "generators"
-                | "division"
-                | "absolute_import"
-                | "with_statement"
-                | "print_function"
-                | "unicode_literals"
-                | "compile_future_features" => {}
-                "generator_stop" => {}
+                "nested_scopes" | "generators" | "division" | "absolute_import"
+                | "with_statement" | "print_function" | "unicode_literals" | "generator_stop" => {}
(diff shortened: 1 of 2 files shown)


compiler/codegen/src/compile.rs
```diff
@@ -2903,9 +2903,15 @@ impl Compiler {
...
-                // "generator_stop" => {}
+                "nested_scopes"
+                | "generators"
+                | "division"
+                | "absolute_import"
+                | "with_statement"
+                | "print_function"
+                | "unicode_literals"
+                | "compile_future_features" => {}
+                "generator_stop" => {}
```
fanninpm, line 2914: Does this work?

```suggestion
                | "compile_future_features"
                | "generator_stop" => {}
```

matheusfgoncalves95: Oops, my autoformatter led to that silly mistake.

To answer your question: it seems to work.

I did a quick check on the Python interpreter and the import succeeds. The module also looks okay via `dir()` inspection.

Anyhow, I will fix it. Thanks for pointing that out.

youknowone (changes requested):
Thank you for contributing! Your patch also fixed a test in `test_contextlib`. Could you also update the test? Removing `@unittest.expectedFailure` will fix the CI failure

youknowone (approved):

PR #5324, opened 2024-05-15
Fix `best_base` to select proper base class

Snowapril:
This revision implements `itemsize` slots and fix `best_base` implementation to select proper base class when generating class.

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -1324,7 +1324,6 @@ fn calculate_meta_class(
     Ok(winner)
 }
 
-
 fn solid_base(typ: &PyTypeRef, vm: &VirtualMachine) -> PyTypeRef {
     let base = if let Some(base) = &typ.base {
         solid_base(base, vm)
@@ -1341,7 +1340,6 @@ fn solid_base(typ: &PyTypeRef, vm: &VirtualMachine) -> PyTypeRef {
 }
 
 fn best_base(bases: &[PyTypeRef], vm: &VirtualMachine) -> PyResult<PyTypeRef> {
-   
     let mut base: Option<PyTypeRef> = None;
     let mut winner: Option<PyTypeRef> = None;
 
@@ -1368,9 +1366,9 @@ fn best_base(bases: &[PyTypeRef], vm: &VirtualMachine) -> PyResult<PyTypeRef> {
             winner = Some(candidate.clone());
             base = Some(base_i.clone());
         } else {
-            return Err(vm.new_type_error(
-                "multiple bases have instance layout conflict".to_string(),
-            ));
+            return Err(
+                vm.new_type_error("multiple bases have instance layout conflict".to_string())
+            );
         }
     }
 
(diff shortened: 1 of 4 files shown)


Lib/test/test_types.py
```diff
@@ -392,8 +392,6 @@ def test(i, format_spec, result):
         test(123456, "1=20", '11111111111111123456')
         test(123456, "*=20", '**************123456')
 
-    @unittest.expectedFailure
-    @run_with_locale('LC_NUMERIC', 'en_US.UTF8')
```
youknowone, line 396: This line is not expected to be removed.

vm/src/builtins/int.rs
```diff
@@ -326,7 +326,8 @@ impl PyInt {
 
 #[pyclass(
     flags(BASETYPE),
-    with(PyRef, Comparable, Hashable, Constructor, AsNumber, Representable)
+    with(PyRef, Comparable, Hashable, Constructor, AsNumber, Representable),
+    itemsize(BigInt)
```
youknowone, line 330: I don't get how this is intended to be used yet. How to handle it if it requires multiple fields?

Snowapril: It was not intended for multiple types to be declared with `itemsize`. 
In the pyclass code, if the number of types declared with itemsize is not one, it is planned to raise an error in a way that developers can understand. Unlike `basicsize`, it seems difficult to automatically infer item type for each class delcaration, so I have chosen this way.

Snowapril: Hmm.. I think `itemsize` implementation is better to be moved to other PRs. It seems it does not cause wrong behavior in `best_base`.

Snowapril:
I can't reproduce `test_types::test_int__format__locale` failure in WSL ubuntu 22.04. It also seems not related to this PR.    . 😂

Lib/test/test_types.py
```diff
@@ -421,9 +421,6 @@ def test_int__format__locale(self):
             self.assertEqual(len(format(0, lfmt)), len(format(x, lfmt)))
             self.assertEqual(len(format(0, cfmt)), len(format(x, cfmt)))
     
-    if sys.platform != "darwin":
-        test_int__format__locale = unittest.expectedFailure(test_int__format__locale)
```
youknowone, line 425: Reverting this will fix the failing test

Snowapril: Oh my.. That code accidentally got included in the commit. I've fixed it!

vm/src/builtins/type.rs
```diff
@@ -1337,28 +1355,25 @@ fn best_base(bases: &[PyTypeRef], vm: &VirtualMachine) -> PyResult<PyTypeRef> {
...
+            return Err(
+                vm.new_type_error("multiple bases have instance layout conflict".to_string())
+            );
+        }
     }
 
-    // FIXME: Ok(base.unwrap()) is expected
-    Ok(bases[0].clone())
+    debug_assert!(base.is_some());
+    Ok(base.unwrap())
```
youknowone, line 1376: `unwrap` implies `panic!`. So `debug_assert!` is duplication.

```suggestion
    Ok(base.expect("best_base must exist"))
```

youknowone (approved):
Thank you!

PR #5325, opened 2024-05-16
Make `__bases__` mutable

dchiquito:
Assigning `__bases__` allows you to alter the class inheritance tree at any point. The updated `typing.py` uses this feature when constructing `NamedTuples`, for reasons unknown to me.

I had to make `PyType.bases` and `PyType.mro` mutable behind a `PyRwLock` in order to enable this feature.

Assigning `__bases__` has several features I wasn't sure how to implement. I left TODOs for these:
* Making sure the new method resolution order is actually a tree, i.e. it contains no cycles
* Removing the modified class from any `__subclass__` lists it used to belong to
* Removing any slots that no longer apply (not sure if this is required?)

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -312,16 +312,6 @@ impl PyType {
         }
     }
 
-    pub fn mro_map_collect<F, R>(&self, f: F) -> Vec<R>
-    where
-        F: Fn(&Self) -> R,
-    {
-        std::iter::once(self)
-            .chain(self.mro.read().iter().map(|cls| -> &PyType { cls }))
-            .map(f)
-            .collect()
-    }
-
     // This is used for class initialisation where the vm is not yet available.
     pub fn set_str_attr<V: Into<PyObjectRef>>(
         &self,
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -34,8 +34,8 @@ use std::{borrow::Borrow, collections::HashSet, fmt, ops::Deref, pin::Pin, ptr::
 #[pyclass(module = false, name = "type", traverse = "manual")]
 pub struct PyType {
     pub base: Option<PyTypeRef>,
-    pub bases: Vec<PyTypeRef>,
-    pub mro: Vec<PyTypeRef>,
+    pub bases: PyRwLock<Vec<PyTypeRef>>,
+    pub mro: PyRwLock<Vec<PyTypeRef>>,
     pub subclasses: PyRwLock<Vec<PyRef<PyWeak>>>,
     pub attributes: PyRwLock<PyAttributes>,
     pub slots: PyTypeSlots,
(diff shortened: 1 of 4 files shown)


dchiquito:
The test failures don't seem related to my changes. @youknowone any insight?

youknowone:
The failing test is about threading, it might be flaky. re-triggered failed jobs.

vm/src/builtins/type.rs
```diff
@@ -308,21 +312,14 @@ impl PyType {
         }
     }
 
-    pub fn iter_mro(&self) -> impl DoubleEndedIterator<Item = &PyType> {
-        std::iter::once(self).chain(self.mro.iter().map(|cls| -> &PyType { cls }))
-    }
-
-    pub(crate) fn mro_find_map<F, R>(&self, f: F) -> Option<R>
+    pub fn mro_map_collect<F, R>(&self, f: F) -> Vec<R>
```
youknowone, line 315: mro_map_collect is once defined here,

vm/src/builtins/type.rs
```diff
@@ -426,11 +428,37 @@ impl Py<PyType> {
     /// so only use this if `cls` is known to have not overridden the base __subclasscheck__ magic
     /// method.
     pub fn fast_issubclass(&self, cls: &impl Borrow<crate::PyObject>) -> bool {
-        self.as_object().is(cls.borrow()) || self.mro.iter().any(|c| c.is(cls.borrow()))
+        self.as_object().is(cls.borrow()) || self.mro.read().iter().any(|c| c.is(cls.borrow()))
+    }
+
+    pub fn mro_map_collect<F, R>(&self, f: F) -> Vec<R>
```
youknowone, line 434: and defined here again.
`&Py<PyType>` derefs to `PyType`. So this one probably is redundant.

Does this method can be `mro_map`(or map_mro) without collect?
Or is it not possible anymore due to RwLock?

youknowone:
> Making sure the new method resolution order is actually a tree, i.e. it contains no cycles
Leaving a comment will be enough for now.

> Removing the modified class from any __subclass__ lists it used to belong to
How is this done in CPython? Does it maintain weakrefs to subclasses?

> Removing any slots that no longer apply (not sure if this is required?)
We don't have this feature yet. To do this, we have to implement slot copy first, which will remove most of `iter_mro` calls in our codebase.

vm/src/builtins/type.rs
```diff
@@ -308,21 +312,14 @@ impl PyType {
...
-        // try std::intrinsics::likely once it is stablized
-        if let Some(r) = f(self) {
-            Some(r)
-        } else {
-            self.mro.iter().find_map(|cls| f(cls))
-        }
+        std::iter::once(self)
+            .chain(self.mro.read().iter().map(|cls| -> &PyType { cls }))
+            .map(f)
+            .collect()
```
youknowone, line 322:
```suggestion
```
Ok, this one is not used one.

vm/src/builtins/type.rs
dchiquito: Yeah, the iterator needs to hold the lock on `mro` so the only way to have a helper method is to collect the result. I think it's clumsy but it's the best way I could find :/

dchiquito: I think it's not redundant, it is used here:
```rust
#[pyclass]
impl Py<PyType> {
    ...
    #[pymethod]
    fn mro(&self) -> Vec<PyObjectRef> {
        self.mro_map_collect(|cls| cls.to_owned().into())
    }
}
```

The `PyType` implementation discards the `Py` wrapper, which is still needed in this case. Is there a better way to write this so that the `PyType::mro_map_collect` can be used?

dchiquito:
> > Removing the modified class from any subclass lists it used to belong to
> 
> How is this done in CPython? Does it maintain weakrefs to subclasses?

Yes, the subclass list is a `PyRwLock<Vec<PyRef<PyWeak>>>`, and CPython is similar-ish. The complicated part is that CPython maintains a dict of subclasses so that they are uniquely identifiable, presumably just so they can be removed or modified. The key function looks complicated to get right, so I left the TODO.

vm/src/builtins/type.rs
dchiquito: Woops, I missed https://github.com/RustPython/RustPython/pull/5325#discussion_r1605958799
^ Ignore that

youknowone (approved):
👍

PR #5337, opened 2024-06-05
Make a shared run function in ThreadedVirtualMachine

MolotovCherry:
This makes it much easier to store a `ThreadedVirtualMachine` somewhere and run it as many times as we want in another thread.

After looking at the code, I see the way things are set up with the current `run` it's not best to change it from taking `self`. However I might suggest renaming it into `into_run` or similar (along with `into_spawn_func`) as it makes the fact it takes ownership clearer. That'd leave us with being able to name the shared run function simply `run`. However these would be breaking changes, so I left the core implementations the same.

For now I've just made a simple `shared_run` instead which accomplishes this without breaking changes.

I am sure this might require quite a bit of discussion about how we really want it and some refactoring, so let me know your thoughts!

### DIFF
--- a/vm/src/vm/thread.rs
+++ b/vm/src/vm/thread.rs
@@ -93,9 +93,9 @@ impl ThreadedVirtualMachine {
     /// to the parent thread and then `join()` on the `JoinHandle` (or similar), there is a possibility that
     /// the current thread will panic as `PyObjectRef`'s `Drop` implementation tries to run the `__del__`
     /// destructor of a python object but finds that it's not in the context of any vm.
-    pub fn run<F, R>(&self, f: F) -> R
+    pub fn run<F, R>(&self, mut f: F) -> R
     where
-        F: Fn(&VirtualMachine) -> R,
+        F: FnMut(&VirtualMachine) -> R,
     {
         let vm = &self.vm;
         enter_vm(vm, || f(vm))
--- a/vm/src/vm/thread.rs
+++ b/vm/src/vm/thread.rs
@@ -100,6 +100,21 @@ impl ThreadedVirtualMachine {
         let vm = &self.vm;
         enter_vm(vm, || f(vm))
     }
+
+    /// Run a function in this thread context. This can be called multiple times.
+    ///
+    /// # Note
+    ///
+    /// If you return a `PyObjectRef` (or a type that contains one) from `F`, and don't return the object
+    /// to the parent thread and then `join()` on the `JoinHandle` (or similar), there is a possibility that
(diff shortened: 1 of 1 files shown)


youknowone:
Thanks for contribution. I'd like to know if breaking change was the only thing you concerned about, or there is something more.

MolotovCherry:
> Thanks for contribution. I'd like to know if breaking change was the only thing you concerned about, or there is something more.

The breaking change is mostly the only thing I'm worried about, since renaming the functions and keeping the same functionality it has is otherwise trivial to do

MolotovCherry:
> Changing `ThreadedVirtualMachine::run` looks fine. It will be a trivial change and our next release will be breaking change. 

Do you mean it's ok to change the current `run` to take `&self` instead? Or did you mean it's ok to rename `shared_run` to run, and rename `run` to something else?

youknowone:
Changing will be okay. We make a lot of breaking changes. And the new one covers superset

MolotovCherry:
I think that should just about do it 😁

MolotovCherry:
Looks like the latest Rust added some clippy lints which caused the build to fail elsewhere. Would you like me to fix it regardless?

youknowone:
@MolotovCherry Could you rebase (git-rebase) the patch? It is fixed on main branch. rebase will fix the broken tests

vm/src/vm/thread.rs
```diff
@@ -93,9 +93,9 @@ impl ThreadedVirtualMachine {
     /// to the parent thread and then `join()` on the `JoinHandle` (or similar), there is a possibility that
     /// the current thread will panic as `PyObjectRef`'s `Drop` implementation tries to run the `__del__`
     /// destructor of a python object but finds that it's not in the context of any vm.
-    pub fn run<F, R>(self, f: F) -> R
+    pub fn run<F, R>(&self, mut f: F) -> R
     where
-        F: FnOnce(&VirtualMachine) -> R,
```
youknowone, line 98: Is keeping FnOnce possible?

MolotovCherry: It appears to compile fine with `FnOnce`, though I'm also wondering if perhaps someone in the future maybe stores an `Fn` or `FnMut` in their type, which would make using this function impossible in such a case. It kind of feels like it goes against the "shared" aspect of it.

MolotovCherry: I added it back as `FnOnce` now

youknowone (approved):
Thank you! And sorry for delayed reviews

MolotovCherry:
> Thank you! And sorry for delayed reviews

No problem! 😁

PR #5347, opened 2024-06-24
CHORE: update as many dependencies that dont cause big breakage

JamesClarke7283:
This PR aims to update as many dependencies as we can, without breaking compatibility with our code.
Other PR's to update clap for example, can deal with the larger updates.

This is just to keep things going smooth.

I will submit this for review tomorrow, when I've done a more comprehensive update.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,12 +58,67 @@ dependencies = [
 ]
 
 [[package]]
+name = "anes"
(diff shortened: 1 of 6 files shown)


youknowone:
Can we cut this a bit not to break the build?

JamesClarke7283:
This PR is stale, its better if i start a new one.

Issue #5355, opened 2024-07-18
Inconsist with cpython in deleting __dict__


Mskxn:
In CPython, the following code can execute normally.
```python
class Foo: pass
f = Foo()
print(f.__dict__)
del f.__dict__
```
But in rustpython, it triggers TypeError:
```
Traceback (most recent call last):
File "temp.py", line 4, in <module>
                  del f.__dict__
TypeError: can't delete attribute
```
I am not sure if it is a design or a bug.
P.S., Are there any docs to explain the consistency between CPython and rustpython, or they should be exactly the same?

youknowone:
I expect that eventually be the same.

I never expect this is possible in CPython.
So this is the way to "clear" `__dict__`.

youknowone:
For contributors:
current `object_set_dict` function signature doesn't support delete.
Changing it to `pub fn object_set_dict(obj: PyObjectRef, dict: PySetterValue<PyDictRef>, vm: &VirtualMachine) -> PyResult<()>` is the starting point.
The remaining steps will be straight-forward by understanding how PySetterValue looks like.

fanninpm:
Just checking — did #5436 actually fix this issue?

maxweilun1989:
@fanninpm  I tried this on the latest commit and found that the issue hasn’t been resolved yet. Could you assign it to me?"

PR #5378, opened 2024-07-29
Support del __dict__

nielsbuwen:
I would like to take over this good first issue: #5355 

It brings the interpreter in line with the cpython behaviour when deleting `__dict__` attributes.

```python
class Thing:
    def __init__(self):
        self.x = 1


t = Thing()

del t.__dict__  # no error

t.x  # this is now an AttributeError
```

Basically this replaces the instance dict with an empty dictionary. This seems to duplicate the cpython behaviour.

Things to improve

- I'm not sure if there is a better place for the test
- ~~I don't like the `let new_dict = match dict` where i take the given dict or and empty one. I could add a `unwrap_or_default` for the `PySetterValue`?~~ (found it)
- ~~I also don't like the `let value = match value {` where i convert the `Assign` from a `PyObjectRef` into a `PyDictRef`. But i could not find a `map*` variant that let's  me pass a function that returns a PyResult~~ (i guess i found it)

I'd love to hear your suggestions

### DIFF
--- a/extra_tests/snippets/classes.py
+++ b/extra_tests/snippets/classes.py
@@ -1,0 +1,17 @@
+from testutils import assert_raises
+
+
+class Thing:
+    def __init__(self):
+        self.x = 1
+
+
+t = Thing()
+assert t.x == 1
+
+del t.__dict__
+
+with assert_raises(AttributeError):
+    _ = t.x
+
+assert t.__dict__ == {}
--- a/vm/src/builtins/object.rs
+++ b/vm/src/builtins/object.rs
@@ -500,12 +500,9 @@ pub fn object_set_dict(
     dict: PySetterValue<PyDictRef>,
     vm: &VirtualMachine,
 ) -> PyResult<()> {
-    let new_dict = match dict {
-        PySetterValue::Delete => PyDict::new_ref(&vm.ctx),
-        PySetterValue::Assign(dict) => dict,
-    };
+    let dict = dict.unwrap_or_else(|| PyDict::new_ref(&vm.ctx));
 
-    obj.set_dict(new_dict)
+    obj.set_dict(dict)
         .map_err(|_| vm.new_attribute_error("This object has no __dict__".to_owned()))
 }
 
--- a/vm/src/builtins/object.rs
+++ b/vm/src/builtins/object.rs
@@ -1,4 +1,4 @@
-use super::{PyDictRef, PyList, PyStr, PyStrRef, PyType, PyTypeRef};
+use super::{PyDict, PyDictRef, PyList, PyStr, PyStrRef, PyType, PyTypeRef};
 use crate::common::hash::PyHash;
 use crate::types::PyTypeFlags;
 use crate::{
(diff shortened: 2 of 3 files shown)


extra_tests/snippets/classes.py
youknowone, line 1: This is the correct directory. Putting this test under `builtin_object.py` is also fine.

youknowone:
The change looks good, but function type seems not allowing to do it.

```
FAIL: test_delete___dict__ (test.test_funcattrs.FunctionDictsTest.test_delete___dict__)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/test_funcattrs.py", line 360, in test_delete___dict__
    self.fail("deleting function dictionary should raise TypeError")
AssertionError: deleting function dictionary should raise TypeError

FAIL: test_setting_dict_to_invalid (test.test_funcattrs.FunctionDictsTest.test_setting_dict_to_invalid)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/test_funcattrs.py", line 331, in test_setting_dict_to_invalid
    self.cannot_set_attr(self.b, '__dict__', None, TypeError)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/test_funcattrs.py", line 42, in cannot_set_attr
    self.fail("shouldn't be able to del %s" % name)
AssertionError: shouldn't be able to del __dict__
```

The related type is `PyFunction`.

Adding a getset is possible like:

```rust
// This impl must be one of `#[pyclass]` annotated
impl PyFunction {
    #[pygetset(setter)]
    fn __dict__(zelf: &Py<Self>, value: dict: PySetterValue<PyDictRef>,, vm: &VirtualMachine) -> ...
}
```

nielsbuwen:
I found out, how to delegate the `__dict__` access to the underlying object. This brings a new problem:

`foo.__dict__`  contains the keys `__doc__`, `__module__` and `__annotations__` in RustPython, but not in CPython (there it's just an empty dict).

I guess this happens, because `execute_make_function` in ` frame.rs` assigns these three, but they are not explicitly added as members in `PyFunction`.

Should i add these three to the `PyFunction` `struct`? And if yes, in another pull request?

youknowone:
Thank you for analysis. That's a good point. Yes, making them as its own getset totally makes sense. And yes, in separated PR would be better.

nielsbuwen:
I was able to remove `__module__` and `__annotations__`  from the function `__dict__`, but the `__doc__`  does not go away, even if i implement a "pygetset" for it. The functions are not even getting called.

I declared them like this

```rust
#[pygetset(magic)]
fn doc(&self) -> PyObjectRef {...}

#[pygetset(magic, setter)]
fn set_doc(&self, value: PyObjectRef) {...}
```

So, something is intercepting the calls to my pygetset. Any idea what that might be? I looked at other pygetset for `__doc__` but none of them is getting called either.

It also does not matter whether the function actually has a doc-string. If i comment out the part in  `compile.rs` where it generates op codes to store the doc-string in the function then the doc-string of the function-type is used when writing `foo.__doc__`.

youknowone:
@nielsbuwen Making a PR of that changes will be helpful for me to understand what's happening. My first guess is `execute_make_function` in `frame.rs`

nielsbuwen:
@youknowone here you go: #5392

PR #5391, opened 2024-08-13
4982 - Implement tm_gmtoff and tm_zone

### LINKED ISSUE #4982: datetime().astimezone() is broken

JazzGlobal:
Fixes #4982 

Implements the missing`tm_gmtoff` and `tm_zone` fields in PyStructTime struct. 

The original issue's code example does not work in python 3.12. I used this instead for testing:

```py
import datetime
sgtTimeDelta = datetime.timedelta(hours=5)
sgtTZObject = datetime.timezone(sgtTimeDelta,
                                name="EST")
print(datetime.datetime(3001, 1, 19, 7, 59, 59, 999999, sgtTZObject).astimezone())
```

### DIFF
--- a/Lib/test/test_time.py
+++ b/Lib/test/test_time.py
@@ -718,9 +718,10 @@ class TestAsctime4dyear(_TestAsctimeYear, _Test4dYear, unittest.TestCase):
 
 
 class TestPytime(unittest.TestCase):
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     @skip_if_buggy_ucrt_strfptime
-    @unittest.skip("TODO: RUSTPYTHON, AttributeError: module 'time' has no attribute '_STRUCT_TM_ITEMS'")
-    # @unittest.skipUnless(time._STRUCT_TM_ITEMS == 11, "needs tm_zone support")
+    @unittest.skipUnless(time._STRUCT_TM_ITEMS == 11, "needs tm_zone support")
     def test_localtime_timezone(self):
 
         # Get the localtime and examine it for the offset and zone.
@@ -755,16 +756,18 @@ class TestPytime(unittest.TestCase):
         self.assertEqual(new_lt.tm_gmtoff, lt.tm_gmtoff)
         self.assertEqual(new_lt9.tm_zone, lt.tm_zone)
     
-    @unittest.skip("TODO: RUSTPYTHON, AttributeError: module 'time' has no attribute '_STRUCT_TM_ITEMS'")
-    # @unittest.skipUnless(time._STRUCT_TM_ITEMS == 11, "needs tm_zone support")
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    @unittest.skipUnless(time._STRUCT_TM_ITEMS == 11, "needs tm_zone support")
     def test_strptime_timezone(self):
         t = time.strptime("UTC", "%Z")
         self.assertEqual(t.tm_zone, 'UTC')
         t = time.strptime("+0500", "%z")
         self.assertEqual(t.tm_gmtoff, 5 * 3600)
     
-    @unittest.skip("TODO: RUSTPYTHON, AttributeError: module 'time' has no attribute '_STRUCT_TM_ITEMS'")
-    # @unittest.skipUnless(time._STRUCT_TM_ITEMS == 11, "needs tm_zone support")
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    @unittest.skipUnless(time._STRUCT_TM_ITEMS == 11, "needs tm_zone support")
     def test_short_times(self):
 
         import pickle
--- a/derive-impl/src/pystructseq.rs
+++ b/derive-impl/src/pystructseq.rs
@@ -80,16 +80,19 @@ fn field_names(input: &mut DeriveInput) -> Result<(Vec<Ident>, Vec<Ident>)> {
 }
 
 pub(crate) fn impl_pystruct_sequence(mut input: DeriveInput) -> Result<TokenStream> {
-    let (not_skipped_fields, _skipped_fields) = field_names(&mut input)?;
+    let (not_skipped_fields, skipped_fields) = field_names(&mut input)?;
     let ty = &input.ident;
     let ret = quote! {
         impl ::rustpython_vm::types::PyStructSequence for #ty {
-            const FIELD_NAMES: &'static [&'static str] = &[#(stringify!(#not_skipped_fields)),*];
+            const REQUIRED_FIELD_NAMES: &'static [&'static str] = &[#(stringify!(#not_skipped_fields),)*];
+            const OPTIONAL_FIELD_NAMES: &'static [&'static str] = &[#(stringify!(#skipped_fields),)*];
             fn into_tuple(self, vm: &::rustpython_vm::VirtualMachine) -> ::rustpython_vm::builtins::PyTuple {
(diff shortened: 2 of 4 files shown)


miviwi:
I was about to say you don't have to `use` types to well... just *use* them.

Buf if you're a noob then I'll try to explain:
`use` - in contrast with other languages like C++ or JS - only creates an alias, but doesn't actually import anything; In fact in Rust you don't need to import types from a crate at all to use them!

In other words: if you have to name a type - say to call an associated method like `OnceCell::new()` - then you'd need:
```rust
use std::cell::OnceCell;
```
...somewhere in the file/function

but if you're only *calling* a method that returns a `OnceCell` then no `use` is needed (or if you'd use the full name - `std::cell::OnceCell::new()` - then no `use` is needed also)

JazzGlobal:
> I was about to say you don't have to `use` types to well... just _use_ them.
> 
> Buf if you're a noob then I'll try to explain: `use` - in contrast with other languages like C++ or JS - only creates an alias, but doesn't actually import anything; In fact in Rust you don't need to import types from a crate at all to use them!
> 
> In other words: if you have to name a type - say to call an associated method like `OnceCell::new()` - then you'd need:
> 
> ```rust
> use std::cell::OnceCell;
> ```
> 
> ...somewhere in the file/function
> 
> but if you're only _calling_ a method that returns a `OnceCell` then no `use` is needed (or if you'd use the full name - `std::cell::OnceCell::new()` - then no `use` is needed also)

Thanks for the feedback. I did so _right_ before this comment 👍

youknowone (commented):
Thank you for contributing. You will want to run some tests about timezones. Here is a few commands not to wait the full CI run.

```
cargo run extra_tests/snippets/stdlib_datetime.py  # most simple one
cargo run Lib/test/test_time.py  # test `time` module
cargo run Lib/test/test_datetime.py  # test `datetime` module
```

miviwi:
@youknowone if I'm right and there's no way to mark a `PyStructSequence` field/s to not get unpacked I don't think you can solve this issue like the author is suggesting

JazzGlobal:
> cargo run extra_tests/snippets/stdlib_datetime.py

I think you're right. I came to conclusion much slower though through messing around with the failing tests.

miviwi:
@JazzGlobal well... this is going to require major changes (completely restructuring the definition of `PyStructTime`) and unless you're up for that - and if you are then great, but it'll be a though one esp. if you're new to Rust - then I'm closing the PR.

So... are you up for it?

JazzGlobal:
> @JazzGlobal well... this is going to require major changes (completely restructuring the definition of `PyStructTime`) and unless you're up for that - and if you are then great, but it'll be a though one esp. if you're new to Rust - then I'm closing the PR.
> 
> So... are you up for it?

yeah I'll fiddle with this for a while long.

youknowone:
@miviwi what are you doing? you don't close the PR. you can't do that.

miviwi:
@youknowone hold on... not closing anything yet, or now that JazzGlobal has spoken not at all 😅

JazzGlobal:
Community is a bit more moderated since I last contributed lol. I'll look at some of the other Traverse implementations for hints on how I should do it for PyStructTime. Thanks guys!

youknowone:
@JazzGlobal Don't worry about it that much. We can solve things step by step, eventually. Nothing changed from the time you visited before.
I agree PyStructureSequence lacks of the feature. The related part is here. https://github.com/RustPython/RustPython/blob/main/derive-impl/src/pystructseq.rs#L5-L28

miviwi:
@JazzGlobal Personally I would first try just leaving `PyStructTime` as a `PyStructSequence` and addding the new fields as getter methods. Yes it'll be slower, but performance and timezones are sort of "opposites" anyway, so... it's worth a shot. Definitely seems easier than uprooting the whole struct

youknowone:
I am not getting how PyTraverse is related to this problem yet. I'd like to fix pystructseq side first.

@JazzGlobal Are you also interested in proc macro? Then the point will be the start point. Otherwise let me make some changes on PyStructSequence. I don't have a good idea how to change it right now though.

miviwi:
@youknowone not as the code is right now, but making the struct a `Traversable` (instead of a `PyStructSequence`) is the only way I could think up to keep those extra fields and be consistent with the docs (i.e. have the tests pass)

youknowone:
Maybe there is confusion. PyTraverse used only for GC. I hope I can solve it with PyStructSequence.

JazzGlobal:
> I am not getting how PyTraverse is related to this problem yet. I'd like to fix pystructseq side first.
> 
> @JazzGlobal Are you also interested in proc macro? Then the point will be the start point. Otherwise let me make some changes on PyStructSequence. I don't have a good idea how to change it right now though.

If utilizing Traverse is not an option then go for it.

youknowone:
I don't think any PR need to be closed due to blocking issues. We always encounter unexpected blocking issues for numerous patches.  We can fix them and come back to the original issue whenever we can. We can review unresolved opened PRs later. But not about unresolved closed PRs. So unclosed PRs are more helpful than closed PRs.

youknowone:
How about adding `#[pystruct(skip)]` to mark fields to ignore?

JazzGlobal:
> How about adding `#[pystruct(skip)]` to mark fields to ignore?

Do you mean `#[pytraverse(skip)]` ?

youknowone:
Yes, adding a new attribute `#[pystruct(<something>)]` 

pytraverse is only used for GC, so not here.

JazzGlobal:
> Yes, adding a new attribute `#[pystruct(<something>)]`
> 
> pytraverse is only used for GC, so not here.

I'm not familiar with Rust macros. Will pystruct("skip") essentially be the same thing implementation-wise as pytraverse?

youknowone:
Unfortunately, it now runs but still tm_gmtoff is not accessible.

youknowone (approved):
@JazzGlobal It look long time, but finally added required support. Thank you for contributing. I am sorry if you felt inconvenience. Please don't worry to be moderated not by CoC but by code. Whenever everybody every code will be welcomed.

JazzGlobal:
> @JazzGlobal It look long time, but finally added required support. Thank you for contributing. I am sorry if you felt inconvenience. Please don't worry to be moderated not by CoC but by code. Whenever everybody every code will be welcomed.

No problem at all. I like contributing to the project and it isn't about me. Thanks for your help in facilitating this!

PR #5392, opened 2024-08-14
Add get/set methods for function doc/module/annotations

nielsbuwen:
WIP

this adds getter and setter for the `doc`, `module` and `annotations` magic attributes for the python function objects.

`module` and `annotations` work without trouble. `doc` is different. The getter and setter were never called until i commented the line `class.set_attr(identifier!(ctx, __doc__), ctx.new_str(doc).into());` in `vm/src/class.rs`.

This will probably break other stuff. But it gets me closer to the goal.

Details

see this python code

```python
def foo():
    """doc string"""

print(foo.__dict__)
```

Output:

- CPython: `{}` - empty dict
- before PR: `{'__doc__': 'doc string', '__module__': '__main__', '__annotations__': {}}`
- without "fix" in `vm/src/class.rs`: `{'__doc__': 'doc string'}`
- with "fix": `{}` - empty dict

Any ideas how to fix this correctly?

### DIFF
--- a/vm/src/builtins/function.rs
+++ b/vm/src/builtins/function.rs
@@ -415,14 +415,19 @@ impl PyFunction {
         *self.name.lock() = name;
     }
 
-    #[pygetset(magic)]
-    fn doc(&self) -> PyObjectRef {
-        self.doc.lock().clone()
+    #[pymember(magic)]
+    fn doc(_vm: &VirtualMachine, zelf: PyObjectRef) -> PyResult {
+        let zelf: PyRef<PyFunction> = zelf.downcast().unwrap_or_else(|_| unreachable!());
+        let doc = zelf.doc.lock();
+        Ok(doc.clone())
     }
 
-    #[pygetset(magic, setter)]
-    fn set_doc(&self, doc: PyObjectRef) {
-        *self.doc.lock() = doc
+    #[pymember(magic, setter)]
+    fn set_doc(vm: &VirtualMachine, zelf: PyObjectRef, value: PySetterValue) -> PyResult<()> {
+        let zelf: PyRef<PyFunction> = zelf.downcast().unwrap_or_else(|_| unreachable!());
+        let value = value.unwrap_or_none(vm);
+        *zelf.doc.lock() = value;
+        Ok(())
     }
 
     #[pygetset(magic)]
--- a/vm/src/builtins/function.rs
+++ b/vm/src/builtins/function.rs
@@ -52,6 +52,7 @@ unsafe impl Traverse for PyFunction {
 }
 
 impl PyFunction {
+    #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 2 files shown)


youknowone:
Uh, that's confusing. Removing `class.set_attr` from `extend_class` affect `foo.__dict__`? that's not a class.

youknowone:
`PyObject::set_attr` seems to have a bug affect by class attr

youknowone:
Ah, got it. `PyFunction` is expected to have `__doc__` as a `pymember`. But not yet.

youknowone:
class `__doc__` is a little bit more complicated. The builtin types doc is read-only. while heap types (usually a python-defined class) is not. To make it, `HeapTypeExt` is expected to have writable `doc` field with a lock.

youknowone:
Either way okay as long as CI passes.

vm/src/builtins/function.rs
```diff
@@ -406,6 +416,41 @@ impl PyFunction {
...
+        Ok(())
+    }
+
+    #[pygetset(magic)]
+    fn module(&self) -> PyObjectRef {
+        self.module.lock().clone()
+    }
+
+    #[pygetset(magic, setter)]
+    fn set_module(&self, module: PyObjectRef) {
```
youknowone, line 440: It's going well! `__module__` need to support delete operation

```suggestion
    fn set_module(&self, module: PySetterValue) {
```
To support deletion, the value type need to be changed. Using `match` statement helps to distinguish set and delete operation

CI Result:
```
ERROR: test_local_lookup_error (test.test_pickle.InMemoryPickleTests.test_local_lookup_error)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/pickletester.py", line 2885, in test_local_lookup_error
    del f.__module__
TypeError: can't delete attribute

ERROR: test_local_lookup_error (test.test_pickle.PyPicklerTests.test_local_lookup_error)
  File "/home/runner/work/RustPython/RustPython/pylib/Lib/test/pickletester.py", line 2885, in test_local_lookup_error
    del f.__module__
TypeError: can't delete attribute
```

youknowone (changes requested):

vm/src/builtins/function.rs
nielsbuwen: Looks like the `__module__` is `None` after deletion in CPython, so i implemented it that way

youknowone (approved):
Thank you so much!
`doc` handling looks much more tricky than it looks like. Let's try to get it.

youknowone:
Please let me know if you blocked by anything.
`PyObject::set_attr` need to be investigated how it handles descriptors.
`execute_make_function` is the part making function on runtime.
`fn __build_class__` is the part making a class on runtime.

`cargo run -- -m dis <source code>` will show compiled python bytecode.

PR #5396, opened 2024-09-07
Fix copysign behavior for NaNs

crazymerlyn:
This allows enabling test_constructor_negative_nans_from_string


crazymerlyn:
The failure in test_re seems independent of the pr as it is also failing on main.

youknowone:
Thank you for checking!

youknowone:
#5398 fixed test_re failure. Could you please rebase it?

crazymerlyn:
Done

youknowone (approved):
Thank you for contributing and the good catch!

PR #5398, opened 2024-09-10
less strict time bound for test_re.test_search_anchor_at_beginning

youknowone:
It recently started to fail.

A bit worrying because 0.1 is already very greater than 0.0003

https://github.com/RustPython/RustPython/actions/runs/10748560557
https://github.com/RustPython/RustPython/actions/runs/10760680587


qingshi163:
It is slowed on slicing the string, the way to improve that can only be done by expose the raw ptr from the sre_engine.

PR #5400, opened 2024-09-14
Add zoneinfo Library from Python 3.12.6

rroblf01:
In this PR I have added zoneinfo with the corresponding test suite.


rroblf01:
> Thank you so much for contributing! Just one more thing. Could you check which Python version did you brought the library from?

Good afternoon,
This library was introduced as is in 3.11, but it is compatible with 3.12 and so far with what has been developed for python 3.13.0rc2

youknowone:
Do you mean you copied them from 3.13.0rc2? That's for record purpose. We usually leave version information the copied version for later.

rroblf01:
> Do you mean you copied them from 3.13.0rc2? That's for record purpose. We usually leave version information the copied version for later.

Hello,

I prefer to say that it is taken from python 3.12.6.

youknowone:
Thank you for confirming!
And welcome to the project.

PR #5405, opened 2024-09-19
Fix set intersection_update implementation

crazymerlyn:
The previous implementation had multiple bugs:

- self intersection would result in nil set (#3881, #3992 )
- Intersecting {1, 2, 3} in {1} and {2} would result in {1} instead of {}.

The new implementation uses the approach chosen by [cpython](https://github.com/python/cpython/blob/5f011115943933ff36adf997c886d73ea88003fb/Objects/setobject.c#L1481-L1495), i.e., create a normal intersection and then replace `self` with the result.

Allows turning on one test in `test_weakset`. The equivalent test in `test_set` still fails due to some of the other set operations also being buggy on self application.

### DIFF
--- a/Lib/test/test_weakset.py
+++ b/Lib/test/test_weakset.py
@@ -322,8 +322,6 @@ class TestWeakSet(unittest.TestCase):
             else:
                 self.assertNotIn(c, self.s)
 
-    # TODO: RUSTPYTHON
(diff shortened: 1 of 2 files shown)


coolreader18 (approved):
Thank you for contributing!

PR #5408, opened 2024-09-21
Avoid allocating a vector of elements when hashing frozenset

crazymerlyn:
Adds a `try_fold_keys` method to `Dict` which allows performing common operations on all elements without needing to create a `Vec` first.

This sped up the `test_hash_effectiveness` test for FrozenSet from 4.5s to 3.2s on my machine.


youknowone (approved):
Nice idea. I added a commit to remove a RC inc/dec

PR #5409, opened 2024-09-21
Cache hash value for FrozenSets

### LINKED ISSUE #5167: Investigate test_set.TestFrozenSet.test_hash_effectiveness performance
## Summary
`test_set.TestFrozenSet.test_hash_effectiveness` verifies some properties of `frozenset.__hash__`. One of the steps involves calculating the hash of `powerset(range(18))`. This part of the test runs so slowly that the parameters had to be reduced for it to finish.

## Expected

The test runs as written in an acceptable time frame.

## Actual

The test has been modified to `powerset(range(13))`.

## Code in question

```
        def powerset(s):
            for i in range(len(s)+1):
                yield from map(frozenset, itertools.combinations(s, i))

        for n in range(18):
            t = 2 ** n
            mask = t - 1
            for nums in (range, zf_range):
                u = len({h & mask for h in map(hash, powerset(nums(n)))})
                self.assertGreater(4*u, t)
```
I suspect the problem is in generating the powerset, not calculating the hash.

crazymerlyn:
Adds a hash field to the `PyFrozenSet` data type in order to avoid recomputing the hash of an immutable object.

Similar to the way it's done in [cpython](https://github.com/python/cpython/blob/5f011115943933ff36adf997c886d73ea88003fb/Objects/setobject.c#L761-L774) but the hash value is stored in the PyFrozenSet struct instead of the inner Set datatype.

Massively speeds up the `test_hash_effectiveness` test and allows using the same limit of 2**18 elements on the test as cpython.

RustPython is still about 5x slower than python at running this test (0.5s vs 2.5s on my machine).

Of course, in practical use cases repeatedly hashing the same frozenset is not common so I doubt this is a performance win in general. Just adding this to be consistent with cpython.

Fixes #5167

### DIFF
--- a/vm/src/builtins/set.rs
+++ b/vm/src/builtins/set.rs
@@ -24,6 +24,10 @@ use crate::{
     AsObject, Context, Py, PyObject, PyObjectRef, PyPayload, PyRef, PyResult, TryFromObject,
 };
 use once_cell::sync::Lazy;
+use rustpython_common::{
+    atomic::{Ordering, PyAtomic, Radium},
+    hash,
+};
 use std::{fmt, ops::Deref};
 
 pub type SetContentType = dictdatatype::Dict<()>;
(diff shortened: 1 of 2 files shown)


vm/src/builtins/set.rs
```diff
@@ -71,9 +72,18 @@ impl PySet {
 }
 
 #[pyclass(module = false, name = "frozenset", unhashable = true)]
-#[derive(Default)]
 pub struct PyFrozenSet {
     inner: PySetInner,
+    hash: PyRwLock<PyHash>,
```
coolreader18, line 77: This could probably be an atomic instead of a lock - you can look at `PyStr` to see how it does it (using `rustpython_common::atomic::PyAtomic<PyHash>`)

crazymerlyn: Good Point. Updated to use a PyAtomic instead. Also replaced the naked constant -1 with hash::SENTINEL

youknowone (approved):
👍

PR #5411, opened 2024-09-22
Fix inconsistent behavior of Decimal.from_float

crazymerlyn:
Rustpython includes the test added in https://github.com/python/cpython/pull/65, but not the behavior, leading to the test being disabled. This patches RustPython with the above pr  to update `from_float` behavior when passed an int in order to pass the test.


coolreader18 (approved):

For posterity: this is updating _pydecimal to a newer version from CPython.

PR #5412, opened 2024-09-23
Remove dependency on uuid-macro-internal

crazymerlyn:
`uuid-macro-internal` is only relevant when using `uuid!` macro to parse uuids at compile time. RustPython doesn't do that so it's unnecessary.


youknowone (approved):
👍

PR #5413, opened 2024-09-23
Remove dependency on proc-macro-crate

crazymerlyn:
This dependency is only useful when modifying the crate name of `num_enum` crate (https://github.com/illicitonion/num_enum/blob/86d359f430ecd12be4d544f55bc68c27a3cda82f/num_enum_derive/src/parsing.rs#L545-L550). RustPython doesn't do that so it's unnecessary.

Gets rid of some transitive dependencies like `toml_edit`, `toml_datetime` and `nom8` as well.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1420,15 +1420,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "nom8"
-version = "0.2.0"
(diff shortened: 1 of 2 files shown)


coolreader18 (approved):
Awesome, thank you!

PR #5418, opened 2024-09-28
Handle MemoryError for sequences

### LINKED ISSUE #1750: list panic when size > isize::MAX

crazymerlyn:
Fixes #1750 

Based on #1779 which seems to be abandoned.

### DIFF
--- a/vm/src/vm/mod.rs
+++ b/vm/src/vm/mod.rs
@@ -53,6 +53,8 @@ pub use interpreter::Interpreter;
 pub(crate) use method::PyMethod;
 pub use setting::Settings;
 
+pub const MAX_MEMORY_SIZE: usize = isize::MAX as usize;
+
 // Objects are live when they are on stack, or referenced by a name (for now)
 
 /// Top level container of a python virtual machine. In theory you could
--- a/vm/src/sequence.rs
+++ b/vm/src/sequence.rs
@@ -1,6 +1,10 @@
 use crate::{
-    builtins::PyIntRef, function::OptionalArg, sliceable::SequenceIndexOp, types::PyComparisonOp,
-    vm::VirtualMachine, AsObject, PyObject, PyObjectRef, PyResult,
+    builtins::PyIntRef,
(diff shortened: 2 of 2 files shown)


vm/src/vm/mod.rs
```diff
@@ -53,6 +53,8 @@ pub use interpreter::Interpreter;
 pub(crate) use method::PyMethod;
 pub use setting::Settings;
 
+pub const MAX_MEMORY_SIZE: usize = (usize::MAX >> 3) + 1;
```
youknowone, line 56: Thank you! Is this constant came from CPython?

youknowone:
Added @coolreader18 as a reviewer who reviewed #1779

vm/src/vm/mod.rs
crazymerlyn: Nah, I just copied it from the last PR. CPython seems to use max usize instead: https://github.com/python/cpython/blob/e0a41a5dd12cb6e9277b05abebac5c70be684dd7/Objects/listobject.c#L740, https://github.com/python/cpython/blob/e0a41a5dd12cb6e9277b05abebac5c70be684dd7/Include/pyport.h#L146

coolreader18: That looks like `isize::MAX`, since the first s in `ssize_t` means signed. Which makes sense, since that's the max size an allocation can be for LLVM. I do think we could maybe just use `isize::MAX`, because otherwise we're unnecessarily constraining allocation size to "just" 512GiB on 32-bit systems.

crazymerlyn: Updated to use `isize::MAX`. Also updated to include the size of the type and fixed the potential integer overflow problem.

coolreader18 (approved):

PR #5420, opened 2024-10-03
Make stdio a feature

danielstuart14:
When embedding RustPython into a graphical windows application (where windows_subsystem = "windows"), it fails to create the RustPython VM.

This happens because the VM will try to open stdin/stdout/stderr when initializing, but stdio isn't available for /SUBSYSTEM:WINDOWS apps.

This commit makes stdio a feature, so it can be disabled when needed.

https://doc.rust-lang.org/reference/runtime.html#the-windows_subsystem-attribute
https://learn.microsoft.com/en-us/cpp/build/reference/subsystem-specify-subsystem?view=msvc-170 https://asawicki.info/news_1768_ways_to_print_and_capture_text_output_of_a_process

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,9 +10,10 @@ repository.workspace = true
 license.workspace = true
 
 [features]
-default = ["threading", "stdlib", "zlib", "importlib"]
+default = ["threading", "stdlib", "stdio", "zlib", "importlib"]
 importlib = ["rustpython-vm/importlib"]
 encodings = ["rustpython-vm/encodings"]
+stdio = ["rustpython-vm/stdio"]
 stdlib = ["rustpython-stdlib", "rustpython-pylib", "encodings"]
 flame-it = ["rustpython-vm/flame-it", "flame", "flamescope"]
 freeze-stdlib = ["stdlib", "rustpython-vm/freeze-stdlib", "rustpython-pylib?/freeze-stdlib"]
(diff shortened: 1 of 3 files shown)


coolreader18:
Probably a better solution would be to implement the `_WindowsConsoleIO` type as CPython does - it seems like it intercepts opening an stdio fd, and instead of using the CRT functions uses the win32 console API.

coolreader18:
Although for now, this is probably blocking you, so a feature flag is fine. Could you make it an off-by-default `no-stdio` flag, though? Probably just on `rustpython-vm`. Given this is pretty niche, it'd be good for it to only be disabled when desired, rather than whenever `default-features = false` is specified.

youknowone:
Since `stdio` fits better than `no-stdio` by common Rust convention, leaving a comment about the decision will be better.
@coolreader18 will you suggest the comment to add?

danielstuart14:
Changed it to set stdios to None, as I found that some imports would fail (ex: subprocess) if not set, and that seems to be an expected value (CPython expects stdin/stdout/stderr to be TextIO or None).

It is possible for the user to still set the stdios to a custom file if needed, after initializing the vm:
image

youknowone:
@coolreader18 could you follow this up? I am not good at windows dev

coolreader18 (approved):
Sorry for the long delay. This seems reasonable to me; thank you!

coolreader18:
Did a manual squash-merge (dabd93c255ba325d742c748e3f6e90c7aa4bfb38) - sorry again for the delay!

PR #5423, opened 2024-10-15
[WIP] Use Ruff Parser

### LINKED ISSUE #5299: Replace parser
## Summary

Check if Ruff parser is able/worth to replace current rustpython-parser

https://github.com/astral-sh/ruff/tree/main/crates/ruff_python_parser

qingshi163:
closes #5299


youknowone:
I am glad to see this, awesome.

qingshi163:
There is a proc macro panic witch I did not understand. Could you try to solve it? @youknowone

vm/src/vm/mod.rs
```diff
@@ -904,22 +904,22 @@ fn core_frozen_inits() -> impl Iterator<Item = (&'static str, FrozenModule)> {
...
-    ext_modules!(
-        iter,
-        dir = "./Lib/python_builtins",
-        crate_name = "rustpython_compiler_core"
-    );
+    // ext_modules!(
+    //     iter,
+    //     dir = "./Lib/python_builtins",
+    //     crate_name = "rustpython_compiler_core"
+    // );
```
youknowone, line 911: I am sorry to be late.
The error caused when compiling Lib/python_builtins/__phello__/spam.py
Parsing step is ok, but compiling fails.

compiler/src/lib.rs
```diff
@@ -1,90 +1,170 @@
...
+            symboltable::SymbolTable::scan_expr(
+                &ast.into_syntax().expect_expression(),
+                source_code.clone(),
+            )
         }
     };
-    res.map_err(|e| e.into_codegen_error(source_path.to_owned()).into())
+    res.map_err(|e| e.into_codegen_error(source_code.path.to_owned()).into())
+}
+
```
youknowone, line 164: I couldn't find the root cause yet, but adding this test will make using debugger easy for the problem

```suggestion
#[test]
fn test_compile_phello() {
    let code = r#"
initialized = True

def main():
    print("Hello world!")

if __name__ == '__main__':
    main()
"#;
    let compiled = compile(&code, Mode::Exec, "<>", CompileOpts::default());
    dbg!(compiled.expect("compile error"));
}
```

etaloof:
Hi, what is the status on PR? I can work on this if needed.

etaloof:
Sure, should I make a new PR when I am done? I don't think Github allows me to add my commits here.

compiler/codegen/src/compile.rs
```diff
@@ -1249,21 +1292,21 @@ impl Compiler {
...
-        type_params: &[located_ast::TypeParam],
+        type_params: Option<&TypeParams>,
     ) -> CompileResult<()> {
         self.prepare_decorators(decorator_list)?;
 
         // If there are type params, we need to push a special symbol table just for them
-        if !type_params.is_empty() {
+        if !type_params.is_some() {
             self.push_symbol_table();
         }
```
dcechano, line 1307: @qingshi163 @youknowone: This is the source of the bug.
As the comment says, `// If there are type params, we need to push a special symbol table just for them`, but the change to `!type_params.is_some()` (as in `type_params.is_none()`) pushes if there are NO params, causing the `panic!` in `push_symbol_table`. On my machine changing to
```rust 
  if type_params.is_some() {
      self.push_symbol_table();
  }
``` 
allows `test_compile_phello` to pass.

youknowone:
oh no... I forget to reset upstream to @etaloof's, now it is pushed here

coolreader18:
Superseded by #5494

PR #5436, opened 2024-11-10
Fix compile error when using `vm-tracing-logging`

buciuMT:
While trying to fix https://github.com/RustPython/RustPython/issues/5355, I stumbled upon a compile error when the `vm-tracing-logging` feature is enabled.

This PR fixes the above mentioned compile error.

One other thing that came up, is that [/vm/src/protocol/object.rs#L172 ](https://github.com/RustPython/RustPython/blob/98d09e781623ccec64643cf5412453cc78ba4f18/vm/src/protocol/object.rs#L172) generates very long traces and I think that it would be a good idea to use another format since it prints all the data of `self` and `value`.


youknowone (approved):
Thank you!

PR #5414, opened 2024-09-23
Update clap to v4

coolreader18:
Been a long time coming, but it wasn't too hard in the end. Also fixed some parity issues that came up while trying to figure out the correct behavior.

closes #4445 #3512

PR #5442, opened 2024-11-13
Miscellaneous cli-related parity fixes

coolreader18:
Part of this is taken from #5414.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -50,13 +50,14 @@ mod interpreter;
 mod settings;
 mod shell;
 
-use atty::Stream;
 use rustpython_vm::{scope::Scope, PyResult, VirtualMachine};
-use std::{env, process::ExitCode};
+use std::env;
+use std::io::IsTerminal;
+use std::process::ExitCode;
 
 pub use interpreter::InterpreterConfig;
 pub use rustpython_vm as vm;
-pub use settings::{opts_with_clap, RunMode};
+pub use settings::{opts_with_clap, InstallPipMode, RunMode};
 
 /// The main cli of the `rustpython` interpreter. This function will return `std::process::ExitCode`
 /// based on the return code of the python code ran through the cli.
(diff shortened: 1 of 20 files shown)


src/lib.rs
```diff
@@ -183,33 +172,46 @@ fn run_rustpython(vm: &VirtualMachine, run_mode: RunMode, quiet: bool) -> PyResu
...
+        );
+
+        eprintln!("Type \"help\", \"copyright\", \"credits\" or \"license\" for more information.");
+    }
+    let res = match run_mode {
         RunMode::Command(command) => {
             debug!("Running command {}", command);
-            vm.run_code_string(scope, &command, "<stdin>".to_owned())?;
+            vm.run_code_string(scope.clone(), &command, "<stdin>".to_owned())
+                .map(drop)
```
youknowone, line 196: Is this a common convention to drop value from Result? It looks like `drop` is important logic here to me. If not, a weak suggestion.

```suggestion
                .map(|_| ())
```

src/settings.rs
```diff
@@ -358,8 +345,13 @@ fn settings_from(matches: &ArgMatches) -> (Settings, RunMode) {
...
 /// Get environment variable and turn it into integer.
-fn get_env_var_value(name: &str) -> Result<u8, std::env::VarError> {
-    env::var(name).map(|value| u8::from_str(&value).unwrap_or(1))
+fn get_env_var_value(name: &str) -> Option<u8> {
+    env::var_os(name).filter(|v| !v.is_empty()).map(|value| {
+        value
+            .to_str()
+            .and_then(|v| v.parse::<u8>().ok())
+            .unwrap_or(1)
+    })
```
youknowone, line 354:
```suggestion
    let value_str = env::var_os(name).filter(|v| !v.is_empty())?;
    let value = value_str
        .to_str()
        .and_then(|v| v.parse::<u8>().ok())
        .unwrap_or(1);
    Some(value)
```

youknowone (approved):

src/settings.rs
coolreader18: Given there's 2 layers of Option here, I think I'd like to keep the logic as it is - looking at the suggested version my instinct was to change it to `.to_str()?.parse().unwrap_or(1)`, but that's different. I think the explicit `map()` makes more clear the control flow and conditions.

src/lib.rs
coolreader18: I've seen `.map(drop)` a decent amount, yeah.

src/settings.rs
youknowone: line 349 and the following lines are unrelated logics. early return reveals line 350 is independent to 349 and code readers can forget about 349 when going to 350.
Using `map` make readers essentially follow the full logic without benefit.

PR #5443, opened 2024-11-15
Implement IncrementalNewlineDecoder in rust

coolreader18:


derive-impl/src/from_args.rs
```diff
@@ -138,7 +139,7 @@ fn generate_field((i, field): (usize, &Field)) -> Result<TokenStream> {
     };
 
     let name = field.ident.as_ref();
-    let name_string = name.map(Ident::to_string);
+    let name_string = name.map(|ident| ident.unraw().to_string());
```
youknowone, line 142: 👍

vm/src/builtins/tuple.rs
```diff
@@ -56,11 +56,30 @@ impl IntoPyTuple for Vec<PyObjectRef> {
...
         impl<$($T: ToPyObject),*> IntoPyTuple for ($($T,)*) {
             fn into_pytuple(self, vm: &VirtualMachine) -> PyTupleRef {
-                PyTuple::new_ref(vec![$(self.$idx.to_pyobject(vm)),*], &vm.ctx)
+                #[allow(non_snake_case)]
+                let ($($T,)*) = self;
+                PyTuple::new_ref(vec![$($T.to_pyobject(vm)),*], &vm.ctx)
+            }
+        }
+
+        // TODO: figure out a way to let PyObjectRef implement TryFromBorrowedObject, and
```
youknowone, line 73: `PyObjectRef` cannot, but `PyObject` can. Since `PyObject` can always be borrowed from `PyObjectRef`, I expect it has same effect.

```rust
impl<'a> TryFromBorrowedObject<'a> for &'a PyObject {
    fn try_from_borrowed_object(_vm: &VirtualMachine, obj: &'a PyObject) -> PyResult<Self> {
        Ok(obj)
    }
}
```

youknowone (approved):
Great, finally we have rust implementation of newline decoder.
Please check comment if you need TryFromBorrowedObject for PyObject.

vm/src/builtins/tuple.rs
youknowone: Could you explain more about it?
For me, it doesn't seem possible without increasing refcount. And increasing a refcount is something not expected from TryFromBorrowedObject

coolreader18: Oh, I didn't realize that was part of the contract for TryFromBorrowedObject. It seems useful to me to have some way of generalizing over a type that can be extracted from a borrowed pyobj, but doesn't necessarily *have* to borrow from it - like serde's Deserialize vs DeserializeOwned. Like this case - the T in extract_tuple<T>() should be *able* to borrow from the pyobjectref in the tuple, but there's no reason we should *enforce* it - a `(PyObjectRef,)` is a perfectly reasonable thing to put there.

youknowone: Thanks, I think I am going to catch up.

PR #5444, opened 2024-11-16
Update fstring from v3.12.7

key262yek:
Update test_fstring.py from CPython v3.12.7

### DIFF
--- a/Lib/test/test_fstring.py
+++ b/Lib/test/test_fstring.py
@@ -8,15 +8,18 @@
 # Unicode identifiers in tests is allowed by PEP 3131.
 
 import ast
+import datetime
 import os
 import re
 import types
 import decimal
 import unittest
+import warnings
+from test import support
 from test.support.os_helper import temp_cwd
-from test.support.script_helper import assert_python_failure
+from test.support.script_helper import assert_python_failure, assert_python_ok
 
-a_global = 'global variable'
+a_global = "global variable"
 
 # You could argue that I'm too strict in looking for specific error
 #  values with assertRaisesRegex, but without it it's way too easy to
@@ -24,6 +27,7 @@ a_global = 'global variable'
 #  triple quotes, raw strings, backslashes, etc. I think it's a
 #  worthwhile tradeoff. When I switched to this method, I found many
 #  examples where I wasn't testing what I thought I was.
+
 
 class TestCase(unittest.TestCase):
     def assertAllRaise(self, exception_type, regex, error_strings):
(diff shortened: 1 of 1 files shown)


Lib/test/test_fstring.py
```diff
@@ -329,13 +334,13 @@ def test_ast_line_numbers_multiline_fstring(self):
         self.assertEqual(t.body[1].lineno, 3)
         self.assertEqual(t.body[1].value.lineno, 3)
         self.assertEqual(t.body[1].value.values[0].lineno, 3)
-        self.assertEqual(t.body[1].value.values[1].lineno, 3)
-        self.assertEqual(t.body[1].value.values[2].lineno, 3)
+        # TODO: RUSTPYTHON self.assertEqual(t.body[1].value.values[1].lineno, 4)
```
youknowone, line 337: we usually do not comment out test lines, but turn off the test itself.
Could you revert this part and mark `test_ast_line_numbers_multiline_fstring` as expectedFailure?

youknowone (changes requested):
Thank you for contributing. Updating test_fstring seems a big task.

Please check the comment and other commented out test codes too to fit failing tests to our preference.

Lib/test/test_fstring.py
```diff
@@ -427,703 +482,1103 @@ def test_ast_line_numbers_with_parentheses(self):
...
+        self.assertEqual(f"" "", "")
+        self.assertEqual(f"" "" f"", "")
+        self.assertEqual(f"" "" f"" "", "")
+
+        # This is not really [f'{'] + [f'}'] since we treat the inside
+        # of braces as a purely new context, so it is actually f'{ and
+        # then eval('  f') (a valid expression) and then }' which would
+        # constitute a valid f-string.
+        # TODO: RUSTPYTHON SyntaxError
+        # self.assertEqual(f'{' f'}', " f")
```
key262yek, line 701: I can't remove some comments, because it makes SyntaxError before test runs (so it cannot be passed by expectedFailure)

Lib/test/test_fstring.py
```diff
@@ -427,703 +482,1103 @@ def test_ast_line_numbers_with_parentheses(self):
...
+        )
+# TODO: RUSTPYTHON SyntaxError
+#         self.assertEqual(
+#             f"""
+# # this is not a comment
+# { # the following operation it's
+# 3 # this is a number
+# * 2}""",
+#             "\n# this is not a comment\n6",
+#         )
```
key262yek, line 769: this one also

Lib/test/test_fstring.py
```diff
@@ -427,703 +482,1103 @@ def test_ast_line_numbers_with_parentheses(self):
...
+        self.assertEqual(f"{10:#{1}0x}", "       0xa")
+        self.assertEqual(f'{10:{"#"}1{0}{"x"}}', "       0xa")
+        self.assertEqual(f'{-10:-{"#"}1{0}x}', "      -0xa")
+        self.assertEqual(f'{-10:{"-"}#{1}0{"x"}}', "      -0xa")
+        self.assertEqual(f"{10:#{3 != {4:5} and width}x}", "       0xa")
+
+        # TODO: RUSTPYTHON SyntaxError
+        # self.assertEqual(
+        #     f"result: {value:{width:{0}}.{precision:1}}", "result:      12.35"
+        # )
```
key262yek, line 842: also here

Lib/test/test_fstring.py
```diff
@@ -427,703 +482,1103 @@ def test_ast_line_numbers_with_parentheses(self):
...
-                             ])
+    def test_backslashes_in_expression_part(self):
+        # TODO: RUSTPYTHON SyntaxError
+        # self.assertEqual(
+        #     f"{(
+        #                 1 +
+        #                 2
+        # )}",
+        #     "3",
+        # )
```
key262yek, line 1096: and here

Lib/test/test_fstring.py
```diff
@@ -427,703 +482,1103 @@ def test_ast_line_numbers_with_parentheses(self):
...
-        # TODO: RUSTPYTHON self.assertEqual(f'{"""x""" """eric"s""" "y"}', 'xeric"sy')
-        # TODO: RUSTPYTHON self.assertEqual(f'{"""x""" """eric"s""" """y"""}', 'xeric"sy')
-        # TODO: RUSTPYTHON self.assertEqual(f'{r"""x""" """eric"s""" """y"""}', 'xeric"sy')
+    #     # Test concatenation within an expression
+    #     self.assertEqual(f'{"x" """eric"s""" "y"}', 'xeric"sy')
+    #     self.assertEqual(f'{"x" """eric"s"""}', 'xeric"s')
+    #     self.assertEqual(f'{"""eric"s""" "y"}', 'eric"sy')
+    #     self.assertEqual(f'{"""x""" """eric"s""" "y"}', 'xeric"sy')
+    #     self.assertEqual(f'{"""x""" """eric"s""" """y"""}', 'xeric"sy')
+    #     self.assertEqual(f'{r"""x""" """eric"s""" """y"""}', 'xeric"sy')
```
key262yek, line 1277: I think this part is the reason why I commented line by line of broken assertion. 
Only one assertion can be run, and others make SyntaxError

key262yek (commented):
I still struggle with rebase.. 
I made new branch with same name. 
I remove comment from test, but some lines cannot be un-commented because of SyntaxError.

Lib/test/test_fstring.py
youknowone: Ah, that makes sense. Thank you

youknowone (approved):
Lets go with it. We can fix it once parser is changed

PR #5454, opened 2024-12-03
Bump MSRV to 1.83

coolreader18:
This has changes from the past few rust versions that I made when they came out but didn't make a PR for :sweat_smile:


youknowone:
redox build seems requiring older version

coolreader18:
Yeah, I'm hoping they'll update to a recent nightly soon.

youknowone (approved):

Issue #5440, opened 2024-11-12
Prebuilt binaries


sn-o-w:
Hi there!

Do you plan to create prebuilt (automated) binaries in the foreseeable future?

I'm running low on storage space on my computer, and it would be nice to be able to install RustPython easily without needing to build it ourselves. 😊

Thank you in advance.

youknowone:
This is a good suggestion. Could anyone please send a patch of github actions release?

theshubhamp:
@youknowone / @sn-o-w - saw this issue open and sent a PR: https://github.com/RustPython/RustPython/pull/5456

Could you folks take a look ? Cutting a new release on every merge to main for now - do we want to do something different ?

theshubhamp:
Ah, so at the moment release is only creating interpreter binaries without the full stdlib wired in. 

You should be able to download the binary and run it as-is. But packaging in the stdlib needs more work atop this.

arihant2math:
I don't think it is possible to wire the stdlib in until there is an actual installer that installs the executable in a folder with the stdlib and then setting the right variables to point at the stdlib.

arihant2math:
Fixed by #5549.

jayvdb:
#5549 was merged but there are no binaries at https://github.com/RustPython/RustPython/releases/tag/0.5.0

I suspect the problem is there are zero (0) releases at https://github.com/RustPython/RustPython/releases

jayvdb:
cargo-quickinstall has prebuilt binaries at https://github.com/cargo-bins/cargo-quickinstall/releases/tag/rustpython-0.5.0

jayvdb:
https://www.npmjs.com/package/rustpython_wasm is 0.1.0-pre-alpha.2 , published 7 years ago in 2019

jayvdb:
I've published rustpython_wasm and a few platforms of rustpython (with feature freeze-stdlib) at https://github.com/edge-toolkit/core/releases/tag/rp-v1

PR #5456, opened 2024-12-07
Add GitHub Binary Release Pipeline for RustPython

theshubhamp:
Add Release Pipeline for RustPython. Pipeline is currently configured to cut a new release on every merge to main.

Sample Release: https://github.com/theshubhamp/RustPython/releases/tag/2024-12-07-gh-rel-rustpython-11
image

Sample Build: https://github.com/theshubhamp/RustPython/actions/runs/12212320619
image

Validated to run on Linux / Fedora:
image

Issue #5440

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -57,10 +57,10 @@ jobs:
         if: runner.os == 'macOS'
 
       - name: Build RustPython
-        run: cargo build --release --verbose --features=threading ${{ env.CARGO_ARGS }}
+        run: cargo build --release --target=${{ matrix.platform.target }} --verbose --features=threading ${{ env.CARGO_ARGS }}
         if: runner.os == 'macOS'
       - name: Build RustPython
-        run: cargo build --release --verbose --features=threading ${{ env.CARGO_ARGS }},jit
+        run: cargo build --release --target=${{ matrix.platform.target }} --verbose --features=threading ${{ env.CARGO_ARGS }},jit
         if: runner.os != 'macOS'
 
       - name: Rename Binary
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,6 @@
 on:
   push:
-    branches: [gh-rel-rustpython]
+    branches: [main]
 
 name: Release
 
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -18,26 +18,26 @@ jobs:
         platform:
           - runner: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-          - runner: ubuntu-latest
-            target: i686-unknown-linux-gnu
-          - runner: ubuntu-latest
-            target: aarch64-unknown-linux-gnu
-          - runner: ubuntu-latest
-            target: armv7-unknown-linux-gnueabi
-          - runner: ubuntu-latest
-            target: s390x-unknown-linux-gnu
-          - runner: ubuntu-latest
-            target: powerpc64le-unknown-linux-gnu
+#          - runner: ubuntu-latest
+#            target: i686-unknown-linux-gnu
+#          - runner: ubuntu-latest
+#            target: aarch64-unknown-linux-gnu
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,78 @@
...
+
+      - name: Build RustPython
+        run: cargo build --release --verbose --features=threading ${{ env.CARGO_ARGS }}
+        if: runner.os == 'macOS'
+      - name: Build RustPython
+        run: cargo build --release --verbose --features=threading ${{ env.CARGO_ARGS }},jit
+        if: runner.os != 'macOS'
+
+      - name: Rename Binary
+        run: cp target/release/rustpython target/release/rustpython-release-${{ matrix.os }}
```
youknowone, line 42: The OS name is not enough to identify binary target. Could you please also add architecture names to the binary too?

e.g. macOS version can be either x86_64 or aarch64 unless universal build. Linux version can be a lot more variant. Windows version can be x86, x86_64 or arm64.

Here is some platform/architecture matrix
https://github.com/youknowone/python-deadlib/blob/main/.github/workflows/_crypt.yml#L20-L37

youknowone (changes requested):
Thank you so much! This is very helpful. I have a little concern about confusing binary names. Please check the comment.

.github/workflows/release.yml
theshubhamp: Good catch! Overlooked this in haste

Here's the new naming: https://github.com/theshubhamp/RustPython/releases/tag/2024-12-11-gh-rel-rustpython-15
image

And the corresponding file types:
image

I added other targets too (commented out), but they need some extra work to be fully cross compile-able. Fine to do that as a follow up PR ?

theshubhamp:
Thanks for your review @youknowone! I have pushed some new changes. PTAL

theshubhamp:
@youknowone could you re-review when you get some time? Thanks!

.github/workflows/release.yml
```diff
@@ -0,0 +1,105 @@
...
+        if: runner.os == 'macOS'
+      - name: Build RustPython
+        run: cargo build --release --target=${{ matrix.platform.target }} --verbose --features=threading ${{ env.CARGO_ARGS }},jit
+        if: runner.os != 'macOS'
+
+      - name: Rename Binary
+        run: cp target/${{ matrix.platform.target }}/release/rustpython target/rustpython-release-${{ runner.os }}-${{ matrix.platform.target }}
+        if: runner.os != 'Windows'
+      - name: Rename Binary
+        run: cp target/${{ matrix.platform.target }}/release/rustpython.exe target/rustpython-release-${{ runner.os }}-${{ matrix.platform.target }}
```
youknowone, line 70:
```suggestion
        run: cp target/${{ matrix.platform.target }}/release/rustpython.exe target/rustpython-release-${{ runner.os }}-${{ matrix.platform.target }}.exe
```
maybe this is missing

youknowone (commented):
Thank you so much for reminding me, and sorry for being late.

.github/workflows/release.yml
youknowone: Changing this may break Uploading artifact step.

theshubhamp: Apologies for being late.

Thanks for pointing this out! I had assumed this would be run via the shell so should be okay. But good to keep .exe for portability

Made the change:
1. https://github.com/theshubhamp/RustPython/releases/tag/2024-12-26-gh-rel-rustpython-16
2. https://github.com/theshubhamp/RustPython/actions/runs/12505126506/job/34888223838

image

image

theshubhamp: Latest with git tagging fixed: https://github.com/theshubhamp/RustPython/releases/tag/2024-12-26-gh-rel-rustpython-17

theshubhamp:
@youknowone thank you for the review. Could you recheck ?

youknowone (approved):
Looks perfect, thank you so much!

PR #5458, opened 2024-12-09
expose run_shell

### LINKED ISSUE #5432: add some kind of setup hook to `rustpython::run` to allow things like module/variable populating
## Summary

Currently `rustpython:run` only accepts a closure that takes `&VirtualMachine` which is executed very early. During experimenting I find only modules can be added here. Things like setting up global variable from rust world, or pre-importing some modules to workaround "static type has not been initialized" error seems hard if not impossible. I suggest add a way to let user access to `vm` and current scope by adding a "late_setup" parameter, and call it before doing the real work.

Possibly here:

https://github.com/RustPython/RustPython/blob/98d09e781623ccec64643cf5412453cc78ba4f18/src/lib.rs#L184-L186

resulting `run` looks like this:
```
pub fn run(init: impl FnOnce(&mut VirtualMachine) + 'static, late_init: impl FnOnce(&VirtualMachine, &Scope) + 'static) -> ExitCode {
```
## Expected use case

We are creating native modules in rust so we can call our logic in python world. Use existing `rustpython::run` code can be an easy way if we also want interactive environment to quickly experiment with the resulting API.  For now two problems is blocking us:

- we need to add a "World" object to the environment (this is init from rust object) to allow script to use before passed to REPL or user script, so we want `scope.globals.set_item` which seems not accessible now
- there are multiple native modules and sometimes we encounter error: "static type has not been initialized". Just importing them beforehand seems to be a solution if we can

Frefreak:
This exposes `shell::run_shell` as per issue discussion

fixes #5432

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,6 +58,7 @@ use std::process::ExitCode;
 pub use interpreter::InterpreterConfig;
 pub use rustpython_vm as vm;
 pub use settings::{opts_with_clap, InstallPipMode, RunMode};
+pub use shell::run_shell;
 
 /// The main cli of the `rustpython` interpreter. This function will return `std::process::ExitCode`
 /// based on the return code of the python code ran through the cli.


youknowone (approved):
Thank you!

Issue #5460, opened 2024-12-19
Boolean methods cause interpreter panic when presented with the wrong type


Masterchef365:
## Summary
You can cause the latest version of the interpreter ([01f7536](https://github.com/RustPython/RustPython/commit/01f7536b36a94b1ab479253b8227c5151136bfff)) to panic with:
```
thread 'main' panicked at [...]rustpython-vm/src/builtins/bool.rs:212:28:
called `Option::unwrap()` on a `None` value
```
By running the following (invalid) code:
```python
x = min # Some type which isn't a boolean
x |= True # Perform a boolean operation
```
## Expected
I was hoping for an exception to be thrown, which I could possibly catch! Python 3 does this.

## Actual
Panics, see above

## Python Documentation
[TypeError exception](https://docs.python.org/3/library/exceptions.html#TypeError)

theshubhamp:
Looks like the or, and & xor operators implemented in bool builtin were missing type check to int.

Sent a PR including tests.

PR #5461, opened 2024-12-26
Check operand types in bool or, and, xor to be PyInt

theshubhamp:
PyNumber methods are expected to type check both arguments.

Dispatch is not done by inverting parameter order for __r<op>__ (example __ror__) when calls are handled via PyNumberMethods

Other changes:
- Add tests to assert a TypeError is raised when either of the parameter types boolean & other is not convertible to int (str or float for example)
- Sync binary operator order comment with actual implementation

Fixes: #5460

### DIFF
--- a/vm/src/vm/vm_ops.rs
+++ b/vm/src/vm/vm_ops.rs
@@ -151,7 +151,7 @@ impl VirtualMachine {
     /// Calling scheme used for binary operations:
     ///
     /// Order operations are tried until either a valid result or error:
-    ///   b.rop(a,b)[*], a.op(a,b), b.rop(a,b)
+    ///   b.rop(b,a)[*], a.op(a,b), b.rop(b,a)
     ///
     /// [*] only when Py_TYPE(a) != Py_TYPE(b) && Py_TYPE(b) is a subclass of Py_TYPE(a)
     pub fn binary_op1(&self, a: &PyObject, b: &PyObject, op_slot: PyNumberBinaryOp) -> PyResult {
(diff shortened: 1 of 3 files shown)


vm/src/vm/vm_ops.rs
```diff
@@ -151,7 +151,7 @@ impl VirtualMachine {
     /// Calling scheme used for binary operations:
     ///
     /// Order operations are tried until either a valid result or error:
-    ///   b.rop(b,a)[*], a.op(a,b), b.rop(b,a)
+    ///   b.rop(a,b)[*], a.op(a,b), b.rop(a,b)
```
youknowone, line 154: `rop` is a reversed version of `op`. So (b,a) is intended.

youknowone (changes requested):
Good catch! Thank you.
Code changes looks perfect. Please check the comment part.

youknowone (commented):
oh, please also remove `get_py_int` from bool.rs which is not used anymore

vm/src/vm/vm_ops.rs
theshubhamp: Done. Had changed the comment looking at https://github.com/python/cpython/blob/main/Objects/abstract.c#L920

theshubhamp:
Thanks for the review! Made both changes, PTAL

vm/src/vm/vm_ops.rs
youknowone: Oh, thank you! we'd better to follow it later.

youknowone (approved):
👍

PR #5462, opened 2024-12-27
Bump result-like to 0.5.0

crazymerlyn:
Removes one more reason to have `syn` 1.0 as a dependency. After this, the only dependency still using `syn` 1.0 is `syn-ext` used by `rustpython-derive-impl`.


youknowone (approved):
👍

PR #5463, opened 2024-12-28
Temporarily disable release notes generation

theshubhamp:
Release creation fails with this error:
```
HTTP 422: Validation Failed (https://api.github.com/repos/RustPython/RustPython/releases)
body is too long (maximum is 125000 characters)
Error: Process completed with exit code 1.
```

Most likely because there's no previous release to diff of from.

Disabling, getting a green release & enabling back might fix this.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -102,5 +102,4 @@ jobs:
               --repo="$GITHUB_REPOSITORY" \
               --title="RustPython Release $today-$tag #$run" \
               --target="$tag" \
-              --generate-notes \
               bin/rustpython-release-*


theshubhamp:
@youknowone can you please check ? This & another follow up PR might fix the release notes character limit we are currently hitting.

youknowone (approved):
Thank you!

youknowone:
@theshubhamp Releasing binary for every PR merge seems too much. Can this be a cron job like a weekly release? Also, to distinguish from periodic release from stable release, can this always-runnning releases be marked as pre-release?

theshubhamp:
@youknowone yes that makes sense. Had added on push main as a starting point, we should be able to swap that with a schedule.

I'll send a PR with weekly schedule and notes generation enabled. Monday GMT morning sounds fine ?

Do you see stable releases being manual short term ?

theshubhamp:
@youknowone handled in https://github.com/RustPython/RustPython/pull/5466. Could you check ?

PR #5464, opened 2024-12-30
Build wasm on release

arihant2math:
Self-explanatory: adds a job before releasing that builds and uploads the wasm artifact


youknowone (approved):
Thank you!

PR #5465, opened 2024-12-31
Bump rust-caseless to 0.2.2

crazymerlyn:
The latest version prebuilds the table (https://github.com/unicode-rs/rust-caseless/pull/21) thus removing the build step and dependency on regex. RustPython still has regex as a dev dependency but it's no longer part of normal builds, speeding them up.


youknowone (approved):
👍

PR #5468, opened 2025-01-06
Handle pre-release flag being empty for schedule triggers in release workflow

theshubhamp:
Handle pre-release flag being empty for schedule triggers in release workflow

GitHub workflow_dispatch input variables are always empty for other triggers

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -127,12 +127,12 @@ jobs:
           tag: ${{ github.ref_name }}
           run: ${{ github.run_number }}
         run: |
-          if [[ "${{ github.event.inputs.pre-release }}" == "true" ]]; then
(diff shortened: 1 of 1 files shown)


theshubhamp:
@youknowone could you check? Also, does it makes sense to mark all other releases until now pre-release?

theshubhamp:
@youknowone could you check this PR when you have some time?

youknowone (approved):
👍

youknowone:
Sorry for delay. I am currently on trip, so access is limited. I will also mark other releases as pre-release later.

theshubhamp:
Ah! Didn't mean to nag. I assumed you missed the notification

youknowone:
Oh, no problem and I actually appreciate to your reminding.
I dont want to forget about review for weeks like before

PR #5469, opened 2025-01-06
docs(readme): fix installation command for cargo

fu050409:
图片

This repo is monorepo so we should use `cargo install --git https://github.com/RustPython/RustPython rustpython` instead


youknowone (approved):
Thank you!

PR #5470, opened 2025-01-08
guard signal-handling init more broadly

bobmcwhirter:
If `install_signal_handlers` is false (due to embedding), the VM still inits the signal stdlib and installs a lot of signal-handling, including touching *ever* signal, including SIGINT.

When running multiple concurrent interpreters with varying inits at varying times, this can break the hosting application's signal-handling so lovingly
set up before starting anything with RustPython.


youknowone:
Thank you for contributing! You may used wrong email (or unmatching email with your github account) while committing. Could you fix it or confirm if it is intended or not?

coolreader18 (approved):

PR #5472, opened 2025-01-09
Fix warnings for Rust 1.84

coolreader18:


youknowone (approved):
I wish wasm fix be separated from warnings fix.
The changes look good anyway

PR #5473, opened 2025-01-10
Support recursion in JIT-ed functions

theshubhamp:
Support recursion in JIT-ed functions

JIT implementation today is experimental & doesn't support calling out (globals, other functions etc.). This PR adds support for recursion.

Summary of Changes:
1. Added a new JitType FuncRef that stores an id-reference to functions imported for use
2. For recursion, a self import is done & passed down to Bytecode Compiler
3. Recursion only works when return type is annotated

On a "non-scientific" benchmark, JIT-ed function appears to be faster than rustpython/intereprted (expected) and cpython 3.13.1
image

Source of fib.py:
```python
import time
import sys

target = int(sys.argv[1])

def fib(n: int) -> int:
  if n == 0 or n == 1:
    return 1
  fib = 0
  return fib(n-1) + fib(n-2)

start_time = time.perf_counter_ns()
val = fib(target)
time_taken = (time.perf_counter_ns() - start_time)/1000000
print("--- fib(%s)=%s --- no jit --- %s ms ---" % (target, val, time_taken))

try:
    fib.__jit__()

    start_time = time.perf_counter_ns()
    fib(target)
    time_taken = (time.perf_counter_ns() - start_time)/1000000
    print("--- fib(%s)=%s --- jit --- %s ms ---" % (target, val, time_taken))
except:
    print("--- fib(%s)=** --- jit --- not supported ---" % (target))
    pass

```

### DIFF
--- a/vm/src/builtins/function/jitfunc.rs
+++ b/vm/src/builtins/function/jitfunc.rs
@@ -52,7 +52,7 @@ fn get_jit_arg_type(dict: &PyDictRef, name: &str, vm: &VirtualMachine) -> PyResu
             Ok(JitType::Bool)
         } else {
             Err(new_jit_error(
-                "Jit requires argument to be either int or float or bool".to_owned(),
+                "Jit requires argument to be either int, float or bool".to_owned(),
                 vm,
             ))
         }
--- a/vm/src/builtins/function/jitfunc.rs
+++ b/vm/src/builtins/function/jitfunc.rs
@@ -52,7 +52,7 @@ fn get_jit_arg_type(dict: &PyDictRef, name: &str, vm: &VirtualMachine) -> PyResu
             Ok(JitType::Bool)
         } else {
             Err(new_jit_error(
-                "Jit requires argument to be either int or float".to_owned(),
+                "Jit requires argument to be either int or float or bool".to_owned(),
                 vm,
             ))
         }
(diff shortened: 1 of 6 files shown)


theshubhamp:
clippy errors are unrelated & mostly because a new rust stable shipped yesterday: https://releases.rs/docs/1.84.0/

youknowone:
@coolreader18 is working for it on #5472

youknowone:
Could you rebase it to make clippy tests pass?

vm/src/builtins/function/jitfunc.rs
```diff
@@ -52,7 +52,7 @@ fn get_jit_arg_type(dict: &PyDictRef, name: &str, vm: &VirtualMachine) -> PyResu
             Ok(JitType::Bool)
         } else {
             Err(new_jit_error(
-                "Jit requires argument to be either int or float".to_owned(),
+                "Jit requires argument to be either int or float or bool".to_owned(),
```
youknowone, line 55:
```suggestion
                "Jit requires argument to be either int, float or bool".to_owned(),
```

youknowone (approved):
Thank you so much! The benchmark is impressive

theshubhamp:
> Thank you so much! The benchmark is impressive

Welcome! Yes, performed much better than I anticipated

PR #5474, opened 2025-01-11
Update contextlib from CPython 3.12

sleiderr:
Implements changes in the `contextlib` library to match CPython 3.12

### DIFF
--- a/Lib/contextlib.py
+++ b/Lib/contextlib.py
@@ -145,14 +145,17 @@ class _GeneratorContextManager(
             except StopIteration:
                 return False
             else:
-                raise RuntimeError("generator didn't stop")
+                try:
+                    raise RuntimeError("generator didn't stop")
+                finally:
+                    self.gen.close()
         else:
             if value is None:
                 # Need to force instantiation so we can reliably
                 # tell if we get the same exception back
                 value = typ()
             try:
-                self.gen.throw(typ, value, traceback)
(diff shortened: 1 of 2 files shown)


youknowone:
Could you rebase it to make clippy tests pass?

sleiderr:
Hi - I can't seem to reproduce the issue locally (all tests are passing), and I couldn't think of any way the changes in `contextlib` could have impacted the `socket` tests. I also did not find any new change to the `socket` tests in Python 3.12 to adapt to any `contextlib` change.

Could it be a spurious test failure, or should I keep investigating ?

youknowone:
Are you using Linux? The test passes on windows and macos but hanging forever on linux. I wish we can find problem on Linux out of CI. problem only in CI environment is always a headache.

>  I couldn't think of any way the changes in contextlib could have impacted the socket tests

`test_socket` test file imports contextlib library. So they might be related.

Let me run it once again if it was a hidden flaky test.

sleiderr:
I'm on macOS but I tried on a Debian as well. All the stuff using `contextlib` is the `test_socket` should not be impacted by the changes (as they are for specific corner cases).

Anyway, looks like tests are passing now !

youknowone (approved):
Thank you so much for contributing and the additional investigation!
We'd better to mark to skip the hanging test later.

PR #5476, opened 2025-01-14
Update zlib from 3.12.6 and _ZlibDecompressor implementation

arihant2math:
After this is merged a few GZip tests are unblocked.

### DIFF
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -532,6 +532,7 @@ is_wasi = sys.platform == "wasi"
 
 has_fork_support = hasattr(os, "fork") and not is_emscripten and not is_wasi
 
+# From python 3.12.6
 is_s390x = hasattr(os, 'uname') and os.uname().machine == 's390x'
 skip_on_s390x = unittest.skipIf(is_s390x, 'skipped on s390x')
 
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -532,6 +532,9 @@ is_wasi = sys.platform == "wasi"
 
 has_fork_support = hasattr(os, "fork") and not is_emscripten and not is_wasi
 
+is_s390x = hasattr(os, 'uname') and os.uname().machine == 's390x'
+skip_on_s390x = unittest.skipIf(is_s390x, 'skipped on s390x')
+
 def requires_fork():
     return unittest.skipUnless(has_fork_support, "requires working os.fork()")
 
--- a/stdlib/src/zlib.rs
+++ b/stdlib/src/zlib.rs
@@ -718,5 +718,11 @@ mod zlib {
             // }
             ret
         }
+
+        // TODO: Wait for getstate pyslot to be fixed
+        // #[pyslot]
+        // fn getstate(zelf: &PyObject, vm: &VirtualMachine) -> PyResult<PyObject> {
+        //     Err(vm.new_type_error("cannot serialize '_ZlibDecompressor' object".to_owned()))
+        // }
     }
 }
--- a/stdlib/src/zlib.rs
+++ b/stdlib/src/zlib.rs
@@ -47,6 +47,7 @@ mod zlib {
     use libz_sys::{
         Z_BLOCK, Z_DEFAULT_STRATEGY, Z_FILTERED, Z_FINISH, Z_FIXED, Z_HUFFMAN_ONLY, Z_RLE, Z_TREES,
     };
+    use rustpython_vm::function::FuncArgs;
     use rustpython_vm::types::Constructor;
 
     // copied from zlibmodule.c (commit 530f506ac91338)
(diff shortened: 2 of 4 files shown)


coolreader18:
To fix the failing test, you can run `cargo fmt` in the repository.

arihant2math:
I can't seem to see how
```
zlibd = zlib._ZlibDecompressor()
self.assertRaises(TypeError, zlibd.decompress)
```
can produce a `TypeError` when using `#[pymethod]` on the rust side, unless it already does and I'm reading something incorrectly.

youknowone:
Not only Constructor, it may also requires Initializer. Constructor corresponds to `__new__` while Initializer corresponds to `__init__`

stdlib/src/zlib.rs
```diff
@@ -587,4 +588,42 @@ mod zlib {
...
+            zlib_decompressor
+                .into_ref_with_type(vm, cls)
+                .map(Into::into)
+        }
+    }
+
+    #[pyclass]
+    impl ZlibDecompressor {
+        #[pymethod]
+        fn decompress(&self, data: ArgBytesLike, vm: &VirtualMachine) -> PyResult<Vec<u8>> {
```
youknowone, line 616: By this definition, `decompress` always requires an ArgBytesLike argument.
When the argument doesn't fit, it raises TypeError. This article has more explanations how it works:
https://github.com/RustPython/RustPython/wiki/Python-Attributes

arihant2math: Specifically
```
FAIL: test_Constructor (test.test_zlib.ZlibDecompressorTest.test_Constructor)
  File "/Users/runner/work/RustPython/RustPython/pylib/Lib/test/test_zlib.py", line 971, in test_Constructor
    self.assertRaises(TypeError, zlib._ZlibDecompressor, "ASDA")
AssertionError: TypeError not raised by _ZlibDecompressor
```

arihant2math: Even though
```rust
impl Constructor for ZlibDecompressor {
        type Args = ();

        fn py_new(cls: PyTypeRef, _args: Self::Args, vm: &VirtualMachine) -> PyResult {
            let decompress = Decompress::new(true);
            let zlib_decompressor = ZlibDecompressor {
                decompress: PyMutex::new(decompress),
            };
            zlib_decompressor
                .into_ref_with_type(vm, cls)
                .map(Into::into)
        }
    }
 ```

**Comment by arihant2math** [comment]
Additionally `_ZlibDecompressor` has to be unpicklable for another test to pass, but I'm fine with just disabling that one.

File: stdlib/src/zlib.rs (hunk 1 above)
Comment (youknowone) [reply stdlib/src/zlib.rs:617]: In CPython,
```
>>> import zlib
>>> zlib._ZlibDecompressor("ASDA")
TypeError: 'str' object cannot be interpreted as an integer
```
_ZlibDecompressor has to reject str as first argument. ArgBytesLike accepts str.

You seem to miss `wbits` argument, which must be the first argument of `_ZlibDecompressor`

Comment (arihant2math) [reply stdlib/src/zlib.rs]: ah ok that explains a lot, thanks

Comment (arihant2math) [reply stdlib/src/zlib.rs]: However, `type Args = ()`, shouldn't that mean it rejects everything?

Comment (youknowone) [reply stdlib/src/zlib.rs]: I added a commit to fix this problem and the payload error. Constructor was not added to impl `_ZlibDecompressor`.

Comment (arihant2math) [reply stdlib/src/zlib.rs]: Thank you so much, that fixed everything.

**Comment by arihant2math** [comment]
Also it seems to my knowledge `__getstate__` isn't a magic method I can use? It would be nice to use it because then I could pass the pickling test rather trivially.

**Comment by youknowone** [comment]
Each class need to implement `__getstate__` to support pickling. pickle library will use getstate in priority. Searching for `fn getstate` will result other types' implementation

**Review by youknowone — COMMENTED** [review]
would you leave a comment which version of CPython did you brought python source from?

File: Lib/test/support/__init__.py [hunk 2]
```diff
@@ -532,6 +532,9 @@ def requires_legacy_unicode_capi():
 
 has_fork_support = hasattr(os, "fork") and not is_emscripten and not is_wasi
 
+is_s390x = hasattr(os, 'uname') and os.uname().machine == 's390x'
+skip_on_s390x = unittest.skipIf(is_s390x, 'skipped on s390x')
```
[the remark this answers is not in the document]
arihant2math: Yes, it's used in the tests.
In the future, a PR to update everything in test/support/ might be something to consider to cut down on this

Lib/test/support/__init__.py
```diff
@@ -532,6 +532,10 @@ def requires_legacy_unicode_capi():
 
 has_fork_support = hasattr(os, "fork") and not is_emscripten and not is_wasi
 
+# From python 3.12.6
```
youknowone, line 535: I am sorry to confuse you. Version was required for commit message, but this is also fine too

youknowone (approved):
Thank you so much!

PR #5478, opened 2025-01-15
Update Logging to 3.12.7

arihant2math:
Turns out logging tests were not being tested, this pr adds them too. This also adds smtplib because that is needed for logging tests


arihant2math:
All tests pass

/cc @youknowone

youknowone (changes requested):
Thank you so much! Everything in the code looks good.
Is this also from CPython 3.12.6? Leaving a specific import version on PR title is helpful to track the current RustPython library versions when we update it to next version.

arihant2math:
Thanks, it's 3.12.7 this time.

youknowone (approved):
👍

PR #5480, opened 2025-01-16
Update gzip & tarfile & xmlrpc & pydoc + associated tests to 3.12.7

arihant2math:
This PR allows the gzip tests to pass at a significantly higher rate thanks to #5476.

This PR has ballooned in size due to random dependencies, but it was originally supposed to be just gzip.
To keep things manageable, I have forgon adding tests for pydoc, I'll follow this PR with them instead.
It also happens to edit parts of the zlib module to get some functionality I forgot to add with #5476


arihant2math:
Somehow by fixing gzip, tarlib and shutil have been broken

arihant2math:
Probably will target 3.13 instead.

Issue #1217, opened 2019-08-07
Implement the ctypes module


windelbouwman:
The ctypes module is pretty interesting. It allows people to load arbitrary dll's into python. This could probably be implemented by using some rust crate which does the same.

This is a nice challenge!

cactorium:
I'm down for trying to implement this one! It looks like `libloading` is probably a good place to start

darleybarreto:
I dont know if it is a good idea to do something like PyPy's [clibffi](https://foss.heptapod.net/pypy/pypy/-/blob/branch/default/rpython/rlib/clibffi.py) and [ _cffi_backend ](https://foss.heptapod.net/pypy/pypy/-/tree/branch/default/pypy/module/_cffi_backend) in the long term. Perhaps to have something kinda working in the near future.

youknowone:
#2364

ANVLISTENER:
how do we include rust in rustpython. please try to illustrate.

Issue #4739, opened 2023-03-22
import ctypes ModuleNotFoundError: No module named 'ctypes'


ParkourLiu:
![image](https://user-images.githubusercontent.com/33681340/226832483-c1f3604f-124c-44c8-bcdf-98d8cb77d4d9.png)

Hello,I want to import 'ctypes',
 am I using the wrong method?

Unable to import 'ctypes' module

ParkourLiu:
![image](https://user-images.githubusercontent.com/33681340/226833252-2d2bd6a4-f534-4261-8191-b19df2bc4ea8.png)

youknowone:
see also #1217 
it is not implemented yet.

youknowone:
And you also didn't correctly set RUSTPYTHONPATH as error message describes.

ParkourLiu:
Thank you for your answer,Are there plans to support it in the future?

youknowone:
I guess so. but not sure about schedule. ctypes is a huge task and our capacity is limited. in general, we need volunteer for new features.

Issue #2671, opened 2021-05-24
--install-pip fails on Windows


oroppas:
```
C:\msys64\home\ryuta\packages\rustpython\rustpython>cargo run --features ssl --release -- --install-pip
    Finished release [optimized] target(s) in 0.69s
     Running `target\release\rustpython.exe --install-pip`
downloading get-pip.py...
Traceback (most recent call last):
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\urllib\request.py", line 1351, in do_open
    encode_chunked=req.has_header('Transfer-encoding'))
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\http\client.py", line 1239, in request
    self._send_request(method, url, body, headers, encode_chunked)
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\http\client.py", line 1285, in _send_request
    self.endheaders(body, encode_chunked=encode_chunked)
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\http\client.py", line 1234, in endheaders
    self._send_output(message_body, encode_chunked=encode_chunked)
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\http\client.py", line 1026, in _send_output
    self.send(msg)
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\http\client.py", line 964, in send
    self.connect()
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\http\client.py", line 1400, in connect
    server_hostname=server_hostname)
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\ssl.py", line 507, in wrap_socket
    session=session
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\ssl.py", line 1043, in _create
    raise
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\ssl.py", line 1040, in _create
    self.do_handshake()
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\ssl.py", line 1311, in do_handshake
    self.settimeout(timeout)
  File "C:\msys64\home\ryuta\packages\rustpython\rustpython\Lib\ssl.py", line 1309, in do_handshake
    self._sslobj.do_handshake()
[...]

PR #5488, opened 2025-01-20
ctypes implementation

arihant2math:
Continuation of #2364. Attempts to just get everything in a building and mergable state.

Fixes: #1217 #2671 #4739


arihant2math:
I've come to the conclusion this implementation is too old to port over, I'll have to start from scratch.

Issue #4668, opened 2023-03-08
gc with cyclic collector?


youknowone:
https://github.com/fitzgen/bacon-rajan-cc
https://github.com/artichoke/cactusref
https://github.com/chc4/samsara

see also https://redvice.org/2023/samsara-garbage-collector/

darleybarreto:
Recently I've found this one called [dumpster](https://github.com/claytonwramsey/dumpster). `Gc` type behaves similar to `Rc`, but the GC can work concurrently if one chooses. What do you think? Would a massive rewrite be needed?

youknowone:
I didn't compare the candidates yet. 
Anyone that doesn't hurt basic RC performance that much will be the best.

arihant2math:
We unfortunately cannot use [https://github.com/artichoke/cactusref](https://github.com/artichoke/cactusref), as it is nightly only.

PR #5493, opened 2025-01-28
Partial Garbage Collection Implementation

arihant2math:
This pr aims to be completed from scratch. It benefits from `Traverse` already existing.
This currently is a tracing stop-the-world gc, mainly because incremental garbage collecting is hard.
The general structure is based off of #4180 but the algorithm is not, as that pr had a blocking bug in it's algorithm.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ bz2 = ["stdlib", "rustpython-stdlib/bz2"]
 sqlite = ["rustpython-stdlib/sqlite"]
 ssl = ["rustpython-stdlib/ssl"]
 ssl-vendor = ["ssl", "rustpython-stdlib/ssl-vendor"]
+gc = ["rustpython-vm/gc"]
 
 [dependencies]
 rustpython-compiler = { workspace = true }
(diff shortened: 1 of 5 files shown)


youknowone:
Are you also interested in RC too? I wish we can try circ for RustPython  because it fits well for RustPython's requirements.

https://crates.io/crates/circ

About GC, if we can use other library, that will be the best. I didn't compare the candidatse, but #4668 has a few links

arihant2math:
> About GC, if we can use other library, that will be the best. I didn't compare the candidatse, but #4668 has a few links
I plan to try to use another crate for the actual meat of the algorithm as the main reason the previous PR failed was having a bug in the algorithm. I will look into the ones in the issue.

arihant2math:
The main issue with using an external library is that we don't actually use `Rc<T>` or `Arc<T>`, making porting difficult.

youknowone:
When i looked in circ, it had internal crate ebr_impl. So it was adaptable for our own rc implementation.
Not sure it also fit for gc, not only rc

arihant2math:
I think for now I will try pulling and running a modified GC algorithm from dumpster mostly because it does gc.

arihant2math:
Ok so I've tried two approaches.

The first, which I didn't push up tried to simply port over the code from the previous pr, but that ran into tons of compiler errors.
The second, which I did push up tried to copy over parts of `dumpster`, and is facing less compile errors, but is missing some key parts.

PR #5482, opened 2025-01-18
Mark version 3.13

youknowone:

PR #5495, opened 2025-02-01
Mark version 3.13.0

arihant2math:
Supersedes #5482 

Also:
- Removes a difflib doctest (this can be added back during a difflib update)
- Strips docstrings like cpython does
- Marks some tests that got activated during 3.13 in `test_enum` as invalid.
- Updates `extra_tests` to pass with cpython 3.13


youknowone:
I will immediately merge separately mergable changes like removing imghdr,sndhdr etc

arihant2math:
@youknowone everything works except some modules, should I disable the tests or update the modules

Lib/test/test_enum.py
```diff
@@ -1369,10 +1369,8 @@ class Inner(Enum):
             [Outer.a, Outer.b, Outer.Inner],
             )
 
-    @unittest.skipIf(
-            python_version < (3, 13),
-            'inner classes are still members',
-            )
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
youknowone, line 1373: Let's keep the original skipIf for 3.13 and add RUSTPYTHON expected failure.
Otherwise the removed lines have to be added again when we update it to 3.13 later

extra_tests/snippets/builtin_object.py
```diff
@@ -7,10 +7,6 @@ class MyObject:
 assert myobj == myobj
 assert not myobj != myobj
 
-object.__subclasshook__() == NotImplemented
-object.__subclasshook__(1) == NotImplemented
```
youknowone, line 11: Line 10 and 12 are less useful. But Line 11 still is valid.

compiler/codegen/src/compile.rs
```diff
@@ -3356,7 +3356,33 @@ fn split_doc<'a>(
...
+                                    }
+                                })
+                                .collect::<Vec<String>>()
+                                .join("\n")
+                        } else {
+                            doc
+                        }
+                    }),
+                    body_rest,
+                );
```
youknowone, line 3385: The  behavior is a bit weird. Please add a test using below code:

```python

def f1():
    """
        x 
    \ty
    """
print(repr(f1.__doc__))

def f2():
    """
\t    x 
\t\ty
    """

print(repr(f2.__doc__))
```

```sh
$ python3.13 x.py
'\nx \ny\n'
'\nx \n    y\n'
```

extra_tests/snippets/stdlib_subprocess.py
```diff
@@ -27,43 +27,8 @@ def sleep(secs):
 assert p.poll() == 0
 assert p.returncode == 0
 
-p = subprocess.Popen(sleep(2))
```
youknowone, line 30: Removing these tests looks like too much loss of test. Is subprocess totally broken?

vm/src/version.rs
```diff
@@ -4,10 +4,10 @@
 use chrono::{prelude::DateTime, Local};
 use std::time::{Duration, UNIX_EPOCH};
 
-// = 3.12.0alpha
+// = 3.13.1alpha
 pub const MAJOR: usize = 3;
-pub const MINOR: usize = 12;
-pub const MICRO: usize = 0;
+pub const MINOR: usize = 13;
+pub const MICRO: usize = 1;
```
youknowone, line 10:
```suggestion
pub const MICRO: usize = 0;
```
we usually keep this zero because we always have release level alpha. This number doesn't affect python libraries' behavior. So 0 is always fine.

youknowone (changes requested):

compiler/codegen/src/compile.rs
arihant2math: Fixed this by switching over to a port of the cpython implementation

extra_tests/snippets/stdlib_subprocess.py
arihant2math: Those tests failed and I couldn't seem to debug it, maybe it is a windows only issue.

extra_tests/snippets/builtin_object.py
arihant2math: 10 and 12 were type errors instead, which conforms to cpython a little better I think

extra_tests/snippets/stdlib_subprocess.py
youknowone: Unlike difflib, this is a core part of python. Let's try to fix it.

arihant2math:
Marking as draft as the docstring whitespace stripping isn't passing all tests.

compiler/codegen/src/compile.rs
```diff
@@ -3368,17 +3368,57 @@ impl EmitArg<bytecode::Label> for ir::BlockIdx {
...
+    // First pass: find minimum indentation of any non-blank lines
+    // after first line.
+    let margin = doc
+        .lines()
+        // Skip the first line as per cpython impl
+        .skip(1)
+        // Skip empty lines
+        .filter(|line| !line.replace('\r', "").is_empty())
+        // Get the 1st line
+        .next()
```
fanninpm, line 3385:
```suggestion
        // Get the 1st non-empty line
        .find(|line| !line.replace('\r', "").is_empty())
```

compiler/codegen/src/compile.rs
youknowone: Please add this test under `extra_tests/snippets/syntax_doc.py` or any other proper place. Everything others looks good.
Thank you so much!

arihant2math: Somehow the first test breaks it; I'll see if this can be fixed

arihant2math: Found the issue: expandtabs isn't as simple as `.replace('\t', '        ')`. 😄

arihant2math:
Ready for merge!

youknowone (approved):
Thank you so much!

arihant2math:
🎉

PR #5497, opened 2025-02-03
Pin malachite versions to avoid API incompatibility

youknowone:
https://github.com/RustPython/malachite-bigint/pull/7


youknowone:
No idea how this still fails on Ubuntu, but better than before

PR #5502, opened 2025-02-04
Fix panic

### LINKED ISSUE #5501: Panic in codecs.rs
Windows x86-64 debug build with default features when running `cargo run -- -m test`.

```
thread 'main' panicked at vm\src\codecs.rs:626:12:
attempt to subtract with overflow
error: process didn't exit successfully: `target\debug\rustpython.exe -m test` (exit code: 101)
```

arihant2math:
Fixes #5501

### DIFF
--- a/vm/src/codecs.rs
+++ b/vm/src/codecs.rs
@@ -623,7 +623,7 @@ fn surrogatepass_errors(err: PyObjectRef, vm: &VirtualMachine) -> PyResult<(PyOb
         // Try decoding a single surrogate character. If there are more,
         // let the codec call us again.
         let p = &s.as_bytes()[range.start..];
-        if p.len() - range.start >= byte_length {
+        if p.len().saturating_sub(range.start) >= byte_length {
             match standard_encoding {
                 StandardEncoding::Utf8 => {
                     if (p[0] as u32 & 0xf0) == 0xe0


vm/src/codecs.rs
```diff
@@ -623,7 +623,7 @@ fn surrogatepass_errors(err: PyObjectRef, vm: &VirtualMachine) -> PyResult<(PyOb
         // Try decoding a single surrogate character. If there are more,
         // let the codec call us again.
         let p = &s.as_bytes()[range.start..];
-        if p.len() - range.start >= byte_length {
+        if (p.len() as isize) - range.start as isize >= byte_length as isize {
```
youknowone, line 626:
```suggestion
        if p.len() >= range.start + byte_length {
```
How about this? is this also panic?

arihant2math: isize should not overflow downwards, I tested it and it fixed the issue.

youknowone: isize casting is usually not an issue on 64bit architecture but sometimes on 32bit one by limiting it half size. We can avoid it when possible. I will take current changes if removing subtraction doesn't fix problem.

arihant2math: I switched to saturating subtraction, hopefully that works as well.

arihant2math:
I think it should be at a better solution at this point.

coolreader18 (approved):
LGTM

PR #5486, opened 2025-01-19
Complement TypeAliasType

moreal:
## Current PR Status

Currently, this pr stuck at the part to implement new syntax introduced at 3.13 syntax.

----

This pull request does:

 - Bump `typing` module to `3.13.1`.
   - Implement `_typing.NoDefault`
 - Complement `TypeAliasType`
   - Set `PyTypeFlags::IMMUTABLETYPE` for `TypeAliasType`.
   - Correct `TypeAliasType.__repr__` implementation.

## References

 - `TypeAliasType` flags: https://github.com/python/cpython/blob/067145177975eadd61a0c907d0d177f7b6a5a3de/Objects/typevarobject.c#L1777

PR #5505, opened 2025-02-06
Complement vm logics related with `ParamSpec` and `TypeVarTuple`

moreal:
This pull request complements compiler, vm logics related [PEP-0695](https://peps.python.org/pep-0695/).

### DIFF
--- a/compiler/codegen/src/compile.rs
+++ b/compiler/codegen/src/compile.rs
@@ -1095,8 +1095,27 @@ impl Compiler {
                         self.store_name(name.as_ref())?;
                     }
                 }
-                located_ast::TypeParam::ParamSpec(_) => todo!(),
-                located_ast::TypeParam::TypeVarTuple(_) => todo!(),
+                located_ast::TypeParam::ParamSpec(located_ast::TypeParamParamSpec {
+                    name, ..
+                }) => {
+                    self.emit_load_const(ConstantData::Str {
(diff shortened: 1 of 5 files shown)


youknowone (commented):
Thank you for contributing!
Is there any related test about this changes in test_typing?

(okay if there is nothing or not the full test is passed yet)

moreal:
@youknowone [PEP-0695](https://peps.python.org/pep-0695/) is applied since Python 3.12  and [current typing module (on main branch)](https://github.com/RustPython/RustPython/blob/3de1c2ab56fed2c3ca059a6ea17c4565ee2e66d1/Lib/typing.py) is on Python 3.10.5. So I guess there is no related tests about this pull request.

Also, this pull request is originally part of #5486 but I am separating this commit from the PR for ease of review. There are more works for marking #5486 as ready-to-review 😞 

```
error: Python compile error from Lib/test/test_typing.py: cannot use starred expression here at row 1081 col 25
  --> pylib/src/lib.rs:14:35
   |
14 |     rustpython_derive::py_freeze!(dir = "./Lib", crate_name = "rustpython_compiler_core");
   |                                   ^^^^^^^^^^^^^

error: could not compile `rustpython-pylib` (lib) due to 1 previous error
```

youknowone (approved):
Oh, got it. Thank you for explanation!

PR #5509, opened 2025-02-10
Implement `del object.__dict__`

key262yek:
I add `object::delete_dict()` which changes Object's dict to `None`, and add `PySetterValue::Delete` match case in `object::object_set_dict`. 
All things fine I think, but minor format issue remains. 

```rust
    pub fn delete_dict(&self) -> Result<(), ()> {
        match self.instance_dict() {
            Some(_) => {
                unsafe {
                    let ptr = self as *const _ as *mut PyObject;
                    (*ptr).0.dict = None;
                }
                Ok(())
            }
            None => Err(()),
        }
    }
```

In this code, since Err can be returned, we should define custom Error type.
I made several options for it.

1. Return `Ok(())` when it does not have __dict__ already
  - Can it satisfy normal convention? or I should raise error when the object to be deleted does not exist?
2. Make simple error type and return it. 
3. Return Option not Result
4. Add something to `delete_dict`'s parameter and return it. 

I cannot decide which one is better. Can you suggest about it?

### DIFF
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -1288,7 +1288,7 @@ fn subtype_set_dict(obj: PyObjectRef, value: PyObjectRef, vm: &VirtualMachine) -
             descr_set(&descr, obj, PySetterValue::Assign(value), vm)
         }
         None => {
-            object::object_set_dict(obj, PySetterValue::Delete, vm)?;
+            object::object_set_dict(obj, PySetterValue::Assign(value.try_into_value(vm)?), vm)?;
             Ok(())
         }
     }
--- a/vm/src/builtins/type.rs
+++ b/vm/src/builtins/type.rs
@@ -1288,7 +1288,7 @@ fn subtype_set_dict(obj: PyObjectRef, value: PyObjectRef, vm: &VirtualMachine) -
             descr_set(&descr, obj, PySetterValue::Assign(value), vm)
         }
         None => {
-            object::object_set_dict(obj, value.try_into_value(vm)?, vm)?;
+            object::object_set_dict(obj, PySetterValue::Delete, vm)?;
             Ok(())
         }
     }
--- a/vm/src/function/getset.rs
+++ b/vm/src/function/getset.rs
@@ -36,8 +36,10 @@ where
 {
     #[inline]
     fn from_setter_value(vm: &VirtualMachine, obj: PySetterValue) -> PyResult<Self> {
-        let obj = obj.ok_or_else(|| vm.new_type_error("can't delete attribute".to_owned()))?;
-        T::try_from_object(vm, obj)
+        match obj {
+            PySetterValue::Assign(obj) => T::try_from_object(vm, obj),
+            PySetterValue::Delete => T::try_from_object(vm, vm.ctx.none()),
+        }
     }
 }
 
--- a/vm/src/object/core.rs
+++ b/vm/src/object/core.rs
@@ -25,6 +25,7 @@ use crate::{
         lock::{PyMutex, PyMutexGuard, PyRwLock},
         refcount::RefCount,
     },
+    function::PySetterValue,
     vm::VirtualMachine,
 };
 use itertools::Itertools;
@@ -707,26 +708,21 @@ impl PyObject {
 
     /// Set the dict field. Returns `Err(dict)` if this object does not have a dict field
     /// in the first place.
-    pub fn set_dict(&self, dict: PyDictRef) -> Result<(), PyDictRef> {
-        match self.instance_dict() {
-            Some(d) => {
+    pub fn set_dict(&self, dict: PySetterValue<PyDictRef>) -> Result<(), PyDictRef> {
+        match (self.instance_dict(), dict) {
+            (Some(d), PySetterValue::Assign(dict)) => {
                 d.set(dict);
                 Ok(())
             }
-            None => Err(dict),
-        }
-    }
-
-    pub fn delete_dict(&self) -> Result<(), ()> {
-        match self.instance_dict() {
-            Some(_) => {
+            (None, PySetterValue::Assign(dict)) => Err(dict),
+            (Some(_), PySetterValue::Delete) => {
                 unsafe {
                     let ptr = self as *const _ as *mut PyObject;
                     (*ptr).0.dict = None;
                 }
                 Ok(())
             }
-            None => Err(()),
+            (None, PySetterValue::Delete) => Ok(()),
         }
     }
 
--- a/vm/src/object/core.rs
+++ b/vm/src/object/core.rs
@@ -717,6 +717,19 @@ impl PyObject {
         }
     }
 
+    pub fn delete_dict(&self) -> Result<(), ()> {
+        match self.instance_dict() {
+            Some(_) => {
+                unsafe {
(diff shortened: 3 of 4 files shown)


vm/src/object/core.rs
```diff
@@ -717,6 +717,19 @@ impl PyObject {
         }
     }
 
+    pub fn delete_dict(&self) -> Result<(), ()> {
```
youknowone, line 720: set_dict in core.rs is only used for object_set_dict. Rather than adding delete_dict, I prefer to add this logic into `set_dict` if other problems not found.

youknowone (commented):
Thanks!

key262yek:
There are few things make the problem complex. 

1. Affect multiple attribute in getset.rs
After change following code,
```rust
impl<T> FromPySetterValue for T
where
    T: Sized + TryFromObject,
{
    #[inline]
    fn from_setter_value(vm: &VirtualMachine, obj: PySetterValue) -> PyResult<Self> {
        match obj {
            PySetterValue::Assign(obj) => T::try_from_object(vm, obj),
            PySetterValue::Delete => T::try_from_object(vm, vm.ctx.none()),
        }
    }
}
```

This test `Lib/test/test_exceptions.py:620 testInvalidAttrs` fails
```python
def testInvalidAttrs(self):
        self.assertRaises(TypeError, setattr, Exception(), '__cause__', 1)
        self.assertRaises(TypeError, delattr, Exception(), '__cause__')
        self.assertRaises(TypeError, setattr, Exception(), '__context__', 1)
        self.assertRaises(TypeError, delattr, Exception(), '__context__')
```
This is because FromPySetterValue trait affect not only __dict__ but also other attributes.

vm/src/builtins/type.rs
```diff
@@ -1288,7 +1288,7 @@ fn subtype_set_dict(obj: PyObjectRef, value: PyObjectRef, vm: &VirtualMachine) -
             descr_set(&descr, obj, PySetterValue::Assign(value), vm)
         }
         None => {
-            object::object_set_dict(obj, value.try_into_value(vm)?, vm)?;
+            object::object_set_dict(obj, PySetterValue::Assign(value.try_into_value(vm)?), vm)?;
```
key262yek, line 1291: 2. Cannot resolve `Lib/test/test_descr.py:3399 test_set_dict`
If I don't change code like this, `object_set_dict` fails to be compiled because now value must be the type `PySetterValue<PyDictRef>`.
There are two option for it, pass the value `PySetterValue::Assign` or `PySetterValue::Delete`

For the case of Assign I get
```bash
ERROR: test_set_dict (__main__.ClassPropertiesAndMethods.test_set_dict)
  File "Lib/test/test_descr.py", line 3415, in test_set_dict
    del a.__dict__ # Deleting __dict__ is allowed
TypeError: Expected type 'dict' but 'NoneType' found.
```

For Delete case,
```bash
ERROR: test_set_dict (__main__.ClassPropertiesAndMethods.test_set_dict)
  File "Lib/test/test_descr.py", line 3404, in test_set_dict
    self.assertEqual(a.b, 1)
AttributeError: 'C' object has no attribute 'b'
```


Test function
```python
def test_set_dict(self):
        # Testing __dict__ assignment...
        class C(object): pass
        a = C()
        a.__dict__ = {'b': 1}
        self.assertEqual(a.b, 1)
        def cant(x, dict):
            try:
                x.__dict__ = dict
            except (AttributeError, TypeError):
                pass
            else:
                self.fail("shouldn't allow %r.__dict__ = %r" % (x, dict))
        cant(a, None)
        cant(a, [])
        cant(a, 1)
        del a.__dict__ # Deleting __dict__ is allowed
```

key262yek (commented):
this problem is more complex than I thought at first.

vm/src/function/getset.rs
```diff
@@ -36,8 +36,10 @@ where
 {
     #[inline]
     fn from_setter_value(vm: &VirtualMachine, obj: PySetterValue) -> PyResult<Self> {
-        let obj = obj.ok_or_else(|| vm.new_type_error("can't delete attribute".to_owned()))?;
-        T::try_from_object(vm, obj)
+        match obj {
+            PySetterValue::Assign(obj) => T::try_from_object(vm, obj),
+            PySetterValue::Delete => T::try_from_object(vm, vm.ctx.none()),
+        }
```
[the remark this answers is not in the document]
key262yek: I found only one way to restrict deletion of attribute except `__dict__` by editing following part
```rust
// vm/src/builtins/getset.rs: 97
     #[pyslot]
    fn descr_set(
        zelf: &PyObject,
        obj: PyObjectRef,
        value: PySetterValue<PyObjectRef>,
        vm: &VirtualMachine,
    ) -> PyResult<()> {
        let zelf = zelf.try_to_ref::<Self>(vm)?;
        if let Some(ref f) = zelf.setter {
            match value {
                PySetterValue::Assign(v) => f(vm, obj, PySetterValue::Assign(v)),
                PySetterValue::Delete if zelf.name == "__dict__" => {
                    f(vm, obj, PySetterValue::Delete)
                }
                _ => Err(vm.new_type_error("can't delete attribute".to_owned())),
            }
        } else {
            Err(vm.new_attribute_error(format!(
                "attribute '{}' of '{}' objects is not writable",
                zelf.name,
                obj.class().name()
            )))
        }
    }
```

vm/src/builtins/type.rs
key262yek: Clarified problem is this.

If I try to delete `__dict__`, following methods are called. 

1. (getset.rs:164) set - `PySetterValue::Delete` becomes `None`
```rust
impl<F, T, V, R> IntoPySetterFunc<(OwnedParam<T>, V, R, VirtualMachine)> for F
where
    F: Fn(T, V, &VirtualMachine) -> R + 'static + Send + Sync,
    T: TryFromObject,
    V: FromPySetterValue,
    R: IntoPyNoResult,
{
    fn set(&self, obj: PyObjectRef, value: PySetterValue, vm: &VirtualMachine) -> PyResult<()> {
        let obj = T::try_from_object(vm, obj)?;
        let value = V::from_setter_value(vm, value)?;  // None = V::from_setter_value(vm, PySetterValue::Delete)?;
        (self)(obj, value, vm).into_noresult()  // subtype_set_dict(obj, None, vm)
    }
}
```
2. (type.rs:1275) subtype_set_dict - value should be PySetterValue again. 
```rust
fn subtype_set_dict(obj: PyObjectRef, value: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
    let cls = obj.class();
    match find_base_dict_descr(cls, vm) {
        Some(descr) => {
            let descr_set = descr
                .class()
                .mro_find_map(|cls| cls.slots.descr_set.load())
                .ok_or_else(|| {
                    vm.new_type_error(format!(
                        "this __dict__ descriptor does not support '{}' objects",
                        cls.name()
                    ))
                })?;
            descr_set(&descr, obj, PySetterValue::Assign(value), vm)
        }
        None => {
            object::object_set_dict(obj, PySetterValue::Assign(value.try_into_value(vm)?), vm)?;
            Ok(())
        }
    }
}
```

Since the first `V::from_setter_value(vm, value)` give same result for `value = PySetterValue::Assign(None)` and `value = PySetterValue::Delete` to None, This kind of error occurs.

PR #5511, opened 2025-02-10
winreg upgrade to 0.55

arihant2math:


arihant2math:
I'm just going to let this merge as is. [https://github.com/microsoft/windows-rs/issues/3093](https://github.com/microsoft/windows-rs/issues/3093) is turning out to be annoying to fix as every time an `isize` is passed in it wants a `*mut cvoid` instead.

For future reference, the "proper" way to do this (to my knowledge) is cast the isize to a `cint`, which can then be cast to the `cvoid`.

coolreader18 (approved):
LGTM, after the clippy lint is fixed.

arihant2math:
Fixed.

PR #5513, opened 2025-02-11
Update openssl to fix possible vulnerability

coolreader18:
I don't think we were actually affected, but better safe than sorry: https://github.com/sfackler/rust-openssl/security/advisories/GHSA-rpmj-rpgj-qmpm

close #5498


youknowone (approved):
👍

PR #5514, opened 2025-02-12
Dependency bumps

arihant2math:


Cargo.lock
```diff
@@ -967,6 +961,15 @@ dependencies = [
  "either",
 ]
 
+[[package]]
+name = "itertools"
+version = "0.12.1"
```
youknowone, line 966: This change makes itertools fragmentation

youknowone (approved):
👍

Cargo.lock
arihant2math: unfortunately I can't bump itertools any higher, because of some deprecation warnings, I'll send another pr to fix this later

PR #5515, opened 2025-02-12
Syn 2.0

arihant2math:
There are two main things that need to be done to get this working

1. `syn_ext` is for syn 1.0, this dependency needs to be updated
2. `syn::meta` has been redone, so all code relating to that has now got to be redone as well.


coolreader18:
I opened https://github.com/youknowone/syn-ext/pull/7 for this, which means that we can update syn while leaving most of our code unchanged, and then incrementally port it to `syn::meta`.

coolreader18:
Here's my impl of this: https://github.com/RustPython/RustPython/pull/5556

coolreader18:
Superseded by #5556

PR #5517, opened 2025-02-13
Importlib upgrade to 3.13.2

arihant2math:


youknowone:
#5500

arihant2math:
Current todo list before this can be retried:

- [ ] Fix `cargo run -r -- -m test .` on windows (it fails right now).
- [ ] Incrementally update `test_support` to 3.13

youknowone:
Thank you for this series of patches. We now targets 3.14. So this is outdated.

PR #5520, opened 2025-02-14
Add _colorize at 3.13.2

arihant2math:
Needed for #5517. I added a dummy for `nt._supports_virtual_terminal` which can be fixed later, but I can't think of a scenario where it'd be false.


youknowone (approved):
👍

PR #5527, opened 2025-02-17
Update rand to 0.9

coolreader18:
`getrandom` is kinda messy now, but hopefully malachite removes its transitive dependency on `getrandom 0.2` soon and things can get simpler.


youknowone:
we are using very recent version of malachite. if we have problem, we have to report it.

youknowone (approved):

PR #5530, opened 2025-02-20
_ctypes pt. 3

arihant2math:
Here comes the meat of the code and the hardest part implementation-wise.
The main goal is to implement (atleast partially) `CFuncPtr`. By partially, I mean that not supporting arguments to functions is probably fine. At any rate the cpython implementation can be used as a reference since they use libffi. The main challenge is determining the return type of the loaded function from the `cvoid` it'll return (unless we don't have to do that for some reason).

- [x] Use libffi to manage functions (Can't pass in arbitrary arguments to functions through libloading)
- [x] Constructor for `CFuncPtr`
- [x] Give inner function class the ability to load functions
- [x] Give it the ability to call them and get the return
- [x] Cast the return properly
- [x] Callable for `CFuncPtr` that encapsulated the entire thing

The next pr can probably just do some cleanup on this and get the basics for struct and union return types working.

# Known issues
- Arguments cause errors to be thrown
- Return type of functions is `c_int` and cannot be changed
- Error handling isn't great.
- `import ctypes` fails
- creating string buffers is broken

### DIFF
--- a/vm/src/stdlib/ctypes/function.rs
+++ b/vm/src/stdlib/ctypes/function.rs
@@ -1,11 +1,10 @@
 use crate::builtins::{PyStr, PyTuple, PyTypeRef};
-use crate::class::StaticType;
 use crate::convert::ToPyObject;
 use crate::function::FuncArgs;
 use crate::stdlib::ctypes::PyCData;
 use crate::stdlib::ctypes::base::{PyCSimple, ffi_type_from_str};
 use crate::types::{Callable, Constructor};
-use crate::{AsObject, Py, PyObjectRef, PyResult, VirtualMachine};
+use crate::{Py, PyObjectRef, PyResult, VirtualMachine};
 use crossbeam_utils::atomic::AtomicCell;
 use libffi::middle::{Arg, Cif, CodePtr, Type};
 use libloading::Symbol;
@@ -37,7 +36,7 @@ impl Function {
     ) -> PyResult<Self> {
         // map each arg to a PyCSimple
         let args = args
-            .into_iter()
+            .iter()
             .map(|arg| {
                 if let Some(data) = arg.downcast_ref::<PyCSimple>() {
                     Ok(ffi_type_from_str(&data._type_).unwrap())
@@ -47,7 +46,7 @@ impl Function {
             })
             .collect::<PyResult<Vec<Type>>>()?;
         let terminated = format!("{}\0", function);
-        let pointer: Symbol<FP> = unsafe {
+        let pointer: Symbol<'_, FP> = unsafe {
             library
                 .get(terminated.as_bytes())
                 .map_err(|err| err.to_string())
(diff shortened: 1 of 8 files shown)


arihant2math:
Some updates:

I've gotten arbitrary function argument passing to work on a test, I still am figuring out return type inference however.
Here is some test code for windows that works:
```rust
use std::ffi::c_void;
use std::path::Path;
use libffi::middle as libffi;
use libloading::Symbol;

type FP = unsafe extern "C" fn ();

fn main() {
    let path = "C:\\Windows\\System32\\kernel32.dll";
    let path = Path::new(path);
    let lib = unsafe { libloading::Library::new(path) }.unwrap();
    let function_bytes = b"Sleep\0";
    let function: Symbol<FP> = unsafe { lib.get(function_bytes) }.unwrap();
    let function = *function;
    let args = vec![libffi::Type::u32()];
    let cif = libffi::Cif::new(args.into_iter(), libffi::Type::void());
    let ms_wait: u32 = 5000;
    println!("Calling function");
    let now = std::time::Instant::now();
    let _res: *const c_void = unsafe { cif.call(libffi::CodePtr(function as *mut _), &[libffi::arg(&ms_wait)]) };
    println!("Function returned in {}ms", now.elapsed().as_millis());
    println!("Function called");
}
```
```
Calling function
Function returned in 5001ms
Function called
```

arihant2math:
Nevermind, it seems like return type needs to be manually specified and the default is cint

arihant2math:
Ready for review.
The current working scope is no-argument functions that return integers.

(this is from the extra test)
```python
libc = cdll.msvcrt
print("rand", libc.rand())
```

/cc @youknowone

youknowone (commented):
Looks good, thank you so much!
Please check if the extra indent is intended or not.

arihant2math:
Done.

youknowone (approved):
👍

vm/src/stdlib/ctypes/function.rs
```diff
@@ -1,24 +1,182 @@
...
+impl Constructor for PyCFuncPtr {
+    type Args = FuncArgs;
+
+    fn py_new(_cls: PyTypeRef, args: Self::Args, vm: &VirtualMachine) -> PyResult {
+        let tuple = args.args.first().ok_or_else(|| {
+            vm.new_type_error("CFuncPtr() takes exactly 1 argument (0 given)".to_string())
+        })?;
+        let tuple: &Py<PyTuple> = tuple
+            .downcast_ref()
+            .ok_or(vm.new_type_error("Expected a tuple".to_string()))?;
```
youknowone, line 120: Oh, just one more thing. Let's do this here. We can more freely split FuncArgs

```suggestion
    type Args = (PyTupleRef, FuncArgs);

    fn py_new(_cls: PyTypeRef, (tuple, _args): Self::Args, vm: &VirtualMachine) -> PyResult {
```

youknowone (changes requested):

youknowone (approved):

PR #5534, opened 2025-02-20
Fix warnings for rust 1.85

coolreader18:


arihant2math:
Whoops I think we snipped each other: #5532 😆 . I can close mine if you want.

coolreader18:
Oh, seems like it lol. Sure, that sounds good.

Issue #3960, opened 2022-07-24
Handle windows test proper way


youknowone:
Example work: https://github.com/RustPython/RustPython/pull/3924/commits/e32523c63219429b08668d8c4ea7afab35471d58

Remove tests from ci.yaml and add proper markers to the test files.

- [ ] test_importlib
- [ ] test_io
- [ ] test_iter
- [ ] test_os
- [ ] test_pathlib
- [ ] test_shutil
- [ ] test_venv

fanninpm:
`test_importilb` should be done in #3919

derlih:
Looks like test_iter was added in e87033c

samchakra0204:
@youknowone I've opened PR #7737 to complete the remaining work here.

The tests that were previously skipped on Windows (test_pathlib and test_posixpath) now have `@unittest.expectedFailure` + `# TODO: RUSTPYTHON` markers. The other files (test_rlcompleter, test_venv) were already properly marked, and no CI skips needed removal in my fork. This should close the issue once merged.

PR #5536, opened 2025-02-21
Windows testing

arihant2math:
This makes it so that failing tests on windows are explicitly marked on CI, and considered platform dependent, and non-failing ones are re-enabled. This also fixes a few bugs to unblock tests and fix broken ones.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -60,7 +60,6 @@ env:
     test_dis
     test_enumerate
     test_exception_variations
-    test_exceptions
     test_float
     test_format
     test_fractions
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,6 +18,7 @@ env:
   CARGO_ARGS: --no-default-features --features stdlib,zlib,importlib,encodings,sqlite,ssl
   # Skip additional tests on Windows. They are checked on Linux and MacOS.
   WINDOWS_SKIPS: >-
+    test_argparse
     test_glob
     test_importlib
     test_io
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -17,14 +17,24 @@ concurrency:
...
+  # test_argparse: UnicodeDecodeError
+  # test_glob: many failing tests
+  # test_io: many failing tests
+  # test_os: many failing tests
+  # test_pathlib: support.rmtree() failing
+  # test_posixpath: OSError: (22, 'The filename, directory name, or volume label syntax is incorrect. (os error 123)')
+  # test_unicode: AttributeError: module '_winapi' has no attribute 'GetACP'
+  # test_venv: couple of failing tests
   WINDOWS_SKIPS: >-
+    test_argparse
```
youknowone, line 29: I am not getting how this one and test_rlcompleter was not tested or didn't fail on windows.

youknowone (commented):
Thank you for catching test_unicode and test_exceptions are platform dependent.

.github/workflows/ci.yaml
arihant2math: It was not tested as it was considered platform independent

youknowone: Thanks. Is this currently failing on windows?
Then this test looks like a platform-dependent tests. it is expected to be not listed on platform dependent tests.
If the test success and it doesn't have platform-dependent tests, we don't need to add it here.
We only add tests here when the tests are not annotated well, so running it on Windows fails.
This list is expected to be removed in future once we annotate every tests enough

youknowone: See also #3960

arihant2math: Yes this fails due to #5542.

youknowone: Thank you for confirming. Then please remove it from platform-independent tests list too

youknowone (approved):
👍

PR #5537, opened 2025-02-21
Split CI

arihant2math:
Use parallel jobs to test snippets and run the cpython testsuite after the build


youknowone:
This is good idea in general. But the snippets tests take comparably a lot less time than Python unittest. Running snipptes took only 30s in this run: https://github.com/RustPython/RustPython/actions/runs/13447996296/job/37577284807
The total running time will not be that much improved

PR #5538, opened 2025-02-21
Ensure pymethod cannot be both magic and named simultaneously + macro documentation

arihant2math:
Originally, this would ignore the `magic`, which is counterintuitive, so it now just throws an error (nothing relies on this, so I think a hard error should be fine). I also documented `pyclass` and `pymodule` with API docs which cover almost everything.
The docs here are mostly just highly specific documentation for every argument and use case. https://github.com/RustPython/RustPython/wiki/Python-Attributes is currently a high level overview and can now be more geared towards that.

It would be nice if the docs for the main branch could be deployed using github pages to display them for users, since the current docs.rs content is highly out of date.


youknowone (approved):
👍

PR #5555, opened 2025-02-24
No panic on symbol lookup fail

arihant2math:


compiler/codegen/src/error.rs
```diff
@@ -33,6 +33,7 @@ pub enum CodegenErrorType {
     DuplicateStore(String),
     InvalidMatchCase,
     NotImplementedYet, // RustPython marker for unimplemented features
+    SymbolLookupError { symbol: String },
```
youknowone, line 36: Because this is not a user-side Python error but our own RustPython implementation bug, let's not expose it here.

youknowone (changes requested):

coolreader18:
Do you have a repr for this getting triggered? The reason it's a panic is because this shouldn't ever happen, the symboltable should be populated with every identifier present in the source file even if it would be a NameError at runtime. It should really be an `unreachable!()`, not a `panic!()`.

arihant2math:
Makes sense, this occurs because I implemented match statements poorly. I feel like the instruction set to be upgraded to 3.12 before match statements are attempted so there is a better feel for specialized instructions.

arihant2math:
Interestingly enough #13161 runs into this as well for unrelated reasons.

arihant2math:
I have found a superior solution

PR #5557, opened 2025-02-25
Enable `unsafe_op_in_unsafe_fn` and `missing_unsafe_on_extern` lints

coolreader18:
These will be enabled by default in rust2024, and they make more clear what is actually unsafe. I didn't add safety comments for the new unsafe blocks required by `unsafe_op_in_unsafe_fn` (except where obvious that requirements are simply forwarded from the caller), but we have plenty of unsafe blocks with no safety comments anyway. As I was adding the new unsafe blocks, I noticed it would make more sense and make the new code clearer if `PyObjectRef::{from,into}_raw()` took/returned `NonNull` instead of raw pointers, so I made that change in its own commit.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -190,6 +190,8 @@ wasm-bindgen = "0.2.100"
 [workspace.lints.rust]
 unsafe_code = "allow"
 unsafe_op_in_unsafe_fn = "deny"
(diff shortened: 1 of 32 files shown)


youknowone (approved):
👍

PR #5558, opened 2025-02-25
 Use non-env-var methods from openssl_probe

coolreader18:
`std::env::{set,remove}_var()` are unsafe in rust2024 due to raciness issues, and this PR treats them as such and also switches away from `openssl_probe::init_openssl_env_vars()`, which also uses `set_var` and is now marked unsafe.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -192,6 +192,7 @@ unsafe_code = "allow"
 unsafe_op_in_unsafe_fn = "deny"
 missing_unsafe_on_extern = "deny"
 unsafe_attr_outside_unsafe = "deny"
+deprecated_safe_2024 = "deny"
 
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository.workspace = true
 license.workspace = true
 
 [features]
-default = ["threading", "stdlib", "zlib", "importlib"]
+default = ["threading", "stdlib", "zlib", "importlib", "ssl"]
```
youknowone, line 13: We removed it due to difficult ssl setup issue on windows. is it now resolved?

coolreader18: Oh, oops, that was accidentally left behind from testing.

youknowone (approved):

PR #5559, opened 2025-02-25
Enable rust2024-incompatible pat and keyword-ident lints

coolreader18:
Open to suggestions for a better "euphemism" for `gen` - could always just go with `gen_` or even `r#gen`, but `jen` felt like it fit in the tradition of `klass` and `zelf`.

### DIFF
--- a/vm/src/stdlib/ast.rs
+++ b/vm/src/stdlib/ast.rs
@@ -3,7 +3,7 @@
 //! This module makes use of the parser logic, and translates all ast nodes
 //! into python ast.AST objects.
 
-mod gen;
+mod r#gen;
 
 use crate::{
     builtins::{self, PyDict, PyModule, PyStrRef, PyType},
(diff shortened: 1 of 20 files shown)


vm/src/stdlib/ast.rs
```diff
@@ -3,7 +3,7 @@
 //! This module makes use of the parser logic, and translates all ast nodes
 //! into python ast.AST objects.
 
-mod gen;
+mod r#gen;
```
youknowone, line 6: Unlike variable names, we don't need shorten `gen` here.

youknowone (approved):
I don't feel comfortable about gen, and maybe I chose a bit longer name like `genor` if I did. Otherwise `zen` because it is very frequently quoted in Python world.
But it may be just because I am not used to `jen` yet. No problem. Let's try.

PR #5561, opened 2025-02-26
Add JIT compilation support for integer multiplication, division, and exponents

dohear:
This PR adds JIT compilation support for:

- Multiplication with two integers
- Division with two integers
- Exponents where the base and exponent are both integers

Summary of Changes:

1. Added support for int multiplication and division in the BinOp match statement in`jit/src/instructions.rs`
2. Created a helper function (`compile_ipow`) for exponents due to additional calculation overhead
3. Implemented unit tests for all three operations in `jit/tests/int_tests.rs`


coolreader18 (approved):
LGTM!

PR #5562, opened 2025-02-26
Switch to libz-rs-sys for zlib implementation

coolreader18:
This is a rust-only reimplementation of the `libz-sys` api, which I discovered [from this blog post](https://trifectatech.org/blog/zlib-rs-is-faster-than-c/). It can build without any C toolchain, which makes it a good pick for us when we're targeting windows and wasm, and the performance isn't anything to scoff at (see blog post). Also, the `zlib` feature now enables the entire `zlib` module, which I feel like just makes more sense.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -15,7 +15,7 @@ concurrency:
   cancel-in-progress: true
 
 env:
-  CARGO_ARGS: --no-default-features --features stdlib,zlib,importlib,encodings,sqlite,ssl
+  CARGO_ARGS: --no-default-features --features stdlib,importlib,encodings,sqlite,ssl
   # Skip additional tests on Windows. They are checked on Linux and MacOS.
   # test_glob: many failing tests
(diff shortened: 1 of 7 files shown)


youknowone:
if it is a pure rust implementation, do we still need a `zlib` feature?

coolreader18:
It's a reasonably heavy dependency (~500KB), so I figured it might be nice to still have the option to turn it off.

youknowone:
We don't suggest feature for every library which can be a redundant dependencies. If the binary size matters, turning off the entire stdlib feature and consist their own well-selected (or we can suggest a few pre-selected) set of libraries will make more sense.

coolreader18:
Fair enough.

youknowone:
Probably we can remove zlib feature from `rustpython` crate but add a lot of features for each library in stdlib as a user interface to support easy library on/off. (not a scope of this pr)

youknowone (approved):

PR #5567, opened 2025-02-27
fix devcontainer.json

arihant2math:
See: https://stackoverflow.com/questions/78216235/cargo-not-found-after-installing-in-a-devcontainer


coolreader18 (approved):
Not super familiar with this but this lgtm

PR #5571, opened 2025-02-28
Update to windows-2025 runner on ci

coolreader18:
This comes preinstalled with openssl 3, so we can cut out building openssl ourselves, which takes 5 minutes every CI run.


arihant2math:
The issue here is env variables that point to the openssl installation are not set.

Setting all the variables specified in https://github.com/sfackler/rust-openssl/issues/1542#issuecomment-2524831738 should resolve this.

coolreader18:
@youknowone you added the `win32_xstat` impl in #5247 - do you know why the tests fixed by that might be failing now with a windows update? It's not like they would've removed the `api-ms-win-core-file-l2-1-4` API set.

youknowone:
I am sorry. I code like a monkey with wniapi. I really don't know well about their details.

bjia56:
I mentioned this link in #5583, but openssl prebuilts are also available from the cpython project here https://github.com/python/cpython-bin-deps/

PR #5579, opened 2025-03-02
Fixes evaluation order in expression lists

### LINKED ISSUE #5566: Incorrect order of evaluation with unpacking in expression lists and call arguments
## Feature

Consider the following:

```py
class Noisy:
    def __iter__(self):
        print('__iter__')
        return iter([])

[*Noisy(), print('arg')]

def sink(*args):
    pass

sink(*Noisy(), print('arg'))
```

CPython prints

```
__iter__
arg
__iter__
arg
```

since it extracts all elements from the unpacked expression before proceeding to the next argument.

Meanwhile, RustPython prints

```
arg
__iter__
arg
__iter__
```

since we currently evaluate all given expressions without unpacking, then unpack and combine the arguments.

https://docs.python.org/3/reference/expressions.html#calls

ivan-shrimp:
Enforce left-to-right evaluation with list/tuple/set expressions and positional arguments in calls.

We cannot iterate through the `*`-unpacked iterables after evaluating all given expressions. This PR fixes the problem by temporarily collecting the unpacked iterable into a tuple before moving on to the next element.

Fixes #5566 

(`**`-unpacking with `dict`s and keyword arguments seem to be more buggy; working on a fix.)

### DIFF
--- a/compiler/codegen/src/compile.rs
+++ b/compiler/codegen/src/compile.rs
@@ -2430,7 +2430,7 @@ impl Compiler<'_> {
             Expr::List(ExprList { elts, .. }) => {
                 let (size, unpack) = self.gather_elements(0, elts)?;
                 if unpack {
-                    emit!(self, Instruction::BuildListUnpack { size });
+                    emit!(self, Instruction::BuildListFromTuples { size });
                 } else {
                     emit!(self, Instruction::BuildList { size });
                 }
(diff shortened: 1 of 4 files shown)


youknowone:
@ivan-shrimp Hi, thank you for contributing. Is this still a draft or done?

youknowone (approved):
The patch looks really good. Thank you so much for contributing!

PR #5581, opened 2025-03-03
Remove winapi dependency

arihant2math:


stdlib/src/socket.rs
```diff
@@ -35,18 +35,13 @@ mod _socket {
...
-            SO_EXCLUSIVEADDRUSE, getprotobyname, getservbyname, getservbyport, getsockopt,
-            setsockopt,
-        };
-        pub use winapi::um::ws2tcpip::{
-            EAI_AGAIN, EAI_BADFLAGS, EAI_FAIL, EAI_FAMILY, EAI_MEMORY, EAI_NODATA, EAI_NONAME,
-            EAI_SERVICE, EAI_SOCKTYPE,
-        };
+        pub use windows_sys::Win32::NetworkManagement::IpHelper::{if_indextoname, if_nametoindex};
+
+        pub const INADDR_ANY: u32 = 0x00000000;
```
youknowone, line 40: This is defined under Win32::Networking::WinSock

https://docs.rs/windows-sys/latest/windows_sys/Win32/Networking/WinSock/constant.INADDR_ANY.html

stdlib/src/socket.rs
```diff
@@ -1755,6 +1761,14 @@ mod _socket {
...
         let cstr_proto = cstr_opt_as_ptr(&cstr_proto);
+        #[cfg(windows)]
+        let serv = unsafe {
+            c::getservbyname(
+                cstr_name.as_ptr() as windows_sys::core::PCSTR,
+                cstr_proto as windows_sys::core::PCSTR,
+            )
+        };
+        #[cfg(not(windows))]
         let serv = unsafe { c::getservbyname(cstr_name.as_ptr(), cstr_proto) };
```
youknowone, line 1772: I hate this part of windows_sys everything is defined with u8, but not with `libc::c_char`

Though this is not always a best idea, these APIs are very stable system and never will be changed. This way can be simpler.
```suggestion
        let serv = unsafe { c::getservbyname(cstr_name.as_ptr() as _, cstr_proto as _) };
```

If this is looking too unsafe, adding a wrapper function in `rustpython_common` somewhere will be a good way for global windows_sys usage.

```rust
/// 

#[cfg(windows)]
type SYS_STR = windows_sys::core::PCSTR;
#[cfg(not(windows)]
type SYS_STR = *const libc::c_char;

trait AsSysStr {
    fn as_sys_str() -> SYS_STR;
}


// mm.. these impls will make conflict. need better idea
impl<T> AsSysStr for T where T: AsRef<std::ffi::CStr> {
    fn as_sys_str(&self) -> SYS_STR {
        self.as_ref().as_ptr() as _
    }
}

impl<T> AsSysStr for Option<T> where T: AsRef<std::ffi::CStr> {
    fn as_sys_str(&self) -> SYS_STR {
        self.map_or_else(std::ptr::null, self.as_ref().as_ptr() as _)
    }
}
```

Then,

```suggestion
        let serv = unsafe { c::getservbyname(cstr_name.as_sys_str(), cstr_proto.into_option().as_sys_str()) };
```

stdlib/src/socket.rs
```diff
@@ -2111,13 +2135,16 @@ mod _socket {
     #[cfg(all(unix, not(target_os = "redox")))]
     type IfIndex = c::c_uint;
     #[cfg(windows)]
-    type IfIndex = winapi::shared::ifdef::NET_IFINDEX;
+    type IfIndex = u32;
```
youknowone, line 2138:
```suggestion
    type IfIndex = u32;  // NET_IFINDEX but windows-sys 0.59 doesn't have it
```

youknowone (dismissed):
Thank you so much! Now we can only have a single windows api source.

fanninpm:
`cargo fmt` should fix the error.

stdlib/src/socket.rs
arihant2math: Had to upgrade to windows-sys 0.59 to do it

coolreader18 (approved):
LGTM, thanks!

PR #5583, opened 2025-03-05
_tkinter pt. 1

arihant2math:
Feature-gated because tk had to be installed in a not-very-easy manner to work on windows.

- [x] Add tkinter code
- [x] Add setup for tkinter feature/module
- [x] Add basic type stubs

### DIFF
--- a/stdlib/src/tkinter.rs
+++ b/stdlib/src/tkinter.rs
@@ -34,12 +34,6 @@ mod _tkinter {
         Ok(main_loop())
     }
 
-    // TODO: Remove once enough has been implemented.
-    #[pyfunction]
-    fn tk_demo() {
-        let _ = demo();
-    }
-
     #[pyattr(once, name = "TclError")]
     fn tcl_error(vm: &VirtualMachine) -> PyTypeRef {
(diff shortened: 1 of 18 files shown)


bjia56:
Would tcl/tk binaries sourced from https://github.com/python/cpython-bin-deps/ be useful for Windows?

arihant2math:
Yup, that is, I'll see if the crates I'm using will play nice with it.

stdlib/src/tkinter.rs
```diff
@@ -0,0 +1,90 @@
...
+            .add_button("btn" - text("quit") - command("destroy ."))?
+            .pack(())?;
+        Ok(main_loop())
+    }
+
+    // TODO: Remove once enough has been implemented.
+    #[pyfunction]
+    fn tk_demo() {
+        let _ = demo();
+    }
```
youknowone, line 41: this test function can be removed from this patch and moved to the next patch branch

youknowone (changes requested):
👍

youknowone (approved):
👍

PR #5580, opened 2025-03-03
Publish demo on weekly release

arihant2math:
Fixes #3751

PR #5585, opened 2025-03-05
Update webpack

coolreader18:
This fixes the notebook deploy from #5580. cc @arihant2math


coolreader18:
Success! https://github.com/RustPython/RustPython/actions/runs/13684794525/job/38265572206 (the other ones failed because I cancelled them, so that they wouldn't make a release)

PR #5587, opened 2025-03-07
Allow surrogates in str

coolreader18:
This switches the backing format of `PyStr` from `Box<str>` to `Box<Wtf8>`. From the motivation I wrote in a doc comment:

An implementation of [WTF-8], a utf8-compatible encoding that allows for unpaired surrogate codepoints. This implementation additionally allows for paired surrogates that are nonetheless treated as two separate codepoints.

RustPython uses this because CPython internally uses a variant of UCS-1/2/4 as its string storage, which treats each `u8`/`u16`/`u32` value (depending on the highest codepoint value in the string) as simply integers, unlike UTF-8 or UTF-16 where some characters are encoded using multi-byte sequences. CPython additionally doesn't disallow the use of surrogates in `str`s (which in UTF-16 pair together to represent codepoints with a value higher than `u16::MAX`) and in fact takes quite extensive advantage of the fact that they're allowed. The `surrogateescape` codec-error handler uses them to represent byte sequences which are invalid in the given codec (e.g.  bytes with their high bit set in ASCII or UTF-8) by mapping them into the surrogate range. `surrogateescape` is the default error handler in Python for interacting with the filesystem, and thus if RustPython is to properly support `surrogateescape`, its `str`s must be able to represent surrogates.

We use WTF-8 over something more similar to CPython's string implementation because of its compatibility with UTF-8, meaning that in the case where a string has no surrogates, it can be viewed as a UTF-8 Rust `str` without needing any copies or re-encoding.

This implementation is mostly copied from the WTF-8 implentation in the Rust standard library, which is used as the backing for `OsStr` on Windows targets. As previously mentioned, however, it is modified to not join two surrogates into one codepoint when concatenating strings, in order to match CPython's behavior.

[WTF-8]: https://simonsapin.github.io/wtf-8

## Method

I opted not to change the signature of `PyStr::as_str()`, since that would mean a truly massive diff. Instead, I have it panic, and my hope is that we can get to a place with this where tests pass, and then we can deal with the rest of the `as_str` calls incrementally.

## Status

Currently panicking in cformat code. Is there still a reason `Parser` et al need to be a separate repo if ruff isn't using it anymore? It'd be really convenient to merge it into the monorepo.

### DIFF
--- a/common/src/wtf8/mod.rs
+++ b/common/src/wtf8/mod.rs
@@ -75,6 +75,12 @@ impl fmt::Debug for CodePoint {
     }
 }
 
+impl fmt::Display for CodePoint {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.to_char_lossy().fmt(f)
+    }
+}
+
 impl CodePoint {
     /// Unsafely creates a new `CodePoint` without checking the value.
     ///
@@ -109,13 +115,13 @@ impl CodePoint {
 
     /// Returns the numeric value of the code point.
     #[inline]
-    pub fn to_u32(&self) -> u32 {
+    pub fn to_u32(self) -> u32 {
         self.value
     }
 
     /// Returns the numeric value of the code point if it is a leading surrogate.
     #[inline]
-    pub fn to_lead_surrogate(&self) -> Option<u16> {
+    pub fn to_lead_surrogate(self) -> Option<u16> {
         match self.value {
             lead @ 0xD800..=0xDBFF => Some(lead as u16),
             _ => None,
@@ -124,7 +130,7 @@ impl CodePoint {
 
     /// Returns the numeric value of the code point if it is a trailing surrogate.
     #[inline]
-    pub fn to_trail_surrogate(&self) -> Option<u16> {
+    pub fn to_trail_surrogate(self) -> Option<u16> {
         match self.value {
             trail @ 0xDC00..=0xDFFF => Some(trail as u16),
             _ => None,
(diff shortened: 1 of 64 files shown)


youknowone:
can we separately merge less wtf8 related changes like format/cformat changes?

coolreader18:
#5621

common/src/wtf8/mod.rs
```diff
@@ -0,0 +1,1481 @@
...
+//! needing any copies or re-encoding.
+//!
+//! This implementation is mostly copied from the WTF-8 implentation in the
+//! Rust standard library, which is used as the backing for [`OsStr`] on
+//! Windows targets. As previously mentioned, however, it is modified to not
+//! join two surrogates into one codepoint when concatenating strings, in order
+//! to match CPython's behavior.
+//!
+//! [WTF-8]: https://simonsapin.github.io/wtf-8
+//! [`OsStr`]: std::ffi::OsStr
```
youknowone, line 32: Let's leave a comment which version of Rust those files are copied from.

vm/src/anystr.rs
```diff
@@ -403,25 +401,33 @@ pub trait AnyStr {
...
-        //  unicode_islower_impl
-        //  unicode_isupper_impl
-        let mut cased = false;
-        for c in self.chars() {
-            if is_opposite(c) {
+    // Unified form of CPython functions:
+    //  _Py_bytes_islower
+    //   Py_bytes_isupper
+    //  unicode_islower_impl
+    //  unicode_isupper_impl
```
youknowone, line 408: since the function is splitted, the comment also need to be splited

vm/sre_engine/Cargo.toml
```diff
@@ -14,7 +14,12 @@ license.workspace = true
 name = "benches"
 harness = false
 
+[features]
+default = ["wtf8"]
+wtf8 = ["rustpython-common"]
```
youknowone, line 19: do we need an optional feature?

youknowone (approved):
👍

vm/sre_engine/Cargo.toml
coolreader18: Not really, I guess.

PR #5595, opened 2025-03-11
partially fix sys.getwindowsversion()

arihant2math:
The `build` field is still somewhat wrong though.


youknowone:
Could you add a test under extra_test to check the versions are aligned to CPython return?

arihant2math:
I can't really judge the results of the function call against anything, so I just ensured it wasn't 6 (which is windows XP).

youknowone (approved):
Ah, right. It is tricky and only writable above snippets level.

PR #5599, opened 2025-03-11
Fix test_sqlite hang

arihant2math:


arihant2math:
doesn't seem like a fix

arihant2math:
the issue is not test_sqlite

PR #5602, opened 2025-03-12
Use lexopt for argument parsing

coolreader18:
Supercedes #5414 -- the python CLI is really not that complicated, and it turns out it's not that hard to just do the flag handling ourselves, with a little help from [`lexopt`](https://docs.rs/lexopt). This actually cuts down on the number of lines of code we have in the codebase!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -52,15 +52,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


fanninpm (approved):
LGTM. @youknowone what do you think?

PR #5603, opened 2025-03-12
Fix hang in test_sqlite3

coolreader18:
I'm not sure why this suddenly started happening, but I'm pretty sure this is the test that's doing it.


arihant2math:
Didn't seem to fix it.

coolreader18:
Ah - seems to be related to the `statistics` update. Might be worth rolling that back or skipping tests in there just to make the tests complete in a reasonable amount of time

PR #5604, opened 2025-03-12
Initial C-api implementation

arihant2math:
This PR doesn't aim to implement any particular feature set/subset of the C-API, rather it tries to get the scaffolding in place, so that most functions could be implemented.

It's currently built as a dll/so, but it probably needs to be renamed later on (packaging this kind of this is not my area of expertise).

This PR implements large parts of the concrete objects layer and a stub for error handling, although it needs to be expanded on.


arihant2math:
Ready for review.

arihant2math:
Should I Just exclude `capi` from the spellchecker.

youknowone:
What kind terms bother you?

I disabled a few files for reasons:
- Not worth to spend too much effort. They are OS/platform/whatever very domain specific terms 
- Too painful to handle right now. Those files will be eventually turned to enabled again

When sharing a few terms with CPython implementation makes problem, adding them to cpython.txt will be better

arihant2math:
Stuff like ccomplex. I'm actually just going to add it to cspell, since it's probably better that way.

PR #5605, opened 2025-03-13
Fix bz2 module

arihant2math:
The new `libbz2-rs-sys` is a rust only implementation of bz2. This is similar to #5562.

Blog post: https://trifectatech.org/blog/translating-bzip2-with-c2rust/


arihant2math:
@youknowone ready for merge

Lib/test/test_bz2.py
```diff
@@ -106,6 +106,8 @@ def testBadArgs(self):
         # compresslevel is keyword-only
         self.assertRaises(TypeError, BZ2File, os.devnull, "r", 3)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
youknowone, line 110: Why should we go this way if it occures these many regressions?
If this can be fixed in future, let's fix it first and merge it later.
Otherwise this must be an optional feature rather for specific requirements.

youknowone (changes requested):

PR #5607, opened 2025-03-13
Remove uu.py and test_uu.py

arihant2math:


Lib/email/message.py
```diff
@@ -101,6 +100,35 @@ def _unquotevalue(value):
         return utils.unquote(value)
 
 
+def _decode_uu(encoded):
```
youknowone, line 103: huh,, so uu library is gone but the decoding feature is still in the standard library

youknowone (approved):

Lib/email/message.py
arihant2math: Yeah I just copied this off the 3.13 stdlib, seems like a weird patch.

PR #5613, opened 2025-03-16
Update test class to cpython 3.13.2

key262yek:
- Copy `Lib/test/test_clsss.py` from Cpython 3.13.2
- Introduce new cpython_only test api `_testlimitedcapi`


youknowone (approved):
Thank you!

PR #5614, opened 2025-03-18
Floating Point Power and While Loop JIT

NicholasPaulick:
## Summary

- Added support for floating-point values in Cranelift.
- Fully re-implemented while loops.
- Upgraded Cranelift from v0.88.0 to v0.116.1.

## Details
- Floating-point support works in most cases, but some values may produce slight deviations from expected results.
- Cases that could result in imaginary outputs were not included, as handling them would require a Tuple, which is yet to be implemented.

Let me know if you'd like any further refinements!


coolreader18 (approved):
Looks awesome, thank you!

PR #5617, opened 2025-03-20
Remove crawl_sourcecode.py

arihant2math:
This seems unneeded as we rely on ruff now. This also updates the output of `demo_closures.py` in the README (although it could also conceivably be removed).


arihant2math:
Seems like it encompasses opcode related stuff too.

PR #5620, opened 2025-03-25
Fix _ctypes.Array base and metacalss

youknowone:
@arihant2math could you review this PR?


arihant2math (commented):
The following causes different results in python and this branch:

```python
from _ctypes import _SimpleCData

class c_short(_SimpleCData):
    _type_ = "h"
print(c_short.__mul__(5))
```
**Python:**
```
<class '__main__.c_short_Array_5'>
```
**RustPython:**
```
  File "tmp.py", line 6, in <module>
    print(c_short.__mul__(5))
TypeError: unexpected payload for __repr__
```

I don't think the python metaclass does anything special, so I assume the rustpython implementation needs repr implemented in a different pr.

arihant2math (approved):

youknowone:
ah, this is #1979

arihant2math:
That seems to no longer occur for that case, but I suppose there might have been a bug in that patch.

coolreader18 (approved):

Issue #5491, opened 2025-01-23
Bytecode updates for match statements


arihant2math:
When implementing #5485 I ran into the problem of opcodes simply not existing. This enumerates the opcodes needed for full support of match statements

- [ ] NOP
- [ ] SWAP
- [ ] MATCH_KEYS
- [ ] MATCH_MAPPING
- [ ] MATCH_SEQUENCE
- [ ] MATCH_CLASS

PR #5628, opened 2025-03-26
Match statements rewrite

arihant2math:
Retains all match statement functionality, but rewrites all the match statement code to reduce bugs and mirror cpython. Also, it adds support for the rest of the patterns (hopefully).

Also this re-enables the codegen tests that the ruff port disabled.

Fixes: #5628 and #5491

### DIFF
--- a/compiler/codegen/src/error.rs
+++ b/compiler/codegen/src/error.rs
@@ -1,6 +1,21 @@
 use ruff_source_file::SourceLocation;
-use std::fmt;
+use std::fmt::{self, Display};
 use thiserror::Error;
+
+#[derive(Debug)]
+pub enum PatternUnreachableReason {
+    NameCapture,
+    Wildcard,
+}
+
+impl Display for PatternUnreachableReason {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::NameCapture => write!(f, "name capture"),
+            Self::Wildcard => write!(f, "wildcard"),
+        }
+    }
+}
 
 // pub type CodegenError = rustpython_parser_core::source_code::LocatedError<CodegenErrorType>;
 
@@ -47,8 +62,9 @@ pub enum CodegenErrorType {
     TooManyStarUnpack,
     EmptyWithItems,
     EmptyWithBody,
+    ForbiddenName,
     DuplicateStore(String),
-    InvalidMatchCase,
+    UnreachablePattern(PatternUnreachableReason),
     NotImplementedYet, // RustPython marker for unimplemented features
 }
 
@@ -94,11 +110,14 @@ impl fmt::Display for CodegenErrorType {
             EmptyWithBody => {
                 write!(f, "empty body on With")
             }
+            ForbiddenName => {
+                write!(f, "forbidden attribute name")
+            }
             DuplicateStore(s) => {
                 write!(f, "duplicate store {s}")
             }
-            InvalidMatchCase => {
-                write!(f, "invalid match case")
+            UnreachablePattern(reason) => {
+                write!(f, "{reason} makes remaining patterns unreachable")
             }
             NotImplementedYet => {
                 write!(f, "RustPython does not implement this feature yet")
--- a/compiler/core/src/bytecode.rs
+++ b/compiler/core/src/bytecode.rs
@@ -438,6 +438,7 @@ pub enum Instruction {
         op: Arg<TestOperator>,
     },
     /// If the argument is true, perform IS NOT. Otherwise perform the IS operation.
+    // TODO: duplication of TestOperator::{Is,IsNot}. Fix later.
     IsOperation(Arg<bool>),
     CompareOperation {
         op: Arg<ComparisonOperator>,
--- a/compiler/core/src/bytecode.rs
+++ b/compiler/core/src/bytecode.rs
@@ -381,6 +381,7 @@ pub type NameIdx = u32;
 #[derive(Debug, Copy, Clone, PartialEq, Eq)]
 #[repr(u8)]
 pub enum Instruction {
+    Nop,
     /// Importing by name
     ImportName {
         idx: Arg<NameIdx>,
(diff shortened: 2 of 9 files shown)


arihant2math:
Currently running into the following:
```
==encodings initialization failed. Only utf-8 encoding will be supported.
  File "[elided]\RustPython\Lib\encodings\__init__.py", line 33, in <module>
    from . import aliases
  File "[elided]\RustPython\Lib\encodings\aliases.py", line 550, in <module>
    'x_mac_trad_chinese'  : 'big5',
TypeError: 'bool' object is not an iterator

RuntimeError: RustPython could not import the encodings module. It usually means something went wrong. Please carefully read the following messages and follow the steps.

Neither RUSTPYTHONPATH nor PYTHONPATH is set. Try setting one of them to the stdlib directory.
If you don't have access to a consistent external environment (e.g. targeting wasm, embedding rustpython in another application), try enabling the `freeze-stdlib` feature.
If this is intended and you want to exclude the encodings module from your interpreter, please remove the `encodings` feature from `rustpython-vm` crate.
[vm\src\frame.rs:2182:9] self = ExecutingFrame {
    code: code: <code object <module> at ??? file "\\\\?\\[elided]\\RustPython\\Lib\\abc.py", line 1>,
    state: FrameState {
        stack: [
            [PyObject code: <code object abstractmethod at ??? file "\\\\?\\[elided]\\RustPython\\Lib\\abc.py", line 7>],
            [PyObject PyStr { value: "abstractmethod", kind: Ascii, hash: -1 }],
            [PyObject code: <code object abstractmethod at ??? file "\\\\?\\[elided]\\RustPython\\Lib\\abc.py", line 7>],
            [PyObject PyStr { value: "abstractmethod", kind: Ascii, hash: -1 }],
            [PyObject code: <code object abstractmethod at ??? file "\\\\?\\[elided]\\RustPython\\Lib\\abc.py", line 7>],
        ],
        blocks: [],
        lasti: 4,
    },
}

thread 'main' panicked at vm\src\frame.rs:526:22:
tried to push value onto stack but overflowed max_stackdepth
error: process didn't exit successfully: `target\release\rustpython.exe ./extra_tests/snippets/syntax_match.py` (exit code: 101)
```

arihant2math:
Note that class and sequence matching is broken, but this is still a significant improvement over the current implementation.

compiler/codegen/src/error.rs
```diff
@@ -1,7 +1,22 @@
 use ruff_source_file::SourceLocation;
-use std::fmt;
+use std::fmt::{self, Display};
 use thiserror::Error;
 
+#[derive(Debug)]
```
youknowone, line 5: you may want to try `derive_more::Display` for this simple Display implementation.
This comment is not a change request

vm/src/frame.rs
```diff
@@ -799,12 +827,28 @@ impl ExecutingFrame<'_> {
...
+                    .push(container.get_item(key.as_object(), vm)?);
+                Ok(None)
+            }
             bytecode::Instruction::LoadAttr { idx } => self.load_attr(vm, idx.get(arg)),
             bytecode::Instruction::StoreAttr { idx } => self.store_attr(vm, idx.get(arg)),
             bytecode::Instruction::DeleteAttr { idx } => self.delete_attr(vm, idx.get(arg)),
             bytecode::Instruction::UnaryOperation { op } => self.execute_unop(vm, op.get(arg)),
             bytecode::Instruction::TestOperation { op } => self.execute_test(vm, op.get(arg)),
             bytecode::Instruction::CompareOperation { op } => self.execute_compare(vm, op.get(arg)),
+            bytecode::Instruction::IsOperation(neg) => {
```
youknowone, line 844: This instruction is duplication of a part of `TestOperation`.
I expect one of them to be eliminated.

youknowone (changes requested):
@coolreader18 Could you review this PR?

vm/src/frame.rs
arihant2math: Is `TestOperation` something that was in cpython bytecode but was removed?

We should really maintain a mapping of cpython instructions and associated instructions

youknowone: I guess so.
CPython changed a lot of instructions for recent a few years. The mapping would be useless very quickly, but better to had one.

youknowone:
@coolreader18 do you have time to review?

vm/src/frame.rs
arihant2math: I think I'll consolidate instructions in later prs, this seems more ergonomic to me too.

compiler/core/src/bytecode.rs
```diff
@@ -1403,15 +1430,22 @@ impl Instruction {
             UnaryOperation { op } => w!(UnaryOperation, ?op),
             BinaryOperation { op } => w!(BinaryOperation, ?op),
             BinaryOperationInplace { op } => w!(BinaryOperationInplace, ?op),
+            BinarySubscript => w!(BinarySubscript),
             LoadAttr { idx } => w!(LoadAttr, name = idx),
             TestOperation { op } => w!(TestOperation, ?op),
+            IsOperation(neg) => w!(IsOperation, neg),
```
youknowone, line 1436:
```suggestion
            IsOperation(neg) => w!(IsOperation, neg),  // TODO: duplication of TestOperation
```

compiler/core/src/bytecode.rs
```diff
@@ -429,21 +430,32 @@ pub enum Instruction {
...
     },
+    BinarySubscript,
     LoadAttr {
         idx: Arg<NameIdx>,
     },
     TestOperation {
         op: Arg<TestOperator>,
     },
+    /// If the argument is true, perform IS NOT. Otherwise perform the IS operation.
+    IsOperation(Arg<bool>),
```
youknowone, line 441:
```suggestion
    // TODO: duplication of TestOperator::{Is,IsNot}. Fix later.
    IsOperation(Arg<bool>),
```

youknowone (approved):

PR #5629, opened 2025-03-27
 Parse surrogates in string literals properly

coolreader18:
Because ruff represents string literals as `str`, we have to reparse every string that might have a surrogate `\uDXXX` escape as wtf8. Then there were a bunch of crashes to fix, because surrogate strings in `test/` actually get parsed properly now instead of being subbed with `REPLACEMENT_CHARACTER`.


youknowone:
interesting, copilot reviewer is added. I tried a test run on this pr

coolreader18:
I guess I'm reviewing the review, lol

> This PR aims to improve the handling of surrogate escapes in string literals by reparsing them into WTF‑8 format so that previously crashing surrogate strings now produce valid output

Close, but reparsing the strings to WTF-8 actually began causing crashes. The second commit then fixes those crashes.

The following are "low confidence" but I guess that's for good reason

> Using unwrap_or_else with an empty match in parse_string_literal may mask unexpected errors during surrogate parsing. Consider handling the error explicitly with a descriptive panic message or returning a proper Result.
> ```
>         .unwrap_or_else(|x| match x {})
> ```

Not how uninhabited types work lol

> In the from_bytes function, always advancing by 3 bytes after decoding a surrogate sequence may be fragile if the assumptions about surrogate lengths change. Consider adding a comment to document this behavior or validating the boundary before slicing.

If UTF-8 changes, we have more issues than just this.

> [nitpick] The code branch that reparses string literals containing the replacement character introduces additional processing. Verify that this extra reparsing is required for proper surrogate handling and that the performance impact is acceptable.

Well, yes, that's why it's behind a `.contains(REPLACEMENT_CHAR)` branch in the first place.

compiler/codegen/src/string_parser.rs
```diff
@@ -0,0 +1,287 @@
+//! A stripped-down version of ruff's string literal parser, modified to
```
youknowone, line 1: Would ruff parser also need to be patched?

youknowone (approved):

compiler/codegen/src/string_parser.rs
coolreader18: Yeah, though I'm not sure whether they'd necessarily feel a need to. I think what might strike a good balance is if there's a flag that can be set like `contains_surrogates`, so we can know whether or not we need to reparse.

MichaReiser: Yeah, probably. There's an open issue but it isn't something that has come up often. https://github.com/astral-sh/ruff/issues/13666

PR #5630, opened 2025-03-27
Update rustpython_literal::escape to support wtf8

coolreader18:
The first commit splits out `rustpython_common::wtf8` into its own crate, so that things like literal and sre_engine don't have to depend on all of common just to implement something for wtf8. I also simplified the crate graph a little by removing `compiler`'s dependency on `common`, since `expandtab` was the only function left that `compiler-codegen` needed, and that's simple enough to just copy over. Originally the point of `common` was for there to be shared code between vm and compiler, but I guess the times change :)

Workspace crate graph before:

image

Workspace crate graph after:

image


youknowone:
How about moving wtf8 to project root? When it is placed under common, it looks like only common can directly use wtf8 but nothing other

coolreader18:
Hmm, fair.

compiler/literal/src/escape.rs
```diff
@@ -35,20 +37,32 @@ pub struct EscapeLayout {
...
         self.layout().len != Some(self.source_len())
     }
 
-    fn write_source(&self, formatter: &mut impl std::fmt::Write) -> std::fmt::Result;
+    /// Write the body of the string directly to the formatter.
+    ///
+    /// # Safety
+    ///
+    /// This string must only contain printable characters.
+    unsafe fn write_source(&self, formatter: &mut impl std::fmt::Write) -> std::fmt::Result;
```
youknowone, line 58: I am confused about this change.
I expect `UnicodeEscape::new_repr(self.as_wtf8())` always safely print the repr. Then the surrogate must be checked somewhere before printing. Doesn't it mean we always print &str, but not wtf8?

coolreader18: We always print `str`, yes. I did this because we were already treating `write_source` as an unsafe function, per the preëxisting comment in `AsciiEscape::write_source`:

```rust
            // SAFETY: this function must be called only when source is printable ascii characters
```

That was technically unsound, since anybody could just call `AsciiEscape::new_repr(b"\xff").write_source()` and invoke library UB.

youknowone (approved):
👍

PR #5634, opened 2025-03-28
Bump cranelift to 0.118

coolreader18:
Removes `hashbrown 0.14` from our dependencies, so now everything just uses `hashbrown 0.15` (except for a proc macro pulled in by tcl :unamused: )


arihant2math:
I'm replacing the tcl crate with raw bindings, which should remove that as well.

youknowone (approved):

PR #5640, opened 2025-03-30
_tkinter pt. 2

arihant2math:
Switches to custom bindgen crates and does some implementation work (`GetVar` is still broken however). 

Also ... ~~building on macos/linux is broken, but that can be fixed on those crates~~ macos/linux require pkg-config now.


youknowone:
The existing tk/tcl libraries seem to use inwelling to generate bindings https://github.com/oooutlk/tcltk/blob/main/tcl/src/lib.rs#L150-L159

If we switch to bindings, how do you think we go similar with those approach to avoid common pitfalls and find a shared point with them in future?

arihant2math:
> The existing tk/tcl libraries seem to use inwelling to generate bindings https://github.com/oooutlk/tcltk/blob/main/tcl/src/lib.rs#L150-L159

I took a look. `inwelling` is simply a manner for upstream crates to send information downstream. `clib` does the building and is just a thin wrapper around bindgen (https://docs.rs/clib/latest/clib/, and https://docs.rs/crate/clib/latest/source/build.rs).

Currently I have a few problems with this approach.
1. `clib` is used as a singular dependency to compile anything, and is depended on twice just by pulling in the tk crate (once for tcl, and the second time for tk). I'm not sure how well cargo handles this, seeing they both call clib with the same version etc.
2. I don't have raw API access, which I seem to need for much of this.
3. `tk` as well as other bindings I have considered all have a bug where `Tcl_DecrRefCount` and `Tcl_IncrRefCount` are not handled properly due to a bindgen limitation in terms of function macros.

I blacklist these macros here: https://github.com/arihant2math/tkinter/blob/d9874a8446cd32dfcd3e71b8e891329f304d2a3b/tk-sys/build.rs#L113

and re-implement them here: https://github.com/arihant2math/tkinter/blob/2957564c74fbb4f25b967f6ac1e2289ed04fd0c0/tk-sys/src/lib.rs#L7

Seeing that they crash the linker if included, unless patched, ... I doubt these libraries are handling memory management well.
Seeing that the right `Free` function isn't called either: https://github.com/search?q=repo%3Aoooutlk%2Ftcltk%20Free&type=code

> If we switch to bindings, how do you think we go similar with those approach to avoid common pitfalls and find a shared point with them in future?

I think with bindings we will be able to mirror the cpython implementation well enough that most issues should be avoided. I think the current set of tk libraries aren't ready for something as broad as tkinter.

arihant2math:
Also I would ideally have contributed to https://github.com/MisterEggnog/tcltk-rs, but it seems to be unmaintained and the author seems to be inactive.

youknowone:
Thank you so much for explanation

youknowone (approved):
The changes look good.
Are you going to individually maintain tk bindings? Otherwise could you consider to transfer the repository to RustPython org?

arihant2math:
I don't have too much of a use for them, so I'll send a transfer request after I clean it up and add basic CI.

arihant2math:
Also since I'm not an admin of the RustPython org, I'd have to transfer it to you first, and you can transfer it to the org.

stdlib/src/tkinter.rs
```diff
@@ -4,83 +4,478 @@ pub(crate) use self::_tkinter::make_module;
...
+    unsafe impl Send for TclObject {}
+    unsafe impl Sync for TclObject {}
+
+    #[pyclass]
+    impl TclObject {}
+
+    static QUIT_MAIN_LOOP: AtomicBool = AtomicBool::new(false);
+
     #[pyattr]
     #[pyclass(name = "tkapp")]
```
youknowone, line 77: please add tkapp to `python-more.txt`

youknowone (changes requested):
Thank you for this headache driving work

youknowone (approved):

PR #5641, opened 2025-03-31
Remove some unncessary dependencies

coolreader18:
Specifically, `rand` and `rustc_version` - we actually use very little of these dependencies' capabilities, and it's quite simple to swap them out with a 2-3 line function.

### DIFF
--- a/vm/build.rs
+++ b/vm/build.rs
@@ -20,10 +20,7 @@ fn main() {
     );
     println!("cargo:rustc-env=RUSTPYTHON_GIT_TAG={}", git_tag());
     println!("cargo:rustc-env=RUSTPYTHON_GIT_BRANCH={}", git_branch());
-    println!(
-        "cargo:rustc-env=RUSTC_VERSION={}",
-        rustc_version::version().unwrap()
-    );
+    println!("cargo:rustc-env=RUSTC_VERSION={}", rustc_version());
 
     println!(
         "cargo:rustc-env=RUSTPYTHON_TARGET_TRIPLE={}",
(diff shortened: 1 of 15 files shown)


vm/build.rs
```diff
@@ -61,7 +58,12 @@ fn git(args: &[&str]) -> String {
     command("git", args)
 }
 
-fn command(cmd: &str, args: &[&str]) -> String {
+fn rustc_version() -> String {
+    let rustc = env::var_os("RUSTC").unwrap_or_else(|| "rustc".into());
+    command(rustc, &["-V"])
+}
```
youknowone, line 64: I agree about rand to reduce runtime efficiency, but do we really need to remove rustc_version? We only use it on build step and we don't need to take this extra code to maintain

coolreader18: It also pulls in the `semver` crate to let you do semantic version matching, which we fully don't use at all. I do agree that a build dependency is not a very big deal, but the part of rust_version we use is basically exactly equivalent to this code, and we already have the `command` function here.

coolreader18: I was actually gonna switch this to use `rustc_version::VersionMeta.short_version_string` instead of the plain 1.X.Y version (it's `rustc 1.85.0 (4d91de4e4 2025-02-17)` which is more similar to what pypy has: `[PyPy 7.3.18 with GCC 14.2.1 20250128]`) and then I realized that `short_version_string` is literally just the output of `rustc -V`, with no parsing of `rustc -Vv` needed at all.

youknowone (approved):

PR #5643, opened 2025-03-31
Fix float parsing

coolreader18:


compiler/literal/src/complex.rs
```diff
@@ -0,0 +1,73 @@
...
+        float::to_string(im)
+    };
+    im_part.push('j');
+
+    // positive empty => return im_part, integer => drop ., fractional => float_ops
+    let re_part = if re == 0.0 {
+        if re.is_sign_positive() {
+            return im_part;
+        } else {
+            re.to_string()
```
youknowone, line 18:
```suggestion
            "0"
```
or
```suggestion
            "0.0"
```
?

Calling specifically `f64::to_string()` only for zero and `float::to_string(re)` for non-zero looks like a bug or unnecessary

youknowone (changes requested):

compiler/literal/src/complex.rs
coolreader18: This was copied over from the old implementation, but I agree.

youknowone: oh, sorry!

youknowone (approved):

PR #5652, opened 2025-04-01
Hmac working without generics, and test not working yet

Shakil582:
This is a draft PR just to show what I have so far, the things I want to do before pulling in the PR are: 

1. Get the unit tests of test_hmac.py working
2. make a generic function to use within the new_hash_hmac() functions for initialization, I just tried very hard to have it work like the hashlib stuff using generics, but I think because CoreProxy needs a type at compile time I can't initialize an hmac<x> dynamically like it's being done for the HashWrapper. Might still play around with that a little, and if anyone know's a way I can use generics I would appreciate the advice, but I couldn't find a way to do it without having to pass a generic to HmacWrapper (as in it would have to be HmacWrapper<T> which then I don't think I can use in PyHmac because I'm guessing it needs to be generic free for the front end). 
3. Play around with NewHmacHashArgs since I get an error when trying to initialize hmac without msg, and digestmod should be able to take a function. I have a feeling the msg thing might be something deeper because it complains about the function in hmac.py expecting 0 arguments. 

Other than that, playing with what I have vs the cpython version they output the same bytes when digesting.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -990,6 +990,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dfa686283ad6dd069f105e5ab091b04c62850d3e4cf5d67debad1933f55023df"
 
 [[package]]
+name = "hmac"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6c49c37c09c17a53d937dfbb742eb3a961d65a994e6bcdcf37e7399d0cc8ab5e"
+dependencies = [
+ "digest",
+]
+
+[[package]]
 name = "home"
 version = "0.5.11"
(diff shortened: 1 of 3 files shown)


arihant2math:
Running `cargo fmt` will fix part of the "Check Rust code with rustfmt and clippy" check.
Running the following and fixing all errors/warnings should fix the other tests:
```
cargo clippy--workspace --all-targets
```

Shakil582:
So I gave up on passing in function for a bit since HmacCore requires strict traits on the class passed into it's generic. Maybe functions can be easily done if we make the function name public and pass in the relevant hash function into it depending on the function name, but that might not be worth it since I'm sure there's reasons not to make pyfunction names public, and also it would only work for the specific hashlib hash functions. I guess the best option would maybe be an HmacCore rewrite where we don't require as strict compile time checks on the passed hash function.

PR #5654, opened 2025-04-01
Cleanup whats_left.py

arihant2math:
Since we target 3.13 now, we don't need to exclude the PEP 594 modules, as they have been removed.
Also adds a features flag to specify which features should be built.

### DIFF
--- a/whats_left.py
+++ b/whats_left.py
@@ -62,46 +62,13 @@ def parse_args():
 
 args = parse_args()
 
-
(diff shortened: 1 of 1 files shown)


youknowone:
did we fully remove PEP 594 modules?

youknowone:
oh, maybe just not necessary because it is used on cpython repository

youknowone (approved):

PR #5658, opened 2025-04-01
Update faulthandler

arihant2math:


arihant2math:
Currently I can't get the `PyMutex` to work in `threading` and `not(threading)` easily.

youknowone:
I'd better to work faulthandler based on this patch.
#6400 and #6406 covers it

PR #5661, opened 2025-04-03
Fix doc warnings

arihant2math:
Also adds doc build to CI to prevent future warnings from cropping up.


arihant2math:
The problem is with cspell checking all the diffed files and finding issues.

PR #5674, opened 2025-04-05
Improve naming from proc macros

arihant2math:
Currently we have `magic` and `name`.

Initially, I was rather confused by `magic` and didn't understand that it just added underscores.
This pr will try to streamline the naming abilities by reducing the usage of `name`, removing `magic`, and adding a single attribute to handing naming modifiers.

This also makes `SimpleItem` allow these.

I try to imitate `serde` when possible.

## Modifiers
- magic` surround with double underscores
- private: prefix with underscore
- (snake|pascal|constant, etc) case: turn into that case
To my knowledge these are basically mutually exclusive.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -403,6 +403,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "convert_case"
+version = "0.8.0"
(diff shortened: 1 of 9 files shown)


arihant2math:
Fair enough, closing.

PR #5678, opened 2025-04-07
Use mimalloc for better performance

### LINKED ISSUE #5526: [RFC] Custom Allocator
## Summary

Consider the use of a faster allocator such jemalloc or mimalloc

## Detailed Explanation

CPython has a custom allocator which is optimized for allocating small objects (PyMalloc), which helps a lot with performance. mimalloc specifically is known to be comparatively fast, and compared to the rest of the executable shouldn't affect binary size too much. This would improve speed when allocating large amounts of heap or just in general.

An issue could be a larger binary size.

## Unresolved Questions

- Which allocator works best for performance (benchmarking this could probably be a good idea)

arihant2math:
I measured around a 10% performance improvement on pystones (better benchmarks coming soon).

Fixes: #5526

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -40,6 +40,8 @@ dirs = { package = "dirs-next", version = "2.0.0" }
 env_logger = { version = "0.9.0", default-features = false, features = ["atty", "termcolor"] }
 flamescope = { version = "0.1.2", optional = true }
 
+mimalloc = "0.1"
+
 [target.'cfg(windows)'.dependencies]
 libc = { workspace = true }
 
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -77,6 +79,7 @@ opt-level = 3
 
 [profile.release]
 lto = "thin"
+debug = true
```
youknowone, line 82: is this intended?

youknowone:
@theoparis Do you have any concern? Please leave a comment if you have opinion about it.

theoparis:
The only concern I had was that there is no way to disable mimalloc without modifying rustpython sources after this is merged. I've encountered crashes with jemalloc and chromium's allocator when another allocator is set globally via LD_PRELOAD for example - but maybe this isn't an issue with mimalloc?

arihant2math:
Not an unreasonable request.
I'll update this.

arihant2math:
I seem to have found a fix, I think I'll just make it environment variable only for now.

PR #5679, opened 2025-04-07
First steps towards native subclassing

coolreader18:
Not very ergonomic yet, and there are definitely some soundness holes atm, but I aim to fix those in a follow-up.

### DIFF
--- a/derive-impl/src/util.rs
+++ b/derive-impl/src/util.rs
@@ -187,6 +187,31 @@ impl ItemMetaInner {
         Ok(value)
     }
 
+    pub fn _optional_parse<T: syn::parse::Parse>(&self, key: &str) -> Result<Option<T>> {
+        let value = if let Some((_, meta)) = self.meta_map.get(key) {
+            let Meta::NameValue(syn::MetaNameValue {
+                value:
+                    syn::Expr::Lit(syn::ExprLit {
+                        lit: syn::Lit::Str(lit),
+                        ..
+                    }),
+                ..
+            }) = meta
+            else {
+                bail_span!(
+                    meta,
+                    "#[{}({} = ...)] must exist as a string",
+                    self.meta_name(),
+                    key
+                )
+            };
+            Some(lit.parse()?)
+        } else {
+            None
+        };
+        Ok(value)
+    }
+
     pub fn _has_key(&self, key: &str) -> Result<bool> {
         Ok(matches!(self.meta_map.get(key), Some((_, _))))
     }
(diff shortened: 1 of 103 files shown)


coolreader18:
The cspell errors seem to be a false positive, since it's flagging words that were there before this PR.

arihant2math:
Yeah it does that a lot. I'll send a pr in to fix it.

arihant2math:
Probably because it checks the entire file instead of just the changes (from what I see).

arihant2math:
The issue is it runs this command, which fails sometimes:
```
Error: Error: Command failed: git diff-tree --no-commit-id --name-only -r 8dc17180029337676395154fc33b513678ea4a8f eb2d2a8b7b4290f893cd352341530e5c3f0339bc
```

stdlib/src/bz2.rs
```diff
@@ -172,7 +172,7 @@ mod _bz2 {
 
     #[pyattr]
     #[pyclass(name = "BZ2Compressor")]
-    #[derive(PyPayload)]
+    #[derive()]
```
youknowone, line 175: empty derive?

extra_tests/snippets/stdlib_ctypes.py
```diff
@@ -266,8 +266,8 @@ def LoadLibrary(self, name):
 
 cdll = LibraryLoader(CDLL)
 
-test_byte_array = create_string_buffer(b"Hello, World!\n")
-assert test_byte_array._length_ == 15
+# test_byte_array = create_string_buffer(b"Hello, World!\n")
```
youknowone, line 269: is this intended?

derive-impl/src/util.rs
```diff
@@ -187,6 +187,51 @@ impl ItemMetaInner {
...
+    }
+
+    pub fn _optional_ident(&self, key: &str) -> Result<Option<&Ident>> {
+        let Some((_, meta)) = self.meta_map.get(key) else {
+            return Ok(None);
+        };
+        if let Meta::NameValue(syn::MetaNameValue {
+            value:
+                syn::Expr::Path(syn::ExprPath {
+                    qself: None, path, ..
```
youknowone, line 218: please add `qself` to `rust-more.txt`

vm/src/class.rs
```diff
@@ -10,9 +12,15 @@ use crate::{
 };
 use rustpython_common::static_cell;
 
-pub trait StaticType {
+/// A type that represents a statically-allocated Python class.
+///
+/// # Safety
+///
+/// `type_id` must be accurate.
```
youknowone, line 19: I understand wrong type_id can cause a bug, but not sure how it can make unsafety. Do you have a scenario?

youknowone (changes requested):

youknowone:
@coolreader18 I was totally forgetting about this. The payload layout is recently heavily changed. Is this still applicable?

PR #5686, opened 2025-04-10
Upgrade parts of test.support

arihant2math:


youknowone:
The support patches look good.
They are pending becuase these patches adding new files but not used anywhere - which means it just can be a not working code due to lack of other dependencies, but not able to test right now.
Then they can be merged at the last minute of updating test.support, or when actually other code requires these files.

Usually keeping a patch unmerged is not a good idea due to gradually increasing conflict and the patch will get old. But adding new files usually don't make conflict, so I think   unmerged state exceptionally give more benefits than merged state about these patches

Issue #5690, opened 2025-04-11
Panic when using Windows GUI app


marcomq:
## Summary

I tried to use RustPython in a windows GUI app, so that python code is running in the background and the result is displayed in a GUI. Unfortunately, it seems that stdin/stdout is required. I get a `essential initialization failed` panic error
```sh
thread 'main' panicked at C:\Users\MyUser\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\rustpython-vm-0.4.0\src\vm\mod.rs:334:30:
essential initialization failed
```

I also have a sample tauri application here that can replicate the issue:
https://github.com/marcomq/tauri-plugin-python/tree/main/examples/plain-javascript

The gui is prevented with
```rust
#![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
```

I could also uncomment this line, but this would then always show a console in the background of the windows app, which is kind of a very ugly workaround.
I will check if I can maybe fix it locally, but I just wanted to create this issue first for documentation.

## Expected

It would be nice if it wouldn't be required to have a console, similar to wasm/wasi mode.

## Actual

The issue can only be replicated on windows release without console, so there is probably no real error visible and the application just doesn't start up.

youknowone:
optional stdio issue may be related

marcomq:
It seems that it is possible to re-enable the console for windows when performing
```rust
        use windows::Win32::System::Console::{AttachConsole, ATTACH_PARENT_PROCESS};
        let _ = unsafe { AttachConsole(ATTACH_PARENT_PROCESS) };
```
But unfortunately, this doesn't seem to fix the panic issue.


So I created this PR:
https://github.com/RustPython/RustPython/pull/5691/files

I also tried to just disable std::err and std::in, but this somehow didn't work and caused the same error.
I actually didn't really check where the panic was coming from. So there is probably a technically better solution out there. But the solution from PR works fine for me. Not sure if somebody actually needs more.

youknowone:
There was also #5420.

@arihant2math @coolreader18 we seem to merge one of them, right? Could you help for decision and reviews?

arihant2math:
I was searching through, since I remember using `println!` in tauri, and found the following:
[https://doc.rust-lang.org/std/io/struct.Stdout.html#note-windows-portability-considerations](https://doc.rust-lang.org/std/io/struct.Stdout.html#note-windows-portability-considerations).

[...]

PR #5691, opened 2025-04-11
Optionally disable console by feature

marcomq:
Going to fix
https://github.com/RustPython/RustPython/issues/5690

Optionally disable stdin, stdout and stderr for example for windows GUI apps, where there is no console and where the previous code panicked.


youknowone (changes requested):
Thanks! Did this patch solved your problem?

One change request. Rust features are expected to be additive. In other words, feature `stdio` is better than `disable-stdio`.

marcomq:
>  Rust features are expected to be additive. In other words, feature stdio is better than disable-stdio.

Oh, thx. You are completely right! The docs actually mention this special case
`A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. A feature should not introduce a SemVer-incompatible change. For example, if you want to optionally support [no_std](https://doc.rust-lang.org/reference/names/preludes.html#the-no_std-attribute) environments, do not use a no_std feature. Instead, use a std feature that enables std.`

Always good to learn something new :) 
I will modify the PR.

marcomq:
Tbh - I think 
https://github.com/RustPython/RustPython/pull/5420
actually looks better.
It looks like it would allow calling "print" and just prevents writing to stdout. My version would raise an error when calling "print".

marcomq:
I guess I will try to resolve the conflict in https://github.com/RustPython/RustPython/pull/5420 and check if it solves my issue.
Also, I'm probably going to name the feature `quiet-stdio` and it should disable the stdio if active.
I would prefer this for following reasons:
- Many projects have just included RustPython-vm and `stdio` would just be a standard feature of RustPython but not of the VM. In those projects, they would loose stdio.
- Some projects have already a "disable-default" and define all features that are needed manually. Those would also loose stdio.
- The "disable-default" doesn't guarantee that a feature is actually disabled. This is usually no problem, but it might cause issues here, as the stdio might be still active due to a foreign dependency.

marcomq:
So - I tried to implement https://github.com/RustPython/RustPython/pull/5420 but it somehow didn't solve my issue. I still got these errors locally during initialization.
I also had a short look on the original issue - the non-UTF8 error message. But I wasn't able to locate where the error message actually originated and what the proper handling would be.

So - this is would currently be my approach of this issue.
Let me know if this is ok or if I still should change it to `stdio` instead of `quiet-stdio`

arihant2math:
Works for me.

arihant2math (approved):

coolreader18:
Sorry for the delay and for the mess that this is in general - I just merged #5420; does that approach work for your problem?

youknowone:
@marcomq could you check your problem is solved or still need more work like this patch?

youknowone:
We now have stdio feature. Thank you so much!

Issue #4072, opened 2022-08-14
TextIOWrapper.detach


youknowone:
## Feature

required for test_mailbox

ShaharNaveh:
Closed by https://github.com/RustPython/RustPython/pull/6267

PR #5695, opened 2025-04-14
Failure marker

arihant2math:
Working script to auto-mark rustpython failures.
Has a few limitations but should work in 90% of the cases.
Ideally I'd fix the class/subclass situation as well as proper ast parsing.


youknowone:
How do you run this code and what does it output?

arihant2math:
Currently I need to patch it a little ... but the idea is that the workflow is the following:

1. Make some changes to Lib/ and Lib/test/
2. Find which tests are failing
3. Run `python ./scripts/fix_test.py --test test_venv --path ./Lib/test/test_venv.py` or equivalent for the test from the project root.
4. Actually fix the test.

This works best for platform-independent tests.

The limitations are:
1. Can't handle unexpected successes
2. Can't handle entire test crashes
3. Can't handle subclasses and nested classes etc. (I'm working on a fix for this though)

arihant2math:
Having a working `--junit-xml` would help with finding failing tests, instead of parsing stdout.
Because some tests like printing out to stdout, ruining the parser.

#4072 is the blocking issue.

youknowone (changes requested):
Please add the comment to the header of the script as documentation

arihant2math:
Done.

youknowone (approved):
👍

PR #5700, opened 2025-04-15
Cleanup match statement codegen

arihant2math:
#5628 was a little chaotic.

- Remove `Instruction::IsOperation`
- Cleans up some error handling (less panicking due to `todo!`)
- Handle stack underflows with more informative error
- Makes code more rust-like
- Cleans up variable names for spellcheck
- Removes match test because we have a test in `extra_tests` and the test only checks for compile.


arihant2math:
Matching classes causes a stack underflow, no clue why. That's for a different pr though.

compiler/codegen/src/compile.rs
```diff
@@ -1944,32 +1944,35 @@ impl Compiler<'_> {
...
-        // Check if the name is forbidden for storing.
-        if self.forbidden_name(name.as_str(), NameUsage::Store)? {
-            return Err(self.compile_error_forbidden_name(name.as_str()));
-        }
+        match n {
+            // If no name is provided, simply pop the top of the stack.
+            None => {
+                emit!(self, Instruction::Pop);
+                Ok(())
+            }
```
coolreader18, line 1952: You could also use let-else, to prevent rightward drift.
```rust
        let Some(name) = n else {
            // If no name is provided, simply pop the top of the stack.
            emit!(self, Instruction::Pop);
            return Ok(());
        };
```

compiler/codegen/src/compile.rs
```diff
@@ -2155,10 +2158,7 @@ impl Compiler<'_> {
             for ident in attrs.iter().take(n_attrs).skip(i + 1) {
```
coolreader18, line 2158: I know this isn't the code you changed, but if you're looking to make it more rusty, you could use `.enumerate()` here for the outer loop (`for (i, attr) in attrs.iter().enumerate()`) and then just `for ident in &attrs[i + 1..]` for the inner loop.

coolreader18 (approved):
lgtm! lmk when you feel it's ready to be merged.

youknowone (approved):

PR #5709, opened 2025-04-17
Switch to `libbz2-rs-sys` and finish bz2 impl

coolreader18:
I realized that [`libbz2-rs-sys`](https://docs.rs/libbz2-rs-sys) exists, by the same folks who make `libz-rs-sys` - it's a Rust reimplementation of `libbz2`. This means we can do that same thing as in #5562, and avoid having to cross-compile C. I then realized after I removed the feature flag that we weren't actually running bz2 regrtests in CI, because we weren't passing it as a feature to cargo build, and so something like half of the tests in `test_bz2.py` were failing. So, I more or less finished up the impl, by making some code in `zlib.rs` generic over the specific `Decompress` struct.


fanninpm:
See also #5605

arihant2math:
Huh I'll switch to this decompressor implementation on my pr.

arihant2math:
This one is better than mine ... `test_shutil` is the only fail (due to 1 new exposed test) and there aren't too many expected failures comparatively. I'll close mine after this is merged.

youknowone:
please check https://github.com/RustPython/RustPython/pull/5605 and rebase on it or pick some idea if possible

coolreader18:
Oh, shoot, sorry - I didn't see that that PR was open.

stdlib/src/bz2.rs
```diff
@@ -12,28 +12,48 @@ mod _bz2 {
         object::{PyPayload, PyResult},
         types::Constructor,
     };
+    use crate::zlib::{
```
arihant2math, line 15: Since there is at-least 1 other decompression module that uses the same format (`_lzma`). I think this should be moved someplace common.

arihant2math (changes requested):

stdlib/src/bz2.rs
coolreader18: I agree - I was planning on doing that in a followup, to reduce the amount of code movement in this pr.

arihant2math (approved):

Lib/test/test_bz2.py
```diff
@@ -1001,6 +1005,8 @@ def test_encoding_error_handler(self):
                 as f:
             self.assertEqual(f.read(), "foobar")
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
youknowone, line 1009: @coolreader18 @arihant2math what cause this regression? could this be fixed in future?

arihant2math: A flag probably needs to be added to the decompressor state. I believe the issue is the `wt` and `rt` formats. Although I'm not to sure, I suppose looking at the cpython source might help.

coolreader18: The reason this needed to be flagged now is because the whole test file wasn't running at all before - bz2 was not included in the `--features=stdlib,threading,...` flag in CI, so the module wasn't even getting compiled or tested at all.

arihant2math:
I checkout out of this pr for #5717 so try not to change the trait too much 😄 .

Lib/test/test_bz2.py
youknowone: Oh, didn't noticed that. Thanks!

stdlib/src/zlib.rs
```diff
@@ -142,18 +145,18 @@ mod zlib {
...
 
     #[derive(Clone)]
-    struct Chunker<'a> {
+    pub(crate) struct Chunker<'a> {
         data1: &'a [u8],
         data2: &'a [u8],
     }
     impl<'a> Chunker<'a> {
-        fn new(data: &'a [u8]) -> Self {
+        pub(crate) fn new(data: &'a [u8]) -> Self {
```
youknowone, line 153:
```suggestion
        pub fn new(data: &'a [u8]) -> Self {
```
When the struct is `pub(crate)`, only `pub` here is automatically `pub(crate)`

youknowone (approved):

stdlib/src/zlib.rs
coolreader18: True, I think I'll change that in the follow-up.

PR #5713, opened 2025-04-18
Fix _suggestions module init

arihant2math:
Renames the module init from suggestions to _suggestions.

My bad


youknowone (approved):

youknowone:
any cpython tests using this module?

PR #5716, opened 2025-04-19
Cleaner panic output for codegen errors

arihant2math:
Replacement of #5555. That attempted to just make it an error. But the main issue is that panicking leads to no source code or even file location. This simply outputs that before the panic (if the appropriate function is called). All the main places I have seen panics have been patched.

### DIFF
--- a/compiler/codegen/src/compile.rs
+++ b/compiler/codegen/src/compile.rs
@@ -212,7 +212,7 @@ macro_rules! emit {
     };
 }
 
-fn print_location(zelf: &Compiler<'_>) {
+fn eprint_location(zelf: &Compiler<'_>) {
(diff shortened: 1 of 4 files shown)


compiler/codegen/src/compile.rs
```diff
@@ -210,6 +212,43 @@ macro_rules! emit {
     };
 }
 
+fn print_location(zelf: &Compiler<'_>) {
```
youknowone, line 215:
```suggestion
fn eprint_location(zelf: &Compiler<'_>) {
```

youknowone:
@coolreader18 Could you review this patch?

coolreader18 (approved):
LGTM!

PR #5719, opened 2025-04-20
copilot-instructions

youknowone:
If you have change requests, do not manually edit the file but ask a language model to add the contents you want to add. That affect machine-readability.

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,186 @@
+# GitHub Copilot Instructions for RustPython
+
+This document provides guidelines for working with GitHub Copilot when contributing to the RustPython project.
+
+## Project Overview
+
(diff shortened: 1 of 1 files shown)


arihant2math:
According to https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot

it seems like it's supposed to be `.github/copilot-instructions.md` not `.github/workflows/copilot-instructions.md`.

youknowone:
huh, Copilot review is so useless not to catch this

auvipy:
> huh, Copilot review is so useless not to catch this

it is indeed! is there any way to use claude?

youknowone:
I never used, so this is not a recommendation.
check products like code rabbit if you are interested in review service.
Doubt for claude at the moment. It is very expensive

PR #5725, opened 2025-04-21
Implement nt.getlogin

arihant2math:


youknowone (approved):
test_getlogin exists in test_os, but marked as skip

PR #5727, opened 2025-04-21
Fix mmap aborting with invalid fd in debug mode

xbjfk:
When running test_mmap without release, it fails with a safety violation. This is because `File::from_raw_fd` is called from an invalid fd (the one from python). Instead, to fix this while avoiding TOCTOU, use `fstat` from the `nix` crate instead
```
test_bad_file_desc (test.test_mmap.MmapTests.test_bad_file_desc) ... fatal runtime error: IO Safety violation: owned file descriptor already closed
zsh: abort      cargo run -- -m test test_mmap -v
```


youknowone (approved):
Thanks, it looks better. Thank you for contributing!

PR #5730, opened 2025-04-21
More nt implementation

arihant2math:
- `nt.listvolumes`
- `nt._path_exists` and similar


youknowone:
`nt._path_exists` doesn't look like to be included here.
I will not retrigger CI because nt is not related to ubuntu build

vm/src/stdlib/nt.rs
```diff
@@ -424,6 +424,39 @@ pub(crate) mod module {
         Ok(vm.ctx.new_list(drives))
     }
 
+    #[pyfunction]
+    fn listvolumes(vm: &VirtualMachine) -> PyResult<PyListRef> {
+        let mut volumes = vec![];
+        let mut buffer = [0u16; 257];
```
youknowone, line 430: 257 is a very specific value. was there any comment about the magic number?

youknowone (commented):
was there any related test for listvolumes?

vm/src/stdlib/nt.rs
arihant2math: MAX_PATH + 1 (windows max path is 256).
I will leave a comment

youknowone: If it is MAX_PATH, can we do like:
```suggestion
        let mut buffer = [0u16; MAX_PATH + 1];
```

vm/src/stdlib/nt.rs
```diff
@@ -445,6 +445,39 @@ pub(crate) mod module {
...
```
youknowone, line 451: If it is related to constant, is this possible?
```suggestion
        let mut buffer = [0u16; MAX_PATH + 1];
```

youknowone:
Thanks! #6373 added it

PR #5739, opened 2025-04-24
Fix openssl error reasons

arihant2math:
This isn't how cpython does it, but it should work just fine.
It can be updated/expanded as needed for tests.


arihant2math:
Somehow hang disappeared for no reason?

youknowone (approved):

PR #5743, opened 2025-04-24
Support multiline function/class definitions in REPL and InteractiveConsole

aneeshdurg:
This makes it possible to define functions/classes interactively.

### DIFF
--- a/Lib/codeop.py
+++ b/Lib/codeop.py
@@ -66,7 +66,10 @@ def _maybe_compile(compiler, source, filename, symbol):
                 compiler(source + "\n", filename, symbol)
                 return None
             except SyntaxError as e:
-                if "incomplete input" in str(e):
+                strerr = str(e)
+                if source.endswith(":") and "expected an indented block" in strerr:
+                    return None
+                elif "incomplete input" in str(e):
                     return None
                 # fallthrough
 
--- a/Lib/codeop.py
+++ b/Lib/codeop.py
@@ -66,6 +66,8 @@ def _maybe_compile(compiler, source, filename, symbol):
                 compiler(source + "\n", filename, symbol)
                 return None
             except SyntaxError as e:
+                # XXX: RustPython; support multiline definitions in REPL
+                # See also: https://github.com/RustPython/RustPython/pull/5743
                 strerr = str(e)
                 if source.endswith(":") and "expected an indented block" in strerr:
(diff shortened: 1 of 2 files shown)


aneeshdurg:
Please let me know if there's a good place to add tests for this!

This is definitely a short-term fix though, and a better long term solution would be do what cPython does, and have a subclass of SyntaxException errors that indicate that it could be incomplete input. Also, it would probably be better to have the shell just execute `code.InteractiveConsole().interact()` instead of maintaining two implementations - I'd love to take a stab at that if that's something you'd like to see!

arihant2math:
Hopefully once ctypes works _pyrepl will become the default, but this is a good fix.

arihant2math:
Actually I would like to test this first.

aneeshdurg:
I'd be happy to add tests if there's already somewhere with tests of the REPL? 
If there isn't, I'd also be interested in helping add those, but that seems like it should be a separate PR.

arihant2math:
I expected the cpython test-suite to have repl tests, but it doesn't. We don't really need tests I suppose.
I apologize for the confusion.

aneeshdurg:
@arihant2math just to confirm is there anything you need from me to proceed with review? I've got some other patches I'm working on to make this work more generally, and to also enable `PY_CF_ALLOW_INCOMPLETE_INPUT` in the python parser/compiler

arihant2math:
LGTM.
@youknowone please review/merge.

youknowone:
I am sorry. Just distracted a bit by recent CI trouble. The patch looks good

Lib/codeop.py
```diff
@@ -66,7 +66,10 @@ def _maybe_compile(compiler, source, filename, symbol):
                 compiler(source + "\n", filename, symbol)
                 return None
             except SyntaxError as e:
-                if "incomplete input" in str(e):
+                strerr = str(e)
```
youknowone, line 69:
```suggestion
                # XXX: RustPython; support multiline definitions in REPL
                # See also: https://github.com/RustPython/RustPython/pull/5743
                strerr = str(e)
```
Because this is a part of standard library from CPython, leaving comment helps not to let them unintentionally deleted by udpate

youknowone (approved):

youknowone:
Thank you for contributing!

PR #5745, opened 2025-04-25
Check what made hang

youknowone:
This PR is not planned to be merged.

just checking what's broken


youknowone:
first trial - pass
second trial - hang after test_yield_from

arihant2math:
Perhaps this is a `regrtest` regression?

youknowone:
enabled platform independent tests on every platform.
first run - nothing hung. try again


unrelated note: `test_types` is platform dependent

youknowone:
second run - nothing hang

arihant2math:
hangs ...

youknowone:
only ubuntu experience problem

youknowone:
only running platform-independent test doesn't break test? what's going on

youknowone:
#5760 looks like the reason

PR #5751, opened 2025-04-27
[Do not merge] Find hang

arihant2math:
Tries to find CI hang


youknowone:
still hang

arihant2math:
test_yield_from is clearly not the issue then.
Also I haven't seen it happen yet with when there is a failing test of some sort.

youknowone:
probably this is not by our change but by CI environment change. Then bisect to the library list can be helpful to find the reason

Issue #5711, opened 2025-04-17
TypeError: '|' not supported between instances of '_CallableGenericAlias' and 'NoneType'


starpit:
## Feature

The following fails on rustpython but succeeds on cpython3.12

```python
from collections.abc import Awaitable, Callable
from typing import TypeVar

T = TypeVar("T")

async def abort_signal_handler(
    fn: Callable[[], Awaitable[T]], on_abort: Callable[[], None] | None = None
) -> T:
    pass
```

> Traceback (most recent call last):
  File "/tmp/foo.py", line 7, in <module>
    fn: Callable[[], Awaitable[T]], on_abort: Callable[[], None] | None = None
TypeError: '|' not supported between instances of '_CallableGenericAlias' and 'NoneType'

## Python Documentation or reference to CPython source code

I'm not sure. I have tried comparing RustPython's typing.py to cpython's. There are differences in the treatment of Callable.

https://github.com/python/cpython/blob/main/Lib/typing.py

arihant2math:
RustPython's typing.py is out of date by 5 minor versions or so (3.9 or 3.10 or something). #5590 should fix it, but there is some ast that needs to be updated as well.

starpit:
Thanks for the attention. I tested the #5590 branch, and still see this error. In addition to the above, here is perhaps a quicker use case:

arihant2math:
Isn't Pattern::MatchOr for match statements (it's implemented now, as per #5628)? This is `BinaryOperation::Or`
<details><summary>Disassembly</summary>
<p>
```
$ ./target/debug/rustpython -m dis tmp.py
  1           0 LoadConst            (0)
              1 LoadConst            (("Awaitable", "Callable"))
              2 ImportName           (0, collections.abc)
              3 ImportFrom           (1, Awaitable)
              4 StoreLocal           (1, Awaitable)
              5 ImportFrom           (2, Callable)
              6 StoreLocal           (2, Callable)
              7 Pop

  2           8 LoadConst            (0)
              9 LoadConst            (("TypeVar"))
             10 ImportName           (3, typing)
             11 ImportFrom           (4, TypeVar)
             12 StoreLocal           (4, TypeVar)
             13 Pop

  4          14 LoadNameAny          (4, TypeVar)
             15 LoadConst            ("T")
             16 CallFunctionPositional(1)
             17 StoreLocal           (5, T)

  7          18 LoadConst            (None)
             19 BuildTuple           (1)

[...]

PR #5752, opened 2025-04-27
Regrtest upgrade

arihant2math:
Attempt to fix hang

Minor patches needed because of #5711


fanninpm:
Superseded by #6410 and #6818.

PR #5753, opened 2025-04-27
Callable __or__ patch

### LINKED ISSUE #5711: TypeError: '|' not supported between instances of '_CallableGenericAlias' and 'NoneType'

arihant2math:
Fixes #5711

### DIFF
--- a/Lib/_collections_abc.py
+++ b/Lib/_collections_abc.py
@@ -512,6 +512,10 @@ class _CallableGenericAlias(GenericAlias):
             new_args = (t_args, t_result)
         return _CallableGenericAlias(Callable, tuple(new_args))
 
+    # TODO: RUSTPYTHON patch for common call
+    def __or__(self, other):
+        super().__or__(other)
+
 def _is_param_expr(obj):
     """Checks if obj matches either a list of types, ``...``, ``ParamSpec`` or
     ``_ConcatenateGenericAlias`` from typing.py
--- a/extra_tests/snippets/stdlib_typing.py
+++ b/extra_tests/snippets/stdlib_typing.py
@@ -1,0 +1,10 @@
+from collections.abc import Awaitable, Callable
+from typing import TypeVar
+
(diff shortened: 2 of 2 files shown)


arihant2math:
No clue why spellcheck decided to fail. But the other test is the usual hang.

Lib/_collections_abc.py
```diff
@@ -508,6 +508,10 @@ def __getitem__(self, item):
             new_args = (t_args, t_result)
         return _CallableGenericAlias(Callable, tuple(new_args))
 
+    # TODO: RUSTPYTHON patch for common call
```
youknowone, line 511:
```suggestion
    # XXX: RUSTPYTHON patch for common call
```
This is a subtle commenting convention for local patches. What does common call here mean? Is there related test? Otherwise adding one under `extra_tests` will be great.

youknowone (changes requested):
What's the

arihant2math:
I think windows tests are flaky here.

youknowone:
Ok, the test name is test_subprocess::test_run_with_an_empty_env. If it is fixed after rebase, let's also mark it as skip/flaky

PR #5733, opened 2025-04-22
Replace puruspe to pymath

youknowone:
Might be lower quality, but better compatibility

Repository: https://github.com/RustPython/pymath

PR #5757, opened 2025-04-29
Increase numerical precision of log1p

kralka:
Use Rust native `ln_1p` to increase numerical precision matching what is expected by `test_mtestfile`.

### DIFF
--- a/stdlib/src/math.rs
+++ b/stdlib/src/math.rs
@@ -147,7 +147,7 @@ mod math {
     fn log1p(x: ArgIntoFloat, vm: &VirtualMachine) -> PyResult<f64> {
         let x = *x;
         if x.is_nan() || x > -1.0_f64 {
-            Ok((x + 1.0_f64).ln())
+            Ok(x.ln_1p())
         } else {
             Err(vm.new_value_error("math domain error".to_owned()))
         }


arihant2math:
Does commenting out that line lead to the test being visibly slower?
`@unittest.skip` unconditionally skips the test everywhere, fyi.

youknowone:
yep, please try remove the line and check your test pass

kralka:
@arihant2math @youknowone 

Removing the line and running on a laptop with SSD and https://hub.docker.com/_/rust (linux). Running `RUSTPYTHONPATH=Lib cargo run --release Lib/test/test_math.py MathTests.test_mtestfile` 

- Skipping the test takes `0.0s`
- Not skipping the test takes `0.003s`.
- The test failing (without the change of `(x+1).ln` to `x.ln_1p`) takes `0.006s`.

This is why I left the test skipping since there was a comment that said "TODO: RUSTPYTHON, Taking a few minutes." because it takes a few minutes somewhere (CI like github actions, different operating system?). Perhaps you as authors of https://github.com/RustPython/RustPython/pull/5733 have more information here?

The downside is that Rust documentation is saying that numerical stability of `ln_1p` (and many other functions) could change: https://doc.rust-lang.org/std/primitive.f64.html#method.ln_1p But I assume that is still better than calling a C library over `pymath`. What do you think?

arihant2math:
Usually we skip a test because of CI. I suggest removing that line and pushing it up, because it is likely no longer relevant due to performance gains for some reason or another. If it takes to long the CI will immediately indicate and it can be reverted.

youknowone:
Oh, the test must be my mistake. I checked it and confirmed the test doesn't take long. Thank you for checking.

I think this patch is good unless it is very different from CPython.

youknowone (approved):

kralka:
ack thank you for checking

PR #5758, opened 2025-04-29
Fix windows freeze stdlib

arihant2math:


arihant2math:
I don't know if we even need the symlink, but this is fine for now.

pylib/src/lib.rs
```diff
@@ -9,6 +9,10 @@ pub const LIB_PATH: &str = match option_env!("win_lib_path") {
...
 };
 
-#[cfg(feature = "freeze-stdlib")]
+#[cfg(feature = "freeze-stdlib", not(target_os = "windows"))]
 pub const FROZEN_STDLIB: &rustpython_compiler_core::frozen::FrozenLib =
     rustpython_derive::py_freeze!(dir = "./Lib", crate_name = "rustpython_compiler_core");
+
+#[cfg(feature = "freeze-stdlib", target_os = "windows")]
+pub const FROZEN_STDLIB: &rustpython_compiler_core::frozen::FrozenLib =
+    rustpython_derive::py_freeze!(dir = "../Lib", crate_name = "rustpython_compiler_core");
```
youknowone, line 18: This will cause problem for cargo publish.

youknowone (changes requested):

pylib/src/lib.rs
arihant2math: Ah ... that explains a lot, however it was already broken on windows coming in. I'll see if I can fix the symlink on windows in that case.

youknowone: There probably was a script written by @coolreader18 to copy them. 
Otherwise now windows also has symlink.
If everything cause problem, I vote for moving them under pylib than this change

youknowone: Suggestion: A good first step to resolve this error is adding a CI job to reproduce this problem

arihant2math:
Closing this, because it's trivially easy to get git to support symlinks properly on windows.

youknowone:
@arihant2math could you add it to README for windows users?

PR #5762, opened 2025-04-30
gitignore, folder fix, like __pycache__ 

Undersk0re:
~Fix the path for better usability.~ We can clean the `__pycache__` directory instead of using the downloaded one. It can sometimes cause errors

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,11 +2,11 @@
 /*/target
 **/*.rs.bk
 **/*.bytecode
-__pycache__
+__pycache__/
 **/*.pytest_cache
 .*sw*
 .repl_history.txt
-.vscode
+.vscode/
 wasm-pack.log
 .idea/
 .envrc


arihant2math (approved):

youknowone (approved):
I agree to the changes because the lines intends to point directories.

But I didn't get what you meant by below. Could you explain it more?

> We can clean the __pycache__ directory instead of using the downloaded one.

Undersk0re:
It's better to make sure the __pycache__ folder is empty, to avoid the risk of cache state being shared between multiple users

youknowone:
Could you tell me more details? I cannot imagine any difference before and after about current setting can make shared cache. I know `/` prefix notate the strict directory, but no idea how it goes further like that.

Undersk0re:
If the __pycache__ folder is unintentionally shared on Git, it can be downloaded by multiple users who might read someone else's cache or mistakenly consider it their own.

Undersk0re:
To exclude a folder (anywhere in the project), use: `folder/` (for Linux, Windows and macOS , Git is platform-independent in this context)

...if this right answer for the question

youknowone:
That's not the answer for the question. `__pycache__` is already in the `.gitignore` file. Directory notation narrow the filter down only to directory, to allow `__pycache__` file to be added to git.
I doubt this change give any kind of "better usability" as described in the PR body in practical scenario. Please leave a comment if you can make one.

I merge this not for better usability but to describe the gitignore intention better they were expected to be directory.

PR #5764, opened 2025-05-01
Support incomplete parsing

aneeshdurg:
This PR adds support for incomplete parsing both in the native REPL and in the stdin REPL (in particular multiline input support is expanded for to allow more cases such as multiline strings). The implementation adds compatibility for cpython's `PY_CF_ALLOW_INCOMPLETE_INPUT` compile flag and adds a new exception type similar to python 3.14.

### DIFF
--- a/Lib/codeop.py
+++ b/Lib/codeop.py
@@ -65,12 +65,10 @@ def _maybe_compile(compiler, source, filename, symbol):
             try:
                 compiler(source + "\n", filename, symbol)
                 return None
+            except _IncompleteInputError as e:
+                return None
             except SyntaxError as e:
-                # XXX: RustPython; support multiline definitions in REPL
-                # See also: https://github.com/RustPython/RustPython/pull/5743
-                strerr = str(e)
-                if isinstance(e, _IncompleteInputError):
-                    return None
+                pass
                 # fallthrough
 
     return compiler(source, filename, symbol, incomplete_input=False)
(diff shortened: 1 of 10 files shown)


aneeshdurg:
This brings the project a step closer to having a single REPL implementation! There's still some deficiencies in the stdlib parser for parsing blocks for functions/classes though.

aneeshdurg:
It seems like the test_pickle* failures might be unrelated to this PR?

Lib/codeop.py
```diff
@@ -69,9 +69,7 @@ def _maybe_compile(compiler, source, filename, symbol):
                 # XXX: RustPython; support multiline definitions in REPL
                 # See also: https://github.com/RustPython/RustPython/pull/5743
                 strerr = str(e)
-                if source.endswith(":") and "expected an indented block" in strerr:
-                    return None
-                elif "incomplete input" in str(e):
+                if isinstance(e, _IncompleteInputError):
```
youknowone, line 72: Does this work? Then partially reverting #5743 to the original code will be better, including the upper comment.
```suggestion
                if "incomplete input" in str(e):
```

aneeshdurg: That doesn't work, since `_IncompleteInputError` doesn't necessarily have that string in the error. I've updated the comment and changed this block to just match python 3.15's implementation at least.

Lib/test/test_baseexception.py
```diff
@@ -83,6 +83,8 @@ def test_inheritance(self):
         exc_set = set(e for e in exc_set if not e.startswith('_'))
         # RUSTPYTHON specific
         exc_set.discard("JitError")
+        # RUSTPYTHON specific
```
youknowone:
```suggestion
        # TODO: RUSTPYTHON; this will be officially introduced in Python 3.15
```

PR #5766, opened 2025-05-01
Update README.md about windows symlink fix

arihant2math:


README.md
```diff
@@ -31,6 +31,11 @@ To build RustPython locally, first, clone the source code:
 git clone https://github.com/RustPython/RustPython
 ```
 
+If on windows, run the following to fix the `Lib/` directory symlink:
```
youknowone, line 34:
```suggestion
RustPython uses symlinks to manage python libraries in `Lib/`. If on windows, run the following helps:
```

Does this make sense?

Also, as I know windows developer mode allows user to create symbolic links. Do you know if it also can be an alternative option of this?

arihant2math: I'm not sure, this solution worked very well for me though.

youknowone: @coolreader18 how about you? Is mentioning developer mode symbolic links support a correct description?

coolreader18: I think this makes sense, I'm pretty sure setting `core.symlinks` makes git make real symlinks using developer mode.

coolreader18 (approved):

PR #5773, opened 2025-05-08
chore: migrate settings to ruff.toml

rexledesma:
## Summary

Extract these settings from CI into a centralized place.

## Test Plan

```
RustPython on  rl/migrate-to-ruff-toml via 🐍 v3.13.2 via 🦀 v1.86.0 
❯ uvx --from "ruff==0.11.8" ruff check                          
All checks passed!
```


rexledesma:
I'm used to stacking diffs, so there might be some duplicate code here because the base branch isn't set properly because I'm contributing from a fork.

The only thing to look at here is https://github.com/RustPython/RustPython/pull/5773/commits/7c8e5c9fcda35c7ee5211fc192917e473204f435.

ruff.toml
```diff
@@ -0,0 +1,20 @@
...
+    "benches",
+]
+
+[lint]
+select = [
+    "E9", # pycodestyle (error)
+    "F63", # pyflakes
+    "F7",
+    "F82",
+]
```
youknowone, line 20:
```suggestion
]

```

youknowone (approved):

PR #5774, opened 2025-05-08
chore: add `ruff format --check`

rexledesma:
Depends on https://github.com/RustPython/RustPython/pull/5773.

## Summary

Add a `ruff format` step for the Python files generated by contributors.

Main file to look at here is https://github.com/RustPython/RustPython/pull/5774/commits/4f14a311bcd4a4fc09325f6d7f45816b427b492c#diff-944291df2c9c06359d37cc8833d182d705c9e8c3108e7cfe132d61a06e9133dd.

Note that we ignore formating `syntax_*.py` and `badsyntax_*.py` files as they are specifically testing non standardized code formating.

## Test Plan

```
RustPython on  rl/add-ruff-format via 🐍 v3.13.2 via 🦀 v1.86.0 
❯ uvx --from "ruff==0.11.8" ruff format --check
217 files already formatted
```


extra_tests/snippets/syntax_assignment.py
```diff
@@ -41,13 +41,15 @@
 assert c == 2
 assert d == 3
 
-a, = [1]
+(a,) = [1]
```
youknowone, line 44: Files with `syntax_` prefix is intentionally testing these kind of syntax. Please review those files not to lose this intentions

youknowone (changes requested):

extra_tests/snippets/syntax_assignment.py
arihant2math: Can't ruff be configured to ignore those lints? That would be preferable in my opinion.

rexledesma: Two options here to make immediate progress:

1. We can ignore syntax_*.py in `ruff.toml` entirely
2. We can add ruff directives to each of these `syntax_*.py` files to ignore formatting for the entire file. + add a TODO to refine the directive to the specific line if someone edits the file in the future.

youknowone: 1 will be simpler. we have to add every exceptions for new syntax for 2.

rexledesma: Cool, this should be resolved here: https://github.com/RustPython/RustPython/pull/5774/commits/9d8c46a5bab7f8d5f41cd6bffe95392929c3cfc0#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989

youknowone (approved):
Thanks!

PR #5783, opened 2025-05-12
Dev container update

arihant2math:
I adapted this configuration from deno: https://github.com/denoland/deno/blob/main/.devcontainer/devcontainer.json.

### DIFF
--- a/.devcontainer/DOCKERFILE
+++ b/.devcontainer/DOCKERFILE
@@ -1,0 +1,6 @@
+FROM mcr.microsoft.com/vscode/devcontainers/rust:1-bullseye
+
+# Install clang
+RUN apt-get update \
+  && apt-get install -y clang \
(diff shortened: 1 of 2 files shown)


arihant2math:
Testing has gotten rather flaky lately.

.devcontainer/DOCKERFILE
```diff
@@ -0,0 +1,6 @@
+FROM mcr.microsoft.com/vscode/devcontainers/rust:1-bullseye
+
+# Install clang
+RUN apt-get update \
+  && apt-get install -y clang \
```
youknowone, line 5: do we need clang?

arihant2math: Yeah I think it's a good idea. Bindgen and other things need it, I think.

PR #5784, opened 2025-05-12
More tkinter

arihant2math:
- Fix broken build
- Fix broken getvar (no more panic and actual null termination)
- Fix threading lookup (actual null termination)
- Add globalgetvar
- Fix theoretical mainloop panic

I still have no idea why getvar can't find tk_version and tcl_version, but that's an issue for another time.

### DIFF
--- a/stdlib/src/tkinter.rs
+++ b/stdlib/src/tkinter.rs
@@ -125,7 +125,7 @@ mod _tkinter {
         interactive: i32,
         #[pyarg(any)]
         wantobjects: i32,
-        #[pyarg(any, default = "true")]
+        #[pyarg(any, default = true)]
         want_tk: bool,
         #[pyarg(any)]
         sync: i32,
--- a/stdlib/src/tkinter.rs
+++ b/stdlib/src/tkinter.rs
@@ -306,6 +306,8 @@ mod _tkinter {
             let threaded = {
                 let part1 = String::from("tcl_platform");
                 let part2 = String::from("threaded");
+                let part1 = ffi::CString::new(part1).unwrap();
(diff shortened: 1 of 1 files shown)


stdlib/src/tkinter.rs
```diff
@@ -247,21 +256,45 @@ mod _tkinter {
...
+            args: TkAppGetVarArgs,
+            flags: u32,
+            vm: &VirtualMachine,
+        ) -> PyResult<PyObjectRef> {
+            let TkAppGetVarArgs { name, name2 } = args;
             // TODO: technically not thread safe
-            let name = varname_converter(arg, vm)?;
+            let name = varname_converter(name, vm)?;
 
+            let name = ffi::CString::new(name).unwrap();
```
youknowone, line 275: We avoid unwrap possibly hit by user input in Python runtime. Instead of unwrap, we have `ToPyException` impl for `NulError`. Please convert it.

stdlib/src/tkinter.rs
```diff
@@ -272,8 +272,8 @@ mod _tkinter {
             // TODO: technically not thread safe
             let name = varname_converter(name, vm)?;
 
-            let name = ffi::CString::new(name).unwrap();
-            let name2 = ffi::CString::new(name2.unwrap_or_default()).unwrap();
+            let name = ffi::CString::new(name)?;
```
youknowone, line 275: I doubt this is buildable.

ToPyException requires vm

youknowone (changes requested):

stdlib/src/tkinter.rs
youknowone: @arihant2math could you check this?

PR #5786, opened 2025-05-15
Fix warnings for rust 1.87

coolreader18:


youknowone:
what's happening in test_format?

youknowone (approved):

coolreader18:
Oh! they changed the range for precision in `std::fmt`, lol

PR #5792, opened 2025-05-19
Remove unnecessary `uv run`in README

arihant2math:
Partial revert of #5771.
Two reasons for this:
1. There is no tangible benefit to using uv here, the script has no dependencies.
2. This breaks the run icons on the side of jetbrains editors for those don't use uv

/cc @rexledesma let me know if I've missed something

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,7 +226,7 @@ To enhance CPython compatibility, try to increase unittest coverage by checking 
 Another approach is to checkout the source code: builtin functions and object
 methods are often the simplest and easiest way to contribute.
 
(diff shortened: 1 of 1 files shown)


rexledesma (changes requested):
`uv` will ensure that the script is run on the correct version of Python. The script requires `python>=3.13`: https://github.com/RustPython/RustPython/blob/main/whats_left.py#L38-L44.

rexledesma (approved):
Approving, subject to this comment: https://github.com/RustPython/RustPython/pull/5792#issuecomment-2900159288

arihant2math:
Alright

PR #5794, opened 2025-05-19
Remove macos skips

arihant2math:
Could not reproduce on an arm mac.

## Summary by CodeRabbit

- **Chores**
  - Updated CI workflow to remove macOS-specific test skips. MacOS test runs now only exclude platform-independent tests, aligning test execution across platforms.


youknowone:
the test was hanging

PR #5803, opened 2025-06-15
Fix nightly clippy warnings

youknowone:


youknowone:
@coolreader18 redoxer-action is broken

PR #5814, opened 2025-06-22
Convert `new_X_error` to use a macro

ShaharNaveh:
## Summary by CodeRabbit

- **Refactor**
  - Streamlined internal exception creation methods to reduce code duplication and improve maintainability. No changes to user-facing functionality or error messages.

### DIFF
--- a/vm/src/vm/vm_new.rs
+++ b/vm/src/vm/vm_new.rs
@@ -13,6 +13,30 @@ use crate::{
     vm::VirtualMachine,
 };
 
+macro_rules! define_exception_fn {
+    // With doc
+    (
+        $fn_name:ident, $attr:ident,
+        doc = $doc:literal
+    ) => {
+        #[doc = $doc]
+        pub fn $fn_name(&self, msg: String) -> PyBaseExceptionRef {
+            let err = self.ctx.exceptions.$attr.to_owned();
+            self.new_exception_msg(err, msg)
+        }
+    };
+
+    // Without doc
+    (
+        $fn_name:ident, $attr:ident
+    ) => {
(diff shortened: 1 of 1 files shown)


vm/src/vm/vm_new.rs
```diff
@@ -607,8 +547,32 @@ impl VirtualMachine {
...
+    // TODO: remove & replace with new_unicode_decode_error_real
+    define_exception_fn!(new_unicode_decode_error, unicode_decode_error);
+
+    // TODO: remove & replace with new_unicode_encode_error_real
+    define_exception_fn!(new_unicode_encode_error, unicode_encode_error);
+
+    define_exception_fn!(
+        new_value_error,
+        value_error,
+        doc = "Create a new python ValueError object.
```
ShaharNaveh: Thought of have this doc as part of the macro and only put the python exception name to template.
so it can look like:
```rust
define_exception_fn!(new_value_error, value_error, ValueError);
define_exception_fn!(new_index_error, index_error, IndexEror);
```

I can do it in a separate PR if you think that it would be beneficial

vm/src/vm/vm_new.rs
```diff
@@ -13,6 +13,30 @@ use crate::{
     vm::VirtualMachine,
 };
 
+macro_rules! define_exception_fn {
+    // With doc
+    (
+        $fn_name:ident, $attr:ident,
```
youknowone:
```suggestion
        fn $fn_name:ident, $attr:ident,
```
Reading code relies on `grep` a lot. Adding `fn` helps it a lot.

vm/src/vm/vm_new.rs
youknowone: I agree to auto-generate docs

PR #5826, opened 2025-06-23
Remove unnecessary string conversions in error message construction

ShaharNaveh:
There are **a lot** more places where we can remove the call of `.to_owned()`, but this is just a start.

Did it here for a couple of files just to show how much code we can remove.

## Summary by CodeRabbit

- **Refactor**
  - Simplified error message creation across multiple modules by removing redundant string conversions.
  - Enhanced internal functions to accept more flexible string inputs for error messages.
  - Improved efficiency and clarity in error handling without changing user-facing behavior or messages.

### DIFF
--- a/vm/src/vm/vm_new.rs
+++ b/vm/src/vm/vm_new.rs
@@ -25,7 +25,7 @@ macro_rules! define_exception_fn {
         pub fn $fn_name(&self, msg: impl Into<String>) -> PyBaseExceptionRef
         {
             let err = self.ctx.exceptions.$attr.to_owned();
-            self.new_exception_msg(err, msg.into())
+            self.new_exception_msg(err, msg)
         }
     };
 }
@@ -117,8 +117,12 @@ impl VirtualMachine {
     /// type is passed in, it may not be fully initialized; try using
     /// [`vm.invoke_exception()`][Self::invoke_exception] or
     /// [`exceptions::ExceptionCtor`][crate::exceptions::ExceptionCtor] instead.
-    pub fn new_exception_msg(&self, exc_type: PyTypeRef, msg: String) -> PyBaseExceptionRef {
-        self.new_exception(exc_type, vec![self.ctx.new_str(msg).into()])
+    pub fn new_exception_msg(
+        &self,
+        exc_type: PyTypeRef,
+        msg: impl Into<String>,
+    ) -> PyBaseExceptionRef {
+        self.new_exception(exc_type, vec![self.ctx.new_str(msg.into()).into()])
     }
 
     /// Instantiate an exception with `msg` as the only argument and `dict` for object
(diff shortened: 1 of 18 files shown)


vm/src/vm/vm_new.rs
```diff
@@ -116,8 +117,12 @@ impl VirtualMachine {
...
     /// [`vm.invoke_exception()`][Self::invoke_exception] or
     /// [`exceptions::ExceptionCtor`][crate::exceptions::ExceptionCtor] instead.
-    pub fn new_exception_msg(&self, exc_type: PyTypeRef, msg: String) -> PyBaseExceptionRef {
-        self.new_exception(exc_type, vec![self.ctx.new_str(msg).into()])
+    pub fn new_exception_msg(
+        &self,
+        exc_type: PyTypeRef,
+        msg: impl Into<String>,
+    ) -> PyBaseExceptionRef {
+        self.new_exception(exc_type, vec![self.ctx.new_str(msg.into()).into()])
```
youknowone: I couldn't find the issue but we once discussed about this changes. We needed a cut to to reduce binary size. I think most of new_*_error is simple wrappers of this function. Then reverting only this one and `new_exception_msg_dict` to using String will be good enough keep the balance of DX and having a single code to create params for new_exception.

ShaharNaveh: If one of the goals of the project is to reduce the binary size, I think this PR should be closed then.

I'm all for a clean, and better code as long as it's not hurting the goals of the project:)

---

If you think it's a good compromise then I'll implement your suggestions, but I'm afraid that you don't want to offend me by just closing the PR lol

youknowone: I think the most outer wrappers becomes more cleaner by your way. By only turning `new_exception_msg` and `new_exception_msg_dict` back to take only String, other changes will not affect binary size.
They are small functions, so must be inlined before.

ShaharNaveh: The binary sizes:

| target | main | branch | diff |
| --- | --- | --- | --- |
| debug | 226548kb (222M) | 226620kb (222M) | +72kb |
| relase | 21836kb (22M) | 21828kb (22M)  | -8kb |


---

This change actually reduces the binary size on the release target? can you please verify that that's also the case on your machine?

youknowone: Interesting.

Build command: `cargo build --release --no-default-features`  (because binary size tweaks usually wants less features)
- Before patch: 15196432
- After patch: 15213472

It increased 17kb for me.

I rebased this patch on current main before build. My arch is aarch64-apple-darwin

ShaharNaveh: Oh, I built mine on `aarch64-linux-android` with `cargo build` (so with the default features).

I can recheck it without the default features, but I don't see the point in that as I assume the majority of RustPython users are not running on an android.

---

How should I proceed with this PR, if at all?

youknowone: I am good with reverting only new_exception_msg and new_exception_msg_dict but keeping anything else as you did.

youknowone: - Before patch: 15212688
- After patch: 15200016

PR #5850, opened 2025-06-27
Upgrade typing to 3.13.5

youknowone:
## Summary by CodeRabbit

* **New Features**
  * Introduced an `override` decorator to mark methods as overrides at runtime.
  * Improved error reporting for protocols with non-method members when using `issubclass()`.

* **Documentation**
  * Expanded docstrings for several typing utilities with new examples and usage guidance.

* **Refactor**
  * Refactored internal instance checking logic for better alignment with expected Python behavior.

* **Tests**
  * Updated test metadata by removing expected failure markers from certain typing tests.

### DIFF
--- a/Lib/typing.py
+++ b/Lib/typing.py
@@ -1987,7 +1987,8 @@ def _allow_reckless_class_checks(depth=2):
     The abc and functools modules indiscriminately call isinstance() and
     issubclass() on the whole MRO of a user class, which may contain protocols.
     """
-    return _caller(depth) in {'abc', 'functools', None}
+    # XXX: RUSTPYTHON; https://github.com/python/cpython/pull/136115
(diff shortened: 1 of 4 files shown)


Lib/typing.py
```diff
@@ -1983,7 +1987,8 @@ def _allow_reckless_class_checks(depth=2):
     The abc and functools modules indiscriminately call isinstance() and
     issubclass() on the whole MRO of a user class, which may contain protocols.
     """
-    return _caller(depth) in {'abc', 'functools', None}
+    # XXX: RUSTPYTHON; https://github.com/python/cpython/pull/136115
```
youknowone, line 1990: https://github.com/python/cpython/issues/136047

PR #5857, opened 2025-06-28
Fix stat module compilation for Redox OS

### LINKED ISSUE #5856: S_IWRITE and S_IEXEC are obsolete and break RustPython on Redox
## Summary

The stat module that was just implemented is using obsolete constants S_IWRITE and S_IEXEC. These constants do not exist on newer OS's like Redox. Please use S_IWUSR and S_IXUSR.

## Details

S_IWRITE and S_IEXEC exist for backwards compatibility reasons and should not be used in new software.

willnode:
Fixes #5856

The replacement constants is exist for redox and should be working: https://github.com/rust-lang/libc/blob/main/src/unix/redox/mod.rs

### DIFF
--- a/vm/src/stdlib/stat.rs
+++ b/vm/src/stdlib/stat.rs
@@ -132,10 +132,10 @@ mod stat {
     #[pyattr]
     pub const S_IWUSR: Mode = 0o0200;
 
-    #[cfg(all(unix, not(target_os = "android")))]
+    #[cfg(all(unix, not(target_os = "android"), not(target_os = "redox")))]
     #[pyattr]
     pub const S_IWRITE: Mode = libc::S_IWRITE;
-    #[cfg(any(not(unix), target_os = "android"))]
+    #[cfg(any(not(unix), target_os = "android", target_os = "redox"))]
     #[pyattr]
     pub const S_IWRITE: Mode = 0o0200;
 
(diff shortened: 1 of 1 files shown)


rw-vanc:
We probably need to fix S_IREAD as well.

I'm not entirely sure why S_IWRITE and S_IEXEC would be needed, but I think you should avoid them if you can.

And why are S_IRUSR, S_IWUSR and S_IXUSR all defined numerically (no libc::) but S_IREAD, S_IWRITE and S_IEXEC are defined using libc:: which will break on many Unixes. Why not define them in terms of S_IRUSR, S_IWUSR, and S_IXUSR?

moreal (approved):

willnode:
@rw-vanc 

I'm not quite understand about `pub const` `S_IWRITE` and `S_IEXEC` is used in the codebase so I'm just following along. The definition for `S_IRUSR`, `S_IWUSR`, and `S_IXUSR` is [already exist](https://github.com/RustPython/RustPython/blob/2bd52121bb5d42f246db709f5f7379fe8c9fbd1f/vm/src/stdlib/stat.rs#L128-L130) in above of my changes and AFAIK `redox` is part of `unix` so no need changes in that part.

The [python constants](https://github.com/RustPython/RustPython/blob/2bd52121bb5d42f246db709f5f7379fe8c9fbd1f/Lib/stat.py#L97) is already using `Unix V7 synonym` though.

rw-vanc:
Yes, I reviewed the original code a little more, and I think a better fix would be to just define S_IREAD, S_IWRITE and S_IEXEC either in terms of S_IRUSR, S_IWUSR and S_IXUSR, or define them numerically always, as was done for the S_I{R,W,X}USR constants.

youknowone:
Thank you! We currently disabled redox CI due to broken github action 

https://github.com/RustPython/RustPython/blob/main/.github/workflows/ci.yaml#L226-L232

Could someone help to fix it? Then this kind of breakage will be prevented

youknowone (approved):

willnode:
I wish I can but currently redoxer itself is broken against recent kernel and the current CI always uses the new kernel. We'll figure out a way to fix it. Thanks for mentioning the CI 👍

PR #5858, opened 2025-06-28
Constify functions & General nitpicks

ShaharNaveh:
If this PR is too large I don't mind splitting it:)

## Summary by CodeRabbit

* **New Features**
  * Many methods and functions across core types, standard library modules, and utilities are now marked as `const`, allowing their use in compile-time constant contexts.
  * Added new method for IPv6 address tuple parsing in the socket module.

* **Refactor**
  * Improved code readability with additional blank lines and minor formatting cleanups.
  * Simplified and modernized some error message constructions and pattern matching expressions.

* **Bug Fixes**
  * Minor corrections to error message formatting and method annotations to ensure proper exposure as Python methods.

* **Style**
  * Enhanced code clarity by reformatting string literals and restructuring some internal logic without changing behavior.

### DIFF
--- a/stdlib/src/socket.rs
+++ b/stdlib/src/socket.rs
@@ -726,19 +726,22 @@ mod _socket {
     }
 
     #[pyfunction]
-    fn htonl(x: u32) -> u32 {
+    const fn htonl(x: u32) -> u32 {
         u32::to_be(x)
     }
+
     #[pyfunction]
-    fn htons(x: u16) -> u16 {
+    const fn htons(x: u16) -> u16 {
         u16::to_be(x)
     }
+
     #[pyfunction]
-    fn ntohl(x: u32) -> u32 {
+    const fn ntohl(x: u32) -> u32 {
         u32::from_be(x)
     }
+
     #[pyfunction]
-    fn ntohs(x: u16) -> u16 {
+    const fn ntohs(x: u16) -> u16 {
         u16::from_be(x)
     }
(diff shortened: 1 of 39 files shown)


vm/src/builtins/mappingproxy.rs
```diff
@@ -194,7 +197,7 @@ impl PyMappingProxy {
     #[pymethod]
     fn __ior__(&self, _args: PyObjectRef, vm: &VirtualMachine) -> PyResult {
         Err(vm.new_type_error(format!(
-            "\"'|=' is not supported by {}; use '|' instead\"",
+            r#""'|=' is not supported by {}; use '|' instead""#,
             Self::class(&vm.ctx)
```
ShaharNaveh, line 201: Are the extra surrounding quotes intended here?

youknowone: That looks weird, probably not

stdlib/src/socket.rs
```diff
@@ -726,19 +726,22 @@ mod _socket {
     }
 
     #[pyfunction]
-    fn htonl(x: u32) -> u32 {
+    const fn htonl(x: u32) -> u32 {
```
youknowone, line 729: #[pyfunction] is always wrapped to be a python function. const on these function will do nothing.

... nothing bad to do it though

youknowone (approved):

PR #5859, opened 2025-06-28
Create ContentItem with a macro

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Reduced code duplication by consolidating several internal struct definitions using a macro. No changes to user-facing functionality or behavior.


derive-impl/src/pyclass.rs
```diff
@@ -653,72 +653,29 @@ pub(crate) fn impl_pyexception_impl(attr: PunctuatedNestedMeta, item: Item) -> R
...
-        &self.inner
-    }
-}
-impl ContentItem for MemberItem {
-    type AttrName = AttrName;
-    fn inner(&self) -> &ContentItemInner<AttrName> {
-        &self.inner
-    }
-}
+define_content_item!(struct MethodItem, "#[pymethod] and #[pyclassmethod]");
```
youknowone: Just in case, is this also possible? 
```suggestion
define_content_item! {
    /// #[pymethod] and #[pyclassmethod]
    struct MethodItem
};
```

ShaharNaveh: Haven't thought of doing that. checking

ShaharNaveh: From my tests it is possible,  had to change the macro a bit to capture the meta attrs (if present), also added the ability to optionally add `pub` (and variants) in front of the struct

PR #5866, opened 2025-06-29
Update issubclass and make `mro` representation fit to CPython

youknowone:


youknowone:
beacuse MRO is currently not weakrefs, this prevent RC=0 to release type objects

PR #5879, opened 2025-06-30
Update mimetypes from 3.13.5

ShaharNaveh:


ShaharNaveh:
Closing as this PR has a much larger scope than originally anticipated

youknowone:
This is how we solve problems. just go deep enough until something actually resolved. Then merging back from the last to the first

PR #5884, opened 2025-07-01
Fix `itertools.pairwise`

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of the pairwise iterator for better code clarity and maintainability. No changes to user-facing behavior.

### DIFF
--- a/Lib/test/test_itertools.py
+++ b/Lib/test/test_itertools.py
@@ -1182,8 +1182,6 @@ class TestBasicOps(unittest.TestCase):
         with self.assertRaises(TypeError):
             pairwise(None)                                  # non-iterable argument
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_pairwise_reenter(self):
         def check(reenter_at, expected):
             class I:
--- a/Lib/test/test_itertools.py
+++ b/Lib/test/test_itertools.py
@@ -1183,7 +1183,7 @@ class TestBasicOps(unittest.TestCase):
             pairwise(None)                                  # non-iterable argument
 
     # TODO: RUSTPYTHON
-    @unittest.skip("TODO: RUSTPYTHON, hangs")
+    @unittest.expectedFailure
     def test_pairwise_reenter(self):
         def check(reenter_at, expected):
             class I:
@@ -1234,8 +1234,6 @@ class TestBasicOps(unittest.TestCase):
             ([5], [6]),
         ])
 
-    # TODO: RUSTPYTHON
-    @unittest.skip("TODO: RUSTPYTHON, hangs")
     def test_pairwise_reenter2(self):
         def check(maxcount, expected):
             class I:
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -1959,11 +1959,11 @@ mod decl {
 
     impl IterNext for PyItertoolsPairwise {
         fn next(zelf: &Py<Self>, vm: &VirtualMachine) -> PyResult<PyIterReturn> {
-            let old_guard = {
+            let old_clone = {
                 let guard = zelf.old.read();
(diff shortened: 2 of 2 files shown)


Lib/test/test_itertools.py
```diff
@@ -1183,7 +1183,7 @@ def test_pairwise(self):
             pairwise(None)                                  # non-iterable argument
 
     # TODO: RUSTPYTHON
-    @unittest.skip("TODO: RUSTPYTHON, hangs")
+    @unittest.expectedFailure
     def test_pairwise_reenter(self):
```
ShaharNaveh, line 1185: I've tried to make this test to pass as well, but for some reason when it's reentering the iter `old` gets reset to None, causing the next `old` to advance the iter one more time than needed.
So for the first assert:
```
Expected: ([1], ([1], [3]))
Got: ([1], ([3], [4]))
```

If someone can guide me where to investigate it would be great:)

vm/src/stdlib/itertools.rs
```diff
@@ -1959,18 +1959,25 @@ mod decl {
 
     impl IterNext for PyItertoolsPairwise {
         fn next(zelf: &Py<Self>, vm: &VirtualMachine) -> PyResult<PyIterReturn> {
-            let old = match zelf.old.read().clone() {
+            let old_guard = {
```
youknowone: This is not a clone of old_guard, but clone of old, right?
```suggestion
            let old = {
```
or

```suggestion
            let old_cloned = {
```

youknowone (changes requested):

Lib/test/test_itertools.py
youknowone: when new returned Err, the value of old need to be stored in `zelf.old` to prevent not to be lost

vm/src/stdlib/itertools.rs
ShaharNaveh: yup, that's a more fitting name

vm/src/stdlib/itertools.rs
```diff
@@ -1959,18 +1959,29 @@ mod decl {
...
+                        obj
+                    }
                     PyIterReturn::StopIteration(v) => return Ok(PyIterReturn::StopIteration(v)),
                 },
                 Some(obj) => obj,
             };
+
             let new = match zelf.iterator.next(vm)? {
                 PyIterReturn::Return(obj) => obj,
                 PyIterReturn::StopIteration(v) => return Ok(PyIterReturn::StopIteration(v)),
```
youknowone, line 1980: old seems like to be stored in this case.

vm/src/stdlib/itertools.rs
```diff
@@ -1959,18 +1959,29 @@ mod decl {
...
+            let old_clone = {
+                let guard = zelf.old.read();
+                guard.clone()
+            };
+            let old = match old_clone {
                 None => match zelf.iterator.next(vm)? {
-                    PyIterReturn::Return(obj) => obj,
+                    PyIterReturn::Return(obj) => {
+                        // Needed for when we reenter
+                        *zelf.old.write() = Some(obj.clone());
```
youknowone, line 1970: instead of immediately after getting the value

ShaharNaveh: IIRC I tried to do it here first but then the tests failed, will recheck

youknowone: oh, you are right. I was misunderstanding the spec of pairwise. Thank you!

youknowone (approved):

vm/src/stdlib/itertools.rs
ShaharNaveh: I tried to save it in the `new` block, then the tests won't pass:/

PR #5885, opened 2025-07-02
Add `raise_if_stop!` macro for `protocol/iter.rs`

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Improved iterator handling for smoother operation and consistent error messaging.
  * Simplified constructor methods and error message formatting for iterator objects.

No changes to user-facing features or functionality.


vm/src/stdlib/itertools.rs
```diff
@@ -1,5 +1,14 @@
 pub(crate) use decl::make_module;
 
+macro_rules! handle_pyiter_return {
```
youknowone, line 3: This macro is expected to be defined on `vm/src/protocol/iter.rs` because it is mainly a function of PyIterReturn.

The name can be simplified as what it means in Python. 
```suggestion
macro_rules! raise_stop {
```

I like this idea 👍

youknowone: This macro is expected to be defined on `vm/src/protocol/iter.rs` because it is mainly a function of PyIterReturn.

The name can be simplified as what it means in Python. 
```suggestion
macro_rules! raise_stop {
```

I like this idea 👍

vm/src/stdlib/itertools.rs
```diff
@@ -569,10 +568,7 @@ mod decl {
             }
 
             // might be StopIteration or anything else, which is propagated upwards
-            let obj = match zelf.iterable.next(vm)? {
-                PyIterReturn::Return(obj) => obj,
-                PyIterReturn::StopIteration(v) => return Ok(PyIterReturn::StopIteration(v)),
-            };
+            let obj = raise_stop!(zelf.iterable.next(vm)?);
```
youknowone: uh, my fault. now returning a value from `raise_stop!`  looks awkward.

```suggestion
            let obj = raise_if_stop!(zelf.iterable.next(vm)?);
```
I will add `if` in the middle

ShaharNaveh: No worries it's just a simple: `:%s/raise_stop/raise_if_stop/g` :wink:

PR #5887, opened 2025-07-02
macro for pyast code depulication

ShaharNaveh:
The entire 2000+ loc file can be written using a single macro (invoked multiple time). 
This PR is currently set to draft to get initial feedback on the macro itself, it's parameters, what should be configurable, etc...

ATM I've changed only couple of structs to use the macro, there are total of 123 structs in this file, I'm going to do it with a script once the macro looks good enough. lmk what you think

### DIFF
--- a/vm/src/stdlib/ast/pyast.rs
+++ b/vm/src/stdlib/ast/pyast.rs
@@ -86,7 +86,7 @@ impl NodeMod {}
 impl_node!(
     #[pyclass(module = "_ast", name = "Module", base = "NodeMod")]
     pub(crate) struct NodeModModule,
-    fields: ["body", "type_ignores"]
+    fields: ["body", "type_ignores"],
 );
 
 impl_node!(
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/ast/pyast.rs
```diff
@@ -2,138 +2,93 @@
 
 use super::*;
 use crate::common::ascii;
+
+macro_rules! impl_node {
+    (
+        $(#[$meta:meta])*
+        $vis:vis struct $name:ident,
+        [$($field:expr),* $(,)?],
+        [$($attr:expr),* $(,)?]
```
youknowone: Because both of them are listed in `[]`

```suggestion
        fields: [$($field:expr),* $(,)?],
        attributes: [$($attr:expr),* $(,)?]
```

attributes looks like empty for many cases. If you don't like to fill `attributes: []` everywhere, adding another match without attributes line will work.

youknowone (commented):
Thanks! I like the idea very much

PR #5894, opened 2025-07-03
Use `const fn` where possible

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Many internal methods and functions across the codebase have been updated to be evaluable at compile time, improving consistency and enabling more usage in constant contexts.
  * These changes do not affect runtime behavior or user-facing features. No new features or bug fixes are introduced in this update.

### DIFF
--- a/vm/src/stdlib/posix.rs
+++ b/vm/src/stdlib/posix.rs
@@ -1228,7 +1228,7 @@ pub mod module {
     }
 
     #[pyfunction]
-    const fn sync() {
+    fn sync() {
         #[cfg(not(any(target_os = "redox", target_os = "android")))]
         unsafe {
             libc::sync();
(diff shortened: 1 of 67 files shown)


vm/src/stdlib/posix.rs
```diff
@@ -1228,7 +1228,7 @@ pub mod module {
     }
 
     #[pyfunction]
-    fn sync() {
+    const fn sync() {
```
youknowone: #[pyfunction], #[pymethod], #[pygetset] etc will have no change to be called in const context. you don't need to spend too much effort about them. 

`const fn` to rust-context functions is a good idea

ShaharNaveh: > #[pyfunction], #[pymethod], #[pygetset] etc will have no change to be called in const context. you don't need to spend too much effort about them.
> 

It's a simple clippy command:)

PR #5903, opened 2025-07-04
Make gemini ignore `Lib/**`

ShaharNaveh:
https://developers.google.com/gemini-code-assist/docs/customize-gemini-behavior-github#custom-configuration
Hopefully this works

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -1,0 +1,2 @@
+ignore_patterns:
+  - "Lib/**"


ShaharNaveh:
@coderabbitai can you please add your nitpick comment to this branch?

.gemini/config.yaml
```diff
@@ -0,0 +1,2 @@
+ignore_patterns:
+  - "Lib/**"
```
ShaharNaveh, line 2: @coderabbitai Have a code suggestion here to add `pylib/**` as well

.gemini/config.yaml
```diff
@@ -0,0 +1,3 @@
...
+  - "pylib/**"
```
youknowone: we don't add this path to git. so the original version is better

```suggestion
```

youknowone (approved):
Thank you, let's try

PR #5906, opened 2025-07-06
Handle negative time.sleep values

### LINKED ISSUE #5895: time.sleep(-1.0) causes panic
## Summary
https://github.com/RustPython/RustPython/pull/5892/files#r2183174081
```
import time
time.sleep(-1)
```

## Expected
raise exception

## Actual
panic

ever0de:
resolved #5895

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation to prevent negative durations when converting from Python objects, ensuring that negative values now raise clear errors.

* **New Features**
  * Enhanced error messaging for invalid sleep durations, providing a more user-friendly message when a negative duration is specified.

* **Refactor**
  * Updated the `sleep` function to accept a wider range of Python objects for specifying sleep duration, increasing flexibility in usage.
  * Unified and simplified platform-specific sleep handling for better consistency across environments.

### DIFF
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -90,8 +90,25 @@ mod decl {
 
     #[cfg(not(unix))]
     #[pyfunction]
-    fn sleep(dur: Duration) {
+    fn sleep(seconds: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
+        let dur = seconds.try_into_value::<Duration>(vm).map_err(|e| {
+            if e.class().is(vm.ctx.exceptions.value_error) {
+                // Check if this is a "negative duration" error by examining the args
+                if let Some(args) = e.args().first() {
+                    if let Ok(s) = args.str(vm) {
+                        if s.as_str() == "negative duration" {
+                            return vm.new_value_error("sleep length must be non-negative");
+                        }
+                    }
+                }
+                e
+            } else {
+                e
+            }
+        })?;
+
         std::thread::sleep(dur);
+        Ok(())
     }
 
     #[cfg(not(target_os = "wasi"))]
@@ -525,7 +542,7 @@ mod platform {
     use super::decl::{SEC_TO_NS, US_TO_NS};
     #[cfg_attr(target_os = "macos", allow(unused_imports))]
     use crate::{
-        PyObject, PyRef, PyResult, TryFromBorrowedObject, VirtualMachine,
+        AsObject, PyObject, PyObjectRef, PyRef, PyResult, TryFromBorrowedObject, VirtualMachine,
         builtins::{PyNamespace, PyStrRef},
         convert::IntoPyException,
     };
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/time.rs
```diff
@@ -90,8 +90,20 @@ mod decl {
...
+        let dur = seconds.try_into_value::<Duration>(vm).map_err(|e| {
+            if e.class().is(vm.ctx.exceptions.value_error) {
+                if let Some(s) = e.args().first().and_then(|arg| arg.str(vm).ok()) {
+                    if s.as_str() == "negative duration" {
+                        return vm.new_value_error("sleep length must be non-negative");
+                    }
+                }
+            }
+            e
+        })?;
```
youknowone, line 102: Now the two sleep functions share the same error handling.
Originally, the reason they were split was because they didn’t share any logic, so keeping them separate was simpler.
But that’s no longer the case.
Let’s remove #[cfg(not(unix))] from this function and make it the sole function responsible for validating the parameters of sleep.
We should move the cfg checks to the actual implementation instead.

vm/src/stdlib/time.rs
```diff
@@ -691,8 +703,17 @@ mod platform {
     }
 
     #[pyfunction]
-    fn sleep(dur: Duration, vm: &VirtualMachine) -> PyResult<()> {
-        // this is basically std::thread::sleep, but that catches interrupts and we don't want to;
```
youknowone, line 695: This comment is not expected to be removed

vm/src/stdlib/time.rs
ever0de: [Currently](https://github.com/RustPython/RustPython/pull/5906/commits/72d612c3ee4159ac3747af26f88fc3fe83770075), the sleep function is implemented directly in the `decl` module with cfg blocks for platform-specific implementations:

```rust
#[pyfunction]
fn sleep(seconds: PyObjectRef, vm: &VirtualMachine) -> PyResult<()> {
    let dur = seconds.try_into_value::<Duration>(vm).map_err(..)?;

    #[cfg(unix)]
    {
        // Unix-specific nanosleep implementation
    }
    #[cfg(not(unix))]
    {
        // std::thread::sleep for other platforms
    }
}
```

Other functions pattern:
```rust
// decl module
#[pyfunction]
fn monotonic(vm: &VirtualMachine) -> PyResult<f64> {
    Ok(get_monotonic_time(vm)?.as_secs_f64())
}

// platform module  
pub(super) fn get_monotonic_time(vm: &VirtualMachine) -> PyResult<Duration> {
    // Platform-specific implementation
}
```

Should we refactor `sleep` to follow the platform pattern for consistency, or it's not necessary since `sleep` isn't reused and would only involve simple wrapping?

youknowone: either way okay

youknowone (approved):
Thank you for contributing!

PR #5914, opened 2025-07-08
Update compileall from 3.13.5

ShaharNaveh:


ShaharNaveh:
@youknowone It constantly hangs on windows, how can I know which tests causes that?
Also, would you be interested in a PR to set a timeout for the CI, I think it's safe to say that if a job ran for more than an hour it failed

youknowone:
No idea by CI result. Do you have a windows machine? Then it will be easier to check.

The previous trial of this test was #5199

Timeout setup for CI is a good idea. Windows test usually finish in 40mins. Other tests in 30mins.

ShaharNaveh:
> No idea by CI result. Do you have a windows machine? Then it will be easier to check.

I don't:/
I tried to search for a flag that will make the unittests to display which test it will run (before actually running it) like pytest does it, but I couldn't find anything:/


> The previous trial of this test was #5199

ty:)
I looked at what that PR tried to do, I copied some parts of it hopefully it will pass now :crossed_fingers: 

> Timeout setup for CI is a good idea. Windows test usually finish in 40mins. Other tests in 30mins.

I've opened an issue on it, I think it's a good first issue

PR #5918, opened 2025-07-09
fix(sqlite): autocommit mode transaction handling to match CPython

ever0de:
### CPython Implementation
- **Connection.c**
  - https://github.com/python/cpython/blob/6ea425828540d7a19296183c3410283897767d9a/Modules/_sqlite/connection.c#L1781-L1806
  - cpython 3.13: https://github.com/python/cpython/blob/c44070b2d51d59f608577b1f6c55c0168d8e416b/Modules/_sqlite/connection.c#L1749-L1771
- **Cursor.c**
  - https://github.com/python/cpython/blob/6ea425828540d7a19296183c3410283897767d9a/Modules/_sqlite/cursor.c#L866-L874

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved transaction management when switching to auto-commit mode by ensuring pending transactions are properly committed.
  * Refined when transactions are started during statement execution, so transactions only begin if an isolation level is explicitly set.

### DIFF
--- a/Lib/test/test_sqlite3/test_transactions.py
+++ b/Lib/test/test_sqlite3/test_transactions.py
@@ -94,8 +94,6 @@ class TransactionTests(unittest.TestCase):
         self.assertEqual(len(res), 1)
         self.assertEqual(res[0][0], 5)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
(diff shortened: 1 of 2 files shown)


youknowone:
@coderabbitai review

youknowone:
Thank you!

PR #5921, opened 2025-07-09
`sys.platform` android & ios

ShaharNaveh:
Starting from 3.13, `sys.platform` now returns 'android' if running on an android device. See notes at https://docs.python.org/3/library/sys.html#sys.platform

Added "ios" support as well

## Summary by CodeRabbit

* **New Features**
  * Improved platform detection to distinguish between Linux, Android, macOS, and iOS systems.
  * Platform string now accurately identifies Android and iOS separately from Linux and macOS.

### DIFF
--- a/vm/src/stdlib/sys.rs
+++ b/vm/src/stdlib/sys.rs
@@ -73,6 +73,8 @@ mod sys {
                 "android"
             } else if #[cfg(target_os = "macos")] {
                 "darwin"
+            } else if #[cfg(target_os = "ios")] {
+                "ios"
             } else if #[cfg(windows)] {
                 "win32"
             } else if #[cfg(target_os = "wasi")] {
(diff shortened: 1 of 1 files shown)


youknowone:
Is there any reference of this change?

vm/src/stdlib/sys.rs
```diff
@@ -67,11 +67,14 @@ mod sys {
     #[pyattr(name = "platform")]
     pub(crate) const PLATFORM: &str = {
         cfg_if::cfg_if! {
-            if #[cfg(any(target_os = "linux", target_os = "android"))] {
-                // Android is linux as well. see https://bugs.python.org/issue32637
```
youknowone, line 71: Ok, unlike this link,  it is finally changed!

youknowone (approved):
Thank you!

PR #5931, opened 2025-07-10
fix(itertools): add re-entrancy guard to tee object

ever0de:
Reference

- https://github.com/python/cpython/blob/f6650f9ad73359051f3e558c2431a109bc016664/Modules/itertoolsmodule.c#L948-L965
- https://github.com/python/cpython/blob/f6650f9ad73359051f3e558c2431a109bc016664/Modules/itertoolsmodule.c#L772-L797

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stability of the tee iterator by preventing concurrent or recursive access, reducing the risk of runtime errors or inconsistent behavior. Users will now receive a clear error if re-entrancy is attempted.

### DIFF
--- a/vm/src/stdlib/itertools.rs
+++ b/vm/src/stdlib/itertools.rs
@@ -1184,33 +1184,28 @@ mod decl {
     #[derive(Debug)]
     struct PyItertoolsTeeData {
         iterable: PyIter,
-        values: PyRwLock<Vec<PyObjectRef>>,
-        locked: AtomicCell<bool>,
+        values: PyMutex<Vec<PyObjectRef>>,
     }
 
     impl PyItertoolsTeeData {
         fn new(iterable: PyIter, _vm: &VirtualMachine) -> PyResult<PyRc<Self>> {
             Ok(PyRc::new(Self {
                 iterable,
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/itertools.rs
```diff
@@ -1185,21 +1185,31 @@ mod decl {
...
+                    return Err(vm.new_runtime_error("cannot re-enter the tee iterator"));
+                }
+
+                let result = self.iterable.next(vm);
+                self.locked.store(false);
+
+                let obj = raise_if_stop!(result?);
+                self.values.write().push(obj);
             }
+
```
youknowone: While traditional locks prefer to lock the code, Rust locks prefer to lock the data.
After changing PyRwLock to PyMutex, this approach will work
```suggestion
            let Some(values) = self.values.try_lock() else {
                return Err(vm.new_runtime_error("cannot re-enter the tee iterator"));
            };
            if values.len() == index {
                let obj = raise_if_stop!(self.iterable.next(vm)?);
                values.push(obj);
            }
            Ok(PyIterReturn::Return(values[index].clone()))
```

youknowone (changes requested):

PR #5935, opened 2025-07-10
Update operator from 3.13.5

ShaharNaveh:
## Summary by CodeRabbit

* **Style**
  * Improved formatting of error messages by adding single quotes around type or object names for better readability.

### DIFF
--- a/Lib/test/test_operator.py
+++ b/Lib/test/test_operator.py
@@ -635,17 +635,17 @@ class PyOperatorTestCase(OperatorTestCase, unittest.TestCase):
 class COperatorTestCase(OperatorTestCase, unittest.TestCase):
     module = c_operator
 
-    #TODO: RUSTPYTHON
+    # TODO: RUSTPYTHON
     @unittest.expectedFailure
     def test_attrgetter_signature(self):
         super().test_attrgetter_signature()
 
-    #TODO: RUSTPYTHON
+    # TODO: RUSTPYTHON
     @unittest.expectedFailure
     def test_itemgetter_signature(self):
         super().test_itemgetter_signature()
 
-    #TODO: RUSTPYTHON
+    # TODO: RUSTPYTHON
     @unittest.expectedFailure
     def test_methodcaller_signature(self):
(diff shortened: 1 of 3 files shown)


Lib/test/test_operator.py
```diff
@@ -555,6 +603,31 @@ def test_dunder_is_original(self):
             if dunder:
                 self.assertIs(dunder, orig)
 
+    @support.requires_docstrings
+    def test_attrgetter_signature(self):
```
ShaharNaveh, line 607: I have tried to mark those signature tests with `unittest.expectedFailure` but then the test fails on "UNEXPECTED SUCCESS", very strange.

I have tried to mess with defining `__text_signature__` but I'm guessing that I'm doing something wrong because it completely ignores it. I would appreciate any suggestion that you have:)

youknowone: Maybe it is a base class of multiple other tests

Lib/test/test_operator.py
```diff
@@ -555,6 +603,43 @@ def test_dunder_is_original(self):
...
+        # TODO: RUSTPYTHON fails only when running inside COperatorTestCase
+        if type(self).__name__ == 'COperatorTestCase':
+            self.skipTest("TODO: RUSTPYTHON")
```
youknowone: This way of handling makes hard to fix it when it is actually fixed.
I know we have a few tests having similar patterns. They must be also fixed.

A recommended pattern is adding another test function under COperatorTestCase.

```
class COperatoorTestCase(...):  # existing code

    # TODO: RUSTPYTHON
    def test_attrgetter_signature(self):
        super().test_attrgetter_signature()
```

youknowone (changes requested):

Lib/test/test_operator.py
ShaharNaveh: > This way of handling makes hard to fix it when it is actually fixed. I know we have a few tests having similar patterns. They must be also fixed.

sure, np

PR #5940, opened 2025-07-10
Don't hang on evil type name

ShaharNaveh:


vm/src/builtins/type.rs
```diff
@@ -833,11 +833,12 @@ impl PyType {
                 value.class().slot_name(),
             ))
         })?;
+
         if name.as_bytes().contains(&0) {
             return Err(vm.new_value_error("type name must not contain null characters"));
         }
 
-        *self.heaptype_ext.as_ref().unwrap().name.write() = name;
```
arihant2math, line 840: Isn't this a bit redundant?

ShaharNaveh: I'm assuming that you meant to comment on the line bellow where I'm creating a new `PyStr` from an existing `PyStr`. I'm still not sure why it fixes it.

I tried to use `clone` but still same thing, this is the only thing that worked 🤷 

If you have better ideas, pleas lmk:)

youknowone: This is really weird

youknowone: I'd better look in this issue more how the original code can be deadlocked.

PR #5954, opened 2025-07-12
`str.replace` support count as keyword arg

ShaharNaveh:
From python 3.13 you can pass `count` as a keyword arg when doing `str.replace`

https://docs.python.org/3.13/library/stdtypes.html#str.replace

### DIFF
--- a/vm/src/builtins/str.rs
+++ b/vm/src/builtins/str.rs
@@ -1017,21 +1017,31 @@ impl PyStr {
         !self.data.is_empty() && self.char_all(char::is_alphabetic)
     }
 
+    /// Return a copy of the string with all occurrences of substring old replaced by new.
+    /// If count is given, only the first count occurrences are replaced.
+    /// If count is not specified or -1, then all occurrences are replaced.
     #[pymethod]
-    fn replace(&self, old: PyStrRef, new: PyStrRef, count: OptionalArg<isize>) -> Wtf8Buf {
+    fn replace(&self, args: ReplaceArgs) -> Wtf8Buf {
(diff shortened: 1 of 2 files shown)


youknowone (approved):
👍

vm/src/builtins/str.rs
```diff
@@ -1017,21 +1017,31 @@ impl PyStr {
         !self.data.is_empty() && self.char_all(char::is_alphabetic)
     }
 
+    /// Return a copy of the string with all occurrences of substring old replaced by new.
+    /// If count is given, only the first count occurrences are replaced.
+    /// If count is not specified or -1, then all occurrences are replaced.
```
youknowone:
```suggestion
```
We have a special document handling for `pymethod`. It automatically copies `__doc__` from Python doc database https://github.com/RustPython/__doc__.
But if we add this comment, this one will have priority

youknowone (changes requested):
👍 Thank you! A minor fix request about our doc system

vm/src/builtins/str.rs
ShaharNaveh: oh, I just looked at https://github.com/RustPython/RustPython/blob/e75aebb967266c02c2b10f9294afe2b8df878efc/vm/src/builtins/str.rs#L1038-L1052
And thought I might as well, so the docs from `str.ispritable` needs to be removed?

youknowone: Yes. being removed is better.
Once we manually added them. But now removing is better when the same function exists in CPython public API

youknowone (approved):

PR #5960, opened 2025-07-13
Fix invalid args count msg

ShaharNaveh:
## Summary by CodeRabbit

* **New Features**
  * Added an arity method to generated argument parsing implementations, allowing users to query the expected argument count range.

* **Style**
  * Updated error messages for argument count mismatches to use a consistent, lowercase format for improved readability.

### DIFF
--- a/derive-impl/src/from_args.rs
+++ b/derive-impl/src/from_args.rs
@@ -180,12 +180,16 @@ fn generate_field((i, field): (usize, &Field)) -> Result<TokenStream> {
 }
 
 pub fn impl_from_args(input: DeriveInput) -> Result<TokenStream> {
-    let fields = match input.data {
-        Data::Struct(syn::DataStruct { fields, .. }) => fields
-            .iter()
-            .enumerate()
-            .map(generate_field)
-            .collect::<Result<TokenStream>>()?,
+    // TODO: Set lower airty bound dynamicly
+    let (fields, arity) = match input.data {
+        Data::Struct(syn::DataStruct { fields, .. }) => (
+            fields
+                .iter()
+                .enumerate()
+                .map(generate_field)
+                .collect::<Result<TokenStream>>()?,
+            fields.len(),
+        ),
         _ => bail_span!(input, "FromArgs input must be a struct"),
     };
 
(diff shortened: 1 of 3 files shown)


vm/src/function/argument.rs
```diff
@@ -213,7 +213,7 @@ impl FuncArgs {
 
         if !self.args.is_empty() {
             Err(vm.new_type_error(format!(
-                "Expected at most {} arguments ({} given)",
+                "expected at most {} arguments, got {}",
```
ShaharNaveh, line 216: The error message needs to have a prefix of the caller name.
For example:

Got:
```
TypeError: expected at most 3 arguments, got 4
```

Expected:
```
TypeError: str expected at most 3 arguments, got 4
```

---

Tried to see how this can be done for quite a while,  I need a hint how to do it😅

ShaharNaveh: @youknowone do you have a suggestion?

youknowone: Not easy. We have to pass more context but you probably found we dont have interface for that.

ShaharNaveh: > you probably found we dont have interface for that.

Correct :/

---

How should this PR continue? I think that it's an improvement overall as now it makes it so the number of argument is correct in the error message. it used to be:

```
at most 0 arguments, got 4
```

Should I restore the test and make this PR ready?

youknowone: Yes please, thank you for deep-diving this issue!

derive-impl/src/from_args.rs
```diff
@@ -180,24 +180,32 @@ fn generate_field((i, field): (usize, &Field)) -> Result<TokenStream> {
...
     };
 
     let name = input.ident;
     let (impl_generics, ty_generics, where_clause) = input.generics.split_for_impl();
     let output = quote! {
         impl #impl_generics ::rustpython_vm::function::FromArgs for #name #ty_generics #where_clause {
+            fn arity() -> ::std::ops::RangeInclusive<usize> {
+                0..=#arity
+            }
+
```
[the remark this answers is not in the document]
ShaharNaveh: That's true but requires some refactoring as we return a Result when we are calling `generate_field` so we lose the attr.kind before we can inspect it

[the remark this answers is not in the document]
ShaharNaveh: Thanks, will do it in a separate PR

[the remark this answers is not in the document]
ShaharNaveh: yes. open an issue about this

PR #5969, opened 2025-07-14
fix(sqlite): validate surrogates in SQL statements

ever0de:


stdlib/src/sqlite.rs
```diff
@@ -2294,6 +2294,7 @@ mod _sqlite {
             sql: &PyStr,
             vm: &VirtualMachine,
         ) -> PyResult<Option<Self>> {
+            let _ = sql.try_to_str(vm)?;
             let sql_cstr = sql.to_cstring(vm)?;
```
youknowone: It means we shouldn't create cstring from wtf8, but have to create one from str.
The underlying logic of this approach is fine. But `let _ = sql.try_to_str(vm)?;` to discard the result of `try_to_str` is weird.
How about adding a new helper method, `fn ensure_utf8(&self) -> PyResult<()>`, which provides basically the same feature but with better naming? I can call `try_to_str` inside.

Another suggestion is a proper way to handle it.

```rust
impl PyRef<PyStr> {
...
    fn into_utf8(self) -> PyRef<PyUtf8Str>
...
}

struct PyUtf8Str(PyStr);
```

This will be useful when we actually need utf8 string instead of wtf8

vm/src/builtins/str.rs
```diff
@@ -80,6 +80,29 @@ impl fmt::Debug for PyStr {
...
+impl PyUtf8Str {
+    /// Returns the underlying string slice. This is safe because the
+    /// type invariant guarantees UTF-8 validity.
+    pub fn as_str(&self) -> &str {
+        debug_assert!(
+            self.0.is_utf8(),
+            "PyUtf8Str invariant violated: inner string is not valid UTF-8"
+        );
+        unsafe { self.0.to_str().unwrap_unchecked() }
+    }
```
youknowone: The safety is trivial for users, but not for devs
```suggestion
    /// Returns the underlying string slice.
    pub fn as_str(&self) -> &str {
        debug_assert!(
            self.0.is_utf8(),
            "PyUtf8Str invariant violated: inner string is not valid UTF-8"
        );
        // Safety: This is safe because the type invariant guarantees UTF-8 validity.
        unsafe { self.0.to_str().unwrap_unchecked() }
    }
```

vm/src/builtins/str.rs
```diff
@@ -80,6 +80,29 @@ impl fmt::Debug for PyStr {
     }
 }
 
+#[repr(transparent)]
+#[derive(Debug)]
+pub struct PyUtf8Str(PyStr);
+
+impl std::ops::Deref for PyUtf8Str {
```
youknowone, line 88:
```suggestion
// TODO: Remove this Deref which may hide missing optimized methods of PyUtf8Str
impl std::ops::Deref for PyUtf8Str {
```

PR #5970, opened 2025-07-14
Impl Drop for PyAtomicRef

youknowone:
@qingshi163 basically everything is PyObjectRef. So this will be fine, right?

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved memory management to ensure that Python object references are safely released when no longer needed, reducing the risk of resource leaks.


youknowone:
#5968 is calling Drop, and it looks fine!

PR #5972, opened 2025-07-14
Provide detailed error for circular `from` imports

ever0de:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages when importing names from a module, providing clearer information in cases of circular imports or partial initialization.

* **New Features**
  * Enhanced detection of modules that are still initializing, helping users identify circular import issues more easily.


ever0de:
ref

- https://github.com/python/cpython/blob/a68ddea3bf7e9bb77d096c613bce2ec1e67a28f4/Lib/importlib/_bootstrap.py#L914-L955

PR #5976, opened 2025-07-15
Implement utf16-le

moreal:
This pull request may implement utf16-le codec but it will be stayed in draft status.

Refer https://github.com/RustPython/RustPython/pull/5976#issuecomment-3072889169.


youknowone:
Since UTF16 is unpopular codecs these days, I'd like to keep it as python implementation to reduce code to maintain.

Here is another encoding work unfinished about UTF32 #5800

moreal:
> Since UTF16 is unpopular codecs these days, I'd like to keep it as python implementation to reduce code to maintain.
> 
> Here is another encoding work unfinished about UTF32 #5800

Ah, I see. Then, this pull request will be stayed in draft status.

PR #5978, opened 2025-07-15
Update `test_exception*.py` from 3.13.5

ShaharNaveh:


Lib/test/test_exceptions.py
```diff
@@ -1765,272 +1885,6 @@ def test_name_error_has_name(self):
         except NameError as exc:
             self.assertEqual("bluch", exc.name)
 
-    def test_name_error_suggestions(self):
```
ShaharNaveh, line 1768: Those tests were moved to `test_traceback`, but when I tried to update `test_traceback` from 3.13.5 nothing ran. Everything was skipped because we don't have `_testcapi`.

How should we proceed?

youknowone: Usually `_testcapi` is used for `@cpython_only` tests.
is PurePythonSuggestionFormattingTests also being skipped?

ShaharNaveh: >  is PurePythonSuggestionFormattingTests also being skipped?

Yes:/

youknowone: Could you also update test_traceback here? I will check if it is a bug of test suite or intended

ShaharNaveh: Now I see that it requires `has_debug_ranges`. which requires the `_testinternalcapi` :/

https://github.com/RustPython/RustPython/blob/a54873d302e2abb5d23d975006e9338d61bbc616/Lib/test/support/__init__.py#L505-L515

youknowone: If those tests actually doesn't require that feature - because they didn't before update - writing a patch for CPython can be a solution

ShaharNaveh: I'll check. my assumption is that only some of them requires it. we'll see

PR #5983, opened 2025-07-15
Correctly implement `isolation_level` setter to handle deletion

ever0de:
ref

- https://github.com/python/cpython/blob/7689407fa4406ab79d7e9e02363f50be4ec35b5e/Modules/_sqlite/connection.c#L1781-L1806

I noticed CPython's error message is just "cannot delete attribute". I made mine more informative (...'isolation_level' attribute...). Should I change it to match CPython for exact compatibility?

### DIFF
--- a/stdlib/src/sqlite.rs
+++ b/stdlib/src/sqlite.rs
@@ -61,6 +61,7 @@ mod _sqlite {
             PyInt, PyIntRef, PySlice, PyStr, PyStrRef, PyTuple, PyTupleRef, PyType, PyTypeRef,
         },
         convert::IntoObject,
+        function::PySetterValue,
         function::{ArgCallable, ArgIterable, FsPath, FuncArgs, OptionalArg, PyComparisonValue},
         object::{Traverse, TraverseFn},
         protocol::{PyBuffer, PyIterReturn, PyMappingMethods, PySequence, PySequenceMethods},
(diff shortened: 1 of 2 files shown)


stdlib/src/sqlite.rs
```diff
@@ -61,6 +61,7 @@ mod _sqlite {
             PyInt, PyIntRef, PySlice, PyStr, PyStrRef, PyTuple, PyTupleRef, PyType, PyTypeRef,
         },
         convert::IntoObject,
+        function::PySetterValue,
         function::{ArgCallable, ArgIterable, FsPath, FuncArgs, OptionalArg, PyComparisonValue},
```
youknowone:
```suggestion
        function::{ArgCallable, ArgIterable, FsPath, FuncArgs, OptionalArg, PyComparisonValue, PySetterValue},
```

youknowone (approved):
👍

stdlib/src/sqlite.rs
```diff
@@ -1350,22 +1353,34 @@ mod _sqlite {
             self.isolation_level.deref().map(|x| x.to_owned())
         }
         #[pygetset(setter)]
-        fn set_isolation_level(&self, val: Option<PyStrRef>, vm: &VirtualMachine) -> PyResult<()> {
-            if let Some(val) = &val {
-                begin_statement_ptr_from_isolation_level(val, vm)?;
-            }
+        fn set_isolation_level(&self, value: PySetterValue, vm: &VirtualMachine) -> PyResult<()> {
```
youknowone:
```suggestion
        fn set_isolation_level(&self, value: PySetterValue<Option<PyStrRef>>, vm: &VirtualMachine) -> PyResult<()> {
```
How about trying this?

youknowone (approved):

youknowone:
Thank you!

PR #5987, opened 2025-07-16
Implement properties on `_TextIOBase` and `StringIO`

ever0de:
ref
- https://github.com/python/cpython/blob/ef053a92d5552bb082bbe1fe7fcaab020cbe6f99/Modules/_io/stringio.c#L959-L965
- https://github.com/python/cpython/blob/ef053a92d5552bb082bbe1fe7fcaab020cbe6f99/Modules/_io/textio.c#L164-L168

## Summary by CodeRabbit

* **New Features**
  * Added `errors` and `line_buffering` properties to text I/O objects, providing default attribute access for improved compatibility.

### DIFF
--- a/vm/src/stdlib/io.rs
+++ b/vm/src/stdlib/io.rs
@@ -743,11 +743,6 @@ mod _io {
         fn errors(_zelf: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
             vm.ctx.none()
         }
-
-        #[pygetset]
-        fn line_buffering(_zelf: PyObjectRef) -> bool {
-            false
-        }
     }
 
     #[derive(FromArgs, Clone)]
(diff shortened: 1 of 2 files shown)


vm/src/stdlib/io.rs
```diff
@@ -738,6 +738,16 @@ mod _io {
         fn encoding(_zelf: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
             vm.ctx.none()
         }
+
+        #[pygetset]
+        fn errors(_zelf: PyObjectRef, vm: &VirtualMachine) -> PyObjectRef {
```
youknowone, line 743: The linked errors is for TextIOBase, so this is looking good

vm/src/stdlib/io.rs
```diff
@@ -738,6 +738,16 @@ mod _io {
...
+            vm.ctx.none()
+        }
+
+        #[pygetset]
+        fn line_buffering(_zelf: PyObjectRef) -> bool {
+            false
+        }
```
youknowone: But this one is different. Linked one is stringIO

youknowone (approved):
👍

PR #5994, opened 2025-07-16
Set `{min,max}_arity`

### LINKED ISSUE #5980: Refactor FromArgs derive to compute accurate arity bounds
## Problem

The current `impl_from_args` function in `derive-impl/src/from_args.rs` computes arity bounds incorrectly. The generated `fn arity()` returns `0..=#arity` (where `#arity` is `fields.len()`), which:

- Always allows zero arguments—even when some fields are required
- Counts keyword-only and defaulted fields toward both bounds
- Does not distinguish between positional and keyword-only parameters

## Root Cause

The current structure makes it challenging to compute accurate arity bounds because `generate_field` encapsulates both attribute parsing and code generation, and we lose access to the `attr.kind` and `attr.default` information after it returns the `TokenStream`.

## Solution

Refactor the code to separate attribute parsing from code generation:

1. **First pass**: Parse attributes from all fields to extract `attr.kind` and `attr.default` 
2. **Second pass**: Generate field code using parsed attributes
3. **Compute bounds**: Calculate accurate arity bounds as:
   - `min_arity` = number of non-default `PositionalOnly` + `PositionalOrKeyword` fields
   - `max_arity` = total number of `PositionalOnly` + `PositionalOrKeyword` fields

## References

- Original PR: https://github.com/RustPython/RustPython/pull/5960
- Discussion: https://github.com/RustPython/RustPython/pull/5960#discussion_r2207717068

## Implementation Notes

The refactoring should split `generate_field` into separate functions for parsing and generation, allowing `impl_from_args` to access the parsed attribute information needed for accurate arity calculation.

ShaharNaveh:
Fixes #5980 

## Summary by CodeRabbit

* **Refactor**
  * Improved error handling and type safety for argument parsing.
  * Enhanced clarity and maintainability of argument extraction logic.
  * Dynamic calculation of required and optional argument counts for better accuracy.

* **Bug Fixes**
  * More robust detection and handling of multiple argument attributes, reducing potential errors in argument parsing.

### DIFF
--- a/derive-impl/src/from_args.rs
+++ b/derive-impl/src/from_args.rs
@@ -6,38 +6,45 @@ use syn::{Attribute, Data, DeriveInput, Expr, Field, Ident, Result, Token, parse
 
 /// The kind of the python parameter, this corresponds to the value of Parameter.kind
 /// (https://docs.python.org/3/library/inspect.html#inspect.Parameter.kind)
+#[derive(Default)]
 enum ParameterKind {
     PositionalOnly,
+    #[default]
     PositionalOrKeyword,
     KeywordOnly,
     Flatten,
 }
 
-impl ParameterKind {
-    fn from_ident(ident: &Ident) -> Option<Self> {
-        match ident.to_string().as_str() {
-            "positional" => Some(Self::PositionalOnly),
-            "any" => Some(Self::PositionalOrKeyword),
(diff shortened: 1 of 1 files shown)


youknowone (approved):
Thank you!

PR #6014, opened 2025-07-21
Construct detailed message on sqlite text decode failure

ever0de:
reference

- https://github.com/python/cpython/blob/ef053a92d5552bb082bbe1fe7fcaab020cbe6f99/Modules/_sqlite/cursor.c#L393-L413


youknowone (approved):

youknowone:
Thanks!

PR #6016, opened 2025-07-21
Replace `compiler::source` module with `ruff_source_file`

ShaharNaveh:
I saw that `compiler/source` was mainly re-exporting items from `ruff_source_file`, so I replaced it with the upstream `ruff_source_file::SourceFile` which is a bit better implementation of what we have currently as it's uses `Arc` internally so it's practically free to clone & replaces the need for a dedicated `SourceCodeOwned` as `SourceFile` is already owned.

This PR contains a lot of noise as I renamed:
- `source_code` -> `source_file`
- `SourceCode` -> `SourceFile`
- `SourceCodeOwned` -> `SourceFile`

---

Feel free to close this PR if this is an unwanted change.

## Summary by CodeRabbit

* **Refactor**
  * Unified the handling of source code representation across the compiler and VM by replacing the previous type with a new, consistent type throughout all relevant modules and interfaces.
  * Updated numerous method and function signatures to accept the new source type, improving maintainability and consistency.
  * Deprecated and removed references to legacy source handling components.
  * Adjusted dependencies to reflect the new approach and removed obsolete packages.

* **Chores**
  * Reformatted configuration files and cleaned up commented-out or unused dependencies.


compiler/source/Cargo.toml
```diff
@@ -1,16 +0,0 @@
-[package]
-name = "rustpython-compiler-source"
-description = "RustPython Source and Index"
-version.workspace = true
```
youknowone, line 4: we have to publish a deprecated version of this crate before removeing it from source code.

```suggestion
version = "0.5.0+deprecated"
```

youknowone (changes requested):
Thanks! This is a lot better. Maybe we'd better to check if those ruff types are exported through public API or not, and add some re-export later though.
Could you check the removed crate to let us to publish a deprecated marked version?

compiler/source/Cargo.toml
ShaharNaveh: ofc, didn't realized that it was published, thought it was completely internal

ShaharNaveh:
> Thanks! This is a lot better. Maybe we'd better to check if those ruff types are exported through public API or not, and add some re-export later though.

Ruff don't release it's crates to https://cates.io so I'm not sure what "public API" you're referring to. Or do you mean our public API?

> Could you check the removed crate to let us to publish a deprecated marked version?

ofc, just did

youknowone:
Public API means RustPython API. It will be confusing if RustPython APIs are directly using ruff's types or functions.

youknowone (approved):
Thanks!

ShaharNaveh:
> Public API means RustPython API. It will be confusing if RustPython APIs are directly using ruff's types or functions.

Completely agree, want me to open a PR reexporting `ruff_source_file:{SourceFile, SourceFileBuilder}`? if so, where should I put it?

youknowone:
probably `compiler/core`

PR #6018, opened 2025-07-22
Cleanup `vm/src/stdlib/stat.rs`

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Enhanced platform compatibility by unifying file mode and attribute constants across Unix, macOS, and Windows.
  * Grouped Windows file attribute constants for streamlined access and consistency.
  * No changes affecting user-facing features or behavior.

### DIFF
--- a/vm/src/stdlib/stat.rs
+++ b/vm/src/stdlib/stat.rs
@@ -10,6 +10,7 @@ mod stat {
     #[cfg(not(any(unix, windows)))]
     type Mode = u32; // Fallback for unknown targets
 
+    // unix_libc_get
     cfg_if::cfg_if! {
         if #[cfg(unix)] {
             macro_rules! unix_libc_get {
@@ -19,6 +20,24 @@ mod stat {
             }
         } else {
             macro_rules! unix_libc_get {
+                ($name:ident, $val:expr) => {
+                    $val
+                };
+            }
+        }
+
+    }
+
+    // macos_libc_get
+    cfg_if::cfg_if! {
+        if #[cfg(target_os = "macos")] {
+            macro_rules! macos_libc_get {
+                ($name:ident, $val:expr) => {
+                    libc::$name
+                };
+            }
+        } else {
+            macro_rules! macos_libc_get {
                 ($name:ident, $val:expr) => {
                     $val
                 };
@@ -316,6 +335,7 @@ mod stat {
     }
 
     // Windows file attributes (if on Windows)
+
     #[cfg(windows)]
     #[pyattr]
     pub use windows_sys::Win32::Storage::FileSystem::{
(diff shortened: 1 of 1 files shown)


vm/src/stdlib/stat.rs
```diff
@@ -13,208 +10,229 @@ mod stat {
...
+        }
+    }
 
-    #[cfg(unix)]
-    #[pyattr]
-    pub const S_IFREG: Mode = libc::S_IFREG;
-    #[cfg(not(unix))]
     #[pyattr]
-    pub const S_IFREG: Mode = 0o100000;
+    pub const S_IFDIR: Mode = unix_libc_get!(S_IFDIR, 0o040000);
```
youknowone: This is a good idea! Looks a lot better than before.

I have a suggestion for the macro interface. How about this?
```suggestion
    pub const S_IFDIR: Mode = libc_const!(#[cfg(unix)]  S_IFDIR, 0o040000);
```

vm/src/stdlib/stat.rs
```diff
@@ -13,208 +10,229 @@ mod stat {
...
+        FILE_ATTRIBUTE_NOT_CONTENT_INDEXED, FILE_ATTRIBUTE_OFFLINE, FILE_ATTRIBUTE_READONLY,
+        FILE_ATTRIBUTE_REPARSE_POINT, FILE_ATTRIBUTE_SPARSE_FILE, FILE_ATTRIBUTE_SYSTEM,
+        FILE_ATTRIBUTE_TEMPORARY, FILE_ATTRIBUTE_VIRTUAL,
+    };
+
+    // Unix file flags (if on Unix)
+
     #[pyattr]
-    pub const S_IEXEC: Mode = 0o0100;
+    pub const UF_NODUMP: u32 = macos_libc_get!(UF_NODUMP, 0x00000001);
```
youknowone: Then this one also will have the same pattern
```suggestion
    pub const UF_NODUMP: u32 = libc_const!(#[cfg(target_os = "macos")] UF_NODUMP, 0x00000001);
```

vm/src/stdlib/stat.rs
ShaharNaveh: I'm on the fence about this one, as it is clear. but will require to put it for every const definition, I tried to make it less verbose overall as there's no reason to do it imo.

That's why I defined two macros that are the same except the condition inside the `cfg_if::cfg_if` macro.

---

I tried to do this cleanup using couple of ways such as:

```rust
#[cfg(unix)]
mod maybe_unix {
  pub use libc::{A, B, C};
}

#[cfg(not(unix))]
mod maybe_unix {
  pub const A:Mode = 1;
  ...
}

mod stat {
  #[pyattr]
  pub use maybe_unix::{A, B, C};
}

```

But I decided to drop it as it would require to sync between 3 places (2 modules, and the `stat` module).

---

I also tried to have the cfg condition to be passed as an optional argument to the macro, but then I encountered some weird issues with `#[pyattr]`. 

---

At the end this is the best compromise that I found, unless you have a better idea

ShaharNaveh: I have tried to implement the macro based on the suggested interface. I'm unsure how this would work. can you help me out here how do you envision the macro to look like?

youknowone: Does current version make sense?

I also tried `libc_const!(@unix S_IFDIR, 0o040000);` and `libc_const!(@macos UF_NODUMP, 0x00000001);`, to make a balance of concise forms, but not confident about readability.

ShaharNaveh (commented):
LGTM:)

vm/src/stdlib/stat.rs
ShaharNaveh: The current version looks way better than mine, clearly you have more experience :laughing:

youknowone (approved):

PR #6019, opened 2025-07-22
Update warnings

ShaharNaveh:


Lib/test/test_warnings/__init__.py
```diff
@@ -0,0 +1,1978 @@
...
+             b"  File \"<string>\", line 1, in <module>",
+             b'    import sys, warnings; sys.stdout.write(str(sys.warnoptions)); warnings.w'
+             b"arn('Message', DeprecationWarning)",
+             b'                                                                  ~~~~~~~~~~'
+             b'~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^',
+             b"DeprecationWarning: Message"])
+
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
+    def test_default_filter_configuration(self):
```
ShaharNaveh, line 1442: @youknowone This test fails here, in the parent test class. but it passes when it runs with the python test class
```PyEnvironmentVariableTests```

How should I mark the parent as failing but the child as passing?

ShaharNaveh:
Can't believe I missread that and tried to mark the root class instead of the child only. Thanks!

youknowone (approved):
That sometimes happens. 😂  Thanks!

PR #6030, opened 2025-07-24
[DRAFT] Add `kde` function and tests to RustPython statistics module

jackoconnordev:
**DRAFT**: I am still not fully confident in my local development environment and want to verify any differences between the tests run remotely and on my local.

# Description

Adds the `kde` function to RustPython's standard library alongside Python unit tests. Function and tests are copied as is from CPython 3.13.

## whats_left.py

Following is the result of diffing the output of `python -I whats_left.py` with and without the kde function added.

```sh
❯ diff with-kde.txt without-kde.txt
1464a1465
> statistics.kde
1465a1467
> statistics.pi
```


Lib/statistics.py
```diff
@@ -136,6 +136,7 @@
 from itertools import groupby, repeat
 from bisect import bisect_left, bisect_right
 from math import hypot, sqrt, fabs, exp, erf, tau, log, fsum
+from math import pi, cos, sin, cosh, atan
```
jackoconnordev: Corresponding line in CPython 3.13: https://github.com/python/cpython/blob/74503acba7d6c563aeef307ccf2d0cee141838b0/Lib/statistics.py#L142

Lib/statistics.py
```diff
@@ -601,6 +602,218 @@ def multimode(data):
     return list(map(itemgetter(0), mode_items))
 
 
+def kde(data, h, kernel='normal', *, cumulative=False):
```
jackoconnordev, line 808: Corresponding line in CPython 3.13: https://github.com/python/cpython/blob/74503acba7d6c563aeef307ccf2d0cee141838b0/Lib/statistics.py#L808-L1017

jackoconnordev:
I don't know if the tests will pass with this. I still haven't fully figured out a fully working local development setup. I can build a rustpython binary which works as excpected, but `cargo test --no-default-features --features stdlib,importlib,stdio,encodings,sqlite,ssl` which I grabbed from the CI yaml file gives me the following output.

```
RUSTPYTHONPATH or PYTHONPATH is set, but it wasn't loaded to `Settings::path_list`. If you are going to customize the RustPython vm/interpreter, those environment variables are not loaded in the Settings struct by default. Please try creating a customized instance of the Settings struct. If you are developing the RustPython interpreter, it might be a bug during development.
If you don't have access to a consistent external environment (e.g. targeting wasm, embedding rustpython in another application), try enabling the `freeze-stdlib` feature.
If this is intended and you want to exclude the encodings module from your interpreter, please remove the `encodings` feature from `rustpython-vm` crate.
```

Any pointers would be appreciated if anyone spots this and is familiar with the message :saluting_face:.

ShaharNaveh:
Hey @jackoconnordev ty for the PR:)

Does:

```bash
cargo run --release -- -m test -j1 -v -u cpu test_statistics -m "TestKDE"
```

works for you? (assuming you're checked to your branch)

---

Also, it would be better if you could update both `Lib/statistics.py` & `Lib/test/test_statistics.py` from CPython 3.13, as we are trying to refrain from doing partial updates of code from CPython.

If any tests fails you can mark them with:

```python
# TODO: RUSTPYTHON
@unittest.expectedFailure
def test_X(self):
  ...
```

jackoconnordev:
Happy to and thanks for the quick review! I tried that test command now and got two AttributeErrors:

- AttributeError: module 'statistics' has no attribute '_kernel_invcdfs'
- AttributeError: module 'statistics' has no attribute 'kde_random'

I'll put this back into Draft and try porting the entire files from CPython :muscle:.

Lib/statistics.py
jackoconnordev: Do not copy Python lib files partially. Instead copy the whole file in one go.

Lib/statistics.py
jackoconnordev: Do not copy Python lib files partially. Instead copy the whole file in one go.

youknowone (approved):

youknowone (dismissed):
Thank you for contributing!

@ShaharNaveh Thank you so much for the review!

youknowone (commented):
test_inv_cdf requires investigation. Adding back `@unittest.expectedFailure` will be enough for `test_slots`

jackoconnordev:
[Relevant CPython PR](https://github.com/python/cpython/pull/95265) which introduced the failing assertion to `test_inv_cdf`. I will try to find the equivalent for the CPython C function `_normal_dist_inv_cdf` in RustPython and try make a similar change.

jackoconnordev:
I've run the full test_statistics suite and it's passing on my laptop now `cargo run --release -- -m test -v -u cpu test_statistics`.

The pytest tests in `extra_snippets` are passing too. I added a hardcoded path to /Lib in `impl Default for Settings` and the cargo tests are passing locally too. Is that the normal solution for running `cargo test`?

youknowone:
I usually don't run `cargo test`, but I expect it works without any local patch.
What was the problem? Fixing the problem will be better

youknowone (approved):
Thank you!

PR #6036, opened 2025-07-25
stdlib(sqlite3): Raise ProgrammingError for missing named parameter

ever0de:
ref 

- https://github.com/python/cpython/blob/f66c75f11d3aeeb614600251fd5d3fe1a34b5ff1/Modules/_sqlite/cursor.c#L741-L748


youknowone (approved):
👍

PR #6037, opened 2025-07-25
Update `Lib/test/support/__init__.py` from 3.13.5

ShaharNaveh:
## Summary by CodeRabbit

* **New Features**
  * The time display function now works consistently across all supported platforms, including WASI.

### DIFF
--- a/vm/src/stdlib/time.rs
+++ b/vm/src/stdlib/time.rs
@@ -121,16 +121,9 @@ mod decl {
         Ok(())
     }
 
-    #[cfg(not(target_os = "wasi"))]
     #[pyfunction]
     fn time_ns(vm: &VirtualMachine) -> PyResult<u64> {
         Ok(duration_since_system_now(vm)?.as_nanos() as u64)
-    }
-
-    #[cfg(target_os = "wasi")]
-    #[pyfunction]
-    fn time_ns(_vm: &VirtualMachine) -> PyResult<u64> {
-        unimplemented!("time.time_ns is not implemented yet for wasi");
     }
 
     #[pyfunction]


Lib/test/support/__init__.py
```diff
@@ -6,8 +6,8 @@
 import contextlib
 import dataclasses
 import functools
-import getpass
-import opcode
+import logging
```
ShaharNaveh, line 9: The CI fails on wasi because `logging` uses `time.time_ns` which is not available on wasi (yet).
How should I approach this? should I patch this here so we won't crash if this raises an `AttributeError`, or should this PR be blocked until we implement `time.time_ns` for wasi?

youknowone: That's up to the situation. In this case, we already have `_time` for wasi, so it must be easy to add time_ns in wasi. (it will be inaccurate though) So I will add time_ns in this case.
If it was not easy, I would try to patch logging a little bit to ignore time_ns or to use time

vm/src/stdlib/time.rs
```diff
@@ -121,7 +121,6 @@ mod decl {
         Ok(())
     }
 
-    #[cfg(not(target_os = "wasi"))]
```
ShaharNaveh, line 124: This is what I ended up doing. This line was inserted 4 years ago. maybe things changed since then? now it passes for wasi 🤷 .

Usure if this is a correct fix for the issue

ShaharNaveh:
@coderabbitai generate unit tests

vm/src/stdlib/time.rs
youknowone: looks good

PR #6039, opened 2025-07-26
feat(stdlib/sqlite): Implement slice assignment for Blob

ever0de:
reference

- https://github.com/python/cpython/blob/d658b9053beaacaae80e318f59a5ddd672aa757a/Modules/_sqlite/blob.c#L508-L552

## Summary by CodeRabbit

* **New Features**
  * Added support for assigning to slices of blobs, allowing users to update multiple bytes at once using slice notation.

### DIFF
--- a/stdlib/src/sqlite.rs
+++ b/stdlib/src/sqlite.rs
@@ -2288,7 +2288,7 @@ mod _sqlite {
                     );
                     self.check(ret, vm)
                 } else {
-                    let span_len = (range.end - range.start) as usize;
+                    let span_len = range.end - range.start;
                     let mut temp_buf = vec![0u8; span_len];
 
                     let ret = inner.blob.read(


ever0de:
cpython version: 3.12.5
```python
import sqlite3

def setup_db_and_blob():
    db = sqlite3.connect(":memory:")
    db.execute("CREATE TABLE test (data BLOB)")
    db.execute("INSERT INTO test (data) VALUES (?)", (b"A" * 50,))
    db.commit()
    blob = db.blobopen("test", "data", 1)
    return db, blob

def test_scenario(description, code_to_run):
    db, blob = setup_db_and_blob()
    error_message = ""
    error_type = ""
    try:
        exec(code_to_run)
    except Exception as e:
        error_type = type(e).__name__
        error_message = str(e)
    finally:
        blob.close()
        db.close()
    
    print(f"--- {description} ---")
    print(f"Code: {code_to_run}")
    print(f"Result: {error_type}('{error_message}')\n")

test_scenario(
    "Non-contiguous Buffer",
    'blob[5:10] = memoryview(b"0123456789")[::2]'
)
# BufferError('memoryview: underlying buffer is not C-contiguous')

test_scenario(
    "Step is Zero",
    'blob[5:10:0] = b"12345"'
)
# ValueError('slice step cannot be zero')

test_scenario(
    "Size Mismatch",
    'blob[5:10] = b"a"'
)
# IndexError('Blob slice assignment is wrong size')
```

PR #6045, opened 2025-07-29
Nitpicks

ShaharNaveh:


wtf8/src/lib.rs
```diff
@@ -91,17 +91,17 @@ impl CodePoint {
     ///
     /// `value` must be less than or equal to 0x10FFFF.
     #[inline]
-    pub const unsafe fn from_u32_unchecked(value: u32) -> CodePoint {
-        CodePoint { value }
+    pub const unsafe fn from_u32_unchecked(value: u32) -> Self {
```
youknowone: wtf8 crate is mostly clone of outer project. I'd like to keep the diff as small as possible

PR #6048, opened 2025-07-29
Add `symtable.py` from 3.13.5

ShaharNaveh:
## Summary by CodeRabbit

* **New Features**
  * Added a comprehensive set of symbol and scope constants to the symbol table module, now accessible as Python attributes.

* **Refactor**
  * Simplified the `PySymbolTable` interface by replacing several explicit methods with Python-style properties.
  * Renamed and adjusted properties for easier access to attributes like name, type, line number, children, identifiers, symbols, nested status, and a new id property.

* **Chores**
  * Updated the module registration name from "symtable" to "_symtable".

### DIFF
--- a/Lib/test/support/__init__.py
+++ b/Lib/test/support/__init__.py
@@ -514,7 +514,12 @@ def has_no_debug_ranges():
     return not bool(config['code_debug_ranges'])
 
 def requires_debug_ranges(reason='requires co_positions / debug_ranges'):
-    return unittest.skipIf(has_no_debug_ranges(), reason)
+    try:
+        skip = has_no_debug_ranges()
+    except unittest.SkipTest as e:
+        skip = True
+        reason = e.args[0] if e.args else reason
+    return unittest.skipIf(skip, reason)
 
 @contextlib.contextmanager
 def suppress_immortalization(suppress=True):
--- a/Lib/test/test_compile.py
+++ b/Lib/test/test_compile.py
@@ -138,7 +138,6 @@ def f(x):
     def test_argument_order(self):
         self.assertRaises(SyntaxError, exec, 'def f(a=1, b): pass')
 
-    @unittest.skip("TODO: RUSTPYTHON, thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseFloatError { kind: Invalid }'")
     def test_float_literals(self):
         # testing bad float literals
         self.assertRaises(SyntaxError, eval, "2e")
@@ -201,6 +200,8 @@ if 1:
         self.assertEqual(eval("0o777"), 511)
         self.assertEqual(eval("-0o0000010"), -8)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_int_literals_too_long(self):
         n = 3000
         source = f"a = 1\nb = 2\nc = {'3'*n}\nd = 4"
@@ -274,6 +275,8 @@ if 1:
             self.assertRaises(SyntaxError, compile, stmt, 'tmp', 'single')
             self.assertRaises(SyntaxError, compile, stmt, 'tmp', 'exec')
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
     def test_import(self):
         succeed = [
             'import sys',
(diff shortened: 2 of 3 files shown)


Lib/test/test_symtable.py
```diff
@@ -54,33 +181,70 @@ class SymtableTest(unittest.TestCase):
...
+
     a_method = find_block(Mine, "a_method")
     spam = find_block(top, "spam")
     internal = find_block(spam, "internal")
     other_internal = find_block(spam, "other_internal")
     foo = find_block(top, "foo")
-
+    Alias = find_block(top, "Alias")
+    GenericAlias = find_block(top, "GenericAlias")
+    # TODO(RUSTPYTHON)
```
youknowone:
```suggestion
    # XXX: RUSTPYTHON
```
We don't have a strict rule for this, but conventionally using this to distinguish local patch from failing tests.

Lib/test/test_symtable.py
```diff
@@ -54,33 +181,70 @@ class SymtableTest(unittest.TestCase):
...
+    GenericMine = find_block(top, "GenericMine")
+    # TODO(RUSTPYTHON)
+    # GenericMine_inner = find_block(GenericMine, "GenericMine")
+    # TODO(RUSTPYTHON)
+    # T = find_block(GenericMine, "T")
+    # TODO(RUSTPYTHON)
+    # U = find_block(GenericMine, "U")
+
+
+    # TODO(RUSTPYTHON)
```
youknowone:
```suggestion
    # TODO: RUSTPYTHON
```
The notation itself looks good.
But we use search tools for this comment. Keeping consistency will be helpful

youknowone (changes requested):
Did we have test_symtable but no symtable.py? How did it work before?

Lib/test/test_symtable.py
ShaharNaveh: Old habits die hard :sweat_smile:

ShaharNaveh:
> Did we have test_symtable but no symtable.py?

Correct

> How did it work before?

We had implemented `symtable.py` in rust:)

Lib/test/test_symtable.py
```diff
@@ -75,12 +230,21 @@ def test_id(self):
...
         self.assertGreater(self.spam.get_id(), 0)
         self.assertGreater(self.internal.get_id(), 0)
+        self.assertGreater(self.foo.get_id(), 0)
+        self.assertGreater(self.Alias.get_id(), 0)
+        self.assertGreater(self.GenericAlias.get_id(), 0)
+        self.assertGreater(self.generic_spam.get_id(), 0)
+        self.assertGreater(self.GenericMine.get_id(), 0)
 
+    # TODO: RUSTPYTHON
+    @unittest.expectedFailure
```
youknowone: If we had test-passing `symtable` before, can these tests be easily fixed?

ShaharNaveh: I think so, I think the main issue is that "symbols" attribute used to return a list, now it's a dict. I'll investigate a bit more about the structure. For now I'll set the PR as a draft

youknowone: Thank you so much!

ShaharNaveh: Now more tests pass. But we still have a difference between our impl and cpython. but this is a change that needs to be made at `compiler/codegen/src/symboltable.rs`, I think it's a bit out of scope for this PR. I'll try to do a followup PR to fix it

vm/src/stdlib/symtable.rs
```diff
@@ -48,57 +141,39 @@ mod symtable {
...
-        const fn is_nested(&self) -> bool {
-            self.symtable.is_nested
-        }
-
-        #[pymethod]
-        const fn is_optimized(&self) -> bool {
-            matches!(
-                self.symtable.typ,
-                CompilerScope::Function | CompilerScope::AsyncFunction
-            )
```
youknowone, line 76: let's not delete these methods but change to getset, just like get_name and get_lineno.
Unlike other more complex cases, the deletion of these 2 methods cause unnecessary regression.

youknowone (changes requested):

vm/src/stdlib/symtable.rs
ShaharNaveh: Fix this. And made `type` to return an int instead of a String, so now more tests pass

youknowone (approved):
Thank you!

PR #6052, opened 2025-08-01
Prevent direct instantiation of `sqlite3.{Statement,Blob}`

ever0de:
ref 

- https://github.com/python/cpython/blob/f66c75f11d3aeeb614600251fd5d3fe1a34b5ff1/Modules/_sqlite/statement.c#L191
- `Py_TPFLAGS_DISALLOW_INSTANTIATION`
- https://github.com/python/cpython/blob/f66c75f11d3aeeb614600251fd5d3fe1a34b5ff1/Objects/typeobject.c#L1661-L1665

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented direct creation of `Blob` and `Statement` objects in the sqlite3 module from Python, ensuring a clear error is raised if attempted.
  * Improved error messages by adding quotes around class names when instantiation is disallowed.

### DIFF
--- a/stdlib/src/sqlite.rs
+++ b/stdlib/src/sqlite.rs
@@ -2036,12 +2036,20 @@ mod _sqlite {
     }
 
     #[pyattr]
-    #[pyclass(name, traverse)]
+    #[pyclass(module = "sqlite3", name = "Blob", traverse)]
     #[derive(Debug, PyPayload)]
     struct Blob {
         connection: PyRef<Connection>,
         #[pytraverse(skip)]
         inner: PyMutex<Option<BlobInner>>,
+    }
+
+    impl Constructor for Blob {
+        type Args = FuncArgs;
+
+        fn py_new(_cls: PyTypeRef, _args: Self::Args, vm: &VirtualMachine) -> PyResult {
+            Err(vm.new_type_error("cannot create 'sqlite3.Blob' instances"))
+        }
     }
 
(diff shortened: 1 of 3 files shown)


stdlib/src/sqlite.rs
```diff
@@ -2036,14 +2036,22 @@ mod _sqlite {
...
-    #[pyclass(name, traverse)]
+    #[pyclass(module = "sqlite3", name = "Blob", traverse)]
     #[derive(Debug, PyPayload)]
     struct Blob {
         connection: PyRef<Connection>,
         #[pytraverse(skip)]
         inner: PyMutex<Option<BlobInner>>,
     }
 
+    impl Constructor for Blob {
```
youknowone: Could you check if Unconstructible works?

```suggestion
    impl Unconstructible for Blob {
```

ever0de: We can proceed with this fix by adding single quotes to the error message. I will apply the patch for now.

https://github.com/ever0de/RustPython/blob/b2d6594bd9cc7a91b85cfe107d0e4ce7916bae62/vm/src/types/slot.rs#L801-L808

```diff
- Err(vm.new_type_error(format!("cannot create {} instances", cls.slot_name())))
+ Err(vm.new_type_error(format!("cannot create '{}' instances", cls.slot_name())))
```

youknowone (approved):
👍

PR #6053, opened 2025-08-01
Update some deps

ShaharNaveh:


stdlib/Cargo.toml
```diff
@@ -80,7 +80,7 @@ adler32 = "1.2.0"
 crc32fast = "1.3.2"
 flate2 = { version = "1.1", default-features = false, features = ["zlib-rs"] }
 libz-sys = { package = "libz-rs-sys", version = "0.5" }
-bzip2 = { version = "0.5", features = ["libbz2-rs-sys"] }
+bzip2 = "0.6"
```
ShaharNaveh, line 83: If I understood it correctly. From 0.6 `libbz2-rs-sys` is enabled by default (but they removed the feature name)

youknowone: Thank you for checking

youknowone (approved):

PR #6055, opened 2025-08-01
Update `bz2.py` from 3.13.5

ShaharNaveh:


Lib/lzma.py
ShaharNaveh, line 1: Updated this from 3.13.5 as well

youknowone (approved):
👍

PR #6061, opened 2025-08-02
Embedded module with an id example

markusmoenig:
I had a bit of a tough time getting this to work API wise so I thought maybe a small example would be helpful.

It creates a module, sets an id attribute to the module, registers the module and calls a global fn which reads out the module and id.

Maybe it is useful.

### DIFF
--- a/examples/embedded_id.rs
+++ b/examples/embedded_id.rs
@@ -1,0 +1,50 @@
+use rustpython::vm::*;
+use rustpython_vm as vm;
+use rustpython_vm::builtins::PyModule;
+
+// A global fn called from Python which extracts the id from the module and prints it.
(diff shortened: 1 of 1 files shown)


examples/embedded_id.rs
youknowone, line 1: How about `examples/module_injection.rs`?

The key technique of this example is creating a module and putting it into sys.module.
Please also add what `__id_module` is and how it can be useful to the top of the file.

youknowone (changes requested):
Thank you so much! The example looks great.

Please check CI failure and the comment about file name/header.

examples/embedded_id.rs
markusmoenig: I think I addressed all issues ?

PR #6071, opened 2025-08-05
Add return annotation to `__annotations__` last

### LINKED ISSUE #6068: Function `__annotations__`  item order does not match CPython
# Summary

RustPython function type annotations order does not match Python when accessed via `__annotations__`.

This causes problems with e.g. functools `singledispatch` which assumes the types in annotations are in order of leftmost parameter to return type. `dataclasses` and `pydantic` are other libraries which could be affected by this.

(It this is fixed a lot of importlib tests can be updated to python 3.13 as they have started using singledispatch in Lib/importlib/resources/_common.py)

## Sample code

`str` parameter and `int` return type.

```python
def func(s: str) -> int:
    return int(s)

print(func.__annotations__)
```

## Expected

```python
{'s': <class 'str'>, 'return': <class 'int'>}
```

## Actual

```python
{'return': <class 'int'>, 's': <class 'str'>}
```

jackoconnordev:
# Description

Functions like `functools.singledispatch` are sensitive to the order of items in the `__annotations__` map. CPython puts return types last.

Resolves #6068 

_Note: I'd prefer to include automated tests from CPython alongside this, but struggled. There are tests from Python 3.13.5's `test_importlib/` which will fail without commit, as they make use of `singledispatch` that would serve to test this code change on every release._

## Test script

```python
def func(s: str) -> int:
    return int(s)


print(func.__annotations__)
```

## RustPython before

```python
{'return': <class 'int'>, 's': <class 'str'>, }
```

## CPython

```python
{'s': <class 'str'>, 'return': <class 'int'>}
```

## RustPython after

```python
{'s': <class 'str'>, 'return': <class 'int'>}
```

## Summary by CodeRabbit

* **New Features**
  * Added a Python snippet demonstrating function type annotations and their retrieval, highlighting the importance of annotation order for certain use cases.  
* **Refactor**
  * Adjusted the processing order of return and parameter annotations during compilation to improve internal handling without affecting functionality.

### DIFF
--- a/extra_tests/snippets/syntax_annotations.py
+++ b/extra_tests/snippets/syntax_annotations.py
@@ -1,0 +1,10 @@
+from typing import get_type_hints
+
+def func(s: str) -> int:
+    return int(s)
+
+hints = get_type_hints(func)
+
+# The order of type hints matters for certain functions
+# e.g. functools.singledispatch
+assert list(hints.items()) == [('s', str), ('return', int)]


youknowone:
Thank you so much! To prevent regression, could you add a test about this? Adding the code described in the issue to `snippets/extra_tests/syntax_annotations.py` will be one of the ways.

jackoconnordev:
Can do 🫡

extra_tests/snippets/syntax_annotations.py
```diff
@@ -0,0 +1,10 @@
+from typing import get_type_hints
+
+def func(s: str) -> int:
+    return int(s)
+
+hints = get_type_hints(func)
+
+# The order of type hints matters for certain functions
+# e.g. functools.singledispatch
+assert hints == {'s': str, 'return': int}
```
youknowone: This test doesn't check the order. It have to be serialized to check the order.

```suggestion
assert list(hints.items()) == [('s', str), ('return': int)]
```

jackoconnordev: You're right, my bad 🤦 . Thanks for the suggestion, I've fixed it now.

youknowone (approved):
Thank you!

PR #6081, opened 2025-08-08
 Fix match mapping pattern

youknowone:
## Summary by CodeRabbit

* **New Features**
  * Added comprehensive support for dictionary (mapping) patterns in structural pattern matching, including detailed validation and error handling for key/value extraction.

* **Bug Fixes**
  * Improved stack handling and matching logic for mapping and sequence patterns to ensure correct behavior and alignment with expected semantics.

* **Tests**
  * Introduced new test cases covering dictionary pattern matching, including multiple keys, wildcards, and comprehensive assertions.

* **Documentation**
  * Added and clarified comments explaining stack operations and pattern matching logic for better code understanding.

### DIFF
--- a/compiler/codegen/src/compile.rs
+++ b/compiler/codegen/src/compile.rs
@@ -26,9 +26,10 @@ use ruff_python_ast::{
     ExprFString, ExprList, ExprName, ExprSlice, ExprStarred, ExprSubscript, ExprTuple, ExprUnaryOp,
     FString, FStringElement, FStringElements, FStringFlags, FStringPart, Identifier, Int, Keyword,
     MatchCase, ModExpression, ModModule, Operator, Parameters, Pattern, PatternMatchAs,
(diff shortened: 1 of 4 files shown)


ShaharNaveh:
Would this allow us to do something like:
```python
match x:
  case str(s):
    print(s)
```
?

or this is more complicated?

youknowone:
That case look like a class case. I mostly edited mapping part. So probably not.

arihant2math (approved):
LGTM

arihant2math:
> Would this allow us to do something like:
> 
> ```python
> match x:
>   case str(s):
>     print(s)
> ```
> 
> ?
> 
> or this is more complicated?

that is handled by `compile_pattern_mapping`, it's slightly broken, as I left some sort of bug in the VM implementation for the `MatchClass` bytecode (probably).

PR #6083, opened 2025-08-08
Update `netrc.py` from 3.13.6 and make `pwd` accesible on Android

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Adjusted platform availability for the password/user-info feature: explicitly excluded iOS, WASI, and Redox from the module, and refined Android handling by gating the getpwall API to non-Android targets to improve platform compatibility and build correctness.

### DIFF
--- a/vm/src/stdlib/pwd.rs
+++ b/vm/src/stdlib/pwd.rs
@@ -2,6 +2,7 @@
 
 pub(crate) use pwd::make_module;
 
+#[allow(unused)]
 #[pymodule]
 mod pwd {
     use crate::{
@@ -26,6 +27,7 @@ mod pwd {
         pw_dir: String,
         pw_shell: String,
     }
+
     #[pyclass(with(PyStructSequence))]
     impl Passwd {}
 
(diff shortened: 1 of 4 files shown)


vm/src/stdlib/pwd.rs
```diff
@@ -2,6 +2,7 @@
 
 pub(crate) use pwd::make_module;
 
+#[allow(unused)]
```
youknowone: Because the conditions in `stdlib/mod.rs` must be aligned, this `allow` will not be used.

youknowone (changes requested):

vm/src/stdlib/pwd.rs
ShaharNaveh: it's only being used on android.
I can do what coderabbitai is suggesting if that's ok

vm/src/stdlib/pwd.rs
```diff
@@ -2,6 +2,7 @@
 
 pub(crate) use pwd::make_module;
 
+#[cfg_attr(target_os = "android", allow(unused_imports))]
```
youknowone:
```suggestion
```

PR #6087, opened 2025-08-10
Update `base64.py` from 3.13.6

ShaharNaveh:


Lib/base64.py
youknowone, line 1: Is this patch only changing the permissions?

ShaharNaveh: yes :shrug:

youknowone: ah, ok. base64.py has a main function and it is a runnable script.

Issue #5974, opened 2025-07-15
[RFC] The way to update Python standard library


moreal:
Preface
======

Recently, while attempting to automate the process of updating Python standard libraries in RustPython, I became exhausted from various challenges and stopped working on it. However, I suddenly wanted to preserve these ideas, so I'm writing this post.
The problems and ideas I considered might not actually be very helpful, so it's fine if they're just treated as "Issue A" and moved on.
Since I wrote this in my native Korean before having it translated, there may be some mistranslations, but I ask for your understanding.

Problems
=======

Regarding Python standard library updates, I see two main issues. One is that updating them requires significant labor, and the other is that it's difficult to easily determine which Python version each file is based on.

From a labor perspective
------------

I'm not sure how others work, but in my case, I would copy files from the CPython repository to the same path in the RustPython repository, then use diff commands to try preserving existing comments.
Manually adding decorators like `@unittest.expectedFailure` was quite labor-intensive.

As a solution, I considered automated scripts running on GitHub Actions. I thought about implementing it with the following structure, but couldn't proceed due to difficulties in setting up different OS hosts and personal exhaustion:

1. Copy desired modules or files from the CPython repository.
1. Run the following process in Windows, macOS, and Linux environments to collect reports:
    1. Use `cargo run -- -m test --list-cases <test>` to get all potentially executable test cases.
    1. Run each test case in a separate process with `cargo run -- -m unittest <test>`, as some cases might cause panics or deadlocks.
    1. Collect test success/failure/panic status and create a JSON report to upload to a separate repository (e.g., S3).
1. Use LibCST to parse test files and build an inheritance graph between test classes.
1. Organize failing or panicking test cases from reports and create an execution plan, considering whether to add decorators to parent classes or create arbitrary methods in child classes that call parent methods.
1. Apply changes using LibCST according to the execution plan and submit a PR.
1. Have humans review and finally merge it.

This approach somewhat ignores:
- Existing comments are disregarded. For example, failure reasons appended to `TODO: RUSTPYTHON` comments aren't preserved. They could be preserved, but I couldn't establish criteria for which comments to keep.

---

While agents like Claude Code can handle most cases well, making such automation somewhat less meaningful, I still believe automating what can be automated is good.
If this idea seems reasonable and there are edge cases I haven't considered, please add comments with additional opinions.

About standard library version management
--------------------------

This idea actually came up while writing the automation script.
[...]

Issue #5529, opened 2025-02-20
Update Python libraries and tests from Python 3.13


youknowone:
Though it will be sometimes blocked by rust side issues, don't worry!
Please just create PR about the broken test and and create issues related to the new bugs.

If you are new to updating cpython libraries, this guide will be useful:
https://github.com/RustPython/RustPython/wiki/How-to-update-test-files

When updating module Lib/X, please don't forget to also update Lib/test/test_X

A few libraries doesn't need to be updated because CPython also kept it same as previous version. In that case, please leave a comment to let us mark it as done. Thank you!

Tips
- To get specific version of CPython source code, e.g. try `git checkout v3.13.1` from CPython repository.
- Please check #4064 for example. Not the entire changes but the each commits. One of the middle size example is https://github.com/RustPython/RustPython/commit/17e12dea1ec11b871a9e5d3a4a26e2f9ad83fde2 . Sometimes it can be really simple like https://github.com/RustPython/RustPython/pull/4064/commits/9571a68c01284191d4b18229694c303dad8992ff
- Please create one PR for one library+test pair whenever possible. Once you stuck by dependencies, go to one of the dependency first and finish it before restarting the stuck one unless they become simpler to be updated at the same time. Normally adding more libraries make passing tests (a lot) harder.
- To run tests, try `cargo run --features ssl -- -m test -v <test_name>`. e.g. If you changed `test_unicode.py`, `cargo run --features ssl -- -m test -v test_unicode`
  - If it runs too slow, try to add `--release` like `cargo run --release --features ssl -- -m test -v test_unicode`

# List of libraries
- [ ] `__future__.py`
  - Related test is  `test_future_stmt`
- [x] _colorize  #5520 #5988
- [ ] `_compression.py` 
- [ ] `_markupbase.py` 
- [x] `_osx_support.py` #5596
- [ ] `_threading_local.py`
- [x] `_weakrefset.py` #6004
- [x] `abc.py` #5865
  - `_collections_abc.py`
  - `_py_abc.py`
- [ ] `aifc.py` removed in 3.13
- [x] `argparse.py` #5874
- [x] `ast.py` #6006
- [x] `asynchat.py`  removed in 3.13 #5688
- [ ] `asyncio`
- [x] `asyncore.py`  removed in 3.13 #5688
- [x] `base64.py` #6087 #5872
- [ ] `bdb.py`
- [x] `bisect.py` not changed
- [x] `bz2.py` #5605 #6055
[...]

PR #6089, opened 2025-08-11
Add tool for easier test updates

ShaharNaveh:
~~The idea is to have a file that defines which tests needs to be marked with what (`expectedFailure`, `skipIf`, etc..), and have a code that automatically applies those patches on a given test file.~~

~~This is a POC only, I have plans to make it more modular using argparse and some flags, but the general gist will remain the same.~~

This tool helps with updating test files from CPython.

You can run it like this:

```fish
./scripts/lib_updater.py --from Lib/test/foo.py --to ~/cpython/Lib/test/foo.py -o Lib/test/foo.py
```

to move the patches found in `Lib/test/foo.py` to ` ~/cpython/Lib/test/foo.py` then write the contents back to `Lib/test/foo.py` (you can omit the `-o` arg to print the new contents to stdout)

You can also run it with:

```fish
./scripts/lib_updater.py --from Lib/test/foo.py --show-patches -o my_patches.json
```

To get a baseline of patches. You can alter the patches file with your favorite tool/script/etc and then reapply it with:
```fish
./scripts/lib_updater.py -p my_patches. json --to Lib/test/foo.py
```

cc: @youknowone @arihant2math @moreal  @reactive-firewall would love to hear your feedback

ref: #5974 

---

This PR also updates `test_os.py` from 3.13.7 btw

## Summary by CodeRabbit

* **New Features**
  * Adds a standalone CLI to analyze unittest tests and generate/apply structured patch blocks.
  * Can derive patches from source or load from JSON; includes a preview mode to show patches without applying.
  * Preserves indentation and insertion points, supports in-place updates or printing results, and emits wrapper entries for inherited tests.
  * Handles platform-specific conditions and warns about missing or unmatched test references.


Lib/test/test_os.py
```diff
@@ -835,9 +829,7 @@ def ns_to_sec(ns):
         # Convert a number of nanosecond (int) to a number of seconds (float).
         # Round towards infinity by adding 0.5 nanosecond to avoid rounding
         # issue, os.utime() rounds towards minus infinity.
-        # XXX: RUSTPYTHON os.utime() use `[Duration::from_secs_f64](https://doc.rust-lang.org/std/time/struct.Duration.html#method.try_from_secs_f64)`
```
ShaharNaveh: I had this crazy idea to store patch files in the repo for patches like this. similar to what Termux is doing for their packages, for example:
https://github.com/termux/termux-packages/blob/9865d958666e04f16201a0774a646f37b6082c80/packages/python/0001-fix-hardcoded-paths.patch

---

That's probably out of scope for this PR tho

tools/test_marker/confs/test_os.toml
```diff
@@ -0,0 +1,103 @@
+path = "Lib/test/test_os.py"
```
ShaharNaveh: Also had an idea of having something like:

```suggestion
path = "Lib/test/test_os.py"
cpython_version = "3.13.6"
```

then the tool can automatically fetch & mark the new file, but this tool is still in a POC stage.

I want to know if this tool is something that is even wanted

tools/test_marker/confs/test_os.toml
```diff
@@ -0,0 +1,103 @@
...
+test_access_denied = { expectedFailureIfWindows = { reason = "os.stat (PermissionError: [Errno 5] Access is denied" } }
+test_stat_block_device = { expectedFailureIfWindows = { reason = "os.stat (PermissionError: [Errno 1] Incorrect function" } }
+
+[UtimeTests]
+test_utime = { expectedFailureIfWindows = { reason = "AssertionError: 2.002003 != 1.002003 within 1e-06 delta (1.0000000000000002 difference)" } }
+test_utime_directory = { expectedFailureIfWindows = { reason = "(AssertionError: 2.002003 != 1.002003 within 1e-06 delta (1.0000000000000002 difference)" } }
+test_utime_current = { expectedFailureIfWindows = { reason = "AssertionError: 3359485824.516508 != 1679742912.516503 within 0.05 delta (1679742912.000005 difference) : st_time=3359485824.516508, current=1679742912.516503, dt=1679742912.000005" } }
+test_utime_current_old = { expectedFailureIfWindows = { reason = "AssertionError: 3359485824.5186944 != 1679742912.5186892 within 0.05 delta (1679742912.0000052 difference) : st_time=3359485824.5186944, current=1679742912.5186892, dt=1679742912.0000052" } }
+test_utime_nonexistent = { expectedFailure = { reason = "" } }
```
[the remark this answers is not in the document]
ShaharNaveh: out of scope for this PR

tools/test_marker/confs/test_os.toml
reactive-firewall: ## Regarding any interest

@ShaharNaveh,

Some interest yes, in my [recent experiment with cross-python test compatibility checking](https://github.com/reactive-firewall/CICD-Patch-RustPython-5974/actions/runs/16895703013), I used a value for the cpython version, and I feel there is value to testing the current versions of python as per [Referance release cycle](https://devguide.python.org/versions/) (e.g., versions 3.10-3.14 atm.)

Further, my experiment is well positioned to lend to validating the status of this config, if that is even wanted :shrug:

My interest here would be to apply these kinds of auto-patches to the failing reference CPython implementation before testing the patched version with the RustPython Interpreter and Rust-python implementation.

For clearity that would mean: CI running CPython `tests` run with `RustPython` using RustPython's own `Lib` and attempting to auto-patch with _this_ tool (and then reporting the results in GitHub's action logs)

This would lead to a feedback-loop of

* Developer makes changes (e.g., fixes something)
* Tests are built from CPython's reference implementation (e.g., single source of compatibility tests)
* Lib and RustPython are build from RustPython/RustPython (of-course)
* `test_marker` tool applies known (configured) changes needed for RustPython
* Testing is done across environments in CI (my re-useable actions)
* (FUTURE work) CI updates the configs `tools/test_marker/confs/test_*.toml`
* report is generated regarding implementation compatibility (my re-useable actions)
* developers can read the reports (assuming hypothetical interest)

## lessons/gotchas from auto-patching

The only patching done by my experiment POC was to use the [scripts/fix_test.py](https://github.com/reactive-firewall/CICD-Patch-RustPython-5974/blob/2bb9cbec411ce849c25f3623937414644310a0ae/.github/actions/CI-5974-Test-RustPython-Integration/action.yaml#L149) where as a tool to mark tests is FAR more inline with what I was looking for. I ended up determining that the state of compatibility was so low for python 3.13 that [the report generated by my experiment](https://github.com/reactive-firewall/CICD-Patch-RustPython-5974/actions/runs/16585856322) would need to be stored somewhere, and I was unsure of how to proceed without [reaching out to you](https://github.com/reactive-firewall/CICD-Patch-RustPython-5974/actions/runs/16585856322/workflow#L38), but you reached-out first. :tada:

the knobs I implemented were:
* rust-python `Lib` path
* c-python `Lib` path
* runner `os` (version, and arch) from matrix
* c-python version
* rust-python version (by git `ref` with a default of `main`)
* sub-shell's (e.g. each test-suite process) time-out threshold (a default of 30 seconds seemed reasonable and kept the total test time down to a few hours)
* the rust-python's [cargo arguments](https://github.com/reactive-firewall/CICD-Patch-RustPython-5974/blob/2bb9cbec411ce849c25f3623937414644310a0ae/.github/workflows/Check_Tests.yml#L26)
* include/exclude patterns (mostly didn't need this one once I added a time-out threshold feature)

## Hints for AI

@coderabbitai, FYI: here are the proof-of-concept GHA components from my experiment:
* [Main workflow template](https://raw.githubusercontent.com/reactive-firewall/CICD-Patch-RustPython-5974/2bb9cbec411ce849c25f3623937414644310a0ae/.github/workflows/Check_Tests.yml)
* [Fetch RustPython action](https://raw.githubusercontent.com/reactive-firewall/CICD-Patch-RustPython-5974/2bb9cbec411ce849c25f3623937414644310a0ae/.github/actions/CI-5974-Fetch-RustPython/action.yaml)
* [Fetch CPython action](https://raw.githubusercontent.com/reactive-firewall/CICD-Patch-RustPython-5974/2bb9cbec411ce849c25f3623937414644310a0ae/.github/actions/CI-5974-Fetch-CPython/action.yaml)
* [Auto Integrate reference tests action](https://raw.githubusercontent.com/reactive-firewall/CICD-Patch-RustPython-5974/2bb9cbec411ce849c25f3623937414644310a0ae/.github/actions/CI-5974-Integrate-CPython/action.yaml)
* [POC Testing Experiment Action](https://raw.githubusercontent.com/reactive-firewall/CICD-Patch-RustPython-5974/2bb9cbec411ce849c25f3623937414644310a0ae/.github/actions/CI-5974-Test-RustPython-Integration/action.yaml)

tools/test_marker/main.py
```diff
@@ -0,0 +1,208 @@
...
+import tomllib
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+COL_OFFSET = 4
+INDENT1 = " " * COL_OFFSET
+INDENT2 = " " * COL_OFFSET * 2
+COMMENT = "TODO: RUSTPYTHON"
```
reactive-firewall: :thinking: This would be nice to have an easy override for (CI may be-able to add more details dynamically that way)

tools/test_marker/main.py
```diff
@@ -0,0 +1,208 @@
...
+import pathlib
+import tomllib
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+COL_OFFSET = 4
+INDENT1 = " " * COL_OFFSET
+INDENT2 = " " * COL_OFFSET * 2
```
reactive-firewall: Consider the easy maintainability win:

```suggestion
INDENT2 = INDENT1 * 2
```

@coderabbitai, please explain this suggestion for posterity

tools/test_marker/main.py
```diff
@@ -0,0 +1,208 @@
...
+    >>> patch = {"expectedFailure": {"reason": "lorem ipsum"}}
+    >>> format_patch(patch)
+    '@unittest.expectedFailure # TODO: RUSTPYTHON; lorem ipsum'
+    """
+    method, conf = next(iter(patch_conf.items()))
+    prefix = f"@unittest.{method}"
+
+    reason = conf.get("reason", "")
+    res = ""
+    match method:
```
reactive-firewall: @coderabbitai, FYI:
Documentation hint: `match` was introduced in python3.10?? (need to verify) This will prevent any support for python3.9 (but I'm fine with that, but we should make note of it)

tools/test_marker/main.py
```diff
@@ -0,0 +1,208 @@
...
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+
+COL_OFFSET = 4
+INDENT1 = " " * COL_OFFSET
+INDENT2 = " " * COL_OFFSET * 2
+COMMENT = "TODO: RUSTPYTHON"
+
+ROOT_DIR = pathlib.Path(__file__).parents[2]
+CONFS = ROOT_DIR / "tools" / "test_marker" / "confs"
```
reactive-firewall: 🧹  NIT: Consider future use of `os.path.sep` (in-place of `/`) to keep this portable for cross-platform use.

tools/test_marker/main.py
```diff
@@ -0,0 +1,208 @@
...
+        Whether or not we got a match.
+    """
+    is_attr = isinstance(node, ast.Attribute)
+    attr_node = node if is_attr else node.func
+
+    if isinstance(attr_node, ast.Name):
+        return False
+
+    if attr_node.value.id != "unittest":
+        return False
```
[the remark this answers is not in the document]
reactive-firewall: @ShaharNaveh The suggestion is a good idea, and probably safer than just the name check above.

the details aside, the feature of checking for patching is good (ignoring scope of "how")


@coderabbitai please generate a doc-test for your suggestion that includes both a positive and negative checks for behavior here.

tools/test_marker/main.py
```diff
@@ -0,0 +1,208 @@
...
+            if not (cls_conf := conf.get(cls_node.name)):
+                continue
+
+            for fn_name, patch_conf in cls_conf.items():
+                patch = format_patch(patch_conf)
+                yield (
+                    cls_node.end_lineno,
+                    f"""
+{INDENT1}{patch}
+{INDENT1}def {fn_name}(self):
```
reactive-firewall: :thinking: I'd be interested to see if this can handle some of the crazier stuff like [`Lib/test/test_abc.py`](https://github.com/python/cpython/blob/main/Lib/test/test_abc.py)

The general approach seems fine to me (edge-cases aside).

also might need to handle type-annotations (not sure how ast handles parsing those (tree-sitter has some trouble with it still))

@coderabbitai what's your take on my line of thought here? did I miss something?

tools/test_marker/main.py
```diff
@@ -0,0 +1,208 @@
...
+
+    Returns
+    -------
+    str
+        Patched raw python code.
+    """
+    lines = contents.splitlines()
+    tree = ast.parse(contents)
+
+    # Going in reverse to not distrupt the lines offset
```
reactive-firewall: :broom: In the future it would be helpful to have `logger.debug()` output with before-after (plus some context lines) of the altered code. But assuming the list mutations are safe... that would be out-of-scope for a POC of-course.

@coderabbitai, please analysis this function (e.g., around lines 172-197) looking for possible improvements that could ensure any worries about list modifications here are mitigated.

tools/test_marker/main.py
reactive-firewall, line 1: @ShaharNaveh, Overall LGTM! With the caveat, that as a POC, it has lots of potential for improvements.

Most of my inline comments are off-the-cuff, I recognize this is a proof-of-concept and I like the concept, so feel-free to dismiss any comments that are off-topic without response.

reactive-firewall (commented):
🤩 I especially like the `ast` based approach to marking tests, and I think this is a good starting point for marking tests that only fail for RustPython.

See the remainder of my comments inline.

Lib/test/test_os.py
reactive-firewall: btw, your idea is not so crazy, `C` is NOT `Rust` after all, it might be needed, but yeah, totally out-of-scope for this PR.

<details><summary>what about an empty patch folder for now? TL;DR</summary>
:thinking: I think patches for stuff like this might be the right idea, git can even generate them for us, especially if it we can make them configureable (knobs like: os/arch pair, patch-id, cpython version, and context lines needed (for preventing regressions in the future, etc.) come to mind) for now we could just have a patch folder (empty, but as a placeholder)
</details>

tools/test_marker/main.py
ShaharNaveh: This uses pathlib.Path which is already cross platform:)

tools/test_marker/main.py
ShaharNaveh: This tool (at it's current state) is aimed to handle 90%+ of the cases, if we still have a handful of files that needs to be manually updated I think it's still a win

tools/test_marker/main.py
ShaharNaveh: ofc, I have plans to make pretty much everything configurable via argparse

tools/test_marker/main.py
reactive-firewall: I concur, this tool (including its current state) it is totally a progress win! This was just nitpicking 🙊 about how to approach the future ~10% edge-cases.

This is not a show stopper.

tools/test_marker/main.py
ShaharNaveh: tbh this should be removed and be discovered dynamically. I was just lazy with implementing this, but should probably invest a bit to do it

tools/test_marker/main.py
ShaharNaveh: Python 3.9 is going to EOL at 2025-10, I don't see a reason to keep compatibility with a version that is only receiving security updates ATM and will be deprecated in 2 months.

Source: https://devguide.python.org/versions/

tools/test_marker/main.py
reactive-firewall: @coderabbitai, I concur with @ShaharNaveh that this should be dynamic; can you please explore the code in the rest of the new tool (e.g., the rest of `tools/test_marker/main.py`) and generate a helper function that takes a snippet of code for the target test functions (e.g., like how the patches are applied with ast and `INDENT1`/`INDENT2` currently) and computes the needed indents for the indentation dynamically? be sure to consider SOLID principles and BCP for testing, and all project conventions when generating code.

ShaharNaveh:
@youknowone can you please review this PR?

tools/rpau/pyproject.toml
```diff
@@ -0,0 +1,17 @@
+[project]
+name = "rpau"
+version = "0.1.0"
+description = "Tool for automaticly update Lib & Tests from CPython"
+readme = "README.md"
+authors = [
+    { name = "RustPython Team", email = "" }
+]
+requires-python = ">=3.12"
```
reactive-firewall: :+1: Explicit requirement here is a good improvement!

:bookmark: Before we forget, this sort of design decision should be documented, probably in the adjacent `./README.md`

tools/rpau/src/rpau/cli.py
```diff
@@ -0,0 +1,134 @@
...
+            return help_msg
+
+    def __init__(self, *, formatter_class=_CustomHelpFormatter, **kwargs):
+        super().__init__(formatter_class=formatter_class, **kwargs)
+
+    def _add_action(self, action):
+        action.default = os.environ.get(
+            f"{self.prog}_{action.dest}".upper(), action.default
+        )
+        return super()._add_action(action)
```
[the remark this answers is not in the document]
reactive-firewall: :broom: NIT - I agree with @coderabbitai on this, if the code is going to implement a custom type-aware arg-parser anyway, might as well be consistent with the types.

tools/rpau/src/rpau/cli.py
reactive-firewall, line 1: :+1: Prototype is off to a good start. :broom: Of course it will probably need cleaned up more in the future for maintainability, but it is much better than before already.

youknowone:
@ShaharNaveh @reactive-firewall Thank you so much.

With this tool, *.toml is the original source and the test files are patched based on *.toml files. right?
Currently the tool itself doesn't look like an enhancement yet, but I agree it will be useful once we can automatically collect the toml files easy.
Though guiding people *NOT* to directly edit python test files may be a challenge, but let's think about it more later.

I have questions about the content of toml files. It currently contains expectedFailure, expectedFailureIfWindows, and skip by condition. Now it makes sense because it is hand-written. But not sure they are always automatically decidable.

A few cases are easy to select
- expectedFailure and expecctedFailureIfWindows can be auto-selectable if we run all the test.
- some skips can be auto-selectable because they simply crash.

The other cases are not.
- Flaky errors are not auto-markable. do we maintain a part of toml files manually? That makes sense to me.
- Syntax errors need to be commented out.

ShaharNaveh:
> With this tool, *.toml is the original source and the test files are patched based on *.toml files. right?

Correct.

> Currently the tool itself doesn't look like an enhancement yet, but I agree it will be useful once we can automatically collect the toml files easy. Though guiding people _NOT_ to directly edit python test files may be a challenge, but let's think about it more later.

By making a comment at the top of each python file to not edit it directly but through the tool, and by adding a CI step that will run with a `--check` flag to ensure no manual edits were made we can enforce it (will need to implement it)

> I have questions about the content of toml files. It currently contains expectedFailure, expectedFailureIfWindows, and skip by condition. Now it makes sense because it is hand-written. But not sure they are always automatically decidable.

The purpose is not to automatically detect it. The purpose is to make the existing patches persistent across version updates. 

> A few cases are easy to select
> 
> * expectedFailure and expecctedFailureIfWindows can be auto-selectable if we run all the test.
> * some skips can be auto-selectable because they simply crash.
> 
> The other cases are not.
> 
> * Flaky errors are not auto-markable. do we maintain a part of toml files manually? That makes sense to me.

> * Syntax errors need to be commented out.

About syntax errors and other patches I've suggested to use "patch" files as stated here: 
https://github.com/RustPython/RustPython/pull/6089#discussion_r2267185548

---

The motivation behind this tool is when you copy a new test file from a newer python version, the test names, and failures are most likely to stay the same. so this tool can easily reapply patches from a predefined configuration.


I will try to automate the creation of toml files for tests from the existing patches that we have, but I'll do it after it's confirmed that this tool is wanted.

@youknowone

youknowone:
Oh, thank you. It is more like a migration tool. That makes sense.

If it is a migration tool, I think there can be a middle ground for easy contribution and easy migration.
Rather than having toml files as a original source, can we have a reverse-procedure of this patch tool? If we have a both tool to retrieving patch from existing files and applying patch to new file, contributors still can easily edit the existing test files, but it will be also easy to applicable to the next upgrade.

ShaharNaveh:
> Oh, thank you. It is more like a migration tool. That makes sense.

Yes! Sorry for the misleading PR title. A migration tool is a better name for sure.

> 
> If it is a migration tool, I think there can be a middle ground for easy contribution and easy migration. Rather than having toml files as a original source, can we have a reverse-procedure of this patch tool? 

ofc! it shouldn't be too hard to implement. But, one of the reasons I chose to have an external conf file is for cases where tests get renamed, moved between classes and even moves between files entirely. It would be very hard to implement the detection of when tests moves, It's easier to move entries in the toml files IMO.

Also, even if we had a completely automated way to mark failing tests, how would you deal with flaky tests (tests that pass sometimes), you'd still have to have a centralized place where you state which tests gets skipped.

> If we have a both tool to retrieving patch from existing files and applying patch to new file, contributors still can easily edit the existing test files, but it will be also easy to applicable to the next upgrade.

I agree. One more reason for choosing an external conf is to keep track of upstream cpython version for modules and tests, it would be easier to see down the line which modules and tests are oldest thus should require more effort to update, where on the other end we can get the low hanging fruits for modules that are needing to be updated for minor version only (i.e `3.13.6 -> 3.13.7`).

Also, we have a problem currently that we cannot update some module and test files due to unsupported syntax or missing rust modules. IMO it's better to not have a completely automated test updates ATM but to ease the update procedure for the majority of modules and tests.

youknowone:
> ofc! it shouldn't be too hard to implement. But, one of the reasons I chose to have an external conf file is for cases where tests get renamed, moved between classes and even moves between files entirely. It would be very hard to implement the detection of when tests moves, It's easier to move entries in the toml files IMO.

Yes, I agree the toml files are useful when we need to edit them. But using it as an intermediate step between prev version to next version looks to have more potential then using it as an or original source. Do you think using toml files as the original source of the patch can be more useful? Please tell me if I am missing something.

> I agree. One more reason for choosing an external conf is to keep track of upstream cpython version for modules and tests, it would be easier to see down the line which modules and tests are oldest thus should require more effort to update, where on the other end we can get the low hanging fruits for modules that are needing to be updated for minor version only 

This is an attractive idea. I didn't think about this point. If we don't use toml as original source of patch, probably adding which version of Python introduced the test to comments could be useful. Not sure this is good enough reason to give up easy contribution for new contributors.

> Also, even if we had a completely automated way to mark failing tests, how would you deal with flaky tests (tests that pass sometimes), you'd still have to have a centralized place where you state which tests gets skipped.

> Also, we have a problem currently that we cannot update some module and test files due to unsupported syntax or missing rust modules. IMO it's better to not have a completely automated test updates ATM but to ease the update procedure for the majority of modules and tests.

I honestly don't have much idea about automated way. If it is not the goal, let's not think about it too much.

ShaharNaveh:
> > ofc! it shouldn't be too hard to implement. But, one of the reasons I chose to have an external conf file is for cases where tests get renamed, moved between classes and even moves between files entirely. It would be very hard to implement the detection of when tests moves, It's easier to move entries in the toml files IMO.
> 
> Yes, I agree the toml files are useful when we need to edit them. But using it as an intermediate step between prev version to next version looks to have more potential then using it as an or original source. Do you think using toml files as the original source of the patch can be more useful? Please tell me if I am missing something.

Yes, you are 100% correct. I think we should have one tool with centralized configuration that manages tests and modules E2E, both for marking tests with decorators and both for patching code where necessary. The original source of the patch will be managed via the tool.

> > I agree. One more reason for choosing an external conf is to keep track of upstream cpython version for modules and tests, it would be easier to see down the line which modules and tests are oldest thus should require more effort to update, where on the other end we can get the low hanging fruits for modules that are needing to be updated for minor version only
> 
> This is an attractive idea. I didn't think about this point. If we don't use toml as original source of patch, probably adding which version of Python introduced the test to comments could be useful.

This will be error prone and very tedious to do, let's say that some tests from `test_x.py` moved to `test_y.py` and only some were renamed/changed. It will be a time consuming task to see which tests were introduced in what version. Not to mention that we still have some test files that were updated months & years ago, you'll need to go through each test and see when was it written, and deduce the appropriate python version it based on the `git blame`, which is a lot of effort with very little gain IMO comparing to updating the whole file and having a single comment which CPython version it's aligned with. 

Even if a module wasn't updated between our current module version and the upstream it's still valuable to know that it's aligned so it can be excluded from a TODO list like this: https://github.com/RustPython/RustPython/issues/5529

> Not sure this is good enough reason to give up easy contribution for new contributors.

A valid point and also one that I care about a lot. 

Everything is a tradeoff, if we choose to go with this tool there's going to be some learning curve of how to use it. On the other hand the current state of how we do things is very tedious and painfully slow. For me at least, the process of updating a module (and its tests) is as follows:

1. Copy the files from CPython
2. Run `git diff`, and re-apply any patches that were present before
3. Watch which tests fails in the CI
4. Patch module/test accordingly 

This tool aims to do steps 1 and 2, which takes the majority of time (and are very boring).

Personally, I think that a new contributor can still provide value with this approach, while also having a standard on how to update modules and tests.

Take for example this PR: https://github.com/RustPython/RustPython/pull/6030
It would have been clearer that it's preferred to update the entire module and not just adding a single function (and we will also know which version was it updated to :wink: )

Do you think that this tool will be a gatekeeper and prevent new contributors from contributing?

> > Also, even if we had a completely automated way to mark failing tests, how would you deal with flaky tests (tests that pass sometimes), you'd still have to have a centralized place where you state which tests gets skipped.
> 
> > Also, we have a problem currently that we cannot update some module and test files due to unsupported syntax or missing rust modules. IMO it's better to not have a completely automated test updates ATM but to ease the update procedure for the majority of modules and tests.
> 
> I honestly don't have much idea about automated way. If it is not the goal, let's not think about it too much.

Let's start with defining what is an "automated way", if we are defining it as:

```
rm -r Lib/
cp -a /path/to/cpython_v3_13_7/Lib/ Lib/
./mark_all_failing_tests.py
git add .
git commit -m 'Updated all tests and modules to 3.13.7'
git push
```

Then I think that it's not practical ATM.

I think that it's more realistic ATM to have an easier migration tool with the ability to select which version each test & module are set to.

youknowone:
Update can be less painful by doing like this: https://github.com/RustPython/RustPython/wiki/How-to-update-test-files
Still worse than this patch though.

> Do you think that this tool will be a gatekeeper and prevent new contributors from contributing?

I wish feature contributors don't care about too much things, if possible.
Yes, there is trade-off. I will try to avoid it if everyone have to learn how to use the new toml files and the new tools to fix a single bug.
On the other hand, asking library updaters to learn the new tool is totally fine because it is too painful without tools. 

I hope the new update workflow to be:
1. Run `./generate_migration Lib/test/test_something.py > test_something.toml`
1. Copy a new library to `Lib/test/test_something.py`
1. Edit `test_something.toml` if required.
1. Run `./apply_migration Lib/test/test_something.py test_something.toml`

Elidable file name is fine. I see how worth this is for the library update tasks.

Step 2-4 must be the same as you intended; otherwise I am misunderstanding something. But not sure what do you think about step 1. In my opinion, trying to skipping step 1 potentially force too many people to learn the tool.

Regardless keeping the toml files in the repository or not, allowing them to be desynced will be better than forcing it to be always synced.

ShaharNaveh:
> Update can be less painful by doing like this: https://github.com/RustPython/RustPython/wiki/How-to-update-test-files Still worse than this patch though.
> 
> > Do you think that this tool will be a gatekeeper and prevent new contributors from contributing?
> 
> I wish feature contributors don't care about too much things, if possible. Yes, there is trade-off. I will try to avoid it if everyone have to learn how to use the new toml files and the new tools to fix a single bug. On the other hand, asking library updaters to learn the new tool is totally fine because it is too painful without tools.
> 
> I hope the new update workflow to be:
> 
> 1. Run `./generate_migration Lib/test/test_something.py > test_something.toml`
> 2. Copy a new library to `Lib/test/test_something.py`
> 3. Edit `test_something.toml` if required.
> 4. Run `./apply_migration Lib/test/test_something.py test_something.toml`
>
> Elidable file name is fine. I see how worth this is for the library update tasks.

I like this idea, will implement it and send you a ping for review.

> Step 2-4 must be the same as you intended; otherwise I am misunderstanding something. 

No, you got everything perfectly:)

> But not sure what do you think about step 1. In my opinion, trying to skipping step 1 potentially force too many people to learn the tool.

Originally the tool wasn't supposed to update the files from CPython, but was later added to demonstrate the potential of it. it might be better to split this functionality into a separate tool where you do something like:

```bash
./update_file.py Lib/test/test_os.py --to 'v3.13.7'
```

And it updates it behind the scene by fetching the new file from CPython and running the migration scripts (I'll do it in a future PR if desired).

> Regardless keeping the toml files in the repository or not, allowing them to be desynced will be better than forcing it to be always synced.

If it's not going to be the source of truth then I'd advise against saving them in the repo, I think that it's better that way so it will not cause confusion. Similar to how code without comments, is better than the same code with misleading comments (IMO at least)

reactive-firewall:
> I honestly don't have much idea about automated way. If it is not the goal, let's not think about it too much.

Please Stand-by, I've been thinking on this, I'll elaborate when I finish my re-review.

tools/rpau/src/rpau/logic.py
```diff
@@ -0,0 +1,210 @@
...
+        path=path, version=version, base_url=base_upstream_url, cache_dir=cache_dir
+    )
+
+    patched_contents = apply_conf(contents, conf)
+    new_contents = f"# upstream_version: {version}\n{patched_contents}"
+
+    output_file = output_dir / path
+    # TODO: Add logic to preserve file permissions if exists
+    output_file.parent.mkdir(parents=True, exist_ok=True)
+    output_file.write_text(f"{new_contents}\n")
```
reactive-firewall: :thinking: To automate this properly we will need to know if this function works or not

Please consider something like:


```suggestion
def run(
    conf: dict,
    path: str,
    version: str,
    output_dir: "pathlib.Path",
    cache_dir: "pathlib.Path | None",
    base_upstream_url: str,
) -> bool:
    run_result: bool = False
    try: 
        contents = get_upstream_contents(
            path=path, version=version, base_url=base_upstream_url, cache_dir=cache_dir
        )

        patched_contents = apply_conf(contents, conf)
        new_contents = f"# upstream_version: {version}\n{patched_contents}"

        output_file = output_dir / path
        # TODO: Add logic to preserve file permissions if exists
        output_file.parent.mkdir(parents=True, exist_ok=True)
        output_file.write_text(f"{new_contents}\n")
        run_result = True  # set success flag
    except BaseException as _cause:
        if logger:
            logger.exception(
                "Error running %s",  # lazy formatting to avoid PYL-W1203
                __name__,
            )
        run_result = False
    return run_result
```

This can similarly be used to improve the outcome test of the main function too.

tools/rpau/src/rpau/logic.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    lines = contents.splitlines()
+    tree = ast.parse(contents)
+
+    # Going in reverse to not distrupt the line offset
+    patches = list(iter_patches(tree, conf))
+    for lineno, patch in sorted(patches, reverse=True):
+        lines.insert(lineno, patch)
+
+    return "\n".join(lines)
```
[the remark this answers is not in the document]
reactive-firewall: :broom: NIT -- I generally agree with @coderabbitai on the use of a deep-copy

tools/rpau/src/rpau/logic.py
```diff
@@ -0,0 +1,210 @@
...
+            for fn_name, patch_conf in cls_conf.items():
+                patch = format_patch(patch_conf)
+                yield (
+                    cls_node.end_lineno,
+                    f"""
+{INDENT1}{patch}
+{INDENT1}def {fn_name}(self):
+{INDENT2}return super().{fn_name}()
+""".rstrip(),
+                )
```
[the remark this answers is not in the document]
reactive-firewall: :broom: I generally agree with @coderabbitai, of-course we'd even want to check the function def's arguments and docstrings once possible, etc.

tools/rpau/src/rpau/logic.py
reactive-firewall, line 1: Regarding the `logic.py` component of `rpau` I have two suggestions:

1. Consider that the `run(...)` function might return a boolean to indicate success (to help detect edge-cases like syntax-errors)

2. of-course re-implement the prototype indentation logic to be based on the ast-parsed function definition being patched dynamicly, as previously discussed already.

Otherwise I generally agree with the direction @coderabbitai is suggesting, albeit nitpicky for this early stage.

tools/rpau/src/rpau/logger.py
reactive-firewall, line 1: :+1: The logging enhancement looks good to me.

tools/rpau/src/rpau/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+
+    conf_dir = args.conf_dir
+    with ProcessPoolExecutor(args.workers) as executor:
+        for conf_file in iter_confs(
+            conf_dir, include=args.include, exclude=args.exclude
+        ):
+            try:
+                conf = tomllib.loads(conf_file.read_text())
+            except tomllib.TOMLDecodeError as err:
+                logger.warn(f"{conf_file}: {err}")
```
reactive-firewall: :broom: NIT - consider logging built-in frame dump feature

consider the debug friendly style:


```suggestion
                logger.warning(
			        "%s: %s",  # lazy formatting to avoid PYL-W1203
			        conf_file,
			        err,  # log as just warning level, instead of exception (error), but still detailed.
			        exc_info=True,
		        )
```

logging functions like `warn`/`debug`/`warning`/etc. take a format string already, but defer it until needed, this ironically makes f-strings into overhead (see [Rationale](https://docs.python.org/3/howto/logging.html#optimization))

(the `exc_info` adds the exception trace details to the message)

tools/rpau/src/rpau/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+            func = functools.partial(
+                run,
+                path=path,
+                conf=conf,
+                cache_dir=cache_dir,
+                version=version,
+                output_dir=args.output_dir,
+                base_upstream_url=args.base_upstream_url,
+            )
+            executor.submit(func)
```
reactive-firewall: to assist automation, it would be helpful to collect the results of the runs (especially when being selective with the `--include` and `--exclude` args where there is little benefit gain from the multiprocessing)

tools/rpau/src/rpau/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+def iter_confs(
+    conf_dir: "pathlib.Path", *, include: "re.Pattern", exclude: "re.Pattern"
+) -> "Iterator[pathlib.Path]":
+    for conf_file in conf_dir.rglob("**/*.toml"):
+        if not conf_file.is_file():
+            continue
+
+        uri = conf_file.as_uri().removeprefix("file://")
+        if not include.match(uri):
+            continue
```
reactive-firewall: :broom: NIT - consider leveraging the logger more in `iter_confs`

while prototyping automation, it would be helpful to have the option for a debug messages for the skipped files.

consider something like:

```suggestion
            if (__DEBUG__):
                logger.debug(
                    "skipping unselected '%s'",  # lazy formatting to avoid PYL-W1203
                    uri,
                )
            continue
```

reactive-firewall (commented):
> [!IMPORTANT]
> Review is at bottom.

## In summation:

:+1: The latest changes look good to me in general (at this stage)

---

## Comments on discussion

@youknowone 

> Also, we have a problem currently that we cannot update some module and test files due to unsupported syntax or missing rust modules. IMO it's better to not have a completely automated test updates ATM but to ease the update procedure for the majority of modules and tests.

I actually mitigated the issue of handling tests where rust-python cannot handle unsupported/missing/etc. type issues during fully automated testing, in my #5974 experiments, by a combination of configuration options (e.g., include/exclude) and time-outs (e.g., handle hangs and crashes as inconclusive tests) and fall-back to LGV (albeit an arbitrary assumption that rustpython main is LGV) the same tricks should apply to the rest of Lib (including some falling back to "missing"). So the comparability tests are not required to be coupled to the `rpau` migrations status.

I think we can get this tool automated in CI with a bit of tinkering (and in a way that it is linked to compatibility testing at the same time) if that is wanted. I'll experiment a little bit either way.

Here's what I think automation would make the dev-cycle look-like:

```mermaid
graph TD
    A[Developer Changes] --> AB[Build CPython Reference Tests]
    AB --> B[Test with RustPython for new compatibility]
    B --> C[Apply test_marker patches (with rpau)]
    C --> D[Re-Test with RustPython for regressions]
    D --> E[Generate Compatibility Report]
    E --> F[Optionally Update TOML configs]
    F --> A
```

CI would coordinate efforts and ensure continuous upstream version tracking for configured files; as for easing into the automation of patching the reference implementation, we can use the configuration (or rather the lack of) to help CI identify what to patch and consider everything else implicitly manual until automated (allowing for an incremental transition path forward).

---

So I generally agree with the discussed points on migration focused design as well.

> This is an attractive idea. I didn't think about this point. If we don't use toml as original source of patch, probably adding which version of Python introduced the test to comments could be useful. Not sure this is good enough reason to give up easy contribution for new contributors.

@youknowone - regarding the source of the patch, why not another compromise? Let's consider configuration options to use patches (CI could detect via git-diff tooling, and then extract the patch via git-format-patch) perhaps stored in a patches/lib/ and then applied,skipped,etc. by config's version options (and possibly a toggle override option)

e.g., this could mean new contributors who want to touch the Lib folder would need to be made aware of the override option, and perhaps a CI-check. The override would basically just be a way to `git auto-stash` until the change/PR is accepted, where CI could eventually be made to automate a conditional re-generation of the patch file (the one in the lib) before any rustPython release. 

The idea is to be able to leverage the VCS (e.g., `git`) for the heavy lifting of custom-patches, while also allowing for new contributors to just hand core-members pull-requests like normal.

does that make sense? Of-course this is probably off-topic for `rpau`'s scope. :shrug:

---

@ShaharNaveh 

> I will try to automate the creation of toml files for tests from the existing patches that we have, but I'll do it after it's confirmed that this tool is wanted.

:star_struck:, the generation of the config files would be helpful.

> Yes, you are 100% correct. I think we should have one tool with centralized configuration that manages tests and modules E2E, both for marking tests with decorators and both for patching code where necessary. The original source of the patch will be managed via the tool.

:1st_place_medal: I concur; I think of the stuff we've been working on to just be components of proving just such a concept.

> Even if a module wasn't updated between our current module version and the upstream it's still valuable to know that it's aligned so it can be excluded from a TODO list like this: https://github.com/RustPython/RustPython/issues/5529

Having experimented with testing across python/OS versions, I agree, the visibility was/is very insightful.

I've already started to attempt to integrate this tool into my experiment from #5974 and my main concern is if we want to automate this to smooth adoption it would be ideal if it was invokable from source (e.g. without installing, but rather from the `.py` source like the test fixing scripts); currently it failed with 'could not find' (due to relative call to module, I'll play around with this a bit more), still something to keep in mind.

The second issue I hit was the lack of failure feedback from the tool for the CLI, so when it fails to write a patched-file or create the cache, it still exits the main function with the default `sys` exit code which indicates success to CI and is only going to complicate and delay automation of migration if we have to check if the tool worked on a given environment or not.

> Originally the tool wasn't supposed to update the files from CPython, but was later added to demonstrate the potential of it. it might be better to split this functionality into a separate tool where you do something like:

I've already implemented a PoC of this as a GitHub action as mentioned in GHI #5974, so I agree that the function to fetch the new file should be separate. If you make this into the"

> 
> ```shell
> ./update_file.py Lib/test/test_os.py --to 'v3.13.7'
> ```


it becomes a drop-in compatible tool with the experiment I did, and trivial to automate in CI.

youknowone:
@ShaharNaveh Great. If you don't mind, let's start first version of this tool as a something like test-decorator-migration-tool.  As both of you and walls said, it looks like to have more potential in future for more tasks. Having a concrete usage and a working scenario will be helpful to expand its ability later.

Also, if you are interested in RustPython maintenance, please poke me on our Discord server. Thanks!

ShaharNaveh:
> @ShaharNaveh Great. If you don't mind, let's start first version of this tool as a something like test-decorator-migration-tool. As both of you and walls said, it looks like to have more potential in future for more tasks. Having a concrete usage and a working scenario will be helpful to expand its ability later.

Just pushed a new version of the tool, now the interface is much simpler and easier to work with. Will post some examples soon on how to use it. I'll write a README for it

> Also, if you are interested in RustPython maintenance, please poke me on our Discord server. Thanks!

ofc, will ping you there:)

scripts/lib_updater.py
```diff
@@ -0,0 +1,297 @@
...
+    reason : str, optional
+        Reason for why the test is patched in this way.
+    """
+
+    ut_method: UtMethod
+    cond: str | None = None
+    reason: str = ""
+
+    def fmt(self) -> str:
+        prefix = prefix = f"@unittest.{self.ut_method}"
```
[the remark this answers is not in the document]
ShaharNaveh: oops

scripts/lib_updater.py
```diff
@@ -0,0 +1,297 @@
...
+        output = json.dumps(patches, indent=4)
+        sys.stdout.write(f"{output}\n")
+        sys.exit(0)
+
+    patched = apply_patches(args.remote_file.read_text(), patches)
+
+    if args.inplace:
+        args.orig_file.write_text(patched)
+    else:
+        sys.stdout.write(patched)
```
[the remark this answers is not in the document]
ShaharNaveh: This is intended

ShaharNaveh:
Example:

```fish
./scripts/lib_updater.py --from 'Lib/test/test_os.py' --show-patches
```

<details>
<summary> Output </summary>

```json
{
    "FileTests": {
        "test_closerange": [
            {
                "ut_method": "skipIf",
                "cond": "sys.platform == 'win32'",
                "reason": ", BrokenPipeError: (32, 'The process cannot access the file because it is being used by another process. (os error 32)')"
            }
        ]
    },
    "StatAttributeTests": {
        "test_file_attributes": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.stat return value doesnt have st_file_attributes attribute"
            }
        ],
        "test_access_denied": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.stat (PermissionError: [Errno 5] Access is denied.)"
            }
        ],
        "test_stat_block_device": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.stat (PermissionError: [Errno 1] Incorrect function.)"
            }
        ]
    },
    "UtimeTests": {
        "test_utime": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(AssertionError: 2.002003 != 1.002003 within 1e-06 delta (1.0000000000000002 difference))"
            }
        ],
        "test_utime_directory": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(AssertionError: 2.002003 != 1.002003 within 1e-06 delta (1.0000000000000002 difference))"
            }
        ],
        "test_utime_current": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(AssertionError: 3359485824.516508 != 1679742912.516503 within 0.05 delta (1679742912.000005 difference) : st_time=3359485824.516508, current=1679742912.516503, dt=1679742912.000005)"
            }
        ],
        "test_utime_current_old": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(AssertionError: 3359485824.5186944 != 1679742912.5186892 within 0.05 delta (1679742912.0000052 difference) : st_time=3359485824.5186944, current=1679742912.5186892, dt=1679742912.0000052)"
            }
        ],
        "test_utime_nonexistent": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ],
        "test_large_time": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(ModuleNotFoundError: No module named '_ctypes')"
            }
        ],
        "test_utime_invalid_arguments": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(AssertionError: NotImplementedError not raised)"
            }
        ]
    },
    "EnvironTests": {
        "test_putenv_unsetenv_error": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(AssertionError: ValueError not raised by putenv)"
            }
        ]
    },
    "BytesWalkTests": {
        "test_compare_to_walk": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": "(TypeError: Can't mix strings and bytes in path components)"
            }
        ],
        "test_dir_fd": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": "(TypeError: Can't mix strings and bytes in path components)"
            }
        ],
        "test_yields_correct_dir_fd": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": "(TypeError: Can't mix strings and bytes in path components)"
            }
        ]
    },
    "MakedirTests": {
        "test_exist_ok_existing_directory": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.umask not implemented yet for all platforms"
            }
        ],
        "test_exist_ok_s_isgid_directory": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.umask not implemented yet for all platforms"
            }
        ]
    },
    "URandomFDTests": {
        "test_urandom_fd_closed": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "on Windows (ModuleNotFoundError: No module named 'os')"
            }
        ],
        "test_urandom_fd_reopened": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(ModuleNotFoundError: No module named 'os'"
            }
        ]
    },
    "ExecTests": {
        "test_execvpe_with_bad_arglist": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.execve not implemented yet for all platforms"
            }
        ],
        "test_execve_invalid_env": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.execve not implemented yet for all platforms"
            }
        ],
        "test_execve_with_empty_path": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.execve not implemented yet for all platforms"
            }
        ]
    },
    "TestInvalidFD": {
        "test_fdopen": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ],
        "test_fpathconf": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ],
        "test_ftruncate": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ],
        "test_lseek": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(OSError: [Errno 18] There are no more files.)"
            }
        ],
        "test_read": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(OSError: [Errno 18] There are no more files.)"
            }
        ],
        "test_write": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(OSError: [Errno 18] There are no more files.)"
            }
        ],
        "test_inheritable": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.get_inheritable not implemented yet for all platforms"
            }
        ],
        "test_fchdir": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ],
        "test_fsync": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ]
    },
    "Win32KillTests": {
        "test_kill_sigterm": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(ModuleNotFoundError: No module named '_ctypes')"
            }
        ],
        "test_kill_int": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(ModuleNotFoundError: No module named '_ctypes')"
            }
        ],
        "test_CTRL_BREAK_EVENT": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ]
    },
    "Win32JunctionTests": {
        "test_create_junction": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(AttributeError: module '_winapi' has no attribute 'CreateJunction')"
            }
        ],
        "test_unlink_removes_junction": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(AttributeError: module '_winapi' has no attribute 'CreateJunction')"
            }
        ]
    },
    "Win32NtTests": {
        "test_stat_unlink_race": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.stat (PermissionError: [Errno 5] Access is denied.)"
            }
        ]
    },
    "PidTests": {
        "test_waitpid_windows": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.spawnv not implemented yet for all platforms"
            }
        ],
        "test_waitstatus_to_exitcode_windows": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(OverflowError: Python int too large to convert to Rust i32)"
            }
        ]
    },
    "SpawnTests": {
        "test_spawnve_bytes": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": "fix spawnv bytes"
            }
        ]
    },
    "OSErrorTests": {
        "test_oserror_filename": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": "(AssertionError: b'@test_22106_tmp\\xe7w\\xf0' is not b'@test_22106_tmp\\xe7w\\xf0' : <built-in function chdir>)"
            }
        ]
    },
    "FDInheritanceTests": {
        "test_get_set_inheritable": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.get_inheritable not implemented yet for all platforms"
            }
        ],
        "test_get_set_inheritable_badf": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.get_inheritable not implemented yet for all platforms"
            }
        ],
        "test_open": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.get_inheritable not implemented yet for all platforms"
            }
        ],
        "test_dup": [
            {
                "ut_method": "skipIf",
                "cond": "sys.platform == 'win32'",
                "reason": "os.dup on windows"
            }
        ],
        "test_dup_standard_stream": [
            {
                "ut_method": "skipIf",
                "cond": "sys.platform == 'win32'",
                "reason": "os.dup on windows"
            }
        ],
        "test_dup_nul": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "os.dup not implemented yet for all platforms"
            }
        ]
    },
    "PathTConverterTests": {
        "test_path_t_converter": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": "(AssertionError: TypeError not raised)"
            }
        ]
    },
    "TestDirEntry": {
        "test_uninstantiable": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": "(AssertionError: TypeError not raised by DirEntry)"
            }
        ],
        "test_unpickable": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": "(pickle.PicklingError: Can't pickle <class '_os.DirEntry'>: it's not found as _os.DirEntry)"
            }
        ]
    },
    "TestScandir": {
        "test_uninstantiable": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": "(AssertionError: TypeError not raised by ScandirIter)"
            }
        ],
        "test_attributes": [
            {
                "ut_method": "skipIf",
                "cond": "sys.platform == 'linux'",
                "reason": ", flaky test"
            }
        ],
        "test_removed_dir": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "entry.is_dir() is False"
            }
        ],
        "test_removed_file": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "entry.is_file() is False"
            }
        ],
        "test_fd": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ],
        "test_empty_path": [
            {
                "ut_method": "expectedFailureIfWindows",
                "cond": null,
                "reason": "(AssertionError: FileNotFoundError not raised by scandir)"
            }
        ],
        "test_resource_warning": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ]
    },
    "TestPEP519": {
        "test_pathlike_subclass_slots": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ],
        "test_fspath_set_to_None": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ]
    },
    "ForkTests": {
        "test_fork_at_finalization": [
            {
                "ut_method": "expectedFailure",
                "cond": null,
                "reason": ""
            }
        ]
    }
}
```

</details>

ShaharNaveh:
@youknowone can you please review this PR? I made the changes as discussed. And updated the top comment to show how to use the tool:)

---

Also, I have pinged you on discord

scripts/lib_updater.py
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env python
```
youknowone, line 1:
```suggestion
#!/usr/bin/env python

# This tool helps with updating test files from CPython.
#
# You can run it like this:
#
# ```sh
# ./scripts/lib_updater.py --from Lib/test/foo.py --to ~/cpython/Lib/test/foo.py -o Lib/test/foo.py
# ```
#
# to move the patches found in `Lib/test/foo.py` to ` ~/cpython/Lib/test/foo.py` then write the contents back to `Lib/test/foo.py` (you can omit the `-o` arg to print the new contents to stdout)
#
# You can also run it with:
# 
# ```sh
# ./scripts/lib_updater.py --from Lib/test/foo.py --show-patches -o my_patches.json
# ```
#
# To get a baseline of patches. You can alter the patches file with your favorite tool/script/etc and then reapply it with:
# ```sh
# ./scripts/lib_updater.py -p my_patches. json --to Lib/test/foo.py
# ```
```

ShaharNaveh: Will add it as a docstring

scripts/lib_updater.py
```diff
@@ -0,0 +1,325 @@
...
+            parent_class = cls_node.name
+            for dec_node in fn_node.decorator_list:
+                if not isinstance(dec_node, (ast.Attribute, ast.Call)):
+                    continue
+
+                attr_node = (
+                    dec_node if isinstance(dec_node, ast.Attribute) else dec_node.func
+                )
+
+                if isinstance(attr_node, ast.Name) or attr_node.value.id != "unittest":
```
youknowone: Is attr_node.value expected to be a `Name`? Then

```suggestion
                if isinstance(attr_node, ast.Name):
                    continue
                if isinstance(attr_node.value, ast.Name) and attr_node.value.id != "unittest":
```


```
$ python ./scripts/lib_updater.py --from Lib/test/test_property.py --to ../cpython/Lib/test/test_property.py  
Attribute(value=Name(id='spam', ctx=Load()), attr='setter', ctx=Load())
Attribute(value=Name(id='spam', ctx=Load()), attr='deleter', ctx=Load())
Attribute(value=Attribute(value=Name(id='BaseClass', ctx=Load()), attr='spam', ctx=Load()), attr='getter', ctx=Load())
  File "/Users/user/Projects/rustpython2/./scripts/lib_updater.py", line 314, in <module>
    patches = build_patch_dict(iter_patches(args.gather_from.read_text()))
  File "/Users/user/Projects/rustpython2/./scripts/lib_updater.py", line 213, in build_patch_dict
    for entry in it:
                 ^^
  File "/Users/user/Projects/rustpython2/./scripts/lib_updater.py", line 208, in iter_patches
    yield from PatchEntry.iter_patch_entires(tree, lines)
  File "/Users/user/Projects/rustpython2/./scripts/lib_updater.py", line 131, in iter_patch_entires
    if isinstance(attr_node, ast.Name) or attr_node.value.id != "unittest":
                                          ^^^^^^^^^^^^^^^^^^
AttributeError: 'Attribute' object has no attribute 'id'
```

scripts/lib_updater.py
```diff
@@ -0,0 +1,325 @@
...
+__doc__ = """
+This tool helps with updating test files from CPython.
+
+Examples
+--------
+To move the patches found in `Lib/test/foo.py` to ` ~/cpython/Lib/test/foo.py` then write the contents back to `Lib/test/foo.py`
+
+>>> ./{fname} --from Lib/test/foo.py --to ~/cpython/Lib/test/foo.py -o Lib/test/foo.py
+
+You can run the same command without `-o` to print the output to stdout:
```
youknowone: How about this change?

```suggestion
You can run the same command without `-o` to override the `--from` path:
```
I found in-place editing is the most useful use case in common scenario.  If a feature to print to stdout is desirable, a conventional fake file name `-` can be used to refer stdout.

youknowone:
Looks great in general. It seems very useful to port lots of test files.

@coolreader18 do you have any comment?

@ShaharNaveh @reactive-firewall How do you think about the next step?

scripts/lib_updater.py
ShaharNaveh: That's a bug :tada: 
Will use `getattr` instead.

And attr_node can be of type Name, but if that's the case then we skip as it's not a unittest patch

scripts/lib_updater.py
ShaharNaveh: Sounds good, will do

ShaharNaveh:
> @ShaharNaveh @reactive-firewall How do you think about the next step?

For me at least is to keep using this tool to update some tests and find edge cases that were not encountered until now

reactive-firewall (approved):
:+1: LGTM!

reactive-firewall:
Prototype of using this new tool in CI (e.g., to check for untouched tests / regressions ) [went smoothly](https://github.com/reactive-firewall/CICD-Patch-RustPython-5974/actions/runs/17448458957). I'd personally say the update tool is ready to merge.

---

> > @ShaharNaveh @reactive-firewall How do you think about the next step?
> 
> For me at least is to keep using this tool to update some tests and find edge cases that were not encountered until now

@shaharNaveh the main edge-case I have run-into is when the user provides a bad path (e.g., real `FileNotFoundError`) the exception is not handled well, e.g.,

https://github.com/RustPython/RustPython/blob/6c615cc727204684a3fdfa7c60dfffd98d6c6379/scripts/lib_updater.py#L318-L318

throws unhandled `FileNotFoundError`

<details><summary>TL;DR - Regarding undiscussed edge-cases</summary>

Also consider taking a look at the edgecases in `Lib/test/test_future_stmt/badsyntax_future*.py` (which won't run with RustPython unittests even with the fix_tests script -- I'm not sure what to do with whole files that fail to parse, myself :shrug:, but they could be patched with some fencing probably)

</details>

> [!TIP]
> Consider searching the output of [this test-run](https://github.com/reactive-firewall/CICD-Patch-RustPython-5974/actions/runs/17448458957) for the term "inconclusive" to highlight all kinds of edge-cases (including tests that hang with RustPython)

More importantly: Great work @ShaharNaveh & thanks for this tool!

Hope this helps :bow:

---

@youknowone

<details><summary>TL;DR - in general</summary>

Next steps might be to mark more "patches" and _perhaps_ a counterpart tool that can bring over trivial changed reference lib modules based on a given reference python version and even possibly apply some kind of patch per function definition like the test tool does, but that is getting ahead for now. My rationale is this would let RustPython target each reference python implementation independently (e.g., main could still focus on latest stable 3.13 series while hopefully any work-in-progress for other reference versions could also have a path towards support, without conflicts) :shrug: but for now it would only need to target one version at a time and that would be future-proof enough for tracking. Of-course 

</details>

**in my corner**
However, I am more interested in the CI/CD infrastructure and, in my spare time, hope to look into shoring up the reporting of automated compliance testing with each reference version (e.g., kinda like the whatsleft cronjob CI, but for reference compliance visibility instead) centered around this tool and will ping you on a separate PR sometime after this tool is merged into main (so that CI reporting workflow can rely on it in RustPython `main`).

---
Ref: GHI #5974

ShaharNaveh:
> @ShaharNaveh the main edge-case I have run-into is when the user provides a bad path (e.g., real `FileNotFoundError`) the exception is not handled well, e.g.,
> 
> https://github.com/RustPython/RustPython/blob/6c615cc727204684a3fdfa7c60dfffd98d6c6379/scripts/lib_updater.py#L318-L318
> 
> throws unhandled `FileNotFoundError`

This is intended. What would you expect to get as a user instead?

scripts/lib_updater.py
```diff
@@ -0,0 +1,333 @@
...
+import json
+import pathlib
+import re
+import sys
+import typing
+
+if typing.TYPE_CHECKING:
+    from collections.abc import Iterator
+
+type Patches = dict[str, dict[str, list["PatchSpec"]]]
```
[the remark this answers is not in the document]
ShaharNaveh: This is fine and can be ignored

scripts/lib_updater.py
```diff
@@ -0,0 +1,333 @@
...
+                    continue
+
+                yield (cls_node, fn_node)
+
+
+def iter_patches(contents: str) -> "Iterator[PatchEntry]":
+    lines = contents.splitlines()
+    tree = ast.parse(contents)
+    yield from PatchEntry.iter_patch_entires(tree, lines)
+
```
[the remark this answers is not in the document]
ShaharNaveh: This is fine and intended. you can ignore it

reactive-firewall:
@ShaharNaveh

> What would you expect to get as a user instead?

TL;DR -- Expected mention about the "--from" argument; As this is a developer tool, the intentional nature also is reasonable.

Consider adding a message:

```python
    else:
        try:
            patches = build_patch_dict(iter_patches(args.gather_from.read_text()))
        except FileNotFoundError as _cause:
            raise FileNotFoundError("--from MUST specify an existing file-path") from _cause
```

or even better, consider just handling the exception by aborting; e.g., please consider:

```python
    else:
        try:
            if args.gather_from:  # simple truthy check
                patches = build_patch_dict(iter_patches(args.gather_from.read_text()))
            else:  # falsy should panic
                raise FileNotFoundError("--from MUST specify an existing file-path") from None
        except FileNotFoundError as _cause:
            if sys.stderr.isatty():  # heuristic - probably run interactively by user
                print(
                    "ERROR - file or directory was not found; --from MUST specify an existing file-path",
                    file=sys.stderr,
                )
                sys.exit(66) # heuristic - common POSIX tool code for File-not-found
            else:
                raise FileNotFoundError("--from MUST specify an existing file-path") from _cause
```

PR #6094, opened 2025-08-14
Fix int respect `sys.set_int_max_str_digits` value

ShaharNaveh:
## Summary by CodeRabbit

* **New Features**
  * Enforces a configurable maximum digit length when converting strings/bytes to integers.
  * Users can adjust the limit via sys.set_int_max_str_digits().
* **Bug Fixes / Error Handling**
  * Clearer, consistent errors for invalid literals, invalid bases, and digit-limit breaches with user-facing messages.
  * Non-string inputs with an explicit base now raise a TypeError instead of a ValueError.

### DIFF
--- a/common/src/int.rs
+++ b/common/src/int.rs
@@ -168,7 +168,7 @@ mod tests {
             (("0_100", 10), BigInt::from(100)),
         ] {
             assert_eq!(
-                bytes_to_int(&buf.as_bytes(), base, DIGIT_LIMIT),
+                bytes_to_int(buf.as_bytes(), base, DIGIT_LIMIT),
                 Ok(expected)
             );
         }
@@ -182,7 +182,7 @@ mod tests {
             (("0_", 2), BytesToIntError::InvalidLiteral { base: 2 }),
         ] {
             assert_eq!(
-                bytes_to_int(&buf.as_bytes(), base, DIGIT_LIMIT),
+                bytes_to_int(buf.as_bytes(), base, DIGIT_LIMIT),
                 Err(expected)
             )
         }
@@ -192,7 +192,7 @@ mod tests {
     fn bytes_to_int_invalid_base() {
         for base in [1, 37] {
             assert_eq!(
-                bytes_to_int(&*b"012345", base, DIGIT_LIMIT),
+                bytes_to_int("012345".as_bytes(), base, DIGIT_LIMIT),
                 Err(BytesToIntError::InvalidBase)
             )
         }
@@ -201,7 +201,7 @@ mod tests {
     #[test]
     fn bytes_to_int_digit_limit() {
         assert_eq!(
-            bytes_to_int(&*b"012345", 10, 5),
+            bytes_to_int("012345".as_bytes(), 10, 5),
             Err(BytesToIntError::DigitLimit { got: 6, limit: 5 })
         );
     }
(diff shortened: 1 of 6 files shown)


vm/src/protocol/number.rs
```diff
@@ -76,7 +76,16 @@ impl PyObject {
...
+            let val = s.as_wtf8().trim().as_bytes();
+            let max_len = vm.state.int_max_str_digits.load();
+            if max_len > 0 {
+                let sign_len = usize::from(matches!(val.first(), Some(b'+' | b'-')));
+                let val_len = val.len();
+                if val_len > max_len + sign_len {
+                    return Err(vm.new_value_error(format!("Exceeds the limit ({max_len} digits) for integer string conversion: value has {val_len} digits; use sys.set_int_max_str_digits() to increase the limit")));
+                }
+            }
+            try_convert(self, val, vm)
```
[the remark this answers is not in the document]
ShaharNaveh: Good catch!

vm/src/protocol/number.rs
```diff
@@ -559,3 +556,25 @@ impl PyNumber<'_> {
         })
     }
 }
+
+pub fn handle_bytes_to_int_err(
```
youknowone, line 560: Maybe implementing `IntoPyException` or `ToPyException` for `BytesToIntError` could be possible.
Then `BytesToIntError::BytesToIntError` will need to include `&PyObject` inside the enum value.

Could you check if this is a possible idea or not?

common/src/int.rs
```diff
@@ -29,113 +29,119 @@ pub fn float_to_ratio(value: f64) -> Option<(BigInt, BigInt)> {
...
+            buf = &buf[1..];
         }
+    }
 
-        if c == b'_' && last == b'_' {
-            return None;
+    // Reject empty strings
+    let mut prev = *buf
+        .first()
+        .ok_or(BytesToIntError::InvalidLiteral { base })?;
```
youknowone, line 107:
```suggestion
    let (mut prev, buf) = buf
        .split_first()
        .ok_or(BytesToIntError::InvalidLiteral { base })?;
```
Then skip() on line 116 will not be required.

youknowone (changes requested):
Thanks! Looks good in general. Please check the comments

vm/src/protocol/number.rs
ShaharNaveh: ofc, will take a look

common/src/int.rs
ShaharNaveh: Good catch!

vm/src/protocol/number.rs
ShaharNaveh: @youknowone I tried to keep `bytes_to_int` in "common", but having `&PyObject` inside the enum value forces me to move it to `vm/src/`, should it be done?

common/src/int.rs
ShaharNaveh: Actually, we can't do it.

because we after we count the digits we pass the unmodified buf to bigint. so unless we want to split it, save the original first element, then iterate over the buffer, and then concat the original first element with the rest of the slice.
I think this option is a more complicated than it needs to be. but it's not a deal breaker for me

vm/src/protocol/number.rs
ShaharNaveh: btw, what's the difference between `IntoPyException` and `ToPyException`?

youknowone: The naming came from https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv
Into consumes the source, but To doesn't.

Issue #6093, opened 2025-08-13
[RFC] Replace rust-timsort so RustPython sorting speed ~= CPython sorting speed


jackoconnordev:
[WIP/Draft PR](https://github.com/RustPython/RustPython/pull/6096)

## Summary

Replace the [single usage of rust-timsort](https://github.com/RustPython/RustPython/blob/a9a9e3bf118f6c631ef99fe7174215254f846097/vm/src/builtins/list.rs#L525-L528) with a more performant sorting algorithm. Either Rust's default sorting algorithm or a crate like [glidesort](https://crates.io/crates/glidesort) which has positive benchmarks against the default implementation.

## Detailed Explanation

As of CPython 3.11, [Timsort is no longer the default sorting algorithm](https://www.wild-inter.net/posts/powersort-in-python-3.11). It is instead an algorithm called Powersort.

### rust-timsort

RustPython currently uses a forked project https://github.com/RustPython/rust-timsort as its default sorting algorithm. To quote rust-timsort's README.

> This is still an extreme work-in-progress, and performance has vast room for improvement.

This performance gap has become noticeable as new unit tests in later Python versions are introduced, which attempt to sort larger and larger lists. Sorting 1 million random numbers with rust-timsort takes 10-20 minutes vs 0.3 seconds for CPython.

```shell
RustPython on  statistics-module-kde-function [$!?] via :snake: v3.13.1 via :crab: v1.88.0 
✦ ❯ time python -c "from random import random; sorted([random() for i in range(1_000_000)]); print('DONE');"
DONE

real    0m0.309s
user    0m0.274s
sys    0m0.036s

RustPython on  statistics-module-kde-function [$!?] via :snake: v3.13.1 via :crab: v1.88.0 
✦ ❯ time cargo run --release -- -c "from random import random; sorted([random() for i in range(1_000_000)]); print('DONE');"
    Finished `release` profile [optimized] target(s) in 0.16s
     Running `target/release/rustpython -c 'from random import random; sorted([random() for i in range(1_000_000)]); print('\''DONE'\'');'`
DONE

real    16m52.217s
user    16m51.926s
sys    0m0.174s
```

### glidesort

[...]

PR #6096, opened 2025-08-15
[DRAFT] Use Rust's `sort_by` instead of `timsort::try_sort_by_gt`

jackoconnordev:
Relates #6093 

## Time improvements
**rust-timsort**
(Timing from Linux laptop - I didn't repeat this experiment on my Mac as it is 16 mins long)
```sh
✦ ❯ time cargo run --release -- -c "from random import random;
sorted([random() for i in range(1_000_000)]); print('DONE');"
    Finished `release` profile [optimized] target(s) in 0.16s
     Running `target/release/rustpython -c 'from random import random;
sorted([random() for i in range(1_000_000)]); print('\''DONE'\'');'`
DONE

real    16m52.217s
user    16m51.926s
sys     0m0.174s
```

**Built-in**
(Timing from MacOS laptop)
```sh
❯ time cargo run --release -- -c "from random import random;
sorted([random() for i in range(1_000_000)]); print('DONE');"
    Finished `release` profile [optimized] target(s) in 0.89s
     Running `target/release/rustpython -c 'from random import random;
sorted([random() for i in range(1_000_000)]); print('\''DONE'\'');'`
DONE
cargo run --release -- -c   0.96s user 0.23s system 54% cpu 2.196 total

```

## Verify list is sorted
```sh
❯ cargo run --release -- -c "from random import random; l =
sorted([random() for i in range(1_000_000)]); print(all(l[i] <= l[i+1]
for i in range(len(l) - 1))               ─╯
);"
TRUE
```

## TODO
- This uses `.unwrap` because I don't know how to handle the error case
- This doesn't use Ordering::Equal so it needs to either be confirmed that this is a stable sorting algorithm, or else make it stable

### DIFF
--- a/vm/src/builtins/list.rs
+++ b/vm/src/builtins/list.rs
@@ -504,8 +504,8 @@ impl Representable for PyList {
     }
 }
 
-use std::cmp::Ordering;
 use crate::exceptions::types::PyBaseExceptionRef;
+use std::cmp::Ordering;
 fn do_sort(
     vm: &VirtualMachine,
     values: &mut Vec<PyObjectRef>,
@@ -520,13 +520,17 @@ fn do_sort(
 
     // If a PyException is encountered stop swapping elements and store the error.
     let mut error_slot: Option<PyBaseExceptionRef> = None;
-    let mut cmp = |a: &PyObjectRef, b: &PyObjectRef| {
-        match a.rich_compare_bool(b, op, vm) {
-            Ok(res) => if res { Ordering::Greater } else { Ordering::Less },
-            Err(e) => {
-                error_slot = Some(e);
-                Ordering::Equal
-            },
+    let mut cmp = |a: &PyObjectRef, b: &PyObjectRef| match a.rich_compare_bool(b, op, vm) {
+        Ok(res) => {
+            if res {
+                Ordering::Greater
+            } else {
+                Ordering::Less
+            }
+        }
+        Err(e) => {
+            error_slot = Some(e);
+            Ordering::Equal
         }
     };
 
--- a/vm/src/builtins/list.rs
+++ b/vm/src/builtins/list.rs
@@ -543,7 +543,7 @@ fn do_sort(
         items.sort_unstable_by(|a, b| cmp(&a.1, &b.1));
         *values = items.into_iter().map(|(val, _)| val).collect();
     } else {
-        values.sort_unstable_by(cmp(a, b));
+        values.sort_unstable_by(cmp);
     }
 
     // After the sort is done, check if an error was captured.
--- a/vm/src/builtins/list.rs
+++ b/vm/src/builtins/list.rs
@@ -504,6 +504,7 @@ impl Representable for PyList {
     }
 }
 
+use std::cmp::Ordering;
 fn do_sort(
     vm: &VirtualMachine,
     values: &mut Vec<PyObjectRef>,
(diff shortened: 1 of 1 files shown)


vm/src/builtins/list.rs
```diff
@@ -515,17 +516,26 @@ fn do_sort(
     } else {
         PyComparisonOp::Gt
     };
-    let cmp = |a: &PyObjectRef, b: &PyObjectRef| a.rich_compare_bool(b, op, vm);
+    let cmp = |a: &PyObjectRef, b: &PyObjectRef| {
+        let res = a.rich_compare_bool(b, op, vm).unwrap();
```
jackoconnordev: @coderabbitai can you think of a good way I can unwrap this and correctly handle the `PyRef<PyBaseException>` in the error case?

vm/src/builtins/list.rs
```diff
@@ -515,17 +516,26 @@ fn do_sort(
...
+    };
 
     if let Some(ref key_func) = key_func {
         let mut items = values
             .iter()
             .map(|x| Ok((x.clone(), key_func.call((x.clone(),), vm)?)))
             .collect::<Result<Vec<_>, _>>()?;
-        timsort::try_sort_by_gt(&mut items, |a, b| cmp(&a.1, &b.1))?;
+        // timsort::try_sort_by_gt(&mut items, |a, b| cmp(&a.1, &b.1))?;
+        items.sort_by(|a, b| cmp(&a.1, &b.1));
```
jackoconnordev: @coderabbitai is rust's builtin sort_by a stable sorting algorithm if only Ordering::Less and Ordering::Greater are returned by `cmp`?

jackoconnordev: Do you know what the True and False return values of `rich_compare_bool` correspond to semantically?

vm/src/builtins/list.rs
ShaharNaveh:
```suggestion
        let res = a.rich_compare_bool(b, op, vm)?;
```

Does this do the trick?

if not you can do it with:
```rust
let res = match a.rich_compare_bool(b, op, vm) {
  OK(v)=> v,
  Err(e) => return e,
};
```

jackoconnordev: There were compiler errors for both versions to do with returning a value that isn't an `Ordering` :-/

ShaharNaveh: > There were compiler errors for both versions to do with returning a value that isn't an `Ordering` :-/


Not so sure why it fails and what's the issue but you can try with:
```suggestion
        let res = a.rich_compare_bool(b, op, vm).unwrap_or(Ordering::Equal);
```

And hope that the CI passes

ShaharNaveh:
@jackoconnordev I've dug a bit more and it seems that we need a failable sorting mechanism. maybe try to use something like https://docs.rs/try-partialord/latest/try_partialord/

jackoconnordev:
I think if this was Rust < 1.81.0 this might actually pass https://users.rust-lang.org/t/panic-on-sort-of-f32-s-with-f32-total-cmp-in-1-81-0/117675

vm/src/builtins/list.rs
```diff
@@ -515,20 +517,40 @@ fn do_sort(
...
+
+    // If a PyException is encountered stop swapping elements and store the error.
+    let mut error_slot: Option<PyBaseExceptionRef> = None;
+    let mut cmp = |a: &PyObjectRef, b: &PyObjectRef| match a.rich_compare_bool(b, op, vm) {
+        Ok(res) => {
+            if res {
+                Ordering::Greater
+            } else {
+                Ordering::Less
+            }
```
youknowone, line 529: The meaning of true/false will be different by `op`

kangdora:
I'd like to pick this up — planning to continue from #6096 .

jackoconnordev:
Go for it! I'd love to learn from a working solution and the speed up is well worth the effort

PR #6102, opened 2025-08-18
lookup slot in hash()

giammirove:
# Description

`str.__hash__` can’t be overwritten, so for exact `PyStr` instances we skip the attribute lookup and MRO walk and call the string hash directly. This reduces overhead on hot paths (e.g., dict/set key hashing) while keeping subclass semantics unchanged.

```
>>> str.__hash__ = None
  File "<python-input-0>", line 1, in <module>
    str.__hash__ = None
    ^^^^^^^^^^^^
TypeError: cannot set '__hash__' attribute of immutable type 'str'
```
in Python 3.13.5.

## Summary by CodeRabbit

* **Refactor**
  * Object hashing now uses type-defined hash behavior via inheritance lookup, ensuring consistent dispatch for custom types.
  * If a type provides no hash slot, an explicit TypeError ("unhashable type: '<TypeName>'") is raised.
  * No public API or signature changes; behavior for most built-in types is preserved.

### DIFF
--- a/vm/src/protocol/object.rs
+++ b/vm/src/protocol/object.rs
@@ -642,6 +642,11 @@ impl PyObject {
     }
 
     pub fn hash(&self, vm: &VirtualMachine) -> PyResult<PyHash> {
+        // __hash__ function in str cannot be overwritten
+        if let Some(pystr) = self.downcast_ref_if_exact::<PyStr>(vm) {
+            return Ok(pystr.hash(vm));
+        }
+
         let hash = self.get_class_attr(identifier!(vm, __hash__)).unwrap();
         if vm.is_none(&hash) {
             return Err(vm.new_exception_msg(
--- a/vm/src/protocol/object.rs
+++ b/vm/src/protocol/object.rs
@@ -646,10 +646,10 @@ impl PyObject {
             return hash(self, vm);
         }
 
-        return Err(vm.new_exception_msg(
+        Err(vm.new_exception_msg(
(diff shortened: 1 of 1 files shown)


youknowone:
Thank you for contributing!

I agree checking `get_class_attr` must be removed to improve performance, but specifically checking `str` doesn't look like the answer. Ideally, we must be able to quickly looking in type slots hash of MRO. Because str directly has hash function in its type, it must be fast enough after removing `get_class_attr`.

giammirove:
Thank you for your feedback.

I intended to reduce to a minimum the number of calls and lock acquisitions.
`str`, as other built-in types, has `__hash__` method, which is immutable.
Even removing `get_class_attr`, we call at least once `cls.slots.hash.load()`, requiring a lock.
Not holding the lock should not violate consistency since the method is known to be immutable. (I'm not a Python expert, correct me in case.)

On your note, my proposal then is to remove `get_class_attr`.
The output of `mro_find_map` will be used to obtain the hash function and check its existence.
This solution maintains unchanged the semantics of the function, since the value returned by `mro_find_map` is obtained using `.unwrap()`.
This works only if the following holds:
`self.get_class_attr(identifier!(vm, __hash__)).unwrap() is not None` <=> `self.class() .mro_find_map(|cls|cls.slots.hash.load()) is not None`

youknowone:
@giammirove oh, I didn't notice new code is pushed. sorry for missing the review

PR #6104, opened 2025-08-20
sqlite: Fix missing ProgrammingError for parameter mismatch

ever0de:
ref 

- https://github.com/python/cpython/blob/d22a7456443415961b93169b851ab7e550be8791/Modules/_sqlite/cursor.c#L640-L683

## Summary by CodeRabbit

* **Bug Fixes**
  * Validates the number of parameters required by prepared SQLite statements before execution to prevent missing or extra bindings.
  * Emits clearer ProgrammingError messages indicating how many bindings are required versus supplied, including when none are provided.
  * Improves reliability of parameterized queries by rejecting mismatched bindings early without changing the public API.


youknowone (approved):
👍

PR #6116, opened 2025-08-31
wasm_js rustlfags on top workspace

youknowone:
## Summary by CodeRabbit

- Bug Fixes
  - Improved reliability of WebAssembly builds by ensuring a browser-compatible random backend is selected for wasm32 targets, reducing initialization errors in non-WASI WASM environments.

- Chores
  - Consolidated WASM build configuration at the workspace level for consistent behavior across platforms.
  - Added a workspace-level WASM feature and updated package feature mappings to streamline cross-package builds and reduce duplication.

### DIFF
--- a/common/Cargo.toml
+++ b/common/Cargo.toml
@@ -10,6 +10,7 @@ license.workspace = true
 
 [features]
 threading = ["parking_lot"]
+wasm_js = ["getrandom/wasm_js"]
 
 [dependencies]
 rustpython-literal = { workspace = true }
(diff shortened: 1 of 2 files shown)


coolreader18 (approved):

coolreader18:
Ah, actually - that test failure is why the config is the way it is. Putting it in the top level means js gets selected even when we're not compiling wasm for browser.

common/Cargo.toml
```diff
@@ -36,6 +36,9 @@ lock_api = "0.4"
 siphasher = "1"
 num-complex.workspace = true
 
+[target.'cfg(all(target_arch = "wasm32", not(target_os = "wasi")))'.dependencies]
+getrandom = { workspace = true, features = ["wasm_js"] }
+
```
youknowone:
```suggestion
```
This is wrong.

coolreader18: Hmm, true. I guess that should be enabled by a `js` feature flag that common exposes.

youknowone:
@coolreader18 better?

coolreader18 (approved):

PR #5795, opened 2025-05-19
Add os constants

arihant2math:

PR #6117, opened 2025-08-31
Attenpt to automate posix consts

ShaharNaveh:
Attempt to automate the definition of posix consts. I'm sure that this will require the script to be more adjustable but IMO this is a good first step in automating something that we do manually. for example: https://github.com/RustPython/RustPython/pull/5795

I ain't so sure how we decide what goes into `os` directly and what into `posix`, if someone can share some knowledge it would be very appreciated:)

### DIFF
--- a/scripts/libc_posix.py
+++ b/scripts/libc_posix.py
@@ -42,6 +42,11 @@ EXCLUDE = frozenset(
     }
 )
 
+EXTRAS = {
+    frozenset({"macos"}): {"COPYFILE_DATA"},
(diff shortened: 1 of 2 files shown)


youknowone:
> I ain't so sure how we decide what goes into os directly and what into posix

I can say it is arbitrary based on history of the module. Currently `os.rs` tends to have shared definitions between `nt`, `posix`, and `posix_compat`, but not strictly. You can safely relocate everything if it is better.

PR #6118, opened 2025-09-01
sqlite3: Support 'size' keyword argument in `Cursor::fetchmany`

ever0de:
## Summary by CodeRabbit

* **New Features**
  * Cursor.fetchmany now accepts a size keyword argument (via an argument bag) and uses arraysize when omitted.
* **Bug Fixes**
  * Clearer handling of invalid argument combinations (too many positional args, unexpected keywords, or duplicate size).
  * Behavior: non‑positive size fetches all remaining rows; positive size limits returned rows.

### DIFF
--- a/stdlib/src/sqlite.rs
+++ b/stdlib/src/sqlite.rs
@@ -1461,6 +1461,12 @@ mod _sqlite {
         statement: Option<PyRef<Statement>>,
     }
 
+    #[derive(FromArgs)]
+    struct FetchManyArgs {
+        #[pyarg(any, name = "size", optional)]
+        size: Option<c_int>,
+    }
+
     #[pyclass(with(Constructor, IterNext, Iterable), flags(BASETYPE))]
     impl Cursor {
         fn new(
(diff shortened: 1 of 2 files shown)


stdlib/src/sqlite.rs
```diff
@@ -1684,14 +1684,40 @@ mod _sqlite {
         #[pymethod]
         fn fetchmany(
             zelf: &Py<Self>,
-            max_rows: OptionalArg<c_int>,
+            mut args: FuncArgs,
```
youknowone: Can `#[derive(FromArgs)]` be applicable? Or is there any blocker not allowing to do it?

ever0de: We can change it. However, I believe we can't use the FromArgs derive macro because the error messages would be different.

[`2e16f51` (#6118)](https://github.com/RustPython/RustPython/pull/6118/commits/2e16f51c68dde8d8c0f79d5745b874ea043f762e)

```diff
 --- Testing with too many positional arguments ---
-fetchmany() takes from 0 to 1 positional arguments but 2 were given  (cpython)
+expected at most 2 arguments, got 3 (FromArgs)
 
 --- Testing with an unexpected keyword argument ---
-fetchmany() got an unexpected keyword argument 'unexpected_kw' (cpython)
+Unexpected keyword argument unexpected_kw (FromArgs)
 
 --- Testing with both positional and keyword for the same argument ---
-fetchmany() got multiple values for argument 'size' (cpython)
+Unexpected keyword argument size (FromArgs)
```

ever0de: If you think the current error messages are acceptable, I believe we can apply this version as is.

coolreader18: Yeah, those error messages are fine - there's discrepancy in so many places, that if we wanted to make the error messages line up we'd just change FromArgs all at once.

PR #6120, opened 2025-09-01
Update nix to 0.30

coolreader18:
Now much more of nix uses IO-safe types!

## Summary by CodeRabbit

- New Features
  - Safer POSIX process spawning with optional setsid support on supported platforms.

- Bug Fixes
  - More consistent FD inheritance and blocking behavior across POSIX and subprocess paths.
  - Clearer error messages for signal non-blocking checks.
  - Reduced FD churn and improved reliability for memory-mapped files.
  - Minor fixes for IO and directory listing from file descriptors.

- Refactor
  - Large migration from raw FD integers to safer, lifetime-aware FD handling; duplication/dup2 semantics improved.

- Chores
  - Updated POSIX-related dependency to a newer minor version.

### DIFF
--- a/stdlib/src/mmap.rs
+++ b/stdlib/src/mmap.rs
@@ -27,11 +27,9 @@ mod mmap {
     use nix::sys::stat::fstat;
     use nix::unistd;
     use num_traits::Signed;
-    use std::fs::File;
+    use rustpython_common::crt_fd;
     use std::io::{self, Write};
     use std::ops::{Deref, DerefMut};
-    #[cfg(unix)]
-    use std::os::unix::io::{FromRawFd, RawFd};
 
     fn advice_try_from_i32(vm: &VirtualMachine, i: i32) -> PyResult<Advice> {
         Ok(match i {
@@ -179,7 +177,7 @@ mod mmap {
     struct PyMmap {
         closed: AtomicCell<bool>,
         mmap: PyMutex<Option<MmapObj>>,
-        fd: RawFd,
+        fd: AtomicCell<i32>,
         offset: libc::off_t,
         size: AtomicCell<usize>,
         pos: AtomicCell<usize>, // relative to offset
@@ -190,7 +188,7 @@ mod mmap {
     #[derive(FromArgs)]
     struct MmapNewArgs {
         #[pyarg(any)]
-        fileno: RawFd,
+        fileno: i32,
         #[pyarg(any)]
         length: isize,
         #[pyarg(any, default = MAP_SHARED)]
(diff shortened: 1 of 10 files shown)


stdlib/src/posixsubprocess.rs
```diff
@@ -121,21 +120,95 @@ impl CharPtrSlice<'_> {
...
 }
 
+#[derive(Copy, Clone)]
+struct Fd(BorrowedFd<'static>);
+
+impl TryFromObject for Fd {
+    fn try_from_object(vm: &VirtualMachine, obj: PyObjectRef) -> PyResult<Self> {
+        match MaybeFd::try_from_object(vm, obj)? {
+            MaybeFd::Some(fd) => Ok(fd),
+            MaybeFd::None => Err(vm.new_value_error("invalid fd".to_owned())),
```
ShaharNaveh:
```suggestion
            MaybeFd::None => Err(vm.new_value_error("invalid fd")),
```

stdlib/src/posixsubprocess.rs
```diff
@@ -121,21 +120,95 @@ impl CharPtrSlice<'_> {
...
+}
+
+impl From<OwnedFd> for Fd {
+    fn from(fd: OwnedFd) -> Self {
+        Self(unsafe { BorrowedFd::borrow_raw(fd.into_raw_fd()) })
+    }
+}
+
+#[derive(Copy, Clone)]
+enum MaybeFd {
```
ShaharNaveh, line 170: Wouldn't it be more correct to have:
```rust
type MaybeFd = Option<(FD type)>
```

or maybe even a newtype:
```rust
struct MaybeFd<T>(Option<T>)
```
?

coolreader18: In this case, None corresponds to `-1`, not `PyNone`

ShaharNaveh: ah, I see. Then this is probably the best I can think about atm

youknowone: Is Some/None naming the best? It looks a bit confusing. 

If it needs option-like operations, `#[derive(result_like::OptionLike)]` will be helpful.

Then this code also work
```
if let Some(value) = maybe_fd.into_option() {
    ...
}
```

coolreader18: Good point. I've renamed the variants, which I think makes it more clear.

stdlib/src/mmap.rs
```diff
@@ -340,7 +338,8 @@ mod mmap {
                 }
             };
 
-            if fd != -1 {
+            let fd = unsafe { crt_fd::Borrowed::try_borrow_raw(fd) }.ok();
+            if let Some(fd) = fd {
```
youknowone:
```suggestion
            let fd = unsafe { crt_fd::Borrowed::try_borrow_raw(fd) };
            if let Ok(fd) = fd {
```
No benefit for conversion

youknowone (approved):
looks good. please check the comment before merge

PR #6121, opened 2025-09-01
`fn unparse_expr` -> `UnparseExpr::new`

ShaharNaveh:


youknowone (approved):
The change looks good.

On the other hand, Ruff also have the corresponding part. Checking if we can share the implement may be worth

PR #6124, opened 2025-09-03
Use ruff for `Expr` unparsing

ShaharNaveh:
Follow up on https://github.com/RustPython/RustPython/pull/6121#pullrequestreview-3176103527

## Summary by CodeRabbit

* **Chores**
  * Updated workspace dependencies to include new Python code generation and parsing libraries.
  * Refactored internal code generation utilities for improved maintainability.

### DIFF
--- a/Lib/test/test_typing.py
+++ b/Lib/test/test_typing.py
@@ -4952,7 +4952,6 @@ class GenericTests(BaseTestCase):
         def barfoo2(x: CT): ...
         self.assertIs(get_type_hints(barfoo2, globals(), locals())['x'], CT)
 
-    @unittest.expectedFailure # TODO: RUSTPYTHON; 'List[list["C2"]]' != "List[list['C2']]"
     def test_generic_pep585_forward_ref(self):
         # See https://bugs.python.org/issue41370
 


compiler/codegen/src/compile.rs
```diff
@@ -147,6 +147,19 @@ enum ComprehensionType {
     Dict,
 }
 
+fn unparse_expr(expr: &Expr) -> String {
+    // Hack, because we can't do `ruff_python_codegen::Indentation::default()`
+    // https://github.com/astral-sh/ruff/pull/20216
```
ShaharNaveh: I see that it was already merged  https://github.com/astral-sh/ruff/pull/20216 (that was fast 😅 )

I'll wait for it be released, and I wanted to upgrade ruff anyway so it will be a good excuse to do so

Lib/test/test_typing.py
```diff
@@ -4952,6 +4952,7 @@ def barfoo(x: AT): ...
         def barfoo2(x: CT): ...
         self.assertIs(get_type_hints(barfoo2, globals(), locals())['x'], CT)
 
+    @unittest.expectedFailure # TODO: RUSTPYTHON; 'List[list["C2"]]' != "List[list['C2']]"
```
youknowone: is there any option to change the quote style?

ShaharNaveh: No:/ not from the `Generator` directly, I think that this needs to be fixed upstream (over Ruff's side)

ShaharNaveh:
ngl, I have mixed feelings about this PR. on one hand it's a major code reduction, and on the other hand there is a regression when it comes to the test cases

Lib/test/test_typing.py
youknowone: Once Ruff forked rustpython-parser, we patched unparse to decide quote by option. That might be removed during refactoring. I have no idea if Ruff maintainers are interested in adding feature only for downstream user. Probably asking if possible be worth. Otherwise fork? but in easier form to sync.

ShaharNaveh: @youknowone It will supported in the next ruff release.
https://github.com/astral-sh/ruff/pull/20434

youknowone: Great, thank you so much!

youknowone (approved):
Less code, no regression. Great!

ShaharNaveh:
> Less code, no regression. Great!

There's regression:/

It generates tuples without `()` so what used to be `(1, 2)` is now `1, 2`. 

I forgot to point that out before, I think it's better to revert the changes and discuss it

PR #6132, opened 2025-09-05
Update `{site,sysconfig}.py` from 3.13.7

ShaharNaveh:


Lib/site.py
```diff
@@ -270,23 +282,26 @@ def check_enableusersite():
 #
 # See https://bugs.python.org/issue29585
 
+# Copy of sysconfig._get_implementation()
+def _get_implementation():
```
youknowone, line 286: 👍 finally this is shipped

vm/src/stdlib/sysconfig.rs
```diff
@@ -0,0 +1,14 @@
+pub(crate) use sysconfig::make_module;
+
+#[pymodule(name = "_sysconfig")]
+pub(crate) mod sysconfig {
+    use crate::{VirtualMachine, builtins::PyDictRef, convert::ToPyObject};
+
+    #[pyfunction]
+    fn config_vars(vm: &VirtualMachine) -> PyDictRef {
+        let vars = vm.ctx.new_dict();
+        vars.set_item("Py_GIL_DISABLED", false.to_pyobject(vm), vm)
```
youknowone:
```suggestion
        vars.set_item("Py_GIL_DISABLED", true.to_pyobject(vm), vm)
```
Does this work with true value? Because we don't implement GIL

youknowone (approved):
👍

vm/src/stdlib/sysconfig.rs
ShaharNaveh: I just put this var as a placeholder. let's see what will be the result

PR #6143, opened 2025-09-10
Fix broken CI on windows

ShaharNaveh:


ShaharNaveh:
Alright, I did a bit of investigation. my conclusion is that github changed the `windows-latest` image from being `windows-2022` to `windows-2025`.

@youknowone how should we proceed? should we merge this PR? or should we pin the CI to use `windows-2022`? or even add another runner so we will check both `2022` and `2025` (if we do it, our patches will need to be slightly different)

PR #6151, opened 2025-09-15
Dependencies cleanup

ShaharNaveh:
- removed some unused dependencies
- update some dependencies
- be less strict on the version of dependencies

## Summary by CodeRabbit

- Chores
  - Updated many dependencies (minor/patch bumps), added a new base utility dev dependency, and removed/narrowed several workspace/dev/runtime dependencies across platforms.

- Tests
  - Benchmarking harness updated to use improved Python GIL handling and standard-library benchmarking helpers; observable behavior unchanged.

- Refactor
  - Dependency and workspace configuration reorganized for clarity; no public API changes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -163,27 +163,27 @@ ruff_python_ast = { git = "https://github.com/astral-sh/ruff.git", tag = "0.11.0
 ruff_text_size = { git = "https://github.com/astral-sh/ruff.git", tag = "0.11.0" }
 ruff_source_file = { git = "https://github.com/astral-sh/ruff.git", tag = "0.11.0" }
 
-ahash = "0.8"
+ahash = "0.8.12"
 ascii = "1.1"
-bitflags = "2.9"
+bitflags = "2.9.4"
 bstr = "1"
 cfg-if = "1.0"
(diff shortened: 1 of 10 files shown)


Cargo.toml
```diff
@@ -163,27 +163,27 @@ ruff_python_ast = { git = "https://github.com/astral-sh/ruff.git", tag = "0.11.0
...
 
-ahash = "0.8.11"
+ahash = "0.8"
 ascii = "1.1"
-bitflags = "2.9.1"
+bitflags = "2.9"
 bstr = "1"
 cfg-if = "1.0"
-chrono = "0.4.39"
+chrono = "0.4"
```
youknowone: We are using chrono features of later versions. e.g. 0.4.10 is not compatible. This will cause dependency resolver success but build fails.

ShaharNaveh: fixed

youknowone: How about other libraries? To remove patch version, we have to ensure it is buildable with x.y.0

ShaharNaveh: good point, I'll revert. And will open a PR for base dependentabot config. this should ease the upgrade process

PR #6156, opened 2025-09-17
Impl `_opcode`. Update `dis` from 3.13.7

ShaharNaveh:


ShaharNaveh:
@coderabbitai review

ShaharNaveh:
@coderabbitai help

stdlib/src/opcode.rs
```diff
@@ -0,0 +1,282 @@
...
+        pub const fn has_const(opcode: i32) -> bool {
+            Self::is_valid(opcode) && matches!(opcode, 83 | 103 | 240)
+        }
+
+        #[must_use]
+        pub const fn has_name(opcode: i32) -> bool {
+            Self::is_valid(opcode)
+                && matches!(
+                    opcode,
+                    63 | 66
```
youknowone, line 64: where the magic numbers came from? will this be changed if we change instruction enum?

youknowone:
finally we get `_opcode`, great!

stdlib/src/opcode.rs
ShaharNaveh: Oh, I just ran:
```py
import opcode

for name in (
    "hasarg",
    "hasconst",
    "hasname",
    "hasjump",
    "hasjrel",
    "hasjabs",
    "hasfree",
    "haslocal",
    "hasexc",
):
    lst = getattr(opcode, name)
    print(f"{name}\n{lst}")
```

And copied the results xD

---

But it's calculated based on https://github.com/python/cpython/blob/bcee1c322115c581da27600f2ae55e5439c027eb/Include/internal/pycore_opcode_metadata.h#L966-L1190

I've put a comment at https://github.com/ShaharNaveh/RustPython/blob/a7b9356da2a85cbc02df1f0fac9eda6a26869b6d/stdlib/src/opcode.rs#L46-L47 stating this:)

---

We should have our own metadata table, but I think that's a bit out of scope for this PR

ShaharNaveh: At https://github.com/RustPython/RustPython/blob/aa0eb4bedf7143334b686a90cf4182a1ebb429e6/compiler/core/src/bytecode.rs#L534-L541

I tried doing something like:
```rust

 #[repr(u8)] 
 pub enum Instruction { 
     Nop = 30, 
     ImportName { 
         idx: Arg<NameIdx>, 
     } = 75, 
     ...
}
```

Based on:
- `Nop`: https://github.com/python/cpython/blob/bcee1c322115c581da27600f2ae55e5439c027eb/Include/opcode_ids.h#L43
- `ImportName`: https://github.com/python/cpython/blob/bcee1c322115c581da27600f2ae55e5439c027eb/Include/opcode_ids.h#L88

But this change alone was enough to break RustPython entirely.  

@youknowone can you please point me to where any of:
https://github.com/RustPython/RustPython/blob/aa0eb4bedf7143334b686a90cf4182a1ebb429e6/compiler/core/src/bytecode.rs#L796-L815

are being called? I've tried looking into it and got a bit lost

youknowone: Thank you for explanation. Since we occasionally break bytecode compatibility, the numbers need to be regenerated.
Could you also attach the updating code around `enum Instruction` not to let it be desynced?

The `From<Instruction>` part is used form marshal. cc @coolreader18

ShaharNaveh: > Since we occasionally break bytecode compatibility

Is there a reason for that? or it's something that's desired but not implemented yet?

> the numbers need to be regenerated. Could you also attach the updating code around `enum Instruction` not to let it be desynced?

I haven't written any code to do that (yet, but was planning to). As I just changed those two enum members and I couldn't even do `cargo run` :/

> 
> The `From<Instruction>` part is used form marshal. cc @coolreader18

ty:) will look into it.

ShaharNaveh: @youknowone I have opened https://github.com/RustPython/RustPython/pull/6174 so we can discuss the next steps there and to not block this PR

Lib/_opcode_metadata.py
```diff
@@ -0,0 +1,343 @@
...
+    'INSTRUMENTED_CALL_KW': 245,
+    'INSTRUMENTED_CALL_FUNCTION_EX': 246,
+    'INSTRUMENTED_INSTRUCTION': 247,
+    'INSTRUMENTED_JUMP_FORWARD': 248,
+    'INSTRUMENTED_JUMP_BACKWARD': 249,
+    'INSTRUMENTED_POP_JUMP_IF_TRUE': 250,
+    'INSTRUMENTED_POP_JUMP_IF_FALSE': 251,
+    'INSTRUMENTED_POP_JUMP_IF_NONE': 252,
+    'INSTRUMENTED_POP_JUMP_IF_NOT_NONE': 253,
+    'JUMP': 256,
```
youknowone, line 328: I didn't know it can be over 256, huh

stdlib/src/opcode.rs
youknowone: > Is there a reason for that? or it's something that's desired but not implemented yet?

No specific reason. We just add or remove instructions by following CPython update. The order changing is unintended side effect. Python doesn't guarantee bytecode compatibiltity anyway.

Does giving specific numbers to each instruction make it easier to be maintained? I guess so.

After #6174, then can we rewrite `63 | 66` to `Instruction::Abc as u8 | Instruction::Def as u8` and so on?

Lib/_opcode_metadata.py
ShaharNaveh: Those are pseudo opcodes:)
Real opcodes can't be more than 256

ShaharNaveh:
For some reason I can't comment on the thread of https://github.com/RustPython/RustPython/pull/6156#discussion_r2375758287 so answering it here:

> Does giving specific numbers to each instruction make it easier to be maintained? I guess so.

Yes! it also makes it easier for writing the helpers of:

- `is_valid`
- `has_arg`
- `has_const`
- `has_name`
- etc...

but that wouldn't be needed as I intend #6174 to generate those automatically (just like CPython does it)

> After #6174, then can we rewrite 63 | 66 to Instruction::Abc as u8 | Instruction::Def as u8 and so on?

Well, yes you could. but I thought that we could have an API that looks similar to this:
```rust
// This is inside the `_opcode` module
fn has_arg(opcode: i32) -> bool {
  Instruction::try_from(opcode).map_or(false, |v| !v.is_pseudo() && v.has_arg())
}
```

where:
- `Instruction::try_from`
- `instruction.is_pseudo()` 
- `instruction.has_arg()` 

are all auto generated.

youknowone (approved):
Looks good in general.

I am worrying about the newly introduced magic numbers. I wish it could be quickly resolved in next patch

ShaharNaveh:
> I am worrying about the newly introduced magic numbers. I wish it could be quickly resolved in next patch

I'll send a quick patch to resolve it in the next couple of days

PR #6157, opened 2025-09-20
Run scheduled CI jobs only on upstream repo

ShaharNaveh:
There's no point in running those tasks on forks anyway imo.
---
I wish Github Actions allowed top level `if`:/
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,8 @@ env:
 jobs:
   build:
     runs-on: ${{ matrix.platform.runner }}
+    # Disable this job when running on a fork.
+    if: github.repository == 'RustPython/RustPython'
     strategy:
       matrix:
         platform:
@@ -88,6 +90,8 @@ jobs:
 
   build-wasm:
     runs-on: ubuntu-latest
+    # Disable this job when running on a fork.
+    if: github.repository == 'RustPython/RustPython'
     steps:
       - uses: actions/checkout@v4
       - uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -21,6 +21,8 @@ env:
 jobs:
   build:
     runs-on: ${{ matrix.platform.runner }}
+    # Disable this job when running on a fork.
+    if: github.repository == 'RustPython/RustPython'
```
youknowone: Though I don't know well about actions. What if we need to edit release.yml? Then isn't it better to run it on the fork?

ShaharNaveh: > Though I don't know well about actions. What if we need to edit release.yml? Then isn't it better to run it on the fork?

Not sure I understand your concern. this only prevents this job from triggering on RustPython forks, like mine:
https://github.com/ShaharNaveh/RustPython/actions/workflows/release.yml

I see that you disabled it manually for yourself lol:
https://github.com/youknowone/RustPython/actions/workflows/cron-ci.yaml

ShaharNaveh: I think that I get what you mean, we can have something like:
```yaml
if: ${{ github.repository == 'RustPython/RustPython' || github.event_name != 'schedule' }}
```

this way it will always run on the upstream repo and can still be triggered with workflow_dispatch on forks

youknowone: Oh, I can't remember why I did it. Probably cron-ci was broken once a while ago. And never be fixed and forgotten.

The event_name check makes sense. Thanks!

youknowone (approved):

PR #6158, opened 2025-09-20
Reconfigure dependabot

ShaharNaveh:


.github/dependabot.yml
```diff
@@ -1,13 +1,10 @@
...
+  - package-ecosystem: cargo
+    directory: /
+    schedule:
+      interval: weekly
   - package-ecosystem: github-actions
     directory: /
-    groups:
-      github-actions:
-        patterns:
-          - "*"  # Group all Actions updates into a single larger pull request
```
ShaharNaveh, line 11: IMO this is something that prevents us from updating our actions gradually and prevents us from finding which action is the problematic one

PR #6170, opened 2025-09-20
Update ruff to 0.11.12

ShaharNaveh:
Not updating to the latest ruff version as the diff will be very large and hard to review. so I'm doing it in stages until we'll get to `>=0.13`


ShaharNaveh:
Will reopen when I resolve the regression

youknowone:
Please feel free to keep it open and working on it. No problem to do that

PR #6177, opened 2025-09-29
Update to ruff 0.13.1

coolreader18:
The biggest contributor to the diff is from ruff adding a `node_index` field to every AST node. Unfortunately, I had to add them all manually, but I'd like to at some point bring back the [asdl_rs.py](https://github.com/RustPython/Parser/blob/main/ast/asdl_rs.py) codegen script to make it less odious to edit.


coolreader18:
I'm thinking about possibly seeing if it'd make sense to publish a `rustpython-ruff-python-parser` package, so that we're able to publish to crates.io again without making ruff worry about maintaining published crates.

youknowone:
see also #6170

ShaharNaveh:
I somehow missed this PR. It's way better than #6170

PR #6193, opened 2025-10-16
Fix PyCode constructor/replace

youknowone:
## Summary by CodeRabbit

* **New Features**
  * Code objects can now be instantiated directly via a public constructor.
  * Bytecode is validated and decoded when creating or replacing code objects, raising errors on invalid input.

* **Improvements**
  * Stronger validation of local variables and symbol tables to ensure consistency.
  * Function objects now serialize access to their code, improving safety when reading or updating __code__.

* **Bug Fixes**
  * Replacing code now re-parses and validates bytecode to prevent inconsistent state.

### DIFF
--- a/vm/src/builtins/code.rs
+++ b/vm/src/builtins/code.rs
@@ -15,8 +15,11 @@ use crate::{
 };
 use malachite_bigint::BigInt;
 use num_traits::Zero;
-use rustpython_compiler_core::OneIndexed;
-use rustpython_compiler_core::bytecode::PyCodeLocationInfoKind;
+use rustpython_compiler_core::{
+    OneIndexed,
+    bytecode::PyCodeLocationInfoKind,
+    marshal::{MarshalError, parse_instructions_from_bytes},
+};
 use std::{borrow::Borrow, fmt, ops::Deref};
 
 /// State for iterating through code address ranges
@@ -454,11 +457,8 @@ impl Constructor for PyCode {
 
         // Parse and validate bytecode from bytes
         let bytecode_bytes = args.co_code.as_bytes();
-        let instructions = parse_bytecode(bytecode_bytes).ok_or_else(|| {
-            vm.new_value_error(
-                "bytecode must have even length and contain only valid opcodes".to_owned(),
-            )
-        })?;
+        let instructions = parse_bytecode(bytecode_bytes)
+            .map_err(|e| vm.new_value_error(format!("invalid bytecode: {}", e)))?;
 
         // Convert constants
         let constants: Box<[Literal]> = args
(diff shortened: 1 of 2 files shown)


youknowone:
@ShaharNaveh could you review the changes?

vm/src/builtins/code.rs
```diff
@@ -866,6 +1036,26 @@ impl ToPyObject for bytecode::CodeObject {
     }
 }
 
+/// Validates and parses bytecode bytes into CodeUnit instructions.
+/// Returns None if bytecode is invalid (odd length or contains invalid opcodes).
+fn parse_bytecode(bytecode_bytes: &[u8]) -> Option<Box<[CodeUnit]>> {
```
ShaharNaveh: Wouldn't it be more appropriate to return
```rust
Result<Box<[CodeUnit]>, MarshalError>
```

vm/src/builtins/code.rs
```diff
@@ -866,6 +1036,26 @@ impl ToPyObject for bytecode::CodeObject {
...
+    // Bytecode must have even length (each instruction is 2 bytes)
+    if !bytecode_bytes.len().is_multiple_of(2) {
+        return None;
+    }
+
+    let mut instructions = Vec::with_capacity(bytecode_bytes.len() / 2);
+
+    for chunk in bytecode_bytes.chunks_exact(2) {
```
ShaharNaveh: We have this logic in multiple places. I think a future enhancement for having a newtype for 
```rust
pub struct CodeUnits(Box<[CodeUnit]>);
```

which implements, `TryFrom<&[u8]>` and returns a MarshalError.
but again, this is for a future PR.

ShaharNaveh (approved):
lgtm

vm/src/builtins/code.rs
youknowone: Is the concept marshal already existing in PyCode level? 
I thought marshal can serialize/deserialize code but it is not necessarily coupled to PyCode itself.

youknowone: hmm, marshal is in compiler and PyCode is in vm. so that feels weird but making sense

ShaharNaveh: Maybe marshal is not the appropriate error, I initially thought of marshal because it has `MarshalError::InvalidBytecode`

ShaharNaveh: I think that this function might belong in compiler-core anyways

youknowone: changed it to marshal error and added a comment that it is not a kind of necessary decision

youknowone:
@ShaharNaveh Thank you for your review 😄

PR #6199, opened 2025-10-20
Fix sqlite Connection initialization check

ever0de:
Add proper __init__ validation for sqlite3.Connection to ensure base class __init__ is called before using connection methods. This fixes the test_connection_constructor_call_check test case.

Changes:
- Modified Connection.py_new to detect subclassing
- For base Connection class, initialization happens immediately in py_new
- For subclassed Connection, db is initialized as None
- Added __init__ method that performs actual database initialization
- Updated _db_lock error message to match CPython: 'Base Connection.__init__ not called.'

## Summary by CodeRabbit

* **Bug Fixes**
  * Clearer error message when attempting to use an uninitialized SQLite connection, making initialization issues easier to diagnose.

* **Improvements**
  * Refined SQLite connection setup so base connections initialize immediately while derived connections defer initialization until explicitly initialized, ensuring correct configuration (path, timeouts, isolation) and preserving threading/locking behavior.

### DIFF
--- a/stdlib/src/sqlite.rs
+++ b/stdlib/src/sqlite.rs
@@ -74,8 +74,8 @@ mod _sqlite {
         },
         sliceable::{SaturatedSliceIter, SliceableSequenceOp},
         types::{
-            AsMapping, AsNumber, AsSequence, Callable, Comparable, Constructor, Hashable, IterNext,
-            Iterable, PyComparisonOp, SelfIter, Unconstructible,
+            AsMapping, AsNumber, AsSequence, Callable, Comparable, Constructor, Hashable,
+            Initializer, IterNext, Iterable, PyComparisonOp, SelfIter, Unconstructible,
         },
         utils::ToCString,
     };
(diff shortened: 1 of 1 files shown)


stdlib/src/sqlite.rs
```diff
@@ -873,25 +897,28 @@ mod _sqlite {
...
-                db: PyMutex::new(Some(db)),
-                detect_types: args.detect_types,
-                isolation_level: PyAtomicRef::from(args.isolation_level),
-                check_same_thread: args.check_same_thread,
-                thread_ident: std::thread::current().id(),
-                row_factory: PyAtomicRef::from(None),
-                text_factory: PyAtomicRef::from(text_factory),
-            })
+        #[pymethod]
+        fn __init__(&self, args: ConnectArgs, vm: &VirtualMachine) -> PyResult<()> {
```
youknowone: We also have an Initializer trait. It’s like the Constructor trait.
Would you apply this trait as well?

ever0de: Oh, I'd forgotten because it's been a while. Thank you

PR #6200, opened 2025-10-20
[boxvec] avoid 0-sized allocations

pieterh-oai:
## Description

Looks like there are occasionally cases where `code.max_stackdepth` is 0. That causes
`Frame::new` to call `BoxVec::new` with layout size 0, which calls the allocator with 0
size and non-zero-sized layout, which is undefined behavior.

The default Rust allocator returns a dangling pointer when called this way
(https://github.com/rust-lang/rust/blob/master/library/alloc/src/alloc.rs#L187 ), which
kind of sidesteps the issue. However, tikv-jemallocator has an `assume!` that causes it
to panic (in debug builds) instead, making it impossible for me to run debug builds
that combine RustPython and tikv-jemallocator
(https://github.com/tikv/jemallocator/blob/main/jemallocator/src/lib.rs#L96 )

## Test plan

```
cargo build ; cargo build --release
cargo clippy
cargo test --workspace --exclude rustpython_wasm
cd extra_tests ; pytest -v ; cd -

# back out change to confirm equivalent test results
git checkout HEAD^ common/src/boxvec.rs
cargo build ; cargo build --release
cargo clippy
cargo test --workspace --exclude rustpython_wasm
cd extra_tests ; pytest -v ; cd -
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved buffer initialization to properly construct empty buffers instead of leaving them uninitialized, ensuring correct behavior for edge cases.

### DIFF
--- a/common/src/boxvec.rs
+++ b/common/src/boxvec.rs
@@ -39,6 +39,12 @@ macro_rules! panic_oob {
 
 impl<T> BoxVec<T> {
     pub fn new(n: usize) -> Self {
+        if n == 0 {
+            return Self {
+                xs: Box::new([]),
+                len: 0,
+            };
+        }
         Self {
             xs: Box::new_uninit_slice(n),
             len: 0,


pieterh-oai:
@coderabbitai generate docstrings

youknowone:
Thank you for contributing! I tried to understand what's happening. The calling chain is:

BoxVec::new ->
Box::new_uninit_slice ->
RawVec::with_capacity ->
RawVecInner::with_capacity ->
RawVecInner::try_allocate_in

`RawVecInner::try_allocate_in` checks `layout.size() == 0` in the middle and skips the allocation.

Am I missing something? I didn't get how it hits the alloc part yet. How did you catch there are allocations?

pieterh-oai:
Oooh you're right.

Here's what happened:

* I added a crate dependency to add RustPython to embed into an existing project.
* The most recent version on crates.io is "0.4.0" (Aug 2024), which predates https://github.com/RustPython/RustPython/commit/c883f0ad8af76a981eaca511fa2c3e334117dbad
* The original code does make a 0-sized request to the allocator directly from `BoxVec::new`
* I didn't realize the crates version was so old, so I ended up modifying the code as-is.
* In parallel, I switched back to the system allocator for now to avoid having to vendor a temporary copy of RustPython.

Can close this unless you feel there's some code cleanliness benefit.

youknowone:
Thank you for explanation. To avoid duplicated logic, I will close it.
Sorry to make confusion for too old release version

PR #6207, opened 2025-10-22
ruff for `Expr` parsing

ShaharNaveh:
Reverts RustPython/RustPython#6206


Lib/test/test_future_stmt/test_future.py
```diff
@@ -198,6 +267,7 @@ def _exec_future(self, code):
         )
         return scope
 
+    @unittest.expectedFailure # TODO: RUSTPYTHON; 'a,' != '(a,)'
```
ShaharNaveh, line 270: @youknowone should I open a feature request to ruff team about this? or should we just maintain our own fork of it? I am less in favor of maintaining a fork as it will require us:
1. maintain a fork
2. prevents us from publishing a new version to crates.io (unless we push the fork to crates.io)

youknowone: make a fork, write a feature but ask to merge it before using the fork. split the fork later when rejected. nothing to loose.
because we (and most of open source projects) always don't have enough engineering power for the project, upstreaming is the better choice when possible.

ShaharNaveh: Let's hope that they are fine with this https://github.com/astral-sh/ruff/pull/21041

Issue #6212, opened 2025-10-24
Panic occurs when formatting with separator and some format specifier


yt2b:
```
>>>>> f"{123456:,g}"

thread 'main' panicked at common/src/format.rs:429:18:
Separators only valid for numbers!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
>>>>> f"{123456:,e}"

thread 'main' panicked at common/src/format.rs:429:18:
Separators only valid for numbers!
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

But in Python 3.13.5, output is below.

```
>>> f"{123456:,g}"
'123,456'
>>> f"{123456:,e}"
'1.234560e+05'
```

I'll fix this bug.

PR #6213, opened 2025-10-24
Panic occurs when formatting with separator and some format specifier

yt2b:
## Summary
#6212
I've fixed get_separator_interval.
And I've added extra tests.

```
>>>>> f"{123456:,g}"
'123,456'
>>>>> f"{123456:,e}"
'1.234560e+05'
>>>>> f"{123456:,%}"
'12,345,600.000000%'
```

## Summary by CodeRabbit

* **New Features**
  * Extended thousands-separator support to decimal, fixed-point, general/scientific, and percentage numeric formats.

* **Tests**
  * Added tests verifying thousands-separator formatting across floating-point and integer specifiers, scientific and percentage formats, and signed/uppercase variations.

### DIFF
--- a/common/src/format.rs
+++ b/common/src/format.rs
@@ -426,7 +426,6 @@ impl FormatSpec {
             Some(FormatType::Binary | FormatType::Octal | FormatType::Hex(_)) => 4,
             Some(
                 FormatType::Decimal
-                | FormatType::Number(_)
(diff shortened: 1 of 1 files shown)


ShaharNaveh:
@yt2b this looks very good, I think that it's worth adding those tests upstream (to cpython). I find it odd that they don't have a test for it

PR #6215, opened 2025-10-25
Enable TestMethodsMutating

kralka:
Make Dict::lookup check if the dictionary has been modified. This is a workaround for a possible segfault. See bpo-46615 for the original discussion in CPython.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved dictionary iteration robustness: operations that modify a collection during iteration now produce a clear runtime error ("set changed size during iteration") instead of risking crashes, preventing silent corruption and improving stability.

### DIFF
--- a/vm/src/dict_inner.rs
+++ b/vm/src/dict_inner.rs
@@ -582,12 +582,13 @@ impl<T: Clone> Dict<T> {
                 });
                 loop {
                     let index_index = idxs.next();
-                    let index_entry_ptr = inner.indices.get(index_index);
-                    if index_entry_ptr.is_none() {
-                        // Dictionary was modified under our hands, see TestMethodsMutating.
-                        continue 'outer;
-                    }
-                    let index_entry = *index_entry_ptr.unwrap();
+                    let index_entry = match inner.indices.get(index_index) {
+                        None => {
+                            // Dictionary was modified under our hands, see TestMethodsMutating.
+                            continue 'outer;
+                        }
+                        Some(v) => *v,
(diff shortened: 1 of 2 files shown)


kralka:
I will investigate later this weekend. Locally (Linux -- rust docker container) the test passes: `target/release/rustpython -m test -j 1 -u all test_set` seems to work for me. Same `RUSTPYTHONPATH=Lib cargo run --release Lib/test/test_set.py` gives OK (skipped=63, expected failures=15).

youknowone:
Thank you!

vm/src/dict_inner.rs
```diff
@@ -582,10 +582,12 @@ impl<T: Clone> Dict<T> {
                 });
                 loop {
                     let index_index = idxs.next();
-                    let index_entry = *unsafe {
-                        // Safety: index_index is generated
-                        inner.indices.get_unchecked(index_index)
-                    };
+                    let index_entry_ptr = inner.indices.get(index_index);
```
ShaharNaveh: Would something like this work?

```rust
let index_entry = match inner.indices.get(index_index) {
  None => continue 'outer,
  Some(v) => v,
};
```

kralka: Yes, done, thank you.

kralka:
Ok, only the Ubuntu is running the platform independent tests (based on the workflow file). And the test is flaky so that it crashed always the same in a single run on my computer. I let it run 1000 times yesterday.

PR #6217, opened 2025-10-26
Update ssl.py from CPython 3.13.9

youknowone:


youknowone:
pip goes to be broken

youknowone:
@ShaharNaveh The *.pem files are removed here

Issue #5726, opened 2025-04-21
PyNumber protocol does not check for implementations in base classes


arihant2math:
`PySequence` and others do this, but `PyNumber` does not leading to #5711 and ctypes data multiplication not working.

arihant2math:
@youknowone I'm curious as to how I could attempt a fix for this. It seems as though PyNumber has been defined slightly differently than the other protocols, and I think this is intentional.

youknowone:
It's been so long time ago I worked on slots. I forgot the most of details.

CPython basically inherits slots not to require subtypes to crawl up the entire mro. But we don't have it.
We check base classes to avoid this limitation. About PyNumber, checking base classes made another problem like referencing PyBaseObject's PyNumber, which shouldn't happen.

If PyNumber slots are fairly well isolated from others, we will be able to implement slot inheritance only for PyNumber

youknowone:
Long rally including #6622 solved hopefully all the slot inheritance issues

PR #6222, opened 2025-10-27
Resolve number slots via MRO in PyNumber and operator, ensure inherit…

YashSuthar983:
…ed and dynamically added methods are found.

Use class().mro_find_map() to mimic the same behaviour as CPython.

issue: #5726

test Cases

**dynamic_patch_test**
```python
# Expected behavior: After patching A.__mul__ at runtime, B() * 2 should succeed.

class A:
    pass

class B(A):
    pass

b = B()
def mul(self, other):
    return ("mul", type(self).__name__, other)

A.__mul__ = mul

res = b * 2
assert res == ("mul", "B", 2)
print("dynamic_patch OK")

```

**Cpython**
```bash
Bravo:~$ python3 ../number_mro_dynamic_patch.py
dynamic_patch OK
```

**RustPython**
```bash
Bravo:~/contri/RustPython$ cargo run --quiet -- tests/number_mro_dynamic_patch.py
  File "tests/number_mro_dynamic_patch.py", line 20, in <module>
    res = b * 2
TypeError: '*' not supported between instances of 'B' and 'int'
Bravo:~/contri/RustPython$ 

```

## Summary by CodeRabbit

* **Bug Fixes**
  * Numeric operations and type conversions now resolve methods by searching the full method resolution order, ensuring inherited or overridden numeric slots are respected.
  * Binary and ternary operation dispatch correctly invokes overridden methods from parent classes when present, preserving previous behavior and error messages.

### DIFF
--- a/vm/src/protocol/number.rs
+++ b/vm/src/protocol/number.rs
@@ -444,6 +444,11 @@ impl<'a> PyNumber<'a> {
     // PyNumber_Check
     pub fn check(obj: &PyObject) -> bool {
         let cls = &obj.class();
+        // TODO: when we finally have a proper slot inheritance, mro_find_map can be removed
+        //    methods.int.load().is_some()
+        //    || methods.index.load().is_some()
+        //    || methods.float.load().is_some()
+        //    || obj.downcastable::<PyComplex>()
         let has_number = cls
             .mro_find_map(|x| {
                 let methods = &x.slots.as_number;


YashSuthar983:
@youknowone @arihant2math  please take a look whenever you get a chance.

arihant2math:
Oh nice, you might consider remove this patch:
https://github.com/RustPython/RustPython/blob/9a2792a44b6362554b7bd76b3f91ee88362299df/Lib/ctypes/__init__.py#L304 (if this fixes that).

vm/src/protocol/number.rs
```diff
@@ -443,117 +443,135 @@ impl<'a> PyNumber<'a> {
 
     // PyNumber_Check
     pub fn check(obj: &PyObject) -> bool {
-        let methods = &obj.class().slots.as_number;
-        methods.int.load().is_some()
-            || methods.index.load().is_some()
-            || methods.float.load().is_some()
-            || obj.downcastable::<PyComplex>()
+        let cls = &obj.class();
```
youknowone, line 446:
```suggestion
        let cls = &obj.class();
        // TODO: when we finally have a proper slot inheritance, mro_find_map can be removed
        //    methods.int.load().is_some()
        //    || methods.index.load().is_some()
        //    || methods.float.load().is_some()
        //    || obj.downcastable::<PyComplex>()
```

PR #6230, opened 2025-10-31
Implement property.__name__ attribute

ever0de:
Add getter and setter for the __name__ attribute on property objects. The getter returns the explicitly set name if available, otherwise falls back to the getter function's __name__. Raises AttributeError if no name is available, matching CPython 3.13 behavior.

The implementation handles edge cases:
- Returns None when explicitly set to None
- Propagates non-AttributeError exceptions from getter's __getattr__
- Raises property-specific AttributeError when getter lacks __name__

This fix enables test_property_name in test_property.py to pass.

## Summary by CodeRabbit

* **New Features**
  * Property objects now expose a readable and writable __name__ attribute.
  * When not explicitly set, __name__ is derived from the property's getter.

* **Bug Fixes**
  * Improved error handling around property names: missing or non-existent names are treated consistently, while unexpected errors are surfaced for clearer diagnostics.

### DIFF
--- a/vm/src/builtins/property.rs
+++ b/vm/src/builtins/property.rs
@@ -74,22 +74,22 @@ impl PyProperty {
             return Ok(Some(name.clone()));
         }
 
-        // Otherwise try to get __name__ from getter
-        if let Some(getter) = self.getter.read().as_ref() {
-            match getter.get_attr("__name__", vm) {
-                Ok(name) => Ok(Some(name)),
-                Err(e) => {
-                    // If it's an AttributeError from the getter, return None
-                    // Otherwise, propagate the original exception (e.g., RuntimeError)
-                    if e.class().is(vm.ctx.exceptions.attribute_error) {
-                        Ok(None)
-                    } else {
-                        Err(e)
-                    }
+        let getter = self.getter.read();
+        let Some(getter) = getter.as_ref() else {
+            return Ok(None);
+        };
+
+        match getter.get_attr("__name__", vm) {
+            Ok(name) => Ok(Some(name)),
+            Err(e) => {
+                // If it's an AttributeError from the getter, return None
+                // Otherwise, propagate the original exception (e.g., RuntimeError)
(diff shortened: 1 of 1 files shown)


youknowone:
Thank you! I thought we already had name handling on pyproperty, but it was `PyProperty::get_property_name`, which was not actually used for `__name__`. Could you check if it is also related to `__name__` impl?

vm/src/builtins/property.rs
```diff
@@ -67,20 +67,30 @@ impl GetDescriptor for PyProperty {
...
+                    // Otherwise, propagate the original exception (e.g., RuntimeError)
+                    if e.class().is(vm.ctx.exceptions.attribute_error) {
+                        Ok(None)
+                    } else {
+                        Err(e)
+                    }
+                }
+            }
+        } else {
+            Ok(None)
```
youknowone:
```suggestion
        let Some(getter) = self.getter.read().as_ref() {
            return Ok(None);
        };
        match getter.get_attr("__name__", vm) {
            Ok(name) => Ok(Some(name)),
            Err(e) => {
                // If it's an AttributeError from the getter, return None
                // Otherwise, propagate the original exception (e.g., RuntimeError)
                if e.class().is(vm.ctx.exceptions.attribute_error) {
                    Ok(None)
                } else {
                    Err(e)
                }
            }
```

ever0de:
I've applied the review. thanks!

vm/src/builtins/property.rs
```diff
@@ -67,20 +67,30 @@ impl GetDescriptor for PyProperty {
...
+            return Ok(Some(name.clone()));
         }
 
-        // Otherwise try to get __name__ from getter
-        if let Some(getter) = self.getter.read().as_ref()
-            && let Ok(name) = getter.get_attr("__name__", vm)
-        {
-            return Some(name);
-        }
+        let getter = self.getter.read();
```
ever0de, line 77: Added an extra variable because a temporary value issue prevented a single-line implementation.

PR #6231, opened 2025-10-31
Move typeid into vtable

coolreader18:
Now that `TypeId::of()` is const, we can move the typeid into `PyObjVTable`

## Summary by CodeRabbit

* **Chores**
  * Updated minimum Rust toolchain requirement to 1.91.0. Users compiling from source must upgrade their Rust toolchain.

* **Refactor**
  * Internal type-identification implementation switched from callable lookups to associated constants. No user-visible behavior changes, only internal API shape.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -131,7 +131,7 @@ members = [
 version = "0.4.0"
 authors = ["RustPython Team"]
 edition = "2024"
-rust-version = "1.89.0"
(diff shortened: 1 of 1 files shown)


youknowone:
No idea how they are related, but re-running CI doesn't fix it.

youknowone:
const PAYLOAD_TYPE_ID evaluates to different values across compilation units

PR #6232, opened 2025-11-01
Remove user defined docstrings for `builtins.str`

ShaharNaveh:
@youknowone would you object to move https://github.com/RustPython/__doc__.git to this project? (in it's own crate)

## Summary by CodeRabbit

* **Chores**
  * Removed internal documentation comments from code with no functional impact or changes to user-facing features.

### DIFF
--- a/vm/src/builtins/str.rs
+++ b/vm/src/builtins/str.rs
@@ -890,10 +890,6 @@ impl PyStr {
         )
     }
 
-    /// Return a str with the given prefix string removed if present.
-    ///
-    /// If the string starts with the prefix string, return string[len(prefix):]
-    /// Otherwise, return a copy of the original string.
     #[pymethod]
     fn removeprefix(&self, pref: PyStrRef) -> Wtf8Buf {
         self.as_wtf8()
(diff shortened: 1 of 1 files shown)


youknowone (approved):

youknowone:
@ShaharNaveh I forgot what was the major reason to split it out. Probably the license and workflow part. To integrate it again, guidelines:
- Please check the generate script is still working well.
- The license must be aligned to CPython
- We'd better to generate them and commit from CI. (manual trigger when upgrading version)

ShaharNaveh:
> @ShaharNaveh I forgot what was the major reason to split it out. Probably the license and workflow part. To integrate it again, guidelines:
> 
> * Please check the generate script is still working well.

I intend to refractor it, either way I'll make it will work.

> * The license must be aligned to CPYTHON

Sure, will copy CPython license to `crates/rustpython_doc_db/LICENSE `

> * We'd better to generate them and commit from CI. (manual trigger when upgrading version)

I thought of having the generate script to have a `--check` flag that will exit with non-zero on any diff, we will check it on the CI.

youknowone:
> I thought of having the generate script to have a --check flag that will exit with non-zero on any diff, we will check it on the CI.

The current script is depends on python runtime.  Each platform generate their own set of document. To cover the entire platform, the results from different platforms must be merged.

ShaharNaveh:
> > I thought of having the generate script to have a --check flag that will exit with non-zero on any diff, we will check it on the CI.
> 
> The current script is depends on python runtime. Each platform generate their own set of document. To cover the entire platform, the results from different platforms must be merged.

Make sense, I don't expect `nt` to be available on Linux. But I don't think that there's a difference between docstring of same object on different platforms, right? i.e, `str.count` is the same on Windows, Linux, MacOS. correct?

youknowone:
Yes, they are.

PR #6234, opened 2025-11-03
Move `__doc__` crate to `crates/doc`

ShaharNaveh:
Moved https://github.com/RustPython/__doc__.git to here, where it's going to be easier to maintain.

I did do some modifications to the original crate, one of them is to not use `once_cell::sync::Lazy` , that's because it's redundant (I think) as it's accessed via the vm & stdlib at compile time, so it's never going to be "lazily evaluated".

Second change I did was to use https://github.com/rust-phf/rust-phf because the rust's HashMap started to choke when inserting ~9K entries

## Summary by CodeRabbit

* **New Features**
  * Centralized documentation database enabling consistent in-app doc lookups.
  * Automated documentation generation workflow to keep standard library and builtins docs up to date.

* **Chores**
  * Workspace reorganization and dependency adjustments to support the doc crate.
  * CI timeout increases for improved test stability.

* **Documentation**
  * Added comprehensive license and ignore rules for generated documentation.

* **Tests**
  * Added a basic retrieval test to validate the doc database.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -113,7 +113,7 @@ jobs:
       RUST_BACKTRACE: full
     name: Run rust tests
     runs-on:  ${{ matrix.os }}
-    timeout-minutes: ${{ contains(matrix.os, 'windows') && 45 || 30 }}
+    timeout-minutes: ${{ contains(matrix.os, 'windows') && 45 || 35 }}
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
@@ -239,7 +239,7 @@ jobs:
       RUST_BACKTRACE: full
     name: Run snippets and cpython tests
     runs-on:  ${{ matrix.os }}
-    timeout-minutes: ${{ contains(matrix.os, 'windows') && 45 || 30 }}
+    timeout-minutes: ${{ contains(matrix.os, 'windows') && 45 || 35 }}
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]


.github/workflows/update-doc-db.yml
```diff
@@ -0,0 +1,70 @@
...
+        run: echo "diff=$(git diff --name-only)" >> $GITHUB_OUTPUT
+
+      - uses: actions/upload-artifact@330a01c490aca151604b8cf639adc76d48f6c5d4 # v5.0.0
+        with:
+          name: doc-db-${{ env.PYTHON_VERSION }}-${{ matrix.os }}
+          path: ${{ steps.diff.outputs.diff }}
+          if-no-files-found: error
+          retention-days: 7
+          overwrite: true
+  pr:
```
ShaharNaveh: I ain't sure if we want to open a PR, or just push the generated artifacts

youknowone: we need to update it only when target Python version is changed.

ShaharNaveh: Yeah, so probably not worth the effort in automating a PR

crates/rustpython_doc_db/src/lib.rs
```diff
@@ -0,0 +1,19 @@
+#[cfg(windows)]
+include!("./win32.inc.rs");
+
+#[cfg(any(target_os = "linux", target_os = "android"))]
+include!("./linux.inc.rs");
+
+#[cfg(any(target_os = "macos", target_os = "ios"))]
```
youknowone:
```suggestion
#[cfg(target_vendor = "apple")]
```

youknowone:
Looks great, I have one more question. Is it possible to merge 3 inc files into a single file? They have subsets but can be merged into a single big file. Then we don't need to worry about missing architecture when building

ShaharNaveh:
> Looks great, I have one more question. Is it possible to merge 3 inc files into a single file? They have subsets but can be merged into a single big file. Then we don't need to worry about missing architecture when building

Yeah sure, we can do that.

Cargo.toml
```diff
@@ -156,13 +161,14 @@ rustpython-pylib = { path = "pylib", version = "0.4.0" }
 rustpython-stdlib = { path = "stdlib", default-features = false, version = "0.4.0" }
 rustpython-sre_engine = { path = "vm/sre_engine", version = "0.4.0" }
 rustpython-wtf8 = { path = "wtf8", version = "0.4.0" }
-rustpython-doc = { git = "https://github.com/RustPython/__doc__", tag = "0.3.0", version = "0.3.0" }
+rustpython_doc_db = { path = "crates/rustpython_doc_db", version = "0.4.0" }
```
youknowone:
```suggestion
rustpython-doc_db = { path = "crates/rustpython_doc_db", version = "0.4.0" }
```
`-` is a namespace split, `_` is a (sort of) whitespace split

youknowone (changes requested):
Thank you so much!

@coolreader18 you have to review crate naming policy not to block these changes

crates/rustpython_doc_db/src/lib.rs
ShaharNaveh: this is outdated. now everything is merged into a single file

Cargo.toml
ShaharNaveh: I think this is worth discussing more. I haven't seen a single project that has both `-` and `_` in it's name. I think we should stick to one separator and use it (I'm in favor of `_`)

youknowone: We already have that names for every crates.
I don't think mixing `rustpython-` and `rustpython_` for different crates gives any value.

Actually I was missing the fundamental question about this change.
We'd better to keep using the previous crate name `rustpython-doc`. Moving crate doesn't mean renaming(creating a new) crate.

ShaharNaveh: > We already have that names for every crates. I don't think mixing `rustpython-` and `rustpython_` for different crates gives any value.

think we should choose one sperator and stick to it, as long as it's consistent.

> Actually I was missing the fundamental question about this change. We'd better to keep using the previous crate name `rustpython-doc`. Moving crate doesn't mean renaming(creating a new) crate.

Yeah, ofc

youknowone: > think we should choose one sperator and stick to it, as long as it's consistent.

I understand that there’s some validity to that point.
However, it’s not fair to justify it simply by calling it “consistent,” because the fact that all RustPython crates share the rustpython- prefix is already a form of consistency in itself.
It’s just a different kind of consistency.
If these two forms of consistency conflict, I would reject introducing a new crate with a different prefix.

ShaharNaveh:
@youknowone I ain't sure about this CI failure. I think it's related to https://github.com/RustPython/RustPython/pull/6237

PR #6240, opened 2025-11-08
Base implementation of `_tokenize` module

ShaharNaveh:
## Summary by CodeRabbit

* **New Features**
  * New Python tokenizer module: Tokenizer iterator that yields token tuples (type, text, start, end, line), accepts a readline callable, optional encoding, and an extra-token mode; emits EOF tokens with position metadata for streaming/tokenization.

* **Chores**
  * Internal dependency declarations updated to support the new tokenizer module.

### DIFF
--- a/crates/stdlib/src/_tokenize.rs
+++ b/crates/stdlib/src/_tokenize.rs
@@ -58,7 +58,7 @@ mod _tokenize {
                     vm.state
                         .codec_registry
                         .decode_text(bytes.into(), encoding, None, vm)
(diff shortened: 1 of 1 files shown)


youknowone:
I missed this, sorry

ShaharNaveh:
> I missed this, sorry

All good. This PR wasn't great anyway. I'm thinking it might be worth to implement the tokenizer by hand instead of relying on ruff as there's multiple issues with it. I'll try to resolve the conflicts and how the newer versions of ruff behave

PR #6252, opened 2025-11-13
Add BaseException.add_note() and __notes__ attribute support

ever0de:
## Summary by CodeRabbit

* **New Features**
  * Exceptions can now have notes attached, letting users add contextual messages to error objects for clearer debugging.
  * When adding a note, the runtime will create any needed internal storage automatically and append the note to the exception's notes list.

### DIFF
--- a/vm/src/exceptions.rs
+++ b/vm/src/exceptions.rs
@@ -659,13 +659,15 @@ impl PyRef<PyBaseException> {
             .dict()
             .ok_or_else(|| vm.new_attribute_error("Exception object has no __dict__"))?;
 
-        let notes = dict
-            .get_item("__notes__", vm)
-            .or_else(|_| {
-                let new_notes = vm.ctx.new_list(vec![]);
-                dict.set_item("__notes__", new_notes.clone().into(), vm)?;
-                Ok(new_notes.into())
-            })?
+        let notes = if let Ok(notes) = dict.get_item("__notes__", vm) {
+            notes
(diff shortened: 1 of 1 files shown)


vm/src/exceptions.rs
```diff
@@ -652,6 +652,27 @@ impl PyRef<PyBaseException> {
...
+
+        let notes = dict
+            .get_item("__notes__", vm)
+            .or_else(|_| {
+                let new_notes = vm.ctx.new_list(vec![]);
+                dict.set_item("__notes__", new_notes.clone().into(), vm)?;
+                Ok(new_notes.into())
+            })?
+            .downcast::<PyList>()
+            .map_err(|_| vm.new_type_error("__notes__ must be a list"))?;
```
youknowone:
```suggestion
        let notes = if let Ok(notes) = dict.get_item("__notes__", vm) {
            notes
        } else {
            let new_notes = vm.ctx.new_list(vec![]);
            dict.set_item("__notes__", new_notes.clone().into(), vm)?;
            new_notes.into()
        };
        let notes = notes
            .downcast::<PyList>()
            .map_err(|_| vm.new_type_error("__notes__ must be a list"))?;
```
I’m a bit concerned about having side-effect code inside or_else.

youknowone (changes requested):
left a suggestion about style

youknowone (approved):
Thank you!

Issue #6214, opened 2025-10-24
Organize our rust code into `crates/` directory


ShaharNaveh:
All of our current crates are scattered in the top level of the repository, it's not very intuitive for newcomers to look at the code and know that `common` directory is for shared rust code, and that `wtf8` is also a crate for example.

I propose that we create a new directory called `crates` where we will have the following:

- [ ] `common` -> `crates/common` #6256 
- [ ] `compiler/core` -> `crates/compiler-core` #6258
- [ ] `compiler/codegen` -> `crates/codegen` #6260 
- [ ] `compiler/literal` -> `crates/literal` #6259
- [ ] `compiler/source` -> `crates/compiler-source` #6261
- [ ] `compiler/src` -> `crates/compiler` #6270
- [ ] `derive` -> `crates/derive` #6264
- [ ] `derive-impl` -> `crates/derive-impl` #6263
- [ ] `jit` -> `crates/jit` #6262
- [ ] `pylib` -> `crates/pylib` #6225 
- [ ] `stdlib` -> `crates/stdlib` #6268
- [ ] `vm/sre_engine` -> `crates/sre_engine` #6265
- [ ] `vm` -> `crates/vm` #6269
- [ ] `wasm/lib` -> `crates/wasm` #6280
- [ ] `wtf8` -> `crates/wtf8` #6257 
- [ ] `src` -> `crates/rustpyton`

I think it be better if we do those changes with many smaller PRs as not all of them would require changes to the CI, or any other changes. while minimizing the merge conflicts frequencies while we are in this transition.

For project examples that uses this workspace structure you can see:

- https://github.com/astral-sh/ruff
- https://github.com/bevyengine/bevy

ShaharNaveh:
@youknowone can you please take a look? feel free to close it if you think this change is not worth doing

youknowone:
I am personally not a fan of that structure when a project is a single build system based.
For RustPython, we only use `cargo`. Also it only has `Lib` to contains Python source code, and every other directories are basically rust crates.

But I am open to follow the convention if this is a well-known good practice. Do you have a good reference to learn why people decide to that or a good discussion what's the benefits of it?

cc @coolreader18

ShaharNaveh:
> But I am open to follow the convention if this is a well-known good practice.

It's not an official convention, or a "good practice" or a "bad practice", but it's something that I think would be beneficial for us.

> Do you have a good reference to learn why people decide to that or a good discussion what's the benefits of it?

[...]

Issue #6204, opened 2025-10-21
Add `cargo fmt --all` and commit to github action


youknowone:
Thanks to @YashSuthar983, we now have auto-format after workflows.

I'd like to check if auto-formatting before workflows is possible

## Summary

This is not a necessary task to be asked to human contributors.
If CI can automatically run it, we can rely on it.

YashSuthar983:
@youknowone i have doubt , as cargo fmt --all formats the code then are we going to make an commit from CI ?

or we can use cargo fmt --all --check ?

youknowone:
@YashSuthar983 Make a commit from CI. We already have `cargo fmt --check`, and the solution is usually running `cargo fmt --all` and commit the changes. It can be trivially autumated.

YashSuthar983:
So what i am thinking is that whenever user open a PR from fork , our CI run cargo fmt --all then if any changes this wil commit to pr of user right?
I don't think it is allowed by github or we need somthing like bot that use PAT of user for this.

youknowone:
I didn't try that yet, but some github actions look like supporting it.

- https://github.com/marketplace/actions/add-commit
- https://github.com/marketplace/actions/git-auto-commit
- https://github.com/marketplace/actions/add-commit-and-push

Not sure which one is best fit for this purpose though.

I can configure some variables on actions if it is required.

YashSuthar983:
@youknowone We can do one thing, whenever a user pushes to their fork, the CI pipeline run cargo fmt --all and if any formatting changes are needed, it should commit in their fork itself.
i tested this in [repo](https://github.com/YashSuthar983/test_CI).  see pull request 
if this satisfy need let me know ?

YashSuthar983:
@youknowone   I’d like to know your thoughts on the  solution, as it’s a hassle to remind everyone to run cargo fmt --all each time.

youknowone:
@YashSuthar983 I checked the repository and it is looking great! Could you apply the same workflow to RustPython too?

YashSuthar983:
@youknowone  i got confused as we can do two thing to be efficient and avoid conflicts .
1. we run auto format at the end of all workflow .
2. we make separate workflow that you can trigger.

youknowone:
I expected auto-format before all other workflow, but auto-format at end of workflow also sounds reasonable.

manual triggering is not a desired one. (better than no auto-format though)

youknowone:
@YashSuthar983 Thank you so much! it looks working great now

[...]

PR #6259, opened 2025-11-14
Move `compiler/literal` -> `crates/literal`

ShaharNaveh:
ref: https://github.com/RustPython/RustPython/issues/6214

## Summary by CodeRabbit

* **Chores**
  * Reorganized internal project structure to streamline workspace configuration.


fanninpm:
@coderabbitai review

ShaharNaveh:
the PR auto-format is reporting issue although everything passes. I ain't so sure what's the added value of it.

fanninpm:
> the PR auto-format is reporting issue although everything passes. I ain't so sure what's the added value of it.

One of the tests on Windows was failing, so I re-ran it. Ideally, the auto-format check should run _first_ (before all the other checks even get a chance to run), so that's something for @YashSuthar983 to keep in mind.

ShaharNaveh:
> > the PR auto-format is reporting issue although everything passes. I ain't so sure what's the added value of it.
> 
> One of the tests on Windows was failing, so I re-ran it.

TY:)
I was wondering how it's marked as OK now. I don't have permissions to re-run a single workflow:/

> Ideally, the auto-format check should run _first_ (before all the other checks even get a chance to run), so that's something for @YashSuthar983 to keep in mind.

What's the added value of formatting things via the CI? why not enforce it like we did it until now?

YashSuthar983:
@fanninpm as I discussed with @youknowone ,we decided that auto formatting should run after all checks passes as if we do before we would have a auto format commit with skip ci and there will be some extra commits too if ci fails after auto format.

So in current implementation auto format waits for all CI to pass and there is timeout , there it every 30 checks if ci passes or not .

youknowone:
> we decided that auto formatting should run after all checks passes 

Ideally formatting before CI will be better. Because CI passes after formatting is more complete. If no choice, this is better than nothing

> why not enforce it like we did it until now?

Auto-formatting is a higher level policy than enforce by policy. User always must format code, which is automatic task run by a command. Then why it shouldn't be automatic? That's one of the major purpose of programming and software.

youknowone:
Please leave more discussion to #6204

youknowone (approved):

ShaharNaveh:
> > we decided that auto formatting should run after all checks passes
> 
> Ideally formatting before CI will be better. Because CI passes after formatting is more complete. If no choice, this is better than nothing

Agree

> > why not enforce it like we did it until now?
> 
> Auto-formatting is a higher level policy than enforce by policy. User always must format code, which is automatic task run by a command. Then why it shouldn't be automatic? That's one of the major purpose of programming and software.

In some way I agree, but I think it makes things more complicated than it needs to be. IMO CI needs to validate things. There's pre-commit hooks to make it easier for developers

PR #6275, opened 2025-11-15
Auto-format before checks

### LINKED ISSUE #6272: CI AutoFormat fails when CI is under load.
The current CI auto-format implementation waits for all jobs to be finished before proceeding, and that wait time is 1 hour.
In the case where the auto-format job gets scheduled before the rest of the other jobs, it will start it's countdown immediately, but there's a chance for the other jobs to still be in queue because some other CI jobs are running for a pushed commit/other PR, which might take more than 1 hour.

It looks somewhat like this:

```mermaid
gantt
    title CI Timeline
    dateformat HH:mm
    axisFormat %H:%M
    PR Auto Format: 00:15, 1h
    Task 1: 00:00, 45m
    Task 2: 00:45, 45m
    Fails if still waiting: milestone, m1, 01:15, 0m
```

---

There are several approaches that I think could work to solve this:
1. increase waiting timeout (or remove the timeout).
2. Make sure that the PR auto format is triggered before any CI jobs.
3. Make the auto-formag opt-in format by triggering it via a comment on the PR, just like we have `take` assigning people to issues (https://github.com/RustPython/RustPython/blob/cc2e84b9fc0af690f3281cd4b02e219b4752bed7/.github/workflows/comment-commands.yml)

I'm leaning towards having the ability to trigger it on-demand. And if not, then move it to the start of the CI as having a job just waiting for 99% of it's time is a waste of a runner that could be used to speed up the CI (at the time writing this we have 6 concurrent auto-pr format running, and 16 CI jobs in queue)

ShaharNaveh:
closes #6272 

## Summary by CodeRabbit

* **Chores**
  * Consolidated PR formatting into CI: new formatting gate runs on PRs, auto-formats code, comments when changes are made, and blocks downstream jobs until resolved.
  * Removed the prior standalone PR auto-format workflow; formatting is now handled in the main pipeline.
  * Job gating now respects cancellations and explicit skip labels for more predictable checks.

* **Style**
  * Minor non-functional code formatting adjustments with no behavior changes.

### DIFF
--- a/crates/vm/src/stdlib/ast/other.rs
+++ b/crates/vm/src/stdlib/ast/other.rs
@@ -14,10 +14,13 @@ impl Node for ruff::ConversionFlag {
     ) -> PyResult<Self> {
         i32::try_from_object(vm, object)?
             .to_u32()
-            .and_then(bytecode::ConversionFlag::from_op_arg)
+            .and_then(
+                bytecode::ConversionFlag::from_op_arg)
             .map(|flag| match flag {
-                bytecode::ConversionFlag::None => Self::None,
-                bytecode::ConversionFlag::Str => Self::Str,
+                bytecode::ConversionFlag::None =>
+                 Self::None,
+                bytecode::ConversionFlag::Str => 
+                Self::Str,
                 bytecode::ConversionFlag::Ascii => Self::Ascii,
                 bytecode::ConversionFlag::Repr => Self::Repr,
             })


youknowone:
@ShaharNaveh could you test this workflow with a mal-formatted commit?

ShaharNaveh:
The main issue I encounter ATM is the fact that github actions prevents you from doing recursive calls to the workflow. I can't get the checks to trigger after the auto-format commit

youknowone:
#6295 replaced it. Thank you for the work!

Issue #6273, opened 2025-11-15
RustPython CLI cannot run .pyc file


youknowone:
## Summary

When running CLI with an argument, Pyhton check if the input source path is `.pyc` and then use special route.
While RustPython does not.

PR #6276, opened 2025-11-15
Add __main__.__cached__

### LINKED ISSUE #6274: __main__ module doesn't set __cached__

youknowone:
fix #6274 

I also re-organize vm.run_script inner functions to prepare #6273

### DIFF
--- a/crates/vm/src/vm/compile.rs
+++ b/crates/vm/src/vm/compile.rs
@@ -69,6 +69,7 @@ impl VirtualMachine {
                 self.ctx.new_str(path).into(),
                 self,
             )?;
+            module_dict.set_item(identifier!(self, __cached__), self.ctx.none(), self)?;
         }
 
         // Consider to use enum to distinguish `path`
--- a/crates/vm/src/vm/context.rs
+++ b/crates/vm/src/vm/context.rs
@@ -98,6 +98,7 @@ declare_const_name! {
     __build_class__,
     __builtins__,
     __bytes__,
+    __cached__,
     __call__,
     __ceil__,
     __cformat__,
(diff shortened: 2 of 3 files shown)


crates/vm/src/vm/compile.rs
```diff
@@ -44,19 +47,57 @@ impl VirtualMachine {
...
         }
 
-        match std::fs::read_to_string(path) {
-            Ok(source) => {
-                self.run_code_string(scope, &source, path.to_owned())?;
+        self.run_any_file(scope, path)
+    }
+
+    fn run_any_file(&self, scope: Scope, path: &str) -> PyResult<()> {
+        let path = if path.is_empty() { "???" } else { path };
```
ShaharNaveh, line 55: Are those `???` is something that CPython is also doing?

youknowone: Yes, `_PyRun_AnyFileObject` does

crates/vm/src/vm/compile.rs
```diff
@@ -44,19 +47,57 @@ impl VirtualMachine {
...
+            )?;
+            module_dict.set_item(identifier!(self, __cached__), self.ctx.none(), self)?;
+        }
+
+        // TODO: check .pyc here
+        let pyc = false;
+        if pyc {
+            todo!("running pyc is not implemented yet");
+        } else {
+            if path != "<stdin>" {
```
ShaharNaveh, line 83: How feasible doing something like:
```rust
enum InputFile {
    StdIn,
    Path(&str),
    Pyc(&str)
}
```

be?

youknowone: Thanks, that sounds like a good suggestion.
Currently, I am not sure how stdin can come here in RustPython. I ported a part of CLI run, but not entirely.
I will leave a comment about this to make sure it can be checked later.

ShaharNaveh (approved):

PR #6278, opened 2025-11-15
Fix wasm32-unknown-unknown target build

moreal:
This pull request fixes `wasm32-unknown-unknown` target build in the following 2 points.

1. It discards wasm-bindgen dependency because it needs to provide `_wbg_xxx` functions through imports object to use WASM build output.
2. Adds `#[unsafe(no_mangle)]` line for `wasm/wasm-unknown-test/src/lib.rs` file because the build without the line, doesn't include `eval` file in the build output. You can check it by running `cd wasm/wasm-unknown-test/; cargo build --target wasm32-unknown-unknown; wasm2wat target/wasm32-unknown-unknown/debug/wasm_unknown_test.wasm`. There may be no `$eval` function if the line doesn't exist.

## Links

- https://doc.rust-lang.org/stable/rustc/platform-support/wasm32-unknown-unknown.html
- https://github.com/WebAssembly/wabt

## Summary by CodeRabbit

* **Chores**
  * Updated dependency configuration to refine feature flags and ensure consistent dependency resolution across the workspace.
  * Simplified platform-specific dependency declarations to streamline WebAssembly target builds.
  * Adjusted WebAssembly export annotations to improve interoperability and runtime stability for wasm consumers.

### DIFF
--- a/crates/vm/Cargo.toml
+++ b/crates/vm/Cargo.toml
@@ -150,7 +150,6 @@ features = [
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 wasm-bindgen = { workspace = true, optional = true }
-getrandom = { workspace = true }
 
 [build-dependencies]
 glob = { workspace = true }


ShaharNaveh:
> ⚠️ Outside diff range comments (1)
> > crates/vm/Cargo.toml (1)> `27-27`: **Critical: Remove `chrono/wasmbind` from the feature definition.**
> > > The `wasmbind` feature references `chrono/wasmbind`, but the workspace configuration for `chrono` (line 165 in root `Cargo.toml`) explicitly disables default features and only enables `["clock", "oldtime", "std"]`. The `wasmbind` feature is not included, so this will cause a build error when the `wasmbind` feature is enabled.
> > > Apply this diff to fix the feature definition:
> > > ```diff
> > > -wasmbind = ["rustpython-common/wasm_js", "chrono/wasmbind", "wasm-bindgen"]
> > > +wasmbind = ["rustpython-common/wasm_js", "wasm-bindgen"]
> > > ```

@moreal

youknowone:
@ShaharNaveh Then adding wasmbind feature to vm will not enable chrono/wasmbind. @coderabbitai is wrong about it

crates/vm/Cargo.toml
```diff
@@ -148,8 +148,10 @@ features = [
   "Win32_UI_WindowsAndMessaging",
 ]
 
-[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
+[target.'cfg(target_arch = "wasm32")'.dependencies]
 wasm-bindgen = { workspace = true, optional = true }
+
+[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 getrandom = { workspace = true }
```
youknowone:
```suggestion
```
getrandom seems duplicated

crates/vm/Cargo.toml
```diff
@@ -148,8 +148,10 @@ features = [
   "Win32_UI_WindowsAndMessaging",
 ]
 
-[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
+[target.'cfg(target_arch = "wasm32")'.dependencies]
```
youknowone: Could you tell me more about this change? It will enable wasm-bindgen on wasi build. But wasi build didn't have problem without it for long time.

crates/vm/Cargo.toml
moreal: I deleted the line because it was a duplicate, even though it was there before.

crates/vm/Cargo.toml
moreal: I apologize for the confusion. That was an unnecessary change. It seems to be a remnant from testing target-specific conditions while working. For this PR, only the chrono-related changes were likely needed.

youknowone (approved):
👍 Thank you!

PR #6286, opened 2025-11-21
fix(sqlite): raise `ProgrammingError` when operating on a blob with a…

### LINKED ISSUE #6285: sqlite: Raise ProgrammingError when operating on a blob with a closed connection
Currently, the `sqlite3` module does not correctly handle cases where a `Blob` object is accessed after its parent database connection has been closed.

**Expected Behavior**
To match CPython's behavior, attempting to `read` (or perform other operations) on a blob after calling `connection.close()` should raise a `sqlite3.ProgrammingError` with the message "Cannot operate on a closed database".

**Context**
This issue prevents the standard library test `test_blob_closed_db_read` from passing. We need to add a check to ensure the connection is still open before processing blob operations.

ever0de:
… closed connection,

Fixed #6285

reference:
- https://github.com/python/cpython/blob/a8733cbc7328dbeeed50e82ef13abb197d511c46/Modules/_sqlite/connection.c#L699-L701
- https://github.com/python/cpython/blob/a8733cbc7328dbeeed50e82ef13abb197d511c46/Modules/_sqlite/blob.c#L70-L73
- https://github.com/python/cpython/blob/a8733cbc7328dbeeed50e82ef13abb197d511c46/Modules/_sqlite/blob.c#L183-L189

## Summary by CodeRabbit

* **New Features**
  * Added ability to check if a database connection is closed

* **Bug Fixes**
  * Database read operations now fail gracefully with error messaging when the connection is closed

### DIFF
--- a/crates/stdlib/src/sqlite.rs
+++ b/crates/stdlib/src/sqlite.rs
@@ -1007,6 +1007,10 @@ mod _sqlite {
             Ok(())
         }
 
+        fn is_closed(&self) -> bool {
+            self.db.lock().is_none()
+        }
+
         #[pymethod]
         fn commit(&self, vm: &VirtualMachine) -> PyResult<()> {
             self.db_lock(vm)?.implicit_commit(vm)
@@ -2169,6 +2173,13 @@ mod _sqlite {
             length: OptionalArg<c_int>,
             vm: &VirtualMachine,
         ) -> PyResult<PyRef<PyBytes>> {
+            if self.connection.is_closed() {
+                return Err(new_programming_error(
+                    vm,
+                    "Cannot operate on a closed database".to_owned(),
+                ));
+            }
+
             let mut length = length.unwrap_or(-1);
             let mut inner = self.inner(vm)?;
             let blob_len = inner.blob.bytes();
--- a/Lib/test/test_sqlite3/test_dbapi.py
+++ b/Lib/test/test_sqlite3/test_dbapi.py
@@ -1469,8 +1469,6 @@ class BlobTests(unittest.TestCase):
             with self.assertRaisesRegex(sqlite.ProgrammingError, msg):
                 blob[0] = b""
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
     def test_blob_closed_db_read(self):
         with memory_database() as cx:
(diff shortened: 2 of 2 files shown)


crates/stdlib/src/sqlite.rs
```diff
@@ -2169,6 +2173,13 @@ mod _sqlite {
             length: OptionalArg<c_int>,
             vm: &VirtualMachine,
         ) -> PyResult<PyRef<PyBytes>> {
+            if self.connection.is_closed() {
```
ShaharNaveh, line 2176: Can you please put the link in the PR here? it will be easier to access rather than going to git blame -> PR -> links

ever0de: Could you please clarify the context? I am currently unable to understand why it's required.

ShaharNaveh: Not required. Just my personal preference

youknowone: This isn’t the official stance of our project, but I’d like to share my personal view on the matter.

If the behavior is something that’s common sense and easily predictable, I don’t think we need to attach a reference. In my opinion, this change falls into that category. Before adding this code, it might have been unclear whether such a check was necessary, but once the code is added, it becomes obvious that it is.

Some behaviors are hard to infer from the code alone. In those cases, having a reference linked is helpful. Depending on the situation, when pointing to the exact code is useful, we link specific line numbers; when it’s more important to refer to the conceptual location, we just mention the function name instead.

crates/stdlib/src/sqlite.rs
```diff
@@ -2169,6 +2173,13 @@ mod _sqlite {
...
+                return Err(new_programming_error(
+                    vm,
+                    "Cannot operate on a closed database".to_owned(),
+                ));
+            }
```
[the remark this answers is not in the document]
youknowone: @ever0de How do you think about this comment?

youknowone (approved):
Thanks!

crates/stdlib/src/sqlite.rs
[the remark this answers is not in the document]
ever0de: @youknowone Can I handle this in a separate issue?

PR #6288, opened 2025-11-21
Fix sqlite connection reinitialization

### LINKED ISSUE #6287: sqlite: Handle state invalidation when Connection.__init__ fails during re-initialization
Currently, `sqlite3` does not correctly update the internal state of a `Connection` object if `__init__` is called a second time and fails (e.g., due to an `OperationalError`).

**Expected Behavior**
If `Connection.__init__` fails (even on an already open connection), the object should be considered uninitialized. Subsequent attempts to use methods like `executemany` should raise a `sqlite3.ProgrammingError` with the message "Base Connection.__init__ not called", matching CPython's behavior.

**Context**
This behavior is required to pass `test_connection_bad_reinit`. We need to ensure the connection's initialized flag is reset or checked properly when re-initialization fails.

ever0de:
Fixed #6287

reference
- https://github.com/python/cpython/blob/49ff8b6cc017bb6e9bdb3bf4918d65f32f7aaed8/Modules/_sqlite/connection.c#L247-L255
- https://github.com/python/cpython/blob/49ff8b6cc017bb6e9bdb3bf4918d65f32f7aaed8/Modules/_sqlite/connection.c#L276-L279
- https://github.com/python/cpython/blob/49ff8b6cc017bb6e9bdb3bf4918d65f32f7aaed8/Modules/_sqlite/connection.c#L344-L350

## Summary by CodeRabbit

* **Refactor**
  * Improved concurrency and initialization safety for the built-in SQLite connection: core connection state now uses concurrency-safe primitives with an explicit initialization guard and atomic state operations.
  * Revised lifecycle and reinitialization flow to open/close safely, enforce thread confinement when enabled, reset row/text factories to defaults on reinit, and ensure failed opens leave the connection uninitialized.

### DIFF
--- a/crates/stdlib/src/sqlite.rs
+++ b/crates/stdlib/src/sqlite.rs
@@ -837,7 +837,7 @@ mod _sqlite {
         detect_types: PyAtomic<c_int>,
         isolation_level: PyAtomicRef<Option<PyStr>>,
         check_same_thread: PyAtomic<bool>,
-        thread_ident: PyMutex<ThreadId>,  // TODO: Use atomic
+        thread_ident: PyMutex<ThreadId>, // TODO: Use atomic
         row_factory: PyAtomicRef<Option<PyObject>>,
         text_factory: PyAtomicRef<PyObject>,
     }


crates/stdlib/src/sqlite.rs
```diff
@@ -833,10 +833,11 @@ mod _sqlite {
...
     struct Connection {
         db: PyMutex<Option<Sqlite>>,
-        detect_types: c_int,
+        initialized: PyAtomic<bool>,
+        detect_types: PyAtomic<c_int>,
         isolation_level: PyAtomicRef<Option<PyStr>>,
-        check_same_thread: bool,
-        thread_ident: ThreadId,
+        check_same_thread: PyAtomic<bool>,
+        thread_ident: PyMutex<ThreadId>,
```
youknowone:
```suggestion
        thread_ident: PyMutex<ThreadId>,  // TODO: Use atomic
```
ThreadId is representable in u64. But not convenient yet.

youknowone:
Could you check this comment? https://github.com/RustPython/RustPython/pull/6288#pullrequestreview-3493579045

This sounds reasonable. And because it is initialization step, we don't have much disadvantage to use more stronger ordering.

PR #6298, opened 2025-11-27
Fix redox compilation in stdlib

willnode:
This fixes compilation on Redox due to cfg filter in this code, and `fcntl::open` here returns OwnedFd already.

This upstream version can't be compiled to Redox due to other reasons, so as a workaround this patch fix is verified in Redox build system using a [backported version](https://gitlab.redox-os.org/redox-os/redox/-/merge_requests/1715).

## Summary by CodeRabbit

* **Refactor**
  * Improved internal file descriptor handling for enhanced code maintainability. No end-user facing changes.


youknowone (approved):
Thanks!

PR #6301, opened 2025-11-28
rustpython-cpython

youknowone:
Test code:

```python
import _cpython

@_cpython.call
def decimal_add(a, b):
    import _decimal
    return _decimal.Decimal(a) + _decimal.Decimal(b)

result = decimal_add('1.1', '2.2')  # '3.3'
print(result)
```

Test command:

```sh
cargo run --features cpython test_cpython.py
```

This might be something awful. Now RustPython can borrow any feature from CPython, as long as CPython is in the environment and the calling function’s arguments and return type are pickle-serializable.

Will this become a new RustPython feature? Otherwise I’ll just create it outside the org repo.

### DIFF
--- a/extra_tests/test_module_pyo3.py
+++ b/extra_tests/test_module_pyo3.py
@@ -1,22 +1,22 @@
 """
-Tests for the _cpython module - RustPython to CPython bridge.
+Tests for the pyo3 module - RustPython to CPython bridge.
 
-This module requires the `cpython` feature to be enabled:
-    cargo build --release --features cpython
+This module requires the `pyo3` feature to be enabled:
+    cargo build --release --features pyo3
 
 Run with:
-    ./target/release/rustpython extra_tests/test_cpython.py
+    ./target/release/rustpython extra_tests/test_pyo3.py
 """
 
-import _cpython
+import pyo3
 
 
-# Test 1: @_cpython.call decorator
+# Test 1: @pyo3.wraps decorator
 
-print("Test 1: @_cpython.call decorator")
+print("Test 1: @pyo3.wraps decorator")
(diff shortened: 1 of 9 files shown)


arihant2math:
We could just replace the `ctypes` implementation with this :). Would make life much easier.

This sort of strays away from the mission of having a pure-rust compiler no?

youknowone:
> We could just replace the ctypes implementation with this :). Would make life much easier.

On environments with available CPythons, that makes sense. But we are more likely on wasm targets.

> This sort of strays away from the mission of having a pure-rust compiler no?

This will be optional feature. Not every data is pickle-serializable. It hopefully will fill gaps until we have rust implementation for them

extra_tests/test_module_pyo3.py
```diff
@@ -0,0 +1,223 @@
...
+print(f"d1 = {d1}")
+print(f"d2 = {d2}")
+
+result = d1 + d2
+print(f"d1 + d2 = {result}")
+assert '3.3' in str(result), f"Expected 3.3, got {result}"
+print("OK!\n")
+
+
+# Test 3: import_module() with numpy
```
youknowone, line 51: Now RustPython can run numpy.
Though it will not be that much practical.

ShaharNaveh (approved):
Overall, very cool!

I do agree that this goes against the project goal in some ways. BUT if we have it behind a feature flag (that's not on by default) then I don't see an issue with letting users choose this functionality.

As an added bonus can we state somewhere in the docs that we have (experimental) support for installing packages that use extension modules (not pure python), like pandas, numpy, pedantic, etc...?

PR #6303, opened 2025-11-28
Remove `Rotate*` & `Duplicate*` instructions

ShaharNaveh:
Implementing https://github.com/python/cpython/pull/30902

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal bytecode instruction set by consolidating stack manipulation operations, replacing redundant instruction variants with more efficient alternatives. These changes streamline compiler code generation and virtual machine execution while maintaining identical external behavior.


ShaharNaveh:
@youknowone When running this on my local machine I had to cleanup all `*.pyc` files before running some tests, as otherwise the VM crashes.
CPython doesn't have this issue as they are changing opcode IDs only for a version change (i.e. 3.12 -> 3.13). How do you think we should address this? Or it's not something worth investing time on as we don't change the opcodes often

youknowone:
> CPython doesn't have this issue as they are changing opcode IDs only for a version change (i.e. 3.12 -> 3.13). How do you think we should address this? Or it's not something worth investing time on as we don't change the opcodes often

We'd better not to change it on same magic number, yep. But we didn't address it anywhere yet.

PR #6312, opened 2025-11-29
Sort `Instruction` variants + Ensure each holds a singular oparg type

ShaharNaveh:
One of the changes made is to sort all variants alphabetically, this will makes it easier for finding if an instruction exist, especially when comparing it to https://github.com/python/cpython/blob/cfcd52490d6531f3b0e8ddd4bb2b40fb6baee854/Include/opcode_ids.h (which is sorted alphabetically).

The second change I did was to change each named field to an associated value because we are only holding the oparg type. 

For example:

```rust
Foo { index: Arg<OpargType> }
```

to:

```rust
Foo(Arg<OpargType>)
```
IMO this has the following benefits:
- signals that the `Instruction` only holds the oparg type not the actual value.
- Will make the `Instruction` easier to auto-generate (what #6174 tried to do).

---

The reason why we had named fields was because in the first iterations of the `Instruction` enum we held the actual oparg value as a named field inside the Instruction, which is no longer the case. 

IMO this enum should be renamed to `Opcode` and have a method called `oparg` that looks like:
```rust
enum Opcode {
     Foo,
     Bar,
     ...
}

impl Opcode {
    fn oparg<T: OpArgType>(self) -> Option<T> { ... }
}
```

but that's a different discussion

## Summary by CodeRabbit

* **Refactor**
  * Standardized the instruction/bytecode representation across the compiler, VM, and JIT for a simpler, more consistent surface.

* **New Features**
  * Broader VM/JIT support for many additional bytecode operations, improving data-structure construction, call patterns, async/with, and iterator flows.

* **Bug Fixes**
  * Improved import/name resolution, mapping/unpacking, exception unwinding, and async/context-manager protocol handling for more correct runtime behavior.

### DIFF
--- a/crates/compiler-core/src/bytecode.rs
+++ b/crates/compiler-core/src/bytecode.rs
@@ -550,42 +550,42 @@ pub type NameIdx = u32;
 #[repr(u8)]
 pub enum Instruction {
     BeforeAsyncWith,
-    BinaryOperation(Arg<BinaryOperator>),
-    BinaryOperationInplace(Arg<BinaryOperator>),
+    BinaryOperation(Arg<BinaryOperator>),        // op
+    BinaryOperationInplace(Arg<BinaryOperator>), // op
     BinarySubscript,
-    Break(Arg<Label>),
-    BuildList(Arg<u32>),
-    BuildListFromTuples(Arg<u32>),
-    BuildMap(Arg<u32>),
-    BuildMapForCall(Arg<u32>),
-    BuildSet(Arg<u32>),
-    BuildSetFromTuples(Arg<u32>),
+    Break(Arg<Label>),             // target
+    BuildList(Arg<u32>),           // size
+    BuildListFromTuples(Arg<u32>), // size
+    BuildMap(Arg<u32>),            // size
+    BuildMapForCall(Arg<u32>),     // size
+    BuildSet(Arg<u32>),            // size
+    BuildSetFromTuples(Arg<u32>),  // size
     /// When holds `true` it will build a slice with a third step argument.
-    BuildSlice(Arg<bool>),
-    BuildString(Arg<u32>),
-    BuildTuple(Arg<u32>),
+    BuildSlice(Arg<bool>), // step
+    BuildString(Arg<u32>),         // size
+    BuildTuple(Arg<u32>),          // size
     BuildTupleFromIter,
-    BuildTupleFromTuples(Arg<u32>),
+    BuildTupleFromTuples(Arg<u32>), // size
     CallFunctionEx(Arg<bool>),
     CallFunctionKeyword(Arg<u32>),
     CallFunctionPositional(Arg<u32>),
     CallIntrinsic1(Arg<IntrinsicFunction1>),
     CallIntrinsic2(Arg<IntrinsicFunction2>),
-    CallMethodEx(Arg<bool>),
-    CallMethodKeyword(Arg<u32>),
-    CallMethodPositional(Arg<u32>),
-    CompareOperation(Arg<ComparisonOperator>),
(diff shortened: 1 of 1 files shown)


crates/compiler-core/src/bytecode.rs
```diff
@@ -549,263 +549,146 @@ pub type NameIdx = u32;
...
+    ReturnConst(Arg<u32>),
+    ReturnValue,
+    Reverse(Arg<u32>),
+    SetAdd(Arg<u32>),
+    SetFunctionAttribute(Arg<MakeFunctionFlags>),
     SetupAnnotation,
-    SetupLoop,
-
+    SetupAsyncWith(Arg<Label>),
+    SetupExcept(Arg<Label>),
```
youknowone: how about at least keeping the name as comment?

e.g.

```suggestion
    SetupExcept(Arg<Label>),  // handler
```

ShaharNaveh: Sure, I can do that

youknowone: Could you also tell me about the motivation to remove the label? They compiled to the same bytecode, but having label is easier to read the code

ShaharNaveh: The motivation is the inconsistenty between opcodes that holds the same `OpArgType`, some holds it as an associated value and some as a named field, which is a bit confusing imo. also, I find it very tedious and just more boilerplate for having different an enum that it's variants hold nothing or a single named field.

I'm not so good with words so here's what I mean:

I can justify having enums like this where different variants hold different fields, even if some overlapping:
```rust
enum Foo {
    A { name: String, age: u8 },
    B { name: String, job: String, married: bool },
    ...
}
```

But I do find it redundant and overly verbose to have something like this:
```rust
enum Foo {
    A { name: String },
    B { name: String },
    C { name: String },
    ...
}
```

Not to mention that we have cases like this:
```rust
enum Foo {
    A { target: Label },
    B { end: Label },
    C { handler: Label },
    ...
}
```

Because `Label` is a newtype for `u32` already, it's just confusing as to why the same newtype is being "treated" differently. I'd much rather to go with newtyprs all the way like:
```rust
enum Foo {
    A(TargetLabel),
    B(EndLabel),
    C(HandlerLabel),
    ...
}
```

ShaharNaveh: @youknowone I can revert that change if that's a deal breaker for you:)

youknowone: I like the new enum idea on #6313. Not sure about Label. Because all the labels are actually label.

If the consistency matters, giving label for every instruction looks better to me. It feels like redundancy while writing code, but we read code more than writing code. Getting information from code is easier than checking document.

The situation is also a bit different to your example. It is more like:
```rust
enum Foo {
    CallMethodPositional { nargs: String },
    ReturnConst { idx: String },
    ...
}
```

Maybe giving different types for them will also make sense though.

cc @coolreader18

ShaharNaveh: > I like the new enum idea on #6313. Not sure about Label. Because all the labels are actually label.
> 
> If the consistency matters, giving label for every instruction looks better to me. It feels like redundancy while writing code, but we read code more than writing code. Getting information from code is easier than checking document.

IMO consistency matters, I can make them consistent in the other way around by having each Oparg to be stored as a named field if that's desired.

> The situation is also a bit different to your example. It is more like:
> 
> ```rust
> enum Foo {
>     CallMethodPositional { nargs: String },
>     ReturnConst { idx: String },
>     ...
> }
> ```
>
> Maybe giving different types for them will also make sense though.

I think that it will, but that is out of scope for this PR, I'll do it in a separate one
> cc @coolreader18

PR #6313, opened 2025-11-29
Ensure `BuildSlice` oparg to be either 2 or 3

ShaharNaveh:
We are currently only checking if the argument count is 3, meaning that even if we pass it `-1`, `0` or `999` it will act as if `2` was passed into it.

This change ensures that the oparg for `BuildSlice` can be either `2` or `3`

## Summary by CodeRabbit

* **Refactor**
  * Slice construction now encodes whether a slice has two or three arguments (instead of a boolean "step" flag), unifying compile-time and runtime handling.
  * Bytecode and execution paths updated to use this explicit argument-count form; behavior is preserved while improving consistency and clarity.

### DIFF
--- a/crates/compiler-core/src/bytecode.rs
+++ b/crates/compiler-core/src/bytecode.rs
@@ -10,7 +10,7 @@ use itertools::Itertools;
 use malachite_bigint::BigInt;
 use num_complex::Complex64;
 use rustpython_wtf8::{Wtf8, Wtf8Buf};
-use std::{collections::BTreeSet, fmt, hash, marker::PhantomData, mem, ops::Deref};
+use std::{collections::BTreeSet, fmt, hash, marker::PhantomData, mem, num::NonZeroU8, ops::Deref};
 
 #[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
 #[repr(i8)]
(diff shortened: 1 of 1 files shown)


crates/compiler-core/src/bytecode.rs
```diff
@@ -1151,6 +1150,22 @@ op_arg_enum!(
...
+    #[derive(Clone, Copy, Debug, Eq, PartialEq)]
+    pub enum BuildSliceArgCount {
+        /// ```py
+        /// x[5:10]
+        /// ```
+        Two = 2,
+        /// ```py
+        /// x[5:10:2]
+        /// ```
+        Three = 3,
```
youknowone:
```suggestion
        /// ```py
        /// x[5:10]
        /// ```
        Two,
        /// ```py
        /// x[5:10:2]
        /// ```
        Three,
```
Removing value not to block optimization. Adding a method will be enough

```rust
impl BuildSliceArgCount {
    const fn argc(&self) -> usize {
        ....
    }
}

youknowone (changes requested):

crates/compiler-core/src/bytecode.rs
ShaharNaveh: Hmmm, but then how this be validated at `stack_effect`?

youknowone: Not by using `as` cast, but something like `step.get(arg).argc()` which return 2 or 3.

youknowone (approved):
👍

PR #6316, opened 2025-11-30
POC: Configurable `Instruction` autogen

ShaharNaveh:
> [!NOTE]
> This is very much at a POC stage ATM 

The idea here is to auto-generate as much as possible, whether it's oparg field name (`namei/consti/target`/etc), the oparg type, oparg ID and opcode name.
+
An important feature using the config file is to have the ability to control every attribute in every opcode, we can exclude opcodes that are not supported yet or add new ones that do not appear in the targeted CPython version (not supported yet, but planning to add it).

This will ensure consistency for the enum variant.

lmk what you think:)

cc: @coolreader18 @youknowone 

## Summary by CodeRabbit

* **New Features**
  * Enhanced bytecode instruction support with improved opcode handling and generation capabilities.

* **Chores**
  * Integrated Python 3.13.9 compatibility and updated build tooling infrastructure.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,5 @@
 [submodule "submodules/cpython-v3.13.9"]
 	path = submodules/cpython-v3.13.9
 	url = https://github.com/python/cpython.git
+	branch = 3.13
     shallow = true


youknowone:
is this now outdated?

ShaharNaveh:
> is this now outdated?

yup

PR #6319, opened 2025-12-02
New subclass payload layout

youknowone:
I am considering new subclass payload layout to provide
1. Actually subclass typed function parameters, like using PyRef<PyBool> instead PyRef<PyInt>
2. adding new fields to the subclass payload, which cause error right now.

I will left open questions on the code.

## Summary by CodeRabbit

* **New Features**
  * Centralized OS/errno error construction with OS-subtype exceptions; SSL/socket errors now surface as OS-derived errors.
  * New subclass/base support: safe upcast/into_base operations and PySubclass support for transparent subclasses.

* **Bug Fixes**
  * Runtime validation for subclass base-field layout and size-aware downcast errors.
  * Exception payloads now expose errno/strerror/filename fields with getters/setters.

* **Refactor**
  * Widespread conversion to transparent-wrapper/base-embedded patterns across ctypes, IO, bool/int, AST, struct-sequence and exception types.

### DIFF
--- a/crates/vm/src/stdlib/io.rs
+++ b/crates/vm/src/stdlib/io.rs
@@ -87,9 +87,7 @@ pub(crate) fn make_module(vm: &VirtualMachine) -> PyRef<PyModule> {
     #[cfg(any(not(target_arch = "wasm32"), target_os = "wasi"))]
     fileio::extend_module(vm, &module).unwrap();
 
-    let unsupported_operation = _io::UNSUPPORTED_OPERATION
-        .get_or_init(|| _io::make_unsupportedop(ctx))
-        .clone();
+    let unsupported_operation = _io::unsupported_operation().to_owned();
     extend_module!(vm, &module, {
         "UnsupportedOperation" => unsupported_operation,
         "BlockingIOError" => ctx.exceptions.blocking_io_error.to_owned(),
@@ -204,7 +202,8 @@ mod _io {
     }
 
     pub fn new_unsupported_operation(vm: &VirtualMachine, msg: String) -> PyBaseExceptionRef {
-        vm.new_exception_msg(UNSUPPORTED_OPERATION.get().unwrap().clone(), msg)
+        vm.new_os_subtype_error(unsupported_operation().to_owned(), None, msg)
+            .upcast()
     }
 
     fn _unsupported<T>(vm: &VirtualMachine, zelf: &PyObject, operation: &str) -> PyResult<T> {
@@ -4243,11 +4242,7 @@ mod _io {
         }
     }
 
-    rustpython_common::static_cell! {
-        pub(super) static UNSUPPORTED_OPERATION: PyTypeRef;
-    }
-
-    pub(super) fn make_unsupportedop(ctx: &Context) -> PyTypeRef {
+    fn create_unsupported_operation(ctx: &Context) -> PyTypeRef {
         use crate::types::PyTypeSlots;
         PyType::new_heap(
             "UnsupportedOperation",
@@ -4261,6 +4256,13 @@ mod _io {
             ctx,
         )
         .unwrap()
+    }
+
+    pub fn unsupported_operation() -> &'static Py<PyType> {
+        rustpython_common::static_cell! {
+            static CELL: PyTypeRef;
+        }
+        CELL.get_or_init(|| create_unsupported_operation(Context::genesis()))
     }
 
     #[pyfunction]
--- a/crates/vm/src/vm/vm_new.rs
+++ b/crates/vm/src/vm/vm_new.rs
@@ -107,6 +107,11 @@ impl VirtualMachine {
         )
     }
 
+    pub fn new_os_error(&self, msg: impl ToPyObject) -> PyRef<PyBaseException> {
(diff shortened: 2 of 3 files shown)


crates/vm/src/builtins/bool.rs
```diff
@@ -78,18 +79,46 @@ impl PyObjectRef {
...
     #[inline]
     fn class(ctx: &Context) -> &'static Py<PyType> {
         ctx.types.bool_type
     }
+
+    /// PyBool reuses PyInt's TypeId
+    #[inline]
+    fn payload_type_id() -> std::any::TypeId {
+        std::any::TypeId::of::<PyInt>()
+    }
```
youknowone: This is okay here, because PyInt and PyBool shares same payload. The only difference is type.

But implementing this makes `PyPayload::downcastable` to be automatically `true` for any kind of `PyInt`. It doesn't make failing tests on this patch because we don't use that pattern. But it obviously will be a sort of trap for future developing.

~~Another question is that `PyPayload::downcastable` only takes &PyObject currently. We can't compare it to actual type... because we don't have vm.ctx.~~

crates/vm/src/builtins/bool.rs
```diff
@@ -78,18 +79,46 @@ impl PyObjectRef {
 }
 
 #[pyclass(name = "bool", module = false, base = PyInt)]
-pub struct PyBool;
+#[repr(transparent)]
+pub struct PyBool(pub PyInt);
```
youknowone, line 90: Now we have to manually define the base layout to be always leading field.
The problem is Rust actually only guarantees it to be actually same layout only when #[repr(transparent)] is used. 

What I didn't check yet: Can using #[repr(C)] helps? Is there any other repr options to keep the first field as same as its original layout?

crates/derive-impl/src/pyclass.rs
```diff
@@ -528,19 +528,46 @@ pub(crate) fn impl_pyclass(attr: PunctuatedNestedMeta, item: Item) -> Result<Tok
...
         quote! {
             impl ::rustpython_vm::PyPayload for #ident {
+                #[inline]
+                fn payload_type_id() -> ::std::any::TypeId {
+                    <#base_type as ::rustpython_vm::PyPayload>::payload_type_id()
+                }
+
+                #[inline]
+                fn validate_downcastable_from(obj: &::rustpython_vm::PyObject) -> bool {
+                    <Self as ::rustpython_vm::class::PyClassDef>::BASICSIZE <= obj.class().slots.basicsize && obj.class().fast_issubclass(<Self as ::rustpython_vm::class::StaticType>::static_type())
```
youknowone, line 631: Maybe I found another way to do with more tricks. Let's check it with OSError

youknowone:
@ShaharNaveh This is almost done! Could you give a look again?

youknowone:
One question: Passing OSError to new_exception is a runtime error. Can this be checked earlier?

crates/derive-impl/src/pyclass.rs
```diff
@@ -305,6 +305,87 @@ pub(crate) fn impl_pyclass_impl(attr: PunctuatedNestedMeta, item: Item) -> Resul
...
+
+    match &item_struct.fields {
+        syn::Fields::Named(fields) => {
+            if fields.named.is_empty() {
+                bail_span!(
+                    item_struct,
+                    "#[pyclass] with base = {base_name} requires the first field to be of type {base_name}, but the struct has no fields"
+                );
+            }
+            let first_field = fields.named.first().unwrap();
```
ShaharNaveh: You can remove the check for `is_empty` above, and have `unwrap_or` on `fields.named.first().unwrap_or(bail_span!(...))`

crates/derive-impl/src/pyclass.rs
```diff
@@ -305,6 +305,87 @@ pub(crate) fn impl_pyclass_impl(attr: PunctuatedNestedMeta, item: Item) -> Resul
...
+            }
+        }
+        syn::Fields::Unnamed(fields) => {
+            if fields.unnamed.is_empty() {
+                bail_span!(
+                    item_struct,
+                    "#[pyclass] with base = {base_name} requires the first field to be of type {base_name}, but the struct has no fields"
+                );
+            }
+            let first_field = fields.unnamed.first().unwrap();
```
ShaharNaveh: same thing

crates/derive-impl/src/pyclass.rs
```diff
@@ -579,14 +715,30 @@ pub(crate) fn impl_pyexception_impl(attr: PunctuatedNestedMeta, item: Item) -> R
...
-        return Err(syn::Error::new_spanned(
-            &attr[0],
-            "#[pyexception] impl doesn't allow attrs. Use #[pyclass] instead.",
-        ));
+    // Check if with(Constructor) is specified
+    let mut has_constructor_trait = false;
+    let mut extra_attrs = Vec::new();
+    for nested in &attr {
+        if let NestedMeta::Meta(Meta::List(MetaList { path, nested, .. })) = nested {
+            if path.is_ident("with") {
```
ShaharNaveh:
```suggestion
            // If we already found the constructor trait, no need to keep looking for it
            if !has_constructor_trait && path.is_ident("with") {
```

Issue #6113, opened 2025-08-26
fn try_to_bool calls `__bool__` method instead of number slot boolean


youknowone:
## Summary

`fn try_to_bool` is expected to call `PyNumberSlots`'s boolean, but currently calling python `__bool__`

youknowone:
@YashSuthar983 How did this work when you tried 50cff20?

YashSuthar983:
@youknowone sorry i was out of town . I will let you know ,let me continue on this.

PR #6328, opened 2025-12-04
prefer nb_bool slot in try_to_bool instead of __bool__

YashSuthar983:
Fixed: #6113

## Summary by CodeRabbit

* **Chores**
  * Improved boolean evaluation performance by adding a faster internal path for boolean checks.
* **Bug Fixes**
  * Preserved and unified error behavior for invalid boolean/length results so type and value errors remain consistent.

### DIFF
--- a/crates/vm/src/builtins/bool.rs
+++ b/crates/vm/src/builtins/bool.rs
@@ -42,6 +42,7 @@ impl PyObjectRef {
         let rs_bool = if let Some(nb_bool) = self.class().slots.as_number.boolean.load() {
             nb_bool(self.as_object().to_number(), vm)?
         } else {
+            // TODO: Fully implement AsNumber and remove this block
             match vm.get_method(self.clone(), identifier!(vm, __bool__)) {
                 Some(method_or_err) => {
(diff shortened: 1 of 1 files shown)


crates/vm/src/builtins/bool.rs
```diff
@@ -39,39 +39,43 @@ impl PyObjectRef {
...
-                        bool_obj.class().name()
-                    )));
-                }
-
-                get_value(&bool_obj)
-            }
-            None => match vm.get_method(self, identifier!(vm, __len__)) {
+        let rs_bool = if let Some(nb_bool) = self.class().slots.as_number.boolean.load() {
+            nb_bool(self.as_object().to_number(), vm)?
+        } else {
```
youknowone, line 44:
```suggestion
        } else {
            // TODO: Fully implement AsNumber and remove this block
```

youknowone (approved):
Thank you!
In principle, if the Python method `__bool__` is defined, there must also be an `nb_bool`, so it seems there’s still some incorrect implementation left somewhere.

PR #6352, opened 2025-12-08
Fix os.remove

youknowone:
## Summary by CodeRabbit

* **Breaking Changes**
  * DirEntry and ScandirIterator objects can no longer be directly instantiated; they are now system-generated only.

* **Bug Fixes**
  * Improved detection and handling of Windows directory symlinks and junctions in file operations.

* **Refactor**
  * Consolidated Windows path string handling across modules for consistency and maintainability.

### DIFF
--- a/Lib/test/test_os.py
+++ b/Lib/test/test_os.py
@@ -736,7 +736,6 @@ class StatAttributeTests(unittest.TestCase):
             result.st_file_attributes & stat.FILE_ATTRIBUTE_DIRECTORY,
             stat.FILE_ATTRIBUTE_DIRECTORY)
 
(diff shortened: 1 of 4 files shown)


fanninpm:
Does this fix `support.rmtree()`? If so, we can likely re-enable `test_pathlib` on the Windows CI.

youknowone:
Good point. Unfortunately not yet, but it is about another error related to surrogate

fanninpm:
> Good point. Unfortunately not yet, but it is about another error related to surrogate

It would be a good idea to mention that in `.github/workflows/ci.yaml`.

PR #6370, opened 2025-12-09
Update CI workflow comment about test_pathlib 

youknowone:
## Release Notes

* **Chores**
  * Updated CI workflow documentation comments to reflect testing focus and improve clarity.


fanninpm (approved):
This is good as it currently stands; however, it would be better if we could fix the problem altogether.

youknowone:
Could anyone give a try? It is about wrong assumption about utf8 str when having surrogate.
Probably matter of time

PR #6375, opened 2025-12-09
[easy][minor] Freeze ruff libraries to the REV that the TAG points to.

slawlor:
Tags can move and be re-aliased to different revisions, and this actually freezes the rev (similar to how a published crate would) so that it only applies to this specific commit hash. (Tag release link: https://github.com/astral-sh/ruff/releases/tag/0.14.1)

This has caused some minor problems when vendoring sources where Ruff is used by other projects and is already pulled in by a rev, causing `cargo vendor` to crash.

Thanks for the awesome project!

## Summary by CodeRabbit

* **Chores**
  * Pinned four internal parsing/text dependencies to specific commit revisions (keeping the same source) to ensure reproducible builds and stability. No other dependencies or configuration changed; no public APIs or runtime behavior altered.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2797,7 +2797,7 @@ dependencies = [
 [[package]]
 name = "ruff_python_ast"
 version = "0.0.0"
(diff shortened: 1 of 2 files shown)


youknowone:
Could you tell me what are you building with RustPython if you don't mind? Just in curiuos.

PR #6403, opened 2025-12-11
TypeData

youknowone:
## Summary by CodeRabbit

* **New Features**
  * Added a public per-type data API for heap-based types with thread-safe, typed attachments.
  * Added a helper to create dynamic array types with attached storage metadata.

* **Refactor**
  * C-types (array/struct/union/pointer/simple) now derive sizing, alignment, and element info from per-type metadata instead of per-instance fields, simplifying creation and multiplication behavior.

### DIFF
--- a/crates/vm/src/stdlib/ctypes.rs
+++ b/crates/vm/src/stdlib/ctypes.rs
@@ -389,37 +389,33 @@ pub(crate) mod _ctypes {
     /// Get the size of a ctypes type or instance
     #[pyfunction(name = "sizeof")]
     pub fn size_of(obj: PyObjectRef, vm: &VirtualMachine) -> PyResult<usize> {
-        use super::array::{PyCArray, PyCArrayType};
         use super::pointer::PyCPointer;
         use super::structure::{PyCStructType, PyCStructure};
-        use super::union::{PyCUnion, PyCUnionType};
+        use super::union::PyCUnionType;
+        use super::util::StgInfo;
+        use crate::builtins::PyType;
 
-        // 1. Instances with stg_info
-        if obj.fast_isinstance(PyCArray::static_type()) {
-            // Get stg_info from the type
-            if let Some(type_obj) = obj.class().as_object().downcast_ref::<PyCArrayType>() {
-                return Ok(type_obj.stg_info.size);
-            }
+        // 1. Check TypeDataSlot on class (for instances)
+        if let Some(stg_info) = obj.class().get_type_data::<StgInfo>() {
+            return Ok(stg_info.size);
         }
+
+        // 2. Check TypeDataSlot on type itself (for type objects)
+        if let Some(type_obj) = obj.downcast_ref::<PyType>()
+            && let Some(stg_info) = type_obj.get_type_data::<StgInfo>()
+        {
+            return Ok(stg_info.size);
+        }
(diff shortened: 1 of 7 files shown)


crates/vm/src/builtins/type.rs
```diff
@@ -32,7 +32,121 @@ use crate::{
...
+    ops::Deref,
+    pin::Pin,
+    ptr::NonNull,
+};
+
+/// Type-erased storage for extension module data attached to heap types.
+pub struct TypeDataSlot {
+    // PyObject_GetTypeData
+    type_id: TypeId,
+    data: Box<dyn Any + Send + Sync>,
```
ShaharNaveh: Hmmm, does it need to be `dyn`? I'd rather have it like:
```rust

struct TypeDataSlot<T> {
  ...
data: Box<T>,
}

impl<T> TypeDataSlot<T> 
where T: Any + Send + Sync {
   ...
}
```

crates/vm/src/builtins/type.rs
```diff
@@ -32,7 +32,121 @@ use crate::{
...
+        }
+    }
+}
+
+impl<T: Any + 'static> std::ops::Deref for TypeDataRef<'_, T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        // SAFETY: try_new already verified the type matches
+        self.guard.as_ref().unwrap().get::<T>().unwrap()
```
ShaharNaveh: Because `try_new` already validated it, we can use `unwrap_unchecked ` right?

crates/vm/src/builtins/type.rs
```diff
@@ -32,7 +32,121 @@ use crate::{
...
+        }
+    }
+}
+
+impl<T: Any + 'static> std::ops::Deref for TypeDataRefMut<'_, T> {
+    type Target = T;
+
+    fn deref(&self) -> &Self::Target {
+        // SAFETY: try_new already verified the type matches
+        self.guard.as_ref().unwrap().get::<T>().unwrap()
```
ShaharNaveh: same

crates/vm/src/builtins/type.rs
```diff
@@ -32,7 +32,121 @@ use crate::{
...
+    }
+}
+
+impl<T: Any + 'static> std::ops::DerefMut for TypeDataRefMut<'_, T> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        // SAFETY: try_new already verified the type matches
+        self.guard.as_mut().unwrap().get_mut::<T>().unwrap()
```
ShaharNaveh: same

crates/vm/src/builtins/type.rs
youknowone: If TypeDataSlot<T> is a generic type, how does it be saved in HeapTypeExt?

ShaharNaveh: > If TypeDataSlot is a generic type, how does it be saved in HeapTypeExt?

Good point, somehow missed that tiny detail :facepalm:

crates/vm/src/builtins/type.rs
youknowone: Good catch! I removed Option from here

PR #6400, opened 2025-12-11
Faulthandler

youknowone:
## Summary by CodeRabbit

* **New Features**
  * Enhanced cross-platform fault handler with richer crash diagnostics and user-facing controls (enable/disable, query status).
  * Stack-trace dumping improvements, including scheduled (timeout-based) dumps, cancellation, and per-signal configurable behavior.

* **Behavior**
  * Error/report labeling changed from "<stdin>" to "<string>" for command-run outputs.

* **Chores**
  * Added Windows diagnostic dependency to support improved Windows crash handling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6406, opened 2025-12-11
Faulthandler

youknowone:
## Summary by CodeRabbit

* **New Features**
  * Added faulthandler feature for diagnostic output on fatal errors
  * Multiple activation methods: use `-X faulthandler` command-line flag, set `PYTHONFAULTHANDLER` environment variable, or automatic activation in development mode

### DIFF
--- a/crates/stdlib/src/faulthandler.rs
+++ b/crates/stdlib/src/faulthandler.rs
@@ -9,7 +9,7 @@ mod decl {
     };
     #[cfg(any(unix, windows))]
     use rustpython_common::os::{get_errno, set_errno};
-    use std::sync::atomic::{AtomicBool, AtomicI32, AtomicUsize, Ordering};
+    use std::sync::atomic::{AtomicBool, AtomicI32, Ordering};
     use std::sync::{Arc, Condvar, Mutex};
     use std::thread;
     use std::time::Duration;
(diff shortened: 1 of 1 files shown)


crates/stdlib/src/faulthandler.rs
```diff
@@ -6,13 +6,114 @@ mod decl {
...
+            name: "Aborted",
+            previous: 0,
+        },
+        FaultHandler {
+            signum: libc::SIGSEGV,
+            enabled: false,
+            name: "Segmentation fault",
+            previous: 0,
+        },
+    ];
```
fanninpm, line 78: When I was workshopping this on my own, prior to #6400, I came up with something similar to this, finding out from the `libc` repository that `SIGILL` does exist on Windows. Would this work?

```suggestion
    // CPython uses static arrays for signal handlers which requires mutable static access.
    // This is safe because:
    // 1. Signal handlers run in a single-threaded context (from the OS perspective)
    // 2. FAULTHANDLER_HANDLERS is only modified during enable/disable operations
    // 3. This matches CPython's faulthandler.c implementation
    #[allow(static_mut_refs)]
    static mut FAULTHANDLER_HANDLERS: [FaultHandler; if cfg!(unix) {5} else {4}] = unsafe {
        [
            #[cfg(unix)]
            FaultHandler {
                signum: libc::SIGBUS,
                enabled: false,
                name: "Bus error",
                previous: std::mem::zeroed(),
            },
            FaultHandler {
                signum: libc::SIGILL,
                enabled: false,
                name: "Illegal instruction",
                previous: std::mem::zeroed(),
            },
            FaultHandler {
                signum: libc::SIGFPE,
                enabled: false,
                name: "Floating-point exception",
                previous: std::mem::zeroed(),
            },
            FaultHandler {
                signum: libc::SIGABRT,
                enabled: false,
                name: "Aborted",
                previous: std::mem::zeroed(),
            },
            FaultHandler {
                signum: libc::SIGSEGV,
                enabled: false,
                name: "Segmentation fault",
                previous: std::mem::zeroed(),
            },
        ]
    };

    /// faulthandler_handlers[]
    /// Number of fatal signals
    const FAULTHANDLER_NSIGNALS: usize = FAULTHANDLER_HANDLERS.len();
```

You might also want a `FaultHandler::default()` implementation to same some repetition in the future.

youknowone: Thanks! applied it

PR #6413, opened 2025-12-12
Tidy `codegen::ir::BlockIdx` api

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Streamlined compiler/internal codegen types and public interfaces to unify block identifiers and improve consistency across the code generation module. This reduces internal surface area and aims to make future maintenance and API evolution safer without changing runtime behavior.

### DIFF
--- a/crates/codegen/src/ir.rs
+++ b/crates/codegen/src/ir.rs
@@ -34,7 +34,7 @@ pub struct BlockIdx(u32);
 impl BlockIdx {
     pub const NULL: Self = Self::new(u32::MAX);
 
-    /// Creates a new instace of [`BlockIdx`] from a [`u32`].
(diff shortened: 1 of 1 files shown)


ShaharNaveh:
> Did you have any motivation to work on it? e.g. Was this not working good for your another patch?

IMO this looks cleaner, but not something critical. Generally, I prefer to not have a tuple struct with a `pub` item inside, this makes it easy to bypass any validation that you might want to add in your `TryFrom` or `::new` method, we can always have `new_unchecked` that does what `BlockIdx(val)` do, (in this example every u32 is valid, but for other structs it might not be always the case)

ShaharNaveh:
@youknowone https://www.howtocodeit.com/guides/ultimate-guide-rust-newtypes this is an excellent article imo, he explains my motivations better lol

PR #6415, opened 2025-12-12
Fix fix_test.py

youknowone:
with better collecting, async support and win32

I'd like to rename the script to `mark_test_failures` or something like that to shows better what it is 

## Summary by CodeRabbit

* **Chores**
  * Improved test automation script with enhanced reliability, better handling of test identification, and cross-platform compatibility improvements.

### DIFF
--- a/scripts/fix_test.py
+++ b/scripts/fix_test.py
@@ -5,16 +5,20 @@ As well as marking the test with a TODO comment.
 
 How to use:
 1. Copy a specific test from the CPython repository to the RustPython repository.
-2. Remove all unexpected failures from the test and skip the tests that hang
-3. Run python ./scripts/fix_test.py --test test_venv --path ./Lib/test/test_venv.py or equivalent for the test from the project root.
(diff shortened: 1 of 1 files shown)


scripts/fix_test.py
```diff
@@ -58,85 +62,87 @@ def parse_results(result):
...
+            for n in node:
                 match n:
                     case ast.ClassDef():
                         if len(test) == 2 and test[0] == n.name:
-                            # look through body for function def
-                            for i, fn in enumerate(n.body):
+                            for fn in n.body:
                                 match fn:
-                                    case ast.FunctionDef():
+                                    case ast.FunctionDef() | ast.AsyncFunctionDef():
```
ShaharNaveh, line 118: We should consider to make this script to generate the json payload `lib_updater` is expecting, and (and use its python API to patch the files). both are doing the same thing (ish)

youknowone: Good idea

PR #6417, opened 2025-12-12
Add regression test for #4505

### LINKED ISSUE #4505: `code.co_names` leaks a `__doc__` entry
when it shouldn't. 

I was able to repro this in cases where a nested function is defined, i.e:

```python
def foo():
    def inner():
        pass
```

In RustPython:

```python
>>>>> foo.__code__.co_names
('__doc__',)
```

while in Python:

```python
>>> foo.__code__.co_names
()
```

and also when compiling a module from a source string:

```python
stmts = """
import blah
 
def foo():
    pass
"""
code = compile(stmts, "<test>", "exec")
```

In RustPython:

```python
>>>>> code.co_names
('blah', '__doc__', 'foo')
```

while in Python:

```python
>>> code.co_names
('blah', 'foo')
```

not entirely sure how easy the fix might be.

ShaharNaveh:
I cannot reproduce this on https://github.com/RustPython/RustPython/commit/75dcf8042e8bbe69312c08a7d45118af9807f3a5.
Closes #4505

## Summary by CodeRabbit

* **Tests**
  * Added regression tests for builtin type handling to improve code reliability and ensure proper behavior of nested function definitions and code object introspection.

### DIFF
--- a/crates/codegen/src/compile.rs
+++ b/crates/codegen/src/compile.rs
@@ -122,7 +122,7 @@ pub struct CompileOpts {
 
 #[derive(Debug, Clone, Copy)]
 struct CompileContext {
-    loop_data: Option<(ir::BlockIdx, ir::BlockIdx)>,
+    loop_data: Option<(BlockIdx, BlockIdx)>,
     in_class: bool,
     func: FunctionContext,
 }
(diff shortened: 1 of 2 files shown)


youknowone:
Oh, I believe this is fixed in #6390 or #6392 

Could you add it to `builtin_type.py`? This is actually specially handled by type.

extra_tests/snippets/builtin_type.py
```diff
@@ -111,8 +111,6 @@ class D:
 with assert_raises(TypeError):
     del int.__qualname__
 
-from testutils import assert_raises
```
ShaharNaveh, line 114: Maybe we should also run `ruff check --select I` in our CI 🤔

youknowone: how about just put it in auto-format and add some files to blocklist

youknowone (approved):

PR #6424, opened 2025-12-12
refactor: small improvement in cp862.py

8of4:
Minor readability and safety improvement.


ShaharNaveh:
@nabil-devs It looks like the wrong content was committed, can you please take a look?

Also, the contents under `Lib/` is copied from official CPython releases. If you want to submit your patch, you should do it upstream at https://github.com/python/cpython

youknowone:
ref: https://github.com/tikv/pd/pull/10059

ShaharNaveh:
> ref: [tikv/pd#10059](https://github.com/tikv/pd/pull/10059)

Yeah, looks like a bot:/

Closing...

PR #6428, opened 2025-12-12
Update `test_sys.py` from 3.13.11

ShaharNaveh:
## Summary by CodeRabbit

* **New Features**
  * Added a new sys module function to query Global Interpreter Lock (GIL) status. Users can now check the GIL implementation state programmatically.

### DIFF
--- a/Lib/test/test_sys.py
+++ b/Lib/test/test_sys.py
@@ -206,6 +206,7 @@ class SysModuleTest(unittest.TestCase):
     def tearDown(self):
         test.support.reap_children()
 
+    @unittest.expectedFailure  # TODO: RUSTPYTHON
     def test_exit(self):
         # call with two arguments
         self.assertRaises(TypeError, sys.exit, 42, 42)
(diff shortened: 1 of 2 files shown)


Lib/test/test_sys.py
```diff
@@ -995,8 +1179,14 @@ def test_getallocatedblocks(self):
         c = sys.getallocatedblocks()
         self.assertIn(c, range(b - 50, b + 50))
 
-    # TODO: RUSTPYTHON, AtExit.__del__ is not invoked because module destruction is missing.
-    @unittest.expectedFailure
+    @unittest.expectedFailure # TODO: RUSTPYTHON; AssertionError: False is not true
```
ShaharNaveh: It seems like 
```py
sysconfig.get_config_var('Py_GIL_DISABLED')
```
returns `None` (should return `1`)

youknowone:
Currently `fn exit` in sys.rs is not checking the range of exit code. Checking it and return -1 for overflow will fix the failing test_exit

PR #6410, opened 2025-12-12
Update `test/libregrtest` to CPython 3.13.10

fanninpm:
- **Update test/libregrtest to CPython 3.13.10**
- **Re-add try block around platform and encodings info**
- **Disable `WindowsLoadTracker` on Windows**
- **Disable `struct._clearcache()`**
- **Disable `saved_test_environment` context manager**
- **Disable `support.record_original_stdout()`**

- [x] `faulthandler.dump_traceback_later()`
- [x] `inspect._shadowed_dict_from_weakref_mro_tuple()` #5377

## Summary by CodeRabbit

* **Chores**
  * Improved CI to detect and verify post-test environment pollution with repeated, OS-specific checks (Linux/macOS/Windows).
  * Added environment-variable toggles to enable or skip pollution verification for relevant jobs.
  * Introduced retry loops (up to 10 attempts), per-test and overall timeouts, and clearer guidance when pollution persists.
  * Normalized CI command formatting and updated minor consistency/formatting details for stability.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6435, opened 2025-12-13
Update `test_smtpnet.py` from 3.13.11

ShaharNaveh:
Related to: #6410


fanninpm:
It seems like `Lib/smtplib.py` also needs to be updated.

PR #6453, opened 2025-12-16
Skip flakey test

ShaharNaveh:
That test passed at https://github.com/RustPython/RustPython/pull/6452 which only added a test to `extra_tests`

### DIFF
--- a/Lib/test/test_subprocess.py
+++ b/Lib/test/test_subprocess.py
@@ -1770,8 +1770,7 @@ class RunFuncTestCase(BaseTestCase):
         res = subprocess.run(args)
         self.assertEqual(res.returncode, 57)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+    @unittest.skipIf(mswindows, 'TODO: RUSTPYTHON; Flakey')
     @unittest.skipUnless(mswindows, "Maybe test trigger a leak on Ubuntu")
     def test_run_with_an_empty_env(self):
         # gh-105436: fix subprocess.run(..., env={}) broken on Windows


Lib/test/test_subprocess.py
```diff
@@ -1770,8 +1770,7 @@ def test_run_with_pathlike_path_and_arguments(self):
         res = subprocess.run(args)
         self.assertEqual(res.returncode, 57)
 
-    # TODO: RUSTPYTHON
-    @unittest.expectedFailure
+    @unittest.skipIf(mswindows, 'TODO: RUSTPYTHON; Flakey')
```
fanninpm, line 1773:
```suggestion
    @unittest.skip("TODO: RUSTPYTHON, Flakey")
```
since the very next line is `@unittest.skipUnless(mswindows, "Maybe test trigger a leak on Ubuntu")`

ShaharNaveh: If CPython decides in a future version to test it on Linux/Darwin, I'd rather not skip it.

Also, regarding the convention of `TODO: RUSTPYTHON...` if it's separated by `;` or `,` it's not consistent ATM. but I went with `;` because that's what https://github.com/RustPython/RustPython/blob/adc2b0dbbe043095be17c037e3daa39b5351c96c/scripts/lib_updater.py does. I prefer the convention of `TODO(RUSTPYTHON): ...`, but our scripts would need to be adjusted accordingly. 

One more thing regarding the single/double quotes is that tool is using `ast.unparse` under the hood which generates string with single quotes (that's why I wrote it here with single quotes)

fanninpm: > If CPython decides in a future version to test it on Linux/Darwin, I'd rather not skip it.

Good point, noted.

> Also, regarding the convention of `TODO: RUSTPYTHON...` if it's separated by `;` or `,` it's not consistent ATM. but I went with `;` because that's what https://github.com/RustPython/RustPython/blob/adc2b0dbbe043095be17c037e3daa39b5351c96c/scripts/lib_updater.py does. I prefer the convention of `TODO(RUSTPYTHON): ...`, but our scripts would need to be adjusted accordingly.
> 
> One more thing regarding the single/double quotes is that tool is using `ast.unparse` under the hood which generates string with single quotes (that's why I wrote it here with single quotes)

In the past, when the `testdata` job in `.github/workflows/cron-ci.yaml` actually worked, a Liquid/Jekyll template would parse the resulting JSON file and [this line of code](https://github.com/RustPython/rustpython.github.io/blob/97beb2d3167f4e4aa8dd5d1a3eceeb7432e28d11/_layouts/regrtests_results.html#L110) would strip away the `TODO: RUSTPYTHON` note and, optionally, a comma or a colon. It seems like something that needs a facelift anyway.

By the way, the reason for #6410 is so that I can have the test suite output an XML file so that (hopefully) it could be converted to something that the RustPython website infrastructure can understand.

ShaharNaveh: I see, may I suggest another website infrastructure such as [Hugo](https://gohugo.io/) which has [many themes](https://themes.gohugo.io/), and supports multi-language blog posts&docs. We can auto generate a JSON file/markdown table(s) that the site will read and display. Hugo supports custom HTML OOB, so you can be as flexible as you want (and many other features)

youknowone (approved):

Lib/test/test_subprocess.py
youknowone: @fanninpm please merge it once the concern is resolved

fanninpm: This isn't a major concern, and this PR can be merged as-is. The better approach to address this would be to standardize around one delimiter and simplify the website code I linked above.

PR #6465, opened 2025-12-20
Make CI test Apple Intel

youknowone:
Now the default macos-latest runner is running on AppleSilicon

## Summary by CodeRabbit

* **Chores**
  * Updated CI/CD configuration for Intel-based macOS builds.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -153,13 +153,13 @@ jobs:
           cargo run --manifest-path example_projects/frozen_stdlib/Cargo.toml
         if: runner.os == 'Linux'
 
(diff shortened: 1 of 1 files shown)


ShaharNaveh:
What's the motivation for this change?

youknowone:
The original intent of this task was test compiling apple silicon target on x86_64 CI runner.
But now the default CI runner is changed to Apple Silicon. So the current task is duplication of the default build,

ShaharNaveh:
> The original intent of this task was test compiling apple silicon target on x86_64 CI runner. But now the default CI runner is changed to Apple Silicon. So the current task is duplication of the default build,

Oh, ty for the explanation

PR #6483, opened 2025-12-24
Update `shelve.py` from 3.13.11

ShaharNaveh:
related to #6410

### DIFF
--- a/Lib/shelve.py
+++ b/Lib/shelve.py
@@ -56,7 +56,7 @@ entries in the cache, and empty the cache (d.sync() also synchronizes
 the persistent dictionary on disk, if feasible).
 """
 
-from pickle import Pickler, Unpickler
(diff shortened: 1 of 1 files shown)


youknowone:
@ShaharNaveh did you also update shelve.py?

ShaharNaveh:
> @ShaharNaveh did you also update shelve.py?

I've tried not to. but I'll try to update it as well

youknowone:
Lib/X.py and Lib/test/test_X.py is pair. they are usually intended to be updated at the same time

ShaharNaveh:
> Lib/X.py and Lib/test/test_X.py is pair. they are usually intended to be updated at the same time

I'm aware, I've tried to minimize the changes in this PR as my only goal was to update `test_shelve.py`

PR #6524, opened 2025-12-25
Reduce usage `JumpIfTrueOrPop` & `JumpIfFalseOrPop` opcdes. Refactor `compile_compare`

ShaharNaveh:
Implements https://github.com/python/cpython/pull/102870

## Summary by CodeRabbit

* **Refactor**
  * Optimized internal compilation logic for comparison operations and boolean expressions, improving code maintainability.

* **Documentation**
  * Enhanced compiler documentation with reference materials and detailed explanatory comments.

### DIFF
--- a/crates/codegen/src/compile.rs
+++ b/crates/codegen/src/compile.rs
@@ -4115,10 +4115,14 @@ impl Compiler {
             self.compile_addcompare(op);
 
             // if comparison result is false, we break with this value; if true, try the next one.
+            /*
             emit!(self, Instruction::CopyItem { index: 1 });
             // emit!(self, Instruction::ToBool); // TODO: Uncomment this
             emit!(self, Instruction::PopJumpIfFalse { target: cleanup });
             emit!(self, Instruction::PopTop);
+            */
+
+            emit!(self, Instruction::JumpIfFalseOrPop { target: cleanup });
         }
 
         self.compile_expression(last_comparator)?;


ShaharNaveh:
This started as an attempt to remove `JumpIfTrueOrPop` & `JumpIfFalseOrPop` opcodes.
While it did work, some tests where failing mainly because of `dis` implementation difference, so I decided to pivot this PR to something else

crates/codegen/src/compile.rs
```diff
@@ -4044,92 +4046,97 @@ impl Compiler {
...
             // if comparison result is false, we break with this value; if true, try the next one.
-            if let Some((break_block, _)) = end_blocks {
-                emit!(
-                    self,
-                    Instruction::JumpIfFalseOrPop {
-                        target: break_block,
-                    }
-                );
-            }
+            /*
```
youknowone, line 4118: what was the blocker of this?

ShaharNaveh: That's the only reason why both opcodes stayed:/

This change caused regression for
https://github.com/RustPython/RustPython/blob/61ddd98b893072d9b067a953318d93a22a68d3ca/Lib/test/test_compile.py#L1073-L1081

Not because it found an `EXTENDED_ARG`, but because iterating over `dis.Bytecode(...)` gave an error

youknowone: if that's because our test_compile is an old one, replacing it will be also good.

youknowone (approved):

PR #6526, opened 2025-12-25
Align opcode names in `dis`

ShaharNaveh:
Also renamed `Instruction:: Pop` -> `Instruction::PopTop`

## Summary by CodeRabbit

* **Refactor**
  * Unified and standardized stack-cleanup instruction across the compiler, VM, and JIT. Behavior remains the same; this is an internal refactor to make stack-pop operations consistent and improve consistency of generated bytecode and disassembly output.

### DIFF
--- a/crates/jit/src/instructions.rs
+++ b/crates/jit/src/instructions.rs
@@ -547,10 +547,6 @@ impl<'a, 'b> FunctionCompiler<'a, 'b> {
                 }
             }
             Instruction::Nop => Ok(()),
-            Instruction::Pop => {
-                self.stack.pop();
-                Ok(())
-            }
             Instruction::PopBlock => {
                 // TODO: block support
                 Ok(())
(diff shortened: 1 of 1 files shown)


crates/compiler-core/src/bytecode.rs
```diff
@@ -1860,112 +1864,112 @@ impl Instruction {
             };
 
         match self {
-            BeforeAsyncWith => w!(BeforeAsyncWith),
+            BeforeAsyncWith => w!(BEFORE_ASYNC_WITH),
```
youknowone, line 1867: Is there reason to change the convention?

ShaharNaveh: Yes. Some tests are checking for the presents/absent of certain opcodes by name like: 
https://github.com/RustPython/RustPython/blob/61ddd98b893072d9b067a953318d93a22a68d3ca/Lib/test/test_compile.py#L1073-L1081

And pretty much all tests under `test_dis.py` (which we have a **VERY** outdated version of ATM), aligning the names of the opcodes just increases our test coverage

Issue #6532, opened 2025-12-25
subprocess.SubprocessError: Exception occurred in preexec_fn.


anki-code:
my_rc_file.xsh:
```xsh
@aliases.register
def _als():
    pass
```

```xsh
rustpython -V
# Python 3.13.0alpha (heads/master:40acd5529

~/.local/bin/rust-xonsh --rc my_rc_file.xsh
als
```
```xsh
Traceback (most recent call last):
  File "/Users/pc/.local/lib/rustpython3.13/site-packages/xonsh/procs/pipelines.py", line 177, in __init__
    proc = spec.run(pipeline_group=pipeline_group)
  File "/Users/pc/.local/lib/rustpython3.13/site-packages/xonsh/procs/specs.py", line 528, in run
    p = self._run_binary(kwargs)
  File "/Users/pc/.local/lib/rustpython3.13/site-packages/xonsh/procs/specs.py", line 585, in _run_binary
    raise xt.XonshError(e) from ex
  File "/Users/pc/.local/lib/rustpython3.13/site-packages/xonsh/procs/specs.py", line 546, in _run_binary
    p = self.cls(cmd, bufsize=bufsize, **kwargs)
  File "/Users/pc/.cargo/git/checkouts/rustpython-63090cdce5e18442/40acd55/crates/pylib/Lib/subprocess.py", line 1062, in __init__
    raise
  File "/Users/pc/.cargo/git/checkouts/rustpython-63090cdce5e18442/40acd55/crates/pylib/Lib/subprocess.py", line 1034, in __init__
    start_new_session, process_group)
  File "/Users/pc/.cargo/git/checkouts/rustpython-63090cdce5e18442/40acd55/crates/pylib/Lib/subprocess.py", line 1958, in _execute_child
    raise child_exception_type(err_msg)
subprocess.SubprocessError: Exception occurred in preexec_fn.
```

PR #6535, opened 2025-12-26
signal timer

youknowone:
related to #6532


youknowone:
patched libc is working great.
waiting for https://github.com/rust-lang/libc/pull/4890

PR #6609, opened 2025-12-31
Updated the importlib + nturl2path + pathlib libraries + associated tests - v3.13.10

terryluan12:


terryluan12:
(Also adding importlib, but that was mainly so a timed out test would rerun)

Lib/importlib/_bootstrap.py
```diff
@@ -527,7 +527,7 @@ def _load_module_shim(self, fullname):
 
     """
     msg = ("the load_module() method is deprecated and slated for removal in "
-          "Python 3.12; use exec_module() instead")
+           "Python 3.15; use exec_module() instead")
```
youknowone, line 530: @terryluan12 which version did you brought the library? 3.15 is not a production version, and we usually don't bring library from alpha version of cpython

youknowone: @terryluan12 I am sorry. I forgot to check the library version from the previous PRs. we currently target 3.13. 3.14 can be also fine if it works without problem. We usually also include the Python version of the library in the PR title.

terryluan12: It is 3.13.10; The message is just  is just stating it will be removed in v3.15

youknowone: Thank you for confirming!

youknowone (approved):
Thank you!

PR #6610, opened 2025-12-31
Updated the pkgutil + unittest libraries + the pkgutil, unittest and import tests - v.3.13.11

terryluan12:
I don't have access to Windows for now, so I've added comments to the top of the relevant `test_pathlib` files

### DIFF
--- a/Lib/test/test_import/__init__.py
+++ b/Lib/test/test_import/__init__.py
@@ -791,6 +791,7 @@ class ImportTests(unittest.TestCase, ExtraAssertions):
         finally:
             del sys.path[0]
 
+    @unittest.expectedFailure  # TODO: RUSTPYTHON; FileNotFoundError: [WinError 2] No such file or directory: 'built-in'
     @unittest.skipUnless(sys.platform == "win32", "Windows-specific")
     def test_dll_dependency_import(self):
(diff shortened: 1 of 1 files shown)


youknowone:
dbm is updated in #6616 
shelve will be updated in #6483

terryluan12:
> dbm is updated in #6616 shelve will be updated in #6483

Sounds good. Removed from the title. Is there anything else I should do?

Also as an FYI, `_test_multiprocessing.py` is failing now, so I'm updating `_test_multiprocessing.py` + the `multiprocessing` library

youknowone:
@terryluan12 Thank you so much! Could you also rebase this to remove test_import, which is already merged? And I recently updated multiprocessing to 3.13.11. Could you also check again if it is patching backward?

terryluan12:
@youknowone taking a look, I think I may completely refactor the history, as pathlib and importlib was already imported, which means quite a lot of refactoring would need to be done anyways. Can you clarify, do you mean importlib? Taking a look, it actually seems like importlib was updated, but not test_import

youknowone:
Using git reset to prev commit and adding only indented files will be helpful. Then cherrypick or rebase will be easy

youknowone (approved):
Thank you!

PR #6611, opened 2025-12-31
Updated the platform + plistlib libraries + associated tests v.3.13.10

terryluan12:


Lib/platform.py
```diff
@@ -1071,20 +1190,21 @@ def _sys_version(sys_version=None):
...
         version, buildno, builddate, buildtime, compiler = \
               match.groups()
-
-        # XXX: RUSTPYTHON support
-        if "RustPython" in sys_version:
-            name = "RustPython"
-        else:
-            name = 'CPython'
-
+        name = 'CPython'
```
ShaharNaveh: Please re-add this patch

terryluan12: Whoops, sorry about that! Just fixed. Thanks for the catch

PR #6633, opened 2026-01-03
Updated token + weakref libraries + weakref test - v3.13.11

terryluan12:


terryluan12:
I could not find a token test file

youknowone:
by searching `import token`, `test_tokenize` seems to test it

terryluan12:
> by searching `import token`, `test_tokenize` seems to test it

I saw that, & assumed that it would test out the tokenize library, but as I'm writing it out I'm realizing I should probably jointly update the two libraries. Will add it, thanks!

terryluan12:
So it seems a rewrite of tokenize has been done. the _tokenize function was moved to a precompiled library, and logic has been changed, so it's not as simple as keeping the older _tokenize version. Is it okay to merge, without making any tokenization migrations?

terryluan12:
It's a bit of a different error, but waiting until  #6635 is merged until debugging

Lib/test/test_ftplib.py
```diff
@@ -15,6 +15,7 @@
     import ssl
 except ImportError:
     ssl = None
+import sys
```
youknowone:
```suggestion
```

youknowone (approved):
👍

PR #6637, opened 2026-01-03
Assign opcode ids

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Internal bytecode instruction names and encodings were reorganized for clearer, strongly-discriminated layouts; this improves consistency in disassembly, error messages, and runtime handling without changing observable behavior.

* **Chores**
  * Opcode metadata generation was automated from source to keep tooling and printed opcode names synchronized.

### DIFF
--- a/Lib/test/test_dis.py
+++ b/Lib/test/test_dis.py
@@ -937,6 +937,7 @@ class DisTests(DisTestBase):
     def test_boundaries(self):
         self.assertEqual(dis.opmap["EXTENDED_ARG"], dis.EXTENDED_ARG)
 
+    @unittest.expectedFailure # TODO: RUSTPYTHON; AssertionError: 29 not less than or equal to 20
     def test_widths(self):
         long_opcodes = set(['JUMP_BACKWARD_NO_INTERRUPT',
                             'INSTRUMENTED_CALL_FUNCTION_EX'])


youknowone:
#6638 will change instruction a lot. Can this be aligned to that patch?

crates/compiler-core/src/bytecode.rs
```diff
@@ -676,460 +676,265 @@ pub type NameIdx = u32;
...
+    SetExcInfo = 134,
+    Subscript = 135,
+    // ===== Pseudo Opcodes (252+) ======
+    Jump {
+        target: Arg<Label>,
+    } = 252, // CPython uses pseudo-op 256
+    LoadClosure(Arg<NameIdx>) = 253, // CPython uses pseudo-op 258
+    LoadMethod {
+        idx: Arg<NameIdx>,
+    } = 254, // CPython uses pseudo-op 259
```
youknowone, line 956: We will actually put these ops on pseudo ops once we get them.
Planning to do it for next task.

crates/compiler-core/src/bytecode.rs
```diff
@@ -676,460 +676,265 @@ pub type NameIdx = u32;
...
     type Error = MarshalError;
 
     #[inline]
     fn try_from(value: u8) -> Result<Self, MarshalError> {
-        if value <= u8::from(LAST_INSTRUCTION) {
-            Ok(unsafe { core::mem::transmute::<u8, Self>(value) })
-        } else {
-            Err(MarshalError::InvalidBytecode)
+        match value {
+            1
```
youknowone: Why not the opposite way? listing reserved may be easier. Reserved ops doesn't have args, so using names is also easy like `Self::Reserved3 as u8`

crates/compiler-core/src/bytecode.rs
```diff
@@ -1857,13 +1662,7 @@ impl Instruction {
             | ForIter { target: l }
             | Break { target: l }
             | Continue { target: l }
-            | Send { target: l }
-            // Dummy jump instructions (placeholders)
-            | JumpBackward { target: l }
```
youknowone, line 1862: If we are going to implement these to replace our own old ones, I hesitate to remove it now to recover it back later.
We will not implement any other instruction in this slot anyway.

crates/compiler-core/src/bytecode.rs
ShaharNaveh: That, will be great:)
If you already have a local branch with the changes, can you please open a draft PR? I might have some inputs

crates/compiler-core/src/bytecode.rs
ShaharNaveh: Not sure I understand. What is the purposed code change?

---

I created this match arm with a simple script

crates/compiler-core/src/bytecode.rs
ShaharNaveh: My philosophy is that we can always add new code, but to have a placeholder code (that can be deleted) is just clutter

crates/compiler-core/src/bytecode.rs
youknowone: not yet. I will start one during weekend

crates/compiler-core/src/bytecode.rs
youknowone: I'd like to avoid using literal numbers in the code. That's not easy to read, and maintainers must not forget to run the 'simple script'. Using name prevent to run the script.

```
if (1..=Self::Resume as u8).contains(value) && ![Self::Reserved3 as u8, Self::Reserved17 as u8, Self::Reserved141 as u8...].contains(value) {
    Ok(core::mem::transmute::<u8, Self>(value))
} else {
    Err(InvalidBytecode)
}
```
This code will automatically error-catachable by breaking build if we change reserved141 to something other. (Though not working the inverse way)

crates/compiler-core/src/bytecode.rs
youknowone: What's the benefit to delete them? They are deletable, but no reason to do.

ShaharNaveh: IMO they are just noise, but I can re-add them, that's not an issue

crates/compiler-core/src/bytecode.rs
ShaharNaveh: But we have "holes" in our IDs, this can lead to undefined behavior.

I would love to use `num_enum` crate, but it doesn't support enum variants that holds a value:/

I do think we should rename the `Instruction ` enum to `Opcode`, and have an `Instruction ` struct that holds:
```rust
struct Instruction {
  opcode: Opcode,
  oparg: Oparg // Oparg enum
}
```
Like a typed `CodeUnit`.
But that's out of scope for this PR, and needs a further dissection

Lib/test/test_dis.py
```diff
@@ -937,6 +937,7 @@ def test_opname(self):
     def test_boundaries(self):
         self.assertEqual(dis.opmap["EXTENDED_ARG"], dis.EXTENDED_ARG)
 
+    @unittest.expectedFailure # TODO: RUSTPYTHON; AssertionError: 29 not less than or equal to 20
```
ShaharNaveh, line 940: Because `BINARY_OP_INPLACE_ADD_UNICODE` opcode was added, we are missing some truncation logic when formatting the `dis` output

crates/compiler-core/src/bytecode.rs
youknowone: Yeah, that could be. At the same time, it is also like TODO

crates/compiler-core/src/bytecode.rs
youknowone: Hmm, we had that form originally. The commits of that change or comments of the PR might have note that why we moved here. 
cc  @coolreader18

crates/vm/src/frame.rs
```diff
@@ -1779,50 +1779,8 @@ impl ExecutingFrame<'_> {
...
-            | bytecode::Instruction::Reserved146
-            | bytecode::Instruction::Reserved147
-            | bytecode::Instruction::Reserved148 => {
-                unreachable!(
-                    "placeholder instruction should not be executed: {:?}",
-                    instruction
-                )
+            _ => {
+                unreachable!("{instruction:?} instruction should not be executed")
             }
```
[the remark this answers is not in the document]
youknowone: @ShaharNaveh uuuuh, this is totally right. I immediately broken build by forgetting change arm during merge. listing all is better.

PR #6639, opened 2026-01-04
Update the zipfile + zipimport libraries + tests - v3.13.11

terryluan12:
Also includes updates to `sre_compile.py`, `sre_constants.py`, `sre_parse.py`

## Summary by CodeRabbit

* **Chores**
  * Updated internal engine version identifier.
  * Removed a deprecated flag option from the engine's public flags.
  * Added an automated script to regenerate engine constants from source definitions.

* **User-visible**
  * Pattern/regex formatting no longer displays the removed flag when showing compiled patterns.

### DIFF
--- a/Lib/test/test_zipfile/test_core.py
+++ b/Lib/test/test_zipfile/test_core.py
@@ -3434,6 +3434,7 @@ class TestExecutablePrependedZip(unittest.TestCase):
         self.assertIn(b'number in executable: 5', output)
 
 
+@unittest.skip("TODO: RUSTPYTHON shift_jis encoding unsupported")
 class EncodedMetadataTests(unittest.TestCase):
     file_names = ['\u4e00', '\u4e8c', '\u4e09']  # Han 'one', 'two', 'three'
     file_content = [


ShaharNaveh:
@terryluan12 can you please split this PR into multiple PRs? Unless there's a connection between all the 20 files. I'd rather review multiple smaller PRs rather than one big PR, this will help with merging what can already be merged and not blocked by other unrelated changes.

And tysm for your contributions<3

youknowone:
probably zipfile and zipimport is related. If splitting re is possible, that will be the best.

terryluan12:
> probably zipfile and zipimport is related. If splitting re is possible, that will be the best.

Sounds good! Just done so on #6648

terryluan12:
(Btw, double checking if skip is needed)

Lib/test/test_zipfile/test_core.py
```diff
@@ -3418,25 +3483,29 @@ def _test_read(self, zipfp, expected_names, expected_content):
             self.assertEqual(info.file_size, len(content))
             self.assertEqual(zipfp.read(name), content)
 
+    @unittest.skip('TODO: RUSTPYTHON')
```
youknowone: what's broken here?

Lib/test/test_except_star.py
```diff
@@ -445,6 +445,7 @@ def test_nested_in_loop(self):
 
 
 class TestExceptStarReraise(ExceptStarTest):
+    @unittest.skip("TODO: RUSTPYTHON")
```
youknowone: it's werid that zip affect except syntax. let me check

youknowone: Please rebase this patch to main and remove every skip.
The failing is related to bytecode changes.
If you keep experience problem, running this command might help: 
`rm -r target/debug/build/rustpython-* && find . | grep -E "\.pyc$" | xargs rm -r`

youknowone (changes requested):

Lib/test/test_except_star.py
terryluan12: Just done

Lib/test/test_zipfile/test_core.py
```diff
@@ -3418,25 +3483,29 @@ def _test_read(self, zipfp, expected_names, expected_content):
...
     def test_read_without_metadata_encoding(self):
         # Read the ZIP archive without metadata_encoding
         expected_names = [name.encode('shift_jis').decode('cp437')
                           for name in self.file_names[:2]] + self.file_names[2:]
         with zipfile.ZipFile(TESTFN, "r") as zipfp:
             self._test_read(zipfp, expected_names, self.file_content)
 
+    @unittest.skip('TODO: RUSTPYTHON')
```
youknowone: These skips too. Actually, I'd like to entirely remove skip in this patch.
Let's see what's happening without skip.

Usually we avoid skip because it hides bugs and problems. If it still fails, I will append the reason what justify skip

Lib/test/test_zipfile/test_core.py
```diff
@@ -3369,16 +3434,17 @@ def test_execute_zip64(self):
         self.assertIn(b'number in executable: 5', output)
 
 
-# TODO: RUSTPYTHON
 @unittest.skip("TODO: RUSTPYTHON shift_jis encoding unsupported")
```
youknowone, line 3437: @terryluan12 Because this is a setUp issue, skipping the entire tests as previous version is better than marking every functions as skip. They are caused by the single reason

PR #6647, opened 2026-01-04
Make use of `Unary` opcodes

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Unary operations (plus, minus, invert, not) are now separate, explicit instructions rather than a single generic unary instruction, improving execution consistency across runtimes.
  * Unary plus is routed through a dedicated intrinsic path for uniform behavior.

* **Documentation / Tests**
  * Tests and docs updated to reflect the explicit unary variants and revised runtime behavior.

### DIFF
--- a/crates/stdlib/src/opcode.rs
+++ b/crates/stdlib/src/opcode.rs
@@ -4,9 +4,8 @@ pub(crate) use opcode::make_module;
 mod opcode {
     use crate::vm::{
         AsObject, PyObjectRef, PyResult, VirtualMachine,
-        builtins::{PyBool, PyInt, PyIntRef, PyNone},
+        builtins::{PyInt, PyIntRef},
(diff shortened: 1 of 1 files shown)


crates/compiler-core/src/bytecode.rs
```diff
@@ -1091,10 +1091,6 @@ pub enum Instruction {
     SetExcInfo,
     // 139: SUBSCRIPT
     Subscript,
-    // 140: UNARY_OP (combines UNARY_*)
-    UnaryOperation {
-        op: Arg<UnaryOperator>,
-    },
     // 141-148: Reserved (padding to keep RESUME at 149)
```
youknowone: Let’s keep this line by replacing it with Reserved140 instead of removing it.

PR #6661, opened 2026-01-06
Fix set in-place operators with self argument

### LINKED ISSUE #3881: `set` `intersection_update`, `difference_update`, `symmetric_difference_update` should check self is equal to given arg

HueCodes:
Fixes #3881
close #3912
close #4424
close #4708

  When a set performs in-place operations with itself (`s &= s`, `s -= s`, `s ^= s`), the operators now handle the self-reference case correctly:

  - `__iand__`: no-op (intersection with self is unchanged)
  - `__isub__`: clear (difference with self is empty)
  - `__ixor__`: clear (symmetric difference with self is empty)

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected behavior of in-place set operations (&=, -=, ^=) when a set operates on itself, preventing unintended modifications.

* **Tests**
  * Added test cases validating in-place set operations when operating on the same set instance.

### DIFF
--- a/crates/vm/src/builtins/set.rs
+++ b/crates/vm/src/builtins/set.rs
@@ -24,6 +24,7 @@ use crate::{
     vm::VirtualMachine,
 };
 use alloc::fmt;
+use core::borrow::Borrow;
 use core::ops::Deref;
 use rustpython_common::{
     atomic::{Ordering, PyAtomic, Radium},
@@ -719,8 +720,10 @@ impl PySet {
     }
 
     fn __iand__(zelf: PyRef<Self>, set: AnySet, vm: &VirtualMachine) -> PyResult<PyRef<Self>> {
-        zelf.inner
-            .intersection_update(core::iter::once(set.into_iterable(vm)?), vm)?;
+        if !set.is(zelf.as_object()) {
+            zelf.inner
+                .intersection_update(core::iter::once(set.into_iterable(vm)?), vm)?;
+        }
         Ok(zelf)
     }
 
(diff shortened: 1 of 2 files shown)


crates/vm/src/builtins/set.rs
```diff
@@ -1305,6 +1315,10 @@ impl AnySet {
         obj.fast_isinstance(ctx.types.set_type) || obj.fast_isinstance(ctx.types.frozenset_type)
     }
 
+    fn is_same_object(&self, other: &PyObject) -> bool {
+        self.object.is(other)
+    }
```
youknowone: We have a mechanism to do this.

`is` is not a method of `PyObject`, but an object of`AsObject`. And PyObject is implementing AsObject through Borrow<PyObject>.

AnySet can do the same thing.
`impl Borrow<PyObject> for AnySet {}`
will enable AsObject for AnySet. Then `set.is_same_object(zelf.as_object())` above can be simply written as `set.is(zelf.as_object())`

HueCodes:
Done.

PR #6675, opened 2026-01-09
Remove `Reverse` bytecode

ShaharNaveh:
## Summary by CodeRabbit

* **Chores**
  * Removed a legacy bytecode instruction and its runtime handling, simplifying the instruction set and trimming related formatting/diagnostic paths.
  * Simplified opcode metadata/name generation logic by removing redundant code paths.
  * CI validation tightened to fail when generated metadata is out of date.

* **Breaking Changes**
  * Potential bytecode compatibility impact for programs or artifacts that relied on the removed instruction; affected items should be regenerated or recompiled.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -334,7 +334,7 @@ jobs:
       - name: Ensure Lib/_opcode_metadata is updated
         run: |
           python scripts/generate_opcode_metadata.py
-          if [ -z "$(git status --porcelain)" ]; then
+          if [ -n "$(git status --porcelain)" ]; then
             exit 1
           fi
 


crates/compiler-core/src/bytecode.rs
```diff
@@ -941,11 +941,8 @@ pub enum Instruction {
     } = 130,
     JumpIfNotExcMatch(Arg<Label>) = 131,
     LoadClassDeref(Arg<NameIdx>) = 132,
-    Reverse {
-        amount: Arg<u32>,
-    } = 133,
-    SetExcInfo = 134,
-    Subscript = 135,
+    SetExcInfo = 133,
+    Subscript = 134,
```
youknowone:
```suggestion
    SetExcInfo = 134,
    Subscript = 135,
```

ShaharNaveh: Hmm, any reason why to revert this change?
this will lead to UB when doing: `Instruction::try_from(133)`

youknowone: we don't need to bytecode compatibility for every commit. Because we didnt' emit Reverse before, keeping the old numbers don't  break bytecode compatibility. But this change breaks.
Yeah, we don't keep bytecode compatibility that much. But we can do it when we necessarily do.

ShaharNaveh:
> **Code has been automatically formatted**
> 
> The code in this PR has been formatted using `cargo fmt --all`. Please pull the latest changes before pushing again:
> 
> ```shell
> git pull origin bytecode-remove-reverse
> ```

Yay #6677 works

youknowone:
Great! The message looks like need to be changed. it is not by `fmt --all`

PR #6693, opened 2026-01-10
Move `Instruction` enum to its own file

ShaharNaveh:
There are no logic changes in this PR.
In a future PR I plan on moving the oparg enums to their own file as well

## Summary by CodeRabbit

* **Refactor**
  * Reorganized bytecode instruction definitions into a dedicated module for improved code maintainability.
  * Renamed instruction variant from `CopyItem` to `Copy` for consistency.
  * Enhanced instruction handling and display logic in the virtual machine.


crates/compiler-core/src/bytecode.rs
```diff
@@ -1087,8 +741,6 @@ impl Deref for CodeUnits {
     }
 }
 
-use self::Instruction::*;
```
ShaharNaveh, line 1088: Due to the removal of this line I've decided to rename `CopyItem` -> `Copy`

youknowone (approved):
@ShaharNaveh I feel like Arg is a part of Instruction. If it needs to be splited, putting in instruction.rs makes more sense to me than putting it on its own sub module.

ShaharNaveh:
> @ShaharNaveh I feel like Arg is a part of Instruction. If it needs to be splited, putting in instruction.rs makes more sense to me than putting it on its own sub module.

No worries, I'll append it in a future PR

PR #6704, opened 2026-01-11
Improve `json.loads` performance

moreal:
First, to minimize movement between Python and Rust, I made it parse with Rust whenever possible. This improved performance from 8.77s on the main branch (by my local benchmark) to 7.37s. Although still slower compared to CPython's 0.2s, it's about a 18% improvement. The execution method was as follows, and I ran it about 3 times, calculating based on the highest value.

~~~~
cargo build --release
time ./target/release/rustpython json_loads.py
~~~~

Additionally, since JSON often has duplicate keys, I also implemented CPython's method of reusing them. However, when measured separately, this didn't show any execution time benefits—it's only expected to help with memory efficiency.

-------

While profiling by taking multiple flamegraph spans, I discovered that scanstring was taking the longest time. The reason was that operations like chars().nth() are O(N) and severely slow. I avoided this performance bottleneck by also operating with byte indices alongside character indices.

Compared to CPython taking 0.2 seconds, it now takes around 0.4 to 0.5 seconds—a huge improvement! This is over 16 times faster!

For the original motivation, `pip install`, what used to take 5 seconds for the `black` package now takes 1.61 seconds. Even when built for debug targets, what used to take 171 seconds now finishes in under 20 seconds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Stronger detection and clearer errors for malformed JSON, invalid escapes, out-of-range positions, and trailing-comma cases.

* **Improvements**
  * More accurate, character-aware error messages for strings and Unicode escapes (including surrogate handling).
  * Reduced allocations for object keys and overall parsing robustness and performance.

* **New Features**
  * Accepts an escaped "/" in strings and improves fallback parsing behavior.

### DIFF
--- a/crates/stdlib/src/json.rs
+++ b/crates/stdlib/src/json.rs
@@ -7,7 +7,7 @@ mod _json {
     use crate::vm::{
         AsObject, Py, PyObjectRef, PyPayload, PyResult, VirtualMachine,
         builtins::{PyBaseExceptionRef, PyStrRef, PyType},
-        convert::{ToPyObject, ToPyResult},
+        convert::ToPyResult,
         function::{IntoFuncArgs, OptionalArg},
         protocol::PyIterReturn,
         types::{Callable, Constructor},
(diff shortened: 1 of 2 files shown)


moreal:
Thank you for resolving the CI issue!

crates/stdlib/src/json.rs
```diff
@@ -189,28 +223,443 @@ mod _json {
...
-                zelf.parse(
-                    chars.as_str(),
-                    pystr.clone(),
-                    idx,
-                    zelf.to_owned().into(),
-                    vm,
-                )
-                .and_then(|x| x.to_pyresult(vm))
-            }
+                match s.char_indices().nth(char_idx) {
```
youknowone, line 645: This part is looking not very performance friendly.
Unlike CPython, our string is not array of chars. 
Can these functions written with only byte_idx instead of char_idx?

moreal: (Please let me know if I misunderstood)

As with tracking byte_idx in this PR, going through the `Chars` iterator has a very negative impact on performance. However, in the Callable implementation you commented on, char_idx has to be used because it's called using char_idx from the Python side.

So this is the part where we convert char_idx to byte_idx before starting parsing. Is there a better way to convert to byte_idx than using `char_indices().nth()`?

youknowone: Oh, thank you for explanation! I was misunderstanding.

youknowone (approved):
Thank you!

PR #6709, opened 2026-01-12
Added a test updater tool

terryluan12:
Hey!

This is adding a small script to:
A. Make better use of the `lib_updater.py` to better update files en masse (along with automatically marking tests as skip or expectedFailure)
B. Test out existing tests which are annotated `@unittest.skip` to see if they can be replaced with `@unittest.expectedFailure`

It is currently still very janky, but it does it's job alright-ish, and while there still is some amount of manual work that needs to be done, it's mainly just checking + undoing actions in git (rather than manual replacing + checking)

I'm probably not going to add too much more (since like I mentioned before it does it's job alright + diminishing returns), but please let me know if it's something of interest to be merged. If so, I’ll likely make some adjustments to make it more platform-agnostic and tidy up the formatting, etc.

## Summary by CodeRabbit

* **Chores**
  * Added a test-management utility to automate syncing between test suites with a comprehensive CLI, parallel processing, safety checks, selective copying/updating, configurable skip handling, dynamic failure annotation, and backup/restore of skip metadata.

### DIFF
--- a/scripts/update_tests.sh
+++ b/scripts/update_tests.sh
@@ -19,7 +19,7 @@ Options:
   -h/--help                  Show this help message and exit
 
 Example Usage: 
-$0 -c ~/cpython -r - . -t 300   # Updates all non-updated tests with a timeout value of 300 seconds
+$0 -c ~/cpython -r . -t 300   # Updates all non-updated tests with a timeout value of 300 seconds
 $0 -c ~/cpython -r . -u -j 5    # Updates all non-updated tests + copies files not in cpython into rpython, with maximum 5 processes active at a time
 $0 -c ~/cpython -r . -a         # Updates all non-updated tests + annotates with @unittest.expectedFailure/@unittest.skip
 $0 -r . -s                      # For all current tests, check if @unittest.skip can be downgraded to @unittest.expectedFailure
@@ -56,11 +56,11 @@ ignored_libraries=("multiprocessing" "concurrent")
 while [[ $# -gt 0 ]]; do
     case "$1" in
         -c|--cpython-path)
-            cpython_path="$2"
+            cpython_path="$2/Lib/test"
             shift 2
             ;;
         -r|--rpython-path)
-            rpython_path="$2"
+            rpython_path="$2/Lib/test"
             shift 2
             ;;
         -u|--copy-untracked)
@@ -93,10 +93,13 @@ while [[ $# -gt 0 ]]; do
             ;;
     esac
 done
+# -------------------------------------- Constants ------------------------------------- #
+RUSTPYTHON_POSSIBLE_SKIP_RE="@unittest.skip.*\([\"']TODO:\s*RUSTPYTHON.*[\"']\)"
+RUSTPYTHON_CANONICAL_SKIP_RE="@unittest.skip\('TODO: RUSTPYTHON; .*'\)"
+RUSTPYTHON_CANONICAL_EX_FAILURE="@unittest.expectedFailure # TODO: RUSTPYTHON"
+RUSTPYTHON_CANONICAL_EX_FAILURE_RE="\s@unittest\.expectedFailure # TODO: RUSTPYTHON.*"
 
-cpython_path="$cpython_path/Lib/test"
-rpython_path="$rpython_path/Lib/test"
-
+# --------------------------------- Updating functions --------------------------------- #
 
 update_tests() {
     local libraries=("$@")
@@ -135,11 +138,13 @@ update_test() {
     fi
 }
 
+# --------------------------------- Downgrade Skips functions --------------------------------- #
+
 check_skips() {
     local libraries=("$@")
     for lib in "${libraries[@]}"
     do
-        if grep -qiE "@unittest.skip.*\('TODO:\s*RUSTPYTHON.*'\)" "$rpython_path/$lib"; then
+        if grep -qiE "$RUSTPYTHON_POSSIBLE_SKIP_RE" "$rpython_path/$lib"; then
             sem
             check_skip "$lib" &
         else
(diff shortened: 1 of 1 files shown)


terryluan12:
> Currently `fix_test.py` have similar categorized feature that appending `expectedFailure` for failing tests. I am not sure what you meant by A yet, but B sounds good. I wish it can be is integrated as `fix_test.py --try-unskip`.
> 
> Could you share examples runs of this script what it solves? The command and `git diff` after the command run.

So to clarify, what I mean by A, is that it can run updates on multiple files at once. As I could tell from the documentation, `lib_updater.py` only worked on one file at a time. 

In the ideal case* you could run `update_tests.sh -r <rust_path> -c <c_path> <folder_path>` on a folder, and update all tests in the folder. Or you can just run `update_tests.sh -r <rust_path> -c <c_path> -u` and it will update all CPython tests *and* copy all untracked tests into RustPython.

Also for that same reason, it's different from `fix_test.py`. Also though, it's difference, since for each test that is annotated, it loops until all tests succeed.  

For example, it first tries to add `@unittest.expectedFailure` and then if it fails, then it adds `@unittest.skip` (no support for `skipIf`/`skipUnless`, so that's a limitation)


* I realize now, I did not add that specific functionality, but if merging, I probably will. The main use case I was just going to use it for was updating multiple specific tests, or all tests

terryluan12:
For example uses cases, see below:
```
update_tests.sh -c ~/cpython -r - . -t 300   # Updates all non-updated tests with a timeout value of 300 seconds
update_tests.sh -c ~/cpython -r . -u -j 5    # Updates all non-updated tests + copies files not in cpython into rpython, with maximum 5 processes active at a time
update_tests.sh -c ~/cpython -r . -a         # Updates all non-updated tests + annotates with @unittest.expectedFailure/@unittest.skip
update_tests.sh -r . -s                      # For all current tests, check if @unittest.skip can be downgraded to @unittest.expectedFailure
```

terryluan12:
In terms of the `git diff`, it very much varies. But from running it (the downgrading feature), I've already verified that it's already downgraded all skips in `test_bdb.py` down to expectedFailures, removed all skips from test_bytes (by technicality), and downgraded some in `test_enum.py`

Like I said it's very very janky, and there are a lot of issues/limitations that it has:

* It deletes skips for intermittent failures
* It replaces @unittest.skipIf with @unittest.skip
* If there are multiple tests with the same name in a file, then it adds the decorator to all of the tests

But it's good enough to at least get candidates for downgrading/removal (and makes the process for downgrading much smoother) imo

youknowone:
I am good for adding automatic unskip, but oppose to automatic skip.
expectedFailure is a normal thing. That means lack of feature compatibility. But skip is a bad sign. We have to understand what's happening and why adding skip. We rarely adds skip. Blindly adding skip is not the best idea to prevent regressions. (That's why we are so picky about skip but not about expectedFailures)
If there is an auto-skip feature, that must be named like `--add-skip-and-i-will-fully-review-the-added-skip-and-leave-enough-description`, never like `--try-unskip`.

The point of `fix_test.py` is, not it is same as this new script, but the purpose and use case is shared. Having multiple tools doing similar task is not the best situation. (tbh I already feel the same thing about fix_test.py and lib_updater.py; they must be merged into single tool in future)

It will be also great if the script is a python script, not a bash because we have to support also windows.

terryluan12:
To clarify, there are two ways that the skips are applied; 
Either:
1. The code has been run once, an expectedFailure has been applied to the test, and it's still failing, or
2. The test is causing the test to crash (We can tell this because the program just stops in the middle of testing)

Are these situations ok to annotate a test as skip, or should I prevent it anyways?

I can make the change for porting it to Python, but it's probably something I'll tackle later as well as merging `fix_test.py` and `lib_updater.py` (I'll close the PR once this conversation is over)

terryluan12:
Also, one last thing:

Using this, there are a lot of tests that can be changed. Should I make a separate PR for each test/group of tests, or just make one single PR?

(So far there are 55 tests that are changed, in reality, there will be less that are true positives)

ShaharNaveh:
> To clarify, there are two ways that the skips are applied; Either:
> 
> 1. The code has been run once, an expectedFailure has been applied to the test, and it's still failing, or
> 2. The test is causing the test to crash (We can tell this because the program just stops in the middle of testing)
> 
> Are these situations ok to annotate a test as skip, or should I prevent it anyways?
> 
> I can make the change for porting it to Python, but it's probably something I'll tackle later as well as merging `fix_test.py` and `lib_updater.py` (I'll close the PR once this conversation is over)

We are skipping tests in one of 4 cases (from most common to most rare):
1. Test hangs forever
2. Test causes panic over rust side
3. Test is flaky (sometimes it passes, sometimes doesn't)
4. Test takes irregular amount of time, and there's no way to tune it down a bit

ShaharNaveh:
> Also, one last thing:
> 
> Using this, there are a lot of tests that can be changed. Should I make a separate PR for each test/group of tests, or just make one single PR?

> (So far there are 55 tests that are changed, in reality, there will be less that are true positives)

Yes! That way we can start benefiting from the tool even before it gets merged. No need to do 1 PR for each test

terryluan12:
> > To clarify, there are two ways that the skips are applied; Either:
> > 
> > 1. The code has been run once, an expectedFailure has been applied to the test, and it's still failing, or
> > 2. The test is causing the test to crash (We can tell this because the program just stops in the middle of testing)
> > 
> > Are these situations ok to annotate a test as skip, or should I prevent it anyways?
> > 
> > I can make the change for porting it to Python, but it's probably something I'll tackle later as well as merging `fix_test.py` and `lib_updater.py` (I'll close the PR once this conversation is over)
> 
> We are skipping tests in one of 4 cases (from most common to most rare):
> 1. Test hangs forever
> 2. Test causes panic over rust side
> 3. Test is flaky (sometimes it passes, sometimes doesn't)
> 4. Test takes irregular amount of time, and there's no way to tune it down a bit

Nice, good to know! 
Re: 1, its addressed, through the use of a timeout, thats customizable.
2 and 3 are also addressed (see above comment)
If I'm understanding what you mean by 4, it's also addressed with the timeout (although the test treats it as a hang)

terryluan12:
> > To clarify, there are two ways that the skips are applied; Either:
> > 
> > 1. The code has been run once, an expectedFailure has been applied to the test, and it's still failing, or
> > 2. The test is causing the test to crash (We can tell this because the program just stops in the middle of testing)
> > 
> > Are these situations ok to annotate a test as skip, or should I prevent it anyways?
> > 
> > I can make the change for porting it to Python, but it's probably something I'll tackle later as well as merging `fix_test.py` and `lib_updater.py` (I'll close the PR once this conversation is over)
> 
> We are skipping tests in one of 4 cases (from most common to most rare):
> 1. Test hangs forever
> 2. Test causes panic over rust side
> 3. Test is flaky (sometimes it passes, sometimes doesn't)
> 4. Test takes irregular amount of time, and there's no way to tune it down a bit

Nice, good to know! 
Re: 1, its addressed, through the use of a timeout, thats customizable.
2 and 3 are also addressed (see above comment)
If I'm understanding what you mean by 4, it's also addressed with the timeout (although the test treats it as a hang, for better or worse)

fanninpm:
> 1. Test hangs forever

A new feature added by #6410 is the `--timeout` flag for `-m test`.

PR #6715, opened 2026-01-12
Bytecode pseudo opcodes

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Internal instruction handling system restructured to improve code organization and maintainability.

### DIFF
--- a/.claude/commands/upgrade-pylib.md
+++ b/.claude/commands/upgrade-pylib.md
@@ -1,0 +1,32 @@
+# Upgrade Python Library from CPython
+
+Upgrade a Python standard library module from CPython to RustPython.
+
+## Arguments
+- `$ARGUMENTS`: Library name to upgrade (e.g., `inspect`, `asyncio`, `json`)
+
+## Steps
+
+1. **Delete existing library in Lib/**
+   - If `Lib/$ARGUMENTS.py` exists, delete it
+   - If `Lib/$ARGUMENTS/` directory exists, delete it
+
+2. **Copy from cpython/Lib/**
+   - If `cpython/Lib/$ARGUMENTS.py` exists, copy it to `Lib/$ARGUMENTS.py`
+   - If `cpython/Lib/$ARGUMENTS/` directory exists, copy it to `Lib/$ARGUMENTS/`
+
+3. **Upgrade tests**
+   - Run: `python lib_updater.py --quick-upgrade cpython/Lib/test/test_$ARGUMENTS`
+   - This will update the test files with appropriate RustPython markers
+
+## Example Usage
+```
+/upgrade-pylib inspect
+/upgrade-pylib json
+/upgrade-pylib asyncio
+```
+
+## Notes
+- The cpython/ directory should contain the CPython source that we're syncing from
+- lib_updater.py handles adding `# TODO: RUSTPYTHON` markers and `@unittest.expectedFailure` decorators
+- After upgrading, you may need to run tests to verify: `cargo run --release -- -m test test_$ARGUMENTS`
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -87,17 +87,22 @@ rm -r target/debug/build/rustpython-* && find . | grep -E "\.pyc$" | xargs rm -r
 # Run Rust unit tests
 cargo test --workspace --exclude rustpython_wasm
 
-# Run Python snippets tests
+# Run Python snippets tests (debug mode recommended for faster compilation)
+cargo run -- extra_tests/snippets/builtin_bytes.py
+
+# Run all Python snippets tests with pytest
 cd extra_tests
 pytest -v
 
-# Run the Python test module
+# Run the Python test module (release mode recommended for better performance)
 cargo run --release -- -m test ${TEST_MODULE}
 cargo run --release -- -m test test_unicode # to test test_unicode.py
 
 # Run the Python test module with specific function
 cargo run --release -- -m test test_unicode -k test_unicode_escape
 ```
+
+**Note**: For `extra_tests/snippets` tests, use debug mode (`cargo run`) as compilation is faster. For `unittest` (`-m test`), use release mode (`cargo run --release`) for better runtime performance.
 
 ### Determining What to Implement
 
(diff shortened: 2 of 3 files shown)


crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -857,7 +777,243 @@ impl Instruction {
...
+    fn try_from(value: u16) -> Result<Self, Self::Error> {
+        match u8::try_from(value) {
+            Ok(v) => v.try_into(),
+            Err(_) => Ok(PseudoInstruction::try_from(value)?.into()),
+        }
+    }
+}
+
+macro_rules! inst_either {
+    (fn $name:ident ( &self $(, $arg:ident : $argty:ty )* ) -> $ret:ty ) => {
```
youknowone:
```suggestion
    (fn $name:ident ( &self $(, $arg:ident : $arg_ty:ty )* ) -> $ret:ty ) => {
```
let's rename it to avoid adding dict

ShaharNaveh:
@youknowone can we ignore the `test_multiprocessing_fork` tests? they seem extremely flaky

crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -424,18 +383,7 @@ impl Instruction {
...
 
-    /// What effect this instruction has on the stack
-    ///
-    /// # Examples
-    ///
-    /// ```
-    /// use rustpython_compiler_core::bytecode::{Arg, Instruction, Label};
-    /// let (target, jump_arg) = Arg::new(Label(0xF));
-    /// let jump_instruction = Instruction::Jump { target };
-    /// assert_eq!(jump_instruction.stack_effect(jump_arg, true), 0);
```
youknowone, line 436: docs are lost

crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -857,7 +777,243 @@ impl Instruction {
...
+        _level: usize,
+    ) -> fmt::Result {
+        unimplemented!()
+    }
+}
+
+const _: () = assert!(mem::size_of::<PseudoInstruction>() == 2);
+
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum Instruction {
```
youknowone: Is this type fit in u16? Otherwise union could be better

youknowone (commented):
Thanks! I am still looking in

crates/compiler-core/src/bytecode/instruction.rs
ShaharNaveh: Hmmm, I've put it in the `InstructionMetadata` trait, but I guess it makes more sense to put it under the `impl`

crates/compiler-core/src/bytecode/instruction.rs
ShaharNaveh: Ain't sure what you mean, could you please explain?

And yes, this type fit in a `u16` it can hold a `RealInstruction` (u8) or a `PseudoInstruction` (u16)

youknowone: Please add `#[repr(u16)]` or `static_assertions::assert_eq_size` to ensure compiler actually recognize it or not.

crates/compiler-core/src/bytecode/instruction.rs
youknowone: oh, sorry. doc will be enough on traits

crates/compiler-core/src/bytecode/instruction.rs
ShaharNaveh: I tried to add `#[repr(u16)]`, but ```core::mem::size_of::<Instruction>()``` gives `4`.
Why do we care about the the size of `Instruction`? It's not intended to be marshaled directly

crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -15,9 +15,9 @@ use crate::{
 /// A Single bytecode instruction.
 /// Instructions are ordered to match CPython 3.13 opcode numbers exactly.
 /// HAVE_ARGUMENT = 44: opcodes 0-43 have no argument, 44+ have arguments.
-#[derive(Debug, Copy, Clone, PartialEq, Eq)]
```
ShaharNaveh, line 18: I've removed `Eq` & `PartialEq` as they are only used once in the code. It's a very tiny optimization for binary size. I can revert it if it's a concern

crates/compiler-core/src/bytecode/instruction.rs
youknowone: Oh, you are right! I didn't catch that.
Still union looks more fit than Instruction because they are just separated types sharing a same space. Will that be too much painful to use union than enum for this?

ShaharNaveh: > Oh, you are right! I didn't catch that. Still union looks more fit than Instruction because they are just separated types sharing a same space. Will that be too much painful to use union than enum for this?

IMO yes it will. And we wouldn't gain a lot from it tbh as the instruction loop will only have `RealInstruction` anyway (which is what we had until now)

ShaharNaveh: Do you want me to remove the `#[repr(u16)]` so the compiler can have some further optimizations?

youknowone: hmm, if this is only used by compiler anyway, probably that could be better idea.

I was originally thinking
```rust
union Instruction {
    real: RealInstruction,
    pseudo: PseudoInstruction,
    raw: u16,
}

impl Instruction {
    fn real(&self) -> Option<RealInstruction> {
        if self.raw < ... { 
            Some(...)
        } else {
             ...
        }
    }
}
```

btw, if The pseudo instruction is only used for compiler, vm only will have `RealInstruction`. That looks a bit awkward. How about keeping the RealInstruction name as Instruction, and the wrapper as AnyInstruction?

```
enum AnyInstruction {
    Real(Instruction),
    Pseudo(PseudoInstruction),
}
```

ShaharNaveh: > hmm, if this is only used by compiler anyway, probably that could be better idea.
> 
> I was originally thinking
> 
> ```rust
> union Instruction {
>     real: RealInstruction,
>     pseudo: PseudoInstruction,
>     raw: u16,
> }
> 
> impl Instruction {
>     fn real(&self) -> Option<RealInstruction> {
>         if self.raw < ... { 
>             Some(...)
>         } else {
>              ...
>         }
>     }
> }
> ```

While it's cool (and doable) I don't think we will gain much from doing that.

> btw, if The pseudo instruction is only used for compiler, vm only will have `RealInstruction`. That looks a bit awkward. How about keeping the RealInstruction name as Instruction, and the wrapper as AnyInstruction?
> 
> ```
> enum AnyInstruction {
>     Real(Instruction),
>     Pseudo(PseudoInstruction),
> }
> ```

Sure, I can do that

PR #6716, opened 2026-01-12
Downgraded skips in tests

terryluan12:
## Summary by CodeRabbit

* **Chores**
  * CI updated to include an additional Windows test in the suite, improving test coverage and reliability for Windows builds.

### DIFF
--- a/Lib/test/test_ftplib.py
+++ b/Lib/test/test_ftplib.py
@@ -902,7 +902,8 @@ class TestIPv6Environment(TestCase):
         self.client.set_pasv(False)
         retr()
 
-@unittest.skip('TODO: RUSTPYTHON; SSL + asyncore has problem; flaky')
+
+@unittest.skip("TODO: RUSTPYTHON; SSL + asyncore has problem")
 @skipUnless(ssl, "SSL not available")
 @requires_subprocess()
 class TestTLS_FTPClassMixin(TestFTPClass):


terryluan12:
Double check tests in `test_class.py`, `test_configparser.py`, `test_context.py`, `test_str.py`, `test_str.py`, `test_subprocess.py`

terryluan12:
Note to self: testing if, given the new timeout, whether timeouts count as failures, or errors (expectedFailure vs skips)

Lib/test/test_sqlite3/test_dbapi.py
```diff
@@ -547,7 +547,6 @@ def test_connection_init_good_isolation_levels(self):
 
     # TODO: RUSTPYTHON
     # @unittest.expectedFailure
```
fanninpm: These can probably be deleted if they are not needed.

Lib/test/test_sqlite3/test_dbapi.py
```diff
@@ -587,11 +586,11 @@ def test_connection_bad_reinit(self):
                                    ((v,) for v in range(3)))
 
 
-@unittest.skip("TODO: RUSTPYTHON")
 class UninitialisedConnectionTests(unittest.TestCase):
     def setUp(self):
         self.cx = sqlite.Connection.__new__(sqlite.Connection)
 
+    @unittest.skip('TODO: RUSTPYTHON')
```
fanninpm, line 591: Any _particular_ reason for the skip?

Lib/test/test_descr.py
```diff
@@ -1834,8 +1834,8 @@ def __init__(self, foo):
         object.__init__(A(3))
         self.assertRaises(TypeError, object.__init__, A(3), 5)
 
-    @unittest.skip('TODO: RUSTPYTHON; This passes, but the `expectedFailure` here is from CPython, so this test is an "UNEXPECTED SUCCESS" (not good)')
-    @unittest.expectedFailure
+    # TODO: RUSTPYTHON; The `expectedFailure` here is from CPython, so this test actually fails. Not run on linux?
+    # @unittest.expectedFailure
```
fanninpm: @youknowone what do you want to do about this? Related: python/cpython#49572

Lib/test/test_marshal.py
```diff
@@ -498,7 +498,7 @@ def testModule(self):
         self.helper(code)
         self.helper3(code)
 
-    @unittest.skip("TODO: RUSTPYTHON")
+    @unittest.skip('TODO: RUSTPYTHON')
```
fanninpm: Any _particular_ reason for the skip?

Lib/test/test_threading.py
```diff
@@ -319,7 +319,7 @@ def f(mutex):
 
     # PyThreadState_SetAsyncExc() is a CPython-only gimmick, not (currently)
     # exposed at the Python level.  This test relies on ctypes to get at it.
-    @unittest.skip('TODO: RUSTPYTHON; expects @cpython_only')
+    @unittest.expectedFailure # TODO: RUSTPYTHON; expects @cpython_only
```
fanninpm: This has already been fixed upstream in python/cpython#143002, so it may be reasonable to do the same here:
```suggestion
    @cpython_only
```

Lib/test/test_sqlite3/test_dbapi.py
terryluan12: I'm not sure. For this PR, I mainly focused on downgrading/specifying the @unittest.skip's, so they were all already notated, I'm just downgrading/removing/specifying them. 

This one specifically was skipping already, except the whole class was skipped, so I just moved it to the specific failing test. It didn't have a reason attached, so the moved reason also didn't.

Lib/test/test_marshal.py
youknowone: this is not regression; unfair to ask to find reason in this PR

terryluan12: Same for this one, it was pre-existing, so I just copied it (in this specific case, I just changed it from double to single quotes)

Lib/test/test_sqlite3/test_dbapi.py
fanninpm: If an underlying reason can be given, it will certainly be helpful when someone tries to fix the problem.

youknowone: @fanninpm i understand the logic, but review is about the changes, not about the code seen

Lib/test/test_io.py
```diff
@@ -1846,7 +1846,6 @@ def test_error_through_destructor(self):
     def test_truncate_on_read_only(self):
         return super().test_truncate_on_read_only()
 
-    @unittest.skip('TODO: RUSTPYTHON; fallible allocation')
     def test_constructor(self):
```
ShaharNaveh: When we have tests that looks like:
```python
@unittest.expectedFailure # TODO: RUSTPYTHON
def foo(self):
    return super().foo()
```

That is a test that we are patching for that specific subclass.
anyway, if there are no other decorators for those kind of tests, they can be deleted entirely

Lib/test/test_io.py
```diff
@@ -2196,7 +2195,6 @@ def test_args_error(self):
     def test_error_through_destructor(self):
         return super().test_error_through_destructor()
 
-    @unittest.skip('TODO: RUSTPYTHON; fallible allocation')
     def test_constructor(self):
         return super().test_constructor()
```
ShaharNaveh: same

Lib/test/test_io.py
```diff
@@ -2695,7 +2693,6 @@ def test_args_error(self):
...
```
ShaharNaveh: same

Lib/test/test_io.py
terryluan12: Ah whoops! Sounds good. Just fixed lol, thanks for the catch

Lib/test/test_tempfile.py
```diff
@@ -1735,7 +1735,7 @@ def test_cleanup_with_symlink_to_a_directory(self):
         d2.cleanup()
 
     @os_helper.skip_unless_symlink
-    @unittest.skip('TODO: RUSTPYTHON; No such file or directory "..."')
+    @unittest.skip('TODO: RUSTPYTHON; alters the execution environment (env changed)')
```
fanninpm: For better tracking:
```suggestion
    @unittest.skipIf(
        'RUSTPYTHON_SKIP_ENV_POLLUTERS' in os.environ,
        'TODO: RUSTPYTHON; alters the execution environment (env changed)'
    )
```
...and when you do this, please add `test_tempfile` to the `ENV_POLLUTING_TESTS_COMMON` list in `.github/workflows/ci.yaml`.

____

You may end up needing to do something like this:

```suggestion
    @unittest.skipIf(
        sys.platform == 'linux' and 'RUSTPYTHON_SKIP_ENV_POLLUTERS' in os.environ,
        'TODO: RUSTPYTHON; alters the execution environment (env changed)'
    )
```

...or this:

```suggestion
    @unittest.skipIf(
        sys.platform in ('darwin', 'linux') and 'RUSTPYTHON_SKIP_ENV_POLLUTERS' in os.environ,
        'TODO: RUSTPYTHON; alters the execution environment (env changed)'
    )
```

...in which case(s), please make sure `test_tempfile` is added to the appropriate OS-specific variable(s) in `.github/workflows/ci.yaml`.

Lib/test/test_descr.py
youknowone: I found our `__new__` behavior is different from Python 3.14. (not checked incompatibility of 3.13 or changed in 3.14)
Let's see that change affect this change

Lib/test/test_ast/test_ast.py
```diff
@@ -2396,7 +2396,7 @@ def test_stdlib_validates(self):
         ast.MatchMapping([], [], rest="_"),
     ]
 
-    @unittest.skip("TODO: RUSTPYTHON; thread 'main' panicked")
+    @unittest.expectedFailure # TODO: RUSTPYTHON; thread 'main' panicked
```
youknowone: If thread 'main' panicked, this test cannot be `expectedFailure`. Please update the reason or delete it.

Lib/test/test_tempfile.py
terryluan12: I'm still trying to nail down the `test_tempfile.py` error to double check, but fyi there may an issue with the `ENV_POLLUTING_TESTS_COMMON` list. 

In [the last test run](https://github.com/RustPython/RustPython/actions/runs/20976681189/job/60292814336?pr=6716), it seems that `test.test_multiprocessing_forkserver.test_threads` and `test.test_multiprocessing_spawn.test_threads` had a similar ENV CHANGED issue, despite the fact that no changes are made to the `test_multiprocessing_spawn` tests

Lib/test/test_zipfile/test_core.py
```diff
@@ -3484,29 +3483,29 @@ def _test_read(self, zipfp, expected_names, expected_content):
             self.assertEqual(info.file_size, len(content))
             self.assertEqual(zipfp.read(name), content)
 
-    @unittest.expectedFailure # TODO: RUSTPYTHON; 
+    @unittest.skip('TODO: RUSTPYTHON; shift_jis encoding unsupported')
```
youknowone: Please revert this file. 
This test is failing because setUp is failing.
`@unittest.skip("TODO: RUSTPYTHON shift_jis encoding unsupported")` to the entire test class effectively do that. we don't need to skip each test. All of them must be unskipped by a single patch.

Lib/test/test_configparser.py
```diff
@@ -762,7 +762,6 @@ def test_read_returns_file_list(self):
         parsed_files = cf.read([], encoding="utf-8")
         self.assertEqual(parsed_files, [])
 
-    @unittest.skip("TODO: RUSTPYTHON, suspected to make CI hang")
```
youknowone, line 765: By the comment of skip, I suggest to leave a comment

```python
    # XXX: RUSTPYTHON; This test might cause CI hang
```
It will be helpful if test_configparser starts to hang again. We upgrade test library every year to new Python library. The warning comment can be removed then if really no problem happens

Lib/test/test_descr.py
```diff
@@ -1834,8 +1834,8 @@ def __init__(self, foo):
         object.__init__(A(3))
         self.assertRaises(TypeError, object.__init__, A(3), 5)
 
-    @unittest.skip('TODO: RUSTPYTHON; This passes, but the `expectedFailure` here is from CPython, so this test is an "UNEXPECTED SUCCESS" (not good)')
-    @unittest.expectedFailure
+    # TODO: RUSTPYTHON; The `expectedFailure` here is from CPython, so this test actually fails
```
youknowone:
```suggestion
    # TODO: RUSTPYTHON; The `expectedFailure` here is from CPython, so this test actually must fail
```

Lib/test/test_ftplib.py
```diff
@@ -920,9 +919,7 @@ def setUp(self, encoding=DEFAULT_ENCODING):
         self.client.auth()
         self.client.prot_p()
 
-
 @skipUnless(ssl, "SSL not available")
-@unittest.skip("TODO: RUSTPYTHON; SSL + asyncore has problem")
```
youknowone, line 925: This is not fixed yet.

Lib/test/test_http_cookiejar.py
```diff
@@ -125,7 +125,6 @@ def test_http2time_garbage(self):
                               "http2time(%s) is not None\n"
                               "http2time(test) %s" % (test, http2time(test)))
 
-    @unittest.skip("TODO: RUSTPYTHON, regressed to cubic complexity")
```
youknowone, line 128: is this now running fast enough?

Lib/test/test_set.py
```diff
@@ -1809,24 +1809,29 @@ def check_set_op_does_not_crash(self, function):
                 self.assertIn("changed size during iteration", str(e))
 
 
-@unittest.skip("TODO: RUSTPYTHON; segfault")
 class TestBinaryOpsMutating(TestOperationsMutating):
 
+    @unittest.skip('TODO: RUSTPYTHON; segfault')
```
youknowone: revert this file please. prev version is better

Lib/test/test_subprocess.py
```diff
@@ -2318,7 +2318,7 @@ def _test_extra_groups_impl(self, *, gid, group_list):
                                       extra_groups=[name_group])
 
     # No skip necessary, this test won't make it to a setgroup() call.
-    @unittest.skip('TODO: RUSTPYTHON; clarify failure condition')
+    @unittest.skip('TODO: RUSTPYTHON')
```
youknowone: change without benefit

Lib/test/test_json/test_recursion.py
```diff
@@ -68,7 +68,7 @@ def default(self, o):
             self.fail("didn't raise ValueError on default recursion")
 
     # TODO: RUSTPYTHON
-    @unittest.skip("TODO: RUSTPYTHON; crashes")
+    @unittest.skip('TODO: RUSTPYTHON; crashes')
```
youknowone: by practical reason, I understand changing formatting the entire file after script run, and the result is actually changed.
But this file only has quotes changes. I am not a fan of this because it increase cost of `git log` and `git blame` when they are necessary

youknowone (changes requested):
Thank you so much! You seem to find many outdated skips in our test code.

Please make sure unskipped tests are not actually flaky. Running multiple times of them will be helpful.
Sometimes they are flaky on specific platforms. I will run the entire CI multiple times once this is done.

Lib/test/test_http_cookiejar.py
terryluan12: It works locally fine! Runs and finishes in 0.089s. If there is an issue though, let me know and I can re-skip it

terryluan12:
@youknowone It should be ok to rereview/run the CI/CD tests. 

Like I said, there seem to be intermittent issues with `test.test_multiprocessing_forkserver.test_threads` + `test.test_multiprocessing_spawn.test_threads`  but it seems to be unrelated.

youknowone:
You’re right. those are making it hard to see the actual results of this test. I’ll fix that and run it again.

youknowone:
the failure looks like unrelated

----

````
ERROR: test_saturation (test.test_concurrent_futures.test_process_pool.ProcessPoolForkProcessPoolExecutorTest.test_saturation)
  File "/Users/runner/work/RustPython/RustPython/crates/pylib/Lib/test/test_concurrent_futures/test_process_pool.py", line 114, in test_saturation
    executor.submit(sem.acquire)
                    ^^^
  File "/Users/runner/work/RustPython/RustPython/crates/pylib/Lib/concurrent/futures/process.py", line 791, in submit
    raise BrokenProcessPool(self._broken)
concurrent.futures.process.BrokenProcessPool: A child process terminated abruptly, the process pool is not usable anymore



2 tests failed:
    test.test_concurrent_futures.test_process_pool
    test.test_concurrent_futures.test_shutdown
````

File: Lib/test/test_ftplib.py [hunk 19]
```diff
@@ -902,8 +902,7 @@ def retr():
         self.client.set_pasv(False)
         retr()
 
-
-@unittest.skip("TODO: RUSTPYTHON; SSL + asyncore has problem")
+@unittest.skip('TODO: RUSTPYTHON; SSL + asyncore has problem; flaky')
```
youknowone:
```suggestion

@unittest.skip("TODO: RUSTPYTHON; SSL + asyncore has problem")
```
revert

PR #6741, opened 2026-01-16
Bytecode instrumented placeholder

ShaharNaveh:
## Summary by CodeRabbit

* **New Features**
  * Expanded bytecode instruction set with specialized and instrumented opcodes to support enhanced code execution capabilities.

* **Refactor**
  * Updated instruction encoding, validation logic, and metadata implementation to accommodate expanded opcode set.
  * Adjusted instruction discriminant values for improved alignment.

### DIFF
--- a/Lib/test/test__opcode.py
+++ b/Lib/test/test__opcode.py
@@ -16,6 +16,7 @@ class OpListTests(unittest.TestCase):
                 self.assertIsInstance(func(op), bool)
                 self.assertEqual(func(op), expected)
 
+    @unittest.expectedFailure # TODO: RUSTPYTHON; Move LoadClosure to psudoes
     def test_invalid_opcodes(self):
         invalid = [-100, -1, 255, 512, 513, 1000]
         self.check_bool_function_result(_opcode.is_valid, invalid, False)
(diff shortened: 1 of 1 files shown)


crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -277,10 +274,106 @@ pub enum Instruction {
         target: Arg<Label>,
     } = 130,
     JumpIfNotExcMatch(Arg<Label>) = 131,
-    SetExcInfo = 134,
-    Subscript = 135,
+    SetExcInfo = 132,
+    Subscript = 133,
```
ShaharNaveh, line 278: The bytecode already has to be purged because `LoadClosure` is no longer 253. So I've moved these as well

youknowone: sure, this will also totally reordered on 3.14

ShaharNaveh:
> is the newly introduced ops all the pseudo ops?

None are. All are either specialized or instrumented opcodes

youknowone:
Thanks! I didn't know about them.

PR #6742, opened 2026-01-16
Complement upgrade-pylib Claude Code command

moreal:
The command was improved based on the process of updating the json module using the `/update-pylib` command. Additionally, the script was modified to add `import unittest` if it is missing.

I worked on https://github.com/RustPython/RustPython/pull/6743 with this pull request's changes.

## Summary by CodeRabbit

* **Documentation**
  * Added a detailed, step-by-step upgrade workflow for migrating tests, including examples, diff-review guidance, explicit restoration criteria for environment-specific markers, and a new example showing how to restore changes.

* **New Features**
  * Enhanced test-upgrade tooling to support single-file and directory upgrades, a verification flow to run and mark tests based on outcomes, and automatic insertion of missing test imports when applying patches.

### DIFF
--- a/scripts/lib_updater.py
+++ b/scripts/lib_updater.py
@@ -282,10 +282,11 @@ def has_unittest_import(tree: ast.Module) -> bool:
 
 def find_import_insert_line(tree: ast.Module) -> int:
     """Find the line number after the last import statement."""
-    last_import_line = 0
+    last_import_line = None
     for node in tree.body:
         if isinstance(node, (ast.Import, ast.ImportFrom)):
             last_import_line = node.end_lineno or node.lineno
+    assert last_import_line is not None
     return last_import_line
 
 


scripts/lib_updater.py
```diff
@@ -270,11 +270,44 @@ def {test_name}(self):
...
 
 
+def has_unittest_import(tree: ast.Module) -> bool:
+    """Check if 'import unittest' is already present in the file."""
+    for node in tree.body:
+        if isinstance(node, ast.Import):
+            for alias in node.names:
+                if alias.name == UT:
+                    return True
+        elif isinstance(node, ast.ImportFrom):
```
ShaharNaveh: I think we should remove this elif since we want to bring `unittest` into scope

moreal: You're right! 😓 I'll fix it now...

ShaharNaveh: we can maybe adjust the decorator to `__import__('unittest').expectedFailure` and not needing this patch at all?

scripts/lib_updater.py
```diff
@@ -270,11 +270,44 @@ def {test_name}(self):
...
+            if node.module == UT:
+                return True
+    return False
```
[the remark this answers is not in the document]
moreal: I fixed it. Could you review again?

scripts/lib_updater.py
```diff
@@ -270,11 +270,44 @@ def {test_name}(self):
...
+
+    # If we have modifications and unittest is not imported, add it
+    if modifications and not has_unittest_import(tree):
+        import_line = find_import_insert_line(tree)
+        modifications.append(
+            (
+                import_line,
+                "import unittest # XXX: RUSTPYTHON; importing to be able to skip tests",
+            )
+        )
```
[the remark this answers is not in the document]
moreal: I fixed it. Could you review again?

scripts/lib_updater.py
moreal: That would certainly allow us to fully resolve the question of where to place the imports, but the code does feel a bit awkward to read, and I also wonder whether calling `__import__` every time—while not significant—might slightly slow down the tests. I don’t have strong convictions or a firm opinion on this.

I’ll follow the maintainers’ decision.

ShaharNaveh: In CPython this will have very minimal performance cost due to modules being cached. I ain't sure about RustPython:/

moreal: Yes, that’s right. I think it’s probably well optimized. It just feels a bit awkward to look at code like the example below.

~~~~ python
@__import__('unittest').expectedFailure
def test_xxx():
  # Something
~~~~

Functionally, there’s no major issue, so I’m mainly interested in hearing the maintainers’ (i.e., the people primarily responsible for maintenance) opinions. Would you be able to share your thoughts on this thread or on this PR? @youknowone

ShaharNaveh: We can maybe check if there's a `import unittest` node in the AST and switch between `@unittest.X` or `__import__('unittest').X`, as I feel like most tests do `import unittest`

youknowone: Well, while that will work perfectly, we still need to keep them human friendly.
Most of people with python-background are familiar with `@unittest.X`. So let's stick on it.

Performance doesn't matter.

youknowone:
@ShaharNaveh As the major author of lib_updater, please be a major reviewer of this PR. I will merge it once you approve it. Thanks!

scripts/lib_updater.py
```diff
@@ -270,11 +270,73 @@ def {test_name}(self):
...
+        ):
+            return first_node.end_lineno or first_node.lineno
+
+    # No imports and no docstring - try to skip shebang/encoding if lines provided
+    # PEP 263: encoding declaration must be in first or second line
+    # and match: ^[ \t\f]*#.*?coding[:=][ \t]*([-_.a-zA-Z0-9]+)
+    if lines:
+        insert_line = 0
+        for i, line in enumerate(lines[:2]):  # Only check first two lines per PEP 263
+            if line.startswith("#!") or re.match(
```
ShaharNaveh: Maybe this entire function is a bit of an overkill? inserting the import before the first:
- class
- function
- constant

do the trick? we don't have lots of files that don't import unittest

moreal: Come to think of it, Python `test.test_*` modules are for testing something, so they inherently have imports. I think the function's complexity is from implementing the suggestion of the below comment, so I'll revert it to first implementation:

https://github.com/RustPython/RustPython/pull/6742#discussion_r2698193354

Like following lines:

``` python
def find_import_insert_line(tree: ast.Module) -> int:
    """Find the line number after the last import statement."""
    last_import_line = None
    for node in tree.body:
        if isinstance(node, (ast.Import, ast.ImportFrom)):
            last_import_line = node.end_lineno or node.lineno
    assert last_import_line is not None  # expect passing because test module might have `import x` to test `x`
    return last_import_line
```

How do you think about it?

ShaharNaveh: This has an issue in the case of an import inside a test.
```py
import posix

def test_posix_import():
    import nt
    ...
```

I'd put a condition to check that it it's indent is 0.

---

Or, insert it after the first import. It won't be pretty but I think it's putting a lot of effort to solve an uncommon problem

moreal: I don't think such a case would be an issue because it iterates `ast.Module.body` not recursively like visitor pattern. So, as long as the import is at the top like the usual pattern, it should be fine.

``` python
>>> import ast
>>> code = """import posix
...
... def test_posix_import():
...     import nt
...     ..."""
>>> ast.parse(code)
Module(body=[Import(names=[alias(name='posix', asname=None)]), FunctionDef(name='test_posix_import', args=arguments(posonlyargs=[], args=[], vararg=None, kwonlyargs=[], kw_defaults=[], kwarg=None, defaults=[]), body=[Import(names=[alias(...)]), Expr(value=Constant(...))], decorator_list=[], returns=None, type_comment=None, type_params=[])], type_ignores=[])
>>> tree = ast.parse(code)
>>> for node in tree.body: print(node)
...
Import(names=[alias(name='posix', asname=None)])
FunctionDef(name='test_posix_import', args=arguments(posonlyargs=[], args=[], vararg=None, kwonlyargs=[], kw_defaults=[], kwarg=None, defaults=[]), body=[Import(names=[alias(name='nt', asname=None)]), Expr(value=Constant(value=Ellipsis, kind=None))], decorator_list=[], returns=None, type_comment=None, type_params=[])
```

ShaharNaveh: Oh, you're right :facepalm: 
let's go with that approach

moreal: @ShaharNaveh I just reverted the previous commit (at [5b54ec9](https://github.com/RustPython/RustPython/pull/6742/commits/5b54ec9d68c1693ff12f08a38537e9dee77cf28a)) and added `assert ...` line for assumption (at [80dfec2](https://github.com/RustPython/RustPython/pull/6742/commits/80dfec2eb50d30677f458259c29aed3becee529f)). Please take a look!

ShaharNaveh (approved):
LGTM!
@youknowone can we merge?

youknowone (approved):
@moreal @ShaharNaveh Thank you so much!

PR #6718, opened 2026-01-13
RustPython version to 3.14

youknowone:
close #6702
close #6543 

Python 3.14 appears to be a release with a significant number of changes.
~This PR currently includes PEP 649 and PEP 750, along with various other behavior changes and deprecations.~

- mark versions 3.14
- Partial implementation of PEP 649 `__annotate__`; this seems coupled to 3.14 changes in many libraries
  - inspect, typing is partially edited to avoid fully upgrade test.support
- incomplete PEP 750 is separated to #6744, but will be activated in this PR
- minor upgrades
  - pickle is removed from itertools
  - upgrade venvlauncher binaries
  - bool(NotImplemented) is fully deprecated
  - {bytes,bytearray}.fromhex behavior is changed
  - int rounding takes None
  - `__init__` behavior of exceptions is changed
  - co_consts is changed
  - site,ensurepip,annotationlib,string,_colorize,argparse,calendar
    - because we implemented ctypes now, ensurepip doesn't require local patch anymore!

## Summary by CodeRabbit

* **New Features**
  * PEP 649-style deferred annotations and lazy __annotate__ support for modules, types, and functions.
  * bytes.fromhex() and bytearray.fromhex() accept bytes-like inputs.

* **Bug Fixes**
  * bool(NotImplemented) now raises TypeError.
  * int.__round__(None) accepted as documented.
  * Docstrings preserved in code objects (CO_HAS_DOCSTRING behavior).

* **Behavior Changes**
  * Some itertools iterator pickle support removed.

* **Chores**
  * Minimum Python requirement bumped to 3.14.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6752, opened 2026-01-17
Add some compiler tests from 3.13.11

ShaharNaveh:


youknowone:
Usually we can take a lower version patch first and update it later. But #6718 heavily updated compiler due to annotation changes, I'd like to ensure this is 3.14 compatible.
Could you please update this patch to Python 3.14?

Or simply rebase and rerun will be also okay

youknowone (approved):
👍

PR #6765, opened 2026-01-18
Await fsm checkpoint

yusufyian:
## Summary by CodeRabbit

* **New Features**
  * Checkpoint and resume functionality to save/restore VM execution state
  * FSM-based async function continuations with decorator-driven transformations
  * Filesystem-backed host API with state persistence, gas accounting, and event logging
  * Deterministic execution mode with configurable stdlib whitelisting
  * Binary renamed from `rustpython` to `pvm`

* **Examples**
  * Smart contract demos: DEX with AMM mechanics, staking rewards, escrow marketplace, batch payroll
  * Checkpoint/resume workflow demonstrations


youknowone (changes requested):
Thank you for patch.

Because this is a huge patch. I need to understand it better. Before starting review, here is a checklist for big patch.

- [ ] Please describe the motivation of the patch.
- [ ] Please ensure pvm is a hard dependency of this feature.
- [ ] This is not a core Python feature. It must be optional
- [ ] Is this the best minimal size of the patch? Usually merging a huge single patch is not easy. Please break it down to smaller ones if possible.

PR #6788, opened 2026-01-18
Introduce compression module and update related modules to 3.14.2

moreal:
Overview
----------

What was previously `Lib/_compression.py` is now `Lib/compression/_common/_streams.py`, and the original file has been removed. During this update process, test files also needed updating, and related compression module files were changed as of version 3.14.2.

As a note: The test_zstd file has been newly added, but since _zstd is not yet supported, it is automatically skipped so the test will not fail.

Context
--------

- https://docs.python.org/3/library/compression.html
  > Added in version 3.14.
- https://github.com/python/cpython/commit/20be6ba61ac8a0a5d6242701c4186579cfa653f0

### DIFF
--- a/Lib/bz2.py
+++ b/Lib/bz2.py
@@ -10,9 +10,9 @@ __all__ = ["BZ2File", "BZ2Compressor", "BZ2Decompressor",
 __author__ = "Nadeem Vawda <nadeem.vawda@gmail.com>"
 
 from builtins import open as _builtin_open
+from compression._common import _streams
 import io
 import os
-import _compression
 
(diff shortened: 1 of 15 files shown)


youknowone:
Are you also interested in xz and zstd support?

moreal:
> Are you also interested in xz and zstd support?

@youknowone I'm not very familiar with compression algorithms, so if I get the chance to study them, I'd be really happy. However, due to other personal things to do, I won't be able to look into this issue for at least three days. So if the issue is still open in a few days, I'll take a look at it one by one.

I've also created a zstd issue! #6793

PR #6792, opened 2026-01-18
Run benchmarks with stdlib native modules

moreal:
Personally, I was checking https://rustpython.github.io/benchmarks to see if benchmarks improved after the `json.loads` enhancement, but there was actually no performance improvement. Upon reviewing this again, I suspect it was running with a Pure Python implementation using an interpreter without stdlib. So if there are no particular issues, how about running the benchmark with a version that includes stdlib?

I'm not very familiar with whether the presence of a standard library has a significant impact on performance benchmarks, so I would be really happy if you could share your opinion!

## Summary by CodeRabbit

* **Chores**
  * Added standard library support to development dependencies to enable richer test environments.
  * Updated development benchmarks to initialize with the standard library, yielding more realistic performance measurements.
  * No changes to public/external APIs or exported signatures.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,6 +54,7 @@ rustyline = { workspace = true }
 [dev-dependencies]
 criterion = { workspace = true }
 pyo3 = { version = "0.27", features = ["auto-initialize"] }
+rustpython-stdlib = { workspace = true }
 
 [[bench]]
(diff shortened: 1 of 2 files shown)


youknowone (approved):
Probably that was written for vm performance at the first time.

Issue #3418, opened 2021-11-06
Introduce `_collections.OrderedDict` type (a.k.a. `collections.OrderedDict`)


moreal:
## Feature

When you see the `collections` module,  there are two versions of `collections.OrderedDict`, pure Python implementation and implemented in CPython C code.

To check compatibility, run the below lines:

```python
from _collections  import OrderedDict
```

It is not incompatible in public exposed APIs so you can close this issue if it is useless.

## Python Documentation

- https://github.com/python/cpython/blob/v3.10.0/Objects/odictobject.c
- https://docs.python.org/3/library/collections.html?highlight=ordereddict#collections.OrderedDict

PR #6799, opened 2026-01-19
Implement `_collections.OrderedDict`

moreal:
This pull request will resolve some parts of #3418. I need to review these code again 🔨


crates/vm/src/builtins/dict.rs
```diff
@@ -325,8 +325,11 @@ impl PyDict {
     }
 
     fn __or__(&self, other: PyObjectRef, vm: &VirtualMachine) -> PyResult {
-        let other_dict: Result<PyDictRef, _> = other.downcast();
-        if let Ok(other) = other_dict {
+        // Only accept exact dict type, not subclasses
+        // This allows subclasses like OrderedDict to handle the operation via __ror__
+        if other.class().is(vm.ctx.types.dict_type)
+            && let Ok(other) = other.downcast::<PyDict>()
```
youknowone: still there is convenience problem, but we have downcast_exact

crates/vm/src/stdlib/collections.rs
```diff
@@ -727,4 +732,1094 @@ mod _collections {
             })
         }
     }
+
+    // ============================================================================
+    // OrderedDict implementation
```
youknowone: If this type is comparably big to this module `crates/vm/src/stdlib/collections/ordered_dict.rs`?

moreal:
It may be replaced with #8780

youknowone:
@moreal hey I missed this is revised. do you want to finish this?

moreal:
> hey I missed this is revised. do you want to finish this?

@youknowone  No, I just remembered a PR I opened a while back and then closed. Since I don't have no relevant context right now, I think it would be best for you—having just opened the PR—to see it through to completion.

PR #6784, opened 2026-01-18
Bump logging, re, fnmatch, glob, mimetypes, urllib, pathlib to 3.14.2

moreal:
This pull request bumps `logging`, `re`, `fnmatch`, `glob`, `mimetypes`, `urllib`, `pathlib` to 3.14.2.

I apologize that the PR has become quite large while trying to pass CI. I tried to split it by commits to make reviewing easier, but if it would be better to split the PR, I will do so.

   - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

How to review
---------------

I'm keeping this as a record because it was useful during the review process, and we also used it when entrusting the task to an LLM and reviewing it.

1.  Ensure there is `cpython/` directory
2.  Run `git show sha` (or `git show HEAD~2`) to check if there are any accidental arbitrary modifications that have disappeared. (It can be reviewed in GitHub diff)
3.  Run `delta Lib/<lib>.py cpython/Lib/<lib>.py` and `delta Lib/test/<lib>.py cpython/Lib/test/<lib>.py`,
     or `delta Lib/<lib>/ cpython/Lib/<lib>/` and `delta Lib/test/test_<lib>/ cpython/Lib/test/test_<lib>/`,
     to ensure that CPython has been updating properly and that no strange values have been inserted.

PR #6808, opened 2026-01-20
Bump mimetypes to 3.14.2

moreal:
This pull request is extracted from #6784

### DIFF
--- a/Lib/test/test_ssl.py
+++ b/Lib/test/test_ssl.py
@@ -3525,6 +3525,7 @@ class ThreadedTests(unittest.TestCase):
             else:
                 s.close()
 
+    @unittest.expectedFailureIfWindows("TODO: RUSTPYTHON")
     def test_socketserver(self):
         """Using socketserver to create and manage SSL connections."""
         server = make_https_server(self, certfile=SIGNED_CERTFILE)


moreal:
The job run on window-2025 has failed with the following lines. It seems not related about updating mimetypes module to 3.14.2 🤔 

```
FAIL: test_socketserver (test.test_ssl.ThreadedTests.test_socketserver)
Using socketserver to create and manage SSL connections.
  File "D:\a\RustPython\RustPython\Lib\test\test_ssl.py", line 3552, in test_socketserver
    self.assertEqual(d1, d2)
                         ^^
AssertionError: b'# Test the support for SSL and sockets\r\[254224 chars]\r\n' != b''

```

Lib/test/test_ssl.py
```diff
@@ -3525,7 +3525,6 @@ def test_starttls(self):
             else:
                 s.close()
 
-    @unittest.expectedFailure # TODO: RUSTPYTHON
```
youknowone, line 3528: please revert this line

moreal: Since I don't have Windows machine, I cannot check the test but it seems to fail in only Windows environment.  I'll mark it as `unittest.expectedFailureIfWindows`.


In my local macOS environment, the following command doesn't fail.
```
cargo run --release -- -m unittest test.test_ssl.ThreadedTests.test_socketserver
```

PR #6827, opened 2026-01-21
Introduce `/upgrade-pylib-next`

moreal:
This pull request introduces `/upgrade-pylib-next` command to find a module to update by running `scripts/update_lib todo` (#6823). It just finds a module to update and delegates other to `/upgrade-pylib`. (cc https://github.com/RustPython/RustPython/issues/6580)

This pull request also has changes about `/upgrade-pylib`. It became to use `auto-mark` subcommand to test newly introduced tests also.

## Summary by CodeRabbit

* **Chores**
  * Added structured guidance document for investigating test failures with step-by-step analysis and issue reporting workflow.
  * Added workflow document for selecting and upgrading Python library modules based on dependency status.
  * Updated Python library upgrade process with streamlined steps, enhanced test-management guidance, and improved examples.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6845, opened 2026-01-23
Update codecs from v3.14.2

moreal:
> [!NOTE]
> I'm dogfooding for #6827

### DIFF
--- a/Lib/test/test_codecs.py
+++ b/Lib/test/test_codecs.py
@@ -3592,7 +3592,7 @@ class CodePageTest(unittest.TestCase):
             (b'\xff', 'strict', '\xff'),
         ))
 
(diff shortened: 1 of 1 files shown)


ShaharNaveh:
@moreal can we merge this?

moreal:
> @moreal can we merge this?

@ShaharNaveh we can but I realized  there are more related files (this PR has changed only codecs and test_codecs). If it is okay to separate them into two PR, this can be merged at this status.

Lib/test/multibytecodec_support.py
youknowone, line 1: adding these new series of tests seems not the best idea for initial updating codec. Could you remove these new files and update only codec? that will be better to update codec right now

moreal: Okay, I'll revert them.

PR #6846, opened 2026-01-23
Align psuedo ops to CPython 3.14.2

youknowone:
## Release Notes

* **New Features**
  * Added support for special method optimization and executor specialization.

* **Refactor**
  * Simplified internal attribute access compilation and bytecode emission.
  * Updated opcode support for Python 3.14.2 alignment.
  * Removed redundant instruction variants for streamlined bytecode generation.

### DIFF
--- a/crates/codegen/src/ir.rs
+++ b/crates/codegen/src/ir.rs
@@ -7,8 +7,7 @@ use rustpython_compiler_core::{
     bytecode::{
         AnyInstruction, Arg, CodeFlags, CodeObject, CodeUnit, CodeUnits, ConstantData,
         ExceptionTableEntry, InstrDisplayContext, Instruction, InstructionMetadata, Label, OpArg,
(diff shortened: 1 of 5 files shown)


crates/codegen/src/ir.rs
```diff
@@ -207,62 +206,18 @@ impl CodeInfo {
...
-                            let encoded =
-                                encode_load_super_attr_arg(idx.get(info.arg), false, true);
-                            info.arg = OpArg(encoded);
-                            info.instr = Instruction::LoadSuperAttr { arg: Arg::marker() }.into();
-                        }
-                        _ => {}
-                    }
-
+                // Real instructions are already encoded by compile.rs
+                if info.instr.real().is_some() {
```
ShaharNaveh: I'd replace the line 214, no need to check twice for if it's real/pseudo

PR #6882, opened 2026-01-26
Docs: add reference to `rustpython_derive` close to example

elmjag:
Include a link for native module macros documentation, after an example where they are used. 

It's not obvious how find this documentation, when reading the example code for the native module.

## Summary by CodeRabbit

* **Documentation**
  * Updated crate documentation to include reference to available macros and their usage.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -41,6 +41,8 @@
 //!
 //! The binary will have all the standard arguments of a python interpreter (including a REPL!) but
 //! it will have your modules loaded into the vm.
+//!
+//! See [`rustpython_derive`](../rustpython_derive/index.html) crate for documentation on macros used in the example above.
 
 #![cfg_attr(all(target_os = "wasi", target_env = "p2"), feature(wasip2))]
 #![allow(clippy::needless_doctest_main)]


Cargo.toml
```diff
@@ -31,6 +31,7 @@ winresource = "0.1"
 
 [dependencies]
 rustpython-compiler = { workspace = true }
+rustpython-derive = { workspace = true }
```
youknowone: Thank you for contributing. This dependency is added only for doc reference. It doesn't feel like ideal. Could you tell me why it is a good idea?

elmjag: Not sure if this is a good idea. This was the only way I could make rustdoc to generate a link to `rustpython_derive` docs here:

https://github.com/RustPython/RustPython/pull/6882/changes#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R45

Maybe there is a better way to make the link work?

Another option, is to skip the link, just mention the crate name. It's is less convenient for the reader, but perhaps not a huge deal.

elmjag: Ok, I just realized that it's also possible to specify relative URLs in rustdoc. So it also possible to make the link work like this:

`//! See ['rustpython_derive'](../rustpython_derive/index.html) crate ...`

In that case, there is no add dependency to `rustpython-derive` in the root crate. 

What do you think about using relative URL instead?

youknowone: Sounds good. Adding dependency is very negative.

elmjag: Updated PR to use relative URL instead. No need anymore for additional dependency.

elmjag:
> welcome to RustPython project.

Thanks!

PR #6889, opened 2026-01-27
Add `ExpatError` and `error` to `pyexpat`

fanninpm:
This PR is one of a series of PRs in preparation for getting https://rustpython.github.io/pages/regression-tests-results.html to work again by using the Python test suite to generate a JUnit-compatible XML file (and later converting said XML a JSON format the current website can accept).
In pursuit of that, I discovered that `xml.parsers.expat.error` needs to be available to the interpreter, and I suspect that it's used to let the Python-side XML machinery report a useful error.
Here, I tried to make `pyexpat.error` available; however, when I submitted this PR, I came up short of my goal, as I currently don't know a useful way of getting `pyexpat.error` to ailias `pyexpat.ExpatError`.
In CPython 3.14, the `pyexpat` module has a `pyexpat_state` struct that gets initialized with, among other things, a new `xml.parsers.expat.ExpatError` exception.
[When the `pyxpat` module is initialized](https://github.com/python/cpython/blob/b82b12bc80373e2cfd0c0bed273e55e67640cdfd/Modules/pyexpat.c#L2191-L2337), the two objects `error` and `ExpatError` are aliased to `state->Error`.
If anyone knows how to replicate this behavior in RustPython on the Rust side, I would be glad to implement that.

## Summary by CodeRabbit

* **New Features**
  * Added a new `ExpatError` exception type for improved error handling in XML parsing operations.

### DIFF
--- a/Lib/test/test_xml_etree.py
+++ b/Lib/test/test_xml_etree.py
@@ -384,7 +384,6 @@ class ElementTreeTest(unittest.TestCase):
         self.serialize_check(element,
                 '<tag key="value"><subtag /><subtag /></tag>')
 
-    @unittest.expectedFailure # TODO: RUSTPYTHON
     def test_cdata(self):
         # Test CDATA handling (etc).
 
@@ -407,7 +406,6 @@ class ElementTreeTest(unittest.TestCase):
         self.assertEqual(tree.find("element/../empty-element").tag,
                 'empty-element')
 
-    @unittest.expectedFailure # TODO: RUSTPYTHON
     def test_path_cache(self):
         # Check that the path cache behaves sanely.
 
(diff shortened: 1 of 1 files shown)


crates/stdlib/src/pyexpat.rs
```diff
@@ -420,6 +420,17 @@ mod _pyexpat {
 
         PyExpatLikeXmlParser::new(ns_sep, args.intern, vm)
     }
+
+    // TODO: Tie this exception to the module's state.
+    // TODO: Alias `error` to `ExpatError`
+    #[pyattr]
```
youknowone, line 425:
```suggestion
    #[pyattr]
    #[pyattr(name = "error")]
```
This will add alias error

fanninpm: Thank you!

youknowone (approved):

Issue #6746, opened 2026-01-16
[RFC] Split oparg type from the opcode enum


ShaharNaveh:
## Summary

Introduce a new `AnyOparg` enum & make `Instruction` + `PseudoInstruction` enums to not hold the oparg type.

## Detailed Explanation

The current `Instruction` enum couples the opcode and its oparg type together. In practice, this creates friction and many boilerplate code that can be avoided.

### Suggested API
New pure `Opcode` enum that will have only the mapping between the opcode name and it's ID, like this:
```rust
enum Opcode {
   Nop = 0,
   IsOp = 1,
   BinaryOp = 2,
   ...
}
```

A new `AnyOparg` enum that will have all oparg types as it's variants:
```rust
enum AnyOparg {
    NameIdx(oparg::NameIdx),
    IntristicFunction1(oparg::IntristicFunction1),
    Invert(oparg::Invert),
    ...
}
```

An `Instruction` struct that is like a typed version of the `CodeUnit` struct:
```rust
struct Instruction {
  opcode: AnyOpcode,
  oparg: AnyOparg,
}

impl Instruction {
  pub fn new<T: Into<AnyOpcode>>(opcode: T, oparg: Option<i32>) -> Result<Self, MarshalError> {
    let oparg = match opcode.into() {
      AnyOpcode::Real(Opcode::Nop) => None,
      AnyOpcode::Real(Opcode::IsOp) => Some(oparg::Invert::try_from(oparg.unwrap())?),
      ...
    };

    Ok(Self { opcode: opcode.into(), oparg: oparg.map(Into::into) })
  }
}
```

### Drawbacks
- Will require a serious refactoring to the code

### Rational
- Reduce boilerplate code (use `num_enum` crate for example).
- Easier to auto-generate the opcodes (and their IDs) in the future.
- Simplify `opcode -> oparg` mapping.

## Unresolved Questions

Performance implifications?
- Will probably increase performance because we will validate the oparg once instead of doing so every time we need to do an operation on the opcode.

Implementation & refactoring efforts?

youknowone:
@coolreader18 I'd like to listen what was the rational to made current layout. And how do you think about the changes.

Python instruction changed a lot since 3.10. Probably this will be a good chance to research what's the best for next a few years

PR #6893, opened 2026-01-29
Split stack_effect into pushed & popped

ShaharNaveh:
Auto generated both `num_popped` and `num_pushed`, and then compared the result with the current `stack_effect` impl. This will make it easier to auto-generate this in the future, and gets us 1 step closer to #6746 as imo it's better to construct the instruction once instead of validate that the oparg is valid at every operation. Also, it's easier to see if we differences in our opcode impls from CPython (if the stack_effect does not match)

## Summary by CodeRabbit

* **Refactor**
  * Consolidated stack-effect handling into a structured push/pop abstraction and exposed it across instruction representations so callers can query pushed/popped counts or a combined effect.
* **Chores**
  * Added bidirectional conversion for opcode argument values to simplify argument handling and adjusted internal call sites accordingly.
* **Impact**
  * Improves consistency and reliability when analyzing and optimizing bytecode.

### DIFF
--- a/Lib/symtable.py
+++ b/Lib/symtable.py
@@ -1,9 +1,16 @@
 """Interface to the compiler's internal symbol tables"""
 
 import _symtable
-from _symtable import (USE, DEF_GLOBAL, DEF_NONLOCAL, DEF_LOCAL, DEF_PARAM,
-     DEF_IMPORT, DEF_BOUND, DEF_ANNOT, SCOPE_OFF, SCOPE_MASK, FREE,
-     LOCAL, GLOBAL_IMPLICIT, GLOBAL_EXPLICIT, CELL)
+from _symtable import (
+    USE,
+    DEF_GLOBAL,  # noqa: F401
+    DEF_NONLOCAL, DEF_LOCAL,
+    DEF_PARAM, DEF_TYPE_PARAM, DEF_FREE_CLASS,
+    DEF_IMPORT, DEF_BOUND, DEF_ANNOT,
+    DEF_COMP_ITER, DEF_COMP_CELL,
+    SCOPE_OFF, SCOPE_MASK,
+    FREE, LOCAL, GLOBAL_IMPLICIT, GLOBAL_EXPLICIT, CELL
+)
 
 import weakref
 from enum import StrEnum
@@ -165,6 +172,10 @@ class SymbolTable:
                 for st in self._table.children]
 
 
+def _get_scope(flags):  # like _PyST_GetScope()
+    return (flags >> SCOPE_OFF) & SCOPE_MASK
+
+
 class Function(SymbolTable):
 
     # Default values for instance variables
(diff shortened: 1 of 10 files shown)


crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -1152,26 +851,33 @@ impl AnyInstruction {
     }
 }
 
-pub trait InstructionMetadata {
+pub trait StackEffect {
+    /// How many items this instruction is popping from the stack.
+    fn num_popped(&self, oparg: i32) -> i32;
+
+    /// How many items this instruction is pushing on the stack.
+    fn num_pushed(&self, oparg: i32) -> i32;
```
youknowone: I like this idea. How do you think about `stack_popped_and_pushed(&self, oparg: i32) -> (i32, i32)` to avoid a whole instruction match twice

crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -1152,26 +851,33 @@ impl AnyInstruction {
     }
 }
 
-pub trait InstructionMetadata {
+pub trait StackEffect {
```
youknowone: What's good if we split this trait? It seems required to implement for the all types implementing InstructionMetadata

crates/compiler-core/src/bytecode/instruction.rs
ShaharNaveh: Sure, I can do it as a followup improvement if this is fine by you

crates/compiler-core/src/bytecode/instruction.rs
ShaharNaveh: No need really, it made more sense for me to split it. but it can be part of `InstructionMetadata`. I'm fine with whatever you want:)

youknowone: Splitting traits as grouping is looking attractive while defining and implementing traits, but it usually leads to import hell for user side. I'd like to keep a single trait if all InstructionMetadata and StackEffect has manual implementation requirements. It can be splitted when one has manual implementation but the other one has blanket implementation (or any kind of different implementation strategy)

crates/compiler-core/src/bytecode/instruction.rs
ShaharNaveh: I've made `StackEffect` into a struct, this will be clearer than returning an unnamed tuple

youknowone: oh, i meant not about changing stack_effect return type but merging num_popped and num_pushed into single function returning a tuple (or StackEffect, that's looking better!) to avoid full instruciton match for each function.
if we are going to use `stack_effect` as a single number for most of places, keeping current `-> i32` will be easier for user side.

crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -1162,17 +1187,14 @@ pub trait InstructionMetadata {
...
-    /// let jump_instruction = Instruction::JumpForward { target };
-    /// assert_eq!(jump_instruction.stack_effect(jump_arg), 0);
-    /// ```
-    fn stack_effect(&self, arg: OpArg) -> i32;
+    /// Stack effect info for how many items are pushed/popped from the stack,
+    /// for this instruction.
+    fn stack_effect_info(&self, oparg: u32) -> StackEffect;
+
+    /// Stack effect of [`Self::stack_effect_info`].
+    fn stack_effect(&self, oparg: u32) -> i32 {
```
youknowone, line 1195: :+1:

PR #6894, opened 2026-01-29
align varnames to CPython

youknowone:


crates/compiler-core/src/bytecode.rs
```diff
@@ -276,12 +276,13 @@ pub struct CodeObject<C: Constant = ConstantData> {
...
+    pub localsplusnames: Box<[C::Name]>,
+    pub localspluskinds: Box<[LocalKind]>,
+    pub nlocals: u32,
+    pub ncellvars: u32,
+    pub nfreevars: u32,
     pub constants: Box<[C]>,
     pub names: Box<[C::Name]>,
-    pub varnames: Box<[C::Name]>,
-    pub cellvars: Box<[C::Name]>,
-    pub freevars: Box<[C::Name]>,
```
youknowone: @coolreader18 I remember you decided to keep this in current design before. I'd like to listen what do you think about the changes.
The motivation is aligning architecture more to CPython to reduce AI engineering cost. Because no actual technical advantage exists, I will keep our own way if you have good reasons not to do.

PR #6897, opened 2026-01-29
Update `py_compile` from 3.14.2

ShaharNaveh:


Lib/test/test_py_compile.py
```diff
@@ -109,18 +109,17 @@ def test_cwd(self):
         self.assertTrue(os.path.exists(self.pyc_path))
         self.assertFalse(os.path.exists(self.cache_path))
 
-    import platform
```
fanninpm, line 112: This may need to be restored with a `TODO: RUSTPYTHON` comment attached to it.

youknowone (approved):

PR #6919, opened 2026-01-31
Update `test_extcall` from v3.14.2

fanninpm:
Since this is a doctest, I had to do some weird things to mark tests as expected failures.


fanninpm:
> I wish we had simpler way to do that. We will loose it easy when upgrading them

@ShaharNaveh you may want to look at what I did to this test to make it "pass".

ShaharNaveh:
> > I wish we had simpler way to do that. We will loose it easy when upgrading them
> 
> @ShaharNaveh you may want to look at what I did to this test to make it "pass".

That's fantastic! ty:)

PR #6910, opened 2026-01-30
gc module internal structure and API

youknowone:
Only interface, without actual GC

split from #6849

## Summary by CodeRabbit

* **New Features**
  * Full garbage collection API: enable/disable, manual collection, and generational behavior
  * GC monitoring: per-generation stats, collection counts, debug flags, and inspectable tracked objects
  * GC controls: configurable thresholds, freeze/unfreeze permanent generation, and GC callbacks/garbage lists
* **Behavioral**
  * Objects can report GC tracking and expose their referents for inspection

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

PR #6923, opened 2026-02-01
Remove RustPython-specific workaround in `timeit` module

moreal:
`XXX: RUSTPYTHON` comment might be removed by implementing `gc` module in #6910.


moreal:
The `timeit` module is already upgraded by #6854. This pull request will remove only RustPython-specific workaround thanks to #6910.

youknowone (approved):

PR #6932, opened 2026-02-01
Update enum from v3.14.2

youknowone:


Lib/test/test_enum.py
```diff
@@ -3063,7 +3040,7 @@ class ThirdFailedStrEnum(StrEnum):
                 one = '1'
                 two = b'2', 'ascii', 9
 
-    @unittest.expectedFailure # TODO: RUSTPYTHON; fails on encoding testing : TypeError: Expected type 'str' but 'builtin_function_or_method' found
+    @unittest.expectedFailure  # TODO: RUSTPYTHON; fails on encoding testing : TypeError: Expected type 'str' but 'builtin_function_or_method' found
```
ShaharNaveh, line 3043: Is the double space before the comment intentional?

```py
# used to be
@unittest.expectedFailure # TODO: RUSTPYTHON; ...
# now it's
@unittest.expectedFailure  # TODO: RUSTPYTHON; ...
```

youknowone: That's the tradition from [PEP-8#inline-comments](https://peps.python.org/pep-0008/#inline-comments)

> ... Inline comments should be separated by at least two spaces from the statement. They should start with a # and a single space.

Lib/test/test_enum.py
```diff
@@ -445,8 +445,7 @@ def spam(cls):
         with self.assertRaises(AttributeError):
             del Season.SPRING.name
 
-    @unittest.expectedFailure # TODO: RUSTPYTHON
-    # RuntimeError: Error calling __set_name__ on '_proto_member' instance failed in 'BadSuper'
+    @unittest.expectedFailure  # TODO: RUSTPYTHON
```
youknowone:
```suggestion
    @unittest.expectedFailure  # TODO: RUSTPYTHON; RuntimeError: Error calling __set_name__ on '_proto_member' instance failed in 'BadSuper'
```

PR #6953, opened 2026-02-02
Update `tabnanny` from v3.14.2

fanninpm:
Should this be part of `PLATFORM_INDEPENDENT_TESTS`?


fanninpm:
CI failure seems unrelated.

youknowone:
@fanninpm I usually don't feel safe to put IO-related library under platform-independent, but I don't believe RustPython users really want to use tabnanny. So probably good idea.

PR #6980, opened 2026-02-03
Use `num_enum` crate for oparg types

ShaharNaveh:
## Summary by CodeRabbit

* **Chores**
  * Standardized opcode/argument conversions across the bytecode layer for greater consistency.
  * Improved error handling for invalid bytecode, reducing ambiguous failures and making errors more actionable.
  * Updated textual formatting for several bytecode entities, making diagnostics and messages clearer.

* **New Features**
  * Exposed and stabilized many bytecode argument types, labels, operator flags and enums with explicit numeric representations to improve serialization, interoperability and tooling.


crates/compiler-core/src/bytecode/oparg.rs
```diff
@@ -112,15 +109,18 @@ impl OpArgState {
...
+#[num_enum(error_type(name = MarshalError, constructor = new_invalid_bytecode))]
+#[repr(u32)]
 pub enum ConvertValueOparg {
     /// No conversion.
     ///
     /// ```python
     /// f"{x}"
     /// f"{x:4}"
     /// ```
+    // Ruff `ConversionFlag::None` is `-1i8`, when its converted to `u8` its value is `u8::MAX`.
```
youknowone, line 142: I encountered similar issue a few days ago. There was a error due to 255 != -1.

crates/compiler-core/src/bytecode/oparg.rs
```diff
@@ -112,15 +109,18 @@ impl OpArgState {
 /// ## See also
 ///
 /// - [CPython FVC_* flags](https://github.com/python/cpython/blob/8183fa5e3f78ca6ab862de7fb8b14f3d929421e0/Include/ceval.h#L129-L132)
-#[repr(u8)]
-#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq)]
+#[derive(Clone, Copy, Debug, Eq, Hash, IntoPrimitive, PartialEq, TryFromPrimitive)]
+#[num_enum(error_type(name = MarshalError, constructor = new_invalid_bytecode))]
+#[repr(u32)]
```
youknowone: I know OpArg is up to u32, so it must be represented u32 for maximum extending.
But at the same time, each OpArg byte requires ExtendedArg instruction. It means if we use wrong-sized OpArg with wrong count of ExtendedArg, it will lead bug.

When we type this as repr(u8), this type is guaranteed to require no ExtendArg. But if repr(u32), we have to check the validity with another method. So this is looking bad idea to me.

I don't know much about OpArg, please correct me if I am missing something.

Oh, I also think this need to be typed i8 if possible.

ShaharNaveh: > I know OpArg is up to u32, so it must be represented u32 for maximum extending. But at the same time, each OpArg byte requires ExtendedArg instruction. It means if we use wrong-sized OpArg with wrong count of ExtendedArg, it will lead bug.

That's correct. I believe the CI will catch this immediately, so we might not need to worry about this too much.

> When we type this as repr(u8), this type is guaranteed to require no ExtendArg. But if repr(u32), we have to check the validity with another method. So this is looking bad idea to me.

The reason for why I changed it to `#[repr(u32)]` is because `num_enum` implements the `From<u32>` which `OpArgType` trait needs, I can revert it to `#[repr(u8)]` and implement `From<u32>` manually for each oparg enum.

> I don't know much about OpArg, please correct me if I am missing something.
> 
> Oh, I also think this need to be typed i8 if possible.

Cpython have a mix for this specific oparg, in the AST they represent it as `b'a' / b'r' / b's' / -1` but in the oparg it can be in `0..=2`, because ruff only cares about the AST they have it as an `i8`, we should keep it as `u8`

crates/compiler-core/src/bytecode/oparg.rs
ShaharNaveh: We should not do the conversion of `ruff_value as u8` but to match Ruff's enum and choose the correct value accordingly

crates/compiler-core/src/bytecode/oparg.rs
youknowone: you seem to made it working with u8, great 👍

crates/compiler-core/src/bytecode/oparg.rs
```diff
@@ -590,66 +572,45 @@ impl fmt::Display for CommonConstant {
...
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[repr(u8)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, TryFromPrimitive, IntoPrimitive)]
+#[num_enum(error_type(name = MarshalError, constructor = new_invalid_bytecode))]
 pub enum BuildSliceArgCount {
     /// ```py
     /// x[5:10]
     /// ```
-    Two,
+    Two = 2,
```
youknowone: I thought we probably switched this un-numbered enum for more room to optimization. Does this need to be numbered again?

ShaharNaveh: Yes, that's because of the use of `num_enum`, I can revert this change if you'd like. Also, why don't we do it for all other opargs?

youknowone: That's fair. I thought about this again. We can define the serialize/deserialize values separate from this values' internal memory representation, right? Then removing values from other types and only defining them for serialize/deserialize make sense.

PR #6982, opened 2026-02-03
Update `pickle.py` from 3.14.2

ShaharNaveh:


Lib/_compat_pickle.py
```diff
@@ -148,14 +147,6 @@
 else:
     PYTHON2_EXCEPTIONS += ("WindowsError",)
 
-# NOTE: RUSTPYTHON exceptions
-try:
-    JitError
-except NameError:
-    pass
-else:
-    PYTHON2_EXCEPTIONS += ("JitError",)
```
youknowone, line 157: probably we still need this for `--features jit`

Lib/test/support/rustpython.py
ShaharNaveh, line 1: @fanninpm can you please review? I've modified this slightly (renamed the class, and didn't use an `if` inside the `check_output` method)

Lib/test/support/rustpython.py
```diff
@@ -0,0 +1,20 @@
...
+import doctest
+
+
+# copied from https://github.com/RustPython/RustPython/pull/6919
+EXPECTED_FAILURE = doctest.register_optionflag("EXPECTED_FAILURE")
+
+
+class DocTestChecker(doctest.OutputChecker):
+    """
+    Custom output checker that lets us to add: `+EXPECTED_FAILURE` for doctest tests.
```
fanninpm: Minor grammar nitpick from a native speaker of English:

```suggestion
    Custom output checker that lets us add: `+EXPECTED_FAILURE` for doctest tests.
```

Also, consider expanding upon what exactly this option string does (allows us to mark doctest examples as currently failing, for accountability).

Lib/test/support/rustpython.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+
+    def check_output(self, want, got, optionflags):
+        if optionflags & EXPECTED_FAILURE:
+            return want != got
```
fanninpm: I realized that if
- you have both `+EXPECTED_FAILURE` and another doctest flag on a test that currently would fail without `+EXPECTED_FAILURE`
- but later, something is changed elsewhere in the interpreter that makes it so that the aforementioned test would pass without the presence of `+EXPECTED_FAILURE`

…then there's a risk of the current logic "covering up" the fact that `+EXPECTED_FAILURE` needs to be removed from the test.

I then came up with this:
```suggestion
            return not super().check_output(want, got, optionflags)
```
See #6945 for an example of it in action.

fanninpm (commented):
> @fanninpm can you please review? I've modified this slightly (renamed the class, and didn't use an `if` inside the `check_output` method)

A few things.

PR #6997, opened 2026-02-04
Better codecs and fix lots of test_io and other expectedFailures

youknowone:
## Summary by CodeRabbit

* **New Features**
  * Added `-X utf8` command-line option to control UTF‑8 mode
  * Added a UTF‑32 decode entry to the codecs module
  * Enabled pickling support for IO wrapper objects and subclasses

* **Improvements**
  * Stronger codec-name validation (NUL and UTF‑8 checks)
  * Better IO finalization with ResourceWarning-like warnings for unclosed files
  * Improved surrogate/encoding error recovery to allow partial output on failure

### DIFF
--- a/Lib/_pycodecs.py
+++ b/Lib/_pycodecs.py
@@ -1109,7 +1109,7 @@ def unicode_call_errorhandler(errors,  encoding,
     else:
         exceptionObject = UnicodeEncodeError(encoding, input, startinpos, endinpos, reason)
     res = errorHandler(exceptionObject)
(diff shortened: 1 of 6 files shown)


crates/common/src/encodings.rs
```diff
@@ -640,3 +640,228 @@ pub mod ascii {
         )
     }
 }
+
+pub mod utf7 {
```
youknowone: no utf7 impl in rust

PR #7003, opened 2026-02-05
Move `pickletester.py` to the correct place

ShaharNaveh:
#6982 put it under `Lib` instead of `Lib/test` by accident


youknowone:
@ShaharNaveh did you manually do it because update_lib didn't cover it?

ShaharNaveh:
> @ShaharNaveh did you manually do it because update_lib didn't cover it?

I did it manually, I didn't even run lib_updater :sweat_smile:

PR #7015, opened 2026-02-05
Fix traceback, syntax errors, and exception handling 

youknowone:
## Summary by CodeRabbit

* **New Features**
  * Frame clearing API with ownership-aware safeguards
  * Traceback objects list tb_frame/tb_next/tb_lasti/tb_lineno via __dir__

* **Bug Fixes / Improvements**
  * Prefer Python traceback formatting when printing exceptions
  * More accurate source-location tracking for subscripts/expressions
  * Enhanced syntax-error metadata and EOF/unclosed-bracket detection
  * Clearer ImportError messages and improved name/import suggestions
  * Safer tb_next setter (deletion prevented) and idempotent AttributeError augmentation
  * Updated duplicate-argument diagnostic wording

### DIFF
--- a/crates/compiler/src/lib.rs
+++ b/crates/compiler/src/lib.rs
@@ -27,6 +27,8 @@ pub struct ParseError {
     pub location: SourceLocation,
     pub end_location: SourceLocation,
     pub source_path: String,
+    /// Set when the error is an unclosed bracket (converted from EOF).
+    pub is_unclosed_bracket: bool,
 }
 
 impl ::core::fmt::Display for ParseError {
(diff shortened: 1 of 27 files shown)


crates/compiler/src/lib.rs
```diff
@@ -102,6 +149,106 @@ impl CompileError {
     }
 }
 
+/// Find the last unclosed opening bracket in source code.
+/// Returns the bracket character and its byte offset, or None if all brackets are balanced.
+fn find_unclosed_bracket(source: &str) -> Option<(char, usize)> {
```
ShaharNaveh, line 154: Maybe ruff has this functionality somewhere?

youknowone: https://github.com/astral-sh/ruff/blob/main/crates/ruff_python_parser/src/parser/expression.rs seems to have a few related functions , but not sure about getting position

PR #7032, opened 2026-02-06
Bytecode oparg optimization

ShaharNaveh:
Discussed at https://github.com/RustPython/RustPython/pull/6980#discussion_r2771659157
Pretty much reverts #6980.

## Summary by CodeRabbit

* **Chores**
  * Removed unused internal dependency
  * Internal code structure improvements and standardization

* **Refactor**
  * Simplified internal type conversion patterns across the compiler

### DIFF
--- a/crates/compiler-core/src/bytecode/oparg.rs
+++ b/crates/compiler-core/src/bytecode/oparg.rs
@@ -209,8 +209,6 @@ macro_rules! impl_oparg_enum {
         }
 
         impl OpArgType for $name {}
-
-        // const _: () = assert!(::core::mem::size_of::<$name>() == ::core::mem::size_of::<u8>());
     };
 }
 


crates/compiler-core/src/bytecode/oparg.rs
```diff
@@ -104,11 +103,82 @@ impl OpArgState {
...
+                $(#[$variant_meta:meta])*
+                $variant:ident = $value:literal $(| $alternatives:expr)*
+            ),* $(,)?
+        }
+    ) => {
+        $(#[$enum_meta])*
+        $vis enum $name {
+            $(
+                $(#[$variant_meta])*
+                $variant, // Do assign value to variant.
```
ShaharNaveh, line 150: @youknowone This breaks RustPython. if I put `$variant = $value` then everything works just fine.
I suspect that there's some unsafe transmuting somewhere, do you have an idea how can I debug this?

ShaharNaveh: After lots of debugging I've figured that the only problematic enum is `ComparisonOperator`. I'm not sure about why yet, but it's not related to an unsafe transmutation like I initially thought

youknowone: ComparisonOperator seems to support bit operation. It doesn't have `0` like other enums

Issue #4210, opened 2022-10-11
[RFC] Support safe embedding


stephanemagnenat:
## Summary

As far as I have seen, RustPython is not yet suitable for "safe" embedding, meaning that executed Python code can block or hurt the caller code, because:

- execution can loop forever,
- memory allocation is not bounded,
- import cannot be disabled.

## Detailed Explanation

I wish to use RustPython as a scripting language within a game engine, running third-party user code. A requirement for me is that this code is run in a safe way. As far as I have seen (but I might have missed some elements), it is currently not the case:

- It is not possible to limit code execution up to a maximum number of instructions. Looking into RustPython's source, if I'm not mistaken, this feature would need to be added to `ExecutingFrame` in its [execution loop](https://github.com/RustPython/RustPython/blob/main/vm/src/frame.rs#L347). Maybe the [`ExecutionResult` type](https://github.com/RustPython/RustPython/blob/main/vm/src/frame.rs#L123) could be extended with an `InstructionBudgetExceeded` variant or similar.
- I couldn't find a way to set limits to heap allocation or stack size, that would also be needed.
- My understanding is that [importlib is always enabled](https://github.com/RustPython/RustPython/blob/main/vm/src/import.rs#L34), but that the OS module is currently disabled in [Wasm32 or Wasi](https://github.com/RustPython/RustPython/blob/main/vm/src/import.rs#L42), is that correct? If so, there should be a way to have finer-control on import lib or even forbid user-defined import statements altogether. Also, I see no reason to link the control of the availability of the `os` module to WASM, maybe a feature flag would be a good addition. Similarly, it should be possible to not enable some [Windows specific code](https://github.com/RustPython/RustPython/blob/main/vm/src/import.rs#L45) and fully disable or controls IOs regardless of the platform (a game engine is a platform in itself).

## Drawbacks, Rationale, and Alternatives

The rationale is to use RustPython as an embedding language within larger software, such as game engines. In these, the software must fully control the scripting environment's limits.

The main drawback is increased code complexity within RustPython, but I believe it can be done cleanly, with some work of course. The split of the Std library (#3102) was already a step towards the direction of embedding.

The alternatives are to not implement this feature, or do it in a fork. A similar issue exists (#3090), but it is more of a question, so I thought an RFC-style new issue is better.

## Unresolved Questions

There are quite some design questions obviously, but I guess first one should agree whether this overall feature makes sense for the project, then the design can be worked out. Probably a unified way to control embedding would be elegant.

[...]

PR #7035, opened 2026-02-07
initial sandbox 

youknowone:
#4210 

## Summary by CodeRabbit

* **New Features**
  * Introduced a host vs sandbox runtime mode: many stdlib modules (os, socket, signal, file I/O, ssl, etc.) are now available only in host mode; sandbox mode is restricted.
  * Added a lightweight sandbox stdio implementation for non-host environments.

* **Tests / Chores**
  * CI now validates sandbox builds and runs a new sandbox smoke test to verify runtime behavior.

### DIFF
--- a/extra_tests/snippets/sandbox_smoke.py
+++ b/extra_tests/snippets/sandbox_smoke.py
@@ -7,10 +7,10 @@ Verifies basic functionality that works in both sandbox and normal mode:
 - open() is properly blocked when FileIO is unavailable (sandbox)
 """
 
-import sys
-import math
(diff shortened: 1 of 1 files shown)


crates/vm/src/stdlib/builtins.rs
```diff
@@ -7,6 +7,8 @@ pub use builtins::{ascii, print, reversed};
 
 #[pymodule]
 mod builtins {
+    use crate::py_io;
```
youknowone: useless diff

PR #7044, opened 2026-02-08
Fix signal

youknowone:
I found this in local but no idea what this is fixing


youknowone:
abandoned trial of #7012

PR #7055, opened 2026-02-08
Replace unsafe pointer cast with PyMutex in PyCode

youknowone:
Add `source_path: PyMutex<&'static PyStrInterned>` field to `PyCode` for interior mutability, replacing the UB-inducing `#[allow(invalid_reference_casting)]` + `write_volatile` pattern in `update_code_filenames`. Update all 12 read sites across the codebase to use the new `source_path()` method.

## Summary by CodeRabbit

* **Refactor**
  * Improved thread-safety and safer handling of code object source paths across the runtime by replacing direct internal field access and in-place mutations with controlled accessors and setters, reducing unsafe mutations and standardizing path retrieval in traceback, import, and warning/reporting flows.

### DIFF
--- a/crates/stdlib/src/faulthandler.rs
+++ b/crates/stdlib/src/faulthandler.rs
@@ -252,7 +252,7 @@ mod decl {
     /// Dump a single frame's info to fd (signal-safe), reading live data.
     #[cfg(any(unix, windows))]
     fn dump_frame_from_raw(fd: i32, frame: &Frame) {
-        let filename = frame.code.source_path.as_str();
+        let filename = frame.code.source_path().as_str();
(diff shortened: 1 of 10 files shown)


jaens:
@youknowone doesn't `PyMutex` add quite a bit of memory overhead? The mutex functionality is not _really_ even used here, as there are no non-atomic operations done inside the lock...

Since the lifetime is `static` anyway, I think something like [atomic_ref](https://docs.rs/atomic_ref/latest/atomic_ref/) should work (or `AtomicPtr` from std, though that would still require a bit of `unsafe`).

What the update code does is actually an atomic compare and exchange, I think.

youknowone:
@jaens Thanks! That sounds right

PR #7092, opened 2026-02-12
raise TypeError when `__iter__` method is `None`

elmjag:
Setting special `__iter__` method to `None` is used to mark objects as non-iterable.

https://docs.python.org/3/reference/datamodel.html#special-method-names

Check if `__iter__` is set to `None` and raise TypeError exception.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced error reporting for iteration operations with clearer, more precise error messages when attempting to iterate over non-iterable objects
  * Improved exception handling to distinguish between different iteration failure scenarios

### DIFF
--- a/Lib/test/test_typing.py
+++ b/Lib/test/test_typing.py
@@ -10916,7 +10916,6 @@ class TypeIterationTests(BaseTestCase):
         Annotated[T, ''],
     )
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
     def test_cannot_iterate(self):
         expected_error_regex = "object is not iterable"
         for test_type in self._UNITERABLE_TYPES:
--- a/crates/vm/src/types/slot.rs
+++ b/crates/vm/src/types/slot.rs
@@ -1,6 +1,7 @@
 use crate::common::lock::{
     PyMappedRwLockReadGuard, PyMappedRwLockWriteGuard, PyRwLockReadGuard, PyRwLockWriteGuard,
 };
+use crate::vm::PyMethod;
 use crate::{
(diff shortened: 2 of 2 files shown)


ShaharNaveh:
Thanks for the PR!
I don't think that the failed CI is related to your change, I'll investigate

ShaharNaveh:
yup: https://github.com/rust-lang/miri/issues/4855

elmjag:
> yup: [rust-lang/miri#4855](https://github.com/rust-lang/miri/issues/4855)

Thanks for investigating!

ShaharNaveh:
@elmjag please merge main, it should pass now

elmjag:
> 
 > `537-562`: **Consider consistency with `__hash__ = None` handling.**
 > The `__hash__ = None` case (lines 758–783) is handled at slot-update time by storing `hash_not_implemented` as the slot function. The `__iter__ = None` case is handled at call time inside the wrapper. Both approaches work, but the inconsistency may confuse future maintainers. If there's no strong reason to differ, consider aligning the pattern (e.g., storing a dedicated `iter_not_implemented` function during slot update, similar to `hash_not_implemented`).
 > That said, CPython's `slot_tp_iter` also checks at call time, so the current approach is arguably more faithful to CPython.

I'll look into using this approach for `__iter__ = None` case. Sounds like a better way to implement it, marking this PR as draft for now.

ShaharNaveh:
@elmjag can you please resolve the conflicts? I think it can be merged

elmjag:
> @elmjag can you please resolve the conflicts? I think it can be merged

This issue is now fixed, via https://github.com/RustPython/RustPython/pull/7091. Closing this PR.

PR #7111, opened 2026-02-13
Fix inspect.signature introspection for attrgetter/itemgetter/methodcaller

moreal:
## Summary by CodeRabbit

* **Documentation**
  * Enhanced signature introspection for operator functions (attrgetter, itemgetter, methodcaller) by adding Python-visible signature hints. These functions now properly reflect their positional-only parameter requirements when users inspect function signatures.

### DIFF
--- a/Lib/test/test_operator.py
+++ b/Lib/test/test_operator.py
@@ -668,15 +668,6 @@ class PyOperatorTestCase(OperatorTestCase, unittest.TestCase):
 class COperatorTestCase(OperatorTestCase, unittest.TestCase):
     module = c_operator
 
-    def test_attrgetter_signature(self):
-        return super().test_attrgetter_signature()
-
-    def test_itemgetter_signature(self):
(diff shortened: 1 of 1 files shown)


Lib/test/test_operator.py
```diff
@@ -668,15 +668,12 @@ class PyOperatorTestCase(OperatorTestCase, unittest.TestCase):
 class COperatorTestCase(OperatorTestCase, unittest.TestCase):
     module = c_operator
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
     def test_attrgetter_signature(self):
```
youknowone: The entire method must be deleted.
running `update_lib quick operator` also will fix it

moreal: I deleted those methods! I've only added a commit to remove unnecessary test overrides without changing any existing content, so I think it's fine to merge if CI resources are tight.

youknowone (approved):

PR #7112, opened 2026-02-13
Remove dummy threading

ShaharNaveh:
I don't see those files in CPython 3.14.3


youknowone:
we use them for wasm build

PR #7126, opened 2026-02-14
Add CI step for checking redundant test patches

ShaharNaveh:
## Chores
* Improved continuous integration pipeline with enhanced Python code quality checks to automatically identify and report redundant test patches.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -467,7 +467,6 @@ jobs:
       - name: Check for redundant test patches
         run: python scripts/check_redundant_patches.py
 
-      - uses: actions/checkout@v6.0.2
       - uses: dtolnay/rust-toolchain@stable
         with:
           components: clippy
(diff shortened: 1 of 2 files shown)


youknowone:
@ShaharNaveh Can this be auto-fix using update_lib instead just lint?

ShaharNaveh:
> @ShaharNaveh Can this be auto-fix using update_lib instead just lint?

Yes, but I'd rather not doing that, for multiple reasons:

- The update_lib is not stable enough imo.
- We would run into issues where someone fixes something else, but this will cause the entire test file to be updated and it will add more stuff to the PR that can be out of scope.

youknowone:
@ShaharNaveh could you check coderabbitai reviews?

youknowone:
What i feel is this is a harmless glitch and automatically fixed when upgrading to next version. I'd rather not to bother contributors by that. Let's try how this work. We can disable it again if it reveals actually annoying.

PR #7139, opened 2026-02-14
Match SymbolTable repr with CPython format

moreal:
## Summary by CodeRabbit

* **Improvements**
  * Symbol table objects now display with detailed information (table name, object ID, and line number) when printed in the Python REPL or error messages.
  * Updated symbol table class identifier for clarity.

### DIFF
--- a/crates/vm/src/stdlib/symtable.rs
+++ b/crates/vm/src/stdlib/symtable.rs
@@ -3,9 +3,10 @@ pub(crate) use _symtable::module_def;
 #[pymodule]
 mod _symtable {
     use crate::{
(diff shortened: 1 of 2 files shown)


crates/vm/src/stdlib/symtable.rs
```diff
@@ -132,7 +133,7 @@ mod _symtable {
     }
 
     #[pyattr]
-    #[pyclass(name = "SymbolTable")]
+    #[pyclass(name = "symtable entry")]
```
youknowone, line 136: SymbolTable itself is not an entry, is this correct?

PR #7176, opened 2026-02-16
Optimize redundant bool check

### LINKED ISSUE #3567: Boolean expression performs redundant bool check
```python
n = 0

class Test:
    def __bool__(self):
        global n
        n += 1
        return False

Test() and False or False
assert n == 1

```

CPython optimizes this check in:
https://github.com/python/cpython/blob/54ab9ad312ea53db40e31712454272e1d4c0315f/Python/compile.c#L8708-L8721

moreal:
Closes #3567 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved short-circuit evaluation for nested boolean expressions so inner truth checks are skipped when an outer expression short-circuits, preventing redundant evaluations and ensuring correct control flow across mixed boolean operators.

* **Tests**
  * Added tests covering nested boolean short-circuit scenarios, including cases that verify inner __bool__ checks are not invoked when outer short-circuiting applies.

### DIFF
--- a/extra_tests/snippets/syntax_short_circuit_bool.py
+++ b/extra_tests/snippets/syntax_short_circuit_bool.py
@@ -33,4 +33,4 @@ while ExplodingBool(False) and False:
 #     pass
 
 # Issue #3567: nested BoolOps should not call __bool__ redundantly
-assert ExplodingBool(False) and False or False
+assert (ExplodingBool(False) and False or False) == False


extra_tests/snippets/builtin_bool.py
```diff
@@ -207,3 +207,17 @@ def __len__(self):
 
 with assert_raises(TypeError):
     bool(TestLenThrowError())
+
+# Issue #3567: nested BoolOps should not call __bool__ redundantly
```
youknowone: this is more like short-circuit than bool. we have syntax_short_circuit_bool.py for that

moreal: I didn't know existence of the file. I moved and simplified the testcase in https://github.com/RustPython/RustPython/pull/7176/changes/c1e7986264b10dbb2271b882bd0bae154d860db2 commit.

```python
# Issue #3567: nested BoolOps should not call __bool__ redundantly
ExplodingBool(False) and False or False
```

extra_tests/snippets/syntax_short_circuit_bool.py
```diff
@@ -31,3 +31,6 @@ def __bool__(self):
 
 # if ExplodingBool(False) and False and True and False:
 #     pass
+
+# Issue #3567: nested BoolOps should not call __bool__ redundantly
+ExplodingBool(False) and False or False
```
youknowone:
```suggestion
assert ExplodingBool(False) and False or False
```

moreal:
As a note about [d6e48ff](https://github.com/RustPython/RustPython/pull/7176/commits/d6e48ff446117f4be494a1b44d39d7b9a61f456b), `assert not (ExplodingBool(False) and False or False)` doesn't raise AssertionError in the current main branch. So I used `==` comparison which raises AssertionError. `assert (ExplodingBool(False) and False or False) == False`

PR #7232, opened 2026-02-27
Fix IOBase.fileno() incorrect error message

ever0de:
fileno() was calling _unsupported with "truncate" instead of "fileno", causing UnsupportedOperation('truncate') to be raised instead of UnsupportedOperation('fileno').

CPython 3.14 reference: Modules/_io/iobase.c calls iobase_unsupported(state, "fileno")

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed error message for unsupported file operations to display the correct operation name instead of an incorrect reference.


youknowone (approved):
👍

PR #7233, opened 2026-02-27
Fix IOBase.isatty() to raise ValueError on closed files

ever0de:
`test_io_after_close` in `Lib/test/test_io.py` calls `f.isatty()` on a closed file, but `f` comes from `io.open()` — a `TextIOWrapper` or `BufferedWriter`. Those classes each have their own `isatty()` that checks for closure, so this test never touches `IOBase.isatty()` at all.

The old code only broke when a subclass didn't override `isatty()`:
```python
class MinimalRaw(io.RawIOBase):
     def readinto(self, b): return 0

f = MinimalRaw()
f.close()
f.isatty()  # returned False; should raise ValueError
```

The fix adds `check_closed()` before returning `False`, matching what CPython 3.14 does in `_io__IOBase_isatty_impl`. `extra_tests/snippets/stdlib_io_isatty.py` tests this path directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the `isatty()` method for IO objects to properly raise an error when called on closed instances, replacing the previous simple behavior.

* **Tests**
  * Added test coverage validating `isatty()` returns False for open IO instances and raises an error for closed instances.


extra_tests/snippets/stdlib_io_isatty.py
youknowone, line 1: Does this file worth to have its own test file? how about appending to `stdlib_io`?

youknowone (changes requested):

extra_tests/snippets/stdlib_io_isatty.py
ever0de: At first, I was planning to work in `stdlib_io`, but I thought it was related to FileIO so I separated the test file. I have now merged them. Thank you!

PR #7237, opened 2026-02-27
sqlite3: handle int type in Connection.timeout parameter

ever0de:
## Summary by CodeRabbit

* **Improvements**
  * SQLite connection timeout now accepts both floating‑point and integer values, making it easier to supply timeouts in seconds or as whole numbers. The change is backward compatible: existing defaults and behavior are preserved while allowing mixed input types for more flexible configuration.

### DIFF
--- a/crates/stdlib/src/_sqlite3.rs
+++ b/crates/stdlib/src/_sqlite3.rs
@@ -61,7 +61,7 @@ mod _sqlite3 {
         },
         convert::IntoObject,
         function::{
-            ArgCallable, ArgIterable, FsPath, FuncArgs, OptionalArg, PyComparisonValue,
+            ArgCallable, ArgIterable, Either, FsPath, FuncArgs, OptionalArg, PyComparisonValue,
             PySetterValue,
         },
         object::{Traverse, TraverseFn},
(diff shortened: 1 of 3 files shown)


crates/stdlib/src/_sqlite3.rs
```diff
@@ -333,8 +333,8 @@ mod _sqlite3 {
     struct ConnectArgs {
         #[pyarg(any)]
         database: FsPath,
-        #[pyarg(any, default = 5.0)]
-        timeout: f64,
+        #[pyarg(any, default = Either::A(5.0))]
+        timeout: Either<f64, i64>,
```
youknowone, line 337: Parsing timeout as `Either<f64, i64>` is a common pattern in Python. If you are interested in, investigating if introducing TimeoutSeconds utility type will be useful or not. (I am not sure yet)

ever0de: I also think it's a great idea. Do you think it would be better to separate it into a new PR instead of working on it here?

youknowone: sure, in separate PR. it is unrelated refactoring

PR #7242, opened 2026-02-27
Update `xml` to 3.14.3

ShaharNaveh:


Lib/test/test_pulldom.py
```diff
@@ -23,8 +23,8 @@
 
 
 class PullDOMTestCase(unittest.TestCase):
-    # TODO: RUSTPYTHON FileNotFoundError: [Errno 2] No such file or directory (os error 2): 'xmltestdata/test.xml' -> 'None'
-    @unittest.expectedFailure
+
+    @unittest.expectedFailure  # TODO: RUSTPYTHON; FileNotFoundError: [Errno 2] No such file or directory (os error 2): 'xmltestdata/test.xml' -> 'None'
```
fanninpm, line 27: We might need to take a look at fixing the `findfile()` function in `test.support`.

PR #7278, opened 2026-02-28
sqlite3: add Cursor.row_factory pygetset and fix Connection.cursor() …

ever0de:
…propagation

Add #[pygetset] getter/setter for Cursor.row_factory so that Python-level attribute access reads/writes the Rust struct field instead of the instance dict.

Fix Connection.cursor() to only propagate the connection's row_factory to the cursor when the connection's row_factory is not None, matching CPython behavior. Previously it unconditionally overwrote the cursor's row_factory, discarding any factory set by a cursor subclass __init__.

## Release Notes

* **New Features**
  * Cursor objects now provide public methods to get and set row factory configurations.

* **Bug Fixes**
  * Improved row factory handling during cursor initialization for increased reliability.

### DIFF
--- a/Lib/test/test_sqlite3/test_factory.py
+++ b/Lib/test/test_sqlite3/test_factory.py
@@ -112,7 +112,6 @@ class CursorFactoryTests(MemoryDatabaseMixin, unittest.TestCase):
 
 class RowFactoryTestsBackwardsCompat(MemoryDatabaseMixin, unittest.TestCase):
 
-    @unittest.expectedFailure # TODO: RUSTPYTHON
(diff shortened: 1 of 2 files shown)


fanninpm:
> Error: 'frozenset' object has no attribute 'discard'

@ShaharNaveh you may want to take a look into this

ShaharNaveh:
> > Error: 'frozenset' object has no attribute 'discard'
> 
> @ShaharNaveh you may want to take a look into this

https://github.com/RustPython/RustPython/pull/7276 :)

PR #7280, opened 2026-02-28
Fix _hashlib.compare_digest to reject non-ASCII strings

ever0de:
Add non-ASCII string check to _hashlib.compare_digest, matching the behavior of _operator._compare_digest. When both arguments are strings, non-ASCII characters now correctly raise TypeError.

Also replace the non-constant-time == comparison with constant_time_eq for proper timing-attack resistance, and return PyResult<bool> instead of PyResult<PyObjectRef>.

## Summary by CodeRabbit

* **Chores**
  * Added a dependency to support constant-time cryptographic comparisons.
* **Bug Fixes**
  * Digest comparison now yields a direct boolean result and uses constant-time comparison under the hood.
  * Stricter input validation with clearer errors for non-ASCII strings and incompatible/bytes-like inputs.

### DIFF
--- a/Lib/test/test_hmac.py
+++ b/Lib/test/test_hmac.py
@@ -1484,18 +1484,10 @@ class HMACCompareDigestTestCase(CompareDigestMixin, unittest.TestCase):
         else:
             self.assertIs(self.compare_digest, operator_compare_digest)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON; AssertionError: TypeError not raised by compare_digest
-    def test_exceptions(self):
-        return super().test_exceptions()
-
 
 @hashlib_helper.requires_hashlib()
 class OpenSSLCompareDigestTestCase(CompareDigestMixin, unittest.TestCase):
     compare_digest = openssl_compare_digest
-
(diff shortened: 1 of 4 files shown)


Lib/test/test_hmac.py
```diff
@@ -1484,7 +1484,6 @@ def test_compare_digest_func(self):
         else:
             self.assertIs(self.compare_digest, operator_compare_digest)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON; AssertionError: TypeError not raised by compare_digest
     def test_exceptions(self):
         return super().test_exceptions()
```
ShaharNaveh: can you please remove the entire test?

Lib/test/test_hmac.py
```diff
@@ -1493,7 +1492,6 @@ def test_exceptions(self):
 class OpenSSLCompareDigestTestCase(CompareDigestMixin, unittest.TestCase):
     compare_digest = openssl_compare_digest
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON; AssertionError: TypeError not raised by compare_digest
     def test_exceptions(self):
         return super().test_exceptions()
```
ShaharNaveh: same:)

PR #7310, opened 2026-03-02
Prevent shell injection

moreal:
Since `github.event.pull_request.head.ref` can be defined by anyone, it can be used in shell injection.
This pull request ensures those values aren't executed as shell scripts by setting them as `env` variables and using parameter expansion syntax to retrieve them.

## Related resources

- https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

## Summary by CodeRabbit

* **Chores**
  * Refactored GitHub Actions workflows to use environment variables for improved configuration management and reliability across release, documentation, and commit automation processes.


fanninpm:
@varunsh-coder does this actually fix your problem? If so, you can close #7311.

PR #7313, opened 2026-03-02
Add method cache for MRO lookups

youknowone:
4096-entry direct-mapped cache keyed by (tp_version_tag, interned_name_ptr), mirroring CPython's type_cache in Objects/typeobject.c.

- find_name_in_mro() checks cache before MRO walk
- Auto-assigns version tags on cache miss (as in _PyType_Lookup)
- Negative caching for attribute-not-found results
- has_name_in_mro() avoids cloning on cache hit
- Invalidation via existing modified() mechanism (tp_version_tag=0)
- get_attr() and has_attr() now route through cached path

## Summary by CodeRabbit

* **Performance**
  * Faster type attribute lookups via a new inline type-attribute cache.
  * Cache is automatically cleared during garbage collection to avoid memory retention and improve cycle collection.

* **New Features**
  * Added a public API to clear the type-attribute cache manually.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -77,6 +77,7 @@
     "kwonly",
     "lossily",
     "makeunicodedata",
+    "mcache",
     "microbenchmark",
     "microbenchmarks",
     "miri",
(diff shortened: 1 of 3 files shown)


youknowone:
@coderabbitai review

PR #7329, opened 2026-03-03
Add vectorcall (PEP 590) dispatch for function calls

youknowone:
Add VectorCallFunc slot to PyTypeSlots and vectorcall dispatch path in the interpreter loop for Call and CallKw instructions.

Implement vectorcall for PyFunction (with fast path for simple positional-only calls that fills fastlocals directly), PyBoundMethod (avoids prepend_arg O(n) shift), and PyNativeFunction.

Add FuncArgs::from_vectorcall helper for fallback conversion. Vectorcall slot is inherited with call slot and cleared when __call__ is overridden in Python subclasses.

## Release Notes

* **Refactor**
  * Implemented vectorcall (PEP 590) support throughout the VM for more efficient function call handling, reducing memory allocations and improving function call performance across builtin functions, regular functions, bound methods, and frame execution.

### DIFF
--- a/crates/vm/src/builtins/builtin_func.rs
+++ b/crates/vm/src/builtins/builtin_func.rs
@@ -243,21 +243,10 @@ fn vectorcall_native_function(
 
     let func_args = if needs_self {
         let self_obj = zelf.zelf.as_ref().unwrap().clone();
-        let total_pos = nargs + 1;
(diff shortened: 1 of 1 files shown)


ShaharNaveh:
Super excited about this one ngl

PR #7334, opened 2026-03-03
Object header slimming: prefix allocation for ObjExt

youknowone:
Extract dict, weak_list, and slots fields from PyInner<T> into a separate ObjExt struct allocated as a prefix before PyInner using Layout::extend(). Objects that don't need these fields (int, str, float, list, tuple, dict, etc.) skip the prefix entirely.

- Add HAS_WEAKREF flag to PyTypeFlags for per-type weakref control
- Add HAS_EXT bit to GcBits indicating prefix presence
- Define ObjExt struct with dict, weak_list, slots
- Shrink PyInner header from ~80-88 bytes to ~32 bytes for lightweight objects
- Update all accessor methods to go through ext_ref()
- Update bootstrap type hierarchy to use prefix allocation
- Add __weakref__ getset descriptor for heap types
- Set HAS_WEAKREF on builtin types that support weak references
- Remove test_weak_keyed_bad_delitem expectedFailure (now passes)

## Summary by CodeRabbit

* **New Features**
  * Broad weak-reference support added to many built-in types (functions, methods, modules, generators/coroutines/async generators, collections, I/O types, memoryview, sets, properties, type objects and type-parameter classes), enabling creation of weak refs to those instances without changing runtime behavior.
* **Chores**
  * Internal object layout and type initialization updated to consistently support weak-reference metadata.

### DIFF
--- a/.cspell.dict/rust-more.txt
+++ b/.cspell.dict/rust-more.txt
@@ -6,6 +6,7 @@ bindgen
 bitand
 bitflags
 bitor
+bitvec
 bitxor
 bstr
 byteorder
@@ -58,6 +59,7 @@ powi
 prepended
 punct
 replacen
+retag
 rmatch
 rposition
 rsplitn
@@ -89,5 +91,3 @@ widestring
 winapi
 winresource
 winsock
-bitvec
-Bitvec
--- a/crates/vm/src/builtins/builtin_func.rs
+++ b/crates/vm/src/builtins/builtin_func.rs
@@ -16,6 +16,8 @@ pub struct PyNativeFunction {
     pub(crate) value: &'static PyMethodDef,
     pub(crate) zelf: Option<PyObjectRef>,
     pub(crate) module: Option<&'static PyStrInterned>, // None for bound method
+    /// Prevent HeapMethodDef from being freed while this function references it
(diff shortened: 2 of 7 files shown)


Lib/test/libregrtest/run_workers.py
```diff
@@ -347,7 +347,10 @@ def read_json(self, json_file: JsonFile, json_tmpfile: TextIO | None,
         except Exception as exc:
             # gh-101634: Catch UnicodeDecodeError if stdout cannot be
             # decoded from encoding
-            err_msg = f"Failed to parse worker process JSON: {exc}"
+            # TODO: RUSTPYTHON; include JSON for debugging flaky worker bugs
```
youknowone: revert test code

Lib/test/libregrtest/result.py
```diff
@@ -238,6 +238,15 @@ def _decode_test_result(data: dict[str, Any]) -> TestResult | dict[str, Any]:
             data['covered_lines'] = [
                 tuple(loc) for loc in data['covered_lines']
             ]
-        return TestResult(**data)
+        # TODO: RUSTPYTHON; diagnostics for flaky worker bug
```
youknowone: revert test code

PR #7345, opened 2026-03-04
Add freelists for PyComplex, PyInt, PyRange

youknowone:
## Summary
- Add freelists for 4 more types matching CPython: PyComplex(100), PyInt(100), PyBoundMethod(20), PyRange(6)
- Use `try_with` instead of `with` in all freelist `push`/`pop` (including existing float/list/dict/slice) to gracefully handle thread-local access during thread teardown

Follow-up to #7337.

## Test plan
- [x] `cargo build --release`
- [x] `cargo clippy` clean
- [x] `test_complex`, `test_int`, `test_range`, `test_class`, `test_descr` all pass

## Release Notes

* **Performance**
  * Improved memory efficiency for numeric and collection objects through optimized object reuse mechanisms.

* **Bug Fixes**
  * Enhanced error handling in memory operations to gracefully degrade under resource contention instead of crashing.

### DIFF
--- a/crates/vm/src/builtins/function.rs
+++ b/crates/vm/src/builtins/function.rs
@@ -20,8 +20,6 @@ use crate::{
         Callable, Comparable, Constructor, GetAttr, GetDescriptor, PyComparisonOp, Representable,
     },
 };
-use core::cell::Cell;
-use core::ptr::NonNull;
 use core::sync::atomic::{AtomicU32, Ordering::Relaxed};
 use itertools::Itertools;
 #[cfg(feature = "jit")]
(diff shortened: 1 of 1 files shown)


youknowone:
## Review response

### Applied

1. **FreeList doc comment** (CodeRabbit): Updated to reflect actual Drop behavior — raw dealloc without running destructors to avoid TLS-after-destruction.

2. **PyBoundMethod `clear = false`** (CodeRabbit Critical / Codex P1): Added `clear = false` to `#[pyclass]` attribute. CPython's `classobject.c` has `tp_clear = 0` (NULL) for bound method objects — they are traversed but never cleared.

### Not applied (with reasoning)

3. **PyRange clear** (Codex P2): CPython's `rangeobject.c` has both `tp_traverse = 0` and `tp_clear = 0` — range objects are not GC-tracked at all. Our PyRange doesn't have traverse either, so no clear is needed.

4. **FreeList::Drop leak concern** (CodeRabbit Major): The raw dealloc path only runs during **thread teardown**, not during normal object lifecycle. During normal operation, objects are properly deallocated with full destructors. Thread teardown leaks are bounded and intentional — same design as CPython which also doesn't free freelist objects on thread destruction.

— *commented by Claude*

PR #7388, opened 2026-03-09
Use matrix for `cargo check`

ShaharNaveh:
## Summary by CodeRabbit

* **Chores**
  * Reduced Linux install footprint by using non-recommending package installs to avoid extra packages.
  * Added additional ARM64 cross-compilation runtime/dev packages when cross-targeting AArch64.
  * Consolidated CI into a matrix-driven compilation check across platforms, centralizing tool/setup steps and per-target build checks to remove duplication.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -240,6 +240,14 @@ jobs:
           ndk-version: r27
           add-to-path: true
 
+      # - name: Prepare repository for redox compilation
+      #   run: bash scripts/redox/uncomment-cargo.sh
+      # - name: Check compilation for Redox
+      #   uses: coolreader18/redoxer-action@v1
+      #   with:
+      #     command: check
+      #     args: --ignore-rust-version
+
       - name: Check compilation
         run: |
           for target in ${{ join(matrix.targets, ' ') }}


.github/workflows/ci.yaml
```diff
@@ -188,102 +188,71 @@ jobs:
...
-        run: cargo check --target x86_64-unknown-freebsd ${{ env.CARGO_ARGS_NO_SSL }}
-
-      - uses: dtolnay/rust-toolchain@stable
-        with:
-          target: wasm32-wasip2
-
-      - name: Check compilation for wasip2
-        run: cargo check --target wasm32-wasip2 ${{ env.CARGO_ARGS_NO_SSL }}
-
-      # - name: Prepare repository for redox compilation
```
youknowone, line 279: where should redox to be put?

ShaharNaveh: In one of the target groups. I can try to do in a separate PR. this PR isn't meant to change existing behavior.

I assume we will have  a special condition for it like we do in the setup-ndk step

youknowone: redox is(was?) actually using rustpython for their python. This build was guaranteeing it is working.
I forgot when this is commented out and not fixed back, (probably redoxer action is broken) but we at least need to keep the comment to revive it later.

PR #7402, opened 2026-03-12
If the `host_env` feature is enabled, then enable the `_ctypes` module on the Android platform.

psyche314:
## Summary by CodeRabbit

* **New Features**
  * Added Android platform support.

* **Chores**
  * Updated native library dependencies for expanded platform support.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ license.workspace = true
 [features]
 default = ["threading", "stdlib", "stdio", "importlib", "ssl-rustls", "host_env"]
 host_env = ["rustpython-vm/host_env", "rustpython-stdlib?/host_env"]
-ctypes = ["rustpython-vm/ctypes"]
 importlib = ["rustpython-vm/importlib"]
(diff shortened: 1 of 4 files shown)


ShaharNaveh:
@youknowone should we start to officially support android as a platform? (like cpython does it from 3.13)

ShaharNaveh (changes requested):
Just line @youknowone said, I don't see the benefit of having a feature of `ctypes`, but to gate it behind the `host_env` feature, unless we would want to go with a feature gate per module (which has it's pros&cons)

youknowone:
How do you define official and unofficial platform? If you mean enhancing CI to guarantee it works, I agree.

ShaharNaveh:
> How do you define official and unofficial platform? If you mean enhancing CI to guarantee it works, I agree.

https://peps.python.org/pep-0738/

---

And sure, I'll try to implement a CI step that checks CI on android as well

psyche314:
Okay, let's simply add `target_os = "android"` to avoid introducing a new `ctypes` feature.

PR #7410, opened 2026-03-12
Fix weakref proxy number protocol delegation

rlaisqls:
- Weakref proxy was missing number protocol forwarding — arithmetic ops like `proxy + 5`, `int(proxy)`, `operator.index(proxy)` all raised `TypeError`
- Implement all number slots (unary, binary, inplace, ternary) by delegating to the underlying object's type slots
- Handle right-side operations (`5 + proxy`, `5 @ proxy`) by detecting which side the proxy is on and using the appropriate `right_* slot`
+ chore: Fix non-iterator proxy error message to match CPython: `"Weakref proxy referenced a non-iterator"`

## Summary by CodeRabbit

* **New Features**
  * Weak proxy objects now support full numeric operations (unary, binary, and ternary) when the target provides them, enabling arithmetic and numeric protocol behavior through proxies.

* **Improvements**
  * More robust operator dispatch with correct operand resolution and clearer errors for unsupported numeric operations on proxies.
  * Runtime validation for proxy-based iteration with explicit errors when the referent is not a valid iterator.

### DIFF
--- a/crates/vm/src/builtins/weakproxy.rs
+++ b/crates/vm/src/builtins/weakproxy.rs
@@ -5,11 +5,11 @@ use crate::{
     class::PyClassImpl,
     common::hash::PyHash,
     function::{OptionalArg, PyComparisonValue, PySetterValue},
-    protocol::{PyIter, PyIterReturn, PyMappingMethods, PyNumber, PyNumberMethods, PySequenceMethods},
+    protocol::{PyIter, PyIterReturn, PyMappingMethods, PyNumberMethods, PySequenceMethods},
     stdlib::builtins::reversed,
     types::{
-        AsMapping, AsNumber, AsSequence, Comparable, Constructor, GetAttr, Hashable,
-        IterNext, Iterable, PyComparisonOp, Representable, SetAttr,
+        AsMapping, AsNumber, AsSequence, Comparable, Constructor, GetAttr, Hashable, IterNext,
+        Iterable, PyComparisonOp, Representable, SetAttr,
     },
 };
 
@@ -136,9 +136,7 @@ impl IterNext for PyWeakProxy {
     fn next(zelf: &Py<Self>, vm: &VirtualMachine) -> PyResult<PyIterReturn> {
         let obj = zelf.try_upgrade(vm)?;
         if obj.class().slots.iternext.load().is_none() {
-            return Err(vm.new_type_error(
-                "Weakref proxy referenced a non-iterator".to_owned(),
-            ));
+            return Err(vm.new_type_error("Weakref proxy referenced a non-iterator".to_owned()));
         }
         PyIter::new(obj).next(vm)
     }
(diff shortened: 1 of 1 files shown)


crates/vm/src/builtins/weakproxy.rs
```diff
@@ -93,6 +93,7 @@ impl PyWeakProxy {
         self.try_upgrade(vm)?.bytes(vm)
     }
 
+
```
youknowone:
```suggestion
```

crates/vm/src/builtins/weakproxy.rs
```diff
@@ -166,14 +172,156 @@ impl SetAttr for PyWeakProxy {
...
 
+fn proxy_upgrade(obj: &PyObject, vm: &VirtualMachine) -> PyResult {
+    obj.downcast_ref::<PyWeakProxy>()
+        .unwrap()
+        .try_upgrade(vm)
+}
+
+fn proxy_upgrade_opt(obj: &PyObject, vm: &VirtualMachine) -> PyResult<Option<PyObjectRef>> {
+    if obj.downcast_ref::<PyWeakProxy>().is_some() {
+        Ok(Some(proxy_upgrade(obj, vm)?))
```
youknowone: calling proxy_upgrade inside this function will cost double for `obj.downcast_ref::<PyWeakProxy>()`

crates/vm/src/builtins/weakproxy.rs
```diff
@@ -166,14 +172,156 @@ impl SetAttr for PyWeakProxy {
...
+                }
+            } else {
+                if let Some(f) = b_ref.class().slots.as_number.$right_slot.load() {
+                    f(a_ref, b_ref, c_ref, vm)
+                } else {
+                    Ok(vm.ctx.not_implemented())
+                }
+            }
+        })
+    };
```
youknowone, line 238: Could you please check if we can retrieve helper functions to minimize the inner side of macro_rules? Small macro footprints help debugging later.

crates/vm/src/builtins/weakproxy.rs
```diff
@@ -166,14 +172,156 @@ impl SetAttr for PyWeakProxy {
...
+            let obj = proxy_upgrade(number.obj, vm)?;
+            let f = obj.class().slots.as_number.$slot.load()
+                .ok_or_else(|| vm.new_type_error(format!(
+                    "bad operand type for unary op: '{}'",
+                    obj.class().name()
+                )))?;
+            let number = PyNumber { obj: &obj };
+            f(number, vm).map(|v| v.into())
+        })
+    };
```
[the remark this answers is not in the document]
rlaisqls: I wasn't aware of the None vs Some(Err) distinction in the fallback chain. Changed int/float/index to call `try_int()`/`try_float()`/`try_index()` on the upgraded object directly, so the full conversion chain works as expected.

crates/vm/src/builtins/weakproxy.rs
rlaisqls: Thanks, I didn't think of that.. Extracted `proxy_binary_op` and `proxy_ternary_op` helpers. Also switched from raw slot calls to VM dispatch (`vm._add()`, etc.), which removed the two-form $right_slot variants since the VM already handles reflected method ordering.

crates/vm/src/builtins/weakproxy.rs
rlaisqls: Thank you for the review. Fixed by using `match` to capture the `downcast_ref` result directly instead of downcasting again inside `proxy_upgrade`.

crates/vm/src/builtins/weakproxy.rs
```diff
@@ -166,14 +172,156 @@ impl SetAttr for PyWeakProxy {
...
+            } else {
+                // Proxy on the right: use right slot
+                if let Some(f) = b_ref.class().slots.as_number.$right_slot.load() {
+                    f(a_ref, b_ref, vm)
+                } else {
+                    Ok(vm.ctx.not_implemented())
+                }
+            }
+        })
+    };
```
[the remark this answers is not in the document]
rlaisqls: Fixed. Delegating through VM dispatch (vm._add(), etc.) naturally handles the subclass reflected method priority, so this is resolved together.

crates/vm/src/builtins/weakproxy.rs
```diff
@@ -166,14 +171,127 @@ impl SetAttr for PyWeakProxy {
...
+}
+
+fn proxy_upgrade_opt(obj: &PyObject, vm: &VirtualMachine) -> PyResult<Option<PyObjectRef>> {
+    match obj.downcast_ref::<PyWeakProxy>() {
+        Some(proxy) => Ok(Some(proxy.try_upgrade(vm)?)),
+        None => Ok(None),
+    }
+}
+
+macro_rules! proxy_unary_slot {
```
youknowone, line 194: proxy_binary_op and proxy_ternary_op are looking good, This one is missing.

youknowone (changes requested):

crates/vm/src/builtins/weakproxy.rs
rlaisqls: I missed the unary case, fixed it to follow the same pattern as the other two 🙇

PR #7412, opened 2026-03-12
Enable PEP 709 inlined comprehensions

youknowone:
## Summary

- Activate the existing `compile_inlined_comprehension()` implementation by fixing 6 bugs that prevented it from working
- List/dict/set comprehensions inside function-like scopes are now compiled inline (no separate code object), matching CPython 3.12+ behavior
- Generator expressions remain non-inlined per PEP 709 spec

## Changes

**`frame.rs`**: `LoadFastAndClear` pushes NULL when slot is empty (not None), `StoreFast` accepts NULL for restore path

**`compile.rs`**: Enable `is_inlined_comprehension_context()`, fix `in_inlined_comp` flag placement, fix `sub_tables.remove(0)` → `next_sub_table` cursor, splice nested sub_tables into parent

**`symboltable.rs`**: Set `comp_inlined` only in function-like parent scopes, conflict detection (nested scopes, bound name conflicts, cross-comprehension reference conflicts)

## Current limitations

- Module/class level comprehensions are not inlined (STORE_NAME incompatible with LOAD_FAST_AND_CLEAR)
- Comprehensions containing nested scopes (lambda, inner function, inner comprehension) are not inlined (Cell/Free handling not yet supported)
- Bound name conflicts with parent scope → falls back to non-inlined

## Test plan

- [x] `test_listcomps` — 59/59 pass
- [x] `test_setcomps` — pass
- [x] `test_symtable` — pass
- [x] `test_dictcomps` — 1 pre-existing failure (unrelated)
- [x] `test_compile`, `test_syntax` — pass
- [x] `cargo test -p rustpython-codegen -p rustpython-compiler` — pass
- [x] Variable isolation, outer references, nested comprehensions, walrus operator — all verified

## Summary by CodeRabbit

* **Refactor**
  * Comprehension inlining tightened: only inline in function-like scopes, skip generator/async comprehensions, and preserve inlined state for nested comprehensions and lambdas.
  * Symbol-table merging hardened: track symbols added by inlining, prevent name conflicts and improper resolutions, and refine scope propagation rules.
  * Runtime locals behavior adjusted: cleared/empty local slots use a nullable representation, aligning push/pop semantics and preserving variable lifecycle consistency.

* **Bug Fixes**
  * Improved handling of scope visibility to avoid incorrect symbol resolution after inlining.

### DIFF
--- a/crates/codegen/src/symboltable.rs
+++ b/crates/codegen/src/symboltable.rs
@@ -299,6 +299,26 @@ fn drop_class_free(symbol_table: &mut SymbolTable, newfree: &mut IndexSet<String
     }
 }
 
+/// Check if an expression contains an `await` node (shallow, not into nested scopes).
+fn expr_contains_await(expr: &ast::Expr) -> bool {
+    use ast::visitor::Visitor;
+    struct AwaitFinder(bool);
+    impl ast::visitor::Visitor<'_> for AwaitFinder {
+        fn visit_expr(&mut self, expr: &ast::Expr) {
+            if !self.0 {
+                if matches!(expr, ast::Expr::Await(_)) {
+                    self.0 = true;
+                } else {
+                    ast::visitor::walk_expr(self, expr);
+                }
+            }
+        }
+    }
+    let mut finder = AwaitFinder(false);
+    finder.visit_expr(expr);
+    finder.0
+}
+
 /// PEP 709: Merge symbols from an inlined comprehension into the parent scope.
 /// Matches symtable.c inline_comprehension().
 fn inline_comprehension(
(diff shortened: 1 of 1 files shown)


moreal:
We may need to rebase this branch on the main branch because some commits seem like already merged by https://github.com/RustPython/RustPython/pull/7354

ShaharNaveh:
Better to merge after #7427

crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -302,6 +302,13 @@ pub enum Instruction {
     YieldValue {
         arg: Arg<u32>,
     } = 120,
+    /// PEP 709: Restore a saved cell object after an inlined comprehension.
+    /// Pops the old cell object from the stack and writes it back to the
+    /// cell slot at `fastlocals[nlocals + i]`, replacing the temporary cell
+    /// created by `MakeCell`.
+    RestoreCell {
```
ShaharNaveh: How does CPython do it? can we not break the full CPython 3.14 bytecode compatibility?

youknowone: I am sorry about this. let me check it.

PR #7416, opened 2026-03-13
Thread-safe type operations: type lock, QSBR type-cache reclamation, GC stop-the-world, and interpreter optimizations

youknowone:
Introduce vm.state.type_mutex to serialize type mutation and version-tag assignment, matching CPython's free-threading model.

- Add with_type_lock() helper and split modified/assign_version_tag into inner (lock-free) and public (lock-acquiring) variants
- Add version_for_specialization() and lookup_ref_and_version_interned() for atomic type-cache lookups under the lock
- Wrap all type-mutating paths (__bases__, __annotations__, __module__, __type_params__, __doc__, SetAttr) with the type lock
- Drop old values outside the lock to prevent deadlock from weakref callbacks that may re-enter specialization
- Fix _ctypes set_attr calls to route through proper SetAttr path
- Reinitialize type_mutex after fork

## Summary by CodeRabbit

* **Bug Fixes**
  * Centralized type-mutation synchronization to reduce races and deadlock risk during bases, annotations, and attribute updates.

* **Performance**
  * More consistent type-versioning and cache lookups, enabling faster and more reliable specialization fastpaths and method/attribute retrievals.

* **Chores**
  * Added and initialized a global type mutex at VM startup and ensured it is reinitialized after fork to support the new synchronization.


crates/vm/src/builtins/type.rs
```diff
@@ -1334,38 +1422,41 @@ impl PyType {
...
-                PyType::resolve_mro(&cls.bases.read()).map_err(|msg| vm.new_type_error(msg))?;
-            // Preserve self (mro[0]) when updating MRO
-            mro.insert(0, cls.mro.read()[0].to_owned());
-            *cls.mro.write() = mro;
-            for subclass in cls.subclasses.write().iter() {
-                let subclass = subclass.upgrade().unwrap();
-                let subclass: &Py<PyType> = subclass.downcast_ref().unwrap();
-                update_mro_recursively(subclass, vm)?;
+        Self::with_type_lock(vm, || {
+            *zelf.bases.write() = bases;
```
youknowone: do we still need bases lock?

crates/vm/src/builtins/type.rs
```diff
@@ -1334,38 +1422,41 @@ impl PyType {
...
+            // Recursively update the mros of this class and all subclasses
+            fn update_mro_recursively(cls: &PyType, vm: &VirtualMachine) -> PyResult<()> {
+                let mut mro =
+                    PyType::resolve_mro(&cls.bases.read()).map_err(|msg| vm.new_type_error(msg))?;
+                // Preserve self (mro[0]) when updating MRO
+                mro.insert(0, cls.mro.read()[0].to_owned());
+                *cls.mro.write() = mro;
```
youknowone: mro?

PR #7421, opened 2026-03-13
Preserve imaginary zero signs when adding real values to complex numbers

moreal:
This pull request lets `complex + float` operations preserve the zero sign of the `imag` part.

Arithmetic operations use `PyComplex::number_op`, which converts `float` (real) to `Complex64` to simplify logic, but it loses the zero sign. For instance, in CPython, `complex(1, -0.0j) + 1` becomes `2-0j`, and `complex(1, -0.0j) + complex(1, 0.0j)` becomes `2+0j`. However, `PyComplex::number_op` converts `float` to `complex`, so it becomes `complex(1, 0.0j)`, which differs from CPython.

## Summary by CodeRabbit

* **Refactor**
  * Internal consolidation of complex-number arithmetic logic for improved maintainability and consistency.
  * No changes to public APIs or observable behavior; users should see no functional differences.

### DIFF
--- a/crates/vm/src/builtins/complex.rs
+++ b/crates/vm/src/builtins/complex.rs
@@ -297,21 +297,21 @@ impl PyComplex {
         vm: &VirtualMachine,
     ) -> PyResult
     where
-        CCF: FnOnce(Complex64, Complex64, &VirtualMachine) -> R,
-        CRF: FnOnce(Complex64, f64, &VirtualMachine) -> R,
-        RCF: FnOnce(f64, Complex64, &VirtualMachine) -> R,
+        CCF: FnOnce(Complex64, Complex64) -> R,
+        CRF: FnOnce(Complex64, f64) -> R,
+        RCF: FnOnce(f64, Complex64) -> R,
         R: ToPyResult,
     {
         let value = match (a.downcast_ref::<PyComplex>(), b.downcast_ref::<PyComplex>()) {
             // complex + complex
-            (Some(a_complex), Some(b_complex)) => cc_op(a_complex.value, b_complex.value, vm),
+            (Some(a_complex), Some(b_complex)) => cc_op(a_complex.value, b_complex.value),
             (Some(a_complex), None) => {
                 let Some(b_real) = float::to_op_float(b, vm)? else {
                     return Ok(vm.ctx.not_implemented());
                 };
 
                 // complex + real
-                cr_op(a_complex.value, b_real, vm)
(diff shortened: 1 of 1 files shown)


crates/vm/src/builtins/complex.rs
```diff
@@ -396,7 +421,7 @@ impl Hashable for PyComplex {
 impl AsNumber for PyComplex {
     fn as_number() -> &'static PyNumberMethods {
         static AS_NUMBER: PyNumberMethods = PyNumberMethods {
-            add: Some(|a, b, vm| PyComplex::number_op(a, b, |a, b, _vm| a + b, vm)),
+            add: Some(complex_add),
             subtract: Some(|a, b, vm| PyComplex::number_op(a, b, |a, b, _vm| a - b, vm)),
```
youknowone: how about subtract? is this okay without change?

moreal: In subtract case, it may work because `-0.0 - 0.0` (i.e., "minus-sign zero" - "plus-sign zero") is evaluated as `-0.0`.

``` rust
fn main() {
    let val: f64 = -0f64 - 0f64;
    println!("{val:?}"); // -0.0
}

fn main() {
    let val: f64 = -0f64 + 0f64;
    println!("{val:?}"); // 0.0
}
```

youknowone:
```
$ python3 -c 'print(1 - (1+0j))'
-0j
$ cargo run --  -c 'print(1 - (1+0j))'
0j
```

moreal: I didn't check `real - complex`. I'll look into it more.

moreal: The issue was that only the real part should have been touched, but because of the conversion process to Complex64, the imaginary part was also being affected; so it was actually strange that there were no issues with subtraction. Thank you.

In CPython, these are handled by writing separate logic for each case, similar to `_Py_{c,cr,rc}_xx`. Accordingly, I created a helper function called `complex_real_binop`, like `number_op`, and used it to write separate logic for each case in addition and subtraction (https://github.com/RustPython/RustPython/pull/7421/changes/df6061723155b423bdc6fcbe3feeb4248ab7a768). (As for the function naming... if you have any alternatives, please let me know.)

```
$ cargo run -q --  -c 'print(1 - (1+0j))'
-0j
```

youknowone (approved):
👍

PR #7431, opened 2026-03-14
Newtype `var_nums` oparg

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Refactored variable index extraction across bytecode instructions, replacing manual bit-level operations with a unified accessor method for improved consistency.
  * Updated instruction field handling to use a consolidated variable numbering approach.
  * Streamlined index encoding and retrieval in the runtime and JIT compiler for better maintainability.

### DIFF
--- a/crates/jit/src/instructions.rs
+++ b/crates/jit/src/instructions.rs
@@ -660,10 +660,9 @@ impl<'a, 'b> FunctionCompiler<'a, 'b> {
             Instruction::LoadFastLoadFast { var_nums }
             | Instruction::LoadFastBorrowLoadFastBorrow { var_nums } => {
                 let oparg = var_nums.get(arg);
-                let idx1 = oparg >> 4;
-                let idx2 = oparg & 0xF;
+                let (idx1, idx2) = oparg.indexes();
                 for idx in [idx1, idx2] {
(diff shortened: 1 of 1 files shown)


crates/compiler-core/src/bytecode/oparg.rs
```diff
@@ -733,36 +733,41 @@ newtype_oparg!(
 newtype_oparg!(
     #[derive(Clone, Copy)]
     #[repr(transparent)]
-    pub struct LoadAttr(u32)
+    pub struct VarNums(u32)
```
youknowone, line 736: This is actually a pair of u16, right? can this be written using union?

ShaharNaveh: Maybe this can be written as a union, I'm not familiar enough with them to properly implement it. Also, I ain't sure how much this will help performance/memory utilization, in comparison with the effort to refactor things (if needed?)

youknowone: I expect nothing changed about performance and memory. The compiler optimize shift ops very well.
It will remove manual shifting and bit ops

PR #7433, opened 2026-03-14
Opcode name

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Unified opcode naming so instruction names render consistently across the compiler and tooling.
  * Reworked formatting/display paths to use the unified opcode name source, replacing hard-coded opcode strings.
  * Standardized rendering for attribute/global/super and instrumented vs. base instruction views so opcode names appear consistently to users.


youknowone (commented):
Could you explain more about the motivation? The metadata seems to be described in a single function before. Now we have to edit 2 places to edit the instruction. There must be benefits to justify this inconvenience

ShaharNaveh:
> Could you explain more about the motivation? The metadata seems to be described in a single function before. Now we have to edit 2 places to edit the instruction. There must be benefits to justify this inconvenience

ofc, I was working on auto-generating the rust code for the opcodes (and methods). This change is extracted from those changes so the final PR wouldn't be that big. I guess it would have made more sense to implement `fmt::Display` instead. LMK what you think

youknowone:
2 questions.
- how about using naming conversion library? https://docs.rs/heck/latest/heck/
- Can the script auto-generate `fmt_dis` rather than only `name`? (if name is only used for fmt_dis)

ShaharNaveh:
> 2 questions.
> 
> * how about using naming conversion library? https://docs.rs/heck/latest/heck/

Wasn't aware of this crate. I can integrate it, it seems better tbh, although I ain't sure it's worth adding a new crate for this single use when there's a simpler (?) approach, with 0 runtime cost.

> * Can the script auto-generate `fmt_dis` rather than only `name`? (if name is only used for fmt_dis)

Yes, the script aims to auto generate an Opcode enum (which implements `fmt::Display` as the CPython name) and the Instruction enum (and the method to get the standalone opcode for said Instruction). I forgot to mention it at https://github.com/RustPython/RustPython/pull/7433#issuecomment-4062767910 :facepalm: which is IMO the most compelling point

youknowone:
you are right.  adding runtime library for const data is not ideal. 
If you are going to generate also fmt_dis, why don't you generate a single function like now? avoiding repeative same match patterns helps human cognitive load. Doesn't it fit in that way?

ShaharNaveh:
> you are right. adding runtime library for const data is not ideal. If you are going to generate also fmt_dis, why don't you generate a single function like now? avoiding repeative same match patterns helps human cognitive load. Doesn't it fit in that way?

I can do it, I think the reduced cognitive load comes when splitting into separate functions, and one of them called `name` that returns a `&str`, it took me some time to realize that the call for `stringify!` was invoked inside the macro. Anyway, feel free to close this PR if you think it's not a good idea (as always)

youknowone:
When we have to edit something, editing single place is better than editing multiple place. We are very fragile to track multi source for single truth.
Guess we have to edit an opcode. (not easy to guess because I am not confident human will do it ever though) If there are multiple match patterns, they must track every guard in multiple place. This is bad for both contributor and maintainer. Contributors can easily forget to check another function. Maintainer also can easily forget that we have another function to check.

So I think this change is justified only when we have more benefit than this disadvantage. 

> Anyway, feel free to close this PR if you think it's not a good idea (as always)

I always would like to listen the writer's opinion. The writer usually got more feedback from editing code. Let me have enough discussion about unclear situations. You actually have to prevent me not to close PRs without enough discussion if I am doing that only because I think it is not a good idea.

ShaharNaveh:
> When we have to edit something, editing single place is better than editing multiple place. We are very fragile to track multi source for single truth. Guess we have to edit an opcode. (not easy to guess because I am not confident human will do it ever though) If there are multiple match patterns, they must track every guard in multiple place. This is bad for both contributor and maintainer. Contributors can easily forget to check another function. Maintainer also can easily forget that we have another function to check.
> 
> So I think this change is justified only when we have more benefit than this disadvantage.

In the current state you're correct.

In the hypothetical scenario of having:

```rust
enum Opcode {
  A = 0,
  B = 1,
}

enum Instruction {
  A,
  B { oparg: u32 },
}

impl From<Instruction> for Opcode {
  ...
}
```

Would  implementing `fmt::Display` for Opcode make sense? And moreover, do we want having two "identical" enums (one with an oparg builder and one without)?

I believe this approach is suited for resolving #6746, as you could do:
```rust
matches!(instruction.opcode(), Opcode::A | Opcode::B)
```

> > Anyway, feel free to close this PR if you think it's not a good idea (as always)
> 
> I always would like to listen the writer's opinion. The writer usually got more feedback from editing code. Let me have enough discussion about unclear situations. You actually have to prevent me not to close PRs without enough discussion if I am doing that only because I think it is not a good idea.

I like your approach

youknowone:
I got the point. Could you include this commit when opcode is actually changed to that way? enum design must come first in that case.

ShaharNaveh:
> I got the point. Could you include this commit when opcode is actually changed to that way? enum design must come first in that case.

sounds better, I'll so that first

youknowone (approved):

PR #7437, opened 2026-03-15
Implement missing slots for NoneType

moreal:
This pull request does two things. First, as the title, it implements missing slots (e.g., `__eq__`) for `NoneType`.

And it lets slot traits (e.g., `Comparable`) implement only the slot (`wrapper_descriptor`), not the method. This task was ongoing, and this commit may complete it (https://github.com/RustPython/RustPython/pull/6486/).

It will remove `NoneType` from whats_left.

---

- NoneType tp_hash: https://github.com/python/cpython/blob/2f4e4ec2e7292901cab0c1466b78f5ddff48208d/Objects/object.c#L2276-L2279
```c
static Py_hash_t none_hash(PyObject *v)
{
    return 0xFCA86420;
}
```

## Summary by CodeRabbit

* **Improvements**
  * None now has a stable, consistent hash value (0xFCA86420 masked to 32 bits) for reliable use in sets and as dict keys.
  * Rich comparisons between None and non-None values now return NotImplemented as expected.

* **Behavior Changes**
  * Some explicit Python-level dunder wrappers (comparison, descriptor, repr, and attribute-call bridges) were removed; behavior now relies on core slot implementations.

* **Tests**
  * Added tests validating None's hashing, comparison behaviors, and presence/type of comparison/hash descriptors.

### DIFF
--- a/crates/vm/src/builtins/singletons.rs
+++ b/crates/vm/src/builtins/singletons.rs
@@ -91,7 +91,7 @@ impl Comparable for PyNone {
 
 impl Hashable for PyNone {
     fn hash(_zelf: &Py<Self>, _vm: &VirtualMachine) -> PyResult<PyHash> {
-        Ok(0xFCA8_6420)
+        Ok(0xFCA86420)
     }
 }
 


moreal:
As a memo, it seems https://github.com/RustPython/RustPython/pull/6562/changes/765133cd0fe99b7c91bd892b6e65620e268a9262 commit added pymethods to `Comparable` again but I can't guess the reason. It may be okay if all tests pass.

youknowone (approved):
Looks good, and thank you for catching legacy python magic methods.
There must be a bunch more of those unnecessary magic methods. Could you investigate more about other methods too?

crates/vm/src/builtins/singletons.rs
```diff
@@ -74,6 +75,26 @@ impl AsNumber for PyNone {
...
+        Ok(op
+            .identical_optimization(zelf, other)
+            .map(PyComparisonValue::Implemented)
+            .unwrap_or(PyComparisonValue::NotImplemented))
+    }
+}
+
+impl Hashable for PyNone {
+    fn hash(_zelf: &Py<Self>, _vm: &VirtualMachine) -> PyResult<PyHash> {
+        Ok(0xFCA8_6420)
```
youknowone:
```suggestion
        Ok(0xFCA86420)
```
Oh, just one glitch. If CPython use `0xFCA86420`, it's better to keep the same notation for search advantage.

youknowone (changes requested):

moreal:
> Looks good, and thank you for catching legacy python magic methods. There must be a bunch more of those unnecessary magic methods. Could you investigate more about other methods too?

Okay, I will do in the next pull request.

PR #7465, opened 2026-03-19
Resume opcode to hold ResumeType

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Improved type safety in bytecode instruction handling by refining argument type representation for resume operations.

### DIFF
--- a/crates/codegen/src/compile.rs
+++ b/crates/codegen/src/compile.rs
@@ -6881,9 +6881,9 @@ impl Compiler {
             self,
             Instruction::Resume {
                 context: if is_await {
-                    u32::from(bytecode::ResumeType::AfterAwait)
+                    bytecode::ResumeType::AfterAwait
                 } else {
-                    u32::from(bytecode::ResumeType::AfterYieldFrom)
+                    bytecode::ResumeType::AfterYieldFrom
                 }
             }
         );
@@ -7055,7 +7055,7 @@ impl Compiler {
                 emit!(
                     self,
                     Instruction::Resume {
-                        context: u32::from(bytecode::ResumeType::AfterYield)
+                        context: bytecode::ResumeType::AfterYield
                     }
                 );
             }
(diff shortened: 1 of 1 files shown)


ShaharNaveh:
@youknowone I'm thinking to create a derive macro called: `#[newtype_oparg]` that can be defined on either struct or enums. this will help streamline the oparg definition while being able to handle slightly less straight forward cases like overriding the `fmt::Display` for specific enum variants like, like we do here:

https://github.com/RustPython/RustPython/blob/b11b8e66ce75f4376f0bce97dbf749800c0f677e/crates/compiler-core/src/bytecode/oparg.rs#L265-L277

and

https://github.com/RustPython/RustPython/blob/b11b8e66ce75f4376f0bce97dbf749800c0f677e/crates/compiler-core/src/bytecode/oparg.rs#L575-L585

(we have more places)


Also with generate the `From<u32>` and `From<OPARG_ENUM> for u32` for enums with a fallback variant, like here with "Other(u32)". `Instruction::GetAwaitable` oparg can also use this pattern

---

lmk what you think:)

youknowone:
@ShaharNaveh sounds good to me. Could you  show me an example how it will look like once the macro is introduced?

ShaharNaveh:
> @ShaharNaveh sounds good to me. Could you show me an example how it will look like once the macro is introduced?

I was thinking something like:
```rust
#[newtype_oparg]
struct VarNum(u32);

#[newtype_oparg]
enum SpecialMethod {
  /// `__enter__` for sync context manager
  #[oparg(display = "__enter__")]
  Enter = 0,
  /// `__exit__` for sync context manager
  #[oparg(display = "__exit__")]
  Exit = 1,
  ...
}

#[newtype_oparg]
enum ResumeType {
  AfterYield = 1
  AfterYieldFrom = 2,
  #[oparg(catch_all)]
  Other(u32)
}
```

ShaharNaveh:
@youknowone loks good?

PR #7473, opened 2026-03-20
Pin deps to their minor version

ShaharNaveh:
When deps don't specify their minor version dependabot only update `Cargo.lock` (see #7470 for example).

This PR ensures that dependabot updates `Cargo.toml` as well.

Closes #7470

## Summary by CodeRabbit

* **Chores**
  * Updated project dependencies to latest stable versions for improved compatibility and performance.


moreal (dismissed):

youknowone:
RustPython is a binary and a library. Library user uses Cargo.toml and application uses uses Cargo.lock. Specifying the latest version is good for application users, but not for library users

ShaharNaveh:
> RustPython is a binary and a library. Library user uses Cargo.toml and application uses uses Cargo.lock. Specifying the latest version is good for application users, but not for library users

That's a good point.
 I believe consistency is good, which way should we choose?

moreal:
When I originally gave the approval, I hadn't considered the library users. Thinking back on it, it's actually normal behavior for Dependabot to only update Cargo.lock in that regard. For the sake of library users, we need to use the most generous (lowest) version possible, as long as it doesn't affect functionality.

So, honestly, I'm leaning toward sticking with the current approach. In the case of something like `bstr = "1"`, it originally allowed versions like `1.0.x` to be used, but now it requires `1.12.1` or higher.

youknowone:
@ShaharNaveh choosing tight versions for Cargo.lock and loose versions for Cargo.toml is a good trade-off and consistent. If we have to really upgrade incompatible, they are being upgraded together.
One of caveat of this approach is we accidently can specify too low version of dependnecy for Cargo.toml. Everything other is good.

PR #7480, opened 2026-03-22
Fix sub_table ordering for nested inlined comprehensions (PEP 709)

LegNeato:
When an inlined comprehension's first iterator expression contains nested scopes (such as a lambda), those scopes' sub_tables appear at the current position in the parent's sub_table list. The previous code spliced the comprehension's own child sub_tables (e.g. inner inlined comprehensions) into that same position before compiling the iterator, which shifted the iterator's sub_tables to wrong indices.

Move the splice after the first iterator is compiled so its sub_tables are consumed at their original positions.

Fixes nested list comprehensions like:
```python
    [[x for _, x in g] for _, g in itertools.groupby(..., lambda x: ...)]
```

Disclosure: I used AI to develop the patch though I was heavily involved.

## Bug Fixes
* Fixed nested comprehension scoping and iteration semantics for complex cases involving lambda expressions and underscore variables.

### DIFF
--- a/crates/codegen/src/compile.rs
+++ b/crates/codegen/src/compile.rs
@@ -8013,23 +8013,31 @@ impl Compiler {
     ) -> CompileResult<()> {
         // PEP 709: Consume the comprehension's sub_table.
         // The symbols are already merged into parent scope by analyze_symbol_table.
-        // Splice the comprehension's children into the parent so nested scopes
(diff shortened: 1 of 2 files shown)


youknowone (approved):
Thank you for contributing! You don't need to worry about AI develop as long as you are in the loop.

PR #7489, opened 2026-03-23
`#[newtype_oparg]` derive macro

ShaharNaveh:
Reason for creating a new crate for this is because `rustpython-derive-impl` has `rustpython-compiler` as a dependency, so it created a circular dependency:/

## Summary by CodeRabbit

* **New Features**
  * Added a procedural-macro crate to generate enum/newtype boilerplate.

* **Refactor**
  * Converted many opcode/newtype declarations to use the new macro-driven approach, standardizing displays and conversions and reducing manual boilerplate.

* **Bug Fixes**
  * Fixed parsing of containment-related bytecode arguments to preserve full argument width.

* **Chores**
  * Added workspace dependency to enable the new macro crate.

### DIFF
--- a/crates/macros/src/newtype_oparg.rs
+++ b/crates/macros/src/newtype_oparg.rs
@@ -97,6 +97,9 @@ impl TryFrom<syn::Variant> for VariantInfo {
     type Error = syn::Error;
 
     fn try_from(variant: syn::Variant) -> Result<Self, Self::Error> {
+        let ident = variant.ident.clone();
+        let discriminant = variant.discriminant.as_ref().map(|(_, expr)| expr.clone());
+
         let mut display = None;
         let mut catch_all = false;
         for attr in &variant.attrs {
@@ -118,8 +121,14 @@ impl TryFrom<syn::Variant> for VariantInfo {
             })?
         }
 
-        let ident = variant.ident.clone();
-        let discriminant = variant.discriminant.as_ref().map(|(_, expr)| expr.clone());
+        if catch_all
+            && !matches!(&variant.fields, syn::Fields::Unnamed(fields) if fields.unnamed.len() == 1)
+        {
+            return Err(Error::new(
+                ident.span(),
+                "`#[oparg(catch_all)]` variant must have exactly one unnamed field, e.g., `Other(u32)`",
+            ));
+        }
 
         if catch_all && display.is_some() {
             return Err(Error::new(
@@ -183,7 +192,7 @@ pub(super) fn handle_enum(item: ItemEnum) -> syn::Result<proc_macro2::TokenStrea
         // Don't assign value. Enables more optimizations by the compiler.
         variant.discriminant = None;
 
-        // Remove `#[oparg(...)`.
+        // Remove `#[oparg(...)]`.
         variant.attrs.retain(|attr| !attr.path().is_ident("oparg"));
 
         variant


crates/macros/src/newtype_oparg.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    let catch_all = variants_info.pop_if(|info| info.catch_all);
+
+    // Ensure a no multiple `#[oparg(catch_all)]`
+    if catch_all.is_some() && variants_info.iter().any(|vinfo| vinfo.catch_all) {
+        return Err(Error::new(
+            item.span(),
+            "Cannot define more than one `#[oparg(catch_all)]`",
+        ));
+    };
```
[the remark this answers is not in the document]
ShaharNaveh: We don't care because we raise an error if there are multiple `catch_all` variants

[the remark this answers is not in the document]
ShaharNaveh: oh, got it. thanks!

crates/compiler-core/src/bytecode/oparg.rs
```diff
@@ -113,273 +116,113 @@ impl OpArgState {
...
+pub enum ConvertValueOparg {
+    /// No conversion.
     ///
-    /// ## See also
+    /// ```python
+    /// f"{x}"
+    /// f"{x:4}"
+    /// ```
+    #[oparg(display = "")]
+    None = 0,
```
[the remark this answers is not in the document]
ShaharNaveh: @youknowone if this a concern? If so I can add something like:
```rust
enum Foo {
  #[oparg(alternatives = [255])]
  A = 0,
}
```

otherwise I don't think it's necessary. This used to rely on converting `-1 as u8` to get 255.
Which is no longer the case:
https://github.com/RustPython/RustPython/blob/82432be96239db7225e836961678f3af26bf9a62/crates/vm/src/stdlib/_ast/other.rs#L6-L17

youknowone:
`rustpython-derive-impl` has dependency of `rustpython-compiler-core` but not about `rustpython-compiler`. rustpython-compiler-core must be very small crate with essential types

ShaharNaveh:
> `rustpython-derive-impl` has dependency of `rustpython-compiler-core` but not about `rustpython-compiler`. rustpython-compiler-core must be very small crate with essential types

Yes, correct. What do you suggest to do in this case?

youknowone:
Sorry, I didn't get the situation yet. If we move this code into the previous macro crate, does it create circular dependency? (So, this new feature create circular dependency?)

ShaharNaveh:
> Sorry, I didn't get the situation yet. If we move this code into the previous macro crate, does it create circular dependency? (So, this new feature create circular dependency?)

Correct, that's why I created a new crate that doesn't depend on `rustpython-compiler-core `

I tried to add a feature called `compiler` to `rustpython-derive-impl` and to put `rustpython-compiler-core` as an optional dependency, but cargo wasn't happy with that still:/

crates/compiler-core/src/bytecode/oparg.rs
```diff
@@ -1,5 +1,8 @@
 use core::fmt;
 
+use crate as rustpython_compiler_core; // Required for newtype_oparg macro
+use rustpython_macros::newtype_oparg;
```
youknowone, line 4: okay, this is circular. unhappy to add a crate only for these a few types

youknowone:
let me think about this a bit.

crates/compiler-core/src/bytecode/oparg.rs
ShaharNaveh: btw, this is only needed because I'm implementing `OpArgType` for any struct/enum that uses the macro.

I can remove that and write:
```rust
// oparg.rs

impl OpArgType for A {}

impl OpArgType for B {}

impl OpArgType for C {}

...
```

youknowone: if avoidable with reasonable effort, not using proc macro is better

ShaharNaveh: I still think the proc macro adds value, but not a hill I'm willing to die on tbh

youknowone: lets go without proc macro. it is worth not to add too many crates when publishing

PR #7503, opened 2026-03-25
fix: Flush stdout on shutdown matching CPython behavior

### LINKED ISSUE #5521: Flush stdout on shutdown
This is done in cpython, and an error is returned if the flush does not succeed.

majiayu000:
Fixes #5521

## Summary

RustPython was not flushing stdout on interpreter shutdown, and when stdout flush failed (e.g. broken pipe), it silently ignored the error instead of returning exit code 120 like CPython does.

**Changes:**

- **`vm_object.rs`**: `flush_std()` now returns `i32` status (0 = ok, -1 = stdout flush failed). On stdout flush failure, calls `run_unraisable()` to report the error to stderr. Added `file_is_closed()` helper to skip flushing already-closed files. Stderr flush errors remain silently ignored (matching CPython).
- **`interpreter.rs`**: `finalize()` tracks flush status. If `exit_code` is 0 and stdout flush failed, returns 120 instead of 0. Non-zero exit codes (e.g. from `SystemExit`) are preserved.
- **`test_cmd_line.py`**: Removed `@unittest.expectedFailure` from `test_stdout_flush_at_shutdown`.

## Test Plan

- `cargo test --workspace --features threading` passes
- `cargo clippy` clean
- `test_stdout_flush_at_shutdown` now passes without `expectedFailure`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved flushing of stdout/stderr during shutdown with better detection of closed streams.
  * Exceptions during stdout flushing are now reported/handled instead of being silently ignored.
  * Shutdown now returns exit code 120 when an earlier flush failure is detected, aligning exit behavior with expectations.

### DIFF
--- a/crates/vm/src/vm/interpreter.rs
+++ b/crates/vm/src/vm/interpreter.rs
@@ -10,6 +10,10 @@ use core::sync::atomic::Ordering;
 
 type InitFunc = Box<dyn FnOnce(&mut VirtualMachine)>;
 
+/// Exit code used when stdout/stderr flush fails during interpreter shutdown.
+/// Matches CPython's behavior (see cpython/Python/pylifecycle.c).
+const EXITCODE_FLUSH_FAILURE: u32 = 120;
+
 /// Configuration builder for constructing an Interpreter.
 ///
 /// This is the preferred way to configure and create an interpreter with custom modules.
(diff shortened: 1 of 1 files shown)


ShaharNaveh (approved):
lgtm!
ty, and welcome to the project:)

PR #7568, opened 2026-04-06
Fix del obj.__dict__ to match CPython behavior (fixes #5355)

### LINKED ISSUE #5355: Inconsist with cpython in deleting __dict__

alok-108:
This is a proposed fix for issue #5355 where deleting `__dict__` on an object failed to match CPython's behavior.

### Goal
Match CPython behavior where `del obj.__dict__` successfully removes the dictionary, and subsequent access to `__dict__` re-creates it lazily as a fresh empty dictionary.

### Technical Changes
1. **Optional Instance Dictionary**: Updated `InstanceDict` in `crates/vm/src/object/core.rs` to hold an `Option<PyDictRef>`. This allows the dictionary field to be set to `None` upon deletion.
2. 2. **Deletion Support**: Modified `object_set_dict` in `crates/vm/src/builtins/object.rs` and relevant parts of `crates/vm/src/builtins/type.rs` to handle `PySetterValue::Delete`.
3. 3. **Lazy Creation**: Implemented logic in `object_get_dict` to re-initialize the dictionary if it was previously deleted, ensuring compatibility with CPython.
### Note on Verification
The logic follows established RustPython patterns (like `transpose()` and `try_into_value()`). However, I was unable to fully verify the build locally due to a missing Windows linker (`link.exe`) in my current environment. I've manually traced the logic and believe it to be correct.

@youknowone - Could you please review this proposed fix?

## Summary by CodeRabbit

* **Refactor**
  * Instance attribute dictionaries are now optional and created on demand; object dict storage and setter semantics were unified to explicitly represent assign vs delete.

* **Bug Fixes**
  * Attribute assignment/deletion and lookup now consistently respect absence of an instance dictionary and raise AttributeError when appropriate.

* **Tests**
  * Added a test confirming lazy __dict__ recreation and that removed attributes raise AttributeError.


youknowone (changes requested):
Please find a test under Lib/test/  and make sure it is working; otherwise put a test code under extra_tests/snippets to ensure this is fixing the reported issue. Also do not submit test files.

also, do NOT submit patch without proper human review. we expect PR authors are responsible for their patches.

alok-108:
Hi @youknowone, thanks for the feedback. I've updated the PR with the following:
- Added a snippet test in extra_tests/snippets/test_del_dict.py to verify the fix for del obj.__dict__.
- - Enabled two passing CPython tests: test_store_attr_deleted_dict and test_rematerialize_object_dict.
- - Fixed a potential GC crash in gc_clear_raw by using replace(None) to keep the InstanceDict container alive.
- - Updated generic_setattr to lazily re-create the dictionary after deletion if a new attribute is set.
About the 'proper human review' - sorry if I skipped a step. I'm still learning the workflow here. Should I have discussed this on the issue before opening the PR? Any guidance on the preferred process would be great. Thanks!

Lib/test/test_class.py
ShaharNaveh, line 1: Please remove the comments if there are no RustPython patches

ShaharNaveh (changes requested):
Thanks for the PR:)
please remove txt files from this PR, they seem like a debug artifact that should not be commited

alok-108:
@ShaharNaveh thanks for the review. I've removed the .txt files from the root and cleaned up the TODO comments in test_class.py. Only the expectedFailure removals remain now.

Lib/test/test_class.py
ShaharNaveh, line 1: Why the file was deleted?

ShaharNaveh (changes requested):

ShaharNaveh (approved):
Great! lgtm if CI passes

ShaharNaveh:
@alok-108 please fix the CI errors and push:)

alok-108:
@youknowone @ShaharNaveh 

All requested changes have been addressed:

- Added snippet test and enabled passing CPython tests
- Fixed GC safety, compilation errors, formatting, lint, and merge conflict
- All 21 CI checks are now green

Could you please re-review? Thanks!

extra_tests/snippets/test_del_dict.py
```diff
@@ -0,0 +1,46 @@
...
+import functools
+
+p = functools.partial(f)
+try:
+    del p.__dict__
+    assert False, "TypeError expected for partial dict deletion"
+except TypeError:
+    pass
+
+print("OK")
```
ShaharNaveh: I don't think this print is necessary

ShaharNaveh (approved):
overall lgtm!
ty:)

extra_tests/snippets/test_del_dict.py
alok-108: yup and i fix that too...

alok-108:
> overall lgtm! ty:)

now its ready to merge

alok-108:
@youknowone All your requested changes from 3 weeks ago (add tests, proper review) have been done. ShaharNaveh approved and all 21 CI checks are green. Could you please re-review and dismiss your old change request? Ready to merge. Thanks!

youknowone (approved):
I am sorry to forget about this patch. I attached a few clean up. Thank you for contributing!

PR #7580, opened 2026-04-10
Bytecode parity - for-loop cleanup, return duplication, late jump threading fix

youknowone:
- Use POP_TOP instead of POP_ITER for for-loop break/return cleanup
- Expand duplicate_end_returns to clone final return for jump predecessors
- Restrict late jump threading pass to unconditional jumps only
- Skip exception blocks in inline/reorder passes
- Simplify threaded_jump_instr NoInterrupt handling

## Summary by CodeRabbit

* **New Features**
  * AST objects gain a public __deepcopy__ for reliable deep copies.

* **Bug Fixes**
  * More robust traceback generation when location data is missing.
  * Fixes to generator/coroutine finalization and frame cleanup.
  * Corrected comparison behavior to honor boolean coercion where expected.
  * Safer OS write loop handling of interruptions.

* **Performance**
  * Improved bytecode control-flow optimizations and jump threading.

* **Tests**
  * Added threading-related tests exercising threaded imports and exec.

### DIFF
--- a/crates/codegen/src/compile.rs
+++ b/crates/codegen/src/compile.rs
@@ -1513,13 +1513,11 @@ impl Compiler {
 
             FBlockType::ForLoop => {
                 // When returning from a for-loop, CPython swaps the preserved
-                // value with the iterator and uses POP_TOP for the iterator slot.
+                // value with the iterator and uses POP_TOP for loop cleanup.
                 if preserve_tos {
                     emit!(self, Instruction::Swap { i: 2 });
-                    emit!(self, Instruction::PopTop);
-                } else {
-                    emit!(self, Instruction::PopIter);
                 }
+                emit!(self, Instruction::PopTop);
             }
 
             FBlockType::TryExcept => {
@@ -3442,7 +3440,11 @@ impl Compiler {
         let handler_block = self.new_block();
         let cleanup_block = self.new_block();
         let end_block = self.new_block();
-        let orelse_block = self.new_block();
+        let orelse_block = if orelse.is_empty() {
+            end_block
+        } else {
+            self.new_block()
+        };
 
         emit!(self, Instruction::Nop);
         emit!(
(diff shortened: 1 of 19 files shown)


ShaharNaveh:
@youknowone 
reviewdog has the following error:
```
time=2026-04-10T11:11:07.604Z level=INFO msg="reviewdog: failed to post a review comment: POST https://api.github.com/repos/RustPython/RustPython/pulls/7580/reviews: 403 Resource not accessible by integration []"
reviewdog: This GitHub Token doesn't have write permission of Review API [1],
so reviewdog will report results via logging command [2] and create annotations similar to
github-pr-check reporter as a fallback.
[1]: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target
[2]: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions
reviewdog: found at least one issue with severity greater than or equal to the given level: error
```

This has something to do with the repository settings, can you please take a look? then it will be able to make committable comments with auto-fixes. It works on forks, I guess there's some default setting that was changed

youknowone:
posting requires write permission. But the permission is read-only

crates/codegen/src/compile.rs
```diff
@@ -1513,12 +1513,12 @@ impl Compiler {
...
             FBlockType::ForLoop => {
                 // When returning from a for-loop, CPython swaps the preserved
-                // value with the iterator and uses POP_TOP for the iterator slot.
+                // value with the iterator and uses POP_TOP for loop cleanup.
                 if preserve_tos {
                     emit!(self, Instruction::Swap { i: 2 });
                     emit!(self, Instruction::PopTop);
                 } else {
-                    emit!(self, Instruction::PopIter);
+                    emit!(self, Instruction::PopTop);
```
youknowone: if both node of branch PopTop, why not out of branch?

crates/vm/src/stdlib/_thread.rs
```diff
@@ -53,8 +53,12 @@ pub(crate) mod _thread {
     const TIMEOUT_MAX_IN_MICROSECONDS: i64 = 0xffffffff * 1_000;
 
     // this is a value in seconds
-    #[pyattr]
-    const TIMEOUT_MAX: f64 = (TIMEOUT_MAX_IN_MICROSECONDS / 1_000_000) as f64;
+    const TIMEOUT_MAX_SECONDS: f64 = (TIMEOUT_MAX_IN_MICROSECONDS / 1_000_000) as f64;
+
+    #[pyattr(name = "TIMEOUT_MAX")]
```
youknowone: what's the point of this change? why not const ?

PR #7562, opened 2026-04-05
Add minimal C-api implementation that builds with Pyo3

bschoenmaeckers:
This is my shot at implementing a minimal Cpython compatible C-api. I've implemented the bare minimum to get the included Pyo3 example running where most of the api is stubbed. I'm not familiar with the rest of the RustPython code base so let me know what you think and where I did stupid things.

Please take extra care reviewing the pylifecycle.rs & pystate.rs files where I try to setup the RustPython interpreter.

xref #5604

PR #7595, opened 2026-04-13
Allow creating `PyCapsule` objects

bschoenmaeckers:
Allow creation of `PyCapsule` objects. The destructor is a `C` function for easy interop with other languages.

## Release Notes

* **New Features**
  * Capsule objects now support storing opaque pointers with optional destructor callbacks for proper resource cleanup
  * Added new APIs to create capsules and retrieve stored pointer data

### DIFF
--- a/crates/vm/src/builtins/capsule.rs
+++ b/crates/vm/src/builtins/capsule.rs
@@ -1,14 +1,19 @@
 use super::PyType;
-use crate::{Context, Py, PyPayload, PyResult, class::PyClassImpl, types::Representable};
+use crate::{
+    AsObject, Context, Py, PyObject, PyPayload, PyResult, VirtualMachine,
+    class::PyClassImpl,
+    types::{Destructor, Representable},
+};
+use core::ffi::c_void;
+use core::sync::atomic::AtomicPtr;
 
 /// PyCapsule - a container for C pointers.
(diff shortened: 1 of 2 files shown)


ShaharNaveh (commented):
welcome to the project:)
can you please share some information about this change? what is it trying to achieve?

crates/vm/src/builtins/capsule.rs
```diff
@@ -18,8 +23,26 @@ impl PyPayload for PyCapsule {
...
+            ptr: ptr.into(),
+            destructor,
+        }
+    }
+
+    pub fn pointer(&self) -> *mut c_void {
+        self.ptr.load(core::sync::atomic::Ordering::Relaxed)
+    }
+
+    fn destructor(&self) -> Option<unsafe extern "C" fn(_: *mut PyObject)> {
```
ShaharNaveh, line 42: does 
```suggestion
    const fn destructor(&self) -> Option<unsafe extern "C" fn(_: *mut PyObject)> {
```
compile?

bschoenmaeckers:
> welcome to the project:)
> can you please share some information about this change? what is it trying to achieve?

I needed this while experimenting with a capi at #7562. But this might be useful for other usecases as well.

youknowone (approved):
Looks good, Thank you!

PR #7603, opened 2026-04-14
Auto-retry flaky MP tests

ShaharNaveh:
This is pretty much what we were doing but manually until now:/

## Summary by CodeRabbit

* **Chores**
  * Improved CI test reliability by adding retry logic for flaky tests (up to five sequential attempts), grouping logs per attempt for clearer diagnostics, and stopping early when a run succeeds to shorten pipeline time.
  * Standardized the execution environment by explicitly setting the shell to bash to ensure consistent behavior and timeout handling across runs.


youknowone:
@fanninpm what do you think about? Would there be a better idea?

fanninpm:
Have you tried looking at libregrtest's `--rerun` flag?

ShaharNaveh:
> Have you tried looking at libregrtest's `--rerun` flag?

I wasn't aware of that flag, that's cool. but from what I can tell it only retry it once? Haven't seen an option to configure the retry count, I go to failed CIs quite a lot tbh, sometimes it takes 3 attempts for the tests to pass:/

.github/workflows/ci.yaml
```diff
@@ -286,8 +286,22 @@ jobs:
...
+            status=$?
+
+            echo "::endgroup:"
+
+            if [ $status -eq 0 ]; then
+              exit 0
+            fi
+          done
+
+          exit 1
```
fanninpm, line 317: If you want to loop through the `FLAKY_MP_TESTS` one by one, attempting multiple times, then here is one approach:

```suggestion
          super_status=0
          for test_module in ${{ env.FLAKY_MP_TESTS }}; do
            for attempt in $(seq 1 5); do
              echo "::group::${test_module} Attempt ${attempt}"

              set +e
              target/release/rustpython -m test -j 1 ${{ join(matrix.extra_test_args, ' ') }} --fail-env-changed --timeout 600 -v "${test_module}"
              status=$?
              set -e

              echo "::endgroup::"
              if [ $status -eq 0 ]; then
                break
              fi

            done
            echo "${test_module} exited with status ${status} on attempt ${attempt}."
            super_status=$((super_status + status))
          done

          exit $super_status
```

@coderabbitai any tips for improvement?

ShaharNaveh: > If you want to loop through the `FLAKY_MP_TESTS` one by one, attempting multiple times, then here is one approach:
> 
> @coderabbitai any tips for improvement?

What is the advantage of it this way over running it all at once multiple times?

ShaharNaveh:
> Have you tried looking at libregrtest's `--rerun` flag?

I'm leaning towards lowering the retry count to 3 (or 2) but with the `--rerun` flag, so it will actually have a possibly of retrying 6 times max but the total runtime would be lower if that makes sense

.github/workflows/ci.yaml
```diff
@@ -299,8 +299,22 @@ jobs:
 
       - name: Run flaky MP CPython tests
         run: |
-          target/release/rustpython -m test -j 1 ${{ join(matrix.extra_test_args, ' ') }} --slowest --fail-env-changed --timeout 600 -v ${{ env.FLAKY_MP_TESTS }}
+          for attempt in $(seq 1 5); do
+            echo "::group::Attempt ${attempt}"
+
+            target/release/rustpython -m test -j 1 ${{ join(matrix.extra_test_args, ' ') }} --slowest --fail-env-changed --timeout 600 -v ${{ env.FLAKY_MP_TESTS }}
+            status=$?
```
fanninpm, line 307:
```suggestion
            set +e
            target/release/rustpython -m test -j 1 ${{ join(matrix.extra_test_args, ' ') }} --slowest --fail-env-changed --timeout 600 -v ${{ env.FLAKY_MP_TESTS }}
            status=$?
            set -e
```

Otherwise when it fails, the whole step will fail.

.github/workflows/ci.yaml
```diff
@@ -299,8 +299,22 @@ jobs:
...
+
+            echo "::endgroup:"
```
fanninpm: Per @coderabbitai:
```suggestion
            echo "::endgroup::"
```

ShaharNaveh:
ty @fanninpm :)

youknowone:
@fanninpm please approve it once it is reasonable to you

fanninpm (approved):
This can always be changed later.

PR #7606, opened 2026-04-14
Update seq tests from 3.14.4

ShaharNaveh:


Lib/test/test_set.py
```diff
@@ -1969,7 +1969,6 @@ def check_set_op_does_not_crash(self, function):
                 self.assertIn("changed size during iteration", str(e))
 
 
-@unittest.skip("TODO: RUSTPYTHON; segfault")
 class TestBinaryOpsMutating(TestOperationsMutating):
```
fanninpm, line 1972: This class no longer segfaults, right?

Lib/test/test_set.py
```diff
@@ -2058,7 +2057,6 @@ class TestBinaryOpsMutating_Subclass_Set(TestBinaryOpsMutating, unittest.TestCas
     constructor2 = set
 
 
-@unittest.skip("TODO: RUSTPYTHON; segfault")
 class TestMethodsMutating(TestOperationsMutating):
```
fanninpm, line 2060: This class no longer segfaults, right?

ShaharNaveh: correct:)

Lib/test/test_set.py
ShaharNaveh: correct:)

fanninpm: Due to this, 60 tests will be reported as no longer being skipped. 🎉

Lib/test/test_set.py
fanninpm: Due to this, 66 tests will be reported as no longer being skipped. 🎉

youknowone (approved):

PR #7610, opened 2026-04-14
Unify CI caches (reduce total cache size by 75% ±)

ShaharNaveh:
ATM we have separate cache for a matrix of `operating systems * job number` (the actual number is less, but not by that much). And the cache of all those jobs is 99% the same.

This PR makes it so we have a single cache for an `${{ runner.os }}-${{ matrix.target }}-${{ matrix.toolchain }}`, which is 3 in total for now. each of them weight 820-900MB.

The issue with the large number of cache size is that once we reach our limit GH starts to purge old caches, so sometimes some `cargo check` jobs didn't have a cache hit:/

Caches on main
Screenshot_2026-04-14-17-24-19-02_e4424258c8b8649f6e67d283a50a2cbc

Caches after change (there will be more entries with more invocations):

Screenshot_2026-04-14-17-23-14-68_e4424258c8b8649f6e67d283a50a2cbc

more targets & toolchains can be easily added if needed

## Summary by CodeRabbit

* **Chores**
  * Standardized and expanded CI caching to improve cache hit rates and build speed (broader restore fallbacks and unified keying).
  * Disabled debug-info generation for build profiles to reduce artifact size.
  * Removed restrictive cache-restore guards to allow more consistent cache usage.
  * Added an automated workflow to refresh and maintain build caches across supported OSes.

### DIFF
--- a/.github/workflows/update-caches.yml
+++ b/.github/workflows/update-caches.yml
@@ -1,0 +1,73 @@
+name: Update Actions Caches
+
+permissions:
+  contents: read
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+env:
+  CARGO_INCREMENTAL: 0
+  CARGO_TERM_COLOR: always
+  CARGO_PROFILE_TEST_DEBUG: 0
+  CARGO_PROFILE_DEV_DEBUG: 0
+  CARGO_PROFILE_RELEASE_DEBUG: 0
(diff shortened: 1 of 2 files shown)


.github/workflows/update-caches.yml
ShaharNaveh, line 1: stole the entire thing from https://github.com/bevyengine/bevy/blob/4553e66916b882ba76be9423f401b79ebc1d7eec/.github/workflows/update-caches.yml

youknowone (approved):

PR #7612, opened 2026-04-15
fix: Python-Rust combining char diff in isalnum

### LINKED ISSUE #7518: Disagreement with CPython about which unicode characters are alnum
## Summary

The regex module interprets \w as matching more characters than it seems like it should (certainly more than CPython does).

## Expected

```
import re

assert not re.match(r"\w", "\u0345"), r"\w should not match U+0345 (category Mn)"
```

This assertion should pass, but instead it fails.

## Python Documentation

https://docs.python.org/3/library/re.html

This says that it should match anything for which isalnum() is true. This turns out to be also an area where RustPython and CPython disagree. `"\u0345".isalnum()` returns False on CPython and True on RustPython.

joshuamegnauth54:
Closes: #7518

Rust and Python differ on alphanumeric characters. Rust follows the Unicode standard closer than Python. This means that `is_alphanumeric` (char function in Rust) is different from `isalnum` (Python). To fix the discrepancy, RustPython needs to mimic Python by rejecting certain characters. Some classes of combining characters count as alphanumeric in Rust but not Python. Combining characters are accent marks that are combined with other characters to create a single grapheme.

It's possible that this PR is not exhaustive. I fixed the combining character issue BUT I don't know the full range of discrepancies.

---

~~This doesn't actually fix #7518, but it fixes a similar issue based on that report.~~ Actually, CodeRabbit pointed me in the right direction so now it _does_ fix #7518.

```rust
assert!('\u{006e}'.is_alphanumeric());
assert!(!'\u{0303}'.is_alphanumeric());
assert!('\u{00f1}'.is_alphanumeric());
assert!('\u{0345}'.is_alphanumeric());

for raw in 0x0363..=0x036f {
    let c = char::from_u32(raw).unwrap();
    assert!(c.is_alphanumeric());
}
```

```python
assert '\u006e'.isalnum()
assert not '\u0303'.isalnum()
assert '\u00f1'.isalnum()
assert not '\u0345'.isalnum()
assert not '\u0363'.isalnum()

for raw in range(0x0363, 0x036f):
    assert not chr(raw).isalnum()
```

^Note the differences which are accounted for by the new tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected str.isalnum() so standalone Unicode combining marks are not treated as alphanumeric, while base letters remain valid.

* **Tests**
  * Expanded tests for str.isalnum() covering combining-character cases.
  * Added regex test ensuring standalone combining marks are not matched as word characters.

### DIFF
--- a/crates/sre_engine/src/string.rs
+++ b/crates/sre_engine/src/string.rs
@@ -1,3 +1,4 @@
+use icu_properties::props::{CanonicalCombiningClass, EnumeratedProperty};
 use rustpython_wtf8::Wtf8;
 
 #[derive(Debug, Clone, Copy)]
@@ -443,7 +444,10 @@ pub(crate) const fn is_uni_linebreak(ch: u32) -> bool {
 pub(crate) fn is_uni_alnum(ch: u32) -> bool {
     // TODO: check with cpython
     char::try_from(ch)
-        .map(|x| x.is_alphanumeric())
+        .map(|x| {
+            x.is_alphanumeric()
+                && CanonicalCombiningClass::for_char(x) == CanonicalCombiningClass::NotReordered
+        })
         .unwrap_or(false)
 }
 #[inline]
--- a/crates/vm/src/builtins/str.rs
+++ b/crates/vm/src/builtins/str.rs
@@ -44,8 +44,12 @@ use rustpython_common::{
     wtf8::{CodePoint, Wtf8, Wtf8Buf, Wtf8Chunk, Wtf8Concat},
 };
 
-use icu_properties::props::{
-    BidiClass, BinaryProperty, EnumeratedProperty, GeneralCategory, XidContinue, XidStart,
+use icu_properties::{
+    CodePointMapData,
+    props::{
+        BidiClass, BinaryProperty, CanonicalCombiningClass, EnumeratedProperty, GeneralCategory,
+        XidContinue, XidStart,
+    },
 };
 use unicode_casing::CharExt;
 
@@ -946,7 +950,11 @@ impl PyStr {
 
     #[pymethod]
     fn isalnum(&self) -> bool {
-        !self.data.is_empty() && self.char_all(char::is_alphanumeric)
+        let map = CodePointMapData::<CanonicalCombiningClass>::new();
+        !self.data.is_empty()
+            && self.char_all(|c| {
+                c.is_alphanumeric() && map.get(c) == CanonicalCombiningClass::NotReordered
+            })
     }
 
     #[pymethod]
--- a/crates/vm/src/builtins/str.rs
+++ b/crates/vm/src/builtins/str.rs
@@ -44,12 +44,9 @@ use rustpython_common::{
     wtf8::{CodePoint, Wtf8, Wtf8Buf, Wtf8Chunk, Wtf8Concat},
 };
 
-use icu_properties::{
-    CodePointMapData,
-    props::{
-        BidiClass, BinaryProperty, CanonicalCombiningClass, EnumeratedProperty, GeneralCategory,
-        XidContinue, XidStart,
-    },
+use icu_properties::props::{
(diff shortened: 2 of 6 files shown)


crates/sre_engine/src/string.rs
```diff
@@ -442,8 +443,9 @@ pub(crate) const fn is_uni_linebreak(ch: u32) -> bool {
 #[inline]
 pub(crate) fn is_uni_alnum(ch: u32) -> bool {
     // TODO: check with cpython
+    let map = CodePointMapData::<CanonicalCombiningClass>::new();
     char::try_from(ch)
-        .map(|x| x.is_alphanumeric())
+        .map(|x| x.is_alphanumeric() && map.get(x) == CanonicalCombiningClass::NotReordered)
```
ShaharNaveh: Remove line 446 and:
```suggestion
        .map(|x| x.is_alphanumeric() && CanonicalCombiningClass::for_char(x) == CanonicalCombiningClass::NotReordered)
```

crates/vm/src/builtins/str.rs
```diff
@@ -946,7 +950,11 @@ impl PyStr {
 
     #[pymethod]
     fn isalnum(&self) -> bool {
-        !self.data.is_empty() && self.char_all(char::is_alphanumeric)
+        let map = CodePointMapData::<CanonicalCombiningClass>::new();
+        !self.data.is_empty()
+            && self.char_all(|c| {
+                c.is_alphanumeric() && map.get(c) == CanonicalCombiningClass::NotReordered
```
ShaharNaveh: remove line 953 and
```suggestion
                c.is_alphanumeric() && CanonicalCombiningClass::for_char(c) == CanonicalCombiningClass::NotReordered
```

extra_tests/snippets/builtin_str.py
```diff
@@ -69,10 +69,17 @@
...
-# assert "Ǳ".title() == "ǲ"
 assert a.isalpha()
 
+# Combining characters differ slightly between Rust and Python
+assert "\u006e".isalnum()
+assert not "\u0303".isalnum()
+assert not "\u006e\u0303".isalnum()
+assert "\u00f1".isalnum()
+assert not "\u0345".isalnum()
+for raw in range(0x0363, 0x036f):
```
ShaharNaveh:
```suggestion
for raw in range(0x0363, 0x036F):
```

joshuamegnauth54:
I fixed the issues and force pushed. :grin:

ShaharNaveh (approved):
Looks great!
tysm:)

joshuamegnauth54:
CodeRabbit's comment is interesting. :thinking: I assumed there were more edge cases. I didn't look at `isalpha()` and the rest yet, but there are issues open for them too. I'll try to fix those in a future patch in a more robust manner.

extra_tests/snippets/builtin_str.py
```diff
@@ -73,6 +73,15 @@
 # assert "Ǳ".title() == "ǲ"
 assert a.isalpha()
 
+# Combining characters differ slightly between Rust and Python
+assert "\u006e".isalnum()
+assert not "\u0303".isalnum()
+assert not "\u006e\u0303".isalnum()
+assert "\u00f1".isalnum()
```
ShaharNaveh, line 80: it may be worth opening a PR for CPython to test those changes to ensure there's no regression on their end as well (it will get to us as well)

youknowone (approved):
> Rust and Python differ on alphanumeric characters. Rust follows the Unicode standard closer than Python. 

If it does, isn't it better to patch CPython to follow unicode standard better?

joshuamegnauth54:
Sorry, I'm not sure if I'm supposed to respond here to you two after PRs have been merged. :sweat_smile: 

I don't fully understand the issue, but I think that the differences are due to Unicode version and how each project chooses to define "alphanumeric." I'm trying to figure out if I can just combine Unicode properties to mimic cpython.

PR #7613, opened 2026-04-15
Simplify cache_entries match statement

ShaharNaveh:
## Summary by CodeRabbit

* **Refactor**
  * Simplified the compiler's instruction cache estimation, reducing internal complexity and improving maintainability.
  * Adjusted cache-count assignments for several instruction types; this may slightly alter cache accounting and optimization behavior in generated bytecode.
  * No public APIs or external interfaces were changed.


crates/compiler-core/src/bytecode/instruction.rs
```diff
@@ -697,124 +697,32 @@ impl Instruction {
     }
 
     /// Number of CACHE code units that follow this instruction.
+    ///
+    /// Instrumented and specelized opcodes have the same cache entries as their base.
```
ShaharNaveh:
```suggestion
    /// Instrumented and specialized opcodes have the same cache entries as their base.
```

youknowone (approved):

PR #7624, opened 2026-04-18
Update `test_base64.py` from 3.14.4

changjoon-park:
Syncs `Lib/test/test_base64.py` with CPython 3.14.4.

Adds coverage for the `ignorechars` parameter of `base64.a85decode`:
- Positive case with `ignorechars=b" \n"`
- `ValueError` when `ignorechars` is empty or missing chars present in the input
- `TypeError` when `ignorechars` is not bytes (`str`, `None`)

Also reorders imports so `unittest` comes first (matches CPython's ordering).

No RUSTPYTHON markers touched.

### DIFF
--- a/Lib/test/test_base64.py
+++ b/Lib/test/test_base64.py
@@ -1,7 +1,7 @@
+import unittest
 import base64
 import binascii
 import os
-import unittest
 from array import array
 from test.support import cpython_only
 from test.support import os_helper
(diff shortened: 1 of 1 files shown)


ShaharNaveh (approved):
ty:)

changjoon-park:
> ty:)

Thank you !

Issue #7527, opened 2026-03-28
Disagreement with CPython for str.istitle/title/isdigit/isidentifier/isnumeric


sunmy2019:
## Python Documentation or reference to CPython source code

tested on current main: 1a9b10ece50ebdbdd3820dafa8f6a630d87be9c0

There are many examples and I picked one.

```py
# the following is True in CPython and False in RustPython
print(
    "\u1c89".istitle() == False,
    "Ǳ".title() == "ǲ",
    "\u0661".isdigit() == True,
    "\u0c5d".isidentifier() == True,
    "\u3405".isnumeric() == True,
    # known issues of isalpha and isalnum omitted
)
```

PR #7626, opened 2026-04-18
Use Unicode properties for alnum, alpha, etc.

joshuamegnauth54:
Rust and Python differ in which properties they use for alphanumeric, numeric, et cetera. Both languages list which properties are used which makes it easy to mimic Python's behavior in Rust.

My previous patch was a bit shortsighted because I filtered out combining characters from is_alphanumeric. Using properties is exact and also much cleaner. It also covers edge cases that my initial approach missed.

Besides isalnum, I also fixed isnumeric and isdigit in the same way by using properties.

---

This contributes a bit to #7527. Unicode issues for isalnum etc. should be solved now (and permanently too!!). Title casing looks harder to fix, but the basic solution of using properties more should apply there too. Also, the RegEx engine can benefit from using properties more too, but I'm terrible at RegEx so I didn't really try to fix issues there. :laughing: 

For extra reading, here is [Python's docs](https://docs.python.org/3/library/stdtypes.html#str.isalnum) which lists the categories used. Look at `isalpha()`, for example.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Unicode classification for string methods (isalnum, isalpha, isdigit, isnumeric) for more accurate behavior across international characters and special Unicode categories.

* **Tests**
  * Added/expanded tests covering edge cases: superscript digits, fractions, Roman numerals, and additional non-Latin characters to validate classification behavior.

### DIFF
--- a/Lib/test/test_str.py
+++ b/Lib/test/test_str.py
@@ -792,7 +792,6 @@ class StrTest(string_tests.StringLikeTest,
         for ch in ['\U0001D7F6', '\U00011066', '\U000104A0']:
             self.assertTrue(ch.isdecimal(), '{!a} is decimal.'.format(ch))
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON; AssertionError: False != True
     def test_isdigit(self):
         super().test_isdigit()
         self.checkequalnofix(True, '\u2460', 'isdigit')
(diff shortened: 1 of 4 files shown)


ShaharNaveh:
> UNEXPECTED SUCCESS: test_isdigit (test.test_str.StrTest.test_isdigit)

:partying_face:

joshuamegnauth54:
I force pushed to handle one of CodeRabbit's suggestions and enable the passing test. CodeRabbit's first suggestion (the one that says to use `matches!`) is incorrect because `NumericType` is not an enum, but the second suggestion cleans up the code and still works! :grin: 

I was very confused why that test kept "failing". :laughing: It's awesome that it passes now.

youknowone (approved):
Thank you! 👍

PR #7633, opened 2026-04-19
Fix process abort on large float format precision

changjoon-park:
## Summary

Formatting a float with a large precision aborts the interpreter instead of raising a Python exception. CPython returns a clean string on the same input.

```python
# Before
$ ./rustpython -c "print(f'{1.5:.1000000}')"
thread 'main' panicked at crates/literal/src/float.rs:135:
Formatting argument out of range              # → exit 101 (process abort)

# After
$ ./rustpython -c "print(f'{1.5:.1000000}')"
1.5
```

## Root cause

Rust's `format!("{:.*}", n, x)` macro panics when `n` exceeds the fmt runtime's internal precision limit. `format_fixed` in `crates/literal/src/float.rs` already caps `n` at `u16::MAX` before calling `format!`, but its sibling functions `format_general` and `format_exponent` — and the `FormatType::Percentage` branch in `crates/common/src/format.rs` — pass user-supplied precision straight through. A one-line user script (`f"{x:.N}"` with N ≳ 65535) triggers the abort.

Affected format types:

| Type | Before | After |
|---|---|---|
| `f` (fixed) | OK (already capped) | OK |
| `e` / `E` (exponential) | **abort** | OK |
| `g` / `G` (general) | **abort** | OK |
| `%` (percent) | **abort** | OK |
| default (no type) | **abort** (routes to `g`) | OK |

## Fix

1. Add `FMT_MAX_PRECISION` + `clamp_fmt_precision()` helper at module level in `float.rs`.
2. Cap is `u16::MAX - 1`, not `u16::MAX` — `{:.*e}` hits a second assertion (`ndigits > 0` in `core::num::flt2dec`) at exactly `u16::MAX`; the smaller value covers both `{:.*}` and `{:.*e}` uniformly.
3. Apply the helper to:
   - `format_fixed` (replacing existing ad-hoc cap — consistency only)
   - `format_exponent` (new, at function entry)
   - `format_general` (new, at each of three internal `format!` calls, with saturating arithmetic on derived precision values)
   - `FormatType::Percentage` branch in `common/src/format.rs` (new)

Complex-number formatting and old-style `%`-formatting dispatch to the same library functions, so they transitively benefit without separate changes.

## Why the cap is safe

f64 carries only ~17 significant decimal digits. Precision beyond ~17 produces padding zeros (for `f`/`e`/`%`) or is silently trimmed (for `g`). Capping at ~65K is far beyond any user-meaningful precision and matches the existing `format_fixed` behavior already shipping in `main`.

## Verification

```console
$ cargo +1.94.0 build --release
    Finished `release` profile [optimized] target(s) in 40.54s

$ ./target/release/rustpython -m test test_float test_fstring test_format
All 3 tests OK.
Total tests: run=162

$ ./target/release/rustpython extra_tests/snippets/builtin_format.py
# (all assertions pass, including 7 new regression cases)
```

### Reliability audit (beyond the trigger path)

Probed after the fix:

- **40 magnitude/type combinations**: 10 values × 4 format types at precision 200_000 — `0.0`, `±1.5`, `±inf`, `nan`, `1e-300`, `1e300`, `f64::MAX`, `5e-324`. All return clean strings.
- **Boundary precisions**: 0 / 1 / 2 for each format type — outputs match expected (`'{:.0g}'.format(1.5) == '2'`, etc.).
- **Complex numbers**: five format specs including `.200000e` — all OK (transitively via `format_exponent`).
- **Old-style `%` formatting**: `'%.200000e' % 1.5` etc. — all OK (transitively via `cformat.rs` → `format_fixed`/`format_exponent`/`format_general`).
- **Combined format specs**: fill + align + width + precision, sign + precision, alternate form + precision, zero-pad + precision, grouping + precision — all OK.
- **Defense-in-depth range**: Rust's parser rejects precision > `i32::MAX` with `ValueError("Precision too big")`, so the guarded interval `[0, i32::MAX]` is now panic-free end-to-end.

## Related

- Prior fix using the same "cap before Rust `format!`" approach: `format_fixed` already did this; this PR extends the same pattern to its siblings.
- Adjacent hardening PRs using the same spirit (translate native failures into Python exceptions): #7630 (cyclic-AST `SIGSEGV` → `RecursionError`), #7632 (deep-JSON `SIGSEGV` → `RecursionError`).

## Summary by CodeRabbit

* **Bug Fixes**
  * Float formatting now handles extreme precisions safely: percent formatting outputs "inf%" for infinite values, precision is clamped for fixed/exponential/general formats, and fractional digits are zero-padded when needed to match expected behavior.

* **Tests**
  * Added regression tests covering very large precision values across f/e/g/% formats to ensure stable, non-crashing output and preserved NaN/Inf representations.

### DIFF
--- a/extra_tests/snippets/builtin_format.py
+++ b/extra_tests/snippets/builtin_format.py
@@ -199,3 +199,31 @@ x = 123456789012345678901234567890
 for i in range(0, 30):
     format(x, ",")
     x = x // 10
+
+
+# Large float precision must not abort the interpreter.
+# Previously these paths hit unguarded `format!("{:.*e}", ...)` in
+# crates/literal/src/float.rs and `crates/common/src/format.rs` (the `%`
+# branch), which panic past Rust's fmt precision limit and killed the
+# process instead of raising a Python exception.
+_big = 1_000_000
+# f-string default (general format) — g-format trims trailing zeros, so
+# high precision returns the short natural representation.
+assert f"{1.5:.{_big}}" == "1.5"
+assert "{:.{}g}".format(1.5, _big) == "1.5"
+assert "{:.{}G}".format(1.5, _big) == "1.5"
+# Exponential and percent types emit padded zeros up to the (internally
+# capped) precision. We don't pin exact length; we only require the call
+# to return a str and not crash the runtime.
+for spec_type in ("e", "E", "%", "f"):
+    out = ("{:." + str(_big) + spec_type + "}").format(1.5)
+    assert isinstance(out, str) and len(out) > 0
+
+# Shallow cases unchanged.
+assert f"{1.5:.5}" == "1.5"
+assert "{:.3f}".format(1.5) == "1.500"
+assert "{:.2%}".format(0.25) == "25.00%"
+assert "{:.4e}".format(1234.5) == "1.2345e+03"
+assert "{:.3g}".format(1234.5) == "1.23e+03"
+assert f"{float('nan'):.10f}" == "nan"
+assert f"{float('inf'):.10f}" == "inf"
--- a/extra_tests/snippets/builtin_format.py
+++ b/extra_tests/snippets/builtin_format.py
@@ -209,24 +209,35 @@ for i in range(0, 30):
 # u16::MAX; output is zero-padded past that boundary to match CPython
 # byte-identically.
 
-# Boundary values around the internal cap (u16::MAX = 65535). Output must
-# match what CPython would produce.
+# Three precision points per format type — below the cap (uncapped
+# path), exactly at the cap (boundary), and one past the cap (the
+# unhappy case, where internal clamping plus zero-padding has to
+# reconstruct CPython's output). All must byte-match CPython.
+
 # f-format pads with trailing zeros up to the requested precision.
-assert "{:.65534f}".format(1.5) == "1." + "5" + "0" * 65533
-assert "{:.65535f}".format(1.5) == "1." + "5" + "0" * 65534
-assert "{:.65536f}".format(1.5) == "1." + "5" + "0" * 65535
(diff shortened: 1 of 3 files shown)


extra_tests/snippets/builtin_format.py
```diff
@@ -199,3 +199,31 @@ def test_zero_padding():
...
     format(x, ",")
     x = x // 10
+
+
+# Large float precision must not abort the interpreter.
+# Previously these paths hit unguarded `format!("{:.*e}", ...)` in
+# crates/literal/src/float.rs and `crates/common/src/format.rs` (the `%`
+# branch), which panic past Rust's fmt precision limit and killed the
+# process instead of raising a Python exception.
+_big = 1_000_000
```
youknowone: Because this patch set the boundary as `pub const FMT_MAX_PRECISION: usize = u16::MAX as usize;`, the test must cover 65535 and 65536 to ensure its behavior follows CPython

youknowone (changes requested):

extra_tests/snippets/builtin_format.py
```diff
@@ -199,3 +199,44 @@ def test_zero_padding():
...
+assert "{:.100000%}".format(1.7976931348623157e308) == "inf%"
+
+# Shallow cases unchanged.
+assert f"{1.5:.5}" == "1.5"
+assert "{:.3f}".format(1.5) == "1.500"
+assert "{:.2%}".format(0.25) == "25.00%"
+assert "{:.4e}".format(1234.5) == "1.2345e+03"
+assert "{:.3g}".format(1234.5) == "1.23e+03"
+assert f"{float('nan'):.10f}" == "nan"
+assert f"{float('inf'):.10f}" == "inf"
```
youknowone, line 253: which test is about unhappy cases, like exceeding the MAX_PRECISION?

youknowone (approved):

PR #7647, opened 2026-04-21
Draft: explore a RustPython-owned ABI facade for pristine PyO3

smarcd:
## Summary

This PR explores a RustPython-owned ABI facade in `crates/capi` so that pristine or near-pristine PyO3 can work against RustPython without requiring a large RustPython-specific backend split inside PyO3.

This direction is explicitly informed by:
- RustPython/RustPython#7562
- PyO3/pyo3#5986 and the maintainer feedback there that the burden should likely sit more on RustPython than on PyO3

## Current verified status

- pristine `pyo3 0.28.x` compiles against `crates/capi`
- unchanged downstream packages:
  - `rpds`: package-owned tests green
  - `jiter`: package-owned tests green
  - `blake3`: package-owned tests green except the `numpy`-dependent case

## Scope and claims

This is not claiming:
- ready abi3 support
- all PyO3 packages now work
- broad drop-in CPython compatibility

What it does claim is:
- the ABI-facade direction is viable
- the compatibility burden can sit mostly in RustPython
- real unchanged PyO3 packages can already be driven through this facade

## Reviewer asks

1. Is the current core/facade boundary acceptable?
2. Are the current core hooks minimal enough?
3. Should `crates/capi` be the preferred direction relative to `#7562`?


bschoenmaeckers:
Why are you sidestepping my MR. Without participating in the api discussion there?

smarcd:
> Why are you sidestepping my MR. Without participating in the api discussion there?

This is a PR taking your ideas and running with them.

crates/capi/src/symbols.rs
```diff
@@ -0,0 +1,335 @@
...
+#[unsafe(export_name = "PyLong_Type")]
+pub static mut PYLONG_TYPE_HANDLE: *mut PyTypeObject = ptr::null_mut();
+#[unsafe(export_name = "PyModule_Type")]
+pub static mut PYMODULE_TYPE_HANDLE: *mut PyTypeObject = ptr::null_mut();
+#[unsafe(export_name = "PyTuple_Type")]
+pub static mut PYTUPLE_TYPE_HANDLE: *mut PyTypeObject = ptr::null_mut();
+#[unsafe(export_name = "PyType_Type")]
+pub static mut PYTYPE_TYPE_HANDLE: *mut PyTypeObject = ptr::null_mut();
+#[unsafe(export_name = "PyUnicode_Type")]
+pub static mut PYUNICODE_TYPE_HANDLE: *mut PyTypeObject = ptr::null_mut();
```
bschoenmaeckers: As discussed in my PR, this will not work.

smarcd: Removed, replaced with https://github.com/RustPython/RustPython/pull/7647/changes/f67a9342710e6add616a65066d8ec572d1a74045#diff-6cb426843cb08f1943a128f6df3e3e46e3862a5eae416e33d45a079529eab6b2

crates/capi/src/abstract_.rs
```diff
@@ -0,0 +1,406 @@
...
+    arg4: *mut PyObject,
+    arg5: *mut PyObject,
+    arg6: *mut PyObject,
+    arg7: *mut PyObject,
+    arg8: *mut PyObject,
+) -> *mut PyObject {
+    let raw = [arg1, arg2, arg3, arg4, arg5, arg6, arg7, arg8];
+    let nargs = raw.iter().position(|arg| arg.is_null()).unwrap_or(raw.len());
+    RustPython_PyObject_CallMethodObjArgsArray(receiver, name, raw.as_ptr(), nargs)
+}
```
smarcd: **Duplicate exported symbol `PyObject_CallMethodObjArgs` — link error / ABI corruption.**

This Rust definition collides with the variadic C definition in [`crates/capi/csrc/pyobject_callmethodobjargs.c#L13-L20`](https://github.com/RustPython/RustPython/blob/f67a9342710e6add616a65066d8ec572d1a74045/crates/capi/csrc/pyobject_callmethodobjargs.c#L13-L20). Both have external linkage with the same name (Rust uses `#[unsafe(no_mangle)]`, C is non-`static`), both end up in the same `rustpython-capi` rlib (the C file is compiled by `cc::Build` in [`crates/capi/build.rs`](https://github.com/RustPython/RustPython/blob/f67a9342710e6add616a65066d8ec572d1a74045/crates/capi/build.rs#L1-L7)), and `lib.rs` even adds `#[used] static KEEP_PYOBJECT_CALL_METHOD_OBJ_ARGS` to force this Rust version to be retained.

Most linkers (gnu-ld with `--whole-archive`, lld in cdylib mode, etc.) will fail with a multiple-definition error. On macOS the linker can silently pick one — and the two signatures are not interchangeable: the C entry is variadic with a NULL sentinel, this Rust entry takes 8 fixed `*mut PyObject` slots. If the Rust version wins, real CPython callers that pass more than 8 args (or that pass the NULL terminator) get garbage in the trailing slots; if the C version wins, this Rust function is dead code.

Since the C wrapper already implements the variadic ABI and just dispatches to `RustPython_PyObject_CallMethodObjArgsArray`, the simplest fix is to delete this Rust function (and the matching `#[used]` static in `lib.rs`).

crates/capi/src/abstract_.rs
```diff
@@ -0,0 +1,406 @@
...
+) -> *mut PyObject {
+    with_vm(|vm| {
+        let args_len = nargsf & !PY_VECTORCALL_ARGUMENTS_OFFSET;
+        let num_positional_args = args_len;
+
+        let args = unsafe { slice::from_raw_parts(args, args_len) }
+            .iter()
+            .map(|arg| unsafe { &*resolve_object_handle(*arg) }.to_owned())
+            .collect::<Vec<_>>();
+
```
smarcd: **Vectorcall drops keyword-argument values — wrong result / runtime panic when any kwargs are passed.**

This slice reads only `args_len` (positional count) entries from the C-supplied `args` buffer. CPython's vectorcall ABI requires that when `kwnames` is non-NULL the buffer holds `nargs + len(kwnames)` entries — positional values at `args[0..nargs]` and keyword values at `args[nargs..nargs+nkwargs]`. See [Python docs / vectorcall](https://docs.python.org/3/c-api/call.html#vectorcall):

> If kwnames is not NULL, it must be a tuple of strings naming the keyword arguments. The values of these arguments are stored in args after the positional arguments.

Downstream, `FuncArgs::from_vectorcall_owned` does `args.drain(nargs..nargs + kw_count)` (see [`crates/vm/src/function/argument.rs#L177-L186`](https://github.com/RustPython/RustPython/blob/f67a9342710e6add616a65066d8ec572d1a74045/crates/vm/src/function/argument.rs#L177-L186)), which will index past the end of this truncated `Vec` whenever `kwnames` is non-empty (debug-asserts in debug, OOB drain panic in release). Any vectorcall through this facade with keyword arguments is broken.

`PyObject_VectorcallMethod` at [L167-L172](https://github.com/RustPython/RustPython/blob/f67a9342710e6add616a65066d8ec572d1a74045/crates/capi/src/abstract_.rs#L167-L172) builds its own slice the same way, but since it then re-enters this function through `args.as_ptr()` (still pointing into the original C buffer where the kwarg values live), fixing only `PyObject_Vectorcall` is enough: resolve `kwnames` length first and slice `args_len + kwnames_len` from the raw pointer.

smarcd (commented):
Two high-confidence issues from a focused review of the new `crates/capi` facade (the rest looks consistent with the draft/exploration intent).

smarcd:
Updated verification from this branch:

- `rpds`: package-owned tests green
- `jiter`: package-owned tests green
- `blake3`: package-owned tests green except the `numpy`-dependent case

I also tightened one concrete ABI regression while doing this pass: the binary must retain the C variadic `PyObject_CallMethodObjArgs` shim export, otherwise `rpds` fails at import-time `dlopen` with the symbol missing.

The remaining work is now mostly ABI surface completeness (`PyType_FromSpec`, slots, buffers, unicode, lifecycle), not basic PyO3 package viability.

bschoenmaeckers:
@youknowone this can be closed right?

PR #7648, opened 2026-04-21
Add minimal capi lifecycle support

bschoenmaeckers:
This add the most basic c api I can think off. This way we can iterate over how we want to handle the interpreter initialization  api's. After this, the road is clear to implement most of the ABI3 api's one-by-one.

This also include my basic pyo3 test harness that exposes the c api's in a safe and battle tested way. To run the tests use `cargo test -p rustpython-capi`.

### References:
- https://docs.python.org/3/c-api/interp-lifecycle.html
- https://docs.python.org/3/c-api/threads.html

xref #7562 

## Summary by CodeRabbit

* **New Features**
  * Adds a C-API compatibility layer to build RustPython as a Python extension, exposing initialization, GIL/thread-state handling, and refcount operations.

* **Chores**
  * Adds build/configuration to support extension-module builds and a new optional "capi" feature to enable the extension.

* **Behavioral Changes**
  * Fork/after-fork reinitialization is now gated by an additional "host_env" feature; when extension support is enabled, runtime execution may use a separate VM thread.

### DIFF
--- a/crates/capi/pyo3-rustpython.config
+++ b/crates/capi/pyo3-rustpython.config
@@ -1,0 +1,5 @@
+implementation=CPython
+version=3.14
+shared=true
+abi3=true
+suppress_build_script_link_lines=true
--- a/crates/capi/src/lib.rs
+++ b/crates/capi/src/lib.rs
@@ -1,8 +1,19 @@
 #![allow(clippy::missing_safety_doc)]
+
+use crate::pylifecycle::MAIN_INTERP;
+use rustpython_vm::Interpreter;
 pub use rustpython_vm::PyObject;
+use std::sync::MutexGuard;
 
 extern crate alloc;
 
 pub mod pylifecycle;
 pub mod pystate;
 pub mod refcount;
+
+/// Get main interpreter of this process. Will be None if it has not been initialized yet.
+pub fn get_main_interpreter() -> MutexGuard<'static, Option<Interpreter>> {
+    MAIN_INTERP
+        .lock()
+        .expect("Failed to lock interpreter mutex")
+}
--- a/crates/capi/src/lib.rs
+++ b/crates/capi/src/lib.rs
@@ -1,3 +1,4 @@
+#![allow(clippy::missing_safety_doc)]
 pub use rustpython_vm::PyObject;
 
 extern crate alloc;
--- a/crates/capi/src/lib.rs
+++ b/crates/capi/src/lib.rs
@@ -1,0 +1,7 @@
+pub use rustpython_vm::PyObject;
+
+extern crate alloc;
+
+pub mod pylifecycle;
+pub mod pystate;
+pub mod refcount;
--- a/crates/capi/src/pylifecycle.rs
+++ b/crates/capi/src/pylifecycle.rs
@@ -1,5 +1,5 @@
 use crate::get_main_interpreter;
-use crate::pystate::attach_vm_to_thread;
+use crate::pystate::ensure_thread_has_vm_attached;
 use core::ffi::c_int;
 use rustpython_vm::Interpreter;
 use rustpython_vm::vm::thread::ThreadedVirtualMachine;
@@ -31,7 +31,7 @@ pub extern "C" fn Py_InitializeEx(_initsigs: c_int) {
     if interp.is_none() {
         *interp = Interpreter::with_init(Default::default(), |_vm| {}).into();
         drop(interp);
-        attach_vm_to_thread();
+        ensure_thread_has_vm_attached();
     }
 }
 
--- a/crates/capi/src/pylifecycle.rs
+++ b/crates/capi/src/pylifecycle.rs
@@ -1,24 +1,23 @@
+use crate::get_main_interpreter;
 use crate::pystate::attach_vm_to_thread;
 use core::ffi::c_int;
 use rustpython_vm::Interpreter;
 use rustpython_vm::vm::thread::ThreadedVirtualMachine;
-use std::sync::{Mutex, OnceLock};
+use std::sync::Mutex;
 
-static MAIN_INTERP: OnceLock<Mutex<Interpreter>> = OnceLock::new();
+pub(crate) static MAIN_INTERP: Mutex<Option<Interpreter>> = Mutex::new(None);
 
 /// Request a thread local vm from the main interpreter
 pub(crate) fn request_vm_from_interpreter() -> ThreadedVirtualMachine {
-    MAIN_INTERP
-        .get()
-        .expect("Interpreter is not initialized")
-        .lock()
-        .expect("Failed to lock interpreter mutex")
+    get_main_interpreter()
+        .as_ref()
+        .expect("Interpreter not initialized")
         .enter(|vm| vm.new_thread())
 }
 
 #[unsafe(no_mangle)]
 pub extern "C" fn Py_IsInitialized() -> c_int {
-    MAIN_INTERP.get().is_some() as c_int
+    get_main_interpreter().is_some() as c_int
 }
 
 #[unsafe(no_mangle)]
@@ -28,9 +27,12 @@ pub extern "C" fn Py_Initialize() {
 
 #[unsafe(no_mangle)]
 pub extern "C" fn Py_InitializeEx(_initsigs: c_int) {
-    MAIN_INTERP.get_or_init(|| Interpreter::with_init(Default::default(), |_vm| {}).into());
-
-    attach_vm_to_thread();
+    let mut interp = get_main_interpreter();
+    if interp.is_none() {
+        *interp = Interpreter::with_init(Default::default(), |_vm| {}).into();
+        drop(interp);
+        attach_vm_to_thread();
+    }
 }
 
 #[unsafe(no_mangle)]
--- a/crates/capi/src/pylifecycle.rs
+++ b/crates/capi/src/pylifecycle.rs
@@ -1,0 +1,49 @@
+use crate::pystate::attach_vm_to_thread;
+use core::ffi::c_int;
+use rustpython_vm::Interpreter;
+use rustpython_vm::vm::thread::ThreadedVirtualMachine;
+use std::sync::{Mutex, OnceLock};
+
+static MAIN_INTERP: OnceLock<Mutex<Interpreter>> = OnceLock::new();
+
+/// Request a thread local vm from the main interpreter
+pub(crate) fn request_vm_from_interpreter() -> ThreadedVirtualMachine {
+    MAIN_INTERP
+        .get()
+        .expect("Interpreter is not initialized")
+        .lock()
+        .expect("Failed to lock interpreter mutex")
+        .enter(|vm| vm.new_thread())
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_IsInitialized() -> c_int {
+    MAIN_INTERP.get().is_some() as c_int
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_Initialize() {
+    Py_InitializeEx(0);
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_InitializeEx(_initsigs: c_int) {
+    MAIN_INTERP.get_or_init(|| Interpreter::with_init(Default::default(), |_vm| {}).into());
+
+    attach_vm_to_thread();
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_Finalize() {
+    let _ = Py_FinalizeEx();
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_FinalizeEx() -> c_int {
+    0
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_IsFinalizing() -> c_int {
+    0
+}
--- a/crates/capi/src/pystate.rs
+++ b/crates/capi/src/pystate.rs
@@ -9,6 +9,7 @@ thread_local! {
     static VM: RefCell<Option<ThreadedVirtualMachine>> = const { RefCell::new(None) };
 }
 
+#[allow(dead_code)]
 pub(crate) fn with_vm<R>(f: impl FnOnce(&VirtualMachine) -> R) -> R {
     if VM_CURRENT.is_set() {
         // We have an active VM set, so use that.
@@ -62,6 +63,7 @@ pub extern "C" fn PyEval_SaveThread() -> *mut PyThreadState {
 
 #[cfg(test)]
 mod tests {
+    use crate::get_main_interpreter;
     use crate::pystate::{VM, with_vm};
     use pyo3::prelude::*;
     use rustpython_vm::vm::thread::VM_CURRENT;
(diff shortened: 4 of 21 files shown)


bschoenmaeckers:
Please feel free to push to this branch.

crates/capi/pyo3-rustpython.config
```diff
@@ -0,0 +1,5 @@
+implementation=CPython
+version=3.14
+shared=true
+abi3=true
```
youknowone, line 4: is there any freethreading marker? we only target free-treading, which represents `abi3t`

youknowone (commented):
good idea to split to small patches 👍

crates/capi/pyo3-rustpython.config
bschoenmaeckers: `abi3t` support is not yet implemented in PyO3. But for now I make sure not to use incompatible api's. For progress see https://github.com/PyO3/pyo3/issues/5786

crates/capi/src/pystate.rs
```diff
@@ -0,0 +1,66 @@
...
+    0
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn PyGILState_Release(_state: PyGILState_STATE) {}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn PyEval_SaveThread() -> *mut PyThreadState {
+    ptr::null_mut()
+}
```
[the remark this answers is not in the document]
bschoenmaeckers: `Py_BEGIN_ALLOW_THREADS` is a no-op by design here.

crates/capi/src/refcount.rs
```diff
@@ -0,0 +1,17 @@
...
+    // By dropping PyObjectRef, we will decrement the reference count.
+    unsafe { PyObjectRef::from_raw(NonNull::new_unchecked(op)) };
+}
+
+#[unsafe(no_mangle)]
+#[allow(clippy::not_unsafe_ptr_arg_deref)]
+pub extern "C" fn _Py_IncRef(op: *mut PyObject) {
+    // Don't drop the owned value, as we just want to increment the refcount.
+    core::mem::forget(unsafe { (*op).to_owned() });
+}
```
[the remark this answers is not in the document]
bschoenmaeckers: Marked as unsafe.

crates/capi/src/pystate.rs
```diff
@@ -0,0 +1,66 @@
...
+
+#[unsafe(no_mangle)]
+pub extern "C" fn PyGILState_Ensure() -> PyGILState_STATE {
+    attach_vm_to_thread();
+
+    0
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn PyGILState_Release(_state: PyGILState_STATE) {}
```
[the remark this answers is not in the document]
bschoenmaeckers: We do not need GIL state tracking because we do not have a GIL.

bschoenmaeckers:
@youknowone Can you help me fix clippy? The `rustpython-capi` crate needs `threading` support to be enabled on the `rustpython-stdlib`. But cargo-shear does not want unused dependencies.

bschoenmaeckers:
> @youknowone Can you help me fix clippy? The `rustpython-capi` crate needs `threading` support to be enabled on the `rustpython-stdlib`. But cargo-shear does not want unused dependencies.

Got it to work.

crates/vm/src/stdlib/_imp.rs
```diff
@@ -75,22 +76,22 @@ mod lock {
...
-#[cfg(all(unix, feature = "threading"))]
+#[cfg(all(unix, feature = "threading", feature = "host_env"))]
 pub(crate) unsafe fn reinit_imp_lock_after_fork() {
     unsafe { lock::reinit_after_fork() }
 }
 
-#[cfg(all(unix, feature = "threading"))]
+#[cfg(all(unix, feature = "threading", feature = "host_env"))]
 pub(crate) unsafe fn after_fork_child_imp_lock_release() {
     unsafe { lock::after_fork_child_reinit_and_release() }
```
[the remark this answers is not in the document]
bschoenmaeckers: posix.rs already requires unix so that clause is redundant right?

crates/capi/src/pylifecycle.rs
```diff
@@ -0,0 +1,71 @@
...
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_Finalize() {
+    let _ = Py_FinalizeEx();
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_FinalizeEx() -> c_int {
+    0
```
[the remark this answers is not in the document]
bschoenmaeckers: Fixed

crates/capi/src/pylifecycle.rs
```diff
@@ -0,0 +1,50 @@
...
+    get_main_interpreter()
+        .as_ref()
+        .expect("Interpreter not initialized")
+        .enter(|vm| vm.new_thread())
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_IsInitialized() -> c_int {
+    get_main_interpreter().is_none() as c_int
+}
```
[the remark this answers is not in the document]
bschoenmaeckers: Fixed

crates/capi/src/pylifecycle.rs
```diff
@@ -0,0 +1,50 @@
...
+    Py_InitializeEx(0);
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn Py_InitializeEx(_initsigs: c_int) {
+    let mut interp = get_main_interpreter();
+    if interp.is_none() {
+        *interp = Interpreter::with_init(Default::default(), |_vm| {}).into();
+        attach_vm_to_thread();
+    }
```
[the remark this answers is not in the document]
bschoenmaeckers: Fixed

crates/capi/src/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+pub use rustpython_vm::PyObject;
+use std::sync::MutexGuard;
+
+extern crate alloc;
+
+pub mod pylifecycle;
+pub mod pystate;
+pub mod refcount;
+
+/// Get main interpreter of this process. Will be None of it has not been initialized yet.
```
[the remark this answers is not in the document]
bschoenmaeckers: Fixed

bschoenmaeckers:
@youknowone I think I have the right setup now for the interpreter. Please let me know what you think.

crates/capi/src/pystate.rs
```diff
@@ -0,0 +1,66 @@
+use crate::pylifecycle::request_vm_from_interpreter;
+use core::cell::RefCell;
+use core::ffi::c_int;
+use core::ptr;
+use rustpython_vm::vm::thread::ThreadedVirtualMachine;
+
+thread_local! {
+    static VM: RefCell<Option<ThreadedVirtualMachine>> = const { RefCell::new(None) };
```
youknowone: my last concern is here. This value is duplication of RustPython's internal thread vm. 
what if a thread is created RustPython inside? then will this be correctly pointing that? Otherwise it doesn't matter because user cannot directly access to the thread via C API?

src/lib.rs
```diff
@@ -118,7 +118,15 @@ pub fn run(mut builder: InterpreterBuilder) -> ExitCode {
     builder = builder.settings(settings);
 
     let interp = builder.interpreter();
-    let exitcode = interp.run(move |vm| run_rustpython(vm, run_mode));
+    let exitcode = cfg_select! {
+        feature = "capi" => {{
```
youknowone, line 122: I have a question to help determine whether the Python C API needs to be supported in the Python executable itself.

Is this C API intended to be used both for embedding Python as a library and for building Python C extensions?
For the former, I assume we do not need to enable the C API in the executable binary itself. (We can publish a library rather than executable)
For the latter, however, it seems unavoidable.

bschoenmaeckers: The C API is intended for both embedding python and for usage in extensions libraries (python packages). 

Libraries will be linked to the exposed symbols in the executable when imported by python, so not at compile time. This way a library does not have any python code embedded into it. You can think of this as the client side of the api.

When embedding the python runtime, you essentially take the implementation side (or server if we continue this though experiment) of the api and embed it into the application. This is what the python executable should be doing because it is running python obviously.  This way a library can also be loaded in dynamically , and be linked the these symbols.

I hope this clears up your understanding.

crates/capi/src/pystate.rs
bschoenmaeckers: In this situation it will create 2 vm's for the same thread. I'm not sure if that is a problem but I can imagine it will be. So maybe we have to find some other way?

bschoenmaeckers: I see the problem now. I might have a solution to this, I will think this over. If you have any insights please don't hesitate to share!

bschoenmaeckers: My last commit addresses this issue. It will now use `VM_CURRENT` when one is already active. But creates a new vm when a module spawned it own new thread.

See 4c8263615318a046f554064656a716f4ac4a8ad1

bschoenmaeckers: I've added a test case as well. To make sure we do not create a new VM when there is already one active.

See 650f588bc53ea757294b9b2e4b7fb71da79f0aab

bschoenmaeckers: @youknowone Could you please take another look? 🙏

crates/capi/src/pystate.rs
```diff
@@ -0,0 +1,124 @@
...
+#[repr(C)]
+pub struct PyThreadState {
+    _interp: *mut core::ffi::c_void,
+}
+
+/// Make sure this thread has a running vm attached. This only creates a new vm if we don't already
+/// have one. So this will only create a new vm when we are in a new thread created outside RustPython.
+pub(crate) fn ensure_thread_has_vm_attached() {
+    if !VM_CURRENT.is_set() {
+        VM.with(|vm| {
```
youknowone: I am still worrying this duplicated VM_CURRENT and VM structure. Even though this is logically true at the moment, this can be fragile in the future.

How about creating a VM to VM_CURRENT, and never create static VM here? We might need to make a few features public form thread side. But that will be more sound for future stability.

How do you think? This is the last blocker for in the patch for me.

crates/capi/src/pystate.rs
youknowone: I am sorry for delayed review. I left a comment. I wish we can entirely remvoe the duplicatied VM store. And that looks possible to me. Please share if there is any blocker I overlooked

crates/capi/src/pystate.rs
bschoenmaeckers: I do not see a way to set `VM_CURRENT` with a new value. Because `VM_CURRENT.set` is a scoped value it cannot be set for a duration longer than the provided closure. And to my knowledge there is no API available to do this.

I see 2 solutions to this:
 - Use the current implementation as is. And maybe in the future change it when required
 - Don't define `VM_CURRENT` as a `scoped_tls` and implement to scoping ourselves.

bschoenmaeckers: See https://docs.rs/scoped-tls/latest/scoped_tls/struct.ScopedKey.html#method.set

youknowone: scoped_tls is a tool. That is used to help us, not to limit the design, at least ideally.
let me check this. thank you for investigating!

bschoenmaeckers: Thanks! Looking forward to your solution.

ShaharNaveh (approved):
I think that the current state is good for now, this PR already brings lots of value and if we see this being fragile we can think of a solution, atm it seems like we're trying to find a solution for a problem we haven't encountered yet (talking about https://github.com/RustPython/RustPython/pull/7648#discussion_r3160036608)

youknowone:
> it seems like we're trying to find a solution for a problem we haven't encountered yet 

disagree. storing same data to multiple storage must be justified by good reason (e.g. cache). this is not acceptable at least without proper proof they don't make problem, and I will try to avoid this design even if there is a proof of that.

bschoenmaeckers:
> disagree. storing same data to multiple storage must be justified by good reason (e.g. cache). this is not acceptable at least without proper proof they don't make problem, and I will try to avoid this design even if there is a proof of that.

For what it's worth; I agree that it is not ideal but I think the current solution will not create problems. This currently blocks other work so I propose to merge this as is and refactor `VM_CURRENT` in a followup PR?

youknowone:
I am trying to figure out how to align it correctly. give me a few more days

youknowone:
@bschoenmaeckers @ShaharNaveh  This is my trial. could you review it please?

crates/vm/src/vm/thread.rs
```diff
@@ -50,6 +50,15 @@ pub type CurrentFrameSlot = Arc<ThreadSlot>;
...
     pub(super) static VM_STACK: RefCell<Vec<NonNull<VirtualMachine>>> = Vec::with_capacity(1).into();
 
+    /// Thread state created through the GILState-style C API.
+    ///
+    /// This is separate from the current VM stack: it only means "attached now",
+    /// while this owns the per-thread VM that may be detached and re-attached.
+    /// Despite the historical CPython "GILState" name, this does not model a
+    /// GIL; it stores the VM used by that compatibility API.
+    #[cfg(feature = "threading")]
+    static GILSTATE_VM: RefCell<Option<Box<ThreadedVirtualMachine>>> = const { RefCell::new(None) };
```
bschoenmaeckers, line 65: What is the reason `ThreadedVirtualMachine` is wrapped in a `Box` here?

bschoenmaeckers:
> @bschoenmaeckers @ShaharNaveh This is my trial. could you review it please?

Overall looks good 👍

ShaharNaveh (approved):
LGTM!
Just need to satisfy cargo-sheer

youknowone (approved):
This is generated review. Let's clear them in follow-up patches.

  - Blocker: crates/capi/src/pylifecycle.rs:44 is a no-op. It is fine to leave
    APIs unimplemented, but exporting a CPython lifecycle symbol while leaving
    the interpreter initialized after Py_FinalizeEx() is incorrect behavior, not
    just an incomplete feature.
  - Blocker: crates/capi/src/pystate.rs:40 always returns null. That is unsafe
    for real C API users because callers expect a restorable thread state.
    Either implement SaveThread/RestoreThread together or do not expose this
    symbol yet.
  - Major: crates/vm/src/vm/thread.rs:175 does not clean up the C API external
    thread’s frame/thread slot. The normal _thread path calls crates/vm/src/
    stdlib/_thread.rs:578, but the C API release path only drops the owner and
    pops VM_STACK. This can leave stale state visible to _current_frames and
    stop-the-world tracking.
  - Major: crates/vm/src/vm/thread.rs:180 drops GILSTATE_VM before popping
    crates/vm/src/vm/thread.rs:183. The intent is understandable, but it exposes
    a VM that is already being dropped as the current VM. Cleanup should happen
    while the VM is still valid, then the stack entry and owner should be
    removed in a well-defined order.
  - Major: The PR needs to explicitly frame this as an opaque ABI3-style subset,
    not general CPython ABI compatibility. RustPython’s crates/vm/src/object/
    core.rs:1228 is not CPython’s object layout, so extension compatibility must
    be limited to code that treats PyObject* opaquely through exported APIs.

bschoenmaeckers:
Thanks!

Issue #7611, opened 2026-04-14
Inventory of Skipped Tests with `TODO: RUSTPYTHON` Annotations (as of 16 April 2026)


fanninpm:
Some of these actually get misreported on the website, which may indicate a bug either in `extra_tests/custom_text_test_runner.py` or in the website code, particularly when class names are identical between files. So far, I only got the ones on Linux, as they're the most visible on the website.

- `Lib/test/_test_eintr.py`
  - class `SelectEINTRTest(EINTRBaseTest)`
    - [ ] `test_poll`: timed out at the 10 minute mark
- `Lib/test/test_asyncio/test_free_threading.py` 
  - class `TestEagerPyFreeThreading(TestPyFreeThreading)` and class `TestPyFreeThreading`
    - [ ] `test_all_tasks_race`: hangs - Python `_current_tasks` dict not thread-safe
- `Lib/test/test_pydoc/test_pydoc.py`
  - class `PydocDocTest`
    - [ ] `test_builtin_with_more_than_four_children`: Panic
  - class `PydocUrlHandlerTest`
    - [ ] `test_content_type_err`: Panic
- `Lib/test/test_array.py`
  - class `LargeArrayTest`
    - [ ] `test_setitem_use_after_clear_with_float_data`: Hangs
    - [ ] `test_setitem_use_after_clear_with_int_data`: Hangs
    - [ ] `test_setitem_use_after_shrink_with_int_data`: Hangs
- `Lib/test/test_ast/test_ast.py`
  - class `ASTHelpers_Test`
    - [ ] `test_recursion_direct`: crash
    - [ ] `test_recursion_indirect`: crash
  - class `ASTValidatorTests`
    - [ ] `test_boolop`: called `Option::unwrap()` on a `None` value
    - [ ] `test_compare`: thread 'main' panicked
- `Lib/test/test_atexit.py`
  - class `FunctionalTest`
    - [ ] `test_atexit_thread_safety`: Flakey on CI
- `Lib/test/test_builtin.py`
  - class `BuiltinTest`
    - [ ] `test_bytearray_join_with_misbehaving_iterator`: Hangs
    - [ ] `test_filter_dealloc`: Segfault
  - class `PtyTests`
    - [ ] `test_input_no_stdout_fileno`: FAILURE, WORKER BUG
  - class `TestType`
    - [ ] `test_namespace_order`: random failure
- `Lib/test/test_bytes.py`
[...]

PR #7656, opened 2026-04-22
Raise ValueError for None in Expr lists to match CPython validator

changjoon-park:
## Summary

CPython raises `ValueError(\"None disallowed in expression list\")` when `compile()` encounters `None` as an element in an `Expr` list position (`Set/List/Tuple.elts`, `BoolOp.values`, `Call.args`, comprehension generators, etc.). RustPython reached the catch-all `TypeError(\"expected some sort of expr, but got {}\")` branch instead, failing a dozen `test_ast` validator tests.

## Fix

One check in `Expr::ast_from_object` (`crates/vm/src/stdlib/_ast/expression.rs`) before the catch-all:

```rust
} else if vm.is_none(&object) {
    return Err(vm.new_value_error("None disallowed in expression list".to_owned()));
}
```

`Option<Expr>` positions (e.g. `Dict.keys` for `**unpack`) are unaffected — `node.rs` `Option<T>::ast_from_object` already short-circuits `None` there, so this change only fires when `None` reaches the non-optional `Vec<Expr>` iteration path.

## Verification

CPython 3.13.4 baseline message:

```
$ python3 -c "import ast; compile(ast.Expression(ast.Set([None])), '<t>', 'eval')"
ValueError: None disallowed in expression list
```

After the fix, `rustpython -c "import ast; compile(ast.Expression(ast.Set([None])), '<t>', 'eval')"` raises the same message.

Full `test.test_ast.test_ast` suite (227 tests):

```
Ran 227 tests in 18.144s
OK (skipped=10, expected failures=39)
```

0 failures, 0 unexpected successes.

## Unmasked tests (12)

In `Lib/test/test_ast/test_ast.py`:

- `test_boolop` — was `@unittest.skip` with stale ``Option::unwrap()`` message
- `test_set`, `test_list`, `test_tuple` — were `@unittest.expectedFailure` / TypeError
- `test_call`, `test_delete`, `test_assign`, `test_dict`, `test_dictcomp`, `test_generatorexp`, `test_listcomp`, `test_setcomp` — were `@unittest.expectedFailure` / \"ValueError not raised\"

Addresses part of the skip-test inventory in #7611.

## Scope notes

The Stmt-level analog — `body=[None]` raising `\"None disallowed in statement list\"`, used by `test_classdef` — needs a sibling change in `statement.rs`. Out of scope here; can be a follow-up PR with the identical pattern.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when a missing/empty (None) value is encountered in an expression list. Users will now see a clearer, specific error indicating None is disallowed in expression lists instead of a generic type mismatch, making invalid-expression diagnostics more actionable.

### DIFF
--- a/crates/vm/src/stdlib/_ast/expression.rs
+++ b/crates/vm/src/stdlib/_ast/expression.rs
@@ -133,6 +133,8 @@ impl Node for ast::Expr {
             let interpolation =
                 string::TStringInterpolation::ast_from_object(vm, source_file, object)?;
             return string::interpolation_to_expr(vm, interpolation);
+        } else if vm.is_none(&object) {
+            return Err(vm.new_value_error("None disallowed in expression list".to_owned()));
         } else {
             return Err(vm.new_type_error(format!(
                 "expected some sort of expr, but got {}",
--- a/crates/vm/src/stdlib/_ast/expression.rs
+++ b/crates/vm/src/stdlib/_ast/expression.rs
@@ -134,7 +134,7 @@ impl Node for ast::Expr {
                 string::TStringInterpolation::ast_from_object(vm, source_file, object)?;
             return string::interpolation_to_expr(vm, interpolation);
         } else if vm.is_none(&object) {
-            return Err(vm.new_value_error("None disallowed in expression list".to_owned()));
+            return Err(vm.new_value_error("None disallowed in expression list"));
         } else {
             return Err(vm.new_type_error(format!(
                 "expected some sort of expr, but got {}",
--- a/Lib/test/test_ast/test_ast.py
+++ b/Lib/test/test_ast/test_ast.py
@@ -2132,14 +2132,12 @@ class ASTValidatorTests(unittest.TestCase):
         self.stmt(cls(decorator_list=[ast.Name("x", ast.Store())]),
                   "must have Load context")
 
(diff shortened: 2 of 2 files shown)


crates/vm/src/stdlib/_ast/expression.rs
```diff
@@ -133,6 +133,8 @@ impl Node for ast::Expr {
             let interpolation =
                 string::TStringInterpolation::ast_from_object(vm, source_file, object)?;
             return string::interpolation_to_expr(vm, interpolation);
+        } else if vm.is_none(&object) {
+            return Err(vm.new_value_error("None disallowed in expression list".to_owned()));
```
ShaharNaveh:
```suggestion
            return Err(vm.new_value_error("None disallowed in expression list"));
```

ShaharNaveh (approved):
Amazing! tysm!

crates/vm/src/stdlib/_ast/expression.rs
changjoon-park: Applied in 95c6566 with you as co-author
i had to commit locally since the suggestion went outdated after a force-push for some reason
thank for catching !

changjoon-park:
> Amazing! tysm!

thanks for the review !

ShaharNaveh (approved):

PR #7657, opened 2026-04-23
Support custom `rustls` providers

### LINKED ISSUE #7059: Support custom crypto backends via rustls-no-provider
I'd like to integrate and test rustpython in another project utilizing https://github.com/ctz/graviola and rustls instead of aws-lc-rs. However currently rustpython hard codes aws-lc-rs in the Cargo.toml. Is this something that could be supported?

reqwest has a rustls-no-provider option where you are able to use graviola like this: 
```rust
fn main() {
  rustls_graviola::default_provider().install_default().unwrap();
  // ... rest of main function
}
```

However I see that there are a *few* mentions of aws_lc_rs imports directly in the rustpython stdlib  🤔 

See https://github.com/RustPython/RustPython/blob/dc2d235e59766d1e671c37e0be6ba7adbf3f93ea/crates/stdlib/Cargo.toml#L125

joshuamegnauth54:
`rustls`'s architecture is very clean and trait-driven. There are many providers for `rustls` including the built-in `aws-lc-rs` and `ring` as well as backends for `boringssl`, `graviola`, `openssl`, `mbedtls`, etc.

The new feature, `ssl-rustls-no-provider`, enables custom rustls providers. By default, `aws-lc-rs` is enabled which matches the old behavior and keeps backward compatibility.

I wrote a new type that abstracts what we need from crypto providers. CryptoExt encapsulates the ticketer as well as cipher suites and KX groups. I wrote fallbacks to help select a reasonable default if a provider is missing features (they all seem to support the same things though).

I also wrote an example to show how to actually use custom providers.

Closes: #7059

---

`aws-lc-rs` is still the default, so this patch should be transparent to all users.

I tested with:

```sh
cargo build
target/debug/rustpython Lib/test/test_ssl.py
```

And:

(todo, test with example!)

## Summary by CodeRabbit

* **New Features**
  * Pluggable SSL/TLS crypto providers so users can choose custom cryptographic backends.
  * New feature option to run Rustls without a bundled provider.
  * Added an example showing how to select and run with different TLS providers.

* **Chores**
  * Updated feature defaults to favor the aws-lc-backed Rustls option.
  * Adjusted build/test workflows and feature wiring to support the new provider options.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7657)

### DIFF
--- a/crates/stdlib/src/ssl.rs
+++ b/crates/stdlib/src/ssl.rs
@@ -85,14 +85,8 @@ mod _ssl {
     };
     use sha2::{Digest, Sha256};
 
-    #[cfg(all(feature = "ssl-rustls-aws-lc-rs", feature = "ssl-rustls-ring"))]
-    compile_error!("Enable only one rustls provider: ssl-rustls-aws-lc-rs or ssl-rustls-ring");
-
-    #[cfg(feature = "ssl-rustls-aws-lc-rs")]
+    #[cfg(feature = "ssl-rustls")]
     use rustls::crypto::aws_lc_rs::{ALL_CIPHER_SUITES, Ticketer, sign};
-
-    #[cfg(feature = "ssl-rustls-ring")]
-    use rustls::crypto::ring::{ALL_CIPHER_SUITES, Ticketer, sign};
 
     // Import certificate operations module
     use super::cert;
--- a/crates/stdlib/src/ssl.rs
+++ b/crates/stdlib/src/ssl.rs
@@ -25,6 +25,9 @@ mod compat;
 // SSL exception types (shared with openssl backend)
 mod error;
 
+// Utilities for setting a Rustls cryptography provider.
+pub mod providers;
+
 pub(crate) use _ssl::module_def;
 
 #[allow(non_snake_case)]
@@ -85,9 +88,6 @@ mod _ssl {
     };
     use sha2::{Digest, Sha256};
 
-    #[cfg(feature = "ssl-rustls")]
-    use rustls::crypto::aws_lc_rs::{ALL_CIPHER_SUITES, Ticketer, sign};
-
     // Import certificate operations module
     use super::cert;
 
@@ -100,6 +100,8 @@ mod _ssl {
         create_client_config, create_server_config, curve_name_to_kx_group, extract_cipher_info,
         get_cipher_encryption_desc, is_blocking_io_error, normalize_cipher_name, ssl_do_handshake,
     };
+
+    use super::providers::CryptoExt;
 
     // Type aliases for better readability
     // Additional type alias for certificate/key pairs (SessionCache and SniCertName defined below)
@@ -638,7 +640,7 @@ mod _ssl {
             return Err("No cipher can be selected".to_string());
         }
 
-        let all_suites = ALL_CIPHER_SUITES;
+        let all_suites = CryptoExt::get_ext().all_ciphers_or_default();
         let mut selected = Vec::new();
 
         for part in cipher_str.split(':') {
@@ -1061,6 +1063,8 @@ mod _ssl {
 
         #[pymethod]
         fn load_cert_chain(&self, args: LoadCertChainArgs, vm: &VirtualMachine) -> PyResult<()> {
+            let crypto_ext = CryptoExt::get_ext();
+
             // Parse certfile argument (str or bytes) to path
             let cert_path = Self::parse_path_arg(&args.certfile, vm)?;
 
@@ -1203,7 +1207,7 @@ mod _ssl {
             }
 
             // Additional validation: Create CertifiedKey to ensure rustls accepts it
-            let signing_key = sign::any_supported_type(&key).map_err(|_| {
+            let signing_key = crypto_ext.any_supported_key(&key).map_err(|_| {
                 vm.new_os_subtype_error(
                     PySSLError::class(&vm.ctx).to_owned(),
                     None,
@@ -1525,7 +1529,8 @@ mod _ssl {
             // Dynamically generate cipher list from rustls ALL_CIPHER_SUITES
             // This automatically includes all cipher suites supported by the current rustls version
 
-            let cipher_list = ALL_CIPHER_SUITES
+            let cipher_list = CryptoExt::get_ext()
+                .all_ciphers_or_default()
                 .iter()
                 .map(|suite| {
                     // Extract cipher information using unified helper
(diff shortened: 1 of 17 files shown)


crates/stdlib/src/ssl.rs
```diff
@@ -77,14 +77,23 @@ mod _ssl {
...
     use sha2::{Digest, Sha256};
 
+    #[cfg(all(feature = "ssl-rustls-aws-lc-rs", feature = "ssl-rustls-ring"))]
+    compile_error!("Enable only one rustls provider: ssl-rustls-aws-lc-rs or ssl-rustls-ring");
+
+    #[cfg(feature = "ssl-rustls-aws-lc-rs")]
+    use rustls::crypto::aws_lc_rs::{ALL_CIPHER_SUITES, Ticketer, sign};
+
+    #[cfg(feature = "ssl-rustls-ring")]
+    use rustls::crypto::ring::{ALL_CIPHER_SUITES, Ticketer, sign};
```
youknowone: every line is repeating the other line and the only difference is submodule name. doesn't rustls really support something like `use rustls::crypto::{ALL_CIPHER_SUITES, Ticketer, sign}`; ?

joshuamegnauth54: Nope. :laughing: The platform itself has to implement those. Also, both of `rustls`' built-in platforms can be enabled at the same time so the crate can't export the items like that.

youknowone: That's unfortune. Then I'd like to ask one more thing. Should we really need to add "ssl-rustls-ring" and "ssl-rustls-aws-ring" feature? There are also many third party providers for rustls. We cannot add all the features for those providers. Wouldn't there be any possible way to specify the crypto provider outside of the library?

joshuamegnauth54: > Should we really need to add "ssl-rustls-ring" and "ssl-rustls-aws-ring" feature?

Nope, and I can remove it if you prefer. :grin: I mostly added `ring` because `rustls` has it built-in, so it was a good way to test if the refactor actually works out.

> Wouldn't there be any possible way to specify the crypto provider outside of the library?

Yes, I think so. To clarify, this PR is primarily a refactor to see how user specified crypto providers would work out. I think I _mostly_ figured it out, which is why `ring` works, but I wanted to check with you and the other maintainers to see if I'm on the right track.

So, my intention is:

This patch -> Preliminary work to reduce explicitly depending on `aws-lc-rs`
Next patch -> Provide a way for user to set a crypto provider (#7059)

youknowone: > Should we really need to add "ssl-rustls-ring" and "ssl-rustls-aws-ring" feature?

This was more like a technical question, "Can ring to be added without adding its own feature?" 

Yeah, this is the right way in general. I try to understand the dedicated features are really required or not as the last step of review.

Another trial: If we turn on rustls-no-provider, can we avoid dedicated features? Like using `#[cfg(all(feature = "ssl-rustls", feature = "ring"))]`

youknowone: @ShaharNaveh Could you help? will this dedicated features be the only way that we can support mulitple provider?

joshuamegnauth54: Sorry, I've been a bit busy so I forgot to respond. :laughing: 

I'll remove `ring`. I think the final iteration will only have `ssl-rustls` (which will use `aws-lc-rs`) and `ssl-rustls-no-provider` (user will have to set a provider).

I'll draft this and work on it next week.

youknowone: Does `ssl-rustls-no-provider` allow users to set the provider ring? Then that sounds perfect. Please tell me I am asking something impossible. I don't know well how to handle rustls features well yet.

This might be out of scope of this PR. Just sharing a note for future implementation.
If `ssl-rustls-no-provider` allows user to select any kind of provider including aws-lc-rs and ring,
I prefer to keep rustpython-stdlib always set to no-provider as default and `rustpython` the shell crate to override it as default.

ShaharNaveh: > @ShaharNaveh Could you help? will this dedicated features be the only way that we can support mulitple provider?

I don't have experience in this area:/
My inputs wouldn't be valuable here:(

joshuamegnauth54:
Temporarily undrafted because I want CodeRabbit's review. :laughing:

crates/stdlib/src/ssl.rs
```diff
@@ -612,7 +615,10 @@ mod _ssl {
             return Err("No cipher can be selected".to_string());
         }
 
-        let all_suites = ALL_CIPHER_SUITES;
+        let crypto_ext = CRYPTO_EXT
+            .get()
+            .expect("A CryptoProvider must be set before SSL");
+        let all_suites = crypto_ext.all_ciphers_or_default();
```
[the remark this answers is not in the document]
joshuamegnauth54: Panicking is intentional. A provider should always be set. Returning an error is misleading in this case.

crates/stdlib/src/ssl/compat.rs
```diff
@@ -57,14 +56,114 @@ pub(super) const VERIFY_X509_PARTIAL_CHAIN: i32 = 0x80000;
...
+                if key.choose_scheme(&[scheme]).is_some() {
+                    return Ok(key);
+                }
+            }
+
+            Err(rustls::Error::General(
+                "failed to parse private key as RSA, ECDSA, or EcDSA".into(),
+            ))
+        })
+    }
```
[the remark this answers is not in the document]
joshuamegnauth54: I copied the error from: https://docs.rs/rustls/latest/src/rustls/crypto/ring/sign.rs.html#21-39 but the typo is my own fault. :laughing:

I disagree with your suggestion about collecting into a vector. I wanted to avoid the allocation.

crates/stdlib/src/ssl.rs
joshuamegnauth54: I made a lot of progress with this but now I'm stuck. :face_with_spiral_eyes: Running the interpreter works and the example I wrote in `examples/tls_providers` works, but I'm not sure why `cargo test` is failing.

The error says that the encodings module can't be found, but I didn't touch anything related to that. :thinking:  I'm not sure if it's feature gate related.

ShaharNaveh (approved):
This looks great! tysm:)

ShaharNaveh:
@youknowone CI fails here because periodic checks took too long. unrelated to the code changes

youknowone (approved):
I am sorry that missing this

youknowone:
@joshuamegnauth54 Thank you so much!

PR #7660, opened 2026-04-23
Remove unused rust impl for formatting dis output

ShaharNaveh:
This was used back when we had a hacky `_dis` module (that was removed at #6690).
I guess we kept maintaining the `fmt_dis` method as it was still technicaly used as we would implement `Display` for PyCode, but since we no longer actually use it anywhere it's dead code.

`Display` of PyCode was called here:
https://github.com/RustPython/RustPython/blob/9e225f52e23ce13342232abb0dea3ab18d729229/crates/stdlib/src/dis.rs#L39-L43

Close #7433

## Summary by CodeRabbit

* **Removed Features**
  * Bytecode disassembly/display formatting and the disassembly example program have been removed.

* **Tests**
  * Several snapshot-based tests were removed.
  * New parameterized unit tests added to validate disassembly output across various language constructs.

* **Chores**
  * Workspace test dependencies updated (insta bumped) and test-case introduced for workspace testing.


ShaharNaveh:
@youknowone seems that the places where we are using our implementation of `dis.dis` is at:
- snapshot testing with cargo-insta.
- [examples/dis.rs](https://github.com/RustPython/RustPython/blob/f2e055f7d6f085b8bb52e80ba2d5de96228189ee/examples/dis.rs)

What do you think will be the best approach with solving this?

youknowone:
lets remove them if obsolete

fanninpm:
> Remove unused dep

🎉

crates/codegen/src/compile.rs
```diff
@@ -11620,26 +11620,6 @@ mod tests {
...
-            )
-        };
-    }
-
-    #[test]
-    fn test_if_ors() {
-        assert_dis_snapshot!(compile_exec(
-            "\
-if True or False or False:
-    pass
```
youknowone, line 12293: oh.. are these test going to be entirely deleted? This is not desired. Can we replace the snapshot test to other way not depending on internal dis?

ShaharNaveh: The only way I can think of is to move them under `crates/vm` or `crates/stdlib`, then parsing the source code into a PyCode, then run it like:
```rust
let module = vm.import("dis", 0)?;
let dis = module.get_attr("dis", vm)?;
let res = dis.call(my_pycode, vm)?;

// <convert res to String and pass to Insta>
```

crates/stdlib/src/_opcode.rs
```diff
@@ -273,3 +273,61 @@ mod _opcode {
...
+            .enter(|vm| {
+                vm.insert_sys_path(vm.new_pyobj(concat!(env!("CARGO_MANIFEST_DIR"), "/../../Lib")))
+                    .unwrap();
+
+                let code = vm
+                    .compile(source, Mode::Exec, "?".into())
+                    .map_err(|err| vm.new_syntax_error(&err, Some(source)))
+                    .unwrap();
+
+                let dis_module = vm.import("dis", 0).unwrap();
```
ShaharNaveh: It seems like no matter what I do I can't import "dis". I'd appreciate some help on this.
@fanninpm @youknowone

crates/codegen/src/compile.rs
ShaharNaveh: @youknowone lmk what you think

Cargo.toml
```diff
@@ -186,7 +186,7 @@ getrandom = { version = "0.3", features = ["std"] }
 glob = "0.3"
 hex = "0.4.3"
 indexmap = { version = "2.14.0", features = ["std"] }
-insta = "1.47"
+insta = "1.46.3"
```
fanninpm: Did you check what version `Cargo.lock` has?

ShaharNaveh: ty! overlooked it. should good now

ShaharNaveh:
Tests pass when using `insta = "1.46.3"`. But fails when using `insta = "1.47.2"`.

bart-ouch-gif-bart-ouch-idiot


I'll open a ticket to them

ShaharNaveh:
Now it's happening with insta 1.46.3 :man_facepalming:

ShaharNaveh:
Looks like setting `CI: true` fixed it

.github/workflows/ci.yaml
```diff
@@ -77,7 +78,7 @@ jobs:
         run: cargo clippy ${{ env.CARGO_ARGS }} --workspace --all-targets ${{ env.WORKSPACE_EXCLUDES }} -- -Dwarnings
 
       - name: run rust tests
-        run: cargo test --workspace ${{ env.WORKSPACE_EXCLUDES }} --verbose --features threading ${{ env.CARGO_ARGS }}
+        run: cargo test --workspace ${{ env.WORKSPACE_EXCLUDES }} --features threading ${{ env.CARGO_ARGS }}
```
fanninpm, line 115:
```suggestion
        run: cargo test --workspace ${{ env.WORKSPACE_EXCLUDES }} --features threading ${{ env.CARGO_ARGS }}
        env:
          CI: true
```

What about here?

ShaharNaveh: I still think there's a value to putting it at the top. other tools in the future (or even exiting ones) might look at this value, it because a standard from what I can tell. but it's not a hill I'm willing to die on

ShaharNaveh:
that still didn't fix it
https://github.com/RustPython/RustPython/actions/runs/25154229577/job/73731856643?pr=7660

:cry:

ShaharNaveh:
@youknowone do we really need those snapshot tests? what are we trying to test? what do you look for in the snapshot?

~~Im having a tough time with insta for some reason, it seems like it fails every second run for some reason. can I convert those tests to be at the codegen, and look for something in the bytecode instead?~~

~~Also, test_peepholer & test_dis are already checking quite a lot, I'm not sure that those snapshots are needed at all~~

nvm, figured it out. Sometimes insta can't detect to workspace root

youknowone (approved):

youknowone:
I expect a few compiler bugs can be detected in unittest, instead of hitting Python test. But that might not be worth enough because `dis` module implies we already ran the full interpreter

ShaharNaveh:
I've fixed the merge conflicts

youknowone:
@ShaharNaveh I am sorry, please rebase it once again. I enqueued it to merge queue

ShaharNaveh:
> @ShaharNaveh I am sorry, please rebase it once again. I enqueued it to merge queue

All good lol, I've merged main

PR #7663, opened 2026-04-23
Fix pyclass memory layout to prevent silent UB in inherited getter dispatch

1ndahous3:
Derived `#[pyclass]` structs that mix pointer-sized and non-pointer-sized fields are silently miscompiled: Rust's default `#[repr(Rust)]` layout algorithm places the base field at a **non-zero offset**, while the inherited getter dispatcher unconditionally assumes it is at offset 0.

The result is undefined behaviour that manifests as a SIGSEGV at runtime.

This PR fixes the root cause by:
1. Auto-inserting `#[repr(C)]` on derived structs that carry no explicit repr (guarantees declaration-order layout, keeping the base field at offset 0).
2. Emitting a `const_assert` via `offset_of!` as belt-and-suspenders: catches the remaining case where a user supplies an explicit repr that does not guarantee offset 0.

Here is my minimal code that causes SIGSEGV/ACCESS_VIOLATION on Windows/Linux:
```rust
use rustpython_vm::Interpreter;
use rustpython_vm::PyObjectRef;
use rustpython_vm::pyclass;
use rustpython_vm::class::PyClassImpl;
use rustpython_vm::PyPayload;

fn main() {

    #[pyclass(module = false, name = "Base")]
    #[derive(Debug, PyPayload)]
    struct Base {
        field: PyObjectRef,
    }

    #[pyclass]
    impl Base {
        #[pygetset]
        fn field(&self) -> PyObjectRef {
            self.field.clone()
        }
    }

    #[pyclass(module = false, name = "Derived", base = Base)]
    #[derive(Debug)]
    struct Derived { // No #[repr(C)] — triggers the bug
        _base: Base,
        extra: Option<u64>,  // displaces _base in memory under #[repr(Rust)]
    }

    #[pyclass]
    impl Derived {
        #[pygetset]
        fn extra(&self) -> Option<u64> {
            self.extra
        }

        // No `field` getter — relies on inherited Base getter -> crash
    }

    let interp = Interpreter::without_stdlib(Default::default());
    interp.enter(|vm| {
        Base::make_static_type();
        Derived::make_static_type();

        let obj = vm.new_pyobj(Derived {
            _base: Base { field: vm.ctx.new_int(42).into() },
            extra: Some(99),
        });

        // Accessing `extra` (own getter) is fine.
        let _extra = obj.get_attr("extra", vm).expect("extra getter ok");

        // Accessing `field` (inherited getter) -> SIGSEGV without the fix
        let _field = obj.get_attr("field", vm).expect("field getter ok");
    });
}
```

## Summary by CodeRabbit

* **Refactor**
  * Strengthened compile-time validation of struct memory layout for base classes, now detecting and validating the primary base field and asserting correct offsets.
  * Automatically enforce a compatible memory representation when a base is declared so derived types meet layout expectations.
  * Removed several explicit memory-layout annotations from internal types to streamline layout handling and reduce redundant declarations.

### DIFF
--- a/crates/derive-impl/src/pyclass.rs
+++ b/crates/derive-impl/src/pyclass.rs
@@ -387,8 +387,7 @@ fn ensure_repr_c(mut item: Item) -> Item {
     };
     let has_repr = s.attrs.iter().any(|attr| attr.path().is_ident("repr"));
     if !has_repr {
-        let repr_c: syn::Attribute = parse_quote!(#[repr(C)]);
-        s.attrs.push(repr_c);
+        s.attrs.push(parse_quote!(#[repr(C)]));
     }
     item
 }
--- a/crates/derive-impl/src/pyclass.rs
+++ b/crates/derive-impl/src/pyclass.rs
@@ -573,14 +573,20 @@ pub(crate) fn impl_pyclass(attr: PunctuatedNestedMeta, item: Item) -> Result<Tok
     let fake_ident = Ident::new("pyclass", item.span());
     let (class_meta, class_name, module_name, base, metaclass, unhashable) = {
         let (ident, _) = pyclass_ident_and_attrs(&item)?;
-        let class_meta =
-            ClassItemMeta::from_nested(ident.clone(), fake_ident, attr.into_iter())?;
+        let class_meta = ClassItemMeta::from_nested(ident.clone(), fake_ident, attr.into_iter())?;
         let class_name = class_meta.class_name()?;
         let module_name = class_meta.module()?;
         let base = class_meta.base()?;
         let metaclass = class_meta.metaclass()?;
         let unhashable = class_meta.unhashable()?;
-        (class_meta, class_name, module_name, base, metaclass, unhashable)
+        (
+            class_meta,
+            class_name,
+            module_name,
+            base,
+            metaclass,
+            unhashable,
+        )
     };
 
     // When a base is specified:
--- a/crates/derive-impl/src/pyclass.rs
+++ b/crates/derive-impl/src/pyclass.rs
@@ -318,12 +318,13 @@ pub(crate) fn impl_pyclass_impl(attr: PunctuatedNestedMeta, item: Item) -> Resul
     Ok(tokens)
 }
 
-/// Validates that when a base class is specified, the struct has the base type as its first field.
-/// This ensures proper memory layout for subclassing (required for #[repr(transparent)] to work correctly).
-fn validate_base_field(item: &Item, base_path: &syn::Path) -> Result<()> {
+/// Validates that when a base class is specified, the struct has the base type as its first
+/// *declared* field.  Returns a token naming that field (e.g. `_base` or `0` for tuple structs)
+/// so the caller can emit a compile-time `offset_of!` assertion, or `None` for non-structs.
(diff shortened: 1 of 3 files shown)


crates/derive-impl/src/pyclass.rs
```diff
@@ -369,8 +373,24 @@ fn validate_base_field(item: &Item, base_path: &syn::Path) -> Result<()> {
...
+/// reorder fields to minimise padding, silently displacing the base field.  `#[repr(C)]`
+/// preserves declaration order, guaranteeing offset 0 for the first field.
+fn ensure_repr_c(mut item: Item) -> Item {
+    let Item::Struct(ref mut s) = item else {
+        return item;
+    };
+    let has_repr = s.attrs.iter().any(|attr| attr.path().is_ident("repr"));
+    if !has_repr {
+        let repr_c: syn::Attribute = parse_quote!(#[repr(C)]);
+        s.attrs.push(repr_c);
```
fanninpm: Would this work?

```suggestion
        s.attrs.push(parse_quote!(#[repr(C)]));
```

fanninpm:
@youknowone @ShaharNaveh on a previous run, reviewdog complained:

> reviewdog: failed to post a review comment: POST https://api.github.com/repos/RustPython/RustPython/pulls/7663/reviews: 403 Resource not accessible by integration []

> reviewdog: This GitHub Token doesn't have write permission of Review API [1],
so reviewdog will report results via logging command [2] and create annotations similar to
github-pr-check reporter as a fallback.
[1]: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target
[2]: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions

youknowone (commented):
I have a question about the current situation. As I understand it, `#[repr(C)]` is only required for types that have derived implementations. Is that correct?
If even base types without any derived implementations also require `#[repr(C)]`, then I would agree with this patch. Otherwise, I would prefer to keep `#[repr(C)]` explicit, and rather than having the macro automatically add `#[repr(C)]`, I would prefer it to emit a warning when a base type is missing `#[repr(C)]`.

1ndahous3:
@youknowone no, this fix is ​​for derived types, not base types (nothing changes for them), but yes, `#[repr(C)]` is set for all derived classes where no `#[repr(` is explicitly specified. Base classes are unaffected. I figured that setting an unconditional `#[repr(C)]` (even on those derived classes where we're lucky and Rust doesn't optimize the layout) wouldn't be a bad thing.

If you don't like this mechanic, I see two other options:
1) Remove the implicit addition of `#[repr(C)]` and leave only the assertion for the actual incorrect offset. Then we'll have to set `#[repr(C)]` manually in derived classes, but not always.
2) Try adding `#[repr(C)]` only if no other `#[repr` is specified (as is the case now) and we've determined that we have an incorrect layout (however, the check will need to be applied twice: as a trigger for adding the `#[repr(C)]` and as a final assert to ensure the final layout is correct.

Honestly, I'd keep the current fix, as I don't see anything wrong with unconditionally `#[repr(C)]` for all derived classes.

youknowone (approved):
Thank you for the explanation.

youknowone:
And welcome to RustPython project!

PR #7673, opened 2026-04-24
Accept surrogates in _json.encode_basestring{,_ascii}

changjoon-park:
## Summary

`json.dumps` raised `UnicodeEncodeError` on any Python `str` containing lone surrogate code points (U+D800–U+DFFF), even though CPython accepts such strings and emits them either as `\uXXXX` escapes (`ensure_ascii=True`) or as raw WTF-8 bytes (`ensure_ascii=False`).

Root cause was the declared argument type of `_json.encode_basestring` and `_json.encode_basestring_ascii`. Both took `PyUtf8StrRef`, which forces a `Python str → &str` conversion that rejects surrogates. The same file's `scanstring` already uses the correct pattern (`PyStrRef` + `s.as_wtf8()`).

See the commit message for the detailed per-file breakdown.

### CPython 3.13.4 byte-identical parity

| Test | Cases | Result |
| --- | --- | --- |
| Targeted manual probes | 16 | All byte-identical |
| Random fuzz (seed=42) | 10,000 strings × 2 modes | All byte-identical |

Fuzz generator sampled across ASCII (25%), BMP pre-surrogate (20%), surrogate range (20%), BMP post-surrogate (20%), and non-BMP (15%) code points. Exposed the two pre-existing `ESCAPE_CHARS` typos fixed in the same commit.

### Test suite

```
$ ./target/release/rustpython -m unittest test.test_json
Ran 214 tests in 108s
OK (skipped=4, expected failures=12)
```

0 failures, 0 unexpected successes.

### Pre-push

- `cargo fmt --all --check` clean
- `cargo clippy -p rustpython-stdlib --all-targets -- -D warnings` clean

## Scope

Decoder path (`JSONDecoder` scanner, `scan_once`, several `pystr: PyUtf8StrRef` entry points) is unchanged in this PR. That is a separate audit, deferred to a follow-up. `test_single_surrogate_decode` remains skipped with an updated reason.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected control-character escaping so JSON output no longer mis-escapes edge-case bytes.
  * Fixed escape mappings for previously mishandled characters.

* **Improvements**
  * JSON string encoding now preserves non-standard Unicode sequences (including lone surrogates) instead of corrupting them.
  * Public JSON encoding functions produce encoding results that safely retain original byte sequences and behave correctly in ASCII-only vs. full-Unicode modes.

### DIFF
--- a/Lib/test/test_json/test_unicode.py
+++ b/Lib/test/test_json/test_unicode.py
@@ -138,14 +138,6 @@ class TestUnicode:
 
 class TestPyUnicode(TestUnicode, PyTest): pass
 class TestCUnicode(TestUnicode, CTest):
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
-    def test_ascii_non_printable_encode(self):
-        return super().test_ascii_non_printable_encode()
-
-    @unittest.skip("TODO: RUSTPYTHON; panics with 'str has surrogates'")
+    @unittest.skip("TODO: RUSTPYTHON; decode path still uses PyUtf8StrRef")
     def test_single_surrogate_decode(self):
         return super().test_single_surrogate_decode()
-
-    @unittest.skip("TODO: RUSTPYTHON; panics with 'str has surrogates'")
-    def test_single_surrogate_encode(self):
-        return super().test_single_surrogate_encode()
--- a/crates/stdlib/src/json.rs
+++ b/crates/stdlib/src/json.rs
@@ -667,24 +667,30 @@ mod _json {
         }
     }
 
-    fn encode_string(s: &str, ascii_only: bool) -> String {
+    fn encode_string(wtf8: &rustpython_common::wtf8::Wtf8, ascii_only: bool) -> Wtf8Buf {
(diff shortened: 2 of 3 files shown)


youknowone (approved):
👍

changjoon-park:
> 👍

thanks for the review !

youknowone:
Thank you!

changjoon-park:
> Thank you!

Thanks for the review :)

changjoon-park:
Retroactive verification against CPython 3.14.4 (RustPython's stated target version per README): re-ran the encoder fuzz at 20,000 cases (10,000 each for `ensure_ascii=True` and `ensure_ascii=False`, seed=42, codepoint distribution covering ASCII / BMP pre-surrogate / surrogate range / BMP post-surrogate / non-BMP). All 20,000 byte-identical with CPython 3.14.4. Original PR body cited 3.13.4 — for completeness, behavior is identical between the two versions for `json.dumps`. No change to merged code; this is a verification-trail comment only.

PR #7675, opened 2026-04-24
Accept surrogates in _json.JsonScanner decode path

changjoon-park:
## Summary

`json.loads` had two distinct failure modes on JSON strings containing lone surrogate code points (U+D800–U+DFFF):

1. **`UnicodeEncodeError`** when the Python `str` passed to `json.loads` itself carried a surrogate (e.g., `json.loads('"\ud83d"')`) — `JsonScanner::Callable::call` called `try_into_utf8` at entry.
2. **Silent `U+FFFD` corruption** when surrogates arrived via JSON `\uXXXX` escapes inside arrays or dict keys (e.g., `json.loads('["\\uD83D"]')` → `['�']`) — `call_scan_once` and `parse_object` collapsed the scanner's `Wtf8Buf` output through `.to_string()`, which routes through `impl Display for Wtf8` (intentionally lossy with `U+FFFD`).

CPython accepts surrogate-bearing strings in both positions and preserves them.

Root-level strings and mid-string surrogates (e.g., `"a\uD83Db"`) already worked because `JsonScanner::parse` put `wtf8_result` directly into a tuple, routing through `impl ToPyObject for Wtf8Buf` which preserves. Surrogate pair combining (`😀` → `😀`) also worked and is unchanged (handled in `machinery::scanstring`).

Switch `JsonScanner`'s five `PyUtf8StrRef` signatures to `PyStrRef`, drop the `try_into_utf8` at `Callable::call` entry, and feed `Wtf8Buf` directly to `new_str` instead of `.to_string()`. Memo keys change from `HashMap<String, PyStrRef>` to `HashMap<Wtf8Buf, PyStrRef>` (`Wtf8Buf` already implements `Hash`). `parse_number` takes `&[u8]` since JSON numbers are ASCII. `machinery::scanstring` is unchanged.

This extends the WTF-8 refactor pattern from #7673 to the decoder side.

### Targeted probes

| Input | Before | After | CPython |
|---|---|---|---|
| `json.loads('"\ud83d"')` | UnicodeEncodeError | `'\ud83d'` | `'\ud83d'` |
| `json.loads('"\\uD83D"')` | `'\ud83d'` | `'\ud83d'` | `'\ud83d'` |
| `json.loads('"\\uD83D\\uDE00"')` | `'😀'` | `'😀'` | `'😀'` |
| `json.loads('["\\uD83D"]')` | `['�']` | `['\ud83d']` | `['\ud83d']` |
[...]

PR #7676, opened 2026-04-24
Report invalid \uXXXX escape position at the u character

changjoon-park:
## Summary

`json.loads` of a string with an invalid `\uXXXX` escape reported a decode error position one character earlier than CPython: at the preceding `\` instead of at the `u` specifier. For surrogate-pair paths (`\uXXXX\uYYYY` where the second escape is invalid), the position was off by more — it landed on the first hex digit of the *first* escape rather than on the second `u`.

Examples (before / after):

| Input | CPython pos | Before | After |
|---|---|---|---|
| `json.loads('"\\uG000"')` | 2 | 1 | 2 |
| `json.loads('"abc\\uG000"')` | 5 | 4 | 5 |
| `json.loads('"\\u1234\\uG000"')` | 8 | 7 | 8 |
| `json.loads('"\\ud83d\\uG000"')` | 8 | 3 | 8 |
| `json.loads('"\\ud83d\\u"')` | 8 | 3 | 8 |
| `json.loads('"\\ud83d\\uDE0G"')` | 8 | 3 | 8 |

## Fix

In ``crates/stdlib/src/json/machinery.rs::scanstring``:

- Primary ``\uXXXX`` call: pass ``char_offset + next_char_i`` (position of the ``u``) instead of ``char_offset + char_i`` (position of the ``\``).
- Surrogate-pair second-escape call: capture the second ``u``'s char index from the ``next_tuple()`` peek and pass ``char_offset + u2_char_i``. The previous ``char_offset + next_char_i + 1`` referred to the *first* escape's position — unrelated to where the second escape fails.

No behavioural change on the success path, on non-``\u`` ``\escape`` errors, on unterminated strings, or on lone-surrogate preservation (pair filter unchanged). Decoded output bit-for-bit identical.

### Targeted probes (13 cases)

Invalid hex digit at position 1/2/3/4, short escapes (``\u``, ``\u1``, ``\u12``, ``\u123``), escape preceded by ASCII, valid-then-invalid, and surrogate-pair error paths. All positions match CPython 3.13.4 after the fix.

### Test suite

```
$ ./target/release/rustpython -m unittest test.test_json
Ran 214 tests in 111s
OK (skipped=4, expected failures=12)
```

No regressions.

### Pre-push

- ``cargo fmt --all --check`` clean
- ``cargo clippy -p rustpython-stdlib --all-targets -- -D warnings`` clean
- ``prek run --all-files`` — all hooks pass

## Scope

Single file, +7/-4 lines (3 substantive + 4 comment lines). Independent from #7675 (decoder WTF-8 refactor) — this PR touches ``machinery::scanstring`` error-reporting path only; #7675 touches ``json.rs`` scanner frontend. Either order of merge is fine.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved JSON parsing error reporting for malformed Unicode escape sequences so error positions point precisely at the invalid escape, matching expected behavior and making debugging clearer.

* **Tests**
  * Added tests that validate invalid Unicode-escape handling and confirm reported error positions match the expected locations.

### DIFF
--- a/crates/stdlib/src/json/machinery.rs
+++ b/crates/stdlib/src/json/machinery.rs
@@ -231,17 +231,20 @@ pub fn scanstring<'a>(
                     'r' => "\r",
                     't' => "\t",
                     'u' => {
-                        let mut uni = decode_unicode(&mut chars, char_offset + char_i)?;
+                        // Error position for an invalid \uXXXX escape points at the
+                        // `u`, not the `\` -- matches CPython's json.decoder.
+                        let mut uni = decode_unicode(&mut chars, char_offset + next_char_i)?;
                         chunk_start = byte_i + 6;
                         if let Some(lead) = uni.to_lead_surrogate() {
                             // uni is a surrogate -- try to find its pair
                             let mut chars2 = chars.clone();
-                            if let Some(((_, (byte_pos2, _)), (_, _))) = chars2
+                            if let Some(((_, (byte_pos2, _)), (u2_char_i, (_, _)))) = chars2
                                 .next_tuple()
                                 .filter(|((_, (_, c1)), (_, (_, c2)))| *c1 == '\\' && *c2 == 'u')
                             {
-                                let uni2 =
-                                    decode_unicode(&mut chars2, char_offset + next_char_i + 1)?;
+                                // u2_char_i is the `u` of the second \uXXXX; same
+                                // position convention as the primary call above.
+                                let uni2 = decode_unicode(&mut chars2, char_offset + u2_char_i)?;
                                 if let Some(trail) = uni2.to_trail_surrogate() {
                                     // ok, we found what we were looking for -- \uXXXX\uXXXX, both surrogates
                                     uni = lead.merge(trail).into();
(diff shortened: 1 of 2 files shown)


youknowone (commented):
was there any test affected by this change?

changjoon-park:
Re-verified against CPython 3.14.4 (RustPython's stated target version): all 13 escape-position probe cases match. Original PR body referenced 3.13.4 because that was the system Python at probe time — for completeness, the error position convention is identical between 3.13 and 3.14 (the validator code in Python/ast.c has been stable for this since at least 3.10).

changjoon-park:
there was no existing cpython test in Lib/test/test_json/ checks the exact "pos" value for invalid \uXXXX escapes — they only regex-match the message text, so this fix doesn't unmask any test

so i added a focused regression test in extra_tests/snippets/stdlib_json.py covering both the leading-position case ("\uXYZW" → pos=2) and the offset case ("abc\uZZZZ" → pos=5). Verified byte-identical against CPython 3.14.4

thanks for catching !

youknowone (approved):
👍

PR #7688, opened 2026-04-26
Enforce int_max_str_digits on int-to-str conversions

changjoon-park:
RustPython enforced `sys.get_int_max_str_digits()` on the str → int direction (via `bytes_to_int`) but not the int → str direction. CPython 3.14 enforces both per PEP 644's DoS mitigation. This adds the missing checks.

## Entry points covered

| Path | Site |
|---|---|
| `repr(int)` | `builtins/int.rs::repr_str` |
| `str(int)` | `protocol/object.rs::str` fast path |
| f-string, `format(int, …)` | `builtins/int.rs::__format__` (decimal/`n`/empty only) |
| `'%d' % int` (str + bytes %) | `vm/cformat.rs` Decimal{D,I,U} branches |

Binary bases (`'x'`, `'o'`, `'b'`) remain exempt per CPython spec.

## Implementation

- New `check_int_to_str_digits` helper in `builtins::int` — bit-count fast path (skip if < 1500 bits ≈ 452 digits), then upper bound `bits × log10(2) + 1` vs the configured limit.
- New `FormatSpec::is_decimal_int_format()` in `common::format` for the decimal-vs-binary branch decision in `__format__`.

## Tests unmasked (8)

- `test_int.IntStrDigitLimitsTests::test_max_str_digits`
- `test_int.IntStrDigitLimitsTests::test_denial_of_service_prevented_int_to_str`
- `test_int.IntStrDigitLimitsTests::test_int_from_other_bases`
- (same 3 mirrored in `IntSubclassStrDigitLimitsTests`)
- `test_ast.AST_Tests::test_repr_large_input_crash`
- `test_reprlib.ReprTests::test_numbers`

## Verification

- 8 modules pass (`test_int test_ast test_reprlib test_decimal test_fractions test_compile test_sys test_cmd_line`), 1,276 tests, no regressions
- Snippet at `extra_tests/snippets/builtin_int.py` covers all 9 decimal paths + 3 binary-base exemptions + disabled-limit; passes on RustPython and CPython 3.14.4
- Boundary cases (4299/4300/4301 digits at limit=4300) byte-identical with CPython 3.14.4

## Summary by CodeRabbit

* **New Features**
  * Decimal integer-to-string conversions now enforce configured digit limits and raise ValueError when exceeded across str(), repr(), format(), f-strings, and percent formatting; non-decimal (hex/oct/bin) formatting remains unaffected.
* **Tests**
  * Added a regression test validating the digit-limit behavior and the ability to disable the limit.

### DIFF
--- a/Lib/test/test_int.py
+++ b/Lib/test/test_int.py
@@ -588,8 +588,8 @@ class IntStrDigitLimitsTests(unittest.TestCase):
         with self.assertRaises(ValueError):
             str(i)
 
-    @unittest.skipUnless(
-        hasattr(time, "get_clock_info"),
+    @unittest.expectedFailureIf(
+        not hasattr(time, "get_clock_info"),
         "TODO: RUSTPYTHON; time.get_clock_info is not available on wasm",
     )
     def test_denial_of_service_prevented_int_to_str(self):
--- a/Lib/test/test_int.py
+++ b/Lib/test/test_int.py
@@ -1,4 +1,5 @@
 import sys
+import time
 
 import unittest
 # TODO: RUSTPYTHON
@@ -587,6 +588,10 @@ class IntStrDigitLimitsTests(unittest.TestCase):
         with self.assertRaises(ValueError):
             str(i)
 
+    @unittest.skipUnless(
+        hasattr(time, "get_clock_info"),
+        "requires time.get_clock_info (not available on wasm)",
+    )
     def test_denial_of_service_prevented_int_to_str(self):
         """Regression test: ensure we fail before performing O(N**2) work."""
         maxdigits = sys.get_int_max_str_digits()
--- a/Lib/test/test_int.py
+++ b/Lib/test/test_int.py
@@ -573,7 +573,6 @@ class IntStrDigitLimitsTests(unittest.TestCase):
             else:
                 self.int_class(i, base)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
     def test_max_str_digits(self):
         maxdigits = sys.get_int_max_str_digits()
 
@@ -588,7 +587,6 @@ class IntStrDigitLimitsTests(unittest.TestCase):
         with self.assertRaises(ValueError):
             str(i)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
     def test_denial_of_service_prevented_int_to_str(self):
         """Regression test: ensure we fail before performing O(N**2) work."""
         maxdigits = sys.get_int_max_str_digits()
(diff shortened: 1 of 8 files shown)


ShaharNaveh (approved):
TYSM for all the PRs recently:)

changjoon-park:
> TYSM for all the PRs recently:)

thanks for checking !  and also, i pushed a follow-up that skips the DoS test on wasm — it depends on time.get_clock_info, which is unix/windows-only in RustPython and surfaces as AttributeError via support.Stopwatch.

thanks for keep watching :)

Lib/test/test_int.py
```diff
@@ -588,7 +588,10 @@ def test_max_str_digits(self):
         with self.assertRaises(ValueError):
             str(i)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
+    @unittest.skipUnless(
+        hasattr(time, "get_clock_info"),
+        "requires time.get_clock_info (not available on wasm)",
```
ShaharNaveh:
```suggestion
        "TODO: RUSTPYTHON; requires time.get_clock_info (not available on wasm)",
```

ShaharNaveh (changes requested):
For custom patches we need to set some type of comment saying that this is our patch so [update_lib](https://github.com/RustPython/RustPython/tree/main/scripts/update_lib) can migrate the custom patches:)

Lib/test/test_int.py
```diff
@@ -588,7 +588,10 @@ def test_max_str_digits(self):
         with self.assertRaises(ValueError):
             str(i)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
+    @unittest.skipUnless(
```
ShaharNaveh: Also, I'd set it to `unittest.expectedFailireIf(not hasattr(time, "get_clock_info"), "TODO: RUSTPYTHON; ...")`

changjoon-park:
thanks and all is done
the "expectedFailureIf" is the right call i should've checked it thanks for catching :)

ShaharNaveh (approved):
ty! looks great:)

youknowone (approved):
👍

Lib/test/test_int.py
```diff
@@ -588,7 +588,10 @@ def test_max_str_digits(self):
         with self.assertRaises(ValueError):
             str(i)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
+    @unittest.expectedFailureIf(
+        not hasattr(time, "get_clock_info"),
+        "TODO: RUSTPYTHON; time.get_clock_info is not available on wasm",
```
youknowone, line 593: how CPython wasm build handle this?

changjoon-park: it's verified via CPython source

the "time_get_clock_info" in `Modules/timemodule.c` (v3.14.0 lines 1630, 1946) has no cfg guards, so cpython exposes it unconditionally including wasm builds. The "expectedFailureIf" will surface as unexpected success once we lift RustPython's #[cfg(unix)]`/`#[cfg(windows)] gating in crates/vm/src/stdlib/time.rs

thanks for catching !

youknowone: Thank you for checking

PR #7697, opened 2026-04-27
Defer staticmethod/classmethod callable storage to __init__

changjoon-park:
## Background

Python's two-phase object construction splits responsibility between `__new__` (allocation, no required state) and `__init__` (initialization). Subclasses can override either phase independently.

CPython's `staticmethod`/`classmethod` follow this contract strictly: `__new__` allocates a bare descriptor with `__func__ = None`, and `__init__` assigns the callable plus copies wrapper attributes (`__doc__`, `__name__`, `__qualname__`, `__module__`, `__annotations__`) via `functools_wraps` (`Objects/funcobject.c::sm_init` / `cm_init`).

RustPython instead did the same work twice — once in `slot_new` and again in the `Initializer` — so a subclass overriding `__init__` without calling `super().__init__()` still had `__func__` set from `slot_new`.

## Repro

```python
class MyStatic(staticmethod):
    def __init__(self, func):
        pass  # deliberately not calling super().__init__()

def f(): pass
sm = MyStatic(f)

# CPython 3.14 (and now RustPython):
#   repr(sm)          == '<staticmethod(None)>'
#   sm.__func__       is None
#   sm.__wrapped__    is None
#
# RustPython before this PR:
#   repr(sm)          == '<staticmethod(<function f at 0x...>)>'
#   sm.__func__       == f
#   sm.__wrapped__    == f
```

This pattern shows up in libraries that subclass the descriptor types to customize their state (e.g. property-like wrappers, mocking libraries, internal CPython tests for descriptor semantics).

## Fix

- `slot_new` now validates the signature (so `staticmethod()` / `staticmethod(a, b)` still error eagerly) but stores `None` for the callable and skips the wrapper-attribute copy.
- `Initializer::init` stores the callable and runs the wrapper copy, mirroring CPython's split.

Net diff: `+35 / -44` (the wrapper-attr setup just moves between methods, and the `slot_new` path collapses).

## Tests unmasked

- `test_descr.ClassPropertiesAndMethods.test_staticmethod_new`
- `test_descr.ClassPropertiesAndMethods.test_classmethod_new`

(Both were marked `@unittest.expectedFailure  # TODO: RUSTPYTHON; Wrong error message` — the failure was actually behavioral, not a wording issue, but the new behavior matches the test's assertions exactly.)

## Verification

- 16 modules pass with no regressions: `test_descr test_decorators test_funcattrs test_inspect test_functools test_class test_super test_abc test_property test_pickle test_dataclasses test_dis test_metaclass test_call test_module test_typing test_keywordonlyarg` (1,071+ tests)
- CPython 3.14.4 byte-identical on the repro and unmask scenarios
- Edge cases probed: 0-arg / 2-arg construction errors, re-init via `sm.__init__(other)`, subclass that *does* call `super().__init__()`, pickling round-trip, `@staticmethod` / `@classmethod` decorator usage — all match CPython

### DIFF
--- a/Lib/test/test_descr.py
+++ b/Lib/test/test_descr.py
@@ -5193,7 +5193,6 @@ class ClassPropertiesAndMethods(unittest.TestCase):
         with self.assertRaisesRegex(NotImplementedError, "BAR"):
             B().foo
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON; Wrong error message
     def test_staticmethod_new(self):
         class MyStaticMethod(staticmethod):
             def __init__(self, func):
@@ -5204,7 +5203,6 @@ class ClassPropertiesAndMethods(unittest.TestCase):
         self.assertIsNone(sm.__func__)
         self.assertIsNone(sm.__wrapped__)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON; Wrong error message
     def test_classmethod_new(self):
         class MyClassMethod(classmethod):
             def __init__(self, func):
--- a/crates/vm/src/builtins/classmethod.rs
+++ b/crates/vm/src/builtins/classmethod.rs
@@ -66,34 +66,15 @@ impl Constructor for PyClassMethod {
     type Args = PyObjectRef;
 
     fn slot_new(cls: PyTypeRef, args: FuncArgs, vm: &VirtualMachine) -> PyResult {
-        let callable: Self::Args = args.bind(vm)?;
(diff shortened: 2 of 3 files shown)


ShaharNaveh (approved):
lgtm!
tysm!

changjoon-park:
> lgtm! tysm!



> lgtm! tysm!

thanks for the review :)

youknowone (approved):
👍

PR #7707, opened 2026-04-27
Reject format spec with width above i32::MAX

changjoon-park:
## Background

CPython rejects format-spec widths that exceed `Py_ssize_t::MAX` with `ValueError: Too many decimal digits in format string`. RustPython's `FormatSpec::_parse` only capped **precision** (via `parse_precision`); width digits were accepted up to `usize::MAX`, so values like `sys.maxsize + 1` silently produced an effectively-ignored width.

## Repro

```python
import sys
fs = '{}f'.format(sys.maxsize + 1)  # '9223372036854775808f'

format(2.34, fs)
# CPython 3.14:    ValueError: Too many decimal digits in format string
# RustPython:      '2.340000'        (silently dropped huge width, before this PR)
```

## Fix

After `parse_number` returns the width in `FormatSpec::_parse`, reject any value above `i32::MAX` with `FormatSpecError::DecimalDigitsTooMany`. This matches the cap precision already uses and produces the byte-identical CPython `"Too many decimal digits in format string"` ValueError.

## Tests unmasked

- `test_str.StrTest.test_format_huge_width`

## Verification

- CPython 3.14.4 byte-identical wording for the huge-width case
- Precision path unchanged (`test_format_huge_precision` still passes with its existing `"Precision too big"` message)
- Normal widths (e.g. `format(2.34, '100.3f')`) still pad correctly
- No regressions across 7 modules: `test_str`, `test_format`, `test_unicode_identifiers`, `test_fstring`, `test_int`, `test_float`, `test_decimal`

## Release Notes

* **Bug Fixes**
  * Improved validation for format specifications to properly reject overly large width values with appropriate error handling during parsing.

### DIFF
--- a/Lib/test/test_str.py
+++ b/Lib/test/test_str.py
@@ -1463,7 +1463,6 @@ class StrTest(string_tests.StringLikeTest,
         with self.assertRaises(ValueError):
             result = format(2.34, format_string)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON; AssertionError: ValueError not raised
     def test_format_huge_width(self):
         format_string = "{}f".format(sys.maxsize + 1)
         with self.assertRaises(ValueError):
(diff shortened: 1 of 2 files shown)


ShaharNaveh (approved):
🥳 
ty:)

youknowone (approved):
👍

PR #7712, opened 2026-04-28
Make updating actions cache synchronous

ShaharNaveh:
If multiple PRs are being merged in a short period of time multiple runs of the workflow can trigger, and cause a little race condition regarding what is considered the latest cache. This PR makes it so no more than 1 run of updating the cache can run at any given time

### DIFF
--- a/.github/workflows/update-caches.yml
+++ b/.github/workflows/update-caches.yml
@@ -9,6 +9,10 @@ on:
     branches:
       - main
 
+concurrency:
+    group: ${{ github.workflow }}
+    cancel-in-progress: false
+
 env:
   CARGO_INCREMENTAL: 0
   CARGO_TERM_COLOR: always


.github/workflows/update-caches.yml
```diff
@@ -9,6 +9,10 @@ on:
     branches:
       - main
 
+concurrency:
+    group: ${{ github.workflow }}
+    cancel-in-progress: false
```
ShaharNaveh, line 14: I'm on the fence regarding setting this to true.

What do you think @fanninpm?

fanninpm: Setting this to true should be safe as long as all the caches are uploaded at the very end of the workflow.

ShaharNaveh: Setting this to true would cause running jobs to be canceled if there's a new one in queue. So in theory, the caches would never update until there will be no pushes to main for long period of time (depends on how many PRs were merged at that time window).

Setting this to false would never cancel already running jobs, but will skip any in-queue job that wasn't started yet.

fanninpm (approved):
This is good as it currently stands.

PR #7717, opened 2026-04-28
Fix title() and capitalize()

joshuamegnauth54:
I fixed `title()` and `capitalize()` as well as enabled their respective tests. I fixed both by following CPython's logic closely, including implementing a workaround for sigma.

Besides that, I removed an unneeded dependency (unicode-casing) by using icu4x instead. This helps with consistency and removes outdated Unicode tables from the final RustPython binary.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Unicode string case behavior (capitalize, title, istitle), including correct handling of Greek sigma, Turkish dotted I, German ß, combining marks, emoji-leading text, and preservation of invalid bytes.

* **Chores**
  * Switched to ICU-based Unicode libraries and added a formatting/helper dependency.

* **Refactor**
  * Reworked titlecasing and ASCII-bytes logic for clarity and reuse.

* **Tests**
  * Enabled and expanded Unicode title/capitalize tests covering multiple edge cases.


Lib/test/string_tests.py
```diff
@@ -1134,8 +1134,9 @@ def test_capitalize_nonascii(self):
         self.checkequal('\u2160\u2171\u2172',
                         '\u2170\u2171\u2172', 'capitalize')
         # check with Ll chars with no upper - nothing changes here
-        self.checkequal('\u019b\u1d00\u1d86\u0221\u1fb7',
-                        '\u019b\u1d00\u1d86\u0221\u1fb7', 'capitalize')
+        # This test doesn't exist anymore in Python 3.15 and fails on 3.15
+        # self.checkequal('\u019b\u1d00\u1d86\u0221\u1fb7',
+        #                 '\u019b\u1d00\u1d86\u0221\u1fb7', 'capitalize')
```
ShaharNaveh, line 1195: Can you please revert this change? We are not targeting CPython 3.15 yet:)

ShaharNaveh (changes requested):
Overall LGTM!

ty:)

Lib/test/string_tests.py
joshuamegnauth54: Sure, but CI might fail because `icu4x` targets the latest Unicode version. :sweat_smile: 

(Also I'll have to do it by Monday since I won't be near my computer for the weekend. :face_with_spiral_eyes: )

ShaharNaveh: > Sure, but CI might fail because `icu4x` targets the latest Unicode version. 😅

That's fine, you can mark that test with:
```py
@unittest.expectedFailure # TODO: RUSTPYTHON; <error message>`.
test_capitalize_nonascii(self):
  ...
```

I've updated this test at #7758 I saw that this file had lots of changes since it was last pulled from upstream

> (Also I'll have to do it by Monday since I won't be near my computer for the weekend. 😵‍💫 )

All good, we're greatful for any effort/time you put into it!

joshuamegnauth54: It turns out that the failing test was updated in *3.14* not 3.15. So, your patch that updated the tests (#7758) fixed the issue. :grinning: 

I reverted the test change and force pushed. Everything should work now.

ShaharNaveh:
@joshuamegnauth54 can you please fix the merge conflicts?

joshuamegnauth54:
Fixed and also I fixed a small clippy lint that I introduced. I'm not sure why it didn't trigger earlier, but one of my helpers should have been `pub(crate)` instead of just `pub`.

ShaharNaveh:
> Fixed and also I fixed a small clippy lint that I introduced. I'm not sure why it didn't trigger earlier, but one of my helpers should have been `pub(crate)` instead of just `pub`.

That's because we have added a new lint for that at #7762 :)

ShaharNaveh (approved):
LGTM!
tysm:)

youknowone:
I am sorry, I merged another PR first and it made conflict. Could you resolve conflict? I will merge it immediately

joshuamegnauth54:
Fixed and force pushed. :grin:

ShaharNaveh (approved):
lgtm!
ty:)

PR #7721, opened 2026-04-28
Bytecode parity CFG normalization and jump cleanup

youknowone:
## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed control flow handling in complex try/except/finally blocks with nested conditionals.
  * Corrected global variable resolution within class body scopes.

* **Refactor**
  * Revised constant folding and expression optimization mechanisms.
  * Changed static attribute collection approach during code compilation.

* **Tests**
  * Extended bytecode generation and control-flow behavior test coverage.

### DIFF
--- a/.cspell.json
+++ b/.cspell.json
@@ -67,7 +67,6 @@
     "jitted",
     "jitting",
     "kwonly",
-    "lolcatz",
     "lossily",
     "mcache",
     "oparg",
--- a/.cspell.json
+++ b/.cspell.json
@@ -67,6 +67,7 @@
     "jitted",
     "jitting",
     "kwonly",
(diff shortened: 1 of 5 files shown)


.cspell.json
```diff
@@ -67,6 +67,7 @@
     "jitted",
     "jitting",
     "kwonly",
+    "lolcatz",
```
youknowone: adding vocabulary only for string in test data is bad idea

fanninpm: Maybe leave it as a single ignore statement in one file?

youknowone: The test was duplication of test_dis.py. removed the test

PR #7734, opened 2026-04-29
Short-circuit identity in rich_compare_bool for Eq/Ne (PyObject_RichCompareBool parity)

changjoon-park:
## Background

CPython distinguishes two comparison entry points (`Objects/object.c`):
- `PyObject_RichCompare` — returns the raw `__eq__` / `__ne__` result; no identity short-circuit
- `PyObject_RichCompareBool` — returns `bool`; **identity implies equality** (and inequality is false on identity), short-circuiting before dispatch

Collection membership / equality (`x in [x]`, `[nan] == [nan]`, set/dict comparisons) go through the bool variant and rely on the short-circuit. RustPython's `rich_compare_bool` skipped the identity check, so a buggy or raising `__eq__` propagated even when the operand was the same object.

## Repro

```python
class BadEq:
    def __hash__(self): return 7
    def __eq__(self, other): raise RuntimeError("eq called")

x = BadEq()
d = {x: x}
v1 = next(iter(d.values()))

v1 in d.values()
# CPython 3.14: True
# RustPython:   RuntimeError("eq called")    (before this PR)
```

NaN exhibits the canonical asymmetry that proves the contract:

```python
n = float('nan')
n == n          # False  (RichCompare path, no shortcut)
[n] == [n]      # True   (RichCompareBool path, identity shortcut)
n in [n]        # True
```

## Fix

Add an identity short-circuit at the top of `rich_compare_bool` for `Eq` (returns `true`) and `Ne` (returns `false`). Ordering ops fall through to `_cmp` because Python does not guarantee reflexivity for `<` / `<=` / `>` / `>=` (e.g. user-defined `__lt__` may legitimately return False for `x < x`). `_cmp` itself is untouched, so `==` / `!=` operators continue to invoke `__eq__` / `__ne__` exactly as before.

## Tests unmasked

- `test_dictviews.TestDictViews.test_compare_error`

## Verification

**53 scenarios across 10 categories — byte-identical with CPython 3.14.4 (`diff` empty):**

- 6 ordering ops with raising dunders (`<`, `<=`, `>`, `>=`) — all raise (no shortcut)
- 7 collection membership types (`list`, `tuple`, `set`, `frozenset`, `dict`, `dict.keys`, `deque`)
- 2 dict view types (`values`, `items`)
- 7 container equality (`==` and `!=` for `list` / `tuple` / `set` / `frozenset` / `dict`)
- 6 NaN edge cases (`n == n` False, `[n] == [n]` True, `n in [n]` True)
- 5 `FalseEq` class cases (where `__eq__` returns `False`)
- 3 list operations (`count`, `index`, `remove`)
- 3 set operations (`add`, `union`, `len`)
- 3 dict operations (`get`, `setdefault`)
- 4 hash collision + identity separation cases

**No regressions** across 14 modules (~2,402 tests): `test_dictviews`, `test_dict`, `test_set`, `test_list`, `test_tuple`, `test_collections`, `test_typing`, `test_descr`, `test_iter`, `test_compare`, `test_class`, `test_userdict`, `test_userlist`, `test_ordered_dict`.

## Out-of-scope finding (recorded for follow-up)

`range.__contains__` and `array.__contains__` return `False` on type mismatch (e.g. `RaisingEq() in range(10)` returns `False` without calling `__eq__`). CPython's specialized `range_contains` and `array_contains` fall back to the default sequence contains for non-fast-path types, which would invoke `__eq__`. This divergence is independent of the `rich_compare_bool` change in this PR — those specialized contains slots bypass `bool_eq` entirely.

## Summary by CodeRabbit

* **Bug Fixes**
  * Equality/inequality now short-circuit when both operands are the same object, improving correctness and performance.
  * Mapping and weak proxy comparisons now evaluate rich-comparison results and coerce them to booleans, ensuring consistent semantics and proper error propagation.
  * Weak references retain identity fallback when a referent cannot be resolved.

### DIFF
--- a/Lib/test/test_dictviews.py
+++ b/Lib/test/test_dictviews.py
@@ -292,7 +292,6 @@ class DictSetTest(unittest.TestCase):
         self.assertRaises(TypeError, copy.copy, d.values())
         self.assertRaises(TypeError, copy.copy, d.items())
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
     def test_compare_error(self):
         class Exc(Exception):
             pass
--- a/crates/vm/src/builtins/mappingproxy.rs
+++ b/crates/vm/src/builtins/mappingproxy.rs
@@ -5,7 +5,7 @@ use crate::{
     class::PyClassImpl,
     common::{hash, lock::LazyLock},
     convert::ToPyObject,
-    function::{ArgMapping, OptionalArg, PyComparisonValue},
+    function::{ArgMapping, OptionalArg, PyArithmeticValue, PyComparisonValue},
     object::{Traverse, TraverseFn},
     protocol::{PyMappingMethods, PyNumberMethods, PySequenceMethods},
     types::{
@@ -211,9 +211,12 @@ impl Comparable for PyMappingProxy {
         vm: &VirtualMachine,
     ) -> PyResult<PyComparisonValue> {
         let obj = zelf.to_object(vm)?;
-        Ok(PyComparisonValue::Implemented(
-            obj.rich_compare_bool(other, op, vm)?,
-        ))
+        // CPython parity (Objects/descrobject.c::mappingproxy_richcompare):
+        // delegate to PyObject_RichCompare on the underlying mapping.
+        let res = obj.rich_compare(other.to_owned(), op, vm)?;
+        PyArithmeticValue::from_object(vm, res)
+            .map(|o| o.try_to_bool(vm))
+            .transpose()
     }
 }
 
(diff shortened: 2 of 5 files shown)


ShaharNaveh (approved):
ty:)
lgtm!

changjoon-park:
> ty:) lgtm!

thanks for the quick check !

changjoon-park:
CI was failing on `test_weak_keyed_cascading_deletes`. Turns out the new short-circuit exposed a latent issue in `weakref` / `weakproxy` / `mappingproxy` — they were going through the bool variant for referent compare, but CPython actually uses `PyObject_RichCompare` there so the referent's `__eq__` still runs even on identity match.

Pushed a follow-up that routes those three through `PyObject_RichCompare`. Tested against CPython 3.14 across a bunch of scenarios (Eq/Ne, ordering ops, dead refs, NotImplemented from referent) and they line up now. Failing test passes locally and no regressions across the nearby modules.

youknowone (approved):
👍

PR #7736, opened 2026-04-29
Resolve PathLike to fspath result for FileIO.name

changjoon-park:
## Background

CPython's `fileio_init` (`Modules/_io/fileio.c`) calls `PyUnicode_FSConverter` on the name argument and stores the resolved str/bytes as `self.name` — not the original PathLike object. RustPython's `FileIO.__init__` stored the raw argument, so `open(PathLike(p)).name` returned the PathLike itself instead of the resolved path. This propagated through `bz2.BZ2File` and `lzma.LZMAFile` (both expose underlying `_fp.name` per Lib/bz2.py:128 / Lib/lzma.py).

## Repro

```python
import bz2, pathlib
p = pathlib.Path("/tmp/foo.bz2")
with bz2.BZ2File(p, "wb") as f:
    type(f.name)
# CPython 3.14: <class 'str'>
# RustPython:   <class 'pathlib.PosixPath'>    (before this PR)
```

User code that does `f.name.upper()`, `f.name.endswith('.bz2')`, or `assertEqual(f.name, expected_str)` silently breaks.

## Fix

Before `set_attr("name", ...)` in `FileIO::init`, convert the name via `__fspath__` when it is neither an int (file descriptor) nor str/bytes:

```rust
let name_to_store = if name.fast_isinstance(vm.ctx.types.int_type)
    || name.fast_isinstance(vm.ctx.types.str_type)
    || name.fast_isinstance(vm.ctx.types.bytes_type)
{
    name
} else {
    vm.call_method(&name, "__fspath__", ())?
};
```

The same `__fspath__` conversion already runs upstream (line 5530, `OsPath::try_from_fspath`) to obtain the file descriptor; this aligns the stored attribute with that resolved value. Errors from `__fspath__` cannot be reached at this point because the upstream conversion would have raised them first.

## Tests unmasked

- `test_bz2.BZ2FileTest.testOpenPathLikeFilename`
- `test_lzma.FileTestCase.test_init_with_PathLike_filename`
- `test_lzma.OpenTestCase.test_with_pathlike_filename`

(All three share the same `<FakePath ...> != '...'` TODO marker, located via the cross-module marker grep.)

## Verification

CPython 3.14.4 byte-identical for 7 input shapes:
- `open(str)` → `.name` is str (unchanged)
- `open(bytes)` → `.name` is bytes (unchanged)
- `open(int_fd)` → `.name` is int (unchanged)
- `open(FakePath(str))` → `.name` is str
- `open(FakePath(bytes))` → `.name` is bytes
- `open(pathlib.Path(...))` → `.name` is str
- `bz2.BZ2File(FakePath(...))` → `.name` is str

`gzip.GzipFile` (different code path) was already correct and remains so.

No regressions across `test_bz2` (101), `test_lzma` (121), `test_gzip` (75), `test_io` (667), `test_fileio` (99), `test_tempfile` (118), `test_pathlib` (1369), `test_os` (373).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file I/O handling for path-like inputs by resolving and normalizing the provided name exactly once before use.
  * Ensures the stored file name is consistent across input types, improving opener behavior and producing more accurate error messages.

### DIFF
--- a/crates/vm/src/stdlib/_io.rs
+++ b/crates/vm/src/stdlib/_io.rs
@@ -5322,9 +5322,9 @@ mod fileio {
         VirtualMachine,
         builtins::{PyBaseExceptionRef, PyUtf8Str, PyUtf8StrRef},
         common::wtf8::Wtf8Buf,
-        convert::{IntoPyException, ToPyException},
+        convert::{IntoPyException, ToPyException, ToPyObject},
         exceptions::OSErrorBuilder,
-        function::{ArgBytesLike, ArgMemoryBuffer, OptionalArg, OptionalOption},
+        function::{ArgBytesLike, ArgMemoryBuffer, FsPath, OptionalArg, OptionalOption},
         ospath::{OsPath, OsPathOrFd},
         stdlib::os,
         types::{Constructor, DefaultConstructor, Destructor, Initializer, Representable},
@@ -5499,6 +5499,17 @@ mod fileio {
                 None
             };
 
+            // CPython parity (Modules/_io/fileio.c::fileio_init): when name is
+            // not an fd, resolve it through os.fspath() exactly once. The
+            // resolved str/bytes is reused for the actual open (or the
+            // user-supplied opener) and stored as self.name, matching CPython's
+            // single PyUnicode_FSConverter call.
+            let name = if arg_fd.is_some() {
+                name
+            } else {
+                FsPath::try_from_path_like(name, true, vm)?.to_pyobject(vm)
+            };
+
             let mode_obj = args
                 .mode
                 .unwrap_or_else(|| PyUtf8Str::from("rb").into_ref(&vm.ctx));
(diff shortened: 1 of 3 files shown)


ShaharNaveh (approved):
thanks for looking into it!

crates/vm/src/stdlib/_io.rs
```diff
@@ -5499,6 +5499,17 @@ mod fileio {
                 None
             };
 
+            // CPython parity (Modules/_io/fileio.c::fileio_init): when name is
+            // not an fd, resolve it through os.fspath() exactly once. The
+            // resolved str/bytes is reused for the actual open (or the
+            // user-supplied opener) and stored as self.name, matching CPython's
+            // single PyUnicode_FSConverter call.
+            let name = if arg_fd.is_some() {
```
ShaharNaveh, line 5507: Can you please use a `match` or `.map_or_else(...)` instead?

changjoon-park:
hi @ShaharNaveh thanks for the point out

i re-checked against cpython 3.14, this fixes the wrong layer. the _io.FileIO already keeps the PathLike on .name like CPython does, the real gap is io_open() not calling os.fspath(). my change makes direct _io.FileIO(PathLike) calls diverge from CPython. 

so i'll close this and open a follow-up against io_open() instead. sorry for the noise.

ShaharNaveh:
> sorry for the noise.

All good lmao, keep up the noise<3

PR #7636, opened 2026-04-20
Use `cfg_select!` macro instead of `cfg_if!`

ShaharNaveh:
Since 1.95 `cfg_select!` macro is stabilized

https://doc.rust-lang.org/stable/std/macro.cfg_select.html

## Summary by CodeRabbit

* **New Requirements**
  * Minimum supported Rust version raised to 1.95.0.

* **Chores**
  * Removed an internal build-time dependency and modernized compile-time condition handling across the codebase.

* **Clarification**
  * No changes to public APIs, runtime behavior, or platform-identifying constants; no user-facing functionality was altered.

PR #7740, opened 2026-04-30
Use cfg_select in a bunch more places

coolreader18:
A sort of follow-up to #7636. I tried to not overdo it here, and I think most of the places I chose are genuinely easier to understand than before.

## Summary by CodeRabbit

* **Refactor**
  * Unified platform- and feature-specific branching across the codebase, consolidating conditional compilation into a single selection mechanism. Changes touch many subsystems (stdlib, stdlib modules, VM, I/O, crypto, signals, time, versioning) with no user-facing behavior changes—maintenance and consistency improvements only.


ShaharNaveh (approved):
I really like this change

PR #7745, opened 2026-04-30
Match CPython's _float_div_mod, fixing divmod and `%` zero-handling

### LINKED ISSUE #7722: `divmod()` returns wrong float result when the remainder is zero
## Summary

`divmod(0, -1.0)` returns `(-1.0, -1.0)` — the quotient and remainder are both off by one. The float `divmod` implementation enters its sign-correction branch even when the remainder is exactly zero, so it adds the divisor to a zero remainder and decrements the quotient, producing values that are arithmetically wrong (`0 == -1.0 * -1.0 + -1.0` is false).

## Expected

```
(-0.0, -0.0)
```

`0 / -1.0` is exactly `-0.0` with no remainder, so `divmod` should return a quotient and remainder that satisfy the identity `a == b * q + r` with `r == 0`.

## Actual

```python
>>> divmod(0, -1.0)
(-1.0, -1.0)
```

The returned values do not satisfy the divmod identity: `-1.0 * -1.0 + -1.0 == 0.0`, which only holds because the quotient and remainder are both wrong by the same amount. The true quotient of `0 / -1.0` is zero, and the true remainder is zero.

## Root Cause

[`crates/common/src/float_ops.rs#L91-L103`](https://github.com/RustPython/RustPython/blob/3f718f9942147811f49a82a1b6c60ca2a86cec85/crates/common/src/float_ops.rs#L91-L103) enters its sign-correction branch whenever `v2.is_sign_negative() != m.is_sign_negative()`, without checking whether the remainder is zero. For `divmod(0.0, -1.0)`, Rust's `%` returns `+0.0` while `v2` is negative, so the branch fires on a zero remainder: it adds the divisor (`m += v2`, giving `-1.0`) and decrements the quotient (`d -= 1.0`, giving `-1.0`) — yielding `(-1.0, -1.0)` instead of `(0.0, 0.0)`.

The function is also missing a quotient-snap step. When the floor of the quotient is zero, the sign of the resulting signed zero should match the sign of the true quotient `v1 / v2`, but the current code just propagates whatever signed zero `(v1 - m) / v2` happens to produce.

## Fix

Guard the sign-correction branch with `m != 0.0`; on the zero-remainder path, set `m = copysign(0.0, v2)`. After the remainder is finalized, snap the quotient to `floor(d)` when non-zero and to `copysign(0.0, v1 / v2)` when zero.

This same function is also responsible for the related signed-zero defect on `divmod(-1.0, -2.0)` returning `(-0.0, -1.0)` instead of `(0.0, -1.0)` — only the quotient-snap branch is missing in that case, while the report above also exercises the zero-remainder branch.

## Python Documentation

[`divmod` reference](https://docs.python.org/3/library/functions.html#divmod): "the quotient and remainder when using integer division ... For floating-point numbers the result is `(q, a % b)`, where `q` is usually `math.floor(a / b)` ... In any case `q * b + r` is very close to `a`, if `r` is non-zero it has the same sign as `b`, and `0 <= abs(r) < abs(b)`."

CPython's `_float_div_mod` ([`Objects/floatobject.c#L620-L656`](https://github.com/python/cpython/blob/v3.14.3/Objects/floatobject.c#L620-L656)) demonstrates the correct structure for reference:

```c
// When the remainder is non-zero, force its sign to match the divisor.
// When it is zero, fall back to copysign — this preserves IEEE-754 signed-zero
// semantics consistently across platforms (fmod's zero sign is platform-defined).
if (*mod) {
    if ((wx < 0) != (*mod < 0)) {
        *mod += wx;
        div -= 1.0;
    }
}
else {
    *mod = copysign(0.0, wx);
}
// Snap the quotient: round toward floor when non-zero;
// when zero, take the sign of the true quotient vx / wx.
if (div) {
    *floordiv = floor(div);
    if (div - *floordiv > 0.5) {
        *floordiv += 1.0;
    }
}
else {
    *floordiv = copysign(0.0, vx / wx);
}
```

## Reproduction

```python
print(divmod(0, -1.0))
```

### Environment

- OS: Debian 12

changjoon-park:
## Background

`crates/common/src/float_ops.rs` had three independent implementations:

- `mod_(v1, v2)` — Python `%` operator
- `floordiv(v1, v2)` — Python `//` operator
- `divmod(v1, v2)` — Python `divmod()` builtin

Each independently converted Rust's dividend-sign `%` to CPython's divisor-sign convention. Two of them (`mod_` and `divmod`) had the same gap — the sign-correction branch fired on zero remainders, producing values that violate Python's spec invariants.

### 1. Zero-remainder + opposite-sign divisor (`divmod` and `%`)

```python
divmod(0.0, -1.0)
# CPython:    (-0.0, -0.0)
# Pre-fix:    (-1.0, -1.0)   ← q and r both off by one
6.0 % -3.0
# CPython:    -0.0
# Pre-fix:    -3.0           ← |r| == |b|, violates 0 <= |r| < |b|
divmod(6.0, -3.0)
# CPython:    (-2.0, -0.0)
# Pre-fix:    (-3.0, -3.0)
```

The dividend-is-non-zero-exact-multiple cases fell into the default sign-correction branch, which adds the divisor to a zero remainder and decrements the quotient — yielding values that violate the magnitude invariant `0 <= |r| < |b|`.

### 2. Signed-zero quotient leak (`divmod`)

```python
divmod(-1.0, -2.0)
# CPython:    (+0.0, -1.0)
# Pre-fix:    (-0.0, -1.0)   ← quotient sign leaked from (v1 - m) / v2
```

When the true quotient is exactly zero, the `(v1 - m) / v2` intermediate produces a signed zero whose sign reflects the intermediate division, not the true quotient `v1 / v2`. CPython's `_float_div_mod` snaps zero quotients via `copysign(0.0, vx / wx)`.

### 3. Internal consistency violation (Python language spec)

Python guarantees `divmod(a, b) == (a // b, a % b)`. RustPython didn't:

```python
a, b = 6.0, -3.0
divmod(a, b)               # (-3.0, -3.0)   ← divmod's own bug
(a // b, a % b)            # (-2.0, -3.0)   ← floordiv correct, mod_ buggy
```

Three values, three different paths, three results that all disagreed.

## Fix

Mirror CPython's `_float_div_mod` (`Objects/floatobject.c`) by making `divmod` the canonical implementation. `mod_` and `floordiv` become thin wrappers that delegate and project, so `divmod(a, b) == (a // b, a % b)` holds **by construction** — no separate paths to keep in sync.

```rust
pub fn mod_(v1: f64, v2: f64) -> Option<f64> {
    divmod(v1, v2).map(|(_, m)| m)
}

pub fn floordiv(v1: f64, v2: f64) -> Option<f64> {
    divmod(v1, v2).map(|(d, _)| d)
}

pub fn divmod(v1: f64, v2: f64) -> Option<(f64, f64)> {
    if v2 == 0.0 {
        return None;
    }
    let mut m = v1 % v2;
    let mut d = (v1 - m) / v2;
    if m != 0.0 {
        if v2.is_sign_negative() != m.is_sign_negative() {
            m += v2;
            d -= 1.0;
        }
    } else {
        m = (0.0_f64).copysign(v2);     // zero remainder: sign of divisor
    }
    let d = if d != 0.0 {
        let f = d.floor();
        if d - f > 0.5 { f + 1.0 } else { f }   // CPython's half-up snap
    } else {
        (0.0_f64).copysign(v1 / v2)     // zero quotient: sign of true quotient
    };
    Some((d, m))
}
```

The half-up snap on non-zero quotient (`if d - f > 0.5 { f + 1.0 } else { f }`) mirrors CPython's safety net for cases where `(v1 - m) / v2` undershoots the true integer quotient by more than half an ULP.

## Verification

CPython 3.14.4 byte-identical for **44 cases** spanning sign permutations, zero edges, exact divisions, IEEE 754 signed zero, and subnormals:

- `divmod`: 29 cases (all pass; pre-fix 8 failed)
- `%` operator (`mod_`): 8 cases (all pass; pre-fix 5 failed)
- `//` operator (`floordiv`): 7 cases (all pass on both)
- Internal consistency `divmod == (//, %)`: 6 cases (all pass; pre-fix all 6 failed)

Regression sweep: `test_float`, `test_int`, `test_complex`, `test_math`, `test_time`, `test_datetime`, `test_fractions`, `test_decimal` — **1,471 tests, 0 regressions**.

A regression test was added to `extra_tests/snippets/builtin_divmod.py` covering the bug class plus the divmod identity and magnitude/sign invariants.

## Notes

`divmod` carries the canonical implementation (rather than introducing a private `_float_div_mod` helper) because all three public functions in this module already share the same `Option<f64>`-based API; a thin-wrapper pattern is the smallest delta that achieves single-source-of-truth without changing the public surface.

The half-up snap was added to match CPython exactly even though no probed case currently exercises it. It is correctness-relevant for extreme magnitudes where the intermediate `(v1 - m) / v2` can lose precision.

Closes #7722

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved floating-point division operations to match Python's standard library behavior more closely, with better handling of edge cases involving signed zeros and operands with opposite signs.

* **Tests**
  * Added comprehensive test coverage for floating-point division edge cases and invariants.

### DIFF
--- a/crates/common/src/float_ops.rs
+++ b/crates/common/src/float_ops.rs
@@ -68,38 +68,42 @@ pub const fn div(v1: f64, v2: f64) -> Option<f64> {
 }
 
 pub fn mod_(v1: f64, v2: f64) -> Option<f64> {
-    if v2 != 0.0 {
-        let val = v1 % v2;
-        match (v1.signum() as i32, v2.signum() as i32) {
-            (1, 1) | (-1, -1) => Some(val),
-            _ if (v1 == 0.0) || (v1.abs() == v2.abs()) => Some(val.copysign(v2)),
-            _ => Some((val + v2).copysign(v2)),
-        }
-    } else {
-        None
-    }
+    divmod(v1, v2).map(|(_, m)| m)
 }
 
 pub fn floordiv(v1: f64, v2: f64) -> Option<f64> {
-    if v2 != 0.0 {
-        Some((v1 / v2).floor())
-    } else {
-        None
-    }
+    divmod(v1, v2).map(|(d, _)| d)
 }
 
+// Canonical (floordiv, mod) for floats matching CPython's _float_div_mod
+// (Objects/floatobject.c). `mod_` and `floordiv` delegate here so that
+// `divmod(a, b) == (a // b, a % b)` holds by construction.
 pub fn divmod(v1: f64, v2: f64) -> Option<(f64, f64)> {
-    if v2 != 0.0 {
-        let mut m = v1 % v2;
-        let mut d = (v1 - m) / v2;
+    if v2 == 0.0 {
+        return None;
+    }
+    let mut m = v1 % v2;
+    let mut d = (v1 - m) / v2;
+    if m != 0.0 {
+        // Non-zero remainder must have the sign of the divisor.
         if v2.is_sign_negative() != m.is_sign_negative() {
             m += v2;
             d -= 1.0;
         }
-        Some((d, m))
     } else {
-        None
+        // Zero remainder: sign matches divisor (IEEE 754 / CPython contract).
+        m = (0.0_f64).copysign(v2);
     }
+    let d = if d != 0.0 {
+        let f = d.floor();
+        // Snap up if (v1 - m) / v2 undershot the true integer quotient by
+        // more than half an ULP (mirrors CPython's `if (div - *floordiv > 0.5)`).
+        if d - f > 0.5 { f + 1.0 } else { f }
+    } else {
+        // Zero quotient: take the sign of the true quotient v1 / v2.
+        (0.0_f64).copysign(v1 / v2)
+    };
+    Some((d, m))
 }
 
 // nextafter algorithm based off of https://gitlab.com/bronsonbdevost/next_afterf
--- a/extra_tests/snippets/builtin_divmod.py
+++ b/extra_tests/snippets/builtin_divmod.py
@@ -1,3 +1,5 @@
+import math
+
 from testutils import assert_raises
 
 assert divmod(11, 3) == (3, 2)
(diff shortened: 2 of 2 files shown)


ShaharNaveh (approved):
looks great!
tysm:)

changjoon-park:
> looks great! tysm:)

thanks for the review !

changjoon-park:
> looks great! tysm:)

thanks !

youknowone (approved):

PR #7746, opened 2026-04-30
Raise OverflowError in float pow when finite-base result overflows

### LINKED ISSUE #7729: `pow(2.0, 2000)` returns `inf` instead of raising `OverflowError`
## Feature

`pow(2.0, 2000)` (and any positive-base × large-exponent `float` exponentiation that overflows IEEE 754 double range) silently returns `inf` in RustPython, but CPython raises `OverflowError`.

### Reproduction

```python
print(pow(2.0, 2000))
```

**RustPython:**

```
inf
```

**CPython:**

```
OverflowError: (34, 'Numerical result out of range')
```

### Root cause

[`float_pow`](https://github.com/RustPython/RustPython/blob/3f718f9942147811f49a82a1b6c60ca2a86cec85/crates/vm/src/builtins/float.rs#L157-L168)'s `else` branch (positive base or integer exponent) returns `v1.powf(v2)` directly without inspecting the result for overflow, so `f64::INFINITY` leaks out as a `float` value. CPython's [`float_pow`](https://github.com/python/cpython/blob/v3.14.3/Objects/floatobject.c#L794-L807) calls `_Py_ADJUST_ERANGE1(ix)` after the libm `pow(iv, iw)` call, which sets `errno = ERANGE` when the result is `±HUGE_VAL`, and then `PyErr_SetFromErrno(PyExc_OverflowError)` raises an `OverflowError` whose message is `strerror(ERANGE)`.

The same overflow guard already exists in the sibling [`complex.rs::inner_pow`](https://github.com/RustPython/RustPython/blob/3f718f9942147811f49a82a1b6c60ca2a86cec85/crates/vm/src/builtins/complex.rs#L153-L171) — `if ans.is_infinite() && !(v1.is_infinite() || v2.is_infinite())` raises `OverflowError("complex exponentiation overflow")` — so RustPython's `complex(2.0) ** 2000` correctly raises while `pow(2.0, 2000)` does not. PR #7727 adds the same guard to `float_pow`'s negative-base × non-integer-exponent branch (delegating to `complex_pow`), but the `else` branch covered here is not addressed by that PR.

### Fix

In `float_pow`'s `else` branch, compute `let ans = v1.powf(v2)` and raise `OverflowError` when `ans.is_infinite() && !(v1.is_infinite() || v2.is_infinite())`. Mirror the inline-overflow pattern already used in `complex.rs::inner_pow`; do not delegate to a shared helper, since CPython itself handles this inline (no shared helper exists in either codebase, and message strings differ per call site — `"complex exponentiation"`, `"math range error"`, `strerror(ERANGE)`).

### Environment

- CPython v3.14.3
- OS: Debian 12

- CPython `float_pow`: [`Objects/floatobject.c#L794-L807`](https://github.com/python/cpython/blob/v3.14.3/Objects/floatobject.c#L794-L807) — `_Py_ADJUST_ERANGE1` forces `errno = ERANGE` when libm `pow` returns `±HUGE_VAL`; `PyErr_SetFromErrno(PyExc_OverflowError)` then raises:

  ```c
  // After the libm pow() call, if the result is ±HUGE_VAL, force errno = ERANGE
  // and raise OverflowError with the libc strerror(ERANGE) message.
  errno = 0;
  ix = pow(iv, iw);
  _Py_ADJUST_ERANGE1(ix);
  if (negate_result)
      ix = -ix;

  if (errno != 0) {
      PyErr_SetFromErrno(errno == ERANGE ? PyExc_OverflowError
                                         : PyExc_ValueError);
      return NULL;
  }
  return PyFloat_FromDouble(ix);
  ```

- CPython `_Py_ADJUST_ERANGE1` macro: [`Include/internal/pycore_pymath.h#L33-L42`](https://github.com/python/cpython/blob/v3.14.3/Include/internal/pycore_pymath.h#L33-L42) — sets `errno = ERANGE` when `x == ±HUGE_VAL`, regardless of whether libm itself set `errno`:

  ```c
  // Some libm implementations don't reliably set errno on overflow; this macro
  // sets ERANGE explicitly when the result is ±HUGE_VAL.
  static inline void _Py_ADJUST_ERANGE1(double x)
  {
      if (Py_IS_INFINITY(x)) {
          if (errno == 0)
              errno = ERANGE;
      }
      else if (errno == ERANGE && x == 0.0) {
          errno = 0;
      }
  }
  ```

- Documentation: [`builtins.pow`](https://docs.python.org/3.14/library/functions.html#pow)

changjoon-park:
## Background

`float_pow` (`crates/vm/src/builtins/float.rs`) has three branches:

1. `0 ** negative` → `ZeroDivisionError`
2. `negative ** non-integer` → delegates to `complex.rs::complex_pow` (which has its own overflow guard, added in #7727)
3. **else** — positive base, or negative base × integer exponent → `v1.powf(v2)` returned directly

Branch 3 is the path covering the most common cases (`2.0 ** 2000`, `10.0 ** 400`, etc.). The result of `f64::powf` was returned without inspecting it, so finite inputs that produced an out-of-range value silently leaked `f64::INFINITY` instead of raising `OverflowError`.

## Reproduction

```python
pow(2.0, 2000)
# CPython:    OverflowError: (34, 'Result too large')
# Pre-fix:    inf

pow(-2.0, 2001)
# CPython:    OverflowError
# Pre-fix:    -inf

pow(0.5, -2000)         # = 1 / (0.5 ** 2000) = 2 ** 2000
# CPython:    OverflowError
# Pre-fix:    inf

pow(1e150, 3)
# CPython:    OverflowError
# Pre-fix:    inf
```

7 of 15 probed inputs failed pre-fix. The intentional-infinity cases (`inf**2`, `2**inf`, `nan**2`) and underflow case (`2**(-2000)` → `0.0`) were already correct.

## Fix

Mirror `complex_pow`'s inline guard pattern (`crates/vm/src/builtins/complex.rs:169-174`):

```rust
} else {
    let ans = v1.powf(v2);
    if ans.is_infinite() && !(v1.is_infinite() || v2.is_infinite()) {
        Err(vm.new_overflow_error("math range error"))
    } else {
        Ok(ans.to_pyobject(vm))
    }
}
```

The two-condition guard distinguishes overflow from intentional infinity:

| `v1` finite | `v2` finite | `ans` inf | Action |
|---|---|---|---|
| ✓ | ✓ | ✓ | `OverflowError` (overflow) |
| ✓ | ✓ | ✗ | normal value |
| inf | ✓ | ✓ | inf (intentional, e.g. `inf ** 2`) |
| ✓ | inf | ✓ | inf (intentional, e.g. `2 ** inf`) |

This matches CPython's `_Py_ADJUST_ERANGE1` macro (`Include/internal/pycore_pymath.h`), which forces `errno = ERANGE` when the result is `±HUGE_VAL`, regardless of whether libm itself set it.

## Note on the error message

CPython's `float_pow` reports `OverflowError: (34, strerror(ERANGE))`, where the message is platform-dependent (`"Result too large"` on macOS, `"Numerical result out of range"` on Linux). RustPython uses `"math range error"` here — the same text already used at `crates/stdlib/src/math.rs:938` for libm `ERANGE`, which CPython's `math.pow` and `math.exp` also raise. The text is platform-stable and consistent with project precedent for math-overflow paths.

## Verification

CPython 3.14.4 byte-identical for **15 cases** spanning:

- Finite-overflow positive base (`2.0 ** 2000`, `10.0 ** 400`, `1.5 ** 10000`, `1e150 ** 3`, `0.5 ** -2000`)
- Finite-overflow negative^integer (`-2.0 ** 2000`, `-2.0 ** 2001`)
- Negative^non-integer (delegates to `complex_pow`, already fixed in #7727)
- No-overflow case (`1e150 ** 2 = 9.99e+299`)
- Underflow (`2.0 ** -2000 = 0.0`)
- Identity (`1.0 ** 1e308 = 1.0`)
- Intentional infinities (`inf ** 2`, `2 ** inf`, `inf ** inf`)
- NaN propagation (`nan ** 2`, `2 ** nan`)

Routing also verified: `pow(int, float)` (e.g. `2 ** 2000.0`) and the int negative-exp path delegate to `float_pow` and now raise correctly. Pure int^int (`pow(2, 2000)` → 603-digit int) is unaffected.

Regression sweep: `test_float`, `test_int`, `test_complex`, `test_math`, `test_decimal` — **809 tests, 0 regressions**.

Regression cases were appended to `extra_tests/snippets/builtin_pow.py` covering the bug class plus the must-not-raise paths (intentional infinities, underflow, identity, NaN).

## Note

Follows up on #7727, which added the same overflow guard pattern to the negative-base × non-integer-exponent branch. This PR completes the coverage by adding it to the remaining else branch. The two PRs together close the float-pow silent-overflow gap entirely.

Closes #7729

## Release Notes

* **Bug Fixes**
  * Fixed float exponentiation to raise OverflowError when computation results exceed limits, instead of returning infinity.

* **Tests**
  * Enhanced test coverage for float exponentiation edge cases including overflow, underflow, and special numeric values.

### DIFF
--- a/crates/vm/src/builtins/float.rs
+++ b/crates/vm/src/builtins/float.rs
@@ -163,7 +163,12 @@ pub fn float_pow(v1: f64, v2: f64, vm: &VirtualMachine) -> PyResult {
         let v2 = Complex64::new(v2, 0.);
         Ok(super::complex::complex_pow(v1, v2, vm)?.to_pyobject(vm))
     } else {
-        Ok(v1.powf(v2).to_pyobject(vm))
+        let ans = v1.powf(v2);
+        if ans.is_infinite() && !(v1.is_infinite() || v2.is_infinite()) {
+            Err(vm.new_overflow_error("math range error"))
+        } else {
+            Ok(ans.to_pyobject(vm))
+        }
     }
 }
 
--- a/extra_tests/snippets/builtin_pow.py
+++ b/extra_tests/snippets/builtin_pow.py
@@ -26,6 +26,36 @@ assert_raises(OverflowError, pow, -2, 2000.5)
 assert_raises(OverflowError, pow, -2.0, 2000.5)
 assert_raises(OverflowError, complex(-2).__pow__, complex(2000.5))
 
+# float pow overflow (positive base or negative^integer) — finite inputs
+# producing a result outside f64 range must raise OverflowError, not silently
+# return inf.
+assert_raises(OverflowError, pow, 2.0, 2000)
+assert_raises(OverflowError, pow, 2.0, 2000.0)
+assert_raises(OverflowError, pow, 10.0, 400)
+assert_raises(OverflowError, pow, 1.5, 10000)
+assert_raises(OverflowError, pow, -2.0, 2000)
+assert_raises(OverflowError, pow, -2.0, 2001)
+assert_raises(OverflowError, pow, 1e150, 3)
+assert_raises(OverflowError, pow, 0.5, -2000)
+
+# Intentional infinities — when an input is already inf, the result inf is
+# expected and must NOT raise OverflowError.
+import math
+
+assert pow(math.inf, 2.0) == math.inf
+assert pow(2.0, math.inf) == math.inf
+assert pow(math.inf, math.inf) == math.inf
+
+# Underflow yields exact zero (not OverflowError).
+assert pow(2.0, -2000) == 0.0
+
+# pow(1.0, anything) == 1.0 — must not trigger overflow path.
+assert pow(1.0, 1e308) == 1.0
+
+# NaN propagates without raising.
+assert math.isnan(pow(math.nan, 2.0))
+assert math.isnan(pow(2.0, math.nan))
+
 assert_raises(TypeError, pow, 1, None, 0)
 assert_raises(TypeError, pow, True, 1.5, False)
 


ShaharNaveh (approved):
Thank you for finding and fixing these!

changjoon-park:
> Thank you for finding and fixing these!

thanks for this quick review !!

youknowone (approved):

PR #7749, opened 2026-05-01
Fix function attribute tests: __defaults__ del, __code__ free vars ch…

ever0de:
…eck, bound method dir()

- Support `del func.__defaults__` and `del func.__kwdefaults__` by changing setter signatures to `PySetterValue<Option<...>>` so Delete is treated as setting the value to None (matching CPython behaviour)
- Validate free-variable count when assigning to `func.__code__`: raise ValueError when the new code object's freevars count doesn't match the existing closure size
- Add `__dir__` to `PyBoundMethod` that delegates to the underlying function's dir(), so attributes set on the function show up in `dir(bound_method)`
- Remove `@unittest.expectedFailure` from the four tests that now pass: test_blank_func_defaults, test_func_default_args, test___code__, test_dir_includes_correct_attrs

## Summary by CodeRabbit

* **New Features**
  * Bound methods now provide an expanded, de-duplicated __dir__ for improved introspection.

* **Enhancements**
  * Replacing a function’s code validates closure-variable counts and errors on mismatch.
  * Replacing code clears compiled caches and ensures the operation completes successfully.
  * Assigning or deleting a function’s __defaults__ / __kwdefaults__ supports explicit deletion to clear stored values.

### DIFF
--- a/Lib/test/test_funcattrs.py
+++ b/Lib/test/test_funcattrs.py
@@ -50,7 +50,6 @@ class FunctionPropertiesTest(FuncAttrsTest):
     def test_module(self):
         self.assertEqual(self.b.__module__, __name__)
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
     def test_dir_includes_correct_attrs(self):
         self.b.known_attr = 7
         self.assertIn('known_attr', dir(self.b),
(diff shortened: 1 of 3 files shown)


ShaharNaveh (approved):
🚀

youknowone (approved):

PR #7761, opened 2026-05-02
Round float at the decimal level to match CPython's _Py_dg_dtoa

changjoon-park:
## Summary

CPython's `float.__round__` (`Objects/floatobject.c`) routes through `_Py_dg_dtoa` and rounds at the **decimal level**. The previous `round_float_digits` multiplied by `10**ndigits` and rounded at the IEEE 754 binary level, which diverges for values that aren't exactly representable.

```python
>>> round(2.675, 2)
# CPython 3.14.4: 2.67
# RustPython main: 2.68 ❌
```

`2.675` stores as `2.67499...`; dtoa correctly rounds it down to `2.67`, but `(2.675 * 100.0).round() / 100.0` lands on `2.68` because the multiplication produces a phantom `267.5` tie that round-half-to-even snaps up.

Rust's `{:.*}` float formatting uses dtoa-style algorithms (Grisu3 + Dragon4 fallback) and matches CPython's `_Py_dg_dtoa` byte-for-byte. Replacing the multiply-then-round step with `format!("{:.*}", ndigits, x).parse()` removes the entire phantom-tie class for `ndigits >= 0`.

For `ndigits < 0` the existing divide-then-round path is preserved because dividing typical inputs by `10**|ndigits|` produces genuine half-integer ties rather than synthesizing them.

| Probe set | Cases | Match |
|---|---|---|
| Targeted half-tie probes (positive ndigits) | 34 | 34/34 |
| Negative ndigits regression | 17 | 17/17 |
| Special values (NaN, ±Inf, ±0.0, NDIGITS bounds) | 12 | 12/12 |
| No-arg `round(x)` (different code path) | 10 | 10/10 |
| Subnormal / denormal | 6 | 6/6 |
| Random fuzz (mantissa × ndigits ∈ {-5..50}) | 4332 | 4332/4332 |

## Tests unmasked

`Lib/test/test_float.py::RoundTestCase`:
- `test_matches_float_format` — rounds against `float(format(x, '.{n}f'))` across many values; this is exactly what the fix does internally.
- `test_previous_round_bugs` — explicit "round-half-even" cases (e.g. `round(25.0, -1) == 20.0`).

## Test runs

```
cargo run --release -- -m test test_float test_builtin test_decimal \
    test_fractions test_complex test_math test_format test_long test_int \
    test_statistics test_re
```

1573 tests pass, 0 regressions. All 188 `extra_tests/snippets/*.py` pass under the CI feature set (`stdlib,importlib,stdio,encodings,sqlite,ssl-rustls,host_env`).

A regression block is added to `extra_tests/snippets/builtin_round.py`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rounding to more closely match CPython’s behavior: better handling of halfway ties (banker’s rounding), negative ndigits, signed zeros, and non-finite inputs.

* **Tests**
  * Expanded rounding tests to cover banker's rounding, negative ndigits, exact-half cases, NaN/infinity behavior, signed zero, and out-of-range ndigits.

### DIFF
--- a/crates/common/src/float_ops.rs
+++ b/crates/common/src/float_ops.rs
@@ -1,6 +1,6 @@
 use core::f64;
 use malachite_bigint::{BigInt, ToBigInt};
-use num_traits::{Float, Signed, ToPrimitive, Zero};
+use num_traits::{Signed, ToPrimitive};
 
 pub const fn decompose_float(value: f64) -> (f64, i32) {
     if 0.0 == value {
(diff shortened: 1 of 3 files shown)


crates/common/src/float_ops.rs
```diff
@@ -208,64 +208,50 @@ pub fn ulp(x: f64) -> f64 {
...
+        let s = format!("{:.*}", ndigits as usize, x);
+        s.parse().ok()?
+    } else {
+        // ndigits < 0: divide-then-round avoids the phantom-tie problem
+        // because dividing typical inputs by 10**|ndigits| produces genuine
+        // half-integer ties rather than synthesizing them.
+        let pow1 = 10.0f64.powi(-ndigits);
+        let y = x / pow1;
+        let z = y.round();
+        #[allow(clippy::float_cmp)]
```
ShaharNaveh:
```suggestion
        #[expect(clippy::float_cmp, reason = "...")]
```

Can you please change it to `expect` and add a reason to this cfg?

ShaharNaveh (changes requested):
overall lgtm!
I had a minor nitpick

tysm for the patches:)

ShaharNaveh (approved):
tysm:)

crates/common/src/float_ops.rs
changjoon-park: good catch i'll keep this convention following prs too thanks !

changjoon-park:
> tysm:)

thanks for the quick review !!

youknowone (approved):

PR #7765, opened 2026-05-02
Match CPython wording for compile() optimize and flags validation

changjoon-park:
## Summary

Two CPython parity gaps in `compile()` argument validation, both in `crates/vm/src/stdlib/builtins.rs`.

### 1. `optimize` validation gap (real bug)

CPython accepts `optimize ∈ {-1, 0, 1, 2}` only; anything else raises `ValueError("compile(): invalid optimize value")` (`Python/bltinmodule.c::builtin_compile_impl`).

The previous logic only validated through `i32::try_into::<u8>()`, which silently accepts every value in `[0, 255]`. So `optimize=3`, `optimize=99`, etc. were silently truncated to a u8 instead of being rejected.

```python
>>> compile('x = 1', '<x>', 'exec', optimize=3)
# CPython 3.14.4: ValueError: compile(): invalid optimize value
# RustPython main: <code object ...>  ❌
```

### 2. `flags` error wording (wording-only)

CPython uses British "unrecognised" with a colon separator: `compile(): unrecognised flags`. The previous wording was American "unrecognized" without the colon.

## Fix

Replace the cast-based optimize validation with a range pattern, and align the flags wording with CPython's.

```rust
let optimize: u8 = match optimize {
    -1 => vm.state.config.settings.optimize,
    0..=2 => optimize as u8,
    _ => return Err(vm.new_value_error("compile(): invalid optimize value")),
};
```

The `optimize=1 << 1000` boundary still raises `OverflowError` via the `ArgPrimitiveIndex<i32>` conversion layer before this check (preserves `Lib/test/test_compile.py:668`).

| Probe | Cases | Match |
|---|---|---|
| optimize boundary (-1, 0..2, 3, 4, 99, 255, 256, 1000, -2, -99) | 12 | 12/12 |
| flags invalid bits (99999, 0x10000) | 2 | 2/2 |
| flags valid bits (0x20000, 0x40000) | 2 | 2/2 |

`cargo run --release -- -m test test_compile test_builtin test_compileall test_codeop test_ast` — 703 tests pass, 0 regressions. All 189 `extra_tests/snippets/*.py` pass under the CI feature set. A new `builtin_compile.py` adds regression coverage for both wordings.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced the compile() function's parameter validation with clearer error messages for invalid optimize values and unrecognized compile flags.

* **Tests**
  * Added comprehensive test coverage for the compile() function, validating accepted parameter values and ensuring proper error handling across different scenarios.

### DIFF
--- a/crates/vm/src/stdlib/builtins.rs
+++ b/crates/vm/src/stdlib/builtins.rs
@@ -268,12 +268,10 @@ mod builtins {
             let mode_str = args.mode.as_str();
 
             let optimize: i32 = args.optimize.map_or(-1, |v| v.value);
-            let optimize: u8 = if optimize == -1 {
-                vm.state.config.settings.optimize
-            } else {
-                optimize
-                    .try_into()
-                    .map_err(|_| vm.new_value_error("compile() optimize value invalid"))?
+            let optimize: u8 = match optimize {
+                -1 => vm.state.config.settings.optimize,
+                0..=2 => optimize as u8,
+                _ => return Err(vm.new_value_error("compile(): invalid optimize value")),
             };
 
             if args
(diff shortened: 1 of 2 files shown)


ShaharNaveh (approved):
Looks great!
tysm:)

youknowone (approved):

PR #7766, opened 2026-05-02
test(str): add regression test for isprintable() unicode 15 chars,

### LINKED ISSUE #7525: Disagreement with CPython for str.isprintable

tested on current main: 1a9b10ece50ebdbdd3820dafa8f6a630d87be9c0

There are many examples and I picked one.

```py
# the following is True in CPython and False in RustPython
print(
    "\u0b55".isprintable() == True,
)
```

ever0de:
Fixes #7525

This PR adds a regression test for #7525.
At the time the issue was reported (commit 1a9b10e), str.isprintable() was returning False for '\u0b55' while CPython returns True.
The root cause was that rustpython_literal::char::is_printable() was using unic-ucd-category 0.9.0, which is based on Unicode 10.0 data. U+0B55 was unassigned (Cn) in Unicode 10.0, so it was incorrectly treated as non-printable.
This has since been fixed by migrating to icu_properties 2.2.0, which uses up-to-date Unicode data and correctly classifies U+0B55 as Mn (Nonspacing Mark), making it printable.

```
RustPython on  HEAD (1a9b10e) [$?] is 📦 v0.5.0 via 🦀 v1.95.0
❯ cargo run -- -c "print('\u0b55'.isprintable())"

False

RustPython on  main (c2910c0) [$?⇡] is 📦 v0.5.0 via 🐍 v3.11.6 via 🦀 v1.95.0
❯ cargo run -- -c "print('\u0b55'.isprintable())"

True
```

## Summary by CodeRabbit

* **Tests**
  * Enhanced test coverage for Unicode character validation and classification.
  * Added comprehensive tests to ensure accurate handling of printable and non-printable Unicode characters across updated Unicode standards.

### DIFF
--- a/crates/vm/src/builtins/str.rs
+++ b/crates/vm/src/builtins/str.rs
@@ -2723,4 +2723,24 @@ mod tests {
             assert_eq!("TypeError", &*translated.unwrap_err().class().name(),);
         })
     }
+
+    #[test]
+    fn str_isprintable_unicode15() {
+        // Regression test for https://github.com/RustPython/RustPython/issues/7525
+        // At the time of the issue, RustPython used unic_ucd_category which had
+        // outdated Unicode data, causing U+0B55 to be misclassified as Unassigned.
+        // Now fixed by migrating to icu_properties with up-to-date Unicode data.
+
+        // Characters that should be printable
+        assert!(PyStr::from("\u{0B55}").isprintable());
+        assert!(PyStr::from("A").isprintable());
+        assert!(PyStr::from(" ").isprintable());
+        assert!(PyStr::from("").isprintable());
+
+        // Characters that should NOT be printable
+        assert!(!PyStr::from("\x00").isprintable());
+        assert!(!PyStr::from("\u{200B}").isprintable());
+        assert!(!PyStr::from("\u{E000}").isprintable());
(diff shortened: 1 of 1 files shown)


ShaharNaveh (approved):
LGTM!

@ever0de what do you say about using [rtest](https://docs.rs/rstest/latest/rstest/)? I see that all the tests in this file are parameterized

ever0de:
@ShaharNaveh 
> LGTM!
> 
> @ever0de what do you say about using [rtest](https://docs.rs/rstest/latest/rstest/)? I see that all the tests in this file are parameterized

Sounds great! Though I notice `rstest` isn't used anywhere else in this crate yet, so we'd need to add it as a dev-dependency. Is that okay?

ShaharNaveh:
> @ShaharNaveh
> 
> > LGTM!
> > @ever0de what do you say about using [rtest](https://docs.rs/rstest/latest/rstest/)? I see that all the tests in this file are parameterized
> 
> Sounds great! Though I notice `rstest` isn't used anywhere else in this crate yet, so we'd need to add it as a dev-dependency. Is that okay?

IMO yes, we can open a ticket about it. perhaps even mark it as a "good first issue"

youknowone (approved):

PR #7767, opened 2026-05-02
Match CPython parity for compile() str source, mode wording, and func_type

changjoon-park:
## Summary

Three CPython parity gaps in `compile()` text-source handling, all in `crates/vm/src/stdlib/builtins.rs`.

### 1. str source path applies bytes-only encoding declaration validation

```python
>>> compile('# -*- coding: badencoding -*-\n"x"\n', 'tmp', 'exec')
# CPython 3.14.4: <code object>
# RustPython main: SyntaxError: unknown encoding for 'tmp': badencoding  ❌
```

Previously, `compile()` collapsed both `str` and bytes-like inputs to bytes via `ArgStrOrBytesLike::borrow_bytes()` and ran the same `decode_source_bytes()` over them, including the "unknown encoding" error path. CPython only applies `# -*- coding: ... -*-` validation when source is bytes — for `str` the source is already decoded and the declaration must be ignored.

Branch on `args.source` type early:
- pure `str` (no surrogates) → skip encoding-declaration validation
- surrogate-bearing `str` → fall back to bytes path (preserves existing SyntaxError)
- bytes-like → unchanged

### 2. `compile()` mode error wording

```python
>>> compile('x=1', '<x>', 'bogus')
# CPython 3.14.4: ValueError: compile() mode must be 'exec', 'eval' or 'single'
# RustPython main: ValueError: mode must be "exec", "eval", or "single"  ❌
```

The parser's `ModeParseError` Display produces a generic message used outside `compile()` too. Replace the `err.to_string()` wrapping at the four `mode_str.parse::<Mode>()` call sites in `builtins::compile` with the exact CPython wording.

### 3. Missing `func_type` validation on text-source path

```python
>>> compile('def f(x): pass', 'f', 'func_type')
# CPython 3.14.4: ValueError: compile() mode 'func_type' requires flag PyCF_ONLY_AST
# RustPython main: ValueError: mode must be "exec", "eval", or "single"  ❌
```

The AST-input path already had this check (lines 287-291); mirror it for text-source input so the user sees the actual cause (missing flag) rather than "invalid mode."

| Probe | Cases | Match |
|---|---|---|
| Source-input combinations: {str, bytes, bytearray, memoryview} × {no decl, valid utf-8, bad decl} | 9 | 9/9 |
| Mode strings (exec, eval, single, func_type, bogus, '', BAD) | 7 | 7/7 |

`cargo run --release -- -m test test_compile test_builtin test_compileall test_codeop test_ast` — 703 tests pass, 0 regressions. All 188 `extra_tests/snippets/*.py` pass under the CI feature set.

## Out of scope

`test_compile.test_encoding`'s `expectedFailure` marker is left in place: the test exercises non-UTF-8 encoding decoding (`# -*- coding: latin1 -*-`, `iso8859-15`) which is a separate pre-existing limitation in `decode_source_bytes` unaffected by this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of source encodings for text vs. binary inputs when compiling code.
  * Enhanced compatibility with CPython compile behavior and error responses.
  * Stricter validation of compile "mode" values, with clearer error handling for invalid modes.
* **Tests**
  * Added tests to verify encoding rules, mode error messages, and special-mode flag requirements.

### DIFF
--- a/crates/vm/src/stdlib/builtins.rs
+++ b/crates/vm/src/stdlib/builtins.rs
@@ -315,9 +315,11 @@ mod builtins {
                 }
                 #[cfg(feature = "rustpython-codegen")]
                 {
-                    let mode = mode_str
-                        .parse::<crate::compiler::Mode>()
-                        .map_err(|err| vm.new_value_error(err.to_string()))?;
+                    use crate::compiler::Mode;
+
+                    let mode = mode_str.parse::<Mode>().map_err(|_| {
+                        vm.new_value_error("compile() mode must be 'exec', 'eval' or 'single'")
+                    })?;
                     return _ast::compile(
                         vm,
                         args.source,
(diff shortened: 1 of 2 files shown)


ShaharNaveh (approved):
Thanks for the patch!

Does this fix something in the CPython test suite? if not, can you add a custom test to prevent regressions?

changjoon-park:
> Thanks for the patch!
> 
> Does this fix something in the CPython test suite? if not, can you add a custom test to prevent regressions?

thanks for catching ! it doesnt unmask an existing test, i added a regression snippet. please check it and thanks for the review :)

ShaharNaveh (approved):
looks great:)

youknowone (changes requested):
I am sorry about missing this patch! Could you please rebase it and resolve the lint failure? Thank you so much for contributing!

PR #7768, opened 2026-05-03
Validate compile() filename type and dont_inherit __bool__ protocol

changjoon-park:
## Summary

Two CPython parity gaps in `compile()` argument handling.

### 1. filename — buffer-protocol types should be rejected

```python
>>> compile('pass', bytearray(b'file.py'), 'exec')
# CPython 3.14.4: TypeError
# RustPython main: <code object>  ❌
```

CPython's `PyUnicode_FSDecoder` accepts `str`, `bytes`, and objects with `__fspath__` only — `bytearray` / `memoryview` / `array.array` raise TypeError. RustPython's `FsPath::TryFromObject` impl falls back to any buffer-protocol object and silently converts to bytes.

Change `CompileArgs::filename` from `FsPath` (which uses the permissive `TryFromObject` impl) to `PyObjectRef`, then call the strict `FsPath::try_from_path_like` at the top of `compile()`. Other `FsPath` consumers are unchanged.

### 2. dont_inherit — should call `__bool__`, not strict-check the bool type

```python
>>> class EvilBool:
...     def __bool__(self): raise ValueError('hi')
>>> compile('pass', 'f', 'exec', dont_inherit=EvilBool())
# CPython 3.14.4: ValueError('hi')
# RustPython main: TypeError('Expected type bool, not EvilBool')  ❌
```

CPython routes `dont_inherit` through `PyObject_IsTrue`, calling `__bool__` on arbitrary objects and propagating any exception raised there. Switch from `OptionalArg<bool>` to `OptionalArg<ArgIntoBool>` (already used elsewhere in `builtins` for `all`, `any`, `print(..., flush=...)`).

| Probe | Cases | Match |
|---|---|---|
| filename basic types (str, bytes, bytearray, memoryview, list, int) | 6 | 6/6 |
| filename edge cases (NUL byte, `pathlib.Path`, custom `__fspath__`, bad `__fspath__`, str/bytes subclass) | 8 | 8/8 |
| dont_inherit basic values (True/False/1/0/""/"yes"/None) + EvilBool | 8 | 8/8 |
| dont_inherit edge cases (int, float, list, dict, non-bool `__bool__`, plain object) | 6 | 6/6 |

## Tests unmasked

`Lib/test/test_compile.py`:
- `test_compile_filename`
- `test_compile_filename_refleak`

## Test runs

`cargo run --release -- -m test test_compile test_builtin test_compileall test_codeop test_ast` — 703 tests pass, 0 regressions. All 188 `extra_tests/snippets/*.py` pass under the CI feature set.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `compile()` function for enhanced CPython compatibility with stricter parameter validation and better truthiness evaluation.

### DIFF
--- a/Lib/test/test_compile.py
+++ b/Lib/test/test_compile.py
@@ -647,7 +647,6 @@ class TestSpecifics(unittest.TestCase):
         d = {f(): f(), f(): f()}
         self.assertEqual(d, {1: 2, 3: 4})
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON; AssertionError: TypeError not raised
     def test_compile_filename(self):
         for filename in 'file.py', b'file.py':
             code = compile('pass', filename, 'exec')
@@ -657,7 +656,6 @@ class TestSpecifics(unittest.TestCase):
                 compile('pass', filename, 'exec')
         self.assertRaises(TypeError, compile, 'pass', list(b'file.py'), 'exec')
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON; TypeError: Expected type bool, not EvilBool
     def test_compile_filename_refleak(self):
         # Regression tests for reference leak in PyUnicode_FSDecoder.
         # See https://github.com/python/cpython/issues/139748.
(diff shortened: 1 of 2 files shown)


ShaharNaveh (approved):
ty:)

youknowone (approved):

PR #7780, opened 2026-05-04
Use icu4x for casefold()

joshuamegnauth54:
For ASCII, casefold() is equivalent to lower casing the string.

Casefolding is a Unicode operation defined by the standard. Both CPython and RustPython follow the standard without any frills. The icu4x set of crates implement casefold as per the standard, so RustPython can use the ICU crates instead of adding in a crate for one operation.

This also has the extra benefit of consistency - consistent Unicode version and consistent crate usage.

## Summary by CodeRabbit

* **Refactor**
  * Enhanced string case-folding with improved Unicode support for handling international characters.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7780)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -186,7 +186,6 @@ bitflags = "2.11.0"
 bitflagset = "0.0.3"
 bstr = "1"
 bzip2 = "0.6"
-caseless = "0.2.2"
 chrono = { version = "0.4.44", default-features = false, features = ["clock", "std"] }
 console_error_panic_hook = "0.1"
 constant_time_eq = "0.4"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -430,15 +430,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "caseless"
-version = "0.2.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8b6fd507454086c8edfd769ca6ada439193cdb209c7681712ef6275cccbfe5d8"
-dependencies = [
- "unicode-normalization",
-]
-
-[[package]]
 name = "cast"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


ShaharNaveh (approved):
Thanks, like that a lot!
Do we have more places that uses `caseless`? maybe worth fixing them in a follow up PR and remove it a dependency if icu4x already providing that functionality

joshuamegnauth54:
Nope, `caseless` is used exactly once, so it's an easy win. :grin: 

The build is intentionally failing because I want to avoid a merge conflict on my end when I rebase on #7717. So after I rebase, I'll remove `caseless` and then force push.

Cargo.toml
```diff
@@ -186,7 +186,6 @@ bitflags = "2.11.0"
 bitflagset = "0.0.3"
 bstr = "1"
 bzip2 = "0.6"
-caseless = "0.2.2"
```
ShaharNaveh, line 189: nice

ShaharNaveh (approved):
TYSM!

fanninpm:
> `752-770`: _⚡ Quick win_
> 
> **Add regression coverage for the new ICU-backed `casefold` paths.**
> 
> This swaps the folding engine and introduces custom WTF-8 chunk handling, but the provided test module still doesn't exercise `casefold`. Please add focused cases for ASCII, expansion folds like `ß -> ss`, Greek sigma behavior, and lone-surrogate preservation so future ICU or Unicode-data bumps don't silently drift behavior.

@coderabbitai Please examine `StrTests.test_casefold()` in `Lib/test/test_str.py`.

fanninpm:
> RustPython does not bundle CPython's standard test suite; `Lib/` is not present in the repo at all

@coderabbitai Are you sure? Looking in `.coderabbit.yml`, it seems the `Lib/` is given special treatment.

youknowone (approved):

PR #7786, opened 2026-05-05
Unchecked code fixes

ShaharNaveh:
While trying to run: `clippy --all-features --workspace` I've seen lots of errors. This PR goal is to break the fixes needed for it to run on the CI, into smaller chunks

## Summary by CodeRabbit

* **Chores**
  * Prevent simultaneous enablement of conflicting SSL implementations; improve SSL configuration safety.
  * Improve portability and internal consistency across native SSL/cert handling and core utilities.
* **Style**
  * Refined debug/tracing output formatting for opcode execution logs.
  * Silenced select false-positive warnings and clarified lint intent.
* **Bug Fixes**
  * Made error messages for certain UI bindings clearer and more accurate.


crates/stdlib/src/ssl/compat.rs
```diff
@@ -8,6 +8,10 @@
 
 // SSL error code data tables (shared with OpenSSL backend for compatibility)
 // These map OpenSSL error codes to human-readable strings
+#[allow(
+    clippy::duplicate_mod,
+    reason = "This is duplicated only when running clippy. The two features are mutually exlusive"
```
fanninpm: Spelling:
```suggestion
    reason = "This is duplicated only when running clippy. The two features are mutually exclusive"
```

crates/stdlib/src/ssl/compat.rs
```diff
@@ -8,6 +8,10 @@
 
 // SSL error code data tables (shared with OpenSSL backend for compatibility)
 // These map OpenSSL error codes to human-readable strings
+#[allow(
```
fanninpm, line 11: Does this work?
```suggestion
#[expect(
```

crates/stdlib/src/ssl/error.rs
```diff
@@ -125,6 +125,7 @@ pub(crate) mod ssl_error {
         )
     }
 
+    #[allow(dead_code, reason = "This seems like a false positive")]
```
fanninpm, line 128: Perhaps more reasonable here (and elsewhere in this file) would be to only let this be compiled if it is used:
```rust
    #[cfg(...)]
```

crates/stdlib/src/openssl.rs
```diff
@@ -3,6 +3,10 @@
 mod cert;
 
 // SSL exception types (shared with rustls backend)
+#[allow(
+    clippy::duplicate_mod,
+    reason = "This is duplicated only when running clippy. The two features are mutually exlusive"
```
fanninpm: Spelling:
```suggestion
    reason = "This is duplicated only when running clippy. The two features are mutually exclusive"
```

crates/stdlib/src/tkinter.rs
```diff
@@ -80,6 +80,7 @@ mod _tkinter {
     static QUIT_MAIN_LOOP: AtomicBool = AtomicBool::new(false);
     static ERROR_IN_CMD: AtomicBool = AtomicBool::new(false);
 
+    #[allow(dead_code, reason = "TODO: Impl more methods")]
```
fanninpm: Does this work?
```suggestion
    #[expect(dead_code, reason = "TODO: Impl more methods")]
```

crates/stdlib/src/ssl/error.rs
ShaharNaveh: I'm passing `--all-targets --all-features`, I haven't looked to much into it :shrug:

fanninpm:
The reason why I ask "Does this work?" is that `#[expect(...)]` will panic if the condition it is given is not true.

ShaharNaveh:
> The reason why I ask "Does this work?" is that `#[expect(...)]` will panic if the condition it is given is not true.

I'm aware, let's see if that works:)

fanninpm:
> > The reason why I ask "Does this work?" is that `#[expect(...)]` will panic if the condition it is given is not true.
> 
> I'm aware, let's see if that works:)

`error: this lint expectation is unfulfilled`

It doesn't work.

ShaharNaveh:
> > > The reason why I ask "Does this work?" is that `#[expect(...)]` will panic if the condition it is given is not true.
> > 
> > 
> > I'm aware, let's see if that works:)
> 
> `error: this lint expectation is unfulfilled`
> 
> It doesn't work.

Right, that's because it isn't reached without passing `--all-features` :facepalm:

PR #7788, opened 2026-05-05
Fix swapcase()

joshuamegnauth54:
The tests for swapcase() were failing for two reasons. The first is '𐐧' casing which should be fixed with modern Unicode tables. The second failure is due to CPython's sigma override, which I implemented in #7117.

---

~~This should work after rebasing, so I'm opening it as a placeholder for now. :smile_cat:~~ Ready since #7117 is merged.

## Summary by CodeRabbit

* **Refactor**
  * Optimized text case operations: faster ASCII fast paths and a new UTF‑8 swapcase routine for improved performance.
  * Byte-level case swapping now uses a streamlined ASCII path; invalid or non-UTF‑8 bytes are preserved during UTF‑8 processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7788)


youknowone (approved):
Thanks!

PR #7790, opened 2026-05-06
Make `determine_changes` step run on `ubuntu-slim`

fanninpm:
Since the `ubuntu-slim` image contains the `git` executable by default, we can use this image and thus improve our actions parallelism.

Ideally, we should consider extracting the `cargo_check` step into a separate GitHub Actions `.yaml` file, possibly obviating the need for the `determine_changes` step.

cc @ShaharNaveh

## Summary by CodeRabbit

* **Chores**
  * Optimized CI workflow configuration for improved build efficiency.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -39,7 +39,7 @@ env:
 jobs:
   determine_changes:
     name: Determine changes
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-slim
     outputs:
       # Flag that is raised when any rust code is changed.
       rust_code: ${{ steps.check_rust_code.outputs.changed }}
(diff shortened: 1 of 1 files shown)


ShaharNaveh:
> Since the `ubuntu-slim` image contains the `git` executable by default, we can use this image and thus improve our actions parallelism.

I like this change, let's see how it playes out for a while.

> Ideally, we should consider extracting the `cargo_check` step into a separate GitHub Actions `.yaml` file, possibly obviating the need for the `determine_changes` step.

I was debating it tbh, but I've decided against it because we have some env that we would need to keep synced (CARGO_EXCLUDE, CARGO_ARG and other cargo), once you start separating the workflows too much it becomes a hell to manage:/

ShaharNaveh (approved):
great!

PR #7792, opened 2026-05-06
Fixed: `unused manifest key: dependencies.libz-sys.package`

ShaharNaveh:
## Summary by CodeRabbit

* **Chores**
  * Updated an internal dependency reference to use the workspace crate directly instead of an aliased entry; versioning and behavior remain unchanged. No functional or user-visible changes.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7792)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -224,7 +224,7 @@ libloading = "0.9"
 liblzma = "0.4"
 liblzma-sys = "0.4"
 libsqlite3-sys = "0.37"
-libz-sys = { package = "libz-rs-sys", version = "0.5" }
+libz-rs-sys = "0.5"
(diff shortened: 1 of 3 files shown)


fanninpm (approved):
This seems good to go as it is.

Alternatively, you can probably replace all mentions of `libz-sys` in RustPython with `libz-rs-sys` to have the same effect.

youknowone (changes requested):
It changed dependency libz-rs-sys -> libz-sys . is this intended?
libz-rs-sys is pure rust replacement of libz-sys

youknowone (approved):

PR #7793, opened 2026-05-06
Fix sys.implementation

### LINKED ISSUE #7770: Differentiate `sys.implementation.version` from `sys.version_info`
## Feature

I'm writing an application that runs a third-party script, which operates on pickles generated by another application. There are different versions of the script for Python 2 and Python 3, and the Python version must match the Python version used by that other application, but the actual implementation doesn't matter -- CPy, RustPy, Jython, IronPy, and PyPy all work properly.

My application allows the user to choose their Python executable of choice. It needs to check whether it implements Python 2 or Python 3 and also collect and show a proof-of-life (i.e. "RustPython 0.5.0 (64bit, for Python 3.14)"). Usually I'd get that from PEP514 registrations, but I also need to generate it if someone browses to an unregistered interpreter like RustPython. I'm getting incorrect output from RustPython when I try to obtain that information.

There are two version numbers in `sys`, and RustPython incorrectly uses the same version information in both:
* `sys.version_info` -- This is the version of CPython that RustPython is trying to copy, and is correctly set to (essentially) `3.14.0alpha`.
* `sys.implementation.version` -- This is the version of RustPython itself. It should be set to (essentially) `0.5.0`, but instead is set to the same version as above.

https://docs.python.org/3/library/sys.html#sys.implementation

joshuamegnauth54:
`sys.version_info` and `sys.implementation` are two different versioning schemes. `version_info` refers to the target Python version while `implementation` refers to the implementation version.

For RustPython, `version_info` should be 3.14 because it currently targets Python 3.14. This was correct and remains unchanged.

`implementation`, however, should be set to RustPython's version. I updated the code to correctly get RustPython's version from the environment. I also ensured values are hard coded at build time which reduces the overall code that needs to be run at runtime as well as removes some unneeded allocations.

Closes: #7770

## Summary by CodeRabbit

* **Refactor**
  * Version and build metadata are now embedded at compile time, exposing implementation-specific major/minor/micro, release level, serial, hex version, git revision and build timestamp.
  * sys._git now returns an additional git revision field.
  * sys.implementation and reported hexversion now reflect the new implementation-specific versioning.
  * Build script improved for more reliable git and compiler metadata extraction.

### DIFF
--- a/crates/vm/src/version.rs
+++ b/crates/vm/src/version.rs
@@ -11,9 +11,36 @@ pub const MICRO: usize = 0;
 pub const RELEASELEVEL: &str = "alpha";
 pub const RELEASELEVEL_N: usize = 0xA;
 pub const SERIAL: usize = 0;
-
 pub const VERSION_HEX: usize =
     (MAJOR << 24) | (MINOR << 16) | (MICRO << 8) | (RELEASELEVEL_N << 4) | SERIAL;
+
+pub const GIT_REVISION: &str = env!("RUSTPYTHON_GIT_HASH");
+const GIT_TAG: &str = env!("RUSTPYTHON_GIT_TAG");
+const GIT_BRANCH: &str = env!("RUSTPYTHON_GIT_BRANCH");
+
+// RustPython version
+pub const MAJOR_IMPL: usize = match usize::from_str_radix(env!("CARGO_PKG_VERSION_MAJOR"), 10) {
+    Ok(v) => v,
+    Err(_) => panic!("Compile with Cargo to get 'CARGO_PKG_VERSION_MAJOR'"),
+};
+pub const MINOR_IMPL: usize = match usize::from_str_radix(env!("CARGO_PKG_VERSION_MINOR"), 10) {
+    Ok(v) => v,
+    Err(_) => panic!("Compile with Cargo to get 'CARGO_PKG_VERSION_MINOR'"),
+};
+pub const MICRO_IMPL: usize = match usize::from_str_radix(env!("CARGO_PKG_VERSION_PATCH"), 10) {
+    Ok(v) => v,
+    Err(_) => panic!("Compile with Cargo to get 'CARGO_PKG_VERSION_PATCH'"),
+};
+pub const RELEASELEVEL_IMPL: &str = env!("RUSTPYTHON_RELEASE_LEVEL");
+pub const SERIAL_IMPL: usize = match usize::from_str_radix(env!("RUSTPYTHON_RELEASE_SERIAL"), 10) {
+    Ok(v) => v,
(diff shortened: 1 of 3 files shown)


crates/vm/src/version.rs
```diff
@@ -11,10 +11,29 @@ pub const MICRO: usize = 0;
...
 
+pub const GIT_REVISION: &str = env!("RUSTPYTHON_GIT_HASH");
+const GIT_TAG: &str = env!("RUSTPYTHON_GIT_TAG");
+const GIT_BRANCH: &str = env!("RUSTPYTHON_GIT_BRANCH");
+
+// RustPython version
+pub const MAJOR_IMPL: usize = match usize::from_str_radix(env!("CARGO_PKG_VERSION_MAJOR"), 10) {
+    Ok(v) => v,
+    Err(_) => panic!("Compile with Cargo to get 'CARGO_PKG_VERSION_MAJOR'"),
+};
```
fanninpm, line 25: Does this work?

```suggestion
pub const MAJOR_IMPL: usize = usize::from_str_radix(env!("CARGO_PKG_VERSION_MAJOR"), 10)
    .expect("Compile with Cargo to get 'CARGO_PKG_VERSION_MAJOR'");
```

If it does, I suggest doing likewise for `MINOR_IMPL` and `MICRO_IMPL`.

fanninpm: Actually, this might not work because `.expect` is not `const`.

joshuamegnauth54: Unfortunately, [Result::unwrap](https://doc.rust-lang.org/std/result/enum.Result.html#method.unwrap), and its variations like `expect()`, as well as [Result::ok](https://doc.rust-lang.org/std/result/enum.Result.html#method.ok) are all const unstable. The code is a little ugly but calculating these values once at compile time is worth it, I think!

joshuamegnauth54:
Force pushed to address the bunny's lints. :rabbit: 

While addressing its lints, I added two env vars to `build.rs` that ideally should come from CI: `RUSTPYTHON_RELEASE_LEVEL` and `RUSTPYTHON_RELEASE_SERIAL`. These correspond to the implementation's `releaselevel` and `serial` respectively. They default to `alpha` and `0` if absent, which matches preexisting behavior. These variables can be passed down from CI. For example, for packaged releases, CI can pass down `stable` and `0` which would be baked in and automatically used by `sys.implementation`.

ShaharNaveh (approved):
ty:)

crates/vm/src/version.rs
fanninpm: The `expect`, `unwrap`, `expect_err`, and `unwrap_err` methods on `Result<T, E>` are not even `const` at all because the internal method `unwrap_failed` itself is not const.

PR #7796, opened 2026-05-07
Move dependency to workspace for example `rustpython-without-js`

fanninpm:
cc @ShaharNaveh

## Summary by CodeRabbit

* **Chores**
  * Updated internal dependency management configuration to use workspace-managed versioning for consistency.


youknowone:
Because this is example, not sure this is good idea

PR #7799, opened 2026-05-07
Use rustix for uname; drop uname crate

joshuamegnauth54:
---

Nothing fancy here. `uname` is a small crate that hasn't been touched in ten years. It doesn't do anything spectacular. [Here](https://github.com/icorderi/rust-uname/blob/master/src/lib.rs) is its source. `rustix`, which is already being used by RustPython, provides a `uname()` wrapper that we can use instead.

I also dropped the `oldtime` feature from `chrono`. According to the `chrono` docs, that feature doesn't do anything anymore.

## Summary by CodeRabbit

* **Chores**
  * Updated time library configuration for improved compatibility
  * Refactored system information retrieval with streamlined dependency management
  * Internal code organization improvements with no functional changes to user-facing features

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3567,7 +3567,6 @@ dependencies = [
  "strum_macros",
  "thiserror 2.0.18",
  "timsort",
- "uname",
  "unicode-casing",
  "wasm-bindgen",
  "which",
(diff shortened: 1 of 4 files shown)


ShaharNaveh (approved):
ty:)

PR #7811, opened 2026-05-09
Update `ensurepip` to 3.14.4

ShaharNaveh:


youknowone (approved):
awesome. the last version of pip whl had local patch, but now we don't need one.

PR #7864, opened 2026-05-12
Update `test_peepholer.py` to 3.14.5

ShaharNaveh:


youknowone:
@ShaharNaveh I don't mind to have multiple libs in single PR when the diff is small enough. That will be helpful to save CI resource

PR #7888, opened 2026-05-17
CI: Extract sandbox tests to their own job

ShaharNaveh:
related to #7883.
I'll continue optimizing the CI so only necessary jobs trigger on their related changes

## Summary by CodeRabbit

* **Tests**
  * CI adds a dedicated sandbox test job that performs Linux compilation checks, sandbox smoke tests, and a separate SSL build step.
  * Sandbox tests run when related code or test areas change and always run on the main branch.
  * Sandbox-related steps were moved out of the previous test flow for clearer, focused validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7888?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -43,6 +43,10 @@ jobs:
     outputs:
       # Flag that is raised when any rust code is changed.
       rust_code: ${{ steps.check_rust_code.outputs.changed }}
+      # Flag that is raised when any upstream lib code is changed.
+      upstream_lib: ${{ steps.check_upstream_lib.outputs.changed }}
+      # Flag that is raised when any extra test is changed.
+      extra_test: ${{ steps.check_extra_test.outputs.changed }}
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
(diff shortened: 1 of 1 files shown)


youknowone:
I am not sure how this help CI optimization. could you explain more?

ShaharNaveh:
> I am not sure how this help CI optimization. could you explain more?

Sure, the idea is to make `cargo test` to run only when there's change to rust related code or when something changes under `Lib/**` but not `Lib/test/**`. atm, the job has some steps that needs to be triggered with different conditions.

I'm decoupling the `rust_test ` job by creating new jobs/moving some steps to a more suited job

youknowone:
isn't host_env expected to share the same triggering condition?

ShaharNaveh:
> isn't host_env expected to share the same triggering condition?

host_env is tested as long as a PR doesn't have the `ci:skip` label (it's triggered no matter what)

youknowone:
I am sorry. Could you explain a bit more? I still didn't get how this is helping optimizing CI. We have limited concurrent runners. and already have too many jobs. I often see CI actions are queued. 
I expect similar tasks, like possibly reusing previous cargo cache running in same job reduce the total build cost. Also I cannot think of with/without sandbox have different condition to be tested or not.

ShaharNaveh:
> I am sorry. Could you explain a bit more? I still didn't get how this is helping optimizing CI. We have limited concurrent runners. and already have too many jobs. I often see CI actions are queued. I expect similar tasks, like possibly reusing previous cargo cache running in same job reduce the total build cost. Also I cannot think of with/without sandbox have different condition to be tested or not.

Maybe the sandbox tests are a bit much. But the goal is to run `cargo test` only when `Lib/**/*.py`(excluding `Lib/test/*`) changes or anything under `crates/*`. For that I need to find the rest of the commands a place to belong to, as they don't match this condition

Issue #6839, opened 2026-01-23
Upgrade Python libraries and test from Python 3.14


youknowone:
## Summary

Check `scripts/update_lib` for tools. As a note, the current latest Python version is `v3.14.3`.

Previous versions' issues as reference
- 3.13: #5529 

## Details

2026-02-11 00:39:24
```shell
$ python3 scripts/update_lib todo --done
```
- [ ] [no deps] `types` (40 dependents) | Δ4
  - [ ] test_types (10 TODO)
- [x] [no deps] `struct` (17 dependents)
  - [ ] test_struct (4 TODO)
- [ ] [no deps] `stat` (14 dependents) | Δ11
  - [ ] test_stat
- [x] [no deps] `operator` (13 dependents)
  - [ ] test_operator (3 TODO)
- [x] [no deps] `linecache` (9 dependents)
  - [x] test_linecache
- [x] [no deps] `keyword` (8 dependents)
  - [x] test_keyword
- [x] [no deps] `codecs` (8 dependents)
  - [ ] test_codecs (11 TODO)
  - [ ] test_codeccallbacks (9 TODO)
  - [ ] test_codecencodings_cn (untracked)
  - [ ] test_codecencodings_hk (untracked)
  - [ ] test_codecencodings_iso2022 (untracked)
  - [ ] test_codecencodings_jp (untracked)
  - [ ] test_codecencodings_kr (untracked)
  - [ ] test_codecencodings_tw (untracked)
  - [ ] test_codecmaps_cn (untracked)
  - [ ] test_codecmaps_hk (untracked)
  - [ ] test_codecmaps_jp (untracked)
  - [ ] test_codecmaps_kr (untracked)
  - [ ] test_codecmaps_tw (untracked)
  - [ ] test_charmapcodec
  - [ ] test_multibytecodec (untracked)
- [x] [no deps] `__future__` (6 dependents)
  - [ ] test_future_stmt (7 TODO)
- [x] [no deps] `copyreg` (6 dependents)
  - [x] test_copyreg
- [x] [no deps] `reprlib` (6 dependents)
  - [x] test_reprlib (2 TODO)
- [x] [no deps] `bisect` (5 dependents)
  - [x] test_bisect
- [x] [no deps] `token` (5 dependents)
- [x] [no deps] `string` (5 dependents)
  - [x] test_string
  - [x] test_userstring
- [x] [no deps] `heapq` (4 dependents)
  - [x] test_heapq
[...]

PR #7902, opened 2026-05-17
Update NOP tests to 3.14.5

ShaharNaveh:
Those tests should be skipped, thus a "NOP test".

The motivation of this PR is to mark those items as done at #6839


youknowone:
is this worth to do? we can simply add blocklist for #6839  too

ShaharNaveh:
> is this worth to do? we can simply add blocklist for #6839 too

I think so, it will provide better tests if/when we'll have `_testcapi` for example

youknowone (approved):

PR #7937, opened 2026-05-20
Add call functions to c-api

bschoenmaeckers:
## Summary by CodeRabbit

* **New Features**
  * Expanded C API object calling interface with new functions supporting multiple invocation patterns. These include standard calls with arguments and keyword arguments, efficient no-argument calls, and optimized vectorcall variants. These additions enable improved performance, better compatibility, and provide developers with flexible mechanisms for object invocation in C-based code.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7937?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/crates/vm/src/function/argument.rs
+++ b/crates/vm/src/function/argument.rs
@@ -1,4 +1,9 @@
-use crate::{AsObject, PyObjectRef, PyPayload, PyRef, PyResult, TryFromObject, VirtualMachine, builtins::{PyBaseExceptionRef, PyDict, PyStr, PyTuple, PyTupleRef, PyTypeRef}, convert::ToPyObject, object::{Traverse, TraverseFn}, Py};
+use crate::{
+    AsObject, Py, PyObjectRef, PyPayload, PyRef, PyResult, TryFromObject, VirtualMachine,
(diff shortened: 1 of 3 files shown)


crates/vm/src/function/argument.rs
```diff
@@ -438,6 +438,29 @@ impl<T> FromIterator<(String, T)> for KwArgs<T> {
     }
 }
 
+impl TryFrom<&Py<PyDict>> for KwArgs<PyObjectRef> {
```
youknowone: `TryFrom<&Py<PyDict>>` and `From<&Py<PyTuple>>` are high-cost conversion and only will be used for c api. rather than adding this conversion, creating a helper function and manually calling them will be reasonable

youknowone (changes requested):

crates/vm/src/function/argument.rs
bschoenmaeckers: Done

bschoenmaeckers:
@coderabbitai  your last comment is not correct. Because the pointer to pos_arg_1 points to the start of the array, and the array still contains the keywords arguments after it. It does not patter we do not count the kwargs in the slice.

youknowone (approved):

PR #7947, opened 2026-05-21
Fix most of `unicodedata`, unmask multiple tests, and remove `ucd`

joshuamegnauth54:
Python bundles an old version of Unicode for compatibility. RustPython tries to mimic supporting that old version by checking the version of individual chars. This is a problem for a few reasons. The first is that the age check adds an additional hit per each char lookup in Unicode data. The check is outdated because the `unic-ucd-age` crate is several versions behind the current Unicode version. The check rejects valid chars because of the version differences.

The check is subtly wrong because it returns properties for Unicode 16.0.0 for Unicode 3.2.0 while checking against a Unicode 10.0.0 database.

Unfortunately, there isn't a crate that can help us here. `icu4x` targets modern Unicode versions. Writing a data provider for `icu4x` for Unicode 3.2.0 is a lot of work for a legacy path. I opted to parse the Unicode 3.2.0 data myself but to skip `icu4x` (mostly) to instead write small lookup tables.

As of this commit, Unicode names is still wrong for 3.2.0. Luckily, the crate RustPython uses is fast and robust for modern Unicode.

## Summary by CodeRabbit

* **New Features**
  * Vendored Unicode 3.2.0 plus new “latest” compatibility datasets for broader coverage.

* **Improvements**
  * Version-aware Unicode lookups (category, bidi, width, decomposition) with reduced allocations.
  * Numeric APIs (digit/decimal/numeric) clarified to return normalized numeric values where applicable.

* **Bug Fixes**
  * Applied normalization/decomposition corrections for historic compatibility.

* **Documentation**
  * Added guidance on vendored Unicode data and update policy.

* **Tests**
  * Updated Unicode tests to validate decompositions and versioned behavior.


crates/literal/src/char.rs
```diff
@@ -10,17 +10,11 @@ use icu_properties::props::{EnumeratedProperty, GeneralCategory};
...
-            | GeneralCategory::LineSeparator
-            | GeneralCategory::ParagraphSeparator
-            | GeneralCategory::Control
-            | GeneralCategory::Format
-            | GeneralCategory::Surrogate
-            | GeneralCategory::PrivateUse
-            | GeneralCategory::Unassigned
-    )
+    c == '\u{0020}' || printable.contains(GeneralCategory::for_char(c))
 }
```
ShaharNaveh, line 26: Probably worth trying to get this patch to ruff as well.

https://github.com/astral-sh/ruff/blob/ceca6028373d31f3a7d44a5e3e00d7c0b2dd8614/crates/ruff_python_literal/src/char.rs

crates/stdlib/src/unicodedata.rs
```diff
@@ -18,6 +22,24 @@ enum NormalizeForm {
...
 
+#[derive(Clone, Copy)]
+pub struct UnicodeVersion {
+    pub major: u8,
+    pub minor: u8,
+    pub micro: u8,
+}
+
+/// Supported Unicode version based on icu4x.
+pub const UNICODE_VERSION: UnicodeVersion = UnicodeVersion {
```
ShaharNaveh: I assume icu4x doesn't expose this:/

Lib/test/test_str.py
```diff
@@ -853,7 +853,6 @@ def test_isprintable(self):
         self.assertTrue('\U0001F46F'.isprintable())
         self.assertFalse('\U000E0020'.isprintable())
 
-    @unittest.expectedFailure  # TODO: RUSTPYTHON
```
ShaharNaveh, line 856: <3

ShaharNaveh (approved):
TYSM!

Overall I really like this change.

The only thing rhat might be worth is to a README.md under `crates/stdlib/unicode/ucd32/README.md` with links for the soure of each file so we can upsate them in the future if needed

crates/literal/src/char.rs
joshuamegnauth54: Maybe! There's a chance I'll revert that change if the Unicode 3.2 fixes also fix that test. :laughing: 

The test seems to be failing because RustPython's `unicodedata` mixes Unicode versions. `ucd` is Unicode 10.0.0 whereas Rust, `icu4x`, and Python 3.14+ are all Unicode 16.0.0.

crates/stdlib/src/unicodedata.rs
joshuamegnauth54: I couldn't find it, but Rust itself exposes its Unicode version so I'm using that instead for now. Both Rust and `icu4x` target the latest Unicode version so it should be fine. I hope. :thinking: :thinking:

joshuamegnauth54:
Also, the Unicode 3.2 files will never change but I agree about linking the sources somewhere. CPython seems to provide 3.2 as a compatibility for an old IDNA version (which is also provided for compatibility :joy_cat: ). All of this work is essentially to fix a 20+ year old module that is likely not used anywhere. :sweat_smile: 

[Unicode 3.2](https://www.unicode.org/reports/tr28/tr28-3.html) - here is the source for now. I'll add it to a README after I finish the patch. :grin:

joshuamegnauth54:
I undrafted for CodeRabbit because I couldn't trigger it. There will likely be a lot of suggestions. :sweat_smile:

extra_tests/snippets/builtin_str_unicode.py
```diff
@@ -20,6 +20,13 @@
 assert unicodedata.bidirectional("a") == "L"
 assert unicodedata.east_asian_width("\u231a") == "W"
 assert unicodedata.normalize("NFC", "bla") == "bla"
+square_aruhua = "<square> 30A2 30EB 30D5 30A1"
+assert unicodedata.decomposition("\u3301") == square_aruhua
+assert unicodedata.ucd_3_2_0.decomposition("\u3301") == square_aruhua
+
+# These are from NormalizationCorrections.txt
+assert unicodedata.decomposition("\uf951") == "964B"
+assert unicodedata.ucd_3_2_0.decomposition("\uf951") == "964B"
```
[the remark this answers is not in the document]
joshuamegnauth54: This is incorrect. Both CPython and RustPython return "964B".

Here's the line in `NormalizationCorrections.txt`:

```
F951;96FB;964B;3.2.0 # Corrigendum 3
```
The character is `\u{F951}`. The original, incorrect decomposition is `\u{96FB}`. The decomposition was correction in version 3.2.0 which means that versions >= 3.2.0 use the correct decomposition.

crates/literal/src/char.rs
joshuamegnauth54: I ended up reverting this change since removing `ucd` and fixing Unicode 3.2.0 support fixed the test. :grin:

youknowone (approved):
Thank you so much!

ShaharNaveh (approved):
this is great!

tysm for work you've put into this<3

joshuamegnauth54:
Sorry, I forgot to draft this again. :joy: I'm almost finished with it though.

fanninpm:
> I undrafted for CodeRabbit because I couldn't trigger it. There will likely be a lot of suggestions. 😅

@coderabbitai Can a CodeRabbit review be triggered by saying `@coderabbitai review`?

crates/stdlib/build.rs
```diff
@@ -3,7 +3,602 @@
...
+        let line = line.trim();
+        if line.is_empty() || line.starts_with('#') {
+            continue;
+        }
+
+        let mut fields = line.split(';');
+        let range = fields.next().expect("Unicode data is missing a char range");
+        let id = fields
+            .nth(field.saturating_sub(1))
+            .expect("Unicode data is missing a property");
```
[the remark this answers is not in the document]
joshuamegnauth54: I intended for `0` and `1` to be treated the same. I expected the caller to pass in a field number, such as `5`, and then internally subtract it to get the real field number. I always consume the first field so the real field is always `field - 1`.

You're right that it seems confusing. :joy: Since the parser unwraps everywhere, there wasn't a risk of using it incorrectly without a panic. To address your review, I changed the fields argument to be a `NonZeroUsize` instead to make it clear that the field should be anything but field `0`.

joshuamegnauth54:
I got as far as I could get with these fixes. The bidirectional test should pass after the tests are bumped. CPython's `numeric` for 3.2.0 is _extremely_ convoluted. It accounts for characters that don't even exist in Unicode. I couldn't figure it out. EastAsianWidth is the same for 3.2.0 - I couldn't figure the logic at all.

ShaharNaveh:
> I got as far as I could get with these fixes. 

:saluting_face: 
TYSM! <3

crates/stdlib/build.rs
```diff
@@ -3,7 +3,604 @@
...
+            }
+        },
+        |writer, mut values| {
+            values.sort_unstable_by_key(|(c, _)| *c);
+            write!(writer, "{values:?};").unwrap();
+        },
+    );
+}
+
+#[allow(clippy::too_many_arguments)]
```
ShaharNaveh:
```suggestion
#[expect(clippy::too_many_arguments)]
```

ShaharNaveh (approved):
TYSM!

PR #7800, opened 2026-05-08
Bump rand, mt19937; fix associated build failures

joshuamegnauth54:
Supercedes: #7716

Bumping rand and mt19937 is not trivial so Dependabot's PR causes build failures.

~~Update: Draft until I figure out why the WebAssembly build is failing. The rest of it is working, though.~~ I reverted the `getrandom` bump. Apparently, the new version is conflicting with version 0.3 which is pulled in by other dependencies.

## Summary by CodeRabbit

* **Chores**
  * Updated random-related dependencies across the workspace to newer, compatible versions.
  * Enabled the JS/random backend and explicitly enabled std+wasm features for WebAssembly targets.

* **Bug Fixes / Behavior Change**
  * When no seed is provided, the default RNG now seeds from the platform thread RNG instead of attempting direct OS entropy.

PR #7954, opened 2026-05-23
Replace `ahash` with `rapidhash`

fanninpm:
Follow-up to #7800.

Back in 2021, `ahash` was added to RustPython in #2397 because it was faster than the Rust standard library hash/`RandomState`. 
Since then, hashing technology has evolved to the point where `rapidhash` beats `ahash` in both the `SMHasher` and `SMHasher3` test suites. 
Additionally, `rapidhash` has very minimal direct dependencies and is portable across CPU architectures.

cc @coolreader18 @joshuamegnauth54

## Summary by CodeRabbit

* **Chores**
  * Replaced the workspace hashing dependency with an alternative implementation and updated usages across crates for consistent hashing behavior.
  * Adjusted dependency grouping so the previous hash crate is no longer included in the automated grouping/configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7954?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -85,7 +85,6 @@ updates:
           - "quote-use*"
       random:
         patterns:
-          - "ahash"
           - "getrandom"
           - "mt19937"
           - "rand*"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2929,6 +2929,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "rapidhash"
+version = "4.4.1"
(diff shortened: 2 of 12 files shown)


ShaharNaveh (approved):
Do we need to adjust dependabot as well?

fanninpm:
> Do we need to adjust dependabot as well?

That will be the last commit in the series.

youknowone (approved):
👍

PR #7963, opened 2026-05-23
docs: expand AGENTS.md with test marker guidance and CI audit notes

JamesClarke7283:
## Summary

Four additive updates to `AGENTS.md` to capture guidance that contributors keep asking about. No existing wording was changed — everything is appended in place.

- **Test marker decision rule.** Documents the four forms (`@unittest.expectedFailure` / `expectedFailureIf` / `skip` / `skipIf`) with the `TODO: RUSTPYTHON; <reason>` format, and explains *when to use which*: prefer `expectedFailure` so a future fix surfaces an unexpected pass; reach for `skip` only when running the test would take down the test process (segfault, Rust panic, abort, hang).
- **Avoiding zizmor's `impostor-commit` audit.** Explains why hash-pinned `uses:` lines can still be rejected when the SHA only exists on a fork, the canonical-repo verification recipe (`gh api repos/<owner>/<repo>/commits/<sha>`), and the repo's existing convention of pinning to a full 40-char SHA with a trailing `# vX.Y.Z` comment.
- **Wiki link** to the "[How to update test files](https://github.com/RustPython/RustPython/wiki/How-to-update-test-files#checkout-cpython-source-code-initial-setup)" guide for syncing tests from upstream CPython.
- **Grep recipe** for locating partly-modified WIP test entries: `grep -d recurse 'TODO: RUSTPYTHON' Lib/test/`.

## Test plan

- [ ] Rendered `AGENTS.md` reads cleanly (headings, code fences, link targets)
- [ ] Existing sections (lines 1–266 in the prior version) are untouched
- [ ] No code or behavior changes — documentation only

## Summary by CodeRabbit

* **Documentation**
  * Added contributor guidance to run pre-commit checks and the full test suite before committing, including re-staging after auto-fixes.
  * Clarified when to use expected-failure vs skip markers for unit tests and how to find WIP test entries.
  * Added a docstring rule for items annotated for Python exposure (avoid `///` on those items).
  * Expanded CI notes to require workflow scanning and linked wiki guidance for syncing Lib tests from upstream.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7963?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -164,6 +164,7 @@ Run `./scripts/whats_left.py` to get a list of unimplemented methods, which is h
 
 - Do not delete or rewrite existing comments unless they are factually wrong or directly contradict the new code.
 - Do not add decorative section separators (e.g. `// -----------`, `// ===`, `/* *** */`). Use `///` doc-comments or short `//` comments only when they add value.
+- Do not put `///` doc comments on items annotated with `#[pyattr]`, `#[pyclass]`, or `#[pyfunction]`. The derive macros pull authoritative docstrings from CPython via the `rustpython-doc` crate; a Rust doc comment overrides that source, and on `#[pyattr]` it is silently dropped.
 
 #### Avoid Duplicate Code in Branches
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -38,6 +38,12 @@ RustPython is a Python 3 interpreter written in Rust, implementing Python 3.14.0
 - Always ask the user before performing any git operations that affect the remote repository
 - Commits can be created locally when requested, but pushing and PR creation require explicit approval
 
+**CRITICAL: Pre-commit Checks**
+- Before creating ANY commit, you MUST run `prek run --all-files` (or `pre-commit run --all-files`) AND the full test suite. Both must pass — do not commit if either fails.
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -258,9 +287,37 @@ See DEVELOPMENT.md "CPython Version Upgrade Checklist" section.
 - Document that it requires PEP 695 support
 - Focus on tests that can be fixed through Rust code changes only
 
+## CI Workflows and Security Audits
+
+PR CI runs [zizmor](https://docs.zizmor.sh/) to audit `.github/workflows/*.yaml` for security issues. The most common failure for contributors editing workflows is the `impostor-commit` audit, which can reject a PR even when the SHA pin "looks correct".
```
ShaharNaveh: I'm not so sure about this section. While it can surely happend, I've also seen it happend once, I don't think it's worth adding a dedicated section about it.

I would leave a section saying that if the agent would modify a workflow file, it needs to pass a zizmor scan

ShaharNaveh (changes requested):

AGENTS.md
JamesClarke7283: Good call — I've trimmed it to a single sentence in 9ea89fbc3 noting that any change under `.github/workflows/` must pass the zizmor scan in CI, and dropped the dedicated impostor-commit subsection.

ShaharNaveh (approved):

PR #7981, opened 2026-05-26
sqlite3: fix Blob.__setitem__ value range validation

ever0de:
Previously, assigning an out-of-range integer (negative or > 255) or an integer too large for i64 (e.g. 2**65) to a Blob index raised OverflowError instead of ValueError.

Mirror CPython's ass_subscript_index logic:
- Convert the value via to_i64(), treating any overflow as -1
- Validate the result is in [0, 255], raising ValueError("byte must be in range(0, 256)") otherwise
- Separate deletion error messages: "item deletion" for index, "slice deletion" for slice

Ref: https://github.com/python/cpython/blob/main/Modules/_sqlite/blob.c

----
### Problem

When assigning an out-of-range integer to a `Blob` index, RustPython raised `OverflowError` instead of `ValueError`:

```python
blob[0] = -1      # OverflowError: Python int too large to convert to Rust u8
blob[0] = 256     # OverflowError
blob[0] = 2**65   # OverflowError
```

CPython raises `ValueError: byte must be in range(0, 256)` for all three cases.

### Root Cause

The previous implementation used `try_to_primitive::<u8>(vm)?` to convert the assigned value, which calls Rust's type conversion and raises a generic `OverflowError` for any value that doesn't fit in `u8` — including negative numbers.

### Fix

Mirrors CPython's `ass_subscript_index` logic:

```c
// CPython blob.c
long val = PyLong_AsLong(value);
if (val == -1 && PyErr_Occurred()) {
    PyErr_Clear();
    val = -1;  // treat overflow as out-of-range
}
if (val < 0 || val > 255) {
    PyErr_SetString(PyExc_ValueError, "byte must be in range(0, 256)");
    return -1;
}
```

In Rust:
- Replace `try_to_primitive::<u8>()` with `as_bigint().to_i64().unwrap_or(-1)`
- Values that overflow `i64` (e.g. `2**65`) are treated as `-1`, which then fails the range check
- Raises `ValueError("byte must be in range(0, 256)")` for any value outside `[0, 255]`

Additionally, aligned the deletion error messages with CPython:
- `del blob[0]` → `"Blob doesn't support item deletion"`
- `del blob[5:10]` → `"Blob doesn't support slice deletion"`

### Changes
- _sqlite3.rs: Fix `Blob::ass_subscript` index branch
- test_dbapi.py: Remove `@unittest.expectedFailure` from `test_blob_set_item_error`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed single-byte BLOB assignment to properly validate byte values and handle integer overflow.
  * Improved BLOB slice assignment so unaffected bytes are preserved for complex slices, including negative-step writes.

* **Tests**
  * Added a regression test for extended-slice BLOB assignments to prevent future regressions.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7981?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/extra_tests/snippets/stdlib_sqlite.py
+++ b/extra_tests/snippets/stdlib_sqlite.py
@@ -53,3 +53,14 @@ class AggrText:
 cx.create_aggregate("aggtxt", 1, AggrText)
 cur.execute("select aggtxt(key) from foo")
 assert cur.fetchone()[0] == "341011"
+
+# Blob extended-slice assignment with negative step
+cx.execute("CREATE TABLE blobtest(b BLOB)")
+data = b"this blob data string is exactly fifty bytes long!"
+cx.execute("INSERT INTO blobtest(b) VALUES (?)", (data,))
+blob = cx.blobopen("blobtest", "b", 1)
+blob[9:0:-2] = b"12345"  # writes to indices 9, 7, 5, 3, 1
+actual = cx.execute("select b from blobtest").fetchone()[0]
+expected = b"t5i4 3l2b1" + data[10:]
+assert actual == expected, f"got {actual!r}, expected {expected!r}"
+blob.close()
--- a/extra_tests/snippets/stdlib_sqlite.py
+++ b/extra_tests/snippets/stdlib_sqlite.py
@@ -58,6 +58,7 @@ assert cur.fetchone()[0] == "341011"
 # Guard: CPython 3.11 has a SystemError bug with negative-step Blob slicing;
 # this test only runs on RustPython where the fix is being validated.
 import sys
+
 if sys.implementation.name == "rustpython":
     cx.execute("CREATE TABLE blobtest(b BLOB)")
     data = b"this blob data string is exactly fifty bytes long!"
--- a/extra_tests/snippets/stdlib_sqlite.py
+++ b/extra_tests/snippets/stdlib_sqlite.py
@@ -57,6 +57,7 @@ assert cur.fetchone()[0] == "341011"
 # Blob extended-slice assignment with negative step
 # Guard: CPython 3.11 has a SystemError bug with negative-step Blob slicing;
 # this test only runs on RustPython where the fix is being validated.
+# TODO: remove this once https://github.com/python/cpython/pull/150450 is released and RustPython CI uses it.
 import sys
 
 if sys.implementation.name == "rustpython":
--- a/extra_tests/snippets/stdlib_sqlite.py
+++ b/extra_tests/snippets/stdlib_sqlite.py
@@ -55,12 +55,16 @@ cur.execute("select aggtxt(key) from foo")
 assert cur.fetchone()[0] == "341011"
 
 # Blob extended-slice assignment with negative step
-cx.execute("CREATE TABLE blobtest(b BLOB)")
-data = b"this blob data string is exactly fifty bytes long!"
-cx.execute("INSERT INTO blobtest(b) VALUES (?)", (data,))
-blob = cx.blobopen("blobtest", "b", 1)
-blob[9:0:-2] = b"12345"  # writes to indices 9, 7, 5, 3, 1
-actual = cx.execute("select b from blobtest").fetchone()[0]
-expected = b"t5i4 3l2b1" + data[10:]
-assert actual == expected, f"got {actual!r}, expected {expected!r}"
-blob.close()
+# Guard: CPython 3.11 has a SystemError bug with negative-step Blob slicing;
+# this test only runs on RustPython where the fix is being validated.
+import sys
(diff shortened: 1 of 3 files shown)


ever0de:
> https://github.com/RustPython/RustPython/actions/runs/26428784978/job/77797608378?pr=7981

### Why the snippet test is guarded by `sys.implementation.name == "rustpython"`

CI was failing because CPython 3.14 also has a bug with negative-step `Blob` slice assignment.

Running `blob[9:0:-2] = b"12345"` under CPython 3.14 raises:

```python
SystemError: Negative size passed to PyBytes_FromStringAndSize
```

The root cause is in CPython's `Modules/_sqlite/blob.c`, inside `ass_subscript_slice`:
> https://github.com/python/cpython/blob/b87c99158c429219fae2d6a6c49076fea68a5cad/Modules/_sqlite/blob.c#L552

```c
// For blob[9:0:-2], PySlice_AdjustIndices gives start=9, stop=0, step=-2
PyObject *blob_bytes = read_multiple(self, stop - start, start);
//                                         0  -  9  = -9  ← negative!
// PyBytes_FromStringAndSize(NULL, -9) → SystemError
```

When `step` is negative, `stop < start`, so `stop - start` is negative, and passing that to `PyBytes_FromStringAndSize` triggers an internal assertion error.

This is a CPython-side bug — both CPython and RustPython had a broken implementation of negative-step `Blob` slice writes (CPython raises `SystemError`, RustPython previously panicked). This PR fixes the RustPython side. The snippet test validates RustPython's fix but cannot run on CPython until the upstream bug is resolved, hence the guard.

### Reference

1. https://docs.python.org/3/library/sqlite3.html#sqlite3.Blob
Use indices and [slices](https://docs.python.org/3/glossary.html#term-slice) for direct access to the blob data.
2. https://docs.python.org/3/library/stdtypes.html#typesseq-common
**`s[i:j:k]`** — The slice of s from i to j with step k is defined as the sequence of items with index `x = i + n*k` such that `0 <= n < (j-i)/k`. **When k is negative**, i and j are reduced to `len(s) - 1` if they are greater.
3. (Similar, but not the same.) https://github.com/python/cpython/issues/142787
4. https://github.com/python/cpython/issues/150449

extra_tests/snippets/stdlib_sqlite.py
```diff
@@ -53,3 +53,19 @@ def finalize(self):
 cx.create_aggregate("aggtxt", 1, AggrText)
 cur.execute("select aggtxt(key) from foo")
 assert cur.fetchone()[0] == "341011"
+
+# Blob extended-slice assignment with negative step
+# Guard: CPython 3.11 has a SystemError bug with negative-step Blob slicing;
+# this test only runs on RustPython where the fix is being validated.
+import sys
+
+if sys.implementation.name == "rustpython":
```
youknowone, line 63: we usually don't add rustpython-only code. is this CPython bug? Then is this reported to CPython?

ever0de: Yes, the negative-step slice crash is a CPython bug — I've filed it upstream as **[python/cpython#150449](https://github.com/python/cpython/issues/150449)** and submitted a fix PR ([python/cpython#150450](https://github.com/python/cpython/pull/150450)).

However, this PR also contains a few pre-existing RustPython-only bugs unrelated to that CPython issue:
- `blob[i] = v` was using `try_to_primitive::<u8>` which silently gave the wrong error for out-of-range values (e.g. `256` or `-1`), instead of CPython's `ValueError: byte must be in range(0, 256)`
- The item-deletion vs. slice-deletion error messages were swapped

Should I split this into two PRs — one for the CPython-mirrored negative-step fix (which could be reverted or marked `TODO: RUSTPYTHON` until upstream lands), and one for the RustPython-specific behavior alignment?

extra_tests/snippets/stdlib_sqlite.py
```diff
@@ -53,3 +53,19 @@ def finalize(self):
 cx.create_aggregate("aggtxt", 1, AggrText)
 cur.execute("select aggtxt(key) from foo")
 assert cur.fetchone()[0] == "341011"
+
+# Blob extended-slice assignment with negative step
+# Guard: CPython 3.11 has a SystemError bug with negative-step Blob slicing;
+# this test only runs on RustPython where the fix is being validated.
```
youknowone, line 59:
```suggestion
# this test only runs on RustPython where the fix is being validated.
# TODO: remove this once https://github.com/python/cpython/pull/150450 is released and RustPython CI uses it.
```

youknowone (approved):
Thanks!

PR #7984, opened 2026-05-26
Add support for creating functions with the c-api

bschoenmaeckers:
## Summary by CodeRabbit

* **New Features**
  * Added C API-compatible support for creating and invoking native Python-callable methods, including multiple calling conventions and bound-self support.

* **Bug Fixes**
  * Improved validation and UTF-8 handling for method names and documentation to reduce runtime errors.

* **Chores**
  * Updated dependencies and internal method-construction plumbing for more reliable native-callable method creation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/RustPython/RustPython/pull/7984?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3322,6 +3322,7 @@ dependencies = [
 name = "rustpython-capi"
 version = "0.5.0"
 dependencies = [
+ "bitflags 2.11.1",
  "num-complex",
  "pyo3",
  "rustpython-stdlib",
--- a/crates/capi/Cargo.toml
+++ b/crates/capi/Cargo.toml
@@ -12,6 +12,7 @@ license.workspace = true
 crate-type = ["cdylib", "rlib"]
 
 [dependencies]
+bitflags = { workspace = true }
 num-complex = { workspace = true }
 rustpython-vm = { workspace = true, features = ["threading", "compiler"] }
 rustpython-stdlib = {workspace = true, features = ["threading"] }
--- a/crates/capi/src/lib.rs
+++ b/crates/capi/src/lib.rs
@@ -18,6 +18,7 @@ pub mod floatobject;
 pub mod import;
 pub mod listobject;
 pub mod longobject;
+pub mod methodobject;
 pub mod object;
 pub mod pycapsule;
 pub mod pyerrors;
--- a/crates/capi/src/methodobject.rs
+++ b/crates/capi/src/methodobject.rs
@@ -30,7 +30,7 @@ pub union PyMethodPointer {
     ) -> *mut PyObject,
     pub PyCFunctionFast: unsafe extern "C" fn(
         slf: *mut PyObject,
-        args: *mut *mut PyObject,
+        args: *const *mut PyObject,
         nargs: isize,
     ) -> *mut PyObject,
     pub PyCFunctionFastWithKeywords: unsafe extern "C" fn(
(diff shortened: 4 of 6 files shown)


JamesClarke7283 (commented):
Nice groundwork on the C API surface. Build is clean and clippy passes the workspace lint set, which already includes pedantic and a fair bit of nursery. Most of what's left is semantic.

**1. VARARGS / FASTCALL paths strip a real user argument as `self` when the function has no bound self.**

```rust
PyMethodFlags::VARARGS => {
    let callable = move |args: PosArgs, vm: &VirtualMachine| unsafe {
        call_function(vm, method, flags, Some(args))
    };
    ...
},
```
```rust
fn take_self_arg(args: &mut FuncArgs, flags: PyMethodFlags) -> Option<PyObjectRef> {
    if flags.contains(PyMethodFlags::STATIC) { None } else { args.take_positional() }
}
```

`Callable::call` for `PyNativeFunction` only prepends `zelf` when `zelf.is_some() && !STATIC`. After `PyCFunction_New(ml, NULL)` with `METH_VARARGS`, calling `fn(a, b)` makes `take_self_arg` consume `a` as `self` and the C function sees `args = (b,)`. Silent data corruption with no error. Same shape for the three other branches at lines 95 to 118. `NOARGS` and `O` already split on `has_self`; the VARARGS / FASTCALL family needs the same treatment, or pass `has_self` into `take_self_arg`.

**2. `module` is silently dropped, so `__module__` is always `None` for anything built through this API.**

```rust
pub unsafe extern "C" fn PyCMethod_New(
    ml: *mut PyMethodDef,
    slf: *mut PyObject,
    _module: *mut PyObject,
    _cls: *mut PyTypeObject,
) -> *mut PyObject {
```

CPython does `op->m_module = Py_XNewRef(module)` and exposes it via `__module__`. Here it is bound to `_module` and never read. `HeapMethodDef::build_function` also lacks a `module` parameter, so even threading it through would not reach `PyNativeFunction.module`. Breaks `inspect.getmodule`, pickling and traceback module attribution.

**3. `assert!` and `.expect()` across an `extern \"C\"` boundary.**

```rust
with_vm(|vm| -> PyResult {
    assert!(
        _cls.is_null(),
        \"PyCMethod_New does not support METH_METHOD on abi3\"
    );
```
```rust
vm.take_raised_exception()
    .expect(\"Native function returned NULL, but there was no exception set\")
```

`with_vm` does not catch panics, and neither function is `extern \"C-unwind\"`, so a panic here aborts the process. Any caller that passes a non-null `cls` (the PEP 573 path with `METH_METHOD | METH_FASTCALL | METH_KEYWORDS`) takes the interpreter down. Likewise for any buggy extension that returns `NULL` without `PyErr_SetString`. Both should be `vm.new_system_error(...)` plus a `NULL` return, matching `_Py_CheckFunctionResult` in CPython.

**4. `PyMethodFlags::from_bits` rejects valid CPython flag combinations.**

```rust
let flags = PyMethodFlags::from_bits(ml.ml_flags as u32)
    .ok_or_else(|| vm.new_system_error(\"PyMethodDef contains unknown flags\"))?;
```

`METH_COEXIST (0x40)` and `METH_STACKLESS (0x100)` are commented out of `PyMethodFlags` in `crates/vm/src/function/method.rs`. CPython masks unknown bits and treats them as ignored, so a textbook pattern like `{\"__contains__\", impl, METH_O | METH_COEXIST, NULL}` (used by several stdlib C modules) fails registration outright. Either switch to `from_bits_truncate` or declare the bits.

**5. `CStr::from_ptr(ml.ml_name)` with no null check.**

```rust
let name = unsafe { CStr::from_ptr(ml.ml_name) }
    .to_str()
    .map_err(|_| vm.new_system_error(\"Method name was not valid UTF-8\"))?;
```

Null `ml_name` is UB. `ml_doc` is already guarded with `NonNull::new` two lines down, so this is also an internal inconsistency. Same guard plus a `SystemError`.

**Minor:** the `#[cfg(false)] mod tests` block at lines 309 to 369 is dead code that references `pyo3` types not declared as a dev dependency. Either wire it into CI or drop it.

youknowone (approved):

youknowone:
@JamesClarke7283 when you share AI generated review, please add a headline to specify it is generated review, not reviewed by you.

@bschoenmaeckers could you share how this c-api journey is going? what do you need more for pyo3 compatibility?

PR #7986, opened 2026-05-27
Fix stack overflow segfault on debug builds for Python threads

### LINKED ISSUE #7941: Some python tests are crashing with segmentation fault when running on debug build
## Summary

```
$ cargo run -- -m test.test_ssl ThreadedTests.test_socketserver
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.11s
     Running `target/debug/rustpython -m test.test_ssl ThreadedTests.test_socketserver`
test_ssl: testing with 'OpenSSL 3.3.0 (rustls/0.23)' (3, 3, 0, 0, 15)
          under 'Linux-6.19.14-400.asahi.fc44.aarch64+16k-aarch64-with-glibc2.43'
          HAS_SNI = True
          OP_ALL = 0x     bfb
          OP_NO_TLSv1_1 = 0x10000000

 server (('127.0.0.1', 34577):34577 ('TLS_AES_256_GCM_SHA384', 'TLSv1.3', 256)):
   [20/May/2026 23:23:31] "GET /test_ssl.py HTTP/1.1" 200 -
Segmentation fault         (core dumped) RUST_BACKTRACE=1 cargo run -- -m test.test_ssl ThreadedTests.test_socketserver
```

Following patch circumvents this, so I believe it is a stack overflow really:

```diff
diff --git a/Lib/test/ssl_servers.py b/Lib/test/ssl_servers.py
index 15b071e04dda..8d9331c6705f 100644
--- a/Lib/test/ssl_servers.py
+++ b/Lib/test/ssl_servers.py
@@ -125,6 +125,7 @@ def __init__(self, context, host=HOST, handler_class=None):
                                   handler_class or RootedHTTPRequestHandler,
                                   context)
         self.port = self.server.server_port
+        threading.stack_size(1024 * 1024 * 128)
         threading.Thread.__init__(self)
         self.daemon = True
```

## Expected

No segmentation fault.

## Actual

Segmentation fault.

## Python Documentation

N/A

JamesClarke7283:
## Summary

- Spawned Python threads were getting Rust's `std::thread::Builder` default stack of 2 MB, which is too small for the call chains the Python stdlib runs on helper threads in debug builds (where Rust stack frames are substantially larger). CPython on glibc Linux relies on pthread's ~8 MB default instead — a 4× difference that explains why this hits us but not CPython.
- Apply 8 MB in `apply_thread_stack_size` whenever the user hasn't explicitly set a value via `threading.stack_size(N)`. This matches CPython's effective default while keeping the Python API contract exact: `threading.stack_size()` still returns `0` ("platform default"), and explicit overrides still take precedence.
- Single-file change in `crates/vm/src/stdlib/_thread.rs`.

Closes #7941

## Test plan

- [x] `cargo build` succeeds.
- [x] Original reproducer passes on a debug build: `cargo run -- -m test.test_ssl ThreadedTests.test_socketserver` → `Ran 1 test in 21.59s, OK` (was segfaulting on `main`).
- [x] Python API contract preserved: `threading.stack_size()` returns `0` by default; setting to 1 MiB then reading back returns `1048576`; previous value reported correctly.
- [x] Basic threaded workload (5 threads doing work + join) still completes normally.

## Summary by CodeRabbit

* **Bug Fixes**
  * Made thread stack sizing more predictable in debug builds by using a consistent fallback when no size is configured.
  * Applied configured stack sizes to newly spawned threads.
  * Release builds continue using the platform default when no stack size is set.


youknowone:
please also add failing test before patch

JamesClarke7283:
> please also add failing test before patch

Will do

JamesClarke7283:
> please also add failing test before patch

Which test? the original one from the original issue that now passes or something different.
Please clarify.

youknowone:
please add a test, which failed before the patch but success after the patch. If adding patch is not possible, please share discussion why it is impossible

JamesClarke7283:
> please add a test, which failed before the patch but success after the patch. If adding patch is not possible, please share discussion why it is impossible

Done

